PR #980, opened 2026-05-14
v0.32.4.1 fix: extract_facts entity_hints array schema + structural guard

aadachi:
## Summary

Hot-fix follow-up to v0.32.4. `extract_facts.entity_hints` declared `type: 'array'` without `items` — JSON Schema requires arrays to declare their item type, and both OpenAI + Anthropic function-calling APIs reject malformed tool schemas with **HTTP 400** the moment any MCP client loads the gbrain HTTP MCP and tries to use any tool. The malformed schema poisons the entire `tools[]` array submitted to the chat provider, so the error isn't even attributable to gbrain at first glance.

The visible failure (reported from a Hermes session against a remote `gbrain serve --http`):

```
Non-retryable error (HTTP 400): Invalid schema for function
'mcp_gbrain_extract_facts': In context=('properties', 'entity_hints'),
array schema missing items.
```

## What changed

- **`src/core/operations.ts:2459`** — `entity_hints` now declares `items: { type: 'string' }`. Matches the handler signature (`p.entity_hints as string[]`) and the existing-good pattern at line 1755 (`pages_updated`).
- **`test/mcp-tool-defs.test.ts`** — structural guard walks every emitted MCP tool def and asserts that any property with `type: 'array'` declares `items`. Runs at the schema-output layer (post-`buildToolDefs`), so future regressions at either the op-def layer OR the schema-builder layer fail the unit suite.

Bug introduced in v0.31 ("Hot memory ops: extract_facts / recall / forget_fact"). All v0.31.x and v0.32.x releases prior to 0.32.4.1 have the broken schema and are unusable over HTTP MCP for any client that strictly validates tool definitions (Hermes, ChatGPT desktop, etc.).

## Test plan

- [x] `bun test test/mcp-tool-defs.test.ts` — 6/6 pass (5 prior + 1 new structural guard)
- [x] `bun run verify` — all 6 pre-checks pass (wasm, admin-build, scope-drift, cli-executable, system-of-record) + tsc
- [x] Full unit suite (`bun run test`): 5582 pass / 9 fail — all 9 failures confirmed pre-existing on master (graph-query.test.ts + doctor-fix.test.ts + run-unit-parallel wrapper tests + 4 shard timeouts unrelated to this diff)
- [ ] After merge, remote `gbrain serve --http` operators run `gbrain upgrade && pkill -f 'gbrain serve --http' && gbrain serve --http &` to pick up the corrected schema. Clients (Hermes / Claude Desktop / etc.) succeed on the first `tools/list` post-restart.


garrytan:
Thanks @aadachi — `extract_facts.entity_hints` (and the other MCP array-schema `items` fixes) already ship in master as of **v0.35.3.0**. The shared mapper is `paramDefToSchema` in `src/mcp/tool-defs.ts` (used by stdio MCP, HTTP MCP, and the subagent registry); `test/mcp-tool-defs.test.ts` walks every array param and fails the suite if any lacks `items.type`. If your install is still hitting this on next `gbrain upgrade`, please reopen with the output of `gbrain doctor --json`.

PR #1002, opened 2026-05-14
fix: supervisor treats code=0 watchdog exits as crashes

garrytan-agents:
## Problem

The RSS watchdog triggers `gracefulShutdown()` which exits with code 0. The supervisor was counting ALL exits < 5min as crashes, including clean code=0 exits. After 10 watchdog-triggered restarts, the supervisor gave up with `max_crashes_exceeded`.

On a 96K-page brain, `process.memoryUsage().rss` reports ~7GB during autopilot sync (likely shared page inflation from git mmap operations). The 4096MB threshold triggers on every autopilot cycle, causing the worker to exit cleanly after ~5 min. The supervisor then counts this as a crash.

**Result:** 73 "crashes" in 24h, all code=0. Supervisor keeps dying and being restarted by external crons.

## Fix

- code=0 exits reset `crashCount` to 0 and restart immediately (no backoff)
- Only code≠0 exits count toward the crash limit
- Clean exits skip the backoff delay entirely

## Separate Issue (not addressed here)

`process.memoryUsage().rss` inflates to 7GB during sync on large repos. The 4096MB watchdog threshold triggers every cycle. This may need a separate fix (use `/proc/self/statm` for accurate RSS, or raise the threshold for large repos).

### DIFF
--- a/src/core/minions/supervisor.ts
+++ b/src/core/minions/supervisor.ts
@@ -408,11 +408,17 @@ export class MinionSupervisor {
         return;
       }
 
+      // Clean exits (crashCount=0) restart immediately with no backoff.
       // crashCount - 1 is the retry-attempt index (0-based exponent for backoff math).
       // On first crash: crashCount=1, backoff exponent=0 → 1s.
       // After stable-run reset: crashCount=1 again → 1s fresh cycle.
       // Test-only: _backoffFloorMs short-circuits to a fixed tiny value so integration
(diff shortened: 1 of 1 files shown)


garrytan:
Moving to base-repo branch for CI secret access (fork PRs from non-collaborator accounts don't receive base-repo secrets). Replacement PR coming up — same title, same body, same commits.

PR #1004, opened 2026-05-14
fix: use RssAnon instead of VmRSS for watchdog threshold

garrytan-agents:
## Problem

`process.memoryUsage().rss` returns VmRSS which includes **file-backed mmap pages**. On repos with large git packfiles (96K+ pages), git operations inflate VmRSS to **7GB+** while actual heap usage is **~100MB**.

The kernel reclaims file-backed pages under memory pressure — they are cache, not real usage. But the watchdog sees 7GB and triggers `gracefulShutdown()` every single autopilot cycle.

## Measurement

| Metric | During git sync | At idle |
|--------|----------------|--------|
| `process.memoryUsage().rss` (VmRSS) | **7,023 MB** | 90 MB |
| `RssAnon` (actual heap) | ~200 MB | 48 MB |
| `RssFile` (file cache) | ~6,800 MB | 42 MB |

## Fix

Replace `process.memoryUsage().rss` with a `getAccurateRss()` helper that reads `/proc/self/status` for `RssAnon + RssShmem` — the anonymous pages that represent actual memory allocation.

Falls back to `process.memoryUsage().rss` on non-Linux (macOS, Windows).

**Before:** Watchdog triggers every autopilot cycle (7GB VmRSS > 4GB threshold)
**After:** Watchdog only triggers on real memory growth (~100MB << 4GB threshold)

## Related

- #1002 — supervisor crash-count fix for the downstream symptom (counting code=0 exits as crashes)

### DIFF
--- a/src/core/minions/supervisor.ts
+++ b/src/core/minions/supervisor.ts
@@ -408,11 +408,17 @@ export class MinionSupervisor {
         return;
       }
 
+      // Clean exits (crashCount=0) restart immediately with no backoff.
       // crashCount - 1 is the retry-attempt index (0-based exponent for backoff math).
       // On first crash: crashCount=1, backoff exponent=0 → 1s.
       // After stable-run reset: crashCount=1 again → 1s fresh cycle.
       // Test-only: _backoffFloorMs short-circuits to a fixed tiny value so integration
       // tests can exercise crash loops in < 1s without waiting for the real curve.
+      if (this.crashCount === 0) {
+        // Clean exit — restart immediately, no backoff
+        this.emit('backoff', { ms: 0, crash_count: 0 });
+        continue;
+      }
       const backoff = this.opts._backoffFloorMs !== undefined
         ? this.opts._backoffFloorMs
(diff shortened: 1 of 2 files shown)


garrytan:
Superseded by #1003 fix wave.

The supervisor commit (d741574e) from this PR duplicated #1003's existing commit. The RssAnon watchdog fix (b81c598f) has been cherry-picked into #1003 (now landed as dab48cdf on that branch).

During plan-eng-review on #1003, the fix wave grew to also:
- preserve flap detection via lastExitCode tracking instead of resetting crashCount on every code=0 exit
- add a clean-restart budget for the macOS / kernel-<4.5 fallback path
- refactor autopilot.ts to share the supervisor core (DRY — Codex caught that autopilot had a parallel implementation with the same bug class)
- fix a field-presence parser bug Codex surfaced in getAccurateRss

Co-Authored-By trailers preserved on the cherry-pick. Thank you for the root-cause diagnosis — RssAnon + RssShmem is the right metric for per-process leak detection.

PR #1007, opened 2026-05-15
rfc: temporal axis for contradiction probe

garrytan:
## What

A report from resolving ~115 HIGH contradiction findings in a production gbrain run, plus a proposal for a 4-phase fix. Documents why a majority of residual findings are temporal false positives and proposes a time-axis enhancement to the contradiction probe.

## Why

The probe treats all claims as timeless. A status change recorded as "trial" in April and "confirmed" in May gets flagged as a contradiction. A role transition recorded in 2017 vs. an updated role in 2025 gets flagged. These are features of a brain that records history, not bugs.

## Use Cases (synthetic placeholders in the RFC text)

1. **Founder tracking** — claimed vs. realized metrics across office hours. Detect regressions, score prediction accuracy, surface narrative drift.
2. **Event disambiguation** — two distinct events within a short window. Two events, not one contradiction.
3. **Role/status changes** — people change roles. Historical records aren't wrong; they're history.
4. **Decision evolution** — multi-step decisions where each step supersedes the previous.

## Proposed Phases

1. Judge prompt enhancement (smallest change, biggest impact — pass dates to judge, add temporal_supersession verdict)
2. Claim trajectory view (e.g. `gbrain eval trajectory "Acme Corp MRR"`)
3. Automatic valid_from/valid_until population during extract_facts
4. Founder scorecard

## Stats

- ~107K pages, ~257K chunks
- ~115 HIGH findings → ~0 genuine contradictions remaining after manual resolution
- A majority of original findings were temporal false positives

This is a report/RFC, not a code change. Phase 1 is the recommended starting point.

### DIFF
--- a/docs/proposals/temporal-contradiction-probe.md
+++ b/docs/proposals/temporal-contradiction-probe.md
@@ -1,0 +1,213 @@
+# Proposal: Temporal Axis for Contradiction Probe
+
+**Status:** Report / RFC
+**Date:** 2026-05-14
+**Context:** A large production run of `gbrain eval suspected-contradictions` surfaced ~115 HIGH findings. Walking through them by hand exposed a structural limitation in the probe.
+
+## The Problem
+
+The contradiction probe (`gbrain eval suspected-contradictions`) treats all claims as timeless. When two chunks make conflicting statements, the judge flags a contradiction regardless of whether both statements were true at their respective points in time.
+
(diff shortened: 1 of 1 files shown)


garrytan:
Closing — branch renamed to drop a private agent name from the head ref per privacy rule. Same commit, same content, reopened as a fresh PR.

PR #1009, opened 2026-05-15
fix(entities): prefix-expansion resolve for bare first names + stub guard

garrytan-agents:
## Problem

resolveEntitySlug(Jared) fell through to slugify → bare jared because pg_trgm scores too low on short strings. writeFactsToFence then stub-created a phantom people/jared.md instead of routing to people/jared-friedman.

## Three-layer fix

1. **Prefix expansion** in resolveEntitySlug: queries people/token-% then companies/token-%, picks highest-connection match
2. **Stub creation guard** in fence-write: refuses to create pages with no directory prefix
3. **Legacy fallback** in backstop: routes blocked facts to DB-only path

## Tests

13 new tests, all passing. Jared→jared-friedman, Diana→diana-hu, Sam→sam-altman, etc.


garrytan:
Moving to base-repo branch so CI secrets are available (garrytan-agents is a fork; pull_request workflows from forks don't get base-repo secrets). Replacement PR coming next from the same branch pushed to garrytan/gbrain.

Issue #935, opened 2026-05-12
OAuth verifyAccessToken returns 500 instead of 401 on invalid bearer token (v0.33.0)


xaviroblessarries:
## Bug

`verifyAccessToken` in `src/core/oauth-provider.ts` throws plain `Error` instances when a bearer token is invalid or expired. The MCP SDK's `requireBearerAuth` middleware (`@modelcontextprotocol/sdk/server/auth/middleware/bearerAuth.js`) classifies thrown errors:
- `InvalidTokenError` → HTTP 401 + WWW-Authenticate (per RFC 6749 / RFC 6750)
- `InsufficientScopeError` → HTTP 403
- `ServerError` → HTTP 500
- **Anything else (plain `Error`) → HTTP 500 with `server_error`**

Because gbrain throws plain `Error`, bearer-token validation failures return HTTP 500 `server_error` instead of the canonical RFC 6750 401 `invalid_token`.

## Reproducer

```bash
# Server v0.33.0 running (postgres engine, OAuth clients registered)
curl -X POST https://your-gbrain/mcp \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer invalid_token" \
  -d '{"jsonrpc":"2.0","id":1,"method":"tools/list"}'

# Actual:
# {"error":"server_error","error_description":"Internal Server Error"}
# HTTP/2 500

# Expected (per RFC 6750 §3.1):
# {"error":"invalid_token","error_description":"Invalid token"}
# HTTP/2 401
# WWW-Authenticate: Bearer error="invalid_token", error_description="Invalid token"
```

Verified on both local (macOS) and remote (Railway) v0.33.0 deployments.

## Root cause

`src/core/oauth-provider.ts`:
- Line ~423: `throw new Error('Token expired');`
- Line ~458: `throw new Error('Invalid token');`

Both inside `verifyAccessToken`. SDK `bearerAuth.js:53-69` catch block routes plain `Error` to the `else` branch:

```js
else {
  const serverError = new ServerError('Internal Server Error');
  res.status(500).json(serverError.toResponseObject());
}
```

## Suggested fix

```diff
 import type { AuthInfo } from '@modelcontextprotocol/sdk/server/auth/types.js';
+import { InvalidTokenError } from '@modelcontextprotocol/sdk/server/auth/errors.js';
 import { hashToken, generateToken, isUndefinedColumnError } from './utils.ts';

   ...
   const expiresAt = coerceTimestamp(row.expires_at);
[...]

PR #1012, opened 2026-05-15
fix(oauth): throw InvalidTokenError so bearerAuth returns 401, not 500

Aashiqe10:
## Problem

`verifyAccessToken` in `src/core/oauth-provider.ts` throws plain `Error` for expired/invalid tokens:

```ts
throw new Error('Token expired')
throw new Error('Invalid token')
```

The MCP SDK's `requireBearerAuth` middleware uses `instanceof` checks to map errors to HTTP responses — only `InvalidTokenError`/`InsufficientScopeError`/`OAuthError` map to proper status codes. Plain `Error` falls through to the catch-all, yielding a **500** with an OAuth-shaped body:

```json
{"error":"server_error","error_description":"Internal Server Error"}
```

## Impact for clients

MCP clients (e.g. rmcp / codex CLI) cannot deserialize that as JSON-RPC and surface cryptic errors like:

`Deserialize error: data did not match any variant of untagged enum JsonRpcMessage, when send initialize request`

This looks like a server crash, not an auth issue. Diagnosed during a real incident where a stale bearer token cost ~6h of debugging before the underlying cause (token expiry) was identified.

## Reproduction

```bash
curl -X POST -H "Authorization: Bearer expired_or_invalid_token" \\
     -H "Content-Type: application/json" \\
     -H "Accept: application/json, text/event-stream" \\
     --data '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-06-18","capabilities":{},"clientInfo":{"name":"x","version":"0"}}}' \\
     https://<your-gbrain>/mcp
# Currently: HTTP 500 {"error":"server_error","error_description":"Internal Server Error"}
# After fix: HTTP 401 with WWW-Authenticate: Bearer error="invalid_token", error_description="Token expired"
```

## Fix

Two lines + one import (see diff). Replace `new Error` with `new InvalidTokenError` from `@modelcontextprotocol/sdk/server/auth/errors.js`.

## No behaviour change for valid tokens

Only error paths affected. Tested against a live v0.31.3 deployment.

### DIFF
--- a/src/core/oauth-provider.ts
+++ b/src/core/oauth-provider.ts
@@ -22,6 +22,7 @@ import type {
 import type { OAuthServerProvider, AuthorizationParams } from '@modelcontextprotocol/sdk/server/auth/provider.js';
 import type { OAuthRegisteredClientsStore } from '@modelcontextprotocol/sdk/server/auth/clients.js';
 import type { AuthInfo } from '@modelcontextprotocol/sdk/server/auth/types.js';
+import { InvalidTokenError } from '@modelcontextprotocol/sdk/server/auth/errors.js';
 import { hashToken, generateToken, isUndefinedColumnError } from './utils.ts';
 import { hasScope, assertAllowedScopes, parseScopeString, InvalidScopeError } from './scope.ts';
 import type { SqlQuery, SqlValue } from './sql-query.ts';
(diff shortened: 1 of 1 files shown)


garrytan:
Thanks for this, @Aashiqe10 — closing as superseded. Your work shipped in **v0.36.1.1**. From the [CHANGELOG](https://github.com/garrytan/gbrain/blob/master/CHANGELOG.md):\n\n> -  —  throws  on expired + invalid bearer; SDK middleware returns 401 not 500. Closes #935. PR #1012.\n\nReal appreciation for the contribution.

PR #1013, opened 2026-05-15
Route dream synthesis through gateway and add embedding dim migration

cxbitz:
## Summary
- Route dream significance/synthesis subagent paths through the existing AI gateway so LiteLLM/OpenAI-compatible routes work without direct Anthropic dependencies.
- Add Ollama bge-m3 embedding support with model-specific 1024d defaults and provider probes.
- Add `gbrain migrate-embedding-dim` to resize existing vector schemas, clear incompatible embeddings/query cache, update config, and re-embed safely.

## Verification
- `git diff --check`
- `bun run typecheck`
- `bun test test/embedding-dim-check.test.ts test/subagent-handler.test.ts test/cycle-synthesize.test.ts test/ai/gateway-chat.test.ts test/ai/gateway.test.ts test/ai/recipes-existing-regression.test.ts test/ai/no-batch-cap-suppression.serial.test.ts` (132 pass / 0 fail)
- Isolated live Ollama flow: 768d init -> migrate to bge-m3 1024d -> dream embed -> query -> stale dry-run clean


time-attack:
Thanks for putting this together. It's still a draft and it bundles at least three independent changes — routing dream synthesis through the gateway, a new embedding-dimension migration command, and provider recipe additions — across 20+ files, which makes it effectively unreviewable as one unit. On top of that, the pieces are feature surface rather than fixes for something currently broken. Closing.

PR #741, opened 2026-05-08
fix: bootstrap forward-references for v39-v41 schema replay

lanceretter:
## Summary

Three column-with-index forward references in the embedded schema blob were missing from `applyForwardReferenceBootstrap`, so any brain at `config.version < 39` (Postgres) or `< 41` (PGLite) wedges before the migration runner can advance.

Reproduced end-to-end on a PlanetScale Postgres brain stuck at `config.version=34` trying to upgrade to `v0.30.0`:

```
ERROR: column "effective_date" does not exist
ERROR: column cc.modality does not exist
```

After hitting these errors, `gbrain search` and `gbrain reindex-frontmatter` both fail because the same columns are referenced by query-path code, but you can't get past the migration step to add them.

## The schema-blob references that crash before migrations run

- **v39** (`multimodal_dual_column_v0_27_1`):
  ```sql
  CREATE INDEX idx_chunks_embedding_image
    ON content_chunks USING hnsw (embedding_image vector_cosine_ops)
    WHERE embedding_image IS NOT NULL;
  ```
- **v41** (`pages_recency_columns`):
  ```sql
  CREATE INDEX pages_coalesce_date_idx
    ON pages ((COALESCE(effective_date, updated_at)));
  ```

## What's in this PR

- **Postgres engine** (`src/core/postgres-engine.ts`): probe + branch for **v39** (modality, embedding_image) — was entirely missing on Postgres, so Postgres brains pre-v39 hit the wedge that PGLite already protected against. PGLite's probe at line 273 had `embedding_image_exists`; Postgres did not.
- **Both engines**: probe + branch for **v40+v41**. Bootstraps all five additive pages columns (`emotional_weight`, `effective_date`, `effective_date_source`, `import_filename`, `salience_touched_at`) gated on `effective_date_exists` as the proxy.
- **`test/schema-bootstrap-coverage.test.ts`**: extends `REQUIRED_BOOTSTRAP_COVERAGE` with the six new columns AND the pre-test DROP block so both the per-target assertion test and the end-to-end "bootstrap + SCHEMA_SQL replay" test exercise the new coverage.

Bootstrap stays additive-columns-only. Indexes are left to schema replay / migrations as before — adding indexes in bootstrap duplicates responsibilities and worsens lock/surprise risk.

## Test plan

- [x] `bun test test/schema-bootstrap-coverage.test.ts` — all 5 tests pass (was 5 before, still 5; expanded coverage caught no regressions in the existing assertions).
[...]

PR #682, opened 2026-05-06
fix(engines): pre-add v0.20 + v0.26.3 forward-reference columns in bootstrap

brandonlipman:
## What

`applyForwardReferenceBootstrap` (both PostgresEngine + PGLiteEngine) covered v0.18 + v0.19 + v0.26.5 forward-references but missed two later groups. Brains upgrading from v0.13/v0.14-era directly to current master crash before the migration ladder runs:

**1. v0.20 Cathedral II — `content_chunks` columns**
- `search_vector`, `parent_symbol_path`, `doc_comment`, `symbol_name_qualified`
- `CREATE INDEX idx_chunks_search_vector` and `CREATE INDEX idx_chunks_symbol_qualified` in `schema.sql:141-143` / `pglite-schema.ts` reference these and crash with `column "search_vector" does not exist` / `column "symbol_name_qualified" does not exist`.

**2. v0.26.3 — `mcp_request_log` columns**
- `agent_name`, `params`, `error_message`
- `CREATE INDEX idx_mcp_log_agent_time ON mcp_request_log(agent_name, ...)` in `schema.sql:420` crashes with `column "agent_name" does not exist`.

## Repro

Brain on schema v13/v14 + run `gbrain init --migrate-only` against current master:

```
NOTICE:  relation "content_chunks" already exists, skipping
NOTICE:  relation "idx_chunks_embedding" already exists, skipping
column "search_vector" does not exist
```

Hit on a real 10K-page Supabase brain upgrading 0.14.2 → 0.27.0. Reproducible in PGLite via the test changes in this PR.

## Fix

Both engines now probe for the missing columns and pre-add them via `ALTER TABLE ADD COLUMN IF NOT EXISTS` before SCHEMA_SQL runs. Same shape as the existing `needsChunksBootstrap` / `needsPagesDeletedAt` patterns. The actual migrations (v26, v27, v33) still run later via `runMigrations` and remain idempotent — they handle backfill on top of the bootstrap-added columns.

## Test coverage

`test/schema-bootstrap-coverage.test.ts` — `REQUIRED_BOOTSTRAP_COVERAGE` now lists 6 new forward-references (4 v0.20 + 1 v0.26.3 in `mcp_request_log` + 1 docstring sweep). The strip-and-rebuild block drops the corresponding indexes/triggers so the test exercises a brain that pre-dates v0.20 + v0.26.3.

```
$ bun test test/schema-bootstrap-coverage.test.ts
2 pass, 0 fail, 12 expect() calls
```

[...]

PR #627, opened 2026-05-04
fix(bootstrap): cover v0.26.3 mcp_request_log columns

ChenyqThu:
## Summary

`applyForwardReferenceBootstrap()` in `src/core/postgres-engine.ts` + `src/core/pglite-engine.ts` covered `pages.deleted_at` (v0.26.5) but missed the v0.26.3 columns on `mcp_request_log` that SCHEMA_SQL forward-references via `CREATE INDEX idx_mcp_log_agent_time ON mcp_request_log(agent_name, created_at DESC)` (`src/schema.sql:420`).

Brains at schema v31 (any v0.22.7+ install that hadn't yet run v0.26.3's migration v33) crash during `gbrain init --migrate-only` with:

```
column "agent_name" does not exist
```

…before migration v33 (`admin_dashboard_columns_v0_26_3`) has a chance to ADD COLUMN. This is exactly the failure mode the bootstrap is structurally meant to prevent (see `test/schema-bootstrap-coverage.test.ts` preamble: #239, #243, #266, #357, #366, #374, #375, #378, #395, #396).

Hit this in the wild on a 2477-page production Postgres brain upgrading from v0.25.0 → v0.26.7. Manual unblock: `ALTER TABLE mcp_request_log ADD COLUMN IF NOT EXISTS agent_name TEXT, params JSONB, error_message TEXT;` — then `gbrain init --migrate-only` proceeds normally and migration v33's idempotent ADD COLUMN no-ops.

## Changes

- `src/core/postgres-engine.ts`: extend the bootstrap probe to detect `mcp_request_log` + the three v0.26.3 columns (`agent_name`, `params`, `error_message`); if `mcp_request_log` exists but any column is missing, ALTER TABLE ADD COLUMN IF NOT EXISTS the missing ones. Migration v33 still runs after and remains idempotent.
- `src/core/pglite-engine.ts`: mirror.
- `test/schema-bootstrap-coverage.test.ts`: extend `REQUIRED_BOOTSTRAP_COVERAGE` with 3 entries (one per new column); both existing test cases drop the new columns + indexes before running bootstrap, so the contract is enforced.
- `test/e2e/postgres-bootstrap.test.ts`: new regression case that mutates a fresh-LATEST brain to a v31 `mcp_request_log` shape (drops 3 columns + 2 indexes), asserts `initSchema()` reaches LATEST_VERSION and bootstraps the columns.

## Test plan

- [x] `bun run typecheck` clean.
[...]

PR #1016, opened 2026-05-15
fix(ai/recipes): declare max_batch_tokens on google embedding recipe

ChenyqThu:
## Why

`google` is the only first-party embedding recipe still missing `max_batch_tokens` after v0.32 (#779) landed the once-per-process startup warning. Operators routing through `google:gemini-embedding-001` (the default-provider path after v0.27 native gateway) see the warning on every `gbrain query`, every kos-compat-api `/ingest` response, and every cron `gbrain` invocation:

```
[ai.gateway] recipe "google" declares an embedding touchpoint without max_batch_tokens; recursion is the only safety net for batch caps.
```

For CJK-dense or large-payload batches the absent field also forces the gateway to discover Google's per-request token cap reactively via recursive halving instead of pre-splitting — one wasted HTTP round-trip per oversized batch.

The existing test `test/ai/no-batch-cap-suppression.serial.test.ts` even explicitly pins this state with the comment _"google should warn (it has fixed-cap models)"_, framing it as a known TODO.

## Patch

Three field additions to `google.touchpoints.embedding`:

```ts
max_batch_tokens: 20_000,
chars_per_token: 2,
// safety_factor left at gateway default 0.8
```

### Rationale

- **`max_batch_tokens: 20_000`** — Google's documented per-text cap is 2 048 tokens; ~20 k tokens per request is the soft cap before `gemini-embedding-001` starts emitting 429s.
- **`chars_per_token: 2`** — English averages ~4 chars/token (OpenAI default), CJK ~1.5. Picking 2 reflects realistic density on mixed-corpora brains so pre-split stays safe on CJK-heavy payloads without over-shrinking on English.
- **`safety_factor`** — left at gateway default 0.8. Pre-split lands at `20 000 × 0.8 / 2 = 8 000` chars/batch.

## Test follow-through

Two regression tests pinned `google` as the canary "real provider with no cap declared":

- `test/ai/no-batch-cap-suppression.serial.test.ts`: assumed google STILL warned. With google capped, the test flips to assert the stronger invariant — **no first-party recipe warns**, because every native/openai-compat recipe now declares either `max_batch_tokens` or `no_batch_cap`.
- `test/ai/adaptive-embed-batch.test.ts`: checked `contractMatch.length > 0`. After this patch the canary set is empty → `toBe(0)`. The once-per-process suppression mechanism is still exercised by the `firstCallCount` stability check earlier in the same test.

Both tests still gate the suppression machinery; they just no longer require any first-party recipe to be the canary. If maintainers prefer a synthetic test-only uncapped recipe to keep a positive "warning fires when expected" assertion, happy to add one in a follow-up.

## Validation

- `bun run typecheck` clean
- `bun test test/ai/` — 144 pass / 0 fail (was 142 pass / 2 fail pre-patch, expected: the two tests above)

## Field repro

Filed downstream from the Jarvis KOS v2 fork (`ChenyqThu/jarvis-knowledge-os-v2`), where the google native gateway is the default embedding provider since the v0.27 M3 cutover. The warning has been load-bearing log noise for ~6 weeks across kos-compat-api responses, cron logs, and operator queries. Same pattern as #627 / upstream fixwave #682+#741 (forward-bootstrap mcp_request_log) — small fork-local patch carried while upstream evaluates.

### DIFF
--- a/src/core/ai/recipes/google.ts
+++ b/src/core/ai/recipes/google.ts
@@ -14,6 +14,13 @@ export const google: Recipe = {
       models: ['gemini-embedding-001'],
       default_dims: 768,
       dims_options: [768, 1536, 3072],
+      // Per-text cap is 2048 tokens; per-request total ~20k tokens before
+      // gemini-embedding-001 starts 429-ing. chars_per_token: 2 covers
+      // CJK-dense corpora (English averages ~4, CJK ~1.5); safety_factor
+      // stays at the gateway default 0.8 so pre-split lands at ~8 000
+      // chars/batch — well under any per-request floor Google publishes.
+      max_batch_tokens: 20_000,
+      chars_per_token: 2,
       cost_per_1m_tokens_usd: 0.15,
       price_last_verified: '2026-04-20',
     },
--- a/test/ai/adaptive-embed-batch.test.ts
+++ b/test/ai/adaptive-embed-batch.test.ts
@@ -375,16 +375,22 @@ describe('startup warning for recipes missing max_batch_tokens', () => {
       console.warn = original;
     }
 
-    // The warning text should match the documented contract.
+    // The warning text contract is still documented; after v0.34.5 (google
+    // declared max_batch_tokens) every first-party native/openai-compat
+    // recipe is capped, so the canary set is empty. The mechanism itself
+    // is still exercised by the once-per-process suppression check above
+    // (firstCallCount stability across re-configure).
     const contractMatch = warnings.filter(w =>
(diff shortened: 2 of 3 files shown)


time-attack:
Thanks — you're right that `google` is still the uncapped first-party embedding recipe on master. We're passing on this one though: the warning is firing correctly by design (recursive halving is the documented safety net), so nothing is actually broken, and the fix hard-codes a vendor cap we can't verify from inside the repo. The inline rationale also miscomputes the effect — `chars_per_token` is a divisor in `splitByTokenBudget`, so 20_000 * 0.8 * 2 is ~32k chars per batch, not the ~8 000 the comment states. Closing.

Issue #974, opened 2026-05-13
Bootstrap gap: applyForwardReferenceBootstrap misses files.source_id + files.page_id (v0.18 forward-refs)


colin477:
## Problem

Migrating a pre-v0.18 brain (in my case, schema version 13 from gbrain v0.14.2) to v0.33.x fails during `init --migrate-only` with:

```
column "page_id" does not exist
```

The wedge fires inside `SCHEMA_SQL` apply, immediately after `idx_files_page` creation. `applyForwardReferenceBootstrap` covers most v0.13–v0.27 forward-references but doesn't pre-add the two columns `files` gained in v0.18.0 Step 7.

## Reproduction

Any brain at schema version <= ~v17 (no `files.source_id` / `files.page_id`) running `init --migrate-only` against `src/core/schema-embedded.ts` from v0.33.x will hit this. Confirmed today on v0.33.1.1 and v0.33.2.0 (same code path).

## Root cause

`applyForwardReferenceBootstrap` docstring in `src/core/postgres-engine.ts` lines 273–296 enumerates what it covers:

> - `sources` table + default seed (FK target of pages.source_id) — v0.18
> - `pages.source_id` column (indexed by `idx_pages_source_id`) — v0.18
> - `links.link_source` column (indexed by `idx_links_source`) — v0.13
> - `links.origin_page_id` column (indexed by `idx_links_origin`) — v0.13
> - `content_chunks.symbol_name` column ... — v0.19
> - ... etc

The two `files` columns added in the same v0.18.0 step as `pages.source_id` are missing:

- `files.source_id TEXT NOT NULL DEFAULT 'default' REFERENCES sources(id) ON DELETE CASCADE`
- `files.page_id INTEGER REFERENCES pages(id) ON DELETE SET NULL`

Both are forward-referenced by unguarded `CREATE INDEX` statements in `src/core/schema-embedded.ts:493-494`:

```sql
CREATE INDEX IF NOT EXISTS idx_files_page_id ON files(page_id);
CREATE INDEX IF NOT EXISTS idx_files_source_id ON files(source_id);
```

`CREATE INDEX IF NOT EXISTS` checks for the index name, not the column — so it tries to index a column that doesn't exist yet on old brains.

## Fix sketch

Three additions to `applyForwardReferenceBootstrap` (mirror in `PGLiteEngine` per the existing pattern):

1. Probe existence:
   ```ts
   files_exists, files_source_id_exists, files_page_id_exists
   ```
[...]

PR #1017, opened 2026-05-15
fix(bootstrap): cover v0.34.1 oauth_clients.{source_id, federated_read} columns

ChenyqThu:
## Why

`applyForwardReferenceBootstrap` in both `PostgresEngine` and `PGLiteEngine` covers 9 forward-reference targets that `schema-embedded.ts` makes (pages.source_id, pages.deleted_at, links.*, content_chunks.*, mcp_request_log.*, subagent_messages.provider_id, ingest_log.source_id, …) — but missed v0.34.1's `oauth_clients.{source_id, federated_read}` additions.

Result: brains upgrading from a pre-v0.34 schema (which has `oauth_clients` but neither column) crash at `schema-embedded.ts` during SCHEMA_SQL replay with:

```
column "source_id" does not exist
```

…on the partial-index `CREATE INDEX idx_oauth_clients_source_id ON oauth_clients(source_id) WHERE source_id IS NOT NULL` and on the parallel `idx_oauth_clients_federated_read` GIN-index DDL. Migrations v60-v65 backfill both columns + RLS policies later, but they can't run until SCHEMA_SQL replay succeeds, so the bootstrap must add enough state up front.

## Patch shape

Same pattern as #627 (superseded by upstream fixwave #682+#741 for `mcp_request_log` v0.26.3 columns):

1. **Probe** rows in the existing `information_schema` EXISTS round-trip — `oauth_clients_exists`, `oauth_source_id_exists`, `oauth_federated_read_exists`.
2. **`needsOauthClientsBootstrap`** when the table exists but either column is missing.
3. **Early-return guard** extended so fresh installs and modern brains still no-op.
4. **`ALTER TABLE` block** at the end of the function alongside the other missing-column bootstraps. `source_id` picks up its FK to `sources(id)` and `federated_read` its `NOT NULL DEFAULT '{}'`.

Mirrored across `src/core/postgres-engine.ts` and `src/core/pglite-engine.ts`. Doc comments at the top of each function updated to list the new coverage.

## Test follow-through

`test/schema-bootstrap-coverage.test.ts` gains two entries in `REQUIRED_BOOTSTRAP_COVERAGE` (`oauth_clients.source_id`, `oauth_clients.federated_read`). The existing contract test strips both columns, runs the bootstrap, then asserts they exist after SCHEMA_SQL replay — same loop as every other v33+ column the test already covers.

E2E test (`test/e2e/postgres-bootstrap.test.ts`) was not extended in this PR — its single "pre-v0.18" mutation scenario is structurally identical for any forward-reference target, and the contract test already validates the per-column bootstrap. Happy to add an oauth-specific E2E case if maintainers prefer.

## Validation

- `bun run typecheck` clean
- `bun test test/schema-bootstrap-coverage.test.ts test/bootstrap.test.ts`:
  - **11 pass / 0 fail** (was 11 pass with 48 `expect()` calls; now 50 — the +2 new probes)

## Field repro

Hit during the Jarvis KOS v2 fork's v0.31.2 → v0.34.4 sync (commit `1b6acd77`):

- `bun install` postinstall ran `gbrain apply-migrations` (requires schema_version ≥ 51 for the v0.32.2 facts orchestrator).
- Schema was at v45. Running `gbrain init --migrate-only` to bump v45 → v66 failed at the `idx_oauth_clients_source_id` DDL.
- Worked around with one manual `ALTER TABLE oauth_clients ADD COLUMN IF NOT EXISTS source_id … / federated_read …`; migrations v60-v65 then ran clean in under a minute.
- Same shape as the `mcp_request_log` trap that produced #627 in our v0.26.7 sync.

### DIFF
--- a/src/core/pglite-engine.ts
+++ b/src/core/pglite-engine.ts
@@ -243,6 +243,8 @@ export class PGLiteEngine implements BrainEngine {
    *     (indexed by `idx_mcp_log_agent_time`) — v0.26.3
    *   - `subagent_messages.provider_id` column (indexed by
    *     `idx_subagent_messages_provider`) — v0.27
+   *   - `oauth_clients.source_id` + `federated_read` columns (indexed by
+   *     `idx_oauth_clients_source_id` and `idx_oauth_clients_federated_read`) — v0.34.1
    *
    * **Maintenance contract:** when a future migration adds a column-with-index
    * or new-table-with-FK referenced by PGLITE_SCHEMA_SQL, extend this method
@@ -288,7 +290,13 @@ export class PGLiteEngine implements BrainEngine {
         EXISTS (SELECT 1 FROM information_schema.tables
                 WHERE table_schema='public' AND table_name='ingest_log') AS ingest_log_exists,
         EXISTS (SELECT 1 FROM information_schema.columns
-                WHERE table_schema='public' AND table_name='ingest_log' AND column_name='source_id') AS ingest_log_source_id_exists
+                WHERE table_schema='public' AND table_name='ingest_log' AND column_name='source_id') AS ingest_log_source_id_exists,
+        EXISTS (SELECT 1 FROM information_schema.tables
+                WHERE table_schema='public' AND table_name='oauth_clients') AS oauth_clients_exists,
+        EXISTS (SELECT 1 FROM information_schema.columns
+                WHERE table_schema='public' AND table_name='oauth_clients' AND column_name='source_id') AS oauth_source_id_exists,
+        EXISTS (SELECT 1 FROM information_schema.columns
+                WHERE table_schema='public' AND table_name='oauth_clients' AND column_name='federated_read') AS oauth_federated_read_exists
     `);
     const probe = rows[0] as {
       pages_exists: boolean;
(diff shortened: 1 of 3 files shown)


ChenyqThu:
Superseded by upstream v0.35.5.0 ([4446e9f9](https://github.com/garrytan/gbrain/commit/4446e9f9d2276ba9d6e524181ce635aaae1ac564), PR #1111).

The v0.35.5.0 `fix(bootstrap): extend probes for files/oauth_clients/sources.archived* + add MIGRATIONS introspection guard` commit is a strict superset of this PR:
- Covers oauth_clients.{source_id, federated_read} (this PR's scope)
- Also covers files.{source_id, page_id} (v0.18 forward refs, #974)
- Also covers sources.{archived, archived_at, archive_expires_at} (v0.26.5 v34 promotion)
- Adds DDL connection threading (Codex P1 catch — bootstrap probes now run inside the advisory lock)
- Adds MIGRATIONS-source introspection contract test that catches the entire column-only forward-ref bug class at PR time

Same pattern as #627 (closed when v0.31.1.1 fixwave superseded with broader v0.20 + v39-v41 column coverage).

Production impact for jarvis-knowledge-os-v2 fork: already manually ALTER-ed on the prod DB during v0.34.4 sync (per docs/JARVIS-ARCHITECTURE.md §6.24); v0.35.6.0 sync landed today (2026-05-17) and `bun install` postinstall confirmed `All migrations up to date` with no manual intervention needed.

PR #1023, opened 2026-05-15
fix(sync): pass no-recurse-submodules to git pull

skalingclouds:
GBrain's shared GIT_SSRF_FLAGS array includes --no-recurse-submodules. That flag is valid for git clone, but when pullRepo spreads it before the git subcommand, git treats it as a top-level git option and fails with 'unknown option: --no-recurse-submodules'.\n\nThis keeps the SSRF config flags shared, but passes --no-recurse-submodules after the pull subcommand where git pull accepts it.\n\nVerified locally with:\n- git -C /Users/chris/brain -c http.followRedirects=false -c protocol.file.allow=never -c protocol.ext.allow=never pull --ff-only --no-recurse-submodules\n- gbrain sync --repo /Users/chris/brain --no-embed\n- /Users/chris/.hermes/scripts/gbrain_live_sync.sh

### DIFF
--- a/src/core/git-remote.ts
+++ b/src/core/git-remote.ts
@@ -26,10 +26,14 @@ import { isInternalUrl } from './url-safety.ts';
  * - protocol.ext.allow=never: no external helpers (`git-remote-foo`)
  * - --no-recurse-submodules: .gitmodules cannot become a second fetch surface
  */
-export const GIT_SSRF_FLAGS = [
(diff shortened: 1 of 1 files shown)


skalingclouds:
Closing this as obsolete: upstream master already contains the same fix in #1053 (split global git config flags from subcommand flags and pass --no-recurse-submodules after clone/pull).

PR #1028, opened 2026-05-15
fix: add items to extract_facts array schema

amreshtech:
## Problem

`extract_facts` exported an MCP array parameter without an `items` schema:

```ts
entity_hints: { type: "array", description: "Existing canonical entity slugs..." }
```

That shape is tolerated by some clients, but stricter OpenAI Codex tool validation rejects it as an invalid function schema:

```
Invalid schema for function "mcp_gbrain_extract_facts": array schema missing items
```

In practice this caused a Hermes Kanban worker on the primary Codex path to fail its first model call and fall back to a secondary model.

## Reproduction

1. Start GBrain as an MCP server with the pre-fix schema.
2. Expose its tools to a strict OpenAI-compatible tool consumer.
3. Inspect the generated tool definition for `extract_facts.entity_hints` or run a worker that loads the toolset.

Observed invalid emitted schema:

```json
{
  "type": "array",
  "description": "Existing canonical entity slugs the agent has already resolved. Helps the extractor pick the right slug."
}
```

Expected shape:

```json
{
  "type": "array",
  "items": { "type": "string" },
  "description": "Existing canonical entity slugs the agent has already resolved. Helps the extractor pick the right slug."
}
```

## Fix

- add `items: { type: "string" }` to `extract_facts.entity_hints` in `src/core/operations.ts`
- add a regression assertion in `test/e2e/mcp.test.ts` so every exported MCP array property must include `items`

No runtime behavior change for valid callers beyond making the MCP tool schema compliant for strict clients.

## Test Plan

- [x] `bun test ./test/e2e/mcp.test.ts`
- [x] `bun run verify`
- [x] post-fix generated MCP tool defs inspected locally: zero exported array properties missing `items`
- [x] post-fix Hermes Kanban worker smoke test completed without the prior Codex schema error

### DIFF
--- a/src/core/operations.ts
+++ b/src/core/operations.ts
@@ -2573,7 +2573,7 @@ const extract_facts: Operation = {
   params: {
     turn_text: { type: 'string', required: true, description: 'The user message or page body to extract facts from. Sanitized via INJECTION_PATTERNS before the LLM call.' },
     session_id: { type: 'string', description: 'Opaque session id (e.g. topic-id from MCP _meta.session_id, or CLI --session). Stored on each fact for the recall --session filter. Not an auth surface.' },
-    entity_hints: { type: 'array', description: 'Existing canonical entity slugs the agent has already resolved. Helps the extractor pick the right slug.' },
+    entity_hints: { type: 'array', items: { type: 'string' }, description: 'Existing canonical entity slugs the agent has already resolved. Helps the extractor pick the right slug.' },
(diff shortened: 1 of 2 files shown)


garrytan:
Thanks @amreshtech — `extract_facts.entity_hints` (and the other MCP array-schema `items` fixes) already ship in master as of **v0.35.3.0**. The shared mapper is `paramDefToSchema` in `src/mcp/tool-defs.ts` (used by stdio MCP, HTTP MCP, and the subagent registry); `test/mcp-tool-defs.test.ts` walks every array param and fails the suite if any lacks `items.type`. If your install is still hitting this on next `gbrain upgrade`, please reopen with the output of `gbrain doctor --json`.

PR #1033, opened 2026-05-15
fix(orphans): exclude soft-deleted pages and inbound links from soft-deleted sources (closes #1021)

### LINKED ISSUE #1021: orphans: soft-deleted pages still counted in orphan scan, inflating orphan count
## Bug Report

**gbrain version:** 0.33.1.0  
**Engine:** PGLite

### Problem

`gbrain orphans` counts soft-deleted pages in its orphan scan. When you soft-delete a page with `gbrain delete <slug>`, the row remains in the `pages` table (recoverable for 72h), and the orphan query:

```sql
SELECT p.slug, ...
FROM pages p
WHERE NOT EXISTS (
  SELECT 1 FROM links l WHERE l.to_page_id = p.id
)
```

...does not filter out soft-deleted rows (`p.deleted_at IS NULL` missing). This means:

1. Soft-deleted orphan pages still inflate the orphan count.
2. Soft-deleting pages that *linked to other pages* can cause previously non-orphaned pages to appear orphaned (their inbound links disappear).

### Reproduction

```bash
# Create a page and link it to another
gbrain put test-page < <(echo '# Test')
gbrain link test-page some-other-page

# Orphan count — some-other-page should not be orphaned (has inbound link)
gbrain orphans --count  # e.g., 100

# Soft-delete test-page
gbrain delete test-page

# Orphan count — some-other-page is now orphaned (inbound link gone)
# AND deleted test-page is still counted as orphan
gbrain orphans --count  # e.g., 101 or higher
```

### Expected Behavior

`gbrain orphans` should filter out soft-deleted pages:
```sql
WHERE p.deleted_at IS NULL
AND NOT EXISTS (SELECT 1 FROM links l WHERE l.to_page_id = p.id)
```

### Impact

In multi-source setups where the same page slugs exist in both `default` and another source (e.g., after `gbrain sync --repo` imports into `default` while the same files are also managed by a named source), soft-deleting the duplicates to clean up does not reduce the orphan count until the 72h hard-purge window expires. The doctor's `sources purge-orphan-clones` command (tracked for v0.32.0) would help here.

### Workaround

Wait 72h for the autopilot hard-purge phase to remove soft-deleted rows. No immediate fix available.

mvanhorn:
## Summary

`gbrain orphans` no longer counts soft-deleted pages, and inbound links from soft-deleted source pages no longer hide downstream pages that have become effectively orphaned. Both engines move together; same fix lands in the `orphan_pages` stats subquery.

## Why this matters

> `gbrain orphans` counts soft-deleted pages in its orphan scan. When you soft-delete a page with `gbrain delete <slug>`, the row remains in the `pages` table (recoverable for 72h), and the orphan query does not filter out soft-deleted rows.

That's @Simona-digital-ops in [#1021](https://github.com/garrytan/gbrain/issues/1021). The reporter pinpointed the SQL and named the missing filter; this PR adds it.

The pages.deleted_at column was introduced in v0.26.5 with `softDeletePage` semantics. CLAUDE.md notes `buildVisibilityClause` in `src/core/search/sql-ranking.ts` is the canonical "hide soft-deleted" pattern that `searchKeyword` / `searchKeywordChunks` / `searchVector` consume. The orphan query and the orphan_pages stats subquery predate soft-delete and were never updated. Result: every `gbrain delete <slug>` increments the orphan count instead of decrementing it, and soft-deleting any linking page silently orphans its downstream neighbors.

## Changes

- `src/core/postgres-engine.ts` and `src/core/pglite-engine.ts` `findOrphanPages` add `WHERE p.deleted_at IS NULL` plus a `JOIN pages from_p ON from_p.id = l.from_page_id WHERE from_p.deleted_at IS NULL` on the inbound-link subquery. The cascade fix is the second half: a target's "inbound link" from a deleted source page is no longer a real link from the rest of the system's perspective.
- The `orphan_pages` count in the stats query (`postgres-engine.ts:3175+`, `pglite-engine.ts:3162+`) gets the same two filters so health-dashboard numbers match what `gbrain orphans` reports.

## Testing

`bun test test/orphans.test.ts`: 38 pass / 0 fail. Two new cases cover (a) soft-deleted pages don't count as orphans and (b) targets linked only from a soft-deleted source DO show up as orphans (the cascade case the reporter named).

Both engines covered. Cross-engine parity gate stays clean.

Fixes #1021.

AI was used for assistance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,10 @@ ZeroEntropy ships two specialized small models that target the two weakest retri
 - New tests: `test/ai/zeroentropy-recipe.test.ts`, `test/ai/dims-zeroentropy.test.ts`, `test/search/rerank.test.ts`, `test/rerank-audit.test.ts` — 42 cases pinning recipe shape, dim allowlist, 4th-arg inputType plumbing, reranker fail-open across every error class, null-vs-undefined topNOut semantics, and the no-success-logging contract.
 - The plan that drove this release went through two Codex rounds (47 source-grounded findings across consult + adversarial challenge) before any code was written. Every must-fix landed; nine deferred bugs (the cache-wrapper double-embed, MiniMax embo-01 asymmetry, openai-compat allowlist gap, cache-hit limit/budget divergence) are documented in the plan file at `~/.claude/plans/system-instruction-you-are-working-linked-moonbeam.md`.
 
+### Fixed
+
+- `gbrain orphans` no longer counts pages you've soft-deleted, and inbound links from soft-deleted pages no longer hide downstream pages that are now effectively orphaned.
+
 ## To take advantage of v0.35.0.0
 
 `gbrain upgrade` does NOT auto-switch your embedding or reranker model. ZeroEntropy is opt-in; you choose when to flip. To try it:
--- a/src/core/pglite-engine.ts
+++ b/src/core/pglite-engine.ts
@@ -1882,9 +1882,14 @@ export class PGLiteEngine implements BrainEngine {
          COALESCE(p.title, p.slug) AS title,
          p.frontmatter->>'domain' AS domain
        FROM pages p
-       WHERE NOT EXISTS (
-         SELECT 1 FROM links l WHERE l.to_page_id = p.id
-       )
+       WHERE p.deleted_at IS NULL
+         AND NOT EXISTS (
+           SELECT 1
+             FROM links l
+             JOIN pages from_p ON from_p.id = l.from_page_id
+            WHERE l.to_page_id = p.id
+              AND from_p.deleted_at IS NULL
+         )
        ORDER BY p.slug`
     );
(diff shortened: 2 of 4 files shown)


IasminIA:
[max turns (5) reached — partial execution]

garrytan:
Thanks @mvanhorn — already shipped in master as of **v0.35.5.0**. `findOrphanPages` in both engines now filters `p.deleted_at IS NULL` on the candidate side AND adds `JOIN pages src ON src.id = l.from_page_id WHERE src.deleted_at IS NULL` to the EXISTS subquery on the link-source side. Closes #1021. Pinned by `test/orphans.test.ts` soft-delete cases.

The reasoning in your PR was sound — independently caught the same bug class as the structural fix that landed. Real appreciation.

mvanhorn:
Thanks for the kind words, and for landing the structural fix - good to know v0.35.5.0 has it covered.

PR #1043, opened 2026-05-15
fix(mcp): add items clause to extract_facts.entity_hints schema

### LINKED ISSUE #1042: extract_facts.entity_hints: array schema missing items (rejected by strict JSON Schema validators)
## Summary

The MCP tool `extract_facts` declares the parameter `entity_hints` as a JSON
Schema array without an `items` clause:

```ts
entity_hints: { type: 'array', description: '...' }
```

OpenAI's tool API and other strict JSON Schema validators reject any "array"
property that lacks an "items" definition. The full toolset upload fails with
HTTP 400, so every gbrain MCP tool (not just `extract_facts`) becomes
unreachable through that provider.

## Reproducer

Connect any client that uploads the gbrain MCP tool definitions to a strict-mode
provider (OpenAI Chat Completions / Responses API with tools, or an
OpenAI-compatible gateway with strict schema enforcement). The toolset upload
returns HTTP 400 with a message of the form:

```
Invalid schema for function 'extract_facts': In context=('properties', 'entity_hints'), array schema missing items.
```

No tool call is attempted. Other gbrain tools are not invokable through the
rejecting provider because the toolset upload fails as a whole.

## Expected

`extract_facts` can be uploaded and called through strict JSON Schema validators.

## Actual

Toolset upload fails with HTTP 400 due to the missing `items` clause on
`entity_hints`. Lenient providers silently accept the schema, which is why this
has gone unnoticed.

## Impact

- Affects gbrain users whose chat or agent layer routes through a strict-mode
  JSON Schema validator. OpenAI's first-party tool API and several
  OpenAI-compatible gateways enforce this.
- The failure mode is silent from gbrain's side — the toolset upload error is
  owned by the provider — so the downstream symptom is "the model never calls
  any gbrain tool."

## Root cause

`src/core/operations.ts:2699` declares `entity_hints` as
`{ type: 'array', description: ... }` with no `items`. The handler at line 2733
already treats the field as `string[]` via
`Array.isArray(p.entity_hints) ? (p.entity_hints as string[]) : undefined`, so
the runtime contract is unambiguous; only the schema is missing the declaration.

The sibling array param `pages_updated` at line 1888 demonstrates the canonical
pattern: `{ type: 'array', required: true, items: { type: 'string' } }`.

## Proposed fix

Add `items: { type: 'string' }` to `entity_hints`, matching the `pages_updated`
precedent. PR with a property-test guard follows.

curtitoo:
Fixes #1042

## Summary

`extract_facts.entity_hints` declares `{ type: 'array' }` with no `items`
clause. OpenAI's tool API and other strict JSON Schema validators reject the
toolset upload with HTTP 400, taking every gbrain tool offline on those
providers.

This adds `items: { type: 'string' }` to match:

- the runtime handler at line 2733 which already casts to `string[]`
- the existing `pages_updated` precedent at line 1888

## Test

A property test in `test/mcp-tool-defs.test.ts` walks every
`buildToolDefs(operations)` entry and asserts that any `array` property has an
`items` clause:

```ts
test('every array property has an items clause (strict JSON Schema)', () => {
  const violations: string[] = [];
  for (const def of buildToolDefs(operations)) {
    for (const [propName, propSchema] of Object.entries(def.inputSchema.properties)) {
      const s = propSchema as { type?: unknown; items?: unknown };
      if (s && typeof s === 'object' && s.type === 'array' && !s.items) {
        violations.push(`${def.name}.${propName}`);
      }
    }
  }
  expect(violations).toEqual([]);
});
```

Verified manually: revert the schema fix, the test fails with
`["extract_facts.entity_hints"]`; re-apply and it passes. With the fix applied
the suite is green (`bun test test/mcp-tool-defs.test.ts`: 6 pass / 0 fail
across all 63 operations). `bun run typecheck` and
`scripts/check-test-isolation.sh` both pass.

## Blast radius

Pre-fix the param was already a `string[]` at the runtime layer; the schema was
just under-specified. Lenient providers accepted it, strict providers rejected
the whole toolset. The fix tightens the schema to match the runtime cast — no
behavior change for accepting clients, fixes HTTP 400 for strict clients.

No new dependencies. No changes outside the schema line and the test.

## Alternatives considered

- A more specific schema (slug pattern, length cap) — opted to mirror
  `pages_updated` for consistency; tightening can land separately.
- A single-case assertion just for `extract_facts.entity_hints` — opted for the
  property test instead so future array params can't regress the same way
  silently.

### DIFF
--- a/src/core/operations.ts
+++ b/src/core/operations.ts
@@ -2696,7 +2696,7 @@ const extract_facts: Operation = {
   params: {
     turn_text: { type: 'string', required: true, description: 'The user message or page body to extract facts from. Sanitized via INJECTION_PATTERNS before the LLM call.' },
     session_id: { type: 'string', description: 'Opaque session id (e.g. topic-id from MCP _meta.session_id, or CLI --session). Stored on each fact for the recall --session filter. Not an auth surface.' },
-    entity_hints: { type: 'array', description: 'Existing canonical entity slugs the agent has already resolved. Helps the extractor pick the right slug.' },
+    entity_hints: { type: 'array', items: { type: 'string' }, description: 'Existing canonical entity slugs the agent has already resolved. Helps the extractor pick the right slug.' },
     is_dream_generated: { type: 'boolean', description: 'When true, extraction is skipped (anti-loop). Caller flips this on for pages with dream_generated:true frontmatter.' },
     visibility: { type: 'string', description: 'Default visibility for extracted facts. private (default) | world.' },
   },
--- a/test/mcp-tool-defs.test.ts
+++ b/test/mcp-tool-defs.test.ts
@@ -64,4 +64,20 @@ describe('buildToolDefs', () => {
       expect(Array.isArray(def.inputSchema.required)).toBe(true);
     }
   });
+
+  test('every array property has an items clause (strict JSON Schema)', () => {
+    // OpenAI's tool API and other strict JSON Schema validators reject an
+    // "array" property that lacks "items". An offender causes the entire
+    // toolset to fail upload with HTTP 400 before any tool is invoked.
+    const violations: string[] = [];
+    for (const def of buildToolDefs(operations)) {
+      for (const [propName, propSchema] of Object.entries(def.inputSchema.properties)) {
+        const s = propSchema as { type?: unknown; items?: unknown };
+        if (s && typeof s === 'object' && s.type === 'array' && !s.items) {
+          violations.push(`${def.name}.${propName}`);
+        }
+      }
+    }
+    expect(violations).toEqual([]);
+  });
 });


garrytan:
Thanks @curtitoo — `extract_facts.entity_hints` (and the other MCP array-schema `items` fixes) already ship in master as of **v0.35.3.0**. The shared mapper is `paramDefToSchema` in `src/mcp/tool-defs.ts` (used by stdio MCP, HTTP MCP, and the subagent registry); `test/mcp-tool-defs.test.ts` walks every array param and fails the suite if any lacks `items.type`. If your install is still hitting this on next `gbrain upgrade`, please reopen with the output of `gbrain doctor --json`.

PR #1050, opened 2026-05-15
fix(doctor): only count non-zero exit codes as worker crashes

garrytan-agents:
## Problem

The doctor's supervisor health check counts ALL `worker_exited` audit events as crashes, including code=0 clean exits (drain after queue empty, health-check restart, etc.). This inflates the crash count to 100+ per day when the actual crash count is 0.

PR #1002 already fixed the supervisor's own crash counting (`crashCount = 0` for code=0). This patch applies the same logic to the doctor's independent audit-log probe.

## Fix

- Filter `worker_exited` events by exit code: only `code !== 0` counts as a crash
- Show clean exit count separately in status message for visibility
- Crash threshold (>3) now only applies to real crashes

## Evidence

Production audit log shows 102 `worker_exited` events in 24h, ALL with `code: 0`. Doctor reports "102 crashes" when actual crashes = 0.

```
worker_exited { code: 0, signal: null, crash_count: 0, reason: "code 0" }
```

Companion to PR #1002 (supervisor side) and PR #1004 (RSS accuracy).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,6 @@ export/
 # Tier 3 PGLite snapshot fixture (built on demand by build:pglite-snapshot)
 test/fixtures/pglite-snapshot.tar
 test/fixtures/pglite-snapshot.version
+ze-*.ts
+embedding-benchmark.ts
+reports/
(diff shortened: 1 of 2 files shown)


garrytan:
Thanks — closing as already-shipped. v0.35.5.0 introduced `summarizeCrashes()` in `src/core/minions/handlers/supervisor-audit.ts` as the single source of truth for crash-event classification; both `gbrain doctor` (`doctor.ts:1011-1043`) and `gbrain jobs supervisor status` (`jobs.ts:803-826`) route through it. The denylist over allowlist design (clean_exit / graceful_shutdown are NON-crashes; everything else including future unrecognized causes counts) is the regression guard the parallel `worker_exited`-filter pattern lacked. Pinned by 14 cases in `test/supervisor-audit.test.ts` + 4 source-grep wiring assertions in `test/doctor.test.ts` that ban the ad-hoc filter from drifting back in.

PR #1051, opened 2026-05-15
Make embedding vector dimension configurable + add LM Studio recipe

### LINKED ISSUE #203: init: --provider flags required even when config.json has persisted embedding settings (DX)
Running a BGE-M3 local embedding setup on v0.10.3 using the pluggable provider layer from #172. `~/.gbrain/config.json` has the embedding settings persisted:

```json
{
  "engine": "pglite",
  "database_path": "/home/user/.gbrain/brain.pglite",
  "embedding": {
    "provider": "ollama",
    "model": "bge-m3",
    "dimensions": 1024,
    "base_url": "http://localhost:8085/v1"
  }
}
```

When PR #188 merged I wanted to apply the new schema migrations, so I ran `gbrain init --pglite` and got:

```
Cannot re-init: existing brain has a different embedding dimension.
  Existing: ollama / bge-m3 (1024d)
  Requested: openai / text-embedding-3-large (1536d)
```

The "Existing" line proves config.json was read somewhere in the flow. The "Requested" line is the baked default, not anything I passed on the CLI. So every re-init fails unless I repeat the flags I already wrote down in the config file.

Workaround:

```
gbrain init --pglite --provider ollama --model bge-m3 --dimensions 1024 --base-url http://localhost:8085/v1
```

Which works but feels like the config file is there for a reason.

Looking at `src/commands/init.ts::resolveProviderWithGuard()`, it calls `resolveEmbeddingConfig(parseEmbeddingFlags(args))` using env vars and CLI flags, without threading the loaded config's embedding section in as defaults.

Rough shape of a fix:

```ts
const existing = loadConfig();
const defaults = existing?.embedding
  ? {
      provider: existing.embedding.provider,
      model: existing.embedding.model,
      dimensions: existing.embedding.dimensions,
      baseUrl: existing.embedding.base_url,
    }
  : {};
const resolved = resolveEmbeddingConfig({ ...defaults, ...parseEmbeddingFlags(args) });
```

Can take a crack at a PR if that shape looks right to you, or if you'd prefer a different fallback order let me know.

vincedk-alt:
## Summary

Two patches for non-OpenAI embedding workflows:

1. **vector(N) schema parameterization** — schema.sql had vector(1536) hardcoded for OpenAI's text-embedding-3-small dim. Changed to read embedding_dimensions from config (default 1536 for backwards compat), so users running local models like nomic-embed-text-v1.5 (768) or mxbai-embed-large (1024) don't have to fork.

2. **LM Studio recipe** — sibling of the ollama recipe at recipes/lmstudio.ts, points at http://localhost:1234/v1 by default. Cleanly separates LM Studio integration from the misleading "ollama at non-Ollama port" workaround pattern.

## Why

Running gbrain on a personal fork to support a hybrid local+cloud setup (LM Studio MLX for embeddings, Anthropic Haiku for chat). Every upstream release requires manual rebase. These patches are generally useful — anyone running gbrain with a non-OpenAI embedding provider needs the dim parameterization.

## Verified by operator

- Running in production on operator's machine for ~10 days with the vector(768) schema patch
- 92 brain pages embedded at 768 dim using nomic-embed-text-v1.5 via LM Studio MLX
- gbrain doctor passes (45/100 brain_score is unrelated — content sparsity, not schema)
- Anthropic Haiku continues to work as the chat/expansion model alongside LM Studio embeddings

## Test plan (not personally re-verified before opening this PR — operator's prior test run is the basis)

The original patches were validated in the operator's earlier work session (2026-05-14). The PR-opener (an agentic helper) is surfacing the patches upstream rather than re-running the test suite. Maintainer may wish to re-validate via:
- Aged-brain test: existing vector(1536) databases continue to work with no migration needed
- Fresh-init test with embedding_dimensions=768 in config produces a working PGLite database
- WAL-stress: 100 sequential gbrain put operations under the new schema path
- LM Studio recipe smoke test: gbrain put then gbrain search round-trip

Branch is 4 commits behind master at time of PR opening. Happy to rebase or iterate on the patch shape if a different abstraction is preferred.

---

**Update 2026-05-16:** This PR's branch now also carries commit `5d75545` (gitignore the gbrain `db_only` auto-managed directories `media/x/`, `media/articles/`, `meetings/transcripts/`). Strictly additive; doesn't conflict with the schema parameterization above.

Separately, see **#1060** (closes #203) — the same operator's branch surfaced a config-first init bug while exercising the v0.35 → v0.35.0.1 rebuild path. That fix is independent of this PR; both can land in any order.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,8 @@ export/
 # Tier 3 PGLite snapshot fixture (built on demand by build:pglite-snapshot)
 test/fixtures/pglite-snapshot.tar
 test/fixtures/pglite-snapshot.version
+
+# Auto-managed by gbrain (db_only directories)
+media/x/
+media/articles/
+meetings/transcripts/
--- a/src/core/ai/recipes/index.ts
+++ b/src/core/ai/recipes/index.ts
@@ -10,6 +10,7 @@ import { openai } from './openai.ts';
 import { google } from './google.ts';
 import { anthropic } from './anthropic.ts';
 import { ollama } from './ollama.ts';
+import { lmstudio } from './lmstudio.ts';
 import { voyage } from './voyage.ts';
 import { litellmProxy } from './litellm-proxy.ts';
 import { deepseek } from './deepseek.ts';
@@ -27,6 +28,7 @@ const ALL: Recipe[] = [
   google,
   anthropic,
   ollama,
+  lmstudio,
   voyage,
   litellmProxy,
   deepseek,
--- a/src/core/ai/recipes/lmstudio.ts
+++ b/src/core/ai/recipes/lmstudio.ts
@@ -1,0 +1,31 @@
+import type { Recipe } from '../types.ts';
+
+export const lmstudio: Recipe = {
+  id: 'lmstudio',
+  name: 'LM Studio (local MLX)',
+  tier: 'openai-compat',
+  implementation: 'openai-compatible',
+  base_url_default: 'http://localhost:1234/v1',
+  auth_env: {
+    required: [], // LM Studio runs unauthenticated locally; users pass `lm-studio` as the key.
+    optional: ['LMSTUDIO_BASE_URL', 'LMSTUDIO_API_KEY'],
+    setup_url: 'https://lmstudio.ai',
+  },
+  touchpoints: {
+    embedding: {
+      models: [
+        'text-embedding-nomic-embed-text-v1.5',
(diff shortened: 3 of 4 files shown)


time-attack:
Closing during backlog triage: this hardcodes the default embedding column to vector(768), which would break default 1536-dim installs. A truly configurable-dimension approach would be welcome as a fresh PR (see the dims_options pattern).

PR #1057, opened 2026-05-16
fix: include items field in array param schemas for HTTP MCP server

DF-FrancoOS:
## Problem

The HTTP MCP server (`serve-http.ts`) has its own inline schema builder that spreads `enum` and `default` from param definitions but omits `items`. This causes OpenAI-strict clients (gpt-4o, etc.) to reject `tools/list` with:

```
400 Invalid schema for function 'gbrain__extract_facts': In context=(properties', 'entity_hints'), array schema missing items.
```

The agent fails on every turn and surfaces an error to the user.

## Root cause

Two issues:

1. `serve-http.ts` inline schema builder does not spread `v.items` (unlike `tool-defs.ts` which handles it correctly)
2. The `entity_hints` param in `operations.ts` was missing `items` entirely

## Fix

- `serve-http.ts`: add `...(v.items ? { items: { type: v.items.type } } : {})` to the inline property mapper — matches the existing pattern for `enum` and `default`
- `operations.ts`: add `items: { type: 'string' }` to `entity_hints`

## Tested

Verified fix against GBrain HTTP MCP server connected to OpenClaw with gpt-4o. `tools/list` accepted, agent responds successfully.

### DIFF
--- a/src/commands/serve-http.ts
+++ b/src/commands/serve-http.ts
@@ -818,6 +818,7 @@ export async function runServeHttp(engine: BrainEngine, options: ServeHttpOption
                 description: v.description,
                 ...(v.enum ? { enum: v.enum } : {}),
                 ...(v.default !== undefined ? { default: v.default } : {}),
+                ...(v.items ? { items: { type: v.items.type } } : {}),
               }]),
             ),
             required: Object.entries(op.params).filter(([, v]) => v.required).map(([k]) => k),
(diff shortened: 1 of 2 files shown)


garrytan:
Thanks @DF-FrancoOS — `extract_facts.entity_hints` (and the other MCP array-schema `items` fixes) already ship in master as of **v0.35.3.0**. The shared mapper is `paramDefToSchema` in `src/mcp/tool-defs.ts` (used by stdio MCP, HTTP MCP, and the subagent registry); `test/mcp-tool-defs.test.ts` walks every array param and fails the suite if any lacks `items.type`. If your install is still hitting this on next `gbrain upgrade`, please reopen with the output of `gbrain doctor --json`.

Issue #1058, opened 2026-05-16
init: GBRAIN_EMBEDDING_MODEL/DIMENSIONS env vars ignored when no config.json AND no DATABASE_URL


vincedk-alt:
## Summary

On a fresh install (no `~/.gbrain/config.json` AND no `DATABASE_URL`), `gbrain init --pglite` ignores `GBRAIN_EMBEDDING_MODEL` and `GBRAIN_EMBEDDING_DIMENSIONS` env vars. The init silently defaults to OpenAI 1536, even when the user has explicitly set their preferred provider via shell env.

## Repro on v0.35.1.0

```bash
# Fresh machine, no config.json, no DATABASE_URL
rm -f ~/.gbrain/config.json
rm -rf ~/.gbrain/brain.pglite

# Set env vars expecting them to take effect
export GBRAIN_EMBEDDING_MODEL=lmstudio:text-embedding-nomic-embed-text-v1.5
export GBRAIN_EMBEDDING_DIMENSIONS=768

# Run plain init
gbrain init --pglite
```

Result: column comes up at `vector(1536)` with `text-embedding-3-large` as the configured model. First embed against LM Studio returns 768-dim, sync explodes with `expected 1536 dimensions, not 768`.

## Root cause

`loadConfig()` at `src/core/config.ts:135` early-returns null when neither `fileConfig` nor `dbUrl` is set:

```ts
if (!fileConfig && !dbUrl) return null;
```

Env-var merge logic below that early-return (config.ts:153-176) never runs. The init's `resolveAIOptions` sees null and falls through to gateway defaults. The env vars ARE honored when there's already a config.json or a DATABASE_URL (both make `loadConfig` return non-null, then the merge step runs) — but not in the env-only cold-start case.

## Workaround

Pre-create config.json before running init. After config.json exists, env vars are honored (loadConfig returns non-null, merge step runs, env vars override file).

## Suggested fix shapes (any of)

**A. Sibling helper `loadAIConfig()`** that reads ONLY embedding/expansion/chat fields from env + file, no `inferredEngine` logic. Called from `resolveAIOptions` in addition to `loadConfig`. Smallest blast radius; doesn't touch any existing call site.

[...]

PR #257, opened 2026-04-20
feat: v0.28.9 pluggable embedding providers — Vercel AI SDK

garrytan:
## Summary

**v0.27 adds pluggable embedding providers to gbrain via Vercel's AI SDK.**

Query expansion (the Haiku call during search) moves to the same seam. Chunking / transcription / enrichment / fail-improve stay on their direct SDK calls for now — those migrate in a follow-up.

The only real takeaway: **we took Vercel's AI SDK. That's it.** No custom provider code, no hand-rolled adapters. Users pick OpenAI / Google / Ollama / Voyage / any OpenAI-compatible endpoint via `provider:model` config strings.

Also fixes a silent-drop bug at three sites where non-OpenAI brains were invisibly dropping embeddings on every `put_page`.

## What's in the PR

1. **feat: AI gateway + 6 provider recipes + silent-drop fix** — `src/core/ai/` gateway module. Patches silent-drop at `operations.ts:237`, `hybrid.ts:81`, `import-file.ts:112`. Schema templating via `getPGLiteSchema(dims, model)` preserves existing 1536-dim brains (OpenAI's API default is actually 3072 — gateway passes `providerOptions.openai.dimensions: 1536` explicitly).
2. **feat: `gbrain providers` CLI + init flags + config** — `list | test | env | explain` subcommands. `--explain=json` emits a schema-versioned choice matrix for agent-driven installs. `gbrain init --embedding-model openai:text-embedding-3-large` etc.
3. **test: AI gateway + silent-drop + schema templating + no-env-mutation** — 28 new unit tests including a 3-site silent-drop regression grep.
4. **chore: bump v0.27.0 + CHANGELOG + deps.**
5. **Merge origin/master** — pulls upstream through v0.21.0 (Code Cathedral II — call-graph edges, two-pass retrieval, parent-scope chunking) plus all interim releases. Conflicts resolved in CHANGELOG, package.json, VERSION, src/cli.ts, src/core/embedding.ts, src/core/import-file.ts.

## Dependencies

One vendor: **Vercel AI SDK** (`ai`, `@ai-sdk/openai`, `@ai-sdk/google`, `@ai-sdk/anthropic`, `@ai-sdk/openai-compatible`). Plus `zod` for schema validation. That's it. No source vendoring; upgrades via `bun update`.

## Test plan

- [x] `bun test` — 1766 pass / 0 fail (up from 1717 pre-merge)
- [x] `bun run test:e2e` per-file with clean DB — 133 pass / 0 fail across 13 files
[...]

PR #1060, opened 2026-05-16
fix(init): seed resolveAIOptions from loadConfig() (closes #203)

### LINKED ISSUE #203: init: --provider flags required even when config.json has persisted embedding settings (DX)

vincedk-alt:
## Summary

Re-init no longer ignores the embedding settings users already wrote down in `~/.gbrain/config.json`. `gbrain init --pglite` (no flags) on a brain whose persisted config declares a non-default `embedding_model` / `embedding_dimensions` / `provider_base_urls` now respects those values instead of silently falling back to OpenAI 1536. Closes #203.

## Why

`resolveAIOptions` in `src/commands/init.ts` only consumed CLI flags. It never read `loadConfig()`, so `aiOpts` came out empty when no flags were passed. That tripped the `if (embedding_model || chat_model)` gate further down in `initPGLite` (line 386), so `configureGateway` was never called, `_config` stayed null, and `initSchema()`'s catch-block fell back to literal `1536` in `pglite-engine.ts:212`.

Net result on every re-init without flags: the user's persisted choice was silently overridden by the OpenAI default. v0.28.5's `embedding-dim-check` catches the resulting column mismatch on the next embed, but only AFTER the wrong-shape brain is already on disk and the user has spent another minute figuring out why their `--embedding-model lmstudio:...` setup keeps producing `vector(1536)` columns.

## What changed

One function (`resolveAIOptions`) in `src/commands/init.ts`. Seeds `out` from `loadConfig()` at the top, before the existing flag-handling code. Flags still win on a per-field basis. New resolution chain:

1. `~/.gbrain/config.json` via `loadConfig()` — NEW
2. `--model SHORTHAND` recipe lookup (sets model + dims)
3. `--embedding-model VERBOSE` (overrides shorthand)
4. `--embedding-dimensions N` (overrides recipe-derived dim)
5. `--expansion-model` / `--chat-model` (additive)

Gateway defaults still apply downstream when nothing in the chain populates a field — **cold-start behavior unchanged**. `claw-test`'s fresh-install scenario keeps working.

Two edge cases caught during review and fixed in this PR:

- **Model swap clears seeded dim.** If config has `lmstudio:nomic 768` and the user reruns init with `--embedding-model openai:text-embedding-3-large` (no `--embedding-dimensions`), the seeded `768` is cleared so the recipe-default derivation re-fires for the new provider. Without this, the user would silently get OpenAI configured at 768-dim.
- **`provider_base_urls` rides along.** Custom endpoint overrides (e.g. `provider_base_urls.lmstudio` pointing at `http://localhost:1234/v1`) are preserved through the seed + `saveConfig` cycle. Without this, re-init dropped the user's endpoint override and the gateway fell back to recipe defaults on next call.

## Tests

`test/init-config-first.test.ts` — 7 cases, hermetic against the runner env via an `isolateEnv()` helper that unsets `DATABASE_URL`, `GBRAIN_DATABASE_URL`, and every `GBRAIN_EMBEDDING_*` / `GBRAIN_*_MODEL` / `OPENAI_API_KEY` / `ANTHROPIC_API_KEY`. Uses `withEnv()` from the canonical test helpers; no `mock.module`, no `*.serial.test.ts` quarantine needed. The 7 cases pin:

1. Config-only, no flags → returns config values
2. Flag-only, no config → existing flag behavior preserved
3. Both flags and config → flags override config per-field, config fills the gaps
4. Empty (cold-start, no config, no flags) → empty object (existing behavior preserved)
5. `--embedding-model` switches provider away from config → seeded dim cleared, recipe-default re-derives
6. `provider_base_urls` preserved through the seed cycle
7. Env-var-only (KNOWN LIMITATION, tracked at #1058)

Case 7 documents a separate bug class not closed by this PR: when there's no config.json AND no `DATABASE_URL`, `loadConfig()` early-returns null at `config.ts:135`, so the env-var merge below that early-return never fires. Setting `GBRAIN_EMBEDDING_DIMENSIONS=768` env-only on a truly cold install still produces a 1536-dim schema. Filed at #1058 for separate fix.

Deliberate-failure run: `DATABASE_URL=... GBRAIN_EMBEDDING_MODEL=... GBRAIN_EMBEDDING_DIMENSIONS=999 bun test test/init-config-first.test.ts` → 7 pass, 0 fail. The `isolateEnv()` is adversarially proven, not just theoretical.

## Verification

- `bun run typecheck` — clean
- `bun run verify` — all 13 pre-checks pass (check:privacy, check:test-isolation with 410 non-serial files scanned, etc.)
- `bun test test/init-config-first.test.ts` — 7 pass, 0 fail
- `bun run test` — full fast loop ran with 0 fails (two shards exceeded the 600s per-shard timeout cap; per gbrain CLAUDE.md that's the documented "infrastructure problem" classification, not test failures; all 4 shards' completed work reports 0 fail; new test is in shard 2 which completed in 1857 pass / 0 fail / 0 skip / rc=0)
- End-to-end smoke against a tempdir GBRAIN_HOME: column came up at `vector(768)` (verified via direct `pg_attribute format_type` probe) on a config-first init with no flags

## Reviewer notes

Two adversarial reviews ran against this branch before opening the PR:

- A devil's-advocate critic loop (4 rounds, quieted with 4 findings raised + resolved — env-var-only path called out as KNOWN LIMITATION + filed as #1058, the `bun -e` smoke-test syntax error caught and re-run with ground truth, the rebase pre-check added before any force-push, and a test-isolation hole closed by `isolateEnv()`).
- `codex review --base upstream/master` 3 rounds: round 1 surfaced the model-swap dim-stale bug (fixed), round 2 surfaced `provider_base_urls` field-loss (fixed), round 3 clean.

Both reviewers were against the same branch tip after each fix; their findings did not overlap, which is the cross-model-review value-add CLAUDE.md Rule 5b describes.

### DIFF
--- a/src/commands/init.ts
+++ b/src/commands/init.ts
@@ -112,21 +112,80 @@ export async function runInit(args: string[]) {
 }
 
 /**
- * Resolve AI provider options from CLI flags. Verbose form (--embedding-model
- * openai:text-embedding-3-large) overrides shorthand (--model openai which
- * expands to the recipe's first embedding model).
+ * Resolve AI provider options from CLI flags AND the persisted config file.
+ *
+ * Resolution chain (each step overrides the previous on a per-field basis):
+ *
+ *   1. `~/.gbrain/config.json` via `loadConfig()` — NEW in v0.35.0.1+
+ *      (closes upstream issue #203). Re-init no longer requires re-passing
+ *      flags the user already wrote down in config.json.
+ *   2. `--model SHORTHAND` (recipe lookup; sets both model + dims).
+ *   3. `--embedding-model VERBOSE` (overrides shorthand).
+ *   4. `--embedding-dimensions N` (overrides recipe-derived dims).
+ *   5. `--expansion-model` / `--chat-model` (additive).
+ *
+ * Gateway defaults still apply downstream for fields nothing in the chain
+ * populates — cold-start behavior unchanged.
+ *
+ * KNOWN LIMITATION (separate bug class, tracked at #1058): if a user sets
+ * `GBRAIN_EMBEDDING_MODEL` + `GBRAIN_EMBEDDING_DIMENSIONS` env vars but has
+ * NO existing config.json AND no `DATABASE_URL`, `loadConfig()` returns null
+ * at `config.ts:135` (the `if (!fileConfig && !dbUrl) return null;` early-
+ * return). The env-var merge below that early-return never runs, so the
+ * seeding step here sees `existing === null` and skips. Env-only callers
+ * still hit OpenAI/1536 gateway defaults on cold-start. Pinned by case 5
+ * in `test/init-config-first.test.ts`. Out of scope for this PR (would
+ * require either reworking loadConfig's `inferredEngine` fallback, or
+ * threading env-var reads through a sibling `loadAIConfig()` helper).
+ *
+ * @internal Exported for `test/init-config-first.test.ts`.
  */
-async function resolveAIOptions(
+type ResolvedAIOptions = {
+  embedding_model?: string;
+  embedding_dimensions?: number;
+  expansion_model?: string;
+  chat_model?: string;
+  provider_base_urls?: Record<string, string>;
+};
+
+export async function resolveAIOptions(
   verbose: string | null,
   shorthand: string | null,
   dimsArg: number | null,
   expansion: string | null,
   chat: string | null,
-): Promise<{ embedding_model?: string; embedding_dimensions?: number; expansion_model?: string; chat_model?: string }> {
-  const out: { embedding_model?: string; embedding_dimensions?: number; expansion_model?: string; chat_model?: string } = {};
+): Promise<ResolvedAIOptions> {
+  const out: ResolvedAIOptions = {};
+
+  // Step 1: seed from persisted config (closes #203). Falsy guards skip
+  // null / undefined / 0 / empty-string so a partial / corrupt config file
+  // degrades to the legacy flag-only path. Env-var overrides (GBRAIN_EMBEDDING_*)
+  // are already merged into the return of `loadConfig()`, so this respects them too.
+  //
+  // `provider_base_urls` rides alongside the AI model fields — without it,
+  // re-init silently strips the user's custom endpoint override (e.g.
+  // `provider_base_urls.dashscope`) and the gateway falls back to the
+  // recipe default endpoint on the next call.
+  const existing = loadConfig();
+  const seededModel = existing?.embedding_model;
+  if (existing?.embedding_model) out.embedding_model = existing.embedding_model;
+  if (existing?.embedding_dimensions) out.embedding_dimensions = existing.embedding_dimensions;
+  if (existing?.expansion_model) out.expansion_model = existing.expansion_model;
+  if (existing?.chat_model) out.chat_model = existing.chat_model;
+  if (existing?.provider_base_urls && Object.keys(existing.provider_base_urls).length > 0) {
+    out.provider_base_urls = existing.provider_base_urls;
+  }
 
   if (verbose) {
     out.embedding_model = verbose;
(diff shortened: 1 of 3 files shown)


vincedk-alt:
## Follow-up commit: v0.10.x nested-shape coverage (close #203 fully)

The original commit closed the case for users who already have flat-field configs (`embedding_model` / `embedding_dimensions` / `provider_base_urls` at the top level). After an audit of the issue body, that left a gap: **jamebobob's actual repro is the v0.10.x nested shape** (`{embedding: {provider, model, dimensions, base_url}}`).

PR #257 flattened the config in v0.27 but no migration was written. Anyone carrying a config.json from before that release silently falls through to gateway defaults (OpenAI 1536) on every command because no code path reads the nested shape.

New commit `1c6fdb4` adds a back-compat read in `loadConfig()` that maps the nested shape to flat fields **in memory** before the env-merge step:
- `embedding.{provider, model}` → `embedding_model: "provider:model"`
- `embedding.dimensions` → `embedding_dimensions`
- `embedding.{provider, base_url}` → `provider_base_urls[provider] = base_url`

The next `saveConfig` serializes only declared flat fields, so the nested object drops out on save. Flat fields win when both shapes are present (defensive against partial manual edits).

Two new test cases in `test/init-config-first.test.ts`:
- **case 6:** jamebobob's exact repro from the issue body → maps to `ollama:bge-m3` / 1024 / base_url override
- **case 7:** flat-wins-over-nested when both present

Full suite now 9 pass / 0 fail. Typecheck clean. `bun run verify` clean.

This closes #203 for both shapes — users on the v0.10.x nested config AND users who've already manually re-saved to the flat shape.

time-attack:
Thanks for the careful diagnosis — this one has since been fixed on master. `resolveAIOptions` now seeds from `loadConfig()` before flag resolution (the D5 block in src/commands/init.ts, landed in 872d4eeb / #3091), and that change also covers the cold-install env-var case your description lists as an out-of-scope known limitation. The PR additionally targets the old positional signature, which master has replaced with an options object. Closing as already fixed; sorry the timing went against you.

PR #1062, opened 2026-05-16
fix: return success for migration list commands

nezovskii:
## Summary
- Make apply-migrations --list exit 0 after printing an up-to-date migration list
- Make --dry-run and no-op success paths exit 0 explicitly
- Fixes skillpack-check falsely reporting apply-migrations failure when migrations are already applied

## Verification
- gbrain apply-migrations --list exits successfully
- gbrain skillpack-check reports healthy: true locally

### DIFF
--- a/src/commands/apply-migrations.ts
+++ b/src/commands/apply-migrations.ts
@@ -328,13 +328,13 @@ export async function runApplyMigrations(args: string[]): Promise<void> {
     process.exit(2);
   }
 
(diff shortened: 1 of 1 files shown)


time-attack:
Thanks for this, @nezovskii! 🙏

What this does: This PR adds explicit process.exit(0) calls to three code paths in the migration command handler: when listing migrations, during dry-run, and when migrations are already up-to-date. This ensures the command exits with success code instead…

PR #1064, opened 2026-05-16
feat(doctor): child-table orphan detection (closes #1063)

### LINKED ISSUE #1063: doctor + autopilot: child-table orphan detection (content_chunks, page_versions, etc.) not surfaced
## Summary

`gbrain dream`'s `orphans` phase only detects orphan PAGES (pages with no inbound links). It does NOT detect orphan CHILD ROWS in FK-bound tables like `content_chunks`, `page_versions`, `tags`, `takes`, etc. Bulk page deletions can leave orphans in these tables that persist indefinitely.

Discovered after a bulk source removal on v0.35.0.0 left 234 orphan `content_chunks` rows. All FK constraints to `pages.id` are correctly declared as `ON DELETE CASCADE` (verified via `pg_constraint` against the live PGLite DB), so the orphans must have come from one of:

A. A historical schema version where the FK was missing (and never retro-cascaded)
B. A bulk-delete code path that bypassed cascade (manual SQL UPDATE without follow-up CASCADE)
C. A race condition during a large CASCADE batch

Regardless of origin, the existing `orphans` phase doesn't catch them. The `find_orphans` MCP op also only scans pages.

## Repro / Detection

```bash
bun -e "
import { PGlite } from '@electric-sql/pglite';
import { vector } from '@electric-sql/pglite/vector';
const db = new PGlite('/Users/<you>/.gbrain/brain.pglite', { extensions: { vector } });
await db.waitReady;
for (const tbl of ['content_chunks', 'page_versions', 'tags', 'takes', 'links', 'raw_data', 'files', 'timeline_entries']) {
  const col = tbl === 'links' ? 'from_page_id' : 'page_id';
  const r = await db.query(\`SELECT COUNT(*) AS c FROM \${tbl} WHERE \${col} IS NOT NULL AND \${col} NOT IN (SELECT id FROM pages)\`);
  console.log(\`\${tbl}: \${r.rows[0].c} orphans\`);
}
"
```

If any table reports non-zero orphans, the existing `orphans` phase didn't catch them.

## Suggested fix shapes

**A. Extend the existing `orphans` phase** to scan all FK-to-pages tables. Could add to `runPhaseOrphans` in `src/core/cycle.ts`; would surface in autopilot output naturally.

**B. New `find_child_orphans` MCP op** (admin-scoped, localOnly) returning a per-table breakdown. Operators can run on-demand without waiting for the next dream cycle.

**C. New `doctor` check** (`child_table_orphans`) — fast, sampled, surfaces non-zero counts with paste-ready cleanup SQL. Cheapest to implement; matches the existing doctor health-check pattern.

I'd vote (C) — it's the smallest blast radius, gives operators immediate signal, and doesn't change autopilot semantics.

## Context

- Live FK constraints are correct (`pg_get_constraintdef` shows `ON DELETE CASCADE` on every page-referencing table including `content_chunks`).
- 234 orphans were cleaned with `DELETE FROM content_chunks WHERE page_id NOT IN (SELECT id FROM pages)` — manual operator intervention.
- Pre-cleanup orphan check could have been auto-fired by the dream cycle but wasn't.

Severity: medium. Doesn't break anything visible (orphans don't poison search since the join eliminates them), but they bloat storage and embedding-index size proportionally to historical deletes. On a brain with months of churn, this can be hundreds of MB of dead vectors.

vincedk-alt:
## Summary

`gbrain doctor` now surfaces orphan rows in FK-child tables (`content_chunks`, `page_versions`, `tags`, `takes`, `raw_data`, `timeline_entries`, `links`, `files`). Before this PR, the autopilot `orphans` phase only detected orphan PAGES; orphan child rows persisted indefinitely after a bulk page delete that bypassed cascade, until manual SQL cleanup.

Closes #1063.

## Why

All ten FK-to-pages columns declare `ON DELETE CASCADE` in live schema — so any orphan row is unexpected by definition. But pre-FK historical rows, races during large cascade batches, or code paths that bypass cascade can leave child orphans that the existing `orphans` phase (page-graph only) and `find_orphans` MCP op don't catch.

Discovered after a real bulk `sources_remove default --confirm-destructive` left 234 orphan `content_chunks` rows on a real brain. Live FK constraint verified correct via `pg_constraint`:

```
content_chunks: FOREIGN KEY (page_id) REFERENCES pages(id) ON DELETE CASCADE
page_versions:  FOREIGN KEY (page_id) REFERENCES pages(id) ON DELETE CASCADE
takes:          FOREIGN KEY (page_id) REFERENCES pages(id) ON DELETE CASCADE
... (10 total)
```

Cascade SHOULD have fired. It didn't. Without a doctor check, the orphans would have sat unnoticed for weeks (or forever).

## What

New `childTableOrphansCheck(engine: BrainEngine): Promise<Check>` helper in `src/commands/doctor.ts`, modeled after `takesWeightGridCheck` and the `jsonb_integrity` pattern:

- Scans 8 unique tables (10 FK-column slots; `links` has 3 FKs to pages)
- Two FKs declared `ON DELETE SET NULL` (`files.page_id`, `links.origin_page_id`) — NULL is a valid post-cascade state there, so the orphan predicate filters `IS NOT NULL AND col NOT IN (SELECT id FROM pages)` for those
- Per-table failure (older schema missing a table) is captured into a `warn` with "Could not check N/10 tables" — NEVER silently masked as `ok`
- Surfaces paste-ready cleanup SQL for every table that has orphans

## Sample output

Clean brain:
```
[ok] child_table_orphans: All FK-child tables clean (10 tables checked)
```

Brain with orphans:
```
[warn] child_table_orphans: 234 orphan row(s) in FK-child tables (content_chunks.page_id=234).
Cleanup: DELETE FROM content_chunks WHERE page_id NOT IN (SELECT id FROM pages);
```

## Tests

`test/doctor-child-orphans.test.ts` — 7 cases (36 expect calls) covering:

1. All clean → status:ok with "10 tables checked"
2. One orphan in content_chunks → status:warn with paste-ready cleanup SQL
3. Multi-table orphans → aggregated breakdown + per-table cleanup
4. Nullable-FK tables filter `IS NOT NULL` (regression guard)
5. `executeRaw` throws on older schemas → warn (no false-ok)
6. Orphans + skipped tables together → orphans branch wins (not masked)
7. All 8 unique target tables get queried (no silent drops)

Structurally-typed `BrainEngine` mock (no `mock.module`, no `*.serial.test.ts` rename) — the check's logic gets exercised cleanly. The full `gbrain doctor` integration with a live engine is covered by the existing 39 cases in `test/doctor.test.ts` (re-ran, all pass).

## Verification

- `bun run typecheck` — clean
- `bun run verify` — clean
- `bun test test/doctor-child-orphans.test.ts` — 7 pass, 0 fail
- `bun test test/doctor.test.ts` — 39 pass, 0 fail (the new check fires `ok` on the clean fixture brain)

## Reviewer notes

The check is read-only — no auto-repair. The fix-hint message gives paste-ready cleanup SQL but the user runs it. Reasoning: bulk DELETEs against a brain database are not the kind of thing doctor should auto-fire even with `--fix`. Operators should see the orphan counts, decide whether they're expected, and run the cleanup themselves. Same posture as `jsonb_integrity` (which points at `gbrain repair-jsonb` but doesn't auto-call it).

### DIFF
--- a/src/commands/doctor.ts
+++ b/src/commands/doctor.ts
@@ -193,6 +193,97 @@ export async function takesWeightGridCheck(engine: BrainEngine): Promise<Check> 
   }
 }
 
+/**
+ * Child-table orphan detection (closes #1063).
+ *
+ * The autopilot `orphans` phase (src/core/cycle.ts:runPhaseOrphans) detects
+ * orphan PAGES (pages with no inbound links via the page-graph). It does NOT
+ * scan FK-child tables for orphan rows. When a bulk page delete leaves
+ * orphans in `content_chunks` / `page_versions` / `tags` / `takes` / etc.
+ * — whether from pre-FK migrations, race conditions, or a code path that
+ * bypassed cascade — they persist indefinitely until manual SQL cleanup.
+ *
+ * All ten FK-to-pages tables declare `ON DELETE CASCADE` in the live schema
+ * (verified via `pg_constraint` snapshot in the issue body), so finding any
+ * orphan row is by definition unexpected. The check ships paste-ready
+ * cleanup SQL when orphans surface.
+ *
+ * Excluded: `files.page_id` and `links.origin_page_id` — both declared as
+ * `ON DELETE SET NULL`, so a NULL value is a valid state (file/link survives
+ * after page deletion); only NOT-NULL-but-page-missing is an orphan there.
+ * The check encodes that distinction for the two SET NULL columns.
+ *
+ * Pure helper for parity with `takesWeightGridCheck` so tests can target it
+ * directly without driving the full `runDoctor` pipeline.
+ */
+export async function childTableOrphansCheck(engine: BrainEngine): Promise<Check> {
+  // (table, fk_column, allow_null). When allow_null=true, NULL is a valid
+  // state (FK was declared ON DELETE SET NULL); the orphan predicate filters
+  // out NULL values. When false, NULL is impossible by NOT NULL constraint;
+  // any value not in pages.id is an orphan.
+  const targets: Array<{ table: string; col: string; allowNull: boolean }> = [
+    { table: 'content_chunks',   col: 'page_id',          allowNull: false },
+    { table: 'page_versions',    col: 'page_id',          allowNull: false },
+    { table: 'tags',             col: 'page_id',          allowNull: false },
+    { table: 'takes',            col: 'page_id',          allowNull: false },
+    { table: 'raw_data',         col: 'page_id',          allowNull: false },
+    { table: 'timeline_entries', col: 'page_id',          allowNull: false },
+    { table: 'links',            col: 'from_page_id',     allowNull: false },
+    { table: 'links',            col: 'to_page_id',       allowNull: false },
+    { table: 'links',            col: 'origin_page_id',   allowNull: true  },
+    { table: 'files',            col: 'page_id',          allowNull: true  },
+  ];
+  let totalOrphans = 0;
+  const breakdown: string[] = [];
+  const cleanupSql: string[] = [];
+  const errors: string[] = [];
+  for (const { table, col, allowNull } of targets) {
+    try {
+      // NOT IN subquery is portable across postgres + PGLite. The `pages.id`
+      // subquery covers every existing parent row.
+      const nullFilter = allowNull ? `${col} IS NOT NULL AND ` : '';
+      const rows = await engine.executeRaw<{ n: string | number }>(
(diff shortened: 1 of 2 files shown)


garrytan:
Thanks for this, @vincedk-alt — closing as superseded. Your work shipped in **v0.36.1.1**. From the [CHANGELOG](https://github.com/garrytan/gbrain/blob/master/CHANGELOG.md):\n\n> -  —  scans 10 FK columns across 8 tables for orphan rows; cascade-violation diagnostic with paste-ready cleanup SQL. Closes #1063. PR #1064.\n\nReal appreciation for the contribution.

PR #1071, opened 2026-05-16
Support DashScope embedding configuration

AzeWZ:
## Summary

- Honor configured `provider_base_urls` when DashScope is selected via `gbrain providers test --model ...`, so China-region smoke tests call the configured endpoint instead of the default international URL.
- Update the DashScope embedding recipe for `text-embedding-v4`, validate v3/v4 dimension options, and pin DashScope embedding requests to `encoding_format: "float"` through the compatibility shim.
- Add a per-provider embedding item cap and set DashScope batches to at most 10 input strings, matching the provider limit observed during `gbrain embed --stale`.
- Document the `provider_base_urls.dashscope` config shape and add regression coverage for request URL/body, recipe metadata, dimension passthrough, and item-cap batch splitting.

## Tests

- `bun test test/ai/recipe-dashscope.test.ts test/providers-dashscope-base-url.test.ts test/ai/gateway.test.ts`
- `bun test test/ai/adaptive-embed-batch.test.ts test/ai/recipe-dashscope.test.ts`
- `bun run typecheck`
- `bun test test/ai/zeroentropy-compat-fetch.test.ts test/voyage-response-cap.test.ts test/ai/dims-zeroentropy.test.ts test/ai/embedQuery.test.ts`
- `git diff --check`

## Manual verification

- `gbrain sync` against a local DashScope v4-backed brain no longer requests `OPENAI_API_KEY`.
- `gbrain embed --stale` completes after splitting DashScope requests into batches of at most 10 inputs.
- `gbrain doctor` reports `embedding_provider: dashscope:text-embedding-v4`, `1536 dims`, DB aligned, with 100% embedding coverage.


time-attack:
DashScope support was reviewed and declined as a direction for now — closing the DashScope PR set together. The generic per-recipe batch-cap concept from this family is being adopted separately with credit. Thank you for the work; if DashScope demand grows this decision can be revisited.

PR #1072, opened 2026-05-16
dims: thread Matryoshka dimensions for Qwen3-Embedding on Ollama

mgandal:
## Summary
- Adds Qwen3-Embedding (Alibaba) to the `openai-compatible` Matryoshka dimensions registry in `src/core/ai/dims.ts`
- Threads `dimensions: N` in the request body for `qwen3-embedding` and tag variants (e.g. `qwen3-embedding:4b`)
- 5 new unit tests; full AI suite 211/211 green

## Problem
Qwen3-Embedding (0.6B / 4B / 8B) is an official Apache-2.0 family in Ollama's trusted `library/` namespace, with strong MTEB scores (74.6 / 70.7 / 75.2 on Eng-v2). All variants are Matryoshka-truncatable and accept the OpenAI-compatible `dimensions` field on `/v1/embeddings`.

Without this passthrough, an existing 1536-dim brain swapping to `ollama:qwen3-embedding:4b` hard-fails with:
```
Embedding dim mismatch: model qwen3-embedding:4b returned 2560 but schema expects 1536.
```
because the Ollama server returns the model's native dim (2560 for 4B) instead of the requested 1536.

Tested locally end-to-end on a 39k-chunk brain: re-embed completes cleanly with `dimensions: 1536` in the request body, vectors verify against direct API calls (cosine = 1.0000), no schema migration needed.

## Why qwen3-embedding (and not gte-Qwen2)
`gte-Qwen2-1.5B-instruct` is the only mainstream model that emits 1536-dim *natively*, but it has no official Ollama tag — community uploads have documented correctness bugs (see ollama/ollama#10824, llama.cpp#14543). Qwen3-Embedding is official, official Ollama tag, and Matryoshka-truncatable. Cleaner provenance.

## Test plan
- [x] `bun test test/ai/recipe-ollama-dims.test.ts` — 5/5 pass (4 new positive cases + 1 regression guard)
- [x] `bun test test/ai/` — 211/211 pass on current upstream master
- [x] End-to-end: ran on a real 39k-chunk PGLite brain. Wiped all embeddings, re-embedded via `ollama:qwen3-embedding:4b` @ 1536-dim. Cosine = 1.0000 vs direct Ollama API call on canary text. Search continues to work post-migration.

## Notes / Future work
- Qwen3-Embedding is asymmetric (supports query/passage prefixes). The v0.35.0.0+ `inputType` parameter is plumbed but not used here — query-vs-document signal could be threaded in a follow-up. The current patch preserves symmetric behavior, matching how `gbrain embed` and `gbrain query` historically treated this provider.
- One cosmetic upstream quirk surfaced during testing: `content_chunks.model` doesn't update on re-embed (vector is correctly the new model's output, but the metadata column keeps the old name. Not in this PR's scope; happy to file a separate issue.

EOF
)


brettdavies:
Hi @mgandal, opened #2155 as a sibling to this one. Same root cause, different layer: this PR fixes the dispatch (threads `dimensions: N` so Ollama actually emits the requested width). #2155 fixes the validation gate (`isCustomDimValidForProvider` in `embedding-dim-check.ts`) that runs first and rejects any non-default Ollama dim before your code ever runs.

Hit it while pointing gbrain at `snowflake-arctic-embed-l-v2` (1024d) and `qwen3-embedding:8b` (4096d). The ollama recipe declares `default_dims: 768` (nomic) with no `dims_options`, so Tier 3 of the dim-check refuses. Same hole for `llama-server` and `litellm-proxy` despite their `user_provided_models: true` flag. #2155 adds a passthrough tier; the PGVECTOR cap and positive-int check still apply.

Both PRs are independently useful and either order of landing works. Wanted to flag in case it's helpful to consider together.

PR #1076, opened 2026-05-16
fix(serve): return 405 on GET /mcp instead of 404

lukejduncan:
## Summary

Per the MCP Streamable HTTP spec, servers that don't support the SSE backchannel MUST respond `405 Method Not Allowed` (with an `Allow` header) on `GET /mcp`. gbrain's transport is stateless and only registers `POST /mcp`, so unhandled `GET /mcp` falls through to Express's default 404 HTML handler.

This breaks claude.ai's Custom Connector flow: its initial reachability probe is a `GET` to the MCP URL, and a 404 makes it conclude the URL isn't an MCP server — surfaced as "Couldn't reach the MCP server" in the UI.

## Fix

Add a no-auth `GET /mcp` handler that returns 405 with a JSON-RPC error body and `Allow: POST, DELETE`. Auth still gates the `POST` path.

## Test plan

- [x] `curl https://server/mcp` returns `405` (was `404`)
- [x] `curl -X POST https://server/mcp ...` still returns `401` without bearer auth
- [x] Authenticated `POST /mcp initialize` still returns valid server info
- [x] claude.ai Custom Connector now reaches gbrain through `/.well-known/oauth-authorization-server` discovery

### DIFF
--- a/src/commands/serve-http.ts
+++ b/src/commands/serve-http.ts
@@ -794,6 +794,15 @@ export async function runServeHttp(engine: BrainEngine, options: ServeHttpOption
   // ---------------------------------------------------------------------------
   const mcpOperations = operations.filter(op => !op.localOnly);
 
+  // MCP Streamable HTTP spec: GET /mcp opens an optional SSE backchannel for
+  // server-initiated messages. gbrain's transport is stateless and doesn't push
+  // server-initiated messages, so per spec we MUST return 405 (not 404) here so
+  // probing clients (claude.ai, others) recognize this as an MCP endpoint.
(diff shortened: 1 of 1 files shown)


garrytan:
Thank you for this work @lukejduncan — closing as already in master at `src/commands/serve-http.ts:1246-1248`. The `GET /mcp` handler returns 405 with `Allow: POST, DELETE` exactly as your PR specified.

Verified during the v0.41.3.0 fix wave (https://github.com/garrytan/gbrain/pull/1403) scope challenge: the handler is wired and the MCP Streamable HTTP spec compliance is in place. Closing for queue cleanup.

PR #1079, opened 2026-05-16
fix: sync all registered sources in autopilot cycle instead of single brainDir

### LINKED ISSUE #1078: Autopilot/dream sync fails with FK constraint error in multi-source brains
## Bug Description

When running `gbrain autopilot` or `gbrain dream` in a **multi-source brain** (e.g. multiple sources registered via `gbrain sources add`), the sync phase fails with a foreign key constraint violation:

```
insert or update on table "pages" violates foreign key constraint "pages_source_id_fkey"
```

This causes **all** file imports in the cycle to fail silently, leaving the brain stale.

## Root Cause

In `src/core/cycle.ts`, `runPhaseSync()` resolves the source ID by matching the autopilot's `--repo` path against `sources.local_path`:

```typescript
// src/core/cycle.ts:558
const sourceId = await resolveSourceForDir(engine, brainDir);
```

`resolveSourceForDir()` does an **exact match**:
```sql
SELECT id FROM sources WHERE local_path = $1 LIMIT 1
```

When `--repo /home/ubuntu/brain` is passed but all sources have subdirectory paths like:
- `logseq` → `/home/ubuntu/onedrive/OneSyncFiles/`
- `wiki` → `/home/ubuntu/brain/wiki/`
- `obsidian` → `/home/ubuntu/onedrive/my_obsidian/`

No source matches, so `sourceId` resolves to `undefined`. The `importFile()` then falls back to the schema DEFAULT for `source_id`, which is `'default'`:

```sql
-- pages.source_id has default 'default'
```

But **no row exists in `sources` with `id = 'default'`**, triggering the FK constraint violation on every insert.

## Reproduction Steps

1. Register multiple sources with different `local_path` values:
   ```bash
   gbrain sources add wiki --path /home/ubuntu/brain/wiki
   gbrain sources add obsidian --path /home/ubuntu/onedrive/my_obsidian
   gbrain sources add logseq --path /home/ubuntu/onedrive/OneSyncFiles
   ```

2. Start autopilot with a parent directory that doesn't match any source:
   ```bash
   gbrain autopilot --repo /home/ubuntu/brain --interval 900
   ```

3. Observe sync phase: all files fail with `pages_source_id_fkey` violation.

## Environment

- gbrain v0.35.1.0
- Engine: postgres (Supabase)
- Sources: 4 (logseq, wiki, obsidian, books)
- OS: Linux arm64

## Proposed Fix

Modify `runPhaseSync()` to enumerate all registered sources from the database and sync each one individually, rather than relying on a single `brainDir` match:

```typescript
// Instead of resolving one source for brainDir:
const sourceId = await resolveSourceForDir(engine, brainDir);

// Enumerate all sources and sync each:
const sources = await engine.executeRaw<{ id: string; local_path: string }>(
  `SELECT id, local_path FROM sources WHERE local_path IS NOT NULL AND local_path != ''`,
);

for (const src of sources) {
  const result = await performSync(engine, {
    repoPath: src.local_path,
    sourceId: src.id,
    // ...
  });
}
```

This ensures each source is synced with its correct `source_id`, avoiding the FK constraint error. The fallback path is preserved for brains with no registered sources.

I can submit a PR with the full patch if this approach is acceptable.

## Workaround

Manually sync each source individually:
```bash
gbrain sync --source logseq
gbrain sync --source wiki
gbrain sync --source obsidian
```

samvilian:
## Summary

Fixes #1078

In multi-source brains, `runPhaseSync()` resolves the source ID by matching `brainDir` against `sources.local_path`. When `--repo` points to a parent directory that doesn't match any source (e.g. `/home/ubuntu/brain` vs `/home/ubuntu/brain/wiki/`), `sourceId` resolves to `undefined`, causing `importFile()` to fall back to the schema DEFAULT `'default'` for `source_id`. Since no row with `id='default'` exists in `sources`, every INSERT violates `pages_source_id_fkey`.

## Changes

- **`src/core/cycle.ts`**: `runPhaseSync()` now enumerates all registered sources from the database and syncs each one individually with its correct `source_id` and `local_path`
- Aggregates per-source results into a unified `SyncPhaseResult`
- Reports per-source status in `details.sourcesSynced` for observability
- Preserves the legacy single-dir fallback when no sources are registered

## Before

```
[cycle.sync] start
[gbrain phase] sync.fullsync.import start strategy=markdown
Warning: skipped wiki/.../file.md: insert or update on table "pages" violates foreign key constraint "pages_source_id_fkey"
  56 files failed
sync_status: blocked_by_failures
```

## After

```
[cycle.sync] start
[gbrain phase] sync.git_pull start  (logseq)
[gbrain phase] sync.git_pull start  (wiki)
[gbrain phase] sync.git_pull start  (obsidian)
sync_status: completed
sourcesSynced: ["logseq: ok", "wiki: ok", "obsidian: ok", "books: error (not a git repo)"]
```

## Test

Verified locally with 4 registered sources (logseq, wiki, obsidian, books) — all FK constraint errors resolved, sync completes successfully across all git-initialized sources.

### DIFF
--- a/src/core/cycle.ts
+++ b/src/core/cycle.ts
@@ -573,39 +573,105 @@ async function runPhaseSync(
 ): Promise<SyncPhaseResult> {
   try {
     const { performSync } = await import('../commands/sync.ts');
-    // Resolve the per-source id so sync reads source-scoped last_commit
-    // instead of the global config key. The global key can drift out of
-    // git history (force push, GC) causing a full reimport of all files.
-    const sourceId = await resolveSourceForDir(engine, brainDir);
-    const result = await performSync(engine, {
-      repoPath: brainDir,
-      sourceId,
-      dryRun,
-      noPull: !pull,
-      noEmbed: true,                       // embed is a separate phase
-      noExtract: willRunExtractPhase,      // dedupe ONLY when cycle's extract phase will also run.
-                                           // If extract isn't scheduled (e.g. `gbrain dream --phase sync`),
-                                           // sync's inline extract still runs to preserve prior behavior.
-    });
-    const syncedCount = result.added + result.modified;
+
+    // Multi-source sync: enumerate all registered sources with local_path
+    // and sync each one individually. This avoids the FK constraint error
+    // that occurs when brainDir doesn't match any source's local_path
+    // (causing sourceId to resolve as undefined, falling back to the
+    // schema DEFAULT 'default' which may not exist in sources table).
+    const sources = await engine.executeRaw<{ id: string; local_path: string }>(
+      `SELECT id, local_path FROM sources WHERE local_path IS NOT NULL AND local_path != ''`,
+    );
+
+    if (sources.length === 0) {
+      // No registered sources — fall back to legacy single-dir sync.
+      const sourceId = await resolveSourceForDir(engine, brainDir);
+      const result = await performSync(engine, {
+        repoPath: brainDir,
+        sourceId,
+        dryRun,
+        noPull: !pull,
+        noEmbed: true,
+        noExtract: willRunExtractPhase,
+      });
+      const syncedCount = result.added + result.modified;
+      return {
+        phase: 'sync',
+        status: result.status === 'blocked_by_failures' ? 'warn' : 'ok',
+        duration_ms: 0,
+        summary: dryRun
+          ? `${syncedCount} page(s) would sync, ${result.deleted} would delete`
+          : `+${result.added} added, ~${result.modified} modified, -${result.deleted} deleted`,
+        details: {
+          added: result.added,
+          modified: result.modified,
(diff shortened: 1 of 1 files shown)


time-attack:
Thanks for the clear repro of #1078 — the FK violation from `sourceId` falling back to a non-existent `'default'` row is real and worth fixing. This shape isn't the one we want, though: it silently redefines the sync phase to touch every registered source regardless of the `--repo` the caller passed, which is a large behavior change for anyone running a cycle against a single repo, and it does so via raw SQL that bypasses our source-scoping helpers, with no tests on a 92-line rewrite of a core phase. The root cause is `resolveSourceForDir` returning undefined and the import silently defaulting instead of failing loudly — a fix at that point would be far smaller. Closing.

PR #1084, opened 2026-05-16
fix: declare entity_hints array item schema

zzdisturbed:
## Summary
- Adds `items: { type: "string" }` to the `extract_facts.params.entity_hints` MCP schema.
- Fixes strict tool-schema validation failures in OpenAI/Codex-compatible clients.

## Context
Without `items`, clients reject the entire tool payload before any model response with:

```
Invalid schema for function mcp_gbrain_extract_facts: In context=('properties', 'entity_hints'), array schema missing items.
```

## Test plan
- `bun run typecheck`
- `hermes mcp test gbrain` locally discovers 70 tools after this patch.


garrytan:
Thanks @zzdisturbed — `extract_facts.entity_hints` (and the other MCP array-schema `items` fixes) already ship in master as of **v0.35.3.0**. The shared mapper is `paramDefToSchema` in `src/mcp/tool-defs.ts` (used by stdio MCP, HTTP MCP, and the subagent registry); `test/mcp-tool-defs.test.ts` walks every array param and fails the suite if any lacks `items.type`. If your install is still hitting this on next `gbrain upgrade`, please reopen with the output of `gbrain doctor --json`.

PR #1088, opened 2026-05-16
fix(migrate): preserve embedding_model and embedding_dimensions on engine switch

samchaudhary:
## Summary

Fixes a silent config-stripping bug in `gbrain migrate --to <engine>`. Existing brains keep working after the migration but new content can't be embedded, because the migration wipes the embedding configuration. I found this last night running my own gbrain.

## Issue

`gbrain migrate --to <engine>` writes a new config that's just `{ engine, database_url }` (or `database_path`) and nothing else. Every other field is silently dropped: `embedding_model`, `embedding_dimensions`, `expansion_model`, `chat_model`, API keys, `storage`, `eval`, `provider_base_urls`.

The migration appears to succeed. The new database has all the pages and vectors. Search still works on existing content. The issue only shows up when you try to embed something new, or when you read `gbrain doctor` output carefully.

## Reproduce

1. Start with a brain on PGLite with custom embeddings:

```json
{
  "engine": "pglite",
  "database_path": "$HOME/.gbrain/brain.pglite",
  "embedding_model": "ollama:nomic-embed-text",
  "embedding_dimensions": 768
}
```

2. Migrate to Postgres:

```bash
gbrain migrate --to postgres --url postgresql://localhost/gbrain
```

3. Check the new config:

```json
{
  "engine": "postgres",
  "database_url": "postgresql://localhost/gbrain"
}
```

`embedding_model` and `embedding_dimensions` are gone. `gbrain doctor` reports:

```
[OK] embedding_provider: Skipped (no provider credentials). Model: openai:text-embedding-3-large.
```

Existing search keeps working because the 768-dim vectors are in place. But new content fails to embed: gbrain tries the OpenAI default model (1536-dim) and breaks either at the API (no key) or at insert (dimension mismatch with the `vector(768)` column).

## Fix

Extracted a pure `buildMigratedConfig()` helper in `src/commands/migrate-engine.ts`. It destructures the three engine-shape fields (`engine`, `database_url`, `database_path`) and spreads the rest of the source config into the new one before stamping the target engine and its connection field. The stale connection field from the source engine is stripped, so a migrated config never holds both `database_url` and `database_path`.

The production migration command calls the helper, so the tests exercise the real production function rather than a duplicate.

## Tests

New file: `test/migrate-engine-config.test.ts`. Nine unit cases:

- pglite ↔ postgres in both directions, embedding fields preserved
- stale connection field stripped on each target engine
- engine-mismatched source (both `database_url` and `database_path` set) gets cleaned up
- `expansion_model`, `chat_model`, and API keys preserved
- nested fields preserved (`storage`, `eval`, `provider_base_urls`)
- minimal config (engine + path only) migrates cleanly
- source config not mutated

9/9 pass. Typecheck clean. 153 tests pass across the broader migrate/config test suite.

## Environment

- macOS 26.3
- gbrain master (also reproducible on v0.34.4)
- Migrating PGLite → local Postgres 17 (also affects Postgres → PGLite)

### DIFF
--- a/src/commands/migrate-engine.ts
+++ b/src/commands/migrate-engine.ts
@@ -74,6 +74,32 @@ function clearManifest(): void {
   if (existsSync(path)) unlinkSync(path);
 }
 
+/**
+ * Build the post-migration file-plane config. Preserves every non-engine
+ * field from the source config (embedding_model, embedding_dimensions,
+ * expansion_model, chat_model, API keys, storage, eval, etc.) and swaps in
+ * the target engine + its connection field. Stale connection fields from the
+ * source engine are stripped so a migrated config never carries both
+ * database_url and database_path.
+ *
+ * Exported for unit testing; called from runMigrateEngine() right before
+ * saveConfig().
+ */
+export function buildMigratedConfig(
+  source: GBrainConfig,
+  targetEngine: 'postgres' | 'pglite',
+  targetConfig: EngineConfig,
+): GBrainConfig {
+  const { database_url: _u, database_path: _p, engine: _e, ...preserved } = source;
+  return {
+    ...preserved,
+    engine: targetEngine,
+    ...(targetEngine === 'postgres'
+      ? { database_url: targetConfig.database_url }
+      : { database_path: targetConfig.database_path }),
+  };
+}
+
 export async function runMigrateEngine(sourceEngine: BrainEngine, args: string[]): Promise<void> {
   const opts = parseArgs(args);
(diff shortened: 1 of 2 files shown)


time-attack:
Thanks for this, @samchaudhary! 🙏

What this does: This PR fixes a bug where migrating between database engines would strip the user's embedding configuration. Master already includes this fix via v0.37.11.0, released 5 days after the PR was opened, making it superseded.

PR #1091, opened 2026-05-17
feat(search): opt-in floor-ratio gate for post-fusion boost stages

jayzalowitz:
## What

Adds an optional `floorRatio?: number` to `applyBacklinkBoost`, `applySalienceBoost`, `applyRecencyBoost`, and `PostFusionOpts`. When set, each boost stage skips results whose pre-boost score is below `floorRatio * topScore` at the moment that stage runs — only the head of the candidate pool receives the multiplicative bonus. Default `undefined` preserves exact prior behavior bit-for-bit.

## The failure mode this addresses

gbrain's bounded boosts (log-compressed salience clipped to `[1.0, ~1.6]`, log-scaled backlink factor, half-life-decayed recency) work as designed on curated test corpora. The existing comments in `hybrid.ts:69` make the design intent explicit: *"a strong boost can't catastrophically flip rankings."*

That guarantee weakens on larger corpora indexed with real high-dimensional embedders (`text-embedding-3-large`, `voyage-3-large`, `voyage-4-large`, `zembed-1`). Baseline vector similarity between topically-unrelated "professional content" is non-trivial — weak-overlap pages routinely land in a query's top-K via vector overlap alone, each receives the multiplicative boost, and on a non-trivial fraction of queries a weak page with high metadata signal climbs above the legitimate primary hit. The per-boost factor still looks harmless in isolation; the compound effect across the long tail of top-K candidates is what shifts ranks.

Concrete example from the new test suite:

| page              | raw RRF score | backlink count | post-boost score |
|-------------------|--------------:|---------------:|-----------------:|
| `strong-primary`  | 1.00          | 0              | 1.000            |
| `weak-with-signal`| 0.50          | 1000           | 0.673            |

`strong-primary` still wins in that pair, but the gap is now ~50% of what the raw text+vector signal said it should be. Drop `strong-primary`'s raw score below ~1.35× the weak page (common in the long tail) and rank flips, even though every component signal said the strong page was the unambiguous match.

## The fix

A boost only fires for results within `floorRatio × topScore` at the moment that stage runs. The long tail keeps its unboosted score and original rank. Stages compose naturally — salience boosts against its own top, then recency runs against the post-salience top, etc. A truly-strong primary always wins regardless of boost magnitude on weaker candidates.

`0.85` is a reasonable starting point for dense embedder corpora — it covers the head of a typical RRF candidate pool while excluding weak-overlap distractors. The value comes from a labeled-retrieval ablation in the [SkyTwin](https://github.com/jayzalowitz/skytwin) twin-memory layer (which consumes gbrain as its default memory backend via [skytwin#250](https://github.com/jayzalowitz/skytwin/pull/250)): `0.85` is the largest ratio that fully eliminated the leapfrog regression on our labeled corpus while preserving baseline rankings on queries that don't have a metadata signal to bias. The relevant ablation is in [skytwin#272](https://github.com/jayzalowitz/skytwin/pull/272).

## Backward compatibility

`floorRatio` defaults to `undefined` → no gate, no threshold computation, exact prior behavior. Existing call sites are untouched; the new parameter is positional-last and optional on each function. `PostFusionOpts.floorRatio` is similarly optional and unset by default.

The gate is opt-in by design — it changes ranking behavior, and existing adopters have tuned around the current post-fusion stack. Surfacing it as a flag lets each consumer evaluate against their own corpus before flipping it on, rather than landing a behavior change in a patch release.

## Tests

7 new cases in `test/search.test.ts`:

- `floorRatio` undefined preserves existing behavior bit-for-bit
- weak page gated out, top page boosted as before
- borderline page at exactly the threshold is eligible (inclusive boundary)
- regression scenario: weak page with strong metadata signal cannot leapfrog strong primary
- `applySalienceBoost` honors the gate (parity with `applyBacklinkBoost`)
- empty results no-op without divide-by-zero
- single-result trivially eligible (page is its own top)

`bun test test/search.test.ts`: **33/33 pass** (was 26/26).
`bun run verify`: pass (typecheck + 12 guard scripts).

## Why I'm sending this

We've been running gbrain as the production memory backend in SkyTwin for the last week, and this came out of the labeled-retrieval ablation we did on top of it. The gate is general-purpose — the shape applies to any per-result boost on metadata orthogonal to the raw text+vector signal — so it seemed worth offering upstream rather than keeping it forked. The opt-in framing makes it cheap to merge if it lands cleanly and equally cheap to leave un-flipped if your shootout corpus doesn't surface the same regression.

### DIFF
--- a/src/core/search/hybrid.ts
+++ b/src/core/search/hybrid.ts
@@ -47,14 +47,72 @@ const DEBUG = process.env.GBRAIN_SEARCH_DEBUG === '1';
  * Apply backlink boost to a result list in place. Mutates each result's score
  * by (1 + BACKLINK_BOOST_COEF * log(1 + count)). Pure data transform; no DB call.
  * Caller fetches counts via engine.getBacklinkCounts.
+ *
+ * `floorRatio` (optional, opt-in): when set, the boost only fires for
+ * results whose pre-boost score is ≥ `floorRatio * topScore`. Pages
+ * below the gate keep their unboosted score. Default undefined → no gate
+ * (preserves existing behavior). See `computeFloorThreshold` for rationale.
  */
-export function applyBacklinkBoost(results: SearchResult[], counts: Map<string, number>): void {
+export function applyBacklinkBoost(
+  results: SearchResult[],
+  counts: Map<string, number>,
+  floorRatio?: number,
+): void {
+  const threshold = computeFloorThreshold(results, floorRatio);
   for (const r of results) {
+    if (r.score < threshold) continue;
     const count = counts.get(r.slug) ?? 0;
     if (count > 0) {
       r.score *= (1.0 + BACKLINK_BOOST_COEF * Math.log(1 + count));
     }
   }
+}
+
+/**
+ * Floor-ratio gate (opt-in, default off).
+ *
+ * Returns `-Infinity` when `floorRatio` is undefined → the gate never
+ * fires (existing behavior preserved). When provided → returns
+ * `topScore * floorRatio`; callers skip the boost for any result with
+ * `r.score < threshold`.
+ *
+ * **Why this exists.** gbrain's bounded boosts (the `[1.0, ~1.6]` clip
+ * range from log-compressed salience, the log-scaled backlink factor,
+ * the half-life-decayed recency factor) keep any single boost from
+ * catastrophically flipping rankings. That guarantee holds for small
+ * curated corpora.
+ *
+ * On larger corpora indexed with real high-dimensional embedders
+ * (text-embedding-3-large, Voyage 3-large, Voyage 4, etc.) the baseline
+ * vector similarity between topically-unrelated "professional content"
+ * is non-trivial. Weak-overlap pages land in a query's top-K via vector
+ * overlap alone, receive the multiplicative boost, and leapfrog the
+ * legitimate primary hit. The boost magnitudes are small enough that
+ * any single comparison still feels "safe," but in the long tail of
+ * top-K shuffles the regression is visible.
+ *
+ * The gate cuts each boost's reach. With `floorRatio: 0.85`, only
+ * results within 85% of `topScore` are eligible; the long tail keeps
+ * its unboosted score and original rank. A truly-strong primary
+ * always wins regardless of boost magnitude on weaker candidates.
+ *
+ * Empirical motivation comes from a labeled-retrieval ablation we ran
+ * in the SkyTwin twin-memory layer that consumes gbrain
+ * (https://github.com/jayzalowitz/skytwin/pull/272 — switched to a
+ * real OpenAI-shape embedder, observed weak-overlap authored pages
+ * climbing into top-K via vector overlap, added the floor-ratio gate
+ * to fix it). Default-off because we cannot reproduce the failure mode
+ * from outside on gbrain's own eval fixtures — surfacing it as an
+ * opt-in lets adopters with large dense corpora opt in without
+ * changing the default behavior anyone else relies on.
+ */
+function computeFloorThreshold(results: SearchResult[], floorRatio?: number): number {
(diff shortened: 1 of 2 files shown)


garrytan:
Hi @jayzalowitz — thank you for this. The labeled-retrieval ablation behind it is the kind of contribution that makes gbrain better than a single-corpus tool, and the failure-mode framing is sharp.

After running this through `/plan-eng-review` and `/codex` (outside-voice adversarial review), we integrated it on top of master in #1129 with some refactoring. Summary of what changed and why:

**Three correctness fixes codex caught:**

1. **Cache contamination via `knobsHash()`.** Without bumping `KNOBS_HASH_VERSION` 2→3 in `src/core/search/mode.ts`, a cached no-floor result gets served to a floor-enabled call. Same bug class CDX-4 v0.32.3 hotfix closed for the other search-lite knobs. Direct ranking-correctness leak — the PR would have shipped with this.
2. **NaN scores bypass the gate.** `NaN < threshold` is false in JS, so a NaN-scored row would have skipped the floor and received boosts. Realistic on Voyage flexible-dim or zembed-1 Matryoshka dim drift across reindexes. Fixed: NaN scores skip the boost entirely.
3. **Negative top scores break "single result trivially eligible".** Your single-result test passed because score was positive; with a negative top score (-0.5), threshold becomes -0.425 and the top itself fails its own floor. Edge case but real on RRF outputs that can go negative.

**Two architectural shape changes:**

- **Single up-front threshold vs per-stage recompute.** Your PR computes the threshold separately at each stage (each of backlink → salience → recency uses `topScore × ratio` from the array state at that moment). Codex argued this makes stage order part of the API — a backlink boost to the leader raises the salience/recency floor and gates out a page that was originally within `0.85 × top`. We refactored to compute the threshold ONCE at `runPostFusionStages` entry from the post-cosine-rescore snapshot, then pass the same number to all three stages. Order-independent.
- **Public surface beyond `PostFusionOpts`.** Wired `SearchOpts.floorRatio` (per-call), `search.floor_ratio` config key (operator default — included in `SEARCH_MODE_CONFIG_KEYS`), and a `floor_ratio` field on `MODE_BUNDLES` (currently `undefined` for all three modes pending gbrain-side ablation). No `GBRAIN_SEARCH_FLOOR_RATIO` env var — `resolveSearchMode()` is pure by design, and a hidden env knob would make `gbrain search modes` lie about the resolved state.

**API surface change:** the three boost functions now take `floorThreshold?: number` (an absolute score floor) instead of `floorRatio?: number`. `runPostFusionStages` is the single ratio→threshold converter. This is what makes the single-baseline semantic enforceable through the type system. Your tests' shape stays similar; they update from passing `0.85` to passing `0.85 * topScore` inline.

**What stays exactly as you shipped:**

- Default-off posture is correct and preserved.
- Empirical framing (dense-embedder targeting, `0.85` starting value, leapfrog regression mechanism) is unchanged — your ablation is what motivates the whole thing.
- Gate scope (metadata-axis boost stages only, exact-match runs independently) matches your design intent.

You're credited via `Co-Authored-By: Jay Zalowitz <jayzalowitz@gmail.com>` on the integration commit, and the CHANGELOG section in #1129 explicitly names you and links your SkyTwin ablation as the empirical motivation. The full 9-decision review trail (D1-D9) is at `~/.claude/plans/swift-sniffing-nygaard.md` in the repo if you want to see exactly which questions came up and how we landed on the integrated shape.

I'll mark this PR as superseded by #1129 (which carries `Closes #1091` so it'll auto-close on merge). If anything about the integration looks wrong, please flag it on #1129 — we have ~24h before merge for any final adjustments.

Thanks again. Filed two follow-up TODOs that grew out of this work: (1) gbrain-side ablation across `longmemeval` / `whoknows` / `suspected-contradictions` / BrainBench-Real before flipping any mode-bundle default, and (2) per-source threshold for federated-read users if real usage surfaces the cross-source suppression codex flagged.

Issue #972, opened 2026-05-13
Feature: opt-in global-basename wikilink resolution for Obsidian-convention vaults


matnotkin:
## Problem

When `gbrain extract links` ingests `[[wikilinks]]` from markdown files, the resolver only matches by:

1. **Parallel directory + same basename** (e.g., a link from `originals/foo.md` resolves `[[bar]]` to `originals/bar.md`)
2. **Same directory + sibling basename** (e.g., a link from `concepts/foo.md` resolves `[[bar]]` to `concepts/bar.md`)

It does **not** do a global-basename lookup. So a wikilink like `[[struktura]]` written in `concepts/knowledge-graph.md` does **not** resolve to `projects/struktura.md`, even though that page exists, has a unique basename in the vault, and resolves correctly in Obsidian.

This silently drops cross-directory bare-basename wikilinks. They're not flagged as broken — they're just missing from gbrain's graph layer.

## Repro

```bash
# Set up a minimal vault with cross-directory link
mkdir -p /tmp/vault/{concepts,projects}
cat > /tmp/vault/projects/struktura.md <<EOF
---
title: Struktura
type: project
---
A project page.
EOF
cat > /tmp/vault/concepts/knowledge-graph.md <<EOF
---
title: Knowledge Graph
type: concept
---
This concept relates to [[struktura]].
EOF

gbrain init --force --dir /tmp/test-brain
GBRAIN_HOME=/tmp/test-brain gbrain import /tmp/vault
GBRAIN_HOME=/tmp/test-brain gbrain extract links
GBRAIN_HOME=/tmp/test-brain gbrain stats
# Links: 0 (expected: 1)
```

Obsidian's graph view correctly links the two nodes. Gbrain's graph layer does not.

## Why it matters

For users who follow Obsidian's standard convention (flat-namespace wikilinks, globally-unique basenames), this silently drops most of their interlinking from gbrain's graph layer. We hit this on a fresh wiki: 71 wikilinks total across 20 pages in Obsidian (avg 3.55/page, dense galaxy), but gbrain's `extract links` only produced 12 — all of which were same-directory. The remaining ~60 cross-directory bare-basename links were dropped silently.

Symptoms downstream:
- `gbrain graph <slug>` returns thin or missing edges that exist in the source files
- `gbrain backlinks <slug>` undercounts incoming references
- Graph-traversal queries (e.g., "what concepts touch this project") return incomplete neighbors
[...]

PR #1101, opened 2026-05-17
fix: support root-level file references in ENTITY_REF_RE

essexcanning:
The Dir-prefix regex (ENTITY_REF_RE) only matched markdown links targeting dir-slugged paths like people/alice or companies/acme. Root-level files like action-tracker.md, decisions-log.md, and home.md — common in Obsidian and personal knowledge vaults — were silently skipped, leaving those pages unlinked in the knowledge graph.

This extends ENTITY_REF_RE with an alternation: after matching the optional ../ or ./ prefix, accept EITHER a dir-slugged path (the existing behavior) OR a bare filename without directory prefix. External URLs are still excluded (the [^)\\s\/:]+ class omits :, so https://... never matches).

The FS-source path (extractMarkdownLinks) already handled root-level files without restriction; this brings the DB-source path (ENTITY_REF_RE) to parity.

All 21 existing extract tests continue to pass.

### DIFF
--- a/src/core/link-extraction.ts
+++ b/src/core/link-extraction.ts
@@ -56,7 +56,7 @@ const DIR_PATTERN = '(?:people|companies|meetings|concepts|deal|civic|project|pr
  * `.md` suffix so the same function works for both filesystem and DB content.
  */
 const ENTITY_REF_RE = new RegExp(
-  `\\[([^\\]]+)\\]\\((?:\\.\\.\\/)*(${DIR_PATTERN}\\/[^)\\s]+?)(?:\\.md)?\\)`,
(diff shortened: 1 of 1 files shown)


time-attack:
Thanks — the gap you describe is real, root-level pages are not matched by ENTITY_REF_RE. The proposed widening is too broad, though: `[^)\s\/:]+` does not exclude `#`, so `[x](#some-heading)` becomes an entity ref, as does `[img](photo.png)` and any relative asset link, and `extractEntityRefs` then derives `dir` from `split('/')[0]` and fabricates a directory for each one. These also skip the `needsResolution` path that the #972 bare-wikilink work added specifically so bare names go through the config-gated basename resolver rather than landing in the graph unresolved. A change here needs to route through that resolver and come with tests covering anchors and asset links. Closing.

PR #1107, opened 2026-05-17
orphan-metric: align find_orphans surfaces with get_health and tighten exclusions

colin477:
## Summary

Four-commit chain that gives `get_health.orphan_pages` honest, accurate semantics and aligns the user-visible orphan surfaces (`gbrain orphans` CLI + `find_orphans` MCP) with it.

| Commit | What | |
|---|---|---|
| 6a945560 | Exclude source-type ingestions (`emails/`, `attachments/`, `0-daily/`, `4-archive/`) from `orphan_pages` SQL | Was: ingestion-by-design pages dominated the count and crushed `brain_score`. |
| a70ef64b | Exclude soft-deleted pages (`deleted_at IS NULL`) from `orphan_pages` SQL | Was: soft-deleted pages still consumed by the SQL count, distorting `no_orphans_score`. |
| 54d349a1 | Exclude pseudo-page hub prefixes (`templates/`, `navigation/`) from `orphan_pages` SQL | Was: vault wayfinding pages reflect `[[wikilink]]` sparsity, not graph rot. |
| 7e6cb77b | Align `find_orphans` (CLI + MCP) with `getHealth.orphan_pages` via shared islanded predicate; add `engine.findIslandedPages()` | Was: CLI used a no-inbound JS filter; MCP get_health used SQL islanded. Different counts. |

Net effect: one authoritative SQL predicate, shared between `getHealth` (`orphan_pages`/`no_orphans_score`/`brain_score`) and `find_orphans` (CLI + MCP). The two surfaces agree by construction.

## The headline alignment fix (#7e6cb77b)

`find_orphans` used to scan for "no inbound wikilinks" then apply a JS-side `shouldExclude` filter. `getHealth.orphan_pages` used an SQL islanded predicate (no in **and** no out) with only the 4–6 source/wayfinding prefixes from earlier commits. Two different definitions → two different counts.

This commit:
- Promotes the full `shouldExclude` set to SQL inside `getHealth.orphan_pages` (both engines): pseudo-page exact slugs (`_atlas`, `_index`, `_stats`, `_orphans`, `_scratch`, `claude`), suffix patterns (`/_index`, `/log`), segment patterns (`/raw/`), additional deny prefixes (`output/`, `dashboards/`, `scripts/`, `openclaw/config/`), and first-segment exclusions (`scratch/`, `thoughts/`, `catalog/`, `entities/`). LIKE escape `'%/#_index' ESCAPE '#'` for the literal `_index` suffix match (bare `_` is a single-char wildcard in LIKE).
- Adds `engine.findIslandedPages(opts?: { includePseudo?: boolean })` to the interface (both engines) sharing the same SQL predicate. `includePseudo=true` relaxes the exclusion set but does **not** change the islanded graph predicate.
- Routes `gbrain orphans` CLI + `find_orphans` MCP through `findIslandedPages`. The JS `shouldExclude` is `@deprecated` and kept exported only for backwards-compat unit tests.
- Keeps `findOrphanPages()` (no-inbound) as the engine-level helper for enrichment-pattern consumers (`queryOrphanPages`, future link-recovery tooling).
- Updates docstrings on `BrainHealth.orphan_pages`, the engine interface, CLI help text + summary line ("N orphans" → "N islanded pages out of K total"), `find_orphans` MCP description, and the runCycle orphans phase summary.
- Drops the post-hoc `total_linkable` + `excluded` fields from `OrphanResult` — they were JS computations over the no-inbound set; with SQL authoritative they degenerate to misleading values (`excluded` = 0, `total_linkable` = `total`). Cycle phase details + cycle.serial.test.ts mock updated accordingly.

## Tests

- Replace fixtures that no longer match islanded semantics (an `alice→bob` link makes neither islanded under the new definition).
- 20-row `test.each` covering every prefix/exact/suffix/segment exclusion.
- LIKE-escape defense: `weird/_zindex` must NOT match the `%/_index` suffix pattern (regression guard for the bare `_` wildcard trap).
- Soft-delete fixture now uses `softDeletePage` and asserts `deleted_at IS NOT NULL` (codex post-impl caught that prior fixture used hard-delete, which would pass even with the `deleted_at IS NULL` predicate removed — false-positive coverage).
- **Critical alignment regression**: `findOrphans(engine).total_orphans === (await engine.getHealth()).orphan_pages`.

60/60 orphan tests pass; 143/143 in adjacent regression files (brain-score-breakdown, doctor, pglite-engine); 28/28 in cycle.serial; typecheck clean.

## Live impact (one brain, ~10K pages, gbrain.colin)

| Metric | Pre | Post | Δ |
|---|---:|---:|---|
| CLI `gbrain orphans --count` | 2722 | 2656 | **-66** |
| `get_health.orphan_pages` | 2699 | **2656** | **-43 (aligned)** |
| `brain_score` | 67 | 67 | 0 |
| `no_orphans_score` | 11/15 | 11/15 | 0 |

`no_orphans_score` coincidentally lands on the same rounding boundary so `brain_score` is unchanged — the fix is **structural alignment**, not metric movement.

## Out of scope (deliberately)

- A shared SQL fragment/helper across both engines. Codex flagged "drift risk" (P2) — predicate is duplicated 4× now (getHealth + findIslandedPages × 2 engines). Worth doing in a follow-up but not a blocker; both engines already keep their getHealth + getStats SQL fully parallel, this is the established pattern.
- Postgres E2E parity test. The SQL is duplicated verbatim between engines; PGLite tests cover the semantics. Postgres-specific test costs vs. confidence is a judgment call best made by maintainer.
- `findOrphanPages` rename. Codex pre-impl agreed keeping it as-is (with updated docstring) avoids churn while metric semantics are changing.

## Test plan

- [x] `bun test test/orphans.test.ts` — 60/60 pass
- [x] `bun test test/pglite-engine.test.ts test/brain-score-breakdown.test.ts test/doctor.test.ts` — 143/143 pass
- [x] `bun test test/core/cycle.serial.test.ts` — 28/28 pass
- [x] Live verification: `gbrain orphans --count` (2656) == `get_health.orphan_pages` (2656)
- [ ] CI on this PR

### DIFF
--- a/src/commands/orphans.ts
+++ b/src/commands/orphans.ts
@@ -50,6 +50,14 @@ const DENY_PREFIXES = [
   'scripts/',
   'templates/',
   'openclaw/config/',
+  // Source-type ingestions: no inbound wikilinks is the healthy default.
+  // These pages are anchored via timeline entries / parent records, not by
+  // being cited from concept pages. Without exclusion they dominate orphan
+  // counts and crush no_orphans_score for reasons unrelated to graph rot.
+  'emails/',
+  'attachments/',
+  '0-daily/',
+  '4-archive/',
 ];
 
 /** First slug segments where no inbound links is expected */
--- a/src/commands/orphans.ts
+++ b/src/commands/orphans.ts
@@ -58,6 +58,13 @@ const DENY_PREFIXES = [
   'attachments/',
   '0-daily/',
   '4-archive/',
+  // Navigation pages are vault-level hub-target pages (Obsidian graph-view
+  // clustering anchors — body literally reads "Notes linking here appear in
+  // the graph view as connected to this hub"). They exist to organize the
+  // graph, not to carry content; missing inbound = sparse [[wikilink]] usage
+  // in the vault, not graph rot. Excluding them keeps brain_score focused on
+  // real content rot.
+  'navigation/',
 ];
 
 /** First slug segments where no inbound links is expected */
--- a/src/commands/orphans.ts
+++ b/src/commands/orphans.ts
@@ -1,15 +1,21 @@
 /**
- * gbrain orphans — Surface pages with no inbound wikilinks.
+ * gbrain orphans — Surface ISLANDED pages (no inbound AND no outbound).
  *
- * Deterministic: zero LLM calls. Queries the links table for pages with
- * no entries where to_page_id = pages.id. By default filters out
- * auto-generated pages and pseudo-pages where no inbound links is expected.
+ * Deterministic: zero LLM calls. Routes through `engine.findIslandedPages()`
+ * which shares its SQL predicate with `getHealth.orphan_pages`, so the
+ * count this command reports equals the `brain_score` orphan_pages count
+ * by construction (v0.33+ step-4 alignment).
+ *
+ * The JS-side `shouldExclude` filter is retained as an exported helper
+ * for backwards-compat tests + agent consumers that may inspect a slug
+ * outside the SQL path, but `findOrphans` no longer runs it — the SQL
+ * predicate is the runtime authority.
  *
  * Usage:
- *   gbrain orphans                  # list orphans grouped by domain
+ *   gbrain orphans                  # list islanded pages grouped by domain
  *   gbrain orphans --json           # JSON output for agent consumption
  *   gbrain orphans --count          # just the number
- *   gbrain orphans --include-pseudo # include auto-generated/pseudo pages
+ *   gbrain orphans --include-pseudo # relax exclusions; islanded predicate unchanged
  */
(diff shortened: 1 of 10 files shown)


time-attack:
Thanks for the careful four-commit writeup — the diagnosis was right, and master now agrees with it. The alignment landed since this PR opened: `src/core/orphan-policy.ts` is the one authoritative exclusion policy and both `gbrain orphans` and `getHealth.orphan_pages` consume it over the same linkable scope with the islanded + `deleted_at IS NULL` predicate (#3023, then #3155 and #3214). Closing as already fixed on master rather than carrying a 467-line conflicting chain against both engines.

PR #1010, opened 2026-05-15
fix(entities): prefix-expansion resolver + stub-guard + dropped-fact audit

garrytan:
## What this PR does

Stops the bug class where bare-name references (\"Alice\", \"Jared\") silently spawned phantom unprefixed entity pages at the brain root. Three independent layers, all addressed.

### Layer 1 — resolver prefix expansion
\`src/core/entities/resolve.ts\`

- New `tryPrefixExpansion` step matches both `<dir>/<token>` exactly AND `<dir>/<token>-%`.
- Bare-name-first priority: \"Alice\" finds `people/alice-example` before exact-matching a leftover phantom.
- Config-driven dir list: `entities.prefix_expansion_dirs` overrides the defaults `['people', 'companies', 'deals', 'topics', 'concepts']`.
- Real-page preservation: bare slugs with non-stub bodies stay intact.
- Fuzzy title matches win over prefix expansion when the bare slug doesn't exist (Liz/Elizabeth case).

### Layer 2 — stub-creation guard
\`src/core/facts/fence-write.ts\`

- Refuses to stub-create phantom pages with no directory prefix.
- Also blocks appends to pre-v0.34.5 phantom files already on disk (consults the DB body too; legitimate DB-only pages get materialized + appended rather than dropped).

### Layer 3 — dropped-fact audit
\`src/core/facts/backstop.ts\` + \`src/core/facts/dropped-audit.ts\`

- When the stub-guard fires, the backstop logs the fact to \`~/.gbrain/facts.dropped.jsonl\` instead of inserting a legacy-shape DB row that would trip the v0.32.2 \`extract_facts\` reconciliation guard.
- Fact text preserved verbatim so a future \`gbrain replay-dropped\` tool can re-process the entries once the canonical entity pages exist.

## Tests
- 35 dedicated tests in \`test/entity-resolve.test.ts\`.
- Full unit suite (6485 tests) green.
- \`bun run verify\` clean.

## Review trail
- \`/plan-eng-review\` with 7 architectural decisions captured in \`~/.claude/plans/mossy-popping-crown.md\`.
- 30 rounds of \`codex review --base master\`. The merge-phantoms cleanup command was built and then stripped (see below).

## Cleanup of existing phantoms — deferred

[...]

PR #1109, opened 2026-05-17
v0.35.3.2 docs(designs): MERGE_PHANTOMS retrospective + future-implementation guide

garrytan:
## Summary

482-line design retrospective on the `gbrain merge-phantoms` operator command that was built and then stripped from closed PR #1010 after 30 rounds of codex review.

The fence-and-cleanup work itself is captured (and superseded) by PR #1085. This PR exists to preserve the institutional knowledge of what went wrong in the cleanup-command attempt so the next implementer doesn't repeat it.

## What's in the doc

- **Problem statement** — what a phantom is, why they exist, why the cleanup matters.
- **What we built** — the merge-phantoms command as it existed before strip.
- **30-round cascade table** — every codex finding (4 P1 data-loss paths + ~40 P2 correctness issues), each fix, each fix-of-fix. The pattern: defensive checks defending the wrong shape.
- **Meta-insight** — the fence is the v0.32.2 system of record. A destructive cleanup that mutates fence + DB independently has to re-implement reconciliation logic that already lives in `extract_facts`. That duplication IS the bug farm.
- **Three alternative-design options** with code-size / risk / operator-burden trade-offs:
  - **Alpha** (~80 lines) — report-only command, manual remediation. Zero data-loss risk.
  - **Beta** (~150 lines) — hook phantom redirect into the existing `extract_facts` cycle phase. **Recommended.** Reuses every drift-handling pathway that's already battle-tested.
  - **Gamma** (~300 lines + migration) — `canonical_of` column for first-class phantom support. Also unlocks general alias support.
- **Regression checklist** — every gotcha codex found, extracted as a test backlog (real fence markers, MAX_SEARCH_LIMIT clamping, stale compiled_truth, embedding driver shape variance, source resolution chain, bi-directional fence drift, etc).
- **Lessons learned** — for the next destructive cleanup command in this codebase. Resist bundling follow-ups into a fix PR. Do design work BEFORE implementation. Codex catches drift; only the human can name when an abstraction is wrong.

## Why this is its own PR

The closed PR #1010 commits are still in git history as the worked example of every gotcha. Treat them as a regression checklist; don't refactor them into the next attempt. PR #1085 supersedes the resolver+stub-guard+backstop fix layers with a structurally cleaner implementation. The merge-phantoms cleanup remains unfinished work — this doc is the bridge from "30 rounds of bug-farm iteration" to "next implementer ships Option Beta cleanly."

## Test plan

- [x] No code changed — design doc only.
- [x] `bun run verify` clean.
- [x] `bun run typecheck` clean.


time-attack:
Closing this one. The retrospective is thoughtful, but it documents a command that was deliberately stripped and whose actual fix shipped elsewhere, so it becomes permanent doc surface describing something the codebase does not have. The VERSION/package.json bump to 0.35.3.2 is also long stale against master. The institutional knowledge is preserved in the PR history and in #1010/#1085 — that's the right home for it.

PR #1112, opened 2026-05-17
Respect public URL for HTTP server links

RandyHaddad:
## Summary\n- Resolve the HTTP server public URL from --public-url or deployment env vars like VERCEL_URL/PUBLIC_URL\n- Use the resolved URL for OAuth issuer, magic links, and startup admin/MCP/health links\n\n## Testing\n- Not run (small URL resolution/server banner change)

### DIFF
--- a/src/commands/serve-http.ts
+++ b/src/commands/serve-http.ts
@@ -177,8 +177,31 @@ interface ServeHttpOptions {
   bind?: string;
 }
 
+function resolvePublicUrl(rawPublicUrl: string | undefined, port: number): string {
+  const candidates = [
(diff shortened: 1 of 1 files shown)


time-attack:
Thanks for this — the behavior you're after is already on master: `--public-url` drives the OAuth issuer, the magic-link base URL, the cookie secure flag, and the startup banner. The remaining delta here is implicit inference from `VERCEL_URL`/`PUBLIC_URL`, and we don't want ambient deploy-platform env vars silently rewriting the OAuth issuer — that's an explicit-flag decision on purpose. Closing as already covered.

PR #1113, opened 2026-05-17
fix(autopilot,cycle): stop respawn-storm from steady-state 'partial' cycles

sergeclaesen:
## Summary

Two interacting design-bugs cause gbrain autopilot to give up — and under
LaunchAgent/systemd `KeepAlive=true` to respawn-storm — every five normal
cycles on any non-trivial brain. Both are one-liner fixes plus source-level
regression guards.

### Bug 1 — autopilot circuit-breaker trips on `'partial'`

`src/commands/autopilot.ts:358` (inline-cycle path):

```ts
if (report.status === 'failed' || report.status === 'partial') {
  cycleOk = false;
}
```

But the `CycleReport` docstring is explicit on the meaning of these statuses:

> - 'partial' : at least one phase warned or failed, others ran
> - 'failed'  : lock acquired but all attempted phases failed

`partial` is a soft signal, not a fatal condition. Treating it as a failure
means a single `warn`-emitting phase trips the consecutive-error counter
every cycle, and after 5 cycles the autopilot gives up. The dispatch path
(`event: 'dispatched'` above) doesn't have this bug — it only looks at
explicit `throw`s from the dispatcher.

**Fix:** only `'failed'` sets `cycleOk = false`. The
`consecutiveErrors >= 5` give-up branch remains as the real-failure
backstop, so the autopilot still stops itself when the DB is unreachable
for 5 cycles in a row.

### Bug 2 — orphan threshold is absolute, not ratio

`src/core/cycle.ts:939`:

```ts
status: count > 20 ? 'warn' : 'ok',
```

The orphans phase is the most common producer of a per-phase warn. On
any brain past a few hundred pages a steady-state orphan count of
50–2000 is normal — entity pages, single-pass syntheses, raw sources
that legitimately have no inbound links yet. A fixed `20`-absolute
floor fires `warn` every cycle, the `cycle.status` rolls up to
`partial`, and combined with Bug 1 above the autopilot stops itself.

In the field repro: 2068 orphans on 5786 pages (≈36% — steady state)
produced a 9 MB respawn-error log inside 2 h under
`KeepAlive=true`.

**Fix:** ratio-based threshold. Warn only when more than half the
corpus is orphaned (the real "your graph fell apart" signal):

```ts
const status: PhaseStatus =
  result.total_pages > 0 && count / result.total_pages > 0.5 ? 'warn' : 'ok';
```

`total_pages == 0` is a defensive `'ok'` (no corpus, nothing to warn
about).

## Tests

`test/autopilot-cycle-failure-classification.test.ts` (new) — 5
source-level regression guards in the same style as `cycle-abort.test.ts`
(autopilot's main loop is hard to unit-test against a real engine; the
existing codebase uses source-grep tests for this layer):

- inline-cycle block must not OR-against `'partial'`
- `consecutiveErrors >= 5` breaker must still exist (we narrowed; we
  didn't remove)
- `runPhaseOrphans` must not use a literal `count > 20` ternary
- `runPhaseOrphans` must reference `total_pages` in its threshold
- both `'warn'` and `'ok'` branches still reachable

Local verification:

```
$ bun test test/autopilot-*.test.ts test/cycle-abort.test.ts test/orphans.test.ts
 47 pass / 0 fail

$ bun run verify
... all gates green incl. typecheck + admin-build ...
```

## Backwards compatibility

- Public API of `CycleReport` / `PhaseResult` / `CyclePhase` unchanged.
- Behaviour change is strictly in the loosening direction: cycles
  previously classified `'partial'` no longer trip the autopilot
  circuit breaker; orphan counts below 50% of total pages no longer
  emit `'warn'`. Brains that genuinely fall apart (`status === 'failed'`
  or orphan ratio > 50%) keep the same signal.
- No migration; no config knob added. The previous absolute threshold
  was undocumented.

## Why one PR, not two

The two fixes are deliberately coupled. Either bug in isolation is
survivable: Bug 1 alone never trips because no phase reaches `partial`
in practice without the orphan-warn; Bug 2 alone just makes
`cycle-inline` log `[partial]` without consequence. Together they
produce the respawn-storm. The new test file documents both arms of
the failure mode in one place.

### DIFF
--- a/src/commands/autopilot.ts
+++ b/src/commands/autopilot.ts
@@ -355,7 +355,13 @@ export async function runAutopilot(engine: BrainEngine, args: string[]) {
             await new Promise(r => setImmediate(r));
           },
         });
-        if (report.status === 'failed' || report.status === 'partial') {
+        // Only 'failed' (every attempted phase failed) trips the autopilot
+        // circuit breaker. 'partial' means at least one phase warned or
+        // failed while others ran — that's a soft signal, not a fatal
+        // condition. Treating 'partial' as failure here caused respawn
+        // storms under KeepAlive=true on brains where a single phase
+        // (typically `orphans`) emits a 'warn' every cycle in steady state.
+        if (report.status === 'failed') {
           cycleOk = false;
         }
         if (jsonMode) {
--- a/src/core/cycle.ts
+++ b/src/core/cycle.ts
@@ -934,9 +934,19 @@ async function runPhaseOrphans(engine: BrainEngine): Promise<PhaseResult> {
     const { findOrphans } = await import('../commands/orphans.ts');
     const result = await findOrphans(engine);
     const count = result.total_orphans;
+    // Orphans are a code-smell signal, not a fatal condition. The
+    // original `count > 20` cutoff was tuned for small dev brains; on
+    // any corpus past a few hundred pages it fires 'warn' every cycle
+    // in steady state. Combined with the autopilot circuit-breaker
+    // historically tripping on cycle.status='partial', that produced
+    // respawn storms under KeepAlive=true. Switch to a ratio: warn
+    // only when more than half the corpus is orphaned (the real "your
+    // graph fell apart" signal). total_pages=0 is a defensive 'ok'.
+    const status: PhaseStatus =
+      result.total_pages > 0 && count / result.total_pages > 0.5 ? 'warn' : 'ok';
     return {
       phase: 'orphans',
-      status: count > 20 ? 'warn' : 'ok',
+      status,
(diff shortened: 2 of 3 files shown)


garrytan:
Thanks for this, @sergeclaesen — closing as superseded. Your work shipped in **v0.36.1.1**. From the [CHANGELOG](https://github.com/garrytan/gbrain/blob/master/CHANGELOG.md):\n\n> - Autopilot stop respawn-storm from steady-state  cycles (#1113). Trips only on  now. Cherry-pick of @sergeclaesen's PR.\n\nReal appreciation for the contribution.

Issue #396, opened 2026-04-24
PGLite initSchema fails on mid-migration installs — source_id index created before column exists


jkrums:
## Summary

`PGLITE_SCHEMA_SQL` in `src/core/pglite-schema.ts` runs `CREATE INDEX IF NOT EXISTS idx_pages_source_id ON pages(source_id)` (line 63) every time `initSchema()` is called. The migration loop that adds `pages.source_id` (migration v21) runs **after** this. On a mid-migration install (any DB that predates v21), the sequence is:

1. `initSchema()` → `db.exec(PGLITE_SCHEMA_SQL)`
2. `CREATE TABLE IF NOT EXISTS pages (...)` → no-op (pages already exists from schema v4)
3. `CREATE INDEX IF NOT EXISTS idx_pages_source_id ON pages(source_id)` → **fails with `column \"source_id\" does not exist`**
4. Migrations never run.

The implicit assumption is that the schema SQL is only executed against either a fresh DB (where `CREATE TABLE` just created `pages` with `source_id`) or an already-current DB (where the column was added by a prior migration run). Any install caught between those two states is wedged.

## Impact

A PGLite install at schema v4 was stuck wedged for 3 days across `gbrain apply-migrations` runs (v0.11.0, v0.12.0, v0.12.2 all recorded as `partial` in `~/.gbrain/migrations/completed.jsonl`). Each failure in `upgrade-errors.jsonl` is a terse `column \"source_id\" does not exist` with no hint that the schema SQL itself is the culprit — all the data-migration orchestrators point at their own schema phase, which just shells out to `gbrain init --migrate-only` and gets the same error.

In our case, the nightly `gbrain sync` launchd job ran successfully enough to mask the wedge (sync skips non-fatally on individual-file errors), so it wasn't obvious until we tried to upgrade the gbrain version.

## Environment

- gbrain `0.20.4` (commit `11abb24`)
- Wedged from schema v4; the failing path affects any install that hasn't already applied migration v21 before picking up a version of the schema SQL that includes the `idx_pages_source_id` line.
- PGLite engine (`~/.gbrain/brain.pglite`)
- Node 24.14.1 / macOS 24.6.0

## Reproduction

Starting state: PGLite DB at schema v4 (i.e. any install from before v21 was introduced).

```
$ npx gbrain init --migrate-only
column \"source_id\" does not exist
```

[...]

Issue #395, opened 2026-04-24
v24 rls_backfill_missing_tables fails on PGLite — missing engine.kind gate


jkrums:
## Summary

Migration `v24` (`rls_backfill_missing_tables`, introduced in v0.18.1) fails on PGLite installs because it unconditionally runs `ALTER TABLE ... ENABLE ROW LEVEL SECURITY` on tables that only exist in the Postgres schema (`subagent_messages`, `access_tokens`, `mcp_request_log`, `subagent_tool_executions`, `subagent_rate_leases`, etc.). PGLite's `pglite-schema.ts` deliberately omits these tables (comment at top: "No access_tokens / mcp_request_log (local-only, no remote auth)").

Other migrations that are Postgres-only correctly guard with `if (engine.kind === 'pglite') return;` (see v23 `files_source_id_page_id_ledger`). v24 is missing that gate.

## Environment

- gbrain `0.20.4` (commit `11abb24`)
- PGLite engine (`~/.gbrain/brain.pglite`)
- Node 24.14.1 / macOS 24.6.0

## Reproduction

Any PGLite install reaching schema v23 will hit this on the next `initSchema()` / `gbrain init --migrate-only`:

```
  [24] rls_backfill_missing_tables...
error: relation "subagent_messages" does not exist
  severity: ERROR
  code: 42P01
  where: 'SQL statement "ALTER TABLE subagent_messages ENABLE ROW LEVEL SECURITY"
          PL/pgSQL function inline_code_block line 23 at SQL statement'
```

## Suggested fix

Add an engine gate at the top of the v24 handler, matching the pattern already used in v23:

```ts
// src/core/migrate.ts around line 748 (v24 entry)
{
  version: 24,
  name: 'rls_backfill_missing_tables',
  sql: '',
  handler: async (engine) => {
    if (engine.kind === 'pglite') return;
    // ... existing RLS backfill SQL
  },
},
```

PGLite is single-process / single-user with no PostgREST exposure, so there's nothing to backfill — a no-op is the correct behavior (already documented in the RLS doctor check: `[OK] rls: Skipped (PGLite — no PostgREST exposure, RLS not applicable)`).

## Workaround

For anyone else hitting this before the fix ships, mark v24 done manually:

```js
// requires @electric-sql/pglite loaded with extensions: { vector, pg_trgm }
await db.query(
[...]

Issue #378, opened 2026-04-24
Upgrade from v0.17.x → v0.18.x on Postgres fails: SCHEMA_SQL runs before migrations, indexes reference not-yet-added columns


valzav:
## Summary

On v0.18.2, upgrading a pre-v0.18 Postgres brain (schema ≤ v19) fails with `column "source_id" does not exist` during `gbrain init --migrate-only`. This blocks the v0.16.0 migration orchestrator (its Phase A shells out to `init --migrate-only`) and by extension any `gbrain apply-migrations --yes` run against an older brain.

Reproduced on a schema v15 Postgres brain upgrading to gbrain 0.18.2.

## Root cause

Ordering bug in `PostgresEngine.initSchema()` (`src/core/postgres-engine.ts:73`):

```ts
async initSchema(): Promise<void> {
  await conn`SELECT pg_advisory_lock(42)`;
  try {
    await conn.unsafe(SCHEMA_SQL);              // ← runs first
    const { applied } = await runMigrations(this);  // ← migrations run after
    ...
  }
}
```

`SCHEMA_SQL` contains (schema.sql:74):

```sql
CREATE INDEX IF NOT EXISTS idx_pages_source_id ON pages(source_id);
```

- On a **fresh** DB, the preceding `CREATE TABLE IF NOT EXISTS pages (...)` defines `source_id` as part of the table (schema.sql:52-66), so the index creation succeeds.
- On an **existing** v15 DB, `CREATE TABLE IF NOT EXISTS pages` is a no-op (NOTICE: "relation pages already exists, skipping"). The `source_id` column is added by migration **v21** (`pages_source_id_composite_unique`) via `ALTER TABLE ... ADD COLUMN IF NOT EXISTS source_id`. That migration only runs **after** `SCHEMA_SQL` finishes — but `SCHEMA_SQL` dies first on the index creation.

Same pattern applies to `files.source_id` (schema.sql:282, column added in v23) and `files.page_id` (schema.sql:`idx_files_page_id`, column added in v23).

## Reproduction

Any Postgres brain with `schema_version < 20` upgrading to 0.18.x:

```
$ gbrain apply-migrations --yes
...
column "source_id" does not exist
Migration v0.16.0 reported status=failed.
```

Full error chain:

- `apply-migrations --yes` invokes migration orchestrator v0.16.0
- v0.16.0 Phase A runs `gbrain init --migrate-only`
- `init --migrate-only` calls `engine.initSchema()`
[...]

Issue #375, opened 2026-04-24
v0.18.2: cold-start upgrade from schema v10 fails with "column source_id does not exist"


gopalpatel:
**What happened?**

Upgrading a PGLite brain from schema v10 (last healthy at v0.12.3) to v0.18.2
fails on the v0.13.0 migration orchestrator's Phase A:

```
=== Applying migration v0.13.0: Frontmatter becomes a graph ===
=== v0.13.0 — Frontmatter relationship indexing ===
column "source_id" does not exist
Migration v0.13.0 reported status=failed.
```

This is a new forward-reference bug that didn't exist in v0.14.x. It's
distinct from #238/#266 (which are also unresolved and block the same path —
see cross-references below).

**What did you expect?**

`gbrain apply-migrations --yes` to walk schema v10 → v24 cleanly, applying
each migration in order. The v0.18.2 changelog claims migration hardening
including atomic commit and integrity-window fixes — this gave high
confidence the orchestrator handled crash-resilience. However, for brains at
schema v10 (pre-v0.13), no migration release between v0.13.0 and v0.18.2
ships a working forward path: each release has a different forward-reference
or execPath failure inside `gbrain init --migrate-only`.

**Steps to reproduce**

Starting from a healthy v0.12.3 brain on PGLite:

1. `gbrain doctor` → `schema_version: Version 10 (latest: 10)`, health_score 80/100
2. Take a cold snapshot (`cp -r ~/.gbrain/brain.pglite ~/.gbrain/brain.pglite.bak`)
3. Stop any `gbrain sync --watch` service
4. In the source clone: `git fetch && git checkout 08b3698e90` (v0.18.2), `bun install`
5. `gbrain --version` → `gbrain 0.18.2`, `gbrain doctor` → `Version 10, latest is 24. Fix: gbrain apply-migrations --yes` (expected)
6. `gbrain apply-migrations --yes` — fails within 1 second with the error above
7. `gbrain doctor` confirms: brain still at schema v10, MINIONS HALF-INSTALLED flag set, all page data intact (2066 pages, embeddings 100%, graph 100%)

The orchestrator correctly rolls back atomically — no partial schema state
reaches disk. That part of v0.18.2 is working as designed.

**Environment**

- gbrain version: **0.18.2** (target), rolled back to **0.12.3** after failure
- OS: Ubuntu 25.10 on aarch64 (NVIDIA DGX Spark, Grace Blackwell)
- Bun: 1.3.12 (700fc117)
[...]

PR #1115, opened 2026-05-17
fix(engine): bootstrap oauth_clients forward-reference columns before SCHEMA_SQL replay

bautrey:
## Problem

`applyForwardReferenceBootstrap()` in both `src/core/postgres-engine.ts` and `src/core/pglite-engine.ts` is the contract that lets `initSchema()` safely replay `(PGLITE_)SCHEMA_SQL` on brains older than the latest schema version. It pre-creates every table/column the latest blob references so the blob's `CREATE INDEX` and FK statements don't crash before the numbered migration chain has a chance to install the real state.

Two columns were missing from that coverage at v0.34.4.0:

- `oauth_clients.source_id` — forward-referenced by `idx_oauth_clients_source_id ON oauth_clients(source_id) WHERE source_id IS NOT NULL` (added v60).
- `oauth_clients.federated_read` — forward-referenced by `idx_oauth_clients_federated_read ON oauth_clients USING GIN (federated_read)` (added v61).

`oauth_clients` itself dates back to v0.30 / v45. Any brain at `schema_version < 60` with the table already installed (so: any v0.30 / v0.31 / v0.32 / v0.33 brain that has the OAuth schema) wedges on the SCHEMA_SQL replay: the `CREATE INDEX` statements reference columns that don't exist yet, the engine throws, and the operator cannot run `gbrain apply-migrations` to install v60-v65 because every command path through `connectEngine()` runs `initSchema()` first.

## Reproduction

Caught upgrading our FortiumClaw deployment from v0.31.3 (schema 46) → v0.34.4.0 (schema 66). On the failing run, `initSchema()` aborted on the `CREATE INDEX idx_oauth_clients_source_id` statement with `column "source_id" does not exist`. The CLI's `--help`, `doctor`, `apply-migrations` — all blocked by the same code path. Manually `ALTER TABLE oauth_clients ADD COLUMN IF NOT EXISTS source_id TEXT; ALTER TABLE oauth_clients ADD COLUMN IF NOT EXISTS federated_read TEXT[] NOT NULL DEFAULT '{}';` unblocked the upgrade and let the v60-v65 migrations run cleanly.

Fresh installs and brains already at v60+ both no-op through the new branches.

## Fix

In both engines:

- Add `oauth_clients_*_exists` probes to the existence-check SELECT (mirrors the existing pattern for `ingest_log`, `mcp_log`, `subagent_messages.provider_id`, etc.).
- Extend the early-return short-circuit so brains that don't need the new bootstrap continue to no-op.
- Add an `ALTER TABLE oauth_clients ADD COLUMN IF NOT EXISTS ...` block at the bottom of each bootstrap function.

The added bootstrap SQL is **intentionally lean** — matches v60's leading `ALTER TABLE ... ADD COLUMN IF NOT EXISTS source_id TEXT` and v61's column add **verbatim**. No FK install, no backfill, no validation; those run later via the numbered migrations (v60-v65) which are already idempotent and handle:

- v60: source_id FK install (NOT VALID) + backfill
- v61: federated_read column add (this PR matches exactly)
- v62: federated_read backfill from existing scopes
- v63: federated_read NOT NULL validation
- v64: source_id FK RESTRICT flip
- v65: source_id FK validation

## Test coverage

`test/schema-bootstrap-coverage.test.ts` is the CI guard that asserts every forward reference declared in `REQUIRED_BOOTSTRAP_COVERAGE` is actually covered by the bootstrap. This PR:

- Adds the two new column entries to `REQUIRED_BOOTSTRAP_COVERAGE` (with comments explaining v60/v61 source).
- Extends the simulator block (pre-bootstrap "old brain" prep) to drop the v60/v61 indexes + FK + columns so the test exercises the new branch.
- Extends the post-bootstrap replay block to do the same.

All 5 tests pass locally (`bun test test/schema-bootstrap-coverage.test.ts`):
```
 5 pass
 0 fail
 37 expect() calls
```

`bun run typecheck` (`tsc --noEmit`) clean.

## Maintenance contract

Both engines kept in sync per the "Keep this in sync with the PGLite version" comment at the top of `applyForwardReferenceBootstrap`. The doc-comment in pglite-engine.ts is also extended to list the new forward references (matches the existing format for prior additions).

## Wedge class

This closes the v60+v61 instance of the recurring wedge class tracked in #239, #243, #266, #357, #366, #374, #375, #378, #395, #396 — same shape, same fix pattern. The structural CI test was added to prevent this class; this PR uses the contract that test enforces.

### DIFF
--- a/src/core/pglite-engine.ts
+++ b/src/core/pglite-engine.ts
@@ -243,6 +243,10 @@ export class PGLiteEngine implements BrainEngine {
    *     (indexed by `idx_mcp_log_agent_time`) — v0.26.3
    *   - `subagent_messages.provider_id` column (indexed by
    *     `idx_subagent_messages_provider`) — v0.27
+   *   - `oauth_clients.source_id` column (indexed by
+   *     `idx_oauth_clients_source_id`) — v0.34.1 (v60)
+   *   - `oauth_clients.federated_read` column (indexed by
+   *     `idx_oauth_clients_federated_read`) — v0.34.1 (v61)
    *
    * **Maintenance contract:** when a future migration adds a column-with-index
    * or new-table-with-FK referenced by PGLITE_SCHEMA_SQL, extend this method
@@ -288,7 +292,13 @@ export class PGLiteEngine implements BrainEngine {
         EXISTS (SELECT 1 FROM information_schema.tables
                 WHERE table_schema='public' AND table_name='ingest_log') AS ingest_log_exists,
         EXISTS (SELECT 1 FROM information_schema.columns
-                WHERE table_schema='public' AND table_name='ingest_log' AND column_name='source_id') AS ingest_log_source_id_exists
+                WHERE table_schema='public' AND table_name='ingest_log' AND column_name='source_id') AS ingest_log_source_id_exists,
+        EXISTS (SELECT 1 FROM information_schema.tables
+                WHERE table_schema='public' AND table_name='oauth_clients') AS oauth_clients_exists,
+        EXISTS (SELECT 1 FROM information_schema.columns
+                WHERE table_schema='public' AND table_name='oauth_clients' AND column_name='source_id') AS oauth_clients_source_id_exists,
+        EXISTS (SELECT 1 FROM information_schema.columns
+                WHERE table_schema='public' AND table_name='oauth_clients' AND column_name='federated_read') AS oauth_clients_federated_read_exists
     `);
     const probe = rows[0] as {
       pages_exists: boolean;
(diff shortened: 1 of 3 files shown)


garrytan:
Thanks @bautrey — already shipped in master as of **v0.35.5.0**. `applyForwardReferenceBootstrap` in both `src/core/pglite-engine.ts` and `src/core/postgres-engine.ts` now probes `files.source_id`, `files.page_id`, `oauth_clients.source_id`, `oauth_clients.federated_read`, `sources.archived`, `sources.archived_at`, and `sources.archive_expires_at` — before SCHEMA_SQL replays. Closes the pre-v18/v34/v60 forward-reference wedges. `test/bootstrap.test.ts` + `test/schema-bootstrap-coverage.test.ts` enforce.

If your install is still wedged on `gbrain upgrade`, please reopen with the output of `gbrain doctor --json` and your starting schema_version.

Closing as already-shipped. Real appreciation for chasing this — the structural fix-set with the schema-bootstrap-coverage CI guard now prevents the entire bug class.

PR #876, opened 2026-05-11
feat(oauth): federated_read column + AuthInfo.allowedSources

sonlndv:
## Summary

Replaces the `oauth_clients.source_id NULL = super-reader` pattern with explicit, per-client read scope. Each row now carries both a single **write target** (`source_id NOT NULL`) and a **read scope** (`federated_read TEXT[]`), giving fine-grained federation without the silent super-reader trap.

**Stacked on #861** (read-side source isolation v2). Targets the same base branch.

## Why

The single-column `source_id` shape conflates two concerns:

- `source_id = 'X'` → writes go to X **and** reads are scoped to X
- `source_id IS NULL` → super-reader (reads everything)

Real federation needs the middle ground: write to ONE source, read from N. Today the only way to give a client that shape is `source_id IS NULL`, which also grants unconstrained write privilege and silently survives source deletes (FK was `ON DELETE SET NULL`).

This PR splits write target and read scope into two columns, lets dept-style clients federate cleanly across `own + shared + wecare`, and closes the silent-super-reader-on-source-delete leak.

## Schema changes

Migrations v47–v52, applied across `src/core/migrate.ts`, `src/core/schema-embedded.ts` (fresh postgres installs), and `src/core/pglite-schema.ts` (fresh pglite installs).

| Version | Change |
|---|---|
| v47 | `ALTER TABLE oauth_clients ADD COLUMN federated_read text[]` (nullable) |
| v48 | Backfill `federated_read = ARRAY[source_id]` for non-NULL `source_id` rows; **`RAISE EXCEPTION` if any NULL `source_id` row exists** (refuse to repair silently — operator must fix via audit skill before retry) |
| v49 | `ALTER COLUMN source_id SET NOT NULL` |
| v50 | `ALTER COLUMN federated_read SET NOT NULL` + `CHECK (source_id = ANY(federated_read))` |
| v51 | Drop FK + recreate with `ON DELETE RESTRICT` (was `SET NULL` — closed silent super-reader promotion) |
| v52 | `CREATE INDEX ON oauth_clients USING GIN (federated_read)` (read-scope lookups) |

`AuthInfo` gains `allowedSources?: string[]` populated from `federated_read`. Engines prefer `allowedSources` (ANY()) over scalar `sourceId` (=) when both are present. Write paths remain scoped to scalar `sourceId`.

## Code path wiring

```
oauth_clients.federated_read
    └─> AuthInfo.allowedSources               (src/core/oauth-provider.ts:verifyAccessToken)
[...]

PR #1116, opened 2026-05-17
fix(schema): bootstrap oauth_clients.source_id + federated_read on upgrade

hesong12:
## Symptom

Upgrading a v0.31.0 brain to v0.34.x or later fails on `initSchema` with:

```
Schema probe/migrate failed: column "source_id" does not exist
  Try: gbrain init --migrate-only
```

`gbrain init --migrate-only` reports the same error. `runMigrations` never gets a chance to apply v60+ because the schema replay aborts first.

## Root cause

`v0.34.1` (#861, #876) added `oauth_clients.source_id` (migration v60) and `oauth_clients.federated_read` (v61), and inlined both columns into `CREATE TABLE IF NOT EXISTS oauth_clients` across `src/schema.sql`, `src/core/schema-embedded.ts`, and `src/core/pglite-schema.ts`. The same schema files emit two standalone `CREATE INDEX` statements that reference the new columns:

```sql
CREATE INDEX IF NOT EXISTS idx_oauth_clients_source_id
  ON oauth_clients(source_id) WHERE source_id IS NOT NULL;
CREATE INDEX IF NOT EXISTS idx_oauth_clients_federated_read
  ON oauth_clients USING GIN (federated_read);
```

On a fresh install everything lands. On an upgrade brain, `CREATE TABLE IF NOT EXISTS` is a no-op so the inline column adds are silently skipped — and the standalone `CREATE INDEX` crashes referencing the missing column. `initSchema` aborts before `runMigrations` runs.

This is the same upgrade-path class already handled for `pages.source_id` (v18), `links.link_source` (v13), `content_chunks.embedding_image` (v39), `pages.effective_date` (v41), and `ingest_log.source_id` (v50) — all bootstrapped via `applyForwardReferenceBootstrap` on both engines. `oauth_clients` was just missed.

## Fix

Extend `applyForwardReferenceBootstrap` in both `PostgresEngine` and `PGLiteEngine` to probe for `oauth_clients.source_id` and `oauth_clients.federated_read` and `ALTER TABLE ADD COLUMN IF NOT EXISTS` them before the schema replay. Migrations v60-v62 then run idempotently through `runMigrations` and pick up the columns this bootstrap provided (they backfill, install the FK, and add the GIN index).

The coverage contract in `test/schema-bootstrap-coverage.test.ts` gains the two new `REQUIRED_BOOTSTRAP_COVERAGE` entries so the static A2 check (covered-by-CREATE-TABLE-OR-bootstrap) fails loudly if anyone removes either `ALTER` from `applyForwardReferenceBootstrap`. The existing PGLite test passed at HEAD because `PGLITE_SCHEMA_SQL` declares both columns inline in `CREATE TABLE` (which the parser accepts as coverage) - that satisfies the fresh-install scenario, but the bootstrap is the actual upgrade-path invariant.

## Verification

Reproduced and fixed on a real v0.31.0 Supabase brain.

Before:
```
$ bun ./src/cli.ts init --migrate-only
... initial schema replay ...
column "source_id" does not exist
```

After (with this patch):
```
$ bun ./src/cli.ts init --migrate-only
Schema version 45 -> 66 (21 migration(s) pending)
[46] mcp_request_log_params_jsonb_normalize... ✓
[47] facts_notability_alter... ✓
...
[66] embed_stale_partial_index... ✓
21 migration(s) applied
```

Tests:

```
bun test test/schema-bootstrap-coverage.test.ts                  # 5 pass
bun test test/mcp-tool-defs.test.ts test/http-transport.test.ts          test/facts-meta-cache.test.ts          test/facts-mcp-allowlist.serial.test.ts                 # 44 pass
```

## Scope

3 files, +80/-3. No public API change. No new migration. Bootstrap-only.

## Notes

Detected during a real v0.31.0 → v0.35.4.0 upgrade. The pre-existing `applyForwardReferenceBootstrap` already follows the exact pattern this PR extends. The maintenance contract in the comment above the function explicitly says:

> when a future migration adds a column-with-index or new-table-with-FK referenced by PGLITE_SCHEMA_SQL, extend this method AND `test/schema-bootstrap-coverage.test.ts`'s `REQUIRED_BOOTSTRAP_COVERAGE`.

`v0.34.1` shipped without that extension. This PR closes that gap.

### DIFF
--- a/src/core/pglite-engine.ts
+++ b/src/core/pglite-engine.ts
@@ -288,7 +288,13 @@ export class PGLiteEngine implements BrainEngine {
         EXISTS (SELECT 1 FROM information_schema.tables
                 WHERE table_schema='public' AND table_name='ingest_log') AS ingest_log_exists,
         EXISTS (SELECT 1 FROM information_schema.columns
-                WHERE table_schema='public' AND table_name='ingest_log' AND column_name='source_id') AS ingest_log_source_id_exists
+                WHERE table_schema='public' AND table_name='ingest_log' AND column_name='source_id') AS ingest_log_source_id_exists,
+        EXISTS (SELECT 1 FROM information_schema.tables
+                WHERE table_schema='public' AND table_name='oauth_clients') AS oauth_clients_exists,
+        EXISTS (SELECT 1 FROM information_schema.columns
+                WHERE table_schema='public' AND table_name='oauth_clients' AND column_name='source_id') AS oauth_clients_source_id_exists,
+        EXISTS (SELECT 1 FROM information_schema.columns
+                WHERE table_schema='public' AND table_name='oauth_clients' AND column_name='federated_read') AS oauth_clients_federated_read_exists
     `);
     const probe = rows[0] as {
       pages_exists: boolean;
@@ -309,6 +315,9 @@ export class PGLiteEngine implements BrainEngine {
       subagent_provider_id_exists: boolean;
       ingest_log_exists: boolean;
       ingest_log_source_id_exists: boolean;
+      oauth_clients_exists: boolean;
+      oauth_clients_source_id_exists: boolean;
+      oauth_clients_federated_read_exists: boolean;
     };
 
     const needsPagesBootstrap = probe.pages_exists && !probe.source_id_exists;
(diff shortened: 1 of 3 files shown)


garrytan:
Thanks @hesong12 — already shipped in master as of **v0.35.5.0**. `applyForwardReferenceBootstrap` in both `src/core/pglite-engine.ts` and `src/core/postgres-engine.ts` now probes `files.source_id`, `files.page_id`, `oauth_clients.source_id`, `oauth_clients.federated_read`, `sources.archived`, `sources.archived_at`, and `sources.archive_expires_at` — before SCHEMA_SQL replays. Closes the pre-v18/v34/v60 forward-reference wedges. `test/bootstrap.test.ts` + `test/schema-bootstrap-coverage.test.ts` enforce.

If your install is still wedged on `gbrain upgrade`, please reopen with the output of `gbrain doctor --json` and your starting schema_version.

Closing as already-shipped. Real appreciation for chasing this — the structural fix-set with the schema-bootstrap-coverage CI guard now prevents the entire bug class.

PR #1124, opened 2026-05-17
fix(query): honor source id with no-expand

hnshah:
## Summary

Multi-source dogfooding exposed two related CLI/query issues:

1. `gbrain query ... --no-expand --source-id <id>` did not actually set `expand=false`.
2. Because `--no-expand` was not recognized as the negated form of the boolean `expand` param, the CLI parser consumed `--source-id` as the value for an unknown `no_expand` flag. That meant `source_id` never reached the query op.

There was also an op-layer ordering bug: explicit `source_id` was computed, but then `sourceScopeOpts(ctx)` was spread later and could overwrite the explicit per-call source.

This patch:

- supports `--no-<boolean>` in the shared CLI arg parser
- keeps `--no-expand` from consuming the next flag
- makes explicit `source_id` win over `ctx.sourceId`
- preserves `source_id=__all__` as the cross-source escape hatch
- applies the same explicit source scope to text and image query paths

## Dogfood evidence

Before, from inside a notes source:

```
gbrain query "freshEmbedSourceScope code source" --no-expand --source-id gstack-code-repo-0e4763c9
```

returned notes results because `--source-id` was swallowed.

After:

```
[0.8660] hello-js -- [JavaScript] hello.js:1-9 export statement summarizeDogfoodFinding
[0.5216] package-json -- [JSON] package.json:1-1 module
```

and the same query scoped to `ren-notes` returns the notes pages.

## Tests

```
bun test test/cli-args.test.ts test/mcp-eval-capture.test.ts test/search-lang-symbol-kind.test.ts --timeout 120000
```

22 passing.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -450,7 +450,7 @@ export function resolveQueryImage(
   return { path: imagePath, base64, mime };
 }
 
-function parseOpArgs(op: Operation, args: string[]): Record<string, unknown> {
+export function parseOpArgs(op: Operation, args: string[]): Record<string, unknown> {
   const params: Record<string, unknown> = {};
   const positional = op.cliHints?.positional || [];
   let posIdx = 0;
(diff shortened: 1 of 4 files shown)


garrytan:
Closing in favor of #1253 (v0.37.7.0 fix wave). The query op already routes via `ctx.sourceId` since v0.34.1.0; the wave adds matching `--source-id` flag parsing to `gbrain import` (which was the originating user-pain shape). Your dogfood patch on the query path is captured via `Co-Authored-By: hnshah` trailer in the relevant commits. Thank you.

PR #1125, opened 2026-05-17
fix(query): drain cache writes before cli exit

hnshah:
## Summary

Local-Ollama dogfooding showed that `gbrain query` could print results and then keep the CLI process alive until externally killed. The same path exited cleanly when cache writes were disabled.

Root cause: `hybridSearchCached` starts query-cache writeback as fire-and-forget work. That is fine for long-running MCP/server processes, but for a one-shot CLI invocation it can keep the process alive while the local engine is already moving toward disconnect.

This keeps cache writeback non-blocking for the search hot path, but tracks pending cache writes and lets the CLI query path drain them before disconnecting.

## Dogfood evidence

Before:

```
timeout 12s gbrain query "freshEmbedSourceScope code source" ...
# prints results, exits via timeout: 124
```

After:

```
EXIT:0
[0.8660] hello-js -- [JavaScript] hello.js:1-9 export statement summarizeDogfoodFinding
[0.5216] package-json -- [JSON] package.json:1-1 module
```

## Tests

```
bun test test/query-cache-knobs-hash.test.ts --timeout 120000
```

9 passing.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -169,6 +169,10 @@ async function main() {
     const result = JSON.parse(JSON.stringify(rawResult));
     const output = formatResult(op.name, result);
     if (output) process.stdout.write(output);
+    if (op.name === 'query') {
+      const { awaitPendingSearchCacheWrites } = await import('./core/search/hybrid.ts');
+      await awaitPendingSearchCacheWrites();
+    }
   } catch (e: unknown) {
     if (e instanceof OperationError) {
       console.error(`Error [${e.code}]: ${e.message}`);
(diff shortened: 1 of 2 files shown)


garrytan:
Closing as superseded — this fix shipped in v0.36.1.1 via #1182 ("fix-wave: community PR triage + 28 atomic fixes"), commit `4173f9e0` ("Drain cache writes before CLI exit (#1125)"). The change is on master. Thanks for the local-Ollama dogfood fix.

Dropping from in-progress fix-wave: warm-narwhal per drop-and-continue atomicity rule (2 drops of 9, below the abort threshold of 4).

PR #1134, opened 2026-05-18
Allow explicit expansion_model to reuse chat-capable providers

warkcod:
## Summary
This keeps dedicated `touchpoints.expansion` semantics intact while allowing an explicitly configured `expansion_model=provider:model` to reuse a chat-capable provider when the recipe does not declare a dedicated expansion touchpoint.

## Why
The openai-compatible runtime path for expansion already instantiates a normal `languageModel(modelId)`, but the recipe-layer capability gate required a dedicated expansion touchpoint. That meant advanced users pointing `expansion_model` at a chat-capable compat gateway got silently downgraded to no expansion even though the transport could serve the request.

## What changed
- allow `assertTouchpoint(..., 'expansion', ...)` to fall back to `chat` only for explicitly configured models
- make `isAvailable('expansion')` recognize the same explicit-config fallback path
- surface fallback support as `EXPAND=via-chat` in `gbrain providers list` / `explain`
- document the new label in provider help text
- add regression coverage for gateway availability, touchpoint validation, and CLI presentation

## Scope boundary
This does **not** mark chat-capable providers as first-class expansion providers by default. Dedicated expansion touchpoints remain the explicit recommended support matrix. The fallback only activates when the user explicitly configures `expansion_model` to that provider:model.

## Validation
- `/Users/mac/.bun/bin/bun test test/ai/gateway.test.ts test/ai/providers-list.test.ts test/ai/gateway-chat.test.ts test/ai/recipes-existing-regression.test.ts`

## Not tested
- full `bun run verify`
- end-to-end query flows against a real external compat gateway


warkcod:
Superseded by a clean rebased branch. Re-closing to avoid carrying the earlier merge/force-push timeline into review.

PR #1137, opened 2026-05-18
docs(shell-jobs): document database_url config requirement for gbrain CLI commands

garrytan-agents:
Shell jobs strip env vars by design (security), so `GBRAIN_DATABASE_URL` doesn't reach child processes. This causes `gbrain` CLI commands submitted as shell jobs to fail with "No database URL" — a common gotcha for agents using Minions for long-running operations.

Documents two workarounds:
1. Write `database_url` to `~/.gbrain/config.json` (recommended)
2. Pass `env` explicitly per job

Also documents the `HOME=/data` gotcha in container deployments where the config file is at `/data/.gbrain/config.json` not `/root/.gbrain/config.json`.

### DIFF
--- a/docs/guides/minions-shell-jobs.md
+++ b/docs/guides/minions-shell-jobs.md
@@ -139,6 +139,62 @@ gbrain jobs list --status waiting --name shell
 
 ---
 
+## Running gbrain CLI commands from shell jobs
+
+A common pattern is submitting shell jobs that call `gbrain` itself:
+
+```bash
+gbrain jobs submit shell \
+  --params '{"cmd":"gbrain sync --skip-failed && gbrain embed --stale","cwd":"/data/gbrain"}'
+```
+
+**This will fail if your database connection is configured only via environment
+variables** (`GBRAIN_DATABASE_URL` or `DATABASE_URL`). Shell jobs receive a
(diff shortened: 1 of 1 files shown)


garrytan:
Thanks @wintermute — your doc made the gap visible. Superseded by v0.35.8.0 which lands code instead of docs:

- Shell jobs grow `inherit: ["database_url"]`, validated **pre-enqueue** in both the CLI (`gbrain jobs submit`) and the `submit_job` op handler so no rejected payload ever lands in `minion_jobs.data`.
- Plain `env: { GBRAIN_DATABASE_URL: ... }` / `env: { DATABASE_URL: ... }` is rejected at validation with a paste-ready hint at the new pattern, so the DB-row leak path doesn't depend on caller hygiene.
- `gbrain doctor` warns when `~/.gbrain/` lives inside a git worktree.
- `~/.gbrain/.gitignore` is now laid down retroactively (post-upgrade + every `saveConfig` call) so existing users don't need to re-init.

/cso audited an earlier, bigger proposal (encrypted vault + Unix-socket broker) and flagged it as security theater for a single-uid topology, so we shipped the small fix that actually defeats the real threats. Credit in v0.35.8.0 CHANGELOG.

PR #1146, opened 2026-05-18
fix: parse skill frontmatter with CRLF newlines

liaosong0187765-maker:
## Summary
- Normalize optional UTF-8 BOM and CRLF line endings before parsing skill YAML frontmatter in check-resolvable.
- Add a Windows checkout regression test so CRLF SKILL.md files still expose their triggers.

## Why
On Windows checkouts, SKILL.md files may use CRLF endings. The existing parser only matched LF-delimited frontmatter, so check-resolvable and doctor could report false mece_gap warnings even when triggers were present.

## Test plan
- bun test test/check-resolvable.test.ts
- bun run typecheck
- gbrain check-resolvable --json
- gbrain doctor --json


garrytan:

PR #1149, opened 2026-05-18
fix(doctor): normalize CRLF in extractTriggers so Windows skill triggers parse

### LINKED ISSUE #1150: gbrain doctor reports 39 false mece_gap warnings on Windows (CRLF) — extractTriggers regex requires LF
## Summary

On Windows with default `core.autocrlf=true`, `gbrain doctor --fast --json` reports every skill not in `OVERLAP_WHITELIST` as a `mece_gap` warning, claiming the skills are missing `triggers:` arrays in their frontmatter — even when those arrays are clearly present.

The same `gbrain doctor` run also reports `skill_conformance: 42/42 skills pass`, so the two checks contradict each other within one report.

## Repro

1. Windows machine with default Git config (`git config --global core.autocrlf` returns `true`).
2. `git clone https://github.com/garrytan/gbrain` — SKILL.md files are checked out with CRLF line endings.
3. `gbrain doctor --fast --json` (any version that uses `src/core/check-resolvable.ts`'s `extractTriggers` function).

Result:

```json
{
  "name": "resolver_health",
  "status": "warn",
  "message": "39 issue(s): 0 error(s), 39 warning(s)",
  "issues": [
    {
      "type": "mece_gap",
      "skill": "query",
      "action": "Add a triggers: array to the frontmatter of skills/query/SKILL.md"
    }
    // ... 38 more
  ]
},
{
  "name": "skill_conformance",
  "status": "ok",
  "message": "42/42 skills pass"
}
```

`skills/query/SKILL.md` and 38 others already have valid `triggers:` blocks. Verifiable with:

```bash
grep -l "^triggers:" skills/**/SKILL.md | wc -l   # 42
```

## Root cause

`src/core/check-resolvable.ts` lines 166 and 169:

```ts
const fmMatch = skillContent.match(/^---\n([\s\S]*?)\n---/);
// ...
const triggersMatch = fm.match(/^triggers:\s*\n((?:\s+-\s+.+\n?)*)/m);
```

Both regexes anchor on literal `\n` (LF). On Windows the file content is `---\r\n` and `triggers:\r\n`, so `^---\n` never matches. The outer frontmatter match fails, `extractTriggers` returns `[]`, and every skill outside `OVERLAP_WHITELIST` (`ingest`, `signal-detector`, `brain-ops`) is flagged. 42 − 3 = 39 — exactly the count observed.

Hexdump of the first bytes of `skills/setup/SKILL.md` on Windows confirms CRLF:

```
0000000   -   -   -  \r  \n   n   a   m   e   :       s   e   t   u   p
```

## Why CI doesn't catch it

`.github/workflows/*.yml` all use `runs-on: ubuntu-latest`. Ubuntu checkouts preserve LF, so the regex works and the test at `test/check-resolvable.test.ts:258-283` ("v0.22.4 regression — actual repo skills/ has 0 errors") passes.

## Suggested fix

One-line CRLF normalization at the top of `extractTriggers`:

```ts
const content = skillContent.replace(/\r\n/g, '\n');
```

Preserves all existing LF behavior. PR coming with patch + unit test for CRLF input.

## Environment

- OS: Windows 11 Enterprise 10.0.26200
- gbrain: 0.35.7.0 (master)
- Bun: 1.3.13
- git: `core.autocrlf=true` (Windows default after Git-for-Windows installer)

samporter-31:
Fixes #1150

## Summary

`extractTriggers` in `src/core/check-resolvable.ts` used LF-anchored regexes (`^---\n` and `^triggers:\s*\n`). On Windows, where `core.autocrlf=true` is the default, SKILL.md files have CRLF endings and the regexes never matched. Result: `gbrain doctor --fast --json` on Windows reported 39 false `mece_gap` warnings — every skill except the 3 in `OVERLAP_WHITELIST` — while `skill_conformance` in the same run reported "42/42 skills pass". See linked issue for full repro.

CI is Ubuntu-only so the bug never surfaced upstream.

## Change

One-line CRLF normalization at the top of `extractTriggers`:

```ts
const content = skillContent.replace(/\r\n/g, '\n');
```

Function also exported so the new unit tests can target it directly (was file-scoped). No call-site changes.

## Tests

Added `describe("extractTriggers")` block in `test/check-resolvable.test.ts` with five cases:

- LF frontmatter parses triggers (baseline)
- **CRLF frontmatter parses triggers (regression case)**
- Missing frontmatter returns `[]`
- Missing `triggers:` field returns `[]`
- Quote-stripping (double, single, unquoted)

All 30 tests in `check-resolvable.test.ts` pass on Windows (`bun test test/check-resolvable.test.ts`).

## Verification on Windows

Before:

```
{"name":"resolver_health","status":"warn","message":"39 issue(s)..."}
health_score: 90
```

After:

```
{"name":"resolver_health","status":"ok","message":"42 skills, all reachable"}
health_score: 95
```

## Notes / follow-ups (out of scope)

- Other `\n`-anchored parsing in the same file (e.g. `parseResolverEntries` using `resolverContent.split('\n')`) appears to tolerate CRLF in practice (line-trailing `\r` is consumed by `trim()`), but a global pass might be worth a separate PR.
- Adding `*.md text eol=lf` to `.gitattributes` would prevent autocrlf conversion entirely for markdown files — orthogonal and stricter fix, but breaks the workflow of Windows-native users who edit in CRLF-only editors. The runtime normalization in this PR is the least-invasive option.
- Consider adding `windows-latest` to the CI matrix to catch future CRLF regressions automatically.


joshwilks111-max:
Independent confirmation on Windows: reproduced on Windows 11, Bun 1.3.13, gbrain 0.36.3.0. On a default `core.autocrlf=true` checkout, `extractTriggers` returns `[]` for every CRLF `SKILL.md` because `^---\n` can't match `---\r\n` — so `gbrain doctor --fast --json` flags every non-whitelisted skill as `mece_gap` while `skill_conformance` in the same run reports all skills passing, exactly as #1150 describes.

The `replace(/\r\n/g, '\n')` normalization here resolves it and is more robust than patching each regex to `\r?\n` individually, since the single up-front normalize also covers a trailing CRLF after the closing `---` fence. Test coverage looks complete (LF / CRLF / no-frontmatter / no-triggers / quote-stripping). +1 to merge.

Issue #449, opened 2026-04-26
Feature: path exclusion for `gbrain sync` (.gbrainignore or sync.exclude)


sahina:
## Problem
`isSyncable()` in `src/core/sync.ts` hardcodes which paths sync. Users with upstream tooling that owns directory conventions (e.g. Docusaurus blog with required `slug:` frontmatter that gbrain rejects) cannot exclude those dirs without:

1. Patching the hardcoded list (lost on upgrade)
2. Renaming to hidden dir (breaks the upstream tool)
3. Spamming `--skip-failed` forever (noise + still re-fails when files edit)

## Concrete case
`docs-site/blog/*.md` in a Docusaurus repo. Each post needs explicit `slug:` for clean URLs. gbrain rejects with `Frontmatter slug X does not match path-derived slug Y`. Repo-wide sync wants to skip the entire blog dir.

## Proposed
Either:
- **`.gbrainignore`** at repo root, gitignore syntax. Filter inside `isSyncable()` after the hardcoded checks.
- **`sync.exclude: string[]`** in gbrain config (glob patterns).

Either lets users opt-out paths without forking core.

## Version
gbrain 0.18.2

magarrent:
+1 — hit this on a private monorepo that holds an Astro marketing-site subdir under `<app>/src/content/posts/`. ~500 posts, all carry an explicit `slug:` in frontmatter because Astro routes against frontmatter, not path. Every `gbrain sync` blocks with **SLUG_MISMATCH** for the entire content collection.

Concrete output (gbrain 0.27.0, still hits on 0.33.x):
```
Skipped <path>/posts/<n>.md: Frontmatter slug "<post-slug>" does not match path-derived slug "<path>/posts/<n>" ...
...
Sync BLOCKED: N file(s) failed to parse.
```

`--skip-failed` works once but gives no persistent manifest, so next sync re-skips the same files (matches #500). Workaround patterns I considered all break the upstream tool:

- Rename to `archive/` → breaks Astro routing (file paths are public URLs).
- Strip `slug:` frontmatter → breaks SEO + breaks every internal link.
- Patch `isSyncable()` → lost on `bun update`.

Strong vote for `.gbrainignore` (gitignore syntax) over `sync.exclude` config, because it lives in the repo + travels with `git push`/`pull`. Docusaurus + Astro both have well-known content dirs (`blog/`, `src/content/posts/`, `src/content/docs/`) that should be one-line opt-outs.

[...]

PR #1159, opened 2026-05-18
feat(sync): opt-in --respect-gitignore for the full-import walker (#1073)

### LINKED ISSUE #1073: sync --strategy code enhancement: honor .gitignore (opt-in flag)
## Request

Add a flag to `gbrain sync --strategy code` that filters the file walker through `.gitignore` (and any other rules `git check-ignore` would honor — `.git/info/exclude`, global excludes). The flag can default off so existing behavior is preserved; users with gitignored build/output directories can opt in.

Today `collectSyncableFiles` in `src/commands/import.ts` only excludes hidden dirs + `node_modules` + `ops`. Repos that put generated artifacts under a gitignored path (`out/`, `dist/`, `build/`, `__pycache__/`, `coverage/`, `data/`, etc.) end up admitting every file in there as "code", because `CODE_EXTENSIONS` in `src/core/sync.ts` covers `.json`, `.yaml`, `.toml`, `.html`, `.css` along with the actual code extensions.

For repos with archived run outputs (ML training, backtests, build artifacts, log captures) this can mean the walker admits tens of thousands of files the operator never intended to index. The DB bloats; embedding cost goes up; search results get polluted by stale fixture data; and if one of those files happens to be pathological (e.g. a single-line giant JSON), it can wedge the chunker.

## Suggested shape

```
gbrain sync --strategy code --respect-gitignore   # opt in
gbrain sync --strategy code --ignore-from FILE    # arbitrary ignore file
```

Or as a config knob:

```yaml
sync:
  respect_gitignore: true
```

Mechanism: use `git ls-files --cached --others --exclude-standard` when the source has a `.git/` (the same shape `sync.ts:248` already uses for manifest building), and fall back to the existing walker for non-git source paths.

## Why not just always default-on

Some users intentionally want gitignored content indexed (e.g. dotfile brains, sensitive notes that live under `.gitignore` to keep them out of git but should still be brain-searchable). Keeping it opt-in respects both modes. A future major release could flip the default once users have had time to migrate.

## Related

- #449 asks for `.gbrainignore` / `sync.exclude` — a different mechanism for the same underlying need ("let me opt files out of sync"). The two can coexist: `.gitignore` integration is the easy win for repos that already express the right intent in their existing gitignore, and the `.gbrainignore` / `sync.exclude` route covers cases where the two intents diverge.

## Environment

- gbrain v0.35.0.0 (commit `baf1a47`)
- Filing on behalf of a user whose `--strategy code` sync stalled on tens of thousands of files it shouldn't have touched.

jetsetterfl:
Closes #1073.

## Problem

`collectSyncableFiles` (`src/commands/import.ts`) only skips dot-dirs / `node_modules` / `ops`. The full / first / `--full` import therefore walks gitignored build output (`dist/`, `out/`, `coverage/`, `__pycache__/`, …) and admits every file in there as "code" — `CODE_EXTENSIONS` covers `.json` / `.yaml` / `.toml` / `.html` / `.css`. On repos with large gitignored trees this bloats the DB, raises embedding cost, pollutes `code-def` / semantic search with stale fixtures/bundles, and can wedge the chunker on a pathological file (e.g. a single-line giant JSON). This is the stall reported in #1073. The incremental sync path is already git-based (`git ls-files --exclude-standard`, `sync.ts`) and excludes these — only the full-import walker diverges.

## Change (opt-in, default off)

- `collectSyncableFiles` gains `respectGitignore` in `CollectOpts`. When set **and** the root is a git work tree, `git ls-files -o -i --exclude-standard --directory` builds the ignored set; entries are pruned at **descent** time so a fully-ignored directory is never recursed into — addressing the walk-stall, not just emit-time filtering. Non-git root / git unavailable → empty set → **zero overhead, exact legacy behavior**.
- **Default OFF.** Preserves behavior for dotfile/secret brains that deliberately keep content out of git but want it brain-searchable (the explicit "why not just default-on" case in #1073).
- Surface area:
  - `gbrain sync --respect-gitignore` (and `--no-respect-gitignore` to override an enabled config)
  - `sync.respect_gitignore` config knob (CLI flag wins)
  - `gbrain import --respect-gitignore`
  - threaded into the `sync --all` cost preview so the estimate matches what will actually be walked
  - documented in `gbrain --help`

## Mechanism note

#1073 suggested `git ls-files --cached --others --exclude-standard` (an allowlist gated at emit time). I used the **ignored set with `--directory`** instead so whole gitignored trees are pruned *before* recursing — the issue's core pain is the walker stalling on tens of thousands of files, which directory-level pruning avoids and emit-time allowlisting does not. All existing walker hardenings (symlink/inode-cycle/max-depth) are untouched. Happy to switch to the allowlist shape if you'd rather.

## Tests

`test/import-walker.test.ts`:
- default (off) still admits gitignored `dist/bundle.js` — legacy behavior preserved
- opt-in prunes `dist/` + `coverage/` while keeping `src/app.ts`
- opt-in on a non-git dir falls back gracefully (no prune, no throw)

`bun run typecheck` clean; `bun test test/import-walker.test.ts` → 7 pass / 0 fail.

## Follow-ups (intentionally out of scope)

`--ignore-from FILE` and `.gbrainignore` / `sync.exclude` (#449) — this PR lands the easy win for repos that already express the intent in `.gitignore`.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -1549,6 +1549,7 @@ CODE INDEXING (v0.19.0 / v0.20.0 Cathedral II)
   reconcile-links [--dry-run]        Batch-recompute doc↔impl edges (v0.20.0)
   reindex-code [--source id] [--yes] Explicit code-page reindex (v0.20.0)
   sync --strategy code               Sync code files into the brain
+  sync --respect-gitignore           Prune git-ignored paths from the walker (#1073, opt-in)
 
 JOBS (Minions)
   jobs submit <name> [--params JSON]  Submit background job [--follow] [--dry-run]
--- a/src/commands/import.ts
+++ b/src/commands/import.ts
@@ -44,11 +44,14 @@ export interface RunImportResult {
 export async function runImport(
   engine: BrainEngine,
   args: string[],
-  opts: { commit?: string; strategy?: SyncStrategy; sourceId?: string } = {},
+  opts: { commit?: string; strategy?: SyncStrategy; sourceId?: string; respectGitignore?: boolean } = {},
 ): Promise<RunImportResult> {
   const noEmbed = args.includes('--no-embed');
   const fresh = args.includes('--fresh');
   const jsonOutput = args.includes('--json');
+  // #1073: opt-in. Honored either programmatically (opts.respectGitignore,
+  // threaded from `gbrain sync`) or directly on the `gbrain import` CLI.
+  const respectGitignore = opts.respectGitignore === true || args.includes('--respect-gitignore');
   // v0.30.x follow-up to PR #707: programmatic sourceId support so internal
   // callers (performFullSync, future Step 6 paths) can route to a named
   // source. The CLI `gbrain import` deliberately has no --source flag per
@@ -73,7 +76,7 @@ export async function runImport(
   const dirArg = args.find((a, i) => !a.startsWith('--') && !flagValues.has(i));
 
   if (!dirArg) {
-    console.error('Usage: gbrain import <dir> [--no-embed] [--workers N] [--fresh] [--json]');
+    console.error('Usage: gbrain import <dir> [--no-embed] [--workers N] [--fresh] [--json] [--respect-gitignore]');
     process.exit(1);
   }
   const dir: string = dirArg;  // narrowed; survives closure capture
@@ -85,7 +88,7 @@ export async function runImport(
   const strategy: SyncStrategy = opts.strategy ?? 'markdown';
   const _walkT0 = Date.now();
   console.error(`[gbrain phase] import.collect_files start dir=${dir} strategy=${strategy}`);
-  const allFiles = collectSyncableFiles(dir, { strategy });
+  const allFiles = collectSyncableFiles(dir, { strategy, respectGitignore });
   console.error(
     `[gbrain phase] import.collect_files done ${Date.now() - _walkT0}ms files=${allFiles.length}`,
   );
@@ -358,6 +361,16 @@ function resolveMaxWalkDepth(): number {
 
 interface CollectOpts {
   strategy?: SyncStrategy;
+  /**
+   * Opt-in (#1073): when true and the walk root is inside a git work tree,
+   * only files git would track or that are untracked-but-not-ignored are
+   * admitted — i.e. anything matched by `.gitignore` / `.git/info/exclude`
+   * / global excludes is pruned. Default `false` preserves legacy behavior
+   * (gitignored content stays indexable; required by dotfile/secret brains
+   * that deliberately keep notes out of git but want them brain-searchable).
+   * No effect for non-git roots (falls back to the legacy walker).
+   */
+  respectGitignore?: boolean;
 }
 
 /**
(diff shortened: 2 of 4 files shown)


spotshare-nick:
**Re: the "requires `ignore` npm package" exclusion in #1253**

Just want to flag for the record that this PR as it stands adds **no npm dependencies**. The diff touches only `src/cli.ts`, `src/commands/import.ts`, `src/commands/sync.ts`, and `test/import-walker.test.ts` — there's no `package.json` change. The gitignore set is built by shelling out to git:

```ts
execFileSync('git', ['-C', root, 'ls-files', '-o', '-i', '--exclude-standard', '--directory'], …)
```

`execFileSync` is a Node/Bun builtin (`child_process`), and `git` is already an assumed runtime tool everywhere else in the sync path (the incremental walker uses `git ls-files --exclude-standard` today). So the "no new deps" gate that excluded this in the v0.37.7.0 wave should actually pass — there's nothing to vendor. I think the exclusion may have conflated this PR with the broader `.gbrainignore` / #449 work, which *would* want a glob library for non-git roots.

**Two paths, your call:**

**Path A — keep the `git` subprocess (this PR).** Zero npm deps, correct gitignore semantics for free (negations, `**`, `.git/info/exclude`, global excludes), and directory-level pruning via `--directory`. The one constraint is it only fires inside a git work tree — non-git roots fall back to legacy behavior, which this PR already documents as intentional.

**Path B — if you'd rather not depend on the `git` executable at all,** the same intent can be done in-repo without a glob library, by reading `.gitignore` and matching it ourselves. Rough sketch:

```ts
// Parse .gitignore (+ .git/info/exclude) into ordered rules.
type Rule = { re: RegExp; negate: boolean; dirOnly: boolean };

function loadRules(root: string): Rule[] {
  const rules: Rule[] = [];
  for (const line of readIgnoreLines(root)) {           // skip '' and '#...'
    let pat = line.trim();
    if (!pat || pat.startsWith('#')) continue;
    const negate = pat.startsWith('!'); if (negate) pat = pat.slice(1);
    const dirOnly = pat.endsWith('/'); if (dirOnly) pat = pat.replace(/\/$/, '');
    const anchored = pat.startsWith('/'); if (anchored) pat = pat.slice(1);
    rules.push({ re: globToRegExp(pat, anchored), negate, dirOnly });
  }
  return rules;
}

// gitignore glob -> RegExp:  ** = any depth, * = within a segment, ? = one char.
// Unanchored patterns match at any path depth; anchored match from root.
function globToRegExp(pat: string, anchored: boolean): RegExp {
  const body = pat
    .replace(/[.+^${}()|[\]\\]/g, '\\$&')   // escape regex metachars
    .replace(/\*\*\//g, '(?:.*/)?')          // **/  -> any number of dirs
    .replace(/\*\*/g, '.*')                  // **   -> anything
    .replace(/\*/g, '[^/]*')                 // *    -> within one segment
    .replace(/\?/g, '[^/]');
  const prefix = anchored ? '^' : '(?:^|/)';
  return new RegExp(`${prefix}${body}(?:/.*)?$`);        // dir matches its children
}

// Last matching rule wins (gitignore precedence); negation re-includes.
function isIgnored(relPath: string, isDir: boolean, rules: Rule[]): boolean {
  let ignored = false;
  for (const r of rules) {
    if (r.dirOnly && !isDir) continue;
    if (r.re.test(relPath)) ignored = !r.negate;
  }
  return ignored;
}
```

Then in the walker, prune at descent time exactly like the current PR:

```ts
const rel = relative(root, full);
if (isIgnored(rel, isDirEntry, rules)) continue;
```

That's ~40 lines, no dependency, and gives the same directory-pruning win for both git and non-git roots — at the cost of owning a (small) slice of gitignore semantics ourselves. It won't be 100% spec-complete the way `git` is (e.g. nested per-directory `.gitignore` files, some edge cases), but it covers the `dist/`/`coverage/`/`out/` pain from #1073.

Happy to push either shape — Path A is already in the diff and is the lower-risk merge; Path B if you want to drop the `git` executable assumption entirely. Either way it lands without a new npm dep.

---

*This comment was drafted with the help of [Claude Code](https://claude.com/claude-code).*

time-attack:
Closing as obsolete — fix already on master (see 7c27fa12 Addresses #1159). Please reopen if still needed.

PR #1177, opened 2026-05-18
feat: --lock-duration and --low-pri-rate-cap flags on gbrain jobs work

### LINKED ISSUE #1014: feat: --lock-duration flag on `gbrain jobs work` to tune the wall-clock dead-letter cap
## Problem

The DB-side wall-clock detector in `handleWallClockTimeouts` uses `lockDuration` as a multiplier for the dead-letter threshold for jobs that don't set `timeout_ms` explicitly:

```sql
EXTRACT(EPOCH FROM (now() - started_at)) * 1000 >
  CASE WHEN timeout_ms IS NOT NULL THEN timeout_ms * 2
       ELSE lockDurationMs * 2 * GREATEST(max_stalled, 1) END
```

At gbrain defaults (`lockDuration=30000`, `max_stalled=3`) that cap is **180s = 3 minutes**, and `gbrain jobs work` doesn't expose either `--lock-duration` or a `GBRAIN_*` env var to bump it.

In practice this is too tight for heavy `claude --print` collector jobs (vision-heavy PDF extraction, multi-page bundle scrapes, orchestrators that fan out serially). On my queue, the slow-tail of successful jobs runs 125-167s — operating with literal seconds of headroom against the cap. Across one overnight cycle I had 7 collector jobs dead-letter with `attempts_made=0` and `error_text='wall-clock timeout exceeded'`. Since wall-clock kills go straight to `status='dead'` (no retry path — they don't increment `attempts_made`), the work is lost; the only recovery is whatever the cron's next fire happens to pick back up.

The workaround today is per-job `timeout_ms` at submit time, but that means touching every job submitter (~10 cron scripts across multiple sibling repos) and remembering to do it for any new wrapper. An operator-side global lever is the right hammer.

## Proposed fix

Add `--lock-duration MS` to `gbrain jobs work`, validated identically to `--health-interval` (positive integer milliseconds, reject suspicious sub-1000ms values that read as unit-confusion typos). Pass through to the existing `MinionWorker` constructor option that's already accepted (`opts?.lockDuration ?? 30000` in `worker.ts`). Startup banner mentions the active value.

```
gbrain jobs work --concurrency 4 --lock-duration 90000
```

Cap then becomes 90s × 2 × 3 = 540s = 9 minutes for non-`timeout_ms` jobs; lock-renewal correctness unchanged.

## Local patch

I have a working local patch (one commit, +35 / -2 in `src/commands/jobs.ts`) running in production for the last hour with `--lock-duration 90000`. Banner reports correctly, validation errors fire on bad input, dead-letter rate dropped to zero overnight.

Happy to open a PR if the shape is welcome — or if you'd prefer a different flag name (`--lock-ttl`?), an env var (`GBRAIN_LOCK_DURATION`?), or to expose `--max-stalled` instead since it's the third multiplier, let me know and I'll re-shape before submitting.

ethanbeard:
Closes #1014.

Adds two CLI flags to `gbrain jobs work` for tuning the wall-clock dead-letter behavior and low-priority pacing.

## `--lock-duration MS`

Exposes the row-lock TTL as a CLI flag (was hardcoded `30_000` in `worker.ts`). The wall-clock dead-letter cap is computed as `lockDuration * 2 * max_stalled`, so at defaults the cap is `30s × 2 × 3 = 180s` — too tight for vision-heavy `claude --print` collectors that legitimately run 5–10 minutes.

Bumping `--lock-duration 90000` raises the cap to 9 min and the wall-clock to 27 min for jobs without an explicit `timeout_ms`. No impact on lock-renewal correctness (worker still renews every `lockDuration / 2`).

Validation mirrors `--health-interval`: rejects NaN, non-positive, and sub-1000ms values (with a unit-confusion error message).

## `--low-pri-rate-cap N`

Rolling-hour cap on starts of jobs with `priority >= 0`. Default unset (no cap). When set, the claim SQL gates low-priority jobs so they collectively never exceed N starts per hour, leaving headroom for `priority < 0` jobs (always claimable, exempt from the cap).

Use case: shared external rate limit where background work shouldn't starve interactive paths. We hit this against the Anthropic Max 5-hour message window (~900 msg/5h on the alt account). Setting `--low-pri-rate-cap 90` caps background at ~50% of that ceiling.

## Compatibility

Both flags are additive and default-unset. `gbrain jobs work` with no new flags behaves exactly as before. No schema changes.

Worker banner now reports both values when set:

```
Minion worker started (queue: default, concurrency: 3, watchdog: 2048MB,
  health-check: 60s, lock-duration: 90s, low-pri-rate-cap: 90/hr (priority >= 0))
```

Validated end-to-end on a Supabase-backed brain (v0.35.8.0) running alongside ~13 launchd-managed wrappers that submit jobs.

### DIFF
--- a/src/commands/jobs.ts
+++ b/src/commands/jobs.ts
@@ -136,6 +136,7 @@ USAGE
   gbrain jobs smoke
   gbrain jobs work [--queue Q] [--concurrency N] [--max-rss MB]
                    [--health-interval MS] [--lock-duration MS]
+                   [--low-pri-rate-cap N]
   gbrain jobs supervisor [start] [--detach] [--json]
                          [--concurrency N] [--queue Q] [--pid-file PATH]
                          [--max-crashes N] [--health-interval N]
@@ -742,12 +743,33 @@ HANDLER TYPES (built in)
         lockDuration = parsed;
       }
 
+      // --low-pri-rate-cap: rolling-hour cap on starts of jobs with priority >= 0.
+      // Default unset (no cap). When > 0, the claim SQL gates low-priority jobs
+      // so they collectively never exceed N starts per hour, leaving headroom
+      // for `priority < 0` jobs (which are exempt and always claimable). This
+      // implements a "background tier shares leftover budget" policy against
+      // external rate limits — e.g. the Anthropic Max 5-hour message window
+      // (~900 msg/5h on the alt account). Setting this to 90 caps low-pri at
+      // 50% of that ceiling, preserving the remaining 50% for interactive
+      // / `priority < 0` work even when a background backlog is draining.
+      const lowPriRateCapRaw = parseFlag(args, '--low-pri-rate-cap');
+      let lowPriRateCap: number | undefined;
+      if (lowPriRateCapRaw !== undefined) {
+        const parsed = parseInt(lowPriRateCapRaw, 10);
+        if (!Number.isFinite(parsed) || parsed < 0) {
+          console.error(`Error: --low-pri-rate-cap must be a non-negative integer (jobs/hour), got "${lowPriRateCapRaw}"`);
+          process.exit(1);
+        }
+        lowPriRateCap = parsed;
+      }
+
       try { await queue.ensureSchema(); }
       catch (e) { console.error(e instanceof Error ? e.message : String(e)); process.exit(1); }
 
       const worker = new MinionWorker(engine, {
         queue: queueName, concurrency, maxRssMb, healthCheckInterval,
         ...(lockDuration !== undefined ? { lockDuration } : {}),
+        ...(lowPriRateCap !== undefined ? { lowPriRateCap } : {}),
       });
       await registerBuiltinHandlers(worker, engine);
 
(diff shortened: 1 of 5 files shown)


ethanbeard:
Bumping for review — about to upgrade our local checkout to v0.40.x and would prefer to drop this patch rather than rebase it forward.

If the design needs work, happy to iterate. If it's just review queue depth, I can rebase onto v0.40.x for a clean merge.

Context: we discovered the wrapper invoking `--lock-duration 90000` has been silently no-op'd since this patch left local master (the CLI's `parseFlag` is a bare `indexOf`, so unknown flags are accepted without error). So either this lands or we eat the default 3-min dead-letter cap on vision-heavy collectors.

time-attack:
Thanks for the detailed writeup and the end-to-end validation — the low-pri rate cap in particular is well thought out. This is new configuration surface on the worker rather than a fix for something broken, and we're keeping the minion queue's knob count deliberately small; a rolling-hour gate in the claim SQL is scheduling policy we'd rather not own per-operator. Closing on that basis, with no criticism of the implementation.

PR #1180, opened 2026-05-18
fix(embed): per-batch token cap for OpenAI + opt-in for LiteLLM

### LINKED ISSUE #1080: OpenAI/no-max_batch_tokens recipes may need per-key token sub-batching
During local rebase of upgrade/v030-patched onto origin/master v0.35.1.0, we dropped the local stale-embedding conflict in favor of upstream v0.34.4+ embed architecture.\n\nKnown gap to verify: local commit 5773631 had per-slug/per-key token-budget sub-batching (~50k tokens) before calling embedBatch. Upstream now has gateway-level adaptive splitting via splitByTokenBudget, but pre-splitting only happens when a provider recipe declares max_batch_tokens. For recipes without max_batch_tokens, including the documented OpenAI fast path, embedSubBatch(texts, ...) is called once and embedBatch still count-splits at 100 texts, not token-budget-splits.\n\nTask: add a focused regression test and, if needed, port conservative per-key token sub-batching into the upstream embedOneKey path using embedBatchWithBackoff(..., { abortSignal }).\n\nContext: this was treated as a follow-up, not a rebase blocker, because upstream v0.34.4+ materially improves stale embedding with cursor pagination, source scoping, rate-limit backoff, wall-clock abort budget, and tests.

kkroo:
Closes #1080. The gateway pre-split was gated on a recipe declaring `max_batch_tokens`. OpenAI was the documented "fast path" with no cap, so a single page whose chunks sum past OpenAI's 300K-tokens-per-request limit was rejected wholesale and dropped from indexing (we hit this on vendored Unicode tables in code-strategy syncs).

Three changes:

1. `recipes/openai.ts`: declare `max_batch_tokens: 300_000`, `chars_per_token: 4`. With the default `safety_factor: 0.8` this gives a 240K-token sub-batch budget, well under OpenAI's documented per-request ceiling. The existing shrink-on-miss recursive halving handles tokenizer-density variance.

2. `gateway.ts`: new `resolveMaxBatchTokens(recipe, embedding, env)` helper. For the `litellm` recipe (which can't ship a static cap — the proxy fronts arbitrary backends), users who know their backend can pin `LITELLM_MAX_BATCH_TOKENS=<N>` to opt into pre-split. Other recipes pass through their declared cap unchanged.

3. Dead warning-suppression branch cleanup: with OpenAI now declaring `max_batch_tokens`, the `id === 'openai'` special-case in `warnRecipesMissingBatchTokens()` is dead code. Removed + updated the surrounding comment.

Tests: updated "OpenAI fast path (D3)" → "OpenAI pre-split at declared max_batch_tokens" to assert the new behavior (large batches pre-split, small batches still single-call). Added a litellm env-override suite covering on/off and invalid-value fallback. 31 pass, 0 fail in the two affected files; 210 pass in test/ai/.

Docs: documented LITELLM_MAX_BATCH_TOKENS in
docs/integrations/embedding-providers.md and the litellm recipe setup_hint.

### DIFF
--- a/docs/integrations/embedding-providers.md
+++ b/docs/integrations/embedding-providers.md
@@ -102,6 +102,8 @@ Run [LiteLLM](https://docs.litellm.ai/docs/proxy/quick_start) in front of any pr
 
 This is the catch-all for "my provider isn't in the list above." Set up LiteLLM, then `gbrain init --embedding-model litellm:<your-model-id> --embedding-dimensions <N>`.
 
+**Large-page handling.** LiteLLM ships without a static batch cap because the proxy can front any backend, and the per-request token limit depends on which backend is wired up. For brains that index large files (vendored Unicode tables, generated code, etc.), pin the downstream backend's cap via `LITELLM_MAX_BATCH_TOKENS=<N>` so gbrain pre-splits each embed batch under that ceiling. Typical values: `300000` when LiteLLM fronts OpenAI, `120000` when fronting Voyage. Without this env var the recipe stays uncapped (current default behavior).
+
 ## Choosing dimensions
 
 Three numbers matter:
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -31,6 +31,7 @@ import { z } from 'zod';
 
 import type {
   AIGatewayConfig,
+  EmbeddingTouchpoint,
   MultimodalInput,
   Recipe,
   TouchpointKind,
@@ -412,13 +413,12 @@ function warnRecipesMissingBatchTokens(): void {
   for (const recipe of listRecipes()) {
     const embedding = recipe.touchpoints?.embedding;
     if (!embedding || embedding.max_batch_tokens !== undefined) continue;
-    // OpenAI is the canonical "no cap declared, fast path is intentional"
-    // recipe; suppress the warning for it. Every other recipe missing the
-    // field is suspicious.
-    if (recipe.id === 'openai') continue;
     // v0.32 (#779): explicit opt-out for dynamic-cap recipes (Ollama,
     // LiteLLM proxy, llama-server) — they ship without a static cap because
     // the cap depends on a user-launched server. Warning is noise for them.
+    // (OpenAI used to be suppressed by id here as the "intentional no-cap
+    // fast path"; #1080 closed that gap by declaring max_batch_tokens, so
+    // OpenAI now exits at the earlier line above.)
     if (embedding.no_batch_cap === true) continue;
     if (_warnedRecipes.has(recipe.id)) continue;
     _warnedRecipes.add(recipe.id);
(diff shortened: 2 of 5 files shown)


kkroo:
Validated end-to-end against a real failing payload from an in-cluster litellm proxy fronting OpenAI `text-embedding-3-large`. Adding a follow-up commit ([f8319c4](https://github.com/Blockcast/gbrain/commit/f8319c4)) that you should review alongside the initial commit — the original PR had a regex gap I found during validation.

### Repro

`Blockcast/pim-multicast-gateway` checks in two vendored Sony chip-register `.h` files (`cxd2880-dvbt2_siena_mrc_a2.h`, 142 chunks, ~5664 chars/chunk; `cxd2880-dvb_siena_mrc_a2_br.h`, 113 chunks). Dense hex/binary register definitions pack ~1.5 chars/token under tiktoken vs the 4-char default. Estimated 200K tokens per page at 4 chars/token is comfortable; real token count is ~500K, well past OpenAI's 300K-per-request cap.

### Pre-fix state (master @ `1dadd9e`, no `LITELLM_MAX_BATCH_TOKENS`)

\`\`\`
gstack-code-pim-cluster | missing=255 | total=13255 | pct=1.92
\`\`\`

Both pages had 0/142 and 0/113 chunks embedded — the entire page's chunks were skipped after one rejected `embedMany` call. `gbrain sync` reported \"ok\" and the failures were invisible to the operator.

### Post-fix state (this PR @ `f8319c4`, `LITELLM_MAX_BATCH_TOKENS=300000`)

\`\`\`
All 7 sources: missing=0
embed-stale job log: \"Embedded 255 chunks across 2 pages\"
\`\`\`

### The regex gap

The original commit (`738e1ad`) added the cap + env override but the safety-net recursion stayed dormant because `isTokenLimitError` didn't match the format litellm actually returns:

\`\`\`
litellm.BadRequestError: OpenAIException - Error code: 400 - {'error': {'message': \"Invalid 'input': maximum request size is 300000 tokens per request.\"}}
\`\`\`

The existing regexes target Voyage-style messages (`max allowed tokens per submitted batch`, `batch contains too many tokens`, `token limit exceeded`). The OpenAI/litellm phrasing — `maximum request size is N tokens per request` — doesn't match any of them. `f8319c4` adds two regex branches for this variant + a unit test using the actual error string.

This is the variant most users will hit, since the pre-split is by tiktoken estimate and dense tokenized content (hex/binary tables, CJK, code with long identifiers) routinely exceeds the per-text estimate.

### Tests

\`\`\`
bun run typecheck — clean
test/ai/adaptive-embed-batch.test.ts — 33 pass / 0 fail
test/ai/ (whole dir) — 212 pass / 0 fail
\`\`\`

time-attack:
Closing: current master already contains this fix. Master already fixed the core bug via a different, deliberately more conservative implementation: src/core/ai/recipes/openai.ts:25 declares `max_batch_tokens: 100_000` (with a rationale comment choosing 100K over the PR's 300K to handle token-dense content under both the 300K/request and 1M TPM ceilings), and pre-split fires at src/core/ai/gateway.ts:1479-1481. LiteLLM is handled via `no_batch_cap: true` (recipes/litellm-proxy.ts:33, honored at gateway.ts:586). PR is OPEN but its change 1 conflicts with master's chosen value; only residuals are the marginal LITELLM_MAX_BATCH_TOKENS env opt-in and removing the now-dead `id === 'openai'` special-case at gateway.ts:~580 — neither justifies merging a superseded diff.

PR #1183, opened 2026-05-18
fix(doctor): run PGLite probes through active engine

duncanclaw:
## Summary

Fix two false-positive `gbrain doctor` warnings for PGLite-backed brains:

- route the `pgvector` probe through the active `BrainEngine`
- route the `jsonb_integrity` probe through the active `BrainEngine`
- add focused PGLite regression coverage for both checks

## Root Cause

The doctor had two remaining direct calls to the module-level Postgres singleton:

- `db.getConnection()` for `pg_extension`
- `db.getConnection().unsafe(...)` for JSONB integrity scanning

That works for Postgres-backed brains, but PGLite uses `PGLiteEngine` and does not initialize the Postgres singleton. The result was a false warning on healthy embedded brains:

```text
[WARN] pgvector: Could not check pgvector extension
[WARN] jsonb_integrity: Could not check JSONB integrity
```

The underlying PGLite engine can answer both SQL probes correctly; the doctor was just bypassing it.

## Real Behavior Proof

- **Behavior addressed:** embedded PGLite brains no longer warn on `pgvector` and `jsonb_integrity` solely because no external Postgres singleton is connected.
- **Real environment tested:** local GBrain checkout with an embedded PGLite engine, plus a PGLite-backed brain used for doctor smoke verification.
- **Exact commands run after this patch:**

```bash
bun test test/doctor.test.ts
bun run typecheck
bash scripts/check-no-legacy-getconnection.sh
bun run check:jsonb
gbrain doctor
```

- **Evidence after fix:**

```text
(pass) doctor command > pgvector and jsonb_integrity checks use the active PGLite engine
49 pass
0 fail

[OK] pgvector: Extension installed
[OK] jsonb_integrity: All JSONB columns store objects/arrays
```

- **Observed result after fix:** the two PGLite-specific false warnings are gone. The local doctor smoke still surfaced an unrelated resolver warning from the checkout, but `pgvector` and `jsonb_integrity` both report OK.
- **What was not tested:** live Supabase/self-hosted Postgres doctor output. The SQL and status semantics are unchanged for Postgres; this PR only routes the probes through the active engine abstraction.

## Regression Test Plan

- **Coverage level that should have caught this:** focused doctor unit coverage against a real PGLite engine.
- **Target test:** `test/doctor.test.ts`
- **Scenario locked in:** `pgvectorCheck(PGLiteEngine)` and `jsonbIntegrityCheck(PGLiteEngine)` both return OK after schema initialization.
- **Why this is the smallest reliable guardrail:** the bug was not pgvector or JSONB behavior; it was the doctor using the wrong connection path.

## Verification

- `bun test test/doctor.test.ts`
- `bun run typecheck`
- `bash scripts/check-no-legacy-getconnection.sh`
- `bun run check:jsonb`

## User-visible Behavior

PGLite users should see fewer false `gbrain doctor` warnings. No CLI flags, config keys, migrations, or data-shape changes.

### DIFF
--- a/src/commands/doctor.ts
+++ b/src/commands/doctor.ts
@@ -146,6 +146,70 @@ export async function whoknowsHealthCheck(_engine: BrainEngine): Promise<Check> 
   }
 }
 
+/**
+ * Doctor check: pgvector availability.
+ *
+ * Use the active engine instead of the module-level Postgres singleton.
+ * PGLite exposes pg_extension through its engine connection, but does not
+ * connect db.ts's Postgres singleton; using db.getConnection() here turns a
+ * healthy PGLite brain into a false warning.
+ */
+export async function pgvectorCheck(engine: BrainEngine): Promise<Check> {
+  try {
+    const ext = await engine.executeRaw<{ extname: string }>(
+      `SELECT extname FROM pg_extension WHERE extname = 'vector'`,
+    );
+    if (ext.length > 0) {
+      return { name: 'pgvector', status: 'ok', message: 'Extension installed' };
+    }
+    return { name: 'pgvector', status: 'fail', message: 'Extension not found. Run: CREATE EXTENSION vector;' };
+  } catch {
+    return { name: 'pgvector', status: 'warn', message: 'Could not check pgvector extension' };
+  }
+}
+
+/**
+ * Doctor check: JSONB columns are not double-encoded as strings.
+ *
+ * This check is valid on both Postgres and PGLite. Route through
+ * engine.executeRaw() so embedded PGLite brains are checked through their
+ * actual connection instead of the unrelated Postgres singleton.
+ */
+export async function jsonbIntegrityCheck(
+  engine: BrainEngine,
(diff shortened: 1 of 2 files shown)


eloe:
Local verification on macOS + PGLite confirms this is still biting current `master` (v0.42.53.0) when using PGLite: `gbrain doctor --json` reports false warnings for `pgvector` and `jsonb_integrity`, which in turn causes watchdog/canary scripts to report degradation even though CLI/search/sync are otherwise healthy.

I carried a local skip/OK patch for those two PGLite-only probes and verified:

- `gbrain doctor --json` health score improves to 95 with only `retrieval_reflex_health` warning remaining
- `python3 ~/.hermes/scripts/gbrain_canary_probe.py` exits 0
- `python3 ~/.hermes/scripts/gbrain_daily_maintenance.py` exits 0

So this PR is the right upstream lane; I’m not opening a duplicate. If maintainers want a fresh rebase/replacement against current `master`, I can provide one, but #1183 is currently still mergeable according to GitHub.

PR #1185, opened 2026-05-18
feat(integrations/doctor): surface dead-jobs from minions queue

ethanbeard:
## Why

`gbrain integrations doctor` currently only runs the per-recipe `health_checks:` declared in each integration's YAML frontmatter. It has zero visibility into the minions job queue, which means any pipeline that submits shell or subagent jobs has a structural blind spot:

1. Cron fires the collector
2. Collector submits a job to the queue → submission succeeds
3. Collector's heartbeat writes `status:ok exit_code:0`
4. Worker picks up the job, runs it, dies
5. Job ends up in `dead` state
6. Nobody knows — the collector stayed green, the doctor didn't surface it

I hit this in the wild: 3,561 shell jobs from one collector had silently died over several days with `{"signal":"error","reason":"empty result from codex"}`. The collector itself was happily heartbeating green the whole time.

## What

Adds one cross-cutting check to `cmdDoctor()` in `src/commands/integrations.ts`. Shells out to `gbrain jobs list --status dead` and counts results grouped by job name. Surfaces under a new `[queue]` section alongside the per-recipe sections:

```
  x-to-brain: ISSUES
  ✗ X API: HTTP 403
  [queue]: ISSUES
  ✗ 3561 dead job(s) in minions queue (shell:3560, autopilot-cycle:1).
    Inspect: gbrain jobs list --status dead | head; details: gbrain jobs get <id>

  OVERALL: 2 issue(s) found
```

JSON mode includes the same check with `integration: "[queue]"`.

## Design notes

- **Shells out instead of opening a DB connection.** The file's top-level comment explicitly states it's a standalone CLI command with no DB connection. I preserved that by piping through the existing `gbrain jobs list` CLI rather than executing SQL directly. Slightly slower (subprocess vs query) but architecturally aligned.
- **Best-effort, silent skip on failure.** If `gbrain jobs` is unavailable (DB unreachable, etc.) the check is dropped and the per-recipe checks still run. Doctor never becomes flakier than before.
- **Global count, not per-integration.** `MinionJob` has no `integration_id` field today, so we can't attribute dead jobs to specific recipes. The name-based breakdown (`shell:N, subagent:M, autopilot-cycle:K`) gives the operator enough to find the broken pipeline. A future enhancement could add `integration_id` to the job schema for proper per-integration attribution — this PR is a stop-gap that gets most of the value at none of the schema-migration risk.
- **Output capped at `--limit 99999`.** Memory-bounded, well above realistic dead-pool sizes; parses in under 1s for typical queues.

## Test plan

- [x] Run `gbrain integrations doctor` with no dead jobs → check is silent, no `[queue]` section appears
- [x] Run with 3,561 real dead jobs → check surfaces under `[queue]: ISSUES` with the breakdown
- [x] JSON mode includes the new check with `integration: "[queue]"`
- [x] No regression to per-recipe checks (X-to-brain still flagged as before)

## Out of scope

- Per-integration job attribution (requires schema migration)
- Stalled-job detection (already covered by `queue_health` check in `doctor.ts`)
- Auto-replay or auto-cleanup of dead jobs

### DIFF
--- a/src/commands/integrations.ts
+++ b/src/commands/integrations.ts
@@ -695,6 +695,20 @@ async function cmdDoctor(args: string[]): Promise<void> {
     }
   }
 
+  // Cross-cutting check: dead jobs in the minions queue. Shell jobs submitted
+  // by collectors (signal-detect, email-collector, art-signals, etc.) die
+  // silently from the per-integration perspective \u2014 the collector's drain
+  // sees the submit succeed and stays green, while the worker job dies on
+  // the queue side. Without this check, an entire collector pipeline can
+  // be broken for days before anyone notices. Empirically validated 2026-05-18
+  // when 3,561 email-collector batches died with "empty result from codex"
+  // and no surface flagged it.
+  //
+  // Implementation deliberately shells out to `gbrain jobs list` to preserve
+  // this file's "no DB connection" design \u2014 see file-level comment.
+  const deadJobsCheck = checkDeadJobs();
+  if (deadJobsCheck) results.push(deadJobsCheck);
+
   if (jsonMode) {
     const fails = results.filter(r => r.status !== 'ok');
     console.log(JSON.stringify({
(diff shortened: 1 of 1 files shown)


ethanbeard:
Bumping for review — still carrying this as a local patch and it surfaced 9 dead jobs in our queue this week that would otherwise have been invisible.

Happy to rebase onto v0.40.x if that helps land it. The change is isolated to `cmdDoctor` in `integrations.ts` and the v0.40.0.0 install/refresh additions are in a different region of the file, so the rebase is clean (verified via `git merge-tree`).

time-attack:
Superseded with credit by #3129 (just merged) — your change landed rebased onto current master with authorship preserved via Co-authored-by. Thank you!

PR #1218, opened 2026-05-20
fix(cycle): cache empty propose_takes scans

AdityaRajeshGadgil:
## Summary

`propose_takes` used `take_proposals` as both the operator-facing queue and the per-page idempotency cache. That works only when the extractor produces at least one proposal. The valid result `[]` left no row behind, so unchanged pages could re-spend extractor tokens every autopilot cycle.

This PR splits those two jobs:

- adds `take_proposal_page_scans` as the successful extractor-call cache, keyed by `(source_id, page_slug, content_hash, prompt_version)` and written even for `[]`
- keeps `take_proposals` as proposal rows only, with dedup moved to `(source_id, page_slug, content_hash, prompt_version, claim_text)` so same-page multi-claim output is not collapsed
- treats malformed or wrong-shape extractor output as retryable failure instead of caching it as empty success
- threads `dryRun` into `propose_takes` so dry runs do not call the LLM or write proposal/cache rows
- adds migration v80 plus fresh Postgres/PGLite schema coverage, including explicit Postgres RLS enablement for the new cache table

## Test Coverage

- `bun test ./test/propose-takes.test.ts ./test/propose-takes-cache-schema.test.ts`
- `bun test ./test/schema-bootstrap-coverage.test.ts ./test/propose-takes.test.ts ./test/propose-takes-cache-schema.test.ts`
- `bun run typecheck`
- `bun run check:all`
- `bun run build`
- `git diff --cached --check`

`bun run ci:local:diff` could not run in this environment because Docker is not installed (`docker: command not found`). The diff-aware gate reached Docker startup before failing.

## Pre-Landing Review

Two independent review passes were run before opening this PR. The first review found two blockers — wrong-shape JSON could still be cached as `[]`, and the new cache table needed explicit RLS coverage — plus an insert-count nit. This version fixes all three:

1. strict production parsing now throws on malformed/wrong-shape output, while literal JSON `[]` remains a valid cacheable empty result
2. migration v80 and fresh schema SQL enable RLS for `take_proposal_page_scans` on Postgres, with a PGLite override for the local engine
3. proposal inserts use `RETURNING id`, so `proposals_inserted` reflects rows actually inserted after `ON CONFLICT DO NOTHING`

The second review passed with no blocking findings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to GBrain will be documented in this file.
 
+## [Unreleased]
+
+### Fixed
+
+- `propose_takes` now caches successful zero-proposal page scans in `take_proposal_page_scans`, so unchanged pages that correctly produce `[]` do not re-spend LLM tokens on every autopilot cycle. The operator-facing `take_proposals` queue remains proposal-only; extractor failures are deliberately uncached so transient model/API errors retry later.
+
 ## [0.37.1.0] - 2026-05-19
 
 **Your brain can now collide its own ideas.**
--- a/src/core/cycle.ts
+++ b/src/core/cycle.ts
@@ -1420,7 +1420,7 @@ export async function runCycle(
           checkAborted(opts.signal);
           progress.start('cycle.propose_takes');
           const { runPhaseProposeTakes } = await import('./cycle/propose-takes.ts');
-          const { result, duration_ms } = await timePhase(() => runPhaseProposeTakes(calibrationCtx, { repoPath: opts.brainDir }) as Promise<PhaseResult>);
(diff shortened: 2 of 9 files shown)


time-attack:
Your diagnosis was exactly right — the valid `[]` result left no idempotency row and unchanged pages re-spent extractor tokens every cycle. Master fixed it in #3319 with a `status='rejected'` tombstone row (`EMPTY_EXTRACTION_TOMBSTONE_TEXT`), and the dedup key already carries `md5(claim_text)` so multi-claim output isn't collapsed either. Closing as already fixed; the separate cache table plus migration v80 would also now collide with master at v125.

PR #1227, opened 2026-05-20
fix(autopilot): scope lock file to GBRAIN_HOME

### LINKED ISSUE #1226: autopilot lock file ignores GBRAIN_HOME — second brain silently respawns forever
## Summary

`gbrain autopilot` uses a lock file at `$HOME/.gbrain/autopilot.lock` regardless of `GBRAIN_HOME`. When two brains run on the same host (each with its own `GBRAIN_HOME` and database), they compete for the same lock — the second autopilot silently exits with code 0 every time, producing a silent respawn loop under launchd/systemd.

## Affected

- v0.35.6 (confirmed; introduced when multi-brain support landed)
- Any setup with multiple brains sharing a host

## Repro

1. Two brains: `~/.gbrain/` and `~/.gbrain-side/`, each with its own `config.json` and Postgres DB.
2. Two launchd `KeepAlive` agents, each setting its own `GBRAIN_HOME` env:
   - `com.gbrain.autopilot` → `GBRAIN_HOME=~/.gbrain`
   - `com.gbrain-side.autopilot` → `GBRAIN_HOME=~/.gbrain-side`
3. Start both.
4. Observe: only the first autopilot to take the lock does any work. The second exits with `"Another autopilot instance is running (lock file is fresh). Exiting."` and launchd respawns it every `ThrottleInterval` seconds.

In our setup this produced **46,388 silent failures in ~3 days** before we traced it.

## Root cause

`src/commands/autopilot.ts`:

```ts
const lockPath = join(process.env.HOME || '', '.gbrain', 'autopilot.lock');
```

The path is hardcoded to `$HOME/.gbrain/`, ignoring `GBRAIN_HOME`. Every other piece of state (config, sync_repo_path, DB) is correctly scoped by `GBRAIN_HOME`; the lock is the one place it isn't.

## Why it stays invisible

- Exit code is `0` (not a crash) → launchd considers it a normal restart.
- Log line is informational, not an error.
- `launchctl list` shows the agent as healthy because it's always "currently running" between respawns.
- Process never gets to `gbrain dream` / extract / embed / orphans → the brain stops evolving but nothing alerts.

## Fix

Scope the lock path to `GBRAIN_HOME` when set. PR: #<TBD>

```ts
const gbrainHome = process.env.GBRAIN_HOME || join(process.env.HOME || '', '.gbrain');
const lockPath = join(gbrainHome, 'autopilot.lock');
```

Each brain gets its own lock; they coexist cleanly. Single-brain installs are unaffected (fallback to `$HOME/.gbrain`).

rafaelreis-r:
Fixes #1226.

## Problem

`gbrain autopilot` hardcodes the lock file to `$HOME/.gbrain/autopilot.lock`, ignoring `GBRAIN_HOME`. When multiple brains share a host, only one autopilot ever runs — the others silently exit on every launchd/systemd respawn, producing tens of thousands of invisible failures.

See #1226 for the full story (46,388 silent failures observed in ~3 days on our dual-brain setup).

## Fix

Two-line change: derive the lock path from `GBRAIN_HOME` when set, fall back to `$HOME/.gbrain` for single-brain installs.

```ts
const gbrainHome = process.env.GBRAIN_HOME || join(process.env.HOME || '', '.gbrain');
const lockPath = join(gbrainHome, 'autopilot.lock');
```

Same fallback updates applied to the `mkdirSync` call so the directory creation is also scoped correctly.

## Verification

- `bun run typecheck` clean.
- Tested locally on a dual-brain mac-mini setup: both autopilots now coexist with separate `autopilot.lock` files under their respective `GBRAIN_HOME` dirs.
- Single-brain installs unaffected (env var unset → same fallback path as before).

## Compatibility

- Backwards compatible: existing single-brain setups behave identically.
- No schema or config changes.
- Stale lock takeover logic (>10min mtime) unchanged.


garrytan:
Closing in favor of #1253 (v0.37.7.0 fix wave) which re-implements the same fix against current master via `gbrainPath('autopilot.lock')` — the canonical GBRAIN_HOME-aware helper from `src/core/config.ts`. Same fix shape, same outcome.

The implementation lives at `src/commands/autopilot.ts:122-130` with a regression test at `test/autopilot-lock-path.test.ts`. Your contribution is acknowledged via `Co-Authored-By: rafaelreis-r` trailer. Thank you for the quick patch — closing #1226 honestly.

PR #1231, opened 2026-05-20
Improve facts extraction observability

shandutta:
## Summary

- make facts extraction check availability for the configured extraction model instead of the default chat model
- log absorbed zero-insert paths (`gateway_unavailable`, `no_candidates`) to `ingest_log` via `facts:absorb`
- pass `sourceId` through the facts backstop path so observability rows are source-scoped
- preserve prefixed slug-shaped entity hints like `projects/gbrain` so facts can fence-write to the intended slug instead of collapsing to `projects-gbrain`

## Why

When the default chat model is unavailable but `facts.extraction_model` is configured to an available model, extraction currently short-circuits and returns zero facts silently. That makes hot-memory failures look like conservative extraction rather than an observable config/runtime issue.

The prefixed-slug fix addresses another hot-memory routing issue: callers can provide a canonical entity hint such as `projects/gbrain`, but the resolver falls through to `slugify()` when no page exists yet, producing `projects-gbrain` and missing the intended fence path.

## Tests

```bash
bun test test/facts-canonicality.test.ts test/facts-backstop-integration.test.ts
```

Result: 17 pass, 0 fail, 47 expectations.

Also live-smoked locally with `facts.extraction_model=openai:gpt-4o-mini`: `extract_facts` inserted and recalled a temporary fact under `entity_slug=projects/gbrain`; cleanup used `forget_fact` and active recall returned no facts.

### DIFF
--- a/src/core/entities/resolve.ts
+++ b/src/core/entities/resolve.ts
@@ -47,11 +47,14 @@ export async function resolveEntitySlug(
   const trimmed = raw.trim();
   if (!trimmed) return null;
 
-  // 1. Exact match on slug. If raw already looks like a slug (or matches
-  //    a row exactly), use it.
+  // 1. Exact match on slug. If raw already carries a directory prefix
+  //    (`people/alice`, `projects/gbrain`), preserve it even before a page
+  //    row exists so fact extraction can fence-write to the intended page
+  //    instead of collapsing `/` to `-` via slugify fallback.
   if (looksLikeSlug(trimmed)) {
     const exact = await tryExactSlug(engine, source_id, trimmed);
     if (exact) return exact;
+    if (trimmed.includes('/')) return trimmed;
   }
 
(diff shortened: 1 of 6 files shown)


time-attack:
Thanks — the underlying finding is genuinely correct: the availability gate checks the default chat model rather than the resolved `facts.extraction_model`, so extraction silently returns zero facts. Unfortunately this patch no longer lands where the gate lives; master refactored to `extractFactsFromTurnWithOutcome` (extract.ts:246) with a typed `chat_unavailable` outcome and `extractFactsFromTurn` is now just a wrapper, so this shape would fix the wrapper and leave the real check untouched. The bundled `resolve.ts` early-return also skips the fuzzy-match arm for any slash-shaped hint, which would stub `people/alice` as a phantom instead of resolving it to an existing page — that arm is deliberate. Closing this one.

Issue #769, opened 2026-05-09
Code chunks land in DB with NULL language / symbol_name / symbol_type across all languages


rayers:
## Symptoms

After `gbrain sync --strategy code` walks a multi-language repo (1882 source files: Java + Python + C + TypeScript + JavaScript + bash + CSS + HTML + YAML + JSON, etc.), the chunks land in `content_chunks` with the chunk text correctly tagged in the body (e.g., `[Java] dview/.../DstreamView.java:5-5 import java.awt.C`), but the `language`, `symbol_name`, `symbol_type`, and `symbol_name_qualified` columns are NULL on essentially every chunk.

`code-def <symbol>` returns 0 hits as a result — the query filters on `symbol_name` which is unpopulated. `code-refs` still works because it's a text-scan over `chunk_text`.

## Concrete numbers from a freshly-synced nvr repo (~31k chunks)

After `gbrain sync --source <id> --strategy code` finishes successfully, plus the autopilot daemon doing a few maintenance cycles plus a manual `gbrain extract all`:

| Source language (inferred from slug) | chunks | with language | with symbol_name | with symbol_type |
|--------------------------------------|--------|---------------|------------------|------------------|
| java | 16,663 | 0 | 0 | 0 |
| python | 10,541 | 0 | 0 | 0 |
| c | 3,024 | 0 | 0 | 0 |
| javascript | 692 | 0 | 0 | 0 |
| c-header | 461 | 0 | 0 | 0 |
| bash | 234 | 0 | 0 | 0 |
| json | 117 | 104 | 0 | 104 |
| html | 59 | 0 | 0 | 0 |
| yaml | 52 | 0 | 0 | 0 |
| css | 50 | 0 | 0 | 0 |
| cpp | 14 | 0 | 0 | 0 |
| typescript | 1 | 0 | 0 | 0 |

JSON's 104 chunks with `symbol_type` are the only language with consistent metadata; everything else is empty.

The chunker code path looks plausible:
- `src/core/chunkers/code.ts:542-579` calls `extractSymbolName(node)` and `normalizeSymbolType(node.type)` to populate metadata.
- `src/core/import-file.ts:127, 470` reads `c.metadata.symbolName` and writes to the `symbol_name` column.
- The chunkable-node Sets at `src/core/chunkers/code.ts:296-300` for Java look right (`method_declaration`, `class_declaration`, `interface_declaration`, etc.).

[...]

PR #768, opened 2026-05-09
fix(sync+embed+extract): code-symbol ingest + graph extraction (#767 + #769 + extract follow-ups)

rayers:
## Summary

Three related fixes for code-symbol ingest and graph extraction:

- **#767** — `gbrain sync --strategy code` silently ran a markdown-only import on first sync.
- **#769** — once code chunks land, every re-embed pass clobbered their metadata to NULL.
- **Extract follow-ups** (no separate issue filed yet) — `gbrain doctor` recommended non-existent commands; body wikilinks like `[[bare-name]]` were silently dropped because they fell outside `DIR_PATTERN` and the resolver wasn't routed through markdown extraction. Surfaced while diagnosing why post-#769 brains showed `graph_coverage 0%` even after running the (correct) extract command.

All three ship together: each is a layer of the same pipeline. Without #767 the chunks don't exist; without #769 their metadata is dead; without the extract fixes the graph stays empty for any brain not formatted in YC/CRM-style `[Name](people/slug)` markup.

### `src/commands/import.ts`
- `runImport` now parses `--strategy markdown|code|auto` (defaults to `markdown` for backward compat).
- Both space-separated and equals-separated forms accepted.
- `auto` strategy is computed as a UNION of `collectMarkdownFiles ∪ collectFilesByStrategy('code')` — strict superset of legacy markdown coverage.
- New `collectFilesByStrategy(dir, strategy)` helper. Mirrors `collectMarkdownFiles` symlink + hidden-dir + node_modules safety.

### `src/commands/sync.ts`
- `performFullSync` threads `opts.strategy` through to `runImport`'s `importArgs`. Same fix in the `dryRun` branch.

### `src/commands/embed.ts`
Three re-upsert call sites (`embedOnePage`, `embedAll` non-stale, `embedAllStale` autopilot path) were stripping the 8 metadata fields. New `preserveCodeMetadata` helper threads them through consistently.

### `src/core/postgres-engine.ts` + `src/core/pglite-engine.ts`
`upsertChunks` ON CONFLICT clause was overwriting metadata columns from `EXCLUDED` unconditionally. Applied the chunk_text-gated CASE pattern to all 8 metadata columns (re-chunk → trust EXCLUDED, re-embed → COALESCE preserve), symmetric with the existing `embedding`/`embedded_at`/`model` pattern.

### `test/embed.serial.test.ts`
Three regression tests cover all three runEmbed paths. Each fails on pre-fix code (8 fields come back undefined), passes on the patched code.

[...]

Issue #767, opened 2026-05-09
sync --strategy code dropped on first sync via performFullSync


rayers:
## Summary

`gbrain sync --strategy code --source <id>` silently does a markdown-only import on first sync (no anchor commit yet). The `--strategy` flag is parsed correctly but dropped when `performFullSync` calls `runImport`, which is hardcoded to walk markdown only via `collectMarkdownFiles`. As a result, registering a fresh code source and running first sync produces 0 code pages — no error, just silent under-coverage. `code-def` / `code-refs` / `code-callers` / `code-callees` then return 0 hits for every symbol in the repo.

## Steps to reproduce

```bash
# Start with a fresh code source
gbrain sources add gstack-code-myrepo --path /path/to/myrepo --federated

# First sync with --strategy code
cd /path/to/myrepo
gbrain sync --source gstack-code-myrepo --strategy code --no-embed

# Output: 0 pages imported, N skipped (where N = the number of .md files
#         that already exist as duplicates in default source)
# Expected: thousands of code pages from .ts/.py/.java/.c/etc.
```

Verify:
```bash
gbrain code-def <known_symbol>  # returns count: 0
```

## Root cause

`src/commands/sync.ts:847` — `performFullSync` calls `runImport` from `./import.ts`:

```ts
const { runImport } = await import('./import.ts');
const importArgs = [repoPath];
if (opts.noEmbed) importArgs.push('--no-embed');
if (fullConcurrency > 1) importArgs.push('--workers', String(fullConcurrency));
const result = await runImport(engine, importArgs, { commit: headCommit });
```

Note that `opts.strategy` is in scope but never threaded into `importArgs`. `runImport` then uses `collectMarkdownFiles(repoPath)` which only enumerates `*.md`. The `walkSyncableFiles(strategy)` path that `incremental` mode uses (sync.ts:111-134) is bypassed entirely on first sync.

The dry-run branch above (sync.ts:858) has the same drop — `collectMarkdownFiles` regardless of `opts.strategy`.

## Affected behavior

- First-time `gbrain sources add <id> --path <repo>` followed by `gbrain sync --strategy code` produces no code pages.
- Subsequent invocations of `sync --strategy code` after the first still hit the same path because `last_commit` was never advanced.
- Workaround `gbrain reindex-code --source <id> --yes` prints `No code pages to reindex` (chicken-and-egg — needs code pages first).
[...]

PR #1233, opened 2026-05-20
fix(extract): resolve bare-name body wikilinks via resolver

rayers:
## Summary

Body wikilinks in wiki/topic/learning content are silently dropped on every `gbrain extract` pass. Three layered issues:

1. `WIKILINK_RE` in `src/core/link-extraction.ts` is gated on `DIR_PATTERN` (`people|companies|meetings|...`). Wiki/topic/learning content uses bare-name wikilinks like `[[Fast-Weigh]]` or `[[2026-05-07-cost-plan]]` which fall outside that allow-list — the regex never matched, so body refs were invisible to extract.
2. Body wikilinks that DID match were only resolved when `--include-frontmatter` was set, because `extractPageLinks` routed ALL refs (body + frontmatter) through `activeResolver` which was set to `nullResolver` when frontmatter was off. Body refs are already free of the cost concern that gated frontmatter — they surface in markdown the user explicitly typed — so they should always resolve.
3. `extract.ts` called `extractPageLinks` with one resolver doing both jobs. Splitting via `opts.skipFrontmatter` lets the body pass keep the real resolver while frontmatter stays opt-in.

## What this PR is

This is the **wikilink-resolver portion of the original PR #768** (which bundled #767 + #769 + extract polish). Two pieces of that bundle have either been absorbed or split off:

- The **#767 sync-strategy first-sync fix** has been absorbed by v0.31.2's `collectSyncableFiles` independently.
- The **doctor hint fix** (`Run: gbrain extract all` instead of the gone-since-v0.16 `gbrain link-extract && gbrain timeline-extract`) has been absorbed by upstream master at `doctor.ts:2503`.
- The **#769 chunk-metadata fix** is in a separate sibling PR.

This PR carries **only** the wikilink resolver — no scope overlap with upstream master.

## Fixes

**`link-extraction.ts`** adds `BARE_WIKILINK_RE` matching `[[<name>(#anchor)?(|display)?]]` shapes outside `DIR_PATTERN`, resolved via the new `resolveBareWikilink(name, resolver)` that walks fuzzy match → bare-name prefix expansion → exact-slug before giving up. Three new exports: `BARE_WIKILINK_RE`, `resolveBareWikilink`, `isBareName` (regex shape guard for the pre-extract candidate check). `extractPageLinks` gains an `opts.skipFrontmatter` parameter — when true, the frontmatter pass is skipped but body wikilinks still resolve through the passed resolver.

**`extract.ts`** threads the always-on `resolver` (not the conditional `nullResolver`) into `extractPageLinks` for the body pass, with `opts.skipFrontmatter` wired off `--include-frontmatter`.

## Tests

`test/link-extraction.test.ts`: 75 lines covering `BARE_WIKILINK_RE` shape (anchor + display variants), `resolveBareWikilink` fuzzy + prefix + exact paths, `isBareName` negative cases (`DIR_PATTERN` prefixes still rejected), and `extractPageLinks` integration with `opts.skipFrontmatter` under both modes.

Local: `bun run verify` clean, `bun test test/link-extraction.test.ts` → 103 pass / 0 fail.

## Scope note

The FS-source path (`extractLinksFromDir`) is NOT updated. It uses a different codepath via `extractMarkdownLinks` + `resolveSlug`; bare-name wikilinks in FS mode still won't resolve. Most users are on `--source db` (autopilot uses it); FS is for offline Obsidian-vault mode. Separate concern.

## Test plan

- [x] `bun run verify` clean
- [x] `bun test test/link-extraction.test.ts` → 103/0/0
- [ ] `bun run test:e2e` (gated on DATABASE_URL)
- [ ] Manual verification on a real wiki corpus — `gbrain extract links` produces non-zero link counts on pages using `[[Bare-Name]]` shapes

### DIFF
--- a/src/commands/extract.ts
+++ b/src/commands/extract.ts
@@ -1066,6 +1066,13 @@ async function extractLinksFromDB(
   // N-thousand API call trap on 46K-page brains. Resolver has a per-run
   // cache so duplicate names (same person appearing on many pages) resolve
   // once, not once per mention.
+  //
+  // Used for BOTH body wikilink resolution (always-on now — see #769 follow-up:
+  // generic `[[bare-name]]` refs need fuzzy matching against canonical slugs)
+  // and frontmatter resolution (gated by --include-frontmatter via skipFrontmatter
+  // opt below). Pre-fix nullResolver was used to suppress frontmatter; that's now
+  // moved into extractPageLinks's opts.skipFrontmatter so body refs always have a
+  // working resolver.
   const resolver = makeResolver(engine, { mode: 'batch' });
   const unresolved: UnresolvedFrontmatterRef[] = [];
   const nullResolver = {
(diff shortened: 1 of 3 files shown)


rayers:
Superseded by #1388 (merged), which is upstream's adoption of this PR's kernel — the generic wikilink regex + slug-tail index pattern — reworked as opt-in via `link_resolution.global_basename`, with multi-match resolution and FS-source coverage. Thanks for taking it forward; closing this as done-a-different-way.

PR #1240, opened 2026-05-20
ai/dims: forward `dimensions` for Qwen3-Embedding on openai-compat

mike7seven:
## Problem

The `ollama` recipe declares an embedding touchpoint, but `dimsProviderOptions` (`src/core/ai/dims.ts`) has no branch matching Qwen3-Embedding model ids. On the `openai-compatible` path it falls through to `return undefined`, so the adapter sends no `dimensions` field.

Qwen3-Embedding is a Matryoshka model with a large native dim (1024 for 0.6B, 2560 for 4B, **4096 for 8B**). Without `dimensions`, the endpoint returns the full native vector, and every embed then fails the dim-consistency check:

```
Error embedding <page>: Embedding dim mismatch:
  model qwen3-embedding:8b-q4_K_M returned 4096 but schema expects 1024.
```

This bites any brain sized to 1024 dims — the natural choice to stay under pgvector HNSW's 2000-dim index cap — when pointed at `ollama:qwen3-embedding:*`. The misconfiguration is silent until first embed: `gbrain config`, `providers list`, and `doctor` all look healthy.

## Fix

Add a `qwen3-embedding*` branch to the `openai-compatible` case, alongside the existing DashScope `text-embedding-v3` / Zhipu `embedding-3` Matryoshka branches:

```ts
if (modelId.startsWith('qwen3-embedding')) {
  return { openaiCompatible: { dimensions: dims } };
}
```

`startsWith` matches both the bare id (`qwen3-embedding-8b`) and the Ollama tag-suffixed form (`qwen3-embedding:8b-q4_K_M`, after the provider prefix is stripped). Ollama's `/v1/embeddings` honors `dimensions` (Matryoshka truncation + renormalize) — verified against ollama 0.24.0. Qwen3-Embedding is symmetric, so `inputType` is ignored.

## Tests

New `test/ai/dims-ollama.test.ts` (4 cases) pins: dimensions forwarded for bare + tag-suffixed ids, a non-default width honored, no `input_type` leak. New + existing `dims-*` / `gateway` tests pass; `tsc --noEmit` and `check-privacy.sh` clean.

### DIFF
--- a/src/core/ai/dims.ts
+++ b/src/core/ai/dims.ts
@@ -220,6 +220,17 @@ export function dimsProviderOptions(
       if (modelId === 'text-embedding-v3' || modelId === 'embedding-3') {
         return { openaiCompatible: { dimensions: dims } };
       }
+      // Qwen3-Embedding (commonly served via Ollama / llama.cpp on the
+      // openai-compatible adapter) is a Matryoshka model with a large native
+      // dim — 1024 for 0.6B, 2560 for 4B, 4096 for 8B. Brains configured for
+      // a narrower width — e.g. 1024 to fit pgvector HNSW's 2000-dim index
+      // cap — must forward `dimensions`, otherwise the endpoint returns the
+      // full native vector and the embed fails the dim-consistency check.
+      // Ollama's /v1/embeddings honors `dimensions` (MRL truncation +
+      // renormalize). Symmetric retrieval — inputType ignored.
+      if (modelId.startsWith('qwen3-embedding')) {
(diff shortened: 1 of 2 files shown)


time-attack:
Thanks for the careful writeup and the Ollama 0.24 verification — the diagnosis was exactly right. This has since landed on master via #1072 (commit 02ba4b4f), which adds the same `qwen3-embedding` / `qwen3-embedding:<tag>` branch to the openai-compatible case in `src/core/ai/dims.ts`. Closing as already fixed; no action needed on your side.

PR #1213, opened 2026-05-20
fix(skillpack): align harvest routing eval with triggers

lost9999:
## Summary
- Rewrite skillpack-harvest routing eval intents so they contain the declared trigger substrings.
- Fixes resolver_health ROUTING_MISS warnings for skillpack-harvest.

## Verification
- 
GBrain Health Check
===================
  [OK] resolver_health: 43 skills, all reachable
  [OK] skill_conformance: 43/43 skills pass
  [OK] sync_failures: 1291 historical sync failure(s), all acknowledged [UNKNOWN=1291].
  [OK] home_dir_in_worktree: gbrain home is outside any enclosing git worktree.
  [WARN] connection: Skipping DB checks (--fast mode, URL present from env:GBRAIN_DATABASE_URL)

Health score: 95/100. All checks OK (some warnings). → resolver_health OK, skill_conformance OK.

## Note
- A broader v0.37 skillpack test subset currently has unrelated failures around reference-pack 10/10 fixture scoring 7/10; this PR only touches the skillpack-harvest routing-eval fixture.

PR #1257, opened 2026-05-21
fix(skillpack): align harvest routing eval with triggers

lost9999:
## Summary
- Align `skillpack-harvest` routing eval intents with the declared resolver triggers.
- Keeps the user phrasing realistic while ensuring every fixture contains a trigger substring the resolver can match.

## Why
`gbrain doctor --fast` currently reports 7 `ROUTING_MISS: skillpack-harvest` warnings on `origin/master` because the eval fixtures use adjacent wording that does not hit the trigger-substring resolver.

## Validation
- `GBRAIN_HOME=/mnt/g/gbrain-runtime gbrain doctor --fast`
- Before: `resolver_health: 7 issue(s): 0 error(s), 7 warning(s)`
- After: `[OK] resolver_health: 43 skills, all reachable`
- Health score improved from `90/100` to `95/100` in the same environment.


garrytan:
Closed in favor of the structural fix in #1458. Same call as #1213 — under the new frontmatter-authoritative contract, the fix moved into broadening `skills/skillpack-harvest/SKILL.md` triggers rather than rewriting fixtures. Thanks @lost9999.

PR #1263, opened 2026-05-21
v0.37.6.0 fix: dynamic embedding column write target (closes #1262)

### LINKED ISSUE #1262: PostgresEngine.upsertChunks ignores embedding_columns registry — writes 2560-dim embeddings into legacy vector(1536), fails with dimension mismatch
## Summary

`PostgresEngine.upsertChunks` in `src/core/postgres-engine.ts` (lines 1577-1678 on `master`, lines 1396-1497 on v0.36.6.0) hardcodes the literal `embedding` column in both the `INSERT` column list and the `ON CONFLICT` update branch. The `$N::vector` cast is also hardcoded. There is no read of the `embedding_columns` registry that PR #1164 introduced for the read side.

Effect: a brain that operates a multi-column registry (e.g. `embedding/vector(1536)` + `embedding_zeroentropy/halfvec(2560)`) with `embedding_model: zeroentropyai:zembed-1` will fail every `put_page` / `gbrain embed` / `gbrain import` / `gbrain sync` write with:

```
expected 1536 dimensions, not 2560
```

…because the embed gateway produces a 2560-dim vector and `upsertChunks` insists on writing it into the legacy 1536-dim column.

The fix on the read side shipped in PR #1164 ("feat: dynamic embedding column selection for search"). The write side is the symmetric gap.

## Reproduction

Postgres backend, gbrain v0.36.6.0 (verified to repro identically on master HEAD `1580c6d1c` / VERSION `0.37.5.0` — `diff` of the `upsertChunks` region across the two versions shows no behavioural changes).

1. Have a brain with an `embedding_columns` registry that advertises both legacy `embedding` and a non-1536 column, e.g.:
   ```sql
   SELECT key, value FROM config WHERE key='embedding_columns';
   -- {"embedding":{"provider":"google:gemini-embedding-001","dimensions":1536,"type":"vector"},
   --  "embedding_zeroentropy":{"provider":"zeroentropyai:zembed-1","dimensions":2560,"type":"halfvec"}}
   ```
   and a `search_embedding_column` pointing at the new column:
   ```sql
   SELECT key, value FROM config WHERE key='search_embedding_column';
   -- embedding_zeroentropy
   ```
2. Switch the file config to the matching model:
   ```json
   {
     "engine": "postgres",
     "database_url": "...",
     "embedding_dimensions": 2560
   }
   ```
   `ZEROENTROPY_API_KEY` set, `gbrain doctor` reports `embedding_column_registry: Registry healthy: ... active='embedding_zeroentropy'`, `embeddings: 100% coverage`.
3. Call `put_page` over MCP (or run `gbrain embed --slug some/page`).
4. Server responds:
   ```json
   {"error":"internal_error","message":"expected 1536 dimensions, not 2560"}
   ```

## Root cause

`src/core/postgres-engine.ts` (master):

- The "cols" string at the top of the row-builder loop is a literal that names `embedding` — no branch on the registry.
- The placeholder for the embedding param is built as `\`\$\${paramIdx++}::vector\`` — `::vector` is hardcoded; nothing routes to `::halfvec` for ZE.
- The `ON CONFLICT` update branch references `embedding` three more times (the value, the COALESCE fallback, the embedded_at reset guard).

No call site (`commands/embed.ts`, `commands/sync.ts`, `core/import-file.ts:351/650/782`, `commands/reindex-multimodal.ts`) supplies an alternate column. `reindex-multimodal.ts` writes only `embedding_multimodal` via a different statement. The `embedding_columns` registry from PR #1164 is consumed only by `src/core/search/embedding-column.ts` and `core/search/hybrid.ts` — read side only.

`backfill-registry.ts` marks `embedding_voyage` as a no-op backfill, and `migrate.ts` only auto-adds `embedding_multimodal` on existing brains — no upstream write code path fills `embedding_zeroentropy` or any user-added alt column.

A GitHub code search for `embedding_zeroentropy` across `garrytan/gbrain` returns only 2 matches: a docstring on `operations.ts:1159` and a unit test in `test/search/embedding-column.test.ts`. Production write path: 0 references.

## Expected behavior

`PostgresEngine.upsertChunks` should resolve a write target column from the `embedding_columns` registry — preferring the entry whose `provider` matches the currently-configured `getEmbeddingModel()` — and use that column + its declared `type` (`vector` | `halfvec`) when emitting the INSERT and ON CONFLICT branch. On any miss it should fall back to legacy `embedding vector(1536)` so pre-registry brains keep working unchanged.

## Suggested fix shape

We're currently running a local patch on our deployment that does this resolution **inside** `PostgresEngine.upsertChunks` by reading the `config` table directly. It works (write lands in `embedding_zeroentropy`, legacy column stays NULL, all tests + our operational golden gate pass) — but we suspect the upstream-preferred shape is to thread a resolved write descriptor through `BrainEngine.upsertChunks` from a higher layer, consistent with how PR #1164 keeps the engine 'config-free' and resolves columns at the `hybridSearch` boundary.

We'd be happy to contribute a PR. Before we send one, would you prefer:

- (a) **Engine-internal resolution** — DB read inside `PostgresEngine.upsertChunks` (the simplest delta — only `postgres-engine.ts` changes, plus `BrainEngine` interface stays as-is).
- (b) **Caller-resolved descriptor** — add an optional `writeColumn?: ResolvedColumn` (or `embeddingColumn` mirror of `SearchOpts`) to the `BrainEngine.upsertChunks` signature, resolve it in `importFromContent` / `embedPage` from the merged config + currently-configured model, and pass it down. Larger diff but symmetric with the read-side architecture.

Happy to implement either — let me know.

## Environment

- gbrain: v0.36.6.0 (production install)
- master verified: HEAD `1580c6d1c`, VERSION `0.37.5.0`
- Backend: Postgres (Supabase pooler)
- pgvector / halfvec installed
- 696 pages / 6605 chunks; 100% ZE coverage from a side-backfill (not gbrain CLI)

Thanks!

DmitryBMsk:
## Summary

Fix dynamic embedding writes for multi-column brains.

Before this PR, `upsertChunks()` always wrote text vectors into the legacy `content_chunks.embedding vector(1536)` column. Brains using a configured alternate column such as `embedding_ze halfvec(2560)` could fail on `put_page`, `import`, `sync`, or `embed` with a dimension mismatch.

This PR:

- resolves the write target once from merged config + current embedding model
- passes that resolved column into `upsertChunks()`
- uses the same resolved column for `embed --stale` count/list scans, so alternate-column rows are not re-embedded forever
- keeps legacy brains on the existing `embedding` column

Closes #1262.

## Tests

- `bun run typecheck`
- `bun test test/search/embedding-column.test.ts`
- `bun test test/e2e/embedding-column-pglite.test.ts`
- `bun test test/e2e/embedding-column-postgres.test.ts` (skips when `DATABASE_URL` is unset)
- `bun test test/embed.serial.test.ts`
- `bun test test/pglite-engine.test.ts --test-name-pattern "countStaleChunks|listStaleChunks"`


time-attack:
Thank you — the diagnosis in #1262 is correct and the write path is still hardcoded to the legacy `embedding` column on master. We're taking #3170 instead, which resolves the same root cause across more call sites (sync, embed-stale, the embed-backfill handler, contextual retrieval) and is cut against the current release line; this PR also carries a stale v0.37.6.0 version bump that no longer applies. Closing as superseded by #3170.

PR #1279, opened 2026-05-21
feat: add ~/.gbrain/recipes/ convention path + --external-dir flag for recipe discovery

ialmeida-jera:
## Summary

Adds two additive entries to the recipe-discovery tier list in `getRecipeDirs()` (in `src/commands/integrations.ts`):

1. **Convention path:** `~/.gbrain/recipes/` — auto-discovered if the directory exists. Drop a `.md` recipe there and it shows up in `gbrain integrations list` with no env-var or cwd gymnastics.
2. **One-shot flag:** `--external-dir <path>` — works on `list`, `show`, `status`, `doctor`, `stats`. Useful for testing a recipe before installing it.

Both new tiers load with `trusted: false`, so the existing B1 trust boundary (`command:` / `http:` / string health_checks blocked for non-embedded recipes via `executeHealthCheck()`) flows through unchanged.

## Why

`gbrain integrations list` currently has four discovery paths (lines 332-344 of `src/commands/integrations.ts`):

1. `../../recipes` (source / dev install) — trusted
2. `~/.bun/install/global/node_modules/gbrain/recipes` (global install) — trusted
3. `$GBRAIN_RECIPES_DIR` — untrusted
4. `./recipes` (cwd) — untrusted

Of those, only (3) and (4) accept user-authored recipes, and both have rough ergonomics:

- `GBRAIN_RECIPES_DIR` needs an export in every shell session (or a systemd `Environment=` line, etc.). Easy to forget; invisible to anyone who didn't set it up.
- `./recipes` only triggers when the caller happens to be `cd`-ed into the parent of a `recipes/` dir. Brittle in cron and MCP contexts where cwd is `/` or `$HOME`.

The net effect: teams authoring recipes outside the upstream `recipes/` dir (org-internal integrations, in-house infra) end up writing wrappers around `gbrain` that re-export the env var, or copying recipes into the package install (gets blown away on upgrade). We hit this at Jera Capital with 4 internal recipes deployed to an agent VM. Any team running gbrain in production with custom integrations probably hits the same friction.

This PR is the minimum-viable affordance for that use case.

## What this PR does (and doesn't) change

The new tier ordering (first-occurrence wins on filename collision, trusted tiers always shadow untrusted):

| # | Path | Trusted | Status |
|---|------|---------|--------|
| 1 | `../../recipes` (source) | yes | unchanged |
| 2 | `~/.bun/.../gbrain/recipes` (global) | yes | unchanged |
| 3 | `$GBRAIN_RECIPES_DIR` | no | unchanged |
| 4 | `--external-dir <path>` | no | **new** |
| 5 | `~/.gbrain/recipes/` | no | **new** |
| 6 | `./recipes` (cwd) | no | unchanged |

**Strictly additive.** With no `~/.gbrain/recipes/` directory present and no `--external-dir` flag, behavior is byte-identical to today. The `getRecipeDirs()` signature gains an optional parameter (`externalDir?: string`) so existing callers (including the test suite) work without change.

**Trust boundary preserved.** Both new tiers are `trusted: false`. Anything loaded via them gets `embedded: false`, and `executeHealthCheck()` continues to hard-block `command:`, `http:`, and string health checks for non-embedded recipes (per the B1 tests at `test/integrations.test.ts:449-509` and `:624-652`, which still pass).

**Why `~/.gbrain/recipes/` and not `~/.gbrain/integrations/`:** `~/.gbrain/integrations/<id>/` is already populated with per-id `heartbeat.jsonl` subdirs. Mixing recipe `.md` files into that tree blurs the data/config separation. A sibling `~/.gbrain/recipes/` keeps the mental model clean.

## Files changed

- `src/commands/integrations.ts` — `getRecipeDirs()` gains optional `externalDir?` param; new `parseExternalDir()` helper plucks the flag (supports both `--external-dir <path>` and `--external-dir=<path>`); all five subcommands (`cmdList`, `cmdShow`, `cmdStatus`, `cmdDoctor`, `cmdStats`) thread the flag through. `printHelp()` updated to document the new discovery tiers and the flag. The tier-list comment block above `getRecipeDirs()` updated to reflect the new ordering.
- `test/integrations.test.ts` — two new `describe` blocks:
  - `getRecipeDirs --external-dir parameter` — 4 tests covering: included when path exists with `trusted: false`; omitted when path does not exist; backwards-compat when no arg passed; B1 regression — external-dir tier is never `trusted: true` even if the recipe markdown is malicious.
  - `getRecipeDirs convention path (~/.gbrain/recipes)` — 1 test asserting that, if the dir appears in the returned list, it is always `trusted: false`. (Tests don't mutate `$HOME` to avoid polluting the user's brain at test time; the invariant is what we care about.)

Net diff: +159 / -23 across 2 files.

## Test plan

- [x] `bun test test/integrations.test.ts` — 115 pass / 0 fail (was 108 pass / 0 fail before this PR; 7 new tests added).
- [x] `bun run typecheck` — clean (existing callers compile against the new optional-param signature without change).
- [ ] Maintainer to run `bun test` and `bun run verify` on their environment.
- [ ] Smoke: `gbrain integrations list` with no `~/.gbrain/recipes/` and no flag — output identical to current master.
- [ ] Smoke: `mkdir -p ~/.gbrain/recipes && cp recipes/email-to-brain.md ~/.gbrain/recipes/foo.md && gbrain integrations list` — new recipe appears with `embedded: false` (verify via `--json`).
- [ ] Smoke: `gbrain integrations list --external-dir /tmp/somerecipes` — recipes from `/tmp/somerecipes` appear; `gbrain integrations doctor --external-dir /tmp/somerecipes` blocks any `command:` / `http:` checks they declare.

## Open questions

1. **Naming.** `~/.gbrain/recipes/` (this PR) vs `~/.gbrain/integrations/` vs XDG (`~/.config/gbrain/recipes/`)? Happy to defer to maintainer preference.
2. **Flag repeatability.** Single `--external-dir` only in v1 (keeps the diff and test surface small). Trivial to add multi-shot later if there's demand.
3. **Config knob in `gbrain.yml`?** Out of scope here — `$GBRAIN_RECIPES_DIR` already covers the override case.
4. **Future: `gbrain integrations install <url>`?** Would clone/fetch external recipes into `~/.gbrain/recipes/`. Mentioned for completeness; not part of this PR.
5. **Trust escalation knob?** Explicit NO from this draft — the B1 boundary exists for a reason. Saying it out loud so reviewers don't have to ask.

## Context

First contribution from Jera Capital. We hit this friction running gbrain in production with org-internal recipes; investigation notes (architecture, alternatives considered, why this shape) are available on request. The TL;DR of the alternatives we rejected: Bun has no entry_points equivalent, symlinks don't avoid the code change, a manifest file is more state for users to maintain, and pulling org-specific recipes into upstream `recipes/` doesn't generalize.

### DIFF
--- a/src/commands/integrations.ts
+++ b/src/commands/integrations.ts
@@ -320,16 +320,20 @@ export function parseRecipe(content: string, filename: string): ParsedRecipe | n
 
 // --- Embedded Recipes ---
 
-// Recipes are loaded from multiple tiers with an explicit trust boundary:
+// Recipes are loaded from multiple tiers with an explicit trust boundary
+// (first-occurrence-wins on filename collision, so TRUSTED always shadows UNTRUSTED):
 //   TRUSTED (embedded=true):  package-bundled recipes shipped with gbrain
-//     - source install: ../../recipes relative to this file
-//     - global install: ~/.bun/install/global/node_modules/gbrain/recipes
+//     1. source install: ../../recipes relative to this file
+//     2. global install: ~/.bun/install/global/node_modules/gbrain/recipes
 //   UNTRUSTED (embedded=false): user-provided recipes discovered at runtime
-//     - $GBRAIN_RECIPES_DIR
-//     - ./recipes in process cwd
-// The trust flag gates command/http health_checks and deprecated string health_checks.
-// An attacker who drops a malicious recipe in ./recipes/ MUST NOT get embedded=true.
-export function getRecipeDirs(): Array<{ dir: string; trusted: boolean }> {
+//     3. $GBRAIN_RECIPES_DIR (env-var override)
+//     4. --external-dir <path> (CLI flag, one-shot)
+//     5. ~/.gbrain/recipes (convention path)
+//     6. ./recipes in process cwd
+// The trust flag gates command/http health_checks and deprecated string health_checks
+// via executeHealthCheck(). An attacker who drops a malicious recipe in any UNTRUSTED
+// tier MUST NOT get embedded=true.
+export function getRecipeDirs(externalDir?: string): Array<{ dir: string; trusted: boolean }> {
   const dirs: Array<{ dir: string; trusted: boolean }> = [];
   const sourceDir = join(import.meta.dir, '../../recipes');
   if (existsSync(sourceDir)) dirs.push({ dir: sourceDir, trusted: true });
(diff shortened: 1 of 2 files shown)


time-attack:

PR #1280, opened 2026-05-21
feat(jobs): --poll-interval flag for `jobs work` subcommand

dcarolan1:
## Summary

Adds a `--poll-interval <ms>` flag to `gbrain jobs work` for tuning the worker's queue-empty sleep duration. Defaults to 5000ms (existing constructor default — purely additive, backward compatible). Bounded `[100, 300000]` with descriptive error messages on invalid input.

## Motivation

`MinionWorker.pollInterval` has existed as a constructor option since the initial worker landed, but had no CLI surface. For low-throughput setups (single-user agent fleet with sparse job traffic), the 5s default is the dominant pickup-latency contributor for small jobs.

Empirical measurement on a 3-host fleet during the last day:
- 3 runs of `gbrain jobs submit subagent {...}` → terminal observation
- Mean worker pickup ~3.6s, worst-case ~5.4s (when the poll cycle had just started — full 5s wait)

Without this flag, operators have to fork the CLI entrypoint to construct `MinionWorker` with a custom `pollInterval`. This PR closes that gap with the same shape as the existing `--health-interval` and `--max-rss` flags.

## Behavior

- New flag: `--poll-interval <ms>` on `gbrain jobs work`
- Default: `5000` (matches existing `MinionWorker` constructor default — no behavior change for existing service units)
- Bounds: `[100, 300000]` ms
  - `100ms` minimum prevents accidental tight-loop CPU burn
  - `5min` maximum is "effectively no polling"; longer intervals likely want a different scheduling pattern entirely
- Validator extracted as exported `parsePollIntervalFlag(args)` following the existing `parseMaxRssFlag` pattern
- Plumbed through to `MinionWorker` constructor option

## Testing

New file `test/jobs-work-poll-interval.test.ts` — 8 test cases, all pass:

| # | Case | Verifies |
|---|------|----------|
| 1 | `parsePollIntervalFlag([])` | undefined when flag absent |
| 2 | Bounds-edge values (100, 5000, 300000) | accepted |
| 3 | Mid-range values (250, 1000, 60000) | accepted |
| 4 | CLI: `--poll-interval 50` | exits 1 with bounds error |
| 5 | CLI: `--poll-interval 0` | exits 1 with bounds error |
| 6 | CLI: `--poll-interval 300001` | exits 1 with bounds error |
| 7 | CLI: `--poll-interval abc` | exits 1 with bounds error |
| 8 | CLI: `--poll-interval 5s` | exits 1 with bounds error (parseInt yields 5, < 100) |

[...]

PR #1282, opened 2026-05-21
feat(subagent): honor subagent_def at runtime — load def.body + def.allowed_tools

dcarolan1:
## Summary

Wires `data.subagent_def` to be load-bearing at job-dispatch time: when set AND the named def exists in the plugin registry, `def.body` becomes the default system prompt and `def.allowed_tools` becomes the default allowed tool set. Both overridable by explicit `data.system` / `data.allowed_tools`. Backward compatible.

## Motivation

The plugin-loader at `src/core/minions/plugin-loader.ts` does substantive work at worker startup:
- Discovers `gbrain.plugin.json` manifests under each `GBRAIN_PLUGIN_PATH` entry
- Validates each subagent def's `allowed_tools` is a subset of the brain-allowlist (fails loudly on typos)
- Resolves left-wins collisions across multiple plugins
- Logs `[plugin-loader] loaded 'plugin-name' v1.0.0 (N subagents)` per discovered plugin

But the loaded `SubagentDefinition[]` data was **never consumed at runtime**: the subagent handler (`src/core/minions/handlers/subagent.ts:170`) reads `data.system ?? DEFAULT_SYSTEM` and `data.allowed_tools ?? registry` directly from job data, with no fallback to whatever the named def specified. The `data.subagent_def` field was set + persisted but never read.

Practical impact: callers had to re-embed the entire system body + allowed_tools list in every `queue.add('subagent', { ... })` submission to get the def's behavior. For a 4,583-byte subagent body, that's ~4.5 KB per job in `minion_jobs.data` plus the per-call complexity on the calling side.

Discovered while building a BAG-side subagent (`aaron-nl-to-shell-ctx`) under an internal stay-close-to-gbrain doctrine — the caller code in our TG dispatcher (Moses) was re-embedding the full subagent body on every submission, defeating the plugin system's stated purpose of letting operators declare reusable subagent contracts.

## Behavior

When `data.subagent_def` is set AND a matching def exists in the plugin registry (populated at worker startup by `loadPluginsFromEnv`):

| Field | Resolution order |
|-------|------------------|
| `systemPrompt` | `data.system` → `def.body` (when non-empty) → `DEFAULT_SYSTEM` |
| `allowed_tools` | `data.allowed_tools` → `def.allowed_tools` → full registry |

Override semantics: **data fields ALWAYS win when explicitly provided.** Callers retain full control; the def is a convenience default.

Fail-open behavior (handler never throws on these):
- `data.subagent_def` set but def not in registry (e.g., worker has a stale plugin path while caller has a newer one) → warn to stderr + fall through to current behavior
- `def.body` empty or whitespace-only → treat as no-body, fall through to `DEFAULT_SYSTEM`
- `data.subagent_def` absent → behavior unchanged (backward compatible)

## Required infrastructure: register the registry

To enable lookup, `loadPluginsFromEnv` now ALSO populates a module-level `_subagentRegistry: Map<string, SubagentDefinition>` alongside its existing log output. Lookup-only public API:

```typescript
export function getSubagentDef(name: string): SubagentDefinition | undefined
```

Registry lives for the worker process lifetime. Left-wins collision policy (already enforced by plugin-loader's existing `subagentByName` Map) is honored on registration.

## Testing

New file `test/subagent-def-runtime-contract.test.ts` — **7 test cases, all PASS in 2.72s**:

| # | Case |
|---|------|
| 1 | `def.body` becomes system prompt when subagent_def set + no data.system override |
| 2 | `data.system` overrides def.body (data wins) |
| 3 | `def.allowed_tools` used when data.allowed_tools absent |
| 4 | `data.allowed_tools` fully overrides def.allowed_tools (NOT intersection — data wins fully) |
| 5 | subagent_def names unknown def → warn + fall through to DEFAULT_SYSTEM + full registry (fail-open) |
| 6 | subagent_def absent → existing behavior unchanged (backward compat) |
| 7 | def.body empty/whitespace-only → treated as no-body, falls to DEFAULT_SYSTEM |

Tests use the same pattern as existing `test/subagent-handler.test.ts` (PGLite in-memory + FakeMessagesClient capturing `params.system` and `params.tools` from the Messages API call). Registry seeding via a new `__testing._seedSubagentDef(def)` + `__testing._clearSubagentRegistry()` surface so tests don't need on-disk plugin fixtures.

`bun run test`: **7967 pass / 15 fail** — 15 failures are the same pre-existing baseline (skillpack-check, check-resolvable, BrainRegistry, ConnectionManager, apply-migrations). Zero failures in plugin-loader / subagent / handler paths.

## Backward Compatibility

Purely additive surface. Existing callers without `data.subagent_def` see no behavior change. The `MinionJobContext` type, the `SubagentHandlerData` shape, and the `SubagentDefinition` shape all unchanged — only the handler's resolution logic + the addition of the registry export.

## Out of Scope

- **Plugin hot-reload.** Registry lives for the worker process lifetime; removed plugins linger until restart. Acceptable for v1; hot-reload deserves its own design discussion.
- **Per-job def versioning.** A future enhancement could pin `data.subagent_def_version`; this PR doesn't add it.
- **Intersection semantics on allowed_tools.** This PR chose "data wins fully" over "intersect" — happy to flip if you prefer; let me know.

## Notes

This is the second BAG → gbrain contribution under our internal stay-close-to-gbrain doctrine. First PR (queued separately) added `--poll-interval` flag to `gbrain jobs work`. Both came out of measurement work for a multi-host agent fleet that uses gbrain's MinionQueue + subagent + plugin systems heavily.

Happy to adjust per repo conventions — the `Co-Authored-By: Claude` trailer is on the commit but easily droppable, intersection-vs-override semantics easily swappable, CHANGELOG entry shape easily adjustable.

🤖 Generated with [Claude Code](https://claude.com/claude-code) — happy to adjust attribution as preferred.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,37 @@
 
 All notable changes to GBrain will be documented in this file.
 
+## [Unreleased]
+
+### Changed
+
+- Subagent handler honors `data.subagent_def` at runtime. When set AND the
+  named def exists in the plugin registry (populated at worker startup by
+  `loadPluginsFromEnv`):
+  - `def.body` is used as the default system prompt — overridable by
+    explicit `data.system`
+  - `def.allowed_tools` is used as the default allowed tool set — overridable
+    by explicit `data.allowed_tools` (data wins fully when set; def is the
+    convenience default)
+  
+  Closes the doctrine-vs-implementation gap where the plugin-loader validated
+  defs at startup (catching typos in `allowed_tools` etc.) but the handler
+  ignored the loaded def at job-dispatch time. Callers previously had to
+  embed the full system body and tool list in every job submission for the
+  named def to take effect.
+  
+  Backward compatible: when `data.subagent_def` is absent, behavior is
+  unchanged. When `data.subagent_def` names a def NOT in the loaded
+  registry (worker has a stale plugin path), the handler logs a warning
+  and falls through to current behavior (fail open, not closed). Empty or
+  whitespace-only `def.body` is treated as no-body and falls through to
+  `DEFAULT_SYSTEM`.
+  
+  Required to establish the registry: `loadPluginsFromEnv` in
+  `plugin-loader.ts` now populates a module-level `_subagentRegistry` Map
+  alongside its existing log-only behavior; lookup-only public access via
+  exported `getSubagentDef(name)`.
+
 ## [0.37.9.0] - 2026-05-20
 
 **Tags get written the same way everywhere now.**
--- a/src/core/minions/handlers/subagent.ts
+++ b/src/core/minions/handlers/subagent.ts
@@ -48,6 +48,7 @@ import {
   logSubagentHeartbeat,
 } from './subagent-audit.ts';
 import { resolveModel, isAnthropicProvider, TIER_DEFAULTS } from '../../model-config.ts';
+import { getSubagentDef } from '../plugin-loader.ts';
 
 // ── Defaults ────────────────────────────────────────────────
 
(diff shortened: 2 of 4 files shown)


dcarolan1:
Hey Garry — quick check-in on this one, no rush on response.

We've been running the subagent_def runtime fix from this PR in production on the BAG fleet for ~48 hours. Branch C escalation across hosts works end-to-end through real interactive Telegram smoke — natural-language queries like "what's the uptime on phineas?" correctly route through the subagent path, hit the runtime contract validation, and dispatch shell commands to the right compute host. The fix is doing exactly what we needed.

Two questions when you have a moment:

1. **Review queue position** — is there anything you'd like changed on the PR before merge, or is it queued for review? Happy to rebase or split if the diff is too broad.

2. **Upgrade path for downstream installs carrying this PR** — we're contemplating an upgrade from our current state (v0.37.9.0 base + this PR branch on Aaron, v0.36.6.0 on Phineas) toward upstream master (now v0.40.6.0). Two paths look reasonable from our side:
   - **Wait** for merge, then upgrade cleanly with the fix already in master
   - **Rebase the PR onto master ourselves** and update this branch, then upgrade to master locally
   
   What would you recommend? We don't want to fork-evolve away from upstream, and we also don't want to sit on stale local commits if we should be moving.

Also pinged on #1280 separately. Both BAG-internal use cases are real and running; just want to make sure we're not blocking ourselves by drifting from main.

Thanks!

time-attack:
Thanks for the detailed resolution table and the fail-open analysis. We're declining this one: nothing is broken today (callers pass `system`/`allowed_tools` explicitly and get exactly what they asked for), and this makes the effective tool allowlist depend implicitly on whichever plugin registry the worker happened to load at startup — a resolution path we don't want on the allowlist decision. Payload size on the calling side isn't enough to justify that trade.

PR #1, opened 2026-04-05
feat: GBrain v0.1.0 — Postgres-native personal knowledge brain

garrytan:
## Summary

GBrain v0.1.0: a personal knowledge brain backed by Postgres + pgvector with world-class hybrid search.

**Foundation**
- Pluggable `BrainEngine` interface with full `PostgresEngine` implementation
- Postgres schema with pgvector HNSW, trigger-based tsvector (spans timeline_entries), pg_trgm fuzzy slug matching
- Markdown parser with frontmatter, compiled_truth/timeline splitting, round-trip serialization

**Search**
- 3-tier chunking: recursive (delimiter-aware), semantic (Savitzky-Golay), LLM-guided (Claude Haiku)
- Hybrid search: vector + keyword merged via Reciprocal Rank Fusion
- Multi-query expansion via Claude Haiku (2 alternative phrasings)
- 4-layer dedup: by source, cosine similarity >0.85, type diversity 60% cap, per-page cap

**CLI + MCP**
- 25+ CLI commands covering CRUD, search, import/export, links, tags, timeline, versions, health, config
- MCP stdio server with 20 tools mirroring all CLI operations
- Smart slug resolution via pg_trgm for reads

**Skills + Docs**
- 6 fat markdown skills: ingest, query, maintain, enrich, briefing, migrate
- ClawHub manifest for skill distribution
- Full design docs: GBRAIN_V0 spec, ENGINES.md (pluggable architecture), SQLITE_ENGINE.md

**Security fixes from adversarial review**
- revertToVersion: added page_id check preventing cross-page data corruption
- traverseGraph: UNION instead of UNION ALL for cycle safety
- embedAll: preserve all chunks when embedding stale subset
- embedding: throw on retry exhaustion instead of returning zero vectors
- putPage: slug validation preventing path traversal on export

## Test Coverage
19 tests passing (markdown parsing + round-trip serialization, recursive chunker splitting + overlap + sizing)

## Pre-Landing Review
5 critical/high issues found and fixed from adversarial review. 0 remaining.

## Test plan
- [x] `bun test` passes (19 tests, 0 failures)
- [x] `gbrain --help` prints usage
- [x] `gbrain --tools-json` outputs tool discovery JSON

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1189, opened 2026-05-19
bug: PGLite apply-migrations wedges with zembed-1 2560d due HNSW index emitted on migrate path


jeades:
## Summary

Fresh PGLite brain initialized with ZeroEntropy `zembed-1` at 2560 dimensions can become wedged when `gbrain apply-migrations` / migration schema replay runs.

The DB itself remains structurally intact, but the migration ledger records repeated partials and doctor reports a wedged migration. The failure appears to be the migrate/schema replay path using default 1536-dim gateway state instead of the configured `embedding_dimensions: 2560`, causing schema replay to emit the HNSW chunk index even though pgvector HNSW caps at 2000 dims.

Related but not identical: #1141. That issue is Postgres/Supabase schema v66 -> v67. This report is the PGLite fresh-install/apply-migrations surface.

## Environment / config

Current upstream inspected while triaging:

- `master`: `3a0e1116e76b1be45a136cf184d1a49d431b6b60`
- `VERSION`: `0.36.1.0`
- Engine: PGLite
- Pages: `0`

Config shape:

```json
{
  "engine": "pglite",
  "database_path": "~/.gbrain/brain.pglite",
  "embedding_model": "zeroentropyai:zembed-1",
  "embedding_dimensions": 2560
}
```

## Observed

Doctor / state after failure:

- `schema_version` remained current / unchanged
- `connection: Connected`
- `pages: 0`
- `search_mode: balanced`
- config preserved
- migration ledger had repeated `partial` entries for `0.11.0`, later also `0.12.0`
- doctor reported `WEDGED MIGRATION(s)` after repeated partials

Representative underlying failure:

```text
column cannot have more than 2000 dimensions for hnsw index
```

## Expected

All schema replay / migration paths should apply the same HNSW policy as fresh schema generation:

- if embedding dimensions <= 2000: create `idx_chunks_embedding` HNSW index
- if embedding dimensions > 2000: skip HNSW index and rely on exact vector scan

No migration should wedge a 2560-dim brain by trying to create the chunk HNSW index.

## Suspected root

`init --migrate-only` loads config and calls `engine.initSchema()`, but does not appear to configure the AI gateway from config before schema generation.

PGLite `initSchema()` reads dimensions from the gateway and falls back to 1536 when the gateway is not configured. That makes `getPGLiteSchema()` apply the <=2000 HNSW branch even though the actual brain config/column is 2560.

[...]

PR #1293, opened 2026-05-22
fix(initSchema): introspect live vector column dim to defend HNSW-2000 cap (#1141, #1189)

### LINKED ISSUE #1141: bug: HNSW 2000-dim cap still hit on upgrade path despite #640 fix (zembed-1 2560d, schema v66 → v67)
## Summary

Follow-up to closed #640 — the schema-side fix (`applyChunkEmbeddingIndexPolicy` + `getPostgresSchema(dims, model)`) works for fresh installs, but **upgrading an existing brain still trips the same `column cannot have more than 2000 dimensions for hnsw index` error**, even when `~/.gbrain/config.json` correctly reports `embedding_dimensions: 2560`.

The fix didn't reach the upgrade path. Migration v67 (and presumably any future migration shipped on `master`) is gated behind a bootstrap stage that still emits the HNSW index.

## Repro

Pre-conditions:
- Brain on Postgres (Supabase Session pooler, port 5432).
- `content_chunks.embedding` was `ALTER COLUMN ... TYPE vector(2560)` for ZeroEntropy `zembed-1`.
- `~/.gbrain/config.json`:
  ```json
  {
    "engine": "postgres",
    "embedding_dimensions": 2560
  }
  ```
- `idx_chunks_embedding` HNSW dropped at switch-time (pgvector refused to keep HNSW on 2560-dim column — that part is well-understood).
- `idx_takes_embedding_hnsw` likewise dropped.
- Schema version: 66 (latest on v0.35.1.1).

Steps:
```bash
cd ~/gbrain
git pull origin master   # → v0.35.7.0 (1dadd9ed)
bun install              # postinstall runs `gbrain apply-migrations` (prints `All migrations up to date.` — misleading, see below)
gbrain init --migrate-only
```

Output (last 30 lines, after ~25 expected `42P07` NOTICEs):
```
{ severity: "NOTICE", code: "42P07",
  message: 'relation "idx_chunks_page" already exists, skipping', ... }
column cannot have more than 2000 dimensions for hnsw index
```
Exit code: 1. Migration v67 never runs.

## Expected

`getPostgresSchema(dims=2560, model='zembed-1')` should be invoked on the upgrade path (same way it's invoked on fresh-install). `applyChunkEmbeddingIndexPolicy` should replace the HNSW `CREATE INDEX` with the skip-comment, and `conn.unsafe(sqlText)` should pass cleanly.

## Hypothesis (for the maintainer to confirm / falsify)

Looking at `src/core/postgres-engine.ts` on `master`:

```ts
let dims = 1536;
let model = 'text-embedding-3-large';
try {
  const gw = await import('./ai/gateway.ts');
  dims = gw.getEmbeddingDimensions();
  model = gw.getEmbeddingModel().split(':').slice(1).join(':') || model;
} catch { /* gateway not yet configured — use defaults */ }

const sqlText = getPostgresSchema(dims, model);
// ...
await this.applyForwardReferenceBootstrap(conn);
await conn.unsafe(sqlText);
```

Two candidate roots:

**(a)** `gw.getEmbeddingDimensions()` returns the default `1536` instead of reading `embedding_dimensions: 2560` from `~/.gbrain/config.json`. The gateway likely sources from a separate config (env var? a different JSON file? a DB-resident setting?) that wasn't synced when the user switched embedders via direct SQL `ALTER COLUMN`.

**(b)** `applyForwardReferenceBootstrap` (which runs *before* `conn.unsafe(sqlText)`) emits an HNSW `CREATE INDEX` somewhere that the policy doesn't touch. Source-text scan against `master` doesn't show it in the bootstrap method, but the order-of-operations means **the error throws before `sqlText` even reaches the connection**, which is consistent with bootstrap being the culprit rather than the policy-applied SCHEMA_SQL.

The misleading `All migrations up to date.` line from the postinstall path is a separate UX bug worth fixing too — the `apply-migrations --yes --non-interactive` shell command appears to swallow the bootstrap failure silently, but `gbrain init --migrate-only` reproduces it deterministically.

## Suggested fixes

1. **Trace and patch the gap.** If hypothesis (a) is correct, make the gateway read `embedding_dimensions` from `~/.gbrain/config.json` as the authoritative source on the init/migrate path (or add an explicit `gbrain config set embedding_dimensions` setter that updates both gateway state and the on-disk JSON). If (b), apply the policy inside `applyForwardReferenceBootstrap` too.

2. **Defensive: drop existing HNSW before re-emit.** Even with (1) fixed, future contributors will reintroduce this class of bug whenever they touch the bootstrap. Add a guard: any time SCHEMA_SQL is about to emit `CREATE INDEX ... USING hnsw` on a `vector(N)` column, introspect `pg_attribute.atttypmod` first and skip when `(typmod - 4) > 2000`. That makes the bootstrap dim-aware **at the SQL layer** instead of relying on the gateway value being threaded correctly all the way down.

3. **Postinstall reporting.** Make `gbrain apply-migrations --yes --non-interactive` exit non-zero when bootstrap fails, instead of printing `All migrations up to date.` over a hidden failure.

I'm happy to draft a PR with fix (2) + a regression test that:
1. Creates a fresh test brain with `embedding vector(2560)`,
2. Drops `idx_chunks_embedding`,
3. Runs `initSchema()`,
4. Asserts no HNSW is created and `runMigrations()` reaches the latest version.

## Why this matters now

`zembed-1` (2560-dim) and `zembed-large` (3072-dim) from ZeroEntropy beat OpenAI's `text-embedding-3-large` on long-document retrieval benchmarks; Voyage `voyage-3-large` is 2048-dim. The era where "1536 is the default and everything is HNSW-friendly" is ending. The #640 fix was the right shape but didn't cover the upgrade path — any user who switched embedders before v0.35.2 (when #640 landed) and now tries to upgrade past v0.35.1.1 will hit this.

## Workaround for affected users

Stay on v0.35.1.1 (DB schema v66 — latest reachable without tripping bootstrap):
```bash
cd ~/gbrain && git reset --hard f004a274 && bun install && bun link
```
v67 (typed-claim columns on `facts`) is the only post-66 migration today, and it's purely additive, so rollback loses no data. But all future migrations are blocked on this getting fixed.

## Environment

- gbrain `master` HEAD = `1dadd9ed` (v0.35.7.0)
- bun 1.3.13, macOS Darwin 25.3.0 (Apple Silicon)
- Postgres: Supabase Session pooler (`aws-1-ap-northeast-1.pooler.supabase.com:5432`), pgvector 0.7+
- `GBRAIN_DISABLE_DIRECT_POOL=1` (IPv6 workaround; immaterial to this bug)
- Install method: `git clone` + `bun link`
- Brain stats: 250 pages / 4887 chunks, all embedded at 2560-dim

metaWuming:
## Summary

Closes #1141, #1189.

Adds a defensive SQL-layer guard at `initSchema` time so the upgrade chain stops wedging when the gateway hasn't been configured yet but the live `content_chunks.embedding` column is wider than 2000 dims (zembed-1 2560d, Qwen3-Embedding-4B 2560d, etc.).

## The bug

When `gbrain init --migrate-only` runs inside `gbrain upgrade`'s Phase A, the gateway hasn't read `~/.gbrain/config.json` yet. `getEmbeddingDimensions()` throws and falls through to `DEFAULT_EMBEDDING_DIMENSIONS` (1536). On brains previously switched to a high-dim embedder, the live column is `vector(2560)`. Schema replay emits the unconditional `idx_chunks_embedding USING hnsw` at the gateway dim, pgvector rejects with `column cannot have more than 2000 dimensions for hnsw index`, and Phase A aborts before any v47+ migration runs. `schema_version` stays where it was; upgrade is wedged forever.

Independent repro confirmed by @Moutonc in #1141 against v0.36.3.0 with `Qwen3-Embedding-4B-mxfp8` (same 2560-dim, different provider): the wedge is **not** specific to ZeroEntropy `zembed-1`. The sibling issue #1189 hits the same root cause on the PGLite fresh-install path. This fix closes both.

## The fix

Two layers, both narrow:

1. **`src/core/vector-introspect.ts` (new)** — pure helper that probes `pg_attribute.format_type(atttypid, atttypmod)` for a live vector column. Returns the dim from `vector(N)` or `null` for any failure mode (column absent, type not vector, query throws). Engine-agnostic via a `QueryRunner` callback. Identifier inputs are gated by a strict regex so the table/column names can never reach the wire as injection.

2. **`resolveSchemaDims(gatewayDims, liveDims)`** — pure decision function. Returns the dim to pass to `getPostgresSchema` / `getPGLiteSchema`. **Only overrides when the HNSW-2000 boundary would flip in the opposite direction from the gateway value.** Silently overriding on every call would mask genuine config drift; the override fires exactly where the cap-2000 bug bites and nowhere else.

3. **Both engines' `initSchema`** now call the probe BEFORE `getPostgresSchema(dims, model)` / `getPGLiteSchema(dims, model)`. When the override fires, a single stderr line documents the substitution.

This layers on top of the existing `applyChunkEmbeddingIndexPolicy` (which already skips HNSW when `dims > 2000`) — the policy was already correct; what was missing was making sure the policy got called with the right `dims`. The schema.sql HNSW emission and the migrate.ts emissions for `idx_takes_embedding_hnsw` / `idx_facts_embedding_hnsw` / `idx_query_cache_embedding_hnsw` are unchanged (those columns are hardcoded `VECTOR(1536)`, never trip the cap).

## Test plan

`test/vector-introspect.test.ts` — 20 cases:

- [x] 8-case `resolveSchemaDims` decision matrix (boundary flips both ways, no override when both sides are above-or-both-below cap, no override when null/equal)
- [x] Identifier-shape gates: rejects semicolons, quotes, mixed-case, empty
- [x] `format_type()` parse contract: `vector(N)` round-trip, `text` returns null, missing column returns null, throw-on-query returns null
- [x] Live PGLite round-trip: creates a real `vector(2560)` column, asserts the helper returns 2560 — the exact failure shape from this issue's repro
- [x] All 133 neighbouring tests (`vector-index-lifecycle`, `bootstrap`, `schema-bootstrap-coverage`, `pglite-engine`) still pass

## Out of scope

The deeper "gateway should be configured before initSchema in Phase A" fix that #1141 also hypothesizes (fix #1) is orthogonal and would require threading the config-read into the migrate-only entry. This PR is the SQL-layer guard (fix #2 from the issue) that makes the contract fail-closed regardless of when gateway configuration races against schema replay.

### DIFF
--- a/src/core/pglite-engine.ts
+++ b/src/core/pglite-engine.ts
@@ -18,6 +18,7 @@ import { MAX_SEARCH_LIMIT, clampSearchLimit } from './engine.ts';
 import { runMigrations } from './migrate.ts';
 import { PGLITE_SCHEMA_SQL, getPGLiteSchema } from './pglite-schema.ts';
 import { DEFAULT_EMBEDDING_MODEL, DEFAULT_EMBEDDING_DIMENSIONS } from './ai/defaults.ts';
+import { getLiveVectorColumnDims, resolveSchemaDims } from './vector-introspect.ts';
 import { acquireLock, releaseLock, type LockHandle } from './pglite-lock.ts';
 import type {
   Page, PageInput, PageFilters, PageType,
@@ -230,6 +231,29 @@ export class PGLiteEngine implements BrainEngine {
       dims = gw.getEmbeddingDimensions();
       model = gw.getEmbeddingModel() || model;
     } catch { /* gateway not configured — use defaults */ }
+
+    // #1141/#1189 guard: see the matching block in PostgresEngine.initSchema
+    // for the full rationale. When the gateway isn't configured (the
+    // `apply-migrations` Phase A path inside `gbrain upgrade`) and the
+    // live content_chunks.embedding column is wider than 2000 dims from
+    // a previous switch to a high-dim embedder, the schema replay would
+    // otherwise emit `idx_chunks_embedding USING hnsw` and pgvector would
+    // reject before any v47+ migration runs. resolveSchemaDims() only
+    // overrides when the HNSW-2000 decision flips, so this can't mask
+    // genuine config drift.
+    const liveDims = await getLiveVectorColumnDims(
+      async (sql, params) => {
+        const result = await this.db.query(sql, params ? Array.from(params) : []);
+        return result.rows as ReadonlyArray<Record<string, unknown>>;
+      },
+      'content_chunks',
+      'embedding',
+    );
+    const resolved = resolveSchemaDims(dims, liveDims);
+    if (resolved.overridden) {
+      process.stderr.write(`[initSchema] ${resolved.reason}\n`);
+      dims = resolved.dims;
+    }
 
     await this.db.exec(getPGLiteSchema(dims, model));
 
--- a/src/core/postgres-engine.ts
+++ b/src/core/postgres-engine.ts
@@ -21,6 +21,7 @@ import { runMigrations } from './migrate.ts';
 import { SCHEMA_SQL } from './schema-embedded.ts';
 import { verifySchema } from './schema-verify.ts';
 import { applyChunkEmbeddingIndexPolicy, dropZombieIndexes } from './vector-index.ts';
+import { getLiveVectorColumnDims, resolveSchemaDims } from './vector-introspect.ts';
 import {
   normalizeEngineColumn,
   buildVectorCastFragment,
@@ -227,6 +228,37 @@ export class PostgresEngine implements BrainEngine {
       dims = gw.getEmbeddingDimensions();
       model = gw.getEmbeddingModel() || model;
     } catch { /* gateway not yet configured — use defaults */ }
+
+    // #1141/#1189 guard: when the gateway isn't configured (the
+    // `gbrain init --migrate-only` path inside `gbrain upgrade`'s Phase A),
+    // `dims` falls back to DEFAULT_EMBEDDING_DIMENSIONS. If the live
+    // content_chunks.embedding column was dimensioned for a high-dim
+    // embedder (zembed-1 2560d, Qwen3-Embedding-4B 2560d, etc.), the
+    // gateway value disagrees with what's actually on disk. Schema replay
+    // then emits `idx_chunks_embedding USING hnsw` via the
+    // applyChunkEmbeddingIndexPolicy at the gateway dim — pgvector
+    // rejects with `column cannot have more than 2000 dimensions for hnsw
+    // index` and Phase A aborts before any subsequent migration runs.
+    //
+    // The defensive read: introspect the column via format_type() and
+    // override `dims` when the live value crosses the HNSW-2000 boundary
+    // in the other direction from the gateway value. resolveSchemaDims()
+    // intentionally only fires the override when the override would
+    // change the HNSW policy decision — otherwise we'd silently mask
+    // genuine model/config drift.
+    const liveDims = await getLiveVectorColumnDims(
+      async (sql, params) => {
+        const rows = await conn.unsafe(sql, (params ? Array.from(params) : []) as never[]);
+        return rows as ReadonlyArray<Record<string, unknown>>;
+      },
+      'content_chunks',
+      'embedding',
+    );
+    const resolved = resolveSchemaDims(dims, liveDims);
+    if (resolved.overridden) {
+      process.stderr.write(`[initSchema] ${resolved.reason}\n`);
+      dims = resolved.dims;
+    }
 
     const sqlText = getPostgresSchema(dims, model);
 
--- a/src/core/vector-introspect.ts
+++ b/src/core/vector-introspect.ts
@@ -1,0 +1,124 @@
+/**
+ * Live introspection of pgvector column widths via `pg_attribute.atttypmod` +
+ * `format_type()`. Used at `initSchema` time to detect when the gateway-
+ * resolved embedding dimension disagrees with the actual on-disk column —
+ * the failure mode behind upstream gbrain #1141 and #1189.
+ *
+ * Closes the bug class where `gbrain init --migrate-only` (Phase A of
+ * `gbrain upgrade`) starts before the gateway has been configured from
+ * `~/.gbrain/config.json`. The gateway falls back to
+ * `DEFAULT_EMBEDDING_DIMENSIONS` (currently 1536); the live column for a
+ * brain previously switched to a high-dim embedder (ZeroEntropy zembed-1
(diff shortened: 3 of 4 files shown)


metaWuming:
Confirming #1293 is the fix-shape this needs — it's exactly the SQL-layer guard proposed in #1141 (introspect the live column typmod, skip HNSW when dims > 2000). 👍

**Still blocked by this on a third independent setup** (alongside the OP's Supabase path and @Moutonc's MLX-served Qwen3 path):

- Engine: Supabase Postgres (Session pooler, pgvector), `GBRAIN_DISABLE_DIRECT_POOL=1`
- Embedder: ZeroEntropy `zembed-1` → `content_chunks.embedding` is `vector(2560)`
- `idx_chunks_embedding` HNSW already dropped at switch-time per `docs/embedding-migrations.md`
- Pinned at **v0.35.1.1** (schema_version 66); every attempt past it wedges in the bootstrap `gbrain init --migrate-only` phase before v67 runs

Verified the root line is still unconditional on current `master` (`src/schema.sql:252` in v0.41.14.0): `CREATE INDEX IF NOT EXISTS idx_chunks_embedding ON content_chunks USING hnsw (...)`. Nothing in the v0.36–v0.41 fix-waves (#1182, #1450) reaches the bootstrap path, so #1293 is the only thing that unblocks 2560d+ embedders.

Happy to test a candidate build against this Supabase + zembed-1 setup. @garrytan any chance of a review pass here?

metaWuming:
Closing as superseded. v0.42.1.0 already resolves the HNSW-2000 cap for high-dim (2560d) embeddings on Postgres+pgvector through two mechanisms that landed independently of this PR:

- **`chunkEmbeddingIndexSql(dims)`** (`src/core/vector-index.ts`) skips the `idx_chunks_embedding` HNSW index when `dims > 2000` and falls back to exact scan. Wired into `getPostgresSchema` / `getPGLiteSchema`, so the main `content_chunks.embedding` column never emits an out-of-range HNSW index at schema time.
- The **`facts` / `query_cache`** embedding columns now use `halfvec` when pgvector ≥ 0.7, whose HNSW indexes support up to 4000 dims — so 2560d builds an index rather than wedging.

Verified empirically against my production setup (the exact case this PR targeted): a `zeroentropyai:zembed-1` **2560-dim** brain on Supabase (**pgvector 0.8.0 / PG 17.6**), pinned at schema **v66**, migrated cleanly **v66 → v111** under v0.42.1.0 — 44 migrations, **zero HNSW wedge**. `gbrain doctor` reports `Schema width (2560d) matches` and `schema_version: 111`.

The live-column-introspection approach in this PR is more general than the config-dims approach that shipped, but the shipped fix covers the real-world #1141 / #1189 cases. Closing.

PR #1294, opened 2026-05-22
fix: Windows compat — CRLF parser, relTarget posix, apply-migr leak, gbrainDir contract

xwang4-svg:
## Summary

Four independent Windows-only bugs surfaced by `gbrain doctor` and the `skillpack-{install,check}` test sweep on a Win11 / bun 1.3.13 install. All four still reproduce on current `master` (v0.37.11.0).

* **Bug A** — CRLF frontmatter parser failure at 4 sites (`skill-frontmatter.ts`, `check-resolvable.ts`, `skill-manifest.ts`, `skillpack.ts`). The literal `\n` after the opening `---` fence can't match `\r`, so every Windows-authored `SKILL.md` silently fell through with no frontmatter. Surfaced as **38 false-positive `mece_gap` warnings** on a 41-skill workspace.
* **Bug C** — `enumerateBundle()` in `bundle.ts` used `path.join` for `relTarget`, but `relTarget` is a portable bundle key — Win32 produced `alpha\SKILL.md` while string-matching callers (managed-block writer, install-plan diffs) expected `alpha/SKILL.md`.
* **Bug D** — `apply-migrations --list` / `--dry-run` ran an engine-connect pre-flight just to print an advisory schema-drift warning. The PGLite WASM teardown leaks worker handles under `execFileSync`, surfacing as spurious **exit-1** to parents like `skillpack-check`. v0.36 papered over the symptom with a PGLite-specific `skipPreflight` + explicit `process.exit(0)`; the architectural fix moves the pre-flight after the read-only early returns so informational subcommands skip the connect+disconnect cycle entirely.
* **Bug F** — `preferences.gbrainDir()` and `config.configDir()` disagreed on `GBRAIN_HOME` semantics: configDir treats it as a parent dir (appends `.gbrain`), preferences returned it directly. Same env var, two paths — meant the doctor's migration ledger and `loadConfig` landed in different directories whenever a test or operator set `GBRAIN_HOME`. The preferences docstring even claimed to match `gbrainPath` but didn't. Unify by delegating to `configDir()`.

## What's new

* `extractFrontmatterBlock()` exported from `src/core/markdown.ts` — content-based, CRLF-tolerant. Routes through `\r\n→\n` normalization once at the top, returns LF-only body so downstream `[^"'\n]+?` value parsers don't swallow trailing `\r`.
* `parseSkillFrontmatter()` in `src/core/skill-frontmatter.ts` gets the same one-line normalization at its entry point.
* `bundle.walkFiles()` uses `posix.join` for `relTarget` only; native `join` stays for absolute filesystem paths.
* `runApplyMigrations` reorders: `--list`/`--dry-run` early-return before the schema-drift pre-flight. PGLite skip preserved for the apply path.
* `preferences.gbrainDir()` is now `return configDir()`.

## Test plan

- [x] New regression: `test/check-resolvable.test.ts` — CRLF-terminated `SKILL.md` must not emit `mece_gap`.
- [x] New regression: `test/skillpack-install.test.ts` — `relTarget` must never contain backslashes.
- [x] `test/preferences.test.ts` + `test/skillpack-check.test.ts` fixtures switched from \`HOME\`-only (broken on Windows because \`os.homedir()\` reads \`USERPROFILE\`) to also setting \`GBRAIN_HOME\` (the platform-neutral hook).
- [x] `test/skillpack-check.test.ts` per-test timeout bumped to 30s — the tests spawn `bun run cli.ts skillpack-check`, which itself spawns child `doctor` + `apply-migrations --list` processes; on Windows each cold-start is ~600ms, blowing the default 5s timeout.

Verified on Windows 11 / bun 1.3.13:

| Suite | Before | After |
|---|---|---|
| `gbrain doctor` mece_gap count | 38 | 0 |
| `gbrain doctor` resolver_health | warn | ok |
| `skillpack-check.test.ts` | 1 pass / 4 fail | 5 / 0 |
| `skillpack-install.test.ts` | 29 / 2 fail | 32 / 0 |
| `check-resolvable.test.ts` | 25 pass | 26 (added CRLF test) |
| 8-file affected sweep | 156 pass / 18 fail | 173 / 1 |

The one remaining failure across the sweep is `savePreferences > writes file with 0o600 perms` — a pre-existing POSIX-only chmod test that's unfixable on Windows by nature.

## Scope notes

* Bug E from the same audit (the `GBRAIN_HOME` POSIX-only `startsWith('/')` check) was already fixed upstream between v0.32 and v0.37 — no action needed.
* The `[ai.gateway] recipe "google" declares an embedding touchpoint without max_batch_tokens` advisory is left untouched — per the gateway comments it's intentional (OpenAI/Google recipes use recursive fallback by design).
* No VERSION / CHANGELOG bump — leaving release plumbing to the maintainer.

### DIFF
--- a/src/commands/apply-migrations.ts
+++ b/src/commands/apply-migrations.ts
@@ -354,6 +354,40 @@ export async function runApplyMigrations(args: string[]): Promise<void> {
     if (cli.forceAll) return; // both surfaces flushed
   }
 
+  const completed = loadCompletedMigrations();
+  const idx = indexCompleted(completed);
+  const plan = buildPlan(idx, installed, cli.specificMigration);
+
+  // Bug 3 — surface wedged migrations as a loud, actionable error.
+  if (plan.wedged.length > 0) {
+    for (const m of plan.wedged) {
+      console.error(
+        `\nMigration v${m.version} is WEDGED (${MAX_CONSECUTIVE_PARTIALS}+ consecutive partials with no completion). ` +
+        `Check ~/.gbrain/upgrade-errors.jsonl for the last failure reasons, fix the underlying issue, then run:\n` +
+        `  gbrain apply-migrations --force-retry ${m.version}\n` +
+        `Then re-run \`gbrain apply-migrations --yes\`.`,
+      );
+    }
+    // Don't exit — applied/partial/pending are still worth reporting and running.
+  }
+
+  if (cli.specificMigration && plan.applied.length + plan.partial.length + plan.pending.length + plan.skippedFuture.length === 0) {
+    console.error(`No migration registered with version "${cli.specificMigration}". Run \`gbrain apply-migrations --list\` to see registered versions.`);
+    process.exit(2);
+  }
+
+  // Read-only branches: --list and --dry-run print plan/state without
+  // touching anything. Skip the schema-drift pre-flight (which opens a DB
+  // connection) — its only purpose is to warn before APPLYING; for
+  // informational commands the connect+disconnect cycle is both wasted
+  // work and a portability hazard. On Windows the PGLite WASM teardown
+  // leaks worker handles in execFileSync contexts, surfacing as a spurious
+  // exit-1 to parent processes that shell out to `apply-migrations --list`
+  // (e.g. `skillpack-check`). The PGLite-specific skip below is preserved
+  // for non-read-only paths.
+  if (cli.list) { printList(plan, installed); process.exit(0); }
+  if (cli.dryRun) { printDryRun(plan, installed); process.exit(0); }
+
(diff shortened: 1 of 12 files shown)


time-attack:
Verified each of these against current master and three still reproduce: the `^---\n` fence match remains CRLF-blind in `skill-frontmatter.ts`, `skill-manifest.ts` and `skillpack.ts`; `bundle.ts` still builds the portable `relTarget` key with native `join`; and `preferences.gbrainDir()` still disagrees with `config.configDir()` on GBRAIN_HOME semantics (its docstring even claims a parity it doesn't have). Merging — one note: `check-resolvable.ts` already picked up the CRLF normalization on master independently, so that hunk conflicts and master's version wins on resolution.

PR #1310, opened 2026-05-23
fix: thread voyage_api_key from config.json to gateway env

eepaul:
Symmetric follow-up to the v0.37.11.0 fix wave (#1286). That wave wired `ZEROENTROPY_API_KEY` from `~/.gbrain/config.json` through `buildGatewayConfig` into the AI gateway env. Voyage users hit the same gap.

Voyage's recipe declares `VOYAGE_API_KEY` in `auth_env.required`, so this PR closes the read-side symmetry in three matching spots:

- `GBrainConfig.voyage_api_key?: string` in `src/core/config.ts`
- `process.env.VOYAGE_API_KEY` merge in `loadConfig()`
- Config field → `envFromConfig.VOYAGE_API_KEY` in `buildGatewayConfig()`

12-line mechanical patch, no behavior change for users who export `VOYAGE_API_KEY` as an env var. Only affects the config.json path.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -1444,6 +1444,7 @@ export function buildGatewayConfig(c: GBrainConfig): AIGatewayConfig {
   // plane field now exists (GBrainConfig type) and gets mapped here, so
   // setting it via `~/.gbrain/config.json` propagates into the gateway.
   if (c.zeroentropy_api_key) envFromConfig.ZEROENTROPY_API_KEY = c.zeroentropy_api_key;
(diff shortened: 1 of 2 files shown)


time-attack:
Thank you — the asymmetry you spotted was real. It has since been closed by #2662, which adds the `voyage_api_key` file-plane slot to `src/core/config.ts`, merges it in `loadConfig()`, and threads it into the gateway env dict exactly as ZeroEntropy's does. Closing as already fixed on master.

PR #1319, opened 2026-05-23
fix(subagent): strip anthropic: prefix in legacy SDK path

zscgeek:
## Summary

v0.38.1.0 made `provider:model` the canonical config format and added a submission-time validator ([`queue.ts:106`](../blob/master/src/core/minions/queue.ts#L106), [`subagent.ts:175`](../blob/master/src/core/minions/handlers/subagent.ts#L175)) that rejects bare model ids. The gateway tool loop (`agent.use_gateway_loop=true`) parses the prefix internally via `gateway.toolLoop()`. But the legacy Anthropic-direct path — **still the default** in v0.38.x — passes the resolved model string straight to `client.messages.create()`. So a perfectly valid config of `anthropic:claude-haiku-4-5-20251001` produces:

```
404 not_found_error: model: anthropic:claude-haiku-4-5-20251001
```

Every subagent job dies on this 404 after exhausting retries.

## Repro

```sh
gbrain config set models.dream.synthesize anthropic:claude-haiku-4-5-20251001
gbrain config set models.tier.subagent anthropic:claude-haiku-4-5-20251001
# leave agent.use_gateway_loop unset (default false)
gbrain dream --phase synthesize
```

All dispatched subagent jobs → DEAD with the 404 above.

## Fix

Add `stripAnthropicPrefix()` to `model-config.ts` (sibling to `isAnthropicProvider`) and call it at the SDK call site in `subagent.ts`. Bare model ids pass through unchanged so `TIER_DEFAULTS`-derived strings and pre-v0.38 configs keep working. Non-Anthropic providers throw — caller must guard with `isAnthropicProvider()`, which is already the case at the legacy SDK call site (the gateway-loop branch handles non-Anthropic above).

```ts
// subagent.ts:440
const params: Anthropic.MessageCreateParamsNonStreaming = {
  model: stripAnthropicPrefix(model),  // ← was: model,
  ...
};
```

## Tests

- 3 new cases in `test/model-config.serial.test.ts` (12 expect() calls) covering prefixed, bare, case-insensitive, whitespace, empty, and non-Anthropic rejection
- All 24 pre-existing `test/subagent-handler.test.ts` + `test/subagent-prompt-too-long.test.ts` cases still pass

## Verified end-to-end

Patched a real brain that had 35 dead subagent jobs from this exact 404. Reloaded the autopilot worker, retried one job — went DEAD → COMPLETED on the first attempt under the patched code, wrote real reflection/idea pages. No other config changes needed.

## Notes for reviewer

- The fix is intentionally narrow. The gateway tool loop also has a separate `[chat(anthropic:…)] schema is not a function` regression when called with the brain-tool registry — that's a different bug in `gateway.toolLoop()` that I haven't dug into. This PR keeps users unblocked on the legacy path (which is the default) without touching the gateway path.
- Considered using `parseModelId()` from `model-resolver.ts` but it throws on bare ids, so it would break the `TIER_DEFAULTS.subagent` fallback (which is still a bare string in v0.38+). A standalone helper next to `isAnthropicProvider()` keeps the legacy path's "either format works" contract.

## Test plan

- [ ] `bun test test/model-config.serial.test.ts` — 21 pass
- [ ] `bun test test/subagent-handler.test.ts test/subagent-prompt-too-long.test.ts` — 24 pass
- [ ] Manual: run `gbrain dream --phase synthesize` with `models.dream.synthesize=anthropic:claude-...` configured, confirm subagent jobs complete instead of 404

### DIFF
--- a/src/core/minions/handlers/subagent.ts
+++ b/src/core/minions/handlers/subagent.ts
@@ -47,7 +47,7 @@ import {
   logSubagentSubmission,
   logSubagentHeartbeat,
 } from './subagent-audit.ts';
-import { resolveModel, isAnthropicProvider, TIER_DEFAULTS } from '../../model-config.ts';
+import { resolveModel, isAnthropicProvider, stripAnthropicPrefix, TIER_DEFAULTS } from '../../model-config.ts';
 import { toolLoop as gatewayToolLoop } from '../../ai/gateway.ts';
 import type { ChatToolDef, ChatMessage, ChatBlock, ChatResult, ToolHandler } from '../../ai/gateway.ts';
 import { classifyCapabilities } from '../../ai/capabilities.ts';
@@ -437,7 +437,11 @@ export function makeSubagentHandler(deps: SubagentDeps) {
       // complexity; for v0.15 we lean on the 120s TTL + abort-on-signal.
       try {
         const params: Anthropic.MessageCreateParamsNonStreaming = {
-          model,
+          // v0.38.1.0 made `provider:model` the canonical config format, but
+          // the Anthropic SDK expects bare model ids. The gateway loop strips
+          // the prefix; the legacy path must do it explicitly. `isAnthropicProvider`
+          // guard upstream guarantees this is safe.
+          model: stripAnthropicPrefix(model),
           max_tokens: 4096,
           system: [
             { type: 'text', text: systemPrompt, cache_control: { type: 'ephemeral' } },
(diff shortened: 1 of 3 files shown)


time-attack:
Thanks for this, @zscgeek! 🙏

What this does: PR #1319 fixes a bug where Anthropic SDK calls fail with 404 errors when model IDs include the anthropic: provider prefix. Master already fixed this issue by adding a stripProviderPrefix() function to subagent.ts on 2026-05-24 that strips…

PR #1320, opened 2026-05-23
fix(ingest): prevent silent duplicate pages on overlapping ingest roo…

### LINKED ISSUE #1309: Ingest produces silent duplicate pages when two ingest roots overlap
# Ingest produces silent duplicate pages when two ingest roots overlap

## Summary

When two ingest invocations target overlapping paths (one rooted at a parent
directory, one rooted at a child subdirectory of it), gbrain creates two separate
pages per file. The duplicates share an identical `content_hash`, identical
`source_id`, and identical `frontmatter.id`, differing only in slug. The
ingester dedupes by slug, not by `(source_id, content_hash)`, so this is
silent — no warning at ingest, no surface in `list_pages`.

## Repro

1. `gbrain ingest /path/to/vault/Subdir/` — creates pages with slugs relative
   to `Subdir/` (e.g. `2026-04-28-foo`).
2. Later, `gbrain ingest /path/to/vault/` — same files re-ingested, now with
   path-qualified slugs (e.g. `subdir/2026-04-28-foo`).
3. Result: two pages per file. Both addressable. Both indexed for search.

## Observed in the wild

Single vault, single `source_id: default`, PGLite engine. `list_pages --type meeting-note`
returned 42 entries for 23 source files on disk. 21 pairs shared:

- identical `content_hash`
- identical `frontmatter.id` (granola UUID in this case)
- identical `source_id`

…with the only difference being the slug shape. Confirmed by sampling pairs
and comparing the `get_page` output.

The two ingest log entries that produced the duplicates:

```
source_ref: /path/to/vault/Subdir/      → Imported 23 pages   (bare slugs)
source_ref: /path/to/vault/              → Imported 486 pages  (path-qualified slugs)
```

Both ran under the same `source_id`. Both succeeded. No collision warning.

## Impact

- Search returns both copies, doubling clutter.
- Wikilinks resolve via `frontmatter.id`, so backlink counts are inflated when
  both copies count as the same link target.
- Storage doubles for the affected files.
- The condition is undetectable from `list_pages` output without a content_hash
  comparison the caller has to write themselves.

## Proposed fix

At ingest time, before inserting a new page, look up by `(source_id, content_hash)`:

- If a match exists with a different slug → either (a) skip + log
  `"duplicate of <existing-slug>"`, or (b) update the existing page's slug to
  the new one (preferring the path-qualified form when one is more specific
  than the other). Either is better than silent duplication.
- Surface a warning in the ingest log summary so it's visible without an
  explicit dedup query.

Optional defense in depth: refuse a second ingest whose `source_ref` is a
descendant of a previously-ingested `source_ref` under the same `source_id`,
or require `--allow-overlap`.

## Environment

- gbrain CLI broken locally on macOS 26.3 (per existing issue #223), but the
  MCP surface reproduces fine.
- Single source (`default`), no remote, PGLite engine.

neuralmanacle:
### **Problem**
Ingesting a nested directory followed by its parent directory creates duplicate database pages for identical files. Since slugs are derived relative to the import root (e.g., `check` vs. `subdir/check`), the database uniqueness constraint on `(source_id, slug)` is bypassed, leading to search clutter, bloated storage, and inflated backlinks.

### **Solution**
* Added a pre-import safeguard in `importFromContent` (`src/core/import-file.ts`) to look up pages by `(source_id, content_hash)`.
* If a match exists under a different slug, it logs a warning and short-circuits with `status: 'skipped'` (saving embedding costs).
* Used optional chaining (`duplicateRows?.[0]`) to maintain compatibility with test suite mock engines.

### **Verification**
* **Automated**: Added a new unit test suite to `test/import-file.test.ts` verifying the skip and short-circuit logic.
* **Manual**: Confirmed that sequential overlapping imports cleanly skip duplicate creations with a logged warning.

**Closes #1309**

### DIFF
--- a/src/core/import-file.ts
+++ b/src/core/import-file.ts
@@ -299,6 +299,21 @@ export async function importFromContent(
     tags: parsed.tags,
   };
 
+  // Check if a page with the same content hash already exists in this source under a different slug.
+  const duplicateRows = await engine.executeRaw<{ slug: string }>(
+    `SELECT slug FROM pages 
+     WHERE source_id = $1 AND content_hash = $2 AND deleted_at IS NULL LIMIT 1`,
+    [sourceId ?? 'default', hash]
+  );
+  const existingDuplicate = duplicateRows?.[0];
+
+  if (existingDuplicate && existingDuplicate.slug !== slug) {
+    console.warn(
+      `[warning] "${slug}" is a duplicate of existing page "${existingDuplicate.slug}" (same content hash). Skipping.`
+    );
+    return { slug, status: 'skipped', chunks: 0, parsedPage };
+  }
+
   const existing = await engine.getPage(slug, sourceId ? { sourceId } : undefined);
   if (existing?.content_hash === hash && !opts.forceRechunk) {
     return { slug, status: 'skipped', chunks: 0, parsedPage };
--- a/test/import-file.test.ts
+++ b/test/import-file.test.ts
@@ -702,3 +702,53 @@ body unchanged
     expect(shortCircuited).toBe(true);
   });
 });
+
+describe('importFromContent — duplicate slug detection (#1309)', () => {
+  test('skips and warns when identical content hash exists under a different slug', async () => {
+    const content = `---
+type: concept
+title: Unique Content
+---
+
+This content is unique and will be duplicate-checked.
+`;
+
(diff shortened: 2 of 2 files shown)


garrytan:

PR #1334, opened 2026-05-24
fix(extract): timeline bullet parser splits on hyphens inside slugs/prose

Chapwi:
## Problem

`extractTimelineFromContent` (the parser used by `gbrain extract`, sync, and the autopilot cycle) splits `- **YYYY-MM-DD** | …` bullets into `source` + `summary` on the first em-dash, en-dash, **or bare hyphen**:

```ts
```

Because a bare hyphen counts as a delimiter, any bullet whose text contains a hyphen inside a word or a markdown-link slug is chopped at the wrong place, and the two halves render reordered in the timeline readout (`${summary} [${source}]`). Examples that mis-split today:

- `| Referenced in [Acme Consulting Group](../companies/acme-consulting-group.md)` → `source = "Referenced in [Acme Consulting Group](../companies/acme"`, `summary = "consulting-group.md)"`
- `| Created during architecture-fix session` → `source = "Created during architecture"`, `summary = "fix session"`

Most real-world bullets are free-form `| text`, not the intended `| Source — Summary`, so the split misfires constantly and silently accumulates scrambled `timeline_entries` via the autopilot/sync path.

This also diverges from the project's *other* timeline parser, `parseTimelineEntries` (`src/core/link-extraction.ts`), used by the `put_page` auto-timeline hook, which already treats the whole line as the summary. The same bullet therefore produces different `timeline_entries` depending on which code path wrote it.

## Fix

Keep the intended `Source — Summary` feature, but only peel off a source label when it is unambiguous: a short label (≤24 chars) containing no markdown bracket/paren syntax, followed by a **spaced** em-dash or en-dash. A bare hyphen is never treated as a delimiter.

This keeps every existing test green (`| Meeting — Discussed partnership` still yields `source: "Meeting"`) and stops the slug/prose mangling.

## Tests

Existing `extractTimelineFromContent` tests unchanged; 3 regressions added (hyphenated link slug, bare-hyphen prose, long prose clause before a spaced em-dash). `bun test test/extract.test.ts` → 24 pass / 0 fail.

## Alternative

A fully equivalent option is to drop the source-split from `extractTimelineFromContent` entirely and unify on `parseTimelineEntries`' whole-line behavior (one parser instead of two). Happy to take that direction instead if you'd prefer it.

### DIFF
--- a/src/commands/extract.ts
+++ b/src/commands/extract.ts
@@ -273,11 +273,28 @@ export async function extractLinksFromFile(
 export function extractTimelineFromContent(content: string, slug: string): ExtractedTimelineEntry[] {
   const entries: ExtractedTimelineEntry[] = [];
 
-  // Format 1: Bullet — - **YYYY-MM-DD** | Source — Summary
-  const bulletPattern = /^-\s+\*\*(\d{4}-\d{2}-\d{2})\*\*\s*\|\s*(.+?)\s*[—–-]\s*(.+)$/gm;
+  // Format 1: Bullet — `- **YYYY-MM-DD** | [Source — ]Summary`
+  // Capture everything after `|`, then peel off an optional leading "Source"
+  // label ONLY when it is unambiguous: a short label (<=24 chars), containing
+  // no markdown link/bracket syntax, followed by a SPACED em-dash or en-dash.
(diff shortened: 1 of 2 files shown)


garrytan:

PR #1259, opened 2026-05-21
fix: drain last-retrieved writes before CLI disconnect

jehoon:
## Summary
- track `last_retrieved_at` fire-and-forget promises in `last-retrieved.ts`
- add `awaitPendingLastRetrievedWrites()` and drain it from the short-lived CLI before `engine.disconnect()`
- add structural regression coverage for the CLI drain and pending-promise tracking

## Root cause
`search` / `query` / `get_page` can call `bumpLastRetrievedAt()` after returning results. That helper intentionally starts a background config read / write-back without awaiting it. In a long-lived MCP process that is fine, but in the local CLI path stdout can be written and then `engine.disconnect()` can race the background PGLite work, leaving the Bun process alive after output.

## Verification
- Reproduced installed `gbrain 0.37.3.0`: `gbrain search AIAgent --limit 1` prints output and does not exit within 5s when run under the Hermes worktree source pin.
- Verified patched CLI from this branch in the same cwd exits successfully:
  - `bun run /Users/agentdeveloper/workspace/gbrain/src/cli.ts search AIAgent --limit 1` → exit 0 in ~1.2s
- Installed patched checkout locally with `bun install -g /Users/agentdeveloper/workspace/gbrain` and verified:
  - `gbrain search AIAgent --limit 1` → exit 0 in ~1.17s
  - no-hit search still exits normally
- `bun test test/fix-wave-structural.test.ts`
- `bun run typecheck`

Fixes #1247

PR #1337, opened 2026-05-24
fix: stop PGLite CLI commands from hanging

matt-dean-git:
## Summary
- release the PGLite advisory/file lock before optional close during disconnect
- let one-shot CLI commands skip the hanging PGLite close path and force-exit after successful completion
- preserve daemon/server behavior by not force-exiting `serve --http`

## Verification
- `bun run typecheck`
- fresh PGLite temp brain: `gbrain list` and `gbrain config show` both exit in <1s instead of hanging
- fresh PGLite temp brain: `gbrain serve --http --port <free>` remains running and serves OAuth metadata
- `bun test test/cycle-pglite-lock-ordering.serial.test.ts test/admin-embed-spawn.serial.test.ts`

## Notes
- Full local `bun run test` was attempted before narrowing; it still has unrelated environment-sensitive serial failures around embedding/provider setup. The regression surfaces touched here are covered by the focused checks above.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -63,6 +63,9 @@ const CLI_ONLY_SELF_HELP = new Set([
 ]);
 
 async function main() {
+  if (shouldForceExitAfterMain()) {
+    process.env.GBRAIN_PGLITE_SKIP_CLOSE_ON_DISCONNECT ??= '1';
+  }
   // Parse global flags (--quiet / --progress-json / --progress-interval)
   // BEFORE command dispatch, so `gbrain --progress-json doctor` works.
   // The stripped argv is what the command sees.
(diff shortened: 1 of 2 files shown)


garrytan:
Thanks for this, @matt-dean-git — closing as superseded. Your work shipped in **v0.41.8.0**. From the [CHANGELOG](https://github.com/garrytan/gbrain/blob/master/CHANGELOG.md):\n\n> Two community PRs proposed competing fixes. [PR #1259](https://github.com/garrytan/gbrain/pull/1259) (jehoon, validated by @eloe, @bcallender, and @61tH0b) added the structural "wait for the background write to finish" pattern — the right approach, mirroring an existing fix we sh…\n\nReal appreciation for the contribution.

PR #1347, opened 2026-05-24
fix: drain last-retrieved writes before CLI disconnect

alexzhu0:
## Summary

- Track in-flight `last_retrieved_at` write-back promises from `bumpLastRetrievedAt()`.
- Drain those best-effort writes briefly during CLI shutdown before `engine.disconnect()`.
- Add regression coverage for normal drain behavior, swallowed write-back failures, and bounded timeout behavior.

## Why

During isolated PGLite smoke testing, `gbrain get`, `gbrain search`, and `gbrain query --no-expand` could print their results and then fail to exit. The common path was retrieval tracking: `bumpLastRetrievedAt()` starts a fire-and-forget write-back, then the CLI closes the engine while that write is still in flight.

This keeps the visible operation path fire-and-forget, but gives short-lived CLI commands a shutdown drain so they do not race PGLite disconnect. The drain has a 2s default timeout so a stuck best-effort write-back cannot keep the CLI alive forever.

## Verification

- `bun test test/last-retrieved.test.ts`
- `bun run typecheck`
- Isolated CLI smoke with `GBRAIN_HOME=/tmp/...`:
  - `bun run src/cli.ts init --pglite --no-embedding --yes`
  - `bun run src/cli.ts put last-retrieved-smoke < /tmp/gbrain-pr-smoke-page.md`
  - `bun run src/cli.ts get last-retrieved-smoke`
  - `bun run src/cli.ts search 'last retrieved smoke'`
  - `bun run src/cli.ts query --no-expand 'last retrieved smoke'`
- `bun run check:test-names`
- Static scan of staged added lines for hardcoded secrets / shell injection / eval / unsafe deserialization / SQL string-formatting patterns: no findings.
- Independent reviewer pass: no blocking security or logic concerns.

Note: I also attempted `bun run verify`; it passed through privacy/proposal-PII/test-names/jsonb/source-id/progress/test-isolation checks, then local `check:wasm` was killed with exit 137 in this environment.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -199,6 +199,12 @@ async function main() {
     console.error(e instanceof Error ? e.message : String(e));
     process.exit(1);
   } finally {
+    try {
+      const { awaitPendingLastRetrievedWrites } = await import('./core/last-retrieved.ts');
+      await awaitPendingLastRetrievedWrites();
+    } catch {
+      // Best-effort shutdown drain only. Never mask the original command result.
+    }
     await engine.disconnect();
   }
 }
--- a/src/core/last-retrieved.ts
+++ b/src/core/last-retrieved.ts
@@ -38,6 +38,49 @@ import { isUndefinedColumnError } from './utils.ts';
 
 let _trackRetrievalCache: { ts: number; enabled: boolean } | null = null;
 const TRACK_RETRIEVAL_CACHE_TTL_MS = 30_000;
+const pendingLastRetrievedWrites = new Set<Promise<unknown>>();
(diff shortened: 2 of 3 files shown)


alexzhu0:
Thanks for taking a look. I noticed #1405 is now carrying a broader PGLite hang fix wave that appears to cover the same `last_retrieved_at` shutdown-drain class this PR targets.

The isolated smoke here still validates the same failure mode (`get` / `search` / `query --no-expand` printing results but not exiting cleanly) and the bounded-drain approach. If #1405 supersedes this PR, I'm happy to close this one and leave it as an additional reproduction / verification signal.

PR #1354, opened 2026-05-24
fix: register reindex command and update skillpack-harvest routing-eval

lost9999:
Two fixes:

1. **reindex CLI registration:** Add `reindex` to `CLI_ONLY` dispatcher set so `gbrain reindex --help` works instead of printing `Unknown command`.
2. **routing-eval fixtures:** Update skillpack-harvest routing-eval.jsonl intents to match current resolver triggers (resolves 7 resolver_health warnings on `gbrain doctor`).


garrytan:
Both fixes landed in #1458. The routing-eval half took the trigger-broadening direction (moved from RESOLVER.md to frontmatter per the v0.41.14.0 structural rework). The `reindex` CLI registration hunk shipped as-is — verified locally that `gbrain reindex --help` was printing 'Unknown command' before this fix. Thanks @lost9999 for the adjacent CLI catch.

PR #1358, opened 2026-05-24
fix: allow contextual synopsis model env override

lost9999:
## Summary
- lets contextual retrieval synopsis use GBRAIN_CONTEXTUAL_SYNOPSIS_MODEL before falling back to Anthropic Haiku
- enables operators without ANTHROPIC_API_KEY to route per_chunk_synopsis through their configured gateway model

## Test Plan
- bun --check src/core/contextual-retrieval-service.ts

Context: local Postgres brain had per_chunk_synopsis jobs falling back to title with audit detail: Anthropic chat requires ANTHROPIC_API_KEY, despite GBRAIN_CONTEXTUAL_SYNOPSIS_MODEL being set in the worker env.

### DIFF
--- a/skills/skillpack-harvest/routing-eval.jsonl
+++ b/skills/skillpack-harvest/routing-eval.jsonl
@@ -1,7 +1,7 @@
-{"intent": "lift this skill back into gbrain so others can use it", "expected_skill": "skillpack-harvest"}
-{"intent": "publish my fork-only skill upstream", "expected_skill": "skillpack-harvest"}
-{"intent": "share this skill with the gbrain bundle", "expected_skill": "skillpack-harvest"}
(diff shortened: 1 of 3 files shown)


garrytan:

ktrawick:
The underlying issue is still live on master @ `1057bf4` (0.42.67.0). I couldn't reopen this PR (not the author), so I've written it up with current file/line evidence in #3597.

Worth noting for whoever picks it up: the env-var default proposed here would no longer be sufficient on its own. `src/core/import-file.ts:754` now passes `haikuModel: 'anthropic:claude-haiku-4-5-20251001'` as a literal at the call site, which overrides any default the service resolves — so the call site has to be fixed too. This also matters more since #3310 landed the `claude-cli` recipe, which gives no-API-key users a working chat path that this code path still can't reach.

PR #1365, opened 2026-05-24
feat(put_page): add --file param to bypass Windows pipe buffer limitation

### LINKED ISSUE #1363: Add --file parameter to bypass Windows pipe buffer limitation
## Problem

On Windows, the default pipe buffer size is limited (approximately 45KB). When passing large content (e.g., files >45KB) to `gbrain` via stdin/pipe, the data is silently truncated or fails without any error message.

### Steps to reproduce

1. Create a markdown file larger than 45KB
2. Run: `gbrain put_page --title "Test" < large_file.md`
3. The content is silently truncated or the operation fails

### Expected behavior

`gbrain` should be able to handle files of any size reliably.

### Actual behavior

Content >45KB gets silently truncated on Windows due to pipe buffer limitations. No error message is shown to the user.

## Proposed Solution

Add a `--file` (or `--input-file`) parameter that accepts a file path and reads the content directly from the file instead of stdin. This completely bypasses the pipe buffer limitation.

### Example usage

```bash
# Instead of (fails on Windows for large files):
gbrain put_page --title "Test" < large_file.md

# Use new --file parameter:
gbrain put_page --title "Test" --file large_file.md
```

## Additional Context

- This issue is **Windows-specific** — Unix systems have much larger pipe buffers so the problem is less visible there
- The `--file` parameter workaround has been manually patched locally and works reliably
- This would make `gbrain` more robust on Windows and also improve performance for large files on all platforms

## Environment

- OS: Windows 10/11
- gbrain version: (please fill in)
- Python version: 3.13

---

Would you be open to accepting a PR that implements this feature?

ecat2010:
## Problem

On Windows, the default pipe buffer size is limited (~45KB). Passing large content to `gbrain put_page` via stdin/pipe silently truncates or fails.

## Solution

Add a `file` parameter to `put_page` operation. When provided, the handler reads content directly from the file, completely bypassing the pipe buffer limitation.

## Changes

- `content` param: `required: false` (was `true`, now optional)
- Add `file` param to `put_page` operation params
- Handler: if `p.file` is set, read file content via `fs.readFile`
- Handler: throw `invalid_request` if neither `content` nor `file` is provided

## Backward Compatibility

- Existing clients that pass `content` only: still work unchanged
- `capture --file` CLI: already works (uses `content` after reading file locally)
- `put_page` with `file` param: new feature for MCP and direct API callers

## Related

Fixes #1363

---
- [x] `content` param backward compatible
- [x] `file` param reads file correctly  
- [x] Error when neither param is provided

### DIFF
--- a/src/core/operations.ts
+++ b/src/core/operations.ts
@@ -459,6 +459,17 @@ const get_page: Operation = {
   },
   handler: async (ctx, p) => {
     const slug = p.slug as string;
+
+    // v0.40: support file parameter to bypass Windows pipe buffer limitation
+    if (p.file) {
+      const fs = await import('node:fs/promises');
+      p.content = await fs.readFile(p.file as string, 'utf-8');
+    }
+
+    if (!p.content) {
+      throw new OperationError('invalid_request', 'put_page requires either content or file parameter');
+    }
+
     const fuzzy = (p.fuzzy as boolean) || false;
     const includeDeleted = (p.include_deleted as boolean) === true;
     // v0.31.8 (D20): thread ctx.sourceId through read-side ops. Only pass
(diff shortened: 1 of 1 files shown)


garrytan:
Closing as redundant with the already-shipped `gbrain capture --file PATH --slug SLUG` (v0.39.3.0). Capture is the canonical file-as-input ingestion path on every platform — it reads files via `readFileSync(path)` (Buffer-first, no encoding), scans the first 8KB for NUL bytes to refuse binary content, decodes to UTF-8 only after the safety check, and adds provenance write-through. The intent behind this PR is real, especially for Windows users hitting pipe-buffer limits.

The v0.41.4.0 wave (#1377) lands a doc improvement that points Windows / pipe-buffer users at capture from `gbrain put_page --help`, so the discoverability gap closes even though the code change itself doesn't ship. Your contributor credit is in the wave's commit body.

Historical context for posterity: the original diff inserted the new handler block into `get_page`'s handler instead of `put_page`'s — every read in the system would have thrown `invalid_request`. The v0.41.4.0 wave's R1 + R2 regression tests pin both behaviors permanently.

PR #1370, opened 2026-05-24
feat: parseResolverEntries supports list-based resolver format

garrytan-agents:
## Problem

`parseResolverEntries` only handles markdown table format:
```
| trigger | `skills/name/SKILL.md` |
```

But OpenClaw agents use a compact list format:
```
- **skill-name**: trigger1 | trigger2 | trigger3
```

This causes `checkResolvable` to find 0 reachable skills → 238 FAIL errors in `gbrain doctor`.

## Fix

Adds a second parser branch for list-format entries. Both formats are handled in the same pass:

- **Format 1** (table): `| trigger | \`skills/name/SKILL.md\` |` — unchanged
- **Format 2** (list): `- **name**: trigger1 | trigger2` — new
  - Bold or plain skill names
  - Pipe-delimited triggers
  - Optional `→ \`skills/...\`` suffix (stripped)
  - Ellipsis markers filtered

Table format continues to work identically. Tested against both formats.

## Impact

- `gbrain doctor` resolver_health: FAIL (238 errors) → 0 errors
- No changes needed to existing table-format resolvers
- OpenClaw compact resolvers now work out of the box

### DIFF
--- a/src/core/check-resolvable.ts
+++ b/src/core/check-resolvable.ts
@@ -127,29 +127,50 @@ export function parseResolverEntries(resolverContent: string): ResolverEntry[] {
       continue;
     }
 
-    // Skip non-table rows
-    if (!line.startsWith('|') || line.includes('---')) continue;
+    // ── Format 1: Markdown table rows ──
+    // | trigger phrase | `skills/<name>/SKILL.md` |
+    if (line.startsWith('|') && !line.includes('---')) {
+      const cols = line.split('|').map(c => c.trim()).filter(Boolean);
+      if (cols.length < 2) continue;
 
-    // Split table columns
-    const cols = line.split('|').map(c => c.trim()).filter(Boolean);
-    if (cols.length < 2) continue;
+      const trigger = cols[0];
+      const skillCol = cols[1];
 
-    const trigger = cols[0];
-    const skillCol = cols[1];
+      // Skip header rows
+      if (trigger.toLowerCase() === 'trigger' || trigger.toLowerCase() === 'skill') continue;
 
-    // Skip header rows
-    if (trigger.toLowerCase() === 'trigger' || trigger.toLowerCase() === 'skill') continue;
+      // Check for GStack entries
+      if (skillCol.startsWith('GStack:') || skillCol.startsWith('Check ') || skillCol.startsWith('Read ')) {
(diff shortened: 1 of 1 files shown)


garrytan:
Thanks for catching this — the diagnosis was exactly right. The compact list-format resolver that OpenClaw agents actually write was being silently treated as zero skills by `parseResolverEntries`, and `gbrain doctor` reported every skill as unreachable as a result.

Absorbed the fix into v0.41.7.0 (#1407) with the same parser branch you proposed, expanded in a few places after a Codex outside-voice review on the plan:

1. **Restructured the parser to an if/else-if shape.** Adding the list branch below the existing `if (!line.startsWith('|') ...) continue;` would have made it dead code — the continue short-circuits every non-table row before the list branch can fire. The shipped version uses the if/else-if pattern so both branches actually execute.

2. **Tightened the skill-name regex to kebab-lowercase** (`[a-z][a-z0-9-]+`). Real-world `AGENTS.md` files have prose bullets like `- **Note**: see [link]` and `- **Convention**: read this first`. The original `[\w-]+` regex would parse those as fake skill rows pointing at `skills/Note/SKILL.md`. The kebab-lowercase tighten kills the noise class at source.

3. **Strip the optional `→` / `->` path suffix instead of capturing it.** Initially planned to honor the explicit path, but Codex caught that two downstream consumers (`routing-eval.ts:skillSlugFromPath` and the manifest check at `check-resolvable.ts:367`) both assume `skills/<name>/SKILL.md` shape. Honoring the explicit path silently breaks their coverage. List format now matches the upstream PR behavior (suffix stripped, path always derived); non-conventional paths use the table format.

4. **Integration regression coverage.** Two new fixtures (`test/fixtures/openclaw-compact-resolver/` and `test/fixtures/openclaw-mixed-merge/`) plus a regression test file (`test/check-resolvable-openclaw-compact.test.ts`) pin the "238 FAILs → 0" outcome AND the v0.31.7 D-CX-14 mixed-merge case (skillpack `RESOLVER.md` + workspace `../AGENTS.md`).

5. **Docs context.** `docs/guides/scaling-skills.md` is a new tutorial explaining the three-tier scaling architecture (always-loaded, resolver-routed, dormant) that makes this parser fix useful at the agent layer.

Closing this PR in favor of v0.41.7.0 (#1407). Credit preserved in the v0.41.7.0 CHANGELOG entry. Thanks for surfacing the gap.

Issue #1371, opened 2026-05-24
OAuth source scope bypass via explicit source_id in query


jlfetter1:
GBrain v0.40.8.x appears to enforce OAuth/federated source scope correctly on default read paths, but explicit source overrides can widen remote MCP caller access.

## Security classification

- CWE-285 Improper Authorization: https://cwe.mitre.org/data/definitions/285.html
- Related: CWE-266 Incorrect Privilege Assignment: https://cwe.mitre.org/data/definitions/266.html
- Related: CWE-20 Improper Input Validation: https://cwe.mitre.org/data/definitions/20.html
- This matches Bishop Fox's "Authorization Bypass through User Controlled Parameter" pattern: https://github.com/BishopFox/llm-testing-findings/blob/main/Authorization-Bypass-through-User-Controlled-Parameter.md

The key issue is that `source_id` is authority-bearing. A remote MCP caller may request a source, but the server must validate the requested source against the caller's token claims (`sourceId` / `allowedSources`) at tool invocation time. Token signature and expiry validation alone are insufficient.

## Observed behavior

In a multi-client OAuth deployment using synthetic canary data:

- `get_page`, `search`, and default `query` did not leak a page written only to another client's source.
- `query` with `source_id=<other-source>` returned the other client's page.
- `query` with `source_id=__all__` also returned the other client's page.

Separately, `get_page` did not expose a safe explicit source selector, so a federated client could not retrieve a canonical page from an allowed shared source without maintaining source-local mirrors.

## Expected behavior

- Remote OAuth callers may only query sources in their allowed source set.
- `source_id=__all__` from remote OAuth callers should be clamped to allowed sources, or rejected.
- Unauthorized explicit source requests should return a generic forbidden/insufficient-scope style error, not an empty success.
- `get_page` should support explicit source selection within the caller's allowed source set.
- Source-scoped tool calls should audit the caller source, requested source, effective source decision, and whether the request was rejected or clamped.

## Reproduction note

No sensitive data was involved in this reproduction. The test used synthetic canary pages and then soft-deleted them.

I have a draft patch ready from a fork and will open a PR referencing this issue.

garrytan:
[...]

PR #1372, opened 2026-05-24
Clamp remote source overrides

jlfetter1:
## Summary

Fixes the OAuth source-scope bypass reported in #1371.

This PR treats remote `source_id` tool parameters as requests, not authority. For remote MCP callers, source selection is now clamped to the caller's OAuth `allowedSources` / `sourceId` claims at tool invocation time.

## Changes

- Clamp remote `query.source_id` to the caller's allowed source set.
- Reject unauthorized explicit source requests with `permission_denied` instead of returning empty success.
- Clamp remote `source_id=__all__` to the caller's allowed sources.
- Add safe explicit `get_page.source_id` selection for allowed shared/canonical sources.
- Reject ambiguous `get_page source_id=__all__`.
- Thread source scope through `think` gather.
- Apply source filters to `searchTakes` / `searchTakesVector` in Postgres and PGLite implementations.
- Add source-scope audit metadata to HTTP MCP request logs without logging query text or page bodies.
- Add focused PGLite regression coverage for unauthorized source override, clamped `__all__`, explicit shared `get_page`, unauthorized `get_page`, and `think` source scoping.

## Security framing

The issue maps to CWE-285 Improper Authorization, with related CWE-266 and CWE-20 concerns. It also matches Bishop Fox's LLM testing pattern "Authorization Bypass through User Controlled Parameter." In this case, `source_id` is authority-bearing and must be validated server-side against token claims.

## Verification

Executed locally in the fork worktree:

```text
npm exec --yes bun -- test test/e2e/source-isolation-pglite.test.ts
```

Result: 19 passed, 0 failed.

```text
npm exec --yes bun -- run typecheck
```

Result: passed.

Also live-tested in a downstream OAuth deployment with synthetic canary pages: unauthorized `query.source_id` no longer leaked cross-source content, `source_id=__all__` clamped to allowed sources, explicit `get_page source_id=shared` worked, and source-scope audit rows were written.

No sensitive data was used in the reproduction or verification.

### DIFF
--- a/src/commands/serve-http.ts
+++ b/src/commands/serve-http.ts
@@ -52,6 +52,78 @@ import {
  */
 export const HEALTH_TIMEOUT_MS = 3000;
 
+type SourceScopeAuditOutcome =
+  | 'default_allowed_scope'
+  | 'default_caller_source'
+  | 'explicit_allowed'
+  | 'clamped_to_allowed_sources'
+  | 'rejected_unauthorized'
+  | 'rejected_ambiguous_all';
+
+interface SourceScopeAudit {
+  caller_source_id: string;
+  requested_source_id: string | null;
+  allowed_sources: string[];
+  effective_source_ids: string[];
+  outcome: SourceScopeAuditOutcome;
+}
+
+function buildSourceScopeAudit(
+  operation: string,
+  params: unknown,
+  authInfo: AuthInfo,
+  tokenSourceId: string,
+): SourceScopeAudit | null {
+  const sourceScopedOps = new Set(['query', 'search', 'list_pages', 'get_page', 'think']);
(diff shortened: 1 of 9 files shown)


time-attack:
Thank you for the careful writeup and the CWE framing — this was a real leak class. It has since been closed on master: `resolveRequestedScope` in `src/core/operations.ts` is now the single trust+grant resolver for every read op that accepts `source_id`/`all_sources`, clamping `__all__` to the caller's grant and raising `permission_denied` on an out-of-grant explicit source, with coverage in `test/source-scope-resolver.test.ts`; commit d574e843 (#2881) also made remote dispatch refuse calls with no resolved source scope. Closing as already fixed rather than for any fault in the patch.

PR #1387, opened 2026-05-24
Make capture default to inbox source

Mikemansour:
## Objective

Make `gbrain capture "thought"` safe as a human inbox command when run from inside repos that have `.gbrain-source`, while preserving explicit source routing for operators.

## Changes

- Plain capture now writes to the default inbox source unless `--source` or `GBRAIN_SOURCE` is set.
- `--source <id>` still routes capture to the explicit source.
- `GBRAIN_SOURCE=<id>` still routes capture to the operator-selected source.
- Capture help now documents the inbox-first policy.
- Existing local branch also includes the CLI hang fix for pending `last_retrieved_at` writebacks.

## Verification

- `bun test test/commands/capture.test.ts test/source-resolver-with-tier.test.ts`: 35 pass, 0 fail
- `git diff --check`

## Context

This came from Axiom issue #9: plain capture from `/Users/mike/axiom` inherited `.gbrain-source` and wrote to the Axiom source mirror instead of the general inbox. For human-facing quick capture, cwd source inheritance is surprising; repo/source writes should be explicit.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -16,6 +16,7 @@ import type { CliOptions } from './core/cli-options.ts';
 import { callRemoteTool, RemoteMcpError, unpackToolResult } from './core/mcp-client.ts';
 import { maybePromptForUpgrade } from './core/thin-client-upgrade-prompt.ts';
 import { VERSION } from './version.ts';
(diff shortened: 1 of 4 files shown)


garrytan:

PR #1388, opened 2026-05-24
v0.42.10.0 feat(extract): opt-in global-basename wikilink resolution (closes #972)

### LINKED ISSUE #972: Feature: opt-in global-basename wikilink resolution for Obsidian-convention vaults

ukd1:
## What this adds

**Opt-in global-basename wikilink resolution** — closes #972.

If you import an Obsidian or Notion vault that uses bare `[[note-name]]` wikilinks, GBrain used to silently drop any link whose target lived in a different folder. Write `[[struktura]]` in `concepts/knowledge-graph.md` while the page lives at `projects/struktura.md`, and the edge never landed — Obsidian showed a dense graph, GBrain showed a thin, broken one. The issue reporter had 71 wikilinks across 20 pages; GBrain captured 12.

This PR adds an **opt-in** mode that resolves bare wikilinks by basename (slug tail). Off by default — existing brains are unchanged.

```bash
gbrain config set link_resolution.global_basename true
gbrain extract links
```

### Behavior
- **Multi-match by design.** If `[[struktura]]` matches both `projects/struktura` and `archive/struktura`, GBrain emits **one edge to each**, tagged with the new `wikilink_basename` link type and `link_source = 'wikilink-resolved'`. No arbitrary winner-takes-all.
- **All three resolution surfaces covered:** `gbrain extract links` (FS-walk — the path the issue's repro hits — and `--source db`), plus the auto-link post-hook on every `put_page`.
- **Discoverable.** `gbrain doctor` adds a `link_resolution_opportunity` check that surfaces a paste-ready enable hint when ≥5 bare wikilinks would resolve at a ≥20% ratio, so you know the payoff before flipping the flag.
- **Resolution order** for the flag: env `GBRAIN_LINK_RESOLUTION_GLOBAL_BASENAME` → DB config `link_resolution.global_basename` → default `false`.

### Schema
- Migration **v110** widens the `links_link_source_check` CHECK to admit `'wikilink-resolved'` (full set: `markdown`, `frontmatter`, `manual`, `mentions`, `wikilink-resolved`). Idempotent (`DROP ... IF EXISTS`), engine-parity `sql` + `sqlFor.pglite`. Mirrored in `schema.sql` + `pglite-schema.ts` for fresh installs.

### Docs
- **README** "Self-wiring knowledge graph" names the Obsidian case + the doctor pre-check.
- **INSTALL_FOR_AGENTS.md → Step 4.5** gets an agent-facing subsection: when to enable, the command, re-running `extract`, the doctor hint, and the multi-match behavior.

### Tests
- `test/link-extraction.test.ts` (+17), `test/extract-fs.test.ts` (+14), `test/doctor.test.ts` (+7), and `test/e2e/global-basename-pglite.test.ts` (7, including the issue's exact `mkdir -p /tmp/vault/{concepts,projects}` repro across FS-source, DB-source, and put_page auto-link under both flag states).

### Credit
PR #1233 from @rayers contributed the kernel — the generic wikilink regex + slug-tail index pattern. This PR keeps that mechanism, makes it opt-in via the config flag, replaces the first-write-wins lookup with multi-match return, and extends coverage to the FS-source path the repro actually hits.

---

Synced to current `master` (v0.41.34.0); the `'wikilink-resolved'` value is unioned with master's `'mentions'` so neither clobbers the other.


rayers:
Heads up — I have a local commit on my fork (rayers/gbrain @ local/integration, `85ba5f1d` from 2026-05-09) that addresses the same bug class by removing the `DIR_PATTERN` gate in `WIKILINK_RE` outright. Approaches diverge on the policy question:

- Yours: opt-in via `link_resolution.global_basename`, default off — strict backward compat.
- Mine: bare wikilinks just work by default — no flag needed.

Worth weighing: the user who hits #972 won't know to flip the config knob, so they'll experience the silent-drop the issue describes until someone tells them about it. The 71→12 wikilink ratio in the original repro feels bad enough to me to advocate default-on.

Happy to test your patch against my ~440K-page brain once it lands and report numbers. Either way thanks for the thorough test coverage — substantially more comprehensive than what I had locally.

ukd1:
@rayers it's working well for mine! LMK!

> The 71→12 wikilink ratio in the original repro feels bad enough to me to advocate default-on.

@garrytan thoughts on default on?

rayers:
@ukd1 nice — glad it's holding up on your vault.

I'm fully on board with **default-on**. The 71→12 ratio in #972 is the tell: bare `[[note-name]]` is the *common* Obsidian/Notion shape, not the exception, so silently dropping cross-folder links is a broken default, not a safe one. Strict backward-compat is protecting behavior nobody actually wants.

Given yours is tested and working, I'd rather converge on it than ship two implementations. My #1233 (and the older `local/integration` commit that just strips the `DIR_PATTERN` gate in `WIKILINK_RE`) solves the same bug class a blunter way. If you flip the flag to default-on, I'll **close #1233 in favor of #1388** — one resolver path is better than two. If @garrytan would rather keep it opt-in for one release and flip the default later, that works too; I'll just rebase #1233 down to whatever's left.

Either way, happy to review the resolver code if a second pair of eyes helps it land.

PR #1394, opened 2026-05-24
fix(mcp): get_page honors per-call source_id parameter

Valentin32:
Prior to this commit, get_page ignored the per-call source_id parameter entirely — only OperationContext.sourceId was respected. MCP clients with a default-source bearer (e.g. hermes-write with source_id=default) could not get_page pages in non-default sources, even when federated_read access was granted. Symptom: page_not_found errors despite the page existing in the DB and being visible via the query-tool.

The query-tool already supported per-call source_id via the --source-id flag (v0.34). This brings get_page into symmetry.

Change:
  - Add `source_id` to params schema (optional, string)
  - effectiveSourceId = p.source_id ?? ctx.sourceId
  - Threads through both the initial getPage lookup and the fuzzy-resolved follow-up.

Regression test: test/get-page-source-id-param.test.ts (5 cases).

Repro (pre-PR):
  # Bearer's ctx.sourceId = default; page lives in priv-karriere-entities
  curl -X POST /mcp -H "Authorization: Bearer <default>" \
    -d '{"method":"tools/call","params":{"name":"get_page", "arguments":{"slug":"X","source_id":"priv-karriere-entities"}}}' # Pre-PR: {"error":"page_not_found"} # Post-PR: returns the page

Backwards compatible: when source_id is unset, behavior matches v0.31.8 (D20).

### DIFF
--- a/src/core/operations.ts
+++ b/src/core/operations.ts
@@ -451,6 +451,7 @@ const get_page: Operation = {
     slug: { type: 'string', required: true, description: 'Page slug' },
     fuzzy: { type: 'boolean', description: 'Enable fuzzy slug resolution (default: false)' },
     include_deleted: { type: 'boolean', description: 'v0.26.5: surface soft-deleted pages with deleted_at populated (default: false). Used by restore workflows.' },
+    source_id: { type: 'string', description: "Scope to a single source (overrides OperationContext.sourceId, consistent with query-tool's --source-id flag)." },
   },
   handler: async (ctx, p) => {
     const slug = p.slug as string;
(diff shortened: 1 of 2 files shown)


time-attack:
Thanks for the contribution, @Valentin32! 🙏

What this does: This PR adds a per-call source_id parameter to the get_page operation, allowing MCP clients to retrieve pages from specific sources. While the use case is real, a more secure implementation already exists in PR #1372 (created earlier the…

Closing this one to keep the queue focused on higher-impact changes. Nothing wrong with the effort — reopen if you'd like to make the case for it and we'll reconsider.

Appreciate your time.

PR #1395, opened 2026-05-24
walker: skip Pods, Carthage, DerivedData (+ register pages in CLI_ONLY)

nubtwerk:
## Summary

Two small fixes the code-sync walker has been missing.

### 1. Skip iOS vendor dirs (Pods / Carthage / DerivedData)

`collectSyncableFiles` already hardcodes skips for `node_modules` and `ops`. It does not honor `.gitignore`. On any React Native / Expo / native iOS project, the `Pods/` directory holds one `.h` per CocoaPods spec, easily into the thousands of files. A real-world sync on my end pulled **8,318 Pod headers** into the brain from a single Expo app, dragging `brain_score` from 76 to 49 (orphan pages 678 → 9,675, link_density score 21 → 3) before I noticed.

These three dirs are universally gitignored in the iOS/macOS ecosystem and never contain user code:

- `Pods` — CocoaPods, regenerated by `pod install` from `Podfile.lock`
- `Carthage` — vendored framework binaries
- `DerivedData` — Xcode build cache

Treating them like `node_modules` matches the existing pattern.

Added a regression test (`test 8` in `sync-walker-symlink.test.ts`) that seeds all three dirs alongside real app code and asserts the walker returns only the real files. Same shape as test 5 for `node_modules` / dot-dirs. All 8 walker tests pass.

### 2. Register `pages` in CLI_ONLY

`src/commands/pages.ts` exists and has a switch case at `cli.ts:1137`, but `pages` was never added to the `CLI_ONLY` set at line 30. Every `gbrain pages <subcommand>` errors out at the top-level dispatch with `"Unknown command: pages"` before reaching the engine-connected switch.

Surfaced when I tried `gbrain pages purge-deleted --older-than 0d` to clean up the leftover soft-deleted pages from the v0.26.5 lifecycle. One-line registration unblocks the operator escape hatch.

## Test plan

- [x] `bun test test/sync-walker-symlink.test.ts` — all 8 tests pass (was 7; new test #8 covers Pods/Carthage/DerivedData)
- [x] Verified end-to-end: `gbrain pages purge-deleted --older-than 0d --dry-run` now works (was: `Unknown command: pages`)
- [x] Confirmed on a real React Native repo: walker no longer descends into `mobile/ios/Pods/`


time-attack:
Merging. Both halves check out against master: `pages` really is absent from the CLI_ONLY set even though `src/commands/pages.ts` and its `case 'pages'` handler already exist, so the command was dead on arrival, and PRUNE_DIR_NAMES had no iOS vendor/build entries — the Pods/Carthage/DerivedData skip is the same shape as the existing `venv` skip from #2020. Thanks for the walker regression test; note the CLI_ONLY line has grown on master since you branched, so expect a rebase there. One small ask: please drop the product name from the test comment before it lands — repo policy keeps real product/company names out of checked-in code.

PR #1396, opened 2026-05-25
fix(cli): support file input for stdin-backed ops

### LINKED ISSUE #1362: `gbrain put <slug>` fails on Windows with ENOENT: /dev/stdin
## Bug Description
`gbrain put <slug>` fails on Windows with `ENOENT: /dev/stdin` error.

## Steps to Reproduce
1. On Windows, run: `echo "some content" | gbrain put test-slug`
2. Observe error

## Expected Behavior
`gbrain put` should read stdin content and create/update the page successfully on all platforms.

## Actual Behavior
```
Error: ENOENT: no such file or directory, open '/dev/stdin'
```

## Root Cause
In `src/cli.ts`, stdin is read via:
```typescript
const stdinContent = readFileSync('/dev/stdin', 'utf-8');
```

`/dev/stdin` does not exist on Windows. This path is Unix-only.

## Proposed Fix
Wrap the stdin read in a try/catch with a Windows fallback using `readSync` on fd 0:

```typescript
// In src/cli.ts, replace:
const stdinContent = readFileSync('/dev/stdin', 'utf-8');

// With:
let stdinContent: string;
try {
  stdinContent = readFileSync('/dev/stdin', 'utf-8');
} catch {
  // Windows fallback: read from stdin fd directly
  const fd = 0;
  const chunks: Buffer[] = [];
  const buf = Buffer.alloc(65536);
  let nread: number;
  while ((nread = readSync(fd, buf, 0, buf.length, null)) > 0) {
    chunks.push(buf.slice(0, nread));
  }
  stdinContent = Buffer.concat(chunks).toString('utf-8');
}
```

Also need to add `readSync` to the import:
```typescript
import { readFileSync, readSync } from 'fs';
```

## Environment
- OS: Windows 11
- gbrain version: 0.39.0
- Runtime: Bun

## Additional Context
This affects all `gbrain put` operations that pipe content via stdin on Windows. The `--file` parameter workaround (using a temp file) works, but stdin piping is a common UX pattern that should work cross-platform.

### LINKED ISSUE #1363: Add --file parameter to bypass Windows pipe buffer limitation

xuezhaolan:
## Summary
- read piped CLI content from fd 0 instead of Unix-only `/dev/stdin`, so stdin-backed ops work on Windows
- add `--file` / `--input-file` for stdin-backed operations such as `gbrain put`, bypassing Windows pipe-buffer limits for large pages
- document the new flag in top-level help and pin argument parsing with tests

Fixes #1362
Fixes #1363

## Tests
- `bun test test/cli-args.test.ts`
- `bun run typecheck`

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -475,6 +475,15 @@ export function resolveQueryImage(
   return { path: imagePath, base64, mime };
 }
 
+const MAX_STDIN_CONTENT_BYTES = 5_000_000; // 5MB
+
+function assertCliContentSize(content: string, source: string): void {
+  if (Buffer.byteLength(content, 'utf-8') > MAX_STDIN_CONTENT_BYTES) {
+    console.error(`Error: ${source} content exceeds ${MAX_STDIN_CONTENT_BYTES} bytes. Split into smaller inputs.`);
+    process.exit(1);
+  }
+}
+
 export function parseOpArgs(op: Operation, args: string[]): Record<string, unknown> {
   const params: Record<string, unknown> = {};
   const positional = op.cliHints?.positional || [];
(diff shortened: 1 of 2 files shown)


time-attack:
Thanks for this — the core of it is already on master. `src/cli.ts` reads piped content from fd 0 rather than `/dev/stdin` as of dab441f5 (v0.41.4.0, #1377, cross-platform stdin), so the Windows failure you describe should be gone on current builds. That leaves the new `--file`/`--input-file` flag, which is added CLI surface we are not taking when `gbrain capture --file PATH` already covers file input. Closing on that basis.

PR #1397, opened 2026-05-25
docs: surface embedding model choice in quickstart

### LINKED ISSUE #1285: Init flag --embedding-model not surfaced in README's Install section; results in silent ZeroEntropy lock-in for sovereignty-focused operators
Hi Garry, first thanks for shipping this. Hybrid search + compiled-truth + timeline is a genuinely clean abstraction over markdown brains.

Filing a doc bug that cost me about 90 minutes on a fresh install.

### Observed

Following the README's **Install** section (Standalone CLI path, `bun add -g gbrain`) and **Setup** section (`gbrain init`), my brain was initialized with an embedding model that I hadn't intentionally chosen. The README's only embedding-related mention is OpenAI text-embedding-3-large in the Prerequisites table, but the install I got actually defaulted to a different provider (ZeroEntropy in my case, since I don't have an OPENAI_API_KEY exported in this environment).

Once the schema is created, the embedding model appears to be baked in. The post-init config command:

```
gbrain config set embedding.model ollama:nomic-embed-text
```

returns success but doesn't change the active model. Re-running `gbrain init` doesn't re-prompt for embedding selection. After some digging I found the actual init flag:

```
gbrain init --embedding-model ollama:nomic-embed-text --embedding-dimensions 768
```

This works correctly and is the only way I've found to lock in a non-default embedding provider at install time.

### Why this matters

For operators choosing gbrain specifically because of its sovereignty posture (markdown-first, library-pluggable, PGLite zero-cloud option), the default init path silently sends embeddings to a third-party cloud provider. Local-only paths (Ollama, llama.cpp llama-server) are advertised as supported recipes but aren't surfaced anywhere in the install flow.

A sovereignty-focused user who follows the README to the letter ends up with cloud egress they didn't intend. That's the opposite of what the rest of the project's design seems to optimize for.

### Suggested fix

In the **Install** or **Setup** section of `README.md`, add an `--embedding-model` paragraph with at least two examples:

```
# OpenAI (default-style cloud, requires OPENAI_API_KEY)
gbrain init --supabase --embedding-model openai:text-embedding-3-large

# Local-only via Ollama (zero egress)
gbrain init --embedding-model ollama:nomic-embed-text --embedding-dimensions 768
```

Optionally list the 14 supported recipes here too, since they're a real differentiator and currently only discoverable by reading source.

### Secondary observation (filed as a note, not a separate ask)

The config key name is inconsistent between the dot-notation accepted at the CLI (`embedding.model`) and what's actually stored under (`embedding_model`, snake_case). The CLI accepts the dot-notation form silently. A short-term mitigation would be making the CLI error on unknown keys; a longer-term fix is settling on one naming convention.

### Environment

- gbrain installed via `bun add -g gbrain` on macOS (Apple Silicon)
- Engine: PGLite (default when no `--supabase` / `--url` provided)
- Discovered while wiring gbrain into a sovereign-local agent substrate; final working init command: `gbrain init --embedding-model ollama:nomic-embed-text --embedding-dimensions 768`

Happy to send a doc-only PR if you'd prefer that route. Let me know.

xuezhaolan:
## Summary
- documents choosing an embedding model during the standalone CLI quickstart
- adds OpenAI and local Ollama init examples
- links to the embedding migration guide for existing brains

Closes #1285

## Test Plan
- git diff --check
- verified docs/embedding-migrations.md exists

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,18 @@ gbrain import ~/notes/   # index your markdown
 gbrain query "what themes show up across my notes?"
 ```
 
+Pick your embedding model during `gbrain init`, before the vector schema is created. For example:
+
+```bash
+# OpenAI cloud embeddings (requires OPENAI_API_KEY)
+gbrain init --pglite --embedding-model openai:text-embedding-3-large
+
+# Local-only embeddings via Ollama (no embedding API egress)
+gbrain init --pglite --embedding-model ollama:nomic-embed-text --embedding-dimensions 768
+```
+
+If you need to switch an existing brain's embedding model or dimensions later, follow [`docs/embedding-migrations.md`](docs/embedding-migrations.md); `gbrain config set embedding_model ...` is not enough because the database vector dimension is fixed at init time.
+
 Postgres-at-scale, Supabase, and thin-client setup paths live in [`docs/INSTALL.md`](docs/INSTALL.md).
 
 ### Connect GBrain to your AI client (MCP)


garrytan:

PR #1398, opened 2026-05-25
fix: drain last_retrieved writes before CLI exit

moeedahmed:
## Summary
- Track pending last_retrieved_at write-backs started by read operations.
- Drain those already-started writes before the local CLI disconnects its engine.
- Add a regression test proving search/get print output and exit cleanly.

## Why
Short-lived CLI commands can print their result but keep the process alive because best-effort last_retrieved_at write-backs are still racing engine disconnect. Callers that wait for process exit then see a hang even though stdout is already present.

## Tests
- bun test test/cli-last-retrieved-lifecycle.test.ts test/search-telemetry.test.ts
- bun run typecheck
- git diff --check
- timeout 5 bun src/cli.ts search 'brain operating contract' --source moeed-brain --limit 1
- timeout 5 bun src/cli.ts get '20-wiki/brain-operating-contract' --source moeed-brain


garrytan:

PR #1400, opened 2026-05-25
fix(gateway): thread input_type to ZE wire body via AsyncLocalStorage

billy-armstrong:
# PR Draft #1: `fix(gateway): thread input_type to ZE wire body via AsyncLocalStorage`

**Branch:** `billy-armstrong/gbrain:fix/ze-input-type-asymmetric` → `garrytan/gbrain:master`
**Files touched:** `src/core/ai/gateway.ts` (+18 / -2)
**Commit:** `732855cd`
**Status:** ready for your review; held off `gh pr create` per your instructions.

---

## Title (for the PR form)

```
fix(gateway): thread input_type to ZE wire body via AsyncLocalStorage
```

## Description

The gateway's data flow for ZE asymmetric encoding goes:

```
embedQuery(text, {embeddingModel, dimensions})
  → embed([text], {inputType: 'query', embeddingModel, dimensions})
  → dimsProviderOptions(impl, modelId, dims, 'query')
  → returns {openaiCompatible: {dimensions, input_type: 'query'}}
  → Vercel AI SDK embedMany(...)             ← input_type DROPPED HERE
  → zeroEntropyCompatFetch(req, init)
  → body parsed: input_type is undefined → default to 'document'
  → POST https://api.zeroentropy.dev/v1/models/embed
     {"model":"zembed-1", "input":[...], "dimensions": N,
      "input_type":"document"}   ← wrong; should be "query"
```

**The break:** the Vercel AI SDK's `openai-compatible` adapter doesn't propagate arbitrary fields from `providerOptions.openaiCompatible` to the request body. Its body schema is fixed (`model`, `input`, `encoding_format`, `dimensions`). `input_type` is silently stripped before `zeroEntropyCompatFetch` receives the request. The shim's "default to `'document'` if undefined" branch then fires, and asymmetric encoding is silently lost on every `embedQuery()` call.

ZE's zembed-1 is specifically distilled for asymmetric retrieval (see [ZE's docs](https://docs.zeroentropy.dev/api-reference/models/embed.md)). Encoding queries as documents collapses retrieval to surface-token overlap — short fragments rank above semantically-correct chunks.

## Fix

Thread `inputType` through `AsyncLocalStorage` so `zeroEntropyCompatFetch` can read it at body-rewrite time. Mirrors the existing `__budgetStore` pattern in the same file.

Three edits to `gateway.ts`:

1. **Declare the store** next to `__budgetStore`:
   ```ts
   export const __embedInputTypeStore = new AsyncLocalStorage<'query' | 'document'>();
   ```

2. **Wrap the SDK call** in `embed()`:
   ```ts
   const result = await __embedInputTypeStore.run(opts?.inputType ?? 'document', () =>
     embedSubBatch(batch, model, providerOpts, expected, recipe, modelId, opts),
   );
   ```

3. **Read the store** in `zeroEntropyCompatFetch`:
   ```ts
   if (parsed.input_type === undefined) {
     parsed.input_type = __embedInputTypeStore.getStore() ?? 'document';
     mutated = true;
   }
   ```

**Document-side path preserved by construction.** When `opts.inputType` is undefined (the common ingest case via plain `embed()`), the store carries `'document'` and the shim reads the same default — embed-side wire bodies are byte-identical pre/post-patch.

## Reproduction (any gbrain checkout with ZE configured)

A ~50-line script that patches `global.fetch` and logs the request body before transmission. The full script lived at `/usr/local/bin/gbrain-ze-wire-probe.ts` on the test host; here's the essential shape:

```ts
import { configureGateway, embed, embedQuery } from "<gbrain>/src/core/ai/gateway.ts";
import { buildGatewayConfig } from "<gbrain>/src/cli.ts";
import { loadConfig } from "<gbrain>/src/core/config.ts";

configureGateway(buildGatewayConfig(loadConfig()));

const origFetch = globalThis.fetch;
globalThis.fetch = async (input, init) => {
  const url = typeof input === "string" ? input : input.toString();
  if (url.includes("zeroentropy") && typeof init?.body === "string") {
    console.log("WIRE:", init.body);
  }
  return origFetch(input, init);
};

// Run either of these; observe the wire body.
await embedQuery("sample query", {
  embeddingModel: "zeroentropyai:zembed-1",
  dimensions: 2560,
});
// Pre-fix wire body:
//   {"model":"zembed-1","input":[...],"encoding_format":"float",
//    "dimensions":2560,"input_type":"document"}
// Post-fix wire body:
//   {"model":"zembed-1","input":[...],"encoding_format":"float",
//    "dimensions":2560,"input_type":"query"}

await embed(["sample doc"], {
  embeddingModel: "zeroentropyai:zembed-1",
  dimensions: 2560,
});
// Both pre-fix and post-fix:
//   {"model":"zembed-1","input":[...],"encoding_format":"float",
//    "dimensions":2560,"input_type":"document"}
```

## Empirical retrieval-quality impact

Tested on a ~17,472-chunk personal brain (long-markdown wikis + transcripts). Two natural-language queries through `mcp__gbrain__query` with `embedding_column` switched between OpenAI and ZE columns:

| Query | OpenAI (768d, HNSW) top-1 | ZE (2560d) top-1 PRE-FIX | ZE (2560d) top-1 POST-FIX |
|---|---|---|---|
| "what conventions do we have around brain-first lookups" | long compiled_truth chunk about MCP routing + GBrain Search Guidance | tiny `[Bash] fence.sh:1-1` chunk | same long compiled_truth chunk OpenAI returns |
| "what's the current Cloudflare mesh setup at home" | `wiki/hubs/cloudflare-mesh` (canonical hub) | tiny `[Bash] ssh server 'which cloudflared'` | `wiki/hubs/cloudflare-mesh` (same as OpenAI) |

ZE-2560 + the existing zerank-2 reranker is now competitive with OpenAI text-embedding-3-large-768 on this brain. Pre-fix, ZE was a regression (surface-token matches beat real semantic answers).

## Caveats

- **AsyncLocalStorage propagation:** the store propagates through `await`/`then` chains, NOT through `Worker` threads or callbacks scheduled outside the current async context. The current AI SDK `openai-compatible` adapter preserves the context; a future SDK change to worker-based batching would silently revert this fix. The document-side default-to-`'document'` fallback would still be correct in that case, so the failure mode degrades to "worse retrieval" not "broken embeddings." Worth noting in the AI SDK changelog if a maintainer is reading.

- **Voyage v3+ may have the same bug class.** Source inspection of `voyageCompatFetch` shows no `input_type` handling at all — the shim only rewrites `encoding_format` (forces base64) and `dimensions → output_dimension`. It does not inject or read `input_type`, relying on Voyage's own API default of `document`. If the AI SDK's openai-compatible adapter is also stripping `providerOptions.openaiCompatible.input_type` on the Voyage path (same code path), Voyage's `embedQuery()` would land as `input_type: "document"` too. I did NOT run the wire-probe against Voyage — out of scope for this PR — but the source pattern is symmetric. Same fix applies if confirmed empirically (read from `__embedInputTypeStore` in `voyageCompatFetch`'s body rewriter alongside the existing `encoding_format` / `dimensions` mutations).

## Sibling note for the maintainer

The wire-probe pattern (one-off `global.fetch` shim) is the cheapest discriminator for any future "this third-party provider's results look wrong" bug class. If the AI SDK shape changes again, the same probe surfaces the gap immediately.

### DIFF
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -967,6 +967,19 @@ const zeroEntropyCompatFetch = (async (input: RequestInfo | URL, init?: RequestI
           parsed.encoding_format = 'float';
           mutated = true;
         }
+        // PATCH:ze-input-type-asymmetric v2 — query-side threading via
+        // __embedInputTypeStore (populated by embed() so opts.inputType
+        // reaches the wire body even after the AI SDK's openai-compatible
+        // adapter strips providerOptions). Runs BEFORE the document-default
+        // so the literal `parsed.input_type = 'document'` line below stays
+        // intact (structural test asserts it remains in source).
+        if (parsed.input_type === undefined) {
+          const threadedType = __embedInputTypeStore.getStore();
+          if (threadedType !== undefined) {
+            parsed.input_type = threadedType;
+            mutated = true;
+          }
+        }
         // Default input_type when caller didn't thread one (document-side
         // embedding is the correct default for ingest paths).
         if (parsed.input_type === undefined) {
@@ -1284,7 +1297,11 @@ export async function embed(texts: string[], opts?: EmbedOpts): Promise<Float32A
   let _embedThrew = false;
   try {
     for (const batch of batches) {
-      const result = await embedSubBatch(batch, model, providerOpts, expected, recipe, modelId, opts);
+      // PATCH:ze-input-type-asymmetric — propagate opts.inputType to the
+      // wire body via __embedInputTypeStore so the ZE shim picks it up.
+      const result = await __embedInputTypeStore.run(opts?.inputType ?? 'document', () =>
+        embedSubBatch(batch, model, providerOpts, expected, recipe, modelId, opts),
+      );
       allEmbeddings.push(...result);
     }
     return allEmbeddings;
@@ -2123,6 +2140,16 @@ export async function generateOcrText(imageBytes: Buffer, mime: string): Promise
 // budget throw → no lease attempted, no rate-lease window held.
 
 const __budgetStore = new AsyncLocalStorage<BudgetTracker>();
+
+// PATCH:ze-input-type-asymmetric v2
+// Threads `input_type: 'query' | 'document'` from embed()/embedQuery()
+// through to zeroEntropyCompatFetch's body-rewriter. The Vercel AI SDK's
+// openai-compatible adapter strips providerOptions.openaiCompatible.
+// input_type from the wire body before our fetch shim sees it, so the
+// shim's `if (parsed.input_type === undefined)` default was always
+// firing — breaking ZE's asymmetric query/document encoding contract.
+// (Sibling to __budgetStore. v0.42+ upstream PR target.)
+export const __embedInputTypeStore = new AsyncLocalStorage<'query' | 'document'>();
 
 export function withBudgetTracker<T>(tracker: BudgetTracker, fn: () => Promise<T>): Promise<T> {
   return __budgetStore.run(tracker, fn);


billy-armstrong:
Reopening: branch was briefly synced to master during the v1 → v2 rebase, which auto-closed the PR. v2 shape is now pushed at `838a0a9a` and CI will pick it up.

PR #1401, opened 2026-05-25
fix(longmemeval): strip provider prefix before passing to Anthropic SDK

billy-armstrong:
# PR Draft #2: `fix(longmemeval): strip provider prefix before passing to Anthropic SDK`

**Branch:** `billy-armstrong/gbrain:fix/longmemeval-strip-provider-prefix` → `garrytan/gbrain:master`
**Files touched:** 4 files, +43 / -25
**Commit:** `dfa56801`
**Status:** ready for your review; held off `gh pr create` per your instructions.

---

## Title (for the PR form)

```
fix(longmemeval): strip provider prefix before passing to Anthropic SDK
```

## Description

`resolveModel()` returns qualified strings like `anthropic:claude-sonnet-4-6` so internal callers can route to the right provider. But provider-specific SDKs only accept bare model ids — the Anthropic SDK rejects the prefixed form with HTTP 404:

```json
{"type":"error","error":{"type":"not_found_error","message":"model: anthropic:claude-sonnet-4-6"}}
```

`stripProviderPrefix()` was originally added during the v0.41 Bug 3 fix to solve this in the subagent path. longmemeval has two more SDK-direct call sites with the same bug:

- `src/commands/eval-longmemeval.ts:354` — answer-gen
- `src/eval/longmemeval/extract.ts:257` — trajectory extractor (v0.40.2.0)

Both hit the same 404 on every call. Result: `gbrain eval longmemeval` returns `hypothesis: ""` and `error: "..."` for every question.

## Fix

Three changes:

1. **Move `stripProviderPrefix` to `src/core/model-config.ts`** — sibling to `resolveModel`, `isAnthropicProvider`. Same implementation, same JSDoc. Canonical home for model-string utilities.

2. **`subagent.ts`** — import from the new home; re-export for back-compat:
   ```ts
   import { ..., stripProviderPrefix } from '../../model-config.ts';
   export { stripProviderPrefix };  // preserves test/subagent-handler.test.ts import path
   ```

3. **`eval-longmemeval.ts` + `extract.ts`** — import + apply at both SDK call sites:
   ```ts
   const response = await client.create({
     model: stripProviderPrefix(model),
     // ...
   });
   ```

## Reproduction

```sh
gbrain eval longmemeval test/fixtures/longmemeval-nightly.jsonl \
  --output /tmp/out.jsonl --by-type --limit 2
head -2 /tmp/out.jsonl
```

**Pre-fix output** (every row):
```json
{"question_id":"nightly-1", ..., "hypothesis":"",
 "error":"404 {\"type\":\"error\",\"error\":{\"type\":\"not_found_error\",
          \"message\":\"model: anthropic:claude-sonnet-4-6\"}}"}
```

**Post-fix output** (real LLM-generated answers for every question):
```json
{"question_id":"nightly-1","question":"Where did alice-example say she lived last year?",
 "hypothesis":"Based on the retrieved session, **alice-example** said she lived in
                **widget-co** last year.",
 "recall_hit":true, ...}
```

10/10 questions return real hypotheses on the public `longmemeval-nightly.jsonl` fixture.

## Test coverage

`test/subagent-handler.test.ts` imports `stripProviderPrefix` from `subagent.ts` and exercises:

- Strips `anthropic:` prefix
- Idempotent on bare names
- Edge inputs: empty string, lone colon, multi-colon

These tests continue to pass via the re-export. No new tests added — the existing suite verifies the helper's contract; this PR is "apply the helper at two more sites."

## Why this stayed broken

The helper lived in `src/core/minions/handlers/subagent.ts` (a handler-specific file) and was never imported outside that file. When longmemeval grew two new direct-SDK call sites in v0.40.1.0 and v0.40.2.0, the helper wasn't discoverable from the canonical location, so the same fix wasn't applied. Moving it to `model-config.ts` (next to `resolveModel`) puts the helper where future contributors will look for it.

## Cost recovery

Anyone running `gbrain eval longmemeval` on a fresh install was burning embedding spend (longmemeval embeds the fixture's haystack sessions into PGLite before each question) and getting zero answer-gen tokens of value. Post-fix, the same embedding spend produces real hypotheses + cross-modal verdicts.

### DIFF
--- a/src/commands/eval-longmemeval.ts
+++ b/src/commands/eval-longmemeval.ts
@@ -17,7 +17,7 @@ import { renderChatBlock, type ChatSessionForPrompt } from '../eval/longmemeval/
 import { importFromContent } from '../core/import-file.ts';
 import { hybridSearch } from '../core/search/hybrid.ts';
 import { expandQuery } from '../core/search/expansion.ts';
-import { resolveModel } from '../core/model-config.ts';
+import { resolveModel, stripProviderPrefix } from '../core/model-config.ts';
 import type { ThinkLLMClient } from '../core/think/index.ts';
 import { createProgress } from '../core/progress.ts';
 import { getCliOptions, cliOptsToProgressOptions } from '../core/cli-options.ts';
@@ -351,8 +351,12 @@ async function generateAnswer(
   const userText =
     `Question:\n${question}\n\n${trajectorySection}Retrieved sessions:\n${rendered}`;
 
+  // resolveModel() returns 'anthropic:claude-sonnet-4-6' with the provider
+  // prefix, but the Anthropic SDK only accepts bare model ids. Without
+  // this strip, every answer-gen call HTTP 404s. Mirrored at the
+  // trajectory-extractor SDK call site in extract.ts.
   const response = await client.create({
-    model,
+    model: stripProviderPrefix(model),
     max_tokens: 512,
     system: systemText,
     messages: [{ role: 'user', content: userText }],
--- a/src/core/minions/handlers/subagent.ts
+++ b/src/core/minions/handlers/subagent.ts
@@ -47,7 +47,7 @@ import {
   logSubagentSubmission,
   logSubagentHeartbeat,
 } from './subagent-audit.ts';
-import { resolveModel, isAnthropicProvider, TIER_DEFAULTS } from '../../model-config.ts';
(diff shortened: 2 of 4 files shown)


billy-armstrong:
Superseded by #1681 — that PR fixes the root cause by routing longmemeval's answer-gen and extractor through the gateway (`tryBuildGatewayClient`, the canonical chat seam `think`/`synthesize` already use) instead of stripping the provider prefix at the call site. It also covers the extractor path this PR didn't, and is verified end-to-end. Closing in favor of #1681.

PR #1408, opened 2026-05-25
fix(postgres-engine): pair pg_advisory_lock acquire+release on same backend via pool.reserve()

rayers:
## Problem

`PostgresEngine.initSchema()` acquires advisory lock 42 and later releases it via bare tag-template calls on the pool. postgres.js pulls a fresh physical connection from the pool on each call, so in **single-pool mode** (`conn = this.sql`) the acquire lands on backend A and the unlock fires on backend B — which silently returns `false` ("not held by current session"). Backend A's session then sits idle in the pool still holding lock 42 indefinitely.

Every subsequent `initSchema()` blocks on `SELECT pg_advisory_lock(42)` until `statement_timeout` fires. Symptom: **CLI cold-start hangs ~5 min** on every command that connects to the engine. Diagnosed against a real 440K-page Postgres brain where one backend had held lock 42 idle for 14h since the autopilot worker's first `initSchema()` call; killing the stale backend dropped cold-start from 5+ min to 1.28s.

## Fix

`pool.reserve()` pins the lock acquire + release pair (and the threaded bootstrap + `SCHEMA_SQL` replay) to the same physical connection, so the unlock actually fires on the session holding the lock. Same pattern `executeRaw` uses elsewhere in this file. `runMigrations`/`verifySchema`/`dropZombieIndexes` keep their own connection management. Postgres-only (the PGLite half is #1337), so engine-parity is unaffected.

## Test

`test/e2e/postgres-initschema-advisory-lock.test.ts` — asserts (1) after one `initSchema()`, lock 42 is held by no backend; (2) two consecutive `initSchema()` calls each complete in <30s (pre-fix the second blocked until statement_timeout).

---
*Rebased onto current `master` (post-`f3ade6c0` ownership-token + `ec5fed29` reconnect rework). The split-backend bug persists in today's `initSchema`, so the fix still applies.*

### DIFF
--- a/src/core/postgres-engine.ts
+++ b/src/core/postgres-engine.ts
@@ -271,9 +271,8 @@ export class PostgresEngine implements BrainEngine {
     // is in dual-pool mode. The pooler's 2-min statement_timeout truncates
     // SCHEMA_SQL replays + migrations on Supabase; the direct pool gets
     // 30min. Lane B replaces the lock primitive with a TTL+heartbeat table
-    // lock; Lane A does the routing and keeps pg_advisory_lock(42) on the
-    // SAME connection so the lock is correct.
-    const conn = this.connectionManager
+    // lock; Lane A does the routing.
+    const pool = this.connectionManager
       ? await this.connectionManager.ddl()
       : this.sql;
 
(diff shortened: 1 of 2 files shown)


garrytan:

PR #1414, opened 2026-05-25
feat(cycle): generalize extract_atoms — page-based extraction + config-plane fallback

garrytan-agents:
## Problem

`extract_atoms` (v0.41 T5) has two issues that prevent it from working on production brains:

### 1. Config-plane mismatch
The phase reads `dream.synthesize.session_corpus_dir` via `loadConfig()`, which only reads the file plane (`~/.gbrain/config.json` + env vars). But `gbrain config set dream.synthesize.*` writes to the DB config table. Result: config appears set but the phase never sees it — silently returns "no transcripts to process."

### 2. Transcript-only design
The phase ONLY processes raw files from filesystem corpus directories. A brain with thousands of existing pages (meetings, sources, articles) has no extraction path — all that content sits in the DB but `extract_atoms` can't reach it.

### Fix 1: `resolveConfigStr()` helper
Walks file config first, falls through to `engine.getConfig(key)` for DB-plane. Matches precedence model used by `loadConfigWithEngine()` elsewhere.

### Fix 2: Page-based discovery
`discoverExtractablePages()` queries `engine.listPages()` for extractable types (`meeting`, `source`, `article`, `video`, `book`, `original`). Filters out already-processed pages (`atoms_extracted: true`). Caps at 50 pages/cycle.

### Fix 3: Dual-source merge
Transcript + page discovery run together. Deduplicated by content hash. Atom frontmatter distinguishes origin (`source_path` for transcripts, `source_slug` for pages). Source pages get `atoms_extracted: true` after extraction.

## Testing
13 new tests — all pass first-green. Covers: page extraction, config fallback, dedup, budget cap, skip-already-extracted, skip-greenfield, skip-short, metadata correctness, backward compat.

### DIFF
--- a/src/core/cycle/extract-atoms.ts
+++ b/src/core/cycle/extract-atoms.ts
@@ -1,48 +1,55 @@
-// v0.41 T5 — extract_atoms cycle phase (minimal-viable implementation).
+// v0.41 T5 — extract_atoms cycle phase.
 //
-// v0.41 ships a working Haiku-driven extract path. The full v0.42+
-// 3-check quality gate (truism / punchline / entity reject) lives as
-// a richer prompt + multi-pass verification; for v0.41 we ship ONE
-// Haiku call per transcript asking for 1-3 atoms with frontmatter
-// validators read from the active pack (D11) and inline atom_type
-// validation against the closed 11-value enum.
+// v0.41.2.1: generalized to extract from BOTH raw transcripts (filesystem
+// corpus dirs) AND existing brain pages (DB query by type). Fixes the
+// config-plane mismatch where `gbrain config set dream.synthesize.*`
+// writes to DB but `loadConfig()` only reads file-plane config.
+//
+// Content sources (checked in order, results merged):
+//   A. Raw transcripts via discoverTranscripts (corpus dirs from config).
+//      Config keys: dream.synthesize.session_corpus_dir,
+//                   dream.synthesize.meeting_transcripts_dir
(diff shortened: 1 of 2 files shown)


garrytan:
Closing in favor of fix-wave incorporation on garrytan/pr-1414-1416-1421 (shipping as v0.41.2.1). Rebuilt with /plan-eng-review + codex outside-voice structural improvements:

- **Source-hash existence check** (idempotency-via-data; survives `gbrain sync --force`; also fixes the pre-existing transcript-side date-stamp duplicate bug) replaces the frontmatter `atoms_extracted: true` marker.
- **Single raw SQL with NOT EXISTS subquery** (one round-trip) replaces the 6 `listPages` iterations + per-candidate atom-existence checks.
- **sourceId threaded through every atom putPage call** (codex caught a real bug here that would have routed non-default-source atoms to 'default').
- **Discovery SQL filters out `dream_generated:true`** to prevent self-consumption, **`content_hash IS NOT NULL`** to prevent crashes, and **`imported_from='markdown-greenfield'`** as before.
- **cycle.ts threads UNION of syncPagesAffected + synthesizeWrittenSlugs** as affectedSlugs (incremental cycles now see pages synthesize just wrote).
- The systemic `dream.*` config merge from #1416 obviates the per-call-site `resolveConfigStr()` helper.

Thank you for the careful PR description and the 13 tests — they were the design source for the rebuild. The wave ships with full credit on the merged commit.

Issue #1413, opened 2026-05-25
autopilot: silent link-row loss + getHealth() breakage from module-singleton null mid-cycle (separate class from #1162 / #465)


archerhpagent:
## Summary

On Postgres engine, the `autopilot --inline` daemon silently loses link rows during the `extract.links_fs` phase, and `engine.getHealth()` then throws `No database connection: connect() has not been called` every cycle. Affects v0.41.0.0 (current at time of filing). Not the same class as #1162 / PR #465 (which is about the reconnect loop crashing the daemon) — this fires even when the daemon stays alive.

## Repro

Stock v0.41.0.0 on a Postgres-engine brain. From an interactive shell with your `GBRAIN_DATABASE_URL` / `ZEROENTROPY_API_KEY` etc. loaded:

```bash
gbrain dream --dir /path/to/your/brain --json 2>&1 | tee /tmp/dream.log
```

Output around the `extract` phase will contain:

```
[cycle.extract] start
[extract.links_fs] 3/3 (100%)
  batch error (2 link rows lost): No database connection: connect() has not been called. Fix: Run gbrain init --supabase or gbrain init --url <connection_string>
[extract.links_fs] 3/3 (100%) done
```

The cycle continues, every subsequent phase that calls `this.sql` and falls back to the module-level `db.getConnection()` will fail the same way (in autopilot's main loop this surfaces as `[health] ERROR ...` after the cycle completes, but the daemon does not crash — `logError` catches it).

## Root cause

`PostgresEngine.connect(config)` without a `poolSize` arg goes to the module-level singleton branch (`postgres-engine.ts:175-189`) — `db.connect(config)` runs, but the engine instance's own `_sql` stays `null`. From then on, `this.sql` returns `db.getConnection()` (the module singleton).

Somewhere in the cycle's call graph (still narrowing — observed firing during the `extract.links_fs` phase, after `synthesize` completes — even on a brain with no transcripts so synthesize is a near-noop), the module-level `sql` in `db.ts` gets nulled. Once that happens, every later engine call routed through the getter throws `connect() has not been called`. The lost-rows behavior in `extract.ts:614-628` is structural — the `flush()` function catches and logs the throw, but `batch.length = 0` in the `finally` clears the un-inserted rows regardless.

This is the same bug class the source comment at `src/commands/auth.ts:52` acknowledges for auth commands:

[...]

PR #465, opened 2026-04-27
fix(autopilot): use engine.reconnect() so DB health check doesn't wipe saved config

jonahberg:
## The bug

Autopilot's per-cycle DB health check (autopilot.ts:248-255) tries to recover from a transient connection failure with:

\`\`\`ts
try {
  await engine.getConfig('version');
} catch {
  try {
    await engine.disconnect();
    await (engine as any).connect?.();   // ← no arguments
  } catch (e) { logError('reconnect', e); }
}
\`\`\`

But \`PostgresEngine.connect\` requires a config:

\`\`\`ts
async connect(config: EngineConfig & { poolSize?: number }): Promise<void> {
  this._savedConfig = config;          // overwrites with undefined
  if (config.poolSize) {                // → \"undefined is not an object (evaluating 'config.poolSize')\"
\`\`\`

So every transient health blip:

1. Wipes \`_savedConfig\` (corrupting state for future supervisor reconnects too).
2. Throws \`undefined is not an object (evaluating 'config.poolSize')\`.
3. Logs \`[reconnect] ERROR ...\`.
4. Leaves the engine without a connection. The next dispatch fails with \`No database connection: connect() has not been called\`.
5. Cycle marks failed. After 5 in a row: \`Autopilot stopping (cycle-failure-cap)\`. launchd respawns ~10 s later. Repeat.

The buggy pattern dates to v0.10.1 (b7e3005), well before v0.22.1's #406 added \`reconnect()\`. It was silent because the catch path only fires on real connection drops; on a healthy pool you never see it.

## Real-world impact (one operator's data)

In ~8 days on a Supabase free-tier brain:

- **54** \`Autopilot stopping (cycle-failure-cap)\` exits.
- **~85** cascading \`Errors.postgres\` parse errors.
- **~70** \`MaxClientsInSessionMode\` pool-saturation errors as orphaned backends piled up between launchd respawns.
- ~2,488 cascading per-page embed failures (separate root cause, but they only kept happening because autopilot kept dying mid-cycle).

Every Supavisor pooler keepalive eviction tripped this code and broke the daemon until launchd respawned it.

## The fix

Route recovery through \`engine.reconnect()\` (added in v0.22.1 #406), which uses the saved config instead of dropping it.

\`\`\`ts
try {
  await engine.getConfig('version');
} catch {
  try {
    if (typeof (engine as any).reconnect === 'function') {
      await (engine as any).reconnect();
    } else {
      await engine.disconnect();
    }
[...]

PR #1415, opened 2026-05-25
fix(autopilot): switch main engine to instance-owned pool so mid-cycle singleton-null doesn't lose rows + break getHealth

archerhpagent:
## The bug

Companion to issue #1413.

On Postgres engine, `gbrain autopilot` silently loses link rows during the `extract.links_fs` phase, and `engine.getHealth()` post-cycle throws `No database connection: connect() has not been called` on every tick.

Root cause: `connectEngine()` in cli.ts calls `engine.connect(config)` without a `poolSize` arg, which routes `PostgresEngine.connect()` to the module-level singleton branch (`postgres-engine.ts:175-189`). The engine's own `_sql` stays null and `this.sql` falls back to `db.getConnection()`. Something in the cycle's call graph nulls that singleton mid-tick, after which every later `this.sql` access throws.

Confirmed reproducible on v0.41.0.0 with `gbrain dream --dir <repo> --json`. The log shows:

```
[cycle.extract] start
[extract.links_fs] 3/3 (100%)
  batch error (2 link rows lost): No database connection: connect() has not been called. Fix: Run gbrain init --supabase or gbrain init --url <connection_string>
[extract.links_fs] 3/3 (100%) done
```

The reported `extracted=N` count in the human log line looks healthy because it reflects intent, not actual rows persisted. `extract.ts:614-628`'s `flush()` catches the throw, logs the line, and `batch.length = 0` in the `finally` clears the un-inserted rows.

Same bug class as the source comment at `src/commands/auth.ts:52` — auth-command instance was fixed by having `withConfiguredSql` call `engine.connect()`; this PR closes the autopilot instance.

## The fix

Re-call `engine.connect({ ...savedConfig, poolSize: 5 })` at autopilot startup so the engine moves to instance-owned pool (`postgres-engine.ts:127-174`). With `poolSize` set, `_sql` is created and the getter returns it directly; the module singleton is no longer load-bearing for this daemon.

PGLite engines are unaffected — `poolSize` is ignored on that engine.

The switch is gated on `engine.kind === 'postgres'` and skipped when `_sql` is already non-null (a future caller could wire the engine with `poolSize` upstream of autopilot; the guard avoids a wasted reconnect). The connect call is wrapped in try/catch so a transient pool-init blip logs and continues rather than killing the supervisor.

```diff
+  if (engine.kind === 'postgres') {
+    const savedCfg = (engine as unknown as { _savedConfig?: EngineConfig })._savedConfig;
+    const hasInstancePool = (engine as unknown as { _sql?: unknown })._sql != null;
+    if (savedCfg && !hasInstancePool) {
+      try {
+        await engine.connect({ ...savedCfg, poolSize: 5 } as EngineConfig & { poolSize: number });
+      } catch (e) {
+        console.error(`[autopilot] could not switch engine to instance pool: ${(e as Error).message}`);
+      }
+    }
+  }
```

Private-field access goes through a narrow structural cast rather than `as any` so the rest of the block keeps its type coverage. If you'd prefer a public method on `BrainEngine` (e.g. `ensureInstancePool(size: number)`), happy to split that into a follow-up — the broader interface change felt out of scope for the first fix.

## Tests

New `test/autopilot-instance-pool.test.ts` follows the static-shape style of `test/autopilot-supervisor-wiring.test.ts` (7 cases): kind gate, `_savedConfig` read shape, `_sql !=  null` guard, `poolSize: 5` cast shape, try/catch wrap, source-ordering before mode/spawn, import of `EngineConfig`. Full behavioral test would need a Postgres fixture; the static guards catch refactor regressions cheaply.

All 94 existing autopilot tests still pass.

## Verified locally

Applied to v0.41.0.0 against a Postgres-engine brain. Every cycle phase runs end-to-end (16 phases including extract_facts / consolidate / propose_takes / grade_takes / calibration_profile / orphans / schema_suggest / purge). No `batch error` lines. No `[health] ERROR` lines. `engine.getHealth()` returns a real `brain_score`. `extract.links_fs` writes real rows.

## Relationship to PR #465

PR #465 (open, ~1 month no review) fixes the `(engine as any).connect?.()` no-config-arg bug in the autopilot reconnect path. It's a real fix and lands when this lands. But #465 alone leaves THIS bug fully present — the daemon stays alive (good — that's what #465 ships) but every cycle still loses rows and `getHealth` still throws.

Independent fixes; can land in either order.

### DIFF
--- a/src/commands/autopilot.ts
+++ b/src/commands/autopilot.ts
@@ -21,6 +21,7 @@ import { existsSync, readFileSync, writeFileSync, mkdirSync, appendFileSync, uti
 import { join } from 'path';
 import { execSync } from 'child_process';
 import type { BrainEngine } from '../core/engine.ts';
+import type { EngineConfig } from '../core/types.ts';
 import { loadPreferences } from '../core/preferences.ts';
 import { loadConfig, gbrainPath as gbrainHomePath } from '../core/config.ts';
 import { ChildWorkerSupervisor } from '../core/minions/child-worker-supervisor.ts';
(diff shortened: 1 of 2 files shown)


archerhpagent:
Closing this — upstream v0.41.x autopilot.ts already eliminates both `(engine as any).connect?.()` and the module-singleton-fallback issue this PR targeted. Upgraded my archer install from v0.41.0.0 → v0.41.11.1 today and confirmed: no batch-error rows, no [health] ERROR lines, cycles run cleanly. The narrow autopilot-only fix this PR proposed is no longer needed. Thanks for the cleanup wave.

archerhpagent:
Superseded by upstream v0.41.x autopilot.ts rewrites. Confirmed end-to-end clean cycles on v0.41.11.1.

PR #1416, opened 2026-05-25
fix: batch retry for PgBouncer connection drops + dream.* DB config merge

garrytan-agents:
## Problem

Two reliability issues surfaced during v0.41.2.0 upgrade testing on a production brain (~96K pages, Supabase PgBouncer on port 6543):

### 1. ~30% batch data loss during extract cycles
The `extract` command's 6 `flush()` functions call `engine.addLinksBatch()` / `engine.addTimelineEntriesBatch()` inside a try/catch. When PgBouncer recycles a backend connection between queries, the batch throws `No database connection: connect() has not been called` and the entire batch (100 rows) is silently dropped. On a 15K-page brain, this lost ~4,500 link rows and ~3,000 timeline rows per cycle.

### 2. `dream.*` config invisible to cycle phases
`loadConfigWithEngine()` merges DB-stored config on top of file-plane config for `embedding_*` and `content_sanity.*`, but NOT for `dream.synthesize.*` or `dream.patterns.*`. So `gbrain config set dream.synthesize.session_corpus_dir /path` writes successfully to DB, `gbrain config get` reads it back, but cycle phases calling `loadConfig()` (or even `loadConfigWithEngine()`) never see it. The `extract_atoms` phase silently skips with "no transcripts to process."

## Error Log
```
[extract.links_fs] 2686/15767 (17%)
  batch error (100 link rows lost): No database connection: connect() has not been called.
[extract.links_fs] 2844/15767 (18%)
  batch error (100 link rows lost): No database connection: connect() has not been called.
...
[extract.timeline_fs] 15642/15767 (99%)
  batch error (100 timeline rows lost): No database connection: connect() has not been called.
```
Pattern: intermittent, affects 20-40% of batches during heavy cycles. More frequent at higher page counts (connection pool pressure).

### Fix 1: `withRetry()` helper for batch inserts
New function catches connection-class errors (`No database connection`, `connect()`, `Connection terminated`) and retries once after 500ms. Non-connection errors (constraint violations, etc.) are NOT retried — they propagate immediately.

Applied to all 6 `flush()` functions in `extract.ts`. The batch array is snapshot-copied before clearing, so the retry operates on the same data.

```typescript
async function withRetry<T>(fn: () => Promise<T>, label: string, jsonMode: boolean): Promise<T> {
  try { return await fn(); }
  catch (firstErr) {
    if (!isConnectionError(firstErr)) throw firstErr;
    await new Promise(r => setTimeout(r, 500));
    return await fn(); // one retry
  }
}
```

### Fix 2: `dream.*` DB config merge in `loadConfigWithEngine()`
Added 7 `dream.*` keys to the sparse-merge block:
- `dream.synthesize.session_corpus_dir`
- `dream.synthesize.meeting_transcripts_dir`
- `dream.synthesize.verdict_model`
- `dream.synthesize.max_prompt_tokens`
- `dream.synthesize.max_chunks_per_transcript`
- `dream.patterns.lookback_days`
- `dream.patterns.min_evidence`

Same precedence as all other merged keys: file/env wins per key; DB fills gaps.

## Testing
5 new tests for `withRetry`: pass-through, retry on connection errors, retry on "Connection terminated", no retry on constraint violations, propagation on double failure. All pass.

## Companion PR
This is the companion to #1414 (generalize extract_atoms — page-based extraction). That PR adds `resolveConfigStr()` as a per-call-site fallback; this PR fixes the systemic gap in `loadConfigWithEngine()` so all callers benefit.

### DIFF
--- a/src/commands/extract.ts
+++ b/src/commands/extract.ts
@@ -37,6 +37,37 @@ import { pathToSlug, pruneDir, isSyncable } from '../core/sync.ts';
 // small (a malformed row aborts at most 100, not thousands).
 const BATCH_SIZE = 100;
 
+/**
+ * Retry a batch operation once after a short delay. PgBouncer transaction-mode
+ * poolers can recycle backend connections between queries, which surfaces as
+ * "No database connection: connect() has not been called" when the postgres.js
+ * client's internal slot was returned to the pool mid-batch. A single retry
+ * after 500ms lets the pool recover a fresh backend.
+ *
+ * v0.41.2.1: added after observing ~30% batch loss during heavy extract cycles
+ * on production brains with Supabase PgBouncer (port 6543).
+ */
+async function withRetry<T>(
+  fn: () => Promise<T>,
+  label: string,
+  jsonMode: boolean,
+): Promise<T> {
+  try {
+    return await fn();
+  } catch (firstErr) {
+    const msg = firstErr instanceof Error ? firstErr.message : String(firstErr);
+    // Only retry on connection-class errors, not constraint violations etc.
+    if (!msg.includes('No database connection') && !msg.includes('connect()') && !msg.includes('Connection terminated')) {
+      throw firstErr;
+    }
+    if (!jsonMode) {
+      process.stderr.write(`  ${label}: connection lost, retrying in 500ms...\n`);
+    }
+    await new Promise((r) => setTimeout(r, 500));
+    return await fn(); // second attempt — if this throws, caller catches it
+  }
+}
+
 // --- Types ---
 
 export interface ExtractedLink {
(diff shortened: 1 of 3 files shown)


garrytan:
Closing in favor of fix-wave incorporation on garrytan/pr-1414-1416-1421 (shipping as v0.41.2.1). Rebuilt with /plan-eng-review structural improvements:

- **withRetry is a pure exported primitive with an `onRetry` callback** (testable in isolation) plus a shared `logBatchRetry()` helper for the 6 flush sites.
- **Classifier is the existing `isRetryableConnError` from `src/core/retry-matcher.ts`** (8-pattern production-grade, extended with the typed `GBrainError{problem:'No database connection'}` shape) rather than the inline 3-substring match.
- **Snapshot-before-clear contract** preserved per your PR description and pinned by a forced-mutation regression test.
- **dream.* DB-config merge** landed in `loadConfigWithEngine()` as you proposed; precedence is `file > DB > defaults` (no env layer since there are no GBRAIN_DREAM_* env vars — claim updated in CHANGELOG to match reality).

Credit on the merged commit.

PR #1417, opened 2026-05-25
fix(lint): only unwrap whole-page code fences, not mid-document blocks

trinh-macbook:
## Problem

`gbrain lint --fix` corrupts any page that contains a ` ```markdown ` (or ` ```md `) fenced block **mid-document**: it strips the block's **closing** fence, leaving an unbalanced/open fence. Because `gbrain autopilot` runs `lint --fix` every cycle, affected notes are re-corrupted continuously — reverting by hand never sticks.

Real-world trigger: Notion-exported notes that fence a config/snippet block partway down the page. Observed on a vault where the autopilot re-stripped the same notes every ~150s cycle.

## Root cause

Detection and the fixer were out of sync in `src/commands/lint.ts`.

**Detection** (`code-fence-wrap` rule) used `/m`-flagged regexes — fires when a ` ```markdown ` opener starts *any line* **and** a closing ` ``` ` ends *any line*:

```ts
if (content.match(/^```(?:markdown|md)\s*\n/m) && content.match(/\n```\s*$/m)) { … }
```

**Fixer** (`fixContent`) used **string-anchored** regexes (no `/m`):

```ts
fixed = fixed.replace(/^```(?:markdown|md)\s*\n/, ''); // only matches at byte 0
fixed = fixed.replace(/\n```\s*$/, '');                // only matches at EOF
```

For a page whose ` ```markdown ` block is **not** at byte 0 (e.g. after frontmatter and prose), the opener-strip no-ops while the closer-strip still fires at EOF — so only the closing fence is removed, and the rule re-flags + re-mangles forever.

## Fix

Detection and fixer now share one `analyzeMarkdownWrap()` helper that treats a page as fence-wrapped **only when the entire body (after frontmatter) is a single fence**: the first non-blank body line opens it, the last non-blank line closes it, and no bare ` ``` ` appears between them (which would mean the fence closed early). A mid-document fenced block is left untouched. Routing both the rule and the fixer through one code path keeps them from drifting apart again.

## Tests

`bun test test/lint.test.ts` → **20 pass**. Two regressions added:

- `lintContent` does **not** flag a mid-document ` ```markdown ` block as a page wrap.
- `fixContent` leaves such a block intact — closing fence preserved, fences balanced.

Existing whole-page-wrap detection/fix tests still pass; `tsc --noEmit` is clean.

---

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,0 +1,5 @@
+{
+  "name": "gbrain",
+  "version": "0.25.1",
+  "description": "Personal knowledge brain skillpack — semantic + keyword retrieval over Postgres + pgvector. Skills include signal-detector, daily-task-prep, concept-synthesis, perplexity-research, briefing, and 60+ others under skills/."
+}
--- a/skills/perplexity-research/SKILL.md
+++ b/skills/perplexity-research/SKILL.md
@@ -203,8 +203,9 @@ trailing whitespace is allowed after the closing fence.
 ````markdown
 ```json
 {
-  "envelope_version": 1,
+  "envelope_version": 2,
   "brain_page_slug": "research/<exact-slug-you-wrote>",
+  "summary": "<2-3 sentence executive summary — same content as the `> Executive summary:` block at the top of the brain page>",
   "citations": [
     {"title": "<source title>", "url": "<source URL>", "accessed": "YYYY-MM-DD"}
   ],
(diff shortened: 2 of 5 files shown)


time-attack:
Closing this one. The bulk of the diff is an unrelated third-party research skill plus its plugin registration and test — we do not take vendor integrations, and bundling one inside a lint fix makes the fix unreviewable on its own terms. Separately, the root cause as described no longer holds on master: the `code-fence-wrap` detector at `src/commands/lint.ts:135` intentionally carries no `/m` flag and has a comment explaining it is kept in sync with `fixContent()`, so mid-document fenced blocks are not flagged. If you hit a residual corruption case on current master, a standalone lint-only patch with the repro would be welcome.

Issue #345, opened 2026-04-22
gbrain import walker doesn't apply isSyncable(); READMEs become orphan pages and drag brain_score (third instance of walker-drift class)


jeffweishaupt:
## Bug

Third instance of the same walker-drift class flagged in #202 (extract walker) and adjacent to #256 (artifact dirs in `isSyncable`): `gbrain import` and `gbrain sync` disagree about what counts as a page.

`isSyncable()` in `src/core/sync.ts:89` excludes `schema.md`, `index.md`, `log.md`, `README.md` — and `gbrain sync` honors this:

```ts
const skipFiles = ['schema.md', 'index.md', 'log.md', 'README.md'];
```

But `gbrain import` (`src/commands/import.ts:50` → `collectMarkdownFiles`) walks every `.md`/`.mdx` and imports it without calling `isSyncable()`:

```ts
} else if (entry.endsWith('.md') || entry.endsWith('.mdx')) {
  files.push(full);
}
```

Effect: every full import re-introduces directory READMEs (`companies/README.md`, `people/README.md`, etc.) as page records. Nothing links to them, `orphans.ts` counts them, and brain_score's `noOrphans` factor drops accordingly.

On a 109-page brain, 6 README orphans = ~5.5% orphan rate = ~1 point lost on the noOrphans component (worth 15 of 100). User-visible cost #202 didn't surface — that issue framed it as "page count is misleading."

**Suggested fix:**

Quick fix: call `isSyncable(relativePath)` in `runImport`'s file loop. Two-line patch.

Long-term (matches what #202's reporter proposed): extract a shared `shouldWalk(path)` helper out of `isSyncable` and have all three walkers (`sync`, `import`, `extract`) call it so they can't drift apart again. This would close #202 and this issue in one PR.

**Reproduction:**

```bash
gbrain query "SELECT slug FROM pages WHERE slug LIKE '%/readme'"
# Returns one row per directory README: companies/readme, people/readme, etc.
```

### Environment

- gbrain v0.17.0 (commit 55ca498)
- Postgres + pgvector engine
- 109 pages

iannwu:
Adding a concrete same-source round-trip repro from `gbrain 0.37.5.0` that shows this walker drift can also delete an existing README page during incremental sync, not just reintroduce README orphans during full import.

## Repro

Brain repo has a nested resolver page:

```text
products/README.md
```

After a prior full sync/import, it exists in the DB as:

```text
products/readme
```

[...]

PR #1419, opened 2026-05-25
fix(extract): opt-in skipMetaFiles so meta-file wikilinks get harvested

zer0me:
## Summary

**`gbrain extract links --source fs` silently drops every outbound wikilink from any file named `index.md`, `log.md`, `README.md`, or `schema.md`** — at every depth in the vault. The walker calls `isSyncable()`, which hard-skips those four basenames. But `gbrain import` *has* created those files as pages via a separate code path. The result is asymmetric: those pages exist in `pages`, get embedded, show up in search — but their outbound edges never make it into the `links` table. They're zombie pages.

This bites hardest when those meta-files are the actual hubs of the brain rather than project boilerplate. On the brain I'm running, `vault/index.md` carries ~70 wikilinks (equipo, productos, clientes, fuentes tables) and `vault/data_room_pre-seed/*/README.md` is the entry point for nine sub-folders. Every one of those was orphan in the graph despite being visibly central in the content.

## Repro

```bash
mkdir -p /tmp/v/wiki
printf '# hub\n- [[wiki/foo]]\n' > /tmp/v/index.md
printf '# foo\n' > /tmp/v/wiki/foo.md

gbrain import /tmp/v
gbrain extract links --source fs --dir /tmp/v
gbrain graph index --depth 1
# → links: []   (expected: 1 → wiki/foo)
```

Same outcome for `log.md`, any `README.md`, or `schema.md`.

## Fix

Add `skipMetaFiles?: boolean` to `SyncableOptions` (default `true` preserves the current remote-sync behavior — `gbrain sync` callers in `commands/sync.ts` should keep skipping these basenames so project boilerplate isn't pushed). Pass `false` from the shared `walkMarkdownFiles` in `commands/extract.ts` so the local link harvester treats them as the pages they really are.

No call-site changes required for downstream consumers — they continue to get the default-true behavior. The flag is purely additive.

## Why opt-in rather than removing the skip entirely

The skip is correct for `gbrain sync`'s push-to-remote path: you don't want every project's `README.md` and `index.md` pushed to a shared sync target as if they were content pages. The skip is wrong only for *local* walkers that treat the brain dir as the source of truth and need every page's edges. `skipMetaFiles: false` lets each caller declare its own intent.

## Observed impact

305-page personal brain, single `extract links --source fs` run after the patch:

| metric | before | after |
|---|---|---|
| pages_processed | 267 | **280** (+13 meta-files) |
| links_created (one run) | 0 (idempotent) | **436** |
| outbound from `index` | 18 (manual `gbrain link`) | **243** |
| outbound from `log` | 0 | **118** |
| orphans | 26 | **10** (remaining are bitácora timestamp leaves) |
| brain_score | 84 | 84 (orphans 14/15, link coverage entities 100%) |

## Test plan

- [x] `bun run typecheck` — no new errors in `src/` (pre-existing fuzz / chokidar / js-yaml issues unrelated)
- [x] End-to-end on a real 305-page brain: import → extract → graph queries show outbound from `index.md` / `log.md` / data-room READMEs (previously all 0)
- [x] Sanity-checked that `gbrain sync` path is unaffected (only `extract.ts` walker opts in)
- [ ] Maintainer to confirm no other call-site should also opt in (frontmatter command? brain-writer? would value a second look)

### DIFF
--- a/src/commands/extract.ts
+++ b/src/commands/extract.ts
@@ -83,7 +83,10 @@ export function walkMarkdownFiles(dir: string): { path: string; relPath: string 
           walk(full);
         } else if (entry.endsWith('.md') && !entry.startsWith('_')) {
           const rel = relative(dir, full);
-          if (!isSyncable(rel, { strategy: 'markdown' })) continue;
+          // skipMetaFiles: false — extract treats index.md / log.md / README.md / schema.md
+          // as real pages so their outbound wikilinks get harvested. The default-true
+          // exclusion still applies to remote sync (push), which has its own callers.
+          if (!isSyncable(rel, { strategy: 'markdown', skipMetaFiles: false })) continue;
           files.push({ path: full, relPath: rel });
         }
       } catch { /* skip unreadable */ }
--- a/src/core/sync.ts
+++ b/src/core/sync.ts
@@ -37,6 +37,14 @@ interface SyncableOptions {
   strategy?: SyncStrategy;
   include?: string[];
   exclude?: string[];
+  /**
+   * Skip meta-pages (`index.md`, `log.md`, `README.md`, `schema.md`) by basename.
+   * Default `true` to preserve historical behavior — sync to remote shouldn't push
+   * project boilerplate. Pass `false` from local-only walkers (extract, link
+   * reconciliation, frontmatter) where these files ARE legitimate pages and their
+   * outbound wikilinks must be captured.
+   */
+  skipMetaFiles?: boolean;
 }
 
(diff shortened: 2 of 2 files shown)


time-attack:
Thank you for the detailed repro and before/after numbers. Closing because the premise no longer matches master: `gbrain import` explicitly excludes the `SYNC_SKIP_FILES` metafiles from bulk import (`src/commands/import.ts` ~line 590, closing #345), so those files are not indexed as pages by either path — harvesting their outbound links would create edges from pages that do not exist. The metafile skip is a deliberate, documented convention, and #1433 already handled the one real residual case by preserving pre-existing metafile pages during sync cleanup. If a particular `index.md` is genuinely a content hub, renaming it is the supported route.

PR #1420, opened 2026-05-25
Use configured embedding dims in providers test

uncfreak1255-code:
## Summary
- preserve configured embedding dimensions when `runProviders test --model ...` targets the active embedding model
- add a regression test for the embedding override path

## Verification
- bun test test/providers-run-test.test.ts


uncfreak1255-code:
Closing this upstream PR. I’m no longer using garrytan/gbrain as my deployment target; local work now lands on my fork operating branch instead.

PR #1421, opened 2026-05-25
fix: warn when env vars override ze-switch embedding config

garrytan-agents:
## Problem

`ze-switch` writes the new embedding model to DB config and `~/.gbrain/config.json`, but `loadConfig()` gives **highest precedence** to `process.env.GBRAIN_EMBEDDING_MODEL`. If the user has this env var set (common in `.env` files loaded at gateway/shell startup), the switch silently does nothing at runtime.

The user sees `Switch status: applied` and thinks it worked. But every gbrain command still uses the old OpenAI model because the env var overrides DB + file config.

**Real-world impact:** 716K chunks had to be re-embedded because the env override meant the switch to ZeroEntropy silently didn't take effect. Doctor reported `openai:text-embedding-3-large` dimension mismatch for hours before the root cause was traced to stale env vars.

### 1. `ze-switch` env override warning

After applying the switch, checks if `GBRAIN_EMBEDDING_MODEL` or `GBRAIN_EMBEDDING_DIMENSIONS` are set in the process environment and differ from the target. Prints a prominent warning box:

```
╔══════════════════════════════════════════════════════════════╗
║  ⚠️  ENV OVERRIDE DETECTED — ACTION REQUIRED                ║
╠══════════════════════════════════════════════════════════════╣
║  GBRAIN_EMBEDDING_MODEL is set in your environment:         ║
║    Current env: openai:text-embedding-3-large               ║
║    Switch target: zeroentropyai:zembed-1                    ║
║                                                              ║
║  The env var takes HIGHEST PRECEDENCE and will override      ║
║  this switch. Update your .env file or shell environment.    ║
║                                                              ║
║  Without this change, the switch has NO EFFECT at runtime.   ║
╚══════════════════════════════════════════════════════════════╝
```

### 2. `doctor` env/DB mismatch detector

New `embedding_env_override` check: if the env var disagrees with DB config, warns with the exact fix. Surfaces the mismatch on every doctor run instead of reporting a confusing dimension error.

## Testing

- Set `GBRAIN_EMBEDDING_MODEL=openai:text-embedding-3-large` in env
- Run `gbrain ze-switch --non-interactive --force`
- Verify warning box appears after "Switch status: applied"
- Run `gbrain doctor --fast`
- Verify `embedding_env_override` warning appears

### DIFF
--- a/src/commands/doctor.ts
+++ b/src/commands/doctor.ts
@@ -3212,6 +3212,38 @@ export async function buildChecks(
     });
   }
 
+  // 8b+. Env-vs-DB embedding config mismatch detector (v0.41 env-override guard).
+  //       If GBRAIN_EMBEDDING_MODEL env var is set and disagrees with the DB config,
+  //       the user likely ran ze-switch but forgot to update their .env / shell env.
+  //       This is the #1 cause of "doctor says wrong model" confusion.
+  try {
+    const envEmbedModel = process.env.GBRAIN_EMBEDDING_MODEL;
+    if (envEmbedModel) {
+      const dbEmbedModel = await engine.getConfig('embedding_model');
+      if (dbEmbedModel && envEmbedModel !== dbEmbedModel) {
+        checks.push({
+          name: 'embedding_env_override',
+          status: 'warn',
+          message:
+            `GBRAIN_EMBEDDING_MODEL env var (${envEmbedModel}) overrides DB config (${dbEmbedModel}). ` +
+            `Env vars take highest precedence in loadConfig(). ` +
+            `Update your .env or shell environment to match: GBRAIN_EMBEDDING_MODEL=${dbEmbedModel}`,
+        });
+      }
+      const envDims = process.env.GBRAIN_EMBEDDING_DIMENSIONS;
+      const dbDims = await engine.getConfig('embedding_dimensions');
+      if (envDims && dbDims && envDims !== dbDims) {
+        checks.push({
+          name: 'embedding_env_override',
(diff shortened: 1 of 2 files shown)


garrytan:
Closing in favor of fix-wave incorporation on garrytan/pr-1414-1416-1421 (shipping as v0.41.2.1). Rebuilt with /plan-eng-review + codex outside-voice structural improvements:

- **Pre-apply refusal** (with `--ignore-env-override` escape hatch) rather than post-apply warning — the 716K-chunk damage incident in your PR description proves the schema transition is the one-way door, so the gate must fire BEFORE any mutation (snapshot write, schema). Test asserts ZERO setConfig calls fire on a refused apply.
- **Gate also fires in `resumeRetrievalUpgrade`** (codex caught the resume bypass that would have re-opened the same failure surface).
- **Gate fires BEFORE `setConfig(KEY_PREVIOUS_SNAPSHOT)` at line ~294** (not just before `runSchemaTransition` at line ~304) — refused = zero side effects.
- **`ApplyResult` tagged union** extends with `{status:'refused', reason:'env_override', warning}`; CLI `ze-switch.ts` renders the ASCII box from `formatEnvOverrideWarning` — planner stays data-pure.
- **`embedding_env_override` doctor check** uses `Check.details` (typed shape per existing `Check.issues` mismatch codex caught) and is wired into BOTH `buildChecks()` AND `doctorReportRemote()` (cross-surface parity).
- **ASCII box convention** (per repo D10) replaces the Unicode box-drawing UI.

Credit on the merged commit.

PR #1423, opened 2026-05-25
fix(ai): route Ollama chat through native API

pabloglzg:
## Summary
- declare Ollama as a local chat and expansion provider, not only embeddings
- route Ollama `/v1/chat/completions` calls through native `/api/chat` with `think:false` so thinking models return visible content for GBrain chat/expansion
- enable structured outputs for Ollama's openai-compatible client so `generateObject()` sends JSON schema requests, then map that schema to Ollama native `format`
- keep embeddings on Ollama's OpenAI-compatible `/v1/embeddings` path and normalize `OLLAMA_BASE_URL` when it omits `/v1`
- tighten the expansion prompt to require the exact `queries` key used by the schema
- update chat/provider tests and add an Ollama-specific compatibility test suite

## Verification
- `bun run typecheck`
- `bun test test/ai/recipe-ollama.test.ts test/ai/gateway-chat.test.ts test/ai/recipes-existing-regression.test.ts test/providers.test.ts test/ai/build-gateway-config.test.ts test/ai/no-batch-cap-suppression.serial.test.ts test/ai/recipe-llama-server.test.ts` (75 pass)
- `bun run src/cli.ts providers list | Select-String -Pattern '^ollama\\s'` shows Ollama with EMBED/EXPAND/CHAT enabled
- local smoke against `ollama:qwen3.6:27b`: `expand('GBrain local embeddings reranker status')` returned 4 rewrites in ~23s with `think:false`

### DIFF
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -342,6 +342,10 @@ export function applyOpenAICompatConfig(
       recipe.setup_hint,
     );
   }
+  if (recipe.id === 'ollama') {
+    const normalized = normalizeOllamaOpenAIBaseURL(baseURL);
+    return { baseURL: normalized, fetch: makeOllamaNativeChatCompatFetch(normalized) };
+  }
   return { baseURL };
 }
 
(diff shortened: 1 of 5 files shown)


time-attack:
Thanks for the careful work — the `think:false` diagnosis and the base-URL normalization are both correct observations about Ollama's API. This lands as a new provider capability rather than a fix: master's Ollama recipe only ever declared an embedding touchpoint, so chat was never a promise we broke. We're not taking ~200 lines of vendor-specific protocol translation into the shared gateway path, so closing on integration-surface policy.

PR #1427, opened 2026-05-25
feat(synopsis): tail-truncate documentText for small-model chat handlers

BenjaminDSmithy:
## Summary

Add `SYNOPSIS_DOC_MAX_CHARS` env-overridable cap (default 32768 chars,
~8K tokens) applied in `buildUserPrompt` for the per-chunk synopsis
prompt. Truncates the TAIL of `documentText` so the head (title,
frontmatter, intro) carries the document-level anchor the synopsis
needs.

## Motivation

Small local chat models (Gemma 4 E2B, Qwen3 4B on Ollama / LM Studio)
get dramatically slower on long contexts even when their declared
window is 131K tokens. A 73K-char page synopsis on Gemma 4 E2B takes
60-120s, exceeding the worker's default 30s `lockDuration` and tripping
`lock-lost` errors during the contextual-retrieval per_chunk_synopsis
backfill.

This is a companion lever to issue #1014 / PR #1177 (`--lock-duration`
on `gbrain jobs work`) — bumping the lock TTL alone helps but lets each
job consume more wall-clock per chunk; capping `documentText` makes each
call complete *faster*, which lets the existing 30s lock survive on
medium-length pages too. Both together get the synopsis backfill running
reliably on small local LLMs without burning Anthropic credit.

## Behavior

- Default `SYNOPSIS_DOC_MAX_CHARS = 32768` (~8K tokens at 4 chars/tok)
- Env override: `GBRAIN_SYNOPSIS_DOC_MAX_CHARS=<int>` with range
  `[512, 1048576]` (clamped via regex; falls back to default on invalid)
- **Anthropic Haiku unaffected** at 32768 cap (prefill cheap, no
  per-chunk slowdown). Bump higher for frontier models if you want
  richer document anchoring.
- Truncation TAIL only — placeholder inserted: `[... N chars truncated
  for synopsis budget ...]`

## Verified

Real 1383-chunk personal brain backfill at:

\`\`\`
GBRAIN_SYNOPSIS_MODEL=lmstudio:google/gemma-4-e2b
GBRAIN_SYNOPSIS_DOC_MAX_CHARS=16384
gbrain jobs work --concurrency 4 --lock-duration 300000
\`\`\`

Pre-change: every transcript page (50-130 chunks each) hit \`lock-lost\` after ~30s. Throughput collapsed.

Post-change: no deaths, no stalls. Steady throughput at ~10 jobs/min during burst windows on a single LM Studio Gemma E2B instance.

## Tests

Manual regression: ran against the production-scale 1383-chunk
backfill described above. Unit test for the truncate behavior in
\`buildUserPrompt\` would require exporting the currently-private
helper or adding a test seam — happy to add in a follow-up if a
maintainer prefers a particular shape.

## Pre-push gate

\`bun run verify\` passes locally (privacy + jsonb + progress +
test-isolation + wasm + admin-build + typecheck).

## Trigger paths

Not retrieval-touching (no \`src/core/search/*\` changes), so
\`gbrain eval replay\` not required per CONTRIBUTING.md.

## Related

- Issue #1014 + PR #1177 — \`--lock-duration\` flag (the companion
  lever; this PR is the documentText half of the belt-and-suspenders
  pair)
- PR #1358 — synopsis model env override (independent; covers the
  AUTH side of the same backfill scenario)


BenjaminDSmithy:
Self-second-opinion via `codex review` flagged a P2 issue: changing `GBRAIN_SYNOPSIS_DOC_MAX_CHARS` shifts the synopsis prompt + downstream embeddings for long documents but the cap was NOT folded into the `computeCorpusGeneration` hash. Same `corpus_generation` for different caps would defeat the v0.40.3.0 D27 P1-5 cache-invalidation contract.

Pushed fix (5f58e90d):
- Export `SYNOPSIS_DOC_MAX_CHARS` from `src/core/page-summary.ts`
- `computeCorpusGeneration` accepts optional `synopsisDocMaxChars` param. Folded into hash as `|doc_cap=<N>` when set; omitted for non-synopsis modes so existing pre-PR caches stay valid for title / none modes
- Service-layer call sites (2 in `contextual-retrieval-service.ts`) pass the cap when `attemptMode/resolution.mode === 'per_chunk_synopsis'`, undefined otherwise
- `import-file.ts` inline path passes undefined (it refuses per_chunk_synopsis upstream)

`bun run verify` clean.

PR #1428, opened 2026-05-25
fix(models): dispatch subcommand reads args[0] not args[1]

BenjaminDSmithy:
## Summary

\`gbrain models doctor\` silently falls through to the read view
instead of running the reachability probe.

\`runModels\` (src/commands/models.ts) checks
\`args[1] === 'doctor'\`, but the caller —
\`handleCliOnly(command, subArgs)\` in \`src/cli.ts:113\` — passes
\`subArgs\` (the leading command token already stripped). So inside
\`runModels\`, \`args[0]\` is the subcommand and \`args[1]\` is
undefined.

The doctor probe path has been unreachable from the CLI since the
\`handleCliOnly\` refactor. \`gbrain models help\` happens to work
by falling through to the \`--help\` flag detection
(\`args.includes('--help')\`), masking the bug.

Two-char fix: \`args[1]\` → \`args[0]\` on both branches of the
ternary.

## Verified

Pre-fix: \`gbrain models doctor\` prints the routing table (read
view) and exits 0 — probe never runs.

Post-fix on a real production brain:

\`\`\`
Model reachability probe:
  ✔ embedding_config  ollama:bge-m3                              ok (0ms)
  ✔ reranker_config   (none)                                     ok (0ms)
  ✘ chat              lmstudio:mistralai/magistral-small-2509    unknown (5012ms)
      [chat(lmstudio:mistralai/magistral-small-2509)] probe timed out after 5s
  ✔ expansion         lmstudio:google/gemma-4-e2b                ok (535ms)

Summary: 3/4 reachable.
\`\`\`

(The chat timeout in this snapshot is the user's LM Studio
having Magistral ejected, not a fix issue.)

## Tests

Manual regression — \`gbrain models doctor\` from a fresh checkout
prints the "Model reachability probe:" header. Unit test would
require mocking the gateway probe paths; happy to add in a
follow-up if a maintainer prefers a particular shape.

## Pre-push gate

\`bun run verify\` passes locally (privacy + jsonb + progress +
test-isolation + wasm + admin-build + typecheck).

## Trigger paths

Not retrieval-touching, so \`gbrain eval replay\` not required per
CONTRIBUTING.md.


BenjaminDSmithy:
Self-second-opinion via `codex review` flagged a P2 regression: the args[1]→args[0] rewrite turned `gbrain models doctor --help` into a network probe run instead of usage print. The original args[1]-shaped ternary happened to dodge this by always falling through to the `args.includes('--help')` branch when args[1] === 'doctor' was false; the new args[0] check fires doctor first.

Pushed fix (64cdbe84): reorder ternary. `hasHelp` computed FIRST from (--help / -h / args[0] === 'help'), then `sub = hasHelp ? 'help' : args[0] === 'doctor' ? 'doctor' : 'read'`.

`bun run verify` clean.

PR #1430, opened 2026-05-25
fix(sync): record sync attempt timestamp on no-op + suppress on pull failure

rayers:
## Problem

doctor's `sync_freshness` check reads `sources.last_sync_at` to decide whether a source is stale. Two gaps:

**(A) No-op syncs never advanced `last_sync_at`.** When `performSync` returns early via the `up_to_date` path (git HEAD unchanged, chunker current, no detached working-tree changes), the column was never touched. So a quiet source (no upstream commits in N days) was misreported as "stale 16d" even though every cron tick polled it and got "Already up to date". The metric users intuit from "last sync" is *last time we attempted a sync*, not *last time we imported new data*.

**(B) `last_sync_at` advanced even when the git pull failed.** A network partition / revoked credential / diverged remote means we never observed remote state — but freshness still advanced, telling doctor/autopilot the source was healthy when it wasn't.

## Fix

- Advance `last_sync_at = now()` on the no-op `up_to_date` early-return (federated `opts.sourceId` path only; legacy `config.sync.last_run` path untouched).
- Gate every `last_commit` write site on `pullAttemptedAndFailed` (threaded through `writeSyncAnchor` + `performFullSync`): `last_commit`/`newest_content_at` still advance against the stale checkout, but the freshness signal is suppressed.
- Operator-skipped offline modes (`--no-pull`, detached HEAD, no origin) are NOT failures and DO advance. `--dry-run` stays side-effect free.

## Test

`test/sync-no-op-touches-last-sync-at.test.ts` (PGLite, 5 cases): no-op advances freshness; no-op without `sourceId` is a no-write; `--dry-run` mutates nothing; no-op + pull-failure does NOT advance; incremental + pull-failure advances `last_commit` but not `last_sync_at`. **5/5 green** on current master.

---
*Rebased onto current `master`. Upstream `#1794` (resumable incremental sync) reworked this path but covers **neither** gap — the no-op early-return at the `up_to_date` branch still skips the freshness write, and there's no pull-failure gate. Both halves of this fix still apply.*

### DIFF
--- a/src/commands/sync.ts
+++ b/src/commands/sync.ts
@@ -601,22 +601,31 @@ async function writeSyncAnchor(
   // git-intrinsic committer time of the HEAD we just synced). `undefined` keeps
   // the legacy 2-column write; `null` clears the column (git unavailable).
   newestContentEpochMs?: number | null,
+  // #1430: when the upstream pull failed this run, advance last_commit (and
+  // newest_content_at — local import may have succeeded against the stale
+  // checkout) but do NOT stamp last_sync_at. doctor/autopilot's sync_freshness
+  // must not read the source as "fresh" when we never observed remote state
+  // (network partition, revoked credentials, diverged remote). Operator-skipped
+  // offline modes (--no-pull, detached HEAD, no origin) are NOT failures and
+  // pass pullFailed=false so freshness advances normally.
+  pullFailed = false,
 ): Promise<void> {
   if (sourceId) {
     const col = which === 'repo_path' ? 'local_path' : 'last_commit';
-    // last_sync_at bookmarked on every last_commit advance.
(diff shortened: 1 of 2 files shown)


time-attack:
Thank you for the precise write-up — your framing of last_sync_at as "last attempt" versus "last import" is exactly the distinction master settled on. Both halves have since landed: the no-op heartbeat is at sync.ts:2283-2294 (v0.42.52.0), and the pull-failure suppression came in with #3068, which returns a `partial`/`pull_failed` result instead of `up_to_date` so freshness stays frozen and the run exits non-zero. Closing as already fixed — your analysis matched the eventual fix on both counts.

PR #1443, opened 2026-05-25
fix: ze-switch preserves embedding_image column dimensions

garrytan-agents:
## Problem

`runSchemaTransition` drops and recreates **both** `embedding` and `embedding_image` columns at the target dimension. But `embedding_image` uses a separate multimodal model (e.g. `voyage-multimodal-3` at 1024d) whose dimensions are independent of the text embedding model.

Switching from OpenAI `text-embedding-3-large` (1536d) to ZeroEntropy `zembed-1` (1280d) also changes `embedding_image` from `vector(1024)` → `vector(1280)`, creating a silent dimension mismatch that prevents the multimodal provider from writing image embeddings.

## Fix

Only transition the primary `embedding` column. Leave `embedding_image` dimensions untouched — just rebuild its HNSW index for safety.

## Impact

Without this fix, every `ze-switch` run silently breaks multimodal search for brains using a separate image embedding model with different dimensions.

### DIFF
--- a/src/core/retrieval-upgrade-planner.ts
+++ b/src/core/retrieval-upgrade-planner.ts
@@ -440,19 +440,40 @@ export async function undoRetrievalUpgrade(engine: BrainEngine): Promise<
  * `--resume`.
  */
 async function runSchemaTransition(engine: BrainEngine, targetDim: number): Promise<void> {
+  // v0.41 fix: only transition the primary text embedding column.
+  // The embedding_image column uses a separate multimodal model (e.g.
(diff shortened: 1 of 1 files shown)


garrytan:
Cherry-picked into worktree branch `garrytan/pr-1443-ze-switch` for landing through the standard fix-wave flow. Closing the fork PR — fork PRs from non-collaborator accounts don't get base-repo secrets, so CI auth checks fail. Original attribution preserved in the cherry-picked commit. Thanks!

PR #1472, opened 2026-05-26
RFC: Parallel Chunked Sync for Large Federated Brains

garrytan-agents:
## Problem

The sync pipeline becomes a reliability bottleneck at 370K+ pages with 4 federated sources. Three compounding failure modes cause sources to fall behind:

1. **Serial timeout cascade** — monolithic `sync --all` times out, killing ALL sources including ones not yet reached
2. **Lock contention** — stale locks from timed-out runs block the next cron wave
3. **Embed blocks sync** — inline embed eats into the timeout budget

**Production data (last 24h):** 8 of 12 sync cron runs timed out at 1803s. Sources going 50+ hours stale.

## Proposal (RFC only — no code changes)

Four additive flags:
- `--timeout <seconds>` — graceful self-termination, commits partial progress
- `--break-lock-if-stale <seconds>` — cron self-healing without manual `--break-lock`
- `--independent` mode — per-source process isolation (one source timing out doesn't affect others)
- Deprecate monolithic `sync --all` as the default cron path

Full RFC: [docs/rfcs/parallel-chunked-sync.md](https://github.com/garrytan-agents/gbrain/blob/feat/parallel-chunked-sync/docs/rfcs/parallel-chunked-sync.md)

## Impact

| Metric | Current | After |
|--------|---------|-------|
| Sync reliability | ~33% (4/12 runs) | ~95%+ |
| Max source staleness | 50+ hours | < 2 hours |
| Recovery from timeout | Manual intervention | Automatic |

All changes are additive — `sync --all` continues to work as-is.

### DIFF
--- a/docs/rfcs/parallel-chunked-sync.md
+++ b/docs/rfcs/parallel-chunked-sync.md
@@ -1,0 +1,121 @@
+# RFC: Parallel Chunked Sync for Large Federated Brains
+
+## Problem Statement
+
+gbrain's sync pipeline becomes a reliability bottleneck at scale (370K+ pages, 4 federated sources). The current architecture has three compounding failure modes that cause sources to fall behind and eventually trigger doctor FAIL alerts.
+
+## Failure Mode 1: Serial Timeout Cascade
+
+**Observed:** The `gbrain-sync-all` cron consistently times out when the total sync+embed cycle exceeds the cron timeout window.
+
+When run via an agent cron (the primary production path), the sync job has a fixed timeout. The pipeline processes sources sequentially in a single agent turn:
+
+```
+git pull (4 sources) → sync default → sync straylight → sync zion → sync media-corpus → embed --stale
+```
+
+Even with `--no-embed`, each source's sync takes variable time (depends on diff size, chunk count, DB write latency). When any source has a large backlog, the total wall time pushes past the timeout, and the **entire run** is killed — including sources that haven't been reached yet.
+
+**Production data (last 24h):**
+- 8 of 12 runs timed out at exactly 1803s (30-min timeout)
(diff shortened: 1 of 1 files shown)


garrytan:
Closing — the actual fix shipped in **[PR #1506](https://github.com/garrytan/gbrain/pull/1506)** as v0.41.15.0.

Your RFC was correct in every diagnosis. The shipped plan reduced scope from your 4 surfaces to 2 surfaces + a documented shell-level cron pattern, after iteration with codex over 3 review passes on cascade-resilience and lock-stealing invariants. Specifically:

- **`--independent`** (Minion fan-out) was rejected because `MinionWorker` is an in-process worker pool, not OS-process-per-source — the "process isolation" claim was oversold. `waitForCompletion` throws on timeout but doesn't cancel the underlying job (leaving a hot lock for the next cron). Shell `timeout(1)` per-source loop gives real OS process isolation with zero new gbrain code. Documented as the recommended cron pattern in CHANGELOG + README.
- **`--break-lock-if-stale`** was replaced by extending the existing `--break-lock` with `--all` + `--max-age`. The semantic changed from "time since acquired" to "time since last refresh" via a new `last_refreshed_at` column (migration v98) so healthy long-running holders are safe by construction. Without that change, the cron pattern would have stolen healthy locks during the upgrade window.
- **`--timeout`** ships as you proposed, with one honest scope statement: it covers pull + delete + rename + import only. Extract + embed run to completion after the bookmark write. Full-sync triggers (first sync, `--full`, chunker-bump rewalk) are deferred to v0.42+.

Credit for surfacing the production data lives in the v0.41.15.0 CHANGELOG entry.

Closing as superseded.

PR #1483, opened 2026-05-26
feat(sync): .gbrainignore + multi-layer per-repo exclusion (v0.40.9.0)

eepaul:
## Summary

Stop indexing files you didn't want indexed. Three additive layers gate every sync walker and incremental diff:

1. **`.gbrainignore`** at the repo root — committed, travels with the repo
2. **`sources.config.excludePatterns`** — persistent per-source (JSONB)
3. **`--exclude <glob>`** CLI flag — one-shot override

Full gitignore parity via the [`ignore`](https://www.npmjs.com/package/ignore) npm lib (same parser ESLint, Prettier, and Husky use). Merge order: `dotfile → source.config → cli`, last wins per gitignore's last-match-wins semantics.

## Why

The user dropped this in plain English: "now when using /sync-gbrain in some project, it syncs every file in the project directory into the brain. Some data file directories shouldn't enter gbrain." Before this PR there was no `.gbrainignore`, no `--exclude` flag, no per-source ignore config in `sources.config`, and `.gitignore` was not consulted. A project with `data/`, `datasets/`, `*.parquet`, `fixtures/`, `models/` got all of it indexed. The only escape hatches were renaming the data dir to start with a dot, moving it under `*.raw/`, or `--no-code`.

## What's new at a glance

```bash
# Drop in any project root:
echo "data/
fixtures/large/
*.parquet" > .gbrainignore

# Next /sync-gbrain or `gbrain sync --source <id>` honors it.
# Existing orphan pages get reaped on next sync (safety-guarded).

# CLI escape hatch for one-offs:
gbrain sync --source <id> --exclude '!data/keep.md'

# Persistent per-source (survives autopilot cycles):
gbrain sources add my-repo --path /path --exclude 'data/' --exclude '*.parquet'
gbrain sources update my-repo --exclude 'fixtures/' --clear-excludes

# Surfaced in:
gbrain doctor                # new gbrainignore_health check
```

## Architecture decisions (locked in `/plan-eng-review`)

| Tag | Decision |
|-----|----------|
| Step 0 | Hold scope — all three layers ship in one wave |
| **A1** | Thread `cfg.excludePatterns` + `cfg.strategy` through `cycle.runPhaseSync` (closes daemon gap) |
| **A2** | Auto-reconcile pages on pattern-hash change; 50%/1000 safety guard refuses runaway deletions |
| **A3** | Negation-aware descent-prune fallback (walker walks dirs when any `!` pattern present) |
| **A4** | Merge order: cli > source.config > .gbrainignore (gitignore-style last-match-wins) |
| **C1** | Single `resolveExclusions()` helper consumed by 4 call sites (DRY) |
| **C2** | mtime-based cache invalidation (autopilot daemon picks up dotfile edits within one cycle) |
| **P1** | Reconciliation streams 1000 rows per query (memory-bounded on 200k-page brains) |

## Failure modes pinned by IRON-RULE tests

The /plan-eng-review IRON RULE captures regressions where silent bugs cost real data. All four are pinned in `test/sync-exclude.test.ts`:

1. **Negation-aware descent-prune fallback** — `data/* + !data/keep.md` correctly emits `keep.md`. Without the `safeForDescentPrune` flag, walker would short-circuit on the parent dir and silently lose the rescue.
2. **50%-deletion safety guard** — refuses to delete >50% of a source's pages without `--reconcile-excludes --yes`. The class of bug this catches: a user typos `**` in `.gbrainignore`, walker excludes everything, reconciliation would otherwise wipe the source.
3. **1000-row absolute guard** — same posture for huge brains where percentage looks small.
4. **1000-row batched reconciliation** — crosses BATCH_SIZE boundary cleanly on a 1100-row seeded fixture. Memory stays bounded.

## File-by-file change list

```
PRODUCTION (8 files)
  src/core/gbrainignore.ts          NEW   load + resolveExclusions + mtime cache (~280 LOC)
  src/core/cycle.ts                 MOD   A1 — thread cfg.excludePatterns + cfg.strategy
  src/commands/import.ts            MOD   walker descent-prune guarded by safeForDescentPrune (A3)
  src/commands/sync.ts              MOD   4 call sites + reconciliation + 1000-row batching
  src/commands/sources.ts           MOD   sources add --exclude, new sources update subcommand
  src/core/sources-ops.ts           MOD   AddSourceOpts.excludePatterns, updateSource()
  src/commands/doctor.ts            MOD   gbrainignore_health check
  (no src/core/sync.ts pruneDir extension — descent-prune is path-aware, lives in walker)

TESTS (4 files, 47 new cases — 150 total green after this PR)
  test/gbrainignore.test.ts         NEW   22 cases (module + cache + mtime)
  test/sync-exclude.test.ts         NEW   15 cases incl. 4 REGRESSION-CRITICAL
  test/sources-exclude.test.ts      NEW   10 cases (CLI + opsUpdate semantics)
  test/doctor-behavioral.test.ts    MOD   +3 cases (gbrainignore_health)

DOCS/META (3 files)
  docs/guides/sync-filtering.md     MOD   new "User-configurable exclusion" section
  CHANGELOG.md                      MOD   v0.40.9.0 wave entry
  package.json + VERSION            MOD   ignore@^7.0.5, bump to 0.40.9.0

DEP
  ignore@^7.0.5                     NEW   MIT, ~17KB packed, zero transitive
```

## Verification

- `bun run typecheck` — clean
- `bun test test/gbrainignore.test.ts test/sync-exclude.test.ts test/sources-exclude.test.ts test/doctor-behavioral.test.ts test/sync-strategy.test.ts test/sync.test.ts test/sync-walker-*.test.ts` — **150 pass / 0 fail / 503 expect()**
- `bun run verify` — clean (privacy, source-id, jsonb, ops-filter-bypass, all 13 guards)

## Honest notes

- Planned `pruneDir(name, parentDir?, dirExcludes?)` extension at `src/core/sync.ts` cancelled during implementation. `pruneDir` operates on single segment names; descent-prune is path-aware so the right home is the walker in `src/commands/import.ts`. Documented in commit message.
- `sources.config` JSONB reads use untyped casts matching the existing `cfg.strategy` pattern. A zod schema for the whole shape is a deferred follow-up — not load-bearing because writers (`sources add` / `sources update`) are typed at entry.
- Autopilot daemon's `runPhaseSync` doesn't accept its own `--exclude` flag for ad-hoc fan-out. The daemon reads `cfg.excludePatterns` per source from the DB. A user wanting a one-off daemon-side exclusion would `sources update --exclude` first. Daemon-side CLI plumbing is a follow-up if anyone hits the use case.

## Gitignore semantics gotcha (documented)

Once a directory is excluded via `data/`, children **cannot** be re-included with `!data/keep.md` — this is documented gitignore behavior, not a gbrain quirk. The rescue pattern requires excluding contents instead: `data/* + !data/keep.md`. Called out in `docs/guides/sync-filtering.md` so users hit it once and never again.


ashtonfeliciano:
I ran into exactly the scenario this PR would fix, and wanted to add a
real-world reproduction case in case it's useful.

**Setup:** GBrain source cloned into a subdirectory of the vault being indexed
(`vault/AI_Configs/GBrain-src/`). Every import run reported 8 failures:

```
Import completed with 8 failure(s). sync.last_commit NOT advanced
```

The verbose output traced them to GBrain's own calibration test fixtures —
`test/fixtures/calibration/holdout/` and `test/fixtures/calibration/extract-takes-corpus/`.
Their hardcoded `slug:` frontmatter doesn't match the path GBrain derives
from their location inside the source tree, so import rejects them on every run.

I read through `src/commands/import.ts` to look for an existing exclude mechanism
and confirmed there isn't one — the walker currently only skips hidden dirs
(`.`-prefixed), `node_modules`, and `ops`.

**Workaround I'm using:** Rename `test/` to `.test/` so the walker skips it.
Works, but gets undone by `git pull` on the source clone.

A `.gbrainignore` or `--exclude` flag would give this a proper native fix.
Supporting the PR — this is a real gap.

time-attack:
Closing as obsolete — fix already on master (see 7c27fa12 Addresses #1483). Please reopen if still needed.

PR #1488, opened 2026-05-26
fix: use jsonSchema() wrapper and 'parameters' key for AI SDK tool definitions

### LINKED ISSUE #1487: gateway.chat(): incorrect tool schema format breaks non-Anthropic providers (agent.use_gateway_loop)
## Bug description

When `agent.use_gateway_loop=true` and the subagent model is a non-Anthropic
provider (e.g. DeepSeek, OpenRouter, any OpenAI-compatible recipe), the gateway
tool loop fails with:

```
[chat(deepseek:deepseek-v4-flash)] schema is not a function. 
(In 'schema()', 'schema' is an instance of Object)
```

## Root cause

In `src/core/ai/gateway.ts`, the `chat()` function passes tool schemas as raw
JSON Schema objects under the `inputSchema` key to the AI SDK's `generateText()`:

```typescript
// gateway.ts ~line 2360
    acc[t.name] = {
      description: t.description,
    };
    return acc;
```

The Vercel AI SDK expects tools to use the `parameters` key with a
`jsonSchema()`-wrapped value (the `jsonSchema` helper from the `ai` package).
The `inputSchema` key is silently ignored for openai-compatible providers.

When the SDK resolves tool definitions, it attempts to call `.schema()` on
whatever lands under `parameters` — which is `undefined` since that key is
never set. This falls through to the `inputSchema` value (a plain Object),
and the `.schema()` call fails.

## Scope of impact

This bug affects **all non-Anthropic providers** using the gateway tool loop
path (`agent.use_gateway_loop=true`). The default Anthropic-direct subagent
path is unaffected because it bypasses `gateway.chat()` entirely.

**Affected versions:** all releases that include `agent.use_gateway_loop`
support — introduced in the v0.38 series and present through current HEAD
(v0.41.14.0).

## Reproduction

Prerequisites: a configured non-Anthropic chat model with API key set.

```bash
# Create a transcript file >2000 chars
gbrain dream --phase synthesize --input <transcript>
```

Result: the significance judge phase completes, but the subagent job dies
with `schema is not a function`.

## Fix

Import `jsonSchema` from the `ai` package and use the correct `parameters`
key with `jsonSchema()` wrapping:

```diff
-import { embed as aiEmbed, embedMany, generateObject, generateText } from 'ai';
+import { embed as aiEmbed, embedMany, generateObject, generateText, jsonSchema } from 'ai';

-      inputSchema: { jsonSchema: t.inputSchema } as any,
+      parameters: jsonSchema(t.inputSchema),
```

## Additional context

Discovered while running a Hermes multi-agent cluster with DeepSeek as the
primary LLM. A fix branch is available at
`justemu/gbrain:fix/gateway-tool-json-schema` for reference.

justemu:
## Summary

Fixes #1487 - the gateway chat() function passes tool schemas in a format incompatible with the Vercel AI SDK for non-Anthropic providers.

## Problem

When agent.use_gateway_loop=true, the subagent handler routes through gateway.toolLoop(), which calls gateway.chat(). The chat() function passes tool schemas as { inputSchema: { jsonSchema: rawObject } } to generateText(). The AI SDK expects { parameters: jsonSchema(rawObject) } where jsonSchema is imported from the ai package.

This causes all non-Anthropic subagent tool loops to fail with: [chat(deepseek:...)] schema is not a function

## Changes

2 lines changed in src/core/ai/gateway.ts. Import jsonSchema from ai, and use parameters: jsonSchema(t.inputSchema) instead of inputSchema: { jsonSchema: t.inputSchema }.

## Testing

Tested with DeepSeek v4 Flash as the subagent model. The Anthropic-direct path is unaffected.

## Affected versions

All releases with agent.use_gateway_loop support (v0.38+ through current HEAD).

### DIFF
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -21,7 +21,7 @@
  *     rotation (via configureGateway()) invalidates stale entries.
  */
 
-import { embed as aiEmbed, embedMany, generateObject, generateText } from 'ai';
+import { embed as aiEmbed, embedMany, generateObject, generateText, jsonSchema } from 'ai';
 import { AsyncLocalStorage } from 'node:async_hooks';
 import { listRecipes } from './recipes/index.ts';
 import { createOpenAI } from '@ai-sdk/openai';
@@ -2357,7 +2357,7 @@ export async function chat(opts: ChatOpts): Promise<ChatResult> {
   const tools = (opts.tools ?? []).reduce((acc, t) => {
     acc[t.name] = {
       description: t.description,
-      inputSchema: { jsonSchema: t.inputSchema } as any,
+      parameters: jsonSchema(t.inputSchema),
     };
     return acc;
   }, {} as Record<string, any>);


Masashi-Ono0611:
Friendly nudge 🙏 — been running this as a local patch on v0.42.26.0 for ~2 weeks, works cleanly. Any chance it could get a review/merge? Thanks for the fix!

garrytan:

PR #1489, opened 2026-05-26
fix(sync): refresh source freshness on no-op sync

sean808080:
## Summary

- Refresh source `last_sync_at` when a sync exits via the `up_to_date` path.
- Preserve the existing global `sync.last_run` behavior for non-source syncs.
- Add a regression test that repeats a same-HEAD source sync and verifies freshness advances.

## Why

Autopilot or other periodic sync callers can successfully check an unchanged source and receive `status: 'up_to_date'`. Before this change, that early return skipped the freshness write, so source-health checks could report the source as stale even though polling was running successfully.

## Test Plan

- `bun test test/sync-last-sync-at.test.ts`
- `bun run typecheck`

### DIFF
--- a/src/commands/sync.ts
+++ b/src/commands/sync.ts
@@ -366,6 +366,20 @@ async function writeSyncAnchor(
   await engine.setConfig(`sync.${which}`, value);
 }
 
+async function markSyncChecked(
+  engine: BrainEngine,
+  sourceId: string | undefined,
+): Promise<void> {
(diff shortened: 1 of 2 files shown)


garrytan:
Thanks for this! Consolidating the sync no-op freshness fixes — we are keeping #1430 because it also handles the pull-failure case. Closing this as a duplicate; appreciated.

PR #1494, opened 2026-05-26
Fix providers test base URL override

guoqunabc:
## Summary
- Preserve configured `provider_base_urls` when `gbrain providers test --model ...` reconfigures the gateway for an isolated provider smoke test.
- Add a hermetic CLI regression test that proves the isolated DashScope model probe uses the configured base URL instead of the recipe default.

## Why
China-region DashScope users configure `provider_base_urls.dashscope` to use `https://dashscope.aliyuncs.com/compatible-mode/v1`. Before this change, passing `--model dashscope:text-embedding-v4` discarded that override and silently fell back to the international endpoint, producing a misleading `Incorrect API key` error even when the key was valid for the configured regional endpoint.

## Verification
- `bun test test/providers-cli-base-url.serial.test.ts`
- `bun test test/providers.test.ts test/ai/recipe-dashscope.test.ts`
- `gbrain providers test --touchpoint embedding --model dashscope:text-embedding-v4`

### DIFF
--- a/src/commands/providers.ts
+++ b/src/commands/providers.ts
@@ -151,6 +151,8 @@ async function runTest(args: string[]): Promise<void> {
     process.exit(1);
   }
 
+  let cfg: ReturnType<typeof loadConfig> | undefined;
+
   // If --model passed, override gateway for this test (touchpoint-aware).
   if (modelArg) {
     const [providerId, ...modelParts] = modelArg.split(':');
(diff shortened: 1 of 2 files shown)


time-attack:
Thanks for reporting and fixing this — the regional DashScope failure mode you describe is real and was worth chasing down. It has since been fixed on master by #2863, which routes the isolated `--model` probe through `buildGatewayConfig` so `provider_base_urls` is preserved; the comment at providers.ts:177-184 describes your exact scenario, misleading "Incorrect API key" and all. Closing as already fixed.

PR #1500, opened 2026-05-26
fix: lint phase engine.disconnect() kills worker shared DB pool

### LINKED ISSUE #1499: Bug: lint phase engine.disconnect() kills worker shared DB pool, causing crash-loop
## Summary

During autopilot cycle execution, the lint phase creates a second PostgresEngine and shares the module-level db.sql singleton via engine.connect({}) (no poolSize). When the phase completes, engine.disconnect() in the finally block destroys the shared pool, rendering the worker unable to operate. All subsequent phases fail, and the worker crashes with an unhandled rejection.

## Impact

- All 9 sources never completed a full cycle
- Brain score stuck at 51 (links 3/25, timeline 2/15, orphans 1/15)
- Worker crash-loops: 51 crash records, max 5 restarts before autopilot gives up

## Root Cause

src/commands/lint.ts:298-319, resolveLintContentSanity():

```typescript
const engine = await createEngine({...});
try {
    await engine.connect({});  // no poolSize -> module singleton path
    const lifted = await loadConfigWithEngine(engine, base);
} finally {
    await engine.disconnect(); // -> db.disconnect() -> sql.end(); sql = null
}
```

engine.connect({}) without poolSize routes through the module-level db.sql singleton. The subsequent engine.disconnect() in the finally block kills the shared pool for ALL callers in the process, including the MinionWorker.

## Diagnosis

Instrumented db.disconnect(), PostgresEngine.disconnect(), and getConnection() with stack traces. The stack trace confirms the disconnect originates from resolveLintContentSanity -> lint.ts:319 -> runLintCore -> runPhaseLint -> runCycle.

## Fix

Pass poolSize: 2 so the lint engine gets its own instance-level pool that does not share state with the module singleton:

```typescript
```

## Related

The /No database connection/i pattern in retry-matcher.ts (added v0.41.2.1) marks this terminal error as retryable, producing misleading connection blip log messages and wasting a 500ms retry. Since the error means the module-level singleton is null, retrying without explicit reconnect is guaranteed to fail.

justemu:
## Problem

During autopilot cycle execution, the lint phase creates a second PostgresEngine and shares the module-level db.sql singleton via engine.connect({}) (no poolSize). When the phase completes, engine.disconnect() in the finally block destroys the shared pool, rendering the MinionWorker unable to operate.

Impact: All 9 sources never completed a full cycle. Brain score stuck at 51. Worker crash-loops with 51 crash records.

## Root Cause

src/commands/lint.ts:298-319 - resolveLintContentSanity(). engine.connect({}) without poolSize routes through the module-level db.sql singleton. The subsequent engine.disconnect() kills the shared pool for ALL callers in the process.

## Changes

1. lint.ts: pass poolSize: 2 so the lint engine gets its own instance-level pool
2. retry-matcher.ts: remove /No database connection/i from retryable patterns

Closes #1499

### DIFF
--- a/src/commands/lint.ts
+++ b/src/commands/lint.ts
@@ -312,7 +312,12 @@ async function resolveLintContentSanity(): Promise<LintContentOpts['contentSanit
         database_path: base!.database_path,
       });
       try {
-        await engine.connect({});
+        // poolSize: 2 forces an instance-level pool so engine.disconnect()
+        // does NOT clobber the module-level db singleton that the worker
+        // (or any other caller sharing the process) relies on.
+        // Without this, a lint phase inside an autopilot cycle kills the
+        // worker's shared DB connection and every subsequent phase fails.
+        await engine.connect({ poolSize: 2 });
         const lifted = await loadConfigWithEngine(engine, base);
         cs = lifted?.content_sanity ?? cs;
       } finally {
--- a/src/core/retry-matcher.ts
+++ b/src/core/retry-matcher.ts
@@ -20,11 +20,13 @@ const CONN_PATTERNS = [
   /connection.*closed/i,
   /server closed the connection/i,
   /could not connect to server/i,
-  // v0.41.2.1: gbrain's own GBrainError thrown by getConnection() when
-  // the singleton pool was nulled (engine.disconnect mid-cycle, or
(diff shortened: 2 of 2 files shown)


garrytan:

PR #1501, opened 2026-05-26
fix(links): two re-orphaning guards — phonetic create-gate + prose mentions

anoopkansupada:
Code half of plans/gbrain-graph-connectivity-fix. Guard 1: phonetic resolve-before-create at putPage (both engines) — rejects NEW people/companies slugs that are near-duplicates (vowel-drift/display-noise/corporate-suffix) of existing canonicals, queues at name_collision_review/. Guard 2: infer_links prose-mention pass over Granola cron meeting bodies (existing entities only, never creates) + deep-flatten fix so triple-nested [[[people/...]]] attendees emit links. 25 new tests pass; 129 existing link/put-page tests still green; live scan of sports-bridge-baci links Anchorage/Lemma/BACI in prose. Closes tasks/post-call-processor-name-sanity-check + tasks/gbrain-granola-cron-entity-resolution.


anoopkansupada:
Wrong base — re-targeting our canonical fork anoopkansupada/gbrain.

PR #1504, opened 2026-05-26
feat(embeddings): add Mistral codestral-embed recipe

VictorGjn:
## Summary

Adds a native Mistral embedding recipe for `codestral-embed`. Mistral's `/v1/embeddings` is OpenAI-shaped on the response, so this ships as a pure openai-compatible provider (recipe file + register, no gateway shim) per the rule in `src/core/ai/recipes/index.ts`.

- `codestral-embed` / `codestral-embed-2505`, 1536d default, code-optimized, `MISTRAL_API_KEY` auth
- pricing entries (`$0.15/1M`) + a row in `docs/integrations/embedding-providers.md`
- hermetic recipe-shape test (mirrors `recipe-minimax` / `zeroentropy-recipe`)

Left VERSION/CHANGELOG to the maintainer per the fix-wave process.

## Verified end-to-end against the live Mistral API

- Raw `/v1/embeddings`: OpenAI-shaped response, **1536d** default.
- Through gbrain's gateway: `init` at 1536d + `put` + `embed` → chunks embedded, **no 422** (the standard openai-compatible path sends a clean request for this recipe).
- Semantic retrieval: query *"what compensates a shipowner for port delay?"* returned the charter-party doc at **0.865** (true vector match, not keyword overlap).
- Confirmed Mistral **rejects** OpenAI's `dimensions` (HTTP 422) and **honors** `output_dimension`. That is why flexible dims (up to 3072) + `mistral-embed` (1024d) are scoped as a follow-up that adds a `mistralCompatFetch` (mirroring `voyageCompatFetch`) rather than shipped in v1. v1 is fixed at the 1536 default.

## Test plan

- [x] `bun test test/ai/recipe-mistral.test.ts` (6 cases) + `recipes-existing-regression` green (17 pass / 0 fail)
- [x] Live embed + semantic query against `api.mistral.ai` (1536d, 0.865 retrieval)
- [ ] CI on a fork PR will not have `MISTRAL_API_KEY`; the bundled test is hermetic by design (recipe shape + pricing lookup, no live call)

### DIFF
--- a/docs/integrations/embedding-providers.md
+++ b/docs/integrations/embedding-providers.md
@@ -30,6 +30,7 @@ The resolved provider + dimensions get persisted to `~/.gbrain/config.json` atom
 | `google` | `GOOGLE_GENERATIVE_AI_API_KEY` | 768 | 0.025 | no | no |
 | `azure-openai` | `AZURE_OPENAI_API_KEY`, `AZURE_OPENAI_ENDPOINT`, `AZURE_OPENAI_DEPLOYMENT` | 1536 | 0.13 | no | no |
 | `minimax` | `MINIMAX_API_KEY` | 1536 | 0.07 | no | no |
+| `mistral` | `MISTRAL_API_KEY` | 1536 | 0.15 | no | no |
 | `dashscope` | `DASHSCOPE_API_KEY` | 1024 | varies | no | no |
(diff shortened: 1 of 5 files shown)


garrytan:

PR #1518, opened 2026-05-26
fix(gateway): handle embed preflight with user-provided models and reasoning model output

### LINKED ISSUE #1516: fix(gateway): diagnoseEmbedding rejects user-provided models for user_provided_models recipes (llama-server)
## Summary

`diagnoseEmbedding()` in `src/core/ai/gateway.ts` returns `user_provided_model_unset` for any recipe where `touchpoints.embedding.models` is an empty array and `user_provided_models === true`, even when the user has already configured an explicit model via `embedding_model`.

## Reproduction

1. Configure a brain with `llama-server` as the embedding provider:
   ```
   embedding_model: llama-server:text-embedding-qwen3-embedding-0.6b
   ```
2. Run `gbrain doctor` or any autopilot cycle that triggers embed preflight:
   ```
   gbrain doctor --remediate --yes
   ```
3. The preflight returns `user_provided_model_unset` despite the model being fully specified.

## Root Cause

`diagnoseEmbedding()` resolves `embedding_model` via `resolveRecipe()`, which parses the model string into `{ providerId, modelId }`. But the subsequent guard at line 670 only checks `tp.models.length === 0` — it never verifies that `parsed.modelId` (the user-provided model id) is actually populated.

```typescript
// Current — incomplete guard
if (
  Array.isArray(tp.models) &&
  tp.models.length === 0 &&
) {
}
```

## Fix

Add `!parsed.modelId` to the condition. If the model string carries a non-empty model id after the colon, the user has provided a model — the preflight should pass.

See PR # for the implementation.

### LINKED ISSUE #1517: fix(gateway): handle reasoning model output falling through response normalization
## Summary

The `chat()` function in `src/core/ai/gateway.ts` normalizes AI SDK v6 `generateText` results into internal `ChatBlock[]`. It processes `result.content` typed parts (text, tool-call), but skips `type: "reasoning"` parts and does not check `result.reasoningText` in the fallback path. When a reasoning model (e.g. DeepSeek with thinking/reasoning enabled) returns content primarily through reasoning channels, the blocks array remains empty and the caller receives an empty text response.

## Reproduction

1. Configure a brain with a DeepSeek reasoning model that returns `reasoning_content`:
   ```
   chat_model: deepseek:deepseek-v4-flash
   ```
2. Run a cycle phase that calls `gateway.chat()` (e.g. proposal, significance judge)
3. The generated text is empty, causing downstream failures.

## Root Cause

The response normalization at line 2408 only handles:
- `part.type === "text"` — text content
- `part.type === "tool-call"` — tool invocation

AI SDK v6 also returns `type: "reasoning"` parts (the `ReasoningPart` type) for reasoning-model responses. The flat `result.text` fallback similarly ignores `result.reasoningText`.

## Fix

- In the `result.content` branch: treat `type: "reasoning"` parts as text blocks
- In the flat fallback: when `result.text` is empty, check `result.reasoningText` as an alternative source

See PR # for the implementation.

justemu:
## Description

Two fixes in `src/core/ai/gateway.ts` addressing bugs found during autopilot diagnosis:

### Fix 1: Embed preflight rejects user-provided models (closes #1516)

When a brain is configured with `embedding_model: llama-server:text-embedding-qwen3-embedding-0.6b`, `diagnoseEmbedding()` returns `user_provided_model_unset` even though the model string is fully specified. The guard checks `tp.models.length === 0` without verifying whether the resolved `parsed.modelId` is populated.

**Fix:** Add `!parsed.modelId` to the condition. A non-empty model id after the colon satisfies the "user provided a model" requirement.

### Fix 2: Reasoning model output falls through response normalization (closes #1517)

The `chat()` response normalization skips `type: "reasoning"` parts in `result.content` and does not check `result.reasoningText` in the flat fallback. When a reasoning model (DeepSeek, etc.) returns content through reasoning channels, the blocks array remains empty.

**Fix:**
- In the `result.content` branch: treat `type: "reasoning"` parts as text blocks
- In the flat fallback: when `result.text` is empty, check `result.reasoningText`

## Related

- Issue #1516
- Issue #1517


justemu:
**Update:** This PR has been rebased onto v0.41.26.0 and expanded with an additional improvement.

### Changes included (now at HEAD 2aed39b3)

1. **Embed preflight fix** — prevents `user_provided_model_unset` false-positive when a user-specified model string is used with a `user_provided_models` touchpoint.

2. **Reasoning model output normalization** — handles AI SDK v6 `reasoning` part types and `reasoningText` fallback so reasoning models produce non-empty text output.

3. **Thinking mode disabled for all openai-compatible providers** — DeepSeek v4-flash defaults to thinking mode (equivalent to the deprecated `deepseek-reasoner`), wasting ~30-50% output tokens on chain-of-thought for non-reasoning use cases. The `@ai-sdk/openai-compatible` provider spreads `providerOptions[providerId]` into the request body, so `"thinking": {"type": "disabled"}` is injected into every chat call. Providers without thinking silently ignore the parameter.

### Rebase status
- Base: v0.41.14.0 → v0.41.26.0 (13 commits ahead)
- Conflict-free merge — `gateway.ts` was untouched between v0.41.14 and v0.41.26
- 9 schema migrations applied (v97→v107)

Best regards

garrytan:
Thanks for this! The embedding-preflight fix is being consolidated into #1938 (covers both reject sites + the dim validator, with tests). This branch also carries ~34k lines of churn that has since landed separately on master (conversation-parser, type-unification, onboard, doctor), so it is unmergeable as-is. Closing as a duplicate/superseded — credit noted. Reopen a scoped version if #1938 misses your case.

PR #1523, opened 2026-05-27
fix(extract): exponential backoff for batch-flush retries on pooler circuit-breaker

garrytan-agents:
## Problem

On a production brain with 16K+ pages, the `extract` dream-cycle phase runs 10+ minutes of sustained batch inserts (links + timeline entries, 100-row batches via `addLinksBatch` / `addTimelineEntriesBatch`). During these long runs, Supabase Supavisor (session-mode pooler, port 5432) periodically recycles backend connections.

The existing retry mechanism (v0.41.2.1, `withRetry`) does a **single 500ms retry** per batch flush. This was designed for PgBouncer's transaction-mode connection recycling where a single retry catches the stale handle. However, when Supavisor's circuit breaker activates (ECIRCUITBREAKER), recovery takes **5-10 seconds** — far longer than 500ms. The result:

- First retry fires at 500ms → circuit breaker still active → batch lost
- Subsequent batches hit the same circuit breaker → cascade of losses
- In production, **~3,000+ rows lost per dream cycle** (links + timeline entries)

```
[extract.links_inc] connection blip, retrying 100 rows in 500ms (Connection terminated unexpectedly)
  link batch error (100 rows lost): Connection terminated unexpectedly
  link batch error (100 rows lost): Connection terminated unexpectedly
  timeline batch error (100 rows lost): Connection terminated unexpectedly
  ...repeats for ~30 batches until connection fully dies...
```

## What We Tried

1. **Running phases sequentially with lock clearing** — helped with lock contention but didn't fix the batch-loss cascade within a single phase
2. **Switching to Supabase direct connection** (port 5432 session mode → `db.<ref>.supabase.co:5432` direct) — improved stability but the pooler path is still the default and needs to be robust
3. **Investigated `GBRAIN_DISABLE_DIRECT_POOL`** — the kill-switch forces all queries through the pooler read pool, which is the default path for most operators

## Solution

Upgrade `withRetry` to support **exponential backoff with configurable retry count**:

### `withRetry` changes (`src/commands/extract.ts`)

```typescript
// New opts
interface WithRetryOpts {
  onRetry?: (attempt: number, err: unknown) => void;
  delayMs?: number;      // base delay, default 500
  delayMaxMs?: number;   // cap, default 8000
  maxRetries?: number;   // default 1 (back-compat)
}

// Retry loop: base * 2^attempt, capped at delayMaxMs
for (let attempt = 0; attempt <= maxRetries; attempt++) {
  try { return await fn(); }
  catch (err) {
    if (!isRetryableConnError(err)) throw err;
    if (attempt >= maxRetries) break;
    const delay = Math.min(baseDelay * 2**attempt, maxDelay);
    await sleep(delay);
  }
}
```

### Extract batch-flush config

All 6 batch-flush sites now use:
```typescript
const BULK_MAX_RETRIES = 3;    // 500ms → 1s → 2s → give up
const BULK_RETRY_BASE_MS = 500;
const BULK_RETRY_MAX_MS = 8_000;
```

This covers the typical Supavisor circuit-breaker recovery window (5-10s) with total backoff time of ~3.5s before exhausting retries.

### Back-compatibility

Default `maxRetries=1` preserves the exact v0.41.2.1 behavior for all existing callers (single 500ms retry). Only the extract batch-flush sites opt in to the higher retry count.

## Testing

20/20 tests pass in `test/extract-batch-retry.test.ts`:

- Existing v0.41.2.1 tests unchanged (default maxRetries=1 behavior)
- New: `maxRetries=3` exhaustion (4 total calls)
- New: `maxRetries=3` recovery on third retry
- New: exponential backoff timing verification
- New: `delayMaxMs` caps delay growth

```
bun test test/extract-batch-retry.test.ts
20 pass, 0 fail, 47 expect() calls
```

## Files Changed

| File | Change |
|------|--------|
| `src/commands/extract.ts` | `withRetry` exponential backoff + bulk retry constants + 6 callsite updates |
| `test/extract-batch-retry.test.ts` | 5 new test cases for multi-retry + backoff behavior |

### DIFF
--- a/src/commands/extract.ts
+++ b/src/commands/extract.ts
@@ -55,6 +55,18 @@ import { parseWorkers, resolveWorkersWithClamp } from '../core/sync-concurrency.
 // small (a malformed row aborts at most 100, not thousands).
 const BATCH_SIZE = 100;
 
+// v0.41.18 — bulk retry config for long-running extract workloads.
+// On a 16K-page brain, extract runs 10+ minutes of sustained batch inserts.
+// Supabase Supavisor (session-mode pooler, port 5432) can recycle the backend
+// connection mid-run; rapid reconnect attempts trigger the pooler's circuit
+// breaker (ECIRCUITBREAKER) which needs 5-10s to recover — far longer than
+// the original single 500ms retry. 3 retries with exponential backoff
+// (500ms → 1s → 2s) covers the typical recovery window without amplifying
+// real outages.
+const BULK_MAX_RETRIES = 3;
+const BULK_RETRY_BASE_MS = 500;
+const BULK_RETRY_MAX_MS = 8_000;
+
 // v0.41.2.1 — batch-flush retry primitive (closes PR #1416's ~30% batch-loss
 // bug). PgBouncer transaction-mode poolers recycle backend connections between
 // queries; the next query through a stale handle throws a retryable connection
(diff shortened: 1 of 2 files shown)


garrytan:
Superseded by **v0.41.18.0 Supavisor Retry Cathedral** (incoming PR on branch `garrytan/fredericton`).

The cathedral wave absorbs your fix verbatim and extends it structurally:

- **Architectural pivot:** wraps retry INSIDE `postgres-engine.ts` + `pglite-engine.ts` batch primitives (`addLinksBatch` / `addTimelineEntriesBatch` / `upsertChunks`) so every caller — current AND future — inherits retry for free without per-site wrapping.
- **Codex-hardened defaults:** `BULK_RETRY_OPTS` becomes `{maxRetries:3, delayMs:1000, delayMaxMs:10000, jitter:'decorrelated'}`. Total worst-case wait ≈ 12s — covers the full 5-10s Supavisor circuit-breaker recovery window your PR identified, where the original `500/1000/2000` shape could still exhaust at ~3.5s.
- **Cathedral scope:** also covers `operations.ts:810` (MCP `put_page` auto-link, hit on every agent write — unprotected today), `extract.ts:1298` (--by-mention path), `backfill-base.ts` (unified retry orchestrator), and `sync.ts`/`reindex.ts`/`reindex-multimodal.ts`.
- **Observability:** audit JSONL via the shared `audit-writer` primitive + `batch_retry_health` doctor check + stress regression test (100 batches × 30% blip × both engines).
- **AbortSignal:** retry sleeps abort cleanly on SIGTERM so deploys aren't blocked.

Your **commit subject + co-authorship** preserved via `Co-Authored-By: @garrytan-agents` trailer on the merge commit. Your 5 new test cases move to `test/core/retry.test.ts` with assertions adjusted for the new defaults.

Plan + 11-task breakdown: see `~/.claude/plans/system-instruction-you-are-working-smooth-gosling.md` in the implementer's worktree.

Thank you for the fix and the writeup — the codebase is better for it.

Issue #1527, opened 2026-05-27
Bug: hybrid-reranker-integration.serial.test.ts is non-hermetic (reads ~/.gbrain config)


tonychen15:
**What happened?**
The four `hybridSearch — reranker enabled (reorder)` tests fail on any machine whose `~/.gbrain/config.json` sets `embedding_dimensions` to something other than the test's hardcoded 1536 (e.g. text-embedding-3-small at 1280). They pass in CI only because a fresh runner has no config file. Deterministic given a config: green in CI, red on a contributor's machine.

Root cause: the test stubs the gateway at 1536, but since v0.36.3.0 `hybridSearch` resolves the embedding column from `loadConfig()`, and the resolver precedence is `cfg.embedding_dimensions > gateway dims > default`. So a real config value outranks the test's gateway dims; the stub's 1536 vector then fails the gateway dim check, the error is swallowed, search falls back to keyword-only, and the reranker never runs (rerankerFn gets 0 docs, rerank_score undefined).

Fix (test-only): isolate `GBRAIN_HOME` in `beforeAll` to a tmpdir whose config matches the stub. PR to follow.

**What did you expect?**
Pass the unit test.

**Steps to reproduce**
1. with `~/.gbrain/config.json` `embedding_dimensions` != 1536:
2. `bun test test/search/hybrid-reranker-integration.serial.test.ts` → 2 pass / 4 fail.
3.

**Environment**
- gbrain version: v0.41.14.0 (test from v0.35.0.0, broken by v0.36.3.0).
- OS:  Ubuntu 25.10
- Bun version: 1.3.11
- Database: Supabase / self-hosted Postgres

**`gbrain doctor --json` output**
```json
(paste output here)
```

PR #1528, opened 2026-05-27
fix(test search): make reranker integration test hermetic.

tonychen15:
Since v0.36.3.0 hybridSearch resolves the embedding column from loadConfig(), so this .serial test (which stubbed only the gateway at 3-large@1536) broke on dev machines whose ~/.gbrain/config.json uses different dims: the stubbed embed threw a dim-mismatch, the vector path swallowed it, search fell back to keyword-only, and the reranker never ran (green on CI, red locally). Isolate GBRAIN_HOME to a tmpdir with a matching config so the resolver agrees with the stub. No product change.
#1527


garrytan:

Issue #1526, opened 2026-05-27
Data loss: `recloneIfMissing` rm-rfs user-managed source path; ungated against arbitrary local_path


Oszkar:
# Data loss: `recloneIfMissing` rm-rfs user-managed source path

**Severity:** Critical / data loss
**Repo:** garrytan/gbrain
**Affected versions:** confirmed in v0.41.14.0 (commit `32f8be96`, where the data loss was observed) AND v0.41.16.0 (commit `f702ec05`, latest at time of filing — same code unchanged) — likely all versions since `recloneIfMissing` was added
**Platform observed:** Windows 11 Pro 10.0.26200, Bun runtime, PowerShell + Git Bash

## Summary

`recloneIfMissing()` in `src/core/sources-ops.ts:645-680` performs an **ungated `rm -rf` against `src.local_path`** when called for a source that has `remote_url` set in its config. There is no check that `src.local_path` is inside a gbrain-managed `~/.gbrain/clones/` directory — the function will destructively re-clone over an arbitrary path, including the user's working tree, if the source row's `local_path` points there.

This wiped my working tree, including `.git/`, `backend/`, `docs/`, `frontend/*`, `Dockerfile`, `Cargo.toml`, `package.json`, `CLAUDE.md`, `AGENTS.md`, and `.env*`. Survivors were dirs that happened to have open file handles during the rm-rf (Windows partial failure).

## Root cause: ungated rm-rf

`src/core/sources-ops.ts:645-680`:

```typescript
export async function recloneIfMissing(engine, id) {
  const src = await fetchSourceRow(engine, id);
  if (!src) throw new SourceOpError('not_found', ...);
  const remoteUrl = getRemoteUrl(src.config);
  if (!remoteUrl || !src.local_path) return false;

  const state = validateRepoState(src.local_path, remoteUrl);
  if (state === 'healthy') return false;

  // Re-clone via temp + rename
  const tempDir = makeTempCloneDir(id);
  mkdirSync(dirname(tempDir), { recursive: true });
  try {
    cloneRepo(remoteUrl, tempDir);
  } catch (e) { /* rmSync tempDir, throw */ }

  // ⬇️ BUG: comment claims "nuke if partially exists"; code unconditionally rm-rf's
  // If the local_path partially exists (e.g., empty dir, file-not-dir), nuke
  // it before the rename so renameSync doesn't fail on a non-empty target.
  rmSync(src.local_path, { recursive: true, force: true });   // line 674
[...]

PR #634, opened 2026-05-05
fix(cli): CLI_ONLY commands should short-circuit on --help instead of executing

tmchow:
## Summary

`gbrain sync --help`, `gbrain doctor --help`, `gbrain init --help`, and other `CLI_ONLY` commands no longer fall through to live execution when invoked with `--help`. They now print a generic usage line and exit 0 without running.

## Why this matters

[Issue #578](https://github.com/garrytan/gbrain/issues/578) diagnosed the dispatcher in `src/cli.ts:57-67`: the per-command `--help` short-circuit only fired when `cliOps.get(command)` returned a shared op. For `CLI_ONLY` commands (`sync`, `doctor`, `init`, `autopilot`, ~50 entries total), `cliOps.get()` returned `undefined`, so the `if (op)` branch never executed and dispatch fell through to `handleCliOnly()`.

@jbaham2 confirmed the same bug class on `gbrain init --help` with destructive impact: their existing postgres-mode `~/.gbrain/config.json` was overwritten by the init default (PGLite), silently disconnecting them from a populated brain. Recovery required setting `GBRAIN_DATABASE_URL` to override the now-broken config.

## Approach

Smallest-diff fix matching the plan's Option A:

1. Added `CLI_ONLY_SELF_HELP` allowlist for commands whose handlers already implement `--help` correctly. Verified by reading each handler's source:
   - `upgrade`, `post-upgrade`, `check-update`
   - `embed`, `config`
   - `skillpack`, `skillpack-check`
   - `integrations`, `friction`
2. When the dispatcher sees `--help`/`-h` for a `CLI_ONLY` command not in the allowlist, it prints a generic usage line via `printCliOnlyHelp(command)` and returns BEFORE `handleCliOnly()` is reached.
3. Allowlisted commands fall through to `handleCliOnly()` so their detailed per-command and per-subcommand help (`gbrain skillpack install --help`, `gbrain integrations show --help`, etc.) stays reachable.

The trade-off: commands not in the allowlist get a generic placeholder rather than hand-rolled per-command usage. That seems like the right default given the size of `CLI_ONLY` (~50 entries) and that the issue specifically called out destructive fall-through as the problem to solve.

If you'd prefer a different cut (e.g., short-circuit only the explicitly-confirmed list of `sync`/`doctor`/`init`/`autopilot`, or expand the allowlist further), happy to adjust.

## Testing

[...]

Issue #578, opened 2026-05-03
bug: `gbrain sync --help` fires a live sync instead of printing help (CLI_ONLY commands without per-command help-handling fall through to live execution)


emanuelubert:
## Summary

Invoking `--help` on `gbrain sync` (and several other `CLI_ONLY` commands) triggers a live operation against the configured brain root instead of displaying help text. This is a structural CLI-dispatch bug: anyone exploring the CLI to understand `sync`'s behavior accidentally triggers a live sync.

## Repro (minimal, sandboxed)

```bash
export GBRAIN_HOME=/tmp/.gbrain-bug-repro
mkdir -p $GBRAIN_HOME && echo "# test" > $GBRAIN_HOME/test.md
gbrain sync --help
```

## Expected

Standard help output: usage, flags, examples. No side effects.

## Actual

Live sync executes against `$GBRAIN_HOME/`:
```
$ gbrain sync --help
Already up to date.
$ echo $?
0
```
No help text printed; sync ran and short-circuited at `up_to_date`. On a fresh / non-empty brain root this would consume embedding compute + wall-clock + DB writes.

## Affected version

gbrain 0.22.6 (`bun bin/gbrain --version`). Confirmed reproduces on local checkout + likely on `master` HEAD `3c032d79` (2026-05-03) — relevant code path unchanged.

## Other affected commands (matrix)

| Command | Class | `--help` behavior |
|---|---|---|
| `gbrain --help` | root | ✅ prints help |
| `gbrain query --help` | shared op (`cliOps`) | ✅ prints help via `printOpHelp(op)` |
| `gbrain embed --help` | `CLI_ONLY` | ✅ prints usage (own handler implements it) |
| `gbrain config --help` | `CLI_ONLY` | ✅ prints usage (own handler implements it) |
| **`gbrain sync --help`** | `CLI_ONLY` | 🐛 **fires live sync** |
| **`gbrain doctor --help`** | `CLI_ONLY` | 🐛 **runs full doctor diagnostic suite** |

The bug affects `CLI_ONLY` commands that don't implement their own `--help` short-circuit in their handler. `sync` and `doctor` are confirmed; the other ~40 entries in `CLI_ONLY` (`autopilot`, `eval`, `agent`, `migrate`, `extract`, `dream`, `serve`, `import`, `export`, ...) likely fall in the same category unless individually audited.

## Root cause (`src/cli.ts` lines 57-70)

```typescript
// Per-command --help
[...]

PR #1529, opened 2026-05-27
fix(autopilot): translate positional subcommands (closes #1525)

### LINKED ISSUE #1525: UX trap: `gbrain autopilot status` starts the daemon instead of reporting status
# `gbrain autopilot status` starts the daemon instead of reporting status

**Severity:** High (UX trap with safety implications)
**Repo:** garrytan/gbrain
**Affected version:** all versions; v0.41.14.0 observed, v0.41.16.0 verified unchanged at HEAD

## Summary

`gbrain autopilot status` (entered as a positional subcommand) silently **starts the autopilot daemon** rather than reporting status. The documented status check is `gbrain autopilot --status` (with the double-dash flag form). The positional form has no special handling, so it falls through to "start daemon" with `status` treated as an extra arg (likely interpreted as repo path or ignored).

This is a high-severity UX bug because:
1. Operators and AI coding assistants intuitively try `gbrain autopilot status` first (most modern CLIs accept subcommand-style status).
2. Starting the daemon has real consequences - it dispatches `sync` jobs against every source in the DB, which can trigger destructive code paths (see companion data-loss issue).
3. There is no warning that you've started the daemon. The output looks like a launch banner, which can be mistaken for status output by a casual reader.

## Reproduction

```bash
# Expectation: report current autopilot state (running? lock file? last cycle?).
gbrain autopilot status

# Actual behavior:
# Autopilot starting. Repo: <cwd>, interval: 300s
# [autopilot] Minions worker spawned (pid: -1, watchdog: 2048MB)
# [dispatch] job #2 sync (freshness: <source-id>; age=<X>min)
# ... daemon stays running ...
```

The user then has to discover that the only way to actually check status is `gbrain autopilot --status`, which is non-obvious from `gbrain autopilot --help`:

Note:  The bug only fires when `sync.repo_path` is set in config (positional `status` falls through, `parseArg(args, '--repo')` returns `undefined`, then falls back to `engine.getConfig('sync.repo_path');` if that's unset, autopilot exits with "No repo path" instead of starting).

```
Usage: gbrain autopilot

gbrain autopilot - run gbrain --help for the full command list.
```

The CLI's own help text doesn't surface `--status`. It's only documented in the source file's JSDoc:

```
src/commands/autopilot.ts lines 14-19:
 * Usage:
 *   gbrain autopilot [--repo <path>] [--interval N] [--json] [--inline]
 *   gbrain autopilot --install [--repo <path>]
 *   gbrain autopilot --uninstall
 *   gbrain autopilot --status [--json]
```

## Real-world impact

During incident investigation, I ran `gbrain autopilot status` to confirm whether autopilot was the culprit for a data-loss event. The command **started a new autopilot instance** that immediately queued a sync job against the source whose `local_path` was a real user repo. The destructive code path in `recloneIfMissing` would have fired again if the spawn-worker chain hadn't been broken on this Windows shell by an unrelated `uv_spawn ENOENT` issue.

Sequence:

```
$ gbrain autopilot status
Autopilot starting. Repo: E:/Programming/degner, interval: 300s
[autopilot] Minions worker spawned (pid: -1, watchdog: 2048MB)
[autopilot] worker spawn failed (async): ENOENT: no such file or directory, uv_spawn '/c/Users/oszka/.bun/bin/gbrain' (code=ENOENT)
[dispatch] job #2 sync (freshness: gstack-code-degner-785bc61d; age=29663579min)
```

If `uv_spawn` had succeeded, the dispatched `sync` job would have run `gbrain sync --strategy code --source gstack-code-degner-785bc61d` and potentially re-triggered the data-loss path on whatever `local_path` was registered for that source.

## Proposed fix

Three options, easy → thorough:

**Minimum:** Accept positional subcommands as aliases for flags. `gbrain autopilot status` → run `--status` handler. `gbrain autopilot start` → start daemon (current behavior). `gbrain autopilot stop` → terminate daemon and clear lock. `gbrain autopilot install` / `uninstall` → corresponding flag handlers. Print a deprecation hint if positional form is used, suggesting the canonical flag form.

**Minimum+:** Like above but fail loudly.

```ts
  // After flag parsing, before lock acquisition:
  const positionals = args.filter(a => !a.startsWith('--') && /* not a flag value */);
  if (positionals.length > 0) {
    console.error(`Unknown subcommand: ${positionals[0]}. Did you mean --${positionals[0]}?`);
    process.exit(2);
  }
```

**Better:** Update `gbrain autopilot --help` so the existing flag form is discoverable. Currently the help block is one line of "run gbrain --help for the full command list" - surface the JSDoc's actual usage block at runtime.

**Best:** Both of the above PLUS require explicit consent before starting the daemon. `gbrain autopilot` alone (no args) should print state + the start command, not just start. Use `gbrain autopilot start` or `gbrain autopilot --daemon` for the actual launch. This shifts the safety burden off the user.

## Cross-references

- **Related: #578** - "`gbrain sync --help` fires a live sync instead of printing help (CLI_ONLY commands without per-command help-handling fall through to live execution)". Same class of UX trap: gbrain's CLI dispatcher falls through to "live execution" when it receives an unexpected positional/flag combo. Different command (`sync` vs `autopilot`), same root cause. A central fix (require recognized subcommand or print help) would close both.
- Critical data-loss issue (companion): `recloneIfMissing` rm-rfs user-managed source paths - this trap triggered the bug when I inadvertently re-spawned autopilot while diagnosing the incident.

> Note: PR #634 fixed the `--help/-h` surface of #578 by adding `CLI_ONLY_SELF_HELP` allowlist + generic help short-circuit in `src/cli.ts:111-121`. #1525 is the unknown-positional surface of the same family, not covered by #634 - `gbrain autopilot status` is `argv = ['status']`, not `argv = ['--help']`, so it bypasses the help short-circuit and reaches `runAutopilot()` which only branches on flags.

Oszkar:
## Summary

- `gbrain autopilot status` silently started the daemon instead of reporting status. `runAutopilot()` only branched on flag forms, so positionals fell through to the default "start daemon" path.
- This PR adds a pure `translatePositionalSubcommands()` function that runs before any side effect (lockfile, daemon spawn, sync dispatch). Known positionals (`status`, `install`, `uninstall`, `start`) translate to their flag equivalents; unknown positionals fail loud with a structured error.
- Closes #1525.

## What's in the patch

`src/commands/autopilot.ts`:
- New exported pure function `translatePositionalSubcommands(args)` returning a discriminated `{ ok: true, args } | { ok: false, error }` so it's trivially testable without IO.
- Value-flag aware: `--repo /foo status` correctly treats `/foo` as the repo path and `status` as the positional, not vice versa. `--repo status` (literal path) passes through unchanged.
- Wired in `runAutopilot` right after the `--help` short-circuit and BEFORE the lockfile/daemon paths, so the fix-loud rejection happens before any side effect.
- `--help` text gains a "Subcommand aliases" block so the canonical flag forms are discoverable from the CLI itself, not just JSDoc.

`test/autopilot-positional-subcommands.test.ts` (new, 22 cases):
- Pure-function tests, no env mutation, no `mock.module` — passes `check-test-isolation.sh`.
- Iron-rule regression guard: first case is the exact ticket repro (`['status']` -> `['--status']`).

## Scope notes

- **`stop` is intentionally NOT aliased.** The ticket lists it as a potential alias, but stopping a running daemon is a new behavior (read PID from lock, send SIGTERM, await drain) that deserves its own design pass. Users typing `gbrain autopilot stop` today get the unknown-positional error rather than the silent daemon-start. Happy to land `stop` as a separate PR if there's interest.
- **Deprecation hints not added.** The ticket suggests a hint when positional form is used; I left it off because it's unclear whether the maintainer wants the positional form to eventually go away or stay supported indefinitely (consistent with modern CLI grammar). Easy to add in review if desired.
- **`--repo VALUE` and `--interval VALUE` only.** Those are the only value-flags in this command. Other autopilot flags are bare. If a future flag needs a value, add it to the `AUTOPILOT_VALUE_FLAGS` set.

## Test plan

- [x] `bun test test/autopilot-positional-subcommands.test.ts` — 22/22 pass
- [x] `bun test test/autopilot-*.test.ts` — all 109 existing autopilot tests pass (no regression)
- [x] `bash scripts/check-test-isolation.sh` — clean (file passes R1/R2/R3/R4 isolation rules)
- [x] `bash scripts/check-jsonb-pattern.sh` — clean
- [x] `bash scripts/check-progress-to-stdout.sh` — clean
- [ ] Manual repro of the original bug pre-fix and post-fix:
  - `gbrain autopilot status` pre-fix → starts daemon (the ticket's bug)
  - `gbrain autopilot status` post-fix → routes to `showStatus()`, prints state, returns
- [ ] Manual `gbrain autopilot stop` post-fix → fails loud with "Unknown subcommand: \`stop\`" + list of allowed subcommands (does NOT start the daemon)

## Related

- Companion data-loss issue: #1526 (independent fix; targeting a separate PR after this one lands so each fix gets its own clean review surface).
- Family parent: #578 (partially fixed by #634 for the `--help` surface; this PR closes the unknown-positional surface).

### DIFF
--- a/src/commands/autopilot.ts
+++ b/src/commands/autopilot.ts
@@ -147,7 +147,11 @@ const AUTOPILOT_POSITIONAL_ALIASES: Record<string, string | null> = {
 
 export type PositionalTranslation =
   | { ok: true; args: string[] }
-  | { ok: false; error: string };
+  | {
+      ok: false;
+      reason: 'unknown_subcommand' | 'multiple_subcommands';
+      message: string;
+    };
 
 export function translatePositionalSubcommands(args: string[]): PositionalTranslation {
   const out: string[] = [];
@@ -178,7 +182,8 @@ export function translatePositionalSubcommands(args: string[]): PositionalTransl
       const known = Object.keys(AUTOPILOT_POSITIONAL_ALIASES).join(', ');
       return {
         ok: false,
-        error: `Multiple subcommands given. Use only one of: ${known}.`,
+        reason: 'multiple_subcommands',
+        message: `Multiple subcommands given. Use only one of: ${known}.`,
       };
     }
     positionalSeen = true;
@@ -191,7 +196,8 @@ export function translatePositionalSubcommands(args: string[]): PositionalTransl
     const known = Object.keys(AUTOPILOT_POSITIONAL_ALIASES).join(', ');
     return {
       ok: false,
-      error:
+      reason: 'unknown_subcommand',
+      message:
         `Unknown subcommand: \`${a}\`.\n` +
         `Allowed subcommands: ${known}.\n` +
         `Or use the flag form: --status, --install, --uninstall.\n` +
@@ -226,7 +232,7 @@ export async function runAutopilot(engine: BrainEngine, args: string[]) {
   // the daemon.
   const translated = translatePositionalSubcommands(args);
   if (!translated.ok) {
-    console.error(translated.error);
+    console.error(translated.message);
     process.exit(2);
   }
   args = translated.args;
--- a/src/commands/autopilot.ts
+++ b/src/commands/autopilot.ts
@@ -116,6 +116,91 @@ export function shouldSpawnAutopilotWorker(args: string[]): boolean {
   return !args.includes('--no-worker');
 }
 
+/**
+ * v0.41.x #1525 — positional subcommand translation.
+ *
+ * Pre-fix, `gbrain autopilot status` silently fell through to "start daemon"
+ * because `runAutopilot()` only branched on flag forms (`--status`, etc.).
+ * `status` was treated as a stray positional and ignored.
+ *
+ * This translator maps known positional subcommands to their flag form so
+ * `autopilot status` is equivalent to `autopilot --status`, then rejects
+ * any unrecognized positional with a fail-loud error before any side
+ * effect (lockfile, daemon spawn, sync dispatch) runs.
+ *
+ * Scope decisions:
+ *   - Known aliases: `status` → `--status`, `install` → `--install`,
+ *     `uninstall` → `--uninstall`, `start` → (drop; default daemon launch).
+ *   - `stop` is intentionally NOT aliased here. Stopping a running daemon
+ *     is a new behavior (read PID from lock, SIGTERM, drain) that deserves
+ *     its own design and PR. Users typing `gbrain autopilot stop` today get
+ *     the unknown-positional error with the canonical alternatives.
+ *   - At most one positional allowed; multiple positionals fail loud.
+ */
+const AUTOPILOT_VALUE_FLAGS = new Set(['--repo', '--interval']);
+const AUTOPILOT_POSITIONAL_ALIASES: Record<string, string | null> = {
+  status: '--status',
+  install: '--install',
+  uninstall: '--uninstall',
+  start: null, // drop the positional; default behavior is daemon launch
+};
+
+export type PositionalTranslation =
+  | { ok: true; args: string[] }
+  | { ok: false; error: string };
+
+export function translatePositionalSubcommands(args: string[]): PositionalTranslation {
+  const out: string[] = [];
+  let positionalSeen = false;
+  let i = 0;
+  while (i < args.length) {
+    const a = args[i];
+    if (AUTOPILOT_VALUE_FLAGS.has(a)) {
+      // Pass through the flag and its value untouched. If the value is
+      // missing at end-of-argv, fall through so the existing parseArg
+      // path can report the broken usage.
+      out.push(a);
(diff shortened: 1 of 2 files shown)


time-attack:
Thanks — you diagnosed the real behavior correctly, `gbrain autopilot status` does fall through to the daemon-start path on master, and the pure-function/no-mock test approach is nicely disciplined. The fix is bigger than the papercut: it introduces a general positional-subcommand grammar with a value-flag-aware tokenizer and an alias table, and it converts previously-ignored stray arguments into a hard error, which is a behavior change for existing scripts. We're keeping the CLI grammar as-is rather than adding a second accepted form, so closing on surface-area policy.

PR #1540, opened 2026-05-27
fix: brainstorm/lsd judge truncation + slash-prefix pricing lookup

garrytan-agents:
Two bugs causing `judge_failed` on every brainstorm/lsd run:

### Bug 1: maxTokens truncation
`maxTokens` hard-coded at 4000 in `judges.ts`. With 36-96 ideas per batch, each producing ~100 tokens of JSON output (id, 5 axis scores, note), the response was consistently truncated mid-JSON → `parseJudgeJSON` failure.

**Fix:** Scale maxTokens to `ideas.length * 150 + 500` (min 4000).

### Bug 2: Slash-prefix pricing lookup
Pricing lookup in `anthropic-pricing.ts` and `budget-tracker.ts` only split on `:` (`anthropic:claude-sonnet-4-6`) but not `/` (`anthropic/claude-sonnet-4-6`). CLI `--judge-model` passes slash-separated IDs → no pricing match → `BudgetExhausted` with reason `no_pricing` when `--max-cost` is set.

**Fix:** Fall through to slash-split when colon-split finds nothing.

### Tested
Brainstorm run with 72 ideas now scores **39 passing** (was 0 of 72, every run since calibration cold-start).

### DIFF
--- a/src/core/anthropic-pricing.ts
+++ b/src/core/anthropic-pricing.ts
@@ -57,6 +57,12 @@ export function estimateMaxCostUsd(
     const tail = modelId.split(':', 2)[1];
     if (tail) p = ANTHROPIC_PRICING[tail];
   }
+  // v0.37.1: also handle slash-prefixed ids (e.g. "anthropic/claude-sonnet-4-6")
+  // which arrive via --judge-model and other CLI overrides.
+  if (!p && modelId.includes('/')) {
+    const tail = modelId.split('/', 2)[1];
+    if (tail) p = ANTHROPIC_PRICING[tail];
+  }
   if (!p) return null;
   return (
     (estimatedInputTokens / 1_000_000) * p.input +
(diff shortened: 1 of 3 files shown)


garrytan:
Superseded by #1562: brought the fix into a production-ready wave that centralizes model-id parsing across all 5 lookup sites (was 2) + extends the gateway resolver (`src/core/ai/model-resolver.ts:parseModelId`) to also accept slash form + adds per-model maxTokens caps.

Codex adversarial review during /ship caught that the pricing fix alone wouldn't close the bug class — it would just shift the failure from `BudgetExhausted no_pricing` to `AIConfigError 'missing a provider prefix'` mid-judge. The shipped version extends the gateway resolver to match.

Thanks @garrytan-agents — your bug report + first-pass diff drove the whole investigation. Credit lives in the v0.41.21.0 CHANGELOG entry and PR #1562 body.

Issue #1535, opened 2026-05-27
gbrain dream: intermittent 4-second exit-1 crash with no error output, leaves stale gbrain-cycle lock


Mr-B-1:
## Summary

`gbrain dream` (invoked via systemd timer cron) intermittently exits with code 1 after ~4 seconds of wall-time and ~7MB memory peak, having written no diagnostic output to its configured `StandardOutput`/`StandardError` log. The crash leaves the `gbrain-cycle` row in `gbrain_cycle_locks` un-released, compounding with #1534.

## Symptoms

From `journalctl -u gbrain-dream.service --no-pager`:

```
May 25 09:00:00 CloudTron systemd[1]: Starting gbrain-dream.service - GBrain nightly dream cycle...
May 25 09:34:07 CloudTron systemd[1]: Finished gbrain-dream.service - GBrain nightly dream cycle.
May 25 09:34:07 CloudTron systemd[1]: gbrain-dream.service: Consumed 27.721s CPU time.
May 26 09:00:01 CloudTron systemd[1]: Starting gbrain-dream.service - GBrain nightly dream cycle...
May 26 09:00:05 CloudTron systemd[1]: gbrain-dream.service: Main process exited, code=exited, status=1/FAILURE
May 26 09:00:05 CloudTron systemd[1]: gbrain-dream.service: Failed with result "exit-code".
May 26 09:00:05 CloudTron systemd[1]: gbrain-dream.service: Consumed 2.618s CPU time, 7.0M memory peak, 0B memory swap peak.
```

A successful run takes 30-35 min wall-time, ~28s CPU, ~440 MB memory peak. The failure mode is a 4-second exit with 7 MB memory peak — the process barely started.

A second instance: cycle attempt at 21:40 PST same day also died (PID 3917507 left a stale lock).

## What I tried

1. **Reproduce interactively** with the exact systemd ExecStart args:
   ```
   /usr/local/bin/gbrain-job.sh dream --dir /data/brain --source default
   ```
   Result: `Skipped: another cycle is already running. (locked)` — exit 0. (The previous crash left the lock held, so this path is benign and exits cleanly.)

2. **`gbrain dream --dry-run --json`** — exit 0, status: skipped, reason: cycle_already_running

3. **`env -i HOME=/root PATH=...` to strip env contamination** — same "Skipped" result

[...]

Issue #1470, opened 2026-05-26
runCycle silently swallows lock.release() errors, stranding rows in gbrain_cycle_locks


BrendanGahan:
## Summary

`runCycle`'s `finally` block at `src/core/cycle.ts` silently swallows errors from `lock.release()`:

```ts
} finally {
  if (lock) {
    try { await lock.release(); } catch { /* best-effort */ }
  }
}
```

When the release SQL throws (e.g. `CONNECTION_ENDED` from `postgres.js` when the pool has been ended out from under the cycle), the row in `gbrain_cycle_locks` persists with no operator-visible signal. The next `gbrain dream` invocation within the 30-minute TTL window then skips with `cycle_already_running` for reasons that look like a phantom — no error message points back to the original failure.

## Reproduction

Easiest path: get any other bug to terminate the postgres pool before the cycle's finally runs (see #C / "shared module singleton clobbered" issue for the cleanest path on `0.41.2.0`). The release attempt then throws `CONNECTION_ENDED`, but the user sees:
- Cycle exits code 0
- No mention of the release failure
- `psql gbrain -c "SELECT * FROM gbrain_cycle_locks;"` shows a stranded row
- Next `gbrain dream` within 30 minutes skips with `status='skipped' reason='cycle_already_running'`

## Suggested fix

Log the error to stderr:

```ts
} finally {
  if (lock) {
    try {
      await lock.release();
    } catch (e) {
      console.error('[cycle] lock.release() threw:', e instanceof Error ? e.message : String(e));
    }
  }
}
```

Doesn't change behavior on the success path. On failure, the operator gets a one-line signal instead of an invisible stranded lock. Diagnostic information turns a "why is dream broken?" debugging session into a one-glance fix.

## Why this matters

The empty-catch hid this codebase's largest single-symptom bug class for the entire life of `0.41.2.0` on Postgres. Operators see stale locks accumulate in `gbrain_cycle_locks` but have no signal that release is failing — and no signal pointing at the underlying connection state issue that's causing the throw. A `console.error` is the difference between a 30-second fix and a 4-hour bisect.

## Environment

- gbrain 0.41.2.0
- Bun 1.3.14
- Postgres (Apple Silicon, localhost)

## Related

See companion issue for the underlying singleton-clobber bug that originally surfaced this swallow.

Full diagnostic write-up + stack traces in our internal brain page; happy to share more detail if useful.

BrendanGahan:
[...]

Issue #1404, opened 2026-05-25
gbrain dream — all DB-dependent phases fail with "connect() has not been called"


SkyairOnline:
**gbrain version:** 0.41.0.0
**Engine:** Postgres (local standalone via Homebrew)
**Runtime:** Bun 1.2.22
**OS:** macOS (arm64)

## Description

Running `gbrain dream` runs all 17 phases, but every DB-dependent phase fails with `"No database connection: connect() has not been called"`. Filesystem-only phases (lint, backlinks) work fine.

## Reproduction

```bash
gbrain dream
```

Output (abbreviated):
```
  ! lint        0 fix(es) applied, 52 remaining          ← OK (FS-only)
  ✓ backlinks   1 missing back-link(s) found              ← OK (FS-only)
  ✗ sync        sync phase failed                        ← FAIL
      [InternalError/UNKNOWN] No database connection: connect() has not been called.
  ✗ synthesize  synthesize phase failed                  ← FAIL
      ...
  ✓ extract     0 link(s), 0 timeline entries            ← batch writes silently lost
  ✗ embed       embed phase failed                       ← FAIL
  ✗ orphans     orphans phase failed                     ← FAIL
  ...
```

## Key Observations

1. **Each individual phase works fine in isolation:**
   ```bash
   gbrain dream --phase sync      # ✅ works
   gbrain dream --phase embed     # ✅ works
   gbrain dream --phase extract   # ✅ works
   gbrain dream --phase orphans   # ✅ works
   ```

2. **`gbrain doctor` connects fine** — all 47 checks pass, 100/100 score.

3. **`gbrain init --url <url>` connects fine** — schema migration succeeds.

4. **Programmatic `runCycle(engine, ...)` also fails** when running all phases, but a single-phase (`['extract']`) run works.

5. **`GBRAIN_POOL_SIZE=2` doesn't help** — the CLI's `toEngineConfig()` doesn't pass `poolSize` to the engine config, so the engine always runs in module mode relying on the global `db.sql` singleton.

## Root Cause Analysis

The error originates from `src/core/db.ts:getConnection()`:

```typescript
export function getConnection() {
  if (!sql) {
    throw new GBrainError(
      'No database connection',
      'connect() has not been called',
      ...
    );
  }
  return sql;
}
```

The module-level `sql` variable is set by `connect()` at startup but becomes `null` by the time any DB phase executes in the full cycle. Individual phases (separate process) work because each gets a fresh module instance.

[...]

PR #1546, opened 2026-05-27
fix(postgres-engine): instance disconnect must not tear down module singleton

billy-armstrong:
## The bug

`PostgresEngine.disconnect()` calls `db.disconnect()` in its `'module'`-style branch:

```typescript
// src/core/postgres-engine.ts (current master, line 207-211)
  await db.disconnect();       // <-- this tears down the shared singleton
  this._connectionStyle = null;
}
```

That call tears down the **module-level singleton** — which is shared across the entire process — every time any instance engine that happens to share the singleton calls its own `disconnect()`.

In practice this fires during the dream cycle's **lint phase**:

1. `runPhaseLint` calls `runLintCore`
2. `runLintCore` calls `resolveLintContentSanity` (`src/commands/lint.ts:319`)
3. `resolveLintContentSanity` creates a fresh engine via `createEngine`, connects it with no explicit URL (so it shares the module singleton), runs a config probe
4. The `finally` block calls `engine.disconnect()`
5. That call kills the singleton the cycle's other phases are using

Every subsequent phase then runs against a dead pool. Visible symptoms:

- `extract.links_fs` silently drops link batches with `"connection blip, retrying 100 rows in 500ms (No database connection: connect() has not been called...)"` — ~191 rows lost per cycle on a medium-size source
- `conversation_facts_backfill` aborts with `"connect() has not been called"` 
- Cycle reports "partial" status; downstream consumers (search, salience, etc) see stale or missing data

The cycle's lint phase fires the disconnect even on `gbrain dream --source <default>` — the bug class triggers any time a fresh engine is created mid-cycle.

## Root cause pinned via diagnostic instrumentation

Added a single line to `src/core/db.ts:disconnect()` capturing a stack trace per call:

```typescript
export async function disconnect(): Promise<void> {
  const stack = new Error("db.disconnect called from").stack;
  process.stderr.write(\`[DIAG] db.disconnect at \${new Date().toISOString()}\n\${stack}\n\`);
  if (sql) { await sql.end(); sql = null; connectedUrl = null; }
}
```

Re-ran a per-source dream cycle. The first `[DIAG]` fires **during** the lint phase:

\`\`\`
[cycle.lint] start
[DIAG] db.disconnect at 2026-05-27T04:47:37.633Z
Error: db.disconnect called from
    at disconnect (src/core/db.ts:229)
    at disconnect (src/core/postgres-engine.ts:207)     ← THIS LINE
    at resolveLintContentSanity (src/commands/lint.ts:319)
    at async runLintCore (src/commands/lint.ts:395)
    at async runPhaseLint (src/core/cycle.ts:622)
    at async runCycle (src/core/cycle.ts:1355)
    at async runDream (src/commands/dream.ts:281)
[cycle.lint] done
\`\`\`

A second `[DIAG]` fires at process exit (the correct disconnect, from `cli.ts:1095` finally{}).

## The fix

Remove the `await db.disconnect()` call from the `'module'`-style branch. The module-level singleton's lifecycle is owned exclusively by the CLI top-level (`src/cli.ts` finally{} block) and explicit test setup — never by individual engine instances that happen to share it.

An instance-level disconnect just releases this instance's reference; the singleton stays alive for any sibling code paths still using it.

## Why v0.28.1's idempotency fix didn't catch this

The v0.28.1 contract comment on the early-return branch reads:

> "After this point, \`_connectionStyle\` stays 'instance' so a second disconnect() is a no-op rather than falling through and clearing the unrelated module-level db singleton."

That fix protected the **SECOND** disconnect call. The **FIRST** disconnect on a `'module'`-style engine still tore down the singleton — that's the bug class this PR closes.

## What about the existing test?

The existing test `test/e2e/postgres-engine-disconnect-idempotency.test.ts` already pins the second-call no-op contract. This PR adds a new test case asserting that an instance engine's **first** disconnect on a 'module'-style instance preserves the singleton:

\`\`\`typescript
test('v0.41+ lint-phase fix: module-style engine disconnect() does NOT tear down singleton (FIRST call)', async () => {
  // Re-establish module-level connection (the cycle's singleton).
  await db.connect({ database_url: DATABASE_URL! });
  // A fresh engine that shares the module singleton (mimics resolveLintContentSanity).
  const lintEngine = new PostgresEngine();
  await lintEngine.connect({ database_url: DATABASE_URL! });
  // The act: lint-engine disconnects on phase completion.
  await lintEngine.disconnect();
  // CRITICAL: module-level singleton is STILL alive.
  // Pre-fix this throws: "No database connection: connect() has not been called."
  const after = await db.getConnection().unsafe('SELECT 1 as ok');
  expect((after[0] as unknown as { ok: number }).ok).toBe(1);
});
\`\`\`

Pre-fix: throws \`No database connection: connect() has not been called\`. Post-fix: passes.

## Verification

Applied the fix on a production-shape brain (multi-source, ZE as primary embedding column, billy-paths schema pack active) and ran the per-source dream cycle that previously failed:

\`\`\`
gbrain dream --source gstack-artifacts-billyarmstrong
\`\`\`

Pre-fix output included \`connection blip, retrying\` messages with batch row loss and \`conversation_facts_backfill\` aborting. Post-fix:

- Zero \`connection blip\` messages
- Zero \`connect() has not been called\` errors
- \`conversation_facts_backfill\` phase completes cleanly
- Full cycle runs end-to-end (~92s for this source)

## Files changed

- \`src/core/postgres-engine.ts\` — single behavioral change (1 line removed) + 20 lines of explanatory comment for future maintainers
- \`test/e2e/postgres-engine-disconnect-idempotency.test.ts\` — new test case (39 lines added)

## Risk

Minimal. The change is purely subtractive on the hot path (removes a call that was breaking the singleton). It cannot newly break anything that previously worked. The behavioral contract change is documented in the comment block:

- Singleton-owning engines (worker pools with their own \`_sql\`) are unchanged (they hit the early-return branch above)
- Instance-level disconnects on module-style engines used to leak a teardown side-effect; that side-effect was always destructive
- Process-exit cleanup is unchanged (\`cli.ts\` calls \`db.disconnect()\` explicitly)

### DIFF
--- a/src/core/db.ts
+++ b/src/core/db.ts
@@ -8,6 +8,28 @@ let sql: ReturnType<typeof postgres> | null = null;
 let connectedUrl: string | null = null;
 
 /**
+ * Refcount for the module-level singleton.
+ *
+ * Multiple PostgresEngine instances can piggyback on the same module
+ * singleton (the CLI's top-level engine plus any ad-hoc engines created
+ * mid-cycle — e.g. `resolveLintContentSanity` at lint.ts:319 spins up
+ * its own engine to probe DB config). Each `connect()` call that reuses
+ * the existing pool bumps this counter; each `disconnect()` call
+ * decrements it. The pool is only closed (via `sql.end()`) when the
+ * counter reaches zero.
+ *
+ * This closes the bug class the v1 #1546 patch chose between:
+ *   - v1 patch said "instance disconnect never closes the singleton"
+ *     → fixed the lint case but orphaned the CLI top-level disconnect
+ *     path; `gbrain init` and every op-dispatch command hung past their
+ *     natural exit until the 15s test SIGTERM or 10s force-exit timer.
+ *   - Pre-v1 behavior said "every instance disconnect closes the
+ *     singleton" → broke the lint case mid-cycle.
+ * Refcount is the discriminator both sides needed.
+ */
+let refCount = 0;
+
+/**
  * Default pool size for Postgres connections. Users on the Supabase transaction
  * pooler (port 6543) or any multi-tenant pooler can lower this to avoid
  * MaxClients errors when `gbrain upgrade` spawns subprocesses that each open
@@ -165,6 +187,10 @@ export async function connect(config: EngineConfig): Promise<void> {
     if (config.database_url && connectedUrl && config.database_url !== connectedUrl) {
       console.warn('[gbrain] connect() called with a different database_url but a connection already exists. Using existing connection.');
     }
+    // Piggyback: add a reference to the existing singleton. The matching
+    // disconnect() decrement keeps the singleton alive until every
+    // caller releases it.
+    refCount++;
     return;
   }
 
@@ -210,11 +236,14 @@ export async function connect(config: EngineConfig): Promise<void> {
     // Test connection
     await sql`SELECT 1`;
     connectedUrl = url;
+    // First successful connect — initial reference held by the caller.
+    refCount = 1;
 
     await setSessionDefaults(sql);
   } catch (e: unknown) {
     sql = null;
     connectedUrl = null;
+    refCount = 0;
     const msg = e instanceof Error ? e.message : String(e);
     throw new GBrainError(
       'Cannot connect to database',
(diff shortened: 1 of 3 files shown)


Mr-B-1:
Confirming the fix works in production on a separate setup. Applied this patch to gbrain v0.41.14.0 on Ubuntu 24.04 / Bun runtime / PostgreSQL 16 / Ollama nomic-embed-text (no Anthropic-paid embeddings).

**Pre-patch symptoms:** `gbrain dream` via systemd timer (nightly cron) consistently died with exit 1 after ~4 seconds, ~7 MB memory peak. Reproduced 3x (May 26 02:00 PST, May 26 21:40 PST, May 27 02:00 PST). Left stale `gbrain-cycle` rows in `gbrain_cycle_locks` every time (compounding with #1470 since the lock release never fires on hard exit).

**Post-patch behavior:** `gbrain dream` now runs cleanly through all cycle phases. 11+ phases completed in a 10-min interactive test (lint, backlinks, sync, synthesize, extract, extract_facts, resolve_symbol_edges, patterns, recompute_emotional_weight, consolidate, propose_takes). Memory peak ~440-477 MB (normal range). Currently running a full systemd-triggered cycle that has been alive for 4+ min with PID stable, lock properly held with valid TTL.

This patch fixed every observable symptom on my setup. Filed the 4-sec exit signature as #1535 before discovering this PR — that issue is the user-visible symptom of #1404. Will close it as duplicate.

LGTM from a downstream-user perspective. Looking forward to the merge.

jkhoffman:
Tested #1546 against my Postgres repro environment.

Context:
- Checked out PR head `3ef4e573` in a clean worktree.
- Backend: full Postgres on NAS via the existing shadow `GBRAIN_HOME` config (`/home/hermes/gbrain-pg-shadow`).
- I did not print or expose the DB URL/password.

Verification:

```sh
bun test test/e2e/postgres-engine-disconnect-idempotency.test.ts
```

Result:

```text
5 pass, 0 fail
```

Live repro:

```sh
# deleted stale cycle locks first
GBRAIN_HOME=/home/hermes/gbrain-pg-shadow bun src/cli.ts dream --source default --json
```

I ran this twice; the second run sourced my normal API-key env so `embed` could exercise the Postgres path too.

Observed result on the second run:

```text
cycle_status=partial
sync=ok
extract=ok
conversation_facts_backfill=skipped (cycle.conversation_facts_backfill.enabled=false in this shadow config)
embed=ok (11 chunks newly embedded)
contains_no_db_connection=False
contains_connection_ended=False
locks_after=0
```

The remaining `partial` status was from ordinary brain-quality warnings in this shadow brain (`lint` and `orphans`), not from the Postgres connection lifecycle bug.

So #1546 fixes the NAS Postgres singleton-clobber repro for me: no `No database connection`, no `CONNECTION_ENDED`, and no stranded `gbrain_cycle_locks` row after the run.

billy-armstrong:
Rebased onto v0.41.38.0 (was based on v0.41.20.0) and squashed the v1+v2 commits into one. The previously-red `test (2)` shard was an unrelated stale-base hermeticity issue — the `facts-engine` cosine-ordering test's embedding-dimension assertion, which is fixed at current master — not anything in this change. CI is now green (20/20) and mergeable is clean. The refcount fix and the disconnect-idempotency suite (5/5) are unchanged, so the downstream production confirmations above still apply.

garrytan:

PR #1547, opened 2026-05-27
feat(link-extraction): consult active schema pack for DIR_PATTERN

billy-armstrong:
## What this fixes

The Schema Cathedral v3 (v0.40.7+) made schema packs the canonical source of truth for "what path prefixes are entity-bearing in this brain." Every other gbrain code path honors that — `parseMarkdown` infers page type from `page_types[].path_prefixes`, `whoknows` / `find_experts` scopes to declared expert types, `extract_facts` gates on declared extractable types, `enrichment-service` routes per primitive.

The link extractor never got wired up. Its `DIR_PATTERN` regex at `src/core/link-extraction.ts:46` is a frozen 16-prefix hardcoded list:

```typescript
const DIR_PATTERN = '(?:people|companies|meetings|concepts|deal|civic|project|projects|source|media|yc|tech|finance|personal|openclaw|entities)';
```

This silently ignores prefixes that gbrain's OWN bundled schema packs declare. Examples from `gbrain-base.yaml`:

| Declared in gbrain-base.yaml | In DIR_PATTERN? |
|---|---|
| `person/` (singular) | ❌ |
| `company/` (singular) | ❌ |
| `writing/` | ❌ |
| `wiki/analysis/`, `wiki/guides/`, `wiki/hardware/`, etc | ❌ |

So a fresh gbrain install with the default `gbrain-base` pack active loses every `[[wiki/analysis/...]]`, `[[wiki/guides/...]]`, `[[writing/...]]`, `[[person/...]]`, `[[company/...]]` wikilink. User-installed packs hit the same gap.

## The fix

Replace the frozen `DIR_PATTERN` const with a lazy resolver that reads the active schema pack via `tryCachedPack()` (sync, documented ~10ns hot path when warm) and unions its declared `path_prefixes` with the canonical fallback list.

The 3 module-level const regexes (`ENTITY_REF_RE`, `WIKILINK_RE`, `QUALIFIED_WIKILINK_RE`) become builder functions that construct fresh regexes per call from the resolved pattern. Existing call sites at `extractEntityRefs` already cloned via `.source`/`.flags` to reset `lastIndex`, so the per-call construction cost is unchanged.

The pattern follows the same `tryCachedPack` + cold-cache fallback shape established by other Schema Cathedral consumers (`whoknows`, `find_experts`, `extract_facts`).

## Behavior

| Condition | Effective DIR_PATTERN |
|---|---|
| No active pack OR registry cache cold | Canonical fallback only (byte-identical to upstream) |
| Active pack with declared `path_prefixes` | Union of canonical + top-level segments of declared prefixes |
| Pack hot-swap | Cache invalidates by pack identity → next call rebuilds |

Cold-cache fallback to canonical preserves byte-identical behavior for fresh installs and pre-warm boundary. Once another code path warms the schema-pack cache via `loadActivePack` (typically the engine's first schema-pack-aware operation), subsequent link-extraction calls pick up the warm cache. Cache key is pack identity (sha8-stable); mismatch on swap triggers a fresh build.

## Limitation (T20)

Per the v0.38 `ResolvedPack` contract, this PR reads ONLY the leaf pack's `manifest.page_types`. Extends-chain merging (so a pack declaring `extends: gbrain-base` would also see base's `wiki/`, `writing/`, etc) is the T20 follow-up. For most users this is fine — they declare what their vault uses at the leaf. The edge case is a custom pack that extends gbrain-base but doesn't redeclare base's prefixes; that case still falls back to the canonical list.

## Verification

Smoke-tested against a production-shape brain with an active user pack (19 declared types). Pre-patch `DIR_PATTERN` had 16 prefixes; post-patch resolved to 29 prefixes (canonical + 13 from the active pack). All 98 cases in `test/link-extraction.test.ts` pass; full `tsc --noEmit` clean.

## Files

- `src/core/link-extraction.ts` — adds 4 imports, replaces 3 const regexes with builder functions, adds `getDirPattern()` resolver with identity-keyed cache (+76 / -24 lines)

## Risk

Subtractive on the hot path — falls back to byte-identical upstream behavior when no pack is active or cache is cold. Cannot newly break anything that previously worked. Strictly adds more wikilink resolution in the warm-cache case (when the user's pack declares more prefixes than canonical alone).

### DIFF
--- a/src/core/link-extraction.ts
+++ b/src/core/link-extraction.ts
@@ -13,6 +13,10 @@
 
 import type { BrainEngine } from './engine.ts';
 import type { PageType } from './types.ts';
+// PATCH:link-extractor-schema-pack-aware v1
+import { resolveActivePackNameOnly } from './schema-pack/load-active.ts';
+import { tryCachedPack, type ResolvedPack } from './schema-pack/registry.ts';
+import { loadConfig } from './config.ts';
 
 // ─── Entity references ──────────────────────────────────────────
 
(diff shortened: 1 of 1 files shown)


billy-armstrong:
Closing for queue hygiene — no review activity since opening (2026-05-27). Note the underlying gap remains in v0.42.40.0: link-extraction's DIR_PATTERN is hardcoded (src/core/link-extraction.ts:86) and doesn't include e.g. wiki/, so wikilinks into wiki/* paths aren't extracted on brains using the stock gbrain-base-v2 layout. Happy to refresh against current master if there's interest.

PR #1553, opened 2026-05-27
fix(doctor): emit accurate --break-lock hint for non-sync stale locks

### LINKED ISSUE #1534: doctor stale_locks: --break-lock hint hard-codes <code>gbrain sync</code> even when lock is <code>gbrain-cycle</code>
## Summary

When `doctor` reports a stale `gbrain-cycle` lock, it advises `gbrain sync --break-lock` — but that command can only address `gbrain-sync:*` locks. There is no shipped CLI handler for `gbrain-cycle` locks, so the only way to clear them is direct SQL into `gbrain_cycle_locks`.

## Repro

1. Have a dream cycle die mid-run (process crash, OOM, sigkill) so the `gbrain-cycle` row stays in `gbrain_cycle_locks` past its TTL
2. Run `gbrain doctor`. The `stale_locks` check correctly detects it and prints:
   `gbrain-cycle (pid <N> on <host>, age <age>) → gbrain sync --break-lock`
3. Run the suggested `gbrain sync --break-lock`
4. Output: `Lock gbrain-sync:default is not held (nothing to break).`
5. The cycle lock is unchanged, and subsequent `cycle_freshness` checks continue to FAIL

## Root cause

`src/commands/doctor.ts:1670` hard-codes the hint:
```ts
const breakHint = source ? `gbrain sync --break-lock --source ${source}` : `gbrain sync --break-lock`;
```

`src/commands/sync.ts:556` scopes `--break-lock` to sync locks only:
```ts
`gbrain sync --break-lock --source ${lockKey.slice('gbrain-sync:'.length)}`
```

There is no analogous handler for `gbrain-cycle` (or any other lock prefix).

## Tested workarounds that did not work

- `gbrain sync --break-lock` — wrong scope (handles `gbrain-sync:*` only)
- `gbrain sync --break-lock --source default` — same
- `gbrain sync --break-lock --source all` — same
- `gbrain dream --break-lock` — flag silently dropped; just runs a cycle (same shape as #1452)

## Suggested fixes

1. **Doctor:** branch the hint based on the lock_name prefix (e.g. `gbrain-cycle` → suggest the cycle breaker)
2. **CLI:** ship a handler. Either:
   - `gbrain dream --break-lock` (route by command that owns the lock kind), or
   - `gbrain admin break-lock <lock-name>` (kind-agnostic)
3. Both should reuse the same liveness + force-break semantics as `sync.ts`s D3 implementation (PID check, TTL window, cross-host safety, `--force-break-lock` opt-out)

## Environment

- gbrain v0.41.14.0
- Hit on `gbrain-cycle` after a dream cycle died at 09:00 UTC; lock TTL expired at 09:30 UTC; PID 3396714 was dead per `ps -p` at 21:30 UTC
- Stale lock blocked all subsequent dream/cycle runs (cycle_freshness FAIL: "last cycled 30h ago")

## Related

- #1452 (dream silently drops --source) — same pattern of flags being dropped
- #1454 (doctor advertises config option that does not exist) — same pattern of doctor hints not matching shipped CLI surface

Sanjays2402:
## Summary

Fixes #1534.

`doctor`'s `stale_locks` check hard-coded `gbrain sync --break-lock` as the recovery hint for every stale row. But that command, per `sync.ts`, only handles `gbrain-sync:*` keys — it silently no-ops on `gbrain-cycle` (and per-source `gbrain-cycle:<source>`) locks with `Lock gbrain-sync:default is not held (nothing to break).`, leaving operators with no shipped path to clear the lock.

## Fix

Extracted a small `formatStaleLockBreakHint(lockId)` helper that branches on the lock-id prefix:

- `gbrain-sync:<src>` → `gbrain sync --break-lock --source <src>`
- `gbrain-sync` → `gbrain sync --break-lock`
- anything else (notably `gbrain-cycle`, `gbrain-cycle:<src>`, future kinds) → an honest SQL-fallback hint:  
  `manual clear required (no CLI breaker for <id>): DELETE FROM gbrain_cycle_locks WHERE id = '<id>';`

The `id` is single-quote-escaped so the printed snippet is safe to copy verbatim.

## Why this shape instead of shipping a new CLI breaker

The issue suggests also adding `gbrain dream --break-lock` or `gbrain admin break-lock <name>`. That's the right next step but it's a CLI-surface decision (PID liveness, force-break semantics, cross-host guards) that's better landed on its own with proper tests. This PR is the minimum honest-hint change so that `doctor` stops actively misleading operators today.

## Risk

Pure formatter change inside the doctor output builder. No DB writes, no new CLI surface, no behavior change for the existing `gbrain-sync:*` cases. Output text changes for cycle locks only.

## Test plan

No unit tests target this string today. Verified the helper output by extracting it into a Node REPL (covers sync, sync-with-source, cycle, cycle-with-source, and single-quote-in-id):

```
gbrain sync --break-lock --source default
gbrain sync --break-lock
manual clear required (no CLI breaker for gbrain-cycle): DELETE FROM gbrain_cycle_locks WHERE id = 'gbrain-cycle';
manual clear required (no CLI breaker for gbrain-cycle:notes): DELETE FROM gbrain_cycle_locks WHERE id = 'gbrain-cycle:notes';
manual clear required (no CLI breaker for weird's-lock): DELETE FROM gbrain_cycle_locks WHERE id = 'weird''s-lock';
```

Did not run the full `bun test` / `bun run verify` suite locally (no Postgres + Bun on this machine); CI will cover.

### DIFF
--- a/src/commands/doctor.ts
+++ b/src/commands/doctor.ts
@@ -1920,6 +1920,29 @@ export function checkAutopilotLockScope(): Check {
 }
 
 /**
+ * Format a precise --break-lock recovery hint for a stale lock id.
+ *
+ * v0.41.6.0 D3 ships `gbrain sync --break-lock` which handles `gbrain-sync:*`
+ * keys only. Other lock kinds (notably `gbrain-cycle` and per-source
+ * `gbrain-cycle:<source>`) do not have a corresponding shipped CLI handler,
+ * so blindly suggesting `gbrain sync --break-lock` for them is wrong and
+ * misleads operators down a dead-end (issue #1534). Until a kind-agnostic
+ * breaker ships, point at the SQL fallback for unsupported kinds so the
+ * hint stays honest.
+ */
+function formatStaleLockBreakHint(lockId: string): string {
+  if (lockId.startsWith('gbrain-sync:')) {
+    return `gbrain sync --break-lock --source ${lockId.slice('gbrain-sync:'.length)}`;
+  }
+  if (lockId === 'gbrain-sync') {
+    return `gbrain sync --break-lock`;
+  }
+  // Cycle / unrecognized kinds: no shipped CLI breaker. Be explicit instead of
+  // suggesting a command that silently no-ops.
+  return `manual clear required (no CLI breaker for ${lockId}): DELETE FROM gbrain_cycle_locks WHERE id = '${lockId.replace(/'/g, "''")}';`;
+}
+
+/**
  * v0.41.6.0 D3 — stale_locks doctor check.
  *
  * Surfaces every row in `gbrain_cycle_locks` whose `ttl_expires_at < NOW()`.
@@ -1947,8 +1970,7 @@ export async function checkStaleLocks(engine: BrainEngine): Promise<Check> {
     }
     const lines = stale.slice(0, 10).map(s => {
       const ageH = Math.floor(s.age_ms / 3600_000);
-      const source = s.id.startsWith('gbrain-sync:') ? s.id.slice('gbrain-sync:'.length) : null;
-      const breakHint = source ? `gbrain sync --break-lock --source ${source}` : `gbrain sync --break-lock`;
+      const breakHint = formatStaleLockBreakHint(s.id);
       return `  ${s.id} (pid ${s.holder_pid} on ${s.holder_host}, age ${ageH}h) → ${breakHint}`;
     });
     const tail = stale.length > 10 ? `  ... and ${stale.length - 10} more.` : null;


iannwu:
Confirmed this same failure mode on a real local Postgres brain after upgrading to `gbrain 0.41.38.0` (and checked that current `origin/master` still has the pre-PR hint shape).

What I saw:

```text
stale_locks warn - 1 stale lock(s) detected (ttl_expires_at < NOW()):
  gbrain-cycle (pid 45905 on <host>, age 24h) → gbrain sync --break-lock
```

Following the hint did not clear the lock because `sync --break-lock` targets `gbrain-sync:<source>` locks:

```text
$ gbrain sync --break-lock --max-age 3600
Lock gbrain-sync:default is not held (nothing to break).
```

I cleared the expired `gbrain-cycle` row manually with direct SQL after verifying it was stale. This PR's "honest SQL-fallback hint for non-sync locks" would have made the operator path clear and prevented the confusing no-op.

Also agree with the PR body that #1591 is the right follow-up for a real cycle-lock breaker CLI; this PR is still valuable as the minimum correction because current doctor output actively points at the wrong command.

time-attack:
Closing — merging this would now be a regression.

When you filed this in May, doctor's `stale_locks` hint wrongly suggested `gbrain sync --break-lock` for gbrain-cycle locks, and your raw-SQL fallback was a reasonable fix. Since then, master commit `7c27fa12` (v0.42.41.0, #2128) rewrote those exact lines with a broader fix: prefix-aware hints (`gbrain dream --break-lock [--source X]` for cycle locks, with a doctor fallback) plus `doctor --fix` auto-reaping dead-holder locks. Issue #1534 is closed.

The branch is CONFLICTING/DIRTY against that rewrite, and resolving the conflict in this PR's favour would **replace master's targeted command hints with a raw SQL DELETE hint** — strictly worse for the user hitting it.

Thanks for the report and the fix; it was right when written, and the eventual fix went further in the same direction.

One thing the review turned up on master while checking this, which I am filing separately rather than holding your PR for: doctor's replacement hint names a flag that needs verifying against the actual command surface. That belongs to the class in #3697 — remediation text naming commands or flags that do not behave as the message claims.

Issue #526, opened 2026-04-30
bun install fails: postinstall script unparseable by Bun's built-in shell


lllcccwww:
## Summary

  Following the README's recommended install path (`git clone && bun install && bun link`), `bun install` fails inside the cloned repo because Bun's built-in mini-shell cannot
  parse the `postinstall` script's redirect syntax. The script never gets a chance to execute its intended logic, and `bun install` exits with code 1.

  This is a *different* Bun pitfall from the one already documented in the README (which warns about global installs blocking the postinstall hook). Here, the hook *does* fire —
  it just fails at parse time.

  ## Environment

  - OS: Windows 11
  - Bun: 1.3.12 (700fc117)
  - Install path: `git clone` + `bun install` (the README-recommended path)

  ## Reproduction

  ```sh
  git clone https://github.com/garrytan/gbrain.git
  cd gbrain
  bun install
  ```

  ## Actual output

  ```
  $ command -v gbrain >/dev/null 2>&1 && gbrain apply-migrations --yes --non-interactive || echo '[gbrain] postinstall skipped. ...' 1>&2
  error: Failed to run script D:\...\gbrain\[eval] due to error expected a command or assignment but got: "Redirect"
  error: postinstall script from "gbrain" exited with 1
  ```

  ## Root cause

  Bun does not invoke the system shell for `package.json` scripts; it uses its own POSIX-ish mini-shell written in Zig. That parser does not currently accept the combination used
  in the postinstall:

  ```sh
  command -v gbrain >/dev/null 2>&1 && gbrain apply-migrations --yes --non-interactive \
    || echo '[gbrain] postinstall skipped...' 1>&2
  ```

  Specifically, the chained redirects (`>/dev/null 2>&1` ... `1>&2`) combined with `&&` / `||` trip the parser. The exact same line runs fine under bash / zsh / Git Bash.

  ## Suggested fix

  Move the postinstall logic into a small Node script and invoke it without shell features:

  ```json
  "scripts": {
    "postinstall": "node scripts/postinstall.mjs"
  }
  ```

  Inside `scripts/postinstall.mjs`, use `child_process.spawnSync` to:

  1. Check whether `gbrain` is on PATH.
  2. If yes, run `gbrain apply-migrations --yes --non-interactive`.
  3. If no, print the existing skip message to stderr.

  This avoids the entire shell-compatibility surface and works identically under bun / npm / pnpm / yarn on every platform.

  ## Workaround for users hitting this today

  ```sh
[...]

Issue #218, opened 2026-04-19
bun install -g github:... produces broken CLI (Aborted in PGLite callMain) + blocked postinstall hides migrations


gopalpatel:
## Summary

`bun install -g github:garrytan/gbrain` produces a CLI where every command that opens PGLite aborts with a WASM `Aborted()`, even on a brand-new empty data dir. The same version installed via the documented `git clone + bun install && bun link` path (per `INSTALL_FOR_AGENTS.md`) works correctly.

Bun's security policy also silently blocks the postinstall hook (`Blocked 1 postinstall. Run 'bun pm -g untrusted' for details.`) so users following the "just upgrade" flow in CHANGELOG don't learn that `apply-migrations` never ran.

Hit this upgrading from v0.12.0 (installed yesterday via `bun install -g`, worked fine) → v0.12.3 (same install command, broken).

## Environment

- **OS:** NVIDIA DGX Spark, Linux ARM64 (Ubuntu)
- **Node:** v22.22.1
- **Bun:** 1.3.12
- **gbrain:** 0.12.3 (commit `013b348`) — also reproduces on 0.12.0 (`81b3f7a`) when reinstalled via the same path
- **@electric-sql/pglite:** 0.4.4 (declared and resolved)

## Repro (broken path)

```bash
bun install -g github:garrytan/gbrain
# "Blocked 1 postinstall. Run `bun pm -g untrusted` for details."
export PATH=$HOME/.bun/bin:$PATH
gbrain --version                # prints: gbrain 0.12.3  (looks fine)
mkdir /tmp/test-brain && cd /tmp/test-brain
gbrain init --pglite --path /tmp/test-brain/brain
# Setting up local brain with PGLite (no server needed)...
# Aborted(). Build with -sASSERTIONS for more info.
```

`gbrain config`, `gbrain stats`, `gbrain apply-migrations --yes` etc. all produce the same `Aborted()` once the CLI tries to open PGLite.

## Works (blessed path)

```bash
bun remove -g gbrain
git clone https://github.com/garrytan/gbrain.git ~/gbrain-tmp
cd ~/gbrain-tmp && bun install && bun link
gbrain --version                # prints: gbrain 0.12.3
mkdir /tmp/test-brain-2 && cd /tmp/test-brain-2
gbrain init --pglite --path /tmp/test-brain-2/brain
# Setting up local brain with PGLite (no server needed)...
#   Migration 2 applied: slugify_existing_pages
#   ...
[...]

PR #1554, opened 2026-05-27
fix(postinstall): cross-platform node shim instead of POSIX shell

### LINKED ISSUE #1486: Windows: bun install fails due to Unix shell redirect syntax in postinstall script
**What happened?**

在 Windows 上执行 `bun install` 时，postinstall 脚本因使用了 Unix shell 重定向语法（`1>&2`、`>/dev/null`、`command -v`）导致 Bun 的内置脚本解析器解析失败，安装报错退出。

**What did you expect?**

`bun install` 应正常完成安装，postinstall 脚本在检测到环境不满足条件时应优雅跳过，而非报错退出。

**Steps to reproduce**
1. 在 Windows 11 环境下 `git clone` gbrain 仓库
2. 执行 `bun install`
3. 观察终端输出：postinstall 阶段报错，`exit code 1`

**Environment**
- gbrain version: 0.41.14.0
- OS: Windows 11 Home China 10.0.26200
- Bun version: 1.3.14
- Database: N/A（`bun install` 阶段，尚未配置数据库）

```
N/A — 安装未完成，gbrain 二进制尚未可用
```

**Additional context**

失败行位于 `package.json` 的 postinstall 脚本：

```json
"postinstall": "command -v gbrain >/dev/null 2>&1 && gbrain apply-migrations --yes --non-interactive || echo '...' 1>&2"
```

三个不兼容 Windows 的语法点：

- `command -v` — POSIX shell 命令，Bun 在 Windows 上的内置解析器不支持
- `>/dev/null 2>&1` — `/dev/null` 在 Windows 上不存在，`2>&1` 被 Bun 解析器拒绝
- `1>&2` — 触发错误 `expected a command or assignment but got: "Redirect"`

临时修复：将 postinstall 替换为 `node -e "try{...}catch(e){...}"` 跨平台写法后，`bun install` 可正常完成。

Sanjays2402:
## Summary

Fixes #1486.

The `postinstall` script used POSIX shell syntax (`command -v`, `>/dev/null 2>&1`, `1>&2`) which Bun's built-in script parser rejects on Windows. `bun install` aborts on Windows 11 with:

```
expected a command or assignment but got: "Redirect"
exit code 1
```

…before `gbrain` is even probed.

## Fix

Replace the POSIX one-liner with a `node -e` shim. `node` is a hard prerequisite of the install (bun is built on top of it on every platform, and `node` is documented in the existing setup) so we can rely on it here.

```js
const {spawnSync} = require('child_process');
const probe = spawnSync('gbrain', ['--version'], {stdio:'ignore', shell: process.platform === 'win32'});
if (probe.status === 0) {
  const r = spawnSync('gbrain', ['apply-migrations','--yes','--non-interactive'], {stdio:'inherit', shell: process.platform === 'win32'});
  process.exit(r.status || 0);
} else {
  process.stderr.write('[gbrain] postinstall skipped... See #218\\n');
}
```

- `shell: true` on Windows lets the spawn resolve `gbrain.cmd` shims (PATHEXT) correctly.
- Skip branch exits 0, so a fresh clone (where `gbrain` is not yet on `PATH`) still completes `bun install` exactly as today.
- POSIX hosts get identical end behavior: probe → migrate or skip.

## Risk

Lowest-risk surface — only the install probe changes. Behavior on macOS / Linux is preserved; Windows goes from hard-fail to working install.

## Test plan

```
$ node -e '<shim>'
[gbrain] postinstall skipped... See #218
$ echo $?
0
```

`package.json` parses (`JSON.parse(fs.readFileSync(...))` ok). Cannot exercise the success branch without a built `gbrain` binary on PATH; that branch is the same `gbrain apply-migrations --yes --non-interactive` call as before, just spawned via Node instead of an inline shell pipe.

I do not have a Windows host to verify end-to-end; if you have CI for that, please rerun. Happy to iterate if there's a preferred shim style (cross-spawn, a real script under `scripts/`, etc.).

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
     "check:fixture-privacy": "scripts/check-fixture-privacy.sh",
     "check:conversation-parser": "bun src/cli.ts eval conversation-parser test/fixtures/conversation-formats/all.jsonl --no-llm",
     "check:source-scope-onboard": "scripts/check-source-scope-onboard.sh",
-    "postinstall": "node -e \"const{spawnSync}=require('child_process');const probe=spawnSync('gbrain',['--version'],{stdio:'ignore',shell:process.platform==='win32'});if(probe.status===0){const r=spawnSync('gbrain',['apply-migrations','--yes','--non-interactive'],{stdio:'inherit',shell:process.platform==='win32'});process.exit(r.status||0)}else{process.stderr.write('[gbrain] postinstall skipped. If installed via bun install -g github:...: run `gbrain doctor` and `gbrain apply-migrations --yes` manually. See https://github.com/garrytan/gbrain/issues/218\\n')}\"",
+    "postinstall": "bun run scripts/postinstall.ts",
     "prepublish:clawhub": "bun run build:all",
     "publish:clawhub": "clawhub package publish . --family bundle-plugin"
   },
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
     "check:fixture-privacy": "scripts/check-fixture-privacy.sh",
     "check:conversation-parser": "bun src/cli.ts eval conversation-parser test/fixtures/conversation-formats/all.jsonl --no-llm",
     "check:source-scope-onboard": "scripts/check-source-scope-onboard.sh",
-    "postinstall": "command -v gbrain >/dev/null 2>&1 && gbrain apply-migrations --yes --non-interactive || echo '[gbrain] postinstall skipped. If installed via bun install -g github:...: run `gbrain doctor` and `gbrain apply-migrations --yes` manually. See https://github.com/garrytan/gbrain/issues/218' 1>&2",
+    "postinstall": "node -e \"const{spawnSync}=require('child_process');const probe=spawnSync('gbrain',['--version'],{stdio:'ignore',shell:process.platform==='win32'});if(probe.status===0){const r=spawnSync('gbrain',['apply-migrations','--yes','--non-interactive'],{stdio:'inherit',shell:process.platform==='win32'});process.exit(r.status||0)}else{process.stderr.write('[gbrain] postinstall skipped. If installed via bun install -g github:...: run `gbrain doctor` and `gbrain apply-migrations --yes` manually. See https://github.com/garrytan/gbrain/issues/218\\n')}\"",
     "prepublish:clawhub": "bun run build:all",
     "publish:clawhub": "clawhub package publish . --family bundle-plugin"
   },
--- a/scripts/postinstall.ts
+++ b/scripts/postinstall.ts
@@ -1,0 +1,49 @@
+#!/usr/bin/env bun
+// scripts/postinstall.ts
+//
+// Postinstall hook: after `bun install`, apply any pending schema migrations so
+// a freshly-installed gbrain is immediately usable. Wired via package.json
+// ("postinstall": "bun run scripts/postinstall.ts") as a real Bun script rather
+// than an inline `node -e` one-liner.
+//
+// Why a script file and not an inline command:
+//   Embedding a program inside the package.json postinstall string lets the
+//   lifecycle shell mangle it. Bun's Windows script-runner expands `\n` in the
+//   hint string into a REAL newline before node sees it, producing
+//   `SyntaxError: Invalid or unexpected token` and aborting the whole install.
+//   `node` is also not guaranteed present under a Bun install (bun is the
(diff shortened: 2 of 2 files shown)


joshwilks111-max:
@
Thanks for picking this up — I hit the same crash and can confirm the root-cause analysis is spot on. But I tested this PR's shim on Windows and **it still fails**, just with a different error than the original. Sharing a full repro so we can get a version that actually lands.

**Environment:** Windows 11 (10.0.26200), Bun 1.3.13, Node v24.14.0, gbrain on `master` (postinstall still uses the POSIX one-liner — unfixed as of today).

### The `node -e` shim doesn't parse under Bun's Windows script-runner

Running this PR's exact `postinstall` via `bun install` / `bun run postinstall` on Windows:

```
$ node -e "const{spawnSync}=require('child_process');...process.stderr.write('[gbrain] postinstall skipped. See https://github.com/garrytan/gbrain/issues/218
')}"
[eval]:1
...issues/218
                  ^^^^
SyntaxError: Invalid or unexpected token
    at makeContextifyScript (node:internal/vm:194:14)
```

The `\n` at the end of the hint string is the problem. Bun's built-in script-runner (which runs lifecycle scripts on Windows — there's no `/bin/sh`) **expands `\n` into a real newline before `node` ever sees the string**. That newline lands inside the single-quoted JS string literal, so `node -e` throws `SyntaxError: Invalid or unexpected token`. Minimal repro, no gbrain involved:

```
# package.json -> "postinstall": "node -e \"process.stderr.write('hi\n')\""
$ bun run postinstall
[eval]:1
process.stderr.write('hi
SyntaxError: Invalid or unexpected token
```

So this is the **same bug class** as the original report — *the lifecycle shell mangles the script string before the interpreter runs it* — just relocated from POSIX redirect grammar to a JS string-literal escape. It likely passed locally because the shell you tested on preserves `\n` as the two characters `\` + `n` (which `node` then reads as a newline escape); Bun's Windows runner expands it instead. It's shell-dependent.

### Two smaller notes on the approach

1. **`node` isn't guaranteed under a Bun install.** The PR body says "bun is built on top of node on every platform" — Bun is actually a from-scratch runtime (Zig + JavaScriptCore) and doesn't require Node. A Bun-only user may not have `node` on PATH. Since this is a `bun install` lifecycle hook, **`bun` is the guaranteed-present runtime, not `node`.** (In practice many machines have both — mine does — so this is a portability nit, not the blocker. The newline above is the blocker.)
2. **`shell: process.platform === 'win32'` re-introduces a shell on Windows**, so `gbrain apply-migrations` gets launched through `cmd.exe` — re-opening a (smaller) quoting surface. Resolving the binary and spawning with an argv array and no shell avoids that.

### Suggestion: move the logic into a file

Both the original and this shim fail because they **embed a program inside a package.json string**, where the lifecycle shell can mangle it. The robust fix is to not put the program in the string at all — keep package.json to `bun run scripts/postinstall.ts` (the pattern the repo already uses for ~19 other scripts) and put the logic in a real file:

```ts
// scripts/postinstall.ts
import { which } from 'bun';
const HINT =
  '[gbrain] postinstall skipped. If installed via `bun install -g github:...`: ' +
  'run `gbrain doctor` and `gbrain apply-migrations --yes` manually. ' +
  'See https://github.com/garrytan/gbrain/issues/218';
const bin = which('gbrain');                 // Windows-aware PATH resolve (finds gbrain.exe)
if (!bin) { console.error(HINT); process.exit(0); }
try {
  const r = Bun.spawnSync([bin, 'apply-migrations', '--yes', '--non-interactive'],
    { stdout: 'inherit', stderr: 'inherit' });  // argv array, no shell
  if (r.exitCode !== 0) console.error(HINT);
} catch { console.error(HINT); }
process.exit(0);                              // never abort the install
```

```diff
-    "postinstall": "command -v gbrain >/dev/null 2>&1 && ... 1>&2",
+    "postinstall": "bun run scripts/postinstall.ts",
```

`bun` is guaranteed present in a Bun lifecycle script, `Bun.which` resolves `gbrain.exe` (the old `command -v` couldn't), the argv array means no shell grammar to mangle, and it keeps the exact #218 hint + the never-fail-the-install contract. **Verified on the environment above:** `bun install` completes, migrations run ("All migrations up to date."), exit 0; and with `gbrain` absent from PATH it prints the hint and still exits 0.

Happy to push this as a commit to your branch or open it as an alternative — whichever you and @garrytan prefer. Also worth linking #526 (an earlier duplicate of the same bug, currently unconnected to any PR).
@

Issue #1531, opened 2026-05-27
extract_atoms silently fails with "No database connection" on Postgres  — loadConfigWithEngine() leaves pool in bad stat


foxhoundinc:
Version: v0.41.18.0
Engine: Postgres 
Schema version: v103

SYMPTOM
gbrain dream --json runs 15/17 phases clean. extract_atoms phase fails with:

  [extract_atoms] page-discovery query failed: No database connection: 
  connect() has not been called.

Phases before and after extract_atoms (including sync, extract_facts, 
resolve_symbol_edges, consolidate, propose_takes) all use engine.executeRaw() 
successfully. Only extract_atoms fails.

ROOT CAUSE (diagnosed via source inspection)
extract-atoms.ts calls loadConfigWithEngine(engine) before discoverExtractablePages(). 
loadConfigWithEngine() calls engine.getConfig() using tagged-template SQL 
(sql`SELECT value FROM config WHERE key = ${key}`). This appears to leave the 
module-level connection singleton (db.ts) in a transient bad state where 
db.getConnection() returns "No database connection: connect() has not been called."

The catch block at line ~193 swallows the error silently, returning []. 
Phase completes with 0 atoms. Later phases recover via fresh pool connections.

IMPACT
- 0 atoms extracted despite 3,125 pages in corpus
- Error is silent — cycle exits clean, user has no visibility
- No brain damage — other phases unaffected

REPRODUCTION
1. Postgres engine (not PGLite)
2. gbrain-creator schema pack active
3. gbrain dream --json
4. Observe extract_atoms phase: page-discovery query failed error on stderr,
   0 atoms in output

EXPECTED
extract_atoms should use the same connection path as surrounding phases, 
or re-establish connection after loadConfigWithEngine() completes.

foxhoundinc:
hermes notes

   This is a persistent upstream GBrain bug in the extract_atoms phase (part of the core cycle engine in the gbrain-latest checkout). We've seen it manifest with multiple different failure modes across sessions:                                                                                                       
                                                                                                                                                                                                                                                                                                                            
     - "No database connection: connect() has not been called."                                                                                                                                                                                                                                                             
     - SIGTERM aborts / 60s timeouts (current error)                                                                                                                                                                                                                                                                        
     - Silent hangs during page discovery                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                            
[...]

Issue #1515, opened 2026-05-26
dream-cycle extract.links_fs/timeline_fs phases silently fail with "No database connection" on Postgres engine (v0.41.14.0)


ChenyqThu:
## Summary

On a healthy Postgres setup (daemon `gbrain serve --http` working fine), invoking `gbrain dream` from CLI causes every batch insert during `extract.links_fs` and `extract.timeline_fs` to fail with `No database connection: connect() has not been called`. The phase nonetheless reports `status: ok` (because successful-count is 0 and no phase-level error is raised), so production cron silently no-ops indefinitely.

## Reproduction

Fork-side gbrain 0.41.14.0, macOS 26.3, Bun 1.3.x, Postgres 16 at `127.0.0.1:5432`, ~3100 markdown pages under `~/brain/`. `~/.gbrain/config.json`:

```json
{
  "engine": "postgres",
  "database_url": "postgresql://USER@127.0.0.1:5432/gbrain",
  "embedding_model": "google:gemini-embedding-001",
  "embedding_dimensions": 1536
}
```

The same Postgres process is used by `gbrain serve --http` (running concurrently and healthy — `/health` returns ok, `put_page` / `query` etc. all work via MCP).

```bash
gbrain dream --json --dir ~/brain
```

Output excerpt (3000+ batches all fail):

```
[cycle.extract] start
[extract.links_fs] start
[extract.links_fs] connection blip, retrying 100 rows in 500ms (No database connection: connect() has not been called. Fix: Run gbrain init --supabase or gbrain init --url <connection_string>)
  batch error (100 link rows lost): No database connection: connect() has not been called. Fix: Run gbrain init --supabase or gbrain init --url <connection_string>
[extract.links_fs] 30/3099 (0%)
[continues for all 3099 pages, every batch fails the same way]
...
[extract.links_fs] 3099/3099 (100%) done
[extract.timeline_fs] start
[same failure pattern]
[extract.timeline_fs] 3099/3099 (100%) done
[cycle.extract] done
```

Final JSON report from `gbrain dream --json`:

```json
{
  "status": "partial",
  "phases": [
    ...
    {
      "phase": "extract",
      "status": "ok",
      "summary": "0 link(s), 0 timeline entries (incremental: 0 slugs)"
    },
    ...
  ]
}
```

[...]

Issue #1491, opened 2026-05-26
autopilot: minion worker crashes with "No database connection: connect() has not been called"


justemu:
## Description

The minion worker subprocess (`gbrain jobs work`) spawned by `gbrain autopilot` crashes repeatedly because `engine.connect()` has not been called before the worker's main loop attempts to access the database.

This prevents ALL autopilot cycles from completing, and causes data loss during the extract phase.

## Environment

- **gbrain version**: 0.41.14.0
- **Engine**: Postgres
- **Config**: `agent.use_gateway_loop=true`, `chat_model: deepseek:deepseek-v4-flash`
- **OS**: Linux (WSL2 Ubuntu)

### 1. Worker crash loop (51 crashes recorded)

```
Promotion error: No database connection: connect() has not been called. Fix: Run gbrain init --supabase or gbrain init --url <connection_string>
[autopilot] worker exited code=1 signal=null after 1409ms, crashCount=4, cause=runtime_error
[autopilot] crash backoff 8111ms (crashCount=4)
```

Error counts:
- **51** worker crashes
- **265** "Promotion error" in `promoteDelayed()`
- **446** total "No database connection" errors
- **52** batch link row losses during extraction

### 2. Data loss during extract.links_fs

```
[extract.links_fs] 60/60 (100%)
[extract.links_fs] connection blip, retrying 26 rows in 500ms (No database connection: connect() has not been called. Fix: Run gbrain init --supabase or gbrain init --url <connection_string>)
  batch error (26 link rows lost): No database connection: connect() has not been called. Fix: Run gbrain init --supabase or gbrain init --url <connection_string>
[extract.links_fs] 60/60 (100%) done
Links: created 0 from 60 pages
```

### 3. Unhandled rejection during conversation_facts_backfill

```
[cycle.conversation_facts_backfill] start
[unhandledRejection] GBrainError: No database connection: connect() has not been called. Fix: Run gbrain init --supabase or gbrain init --url <connection_string>
    at getConnection (/.../gbrain/src/core/db.ts:153:15)
    at transaction (/.../gbrain/src/core/postgres-engine.ts:764:34)
    at failJob (/.../gbrain/src/core/minions/queue.ts:855:24)
    at executeJob (/.../gbrain/src/core/minions/worker.ts:831:39)
    at processTicksAndRejections (native:7:39)
```

### 4. All sources never complete a full cycle

```
[FAIL] cycle_freshness: Source 'aevum' has never completed a full cycle;
[...]

Issue #1462, opened 2026-05-26
facts:absorb logging can fail on disconnected engine handle


shandutta:
## Bug Description

`facts:absorb` is intended to make best-effort facts/backstop failures observable via `ingest_log` reason-coded rows. In at least the page-write / CLI facts-backstop queue path, the observability path itself can fail because `writeFactsAbsorbLog()` calls `engine.logIngest()` on an engine handle that is not connected.

The user-facing page write succeeds, but instead of a durable `ingest_log` row, stderr prints a one-off warning:

```text
[facts:absorb] suppressed: 'No database connection' fires on a separate engine handle (known WARN-4 in v0.38; subsequent occurrences silent this process). First-occurrence trace for v0.38.4 diagnosis:
GBrainError: No database connection: connect() has not been called. Fix: Run gbrain init --supabase or gbrain init --url <connection_string>
    at getConnection (.../src/core/db.ts:153:15)
    at logIngest (.../src/core/postgres-engine.ts:4014:17)
    at writeFactsAbsorbLog (.../src/core/facts/absorb-log.ts:81:18)
    at <anonymous> (.../src/core/facts/backstop.ts:173:15)
    at async runEntry (.../src/core/facts/queue.ts:166:19)
```

## Steps to Reproduce

Observed on `gbrain 0.41.11.1` during a normal page write where facts backstop was queued:

```bash
gbrain put personal/local-favorites < /tmp/gbrain-local-favorites.md \
  && gbrain get personal/local-favorites
```

The `put` response included successful durable write-through plus queued facts backstop:

```json
{
  "status": "created_or_updated",
  "facts_backstop": { "queued": true },
  "write_through": {
    "written": true,
    "path": "/home/.../brain/knowledge/personal/local-favorites.md"
  }
}
```

Then the warning above printed from the facts absorb logger.

## Expected Behavior

- The page write should continue to succeed. ✅
- If the optional facts backstop hits a failure / no-op / absorbed condition, the reason should be durably observable as an `ingest_log` row with:
  - `source_type='facts:absorb'`
  - stable reason-coded `summary` (`gateway_unavailable`, `parse_failure`, `embed_failure`, `no_candidates`, `pipeline_error`, etc.)
  - the relevant source/page/session ref
- Doctor/admin surfaces should be able to read those rows.
- No repeated noisy stderr spam.

## Actual Behavior

- The page write succeeds. ✅
- The facts backstop logger attempts to call `engine.logIngest(...)`.
[...]

PR #1555, opened 2026-05-27
[fix] Preserve shared Postgres connection during lint probes

### LINKED ISSUE #1471: resolveLintContentSanity disconnects shared module-level db singleton, killing the cycle's main engine connection
## Summary

When `gbrain dream` runs on Postgres, `resolveLintContentSanity` (`src/commands/lint.ts:298-339`) creates a fresh `PostgresEngine` to probe DB-plane lint config, then disconnects it in `finally`. That disconnect kills the module-level `db.ts` connection singleton that the **cycle's main engine** is still using — every subsequent DB-touching phase then throws `No database connection: connect() has not been called` and `lock.release()` throws `CONNECTION_ENDED` from `postgres.js`, stranding the row in `gbrain_cycle_locks`.

## Mechanism (root cause confirmed via stack-trace instrumentation)

1. CLI's `connectEngine()` creates `PostgresEngine` instance #1 (the cycle engine).
2. `#1.connect()` falls through to the `else` branch in `src/core/postgres-engine.ts:175` (no `poolSize` passed), calls `db.connect(config)`, which creates the module-level `sql` singleton. Sets `#1._connectionStyle = 'module'`.
3. Cycle runs lint as its first phase; `runLintCore` calls `resolveLintContentSanity()`.
4. `resolveLintContentSanity` calls `createEngine(...)` → `PostgresEngine` instance #2 to probe lifted config.
5. `#2.connect({})` also falls through to the `else` branch. `db.connect()` sees `sql` already set and returns early. Sets `#2._connectionStyle = 'module'`.
6. `lint.ts:319` calls `engine.disconnect()` on #2 in `finally`.
7. `#2.disconnect()` (`postgres-engine.ts:192-211`) sees `_connectionStyle === 'module'` and calls `db.disconnect()` → nulls the shared `sql` singleton.
8. Engine #1 (still in use by the cycle) is now broken. Every later phase throws `connect() has not been called`. `runCycle`'s finally then throws `CONNECTION_ENDED` on `lock.release()`.

The comment at `postgres-engine.ts:94-100` explicitly anticipates this category of bug, but the existing `_connectionStyle: 'instance' | 'module' | null` flag doesn't distinguish a singleton **owner** from a singleton **borrower** — both call `db.disconnect()`.

## Repro

```sh
psql gbrain -c "DELETE FROM gbrain_cycle_locks;"
gbrain dream --dir ~/path/to/brain
psql gbrain -c "SELECT id, holder_pid FROM gbrain_cycle_locks;"
```

Expect: 1.4s run, every DB phase reports `No database connection: connect() has not been called`, stranded row in `gbrain_cycle_locks`.

(In a vanilla 0.41.2.0 install this currently surfaces differently because of a separate Bun WebCrypto issue that masks the post-extract_facts phases — see linked Bun issue draft. Switching `src/core/schema-pack/manifest-v1.ts:274` and `closure.ts:175` to sync `node:crypto` exposes the singleton-clobber bug cleanly.)

## Suggested fix

Track singleton ownership in `PostgresEngine`. Only the instance whose `connect()` call actually created the singleton may disconnect it. Borrowers just clear their own marker:

```ts
// postgres-engine.ts — add field
private _ownsModuleSingleton: boolean = false;

// in connect(), else branch (the no-poolSize / module-singleton path):
} else {
  // Determine ownership BEFORE delegating to db.connect.
  let alreadyConnected = false;
  try { db.getConnection(); alreadyConnected = true; } catch { /* sql null = we'll be the owner */ }
  await db.connect(config);
  this._ownsModuleSingleton = !alreadyConnected;
  // ... existing connectionManager wiring
}

// in disconnect():
  if (this._ownsModuleSingleton) {
    await db.disconnect();
    this._ownsModuleSingleton = false;
  }
  this._connectionStyle = null;
}
```

This fix is minimal, backward-compatible, and addresses the root cause rather than symptoms. The CLI's `connectEngine()` engine becomes the owner; any helper that creates a probe engine (lint, doctor, future probes) becomes a borrower automatically.

An alternative — refactor `resolveLintContentSanity` to accept an engine from the caller and skip its own probe — is also defensible but spreads the API contract through more call sites. The ownership flag contains the bug to the place it was introduced.

## Verification

After the fix:
- Dream completes all 19 phases doing real DB work (sync adds pages, extract_facts reconciles, embed runs, etc.)
- Lock row released cleanly each run (zero rows in `gbrain_cycle_locks`)
- Per-phase report shows `✓` on every DB phase, not `✗ ... connect() has not been called`

Tested locally on a 12,500-page brain over multiple runs. Full diagnostic write-up + stack traces available if useful.

## Environment

- gbrain 0.41.2.0
- Bun 1.3.14
- Postgres on macOS (Apple Silicon, localhost)

## Related

Companion issue: silent `lock.release()` swallow (made this bug invisible for an unknown time).

### LINKED ISSUE #1492: dream: phases fail with "No database connection" on Postgres backend (lint+backlinks succeed in same run)
### Summary

`gbrain dream` (and `gbrain dream --dry-run`) consistently fails with `No database connection: connect() has not been called` on every phase that needs the brain engine, even though the serve process at the same `DATABASE_URL` works fine. `lint` and `backlinks` (which presumably use a different connect path) complete successfully in the same invocation.

### Environment

- gbrain v0.41.10.1, commit `d036a97`
- Backend: Postgres (Supabase self-hosted; not PGLite)
- `gbrain init --url <postgres-url>` ran cleanly on container first-boot; `serve --http --port 3131` works, OAuth flow works, `query`/`search`/`extract links` all hit the DB normally
- `/var/lib/gbrain/.gbrain/config.json` contains `engine` + `database_url` fields (6 lines total)

### Repro

```bash
docker exec gbrain gbrain dream --dry-run
```

Output (abridged):

```
  ! lint        523 issue(s) found (dry-run, no writes)
  ✓ backlinks   no missing back-links found
  ✗ sync        sync phase failed
      [InternalError/UNKNOWN] No database connection: connect() has not been called. Fix: Run gbrain init --supabase or gbrain init --url <connection_string>
  ✗ synthesize  synthesize phase failed
  ✗ extract_facts ...
  ✗ patterns ...
  ✗ recompute_emotional_weight ...
  ✗ consolidate ...
  ✗ propose_takes ...
  ✗ grade_takes ...
  ✗ calibration_profile ...
  ✗ embed ...
  ✗ orphans ...
  ✗ schema-suggest ...
```

The same failure occurs with `HOME=/var/lib/gbrain` explicitly passed via `docker exec -e HOME=...`. The cycle lock IS acquired (a row appears in `gbrain_cycle_locks`), so SOME DB connection is being established — just not the one each phase needs.

### Impact

Dream cycle (the documented backbone of "consolidate → propose_takes → grade_takes → calibration_profile") is unusable on Postgres backends. Brains running on the Postgres engine can't populate the `facts` table, `recall` returns empty, autopilot doesn't help (presumably hits the same path).

### Guess at root cause

The phase runner appears to construct its own brain-engine instance without inheriting the connection pool the parent `dream` process used to acquire the cycle lock. Looks like a DI gap — the lock-subsystem connect happens, the per-phase engine instances don't reuse it and don't reconnect from config.

### Workaround used

Switched to running individual extract-and-embed steps via host cron (`gbrain extract links --source db && gbrain embed --stale`) until this lands. Drops fact consolidation / take grading entirely.

### Related

- The lock row in `gbrain_cycle_locks` has a 30-min TTL but stays held on phase failure (acquired by failed dream-cycle, not released on the engine-side crash). Workaround: `DELETE FROM gbrain_cycle_locks WHERE id='gbrain-cycle';` as superuser. Possibly worth a separate `gbrain dream --abort` CLI.

### LINKED ISSUE #1499: Bug: lint phase engine.disconnect() kills worker shared DB pool, causing crash-loop

flamerged:
## Summary

This fixes a Postgres connection-lifecycle bug where short-lived helper engines could tear down the shared module-level database pool that the main `gbrain dream` cycle still needed.

In the dream cycle, lint runs early and briefly opens an engine to lift DB-backed content-sanity config. On Postgres, that helper can reuse the existing module-level singleton. Before this change, its `disconnect()` path always called `db.disconnect()` for module-mode engines, even when the helper did not create the singleton. That leaves later phases with `No database connection` / `connect() has not been called`.

## Issue cluster

Fixes #1492.
Fixes #1471.
Fixes #1499.

Those reports converge on the same root cause: an early lint/config helper borrows the shared Postgres module singleton, then disconnects it out from under the still-running dream cycle or worker pool.

This also targets the same Postgres "later phase loses the DB connection after an earlier helper phase ran" symptom cluster reported in #1491, #1515, #1531, #1462, and #1422. I am not marking every adjacent report as auto-closed because some include follow-on error-reporting, worker-startup, idle-timeout, or phase-specific details, but this PR removes the shared-pool teardown bug they converge on.

Related but likely separate: #1426 appears to involve PGLite pending-write lifecycle behavior, so this PR does not claim to fix it. #1404 and #1413 have overlapping `connect() has not been called` evidence, but their later comments also discuss additional lifecycle paths outside this narrow ownership fix.

## Why this matters

This is a small lifecycle fix with high operational impact:

- Keeps long-running dream cycles from losing their Postgres connection after the lint/config probe.
- Preserves the existing module-singleton compatibility path for CLI flows.
- Keeps instance-owned worker pools unchanged.
- Makes ownership explicit: only the engine that created the module singleton can close it; borrower engines clean up their own routing manager but leave the shared pool alive.

For Postgres users, this turns a whole class of hard-to-diagnose dream-cycle crashes back into stable multi-phase runs without changing query semantics or pool setup.

## What changed

- Added module-singleton ownership tracking to `PostgresEngine`.
- Updated `disconnect()` so module-mode borrowers do not call `db.disconnect()`.
- Expanded the Postgres disconnect E2E coverage to pin:
  - instance-pool double-disconnect safety,
  - module-borrower disconnect safety,
  - the lint DB-config probe path,
  - module-owner disconnect idempotency.

## Validation

- `DATABASE_URL=postgresql://postgres:postgres@127.0.0.1:54322/<scratch-db> bun test test/e2e/postgres-engine-disconnect-idempotency.test.ts`
- `bun test test/cli-dream-engine-warn.test.ts test/dream.test.ts`
- `bun run typecheck`
- `git diff --check`
- `git show --check HEAD`
- `scripts/check-no-double-retry.sh && scripts/check-batch-audit-site.sh && scripts/check-test-isolation.sh`
- `gitleaks dir . --redact --no-banner`
- `gitleaks git . --redact --no-banner --log-opts="origin/master..HEAD"`
- CodeRabbit local review: 0 findings
- Claude Code prompt review: no blockers; low-severity comment/test-context nits addressed

### DIFF
--- a/src/core/postgres-engine.ts
+++ b/src/core/postgres-engine.ts
@@ -97,12 +97,19 @@ export class PostgresEngine implements BrainEngine {
   /**
    * Tracks which connection path this engine is using so disconnect() is
    * idempotent. 'instance' = own _sql pool (poolSize was set);
-   * 'module' = the module-level db singleton (backward compat path).
+   * 'module' = the module-level db singleton (owned or borrowed).
    * null = never connected, or already disconnected. Without this, a second
    * disconnect() on an instance-pool engine would fall through to
    * db.disconnect() and clobber the unrelated module-level connection.
    */
   private _connectionStyle: 'instance' | 'module' | null = null;
+  /**
+   * True only for the module-style engine instance that created the db.ts
+   * singleton. Other module-style engines are borrowers and must not tear down
+   * the shared pool in disconnect(). Borrowers assume the owner manages the
+   * module singleton lifetime, as in the CLI main-engine plus helper-probe flow.
+   */
+  private _ownsModuleConnection = false;
 
   /**
    * v0.30.1 (Fix 1 + X1 + T5): instance-owned ConnectionManager.
@@ -165,6 +172,7 @@ export class PostgresEngine implements BrainEngine {
       await this._sql`SELECT 1`;
       await db.setSessionDefaults(this._sql);
       this._connectionStyle = 'instance';
+      this._ownsModuleConnection = false;
 
       // v0.30.1: instance-owned ConnectionManager wraps the read pool we just
       // built. Parent inheritance (T5/X1): worker engines pass their parent's
@@ -177,8 +185,16 @@ export class PostgresEngine implements BrainEngine {
       this.connectionManager.setReadPool(this._sql);
     } else {
       // Module-level singleton (backward compat for CLI main engine)
+      let moduleConnectionExists = true;
+      try {
+        db.getConnection();
+      } catch {
+        moduleConnectionExists = false;
+      }
+      const alreadyOwnsModuleConnection = this._connectionStyle === 'module' && this._ownsModuleConnection;
       await db.connect(config);
       this._connectionStyle = 'module';
+      this._ownsModuleConnection = alreadyOwnsModuleConnection || !moduleConnectionExists;
 
       // v0.30.1: connection-manager wraps the module singleton.
       if (url) {
@@ -201,14 +217,18 @@ export class PostgresEngine implements BrainEngine {
     if (this._sql) {
       await this._sql.end();
       this._sql = null;
+      this._ownsModuleConnection = false;
       // After this point, _connectionStyle stays 'instance' so a second
       // disconnect() is a no-op rather than falling through and clearing
       // the unrelated module-level db singleton.
       return;
     }
     if (this._connectionStyle === 'module') {
-      await db.disconnect();
+      if (this._ownsModuleConnection) {
+        await db.disconnect();
+      }
       this._connectionStyle = null;
+      this._ownsModuleConnection = false;
     }
     // else: nothing to disconnect (already done or never connected)
   }
--- a/test/e2e/postgres-engine-disconnect-idempotency.test.ts
+++ b/test/e2e/postgres-engine-disconnect-idempotency.test.ts
@@ -11,21 +11,29 @@
  * connection" on the next beforeEach).
  *
  * The fix: PostgresEngine tracks `_connectionStyle` ('instance' | 'module')
- * and only calls db.disconnect() when it actually owns the module-level
- * connection. Second disconnect on an instance-pool engine is a no-op.
+ * plus module-singleton ownership. Only the engine that created the module-level
+ * connection calls db.disconnect(); borrowers leave the shared pool alive.
+ * Second disconnect on an instance-pool engine is a no-op.
  *
  * This test pins the contract so future refactors of disconnect() can't
  * silently regress (it's exactly the bug class that took an hour of E2E
- * debugging to find). Two cases:
+ * debugging to find). Four cases:
  *   1. instance-pool engine: connect → disconnect → disconnect must NOT
  *      affect the module-level connection.
- *   2. module-singleton engine: connect → disconnect → disconnect is safe
- *      (second call no-ops).
+ *   2. module-singleton borrower: connect → disconnect must NOT affect the
+ *      owning module-level connection.
+ *   3. lint's DB-config probe must leave the owner alive.
+ *   4. module-singleton owner: connect → disconnect → disconnect is safe
+ *      (first call closes the singleton; second call no-ops).
  */
 
-import { describe, test, expect, beforeAll, afterAll } from 'bun:test';
+import { describe, test, expect, beforeEach, afterEach } from 'bun:test';
+import { mkdtempSync, rmSync, writeFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
 import { PostgresEngine } from '../../src/core/postgres-engine.ts';
 import * as db from '../../src/core/db.ts';
+import { runLintCore } from '../../src/commands/lint.ts';
 
 const DATABASE_URL = process.env.DATABASE_URL;
 const skip = !DATABASE_URL;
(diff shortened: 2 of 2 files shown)


garrytan:

Issue #1036, opened 2026-05-15
admin: dashboard at :3131/admin does not expose redirect_uris field for OAuth client registration


dcarolan1:
Affected version 
gbrain 0.33.1.1 — HEAD cb8d6d8 
Steps to reproduce 
Stand up a gbrain serve --http --port 3131 instance with admin enabled
Navigate to https://<host>:3131/admin in a browser
Open the OAuth client registration form 
Expected behavior 
The form includes a redirect_uris field (text input, comma-separated or one-per-line), required for authorization_code flow clients, optional
otherwise. Per the OAuth 2.1 spec, authorization_code clients without registered redirect_uris are non-functional for browser flows. 
Actual behavior 
The form omits redirect_uris entirely. Submitting the form registers the client with redirect_uris = [], and any subsequent
authorization_code request from that client fails at the redirect_uri match step in the IdP. 
Workaround 
Skip the admin UI entirely. Register clients via: - The CLI (gbrain auth register-client) — but the CLI has its own redirect_uris bug, see issue
[#XXX-bug-1] (auth: CLI register-client hardcodes empty redirect_uris regardless of --redirect-uri input) - A direct Bun
script invoking GBrainOAuthProvider.registerClientManual() with the four-argument form including redirect_uris 
Used the Bun-script workaround Day 19 (2026-05-14) during MCP bridge stand-up. Both workarounds bypass the admin UI; neither is acceptable for
self-service OAuth client onboarding by non-operators. 
Severity / impact 
Functional blocker for production OAuth client onboarding via the admin UI. 
The admin UI is the documented self-service path for non-operator users
Both fallback paths (CLI + direct API) have their own gaps; CLI is broken (issue 1), direct API requires DB access
Forces every OAuth client setup through an operator-with-DB-access workflow, defeating the purpose of an admin UI 
Suggested fix 
Add a redirect_uris field to the admin OAuth client registration form (text input, one URI per line or comma-separated), validated against URL
syntax, forwarded to provider.registerClientManual()’s fourth argument. 
Related 
See issue [#XXX-bug-1] for the matching CLI bug — they share the same redirect_uris plumbing gap and may share a fix.

time-attack:
[...]

PR #1558, opened 2026-05-27
[fix] Add source scoping to admin OAuth registration

### LINKED ISSUE #1490: admin: "Register Agent" form does not expose source_id / federated_read — multi-source clients silently broken
### Summary

The "Register Agent" form in the admin dashboard at `:3131/admin` exposes only three fields (Agent Name, Scopes, Token Lifetime), but `oauth_clients` has ~15 columns. For brains with anything other than a single `default` source, this silently produces clients that **cannot see any content** unless the operator knows to run a SQL `UPDATE` or recreate the client via CLI.

### Repro

1. Install gbrain (any v0.34.1+). Add federated sources, e.g.:
   ```bash
   gbrain sources add notion-active --path /data/repos/notion-active --federated
   gbrain sources add notion-archive --path /data/repos/notion-archive --federated
   gbrain sync --source notion-active --no-pull
   gbrain sync --source notion-archive --no-pull
   ```
2. Open the admin dashboard → click `+ OAuth Client` → fill in `Name: claude-test`, check `read` + `write`, leave token lifetime default → Register.
3. Use the returned `client_id`/`client_secret` to do `client_credentials` exchange → token returns scope `read write` ✓.
4. Call `tools/list` → returns 76 tools ✓.
5. Call `tools/call` `list_pages` (no `source_id`) → returns `[]`.
6. Call `query` with any string → `[]`.
7. Even passing `source_id: "__all__"` → still `[]`.

### Root cause

`src/core/oauth-provider.ts` `registerClient` (the handler behind `POST /admin/api/register-client`) hardcodes the row:

```ts
INSERT INTO oauth_clients (..., source_id, federated_read)
VALUES (..., 'default', ARRAY['default'])
```

The form's body has no path to override this. The brain has content in `notion-active` and `notion-archive` but the new client's `federated_read = {default}` filters all rows out.

### Why this matters

Per `docs/tutorials/company-brain.md`, multi-source / multi-user setups are an upstream-supported pattern. The tutorial **does** show how to register clients correctly via `gbrain auth register-client --source X --federated-read Y,Z`. But the admin UI dashboard appears to be the obvious entry point for new operators, and it silently steers them into a broken state with no warning. The form succeeds, the token exchanges fine, scope reports `read write` — and reads still return `[]`. Debugging takes hours.

### Adjacent gap

There is no `gbrain auth update-client` / `rescope` subcommand. The comment in `src/core/oauth-provider.ts` even says:

> "Operators who need narrower / wider scope rescope via the CLI later."

— but the rescope CLI isn't shipped. Currently the only path to fix an already-registered client is direct `UPDATE oauth_clients SET federated_read = ...` against Postgres. The same gap is acknowledged in `TODOS.md` for the threading work in 14 other read-side ops.

### Proposed fix

Either:
1. **Add the missing fields to the Register Agent form**: `source_id` (dropdown of known sources), `federated_read` (multi-select of known sources), and optionally the `bound_*` knobs.
2. **Add a Rescope/Edit button + dialog on the agent row** in the Agents page, since post-create updates are also needed when sources are added.
3. At minimum, **warn in the form**: "Multi-source brain detected — clients created here will default to the `default` source. Use `gbrain auth register-client --source X --federated-read Y,Z` for federated scoping." Cheap to ship, gets operators to the documented path.

Related: #1036 (same class of bug — `redirect_uris` field missing from the same form).

### Workaround for operators reading this

```bash
docker exec gbrain gbrain auth register-client "<name>" \
  --scopes "read write" \
  --source <primary-source-for-writes> \
  --federated-read <source1>,<source2>,<...>
```

— or after-the-fact, the SQL one-liner:

```sql
UPDATE oauth_clients
SET federated_read = ARRAY['source1','source2','default'],
    source_id      = 'source1'
WHERE client_id = 'gbrain_cl_<...>';
```

The `federated_read` column is JOINed live from `oauth_clients` on every auth lookup (see `oauth-provider.ts` `validateToken`), so the SQL takes effect immediately on the existing access token — no client rotation needed.

### Environment

- gbrain v0.41.10.1 (commit d036a97)
- engine: postgres (Supabase)
- transport: HTTP MCP
- multi-source brain with 2 federated sources

flamerged:
## Summary

- Add write-source and read-source controls to the admin **Register Agent** form.
- Add an authenticated `/admin/api/sources` endpoint so the UI can offer the same known source IDs the engine uses.
- Validate `source_id` and `federated_read` on the admin registration endpoint before creating the OAuth client.
- Surface source scope in the Agents table, drawer details, and JSON export.

## Why this matters

The admin dashboard is the obvious path for first-time operators creating MCP clients. Before this change, clients created there always landed on the `default` source with `federated_read = {default}`. On multi-source brains that makes the token exchange look successful while reads like `list_pages` and `query` return empty results for content stored in other sources.

This brings the admin path up to parity with the CLI registration path: an operator can create a client that writes to the intended source and reads across the selected federated sources without hand-editing `oauth_clients` afterward.

## Notes

- Existing default behavior is preserved: if no source fields are supplied, the admin endpoint uses `default` when available.
- API keys still show `all` in the UI because their existing bearer-token behavior is not changed.
- This does not try to solve the adjacent redirect URI editing gap from #1036.

Fixes #1490.

## Verification

- `git diff --check`
- `bun run typecheck`
- `bun test test/admin-source-scope.test.ts test/oauth.test.ts test/fix-wave-structural.test.ts`
- `bun run build:admin`
- `bun run check:admin-build`
- `bun run check:admin-embedded`
- `gitleaks dir . --redact --no-banner`
- `gitleaks git . --redact --no-banner --log-opts="origin/master..HEAD"`
- `coderabbit review --plain` (no findings)


time-attack:
Solid writeup, and the underlying concern was legitimate — a client silently pinned to `default` on a multi-source brain looks like a successful token exchange while every read comes back empty. Master has since closed the blocking part of that: `/admin/api/sources` is served at `src/commands/serve-http.ts:1185`, the agents endpoint returns `source_id` and `federated_read`, and `admin/src/pages/Agents.tsx:416` ships a `SourceAccessEditor` that rescopes both from the agent drawer, so no raw SQL is needed. What's left here is setting scope at registration time rather than immediately after, which is convenience surface plus rebuilt `admin/dist` bundles that conflict on every admin build, so I'm closing it.

PR #1559, opened 2026-05-27
fix: dream --source passes sourceId to runCycle, writes cycle timestamp

garrytan-agents:
## Problem

`gbrain dream --source default` silently ignores the `--source` flag. The flag was never parsed in `parseArgs` and never forwarded to `runCycle`.

**Impact:** `runCycle` writes `last_full_cycle_at` to the source config ONLY when `opts.sourceId` is set (line 1961 of cycle.ts). Without it, the cycle completes successfully but the doctor cycle_freshness check always reports stale — because no timestamp was ever written.

This means:
- `gbrain dream --source default` via cron → cycle runs → doctor says 60h stale → operator re-runs → cycle runs → still stale → infinite loop
- Only `gbrain autopilot` (which has its own fanout logic) could write the timestamp
- Every manual or cron-triggered dream run was invisible to doctor

## Fix

1. Parse `--source <id>` in `parseArgs` (was completely missing)
2. Parse `--max-pages <n>` (was silently ignored)
3. Forward both to `runCycle` opts
4. Document in `--help`

One-line root cause: `sourceId` was in `CycleOpts` but `dream.ts` never set it.

## Testing

```bash
# Before: cycle completes, no timestamp written
gbrain dream --source default --json | jq .status  # "ok"
# sources.config still shows old last_full_cycle_at

# After: cycle completes, timestamp written
gbrain dream --source default --json | jq .status  # "ok"
# sources.config.last_full_cycle_at = now
# doctor cycle_freshness check passes
```


garrytan:
Superseded by #1571: same intent, canonical `resolveSourceId` + `fetchSource` wiring, + engine-null and archived-source guards, + typed-error try/catch via predicate (so TypeError doesn't get hidden as operator-error), + 13 PGLite integration cases (back-compat, alias equivalence, repetition rules, --help short-circuit ordering, D5 end-to-end dream→checkCycleFreshness column-name drift guard). Drops the silently-dead --max-pages flag (cycle phases don't honor it today — shipping the flag would have been a lying flag). The --max-pages plumbing is filed as v0.41+ TODO-V13-A.

The shared surrogate-safe commit ships via the canonical `safeSplitIndex` helper instead of `safeSliceEnd` — see `src/core/text-safe.ts` module docstring for why the canonical helper exists (the case-3 bug the agent-authored `safeSliceEnd` re-introduces).

Thank you for catching the bug.

PR #1561, opened 2026-05-27
fix: expand error_page_title pattern to catch scraper artifacts at ingest

garrytan-agents:
232 scraper error pages (202 from straylight-brain) persisted in the DB because the error_page_title content-sanity pattern only matched bare numeric codes. Expanded to catch Just a moment, Access Denied, Error, Forbidden, Service Unavailable, Robot Check. All anchored so legitimate pages wont trip.


garrytan:
Superseded by #1571: same intent, with three differences. (1) **Distinct pattern names**: `error_page_title` keeps the bare numeric codes and the new title-only phrases (Forbidden / Access Denied / Service Unavailable / Robot Check / Verify You Are Human); the Cloudflare title gets its own name (`cloudflare_challenge_title`) so audit JSONL aggregation stays diagnosable (the original PR reused `error_page_title` for both and collapsed the audit signal). (2) **Drops the bare `error` matcher** — too aggressive on legitimate concept/taxonomy pages titled exactly 'Error'. (3) **Over-match regression guard tests**: 'How to Handle Access Denied Errors', 'Error Boundary in React', 'Service Unavailable Pattern', 'Forbidden Knowledge' all verified to still ingest cleanly.

The shared surrogate-safe commit ships via the canonical `safeSplitIndex` helper instead of `safeSliceEnd` — see #1571's commit `02b1f5c` for the explanation.

`gbrain pages audit-junk-titles` legacy cleanup is filed as v0.41+ TODO-V13-C (full spec preserved). Deferred from this PR per cross-model tension review for ship-and-validate-matchers-first discipline — destructive cleanup gets its own observation window after the matcher proves itself in production for ~1 week.

Thank you for catching the bug.

PR #1564, opened 2026-05-27
fix: skip sync/cycle freshness warnings when upstream repo is unchanged

garrytan-agents:
## Problem

Doctor flags sources like `media-corpus` as stale after 40h even when zero commits have changed. This creates false-positive warnings and unnecessary sync/cycle runs.

Garry: "if there have been no commits to that particular corpus, then it's not stale, actually."

## Fix

Add `isSourceUnchangedSinceSync()` — a fast `git rev-parse HEAD` check that compares the repo's current HEAD against the `last_commit` stored in the DB. If they match, the source is up-to-date regardless of wall-clock age.

Wired into both `checkSyncFreshness` and `checkCycleFreshness`.

## Changes

- `doctor.ts`: add `isSourceUnchangedSinceSync()` helper, wire into both freshness checks
- `engine.ts`: add `last_commit` to `SourceRow` interface
- `postgres-engine.ts`: include `last_commit` in `listAllSources` query
- `pglite-engine.ts`: same for PGLite implementation

## Safety

Fail-open: any error in the git check (missing repo, corrupt state) falls through to the normal time-based freshness logic. No behavior change for non-git sources.


garrytan:
Superseded by production-quality rebuild in workspace `irvine-v1` (branch `garrytan/sync-freshness-git-check`). The fix is correct in shape but the rebuild adds:

1. **Shell-injection safety** — `execSync('git -C ${JSON.stringify(path)} ...')` runs through `/bin/sh -c`; JSON.stringify escapes for JSON, not shell. Rebuild uses `execFileSync('git', ['-C', path, 'rev-parse', 'HEAD'])` — no shell, no injection surface.
2. **Unit + integration test coverage** — helper has its own `test/core/git-head.test.ts`; `checkSyncFreshness` + `checkCycleFreshness` tests in `test/doctor.test.ts` extended with unchanged-source / mismatched-HEAD / NULL-last_commit / non-git-path / mixed-source cases.
3. **Shared primitive** — `isSourceUnchangedSinceSync` lives in `src/core/git-head.ts` (reusable by autopilot decision paths) instead of buried in doctor.ts.
4. **More honest ok message** — when sources are skipped via git-check vs time-check, the ok line breaks down both counts so the operator sees what's happening.

Surrogate-pair fix from commit 78b93f3f is NOT brought over — already on master via `safeSplitIndex` in v0.42.0.0.

Thank you for the contribution @garrytan-agents; Co-Authored-By preserved in the rebuild commits.

PR #1565, opened 2026-05-27
feat: add opt-in plain-text entity extraction for auto-link

donogeme:
## Summary

Adds a new config flag `auto_link_plain_text` (default: `false`) that enables automatic extraction of plain-text entity names from markdown content during `runAutoLink`.

## Problem

Currently, auto-link only creates graph edges for explicit `[[wikilinks]]`. Content that mentions people by plain text (e.g., "Patrick Skinner mentioned that...") doesn't get linked to their person pages, even when those pages exist.

## Solution

When `auto_link_plain_text = true`:
1. After extracting explicit wikilinks, the system runs `extractEntities()` on the markdown body
2. This uses a regex NER approach to find capitalized multi-word names (e.g., "Patrick Skinner", "John Chen")
3. For each candidate, it generates an expected slug via `slugifyEntity()` (e.g., `people/patrick-skinner`)
4. If the slug resolves to an existing page, a `mentions` edge is created with `link_source='plain_text'`
5. Dedupes against explicit wikilinks to avoid double-linking

## Changes

- `src/core/link-extraction.ts`: Added `isPlainTextEntityExtractionEnabled()` config helper
- `src/core/operations.ts`: Wire `extractEntities` into `runAutoLink`, add `plain_text` to reconcilable sources
- `test/link-extraction.test.ts`: 7 new tests for config flag behavior

## Testing

```bash
bun test test/link-extraction.test.ts
# 105/105 pass
```

## Usage

```bash
gbrain config set auto_link_plain_text true
gbrain sync --source <source> --full  # re-process existing pages
```

## Notes

- Opt-in by default to preserve backward compatibility
- The regex pattern `/\b([A-Z][a-z]+(?:\s+[A-Z][a-z]+)+)\b/g` catches most Western-style names
- Could be extended with more sophisticated NER in the future

### DIFF
--- a/src/core/link-extraction.ts
+++ b/src/core/link-extraction.ts
@@ -13,6 +13,7 @@
 
 import type { BrainEngine } from './engine.ts';
 import type { PageType } from './types.ts';
+import { extractEntities } from './enrichment-service.ts';
 
 // ─── Entity references ──────────────────────────────────────────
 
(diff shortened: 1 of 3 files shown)


time-attack:
Thanks for the clean, opt-in-by-default framing and the test coverage. This is new feature surface rather than a fix: explicit `[[wikilink]]` extraction works as designed, so nothing is broken today. The regex heuristic also worries me for a graph that users trust — `/\b([A-Z][a-z]+(?:\s+[A-Z][a-z]+)+)\b/g` matches any title-cased phrase, so things like place names or sentence-initial pairs can slug-collide with a real person page and write false `mentions` edges that are tedious to unwind. Closing as feature surface we're not taking.

PR #1572, opened 2026-05-27
v0.41.26.1 fix: lock-renewal cathedral — closes ~39 worker crashes/day (supersedes #1567)

garrytan:
## Summary

- **Closes the production worker crash class** that was dropping ~39 daemons/day on Supabase + PgBouncer with `unhandledRejection at renewLock`.
- **Closes a second unhandledRejection vector** the inside-review missed: the stored `executeJob(...).finally(...)` promise had no `.catch()`, so failJob/completeJob throws during the same DB outage would crash the daemon too.
- **Privacy backfill** for audit JSONL — new shared `redactConnectionInfo` helper strips Postgres URLs/host/user/password/IPv4 from error messages before they hit disk. Wired into both new lock-renewal-audit AND existing batch-retry-audit (same risk class).

## Why this supersedes #1567

PR #1567 (contributor `@garrytan-agents`) proposed the right try/catch shape but was missing:
- Re-entrancy guard (overlapping ticks during PgBouncer stalls would pile concurrent connection acquisitions).
- Per-call timeout (hung renewLock would wedge the worker forever).
- Time-based abort (count-based 3-strike at lockDuration=30s let other workers reclaim BEFORE we aborted — concurrent execution risk).
- Infrastructure-reason guard (PgBouncer blips would dead-letter healthy jobs).
- Universal grace-eviction (the 30s force-evict safety net only fired for `job.timeout_ms`, not lock-renewal aborts).
- Privacy redactor (audit JSONL leaked credentials).
- Tests (zero coverage on the regression class).
- Bundled unrelated work (`safeSliceEnd` UTF-16 surrogate fix — different bug class, filed as TODO-LR-1).

The plan-eng-review surfaced 4 of these gaps inside, and codex outside-voice surfaced the other 4. All 8 absorbed via 9 locked design decisions (D1-D9) captured in the plan file.

## Architecture

Pure-function extraction is the load-bearing change:
- `src/core/minions/lock-renewal-tick.ts` — `runLockRenewalTick(deps, state)` returns a tagged union (`'ok' | 'cancelled' | 'lock_lost' | 'should_abort'`). All side effects (clearInterval, abort.abort, finally cleanup) stay in the worker's closure scope. Tests drive the state machine with fake deps — no `setInterval` plumbing, no PGLite, no `mock.module`.
- `worker.ts:launchJob` reduces to a thin `setInterval(() => void runLockRenewalTick(...))` sync wrapper.
- `scripts/check-worker-lock-renewal-shape.sh` CI guard wires into `bun run verify`. Asserts the bug pattern (`lockTimer = setInterval(async ...)`) stays absent AND the call site survives refactors. Narrowed via `lockTimer =` prefix so unrelated `setInterval(async)` calls elsewhere in the file (e.g. stall detector — codex C13, filed separately as TODO-LR-4) don't false-fire.

## Test plan

- [x] 18 hermetic pure-function tests on `runLockRenewalTick` (state machine paths, cancellation, time-based abort, hung-renewLock timeout, audit defense-in-depth, env-knob resolution)
- [x] 11 audit module tests (4-outcome contract, redactor wired, readback, corrupted-line tolerance, pruning)
- [x] 15 redactor unit tests (all 5 patterns, IPv4-vs-version-number defense, idempotency, real Supabase/ENOTFOUND fixtures)
- [x] 3 privacy backfill regression tests on `batch-retry-audit` (snapshot JSONL — no DSN/host/password)
- [x] 5 CI guard meta-tests (good shape passes, bug pattern fails, missing call-site fails, false-positive defense for stall-detector pattern elsewhere)
- [x] 12 existing batch-retry audit tests still pass (privacy backfill non-breaking)
- [x] 169 existing minion tests still pass
- [x] 13 existing worker-shutdown / worker-rss tests still pass
- [x] `bun run verify` green (29 checks including the new shape guard)

Total: 64 new cases + 182 existing tests verified. All hermetic.

Test failures in the parallel suite (longmemeval timeouts, schema-pack timeouts, bootstrap PGLite cold-start timeouts, query-cache flakes, eval-prune timeouts) are pre-existing infrastructure-flake under heavy load — none touch any code in this PR.

## Operator verification after upgrade

```bash
# Confirm the shape guard is wired.
bun run check:worker-lock-renewal-shape  # → "lock-renewal shape OK"

# Watch the audit channel during a real Supabase reconnect.
tail -F ~/.gbrain/audit/lock-renewal-*.jsonl

# Force a blip if you can:
docker exec your-pgbouncer-container pkill -HUP pgbouncer
# Expected: 1-2 `failure` events then `success_after_failure` on recovery.
# Worker process MUST NOT exit. If it does, file an issue.
```

## Closes

- #1567 (incorporates contributor's try/catch shape with full attribution; closes the bug class structurally)


garrytan:
**Test coverage gap-fill landed in a8b282d4** (closes the post-ship gap audit covering A-H).

Two new test files, 19 new cases, 5.2s wallclock:

- `test/worker-lock-renewal-e2e.serial.test.ts` (1 test, gap **H** — gold-standard regression): real PGLite + real MinionWorker + executeRaw wrap that injects renewLock failures. Pins the headline behavioral contract: worker process MUST NOT crash via unhandledRejection under sustained renewLock throws (the exact v0.41.22.1 production bug class), the handler observes `abort.signal.aborted` with reason `lock-renewal-failed`, and the audit JSONL contains both `failure` and `gave_up` events.

- `test/worker-lock-renewal-shape.test.ts` (18 tests, gaps **A, B, C, D, E, F, G**): source-shape behavioral pins. Greps worker.ts function bodies for the patterns the locked decisions promised — launchJob calls `runLockRenewalTick` + `resolveLockRenewalKnobs` + `lockRenewalAudit`; `tickInFlight` declared and gated; stored executeJob promise has `.catch` with `logExecuteJobRejected` + console stderr; `abort.signal.addEventListener('abort', ...)` fires for any abort reason; `INFRASTRUCTURE_ABORT_REASONS` consulted inside executeJob's catch with return-early shape. Bug-pattern-specific by design.

Why E2E sits in its own single-test file: bun:test's serial runner has an unresolved interaction with PGLite when multiple MinionWorker-driven tests share a file (the second test's `queue.add` hangs indefinitely even with cleanly reset state). Isolating H + using source-shape pins for the wiring (which are stronger regression guards anyway — they survive refactors that change implementation shape but preserve the contract) is the cleanest split.

**Total wave coverage:** 83 cases across 8 files, all hermetic, all pass in 5.2s. 205 existing minion + worker tests still green.

Issue #1570, opened 2026-05-27
bug: withRetry doesn't reconnect on 'No database connection' — batch rows lost every dream cycle


garrytan-agents:
## Problem

`gbrain dream` cycle loses ~150 rows per run to batch errors and `conversation_facts_backfill` crashes completely. The error is always:

```
No database connection: connect() has not been called.
Fix: Run gbrain init --supabase or gbrain init --url <connection_string>
```

### What happens

1. Dream cycle starts, connects to Supabase pooler (port 5432)
2. Extract phase processes ~300 pages
3. At ~68% through, the batch flush hits a connection drop
4. `withRetry` (v0.41.18) correctly classifies the error as retryable via `isRetryableConnError`
5. Retry fires 3x with decorrelated jitter — **but the connection is never re-established**
6. All 3 retries call `getConnection()` which throws again because `sql` is null
7. 100-200 link/timeline rows lost per flush
8. `conversation_facts_backfill` phase opens a secondary engine handle that was never connected → immediate crash

### Root cause

The retry loop in `withRetry` (src/core/retry.ts) re-invokes the original function, which calls `getConnection()`. But `getConnection()` just reads the `sql` singleton from `db.ts`. When the postgres.js client's internal connection drops and `sql` gets nulled (line 216 in db.ts), no code path calls `connect()` again. The retry retries on a dead connection.

The `isRetryableConnError` check at line 227 correctly identifies the error. The problem is between lines 227 and the retry sleep — there's no reconnection step.

### Production data

From a 192K-page brain on Supabase (Supavisor transaction pooler, port 5432):

```
[extract.links_fs] 200/296 (67%)
[extract.links_fs] connection blip, retrying (attempt 1/3): No database connection: connect() has not been called
[extract.links_fs] connection blip, retrying (attempt 2/3): No database connection: connect() has not been called
[extract.links_fs] connection blip, retrying (attempt 3/3): No database connection: connect() has not been called
  batch error (100 link rows lost): No database connection: connect() has not been called
```

This repeats at the final flush too. Total: ~152 rows lost per dream cycle, and `conversation_facts_backfill` always crashes.

### Proposed fix

In `withRetry` (src/core/retry.ts), after classifying as retryable and before the retry sleep, check if the error is "No database connection" and call `connect()` to re-establish the singleton:

```typescript
if (!isRetryableConnError(err)) throw err;

// Reconnect on null-singleton errors before retrying
if (err?.problem === 'No database connection') {
  try {
[...]

PR #1578, opened 2026-05-27
fix(jobs work): switch worker engine to instance-owned pool (Issue #1413 worker-side)

jonahberg:
## Summary

Forces the queue-worker subprocess engine into instance-owned-pool mode at startup, mirroring the pattern @archerhpagent shipped in #1415 for the autopilot main loop, but applied at the separate worker code path (`bun gbrain jobs work --max-rss N`).

## The bug

When a transient Supabase pool eviction nulls the module-level `sql` singleton (the catch path at `src/core/db.ts:216` sets `sql = null` on connect failure), the next firing of `src/core/minions/worker.ts:619`'s `lockTimer` setInterval calls `queue.ts:renewLock` → `engine.executeRaw` → `db.ts:getConnection`, which throws:

```
GBrainError: No database connection: connect() has not been called. Fix: Run gbrain init --supabase or gbrain init --url <connection_string>
    at getConnection (src/core/db.ts:153:15)
    at executeRaw (src/core/postgres-engine.ts:4349:18)
    at renewLock (src/core/minions/queue.ts:1024:36)
    at <anonymous> (src/core/minions/worker.ts:619:40)
```

That either:
1. **Unhandled-reject-crashes the worker** — autopilot KeepAlive respawns. Observed 2x in 90s on a real production brain.
2. **Degrades the worker silently** — `queue.poll()` returns nothing, the cycle stops making progress, but the daemon process stays alive and `gbrain autopilot --status` reports `installed`. Observed 110+ minutes of zero ingest progression on the same brain before manual restart.

Both states observed on v0.41.18.0 against a 42k-page Supabase pooler brain (Autopilot run, `extract.links_fs` phase). Logs at https://github.com/garrytan/gbrain/issues/1413#issuecomment-4557640845 (my comment in the same incident thread).

## Relationship to #1415 and #1413

@archerhpagent's #1415 fixed the same module-singleton-null bug class for the autopilot main loop by forcing `engine.connect({...savedConfig, poolSize: 5})` at autopilot startup. That PR was closed 2026-05-26 on the claim that v0.41.11.x autopilot.ts rewrite eliminated the bug — which is true for the autopilot path, but the **queue worker subprocess is a separate process** with its own engine instance and its own `connectEngine()` call. Per-subprocess engine initialization defaults to the module-singleton fallback (`db.connect(config)` → module-level `sql` shared with the worker's `executeRaw` callers), so the worker inherits the same bug class.

This PR closes the bug at the worker startup site. Same pattern, different code path.

## The fix

```ts
if (engine.kind === 'postgres') {
  const eng = engine as unknown as {
    _savedConfig?: EngineConfig & { poolSize?: number };
    _connectionStyle?: 'instance' | 'module' | null;
  };
  if (eng._savedConfig && eng._connectionStyle !== 'instance') {
    try {
      await engine.connect({ ...eng._savedConfig, poolSize: 5 } as EngineConfig & { poolSize: number });
    } catch (e) {
      console.error(`[jobs work] could not switch engine to instance pool: ${(e as Error).message}`);
    }
  }
}
```

Inserted right before `new MinionWorker(engine, {...})` in `case 'work':` (around `jobs.ts:814`). 14 LOC of executable code + 17 LOC of explanatory comment.

**Properties:**
- Gate on `engine.kind === 'postgres'` — PGLite engines unaffected (they ignore `poolSize`).
- Read `_savedConfig` via narrow structural cast (mirrors PR #1415 style; no `as any` blast radius).
- Skip when `_connectionStyle === 'instance'` — idempotent on re-run, and a no-op when a future caller wires the engine with `poolSize` upstream of `case 'work':`.
- `try/catch` wraps the `connect()` call so a transient pool-init blip logs and continues rather than killing the worker startup. The error message names the daemon process the operator's eyes are already on.
- No new dependencies. No new public API.

## Verification

Confirmed locally on the same production brain that hit the bug:

- **Before patch:** Worker subprocess crashes within 90s of restart with the trace above. 2 unhandled-rejection crashes in the first run. Subsequent restarts also crash within minutes. Brain stays 100+ minutes behind real-time because each catch-up sync hits the bug mid-`extract.links_fs`.
- **After patch:** Autopilot log shows `[jobs work] switched engine to instance-owned pool (size=5)` once at startup. Worker survives `extract.links_fs` (42k pages) cleanly. New ingest entries land on schedule. (Monitor still running at PR-submit time — will update this PR with the final clean-runtime number when it lands.)

## Tests

The bug is hard to unit-test deterministically — reproducing it needs real Supabase + PgBouncer transaction-mode + connection pool eviction (or a brittle simulation of the same). I did NOT add a behavioral test for the new code path; the static guard `_connectionStyle !== 'instance' && _savedConfig defined → call engine.connect with poolSize` is exercised by the engine's existing `connect()` tests via the `poolSize` branch.

Happy to add a static-shape test (mirror of `test/autopilot-supervisor-wiring.test.ts`) that asserts the patch wires `engine.connect({..., poolSize: 5})` between `queue.ensureSchema()` and `new MinionWorker()`, if that'd help unblock review.

## Test plan

- [ ] Static-shape test for the patch (optional; see Tests section above)
- [x] Typecheck clean (`bun run typecheck` on the branch — no errors introduced)
- [x] Manual repro confirmation on the v0.41.18.0 → patched scenario

### DIFF
--- a/src/commands/jobs.ts
+++ b/src/commands/jobs.ts
@@ -4,6 +4,7 @@
  */
 
 import type { BrainEngine } from '../core/engine.ts';
+import type { EngineConfig } from '../core/types.ts';
 import { MinionQueue } from '../core/minions/queue.ts';
 import { MinionWorker } from '../core/minions/worker.ts';
 import type { MinionJob, MinionJobStatus } from '../core/minions/types.ts';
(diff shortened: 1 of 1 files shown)


jonahberg:
Closing as obsolete: upstream v0.41.28.0's retry self-heal (reconnect() wired into batchRetry, #1570) supersedes this worker-side patch — verified in production that connection blips now recover at attempt 1/3 with zero row loss, so the instance-pool workaround is no longer needed.

PR #1582, opened 2026-05-28
fix(lint): preserve shared Postgres connection during config lift

warkcod:
## Summary

Use an instance-owned Postgres pool for lint's temporary DB-backed config lift.

## Problem

`runLintCore()` can open a short-lived engine to read DB-backed content-sanity
config. On Postgres, that temporary engine was using the module-level singleton
connection and disconnecting it on cleanup.

In long-running workflows that share a `PostgresEngine` instance, this could
tear down the caller's connection mid-run.

Observed downstream symptom:
- `gbrain dream` reached later phases and then failed with
  `No database connection: connect() has not been called`

## Fix

- connect the temporary lint probe engine with `poolSize: 1`
- keep the connection instance-owned so cleanup cannot clobber the shared
  module-level singleton

## Tests

- add E2E regression coverage proving `runLintCore()` does not disconnect a
  shared `PostgresEngine`

### DIFF
--- a/src/commands/lint.ts
+++ b/src/commands/lint.ts
@@ -312,7 +312,10 @@ async function resolveLintContentSanity(): Promise<LintContentOpts['contentSanit
         database_path: base!.database_path,
       });
       try {
-        await engine.connect({});
+        // Use an instance-owned pool for this short-lived config probe so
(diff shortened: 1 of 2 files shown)


garrytan:

PR #1583, opened 2026-05-28
fix(gateway): accept explicit litellm embedding model ids

warkcod:
## Summary

Do not reject `litellm:<model>` embedding configs when the model id is already
explicitly present.

## Problem

Embedding preflight treated LiteLLM's empty allowlist as equivalent to
"no model specified", even when the configured embedding model already had a
concrete suffix like `litellm:bge-m3-emb`.

That caused embedding preflight to fail with the user-provided-model-unset
diagnosis even though the configuration was valid.

## Fix

- keep the user-provided-model guard
- only fire it when the parsed model id is actually empty
- allow explicit `provider:model` strings through

## Tests

- add regression coverage for `litellm:bge-m3-emb`


garrytan:

PR #1584, opened 2026-05-28
fix(sources): stop double-encoding source config on Postgres

warkcod:
## Summary

Write `sources.config` as real JSON on Postgres instead of
`JSON.stringify(...)+::jsonb`.

## Problem

Several source-management write paths persisted config through
`JSON.stringify(config)` into a `::jsonb` cast.

On Postgres this can store a JSON string shape instead of an object. Later
config merges then degrade into arrays, and object-path reads like
`config->>'last_full_cycle_at'` stop working.

This breaks doctor/autopilot freshness checks and corrupts source metadata.

## Fix

- add a safe Postgres config-write path using `sql.json(...)`
- use it in:
  - `sources` command config updates
  - `sources-ops` source creation paths
- preserve the existing PGLite behavior

## Tests

- extend Postgres parity coverage for `updateSourceConfig()`
- keep `sources-ops` coverage on add/create paths
- assert merged config stays a JSON object, not an array

### DIFF
--- a/src/commands/sources.ts
+++ b/src/commands/sources.ts
@@ -26,6 +26,7 @@
 import { writeFileSync, unlinkSync, existsSync } from 'fs';
 import { join } from 'path';
 import type { BrainEngine } from '../core/engine.ts';
+import { PostgresEngine } from '../core/postgres-engine.ts';
 import {
   assessDestructiveImpact,
   checkDestructiveConfirmation,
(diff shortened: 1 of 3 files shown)


time-attack:
Thanks for chasing this down — you correctly identified a real and nasty bug class (JSON string scalar stored where an object belongs, which then breaks `config->>'last_full_cycle_at'` reads). It's already fixed on master: the sweep in #2375 converted every `sources.config` write to the `$N::text::jsonb` binding (see `src/core/sources-ops.ts:437`/`:514` and the six `UPDATE sources SET config` sites in `src/commands/sources.ts`), and #3334 closed the re-wrapping follow-up. Both CI guards now block regressions, so I'm closing this as already fixed rather than adding a third spelling of the same fix.

PR #1596, opened 2026-05-28
fix(synthesize): promote subagent timeouts to config keys (#1594)

### LINKED ISSUE #1594: Dream synthesize: hardcoded 30-min subagent timeout silently drops transcripts
# Dream synthesize: hardcoded 30-min subagent timeout fails on dense session transcripts

## Summary

`gbrain dream --phase synthesize` consistently fails with `wall-clock timeout exceeded` on transcripts that exceed ~20k tokens, even when the subagent is making valid progress. The 30-minute per-chunk timeout is **hardcoded as a literal** in `src/core/cycle/synthesize.ts` (line 481) rather than being exposed as a config knob, so users cannot raise it without forking.

## Environment

- gbrain: v0.41.14.0 (Bun global install)
- Engine: Supabase (PostgreSQL 17.6, pgvector, transaction pooler @ port 6543)
- macOS arm64, Bun 1.3.14
- Embedding: `openai:text-embedding-3-small` (1536d)
- Subagent model: `anthropic:claude-sonnet-4-6`

## Reproduction

1. Configure `dream.synthesize.session_corpus_dir` to a directory of `.txt` transcripts, where each transcript is a dense Telegram/chat-style log (~20k–50k tokens).
2. Run `gbrain dream --dir <brain> --phase synthesize --json`.
3. Watch jobs queue as `subagent` rows; each gets the hardcoded `timeout_ms: 30 * 60 * 1000`.
4. Every job runs to wall-clock timeout, retries 3x, ends `dead`. Zero `wiki/personal/reflections/*` pages get written.

In my run, 6/6 worthwhile transcripts (per the verdict pass) hit `wall-clock timeout exceeded` on all 3 attempts.

## Relevant code

`src/core/cycle/synthesize.ts:481`
```ts
const submitOpts: Partial<MinionJobInput> = {
  max_stalled: 3,
  on_child_fail: 'continue',
  idempotency_key,
  timeout_ms: 30 * 60 * 1000, // 30 min per chunk    <-- hardcoded literal
};
```

`src/core/cycle/synthesize.ts:502`
```ts
const job = await waitForCompletion(queue, jobId, {
  timeoutMs: 35 * 60 * 1000,                          <-- also hardcoded
  pollMs: 5 * 1000,
});
```

Attempting `gbrain config set subagent.timeout_ms 600000` returns `[config] No similar known key`, confirming there is no published override path.

## Why 30 min is not enough for some users

The subagent is allowed `max_turns: 30`. With ~25k–50k input tokens per transcript:
- Sonnet spends most turns on `search` + `get_page` exploring existing brain content before calling `put_page` (the prompt mandates wikilink discovery first).
- Each tool round-trip can be 30–90 sec on Anthropic.
- A 25–30 turn read-heavy session can legitimately blow past 30 min wall-clock even when the subagent is making real progress.

The result is **silent data loss**: the verdict pass identifies worthwhile transcripts, but the brain never receives the synthesized pages.

## Proposed fix

Promote both literals to config keys with sane defaults:

```ts
const submitTimeoutMs =
  Number(await engine.getConfig('dream.synthesize.subagent_timeout_ms')) ||
  30 * 60 * 1000;

const waitTimeoutMs =
  Number(await engine.getConfig('dream.synthesize.subagent_wait_timeout_ms')) ||
  35 * 60 * 1000;
```

Register both keys in the config schema validator so `gbrain config set …` accepts them without `--force`.

Optional follow-ups (lower priority):
- Surface `dream.synthesize.max_turns` (currently fixed at 30 inline in the synthesize prompt builder).
- Emit a structured event when a synthesize subagent is **terminated by timeout while still making tool calls** so operators can distinguish "timed out but progressing" from "stuck/looping".

## Workaround

Disable the phase entirely:
```bash
gbrain config set dream.synthesize.enabled false
```
…which loses the whole feature.

## Will to contribute

Happy to PR this if you want — straightforward 10-line patch + a unit test that asserts the override is honored. Tag me if you'd prefer it shipped under a different config namespace.

ai920wisco:
Fixes #1594

Promotes two hardcoded subagent timeout literals in `src/core/cycle/synthesize.ts` to config keys (defaults preserve original behavior):

- `dream.synthesize.subagent_timeout_ms` (default 1800000 = 30 min)
- `dream.synthesize.subagent_wait_timeout_ms` (default 2100000 = 35 min)

Registered in config schema + loadConfig merge. Unit test added.

Verification: targeted test ✅, tsc --noEmit ✅, diff --check ✅

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to GBrain will be documented in this file.
 
+## [0.41.15.0] - 2026-05-28
+
+- `dream.synthesize.subagent_timeout_ms` and
+  `dream.synthesize.subagent_wait_timeout_ms` now control synthesize
+  subagent job and wait timeouts. Defaults remain 30 and 35 minutes.
+
 ## [0.41.14.0] - 2026-05-25
 
 **Your gbrain skills can declare their own routing triggers in their
--- a/src/core/config.ts
+++ b/src/core/config.ts
@@ -190,6 +190,8 @@ export interface GBrainConfig {
       verdict_model?: string;
       max_prompt_tokens?: number;
       max_chunks_per_transcript?: number;
+      subagent_timeout_ms?: number;
+      subagent_wait_timeout_ms?: number;
     };
     patterns?: {
       lookback_days?: number;
@@ -497,6 +499,12 @@ export async function loadConfigWithEngine(
     const n = parseInt(v, 10);
     return Number.isFinite(n) && n > 0 ? n : undefined;
   }
+  async function dbNum(key: string): Promise<number | undefined> {
+    const v = await dbStr(key);
+    if (v === undefined) return undefined;
+    const n = Number(v);
+    return Number.isNaN(n) ? undefined : n;
+  }
   const dbWarnBytes = await dbInt('content_sanity.bytes_warn');
   const dbBlockBytes = await dbInt('content_sanity.bytes_block');
   const dbJunkEnabled = await dbBool('content_sanity.junk_patterns_enabled');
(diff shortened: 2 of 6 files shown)


time-attack:
Superseded with credit by #2937 (just merged): your config-key promotion of the synthesize subagent timeouts was ported onto current master as part of the subagent-orchestration wave (dream.synthesize.subagent_timeout_ms / subagent_wait_timeout_ms), with your authorship preserved via Co-authored-by. Fixes #1594. Thank you!

PR #1598, opened 2026-05-28
(fix) typo with alphaclaw.com url (should be alphaclaw.md)

jordanschau:


jordanschau:
https://alphaclaw.com/ is a pet site 😆

garrytan:
Thanks for spotting the broken link! One catch: alphaclaw.com is the real site, so rewriting it to alphaclaw.md would not resolve as a URL. The correct fix is being handled in #1828. Closing this one to avoid the conflicting change — appreciated.

PR #1613, opened 2026-05-29
fix(conversation-parser): add bold-name-no-timestamp builtin (Circleback/Granola/Zoom)

garrytan-agents:
## Problem

gbrain's conversation parser ships 13 builtin patterns in `src/core/conversation-parser/builtins.ts` — **every one requires a per-line timestamp** (an inline date or a time anchor like `(HH:MM)` / `[HH:MM]`).

Modern meeting-transcription tools — **Circleback, Granola, Zoom** — emit transcripts as `**Speaker Name:** message text` with **no per-line timestamp**:

```
**Garry Tan:** Okay, start on. And then weirdly like zoom doesn't...
**Participant 2:** he tried to reset it remotely the other night. Let me ask him.
**Garry Tan:** I mean it's really just like we need to get zoom to fix this.
**Participant 2:** Okay, let me.
```

With no time anchor, this shape matches no builtin. A single accidental anchor on a long page scores ~`0.002`, well below the `SCORING_MIN_ACCEPTANCE` gate of `0.05` in `parse.ts`, so the page parses to **0 messages** and extracts **zero conversation-facts**.

### Real-world impact

In a production brain, the coverage scan reported `conversation_format_coverage: 100% _no_match` — **104/104 conversation pages** plus **3,423 eligible pages** matched no builtin pattern and silently yielded nothing.

## The fix

New **additive** builtin `bold-name-no-time`:

- regex `/^\*\*(.+?):\*\*\s*(.*)$/` — captures the speaker **inside** the bold markers (`**Name:**`), message text in group 2.
- No time capture. `date_source: 'frontmatter'` with `hour_group` undefined routes through `parse.ts`'s existing no-time branch (the same convention as `irc-classic`), anchoring every message at `00:00:00` of the page's frontmatter date. **No wall-clock time is fabricated**; intra-day ordering is preserved by line order.
- Declared **after** the timestamped bold patterns (`imessage-slack`, `telegram-bracket`, `bold-paren-time`) so it can never shadow them. The scorer tie-breaks on declaration index, and the regex requires the colon **inside** the bold markers, so the `**Name** (time):` paren-time shape (colon outside) still matches `bold-paren-time`, not this pattern.

No existing pattern is modified.

## Tests

Added a dedicated `bold-name-no-time` test block in `test/conversation-parser/parse.test.ts`:
- parses the real 4-line `**Speaker:** text` sample → 4 messages with correct speaker/text, anchored at `00:00:00` of the frontmatter date;
- scores above the `0.05` acceptance floor on a pure bold-name transcript (epoch-default date when none supplied);
- **regression**: `**Garry** (HH:MM): text` still matches `bold-paren-time` (not the new pattern);
- **regression**: a pure `telegram-bracket` transcript still matches `telegram-bracket` (wins the score tie via lower declaration index).

Updated the existing `disabledBuiltinIds` test: disabling `telegram-bracket` now correctly falls through to `bold-name-no-time` (the bracket line's colon is inside the bold markers); disabling both yields the genuine `no_match`.

Full suite result:

```
bun test test/conversation-parser/ test/conversation-parser-cli.test.ts test/extract-conversation-facts.test.ts
131 pass / 134
```

The 3 failures are **pre-existing on master** and env-driven (`probeLlmAvailability returns null when ANTHROPIC_API_KEY is unset` and two sibling "provider unavailable" fail-open tests) — they fail identically on clean `master` in this sandbox because an `ANTHROPIC_API_KEY` is present in the environment. They are unrelated to this change. `eval-conversation-parser-cli` and `extract-conversation-facts-workers` are fully green (32/32). All 14 builtins (13 existing + new) pass their module-load `validatePatternEntry` checks and their `test_positive` parse loop.

---

### Separately noticed (not in this PR)

`doctor --source <id>` does **not** scope the `orphan_ratio` check. In `src/commands/doctor.ts` (~line 4559), `getOrphansData(engine, { includePseudo: false })` takes no source filter, so `--source default` still reports a brain-wide orphan ratio across all federated sources. Suggested follow-up: thread `sourceId` into `getOrphansData`. (Note only — intentionally not fixed here.)

### DIFF
--- a/src/core/conversation-parser/builtins.ts
+++ b/src/core/conversation-parser/builtins.ts
@@ -1,7 +1,7 @@
 /**
  * v0.41.16.0 — Built-in conversation parser pattern registry.
  *
- * Twelve hand-vetted patterns covering the chat-export formats this
+ * Fourteen hand-vetted patterns covering the chat-export formats this
  * codebase is most likely to encounter. Each pattern's regex was
  * derived from a public format reference (source_doc field) so future
  * maintainers can verify against the wild shape.
@@ -50,7 +50,7 @@ export function cleanSpeaker(raw: string, override?: RegExp): string {
   return stripped || raw.trim();
 }
 
-/** The 12 hand-vetted built-in patterns. */
+/** The 14 hand-vetted built-in patterns. */
 export const BUILTIN_PATTERNS: readonly PatternEntry[] = [
   // -------------------------------------------------------------------
   // INLINE-DATE patterns (date in every line; less ambiguous; tried first).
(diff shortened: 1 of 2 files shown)


garrytan:
Superseded by #1620 (landing as v0.41.29.0). Thank you for this — the `bold-name-no-time` pattern is the real unlock for the 104 conversation pages + 3,423 eligible pages that were silently extracting nothing.

Re-homed into a base-repo branch because fork PRs from this account don't get CI secrets. Folded in with these changes during the review (`/plan-eng-review` + `/codex` outside-voice):

- **Privacy scrub** — `test_positive` samples used real names (`Garry Tan`, `Alex Graveley`), which the project privacy rule forbids in checked-in code; replaced with placeholders. Also scrubbed the same names that were already in `bold-paren-time`.
- **`score_full_body` guard** — the regex matches any `**Label:** text` (a common prose idiom). Codex verified the 0.05 floor does NOT protect a notes page with bold labels clustered in its first 10 lines (head score 0.3 skips the rescore). Added a full-body acceptance recompute so such a page stays `no_match` instead of mis-parsing as a conversation.
- **`(?!\[)` lookahead** — so disabling telegram-bracket yields an honest `no_match` instead of capturing `speaker="[18:37] Name"`. The `disabledBuiltinIds` test stays at `no_match`.
- Added an eval fixture + clustered-head adversarial regression.

Your original `bold-name-no-time` pattern is preserved with `Co-Authored-By` on the parser commit. Bundled an unrelated `orphan_ratio --source` scoping fix into the same release. Closing in favor of #1620.

PR #1616, opened 2026-05-29
docs: retrieval incident — vector search returns pages by their weakest chunk (RFC + fix menu + eval design)

garrytan-agents:
**Docs-only PR (no code).** Full incident writeup + boil-the-ocean fix menu + eval design for a brain-wide retrieval defect.

## TL;DR
Querying **"Greek amphitheater"** failed to confidently surface the page titled *"The Mingtang (明堂) — Indoor **Greek amphitheater**…"*, returning it at cosine **0.64** via a weak body chunk. The same page's **title chunk scores 0.9866** on a richer query — the embedding is fine, the retrieval aggregation is broken.

**Root cause:** `searchVector` (both postgres + pglite engines) returns top-k *chunks* with **no per-page max-pool**. A page is represented by whichever chunk survives the candidate cut, not its best chunk. The keyword path (`searchKeyword`) already does the correct `DISTINCT ON (slug) ORDER BY score DESC`. The vector path does not.

**Secondary:** `reranker_enabled` / `expansion` are a **no-op** on this query across all modes — suggesting the CLI default does not exercise the documented hybrid stack (RETRIEVAL.md claims 49.1 P@5 / 97.9 R@5). Doc/reality gap flagged.

## In the doc
- Incident narrative + reproduction with exact CLI output
- Code-level root cause with line refs (searchVector vs searchKeyword asymmetry, innerLimit, pglite parity)
- Boil-the-ocean fix menu: Tier 1 (per-page max-pool — the actual bug, ~1 day) → Tier 6 (graph-assisted named-entity recall)
- Eval design ("NamedThingBench") on the existing eval/code-retrieval harness + baseline→compare→gate: title-substring, generic-label→named-entity, alias/synonym, multi-chunk dilution, short-vs-rich pairs, hard negatives; Hit@1/3, MRR, nDCG@5, score calibration vs the 0.85 dedup threshold; CI gate + engine parity

## Recommendation
Ship Tier 1, **measure**, then decide Tiers 2–6 by data. No code in this PR by request — review the plan before implementation.

### DIFF
--- a/docs/architecture/RETRIEVAL_MAXPOOL_INCIDENT.md
+++ b/docs/architecture/RETRIEVAL_MAXPOOL_INCIDENT.md
@@ -1,0 +1,259 @@
+# Retrieval Quality Incident: Vector Search Returns Pages by Their Weakest Chunk
+
+**Status:** Proposal / RFC (no code in this PR — documentation + remediation plan + eval design)
+**Author:** Wintermute (Garry Tan's agent)
+**Date:** 2026-05-29
+**gbrain version at time of incident:** 0.41.26.1
+**Severity:** High — brain-wide retrieval quality defect affecting every multi-chunk page
+**Related:** [`docs/architecture/RETRIEVAL.md`](./RETRIEVAL.md), [`docs/eval/METRIC_GLOSSARY.md`](../eval/METRIC_GLOSSARY.md), [`docs/eval/SEARCH_MODE_METHODOLOGY.md`](../eval/SEARCH_MODE_METHODOLOGY.md)
+
+---
+
+## 0. TL;DR
+
+A real query missed a real page. Querying **"Greek amphitheater"** failed to surface the page literally titled *"The Mingtang (明堂) — Indoor **Greek amphitheater**…"* with enough confidence to be used, returning it at cosine **0.64** via a mediocre body chunk. The same page's **title chunk scores 0.9866** on a richer query — so the embedding is fine; the *retrieval aggregation* is broken.
+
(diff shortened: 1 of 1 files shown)


garrytan:
Superseded by the retrieval-cathedral wave (branch garrytan/vector-search-max-pool).

The docs-only diagnosis here is folded in and **corrected** against verified ground truth from the code:
- `gbrain search` is keyword-only (ts_rank); the 0.64/0.98 cosine scores came from the hybrid `query`/MCP path Wintermute actually hit — the repro command was mislabeled.
- `--mode` is not a param of either op; mode resolves server-side from the `search.mode` config key, which is why all three 'modes' returned identical results (the flag is silently dropped; `thorough` isn't a real mode). Rerank/expansion also fail-open to no-ops when no provider is configured.
- hybridSearch ALREADY max-pools per page at the dedup layer (top-3 → cap-2, score-sorted), so the per-page max-pool fix's real win is candidate-set page recall, not final-rank — and it is necessary-but-not-sufficient (the title chunk may simply score below the body chunk on a 2-word query, or fall outside the candidate pool).

The replacement wave does the complete cure, not the proximate patch: a Phase-0 stage-by-stage diagnostic, per-page max-pool (both engines, shared SQL builder), page.title as a first-class boost, a free-text alias-resolution layer (new page_aliases table), an evidence-based confidence contract (evidence enum + create_safety, so the agent's don't-duplicate decision keys off WHY a page matched, not a blended score), `gbrain search`→hybrid with a remote-provider policy + escape hatch, and NamedThingBench (a hard CI gate). The narrative here will be carried into a corrected docs/architecture/RETRIEVAL_MAXPOOL_INCIDENT.md with Co-Authored-By preserved.

Thank you for the writeup — it was the right call to file it; closing only because the fix supersedes the doc-only RFC.

PR #1622, opened 2026-05-29
fix(postgres): don't close a module singleton an engine only joined (deeper #1570 root cause)

pabloglzg:
## What

The deeper root cause behind #1570 / #1608. A module-style `PostgresEngine` that **joined** an already-open `db.ts` singleton calls `db.disconnect()` (→ `sql.end()`) on its own `disconnect()`, ending the **shared** connection for the whole process.

#1608 added a `withRetry` self-heal (reconnect between attempts) + a disconnect audit, and its message notes: *"The deeper bug (who's calling disconnect mid-cycle) is left [as follow-up]."* This PR fixes that deeper bug and names the culprit.

## The culprit

`gbrain dream` on Postgres: the **lint** phase's content-sanity probe `resolveLintContentSanity` (`src/commands/lint.ts:315`/`319`) opens a module-style engine to lift `content_sanity` config from the DB, then `disconnect()`s it. That engine merely joined the cycle's already-open singleton (its `db.connect()` was a no-op), so the disconnect ends the shared connection. Every later DB phase (extract write, embed, conversation_facts_backfill) then throws `connect() has not been called` → the cycle exits 1 and silently loses link/timeline writes. `--phase lint` alone "works" because nothing runs after it.

## The fix

Ownership tracking — same pattern as the v0.28.1 jobs-path idempotency fix:
- `db.ts`: new `isConnected()`.
- `postgres-engine.ts`: new `_ownsModuleSingleton`, set to `!db.isConnected()` **before** `db.connect()`. Module-style `disconnect()` only calls `db.disconnect()` when it owns the singleton; a joiner leaves it for its opener to close.

Complements #1608 (prevents the clobber instead of recovering from it).

## Verification

- Full `gbrain dream --source default` exits 0 (was exit 1); extract/embed writes land.
- `test/e2e/postgres-engine-disconnect-idempotency.test.ts`: 3/3 pass incl. a new regression case (`module engine that JOINED an open singleton does NOT clobber it`).
- Typecheck clean.

### DIFF
--- a/src/core/db.ts
+++ b/src/core/db.ts
@@ -159,6 +159,16 @@ export function getConnection(): ReturnType<typeof postgres> {
   return sql;
 }
 
+/**
+ * True when the module-level singleton connection is currently open.
+ * Lets a PostgresEngine that joins an already-open singleton (e.g. a config
+ * probe created inside the dream cycle) know it does NOT own the connection,
+ * so its disconnect() won't end the shared pool out from under the owner.
+ */
+export function isConnected(): boolean {
+  return sql !== null;
+}
+
 export async function connect(config: EngineConfig): Promise<void> {
   if (sql) {
     // Warn if a different URL is passed — the old connection is still in use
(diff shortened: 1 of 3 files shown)


garrytan:

PR #1623, opened 2026-05-29
fix(staleness): measure sync lag relative to newest committed content (not wall-clock)

garrytan-agents:
## Problem

Source staleness in `gbrain doctor` (and `sources status`) was measured as raw wall-clock time since the last sync:

```
lag_seconds = now - last_sync_at
```

This flags **quiet, fully-caught-up repos as severely stale** even when nothing new has been committed since the last sync. A federated source that hasn't received a commit in days is *not* stale for search purposes — the sync has everything the repo contains. But the wall-clock metric kept escalating, producing false `SEVERELY STALE` alerts.

### Error Log (anonymized)

A daily health check repeatedly fired against a low-churn source:

```
GBrain Doctor — Score: 0/100
Source Health:
  - source-x — SEVERELY STALE (~86h since last sync, last seen <date> 21:14 UTC)
Threshold exceeded: >72h
```

Yet the underlying repo was caught up:

```
last_sync_at:  <date+1> 21:14 UTC
HEAD commit:   <date> 17:48 UTC   (older than last_sync_at → nothing new to sync)
working tree:  only untracked dirs (?? companies/, ?? media/) — never committed
```

The newest **committed** content predated the last sync by ~28h, so the source was fully synced. The alert was pure wall-clock noise.

## Root Cause

Three independent consumers all derived staleness from `now - last_sync_at`:

1. `buildSyncStatusReport` (sync.ts) → `staleness_hours` / dashboard
2. `computeAllSourceMetrics` + `isSourceStale` (source-health.ts) → `lag_seconds`, federation health
3. `checkSyncFreshness` + `checkCycleFreshness` (doctor.ts)

None compared against what the repo actually contains.

## What We Tried

- **Considered raising the threshold** — rejected. A higher threshold just delays the false positive; it doesn't fix the semantics. A genuinely stale active repo would also be masked.
- **Considered counting untracked files as "content"** — rejected after observing that untracked dirs (`?? companies/`, `?? media/`) inflated the lag. Untracked files are not part of the repo; "last committed to the repo" must mean committed/tracked state only.
- **Final approach** — make lag *content-relative*: 0 when the newest committed/tracked content is at or before `last_sync_at`, else the wall-clock delta. Fall back to wall-clock when content can't be probed (non-git / unreadable path) so detection never regresses.

## Solution

New helpers in `src/core/source-health.ts`:

- `newestContentMs(localPath)` — newest *committed/tracked* mtime: `git log -1 --format=%ct` (HEAD commit time) combined with the newest tracked working-tree modification (`git status --porcelain -z`, **excluding** untracked `??` entries). Returns `null` for non-git/unreadable paths.
- `contentRelativeLagSeconds(localPath, lastSyncMs, nowMs)`:
  - `null` when `last_sync_at` is unknown.
  - **Negative** wall-clock (future `last_sync_at`) is surfaced as-is so clock-skew detection upstream still fires.
  - When content can be probed: `0` if newest content ≤ last sync (caught up), else wall-clock delta.
  - When content can't be probed: wall-clock fallback.

Wired into all three consumers. In `checkSyncFreshness`, it composes cleanly with the v0.41.27.0 `localOnly` git short-circuit — the short-circuit is a strict "definitely unchanged" gate (HEAD === last_commit + clean tree + chunker match) for the local CLI; content-relative lag covers the general/remote path where the short-circuit doesn't fire. `cycle_freshness` is also made content-relative (a source with no new committed content doesn't need re-cycling).

### Behavior Matrix

| Scenario | Newest committed content | Old metric | New metric |
|---|---|---|---|
| Quiet repo, caught up | ≤ last_sync | grows forever → stale | **0 (ok)** |
| Active repo, behind | > last_sync | wall-clock | wall-clock (stale) |
| Non-git / unreadable path | n/a | wall-clock | wall-clock (unchanged) |
| Future last_sync (clock skew) | any | negative → skew warn | negative → skew warn (unchanged) |

## Results

Re-running `sources status --json` against a real low-churn source after the fix:

```
before:  lag ≈ 87.6h  → SEVERELY STALE (false positive)
after:   lag = 0.0h   → ok (newest commit predates last sync = caught up)
```

Other active sources continue to report correct non-zero lag.

## Testing

- `tsc --noEmit`: clean.
- New regression tests in `test/source-health.test.ts` (quiet-repo → 0, behind-repo → wall-clock, non-git fallback, tracked-edit counts, untracked excluded) and `test/sync-all-parallel.test.ts` (caught-up → 0 staleness).
- Full local run of `source-health`, `sync-all-parallel`, `doctor-cycle-freshness`, `doctor`: **115 pass / 0 fail**.

### DIFF
--- a/src/commands/doctor.ts
+++ b/src/commands/doctor.ts
@@ -2750,6 +2750,16 @@ export async function checkSyncFreshness(
     // a TEXT column storing String(CHUNKER_VERSION).
     const currentChunkerVersion = String(CHUNKER_VERSION);
 
+    // Content-relative staleness: age is the gap to the source's newest
+    // committed/tracked content ("what was last committed to the repo"), not
+    // wall-clock since the last sync. A quiet repo whose newest commit
+    // predates its last sync is caught up and must NOT be flagged stale.
+    // Falls back to wall-clock when content can't be probed (non-git path).
+    // Complements the localOnly git short-circuit below: the short-circuit is
+    // a strict "definitely unchanged" gate (HEAD===last_commit + clean tree +
+    // chunker match) for the local CLI; content-relative lag covers the
+    // general/remote path where the short-circuit does not fire.
+    const { contentRelativeLagSeconds } = await import('../core/source-health.ts');
     const issues: string[] = [];
     // v0.41.27.0: D6 three-bucket count math. Every source falls into
     // EXACTLY ONE bucket per iteration. Invariant pinned by unit test:
@@ -2777,7 +2787,8 @@ export async function checkSyncFreshness(
       }
 
       const lastSync = new Date(source.last_sync_at).getTime();
-      const ageMs = now - lastSync;
+      const lagSeconds = contentRelativeLagSeconds(source.local_path, lastSync, now);
+      const ageMs = lagSeconds === null ? now - lastSync : lagSeconds * 1000;
 
       if (ageMs < 0) {
         issues.push(
@@ -2968,6 +2979,11 @@ export async function checkCycleFreshness(
     const failMs = failHours * 60 * 60 * 1000;
     const now = opts?.nowMs ?? Date.now();
 
+    // Content-relative: a source whose newest committed content predates its
+    // last full cycle is caught up — a quiet repo doesn't need re-cycling and
+    // must not be flagged. Falls back to wall-clock when content can't be
+    // probed (non-git path).
+    const { contentRelativeLagSeconds } = await import('../core/source-health.ts');
     const issues: string[] = [];
     let hasWarnings = false;
     let hasFailures = false;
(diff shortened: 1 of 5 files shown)


garrytan:
Superseded by #1656, re-implemented in the base repo (garrytan-agents PRs run in a fork without secret access, so CI can't gate them — base-repo branch fixes that).

Two substantive changes vs this PR:
1. **Correctness:** staleness keys off the commit HASH (`HEAD == last_commit`, untracked ignored), not a content timestamp. The timestamp comparison here false-reports "caught up" when HEAD moves to an older-*dated* commit (rebase preserving dates, branch rewind, imported old commit). #1656 also drops the fragile `git status --porcelain` mtime parse.
2. **Trust boundary:** this PR wired a live `git` subprocess into three remote-reachable consumers (doctorReportRemote's checkSyncFreshness + federation_health, and the get_status_snapshot MCP op), re-opening the v0.41.27.0 boundary (no subprocess on a DB-supplied local_path from remote callers). #1656 keeps local on live git and routes remote through a durable `sources.newest_content_at` column written at sync time — so the boundary stays intact AND the remote false-positive is still fixed.

Your root-cause analysis (untracked dirs defeating the clean-tree gate) was exactly right and is the headline fix. Thanks for it.

Issue #1061, opened 2026-05-16
mcp__gbrain__whoami throws unknown_transport on stdio MCP calls (v0.33 → v0.35.1.0)


vincedk-alt:
## Summary

`mcp__gbrain__whoami` (the `whoami` MCP tool exposed by the gbrain MCP server) consistently throws `unknown_transport` when invoked via the Claude Code MCP client on a local stdio-bound gbrain serve process. The error message claims to be a transport-layer guard:

```
{
  "error": "unknown_transport",
  "message": "whoami called over a remote transport that did not thread ctx.auth. This is a transport bug — every remote call site must populate ctx.auth or set ctx.remote === false."
}
```

The error message labels itself as a transport bug ("every remote call site must populate ctx.auth or set ctx.remote === false") but the bug has persisted across v0.33 → v0.35.0.0 → v0.35.1.0 without resolution.

## Repro on v0.35.1.0 against a healthy brain

```
# Brain is healthy and recently rebuilt
$ gbrain doctor | grep -E "Health|embedding_provider"
  [OK] embedding_provider: lmstudio:text-embedding-nomic-embed-text-v1.5 ✓ 158ms, 768 dims, DB aligned
Health score: 95/100.

# Other MCP ops work fine
$ # via Claude Code MCP: mcp__gbrain__sources_list → returns 2 sources
$ # via Claude Code MCP: mcp__gbrain__get_stats → returns page_count, chunk_count, ...

# But whoami throws:
$ # via Claude Code MCP: mcp__gbrain__whoami → unknown_transport
```

## Root cause hypothesis

Per the error message itself: the `whoami` op handler asserts `ctx.auth || ctx.remote === false` and throws `unknown_transport` when both are falsy. The Claude Code stdio MCP client connects via the standard `gbrain serve` stdio transport, which historically sets `ctx.remote = false` (trusted local CLI). One of two things must have regressed:

A. The stdio dispatch path in `src/mcp/server.ts` stopped setting `ctx.remote = false` somewhere in the v0.33 → v0.34 wave.
B. The `whoami` op handler hardened to require `ctx.auth` AND `ctx.remote === false` (logical AND), where it previously accepted either alone, and the stdio path now fails the new test.

Either way, the result is that local-CLI consumers (which is most MCP users) can't call `whoami` at all.

## Suggested fix

[...]

Issue #713, opened 2026-05-07
whoami returns unknown_transport over stdio MCP transport (v0.28.6)


lucha0404:
**Repro**:

1. `gbrain serve` (stdio MCP server, no flags)
2. From MCP client (Claude Code), call `mcp__gbrain__whoami`

**Expected**: `{transport: "local", scopes: []}`

**Actual**:
```
{
  "error": "unknown_transport",
  "message": "whoami called over a remote transport that did not thread ctx.auth. This is a transport bug — every remote call site must populate ctx.auth or set ctx.remote === false."
}
```

**Other tools work fine** over the same stdio transport — `find_orphans`, `takes_list`, `sources_list`, `get_stats`, `list_pages` all return correct responses. Only `whoami` mis-detects the transport.

**Likely cause** (from quick grep on `src/mcp/server.ts`):
- Line 36 sets `remote: true` (HTTP server context)
- Line 58 sets `remote: false` (stdio context)

The `whoami` handler probably checks `ctx.remote === true` first and bails into the unknown_transport branch before reaching the local-shape return — OR the stdio entry point doesn't actually reach line 58 in some code path.

**Environment**: gbrain v0.28.6 (commit b325f28) / Bun 1.3.13 / Linux (Ubuntu 22.04, Vertex AI ADC fork) + Windows 11 client. Both clients exhibit the same error over their respective stdio MCP server.

Happy to PR a fix if you confirm the diagnosis direction.

Hugo23WF:
**Confirming repro on v0.30.0** — same `unknown_transport` over stdio MCP, with the diagnosis in OP still accurate. Adding context that may help disambiguate the fix.

## v0.30.0 codebase observation

In current `src/mcp/server.ts:30-42`, the stdio dispatch deliberately sets `remote: true` (not `false` as OP's grep on v0.28.6 suggested):

```ts
return dispatchToolCall(engine, name, params, {
  remote: true,
  takesHoldersAllowList: ['world'],
});
```

Per the in-line comment, this is **intentional** — it limits what `takes_list`, `takes_search`, and `query` expose to a stdio agent caller (private hunches stay private; only `world`-holder takes leak through). Flipping stdio to `remote: false` would regress the v0.28 takes-protection contract.

## Why only `whoami` mis-detects

[...]

PR #1625, opened 2026-05-29
fix(whoami): return a defined identity for stdio MCP instead of throwing unknown_transport

bluk1020:
## Problem

Calling `whoami` over the **stdio MCP transport** throws `unknown_transport`:

```
{"error":"unknown_transport","message":"whoami called over a remote transport that did not thread ctx.auth. ..."}
```

stdio is the canonical transport for the MCP server (`gbrain serve`), so the primary transport can't introspect its own identity — every `whoami` from Claude Desktop / Claude Code / any stdio client fails.

## Root cause

`ctx.remote` is the sole authority gate, and stdio is deliberately `remote: true` with no `ctx.auth` (so it stays visibility-filtered to `world`). But `whoami` only had identity shapes for:

- `local` — `remote === false` (trusted CLI)
- `oauth` / `legacy` — `ctx.auth` present (HTTP)
- else → throw `unknown_transport`

stdio matches none of these (`remote: true`, no auth), so it falls through to the throw that's meant to catch an **HTTP** call that lost its `ctx.auth`.

## Fix

Add a transport-**identity** discriminator `OperationContext.transport: 'stdio' | 'http'`, threaded from each MCP dispatch call site. `whoami` returns a defined, zero-authority shape for the stdio pipe:

```json
{ "transport": "stdio", "scopes": [] }
```

The `unknown_transport` throw is now reserved for the genuine bug case: a remote **HTTP** call that reached dispatch without `ctx.auth`.

## Why this is safe (trust boundary unchanged)

- **`remote` is untouched** and remains the sole access gate. stdio is still `remote: true` → still visibility-filtered to `world`, still scope-enforced, still gets the provenance fence.
- **`transport` is never read for an access decision** — its only consumer is `whoami`. It's a pure identity discriminator, documented as NOT a trust signal.
- **Unspoofable from client input** — `transport` is hardcoded at each server-side dispatch call site (`opts.transport`), never derived from request params/args. Added a test that sends `{"transport":"stdio"}` as a `whoami` *argument* and asserts it does **not** yield the stdio shape.
- **stdio returns empty scopes** — no fabricated authority; a caller gating on `scopes.includes('admin')` gets nothing.
- The stdio branch runs *before* the auth check, so even a stray leaked `auth` blob can't promote stdio to oauth/legacy (defense-in-depth test included).

## Testing

- `test/whoami.test.ts`: +3 tests — stdio shape without auth, stale-auth defense-in-depth, and no-spoof-via-client-param. 12/12 pass.
- Existing `unknown_transport` throw tests still pass (HTTP-without-auth path unchanged).
- Full integration suite (`test/integrations.test.ts`) 110/110; `tsc --noEmit` clean.
- Verified end-to-end by driving a fresh `gbrain serve` stdio process: `tools/call whoami` → `{"transport":"stdio","scopes":[]}` (no throw).

## Notes

- Did **not** bump the version / `package.json` — left for your release process.
- HTTP MCP transports (`serve-http.ts`, `http-transport.ts`) now also pass `transport: 'http'` so the discriminator is reliably populated at every call site (mirrors how `remote` is set on every transport).

### DIFF
--- a/src/commands/serve-http.ts
+++ b/src/commands/serve-http.ts
@@ -1535,6 +1535,7 @@ export async function runServeHttp(engine: BrainEngine, options: ServeHttpOption
       try {
         toolResult = await dispatchToolCall(engine, name, params as Record<string, unknown> | undefined, {
           remote: true,
+          transport: 'http',
           takesHoldersAllowList: tokenAllowList,
           sourceId: tokenSourceId,
           metaHook: getBrainHotMemoryMeta,
--- a/src/core/operations.ts
+++ b/src/core/operations.ts
@@ -296,6 +296,21 @@ export interface OperationContext {
    */
   remote: boolean;
   /**
+   * Which MCP transport carried this call. Set explicitly by the two
+   * remote/untrusted MCP transports — `'stdio'` by the local stdio server
+   * (src/mcp/server.ts), `'http'` by the HTTP MCP transports. Left undefined
+   * for the local CLI path (`gbrain call`, remote=false), which has no MCP
+   * transport.
+   *
+   * This is a transport-IDENTITY discriminator only — it is NOT a trust
+   * signal and never widens access. `remote` remains the sole authority
+   * gate; stdio MCP is still `remote: true` and sees only `visibility='world'`.
+   * The single consumer today is `whoami`, which needs to tell stdio (a
+   * known, auth-less local pipe) apart from an HTTP call that lost its
+   * `ctx.auth` (a genuine threading bug worth throwing on).
+   */
+  transport?: 'stdio' | 'http';
+  /**
    * Subagent runtime context (v0.16+). Set by the subagent tool dispatcher when
    * dispatching an op as a tool call from an LLM loop. Used to enforce per-op
    * agent policy (e.g. put_page namespace rule).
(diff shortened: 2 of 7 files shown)


bluk1020:
## Independent agent review — ✅ SHIP

This PR was reviewed by **Hermes** (persona "Donna"), an independent review agent running on a different model + toolchain from the author. It read the diff directly, searched the codebase, and ran the tests itself. Verdict: **SHIP**, no findings.

- **Trust boundary intact** — `ctx.transport` is consumed *only* by `whoami`; it is not read by any authorization, visibility-filtering, file/source-access, mutation, or scope check. `ctx.remote === false` remains the sole trusted-local path.
- **Not spoofable** — `transport` is threaded from server-side `DispatchOpts`, never from client params. The added test confirms `{transport:'stdio'}` passed as *user args* over HTTP still fails closed.
- **Stdio still visibility-scoped** — remains `remote: true` with `takesHoldersAllowList: ['world']`; the new field doesn't touch source/takes filtering.
- **HTTP-without-auth** — still throws `unknown_transport` unless the server-side call site explicitly marks stdio; authenticated HTTP still resolves to OAuth/legacy identity.
- **Verification run:** `bun test test/whoami.test.ts` → 12 pass, 0 fail.

**One caveat it raised** (not a blocker): coverage was unit/dispatcher-level, not a full stdio MCP *process* e2e. That is now addressed in `test/e2e/whoami-stdio-mcp.test.ts` (commit 46ab65d) — it spawns a real `gbrain serve` process and drives it through the MCP SDK stdio client, asserting `whoami` → `{transport:'stdio', scopes:[]}`. Hermetic (PGLite temp brain, no DATABASE_URL, no API keys), ~4–6s, stable across runs.

daragao3:
Confirming this is still needed at current `master` (HEAD `9bf96db8`, **v0.42.51.0**) — verified by code inspection, not just an older tag:

- `src/mcp/server.ts` still dispatches stdio with `remote: true` and no `auth`: `dispatchToolCall(engine, name, params, { remote: true, takesHoldersAllowList: ['world'], … })`.
- `buildOperationContext` (`src/mcp/dispatch.ts`) defaults `remote: opts.remote ?? true` and only sets `auth: opts.auth`, so over stdio `ctx.remote === true` and `ctx.auth === undefined`.
- The `whoami` handler (`src/core/operations.ts`) returns the `local` shape **only** for `ctx.remote === false`, then throws `unknown_transport` on `!ctx.auth` — so the stdio pipe still falls into the throw meant for HTTP-without-auth.

Repro chain is now v0.28.6 → v0.30.0 → v0.28.12 → v0.35.1.0 → v0.42.51.0; the referenced issues #713 and #1061 are both still open.

Independent corroboration on the approach: I hit this on v0.28.6 and, before finding this PR, implemented the identical seam — an `OperationContext.transport: 'stdio' | 'http'` tag threaded server-side from each dispatch call site, `whoami` returning the zero-authority local shape for stdio while `remote` stays `true` (takes/visibility/filesystem boundaries untouched) and HTTP-without-auth still fails closed. Two people landing on the same `transport`-discriminator design separately is a decent signal it's the right cut. The `test/e2e/whoami-stdio-mcp.test.ts` real-process test here is the piece my local version lacked.

+1 to merging (or folding into a fix-wave).

PR #1628, opened 2026-05-29
fix(code-def): include method/constructor/field/struct/protocol definitions

rayers:
## Problem

`gbrain code-def <symbol>` returns **0 results for methods, constructors, fields, C structs, and Swift protocols** — i.e. most of the symbols in an object-oriented codebase. `code-def <class>` and `code-def <function>` work fine. So on a Java / C++ / Swift project, `code-def` silently can't find a method even though it's fully indexed and `code-refs` finds it.

## Root cause

`content_chunks.symbol_type` is produced by `normalizeSymbolType()` in `src/core/chunkers/code.ts`, which canonicalizes only a subset of tree-sitter node types and lets the rest fall through:

```js
return type.replace(/_/g, ' ');
```

So node types land as space-normalized strings: `method_declaration` → `'method declaration'`, `method_definition` → `'method definition'`, `constructor_declaration` → `'constructor declaration'`, `field_definition` → `'field definition'`, `struct_specifier` → `'struct specifier'`, `protocol_declaration` → `'protocol declaration'`.

`code-def`'s `DEF_TYPES` filter in `src/commands/code-def.ts` only lists the canonical names, so `WHERE cc.symbol_type IN (...)` excludes every one of those. (The plain `'struct'` entry already in `DEF_TYPES` never matched anything either — C structs are stored as `'struct specifier'`.)

## Fix

Add the fallthrough definition forms to `DEF_TYPES`. Read-path only — no schema change, no reindex; it surfaces symbols already present in `content_chunks`.

## Verification

On a brain with an indexed Java codebase, `code-def <methodName>` went 0 → N, returning real `[java] path/File.java:line (method declaration)` definition sites. Class/function lookups unchanged. `bun test test/code-def-refs.test.ts test/code-edges.test.ts` → 18 pass / 0 fail; `tsc --noEmit` clean.

Follow-up (not in this PR): qualified queries like `code-def Class.method` still miss because `symbol_name` stores the short name; matching the trailing dotted segment would close that.


rayers:
Nudge on this — it's tiny (+10/-0, additive, no change to how `class`/`function` resolve today). Without it, `code-def` silently returns nothing for methods, constructors, fields, C structs, and Swift protocols, even though `code-refs` finds them fine — so on any OO codebase (Java/C++/Swift) most symbols are invisible to `code-def`. The fix just adds the tree-sitter node types `normalizeSymbolType()` currently lets fall through.

Been running on my fork for a couple weeks. Let me know if you'd like anything changed to land it.

PR #1630, opened 2026-05-29
fix(cost): embedding cost preview uses configured model rate, not hardcoded OpenAI

garrytan-agents:
## Problem

`gbrain sync --all` priced its embedding-cost gate using a hardcoded OpenAI rate, regardless of the actually-configured embedding model.

- `estimateEmbeddingCostUsd` computed spend from `EMBEDDING_COST_PER_1K_TOKENS = 0.00013` (OpenAI `text-embedding-3-large`).
- `sync.ts` labeled the preview with the back-compat `EMBEDDING_MODEL` constant (`text-embedding-3-large`) in the human message, the `--dry-run` JSON, and the `ConfirmationRequired` envelope JSON.

For a brain configured with a cheaper model (e.g. `zeroentropyai:zembed-1` @ $0.05/MTok), the gate named the **wrong provider** and **over-stated spend ~2.6x**: a 2.6B-token preview reported **$336.79 (text-embedding-3-large)** when the configured model's true cost is **~$129.54**.

## Root cause

A missed migration, not a config error.

- The estimator was introduced in **#422** (v0.21.0) with OpenAI hardcoded — correct at the time, since 3-large was the only supported embedding model.
- **#898** (v0.32.7) added a proper per-model pricing table (`src/core/embedding-pricing.ts`) with a `lookupEmbeddingPrice()` resolver, and migrated its consumers (`budget-tracker`, `post-upgrade-reembed`, `retrieval-upgrade-planner`, `brain-score-recommendations`) to it.
- The `sync --all` cost estimator was the **one consumer never migrated**, so it kept the original hardcode. Classic "five of six call sites updated."

Note: this only affected the **cost preview**. Actual embedding always used the configured model — the estimator just mispriced and mislabeled it.

## Fix

- `estimateEmbeddingCostUsd` now resolves the live configured model via the gateway and prices it through `embedding-pricing.ts` — the same table the other consumers use.
- New helper `currentEmbeddingPricePerMTok()` returns the configured model's rate, falling back to the OpenAI `text-embedding-3-large` rate **only** when the gateway is unconfigured (e.g. unit-test context) or the model is unknown to the pricing table.
- `sync.ts` surfaces the real model name (`getEmbeddingModelName()`) in the preview message, `--dry-run` JSON, and the envelope JSON.
- `EMBEDDING_COST_PER_1K_TOKENS` is retained for back-compat imports.

## Tests

- Rewrote `test/sync-cost-preview.test.ts`:
  - Clarified the back-compat constant test (it asserts the OpenAI rate, which is now only the fallback).
  - Documented that the linear-scaling test exercises the gateway-unconfigured fallback path.
  - **Added a regression test** asserting model-aware pricing: `openai:text-embedding-3-large` and `zeroentropyai:zembed-1` must produce materially different previews on the same token count. A fix that re-collapses both to the OpenAI number fails the assertion.
- `bunx tsc --noEmit`: clean.
- `bun test test/sync-cost-preview.test.ts test/embedding-pricing.test.ts`: 18 pass / 0 fail.

### DIFF
--- a/src/commands/sync.ts
+++ b/src/commands/sync.ts
@@ -17,7 +17,7 @@ import {
   formatCodeBreakdown,
 } from '../core/sync.ts';
 import { estimateTokens, CHUNKER_VERSION } from '../core/chunkers/code.ts';
-import { EMBEDDING_MODEL, estimateEmbeddingCostUsd } from '../core/embedding.ts';
+import { estimateEmbeddingCostUsd, getEmbeddingModelName } from '../core/embedding.ts';
 import { errorFor, serializeError } from '../core/errors.ts';
 import type { SyncManifest } from '../core/sync.ts';
 import { createProgress } from '../core/progress.ts';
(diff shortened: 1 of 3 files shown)


garrytan:
Superseded by #1632 — moved from the garrytan-agents fork into base-repo branch `garrytan/embedding-cost-model-fix` so PR CI can access secrets. Single commit cherry-picked verbatim (+64/−13), typecheck clean, 18/18 cost-preview + embedding-pricing tests pass. Co-author attribution preserved on #1632.

PR #1631, opened 2026-05-29
fix(security): bound __all__/all_sources to caller's authorized sources

confersolutions:
## The leak
A remote, source-scoped OAuth client can read content out of **isolated** sources it has no `federated_read` grant for, via `query(source_id="__all__")` (and `all_sources:true` on `code_callers`/`code_callees`, and `__all__` on `search_by_image`).

**Repro (live, v0.41.29):** as `hermes-argos` (a normal agent, `read,write`):
- `get_page hr-restricted/...` → blocked ✅
- default-scope `query` → does not surface restricted content ✅
- `query(source_id="__all__")` → **returns the `hr-restricted` page** ❌

Root cause: the `__all__` branch resolved to `{}` (no filter) instead of the caller's allowed scope — the exact escape `sourceScopeOpts` guards against for empty `allowedSources` (operations.ts:416-419).

## The fix
New `crossSourceScope(ctx)`:
- local CLI (`remote===false`, OS-trusted) → unrestricted `{}`
- remote admin/unrestricted (no `allowedSources`/`sourceId`) → `{}` (still cross-source)
- remote source-scoped client → bounded to `allowedSources`/`sourceId`

Applied at all four `__all__`/`all_sources` escape sites (`query`, `code_callers`, `code_callees`, `search_by_image`). The `query` op fully honors a federated `sourceIds` set; the scalar/bool engine APIs (`code_*`, `search_by_image`) are pinned to the caller's primary source for scoped remotes (safe; a follow-up can thread `sourceIds` for multi-source completeness on those).

## Tests
- New `test/cross-source-scope-authz.test.ts` (5 cases incl. empty-`allowedSources` must-not-widen).
- Existing source-scope suites pass (30/30); `tsc --noEmit` clean on edited files.

## Why it matters
This is the gating blocker for using gbrain "restricted" sources (`hr-restricted`, `finance-restricted`, `security-restricted`) for sensitive data — without it, `federated:false` is not a confidentiality boundary against search.

## Redeploy
Needs a brain rebuild + deploy (brain-api Coolify app + worker on S1). Recommend: review → merge → rebuild image → roll the worker/brain-api → re-run the live repro (argos `query __all__` must no longer see `hr-restricted`).


confersolutions:
Closing — opened against the upstream public repo by mistake (fork-default base). Re-targeting Confer's fork; will upstream a sanitized version separately if desired.

PR #1637, opened 2026-05-30
feat: CJK entity extraction for Chinese/Japanese/Korean names

alkalide:
## Summary

Adds support for Chinese, Japanese, and Korean (CJK) entity names in the mention extraction pipeline. Currently, CJK names are invisible to the gazetteer because:

1. **TOKEN_RE** (`/[a-zA-Z0-9]+/g`) cannot tokenize CJK characters
2. **MIN_NAME_LENGTH = 4** filters out all CJK names (typically 2-3 chars)
3. **tokenizeTitle()** returns `[]` for pure CJK titles, excluding them from the gazetteer

## Changes

| File | Change |
|------|--------|
| `src/core/by-mention.ts` | Add `hasCJK()` / `cjkCharCount()` helpers; lower min length to 2 for CJK; fix `tokenizeTitle()` to treat pure CJK titles as single tokens; add CJK substring matching pass in `findMentionedEntities` |
| `src/core/extract-ner.ts` | NER extraction falls back to plain `mentions` links when no schema pack is available (previously returned 0 links) |

## Verification

```
$ gbrain extract links --by-mention --source db
Mentions: created 27 links from 456 pages against gazetteer of 3 first-token buckets

$ gbrain list --type person
people/shuang-xuetao    person  双雪涛
people/naval            person  纳瓦尔
```

## Backward Compatibility

- English entity extraction is completely unchanged
- The CJK scan pass only activates when CJK-titled entities exist in the gazetteer
- NER fallback to `mentions` type only triggers when no schema pack with inference regex is available


irresi:
Great findings! @alkalide — verified the issues you identified are real. 
A few concerns about the fix:
- Tests should be added.
- The CJK substring pass adds O(P × C × N) on top of the existing
  O(P × N) walk (P = pages, C = CJK entries, N = body length).
- The title says CJK but support is asymmetric — tokenization covers all three (Chinese/Japanese/Korean), 
but the new `CN_*_RE` verbs and `TIMELINE_LINE_RE_CN` in `link-extraction.ts` are Chinese-only.
- The CJK bucket key is the full title (e.g. `"纳瓦尔"`), and `tokenizeForScan` is still ASCII-only — 
so [by-mention.ts:245](src/core/by-mention.ts#L245) never reaches the CJK bucket via the normal scan path.

alkalide:
Thanks for the thorough review @irresi! Addressed all 4 concerns:

**1. Tests added** — 12 new CJK-specific tests (10 pure unit + 2 engine integration), all passing alongside the existing 39.

**2. O(P×C×N) removed** — The CJK substring pass has been deleted. CJK entities now go through the normal maximal-munch scan path at O(P × N_tokens), same complexity as ASCII.

**3. Asymmetry** — Renamed `CN_*_RE` → `ZH_*_RE` (ISO 639-1), added a comment noting these are Chinese-only for link *type* extraction. Entity *name* extraction in `by-mention.ts` covers all three scripts (CJK = Chinese/Japanese/Korean) via Unicode codepoint ranges.

**4. Bucket key reachable** — The root cause. `tokenizeForScan` now emits individual CJK characters as tokens, matching `tokenizeTitle` which splits pure CJK titles into character-level tokens (e.g. "纳瓦尔" → ["纳","瓦","尔"], key = "纳"). The normal scan path finds the bucket at the "纳" token and matches subsequent "瓦"+"尔" via maximal-munch. No special pass needed.

PR #1639, opened 2026-05-30
fix: keep maintenance jobs source-local

JiraiyaETH:
## Summary

- Make source-scoped maintenance/write-through paths resolve from `sources.local_path` instead of borrowing global `sync.repo_path` or falling back to `.`.
- Keep `sync.repo_path` only as a legacy fallback for literal default-source installs.
- Update autopilot fan-out, job handlers, dream/autopilot/takes/notability/features/remediation path resolution, and source-scoped import metadata writes.
- Add regression coverage for source-local job resolution, fan-out repo paths, source-scoped import config behavior, and `put_page` write-through routing.

## Test plan

- `bun install --ignore-scripts`
- `bun run typecheck`
- `bun run verify` — 29/29 checks green
- `bun test test/source-resolver.test.ts test/jobs-source-path-resolution.test.ts test/autopilot-fanout.test.ts test/import-source-id.test.ts test/ingestion/put-page-write-through.test.ts test/import-file.test.ts test/features.test.ts` — 108 pass / 0 fail

## Notes

This was ported onto current `origin/master` in a clean side worktree. The branch intentionally does not enable embeddings, autopilot daemons, recurring jobs, or any live sync loop.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -24,6 +24,7 @@ import type { CliOptions } from './core/cli-options.ts';
 import { callRemoteTool, RemoteMcpError, unpackToolResult } from './core/mcp-client.ts';
 import { maybePromptForUpgrade } from './core/thin-client-upgrade-prompt.ts';
 import { VERSION } from './version.ts';
+import { getDefaultSourcePath } from './core/source-resolver.ts';
 
 // Build CLI name -> operation lookup
 const cliOps = new Map<string, Operation>();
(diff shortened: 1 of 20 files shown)


garrytan:
Thanks for this! We are consolidating the source-scoped dream / doctor-freshness hardening — #1695 is the broader version we are keeping. Closing this as subsumed; credit noted. Please reopen if #1695 misses something.

PR #1648, opened 2026-05-30
fix: legacy bearer tokens respect GBRAIN_SOURCE env for sourceId

### LINKED ISSUE #1336: Legacy access_tokens ignore `permissions.source_id`; MCP reads silently return empty in non-default brains
## Environment
- v0.40.7.2, postgres engine
- Single brain with one non-default source (e.g. `source_id='brain'`, no `sources.default` set or set to that source)
- Legacy `access_tokens` row (pre-OAuth, bearer token used by an MCP client)

## Symptom
All MCP read ops silently return empty:
- `list_pages` → `[]`
- `get_page` → not found
- `query` / `search` → no hits

…even though the brain has hundreds of pages on disk and in DB. Writes via MCP also land in the wrong place (under a `default` source dir) until the routing is fixed.

## Root cause
`src/core/oauth-provider.ts` (around line 596 on `0a03ca82`) hardcodes `sourceId: 'default'` for the legacy bearer-token fallback, regardless of what's in `access_tokens.permissions.source_id`:

```ts
return {
  token,
  clientId: name,
  clientName: name,
  scopes: ['read', 'write', 'admin'],
  expiresAt: …,
  sourceId: 'default',   // ← ignores permissions.source_id
} as AuthInfo;
```

The CLI resolver path already honors a three-tier fallback (token perms → brain default → `'default'`); the MCP/HTTP path doesn't.

## Workaround (operator-side, what we did)
1. `gbrain sources default <id>` to persist a brain-wide default.
2. Hand-patch `oauth-provider.ts` to pull `permissions` from the access_tokens row and resolve `sourceId = permissions?.source_id ?? brain_default ?? 'default'`.
3. `UPDATE gbrain.access_tokens SET permissions = permissions || '{"source_id":"<id>"}' WHERE …` for the legacy token.

After (1)+(2)+(3) all MCP reads/writes scope to the correct source.

## Proposed fix
Make the legacy-token branch of `oauth-provider.ts` mirror the CLI resolver's three-tier semantics:

```ts
const perms = row.permissions ?? {};
const brainDefault = await this.getBrainDefaultSourceId(); // existing helper
const sourceId = perms.source_id ?? brainDefault ?? 'default';
```

## Bonus drive-by: `gbrain reindex` unreachable
`'reindex'` is missing from `CLI_ONLY` in `src/cli.ts` (~line 30) even though `case 'reindex'` exists in the dispatch at ~line 1222. The command falls through and errors out. One-line add to `CLI_ONLY` fixes it.

---

Happy to send a PR if useful — let me know if you'd prefer to keep the legacy-token path explicitly `'default'` and force operators onto OAuth for non-default sources instead.

JavanC:
## Summary

Fixes #1336

- Legacy bearer tokens (access_tokens transport) hard-coded `sourceId` to `'default'`, causing all MCP queries to return empty results for brains using a non-default source via `GBRAIN_SOURCE`
- Adds fallback chain `process.env.GBRAIN_SOURCE || 'default'` at all three auth resolution sites
- Matches existing behavior of stdio MCP server (`server.ts:42`) which already reads `GBRAIN_SOURCE`
- OAuth clients are unaffected — their `sourceId` comes from `oauth_clients.source_id` (migration v60)

## Reproduction

1. Set up a multi-source brain with `gbrain sources add my-source --path /path`
2. Run `gbrain serve --http` with `GBRAIN_SOURCE=my-source`
3. Query via bearer token → returns empty `[]`
4. Same query via CLI (`gbrain call`) → returns correct results

## Files changed

- `src/core/oauth-provider.ts` — primary `verifyAccessToken` path
- `src/mcp/http-transport.ts` — v0.22.7 transport path  
- `src/commands/serve-http.ts` — `tokenSourceId` fallback chain

## Test plan

- [x] Verified `list_pages` returns data via HTTP bearer token with `GBRAIN_SOURCE=javan-brain`
- [x] Verified `query` semantic search returns results through cloudflared tunnel
- [x] Confirmed OAuth clients unaffected (source_id from DB, not env)

### DIFF
--- a/src/commands/serve-http.ts
+++ b/src/commands/serve-http.ts
@@ -1529,7 +1529,7 @@ export async function runServeHttp(engine: BrainEngine, options: ServeHttpOption
       // path codex flagged in plan review. Post-v60, every OAuth client
       // has source_id set; legacy bearer tokens default to 'default' in
       // verifyAccessToken. The env-fallback is gone.
-      const tokenSourceId = authInfo.sourceId ?? 'default';
+      const tokenSourceId = authInfo.sourceId ?? process.env.GBRAIN_SOURCE ?? 'default';
 
       let toolResult: Awaited<ReturnType<typeof dispatchToolCall>>;
       try {
--- a/src/core/oauth-provider.ts
+++ b/src/core/oauth-provider.ts
@@ -657,7 +657,7 @@ export class GBrainOAuthProvider implements OAuthServerProvider {
         // serve-http transport fell back to GBRAIN_SOURCE/'default' for
         // any caller without explicit scope. Operators who want a
         // narrower scope for legacy tokens migrate to OAuth.
-        sourceId: 'default',
+        sourceId: process.env.GBRAIN_SOURCE || 'default',
       } as AuthInfo;
     }
 
--- a/src/core/operations.ts
+++ b/src/core/operations.ts
@@ -779,7 +779,12 @@ const put_page: Operation = {
     const trustedWorkspace = ctx.viaSubagent === true
       && Array.isArray(ctx.allowedSlugPrefixes)
       && ctx.allowedSlugPrefixes.length > 0;
-    if (ctx.remote !== false && !trustedWorkspace) {
+    // JavanC fork: env-gated escape hatch for single-user trusted deployments
+    // (Mac mini canonical host where every MCP caller is an internal agent
+    // gated by bearer token + Cloudflare WAF). Opt-in only; default behavior
+    // preserves upstream's security boundary for untrusted webhook input.
+    const trustedRemote = process.env.GBRAIN_TRUSTED_REMOTE === '1';
+    if (ctx.remote !== false && !trustedWorkspace && !trustedRemote) {
       autoLinks = { skipped: 'remote' };
       autoTimeline = { skipped: 'remote' };
     } else if (result.parsedPage) {
--- a/src/mcp/http-transport.ts
+++ b/src/mcp/http-transport.ts
@@ -207,7 +207,7 @@ export async function startHttpTransport(opts: HttpTransportOptions) {
         // serve-http fallback chain that was removed for OAuth clients
         // (migration v60 backfills oauth_clients.source_id). This path
         // is for the older v0.22.7 access_tokens transport.
-        sourceId: 'default',
+        sourceId: process.env.GBRAIN_SOURCE || 'default',
       };
     } catch {
       return { ok: false };


jacklau1993:
Hey — sharing a related use case from a downstream user running gbrain as a multi-agent shared brain via Zeabur, in case it helps future readers.

**Same bug class, different transport.** We hit the silent default-to-`default` symptom on a multi-source brain (canonical source `nikke-memory`, federated setup) but for the **stdio MCP** transport rather than HTTP bearer. Our `v0.42.8.0` `src/mcp/server.ts:42` already reads `process.env.GBRAIN_SOURCE`, so the stdio side is technically covered — the underlying issue for us was that the env var was simply not being set in the subprocess at all (openab spawns agents with `inherit_env = []`, and the `[mcp_servers.gbrain].environment` block in our `~/.grok/config.toml` wasn't being honored by the Grok MCP client). Same blind-spot shape, different layer.

**3-layer defense that worked for us** (all on the calling side, no gbrain patch needed):

1. **Wrapper that hard-codes the env** — `/home/agent/bin/gbrain`:
   ```sh
   #!/bin/sh
   export HOME=/home/agent
   export PATH="/home/agent/.bun/bin:$PATH"
   : "${GBRAIN_SOURCE:=nikke-memory}"
   export GBRAIN_SOURCE
   exec /home/agent/.bun/bin/gbrain "$@"
   ```
   POSIX default-substitution `:=` means even a clean parent env (`env -i`) gets the right value. Defense in depth: doesn't rely on the MCP client correctly threading `[mcp_servers.gbrain].environment` to subprocess.

2. **Grok config.toml explicit env block** — `~/.grok/config.toml`:
   ```toml
   [mcp_servers.gbrain]
   command = "/home/agent/bin/gbrain"
   args = ["serve"]
   enabled = true

   [mcp_servers.gbrain.environment]
   GBRAIN_SOURCE = "nikke-memory"
   ```

3. **`~/.gbrain-source = nikke-memory` dotfile** — CLI fallback for sanity, since the MCP stdio transport in our version (`v0.42.8.0 src/mcp/server.ts:42`) does NOT consult the dotfile the way the CLI 6-tier `core/source-resolver.ts` does. **This is the only real gap on the stdio side**, and it's the analog of your #1648 fix for the stdio path. A 2-line patch calling `resolveSourceId(engine, explicit=null, cwd=process.cwd())` from `server.ts:42` would close it; we'd love a follow-up for it.

**End-to-end verification** (the gold-standard proof for routing): the real Grok MCP call returned an error whose body contained the string `(source=nikke-memory)` — printed by `src/core/operations.ts:549` from `ctx.sourceId`, which `mcp/server.ts:42` set from `process.env.GBRAIN_SOURCE`. The error message itself is the proof that the env plumbing through the entire Grok → openab → wrapper → gbrain chain is correct.

**Two adjacent observations for v0.42.8.0** (independent of your PR, FYI):

1. The new content-sanity gate (`v0.42.8.0 src/core/import-file.ts:469` `assessContentSanity`) hard-blocks minimal test slugs (e.g. `ops/test/...`) and surfaces as `Page not found: <slug> (source=...)` from `_upsertChunksOnce` (the page INSERT never durably commits, so the chunk SELECT returns 0 rows). Both CLI and MCP fail identically. It looks like a routing bug at first glance but isn't — the actual cause is the gate. Surfacing the gate verdict explicitly (e.g. `ContentSanityBlocked` with the `assessContentSanity` reason) would save a lot of debug time for downstream users.

2. Your second commit (`feat(put_page): add GBRAIN_TRUSTED_REMOTE env override for auto_link/auto_timeline`) is directly relevant for our case. We have a single-user trusted deployment behind Zeabur internal DNS; every MCP call is internal, so the security-skip on `auto_link`/`auto_timeline` is pure cost for us. Will pick up the opt-in once the PR lands.

**Reference**: I documented our full 3-layer pattern in `ops/diagnostics/sakura-gbrain-mcp-source-routing-hardening-2026-06-03` (will be ingested into our gbrain via the team sync worker). Cross-linking to PR #1648 from there.

Thanks for the fix — looking forward to it landing.

time-attack:
Thanks for the contribution, @JavanC! 🙏

What this does: This PR fixes a genuine issue where legacy bearer tokens incorrectly hard-code sourceId to 'default', causing MCP queries to return empty results in multi-source brains. However, this issue was already resolved in master (v0.42.41.0+) with…

Closing this one to keep the queue focused on higher-impact changes. Nothing wrong with the effort — reopen if you'd like to make the case for it and we'll reconsider.

Appreciate your time.

Issue #1619, opened 2026-05-29
gbrain dream: extract phase silently drops all link/timeline batches when the module-singleton DB connection is null mid-cycle (retry-on-null can't recover)


xaviroblessarries:
## Summary

On a Postgres (Supabase) brain-host running the nightly `gbrain dream` cycle, the `extract` phase reliably writes **zero** links and timeline entries:

```
[extract.links_fs] connection blip, retrying (attempt 1/3): No database connection: connect() has not been called.
[extract.links_fs] connection blip, retrying (attempt 2/3): No database connection: connect() has not been called.
[extract.links_fs] connection blip, retrying (attempt 3/3): No database connection: connect() has not been called.
  batch error (N link rows lost): No database connection: connect() has not been called.
Links: created 0 from 346 pages
Timeline: created 0 entries from 346 pages
```

The cycle's `sync` phase (earlier in the same process) works fine, and a separate `gbrain extract all` invocation (fresh process) writes the links correctly. So the brain repo + content are fine — the problem is the cycle's shared DB connection going null before the `extract` phase.

## What's certain

`PostgresEngine.sql` getter returns `this._sql ?? db.getConnection()`. For the module-style engine `gbrain dream` uses, `_sql` is null by design, so it falls through to `db.getConnection()` (`src/core/db.ts:151`), which throws `connect() has not been called` when the module singleton `sql` (`src/core/db.ts:7`) is null.

`addLinksBatch` → `batchRetry` → `withRetry` re-invokes `_addLinksBatchOnce`, which re-reads `const sql = this.sql` on **every** attempt. Since both `_sql` and the singleton are null, all 3 retries hit the identical error and the batch is dropped. The v0.41.19.0 / v0.41.2.1 retry work only recovers a *transient* pool drop where postgres.js reconnects the existing `sql` object — it structurally cannot recover a permanently-nulled singleton (retry-on-null re-reads the same null).

The extract phase also swallows the per-batch error and reports `created 0` rather than failing loudly (`src/commands/extract.ts`), so the cycle exits 0 and the data loss is silent.

## Likely trigger (hypothesis)

[...]

PR #1651, opened 2026-05-30
fix(postgres-engine): only the singleton owner may disconnect it (#1471)

### LINKED ISSUE #1471: resolveLintContentSanity disconnects shared module-level db singleton, killing the cycle's main engine connection

nullhex-io:
## Problem

Fixes #1471 (root cause). Should also resolve the symptom issues that stem from the same module-singleton-null-mid-cycle defect: #1404, #1413, #1491, #1619.

A `PostgresEngine` connecting via the module-singleton path (no `poolSize`) is tagged `_connectionStyle = 'module'` whether it CREATED the shared `db.ts` `sql` singleton (the **owner** - the CLI cycle engine) or merely found it already connected (a **borrower** - the probe engine in `resolveLintContentSanity`, doctor, future probes). The old `disconnect()` called `db.disconnect()` for BOTH, so a short-lived borrower's `finally`-block teardown nulled the singleton the owner cycle engine was still using. Every later phase then threw `No database connection: connect() has not been called`, and `runCycle`'s finally stranded a row in `gbrain_cycle_locks`.

`_connectionStyle` alone can't separate owner from borrower - both are `'module'`. This implements the ownership-flag fix proposed in #1471.

## Fix

- **`db.ts`**: add `isConnected()` so an engine can tell, at `connect()` time, whether the singleton already exists.
- **`postgres-engine.ts`**: add `_ownsModuleSingleton`. `connect()` samples ownership *before* delegating to `db.connect()`; `disconnect()` calls `db.disconnect()` only when this engine owns the singleton. A borrower clears its own marker without touching the shared connection.

Scope:
- Instance-pool engines (`poolSize` set) are unaffected - they never take the module branch.
- `reconnect()` stays correct: an owner's disconnect+reconnect re-acquires ownership (the singleton is null right after its own `db.disconnect()`), and a borrower re-borrows the owner's (possibly refreshed) singleton.

## Tests

- New `test/postgres-engine-singleton-ownership.test.ts`: source-level guardrails (DB-free, matching the existing `postgres-engine.test.ts` convention, since runtime-mocking postgres.js's tagged-template interface is painful under bun ESM). Asserts `db.isConnected()` is sampled before `db.connect()`, and that the only `db.disconnect()` in `disconnect()` is guarded by `_ownsModuleSingleton`.
- `bun run typecheck`: clean.
- Regression (59 tests green): `postgres-engine`, `db-disconnect-audit`, `connection-manager.serial`, `autopilot-reconnect-classifier`, `worker-shutdown-disconnect`.

The live `gbrain dream` cycle is the e2e exercise of this path; the ownership logic above is what keeps the owner's singleton alive across a borrower probe's teardown.

### DIFF
--- a/src/core/db.ts
+++ b/src/core/db.ts
@@ -148,6 +148,16 @@ export async function setSessionDefaults(_sql: ReturnType<typeof postgres>): Pro
   // No-op: timeouts are now applied as startup parameters in resolveSessionTimeouts().
 }
 
+/**
+ * Whether the module-level singleton is currently connected. Lets a
+ * PostgresEngine decide, at connect() time, whether it is creating the
+ * singleton (owner) or borrowing an existing one (borrower) - so only the
+ * owner disconnects it later. See #1471.
+ */
+export function isConnected(): boolean {
+  return sql !== null;
+}
+
 export function getConnection(): ReturnType<typeof postgres> {
   if (!sql) {
     throw new GBrainError(
--- a/src/core/postgres-engine.ts
+++ b/src/core/postgres-engine.ts
@@ -106,6 +106,17 @@ export class PostgresEngine implements BrainEngine {
   private _connectionStyle: 'instance' | 'module' | null = null;
 
   /**
+   * #1471: module-singleton OWNERSHIP. Only the engine whose connect() actually
+   * created the shared db.ts `sql` singleton may disconnect it. A borrower (a
+   * probe engine constructed while the singleton already exists - e.g.
+   * resolveLintContentSanity or doctor) must clear its own marker WITHOUT
+   * calling db.disconnect(), or it nulls the `sql` the owner cycle engine is
+   * still using and every later phase throws "connect() has not been called".
+   * `_connectionStyle` alone can't separate owner from borrower: both are 'module'.
+   */
+  private _ownsModuleSingleton = false;
+
+  /**
    * v0.30.1 (Fix 1 + X1 + T5): instance-owned ConnectionManager.
    * - INSTANCE-owned: each PostgresEngine constructs its own.
    * - Worker engines (cycle, sync) inherit via opts.parentConnectionManager.
@@ -177,9 +188,14 @@ export class PostgresEngine implements BrainEngine {
       });
       this.connectionManager.setReadPool(this._sql);
     } else {
-      // Module-level singleton (backward compat for CLI main engine)
+      // Module-level singleton (backward compat for CLI main engine).
+      // #1471: sample ownership BEFORE delegating to db.connect(). If the
+      // singleton already exists we are a borrower and must never disconnect
+      // it; only the engine that creates it owns its teardown.
+      const ownsSingleton = !db.isConnected();
       await db.connect(config);
       this._connectionStyle = 'module';
+      this._ownsModuleSingleton = ownsSingleton;
 
       // v0.30.1: connection-manager wraps the module singleton.
       if (url) {
@@ -220,7 +236,13 @@ export class PostgresEngine implements BrainEngine {
       return;
     }
     if (this._connectionStyle === 'module') {
-      await db.disconnect();
+      // #1471: only the owner disconnects the shared singleton; a borrower
+      // clears its markers so a probe engine's teardown can't clobber the
+      // owner's live connection.
+      if (this._ownsModuleSingleton) {
+        await db.disconnect();
+        this._ownsModuleSingleton = false;
+      }
       this._connectionStyle = null;
     }
     // else: nothing to disconnect (already done or never connected)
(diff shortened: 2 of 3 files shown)


nullhex-io:
#1805 looks like the canonical landing for this bug class and credits this approach - appreciated. Happy to close this in favor of it; leaving it open for now since #1805 still shows conflicts against master. Say the word if you'd rather I rebase this one instead.

garrytan:

PR #1652, opened 2026-05-30
feat(guardrails): vendor-neutral content guardrail seams

garrytan-agents:
## What

Exposes **vendor-neutral content guardrail seams** at the five boundaries where external content enters GBrain's retrieval layer and where queries/tool-inputs enter the LLM gateway. Lets a content firewall (prompt-injection / RAG-poison detector, PII scrubber, etc.) be hooked in **without binding GBrain to any specific vendor**.

OSS ships **inert** — zero guardrails registered by default, every seam is a no-op until an operator registers a provider.

## Why

Content poisoning (a malicious page, a booby-trapped tweet) only becomes dangerous at the moment it's ingested into the retrieval layer and made searchable. That import boundary — plus the gateway's own LLM calls — is the right place to let an external classifier observe. Rather than wire one vendor into core, this adds a generic seam that any guardrail backend implements against.

## The seam

New module `src/core/guardrails.ts`:
- `runGuardrails({ hook, content, metadata }): Promise<void>`
- `registerGuardrailProvider` / `unregisterGuardrailProvider`
- `hasGuardrails()` fast-path guard for hot paths

### Hard invariants (enforced by `test/guardrails.test.ts`)
- **Observe-only** — returns `void`; callers never branch on a verdict. Cannot block/rewrite/drop/retry. Enforcement, if ever added, gets its own RFC-gated seam.
- **Fail open** — provider throw/reject/timeout/network-error is swallowed; a broken guardrail never breaks an ingest, query, or tool call.
- **Inline await** — provider sees content at the exact pre-persist / pre-inference moment.
- **No verdict persistence** — providers own their own audit trail.
- **Content boundaries** — passes only the ingest/user payload; never system prompts, full history, tool output, LLM output, embeddings, or multimodal/OCR/rerank.

### Hooks

| `hook` | Location | Fires |
| --- | --- | --- |
| `file_storage.markdown` | `importFromContent` | after parse + size guard, before sanity/hash/chunk/embed/write |
| `file_storage.code` | `importCodeFile` | after size guard, before hash/chunk/embed/write |
| `ai_gateway.chat` | `chat` | latest user message only, before inference |
| `ai_gateway.expand` | `expand` | query, before expansion model call |
| `ai_gateway.tool_input` | `toolLoop` | `{toolName, input}`, before pending-persist + execution |

## Tests

- `test/guardrails.test.ts` — 14 tests: inert-by-default, register/unregister/idempotent, fail-open isolation, inline-await, verdict-ignored, empty/blank short-circuit, content+metadata pass-through.
- Existing `import-file.test.ts` + `import-file-content-sanity.test.ts` — 40 tests still green (hot path undisturbed).
- `tsc --noEmit` clean across the repo.

## Docs

`docs/guardrails.md` — contract, seam table, provider-authoring guide.

## Scope / non-goals

- No enforcement. This is the observe seam only.
- No bundled vendor. A guardrail provider lives in its own package and registers at init.

### DIFF
--- a/docs/guardrails.md
+++ b/docs/guardrails.md
@@ -1,0 +1,102 @@
+# Content Guardrail Seams
+
+GBrain exposes **vendor-neutral guardrail seams** at the boundaries where
+external content enters the retrieval layer and where queries/tool-inputs enter
+the LLM gateway. A guardrail is any external classifier — a content firewall, a
+prompt-injection detector, a PII scrubber — that wants to *observe* content at
+those boundaries.
+
+The OSS distribution ships **inert**: zero guardrails are registered by default,
+and every seam is a no-op until an operator registers a provider.
+
+## Design contract (hard invariants)
+
+These hold for every seam and are enforced by `test/guardrails.test.ts`:
+
+- **Observe-only.** `runGuardrails()` returns `void`. Callers never branch on a
+  provider verdict. A guardrail registered through this interface *cannot*
+  block, rewrite, drop, retry, or reorder GBrain behavior. Enforcement, if ever
+  added, will get its own explicitly-named seam and its own RFC — it will not
+  silently reuse this one.
+- **Fail open.** Missing config, provider throw/reject, timeout, and network
+  error are all swallowed. A broken guardrail never breaks an ingest, a query,
+  or a tool call.
+- **Inline await.** Hooks await the provider before proceeding, so the
+  classifier sees content at the exact pre-persist / pre-inference moment.
+- **No verdict persistence.** GBrain writes no guardrail rows. Providers own
+  their own audit trail.
+- **Content boundaries.** Hooks pass only the ingest/user-facing payload — the
+  markdown/code body, the last user message, the expansion query, the tool
+  input. They never pass system prompts, full chat history, tool *output*, LLM
+  output, embeddings, or multimodal/OCR/rerank payloads.
+
+## The five seams
+
+All seams call `runGuardrails({ hook, content, metadata })` from
+`src/core/guardrails.ts`.
+
+| `hook` | Location | Fires |
+| --- | --- | --- |
+| `file_storage.markdown` | `import-file.ts` → `importFromContent` | After `parseMarkdown` + size guard, **before** content-sanity, hashing, chunking, embedding, DB write |
+| `file_storage.code` | `import-file.ts` → `importCodeFile` | After code size guard, **before** hashing, code-chunking, embedding, DB write |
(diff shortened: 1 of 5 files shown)


garrytan-agents:
## ✅ Pre-merge review gate

Ran the GStack `/review` critical pass + tests before marking ready.

**Tests:** `test/guardrails.test.ts` — 14/14 pass. Existing `import-file.test.ts` + `import-file-content-sanity.test.ts` — 40/40 pass (hot path undisturbed). `tsc --noEmit` clean across the repo.

**`/review` critical pass (5 categories):**
- **SQL & Data Safety** — ✅ clean. No SQL in diff; seam writes nothing to DB/vector store (verdicts explicitly not persisted).
- **Race Conditions & Concurrency** — ✅ clean. `providers` Map is snapshotted via `Array.from()` before `Promise.all` iteration, so mid-flight register/unregister can't mutate the iteration set. No find-or-create / check-then-write.
- **LLM Output Trust Boundary** — ✅ clean, and on-point: this seam *observes* inbound content pre-persist; it never consumes LLM output, and the provider return value is typed `unknown` and ignored. No path where a verdict influences a DB write, mailer, or fetch. This is the observation point *for* the stored-prompt-injection class, not a new instance of it.
- **Shell Injection** — ✅ clean. No exec/eval/subprocess/shell.
- **Enum & Value Completeness** — ✅ clean. New `GuardrailHook` union (5 values); traced every consumer — each value emitted by exactly one seam caller, no `switch`/`case` consumes it (providers get it as opaque metadata), so no exhaustiveness gap.

**Informational:** redundant `hasGuardrails()` check in the gateway wrapper is intentional (skips the message-array walk on every chat call in the common zero-guardrail case). Import is used. No slop.

**Codex review:** attempted `codex exec review`; hung in this container's read-only sandbox (no bubblewrap) without emitting findings — environment limitation, not a code signal. `/review` is the authoritative gate here.

**Verdict: ready to pick up.**

garrytan:
Superseded by #1660. Rebased into a base-repo branch (`garrytan/guardrails-seam`) so CI gets secret access per the `garrytan-agents` workflow in CLAUDE.md, and folded in the v0.41.34.0 release bookkeeping (VERSION + CHANGELOG). Your feature commit was cherry-picked verbatim — authorship preserved, code byte-identical. Verified on the new branch: `bun run verify` 29/29 green, 14/14 guardrail tests, import-file hot path undisturbed. Thank you for this — the five-seam shape is exactly right, and it closes the injection-via-filesystem gap. Closing this one.

PR #966, opened 2026-05-13
fix(autopilot --install): source ~/.zshenv before zshrc/bashrc

p3ob7o:
## Problem

The autopilot wrapper script generated by `gbrain autopilot --install` sources `~/.zshrc` (with `~/.bashrc` fallback) but never `~/.zshenv`. In zsh, `~/.zshenv` is where environment variables belong — it runs for every shell invocation, including non-interactive subprocesses. `~/.zshrc` only runs for interactive shells.

macOS LaunchAgents run as non-interactive subprocesses, so secrets exported in `~/.zshenv` (`GBRAIN_DATABASE_URL`, `OPENAI_API_KEY`, `ANTHROPIC_API_KEY`) never reach autopilot. Autopilot then can't connect to the DB.

## Fix

Add `[ -f ~/.zshenv ] && source ~/.zshenv 2>/dev/null` before the existing `source ~/.zshrc`. The `-f` guard makes it a no-op for anyone not using zshenv.

## Verification

Tested on macOS 26.5 with zshenv-based env management. Autopilot picks up `GBRAIN_DATABASE_URL` and connects to Supabase Postgres on LaunchAgent startup. Confirmed across multiple `gbrain autopilot --install` re-runs.

---
<a href="https://app.blacksmith.sh/garrytan/codesmith/gbrain/pr/966"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need.</sup>

- [ ] Let Codesmith autofix CI failures and bot reviews

PR #1654, opened 2026-05-30
fix(autopilot): put the gbrain CLI dir on PATH in the generated wrapper

nullhex-io:
## Problem

`gbrain autopilot --install` generates `~/.gbrain/autopilot-run.sh`, which sources `~/.zshenv` + `~/.zshrc`/`~/.bashrc` to inherit PATH (and secrets). But interactive shell rc files commonly early-return for non-interactive shells, or guard the PATH line behind interactivity - so the gbrain CLI's own directory (`~/.bun/bin` for a bun install, an nvm bin, a custom prefix, etc.) never reaches PATH when the wrapper runs under launchd / systemd / cron.

Result: autopilot starts (the wrapper `exec`s the absolute CLI path it baked in), but its internal subprocess spawns (`gbrain <subcommand>`) can't resolve the CLI:

```
Could not resolve the gbrain CLI path. Install gbrain so it is on $PATH (e.g. /usr/local/bin/gbrain) ...
```

The daemon shows `active`, but every cycle's subprocess work silently no-ops - continuous capture quietly does nothing. (Observed on a Linux systemd user-service install with a bun-installed CLI whose `~/.bashrc` early-returns for non-interactive shells.)

## Fix

`writeWrapperScript()` already resolves the absolute CLI path for `exec`. Reuse it: export `dirname(resolvedCliPath)` onto PATH, placed AFTER the rc sourcing (so rc can't clobber it) and BEFORE `exec`. Derived from the path the installer already found, so it works for any install location (bun, npm-global, nvm, custom prefix):

```sh
export PATH='<cli-dir>':"$PATH"
exec '<cli-path>' autopilot --repo '<repo>'
```

## Tests

- `test/autopilot-install.test.ts`: source-level guardrail (matching the existing `#966` env-source-order wrapper test) - asserts the wrapper exports the CLI dir on PATH, after rc sourcing and before `exec`.
- `bun run typecheck`: clean.
- Existing autopilot-install suite: green (6 tests).


time-attack:
Thank you for the detailed diagnosis — the failure mode you documented (bashrc's non-interactive early-return starving the wrapper's PATH under launchd/systemd) is exactly right, and it has since been fixed on master. #3397 (commit f0a28eb2) now derives the runtime dir from `process.execPath` at install time and exports it onto PATH after the rc sourcing and before the exec, with `$HOME/.bun/bin` kept as a fallback; #2747 separately fixed `resolveGbrainCliPath` to forward `process.env` to its `which` call. Closing as already fixed, but the write-up was genuinely useful.

Issue #1556, opened 2026-05-27
sync crashes on inferred date-like titles from bare markdown


hassan-albalawi:
## Summary

`gbrain sync` can crash on bare markdown files whose inferred title is a date-looking string, e.g. `# 2026-05-26` / `memory/2026-05-26.md`.

The failure is recorded as `UNKNOWN` in `~/.gbrain/sync-failures.jsonl`:

```text
opts.title.toLowerCase is not a function. (In 'opts.title.toLowerCase()', 'opts.title.toLowerCase' is undefined)
```

## Environment

- gbrain: `0.41.14.0`
- local repo HEAD: `32f8be96` (`v0.41.14.0 fix(#1451): close RESOLVER.md drift bug class structurally (#1458)`)
- `origin/master` checked locally also appears affected as of `10816cba`
- Engine: Postgres + pgvector
- Embedding provider: `zeroentropyai:zembed-1`

## Repro

A bare markdown file with no YAML frontmatter:

```markdown
# 2026-05-26

- any content
```

When import/sync runs frontmatter inference, `applyInference()` prepends:

```yaml
---
title: 2026-05-26
type: note
date: "2026-05-26"
---
```

Then `parseMarkdown()` returns `title` as a runtime `Date` object because YAML parses unquoted `2026-05-26` as a timestamp/date. Later `assessContentSanity()` calls:

```ts
const titleLower = opts.title.toLowerCase();
```

and crashes because `opts.title` is not a string.

Minimal local proof from the affected install:

```ts
import { readFileSync } from "fs";
import { parseMarkdown } from "./src/core/markdown.ts";
import { applyInference } from "./src/core/frontmatter-inference.ts";

const rel = "memory/2026-05-26.md";
const raw = readFileSync(`/home/hassan/.openclaw/workspace/${rel}`, "utf8");
const inf = applyInference(rel, raw);
console.log(inf.content.split("\n").slice(0, 6).join("\n"));

const parsed = parseMarkdown(inf.content, rel);
console.log(typeof parsed.title, parsed.title?.constructor?.name, parsed.title);
```

Observed:

```text
---
title: 2026-05-26
type: note
date: "2026-05-26"
---

object Date 2026-05-26T00:00:00.000Z
```

## Expected

`title` should remain a string throughout the ingestion pipeline, and sync should not block on date-heading/daily-note files.

## Likely fix

One or both of these would close the bug class:

[...]

PR #1659, opened 2026-05-30
fix(content-sanity): null-safe title in assessContentSanity (fixes opts.title.toLowerCase sync crash)

### LINKED ISSUE #1658: sync crashes in assessContentSanity on pages with undefined title (opts.title.toLowerCase) — confirms #1556 on current master
## Summary

`gbrain sync`/ingest crashes in `assessContentSanity()` when a page's parsed title is `undefined`:

```text
opts.title.toLowerCase is not a function. (In 'opts.title.toLowerCase()', 'opts.title.toLowerCase' is undefined)
```

This is the crash from #1556, which the original reporter closed as "premature, filed while validating a local install — will file fresh after confirming against a clean/current install." Confirming here against current `master`.

## Environment

- gbrain HEAD: `041d89b` (`v0.41.29.0`)
- Engine: Postgres + pgvector (Supabase), embedding `openai:text-embedding-3-small`
- Reproduced: 988 failure events across 247 distinct pages logged to `~/.gbrain/sync-failures.jsonl`, all dated right after `v0.41.26.0` (#1571) landed the junk-title content-sanity check.

## Root cause

`src/core/content-sanity.ts` types the param as `title: string` (non-optional):

```ts
export function assessContentSanity(opts: {
  ...
  title: string;            // <- callers can pass undefined at runtime
  ...
}) {
  ...
  const titleLower = opts.title.toLowerCase();          // throws
  ...
  if (p.pattern.test(opts.title)) ...                   // also unsafe
}
```

But `src/core/import-file.ts` (and `sources.ts`, `lint.ts`, `doctor.ts`) call it with `title: parsed.title`, and `parsed.title` is `undefined` whenever a page has no `title:` frontmatter field. This bites two ways:

1. **Title-less pages** (e.g. daily/date notes with no `title:`).
2. **History replay:** `sync` replays git history commit-by-commit; a page that has a `title:` *today* did not at the historical commits being replayed, so `parsed.title` is genuinely `undefined` for those revisions and the whole sync run aborts.

## Repro

A markdown file with no `title:` in frontmatter, then `gbrain sync` (or import). `assessContentSanity` throws on the first such page.

## Fix

Coerce defensively at the single choke point (covers all four callers) and make the type honest:

```ts
title?: string | null;
// ...
const title = String(opts.title ?? '');
const titleLower = title.toLowerCase();
// ...
if (p.pattern.test(title)) ...
```

All 76 existing `content-sanity` tests pass; `null`/`undefined`/number/empty/string inputs all return a result instead of throwing. PR incoming.

seaneoliver:
## What

Makes `assessContentSanity()` null-safe on `title`.

`opts.title` is typed `string`, but every caller passes `parsed.title`
(`import-file.ts`, `sources.ts`, `lint.ts`, `doctor.ts`), which is
`undefined` whenever a page has no `title:` frontmatter field. Two paths hit it:

1. Title-less pages (e.g. date/daily notes).
2. `sync` replays git history commit-by-commit — a page with a `title:`
   today did not at older revisions, so `parsed.title` is `undefined` for
   those and the whole sync aborts.

`opts.title.toLowerCase()` (and the title-scope `pattern.test(opts.title)`)
then throw:

```text
opts.title.toLowerCase is not a function. (In 'opts.title.toLowerCase()', 'opts.title.toLowerCase' is undefined)
```

## Fix

Coerce once at the single choke point and make the type honest:

- `title?: string | null` in the opts type
- `const title = String(opts.title ?? '')`, reused for `titleLower` and the title-scope pattern test

## Testing

- `bun test test/content-sanity.test.ts test/content-sanity-literals.test.ts` → 76 pass / 0 fail
- `null` / `undefined` / number / empty / normal string inputs all return a `ContentSanityResult` instead of throwing.

Closes #1658. Confirms and supersedes #1556 (which the original reporter closed as premature pending confirmation against current master — confirmed here on `041d89b` / v0.41.29.0).

### DIFF
--- a/src/core/content-sanity.ts
+++ b/src/core/content-sanity.ts
@@ -232,8 +232,11 @@ export function assessContentSanity(opts: {
   compiled_truth: string;
   /** Post-parseMarkdown timeline section (empty string if no sentinel). */
   timeline: string;
-  /** Post-parseMarkdown title. Some patterns key on title alone. */
-  title: string;
+  /** Post-parseMarkdown title. Some patterns key on title alone.
+   *  May be undefined/null: pages without a `title:` frontmatter field
+   *  (or git-history commits replayed by `sync` that predate the field)
+   *  yield `parsed.title === undefined`. Coerced defensively below. */
+  title?: string | null;
   /** Effective warn threshold; defaults to DEFAULT_BYTES_WARN. */
   bytes_warn?: number;
   /** Effective block threshold; defaults to DEFAULT_BYTES_BLOCK. */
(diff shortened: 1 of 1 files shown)


garrytan:

PR #1669, opened 2026-05-30
fix(db-lock): self-heal cycle-lock refresh on pooler-reaped CONNECTION_ENDED

garrytan-agents:
## Problem

A production worker logged **400+ "crashes" / 24h** — one every ~5 minutes — all with the same chain:

```
[cycle.lint] done
Promotion error: No database connection: connect() has not been called.
Job NNNN (autopilot-cycle) released after infrastructure abort (lock-renewal-failed); stall detector will requeue (no attempt burned)
write CONNECTION_ENDED aws-1-...pooler.supabase.com:6543
worker_exited code=1 ... likely_cause=runtime_error
```

These aren't data-losing (jobs requeue, no attempt burned — `withRefreshingLock`'s designed safety release) but they are noisy and throughput-killing: the worker restarts every cycle and the autopilot queue backs up.

## Root cause

`withRefreshingLock`'s refresh tick fires every `(TTL/6)` ms — **~5 min** for the default 30-min cycle lock. But the read pool's `idle_timeout` is **20s** (`connection-manager.ts`), so the connection is *always* reaped between ticks. Against a transaction-mode pooler (PgBouncer / Supavisor :6543) the reconnect races and postgres.js raises **`write CONNECTION_ENDED`**.

That throw escapes both `engineSelectOne` (the SELECT-1 heartbeat) and `handle.refresh()`, trips `withRefreshingLock`'s catch, sets `healthOk=false`, and the worker voluntarily releases + exits code 1.

The existing retry matcher (`retry-matcher.ts`) caught the **downstream** `No database connection` GBrainError — but only *after* the pool was already nulled. The **root** `CONNECTION_ENDED` had no matching pattern, so the first (recoverable) error escaped retry entirely.

## Fix (two minimal, idiomatic changes)

1. **`retry-matcher.ts`** — add `CONNECTION_ENDED` (postgres.js error *code* + message pattern). Benefits every `withRetry` call site, not just locks.

2. **`db-lock.ts`** — wrap `refresh()` and the SELECT-1 heartbeat in `withRetry` with a duck-typed `reconnect()` callback (PostgresEngine-private; `undefined` for PGLite, which can't lose a socket). A reaped connection self-heals (rebuild pool, retry) instead of killing the lock holder. Mirrors the engine-level `batchRetry` idiom from #1570.

## Tests

- `+2` CONNECTION_ENDED cases in `retry-matcher.test.ts` (code form + message form).
- `tsc --noEmit` clean.
- `retry-matcher` (17 pass) + `db-lock-refresh` + `worker-lock-renewal` + `worker-lock-renewal-shape` suites green.

## Notes / alternatives considered

- **Bumping `idle_timeout`**: rejected — fights the pooler's own reaping and just moves the race.
- **Pinning refresh to `withReservedConnection`** (the `db-lock.ts:552` "Lane B follow-up" TODO): heavier surface; `withRetry`+reconnect achieves the same self-heal with far less risk and reuses the existing primitive. The reserved-connection pin is still a reasonable future hardening.

### DIFF
--- a/src/core/db-lock.ts
+++ b/src/core/db-lock.ts
@@ -23,6 +23,21 @@
  */
 import { hostname } from 'os';
 import type { BrainEngine } from './engine.ts';
+import { withRetry } from './retry.ts';
+
+/**
+ * Resolve a best-effort reconnect callback for `withRetry`. `reconnect()`
+ * is a PostgresEngine-private method (not on the BrainEngine interface), so
+ * we duck-type it. Returns undefined for engines (PGLite) that don't expose
+ * it — withRetry then simply retries against the existing pool, which is the
+ * correct behavior for an in-process engine that can't lose a socket.
+ */
+function resolveReconnect(engine: BrainEngine): (() => Promise<void>) | undefined {
+  const maybe = engine as unknown as { reconnect?: () => Promise<void> };
+  return typeof maybe.reconnect === 'function'
+    ? () => maybe.reconnect!()
+    : undefined;
+}
 
 export interface DbLockHandle {
   id: string;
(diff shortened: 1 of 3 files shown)


time-attack:
Closing: current master already contains this fix. Master already self-heals this exact failure chain by a different implementation: src/core/retry-matcher.ts:29-33,114,142-144 matches CONNECTION_ENDED (library code + message form, incl. isConnectionEndedError); src/core/db-lock.ts:823-846 (#1794 rework) refreshes via the direct session pool and on failure logs '[lock-refresh] ...; will retry next tick' instead of releasing + exiting code 1; src/core/postgres-engine.ts:5640-5680 has reconnect() with pooler-reap classification + pool-recovery audit (#1685). The PR's mechanism (withRetry around handle.refresh) was superseded.

PR #1673, opened 2026-05-31
fix(embed): only reject user_provided_model_unset when modelId is genuinely absent

### LINKED ISSUE #1672: llama-server embedding always returns user_provided_model_unset even when model name is configured
### Summary

`diagnoseEmbedding()` in `src/core/ai/gateway.ts` always returns `user_provided_model_unset` for any provider with `user_provided_models: true` and an empty `models` array, even when the user has specified a model name via `embedding_model`.

### Reproduction

1. Set `embedding_model: "llama-server:text-embedding-qwen3-embedding-0.6b"` in config.json
2. Start llama-server / LM Studio with the model loaded
3. Run `gbrain embed --stale`
4. Observe: `Provider "llama-server" requires a specific model name to be configured.`

### Root cause

`src/core/ai/gateway.ts` lines 670-682 checks `tp.models.length === 0 && isUserProvided` but never inspects `parsed.modelId`. For `llama-server:text-embedding-qwen3-embedding-0.6b`, `parseModelId()` correctly extracts `modelId = "text-embedding-qwen3-embedding-0.6b"`, but the guard ignores it.

### Fix

Add `!parsed.modelId` to the condition so `user_provided_model_unset` only fires when the model name is genuinely absent:

```diff
if (
  Array.isArray(tp.models) &&
  tp.models.length === 0 &&
  (recipe.id === "litellm" || isUserProvided) &&
+ !parsed.modelId
) {
```

### Workaround

Use `--no-embed` on sync/import operations until fix is merged.

justemu:
### Problem

`diagnoseEmbedding()` unconditionally returns `user_provided_model_unset` for any recipe with `user_provided_models: true` and an empty `models` array (e.g. llama-server). It never checks whether the user has actually supplied a model name via the `embedding_model` config string.

### Root cause

`src/core/ai/gateway.ts` lines 670-682 checks `tp.models.length === 0 && isUserProvided` but never inspects `parsed.modelId`. For `llama-server:text-embedding-qwen3-embedding-0.6b`, `parseModelId()` correctly extracts `modelId = "text-embedding-qwen3-embedding-0.6b"`, but the guard ignores it.

### Fix

Add `!parsed.modelId` to the condition so `user_provided_model_unset` only fires when the model name is genuinely absent.

Fixes #1672

### DIFF
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -671,7 +671,8 @@ export function diagnoseEmbedding(modelOverride?: string): EmbeddingDiagnosis {
   if (
     Array.isArray(tp.models) &&
     tp.models.length === 0 &&
-    (recipe.id === 'litellm' || isUserProvided)
+    (recipe.id === 'litellm' || isUserProvided) &&
+    !parsed.modelId
   ) {
     return {
       ok: false,


garrytan:

PR #1675, opened 2026-05-31
fix(chunker): cap oversized code chunks so they stay embeddable

lubosxyz:
## Problem

The code chunker splits large nodes via `splitLargeNode`, which only works when a node exposes a `body` with ≥2 named children. A node *without* one — a giant object/array literal, a single huge assignment, a massive template literal — falls through and is emitted **whole** (`code.ts` ~L666). On real source this produces a chunk far larger than the embedder's context window. The embedder then rejects it (`input exceeds context length` / times out) and the chunk is **never embedded** — it silently drops out of semantic search forever.

Concrete case that surfaced this: a 372 KB service file chunked into 113 chunks, **one of which was a single 281 KB (~70k-token) node**. That one chunk could not embed, so the whole symbol was unsearchable by vector. Across a few repos this left a long tail of large real source files (services, big UI components) unembedded.

## Fix

A final safety-net pass, `capOversizedChunks`, applied to both emit paths. Any chunk whose estimated tokens exceed a budget (default **2000**, configurable via the new `CodeChunkOptions.maxChunkTokens`) is recursively re-split with the existing `recursiveChunk`, falling back to a hard character split for pathological no-whitespace content (e.g. a minified one-liner) where word/line splitting can't get under budget. Chunks already under budget pass through untouched, so normal files are unaffected.

## Verification

```
heartbeat.ts (372 KB):  before 113 chunks, max 281407 chars (1 oversized)
                        after  188 chunks, max   6094 chars (0 oversized)
issues.ts   (227 KB):   after  162 chunks, max   7798 chars (0 oversized)
routines.ts ( 93 KB):   after   49 chunks, max   6089 chars (0 oversized)
small .d.ts file:       1 chunk, unchanged (no regression)
```

Default 2000 fits the smallest common embedder context (e.g. nomic-embed-text, 2048); larger-context models can raise it via `maxChunkTokens`.

### DIFF
--- a/src/core/chunkers/code.ts
+++ b/src/core/chunkers/code.ts
@@ -168,6 +168,15 @@ export interface CodeChunkOptions {
   largeChunkThresholdTokens?: number;
   fallbackChunkSizeWords?: number;
   fallbackOverlapWords?: number;
+  /**
+   * Hard upper bound (estimated tokens) on any single emitted chunk. A node
+   * the AST splitter can't break up (a giant object/array literal, a single
+   * huge assignment, a massive template literal) would otherwise be emitted
+   * whole and rejected by the embedder ("input exceeds context length").
+   * Chunks over this budget are recursively re-split. Default 2000 fits the
+   * smallest common embedder context (e.g. nomic-embed-text, 2048).
+   */
+  maxChunkTokens?: number;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


lubosxyz:
One implementation note for your consideration: existing brains won't re-chunk the affected files until `CHUNKER_VERSION` bumps (it gates the content-hash re-walk in `sync.ts`), since git HEAD for those files hasn't moved. I left it at 4 because bumping forces a full re-chunk for *all* users (the fix only changes output for the rare oversized node) — happy to bump it in this PR if you'd rather existing brains pick it up automatically on upgrade.

PR #1679, opened 2026-05-31
docs: clarify HTTP localOnly boundary

### LINKED ISSUE #1612: Docs: DEPLOY.md contradicts localOnly remote availability
Thank you for maintaining gbrain. I noticed a small docs mismatch around the MCP trust boundary.

In `docs/mcp/DEPLOY.md`, the `Scopes and localOnly` section says these operations are rejected over HTTP regardless of scope:

- `sync_brain`
- `file_upload`
- `file_list`
- `file_url`

That matches `src/commands/serve-http.ts`, where HTTP MCP tools are built from `operations.filter(op => !op.localOnly)`. It also matches `docs/mcp/CHATGPT.md`, which says the same four operations are rejected over HTTP.

Later in `docs/mcp/DEPLOY.md`, the `Operations` section says all operations are available remotely, including `sync_brain` and `file_upload`, and describes remote `file_upload` confinement. That looks stale relative to the current HTTP behavior.

Suggested fix: update or remove the later `Operations` paragraph so it says HTTP remote clients do not see `localOnly` operations, and point local users to stdio or CLI for those local filesystem surfaces.

I did not open a PR because I wanted to confirm the intended docs wording first, but this looks like a documentation-only correction.

TheLordArgus:
## Summary

Fixes #1612.

`docs/mcp/DEPLOY.md` already states in the Scopes section that HTTP rejects the four `localOnly` operations regardless of scope, which matches `serve-http.ts` filtering `operations.filter(op => !op.localOnly)` and `docs/mcp/CHATGPT.md`.

Later in the same deploy doc, the Operations section still said all operations are remote, including `sync_brain` and `file_upload`, and described remote `file_upload` confinement. This updates that stale section so the doc consistently says:

- HTTP exposes scoped, non-`localOnly` operations
- `sync_brain`, `file_upload`, `file_list`, and `file_url` are hidden from remote MCP clients even with `admin`
- local stdio/CLI remains the path for local filesystem surfaces

I also changed the scope table's `admin` row from "local-only ops" to "remote-safe admin ops" so it no longer contradicts the fail-closed HTTP boundary.

## Validation

- `bun test test/config-set.test.ts`
- `git diff --check`

### DIFF
--- a/docs/mcp/DEPLOY.md
+++ b/docs/mcp/DEPLOY.md
@@ -184,7 +184,7 @@ filesystem surface area.
 |-------|---------------|
 | `read` | `search`, `query`, `get_page`, `list_pages`, graph traversal |
 | `write` | `put_page`, `delete_page`, `add_link`, `add_timeline_entry` |
-| `admin` | Client management, token revocation, sweep, local-only ops |
+| `admin` | Client management, token revocation, sweep, remote-safe admin ops |
 
 ## Legacy Bearer Token Setup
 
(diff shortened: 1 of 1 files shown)


time-attack:
Merging — this isn't cosmetic doc polish, it's a doc that actively misstates the remote trust boundary. Verified on master that `serve-http.ts` filters `!op.localOnly` and that `file_upload`/`file_list`/`file_url` are all `localOnly: true`, so the Operations section promising them to remote clients (and describing confinement rules that can never apply) is simply wrong. One note for a follow-up if you're inclined: the localOnly set is actually broader than the four ops you name — `purge_deleted_pages`, `get_recent_transcripts`, `code_traversal_cache_clear`, and `chronicle_backfill` are also hidden — so the general rule you lead with is the durable phrasing.

PR #1680, opened 2026-05-31
fix: allow gateway loop config flag

### LINKED ISSUE #1454: doctor advertises agent.use_gateway_loop=true as fix, but config CLI rejects it
## Symptom
Doctor's `subagent_capability` warn message recommends `gbrain config set agent.use_gateway_loop true` as alternative to setting ANTHROPIC_API_KEY. The config CLI rejects it:

```
$ gbrain config set agent.use_gateway_loop true
[config] Unknown config key "agent.use_gateway_loop".
```

## Diagnosis
Either:
- The key name has drifted from what doctor advertises
- The config CLI needs `--force` to accept new keys (then doctor should say so)
- The feature was renamed/removed but doctor message wasn't updated

## Suggested fix
Either restore the config key OR update doctor's recommendation to match the current config surface.

TheLordArgus:
## Summary
- add `agent.use_gateway_loop` to the known config key registry
- cover the key in the config-set gate regression tests

Closes #1454

## Test plan
- `bun test test/config-set.test.ts`
- `bun run typecheck`
- `scripts/check-test-isolation.sh`
- `git diff --check`

### DIFF
--- a/src/core/config.ts
+++ b/src/core/config.ts
@@ -673,6 +673,8 @@ export const KNOWN_CONFIG_KEYS: readonly string[] = [
   'models.expansion',
   'models.chat',
   'models.eval.longmemeval',
+  // Agent feature flags
+  'agent.use_gateway_loop',
   'facts.extraction_model',
   // Dream cycle config
   'dream.synthesize.session_corpus_dir',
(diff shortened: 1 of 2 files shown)


time-attack:
Closing: current master already contains this fix. Both changes in the PR already exist verbatim on master: the registry entry at src/core/config.ts:872 and the config-set regression test at test/config-set.test.ts:49-51. The PR (Closes #1454) is fully redundant; #1454 itself is also fixed.

PR #1681, opened 2026-05-31
fix(eval-longmemeval): route answer-gen + extractor through the gateway (supersedes #1401)

billy-armstrong:
## Problem

`gbrain eval longmemeval` constructed `new Anthropic()` and passed
`resolveModel()`'s output straight to `.messages.create({ model })`.
`resolveModel` returns a **provider-prefixed** id (e.g.
`anthropic:claude-sonnet-4-6` from the `sonnet` fallback / `DEFAULT_ALIASES`).
The Anthropic SDK only accepts a bare model id, so it 404s on the prefixed
form — every question fails unless a bare id is forced via `--model`. The
trajectory extractor (`tier: 'utility'` → `anthropic:claude-haiku-4-5-...`)
hits the same wall.

`stripProviderPrefix` exists upstream for the subagent path (`subagent.ts`)
but was never applied at these SDK-direct call sites.

## Fix

Route both the answer-gen and the trajectory extractor through the **gateway**
(`tryBuildGatewayClient`, the adapter `think` has used since v0.35.5.0) instead
of constructing `new Anthropic()` directly. The gateway:

- normalizes the model id — a bare `claude-sonnet-4-6` **and** a prefixed
  `anthropic:claude-sonnet-4-6` both resolve;
- owns auth + provider routing.

So the 404 class is fixed at the root for every provider, not stripped at one
call site.

Behavioral details:
- **Answer-gen**: hard-fails with a clear, actionable stderr message when no
  chat model is available (was an opaque per-question SDK throw).
- **Extractor**: a null gateway client (no key / unknown provider for the
  extractor model) disables trajectory routing for the run rather than failing
  the whole eval — trajectory is optional enrichment.
- Tests still inject via `runOpts.client` / `runOpts.extractorClient`,
  short-circuiting the gateway path.

## Why this supersedes #1401

#1401 applied `stripProviderPrefix` at the longmemeval call sites — a narrow
workaround. This routes through the canonical gateway seam instead, which is how
every other LLM call site in gbrain already works (chat, expansion, synthesize,
think). It fixes the root cause and aligns longmemeval with the rest of the
codebase. Recommend closing #1401 in favor of this.

## Testing

- `test/longmemeval-gateway-routing.test.ts` (new, 3 cases) exercises the
  **default** (non-injected) path with a stubbed gateway transport:
  - answer-gen sends the prefixed `anthropic:claude-sonnet-4-6` through
    `gateway.chat` (the exact id the SDK 404'd on);
  - a bare `--model claude-sonnet-4-6` normalizes to the same prefixed form;
  - **trajectory ON** (the default the nightly probe runs): extractor sends a
    prefixed `anthropic:claude-haiku-4-5-...` AND answer-gen sends prefixed
    sonnet, and the stubbed answer round-trips into the output `hypothesis`.
- Existing `longmemeval-*` tests pass unchanged (the `runOpts.client` injection
  short-circuit is intact).
- `tsc --noEmit` clean.
- **Verified end-to-end**: a live `gbrain eval longmemeval --limit 1` with NO
  `--model` (resolves to the prefixed `anthropic:claude-sonnet-4-6` — the exact
  id the old `new Anthropic()` path 404'd on) completed with `0 errors` and
  returned a valid answer against the real Anthropic API.

### DIFF
--- a/src/commands/eval-longmemeval.ts
+++ b/src/commands/eval-longmemeval.ts
@@ -10,7 +10,6 @@
  */
 
 import { readFileSync, existsSync, openSync, writeSync, closeSync, writeFileSync } from 'fs';
-import Anthropic from '@anthropic-ai/sdk';
 import { withBenchmarkBrain, resetTables } from '../eval/longmemeval/harness.ts';
 import { haystackToPages, type LongMemEvalQuestion } from '../eval/longmemeval/adapter.ts';
 import { renderChatBlock, type ChatSessionForPrompt } from '../eval/longmemeval/sanitize.ts';
@@ -18,7 +17,7 @@ import { importFromContent } from '../core/import-file.ts';
 import { hybridSearch } from '../core/search/hybrid.ts';
 import { expandQuery } from '../core/search/expansion.ts';
 import { resolveModel } from '../core/model-config.ts';
-import type { ThinkLLMClient } from '../core/think/index.ts';
+import { __thinkAdapter, type ThinkLLMClient } from '../core/think/index.ts';
 import { createProgress } from '../core/progress.ts';
 import { getCliOptions, cliOptsToProgressOptions } from '../core/cli-options.ts';
 import type { PGLiteEngine } from '../core/pglite-engine.ts';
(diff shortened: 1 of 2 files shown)


billy-armstrong:
Closing for queue hygiene — no review activity since opening (2026-05-31). The gap it addressed still exists as of v0.42.40.0 (longmemeval answer-gen/extractor don't route through the gateway), so feel free to reopen or cherry-pick if it becomes relevant.

PR #1684, opened 2026-05-31
fix(engine): batch upsertChunks INSERT under the int16 Bind param ceiling

alexhawkins:
## Problem

`upsertChunks` can **silently lose every chunk of a page** with no error thrown.

PGLite binds a statement's parameters through a **signed int16**, so a single multi-row `INSERT` silently inserts **0 rows** once it exceeds **32767** bound params. Verified empirically against the real `PGLiteEngine` (in-memory):

| bound params | rows actually stored |
| ------------ | -------------------- |
| 32760        | all                  |
| 32775        | **0 (no error)**     |

`upsertChunks` built one unbounded multi-row `INSERT` at ~15 params/row, so any page that chunks past **~2184 rows** silently dropped **all** of its chunks. Large generated files hit this in normal indexing, and at higher counts it escalated to the server-side `bind message has N parameter formats but 0 parameters` error, which aborts the surrounding transaction and (in a sync loop) pins the git anchor so every run re-imports from scratch.

## Fix

Split the multi-row `INSERT` into `<=30000`-param batches (comfortable margin under 32767) by extracting the per-row build + `INSERT ... ON CONFLICT` into a new `_insertChunkBatch` helper. Applied to **both** `pglite-engine.ts` and `postgres-engine.ts` for parity (postgres.js rejects loudly at 65534 rather than silently truncating, but shares the same conservative budget).

The orphan-prune `DELETE` and the empty-chunks early `return` stay in `_upsertChunksOnce`, so they still run **exactly once** per call -- only the `INSERT` is batched.

## Test

Adds a regression test that upserts **2500 text chunks** (~37500 params, forcing more than one batch) and asserts all are stored with contiguous content across the batch boundary.

- **Fails** on the pre-fix engine (stores 0 rows).
- **Passes** after the fix.

## Verification

- `npx tsc --noEmit` -> 0 errors.
- New regression test passes; verified it fails when the engine fix is reverted but the test is kept.
- The 4 pre-existing `pglite-engine.test.ts` failures (search_vector trigger / stats) are present on `master` before this change and are unrelated.

### DIFF
--- a/src/core/pglite-engine.ts
+++ b/src/core/pglite-engine.ts
@@ -1983,6 +1983,36 @@ export class PGLiteEngine implements BrainEngine {
     // (1024-dim Voyage). Text/code chunks pass embedding=Float32Array +
     // embedding_image=null. Default modality='text' when omitted.
     const cols = '(page_id, chunk_index, chunk_text, chunk_source, embedding, model, token_count, embedded_at, language, symbol_name, symbol_type, start_line, end_line, parent_symbol_path, doc_comment, symbol_name_qualified, modality, embedding_image)';
+
+    // Batch the multi-row INSERT to stay under the Bind parameter ceiling. The
+    // PGLite engine binds params through a SIGNED int16, so a single INSERT
+    // silently no-ops (0 rows, NO error thrown) above 32767 params; the
+    // generated __generated__/graphql.ts chunked past that and pinned
+    // sync.last_commit. Both engines share this conservative budget. (Fix 6)
+    const MAX_BIND_PARAMS = 30000;
+    let batch: ChunkInput[] = [];
+    let batchParams = 0;
+    for (const chunk of chunks) {
+      // Base row = 15 params + 1 per present embedding (text / image vector).
+      const rowParamCount =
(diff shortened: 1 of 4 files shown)


alexhawkins:
Added a Postgres-side E2E regression and verified the fix against a real Postgres (`pgvector/pgvector:pg16`):

- **Postgres engine suite** (`engine-parity` + `embedding-column-postgres`): **18 pass / 0 fail** — the batching refactor produces identical search/provenance results to PGLite.
- **New Postgres regression** (`test/e2e/upsertchunks-bind-ceiling-postgres.test.ts`): upserts 5000 text chunks (~75000 params, past postgres.js's hard 65534 cap).
  - On the pre-fix engine it throws `MAX_PARAMETERS_EXCEEDED: Max number of parameters (65534) exceeded`.
  - After the fix it stores all 5000 with contiguous content across batch boundaries.

So both engines now have a regression test that fails before the fix and passes after — PGLite (silent loss at 32767) and Postgres (loud rejection at 65534). `tsc --noEmit`: 0 errors.

garrytan:

Issue #1178, opened 2026-05-18
bug: migration v66 (embed_stale_partial_index) fails with 'DROP INDEX CONCURRENTLY cannot be executed from a function' when invalid index exists from prior failed run


ethanbeard:
## Summary

Migration `v66 / embed_stale_partial_index` in `src/core/migrate.ts` wraps `DROP INDEX CONCURRENTLY IF EXISTS idx_chunks_embedding_null` inside a `DO $$ ... EXECUTE '...' END $$` block. Postgres rejects `CONCURRENTLY` inside *any* function context — including statements run via `EXECUTE` in a `DO` block. So when the EXISTS guard triggers (because a prior `CREATE INDEX CONCURRENTLY` left an `indisvalid = false` index), the migration fails:

```
DROP INDEX CONCURRENTLY cannot be executed from a function
```

The migration completes fine on brains that never had a failed `CREATE INDEX CONCURRENTLY` for that index name — the EXISTS guard returns false and `EXECUTE` is never called. The bug only manifests for brains carrying invalid-index leftovers.

## Repro

On a Postgres-backed brain:

```sql
-- Create an invalid leftover (simulates a prior failed CREATE INDEX CONCURRENTLY)
CREATE INDEX idx_chunks_embedding_null ON content_chunks (page_id, chunk_index) WHERE embedding IS NULL;
UPDATE pg_index SET indisvalid = false WHERE indexrelid = 'idx_chunks_embedding_null'::regclass;
```

Then run `gbrain init --migrate-only` (with schema version <= 65). Migration v66 fails.

I hit this organically while upgrading from v0.32.0 → v0.35.8.0 on a Supabase-backed brain — the invalid index was already present (presumably from a much earlier failed migration attempt). Workaround was:

```sql
DROP INDEX IF EXISTS idx_chunks_embedding_null;
```

then re-run `gbrain init --migrate-only`. Worked cleanly.

## Why the current shape doesn't work

[...]

PR #1686, opened 2026-05-31
fix(migrations): avoid concurrent index drop inside DO blocks

shandutta:
## Summary
- use plain `DROP INDEX IF EXISTS` for invalid concurrent-index cleanup inside migration DO blocks
- keep `CREATE INDEX CONCURRENTLY` for Postgres index creation outside a transaction
- update migration structural tests to assert transaction-safe cleanup

## Test plan
- `bun test test/migrate.test.ts test/migrate-retry.test.ts test/migrations-registry.test.ts`

### DIFF
--- a/src/core/migrate.ts
+++ b/src/core/migrate.ts
@@ -505,7 +505,7 @@ export const MIGRATIONS: Migration[] = [
                JOIN pg_class c ON c.oid = i.indexrelid
                WHERE c.relname = 'idx_pages_updated_at_desc' AND NOT i.indisvalid
              ) THEN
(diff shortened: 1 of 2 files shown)


time-attack:
Thanks — merging this. The DO-block `EXECUTE 'DROP INDEX CONCURRENTLY ...'` sites are a genuine latent migration abort: Postgres rejects CONCURRENTLY from any function/EXECUTE context, which `migrate.ts`'s own header comment (from #1178) already documents, and the `dropInvalidIndex` helper already handles correctly. Plain `DROP INDEX IF EXISTS` is the right call inside the guard since it only ever fires against an already-invalid index remnant. Please rebase on master — one additional site (`pages_links_extracted_at_idx`) landed after you opened this and needs the same treatment.

PR #1690, opened 2026-05-31
fix(extract): parse simple timeline bullets

shandutta:
## Summary

Many hand-authored brain pages use the lightweight form `- YYYY-MM-DD: Summary` under `## Timeline` headings, but `extractTimelineFromContent` only handled the pipe-delimited `- **YYYY-MM-DD** | Source — Summary` format (Format 1) and the heading format (Format 2). The simple bullet form was silently dropped during extraction.

This adds a Format 1b regex that captures `- YYYY-MM-DD: Summary` entries with `source: 'markdown'`, filling the coverage gap without affecting existing extraction paths.

## Changes

- `src/commands/extract.ts`: adds `simpleBulletPattern` regex after Format 1 extraction
- `test/extract.test.ts`: 2 new test cases covering the new format

## Test plan

- [x] `bun test test/extract.test.ts` — 22 pass, 0 fail
- [x] Existing timeline extraction tests unaffected
- [x] New cases verify both date parsing and `source: 'markdown'` attribution

### DIFF
--- a/src/commands/extract.ts
+++ b/src/commands/extract.ts
@@ -327,6 +327,15 @@ export function extractTimelineFromContent(content: string, slug: string): Extra
     entries.push({ slug, date: match[1], source: match[2].trim(), summary: match[3].trim() });
   }
 
+  // Format 1b: Simple timeline bullet — - YYYY-MM-DD: Summary
+  // Many hand-authored brain pages use this lightweight form under
+  // "## Timeline"; treating it as timeline data fixes coverage without
+  // forcing noisy rewrites into the heavier source-delimited format.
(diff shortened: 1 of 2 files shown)


time-attack:
Thanks for the clean patch and the tests — the format-coverage observation is fair. I'm closing it as new extraction surface rather than a fix, though: `extractTimelineFromContent` has always implemented exactly the two documented formats, so `- YYYY-MM-DD: Summary` isn't being dropped so much as never having been a supported shape. My concrete worry is false positives — a bare dated-bullet regex will also match changelog lines, meeting-note bullets, and dated task items anywhere on a page, and spurious timeline entries are painful to unwind once they're in a brain. Timeline extraction formats are a surface we're keeping deliberately narrow.

PR #1691, opened 2026-05-31
fix: mark clean local imports as synced

MonteMacMini:
## Summary
- Mark clean source-routed directory imports as synced by updating `sources.last_sync_at`
- Preserve failure visibility by not advancing freshness when import failures occur
- Extend directory walker pruning to skip common generated/dependency directories (`venv`, `.venv`, `__pycache__`, `dist`, `build`)

## Why
Non-git local folder sources can be refreshed with `gbrain import <path> --source-id <id> --no-embed --fresh`, but previously those imports populated pages without updating source freshness metadata. That left valid local sources stuck at `LAST SYNC: never` even when their content was current.

This makes a clean source-routed import a valid freshness event for local-folder sources, while still refusing to hide partial import failures.

## Test Plan
- [x] `bun run typecheck`
- [x] Verified locally with a non-git source: `gbrain import /Users/clawagents/hermes-data/Squeeze-Scanner --source-id squeeze-scanner --no-embed --fresh --json`
- [x] Confirmed `gbrain sources status` advanced `squeeze-scanner` from `never` to a real `LAST SYNC` timestamp
- [x] Ran Mac Mini hourly GBrain runner and confirmed `gbrain doctor` reports `sync_freshness: ok` for all federated sources

### DIFF
--- a/src/commands/import.ts
+++ b/src/commands/import.ts
@@ -425,6 +425,26 @@ export async function runImport(
     summary: `Imported ${imported} pages, ${skipped} skipped, ${chunksCreated} chunks`,
   });
 
+  // Local-folder source freshness: a successful source-routed import is a
+  // real source sync even when the directory is not a git repo. Previously
+  // only git-backed imports/syncs advanced sources.last_sync_at, so valid
+  // local project sources stayed stuck at "never synced" forever despite
+  // their pages being current. Advance the source freshness bookmark only on
+  // a clean run; per-file failures should remain visible in doctor/status.
+  if (sourceId) {
+    if (failures.length === 0) {
+      await engine.executeRaw(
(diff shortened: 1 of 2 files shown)


time-attack:
Thanks for this, and apologies for the slow response — the situation has moved. The directory-pruning half is already on master: `PRUNE_DIR_NAMES` in `src/core/sync.ts` now carries `node_modules`, `vendor`, `dist`, `build`, and `venv` via #2020/#1483. The blocking problem is that your branch still has `'ops'` in that set and in `collectSyncableFiles`; master deliberately removed it in #2404 because matching the bare segment pruned every user `ops/` directory at any depth and silently deleted `ops/*` pages, so landing this would reintroduce a data-loss bug. The `sources.last_sync_at` gap for non-git local sources is real and still unfixed — if you want to open a fresh, focused PR against current master with just that change (and no `PRUNE_DIR_NAMES` edits), that one has a clear story.

PR #1694, opened 2026-05-31
perf(config): batch + cache engine.getConfig to kill ~85 round-trips per query

### LINKED ISSUE #1693: perf(config): engine.getConfig issues ~85 round-trips per query; batch+cache to fix >5s latency + empty-stdout on remote Postgres
## Summary

On a **remote Postgres / Supabase pooler**, a single `gbrain query` / `gbrain search` issues **~85 separate `SELECT value FROM config WHERE key = $1` round-trips**. Each is a network round-trip; serial, they dominate query latency (~5s) and push the op handler past `cli.ts`'s hard-coded **10s disconnect force-exit timer** (`DISCONNECT_HARD_DEADLINE_MS`), which fires `process.exit(0)` **mid-handler** → **empty stdout** + a misleading `engine.disconnect() did not return within 10000ms` line (disconnect was never reached).

Invisible on local PGLite (sub-ms reads); only manifests on remote DBs.

## Evidence (gbrain 0.41.38.0, Supabase pooler, 2486-page brain)

Instrumented the engine (Proxy timing every method) during one `hybridSearch`:
- `getConfig()` called **~85×** (`loadConfigWithEngine` reads 15 keys and is invoked 2× per query; mode/cache/intent/rerank/graph-signals resolvers add more), each ~165ms on the pooler.
- Per-stage SQL is fast: `searchVector` 0.75s, `searchKeyword` 0.6s, `embed()` 0.31s — but full `hybridSearch` ~7–8s. The delta is config N+1.

Validated fix (preload whole `config` table in ONE query, serve `getConfig` from a Map):
```
preloaded 11 config rows in ONE query
handler returned 5 results in 3.71s (getConfig served from cache 85x)
```
Handler **~8s → 3.7s**; `gbrain query` end-to-end ~11s → ~5s, removing the force-exit/empty-stdout failure mode.

## Proposed fix

Process-lifetime config cache on `PostgresEngine.getConfig()`: one batch `SELECT key, value FROM config` on first read → Map; write-through on `setConfig`/`unsetConfig`; coarse TTL (`GBRAIN_CONFIG_CACHE_TTL_MS`, default 30s, 0 disables). Fixes all 56 `getConfig` callers. Precedent already exists in `query-cache.ts` (`WHERE key = ANY($1)`).

## Secondary (related) — force-exit timer scope

`cli.ts` arms the unref'd 10s force-exit timer **before** the op handler, so a legitimately slow remote handler (>10s) is killed mid-flight with an empty result. Consider arming it only around `engine.disconnect()` (its documented purpose), or making the deadline env-configurable, so a slow-but-correct handler isn't truncated.

Patch available; happy to open a PR.

Omerbahari:
Fixes #1693.

## Problem

A single `gbrain query <q>` / `gbrain search <q>` issues **~85 separate `SELECT value FROM config WHERE key = $1` round-trips**. `loadConfigWithEngine` reads 15 keys and is called twice per query; the mode/cache/intent/rerank/graph-signals resolvers each re-read more.

On a local PGLite brain each read is sub-millisecond, so this is invisible. On a **remote Postgres / Supabase pooler** each read is a network round-trip (amplified under transaction-mode `prepare:false`, where every statement re-parses). 85 serial reads = several seconds of pure config latency on every search.

Two user-visible symptoms downstream:

1. **Empty stdout + misleading force-exit.** The CLI's hard-coded 10s disconnect force-exit timer (`cli.ts` `DISCONNECT_HARD_DEADLINE_MS`) is armed *before* the op handler. When the config-read tax pushes the handler past 10s, the unref'd timer fires `process.exit(0)` **mid-handler** → empty stdout + a misleading `engine.disconnect() did not return within 10000ms` line (disconnect was never even reached).
2. **Latency dominated by config, not retrieval.** Measured on a 2486-page brain: `searchVector` 0.75s, `searchKeyword` 0.6s, `embed` 0.31s — but full `hybridSearch` ~7–8s. The delta is config N+1.

## Fix

Add a process-lifetime config cache to `PostgresEngine.getConfig()`:

- First read triggers **one** batch `SELECT key, value FROM config` that populates a `Map`; subsequent `getConfig()` calls are served from memory.
- `setConfig()` / `unsetConfig()` update the Map in place (write-through), so a long-lived process (`serve` / autopilot) never serves stale config.
- A coarse TTL (default 30s, env `GBRAIN_CONFIG_CACHE_TTL_MS`) bounds staleness for multi-writer setups; set `0` to disable caching entirely and restore the original per-key query.

`Map.has` distinguishes "known-absent" (cached null) from "not yet loaded", so a missing key is not re-fetched on every call.

## Result (Supabase session-mode pooler, 2486-page brain)

| Metric | Before | After |
|---|---|---|
| getConfig round-trips / query | ~85 | 1 |
| hybridSearch handler | ~8s | ~3.7s |
| `gbrain query` end-to-end | ~11s | ~5s |

This brings the handler comfortably under the 10s force-exit deadline, removing the empty-stdout failure mode.

## Precedent

`query-cache.ts` already batch-reads config via `WHERE key = ANY($1)`; this PR generalizes that pattern to the hot `getConfig` path.

## Notes

- Behaviour is unchanged with `GBRAIN_CONFIG_CACHE_TTL_MS=0` (falls back to the original single-key query).
- Write-through keeps `serve`/autopilot processes correct without waiting for the TTL.

### DIFF
--- a/src/core/postgres-engine.ts
+++ b/src/core/postgres-engine.ts
@@ -91,6 +91,25 @@ export function getPostgresSchema(
 export class PostgresEngine implements BrainEngine {
   readonly kind = 'postgres' as const;
   private _sql: ReturnType<typeof postgres> | null = null;
+  /**
+   * perf: process-lifetime config cache. A single search fires ~85
+   * getConfig() reads (loadConfigWithEngine x2 = 30 keys, plus mode/cache/
+   * intent/rerank/graph-signals resolvers). On a remote pooler each read is
+   * a round-trip; serial they dominate query latency (~5s) and can push the
+   * op handler past cli.ts's 10s disconnect force-exit, truncating stdout.
+   * First read batch-loads the whole `config` table into this Map; writes
+   * update it in place. TTL bounds staleness for multi-writer processes.
+   */
+  private _configCache: Map<string, string | null> | null = null;
+  private _configCacheLoadedAt = 0;
+  private get _configCacheTtlMs(): number {
+    const raw = process.env.GBRAIN_CONFIG_CACHE_TTL_MS;
+    if (raw !== undefined) {
+      const n = parseInt(raw, 10);
+      if (Number.isFinite(n) && n >= 0) return n;
+    }
+    return 30_000;
+  }
   /** Saved config for reconnection. */
   private _savedConfig: (EngineConfig & { poolSize?: number; parentConnectionManager?: ConnectionManager }) | null = null;
   /** Whether a reconnect is in progress (prevents concurrent reconnects). */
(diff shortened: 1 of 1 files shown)


time-attack:
Thank you for the careful measurement work — the config N+1 in `loadConfigWithEngine` is a real cost on a remote pooler and your latency breakdown is useful. I'm not taking this shape, though. A process-lifetime cache in the engine introduces a 30s staleness window where a long-lived MCP server keeps serving old config after another process runs `gbrain config set`, and it lands only on the Postgres side, which breaks the engine-parity invariant that `engine-parity.test.ts` pins. Master has also moved: `getConfig` is now wrapped in `connRetry` (#1603, which addressed the empty-stdout symptom you cite), and `query-cache.ts` already batches the hot path with `WHERE key = ANY($1)`. Batching inside `loadConfigWithEngine`'s `dbBool`/`dbStr` helpers would attack the same N+1 with no cache-coherency surface at all.

PR #1703, opened 2026-05-31
fix(links): exclude soft-deleted endpoints from getLinks/getBacklinks (closes #1702, sibling of #1021)

### LINKED ISSUE #1702: get_links / get_backlinks return links for soft-deleted endpoints (sibling of #1021)
## Bug Report

**gbrain version:** 0.41.38.0 (master)
**Engine:** PGLite (also reproducible on Postgres by code inspection)

### Problem

\`getLinks\` and \`getBacklinks\` do not filter \`pages.deleted_at IS NULL\` on the JOIN to pages, so during the 72h soft-delete recovery window:

- A page that \`delete_page\` soft-deleted still appears in \`get_backlinks(target)\` for any page it linked to
- A page whose link target was soft-deleted still appears in \`get_links(source)\`

This is the same class of bug as #1021 (\"orphans: soft-deleted pages still counted in orphan scan\"), fixed by PR #1033. PR #1033 only patched orphans + the \`dead_links\` counter in \`get_health\`; the read-side \`getLinks\` / \`getBacklinks\` queries in both \`src/core/pglite-engine.ts\` and \`src/core/postgres-engine.ts\` still need the same filter.

### Reproduction

\`\`\`bash
# In an MCP session (or via gbrain CLI):
gbrain put inbox/test-orphan-link <<'MD'
---
title: "Test"
type: note
---
Linking [[concepts/some-target]] for repro.
MD

gbrain delete inbox/test-orphan-link
# Page is soft-deleted (recoverable for 72h)
gbrain backlinks concepts/some-target
# Expected: empty (or no row from inbox/test-orphan-link)
# Actual:   row still present, naming the deleted source
\`\`\`

Or via MCP:

\`\`\`json
// 1. put_page with a wikilink
{ \"method\": \"tools/call\", \"params\": { \"name\": \"put_page\",
  \"arguments\": { \"slug\": \"inbox/test\", \"content\": \"... [[concepts/x]] ...\" }}}

// 2. delete_page (= softDeletePage)
{ \"method\": \"tools/call\", \"params\": { \"name\": \"delete_page\",
  \"arguments\": { \"slug\": \"inbox/test\" }}}

// 3. get_backlinks on the target
{ \"method\": \"tools/call\", \"params\": { \"name\": \"get_backlinks\",
  \"arguments\": { \"slug\": \"concepts/x\" }}}
// Returns inbox/test even though the source page is soft-deleted.
\`\`\`

### Expected Behavior

\`get_links\` and \`get_backlinks\` should hide rows whose endpoint pages are soft-deleted, matching the contract already enforced by \`get_page\` / \`list_pages\` / \`search\`. \`restore_page\` should bring them back into view. Hard delete (autopilot \`purgeDeletedPages\`) is already correct because the FK on \`links\` has \`ON DELETE CASCADE\`.

### Schema confirms this is a bug, not by design

\`src/schema.sql\` lines 401-402:

\`\`\`sql
CREATE TABLE IF NOT EXISTS links (
  ...
  from_page_id INTEGER NOT NULL REFERENCES pages(id) ON DELETE CASCADE,
  to_page_id   INTEGER NOT NULL REFERENCES pages(id) ON DELETE CASCADE,
  ...
);
\`\`\`

And the \`purgeDeletedPages\` comment at \`src/core/postgres-engine.ts:1004\`:

> The DELETE cascades through content_chunks, page_links, chunk_relations via FKs.

The schema author clearly intended links to disappear when their endpoint page does — that contract is satisfied on hard delete via FK cascade, but the soft-delete read path leaks them.

### Impact

- \`get_backlinks\` reports stale \"who links to me\" for 72h after a delete
- \`get_links\` reports outbound links to invisible targets
- Graph / recommendation surfaces that consume these queries inherit the staleness
- For agents that programmatically create-then-delete pages over MCP (test harnesses, scratch pages, autopilot-cycle subagent writes), the link table appears to accumulate orphans even though the underlying schema is correct

### Affected sites

By code inspection, both queries have the same omission in both engines:

- \`src/core/pglite-engine.ts\` — \`getLinks\` (line ~2430), \`getBacklinks\` (line ~2464)
- \`src/core/postgres-engine.ts\` — \`getLinks\` (line ~2477), \`getBacklinks\` (line ~2508)

The graph-traversal queries (\`traverseGraph\`, the v0.20.0 graph-query path) JOIN pages on \`l.to_page_id\` for the recursive step too — those probably want the same filter for consistency, but they are out of scope for this issue/PR to keep the change reviewable.

### Patch

PR coming separately. Mirrors PR #1033's pattern: add \`AND f.deleted_at IS NULL AND t.deleted_at IS NULL\` to the page JOINs in both branches (with and without \`opts.sourceId\`) of both functions in both engines. New regression test \`test/links-soft-deleted.test.ts\` covers backlinks + links + restore.

### Workaround

Wait 72h for autopilot \`purgeDeletedPages\` to hard-delete soft-deleted pages and let the FK cascade clean up the links. Or manually \`remove_link\` after every \`delete_page\`.

starpincer:
Closes #1702. Sibling of #1021 / PR #1033.

## Problem

`getLinks` and `getBacklinks` JOIN `pages` on the link FKs but do not filter `pages.deleted_at IS NULL`. During the 72h soft-delete recovery window, links to or from a soft-deleted page still appear in the read-side results even though the page itself is hidden from `get_page` / `list_pages` / `search`.

This is the same class of bug PR #1033 fixed for `orphans` and `dead_links`. #1033 didn't touch `getLinks` / `getBacklinks`, so the read-side leak remained.

`delete_page` (MCP op) maps to `softDeletePage`, so any agent that creates and then deletes a page over MCP leaves apparent links behind for 72h, until the autopilot `purgeDeletedPages` hard-deletes the rows and the FK on `links` (`ON DELETE CASCADE`) cleans up. The schema clearly intends links to disappear when their endpoint does — that contract is met on hard delete, not on soft delete.

## Change

Add `AND f.deleted_at IS NULL AND t.deleted_at IS NULL` to the page JOINs in both branches (with and without `opts.sourceId`) of:

- `PGLiteEngine.getLinks`
- `PGLiteEngine.getBacklinks`
- `PostgresEngine.getLinks`
- `PostgresEngine.getBacklinks`

Filtering both endpoints (not just the from-side, which is the visible repro) keeps the read contract symmetric and defends against the case where a target is soft-deleted but a link is queried via the source.

## Why this shape

Followed PR #1033's pattern:
- Same query-level fix (no schema migration, no behavior change for hard delete)
- Same defense-in-depth philosophy (filter both endpoints)
- Same test scaffolding (engine-injected `bun:test`, `softDeletePage` + assertion on the read path)

The graph traversal queries (`traverseGraph`, the recursive `JOIN pages p2 ON p2.id = l.to_page_id` paths) have the same omission, but they are out of scope for this PR to keep the diff reviewable. Happy to fold those into a follow-up.

## Tests

New `test/links-soft-deleted.test.ts` (5 cases):

- getBacklinks hides links from soft-deleted source
- getBacklinks hides links targeting soft-deleted page (defense-in-depth)
- getLinks hides outbound row when target is soft-deleted
- getLinks hides outbound row when source itself is soft-deleted
- `restore_page` brings the row back into view

Existing `test/orphans.test.ts` (57 cases across the related suites) and `test/backlinks.test.ts` continue to pass.

\`\`\`
$ bun test test/links-soft-deleted.test.ts
 5 pass | 0 fail | 9 expect() calls

$ bun test test/orphans.test.ts test/backlinks.test.ts test/reconcile-links.serial.test.ts
 57 pass | 0 fail | 93 expect() calls
\`\`\`

## Compat

- No schema migration
- No public API / op shape change
- Hard delete behavior unchanged (FK cascade still applies)
- `restore_page` semantics preserved (un-hiding a soft-deleted page brings its links back)
- Behavior change is limited to: soft-deleted-but-not-yet-purged endpoints no longer leak into read results

## Discovered while

Building a multi-host gbrain topology and noticing residual backlinks after MCP put_page + delete_page round-trips in a JavanC fork. Wrote up the topology + the bug-class lineage in #1692 (separate JavanC opt-in patch for `GBRAIN_TRUSTED_REMOTE`) before tracking this one down.

### DIFF
--- a/src/core/pglite-engine.ts
+++ b/src/core/pglite-engine.ts
@@ -2433,14 +2433,21 @@ export class PGLiteEngine implements BrainEngine {
     // and read-side op handlers that haven't threaded sourceId yet). With
     // opts.sourceId, scope to that source — used by reconcileLinks and any
     // ctx.sourceId-aware read op (D20).
+    //
+    // Soft-delete filter (siblings of #1021/PR #1033): hide rows where either
+    // endpoint page is soft-deleted. Without this, links survive the 72h
+    // recovery window in read-side results even though the endpoint pages are
+    // hidden from get_page / list_pages / search. Hard purge
+    // (purgeDeletedPages) still cascades via FK; this filter only governs the
+    // soft-delete window.
     if (opts?.sourceId) {
       const { rows } = await this.db.query(
         `SELECT f.slug as from_slug, t.slug as to_slug,
                 l.link_type, l.context, l.link_source,
                 o.slug as origin_slug, l.origin_field
          FROM links l
-         JOIN pages f ON f.id = l.from_page_id
-         JOIN pages t ON t.id = l.to_page_id
+         JOIN pages f ON f.id = l.from_page_id AND f.deleted_at IS NULL
+         JOIN pages t ON t.id = l.to_page_id   AND t.deleted_at IS NULL
          LEFT JOIN pages o ON o.id = l.origin_page_id
          WHERE f.slug = $1 AND f.source_id = $2`,
         [slug, opts.sourceId]
@@ -2452,8 +2459,8 @@ export class PGLiteEngine implements BrainEngine {
               l.link_type, l.context, l.link_source,
               o.slug as origin_slug, l.origin_field
        FROM links l
-       JOIN pages f ON f.id = l.from_page_id
-       JOIN pages t ON t.id = l.to_page_id
+       JOIN pages f ON f.id = l.from_page_id AND f.deleted_at IS NULL
+       JOIN pages t ON t.id = l.to_page_id   AND t.deleted_at IS NULL
        LEFT JOIN pages o ON o.id = l.origin_page_id
        WHERE f.slug = $1`,
       [slug]
@@ -2463,14 +2470,15 @@ export class PGLiteEngine implements BrainEngine {
 
   async getBacklinks(slug: string, opts?: { sourceId?: string }): Promise<Link[]> {
     // v0.31.8 (D16): two-branch query. See getLinks() comment.
+    // Soft-delete filter: see getLinks() above.
     if (opts?.sourceId) {
       const { rows } = await this.db.query(
         `SELECT f.slug as from_slug, t.slug as to_slug,
                 l.link_type, l.context, l.link_source,
                 o.slug as origin_slug, l.origin_field
          FROM links l
-         JOIN pages f ON f.id = l.from_page_id
-         JOIN pages t ON t.id = l.to_page_id
+         JOIN pages f ON f.id = l.from_page_id AND f.deleted_at IS NULL
+         JOIN pages t ON t.id = l.to_page_id   AND t.deleted_at IS NULL
          LEFT JOIN pages o ON o.id = l.origin_page_id
          WHERE t.slug = $1 AND t.source_id = $2`,
         [slug, opts.sourceId]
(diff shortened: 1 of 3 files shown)


time-attack:
Closing as fix_needed. Currently conflicts with master (dirty), and independent of that it's stale: tonight's source-scope wave added a third federated branch to getLinks/getBacklinks in both engines (the one remote MCP clients always hit per its own comment) that this PR never patches. Re-opening #1702 for a fresh patch covering all three branches x two functions x two engines.

PR #1704, opened 2026-05-31
test(e2e): harden suite — kill flakes, no-op assertions, cross-test coupling

auroracapital:
## What this is

A test-only hardening pass over the E2E suite (`test/e2e/`). No product/runtime
code changes — every commit touches `test/e2e/*.test.ts` only. The goal: remove
flaky patterns, kill assertions that passed without proving anything, eliminate
cross-test/cross-file state coupling, and close a few coverage gaps. Full suite
verified green against real Postgres+pgvector: **18 files, 190 tests, 0 fail**
(Tier-2 `skills` skips cleanly without API keys).

## Commits (grouped by area)

**Flake removal**
- `minions-resilience`: dropped the two tight wall-clock upper bounds
  (`deadAt-started < 2000ms`, `cancelElapsed < 3000ms`). Terminal state
  (`dead`, all `cancelled`, `abortedChildren===10`) already proves correctness;
  the time bounds only flake on loaded runners. Inner poll deadlines still cap runtime.

**Cross-test / cross-file state coupling**
- `mechanical`: `traverse_graph` no longer depends on a prior `it()`'s `add_link`
  (re-adds its own idempotent link, asserts the linked page is reachable);
  `file_list`-without-slug seeds its own >100 rows instead of relying on the
  previous test's survivors.
- `multi-source`: `performSync`-no-sourceId sets its own `sync.repo_path` instead
  of inheriting a prior test's side effect.
- `migration-flow`: the suite repoints `process.env.HOME`/`PATH` to a temp dir and
  only restored them in `afterAll`; a mid-test throw left `HOME` dead for the rest
  of the bun process and silently broke sibling suites. Now throw-safe via per-test
  try/finally + a defensive `beforeEach` restore.

**No-op / weak assertions**
- `mechanical`: `precision@5` had zero assertions (a 0% result passed) — added a
  loose floor (every known-item query surfaces >=1 truth doc in top-5); `get_health`
  now asserts value bounds (`page_count==16`, `embed_coverage` in 0..1) not just
  `typeof`; `get_chunks` asserts non-empty text + non-decreasing `chunk_index` +
  the page name appears, instead of `toBeTruthy(chunks[0])`.
- `search-quality`: fixed the vacuous `detail=low` vector test (an empty result set
  passed the all-`compiled_truth` loop); assert pedro returns >=2 chunks; assert
  `detail=high` includes the timeline chunk; added empty-query + zero-vector
  no-throw edge tests.
- `graph-quality`: replaced `toBeGreaterThan(0)` link/timeline floors with
  fixture-derived minimums; assert the exact attendee slugs are `attended` instead
  of `.every()` over a maybe-empty set; pin `autoLinks.created` to the provable 2.

**Coverage additions**
- `graph-quality`: added `direction: out|both` + `depth:2` multi-hop traversal and
  a cycle-safety test (A to B to A terminates, bounded results).
- `multi-source`: added the missing `file_migration_ledger` cascade-delete assertion.

**Correctness / contract**
- `upgrade`: the "no-releases" test hard-asserted `update_available === false`,
  which flips to failing the moment the repo has a published release. Now asserts
  the check-update JSON contract shape (boolean `update_available`,
  `current_version === VERSION`, typed optional fields).
- `mcp`: added `find_orphans` to the asserted generated tool names.

**Skip hygiene**
- `jsonb-roundtrip` + `doctor-progress`: both skipped silently without
  `DATABASE_URL` (zero signal the regression guard ran) — added the loud-skip
  console line matching sibling files.

**Config-leak fix**
- `graph-quality`: `truncateAll` now truncates + reseeds `config`, so a test that
  sets `auto_link`/`auto_timeline` and throws before its finally can't bleed into
  later tests.

## Intentionally out of scope
- `cycle` lock-release-on-error coverage was scoped but left out: wiring a clean
  phase-throw was not unambiguous from the file's existing patterns, so it was
  skipped rather than guessed.
- No version bump / CHANGELOG entry: these are test-only changes with no
  user-facing behavior.

## Verification
Each file run individually against a throwaway `pgvector/pgvector:pg16` container,
then the full suite via `scripts/run-e2e.sh` (sequential): 18/18 files green.

### DIFF
--- a/test/e2e/doctor-progress.test.ts
+++ b/test/e2e/doctor-progress.test.ts
@@ -20,6 +20,10 @@ import {
 const skip = !hasDatabase();
 const describeE2E = skip ? describe.skip : describe;
 
+if (skip) {
+  console.log('Skipping E2E doctor --progress-json tests (DATABASE_URL not set)');
+}
+
 const CLI = join(import.meta.dir, '..', '..', 'src', 'cli.ts');
 
 describeE2E('gbrain doctor --progress-json (E2E)', () => {
--- a/test/e2e/graph-quality.test.ts
+++ b/test/e2e/graph-quality.test.ts
@@ -29,9 +29,15 @@ afterAll(async () => {
 });
 
 async function truncateAll() {
-  for (const t of ['content_chunks', 'links', 'tags', 'raw_data', 'timeline_entries', 'page_versions', 'ingest_log', 'pages']) {
+  for (const t of ['content_chunks', 'links', 'tags', 'raw_data', 'timeline_entries', 'page_versions', 'ingest_log', 'config', 'pages']) {
     await (engine as any).db.exec(`DELETE FROM ${t}`);
   }
+  // Re-seed the two config keys this file touches back to their documented
+  // defaults (both default to ON). This makes every test deterministic even if
+  // an earlier test threw before its finally restored auto_link/auto_timeline,
+  // and even though absent-key already resolves truthy via isAuto*Enabled.
+  await engine.setConfig('auto_link', 'true');
+  await engine.setConfig('auto_timeline', 'true');
 }
 
 function makeContext(): OperationContext {
@@ -77,10 +83,12 @@ describe('E2E graph quality (v0.10.1 pipeline)', () => {
     await runExtract(engine, ['links', '--source', 'db']);
     await runExtract(engine, ['timeline', '--source', 'db']);
 
-    // Verify graph populated.
+    // Verify graph populated. Concrete floors derived from the seeded fixtures:
+    //   resolvable entity refs: alice->acme, bob->acme, standup->alice, standup->bob = 4
+    //   timeline lines: alice(2) + bob(1) + acme(1) + standup(1) = 5
     const stats = await engine.getStats();
-    expect(stats.link_count).toBeGreaterThan(0);
-    expect(stats.timeline_entry_count).toBeGreaterThan(0);
+    expect(stats.link_count).toBeGreaterThanOrEqual(4);
+    expect(stats.timeline_entry_count).toBeGreaterThanOrEqual(5);
 
     // Verify typed link inference.
     const aliceLinks = await engine.getLinks('people/alice');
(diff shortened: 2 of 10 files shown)


auroracapital:
Rebased onto upstream `master` (now at v0.42.1.0) — this branch was originally cut from a stale local base ~188 commits behind, which is what caused the CONFLICTING state. It is now **MERGEABLE**.

**What changed during the rebase**
- Only `mechanical.test.ts` needed a manual conflict resolution (one hunk in `get_chunks`): kept the stronger content/ordering assertions and upstream's added `30_000` timeout. The other six commits replayed cleanly.
- `search-quality.test.ts`: two assertions an earlier pass added assumed the old keyword-search behavior (multiple chunks per page, timeline chunk surfaced for a keyword match). v0.42 returns one chunk per page, so those two were reverted to upstream's behavior-correct form. The valid additions (empty-query / zero-vector no-throw, detail=low vector non-empty floor) stayed.
- Ran `bun install` to pick up new deps the 188 upstream commits added (`src/core/ai/gateway.ts` → `ai` package, etc.).

**Test signal (upstream has no CI, so this is the verification)**
Every file this branch touches was run individually against a fresh `pgvector/pgvector:pg16` container on the v0.42 schema (full v1→v111 migration chain), all green:

| file | result |
|---|---|
| minions-resilience | 5 pass |
| mcp | 2 pass |
| graph-quality | 11 pass |
| search-quality | 14 pass |
| multi-source | 20 pass |
| migration-flow | 4 pass |
| jsonb-roundtrip | 5 pass |
| doctor-progress | 3 pass |
| mechanical | 78 pass |
| upgrade | 5 pass |

`mechanical.test.ts` was also diffed for isolation: it passes 78/78 alone on a pristine DB for **both** this branch and upstream `master`. When several Postgres E2E files run before it against one shared DB it shows fixture-count failures, but upstream's unmodified `mechanical.test.ts` fails identically in that ordering — it's a pre-existing shared-DB cross-file isolation quirk in the suite, not introduced by this diff. Left out of scope to keep the PR to its stated purpose.

Net diff is 10 test files, +356/-135, no product/runtime code.

PR #1705, opened 2026-05-31
sources: gate recloneIfMissing against paths outside the managed clone root

### LINKED ISSUE #1526: Data loss: `recloneIfMissing` rm-rfs user-managed source path; ungated against arbitrary local_path

mvanhorn:
## Summary

`recloneIfMissing` recovers a degraded source by re-cloning via a destructive temp + rename over `src.local_path`, but it never verified that `local_path` is actually gbrain-managed storage. A source that was re-added or repointed at a user-managed directory could therefore have that directory wiped during recovery.

This gates the destructive reclone behind `isManagedRecloneTarget()`:

- Rejects any raw `..` path segment first, so a symlinked segment followed by `..` cannot make the OS-resolved `rm`/`rename` target diverge from the lexically-checked path.
- Requires the resolved path to be strictly contained in the gbrain clone root.
- Is symlink-aware: it `realpath`s the nearest existing ancestor and confirms it resolves within the realpathed clone root, so a symlinked parent pointing outside managed storage is rejected.

When the target is not managed it throws `unsafe_reclone_path` instead of deleting anything, and tells the user to repoint or re-add the source.

## Testing

- `bun test test/sources-ops.test.ts` -> 28 pass, covering: non-managed path, relative/trailing-slash normalization, symlink parent escape, and raw `..` traversal. Each asserts the outside victim file survives.
- `bun run verify` -> 29/29.

Closes #1526

### DIFF
--- a/src/core/sources-ops.ts
+++ b/src/core/sources-ops.ts
@@ -38,7 +38,7 @@
 
 import { existsSync, mkdirSync, renameSync, rmSync, lstatSync } from 'fs';
 import { realpathSync } from 'fs';
-import { join, dirname, resolve as resolvePath } from 'path';
+import { join, dirname, relative, sep, isAbsolute, resolve as resolvePath } from 'path';
 import { randomBytes } from 'crypto';
 import type { BrainEngine } from './engine.ts';
 import {
@@ -66,6 +66,7 @@ export type SourceOpErrorCode =
   | 'not_found'
   | 'protected_id'
   | 'clone_dir_outside_gbrain'
+  | 'unsafe_reclone_path'
   | 'symlink_escape';
 
 export class SourceOpError extends Error {
@@ -251,6 +252,64 @@ export function isPathContained(child: string, parent: string): boolean {
   // Append a separator to parent so /foo doesn't match /foobar.
   const parentWithSep = resolvedParent.endsWith('/') ? resolvedParent : resolvedParent + '/';
   return resolvedChild === resolvedParent || resolvedChild.startsWith(parentWithSep);
+}
+
+function isResolvedPathWithin(
+  childAbs: string,
+  parentAbs: string,
+  allowEqual = true,
+): boolean {
+  const rel = relative(parentAbs, childAbs);
+  if (rel === '') return allowEqual;
+  return (
+    rel !== '..' &&
+    !rel.startsWith(`..${sep}`) &&
+    !isAbsolute(rel) &&
+    resolvePath(parentAbs, rel) === childAbs
+  );
+}
+
+function pathExistsByLstat(path: string): boolean {
+  try {
+    lstatSync(path);
+    return true;
+  } catch {
+    return false;
+  }
+}
+
+function hasRawParentSegment(path: string): boolean {
+  return path.split(sep).some((segment) => segment === '..');
+}
+
+function isManagedRecloneTarget(localPath: string, cloneRoot: string): boolean {
+  if (hasRawParentSegment(localPath)) return false;
+
+  const rootAbs = resolvePath(cloneRoot);
+  const targetAbs = resolvePath(localPath);
+  if (!isResolvedPathWithin(targetAbs, rootAbs, false)) return false;
+
+  let rootReal: string;
+  try {
+    rootReal = realpathSync(rootAbs);
+  } catch {
+    return false;
+  }
+
+  let ancestor = targetAbs;
+  while (!pathExistsByLstat(ancestor)) {
+    const parent = dirname(ancestor);
+    if (parent === ancestor) return false;
+    ancestor = parent;
+  }
+
+  let ancestorReal: string;
+  try {
+    ancestorReal = realpathSync(ancestor);
+  } catch {
+    return false;
+  }
+  return isResolvedPathWithin(ancestorReal, rootReal);
 }
 
 // ── addSource ───────────────────────────────────────────────────────────────
@@ -697,6 +756,17 @@ export async function recloneIfMissing(
 
   const state = validateRepoState(src.local_path, remoteUrl);
   if (state === 'healthy') return false;
+
+  const cloneRoot = gbrainPath('clones');
+  mkdirSync(cloneRoot, { recursive: true });
+  if (!isManagedRecloneTarget(src.local_path, cloneRoot)) {
+    throw new SourceOpError(
+      'unsafe_reclone_path',
+      `Refusing to re-clone source "${id}" into ${resolvePath(src.local_path)}: ` +
+        `path is outside the gbrain-managed clone root ${resolvePath(cloneRoot)}. ` +
+        `Repoint or re-add the source instead.`,
+    );
+  }
 
   // Re-clone via temp + rename, mirroring addSource's atomicity contract.
   const tempDir = makeTempCloneDir(id);
--- a/test/sources-ops.test.ts
+++ b/test/sources-ops.test.ts
@@ -13,7 +13,7 @@ import {
   chmodSync,
   existsSync,
 } from 'fs';
-import { join } from 'path';
+import { join, relative, sep } from 'path';
 import { tmpdir } from 'os';
 import { PGLiteEngine } from '../src/core/pglite-engine.ts';
 import {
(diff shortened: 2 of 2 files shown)


garrytan:
Thanks @mvanhorn — the recloneIfMissing data-loss path is sealed in master via #1881 (v0.42.33.0) with an ownership guard (`isOwnedClone`, managed_clone marker / exact default-clone-dir match), a TOCTOU re-check, and a symlink-leaf guard — stronger than path-containment alone (a user tree under $GBRAIN_HOME is owned-false). Closing as superseded; thank you for flagging it.

PR #1706, opened 2026-05-31
query: filter since/until on effective date instead of updated_at

### LINKED ISSUE #1520: query: since/until filters on updated_at instead of effective_date
## Bug

The `since` / `until` parameters on the `query` operation are documented as filtering on `effective_date`:

> `since`: v0.29.1 — filter to pages whose effective_date is >= this.
> `until`: v0.29.1 — filter to effective_date <= this.

But the actual SQL in `postgres-engine.ts` (lines 1401, 1406, 1554, 1559, 1676, 1681) filters on `COALESCE(p.updated_at, p.created_at)`:

```sql
AND COALESCE(p.updated_at, p.created_at) > $N::timestamptz
AND COALESCE(p.updated_at, p.created_at) < $N::timestamptz
```

## Impact

For pages ingested via `sync` or `import`, `updated_at` reflects **import time**, not the content's actual date. Email pages with filenames like `2026-05-22_$

In practice, `query --since 2026-05-18 --until 2026-05-25` returns **zero results** even though dozens of pages have `effective_date` within that range. The $

## Reproduction

1. Import a markdown directory with date-prefixed filenames
2. Verify `effective_date` is populated: `SELECT effective_date, slug FROM pages WHERE effective_date IS NOT NULL LIMIT 5;`
3. Run `gbrain query "anything" --since 2026-05-01 --until 2026-05-31` — returns empty
4. Run `gbrain query "anything"` (no date filter) — returns pages with `effective_date` in that range

## Fix

Change all 6 occurrences in `postgres-engine.ts` from:
```sql
COALESCE(p.updated_at, p.created_at)
```
to:
```sql
COALESCE(p.effective_date, p.updated_at, p.created_at)
```

This respects `effective_date` when set (filename-derived, frontmatter, etc.) and falls back to `updated_at` / `created_at` for pages without one.

The same bug likely exists in `pglite-engine.ts` (same pattern at lines ~1398–1728).

## Environment
- gbrain v0.41.14.0
- Postgres engine with pgvector
- 1,416 pages (1,227 emails with filename-derived effective_date, 189 concepts)

mvanhorn:
## Summary

`since`/`until` range filters were applied against `COALESCE(updated_at, created_at)`, so an entry that was edited recently but whose content is dated to the past (or vice versa) was filtered by its row-modification time rather than its effective date. Time-bounded queries therefore returned the wrong set.

Filter on the effective date instead: `COALESCE(effective_date, updated_at, created_at)`, applied consistently across all three query paths. The `updated_at`/`created_at` fallback preserves current behavior for rows without an `effective_date`.

## Testing

- `bun test test/postgres-engine.test.ts` passes, including a DB-free regression guard asserting the since/until clauses filter on `effective_date`.
- `bun run verify` -> 29/29.

Closes #1520

### DIFF
--- a/src/core/postgres-engine.ts
+++ b/src/core/postgres-engine.ts
@@ -1482,16 +1482,16 @@ export class PostgresEngine implements BrainEngine {
       params.push(symbolKind);
       symbolKindClause = `AND cc.symbol_type = $${params.length}`;
     }
-    // v0.27.0: date filtering support
+    // v0.29.1: since/until filter by effective date, with import-time fallback.
     let afterDateClause = '';
     if (opts?.afterDate) {
       params.push(opts.afterDate);
-      afterDateClause = `AND COALESCE(p.updated_at, p.created_at) > $${params.length}::timestamptz`;
+      afterDateClause = `AND COALESCE(p.effective_date, p.updated_at, p.created_at) > $${params.length}::timestamptz`;
     }
     let beforeDateClause = '';
     if (opts?.beforeDate) {
       params.push(opts.beforeDate);
-      beforeDateClause = `AND COALESCE(p.updated_at, p.created_at) < $${params.length}::timestamptz`;
+      beforeDateClause = `AND COALESCE(p.effective_date, p.updated_at, p.created_at) < $${params.length}::timestamptz`;
     }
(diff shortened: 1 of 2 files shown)


mvanhorn:
Thank you as well for #1706, @time-attack - since/until filtering on effective date matches what users actually mean.

PR #1707, opened 2026-05-31
schema: resolve all bundled packs in schema use, not just gbrain-base

### LINKED ISSUE #1574: schema fork/show/diff/lint/validate `gbrain-base-v2` fails with "Unknown pack" — `packPathByName` only resolves `gbrain-base`
## Summary

`packPathByName` in [`src/commands/schema.ts:368-387`](src/commands/schema.ts#L368-L387) only hardcodes a bundled-resolution branch for `gbrain-base`, even though `BUNDLED_PACK_NAMES` in [`src/core/schema-pack/mutate.ts:96`](src/core/schema-pack/mutate.ts#L96) (and `load-active.ts:114`) lists three bundled packs:

```ts
export const BUNDLED_PACK_NAMES = new Set(['gbrain-base', 'gbrain-recommended', 'gbrain-base-v2']);
```

As a result, every `schema` subcommand that resolves by name via `packPathByName` fails for `gbrain-base-v2` (and presumably `gbrain-recommended`) with `Unknown pack`.

## Repro

```
$ gbrain schema show gbrain-base-v2
Error: Unknown pack: gbrain-base-v2

$ gbrain schema fork gbrain-base-v2 morty
Error: Unknown pack: gbrain-base-v2
```

The `unify-types` job is unaffected because it uses `loadActivePack` (which DOES know about all bundled packs), not `packPathByName`.

## Affected CLI surfaces

- `gbrain schema show <bundled-pack>`
- `gbrain schema fork <bundled-pack> <new>`
- `gbrain schema diff <bundled-pack> <other>`
- `gbrain schema lint <bundled-pack>`
- `gbrain schema validate <bundled-pack>`

…all of the above, for any pack in `BUNDLED_PACK_NAMES` other than `gbrain-base`.

## Proposed fix

Generalize the bundled-resolution branch to iterate over `BUNDLED_PACK_NAMES` and look up `<name>.yaml`:

```ts
function packPathByName(name: string): string | null {
  if (BUNDLED_PACK_NAMES.has(name)) {
    const here = dirname(new URL(import.meta.url).pathname);
    const candidates = [
      join(here, '..', 'core', 'schema-pack', 'base', `${name}.yaml`),
      join(here, '..', '..', 'src', 'core', 'schema-pack', 'base', `${name}.yaml`),
    ];
    for (const c of candidates) {
      if (existsSync(c)) return c;
    }
    return null;
  }
  const baseDir = gbrainPath('schema-packs', name);
  for (const c of ['pack.yaml', 'pack.yml', 'pack.json']) {
    const candidate = join(baseDir, c);
    if (existsSync(candidate)) return candidate;
  }
  return null;
}
```

## Operator workaround

Manual `cp` from the gbrain source dir:

```
cp ~/Projects/gbrain/src/core/schema-pack/base/gbrain-base-v2.yaml ~/.gbrain/schema-packs/morty/pack.yaml
```

…then continue with `gbrain schema validate morty` (which works because by then the pack file exists at the user-pack path).

## Context

Hit during a morty-v3 pack-activation session (2026-05-27) where the intended path was `gbrain schema fork gbrain-base-v2 morty`. Detailed write-up internal to that project, but the bug-class is generic — anyone trying to use `schema fork/show/diff/lint/validate` against a bundled pack other than `gbrain-base` will trip it.

mvanhorn:
## Summary

`gbrain schema use <name>` hardcoded `gbrain-base` (and a fixed two-element bundled list elsewhere), so other bundled packs (e.g. `gbrain-recommended`, `gbrain-base-v2`) could not be selected by name even though they ship with gbrain.

Resolve bundled packs through the shared `BUNDLED_PACK_NAMES` set and look up `<name>.yaml` generically, so every bundled pack activates by name.

## Testing

- `bun test test/commands/schema-packpath.test.ts` -> 3 pass (bundled-name resolution + the existing non-bundled fallthrough cases).

Closes #1574

### DIFF
--- a/src/commands/schema.ts
+++ b/src/commands/schema.ts
@@ -23,6 +23,7 @@ import {
   addAliasToType,
   addLinkTypeToPack,
   addPrefixToType,
+  BUNDLED_PACK_NAMES,
   addTypeToPack,
   invalidatePackCache,
   loadActivePack,
@@ -179,7 +180,7 @@ async function runActive(_args: string[]): Promise<void> {
 }
 
 function runList(_args: string[]): void {
-  const bundled = ['gbrain-base', 'gbrain-recommended'];
+  const bundled = [...BUNDLED_PACK_NAMES];
   const installedDir = gbrainPath('schema-packs');
   const installed: string[] = [];
   if (existsSync(installedDir)) {
@@ -366,12 +367,12 @@ function runUse(args: string[]): void {
 }
 
 function packPathByName(name: string): string | null {
-  if (name === 'gbrain-base') {
+  if (BUNDLED_PACK_NAMES.has(name)) {
     // Resolve bundled YAML — try a few locations.
     const here = dirname(new URL(import.meta.url).pathname);
     const candidates = [
-      join(here, '..', 'core', 'schema-pack', 'base', 'gbrain-base.yaml'),
-      join(here, '..', '..', 'src', 'core', 'schema-pack', 'base', 'gbrain-base.yaml'),
+      join(here, '..', 'core', 'schema-pack', 'base', `${name}.yaml`),
+      join(here, '..', '..', 'src', 'core', 'schema-pack', 'base', `${name}.yaml`),
     ];
     for (const c of candidates) {
       if (existsSync(c)) return c;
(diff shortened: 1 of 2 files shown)


mvanhorn:
Thanks for taking both of these, @time-attack - resolving all bundled packs in schema use closes a real gap.

PR #1708, opened 2026-05-31
fix(skillopt): support AI SDK v6 tool loops

JE4NVRG:
## Summary

- Wrap raw JSON schemas with AI SDK v6 `jsonSchema()` when building gateway tools.
- Feed tool results back as `role: 'tool'` messages with v6-compatible `{ type: 'json' | 'error-text', value }` outputs.
- Reuse the central `paramDefToSchema()` mapper for SkillOpt rollout tools so enum/default/items metadata survives.
- Persist the accepted proposal artifact in `--no-mutate` SkillOpt runs by writing `skillopt/best.md` atomically.

## Why

While testing `gbrain skillopt` with an OpenAI-compatible chat provider, tool calls failed with schema/tool-result shape errors under AI SDK v6. After the gateway fixes, chat, forced toolLoop, and one SkillOpt rollout all complete successfully.

## Test Plan

- `bun --check src/core/ai/gateway.ts src/core/skillopt/rollout.ts src/core/skillopt/orchestrator.ts test/ai/gateway-tool-loop.test.ts test/skillopt/rollout-schema.test.ts`
- `bun test test/ai/gateway-tool-loop.test.ts test/skillopt/rollout-schema.test.ts`
- `bun run typecheck`

Local note: full `bun run verify` is not reliable in my Windows/MSYS checkout because many package scripts invoke bare `scripts/*.sh` paths that Bun for Windows cannot dispatch directly; the focused checks and typecheck above are green.

### DIFF
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -21,7 +21,7 @@
  *     rotation (via configureGateway()) invalidates stale entries.
  */
 
-import { embed as aiEmbed, embedMany, generateObject, generateText } from 'ai';
+import { embed as aiEmbed, embedMany, generateObject, generateText, jsonSchema } from 'ai';
 import { AsyncLocalStorage } from 'node:async_hooks';
 import { listRecipes } from './recipes/index.ts';
 import { createOpenAI } from '@ai-sdk/openai';
@@ -2457,7 +2457,7 @@ export async function chat(opts: ChatOpts): Promise<ChatResult> {
   const tools = (opts.tools ?? []).reduce((acc, t) => {
     acc[t.name] = {
       description: t.description,
-      inputSchema: { jsonSchema: t.inputSchema } as any,
+      inputSchema: jsonSchema(t.inputSchema as any) as any,
     };
     return acc;
   }, {} as Record<string, any>);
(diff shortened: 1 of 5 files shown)


JE4NVRG:
Follow-up verification from Windows sandbox:\n\n- Branch is current with upstream master (0 behind / 1 ahead).\n- Focused regression tests pass: `bun test test/ai/gateway-tool-loop.test.ts test/skillopt/rollout-schema.test.ts` → 9 pass / 0 fail.\n- TypeScript check passes: `bun run typecheck`.\n\nGitHub status checks are not currently present on the PR, so local focused verification is green.

garrytan:
Landed in #1809 (v0.42.19.0). Your tool-result-shape work (`role:'tool'` + typed `output`) is the heart of the fix — thank you. We reworked it into a single `toModelMessages()` boundary adapter at the `chat()` edge (instead of scattering the conversion across toolLoop's push sites), which keeps toolLoop provider-neutral and also handles output normalization + dropping the v6-invalid `isError` field. Crediting you as Co-Author. The skillopt best.md change is filed as a P2 follow-up. Closing in favor of the combined wave.

PR #1713, opened 2026-05-31
fix(skill-catalog): fold block-scalar descriptions instead of leaking `|`/`>` (#1711)

### LINKED ISSUE #1711: list_skills/get_skill: catalog description shows block-scalar indicator (`|`/`>`) for skills using `description: |`
**What happened?**

`list_skills` / `get_skill` return the literal YAML block-scalar indicator (`|` or `>`) as a skill's description when the skill's `SKILL.md` frontmatter uses a block scalar:

```yaml
---
name: browse
description: |
  Fast headless browser for QA testing and site dogfooding. Navigate any URL...
---
```

→ catalog entry comes back as `"description": "|"`. Affects every skill authored with `description: |` or `description: >`. Single-line `description: ...` skills are fine. `triggers`/`tools` parse correctly, and `get_skill` still returns the full prose body — so it's a catalog-display issue, not data loss.

**What did you expect?**

The folded block-scalar content as the one-line description (or, minimally, the body's first prose line via the existing `firstProseLine` fallback).

**Steps to reproduce**

1. Author a skill whose `SKILL.md` frontmatter uses `description: |` (block scalar) with the text on the indented lines below.
2. Enable `mcp.publish_skills` and point `GBRAIN_SKILLS_DIR` at the skills dir.
3. Call `list_skills` over MCP → that skill's `description` is `"|"` (or `">"`).

**Environment**

- gbrain version: 0.42.1.0
- OS: macOS 26.5 (Darwin 25.5.0)
- Bun version: 1.3.13
- Database: self-hosted (PGLite)

```
N/A — this is a static-analysis bug in src/core/skill-catalog.ts (frontmatter
description parsing), independent of brain health/connection. doctor reports the
brain connected and healthy; no doctor check covers the list_skills catalog.
```

---

### Root cause

`src/core/skill-catalog.ts`:

```js
function parseDescriptionField(raw) {
  const m = raw.match(/^description:\s*["']?(.+?)["']?\s*$/m);
  if (!m) return undefined;
  const v = m[1].trim();
  return v.length > 0 ? v : undefined;   // returns "|" for `description: |`
}

export function oneLineDescription(raw, body) {
  return parseDescriptionField(raw) ?? firstProseLine(body);
}
```

For `description: |`, the regex captures `|` (length 1 > 0), so `parseDescriptionField` returns `"|"` and the `?? firstProseLine(body)` fallback never fires. Same for `>` and chomping/indent variants (`|-`, `>2`, etc.).

### Suggested fix

In `parseDescriptionField`, when the captured value is a bare block-scalar indicator (`|`/`>`, optionally followed by `+`/`-`/digits), either:
- **(minimal)** return `undefined` so `oneLineDescription` falls through to `firstProseLine(body)`, or
- **(complete)** read the indented continuation lines and fold them into a one-liner.

Happy to send a PR with a `test/` case if useful.

twidtwid:
## Summary

Fixes #1711 — `list_skills` / `get_skill` returned the literal YAML block-scalar indicator (`|` or `>`) as a skill's description for any `SKILL.md` authored with `description: |` (or `>`).

## Root cause

`src/core/skill-catalog.ts` → `parseDescriptionField`:

```js
const m = raw.match(/^description:\s*["']?(.+?)["']?\s*$/m);
...
return v.length > 0 ? v : undefined;   // returns "|" for `description: |`
```

The single-line regex captured the block indicator `|` (length 1 > 0), so `oneLineDescription`'s `?? firstProseLine(body)` fallback never fired.

## Fix

`parseDescriptionField` now detects a block-scalar header (`|`/`>` with optional chomping `+`/`-` and indent digit, optional trailing comment), folds the following indented lines into a single line, and returns `undefined` for an empty block so the body fallback still works. Inline (and quoted) descriptions are unchanged.

## Tests

Added 4 cases to the existing `oneLineDescription` suite in `test/skill-catalog.test.ts`:
- folds a literal block scalar (`|`)
- folds a folded block scalar (`>`)
- handles chomping indicators (`|-`)
- never leaks a bare indicator; empty block falls back to body (regression)

```
bun test test/skill-catalog.test.ts   → 19 pass / 0 fail
bun run verify                         → 29 checks, all green
```

Scope: catalog-display only; no data path or `get_skill` body change.

### DIFF
--- a/src/core/skill-catalog.ts
+++ b/src/core/skill-catalog.ts
@@ -326,11 +326,36 @@ function availableBrainTools(ctx: OperationContext): string[] {
 // Frontmatter projection + description
 // ---------------------------------------------------------------------------
 
-/** Parse a single-line `description:` from raw frontmatter (best-effort). */
+/**
+ * Parse the `description:` field from raw frontmatter (best-effort). Handles
+ * both inline (`description: text`) and YAML block scalars (`description: |`
+ * or `description: >`, including chomping/indent indicators like `|-`, `>2`),
+ * folding block content into a single line. Returns undefined when there's no
+ * usable value so the caller falls back to the body's first prose line — a bare
+ * block indicator must NOT leak through as the literal `|`/`>`.
+ */
 function parseDescriptionField(raw: string): string | undefined {
-  const m = raw.match(/^description:\s*["']?(.+?)["']?\s*$/m);
-  if (!m) return undefined;
-  const v = m[1].trim();
+  const lines = raw.split('\n');
+  const idx = lines.findIndex(l => /^description:/.test(l));
+  if (idx === -1) return undefined;
+
+  const after = lines[idx].replace(/^description:[ \t]*/, '');
+  // Block scalar: `|` or `>`, optional chomping (+/-) and/or indent digit,
+  // optional trailing comment. Fold the following indented lines into one line.
+  if (/^[|>][+-]?\d*[ \t]*(#.*)?$/.test(after.trim())) {
+    const folded: string[] = [];
+    for (let i = idx + 1; i < lines.length; i++) {
+      const line = lines[i];
+      if (line.trim() === '') { folded.push(''); continue; }
+      if (/^[ \t]/.test(line)) { folded.push(line.trim()); continue; }
+      break; // dedent to column 0 → next top-level key ends the block
+    }
+    const v = folded.join(' ').replace(/\s+/g, ' ').trim();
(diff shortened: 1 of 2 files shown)


garrytan:
Thanks @twidtwid — block-scalar description folding landed in master via #1999 (v0.42.37.0): `parseDescriptionField` detects `|`/`>` scalars, folds continuation lines, and falls back to body prose on an empty block. Closing as superseded (closes #1711 too).

PR #1718, opened 2026-06-01
fix: preserve cycle engine during lint phase

whobin1982:
## Summary
- Thread the cycle-owned BrainEngine into the lint phase.
- Reuse caller-owned engine when resolving content-sanity config.
- Avoid creating and disconnecting a temporary engine during full dream/cycle lint.

## Test Plan
- bun test test/core/cycle.serial.test.ts --test-name-pattern 'cycle-owned engine is threaded into lint during full cycles'
- bun test test/core/cycle.serial.test.ts
- bun test test/lint.test.ts test/lint-content-sanity.test.ts test/post-write-lint.test.ts
- bun run verify


garrytan:

PR #1723, opened 2026-06-01
fix(code-graph): TS/TSX call edges now resolve (const-arrow naming + source_id on edge writes)

sonlndv:
## Summary

`code-callers` / `code-callees` return **0** for TypeScript/TSX repositories even after a
full `gbrain sync --strategy code`. Two coupled bugs in the code-ingestion path cause it.
This PR fixes both. Verified on a real React Native / TSX repo: `code-callers useTheme`
goes from **0 → 64 resolved callers**.

### Bug 1 — `extractSymbolName` doesn't name `const`/arrow-bound symbols
`src/core/chunkers/code.ts` `extractSymbolName` handles `function_declaration` /
`class_declaration` / `export_statement` (via the `declaration` field recursion) but never
descends into a `variable_declarator`. So the dominant React/RN form

```ts
export const Foo = () => { … }
```

produces a chunk with `symbol_name = null` → `symbol_name_qualified = null`. Its outbound
call edges are then dropped at the qualified-name gate in
`src/core/import-file.ts` (`if (!from.id || !from.symbol_name_qualified) continue;`).

**Fix:** name `lexical_declaration` / `variable_declaration` via the declarator's `name`
field. Uses **exact `node.type ===` equality** (never a substring / `endsWith('declaration')`
test, which would wrongly match Go's `var_declaration`/`const_declaration`,
`function_declaration`, etc. across the other grammars).

### Bug 2 — edge writes omit `source_id`
`src/core/import-file.ts` builds each `CodeEdgeInput` without `source_id`, even though
`sourceId` is in scope (`const sourceId = opts.sourceId;`). Every row therefore lands in
`code_edges_symbol` with `source_id = NULL`, and every source-scoped reader
(`getCallersOf`/`getCalleesOf` → `AND source_id = $scoped`; the resolver's edge-load query)
silently misses them. So even once Bug 1 is fixed and edges pass the gate, scoped
`code-callers --source X` still returns 0.

**Fix:** thread `source_id: sourceId ?? null` into the edge input. Default brain (no
`sourceId`) → `null`, which matches unscoped reads; a named source → its id, which matches
`--source` reads.

### Watermark
`EDGE_EXTRACTOR_VERSION_TS` is bumped so `resolveSymbolEdgesIncremental` /
`edges-backfill` re-walk chunks whose `edges_backfilled_at` was already stamped by a prior
(broken) run.

## Verification

On a real RN/TSX source (~194 files, ~850 code chunks):

| | before | after |
|---|---|---|
| `code-callers useTheme` | 0 | **64** (HeroChip, Chip, Card, …) |
| `code-callees DiaryScreen` | 0 | **66** |
| `edges-backfill` | 0 walked, 0 resolved | 2000 walked, 30+ resolved |

`bun test test/chunkers/code.test.ts` — **33 pass** (31 existing + 2 new pins asserting
`export const Foo = () => {}` and a plain `const Bar = () => {}` get `symbolName` +
`symbolNameQualified`).

## Notes / scope

- Naming **all** top-level const/var bindings (not only arrow functions) is intentional and
  harmless — it also makes `code-def` find `export const config = …`.
- `CHUNKER_VERSION` is **not** bumped here (would force a fleet-wide re-chunk+re-embed). New
  or re-synced pages pick the fix up naturally; existing corpora can re-index when desired.
- Small files that collapse via `mergeSmallSiblings` still produce a `merged`/null chunk —
  a pre-existing, separate limitation, out of scope for this fix.


garrytan:

PR #1725, opened 2026-06-01
fix(skillopt): surface empty rollout failures

ygc3817922006-sketch:
## Summary
- fail loud when a SkillOpt validation gate produces zero scored rollouts instead of folding every rollout error into score 0
- preserve the original rollout errors in the run receipt, audit log, and CLI output for aborted/errored runs
- add unit and E2E coverage so all-target-rollout failures no longer report as `no_improvement`

## Why
A live `gbrain skillopt` run could return `no_improvement` with `sel_score_runs: []` and `$0.00` spend even though no target rollouts were actually scored. That makes a transport/config/budget failure look like a valid optimization result.

With this change, all-rollout failure surfaces as a diagnostic failure. In the local smoke case, the hidden cause became visible: the default DeepSeek model had no pricing entry under `--max-cost`, so BudgetTracker blocked every rollout.

## Validation
- `bun test test/skillopt/validate-gate.test.ts`
- `bun test test/e2e/skillopt-loop.serial.test.ts`
- `bun test test/skillopt test/e2e/skillopt-loop.serial.test.ts`

## Note
`bun run typecheck` still fails in this checkout on pre-existing missing type/dependency issues for `chokidar`, `fast-check`, and `js-yaml`; those errors are unrelated to this patch.


ygc3817922006-sketch:
Closing this for now; not pursuing the PR path at this time.

PR #1730, opened 2026-06-01
Suppress NER onboarding when active pack cannot infer NER links

shawnduggan:
## Summary
- Gate the `onboard.extract_ner_links` remediation on the active schema pack actually declaring at least one `link_types[].inference.regex` rule.
- Keep `entity_link_coverage` as a warning when graph coverage is low, but avoid submitting a no-op `extract-ner` job under packs that cannot infer NER links.
- Add focused PGLite coverage for both sides of the contract: `gbrain-base` still recommends NER, while `gbrain-base-v2` does not.

## Why
`extract-ner` only does useful work when the active schema pack has regex-backed inference rules. The bundled `gbrain-base` pack has those rules for verbs like `founded`, `invested_in`, and `works_at`; `gbrain-base-v2` currently declares link verbs/inverses but does not declare `inference.regex` metadata.

Without this guard, `gbrain onboard --check` recommends `extract-ner` solely because entity-link coverage is low, even when the active pack gives the NER extractor no regex rules to run. On a `gbrain-base-v2` brain this produces a misleading remediation: the CLI NER path cleanly no-ops with 0 pages / 0 links, while onboard still suggests it as an auto-remediation.

This PR makes the recommendation capability-aware. It does not try to reintroduce broad regex inference into `gbrain-base-v2`; that is a separate schema-design decision.

## Tests
- `bun test test/onboard-entity-link-coverage.test.ts`
- `bun test test/e2e/onboard-full-flow.test.ts`
- `bun run typecheck`

### DIFF
--- a/src/core/onboard/checks.ts
+++ b/src/core/onboard/checks.ts
@@ -18,6 +18,7 @@
 import type { BrainEngine } from '../engine.ts';
 import type { RemediationStep } from '../remediation-step.ts';
 import { makeRemediationStep } from '../remediation-step.ts';
+import { loadConfig } from '../config.ts';
 
 /** Shared shape returned by all four checks. */
 export interface OnboardCheckResult {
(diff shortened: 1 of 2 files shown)


time-attack:
Thanks @shawnduggan! 🙏

What this does: This PR prevents the onboard check from recommending extract-ner when the active schema pack cannot actually infer NER links (no link_types[].inference.regex rules). It adds a local gate in the onboard module to check this capability…

This overlaps with #2362, which we're carrying forward, so I'm closing this in its favor to keep the review queue focused. If yours covers a case #2362 misses, reopen with that detail and we'll reconcile them.

PR #1731, opened 2026-06-01
fix(import): make checkpoints staging-first (#1728)

### LINKED ISSUE #1728: import-checkpoint.json `dir` can resolve to CWD/repo root on SIGTERM — enforce staging-first checkpointing
Spun out of [garrytan/gstack#1802](https://github.com/garrytan/gstack/issues/1802), where gstack's `/sync-gbrain` resume path `rm -rf`'d a user's repo because `import-checkpoint.json`'s `dir` field was the repo root.

## What happened downstream

gstack stages prepared pages to `~/.gstack/.staging-ingest-*`, runs `gbrain import <stagingDir>`, and on completion deletes that dir. Its resume feature reused the checkpoint's `dir` as the staging dir. When an autopilot `gbrain import` was SIGTERM'd on timeout **while the working directory was the user's repo**, the persisted checkpoint `dir` resolved to the repo root rather than the import-target staging dir. The next run adopted it and deleted it.

gstack is fixing its own `rm -rf` boundary (fail-closed ownership guard — gstack#1802 / its PR). But the durable fix is here: **a checkpoint should only ever name a path gbrain created and owns.**

## Requests

1. **Persist `dir` as the absolute, resolved import-target path captured at import start** — never re-derive it from CWD on interrupt/SIGTERM. The poison arose because the interrupted write resolved `dir` to the process CWD instead of the argument passed to `gbrain import`.
2. **Self-describing checkpoint** — consider a `schema_version` plus an `owner`/`kind` field so consumers can validate a checkpoint before acting on it, and reject malformed/old-format files rather than parsing garbage.
3. **Document the checkpoint contract** in the sync docs: what `dir` means, who owns cleanup of it, and the invariant that it is always a gbrain-minted path.

## Severity

`rm -rf` of a user's working tree (including untracked/uncommitted work) with no confirmation. Commits survived on the remote, but the blast radius is user data. High.

## Repro shape

1. `gbrain autopilot` / `gbrain import <dir>` interrupted via SIGTERM mid-run while CWD is a repo.
2. Inspect `~/.gbrain/import-checkpoint.json` → `dir` is the repo root, not the import target.

Filed from the same investigation as gstack#1802; happy to provide the `fs_usage` trace and the gstack-side mitigation PR for cross-reference.

diazMelgarejo:
Fixes #1728.

## Summary
- Capture `gbrain import <dir>` target once at import start as an absolute real path, then use that canonical value for collection, checkpoint load/save, ingest logging, and git continuity.
- Write self-describing import checkpoints with `schema_version: 1`, `owner: "gbrain"`, and `kind: "import"` while preserving safe legacy path-based checkpoint reads.
- Reject relative checkpoint dirs and wrong metadata, and document the downstream checkpoint contract in `docs/guides/live-sync.md`.
- Adds `docs/reference/gbrain-1728-submission-package.md` with the autoplan-style package matching the gstack #1802 companion.

## Tests
- `bun test test/import-checkpoint.test.ts` (23 pass)
- `bun run typecheck` currently fails on pre-existing `test/e2e/openclaw-plugin-load-real.test.ts:281` parameter type incompatibility, unrelated to this change.

## Notes
This is the upstream prevention half of the gstack #1802 incident. gstack PR #1827 guards its own deletion boundary; this PR fixes the producer-side checkpoint identity contract.

## Cross-links
- gstack PR: https://github.com/garrytan/gstack/pull/1827
- gstack issue: https://github.com/garrytan/gstack/issues/1802
- gbrain issue: https://github.com/garrytan/gbrain/issues/1728

## Cross-links
- gstack PR: https://github.com/garrytan/gstack/pull/1827
- gstack issue: https://github.com/garrytan/gstack/issues/1802
- gbrain issue: https://github.com/garrytan/gbrain/issues/1728


time-attack:
Superseded with credit by #2935 (just merged): your staging-first checkpoint fix — canonical resolved target dir captured once, self-describing checkpoint metadata, relative-dir rejection — ported onto current master with your authorship preserved via Co-authored-by (the unrelated env hunk was dropped). Fixes #1728. Thanks for the careful root-cause work.

PR #1733, opened 2026-06-01
Add local iPhone backup importer

H4RR1SON:
## Summary

Adds a local-CLI-only `gbrain import-iphone` path for already-decrypted iPhone backups. The importer reads `Manifest.db`, `sms.db`, and `AddressBook.sqlitedb`, renders contacts as `person` pages and iMessage threads as `conversation` pages, then imports them with `noEmbed:true` so the existing `extract-conversation-facts` → `embed --stale` pipeline can take over.

The command is intentionally not exposed as an MCP operation. It is also refused on thin-client installs because decrypted iPhone backups must be handled on the host machine with the local brain.

## What changed

- Added pure iPhone backup modules under `src/core/iphone/`:
  - Manifest path resolution with backup-root containment and symlink rejection
  - SMS parsing with Apple timestamp conversion, optional-column fallback, row cap, and message text cap
  - Contacts parsing with phone/email normalization
  - Name resolution and markdown rendering in the existing `imessage-slack` parser format
- Added `IphoneBackupSource` as a migration-mode `IngestionSource`.
- Added `gbrain import-iphone --backup <decrypted-dir> [--source-id <id>] [--self-name ...] [--limit N] [--max-messages N] [--dry-run] [--json]`.
- Added overwrite guards so existing non-iPhone pages are not clobbered.
- Added synthetic iPhone backup fixtures and extended fixture privacy scanning.
- Added diff-aware E2E mapping for the new importer.

## Safety notes

- No device-pull or backup decryption code is included in this PR.
- No new fact extraction or search path is added.
- The importer refuses implicit `seed_default` source routing unless the caller passes `--source-id`.
- SMS cap hits do not import partial conversation pages.

## Test plan

Passed:

```bash
bun test test/commands/import-iphone.test.ts test/iphone test/ingestion/iphone-backup.test.ts test/e2e/iphone-import-pglite.test.ts
bun run typecheck
bun run verify
bun run check:fixture-privacy
bun run check:test-isolation
```

Also ran `/review` with specialist/adversarial/Codex passes. Issues found there were fixed before opening the PR.

Known: full `bun run test` was attempted and failed in unrelated existing env-sensitive AI/search tests after sourcing `~/.zshrc`; the new iPhone importer tests passed.


time-attack:
This is careful, well-structured work — the backup-root containment, symlink rejection, thin-client refusal, and the decision to keep it off the MCP surface all show real judgment about the risk. I'm still closing it. A 2000-line net-new ingestion subsystem for one vendor's device backup format is a permanent maintenance obligation (Apple's backup schema shifts across iOS releases, and the SMS/contacts parsing carries the drift) that sits well outside where this project is investing, which is retrieval and memory quality rather than device-specific importers. It would also work well as a standalone tool that renders markdown for `gbrain import` to consume, which keeps the parsing burden out of the core and loses you nothing.

PR #1742, opened 2026-06-01
fix: use jsonSchema() wrapper and parameters key for AI SDK tool definitions

### LINKED ISSUE #1739: Non-Anthropic subagent tool loops fail with 'schema is not a function'
## What happened?

When `agent.use_gateway_loop=true` is enabled for non-Anthropic providers (DeepSeek, OpenAI, OpenRouter, etc.), subagent tool loops fail with:

```
[chat(deepseek:...)] schema is not a function
```

## Root Cause

In `src/core/ai/gateway.ts` line 2460, the `chat()` function passes tool definitions to `generateText()` using:

```typescript
```

The Vercel AI SDK expects the `parameters` key with `jsonSchema()`-wrapped values. The `inputSchema` key is silently ignored for openai-compatible providers, causing the SDK to fall through to calling `.schema()` on `undefined`.

## Expected Behavior

Non-Anthropic subagent tool loops should work correctly with the gateway loop.

## Fix

```typescript
// Import jsonSchema from 'ai'
import { embed as aiEmbed, embedMany, generateObject, generateText, jsonSchema } from 'ai';

// Use parameters key with jsonSchema() wrapper
parameters: jsonSchema(t.inputSchema),
```

## Environment
- gbrain version: v0.42.1.0 (bug present since v0.38+ when gateway_loop was added)
- Provider: any non-Anthropic (DeepSeek, OpenAI, OpenRouter, Groq, etc.)
- Config: `agent.use_gateway_loop: true`

justemu:
## Problem

When `agent.use_gateway_loop=true` is enabled for non-Anthropic providers, subagent tool loops fail with `schema is not a function`.

## Root Cause

`gateway.chat()` passes tool definitions to `generateText()` using `inputSchema: { jsonSchema: t.inputSchema }`. The Vercel AI SDK expects the `parameters` key with `jsonSchema()`-wrapped values. The `inputSchema` key is silently ignored for openai-compatible providers.

## Fix

- Import `jsonSchema` from the `ai` package
- Replace `inputSchema: { jsonSchema: t.inputSchema }` with `parameters: jsonSchema(t.inputSchema)`

## Testing

Closes #1739

### DIFF
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -21,7 +21,7 @@
  *     rotation (via configureGateway()) invalidates stale entries.
  */
 
-import { embed as aiEmbed, embedMany, generateObject, generateText } from 'ai';
+import { embed as aiEmbed, embedMany, generateObject, generateText, jsonSchema } from 'ai';
 import { AsyncLocalStorage } from 'node:async_hooks';
 import { listRecipes } from './recipes/index.ts';
 import { createOpenAI } from '@ai-sdk/openai';
@@ -2457,7 +2457,7 @@ export async function chat(opts: ChatOpts): Promise<ChatResult> {
   const tools = (opts.tools ?? []).reduce((acc, t) => {
     acc[t.name] = {
       description: t.description,
-      inputSchema: { jsonSchema: t.inputSchema } as any,
+      parameters: jsonSchema(t.inputSchema),
     };
     return acc;
   }, {} as Record<string, any>);


garrytan:
Thanks for the diagnosis — closing in favor of #1809 (v0.42.19.0). One correction worth noting: this PR switches the key to `parameters`, which is the AI SDK **v4** key. In v6 (`ai@6.0.174`) the tool field is `inputSchema`, and `parameters` is silently ignored, so it would drop the schema. The landed fix keeps `inputSchema` and wraps it with `jsonSchema()`. Crediting your independent diagnosis as Co-Author.

Issue #1729, opened 2026-06-01
gbrain dream disconnects module Postgres singleton mid-cycle before conversation_facts_backfill


mini-sitehub:
**What happened?**
The failure occurs after calibration_profile completes and before conversation_facts_backfill can finish:

  [cycle.calibration_profile] done
  [cycle.conversation_facts_backfill] start
  No database connection: connect() has not been called. Fix: Run gbrain init --supabase or gbrain init --url
  <connection_string>

  The disconnect audit shows an early module-singleton disconnect before normal CLI cleanup:

  {"command":"dream","pid":122068,"caller_stack":"    at disconnect (.../src/core/postgres-engine.ts:207:7)\n    at
  processTicksAndRejections (native:7:39)"}
  {"command":"dream","pid":122068,"caller_stack":"    at disconnect (.../src/core/db.ts:237:5)\n    at async disconnect
  (.../src/core/postgres-engine.ts:223:16)\n    at processTicksAndRejections (native:7:39)"}

  Then normal CLI cleanup runs later:

  {"command":"dream","pid":122068,"caller_stack":"    at disconnect (.../src/core/postgres-engine.ts:207:7)\n    at async
  handleCliOnly (.../src/cli.ts:1143:26)\n    at async main (.../src/cli.ts:160:11)"}

 A full dream cycle should complete successfully, skip disabled phases cleanly, and update source freshness for default.

  At minimum, if a phase fails, the cycle should not lose the active Postgres connection mid-process or leave the
  following phase with:

  No database connection: connect() has not been called.

  Steps to reproduce

  1. Source the environment and run full dream with pipefail:

     set -a; . ~/.hermes/.env; set +a
     set -o pipefail

     gbrain dream --source default 2>&1 \
       | tee /tmp/dream-full-audit-$(date +%s).log
     echo "gbrain=${PIPESTATUS[0]}"

  2. Observe the failure:

     [cycle.conversation_facts_backfill] start
     No database connection: connect() has not been called. Fix: Run gbrain init --supabase or gbrain init --url
     <connection_string>
     gbrain=1

  3. Compare with standalone phase runs, which succeed:

     gbrain dream --source default --phase conversation_facts_backfill
     gbrain dream --source default --phase calibration_profile
     gbrain dream --source default --phase grade_takes
     gbrain dream --source default --phase propose_takes

     Each exits 0 and only emits normal final cleanup disconnect audit rows.

  Environment

  - gbrain version: gbrain 0.42.1.0
[...]

PR #1743, opened 2026-06-01
fix(lint): use dedicated pool (poolSize: 2) to avoid consuming module-level singleton

justemu:
## Problem

When `gbrain lint` runs inside an autopilot cycle, `resolveLintContentSanity()` connects to the module-level singleton pool via `engine.connect({})`. When `engine.disconnect()` runs in the `finally` block, it tears down the shared pool, breaking subsequent phases that depend on the DB connection.

This is a specific trigger of the broader singleton-disconnect pattern reported in #1729.

## Fix

Use `engine.connect({ poolSize: 2 })` to create an instance-level pool. The `disconnect()` in the `finally` block then safely closes only the instance pool without affecting the module singleton.

```typescript
// Before
await engine.connect({});

// After
```

## Testing

Refs #1729

### DIFF
--- a/src/commands/lint.ts
+++ b/src/commands/lint.ts
@@ -312,7 +312,12 @@ async function resolveLintContentSanity(): Promise<LintContentOpts['contentSanit
         database_path: base!.database_path,
       });
       try {
-        await engine.connect({});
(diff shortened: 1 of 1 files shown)


garrytan:

PR #1744, opened 2026-06-01
fix(embed): only reject as user_provided_model_unset when modelId is truly absent

### LINKED ISSUE #1740: embedding preflight: user_provided_model_unset rejects valid user-specified modelId
## What happened?

`gbrain init` with a correctly configured llama-server embedding model fails with `user_provided_model_unset`, even though the model string is fully specified.

Example: `llama-server:text-embedding-qwen3-embedding-0.6b`

## Root Cause

In `diagnoseEmbedding()` (gateway.ts ~line 671), the guard checks:

```typescript
if (
    Array.isArray(tp.models) &&
    tp.models.length === 0 &&
)
```

This only inspects `tp.models` (the touchpoint's static model list) but ignores `parsed.modelId` (the user's actual model name from the config string). When the user specifies a complete model string like `llama-server:text-embedding-qwen3-embedding-0.6b`, `parseModelId()` correctly extracts the modelId, but the guard unconditionally rejects because the touchpoint has an empty static models array.

This is distinct from #1674 (dimensions mismatch error) — this bug triggers even when dimensions are correctly specified.

## Steps to Reproduce

```bash
gbrain init --embedding-provider llama-server \
  --embedding-model llama-server:text-embedding-qwen3-embedding-0.6b \
  --embedding-dimensions 1024
```

Result: `user_provided_model_unset` error despite having a valid modelId.

## Expected Behavior

Should proceed to model instantiation when the user has provided a valid model string with a parseable modelId.

## Fix

Add `!parsed.modelId` to the condition so it only rejects when the user truly hasn't supplied a model:

```typescript
if (
    Array.isArray(tp.models) &&
    tp.models.length === 0 &&
    !parsed.modelId  // only reject when modelId is truly absent
)
```

## Environment
- gbrain version: v0.42.1.0
- Provider: llama-server or any openai-compatible with `user_provided_models: true`

justemu:
## Problem

`gbrain init` with a correctly configured embedding model (e.g., `llama-server:text-embedding-qwen3-embedding-0.6b`) fails with `user_provided_model_unset`, even though the user has supplied a valid model string.

## Root Cause

In `diagnoseEmbedding()`, the `user_provided_model_unset` guard only inspects `tp.models` (the touchpoint static model list) but ignores `parsed.modelId` (the user actual model name). `parseModelId()` correctly extracts the modelId, but the guard unconditionally rejects because the touchpoint has an empty static models array.

## Fix

Add `!parsed.modelId` to the condition so it only rejects when the user truly has not supplied a model:

```typescript
if (
    Array.isArray(tp.models) &&
    tp.models.length === 0 &&
    !parsed.modelId  // only reject when modelId is truly absent
)
```

## Testing

Closes #1740

### DIFF
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -671,7 +671,8 @@ export function diagnoseEmbedding(modelOverride?: string): EmbeddingDiagnosis {
   if (
     Array.isArray(tp.models) &&
     tp.models.length === 0 &&
-    (recipe.id === 'litellm' || isUserProvided)
+    (recipe.id === 'litellm' || isUserProvided) &&
+    !parsed.modelId
   ) {
     return {
       ok: false,


garrytan:

PR #1758, opened 2026-06-02
fix(db): lazy-reconnect module singleton to survive mid-cycle teardown (#1570)

sonlndv:
## Problem

The dream cycle runs its phases sharing the `db.ts` module-singleton postgres connection. A mid-cycle callee (a **module-style engine disconnect between phases**) nulls the singleton, so the later **sync** and **synthesize** phases throw `No database connection: connect() has not been called` and the whole cycle reports `partial`.

Knock-on effect: `partial` (well, the failed phases) meant the freshness signal was unreliable, and `doctor`'s `cycle_freshness` check stayed FAIL ("Source X has never completed a full cycle") because the broken phases dominated the run.

This is the singleton-null bug the disconnect-audit instrumentation in #1570 was added to chase. I captured the culprit by logging full stacks in `db.disconnect()` during a real `gbrain dream --source <id>` run (interpreted, unminified): a module-style `PostgresEngine.disconnect()` fires mid-cycle, nulling `sql` before sync/synthesize reach `getConnection()`.

## Fix

`getConnection()` now **lazily rebuilds** the pool from a remembered URL (`lastConnectedUrl`) instead of throwing when the singleton was torn down earlier in the same process. `postgres()` opens no socket until the first query, so rebuilding from the synchronous `getConnection()` path is safe. Extracted `buildPool()` so `connect()` and the lazy path share identical options.

This is defensive: it self-heals regardless of *which* phase drops the connection, without re-plumbing every caller's engine.

## Verification

- `gbrain dream --source {helmsman,perfeat-mobile,default}` — sync + synthesize now run to completion, exit 0, **0 failed phases** (were 2: sync + synthesize).
- `doctor` `cycle_freshness` flips to **OK** across all 3 sources.
- **52** connection-resilience tests + **64** chunker tests pass.

Refs #1570.


sonlndv:
Closing as superseded. Upstream already fixes this exact singleton-null mid-cycle bug (#1570) via two landed commits:

- v0.41.27.0 / #1608 — "withRetry self-heals on null singleton + disconnect audit (closes #1570)"
- v0.42.5.0 / #1678 — "cycle lint DB-disconnect": the lint phase's resolveLintContentSanity created+disconnected a module-style engine that nulled the shared db singleton mid-cycle, breaking later phases with 'connect() has not been called'. Fixed by reusing the shared connection.

Upstream's caller-side fix is canonical; my lazy-reconnect in getConnection() is redundant. Thanks!

PR #1760, opened 2026-06-02
fix(cycle): cache empty propose_takes results — stop re-extracting unchanged pages every cycle

jonahberg:
## Summary

The dream-cycle `propose_takes` phase re-extracts the **same unchanged pages on every autopilot tick**, because pages whose extraction returns no proposals never enter the idempotency cache.

`runPhase()` walks the top-100 most-recently-updated pages, checks `take_proposals` for `(source_id, page_slug, content_hash, prompt_version)`, and calls the LLM on cache misses. But the INSERT only happens inside `for (const p of proposals)` — **when `proposals` is `[]`, no row is written**, so the same page is a cache miss again on the next cycle, ~5 minutes later, forever.

Observed on a ~42K-page production brain (Supabase, autopilot every ~5 min): **57,885 propose_takes call submissions over 11 days** — near-identical batches of ~95 calls in 92% of cycles, constant volume regardless of new content — with **zero takes produced** (`gbrain takes list` empty). Ledger evidence (daily call counts from `~/.gbrain/audit/dream-budget-*.jsonl`) available on request; it contains no page content.

The per-cycle `$5` budget cap never fires on this pattern: each cycle self-limits at ~$1.14 (batch size × per-call estimate), so the damage accrues across 25-219 cycles/day rather than within one cycle.

## The fix

When the extractor returns an empty array, insert one **sentinel row** under the same composite idempotency key:

- `kind = 'empty_result'` — distinct from every real proposal kind (`fact`/`take`/`bet`/`hunch`), so it can never be confused with a real proposal and any future reader can filter it
- `claim_text = ''`, `weight = 0`, `holder = 'system'`
- `status = 'superseded'` — terminal, so it never surfaces as pending review

The page is then skipped by the existing idempotency check until its `content_hash` actually changes. Safe by construction: the only reader of `take_proposals` in the codebase is this phase's own idempotency check.

Also adds an `empty_results_cached` counter to `ProposeTakesResult` and the phase summary line, so operators can see the cache working.

## Tests

Two new cases in `test/propose-takes.test.ts` (existing mock-engine conventions):

1. **empty extraction result writes a sentinel cache row** — asserts exactly one INSERT with `kind='empty_result'`, `claim_text=''`, `status='superseded'`, and `empty_results_cached=1`
2. **second run after an empty result is a cache hit** — asserts the extractor (LLM) is never called again and no further INSERTs happen

All 27 tests in the file pass; `bun run verify` green (29/29 checks).

## Notes

- No schema changes — the sentinel uses existing columns and the existing unique index.
- Existing rows and real proposals are unaffected; `ON CONFLICT DO NOTHING` preserves race safety.
- A natural follow-up (not in this PR): gate proposal on `content_hash` change rather than `updated_at` recency, and/or a daily (not per-cycle) dream-cycle budget cap, since per-cycle caps structurally cannot catch per-tick grinders.

### DIFF
--- a/src/core/cycle/propose-takes.ts
+++ b/src/core/cycle/propose-takes.ts
@@ -152,6 +152,9 @@ export interface ProposeTakesResult {
   cache_hits: number;
   cache_misses: number;
   proposals_inserted: number;
+  /** Pages whose extraction returned no proposals, recorded as sentinel cache
+   *  rows so they are not re-extracted until their content changes. */
+  empty_results_cached: number;
   budget_exhausted: boolean;
   warnings: string[];
 }
@@ -314,6 +317,7 @@ class ProposeTakesPhase extends BaseCyclePhase {
       cache_hits: 0,
       cache_misses: 0,
       proposals_inserted: 0,
+      empty_results_cached: 0,
       budget_exhausted: false,
       warnings: [],
     };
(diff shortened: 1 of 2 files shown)


rdguidry:
Production confirmation (0.42.39-era master, Postgres engine, autopilot running continuously): before this patch, propose_takes was re-extracting ~90 zero-proposal pages every cycle — 3,578 dream-budget submissions over Jun 8–11 to produce a queue of 93 proposals across 81 pages.

Cherry-picked 149b282b: clean apply, 27/27 propose-takes tests pass. Measured from the dream-budget ledger after restart: first cycle = 91 calls (the expected one-time pass that writes the sentinel rows), next cycle = 1 call (a genuinely new page), every cycle since = 0. The hourly re-extraction is gone.

One interaction heads-up for whoever merges: #1931 fixes the multi-claim drop by making take_proposals_idempotency_idx non-unique. Once that index is non-unique, this PR's sentinel `INSERT … ON CONFLICT (…) DO NOTHING` raises in Postgres (ON CONFLICT requires a unique or exclusion constraint). Both fixes are correct individually; together they need the ON CONFLICT clause dropped from the sentinel insert — the cache check immediately above already guarantees the tuple is absent. We run both that way.

PR #1763, opened 2026-06-02
fix(cli): drain fire-and-forget facts queue (and bound embed fetch) so multi-chunk capture can't pin the PGLite lock

### LINKED ISSUE #1762: CLI_ONLY commands (e.g. `gbrain capture`) hang after completing on multi-chunk pages, pinning the PGLite single-writer lock
> **⚠️ UPDATE (2026-06-02): the “Root cause” section below was DISPROVEN.** The real cause is a fire-and-forget `facts:absorb` Haiku job that is not drained before `engine.disconnect()`, spinning PGLite’s `db.close()` into a 100%-CPU busy-loop that pins the lock (the force-exit timer can never fire, so it cannot fix this). See the corrected analysis in the comments and in #1763. Original report preserved below for history.

---

### Version
gbrain 0.42.1.0 (local `bun-link` install, PGLite engine, `search.mode=tokenmax`)

### Summary
`gbrain capture` on a **multi-chunk page** finishes all its work — writes the page, embeds the chunks, prints the receipt — and then **never exits**. The process keeps PGLite's single-writer lock held, so every subsequent gbrain command fails with `Timed out waiting for PGLite lock` until the zombie process is `SIGKILL`ed. Single-chunk pages (person/company) exit cleanly.

### Repro
1. `gbrain config set search.mode tokenmax`
2. `gbrain capture --file <a long, multi-chunk markdown page> --slug meetings/foo --type meeting --quiet`
3. Observe: the receipt prints, but the process hangs indefinitely (does not return to the shell).
4. In another shell, any `gbrain` command now fails: `Timed out waiting for PGLite lock`.
5. `kill -KILL` the hung capture → the lock frees and gbrain works again.

Notably the content + embeddings DO land before the hang (`gbrain embed --stale --dry-run` reports 0 stale after the kill), so this is purely a **process-teardown** problem, not a write problem.

### Root cause
The `CLI_ONLY` dispatch path (`handleCliOnly` in `src/cli.ts`) disconnects the engine in its `finally`, but — unlike the op-dispatch path (`src/cli.ts` ~244-256 / 315-316, the `shouldForceExitAfterMain()` + unref'd hard-deadline + post-disconnect `process.exit(0)` contract) — it has **no force-exit-after-main safety net**. So when a lingering async handle stays open (or `engine.disconnect()` itself hangs on PGLite, exactly the C13 case the op-dispatch comment at ~239 warns about), nothing forces the process to exit, and the held PGLite connection keeps its exclusive directory lock forever.

Per-chunk Haiku synopsis is *not* the cause — the inline import path downgrades `per_chunk_synopsis` to title-tier (`src/core/import-file.ts:575`) and makes no Haiku calls. And `capture`/`put_page` never start a MinionWorker, so it isn't the worker's timers either. It's specifically the missing teardown contract on the `CLI_ONLY` path, triggered by whatever extra handle multi-chunk pages leave open (embedder keep-alive socket is the prime suspect).

### Fix
Give `handleCliOnly`'s `finally` the same force-exit-after-main contract the op-dispatch path already has: gate on `shouldForceExitAfterMain()` (already imported; excludes the `serve` daemon), install an unref'd 10s hard-deadline `setTimeout(() => process.exit(0))` so a hung `disconnect()` can't defeat exit, then `process.exit(0)` after `engine.disconnect()` returns. Keeps the existing `command !== 'serve'` no-disconnect rule.

I've applied this locally and verified the previously-hanging multi-chunk capture now self-exits with code 0, no zombie, lock immediately free. Happy to open a PR if useful.

### Impact
Hits anyone on `tokenmax` (or any config that produces multi-chunk pages) ingesting larger pages via `capture`/`import` — one hung ingest bricks all subsequent gbrain commands until manually killed.

ElliotDrel:
Closes #1762.

> **Update (2026-06-02):** the original root-cause analysis in this PR and in #1762 was **disproven** by deeper instrumentation. The real cause is a fire-and-forget facts-queue job, not a missing teardown contract / leaked embedder socket. Full corrected analysis below; the force-exit patch is retained only as defense-in-depth.

### Problem
`gbrain capture` (and other `CLI_ONLY` write commands) on a **multi-chunk page** writes the page, embeds the chunks, prints the receipt — then **never exits**. The process keeps PGLite's single-writer lock held, so every later `gbrain` command fails `Timed out waiting for PGLite lock` until it is `SIGKILL`ed. Single-chunk pages exit cleanly. Content + embeddings land before the hang, so it is purely a teardown problem.

### Root cause (corrected)
`put_page` fires a **fire-and-forget `facts:absorb` Haiku job** into the bounded `FactsQueue` (`mode:'queue'`) *after* printing the receipt. On a multi-chunk page that job is still in flight when `handleCliOnly`'s `finally` tears the engine down. Unlike the op-dispatch `finally`, `handleCliOnly` did **not drain that queue** before `engine.disconnect()`.

`disconnect()` nulls `_db` out from under the in-flight job. The job's "PGLite not connected" error path re-pumps the queue via `queueMicrotask`, and that microtask storm interleaving with PGLite's WASM message pump spins `db.close()` into a **100%-CPU userspace busy-loop that never returns** (in isolation `db.close()` returns in ~9ms — confirmed via `/proc/<pid>/stat` CPU sampling).

Because the busy-loop pins the single JS thread, the unref'd 10s force-exit `setTimeout` from the earlier patch **can never fire** (a same-thread timer needs an event-loop turn the busy WASM call never yields; a worker-thread watchdog can't preempt it either). So `disconnect()`'s lock-release in its own `finally` is never reached, and the PGLite single-writer lock stays pinned indefinitely.

This is why the force-exit-after-main approach (the prior commit) could not fix it — it treats the symptom (process won't exit) but the thread is wedged before any exit path can run.

> Note: this corrects the earlier claim that "per-chunk Haiku synopsis is not the cause." A Haiku job **is** the trigger — but it is the `facts:absorb` queue job, not the per-chunk synopsis path.

### Fix
1. In `handleCliOnly`'s `finally`, **drain the fire-and-forget work before disconnect** — `awaitPendingLastRetrievedWrites()` + `getFactsQueue().drainPending({ timeout: 2000 })` — mirroring the op-dispatch `finally`. The Haiku job then settles against a still-live engine and `db.close()` returns in ~10ms instead of busy-looping.
2. The force-exit hard-deadline from the prior commit is **kept as defense-in-depth** (it just can't be the only safeguard).
3. **`gateway.ts` `withEmbedFetchTimeout`** — fixes a separate latent hang: a stalled `openai-compatible` embed socket inside the inline `embedBatch` `await` hangs capture *during* work (before disconnect), also pinning the lock. The Vercel AI SDK `maxRetries` only fires on a *settled* error, so a half-open socket is unbounded without a per-request timeout. Adds a 60s `AbortSignal.timeout`, composed via `AbortSignal.any` so it never clobbers a caller-supplied signal.

`src/cli.ts` + `src/core/ai/gateway.ts` only.

### Testing
Reproduced the hang on a multi-chunk meeting-page capture (`tokenmax`): pre-fix it hangs every run (SIGKILLed at timeout, lock pinned). Post-fix: **5/5 clean** — exit 0, ~3s, no zombie process, PGLite lock immediately free (`gbrain get`/`doctor` run right after with no timeout). Root cause was localized with `MARK` probes (since removed) + `/proc` CPU sampling.

### Impact
Hits anyone whose config produces multi-chunk pages (e.g. `tokenmax`) ingesting larger pages via `capture`/`import` — one hung ingest bricks all subsequent `gbrain` commands until manually killed.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -1747,16 +1747,50 @@ async function handleCliOnly(command: string, args: string[]) {
       }
     }
   } finally {
-    // v0.42.1.0 LOCAL PATCH (see PATCH.md): give the CLI_ONLY dispatch path
-    // the same force-exit-after-main contract the op-dispatch path already has
-    // (cli.ts ~244-256 / 315-316). Without it, a write command like
-    // `gbrain capture` on a MULTI-CHUNK page leaves an async handle open (or
-    // engine.disconnect() itself hangs on PGLite — see the C13 note at ~239),
-    // so the process never exits and pins PGLite's single-writer lock. Every
-    // later gbrain command then dies with "Timed out waiting for PGLite lock"
-    // until the zombie is SIGKILLed. Bound disconnect with an unref'd hard
-    // deadline, then force-exit once it returns. Daemons (serve) are excluded
-    // by shouldForceExitAfterMain(); serve also keeps its no-disconnect rule.
+    // v0.42.1.0 LOCAL PATCH (see PATCH.md item #3): the CLI_ONLY dispatch path
+    // (which owns `gbrain capture`) lacked the op-dispatch finally's
+    // drain-before-disconnect contract (cli.ts ~290-317). That was the TRUE
+    // root cause of the multi-chunk capture hang:
+    //
+    //   `put_page` fires a fire-and-forget facts:absorb job into the bounded
+    //   FactsQueue (operations.ts ~875, mode:'queue') AFTER printing the
+    //   receipt. On a multi-chunk page that in-flight Haiku job is still
+    //   running when this finally tears the engine down. `engine.disconnect()`
+    //   nulls `_db` out from under the job; the job's "PGLite not connected"
+    //   error path re-pumps the queue via queueMicrotask, and that microtask
+    //   storm interleaving with PGLite's WASM message pump spins `db.close()`
+    //   in a 100%-CPU userspace busy-loop that NEVER returns. Because the
+    //   busy-loop pins the JS thread, the unref'd hard-deadline setTimeout
+    //   below can never fire (a same-thread timer needs an event-loop turn the
+    //   busy WASM call never yields — verified: a Worker-thread watchdog can't
+    //   preempt it either). disconnect()'s lock-release in its finally is
+    //   therefore never reached, so the PGLite single-writer lock stays pinned
+    //   and every later gbrain command dies "Timed out waiting for PGLite lock"
+    //   until the zombie is SIGKILLed.
+    //
+    // Fix: DRAIN the fire-and-forget work (facts queue + any pending
+    // last-retrieved writes) BEFORE disconnect, so the job finishes cleanly
+    // against a live engine and `db.close()` returns in ~10ms instead of
+    // busy-looping. This mirrors the op-dispatch finally exactly. The
+    // force-exit hard-deadline below is KEPT as defense-in-depth, but it is no
+    // longer the only thing standing between the user and a wedged brain — and
+    // it can't be (it can't preempt a WASM busy-loop). Daemons (serve) keep
+    // their no-disconnect rule and are excluded from force-exit by
+    // shouldForceExitAfterMain().
+    if (command !== 'serve') {
+      try {
+        const { awaitPendingLastRetrievedWrites } = await import('./core/last-retrieved.ts');
+        await awaitPendingLastRetrievedWrites();
+      } catch { /* best-effort; never block disconnect on drain failure */ }
+      try {
+        // 2s timeout: pages that enqueue no facts pay one ~0ms check; capture/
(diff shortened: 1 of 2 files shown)


garrytan:
Superseded by #1810 (v0.42.11.0). That PR incorporates your drain-before-disconnect fix and the embed-fetch timeout, and hardens both: the four fire-and-forget sinks (last-retrieved, facts, search-cache, eval-capture) are unified behind a background-work registry drained on every CLI exit path, and the residual hung-Haiku hole is closed by an awaited facts `shutdown()` abort belt. Your corrected root-cause diagnosis and the fix are credited via `Co-Authored-By: ElliotDrel`. Thank you for the precise instrumentation that found the real cause. Closing in favor of #1810.

PR #1765, opened 2026-06-02
fix: ignore clone health for path-only sources

Goose135:
## Summary
- Treat path-only local sources without `remote_url` as `clone_state: not-applicable`
- Preserve clone health validation for remote-managed sources with `remote_url`
- Update `sources-ops` regression coverage for the path-only contract

## Verification
- Static scan on scoped diff: clean
- Independent reviewer verdict: passed, no security concerns or logic errors
- `bun test test/sources-ops.test.ts --timeout 30000` → 24 pass / 0 fail

## Context
This addresses the default local GBrain source being reported as `clone_state: corrupted` even though it is a path-only source rather than a managed clone.


Goose135:
Closing at Goose's request.

PR #1766, opened 2026-06-02
fix: allow user-provided embedding dimensions

NYTEMODEONLY:
## Summary

Fixes local/user-provided embedding backends whose models are not statically enumerable by GBrain recipes.

`llama-server` and `litellm` intentionally declare `models: []` and `user_provided_models: true` because the served embedding model comes from the user's local server/proxy configuration. These recipes already require an explicit `provider:model` and `--embedding-dimensions <N>`, but the current preflight path rejects that explicit dimension as an unsupported "custom dimension" before init/reinit can create the schema.

This PR treats the explicit dimension for `user_provided_models` recipes as the schema contract while preserving the existing positive-integer and pgvector max-dimension guards. Runtime vector-length validation still protects storage if the backend returns vectors with a different length.

It also fixes the embedding gateway availability diagnostic so an explicit user-provided model id like `llama-server:bge-small-en-v1.5` is not reported as `user_provided_model_unset` just because the recipe's static model list is empty.

## Changes

- Allow `resolveSchemaEmbeddingDim()` to accept explicit dimensions for `user_provided_models` recipes.
- Keep missing/invalid dimensions rejected for local/user-provided recipes.
- Make `diagnoseEmbedding()` consider empty-model recipes available when `provider:model` includes an explicit model id.
- Add unit coverage for `llama-server` and `litellm` user-provided model/dimension behavior.
- Add an E2E init regression test for `llama-server:bge-small-en-v1.5 --embedding-dimensions 384`.
- Clarify local `llama-server` docs for OpenAI-compatible local services/proxies.

## Real-world validation

This was found while wiring GBrain to an existing local OpenAI-compatible BGE embedding service. After the local patch, a PGLite brain successfully initialized and embedded 1,327/1,327 chunks using:

```bash
LLAMA_SERVER_BASE_URL=http://127.0.0.1:8090/v1 \
  gbrain reinit-pglite \
  --embedding-model llama-server:bge-small-en-v1.5 \
  --embedding-dimensions 384 \
  --yes
```

## Test plan

- [x] `bun test --timeout 30000 test/embedding-dim-check.test.ts test/gateway-embed-model-override.test.ts`
- [x] `bun test --timeout 30000 test/e2e/init-fresh-pglite.test.ts --test-name-pattern 'user-provided local embedding model accepts explicit dimensions'`
- [x] `NODE_OPTIONS=--max-old-space-size=4096 bun run typecheck`
- [x] Static added-line scan: no secret-looking assignments; no added `eval`/`exec`/shell-exec patterns.
- [x] Independent reviewer pass: no security concerns or logic errors.

### DIFF
--- a/docs/integrations/embedding-providers.md
+++ b/docs/integrations/embedding-providers.md
@@ -147,7 +147,7 @@ Recipe ships with `nomic-embed-text` (768d, recommended), `mxbai-embed-large` (1
 
 `llama.cpp`'s `llama-server --embeddings` endpoint. No env required. Optional `LLAMA_SERVER_BASE_URL` (default `http://localhost:8080/v1`) and `LLAMA_SERVER_API_KEY`.
 
-User-driven models: launch llama-server with `--model <gguf-path> --embeddings`, then run `gbrain init --embedding-model llama-server:<your-id> --embedding-dimensions <N>`. The recipe refuses the implicit shorthand `--model llama-server` because there's no canonical first model.
+User-driven models: launch llama-server with `--model <gguf-path> --embeddings`, then run `gbrain init --embedding-model llama-server:<your-id> --embedding-dimensions <N>`. The recipe refuses the implicit shorthand `--model llama-server` because there's no canonical first model. For OpenAI-compatible local services such as BGE/TEI-compatible proxies, set `LLAMA_SERVER_BASE_URL` to the service's `/v1` endpoint and use the model id reported by `/v1/models`; the explicit `--embedding-dimensions` value becomes the schema contract.
 
 ### LiteLLM proxy (universal escape hatch)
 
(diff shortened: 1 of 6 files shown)


felipeyanez:
Hitting this too. The `llama-server` recipe can't init for a non-registry dim: it requires explicit dims and rejects every value.

```
gbrain init --pglite --embedding-model llama-server:all-minilm-l6-v2 --embedding-dimensions 384
  -> "does not support custom dimensions 384 (this model only emits its default vector size)"

gbrain init --pglite --embedding-model llama-server:all-minilm-l6-v2
  -> "requires --embedding-dimensions <N>"
```

Regression bisects to v0.37.10.0 (#1278): it added the `resolveSchemaEmbeddingDim` preflight, which validates the requested dim against the recipe's `default_dims`. Recipes with `default_dims: 0` (llama-server, litellm) then have no valid dim, so every value is rejected. Broken from v0.37.10.0 through 0.42.8.0.

Workaround for now: pin to v0.37.9.0, the last release before the regression.

garrytan:

PR #1769, opened 2026-06-02
[codex] fix init schema-pack validation

cosin2077:
## What changed

- Route `gbrain init --help` through the init command's detailed help instead of the generic CLI-only stub.
- Validate `--schema-pack` during init before creating the brain or writing config.
- Surface a structured `invalid_schema_pack` JSON error for `--json` callers and add regression coverage.
- Update init help text for `--schema-pack`, `--mcp-only`, and the large-brain Supabase recommendation behavior.

## Why

`gbrain init --schema-pack <typo>` previously exited successfully and persisted a broken `schema_pack`, leaving follow-up schema-pack-aware commands to fail later. Separately, the detailed `init --help` handler existed but was unreachable because CLI dispatch intercepted it with the generic stub.

## Validation

- `bun test test/cli.test.ts --timeout 120000`
- `bun test test/e2e/init-fresh-pglite.test.ts --timeout 240000`
- `bun run typecheck`

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -40,6 +40,9 @@ const CLI_ONLY = new Set(['init', 'reinit-pglite', 'upgrade', 'post-upgrade', 'c
 // excluded from the generic short-circuit so detailed per-command and
 // per-subcommand usage stays reachable.
 const CLI_ONLY_SELF_HELP = new Set([
+  // `gbrain init --help` has a detailed, state-safe help path in runInit.
+  // Keep it reachable instead of the generic CLI-only one-line stub.
+  'init',
   'upgrade', 'post-upgrade', 'check-update',
   'embed', 'config',
   'skillpack', 'skillpack-check',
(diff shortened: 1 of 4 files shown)


time-attack:
Merging. Confirmed on master that `--schema-pack` is read and persisted with zero validation, so a typo silently writes a broken `schema_pack` and every downstream schema-pack-aware command fails later with a confusing error — that's a real correctness bug, and reusing the existing `loadActivePack`/`UnknownPackError` path instead of inventing a new validator is the right shape. The e2e assertion that no config.json is written on the failure path is the part that sells it. Please rebase and re-run `test/cli.test.ts` since `init --help` text has drifted since you opened this.

cosin2077:
Hi @time-attack — thanks again for confirming the bug and the approach. I noticed the PR was closed immediately after the “Merging” comment, but it was not merged. Was that close intentional? If the project still wants the fix, I am happy to rebase onto current main and rerun test/cli.test.ts. Please let me know whether you would like that update and whether the PR should be reopened.

PR #1771, opened 2026-06-02
fix(write-through): route writes to the page's own source repo (not global sync.repo_path)

sonlndv:
## Problem

`writePageThrough()` resolves the write target from the single global `sync.repo_path` config key. That key is **last-writer-wins** — every `gbrain sync <dir>` / import overwrites it (`import.ts: setConfig('sync.repo_path', dir)`). On a multi-source brain, whichever source synced most recently captures the global write path.

Result: a page belonging to source **A** gets its `.md` written into source **B**'s working tree — e.g. `<repoB>/.sources/A/<slug>.md`. A cross-source file leak that the per-source isolation model is supposed to prevent. (Found in production on a 5-source brain: an `agent-memory/backend/*` page landed under a different project's `.sources/`.)

## Fix

Resolve the base path from the **page's own source** (`sources.local_path` for `opts.sourceId`) and write at the plain `<repo>/<slug>.md` path. Fall back to the global `sync.repo_path` — keeping the legacy `.sources/<id>/` namespacing — only for pure-DB sources with no `local_path`, so pre-existing single-repo behavior is preserved.

This makes the write self-correct regardless of which source synced last, which device/agent wrote, or CLI-vs-MCP — enforcing per-source isolation at the write layer and scaling to any number of registered sources with zero config.

## Verification

Live on a 5-source brain:
- page scoped to `perfeat-mobile` → `<perfeat-mobile>/<slug>.md` (own repo, no `.sources/` nesting)
- page scoped to `default` → `<default-repo>/<slug>.md`

Note: `test/write-through.test.ts` uses `PGLiteEngine`, which times out in my environment **on clean master too** (pre-existing, unrelated) — so I verified functionally against the live Postgres brain.

Relates to the multi-source isolation work (#861, #876).

### DIFF
--- a/src/core/write-through.ts
+++ b/src/core/write-through.ts
@@ -3,9 +3,11 @@
  *
  * After a page row lands in the DB (via importFromContent / putPage), this
  * renders the row to markdown via `serializePageToMarkdown` and writes it to
- * `sync.repo_path` so the brain repo has a committable `.md` artifact that
- * round-trips cleanly through `gbrain sync`. The file is rendered FROM the DB
- * row, so the two sinks cannot diverge.
+ * the PAGE'S SOURCE repo (`sources.local_path` for `opts.sourceId`), falling
+ * back to the global `sync.repo_path` only for pure-DB sources with no
+ * local_path. The file is rendered FROM the DB row, so the two sinks cannot
+ * diverge. Routing per-source (rather than off the single global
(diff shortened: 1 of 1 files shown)


time-attack:
Thanks for this, @sonlndv! 🙏

What this does: This PR fixes a data-leak bug where pages from one source could be written into another source's working tree on multi-source brains. It routes write-through disk operations to each page's assigned source repo instead of the global…

PR #1778, opened 2026-06-02
fix(db-lock): bind release to lock acquisition ownership

shandutta:
## Summary
- Bind DB lock `refresh()` / `release()` cleanup to the specific acquisition timestamp, not just `(id, holder_pid)`.
- Prevent a stale same-process lock handle from deleting a newer live takeover of the same lock row.
- Add a regression test covering same-PID expired-lock takeover behavior.

## Why
Long-running worker processes can execute multiple jobs under the same PID. If an old handle's `finally` path runs after a same-PID takeover, the previous predicate could delete the newer live lock because it matched only `(id, holder_pid)`. Including `acquired_at` makes the lock handle an ownership token for the specific acquisition.

## Test plan
- `bun install`
- `bun test test/db-lock-release-ownership.test.ts test/db-lock-per-source.test.ts`

### DIFF
--- a/src/core/db-lock.ts
+++ b/src/core/db-lock.ts
@@ -80,7 +80,7 @@ export async function tryAcquireDbLock(
     // `gbrain sync --break-lock --max-age <s>` uses last_refreshed_at (not
     // acquired_at) to identify wedged-but-alive holders without stealing
     // healthy long-running holders that are actively refreshing.
-    const rows: Array<{ id: string }> = await sql`
+    const rows: Array<{ id: string; acquired_at: Date | string }> = await sql`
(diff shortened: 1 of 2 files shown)


time-attack:
Merging this one — the ownership hole is real and still present on master: every release/refresh/cleanup predicate keys on `(id, holder_pid)` alone, so a starved same-PID handle can delete a newer live takeover and drop mutual exclusion on a source. Binding the handle to its `acquired_at` is the right-sized fix and the regression test pins it. One thing to fix on rebase: master gained the `last_refreshed_at < NOW() - stealGraceSeconds` steal-grace clause, so your test's `NOW() - INTERVAL '1 second'` backdate won't allow the takeover anymore — widen it past the grace window. Thanks for the careful write-up.

PR #1790, opened 2026-06-02
fix: scope exact get_page to federated sources

colin-atlas:
## Summary

Fixes exact `get_page` reads so they honor OAuth federated read scope (`auth.allowedSources`) instead of only the scalar `ctx.sourceId`.

## Problem

Federated OAuth clients can read broader source sets through list/search paths, but exact `get_page` was still scoped only by `ctx.sourceId`. A client with `source=shared` and `federated_read=shared,internal` could not exact-read an internal page by id, even though broader read surfaces honored the federated scope.

## Changes

- Thread `sourceScopeOpts(ctx)` into exact `get_page`.
- Add `sourceIds` support to PGLite/Postgres `getPage`.
- Preserve back-compat for unscoped internal callers.
- Load tags from the concrete returned page's `source_id`.
- Add regression coverage for exact `get_page` honoring `AuthInfo.allowedSources` over scalar `ctx.sourceId`.

## Verification

- `bun test test/operations-fuzzy-source-scope.test.ts` — 7 passed
- `bun test test/e2e/source-isolation-pglite.test.ts` — 14 passed

### DIFF
--- a/src/core/operations.ts
+++ b/src/core/operations.ts
@@ -517,25 +517,17 @@ const get_page: Operation = {
     const slug = p.slug as string;
     const fuzzy = (p.fuzzy as boolean) || false;
     const includeDeleted = (p.include_deleted as boolean) === true;
-    // v0.31.8 (D20): thread ctx.sourceId through read-side ops. Only pass
-    // sourceId when it's set on ctx — when unset (local CLI default chain
-    // resolves to no source), the engine two-branch query falls through to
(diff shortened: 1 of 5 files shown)


garrytan:
Thanks @colin-atlas — exact get_page federated scoping landed in master via #1999 (v0.42.37.0): get_page routes through `sourceScopeOpts(ctx)` and the engine getPage honors `source_id = ANY(...)`, with federated-scope tests. Your diff is effectively already in master, so closing as superseded.

Issue #1738, opened 2026-06-01
gbrain agent run: --detach / flags after the prompt get swallowed into the prompt string


garrytan-agents:
## Summary
`gbrain agent run "<prompt>" --detach` submits a subagent job whose prompt is `"<prompt> --detach"` — the flag is captured as part of the prompt instead of being consumed as a flag.

## Repro
```
$ gbrain agent run "test prompt" --detach
submitted: job 11407 (subagent)
11407

$ gbrain jobs get 11407
Job #11407: subagent (WAITING)
  Data: {"prompt":"test prompt --detach"}   <-- flag leaked into prompt
```

## Cause (src/commands/agent.ts)
The prompt is assembled by joining positional args, but the flag scanner doesn't strip known flags from the positional set before joining. `isKnownFlag()` only matches `s.startsWith('--')`, and `--detach` *is* `--`-prefixed, so it should be excluded — but in the `run` path the prompt is built from the raw remainder, so trailing flags land in the prompt. (`--follow`/`--no-follow`/`--detach` are documented in the help and parsed in the flag switch at agent.ts:122-124, but the prompt-assembly step doesn't exclude them.)

## Expected
Flags anywhere in the argv (before OR after the prompt) are consumed as flags; the prompt is only the non-flag positional text. `gbrain agent run "x" --detach` should yield `{"prompt":"x"}` and detach.

## Severity
Low (cosmetic prompt pollution) but it silently corrupts the agent's instruction and makes `--detach`/`--follow` unreliable depending on arg order — worth a quick fix since `agent run` is the documented canonical subagent entrypoint.

## Note
Separately, the docs/skill describe `gbrain agent run` as the canonical LLM-job entrypoint; it works, but `gbrain agent --help` (no subcommand) prints only the generic top-level stub rather than the agent help block — minor discoverability gap (the real help is reachable via `gbrain agent run --help` path quirks). Bundling here as an FYI, not the main ask.

time-attack:
Closing as fixed on master (see bb2e88c4 / #2287). Please reopen if still reproducible on current HEAD.

Issue #1717, opened 2026-06-01
Embedding: `gbrain embed`/`import` mislabels content_chunks.model as the gateway default (v0.41.x)


ChenyqThu:
**Version**: gbrain 0.41.14.0 (Postgres engine)

## Summary

After `gbrain embed --all` / `gbrain embed <slug>` / `gbrain import`, every embedded chunk's `content_chunks.model` column is written as `DEFAULT_EMBEDDING_MODEL` (`zeroentropyai:zembed-1` in v0.36+) **regardless of the model that actually produced the vector**. The vectors themselves are correct (they go through `getEmbeddingModel()`); only the stored `model` label is wrong.

## Root cause

`src/commands/embed.ts` builds the `ChunkInput[]` for `upsertChunks` **without a `model` field** (e.g. `embedPage` ~336-342, `embedOnePage` ~435-441):

```ts
const updated: ChunkInput[] = chunks.map(c => ({
  chunk_index: c.chunk_index,
  chunk_text: c.chunk_text,
  chunk_source: c.chunk_source,
  embedding: embeddingMap.get(c.chunk_index) ?? undefined,
  token_count: c.token_count || Math.ceil(c.chunk_text.length / 4),
  // model: <omitted>
}));
```

`upsertChunks` (`src/core/postgres-engine.ts:1868`) then writes `chunk.model || DEFAULT_EMBEDDING_MODEL`, so the column is mislabeled with the gateway default. This is misleading for `gbrain doctor`, for mixed-model detection, and for anyone reading the `model` column to learn which embedding space a chunk lives in (it can read e.g. `zeroentropyai:zembed-1` while the vector is actually OpenAI `text-embedding-3-large`).

## Suggested fix

Thread the resolved model (`getEmbeddingModel()`, the same value used for the batch) into the `model` field of the upserted chunks.

time-attack:
Great catch, and it's since been fixed — the fix went in exactly where you pointed, just on the sink side rather than the caller side. `upsertChunks` now resolves the live gateway model at runtime (`src/core/postgres-engine.ts:2436-2441`, `gw.getEmbeddingModel()`) and stamps `chunk.model || resolvedModel` at `:2473`, so a `ChunkInput` built without a `model` field gets the model that actually produced the vector instead of the compiled `DEFAULT_EMBEDDING_MODEL`. Shipped in `5aa4795c` (#2846) and relanded as `e1919fab` (#3343). Closing as already fixed — thanks for the precise root-cause writeup.

PR #1799, opened 2026-06-03
fix(schema-pack): parseYamlMini fails loud on block scalars instead of silently truncating (#1750)

### LINKED ISSUE #1750: parseYamlMini silently drops every key after a block scalar — gbrain-recommended ships with 0 page_types
## Symptom

```
```

## Root cause

## Verified fix

```
```

## Why this matters

## Repro

```
Page types: 0
```
```ts
raw.page_types   // undefined
raw.description  // "|"
```

harjothkhara:
## Problem

Fixes #1750.

`parseYamlMini` (`src/core/schema-pack/loader.ts`) documents block scalars (`|`, `>`) as unsupported, but it didn't actually *reject* them — it silently mis-parsed. A `description: |` line was read as the literal string `"|"`, then the indented block body had `indent > baseIndent` and tripped the `break` in `parseMapping`, **silently dropping `page_types` and every later top-level key** with no error.

This bit the bundled **`gbrain-recommended`** pack — the documented "operational brain" starter. It uses `description: |`, so:

```
Page types: 0          # declares 12; all silently dropped
```

```ts
parseYamlMini(recommended).page_types  // undefined
parseYamlMini(recommended).description // "|"
```

The parser's own header lists block scalars under "Rejected by design… ship JSON" — but it didn't enforce that; it corrupted the rest of the manifest instead.

## Fix

1. **`parseScalar` now throws a clear, actionable error** when a bare block-scalar indicator (`|`, `>`, including chomping/indent forms like `|-`, `>+`, `|2`) reaches it — enforcing the parser's documented contract and turning a silent footgun into a one-line diagnostic for the next pack author.
2. **Inline `gbrain-recommended.yaml`'s description** so the shipped pack loads its 12 page_types.

I chose *fail-loud + fix the one shipped pack* over expanding the parser to support block scalars, to stay aligned with the parser's stated narrow-by-design intent. The plain inline-only fix was rejected because it leaves the silent-truncation footgun for the next author.

## Safety

- **Round-trip safe:** `emitYamlScalar` (`mutate.ts`) already emits any multi-line / `|` / `>` string as a JSON-quoted single line — never as a block scalar — so the `mutate.ts` emit→parse round-trip is unaffected.
- **No false positives:** the guard only matches a *lone* indicator (`^[|>][0-9+-]*$`). A real value like `> 50% growth` or `>=x` is unaffected, and a quoted `"|"` still parses to the string `"|"` (handled earlier in `parseScalar`).
- Only `gbrain-recommended.yaml` used a block scalar; the other 6 bundled packs are unaffected.

## Tests (`test/schema-pack-loader.test.ts`)

- Block scalars throw a clear error (4 forms: `|`, `>`, `|-`, `|2`) and no longer silently drop following keys.
- Quoted `"|"` still parses as the string `"|"` (non-regression).
- The bundled `gbrain-recommended.yaml` parses with its full page_types (regression guard for the shipped pack).

```
bun test test/schema-pack-loader.test.ts   # 41 pass / 0 fail
bun node_modules/typescript/bin/tsc --noEmit  # 0 errors
```


harjothkhara:
Gentle nudge when you have a moment, @garrytan — this fix for #1750, along with #1800 (#1738) and #1803 (#1717), are all green and mergeable. No urgency; just flagging since they've been quiet for a few days. Happy to rebase if any have drifted from `main`.

harjothkhara:
Gentle follow-up, @garrytan — this (#1750) plus #1800 (#1738) and #1803 (#1717) have been green and mergeable for ~2 weeks now. No rush; flagging again in case they slipped off the radar. Happy to rebase if any have drifted from `main`.

time-attack:
Thanks for the thorough diagnosis of #1750 — the silent-truncation analysis was exactly right. Since you opened this, master went the other direction and taught the parser to actually handle block scalars (`parseBlockScalar` in loader.ts, wired into `parseMapping` for both `|` and `>`), so `gbrain-recommended`'s `description: |` now parses and its page_types load. Closing as already fixed; the fail-loud guard would now reject valid manifests.

PR #1800, opened 2026-06-03
fix(agent): consume `agent run` flags that appear after the prompt (#1738)

### LINKED ISSUE #1738: gbrain agent run: --detach / flags after the prompt get swallowed into the prompt string

harjothkhara:
## Problem

Fixes #1738.

`gbrain agent run` is documented as the canonical subagent entrypoint, but `parseRunFlags` (`src/commands/agent.ts`) stopped scanning at the first non-flag token and returned `args.slice(i)` as the prompt. Any flag placed **after** the prompt leaked into the prompt string instead of being parsed:

```
$ gbrain agent run "test prompt" --detach
$ gbrain jobs get <id>
  Data: {"prompt":"test prompt --detach"}   # --detach swallowed into the prompt, never applied
```

This silently corrupts the agent's instruction and makes `--detach` / `--follow` behave unreliably depending on argument order.

## Fix

`parseRunFlags` now collects non-flag tokens into the prompt and **keeps scanning**, so known flags are consumed wherever they appear in argv (before, after, or interleaved with the prompt). Behaviour preserved:

- `--` still terminates flag parsing — a prompt that legitimately contains flag-shaped words can be passed literally: `gbrain agent run -- explain the --verbose option`.
- Unknown flags still throw; value-flags still consume their value; prompt token order is preserved.

Help text updated to match the new (and now accurate) contract.

## Tests (`test/agent-cli.test.ts`)

Added to the existing `parseRunFlags` block:
- a flag after the prompt is consumed, not leaked (the issue's exact repro)
- value-flags after the prompt consume their value (`--model m`, `--max-turns 5`)
- flags on both sides of the prompt are all consumed
- `--` after a positional keeps flag-shaped tokens as literal prompt

All 8 pre-existing `parseRunFlags` cases still pass.

```
bun test test/agent-cli.test.ts                # 31 pass / 0 fail
bun test test/agent-runner.test.ts test/submit-agent.test.ts   # green
bun node_modules/typescript/bin/tsc --noEmit   # 0 errors
```

### DIFF
--- a/src/commands/agent.ts
+++ b/src/commands/agent.ts
@@ -74,8 +74,9 @@ SUBMITTING
     --follow                     Tail status until terminal (default on TTY)
     --detach                     Submit + print job id, exit immediately
 
-  Flags after \`run\` up to the first unrecognized token are parsed; the
-  remainder is the prompt. Use \`--\` to explicitly terminate flag parsing.
+  Flags may appear anywhere after \`run\` (before or after the prompt); the
+  non-flag tokens form the prompt. Use \`--\` to terminate flag parsing so
+  the remainder is treated as literal prompt text.
 
 VIEWING
   gbrain agent logs <job_id>
@@ -107,11 +108,24 @@ function parseRunFlags(args: string[]): { flags: RunFlags; rest: string[] } {
     follow: process.stdout.isTTY === true,
     detach: false,
   };
+  const rest: string[] = [];
   let i = 0;
   while (i < args.length) {
     const a = args[i];
-    if (a === '--') { i++; break; }
-    if (!isKnownFlag(a!)) break;
+    if (a === '--') {
+      // Explicit terminator: everything after `--` is literal prompt text,
+      // even if it looks like a flag.
+      rest.push(...args.slice(i + 1));
+      break;
+    }
+    if (!isKnownFlag(a!)) {
+      // Positional token → part of the prompt. Keep scanning so flags that
+      // appear AFTER the prompt (e.g. `agent run "x" --detach`) are still
+      // consumed instead of leaking into the prompt string (#1738).
+      rest.push(a!);
+      i++;
+      continue;
+    }
     switch (a) {
       case '--subagent-def': flags.subagentDef = args[++i]; i++; break;
       case '--model':        flags.model = args[++i]; i++; break;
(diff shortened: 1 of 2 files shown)


time-attack:
Thanks for chasing #1738 down to `parseRunFlags`. Master now consumes trailing boolean flags after the prompt (the `BOOLEAN_TAIL_FLAGS` pop, plus a literal `--` escape hatch), so the exact repro in the issue — `agent run "prompt" --detach` — no longer leaks into the prompt. Closing as already fixed; full interleaved value-flag parsing is a parser-surface change we're not taking on top of that.

PR #1803, opened 2026-06-03
fix(embed): label content_chunks.model with the model that produced the vector (#1717)

### LINKED ISSUE #1717: Embedding: `gbrain embed`/`import` mislabels content_chunks.model as the gateway default (v0.41.x)

harjothkhara:
## Problem

Fixes #1717.

`gbrain embed --all` / `gbrain embed <slug>` / `gbrain import` built the `ChunkInput[]` for `upsertChunks` **without a `model` field**. `upsertChunks` writes `chunk.model || DEFAULT_EMBEDDING_MODEL`, and its `ON CONFLICT` clause does `model = COALESCE(EXCLUDED.model, content_chunks.model)` where `EXCLUDED.model` is always non-null — so every embedded chunk's `model` column was stamped with the **gateway default** (e.g. `zeroentropyai:zembed-1`) regardless of the model that actually produced the vector.

The vectors themselves are correct (they go through `getEmbeddingModel()`); only the stored `model` label is wrong. This misleads `gbrain doctor`, mixed-model detection, and anyone reading the `model` column to learn which embedding space a chunk lives in (it can read `zeroentropyai:zembed-1` while the vector is actually OpenAI `text-embedding-3-large`).

## Fix

Thread the resolved embedding model (`getEmbeddingModel()` — the same value the batch was embedded with) into the `model` field at all three write sites: `embedPage`, `embedAll` → `embedOnePage`, and `embedAllStale`.

Labeling is **per-chunk**, which the `COALESCE` semantics require for correctness:
- chunks **(re)embedded this pass** → the resolved model;
- chunks **left untouched** (preserved in the same upsert) → keep their existing `model`, so a partially-stale / mixed-model page isn't relabeled wholesale.

Falls back to the chunk's existing model if the gateway is unconfigured (never mislabels).

## Test (`test/embed.serial.test.ts`)

A page with one stale chunk + one already-embedded chunk (with a *different* model):
- the re-embedded chunk is labeled with the configured model (was `DEFAULT_EMBEDDING_MODEL` before the fix);
- the untouched chunk keeps its original model (no wholesale relabel).

```
bun test test/embed.serial.test.ts          # 27 pass / 0 fail
bun test test/embed-stale.serial.test.ts test/operations-embedding-column.test.ts \
         test/embedding-signature-stale.test.ts test/embed-multimodal-batching.test.ts   # 32 pass
bun node_modules/typescript/bin/tsc --noEmit  # 0 errors
```

### DIFF
--- a/src/commands/embed.ts
+++ b/src/commands/embed.ts
@@ -303,6 +303,22 @@ export async function runEmbed(engine: BrainEngine, args: string[]): Promise<Emb
   }
 }
 
+/**
+ * Resolve the embedding model label (`provider:model`) to stamp onto
+ * `content_chunks.model`, so each chunk records the model that actually
+ * produced its vector instead of the gateway default (#1717). Returns
+ * undefined if the gateway is unconfigured; callers then fall back to the
+ * chunk's existing model rather than mislabeling it.
+ */
+async function resolveEmbeddingModelLabel(): Promise<string | undefined> {
+  try {
+    const { getEmbeddingModel } = await import('../core/ai/gateway.ts');
+    return getEmbeddingModel();
+  } catch {
+    return undefined;
+  }
+}
+
 async function embedPage(
   engine: BrainEngine,
   slug: string,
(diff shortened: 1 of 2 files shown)


time-attack:
Great root-cause writeup on #1717 — the COALESCE/EXCLUDED analysis was correct. Master has since fixed it at the engine layer instead of the call sites: `upsertChunks` now resolves the gateway's live embedding model and stamps that rather than the compile-time `DEFAULT_EMBEDDING_MODEL`, which covers `embed`, `import`, and every other caller in one place. Closing as already fixed.

PR #1811, opened 2026-06-03
fix(content-sanity): guard against non-string page titles

p3ob7o:
## What

`assessContentSanity` (`src/core/content-sanity.ts`) is typed to take a string `title`, but at runtime YAML hands it a **number** for numeric-named pages. The clearest trigger: a contact named by a bare phone number, whose page slug is all digits (e.g. `33384818832`). `opts.title.toLowerCase()` then throws:

```
```

Because the sanity check runs in the import path, that exception aborts the whole `gbrain sync` run, and the offending page never imports — it shows up as a recurring, un-clearable failure in `gbrain doctor` (`--retry-failed` just re-throws).

## Fix

Coerce `title` to a string once at the top of the function and reuse it for both the lowercase scan and the junk-pattern `.test()`:

```ts
const title = typeof opts.title === 'string' ? opts.title : String(opts.title ?? '');
```

One defensive line at the consumer covers every caller (import, lint, doctor, quarantine, sources) without touching the public type.

## Test

Adds a regression block to `test/content-sanity.test.ts` asserting that numeric and null/undefined titles don't throw. `bun test test/content-sanity.test.ts`: 79 pass, 0 fail.

Found on a live brain after upgrading to 0.42.x (the content-quality gate landed in 0.42.8.0).

### DIFF
--- a/src/core/content-sanity.ts
+++ b/src/core/content-sanity.ts
@@ -376,14 +376,20 @@ export function assessContentSanity(opts: {
   // doesn't repeat the lowercase per literal.
   const bodyHead = body.slice(0, SCAN_HEAD_BYTES);
   const bodyHeadLower = bodyHead.toLowerCase();
-  const titleLower = opts.title.toLowerCase();
+  // Defend against non-string titles. YAML coerces an all-digit / leading-"+"
+  // title (e.g. a contact named by a bare phone number, slug "33384818832")
+  // to a number, so `opts.title` can be a number at runtime despite the
+  // declared `string` type. Calling `.toLowerCase()` on it throws and aborts
+  // the entire sync run. Normalize once and reuse.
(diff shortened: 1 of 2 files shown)


sanchitmonga22:
Confirming an independent reproduction of #1658 on v0.42.23.0: when a page's frontmatter `title:` parses to a non-string (a bare numeric or date title), `assessContentSanity` throws `opts.title.toLowerCase is not a function` at content-sanity.ts ~L379, and the same unguarded value also reaches the `p.pattern.test(opts.title)` call ~L386. This blocks the dream/sanity pass with `cycle_freshness FAIL`. This PR's one-shot coercion plus the regression test look like the right fix (vs. guarding only the single call site). Is there a timeline for getting it merged? Happy to help if anything's blocking.

garrytan:

PR #1813, opened 2026-06-03
fix(code-graph): cross-file edge resolution + arrow-const defs + bash calls

sonlndv:
## What
Completes the TS/TSX call-graph work (commit `28e8cdf6`, on which this stacks): edges were named + source_id'd but stayed `resolved:false` because (a) resolution was **same-file only** and (b) resolved edges were never **promoted into `code_edges_chunk`** — the table the read path (`getCallersOf`/`getCalleesOf`) treats as resolved.

> Note: this branch is stacked on prior helmsman-build code-graph commits; the **only new commit here is `0e76bef1`**. Happy to rebase onto a clean base if preferred.

## Three coupled fixes (the new commit)
1. **Cross-file resolution** (`symbol-resolver.ts`): a source-wide bare-name definition index resolves an imported/cross-file callee to its unique definition anywhere in the source (not just same file). Resolved edges are PROMOTED into `code_edges_chunk` so reads report `resolved:true` with a real `to_chunk_id`. Keyed on the **bare** symbol name (`bareSymbolName`) to bridge the `::` (call side) vs `.` (def side) qualified-name schemes. 2+ defs → ambiguous (honest), 0 → unmatched.
2. **Arrow/const function defs** (`code.ts` `normalizeSymbolType`): `const Foo = () => {}` / `export const Bar = function(){}` now normalize to `symbol_type='function'` (was `'lexical declaration'`, which `code-def`'s DEF_TYPES dropped). Surfaces the dominant modern JS/TS/React definition form.
3. **bash call edges** (`edge-extractor.ts`): added `bash` to `CALL_CONFIG` (`command` nodes); relaxed the `import-file.ts` edge-persistence gate so call sites in unnamed/merged chunks (common in .js/.mjs) aren't silently dropped.

Plus: `deleteCodeEdgesForChunks` resets `edges_backfilled_at` so a reindex re-triggers resolution (pg + pglite parity); bumped `EDGE_EXTRACTOR_VERSION_TS`.

## Verified on a real RN/TSX repo
- `edges-backfill`: **528 resolved** (was 0).
- `code-callees DiaryScreen` → 76 callees with `resolved:true` + non-null `to_chunk_id` (cross-file: mealTimeLabel, fmtTime, dayOfMonth…).
- `code-def Button/HeroChip/Card` → 1 (was 0).
- bash extracts `[git, myfunc, echo]` (was `[]`).
- 286 existing chunker/edge/resolver tests pass + new pins for all three fixes.

## Risk
Engine-core (chunker + resolver run on every sync). typecheck clean; full unit suite green; non-code text sync/query verified unaffected.


sonlndv:
Superseded by #1814 — same fix on a clean branch that stacks only on #1723 (this one accidentally bundled 4 already-open PRs' commits + a private version stamp, and conflicted). Please review #1814 instead.

PR #1814, opened 2026-06-03
fix(code-graph): cross-file edge resolution + arrow-const defs + bash calls

sonlndv:
## What
Completes the TS/TSX call-graph work. **Stacks on #1723** (`0aba3d38` TS/TSX call edges) — the only NEW commit here is `398ea089`. Merge #1723 first, or merge this and #1723 is absorbed.

After #1723, edges were named + source_id'd but still `resolved:false` because (a) resolution was **same-file only** and (b) resolved edges were never **promoted into `code_edges_chunk`** (the table `getCallersOf`/`getCalleesOf` treats as resolved).

## Three coupled fixes
1. **Cross-file resolution** (`symbol-resolver.ts`): source-wide bare-name def index resolves imported/cross-file callees to their unique definition; resolved edges PROMOTED into `code_edges_chunk` → reads report `resolved:true` + real `to_chunk_id`. Bare-name key (`bareSymbolName`) bridges `::` (call) vs `.` (def) schemes. 2+ defs → ambiguous; 0 → unmatched.
2. **Arrow/const defs** (`code.ts normalizeSymbolType`): `const Foo = () => {}` / `export const Bar = function(){}` → `symbol_type='function'` (was `'lexical declaration'`, dropped by code-def DEF_TYPES).
3. **bash call edges** (`edge-extractor.ts`): bash added to CALL_CONFIG; relaxed the import-file edge-persistence gate so call sites in unnamed/merged .js/.mjs chunks aren't dropped.

Plus: `deleteCodeEdgesForChunks` resets `edges_backfilled_at` so reindex re-triggers resolution (pg + pglite); bumped `EDGE_EXTRACTOR_VERSION_TS`.

## Verified on a real RN/TSX repo
- `edges-backfill`: **528 resolved** (was 0).
- `code-callees DiaryScreen` → 76 with `resolved:true` + non-null `to_chunk_id` (cross-file: mealTimeLabel, fmtTime…).
- `code-def Button/HeroChip/Card` → 1 (was 0).
- bash → `[git, myfunc, echo]` (was `[]`).
- 286 existing tests pass + new pins for all three fixes; typecheck clean.

## Risk
Engine-core (chunker + resolver run on every sync). typecheck clean; full unit suite green; non-code text sync/query verified unaffected.


time-attack:
Thanks for the detailed verification numbers on a real repo — the cross-file resolution work is clearly careful. Closing this one: it's explicitly stacked on #1723, which was closed rather than merged, so it can't land as-is, and the remaining change is 190 lines bundling three separate concerns across chunker + both engines + import-file, i.e. engine-core paths that run on every sync. Code-graph edge extraction has also moved on master since (Kotlin call edges landed via #2574), so this would need a rebase and a split regardless.

PR #1816, opened 2026-06-03
fix(minions): route lock claim/renewLock through direct session pool

garrytan-agents:
## The wedge, root-caused

The Minion lock hot-path — `claim()` and `renewLock()` in `queue.ts` — ran every UPDATE through `engine.executeRaw()`, which is hardcoded to the read pool (`this.sql`). On Supabase that's the **transaction-mode pooler (port 6543)**, which recycles connections per-transaction.

A lock heartbeat is held open for minutes. So the pooler periodically reaps its socket mid-flight → bun's postgres client throws `CONNECTION_ENDED` → the lock looks expired → the worker force-evicts the job → the claim loop wedges (process alive, no errors, claiming nothing). This showed up as repeated silent worker wedges, especially under `enrich` fan-out load.

## The fix was already half-built

gbrain already ships a **dual-pool** design: `ConnectionManager.ddl()/bulk()` route to the direct **session-mode** pool (port 5432, `GBRAIN_DIRECT_DATABASE_URL`), which holds a connection for the worker's lifetime. The lock path just never used it.

Verified empirically: a single connection heartbeating every 2s survived **4/4 beats over 8s on 5432**, vs. `CONNECTION_ENDED` storms on 6543.

## What this PR does

- Adds `BrainEngine.executeRawDirect()` — same contract as `executeRaw`, but routes to the direct pool when dual-pool is active. No-op delegation on PGLite / non-Supabase / kill-switch (`GBRAIN_DISABLE_DIRECT_POOL`).
- Points `claim()` and `renewLock()` at it.
- Single-statement UPDATEs only — same idempotency guarantees, so the **Codex #1 double-claim guard** and the **renewLock no-inline-retry** contract are preserved.
- Transaction-scoped engine clones keep using their tx connection (never reroute a statement off an open transaction — that would break atomicity). The lock hot-path doesn't run inside `transaction()`, so it always reaches the direct pool there.

## Testing

- `test/minions.test.ts` — **169/169 pass** (claim/renewLock via real `PGLiteEngine` delegation path)
- `test/queue-lock-retry.test.ts` — pass; mock updated to `executeRawDirect`, plus a new guard that `claim` never falls back to `executeRaw`
- `tsc --noEmit` — clean
- Guards green: `check-worker-pool-atomicity`, `check-worker-lock-renewal-shape`, `check-no-double-retry`, `check-no-legacy-getconnection`

## Note

No new infra needed — `GBRAIN_DIRECT_DATABASE_URL` is already configured in our env. This just makes the lock path use the pool that was built for exactly this workload.

### DIFF
--- a/src/core/engine.ts
+++ b/src/core/engine.ts
@@ -1912,6 +1912,24 @@ export interface BrainEngine {
     opts?: { signal?: AbortSignal },
   ): Promise<T[]>;
 
+  /**
+   * Like `executeRaw`, but routes through the DIRECT (session-mode) pool when
+   * dual-pool is active (Supabase: port 5432), falling back to the read pool
+   * otherwise. Use this for the Minion lock hot-path (`claim`/`renewLock`):
+   * those statements heartbeat a lock over many seconds, and the
+   * transaction-mode pooler (port 6543) recycles connections per-transaction,
+   * which surfaces as `CONNECTION_ENDED` mid-heartbeat → orphaned locks →
+   * silent worker wedge. The direct session pool holds the connection open for
+   * the life of the worker, so heartbeats survive. Single-statement UPDATEs
+   * only — same idempotency contract as `executeRaw`. On PGLite (no pooler)
+   * this is identical to `executeRaw`.
+   */
+  executeRawDirect<T = Record<string, unknown>>(
+    sql: string,
+    params?: unknown[],
+    opts?: { signal?: AbortSignal },
+  ): Promise<T[]>;
+
   // ============================================================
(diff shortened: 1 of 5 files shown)


garrytan:
Superseded: the full change set has been pulled into the `garrytan/minion-locks-session-pool` worktree (Conductor workspace `cape-town-v6`), where it will ship as part of that branch's PR. Closing this one to avoid a duplicate. All five files apply cleanly and the verification (typecheck, minions.test.ts, queue-lock-retry, four lock/pool guards) reproduces green there.

PR #1821, opened 2026-06-03
feat(put_page): rich-page overwrite guard for agent callers (force=true override)

georgell-ceo:
## What

Adds a tool-boundary guard to `put_page`: remote/agent callers can't overwrite a page whose existing `compiled_truth` is substantial (>= 100 chars) unless they pass `force: true`. Trusted local CLI callers (`ctx.remote === false`) are unaffected and the guard does not even read the existing page for them.

## Why

Agents follow a "helpful upsert" instinct: they read a page, decide they can improve it, and call `put_page` with a rewritten version, silently discarding the prior content. Prompt-level guidance ("read the page first; don't overwrite a rich page") does not reliably hold across models. Enforcing at the tool boundary turns an accidental clobber into an explicit, opt-in act.

The thrown error (`rich_page_overwrite_blocked`) names the alternatives so a blocked agent has a clear path forward:

> Use add_timeline_entry for new dated facts, add_link for new relationships, or pass force=true to deliberately overwrite.

## Details

- New `force` boolean param on `put_page` — propagates to the MCP tool schema automatically via the contract-first op definitions.
- New `ErrorCode` member `rich_page_overwrite_blocked`.
- New exported `RICH_PAGE_MIN_CHARS = 100` threshold.
- Guard runs **before** the dry-run short-circuit, so a dry-run preview surfaces the same block (matches the existing subagent-namespace check precedent).
- **Fail-closed:** anything not strictly `ctx.remote === false` is treated as remote (matches the CV6 / v0.26.9 F7b trust posture already used in this handler for provenance stamping).
- Multi-source aware: reads the existing page with `ctx.sourceId`.

## Tests

- New hermetic regression suite `test/operations-put-page-overwrite-guard.test.ts` (6 cases): blocks at the threshold boundary, blocks on dry-run preview, allows thin pages, allows when no page exists, allows with `force=true`, and allows local callers without even reading the page.
- `bun run typecheck` clean.
- parity / mcp-tool-defs / public-exports / operations-* suites green (157 tests).

## Notes

- VERSION / CHANGELOG intentionally left untouched for the maintainer's release process.

### DIFF
--- a/src/core/operations.ts
+++ b/src/core/operations.ts
@@ -69,6 +69,7 @@ export type ErrorCode =
   | 'rate_limited'      // v0.31: gateway rate-limit upstream
   | 'extraction_failed' // v0.31: facts extractor failed (refusal, parse, abort)
   | 'fact_not_found'    // v0.31: forget_fact / recall on unknown id
+  | 'rich_page_overwrite_blocked' // put_page guard: agent overwrite of a substantial page without force=true
   // eslint-disable-next-line @typescript-eslint/ban-types
   | (string & {});      // OPEN union for forward-compat (eE7 / D13)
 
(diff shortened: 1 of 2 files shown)


garrytan:

PR #1823, opened 2026-06-03
Fix propose_takes dry-run writes

uncfreak1255-code:
## Summary
- prevent `cycle propose_takes --dry-run` from inserting take proposals, receipts, or rollups
- pass `dryRun` through cycle calibration phases so the pipeline honors dry-run mode
- add a regression test that proves dry-run extraction does not persist proposals

## Verification
- `bun test test/propose-takes.test.ts`
- `bun run typecheck`
- local runtime smoke: dry-run proposal count stayed `0 -> 0` after cleanup
- global `gbrain` relinked to the patched local checkout and reports `0.42.8.0`

## Notes
- This is a narrow behavioral fix only. Dream-lint side effects from the runtime run were separated out and not included in this branch.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,8 @@ test/fixtures/pglite-snapshot.version
 
 # Private brain reports — never check these in (per CLAUDE.md privacy rule)
 reports/network-intelligence/
+
+# Auto-managed by gbrain (db_only directories)
+media/x/
+media/articles/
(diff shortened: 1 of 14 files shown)


uncfreak1255-code:
Closing from the contributor side. This was opened for a local runtime fix path, but we do not need an upstream PR for this workflow.

PR #1826, opened 2026-06-03
feat(gateway): Amazon Bedrock provider — embedding, reranker, chat

naterchrdsn:
## What

Adds `bedrock` as a native provider, filling the only major enterprise
hosting gap in the recipe registry. AWS shops running on Bedrock can now
use gbrain without a LiteLLM proxy or a second API key vendor.

## Provider surface

**Embedding** — Titan Embed Text v2 (1536d, Matryoshka 256/512/1024/1536),
Titan v1, Cohere Embed English v3, Cohere Embed Multilingual v3.

**Reranker** — `amazon.rerank-v1:0` and `cohere.rerank-v3-5:0`. Uses
`@ai-sdk/amazon-bedrock`'s `RerankingModelV4.doRerank()` instead of the
ZeroEntropy/llama.cpp wire shape — Bedrock's rerank API is structurally
different (queries array, not a flat documents list), so the native path
branches early in `rerank()` before the raw HTTP fetch.

**Chat** — Claude on Bedrock (cross-region inference prefixes included),
Nova Pro/Lite/Micro.

## Auth

AWS_REGION is required. Key-based auth (ACCESS_KEY_ID + SECRET +
optional SESSION_TOKEN) and IAM instance profile/role are both supported.
The pattern matches how the rest of the stack handles optional key env vars
— no keys present means the SDK falls back to the credential provider chain.

## Implementation notes

- `native-bedrock` added to the `Implementation` union in `types.ts`.
  Three switch sites in `gateway.ts` get the new case (instantiateEmbedding,
  instantiateExpansion, instantiateChat). Credential resolution extracted into
  `buildBedrockClientConfig(cfg)` to avoid repeating the optional-key spread
  across all four call sites (the fourth being the rerank branch).
- `@ai-sdk/amazon-bedrock` is a static import, same as the existing
  native-openai / native-google providers.
- `supports_subagent_loop: false`. The gateway-native tool loop
  (`agent.use_gateway_loop = true`) already supports any recipe with
  `supports_tools: true`, so Bedrock Claude works for subagents today with
  that flag. The flag is false by default because crash-replay `tool_call_id`
  stability across the Bedrock converse API hasn't been verified the way
  native-anthropic has.

## Not in this PR

- No Titan Rerank — not available on Bedrock as of this writing.
- No multimodal embedding — Titan v2 is text-only.
- Subagent crash-replay verification for `supports_subagent_loop: true` — follow-up.

## Verification

```
bun run typecheck
bun test test/ai/gateway-bedrock.test.ts   # 11 pass
bun test test/ai/gateway.test.ts           # 36 pass, no regressions
bash scripts/check-gateway-routed-no-direct-anthropic.sh
bash scripts/check-privacy.sh
bash scripts/check-key-files-current-state.sh
npx fallow audit --changed-since master    # new clone groups: 0
```

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -5,6 +5,7 @@
     "": {
       "name": "gbrain",
       "dependencies": {
+        "@ai-sdk/amazon-bedrock": "^4.0.112",
         "@ai-sdk/anthropic": "^3.0.71",
         "@ai-sdk/google": "^3.0.64",
         "@ai-sdk/openai": "^3.0.53",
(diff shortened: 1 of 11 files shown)


time-attack:
This is well-built and clearly documented — thank you for the effort. We're not taking new cloud-vendor provider integrations, though: each one adds a permanent third-party runtime dependency (`@ai-sdk/amazon-bedrock`), new branches in the gateway switch sites and the rerank path, and a provider surface we then have to keep priced and tested forever. Closing on that policy rather than on the quality of the work.

PR #1832, opened 2026-06-04
fix: litellm/user_provided_models embedding validation (two bugs)

sternryan:
## Summary

Two bugs in the `litellm` (and `llama-server`) embedding-validation path block anyone from using a litellm-proxied embedding model at a custom dimension. Both bugs live in the same under-tested code path — recipes that declare `user_provided_models: true` (no static model list, user must supply model and dimensions at init time). Nobody uses this path via the openai/voyage/zeroentropyai first-party providers, so both bugs were invisible until a litellm user hit them.

### Bug 1 — `src/core/embedding-dim-check.ts`, `isCustomDimValidForProvider()`

**Problem:** `gbrain init --embedding-model litellm:embeddings --embedding-dimensions 1024` fails with a dim-validation error, making it impossible to initialize a brain on a litellm-proxied model.

The `isCustomDimValidForProvider()` function has no path for `user_provided_models` recipes. These recipes declare `default_dims: 0` as a sentinel meaning "the user must supply dimensions at init time." When the user does supply `--embedding-dimensions N`, the function should accept it immediately — there is no server-side allow-list gbrain can know about for an arbitrary litellm backend. Instead, it falls through to the Tier 3 Matryoshka rejection ("this model only emits its default dim"), and returns `valid: false` for every custom dim.

**Fix:** Add an early-return guard before Tier 1. When `recipe.touchpoints.embedding.user_provided_models === true`, the user-supplied dimension is the complete contract; return `{ valid: true, error: '' }` and skip all Matryoshka tiers.

### Bug 2 — `src/core/ai/gateway.ts`, `diagnoseEmbedding()`

**Problem:** `diagnoseEmbedding()` returns `user_provided_model_unset` whenever the recipe's static `tp.models` array is empty — which is always the case for litellm by design. The check does not look at whether the user actually supplied a model name after the provider prefix (e.g. `litellm:embeddings`). So `isAvailable('embedding')` returns `false` and the embed path is blocked even when the user has correctly configured `embedding_model: 'litellm:embeddings'`.

**Fix:** Extract the user-model portion of `modelStr` (the substring after the first `:`). Only reject with `user_provided_model_unset` when that portion is absent — preserving the guard for a genuinely unconfigured provider while letting fully-specified strings through.

## Changes

- `src/core/embedding-dim-check.ts`: early-return in `isCustomDimValidForProvider()` for `user_provided_models` recipes
- `src/core/ai/gateway.ts`: check for non-empty user model name in `diagnoseEmbedding()` before returning `user_provided_model_unset`
- `test/embedding-dim-check.test.ts`: new `describe` block — 5 tests for Bug 1, including regression pins that standard providers (OpenAI, ZeroEntropy) still reject invalid dims
- `test/ai/gateway.test.ts`: new `describe` block — 4 tests for Bug 2, covering the happy path, compound model ids, and a no-regression pin for unknown providers

## Testing

```
bun test test/embedding-dim-check.test.ts    # 31 pass, 0 fail (new tests included)
bun test test/ai/gateway.test.ts             # 40 pass, 0 fail (new tests included)
```

All new tests fail against the unfixed code and pass after the fix (verified by reverting the source changes, running `bun test` on the two files, re-applying, and confirming green).

## Notes

This is a correctness fix for an under-tested path. The litellm and llama-server providers have been present since v0.32 but have no test coverage for the init/embed flow at a custom dimension. The fix is minimal and does not change behavior for any existing provider.

### DIFF
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -704,11 +704,17 @@ export function diagnoseEmbedding(modelOverride?: string): EmbeddingDiagnosis {
     };
   }
 
-  // Openai-compat recipes with empty models list require a user-provided model.
+  // Openai-compat recipes with empty models list require a user-provided model
+  // (e.g. `litellm:embeddings`). The user supplies the model as the portion after
+  // the first colon in the embedding_model string. Only reject when that portion
+  // is absent — not merely because the recipe's static models[] is empty (it is
+  // always empty for user_provided_models recipes by design).
   const isUserProvided = (tp as any).user_provided_models === true;
+  const userModelName = modelStr.includes(':') ? modelStr.slice(modelStr.indexOf(':') + 1).trim() : '';
   if (
     Array.isArray(tp.models) &&
     tp.models.length === 0 &&
+    !userModelName &&
     (recipe.id === 'litellm' || isUserProvided)
   ) {
     return {
--- a/src/core/embedding-dim-check.ts
+++ b/src/core/embedding-dim-check.ts
@@ -409,6 +409,15 @@ function isCustomDimValidForProvider(
   requestedDims: number,
   dimsOptions: number[] | undefined,
 ): CustomDimCheck {
+  // Tier 0: user_provided_models recipes (e.g. litellm, llama-server) have no
+  // static model list and declare default_dims=0 as a sentinel meaning "the user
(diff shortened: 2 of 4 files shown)


garrytan:

PR #1833, opened 2026-06-04
Fix PGLite stale lock recovery for stdio serve

zach22-1999:
## Summary
- stop `startMcpServer()` from owning stdio/signal shutdown so `runServe()` is the single lifecycle owner
- preserve dead-PID/corrupt-lock reclaim while refusing to reap live PGLite holders
- add regression coverage for stale-lock self-heal, single-owner stdio lifecycle, and slower serial disconnect tests

## Root cause
`gbrain serve` already installs stdio/signal/parent-watchdog shutdown in `runServe()`, but `startMcpServer()` also installed its own stdin, transport, and signal shutdown hooks. That created a double-owner cleanup path around `engine.disconnect()` / `process.exit()`. In parallel, `pglite-lock.ts` still contained the older "reap long-held live lock" behavior, which is unsafe for embedded Postgres and not needed once dead-PID stale locks are reclaimed reliably on the next connect.

## Fix
- remove the duplicate stdio/signal shutdown hooks from `src/mcp/server.ts`
- keep stale-lock recovery in `acquireLock()` limited to dead PIDs and corrupt lockfiles
- add a serial CLI regression that seeds a dead-PID stale lock and proves `gbrain stats` reaps it and exits 0
- add a lightweight regression test that prevents future reintroduction of a second stdio lifecycle owner
- extend PGLite lock coverage and raise timeouts on the heavy disconnect serial tests so the intended invariants stay green under cold migration startup

## Validation
- `bun test test/pglite-lock.test.ts`
- `bun test test/mcp-stdio-owner-regression.test.ts`
- `bun test test/serve-stdio-lifecycle.test.ts`
- `bun test test/pglite-engine-disconnect.serial.test.ts`
- `bun test test/e2e/pglite-cli-exit.serial.test.ts`

### DIFF
--- a/src/core/pglite-lock.ts
+++ b/src/core/pglite-lock.ts
@@ -14,12 +14,11 @@
  *   try { ... } finally { await releaseLock(lock); }
  */
 
-import { mkdirSync, existsSync, readFileSync, writeFileSync, rmSync, statSync } from 'fs';
+import { mkdirSync, existsSync, readFileSync, writeFileSync, rmSync } from 'fs';
 import { join } from 'path';
 
 const LOCK_DIR_NAME = '.gbrain-lock';
 const LOCK_FILE = 'lock';
-const STALE_THRESHOLD_MS = 5 * 60 * 1000; // 5 minutes — embed jobs can be long
 
 export interface LockHandle {
   lockDir: string;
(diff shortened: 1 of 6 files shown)


garrytan:

PR #1837, opened 2026-06-04
fix: make dream extract_facts idempotent so fence rows don't duplicate each cycle

### LINKED ISSUE #1781: dream extract_facts non-idempotent: facts + fence rows duplicate each cycle (Postgres engine)
### Summary

On a Postgres-engine brain, the `extract_facts` phase of `gbrain dream` appears **non-idempotent**: re-running the cycle on unchanged content re-inserts the same facts as new rows (and new `## Facts` fence rows), so duplicates accumulate roughly linearly with the number of cycle runs. Over weeks of a nightly `dream`, the `facts` table grows to many multiples of the distinct fact set, and individual entity-page fences end up containing dozens of byte-identical rows.

### Environment

- gbrain `v0.42.8.0` (also observed on the state inherited from `v0.37.2.0`)
- Engine: Postgres (`postgres-engine`), pgvector, `text-embedding-3-large` @ 1536d
- Multi-source brain (`default` + one federated source), nightly `gbrain dream` via cron
- Schema version 112

### Observed

```
SELECT count(*) total,
       count(DISTINCT (entity_slug, fact)) distinct_facts,
       count(*) - count(DISTINCT (entity_slug, fact)) dupes
FROM facts;
-- total=4360  distinct_facts=929  dupes=3431
```

A single fact's text repeats as many identical **fenced** rows on its entity page (`row_num IS NOT NULL`), e.g. one company page's `## Facts` fence held ~85 byte-identical rows of the same claim. ~3142 of the 3431 duplicate rows are fenced (in the markdown system-of-record), not just stray DB rows.

The duplicate count tracks the number of `dream` runs (≈ one extra copy of each fact per cycle), which is consistent with `extract_facts` re-extracting + re-appending facts that already exist rather than treating them as already-known.

`gbrain sync` alone (no `dream`) does **not** add duplicates — only cycles that run `extract_facts` do.

### Expected

Re-running `dream` on unchanged content should be idempotent: an already-extracted fact (already present in the DB and/or the page's `## Facts` fence) should not be re-inserted / re-appended. The fence's row-level dedup (the `(claim, source)` key used by `v0.32.2`'s `upsertFactRow`) does not appear to be applied on the `extract_facts` write path.

### Impact

- `facts` table and entity-page fences grow unbounded over nightly cycles.
- Search returns many byte-identical fact chunks for a single query (diversity loss, wasted context tokens), and every duplicate is embedded (cost).
- There is no `gbrain facts dedup` command to recover, and a DB-only dedup would desync the markdown fences (which contain the duplicates), risking re-wedging the `v0.32.2` verify.

### Questions

1. Is `extract_facts` intended to dedup against existing facts (by `content_hash` / `(claim, source)`)? If so, what conditions cause it to miss exact-text repeats?
2. Is there a supported way to dedup an existing brain's facts + fences in lockstep?

Happy to provide more detail or a minimal repro brain.

mvanhorn:
## Summary

Re-running a `gbrain dream` cycle on unchanged content no longer accumulates duplicate facts. The `extract_facts` phase in `src/core/cycle/extract-facts.ts` now de-duplicates extracted facts against existing fence rows by the canonical `(claim, source)` content key before appending or calling `engine.insertFacts(...)`, so a no-change cycle is a no-op.

## Why this matters

Issue #1781 reported that on a Postgres-engine brain the `extract_facts` phase was non-idempotent: each cycle re-inserted the same facts and re-appended byte-identical `## Facts` fence rows, growing roughly linearly with the run count (the reporter saw total=4360 / distinct=929 / dupes=3431, with one entity page carrying ~85 identical fenced rows of the same claim). `gbrain sync` alone did not duplicate, and `consolidate` did not recover the duplicates. The row-level dedup key `(claim, source)` used by `upsertFactRow` was not applied on the `extract_facts` write path; `dedupeFactsByContentKey` and `factContentKey` in `extract-facts.ts` now apply it, and the existing empty-fence guard and legacy NULL-source pre-check are preserved.

## Testing

Covered by the new cases in `test/extract-facts-phase.test.ts`: extracting twice on an unchanged page inserts 0 the second time, same-claim-different-source is not treated as a duplicate, a legitimately new fact is inserted exactly once, and the empty-fence / legacy-row guards still short-circuit. Full suite runs in CI.

Fixes #1781

### DIFF
--- a/src/core/cycle/extract-facts.ts
+++ b/src/core/cycle/extract-facts.ts
@@ -11,15 +11,16 @@
  *   1. Reads the markdown body (DB-side fetch via engine.getPage).
  *   2. Parses the `## Facts` fence with parseFactsFence.
  *   3. Maps ParsedFact → FenceExtractedFact via extractFactsFromFenceText.
- *   4. Wipes the page's DB index via deleteFactsForPage.
- *   5. Re-inserts via engine.insertFacts batch.
+ *   4. De-dupes rows by canonical (claim, source) content key.
+ *   5. Reconciles the page-scoped DB index: no-op when already in sync,
+ *      insert only missing keys when possible, or wipe/reinsert when stale
+ *      DB rows need cleanup.
  *
- * After the phase, the DB index for every affected page byte-matches
- * the fence (modulo embeddings + runtime-derived fields). Pages with
- * no fence go through delete-then-empty-insert — DB rows for that
- * page coordinate are wiped; legacy NULL-source_markdown_slug rows
- * survive because deleteFactsForPage targets source_markdown_slug =
- * slug only.
+ * After the phase, the DB index for every affected page matches the
+ * fence's canonical (claim, source) row set (modulo embeddings +
+ * runtime-derived fields). Pages with no fence wipe DB rows for that
+ * page coordinate only; legacy NULL-source_markdown_slug rows survive
+ * because deleteFactsForPage targets source_markdown_slug = slug only.
  *
  * Empty-fence guard (Codex R2-#7): the phase refuses to do its
  * destructive reconciliation pass when legacy rows (row_num IS NULL,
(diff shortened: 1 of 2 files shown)


time-attack:
Superseded with credit by #2932 (just merged): your idempotency fix — dedupe by (claim, source) with reconcile instead of wipe-and-reinsert — ported onto current master preserving the #1972 abort-signal threading and #1928 cli-prefix protections, authorship on the commit via Co-authored-by. Fixes #1781. Thank you!

PR #1845, opened 2026-06-04
fix(doctor): resolve Windows storage paths under WSL

lost9999:
## Summary
- Add a small storage-path resolver for persisted Windows drive paths when doctor runs on non-Windows hosts such as WSL.
- Use that resolver in the `image_assets` doctor check instead of directly statting the persisted path.
- Add candidate-generation unit coverage for Windows drive paths, Windows runtime behavior, and native paths.

## Why
A brain can index images from a Windows host as paths like `D:/vault/image.jpg`. When `gbrain doctor` later runs under WSL, Node cannot stat that literal path even though the file exists at `/mnt/d/vault/image.jpg`. The previous `image_assets` check therefore reported all indexed images as missing and suggested `gbrain sync --skip-failed`, even though no asset was actually missing.

## Validation
- Local function smoke check:
  - `storagePathCandidates("D:/vault/image.jpg", "linux")` -> `["D:/vault/image.jpg", "/mnt/d/vault/image.jpg"]`
- Local live doctor against a WSL/Windows vault:
  - `image_assets`: `ok`, `248 image(s) all present on disk`
- `git diff --check` passed.

Note: this local WSL machine has a Bun 1.3.14 runner crash unrelated to this patch (`panic(main thread): Segmentation fault at address 0x0`) when invoking `bun test`; the included test is pure and should run under CI.

### DIFF
--- a/src/commands/doctor.ts
+++ b/src/commands/doctor.ts
@@ -41,6 +41,7 @@ import { lagFromContentMs } from '../core/source-health.ts';
 import { CHUNKER_VERSION } from '../core/chunkers/code.ts';
 import { LINK_EXTRACTOR_VERSION_TS } from '../core/link-extraction.ts';
 import { isUndefinedColumnError } from '../core/utils.ts';
+import { resolveExistingStoragePath } from '../core/path-resolve.ts';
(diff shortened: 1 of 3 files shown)


garrytan:

PR #1848, opened 2026-06-04
fix: updated personal brain tutorial to new Supabase UI

FilipHarald:
Thanks for this awesome project @garrytan .

I'm going through the tutorial now and it seems that some parts in tutorial are not up to date.

For me I got the connection string by the Navigation bar:
screenshot-2026-06-04_13-14-34

And there are now three options. I did not try with the session pooler, and I don't use Render atm, but it might work as well for the IPv4-problem. Anyway, transaction pooler seems to be what your tutorial was proposing before.
screenshot-2026-06-04_13-12-10


FilipHarald:
#1848 fixes this.

Issue #1779, opened 2026-06-02
Make the reasoning/expansion LLM configurable via provider recipes (parity with embedding providers)


danieltartaro:
Summary

gbrain already lets me choose my embedding provider from 16 recipes — I run llama-server:<model> against a local llama.cpp endpoint and it works great. But the reasoning LLM (multi-query expansion, extract-facts, the dream cycle, the Minions/jobs work worker) appears hardcoded to Anthropic Claude — ANTHROPIC_API_KEY is required and config show exposes no chat/reasoning model key. This asks for the reasoning tier to be configurable the same way embeddings already are.

Use case

The reasoning calls split cleanly into two latency classes:

Hot / interactive (query expansion when a human is waiting) — wants a fast, high-quality hosted model. Anthropic is a great default here.
Latency-tolerant batch (dream, extract-facts, overnight enrichment via the worker) — nobody is waiting. For these I want to run a local or cheaper model (e.g. a quantized OSS model behind an OpenAI-compatible endpoint like llama.cpp llama-server / llama-swap) for cost control and data locality, while keeping Anthropic for the hot path.

Right now the only way to get any of gbrain's cognition off Anthropic is to not use it — there's no "parallel"/secondary reasoning model.

Proposal

Mirror the embedding-provider design on the reasoning side:

Add a reasoning_model config key (optionally a separate expansion_model) resolved through the same recipe system that already powers embedding_model (llama-server:, openai:, litellm:, ollama:, …). Default = the current anthropic:claude-haiku-* so existing behavior is unchanged.
Honor an OpenAI-compatible base-URL override for the chat endpoint, exactly as embeddings already do (the LLAMA_SERVER_BASE_URL pattern).
Route expansion / extract-facts / dream / the Minions worker through the configured reasoning model instead of the hardcoded client.

Why this is low-risk

Default unchanged — Anthropic stays the out-of-box reasoning provider; this is purely additive and opt-in.
Same pattern you already shipped — the recipe + base-URL machinery exists for embeddings; this reuses it for chat/completions rather than inventing anything.
Unblocks the local-batch + cost + data-locality story without touching the interactive quality bar.

Precedent
You already did exactly this for embeddings — 16 provider recipes including local llama-server and ollama. This is the same treatment for the reasoning tier.
Happy to open a PR if you're open to the shape above.

time-attack:
[...]

Issue #1307, opened 2026-05-22
brainstorm/lsd cross-generation hardcoded to anthropic:claude-sonnet-4-6 — unrunnable without ANTHROPIC_API_KEY despite chat_model config


rvdlaar:
## Problem

On `0.39.0.0`, `gbrain brainstorm` and `gbrain lsd` are unrunnable on any stack that lacks `ANTHROPIC_API_KEY` — even with a valid `chat_model` set in `~/.gbrain/config.json` and even with `--judge-model` passed on the CLI.

## Repro

```bash
# Config: local llama-server for embeds, openrouter for chat
$ cat ~/.gbrain/config.json
{
  "chat_model": "openrouter:openai/gpt-oss-120b",
  "embedding_model": "llama-server:embeddinggemma",
  ...
}
$ env | grep -E '^(ANTHROPIC|OPENROUTER)_API_KEY'
OPENROUTER_API_KEY=sk-or-...
# ANTHROPIC_API_KEY: absent

$ gbrain brainstorm "test question" --limit 4
[brainstorm] WARN: cross [a/b] × [c/d] failed: [chat(openrouter:claude-sonnet-4-6)] claude-sonnet-4-6 is not a valid model ID
... (all 16 crosses fail identically)
brainstorm: no ideas generated across 4 crosses.
```

## Root cause — three compounding issues

1. **Cross-generation hardcode** (`src/core/brainstorm/orchestrator.ts:501`):
   ```ts
   const modelStr = opts.modelOverride ?? 'anthropic:claude-sonnet-4-6';
   ```
   `opts.modelOverride` is never set from the CLI — the `--judge-model` flag only feeds `opts.judgeModel` which is then forwarded ONLY into the judge phase (line 810). The cross-generation phase (line 686: `let crossModel = modelStr`) sees the hardcode unchanged.

2. **`chat_model` config is bypassed by the hardcode.** The orchestrator's `modelStr` falls back to a string literal rather than to the gateway-configured chat model. So setting `chat_model: openrouter:openai/gpt-oss-120b` in config does nothing for brainstorm.

3. **Buggy `anthropic:` → `openrouter:` provider rewrite.** When `ANTHROPIC_API_KEY` is missing but `OPENROUTER_API_KEY` is present, the gateway rewrites the provider prefix but keeps the bare model id, producing `openrouter:claude-sonnet-4-6`. OpenRouter's catalog uses `anthropic/claude-sonnet-4.6` — the bare id is rejected.

[...]

PR #1854, opened 2026-06-04
fix: 5 Ollama compatibility fixes for local-first usage

starm2010:
## Problem

Several features break when using Ollama or other local inference providers instead of Anthropic:

1. **`--max-cost` hard-fails for brainstorm/lsd**: `lookupPricing` returns `null` for local chat providers (Ollama, llama-server), causing budget-gated operations to fail even though the user explicitly configured a free local provider.

2. **Numeric idea IDs crash brainstorm judges**: Some models (e.g. Kimi K2 via Ollama) return numeric `id` fields in idea JSON instead of the expected string format like `"Idea 01"`, causing validation to silently reject all ideas.

3. **False `NO_ANTHROPIC_API_KEY` warning**: The warning fires even when the user has Ollama or OpenRouter configured — it's only relevant for the Anthropic provider.

4. **Unhelpful "no LLM available" message**: The error message only mentions `anthropic_api_key`, not the other config options (`chat_model`, `expansion_model`) that are the correct way to configure non-Anthropic providers.

5. **Ollama recipe missing chat/expansion touchpoints**: The Ollama recipe only defines embedding touchpoints. Without chat and expansion, the recipe system cannot route reasoning calls to Ollama, and `supports_tools` is not declared.

## Solution

1. Add `FREE_LOCAL_CHAT_PROVIDERS` set (parallel to existing `FREE_LOCAL_EMBED_PROVIDERS`) — returns zero pricing for local chat providers, fixing `--max-cost` budget checks.

2. Coerce numeric `id` to string format (`5` → `"Idea 05"`) in `validateIdeaShape` before the string type check.

3. Scope `NO_ANTHROPIC_API_KEY` warning to only fire when `providerId === 'anthropic'`.

4. Update "no LLM available" message to mention `chat_model`, `expansion_model`, or `anthropic_api_key`.

5. Add `chat` and `expansion` touchpoints to the Ollama recipe with generic models (`llama3.3`, `qwen3`, `deepseek-r1`), `supports_tools: true`, and zero cost.

## Testing

- TypeScript compiles cleanly (`tsc --noEmit` passes)
- No debug/temporary code included
- Backward compatible — all changes are additive; existing defaults unchanged

## Related Issues

Related to #1307 (hardcoded Anthropic model in brainstorm)
Related to #1779 (configurable reasoning/expansion LLM via recipes)

### DIFF
--- a/src/core/ai/recipes/ollama.ts
+++ b/src/core/ai/recipes/ollama.ts
@@ -21,6 +21,20 @@ export const ollama: Recipe = {
       // OLLAMA_NUM_PARALLEL config; no static cap to declare. v0.32 (#779).
       no_batch_cap: true,
     },
+    chat: {
+      models: ['llama3.3', 'qwen3', 'deepseek-r1'],
+      supports_tools: true,
+      supports_subagent_loop: false,
+      supports_prompt_cache: false,
+      cost_per_1m_input_usd: 0,
+      cost_per_1m_output_usd: 0,
+      price_last_verified: '2026-06-02',
+    },
+    expansion: {
+      models: ['llama3.3', 'qwen3', 'deepseek-r1'],
+      cost_per_1m_tokens_usd: 0,
+      price_last_verified: '2026-06-02',
+    },
   },
   setup_hint: 'Install Ollama from https://ollama.ai, then `ollama pull nomic-embed-text` and `ollama serve`.',
 };
--- a/src/core/brainstorm/judges.ts
+++ b/src/core/brainstorm/judges.ts
@@ -361,6 +361,11 @@ function isAxisScoreInRange(n: unknown): n is number {
 function validateIdeaShape(raw: unknown): { id: string; scores: JudgeAxisScores; note: string } | null {
   if (typeof raw !== 'object' || raw === null) return null;
   const r = raw as Record<string, unknown>;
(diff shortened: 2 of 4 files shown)


starm2010:
Friendly ping — this PR has been open for 6 weeks with no review. The underlying issues are still live on master:

- `--max-cost` crashes with local providers (Ollama/llama-server return null from `lookupPricing`)
- Numeric idea IDs from models like Kimi K2 crash brainstorm judges
- `brainstorm`/`lsd` hardcoded to `anthropic:claude-sonnet-4-6` (issue #1307 still OPEN)

Both PRs (#1854 and #1855) are mergeable with no conflicts. Happy to make any adjustments if there's feedback. Thanks for the great work on gbrain!

time-attack:
Thanks for the local-first testing — real usage reports like this are valuable. Closing as filed, though: five unrelated changes in one PR with no tests, and two of them are already stale (master's `think` path now distinguishes MODEL_NOT_USABLE from NO_ANTHROPIC_API_KEY via a model probe, with different message text). The Ollama recipe chat/expansion touchpoints are provider surface we're not adding, and the numeric-id coercion invents an ordering rather than fixing the prompt contract. Item 1 alone — `FREE_LOCAL_CHAT_PROVIDERS` so `--max-cost` doesn't hard-fail on local chat providers — is a genuine bug that's still present on master; a small standalone PR with one test for just that would be considered on its merits.

PR #1858, opened 2026-06-04
fix(takes): honor facts.extraction_model in takes extract --from-pages

### LINKED ISSUE #1857: takes extract --from-pages hardcodes anthropic:claude-haiku-4-5, silently returns 0 claims on OpenAI-only brains
### Summary
`gbrain takes extract --from-pages` returns `0 claim(s) from N page(s)` (no error) on a brain configured for OpenAI that has no `ANTHROPIC_API_KEY`, even though the chat gateway is healthy and `facts.extraction_model` is `openai:gpt-4.1-mini`.

### Root cause
`src/core/extract-takes-from-pages.ts` hardcodes the model on the per-page classifier call:

```ts
// src/core/extract-takes-from-pages.ts:177
response = await chat({
  model: opts.model ?? 'anthropic:claude-haiku-4-5',
  ...
});
```

But the documented contract for that param says it should default to the configured model:

```ts
// src/core/extract-takes-from-pages.ts:51
/** Model override; defaults to facts.extraction_model. */
model?: string;
```

Neither caller sets `opts.model`:
- `src/commands/takes.ts` `cmdExtract` (the `takes extract --from-pages` CLI handler)
- `src/commands/jobs.ts` (the `extract-takes-from-pages` Minion handler)

So every invocation resolves to the hardcoded Anthropic id regardless of `facts.extraction_model`.

### Why it fails silently
1. The top-level `isAvailable('chat')` guard checks the gateway's configured chat model (OpenAI here), so it passes.
2. Per page, `chat({ model: 'anthropic:claude-haiku-4-5', ... })` is attempted against Anthropic, which has no key configured.
3. The per-page `try/catch` swallows the failure and `continue`s → 0 claims for that page.
4. The function returns `{ pages_scanned: N, claims_extracted: 0, llm_unavailable: false }`. `llm_unavailable` stays false because the gateway-level guard passed, so the CLI never prints the "chat gateway unavailable" message.

Result: `takes extract --from-pages: 0 claim(s) from N page(s)` with no indication anything went wrong.

### Repro
On a brain with `facts.extraction_model = openai:gpt-4.1-mini` and **no** `ANTHROPIC_API_KEY`:

```
gbrain takes extract --from-pages --source-id default --max-pages 5 --yes
# -> "takes extract --from-pages: 0 claim(s) from 5 page(s)"
```

Set an Anthropic key and the same command extracts claims, confirming the model routing is the cause.

### Suggested fix
Honor `facts.extraction_model` as the JSDoc already promises, keeping the Anthropic Haiku id only as the last-resort fallback for brains that never set it (backward compatible):

```ts
const model =
  opts.model ??
  (await engine.getConfig('facts.extraction_model')) ??
  'anthropic:claude-haiku-4-5';
```

Resolving in `extractTakesFromPages` itself fixes both the CLI and the Minion handler in one place. The interactive consent prompt in `cmdExtract` ("...sends ... page content to Haiku.") should also be updated since the model is now configurable.

PR to follow.

rodbland2021:
Fixes #1857.

## Problem
`gbrain takes extract --from-pages` returns `0 claim(s) from N page(s)` (no error) on an OpenAI-configured brain with no `ANTHROPIC_API_KEY`, even though the chat gateway is healthy and `facts.extraction_model` is `openai:gpt-4.1-mini`.

The per-page classifier hardcoded the model:

```ts
// src/core/extract-takes-from-pages.ts
model: opts.model ?? 'anthropic:claude-haiku-4-5',
```

but the `model?` JSDoc on `ExtractTakesFromPagesOpts` says it "defaults to `facts.extraction_model`", and neither caller sets `opts.model`:
- `src/commands/takes.ts` `cmdExtract` (the CLI handler)
- `src/commands/jobs.ts` (the `extract-takes-from-pages` Minion handler)

So every run used the hardcoded Anthropic id. With no Anthropic key, each per-page `chat()` threw, got swallowed by the per-page `try/catch`, and produced 0 claims. `llm_unavailable` stayed `false` because the gateway-level `isAvailable('chat')` guard passed against the configured OpenAI model — so nothing was printed.

## Fix
Resolve the model from `facts.extraction_model` inside `extractTakesFromPages` (fixes both callers in one place), keeping `anthropic:claude-haiku-4-5` only as the last-resort fallback for brains that never set the config key — backward compatible:

```ts
const model =
  opts.model ??
  (await engine.getConfig('facts.extraction_model')) ??
  'anthropic:claude-haiku-4-5';
```

Also updates the interactive consent prompt in `cmdExtract`, which hardcoded "Haiku" in its text, to name the configured model instead.

## Behavior change
- Brains with `facts.extraction_model` set (most): takes extraction now uses that model, as documented. (Previously: always Anthropic Haiku.)
- Brains that never set `facts.extraction_model` AND have an Anthropic key: unchanged (fallback is still Haiku).
- OpenAI-only brains: now works instead of silently returning 0 claims.

## Verification
On an OpenAI-only brain (`facts.extraction_model = openai:gpt-4.1-mini`, no `ANTHROPIC_API_KEY`):

```
gbrain takes extract --from-pages --source-id default --max-pages 5 --yes
# before: 0 claim(s) from 5 page(s)
# after:  32 claim(s) from 5 page(s)
```

### DIFF
--- a/src/commands/takes.ts
+++ b/src/commands/takes.ts
@@ -616,7 +616,7 @@ async function cmdExtract(engine: BrainEngine, rest: string[]): Promise<void> {
   }
   if (!dryRun && !skipConfirm) {
     process.stderr.write(
-      `[takes extract] sends concept/atom/lore/briefing/writing/originals page content to Haiku.\n` +
+      `[takes extract] sends concept/atom/lore/briefing/writing/originals page content to the configured facts.extraction_model (Haiku if unset).\n` +
       `Pass --yes to proceed (or --dry-run to preview).\n`,
     );
     process.exit(1);
--- a/src/core/extract-takes-from-pages.ts
+++ b/src/core/extract-takes-from-pages.ts
@@ -126,6 +126,15 @@ export async function extractTakesFromPages(
   const dryRun = opts.dryRun ?? false;
   const maxPages = opts.maxPages ?? 50;
   const holder = opts.holder ?? 'system';
+  // Honor facts.extraction_model (the documented default per the `model?`
+  // JSDoc above) instead of a hardcoded Anthropic id, so a brain configured
+  // for OpenAI doesn't silently require an Anthropic key just for takes
+  // bootstrap. The hardcoded Haiku stays only as the last-resort fallback for
+  // brains that never set facts.extraction_model (backward compatibility).
+  const model =
+    opts.model ??
+    (await engine.getConfig('facts.extraction_model')) ??
+    'anthropic:claude-haiku-4-5';
   const sourceFilter = opts.sourceIdFilter ? `AND source_id = $1` : '';
   const params = opts.sourceIdFilter ? [opts.sourceIdFilter] : [];
 
@@ -174,7 +183,7 @@ export async function extractTakesFromPages(
     let response: { text: string };
     try {
       response = await chat({
-        model: opts.model ?? 'anthropic:claude-haiku-4-5',
+        model,
         system: CLASSIFIER_SYSTEM,
         messages: [
           {


Masashi-Ono0611:
Ran into the adjacent version of this today, want to share a data point in case it's still useful here even though the underlying model-routing bug already shipped via a different path (#3021, takeover of #2997 — resolves to `getChatModel()` / the file-plane `chat_model` config, not `facts.extraction_model` as proposed in this PR).

Real-world finding: on a brain that never explicitly sets `chat_model`, `getChatModel()` falls back to `DEFAULT_CHAT_MODEL = 'anthropic:claude-sonnet-4-6'` — **not Haiku**. So the original hardcoded-Haiku behavior this issue reported as a bug (silently 0 claims on non-Anthropic brains) got fixed, but on a default/unconfigured install the takes classifier now silently runs on Sonnet pricing ($3/$15 per M tok) instead of the Haiku pricing ($1/$5) it used before #3021. Measured this directly: 143 pages / 1176 takes came out to ~$1.15 by my estimate, about 3x what I'd budgeted assuming Haiku.

Not filing this as a fresh bug — more flagging that `facts.extraction_model` (this PR's approach) would have been a cheaper default for the common case (extraction is a cheap-model task) than routing through the general `chat_model`, which most installs set to a stronger model for actual conversation. Worth considering whether this PR (or a lighter-weight version of it) is still worth landing on top of #3021, or whether it's now superseded — you'd know the codebase intent here better than I would. Either way, the actual production behavior post-#3021 is solid otherwise (great claim quality on concept/atom pages, zero exact-duplicate claims across ~1200 generated).

time-attack:
Thanks for the precise root-cause writeup and the before/after verification — this was a real bug. It has since been fixed on master by dbf2b3f5 (#3021), which resolves the classifier model via `getChatModel()` instead of the hardcoded `anthropic:claude-haiku-4-5`, so OpenAI-configured brains no longer silently return 0 claims. Closing as already fixed; no action needed on your side.

Masashi-Ono0611:
Good to know — glad #3021 already covers this path. Thanks for tracing it back.

PR #1860, opened 2026-06-04
feat(auth): enforce bound_slug_prefixes on the READ side — per-prefix…

choomz:
## TL;DR

  `oauth_clients.bound_slug_prefixes` (migration v85, agent-binding wave) already
  bounds what a client may delegate to subagents on the WRITE side
  (`submit_agent`). This PR completes the binding on the READ side: a client
  bound to e.g. `['clients/*', 'people/*']` is now confined to matching slugs
  across every read surface. One column, one concept — the slug-prefix authority
  of the OAuth client.

  > Design-first PR: we built this because we needed it in production (one
  > company brain, per-teammate tokens), and we tried to follow the house
  > patterns closely (`sourceScopeOpts` ladder, fail-closed trust, engine
  > parity, no-oracle errors). If you'd rather fold per-prefix read ACL into
  > the mounts/brains access-policy story or a different shape, happy to
  > rework — flagging the need + a working implementation either way.

  ## Use case

  One brain, one vault, several team members with different access levels.
  Splitting the vault into sources breaks every `[[wikilink]]` (slugs lose
  their prefix); per-prefix token binding keeps the vault intact:

  ```bash
  gbrain auth register-client teammate-agent \
    --scopes read \
    --bound-slug-prefixes "clients/*,people/*,tools/*"

  The teammate's agent (a downstream OpenClaw, Claude Code over MCP, …) can
  search, read, and traverse clients/** and people/**, and never sees —
  or learns the existence of — anything under other prefixes.

  Semantics

  - Glob semantics are exactly matchesSlugAllowList (the existing
  write-side matcher): clients/* = all descendants (not the bare slug
  clients), a bare entry = exactly that slug. SQL side and TS side share
  the same contract, pinned by tests.
  - bound_slug_prefixes = NULL (every existing client) → unrestricted.
  Zero behavior change for current deployments; no migration needed (v85
  column already exists).
  - [] (explicit empty) → fail-closed, sees nothing — mirror of the
  allowedSources empty-array rule in sourceScopeOpts.
  - Out-of-binding single-slug reads throw page_not_found with the standard
  message — no existence oracle: indistinguishable from a missing page.
  - No per-call opt-out (deliberately unlike source_id='__all__'): the
  binding is a trust boundary, not a routing convenience.

  What's in the diff (by area)

  Auth plumbing
  - AuthInfo.boundSlugPrefixes projected by verifyAccessToken from
  oauth_clients.bound_slug_prefixes. Fallback chain extended so
  v61..v84 brains (federated_read present, binding column absent) retain
  their federated scope instead of degrading to the v60 projection.
  - registerClientManual + CLI gbrain auth register-client --bound-slug-prefixes p1,p2. On a
  pre-v85 schema it refuses loudly
  rather than silently minting an unbound client.

  Op layer (src/core/operations.ts) — slug-axis siblings of sourceScopeOpts
  - slugScopeOpts(ctx) — the canonical accessor (null = unrestricted).
  - assertSlugReadable(ctx, slug) — throws page_not_found (no oracle).
  - filterRowsBySlugBinding(ctx, rows, slugOf) — post-filter for
  list-shaped results.
  - Wired into: get_page (incl. fuzzy candidates), search, query (text
    - image branches), list_pages, get_tags, get_links,
  get_backlinks, traverse_graph (nodes pruned AND visible nodes' edges
  into hidden prefixes pruned), get_timeline, get_versions,
  get_raw_data, resolve_slugs, get_chunks, get_ingest_log,
  recall (entity_slug; NULL-entity rows dropped fail-closed for bound
  callers), takes_list, takes_search, find_experts, find_orphans,
  get_recent_salience, find_anomalies, find_trajectory,
  find_contradictions (both sides must be visible), think (gather
  streams: pages SQL-side, takes + graph post-filtered).

  SQL layer
  - SearchOpts.restrict_slug_prefixes (trust filter, distinct from the
  include/exclude ranking-policy knobs) + buildSlugAllowClause in
  sql-ranking.ts (positive inverse of buildHardExcludeClause,
  LIKE-escaped). Folded into the same interpolation site as the
  hard-exclude clause in searchKeyword / searchVector /
  searchKeywordChunks, both engines — every query shape that honors
  hard-excludes honors the binding by construction.
  - PageFilters.restrictSlugPrefixes on listPages, both engines.
  - hybrid.ts: the field is added to the explicit searchOpts pick (per
  the MUST-add comment for source-scoping fields), and restricted calls
  bypass query_cache entirely (binding is not in the cache key —
  same posture as adaptiveReturn; folding it into knobsHash is the
  natural follow-up if you'd rather cache).

  Tests

  - New test/e2e/slug-isolation-pglite.test.ts (18 cases), modeled on
  source-isolation-pglite.test.ts: engine layer + op layer, confinement
  on every surface, bare-vs-glob semantics, empty-binding fail-closed,
  LIKE-metachar escaping (a _ in a prefix can't act as a wildcard),
  link/backlink/graph leak guards, no-existence-oracle, unbound
  back-compat.
  - 2 new AuthInfo projection cases in test/oauth.test.ts.
  - bun run typecheck clean; oauth (93), source-isolation,
  sql-ranking, search, search-exclude, operations-trust-boundary
  suites green (251 cases).
  - Validated end-to-end on a production brain (~1.7K pages): bound
  client_credentials token → /token → /mcp — broad + targeted
  searches, direct get_page on hidden slugs, list_pages,
  resolve_slugs: zero leaks; legacy/unbound tokens unaffected.

  Known v1 limitation (documented in code)

  traverse_graph drops hidden nodes and prunes visible nodes' edges into
  hidden prefixes, but a multi-hop walk THROUGH a hidden node can still
  surface a visible far node — i.e. the existence of some path, never
  content. Stopping traversal at the binding frontier would be the stricter
  alternative; we kept v1 simple and called it out.

  Compatibility

  - No schema migration (v85 column reused).
  - No behavior change for any existing client (NULL binding = unrestricted).
  - New SearchOpts/PageFilters fields are optional and ignored when unset.

### DIFF
--- a/src/commands/auth.ts
+++ b/src/commands/auth.ts
@@ -344,6 +344,7 @@ interface RegisterClientArgs {
   scopes: string;
   sourceId: string;
   federatedRead: string[] | undefined;
+  boundSlugPrefixes: string[] | undefined;
   redirectUris: string[];
   tokenEndpointAuthMethod: string | undefined;
 }
@@ -354,6 +355,7 @@ export function parseRegisterClientArgs(args: string[]): RegisterClientArgs {
     scopes: 'read',
     sourceId: 'default',
     federatedRead: undefined,
+    boundSlugPrefixes: undefined,
     redirectUris: [],
     tokenEndpointAuthMethod: undefined,
   };
@@ -383,6 +385,14 @@ export function parseRegisterClientArgs(args: string[]): RegisterClientArgs {
         out.federatedRead = v.split(',').map(s => s.trim()).filter(Boolean);
         i += 2; break;
       }
+      case '--bound-slug-prefixes': {
+        // Read+write slug-prefix binding. Glob semantics of
+        // matchesSlugAllowList: `clients/*` = all descendants of clients/,
+        // a bare entry = exactly that slug. Omit for an unrestricted client.
+        const v = requireValue();
+        out.boundSlugPrefixes = v.split(',').map(s => s.trim()).filter(Boolean);
+        i += 2; break;
+      }
       case '--redirect-uri':
         out.redirectUris.push(requireValue());
         i += 2; break;
@@ -405,7 +415,7 @@ export function parseRegisterClientArgs(args: string[]): RegisterClientArgs {
 
 async function registerClient(name: string, args: string[]) {
   if (!name) {
-    console.error('Usage: auth register-client <name> [--grant-types G] [--scopes S] [--source SOURCE] [--federated-read SRC1,SRC2,...] [--redirect-uri URI ...] [--token-endpoint-auth-method client_secret_post|client_secret_basic|none]');
+    console.error('Usage: auth register-client <name> [--grant-types G] [--scopes S] [--source SOURCE] [--federated-read SRC1,SRC2,...] [--bound-slug-prefixes P1,P2,...] [--redirect-uri URI ...] [--token-endpoint-auth-method client_secret_post|client_secret_basic|none]');
     process.exit(1);
   }
   let parsed: RegisterClientArgs;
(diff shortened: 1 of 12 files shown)


time-attack:
Thank you for the careful, house-pattern-aware implementation and for being upfront that this is a design-first proposal. Per-prefix read ACLs are a new authorization surface rather than a fix to something broken, and a 733-line change spanning operations.ts, both engines, and the search path is more feature surface than we're taking on right now — access scoping is intended to live in the brains/mounts access-policy story. Closing as a feature we're not adopting; the write-side binding stays as-is.

PR #1862, opened 2026-06-04
fix(sync): coerce YAML-auto-parsed frontmatter scalars to string

michaeladair44:
## Summary

YAML auto-resolves unquoted scalars by tag — `title: 2026-06-03` therefore parses to a JavaScript `Date` object (not a string), `title: 2026` to a number, `title: true` to a boolean. `parseMarkdown` returned the raw value behind an `as string` cast that was a compile-time-only lie. Every consumer that called `.toLowerCase()` / `.replace()` / `.trim()` on `parsed.title` (`assessContentSanity`, `sanitizeTitle`, slug derivation, the intent-weights ranker) crashed with:

```
```

For `gbrain sync --source <obsidian-vault>`, the symptom was silent: each file with an unquoted ISO-date title (every daily-journal entry) was caught by `--skip-failed` and dropped. ~28 journal entries vanished from a single overnight sync run before the pattern was identified. The contained reproduction:

```yaml
---
title: 2026-06-03
type: journal
---
```

…crashes `assessContentSanity` at the line `opts.title.toLowerCase()`.

## Fix

- `parseMarkdown` now coerces `frontmatter.title` / `.slug` / `.type` via a shared `coerceFrontmatterString` helper so the `ParsedMarkdown.title: string` contract is honored at runtime:
  - `string` → trimmed string (passthrough)
  - `Date` → ISO `YYYY-MM-DD` (matches user's literal intent)
  - `number` / `boolean` → `String(...)`
  - `null` / `undefined` / `''` → `''` (caller's `||` fallback then fires)
  - `object` / `array` → `''` (falls through to `inferTitle` / `inferSlug` / `inferType`)
- `assessContentSanity` adds a defense-in-depth coercion at the title callsite. It has direct callers (MCP `put_page`, `lint`, `doctor`, `sources audit`, `quarantine`) that don't all flow through `parseMarkdown`, and the cost of a one-line `String(...)` guard is zero compared to a silent page-skip.

## Test plan

- [x] `test/markdown.test.ts` — new `frontmatter scalar coercion (YAML auto-typing)` describe block covers: unquoted ISO date, full timestamp, quoted ISO date (no-regression), numeric title, boolean title, null title (falls back to inference), unquoted ISO-date slug.
- [x] `test/content-sanity.test.ts` — new `non-string title coercion (defense-in-depth)` block exercises Date, `undefined`, and numeric titles handed directly to `assessContentSanity`.
- [x] 231/231 tests pass across the affected surfaces (`markdown`, `markdown-validation`, `markdown-serializer`, `content-sanity`, `content-sanity-literals`, `import-file-content-sanity`, `lint-content-sanity`, `ingestion/markdown-greenfield`).
- [x] `bunx tsc --noEmit` clean.
- [x] `bun run verify` — 29/30 checks pass. The one failure (`check:resolver`) is pre-existing on master and unrelated (orphan-trigger entries in a local skills dir).
- [x] Manual reproduction: minimal date-titled fixture crashed `assessContentSanity` before the fix; same fixture parses cleanly to `title: "2026-06-03"` (typeof string) after.

## Out of scope (filed if it survives investigation)

Mike's overnight sync also surfaced a trailing `The "data" argument must be of type string or an instance of Buffer, TypedArray, or DataView. Received undefined` error during sync finalization that leaves `sources.last_sync_at` un-advanced. No stack trace was captured in the original run; needs a reproduction before a targeted fix. Investigation narrowed it to the post-import phase (`extract` / `facts` / `embed`) but a real reproduction is required to land a clean fix.

Co-Authored-By: Michael Adair <michaeladair44@gmail.com>

### DIFF
--- a/src/core/content-sanity.ts
+++ b/src/core/content-sanity.ts
@@ -376,14 +376,22 @@ export function assessContentSanity(opts: {
   // doesn't repeat the lowercase per literal.
   const bodyHead = body.slice(0, SCAN_HEAD_BYTES);
   const bodyHeadLower = bodyHead.toLowerCase();
-  const titleLower = opts.title.toLowerCase();
+  // Defense-in-depth coercion. parseMarkdown now normalizes frontmatter
+  // titles via coerceFrontmatterString (e.g. YAML auto-parses unquoted
+  // `title: 2026-06-03` to a Date), but assessContentSanity is also called
+  // directly from MCP put_page, lint, sources audit, doctor, and quarantine
+  // — any of which could be handed a non-string title from a caller that
+  // hasn't gone through parseMarkdown. Skipping a journal entry because a
+  // .toLowerCase() crashed is a worse failure than a one-line String() guard.
+  const titleStr = typeof opts.title === 'string' ? opts.title : String(opts.title ?? '');
+  const titleLower = titleStr.toLowerCase();
 
   const junk_pattern_matches: string[] = [];
   for (const p of BUILT_IN_JUNK_PATTERNS) {
     const scope = p.applies_to ?? 'both';
     let matched = false;
     if (scope === 'title' || scope === 'both') {
-      if (p.pattern.test(opts.title)) matched = true;
+      if (p.pattern.test(titleStr)) matched = true;
     }
     if (!matched && (scope === 'body' || scope === 'both')) {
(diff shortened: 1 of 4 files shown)


garrytan:

PR #1863, opened 2026-06-04
fix(search): weak-top floor on autocut to stop cross-source result collapse

rayers:
## Problem

On a real multi-source brain, `gbrain query "<rare specific terms>" --source __all__` collapses to **1 result**, while a common-term query (`playback`) returns the full set. A single-source scoped query for the same rare terms returns ~12 results — so results that appear when scoped **vanish** when searching all sources (`scoped ⊄ __all__`).

## Root cause

`applyAutocut` (`src/core/search/autocut.ts`) trims the ranked list at the largest gap in cross-encoder rerank scores, where the gap is **normalized by the top score**. On a rare-term cross-source pool the top rerank score is itself *weak* (measured: `0.317`), but normalizing rescales it to `1.0`, so a normal decay below it looks like a confident cliff. autocut then declares an "obvious single answer" and collapses the pool to 1. Single-source pools have flat scores → no cliff → autocut correctly keeps many, which is why the two paths diverge.

Calibration across query classes on a live brain (zerank-2):

| query | top rerank | autocut result | correct? |
|---|---|---|---|
| rare multi-term | **0.317** | collapses to 1 | ✗ bug |
| keyword-#1 lookup | 0.981 | keeps 2 | ✓ |
| broad ("playback") | 0.956 | keeps 12 | ✓ |
| genuine single answer | 0.984 | collapses to 1 | ✓ (real cliff) |

Every legitimate result has `top ≥ 0.95`; the only bug case is `0.317`. The reranker is bimodal (real ≈0.95+, weak ≈0.3).

## Fix

Add `AutocutConfig.minTopScore` (default **0.5**): when the top rerank score is below the floor, autocut no-ops and preserves recall; at/above it, behavior is unchanged. `0.5` sits in the empty middle of the bimodal distribution. Tunable via `search.autocut_min_top_score`, threaded exactly like `autocut_jump` (mode bundle → config override → pick-chain → knobsHash). `KNOBS_HASH_VERSION` 9→10 (the floor changes the trimmed set, so cache rows must not cross-serve; `acmts` is hashed at 4 decimals so near floors don't collide).

### Reranker-scale safety

`rerank_score` is the provider's **raw** relevance score (no normalization in the gateway). The `[0,1]` floor is only meaningful for normalized rerankers (ZE `zerank-*`, Cohere, Voyage), **not** the shipped raw-logit `llama-server-reranker` / Qwen3-Reranker path (unbounded logits). So `hybridSearch` gates the floor on `rerankerEmitsNormalizedScores(<effective reranker model>)` — honoring a per-call `SearchOpts.reranker.model` override, not just the bundle default — and passes `0` (disabled) for raw-logit rerankers. The compute-path default mirrors the knobsHash default (`0.5`) so a partial-knobs caller can't compute a floored set while stamping the cache as unfloored.

## Tests

- Pure-fn: weak-top → no trim; strong-top → still trims; partial-knobs default agreement; `minTopScore:0` disables.
- End-to-end (`hybridSearch` + stubbed reranker): weak-top no-trim, strong-top trim, **per-call raw-logit model disables the floor**.
- Plumbing: bundle defaults, per-call/config/override precedence, `loadOverridesFromConfig` clamp `[0,1]`, knobsHash separation, `SEARCH_MODE_CONFIG_KEYS`.
- `rerankerEmitsNormalizedScores` unit coverage (cloud vs raw-logit, case-insensitive).

Live verify: the repro query goes 1 → 20 results via `--source __all__`; broad/single-answer/keyword-#1 queries unchanged.

### DIFF
--- a/src/commands/search.ts
+++ b/src/commands/search.ts
@@ -71,6 +71,7 @@ const KNOB_DESCRIPTIONS: Record<keyof ModeBundle, string> = {
   // v0.42.3.0 autocut
   autocut: 'Score-discontinuity result-sizing (cuts at the rerank-score cliff; no-op without a reranker)',
   autocut_jump: 'Autocut sensitivity: min normalized score gap that counts as a cliff (0..1, 0.20 default)',
+  autocut_min_top_score: 'Autocut weak-top floor: min top rerank score to trust the cliff (0..1, 0.50 default; 0 disables; below it autocut no-ops to preserve recall)',
   // v0.43 relational recall
   relationalRetrieval: 'Typed-edge relational recall arm (relational queries walk the graph; no-op otherwise)',
   relational_retrieval_depth: 'Max hops for relational traversal (1..3, 2 default)',
--- a/src/core/search/autocut.ts
+++ b/src/core/search/autocut.ts
@@ -33,6 +33,19 @@ export interface AutocutConfig {
   jumpRatio: number;
   /** Failsafe: never return fewer than this when candidates exist (≥1). */
   minKeep: number;
+  /**
+   * Weak-top floor: the minimum TOP rerank score for the cliff test to be
+   * trusted. The cliff is measured on scores NORMALIZED by the top, so a weak
+   * top (e.g. 0.317 on a rare-term cross-source query) rescales to 1.0 and a
+   * normal decay below it masquerades as a confident cliff — collapsing a rich
+   * pool to 1. When the top score is below this floor, autocut no-ops (recall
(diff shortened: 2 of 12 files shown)


time-attack:
Great diagnosis — the bimodal reranker calibration table made the weak-top failure mode obvious, and the bug is genuinely still present on master. The work has been carried forward in #3131, which is the same fix rebased onto current master (your version bumps KNOBS_HASH_VERSION 9→10, but master is now at 12). Closing as superseded by #3131 so the two don't conflict; thanks for finding and characterizing this.

Issue #1162, opened 2026-05-18
autopilot: reconnect loop after 5 consecutive failures logs "config.database_url undefined" indefinitely, never exits


colin477:
## Summary

When the autopilot hits its 5-consecutive-cycle-failure threshold and logs `5 consecutive cycle failures. Stopping autopilot.`, the process **does not actually exit**. Instead it enters a reconnect-retry loop that logs three lines every 5 minutes forever:

```
[2026-05-18T15:16:53] [reconnect] ERROR: undefined is not an object (evaluating 'config.database_url')
[2026-05-18T15:16:53] [dispatch] ERROR: No database connection: connect() has not been called. Fix: Run gbrain init --supabase or gbrain init --url <connection_string>
[2026-05-18T15:16:53] [health]   ERROR: No database connection: connect() has not been called. Fix: Run gbrain init --supabase or gbrain init --url <connection_string>
```

`~/.gbrain/config.json` on disk is still valid throughout (e.g. `gbrain query` from the same machine works fine). The autopilot's in-process config reference has gone undefined and the reconnect path doesn't re-read from disk before retrying.

## Repro

1. Create connection pressure that causes autopilot cycles to fail. In my case the trigger was Supabase session-pool exhaustion (`(EMAXCONNSESSION) max clients reached in session mode - max clients are limited to pool_size: 15`) from accumulated `gbrain serve` processes across days — see companion issue **[link to serve-accumulation issue]**. Any sustained DB-unreachability of similar shape should reproduce.
2. Wait for 5 consecutive `autopilot-cycle` job failures.
3. Observe `5 consecutive cycle failures. Stopping autopilot.` in `~/.gbrain/autopilot.err`.
4. `ps aux | grep gbrain.*autopilot` — the process is still running.
5. `tail -f ~/.gbrain/autopilot.err` — the three-line reconnect/dispatch/health error block repeats every 5 minutes (the autopilot interval), forever.

## Diagnosis

Stack trace from the most recent occurrence:

```
GBrainError: No database connection: connect() has not been called.
    at getConnection (src/core/db.ts:153)
    at transaction (src/core/postgres-engine.ts:524)
    at failJob (src/core/minions/queue.ts:846)
    at executeJob (src/core/minions/worker.ts:715)
```

[...]

PR #1865, opened 2026-06-04
fix: wikilink slug case-normalization (silently dropped edges) + autopilot reconnect recovery

awilkinson:
Two small but high-impact bug fixes, both found and verified against a live ~1,900-page Postgres brain.

## 1. `link-extraction`: wikilink/markdown slug targets aren't case-normalized

`addLinksBatch` inserts edges with `INSERT ... SELECT FROM unnest(...) JOIN pages ON p.slug = target`. That JOIN is case-sensitive, so any link whose target isn't already lowercased silently produces **no row** — the edge is dropped and the target page looks orphaned.

In practice this hits every index/hub link written in mixed case, e.g. `[[people/INDEX]]` → target `people/INDEX` → no match for the real page `people/index` → edge dropped. `dead_links` stays 0 (nothing is stored), so the failure is invisible; the page just shows up as an orphan.

**Fix:** lowercase the slug (not the display name) in all three `extractEntityRefs` passes (markdown link, qualified wikilink, unqualified wikilink). Canonical slugs are already lowercase (the resolver regex enforces it), so this is safe.

Also extends `DIR_PATTERN` with additional knowledge-base directories that were silently un-extractable (`deals` — note the existing entry was singular `deal` only — plus `analysis`, `medical`, `writing`, `group-chats`, `tools`, `vendors`, `books`). Longer term `DIR_PATTERN` would be better as configurable rather than hardcoded; this is the immediate fix.

**Verified in production:** after deploying, pages with inbound links appeared in directories that previously had zero (`deals/` 42, `medical/` 28, `group-chats/` 123, `analysis/` 11, `books/` 11), with no change to the already-working `people/`/`companies/`.

## 2. `autopilot`: reconnect path calls `connect()` with no arguments

In the cycle DB-health check, on a dropped connection the recovery path calls `await (engine as any).connect?.()` with **no config**, so `PostgresEngine.connect(config)` immediately evaluates `config.poolSize` on `undefined` and throws `undefined is not an object (evaluating 'config.poolSize')`. The daemon can never recover from a dropped connection and trips the consecutive-failure cap. (The `#1162` error-classification wraps this call but doesn't fix the root cause.)

**Fix:** call `engine.reconnect()`, which already exists and tears down the dead pool + re-dials using the config saved at startup.

Observed in production as the daemon going dead for ~2 weeks after a connection blip; after the fix it survives drops and resumes cycling.

---

Happy to add unit tests for the slug case-normalization (the `extractEntityRefs` lowercasing is straightforward to cover) and split into two PRs if preferred.

### DIFF
--- a/src/commands/autopilot.ts
+++ b/src/commands/autopilot.ts
@@ -529,8 +529,8 @@ export async function runAutopilot(engine: BrainEngine, args: string[]) {
       autopilotReconnectFails = 0; // reset on success
     } catch (probeErr) {
       try {
-        await engine.disconnect();
-        await (engine as any).connect?.();
+        // Reuse saved config; connect() with no args crashed on config.poolSize.
+        await (engine as any).reconnect?.();
         autopilotReconnectFails = 0;
       } catch (e) {
         logError('reconnect', e);
--- a/src/core/link-extraction.ts
+++ b/src/core/link-extraction.ts
@@ -82,7 +82,7 @@ export type LinkResolutionType = 'qualified' | 'unqualified';
  *   - Our domain extensions: tech, finance, personal, openclaw (domain-organized wikis)
  *   - Our entity prefix: entities (we kept some legacy entities/projects/ pages)
  */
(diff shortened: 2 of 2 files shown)


garrytan:

PR #1877, opened 2026-06-05
fix(import): normalize slug casing at importFromContent entry (mixed-case slugs no longer 404)

zach22-1999:
## What

`importFromContent` now normalizes the slug once at its entry point (`slug = validateSlug(slug)`), so every per-page tx call agrees on casing.

## Why (root cause)

`putPage` runs `validateSlug` (→ lowercase) internally, but `addTag` and `upsertChunks` do **not** — and `importFromContent` passed the caller's original slug straight through to all of them.

So a **mixed-case slug** wrote the page row under a lowercased key, while `addTag` queried `WHERE slug = '<original-case>'` (a miss) and threw `addTag failed: page "..." not found`, rolling back the whole import. The row then became **permanently unwritable** under that slug.

This bites every ingestion source whose IDs carry uppercase — Lark/Feishu doc tokens (mixed-case base62), GitHub repo paths, etc. — surfacing as a spurious "Page not found" from `gbrain capture` / `put_page`.

## Minimal repro (before this PR)

```
echo body | gbrain capture --stdin --slug Mixed-CASE   # put_page failed: Page not found: Mixed-CASE
echo body | gbrain capture --stdin --slug mixed-case   # works
```

## Fix

One-line normalization at the `importFromContent` entry point — covers `capture` / `put_page` / `sync` (all ingestion flows through here). `putPage`'s internal `validateSlug` stays (idempotent).

## Test

`test/slug-case-regression.test.ts` — imports a mixed-case slug with a tag and asserts it round-trips (retrievable lowercased, tag reconciled). Pre-fix this threw at `addTag`.

## Note

The deeper inconsistency is at the engine layer: `addTag` / `upsertChunks` don't run `validateSlug` while `putPage` does. Normalizing at the `importFromContent` entry fixes all ingestion paths with the smallest diff; glad to also add `validateSlug` to `addTag`/`upsertChunks` in both engines if you'd prefer defense-in-depth.

### DIFF
--- a/src/core/import-file.ts
+++ b/src/core/import-file.ts
@@ -36,7 +36,7 @@ import {
 } from './embedding-context.ts';
 import { loadSearchModeConfig, resolveSearchMode } from './search/mode.ts';
 import { normalizeAliasList } from './search/alias-normalize.ts';
-import { isUndefinedTableError, warnOncePerProcess } from './utils.ts';
+import { isUndefinedTableError, warnOncePerProcess, validateSlug } from './utils.ts';
 import { computeCorpusGeneration } from './contextual-retrieval-service.ts';
 import { runGuardrails } from './guardrails.ts';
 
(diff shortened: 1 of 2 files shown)


time-attack:
Thanks @zach22-1999! 🙏

What this does: This PR fixes a bug where mixed-case slugs would fail with "Page not found" errors during import, because putPage lowercases them internally but addTag and upsertChunks do not. The fix normalizes the slug once at the importFromContent…

This overlaps with #2436, which we're carrying forward, so I'm closing this in its favor to keep the review queue focused. If yours covers a case #2436 misses, reopen with that detail and we'll reconcile them.

Issue #427, opened 2026-04-25
bun auto-loads project .env from CWD — DATABASE_URL silently overrides gbrain config with no warning


zek01svg:
## Summary

If you run `gbrain` from a project directory that has `DATABASE_URL` in its `.env` file (common in Next.js, Hono, Supabase, and most Node/bun web apps), bun automatically loads that `.env` and `DATABASE_URL` silently overrides the URL in `~/.gbrain/config.json`. gbrain connects to the wrong database with no indication that a project `.env` is involved.

The error message points the user at the wrong place:

```
Cannot connect to database: Tenant or user not found.
Fix: Check your connection URL in ~/.gbrain/config.json
```

The URL in `config.json` is correct. The `.env` file is the culprit — but nothing says so.

## Reproduction

1. Have a working gbrain install with `~/.gbrain/config.json` pointing at your brain (e.g. Supabase project A).
2. Create a web project with a `.env` containing `DATABASE_URL=<some other postgres URL>` (e.g. the app's own Supabase project B).
3. `cd` into that project and run `gbrain stats`.
4. gbrain connects to project B (or fails with "Tenant or user not found" if project B does not exist), with no warning.

## Environment

- OS: Windows 11 (also affects any OS where the project `.env` has `DATABASE_URL`)
- Bun: 1.x (bun auto-loads `.env` from CWD by default)
- gbrain: 0.18.2
- Brain backend: Supabase postgres pooler (port 6543)

## Root cause

`loadConfig()` in `src/core/config.ts`:

```ts
const dbUrl = process.env.GBRAIN_DATABASE_URL || process.env.DATABASE_URL;
```

`DATABASE_URL` is used by virtually every web framework that talks to Postgres (Prisma, Drizzle, Supabase client, etc.). When bun loads the project `.env`, it injects `DATABASE_URL` into `process.env` before gbrain reads it, so it wins over `config.json`.

## Why this is hard to debug

- The error says "Check your connection URL in `~/.gbrain/config.json`" — config.json is fine.
- No indication that an env var is active or where it came from.
- Took ~2 hours to isolate by adding `console.log` to `loadConfig()`.

## Suggested fix

When `DATABASE_URL` (not `GBRAIN_DATABASE_URL`) is the active source, emit a one-time warning:

```
[gbrain] Warning: DATABASE_URL env var is overriding ~/.gbrain/config.json.
         This may be from a project .env file. Use GBRAIN_DATABASE_URL to
         avoid conflicts with application databases.
[...]

PR #1884, opened 2026-06-05
fix(config): generic DATABASE_URL silently overrides a configured brain

awilkinson:
## Problem
`loadConfig()` resolves the database URL as:
```ts
const dbUrl = process.env.GBRAIN_DATABASE_URL || process.env.DATABASE_URL;
```
and that value then **overrides `config.json`** in the merge below it.

The generic `DATABASE_URL` is an extremely common env var — many apps/frameworks set it. If gbrain runs in an environment where another project's `DATABASE_URL` is present (a co-located app, a sourced `.env`, a dev shell), gbrain **silently connects to that other database instead of the brain configured in `config.json`** — no error, just wrong data (e.g. `gbrain stats` showing 0 pages, or worse, auto-running migrations against an unrelated database on connect).

We hit this in a real fleet: a machine whose `config.json` correctly pointed at the shared brain still resolved to a different Supabase project because a `DATABASE_URL` for a separate app was in the environment.

## Fix
Make precedence: **`GBRAIN_DATABASE_URL` > `config.json` `database_url` > generic `DATABASE_URL`.**
```ts
const dbUrl = process.env.GBRAIN_DATABASE_URL || fileConfig?.database_url || process.env.DATABASE_URL;
```
- `GBRAIN_DATABASE_URL` stays the intentional, gbrain-specific operator override (unchanged).
- A configured brain (`config.json.database_url`) is no longer hijacked by a co-located `DATABASE_URL`.
- PGLite brains have no `config.json.database_url`, so a bare `DATABASE_URL` still upgrades them to Postgres — the existing operator escape hatch (and the engine-inference comment) is preserved.

One-line change in `src/core/config.ts`. Happy to add a `test/config.test.ts` case asserting config.json wins over a stray `DATABASE_URL`.

### DIFF
--- a/src/core/config.ts
+++ b/src/core/config.ts
@@ -383,8 +383,13 @@ export function loadConfig(): GBrainConfig | null {
     fileConfig = migrateLegacyEmbeddingConfig(parsed) as unknown as GBrainConfig;
   } catch { /* no config file */ }
 
-  // Try env vars
-  const dbUrl = process.env.GBRAIN_DATABASE_URL || process.env.DATABASE_URL;
+  // Try env vars. GBRAIN_DATABASE_URL is the intentional gbrain-specific operator
+  // override and always wins. The generic DATABASE_URL is only a LAST-RESORT
+  // fallback — it must not override a brain already configured in config.json, or
+  // a co-located app's DATABASE_URL in the same environment silently repoints
+  // gbrain at the wrong database. (PGLite brains have no config database_url, so a
+  // bare DATABASE_URL still upgrades them to Postgres — the operator escape hatch.)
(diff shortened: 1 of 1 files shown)


time-attack:
Thanks for the report — the wrong-database failure mode is nasty and we take it seriously. Master already guards this class via the #427 fix in `src/core/config.ts`: a `DATABASE_URL` whose value comes from a cwd `.env` is ignored with a stderr notice, which covers the co-located-app scenario you describe. The remaining case (a deliberately exported `DATABASE_URL`) is intentional operator precedence — inverting it would also repoint the `DATABASE_URL`-gated e2e Postgres tests at a configured brain — so we're keeping the current ordering and closing this.

PR #1885, opened 2026-06-05
fix(autopilot): stamp federated freshness cycles

DenverEvolution:
## Summary

- Dispatch federated freshness work as per-source `autopilot-cycle` jobs instead of sync-only jobs, so `sources.config.last_full_cycle_at` is stamped and `doctor` cycle freshness reflects real full-cycle state.
- Avoid duplicate targeted remediation after freshness-cycle fanout dispatch.
- Treat fresh autopilot lock files as active only when their recorded PID is still alive, so dead fresh locks do not block restart.
- Add an `extract_facts` full-walk fast-path for sources with no fence facts and no pages containing a Facts fence, avoiding thousands of no-op per-page DELETE round trips on remote Postgres/Supabase.

## Why

A live federated setup had autopilot running but `gbrain doctor` still failed `cycle_freshness`: the freshness gate submitted `sync` jobs, which update `last_sync_at` but never update `last_full_cycle_at`. The same incident exposed stale fresh lock files after restarts, and very slow no-op full-walk `extract_facts` scans over thousands of pages with an empty facts table.

## Tests

```sh
timeout 180 bun test \
  test/extract-facts-phase.test.ts \
  test/autopilot-fanout-wiring.test.ts \
  test/autopilot-fanout.test.ts \
  test/cycle-last-full-cycle-at.test.ts
```

Result: 54 pass, 0 fail.

## Live verification

On the affected runtime after applying the hotfix:

- `catalyst-os` and `catalyst-repo` both stamped fresh `last_full_cycle_at`.
- `gbrain --progress-json doctor --json` reports `cycle_freshness` as `ok`.
- Remaining doctor warning is unrelated: `takes_count` because takes bootstrap is disabled.

### DIFF
--- a/src/commands/autopilot.ts
+++ b/src/commands/autopilot.ts
@@ -351,13 +351,28 @@ export async function runAutopilot(engine: BrainEngine, args: string[]) {
   try {
     mkdirSync(gbrainHomePath(), { recursive: true });
     if (existsSync(lockPath)) {
-      const stat = require('fs').statSync(lockPath);
+      const fs = require('fs');
+      const stat = fs.statSync(lockPath);
       const ageMinutes = (Date.now() - stat.mtimeMs) / 60000;
-      if (ageMinutes < 10) {
+      const lockPidRaw = fs.readFileSync(lockPath, 'utf8').trim();
+      const lockPid = Number(lockPidRaw);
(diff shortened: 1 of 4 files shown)


DenverEvolution:
Closing: this hotfix is intended to remain local to Chris's GBrain runtime, not submitted upstream.

PR #1894, opened 2026-06-05
feat(sync): expose repeatable --exclude <glob> on gbrain sync

billy-armstrong:
## What

Adds a repeatable `--exclude <glob>` flag to the single-source `gbrain sync` CLI path.

## Why

`core/sync.ts` already supports `SyncableOptions.exclude` — `classifySync()` returns `'exclude-glob-hit'` via `matchesAnyGlob()`, and it's covered by `test/sync-isSyncable-shape.test.ts` (e.g. `isSyncable('drafts/wip.md', { exclude: ['drafts/**'] }) === false`). But the `gbrain sync` command never exposed it, so the capability was unreachable from the CLI.

This is useful when a single git clone holds subtrees that belong to **different sources**. Without `--exclude`, `gbrain sync --source <parent>` walks the whole working tree and routes every file into the parent source. With it, the parent sync can skip a subtree, and that subtree is imported separately under its own id:

```
gbrain sync   --source parent-source --exclude 'subtree-a/**' --exclude 'subtree-b/**'
gbrain import subtree-a/ --source-id source-a
gbrain import subtree-b/ --source-id source-b
```

## How

Five small edits in `src/commands/sync.ts`, all wiring (no new logic — the filter already exists in core):

1. `SyncOpts.exclude?: string[]` field
2. thread it into the `isSyncable()` filter options
3. `--exclude` help text
4. parse the repeatable flag from argv
5. pass the parsed globs into the single-source `SyncOpts`

## Scope

Single-source path only. `--all` and `sync trigger` are intentionally unchanged — global exclusion across heterogeneous sources, and exclusion persisted into an enqueued job payload, are different concerns best handled separately.

## Test / verify

`bun run typecheck` clean. Core filter behavior is already covered by `test/sync-isSyncable-shape.test.ts`. Manual:

```
gbrain sync --help | grep -- --exclude
gbrain sync --source <id> --dry-run --exclude 'some/subtree/**'
```

### DIFF
--- a/src/commands/sync.ts
+++ b/src/commands/sync.ts
@@ -389,6 +389,15 @@ export interface SyncOpts {
    * Precedent: CycleOpts.signal at src/core/cycle.ts (v0.22.1 #403).
    */
   signal?: AbortSignal;
+  /**
+   * Repeatable `--exclude <glob>` patterns, threaded into the `isSyncable()`
+   * filter. Lets a single git clone fan out into multiple sources: a subtree
+   * that should be imported under a different `--source-id` is skipped by the
+   * sync of the parent source. Globs use the same `matchesAnyGlob()` syntax as
+   * `SyncableOptions.exclude` in core/sync.ts (`**` = any path segments,
+   * `*` = non-slash chars). Honored on the single-source CLI path.
+   */
+  exclude?: string[];
 }
 
 /**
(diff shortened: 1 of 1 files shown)


billy-armstrong:
Closing — handling this as a local patch instead of upstreaming. Thanks.

PR #1896, opened 2026-06-05
fix(extract): recognize plain-bullet timeline format (- YYYY-MM-DD — Summary)

### LINKED ISSUE #1895: extractTimelineFromContent ignores enrich skill's plain-bullet timeline format → timeline_coverage stuck at 0%
## Root cause

`extractTimelineFromContent` (`src/commands/extract.ts`) only recognized two timeline line formats:

- `- **DATE** | Source — Summary`
- `### DATE — Title`

But gbrain's own `enrich` skill writes plain bullets: `- YYYY-MM-DD — Summary` (no bold, no source pipe). As a result every brain-authored timeline entry was invisible to the extractor.

## Impact

`gbrain extract all` created **0** timeline entries from 41 pages, and `doctor` reported `timeline_coverage` pinned at **0%** — even on a brain that was diligently maintaining timelines via the enrich skill.

## Suggested fix

Add a third recognized format for the plain bullet (`- YYYY-MM-DD — Summary`), anchored at line start so a date inside a summary/link can't start a spurious entry, and so it can't double-count the bold Format-1 lines.

## Verification

After adding the plain-bullet format: `gbrain extract timeline` created **26** entries from 41 pages; `doctor` `timeline_coverage` went **0% → 59%**, brain score **78 → 84**.

PR to follow.

ElliotDrel:
Closes #1895

## Root cause

`extractTimelineFromContent` (`src/commands/extract.ts`) only matched `- **DATE** | Source — Summary` and `### DATE — Title`, but gbrain's own `enrich` skill writes plain bullets (`- YYYY-MM-DD — Summary`, no bold, no source pipe). Every brain-authored timeline entry was therefore invisible: `gbrain extract all` created 0 timeline entries from 41 pages and `timeline_coverage` was pinned at 0%.

## Fix

Add a third format (Format 3) for the plain bullet. Anchored at line start so a date inside a summary/link can't start a spurious entry; bold Format-1 lines can't match (a `*` follows the bullet, not a digit) so the patterns never double-count.

## Verification

`gbrain extract timeline` now creates 26 entries from 41 pages; `doctor` `timeline_coverage` 0% → 59%, brain score 78 → 84. 5 unit tests added (`test/extract.test.ts`).

### DIFF
--- a/src/commands/extract.ts
+++ b/src/commands/extract.ts
@@ -492,6 +492,18 @@ export function extractTimelineFromContent(content: string, slug: string): Extra
     entries.push({ slug, date: match[1], source: 'markdown', summary: match[2].trim(), detail: detail || undefined });
   }
 
+  // Format 3: Plain bullet — - YYYY-MM-DD — Summary
+  // This is the format gbrain's own enrich skill writes (no bold, no source
+  // pipe). Without it, every brain-authored timeline entry is invisible to
+  // extraction and timeline_coverage stays at 0%. Anchored at line start so a
+  // date inside a summary/link cannot start a spurious entry; a bold Format-1
+  // line (`- **…**`) cannot match here (a `*` follows the bullet, not a digit),
+  // so the two patterns never double-count.
+  const plainBulletPattern = /^-\s+(\d{4}-\d{2}-\d{2})\s*[—–-]\s*(.+)$/gm;
+  while ((match = plainBulletPattern.exec(content)) !== null) {
+    entries.push({ slug, date: match[1], source: 'markdown', summary: match[2].trim() });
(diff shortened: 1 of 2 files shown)


time-attack:
Good catch — the plain-bullet timeline format really was invisible to the extractor, and your regex anchoring is careful. Master has since added an inline-citation arm, which means this pattern as written would double-count any plain bullet that carries its own `[Source: …]` citation; #3079 carries your fix forward with that dedupe skip plus a test for it. Closing as superseded by #3079 — thanks for the find. (Minor note for future PRs: test fixtures should use generic placeholder names rather than real people.)

PR #1900, opened 2026-06-05
feat(reference): reference-only entity flag, exempt from coverage metrics

### LINKED ISSUE #1899: Canon/reference figures (imported from books/articles) sit permanently red in entity-coverage metrics with no actionable fix
## Problem

Canon/reference figures imported from books and articles — Andy Grove, Kleiner Perkins, etc. — are real `type: person`/`company` pages. They're searchable, enrichable, and linkable, which is correct. But they have **no dated history in the user's own life**, so they sit permanently red in `timeline_coverage` and `entity_link_coverage`.

There's no actionable fix for these pages: the user can't add a personal timeline to someone they only read *about*, yet the metric keeps flagging them. This drags coverage scores down and trains the user to ignore the metric.

## Suggested fix

An opt-in `reference: true` frontmatter flag that exempts a page from entity-coverage metrics **only** (timeline_coverage, entity_link_coverage), while leaving `type: person`/`company` intact so retrieval/linking is unchanged. A small CLI command to set/unset the flag, applied to both the markdown frontmatter (durable) and the engine JSONB (immediate).

## Verification

After marking 11 canon pages on a live PGLite brain: `timeline_coverage` 59% → 90% (WARN→OK), `entity_link_coverage` → 86% (OK), graph_coverage OK.

PR to follow.

ElliotDrel:
Closes #1899

## Root cause

Canon/reference figures imported from books/articles (Andy Grove, Kleiner Perkins, …) are real `type: person`/`company` pages — searchable, enrichable, linkable — but have no dated history in the user's own life, so they sit permanently red in `timeline_coverage` / `entity_link_coverage` with no actionable fix.

## Fix

Opt-in `reference: true` frontmatter flag that exempts a page from those entity-coverage metrics ONLY; nothing about retrieval changes.

- `src/core/reference-flag.ts`: single source of truth for the exclusion predicate `(frontmatter->>'reference') IS DISTINCT FROM 'true'` (`referenceExclusionSql`), backed by the existing GIN index on `pages.frontmatter`. Absent/false/non-true = normal entity (default).
- Apply the predicate to BOTH numerator and denominator at every entity-coverage site: `getHealth()` entity_pages CTE (pglite + postgres engines), and `onboard/checks.ts` + `init-nudge.ts` (timeline_coverage, entity_link_coverage). postgres-engine inlines the literal (tagged-template can't take raw-SQL `${}`). `most_connected` (top-N display) is intentionally NOT filtered.
- New `gbrain reference <slug> [--unset]` command: writes the flag to BOTH the markdown frontmatter (durable, survives re-ingest/rebuild) and the engine JSONB (immediate, no re-sync). Minimal-diff frontmatter edit.
- Skills: `enrich` gets a "Reference entities" convention; `book-mirror` & `article-enrichment` get a follow-on to mark figures the reader only reads ABOUT. Canonical convention doc at `skills/conventions/reference-entities.md`; `maintain` + `docs/architecture/KEY_FILES.md` updated.
- Tests: `referenceExclusionSql` output + `applyReferenceFrontmatter` (add, idempotent, replace-false, unset, no-fm, YAML-special-char body). 10/10.

## Verification

Live on the PGLite brain: after marking 11 canon pages, `timeline_coverage` 59% → 90% (WARN→OK), `entity_link_coverage` → 86% (OK), graph_coverage OK. (brain_score's timeline component is a separate whole-brain density metric and is intentionally unaffected.)


time-attack:
Thanks for thinking about coverage-metric accuracy — reference-only entities skewing the score is a real annoyance. That said, this is a 337-line new feature adding an entity flag and metric-exemption path rather than a fix to something broken, and we're holding the line against new surface area in the coverage/metrics layer. Closing as a feature we're not taking; the metric can stay imperfect rather than grow a flag to maintain.

Issue #1698, opened 2026-05-31
think --model silently degrades to 'no LLM available' and SAVES an empty synthesis instead of erroring


garrytan:
## Symptom

Running a batch of `gbrain think --anchor <slug> --take --save --model "anthropic/claude-sonnet-4-6"` (slash form) on v0.41.38.0, the **first pass silently produced empty synthesis pages**. No model ran, no error was thrown, and `--save` still persisted a near-empty synthesis page + take row. The failure only became visible because the *output* was obviously empty on inspection — the command exited 0.

Re-running the identical batch with the colon form `--model "anthropic:claude-sonnet-4-6"` worked perfectly: 95/95 people + 104/104 company pages enriched with real, cited synthesis.

## Why this is surprising

v0.41.21.0's `parseModelId` (src/core/ai/model-resolver.ts) explicitly added slash-form support, with a comment that this 'Closes the end-to-end bug class.' Yet on v0.41.38.0 the slash form still failed for `think` — either:
- `think`'s model-resolution path doesn't route through the fixed `parseModelId`, or
- it catches the resolution failure and falls back to a no-LLM/title-only path instead of surfacing it.

Either way the **fail-silent + save-anyway behavior is the real bug**, independent of the slash/colon parsing question. A model override that can't be resolved should be a hard error, never a silent degrade that still writes persisted pages.

## Proposed fixes

1. **Fail loud on unresolvable `--model`.** If the operator explicitly passed `--model X` and X can't be resolved to a working recipe/provider, throw `AIConfigError` and exit non-zero. Never fall back to no-LLM when the model was *explicitly requested*. (Implicit/default model fallback is fine; explicit override is a contract.)
2. **Never `--save` an empty synthesis.** Guard `persistSynthesis`: if the synthesis body is empty / the LLM didn't actually run (`Pages: N | Citations: 0` AND zero generated prose), refuse to persist and warn. Writing empty synthesis pages pollutes the brain and is worse than no page.
3. **Normalize slash→colon at every CLI entry point** so `--model anthropic/claude-sonnet-4-6` and `--model anthropic:claude-sonnet-4-6` are interchangeable everywhere, matching the documented `--model <alias or full id>` help text. Confirm `think` specifically routes through `parseModelId` post-v0.41.21.0.
4. **Surface the resolved model in output** even on success (`Model: anthropic:claude-sonnet-4-6 (resolved from 'anthropic/claude-sonnet-4-6')`) so silent degrades are caught immediately in logs.

[...]

PR #1901, opened 2026-06-05
test(conversation-parser): make LLM "no key" tests hermetic

ElliotDrel:
## Root cause

The `probeLlmAvailability` / `runLlmCall` / `runLlmFallback` "no key" tests cleared only `ANTHROPIC_API_KEY`, but `hasAnthropicKey()` also reads `loadConfig().anthropic_api_key` (from `~/.gbrain/config.json`, added in #1698 after these tests were written in v0.41.16.0). On any machine with a real config key (e.g. via `gbrain config set anthropic_api_key`) the probe still saw a key and 3 tests failed.

## Fix

Point `GBRAIN_HOME` at an empty temp dir inside the `withEnv` override so `configPath()` resolves to a dir with no `config.json` → `loadConfig()` returns null → both key sources are clean → probe returns null. Hermetic everywhere now.

## Verification

`conversation-parser` dir: 101 pass / 0 fail (was 98/3).

### DIFF
--- a/test/conversation-parser/llm-base.test.ts
+++ b/test/conversation-parser/llm-base.test.ts
@@ -12,6 +12,9 @@
  */
 
 import { describe, expect, test, beforeEach } from 'bun:test';
+import { mkdtempSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
 import { withEnv } from '../helpers/with-env.ts';
 import {
   runLlmCall,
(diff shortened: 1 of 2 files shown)


time-attack:
Thanks for this, @ElliotDrel! 🙏

What this does: This PR adds GBRAIN_HOME env overrides to "no key" tests so they neutralize both the env var and config file key sources, making tests hermetic on machines with real config keys. Master already has this exact fix implemented via a shared…

PR #1908, opened 2026-06-05
fix: coerce numeric YAML frontmatter title/type/slug to string

mgunnin:
## Problem

YAML parses an unquoted scalar like `title: 20260517_131712_214848` as a **number**, not a string. `parseMarkdown` casts it with `frontmatter.title as string`, which is a compile-time-only assertion and does **not** coerce at runtime. The numeric value flows through, and `assessContentSanity()` then calls `opts.title.toLowerCase()`, throwing:

```
```

In my brain this silently failed import for ~126 pages every sync — mostly chat/transcript exports whose session-id titles (e.g. `20260517_131712_214848`) parse as numbers.

## Fix

- **Root cause** (`src/core/markdown.ts`): coerce `title`, `type`, and `slug` to string at parse time instead of an inert `as string` cast.
- **Defense in depth** (`src/core/content-sanity.ts`): coerce `opts.title` before `.toLowerCase()` so the gate can never throw on a non-string title.
- **Tests** (`test/numeric-title-regression.test.ts`): cover both layers.

## Verification

- `bun test test/markdown.test.ts` → 33 pass
- `bun test test/import-file-content-sanity.test.ts test/numeric-title-regression.test.ts` → 25 pass
- Full `--no-embed` re-sync of a ~9.4k-page brain: numeric-title pages now import cleanly.

🤖 Generated with Claude Code


garrytan:

PR #1912, opened 2026-06-06
docs: clarify that .env is not read by systemd, use ~/.profile instead

UncleFruit:
When gbrain runs via systemd (autopilot --install, MCP serve),
.env files and .bashrc are not sourced. API keys placed in .env
per the current docs will be empty, causing silent auth failures.


time-attack:
Thanks for flagging the env-under-systemd confusion — it's a real source of silent auth failures. The specific advice doesn't hold for gbrain though: `autopilot --install` generates a wrapper script that sources `~/.zshenv` then `~/.zshrc`/`~/.bashrc` and never `~/.profile`, so keys placed in `~/.profile` would still be missing while `~/.bashrc` does get read. Closing since the guidance would point people the wrong way (the patch also drops a space in "services.Keys").

PR #1919, opened 2026-06-06
fix(markdown): add case hint for slug mismatches

### LINKED ISSUE #1916: Mixed-case slug: declared with mixed case fails SLUG_MISMATCH against lowercased default with no actionable error
## Symptom

A markdown file at path `projects/README.md` with frontmatter:

```yaml
slug: projects/README
```

is refused with `SLUG_MISMATCH` because gbrain lowercases path-derived slugs (the default would be `projects/readme`) and the declared slug doesn't match.

The error message says `SLUG_MISMATCH` but doesn't say *why* — capitalization isn't obvious to scan for, especially when both strings look identical to a quick eye.

## Reproduction

1. Create a file `projects/README.md` with frontmatter `slug: projects/README` (preserving case)
2. Run `gbrain sync` (or any path that resolves the slug)
3. Sync fails with `SLUG_MISMATCH`

## Workaround

Lowercase the declared slug, OR remove the `slug:` line entirely and let the default kick in.

## Proposed fix

- (a) Auto-normalize the declared slug to lowercase before the comparison, with a single info-level log line: `normalized declared slug "projects/README" → "projects/readme"`, OR
- (b) Replace the bare `SLUG_MISMATCH` with: `SLUG_CASE_MISMATCH: declared slug "projects/README" doesn't match lowercased path-derived default "projects/readme". Use lowercase in the slug: field, or omit the slug: line.`

## Environment

- gbrain 0.42.23.0
- First hit 2026-06-05 during repo cleanup (`README.md` vs `readme.md` collision on macOS APFS)

yaohao159123:
## Summary

- clarify `SLUG_MISMATCH` when the declared frontmatter slug only differs from the path-derived slug by capitalization
- tell the user to use the lowercase path-derived slug or remove `slug:` so the default is used
- add a regression test for `projects/README.md` declaring `slug: projects/README`

Closes #1916.

## Verification

- `bun test test/markdown-validation.test.ts`
- `bun test test/markdown.test.ts test/markdown-validation.test.ts`

### DIFF
--- a/src/core/markdown.ts
+++ b/src/core/markdown.ts
@@ -283,9 +283,12 @@ function collectValidationErrors(
   if (ctx.expectedSlug && typeof ctx.parsedFrontmatter.slug === 'string') {
     const declared = ctx.parsedFrontmatter.slug as string;
     if (declared !== ctx.expectedSlug) {
+      const caseHint = declared.toLowerCase() === ctx.expectedSlug
+        ? ` Slugs are case-sensitive; use "${ctx.expectedSlug}" or remove the slug field to use the path-derived value.`
+        : '';
       errors.push({
         code: 'SLUG_MISMATCH',
-        message: `Frontmatter slug "${declared}" does not match path-derived slug "${ctx.expectedSlug}"`,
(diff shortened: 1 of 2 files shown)


garrytan:

Issue #1849, opened 2026-06-04
jobs supervisor singleton is pidfile-path-keyed (HOME-relative default) → two supervisors run on the same queue with conflicting --max-rss


garrytan-agents:
## Summary

The `jobs supervisor` singleton guard is keyed **only on the `--pid-file` path**, and that path defaults to `${HOME}/.gbrain/supervisor.pid`. Because the default is `$HOME`-relative and there is **no DB/queue-level mutual exclusion**, two supervisors launched with different `HOME` (or different `--pid-file`) both pass the `O_CREAT|O_EXCL` EEXIST check and **run simultaneously against the same `default` queue** — with independent, possibly-conflicting `--max-rss` caps.

In production this manifested as a "ghost" supervisor at `--max-rss 4096` co-existing with the operator's intended `--max-rss 16384` tree on the same box. The 4GB cap's RSS watchdog kept killing legit `autopilot-cycle` jobs mid-run (**30 dead in 6h**, all `max stalled count exceeded` / `aborted: watchdog`), while every liveness/`doctor` check reported the queue "healthy" because *a* supervisor + worker were alive.

## Why the existing guard misses it

`MinionSupervisor.acquirePidLock()` → `tryAtomicCreate()` (`src/core/minions/supervisor.ts`) atomically creates `this.opts.pidFile`. That's correct mutual exclusion **for one pidfile path**. But:

```ts
export const DEFAULT_PID_FILE: string = (() => {
  const envOverride = process.env.GBRAIN_SUPERVISOR_PID_FILE;
  if (envOverride && envOverride.length > 0) return envOverride;
  const home = process.env.HOME ?? '/tmp';
  return `${home}/.gbrain/supervisor.pid`;
})();
```

So `HOME=/data` → `/data/.gbrain/supervisor.pid`, `HOME=/root` → `/root/.gbrain/supervisor.pid`. Two **different lock files → both acquire → two supervisors, same queue**. The guard's mutual-exclusion domain (a filesystem path) is narrower than the resource being protected (the `(database, queue)` pair). There is no `pg_advisory_lock` or `(queue)`-scoped row lock anywhere in the supervise path — confirmed by grep: the only "lock" on the queue side is per-job `lock_token`/`lock_until`, nothing supervisor-scoped.

## Repro

```bash
# same brain, same DATABASE_URL, same queue — different HOME
HOME=/data gbrain jobs supervisor --queue default --max-rss 16384 &
[...]

PR #1925, opened 2026-06-06
v0.42.27.0 fix(autopilot): harden runaway safeguards (#1678)

AkiGarage:
## Summary

Hardens GBrain autopilot against the incident class from #1678: duplicate cycles, stale/fresh lock confusion, and unbounded fan-out requeue pressure that can turn into CPU and memory runaway.

This keeps GBrain's useful background work alive. The fix deduplicates and bounds work per source instead of disabling autopilot, workers, or memory compounding.

Refs #1678
Related #1849

## Changes made

- Add per-source pending-cycle detection before autopilot fan-out dispatches new `autopilot-cycle` jobs.
- Add queue-level `liveSingletonKey` support so live nonterminal jobs are coalesced even when the hourly idempotency key changes.
- Make `autopilot.lock` PID-aware and fail closed:
  - fresh live owner blocks startup;
  - fresh invalid lock blocks startup instead of truncating unknown state;
  - stale locks can be recovered under a takeover lock;
  - cleanup and heartbeat only mutate the lock if the current process still owns it.
- Preserve fan-out throughput by ensuring already-pending sources do not consume fanout slots for other stale sources.
- Add regression tests for fan-out pending guards, source-less legacy blockers, lock takeover behavior, and queue singleton behavior.
- Ignore `.autoreview-reports/` generated by the local review hook.

## Tests/checks run

- `git diff --check origin/master...HEAD` - passed
- `bun run typecheck` - passed (`tsc --noEmit`)
- `bun test test/autopilot-fanout.test.ts test/autopilot-lock-path.test.ts test/autopilot-supervisor-wiring.test.ts test/minions-lease-full-retry.test.ts` - passed (`61 pass`, `0 fail`, `170 expect() calls`)
- `bun test test/e2e/autopilot-fanout-postgres.test.ts` - skipped locally (`0 pass`, `6 skip`, no `DATABASE_URL`)
- Autoreview final pass: clean, no accepted/actionable findings (`overall: patch is correct (0.82)`)
- Quality gate final focused pass: passed. Full omission/adversarial reports were retained separately; those phases produced false positives around existing secret-warning text and diff counters, so the final gate used the non-noisy test/review path after manual triage.

## Risks or unfinished items

- Postgres E2E still needs a real `DATABASE_URL` run in CI or a trusted local database to verify the SQL path end-to-end.
- This PR addresses the autopilot/job coalescing and lock-safety incident path. It does not fully resolve broader launchd/browser/computer-use policy questions from the local incident report.
- Because #1678 is broader than this patch, this PR references it instead of auto-closing it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,9 @@ export/
 # .context/test-shards/. Workspace-local by design — never committed.
 .context/
 
+# Local autoreview hook reports are generated during agent commits/reviews.
+.autoreview-reports/
+
 # Tier 3 PGLite snapshot fixture (built on demand by build:pglite-snapshot)
 test/fixtures/pglite-snapshot.tar
 test/fixtures/pglite-snapshot.version
--- a/src/commands/autopilot-fanout.ts
+++ b/src/commands/autopilot-fanout.ts
@@ -56,6 +56,7 @@ export interface FanoutResult {
   dispatched: string[];
   /** Source ids skipped because their last_full_cycle_at is still fresh. */
   skipped_fresh: string[];
+  skipped_pending: string[];
   /** Source ids beyond the fanoutMax cap (will retry next tick). */
   skipped_cap: string[];
   /** True when this tick fell back to the legacy single-job path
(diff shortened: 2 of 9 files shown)


garrytan:

PR #1930, opened 2026-06-07
Add Jina local embedding provider

Whamp:
## Summary

Adds a first-class `jina` embedding provider for local/OpenAI-compatible Jina retrieval embedding services.

- Registers `src/core/ai/recipes/jina.ts` with `jinaai/jina-embeddings-v5-omni-small` and `jina-embeddings-v5-text-small`.
- Supports 1024d default plus Jina v5 Matryoshka dimensions `[32,64,128,256,512,768,1024]`.
- Threads asymmetric retrieval side through `embed()` / `embedQuery()` so document chunks send `input_type=document` and query embeddings send `input_type=query`.
- Wires `JINA_BASE_URL` into the gateway config builder.
- Documents local/proxy usage and the stable embedding signature path for later proxy removal without re-embedding.

## Notes

The AI SDK openai-compatible embedding path forwards `dimensions` but drops arbitrary custom fields like `input_type`. This PR uses a reserved `user: gbrain-input-type:<side>` sentinel in `dimsProviderOptions()` and rewrites it in `jinaCompatFetch` immediately before fetch, removing `user` from the outbound body.

Default base URL is local (`http://localhost:8081/v1`) so no-auth availability behaves like other local recipes. Hosted Jina users can set `JINA_BASE_URL=https://api.jina.ai/v1` and `JINA_API_KEY`.

## Validation

- `bun test test/ai/recipe-jina.test.ts test/ai/build-gateway-config.test.ts test/ai/recipes-existing-regression.test.ts test/ai/gateway.test.ts`
- `bun run typecheck`
- `bun run build:llms`
- `bash scripts/check-key-files-current-state.sh`
- `bun test test/build-llms.test.ts`
- Live local smoke through sidecar proxy:
  - `embed()` returned 1024d document vector
  - `embedQuery()` returned 1024d query vector


time-attack:
Thanks for taking the time to wire this up cleanly. We keep the embedding-provider set deliberately narrow — each additional vendor backend is permanent maintenance and test surface we have to carry — so we're not accepting new third-party provider integrations. Closing on that policy basis rather than anything about the code quality.

PR #1931, opened 2026-06-07
bug fix: multiple proposals for a page are silently dropped after extraction

chasezheng:
When a page produces multiple proposals, only the first is saved to take_proposals — the rest are silently dropped.

The idempotency index imposed a uniqueness constraint that doesn't allow multiple proposals per page. Make it non-unique (migration 114 + fresh-install schema) and drop the INSERT's ON CONFLICT clause.

### DIFF
--- a/src/core/cycle/propose-takes.ts
+++ b/src/core/cycle/propose-takes.ts
@@ -6,7 +6,7 @@
  * `take_proposals` queue. User accepts/rejects via `gbrain takes propose`.
  *
  * Idempotency contract (D17 schema spec):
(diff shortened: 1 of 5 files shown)


time-attack:
Thanks for catching this and writing it up clearly — the symptom you described was real. It has since been fixed on master via migration v125 (`take_proposals_per_claim_idempotency`), which folds `md5(claim_text)` into the idempotency key so every claim on a multi-claim page is kept while re-runs stay idempotent. Your approach of making the index non-unique and removing `ON CONFLICT` would keep all claims but lose that re-run protection, so I'm closing in favor of the shipped fix.

Issue #1720, opened 2026-06-01
autopilot worker crash-loops on pooler CONNECTION_CLOSED: background queue loops have no in-process reconnect


mdcruz88:
**Version:** gbrain 0.41.11.1 (Postgres engine, `gbrain autopilot` continuous daemon)

## TL;DR

When the Postgres connection drops under a long-running `gbrain autopilot` worker (e.g. a Supabase Supavisor session-pooler closing the socket), the autopilot-managed child worker (`gbrain jobs work`) never re-establishes the connection in-process. Its background interval loops (`promoteDelayed`, stall-detection) and the failure-recording path (`failJob`) just throw `No database connection` repeatedly, the original job error gets masked, and the worker exits `code=1` and respawns. On a brain that sees periodic pooler closes this becomes a crash-loop.

This is **not** the RSS watchdog — no `[watchdog rss=…]` events occur. The initiating failure is a socket/pooler close.

## Observed (sanitized)

Continuous `gbrain autopilot` daemon, child worker `gbrain jobs work --max-rss 2048`, Supabase Supavisor **session pooler** (`<project>.pooler.supabase.com:5432`).

```
[extract.links_fs] connection blip, retrying 7 rows in 500ms (No database connection: connect() has not been called…)
Promotion error: write CONNECTION_CLOSED <project>.pooler.supabase.com:5432
Stall detection error: write CONNECTION_CLOSED <project>.pooler.supabase.com:5432
  batch error (7 link rows lost): No database connection: connect() has not been called…
```

Over the daemon's lifetime: `No database connection` ×6,350, `Promotion error` ×20,064 (background interval loop spamming the dead pool), `worker exited code=1` ×547, `crashCount` reaching 5/5. The errors spray across *every* cycle phase (lint, sync, extract, embed…), confirming a global pool-state failure, not a single bad phase.

## Root cause

1. The DB connection is a single mutable module-level pool singleton: `let sql` (`src/core/db.ts:7`). `getConnection()` throws `No database connection` when it's null (`db.ts:151`).
2. A raw `CONNECTION_CLOSED` does **not** null the singleton — the object stays non-null but its underlying socket is dead. And `connect()` is a **blind no-op when `sql` is already set** (`db.ts:162`), so nothing repairs a dead-but-non-null pool.
3. The autopilot child is plain `gbrain jobs work` — **not** `MinionSupervisor` — so it lacks the reconnect health logic in `supervisor.ts:589`. Its background paths only log:
[...]

PR #1935, opened 2026-06-07
fix(autopilot): route DB reconnect through engine.reconnect(); stop crash-loop on pooler drop (#1720)

mdcruz88:
## Problem (#1720)

The autopilot daemon crash-loops against the Supabase pooler. On a DB-health-probe failure, the catch block hand-rolls a reconnect:

```ts
await engine.disconnect();
await (engine as any).connect?.();   // <- called with NO config argument
```

`PostgresEngine.connect(config)` immediately does `const url = config.database_url`, so the argless call throws `undefined is not an object (evaluating 'config.database_url')`. `classifyReconnectError` then matches `database_url` + `undefined` and returns `unrecoverable` → the daemon logs FATAL and exits → launchd `ThrottleInterval` respawns it → the pooler drops the connection again → same crash. Observed: ~25k `CONNECTION_CLOSED` events and a brain degraded for days.

Root cause: `PostgresEngine` already exposes a correct `reconnect()` (reuses its saved config, rebuilds the instance pool, logs pool-recovery audits), but the autopilot catch **predates** it (catch from 2026-04-14; `reconnect()` added later) and never adopted it.

## Fix

- **New `reconnectAndProbe()` helper** routes through `engine.reconnect()` when present (PostgresEngine), and **rethrows the original probe error** for engines without it (the `BrainEngine` interface / `PGLiteEngine`) instead of crashing on an argless `connect()`. It then **re-probes health** (`getConfig`) before the caller clears its failure counter, so a silent no-op reconnect (e.g. a concurrent reconnect already in flight) can't falsely reset it.
- **`classifyReconnectError` hardened**: a raw JS runtime TypeError (`is not an object`, `is not a function`, `is not a constructor`, `cannot read prop`) is now classified **recoverable** — a code bug must not masquerade as fatal misconfiguration and crash-loop the daemon; `AUTOPILOT_MAX_RECONNECT_FAILS` still bounds retries. `undefined` is removed from the unrecoverable `database_url` branch (its only producer was the TypeError above); genuine `missing`/`empty` config stays fatal.

## Tests

- Classifier regression: the real TypeError + variants + bare text → recoverable; genuine missing/empty `database_url` stays unrecoverable.
- New `test/autopilot-reconnect-helper.test.ts`: reconnect→reprobe ordering, the `{ error }` arg passthrough, the never-argless-connect guarantee, the PGLite rethrow path, and the no-op-reconnect → reprobe-throws path.
- Full autopilot suite (107 tests) green; `tsc --noEmit` clean.

### DIFF
--- a/src/commands/autopilot.ts
+++ b/src/commands/autopilot.ts
@@ -54,7 +54,26 @@ import { inspectLock } from '../core/db-lock.ts';
  */
 export function classifyReconnectError(err: unknown): 'recoverable' | 'unrecoverable' {
   const msg = (err instanceof Error ? err.message : String(err ?? '')).toLowerCase();
-  if (msg.includes('database_url') && (msg.includes('undefined') || msg.includes('missing') || msg.includes('empty') || msg.includes('not set'))) {
+  // #1720: a raw JS runtime TypeError (e.g. "undefined is not an object
+  // (evaluating 'config.database_url')", "x is not a function", "cannot read
+  // properties of undefined") is a PROGRAMMING error in the reconnect path, not
+  // a genuine user misconfiguration. Treat it as recoverable: a transient blip
+  // that tripped a code bug must not crash-loop the daemon forever, and the
+  // AUTOPILOT_MAX_RECONNECT_FAILS cap still bounds retries. This check runs
+  // FIRST because such a message can also contain "database_url".
+  if (
+    msg.includes('is not an object') ||
+    msg.includes('is not a function') ||
+    msg.includes('is not a constructor') ||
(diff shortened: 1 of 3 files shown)


mdcruz88:
Thanks @garrytan, and no worries on the triage timing. Closing this as superseded: #1805 (module-singleton ownership) and #1810 (the reliability wave's Postgres reconnect-race fix) cover this crash-loop class on master. I also take the point recorded in COMMUNITY_IDEAS.md that a disconnect-before-connect reconnect() is the wrong lever here. Appreciate the canonical fix landing 🙏

PR #1936, opened 2026-06-07
fix(postgres-engine): tear down half-open client when instance connect() fails (#1720 follow-up)

mdcruz88:
## Problem

Instance-path `PostgresEngine.connect()` (called with `poolSize`, e.g. `gbrain jobs work` / autopilot worker engines) assigns the client before it probes:

```ts
this._sql = postgres(url, opts);
await this._sql`SELECT 1`;   // <- if this throws, _sql is already set
```

If the probe throws — e.g. `EMAXCONNSESSION` when the Supabase **session-mode pooler is at its client cap**, or `ECONNREFUSED` mid-reconnect — the method propagates the error but leaves `_sql` holding a **half-open postgres.js client**. That client occupies a pooler client slot until the *next* `reconnect()`'s `disconnect()` tears it down. Under a tight session-mode client cap, that transient retention compounds connection pressure.

Surfaced in adversarial review during live verification of the #1720 autopilot crash-loop fix (#1935): the daemon, now staying alive and retrying reconnects (instead of crash-exiting and freeing everything), exercises this path repeatedly during a pooler-saturation window.

## Fix (instance path only)

- Wrap the instance rebuild in `try/catch`. On failure: best-effort `connectionManager?.disconnect()` + null it, then `_sql.end({ timeout: 5 })` + null it, before rethrowing. A failed connect now leaves the engine holding **zero** clients.
- Mark `_connectionStyle = 'instance'` **before** the awaited probe. `reconnect()` branches on `_connectionStyle !== 'instance'`; if a `poolSize`-configured engine's *first* connect failed with the marker still `null`, the next `reconnect()` would silently divert into module-singleton recovery. The `sql` getter's instance self-heal (#1678) already throws a retryable error for instance-style + null `_sql`, so this state is handled.

## Scope / known follow-up

**Instance path only.** The module-singleton path (`db.connect`, `src/core/db.ts:222-231`) has the **same** construct-then-probe shape, with a catch that nulls `sql` *without* `await s.end()`. That is a **separate, named follow-up** — this PR does **not** address it and does **not** claim the module path is safe.

## Tests

`test/postgres-engine-connect-cleanup.test.ts` drives an instance `connect()` to a refused local port; asserts `_sql` and `connectionManager` are null and `_connectionStyle === 'instance'` after the throw (all RED pre-fix). 66 engine/connection/reconnect tests pass; `tsc --noEmit` clean.

### DIFF
--- a/src/core/postgres-engine.ts
+++ b/src/core/postgres-engine.ts
@@ -192,20 +192,43 @@ export class PostgresEngine implements BrainEngine {
       if (typeof prepare === 'boolean') {
         opts.prepare = prepare;
       }
-      this._sql = postgres(url, opts);
-      await this._sql`SELECT 1`;
-      await db.setSessionDefaults(this._sql);
-      this._connectionStyle = 'instance';
+      try {
+        // Mark instance intent BEFORE the awaited probe. reconnect() branches on
+        // `_connectionStyle !== 'instance'`; if the first connect's probe throws
+        // and this were still null, the next reconnect() would silently divert a
+        // poolSize-configured engine into module-singleton recovery (#1720
+        // follow-up, Data R1).
+        this._connectionStyle = 'instance';
+        this._sql = postgres(url, opts);
+        await this._sql`SELECT 1`;
+        await db.setSessionDefaults(this._sql);
 
-      // v0.30.1: instance-owned ConnectionManager wraps the read pool we just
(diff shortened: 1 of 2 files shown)


mdcruz88:
Closing as superseded by the master reliability work (#1805 / #1810) — the half-open client teardown is covered by the canonical module-singleton + reconnect-race fixes. Thanks for the thorough backlog pass 🙏

PR #1937, opened 2026-06-07
fix(db): validate module singleton before publish — leak + connect/connect + disconnect/connect races (#1720 follow-up)

mdcruz88:
## Problem

`db.connect()` (the module-singleton create path) builds the client, publishes it to the module `sql`, and *then* probes it:

```ts
sql = postgres(url, opts);
await sql`SELECT 1`;   // <- publish happens BEFORE validation
```

Three defects fall out of that ordering — all surfaced in adversarial review of the #1720 autopilot reconnect fix:

1. **Leak.** The catch did `sql = null` **without** `await sql.end()`. A thrown probe (`EMAXCONNSESSION` at the session-mode pooler client cap, `ECONNREFUSED`, …) orphans a postgres.js client that holds a pooler client slot, with no reference left to close it.
2. **Join-the-dead-client race.** Because the unvalidated `sql` was published before the probe, a concurrent `connect()` arriving mid-probe sees `sql` truthy and returns `false` (borrower) against a client that may be about to fail — and records ownership/manager state for an invalid singleton.
3. **Disconnect-vs-in-flight race.** A `disconnect()` racing an in-flight `connect()` sees `sql` still null (probe pending), tears nothing down, and is then silently undone when the probe later publishes the singleton.

## Fix — validate before publish

- **`connecting`** — the in-flight create's probe promise. `connect()` decides atomically across **three** states (live `sql` / in-flight `connecting` / absent): the checks and the synchronous `connecting = (...)()` assignment run with no `await` between them, so two concurrent callers can't both become creator. `sql` is published **only after** the `SELECT 1` probe validates the client; latecomers `await connecting` instead of the unvalidated `sql`. On failure the client is ended and every waiter throws; on success exactly one caller gets `true`, waiters get `false`. (#1471 create-vs-join ownership preserved.)
- **`connectEpoch`** — bumped by `disconnect()` before its snapshot. The creator captures the epoch before its probe and refuses to publish if it changed, ending the client and rejecting — so a completed `disconnect()` can't be resurrected.

## Scope

Module-singleton **create / borrow / failed-probe / disconnect** lifecycle. The separately-tracked `PostgresEngine.reconnect()` module-mode *"wedged-but-non-null pool"* rebuild is **not** addressed here. (The instance-path `connect()` got the analogous end-on-failed-probe cleanup in the sibling PR.)

## Tests

`test/db-connect-singleton.test.ts` — deterministic via a gated mock probe (`mock.module('postgres')`): failed-probe ends the client + nulls the singleton; single success; concurrent success (one creator / one borrower, single client); concurrent failure (both throw, single client, ended once); `disconnect()`-during-in-flight aborts the publish; retry-after-failure creates a fresh singleton. 6 tests green; 54+ db-lock / connection / ownership regression tests pass; `tsc --noEmit` clean.

### DIFF
--- a/src/core/db.ts
+++ b/src/core/db.ts
@@ -6,6 +6,18 @@ import { verifySchema } from './schema-verify.ts';
 
 let sql: ReturnType<typeof postgres> | null = null;
 let connectedUrl: string | null = null;
+// #1720: the in-flight create's probe promise. Concurrent connect() callers
+// await THIS — not the not-yet-validated `sql` — so a latecomer never joins a
+// client whose SELECT 1 probe is still pending or about to fail. Null when no
+// create is in flight; resolves once `sql` is published (validated); rejects
+// (with the GBrainError) when the probe fails.
+let connecting: Promise<void> | null = null;
+// #1720: bumped by disconnect() (and by a new connect's teardown) to invalidate
+// an in-flight create. The creator captures this epoch before its probe and, if
+// it changed by the time the probe resolves, aborts WITHOUT publishing `sql` —
+// otherwise a disconnect() that raced the probe (saw `sql` still null, closed
+// nothing) would be silently undone when the probe later publishes a singleton.
+let connectEpoch = 0;
 
 /**
  * Default pool size for Postgres connections. Users on the Supabase transaction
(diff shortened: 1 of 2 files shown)


garrytan:

PR #1944, opened 2026-06-07
fix(importer): coerce non-string frontmatter title before toLowerCase (#1939)

### LINKED ISSUE #1939: Sync bookmark permanently stuck when importer throws on non-string frontmatter `title` (content-sanity.ts:379) — silent indexing outage
## Summary

The `default` source sync bookmark can get **permanently stuck** when a small number of files crash the importer. A stuck bookmark forces every subsequent `gbrain sync` to restart the full repo walk (200K+ files) and never reach HEAD, so new commits silently stop being indexed. On our brain the `default` source (`/data/brain`) was stuck at **2026-06-04** for 3+ days while commits kept landing; `gbrain get <new-page-slug>` returned `page_not_found` even though the page was committed and pushed.

## Root cause

Two interacting issues:

### 1. Importer throws on non-string `title` (primary)

`src/core/content-sanity.ts:379`:

```ts
const titleLower = opts.title.toLowerCase();
```

`opts.title` is assumed to be a string, but YAML frontmatter like:

```yaml
title: 2024-06-01
```

parses `title` as a **Date** (or a number, e.g. `title: 1458`), so `opts.title.toLowerCase` is `undefined` and the import throws:

```
opts.title.toLowerCase is not a function. (In 'opts.title.toLowerCase()', 'opts.title.toLowerCase' is undefined)
```

This reliably fires on apple-notes-style pages whose filename/title is a bare date or number. Repro files in our brain include `sources/apple-notes/Politics/2024-06-01 8189165238.md`, `sources/apple-notes/YC/Talks YC/2023-04-25 1458.md`.

### 2. A handful of throwing files block bookmark advancement

Because these files throw (rather than being recorded as skippable failures and advanced past), the incremental bookmark never moves past them, so each run re-walks the entire tree from the stuck point. `sync --skip-failed` is the documented escape hatch, but it only helps if something actually runs it to completion — a normal `sync` never drains the backlog on its own.

## Fix

1. **Coerce title to string before `.toLowerCase()`** in `content-sanity.ts` (and audit other `opts.title`/frontmatter string-method call sites — e.g. `frontmatter-inference.ts:298`). Something like `String(opts.title ?? '')`. A non-string `title` should degrade gracefully (or be normalized at frontmatter-parse time so a YAML date/number title becomes its source string), never throw and wedge the importer.
2. **Don't let a throwing file freeze the bookmark.** A parse/throw on one file should be recorded as a failure and the bookmark allowed to advance past it (the `--skip-failed` semantics) so a poison file can't silently stop all indexing.

## Suggested guardrail (doctor)

`gbrain doctor` should add a **sync-freshness / stuck-bookmark check**: FAIL/WARN when a source's `last_sync_at` is older than ~2× the sync cadence while the repo HEAD has advanced (i.e. unindexed commits exist). Today `sources status` exposes `last_sync_at` but nothing alerts when it goes stale, so a stuck bookmark is invisible until someone notices a missing page. This is the signal that would have caught the 3-day stall immediately.

## Impact

High: silent indexing outage. Commits look successful, GitHub shows the page, but the brain never indexes it and search/`get` return nothing. No surfaced error.

## Environment

- gbrain ~v0.42.26.0
- PgBouncer transaction-mode pooler (port 6543), prepared statements disabled
- Multi-source brain (default + 3 federated sources)

garrytan-agents:
## Summary

Fixes #1939 (and the convergence half of #1794): the gbrain importer crashes on any page whose frontmatter `title` parses as a non-string, silently skipping the page and **wedging the source sync bookmark** — which forces a full re-walk of the whole repo that never converges, so new commits stop being indexed.

## Root cause

`assessContentSanity` (src/core/content-sanity.ts:379) is typed `title: string`, but the importer hands it the raw YAML-parsed frontmatter value:

```yaml
title: 404          # -> number
title: 2024-06-01   # -> Date
```

`opts.title.toLowerCase()` then throws `opts.title.toLowerCase is not a function`, the importer skips the file, and per the #1939 mechanism a thrown importer error leaves the `default` source bookmark stuck.

## Live evidence (garrytan/brain)

- `default` source bookmark stuck at **2026-06-04** for 3+ days while commits kept landing.
- A 2.5h sync sat at **9% (18.5K/204K files)** re-walking from scratch, never converging (#1794), burning ~27% CPU continuously.
- **21 real poison files** on the brain (titles like `404`, `40`, `222`, `529`, numeric apple-notes). Drain log: `opts.title.toLowerCase is not a function ... (suppressing further ...)`.
- Data-side already fixed on our brain by quoting those 21 titles; this PR is the durable code fix so it can't recur from a single bad title.

## Fix

```ts
const titleLower = String(opts.title ?? '').toLowerCase();
```

One line, defensive at the use site. Added 3 regression tests (numeric / Date / null titles no longer throw). Full suite: **80 pass / 0 fail**.

## Note on #1794

This removes the *trigger* (a single poison file should not wedge the bookmark), but the underlying "full sync never checkpoints, killed = lose all progress" invariant in #1794 is still worth a separate fix (intermediate `last_commit` checkpointing). This PR makes the common case stop firing.

### DIFF
--- a/src/core/content-sanity.ts
+++ b/src/core/content-sanity.ts
@@ -376,7 +376,13 @@ export function assessContentSanity(opts: {
   // doesn't repeat the lowercase per literal.
   const bodyHead = body.slice(0, SCAN_HEAD_BYTES);
   const bodyHeadLower = bodyHead.toLowerCase();
-  const titleLower = opts.title.toLowerCase();
+  // Defensive: `opts.title` is typed `string`, but the importer feeds the raw
+  // YAML-parsed frontmatter value. `title: 404` parses as a number and
+  // `title: 2024-06-01` as a Date, so a bare `.toLowerCase()` throws
+  // ("opts.title.toLowerCase is not a function") and the importer skips the
+  // page. A thrown importer error wedges the source bookmark (#1939), forcing a
+  // full re-walk that never converges (#1794). Coerce to string first.
+  const titleLower = String(opts.title ?? '').toLowerCase();
 
   const junk_pattern_matches: string[] = [];
   for (const p of BUILT_IN_JUNK_PATTERNS) {
--- a/test/content-sanity.test.ts
+++ b/test/content-sanity.test.ts
@@ -640,3 +640,34 @@ describe('assessContentSanity — confidence split (Q1=A)', () => {
     expect(r.shouldFlag).toBe(false);
   });
 });
+
+// ─── REGRESSION: non-string title (#1939) ─────────────────────
+// The importer feeds the raw YAML-parsed frontmatter value. `title: 404`
+// parses as a number and `title: 2024-06-01` as a Date, so a bare
+// `opts.title.toLowerCase()` threw and skipped the page — wedging the source
+// bookmark (#1939) and forcing a never-converging full re-walk (#1794).
+describe('assessContentSanity — non-string title coercion (#1939)', () => {
+  test('numeric title does not throw', () => {
+    expect(() =>
+      // @ts-expect-error — simulate importer passing a YAML number
+      assessContentSanity({ compiled_truth: 'hello world', timeline: '', title: 404 }),
+    ).not.toThrow();
+  });
+
+  test('Date title does not throw', () => {
+    expect(() =>
+      // @ts-expect-error — simulate importer passing a YAML Date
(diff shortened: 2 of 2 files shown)


garrytan:

PR #1947, opened 2026-06-07
fix(embed): keep --catch-up running instead of aborting on 32-bit setTimeout overflow

### LINKED ISSUE #1946: embed --catch-up aborts after one batch (32-bit setTimeout overflow)
## Summary
`gbrain embed --stale --catch-up` is meant to run until `countStaleChunks()` returns 0. Instead it aborts almost immediately after the first batch, so large stale backlogs never get embedded.

## Cause
In `embedAllStale` (`src/commands/embed.ts`), the `--catch-up` path sets the wall-clock budget to an effectively-unbounded value and passes it directly to `setTimeout` to arm the budget `AbortController`:

```ts
const BUDGET_MS = staleOpts?.catchUp ? Number.MAX_SAFE_INTEGER : /* 30 min default */;
const budgetTimer = setTimeout(() => budgetController.abort(), BUDGET_MS);
```

JS timer delays are stored as signed 32-bit milliseconds (max 2,147,483,647 ms ≈ 24.8 days). `Number.MAX_SAFE_INTEGER` overflows that; Node/Bun clamp it and the timer fires almost immediately, so the controller aborts the run after the first batch.

## Impact
On a brain with a large stale backlog, `--catch-up` never catches up — it silently stops and the backlog persists across runs.

## Expected
`--catch-up` runs until there are no stale chunks left.

## Fix
Arm the deadline via a helper that never passes an out-of-range delay to `setTimeout` (re-arm within the 32-bit cap). PR incoming.

100menotu001:
## Problem
`embed --stale --catch-up` aborts after one batch instead of running until the stale backlog is cleared. The unbounded catch-up budget is passed straight to `setTimeout`, which clamps out-of-range (> 2³¹−1 ms) delays and fires the budget `AbortController` immediately. Details in #1946.

## Fix
- Add `scheduleAbortAfterDelay(controller, ms)`: arms the abort deadline and **re-arms within the 32-bit timer cap** so long/unbounded budgets behave correctly (aborts immediately for `ms <= 0`, no-ops for non-finite).
- Use it in `embedAllStale` in place of the single `setTimeout`.
- Regression coverage in `test/embed.serial.test.ts`.

No behavior change for the default bounded (30 min) budget; the fix only affects the unbounded/long-budget path.

Closes #1946

### DIFF
--- a/src/commands/embed.ts
+++ b/src/commands/embed.ts
@@ -73,6 +73,51 @@ export interface EmbedOpts {
   signal?: AbortSignal;
 }
 
+export const MAX_JS_TIMER_DELAY_MS = 2_147_483_647;
+
+export interface AbortTimerHandle {
+  clear(): void;
+}
+
+/**
+ * Schedule an AbortController deadline without ever passing an out-of-range
+ * delay to setTimeout. JS runtimes store timer delays as signed 32-bit
+ * milliseconds; larger values are clamped by Node/Bun and can fire at once.
+ *
+ * @internal exported for targeted regression tests.
+ */
+export function scheduleAbortAfterDelay(controller: AbortController, ms: number): AbortTimerHandle | undefined {
+  if (!Number.isFinite(ms)) return undefined;
+  if (ms <= 0) {
+    controller.abort();
+    return undefined;
+  }
+
+  let timer: ReturnType<typeof setTimeout> | undefined;
+  let cancelled = false;
+  const startedAt = Date.now();
+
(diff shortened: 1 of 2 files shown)


time-attack:
Thanks for this, @100menotu001! 🙏

What this does: This PR fixes a bug where `gbrain embed --stale --catch-up` aborts after one batch because the unbounded budget overflows JavaScript's 32-bit setTimeout limit. The fix implements a helper function that chunks setTimeout calls to handle…

PR #1949, opened 2026-06-07
fix(markdown): coerce non-string frontmatter title/type/slug instead of mis-typing

### LINKED ISSUE #1948: Non-string frontmatter title/type/slug mis-types pages on import
## Summary
`parseMarkdown` (`src/core/markdown.ts`) and `importFromContent` (`src/core/import-file.ts`) assume frontmatter `title`, `type`, and `slug` are strings and cast them with `as string`. When a page carries a non-string frontmatter value — e.g. a numeric `title` from a structured JSON import — the unchecked cast lets the wrong type flow into page typing and the content-sanity title check.

## Expected
Non-string frontmatter values are coerced to a string (or ignored in favor of the path-derived inference) rather than propagated as the wrong type.

## Fix
Defensive coercion in both functions. PR incoming.

100menotu001:
## Problem
`parseMarkdown` / `importFromContent` cast frontmatter `title`/`type`/`slug` with `as string`. A non-string value (e.g. a numeric title from a structured JSON import) flows through as the wrong type into page typing and the content-sanity title check. Details in #1948.

## Fix
- `markdown.ts`: use the frontmatter value only when it is a string; otherwise fall back to the path-derived `inferType` / `inferTitle` / `inferSlug`.
- `import-file.ts`: coerce the title passed to `assessContentSanity` to a string.

Small defensive type-safety change; no behavior change when frontmatter values are already strings.

Closes #1948

### DIFF
--- a/src/core/import-file.ts
+++ b/src/core/import-file.ts
@@ -400,10 +400,16 @@ export async function importFromContent(
     // default, or reject (throw → sync-failure) when the operator opts in.
     const junkDisposition: 'quarantine' | 'reject' =
       cs.junk_disposition === 'reject' ? 'reject' : 'quarantine';
+    const sanityTitle =
+      typeof parsed.title === 'string'
+        ? parsed.title
+        : parsed.title == null
+          ? ''
+          : String(parsed.title);
     const sanityResult = assessContentSanity({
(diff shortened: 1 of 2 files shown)


garrytan:

PR #1952, opened 2026-06-07
fix(pglite): populate chunk FTS vectors on upsert

heyimcarlos:
## Summary
- populate PGLite content_chunks.search_vector directly after upsertChunks
- add regression that drops the chunk FTS trigger and verifies fresh chunks still search

## Tests
- bun test test/pglite-engine.test.ts --timeout 20000 -t "upsertChunks remains keyword-searchable"
- bun test test/pglite-engine.test.ts --timeout 30000

### DIFF
--- a/src/core/pglite-engine.ts
+++ b/src/core/pglite-engine.ts
@@ -1952,6 +1952,22 @@ export class PGLiteEngine implements BrainEngine {
          embedding_image = COALESCE(EXCLUDED.embedding_image, content_chunks.embedding_image)`,
       params
     );
+
+    // Defense in depth for PGLite fresh installs whose chunk FTS trigger is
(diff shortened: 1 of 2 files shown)


time-attack:
Thanks for the careful patch and the regression test. I traced this on master: migration v27 installs `chunk_search_vector_trigger`, and PGLite runs the full migration array after the schema blob, so the vector is populated on fresh installs — your test only fails because it drops the trigger first. That makes this an extra unconditional UPDATE of every chunk on every upsert on the hot ingest path to guard a state the schema already guarantees, so I'm closing it; if you have a reproducible install where the trigger really is missing, that would be a schema-bootstrap bug worth reporting on its own.

PR #1953, opened 2026-06-07
fix(ai): litellm proxy unusable for chat — add chat touchpoint + fold proxy api keys

miroslavb:
## Problem

The `litellm-proxy` recipe declares only an `embedding` touchpoint — no `chat`.
So any `chat_model` / `models.default = litellm:<model>` fails `validateModelId`
with *"Provider \"litellm\" does not support touchpoint \"chat\""*, and the think
synthesis path then degrades silently to **`(no LLM available — set ANTHROPIC_API_KEY)`**
(misleading: the user configured litellm, not anthropic). This makes LiteLLM —
whose primary purpose is chat — unusable for chat/synthesis/subagent in gbrain.

A second gap: `build-gateway-config.ts` folds `openai_api_key` / `anthropic_api_key` /
`zeroentropy_api_key` from config into the gateway env, but **not** the openai-compat
proxy keys. So `gbrain config set litellm_api_key …` writes the DB plane and never
reaches the proxy call → it goes out unauthenticated. (Same class as the v0.37
ZeroEntropy fold fix.)

## Fix

1. **`recipes/litellm-proxy.ts`** — add a `chat` touchpoint (openai-compatible,
   `user_provided_models: true`, `supports_tools: true`) mirroring the embedding one.
   The proxy serves arbitrary models; the openai-compat tier doesn't enforce the
   list at runtime.
2. **`build-gateway-config.ts`** — fold `litellm_api_key` / `openrouter_api_key` /
   `together_api_key` from config into the gateway env (`LITELLM_API_KEY`, …), the
   same way openai/anthropic/zeroentropy are folded. Their auth resolves to
   `${RECIPE_ID}_API_KEY` via `applyResolveAuth`.

## Verification (against a live authenticated LiteLLM proxy)

- `probeChatModel('litellm:qwen-3.5-plus')` → `{ ok: true }` (was `unknown_model`).
- `gbrain think "…"` synthesizes a real answer through the proxy (was "no LLM available").
- The proxy model returns OpenAI-style `tool_calls`, and with
  `agent.use_gateway_loop=true` a subagent job completes through `gateway.toolLoop()`
  on the litellm model.

No behavior change for users who don't use litellm/openrouter/together.
Happy to add a recipe-shape test pinning the new chat touchpoint if desired.

### DIFF
--- a/src/core/ai/build-gateway-config.ts
+++ b/src/core/ai/build-gateway-config.ts
@@ -34,6 +34,15 @@ export function buildGatewayConfig(c: GBrainConfig): AIGatewayConfig {
   // plane field now exists (GBrainConfig type) and gets mapped here, so
   // setting it via `~/.gbrain/config.json` propagates into the gateway.
   if (c.zeroentropy_api_key) envFromConfig.ZEROENTROPY_API_KEY = c.zeroentropy_api_key;
+  // Same fold for the openai-compat proxy recipes (litellm / openrouter /
+  // together): their auth resolves to `${RECIPE_ID}_API_KEY` via applyResolveAuth,
+  // so a `gbrain config set litellm_api_key …` (or a config.json field) must be
+  // mapped into the gateway env or the proxy call goes out unauthenticated and
+  // chat_model=litellm:<m> degrades to "no LLM available". Mirrors the ZE fix.
+  const cAny = c as Record<string, unknown>;
+  if (typeof cAny.litellm_api_key === 'string') envFromConfig.LITELLM_API_KEY = cAny.litellm_api_key;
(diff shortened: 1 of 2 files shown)


time-attack:
Thank you for the detailed report and for verifying against a live proxy — that write-up was genuinely useful. The main gap has since been closed on master: the litellm recipe now declares a `chat` touchpoint (plus `expansion`), and the OpenRouter key fold in `build-gateway-config.ts` also landed. What remains from your patch is the litellm/together key folding, which here goes through an untyped `as Record<string, unknown>` cast for fields not declared on `GBrainConfig`, so I'm closing this rather than merging around the conflict.

PR #1962, opened 2026-06-08
fix(embed): don't arm wall-clock timer in --catch-up (#1946)

### LINKED ISSUE #1946: embed --catch-up aborts after one batch (32-bit setTimeout overflow)

TreyLawrence:
## Summary

Closes #1946.

`embedAllStale` armed the budget timer with `setTimeout(..., Number.MAX_SAFE_INTEGER)` when `--catch-up` was set. JS timer delays are clamped to a signed 32-bit ms range (~24.8 days); any out-of-range delay fires on the next tick. So the `AbortController` aborted after the first batch, and large stale backlogs silently never caught up across runs.

## Fix

Skip arming the timer entirely in catch-up mode:

```ts
const BUDGET_MS = staleOpts?.catchUp
  ? null
  : parseInt(process.env.GBRAIN_EMBED_TIME_BUDGET_MS || `${30 * 60 * 1000}`, 10);
const budgetController = new AbortController();
const budgetTimer = BUDGET_MS == null
  ? null
  : setTimeout(() => budgetController.abort(), BUDGET_MS);
```

The `AbortController` is still constructed and still composed via `anySignal(budgetSignal, externalSignal)`, so SIGINT / worker-timeout / lock-loss propagation is unchanged — only the internal wall-clock timer is absent. `budgetSignal.aborted` stays `false` throughout catch-up, so the existing budget-exceeded log branch is unreachable in that mode (correct: catch-up has no budget to exceed).

Alternative shapes considered:
- **Clamp to `2**31 - 1` (~24.8 days)** — works in practice but isn't semantically "until `countStaleChunks()` returns 0", and a backlog that legitimately exceeds 24 days would re-introduce the silent abort.
- **Re-arm pattern** — adds complexity for no behavior gain over "just don't arm it."

## Test plan

- [x] `bun test test/embed.serial.test.ts` — 29/29 pass, including the existing `GBRAIN_EMBED_TIME_BUDGET_MS=N` budget test (unchanged behavior on the non-catch-up path)
- [x] New regression test: `--catch-up does NOT abort after the first batch (32-bit setTimeout overflow)` — asserts all 5 batches process through to completion under catch-up

Diff: 2 files, +58 / -7. `src/commands/embed.ts` + `test/embed.serial.test.ts` only. No VERSION / CHANGELOG / package.json changes — per the fix-wave workflow in `docs/RELEASING.md`, those are the wave shipper's concern.

### DIFF
--- a/src/commands/embed.ts
+++ b/src/commands/embed.ts
@@ -629,15 +629,19 @@ async function embedAllStale(
   const CONCURRENCY = parseInt(process.env.GBRAIN_EMBED_CONCURRENCY || '20', 10);
 
   // D3 + D3a + D8: wall-clock budget. 30 min default; env override.
-  // v0.41.18.0 (A13): --catch-up removes the wall-clock cap entirely so the
-  // handler runs until countStaleChunks() returns 0. Use Number.MAX_SAFE_INTEGER
-  // (effectively unbounded) instead of the 30-min default. The AbortController
-  // still wraps for SIGINT propagation; just the timer never fires.
+  // #1946: --catch-up removes the wall-clock cap entirely. Earlier versions
+  // used `setTimeout(…, Number.MAX_SAFE_INTEGER)`, but JS timer delays are
+  // clamped to a signed-32-bit ms range (~24.8 days); an out-of-range delay
+  // fires on the next tick and aborts the run after the first batch. Skip
+  // arming the timer in catch-up mode instead. The AbortController stays
+  // around so `budgetSignal.aborted` keeps its semantics (never true here).
   const BUDGET_MS = staleOpts?.catchUp
-    ? Number.MAX_SAFE_INTEGER
+    ? null
     : parseInt(process.env.GBRAIN_EMBED_TIME_BUDGET_MS || `${30 * 60 * 1000}`, 10);
   const budgetController = new AbortController();
-  const budgetTimer = setTimeout(() => budgetController.abort(), BUDGET_MS);
+  const budgetTimer = BUDGET_MS == null
+    ? null
+    : setTimeout(() => budgetController.abort(), BUDGET_MS);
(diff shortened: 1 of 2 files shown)


time-attack:
Closing as superseded — this exact fix (skip arming the wall-clock timer under `--catch-up` instead of overflowing `setTimeout`'s 32-bit delay) is already on master in `src/commands/embed.ts` (landed via #1521, v0.41.18.0), with an inline comment citing #1946. Thanks for the clear diagnosis and regression test — please reopen if you can still reproduce the silent-abort on a current build.

PR #1965, opened 2026-06-08
fix(takes): resolve chat model via facts.extraction_model, not hardcoded Haiku (#1857)

### LINKED ISSUE #1857: takes extract --from-pages hardcodes anthropic:claude-haiku-4-5, silently returns 0 claims on OpenAI-only brains

TreyLawrence:
## Summary

Closes #1857.

`extractTakesFromPages`'s per-page `chat()` call hardcoded `opts.model ?? 'anthropic:claude-haiku-4-5'`. The JSDoc on `opts.model` said *"defaults to facts.extraction_model"*, but no caller (`cmdExtract` in `commands/takes.ts`, the `extract-takes-from-pages` Minion handler in `commands/jobs.ts`) set `opts.model`, so every invocation hit the hardcoded Anthropic id regardless of how the brain was configured.

On an OpenAI-only brain without `ANTHROPIC_API_KEY` the failure was silent: the top-level `isAvailable('chat')` gate passed (gateway-configured chat model = OpenAI), but the per-page Anthropic call threw, the surrounding try/catch swallowed it, and the command reported `0 claim(s) from N page(s)` with no error.

## Fix

Resolve the model once via `getFactsExtractionModel(engine)` when `opts.model` is unset — same canonical `resolveModel` chain (`facts.extraction_model` config key → `models.default` → tier `'reasoning'` → env → fallback Sonnet) that `facts/extract.ts` already uses. Callers stay untouched; the fix lives entirely inside the core function so the public contract finally matches its JSDoc.

```ts
const resolvedModel = opts.model ?? await getFactsExtractionModel(engine);
// ...
response = await chat({ model: resolvedModel, ... });
```

## Test plan

New file `test/extract-takes-from-pages.test.ts` pins four points on the resolution chain via the gateway's `__setChatTransportForTests` seam:

- [x] `facts.extraction_model` from engine config (the exact OpenAI-only repro from the issue)
- [x] `models.default` fallthrough when `facts.extraction_model` is unset
- [x] `opts.model` explicit override still wins
- [x] **Regression assertion**: the resolved model is never the pre-fix hardcoded `anthropic:claude-haiku-4-5`

```
4 pass / 0 fail
```

- [x] Adjacent test files unaffected: `extract-takes.test.ts`, `extract-takes-holder-producer-seam.test.ts`, `extract-conversation-facts.test.ts` — 39/39 pass

## Non-goals

This PR keeps scope tight. Two adjacent latent issues the original report touches on but this PR does NOT change:

- **Silent per-page error swallow** (`try { chat(...) } catch { continue }` at `extract-takes-from-pages.ts:187`). After this fix the OpenAI-only repro stops hitting that path because the model now resolves correctly, but the catch is still there for legitimate rate-limit / content-filter transients. A future PR could surface a single aggregated warning when *every* page fails.
- **Choice of fallback model** (`getFactsExtractionModel` falls back to Sonnet). Higher quality than Haiku but ~3× the cost per page. Worth its own discussion since it's a behavior change for brains that were happy with Haiku's defaults pre-fix; matches the existing JSDoc + `facts.extract.ts` pipeline either way.

Diff: 2 files, +150 / -1. No VERSION / CHANGELOG / package.json changes — per the fix-wave workflow in `docs/RELEASING.md`, those are the wave shipper's concern.

### DIFF
--- a/src/core/extract-takes-from-pages.ts
+++ b/src/core/extract-takes-from-pages.ts
@@ -16,6 +16,7 @@
 import type { BrainEngine } from './engine.ts';
 import type { TakeBatchInput, TakeKind } from './engine.ts';
 import { chat, isAvailable } from './ai/gateway.ts';
+import { getFactsExtractionModel } from './facts/extract.ts';
 
 export const ALLOWED_PAGE_TYPES = [
   'concept', 'atom', 'lore', 'briefing', 'writing', 'originals',
@@ -129,6 +130,15 @@ export async function extractTakesFromPages(
   const sourceFilter = opts.sourceIdFilter ? `AND source_id = $1` : '';
   const params = opts.sourceIdFilter ? [opts.sourceIdFilter] : [];
 
+  // #1857: resolve the model once via the canonical facts.extraction_model
+  // chain. Pre-fix this was hardcoded to `anthropic:claude-haiku-4-5`, which
+  // silently returned 0 claims on OpenAI-only brains: per-page chat() would
+  // hit Anthropic with no key and the try/catch below would swallow each
+  // failure. The JSDoc on opts.model said "defaults to facts.extraction_model"
+  // but no caller resolved that — both cmdExtract (commands/takes.ts) and the
+  // Minion handler in commands/jobs.ts leave opts.model unset.
+  const resolvedModel = opts.model ?? await getFactsExtractionModel(engine);
+
   // Fetch eligible pages. Order by updated_at DESC so recently-edited
   // pages get bootstrapped first.
   const typesList = ALLOWED_PAGE_TYPES.map((t) => `'${t}'`).join(', ');
@@ -174,7 +184,7 @@ export async function extractTakesFromPages(
     let response: { text: string };
     try {
       response = await chat({
-        model: opts.model ?? 'anthropic:claude-haiku-4-5',
+        model: resolvedModel,
         system: CLASSIFIER_SYSTEM,
         messages: [
           {
--- a/test/extract-takes-from-pages.test.ts
+++ b/test/extract-takes-from-pages.test.ts
@@ -1,0 +1,139 @@
+/**
+ * Regression tests for #1857 — `extractTakesFromPages` model resolution.
+ *
+ * Pre-fix the per-page classifier call hardcoded
+ * `model: opts.model ?? 'anthropic:claude-haiku-4-5'` even though the JSDoc
+ * said "defaults to facts.extraction_model". Neither caller (cmdExtract,
+ * the Minion handler) set opts.model, so OpenAI-only brains attempted
+ * Anthropic with no key and the per-page try/catch silently continued —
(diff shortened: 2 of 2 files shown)


time-attack:
Closing as duplicate of #1858.

Competing PR for facts.extraction_model in takes extract. Keep #1858.

PR #1967, opened 2026-06-08
fix(facts): cycle reconcile preserves cli-sourced facts (#1928)

### LINKED ISSUE #1928: extract_facts cycle phase hard-deletes conversation facts (deleteFactsForPage has no source filter; failed-sync fallback turns it into a full-brain wipe)
### Summary

The `extract_facts` cycle phase hard-deletes **non-fence facts** (e.g. rows written by `gbrain extract-conversation-facts`) for every page it reconciles. Its own header comment claims these rows survive; the SQL says otherwise. Combined with a failed-sync fallback that turns the phase into a full-brain walk, one autopilot cycle deleted **all 1829 facts** in our brain (the entire conversation-facts backfill) with `factsInserted: 0`.

Observed on v0.42.8.0 (the deleting cycle) and the code path is byte-identical through v0.42.26.0 and current `master`.

### The contradiction

`src/core/cycle/extract-facts.ts` header (lines ~18-22) says:

> Pages with no fence go through delete-then-empty-insert — DB rows for that page coordinate are wiped; legacy NULL-source_markdown_slug rows **survive because deleteFactsForPage targets source_markdown_slug = slug only**.

That protects only rows with `source_markdown_slug IS NULL`. But `extract-conversation-facts` deliberately stamps `source_markdown_slug` with the transcript page slug (page-global `row_num` contract, `src/commands/extract-conversation-facts.ts`). So conversation facts live exactly on the coordinate the wipe targets:

```ts
// src/core/postgres-engine.ts
async deleteFactsForPage(slug: string, source_id: string): Promise<{ deleted: number }> {
  const result = await sql`
    DELETE FROM facts WHERE source_id = ${source_id} AND source_markdown_slug = ${slug}
  `;
}
```

No `source` filter. Since conversation pages carry no `## Facts` fence (by design — "the chat-log shape is the source-of-truth"), the reconcile for a transcript page is: delete every fact, reinsert nothing.

Notably, `extract-conversation-facts` itself already knows how to do this safely — its `deleteOrphanFactsForPage` scopes its delete with `AND source LIKE 'cli:extract-conversation-facts%'`. The cycle phase lacks the inverse scoping.

### The amplifier: failed sync ⇒ full-brain walk

On an ordinary night the phase only reconciles `syncPagesAffected`, so transcript pages are rarely touched and the bug stays hidden. But in `src/core/cycle.ts` (~line 1608):

```
// If sync didn't run (phases exclude it) or failed, syncPagesAffected
// is undefined → extract falls back to full walk (safe default).
```

`undefined` → full walk is a safe default for link **extraction**; for a **destructive wipe-and-reinsert** it is the opposite. In our case a standalone `sync` job and the autopilot cycle started ~40 ms apart; the cycle's sync phase failed on the sync lock ("Another sync is in progress"), `syncPagesAffected` came back `undefined`, and `extract_facts` walked all 1587 pages:

```json
{ "phase": "extract_facts", "status": "ok",
  "details": { "factsDeleted": 1829, "pagesScanned": 1587, "factsInserted": 0, "pagesWithFacts": 0 },
  "summary": "0 fact(s) reconciled across 1587 page(s)" }
```

Status `ok`, summary reads like a no-op — 1829 rows gone.

### Repro sketch

1. `gbrain extract-conversation-facts` over any conversation page → facts rows with `source = 'cli:extract-conversation-facts'` and `source_markdown_slug = <page slug>`.
2. Run a cycle whose sync phase fails (e.g. hold the sync lock), or any full-walk `extract_facts`.
3. The conversation facts are deleted; nothing is reinserted.

### Suggested fix

Scope the cycle's wipe to fence-owned rows only — e.g. at the call-site in `extract-facts.ts`, replace `deleteFactsForPage` with a delete carrying `AND source NOT LIKE 'cli:%'` (mirroring `deleteOrphanFactsForPage`'s scoping), or add a source-prefix exclusion parameter to `deleteFactsForPage`. We run this exact call-site guard locally and fence reconciliation (insert/edit/shrink, idempotent re-runs) still byte-matches while `cli:%` rows survive full walks.

Secondary hardening worth considering: a destructive phase probably shouldn't inherit the "failed sync ⇒ full walk" fallback, and `factsDeleted >> factsInserted` on a full walk could warrant at least a `warn` status.

TreyLawrence:
## Summary

Closes #1928. **Data-loss class bug** per the issue (one cycle wiped 1829 conversation-facts rows in the reporter's brain).

The `extract_facts` cycle phase wiped every fact row on each reconciled page's `(source_id, source_markdown_slug)` coordinate. CLI deposits — notably `gbrain extract-conversation-facts`, which writes rows with `source LIKE 'cli:extract-conversation-facts%'` against transcript pages that carry **no `## Facts` fence by design** — live on that exact coordinate. So a reconcile for a transcript page was: delete every fact, reinsert nothing.

Combined with the failed-sync ⇒ full-walk fallback in `cycle.ts` (`syncPagesAffected = undefined` → walk every page), one autopilot cycle could wipe an entire conversation-facts backfill while reporting `status: 'ok'` and `0 fact(s) reconciled across N page(s)`.

## Fix

Add an optional `excludeSourcePrefixes` option to `deleteFactsForPage`. The cycle phase passes `['cli:']` so CLI-deposited rows survive the wipe.

### Files

- `src/core/engine.ts` — interface: optional `opts?: { excludeSourcePrefixes?: string[] }` on `deleteFactsForPage`. JSDoc updated.
- `src/core/postgres-engine.ts` + `src/core/pglite-engine.ts` — engine parity. Both implementations short-circuit to the existing single-statement DELETE when no prefixes are passed (back-compat) and add `AND NOT (source LIKE ANY($N::text[]))` when prefixes are present.
- `src/core/cycle/extract-facts.ts` — call site passes `excludeSourcePrefixes: ['cli:']`. File header updated to document the exclusion.

The reporter's suggested-fix paragraph from the issue mentions running this exact call-site guard locally and confirming fence reconciliation still byte-matches.

## Non-goals (issue's "secondary hardening" — not in this PR)

The issue suggests two additional hardenings worth considering separately:

1. **Destructive phase shouldn't inherit the "failed-sync ⇒ full-walk" fallback.** That's a `cycle.ts` change at a different layer and worth its own PR — the fix here makes the data loss impossible regardless of whether the fallback runs, so it's strictly additive.
2. **`factsDeleted >> factsInserted` should `warn` status.** Useful canary, but signal-only — also strictly additive on top of this fix.

Kept the scope tight to the data-loss fix; happy to follow up on either if you'd like.

## Test plan

`test/extract-facts-phase.test.ts` gains a new `#1928 CLI-deposited facts survive the wipe` describe with five tests:

- [x] CLI-sourced fact on a fenceless page survives slug-scoped reconcile (`factsDeleted: 0`)
- [x] Full-walk fallback (no `slugs` filter) does NOT wipe CLI deposits brain-wide (the "amplifier" scenario from the issue)
- [x] Non-cli rows on the same `(source_id, source_markdown_slug)` coordinate ARE still wiped (reconcile semantics preserved for `sync:import` / `mcp:put_page` fence rows)
- [x] Engine-level pin: `deleteFactsForPage` honors the option directly (1 deleted, 1 cli row survives)
- [x] Back-compat: no opts deletes every coordinate row (pre-#1928 behavior unchanged)

Result:

```
18 pass / 0 fail in extract-facts-phase.test.ts
96 pass / 0 fail across extract-facts-phase, insert-facts-batch, phantom-redirect, extract-conversation-facts
```

Diff: 5 files, +211 / -8.

### DIFF
--- a/src/core/cycle/extract-facts.ts
+++ b/src/core/cycle/extract-facts.ts
@@ -19,7 +19,12 @@
  * no fence go through delete-then-empty-insert — DB rows for that
  * page coordinate are wiped; legacy NULL-source_markdown_slug rows
  * survive because deleteFactsForPage targets source_markdown_slug =
- * slug only.
+ * slug only. #1928 — CLI-deposited rows (`source LIKE 'cli:%'`) also
+ * survive: `gbrain extract-conversation-facts` writes to transcript
+ * pages that carry no `## Facts` fence by design, so the wipe at the
+ * fence-reconcile coordinate must skip rows it doesn't own. The
+ * exclusion is passed at the call site via deleteFactsForPage's
+ * `excludeSourcePrefixes` option.
  *
  * Empty-fence guard (Codex R2-#7): the phase refuses to do its
  * destructive reconciliation pass when legacy rows (row_num IS NULL,
@@ -212,7 +217,14 @@ export async function runExtractFacts(
     // Wipe-and-reinsert per page. The deleteFactsForPage call targets
     // source_markdown_slug = slug only, so NULL-source_markdown_slug
     // legacy rows survive (the partial-UNIQUE-index keyspace).
-    const deleted = await engine.deleteFactsForPage(slug, sourceId);
+    // #1928: also exclude `cli:%`-sourced rows so out-of-band CLI deposits
+    // (e.g. `cli:extract-conversation-facts`, which write to conversation
+    // pages that carry no `## Facts` fence by design) survive the reconcile.
+    // Without this, a full-walk reconcile (which the cycle falls back to
+    // when sync fails — see cycle.ts) wipes every conversation-fact row.
+    const deleted = await engine.deleteFactsForPage(slug, sourceId, {
+      excludeSourcePrefixes: ['cli:'],
+    });
     result.factsDeleted += deleted.deleted;
 
     if (parsed.facts.length === 0) continue;
--- a/src/core/engine.ts
+++ b/src/core/engine.ts
@@ -1653,8 +1653,18 @@ export interface BrainEngine {
    * structural guarantee that legacy rows live in a different keyspace
    * until the v0_32_2 migration backfills them. Cycle-phase callers in
    * commit 7 add the empty-fence-guard as a belt-and-suspenders check.
+   *
+   * #1928 — `opts.excludeSourcePrefixes` skips rows whose `source` matches
+   * any of the given LIKE prefixes (each prefix is matched as `prefix%`).
+   * The cycle's `extract_facts` phase passes `['cli:']` so CLI-deposited
+   * rows (e.g. `cli:extract-conversation-facts`, which write to pages
+   * that have no `## Facts` fence by design) survive the fence reconcile.
    */
-  deleteFactsForPage(slug: string, source_id: string): Promise<{ deleted: number }>;
+  deleteFactsForPage(
+    slug: string,
+    source_id: string,
+    opts?: { excludeSourcePrefixes?: string[] },
+  ): Promise<{ deleted: number }>;
 
   /**
    * Mark a fact expired. Never DELETE. Returns true iff a row was updated.
--- a/src/core/pglite-engine.ts
+++ b/src/core/pglite-engine.ts
@@ -3506,10 +3506,26 @@ export class PGLiteEngine implements BrainEngine {
     return { inserted: ids.length, ids };
   }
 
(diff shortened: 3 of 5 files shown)


time-attack:
Thanks for this, @TreyLawrence! 🙏

What this does: This PR fixes a data-loss bug where the extract_facts cycle would wipe CLI-deposited facts (like conversation facts). The fix adds excludeSourcePrefixes to prevent deletion of rows from CLI sources. This exact fix is already on master with…

PR #1968, opened 2026-06-08
fix(dream): persist oversize-after-split skip to dream_verdicts (#1879)

### LINKED ISSUE #1879: dream synthesize: oversized session transcripts silently burn tokens on timeout (need a size guard / pre-chunk)
On a Supabase/pooler thin-client brain, the dream `synthesize` phase repeatedly fails to process large session transcripts: the per-transcript subagent ingests the whole file, times out, and writes nothing. The nightly dream summary shows e.g. `Children: 0 completed, 2 failed/timeout. Pages written: 0` for consecutive nights.

**Concrete case:** a session-corpus transcript of ~5 MB (~1.4M tokens, 82k lines). The synthesize subagent (`anthropic:claude-sonnet-4-6`) is handed the full file, ingests it, and times out before producing output. Because the corpus entry is keyed by `(file_path, content_hash)` for idempotency, the same oversized file is retried every night and fails every night — so it never gets smaller and never succeeds.

**The cost angle is the real issue:** each failed attempt still pays for the ingested input (~$/MB of input tokens), recurring nightly, for zero pages written. And a timed-out child is not recorded in `subagent_messages` (only completed children are), so the spend is invisible in the brain's own accounting (`minion_budget_log` / `budget_ledger` are empty) — you only see it on the provider bill.

**Ask:** a configurable size guard in the synthesize phase, e.g.:
- Skip + warn (record a skip-reason in the dream summary / `dream_verdicts`) when a transcript exceeds a token/byte threshold, instead of handing it to the subagent and timing out.
- Or pre-chunk / map-reduce oversized transcripts (synthesize per chunk, then merge), so large inputs still produce pages without a single oversized call.
- Either way, never retry the same oversized transcript indefinitely with no backoff or cap.

This is defense-in-depth: producers should keep transcripts lean too, but the synthesize phase should not silently burn tokens on inputs it cannot complete.

TreyLawrence:
## Summary

Closes #1879.

The synthesize phase's `chunks > max_chunks_per_transcript` skip (`oversize_after_split`) was logged + tracked in `skipReports`, but deliberately NOT cached in `dream_verdicts`. The original SynthConfig comment cited the *"poison-pill class"*: next cycle should re-attempt under whatever budget is then current.

In practice that means the same oversize transcript gets re-chunked and re-dispatched every cycle. Each per-chunk subagent (Sonnet, 30-min timeout, 30 turns) burns input tokens. **Timed-out children are not recorded in `subagent_messages`**, so the spend doesn't show up in `minion_budget_log` / `budget_ledger` — the recurring cost is invisible in gbrain's own accounting; you only see it on the provider bill. That's the issue's actual complaint.

## Fix

Add `dream.synthesize.persist_oversize_skip` config (default **true**). When the cap fires, also write a `worth_processing: false` row to `dream_verdicts` with reasons `['oversize_after_split: N/cap=M']`. The next cycle's verdict-cache check short-circuits **before** chunking + fan-out: zero subagent jobs submitted, zero recurring cost.

The cache key is `(file_path, content_hash)` — an edited transcript changes the hash and gets re-evaluated naturally. Operators who want the legacy "re-evaluate every cycle under possibly-different budget" behavior can opt out:

```sh
gbrain config set dream.synthesize.persist_oversize_skip false
```

When the persist fires, stderr also prints a one-liner pointing at the opt-out + the content-edit retry path, so the behavior is discoverable.

### Files
- `src/core/cycle/synthesize.ts` — new `SynthConfig.persistOversizeSkip` field, read from config, branched at the skip site. Existing `skipReports` push is unchanged.
- `src/core/config.ts` — TypeScript type field, DB-plane merge, canonical key list.
- `test/e2e/dream-synthesize-chunking.test.ts` — tests updated + extended (see below).

## Test plan

`test/e2e/dream-synthesize-chunking.test.ts` — 3 tests under the existing `D5 cap hit` describe:

- [x] **New default behavior** — `chunks > max_chunks_per_transcript` upserts the verdict row to `worth_processing: false` with reasons including `oversize_after_split`. Updated from the old "no verdict-cache write" assertion.
- [x] **Legacy escape hatch** — with `persist_oversize_skip=false`, the verdict cache is untouched (the seeded `worth_processing: true` row survives the cycle). Pins the opt-out path so future work doesn't silently drop it.
- [x] **Follow-through** — after the persist, the **subsequent** cycle short-circuits at the verdict-cache check: `transcripts_processed: 0`, zero subagent jobs submitted, no fan-out. This is the actual cost-savings path.

```
6 pass / 0 fail in dream-synthesize-chunking.test.ts
94 pass / 0 fail across cycle-synthesize{,-chunker,-slug-collection}, loadConfig-merge, e2e dream-synthesize-chunking
```

## Non-goals

The issue mentions two related shapes left out of this PR:

1. **Map-reduce / per-chunk merge** for oversized transcripts so large inputs still produce pages. Bigger architectural change — worth its own PR.
2. **`gbrain dream --phase synthesize --input <file> --bypass-verdict-cache`** flag for operators who want to retry a locked-out transcript without editing it or deleting the verdict row. The existing `--input` already bypasses cooldown; a parallel verdict-cache bypass would be the natural follow-up. Also out of scope here.

The fix here stops the cost recurrence; the follow-ups expand what's recoverable. Happy to do either if you'd like.

Diff: 3 files, +185 / -17.

### DIFF
--- a/src/core/config.ts
+++ b/src/core/config.ts
@@ -240,6 +240,13 @@ export interface GBrainConfig {
       verdict_model?: string;
       max_prompt_tokens?: number;
       max_chunks_per_transcript?: number;
+      /**
+       * #1879 — when true (default), the oversize_after_split skip in the
+       * synthesize phase persists to `dream_verdicts` so subsequent runs
+       * short-circuit on the cached verdict instead of re-chunking and
+       * burning per-chunk subagent timeouts every cycle.
+       */
+      persist_oversize_skip?: boolean;
     };
     patterns?: {
       lookback_days?: number;
@@ -623,6 +630,7 @@ export async function loadConfigWithEngine(
   const dbVerdictModel = await dbStr('dream.synthesize.verdict_model');
   const dbMaxPromptTokens = await dbInt('dream.synthesize.max_prompt_tokens');
   const dbMaxChunksPerTranscript = await dbInt('dream.synthesize.max_chunks_per_transcript');
+  const dbPersistOversizeSkip = await dbBool('dream.synthesize.persist_oversize_skip');
   const dbLookbackDays = await dbInt('dream.patterns.lookback_days');
   const dbMinEvidence = await dbInt('dream.patterns.min_evidence');
 
@@ -646,6 +654,9 @@ export async function loadConfigWithEngine(
   }
   if (mergedSynth.max_chunks_per_transcript === undefined && dbMaxChunksPerTranscript !== undefined) {
     mergedSynth.max_chunks_per_transcript = dbMaxChunksPerTranscript;
+  }
+  if (mergedSynth.persist_oversize_skip === undefined && dbPersistOversizeSkip !== undefined) {
+    mergedSynth.persist_oversize_skip = dbPersistOversizeSkip;
   }
   if (mergedPatterns.lookback_days === undefined && dbLookbackDays !== undefined) {
     mergedPatterns.lookback_days = dbLookbackDays;
@@ -751,6 +762,7 @@ export const KNOWN_CONFIG_KEYS: readonly string[] = [
   'dream.synthesize.verdict_model',
   'dream.synthesize.max_prompt_tokens',
   'dream.synthesize.max_chunks_per_transcript',
+  'dream.synthesize.persist_oversize_skip',
   'dream.patterns.lookback_days',
   'dream.patterns.min_evidence',
   // Emotional weight (v0.29)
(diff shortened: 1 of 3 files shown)


time-attack:
Closing after review: on current master the cap-hit path in synthesize.ts already skips with continue BEFORE the subagent fan-out, and the significance verdict is cached — the double-spend this PR guards against cannot occur. Thanks for the careful read of the loop.

Issue #1971, opened 2026-06-08
gbrain auth register-client missing --bound-* flags for agent-scoped clients


TreyLawrence:
## Summary

`gbrain auth register-client` does not expose flags to set the `bound_*` columns on `oauth_clients`. As a result, freshly registered agent-scoped clients can't actually call `submit_agent` — they get rejected with a `permission_denied` error that references CLI flags that don't exist on the command.

## Reproduce

1. Register a client with the `agent` scope:
   ```bash
   gbrain auth register-client my-agent \
     --grant-types client_credentials \
     --scopes 'read write agent' \
     --source default
   ```
2. Use the resulting client to call `submit_agent` via MCP.
3. Error returned:
   ```
   submit_agent: client gbrain_cl_xxx has the agent scope but no bindings.
   Re-register with --bound-tools, --bound-source, --bound-slug-prefixes,
   --bound-max-concurrent, --budget-usd-per-day.
   ```
4. But `gbrain auth register-client --help` does **not** document any of those flags, and `RegisterClientArgs` in `src/commands/auth.ts` has only:
   - `grantTypes`, `scopes`, `sourceId`, `federatedRead`, `redirectUris`, `tokenEndpointAuthMethod`

## What's happening

The schema (`src/core/pglite-schema.ts`, `oauth_clients` table) includes:
```
budget_usd_per_day      NUMERIC(10, 2) NULL,
bound_tools             TEXT[] NULL,
bound_source_id         TEXT NULL,
bound_brain_id          TEXT NULL,
bound_slug_prefixes     TEXT[] NULL,
bound_max_concurrent    INTEGER NOT NULL DEFAULT 1,
```

…and `submit_agent`'s check in `src/core/operations.ts` (line ~2658) reads these to decide whether to allow the call. But the CLI registration path never populates them, so they're always NULL on freshly registered clients, and the error message tells the user to use CLI flags that haven't been wired up.

## Affected version

`gbrain 0.42.32.0` (probably any version since `v85` migration introduced the `bound_*` columns).

## Workaround

Direct SQL UPDATE on the `oauth_clients` row works:

```sql
UPDATE oauth_clients SET
  bound_tools = ARRAY['get_skill','search','query','list_pages','get_page','put_page',...],
  bound_source_id = 'default',
  bound_slug_prefixes = ARRAY['companies/','projects/','topics/','meetings/','reports/'],
  bound_max_concurrent = 1,
  budget_usd_per_day = 5
[...]

Issue #1945, opened 2026-06-07
submit_agent bindings gate (operations.ts:2658) is unsatisfiable — --bound-* flags don't exist in `gbrain auth register-client`


dcarolan1:
**Version:** 0.42.29.0

### Summary

`submit_agent` enforces two gates on the calling OAuth client:

1. `operations.ts:2625` — requires the `agent` scope. Satisfiable via `gbrain auth register-client --scopes "read write agent"`.
2. `operations.ts:2658` — requires *bindings*, throwing:
   `submit_agent: client <id> has the agent scope but no bindings. Re-register with --bound-tools, --bound-source, --bound-slug-prefixes, --bound-max-concurrent, --budget-usd-per-day.`

The problem: **none of those `--bound-*` flags exist in `gbrain auth register-client`** in 0.42.29.0. The CLI rejects `--bound-tools` as an unknown flag. The available flags are only: `--grant-types`, `--scopes`, `--source`, `--federated-read`, `--redirect-uri`, `--token-endpoint-auth-method`.

The `--bound-*` names appear **only** in the gate-2 throw string — there is no CLI path to set the bindings the throw demands. The feature looks half-shipped (design tagged `v0.38 Slice 3 — D13 — remote-callable submit_agent with registration-time [bindings]`): the enforcement landed; the registration surface didn't.

### Impact

An `agent`-scope client minted without bindings clears gate 1, then hits gate 2 on the first `submit_agent` call and fails. So remote-MCP agent-to-agent dispatch (the documented Minions use case — one agent submits a subagent job to another agent's queue) can't be enabled the documented way. The only workarounds are shelling out via `gbrain agent run` (bypasses the OAuth-bindings discipline entirely) or hand-patching the `bound_*` DB columns (writes data the CLI should own) — neither desirable.

### Question

Is the `--bound-*` registration surface planned for a near release, or is there an alternate documented way to attach bindings to an `agent`-scope client (a separate `gbrain auth set-bindings <client>` command, a config block, an API call) that I've missed?

### Repro

​```
gbrain auth register-client test-agent \
  --grant-types authorization_code,refresh_token \
  --scopes "read write agent" --source X --federated-read X,default
# then a submit_agent MCP call from that client →
# "has the agent scope but no bindings. Re-register with --bound-tools ..."

gbrain auth register-client test-agent2 ... --bound-tools search,query
# → Error: Unknown flag --bound-tools
​```

dcarolan1:
[...]

PR #1976, opened 2026-06-08
fix(auth): add register-client agent binding flags

mzkarami:
## Summary
- Adds `gbrain auth register-client` flags for submit_agent binding fields: `--bound-tools`, `--bound-source`, `--bound-brain`, `--bound-slug-prefixes`, `--bound-max-concurrent`, and `--budget-usd-per-day`.
- Persists those values into the existing OAuth client binding columns when the flags are supplied.
- Adds focused parser and OAuth persistence coverage for the new registration path.

## Rationale
Refs #1945 and #1971.

`submit_agent` registration currently checks for OAuth client bindings, but the manual `gbrain auth register-client` path does not expose the corresponding `--bound-*` flags. That makes it difficult for operators to create agent-scoped clients that satisfy the existing registration-time binding gate.

This keeps existing client registration behavior unchanged when no binding flags are supplied, and only writes the binding columns for registrations that opt into the new flags.

## Test plan
- [x] `git diff --check`
- [x] `bun test test/auth-register-client-args.test.ts`
- [x] `bun test test/oauth.test.ts`
- [x] `bun run typecheck`

## Risk
Low. The change extends manual OAuth client registration and preserves the old insert path for no-binding registrations. Existing CLI flags and DCR behavior are unchanged.

### DIFF
--- a/src/commands/auth.ts
+++ b/src/commands/auth.ts
@@ -346,6 +346,12 @@ interface RegisterClientArgs {
   federatedRead: string[] | undefined;
   redirectUris: string[];
   tokenEndpointAuthMethod: string | undefined;
+  boundTools: string[] | undefined;
+  boundSourceId: string | undefined;
+  boundBrainId: string | undefined;
+  boundSlugPrefixes: string[] | undefined;
+  boundMaxConcurrent: number | undefined;
+  budgetUsdPerDay: string | undefined;
 }
 
 export function parseRegisterClientArgs(args: string[]): RegisterClientArgs {
@@ -356,6 +362,12 @@ export function parseRegisterClientArgs(args: string[]): RegisterClientArgs {
     federatedRead: undefined,
     redirectUris: [],
     tokenEndpointAuthMethod: undefined,
+    boundTools: undefined,
+    boundSourceId: undefined,
+    boundBrainId: undefined,
+    boundSlugPrefixes: undefined,
+    boundMaxConcurrent: undefined,
+    budgetUsdPerDay: undefined,
   };
   let i = 0;
   let grantTypesSet = false;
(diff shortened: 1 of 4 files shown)


dcarolan1:
Thanks for putting this together, @mzkarami — this is exactly the flag set #1945 was missing. We're running a multi-agent gbrain fleet and submit_agent delegation is blocked on precisely these register-client binding flags (--bound-tools, --bound-source, --bound-slug-prefixes, --bound-max-concurrent, --budget-usd-per-day). Happy to test against our setup once it lands. Would love to see this reviewed when you get a chance, @garrytan.

dcarolan1:
Chiming in as a downstream user who's been blocked on exactly this. We run a small multi-agent gbrain fleet (4 agents on a shared Supabase Postgres engine, v0.42.29.0), and submit_agent delegation is the one capability we've had to leave closed — an agent-scope client clears the scope gate at operations.ts:2625 but hits the bindings gate at operations.ts:2658, and there's no CLI surface to satisfy it because gbrain auth register-client doesn't accept the --bound-* flags the throw string asks for. That's the gap I filed as #1945.
This PR adds precisely those flags (--bound-tools, --bound-source, --bound-brain, --bound-slug-prefixes, --bound-max-concurrent, --budget-usd-per-day) and persists them into the existing OAuth binding columns — which is exactly the registration-time path the gate-2 throw points to. We've deliberately held off running the branch directly (we keep the fleet on released, merged paths only), so a review + merge + tagged release would unblock us cleanly.
Thank you @mzkarami for putting this together, and thanks @garrytan for the project. Happy to test against our fleet once it lands if that's useful. 🙏

mzkarami:
YW @dcarolan1 . Happy to help further @garrytan if merging this PR needs further help/support.

PR #1982, opened 2026-06-08
docs(install): surface OPENROUTER_API_KEY as the recommended first option in Step 2

ttomiczek:
## What

The install doc currently leads with three provider-direct keys (ZeroEntropy, OpenAI, Anthropic) without mentioning that gbrain has full OpenRouter support already documented in `docs/integrations/embedding-providers.md`.

For new users who don't have a strong preference, a single `OPENROUTER_API_KEY` covers embeddings, chat, and rerankers across 100+ upstream providers with one bill and one payment endpoint. That's a meaningfully better first-install experience than handing the user a 3-key shopping list.

## Why I hit this

Installing gbrain for an agent stack that already had a few LLM endpoints configured. The setup assumed I'd want to add a separate API key for each provider family. I would have preferred one OpenRouter key for everything. The recipe was already there — just not surfaced at the install-front-door.

## Changes

**`INSTALL_FOR_AGENTS.md` Step 2:**
- Add a TL;DR callout putting `OPENROUTER_API_KEY` first
- Add config defaults for the OR-only case (`embedding_model = openrouter:openai/text-embedding-3-small`, `chat_model` left at gbrain default for subagent compatibility)
- Add the `OPENROUTER_REFERER` / `OPENROUTER_TITLE` attribution env vars so forks/operators get credited on OR's leaderboard
- Keep the existing provider-direct options as a fallback path

**`AGENTS.md` install step 3:**
- Insert a STOP-and-ask-for-API-keys step before the existing STOP-and-ask-for-search-mode step, linking to the OpenRouter recipe

## Notes

The full provider matrix, model list, subagent caveat, and pricing stay where they are in `docs/integrations/embedding-providers.md`. This PR just makes sure new agents don't miss the option at install time.

Happy to revise phrasing, restructure as a separate `Step 1.5: pick your LLM gateway`, or split the AGENTS.md change into a separate PR if maintainers prefer. Just let me know.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -18,7 +18,13 @@ start here.
    deterministic install: `git clone https://github.com/garrytan/gbrain.git ~/gbrain && cd ~/gbrain && bun install && bun link`.
 2. Init the brain: `gbrain init` (defaults to PGLite, zero-config). For 1000+ files or
    multi-machine sync, init suggests Postgres + pgvector via Supabase.
-3. **STOP — ask the user about search mode.** `gbrain init` auto-applied a
+3. **STOP — ask the user about API keys.** Before asking, surface that
+   `OPENROUTER_API_KEY` is the recommended first option: one key covers embeddings,
+   chat, and rerankers across 100+ upstream providers. Full recipe and model list at
+   [`./docs/integrations/embedding-providers.md`](./docs/integrations/embedding-providers.md#openrouter).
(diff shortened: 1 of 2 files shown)


time-attack:
Thanks for the suggestion and for explaining the install experience that prompted it. We don't want the install front door to recommend one specific third-party aggregator as the default path — especially with attribution/referrer env vars attached — so I'm closing this. OpenRouter remains fully supported and documented in `docs/integrations/embedding-providers.md` for anyone who prefers a single key.

PR #1983, opened 2026-06-08
fix(links): resolve [[wikilink]] + exact slug-path values in frontmatter link fields

spiky02plateau:
## Problem

Frontmatter link fields (`related:`, `see_also:`, …) silently drop two common value shapes that point at **exact-existing pages**:

1. **Obsidian `[[wikilink]]` values.** Mainstream Obsidian authors frontmatter links as `related: ["[[Page]]"]`. The resolver treats the brackets as part of the value, so `[[90-people/nicolai]]` is normalized to `90peoplenicolai` and never resolves.
2. **Digit-leading / nested slug paths.** The step-1 slug regex `^[a-z][a-z0-9-]*\/[a-z0-9][a-z0-9-]*$` rejects PARA-style paths like `90-people/nicolai` or `01-trading/wiki/strategies/opening-range-breakout` (leading digit / more than two segments). Full-path fuzzy then scores below the 0.55 threshold against the bare title, so a page that exists *exactly* is missed.

These wear one symptom — frontmatter `related:` blocks that produce zero edges even though the target page is present.

## Fix

- **`unwrapWikilink(value)`** — unwrap a wholly-wrapped `[[…]]` frontmatter value to its bare link target (dropping a trailing `|alias`, `#heading`, or `^block`) before resolution. The regex is anchored (`^\s*\[\[…\]\]\s*$`), so bare titles and any partially-wrapped text pass through unchanged — existing behavior is preserved exactly.
- **Broaden resolver step 1** to attempt an exact `getPage()` for any slug-shaped value (contains `/`, slug charset), covering digit-leading and nested paths. This is **exact-match only — no fuzzy** — so it can never produce a false positive; a non-existent slug just falls through to the existing dir-hint / fuzzy steps.

Bare-title and wrapped-title (fuzzy) resolution are unchanged; the change only *adds* the previously-dropped exact-slug recoveries.

## Evidence

On a real ~1,900-page PARA-numbered vault, this recovered **~81 frontmatter `related_to` edges** that each resolved to an exact-existing page (zero false positives — every recovered value was an exact `getPage()` hit). Before: those values landed in the unresolved set; after: they materialize as edges.

## Tests

Adds unit tests (no DB required, `bun test test/link-extraction.test.ts`) covering:

- `unwrapWikilink`: wrapped title, wrapped slug-path (digit-leading + nested), `|alias` / `#heading` / `^block` stripping, surrounding whitespace, bare-passthrough, and partially-wrapped (not unwrapped).
- `makeResolver` step 1: digit-leading slug, nested slug, single-segment regression, **exact-only safety** (slug-shaped but absent → `null`, no false positive), non-slug → fuzzy.
- End-to-end `extractFrontmatterLinks`: wrapped slug-path resolves (the core win), nested, `|alias`, bare-slug regression, wrapped-title-via-fuzzy regression, and unknown wrapped slug → unresolved with the original value preserved (no crash).

All 147 tests in `test/link-extraction.test.ts` pass.

### DIFF
--- a/src/core/link-extraction.ts
+++ b/src/core/link-extraction.ts
@@ -931,8 +931,17 @@ export function makeResolver(
 
       const hints = Array.isArray(dirHint) ? dirHint : (dirHint ? [dirHint] : []);
 
-      // Step 1: already a slug? (dir/name shape, lowercase, hyphenated)
-      if (/^[a-z][a-z0-9-]*\/[a-z0-9][a-z0-9-]*$/.test(trimmed)) {
+      // Step 1: already a slug? Try an exact page lookup for any slug-shaped
+      // value (contains '/', slug charset). Broadened beyond the original
+      // single-segment lowercase-leading form (`^[a-z][a-z0-9-]*\/[a-z0-9]...`)
+      // to also accept digit-leading folders (`90-people/nicolai`,
+      // `01-trading/...`) and nested paths (`a/b/c`) — common in PARA-numbered
+      // vaults. This is an EXACT getPage match only — no fuzzy — so it never
+      // produces a false positive; a non-existent slug just falls through to
+      // the steps below. Fixes frontmatter `related: [[dir/slug]]` values
+      // (unwrapped by unwrapWikilink) that name a real page the strict regex
+      // could not reach and whose full-path fuzzy score is below threshold.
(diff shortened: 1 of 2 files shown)


time-attack:
Superseded with credit by #3087 (just merged) — your change landed rebased onto current master with authorship preserved via Co-authored-by. Thank you!

PR #1990, opened 2026-06-08
Withdrawn

genrae-actions:
Withdrawn by the author. The source branch and public fork have been deleted.


time-attack:
Thank you for the investigation — the source-scoping observation in the FS extraction path is a legitimate concern and the write-up was clear. However, the bulk of this PR is a new 383-line Python evaluator plus a 201-line example config and a package.json entry, which is eval tooling and a new runtime we're not adding to this repo, and the example config carries deployment-specific terms that look like a real vault's layout. Because the extract change is bundled with all of that, I'm closing this rather than merging it as-is.

PR #1991, opened 2026-06-08
feat(ai): declare chat touchpoint for ollama recipe

michaeladair44:
## Summary

The `ollama` recipe in `src/core/ai/recipes/ollama.ts` only declares an `embedding` touchpoint, even though Ollama serves any pulled model via OpenAI-compatible `/v1/chat/completions`. As a result, routing `tier.utility` (or any chat tier) at `ollama:<model>` throws:

```
AIConfigError: Provider "ollama" does not offer a chat touchpoint.
```

…despite `capabilities.ts:80`'s own error hint already listing `ollama` in the known chat-providers list.

This PR closes that gap so users with a local Ollama-served model (Qwen2.5, Llama 3.1, Mistral, etc.) can route gbrain's utility/subagent tier to local inference and eliminate Anthropic spend on Haiku-class background work (expansion, dream verdict, contradictions judge).

## Changes

- **`src/core/ai/recipes/ollama.ts`** — add `chat` touchpoint with:
  - `models: []` — user-driven (openai-compat tier does not enforce list at runtime; matches `openrouter.ts` and `litellm-proxy.ts` pattern)
  - `supports_tools: true` — Qwen2.5, Llama 3.1+, Mistral, Granite all support `/v1` tool calls via Ollama's OpenAI-compat layer
  - `supports_subagent_loop: false` — real gate is `isAnthropicProvider()` upstream; informational only (same as openrouter)
  - `supports_prompt_cache: false`
  - `cost_per_1m_input_usd: 0`, `cost_per_1m_output_usd: 0`
- **`src/core/ai/model-resolver.ts`** — drop `ollama` from the embedding-only fix-hint branch (voyage stays — it really is embedding-only)
- **`test/ai/gateway-chat.test.ts`** — replace the negative-assertion test ("ollama does NOT declare chat") with positive assertions that ollama declares chat with the expected user-driven shape, and that `assertTouchpoint` accepts arbitrary local model ids (`qwen2.5:14b`, `llama3.1:8b`)

## Tested

Locally against Qwen2.5-14B on RTX 4080 SUPER:

```bash
gbrain config set models.tier.utility ollama:qwen2.5:14b
```

Routing dashboard correctly cascades `models.expansion`, `models.dream.synthesize_verdict`, and `models.eval.contradictions_judge` to local Ollama.

```bash
bun test test/providers.test.ts \
         test/ai/recipes-existing-regression.test.ts \
         test/ai/gateway-chat.test.ts \
         test/ai/build-gateway-config.test.ts
# 58 pass / 0 fail.
```

## Test plan

- [x] All targeted unit tests pass (58/58)
- [x] `gbrain config set models.tier.utility ollama:qwen2.5:14b` succeeds (was previously rejected at config-write or first-call time)
- [x] `gbrain models` shows utility tier + cascaded per-task overrides pointing at `ollama:qwen2.5:14b`
- [x] Direct smoke test of Ollama `/v1/chat/completions` returns valid OpenAI-format response

### DIFF
--- a/src/core/ai/model-resolver.ts
+++ b/src/core/ai/model-resolver.ts
@@ -117,7 +117,7 @@ export function assertTouchpoint(
       `Provider "${recipe.id}" does not support touchpoint "${touchpoint}".`,
       touchpoint === 'embedding' && recipe.id === 'anthropic'
         ? 'Anthropic has no embedding model. Use openai or google for embeddings.'
-        : touchpoint === 'chat' && (recipe.id === 'voyage' || recipe.id === 'ollama')
+        : touchpoint === 'chat' && recipe.id === 'voyage'
           ? `${recipe.name} is configured here only for embeddings. Use openai/anthropic/google/deepseek/groq/together for chat.`
           : undefined,
     );
--- a/src/core/ai/recipes/ollama.ts
+++ b/src/core/ai/recipes/ollama.ts
@@ -21,6 +21,32 @@ export const ollama: Recipe = {
       // OLLAMA_NUM_PARALLEL config; no static cap to declare. v0.32 (#779).
       no_batch_cap: true,
     },
+    chat: {
+      // Ollama serves whatever model the user has pulled. The openai-compat
+      // tier does NOT enforce this list at runtime — users pass any local
+      // model id (e.g. `ollama:qwen2.5:14b`, `ollama:llama3.1:8b`). The list
+      // below is a curated entry point for the wizard, not an allow-list.
+      // Matches openrouter.ts pattern (catalog spans many models).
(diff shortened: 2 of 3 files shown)


time-attack:
Thank you for the thorough write-up — the reasoning about Ollama's OpenAI-compatible `/v1/chat/completions` is accurate. That said, the ollama recipe is intentionally scoped to embeddings today (the resolver hint says so explicitly), so this is new provider capability surface rather than a bug fix, and `supports_tools: true` would be asserting tool-call behavior across whatever arbitrary model a user has pulled locally. We're not expanding provider chat surface right now, so I'm closing this one.

PR #1993, opened 2026-06-08
fix(cycle): stamp last_full_cycle_at for the source resolved from brainDir

### LINKED ISSUE #1992: Autopilot keeps cycle_freshness stale: inline cycle never stamps last_full_cycle_at
## Summary
On a brain maintained by `gbrain autopilot`, `gbrain doctor`'s `cycle_freshness` check goes stale (warn at 6h, fail at 24h) even though the autopilot runs a full cycle every interval. The autopilot keeps sync/extract/embed fresh, but the per-source cycle-completion timestamp is never advanced.

## Cause
`runCycle` (`src/core/cycle.ts`) writes `sources.config.last_full_cycle_at` only when `opts.sourceId` is set:

```ts
if (opts.sourceId && engine && !dryRun && (status === 'ok' || 'clean' || 'partial')) {
  await engine.updateSourceConfig(opts.sourceId, { last_full_cycle_at: ... });
}
```

But the autopilot's inline cycle (`src/commands/autopilot.ts`) calls `runCycle({ brainDir: repoPath, ... })` with **no** `opts.sourceId`, so the stamp is skipped on every cycle and `cycle_freshness` reports the brain stale.

Notably `runCycle` already computes `cycleSourceId = opts.sourceId ?? resolveSourceForDir(engine, brainDir)` and uses it for the cycle lock + phase scoping — the stamp is the one place still keyed off raw `opts.sourceId`.

## Expected
A cycle that runs against a resolvable source (explicit `sourceId`, or `brainDir` matching a source's `local_path`) advances `last_full_cycle_at` for that source, so `cycle_freshness` stays fresh while the autopilot runs.

## Fix
Key the stamp off `cycleSourceId` (the already-resolved id) instead of raw `opts.sourceId`. PR incoming.

100menotu001:
## Problem
`gbrain doctor`'s `cycle_freshness` goes stale on autopilot-maintained brains even though the autopilot cycles every interval. `runCycle` only stamps `last_full_cycle_at` when an explicit `opts.sourceId` is passed, but the autopilot's inline cycle sets `brainDir` and no `sourceId`. Details in #1992.

## Fix
`runCycle` already resolves `cycleSourceId = opts.sourceId ?? resolveSourceForDir(engine, brainDir)` and uses it for the cycle lock + phase scoping. This keys the `last_full_cycle_at` stamp off that same id instead of raw `opts.sourceId`:
- autopilot inline cycle (brainDir set, no sourceId) → resolves + stamps the source ✓
- `dream --source X` (explicit sourceId) → unchanged ✓
- no checkout + no sourceId → `cycleSourceId` undefined → no stamp (unchanged) ✓

No change to cycle scope — `cycleSourceId` is the same id the cycle already locked + scoped its phases to.

## Tests
`test/cycle-last-full-cycle-at.test.ts`: the prior "legacy caller (no sourceId) does NOT write" case encoded the bug (it seeded a source at `brainDir`). It now asserts the resolved-from-`brainDir` source IS stamped, plus a new case that a `brainDir` matching no source still does not write.

Closes #1992

### DIFF
--- a/src/core/cycle.ts
+++ b/src/core/cycle.ts
@@ -2190,24 +2190,30 @@ export async function runCycle(
   const status = deriveStatus(phaseResults, totals);
 
   // v0.38 (codex r1 P0-5): persist per-source cycle completion timestamp
-  // when the cycle ran successfully against an explicit source. Read by
-  // autopilot's per-source freshness gate next tick. Skipped when:
-  //   - opts.sourceId is unset (legacy callers — autopilot still here)
-  //   - engine is null (no-DB path)
+  // when the cycle ran successfully against a resolvable source. Read by
+  // autopilot's per-source freshness gate next tick.
+  //
+  // Uses `cycleSourceId` (opts.sourceId ?? the source resolved from brainDir),
+  // the SAME id the cycle locked + scoped its phases to — NOT raw opts.sourceId.
+  // The autopilot's inline cycle sets brainDir but passes no explicit sourceId,
+  // so keying the stamp off opts.sourceId alone never advanced last_full_cycle_at
+  // and `gbrain doctor` reported cycle_freshness stale even while the autopilot
+  // cycled every interval. Skipped when:
+  //   - no source resolves (engine null, or no checkout AND no opts.sourceId)
   //   - status is 'failed' or 'skipped' (don't mark a non-run as fresh)
   //   - dryRun (writes are out of scope)
   //
   // Best-effort: a write failure does NOT change the CycleReport status.
   // The cost of writing the wrong timestamp post-failure is higher than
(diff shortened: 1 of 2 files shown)


time-attack:
Thanks for chasing this down — the diagnosis in #1992 is correct and the gap is still on master. The blocker is that the patch predates the cycle split: that same write block now also stamps `last_source_cycle_at`, which became the gate for per-source autopilot dispatch, and `autopilot-global-maintenance` calls `runCycle` with `brainDir` and no `sourceId` while running only the brain-wide phase set. Keying the stamp off `cycleSourceId` would therefore mark a source's per-source cycle fresh after a global-only pass and suppress its next dispatch, so I'm closing this rather than landing it in its current shape.

PR #2000, opened 2026-06-09
fix: accept explicit LiteLLM embedding models

nahkhsinad:
## Summary
- Accept explicit model IDs for user-provided OpenAI-compatible embedding recipes like LiteLLM and llama-server
- Treat declared dimensions for those recipes as native schema sizing, not unsupported custom-dimension requests
- Add regression tests for LiteLLM diagnosis and LiteLLM/llama-server schema dimension resolution

## Test Plan
- `bun test test/litellm-embedding-config.test.ts`

## Real behavior proof
Before this patch, a valid config like `embedding_model=litellm:text-embedding-qwen3-embedding-0.6b` could be rejected during embedding diagnosis because the LiteLLM recipe has an empty static model list. During a real local LM Studio/LiteLLM-backed Documents ingest, this blocked a user-provided embedding model even though the provider and concrete model were configured. The new test pins the valid explicit-model path while preserving rejection for provider-only shorthand.

### DIFF
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -704,12 +704,16 @@ export function diagnoseEmbedding(modelOverride?: string): EmbeddingDiagnosis {
     };
   }
 
-  // Openai-compat recipes with empty models list require a user-provided model.
+  // OpenAI-compatible recipes with empty model lists require a concrete
+  // provider:model string, but an already-parsed non-empty modelId satisfies
+  // that contract. The old check rejected every litellm/llama-server model,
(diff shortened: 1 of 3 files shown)


time-attack:
Thank you for the detailed real-behavior write-up — the failure you hit was genuine. Both halves have since landed on master independently: the `user_provided_model_unset` guard in `gateway.ts` was removed as structurally unreachable, and `embedding-dim-check.ts` gained the #2271 passthrough tier that trusts declared dims for litellm/llama-server/ollama. Closing as already fixed; please re-test against master and reopen a fresh report if an explicit LiteLLM model is still rejected.

PR #2001, opened 2026-06-09
feat: emit import heartbeat progress

nahkhsinad:
## Summary
- Emit a periodic stderr heartbeat during long-running imports
- Report processed/total, imported/skipped/errors, chunks, active worker count, elapsed time, ETA, last successful file, and slow active files
- Add a unit test for the heartbeat formatter

## Test Plan
- `bun test test/import-heartbeat.test.ts`

## Real behavior proof
During a 30k-file Documents import backed by a LAN OpenAI-compatible embedding server, normal progress output stopped updating for hours while the parent process still appeared alive. This heartbeat gives operators enough state to tell slow embeddings from a stuck import without polluting stdout or JSON output.

### DIFF
--- a/src/commands/import.ts
+++ b/src/commands/import.ts
@@ -41,6 +41,56 @@ export interface RunImportResult {
   failures: Array<{ path: string; error: string }>;
 }
 
+export interface ImportHeartbeatState {
+  processed: number;
+  total: number;
+  imported: number;
+  skipped: number;
+  errors: number;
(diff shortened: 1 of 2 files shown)


time-attack:
Reviewed and declined for now: large-import operability additions are paused as a set. The design was sound; if big-corpus imports become a priority lane this is the first family to revisit. Thank you.

PR #2002, opened 2026-06-09
feat: add import planning summary

nahkhsinad:
## Summary
- Add `gbrain import --plan` / `--dry-run` to summarize a corpus before importing
- Report source, strategy, embedding mode, worker count, total files/bytes, and top extensions
- Keep JSON-mode progress chatter on stderr so the plan payload stays machine-readable
- Add a unit test for the planning summary

## Test Plan
- `bun test test/import-plan.test.ts`

## Real behavior proof
For a large Documents ingest, I had to write an external staging/planning script just to know what would be imported before spending embedding calls. This PR adds the core planning summary directly to import so operators can inspect large trees first.

### DIFF
--- a/src/commands/import.ts
+++ b/src/commands/import.ts
@@ -1,6 +1,6 @@
 import { readdirSync, lstatSync, existsSync } from 'fs';
 import { execFileSync } from 'child_process';
-import { join, relative } from 'path';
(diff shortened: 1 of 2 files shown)


time-attack:

PR #2003, opened 2026-06-09
feat: write import manifest records

nahkhsinad:
## Summary
- Add `gbrain import --manifest path.jsonl` for append-only per-file import audit records
- Record imported/skipped/error status, source id, slug, chunks, errors, and duration
- Add a unit test for JSONL manifest appends and parent directory creation

## Test Plan
- `bun test test/import-manifest.test.ts`

## Real behavior proof
During a large Documents import, the only durable failure trail was stderr plus an external script. A JSONL manifest gives operators a machine-readable record of what imported, skipped, or failed so large imports can be resumed and audited without scraping terminal output.


time-attack:

PR #2004, opened 2026-06-09
feat: support auto import workers

nahkhsinad:
## Summary
- Allow `gbrain import --workers auto` to opt into the shared auto-concurrency policy
- Keep the default import path serial for compatibility
- Add a regression test that pins the CLI wiring to `autoConcurrency`

## Test Plan
- `bun test test/import-workers-auto.test.ts`

## Real behavior proof
For a large Documents import, I had to manually choose worker counts. Sync already has a shared auto-concurrency policy, but import could only accept explicit integers. This makes the operator path less footgunny without changing default serial behavior.


time-attack:

PR #2007, opened 2026-06-09
fix: jsonb config double-encode (cycle_freshness), autopilot lock liveness, extraction-lag versionTs floor, doctor papercuts

Jey2311:
Found while remediating a live postgres/Supabase brain (voyage-4-large, 2048d, ~990 pages) whose `gbrain doctor` reported a permanent `cycle_freshness` FAIL plus a wall of warnings. Five independent fixes; all validated against that live brain (patched doctor: 75/100, 0 FAIL, 0 `unknown check name` warnings).

## Fixes

**1. `addSource` stored federated `config` as a double-encoded JSONB string → permanent `cycle_freshness` FAIL.**
Both INSERT paths used `executeRaw($4::jsonb, [JSON.stringify(config)])`. Through postgres-js the stringified param + `::jsonb` cast double-encodes into a JSONB **string scalar** (`"{\"federated\":true}"`), not an object. Every per-cycle `config || {last_full_cycle_at}` merge then produced a JSONB **array** instead of setting the key, so `config->>'last_full_cycle_at'` read null forever and `cycle_freshness` FAILed for every federated source (the `default` source, created via an inline jsonb literal, was unaffected). Routed both INSERTs through `executeRawJsonb` (the gbrain#1861 contract). `check-jsonb-pattern.sh` missed this because it scans for `${..}::jsonb` *interpolation*, not a `JSON.stringify` *parameter*.

**2. Autopilot lock liveness — systemd flap.**
The lock check yielded on a fresh mtime alone, never verifying the stored pid is alive. A SIGTERM'd autopilot (`systemctl restart`) leaves a fresh-mtime lock with a dead pid, so every relaunch `exit(0)`s and the unit flaps in `activating (auto-restart)`. Now probes the holder pid with `process.kill(pid, 0)` and takes over when it's dead. (Reproduced twice on the live host.)

**3. `links_extraction_lag` pinned near 100% on any pre-versionTs corpus.**
`extract --stale` stamps `links_extracted_at = page.updated_at` (a deliberate D4 race fix). But the staleness predicate also re-stales when `links_extracted_at < LINK_EXTRACTOR_VERSION_TS`. Any page whose content predates versionTs has `updated_at < versionTs`, so the stamp lands below it and the version arm re-flags the page forever — the sweep re-processes the same pages every run. Verified live: 956/991 pages, `updated_at < 2026-05-31`, never cleared. Now stamps `max(updated_at, versionTs)`, which records that the current extractor version processed the page and preserves D4 race-safety. Live: 956 → 0 stale. +regression test.

**4. doctor `embedding_column_registry` recommended an impossible HNSW DDL.**
For a `vector` column wider than pgvector's 2000-dim HNSW cap it told operators to run `CREATE INDEX ... vector_cosine_ops`, which always errors at >2000 dims (gbrain already skips that index by policy). Now explains seq scan is expected and points to the halfvec migration path. Actually enabling a halfvec index (column migration + recall eval) is left to coordinate with `vector-search-max-pool` rather than landing a competing change.

**5. doctor-categories drift for onboard checks.**
The 7 onboard remediation checks (`embed_staleness`, `entity_link_coverage`, `timeline_coverage`, `takes_count`, `type_proliferation`, `pack_upgrade_available`, `dangling_aliases`) flow through `categorizeCheck` but were uncategorized → once-per-run stderr warn + `meta` fallback. Added them and extended the drift-guard test to also scan `onboard/checks.ts` (both directions of the guard pass).

## Tests
- doctor-categories drift guard: 8/8
- sources-ops + autopilot-lock-path: 41/41
- extract-stale (incl. new regression): 15/15
- vector-index + embedding-column: 24/24
- typecheck clean on all changed files

`docs/designs/doctor_config_lock_fixes.md` documents the root-cause analysis and what's deferred.


Jey2311:
Added a 4th commit (552d9d0f) — a 6th fix surfaced from the same live brain:

**`page_links` view is SECURITY DEFINER (Supabase Critical lint).** Migration v86 created the `page_links` alias view without `security_invoker`, and gbrain sets it on no views anywhere, so on Postgres the view runs with the owner's privileges and bypasses the querying role's RLS on `links`. Live view matches source exactly → upstream defect, not drift. Added migration **v116** (`page_links_view_security_invoker`, idempotent `CREATE OR REPLACE ... WITH (security_invoker = true)`, runs on both engines, supersedes v86, fixes existing brains) plus the same option in `pglite-schema.ts`. PGLite (PG15+) accepts the option (verified). No behavior change for the engine's own service-role JOINs. v116 applies cleanly in the migrate test (`[116] ✓`).

PR #2014, opened 2026-06-09
feat(stdio-mcp): GBRAIN_SOURCES env var for multi-source federated reads

donogeme:
## What

Adds `GBRAIN_SOURCES` (plural, comma-separated) env var for stdio MCP, so callers on a multi-source brain can read across every listed source.

## Why

Stdio MCP (the `gbrain serve` transport Claude Desktop / Code / Cursor speak) can only scope to ONE source via `GBRAIN_SOURCE` (singular). On a multi-source brain, every read scopes to `default` and pages in other sources come back `page_not_found` / empty. The HTTP transport already solves this — it populates `auth.allowedSources` from `oauth_clients.federated_read` and routes reads through `sourceScopeOpts(ctx)`. Stdio had no equivalent.

This bridges that gap: `GBRAIN_SOURCES` populates a synthetic `AuthInfo.allowedSources` for stdio callers, so the **existing** `sourceScopeOpts(ctx)` federated-read path (the same one HTTP uses) works on stdio too. No new scoping logic — it reuses what's already there.

## Behavior

- `GBRAIN_SOURCE` (singular) unchanged — scalar write-scope fallback for single-source setups.
- When both set, `GBRAIN_SOURCES` wins for reads; its first entry is the write-scope fallback when `GBRAIN_SOURCE` is unset.
- Parser trims whitespace, drops dupes, preserves first-seen order, treats empty input as "no federated scope" (falls through to scalar).
- Purely additive — brains that don't set `GBRAIN_SOURCES` behave exactly as before.

## Example

\`\`\`json
"gbrain": {
  "command": "gbrain",
  "args": ["serve"],
  "env": { "GBRAIN_SOURCES": "default,docc,popify,personal-brain" }
}
\`\`\`

## Tests

`test/stdio-mcp-allowed-sources.test.ts` — 10 cases pinning the parser contract (trim / dedupe / order / empty → scalar fallback). `bun run typecheck` clean.

## Context

Maintained as a fork-local patch (donogeme/gbrain) since v0.41.26; upstreaming so multi-source stdio users don't need a fork. Versioning/CHANGELOG left to maintainer discretion.

### DIFF
--- a/src/mcp/server.ts
+++ b/src/mcp/server.ts
@@ -8,6 +8,31 @@ import { buildToolDefs } from './tool-defs.ts';
 import { dispatchToolCall, validateParams, buildOperationContext } from './dispatch.ts';
 import { getBrainHotMemoryMeta } from '../core/facts/meta-hook.ts';
 
+/**
+ * v0.36.0.1: Parse `GBRAIN_SOURCES` (plural, comma-separated) into a normalized
+ * array of source ids for stdio MCP federated reads. Returns `undefined` when
+ * the env var is unset OR resolves to an empty list (so callers fall back to
+ * the scalar `GBRAIN_SOURCE` path).
+ *
+ * Exported for unit testing — the parsing rules (trim, drop empties, drop
+ * dupes) are the same regression-class as the federated_read array
+ * normalization. Drift here would silently re-introduce the docc-source-id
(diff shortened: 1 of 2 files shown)


time-attack:

PR #2016, opened 2026-06-09
feat(serve-http): honor source + federatedRead in admin register-client

guim4dev:
## Problem

`POST /admin/api/register-client` hardcodes `source_id='default'` and `federated_read=undefined`:

```ts
const result = await oauthProvider.registerClientManual(
  name, grants, scopeString, uris, 'default', undefined, validatedAuthMethod,
);
```

So a client minted **over HTTP** — by the admin SPA, or by a proxy that provisions per-user OAuth clients programmatically — can never be bound to a non-`default` brain source. Only the CLI (`gbrain auth register-client --source <id> --federated-read <a,b>`) can, because it calls `registerClientManual` with the real `sourceId` / `federatedRead` args.

Concretely: a downstream MCP→MCP proxy that mints a `client_credentials` client per user via this endpoint gets a client pinned to `source_id='default'`, so every read resolves to the `default` source regardless of which brain source actually holds the corpus. There is no HTTP path to fix it after the fact (only `update-client-ttl` / `revoke-client` exist).

## Change

Thread the already-supported `sourceId` / `federatedRead` parameters of `registerClientManual` through from the request body:

- `source` (string) → write `source_id`
- `federatedRead` (string[]) → read source set

Mirrors the CLI's `--source` / `--federated-read` flags.

**Backward-compatible:** omitting `source` keeps `source_id='default'`; omitting `federatedRead` keeps `federated_read=[source_id]` (the existing non-federated default). Both fields are validated against the canonical `source_id` regex (`src/core/source-id.ts`) and surface a structured `400 invalid_source` on bad input — no more 500-at-INSERT for a malformed id.

Validation lives in two small pure helpers (`normalizeSourceInput`, `normalizeFederatedReadInput`) next to the canonical `source_id` validators, following the `normalizeScopesInput` precedent.

## Tests

New hermetic unit test `test/register-client-source-normalize.test.ts` (15 cases): omitted→`'default'`, valid/invalid source ids, federated array shapes, empty-array and bad-element rejection.

```
bun test test/register-client-source-normalize.test.ts   # 15 pass
bun test test/source-id.test.ts                          # 19 pass (no regression)
bun test test/scope-normalize.test.ts                    # 28 pass (adjacent endpoint logic)
bun build src/cli.ts                                     # builds clean
```

(Full e2e in `test/e2e/serve-http-oauth.test.ts` needs a `DATABASE_URL`; not run locally.)

### DIFF
--- a/src/commands/serve-http.ts
+++ b/src/commands/serve-http.ts
@@ -28,6 +28,7 @@ import type { OperationContext, AuthInfo } from '../core/operations.ts';
 import { GBrainOAuthProvider, validateTokenEndpointAuthMethod } from '../core/oauth-provider.ts';
 import type { SqlQuery } from '../core/oauth-provider.ts';
 import { hasScope, ALLOWED_SCOPES_LIST, normalizeScopesInput } from '../core/scope.ts';
+import { normalizeSourceInput, normalizeFederatedReadInput } from '../core/source-id.ts';
 import { summarizeMcpParams, dispatchToolCall } from '../mcp/dispatch.ts';
 import { paramDefToSchema } from '../mcp/tool-defs.ts';
 import { getBrainHotMemoryMeta } from '../core/facts/meta-hook.ts';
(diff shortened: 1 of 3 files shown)


caviquioli (approved):

time-attack:

PR #2017, opened 2026-06-09
fix(schema): list all bundled packs

arisgysel-design:
## Summary
- make the schema CLI use the shared bundled-pack registry instead of a stale hardcoded two-pack list
- allow bundled packs such as gbrain-base-v2 and the lens packs to appear in schema list/show/use
- extend schema CLI and mutation tests for the bundled successor pack

## Root Cause
The repo already shipped gbrain-base-v2 and the loader/mutation surfaces knew about it, but src/commands/schema.ts kept its own stale bundled-pack list/path resolver. This made official docs/onboard recommendations point to a pack that schema show/use reported as unknown.

## Validation
- bun test test/schema-cli.test.ts
- bun test test/schema-pack-mutate.test.ts
- bun test test/schema-pack-find-pack-successors.serial.test.ts test/onboard-pack-upgrade-checks.test.ts
- bun run typecheck
- live CLI smoke: gbrain schema list, gbrain schema show gbrain-base-v2, isolated GBRAIN_HOME gbrain schema use gbrain-base-v2

### DIFF
--- a/src/commands/schema.ts
+++ b/src/commands/schema.ts
@@ -41,6 +41,7 @@ import {
   setExpertRoutingOnType,
   setExtractableOnType,
   UnknownPackError,
+  BUNDLED_PACK_NAMES,
   updateTypeOnPack,
   __setPackLocatorForTests,
   _resetPackLocatorForTests,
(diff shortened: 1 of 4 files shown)


arisgysel-design:
Closing as superseded: current master now uses the shared BUNDLED_PACK_NAMES registry across schema list/show/use and the mutation guard, so the original defect is already resolved upstream.

PR #2022, opened 2026-06-09
fix(resolver): tolerate CRLF line endings in SKILL.md frontmatter parsing

joshwilks111-max:
## Problem

The frontmatter-fence and trigger-block regexes in `src/core/check-resolvable.ts` (`extractTriggers`) and `src/core/skill-frontmatter.ts` (`parseSkillFrontmatter` / `parseArrayField`) require bare-LF line endings. On Windows checkouts (`core.autocrlf=true` is the default), SKILL.md files and the committed test fixtures carry CRLF in the working tree, so:

- frontmatter silently fails to parse → `triggers:` arrays vanish
- `gbrain check-resolvable` / `gbrain doctor` report bogus `unreachable` + `mece_gap` issues against perfectly valid skill trees (reproduced live: a 52-skill tree that validates clean on LF reports 28 issues when files are CRLF)
- **5 existing tests fail out of the box on Windows** (`analyzeSkillBrainFirst` fixture-corpus tests in `test/skill-brain-first.test.ts`) — CI never sees this because Linux checkouts are LF

## Fix

CRLF-tolerant parsing, no behavior change for LF content:

- fence match `/^---\n…/` → `/^---\r?\n…/`
- trigger/list block matchers allow `\r?\n` at boundaries and exclude `\r` from item captures
- line splits use `/\r?\n/` instead of `'\n'`

One subtlety: `parseArrayField`'s `blockRe` is built via a template literal passed to `new RegExp()`, so the escapes are doubled (`\s`, `\r?\n`) — a single-backslash `\s` in a template literal silently degrades to the letter `s`.

## Tests

Adds CRLF regression tests for both parsers:

- `test/skill-brain-first.test.ts` — `parseSkillFrontmatter` parses a CRLF document identically to LF (name, booleans, block lists, triggers)
- `test/check-resolvable.test.ts` — `checkResolvable` on a fixture whose SKILL.md is CRLF yields 0 unreachable and no `mece_gap`

Verified both directions: with the fix reverted, the new tests fail alongside the 5 pre-existing Windows failures (7 fail / 87 pass); with the fix, the full files pass (94 pass / 0 fail). `bun run typecheck` clean.

### DIFF
--- a/src/core/check-resolvable.ts
+++ b/src/core/check-resolvable.ts
@@ -219,13 +219,13 @@ export function parseResolverEntries(resolverContent: string): ResolverEntry[] {
 
 /** Simple YAML frontmatter parser — extracts triggers array if present. */
 function extractTriggers(skillContent: string): string[] {
-  const fmMatch = skillContent.match(/^---\n([\s\S]*?)\n---/);
+  const fmMatch = skillContent.match(/^---\r?\n([\s\S]*?)\r?\n---/);
   if (!fmMatch) return [];
   const fm = fmMatch[1];
-  const triggersMatch = fm.match(/^triggers:\s*\n((?:\s+-\s+.+\n?)*)/m);
+  const triggersMatch = fm.match(/^triggers:\s*\r?\n((?:\s+-\s+.+(?:\r?\n|$))*)/m);
   if (!triggersMatch) return [];
   return triggersMatch[1]
-    .split('\n')
+    .split(/\r?\n/)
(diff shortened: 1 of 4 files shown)


time-attack:
Thanks for the thorough repro — the Windows CRLF class is real and CI genuinely can't see it. The `check-resolvable` half already landed on master via #1149, which normalizes CRLF→LF once before parsing rather than making each regex CRLF-aware, so this diff conflicts with the approach that shipped. Closing on that basis; what's left in `skill-frontmatter.ts` is the same one-line normalize at the top of the parser rather than the regex rewrite here.

PR #2023, opened 2026-06-09
fix(cli): route thin-client CLI-only commands before DB connect

WillDent:
Fixes thin-client CLI-only commands that were falling through to local connectEngine() and reporting missing database_url even though thin-client installs intentionally have no local DB.

Changes:
- route capture, jobs list/get, recall/forget, and config show before engine connect when remote_mcp is configured
- cleanly refuse DB-bound host-only CLI commands on thin clients with actionable hints
- coerce remote job timestamp strings back to Date before rendering
- extend thin-client routing audit coverage for the newly refused commands and remote job date coercion

Verification:
- bun run typecheck
- bun test test/thin-client-routing-audit.test.ts test/cli-dispatch-thin-client.test.ts test/capture-runcapture.test.ts
- live thin-client smoke: capture, jobs list, recall --today, config show, features refusal

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -892,6 +892,11 @@ function formatResult(opName: string, result: unknown): string {
 const THIN_CLIENT_REFUSED_COMMANDS = new Set([
   'sync', 'embed', 'extract', 'extract-conversation-facts', 'enrich', 'migrate', 'apply-migrations',
   'repair-jsonb', 'orphans', 'integrity', 'serve',
+  // Thin clients have no local engine. Refuse DB-bound CLI-only commands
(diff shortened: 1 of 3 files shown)


time-attack:
Thanks for the careful audit and the live smoke run. The `jobs` list/get pre-engine branch already landed on master, so that part conflicts, and most of the rest expands the thin-client CLI surface (17 new refusal hints, additional routed commands via `null as never` engine casts) rather than repairing a break. The one piece here that fixes a real crash is the remote job date coercion — `formatJob` calls `.toISOString()` on strings coming back from `list_jobs` — and that would be far easier to evaluate on its own; closing this one as-is.

Issue #223, opened 2026-04-19
PGLite WASM crash on macOS 26.3 with Bun 1.3.11


AndreLYL:
## Bug Description

PGLite-based commands crash immediately with `Aborted(). Build with -sASSERTIONS for more info.` on macOS 26.3 (Darwin kernel 25.3.0) with Bun 1.3.11 and PGLite 0.4.4.

## Steps to Reproduce

1. Upgrade macOS to 26.3 (Darwin kernel 25.3.0, arm64)
2. Run any gbrain CLI command that needs PGLite:
   - `gbrain serve`
   - `gbrain export`
   - `gbrain health`
   - `gbrain query`

All crash with the same error:
```
RuntimeError: Aborted().
  at abort (node_modules/@electric-sql/pglite/dist/index.js:1:77811)
  at invoke_viiiiii
  at async create
```

## Environment

- macOS 26.3 (Darwin 25.3.0, arm64)
- Bun 1.3.11
- gbrain 0.10.1
- PGLite 0.4.4

## Key Observation

Running `gbrain serve` processes started **before** the macOS upgrade continue to work. The issue only affects **new** PGLite instances created after the upgrade.

## Analysis

This suggests a WASM compatibility issue with macOS 26.3. The crash happens during PGlite initialization (`PGlite.create()` with extensions: {vector, pg_trgm}), before any database operations. The `postmaster.pid` file confirms Postgres fails to start inside WASM.

## Possible Fixes

1. Pin to an older PGLite version known to work on macOS 25.x
2. Add explicit WASM SIMD/compatibility flags
3. Investigate macOS 26.3 changes affecting WASM syscalls

devlsh:
getting this in linux now after upgrading - clearing `~/gbrain` and `~/.gbrain`, setting it up again and embedding etc. works for a short time, but then this issue comes back and doesn't go away

papiofficial:
**Update (v0.19.0 regression):** After upgrading from v0.18.2 → v0.19.0, this went from intermittent to **consistently broken** on macOS 25.3 arm64 / Bun 1.3.12. In v0.18.2, some commands like `gbrain embed --all` would successfully initialize PGLite and run. After the v0.19 upgrade, every single PGLite command fails immediately with the WASM crash — no exceptions. Filed as #391. Environment: Darwin 25.3.0 arm64, Bun 1.3.12, PGLite 0.4.3.

WattyouSay:
[...]

PR #2024, opened 2026-06-09
fix: route Windows PGLite init diagnostics

### LINKED ISSUE #2008: Windows: username with apostrophe in path breaks gbrain migration — NativeCommandError + WASM/module errors are red herrings
## Summary

If your Windows username contains an apostrophe (e.g. `Arnuv'`), gbrain migrations fail with confusing errors. The path `C:\Users\Arnuv'\` breaks PowerShell string parsing mid-command, causing downstream failures that look like bun issues, missing modules, or PGLite WASM crashes — none of which are the real cause.

## Environment

- OS: Windows 11 Pro 10.0.26200
- Shell: PowerShell 5.1
- gbrain: v0.42.x (PGLite)

## Root cause

The migration script runs bun inside a PowerShell context with the home path embedded. An apostrophe in the username (`'`) terminates the PowerShell string prematurely, corrupting the command before bun even runs.

## Errors seen (all red herrings)

- `Cannot find module '@electric-sql/pglite/vector'`
- `PGLite failed to initialize its WASM runtime. Aborted()`
- `NativeCommandError` in PowerShell
- Migration WEDGED after 3 consecutive failures with no clear recovery path

Example from `$env:TEMP\gbrain-migrate.txt`:

```
bun :
At line:1 char:31
+ ... '\.gbrain"; bun run src/cli.ts apply-migrations --yes > "$env:TEMP\gb...
+                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError

Migration v0.12.0 is WEDGED (3+ consecutive partials with no completion).
Phase A (schema) failed: PGLite failed to initialize its WASM runtime.
    Most common cause: the macOS 26.3 WASM bug
    Original error: Aborted(). Build with -sASSERTIONS for more info.
```

Note: the error blames a macOS bug — the user is on Windows 11.

## Fix

Fresh install to a path without the apostrophe. The `--force-retry` recovery instructions shown in the WEDGE message do not help while the path issue exists.

## Impact

Affects any Windows user whose username contains `'` — common in Irish, French, and Italian surnames (O'Brien, O'Connor, D'Angelo, etc.). All error messages point away from the real cause, making self-diagnosis very difficult.

Related: garrytan/gstack#1731 (Windows shell spawning), garrytan/gbrain#1605 (Windows migration subprocess failures)

maxpetrusenkoagent:
## Summary
- add Windows-aware PGLite init-error verdicts so Windows failures no longer fall through to the macOS 26.3 / #223 hint
- route the #2008 PowerShell / NativeCommandError + apostrophe-path shape to a Windows path-quoting message that names #2008
- keep generic Windows WASM / `Aborted()` / missing-module failures on a Windows-generic path instead of prescribing the apostrophe workaround

Closes #2008.

## Scope note
PR #1554 already covers the package.json postinstall shell shim. This PR intentionally avoids that surface and fixes the remaining diagnostic-routing failure so Windows users do not get pointed at the wrong macOS issue.

## Verification
- RED: `bun test test/pglite-init-classifier.test.ts` initially failed because the #2008 fixture routed to `macos-26-3`
- GREEN: `bun test test/pglite-init-classifier.test.ts test/fix-wave-structural.test.ts`
- `bun run typecheck`
- `bun run build`
- Autoreview: Codex found an overbroad Windows `Aborted()` classifier and a privacy issue in the fixture; both were fixed. Final Codex review found no blockers.

### DIFF
--- a/src/core/pglite-engine.ts
+++ b/src/core/pglite-engine.ts
@@ -170,9 +170,8 @@ export function classifyPgliteInitError(
 ): PgliteInitFailure {
   if (/\$\$bunfs|ENOENT[\s\S]*pglite\.data/i.test(message)) return 'bunfs';
   if (platform === 'win32') {
-    const hasShellEvidence = /NativeCommandError|FullyQualifiedErrorId\s*:\s*NativeCommandError|At line:\d+ char:\d+|CategoryInfo\s*:|PowerShell/i.test(message);
     const hasWindowsApostrophePath = /[A-Z]:\\[^\n\r]*'[^\n\r]*/i.test(message);
-    if (hasShellEvidence || hasWindowsApostrophePath) return 'windows-path-shell';
+    if (hasWindowsApostrophePath) return 'windows-path-shell';
     if (/abort.*runtime|wasm.*runtime|Aborted\(\)|Cannot find module ['"]?@electric-sql\/pglite/i.test(message)) {
       return 'windows-pglite-unknown';
     }
--- a/src/core/pglite-engine.ts
+++ b/src/core/pglite-engine.ts
@@ -170,8 +170,9 @@ export function classifyPgliteInitError(
 ): PgliteInitFailure {
   if (/\$\$bunfs|ENOENT[\s\S]*pglite\.data/i.test(message)) return 'bunfs';
   if (platform === 'win32') {
+    const hasShellEvidence = /NativeCommandError|FullyQualifiedErrorId\s*:\s*NativeCommandError|At line:\d+ char:\d+|CategoryInfo\s*:|PowerShell/i.test(message);
     const hasWindowsApostrophePath = /[A-Z]:\\[^\n\r]*'[^\n\r]*/i.test(message);
-    if (hasWindowsApostrophePath) return 'windows-path-shell';
+    if (hasShellEvidence && hasWindowsApostrophePath) return 'windows-path-shell';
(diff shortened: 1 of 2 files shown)


maxpetrusenkoagent:
Verification update after exact-PR-diff autoreview:

- `bun test test/pglite-init-classifier.test.ts test/fix-wave-structural.test.ts` → 35 pass, 0 fail, 95 expect() calls
- `bun run typecheck` → passed
- `bun run build` → passed, compiled `bin/gbrain`
- Final Codex autoreview of `gh pr diff 2024` → BLOCKERS: none

Reviewer-blocker fixes handled in this branch:
- Windows `abort/runtime` and `Aborted()` no longer fall through to the macOS #223 hint on `win32`.
- `windows-path-shell` now requires both shell evidence and an apostrophe-bearing Windows path.
- Shell-only and apostrophe-only fixtures stay on the generic Windows PGLite hint.
- Test fixtures use synthetic names only (`quote-test`, `quote-safe`).

time-attack:
Thanks for the careful classifier work and for keeping it off PR #1554's surface. This is still a draft and it changes which diagnostic hint is printed rather than fixing the underlying Windows PGLite init failure, so it doesn't clear the bar for landing error-message routing plus 100 lines of classifier fixtures. Closing; the actual Windows install fix is the surface worth pursuing.

PR #2026, opened 2026-06-10
Update OpenClaw deployment link in README

timgl:
The current link didn't actually work


time-attack:
Thanks for flagging this. Both the old and new URLs currently resolve for me, so I can't confirm the breakage from here, and the patch repoints the button at a different third-party template repo while the link text still reads "AlphaClaw on Render" — so the label and target no longer agree. Closing for now; a report with the exact error the old link produced would be the useful next step.

PR #2029, opened 2026-06-10
fix(schema): make bundled pack inspection truthful

JiraiyaETH:
## Problem

`gbrain schema` inspection lied about bundled packs in two ways:

1. **`parseYamlMini` block-scalar bug** — a `description: |` block scalar swallowed the keys that followed it, so `gbrain-recommended` loaded with effectively **zero page types** even when configured as the active pack. `schema-author` and anything downstream of the active pack was unsafe to rely on.
2. **Drifted bundled-pack lists** — the CLI `schema list/show` path and the MCP `list_schema_packs` operation each carried their own idea of which packs ship bundled, and both had drifted from the real bundled pack files.

## Fix

- Added a single bundled-pack registry (`src/core/schema-pack/bundled.ts`) and pointed the CLI list/show path, the active-pack loader, the mutation readonly guard, and the MCP list operation at it.
- Fixed `parseYamlMini` block-scalar handling so `description: |` no longer swallows following keys like `page_types`.

## Tests

Regression coverage added (79 lines in `schema-pack-loader.test.ts`, 39 in `schema-cli.test.ts`):

- block scalar parsing
- `gbrain-recommended` page types actually loading
- `schema list` showing all bundled packs
- `schema show/validate gbrain-recommended` and `schema show gbrain-base-v2`
- MCP `list_schema_packs` exposing all bundled packs

`bun test test/schema-pack-loader.test.ts test/schema-cli.test.ts test/operations-schema-pack.test.ts` → 78 pass / 0 fail. `bun run typecheck` → pass.

Verified live: `gbrain schema list` shows all seven bundled packs; `gbrain schema active` shows `gbrain-recommended` with its 12 page types.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2030, opened 2026-06-10
feat(extract): schema-pack frontmatter_links drive extraction, with D4 fail-empty semantics

JiraiyaETH:
## Summary

- Fixes D4 fail-empty semantics for pack-fed frontmatter link extraction.
- Uses the shared `loadActivePackBestEffort` path from `extract.ts` instead of an open-coded loader.
- Preserves legacy `FRONTMATTER_LINK_MAP` fallback only when no schema pack is configured or the configured/active pack declares zero `frontmatter_links`.
- Skips frontmatter link extraction, with a loud stderr warning, when a configured schema pack fails to load so extraction never silently emits legacy-verb edges against packed-out user intent.

## Why this matches Garry's migration pattern

This completes the follow-through described in `TODOS.md`:

- `T1.5 follow-through` says runtime call sites should consume pack-aware type/filter helpers through the shared best-effort loader rather than retaining hardcoded defaults.
- `TODO-1 (P2)` repeats the same migration shape for the later `--by-mention` gazetteer: use `loadActivePackBestEffort(ctx)` first, then fall back only for unconfigured/back-compat cases.

The important contract is D4: pack load failure is not a signal to resurrect hardcoded legacy behavior. It is a fail-empty condition for that pack-fed extraction surface.

## Notes

The existing `extractStaleFromDB` `skipFrontmatter` threading is intentionally retained. It keeps stale extraction's frontmatter pass behind the same explicit `includeFrontmatter` gate while still allowing body/wikilink extraction to proceed.

## Verification

- `bun test test/link-extraction.test.ts`
- `bun test test/schema-pack-best-effort.test.ts`
- `bun test test/schema-pack-load-active.serial.test.ts`
- `bun test test/put-page-provenance.test.ts`
- `bun run typecheck`

> Stacked on #2029 (`fix(schema): make bundled pack inspection truthful`) — the first commit here is that PR; review the second commit (`fix(extract): ...`) plus the base wiring commit for this change.

### DIFF
--- a/src/commands/extract.ts
+++ b/src/commands/extract.ts
@@ -61,6 +61,9 @@ import { createHash } from 'crypto';
 // shared sliding-pool helper + PGLite-clamp wrapper.
 import { runSlidingPool } from '../core/worker-pool.ts';
 import { parseWorkers, resolveWorkersWithClamp } from '../core/sync-concurrency.ts';
+import { loadActivePack } from '../core/schema-pack/load-active.ts';
+import { loadConfig } from '../core/config.ts';
+import type { SchemaPackManifest } from '../core/schema-pack/manifest-v1.ts';
 
 // Batch size for addLinksBatch / addTimelineEntriesBatch.
 // Postgres bind-parameter limit is 65535. Links use 4 cols/row → 16K hard ceiling;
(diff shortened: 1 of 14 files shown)


time-attack:
Thanks for the careful write-up and the D4 reasoning — closing this one as superseded by your own #2040, which contains this change set plus the flag on top, so there is no point reviewing the same wiring twice. To be straight with you: #2040 is also being closed, because reshaping frontmatter-link extraction semantics (fail-empty on pack load failure, dropping the legacy fallback) is new feature surface on a hot path, and we are only taking bug/correctness/security fixes right now. If you hit a concrete case where extraction emits a *wrong* edge today, a small PR fixing just that case would be very welcome.

PR #2032, opened 2026-06-10
fix(cli): arm the disconnect hard-deadline at teardown, not before the op handler (#1775)

### LINKED ISSUE #1775: search/query render no output + engine.disconnect() 10s force-exit on 0.42.8.0 (Postgres brain); 0.22.8 works
## Summary

On gbrain **0.42.8.0** (installed via `bun install -g github:garrytan/gbrain`, `main` HEAD), `gbrain search <q>` and `gbrain query <q>` print **no results** and exit with:

```
[cli] engine.disconnect() did not return within 10000ms — force-exiting
```

The **same brain DB** queried with gbrain **0.22.8** returns correct ranked results. Only `search`/`query` are affected — `list`, `get`, `sync`, `doctor`, `status` all work and render output normally on 0.42.8.0.

## Environment

- gbrain **0.42.8.0** (`bun install -g github:garrytan/gbrain`)
- Runtime: **Bun 1.3.14** on **WSL2 / Ubuntu 24.04**
- Brain: **Postgres** (remote), schema **v112** (migrated up from v29 by 0.42.8.0)
- Embedding model: `zeroentropyai:zembed-1`

## Repro

```
$ gbrain search trak
[cli] engine.disconnect() did not return within 10000ms — force-exiting
$            # no results printed
```

## Expected

Ranked keyword results, as **gbrain 0.22.8** produces against the identical DB:

```
$ gbrain search trak
[0.4493] companies/shiftcare -- # ShiftCare ...
[0.3559] products/trak/trak -- ...
... (7 hits)
```

## Evidence the data + index are healthy (the regression is in the search/query CLI path, not the data)

- `gbrain list -n 5` returns pages including `products/trak/trak`.
- `gbrain doctor` reports all checks OK; embeddings present; no stale chunks.
- Direct SQL against the same DB confirms FTS is fully populated:
  - `SELECT count(*) FROM content_chunks WHERE search_vector @@ to_tsquery('english','trak')` → **19**
  - `SELECT count(*) FROM pages WHERE search_vector @@ to_tsquery('english','trak')` → **9**
- gbrain **0.22.8** renders the results correctly against the same DB, captured the same (non-TTY) way.

## Notes / narrowing

- **Not stdout buffering.** The empty output persists under a PTY (`script -qec 'gbrain search trak' /dev/null`), so it isn't results lost to an unflushed buffer at force-exit.
- The `engine.disconnect() did not return within 10000ms` line appears **only** on `search`/`query`, never on `list`/`status`/`doctor`/`sync` — suggesting the search/query code path opens a connection/pool that never closes, and either (a) the process force-exits before printing results, or (b) the query returns empty due to a scoping/RLS change introduced in 0.4x.

Happy to provide more detail, logs, or test a patch.

pabloglzg:
Fixes #1775.

## The bug

The 10s "disconnect hang guard" in `src/cli.ts` is armed at the top of the op-dispatch path, **before the op handler runs**, and only cleared in the `finally` after `engine.disconnect()` returns. That makes it accidentally a **whole-run deadline, not a disconnect guard**: any shared op whose total wall time exceeds 10s gets `process.exit(process.exitCode ?? 0)` mid-handler. The user sees empty stdout, exit 0, and a misleading `engine.disconnect() did not return within 10000ms` warning for a disconnect that never ran.

This is why #1775 reproduces only on `search`/`query`: they're the shared ops that routinely exceed 10s (remote Postgres over WAN, reranker passes, LLM query expansion), while `list`/`get`/`status` finish fast, and bulk commands (`embed`, `sync`) dispatch through the CLI-only path that never hits this timer. Root cause was diagnosed exactly by @danashburn in the issue thread — this PR is that one-move fix, plus a regression test.

Note: #2015 (`endPoolBounded`) fixes a *different* bug in the same blast radius — a genuinely hung `pool.end()` against PgBouncer. It doesn't touch this timer; with #2015 merged, an op handler slower than 10s still gets killed mid-run. The two compose.

## The fix

Arm the timer in the `finally`, immediately before `drainAllBackgroundWorkForCliExit()` + `engine.disconnect()` — the teardown it exists to bound (its warning message already says so). The C13 defense-in-depth contract ("a hung disconnect cannot defeat the force-exit") is fully preserved: the timer is armed synchronously at teardown start, before the drain/disconnect awaits. This also brings the op-dispatch path to the same shape `handleCliOnly`'s teardown has used since v0.42.20.0 (#1762) — the pre-handler placement was the outlier, not the design.

**Trade-off, stated honestly:** a hung op *handler* is no longer killed at 10s. There is deliberately no uniform whole-run wallclock after this change — handlers are bounded piecemeal by the layers underneath (6s query-embed deadline in hybrid.ts, 60s/300s AI gateway deadlines via `withDefaultTimeout`, Postgres `statement_timeout` default 5min), so the worst-case wedge for a pathological case (e.g. client-side TCP blackhole mid-query) moves from 10s to ~minutes. The old behavior was strictly worse for that case too: it reported the failure as an exit-0 success with empty output. If a uniform backstop is wanted for unattended callers, a generous (minutes-scale) pre-handler deadline with a distinct warning and nonzero exit would be a separate, additive change — happy to file it.

## Tests

`test/cli-disconnect-deadline-placement.test.ts` — source-shape regression test pinning that the arming sits after `op.handler(...)` and before the drain, with exactly one arming site in `main()`. Placement is the load-bearing property, so this follows the pattern the repo endorses in `test/fix-wave-structural.test.ts` (a behavioral repro would need a >10s op against a live engine just to observe the timer fire).

Also updates the one `KEY_FILES.md` phrase that described the old whole-run behavior ("hanging past the 10s force-exit") to the current truth.

## Verification

- New test + `test/cli-should-force-exit.test.ts`, `test/cli.test.ts`, `test/cli-options.test.ts`, `test/cli-dispatch-thin-client.test.ts`, `test/fix-wave-structural.test.ts`, `test/build-llms.test.ts`: all pass. The placement test fails on current master, passes with the fix.
- `bun run typecheck`: clean.
- Manually verified on a Postgres brain with a local reranker where `gbrain query` takes 13–28s: before the fix, empty stdout + exit 0 + the misleading warning at exactly 10s; after, full ranked results print and the process exits cleanly.


time-attack:
Thanks for this, @pabloglzg! 🙏

What this does: PR #2032 attempts to fix issue #1775 by moving a 10-second force-exit timer from before the operation handler to the teardown phase, preventing long search/query operations from being killed mid-execution. The root cause has already been…

PR #2033, opened 2026-06-10
fix(gateway): carry asymmetric input_type across the AI SDK to the wire body (supersedes #1400)

pabloglzg:
Supersedes #1400 (closed unmerged) — same AsyncLocalStorage mechanism @billy-armstrong proposed, extended to cover Voyage and every openai-compatible recipe (llama-server, litellm, ollama, …). Full credit to him for the original diagnosis.

## The bug

```
embedQuery(text)
  → embed([text], {inputType: 'query'})
  → dimsProviderOptions(..., 'query')
  → {openaiCompatible: {dimensions: N, input_type: 'query'}}   ✓ threaded
  → AI SDK embedMany / openai-compatible adapter                ✗ input_type DROPPED
  → per-recipe fetch shim sees a body with no input_type
  → wire: input_type: 'document' (ZE default) / absent (everyone else)
```

The AI SDK's openai-compatible adapter validates `providerOptions.openaiCompatible` against a fixed schema (`dimensions`, `user`) and silently drops every other field before building the request body. So for asymmetric embedding models — where queries and documents are encoded with different prompts — **every query was encoded document-side**, and asymmetric retrieval silently collapsed toward surface-token overlap. The failure is invisible at the providerOptions layer: `test/asymmetric-encoding-contract.test.ts` pins that `input_type: 'query'` reaches the transport's providerOptions, and it stays green while the wire carries the wrong value. Embeddings remain valid vectors, hybrid search still returns results — they're just systematically worse, with the BM25 arm masking the vector-arm failure on keyword-y queries.

Affected consumers on master:
- **ZeroEntropy hosted** (`zeroentropyai:zembed-1`): the shim's `input_type = 'document'` default fires for queries too.
- **Voyage v3+** (`input_type` is part of Voyage's retrieval API): the field never reaches the wire at all.
- **llama-server / litellm / ollama** (local/proxy asymmetric models): no shim existed, so those deployments had no way to receive the signal.

## The fix

One mechanism, all consumers. A module-level `AsyncLocalStorage` (`__embedInputTypeStore`, mirroring the existing `__budgetStore` pattern in the same file) is populated in `embedSubBatch()` **only when `providerOptions.openaiCompatible.input_type` is actually present**, and read at body-rewrite time by the fetch shims:

- `zeroEntropyCompatFetch`: recovers the threaded value; the `'document'` default is preserved byte-for-byte for paths that didn't thread one (ingest).
- `voyageCompatFetch`: opt-in, mirroring the dims.ts Voyage branch — injects only when a caller threaded one; otherwise the field stays off the wire (pre-v0.35.0.0 behavior preserved).
- `openAICompatAsymmetricFetch` (new): the fallthrough default for every openai-compatible recipe without its own compat fetch (llama-server, litellm, ollama, …) — the canonical local/proxy paths for asymmetric models (a zembed-1 behind llama.cpp, vLLM, or an OpenAI-compatible proxy). Strictly opt-in: when nothing was threaded the request passes through untouched, so symmetric deployments see **zero wire change** (dims.ts threads input_type by model id, never for Azure/DashScope/Zhipu-style fixed models); when threaded, the field is injected (OpenAI-compatible servers ignore unknown body keys; asymmetric-aware servers/proxies read it). Recipes with their own compat fetch (azure) keep it via the `compat.fetch ??` precedence.

**`KNOBS_HASH_VERSION` bumped 10→11**: the semantic query cache keys rows by query-embedding similarity, and every pre-fix row was written with document-side query vectors. Without the bump, pre-fix rows could be served to post-fix lookups for up to `cache.ttl_seconds`. Same convention as the v=3 embedding-provider bump; one-time global cold-miss on upgrade, refills within TTL. The four test pins on the version were updated accordingly.

## Tests

`test/embed-input-type-wire.test.ts` — runs the **real SDK transport** with a mocked global `fetch` and asserts on the outbound body, the only layer where this regression is observable (one layer below the existing contract test):

- ZE hosted: query → `input_type: 'query'`; index path → `'document'`.
- llama-server + zembed-1: query → `'query'`; index → `'document'`; non-asymmetric model → **no** `input_type` (strict pass-through pinned).
- litellm + ollama serving zembed-1: query → `'query'` (the fallthrough shim covers them).
- Voyage: query → `'query'` (alongside the existing `output_dimension`/`encoding_format` translations); index path → field stays off the wire (opt-in pinned).

**4 of the original 7 assertions fail on current master**, proving the test pins the actual bug. One structural pin in `test/ai/zeroentropy-compat-fetch.test.ts` updated to the new line shape (same semantic: document default preserved). `KEY_FILES.md`'s gateway.ts entry updated to describe the new mechanism.

## Verification

- New file + `test/asymmetric-encoding-contract.test.ts`, `test/ai/*`, `test/search-mode.test.ts`, `test/cross-modal-phase1.test.ts`, `test/search-alias-resolved-boost.test.ts`, `test/search/knobs-hash-reranker.test.ts`, `test/query-cache-knobs-hash.serial.test.ts`, `test/embedding-dim-check.test.ts`, `test/embed-multimodal-batching.test.ts`, `test/openai-compat-multimodal.test.ts`, `test/doctor-ze-checks.test.ts`, `test/build-llms.test.ts`: **543 pass / 0 fail**.
- `bun run typecheck`: clean.
- Verified end-to-end on a real brain running zembed-1 locally behind llama-server: before the fix, semantically-phrased queries (no keyword overlap with the target page) ranked near-empty stub pages above the exact page; after, the query-side encoding reaches the serving layer and the exact page ranks top-1 consistently.


time-attack:
Thanks for this, @pabloglzg! 🙏

What this does: This PR fixes a bug where asymmetric embedding models (ZeroEntropy zembed-1, Voyage v3+, llama-server) had their query-side input_type stripped by the AI SDK before building the HTTP body, causing all queries to encode as documents and…

PR #2041, opened 2026-06-10
fix(cli): update post-install advisory to suggest 'scaffold' instead …

batman-tdk:
had installed gbrain manually in docker

running the command
`docker compose run --rm gbrain gbrain skillpack install --all`

gives the error 

```
Error: 'gbrain skillpack install' was removed in v0.33. Use 'gbrain skillpack scaffold <name>' instead.
If you're upgrading from an older release, run 'gbrain skillpack migrate-fence' once to strip the legacy managed block, then scaffold any new skills.
```


time-attack:
Thanks for reporting this from a real Docker install — the advisory pointing at a removed command was a genuine papercut. It has since been fixed on master: `post-install-advisory.ts` now renders `gbrain skillpack scaffold ...` through the shared renderer so the advisory and `gbrain advisor` stay in agreement. Closing as already fixed; the only other changes here are comment rewording, which we would rather not churn.

PR #2046, opened 2026-06-10
fix(links): collapse hyphen runs in normalizeBasename

DylPorter:
## Bug

Under global-basename resolution, a wikilink whose target title contains `" - "` (space-hyphen-space) never resolves, so the edge is silently dropped. `[[Note - Quarterly Review]]` pointing at the page slugged `notes/note-quarterly-review` produces no link.

## Cause

`normalizeBasename` (src/core/link-extraction.ts) does `.replace(/\s+/g, '-')`, which collapses whitespace runs but not hyphen runs. `" - "` has a literal hyphen between two spaces, so it normalizes to a triple hyphen:

```
"Note - Quarterly Review"            ->  "note---quarterly-review"
slug tail keyed by buildBasenameIndex ->  "note-quarterly-review"
```

They never match. The slug generator already collapses separators, so the lookup side is asymmetric. This hits every page using the common `Type - Name` title convention.

## Fix

Collapse `[\s-]+` to one hyphen and trim stray edge hyphens, matching how slugs are built. Single-space titles are unaffected. Also bumps `LINK_EXTRACTOR_VERSION_TS` per the in-file convention so existing brains re-extract links on update.

## Testing

Added cases to `test/link-extraction.test.ts` for the hyphen-run collapse plus a no-regression check. `bun test test/link-extraction.test.ts` passes, `tsc --noEmit` clean.

### DIFF
--- a/src/core/link-extraction.ts
+++ b/src/core/link-extraction.ts
@@ -27,7 +27,7 @@ import type { PageType } from './types.ts';
  * OR updated_at > links_extracted_at`. It is an ISO-8601 string (NOT a number) —
  * the column is TIMESTAMPTZ and the predicate binds it as `::timestamptz`.
  */
-export const LINK_EXTRACTOR_VERSION_TS = '2026-05-31T00:00:00Z';
+export const LINK_EXTRACTOR_VERSION_TS = '2026-06-10T00:00:00Z';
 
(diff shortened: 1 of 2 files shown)


DylPorter:
Closing — re-doing with synthetic test fixtures instead of real data. Will reopen clean.

PR #2047, opened 2026-06-10
fix(links): collapse hyphen runs in normalizeBasename

DylPorter:
## Bug

Under global-basename resolution, a wikilink whose target title contains `" - "` (space-hyphen-space) never resolves, so the edge is silently dropped. `[[Project - API Gateway]]` pointing at the page slugged `guides/project-api-gateway` produces no link.

## Cause

`normalizeBasename` (src/core/link-extraction.ts) does `.replace(/\s+/g, '-')`, which collapses whitespace runs but not hyphen runs. `" - "` has a literal hyphen between two spaces, so it normalizes to a triple hyphen:

```
"Project - API Gateway"               ->  "project---api-gateway"
slug tail keyed by buildBasenameIndex ->  "project-api-gateway"
```

They never match. The slug generator already collapses separators, so the lookup side is asymmetric. This hits every page using the common `Type - Name` title convention.

## Fix

Collapse `[\s-]+` to one hyphen and trim stray edge hyphens, matching how slugs are built. Single-space titles are unaffected. Also bumps `LINK_EXTRACTOR_VERSION_TS` per the in-file convention so existing brains re-extract links on update and pick up the recovered edges.

## Testing

Added cases to `test/link-extraction.test.ts` for the hyphen-run collapse plus a no-regression check.

- `bun test test/link-extraction.test.ts` passes
- `tsc --noEmit` clean

On a ~230-page brain that used the `Type - Name` convention heavily, a full re-extract after the fix recovered ~600 previously-dropped edges and roughly halved the orphan count.

## Aside (not in this PR)

Path-style links with a non-canonical (capitalized) dir prefix, like `[[Some Folder/Some Page]]`, also miss in this resolver because `normalizeBasename` strips the `/` and mashes the segments. The filesystem extract path resolves those via relative-path logic, but the DB/basename resolver does not. Happy to send a follow-up that falls back to the final path segment if that is of interest.

### DIFF
--- a/src/core/link-extraction.ts
+++ b/src/core/link-extraction.ts
@@ -27,7 +27,7 @@ import type { PageType } from './types.ts';
  * OR updated_at > links_extracted_at`. It is an ISO-8601 string (NOT a number) —
  * the column is TIMESTAMPTZ and the predicate binds it as `::timestamptz`.
  */
-export const LINK_EXTRACTOR_VERSION_TS = '2026-05-31T00:00:00Z';
+export const LINK_EXTRACTOR_VERSION_TS = '2026-06-10T00:00:00Z';
 
 // ─── Entity references ──────────────────────────────────────────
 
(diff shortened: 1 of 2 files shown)


time-attack:
Great bug report and a correctly-scoped fix — confirmed still present on master: `normalizeBasename` collapses whitespace runs but not hyphen runs, while `slugifySegment` in sync.ts already does `.replace(/-+/g,'-')` plus edge-trim, so the lookup side was asymmetric and dropped every `Type - Name` wikilink. Merging. One thing on rebase: master's `LINK_EXTRACTOR_VERSION_TS` has moved to `2026-07-10T00:00:00Z`, so please bump above that rather than back down to `2026-06-10`, or brains will not re-extract. The path-prefix case in your aside is a separate discussion; please don't stack it here.

PR #2040, opened 2026-06-10
feat(extract): --frontmatter-only — schema-pack-driven link extraction without legacy typed inference

JiraiyaETH:
## Summary

Adds `--frontmatter-only` for `gbrain extract links` so the strict Phase 2 graph backfill lane can run deterministic frontmatter extraction without also firing legacy markdown/bare-link typed inference.

The new flag is valid only with `--include-frontmatter`; otherwise the CLI exits with a clear usage error. When enabled, `runExtract -> extractLinksFromDB -> extractPageLinks` threads `frontmatterOnly: true`, skipping the markdown entity-ref and bare-slug passes while preserving `extractFrontmatterLinks` and active schema-pack `frontmatter_links` semantics.

I checked `src/core/operations.ts`; `extract links` is not exposed as a shared MCP operation, so this is CLI-only for now. That keeps scope tight and avoids inventing a second operation surface.

## Rationale

This implements the preferred shape from the Phase 2 strict extraction decision packet. The packet blocked full Option A because `--include-frontmatter` still ran the legacy typed `inferLinkType` path, where previous false `invested_in` / `founded` risk lived. This patch creates the native strict switch needed before approval gate #3.

It also matches the schema/ontology calibration direction: Option A remains the operationally safe first-write posture, while Option C is the durable schema-pack-first typed extraction direction. `--frontmatter-only` is the small bridge: it lets the Option A gate use pack-declared frontmatter semantics now, without broad legacy inference noise.

Lineage:
- Phase 2 decision packet: `phase2-strict-extraction-decision-packet-20260610T000000Z/phase2-decision-packet.md`
- Calibration plan: `90-schema-ontology-calibration-plan.md` Option A -> Option C direction
- TODO lineage: `TODOS.md` T1.5 / TODO-1 pack-aware runtime migration pattern

## Tests

- `bun test test/link-extraction.test.ts test/extract-db.test.ts` -> 142 pass / 0 fail
- `bun run typecheck` -> exit 0

## Phase 2 dry-run receipt

Command:

```bash
bun src/cli.ts extract links --source db --source-id default --include-frontmatter --frontmatter-only --dry-run --json
```

Result: 265 frontmatter edges, all `link_source=frontmatter`.

By type:
- `about`: 231
- `related_to`: 14
- `derived_from`: 11
- `sourced_from`: 5
- `discussed_in`: 2
- `works_at`: 2
- `attended`: 0

Receipts are in the Phase 2 specs directory:
[...]

PR #2048, opened 2026-06-10
fix(dream,minions): allow-list-derived paths + config-aware subagent auth — dream cycle on custom schema packs

JiraiyaETH:
# PR draft — Patch #6: subagent auth honors gbrain config

## Summary
- Make the Anthropic-direct subagent handler resolve API keys the same way the gateway does: `ANTHROPIC_API_KEY` first, then `gbrain config set anthropic_api_key ...`.
- Pass the resolved key into the SDK constructor for the default `sdk.messages` client path.
- Keep the existing clear no-key failure shape when both env and config are absent.

## Why
The attempt-2/attempt-3 dream-cycle receipts showed synthesize children failing under launchd/MCP-style environments where shell env was absent even though `anthropic_api_key` was configured in gbrain. The gateway path already honored config; the legacy Anthropic-direct subagent loop did not.

## Tests
- `bun test test/subagent-handler.test.ts`

## Notes
No wrapper changes in this PR. Production activation should be done by pointing the wrapper at this carried branch after review.
# PR draft — Patch #5: synthesize prompt examples come from allow-list

## Summary
- Derive reflection/original slug example prefixes from the threaded `allowed_slug_prefixes` list sent to synthesize children.
- Strip glob suffixes before rendering examples.
- Preserve stock `wiki/personal/reflections` and `wiki/originals/ideas` examples only when the allow-list is empty.

## Why
The attempt-2/attempt-3 receipts showed the dream orchestrator correctly threading an allow-list, but the child prompt still nudged the model toward stock wiki paths. That creates avoidable conflict for carried/local installs with different filing prefixes.

## Tests
- `bun test test/cycle-synthesize.test.ts`

## Notes
This changes prompt examples only; server-side `brain_put_page` enforcement still comes from the same allow-list.
# PR draft — Patch #4: dream summary slug follows allow-list

## Summary
- Derive the dream-cycle summary prefix from the configured dream synthesize allow-list when a `dream-cycle` glob is present.
- Add `dream.synthesize.summary_slug_prefix` as the fallback config key for custom installs without a dream-cycle allow-list entry.
- Preserve the stock `dream-cycle-summaries/YYYY-MM-DD` default for back-compat.

## Why
The attempt-2/attempt-3 receipts showed child outputs could be safely relocated via allow-list, but the orchestrator summary stayed hardcoded to `dream-cycle-summaries/${date}`. This closes that remaining local-carry path mismatch without breaking stock installs.

## Tests
- `bun test test/cycle-synthesize.test.ts`

## Notes
The orchestrator still validates the final summary slug against the existing slug regex before writing.

> Stacked on #2040 → #2030 → #2029. Three commits, one theme: dream-cycle paths and auth honor operator configuration (active-pack allow-list + config-stored key) instead of hardcoded stock-brain assumptions. Found running dream on a custom schema pack — full forensic receipts available.


time-attack:
Superseded with credit by #3110 (just merged) — your change landed rebased onto current master with authorship preserved via Co-authored-by. Thank you!

PR #2052, opened 2026-06-10
fix(sync): honor .gitignore in code walk; prune vendor/dist/build

aphaiboon:
## Problem

`gbrain sync --strategy code` indexes git-ignored trees. On a Laravel/PHP repo this means it descends into `vendor/` (~50k Composer files), `storage/`, and `public/build/`, trying to import **52,164 files** where only ~1,000 are real source. The result: the full sync never finishes (watchdog SIGTERM at ~3%), and whatever it does import floods the index with library internals (`vendor/symfony/...`, `vendor/psy/psysh/...`), so semantic code search returns dependency files instead of project code.

Root cause: `collectSyncableFiles` (the full-sync / dry-run enumerator in `src/commands/import.ts`) reimplemented its own descent skip list inline — `if (entry === 'node_modules' || entry === 'ops') continue;` — bypassing the canonical `pruneDir` gate in `src/core/sync.ts` and ignoring `.gitignore` entirely. The two skip lists drifted, and neither covered `vendor`.

This is the concrete bug behind the open `--respect-gitignore` (#1159) / `.gbrainignore` (#1483) requests and overlaps the maintainer's #1942 vendor/dist/build prune.

## Fix

1. **Git-aware enumeration.** When `dir` is a git work tree, `collectSyncableFiles` now lists files via `git ls-files --cached --others --exclude-standard -z` (tracked + untracked-not-ignored), so the walk honors `.gitignore`. Non-git dirs fall through to the existing recursive FS walk unchanged. EroLab (a Laravel repo): **52,164 → 1,028 files**.
2. **Dedupe the FS fallback to the canonical gate.** The fallback walk now prunes through `pruneDir()` instead of the drifted inline `node_modules || ops` list, so the two can't diverge again.
3. **`PRUNE_DIR_NAMES` gains `vendor`, `dist`, `build`** — dependency/build-output trees that are git-ignored on virtually every repo. This covers the FS-fallback + incremental (`classifySync`) paths.

## Testing

- `git ls-files` path verified end-to-end via `--dry-run` on a real Laravel repo: 52,164 → 1,028.
- Walker regression suites green (90 pass, 0 fail): `test/sync-walker-symlink.test.ts`, `test/brain-writer-walk-prune.test.ts`, `test/sync.test.ts`, `test/sync-walker-submodule.test.ts`.
- `tsc --noEmit` clean.

Symlink policy and submodule-gitlink skipping are preserved (each `ls-files` entry is `lstat`-checked).

### DIFF
--- a/src/commands/import.ts
+++ b/src/commands/import.ts
@@ -11,6 +11,7 @@ import {
   isCodeFilePath,
   isMarkdownFilePath,
   isImageFilePath as isImageFilePathFromSync,
+  pruneDir,
   type SyncStrategy,
 } from '../core/sync.ts';
 import { sortNewestFirst } from '../core/sort-newest-first.ts';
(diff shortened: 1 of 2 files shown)


spotshare-nick:
Nice fix — the root-cause framing is right. The real bug here isn't just "doesn't honor `.gitignore`," it's that `collectSyncableFiles` carried its own inline skip list (`node_modules || ops`) that drifted from the canonical `pruneDir()` gate and covered neither `vendor` nor anything else. Routing the FS fallback back through `pruneDir()` so the two can't diverge again is the part #1159 doesn't do, and adding `vendor`/`dist`/`build` to `PRUNE_DIR_NAMES` helps the incremental (`classifySync`) path too, not just the full walk. The Laravel 52,164 → 1,028 repro is a clean demonstration of the #1073 stall.

Worth noting for the maintainer's triage: like #1159, this adds **no npm dependency** — it shells out via `execFileSync('git', …)`, so the "requires the `ignore` package" concern that kept #1159 out of the v0.37.7.0 wave doesn't apply here either.

**One behavioral risk to flag:** this is **default-on** — whenever `dir` is a git work tree the walk switches to `git ls-files`, with no flag or config gate. That silently changes behavior for **dotfile/secret brains that deliberately `.gitignore` content but still want it brain-searchable** — the exact "why not just default-on?" case raised in #1073, which #1159 preserved by putting the behavior behind a default-off `--respect-gitignore` flag. Repos like this one clearly *want* the prune, but those brains would lose indexed content without opting in.

I don't have a strong opinion on how to resolve that — whether to blend the two (this PR's `pruneDir` dedupe + `PRUNE_DIR_NAMES` additions, gated behind #1159's opt-in flag), accept this one as-is and treat the secret-brain case as a follow-up, or pick one PR over the other — is entirely the maintainer's call, and it overlaps the in-flight #1942 vendor/dist/build prune as well. Just wanted the always-on tradeoff to be visible alongside the (real) improvements so it's a deliberate choice.

---

*This comment was drafted with the help of [Claude Code](https://claude.com/claude-code).*

PR #2054, opened 2026-06-10
fix(doctor): avoid misleading PGLite and gateway-loop warnings

mollusk:
## Summary
- Treat pgvector and JSONB integrity checks as OK/not-applicable on PGLite instead of warning on Postgres-only probes.
- Honor agent.use_gateway_loop=true in subagent capability doctor checks before warning about missing ANTHROPIC_API_KEY for non-Anthropic chat models.
- Add agent.use_gateway_loop to known config keys.
- Add regression coverage for PGLite doctor checks and gateway-loop subagent configuration.

## Test Plan
- bun test test/doctor-pglite-and-gateway-loop.test.ts
- bun test test/doctor-subagent-health.test.ts test/doctor.test.ts
- bun run typecheck

## Local verification
- Andrew's local PGLite brain now reports pgvector/jsonb_integrity/subagent_capability as OK in doctor output; remaining warning is only takes_count.

### DIFF
--- a/src/commands/doctor.ts
+++ b/src/commands/doctor.ts
@@ -2672,8 +2672,18 @@ async function checkSubagentCapability(engine: BrainEngine): Promise<Check> {
       const { loadConfig } = await import('../core/config.ts');
       const cfg = loadConfig();
       const chatModel = cfg?.chat_model;
+      const useGatewayLoopRaw = await engine.getConfig('agent.use_gateway_loop').catch(() => null);
+      const useGatewayLoop = useGatewayLoopRaw === 'true' || useGatewayLoopRaw === '1';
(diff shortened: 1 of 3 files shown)


time-attack:
Thanks for chasing down the false warnings — they were real, and they have since been fixed on master along the same lines you proposed. `pgvectorCheck` and `jsonbIntegrityCheck` now run through `engine.executeRaw()` so PGLite brains are probed on their own connection, the subagent capability check reads `agent.use_gateway_loop` through `isConfigTruthy` before warning about `ANTHROPIC_API_KEY`, and that key is in the known-config-keys list. Closing as already fixed — please do re-run `gbrain doctor` on latest and open a fresh issue if anything still reads wrong on PGLite.

PR #2061, opened 2026-06-10
[codex] fix Supabase pgvector doctor checks

crumeb:
## Summary

- avoid explicit pgvector casts in Supabase Postgres paths where restricted roles cannot use the private extensions schema
- teach doctor to parse schema-qualified pgvector types and respect agent.use_gateway_loop
- treat an external autopilot lock as healthy when its owner PID is still running
- add targeted regression tests for the doctor and Postgres-engine paths

## Validation

- bun test test/doctor-v0_37_7_checks.test.ts test/doctor.test.ts test/doctor-subagent-capability.test.ts test/postgres-engine.test.ts
- bun run typecheck


crumeb:
Closing this PR. It was opened against the public upstream by mistake; this work should only go through our private repository workflow.

PR #2062, opened 2026-06-10
fix(gateway): reconcile dangling tool-call replay

### LINKED ISSUE #2039: gateway toolLoop: no resume reconciliation — crash mid-tool-round leaves a dangling tool-call tail that dead-letters every retry (ModelMessage[] schema / missing tool results)
## Summary

Two gaps in the gateway-native tool loop (`agent.use_gateway_loop=true`) make subagent retries fail **deterministically** after a mid-tool-round crash. Closely related to #1886 (which covers persisting the tool-results message in-loop); this is the **resume** half plus a history-poisoning class that persistence alone doesn't fix. We carry a local patch (described below, with unit tests) and are happy to PR.

## Environment

- gbrain v0.42.37.0, Postgres engine (Supabase session pooler)
- `agent.use_gateway_loop=true`, model `openrouter:deepseek/deepseek-v4-pro`
- Carrying the #1886-style persistence fix locally (`onToolResultsTurn` → `persistMessage`) — these failures occur **with** that fix in place

## Production evidence

Dream `synthesize` fan-out, 14 subagent children: **8 dead after 3 attempts each**. `error_text` on the dead jobs:

- 7×: `[chat(openrouter:deepseek/deepseek-v4-pro)] Invalid prompt: The messages do not match the ModelMessage[] schema.`
- 1×: `[chat(openrouter:deepseek/deepseek-v4-pro)] Tool results are missing for tool calls call_<id>, call_<id>, call_<id>.`

Telling detail: **5 of the 6 jobs that eventually completed show the same errors on early attempts** — when the crash point left a *consistent* persisted history, the retry succeeded; when it left a dangling tail, every retry failed identically until `max_attempts`. The failure is a deterministic property of the persisted history, so raising `max_attempts` cannot help.

## Root cause 1 — no resume reconciliation in `toolLoop`

`toolLoop` seeds the conversation straight from `replayState.priorMessages` and immediately calls `chat()`:

```ts
const messages: ChatMessage[] = opts.replayState
  ? [...opts.replayState.priorMessages]
  : [...opts.initialMessages];
```

Per D11 write-ordering, the assistant turn is persisted **before** tool dispatch, but the tool-results user message is only persisted **after the entire tool round completes**. A tool round can run for minutes (`put_page`, `search`, …), so a stall-kill / timeout / worker crash in that window leaves the persisted history ending in an assistant message with dangling `tool-call` blocks. On resume, the very first `chat()` sends that prompt; AI SDK v6 rejects it (`MissingToolResultError` / ModelMessage zod validation); the attempt fails before reaching any code that could repair the tail — and every subsequent retry replays the identical poisoned history.

The legacy Anthropic-direct path in `src/core/minions/handlers/subagent.ts` has had exactly this protection for a long time ("Replay reconciliation": settle dangling `tool_use` blocks from `subagent_tool_executions` outcomes, synthesize the tool-results user turn, then resume). The gateway path never got the equivalent, and all infrastructure it needs (`replayState.priorTools`, `onToolCallStart`'s ON-CONFLICT re-claim, idempotency flags) is already there.

A second, smaller miss in the same family: when the prior history ends in a **terminal** assistant turn with *no* tool calls (crash after persisting the answer but before job bookkeeping), the gateway resume calls `chat()` with a trailing assistant message — assistant-prefill, which several providers reject. The legacy path returns the persisted text instead.

## Root cause 2 — persisted-history poisoning passes through `toModelMessages`

`toModelMessages` (added for v6 in #1759/#1809) fixed the tool-result/`role:'tool'` conversion but still lets two shapes through:

1. **`tool-call` blocks with `input: undefined`.** `chat()` normalizes provider parts with `input: part.input ?? part.args` — DeepSeek emits zero-arg tool calls where both are absent, so `undefined` is persisted (JSONB drops the key) and replayed forever.
2. **Unknown block types fall through `return b` as-is** into the v6 messages array.

Either one, once persisted, fails zod validation on every later turn and every retry of that job → `Invalid prompt: The messages do not match the ModelMessage[] schema`.

## Fix we run locally

1. **Resume reconciliation in `toolLoop`** (mirrors the legacy path): if `replayState.priorMessages` ends in an assistant message with `tool-call` blocks, settle each call before the first `chat()` — prior `complete` → replay output; prior `failed` → error result; `pending` + idempotent → re-execute; `pending` + non-idempotent → unrecoverable (same contract as the in-loop path); unregistered tool → error result. Persist the synthesized tool-results turn via `onToolResultsTurn` so the *next* resume is also consistent. If the tail is a terminal assistant turn (no tool calls), return its text instead of assistant-prefilling.
2. **Defensive `toModelMessages`**: coerce missing `tool-call` input to `{}`, map `reasoning` text parts through, drop unknown block types, and drop messages whose content sanitizes to nothing.
3. **`?? {}` at both `chat()` normalization sites** so undefined input is never persisted in the first place.
4. **Diagnosability**: on prompt-validation failures, the `llm_call_failed` heartbeat now includes a `message_shapes` summary (roles + block types only, never content), so a poisoned history can be diagnosed from logs instead of needing a live repro.

Unit tests added (all green, plus the existing 15 pinned cases unchanged):

- `toolLoop` resume: settles dangling calls from prior outcomes before first `chat()` (no re-execution for `complete`)
- `toolLoop` resume: re-executes idempotent `pending`, maps `failed` to an error tool-result
- `toolLoop` resume: terminal assistant tail returns persisted text without calling `chat()`
- `toModelMessages`: undefined input → `{}`; unknown blocks dropped; `reasoning` preserved; empty messages dropped

Happy to open a PR with the full diff if maintainers want it.

maxpetrusenkoagent:
## Summary
- reconcile replay history when a prior assistant tool-call tail is missing its tool result
- preserve terminal assistant replay tails without pre-filling the next provider call
- surface unrecoverable pending non-idempotent tool replays instead of dead-lettering retry attempts
- add regression coverage for gateway tool-loop replay and ModelMessage tool-result conversion

Fixes #2039.

## Verification
- `PATH=/Users/maxsmacmini/.local/bin:/opt/homebrew/bin:/Users/maxsmacmini/.bun/bin:$PATH bun test test/ai/gateway-tool-loop.test.ts test/gateway-model-messages.test.ts`
- Result: 19 pass, 0 fail


maxpetrusenkoagent:
Verification from Mac mini catch-up lane:

- Focused tests: `PATH=/Users/maxsmacmini/.local/bin:/opt/homebrew/bin:/Users/maxsmacmini/.bun/bin:$PATH bun test test/ai/gateway-tool-loop.test.ts test/gateway-model-messages.test.ts`
- Result: 19 pass, 0 fail
- Duplicate check: no existing PR found for issue #2039 before opening this draft

GitHub currently reports no status check suite for this draft PR.

time-attack:

PR #2063, opened 2026-06-10
test(ai): pin gateway tool schema conversion

### LINKED ISSUE #1487: gateway.chat(): incorrect tool schema format breaks non-Anthropic providers (agent.use_gateway_loop)

maxpetrusenkoagent:
## Summary
- Extract gateway tool conversion into `toAISDKTools()` so `chat()` and tests exercise the same schema builder
- Keep AI SDK v6 tool schemas wrapped with `jsonSchema()` instead of the legacy bare `{ jsonSchema }` object that triggers `schema is not a function`
- Strengthen the real `generateText()` regression test for the gateway tool loop path

Closes #1487

## Test Plan
- `bun test test/ai/gateway-tools-schema.test.ts`
- `bun test test/ai/gateway-tools-schema.test.ts test/ai/gateway-chat.test.ts test/ai/gateway-tool-loop.test.ts`
- `bun run verify`

### DIFF
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -2532,6 +2532,22 @@ async function classifyGatewayGuardrail(input: {
   }
 }
 
+export function toAISDKTools(tools: ChatToolDef[] | undefined): Record<string, any> | undefined {
+  if (!tools || tools.length === 0) return undefined;
+  return tools.reduce((acc, t) => {
+    acc[t.name] = {
+      description: t.description,
+      // AI SDK v6 requires a Schema (carrying the schema symbol), not a plain
+      // `{jsonSchema}` object — the bare object makes asSchema() treat it as a
+      // thunk and call schema(), throwing "schema is not a function". Wrap the
+      // raw JSON Schema with the SDK's jsonSchema() helper so tool calls work
+      // through the real toolLoop (skillopt rollouts + subagent jobs).
+      inputSchema: jsonSchema(t.inputSchema as any),
+    };
+    return acc;
+  }, {} as Record<string, any>);
+}
+
 export async function chat(opts: ChatOpts): Promise<ChatResult> {
   const tracker = __budgetStore.getStore() ?? null;
   const modelStrEarly = opts.model ?? getChatModel();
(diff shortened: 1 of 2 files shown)


maxpetrusenkoagent:
Verification from Mac mini catch-up lane:

- Focused tests: `PATH=/Users/maxsmacmini/.local/bin:/opt/homebrew/bin:/Users/maxsmacmini/.bun/bin:$PATH bun test test/ai/gateway-tools-schema.test.ts test/ai/gateway-chat.test.ts test/ai/gateway-tool-loop.test.ts`
- Result: 32 pass, 0 fail

GitHub currently reports no status check suite for this draft PR.

PR #2067, opened 2026-06-10
fix(sync): stamp last_sync_at on no-op up_to_date sync

mgunnin:
When `performSync` finds git HEAD unchanged, it returns early with status `up_to_date` but never calls `writeSyncAnchor`, so `sources.last_sync_at` is never refreshed.

doctor's `sync_freshness` check reads `sources.last_sync_at` directly (`doctor.ts`), so a quiet, fully-caught-up source shows as **stale indefinitely** — a successful no-op sync is *proof* of freshness but left the timestamp untouched. In practice this surfaces as a `sync_freshness` warning that never clears no matter how many times you run `gbrain sync`.

### Fix
Stamp the freshness anchor before the `up_to_date` return (`src/commands/sync.ts`) by reusing the existing `writeSyncAnchor(engine, sourceId, 'last_commit', headCommit, commitTimeMs(repoPath, headCommit))` path. Re-writing the same `last_commit` value is value-neutral; the point is the `now()` timestamp + `newest_content_at` refresh in the same atomic UPDATE that branch already performs.

### Test
Adds a regression test in `test/sync.test.ts` (`no-op up_to_date sync stamps last_sync_at`) asserting `last_sync_at` strictly advances across a no-op `up_to_date` sync. Fails against current `master`, passes with the fix. Full `test/sync.test.ts`: **67 pass, 0 fail**. `tsc --noEmit`: clean.

### Scope
Single short-circuit branch; no behavior change to `synced` / `first_sync` / `dry_run` / `partial` paths (those already stamp via their own `writeSyncAnchor` calls).


time-attack:
Closing as superseded — the exact fix (stamping the freshness anchor on the `up_to_date` no-op short-circuit in `performSyncInner`) already landed on master via #2335 (3a2033e8, v0.42.52.0). `sources.last_sync_at` now advances on a quiet, fully-caught-up sync. Thanks for the clean repro + regression test — feel free to check master and reopen if the staleness still reproduces.

Issue #1715, opened 2026-06-01
bug: doctor-categories missing emitted check names, causing unknown-check warnings


iannwu:
## Summary

`gbrain doctor --json` emits stderr warnings for several doctor checks whose names are not registered in `src/core/doctor-categories.ts`:

```text
[doctor-categories] unknown check name 'embed_staleness' — defaulting to 'meta'. Add it to src/core/doctor-categories.ts.
[doctor-categories] unknown check name 'entity_link_coverage' — defaulting to 'meta'. Add it to src/core/doctor-categories.ts.
[doctor-categories] unknown check name 'timeline_coverage' — defaulting to 'meta'. Add it to src/core/doctor-categories.ts.
[doctor-categories] unknown check name 'takes_count' — defaulting to 'meta'. Add it to src/core/doctor-categories.ts.
[doctor-categories] unknown check name 'pack_upgrade_available' — defaulting to 'meta'. Add it to src/core/doctor-categories.ts.
[doctor-categories] unknown check name 'type_proliferation' — defaulting to 'meta'. Add it to src/core/doctor-categories.ts.
[doctor-categories] unknown check name 'dangling_aliases' — defaulting to 'meta'. Add it to src/core/doctor-categories.ts.
```

This looks like a doctor check/category registry drift: `doctor` emits these check names, but `doctor-categories.ts` does not include them in `BRAIN_CHECK_NAMES`, `OPS_CHECK_NAMES`, or `META_CHECK_NAMES`.

## Environment

- Installed CLI: `gbrain 0.41.38.0`
- Checked latest upstream source: `origin/master` at `eefe8b57`, `VERSION` = `0.42.1.0`
- Engine: Postgres + pgvector
- Command:

```bash
gbrain doctor --json > /tmp/gbrain-doctor.json 2> /tmp/gbrain-doctor.err
rg 'doctor-categories' /tmp/gbrain-doctor.err
```

## Observed behavior

Doctor completes, but stderr logs the unknown-check warnings above. The JSON output then defaults those checks to category `meta`.

## Expected behavior

Every doctor check emitted by `doctor` should be registered in `src/core/doctor-categories.ts`, or the categorizer should intentionally recognize that group without warning. A normal doctor run should not produce unknown-check warnings for first-party checks.

## Impact

- No data loss.
- No install blocker.
- Noisy stderr for operators/agents parsing doctor output.
- Category reporting can be misleading because these checks silently fall through to `meta`.

## Suggested fix

[...]

PR #1839, opened 2026-06-04
fix: register missing doctor check names in doctor-categories to stop unknown-check warnings

mvanhorn:
## Summary
`gbrain doctor --json` no longer prints `[doctor-categories] unknown check name` warnings for emitted checks. The missing names (`embed_staleness`, `entity_link_coverage`, `timeline_coverage`, `takes_count`, `dangling_aliases`, `pack_upgrade_available`, `type_proliferation`) are now registered to their correct category sets in `src/core/doctor-categories.ts` instead of falling through to `meta`.

## Why this matters
Issue #1715 reports that several doctor checks emit a stderr warning on every run because their names are absent from the registry and default to `meta`. The warning text itself prescribes the fix ("Add it to src/core/doctor-categories.ts"). The drift guard in `test/doctor-categories.test.ts` only scanned `doctor.ts`, so checks emitted from `src/core/onboard/checks.ts` slipped past it.

## Testing
- `test/doctor-categories.test.ts` now scans both `doctor.ts` and `onboard/checks.ts` emitter sources, and asserts the newly registered names resolve to `brain`/`meta` rather than `meta`-with-warning.
- A genuinely unknown name still falls through to `meta` with one warning, and already-registered names keep their categories.

Fixes #1715

PR #2069, opened 2026-06-10
test(doctor): include onboard checks in category drift guard

mgunnin:
## Summary
- Include `src/core/onboard/checks.ts` in the doctor category drift guard source scan.
- Categorize onboard-emitted check names that doctor surfaces through the onboard check path.
- Update drift guard failure text to reflect that it now scans both doctor and onboard check sources.

## Why
Doctor category scoring can drift when checks are emitted through imported onboard check sources rather than appearing directly in `doctor.ts`. The drift guard should scan both places so category scoring remains stable as check sources are split out.

## Verification
- `bun test test/doctor-categories.test.ts`
- `bun run typecheck`
- `~/.agents/skills/autoreview/scripts/autoreview --mode auto`


time-attack:
Thanks @mgunnin! 🙏

What this does: This PR fixes issue #1715 by registering 7 missing doctor check names (embed_staleness, entity_link_coverage, timeline_coverage, takes_count, dangling_aliases, pack_upgrade_available, type_proliferation) in doctor-categories.ts and…

This overlaps with #1839, which we're carrying forward, so I'm closing this in its favor to keep the review queue focused. If yours covers a case #1839 misses, reopen with that detail and we'll reconcile them.

PR #2073, opened 2026-06-10
v0.42.39.0 fix(cli,code-graph): search silently killed at 10s on slow poolers + scoped call-graph queries returned 0 rows + dim-migration recipe order

jbarol:
## Summary

Two silent-failure bugs found and traced live on a production brain behind a Supabase transaction pooler (6-10s per fresh connection), plus a recipe-order defect found three times in the same codebase.

**CLI: search silently killed at 10s**
- `fix(cli)`: the op-dispatch disconnect hard-deadline timer was armed BEFORE the try block, so any op whose handler ran past 10s wall-clock was force-exited mid-flight with code 0 and zero stdout. On a slow pooler every `gbrain search` hit this — an empty "success" indistinguishable from no results. The timer now arms at teardown entry (same placement as the fall-through owner-disconnect site). Verified: a search that died at 10s now returns correct results at ~21s.
- `fix(cli)`: read-scope handlers and context build (DB I/O for every op) get a real wallclock bound — 180s default, `--timeout=Ns`, exit 124 — and the timeout path hard-exits after teardown so the abandoned handler (withTimeout races, never cancels) can't hold sockets that zombify the process. This also completes the TODOS.md item "(v0.42.20.0 follow-up) Decouple the op-dispatch force-exit timer".

**Code graph: scoped `code-callers`/`code-callees` returned 0 rows**
- `fix(import)`: `importCodeFile` never stamped `source_id` on extracted edges → every edge NULL → the scoped filter `AND source_id = ...` never matched. Edges are stamped with a once-normalized source (`sourceId || 'default'`, matching the pages-table default; also closes an `''`-source FK-violation door that silently dropped a file's whole call graph, and scopes the chunk lookup to prevent cross-source slug fan-out).
- `fix(engine)`: both engines default `source_id` to `'default'` at the insert layer as a last line of defense (parity).
- `feat(migrate)`: v116 backfills every pre-existing NULL edge from its own chunk's page, and indexes `from_symbol_qualified` on both edge tables (getCalleesOf had no index → seq scan per BFS node). Backfill SQL verified live on a 2,122-edge production brain. Mirrored in schema.sql; schema-embedded regenerated.

**Embedding dim-migration recipe order (×3)**
- `docs(migrations)` + `fix(core)`: the doc recipe, the runtime `embeddingMismatchMessage` that prints it, and `buildFactsAlterRecipe` all ordered `ALTER COLUMN TYPE vector(N)` before wiping old embeddings — pgvector refuses cross-dimension casts ("expected N dimensions, not M") and aborts, on exactly the brains that need the recipe. All three now NULL first; same-dim type swaps (halfvec ↔ vector) keep the lossless cast and preserve data. Orders pinned by tests so they can't drift apart again.

## Test Coverage

```
CODE PATHS (docs commits: no code paths)

src/cli.ts — force-exit timer relocation + wallclock bounds
├─ [★] no arming before op-body try (slow op survives >10s) — cli-force-exit-teardown-arming.test.ts (verified fails pre-fix)
├─ [★] arming at teardown entry, before drain+disconnect
├─ [★] daemon guard / unref / clearTimeout intact
├─ [GAP] timer fires on hung teardown — intentionally uncovered: needs >=10s wall-clock + wedged disconnect in a spawned CLI, flaky by construction; firing logic unchanged, only relocated
└─ [★] clean teardown — e2e/pglite-cli-exit.serial.test.ts

src/core/import-file.ts — edge source stamping
├─ [★] scoped import stamps edges; scoped getCallersOf/getCalleesOf match — import-code-edges-source-id.test.ts (verified fails pre-fix)
├─ [★] UNSCOPED import stamps 'default' (not NULL) — same file
├─ [★] cross-source isolation holds
└─ [★] engine-layer default — engine-parity suite green

src/core/embedding-dim-check.ts — recipe order
├─ [★] content_chunks recipe: UPDATE precedes ALTER — embedding-dim-check.test.ts (ordering pinned)
├─ [★] facts recipe dims-change: NULL-first — embedding-dim-check-facts.test.ts
└─ [★] facts same-dim type swap: PRESERVES data (no wipe)

COVERAGE: 11/12 paths (92%)
```
Tests: 1,137 → 1,139 files (+2 new); affected suites 297 pass / 0 fail; `tsc --noEmit` clean.

## Pre-Landing Review

10 findings (2 critical) from 3 specialists; 7 fixed in this branch, 3 skipped with rationale:
- [FIXED][CRITICAL] runtime dim recipe printed the same broken SQL order the doc fix corrected (and linked to the corrected doc)
- [FIXED][CRITICAL] `?? null` left the unscoped-import path stranding edges (pages default to 'default')
- [FIXED] no upgrade path for legacy NULL edges → migration v116 backfill
- [FIXED] `from_symbol_qualified` unindexed on both edge tables → v116 indexes
- [FIXED] shared-op handlers had no wallclock bound after the timer relocation
- [FIXED] stale "outer hard-deadline timer" comment
- [FIXED] doc rationale list contradicted the corrected recipe; fragile numeric cross-ref
- [SKIPPED] DRY: extract a shared teardown helper for the two now-identical finally sites — touches pre-existing code; left to maintainer taste
- [SKIPPED] declaration scope-tightening — the structural regression test deliberately anchors on the current shape
- [SKIPPED] `findChunkForOffset` O(edges×fileSize) CPU in import — pre-existing; suggest precomputing line-start offsets + binary search as a follow-up

## Adversarial Review

Claude adversarial subagent (Codex unavailable on this machine): 12 findings, 6 fixed here — including two HIGHs: the timeout path cleared the force-exit backstop and could zombify the process through the new code itself (fixed: hard-exit after teardown), and `buildFactsAlterRecipe` shipped the same NULL-before-ALTER defect class (fixed, with same-dim preservation). Remaining notes for the maintainer: a doctor probe for `source_id IS NULL` edge rows would catch old binaries writing to shared brains; v116's backfill is a full-table rewrite in one transaction (fine on modest edge tables; consider batching for very large brains); `--timeout` is silently ignored for write/admin ops (documented behavior choice); no behavioral test of the 124 timeout path (needs deterministic wall-clock control).

## Scope Drift

Scope Check: CLEAN — intent (fix the three traced defects) matches the diff; no unrelated changes.

## Plan Completion

No plan file detected.

## TODOS

- Completed: "(v0.42.20.0 follow-up) Decouple the op-dispatch force-exit timer" — v0.42.39.0 (2026-06-10)

## Test plan

- [x] `tsc --noEmit` clean
- [x] Affected suites: 297 pass / 0 fail (cli, code-edges, import-file×2, edge-extractor, migrate, schema-bootstrap-coverage, embedding-dim-check×2, engine-parity, + 2 new regression files)
- [x] New regression tests verified to FAIL on pre-fix code (temporary revert)
- [x] `bun test ./test/build-llms.test.ts` 12 pass (bundle fresh)
- [x] v116 backfill SQL verified live on a production Supabase brain (2,122 edges, scoped code-callers went 0 → 21)
- Note: the full unit suite on the test machine shows ~590 env/load-dependent failures (missing API keys, 1,128-file contention timeouts) — verified identical in isolation on master vs this branch (51/51 both sides on the top-failing files); not introduced by this diff.

Fork PR note: base-repo secrets don't reach fork PRs, so eval CI jobs will fail with empty-env auth per the repo's documented fork-PR behavior; check-freshness/lint/unit lanes should be green.


time-attack:
Thanks for this, @jbarol! 🙏

What this does: This PR (dated 2026-06-10) fixes three production bugs: a CLI force-exit timer killing search requests after 10 seconds on slow database connections, code-graph queries returning zero rows on source-scoped brains, and an embedding…

PR #2074, opened 2026-06-11
fix(postgres-engine): atomic JSONB merge in updateSourceConfig — eliminate lost-update race

pai-scaffolde:
## Problem

`updateSourceConfig` in `postgres-engine.ts` used a read-then-write pattern:
read the current `config` row, normalize it in JavaScript, then write the merged
result back with `SET config = <normalized> || <patch>`.

Under concurrent callers (two background autopilot or cycle paths patching
different keys simultaneously), both callers can read the same stale row. The
later `SET config = ...` then clobbers the earlier patch, silently dropping
whatever keys the first caller wrote. Running `gbrain dream` + `gbrain autopilot`
in parallel reproduced 21/25 lost-update events on real Postgres.

Note that `pglite-engine.updateSourceConfig` already used an atomic `||` merge; this
PR closes the parity gap in the Postgres engine.

## Fix

Fold the normalization and merge into a single atomic `UPDATE … SET config = CASE … END || patch` statement. Because the `SET` expression evaluates against the row-locked latest version of `config`, there is no snapshot window between read and write. Concurrent callers converge correctly.

The `CASE` normalizes historical bad JSONB shapes inline:

- `object` — used as-is
- `string` — double-encoded config; inner text is parsed using the SQL `IS JSON` predicate (Postgres 16+) to distinguish valid JSON (parsed and merged) from unparseable text (reset to `{}`, avoiding an `invalid input syntax for type json` error that would abort the whole `UPDATE`)
- `array` — flattened via `jsonb_object_agg`
- anything else — reset to `{}`

## Tests

Added two test cases to `test/list-all-sources.test.ts`:

1. JSONB string holding non-JSON text normalizes to `{}` without throwing
2. JSONB string holding double-encoded valid JSON is parsed then merged with the patch

```
12 pass, 0 fail
```

These run against PGLite (in-memory Postgres 17), which supports the `IS JSON` predicate at parity with Postgres 16+.

### DIFF
--- a/src/core/postgres-engine.ts
+++ b/src/core/postgres-engine.ts
@@ -1276,11 +1276,28 @@ export class PostgresEngine implements BrainEngine {
   }
 
   async updateSourceConfig(sourceId: string, patch: Record<string, unknown>): Promise<boolean> {
-    // v0.38: atomic JSONB merge. `||` is the Postgres concat operator —
-    // for jsonb, right-side keys overwrite left-side; nested object keys
-    // are NOT deep-merged (use jsonb_set for nested paths). The patch
-    // shape this autopilot wave uses is flat (`last_full_cycle_at`,
-    // `archive_*`, etc.) so concat is sufficient. Idempotent on re-run.
+    // Atomic single-statement merge. The previous read-then-write form dropped
+    // concurrent updates: two callers patching different keys could both read
+    // the same old config and the later `SET config = ...` clobbered the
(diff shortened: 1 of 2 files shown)


time-attack:
Thanks for this, @pai-scaffolde! 🙏

What this does: This PR fixes a lost-update race condition in the `updateSourceConfig()` method by moving the JSONB merge operation inside the SQL UPDATE statement, and adding a guard to handle historically malformed config shapes (string-encoded JSON).…

PR #2075, opened 2026-06-11
fix(doctor): five correctness fixes — stale locks, content sanity, graph coverage, exit code, gateway guard

pai-scaffolde:
Five independent correctness fixes for `gbrain doctor`, all found while running it
in production. Each is a small, isolated change; grouped here because they all touch
`src/commands/doctor.ts`.

---

### 1. Stale lock break hints cover `gbrain-cycle:` keys

The stale-lock report only recognized `gbrain-sync:` prefixes; everything else fell
back to `gbrain sync --break-lock`, which is wrong for cycle/dream locks. A
`gbrain-cycle:<source>` lock now suggests `gbrain dream --break-lock --source <name>`,
bare `gbrain-cycle` suggests `gbrain dream --break-lock`, and truly unknown lock shapes
fall back to `gbrain doctor` instead of a misleading sync command.

### 2. `content_sanity_audit_recent` counts `reject` and `quarantine` as hard failures

v0.42 renamed the hard-disposition path: rejected pages emit a `reject` event and
quarantined junk pages emit `quarantine`; `hard_block` is now only the pre-v0.42 legacy
alias. The status check only counted `hard_block`, so fresh `reject`/`quarantine` events
cleared as `ok` whenever fewer than 10 events existed. The check now sums
`hard_block + reject + quarantine` for the hard count and `soft_block + flag` for soft.

### 3. `graph_coverage` excludes test-fixture entity pages from the denominator

Brains seeded with code sources can accumulate test-fixture pages typed as `entity` or
`person` (e.g. fixtures under `tools/gbrain/test/`). Including these in the entity-count
denominator diluted coverage and produced spurious warnings. The check now computes
coverage against an eligible-entity CTE that excludes `tools/gbrain/test/*` slugs and
the `templates/new-person` stub, with an additional short-circuit for the all-fixture
case (`eligibleEntityCount = 0 → ok`).

### 4. `process.exitCode` instead of `process.exit` at doctor's main exit point

`process.exit(hasFail ? 1 : 0)` was a hard kill that prevented cleanup handlers (Bun
unload events, open DB connections) from running. `process.exitCode = hasFail ? 1 : 0`
defers termination until the end of the event loop so cleanup completes normally.

### 5. `checkSubagentCapability` exported + gateway-loop guard

The function was private, making it untestable as a standalone unit. It is now exported.

Additionally, when `agent.use_gateway_loop=true` is configured, subagents route via
the gateway and do not need `ANTHROPIC_API_KEY` directly. The previous check fired a
spurious warning for users who had a non-Anthropic `chat_model` and had the gateway
loop enabled.

---

## Tests

`bun test test/doctor.test.ts` → 77 pass, 0 fail (no regressions from any of these changes).

### DIFF
--- a/src/commands/doctor.ts
+++ b/src/commands/doctor.ts
@@ -2382,8 +2382,14 @@ export async function checkStaleLocks(
     }
     const lines = stale.slice(0, 10).map(s => {
       const ageH = Math.floor(s.age_ms / 3600_000);
-      const source = s.id.startsWith('gbrain-sync:') ? s.id.slice('gbrain-sync:'.length) : null;
-      const breakHint = source ? `gbrain sync --break-lock --source ${source}` : `gbrain sync --break-lock`;
+      let breakHint = 'gbrain doctor';
+      if (s.id.startsWith('gbrain-sync:')) {
+        breakHint = `gbrain sync --break-lock --source ${s.id.slice('gbrain-sync:'.length)}`;
+      } else if (s.id.startsWith('gbrain-cycle:')) {
+        breakHint = `gbrain dream --break-lock --source ${s.id.slice('gbrain-cycle:'.length)}`;
+      } else if (s.id === 'gbrain-cycle') {
+        breakHint = 'gbrain dream --break-lock';
+      }
       return `  ${s.id} (pid ${s.holder_pid} on ${s.holder_host}, age ${ageH}h) → ${breakHint}`;
     });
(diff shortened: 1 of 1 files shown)


time-attack:
Thanks for this, @pai-scaffolde! 🙏

What this does: PR #2075 adds five independent correctness fixes for the gbrain doctor command: improved stale-lock break hints for cycle locks, harder-failure detection for rejected/quarantined content, excluding test fixtures from entity-coverage…

Issue #1469, opened 2026-05-26
hybrid `query` returns identical results regardless of input (vector path); `search` (BM25) correct


melodjk:
## Summary
`gbrain query` (hybrid) returns the **same** top documents with **identical scores** for completely unrelated queries — including nonsense input. `gbrain search` (BM25) is query-responsive and correct. The hybrid/vector path appears to ignore the query.

## Environment
- gbrain **0.41.11.1** (also reproduced on 0.35.6.0 before upgrading)
- Engine: postgres
- Embedding: `ollama:nomic-embed-text`, 768d (`gbrain doctor`: `embedding_provider ✓`, `embedding_width_consistency` OK)
- macOS / bun

## Repro
```
gbrain query "wyckoff accumulation"     # top-3 docs A,B,C @ ~0.99/0.98/0.97
gbrain query "banana recipe quantum"    # IDENTICAL top-3 A,B,C @ IDENTICAL scores
gbrain query "..." --no-expand          # still identical
gbrain search "wyckoff accumulation"    # correct, query-responsive
gbrain search "banana recipe quantum"   # different/correct results
```
Different queries → identical `query` results **and identical cosine scores** ⇒ the query vector used in the hybrid search appears fixed/independent of the input.

## Ruled out (investigated)
- **Embedding model**: ollama `nomic-embed-text` returns distinct vectors for distinct inputs (cosine ≈0.47 between the two example queries, via both `/api/embeddings` and `/v1/embeddings`, with and without `input_type`).
- **Semantic query cache**: `gbrain cache clear --yes`, then fresh (cache-miss) queries are still identical.
- **Query expansion**: `--no-expand` still identical.
- **BM25 (`search`)**: correct and query-responsive — data + BM25 path are fine.

## Hypothesis
The query-side embedding (`embedQuery` → gateway `embed([text], {inputType:'query'})`) or its use in `core/search/hybrid.ts` may produce/use a fixed query vector. On every query this warning prints: `[ai.gateway] recipe "google" declares an embedding touchpoint without max_batch_tokens` — possibly related to query-embedding routing, even though the configured embedding model is `ollama:nomic-embed-text`.

Happy to provide more detail or test a patch.

melodjk:
# Correction / root-cause update for #1469 (gbrain "degenerate hybrid query" on Korean)

[...]

PR #2083, opened 2026-06-11
fix(gateway): thread asymmetric input_type to ZeroEntropy via request header — AI-SDK adapter drops providerOptions.openaiCompatible.input_type

victorfteha:
## Problem

`embedQuery()` exists and threads `inputType: 'query'` into `providerOptions.openaiCompatible` — but the AI-SDK's openai-compatible adapter **silently drops unrecognized providerOptions fields**. `dimensions` survives (known param); `input_type` doesn't. Every embed request therefore reached the ZeroEntropy compat shim without an `input_type`, and the shim's correct-for-ingest default (`'document'`) applied to **queries too**.

Net effect: the entire vector arm performs asymmetric retrieval with symmetric (document-typed) query vectors. Nothing errors — rankings are just quietly wrong, especially for natural-question phrasings against note-style passages (exactly what asymmetric encoding exists to fix).

## How this was found (evidence chain)

On a real ~2.3k-page / ~9.7k-chunk corpus (zembed-1 @ 1280, Postgres engine):

1. Hybrid search consistently missed targets for natural-question queries that **pure-vector SQL with a properly query-typed vector ranked #1**.
2. A manually **document-typed** query embedding reproduced production's broken rankings **byte-identically** (cosine sims equal to 16 decimal places) — production was provably searching with document-typed query vectors.
3. `gbrain`'s own `embedQuery()` output matched that document-typed vector, not the query-typed one.
4. Post-fix: 8/8 previously-missing natural-question probes rank their target #1 on the same corpus. (An unrelated dims experiment — 1280 vs native 2560 — measured a flat wash, which was the tell: dims should have moved something. The model was never the problem.)

Worth noting: `test/asymmetric-encoding-contract.test.ts` stayed **green** through all of this — it captures `providerOptions` at the `__setEmbedTransportForTests` seam, *upstream* of the adapter that drops the field. The contract it pins is real but doesn't reach the wire.

## Fix

- `embedSubBatch()` threads the input type as an **`x-gbrain-input-type` request header** — headers survive the adapter, travel with the request (race-free), and need no module state.
- `zeroEntropyCompatFetch` consumes + strips the header during its existing body rewrite and writes `input_type` into the request body.
- No API change. The Voyage path is unaffected (it builds its request body directly rather than going through the openai-compatible adapter — `input_type` already reaches the wire there).

## Test

New `test/ze-input-type-wire.test.ts` drives `embed()`/`embedQuery()` through the **real** adapter via a new `__setZeFetchForTests` terminal-fetch seam (mirrors the existing `__set*ForTests` conventions) and asserts the final wire body:

- `embedQuery(...)` → body `input_type: 'query'`, threading header stripped, URL rewrite ran
- `embed([...])` → body `input_type: 'document'`

**Verified to fail without the fix** (the query-side assertion sees `'document'`). `bun run verify`: 30/30 green. Full unit suite: **no new failures vs master** — a set of provider-env-sensitive serial tests (reranker/cross-modal/think-gateway) fails identically on a clean master checkout in my environment (host injects `ANTHROPIC_*` env vars), so I'm expecting CI's clean env to be green.

## Operational note for existing installs

Any cached query-embedding artifacts produced while the bug was live are document-typed — after upgrading, clear the query cache and restart long-running `gbrain serve` processes so they pick up the fixed code.

Possibly related: #1469 reports the hybrid vector path returning identical results regardless of input (ollama). Different symptom — a fixed/ignored query vector vs. a wrong-typed one — so this PR does not claim to fix it, but it's the same subsystem and worth re-testing against this change.

### DIFF
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -943,9 +943,11 @@ const voyageCompatFetch = (async (input: RequestInfo | URL, init?: RequestInit) 
  *  - Path: AI SDK adapter calls `${base_url}/embeddings`; ZE wants
  *    `${base_url}/models/embed`. Rewrite the URL path.
  *  - Body: inject `input_type: 'document'` (or `'query'` when threaded via
- *    providerOptions.openaiCompatible.input_type) and `encoding_format:
- *    'float'` (don't trust SDK default; strip any base64 caller injected
- *    to keep the response rewriter simple).
+ *    the `x-gbrain-input-type` request header — NOT providerOptions:
+ *    `openaiCompatible.input_type` is silently dropped by the AI-SDK
+ *    adapter) and `encoding_format: 'float'` (don't trust SDK default;
+ *    strip any base64 caller injected to keep the response rewriter
+ *    simple).
  *  - Response: ZE returns `{results: [{embedding: float[]}], usage:
  *    {total_bytes, total_tokens}}`. AI SDK's openai-compatible Zod schema
  *    expects `{data: [{embedding, index}], usage: {prompt_tokens, ...}}`.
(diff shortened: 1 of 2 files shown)


garrytan:
Superseded by #2033 (merged into the v0.42.41.0 triage wave, PR #2128), which fixes the same asymmetric `input_type` bug across the whole openai-compatible recipe class rather than ZeroEntropy-only. Thanks for the fix here — closing as superseded.

PR #2086, opened 2026-06-11
fix: bound subagent tool stalls and retry timestamps

JiraiyaETH:
## Summary
- bound subagent tool calls with a per-tool timeout and one retry through the AI gateway
- make the subagent handler pass retry/timeout policy into tool execution
- reset `started_at`/`timeout_at` when stalled jobs are retried so resurrected jobs get a fresh lease

## Tests
- `bun install --ignore-scripts`
- `bun test test/subagent-tool-timeout.test.ts test/minions.test.ts --max-concurrency=2`
- `bun run typecheck`


time-attack:
Appreciate the detail here, especially the abort-plumbing care in `executeSingleToolAttempt`. We are closing it: there is no linked issue or reproduction of a tool call actually stalling past the existing job-level bounds (`handleStalled` → `handleTimeouts` → `handleWallClockTimeouts`), and the lease-reset piece is already on master in `retryJob` (#2783). The part that gives us most pause is reusing the `idempotent` flag — today it only gates crash-replay re-execution — as a live retry signal, which risks double side effects. If you can attach a concrete stalled-job trace, a narrow timeout with no retry layer would be a much easier conversation.

PR #2090, opened 2026-06-11
feat(minimax): add chat touchpoint (MiniMax-M3 / M2.5) with tool calling

rafaelreis-r:
## What

Adds a `chat` touchpoint to the MiniMax recipe (currently embeddings-only).

## Why

The capabilities hint in `src/core/ai/capabilities.ts` already lists minimax among "Known providers with chat", but the recipe never declared the touchpoint — so `minimax:<model>` fails at `getProviderCapabilities()` with *does not offer a chat touchpoint*, and the model can't be used for `models.tier.subagent` / dream synthesis via the gateway tool loop. This makes the recipe match the hint.

## How

- Same OpenAI-compatible `base_url` the embedding touchpoint already uses.
- `supports_tools: true` — MiniMax's `/chat/completions` supports function calling, which is what `agent.use_gateway_loop` needs.
- `supports_prompt_cache: false` — Anthropic-style `cache_control` is only honored on their anthropic-compatible endpoint, not the OpenAI-compatible one.
- Models list (`MiniMax-M3`, `MiniMax-M2.5`) is informational, same contract as the openrouter recipe (openai-compat tier doesn't enforce the list at runtime).
- `max_context_tokens: 200_000` declares the conservative floor (M2.5); M3 advertises 1M, handled per-model by callers.

## Testing

Verified against `api.minimax.io` with `MiniMax-M3`: tool calling and plain text both return correctly (we run dream synthesis on it in production on our install).

Context: we run gbrain's dream pipeline against MiniMax in production after our Anthropic OAuth proxy setup started rejecting raw-SDK subagent traffic — happy to add anything else needed here.

### DIFF
--- a/src/core/ai/recipes/minimax.ts
+++ b/src/core/ai/recipes/minimax.ts
@@ -38,6 +38,28 @@ export const minimax: Recipe = {
       // halving in the gateway catches token-limit errors at runtime.
       max_batch_tokens: 4096,
     },
+    chat: {
+      // MiniMax text models via the OpenAI-compatible /chat/completions
+      // endpoint (same base_url). Function calling is supported, which is
+      // what the gateway tool loop needs. The capabilities hint in
+      // src/core/ai/capabilities.ts already lists minimax among chat
+      // providers — this touchpoint makes the recipe match it.
+      // openai-compat tier: the models list is informational, arbitrary
+      // ids are accepted at runtime (same contract as openrouter).
+      models: ['MiniMax-M3', 'MiniMax-M2.5'],
+      supports_tools: true,
+      supports_subagent_loop: true,
+      // Anthropic-style cache_control is only honored on MiniMax's
+      // anthropic-compatible endpoint, not the OpenAI-compatible one.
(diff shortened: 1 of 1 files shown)


rafaelreis-r:
@garrytan friendly ping for review on this one — it's been validated in production for a day now (MiniMax-M3 chat touchpoint driving dream synthesize + verdicts; 41 verdicts in ~100s, zero 429s, prompt-cache honored via anthropic-compat). Happy to adjust anything.

time-attack:
Thanks @rafaelreis-r! 🙏

What this does: PR #2090 adds chat endpoint configuration to the MiniMax recipe, enabling MiniMax-M3 and M2.5 models for synthesis and reasoning tasks. PR #2220 adds the identical chat touchpoint but as part of a larger full-stack implementation that also…

This overlaps with #2220, which we're carrying forward, so I'm closing this in its favor to keep the review queue focused. If yours covers a case #2220 misses, reopen with that detail and we'll reconcile them.

Issue #2058, opened 2026-06-10
pglite-lock force-removes locks of LIVE processes after 5 minutes (no heartbeat) — concurrent serve processes share the single-writer data dir; WAL corruption observed


na-dev-12:
Found during a deep audit of gbrain v0.42.26.0 after a real PGLite WAL corruption event.

## Mechanics (src/core/pglite-lock.ts)
- `STALE_THRESHOLD_MS = 5 min` (line ~22); the acquire loop force-removes a lock whose holder PID is **alive** once the lock is older than 5 minutes ("Still alive but probably stuck — force remove", lines ~82-85).
- `acquired_at` is written once at acquisition (lines ~102-106) and never refreshed — there is no heartbeat, so **every** healthy long-lived holder is classified stale after 5 minutes.
- `releaseLock` (lines ~140-148) `rmSync`s the shared lock dir without checking ownership, so a finishing short-lived CLI process deletes whichever holder's marker is present.

## Consequence
`gbrain serve` connects the engine at startup and holds it for the process lifetime (cli.ts "serve doesn't disconnect"). With per-client stdio MCP registration (Claude Desktop + every Claude Code/Codex session spawns its own serve), serve #2 starting >5 min after serve #1 steals the lock, and so on. Observed live: **4 concurrent `gbrain serve` processes each holding 40+ read-write FDs inside the same `brain.pglite` data dir (two sharing the same WAL segment), with no `.gbrain-lock` dir present at all** — exactly the concurrent access the module's own header says PGLite cannot survive (`Aborted()`). We experienced a WAL/checkpoint corruption that required a full rename-aside + reimport, plus recurring "PGLite lock contention persisted after 3 attempts" failures in scheduled sync runs.

## Suggested fix
Mirror the design already present in `src/core/db-lock.ts` (`last_refreshed_at`, lines ~154-160):
1. Heartbeat: refresh a timestamp in the lock file every ~60s (unref'd interval) while the engine is connected.
2. Steal only when the holder PID is dead OR the heartbeat (not `acquired_at`) is stale.
3. `releaseLock` should verify the lock file's pid === process.pid before `rmSync`.

Note the behavior change: a second stdio serve then fails fast with the actionable 30s timeout instead of silently sharing the data dir — correct for a single-writer database, and it makes the documented `serve --http` single-daemon mode the explicit multi-client answer.

time-attack:
[...]

Issue #1981, opened 2026-06-08
Retrieval Reflex: teach the agent when/what to retrieve (default behavior when gbrain installed)


garrytan:
# Feature: Retrieval Reflex — teach the agent *when* to retrieve and *what* to retrieve (default behavior when gbrain is installed)

## Problem

gbrain is great at **storing** knowledge and at **injecting deterministic context** (the `gbrain-context` engine injects temporal/spatial state every turn, zero LLM calls). But it does **not** teach the host agent the *policy* of retrieval: **when** to look something up, and **what** to pull when it does.

Today that policy lives in each user's hand-rolled agent instructions (e.g. an AGENTS.md "on any proper noun, run a brain lookup" gate). That is fragile and inconsistent, and it fails in exactly the way a human never would.

**Motivating failure (real, 2026-06-08):** the agent discussed the user's closest friend — a person with a rich, well-maintained brain page — for **four messages** without ever opening that page, then gave generic answers about facts the brain already knew. The *write* side (ambient capture into the brain) worked perfectly; the *read* side (recall at the right moment) had no enforced policy, so it didn't fire.

**The human model we want to encode** (user's framing): *"If you see a name you don't recognize, the brain probably has the data — do a quick retrieval for it."* A person doesn't bulk-load their whole address book into working memory on every conversation. They retrieve **on demand**, when an entity becomes **salient**, and then drop it. We want gbrain to teach the agent that reflex.

## Anti-goal (equally important)

The naive fix — warm-load the user's whole inner circle / top-N entities into context every turn — is **wrong**. It pollutes context, you don't know the task in advance, and it scales terribly. The reflex must be **on-demand, salience-triggered, and judgment-gated**, not a standing cache. (We tried the warm-load version first and the user correctly rejected it.)

## Proposed solution

Add a **Retrieval Reflex** to gbrain — a first-class, installed-by-default policy that ships with the gbrain-context engine and tells the host agent two things: a **trigger policy** (when to retrieve) and a **retrieval spec** (what to pull, how much, and when to stop). Two delivery layers, because gbrain supports both:

### 1. Deterministic layer — entity salience detector in the context engine
Extend `core/context-engine.ts` `assemble()` to do a cheap, **zero-LLM** scan of the current turn's user message(s) for candidate entities (capitalized tokens, known-alias matches, `@handles`), then:
- Resolve each candidate against the brain index (the same hybrid `query` path, but a fast metadata/alias lookup, not a full semantic pull).
[...]

PR #2092, opened 2026-06-11
fix(serve): answer the MCP initialize handshake before the engine connects — lazy PGLite boot, retryable lock contention (#2091)

### LINKED ISSUE #2091: serve (stdio): engine connects eagerly before the MCP initialize handshake — slow PGLite boot / lock contention kills the host connection at its 30s timeout, and shutdown racing an in-flight connect leaks the lock dir
## Summary

`gbrain serve` (stdio) cannot answer the MCP `initialize` handshake until the engine is fully connected, because `cli.ts` awaits `connectWithRetry(engine, ...)` before command dispatch (`src/cli.ts` ~1992-1997). For the PGLite engine, "connected" means the full WASM Postgres boot **plus** advisory write-lock acquisition (`src/core/pglite-engine.ts:217` → `acquireLock`, which itself waits up to 30s on contention).

MCP hosts enforce a connect timeout on `initialize` — Claude Code's default is **30s**. Whenever PGLite boot + lock-wait exceeds that, the host kills the server and reports an opaque `connection timed out after 30000ms`. The user never sees the real reason (slow WASM boot, or "another serve holds the write lock").

Related but distinct from #677 (cooperative single-owner mode) and #2058 (no lock heartbeat / live-holder steal): this is about the *handshake ordering* — even with a perfect lock protocol, any engine boot slower than the host timeout makes stdio MCP registration unusable, and lock contention surfaces as a host-side timeout instead of an actionable error.

## Environment

- gbrain `0.42.37.0`, PGLite engine (~2k pages / 6.6k chunks)
- Bun 1.4.0, Windows 11
- MCP host: Claude Code (stdio transport via a `.bat` wrapper, 30s connect timeout)

## Observed timeline (real incident, UTC)

| Time | Event |
|---|---|
| 13:42:36 | Claude Code session A spawns `gbrain serve` (stdio), 30s connect timer starts |
| 13:42:38 | Claude Code session B (second window) spawns its own `gbrain serve` |
| 13:43:06 | Session B: `Connection timeout triggered after 30038ms` → host kills its server |
| 13:43:08 | Session A: same timeout → host kills its server |
| 13:43:16 | `.gbrain-lock/lock` written by one of the (now-orphaned) serve processes — **8-10s after both parents were killed** |
| later | Lock dir left behind; holder PID dead; next boot pays the stale-lock cleanup path |

Two engines booting PGLite WASM concurrently (CPU contention) plus the loser waiting in `acquireLock`'s retry loop pushed both past 30s. Neither ever got to answer `initialize`.

## Measurements

Time from spawn to `initialize` response on the same brain, same machine (warm FS cache, single instance):

| Config | time-to-initialize |
|---|---|
| Bun default (JIT on) | **~2.3s** |
| `BUN_JSC_useJIT=0` (interpreted WASM) | **~16s** |

The JIT-off row is a local quirk of this machine's wrapper, but it demonstrates the variance class: cold cache, large brains, slow disks, CPU contention, or a second session waiting on the lock can all push an *eager* connect past a host's handshake deadline. (For scale: `acquireLock`'s own default wait is 30s — identical to the host timeout, so a locked brain is *guaranteed* to lose this race.)

## Secondary bug: shutdown racing an in-flight connect leaks the lock dir

The 13:43:16 lock write above happened **after** shutdown had already begun. Sequence:

1. Host kills the `.bat`/`cmd.exe` wrapper → stdin `close` fires (and/or the #2049 parent watchdog catches it) → `beginShutdown` → `engine.disconnect()`.
2. But `engine.connect()` is still in flight: `this._lock` isn't set yet, so `disconnect()` releases nothing and the process exits.
3. The in-flight `acquireLock` then completes (or completed between the disconnect check and exit), writing a lock dir that no live process owns and nothing ever releases.

It self-heals on the next boot via the dead-PID stale check, but under Windows PID recycling the next boot can mistake the recycled PID for a live holder and burn the full 5-minute `STALE_THRESHOLD_MS` (or, post-#2058, whatever replaces it). `beginShutdown` should await/cancel an in-flight connect before calling `disconnect()` (or `connect()` should check a shutdown flag before and after `acquireLock` and release immediately).

## Proposed fix

Decouple the MCP handshake from engine readiness for the stdio path:

1. **Answer `initialize` (and `tools/list`) immediately.** The tool list is generated from `src/core/operations.ts` — no DB needed. Start the engine connect in the background when `serve` dispatches.
2. **First `tools/call` awaits engine-ready** (with a generous internal deadline). If the engine can't connect because the write lock is held, return a structured tool error — e.g. `brain is locked by pid N (command: ...)` — instead of letting the host time out the whole server. The host stays connected, the user sees an actionable message, and a later call can retry once the lock frees.
3. **Close the shutdown/connect race** so a killed-during-boot serve never leaves a lock dir behind.

This would also make the multi-session story (per-client stdio registration spawning N serves — the exact topology #2058 observed) degrade gracefully: extra sessions connect fine at the MCP layer and get a clear per-call error, rather than dying as opaque host timeouts.

Happy to PR this (follow-up to #2049, which this incident's lock-leak analysis builds on).

abyss-node:
Fixes #2091. Follow-up in spirit to #2049 (same incident family: MCP hosts killing/orphaning stdio serves).

## Problem

The stdio serve path received an eagerly-connected engine (`cli.ts` awaited `connectEngine()` before dispatch), so the MCP `initialize` handshake could not be answered until the full PGLite WASM boot **plus** advisory write-lock acquisition (itself up to 30s under contention) had finished. MCP hosts enforce a connect timeout on the handshake (Claude Code: 30s), which produced two failure modes documented with a real-incident timeline in #2091:

1. Any boot slower than the host timeout → opaque `connection timed out after 30000ms`, real cause invisible. A second concurrent host session was *guaranteed* to lose: `acquireLock`'s 30s wait equals the host's 30s deadline.
2. The killed-at-timeout orphan leaked the lock dir when shutdown raced the still-in-flight connect — `engine.disconnect()` ran before `this._lock` was set, then the in-flight `acquireLock` completed with nobody left to release it (observed: lock written 8–10s after the host killed the parent).

## Fix

New `src/core/engine-provider.ts` — a small lazy provider with three load-bearing behaviors:

- **Handshake decoupled from boot.** `cli.ts` dispatches `serve` before the eager connect and hands `runServe` a lazy connector. The stdio path answers `initialize` / `tools/list` immediately (the tool list is generated from `operations.ts`, no DB needed) while the engine connects in the background. Cold boot, big brains, and lock waits no longer race the host's deadline.
- **Lock contention becomes a per-call, retryable error.** Each `tools/call` awaits `provider.ensure()`. A connect failure returns a structured `engine_unavailable` tool error (with the underlying message, e.g. the lock-holder diagnostics from `pglite-lock.ts`) and `ensure()` un-memoizes rejected attempts so the next call retries. Per-client stdio registration spawning N serves (the topology #2058 observed) now degrades gracefully: extra sessions stay connected at the MCP layer and see an actionable error instead of dying as host-side timeouts.
- **Shutdown-during-boot race closed.** `provider.disconnect()` awaits any in-flight connect before tearing down, so the lock the connect acquires is the lock the shutdown releases. Shutdown on a never-started provider is a no-op (a serve that dies before first use no longer boots the engine just to tear it down). The existing 5s cleanup deadline still bounds a connect wedged on lock contention — if it fires, the lock either wasn't acquired yet or is reclaimed by the dead-PID stale check.

Scope notes:
- `--http` keeps the eager connect (its hosts dial TCP after the listener is up; no handshake deadline).
- The Retrieval Reflex IPC socket (#1981) binds once the background connect lands — it needs a live engine, and binding earlier would advertise a resolver that can only fail.
- Bare-engine callers of `startMcpServer` / `runServe` (tests, legacy paths) are wrapped in a pre-resolved provider; behavior unchanged.

## Tests

- New `test/serve-lazy-engine-connect.test.ts` (12 cases): provider memoization / retry-after-failure / `disconnect`-awaits-in-flight-connect semantics, plus `runServe` stdio integration with a deferred connector — including the shutdown-racing-connect lock-leak regression and the connector-rejection → clean-exit path. Same stubbing discipline as `serve-stdio-lifecycle.test.ts` (no real Bun child, no real MCP SDK).
- `test/serve-stdio-lifecycle.test.ts`: 24 pass unchanged.
- `test/e2e/serve-stdio-roundtrip.test.ts` (real spawned serve against a fresh PGLite brain, full `initialize` → `tools/list` → `tools/call`): 3 pass.
- `bun run typecheck` clean; `bun run build:llms` regenerated (no content drift — KEY_FILES is linked, not inlined).

`docs/architecture/KEY_FILES.md` entries for `serve.ts` / `mcp/server.ts` updated to current behavior and a new entry added for `engine-provider.ts`.


eloe:
Opened a rebased replacement against current master: #2456. It keeps the lazy stdio engine-connect fix from this PR and resolves the current master conflicts, including the latest Retrieval Reflex IPC suppression/delivery logging behavior and KEY_FILES current-state docs.

time-attack:
Confirmed against master: `cli.ts` still awaits `connectEngine()` before dispatching `serve`, so the stdio `initialize` handshake sits behind the PGLite boot and the write-lock wait — a second concurrent session is deterministically lost at the host's 30s deadline, exactly as #2091 describes. Merging: this fixes a broken core promise, the new `engine-provider.ts` is small and single-purpose, `--http` correctly keeps the eager connect, and the shutdown-awaits-in-flight-connect case that leaked the lock dir is covered by tests. Thanks for the incident timeline and for updating KEY_FILES to current-state rather than appending release narration.

PR #2093, opened 2026-06-11
fix(pglite): heartbeat file lock without stealing live holders (#2058)

### LINKED ISSUE #2058: pglite-lock force-removes locks of LIVE processes after 5 minutes (no heartbeat) — concurrent serve processes share the single-writer data dir; WAL corruption observed

venturejakef:
## What & why

Closes #2058.

The current PGLite lock can remove a lock from a still-live holder once `acquired_at` crosses the stale threshold. That is unsafe for embedded PGLite: a paused or starved process can still hold open database files even if its JS timers are not making progress, so a second process deleting the lock can recreate concurrent access/WAL corruption risk.

Before this PR, a second process could delete a live holder's lock after five minutes. After this PR, stale heartbeat metadata is diagnostic only; the lock is not recovered unless the same-host PID is provably dead and past a short PID-reuse grace window, or the lock directory is an old orphan with missing/corrupt metadata.

## Why this shape

This intentionally does not use a stale heartbeat as a takeover signal. For PGLite, a starved or paused JS process can still hold database files open, so heartbeat age is useful diagnostic data but not proof that it is safe for another process to enter.

The safety choices here are meant to close the race windows around that:

- `owner_id` prevents an old handle from releasing a newer lock created by the same PID.
- `last_refreshed_at` gives waiters better diagnostics without becoming an unsafe steal condition.
- Same-host dead-PID recovery keeps normal crash recovery working, but the short grace window reduces PID-reuse risk.
- Cross-host, `EPERM`, and unknown PID-probe results fail closed because they cannot prove the holder is gone.
- Quarantine rename avoids deleting the active lock directory after reading stale metadata.

## The fix

- Add per-acquisition ownership metadata: `owner_id`, `pid`, `host`, `acquired_at`, `last_refreshed_at`, and `command`.
- Add an owner-checked heartbeat that refreshes `last_refreshed_at` via temp-file write plus atomic rename.
- Fail closed for live PIDs, stale heartbeats, cross-host metadata, `EPERM`, and unknown PID probe results.
- Recover only dead same-host PIDs after a grace window, or old missing/corrupt orphan metadata after initialization grace.
- Replace direct lock-directory deletion during recovery with an atomic quarantine rename followed by best-effort removal.
- Make release owner-checked so a stale handle cannot remove a replacement lock from the same PID.

## Scope decisions

- No release metadata (`VERSION`, `CHANGELOG.md`, `package.json`).
- No public exports, CLI flags, docs, or generated `llms` files.
- No changes to `db-lock.ts` or `pglite-engine.ts`.
- Test seams are internal to `acquireLock` options only.

## Tests

Added deterministic coverage for:

- live PID with old `acquired_at` and fresh `last_refreshed_at` is not stolen
- live PID with stale `last_refreshed_at` is not stolen
- old-format live lock without heartbeat metadata is not force-stolen
- dead same-host PID past grace is cleaned up
- dead same-host PID inside grace is not cleaned up
- unknown and EPERM-equivalent PID probe results are treated as unsafe/alive
- cross-host locks are not auto-stolen
- fresh missing metadata inside initialization grace is not removed
- old corrupt metadata is recovered as orphaned
- acquired locks write `owner_id`, `host`, and `last_refreshed_at`
- heartbeat refreshes `last_refreshed_at` and stops after release
- release removes own lock
- release does not remove same-PID different-owner lock
- release no-ops for corrupt, missing-file, and missing-directory lock state

## Verification

- `bun test test/pglite-lock.test.ts` — 19 pass, 0 fail
- Docker: `docker compose -f docker-compose.ci.yml run --rm --no-deps runner bash -lc 'bun test test/pglite-lock.test.ts'` — 19 pass, 0 fail
- `bun run typecheck` — pass
- `bun run verify` — 30 pass, 0 fail
- `gitleaks detect --source . --redact --no-banner` — no leaks found

I also ran `bun run ci:local -- --no-pull` against both this branch and a clean upstream checkout under local Docker/Colima. Both runs reproduced the same current local-Docker baseline failures, including the same `validateUploadPath` fuzz counterexample (`./../../../tmp`) and shard 3 exit 137, so I did not treat that as branch-specific evidence.

### DIFF
--- a/src/core/pglite-lock.ts
+++ b/src/core/pglite-lock.ts
@@ -14,16 +14,48 @@
  *   try { ... } finally { await releaseLock(lock); }
  */
 
-import { mkdirSync, existsSync, readFileSync, writeFileSync, rmSync, statSync } from 'fs';
+import { randomBytes } from 'crypto';
+import { mkdirSync, existsSync, readFileSync, writeFileSync, rmSync, statSync, renameSync } from 'fs';
+import { hostname } from 'os';
 import { join } from 'path';
 
 const LOCK_DIR_NAME = '.gbrain-lock';
 const LOCK_FILE = 'lock';
 const STALE_THRESHOLD_MS = 5 * 60 * 1000; // 5 minutes — embed jobs can be long
+const HEARTBEAT_INTERVAL_MS = 60 * 1000;
+const INIT_GRACE_MS = 1000;
+const DEAD_PID_GRACE_MS = 60 * 1000;
+const WAIT_RETRY_MS = 50;
+
+type ProcessState = 'alive' | 'dead' | 'unknown';
+
+interface LockRecord {
+  pid: number;
+  acquired_at: number;
+  command?: string;
+  owner_id?: string;
+  host?: string;
+  last_refreshed_at?: number;
+}
+
+interface AcquireLockOpts {
+  timeoutMs?: number;
+  heartbeatIntervalMs?: number;
+  staleThresholdMs?: number;
+  initGraceMs?: number;
+  deadPidGraceMs?: number;
+  now?: () => number;
+  isProcessAlive?: (pid: number) => ProcessState;
+  host?: string;
+}
 
 export interface LockHandle {
   lockDir: string;
   acquired: boolean;
+  ownerId?: string;
+  pid?: number;
+  host?: string;
+  heartbeat?: ReturnType<typeof setInterval>;
 }
 
 function getLockDir(dataDir: string | undefined): string {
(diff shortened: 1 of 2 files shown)


venturejakef:
Closing this pre-v0.42.41 version in favor of the smaller current-master follow-up in #2154. The replacement keeps Garry's merged #2058 heartbeat/owner-token shape and only fixes the remaining stale-heartbeat + live-PID takeover edge.

PR #2096, opened 2026-06-11
feat(oauth): opt-in Google Workspace login gate for the browser authorization_code flow

zaycruz:
## What

Adds an **opt-in** Google-Workspace login gate to the HTTP server's OAuth `authorization_code` flow. When enabled, a browser-driven MCP client (Claude Code, Cursor, etc.) doing the remote-MCP OAuth dance must sign in with a Google account on an allowed hosted domain (`hd`) **before** `/authorize` issues an authorization code.

## Why

Today `oauth-provider.ts authorize()` issues a code to any caller, and `/register` (DCR) is unauthenticated. So turning on `--enable-dcr` for a publicly-reachable deployment means anyone can self-register and read the brain. This gate makes DCR safe to enable for a team deployment: only signed-in users on your Workspace domain can complete the flow.

## Backward compatibility

100% opt-in. Dormant unless `--oauth-login-hd <domain>` is set (each flag also reads a `GBRAIN_OAUTH_*` env var so secrets can be supplied out-of-band, e.g. via a secret manager rather than argv). With it unset, behavior is byte-for-byte unchanged — verified against the existing oauth test suites.

## How

New `src/core/oauth-google-login.ts` (dependency-light — no `jose`; RS256 verified with `node:crypto` against Google's JWKS):
- `GET /login/google` + `GET /login/google/callback`; gate middleware on `/authorize` registered before `mcpAuthRouter`.
- `id_token` verification: RS256 signature via JWKS, plus `iss`/`aud`/`exp` (with skew)/`hd`/`email_verified` assertions. The id_token from Google's `/token` is verified, not trusted.
- **PKCE (S256) + OIDC `nonce`** on the server↔Google leg, bound to the browser via a signed `HttpOnly; Secure; SameSite=Lax` cookie (closes state-replay / login-CSRF / session-fixation).
- Signed, expiring `state` as the return-url carrier with an **open-redirect guard** (same-origin + `/authorize`-prefixed).
- Session cookie `{email, exp}`, HMAC-signed, constant-time compare.

**DCR hardening (only when the gate is enabled):** every self-registered `/register` client is forced to public (`token_endpoint_auth_method=none`) + `authorization_code`/`refresh_token` + `read` scope, stripping `client_credentials` so a self-registered client cannot mint a privileged token via `/token` and bypass `/authorize`. The operator path (`registerClientManual`) is untouched, so admin-minted machine clients keep full grants/scope.

Also adds `resource_metadata` to the `/mcp` 401 `WWW-Authenticate` (RFC 9728 discovery hint).

## Tests

New unit + integration tests (all green; full suite shows only pre-existing env-gated failures): id_token verification (wrong aud/iss/expired/wrong hd/`email_verified`=false/bad sig/bad alg/unknown kid/nonce mismatch), state/session/PKCE sign+verify with tamper+expiry, open-redirect guard, gate middleware (off→passthrough, on+no-cookie→redirect, on+valid→passthrough), and the DCR clamp including the confidential-`client_credentials`-admin bypass case.

## Config

```
serve --http --oauth-login-hd your-workspace.com \
  --oauth-google-client-id <id> \
  --oauth-google-client-secret <secret> \
  --oauth-session-secret <random> \
  --enable-dcr
```

Happy to adjust naming/version/CHANGELOG to match your release process.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -2236,6 +2236,11 @@ ADMIN
     --token-ttl N                    Access token TTL in seconds (default: 3600)
     --enable-dcr                     Enable Dynamic Client Registration
     --public-url URL                 Public issuer URL (required behind proxy/tunnel)
+    --oauth-login-hd DOMAIN          Require Google sign-in on this hosted domain
+                                     before /authorize issues a code (opt-in gate)
+    --oauth-google-client-id ID      Google OAuth client id   (required with --oauth-login-hd)
+    --oauth-google-client-secret S   Google OAuth client secret (required with --oauth-login-hd)
+    --oauth-session-secret SECRET    HMAC key for login session+state cookies (required with --oauth-login-hd)
   connect <mcp-url> --token <t>      Wire Claude Code to a remote gbrain (bearer token)
         [--install] [--json]         Print the paste-ready command, or --install to run it
   call <tool> '<json>'               Raw tool invocation
--- a/src/commands/serve-http.ts
+++ b/src/commands/serve-http.ts
@@ -27,6 +27,29 @@ import { operations, OperationError } from '../core/operations.ts';
 import type { OperationContext, AuthInfo } from '../core/operations.ts';
 import { GBrainOAuthProvider, validateTokenEndpointAuthMethod } from '../core/oauth-provider.ts';
 import type { SqlQuery } from '../core/oauth-provider.ts';
+import {
+  resolveGoogleLoginConfig,
+  buildGoogleAuthUrl,
+  signState,
+  verifyState,
+  signSession,
+  verifySession,
+  signPkce,
+  verifyPkce,
+  generateCodeVerifier,
+  codeChallengeS256,
+  generateNonce,
+  safeReturnUrl,
+  exchangeGoogleCode,
+  verifyGoogleIdToken,
+  GoogleJwksCache,
+  LOGIN_SESSION_COOKIE,
+  LOGIN_PKCE_COOKIE,
+  SESSION_TTL_SECONDS,
+  STATE_TTL_SECONDS,
+} from '../core/oauth-google-login.ts';
+import type { GoogleLoginConfig } from '../core/oauth-google-login.ts';
+import { randomUUID } from 'crypto';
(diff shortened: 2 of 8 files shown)


zaycruz:
Ran a multi-agent code review (security / correctness / adversarial / testing / maintainability) on this branch. **No P0/P1 — no working bypass found**; all three of security/correctness/adversarial independently confirmed the gate holds (RS256 alg-pinning rejects none/HS256 confusion, JWKS sig verified before claims, full iss/aud/exp/hd/email_verified/nonce checks, HMAC cookies constant-time, open-redirect guard, gate runs before the SDK router with identical path matching, fails closed).

Addressed in the latest commit:
- **Backward-compat:** the DCR public-client `read` clamp now fires only when the gate is enabled, so gate-OFF behavior is byte-for-byte upstream (was clamping unconditionally).
- **Comment accuracy:** noted the DCR hardening is registration-time, not retroactive (a confidential `client_credentials` client registered before the gate was enabled keeps working at `/token`; enable from the start or revoke pre-existing DCR clients).
- **Tests:** added explicit `alg=HS256` and `email_verified`-absent id_token rejection cases.
- Deduped a duplicate `crypto` import.

Optional follow-ups noted but not applied (happy to if you'd prefer): extract the three near-identical `sign*/verify*` HMAC helper pairs into a generic `signPayload/verifyPayload`; lift the ~90-line inline `/login/google/callback` handler to a named module-level function for unit-testability; add HTTP-level tests for the callback rejection paths. Left these out to keep the security-verified code stable — your call on house-style preference.

time-attack:
This is clearly careful work — verifying the id_token against Google's JWKS rather than trusting it, PKCE plus nonce on the server-to-Google leg, and the open-redirect guard all show real attention. We still have to close it: it binds a vendor-specific identity provider and roughly 1800 lines of new auth surface into the serve path, and third-party integrations of that size are outside what this project takes on, opt-in or not. If the concern is that `--enable-dcr` is too permissive on a public deployment, a small PR tightening or gating DCR itself — no external IdP involved — is the version we would read closely.

Issue #1292, opened 2026-05-22
fix(gateway): isAvailable returns false for litellm when user specifies a model id


ricky86102:
### Problem

When using the `litellm` recipe with a concrete embedding model
(e.g. `litellm:bge-large`), `isAvailable('embedding')` incorrectly
returns `false`. This causes hybrid search to silently fall back to
keyword-only mode (`vector_enabled: false` in search meta), which in
`conservative` mode returns no results — even though vectors are
correctly stored and direct SQL cosine search works fine.

**Symptom:** `gbrain query "..."` returns `No results.` despite the
embedding pipeline being fully functional.

**Affected users:** Anyone using gbrain with a LiteLLM proxy in front
of Bedrock, Vertex AI, Azure OpenAI, or any other OpenAI-compatible
endpoint.

### Root Cause

The empty-models guard in `isAvailable()`:

    if (
      Array.isArray(touchpointConfig.models) &&
      touchpointConfig.models.length === 0 &&
      (recipe.id === 'litellm' || isUserProvided)
    ) return false;

The `litellm` recipe declares `models: []` and
`user_provided_models: true` — meaning "the user will supply the model
id". The guard was designed to catch the case where the user *hasn't*
supplied one yet. But it fired unconditionally, even when
`embedding_model` was already set to `litellm:bge-large`.

### Fix

Extract `parsed` from `resolveRecipe()` and add
`!hasUserSpecifiedModel` to the guard. When `parsed.modelId` is
non-empty, the `user_provided_models` contract is already satisfied —
skip the guard.

    const { parsed, recipe } = resolveRecipe(modelStr);
    // ...
    const hasUserSpecifiedModel = parsed.modelId.length > 0;
    if (
      Array.isArray(touchpointConfig.models) &&
      touchpointConfig.models.length === 0 &&
      (recipe.id === 'litellm' || isUserProvided) &&
      !hasUserSpecifiedModel
    ) return false;

### Tests

Three new cases in `test/ai/gateway.test.ts`:
- `litellm:bge-large` → `isAvailable('embedding')` returns `true` (was `false`)
- `litellm:azure:text-embedding-ada-002` (compound model id) → `true`
- No `LITELLM_API_KEY` env var → still `true` (key is declared `optional` in the recipe)

time-attack:
Closing as fixed on master (see a25209bb / #2627). Please reopen if still reproducible on current HEAD.

PR #2107, opened 2026-06-11
fix: allow user-provided llama-server embedding model

### LINKED ISSUE #1740: embedding preflight: user_provided_model_unset rejects valid user-specified modelId

maxpetrusenkoagent:
## Summary
- fix embedding preflight for user-provided llama-server embedding model ids
- add regression coverage for `llama-server:text-embedding-qwen3-embedding-0.6b`

Fixes #1740

## Verification
- `bun test test/embed-preflight.test.ts --test-name-pattern 'passes for llama-server'` ✅
- `bun test test/embed-preflight.test.ts` ✅
- `bun run typecheck` ✅
- `bun run verify` ⚠️ 29/30 passed locally; `check:wasm` failed in this worktree with `bun build --compile` killed/exit 137 after the machine reported the Data volume at 100% capacity. Focused tests and typecheck pass.

### DIFF
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -709,7 +709,8 @@ export function diagnoseEmbedding(modelOverride?: string): EmbeddingDiagnosis {
   if (
     Array.isArray(tp.models) &&
     tp.models.length === 0 &&
-    (recipe.id === 'litellm' || isUserProvided)
+    (recipe.id === 'litellm' || isUserProvided) &&
+    !parsed.modelId
   ) {
     return {
       ok: false,
--- a/test/embed-preflight.test.ts
+++ b/test/embed-preflight.test.ts
@@ -6,7 +6,7 @@
  * process.env.
  */
 import { describe, test, expect, beforeEach, afterAll } from 'bun:test';
-import { configureGateway, resetGateway } from '../src/core/ai/gateway.ts';
+import { configureGateway, diagnoseEmbedding, resetGateway } from '../src/core/ai/gateway.ts';
 import {
   validateEmbeddingCreds,
   formatEmbeddingCredsError,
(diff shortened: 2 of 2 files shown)


time-attack:
Thanks for the regression test and the clear repro on the llama-server model id. That preflight branch has since been removed on master — diagnoseEmbedding's old user_provided_model_unset check was replaced by the #1292/D6 dims guard, so a fully-specified llama-server:<model> with embedding_dimensions set passes today. Closing as already fixed; if you still see a false negative against current master, a fresh issue with the exact config would be welcome.

PR #2112, opened 2026-06-12
Fix gateway subagent replay reconciliation

### LINKED ISSUE #2110: Gateway-native tool loop loses tool results on deepseek parallel tool calls; subagent model config keys inconsistent (models.subagent vs models.tier.subagent)
## Summary
With `agent.use_gateway_loop=true`, subagent jobs running on `openrouter:deepseek/deepseek-v4-flash` die with:

```
[chat(openrouter:deepseek/deepseek-v4-flash)] Tool results are missing for tool calls call_xxx, call_yyy.
```

The error message users hit first (`subagent job: resolved model ... is non-Anthropic but agent.use_gateway_loop is not enabled`) explicitly recommends enabling the gateway loop for non-Anthropic providers — but on deepseek the gateway loop then fails mid-conversation when the model emits parallel tool calls and the loop doesn't return results for all of them. Observed 6/8 dream-pipeline subagent jobs dead with this signature (v0.42.26.0).

## Repro
1. `gbrain config set models.subagent openrouter:deepseek/deepseek-v4-flash` (or models.default resolving there)
2. `gbrain config set agent.use_gateway_loop true`
3. Let the dream cycle enqueue a subagent job (multi-turn, tool-using synthesis prompt)
4. Job dead-letters after retries with "Tool results are missing for tool calls …"

## Expected
gateway.toolLoop() should return a tool_result for every tool_call id in the assistant turn (including parallel tool calls), or the gateway loop docs/error text should not recommend providers it can't drive.

## Side note — config key inconsistency
While remediating: the subagent handler reads `models.subagent` (`src/core/minions/handlers/subagent.ts:211` → `resolveModel(..., configKey: 'models.subagent')`) but the `subagent_capability` doctor hint tells users to set `models.tier.subagent`. One of the two should be canonical. Also `gbrain config set agent.use_gateway_loop true` warns `Nothing in gbrain reads this` (unregistered key) even though the subagent handler reads exactly that key via `engine.getConfig('agent.use_gateway_loop')`.

## Workaround used
Reverted the gateway loop and moved subagent jobs to `anthropic:claude-sonnet-4-6` (set on BOTH `models.subagent` and `models.tier.subagent`).

maxpetrusenkoagent:
Fixes #2110

## Summary
- reconcile replayed assistant tool-call turns before the next gateway model call so providers like DeepSeek receive matching tool results
- reuse existing legacy/v2 tool execution rows during replay instead of inserting duplicate pending rows
- align subagent config handling around `models.subagent` and register `agent.use_gateway_loop` as a known config key

## Verification
- `bun test test/ai/gateway-tool-loop.test.ts test/gateway-model-messages.test.ts test/model-config.serial.test.ts test/config-env.test.ts test/e2e/subagent-gateway-path.test.ts test/e2e/subagent-crash-replay-multi-provider.test.ts`
- `bun run typecheck`


time-attack:
Closing as superseded: the merged fix-wave A (#2820) covers this more broadly — tool-result turns are persisted per round and reconcileGatewayReplay heals every dangling turn (not just the trailing one), and the config key registration is already on master. The one piece #2820 did NOT cover — your doctor.ts checkSubagentCapability fix (explicit models.subagent before models.tier.subagent) — still applies cleanly and is tracked in TODOS.md as a P2 cherry-pick. A tiny standalone PR with just that hunk (+ a test) would be very welcome. Thanks for the thorough work here — it shaped the consolidated fix.

PR #2124, opened 2026-06-12
fix(cycle): extract_atoms stamps concepts so synthesize_concepts has material (#2123)

### LINKED ISSUE #2123: extract_atoms never stamps frontmatter.concepts — synthesize_concepts permanently skips ("no atoms with concept refs")
## Summary

`extract_atoms` never writes the `concepts` frontmatter field that `synthesize_concepts` consumes, so the atoms → concepts pipeline is broken end-to-end: every cycle reports `synthesize_concepts: skipped — no atoms with concept refs`, no matter how many atoms accumulate.

## Evidence (v0.42.37.0 / current master)

**The consumer documents that the producer should stamp the field** — `src/core/cycle/synthesize-concepts.ts` header:

> 2. Group by `concepts:` frontmatter field on each atom (when the Haiku 3-check from extract_atoms decides "this atom is about concept X", it stamps the field).

and consumes ONLY that field (`synthesize-concepts.ts:92`):

```ts
concept_refs: r.frontmatter!.concepts!,
```

with atoms filtered by `Array.isArray(r.frontmatter?.concepts)`.

**The producer never stamps it** — `src/core/cycle/extract-atoms.ts` writes `type` / `atom_type` / `source_hash` / `source_slug` / `source_quote` / `lesson` / `virality_score` / `emotional_register` / `extracted_at` / `extracted_by`, but `grep -n "concepts" src/core/cycle/extract-atoms.ts` has zero hits in the frontmatter write path (page path and transcript path both).

## Reproduction (production brain, postgres engine, pack `gbrain-everything`)

1. Run cycles until atoms accumulate — we have **696 page-derived atoms** (`extracted_by: extract_atoms-v0.41.2.1`).
2. `SELECT COUNT(*) FROM pages WHERE type='atom' AND frontmatter ? 'concepts'` → **0**.
3. Every `gbrain dream` reports `synthesize_concepts: skipped — no atoms with concept refs`.

## The consumer side works once fed

We validated the consumer by stamping `concepts` externally (1-3 kebab-case shared-vocabulary topic labels per atom, batch-labeled with one LLM call per 40 atoms): the very next `gbrain dream --phase synthesize_concepts` wrote **33 concept pages (T2=7, T3=26) from 60 stamped atoms**, $0.044, zero failures. So the fix is purely on the extractor side.

## Proposed fix (PR to follow)

Add `concepts` to the extractor's per-atom JSON contract — `"concepts": 1-3 kebab-case topic labels, reuse labels already seen in this run"` — validate kebab shape, and stamp the field in the same frontmatter write. One extra field in the existing extraction call; no new LLM round-trip. The shared-vocabulary instruction matters: labels must cluster (synthesize_concepts only materializes groups of ≥2), so the prompt should bias reuse-over-coinage.

Happy to send the PR — we're running this shape in production via an external backfill script and the tiering/synthesis downstream behaves exactly as the v0.41 design comment describes.

ChenyqThu:
Fixes #2123.

## Problem

`synthesize-concepts.ts`'s own design comment says `extract_atoms` stamps a `concepts:` frontmatter field on each atom ("when the Haiku 3-check from extract_atoms decides 'this atom is about concept X', it stamps the field") and `:92` consumes ONLY that field — but the extractor never wrote it. The atoms → concepts pipeline was therefore dead end-to-end: every cycle reports `synthesize_concepts: skipped — no atoms with concept refs`, regardless of how many atoms accumulate (696 page-derived atoms / 0 with `concepts` on our production brain before an external backfill).

## Fix (extractor side only, 2 files / +92 −3)

- **`EXTRACT_PROMPT`** asks for `concepts` (1-3 kebab-case TOPIC labels, never entity names) with an explicit reuse-over-coinage instruction — labels must cluster, since `synthesize_concepts` only materializes groups of ≥2 (T3).
- **`parseAtomsResponse`** validates labels (kebab regex, max 3, drops invalid entries; empty → `undefined` so no empty arrays get stamped).
- **The `putPage` frontmatter write** stamps `concepts` alongside `lesson` / `source_quote`. No change needed on the synthesize side.

`extracted_by` stamp left at `extract_atoms-v0.41.2.1` for the maintainer to bump with whichever release this lands in — atoms with/without labels are distinguishable via `frontmatter ? 'concepts'` either way.

## Tests

`bun test test/cycle/extract-atoms-synthesize-concepts.test.ts` → **25 pass / 0 fail / 88 expect()** (was 20). `tsc --noEmit` clean.

- 4 new parse cases (valid kept / non-kebab filtered / >3 truncated / absent→undefined).
- 1 new end-to-end regression: extractor → **real frontmatter** → `synthesize_concepts`' **own DB query path** → concept page. The existing tests fed synthesize via the `_atoms` seam, which is exactly how this gap survived CI.

## Production validation

Before this PR we validated the consumer by stamping the same shape externally on a production brain (postgres engine, pack `gbrain-everything`): the very next `gbrain dream --phase synthesize_concepts` wrote **33 concept pages (T2=7, T3=26) from 60 stamped atoms**, $0.044, zero failures — narratives are genuine cross-atom synthesis. The extractor was the only missing link.

### DIFF
--- a/src/core/cycle/extract-atoms.ts
+++ b/src/core/cycle/extract-atoms.ts
@@ -118,9 +118,19 @@ interface ExtractedAtom {
   body: string;
   source_quote?: string;
   lesson?: string;
+  /**
+   * 1-3 kebab-case topic labels for concept clustering. Consumed by
+   * synthesize_concepts (groups atoms by `frontmatter.concepts`; only
+   * labels shared by >=2 atoms materialize a concept page, so the prompt
+   * biases reuse-over-coinage). #2123.
+   */
+  concepts?: string[];
   virality_score?: number;
   emotional_register?: string;
 }
+
+/** kebab-case validator for concept labels ("captive-portal", "channel-pricing"). */
+const CONCEPT_LABEL_RE = /^[a-z0-9]+(-[a-z0-9]+)*$/;
 
 const EXTRACT_PROMPT = `You extract atomic content nuggets from a transcript.
 
@@ -132,11 +142,16 @@ quote, or short essay angle. Each atom must:
 
 Output a JSON array of atoms (1-3 per transcript, never more than 3).
 Each atom: {title (≤80 chars), atom_type, body (2-4 sentences),
-source_quote (verbatim ≤200 chars), lesson (one sentence), virality_score
-(0-100), emotional_register (one of: shocking, inspiring, funny, sobering,
-practical, controversial)}.
+source_quote (verbatim ≤200 chars), lesson (one sentence), concepts
+(1-3 topic labels), virality_score (0-100), emotional_register (one of:
+shocking, inspiring, funny, sobering, practical, controversial)}.
 
 atom_type MUST be one of: ${ATOM_TYPES.join(', ')}.
+
+concepts are kebab-case English TOPIC labels used to cluster atoms into
+concept pages (e.g. "captive-portal", "channel-pricing-strategy") — never
+entity or brand names. Use the same label for the same topic across atoms;
+prefer a label you already used over coining a near-synonym.
 
 Output ONLY the JSON array, no prose.`;
 
@@ -538,6 +553,7 @@ export async function runPhaseExtractAtoms(
                 source_hash: item.contentHash.slice(0, 16),
                 ...(atom.source_quote && { source_quote: atom.source_quote }),
                 ...(atom.lesson && { lesson: atom.lesson }),
+                ...(atom.concepts && atom.concepts.length > 0 && { concepts: atom.concepts }),
                 ...(atom.virality_score !== undefined && { virality_score: atom.virality_score }),
                 ...(atom.emotional_register && { emotional_register: atom.emotional_register }),
                 extracted_at: new Date().toISOString(),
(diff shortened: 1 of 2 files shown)


asenkovskiy:
Production validation for this fix - it's load-bearing and the last missing link now that #2163 is merged.

We hit the exact gap this PR fixes: `synthesize_concepts` consumes only the `concepts:` frontmatter field, but `extract_atoms` never wrote it, so every cycle logged *"skipped - no atoms with concept refs"* no matter how many atoms accumulated. We've run an equivalent extractor-side patch (stamp 1–3 kebab-case concept labels per atom, reuse-over-coinage) in production for ~2 weeks on a ~230-atom brain. Result: concept synthesis went from **dead to 36 concepts**, tiered T2/T3, regenerating on cadence — same shape as the PR author's 33-concepts-from-60-atoms result. Two independent reproductions, same root cause, same fix.

Worth flagging for whoever reviews: **with #2939 (#2163) now merged, this PR completes the pipeline end-to-end.** #2163 made `synthesize_concepts` write concept pages through the chunk+embed path so they're retrievable - but that fix has nothing to synthesize *from* until atoms carry concept refs, which is exactly what this PR provides. Producer (this) + consumer (#2939) together = atoms→concepts→retrieval actually works. Separately merged, each is inert without the other.

Would love to see this land - happy to share our extractor prompt/validation shape if useful for comparison.

PR #2126, opened 2026-06-12
feat(recipes): add MLX embedding + MLX chat recipes for Apple silicon

hunterbohm:
Adds first-class recipes for local inference on Apple silicon via `mlx_lm.server` / MLX OpenAI-compatible shims:

- **`mlx-embedding`** — local embedding provider. Running in production with `Qwen3-Embedding-8B-4bit-DWQ` at 4096 dimensions.
- **`mlx-gemma4`** — local chat/expansion provider for MLX-served chat models.

Both follow the existing recipe shape (`base_url_default` + env override), and `buildGatewayConfig` gains `MLX_EMBEDDING_BASE_URL` / `MLX_GEMMA4_BASE_URL` passthrough, mirroring the existing llama-server and Ollama entries.

No behavior change for anyone not setting these env vars.

**Tested:** running since April 2026 on a 24GB M-series Mac mini — embeddings, sync, and autopilot extract cycles against a ~150-page brain. `test/ai/build-gateway-config.test.ts` passes (7/7).


time-attack:
Thanks for sharing this — running it in production since April on real hardware is a genuinely useful signal. We are not adding new provider recipes: each one is permanent surface to keep current, and an MLX OpenAI-compatible server is already reachable through the existing user-provided/llama-server style recipes with a base-url override. Closing as provider surface we are not taking, with appreciation for the clean, minimal shape of the patch.

PR #2127, opened 2026-06-12
fix(budget): price non-Anthropic chat models via the canonical table

troyhoffman-oss:
## Problem

`lookupChatPrice` in `budget-tracker.ts` only consults `ANTHROPIC_PRICING`. Any non-Anthropic chat model (`openai:gpt-5.5`, `openrouter:*`, `google:*`) resolves to null pricing, and because a default cost cap is always present, `reserve()` throws `BudgetExhausted (no_pricing)` **before the provider call**. Facts extraction silently produces nothing for those providers — no error surfaces to the operator.

Hit this in production running gbrain extraction through a subscription proxy (OpenAI-compatible endpoint) — extraction was silently dead until we traced it to the budget gate.

## Fix

- `lookupChatPrice` falls through to the existing exported `canonicalLookup` from `model-pricing.ts` after the Anthropic views miss.
- Add $0 canonical entries for subscription-proxy OpenRouter endpoints (`openrouter:gpt-5.5`, `openrouter:gpt-5.4-mini`) so the TX2 hard-fail doesn't trip on per-token-free routes.

Running patched in production since 2026-06-11 (two daemons, extraction + dream drills green).

### DIFF
--- a/src/core/budget/budget-tracker.ts
+++ b/src/core/budget/budget-tracker.ts
@@ -32,6 +32,7 @@ import { mkdirSync, appendFileSync } from 'node:fs';
 import { dirname } from 'node:path';
 import { gbrainPath } from '../config.ts';
 import { ANTHROPIC_PRICING, type ModelPricing } from '../anthropic-pricing.ts';
+import { canonicalLookup } from '../model-pricing.ts';
 import { EMBEDDING_PRICING, lookupEmbeddingPrice } from '../embedding-pricing.ts';
 import { splitProviderModelId } from '../model-id.ts';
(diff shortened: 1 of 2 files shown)


time-attack:
Thanks for tracing this to the budget gate — silent extraction death is the worst failure mode and you found the real cause. Master's lookupPricing already falls through to canonicalLookup after the Anthropic views miss, so non-Anthropic chat models resolve instead of TX2 no_pricing hard-failing. Closing as already fixed; the $0 openrouter subscription-proxy entries are install-specific pricing and belong in your own config rather than the canonical table.

PR #2131, opened 2026-06-12
feat: add Upstage Solar embeddings recipe

GoBeromsu:
## Summary
- add an Upstage OpenAI-compatible recipe for Solar embeddings
- route document/query input sides to Solar passage/query model names through an Upstage compat fetch shim
- validate fixed 4096-dimensional embeddings and auth via `UPSTAGE_API_KEY`

## Tests
- `bun test test/ai/recipe-upstage.test.ts test/ai/upstage-compat-fetch.test.ts test/ai/recipes-existing-regression.test.ts test/ai/gateway.test.ts test/ai/embedQuery.test.ts`
- live smoke with a local Keychain-provided `UPSTAGE_API_KEY`: `{"doc_count":1,"doc_dim":4096,"query_dim":4096}`

## Notes
- `bun run typecheck` currently fails on an existing unrelated `test/e2e/openclaw-plugin-load-real.test.ts(281,9)` factory type mismatch.

### DIFF
--- a/src/core/ai/dims.ts
+++ b/src/core/ai/dims.ts
@@ -65,6 +65,22 @@ export function isValidZeroEntropyDim(dims: number): boolean {
   return (ZEROENTROPY_VALID_DIMS as readonly number[]).includes(dims);
 }
 
+// Upstage Solar Embeddings are a fixed 4096-dim asymmetric pair in one space:
+// passage for document/indexing and query for search/query encoding.
+const UPSTAGE_SOLAR_EMBEDDING_MODELS = new Set([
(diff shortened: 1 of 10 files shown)


time-attack:
Thanks for the thorough work — the tests and the live 4096-dim smoke check are more than most provider PRs bring. We are not taking new third-party provider integrations: each one adds permanent maintenance surface in the shared gateway, and this one needs a provider-specific compat fetch shim plus changes in dims, the postgres engine, and the vector index. Closing on that policy, not on the quality of the patch.

Issue #2097, opened 2026-06-11
prefix grant for federated access


pavelpp-topia:
**What problem does this solve?**

My usecase - a shared readonly mcp for all developers accessed via VPN. The brain would contain knowledgebase of multiple code repositories. The number of repositories may grow at a random moment or some repo might be dropped. Currently federated access has to list all sources and there does not seem to be a straightforward way to update access after the fact

**What does the solution look like?**

I propose prefix access support like `mycompany/codebase/*`, which would automatically allow devs access to all sources in `mycompany/codebase/app-a`, `mycompany/codebase/app-b` ... etc. Separating it from `mycompany/wiki`. We can still restrict pure wildcard access like `*`, although I do see some usecases in a trusted environment, where automatic access to all sources could be useful.

**Alternatives considered**

I could just fork and just implement it for us, but I think it will be useful for the whole community

pavelpp-topia:
I covered wildcard scenario in [https://github.com/garrytan/gbrain/pull/2133](PR)

time-attack:
Thanks for the thoughtful write-up and for taking a swing at it in #2133 — prefix grants like `mycompany/codebase/*` are a reasonable shape for a growing multi-repo brain. That said, this is a new feature surface rather than a defect in the current federated-grant behavior (explicit source lists work as designed), and the repo isn't taking new feature surface right now. Closing the issue; if the maintainer picks up the wildcard work, the discussion will happen on the PR.

pavelpp-topia:
@time-attack I disagree that it's a whole new feature, more like an enhancement. We've been running this version internally from a fork, the downside is that we will have to maintain this fork for as long as it's not merged in. Hopefully you will reconsider.

PR #2133, opened 2026-06-12
feat(security): '*' wildcard in federated_read for all-sources read scope

pavelpp-topia:
## Problem

Federated read grants — `oauth_clients.federated_read`, or a legacy bearer token's `permissions.source_id` array — are an **exact-match source allow-list that is frozen at client creation**. There is no command to amend a grant after the fact (`auth permissions` only edits `takes_holders`), and `sources add` never touches existing grants. The grant resolves to `source_id = ANY($array)` (`postgres-engine.ts`), pure equality with no prefix/wildcard.

Consequence: in a multi-source brain, **every newly-synced source is invisible to existing read clients** until they are revoked and re-issued with an expanded list. For a brain whose sources grow over time (e.g. a sync job onboarding repos), that's a standing operational tax and a silent-failure trap (new source syncs fine, but reads return empty with no error).

## Fix

A `'*'` entry in a federated_read grant means **"read every source, current and future."** One contained change, routed through the existing single resolver so every read op honors it without per-handler drift:

- **`sourceScopeOpts`** returns the no-filter shape `{}` when the grant holds `'*'` — checked *before* the exact-list branch, so it never degrades to a literal `source_id = ANY(['*'])` (which would match a nonexistent source named `*`).
- **`resolveRequestedScope`** lets a `'*'` client narrow to any explicit `source_id` without the out-of-grant rejection.
- **`parseLegacyTokenScope`** keeps `'*'` out of the scalar **write** floor (falls back to `'default'`).

## Safety

Built to ride *with* the source-isolation contract, not around it:

- **Read scope only.** `'*'` lives on the read axis; write authority is the FK-backed scalar source, which rejects `'*'`. A wildcard reader can never write across sources.
- **Deliberate, never implicit.** Only an explicit `'*'` widens; an empty/garbage grant stays fail-closed to `'default'`. The existing `[]`-must-not-widen guard is untouched.
- **Single chokepoint.** All read ops resolve through `resolveRequestedScope` / `sourceScopeOpts`, so there's no per-handler `__all__` re-inlining (the bug class the existing resolver was built to prevent).

## Usage

```bash
gbrain auth register-client kb-reader \
  --scopes read \
  --federated-read '*'
```

## Tests

13 new cases across `test/source-scope-resolver.test.ts`, `test/legacy-token-federated-scope.test.ts`, and `test/get-page-federated-scope.test.ts`: wildcard spans all sources; mixed-with-explicit spans all; narrows to any source; read-only write floor; code-traversal stays single-source; **empty/garbage still fail-closed**; non-wildcard arrays unchanged; end-to-end `get_page` reads any source. Verified green alongside the adjacent source-isolation suites (incl. `test/e2e/source-isolation-pglite.test.ts`). `typecheck` and all `bun run verify` checks pass.

Docs: `docs/mcp/DEPLOY.md` documents the flag, the read-only/deliberate/narrowing properties, and the "sources added later" lifecycle gap this closes.

## Notes for the maintainer

No `VERSION` / `package.json` / `CHANGELOG` changes — left to your release/fix-wave process per `docs/RELEASING.md`.

### DIFF
--- a/docs/mcp/DEPLOY.md
+++ b/docs/mcp/DEPLOY.md
@@ -135,6 +135,34 @@ queries return rows from any of the listed sources. Omit both flags for the
 v0.33-compatible super-client shape. Pre-v0.34 clients are backfilled to
 `source_id='default'` on `gbrain upgrade`.
 
+**Read-everything with `--federated-read '*'`.** For a read-only client that
+should see *every* source — current and future — pass the `'*'` wildcard
+instead of an explicit list:
+
+```bash
+gbrain auth register-client kb-reader \
+  --grant-types client_credentials \
+  --scopes read \
+  --federated-read '*'
+```
+
+This is the answer to "how do I grant access to sources added later?": an
+explicit source list is **frozen at registration** (there is no command to amend
+it — you'd revoke and re-issue), so a `'*'` grant is the only shape that auto-
+includes repos synced after the client was created. Properties:
+
+- **Read scope only.** `'*'` lives on the read axis; write authority still comes
+  from the scalar `--source` (a `'*'` write source is rejected by the FK), so a
+  wildcard reader can never write across sources.
+- **Deliberate, not a default.** Only an explicit `'*'` widens to all; an
+  empty/garbage grant stays fail-closed to `default` (the v0.42.37 isolation
+  contract is unchanged). Use it when "anyone who can reach this MCP endpoint
+  may read the whole brain" is the intended posture (e.g. a company code KB
+  behind a VPN/internal ALB).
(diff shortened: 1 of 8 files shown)


pavelpp-topia:
partially addresses #2097

time-attack:
Thank you for the careful design work here — routing through the single resolver and keeping '*' off the write axis is the right instinct if this were going to be built. We are not taking it: source isolation is a must-never-violate invariant in this codebase, and a magic-string grant meaning 'every source, current and future' converts a fail-closed allow-list into a standing blanket read. The re-issue tax on adding a source is a real cost we would rather pay than hold that grant shape. Closing as a feature surface we are not adding.

Issue #2108, opened 2026-06-11
facts:absorb fire-and-forget jobs are aborted by the 1s exit-drain — silent 100% fact loss on batch put_page ingest with non-instant chat models


felix1904:
## Summary

`put_page`-triggered `facts:absorb` jobs run fire-and-forget (`runFactsBackstop` mode `'queue'`). On process exit, the background-work drain waits only **1 second** (`drainPending` default, `src/core/facts/queue.ts:155`) before calling `abort()` on the facts queue (`src/core/background-work.ts:101-105`). Any in-flight `chat()` call slower than that window is killed and the extraction is **silently lost** — the page lands fine, but no facts are ever extracted for it.

With a slower expansion-tier chat model (`openrouter:deepseek/deepseek-v4-pro`) and proposal-sized pages (several KB of CJK text), the loss rate on a batch ingest was **100%**: 40/40 absorb attempts aborted across 30 distinct pages, and the `facts` table gained **zero rows** on two consecutive days of heavy ingest.

Version `0.42.37.0` @ `1eb430a2`.

## Observed

`ingest_log` over a ~4.5h batch-ingest window (2026-06-11 05:45–10:15 UTC), all 40 rows identical in shape:

```
pipeline_error: [chat(openrouter:deepseek/deepseek-v4-pro)] The operation was aborted.
```

- 40 rows, 30 distinct `source_ref` pages (some pages aborted 2–3×).
- Affected pages were written via MCP `put_page` during batch knowledge ingest (full proposal documents, mostly Traditional Chinese, multi-KB content each).
- `SELECT date_trunc('day', created_at)::date, count(*) FROM facts GROUP BY 1` shows **0 facts** landed on either ingest day (most recent fact row predates the batch by 5 days). Pages, embeddings, links, search: all fine. Hot-facts layer: empty for the entire batch.

## Root cause (code-path walked)

1. `put_page` → `runFactsBackstop` mode `'queue'` — fire-and-forget by design (`src/core/facts/backstop.ts` header: "Caller's await is ~zero").
2. Host process eventually exits (one-shot CLI, or MCP server shutting down at session end). The background-work drain registry runs: `drainPending({ timeout })` with the **default 1000ms** (`src/core/facts/queue.ts:155`).
[...]

Issue #2084, opened 2026-06-11
query path still hits 10s disconnect force-exit on PgBouncer txn-mode — search-arm teardown bypasses endPoolBounded (residual of #1972)


garrytan-agents:
## Residual from #1972 (Bug 2): `gbrain query` still hits the 10s disconnect force-exit on PgBouncer txn-mode — a teardown path bypasses `endPoolBounded`

**Version:** 0.42.39.0 @ `8f45624e`. Same deployment as #1972 (386K pages, Supabase split-pool: direct `:5432` + read `:6543` PgBouncer transaction-mode, federation `default` + `straylight-brain`).

The 0.42.37 fix (#2015) landed `endPoolBounded` (db.ts: 2s per-pool cap) and wired it through `ConnectionManager.disconnect()` (concurrent, non-stacking) and the db.ts module-singleton `disconnect()`. **It works** — the worst symptom from #1972 Bug 2 (force-exit racing/swallowing CLI stdout) is gone, and simple commands now exit clean:

```
gbrain status            → 6s wall, no force-exit ✅
gbrain query "x" --limit 1 → 15s wall, force-exit banner EVERY time ❌
   [cli] engine.disconnect() did not return within 10000ms — force-exiting
```

100% reproducible on the query path; 0% on status. Output is intact (results print fully before teardown — the damage half is fixed), but every query pays a flat **10s teardown tax** and exits via the hard-deadline backstop instead of cleanly.

### Where the bound is being bypassed

The teardown sequence (cli.ts `finally`, ~line 408) is:
1. `drainAllBackgroundWorkForCliExit({ timeoutMs: 1000 })` — bounded ✅
2. `await engine.disconnect()` — this is what exceeds 10s on the query path ❌

Since `ConnectionManager.disconnect()` and db.ts module `disconnect()` are both bounded (2s each, concurrent → worst case ~2.5s), whatever hangs must be a connection/handle the query path opens that doesn't drain through either:

Candidates, in likely order:
- **Federation / brain-registry path:** `BrainRegistry.disconnectAll()` (brain-registry.ts:387) does `Promise.allSettled(handles.map(h => h.engine.disconnect()))` — bounded *per engine*, but a multi-source query (`default` + `straylight-brain` here) initializes multiple engines; if any engine's disconnect resolves the bound but leaves a socket open, Bun's event loop stays alive and only the unref'd force-exit timer ends the process.
[...]

PR #2135, opened 2026-06-12
fix(cli): GBRAIN_EXIT_DRAIN_MS exit-drain knob + arm disconnect watchdog after drain (#2108)

claudiobedino:
## Summary

Closes the silent fact-loss class from #2108 (ask 2): `facts:absorb` fire-and-forget jobs are aborted by the 1-2s exit-drain whenever the in-flight LLM round-trip is slower than the window. Reproduced on the **`gbrain sync` CLI path** as well (not only MCP `put_page`): 17/17 absorbs aborted on a single batch sync, `ingest_log` filling with `gateway_error: [chat(...)] The operation was aborted.`

Two changes, both required:

1. **`src/core/background-work.ts` — `GBRAIN_EXIT_DRAIN_MS` env knob** for the exit-drain timeout. Env wins over call-site values so one knob covers both exit paths (op-dispatch passes `1000` explicitly; the CLI_ONLY teardown passes nothing). `drainPending` settles as soon as in-flight work finishes, so idle runs still exit instantly regardless of the value — the knob only buys patience when there IS work to save.
2. **`src/cli.ts` (CLI_ONLY `finally`) — arm the 10s disconnect watchdog *after* the drain**, not before. Armed early, it force-exits any drain longer than 10s with the misleading `engine.disconnect() did not return within 10000ms` warning — silently defeating any longer drain window (same early-arming class as #2105). Armed after the drain, it bounds only the disconnect, matching its message.

## Validation

- Production Postgres brain, `0.42.40.0`, `facts.extraction_model=anthropic:claude-haiku-4-5`: a 195-file sync (469 chunks) with `GBRAIN_EXIT_DRAIN_MS=600000` completed with **0 aborted absorb rows**, vs **17/17 aborted** immediately before the patch on the same brain.
- Two new unit tests pin the contract: env-wins-over-call-site, and invalid env (`abc`/`0`/`-5`/empty) falls back to call-site/default.
- `tsc --noEmit` clean; drain-related suites (`test/core/background-work.test.ts`, `test/eval-capture-drain.test.ts`, `test/fix-wave-structural.test.ts`, `test/cli-should-force-exit.test.ts`) pass 36/36.

## Notes

- No VERSION/CHANGELOG bump included — left to the maintainer's release wave per the repo's versioning workflow.
- Refs #2108, #2105. Context and field validation also posted at https://github.com/garrytan/gbrain/issues/2108#issuecomment-4689648821

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -1926,6 +1926,12 @@ async function handleCliOnly(command: string, args: string[]) {
     // #1471: this is also the fall-through OWNER-disconnect — the owner is torn
     // down LAST (after the drain), so module-singleton borrowers never outlive it.
     if (command !== 'serve') {
+      // Drain FIRST, then arm the disconnect watchdog: the timer's contract is
+      // bounding engine.disconnect() (see its message), but armed before the
+      // drain it also killed any drain longer than 10s — which defeats a
+      // GBRAIN_EXIT_DRAIN_MS-extended drain waiting on in-flight facts:absorb
+      // (#2108) and force-exited mid-settle with a misleading warning.
+      await drainAllBackgroundWorkForCliExit();
       const forceExit = shouldForceExitAfterMain();
       let hardExitTimer: ReturnType<typeof setTimeout> | undefined;
       if (forceExit) {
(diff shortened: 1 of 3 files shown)


time-attack:
Thank you for the detailed field validation — the 17/17 aborted absorbs is exactly the right kind of evidence. Both changes have since landed on master through #2084: the per-sink drain budget is env-overridable via GBRAIN_DRAIN_TIMEOUT_MS in src/core/cli-force-exit.ts, and the teardown backstop is now armed after the handler at teardown start with a deadline computed from the bounds it guards, so a longer drain no longer trips the misleading disconnect warning. Closing as already fixed; the cli.ts hunk no longer applies since that finally block is now finishCliTeardown().

Issue #1433, opened 2026-05-25
`gbrain sync --skip-failed` drops bare-markdown log files from the index


infiniteGameExp:
_Filed via brain-janitor from `Learning and Strategy/Wiki/reference/gbrain-known-bugs.md` (bug #6)._

- **Filed**: 2026-05-24 (pending upstream filing — not yet posted)
- **Symptom**: After running `gbrain sync --skip-failed`, a domain `log.md` that was previously indexed is silently removed from the gbrain page store. The sync output includes a single line `Deleted un-syncable page: <slug>` and proceeds. The on-disk file is untouched (all entries intact), but `mcp__gbrain__get_page` for that slug returns `page_not_found` and the page no longer surfaces in retrieval. Distinct from bug #7 (createVersion) — that fails on edit, this one *deletes* on `--skip-failed`.
- **Root cause** (suspected): the sync's `--skip-failed` parse-failure handler treats "no YAML frontmatter" as un-syncable rather than as bare-markdown-valid. Domain `log.md` files are valid per the vault schema (`Append-only chronological record. Format: ## [YYYY-MM-DD] <op> | <subject>`) and intentionally have no frontmatter — they are operational logs, not typed pages. The sync collapses two different conditions (genuinely malformed vs. valid-but-untyped) into the same deletion path.
- **India's encounter**: Ran `gbrain sync --skip-failed` on 2026-05-24 after the [[phase-6-architecture-blind-spots-2026-05-23]] decision-log landed. Sync output: `Deleted un-syncable page: learning-and-strategy/log`. The file at `Learning and Strategy/log.md` (3755 bytes, 37 lines, two recent appends from the same session) was unchanged on disk but `mcp__gbrain__get_page` returned 404 for `learning-and-strategy/log` immediately after.
- **Workaround**: None applied yet. Two options India can pick from:
  1. Add a minimal `---\ntype: log\nid: <slug>\n---\n` frontmatter to every domain `log.md` so gbrain parses them as typed pages. Requires extending the schema with a `log` page type (or repurposing `reference-list`).
  2. Pin the affected log files in an `~/.gbrain/sync-exclude` (if such a mechanism exists in this version) so sync ignores them entirely, preserving the bare-markdown convention.
[...]

PR #2157, opened 2026-06-12
feat(sources): persist + honor --include / --exclude globs across sync and lint (#2156)

### LINKED ISSUE #2156: sources: add --include / --exclude glob filters honored by sync AND lint
**What happened?**

`sources.config.{include_globs,exclude_globs}` had storage but no readers. `gbrain sources add` had no flag to populate them, and the two surfaces that walk a source directory (`gbrain sync` and `gbrain lint`) both ignored the field. The internal sync classifier in `src/core/sync.ts` (`SyncableOptions`, `unsyncableReason`, the `'include-glob-miss'` / `'exclude-glob-hit'` reason codes shipped in v0.41.13 / #1433) supported include + exclude globs from day one; the CLI surface did not. So an Obsidian-style vault with authoring scaffolding under `Templates/`, third-party documentation dumps under `Resources/api-specs/`, or older content under `Archive/` saw every page chunked, embedded, surfaced in search results, AND emitted as lint findings the user had no way to opt out of.

The workarounds all break the host tool: renaming `Templates/` confuses Obsidian's "create note from template" flow, hiding it with a dot-prefix breaks the template picker, and `.gitignore` is the wrong place (the file is meant to be in the vault, not the index).

On a mixed-content vault, a no-filter `gbrain lint --dry-run` reports an order of magnitude more findings than the user actually wants to triage; the gap is entirely from subtrees the user already considers out of scope (vendor API specs, historical scratch, scaffolding templates).

**What did you expect?**

`gbrain sources add vault --path ~/vault --exclude 'Templates/**' --exclude '.smart-env/**'` should persist those globs so every subsequent `gbrain sync` AND `gbrain lint <vault>` honors them. Same for the symmetric `--include 'people/**'` case (CRM vaults where only a subtree is brain material). The cycle's lint phase, which calls into the same path, should pick up the same scope without per-invocation flag plumbing.

**Steps to reproduce**

1. `gbrain init` against a fresh engine.
2. `gbrain sources add vault --path /some/obsidian-vault --federated`.
3. Add an `obsidian-vault/Templates/Meeting.md` file.
4. `gbrain sync --source vault`.
5. `gbrain search "meeting template"`: the template's pages appear in results despite being scaffolding.
6. `gbrain lint --dry-run /some/obsidian-vault`: the template's pages appear in the issue list.

There is no CLI surface today to opt out of indexing or linting `Templates/**` short of renaming the directory or carrying a fork patch.

**Environment**

- gbrain version: 0.42.42.0 (reproduces against current `master` at 4ee530f3).
- OS: Linux (Ubuntu) and macOS both reproduce.
- Bun version: 1.3.14.
- Database: self-hosted Postgres 16, also reproduces on PGLite.

**Proposed fix**

Two-half change with no schema migration:

1. **Write side (`gbrain sources add`).** New repeatable `--include <glob>` / `--exclude <glob>` flags. Values land in `sources.config.include_globs` / `sources.config.exclude_globs` arrays at INSERT time.

2. **Read side, surface A (`gbrain sync`).** `commands/sync.ts` reads those arrays at the three sync entry points (`--all` fan-out's `runOne`, single-source CLI path, `syncOneSource`) and threads them into `SyncOpts.include` / `SyncOpts.exclude`. `performSync`'s filter at `commands/sync.ts:1454` constructs `syncOpts` with all three knobs and passes them into `isSyncable` / `unsyncableReason`.

3. **Read side, surface B (`gbrain lint`).** `runLintCore` lifts the same `sources.config.{include,exclude}_globs` arrays for the matching source row, threads them through `collectPages`. New `--include <glob>` / `--exclude <glob>` flags (repeatable, parallel to the sync surface) for ad-hoc scoping. The cycle's lint phase already passes `opts.engine`, so it gets auto-resolution without further wiring. Explicit caller-supplied flags always win over the source-row lift.

4. **Shared matcher.** Export `matchesAnyGlob` from `src/core/sync.ts` so lint reuses the exact same `globToRegex` semantics sync ships, rather than re-declaring the matcher. Keeps the two surfaces in lockstep.

Adjusting globs on an existing source is deliberately out of scope. The existing `gbrain sources set-cr-mode` is the precedent for a future narrow-verb `sources set-globs` surface; `UPDATE sources SET config = config || '...'::jsonb` is the documented escape hatch in the meantime.

PR ready for review: #2157.

brettdavies:
Closes #2156.

## Summary

**The bug.** `sources.config.{include_globs,exclude_globs}` had no readers. `gbrain sources add` had no flag to populate them; `gbrain sync` and `gbrain lint` both walked the source dir blind. The classifier at `src/core/sync.ts` has carried the `include` / `exclude` contract on `SyncableOptions` since v0.41.13 (#1433). `unsyncableReason` returns `'include-glob-miss'` / `'exclude-glob-hit'`, and `test/sync-isSyncable-shape.test.ts` already pins those reason codes. But the runtime never populated either field on either surface. An Obsidian-style vault with authoring scaffolding under `Templates/`, third-party documentation dumps under `Resources/api-specs/`, or scratch content under `Archive/` saw every page chunked, embedded, surfaced in search results, AND emitted as lint findings the user had no way to opt out of. The workarounds break the host tool: renaming `Templates/` confuses Obsidian's "create note from template" flow, hiding it with a dot-prefix breaks the template picker.

On a mixed-content vault, no-filter `gbrain lint --dry-run` reports an order of magnitude more findings than the user wants to triage; the gap is entirely from subtrees the user already considers out of scope.

**The fix.** Four-surface change. The first three surfaces wire up the storage + reader pair on sources, sync, and lint. The fourth (added in the follow-on commit) closes a contract gap the first three exposed: changing `sources.config.exclude_globs` mid-life had no observable effect on the next sync because the checkpoint gate ignored config drift. One schema migration (v117).

1. `gbrain sources add` gains repeatable `--include <glob>` / `--exclude <glob>` flags. Values land in `sources.config.include_globs` / `sources.config.exclude_globs` arrays at INSERT time. Both INSERT paths (`--url` Path A and `--path` Path B) write the arrays when non-empty; omitted flags write nothing, so a brain that doesn't opt in keeps the same `sources.config` shape as today. `runAdd` refuses empty / flag-like argument values (e.g. `--include --federated` exits 2 with a clear message instead of silently swallowing the next flag).

2. `commands/sync.ts` reads those arrays at every sync entry point (the `--all` fan-out's `runOne` closure, the single-source CLI path, `syncOneSource`) and threads them into `SyncOpts.include` / `SyncOpts.exclude`. `performSync`'s filter at line ~1454 now constructs `syncOpts` with all three knobs (`strategy`, `include`, `exclude`) and passes them into `isSyncable` / `unsyncableReason` on every walked path.

3. `gbrain lint` gains the same `--include <glob>` / `--exclude <glob>` flags. `runLintCore` lifts `sources.config.{include,exclude}_globs` from the row whose `local_path` resolves to the same absolute dir as `target`. The cycle.lint phase + Minion lint handlers already pass `opts.engine`; they get auto-resolution without further wiring. Explicit caller-supplied `include` / `exclude` always win over the source-row lift. The CLI's summary-line `runLintCore` re-invocation passes the same flags through so the summary scope matches the per-page detail walk.

4. `matchesAnyGlob` is exported from `src/core/sync.ts` so lint reuses the exact `globToRegex` semantics sync ships, rather than re-declaring the matcher. Keeps the two surfaces in lockstep.

5. The CLI auto-resolve engine for the standalone `gbrain lint <dir>` path now uses `connectWithRetry` (same pattern `connectEngine` in `cli.ts` uses for every other CLI invocation) with `noRetry: true`. The previous `engine.connect({})` shape dropped the URL silently; auto-resolve degraded to no-filter even when a brain WAS reachable. `noRetry: true` keeps standalone lint snappy when the brain happens to be unreachable.

**Defensive parse helper** (`parseGlobList`): the `sources.config` JSONB column is free-form. The helper coerces to `string[] | undefined`, dropping non-strings + empty strings and collapsing empty arrays to `undefined`. A literal `[]` must not engage the filter loop or every path classifies as `include-glob-miss`.

**Follow-on: checkpoint invalidation on `sources.config` drift.**

The "Already up to date" branch at `performSync`'s git-HEAD equality check honored `chunker_version` match but ignored `sources.config` drift. A user who ran `gbrain sources add default --exclude 'Templates/**'` AFTER an initial sync got "Already up to date" on the next pass because git HEAD was unchanged. The new exclusion never reached the walk until `gbrain sync --full`. The first three surfaces would have shipped a knob whose contract was "set it and the next sync respects it"; in practice the next sync no-op'd until `--full`.

Migration v117 adds `sources.config_fingerprint TEXT`, a SHA-256 hash of the walk-affecting fields in `sources.config`: `strategy`, `include_globs` (canonicalized via `parseGlobList` then sorted), `exclude_globs` (same). The gate compares stored vs current; mismatches force a full re-walk via the same code path as a `chunker_version` bump and stamp the new fingerprint on success. Pattern mirrors the existing `chunker_version` gate exactly so both kinds of drift route through one recovery path.

Fields NOT in the fingerprint:
- `federated`: affects search visibility, not the walk set.
- CLI-supplied `--include` / `--exclude` overrides: one-off scope changes, not source state. A user running `gbrain sync --exclude X` on a row whose stored config has no X is intentionally narrowing this one pass; the row's checkpoint stays valid for the next no-flags sync.

NULL on pre-v117 rows is treated as "never stamped", so the first post-upgrade sync stamps quietly without spuriously force-fulling. Array order is normalized (alphabetical, post-`parseGlobList`), so re-ordering an exclude list doesn't trigger a spurious re-walk. Add-then-revert returns to the original fingerprint (toggle is a no-op).

`sources.config_fingerprint` joins `sources.chunker_version` in `COLUMN_EXEMPTIONS` in `test/schema-bootstrap-coverage.test.ts`: TEXT column on `sources`, not referenced by any CREATE INDEX, idempotent `ALTER ... IF NOT EXISTS` in the migration covers every upgrade path (fresh install lands via CREATE TABLE in `src/schema.sql` + `src/core/schema-embedded.ts`; pre-existing brains get the ALTER).

## Tests

- `bun test test/sources.test.ts`: 21/21 pass (existing 17 + new add-CLI cases for single `--exclude`, single `--include`, repeatable `--exclude`, composed `--include` + `--exclude`, omitted-flags regression guard, two arg-required negative cases).
- `bun test test/sync-source-globs.test.ts`: 22/22 pass (7 existing `parseGlobList` cases + 15 new unit cases on `computeSourceConfigFingerprint`: determinism, order-insensitivity, per-field sensitivity, toggle-and-revert, defensive coercion, non-walk-affecting fields ignored, SHA-256 output shape).
- `bun test test/sync-config-fingerprint.test.ts`: 14/14 pass; PGLite-backed end-to-end coverage of migration v117 column shape, read/write/overwrite round-trips, NULL semantics for legacy rows, drift simulation across all three walk-affecting fields, federated-toggle non-invalidation, double-encoded JSONB round-trip parity (the `gbrain sources add` `JSON.stringify(config)::jsonb` shape from #2156 still hashes to the same value after parse).
- `bun test test/sync-isSyncable-shape.test.ts`: 12/12 pass; the existing `include` / `exclude` contract is untouched.
- `bun test test/lint-source-globs.test.ts`: 8/8 pass; walker contract pinned across no-filter regression guard, include / exclude / both, multi-pattern union, exclude-after-include precedence, empty-array guard (symmetric with `parseGlobList`), single-file bypass, and the `**` cross-segment semantics sync uses.
- `bun test test/schema-bootstrap-coverage.test.ts`: 9/9 pass; the new column joins the existing `sources.chunker_version` exemption with the same rationale.
- Adjacent suites also pass: `test/sync.test.ts`, `test/sync-strategy.test.ts`, `test/sync-metafile-skip.serial.test.ts`, `test/sync-classifier-widening.test.ts`, every other `lint-*.test.ts`.
- `bun run typecheck` clean. `bun run verify` clean (30 parallel checks).

Smoke (end-to-end against a synthetic vault layout):

```text
gbrain sources add vault --path ~/vault --exclude 'Templates/**' --exclude 'Resources/api-specs/**'
gbrain sync --source vault       # writes only the in-scope subtree
gbrain lint /home/user/vault     # standalone CLI lifts the same persisted globs
```

Drift smoke (the follow-on commit's contract):

```text
gbrain sources add vault --path ~/vault                     # initial sync, no excludes
gbrain sync --source vault                                  # walks every file under ~/vault
# operator updates the row via direct JSONB (or a future `sources set-globs`):
psql "$DATABASE_URL" -c "UPDATE sources SET config = jsonb_set(config, '{exclude_globs}', '[\"Templates/**\"]') WHERE id = 'vault'"
gbrain sync --source vault                                  # logs `[sync] full re-walk forced (config_fingerprint=...→...)` and re-walks
```

Once the globs persist on the source, the cycle.lint phase + every nightly cron pick up the same scope automatically. No per-invocation flag plumbing, and no manual `--full` after a config change.

## Notes

- `gbrain sources set-globs` for modifying globs on an existing source is out of scope here. The precedent in this repo is the narrow-verb `sources set-cr-mode`; worth filing as a follow-up once 3+ writable per-source fields exist. Until then, the documented escape hatch is a direct JSONB update (see `docs/guides/multi-source-brains.md`). Whenever it lands, the fingerprint gate already handles whatever shape it writes.
- `.gbrainignore` at the source root is also out of scope. The JSONB column is the right single source of truth for the first cut; a `.gbrainignore` parser layers cleanly on top once the persistence is settled. The fingerprint would need to extend to cover `.gbrainignore` content as well if that ever lands; structurally the same problem.
- The `JSON.stringify(config)::jsonb` shape `gbrain sources add` writes (a JSON-string scalar rather than a proper JSONB object) is tolerated end-to-end. The defensive reader in `postgres-engine.ts:1274` parses the string back, and `readSourceConfig` in the fingerprint path applies the same coercion. Fixing the writer to use `engine.executeRawJsonb` (per the cross-cutting CLAUDE.md invariant) is a separate cleanup that does not change observable behavior.

### DIFF
--- a/docs/guides/multi-source-brains.md
+++ b/docs/guides/multi-source-brains.md
@@ -1,18 +1,15 @@
 # Multi-source brains
 
-**A single gbrain database can hold multiple knowledge repos.** Each one
-is a `source`: a logical brain-within-the-brain with its own slug
-namespace, its own sync state, and its own federation policy. The rest
-of this guide walks the three canonical scenarios.
+**A single gbrain database can hold multiple knowledge repos.** Each one is a `source`: a logical brain-within-the-brain
+with its own slug namespace, its own sync state, and its own federation policy. The rest of this guide walks the three
+canonical scenarios.
 
 ## The three scenarios
 
 ### 1. Unified knowledge recall (wiki + gstack)
 
-You have a personal wiki and a `gstack` checkout. Both belong to you,
-both are knowledge you want your agent to recall across. When you ask
-"what did I learn about X?" you want the best hit whether it lives in
-the wiki or in a gstack plan.
+You have a personal wiki and a `gstack` checkout. Both belong to you, both are knowledge you want your agent to recall
+across. When you ask "what did I learn about X?" you want the best hit whether it lives in the wiki or in a gstack plan.
 
 ```bash
 # Register the gstack source, federate so it joins cross-source search
@@ -28,15 +25,14 @@ gbrain sync --source gstack
 # gstack. Each result includes source_id so the agent can cite properly.
 ```
 
-Result: wiki pages and gstack plans are separate (different source_ids,
-different slug namespaces) but share the search surface.
+Result: wiki pages and gstack plans are separate (different source_ids, different slug namespaces) but share the search
+surface.
 
 ### 2. Purpose-separated brains (yc-media + garrys-list)
 
-You run two completely different content pipelines on the same backend.
-YC Media covers portfolio news and founder profiles. Garry's List is
-personal writing. You explicitly DON'T want them mixed in search — YC
-portfolio content leaking into essay searches is a bug, not a feature.
+You run two completely different content pipelines on the same backend. YC Media covers portfolio news and founder
+profiles. Garry's List is personal writing. You explicitly DON'T want them mixed in search — YC portfolio content
+leaking into essay searches is a bug, not a feature.
 
 ```bash
 # Two sources, both isolated (federated=false)
@@ -52,10 +48,9 @@ gbrain sync --source yc-media
 gbrain sync --source garrys-list
 ```
 
-Result: searching from neither directory returns the `default` source
-(your main brain). Searching from inside `~/yc-media` returns only yc-
-media hits. Searching from inside `~/writing` returns only garrys-list.
-Federation is opt-in, not leaked.
+Result: searching from neither directory returns the `default` source (your main brain). Searching from inside
+`~/yc-media` returns only yc- media hits. Searching from inside `~/writing` returns only garrys-list. Federation is
+opt-in, not leaked.
 
 To search across them explicitly on demand:
 
@@ -65,9 +60,8 @@ gbrain search "tech layoffs" --source yc-media,garrys-list
 
 ### 3. Mixed (wiki federated + sessions isolated)
 
-Your main wiki is federated with a few trusted sources. Your session
-transcripts (coming in v0.18) land in a separate isolated source so
-they don't dominate every search result.
+Your main wiki is federated with a few trusted sources. Your session transcripts (coming in v0.18) land in a separate
+isolated source so they don't dominate every search result.
 
 ```bash
 # Federated sources
@@ -79,33 +73,30 @@ gbrain sources add sessions --path ~/.claude/sessions --no-federated
 
 ## Resolution priority
 
-When any command needs to pick a source, gbrain walks this list (highest
-first):
+When any command needs to pick a source, gbrain walks this list (highest first):
 
 1. Explicit `--source <id>` flag.
 2. `GBRAIN_SOURCE` environment variable.
 3. `.gbrain-source` dotfile in CWD or any ancestor directory.
-4. A registered source whose `local_path` contains the CWD (longest
-   prefix wins for nested checkouts).
+4. A registered source whose `local_path` contains the CWD (longest prefix wins for nested checkouts).
 5. The brain-level default set via `gbrain sources default <id>`.
 6. The seeded `default` source.
 
-So inside `~/.gstack/plans/` on a brain that pinned `gstack` to
-`~/.gstack` via `.gbrain-source`, `gbrain put-page` implicitly writes to
-the `gstack` source. Outside any registered directory with no env/dotfile
-set, it writes to the default.
+So inside `~/.gstack/plans/` on a brain that pinned `gstack` to `~/.gstack` via `.gbrain-source`, `gbrain put-page`
+implicitly writes to the `gstack` source. Outside any registered directory with no env/dotfile set, it writes to the
+default.
 
 ## Federation flag
 
 Every source row stores `config.federated: boolean` in its JSONB config.
 
-| Value | Meaning |
-|-------|---------|
-| `true` | Source participates in unqualified `gbrain search "X"` results. |
+| Value                             | Meaning                                                                               |
+| --------------------------------- | ------------------------------------------------------------------------------------- |
+| `true`                            | Source participates in unqualified `gbrain search "X"` results.                       |
 | `false` (default for new sources) | Source only searched when explicitly named via `--source <id>` or qualified citation. |
 
-The seeded `default` source is `federated=true` so pre-v0.17 brains
-behave exactly as before — every page appears in search.
+The seeded `default` source is `federated=true` so pre-v0.17 brains behave exactly as before — every page appears in
+search.
 
 Flip later with `gbrain sources federate <id>` / `unfederate <id>`.
 
@@ -113,8 +104,9 @@ Flip later with `gbrain sources federate <id>` / `unfederate <id>`.
 
 Full subcommand reference:
 
-```
+```text
 gbrain sources add <id> --path <p> [--name <n>] [--federated|--no-federated]
+                                  [--include <glob>...] [--exclude <glob>...]
                                Register a source. id: [a-z0-9](?:[a-z0-9-]{0,30}[a-z0-9])?
 gbrain sources list [--json]   List all sources with page counts + federation state.
 gbrain sources remove <id> [--yes] [--dry-run] [--keep-storage]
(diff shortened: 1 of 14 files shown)


time-attack:
Closing this one, and it isn't a knock on the work — the write-up and the tests are genuinely strong. It's a ~1300-line feature that adds new user-facing flags to two commands plus a schema migration, which is more surface than we're taking right now, and by your own note the edit path for existing sources is left as a raw `UPDATE ... jsonb` escape hatch, so the flag story lands incomplete. The dangling `include`/`exclude` fields on SyncableOptions are a fair thing to point at; we'd rather resolve that by narrowing the contract than by growing two CLI surfaces. Thank you for the thorough investigation.

PR #2159, opened 2026-06-13
fix(sync): let tokenmax cost gates proceed

### LINKED ISSUE #2139: feat: cost gates should respect tokenmax posture — blocking ConfirmationRequired wedges non-interactive syncs
## Problem

The sync inline-embed cost gate (`sync.cost_gate_min_usd`, default $0.50) fires on every serial recovery sync of a large active brain, and there is no first-class way to say "this operator has decided cost is not the constraint — never gate."

On a 193K-page production brain in **tokenmax** posture (operator explicitly optimizing for maximum knowledge quality, not spend):

- `--skip-failed` is incompatible with `--parallel` (v0.42.x: failures log is brain-global, parallel acks race), so any sync that needs failure acknowledgment must drop to `--serial`
- `--serial` embeds **inline**, which arms the blocking `ConfirmationRequired` gate
- Routine daily syncs estimate ~$8 (155-160M new tokens), far above the $0.50 floor → **every cron sync blocks on a human confirmation**, and the brain silently stops embedding new content until someone notices

The operator's instruction was literally: "disable cost gates when token maxing mode." Today that takes archaeology: discovering `sync.cost_gate_min_usd` + `embed.backfill_max_usd_per_source_24h` (both `--force`-gated, undocumented in `config show`), and setting both to absurd sentinel values (100000) because `0`/`off` is rejected.

## Proposal

1. **Mode-aware gating (the right fix).** The brain already has a tokenmax search/quality mode concept. When the operator has expressed tokenmax intent — e.g. `gbrain config set spend_posture tokenmax` — *all* cost gates (sync inline gate, embed backfill 24h cap, enrich/onboard `--max-usd` prompts) should default to **informational FYI, never blocking**. Print the estimate, proceed, log to the spend ledger.

2. **First-class off switch.** Accept `0` / `off` / `unlimited` on `sync.cost_gate_min_usd` and `embed.backfill_max_usd_per_source_24h` instead of silently falling back to defaults — sentinel-large-number configuration is a smell.

3. **Non-interactive contexts should defer, not block.** When stdin is not a TTY (cron/agent), a blocking ConfirmationRequired gate guarantees a wedged pipeline. The deferred-backfill path (federated_v2) already has the right shape — make non-TTY serial syncs auto-defer embeds to the capped backfill job instead of exiting 2.

4. **Document the spend-control surface** in one place: every gate, its key, default, and how to widen/disable.

## Related
- The `--skip-failed` x `--parallel` incompatibility is what forces serial (and thus inline-embed gating) in the first place — fixing that (per-source failure acks?) would also defuse this.

## Environment
- gbrain 0.42.39.0, Postgres engine, 193K pages, 4 sources
- Daily `sync --all` cron; gate fired 4 consecutive days; embeds silently stalled each time until manually approved

maxpetrusenkoagent:
## Summary
- Treat `search.mode=tokenmax` as a max-quality posture for `gbrain sync --all --serial` cost previews.
- Keep the inline cost estimate visible as a structured `tokenmax_notice`, but do not return `ConfirmationRequired` / exit 2 for unattended tokenmax syncs.
- Add a regression test proving an above-floor tokenmax inline sync emits the notice and does not wedge non-TTY callers.

Fixes #2139.

## Verification
- `bun test test/sync-cost-gate.serial.test.ts`
- `bun run typecheck`
- `bun run build`
- `bun run verify`
- `git diff --check`
- Codex read-only autoreview against the final diff: no blockers, verdict ready for draft PR.

## Notes
- Full `bun run test` is running locally in the persistent worktree; focused regression and repo verify gates are green.


time-attack:
Closing during backlog triage: the PR diff is empty — the merge commit appears to have discarded the fix commit's changes. Please re-open with the actual changes if still relevant.

PR #2168, opened 2026-06-13
fix(ai): pass `dimensions` for Google embeddings on openai-compat

Hippityy:
## What

Adds a single new branch in `src/core/ai/dims.ts` (openai-compatible case) so `dimsProviderOptions()` emits `dimensions: N` for Google embedding models on the openai-compat adapter. Mirrors the existing pattern for DashScope `text-embedding-v3`, Zhipu `embedding-3`, and OpenAI `text-embedding-3-*`.

## Why

Google's `gemini-embedding-2` (and legacy `-001` / `text-embedding-004`) supports Matryoshka-style output dim truncation via the `dimensions` request field — range 128..3072, recommended 768 / 1536 / 3072. Without this branch, `dimsProviderOptions()` returned `undefined` for Google models, so the gateway sent no `dimensions` field, the API returned its native 3072, and the dim-mismatch validator at `gateway.ts:1981` hard-failed on first embed for any brain configured at a non-default dim.

Reproducer (pre-fix, with brain at `embedding_dimensions: 1536`):

```bash
gbrain config set embedding_model openrouter:google/gemini-embedding-2
gbrain embed --stale
# → gateway error: vector dim mismatch (expected 1536, got 3072)
```

After this patch, gbrain sends `{"dimensions": 1536}` to the openai-compat `/v1/embeddings` endpoint, OpenRouter forwards to Google, Google returns 1536-dim vectors, gateway validation passes.

## Diff

```diff
+      // Google gemini-embedding-2 (and -embedding-001) on openai-compat
+      // proxies (OpenRouter etc.) accept `dimensions` for Matryoshka-style
+      // output truncation. Per Google's docs the supported range is
+      // 128..3072 with recommended values 768 / 1536 / 3072. ...
+      if (
+        modelId.startsWith('google/gemini-embedding') ||
+        modelId === 'text-embedding-004'
+      ) {
+        return { openaiCompatible: { dimensions: dims } };
+      }
```

Plus a new test file `test/ai/dims-google.test.ts` (7 cases).

## Symmetry note

Google embeddings are symmetric — `input_type` is intentionally NOT emitted (would be rejected provider-side). The test pins this: `inputType='query'` and `inputType='document'` produce identical output blobs.

## Negative pin

The branch uses `startsWith('google/gemini-embedding')` — does NOT match `google/gemini-3-flash-preview` (a chat model). If dimsProviderOptions started matching chat model IDs, `embedMany` would send `dimensions` to a chat endpoint and break. The last test case in `dims-google.test.ts` pins this.

## Verification

- `bun run verify` → **30/30 green**
- 143 targeted tests pass across `dims-openai / dims-zeroentropy / recipe-zhipu / recipe-dashscope / recipe-minimax / recipe-azure-openai / embedQuery / gateway / recipes-existing-regression / asymmetric-encoding-contract / gateway-embed-model-override` (0 regressions)
- 7/7 new tests pass

## Related

- Companion to #2164 (reranker touchpoint — already open) which made the recipe's existing openrouter model ID `google/gemini-embedding-2` actually usable end-to-end.
- Doesn't touch any recipe file — purely a gateway helper change. `dims_options: [512, 768, 1024, 1536]` in the openrouter recipe still gates init (1536 is in-range; users who want 3072 need a separate recipe edit).

### DIFF
--- a/src/core/ai/dims.ts
+++ b/src/core/ai/dims.ts
@@ -220,6 +220,19 @@ export function dimsProviderOptions(
       if (modelId === 'text-embedding-v3' || modelId === 'embedding-3') {
         return { openaiCompatible: { dimensions: dims } };
       }
+      // Google gemini-embedding-2 (and -embedding-001) on openai-compat
+      // proxies (OpenRouter etc.) accept `dimensions` for Matryoshka-style
+      // output truncation. Per Google's docs the supported range is
+      // 128..3072 with recommended values 768 / 1536 / 3072. Without this,
+      // a user who picks e.g. 1536 silently gets back 3072-dim vectors and
+      // the gateway dim-mismatch validator at gateway.ts:1981 hard-fails
+      // on first embed. Symmetric retrieval — inputType ignored.
+      if (
+        modelId.startsWith('google/gemini-embedding') ||
+        modelId === 'text-embedding-004'
+      ) {
+        return { openaiCompatible: { dimensions: dims } };
+      }
       // MiniMax embo-01 takes a `type: 'db' | 'query'` field for asymmetric
       // retrieval. Today still hardcoded to 'db' for back-compat — opting
       // into the new inputType seam is a follow-up (see plan's deferred
--- a/test/ai/dims-google.test.ts
+++ b/test/ai/dims-google.test.ts
@@ -1,0 +1,99 @@
+/**
+ * Google embedding Matryoshka dim plumbing for openai-compat recipes.
+ *
+ * Pins:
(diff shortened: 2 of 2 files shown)


time-attack:
Merging. I confirmed on current master that the openai-compat branch of `dimsProviderOptions()` still returns `undefined` for Google embedding models, so brains configured at a non-default dim hard-fail on first embed against a model the OpenRouter recipe already advertises. The fix mirrors the existing DashScope/Zhipu/OpenAI branches exactly and the negative test pinning that chat model IDs stay unmatched is the right guard. Thanks for the tight diff and the reproducer.

PR #2181, opened 2026-06-14
feat: support flexible embedding providers

huttey:
Summary:
- add embedding provider dimension normalization and chunk cleaning support
- update operations/search paths and tests for provider-flexible embeddings
- document install guidance for the provider behavior

Validation:
- bun run typecheck
- bun test test/embedding-provider.test.ts --timeout=60000

### DIFF
--- a/skills/install/SKILL.md
+++ b/skills/install/SKILL.md
@@ -1,3 +1,18 @@
+---
+name: install
+description: Deprecated alias for setup when the user asks to install GBrain
+triggers:
+	- "install gbrain"
+	- "gbrain install"
+	- "install the brain"
+tools:
+	- get_stats
+	- get_health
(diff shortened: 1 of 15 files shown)


time-attack:
Closing. This bundles several unrelated changes under one title — a new env-gated embed-text cleaner, frontmatter added to a deprecated alias skill, a CRLF parser tweak — and it's built against a test file that isn't on master. The part that worries me most is the unconditional `setConfig('embedding_model'/'embedding_dimensions')` at the top of `runEmbedCore`: overwriting the brain's recorded embedding config from module constants on every run would hide exactly the dim-mismatch failures we want loud. Thanks for the effort here; a single focused change with a clear failure it fixes is much easier for us to take.

Issue #2186, opened 2026-06-14
Task list has conflicting write/read contracts — tasks added via daily-task-manager never surface in live context


bcrawford18:
`resolveTodayTasks()` (the OpenClaw live-context surfacer) and `daily-task-manager` (the documented task writer) disagree on both *what heading* and *what store* the task list uses, so tasks added the documented way silently never appear. Verified against v0.42.43.0.

## The chain, traced

1. `skills/daily-task-manager/SKILL.md` saves the list via `gbrain put ops/tasks` (`put_page`) and emits headings `## P0 — Urgent` / `## P1 — Today` / `## P2` / `## P3`. `put_page` writes to the DB `pages` table only — there is no DB→disk writeback.
2. `src/core/sync.ts` (`PRUNE_DIR_NAMES`, ~L258) prunes `ops/` at any depth, so the disk file is never ingested into the DB and the DB page is never reconciled back to disk.
3. `src/core/context-engine.ts` `resolveTodayTasks()` reads the **disk file** `ops/tasks.md` via `readFileSync` and matches only a section headed literally `## Today`, lines shaped `- [ ] **bold**`.
4. `skills/briefing/SKILL.md` never reads the task file at all — action items come from `gbrain query` + timeline.

## Two independent breaks

- **Heading:** the writer emits `## P1 — Today`; the reader requires the substring `## Today`, which never occurs in `## P1 — Today`. Match fails → `[]`.
- **Storage:** even with matching headings, the writer persists to the DB while the reader reads disk, and `ops/` is pruned from sync — so the two stores never reconcile.

Net: a task added via the documented skill reaches neither the file the reader reads nor (for briefing) any path that's consulted. The only thing matching the reader's format is the test fixtures, which fabricate `## Today` input inline — so the suite passes while the real write path produces nothing the reader can see.

## Repro

1. Add a task via daily-task-manager → lands as a DB page under `## P1 — Today`.
2. `gbrain get ops/tasks` → `page_not_found` (ops/ pruned from sync).
3. `resolveTodayTasks()` → `[]` (no `## Today` on disk).
4. Briefing shows no action items.

## Suggested direction — converge on one contract

Reading the disk file by path looks correct given the deliberate `ops/` sync exclusion, so:

- Make `daily-task-manager` read/write `ops/tasks.md` by path instead of `gbrain get/put` (fixes the storage split).
[...]

PR #2188, opened 2026-06-15
Fix task manager live-context contract

caioribeiroclw-pixel:
## Summary

Fixes the task-list live-context contract described in #2186.

Changes:
- Teach `resolveTodayTasks()` to read the documented `## P1 — Today` section as well as the older `## Today` fixture shape.
- Accept both documented plain task lines (`- [ ] task`) and legacy bold task lines (`- [ ] **task**`).
- Update task-related skill docs so daily-task-manager, daily-task-prep, and briefing all use the same disk-backed `ops/tasks.md` contract instead of `gbrain get/put ops/tasks`.
- Add a regression test for the documented P1 Today format.

## Test

- `git diff --check`
- `npm exec --yes bun -- test test/context-engine.test.ts`

### DIFF
--- a/skills/briefing/SKILL.md
+++ b/skills/briefing/SKILL.md
@@ -65,12 +65,13 @@ Compile a daily briefing from brain context.
 3. **Time-sensitive threads.** Open items from timeline entries:
    - Items with deadlines in the next 48 hours
    - Follow-ups that are overdue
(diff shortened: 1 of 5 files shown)


time-attack:
Superseded with credit by #3124 (just merged) — your change landed rebased onto current master with authorship preserved via Co-authored-by. Thank you!

PR #2191, opened 2026-06-15
[codex] Expose query type filters

TheLordArgus:
## Summary

Exposes the existing `SearchOpts.types` filter on the public `query` operation so MCP/remote callers can request evidence-only or type-scoped retrieval directly.

## Root Cause

The lower search layer already supports page-type filtering, but the public `query` op schema did not accept or thread a `types` parameter. Remote agents therefore had to retrieve all page types and filter client-side, which is both noisier and less reliable for evidence-only temporal retrieval.

## Impact

Callers can now ask for queries such as `types: ["source", "conversation"]` and avoid mixing raw evidence with derived pages when they need source-grounded answers. The same filter is threaded through the text and image/vector query paths.

## Changes

- Add `types` to the public `query` operation params.
- Validate it as a string array and pass it into `hybridSearch` / vector search options.
- Add a PGLite e2e regression proving a source page is returned while a same-keyword derived company page is excluded.

## Validation

- `bun test test/e2e/source-isolation-pglite.test.ts`

### DIFF
--- a/src/core/operations.ts
+++ b/src/core/operations.ts
@@ -1435,6 +1435,12 @@ const query: Operation = {
     image_mime: { type: 'string', description: 'MIME type for the image bytes (auto-derived from path on CLI; required when calling op directly).' },
     limit: { type: 'number', description: 'Max results (default 20)' },
     offset: { type: 'number', description: 'Skip first N results (for pagination)' },
+    types: {
+      type: 'array',
+      description:
+        "Filter to page types, e.g. ['source','conversation'] for evidence-only temporal retrieval. " +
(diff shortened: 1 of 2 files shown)


time-attack:
Closing — this adds a new parameter to the public `query` contract, and since CLI and MCP tool schemas are both generated from operations.ts, that's permanent surface on the most-used operation rather than an internal change. The underlying SearchOpts filter isn't broken; this is a convenience exposure, and we're holding the line on query-surface growth. Thanks for the clean threading and for including an isolation test.

PR #2192, opened 2026-06-15
[codex] Fix search date filters to use event dates

TheLordArgus:
## Summary

Fixes temporal search filtering so `since` / `until` operate on page event time first, using `effective_date` before falling back to `updated_at` / `created_at`. It also makes the lower bound inclusive and normalizes plain `YYYY-MM-DD` `until` values to the next day for the existing exclusive upper-bound comparison.

## Root Cause

The Postgres search paths were filtering dates against `COALESCE(p.updated_at, p.created_at)`, so recently reimported or refreshed pages could appear in a recent-event query even when their actual `effective_date` was old. PGLite already used `effective_date`, but its lower bound was exclusive.

## Impact

Temporal queries such as "what happened since May 20" can otherwise include old pages that were merely touched recently, or miss events that fall exactly on the lower-bound date.

## Changes

- Use `COALESCE(p.effective_date, p.updated_at, p.created_at)` in Postgres search date filters.
- Make lower bounds inclusive in Postgres and PGLite.
- Normalize plain-date `until` values to the next-day timestamp before passing them to the exclusive upper-bound engine filter.
- Add a regression that inserts an old event refreshed during the test and verifies it is excluded by a May `afterDate`, while the boundary date is included.

## Validation

- `bun test test/chunk-grain-fts.test.ts`

### DIFF
--- a/src/core/pglite-engine.ts
+++ b/src/core/pglite-engine.ts
@@ -1548,7 +1548,7 @@ export class PGLiteEngine implements BrainEngine {
     // got reimported). Same param shape as Postgres engine.
     if (opts?.afterDate) {
       params.push(opts.afterDate);
-      extraFilter += ` AND COALESCE(p.effective_date, p.updated_at, p.created_at) > $${params.length}::timestamptz`;
+      extraFilter += ` AND COALESCE(p.effective_date, p.updated_at, p.created_at) >= $${params.length}::timestamptz`;
     }
     if (opts?.beforeDate) {
       params.push(opts.beforeDate);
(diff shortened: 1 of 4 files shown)


time-attack:
Closing — the engine-parity bug this targets was already fixed on master by #1706 (commit d33aee84), which moved all three postgres-engine since/until clauses onto `COALESCE(p.effective_date, p.updated_at, p.created_at)`. The remaining delta here is an inclusive/exclusive boundary flip plus a new date-rounding helper, which is a semantics change on a shipped filter rather than the parity fix described, so I'd rather not take it as part of this draft. Thanks for digging into the engine parity — that instinct was right, it just landed elsewhere first.

PR #2197, opened 2026-06-15
fix(facts): end facts:absorb sync abort race + 2 related synthesis-tier fixes

hypergrowthagents:
## Summary

Three related fixes to the `facts:absorb` synthesis tier, the largest of which (RC2) ends a shutdown-time abort race that makes `gbrain sync`'s high-only fact synthesis fail almost every time.

> **Draft / review requested.** The abort race depends on the autopilot/CLI-exit timing and is hard to exercise deterministically in a unit test, so this is opened as a draft. Unit/typecheck results are below; I'm asking a maintainer to confirm the design choice (inline-await vs. raising the drain grace) and to validate against a live sync cycle. Happy to adjust.

## Bug 1 (RC2) — `facts:absorb` abort race during sync

**Diagnosis.** `sync.ts` enqueues the high-only fact absorb fire-and-forget (`mode:'queue'`, `notabilityFilter:'high-only'`). On CLI exit, `drainAllBackgroundWorkForCliExit` (`src/core/background-work.ts`) gives the facts queue only ~1–2s (the two call sites in `src/cli.ts` pass `timeoutMs: 1000` and the default `2000`), then calls the registered `abort()` → `FactsQueue.shutdown()` (`src/core/facts/queue.ts`), which sets `shuttingDown=true` and fires `this.internalAbort.abort()`. That signal is threaded into the in-flight `chat()` call, cancelling synthesis mid-flight. A single-shot synthesis call routinely takes longer than the grace, so for sync the race is effectively deterministic.

**Evidence (live brain, v0.42.26.0).** `gbrain doctor` → `facts_extraction_health`:

```
[OK] facts_extraction_health: Facts:absorb activity in last 24h (under threshold 10): default: 4 pipeline_error.
```

The underlying ingest_log rows are the recurring shape:

```
pipeline_error: [chat(openrouter:deepseek/deepseek-v4-flash)] The operation was aborted
```

**Fix (chosen: inline-await, the minimal localized change).** Switch the sync high-only absorb from `mode:'queue'` to `mode:'inline'` so the synthesis completes *within* the sync command, before the CLI-exit drain ever runs — it can no longer be raced by `shutdown()`. Sync now pays synthesis latency inline, but that is bounded by the two guards already on this block: `pagesAffected.length <= 50` and the `high-only` notability filter (only notable pages reach the LLM). Per-turn chat keeps fire-and-forget `queue` mode where latency matters.

I preferred this over raising the facts-queue drain grace to ~30s because (a) it removes the race entirely on the affected path rather than widening a timing window that can still lose under a slow provider, (b) it's localized to the one surface that was enqueuing synthesis fire-and-forget, and (c) raising the grace would make *every* CLI command that enqueues facts (capture, import, sync) hang up to ~30s on exit. Open to the grace approach instead if maintainers prefer to keep sync's facts path non-blocking — happy to switch.

Inline mode bubbles pipeline errors to the caller (queue mode absorb-logs them in its worker), so the catch block now writes the same `ingest_log` row via `classifyFactsAbsorbError` + `writeFactsAbsorbLog`, preserving the D5 failure-visibility contract for doctor + dashboard. Best-effort — a sync never fails because one page's synthesis did.

## Bug 2 — `resolveModel` tier override precedence

`resolveModel` (`src/core/model-config.ts`) checked `models.default` *before* the tier override, so a `tier:'reasoning'` (or any tier) caller could never override a set `models.default` — the broad default swallowed every tier-specific config. Reordered so `models.tier.<tier>` is consulted before `models.default`. The CLI flag and explicit configKey / deprecated key still outrank the tier override, and the subagent capability gate (`enforceSubagentCapable`) still runs on each resolved value — including the `models.default` fallthrough — so subagent routing is unchanged. Doc comments + the tier-system test updated to the corrected precedence.

## Bug 3 — abort errors misclassified as `pipeline_error`

`classifyFactsAbsorbError` (`src/core/facts/absorb-log.ts`) bucketed AbortError / "operation was aborted" as the catch-all `pipeline_error`, conflating a benign shutdown race with a genuine synthesis failure (and inflating the pipeline-error rate doctor reads). Added a branch mapping `/operation was aborted|AbortError/i` (and the `AbortError` name) to the existing `queue_shutdown` reason — it *is* a shutdown-driven cancellation, and reusing the existing reason avoids a new reason code / doctor threshold.

Note: with Bug 1 fixed, the sync path should stop producing these aborted errors entirely; Bug 3 still correctly classifies any abort race that surfaces from other paths or timing.

## Files changed

| File | Change |
|---|---|
| `src/commands/sync.ts` | high-only sync absorb `mode:'queue'` → `mode:'inline'`; catch block writes absorb-log row to preserve failure visibility |
| `src/core/model-config.ts` | reorder `resolveModel` so tier override beats `models.default`; doc comments |
| `src/core/facts/absorb-log.ts` | `classifyFactsAbsorbError`: map abort errors → `queue_shutdown` |
| `test/model-config.serial.test.ts` | flip "models.default beats tier override" → "tier override beats models.default"; add no-tier-config fallthrough test |
| `test/facts-absorb-log.test.ts` | add abort-race classification test using the live-log error shape |

## What was tested locally

- `bun run typecheck` (tsc --noEmit) — **clean**.
- `bun test` on the affected suites — **all pass**:
  - `test/model-config.serial.test.ts` — 21 pass
  - `test/facts-absorb-log.test.ts` — 17 pass
  - `test/facts-queue.test.ts` + `test/facts-queue-drain-pending.test.ts` — 13 pass
  - `test/core/background-work.test.ts` — 6 pass
  - `test/facts-backstop*.test.ts` (3 files) — 24 pass
  - `test/sync.test.ts` + `test/sync-inline-extract-stamps.serial.test.ts` — 61 pass

## What was NOT tested locally

- **The RC2 race itself end-to-end.** It depends on the autopilot/CLI-exit drain timing and isn't deterministically reproducible in a unit test. The fix is reasoned from the code path + the live ingest-log evidence above; I have not observed a post-fix sync cycle confirming the `aborted` rows stop. Maintainer validation against a live sync is requested.
- No e2e suite run (no `DATABASE_URL` in this environment).
- One pre-existing repo lint failure (`check:trailing-newline` on `test/fixtures/e5-lease-cap-ab/2026-05-24-baseline-dry-run.json`) is unrelated to this PR and untouched.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/commands/sync.ts
+++ b/src/commands/sync.ts
@@ -2119,14 +2119,28 @@ async function performSyncInner(engine: BrainEngine, opts: SyncOpts): Promise<Sy
     } catch { /* extraction is best-effort */ }
   }
 
-  // v0.31.2: facts extraction now routes through the shared
-  // src/core/facts/backstop.ts helper (PR1 commit 6). Sync uses
-  // queue mode (fire-and-forget) + 'high-only' filter so a 50-page
-  // sync doesn't block on N sequential Sonnet calls. The pre-fix
-  // inline loop is gone — it carried (a) a dead-code type filter
-  // ('conversation'/'transcript'/'therapy'/'call' aren't real
-  // PageTypes), (b) a divergent eligibility shape from put_page,
-  // and (c) raw extract→insert without dedup/supersede.
+  // v0.31.2: facts extraction routes through the shared
+  // src/core/facts/backstop.ts helper (PR1 commit 6) with the 'high-only'
+  // notability filter (HIGH lands now, MEDIUM waits for the dream cycle,
+  // LOW dropped at the LLM layer). The pre-fix inline loop is gone — it
+  // carried (a) a dead-code type filter ('conversation'/'transcript'/
+  // 'therapy'/'call' aren't real PageTypes), (b) a divergent eligibility
+  // shape from put_page, and (c) raw extract→insert without dedup/supersede.
+  //
+  // RC2 fix (facts:absorb abort race): this runs INLINE (await the full
+  // pipeline) rather than fire-and-forget 'queue' mode. Sync was the only
+  // surface enqueuing high-only synthesis fire-and-forget, and on CLI exit
+  // `drainAllBackgroundWorkForCliExit` gives the facts queue only a short
+  // grace before `queue.shutdown()` fires `internalAbort.abort()` — which
+  // cancels the in-flight chat() mid-synthesis and lands a recurring
+  // `pipeline_error: ... The operation was aborted` in the ingest log.
+  // A single-shot synthesis call routinely outlives that grace, so the
+  // race was effectively deterministic for sync. Awaiting here completes
+  // the synthesis WITHIN the sync command (before the exit drain runs), so
+  // it can't be raced by shutdown. Cost: sync now pays synthesis latency
+  // inline, but it's bounded by the existing `pagesAffected.length <= 50`
+  // guard AND the 'high-only' filter (only notable pages reach the LLM).
+  // Per-turn chat keeps fire-and-forget 'queue' mode where latency matters.
   if (!opts.noExtract && pagesAffected.length > 0 && pagesAffected.length <= 50) {
     const { runFactsBackstop } = await import('../core/facts/backstop.ts');
     const factsSourceId = opts.sourceId ?? 'default';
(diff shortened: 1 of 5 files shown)


time-attack:
Closing this one. The inline-await change trades a log-noise problem for making `gbrain sync` block on up to 50 sequential synthesis calls, and the `resolveModel` reorder changes behavior that master documents as intentional (`models.default` above the tier override) rather than fixing a bug — that's a routing change that would silently move traffic for anyone who sets both keys. The AbortError-to-`queue_shutdown` classification tweak is reasonable on its own but not enough to carry the other two. Thanks for the careful diagnosis write-up; the ingest_log evidence was genuinely useful.

PR #2203, opened 2026-06-15
fix(gateway): disable thinking for all openai-compatible providers (#2202)

### LINKED ISSUE #2202: OpenAI-compatible providers with default thinking (e.g. DeepSeek) waste ~50% tokens on non-reasoning calls
## Problem

Several OpenAI-compatible providers, notably **DeepSeek v4-flash**, default to thinking/reasoning mode. When gbrain uses these models for non-reasoning calls (embeddings, search expansion, linting, etc.), every response carries unnecessary thinking tokens, inflating token costs by roughly 50%.

DeepSeek v4-flash is a popular provider for gbrain because of its competitive pricing and good performance, but the default thinking mode silently increases billable tokens for every chat call that gbrain makes — including short, fact-based queries that gain nothing from reasoning.

## Reproduction

1. Configure gbrain with a DeepSeek provider (or any openai-compatible provider that supports thinking)
2. Run any gbrain command that triggers a chat call: `gbrain query`, `gbrain search`, `gbrain embed`, etc.
3. Observe that the response includes reasoning/thinking tokens for every call

## Root Cause

The `chat()` function in `src/core/ai/gateway.ts` does not explicitly disable thinking for openai-compatible providers. Unlike Anthropic Claude (which uses a different parameter `thinking` under `providerOptions.anthropic`) and OpenAI o-series (which uses `reasoning_effort`), openai-compatible providers pass their thinking parameter through `providerOptions[recipe.id]`, which gets spread into the request body. If the provider defaults to thinking enabled, gbrain has no way to opt out.

## Proposed Fix

Explicitly set `thinking: { type: "disabled" }` for all openai-compatible providers (`recipe.implementation === "openai-compatible"`) in the `chat()` function. The `@ai-sdk/openai-compatible` provider spreads `providerOptions[recipe.id]` into the request body; providers that do not support thinking silently ignore the parameter — so the fix is safe for ALL openai-compatible providers, not just DeepSeek.

This does not affect native providers (OpenAI o-series, Anthropic Claude, Google Gemini), which use different parameter names for reasoning control.

## Impact

- Token waste: ~50% reduction in unnecessary output tokens for DeepSeek users
- Backward compatible: no change in behavior for providers that don't support thinking
- Future-proof: covers providers (Ollama, OpenRouter, Groq, Together, llama-server) that may add thinking support later

justemu:
Closes #2202

## Summary

Explicitly disables thinking/reasoning mode for all openai-compatible providers. DeepSeek v4-flash defaults to thinking mode, producing ~50% unnecessary output tokens for gbrains non-reasoning use cases (search, embed, lint, etc.).

## Change

In `src/core/ai/gateway.ts`, `chat()` function sets `thinking: { type: "disabled" }` via `providerOptions.openaiCompatible` when `recipe.implementation === "openai-compatible"`.

## Safety

- The `@ai-sdk/openai-compatible` provider spreads `providerOptions[recipe.id]` into the request body
- Providers that do not support thinking silently ignore the parameter - 100% safe
- Does not affect native providers (OpenAI o-series, Anthropic Claude, Google Gemini)

cc: @justemu

### DIFF
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -2735,6 +2735,20 @@ export async function chat(opts: ChatOpts): Promise<ChatResult> {
     providerOptions.anthropic = { cacheControl: { type: 'ephemeral' } };
   }
 
+  // Disable thinking/reasoning for all openai-compatible providers.
+  // DeepSeek v4-flash defaults to thinking mode, producing ~50% unnecessary
+  // output tokens for gbrain's non-reasoning use cases. Other providers
+  // (Ollama, OpenRouter, Groq, Together, llama-server) may add thinking in
+  // the future — the parameter is silently ignored by unrecognizing providers.
+  // Does not affect native providers (OpenAI o-series, Anthropic Claude,
+  // Google Gemini) which use different parameter names for reasoning control.
+  if (recipe.implementation === 'openai-compatible') {
+    providerOptions.openaiCompatible = {
+      ...((providerOptions as any).openaiCompatible ?? {}),
+      thinking: { type: 'disabled' as const },
+    };
+  }
+
   let _budgetRecorded = false;
   const _recordBudget = (modelLabel: string, inputTokens: number, outputTokens: number): void => {
     if (!tracker || _budgetRecorded) return;


time-attack:
Thanks @justemu! 🙏

What this does: PR #2203 attempts to disable reasoning/thinking mode for DeepSeek and other OpenAI-compatible providers to save ~50% wasted tokens on non-reasoning tasks. However, it uses the wrong configuration key: providerOptions.openaiCompatible…

This overlaps with #2670, which we're carrying forward, so I'm closing this in its favor to keep the review queue focused. If yours covers a case #2670 misses, reopen with that detail and we'll reconcile them.

time-attack:
right now im kind of conflicted on what to do about this issue. right now there are many different ways of calling AI in gbrain and i still am not keen on the current solution. i figure whatever i think of later on will fix this reasoning issue. 

also i think this issue is super weird, idk why deepseek works this way. 

but if you have another useful PRs please add them!!!

PR #2205, opened 2026-06-15
fix(embed): only reject as user_provided_model_unset when modelId is absent

### LINKED ISSUE #2204: diagnoseEmbedding() falsely returns user_provided_model_unset when modelId is parsed from config string
## Problem

When using a non-default embedding provider with `user_provided_models: true` (such as a local LM Studio / llama-server embedding endpoint), `diagnoseEmbedding()` in `src/core/ai/gateway.ts` unconditionally returns `user_provided_model_unset` if the recipe has an empty `models: []` array, even when a valid model name is supplied via the config string.

For example, a config like:
```
embedding_model: "llama-server:text-embedding-qwen3-embedding-0.6b"
```
correctly parses `modelId = "text-embedding-qwen3-embedding-0.6b"`, but the preflight guard ignores it and rejects the call.

This makes custom embedding providers (LM Studio, Ollama, llama-server, vLLM) entirely non-functional when configured through `user_provided_models`.

## Reproduction

1. Set up gbrain with a custom embedding provider (e.g. LM Studio running on localhost):
   ```
   embedding_model: "llama-server:text-embedding-model-name"
   ```
   with provider config:
   ```json
   {
     "providers": {
       "llama-server": {
         "base_url": "http://localhost:1234/v1",
         "user_provided_models": true,
         "models": []
       }
     }
   }
   ```
2. Run any command that triggers embedding: `gbrain search`, `gbrain embed`
3. Observe: embedding fails with `user_provided_model_unset`

## Root Cause

In `src/core/ai/gateway.ts`, the `diagnoseEmbedding()` function checks:
```typescript
if (
  Array.isArray(tp.models) &&
  tp.models.length === 0 &&
) {
}
```

The guard rejects based solely on `models` array being empty and `isUserProvided` being true, without checking whether `parsed.modelId` was successfully extracted from the config string. If the user has specified a model name in the embedding_model config, the guard should accept it.

## Proposed Fix

Add `!parsed.modelId` to the condition:
```typescript
if (
  Array.isArray(tp.models) &&
  tp.models.length === 0 &&
  !parsed.modelId
) {
}
```

When a valid modelId has been parsed from the config string, the guard no longer rejects.

## Impact

- Fixes local LM Studio / llama-server / Ollama / vLLM embedding for all gbrain users
- Minimal change: one extra conjunct in an existing condition
- No behavioral change for other embedding configurations

justemu:
Closes #2204

## Summary

Fixes `diagnoseEmbedding()` returning `user_provided_model_unset` when a valid modelId has been parsed from the config string. This makes custom embedding providers (LM Studio, llama-server, Ollama, vLLM) functional when configured through `user_provided_models`.

## Change

In `src/core/ai/gateway.ts`, adds `!parsed.modelId` to the preflight rejection condition so that a successfully parsed model name is not falsely rejected.

## Root Cause

The guard checked `models.length === 0 && isUserProvided` without checking whether `parsed.modelId` was populated. A valid config string like `llama-server:text-embedding-qwen3-0.6b` correctly parses a modelId, but the guard ignored it and rejected anyway.

## Testing

- Auto-merge was clean against v0.42.44.0 (no conflicts)
- Diff: +1 line, -1 line in a single condition
- Covers both `litellm` and generic `user_provided_models` paths

cc: @justemu

### DIFF
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -709,7 +709,8 @@ export function diagnoseEmbedding(modelOverride?: string): EmbeddingDiagnosis {
   if (
     Array.isArray(tp.models) &&
     tp.models.length === 0 &&
-    (recipe.id === 'litellm' || isUserProvided)
+    (recipe.id === 'litellm' || isUserProvided) &&
+    !parsed.modelId
   ) {
     return {
       ok: false,


justemu:
#1938 is dealing with the same issue

time-attack:
Thanks for this, @justemu! 🙏

What this does: This PR adds a !parsed.modelId check to fix llama-server embedding by preventing false-positive user_provided_model_unset errors. However, that entire guard has been removed from master (v0.42.58.0, released July 10) and replaced with a…

PR #2210, opened 2026-06-15
docs(ai/recipes/litellm): document /v1 suffix in setup_hint (#2209)

### LINKED ISSUE #2209: litellm: setup_hint silent on /v1 path convention; OpenAI-shaped proxies 401 with no signal
**What happened?**

The litellm recipe's `setup_hint` says only "set LITELLM_BASE_URL" with no guidance on whether to include a path prefix. Vanilla LiteLLM serves `/chat/completions` at the host root, so `LITELLM_BASE_URL=http://host:port` works. But OpenAI-shaped proxies (codex-proxy, some Azure-OpenAI mirrors, third-party translators) serve only `/v1/chat/completions` and reject the bare path with **401 Unauthorized** (not 404, which would be the diagnostic-correct answer).

The AI SDK's `createOpenAICompatible` appends `/chat/completions` to the base URL. When the proxy expects `/v1/chat/completions` and gets `/chat/completions`, it returns a generic 401. The AI SDK surfaces this as `AIConfigError: Unauthorized — Check your API key`, so the user's natural diagnostic move is to rotate or re-check the bearer token. That dead-ends. The actual fix is appending `/v1` to `LITELLM_BASE_URL`.

**What did you expect?**

The setup_hint should call out the path convention difference so users hitting 401s with codex-proxy-class endpoints know to check the URL shape, not the bearer.

**Steps to reproduce**

1. Run codex-proxy (or any openai-shaped endpoint that serves `/v1/...` only) on `:8080`.
2. `export LITELLM_BASE_URL=http://localhost:8080` (no `/v1`).
3. `gbrain config set models.tier.reasoning litellm:gpt-5.4`.
4. Any chat dispatch → `AIConfigError: Unauthorized`. `curl $LITELLM_BASE_URL/models` returns 200 with the same bearer.
5. `export LITELLM_BASE_URL=http://localhost:8080/v1` → success.

**Environment**

- gbrain version: 0.42.42.0+ (reproduces against current `master`).
- Reproduces with any openai-shaped reverse proxy (codex-proxy and similar translators); vanilla LiteLLM is unaffected.

**Proposed fix (doc-only)**

Append path-convention guidance to `src/core/ai/recipes/litellm-proxy.ts`'s `setup_hint`. Suggested shape (matches the existing one-paragraph string concatenation pattern from `llama-server-reranker.ts`):

```text
Run LiteLLM (https://docs.litellm.ai) in front of any provider; set
LITELLM_BASE_URL + pass --embedding-model litellm:<model> and
--embedding-dimensions <N>. Path convention: vanilla LiteLLM serves
/chat/completions at the host root (LITELLM_BASE_URL=http://host:port).
OpenAI-shaped proxies (codex-proxy, some Azure-OpenAI mirrors,
third-party translators) serve /v1/chat/completions and require the
/v1 suffix (LITELLM_BASE_URL=http://host:port/v1). If chat dispatch
returns 401 Unauthorized but `curl $LITELLM_BASE_URL/models` returns
200 with a valid bearer, the /v1 suffix is the missing piece.
```

**Adjacent surface (worth flagging or splitting off):** `gbrain models doctor` could probe `${LITELLM_BASE_URL}/chat/completions` and `${LITELLM_BASE_URL}/v1/chat/completions` with the configured bearer and surface the path-shape mismatch explicitly. Two-line diff to the doctor probe logic; saves the next user 20 minutes of confusion. Out of scope for the doc-only PR; happy to file as a follow-up.

PR ready for review: #2210 .

brettdavies:
Closes #2209.

## Summary

**The bug.** The litellm `setup_hint` says "set LITELLM_BASE_URL" with no guidance on the path prefix. Vanilla LiteLLM serves `/chat/completions` at the host root, so `LITELLM_BASE_URL=http://host:port` works. OpenAI-shaped proxies (codex-proxy, some Azure-OpenAI mirrors, third-party translators) serve only `/v1/chat/completions` and reject the bare path with 401 Unauthorized. The AI SDK surfaces the 401 as `AIConfigError: Unauthorized — Check your API key`, so the diagnostic dead-ends on the bearer rather than the URL shape. `curl $LITELLM_BASE_URL/models` returns 200 with the same bearer, which makes the 401 doubly misleading.

**The fix.** Append three sentences to `src/core/ai/recipes/litellm-proxy.ts`'s `setup_hint`: the vanilla form (root), the OpenAI-shaped form (`/v1` suffix), and the cross-check (`curl $LITELLM_BASE_URL/models` returns 200 with the bearer → `/v1` is the missing piece). One-paragraph string concatenation matches the existing setup_hint shape; `llama-server-reranker.ts` is the precedent.

**Adjacent surface (worth filing as a follow-up, not bundled here):** `gbrain models doctor` could probe `${LITELLM_BASE_URL}/chat/completions` and `${LITELLM_BASE_URL}/v1/chat/completions` with the configured bearer and surface the path-shape mismatch explicitly. Out of scope for this doc-only PR.

## Tests

- No behavior change; the only edit is the `setup_hint` string. Recipe-level tests (`test/ai/recipes-existing-regression.test.ts`, `test/ai/gateway.test.ts`): 53/53 pass.
- `bun run typecheck` clean. `bun run verify` clean (30 parallel checks).

### DIFF
--- a/src/core/ai/recipes/litellm-proxy.ts
+++ b/src/core/ai/recipes/litellm-proxy.ts
@@ -41,5 +41,9 @@ export const litellmProxy: Recipe = {
       supports_multimodal: true,
     },
   },
-  setup_hint: 'Run LiteLLM (https://docs.litellm.ai) in front of any provider; set LITELLM_BASE_URL + pass --embedding-model litellm:<model> and --embedding-dimensions <N>.',
+  setup_hint:
+    'Run LiteLLM (https://docs.litellm.ai) in front of any provider; set LITELLM_BASE_URL + pass --embedding-model litellm:<model> and --embedding-dimensions <N>. ' +
+    'Path convention: vanilla LiteLLM serves /chat/completions at the host root (LITELLM_BASE_URL=http://host:port). ' +
+    'OpenAI-shaped proxies (codex-proxy, some Azure-OpenAI mirrors, third-party translators) serve /v1/chat/completions and require the /v1 suffix (LITELLM_BASE_URL=http://host:port/v1). ' +
+    'If chat dispatch returns 401 Unauthorized but `curl $LITELLM_BASE_URL/models` returns 200 with a valid bearer, the /v1 suffix is the missing piece.',
 };


time-attack:
Thanks for chasing this down — the diagnosis in #2209 was exactly right, and the misleading 401 was a real trap. This has since landed on master: commit a25209bb (v0.42.58.0, #2627) added the `/v1` suffix guidance to the litellm recipe's `setup_hint`, citing your issue. Closing as already fixed; the follow-up you suggested about `gbrain models doctor` probing both path shapes is the more interesting remaining piece.

Issue #2213, opened 2026-06-16
Dream synthesize can reverse-write nested wiki/ paths for source-aware wiki repos


barronlroth:
## Summary

`gbrain dream --phase synthesize` can reverse-write generated pages into a nested `wiki/` directory when the active source id is already `wiki` and the source root is the wiki checkout.

In a source-aware setup, dream synthesis currently prompts/allows child jobs to write slugs shaped like:

```text
wiki/personal/reflections/...
wiki/originals/ideas/...
```

If source `wiki` maps to `/path/to/wiki`, reverse-rendering those slugs to disk produces:

```text
/path/to/wiki/wiki/personal/reflections/...
/path/to/wiki/wiki/originals/ideas/...
```

That conflates the **source id** (`wiki`) with the **slug/path inside the source** (`personal/reflections/...`).

## Why this seems wrong

The documented brains/sources model says:

- a source is the named repo inside a brain (`source_id = wiki`)
- slugs are scoped per source
- slug uniqueness is `(source_id, slug)`, not global slug strings

So for a wiki source rooted at `/path/to/wiki`, the source-aware representation should be one of:

```text
source_id = wiki
slug = personal/reflections/...
```

not:

```text
source_id = default or wiki
slug = wiki/personal/reflections/...
```

## Reproduction shape

1. Configure/add a source named `wiki` whose local path is the wiki repo root, e.g. `/Users/example/wiki`.
2. Run a non-dry dream synthesize pass that writes reflection/original pages.
3. Observe generated markdown under a nested `wiki/` directory inside the repo root.

Observed locally after a controlled smoke run:

```text
/Users/.../wiki/wiki/personal/reflections/...
/Users/.../wiki/wiki/originals/ideas/...
```

## Expected behavior

Dream synthesis should keep source id and in-source slug separate, so generated files land under:

```text
/path/to/wiki/personal/reflections/...
/path/to/wiki/originals/ideas/...
```

## Possible fix direction

Avoid a global string hack. The safe fix should be source-aware:

- either make synthesize prompts/allow-lists use in-source slugs while passing the resolved `sourceId` to subagent tools / reverse writes, or
- normalize only at the source boundary: when reverse-writing to a source root, treat a leading `<sourceId>/` prefix as a virtual prefix and strip it before joining to the source root, with regression tests.

Regression coverage should include:

- `source_id = wiki`, `slug = wiki/personal/reflections/foo` does not write `/wiki/wiki/...`
[...]

PR #2214, opened 2026-06-16
dream: support Codex OAuth synthesize on PGLite

barronlroth:
## Summary

- add `openai-codex` as a chat recipe backed by local Codex OAuth credentials
- run PGLite dream synthesize subagent jobs through a scoped inline worker instead of waiting for an external worker that cannot open the same PGLite file
- force Codex Responses chat calls into stateless `store:false` mode so GBrain's manual tool loop can round-trip tool results
- prevent dream reverse-writes from creating nested `wiki/wiki/...` paths when `.gbrain-source` already names the active source

## Why

This fixes the local failure mode where `gbrain dream --phase synthesize` could judge transcripts successfully but never drain subagent jobs on PGLite, and then, once drained, reverse-write dream pages under a nested `wiki/` directory for source-aware wiki repos.

Related issue: #2213

## Verification

```bash
bun test test/gateway-model-messages.test.ts test/cycle-synthesize-chunker.test.ts test/core/ai/codex-oauth.test.ts
bun run typecheck
```

Also smoke-tested locally with a controlled non-dry synthesize run using `openai-codex:gpt-5.5`; it completed one subagent and reverse-wrote two pages to `originals/...` and `personal/...`, not nested `wiki/...`.

### DIFF
--- a/src/core/ai/codex-oauth.ts
+++ b/src/core/ai/codex-oauth.ts
@@ -1,0 +1,376 @@
+import { existsSync, mkdirSync, readFileSync, renameSync, writeFileSync } from 'node:fs';
+import { dirname, join } from 'node:path';
+import { homedir } from 'node:os';
+
+import { AIConfigError, AITransientError } from './errors.ts';
+
+const DEFAULT_CODEX_BASE_URL = 'https://chatgpt.com/backend-api/codex';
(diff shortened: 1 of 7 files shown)


time-attack:
Closing after maintainer review — decision: not accepting this. Close it and ask the author to resubmit only part (2) — the stalling fix — on its own; that piece is confirmed still missing and valuable.

PR #2216, opened 2026-06-16
feat(orphans): user-extensible exclude prefixes via brain config (#2215)

### LINKED ISSUE #2215: gbrain orphans / orphan_ratio overcount on PARA / Obsidian / agent-organized brains — no user-extensible deny list
## Symptom

`gbrain orphans` reports thousands of "orphan" pages on brains organized by anything other than gbrain's own internal conventions. The same overcount drives `orphan_ratio` in `gbrain doctor`'s brain-score, so the user's brain-score is dominated by reference material that is orphan-by-design.

Two distinct overcount sources contribute, each living at a different filter layer:

1. **Slug-prefix layer**: PARA brains have `resources/`, `archives/`, `inbox/`; Obsidian brains have `clips/`, `attachments/`; code corpora have `transcripts/` (gstack-memory-ingest landing dir); agent runtimes have `agents/<name>/state`. The shipped `DENY_PREFIXES` doesn't cover any of these.
2. **Page-type layer**: code-symbol extractions (`pages.type = 'code'`), system artifacts (`extract_receipt`), gstack-memory-ingest commit-summary pages (`feat`, `fix`, `refactor`, `chore`). These have no inbound-link semantics regardless of slug.

## Concrete numbers

A 2K-page Obsidian vault organized PARA-style plus 11 federated code-stage source repos (gstack-memory-ingest) plus a `transcripts/` corpus:

```text
$ gbrain orphans --count
2906       # raw (shipped DENY_PREFIXES only)

# Adding PARA convention dirs to the prefix knob:
$ gbrain config set orphans.exclude_prefixes 'resources/,transcripts/,agents/,clips/'
$ gbrain orphans --count
1418       # dropped 1488 PARA convention pages

# Adding default page-type filter (post-fix):
$ gbrain orphans --count
928        # dropped 490 code + system + commit-summary pages
```

928 actual signal vs 2906 raw. The 928 remaining are dominated by `concept` (414), `project` (291), and `note` (204) types: pages the user actually wrote and might want to graph.

## Diagnosis

`src/commands/orphans.ts:47-56` carries the shipped exclusion lists:

```ts
const DENY_PREFIXES = ['output/', 'dashboards/', 'scripts/', 'templates/', 'openclaw/config/'];
const FIRST_SEGMENT_EXCLUSIONS = new Set(['scratch', 'thoughts', 'catalog', 'entities']);
```

These cover gbrain-internal convention dirs but miss the convention layer AND the page-type layer entirely. There's no mechanism for a user to extend the deny list or filter by page type short of patching source and rebuilding.

## Proposed fix

Two complementary per-brain config knobs, layered at the surface where the noise actually lives:

```text
# Slug-prefix layer (opt-in; additive to shipped DENY_PREFIXES):
gbrain config set orphans.exclude_prefixes 'resources/,transcripts/,agents/'

# Page-type layer (shipped defaults apply automatically; additive via config):
gbrain config set orphans.exclude_types 'guide,reference'
```

Shipped `DEFAULT_EXCLUDE_TYPES` covers types whose pages are orphan-by-construction on every brain shape:

- `code`: code-symbol extractions. Code files don't wikilink to code files.
- `extract_receipt`: system artifact from extract phases.
- `feat`, `fix`, `refactor`, `chore`: gstack-memory-ingest commit-summary types.

Same one-directional contract as `DENY_PREFIXES`: shipped entries always apply; user config extends but cannot remove.

The fix touches `src/commands/orphans.ts` (helpers + filter wiring), `src/core/engine.ts` + both engines (`findOrphanPages` gains optional `excludeTypes`), `src/core/config.ts` (registry), plus tests.

## Why this fixes more than just `gbrain orphans --count`

`getOrphansData` is the canonical "what counts as an orphan in this brain" pure-data function shared between `gbrain orphans` and doctor's `orphan_ratio` check (per the v0.42.0.0 D1 unification). Both knobs therefore correct the brain-score input. A brain federating code-stage sources stops being penalized for code-symbol extractions, and a PARA-organized brain stops being penalized for web clips, the moment the config lands.

## Behavior change to call out

The `DEFAULT_EXCLUDE_TYPES` list shifts orphan counts downward for every brain that has any of those six types. The orphan-ratio doctor check stops counting code-symbol and commit-summary pages. This is a one-time correction, not a regression. A code-symbol page never had inbound-link semantics; the count was just lying about it.

## Future surface (out of scope here, worth noting)

- **Frontmatter-level `orphan_expected: true` per-page opt-out.** Complementary to dir-level prefixes + type-level filters; useful for one-off pages where neither rule fits. Worth a follow-up.
- **Glob support** for `orphans.exclude_prefixes`. The existing semantics are `startsWith`; adding globs would need a wider rework. Prefix-only is sufficient for the patterns this PR targets.

Adjacent issue from the same operational-health theme: `orphan_ratio` recommends `extract links --by-mention` as the remediation, but backlink counting excludes mention-only links so the recommendation is structurally unactionable. Out of scope here; flagging for theme awareness.

brettdavies:
Closes #2215.

## Summary

`gbrain orphans` ships a hardcoded `DENY_PREFIXES` list in `src/commands/orphans.ts` covering gbrain-internal convention dirs (`output/`, `dashboards/`, `scripts/`, `templates/`, `openclaw/config/`). Brains organized by other conventions OR carrying code-symbol / system-artifact pages surface thousands of "orphans" that are reference material or structurally noise-by-construction. Two distinct overcount sources contribute, each living at a different filter layer:

1. **Slug-prefix layer**: PARA brains have `resources/`, `archives/`, `inbox/`; Obsidian brains have `clips/`, `attachments/`; code corpora have `transcripts/`; agent runtimes have `agents/<name>/state`.
2. **Page-type layer**: code-symbol extractions (`pages.type = 'code'`), system artifacts (`extract_receipt`), gstack-memory-ingest commit-summary pages (`feat`, `fix`, `refactor`, `chore`). These have no inbound-link semantics regardless of where they live in the slug tree.

On a 2K-page PARA vault + 11 federated code-stage source repos: 2906 raw orphans, of which 928 are real signal. The other ~1980 are noise from the two patterns above.

## The fix

Two complementary per-brain config knobs, layered at the surface where the noise lives:

### Knob 1: slug-prefix layer (opt-in; additive to shipped DENY_PREFIXES)

```text
gbrain config set orphans.exclude_prefixes 'resources/,transcripts/,agents/'
```

### Knob 2: page-type layer (shipped defaults apply automatically; additive via config)

```text
gbrain config set orphans.exclude_types 'guide,reference'
```

The shipped `DEFAULT_EXCLUDE_TYPES` covers types whose pages are orphan-by-construction on every brain shape: `code`, `extract_receipt`, `feat`, `fix`, `refactor`, `chore`. Same one-directional contract as `DENY_PREFIXES` — shipped entries always apply; user config extends but cannot remove.

Both knobs compose: a page is excluded if it matches EITHER the prefix filter OR the type filter. Both apply to the orphan list AND the F6 denominator so `total_linkable` matches the candidate set.

### Knob 1: `orphans.exclude_prefixes`

1. `shouldExclude(slug, userPrefixes = [])` gains an optional second parameter. Same `startsWith` semantics as `DENY_PREFIXES`. Runs after the shipped list, before the first-segment exclusions.
2. `findOrphans` reads the config once via `readUserExcludePrefixes(engine)` and threads it into both the orphan-filter pass AND the F6 denominator pass.
3. `parseUserExcludePrefixes(raw: unknown)` is the pure-function parser. Tolerates whitespace, drops empties, returns `[]` for any non-string input.

### Knob 2: `orphans.exclude_types`

1. `DEFAULT_EXCLUDE_TYPES` (exported readonly list) carries the shipped types.
2. `readExcludeTypes(engine)` merges `DEFAULT_EXCLUDE_TYPES` with user config (deduplicated union). Returns the merged set, never empty.
3. `findOrphanPages` (engine method) gains optional `excludeTypes?: string[]`. SQL filter `AND p.type <> ALL($N::text[])` applied AFTER the existing inclusion filter. Both engines updated in lockstep per the engine-parity invariant.
4. `findOrphans` threads the merged list to `engine.findOrphanPages({ excludeTypes })` AND the live-rows denominator query.

### Shared

- `USER_EXCLUDE_PREFIXES_CONFIG_KEY` and `EXCLUDE_TYPES_CONFIG_KEY` exported as the single source of truth.
- `gbrain orphans --help` documents both knobs inline.
- `src/core/config.ts` registers both keys in `KNOWN_CONFIG_KEYS`.

## Behavior change to call out

`DEFAULT_EXCLUDE_TYPES` shifts orphan counts downward for every brain that has any of those six types. The `orphan_ratio` doctor check stops counting code-symbol and commit-summary pages. This is a one-time correction, not a regression: a code-symbol extraction never had inbound-link semantics; the previous count was lying about it. Brains without those types see no change.

## Backward compatibility

- Existing call sites (CLI, doctor's `orphan_ratio` check, every test fixture, `getOrphansData` consumers) pass no second argument or no excludeTypes. `shouldExclude` and `findOrphanPages` keep their prior shape for callers that don't opt in.
- Unset prefix config returns `[]`. Unset type config returns the defaults — that's the change.

## Tests

- `bun test test/orphans.test.ts`: 12 new cases on Knob 1's pure-function surface (`parseUserExcludePrefixes` + `shouldExclude` with userPrefixes). Plus 22 existing cases unchanged.
- `bun test test/orphans-user-exclude-prefixes.test.ts`: 22 PGLite-backed integration cases total. 8 for Knob 1. 14 for Knob 2 (parser, readExcludeTypes-merges-with-defaults including dedup, end-to-end findOrphans covering default-type drops, user-extension, F6 denominator symmetry, composition with Knob 1, source-scoped).
- `bun test test/orphans-pure-fn.test.ts test/orphans-source-scope.test.ts`: 22 existing cases unchanged.
- `bun test test/config-set.test.ts`: regression guards on both new keys.
- 115 tests pass across the orphans + config-set suites. `bun run typecheck` clean.

## Smoke test on a real brain

```bash
# Baseline on a 2K-page PARA vault + 11 federated code-stage source repos:
gbrain orphans --count
# 2906

gbrain config set orphans.exclude_prefixes 'resources/,transcripts/,agents/,clips/'
gbrain orphans --count
# 1418 (dropped 1488 PARA convention pages)

# Then DEFAULT_EXCLUDE_TYPES fires automatically (no operator action):
gbrain orphans --count
# 928 (dropped 490 code + system + commit-summary pages)

# Operator extends for any local types they don't want surfaced:
gbrain config set orphans.exclude_types 'guide,reference'
gbrain orphans --count
# ~920
```

928 actual signal vs 2906 raw. The dream cycle's `orphan_ratio` health check tracks the same number, so the brain-score stops being dominated by reference material and structurally-orphan page types the moment the patch lands.

## Notes

- **Glob support** (e.g. `**/templates/**` for nested deny patterns) is out of scope. The existing `DENY_PREFIXES` uses `startsWith`; adding globs would need a wider rework.
- **Frontmatter-level `orphan_expected: true` per-page opt-out** is also out of scope. Complementary to the dir-level + page-type surfaces, useful for one-off pages. Worth a follow-up.
- The first iteration of this PR tried `orphans.skip_code_sources` keying on `sources.config.strategy === 'code'`. That field is almost never populated in practice (gstack-memory-ingest writes `federated: true` + runtime fields only), so the toggle was a no-op on real brains. Reverted in `0a6bf3ad` and replaced with the page-type knob, which reliably fires because `pages.type` is set by the schema-pack inference + frontmatter pipeline on every page.


time-attack:
This is a thoughtful and well-tested PR, and the 2906-vs-928 orphan overcount you measured is a real annoyance. We're still closing it: two new per-brain config keys plus engine-interface changes is permanent configuration surface added to tune a reporting heuristic, and we're holding the line hard on new knobs. If shipped DENY_PREFIXES is genuinely missing entries that are noise-by-construction on every brain shape (not just PARA/Obsidian layouts), a few-line addition to that hardcoded list is something we'd consider.

PR #2218, opened 2026-06-16
fix(provider): honor provider_base_urls for native OpenAI and providers test

hazeone:
## Summary

- **Native OpenAI gateway path** (`chat`, `expansion`, `embed`) now passes `provider_base_urls.openai` into the AI SDK client, so custom compatible-mode proxies (e.g. third-party OpenAI gateways) work instead of always calling `api.openai.com`.
- **`gbrain providers test --model`** now keeps file-plane `provider_base_urls` and `embedding_dimensions` from `~/.gbrain/config.json` when overriding the model, so isolated probes hit the same endpoint as the configured brain.

## Problem

Users configuring:

```json
"provider_base_urls": { "openai": "https://example.com/v1" },
"chat_model": "openai:gpt-5.4"
```

saw chat/expansion fail with OpenAI official API key errors because `createOpenAI({ apiKey })` ignored the custom base URL. DashScope embedding worked (openai-compatible recipe) but OpenAI chat did not (native-openai recipe).

Separately, `gbrain providers test --model dashscope:...` reconfigured the gateway without `base_urls`, silently falling back to recipe defaults (e.g. dashscope-intl).

## Test plan

- [x] `bun run typecheck`
- [x] `bun test test/providers.test.ts test/ai/build-gateway-config.test.ts`
- [x] Manual: set `provider_base_urls.openai` to a compatible proxy, run `gbrain providers test --touchpoint chat`
- [x] Manual: set custom `provider_base_urls.dashscope`, run `gbrain providers test --model dashscope:text-embedding-v4`

### DIFF
--- a/src/commands/providers.ts
+++ b/src/commands/providers.ts
@@ -180,16 +180,29 @@ async function runTest(args: string[]): Promise<void> {
       }
     } catch { /* loadConfig throws when no brain configured — first-time install path; the no-config branch above handles it. */ }
 
+    const cfg = loadConfig();
     if (tpArg === 'embedding') {
-      const dims = recipe?.touchpoints.embedding?.default_dims ?? 1536;
+      const dims =
+        cfg?.embedding_dimensions ??
+        recipe?.touchpoints.embedding?.default_dims ??
+        1536;
       configureGateway({
         embedding_model: modelArg,
         embedding_dimensions: dims,
+        expansion_model: cfg?.expansion_model,
+        chat_model: cfg?.chat_model,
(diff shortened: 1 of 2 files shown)


time-attack:
Good catch and an accurate diagnosis — `createOpenAI({ apiKey })` with no baseURL really did strand custom OpenAI-compatible proxies. Master has since fixed exactly this: a25209bb (v0.42.58.0, #2627) added `resolveNativeBaseUrl` and threads `provider_base_urls` into every native OpenAI client, and `gbrain providers test` now preserves base_urls and embedding_dimensions when overriding the model. Closing as already fixed — thanks for reporting it.

PR #2220, opened 2026-06-16
feat(ai): MiniMax full-stack — embeddings compat shim, asymmetric type, chat touchpoint

tcflying:
## Problem

The `minimax` recipe declared only an `embedding` touchpoint and was tagged
`openai-compatible`, but MiniMax's `/v1/embeddings` endpoint is **not** OpenAI-shaped
at the wire level — it requires `{model, texts, type}` and returns
`{vectors, base_resp}`. With no fetch shim, the AI SDK sent `{input}` and MiniMax
replied `status_code 2013 "missing required parameter texts"`, surfacing as
`[embed(minimax:embo-01)] Invalid JSON response` and `Embedded 0 chunks`. MiniMax was
effectively unusable for embeddings, and there was no chat touchpoint at all.

## Fix (3 commits)

1. **`gateway.ts` — `minimaxCompatFetch`** (mirrors `voyageCompatFetch` /
   `zeroEntropyCompatFetch`): OUTBOUND maps `input → texts`, derives `type` from the
   threaded `input_type`, strips OpenAI-only keys; INBOUND rewrites
   `{vectors} → {data:[{embedding,index}]}` and surfaces a non-zero
   `base_resp.status_code` as a thrown error (MiniMax returns HTTP 200 on logical
   errors). Wired into `instantiateEmbedding`'s `fetchWrapper` dispatch for
   `recipe.id === 'minimax'`.
2. **`dims.ts` — asymmetric retrieval.** embo-01 now emits `input_type` (same seam as
   ZE/Voyage) instead of a hardcoded `type:'db'`, so `embedQuery()` threads `'query'`
   and `embed()` threads `'document'`; the shim maps that to MiniMax's `type: db|query`
   wire field.
3. **`recipes/minimax.ts` — chat touchpoint.** MiniMax exposes an OpenAI-compatible
   `/v1/chat/completions` endpoint (same key); M3 function-calls and returns
   OpenAI-shaped responses (no shim needed). Adds `MiniMax-M3` / `MiniMax-M2.7`
   (`supports_tools`, 800K ctx). With this plus the embedding fix, gbrain runs
   end-to-end — embed + search + graph + chat synthesis — on MiniMax alone, no
   OpenAI/Anthropic key required.

## Testing

- Verified end-to-end against the live API with a coding-plan key: `embo-01` embeds
  (1536d), asymmetric `db`/`query` both succeed, `gbrain think` synthesizes from a
  brain page with a citation (`Model: minimax:MiniMax-M3`).
- `test/ai/recipe-minimax.test.ts` updated for the new `input_type` behavior (document
  default + query path) and a new chat-touchpoint assertion.
- Green locally: `recipe-minimax`, `recipes-existing-regression`, `gateway`,
  `gateway-chat`, `dims-openai`, `dims-zeroentropy` (89+ tests, 0 fail);
  `bun run typecheck` clean.


tcflying:
Added one more commit (`fix(cli): windowsHide on the startup check-update spawn`): the detached `check-update --refresh-cache` refresh had no `windowsHide`, so on Windows it flashed a console window / launch-error dialog. Same spawn-hygiene class. Happy to split into its own PR if you'd prefer it separate from the MiniMax work.

tcflying:
This branch now carries three related fixes, verified locally on both PGLite and Postgres 17:

**1. MiniMax full-stack (original commits)** — embeddings compat shim (`input`->`texts`, `vectors`->`data[]`), asymmetric `type: db|query`, and a MiniMax-M3 chat touchpoint. Runs embeddings + think/dream/synthesis on MiniMax with no OpenAI/Anthropic key.

**2. `fix(cycle)`: maintenance flag so serve defers during dream.** PGLite is single-writer. A concurrent `gbrain serve` (MCP) contends the lock and the long `propose_takes` phase could hit a native PGLite abort (silent exit). `gbrain dream` now claims a maintenance window (`~/.gbrain/.maintenance`, TTL backstop); `serve` checks it at startup + on a poll and gracefully defers (releases lock + exits 0). Adds `gbrain dream --exclude-takes` so a wrapper can run the takes phases as isolated subprocesses.

**3. `fix(migrate)`: copy source rows before pages.** `gbrain migrate` copied pages but never the `sources` table, so a multi-source brain aborted with `pages_source_id_fkey` on the first non-default-source page (target `initSchema` seeds only `default`). Now copies non-default sources (`INSERT ... ON CONFLICT DO NOTHING`) before the page loop. Verified: a 5-source / 4288-page / ~30k-chunk brain migrated PGLite->Postgres cleanly.

Happy to split into separate PRs if you prefer.

time-attack:
Reviewed and declined for now: paid-cloud vendor catalog growth beyond the majors is paused (each recipe becomes a maintenance commitment we can't test). The LiteLLM/openai-compat generic path reaches MiniMax today. Thank you for the substantial work.

Issue #2221, opened 2026-06-16
Improve facts extractor reliability for typed metrics, multi-claim turns, and local models


sene1337:
## Suggestion

Improve the facts extractor as one cohesive quality pass, rather than treating this as separate changes for metrics vs. local models.

The main opportunity is to make extractor output more schema-consistent and robust across both frontier and local models, especially for:

1. Typed metric claims that are already emitted with `metric/value/unit/period` fields.
2. Multi-claim user turns that should split into multiple atomic facts.
3. Low-value/routine messages that should be skipped even when they contain numbers.
4. Local-model pipelines, where a strict second-pass judge/normalizer can help but should not lower frontier-model quality.

## Why this came up

I ran a no-insert benchmark against the current facts extractor behavior using 70 human-curated, Telegram/GBrain-style fixture messages. The fixtures were actual-style/anonymized test cases, not private conversation exports.

Environment under test:

```text
GBrain package version: 0.42.38.0
Repo commit tested:     03ffc6eb
Default facts model:    anthropic:claude-sonnet-4-6
Local extractor model:  Gemma 26B via an OpenAI-compatible local endpoint
Local judge model:      atla/selene-mini via Ollama
```

The current extractor prompt already asks models to emit typed metric fields:

```text
metric, value, unit, period
```

But when models correctly emit those typed fields, they sometimes still classify achieved/current metric claims as `event` instead of `fact`.

## Benchmark summary

```text
Pipeline                            Score      Notes
----------------------------------  ---------  -----------------------------------------------
Sonnet raw                          59/70      Current default facts model behavior
Sonnet + metric normalizer          63/70      Metric positives fixed: 6/10 -> 10/10
Local Gemma raw                     56/70      Local extractor, no judge
Local Gemma + metric normalizer     59/70      Metric positives fixed: 7/10 -> 10/10
Gemma + Selene judge                59/70      Fixed commitments, but regressed one static fact
Gemma + Selene + metric normalizer  62/70      Best local pipeline in this run
```

The most reliable improvement was a narrow deterministic post-processing rule:

```text
If the extractor emits a typed metric claim for an achieved/current/historical metric,
normalize kind to `fact` unless the source/fact text is a question, goal, target,
forecast, hope, or other future/uncertain statement.
```

This improved both the frontier baseline and local-model results.

### 1. Achieved/current metrics should normalize to `fact`

These are objective metric facts, not events, if the model already emitted typed metric fields:

```text
[...]

PR #2233, opened 2026-06-16
fix(facts): harden extraction gating and provider config

sene1337:
## Summary

This PR hardens facts extraction around the failure modes discussed in https://github.com/garrytan/gbrain/issues/2221#issuecomment-4719397290.

It adds deterministic guardrails around the existing LLM extractor rather than changing the default model:

- skip obvious transient task/question/ack/exact-reply smoke-test turns before calling chat
- probe chat availability using the effective facts extraction model instead of the global chat model
- normalize high-signal extracted kinds for preferences, commitments, beliefs, and typed metric claims
- avoid smearing a decision marker across unrelated facts in multi-claim turns
- merge DB-plane `provider_base_urls.<provider>` entries into runtime config so configured OpenAI-compatible/private endpoints actually reach the gateway

## Why

The issue thread documents extractor-quality gaps around local/private model usage, transient operational chatter, and provider routing. The narrow bug fixed here is that facts extraction could either call the model on obvious non-memory turns or silently no-op when the configured facts model differed from the global chat model.

## Tests

- `git diff --check`
- `bun test test/facts-extract-smoke.test.ts test/facts-extract-silent-no-op.test.ts test/loadConfig-merge.test.ts`
  - `37 pass / 0 fail / 81 expect() calls`

Also ran `bun run typecheck`; it still fails on existing OpenClaw SDK typing issues unrelated to this patch:

```text
src/core/context-engine.ts(90,44): Property 'buildMemorySystemPromptAddition' does not exist on type 'typeof import(".../openclaw/dist/plugin-sdk/core")'.
test/e2e/openclaw-plugin-load-real.test.ts(281,9): Type 'unknown' is not assignable to type 'ContextEngine | Promise<ContextEngine>'.
```

## Notes

No model defaults are changed by this PR. It only makes extraction safer and makes existing provider/base-url config behave as advertised.

### DIFF
--- a/src/core/ai/build-gateway-config.ts
+++ b/src/core/ai/build-gateway-config.ts
@@ -21,13 +21,15 @@ import type { AIGatewayConfig } from './types.ts';
 
 export function buildGatewayConfig(c: GBrainConfig): AIGatewayConfig {
   // v0.32 (#121 reworked): when ~/.gbrain/config.json declares
-  // openai_api_key / anthropic_api_key, fold them into the gateway env so
-  // recipes that read OPENAI_API_KEY / ANTHROPIC_API_KEY find them. Process
+  // openai_api_key / anthropic_api_key / openrouter_api_key, fold them into
+  // the gateway env so recipes that read OPENAI_API_KEY / ANTHROPIC_API_KEY /
+  // OPENROUTER_API_KEY find them. Process
   // env still wins (it's loaded last) — this is a fallback for daemons /
   // launchd-spawned subprocesses that don't propagate ~/.zshrc-sourced keys.
   const envFromConfig: Record<string, string> = {};
   if (c.openai_api_key) envFromConfig.OPENAI_API_KEY = c.openai_api_key;
   if (c.anthropic_api_key) envFromConfig.ANTHROPIC_API_KEY = c.anthropic_api_key;
+  if (c.openrouter_api_key) envFromConfig.OPENROUTER_API_KEY = c.openrouter_api_key;
   // v0.37 fix wave (CDX2-5+6): ZE became the default provider in v0.36 but
   // the env-mapping at this seam never picked it up. `gbrain config set
   // zeroentropy_api_key X` wrote DB plane (ignored by gateway). The file-
--- a/src/core/config.ts
+++ b/src/core/config.ts
@@ -601,7 +601,10 @@ export function loadConfig(): GBrainConfig | null {
  * size the schema and must be stable across engine connect.
  */
 export async function loadConfigWithEngine(
-  engine: { getConfig(key: string): Promise<string | null | undefined> },
+  engine: {
+    getConfig(key: string): Promise<string | null | undefined>;
+    listConfigKeys?(prefix?: string): Promise<string[]>;
+  },
   base?: GBrainConfig | null,
 ): Promise<GBrainConfig | null> {
   // Codex /ship finding #3: when there's no file config AND no env DB URL,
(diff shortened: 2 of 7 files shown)


darbsllim:
the clanker found some bugs and fixed them but didn't mention the bug fixes in here, hold up let me get him to add a comment with the bug fixes.

sene1337:
Adding more concrete bug detail for reviewers.

This patch fixes several bugs/footguns I hit while testing facts extraction against real conversation-style turns and private/OpenAI-compatible provider routing.

### 1. `facts.extraction_model` could be configured but extraction still silently no-oped

Before this PR, `extractFactsFromTurn()` checked:

```ts
isAvailable('chat')
```

before resolving the effective facts extraction model. That means availability was tested against the global/default chat model, not the model actually used for facts extraction.

Failure mode:

- global/default chat model points at provider A
- `facts.extraction_model` points at provider B
- provider A is not configured/available
- provider B is configured/available
- extraction returns `[]` before it ever calls chat

This is especially visible for local/private/OpenAI-compatible routing, where a user may want facts extraction to use one local/private model while leaving the global chat model alone.

Fix in this PR:

- resolve `const model = input.model ?? await getFactsExtractionModel(input.engine)` first
- test `isAvailable('chat', model)` against that effective model
- pass the same effective model into `chat()`

### 2. `provider_base_urls.<provider>` was accepted as config but not merged into runtime gateway config

`gbrain config set provider_base_urls.openrouter http://...` is allowed by the known config-key prefix, but `loadConfigWithEngine()` was not reading/merging those DB-plane prefixed keys into `provider_base_urls`.

Failure mode:

- user configures an OpenAI-compatible local/private endpoint via `provider_base_urls.openrouter`
- config write succeeds
- gateway config never sees that base URL after engine merge
- facts extraction/provider tests can still route to the default provider URL or appear unavailable

Fix in this PR:

- if the engine supports `listConfigKeys`, load keys under `provider_base_urls.`
- merge them into `merged.provider_base_urls`
- keep file-plane config precedence over DB-plane values

### 3. Obvious non-memory turns were still sent to the LLM extractor

The prompt tells the model to skip questions/operational chatter, but in practice this is exactly the class of thing that benefits from deterministic filtering before the model call.

Failure mode:

- exact-reply probes, smoke tests, short acks, and ordinary task requests go to the LLM
- weaker/local models sometimes convert them into durable facts/commitments
- unnecessary model calls add latency/cost and create hot-memory noise

Fix in this PR:

- add a conservative `shouldSkipTurnForFactsExtraction()` pre-filter
- skip pure acks, exact-reply/smoke-test probes, wrappers, and ordinary task/question turns
- preserve turns with durable preference/rule markers such as `I prefer`, `make sure`, `always`, `never`, `source of truth`, etc.

### 4. Kind classification was too model-dependent for common high-signal first-person markers

In real conversation turns, the model may extract the right text but assign a weak/wrong kind, especially for local/private models.

Examples:

- `I prefer ...` should be `preference`
- `make sure ...`, `we decided ...`, `always/never/do not ...` should generally be `commitment`
- `I think/I believe ...` should be `belief`
- typed metric assertions should remain `fact`

Fix in this PR:

- add `normalizeExtractedFactKind()` as a deterministic post-guard around model output
- force typed metric claims to `fact`
- normalize obvious preference/commitment/belief markers

### 5. Multi-claim turns could smear one marker across unrelated extracted facts

A turn can contain both a preference and a decision, e.g.:

```text
I prefer concise extraction candidates. I decided to use a private model for facts extraction.
```

A naive whole-turn normalizer would see `decided` and mark both facts as `commitment`.

Fix in this PR:

- prefer evidence in the extracted fact text itself
- only fall back to whole-turn evidence when the turn has a single unambiguous marker class
- added a regression test so the preference stays `preference` and the decision stays `commitment`

### Test coverage added

The PR adds targeted coverage for:

- transient pre-filter skips before chat is called
- durable style/rule feedback is not skipped
- high-signal kind normalization
- multi-claim kind-smear prevention
- facts-model availability independent of global chat model
- DB-plane `provider_base_urls.<provider>` runtime merge
- file-plane provider base URL precedence

Local verification before opening the PR:

```text
git diff --check
bun test test/facts-extract-smoke.test.ts test/facts-extract-silent-no-op.test.ts test/loadConfig-merge.test.ts

37 pass
0 fail
81 expect() calls
```

`bun run typecheck` still fails on existing OpenClaw SDK type errors unrelated to this patch; I noted those in the PR body.

sene1337:
Update from durable local/private proxy setup:

While wiring this into a LaunchAgent-backed local OpenAI-compatible proxy, I found one more config bug and pushed a second commit (`d034be78`).

### Additional bug fixed: OpenRouter had no file-plane API key mapping

`buildGatewayConfig()` already maps these file-plane config fields into gateway env for daemon/launchd paths:

- `openai_api_key` → `OPENAI_API_KEY`
- `anthropic_api_key` → `ANTHROPIC_API_KEY`
- `zeroentropy_api_key` → `ZEROENTROPY_API_KEY`

But OpenRouter did not have the same path. That meant a daemon or service could declare `provider_base_urls.openrouter` and `facts.extraction_model=openrouter:...`, but still require `OPENROUTER_API_KEY` in the process environment. In a noninteractive service context, the extractor could still silently return no facts even though the base URL/model config was correct.

The new commit adds:

- `openrouter_api_key?: string` to `GBrainConfig`
- env override import from `OPENROUTER_API_KEY`
- `buildGatewayConfig()` mapping to `OPENROUTER_API_KEY`
- `openrouter_api_key` to known config keys
- a unit test that verifies `openrouter_api_key` + `provider_base_urls.openrouter` reach gateway config together

This keeps behavior parallel with the existing provider key fields and makes OpenRouter-compatible local/private proxy use viable from daemon contexts without depending on shell startup files.

time-attack:
Thanks for engaging with the failure modes in the #2221 thread. Two reasons we're closing: the provider-routing half is already on master (config.ts merges the DB-plane `provider_base_urls.*` prefix map, and the native-gateway base-URL path landed in a25209bb / #2627), and the remaining half is a set of hardcoded heuristics for classifying 'transient' turns plus kind normalization — tunable extractor policy rather than a bug fix, and the kind of surface that quietly starts dropping real memories. Probing chat availability against the effective facts model rather than the global chat model is the one narrow idea worth filing as its own small issue.

Issue #1340, opened 2026-05-24
PGLite WASM initialization fails with ENOENT: //root/pglite.data on macOS 12.7.6 + Bun 1.3.14


788s4jnmgr-cloud:
## Bug Description

PGLite-based gbrain commands crash immediately with:
```
PGLite failed to initialize its WASM runtime.
  This is most commonly the macOS 26.3 WASM bug: https://github.com/garrytan/gbrain/issues/223
  Run \`gbrain doctor\` for a full diagnosis.
  Original error: ENOENT: no such file or directory, open '/$$bunfs/root/pglite.data'.
```

Unlike issue #223 (which is macOS 26.3 specific), this occurs on **macOS 12.7.6 (Darwin 21H1320)**.

## Steps to Reproduce

1. \`gbrain sync --skip-failed\`
2. \`gbrain sources list\`
3. \`gbrain apply-migrations --yes\`
4. Any command that triggers \`engine.connect()\`

All crash with the same PGLite WASM initialization error.

Commands that work: \`gbrain doctor --fast\` (skips DB connection check).

## Environment

- macOS: 12.7.6 (Darwin 21H1320, macOS Monterey)
- Bun: 1.3.14
- Node: v24.14.1
- gbrain: 0.40.8.0 (source build) and 0.40.2.0 (official binary)
- PGLite: 0.4.3
- Architecture: arm64 (Apple Silicon)

## Analysis

The error path \`/$$bunfs/root/pglite.data\` is a Bun virtual filesystem mount point. On this macOS version, \`/$$bunfs/root\` is a read-only directory:

```bash
$ ls / | grep bunfs
# → nothing (directory does not exist or is empty/mounted as read-only)

$ mkdir -p /$$bunfs/root
# → mkdir: /$$bunfs: Read-only file system
```

This means PGLite cannot write its \`pglite.data\` file during \`PGlite.create()\`, causing immediate Abort.

\`gbrain doctor\` works because the \`--fast\` flag skips the DB connection phase entirely.

## Expected Behavior

PGLite should initialize successfully, or gbrain should detect this path issue and fall back gracefully.

## Additional Context

- Issue #223 mentions macOS 26.3, but this machine is on macOS 12.7.6
- Tested with both official gbrain binary (0.40.2.0) and source-built binary (0.40.8.0) — both fail
- Tested with \`bun run src/cli.ts\` (direct source) — also fails
- \`/$$bunfs/root\` being read-only appears to be the root cause

[...]

PR #2246, opened 2026-06-17
Diagnose & recover unclean PGLite data dirs (refs #223)

tlconde:
## Summary

Addresses recurring `Aborted()` crashes from PGLite (#223) that are usually unclean shutdowns / WAL recovery candidates rather than the originally-suspected macOS 26.3 WASM bug (per WilliamCourterWelch's diagnosis in the thread).

Two commits, complementary:

- **e2fc66c5** — adds a conservative `pglite-data-dir-recovery-candidate` verdict in `src/core/pglite-engine.ts`. Takes a preflight filesystem snapshot before `PGlite.create()` so post-failure classification doesn't conflate startup residue with pre-existing crash residue. Wrapped error now pastes the recovery command.
- **4861e00d** — adds opt-in `gbrain doctor --rebuild-pglite`, routed in `src/cli.ts` **before** the normal doctor engine connect (mirrors `reinit-pglite`). Extracts a lock-safe preserve/reinit helper that both commands share. Adds a sibling maintenance lock to prevent rename/reinit races. Refuses on thin-client config. Real subprocess smoke test for the rebuild JSON behavior.

Per the maintainer's request in the issue thread for a `gbrain doctor --rebuild-on-pglite-wasm-abort` recovery path and clearer error wrap in `pglite-engine.ts connect()`.

## Test plan

- [x] `git diff --check` — clean
- [x] `bun test test/pglite-init-classifier.test.ts test/pglite-lock.test.ts test/doctor-rebuild-pglite.serial.test.ts` — green
- [x] `bun test test/v0_37_gap_fill.serial.test.ts test/fix-wave-structural.test.ts` — green (broader smoke, no regressions)
- [x] Manual repro: hit `Aborted()` on a corrupted data dir → wrapped error now names the rebuild command → `gbrain doctor --rebuild-pglite --yes` recovers via put+search round-trip

## Design notes

- Verdict deliberately named "recovery-candidate" + uses "may be" wording rather than overclaiming WAL corruption — `postmaster.pid` alone, or non-empty `pg_wal/`, are not reliable corruption signals.
- Rebuild is opt-in only; no prompts from engine code (would hang `gbrain serve` + other noninteractive callers).
- Lock released **before** the data-dir rename so the in-dir `.gbrain-lock` doesn't get carried into the `.broken-<ts>` backup as a stale lock.
- Consistent with upstream electric-sql/pglite#994 declining automatic library-level repair — recovery is an explicit user-driven tool here.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -1217,6 +1217,16 @@ async function handleCliOnly(command: string, args: string[]) {
     return;
   }
   if (command === 'doctor') {
+    if (args.includes('--rebuild-pglite')) {
+      const cfgForRebuild = loadConfig();
+      if (isThinClient(cfgForRebuild)) {
+        refuseThinClient('doctor --rebuild-pglite', cfgForRebuild!.remote_mcp!.mcp_url);
+      }
+      const { runDoctorRebuildPglite } = await import('./commands/doctor.ts');
+      await runDoctorRebuildPglite(args);
+      return;
+    }
+
     // Multi-topology v1: thin-client doctor. When `~/.gbrain/config.json`
     // has remote_mcp set, every DB-bound check is irrelevant. Route to the
     // outbound-HTTP probe set in `src/core/doctor-remote.ts` and return
@@ -2183,6 +2193,7 @@ SETUP
   upgrade                            Self-update
   check-update [--json]              Check for new versions
   doctor [--json] [--fast]            Health check (resolver, skills, pgvector, RLS, embeddings)
+  doctor --rebuild-pglite [--yes]     Move aside a broken PGLite data dir and re-init
   integrations [subcommand]          Manage integration recipes (senses + reflexes)
 
 PAGES
--- a/src/commands/doctor.ts
+++ b/src/commands/doctor.ts
@@ -7232,6 +7232,99 @@ export async function runDoctor(
   setCliExitVerdict(hasFail ? 1 : 0);
 }
 
+export async function runDoctorRebuildPglite(args: string[]): Promise<void> {
+  const jsonOutput = args.includes('--json');
+  const yes = args.includes('--yes') || args.includes('-y');
+  const noSync = args.includes('--no-sync');
+  const cfg = loadConfig();
+
+  if (cfg?.engine !== 'pglite') {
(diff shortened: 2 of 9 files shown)


tlconde:
**Manual repro verified end-to-end** ✅

Reproduced on an actual corrupted PGLite data dir (a real `brain.pglite.broken-*` from this user's history of hitting #223):

```text
$ GBRAIN_HOME=/tmp/test-broken-home bun run src/cli.ts recall "test"
PGLite failed to initialize its WASM runtime.
  The existing PGLite data directory may need recovery/rebuild.
  Fix: run `gbrain doctor --rebuild-pglite` and retry.
  Original error: Aborted(). Build with -sASSERTIONS for more info.

$ GBRAIN_HOME=/tmp/test-broken-home bun run src/cli.ts doctor --rebuild-pglite --yes
...
PGLite rebuild complete.
  Fresh brain: /tmp/test-broken-home/.gbrain/brain.pglite
  Preserved old data dir: /tmp/test-broken-home/.gbrain/brain.pglite.broken-20260617T121940Z
  Sync skipped or did not complete. Run `gbrain sync` to repopulate from sources.

$ GBRAIN_HOME=/tmp/test-broken-home bun run src/cli.ts recall "anything"
No matching facts.    # ← fresh brain works, no Aborted()
```

Also ran the full CONTRIBUTING-specified pre-push gate:
- `bun run verify` — clean (exit 0)
- `bash scripts/check-test-isolation.sh` — `OK (929 non-serial unit files scanned)` — new non-serial test files (`pglite-init-classifier.test.ts`, `pglite-lock.test.ts`) comply with R1–R4.

Last unchecked test-plan box in the PR description is now verified.

time-attack:
This is careful work — the preflight filesystem snapshot to avoid conflating startup residue with pre-existing crash residue is a genuinely sharp detail, and the 'recovery-candidate' hedging is the right instinct. We're still closing it: master already carries `classifyPgliteInitError` (v0.41.8.0, #1340) and a `reinit-pglite` command, so this adds a second parallel recovery path plus a new maintenance lock across 634 lines. The piece with the best ratio would be a standalone few-line change that makes the wrapped `Aborted()` error name the existing `gbrain reinit-pglite` recovery command — that we'd read.

PR #2248, opened 2026-06-17
feat(auth): federated_write — per-write source authorization (write-side mirror of federated_read)

pablofavre12:
## What
Adds `federated_write TEXT[]` to `oauth_clients`: the set of sources a token may write to, beyond its single `source_id`. Write ops choose a target source per call, authorized against that set — so one token (e.g. a "directorio" curator) can write to campo / lideres / directorio without re-minting tokens.

## Mirror of federated_read
This is the WRITE-side mirror of `federated_read`, implemented at every site federated_read touches:
- column + GIN index in all 3 schema sources (`schema.sql`, `pglite-schema.ts`, `schema-embedded.ts`) + pre-bootstrap on both engines
- idempotent migrations (v118 column, v119 GIN index), no backfill
- `--federated-write SRC1,SRC2,…` CLI flag on `register-client`
- oauth-provider persist + SELECT/normalize, with the same pre-version back-compat fallback shape
- `AuthInfo.federatedWrite` (mirror of `allowedSources`)

## The one non-mirror part: per-write authorization
New `resolveWriteSource(ctx, requested?)` helper, fail-closed like the read-side scope resolution:
- no `requested` → `ctx.sourceId` (UNCHANGED — pre-v118 clients never sent one)
- `requested` ∈ `{sourceId} ∪ federatedWrite` → allowed
- otherwise → `permission_denied` (`source 'X' not in this client's federated_write set`)

Empty/undefined `federated_write` collapses the allowed set to just `source_id`, so omitting the grant exactly preserves the single-source write lock. Server-stamped provenance (`source_kind`/`source_uri`/`ingested_via`) is untouched — the new `source` is a separate, authorized write-target field.

`put_page` gets an optional `source`; `put_raw_data` gets `write_source` (its existing `source` param is the data-provenance label and was left intact).

## Tests
New `test/put-page-federated-write.test.ts` covers: no `source` → `source_id`; `source` in set → allowed; `source` not in set → rejected; attacker-supplied empty `[]` does not widen scope; `put_raw_data` `write_source` vs data-`source` distinction. `tsc --noEmit` clean; existing oauth/source-isolation guard checks pass.

## Compatibility
Default `'{}'` (empty) for every existing client → writes stay locked to `source_id`. No backfill. Pre-v118 brains fall back to the federated_read projection until `apply-migrations` runs.

## Usage
```
gbrain auth register-client director --scopes read,write --source directorio \
  --federated-read campo,lideres,directorio --federated-write campo,lideres,directorio
```

### DIFF
--- a/src/commands/auth.ts
+++ b/src/commands/auth.ts
@@ -344,6 +344,7 @@ interface RegisterClientArgs {
   scopes: string;
   sourceId: string;
   federatedRead: string[] | undefined;
+  federatedWrite: string[] | undefined;
   redirectUris: string[];
   tokenEndpointAuthMethod: string | undefined;
 }
@@ -354,6 +355,7 @@ export function parseRegisterClientArgs(args: string[]): RegisterClientArgs {
     scopes: 'read',
     sourceId: 'default',
     federatedRead: undefined,
+    federatedWrite: undefined,
     redirectUris: [],
     tokenEndpointAuthMethod: undefined,
   };
@@ -383,6 +385,11 @@ export function parseRegisterClientArgs(args: string[]): RegisterClientArgs {
         out.federatedRead = v.split(',').map(s => s.trim()).filter(Boolean);
         i += 2; break;
       }
+      case '--federated-write': {
+        const v = requireValue();
+        out.federatedWrite = v.split(',').map(s => s.trim()).filter(Boolean);
+        i += 2; break;
+      }
       case '--redirect-uri':
         out.redirectUris.push(requireValue());
         i += 2; break;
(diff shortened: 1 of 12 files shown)


time-attack:
This is careful work — the fail-closed default, the empty-set-collapses-to-source_id behavior, and mirroring federated_read at every site it touches are all the right instincts. But it is new authorization surface rather than a fix for something broken: single-source write locking behaves as designed today, and this widens the write-side trust boundary permanently. We're holding the line on new feature surface, especially in auth, so I'm closing this. Thanks for the thorough write-up and tests.

PR #2257, opened 2026-06-17
fix(subagent): reliable gateway tool-loop for non-Anthropic providers (persist tool-result messages + JSON-normalize outputs)

### LINKED ISSUE #2256: Gateway subagent loop unreliable for multi-turn tool calls on non-Anthropic providers (resume loop + Date-in-output rejection)
## Summary

The gateway-native subagent loop (`agent.use_gateway_loop=true`, mandatory for non-Anthropic providers) is unreliable for any multi-turn tool conversation. Two independent bugs combine so that subagent jobs using a local/OpenAI-compatible model (e.g. a `llama-server` qwen) loop on retries and never complete:

1. **Tool-result messages are never persisted** → every *resume* rebuilds an invalid history.
2. **Tool outputs carrying non-JSON values (e.g. `Date`) are rejected by AI SDK v6** → fails *live*, mid-run.

Both surface as the same family of AI SDK v6 errors:
- `Invalid prompt: The messages do not match the ModelMessage[] schema.`
- `[chat(<provider:model>)] Tool results are missing for tool calls <ids>`

The Anthropic-direct path is unaffected, so this only bites users who enable the gateway loop for non-Anthropic models.

## Bug 1 — gateway loop never persists the tool-result message

`toolLoop()` (`src/core/ai/gateway.ts`) feeds each turn's tool results back as a `role:'user'` message:

```js
const userMessageIdx = messageIdx++;
void userMessageIdx;                                  // <-- dropped
messages.push({ role: 'user', content: toolResultBlocks });
```

It persists assistant turns (`onAssistantTurn`) and per-tool execution rows (`onToolCall*`), but there is **no callback to persist this user/tool-result message**, and `runSubagentViaGateway` discards `ToolLoopResult.messages`. The legacy Anthropic-direct path *does* persist it (`subagent.ts`, the `userIdx`/`persistMessage(role:'user', content_blocks: toolResults)` write).

Consequence: `subagent_messages` ends up with only odd `message_idx` (assistant turns at 1, 3, 5…) and gaps where the tool-result messages (2, 4…) should be. Within one uninterrupted worker run this is invisible (the messages live in the in-memory array). But on **any resume** (worker restart / re-claim / timeout), `loadPriorMessages` rebuilds the conversation from `subagent_messages` alone — now `[user, assistant(tool-calls A), assistant(tool-calls B), …]` with dangling tool-calls. AI SDK v6 rejects it before the loop's `replayState.priorTools` reconciler ever runs (the reconciler only fires inside tool-dispatch, *after* `chat()`), and the job retries the same broken history until `max_attempts`.

Observed: 23/23 stuck jobs had this gap pattern; the missing-tool-call ids in the error matched the persisted assistant turn's `tool_use_id`s exactly.

## Bug 2 — non-JSON tool outputs rejected by AI SDK v6

AI SDK v6 validates each tool-result `json` output against a strict `JSONValue` Zod schema. gbrain tool outputs routinely contain non-JSON values: **node-postgres returns `timestamptz` columns as JS `Date`**, so `brain_get_page` / `brain_list_pages` rows include Date-typed `updated_at`/`created_at`. A raw `Date` (also `undefined`/`bigint`) makes the entire tool message fail the `ModelMessage` union → `Invalid prompt: The messages do not match the ModelMessage[] schema` on the turn *after* such a tool runs.

Crash-replay masks it: on resume the value has already been round-tripped through the `jsonb` column to an ISO string, so a resumed run validates — until the next live tool call produces a fresh `Date`. (`brain_search`, which returns no Date fields, never trips it.)

## Reproduction

With `agent.use_gateway_loop=true` and any non-Anthropic model, run a subagent that makes multi-turn tool calls including `brain_list_pages`/`brain_get_page`, and restart the worker mid-run. Bug 2 fails the live run; Bug 1 then loops every retry.

## Fix

PR linked below:
- `toolLoop`: add an `onToolResultMessage` callback (fired before the in-memory push, write-before-use), wired in `runSubagentViaGateway` to persist the tool-result message. `runSubagentViaGateway` also self-heals already-corrupted jobs by reconstructing each missing tool-result message from settled `subagent_tool_executions` rows.
- `toModelMessages`: normalize tool-result `json` values to plain JSON (`Date`→ISO, `undefined` dropped) at the AI SDK boundary.

Tests: forward persistence + self-healing resume (no tool re-execution); `Date`→ISO normalization + a real `generateText` validation accepting a Date-bearing tool output.

Validated end-to-end against a live `llama-server:qwen36b` dream backlog: jobs that previously looped now run the full multi-turn synthesis to completion with contiguous `message_idx`.

rafaelreis-r:
Fixes #2256.

The gateway-native subagent loop (`agent.use_gateway_loop=true`, mandatory for non-Anthropic providers) couldn't reliably complete a multi-turn tool conversation. Two independent bugs combined to make subagent jobs loop on retries forever. Both surface as AI SDK v6 errors: `Invalid prompt: The messages do not match the ModelMessage[] schema` and `Tool results are missing for tool calls <ids>`. The Anthropic-direct path is unaffected.

## Commit 1 — persist tool-result messages; self-heal on resume

`toolLoop()` fed each turn's tool results back as a `role:'user'` message but never persisted it (`void userMessageIdx`), and `runSubagentViaGateway` discarded `ToolLoopResult.messages`. So `subagent_messages` had gaps at every even `message_idx`. On any resume, `loadPriorMessages` rebuilt a history of adjacent assistant turns with dangling tool-calls, which AI SDK v6 rejects — before the `replayState.priorTools` reconciler (which only runs inside tool-dispatch, after `chat()`) ever gets a chance. The job then looped to `max_attempts`. The legacy Anthropic-direct path always persisted this message.

- `toolLoop`: new `onToolResultMessage` callback, fired before the in-memory push (write-before-use).
- `runSubagentViaGateway`: wires it to persist the tool-result message; and `rebuildReplayHistory()` self-heals already-corrupted jobs by reconstructing each missing tool-result message from settled `subagent_tool_executions` rows (matched by provider `toolCallId`) and re-persisting it. Conservative: a still-pending/absent exec leaves the turn un-healed rather than fabricating data. `nextMessageIdx` now derives from `max(known idx)+1` so a healed gap can't collide under the unique `(job_id, message_idx)` constraint.

## Commit 2 — JSON-normalize tool-result outputs

AI SDK v6 validates each tool-result `json` output against a strict `JSONValue` schema. node-postgres returns `timestamptz` as JS `Date`, so `brain_get_page`/`brain_list_pages` rows carry Date-typed `updated_at`/`created_at`; a raw `Date` (also `undefined`/`bigint`) fails the `ModelMessage` union the turn after such a tool runs. Replay masked it (the jsonb round-trip had already stringified the Date).

- `toModelMessages`: normalize the `json` value via `toJsonValue()` (JSON round-trip: `Date`→ISO, `undefined` dropped, non-serializable→string), plus a non-throwing `safeStringify()` for the error-text branch. Persisted blocks are unaffected (write path already JSON-stringifies, so stored and sent forms now agree).

## Tests
- `test/e2e/subagent-gateway-resume.test.ts` — forward persistence (contiguous `message_idx`, no gaps) + self-healing resume from a pre-fix corrupted job without re-executing the tool.
- `test/gateway-model-messages.test.ts` — `Date`→ISO normalization, `undefined` drop, and a real `generateText` validation accepting a Date-bearing tool output (no `ModelMessage` rejection).
- Existing `subagent-gateway-path` and cross-provider `subagent-crash-replay-multi-provider` suites stay green (33 pass on this branch).

## Validation
Deployed against a live `llama-server:qwen36b` dream backlog where 23/23 multi-turn jobs were stuck looping. After the fix, jobs that previously failed every retry run the full multi-turn synthesis to completion with contiguous `message_idx` (verified one job: 16 messages, idx 0–15, real synthesis output; the rest drained the same way). The only non-completions were pre-existing `wall-clock timeout` deaths on the heaviest transcripts that had already exhausted most of their retry budget on the old bug — not a regression.

### DIFF
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -2908,6 +2908,22 @@ export interface ToolLoopOpts {
   ) => Promise<{ gbrainToolUseId: string }>;
   onToolCallComplete?: (gbrainToolUseId: string, output: unknown) => Promise<void>;
   onToolCallFailed?: (gbrainToolUseId: string, error: string) => Promise<void>;
+  /**
+   * Persist the user-role message that carries this turn's tool results back to
+   * the model. Fires once per turn, AFTER every tool execution settles and
+   * BEFORE the results are appended to the in-memory history (write-before-use).
+   *
+   * The legacy Anthropic-direct path always persisted this message; the
+   * gateway path originally dropped it (`void userMessageIdx`), so on
+   * crash-replay `loadPriorMessages` rebuilt a history that ended with an
+   * assistant turn whose tool-calls had NO matching tool-result message. AI SDK
+   * v6 then rejected the prompt — "Tool results are missing for tool calls ..."
+   * (when another assistant turn followed) or "messages do not match the
+   * ModelMessage[] schema" (when it was the last message) — and the job retried
+   * the same broken history until max_attempts. Pinned by
+   * test/e2e/subagent-gateway-resume.test.ts.
+   */
+  onToolResultMessage?: (messageIdx: number, blocks: ChatBlock[]) => Promise<void>;
 
   /** Optional per-call heartbeat for observability. */
   onHeartbeat?: (event: string, data: Record<string, unknown>) => void;
@@ -3131,9 +3147,13 @@ export async function toolLoop(opts: ToolLoopOpts): Promise<ToolLoopResult> {
 
     if (stopReason === 'aborted') break;
 
-    // Feed all tool results back as a single user message.
+    // Feed all tool results back as a single user message. Persist it BEFORE
+    // appending to the in-memory history (write-before-use) so a crash here
+    // leaves the DB holding a complete, replayable turn: the assistant's
+    // tool-calls AND the matching tool-result message. Dropping this write was
+    // the root cause of the gateway-path resume failures (see onToolResultMessage).
     const userMessageIdx = messageIdx++;
-    void userMessageIdx;
+    await opts.onToolResultMessage?.(userMessageIdx, toolResultBlocks);
     messages.push({ role: 'user', content: toolResultBlocks });
 
     turnIdx++;
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -2323,6 +2323,23 @@ export interface ChatToolDef {
   inputSchema: Record<string, unknown>;
 }
 
+/** JSON.stringify that never throws (bigint / circular → best-effort string). */
+function safeStringify(v: unknown): string {
+  try { return JSON.stringify(v) ?? 'null'; } catch { return String(v); }
+}
+
+/**
+ * Coerce an arbitrary tool-output value into a plain JSON value the AI SDK's
+ * strict JSONValue schema accepts. Round-tripping through JSON converts `Date`
+ * to an ISO string, drops `undefined` object properties, and normalizes nested
+ * structures. Non-serializable inputs (bigint, circular) fall back to a string.
+ * Returns `null` for nullish input so the json part always has a defined value.
+ */
+function toJsonValue(v: unknown): unknown {
+  if (v === null || v === undefined) return null;
+  try { return JSON.parse(JSON.stringify(v)); } catch { return String(v); }
+}
+
 /**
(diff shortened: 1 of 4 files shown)


time-attack:

PR #2259, opened 2026-06-17
feat(sync): GBRAIN_NO_PRUNE_DIRS — un-prune a content dir (e.g. an ops/ KB)

donogeme:
## What

Adds `GBRAIN_NO_PRUNE_DIRS` (comma-separated env var) that removes directory names from `PRUNE_DIR_NAMES` at walk time, so a brain can sync a tree whose top-level name happens to collide with a pruned name.

## Why

`PRUNE_DIR_NAMES = {node_modules, .raw, ops}` is hardcoded. `ops` in particular is a *semantic* prune ("operations, not knowledge") — great for repos that use `ops/` for throwaway operational files. But some brains use `ops/` as a genuine **knowledge base** (e.g. `ops/_kb/{compliance,finance,partners,...}`). For them, the entire tree is silently invisible to search with no opt-out — a single hardcoded name decides it.

This gives an escape hatch without changing any default: set `GBRAIN_NO_PRUNE_DIRS=ops` and that tree syncs.

## Behavior

- Comma-separated; trims whitespace, drops empties/dupes.
- Only un-prunes names in `PRUNE_DIR_NAMES` (and the `*.raw` suffix rule).
- **Dot-prefixed dirs (`.git`, `.obsidian`, …) and git submodules stay pruned regardless** — those are never content.
- Unset/empty = exactly today's behavior. Purely additive, no default change.
- Routes through `pruneDir`, so every walker (sync, extract, transcript-discovery) honors it consistently.

## Example

```bash
GBRAIN_NO_PRUNE_DIRS=ops gbrain sync --all
```

## Tests

`test/sync-no-prune-dirs.test.ts` — parser contract + prune/un-prune matrix (ops un-pruned, node_modules still pruned, dot-dirs always pruned). `bun run typecheck` clean.

## Context

Env-var form mirrors the existing `GBRAIN_SOURCE`/`GBRAIN_SOURCES` convention. Happy to reshape as a `sync.*` config key instead if you'd prefer config-driven over env.

### DIFF
--- a/src/core/sync.ts
+++ b/src/core/sync.ts
@@ -261,14 +261,35 @@ const PRUNE_DIR_NAMES = new Set<string>([
  * existing callers stay back-compat; new callers (sync walker, extract
  * walker) thread it through.
  */
+/**
+ * Names to NOT prune even though they're in PRUNE_DIR_NAMES (or match the
+ * `.raw` sidecar rule) — an escape hatch for brains that legitimately use one
+ * of these names as a content directory (e.g. an `ops/` knowledge base instead
+ * of throwaway operational files). Set `GBRAIN_NO_PRUNE_DIRS` to a comma-
+ * separated list (e.g. `ops`). Empty/unset = default pruning. Dot-prefixed dirs
+ * and git submodules are ALWAYS pruned regardless (genuine non-content).
+ */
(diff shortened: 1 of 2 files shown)


time-attack:
Reviewed and declined for now with the content-scope set, pending one unified mechanism. Thank you.

PR #2261, opened 2026-06-17
fix(propose-takes): insert sentinel rows for zero-proposal pages to avoid re-extraction every cycle

### LINKED ISSUE #2106: propose_takes re-extracts every zero-take page on every cycle — no negative-result cache
**Type:** bug
**Component:** `dream` cycle / `propose_takes`
**Version:** gbrain `0.42.40.0` (master HEAD, commit `ecd6ae8`)
**Engine:** PostgreSQL

## Summary

`propose_takes` claims a content-hash idempotency contract — *"an unchanged page never re-spends LLM tokens"* — but that contract only holds for pages that produce **≥1 take**. A page that yields **zero** proposals never gets a `take_proposals` row, so its idempotency key is never stored, so it is **re-sent to the LLM on every cycle** in which it falls inside the scan window. In a steady brain, the *majority* of pages have no takeable claims (concept pages, stubs, transcripts, reference), so the phase spends the large majority of its tokens re-extracting pages that will never produce a take.

## The idempotency contract (and why it's false for zero-take pages)

`src/core/cycle/propose-takes.ts` JSDoc:

> *Idempotency contract (D17 schema spec): The unique index on (source_id, page_slug, content_hash, prompt_version) means an unchanged page never re-spends LLM tokens.*

The unique index does guarantee that — **but only because a row gets written.** The check and the write are asymmetric:

- **Check** (`propose-takes.ts:349`): skip the page iff a row exists for `(source_id, page_slug, content_hash, prompt_version)`:
  ```ts
  const cached = await engine.executeRaw(
    `SELECT id FROM take_proposals
     WHERE source_id = $1 AND page_slug = $2 AND content_hash = $3 AND prompt_version = $4
     LIMIT 1`, [sourceId, page.slug, ch, promptVersion]);
  if (cached.length > 0) { result.cache_hits += 1; continue; }
  ```
- **Write** (`propose-takes.ts:392-398`): the *only* writer of that key is the per-proposal loop:
  ```ts
  for (const p of proposals) {
    await engine.executeRaw(
      `INSERT INTO take_proposals (...) VALUES (...)
       ON CONFLICT (source_id, page_slug, content_hash, prompt_version) DO NOTHING`, [...]);
  }
  ```

If `proposals.length === 0`, the loop body never runs, **no row is written, and the page is a permanent cache-miss.** Nothing else writes that key (verified: `take_proposals` is referenced only by this producer + the schema DDL across the whole tree).

This compounds with the scan strategy: the phase scans `sort: 'updated_desc'`, `pageLimit` default **100** (`propose-takes.ts:308,325`), `skipPagesWithFence` default **false** (`:309`), and runs on the autopilot every cycle (it's in `ALL_PHASES`, gated only by `phases.includes('propose_takes')` at `cycle.ts:1962`, not pack-gated). Any churn that bumps `updated_at` (sync/embed/consolidate, ingest) keeps rotating the same zero-take pages back through the top-100 window, and each appearance is a fresh LLM call.

## Evidence (one production brain, Postgres, ~3 weeks)

- Total live pages: **2,677**; pages with a non-empty body but **zero** `take_proposals` rows: **1,659**.
- Rows ever written to `take_proposals`: **1,466** (across 1,038 distinct `page_slug`); single `prompt_version` (`v0.36.1.0-tuned-cat15`) — i.e. **no** prompt-version invalidation.
- `propose_takes` LLM submits over 3 weeks (from the dream-budget audit log): **~13,600** (one week peaked at 8,671; one hour peaked at 596 calls).
- So only **~11%** of calls produced a cacheable result; **~85–89%** were re-extraction of pages that yield no take and never cache. The *positive* cache works perfectly (no productive `(slug, content_hash)` pair is ever re-extracted across runs) — the entire leak is the missing **negative** cache.
- At the logged ~$0.012/call this is **~$100/week** of pure waste in this deployment; it also defeats the per-cycle budget cap on unpriced models (e.g. OpenRouter), since those calls log `submit_unpriced` and never project a cost to trip the cap.

## Reproduce

1. A brain with many body-bearing pages that contain no gradeable claims (concepts/reference/transcripts).
2. Run the dream cycle repeatedly (or let autopilot run).
3. Observe `take_proposals` row count stays flat while `propose_takes` LLM submits keep accruing each cycle; the zero-take pages are re-sent every run they fall in the `updated_desc` top-100 window.

## Suggested fix

Cache the **negative** result: after the extractor returns, if `proposals.length === 0`, still stamp the `(source_id, page_slug, content_hash, prompt_version)` key so the `:349` SELECT short-circuits next cycle. The composite key already self-invalidates correctly — a page whose body later changes gets a new `content_hash` (SELECT miss → re-extract), and a `PROMPT_VERSION` bump re-scans everything — so no legitimately-new take is ever suppressed.

Two implementation options:
- **Sentinel row in `take_proposals`** (no migration): insert one marker row on the zero-proposal path via the same `ON CONFLICT DO NOTHING` upsert. To keep it out of the operator's pending view (`take_proposals_pending_idx` is `WHERE status='pending'`), set `status='superseded'` (an already-legal CHECK value) and a recognizable marker (`kind='_skip'`, `claim_text=''`, `proposal_run_id` prefixed `propose-skip-…`). Satisfies the NOT NULL columns; adds nothing to the queue.
- **Dedicated marker** (cleaner, small migration): a `take_scan_marker(source_id, page_slug, content_hash, prompt_version, scanned_at)` table (or a `last_take_scan_hash` column), written unconditionally after each extract and consulted alongside the existing SELECT. Keeps `take_proposals` purely positive; costs a migration.

Either removes ~85% of `propose_takes` token spend with no loss of coverage.

## Related / adjacent issues

- [#1467 — Take proposal workflow: configurable provider, multiple proposals, review CLI](https://github.com/garrytan/gbrain/issues/1467) — the missing consumer/review surface. *Different concern*: this report is purely the producer-side token waste and is orthogonal to whether a review surface exists.
- [#1412 — `propose_takes` phase times out on large corpus (3,125 pages); no configurable timeout or standalone invocation](https://github.com/garrytan/gbrain/issues/1412) — related symptom (the phase is expensive / slow) but a different root cause.
- [#1781 — dream `extract_facts` non-idempotent: facts + fence rows duplicate each cycle (Postgres)](https://github.com/garrytan/gbrain/issues/1781) — *sibling* class of "dream phase re-does work every cycle"; this is the `propose_takes` analogue.
- [#2088 — long-running jobs worker never refreshes AI gateway config; `propose_takes` hard-fails per page on the hardcoded anthropic default](https://github.com/garrytan/gbrain/issues/2088) — same phase, adjacent failure mode (model resolution), independent of this caching bug.

LittleRandom:
Closes #2106

## Problem

`propose_takes` claims an idempotency contract — "an unchanged page never re-spends LLM tokens" — but this only holds for pages that produce ≥1 take. A page yielding zero proposals never gets a `take_proposals` row, so its idempotency key is never stored. It is re-sent to the LLM every cycle it falls inside the scan window.

In a steady brain, the majority of pages have no takeable claims (concept pages, stubs, transcripts, reference), so the phase spends ~85-90% of its tokens re-extracting pages that will never produce a take.

## Fix

Insert a sentinel row with `claim_text = '__no_proposals__'` when the extractor returns `[]`. The composite unique index on `(source_id, page_slug, content_hash, prompt_version)` catches it on future cycles:

- Page unchanged → same content_hash → SELECT finds the sentinel → cache hit → skip LLM call
- Page edited → new content_hash → SELECT misses → re-extract (legitimate)
- Prompt version bumped → new prompt_version → full re-scan (fresh prompt)

## Verification

Before fix: 89 pages scanned, 0 cached, **1060 seconds**
After fix: 94 pages scanned, 93 cached, **6.8 seconds total cycle**

~156x faster. Verified on a production brain with PostgreSQL engine.


time-attack:
Thank you for the diagnosis and the before/after timing numbers — the analysis was correct and the idle-cost bug was real. It has since been fixed on master by commit 3fcca330 (#3319, from #2514), which writes an empty-extraction tombstone row on the same `(source_id, page_slug, content_hash, prompt_version, md5(claim_text))` tuple. Closing as already fixed; no action needed on your side.

PR #2262, opened 2026-06-18
fix(cycle): bound propose_takes phase with per-call + phase-level deadlines

tschew72:
## Summary

The \`propose_takes\` phase was being killed by SIGTERM from outer \`timeout 600\` wrappers in cron-driven dream runs. Root cause: \`pageLimit=100\` default combined with the unbounded \`gateway.chat\` call in \`defaultExtractor\` made the phase routinely take 50+ minutes.

This PR introduces three layered bounds that turn a hard-kill SIGTERM into a clean partial result:

1. **\`EXTRACTOR_CALL_TIMEOUT_MS = 90_000\`** — per-call \`AbortSignal.timeout\` on the extractor. A stalled provider socket aborts the single call; the page is logged as a warning; the phase continues.
2. **\`PHASE_DEADLINE_MS = 30 * 60 * 1000\`** — phase wall-clock deadline checked inside the page loop. Phase returns \`deadline_hit: true\` instead of being killed by an outer wrapper.
3. **\`pageLimit\` default 100 → 30** — 30 pages × ~30s = 15 min fits inside the 30-min phase deadline and a $5 budget.

## What changes

- \`src/core/cycle/propose-takes.ts\` — 53 insertions, 1 deletion
- \`ProposeTakesResult\` gains optional \`deadline_hit?: boolean\` field (older consumers ignore it)

## Why this matters

Surfaced via the \`propose_takes aborted SIGTERM\` pattern that hit the nightly dream cycle from 2026-06-15 onward (post-v0.42.20.0 ship, when extraction prompts moved to the production path). After this fix, the phase returns a partial result with whatever proposals it managed to insert before the deadline — making dream runs **resumable** and the lost proposals **observable** in the dream log.

## Test plan

- [x] Live test on our 219-page brain: phase completed cleanly in ~4 min with \`pageLimit=30\`
- [x] No \`propose_takes aborted SIGTERM\` in dream.log since 2026-06-18
- [x] Brain score 83/100 preserved through the change
- [ ] Need upstream CI to confirm on the larger hermetic test set

## Reproduction (before fix)

\`\`\`bash
 On a brain with >100 pages, run:
gbrain dream --phase propose_takes
 Watch dream.log tail — see `propose_takes aborted SIGTERM` after ~10 min
 (cron wrapper kills the process before page loop completes)
\`\`\`

## Fix verification

\`\`\`bash
After this PR:
gbrain dream --phase propose_takes
Phase completes in ~4 min with proposals_inserted: 28
Or, on a slow LLM day, returns deadline_hit: true with whatever it got
\`\`\`

### DIFF
--- a/src/core/cycle/propose-takes.ts
+++ b/src/core/cycle/propose-takes.ts
@@ -153,6 +153,8 @@ export interface ProposeTakesResult {
   cache_misses: number;
   proposals_inserted: number;
   budget_exhausted: boolean;
+  /** v0.42.20.0+ — true when PHASE_DEADLINE_MS fired before page loop completed. */
+  deadline_hit?: boolean;
   warnings: string[];
 }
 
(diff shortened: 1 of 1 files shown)


time-attack:

PR #2267, opened 2026-06-18
fix(schema-pack): embed bundled pack YAMLs in the compiled binary (+ entity-slug orphan floor)

JiraiyaETH:
## The bug (Layer 0 — the foundational outage)
Bundled schema-pack YAMLs are **never embedded** in `bun build --compile` binaries. The pack locators resolve paths via `fileURLToPath(import.meta.url) + existsSync`, which points at `/$bunfs/root/...` in a compiled binary where the YAML is **not on disk**. Symptoms in a deployed binary:
- `gbrain schema active` → `unknown schema pack: gbrain-base`. Any active pack that `extends: gbrain-base` fails to resolve because the bundled parent can't load, taking the whole resolution down.
- `put_page` **silently degraded** — the pack-load `try/catch` swallowed the failure, reverted to a hardcoded prefix table, and **skipped type validation** → accumulating page-type sprawl.
- `gbrain schema show gbrain-base` and the lens packs (`gbrain-creator`/`investor`/`engineer`/`everything`) were unreachable from the CLI/MCP.

`build:schema` is **not** the cause (it regenerates Postgres DDL); a plain rebuild reproduces the bug. The fix is a source-level embed.

## The fix
- **New central registry `src/core/schema-pack/bundled-packs.ts`** imports all 7 bundled pack YAMLs via `import … with { type: 'file' }` (Bun embeds the bytes; the import evaluates to a readable path in **both** `bun run` and `--compile`). Mirrors the existing WASM-embed pattern (`src/core/chunkers/code.ts` + `scripts/check-wasm-embedded.sh`). Exports `BUNDLED_PACK_PATHS`, `BUNDLED_PACK_LIST`, `BUNDLED_PACK_NAMES`, `bundledPackPath()`.
- **Routed every pack-load surface through it** (the previous code had four divergent locators): `load-active.ts` `defaultPackLocator`, the CLI `schema show/validate/list/use/fork` (`packPathByName` + `runList`), MCP `list_schema_packs` + `schema_lint` named-pack, and the read-only mutability set in `mutate.ts` (was 3 names, now all 7).
- **Fail-closed `put_page`** — a *configured* pack (resolution source ≠ `default`) that won't load now throws instead of silently degrading; a default-pack load failure loud-warns.
- **Functional regression guard** `scripts/check-packs-embedded.sh` + `scripts/packs-smoketest.ts`: compiles a probe and loads every bundled pack **through the compiled binary** (not `strings | grep`). Wired into `check:all`, `run-verify-parallel`, and `ci-local`. Updated the `BUNDLED_PACK_NAMES` size test 3 → 7. Docs refreshed.

## Layer 1a — entity-slug floor (no literal-`null`/hyphen-flattened fact orphans)
- `extract.ts`: a model-emitted `entity` of `"null"`/`"none"`/whitespace is coerced to JSON null (the fact stays unbound, never a `null`-slug orphan).
- `entities/resolve.ts`: non-entity tokens → null binding (kept unbound via the existing legacy bucket, not dropped); the floor uses `slugifyEntityPath`, which preserves an explicit **entity-prefix** path (`companies/Acme Co` → `companies/acme-co`) but flattens non-prefix slashes (`A/B Partners` → `a-b-partners`) so it can't mint arbitrary nested pages past the stub guard.
- Read path (`recall` + MCP `list_facts`): a null resolution returns no facts instead of querying the raw string and surfacing legacy orphan rows.

## Deferred (called out, not in this PR)
- `registry.ts` `resolvePack` returns the child manifest unchanged for `extends`/`borrow_from`, so `gbrain-recommended`/`gbrain-everything` resolve with 0 page types. Not the outage (operational packs declare their types directly), but worth a follow-up to honor the lens-pack design.
- Deriving the entity prefix set from the active pack's `path_prefixes` instead of the hardcoded `people/`+`companies/`.

## Tests
typecheck clean · `scripts/check-packs-embedded.sh` passes (compiled-binary probe) · the schema-pack/facts/entity suites pass (246 tests). Verified live: a freshly compiled binary resolves the configured active pack and `schema validate <lens-pack>` reads its embedded `/$bunfs/root/...yaml`.

### DIFF
--- a/src/commands/schema.ts
+++ b/src/commands/schema.ts
@@ -19,6 +19,7 @@
 
 import { existsSync, readdirSync, readFileSync, writeFileSync, mkdirSync } from 'node:fs';
 import { dirname, join } from 'node:path';
+import { bundledPackNames, bundledPackPath } from '../core/schema-pack/bundled-packs.ts';
 import {
   addAliasToType,
   addLinkTypeToPack,
@@ -179,7 +180,7 @@ async function runActive(_args: string[]): Promise<void> {
 }
 
 function runList(_args: string[]): void {
-  const bundled = ['gbrain-base', 'gbrain-recommended'];
+  const bundled = bundledPackNames();
   const installedDir = gbrainPath('schema-packs');
   const installed: string[] = [];
   if (existsSync(installedDir)) {
@@ -366,18 +367,11 @@ function runUse(args: string[]): void {
 }
 
 function packPathByName(name: string): string | null {
-  if (name === 'gbrain-base') {
-    // Resolve bundled YAML — try a few locations.
-    const here = dirname(new URL(import.meta.url).pathname);
-    const candidates = [
-      join(here, '..', 'core', 'schema-pack', 'base', 'gbrain-base.yaml'),
-      join(here, '..', '..', 'src', 'core', 'schema-pack', 'base', 'gbrain-base.yaml'),
-    ];
-    for (const c of candidates) {
-      if (existsSync(c)) return c;
-    }
-    return null;
-  }
+  // Bundled packs resolve to their embedded YAML (bundled-packs.ts) — works
+  // in dev AND the compiled binary, for EVERY bundled pack. The old resolver
+  // hardcoded only gbrain-base, so `show gbrain-base-v2` failed even in dev.
+  const embedded = bundledPackPath(name);
+  if (embedded) return embedded;
   const baseDir = gbrainPath('schema-packs', name);
   for (const c of ['pack.yaml', 'pack.yml', 'pack.json']) {
     const candidate = join(baseDir, c);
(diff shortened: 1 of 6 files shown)


time-attack:
Merging this one. I confirmed on current master that bundled pack YAMLs are still located via `fileURLToPath(import.meta.url) + existsSync` with no `type: 'file'` embed anywhere, so they genuinely are not present in the `bun build --compile` artifact we ship — which means schema-pack resolution is broken in the binary and put_page silently skips type validation. The embed approach matches the existing WASM/admin-embed precedent and the compiled-binary smoketest is the right kind of guard. One thing before it lands: the registry-consolidation portion already arrived on master as `src/core/schema-pack/bundled.ts` (via #3110), so please rebase and fold `bundled-packs.ts` into that file instead of adding a second registry.

PR #2268, opened 2026-06-18
fix(autopilot): resilience + hot-memory hardening (per-phase timeout, dispatch probe, op-checkpoint, facts visibility)

JiraiyaETH:
Autopilot resilience + hot-memory hardening (6 commits + a test fix). All off `origin/master`.

## Commits
- **minions:** env-resolvable + clamped DB-probe self-defense thresholds.
- **frontmatter:** derive the validate slug from the brain root (`git rev-parse`) so it's stable regardless of CWD.
- **cycle:** per-phase soft timeout + an autopilot git-commit phase — a slow phase aborts at its next batch boundary, self-truncates (`status:'skipped', reason:'phase_timeout'`), and the cycle continues instead of dying at the job wall-clock. Partial work is durable (watermarks) and resumes next cycle.
- **autopilot:** bound the dispatch health-probe + enable the per-phase timeout on cycles.
- **facts:** `facts.default_visibility` knob (configurable hot-memory default).
- **op-checkpoint:** bind `completed_keys` via `to_jsonb(text[])` instead of `JSON.stringify` + `$3::jsonb` (the latter double-encodes → jsonb string scalar → violates the `op_checkpoints_completed_keys_array` CHECK). (#2255 follow-up.)
- **test(cycle):** pass a real `AbortSignal` to the autopilot-cycle handler tests. `runPhaseWithTimeout` registers an abort listener via `parentSignal.addEventListener`; the tests passed a fake `{ aborted: false } as any` mock with no `addEventListener`, so they threw once the timeout path engaged. The real worker passes a genuine `AbortSignal` (production cycles run fine). No production code change.

## Tests
Full suite green in a clean environment. (Two tests are environment-sensitive only when run against a live machine: `resolveBoostMap` reads `GBRAIN_SOURCE_BOOST` from the ambient shell, and `worker-registry.serial` contends with a running autopilot writing under `~/.gbrain` — both pass in isolated CI.)


time-attack:
Thanks for putting this together — several of the individual commits are reasonable. The problem is packaging: 1044 lines across 21 files spanning six unrelated concerns, including genuinely new surface (an autopilot git-commit phase, a `facts.default_visibility` knob) alongside behavior changes, which makes it unreviewable as one unit. The one clear bug fix in the set, the op-checkpoint `completed_keys` double-encoding, has already landed on master via the `$3::text::jsonb` binding (#2339). Closing this; if the per-phase timeout is causing you real production pain, a standalone PR for just that would be a much easier read.

PR #2274, opened 2026-06-18
fix(subagent): persist + reconcile tool-result turns in the gateway loop

### LINKED ISSUE #2273: gateway-loop subagent path never persists tool-result user turns — every resume rebuilds an unbalanced conversation that next chat() rejects
**What happened?**

Any subagent job that takes the gateway-loop path (`agent.use_gateway_loop = true`, the path every non-Anthropic recipe uses) dies deterministically on the next `chat()` dispatch after ANY mid-conversation interruption with:

```text
[chat(litellm:gpt-X.Y)] Tool results are missing for tool calls call_<id>, call_<id>, ...
```

The failure scales with parallel-tool-call count (gpt-5.5 dropped 10 ids per dispatch in production, gpt-5.4 dropped 6, gpt-5.4-mini dropped 2), which initially reads as a provider-side bridge bug. It is not. The missing-tool-results error comes from the NEXT `chat()` dispatch operating on a stale in-memory message array, not from the wire response of the prior call.

**Diagnosis**

`gateway.toolLoop` builds the tool-result user turn in memory after the tool executions settle, then throws the index away. `runSubagentViaGateway` wires no callback to persist that turn. On any interruption (worker recycle, rate-lease renewal, RSS watchdog, abort, even a clean shutdown), replay rebuilds the conversation from `subagent_messages` without the tool-result row, an unbalanced shape with assistant `tool_call` turns and no matching `tool` user turns.

Unlike the legacy Anthropic-direct path (which reconciles the turn on replay), the gateway-loop path has no reconciliation. The next `chat()` dispatch hits the provider's validator with the unbalanced shape and the provider returns the "Tool results are missing for tool calls ..." error, which the worker treats as a hard failure, retries the same unbalanced shape, and dead-letters the job.

The existing crash-replay test suite missed this because it stubs the chat transport, so the real provider validator never runs and message balance went unasserted.

**What did you expect?**

The gateway-loop path should persist the synthesized tool-result user turn before the next dispatch, matching what the legacy Anthropic-direct path already does. On replay, it should reconcile the missing turn from the settled `subagent_tool_executions` rows (the tool outputs were persisted; only the wrapping user turn was lost).

**Steps to reproduce**

1. Brain configured for any non-Anthropic chat provider (litellm-proxy, codex-proxy, ollama, openrouter, openai directly, …). Set `agent.use_gateway_loop = true`.
2. Submit any subagent job whose dispatch fires parallel tool calls (`gbrain dream --phase patterns` is the deterministic local trigger; any subagent that calls a tool reproduces).
3. Interrupt mid-conversation: kill the worker, force a rate-lease renewal, or let the RSS watchdog recycle the process.
4. The worker re-claims the job and replays. Replay rebuilds the conversation from persisted rows.
5. The next `chat()` throws `Tool results are missing for tool calls call_<id>, ...`. The worker retries; same error; job dead-letters.

Observed on gbrain 0.42.43.0 through 0.42.51.0 under codex-proxy / litellm-proxy. Reproduces on every non-Anthropic chat recipe when `agent.use_gateway_loop = true`.

**Environment**

- Bun: 1.3.14.

**Proposed fix**

Two pieces, mirroring what the legacy Anthropic path already does:

1. `gateway.toolLoop` gains an `onToolResultTurn` callback fired when the tool-result user turn is built, so the caller can persist the turn before the next dispatch.
2. `runSubagentViaGateway` wires the callback to persist the turn AND adds reconciliation on replay: when prior messages end with unanswered tool-calls, re-synthesize the tool-result turn from the settled `subagent_tool_executions` rows. Bail if any tool is still unsettled so a non-idempotent tool is never re-run.

Regression test: a new `test/e2e/subagent-gateway-toolresult-replay.test.ts` that captures the real transport input and asserts the resumed conversation is balanced. Fails without the change; passes with it.

Paired PR: #2274.

brettdavies:
Closes #2273

## Summary

**The bug.** Subagent jobs that take the gateway-loop path (`agent.use_gateway_loop = true`, used by every non-Anthropic recipe) die deterministically after any mid-conversation interruption with `[chat(litellm:gpt-X.Y)] Tool results are missing for tool calls call_<id>, ...`. The failure scaled with parallel-tool-call count, which read as a provider-bridge bug but was not. The error came from the next `chat()` dispatch operating on a stale in-memory message array, not from the wire response of the prior call.

**The fix.** `gateway.toolLoop` gains an `onToolResultTurn` callback fired when the tool-result user turn is built. `runSubagentViaGateway` wires the callback to persist the turn before the next dispatch AND reconciles on replay: when prior messages end with unanswered tool-calls, it re-synthesizes the tool-result turn from the settled `subagent_tool_executions` rows. Bails if any tool is still unsettled, so a non-idempotent tool is never re-run.

This mirrors the legacy Anthropic-direct path's existing reconciliation behavior; the gateway path was the only dispatch shape that lacked it.

## Diagnosis

- `gateway.toolLoop` (pre-patch) built the tool-result user turn in memory after the tool executions settled, then discarded the index. `runSubagentViaGateway` had no persistence hook, so the turn never reached `subagent_messages`.
- On any interruption (worker recycle, rate-lease renewal, RSS watchdog, abort, clean shutdown), replay rebuilt the conversation from `subagent_messages` without the tool-result row. The shape was unbalanced: assistant `tool_call` turns with no matching `tool` user turns.
- The next `chat()` dispatch hit the provider's validator with the unbalanced shape; provider returned the "Tool results are missing for tool calls ..." error; worker treated as hard failure, retried the same unbalanced shape, dead-lettered the job.
- The legacy Anthropic-direct path in `src/core/minions/handlers/subagent.ts` already persists the user-turn reconciliation; the gateway path was simply never extended.

## Reproduction

1. Brain configured for any non-Anthropic chat provider (litellm-proxy, ollama, openrouter, openai, deepseek, groq). Set `agent.use_gateway_loop = true`.
2. Submit a subagent job whose dispatch fires parallel tool calls (`gbrain dream --phase patterns` is the deterministic local trigger).
3. Interrupt mid-conversation: `kill -TERM` the worker, or let the RSS watchdog recycle the process.
4. The worker re-claims the job and replays. Replay rebuilds the conversation from persisted rows.
5. Next `chat()` throws `Tool results are missing for tool calls call_<id>, ...`. Job dead-letters.

## Tests

- New `test/e2e/subagent-gateway-toolresult-replay.test.ts`: captures the real transport input shape and asserts the resumed conversation is balanced, parametrized across the recognized tool-capable non-Anthropic providers (openai, deepseek, groq, together) so the provider-agnostic property is proven per provider, not assumed. Fails on master without the patch; passes with it. ~5s runtime.
- Verified end-to-end on a production-shaped brain routing through an OpenAI-compatible proxy. Pre-patch: every interruption dead-lettered the active jobs; post-patch: every interruption resumed cleanly across worker recycles, RSS-watchdog kills, and clean restarts.

`agent.use_gateway_loop` is missing from `KNOWN_CONFIG_KEYS` (operators have to discover the `--force` escape hatch from the error message). And with the gateway path's recipe coverage now broad (every non-Anthropic provider plus Anthropic itself via `native-anthropic`), the default could flip from `false` to `true`. Both worth follow-up discussion; not in scope here.

### DIFF
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -2958,6 +2958,16 @@ export interface ToolLoopOpts {
   onToolCallComplete?: (gbrainToolUseId: string, output: unknown) => Promise<void>;
   onToolCallFailed?: (gbrainToolUseId: string, error: string) => Promise<void>;
 
+  /**
+   * Persist the synthesized tool-result user turn AFTER every tool in the turn
+   * settles. Without it the assistant tool-call turn is persisted (onAssistantTurn)
+   * but its results are not, so crash-replay rebuilds an unbalanced conversation
+   * (assistant tool-calls with no matching tool-results) and the next chat()
+   * dispatch throws "Tool results are missing". Mirrors the legacy Anthropic
+   * path's user-turn persist in subagent.ts.
+   */
+  onToolResultTurn?: (turnIdx: number, messageIdx: number, blocks: ChatBlock[]) => Promise<void>;
+
   /** Optional per-call heartbeat for observability. */
   onHeartbeat?: (event: string, data: Record<string, unknown>) => void;
 }
@@ -3180,9 +3190,12 @@ export async function toolLoop(opts: ToolLoopOpts): Promise<ToolLoopResult> {
 
     if (stopReason === 'aborted') break;
 
-    // Feed all tool results back as a single user message.
+    // Feed all tool results back as a single user message. Persist it BEFORE
+    // the next dispatch so crash-replay rebuilds a balanced conversation —
+    // without this, resume sends assistant tool-calls with no matching
+    // tool-results and chat() throws "Tool results are missing".
     const userMessageIdx = messageIdx++;
-    void userMessageIdx;
+    await opts.onToolResultTurn?.(turnIdx, userMessageIdx, toolResultBlocks);
     messages.push({ role: 'user', content: toolResultBlocks });
 
     turnIdx++;
--- a/src/core/minions/handlers/subagent.ts
+++ b/src/core/minions/handlers/subagent.ts
@@ -806,6 +806,51 @@ async function runSubagentViaGateway(args: GatewayRunArgs): Promise<SubagentResu
     nextMessageIdx = 1;
   }
 
+  // Crash-replay reconciliation. The gateway loop persists the assistant
+  // tool-call turn (onAssistantTurn) before executing tools and persisting the
+  // tool-result turn (onToolResultTurn). A crash in that window leaves the last
+  // persisted message an assistant turn with unanswered tool-calls, so the first
+  // chat() on resume throws "Tool results are missing". Re-synthesize the
+  // tool-result turn from the settled execution rows (keyed by provider id).
+  // Mirrors the legacy Anthropic path's reconciliation.
+  const lastPrior = priorChatMessages[priorChatMessages.length - 1];
+  if (lastPrior && lastPrior.role === 'assistant' && Array.isArray(lastPrior.content)) {
+    const pendingCalls = lastPrior.content.filter(
+      (b): b is Extract<ChatBlock, { type: 'tool-call' }> =>
+        typeof b === 'object' && b !== null && (b as { type?: unknown }).type === 'tool-call',
+    );
+    if (pendingCalls.length > 0) {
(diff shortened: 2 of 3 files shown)


time-attack:

PR #2277, opened 2026-06-18
feat(ai): claude-cli recipe for native gateway-based subagent dispatch

### LINKED ISSUE #334: feat: Support claude-cli as alternative to Anthropic API for Minions subagent handler
## Problem

The Minions subagent handler (`src/core/minions/handlers/subagent.ts`) hardcodes the Anthropic SDK:

```typescript
import Anthropic from '@anthropic-ai/sdk';
const client: MessagesClient = deps.client ?? makeAnthropic().messages;
```

This requires `ANTHROPIC_API_KEY` with API credits. Users on **Claude Max subscription** (which uses OAuth via `claude-cli`) cannot use Minions without paying additional API costs.

## Proposed Solution

Add support for `claude-cli` as an alternative LLM backend. When `GBRAIN_USE_CLAUDE_CLI=1` is set, inject a `MessagesClient` adapter that shells out to `claude --print` instead of calling the Anthropic API directly.

### Implementation

I've implemented this locally with two changes:

1. **New file: `src/core/minions/handlers/claude-cli-adapter.ts`**
   - Implements `MessagesClient` interface
   - Shells out to `claude --print --model <model> --output-format json`
   - Parses response into Anthropic Message format

2. **Modified: `src/commands/jobs.ts` (~line 628)**
   ```typescript
   let subagentClient: MessagesClient | undefined;
   if (process.env.GBRAIN_USE_CLAUDE_CLI === '1') {
     const { makeClaudeCliClient } = await import('../core/minions/handlers/claude-cli-adapter.ts');
     subagentClient = makeClaudeCliClient();
   }
   worker.register('subagent', makeSubagentHandler({ engine, client: subagentClient }));
   ```

### Benefits

- Claude Max subscribers can use Minions at no additional cost
- No changes required for users with API keys (default behavior unchanged)
- Simple env var toggle

### Limitations of current local patch

- Tool use not yet supported (would need to parse tool_use blocks and handle tool_result injection)
- Token counting is approximate (claude-cli doesn't always return usage stats)

Happy to submit a PR if this approach looks reasonable.

---

**Environment:**
- gbrain 0.17.0 (commit 55ca498)
- macOS, bun runtime
- Claude Max subscription via claude-cli OAuth

### LINKED ISSUE #2276: feat(ai): claude-cli recipe for native gateway-based subagent dispatch — Claude Max subscription without ANTHROPIC_API_KEY
Related: [#334](https://github.com/garrytan/gbrain/issues/334) (jarvisdoes proposed an env-var-gated `MessagesClient` adapter for the same need). This issue describes the same end-user need framed as a recipe so the gateway-loop dispatch path handles claude-cli the same way it handles every other non-Anthropic recipe.

Operators with a Claude Max subscription want subagent dispatch to run against their existing OAuth session via the `claude` CLI binary, not against a separate per-token `ANTHROPIC_API_KEY` billing relationship. Today:

- `agent.use_gateway_loop = true` is the production path for every non-Anthropic recipe.
- There is no claude-cli recipe in the registry, so `models.tier.subagent = claude:claude-sonnet-4-6` (or any CLI-routed alias) has no dispatch path.
- The original proposal in #334 was a `GBRAIN_USE_CLAUDE_CLI=1` env var injecting a custom `MessagesClient` adapter at jobs.ts worker startup. That adapter only runs on the legacy Anthropic-direct path; under the gateway-loop path it never fires.

A native `claude-cli` recipe in `src/core/ai/recipes/` so dispatch goes through the same `gateway.toolLoop()` path every other recipe uses. The recipe:

- Declares `tier: 'native'`, chat-only (no embedding / no expansion).
- Spawns the local `claude` binary with `--print --output-format json --model <X> --disable-slash-commands --system-prompt <gbrain prompt>` from a dedicated tmpdir (skips local `CLAUDE.md` auto-discovery).
- Parses the JSON envelope and extracts `<use_tools>` blocks back into native ai-sdk `tool-call` content parts. Parallel tool calls in one block round-trip cleanly.
- Reports `auth_env.required: []` because the CLI owns auth via its own OAuth session.

Per-call routing then works the same way every other recipe does: `models.dream.patterns = claude-cli:claude-sonnet-4-6` routes through the recipe; `models.tier.subagent = litellm:gpt-5.4` continues through litellm-proxy in the same worker. No env-var switch, no global flag.

**Alternatives considered**

1. **Env-var-gated MessagesClient adapter (the #334 proposal).** Works only on the legacy Anthropic-direct path. Under the gateway-loop path (the production default for every non-Anthropic recipe), the adapter never runs because dispatch flows through `gateway.toolLoop()` not `client.create()` at `subagent.ts:506`. Per-call routing via a proper recipe is the architecture that extends correctly into gateway-loop world.
2. **Anthropic-direct adapter wrapping the `claude` binary.** Same blocker as #1: wrong dispatch path.
3. **Subscriber-only forwarder service (a small local HTTP proxy that authenticates with Claude Max and exposes an OpenAI-shaped API).** Possible but introduces an external service the user has to keep running. The recipe path keeps everything in-process.

**Context isolation knobs**

The recipe spawns from a clean tmpdir + `--disable-slash-commands` + `--system-prompt`, which is the maximum suppression preserving OAuth/Max-subscription auth. `--bare` would also strip `~/.claude/CLAUDE.md` but forces `ANTHROPIC_API_KEY` auth, defeating the whole point. The ~42K cached tokens from user-level instructions stay as an unavoidable trade-off on the subscription path.

Paired PR: #2277. Closes upstream #334.

brettdavies:
Closes #2276, closes #334

## Summary

Operators with a Claude Max subscription can now route subagent dispatch to their existing OAuth session via the local `claude` binary, without an `ANTHROPIC_API_KEY`. After this PR, `gbrain config set models.tier.subagent claude-cli:claude-sonnet-4-6` (or any per-task key like `models.dream.patterns`) dispatches through the gateway-loop the same way every other non-Anthropic recipe does. Per-call routing is preserved: a worker can serve `claude-cli:*` for one job and `litellm:gpt-5.4` for the next, no env-var switch.

## The fix

A native `claude-cli` recipe in the registry. The recipe + its ai-sdk `LanguageModelV2` provider implementation let `gateway.toolLoop()` dispatch to the local `claude` binary directly, parse the JSON envelope, and surface parallel `<use_tools>` blocks as native ai-sdk `tool-call` content parts. End-to-end:

- **Recipe** (`src/core/ai/recipes/claude-cli.ts`): `tier: 'native'`, chat-only (no embedding / no expansion). Models: `claude-opus-4-7` / `claude-sonnet-4-6` / `claude-haiku-4-5-20251001`. `supports_tools: true`, `supports_subagent_loop: true`, `auth_env.required: []` (CLI owns auth via its OAuth session).
- **Provider** (`src/core/ai/providers/claude-cli-language-model.ts`): ai-sdk `LanguageModelV2`. `doGenerate` spawns `claude --print --output-format json --model <X> --disable-slash-commands --system-prompt <gbrain prompt>` from a dedicated tmpdir (skips local `CLAUDE.md` auto-discovery), parses the JSON envelope, extracts `<use_tools>` blocks back into native `tool-call` content parts. Parallel tool calls in one block round-trip cleanly.
- **Registry + gateway wiring** (`src/core/ai/recipes/index.ts`, `src/core/ai/types.ts`, `src/core/ai/gateway.ts`): `claude-cli` joins the `Implementation` union; `instantiateChat` / `instantiateExpansion` / `instantiateEmbedding` gain `case 'claude-cli':` (chat returns the `LanguageModelV2`; expansion mirrors it for a future expansion touchpoint; embedding throws like `native-anthropic` does).

## Why a recipe instead of #334's MessagesClient adapter

The original #334 proposal injected a `MessagesClient` adapter at jobs.ts worker startup behind `GBRAIN_USE_CLAUDE_CLI=1`. That adapter only runs on the legacy Anthropic-direct path. Under `agent.use_gateway_loop = true` (the production default for every non-Anthropic recipe), dispatch flows through `gateway.toolLoop()` not `client.create()` at `subagent.ts:506`, so the adapter never fires.

Per-call routing via a proper recipe is the architecture that extends correctly into gateway-loop world. It also gives the operator a clean per-call routing knob between OAuth-subscription dispatch and API-key dispatch on the same model name, in the same worker, without a process-level switch.

## Context isolation

The recipe spawns from a clean tmpdir + `--disable-slash-commands` + `--system-prompt`, which is the maximum suppression preserving OAuth/Max-subscription auth. `--bare` would also strip `~/.claude/CLAUDE.md` but forces `ANTHROPIC_API_KEY` auth, defeating the whole point. The ~42K cached tokens from user-level instructions stay as an unavoidable trade-off on the subscription path.

## Reproduction (before this PR)

1. Brain with Claude Max subscription and `claude` CLI logged in. No `ANTHROPIC_API_KEY` in env.
2. `gbrain config set models.tier.subagent claude-cli:claude-sonnet-4-6`.
3. `gbrain models` → warns "unrecognized provider" and falls back to `TIER_DEFAULTS.subagent` (`anthropic:claude-sonnet-4-6`).
4. Submit any subagent job → throws `AIConfigError: Anthropic ... requires ANTHROPIC_API_KEY`.

After this PR: step 3 resolves to the claude-cli recipe; step 4 dispatches through the local `claude` binary; the job runs against the OAuth session.

## Tests

- `test/claude-cli-recipe.test.ts`: 16 tests, 46 assertions. Covers recipe registration; spawn args + tmpdir isolation; JSON envelope parse (text-only, tool-call, parallel tool-call, mixed); `<use_tools>` block extraction back to ai-sdk `tool-call` content parts; "model offered tools but declines to use them" case asserting clean text-only return so gateway.toolLoop treats it as a final answer rather than wedging; auth-env declared as empty array.
- Verified end-to-end on a production brain (Claude Max subscription, no `ANTHROPIC_API_KEY`): `gbrain dream --phase patterns` with `models.dream.patterns = claude-cli:claude-sonnet-4-6` runs through parallel tool calls, persists turns, resumes cleanly across worker recycles.

## Coordination with the gateway-loop reliability fix

The gateway-loop tool-result-turn persistence fix (separate PR, separate issue; see `fix(subagent): persist + reconcile tool-result turns in the gateway loop`) is REQUIRED for this recipe to work under load. Without that fix, parallel tool calls dead-letter on any mid-conversation interruption regardless of which recipe is dispatching. Recommend landing that fix first; either order works for the actual code change (this PR's gateway.ts touches are additive only).


loweaxerium:
Ran this recipe in local production against a Claude Max subscription — works, and it's genuinely useful. Four readiness fixes we applied to get it running cleanly, offered as review:

1. **Verbose-array parse (out-of-the-box break)**: with `verbose: true` in `~/.claude/settings.json`, the CLI emits an *array* of events, not the `{type:"result"}` object the parser expects → crash. `--settings '{}'` merges, not replaces, so there's no escape flag. The parser needs to handle both shapes.
2. **Env scrub**: the subprocess inherits the full env — a present `ANTHROPIC_API_KEY` makes the CLI bill the API instead of the subscription, which defeats the PR's stated purpose. Scrub `ANTHROPIC_API_KEY` / `ANTHROPIC_AUTH_TOKEN` / `ANTHROPIC_BASE_URL` before spawn.
3. **Agent isolation**: without `--tools '' --strict-mcp-config` the "model" loads the user's hooks/MCP servers; on a machine running the gbrain MCP itself this recursed. Verified `"tools":[]` / `"mcp_servers":[]` after adding both flags.
4. **stdin error handler**: `stdin.write` with no error listener crashes on ENOENT/EPIPE when the binary is missing.

With these applied: 18/18 recipe tests green, E2E smoke green. Happy to share the diffs if useful.

brettdavies:
Hi @loweaxerium thanks for the feedback. Glad it works for you as well! 

Please feel free to stack your commits on top in this feature branch. 

If you're not sure how to do that, send me your working branch and I'll cherry pick with your authorship preserved.

loweaxerium:
Opened it as a PR against your feature branch so it's a one-click merge (authorship carries over on its own): https://github.com/brettdavies/gbrain/pull/1

Single commit — the four fixes plus two extra tests, all 18 recipe tests green:
- **env scrub** — strips `ANTHROPIC_API_KEY`/`AUTH_TOKEN`/`BASE_URL` from the child env, so the CLI always runs on subscription OAuth and never silently bills an API key
- **agent isolation** — `--tools ""` + `--strict-mcp-config`, prevents recursion when gbrain itself is registered as an MCP server
- **stdin error handler** — `EPIPE`/`ENOENT` on write no longer crashes the worker
- **verbose-array compat** — tolerates the JSON event *array* envelope the CLI emits when `"verbose": true` is set in `~/.claude/settings.json` (no CLI flag disables it; verified on 2.1.145). This one's worth the closest look — it touches the stream parsing path.

brettdavies:
@loweaxerium Great, thank you. I've merged it into this forked branch. Added a couple of additional tests of my own on top.

PR #2279, opened 2026-06-18
fix(dream): drop ANTHROPIC_API_KEY gate from patterns phase

### LINKED ISSUE #2278: dream patterns phase skips with no_api_key on non-Anthropic stacks even though the subagent submission below routes through the gateway
**What happened?**

`gbrain dream --phase patterns` short-circuits with `{ status: 'skipped', reason: 'no_api_key', summary: 'ANTHROPIC_API_KEY unset; pattern detection skipped' }` on any non-Anthropic-routed stack (litellm-proxy / codex-proxy / ollama / openrouter / any openai-compatible deployment), even though the subagent submission immediately below the gate dispatches via the gateway model-tier resolver which already handles non-Anthropic providers correctly.

The gate is `src/core/cycle/patterns.ts:67-69`:

```ts
if (!process.env.ANTHROPIC_API_KEY) {
  return skipped('no_api_key', 'ANTHROPIC_API_KEY unset; pattern detection skipped');
}
```

The sibling synthesize phase already removed its equivalent gate. The `makeJudgeClient` docstring at `src/core/cycle/synthesize.ts:804` documents the rationale: "preserves the legacy 'no ANTHROPIC_API_KEY' cheap-skip semantics" while delegating real auth probing to the gateway recipe's `auth_env.required` machinery, with `AIConfigError` at `gateway.chat()` time caught and surfaced per-call. Patterns carries a similar literal env-var check in its body that never got updated.

**What did you expect?**

`gbrain dream --phase patterns` should dispatch through whatever gateway recipe `models.dream.patterns` (or the applicable tier default) resolves to, exactly as synthesize does. When the gateway IS misconfigured, dispatch should surface a clear `AIConfigError` per-call rather than a silent `skipped: no_api_key` verdict that misclassifies the failure shape.

**Steps to reproduce**

1. Fresh brain with no `ANTHROPIC_API_KEY` in env, paid traffic routed through codex-proxy / litellm / claude-cli / any non-Anthropic recipe.
2. `gbrain config set models.dream.patterns litellm:gpt-5.4` (or any non-`anthropic:*` model).
3. `gbrain dream --phase patterns --json`.
4. Expected: subagent dispatches via the gateway and returns pattern detections, or surfaces `AIConfigError` if the gateway is misconfigured.
5. Actual: `{ status: 'skipped', reason: 'no_api_key', summary: 'ANTHROPIC_API_KEY unset; pattern detection skipped' }` regardless of the resolved tier.

Observed on gbrain 0.42.44.0 through 0.42.51.0.

**Environment**

- gbrain version: 0.42.44.0+ (reproduces against current `master`).
- Bun: 1.3.14.

Not material to the bug. `doctor` reports a healthy brain; the gate fires inside `cycle/patterns.ts` before any dispatch and never reaches the recipe layer.

**Proposed fix**

Drop the gate. The subagent submission immediately below the gate uses the same gateway routing that synthesize already proves works on non-Anthropic stacks. Five-line diff in one file.

Paired PR: #2279.

brettdavies:
Closes #2278, #2782

## Summary

**The bug.** `gbrain dream --phase patterns` returns `{ status: 'skipped', reason: 'no_api_key' }` on any non-Anthropic stack (litellm-proxy, codex-proxy, ollama, openrouter, any openai-compat deployment) regardless of how `models.dream.patterns` resolves. The phase never reaches its subagent submission below, which would have routed through the gateway and dispatched correctly.

**The fix.** Drop the `if (!process.env.ANTHROPIC_API_KEY)` short-circuit in `src/core/cycle/patterns.ts`. The subagent submission below already uses the same gateway model-tier resolver that synthesize uses; when the gateway is misconfigured, dispatch surfaces a clear `AIConfigError` per-call rather than a misclassified `skipped: no_api_key` verdict.

## Diagnosis

- `src/core/cycle/patterns.ts:67-69` (pre-patch) short-circuits when `ANTHROPIC_API_KEY` is unset:

  ```ts   if (!process.env.ANTHROPIC_API_KEY) {     return skipped('no_api_key', 'ANTHROPIC_API_KEY unset; pattern detection skipped');   }   ```

- The submission immediately below the gate calls `submitSubagent(...)`, which routes through `gateway.toolLoop()` via the recipe registered for whichever model the resolver selected. That path already supports every non-Anthropic recipe (litellm, ollama, openrouter, openai, claude-cli, …).
- The sibling synthesize phase removed its equivalent gate (see the `makeJudgeClient` docstring at `src/core/cycle/synthesize.ts:804`, which "preserves the legacy 'no ANTHROPIC_API_KEY' cheap-skip semantics" while delegating real auth probing to the gateway recipe's `auth_env.required` machinery).

The patterns phase carries the same pattern in its body but never had the gate updated when the rest of the cycle moved to gateway-routed dispatch.

## Reproduction

1. Fresh brain with no `ANTHROPIC_API_KEY` in env, paid traffic routed through any non-Anthropic recipe.
2. `gbrain config set models.dream.patterns litellm:gpt-5.4` (or any non-`anthropic:*` model).
3. `gbrain dream --phase patterns --json`.
4. Expected: subagent dispatches and returns pattern detections.
5. Actual (pre-patch): `{ status: 'skipped', reason: 'no_api_key' }`.

## Tests

- Manual verification on a brain with `models.dream.patterns = litellm:gpt-5.4` and no `ANTHROPIC_API_KEY` in env: pre-patch returned the `skipped` verdict; post-patch dispatched through the litellm route and returned pattern detections.
- No new unit test added; the change deletes a literal env-var check and the existing patterns-phase tests cover the dispatch path. Happy to add a regression test that asserts the phase reaches the gateway when `ANTHROPIC_API_KEY` is unset if the maintainer prefers.

## Adjacent observation (follow-up, not in this PR)

Other phases / commands may carry similar stale `ANTHROPIC_API_KEY` checks that pre-date the gateway-routed dispatch. Worth a `rg -n 'ANTHROPIC_API_KEY' src/core/cycle src/commands` sweep to catch siblings. The synthesize phase is already clean; patterns is the one this PR fixes. Anything else found would be a separate small PR per site.

### DIFF
--- a/src/core/cycle/patterns.ts
+++ b/src/core/cycle/patterns.ts
@@ -63,11 +63,12 @@ export async function runPhasePatterns(
       });
     }
 
-    // Submit one subagent for pattern detection.
-    if (!process.env.ANTHROPIC_API_KEY) {
-      return skipped('no_api_key', 'ANTHROPIC_API_KEY unset; pattern detection skipped');
-    }
-
+    // Submit one subagent for pattern detection. The subagent dispatches via
+    // the gateway model-tier resolver, so a hardcoded ANTHROPIC_API_KEY gate
+    // here misclassifies non-Anthropic stacks (litellm/codex-proxy) as
+    // "no upstream" even though synthesize routes through the same gateway
+    // happily. Synthesize already dropped its cheap-skip gate (see
+    // makeJudgeClient docstring); patterns now matches.
     const allowedSlugPrefixes = await loadAllowedSlugPrefixes();
     if (allowedSlugPrefixes.length === 0) {
       return failed(makeError('InternalError', 'NO_ALLOWLIST',


brettdavies:
#2782 re-reports this same bug. Added it to `Closes`. Still the one-line gate removal, ready to land.

time-attack:
Superseded with credit by #2936 (just merged), which absorbs this PR's intent with the post-#2208 end-state: the patterns phase now gates on provider reachability (probeChatModel, same predicate as think/synthesize) instead of ANTHROPIC_API_KEY specifically — non-Anthropic gateway stacks run patterns, keyless installs still skip cleanly (no_provider), and config-file Anthropic keys count. Authorship preserved via Co-authored-by. Thanks — you flagged the right gate to kill.

PR #2284, opened 2026-06-18
feat(dream): synthesize backfill loop hardening — configured zeros, bigint child ids, orchestrator-owned frontmatter, stderr heartbeats

### LINKED ISSUE #2283: dream synthesize backfill loop drops configured zeros, bigint job ids, deterministic frontmatter, and operator visibility on long runs
**What happened?**

Four failure modes in `src/core/cycle/synthesize.ts` surface together during long-running operator-driven backfills (`gbrain dream --phase synthesize --from <day> --to <day>` repeated across many days). Each is a different bug class, but they cluster on the same wait-loop + dual-write path and reproduce together once the loop runs long enough to hit any of them. None reproduces in a short run.

1. **Configured `cooldown_hours = 0` is silently coerced to the 12h default.** The reader uses `parseInt(str, 10) || DEFAULT` for `cooldown_hours` and `min_chars`; `parseInt('0', 10) || 12` evaluates as `0 || 12 = 12` because `0` is falsy. A backfill loop that wants every-cycle synthesis (no cooldown) can't actually disable it. `gbrain config get dream.synthesize.cooldown_hours` returns `0` but the phase still reports `12h cooldown` and skips re-runs accordingly.
2. **Subagent job ids silently mis-look-up in `childMeta`.** postgres-js returns `INT8` as a JS `bigint` by default. The orchestrator builds `childMeta` as `Map<number, ChildMeta>` keyed off `child.id` (a JS `number`); reverse-write reads `subagent_tool_executions.job_id` (bigint) and calls `childMeta.get(jobId)`. `Map.get` uses SameValueZero equality and `2719n !== 2719`, so the lookup silently returns `undefined`; deterministic frontmatter overrides default to `{}`; `buildDeterministicFrontmatter(meta, cycleDate)` is never reached. No error, no warning. Every dual-written page lands with truncated frontmatter (only the unconditional fallbacks `dream_generated` and `dream_cycle_date` survive).
3. **The orchestrator doesn't own deterministic frontmatter on dual-write.** Subagents own their full output frontmatter today including fields the operator wants the orchestrator to stamp authoritatively (`transcript_id`, `transcript_source`, `effective_date`, `date`, `chunk`, `dream_cycle_date`). Subagent drift (typos, omissions, model-side reformatting) leaks into the persisted row and the disk markdown. The orchestrator already has the right values; it just doesn't merge them.
4. **The operator can't see live progress between phase boundaries.** `gbrain dream` runs silently while the synthesize wait loop polls the queue and only logs on state transitions. With `--json` the stdout pipe is silent (envelope only at completion); operators have no way to see "the orchestrator is still working through 47 children" without opening the DB.

**What did you expect?**

Each failure mode has an obvious correct shape:

1. A configured `0` should mean `0`, not be silently replaced with the default. Operators who set the value explicitly to disable a behavior should get that behavior disabled.
2. The bigint-to-number boundary should be defensively coerced at consumer sites so `Map<number, V>` lookups succeed.
3. The orchestrator should re-stamp the deterministic fields after each subagent's `put_page`, so subagent drift can't leak.
4. The orchestrator should emit verbose stderr heartbeats (e.g. `[dream:sync] N children pending, M active, K completed`) every ~20s during the wait, so a tmux pane running `gbrain dream --phase synthesize ... --json > /tmp/log` shows live progress via stderr while stdout stays clean for the JSON consumer.

**Steps to reproduce**

The cluster reproduces during any operator-driven multi-day backfill loop. A short one-day run does not surface issues 2 and 3 because the deterministic frontmatter looks "good enough" from default fallbacks; it takes a few days before the operator notices that `transcript_id` / `effective_date` are missing on every persisted page.

For each failure mode individually:

1. `gbrain config set dream.synthesize.cooldown_hours 0` → run synthesize twice on the same day → second run says "12h cooldown not elapsed" and skips.
2. After any synthesize run with subagents, query the DB for the persisted pages → every page is missing `transcript_id`, `transcript_source`, `effective_date`, `chunk`. Confirmed by adding a `console.log` at the `childMeta.get(jobId)` site: always returns `undefined`.
3. (same root → same observed symptom as issue 2; the orchestrator-frontmatter fix subsumes the bigint fix in practice, but each is required because either alone leaves a different gap.)
4. Run `gbrain dream --phase synthesize --json > /tmp/log` in a tmux pane → stderr is silent for the entire wait period, sometimes 10+ minutes; operator has no signal whether the orchestrator is alive or stuck.

**Environment**

- Bun: 1.3.14.
- Database: Postgres 16 (the bigint mis-lookup is Postgres-specific; PGLite returns numbers for INT8 by default).

**Proposed fix**

Four behavior fixes in `src/core/cycle/synthesize.ts` plus one structural prereq in `src/core/cycle/transcript-discovery.ts`:

1. Replace the falsy-OR config-default pattern with an explicit `Number.isFinite` check via an IIFE for `cooldown_hours` and the adjacent `min_chars` reader.
2. Coerce `jobId` to `Number` at every `childMeta.get()` site (three call sites: `reverseWriteRefs`, `collectChildPutPageSlugs`, the slug-map re-insert).
3. After each subagent's `put_page` lands its row, the orchestrator re-renders the row to disk markdown via `reverseWriteRefs` AND merges the deterministic frontmatter into the DB row via `engine.putPage`. Same writes on both sides so disk and DB stay in lockstep. Subagents still own the content-semantic fields (`type`, `title`, `tags`).
4. Emit `[dream:sync] N children pending, M active, K completed` heartbeats to stderr every ~20s during the wait. Coarse enough not to spam; frequent enough that an operator watching the pane sees liveness.
5. Add a `transcriptSource` field to `DiscoveredTranscript`, derived from the grandparent dir name during discovery (validated against a date-bucket convention so non-bucketed corpora don't trip it). Fix 3 stamps this onto the orchestrator-owned frontmatter; without the field-add it has nothing to write. A parallel transcript-metadata PR also ships this commit; duplicate-commit landing is fine, second PR sees it as a no-op.

Coupling: fix 3 depends on fix 2 because without the bigint coercion, the orchestrator's merge silently drops the deterministic fields when the `childMeta` Map lookup misses. Fix 3 also reads fix 5's `transcriptSource` field. The fixes ship together as one PR because they all reproduce together once the loop runs long enough, and the dependency chain is internal to this PR's scope.

Paired PR: #2284.

brettdavies:
Closes #2283

## Summary

A hardening of `src/core/cycle/synthesize.ts` that closes the cluster of failure modes the operator hits during long-running backfills. The PR is one topic, "the dream synthesize backfill loop's reliability under operator-driven multi-day runs," with four behavior changes plus one structural prereq that fall under it:

- Configured `dream.synthesize.cooldown_hours = 0` and `min_chars = 0` are honored instead of being silently coerced to defaults. Operators who set these to zero to disable the behavior get the behavior disabled.
- The orchestrator's `Map<number, ChildMeta>` lookup is bigint-safe. `postgres-js` returns `INT8` as `bigint`; the lookup now coerces to `Number` at every consumer site so it doesn't silently miss.
- The orchestrator authoritatively stamps deterministic frontmatter (`transcript_id`, `transcript_source`, `effective_date`, `date`, `chunk`, `dream_cycle_date`) onto every dual-written page after each subagent's `put_page` settles. Subagents still own `type` / `title` / `tags` / body; the orchestrator owns only the fields it computes.
- The orchestrator emits `[dream:sync] N children pending, M active, K completed` heartbeats to stderr every ~20s during the wait loop. Operators tailing `gbrain dream --phase synthesize --json > /tmp/log` now see liveness on stderr while stdout stays clean for the JSON consumer.
- Structural prereq: `DiscoveredTranscript` gains a `transcriptSource` field (inferred from the grandparent dir during discovery, validated against a date-bucket convention) so the orchestrator-frontmatter merge has the field to stamp. A parallel PR (transcript-metadata pack) also ships this commit; whichever PR lands second sees it as a no-op (identical patch) and git auto-resolves.

## Before this PR

| Knob / surface                             | Reported state                           | Actual behavior                                                                                                                               |
| ------------------------------------------ | ---------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------- |
| `dream.synthesize.cooldown_hours = 0`      | `gbrain config get` returns `0`          | Phase reports `12h cooldown`, skips re-runs accordingly                                                                                       |
| `dream.synthesize.min_chars = 0`           | Same                                     | Same: the default replaces the configured zero                                                                                                |
| Persisted page frontmatter on Postgres     | "looks like a real page"                 | Missing `transcript_id`, `transcript_source`, `effective_date`, `date`, `chunk` on every page; only the unconditional fallbacks survive       |
| Orchestrator frontmatter merge             | Documented to merge deterministic fields | Silently no-ops on Postgres because `childMeta.get(bigint)` misses (`2719n !== 2719`); on PGLite it works because the engine returns `number` |
| Operator visibility during synthesize wait | None                                     | Stderr silent for entire wait period (10+ min on big backfills)                                                                               |

## After this PR

| Knob / surface                        | Behavior                                                                                             |
| ------------------------------------- | ---------------------------------------------------------------------------------------------------- |
| `dream.synthesize.cooldown_hours = 0` | Disables cooldown; every cycle synthesizes                                                           |
| `dream.synthesize.min_chars = 0`      | Disables the minimum-length skip; every transcript synthesizes                                       |
| Persisted page frontmatter            | Deterministic fields stamped by orchestrator, lockstep with disk markdown; subagent drift can't leak |
| Orchestrator frontmatter merge        | Works identically on Postgres and PGLite                                                             |
| Operator visibility                   | Heartbeat line on stderr every ~20s during the wait                                                  |

## Diagnosis (per behavior change)

**Configured zeros.** The reader uses `parseInt(str, 10) || DEFAULT`. `parseInt('0', 10) || 12` evaluates as `0 || 12 = 12` because `0` is falsy. Replaced with an explicit `Number.isFinite` check via an IIFE so a configured `0` makes it through. Same pattern at the adjacent `min_chars` site.

**Bigint child ids.** The synthesize orchestrator builds `childMeta` as `Map<number, ChildMeta>` keyed off `child.id` (a JS `number`). At reverse-write time the orchestrator reads `subagent_tool_executions.job_id` (INT8) back from postgres-js and looks up via `childMeta.get(jobId)`. postgres-js returns INT8 as `bigint` by default; `Map.get` uses SameValueZero equality, and `2719n !== 2719`, so every lookup silently returns `undefined`, `overrides` defaulted to `{}`, and `buildDeterministicFrontmatter(meta, cycleDate)` was never reached. Fixed by coercing to `Number` at every `Map.get` site (`reverseWriteRefs`, `collectChildPutPageSlugs`, the slug-map re-insert).

**Orchestrator-owned frontmatter.** Subagents previously owned their full output frontmatter, including fields the operator wants the orchestrator to stamp authoritatively. Subagent drift (typos, omissions, model-side reformatting) leaked into the persisted row and the disk markdown. Fixed by re-rendering each row to disk markdown via `reverseWriteRefs` AND merging the deterministic frontmatter into the DB row via `engine.putPage` after the subagent's `put_page` lands. Both writes happen on both sides so disk and DB stay in lockstep. Subagents still own the content-semantic fields.

**Stderr heartbeats.** The wait loop polls the queue at intervals and only logs on state transitions. With `--json` the stdout pipe is silent (envelope only at completion). Added a ~20s stderr heartbeat `[dream:sync] N children pending, M active, K completed` so a tmux pane shows liveness without opening the DB.

## Reproduction

Cluster reproduces during operator-driven multi-day backfills. Per-failure-mode repros:

1. `gbrain config set dream.synthesize.cooldown_hours 0`, run synthesize twice on the same day, observe second run skips with "12h cooldown not elapsed."
2. Run any synthesize cycle against Postgres, query persisted pages, observe missing `transcript_id` / `transcript_source` / `effective_date` / `date` / `chunk` on every page.
3. Same DB query after the patch shows the fields populated and matching what the orchestrator computed.
4. `gbrain dream --phase synthesize --json > /tmp/log` in a tmux pane; stderr silent during wait pre-patch, heartbeats every ~20s post-patch.

## Tests

- `test/cycle-synthesize.test.ts`: existing wait-loop tests continue to pass; new cases cover the configured-zero reads, the bigint `Map.get` coercion, and the orchestrator-merge dual-write.
- Verified end-to-end on a production-shaped brain (Postgres 16) running a multi-day backfill loop: every persisted page lands with full deterministic frontmatter; stderr heartbeats fire on every wait window.

## Why one PR for four changes

All four live in `src/core/cycle/synthesize.ts`. All four reproduce together during the same operator-driven backfill the moment any one surfaces. The orchestrator-frontmatter fix is functionally dependent on the bigint coercion (without the coercion, the merge silently no-ops on Postgres). The configured-zeros and stderr-heartbeat fixes are operator-facing and orthogonal in scope, but they ride with the others because the operator that hits the dual-write bug is the same operator running the long backfill with `cooldown_hours = 0`.

Splitting would create a four-PR chain with an interleaved-merge dependency on shipment order. Bundling them as "the synthesize backfill loop hardening pack" lands the same code with one review surface and one regression test bump.

The synthesize phase currently batches dual-write at end of chunk instead of per-subagent-completion. When the orchestrator dies mid-chunk, every completed subagent's frontmatter merge AND disk write is lost. The per-child flush is a separate architectural change (separate PR), but it would compose cleanly with the orchestrator-owned frontmatter logic in this PR.

### DIFF
--- a/src/core/cycle/synthesize.ts
+++ b/src/core/cycle/synthesize.ts
@@ -651,11 +651,17 @@ async function loadSynthConfig(engine: BrainEngine): Promise<SynthConfig> {
     enabled,
     corpusDir: corpusDir ?? null,
     meetingTranscriptsDir: meetingTranscriptsDir ?? null,
-    minChars: minCharsStr ? Math.max(0, parseInt(minCharsStr, 10) || 2000) : 2000,
+    minChars: (() => {
+      const parsed = parseInt(minCharsStr ?? '', 10);
+      return Number.isFinite(parsed) ? Math.max(0, parsed) : 2000;
+    })(),
     excludePatterns,
     model,
     verdictModel,
-    cooldownHours: cooldownHoursStr ? Math.max(0, parseInt(cooldownHoursStr, 10) || 12) : 12,
+    cooldownHours: (() => {
+      const parsed = parseInt(cooldownHoursStr ?? '', 10);
+      return Number.isFinite(parsed) ? Math.max(0, parsed) : 12;
+    })(),
     maxPromptTokens,
     maxChunksPerTranscript,
   };
--- a/src/core/cycle/synthesize.ts
+++ b/src/core/cycle/synthesize.ts
@@ -506,18 +506,61 @@ export async function runPhaseSynthesize(
 
     // Wait for every child to reach a terminal state. Tick yieldDuringPhase
     // every 5 min so the cycle lock TTL refreshes.
+    //
+    // Verbose progress is emitted to stderr (so `--json` mode still gets a
+    // clean stdout). Without this, an operator running the CLI in a tmux
+    // pane sees nothing between `[cycle.synthesize] start` and the final
+    // JSON payload — for a 50-subagent chunk that can be 15-25 minutes of
+    // silence, which makes it hard to tell a slow chunk apart from a stuck
+    // one. The per-child poll runs serially in childIds order, so emissions
+    // bunch when several children finish during a long wait on an earlier
+    // sibling; a background heartbeat fills the gap with queue snapshots.
+    process.stderr.write(`[dream:sync] orchestrator waiting on ${childIds.length} subagent(s)\n`);
+
+    const phaseStartMs = Date.now();
+    const heartbeat = setInterval(() => {
+      // Best-effort: query the queue for a fresh snapshot, no throw.
+      engine.executeRaw<{ status: string; n: string }>(
+        `SELECT status, COUNT(*)::text AS n
+           FROM minion_jobs
+          WHERE id = ANY($1::bigint[])
+          GROUP BY status`,
+        [childIds],
+      ).then((rows) => {
+        const c: Record<string, number> = { waiting: 0, active: 0, completed: 0, dead: 0, cancelled: 0, failed: 0 };
+        for (const r of rows) {
+          c[r.status] = parseInt(r.n, 10);
+        }
+        const elapsed = Math.round((Date.now() - phaseStartMs) / 1000);
+        process.stderr.write(
+          `[dream:sync] heartbeat (${elapsed}s) waiting=${c.waiting} active=${c.active} ` +
+          `completed=${c.completed} dead=${c.dead} cancelled=${c.cancelled} failed=${c.failed}\n`,
+        );
+      }).catch(() => { /* best-effort, never throw from heartbeat */ });
+    }, 20_000);
+
     const childOutcomes: Array<{ jobId: number; status: string }> = [];
     for (const jobId of childIds) {
+      const startWait = Date.now();
       try {
         const job = await waitForCompletion(queue, jobId, {
           timeoutMs: 35 * 60 * 1000,
           pollMs: 5 * 1000,
         });
         childOutcomes.push({ jobId, status: job.status });
+        const dur = Math.round((Date.now() - startWait) / 1000);
+        process.stderr.write(
+          `[dream:sync] subagent ${childOutcomes.length}/${childIds.length} ${job.status} ` +
+          `(job=${jobId} duration=${dur}s)\n`,
+        );
       } catch (e) {
         if (e instanceof TimeoutError) {
           childOutcomes.push({ jobId, status: 'timeout' });
+          process.stderr.write(
+            `[dream:sync] subagent ${childOutcomes.length}/${childIds.length} TIMEOUT (job=${jobId})\n`,
+          );
         } else {
+          clearInterval(heartbeat);
           throw e;
         }
       }
@@ -527,6 +570,8 @@ export async function runPhaseSynthesize(
       }
     }
 
+    clearInterval(heartbeat);
+
     // Collect slugs from put_page tool executions across the children
     // (codex finding #2: deterministic provenance, NOT pages.updated_at).
     // D6 orchestrator slug rewrite: chunkInfo drives post-hoc rewrite of
@@ -535,6 +580,9 @@ export async function runPhaseSynthesize(
     // v0.32.8: refs carry source_id so reverseWriteRefs picks the correct
     // (source, slug) row (currently always 'default' from subagent put_page).
     const writtenRefs = await collectChildPutPageSlugs(engine, childIds, childMeta);
+    process.stderr.write(
+      `[dream:sync] all subagents settled; dual-writing ${writtenRefs.length} page(s) to disk\n`,
+    );
 
     // Dual-write: reverse-render each DB row → markdown file, with the
     // orchestrator-computed deterministic frontmatter merged in (date,
@@ -548,6 +596,9 @@ export async function runPhaseSynthesize(
       writtenRefs,
       childMeta,
       cycleDate,
+    );
+    process.stderr.write(
+      `[dream:sync] dual-write complete (${reverseWriteCount} disk page(s))\n`,
     );
 
     // Summary index page (deterministic; orchestrator-written via direct
--- a/src/core/cycle/synthesize.ts
+++ b/src/core/cycle/synthesize.ts
@@ -1068,11 +1068,17 @@ async function collectChildPutPageSlugs(
   const rewritten = new Map<string, number>();
   for (const r of rows) {
     if (typeof r.slug !== 'string' || r.slug.length === 0) continue;
-    const meta = childMeta.get(r.job_id);
+    // postgres.js returns `job_id` as bigint; coerce to Number so the
+    // childMeta keys (set from queue.add's number-typed child.id) actually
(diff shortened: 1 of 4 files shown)


time-attack:
Thanks for the detailed before/after table — the `cooldown_hours = 0` coercion is a legitimate complaint. But this bundles that with three behavior/feature changes (orchestrator taking frontmatter ownership from subagents, a new stderr heartbeat channel, a new `transcriptSource` discovery field), and it substantially overlaps your own #2286, which ships the same commits against the same two files. That combination is more new surface than we're taking in the dream phase right now, so I'm closing it. If the configured-zeros coercion alone is blocking your runs, a one-file fix for just that would be a much narrower ask.

PR #2286, opened 2026-06-18
feat(dream): orchestrator-owned transcript metadata + transcriptSource discovery field

### LINKED ISSUE #2285: dream transcript metadata: discovery exposes no source label, date inference fights with content metadata, orchestrator does not own deterministic frontmatter
**What happened?**

Three coupled gaps in how `gbrain dream --phase synthesize` carries transcript metadata from discovery through the dual-write. All three reproduce together on any brain that ingests transcripts from multiple sources or whose file timestamps drift relative to conversation start.

1. **Transcript discovery infers source from the grandparent dir but never exposes it.** A transcript at `~/.gbrain/transcripts/claude-code/2026-06-12/<file>.md` is clearly a claude-code transcript and a transcript at `~/.gbrain/transcripts/voice-notes/2026-06-12/<file>.md` is clearly a voice note, but `DiscoveredTranscript` carries no `transcriptSource` field. Downstream consumers (the orchestrator's deterministic frontmatter merge, any retrieval surface that wants to filter by ingestion pipeline) have no clean way to ask "show me transcripts ingested from claude-code vs from voice notes."
2. **Transcript date inference uses file mtime, which drifts.** Mtime is re-stamped by every sync tool that touches the file (rsync, Dropbox, Syncthing, B2). A transcript whose conversation started 2026-05-15 but whose file was last `mv`d in the morning of 2026-06-17 lands with `effective_date: 2026-06-17`, dropping it into the wrong cycle window. Claude-code transcripts ship a `## Metadata` table with a `| First message | 2026-05-15T03:51:11.584Z |` row that is stable across copies and re-syncs and would give the correct date directly.
3. **The orchestrator does not own deterministic frontmatter on dual-write.** Subagents own their full output frontmatter today, including the fields the operator wants the orchestrator to stamp authoritatively (`transcript_id`, `transcript_source`, `effective_date`, `date`, `chunk`, `dream_cycle_date`). Subagent drift (typos, omissions, model-side reformatting) leaks into the persisted row and the disk markdown. The orchestrator already has the right values; it just doesn't merge them.

**What did you expect?**

A unified metadata pipeline for transcripts:

- `DiscoveredTranscript.transcriptSource` populated from the grandparent dir name during discovery.
- Date inference reads the first-message timestamp from the transcript body when available and falls back to file mtime when absent. Stable across copies and re-syncs.
- The orchestrator merges `transcript_id` / `transcript_source` / `effective_date` / `date` / `chunk` / `dream_cycle_date` onto every dual-written page authoritatively. Subagents still own `type` / `title` / `tags` and the body content.

**Steps to reproduce**

1. Ingest two transcripts under different source dirs: `~/.gbrain/transcripts/claude-code/2026-06-12/<a>.md` and `~/.gbrain/transcripts/voice-notes/2026-06-12/<b>.md`.
2. Run `gbrain dream --phase synthesize` against both. Inspect the resulting pages: `transcript_source` is empty on both because discovery never exposed it.
3. Touch the file (`touch ~/.gbrain/transcripts/claude-code/2026-06-12/<a>.md`). Re-run synthesize. The page's `effective_date` is now today, not the original conversation date.
4. Inspect any dual-written page: `transcript_id`, `transcript_source`, `effective_date`, `date`, `chunk` are missing or drift from the orchestrator's computed values.

**Environment**

- Bun: 1.3.14.

**Proposed fix**

Three coupled changes, all in `src/core/cycle/transcript-discovery.ts` + `src/core/cycle/synthesize.ts`:

1. Add `transcriptSource: string` to `DiscoveredTranscript`. Populate from `basename(dirname(dirname(filePath)))` during discovery so the source label survives into the downstream pipeline.
2. Infer the transcript date by parsing the `## Metadata` `| First message |` row inside the body when present; fall back to file mtime otherwise. Preserve upstream's existing filename-regex hint as an intermediate fallback so brains that name their transcripts by date keep working.
3. After each subagent's `put_page` lands, the orchestrator merges deterministic fields into the DB row via `engine.putPage` AND re-renders the row to disk via `reverseWriteRefs`. Subagents still own content fields; the orchestrator owns metadata.

Paired PR: #2286.

brettdavies:
Closes #2285

## Summary

The dream synthesize phase now carries transcript metadata end-to-end without subagent drift and without dropping the conversation's actual date when the file gets re-synced. Three coupled changes that all live in transcript discovery + the synthesize orchestrator:

- `DiscoveredTranscript` carries a `transcriptSource` label (`claude-code`, `voice-notes`, etc.) inferred from the grandparent dir during discovery, so downstream consumers can filter by ingestion pipeline.
- Transcript date inference uses the `## Metadata` `| First message |` row from the transcript body when available, falling back to the filename-regex date and then to file mtime. Stable across `rsync` / Dropbox / Syncthing / B2 re-syncs that re-stamp mtime.
- The orchestrator stamps deterministic frontmatter (`transcript_id`, `transcript_source`, `effective_date`, `date`, `chunk`, `dream_cycle_date`) onto every dual-written page after each subagent's `put_page` settles. Subagents still own `type` / `title` / `tags` / body; only the metadata fields are owned by the orchestrator.

## Before this PR

| Surface                                 | Behavior                                                                                        |
| --------------------------------------- | ----------------------------------------------------------------------------------------------- |
| `DiscoveredTranscript.transcriptSource` | Does not exist; no clean way to filter by ingestion source                                      |
| Transcript date                         | Inferred from file mtime; drifts when sync tools re-stamp                                       |
| Persisted page frontmatter              | Subagent owns full frontmatter; orchestrator drift (typos, omissions, model reformatting) leaks |

## After this PR

| Surface                                 | Behavior                                                                                                                                                     |
| --------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| `DiscoveredTranscript.transcriptSource` | Populated from `basename(dirname(dirname(filePath)))` during discovery; available throughout the downstream pipeline                                         |
| Transcript date                         | Content-metadata wins when the `## Metadata` first-message-row is present; filename-regex applies next; mtime is the final fallback. Stable across re-syncs. |
| Persisted page frontmatter              | Orchestrator stamps deterministic fields authoritatively, lockstep across DB and disk; subagent drift on those fields can't leak                             |

## Diagnosis (per change)

**transcriptSource field.** Discovery has the information (the dir name immediately above `<date>/<file>.md` is the source label by convention) but `DiscoveredTranscript` doesn't expose it. Adding the field + the `basename(dirname(dirname(filePath)))` extraction makes it available to every consumer that already destructures the discovered transcript object.

**Content-based date inference.** Claude-code transcripts include a `## Metadata` table with a `| First message | 2026-05-15T03:51:11.584Z |` row. That timestamp is stable across copies and re-syncs because every sync tool that re-stamps mtime preserves file content. Parsing the row when present gives a date that matches the operator's mental model ("when did this conversation happen?") rather than the file-on-disk's metadata fingerprint. Upstream's existing filename-regex date inference at `transcript-discovery.ts:~261` is preserved as the intermediate fallback so brains that name transcripts by date keep working without behavior change.

**Orchestrator-owned frontmatter.** Subagents previously owned full frontmatter, including the fields the orchestrator computes deterministically. Drift (typos, omissions, model-side reformatting) leaked into the persisted row and the disk markdown. Fixed by re-rendering each row to disk markdown via `reverseWriteRefs` AND merging the deterministic frontmatter into the DB row via `engine.putPage` after the subagent's `put_page` lands.

## Reproduction

1. Ingest two transcripts under different source dirs: `~/.gbrain/transcripts/claude-code/2026-06-12/<a>.md` and `~/.gbrain/transcripts/voice-notes/2026-06-12/<b>.md`. Run synthesize; pre-patch the resulting pages have no `transcript_source` field.
2. `touch ~/.gbrain/transcripts/claude-code/2026-06-12/<a>.md` between syncs. Re-run synthesize; pre-patch the `effective_date` is today's date rather than the original conversation date. Post-patch the page reads the `## Metadata` `| First message |` row and keeps the conversation date.
3. Inspect any dual-written page; pre-patch the deterministic fields are missing or drift from what the orchestrator computed; post-patch they're populated and matching.

## Tests

- `test/cycle-synthesize.test.ts`: new cases for the transcriptSource field, the content-date parser, the orchestrator merge.
- `test/cycle/transcript-discovery.test.ts`: new cases for the date inference cascade (content wins, filename fallback, mtime fallback).
- Verified end-to-end on a brain that ingests both claude-code and voice-note sources: every persisted page lands with the correct `transcript_source` label and a date that matches the conversation rather than the file metadata.

## Coordination with the synthesize hardening pack

The `feat(dream): synthesize backfill loop hardening` PR (separate branch `feat/dream-synthesize-hardening`) also ships the orchestrator-owned frontmatter commit. Per "duplicate commits across branches are fine, the second PR's duplicate becomes a no-op," whichever PR merges second sees that commit auto-resolve. Both PRs are defensible as single-topic: this PR is "transcript metadata pipeline," the other is "synthesize backfill loop hardening."

## Why one PR for three changes

All three live in the transcript-discovery → synthesize dual-write path. All three reproduce together on any brain that ingests transcripts from multiple sources or whose file timestamps drift relative to conversation start. Splitting would create a three-PR chain with an interleaved-merge dependency on shipment order (the orchestrator merge needs `transcriptSource` to stamp it). Bundling them as "the transcript metadata pipeline" lands the same code with one review surface.

### DIFF
--- a/src/core/cycle/synthesize.ts
+++ b/src/core/cycle/synthesize.ts
@@ -405,8 +405,14 @@ export async function runPhaseSynthesize(
 
     const queue = new MinionQueue(engine);
     const childIds: number[] = [];
-    /** Map child job_id → chunk metadata for D6 orchestrator-side slug rewrite. */
-    const chunkInfo = new Map<number, { idx: number; hash6: string }>();
+    /**
+     * Map child job_id → transcript metadata. Drives D6 orchestrator-side slug
+     * rewrite for chunked transcripts AND the deterministic frontmatter the
+     * dual-write builds at reverse-render time. Populated for every child
+     * (single-chunk children carry chunkTotal=1; chunked children carry
+     * idx/total per chunk).
+     */
+    const childMeta = new Map<number, ChildMeta>();
     /** Skip reasons for the cycle report (D5 cap hits, D8 legacy-key skips). */
     const skipReports: Array<{ filePath: string; reason: string }> = [];
 
@@ -487,9 +493,14 @@ export async function runPhaseSynthesize(
           { allowProtectedSubmit: true },
         );
         childIds.push(child.id);
-        if (isChunked) {
-          chunkInfo.set(child.id, { idx: i, hash6 });
-        }
+        childMeta.set(child.id, {
+          idx: i,
+          hash6,
+          chunkTotal: chunks.length,
+          transcriptSource: t.transcriptSource,
+          transcriptId: stripContentVersionSuffix(t.basename),
+          inferredDate: t.inferredDate,
+        });
       }
     }
 
@@ -523,10 +534,21 @@ export async function runPhaseSynthesize(
     // even if Sonnet drops the chunk suffix.
     // v0.32.8: refs carry source_id so reverseWriteRefs picks the correct
     // (source, slug) row (currently always 'default' from subagent put_page).
-    const writtenRefs = await collectChildPutPageSlugs(engine, childIds, chunkInfo);
+    const writtenRefs = await collectChildPutPageSlugs(engine, childIds, childMeta);
 
-    // Dual-write: reverse-render each DB row → markdown file.
-    const reverseWriteCount = await reverseWriteRefs(engine, opts.brainDir, writtenRefs);
+    // Dual-write: reverse-render each DB row → markdown file, with the
+    // orchestrator-computed deterministic frontmatter merged in (date,
+    // effective_date, transcript_source, transcript_id, transcript_hash,
+    // chunk, dream_generated, dream_cycle_date). Subagent retains
+    // authority over title and tags; everything else is computed.
+    const cycleDate = today();
+    const reverseWriteCount = await reverseWriteRefs(
+      engine,
+      opts.brainDir,
+      writtenRefs,
+      childMeta,
+      cycleDate,
+    );
 
     // Summary index page (deterministic; orchestrator-written via direct
     // engine.putPage so no allow-list path needed).
@@ -1010,8 +1032,8 @@ function sanitizeForSlug(s: string): string {
 async function collectChildPutPageSlugs(
   engine: BrainEngine,
   childIds: number[],
-  chunkInfo: Map<number, { idx: number; hash6: string }>,
-): Promise<Array<{ slug: string; source_id: string }>> {
+  childMeta: Map<number, ChildMeta>,
+): Promise<Array<{ slug: string; source_id: string; jobId: number }>> {
   if (childIds.length === 0) return [];
   // Raw fetch — NO SELECT DISTINCT. Preserves per-child slug duplicates so
   // the orchestrator sees what each child wrote. COALESCE handles both
@@ -1024,6 +1046,10 @@ async function collectChildPutPageSlugs(
   // product behavior. Threading the source_id through reverseWriteRefs
   // guarantees getPage targets the correct (source, slug) row instead of
   // the first DB match.
+  //
+  // Also threads job_id so reverseWriteRefs can pair each slug back to the
+  // childMeta entry that produced it, which drives the deterministic
+  // frontmatter the dual-write injects on top of the subagent's row.
   const rows = await engine.executeRaw<{ job_id: number; slug: string }>(
     `SELECT job_id,
             COALESCE(input->>'slug', (input #>> '{}')::jsonb->>'slug') AS slug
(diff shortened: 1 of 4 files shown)


time-attack:
Thanks for this — the mtime-drift problem with rsync/Syncthing re-stamping is a real observation. That said, all three changes here are new behavior in the dream synthesize phase rather than fixes to something broken, and they overlap heavily with your own #2284 (same two source files, same two test files, and #2284 explicitly ships the transcriptSource commit too). We're not taking new surface in this area, so I'm closing this one along with #2284. Appreciate the care in the write-up.

PR #2288, opened 2026-06-18
Fix op checkpoint JSONB array binding

EagleEyez1:
## Summary

Fixes a live Postgres sync failure where `recordCompleted()` wrote `completed_keys` as a JSON string instead of a JSON array through a `postgres.js` `unsafe` placeholder.

In production this surfaced as:

```text
new row for relation "op_checkpoints" violates check constraint "op_checkpoints_completed_keys_array"
```

The database constraint was correct; the writer was double-encoding the value by passing `JSON.stringify(sorted)` to `$3::jsonb`.

## Change

- Pass the JavaScript `string[]` directly to the `$3::jsonb` placeholder.
- Keep the defensive fallback as an empty JavaScript array.
- This preserves the v119 `completed_keys` array-shape guard and lets postgres.js serialize the JSONB value correctly.

## Validation

```text
bun test test/op-checkpoint*.test.ts
37 pass
0 fail
```

Also validated against the production Artie brain after the same fix:

- `gbrain sync --no-pull --no-embed` completed successfully after previously aborting before import.
- `gbrain embed --stale` restored missing embeddings to 0.
- Retrieval gate remained healthy.

### DIFF
--- a/src/core/op-checkpoint.ts
+++ b/src/core/op-checkpoint.ts
@@ -180,9 +180,12 @@ export async function recordCompleted(
   // extract-conversation-facts serialize a MUTABLE map through here and rely on
   // stale keys being REMOVED; an append would make them unremovable. The full
   // set lands in the parent `completed_keys` JSONB column via a single UPSERT —
-  // exactly as before. JSON.stringify into `$3::jsonb` is correct (the text→jsonb
-  // cast yields a proper array; NOT the double-encode trap, which is the template
-  // form). Sync uses `appendCompleted` (below) instead, never this.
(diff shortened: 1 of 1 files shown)


garrytan-agents:

Thanks for the report/fix on `op_checkpoints jsonb array binding`. Verified fixed on 0.42.53.0.

garrytan:

Thank you @EagleEyez1 for the independent fix — several of you found this same bug in parallel. Crediting contributors in the v0.42.53.0 release notes. Closing as superseded.

PR #2292, opened 2026-06-18
Fix op checkpoint JSONB binding

uncfreak1255-code:
## Summary
- Fix `recordCompleted` so `op_checkpoints.completed_keys` is written as a real JSONB array on Postgres, not a double-encoded scalar string.
- Add PGLite and real-Postgres regression coverage for the completed_keys shape.
- Tighten the JSONB static guard for `executeRawDirect` and fix the admin build gate that was timing out in verify.

## Proof
- `bun run verify` -> 30/30 checks green
- `bun test test/op-checkpoint.test.ts` -> 31 pass, 0 fail
- Fresh Postgres `pgvector/pgvector:pg16` E2E: `DATABASE_URL=postgresql://postgres:postgres@localhost:5435/gbrain_test bun test test/e2e/op-checkpoint-jsonb-postgres.test.ts` -> 1 pass, 0 fail
- Direct Postgres readback: `array|chunk-a|2`

## Notes
Compared with upstream PR #2288, this keeps the fix narrow but avoids passing a bare top-level JS array into `$3::jsonb`; the value is wrapped in an object and extracted server-side with `($3::jsonb)->'completed_keys'`.


uncfreak1255-code:
Closing this one because it was opened from the fork master lineage and includes unrelated fork history. I am rebuilding the branch from upstream/master and will reopen a narrow PR.

PR #2293, opened 2026-06-18
Fix op checkpoint JSONB binding

uncfreak1255-code:
## Summary
- Fix `recordCompleted` so `op_checkpoints.completed_keys` is written as a real JSONB array on Postgres, not a JSON string scalar.
- Add focused PGLite coverage plus a real Postgres E2E regression for the postgres.js direct binding path.
- Narrow the JSONB pattern guard to this checkpoint regression and fix the admin build check so local verify does not hang behind the Bun wrapper.

## Upstream comparison
Compared against #2288: same root cause, but this branch binds a wrapper object and extracts `completed_keys` server-side via `($3::jsonb)->'completed_keys'`. That keeps the existing direct-pool write path while avoiding `JSON.stringify(sorted)` double-encoding.

The first PR I opened from this machine (#2292) was closed because it was based on the fork master lineage and included unrelated history. This is the rebuilt upstream-based PR.

## Proof receipts
- `bun run verify` -> `[verify-parallel] elapsed=95s | pass=30 fail=0 | all checks green`
- `bun test test/op-checkpoint.test.ts` -> `31 pass, 0 fail`
- `scripts/check-jsonb-pattern.sh` -> all three JSONB/max_stalled checks OK
- Fresh `pgvector/pgvector:pg16` Postgres E2E: `DATABASE_URL=postgresql://postgres:postgres@localhost:5435/gbrain_test bun test test/e2e/op-checkpoint-jsonb-postgres.test.ts` -> `1 pass, 0 fail`
- Direct Postgres readback: `SELECT jsonb_typeof(completed_keys), completed_keys->>0, jsonb_array_length(completed_keys) ...` -> `array|chunk-a|2`

### DIFF
--- a/admin/src/pages/Agents.tsx
+++ b/admin/src/pages/Agents.tsx
@@ -381,8 +381,15 @@ function CredentialsModal({ credentials, onClose }: {
   );
 }
 
+type ConfigExportTab = 'claude-code' | 'chatgpt' | 'claude-cowork' | 'perplexity' | 'cursor' | 'json';
+const OAUTH_ONLY_TAB_LABELS: Partial<Record<ConfigExportTab, string>> = {
+  chatgpt: 'ChatGPT',
+  'claude-cowork': 'Claude.ai',
+  perplexity: 'Perplexity',
+};
+
 function AgentDrawer({ agent, onClose, onRevoked }: { agent: Agent; onClose: () => void; onRevoked: () => void }) {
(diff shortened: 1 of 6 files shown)


garrytan-agents:

Thanks for the report/fix on `op_checkpoint jsonb binding`. Verified fixed on 0.42.53.0.

PR #2299, opened 2026-06-19
feat(code-intel): MQL (MQL4/5) + C++ call-graph code intelligence

immy2good:
## What this delivers

Call-graph **code intelligence for MQL4/MQL5** (MetaTrader) plus the enabling **C++ receiver-type resolution** — `code_callers` / `code_callees` / `code_blast` / `code_flow` over a real MQL corpus. MQL has no dedicated tree-sitter grammar and, as far as we can find, no call-graph tooling anywhere; tagging `.mq4/.mq5/.mqh` as `mql` over the already-shipped tree-sitter-cpp grammar gives gbrain that coverage at zero grammar-maintenance cost.

## Self-contained on fresh checkout

Rebased on current `master` and works **end-to-end**, not just at the unit level:

- `.mq4/.mq5/.mqh` are added to `CODE_EXTENSIONS`, so `gbrain sync --strategy code` actually routes MetaTrader files to the code chunker (guarded by a regression test in `sync-classifier-widening.test.ts`). Without it, sync silently dropped every MQL file.
- Coverage is honest on both ends: **admission** (the classifier test) + **edge extraction** (the `importCodeFile` → `getCallersOf`/`getCalleesOf` tests with committed MQL fixtures, so edge counts are reproducible in CI).

## Receiver-type resolution

Explicit and implicit receivers resolve to the receiver's declared class type, so the graph spans files:

- implicit-self (`m()`, `this->m()`) → enclosing class (exactly-one-declaration rule)
- out-of-line (`C::m(){...}`) → class named by the definition's qualified scope
- member-field / local-object (`m_x.Method()`, `obj.Method()`) → the field/local's declared class type
- **file-scope global objects** (`CDash g_dashboard; … g_dashboard.M()`) → the global's declared class type. Real MQL indicators wire their entry layer (`OnInit`/`OnCalculate`/`OnChartEvent`/`OnTimer`) to class internals through global singletons — this is what connects that boundary to the graph. (`mql-global-receiver-callers.test.ts`)
- **Honest precision boundary:** genuine globals/library calls, overloads, and namespaced/reference-typed receivers stay **bare** — no confident false edges.

## Free-function entry handlers are first-class call-graph nodes

[...]

PR #2309, opened 2026-06-19
fix(sync): op_checkpoints completed_keys double-encode blocks all sync

### LINKED ISSUE #2305: sync always fails: recordCompleted writes completed_keys as a jsonb string (double-encode), violating op_checkpoints_completed_keys_array
## Summary

On a Postgres/Supabase brain (v0.42.51), **every `gbrain sync` aborts before importing anything**: the `sync-target` checkpoint write violates the `op_checkpoints_completed_keys_array` CHECK constraint. Root cause — `recordCompleted()` binds `JSON.stringify(keys)` to a `$N::jsonb` parameter via `executeRawDirect` (`conn.unsafe`), which (with `{ prepare: false }`) lands as a jsonb **string scalar** (`"[\"...\"]"`), **not** a jsonb **array**. The CHECK requires `jsonb_typeof(completed_keys) = 'array'`, so the write is rejected and sync can never bank a checkpoint.

## Impact

- All sync/ingestion blocked (0 files imported, `last_commit` never advances; "safe to retry" but every retry hits the same).
- Observed on v0.42.51.0, Postgres backend behind Supabase Supavisor (pooler), `{ prepare: false }`.

## Error

```
[op-checkpoint] write failed (sync-target, <fp>): new row for relation "op_checkpoints"
[sync] checkpoint target write failed — aborting before import; nothing drained, next run retries.
Sync PARTIAL at <commit>: imported 0 of N file(s), reason=checkpoint_unavailable.
```

## Root cause + minimal reproduction

`recordCompleted` (`src/core/op-checkpoint.ts:174`):

```ts
const sorted = [...keys].sort();
engine.executeRawDirect(
   VALUES ($1, $2, $3::jsonb, now()) ON CONFLICT (op, fingerprint) DO UPDATE
     SET completed_keys = EXCLUDED.completed_keys, updated_at = now()`,
  [key.op, key.fingerprint, JSON.stringify(sorted)],   // <-- JSON.stringify + $3::jsonb
);
```

`executeRawDirect` → `runUnsafe` → `conn.unsafe(sql, params)` (`src/core/postgres-engine.ts:5158`). The in-code comment asserts this is safe ("JSON.stringify into `$3::jsonb` is correct … NOT the double-encode trap, which is the template form"). **It is not, with `.unsafe()` + `{ prepare: false }`.** Minimal repro on the same engine:

```js
const sql = postgres(DATABASE_URL, { prepare: false });
const param = JSON.stringify(["6bf9661b"]);
await sql.unsafe("select jsonb_typeof($1::jsonb) as t, $1::jsonb as v", [param]);
// => [{ t: "string", v: "[\"6bf9661b\"]" }]      // jsonb STRING scalar, NOT an array
```

So `$3::jsonb` receives the JSON-encoded *string* and parses it as a jsonb scalar string → `jsonb_typeof = 'string'` → violates the `'array'` CHECK.

Failing path: `src/commands/sync.ts:1956` → `recordCompleted(engine, ckpt.target, [pin])` (documented intent at `sync.ts:96`: `op 'sync-target' -> completed_keys = [pinnedTargetCommit]`).

## Why it surfaces now

The `op_checkpoints_completed_keys_array` CHECK (`jsonb_typeof = 'array'`) was added in the v0.42.x (#1794) "checkpoint integrity" work (`src/core/migrate.ts:5332-5365`), and checkpoint writes were routed to the direct session pool (`executeRawDirect`) with `{ prepare: false }`. Before the CHECK, the malformed jsonb-string value inserted silently (the loader at `op-checkpoint.ts:120-156` even carries defensive handling for non-array `completed_keys`). The CHECK now correctly rejects it, turning the latent double-encode into a hard, total sync failure.

## Suggested fix

- Bind the JS value directly and let postgres.js encode it as jsonb, instead of `JSON.stringify(...)` + `::jsonb`: pass `sorted` (the JS array) as the param (postgres.js serializes JS arrays/objects to jsonb), or use postgres.js's `sql.json(sorted)` helper; drop the manual `JSON.stringify` and the `$3::jsonb` cast.
- Add a regression test asserting `jsonb_typeof(completed_keys) = 'array'` after `recordCompleted`.
- Audit other `JSON.stringify(x)` values bound to `::jsonb` params via `.unsafe()` for the same trap.

## Environment

gbrain 0.42.51.0 · Postgres (Supabase) backend · Supavisor pooler · `{ prepare: false }`.

immy2good:
## Problem

`recordCompleted` (`src/core/op-checkpoint.ts`) blocks **all** `gbrain sync` on Postgres. It binds `JSON.stringify(keys)` to a `$3::jsonb` parameter via `executeRawDirect` → `conn.unsafe(sql, params)` with `{ prepare: false }`. postgres.js v3 sends the JSON string as a text param, so `$3::jsonb` parses it into a jsonb **string scalar** (`"[\"k\"]"`), not an array. The `op_checkpoints_completed_keys_array` CHECK (`jsonb_typeof = 'array'`) then rejects every `sync-target` checkpoint write:

```
[op-checkpoint] write failed (sync-target, <fp>): new row for relation "op_checkpoints"
[sync] checkpoint target write failed — aborting before import; nothing drained.
Sync PARTIAL: imported 0 of N file(s), reason=checkpoint_unavailable.
```

Result: 0 files imported, `last_commit` never advances; "safe to retry" but every retry hits the same. Fixes #2305.

This is the **parameterized twin** of the `${JSON.stringify(x)}::jsonb` double-encode that `scripts/check-jsonb-pattern.sh` guards against — the grep guard only catches the template form, so this slipped through. The in-code comment even claimed the parameterized form was safe ("NOT the double-encode trap"); it isn't, under `.unsafe()` + `{ prepare: false }`.

Minimal repro:

```js
const sql = postgres(DATABASE_URL, { prepare: false });
await sql.unsafe("select jsonb_typeof($1::jsonb) as t", [JSON.stringify(["x"])]);
// => [{ t: "string" }]   // jsonb STRING scalar, not an array
```

## Fix

Bind the JS `string[]` as a Postgres `text[]` and build the jsonb in SQL via `to_jsonb($3::text[])` — the same native binding `appendCompleted` already relies on (works on both Postgres and PGLite). Drops the `JSON.stringify` + `$3::jsonb`.

## Test

Adds an e2e regression test in `test/e2e/postgres-jsonb.test.ts` asserting `recordCompleted` stores a jsonb **array** (`jsonb_typeof = 'array'`). The bug only manifests on real Postgres (PGLite uses a different driver path and hides it), so the test lives in the Postgres e2e suite.

Verified RED→GREEN against a real Postgres; existing PGLite op-checkpoint unit suite passes (30/0); `tsc --noEmit` clean.

---
**Related contributions** (from the iTradeAIMS fork; reviewed cross-model before opening):
- #2309 — `op_checkpoints` `completed_keys` double-encode sync fix (this PR; fixes #2305)
- #2299 — MQL4/5 + C++ call-graph code intelligence (feature)

### DIFF
--- a/src/core/op-checkpoint.ts
+++ b/src/core/op-checkpoint.ts
@@ -179,19 +179,26 @@ export async function recordCompleted(
   // REPLACE semantics (kept deliberately — #1794 V3). Callers like
   // extract-conversation-facts serialize a MUTABLE map through here and rely on
   // stale keys being REMOVED; an append would make them unremovable. The full
-  // set lands in the parent `completed_keys` JSONB column via a single UPSERT —
-  // exactly as before. JSON.stringify into `$3::jsonb` is correct (the text→jsonb
-  // cast yields a proper array; NOT the double-encode trap, which is the template
-  // form). Sync uses `appendCompleted` (below) instead, never this.
+  // set lands in the parent `completed_keys` JSONB column via a single UPSERT.
+  //
+  // #2305: bind the JS string[] as a Postgres text[] and build the jsonb in SQL
+  // via `to_jsonb($3::text[])`. The previous `JSON.stringify(sorted)` + `$3::jsonb`
+  // form is the *parameterized* twin of the `${JSON.stringify(x)}::jsonb`
+  // double-encode trap: postgres.js (`conn.unsafe`, prepare:false) sends the JSON
+  // string as a text param, so `$3::jsonb` parses it into a jsonb STRING scalar
+  // (`"[\"k\"]"`), not an array — which the `op_checkpoints_completed_keys_array`
+  // CHECK then rejects, failing every sync-target checkpoint write. The text[] →
+  // to_jsonb binding is the same one `appendCompleted` already relies on (native
+  // on both Postgres and PGLite). Sync uses `appendCompleted` (below), not this.
   const sorted = [...keys].sort();
   return durableWrite(engine, key, 'write', () =>
     engine.executeRawDirect(
       `INSERT INTO op_checkpoints (op, fingerprint, completed_keys, updated_at)
-       VALUES ($1, $2, $3::jsonb, now())
+       VALUES ($1, $2, to_jsonb($3::text[]), now())
        ON CONFLICT (op, fingerprint) DO UPDATE
          SET completed_keys = EXCLUDED.completed_keys,
              updated_at     = now()`,
-      [key.op, key.fingerprint, JSON.stringify(sorted)],
+      [key.op, key.fingerprint, sorted],
     ));
 }
 
--- a/test/e2e/postgres-jsonb.test.ts
+++ b/test/e2e/postgres-jsonb.test.ts
@@ -26,6 +26,7 @@ import { describe, test, expect, beforeAll, afterAll } from 'bun:test';
 import {
   hasDatabase, setupDB, teardownDB, getEngine, getConn,
 } from './helpers.ts';
+import { recordCompleted } from '../../src/core/op-checkpoint.ts';
 
 const skip = !hasDatabase();
 const describeE2E = skip ? describe.skip : describe;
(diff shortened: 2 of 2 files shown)


immy2good:
**Impact scope.** This isn't a marginal fix — the `completed_keys` double-encode blocks **100% of Postgres sync** (0 files imported, `last_commit` never advances, every retry re-fails the same checkpoint write), so any Postgres-backed gbrain is stuck until it lands. PGLite hides it (different driver path), which is why it slipped past.

It also surfaces a guard gap: `scripts/check-jsonb-pattern.sh` catches the template-literal `${JSON.stringify(x)}::jsonb` form but not the parameterized `.unsafe()` + `{ prepare: false }` twin this PR fixes. Extending that guard to the parameterized form would prevent the next instance (happy to follow up).

*Authored with AI assistance (Claude) under the contributor's direction.*

garrytan-agents:

Thanks for the report/fix on `op_checkpoints completed_keys double-encode`. Verified fixed on 0.42.53.0.

garrytan:

Thank you @immy2good for the independent fix — several of you found this same bug in parallel. Crediting contributors in the v0.42.53.0 release notes. Closing as superseded.

PR #2310, opened 2026-06-19
fix(capture): make fallback title truncation explicit

xd-Neji:
## Summary

`gbrain capture` already derives fallback titles from the first non-empty body line and caps them at 80 characters. This keeps that existing cap, but makes truncation safer and visible.

Changes:

- replace silent `.slice(0, 80)` truncation with an explicit ellipsis when fallback titles are overlong
- keep the existing first-line title semantics and 80-character cap
- truncate with bounded codepoint-aware iteration so astral Unicode is not split
- avoid materializing an entire overlong first line just to derive the capped title
- preserve useful URL prefixes instead of collapsing long URLs at separators
- update capture skill docs and regression coverage

## Testing

- `env -u GBRAIN_HOME HOME="$(mktemp -d)" bun test test/capture-build-content.test.ts test/commands/capture.test.ts`
  - 45 pass, 0 fail
- `bun run typecheck`
- `bun run verify`
  - 30 checks green
- `git diff --check`

Also manually checked derived-title behavior for long single-token text, long URLs, astral Unicode, separator-only first lines, and dotted title text.


time-attack:

PR #2313, opened 2026-06-20
fix(entities): resolver fallback preserves directory prefix instead of collapsing the slash

ElliotDrel:
## Problem
The step-4 deterministic-slugify fallback in `resolveEntitySlug` / `resolveEntitySlugWithSource` (`src/core/entities/resolve.ts`) runs plain `slugify()`, which collapses the slash in an already-prefixed reference: `people/Some Name` → `people-some-name`. That mangled, unprefixed slug can never match the real `people/some-name` page (and is rejected by the fence-write stub-guard), so the resolved fact silently drops to DB-only instead of fencing onto the entity page.

## Fix
Add `slugifyPreservingPrefix()` — split on the first slash, slugify each side, rejoin — and use it in the two fallback call sites. The exported flat `slugify()` contract is unchanged; only the prefixed-reference fallback path is affected.

## Repro
1. Brain with a real page `people/some-person.md` and no fuzzy-matching title.
2. `resolveEntitySlug(engine, source_id, "people/Some Person")` (already-prefixed free-form ref that misses exact/fuzzy/expansion).
3. Before: returns `people-some-person` (slash collapsed) → fact drops to DB-only. After: returns `people/some-person` → fences onto the page.

## Tests
Adds `slugifyPreservingPrefix` cases to `test/entity-resolve.test.ts`.

### DIFF
--- a/src/core/entities/resolve.ts
+++ b/src/core/entities/resolve.ts
@@ -71,8 +71,38 @@ export async function resolveEntitySlug(
     if (expanded) return expanded;
   }
 
-  // 4. Fallback: deterministic slugify.
-  return slugify(trimmed);
+  // 4. Fallback: deterministic slugify, PRESERVING any directory prefix.
+  //    Plain slugify("people/Some Name") → "people-some-name" mangles the
+  //    slash, which (a) can never match the real people/some-name page and
+  //    (b) makes the fence-write stub-guard refuse it as unprefixed → the fact
+  //    silently drops to DB-only instead of being fenced onto a person page.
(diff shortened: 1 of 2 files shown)


time-attack:
Thanks — verified this one against current master and it is a real, still-present correctness bug: the step-4 fallback in resolveEntitySlug/resolveEntitySlugWithSource collapses the slash in an already-prefixed reference, producing a flat slug that can never match the real prefixed page. The fix is well scoped: a local helper used only in the two fallback call sites, the exported slugify() contract unchanged, and the new cases cover the no-slash and extra-segment paths. Merging.

PR #2314, opened 2026-06-20
fix(doctor): serialize onboard checks + add timeout to avoid connection-pool deadlock on remote engines

ElliotDrel:
## Problem
`runAllOnboardChecks` (`src/core/onboard/checks.ts`) runs its 7 DB checks concurrently via `Promise.all`. Each check acquires at least one connection, so on a pooled remote engine (Postgres/Supabase) with a small pool (e.g. `GBRAIN_POOL_SIZE=2`), the concurrent acquisitions exhaust the pool and deadlock — `gbrain doctor` hangs indefinitely (observed >20 min vs ~10s after the fix). It's invisible on PGLite, which has no connection pool.

## Fix
- Run the onboard checks **sequentially** so in-flight connections never exceed the pool size. The checks are cheap serial counts and each already try/catches internally, so this costs ~milliseconds.
- Defensively bound the call from `doctor.ts` with a **30s timeout** that falls through to a `warn` check rather than hanging the whole `doctor` run.

## Repro
1. Point gbrain at a remote Postgres/Supabase brain with a small pool (`GBRAIN_POOL_SIZE=2`).
2. Run `gbrain doctor`. Before: the onboard-checks phase hangs (pool exhausted). After: completes in seconds.

## Notes
Open PR #2007 also touches these same 7 onboard checks (for a different `categorizeCheck` reason) — maintainer may want to land both together.

### DIFF
--- a/src/commands/doctor.ts
+++ b/src/commands/doctor.ts
@@ -7219,7 +7219,30 @@ export async function buildChecks(
     // for embed_staleness, TABLESAMPLE on PG >50K for the coverage pair).
     progress.heartbeat('onboard_checks');
     const { runAllOnboardChecks } = await import('../core/onboard/checks.ts');
-    const onboardResults = await runAllOnboardChecks(engine);
+    // Defensive cap: runAllOnboardChecks talks to the DB and could in principle
+    // stall on a wedged pooled connection. Never let it hang the entire doctor
+    // run — bound it and fall through with a warn if it blows the budget.
+    const ONBOARD_TIMEOUT_MS = 30_000;
+    let onboardResults: { check: Check }[] = [];
+    try {
+      onboardResults = await Promise.race([
+        runAllOnboardChecks(engine),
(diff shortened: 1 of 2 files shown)


time-attack:
Thank you for the detailed writeup, but the diagnosis does not hold up on inspection: every onboard check routes through safeCount -> engine.executeRaw, which uses the shared pool with no reserved connection and no nested acquisition, so postgres.js queues the excess concurrent queries rather than deadlocking — seven cheap COUNT queries cannot deadlock a pool of two, and this repo's genuine pool-exhaustion failure mode is an EMAXCONNSESSION error rather than an indefinite hang. That leaves the 30s doctor timeout as a band-aid introducing a new silent-warn path for an unestablished cause, so I am closing this. If you can capture a stack or pool trace from the hang showing where the connection is actually held, that evidence would be the thing to open a new issue with.

PR #2316, opened 2026-06-20
fix(files): persist small raw uploads into a per-page .raw/ git sidecar (closes #2297)

### LINKED ISSUE #2297: files upload-raw silently no-ops for small text/PDF (returns success, persists nothing)
## Summary

`gbrain files upload-raw <file> --page <slug>` reports success for small (<100MB) non-media files but never persists them: no copy into the brain repo, and no row in the `files` table. Provenance for every small text/PDF "raw" is silently lost while the command returns `success: true`.

## Version

gbrain 0.42.51.0

## Repro

```
printf 'hello\n' > /tmp/r.txt
gbrain files upload-raw /tmp/r.txt --page people/example
# prints: {"success":true,"storage":"git","path":"/tmp/r.txt","size":6,...}
# but: nothing is written anywhere under the brain repo, and
#      SELECT * FROM files WHERE page_slug='people/example' returns no row
```

## Cause

`src/commands/files.ts`, the `!needsCloud` (small-file) branch. It returns a success JSON and returns early, without ever copying the file into the repo or inserting a DB record. The reported `path` is the input path, not a destination inside the brain repo. The doc comment says small files "stays in git (brain repo)", but the code never writes there:

```ts
if (!needsCloud) {
  // Small text/PDF files stay in git
  console.log(JSON.stringify({
    success: true,
    storage: 'git',
    path: filePath,   // input path, not an in-repo destination
    size: stat.size,
    ...
  }));
  return;             // no copy into the repo, no DB insert
}
```

## Expected

For the git branch, actually copy the file into a deterministic in-repo location (for example a `.raw/` sidecar next to the page, `<category>/.raw/<page-name>/<filename>`, which matches the convention the docs describe) and/or insert a `files` row, so the raw source is preserved and gets committed.

## Note on the usual workaround

`gbrain files upload` (without `-raw`) is sometimes suggested as the workaround, but under a `local` storage backend it also produced no file and no output in our setup, so there is currently no working CLI path for small-text raw preservation. We route around it with a small helper that writes the `.raw/` sidecar directly and verifies the copy.

## Impact

Any skill or automation that relies on `upload-raw` for provenance (the bundled ingest skills do) silently loses raw sources while reporting success, so the failure is invisible until someone goes looking for a `.raw` file that was never written.

ElliotDrel:
Closes #2297.

## Problem
`gbrain files upload-raw` silently no-ops on small (<100MB) non-media files: it returns `success: true` while writing nothing to disk or the `files` table, losing the raw-source provenance the command is supposed to capture.

## Fix
For small raw uploads, resolve the default brain repo path, derive the owning page, create a `<page>.raw/` sidecar directory beside the page, copy the file in, and register the stored path in the `files` table (upsert on `storage_path`). This makes the documented "small files stay in git" path actually work, keeping raw source versioned alongside the page. Also disambiguates `localHash` vs `cloudHash`, and restores `uploadFile` to its prior behavior (a stray rename had left dangling references that would throw on `gbrain files upload`).

## Repro
`gbrain files upload-raw <small .txt/.pdf> --page <slug>`. Before: returns success, nothing on disk, no `files` row. After: file lands in `<page>.raw/` and a `files` row is created.

## Note for maintainer
#2297 suggested a `<category>/.raw/<page-name>/` path; this implements `<page>.raw/` beside the page. Same sidecar concept — happy to align the path shape to your preference.

### DIFF
--- a/src/commands/files.ts
+++ b/src/commands/files.ts
@@ -1,4 +1,4 @@
-import { readFileSync, readdirSync, statSync, lstatSync, existsSync, writeFileSync, unlinkSync, mkdirSync } from 'fs';
+import { readFileSync, readdirSync, statSync, lstatSync, existsSync, writeFileSync, unlinkSync, mkdirSync, copyFileSync } from 'fs';
 import { join, relative, extname, basename, dirname } from 'path';
 import { createHash } from 'crypto';
 import type { BrainEngine } from '../core/engine.ts';
@@ -6,6 +6,7 @@ import { sqlQueryForEngine, executeRawJsonb } from '../core/sql-query.ts';
 import { humanSize } from '../core/file-resolver.ts';
 import { createProgress } from '../core/progress.ts';
 import { getCliOptions, cliOptsToProgressOptions } from '../core/cli-options.ts';
+import { getDefaultSourcePath } from '../core/source-resolver.ts';
 
 /** Size threshold: files >= 100 MB use TUS resumable upload */
 const SIZE_THRESHOLD = 100 * 1024 * 1024;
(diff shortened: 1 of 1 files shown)


time-attack:
Closing as a duplicate of #2398, which covers this. Please follow along there — and thank you for the report.

Issue #1893, opened 2026-06-05
doctor content_sanity_audit_recent FAIL recurs weekly because audit JSONLs have no native rotation / retention policy


Mr-B-1:
## Problem

`doctor` check `content_sanity_audit_recent` fails as soon as the rolling window's `/root/.gbrain/audit/content-sanity-*.jsonl` files accumulate more than the doctor's threshold of events. For a corpus with chronic oversized pages (transcripts, long docs), the same ~2,000 pages re-emit `PAGE_OVERSIZE_WARN` on every weekly audit run, so the event count climbs back above threshold every Monday morning even after the operator has acknowledged the underlying pages are intentionally long.

## Current behavior

- v0.42.8.0 shipped `feat: content-quality gate on sync — quarantine junk + flag boilerplate` which prevents NEW junk pages from being ingested. ✓
- But the audit LOG of past events accumulates with no native rotation. Doctor reads everything in `/root/.gbrain/audit/content-sanity-*.jsonl`.
- The doctor's own message ("audit JSONL is local to this host; share GBRAIN_AUDIT_DIR for multi-host visibility") implies operators should manage these files manually.
- There is no native command to:
  - Set a rolling retention window (e.g. "only events from last N days count")
  - Rotate old audit files (`gbrain audit rotate --older-than 7d`)
  - Re-baseline / dismiss historical chronic-page events without deleting the file

## Reproduction

1. Brain with ≥ 20 chronic oversized pages (any pages >50KB body)
2. Wait for the weekly content-sanity audit run
3. `gbrain doctor` → FAIL `content_sanity_audit_recent: N events (hard=0 soft=X warn=Y)` where Y is dominated by `PAGE_OVERSIZE_WARN`
4. Operator archives `/root/.gbrain/audit/content-sanity-*.jsonl` manually → FAIL clears
5. Next weekly audit run regenerates same events → FAIL recurs

## Requested

One of:

1. Native `gbrain audit rotate [--older-than 7d]` command + config key for default retention.
2. Doctor reads a rolling window (`content_sanity_audit.window_days`, default 7) instead of all JSONLs in the dir.
3. Doctor distinguishes "chronic recurring" event signatures from "new" ones — same page+reason pair flagged 4+ weeks in a row → dedupe in score.
4. A `gbrain audit ack --pattern PAGE_OVERSIZE_WARN --slug-prefix <X>` to acknowledge chronic warnings (similar to how lint suppression works).

## Workaround we shipped

[...]

PR #2317, opened 2026-06-20
test(import-file): isolate GBRAIN_AUDIT_DIR so oversize fixtures don't pollute the operator's real audit log

ElliotDrel:
## Problem
The "accepts in-memory content just under MAX_FILE_SIZE" test in `test/import-file.test.ts` imports a 4.9MB page that trips the content-sanity oversize soft-block path. `resolveAuditDir()` honors `GBRAIN_AUDIT_DIR` and defaults to `~/.gbrain/audit`, so the test writes that soft_block event into the operator's REAL audit log on every run — a phantom event that surfaces as a recurring `content_sanity_audit_recent` WARN in `gbrain doctor` (related: #1893).

## Fix
Point `GBRAIN_AUDIT_DIR` at the test's temp dir in `beforeAll` and restore the prior value in `afterAll`. Self-contained; no behavior change to the code under test.

## Repro
`bun test test/import-file.test.ts`, then `gbrain doctor` → a `content_sanity` event appears in `~/.gbrain/audit/content-sanity-<week>.jsonl` and feeds the recurring WARN. After the fix, the run writes only to the temp dir.

### DIFF
--- a/test/import-file.test.ts
+++ b/test/import-file.test.ts
@@ -29,12 +29,24 @@ function mockEngine(overrides: Partial<Record<string, any>> = {}): BrainEngine {
   return engine;
 }
 
+let priorAuditDir: string | undefined;
+
 beforeAll(() => {
   mkdirSync(TMP, { recursive: true });
+  // Isolate audit writes to the temp dir. The oversized-content fixtures below
+  // (e.g. the 4.9MB `borderline-slug` import) trip the content-sanity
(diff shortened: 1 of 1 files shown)


time-attack:
Good catch, and verified: the test currently writes a real soft_block event into the operator's ~/.gbrain audit log because resolveAuditDir() falls back to the home directory, which is where that recurring doctor WARN comes from. Twelve lines of env isolation with proper restore in afterAll, no change to the code under test. Merging.

PR #2318, opened 2026-06-20
feat(tags): native gbrain tags command for tag-vocabulary hygiene (list/audit/merge)

ElliotDrel:
## What & why
Adds a first-class `gbrain tags` subcommand for whole-brain tag-vocabulary hygiene (sibling to `check-backlinks`). Tag drift — `ai`/`AI`/`a-i`, `startup`/`startups` — silently fragments a brain's graph into synonym islands; today there's no deterministic way to spot or consolidate it.

- `tags list` — prints the tag vocabulary with per-tag page counts.
- `tags audit` — clusters mechanical near-duplicates (case / separator / plural / 1-char typo via Levenshtein), read-only.
- `tags merge <from> <to> --apply` — repoints every page's frontmatter from `<from>` to `<to>` and reconciles the engine in-process (`removeTag` + `addTag`, since sync only adds). Dry-run by default.

## Design notes
- Filesystem is the source of truth (reads git-tracked `.md` frontmatter); engine is reconciled after.
- The `src/cli.ts` wiring is a small guarded intercept for `tags list|audit|merge` that preserves the existing bare `tags <slug>` → `get_tags` behavior (no regression).
- Self-contained: `src/commands/tags.ts` imports only `fs`/`path`/`child_process` (`git ls-files`) and existing core (`config`, `engine-factory`, `engine`).

## Tests / verification
- `bun build src/commands/tags.ts src/cli.ts --target=bun --outdir=...` → succeeds (BUILD_OK). (`--outdir` is required by Bun when passing two entry points.)
- Exercised offline against a tiny temp brain (three `.md` pages mixing block-style and inline `tags:`):
  - `tags list` and `tags list --json` — correct counts (`ai`×2, `AI`, `startup`, `startups`, `writing`).
  - `tags audit` — correctly clustered the case dupe (`AI`→`ai`) and the plural dupe (`startups`→`startup`), and emitted ready-to-run `merge` commands.
  - `tags merge AI ai` (dry-run) — reported the 1 affected page without writing.
- Regression check: a bare `gbrain tags <slug>` was confirmed to fall through to the existing `get_tags` op (the intercept only matches `list|audit|merge`), not the new command.
- Not yet exercised in this PR: a live `merge --apply` round-trip against a connected engine (the offline temp brain has no DB). The frontmatter-rewrite path is covered by the dry-run; the engine-reconcile path (`removeTag`/`addTag`) is straightforward but unverified here.

Happy to adjust scope/UX to fit project conventions.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -235,6 +235,17 @@ async function main() {
     command = 'query';
   }
 
+  // `gbrain tags list|audit|merge` is whole-brain tag-vocabulary hygiene,
+  // distinct from the `tags <slug>` op (get_tags — list tags for ONE page).
+  // Only these three subcommands route here; a bare `gbrain tags <slug>`
+  // falls through to get_tags below. (A page literally slugged list/audit/
+  // merge would be shadowed — acceptable; none exist in practice.)
+  if (command === 'tags' && ['list', 'audit', 'merge'].includes(subArgs[0] ?? '')) {
+    const { runTags } = await import('./commands/tags.ts');
+    await runTags(subArgs);
+    return;
+  }
+
   // T5 — `gbrain search modes|stats|tune` is the read-only config dashboard,
   // NOT a free-text search for the literal word "modes". Free-text
   // `gbrain search "<query>"` falls through to the cheap-hybrid `search` op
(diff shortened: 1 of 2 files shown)


time-attack:
Thanks for the careful work here, but this is new command surface rather than a fix, and feature additions are closed by default to keep the CLI from accreting. It also sits outside the contract-first path — CLI and MCP are both generated from src/core/operations.ts, and this adds a CLI-only command via a hand-rolled dispatch special case that shadows `tags <slug>` for pages slugged list/audit/merge. Closing; no reflection on the code quality.

Issue #2295, opened 2026-06-19
Bug: diagnoseEmbedding() blocks ollama embedding provider


zunami:
### Problem
When using ollama as embedding provider in gbrain v0.42.51.0, `diagnoseEmbedding()` unconditionally returns `user_provided_model_unset` because the ollama recipe has an empty `models: []` array. The config correctly parses `modelId = "nomic-embed-text"` from `embedding_model: ollama:nomic-embed-text`, but the preflight guard ignores the parsed modelId.

### Reproduction
```bash
export OLLAMA_BASE_URL=http://192.168.1.245:11434
gbrain init --pglite --embedding-model ollama:nomic-embed-text
gbrain providers test --model ollama:nomic-embed-text
# → ✗ transient error: Cannot connect to API
```
While direct curl succeeds:
```bash
curl http://192.168.1.245:11434/v1/embeddings -d '{"model":"nomic-embed-text","input":"test"}'
# → 200 OK with valid embedding
```

### Root Cause
`diagnoseEmbedding()` in `src/core/ai/gateway.ts` rejects on empty `models: []` when `isUserProvided: true`, ignoring the parsed `modelId`.

### Proposed Fix
```typescript
// Add !parsed.modelId to the condition
if (!parsed.modelId && ...)  // guard no longer rejects when modelId is parsed
```

### Workaround
`gbrain init --pglite --no-embedding` + `gbrain import <dir> --no-embed`

### Environment
- gbrain 0.42.51.0
- Ollama with nomic-embed-text on remote host (192.168.1.245)
- PGLite engine

hermes agent 0.16

stevegrubb:
I'm seeing the exact same issue but hermes agent 0.17 and gbrain  0.42.53.0.

FaisalMSiddiq:
I applied a different workaround changed the OPENAI URL to local hosted OLLAMA URL, initialized the GBrain instance with openai:{models-served-by-ollama} it worked for me

time-attack:
[...]

PR #2321, opened 2026-06-20
fix(ai): honor user-provided embedding model ids

### LINKED ISSUE #2204: diagnoseEmbedding() falsely returns user_provided_model_unset when modelId is parsed from config string

moneycaringcoder:
## Summary
- allow `diagnoseEmbedding()` to accept user-provided embedding model ids parsed from the `provider:model` config string
- keep the existing `user_provided_model_unset` diagnostic for empty-model recipes only when no model id was provided
- add a regression test for a `llama-server:<model>` embedding config

## Test Plan
- `npm exec --yes bun@1.3.13 -- test test/ai/gateway.test.ts`
- `npm exec --yes bun@1.3.13 -- run typecheck`

Fixes #2204.
Related to #2295.

### DIFF
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -709,7 +709,8 @@ export function diagnoseEmbedding(modelOverride?: string): EmbeddingDiagnosis {
   if (
     Array.isArray(tp.models) &&
     tp.models.length === 0 &&
-    (recipe.id === 'litellm' || isUserProvided)
+    (recipe.id === 'litellm' || isUserProvided) &&
+    !parsed.modelId
   ) {
     return {
       ok: false,
--- a/test/ai/gateway.test.ts
+++ b/test/ai/gateway.test.ts
@@ -8,6 +8,7 @@ import {
   getEmbeddingDimensions,
   getExpansionModel,
   VoyageResponseTooLargeError,
+  diagnoseEmbedding,
 } from '../../src/core/ai/gateway.ts';
 
 // v0.39.x ship-wave fix: gateway module is process-scoped. Without an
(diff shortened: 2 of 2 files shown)


Willardgmoore:
Watching this — blocked on the same issue with `llama-server:Qwen3-Embedding-0.6B`. Hoping this lands soon.

time-attack:
Thanks for this, @moneycaringcoder! 🙏

What this does: This PR fixes a bug where user-provided embedding models (like llama-server) with fully-specified model IDs fail validation even when embedding dimensions are explicitly set. The fix adds a modelId check to prevent false positives.

PR #2324, opened 2026-06-20
fix(db,search,eval): jsonb double-encode — contradiction-probe alias-hop crash, repo-wide audit + CI guard

DarkNightForge:
## What & why

Fixes a reproducible `gbrain eval suspected-contradictions` crash and the broader jsonb double-encode bug it surfaced. Two logical commits:

### 1. Contradiction-probe crash (`fix(search,eval)`)
A whole-brain probe aborted with `UNDEFINED_VALUE: Undefined values are not allowed`. Root cause traced via `bun src/cli.ts`:

```
postgres arraySerializer (undefined in ::int[] bind)
  → listActiveTakesForPages (postgres-engine.ts, WHERE t.page_id = ANY($1::int[]))
  → generateIntraPagePairs (eval-contradictions/runner.ts)
```

- **`src/core/search/hybrid.ts` `applyAliasHop`** — when a query's text is an exact page alias and the page isn't already in the organic results, the injected synthetic `SearchResult` omitted `page_id` (the `as SearchResult` cast hid the missing required field). Any consumer reading `page_id` off an alias-injected result got `undefined`. Fix: include `page_id: page.id` (the page is already fetched).
- **`src/core/eval-contradictions/runner.ts` `generateIntraPagePairs`** — mapped `r.page_id` straight into the `ANY($1::int[])` bind, so one undefined aborted the whole probe. Fix: filter to finite numbers before the bind (mirrors the guard in `query-cache.ts`).

### 2. Repo-wide jsonb double-encode + CI guard (`fix(db)`)
The crash investigation exposed that `query_cache.{results,meta,page_generations}` were stored as jsonb **string scalars**, not arrays/objects — the "never `JSON.stringify` into a `::jsonb` cast" footgun from CLAUDE.md, but in **positional-param** form (`executeRaw(\`…$N::jsonb…\`, [JSON.stringify(x)])`) that the existing `check-jsonb-pattern.sh` grep doesn't catch. postgres.js re-encodes the string into a jsonb string scalar (PGLite hides it), breaking `jsonb_each` / `->>` / `jsonb_array_elements`.

An empirical `jsonb_typeof` sweep of all 35 jsonb columns plus a new static scanner found the full set — **23 sites across 16 files**. Confirmed live: `take_proposals.dedup_against_fence_rows` (stored as string). The rest were latent (empty/PGLite-tolerated columns); several are PGLite-engine twins of postgres-engine writes that already used `sql.json()`, so this also restores engine parity.

Fix everywhere: pass the **raw object/array**, never `JSON.stringify` it. Readers were already defensive (`typeof === 'string' ? JSON.parse : value`), so existing string rows still read back, and rows self-heal (overwrite/TTL).

**New guard:** `scripts/check-jsonb-params.mjs` extracts each `executeRaw`/`.unsafe`/`.query` call's balanced arg span (comments stripped, strings/templates skipped) and fails when it binds `JSON.stringify(...)` into a `$N::jsonb` param; the intentional `to_jsonb($N::text)` form is allowed. Wired into `check-jsonb-pattern.sh`.

## Verification
- Previously-crashing `--query "Business Brain"` now completes (and surfaces a temporal finding the crash had masked).
- Full whole-brain probe runs end-to-end (124 pairs, 0 contradictions, 0 judge errors).
- Live round-trips on Postgres confirm fresh writes now store `array`/`object` (were `string`).
- New guard catches a planted reintroduction; clean on the fixed tree.
- `tsc --noEmit` clean.

## Notes for maintainers
- `VERSION` / `package.json` / `CHANGELOG.md` intentionally untouched — left to your `/ship` release flow.
- Existing double-encoded rows in live brains self-heal; a one-time `UPDATE … SET c = (c #>> '{}')::jsonb WHERE jsonb_typeof(c)='string'` can repair them eagerly if desired.


garrytan:
Delivered by #2375 (v0.42.53.0): the repo-wide positional jsonb double-encode sweep, the AST guard (`scripts/check-jsonb-params.mjs`) wired into `check-jsonb-pattern.sh` and CI, and a DATABASE_URL-gated Postgres parity job. Thanks @DarkNightForge for scoping the umbrella. Closing as delivered.

PR #2333, opened 2026-06-21
fix(sync): bind op_checkpoints.completed_keys as text[] to avoid jsonb double-encode on native Postgres

### LINKED ISSUE #2332: recordCompleted double-encodes completed_keys on native Postgres → every incremental sync aborts PARTIAL (checkpoint_unavailable)
### Summary

On the native Postgres engine (not PGLite), `recordCompleted()` writes `op_checkpoints.completed_keys` as a JSON-stringified array bound to a `$3::jsonb` param. postgres.js JSON-encodes the JS string **a second time**, so the value lands as a jsonb *string* (`"[\"abc\"]"`) instead of a jsonb *array*. That violates the table's own `CHECK (jsonb_typeof(completed_keys) = 'array')`, the checkpoint INSERT is rejected, and the sync aborts:

```
[op-checkpoint] write failed (sync-target, <fp>): new row for relation
  "op_checkpoints" violates check constraint "op_checkpoints_completed_keys_array"
[sync] checkpoint target write failed (pool unavailable) — aborting before import
Sync PARTIAL at <commit>: imported 0 of 1 file(s), reason=checkpoint_unavailable.
  Re-run 'gbrain sync' to continue (last_commit unchanged; safe to retry).
```

Because the mint of the run's checkpoint is what fails, **no incremental sync can ever advance `sources.last_commit`** — the index silently goes stale. The "safe to retry" hint is misleading: the failure is deterministic, so re-running never clears it. `gbrain sync --full` is the only thing that works (it doesn't mint this checkpoint).

### Environment

- gbrain `0.42.51.0` (HEAD `9bf96db8`)
- engine: native Postgres 17 (Homebrew), direct `database_url` (session pooler)
- postgres.js `3.4.9`
- macOS `27.0` — relevant only because the macOS 26/27 PGLite WASM bug (#223/#1670) is what pushes users onto native Postgres, where this surfaces.

### Root cause

`src/core/op-checkpoint.ts`, `recordCompleted()`:

```ts
engine.executeRawDirect(
   VALUES ($1, $2, $3::jsonb, now())
   ON CONFLICT (op, fingerprint) DO UPDATE
     SET completed_keys = EXCLUDED.completed_keys, updated_at = now()`,
  [key.op, key.fingerprint, JSON.stringify(sorted)],   // <-- double-encoded
);
```

The inline comment asserts this form is safe ("JSON.stringify into `$3::jsonb` is correct ... NOT the double-encode trap"). That holds on PGLite but **not** on the postgres.js `.unsafe(sql, params)` direct-pool path used by `executeRawDirect` on the native engine.

### Minimal reproduction

```js
// node/bun, from the gbrain checkout
import postgres from 'postgres';
const sql = postgres(process.env.GBRAIN_DATABASE_URL, { max: 1 });
await sql.unsafe(`CREATE TEMP TABLE _p (label text, k jsonb)`);
await sql.unsafe(`INSERT INTO _p VALUES ('current', $1::jsonb)`, [JSON.stringify(['x','y'])]);
await sql.unsafe(`INSERT INTO _p VALUES ('fixed',   to_jsonb($1::text[]))`, [['x','y']]);
console.log(await sql.unsafe(`SELECT label, jsonb_typeof(k) typeof, k::text FROM _p ORDER BY label`));
await sql.end();
```

Output:

```
current  typeof=string   "[\"x\",\"y\"]"     <-- fails CHECK jsonb_typeof='array'
fixed    typeof=array     ["x", "y"]
```

### Why PGLite didn't hit it

PGLite's param path doesn't re-serialize the already-stringified JSON, so the text→jsonb cast yields an array there. Only the native postgres.js direct pool double-encodes.

### Impact

- Every git-backed source on the native engine: `code`, curated `default`, any allowlist repo. Autopilot's incremental cycles and any cron wrapping `gbrain sync` are all affected.
- Failure is silent unless the operator parses sync output — `gbrain sync` exits `0` on the "safe to retry" partial, so wrappers that trust the exit code report green while the index is stale.

### Proposed fix

Bind the keys as a Postgres `text[]` and build the jsonb with `to_jsonb()` — exactly what `APPEND_PATHS_SQL` (the resumable-sync append path) already does and documents as the way to "sidestep executeRawJsonb's array rejection." One-line change; PR incoming.

### Optional follow-ups (separate)

- One-time data repair for any rows already stuck as non-array.
- Have `gbrain sync` exit non-zero (or a distinct code) on `checkpoint_unavailable` so cron/CI wrappers don't read a partial as success.

pedrospereira:
Fixes #2332.

`recordCompleted()` bound `completed_keys` as `JSON.stringify(sorted)` → `$3::jsonb`. On the native postgres.js direct pool this double-encodes the JS string into a jsonb *string*, violating `CHECK (jsonb_typeof = 'array')`, so the checkpoint INSERT fails and every incremental sync aborts PARTIAL (`checkpoint_unavailable`) without advancing `last_commit`. PGLite tolerated the old form; native Postgres does not.

This aligns `recordCompleted` with the already-proven `text[]` binding used by `APPEND_PATHS_SQL`: bind the keys as a Postgres `text[]` and build the jsonb with `to_jsonb()`.

### Verification

- **Driver probe** (postgres.js 3.4.9): old form → `jsonb_typeof = 'string'`; new form → `'array'`.
  ```
  current  typeof=string   "[\"x\",\"y\"]"
  fixed    typeof=array     ["x", "y"]
  ```
- **End-to-end** on native Postgres 17: a previously-failing incremental `gbrain sync` now advances `last_commit`; `op_checkpoints` holds valid array rows and clears on clean exit.

### Optional follow-ups (separate, not in this PR)

- One-time data repair for any rows already stuck as a non-array scalar.
- Have `gbrain sync` exit non-zero (or a distinct code) on `checkpoint_unavailable` so cron/CI wrappers don't read a partial as success.

### DIFF
--- a/src/core/op-checkpoint.ts
+++ b/src/core/op-checkpoint.ts
@@ -180,18 +180,22 @@ export async function recordCompleted(
   // extract-conversation-facts serialize a MUTABLE map through here and rely on
   // stale keys being REMOVED; an append would make them unremovable. The full
   // set lands in the parent `completed_keys` JSONB column via a single UPSERT —
-  // exactly as before. JSON.stringify into `$3::jsonb` is correct (the text→jsonb
-  // cast yields a proper array; NOT the double-encode trap, which is the template
-  // form). Sync uses `appendCompleted` (below) instead, never this.
+  // exactly as before. completed_keys is built with `to_jsonb($3::text[])`: the
+  // keys bind as a Postgres text[] and to_jsonb yields a real jsonb ARRAY. The
+  // older `$3::jsonb` form with JSON.stringify(keys) double-encodes on the native
+  // postgres.js direct pool — the JS string gets JSON-serialized AGAIN, landing a
+  // jsonb *string* that fails CHECK (jsonb_typeof = 'array') and aborts every
+  // incremental sync as PARTIAL (PGLite tolerated it; native Postgres does not).
+  // Sync uses `appendCompleted` (below) instead, never this.
   const sorted = [...keys].sort();
   return durableWrite(engine, key, 'write', () =>
     engine.executeRawDirect(
       `INSERT INTO op_checkpoints (op, fingerprint, completed_keys, updated_at)
-       VALUES ($1, $2, $3::jsonb, now())
+       VALUES ($1, $2, to_jsonb($3::text[]), now())
        ON CONFLICT (op, fingerprint) DO UPDATE
          SET completed_keys = EXCLUDED.completed_keys,
              updated_at     = now()`,
-      [key.op, key.fingerprint, JSON.stringify(sorted)],
+      [key.op, key.fingerprint, sorted],
     ));
 }
 


garrytan-agents:

Thanks for the report/fix on `op_checkpoints.completed_keys text[] binding`. Verified fixed on 0.42.53.0.

garrytan:

Thank you @pedrospereira for the independent fix — several of you found this same bug in parallel. Crediting contributors in the v0.42.53.0 release notes. Closing as superseded.

PR #2336, opened 2026-06-21
fix(gateway): recover unbalanced tool-call histories instead of hard-failing the job

fbal23:
## Problem

The provider-agnostic tool loop (`gateway.chat` → `generateText`, used for **all non-Anthropic models** — including local vLLM models in `dream` synthesis) crashes with AI SDK v6 `AI_MissingToolResultsError`:

> `[chat(openrouter:…)] Tool results are missing for tool calls: chatcmpl-tool-…`

`generateText` validates that every assistant `tool-call` is resolved by a `tool-result` in the **immediately-following** message. When that invariant is violated the whole subagent job dies — **and every retry too**, because the persisted history is replayed verbatim and re-throws the identical error (the classic `retrying… permanently failed` cascade). On a local-model dream setup this killed the large majority of synthesis subagents (~89% dead over 10 days in my brain).

Three real triggers produce an unbalanced history:

1. **Crash / force-evict mid-turn** — the assistant turn (N tool-calls) is persisted but its tool-result message never is, so replay loads an assistant turn with no following results.
2. **Truncation** (`finishReason: 'length'`) emitting a partial parallel tool-call batch.
3. **Provider-duplicated / dropped tool-call IDs** — observed with vLLM parallel tool-use on local models (e.g. Qwen-NVFP4).

## Fix

Add `repairToolPairing()`, a normalization pass applied right before `toModelMessages()` in `chat()`. For every assistant tool-call with no matching result in the next message, it injects a synthetic `error-text` tool-result so the model recovers gracefully on the next turn instead of the job dying. It is a **no-op on already-balanced histories** (the overwhelmingly common case).

This mirrors the AI SDK's own guidance for `AI_MissingToolResultsError` (every tool-call must be resolved before the next model call) and matches the robustness the native-Anthropic path already had.

## Tests

`test/ai/gateway-toolcall-pairing.test.ts` — drives the **real** AI SDK v6 `generateText` via `MockLanguageModelV3` (no network/keys), same pattern as the existing `gateway-tools-schema.test.ts`:

- no-op on a balanced history
- synthesizes results when the assistant turn has no following tool message (crash/evict)
- back-fills **only the missing IDs** of a partial parallel batch (truncation / dropped IDs)
- a previously-fatal unbalanced history now passes `generateText`

All 4 new tests pass; the 32 existing `test/ai/gateway-*` tests remain green.

## Possible follow-up

Reducing parallel tool-call fan-out for local models (the trigger amplifier) would further cut churn, but the invariant repair is the correctness fix and stands alone.


time-attack:

PR #2338, opened 2026-06-21
fix(sync): pin write double-encodes jsonb scalar — every sync aborts at checkpoint

garrytan-agents:
## Problem

Multi-source brains stop syncing entirely after a recent change to the durable checkpoint path. Every `gbrain sync` aborts at the very first checkpoint write with:

```
[op-checkpoint] write failed (sync-target, <fp>): new row for relation "op_checkpoints" violates check constraint "op_checkpoints_completed_keys_array"
[sync] checkpoint target write failed (pool unavailable) — aborting before import; nothing drained, next run retries.
Sync PARTIAL at <commit>: imported 0 of N file(s), reason=checkpoint_unavailable.
```

The `(pool unavailable)` text is a **misleading wrapper** — the pool is healthy. The real error is SQLSTATE `23514` (CHECK violation) on the `completed_keys` column.

## Root cause

`recordCompleted()` in `src/core/op-checkpoint.ts` persists the sync-target pin as:

```ts
 VALUES ($1, $2, $3::jsonb, now()) ...`,
[key.op, key.fingerprint, JSON.stringify(sorted)]
```

Under `postgres.js` `.unsafe(sql, params)`, binding a JS **string** to a `$3::jsonb` param double-encodes it: the text->jsonb cast wraps the already-JSON string into a jsonb **scalar string** rather than parsing it into a jsonb array. That fails the table CHECK:

```sql
CONSTRAINT op_checkpoints_completed_keys_array CHECK (jsonb_typeof(completed_keys) = 'array')
```

`sync.ts` persists the pin via `recordCompleted(ckpt.target, [pin])` *before* the first path flush (by design, so a crash resumes to the pinned target). Because that write throws on the first attempt, **no source can ever anchor a sync** — `appendCompleted` (the `text[]` path-flush) is never reached. The constraint that was added to protect the load is now rejecting every legitimate write.

## Fix

Cast the param through `text` first so it binds as text and the `text->jsonb` cast parses it into a real array:

```ts
VALUES ($1, $2, $3::text::jsonb, now())
```

Verified bindings (live, against Supabase Postgres via postgres.js `.unsafe`):

| binding | `jsonb_typeof(completed_keys)` |
|---|---|
| `$3::jsonb` + `JSON.stringify([...])` | ❌ throws 23514 (scalar) |
| `$3::text::jsonb` + `JSON.stringify([...])` | ✅ `array` |
| `$3` + `sql.json([...])` | ✅ `array` |
| `to_jsonb($3::text[])` + `[...]` | ✅ `array` |

`appendCompleted`'s `unnest($3::text[])` path was already correct and is unaffected; only the single-shot pin write in `recordCompleted` needed the cast.

## Verification

Reproduced and fixed on a 5-source brain that had been stuck **84h** (GBrain Doctor 0/100, all sources `severely stale`). After the one-line change, the smallest source synced on the first attempt:

```
Synced 2018def9..642558cf:
  +19 added, ~0 modified, -0 deleted
  42 chunks created, 19 pages embedded
```

Scope: one-line, lowest-risk cast on the pin-write statement. No schema change, no behavior change beyond letting the write land.

### DIFF
--- a/src/core/op-checkpoint.ts
+++ b/src/core/op-checkpoint.ts
@@ -187,7 +187,7 @@ export async function recordCompleted(
   return durableWrite(engine, key, 'write', () =>
     engine.executeRawDirect(
       `INSERT INTO op_checkpoints (op, fingerprint, completed_keys, updated_at)
-       VALUES ($1, $2, $3::jsonb, now())
+       VALUES ($1, $2, $3::text::jsonb, now())
        ON CONFLICT (op, fingerprint) DO UPDATE
          SET completed_keys = EXCLUDED.completed_keys,
              updated_at     = now()`,


garrytan-agents:
## Re-confirmed live after v0.42.52.0 — this is NOT a false alarm

v0.42.52.0 (#2287) reintroduced `$3::jsonb` in `recordCompleted` with a code comment asserting it is correct ("JSON.stringify into `$3::jsonb` is correct ... NOT the double-encode trap"). That assertion is empirically false against a real Postgres pool. Running a multi-source sync on that exact build:

```
[op-checkpoint] write failed (sync-target, 4f21bbc1): new row for relation "op_checkpoints" violates check constraint "op_checkpoints_completed_keys_array"
[sync] checkpoint target write failed (pool unavailable) — aborting before import; nothing drained, next run retries.
Sync PARTIAL: imported 0 of 4682 file(s), reason=checkpoint_unavailable.
```

The reason `recordCompleted` differs from the working `appendCompleted`: `appendCompleted` binds a `text[]` and does `unnest($3::text[])`, so the array is constructed server-side. `recordCompleted` binds a single JS **string** (`JSON.stringify(sorted)`) to a `$3::jsonb` param. Under postgres.js `.unsafe()`, a string bound to `::jsonb` takes the `text → jsonb` cast path, which produces a jsonb **scalar string** (`"[...]"`), not a jsonb array — failing `CHECK (jsonb_typeof(completed_keys) = 'array')`.

Forcing the param to bind as text first (`$3::text::jsonb`) makes the cast parse the JSON into a real array. Verified: straylight-brain (4682 files) syncs clean immediately after the cast change; reverts to the constraint failure without it.

Suggest also deleting/correcting the inline comment defending `$3::jsonb`, since it will invite another revert.

DiegoRetamalPacheco:
### Independent repro — confirmed on Supabase + Bun

Hit this on `gbrain 0.42.52.0` against a managed **Supabase** Postgres (Supavisor pooler), runtime **Bun 1.3.14**, **postgres.js 3.4.9**. Every `gbrain sync` of a source *with changes* aborted before import:

```
[op-checkpoint] write failed (sync-target, b4895013): new row for relation "op_checkpoints" violates check constraint "op_checkpoints_completed_keys_array"
[sync] checkpoint target write failed (pool unavailable) — aborting before import; nothing drained, next run retries.
Sync PARTIAL at c1dbf3c4: imported 0 of 1 file(s), reason=checkpoint_unavailable.
```

Sources with **0 changes** synced fine (they never reach the pin write), which is what masks it at first.

Isolated the binding against the live DB to rule out anything app-side:

```js
// payload = JSON.stringify(['x']) === '["x"]'
sql.unsafe(`SELECT jsonb_typeof($1::jsonb) t`, [payload])  // -> { t: 'string' }  ❌ scalar
sql`SELECT jsonb_typeof(${payload}::jsonb) t`              // -> { t: 'string' }  ❌ scalar (template too)
sql.unsafe(`SELECT jsonb_typeof('["x"]'::jsonb) t`)        // -> { t: 'array'  }  ✅ literal only
```

So under this runtime *both* the positional `.unsafe()` form and the template form double-encode — the inline comment in `recordCompleted` asserting the positional form is safe is empirically false here, confirming your note about deleting it before it invites another revert.

Verified both proposed fixes resolve it (rollback-tx test, then full sync):
- this PR's `$3::text::jsonb`
- the `to_jsonb($3::text[])` variant (matches the `appendCompleted` text[] path)

Both produce `jsonb_typeof = 'array'` and the sync drains cleanly. 👍 for getting one of these merged — the duplicate-PR pile suggests it's biting a lot of people.

garrytan-agents:

Thanks for the report/fix on `pin write double-encodes jsonb scalar`. Verified fixed on 0.42.53.0.

garrytan:

Thank you @garrytan-agents for the independent fix — several of you found this same bug in parallel. Crediting contributors in the v0.42.53.0 release notes. Closing as superseded.

PR #2342, opened 2026-06-21
fix(jobs): retry resets started_at + attempts counters

ethanbeard:
## Problem

`gbrain jobs retry <id>` left `started_at` populated from the original run. The DB-side wall-clock detector (`handleWallClockTimeouts`) compares `now() - started_at` against `timeout_ms * 2` and dead-letters any job past that window. For a retry of a *dead* job, the elapsed time since the original start is essentially always `> timeout_ms * 2` — an operator only retries a job after noticing it died, which takes longer than the timeout window itself.

Result: the retry was immediately re-dead-lettered by the next worker tick before any handler ran.

Symptom:
```
gbrain jobs retry 123
# status flips waiting -> dead within ~60s
# error_text = "wall-clock timeout exceeded", attempts_made unchanged
```

## Fix

`retryJob` now resets three fields:

- `started_at = NULL` — the core fix. `claim()` does `started_at = COALESCE(started_at, now())`, so a fresh start is recorded only when the retry is actually claimed.
- `attempts_made = 0` — gives the retry a fresh `max_attempts` round (otherwise a 3/3 dead job dead-letters on the first failure of the retry).
- `attempts_started = 0` — consistency with `attempts_made`.

`stacktrace` is preserved to retain debug history across retries.

This is adjacent to the attempt-counter correctness work in the recent reliability releases (e.g. counting timed-out runs as spent attempts) — it makes operator-initiated retry behave consistently with that model.

## Test

Extends the existing `retry dead job re-queues` test to assert the three resets. Original assertions still pass.

### DIFF
--- a/src/core/minions/queue.ts
+++ b/src/core/minions/queue.ts
@@ -450,12 +450,25 @@ export class MinionQueue {
     });
   }
 
-  /** Re-queue a failed or dead job for retry. */
+  /**
+   * Re-queue a failed or dead job for retry.
+   *
+   * Resets `started_at` and the attempts counters so the user-initiated retry
+   * gets a fresh wall-clock budget and a fresh max_attempts round. Without
+   * resetting `started_at`, the DB-side wall-clock detector
+   * (handleWallClockTimeouts) compares `now() - started_at` against
+   * `timeout_ms * 2` and immediately dead-letters any retry where more time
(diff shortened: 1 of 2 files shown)


ethanbeard:
Closing as a duplicate of #1376, which already covers this exact fix. Apologies for the noise.

PR #2343, opened 2026-06-21
fix(recipes/x-to-brain): use /users/by/username for app-only bearer health check

ethanbeard:
## Problem

The `x-to-brain` recipe's health check probes `https://api.x.com/2/users/me`, which requires **user-context OAuth**. Under an app-only bearer token (what the rest of the recipe uses), `/users/me` returns 401 — so the health check fails on every probe even when the token is valid.

## Fix

Switch the probe to `https://api.x.com/2/users/by/username/$X_HANDLE`, which works under app-only bearer and matches how the rest of the recipe authenticates.

Requires the operator to set `$X_HANDLE` alongside `$X_BEARER_TOKEN`.


time-attack:

PR #2353, opened 2026-06-22
docs(schema): list all 7 bundled read-only packs, not just 2

JiraiyaETH:
The `schema-evolution` and `schema-author` skill docs only list `gbrain-base` and `gbrain-recommended` as the read-only bundled packs to fork before mutating. gbrain now ships **7** bundled read-only packs, so the "fork first" guidance + matching anti-patterns are stale — an agent reading them could try to mutate one of the other 5 (`gbrain-base-v2`, `gbrain-creator`, `gbrain-investor`, `gbrain-engineer`, `gbrain-everything`) and hit a confusing `PACK_READONLY` refusal the docs never warned about.

Updates both docs to list all 7. Docs-only; no behavior change.

### DIFF
--- a/skills/conventions/schema-evolution.md
+++ b/skills/conventions/schema-evolution.md
@@ -89,7 +89,7 @@ imports under `meetings/` infer `meeting` type via the pack.
   `expert_routing_without_prefix` lint rule warns about this exact
   shape: an expert-routed type with no prefix never matches a put_page
   inference, so `whoknows` silently never surfaces it.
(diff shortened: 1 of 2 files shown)


time-attack:
Confirmed against master: `BUNDLED_PACK_NAMES` in `src/core/schema-pack/bundled.ts` ships all 7 packs and `mutate.ts` refuses every one of them with `PACK_READONLY`, so both skill docs were genuinely stale. Since skills are instructions agents act on, a wrong pack list here produces real failed mutations — merging. Thanks for keeping the list tied to the actual registry.

PR #2355, opened 2026-06-22
fix(op-checkpoint): build completed_keys via to_jsonb($3::text[]) — satisfy v119 array CHECK

chriskedz:
## Problem

`recordCompleted()` in `src/core/op-checkpoint.ts` bound `JSON.stringify(sorted)` to a `$3::jsonb` cast. On the direct session pool, `executeRawDirect → runUnsafe → postgres.js conn.unsafe(sql, params)` JSON-encodes the string parameter a **second** time, so the cast yields a jsonb **scalar string** (`"[...]"`, `jsonb_typeof = 'string'`) instead of an array.

Migration **v119** (`op_checkpoints_completed_keys_array_check`) added `CHECK (jsonb_typeof(completed_keys) = 'array')`. So after v119, every `recordCompleted` write of a non-empty array fails the constraint:

```
[op-checkpoint] write failed (sync-target, …): new row for relation "op_checkpoints"
[sync] checkpoint target write failed (pool unavailable) — aborting before import; banked 0 file(s)
```

This breaks **`gbrain sync`'s pin write** (the `sync-target` checkpoint at `src/commands/sync.ts:1956`) — large syncs abort before importing and bank 0 files every run — and every other `recordCompleted` caller (embed/extract checkpoints). The `"pool unavailable"` wording is misleading; the underlying error is the CHECK violation, not pool exhaustion.

The existing in-code comment claiming the `JSON.stringify`-into-`$3::jsonb` form was correct ("NOT the double-encode trap") was wrong for the `.unsafe()` path.

## Fix

Bind the array as a Postgres `text[]` and build the JSONB array in SQL with `to_jsonb($3::text[])` — the **same binding `APPEND_PATHS_SQL` already uses** in this file, and the pattern `CLAUDE.md` mandates ("JSONB: never `JSON.stringify` into a `::jsonb` cast"). One line of SQL + pass the array directly instead of pre-stringifying.

## Why the unit suite didn't catch it

PGLite does **not** double-encode, so `test/op-checkpoint.test.ts` passes on both the broken and fixed code — an engine-parity gap. This PR adds `test/e2e/op-checkpoint-postgres.test.ts` (skips when `DATABASE_URL` is unset; runs in the Postgres CI matrix) asserting `recordCompleted` persists `jsonb_typeof = 'array'` and round-trips through the loader, including the REPLACE-semantics path. It's scoped to a reserved `__test_ckpt_pg__` op and self-cleans.

## Verification

- Empirically reproduced against a real Postgres brain: the old `JSON.stringify`→`$3::jsonb` form is **REJECTED** by the v119 CHECK; the new `to_jsonb($3::text[])` form stores a proper `array`. After the fix, live `op_checkpoints` rows land as arrays and `gbrain sync` resumes banking paths.
- `bun run typecheck` clean.
- `test/op-checkpoint.test.ts` (PGLite): 30 tests pass.
- The new e2e test passes the skip path locally (no `DATABASE_URL`); it could not be exercised against Postgres on my machine because the shared session pooler was saturated by a concurrent backfill, but its assertions mirror the reproduced behavior and the established `*-postgres.test.ts` pattern.

## Note for maintainer

I don't own the release queue, so I did **not** bump `VERSION`/`package.json` or add a `CHANGELOG` header — please assign the version and prefix the PR title per the version-first convention on merge.

### DIFF
--- a/src/core/op-checkpoint.ts
+++ b/src/core/op-checkpoint.ts
@@ -179,19 +179,26 @@ export async function recordCompleted(
   // REPLACE semantics (kept deliberately — #1794 V3). Callers like
   // extract-conversation-facts serialize a MUTABLE map through here and rely on
   // stale keys being REMOVED; an append would make them unremovable. The full
-  // set lands in the parent `completed_keys` JSONB column via a single UPSERT —
-  // exactly as before. JSON.stringify into `$3::jsonb` is correct (the text→jsonb
-  // cast yields a proper array; NOT the double-encode trap, which is the template
-  // form). Sync uses `appendCompleted` (below) instead, never this.
+  // set lands in the parent `completed_keys` JSONB column via a single UPSERT.
+  //
+  // Bind the array as a Postgres `text[]` and build the JSONB array in SQL with
+  // `to_jsonb($3::text[])` — the SAME binding APPEND_PATHS_SQL uses. Do NOT
+  // `JSON.stringify(sorted)` into a `$3::jsonb` cast: postgres.js's `.unsafe()`
+  // path (executeRawDirect → runUnsafe) JSON-encodes the string parameter a
+  // SECOND time, so the cast yields a jsonb *scalar string* (`"[...]"`), not an
+  // array. That trips migration v119's `jsonb_typeof(completed_keys) = 'array'`
+  // CHECK and the whole write fails (empirically confirmed; was silently fine
+  // pre-v119 and on PGLite, which hides the double-encode). Sync uses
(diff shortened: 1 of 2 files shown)


garrytan-agents:

Thanks for the report/fix on `completed_keys via to_jsonb($3::text[])`. Verified fixed on 0.42.53.0.

garrytan:

Thank you @chriskedz for the independent fix — several of you found this same bug in parallel. Crediting contributors in the v0.42.53.0 release notes. Closing as superseded.

PR #2358, opened 2026-06-22
fix(sync): op_checkpoints write a JSONB array, not a string scalar, on Postgres

lubosxyz:
## Problem

Every incremental `gbrain sync` against a **native-Postgres** brain that has real work to import hard-fails:

```
[op-checkpoint] write failed (sync-target, <fp>): new row for relation "op_checkpoints"
[sync] checkpoint target write failed (pool unavailable) — aborting before import; nothing drained
Sync PARTIAL at <commit>: imported 0 of N file(s), reason=checkpoint_unavailable
```

The `pool unavailable` text is a hardcoded fallback reason in `sync.ts`; the real error is the constraint violation. Sync never advances `last_commit`, so the brain silently stops ingesting new content (no-op syncs and initial full syncs still "succeed", which is why this can hide for a while).

## Root cause

`recordCompleted()` (`src/core/op-checkpoint.ts`) — the sync-target pin write — binds `JSON.stringify(sorted)` as a **positional parameter** to a `$N::jsonb` cast:

```ts
engine.executeRawDirect(
   VALUES ($1, $2, $3::jsonb, now()) ON CONFLICT ...`,
  [key.op, key.fingerprint, JSON.stringify(sorted)]);  // e.g. '["<commit-sha>"]'
```

`PostgresEngine.executeRawDirect` → `postgres.js` `conn.unsafe(sql, params)`. `.unsafe()` sends inferred type `0` for the JS string, the server then **describes the `$N::jsonb` parameter as json/jsonb via the cast**, and `postgres.js` serializes the bind value with its JSON serializer (`JSON.stringify`). The already-stringified JSON array gets **double-encoded into a JSONB string scalar** (`"[\"<commit-sha>\"]"`), so `jsonb_typeof = 'string'`.

Migration **v119** added `CHECK (jsonb_typeof(completed_keys) = 'array')`, which then rejects the write. (Pre-v119 it silently stored a scalar; v119 turned the latent corruption into a hard sync stop.)

This is **postgres.js library behaviour**, not a connection-config quirk — the pools only configure `bigint` handling / prepare / timeouts, no custom JSON serializer. PGLite does **not** double-encode, so the PGLite path and the PGLite test coverage are unaffected.

Reproduced directly against Postgres:

```
sql.unsafe('SELECT jsonb_typeof($1::jsonb) t',       [JSON.stringify(['x'])]) -> 'string'   ← bug
sql.unsafe('SELECT jsonb_typeof($1::text::jsonb) t', [JSON.stringify(['x'])]) -> 'array'     ← fix
```

## Fix

Cast `$3::text::jsonb` so the parameter is sent as text and Postgres parses the JSON text into a real array. One-token change; `NULL` / `ON CONFLICT ... EXCLUDED` paths are unaffected (`recordCompleted` always passes `JSON.stringify(sorted)` of a string array). This is more conservative than passing the JS array directly — `executeRawJsonb()` already rejects top-level arrays as a cross-driver footgun, and `::text::jsonb` avoids array-binding ambiguity entirely.

The read/append paths are **not** affected: `loadOpCheckpoint()` guards legacy scalar rows before `jsonb_array_elements_text`, and `APPEND_PATHS_SQL` binds paths as `text[]`, not jsonb.

## Test

Adds `test/e2e/op-checkpoint-jsonb-postgres.test.ts` (skipped without `DATABASE_URL`, like every other `test/e2e/` file) asserting `jsonb_typeof(completed_keys) = 'array'` after `recordCompleted()` and a `loadOpCheckpoint` round-trip. Verified against a real Postgres:

- old `$N::jsonb` → **2 fail** (`... violates check constraint "op_checkpoints_completed_keys_array"`)
- new `$N::text::jsonb` → **2 pass**

PGLite-based `test/op-checkpoint.test.ts` cannot cover this (no double-encode on PGLite), hence the dedicated Postgres e2e file.

## Scope note (deliberately focused)

This PR fixes the **checkpoint hard-stop** only. The same positional-`JSON.stringify(...)` + `$N::jsonb` pattern exists at other sites (e.g. `conversation-parser/llm-base.ts`, `commands/agent.ts`, `commands/sources.ts`, `core/sources-ops.ts`, `minions/handlers/subagent.ts`, `search/query-cache.ts`, `code-intel/traversal-cache.ts`). Those don't hard-fail (no array/object CHECK) but can **silently** store JSONB string scalars on Postgres; some readers defensively parse them, which masks rather than fixes. Worth a follow-up — and note `scripts/check-jsonb-pattern.sh` only catches interpolation-style `JSON.stringify(x)::jsonb`, not positional `executeRaw(..., [JSON.stringify(x)])`, so it doesn't flag this class. Left out of this PR to keep the sync fix small and reviewable.

### DIFF
--- a/src/core/op-checkpoint.ts
+++ b/src/core/op-checkpoint.ts
@@ -180,14 +180,17 @@ export async function recordCompleted(
   // extract-conversation-facts serialize a MUTABLE map through here and rely on
   // stale keys being REMOVED; an append would make them unremovable. The full
   // set lands in the parent `completed_keys` JSONB column via a single UPSERT —
-  // exactly as before. JSON.stringify into `$3::jsonb` is correct (the text→jsonb
-  // cast yields a proper array; NOT the double-encode trap, which is the template
-  // form). Sync uses `appendCompleted` (below) instead, never this.
+  // exactly as before. JSON.stringify the array, then cast `$3::text::jsonb`:
+  // postgres.js (current pin) serializes a bare string param destined for jsonb
+  // AS a jsonb string scalar (double-encode), so `$3::jsonb` yielded
+  // jsonb_typeof='string' and tripped migration v119's array CHECK — hard-failing
+  // every sync-target write. Forcing `::text::jsonb` makes Postgres parse the
+  // JSON text into a real array. Verified: $3::jsonb→'string', $3::text::jsonb→'array'.
   const sorted = [...keys].sort();
   return durableWrite(engine, key, 'write', () =>
     engine.executeRawDirect(
       `INSERT INTO op_checkpoints (op, fingerprint, completed_keys, updated_at)
-       VALUES ($1, $2, $3::jsonb, now())
+       VALUES ($1, $2, $3::text::jsonb, now())
        ON CONFLICT (op, fingerprint) DO UPDATE
          SET completed_keys = EXCLUDED.completed_keys,
              updated_at     = now()`,
--- a/test/e2e/op-checkpoint-jsonb-postgres.test.ts
+++ b/test/e2e/op-checkpoint-jsonb-postgres.test.ts
@@ -1,0 +1,82 @@
+/**
+ * Regression: op_checkpoints.completed_keys must be a JSONB ARRAY on Postgres.
+ *
+ * recordCompleted() binds `JSON.stringify(sorted)` as a positional parameter to
+ * a `$N::jsonb` cast. On postgres.js (`^3.4.0`) the server describes that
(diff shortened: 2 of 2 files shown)


lubosxyz:
Superseded by #2360, which is the single complete fix: the same one-line `op_checkpoints` `$N::text::jsonb` cast, the Postgres e2e regression test (now folded in), the full sweep of the other 25 positional `JSON.stringify(...)`→`$N::jsonb` sites in the same bug class, and a hardened `scripts/check-jsonb-pattern.sh` that catches the positional form. Closing this in favor of the consolidated PR.

Issue #1861, opened 2026-06-04
extract links --stale crashes with `malformed array literal` on calendar/meeting pages (addLinksBatch text[] serialization)


garrytan:
## Summary

`bun run src/cli.ts extract links --source brain --stale` crashes partway through (~page 2200/353K) with a Postgres `malformed array literal` error. The payload in the error is the serialized `context` array from `addLinksBatch`, full of calendar-event text: Zoom URLs (`?pwd=...`), commas, em-dashes, and brace-looking fragments.

This blocks the **entire** `--stale` re-extraction sweep — one bad batch kills the run, so no edges get reconciled. It surfaced today when `LINK_EXTRACTOR_VERSION_TS` was bumped to `2026-05-31`, marking everything stale and forcing a full re-extract.

## Crash signature

```
malformed array literal: "{"(YC) [pin] https://ycombinator.zoom.us/j/95178948505?pwd=YmdFRWxXbWZadlNkaG9iNC9CYW12QT09, YC-SF-560-2-3 (15) [Z] — with [Mark Thurman](../../people/mark-thurman.md), [Doug Duhaime](../../people/doug-duhaime.md), [Kat Bernstein](../../people/..."
  at _addLinksBatchOnce -> sql` ... unnest(${contexts}::text[]) ... `
```

The braces in the error are postgres-js's own serialization of the JS `contexts: string[]` into a Postgres `text[]` literal. One or more `context` strings contain characters (embedded quotes, backslashes, the `,` delimiters inside long calendar event descriptions) that break the `::text[]` cast — Postgres parses the serialized literal and rejects it as malformed.

## Location

`src/core/postgres-engine.ts` -> `_addLinksBatchOnce()` (~line 2528):

```ts
const contexts = links.map(l => l.context || '');
...
FROM unnest(
  ${fromSlugs}::text[], ${toSlugs}::text[], ${linkTypes}::text[],
  ${contexts}::text[], ...
)
```

The `context` field for calendar/meeting edges carries the full raw event line (location + Zoom URL + attendee link list), which is exactly the kind of string that trips array-literal escaping.

## Why it's a real bug

1. **One poisoned batch aborts the whole `--stale` sweep.** No partial progress, no skip — the run dies. Today it meant a graph re-sync (needed to drop stale edges) could not complete at all.
2. It's data-dependent and silent until a calendar page lands in a batch, so it recurs on every future `--stale` run / version bump.

## Repro

```bash
cd /data/gbrain
[...]

PR #2360, opened 2026-06-22
fix(postgres): cast $N::text::jsonb for all positional JSON.stringify jsonb writes + harden guard

lubosxyz:
Single, complete fix for a postgres.js JSONB double-encode bug class. (Supersedes #2358, now closed — its one-line checkpoint fix and its Postgres e2e regression test are folded in here.)

## The bug class

postgres.js describes a `$N::jsonb` **positional parameter** as json/jsonb (via the cast), then serializes the bind value with its JSON serializer (`JSON.stringify`). When the value at position N is **already a JSON string** (`JSON.stringify(...)`), it gets serialized again — double-encoded into a JSONB **string scalar** on native Postgres.

- On most columns this is **silent corruption**: `jsonb_typeof` is `'string'` instead of `'object'`/`'array'`; readers that defensively `JSON.parse` mask it.
- On `op_checkpoints.completed_keys` it is a **hard failure**: migration v119's `CHECK (jsonb_typeof(completed_keys) = 'array')` rejects the write, so every incremental `gbrain sync` with real work aborts (`checkpoint target write failed (pool unavailable)`, imports 0). This silently stops a Postgres brain from ingesting new content.

PGLite does not double-encode, so PGLite paths and PGLite test coverage never saw it.

Mechanism, confirmed by reading the installed `postgres@3.4.9` source: `.unsafe(sql, params)` sends inferred type `0` for a JS string; the server reports the `$N::jsonb` parameter as json/jsonb via the cast; postgres.js then serializes the bind value with `JSON.stringify`. A pre-stringified array `["sha"]` becomes the JSONB string scalar `"[\"sha\"]"`.

## The fix

Cast `$N::text::jsonb` at **every** positional `JSON.stringify(...)` → `$N::jsonb` site, so the parameter is sent as text and Postgres parses the JSON text into real jsonb:

| area | file |
|---|---|
| sync pin (the hard-stop) | `core/op-checkpoint.ts` |
| source config | `commands/sources.ts` (6), `core/sources-ops.ts` (2) |
| minion/subagent | `core/minions/handlers/subagent.ts` (6), `commands/agent.ts` |
| caches | `core/search/query-cache.ts` (3), `core/conversation-parser/llm-base.ts`, `core/code-intel/traversal-cache.ts`, `core/chunkers/symbol-resolver.ts` |
| misc | `core/onboard/impact-capture.ts`, `core/cycle/calibration-profile.ts`, `core/takes-quality-eval/receipt-write.ts`, `core/destructive-guard.ts` |

**Why `::text::jsonb` and not "pass the object directly":** `executeRawJsonb()` already does the object-direct path correctly and is **unaffected** (it passes JS objects, serialized once → `jsonb_typeof='object'`). Several payloads here are top-level **arrays** (checkpoint keys, candidate ids, search results), which `executeRawJsonb` deliberately rejects as a Postgres-array-literal footgun (#1861). `::text::jsonb` keeps the existing `JSON.stringify`, works for both object and array payloads, and is the smallest possible per-site diff.

## Test

`test/e2e/op-checkpoint-jsonb-postgres.test.ts` (skipped without `DATABASE_URL`, like every other `test/e2e/` file) asserts `jsonb_typeof(completed_keys) = 'array'` after `recordCompleted()` plus a `loadOpCheckpoint` round-trip. Verified on real Postgres: **fails on bare `$N::jsonb` (2 fail, the v119 CHECK violation), passes on `$N::text::jsonb` (2 pass).**

## Guard

`scripts/check-jsonb-pattern.sh` only caught the `${JSON.stringify(x)}::jsonb` **template-interpolation** form — which is exactly why this **positional** class slipped through. Hardened with a call-scoped check that flags any `executeRaw`/`executeRawDirect` binding a bare `$N::jsonb` together with `JSON.stringify(` (`executeRawJsonb(` is exempt by construction). The guard found `symbol-resolver.ts` after a first manual pass missed it; it now passes (`OK: no positional JSON.stringify + $N::jsonb double-encode pattern`).

## Note

`core/chunkers/symbol-resolver.ts` is flagged **binary by git already on `master`** (pre-existing, not introduced here), so GitHub renders "Binary files differ". The actual change is one line: `'candidates', $1::jsonb` → `'candidates', $1::text::jsonb` inside the `jsonb_build_object(...)` UPDATE. All edits are inside SQL template strings (no logic change).


garrytan-agents:

Thanks for the report/fix on `cast $N::text::jsonb for all positional sites`. Verified fixed on 0.42.53.0.

garrytan:
Superseded by #2375 (v0.42.53.0), which shipped this exact approach: the repo-wide positional `$N::text::jsonb` sweep plus the AST guard (`scripts/check-jsonb-params.mjs`) wired into CI. Verified on master — the guard runs clean and every site this PR targeted already casts correctly (`src/core/destructive-guard.ts:193` uses a literal jsonb concat, not a double-encode, so it was never affected).

Thanks @lubosxyz for the thorough sweep + guard design — crediting in the v0.42.53.0 release notes. Closing as superseded.

PR #2361, opened 2026-06-22
fix(remediation): terminate onboard --auto loop on a terminally-failed step

DarkNightForge:
## Problem

`gbrain onboard --auto` (and `doctor --remediate`) can spin forever, making no forward progress, when a remediation step's job reaches a terminal **non-completed** state — most easily reproduced by **cancelling** an in-flight job mid-run, but the same trap fires for any `failed`/`dead` step whose underlying health metric never improves.

## Root cause

The orchestrator loop (`runRemediation`, `src/core/remediation/run.ts`) does a per-step recheck (D7): after each step it recomputes recommendations from a fresh health snapshot and re-introduces any recommendation that still fires. A step that terminally failed is recorded in `abortedIds`, but the recompute **did not exclude it**:

```ts
recs = computeRecommendations(freshHealth, ctx).filter((r) => r.status === 'remediable');
```

Because idempotency keys are content-stable (same `job`+`params` → same key) and `queue.add` is already called with `max_attempts: 2`, re-dispatching an aborted step hits the queue's exactly-once fast path and returns the **same terminal row** every iteration. `waitForCompletion` returns it immediately, the step is re-aborted, the recompute re-emits it… and with `maxJobs` defaulting to `Infinity` the loop never terminates.

## Fix (caller-side; queue contract untouched)

The `idempotency_key` comment always said replay is *"caller responsibility — handled by the --remediate loop."* So the fix lives in the loop, not the shared primitive:

- **`selectActiveRecs(recs, abortedIds)`** (new, pure, in `remediation-step.ts`) — drops non-remediable steps **and** ids already in `abortedIds`. Used at the D7 recompute so a terminally-failed step is attempted **at most once per run**, then abandoned until the next scheduled run (which starts with a fresh abort set). This makes loop progress monotonic.
- **Top-of-loop guard** (`skipped_already_aborted`) — an explicit, locally-checkable backstop.
- **Queue idempotency left as-is.** `minions/queue.ts`'s exactly-once contract (unique partial index + `ON CONFLICT DO NOTHING`, with the deliberate "re-submit must not mutate the existing row" note) is intentional and shared by every caller (autopilot fan-out, crons, …). Changing it would be the wrong layer with a wide blast radius. The `:r<N>` retry-suffix replay referenced by several stale comments was **never implemented** and is unnecessary given queue-level `max_attempts`; those comments are updated to match reality.

## Testing

- `test/remediation-loop-termination.test.ts` (new) pins the termination invariant at the decision boundary: a re-emitted aborted step is filtered out so the active set strictly shrinks. (Full-orchestrator e2e with real Minion handlers stays deferred upstream per the existing TODO in `test/e2e/onboard-full-flow.test.ts` — the per-handler stub seam isn't landed yet.)
- `bun run typecheck` clean.
- `bash scripts/run-verify-parallel.sh` — 30/30 checks green (incl. `check:no-double-retry`).
- `bun test` on `remediation-step`, `brain-score-recommendations`, `doctor`, `minions`, `onboard-full-flow` + the new test — all pass (0 fail).

Versioning / CHANGELOG left to the maintainer's ship flow (community contribution).


time-attack:
This PR's branch has fallen behind a fast-moving master (fork branch, cannot be updated in-repo). The change is still needed — the fix has NOT landed on master — so it was taken over: re-applied 3-way onto current origin/master with author credit (Co-authored-by) in #3255, which supersedes this PR. All local gates green (typecheck, touched unit tests 137/137, onboard-full-flow e2e 13/13).

DarkNightForge:
Thanks for carrying these forward!

time-attack:
Great diagnosis — this was a real hang, but master already closes it from the other side: the D7 recompute now filters on `attemptedIds` (`run.ts`, landed in #2854 and relanded as #3342), and since every terminal path adds the step id to `attemptedIds`, an aborted step can't be re-emitted. Your `selectActiveRecs` would be a narrower version of the same guard, so we're closing as already fixed. Thanks for the thorough write-up.

PR #2362, opened 2026-06-22
fix(onboard): gate extract-ner / timeline recs on whether the action can actually help

DarkNightForge:
## Problem

`gbrain onboard --check --explain` keeps surfacing remediations that can **never clear** on a given brain, because they're emitted purely on a coverage threshold with no check that the recommended job is *capable* of moving that metric:

- **`extract-ner`** is recommended whenever `entity_link_coverage` is low — even when the active schema pack declares no `link_types[].inference.regex` rules. `extractNerLinks` then early-returns `pack_unavailable` and creates 0 links, so coverage never moves and the rec re-fires forever. (Reproduces on brains resolving to **`gbrain-base-v2`**, which declares **zero** `inference.regex` rules. Note `gbrain-base` — the tier-7 default — *does* declare four, at `gbrain-base.yaml:337-350`; an earlier revision of this PR had these two packs backwards.)
- **`extract-timeline-from-meetings`** is recommended whenever `timeline_coverage` is low — even when there are **no dated meeting pages** to extract from. The handler skips meetings without `effective_date`, so it creates 0 entries and the rec persists.

Net effect: `onboard --check` nags with auto-eligible recommendations that are structural no-ops, even when `doctor` is otherwise green. An operator (or an automated remediation agent) runs the direct equivalents, sees them no-op, and the recommendations stay — confusing, and it makes `onboard --auto` burn cycles on work that can't help.

## Fix — gate each rec on whether the action can do anything

- **`packSupportsNerInference(pack)`** — extracted into `src/core/schema-pack/best-effort.ts` (co-located with the pack loader). It's the exact capability `extractNerLinks` already checks (`some link_type with inference.regex`), and the handler now calls it too, so recommender and handler cannot disagree about what a given pack can do.
- **`loadActivePackForLocalEngine(engine)`** — new shared resolver in `best-effort.ts`, used by **both** the recommender and the `extract-ner` handler. Pins `remote: false` and pairs `loadConfigFileOnly()` with the engine's DB-side `schema_pack`. Sharing only the predicate was not enough: the two sides could still resolve *different packs* and disagree about the same brain (see the DB-flip test below). `loadActivePackBestEffort` is untouched — its ~12 other callers keep their existing contract.
- **`resolveNerInferenceCapability(engine)`** (file-local to `checks.ts`) wraps that resolver and returns **three** states, not a boolean: `supported` / `no_rules` / `unresolved`.
- **Timeline gate** — `checkTimelineCoverage` only emits the rec when `COUNT(*) FROM pages WHERE type='meeting' AND effective_date IS NOT NULL > 0`, matching the handler's own precondition.

When a gate withholds the rec, the check **stays WARN** (coverage really is low — honest, and `doctor`'s exit-code contract is unchanged) and the message explains *why* there's no auto-fix. `no_rules` and `unresolved` produce **different** messages, so a pack that failed to load can never masquerade as a pack that simply lacks NER rules. The merged coverage branches preserve the original severity/est_seconds tiers via ternaries.

## Review round 2 — the three blockers

1. **Trust context.** The gate no longer calls `loadActivePackBestEffort({ engine } as never)`. That helper defaults `remote: ctx.remote ?? true`, which put a *local* onboard check behind remote trust gating, and a trust-gate rejection then returned `null` and silently suppressed a legitimate rec. Now `remote: false`, and — separately — the residual `null` case is **surfaced with its own message** rather than collapsed into "no rules". The original objection was that the fix traded a nag for a silent failure; the three-state return is the answer to that, not just the `remote` flag.
2. **Pack-resolution path.** Now `loadConfigFileOnly()` + `engine.getConfig('schema_pack')` + `remote: false`, matching `checkPackUpgradeAvailable` / `checkTypeProliferation` two functions down, so a DB-side pack flip resolves identically for all three. Deliberately *not* refactored into one helper shared with those two siblings: their outer `catch` returns a distinguishable `Check skipped: <message>`, which a null-swallowing helper would flatten into "No active pack". Documented at the call site.

   Chasing this exposed a fourth defect you didn't list, and it inverts the bug: `extract-ner.ts` used the same `loadActivePackBestEffort({ engine } as never)` shape (pre-existing on master). With the recommender fixed and the handler left alone, a **tier-4 DB-side pack flip** made them resolve different packs — recommender sees `gbrain-base-v2` and withholds, handler resolves the tier-7 default `gbrain-base` and would happily run. A *missed* recommendation: the phantom's mirror image. Both sides now share `loadActivePackForLocalEngine`, and there's a test that fails if either side drifts off it.
3. **Env-dependent test.** Every pack-sensitive case now runs inside `withEnv({ GBRAIN_HOME: emptyHome(), GBRAIN_SCHEMA_PACK: <pack> })`. `GBRAIN_SCHEMA_PACK` is **tier 2** of the 7-tier chain, so it outranks the DB config (tier 4), the home config (tier 6) *and* the default (tier 7) — the pin holds on any machine. Both fixture packs are already bundled, so no fixture files were added. The missing **positive direction** (pack *with* regex → rec emitted) is now covered, plus two pin-sanity tests asserting each fixture pack's premise directly, so the withheld-rec case can't start passing vacuously if `gbrain-base-v2` ever gains a regex rule.

Also fixed: the `gbrain-base` / `gbrain-base-v2` mix-up in this PR body and in the test comments (verified on disk: `gbrain-base.yaml` has 4 `regex:` rules, `gbrain-base-v2.yaml` has 0). Title still lacks the version-first prefix — left for the maintainer's `/ship` with the version bump, as before.

## Testing

Rebased onto current `master` (the branch was 681 commits behind; the bug is unchanged on master, and the rebase was conflict-free).

- `bun run typecheck` → clean.
- `test/onboard-phantom-rec-gates.test.ts` → **14 pass / 0 fail**, and identically green in three environments: this machine (home pins `gbrain-base-v2`), a **clean `GBRAIN_HOME`** (the probe that failed the previous revision 6/1), and a hostile outer env that pins `GBRAIN_SCHEMA_PACK=gbrain-base` before the run.
- **Discrimination:** restoring `origin/master`'s `checks.ts` (the always-recommend behavior) under a clean `GBRAIN_HOME` → **EXIT=1, 8 pass / 3 fail** — the two withheld-rec gates plus the `unresolved` message. It now discriminates for the right reason rather than because of the author's home config.
- **Parity, the anti-drift invariant:** paired assertions that `rec emitted ⇔ handler can run`, on `gbrain-base`, on `gbrain-base-v2`, and across a DB-side pack flip. Reverting *only* the handler to `loadActivePackBestEffort` fails the DB-flip case (`recommender can-help=false` vs `handler can-run=true`), so the test discriminates on the shared resolver rather than restating the predicate.
- Neighbours green under clean home: `onboard-pack-upgrade-checks` (9), `remediation-run-extras.serial` (1), `links-timeline-jsonb-poison` (21), `e2e/onboard-full-flow` (13), `e2e/type-unification-full-flow` (3). All six files in **one** process → 61 pass / 0 fail; `schema-pack-sync.test.ts` first (the `__setPackLocatorForTests` shard hazard) → 24 pass / 0 fail.
- Broad sweep over all `schema-pack*` + `*extract*` suites (73 files) → **1080 pass / 1 fail**. The single failure is `test/extract-fs-source-id.test.ts` › "CLI fs path auto-resolves the source from the registered dir", and it **reproduces on clean `origin/master`** with my changes stashed — pre-existing, not from this PR.
- `scripts/run-verify-parallel.sh` **not** re-run this round (forks 30+ parallel checks).

Companion to #2361 (which terminated the `onboard --auto` *loop*); this stops the loop from being *fed* un-satisfiable recommendations in the first place. Versioning/CHANGELOG left to the maintainer's ship.


time-attack:

The fix itself is sound: a shared packSupportsNerInference predicate gates both the onboard recommender (checks.ts:182) and the extract-ner handler (extract-ner.ts:111), and the new timeline gate query exactly matches the handler's preconditions (type='meeting', effective_date IS NOT NULL, deleted_at IS NULL; handler skips undated meetings at extract-timeline-from-meetings.ts:147). Discrimination test passed: stubbing the old always-recommend behavior back into checks.ts makes 2 of the 3 engine tests fail (EXIT=1). BUT the headline NER-gate test is environment-dependent: it never isolates GBRAIN_HOME or pins the schema pack, so it silently reads the operator's real ~/.gbrain/config.json. On a clean machine/CI runner the active pack defaults to gbrain-base, which DECLARES four inference.regex link_types (gbrain-base.yaml:337-350) — the rec IS emitted and the test fails at onboard-phantom-rec-gates.test.ts:91 (verified: GBRAIN_HOME=/tmp/clean → 6 pass / 1 fail). The PR body's reproduction claim ('gbrain-base, which has no NER inference rules') is factually wrong — the regex-less pack is gbrain-base-v2; the test only passes on machines whose home config happens to pin a regex-less pack (the author's, and this reviewer's). No CI checks ever ran on this branch, so this was never caught. Sibling suite onboard-pack-upgrade-checks.test.ts shows the exact correct pattern: withEnv({ GBRAIN_HOME: emptyHome(), GBRAIN_SCHEMA_PACK: undefined }) plus __setPackLocatorForTests. Required changes: (1) isolate the NER-gate test's pack resolution (withEnv + pin a regex-less fixture pack via GBRAIN_SCHEMA_PACK or __setPackLocatorForTests); (2) add the missing positive case (pack WITH regex → rec emitted), which would have caught this instantly — the timeline gate has both directions, the NER gate only one; (3) correct the gbrain-base claim in the PR body and test comments to gbrain-base-v2; (4) title lacks the version-first prefix per the repo IRON RULE (maintainer /ship can handle with the version bump, which the PR intentionally defers). No invariant violations: no engine/JSONB/migration/knobs-hash surface touched, brain-wide SQL matches the file's documented sourcescope:file-brain-wide policy, typecheck clean, all neighboring onboard suites green on this machine.

### Discrimination test

Stubbed the old always-recommend behavior back into src/core/onboard/checks.ts (if (true || packSupportsNerInference(pack)) and if (true || datableMeetings > 0)), re-ran bun test test/onboard-phantom-rec-gates.test.ts: EXIT=1, both withheld-rec gate tests failed (5 pass / 2 fail). The tests discriminate on this machine — but the NER one discriminates for an accidental reason: this machine's ~/.gbrain/config.json pins gbrain-base-v2 (no regex rules). Second probe with GBRAIN_HOME pointed at an empty dir (simulating clean CI): the unmodified PR branch itself fails that test (6 pass / 1 fail at line 91), proving the test is environment-dependent rather than a portable pin of the fix.

**Mergeability at review time:** MERGEABLE / UNSTABLE (no checks reported on the branch — CI never ran)

### Blockers

**1.** test/onboard-phantom-rec-gates.test.ts:85-94 — the NER-gate test reads the operator's real ~/.gbrain/config.json (no GBRAIN_HOME isolation, no GBRAIN_SCHEMA_PACK pin, no __setPackLocatorForTests). Failing input: any clean machine/CI runner with no home config — pack resolution defaults to gbrain-base, packSupportsNerInference returns TRUE (gbrain-base.yaml:337-350 declares regex inference for founded/invested_in/advises/works_at), the rec is emitted, and expect(remediations).toHaveLength(0) fails at line 91. Reproduced: GBRAIN_HOME=/tmp/pr2362-clean-home bun test → 6 pass / 1 fail.

**2.** PR body + test comment misidentify the reproduction fixture: 'gbrain-base ... has no NER inference rules' is false; the regex-less pack is gbrain-base-v2. On a truly fresh brain (default pack gbrain-base) the NER rec behavior is unchanged by this PR — the phantom only affects brains pinned to regex-less packs.

**3.** Missing positive-direction engine test for the NER gate (pack WITH inference.regex → rec emitted). The timeline gate tests both directions; the NER gate tests only the withheld path, which is why the environment dependence went unnoticed.

### Not verified

- scripts/run-verify-parallel.sh 30/30 claim — not reproduced (execution-footprint rule: forks 30+ parallel checks)
- Real-Postgres behavior of the new COUNT query — PGLite-only here; query shape is trivial and mirrors existing safeCount sites, low risk
- Whether gbrain init writes schema_pack: gbrain-base-v2 to fresh configs (would determine how many real fresh brains actually hit the phantom) — not traced

- `gh pr view 2362 --json mergeable,mergeStateStatus → MERGEABLE / UNSTABLE (no CI checks ever reported on the branch)`
- `gh pr diff 2362 (read all 312 lines)`
- `git fetch origin pull/2362/head:pr-2362 && git checkout pr-2362 (32 commits behind master, merges cleanly per GitHub)`
- `bun run typecheck → EXIT=0`
- `bun test test/onboard-phantom-rec-gates.test.ts → 7 pass / 0 fail (author-machine-like env)`
- `discrimination stub in checks.ts + re-run → EXIT=1, 2 gate tests fail; stub reverted`
- `bun test test/onboard-phantom-rec-gates.test.ts test/remediation-run-extras.serial.test.ts → 8 pass / 0 fail`
- `bun test test/onboard-pack-upgrade-checks.test.ts test/e2e/onboard-full-flow.test.ts → 22 pass / 0 fail`
- `GBRAIN_HOME=/tmp/pr2362-clean-home bun test test/onboard-phantom-rec-gates.test.ts → EXIT=1, 6 pass / 1 fail (NER gate test fails at line 91 on clean home)`
- `bun probe-pack.ts → active pack resolves via home-config to gbrain-base-v2 (supportsNer=false) on this machine; default resolution is gbrain-base (registry.ts:120), which has regex rules`
- `grep audits: extractNerLinks callers (extract.ts:949, jobs.ts:2134 — both route through the internal gate), extract-ner recommender sites (only checks.ts), timeline handler preconditions (extract-timeline-from-meetings.ts:72-78,147), gbrain-base.yaml:322-350 link_types, test seams (__setPackLocatorForTests, withEnv/emptyHome pattern in onboard-pack-upgrade-checks.test.ts)`
- `cat VERSION → 0.42.66.1 on branch (no bump; PR defers versioning to maintainer /ship)`

</details>

---

time-attack:
Review finding (blocking): the phantom-recommendation bug is real and still on master, and the shared-predicate shape (one `packSupportsNerInference` consumed by both recommender and handler) is the right root-cause fix. Three defects block it:

1. The gate calls `loadActivePackBestEffort({ engine } as never)`, and that helper defaults `remote: ctx.remote ?? true` — so a **local** onboard check runs under remote trust gating, and a pack-load rejection returns null and silently suppresses a legitimate recommendation. That trades the nag for a new silent failure, against the fail-closed trust invariant. Pass a real local context (`remote: false`).
2. It bypasses the pack-resolution path #2538 established for this exact file (`loadConfigFileOnly()` + `engine.getConfig('schema_pack')`), so a DB-side pack flip resolves differently for your gate than for the sibling checks two functions down.
3. The new test never sandboxes `GBRAIN_HOME`/`withEnv`, so it reads the real `~/.gbrain/config.json` — on any runner whose config pins a pack with NER inference rules the test fails. That can turn master red; same env-dependence #2538's second commit had to retrofit onto the neighboring tests.

time-attack:
Status update: reviewed in the 2026-08-18 full-backlog triage — still wanted, but not yet mergeable: Bug still on master and fix shape endorsed, but three blocking review defects (trust ctx, pack-resolution path, env-dependent test) are unanswered. If you update it, it will be re-reviewed promptly; alternatively the fix may be adopted with credit in an upcoming fix wave. Thanks for the contribution and the patience.

DarkNightForge:
Thanks for the triage ping — all three blockers are addressed, and the branch is rebased onto current `master` (it was 681 commits behind; conflict-free, and the bug is unchanged on master).

**1. Trust context.** Dropped `loadActivePackBestEffort({ engine } as never)`. But the `remote: false` flag alone doesn't answer your objection — you said it "trades the nag for a new silent failure," and a genuine pack-load failure would still return `null` and still silently eat the rec. So the gate now returns **three** states, not a boolean: `supported` / `no_rules` / `unresolved`, with a **distinct message** for `unresolved`. A pack that failed to load can no longer masquerade as a pack that simply declares no NER rules. There's a test pinning exactly that.

**2. Pack-resolution path.** Now `loadConfigFileOnly()` + `engine.getConfig('schema_pack')` + `remote: false`, matching `checkPackUpgradeAvailable` / `checkTypeProliferation` two functions down. One judgement call: I did **not** fold those two siblings into the shared helper, because `checkPackUpgradeAvailable`'s outer `catch` returns a distinguishable `Check skipped: <message>` that a null-swallowing helper would flatten into "No active pack". Happy to unify if you'd rather take that behavior change.

**2b. A fourth defect, found by chasing #2 — and it inverts the bug.** `extract-ner.ts` used the same `loadActivePackBestEffort({ engine } as never)` shape (pre-existing on master, which is why I'd first scoped it out). But fixing only the recommender makes the two sides resolve the pack by *different paths*, and the shared predicate does not save you: on a **tier-4 DB-side pack flip**, the recommender sees `gbrain-base-v2` and withholds while the handler resolves the tier-7 default `gbrain-base` and would happily run. That's a *missed* recommendation — the phantom's mirror image. Reproduced, then fixed: both now go through a new `loadActivePackForLocalEngine` in `best-effort.ts` (`remote: false` + file-only config + DB-side `schema_pack`). `loadActivePackBestEffort` itself is untouched, so its ~12 other callers keep their contract.

**3. Env-dependent test.** Your reproduction was correct and I reproduced it before touching anything. All pack-sensitive cases now run inside `withEnv({ GBRAIN_HOME: emptyHome(), GBRAIN_SCHEMA_PACK: <pack> })`. I pinned via `GBRAIN_SCHEMA_PACK` rather than `__setPackLocatorForTests` because it's **tier 2** of the chain — it outranks the DB config (4), home config (6) and the default (7), so the pin holds regardless of the runner, and both fixture packs are already bundled so no fixture files were needed. Added the missing positive direction (pack *with* regex → rec emitted), plus two pin-sanity tests that assert each fixture pack's premise directly, so the withheld-rec case can't start passing vacuously if `gbrain-base-v2` ever gains a regex rule.

Verified green in three environments — this machine (`gbrain-base-v2` in home config), a **clean `GBRAIN_HOME`** (your 6/1 probe), and a hostile env pinning `GBRAIN_SCHEMA_PACK=gbrain-base` before the run — 14 pass / 0 fail in all three.

The anti-drift claim is now asserted as a paired invariant (`rec emitted ⇔ handler can run`) on both packs and across a DB-side flip. Reverting *only* the handler to `loadActivePackBestEffort` fails the DB-flip case, so it discriminates on the shared resolver rather than restating the predicate.

**Discrimination, re-done under a clean home:** restoring `origin/master`'s `checks.ts` → `EXIT=1, 8 pass / 3 fail` (both withheld-rec gates + the `unresolved` message). It now discriminates on the fix rather than on my home config.

Neighbours under clean home: `onboard-pack-upgrade-checks` 9, `remediation-run-extras.serial` 1, `links-timeline-jsonb-poison` 21, `e2e/onboard-full-flow` 13, `e2e/type-unification-full-flow` 3 — all 0 fail. All six in one process → 61/0; with `schema-pack-sync.test.ts` first (the locator-stub shard hazard) → 24/0. Broad sweep over all `schema-pack*` + `*extract*` suites (73 files) → 1080 pass / 1 fail; that one is `test/extract-fs-source-id.test.ts` › "CLI fs path auto-resolves the source from the registered dir", which **reproduces on clean `origin/master`** with my changes stashed — pre-existing, flagging rather than fixing it here. `typecheck` clean. I did not re-run `run-verify-parallel.sh` this round.

**Your factual correction was right, and it was load-bearing:** `gbrain-base.yaml:337-350` declares 4 `inference.regex` rules; `gbrain-base-v2.yaml` declares 0. Since `gbrain-base` is the tier-7 default, the phantom only bites brains pinned to a regex-less pack — the opposite of what the PR body claimed. Body and test comments corrected.

One thing left alone deliberately: the title's version-first prefix, yours at `/ship` with the version bump, as before.

DarkNightForge:
One follow-up on a loose end from your review — *"No CI checks ever ran on this branch, so this was never caught."* Now I know why, and it isn't something I can fix from this side.

The push just queued four runs on `80baad95` (Test, E2E Tests, Heavy Tests, Semgrep), and all four immediately came back `conclusion: action_required` with a 0s duration. Same for the 2026-07-28 runs. That's the fork-PR approval gate: workflows on pull requests from forks don't execute until a maintainer approves them, so `gh pr checks` reports "no checks reported" and `mergeStateStatus` sits at `UNSTABLE` forever.

So the env-dependence you found was structurally *un*catchable by CI on this branch — the gate meant the test never ran anywhere but the author's machine and yours, which is exactly the condition that let it hide. Worth knowing that it applies to every fork PR in the backlog, not just this one.

If you approve the runs on `80baad95` I'll act on whatever they turn up. My local numbers, for comparison: `typecheck` clean, target suite 14/0 under a clean `GBRAIN_HOME`, six affected files 61/0 in one process.

DarkNightForge:
Correcting my own last comment: **ignore the request to approve the runs.** I asked for the wrong thing.

Two reasons, both from `docs/RELEASING.md`. Fork PRs don't receive base-repo secrets, so approving those runs would just produce empty-env auth failures on the key-dependent jobs — a GitHub security default, not a config gap. And per "Community PR wave process," external PRs aren't merged directly anyway; the fix gets cherry-picked or re-implemented into a wave with `Co-Authored-By` credit and this PR gets closed. So a green check here was never the thing standing between the fix and master, and my "approve and I'll act on it" framing pointed at a dead end.

The mechanics in that comment still hold, and they're the part worth keeping: the four runs on `80baad95` are `action_required`, not absent, so `mergeStateStatus: UNSTABLE` and an empty check-runs list are the gate — not a failed push. Worth noting because *"missing CI at head"* is a triage rubric bullet a fork PR structurally cannot satisfy on its own.

What that leaves is local verification, which is the part I can actually own. Re-run with exit codes captured to a file rather than piped:

```
bun run typecheck                                  → EXIT=0
bun test test/onboard-phantom-rec-gates.test.ts    → EXIT=0   (14 pass / 0 fail)
bun test <the six affected files, one process>     → EXIT=0   (61 pass / 0 fail)
```

Clean `GBRAIN_HOME` throughout. Discrimination and the DB-flip parity result are in my previous comment. Nothing needed from you beyond the normal wave triage.

time-attack:

PR #2363, opened 2026-06-22
fix(remediation): gate sync/extract recs on real extraction lag, not the stale_pages proxy

DarkNightForge:
## Problem

The `sync.repo` and `extract.all` remediations (and `embed.stale`'s sync prerequisite) gate on `health.stale_pages`, defined in both engines as:

```sql
count(*) FROM pages p WHERE p.updated_at < (SELECT MAX(te.created_at) FROM timeline_entries te WHERE te.page_id = p.id)
```

i.e. "pages whose content `updated_at` predates their newest timeline entry." That was a meaningful proxy only while a trigger bumped `pages.updated_at` on every timeline insert — **a trigger dropped in migration v10** (`drop_timeline_search_trigger`). Since then the metric reflects a near-never condition, yet it still fires the two recommendations with the rationale **"N stale pages on disk"** — which is not what it measures and disagrees with both `gbrain sync` (commit-clean) and `doctor`'s `links_extraction_lag`. Operators see sync/extract recs that don't map to real work.

## Fix — gate on the real extraction lag

- `RecommendationContext` gains **`extractionLagPages`**, populated by `loadRecommendationContext` from **`engine.countStalePagesForExtraction`** — the *same* counter the `gbrain extract --stale` walk processes and doctor's `links_extraction_lag` check reports. So the recommendation fires **iff running `extract` will actually reduce it**, and clears once extraction is current. No re-implemented predicate → no drift between the recommender and the job it recommends.
- `sync.repo` / `extract.all` / `embed.stale`'s sync-dependency now gate on `extractionLagPages`; rationale text is honest (`"N page(s) with stale link/timeline edges"`). The sync→extract pipeline shape is unchanged (sync is the prerequisite that brings pages current before extract materializes their edges).
- `health.stale_pages` is **retained** (CLI health line + back-compat — no API/type break) but marked **legacy** in its docstring and **no longer gates any recommendation**.

## Testing

- `test/brain-score-recommendations.test.ts` updated to drive the gate via `ctx.extractionLagPages` (the pure-function pins for `sync.repo`/`extract.all`/`embed.stale` dependencies + severity ordering + determinism).
- `test/remediation-context-extraction-lag.test.ts` (new): PGLite engine test pins that `loadRecommendationContext` reports 0 on an empty brain and >0 once a page needs extraction.
- `bun run typecheck` clean · `bash scripts/run-verify-parallel.sh` 30/30 · recommender + onboard + advisor/features suites green (0 fail).

Companion to #2361 (terminate the `onboard --auto` loop) and #2362 (capability/content gates) — together they make `onboard --auto` recommend only work that can actually clear. Versioning/CHANGELOG left to the maintainer's ship.

### DIFF
--- a/src/core/brain-score-recommendations.ts
+++ b/src/core/brain-score-recommendations.ts
@@ -146,6 +146,16 @@ export interface RecommendationContext {
   chatModel?: string;
   /** Whether the chat provider has a usable API key. */
   hasChatApiKey?: boolean;
+  /**
+   * Count of pages needing link/timeline extraction — the SAME staleness the
+   * `gbrain extract --stale` walk and doctor's `links_extraction_lag` check use
+   * (`engine.countStalePagesForExtraction`). Gates the sync→extract pipeline
+   * (sync.repo / extract.all). Replaces the old `health.stale_pages` gate, which
+   * counted "pages whose updated_at predates their newest timeline entry" — a
+   * proxy that broke when the updated_at-on-timeline-insert trigger was dropped
+   * (migration v10) and never reflected real extraction work.
+   */
+  extractionLagPages?: number;
 }
 
 /** Triage result for one check. */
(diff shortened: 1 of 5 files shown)


time-attack:
Confirmed the root cause on master: `stale_pages` depends on an `updated_at` bump from the trigger dropped in `drop_timeline_search_trigger` (migrate.ts), so the sync/extract recs fire off a metric that measures nothing and print a rationale that isn't true. Gating on `countStalePagesForExtraction` — the same counter `extract --stale` and doctor's `links_extraction_lag` use — is the right fix, and retaining `stale_pages` for the health line avoids a break. One follow-up nit: `loadRecommendationContext` calls the counter without `versionTs`/`sourceId` while doctor passes `LINK_EXTRACTOR_VERSION_TS`, so the rec can under-fire after an extractor-version bump. Merging.

PR #2370, opened 2026-06-23
v0.42.52.1 fix(doctor): categorize 7 onboard checks so doctor stops warning

johnybradshaw:
## What

`gbrain doctor` prints a `[doctor-categories] unknown check name … defaulting to 'meta'` stderr warning on **every run** — once for each of 7 checks. This categorizes them and closes the gap that let them slip through CI.

## Root cause

`src/commands/doctor.ts` merges `runAllOnboardChecks()` output into the report and categorizes each via `categorizeCheck(c.name)`. But those 7 checks are defined in `src/core/onboard/checks.ts`, and their names were never added to the category sets in `src/core/doctor-categories.ts` — so they fell through to `'meta'` with a warning, and were scored under the wrong category. The drift-guard test only scanned `doctor.ts`, so CI never noticed.

## Change

- Categorize the 7 onboard checks:
  - **brain** (data-health): `embed_staleness`, `entity_link_coverage`, `timeline_coverage`, `takes_count`
  - **meta** (gbrain coherence): `pack_upgrade_available`, `type_proliferation`, `dangling_aliases`
- Extend `test/doctor-categories.test.ts` to also scan `src/core/onboard/checks.ts`, so onboard checks are covered by the drift guard and a future uncategorized onboard check **fails the build instead of warning silently in production**.

## Test plan

- `bun run typecheck` — clean
- `bun test test/doctor-categories.test.ts` — 8 pass, 0 fail
- `bun test` doctor-category-dependent files (brain-checks-score, cause-rank, hidden-by-search-policy) — 39 pass, 0 fail
- `gbrain doctor` — the 7 `unknown check name` lines are gone (0 remaining)

Pre-existing, unrelated `.serial.test.ts` failures in the local environment (worker-registry, brain-durability, brain-registry, doctor-cli-smoke, hybrid-meta, migration) were confirmed identical on clean `master` with these changes stashed — not introduced by this PR (they stem from a live local brain + `OPENAI_API_KEY` set in the shell).


time-attack:
Thanks for chasing this down — the diagnosis was exactly right, but master already carries the fix: all 7 onboard check names are registered in `src/core/doctor-categories.ts` as of #3075 (commit d6fe4863), so `gbrain doctor` no longer emits the unknown-check warnings. Closing as already fixed.

PR #2376, opened 2026-06-23
fix(mcp): whoami over stdio returns local identity instead of unknown_transport

Facundo-Moreno-Dev:
# PR title

fix(mcp): whoami over stdio returns local identity instead of unknown_transport

> Note: this repo's convention prefixes PR titles with the next version
> (`vX.Y.Z.W ...`). I left that off because an external branch can't know the
> next `/ship` version. Add it when you bump, e.g.
> `v0.42.53.0 fix(mcp): whoami over stdio ...`.

# PR body

## Problem

Calling `whoami` over the **stdio MCP transport** (`gbrain serve`, the
Claude Desktop / Claude Code default) always throws:

```
unknown_transport: whoami called over a remote transport that did not thread
ctx.auth. This is a transport bug — every remote call site must populate
ctx.auth or set ctx.remote === false.
```

The stdio pipe is `remote: true` (untrusted, per the trust boundary) but
carries **no per-token auth by design** — a local pipe has no OAuth/bearer
token (see the comment at `src/mcp/server.ts`). `whoami` required `ctx.auth`
for any `remote !== false` call, so it could never succeed over stdio. The
contract had no way for stdio to satisfy it: it can't set `remote: false`
(that would loosen trust) and it has no auth to thread.

## Fix

Thread an explicit `transport: 'stdio' | 'http'` signal from the MCP dispatch
sites into `OperationContext`. `whoami` now returns a local, unprivileged
identity for the stdio pipe:

```json
{ "transport": "local", "scopes": [] }
```

Why this is safe:

- **No trust change.** `ctx.remote` stays `true` for stdio — every
  untrusted-caller protection (`file_upload` confinement, the `takes`
  holder allow-list `['world']`, source isolation) is unchanged.
- **No privilege.** `scopes: []` grants nothing; consumers that gate on
  scopes get zero.
- **HTTP stays fail-closed.** The `unknown_transport` throw is preserved for
  an auth-less **HTTP** transport, which is a genuine bug — the relaxation is
  scoped to the stdio pipe only.
- Anyone who can talk to the stdio pipe already has local execution as the
  user (they spawned the process), so reporting "local" reveals/grants
  nothing new.

## Changes

| File | Change |
|---|---|
| `src/core/operations.ts` | add `OperationContext.transport`; `whoami` stdio branch returns local identity |
| `src/mcp/dispatch.ts` | `DispatchOpts.transport` + thread onto context in `buildOperationContext` |
| `src/mcp/server.ts` | stdio server tags `transport: 'stdio'` (remote stays true) |
| `src/mcp/http-transport.ts`, `src/commands/serve-http.ts` | tag `transport: 'http'` for consistency |
| `test/whoami.test.ts` | stdio returns local + grants nothing; HTTP without auth still fail-closed |

## Tests

- `bun test test/whoami.test.ts` → 12 pass / 0 fail (9 existing + 3 new).
- `bun run typecheck` → clean.

No `VERSION` / `CHANGELOG` bump — left to your `/ship` flow to avoid conflicts.

### DIFF
--- a/src/commands/serve-http.ts
+++ b/src/commands/serve-http.ts
@@ -1601,6 +1601,7 @@ export async function runServeHttp(engine: BrainEngine, options: ServeHttpOption
       try {
         toolResult = await dispatchToolCall(engine, name, params as Record<string, unknown> | undefined, {
           remote: true,
+          transport: 'http',
           takesHoldersAllowList: tokenAllowList,
           sourceId: tokenSourceId,
           metaHook: getBrainHotMemoryMeta,
--- a/src/core/operations.ts
+++ b/src/core/operations.ts
@@ -300,6 +300,19 @@ export interface OperationContext {
    */
   remote: boolean;
   /**
+   * Which MCP transport delivered this call, when it came over MCP.
+   * `'stdio'` is the local pipe (`gbrain serve`) — remote/untrusted by the
+   * trust boundary, but auth-less BY DESIGN (no per-token auth on a local
+   * pipe; see src/mcp/server.ts). `'http'` is the OAuth/bearer HTTP transport,
+   * which threads `auth`. Unset for the trusted local CLI path.
+   *
+   * whoami uses this to distinguish "stdio pipe, legitimately auth-less"
+   * (report transport: 'local') from "HTTP transport that dropped auth"
+   * (a real bug → fail-closed unknown_transport). Do NOT use this to grant
+   * trust — `remote` remains the trust discriminator.
+   */
+  transport?: 'stdio' | 'http';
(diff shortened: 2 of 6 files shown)


time-attack:
Your analysis matches what master already does: `OperationContext.transport` exists and `whoami` returns `{ transport: 'stdio', scopes: [] }` for the auth-less local pipe (landed for #1061), with the `unknown_transport` throw kept for auth-less HTTP and `ctx.remote` still `true`. Since the behavior you're fixing is already in place, we're closing this one. Thanks for the careful trust-boundary reasoning.

PR #2377, opened 2026-06-24
feat(autopilot): operator opt-out — exclude a source from the knowledge cycle via config.autopilot_cycle=false

JiraiyaETH:
## What

Adds a per-source operator opt-out from the autopilot knowledge cycle. A source whose config sets `autopilot_cycle: false` is excluded from autopilot's per-source work, so a repository kept current by a **dedicated external sync pipeline** is never re-dispatched by autopilot — regardless of staleness.

## Why

When a source repo is maintained by its own out-of-band sync (e.g. a nightly `gbrain sync --strategy code` job), autopilot's per-source cycle becomes a second writer: it re-imports with the default `markdown` strategy and runs per-source extract / atom-drain on it. That fights the dedicated pipeline over the `last_commit` bookmark and pulls unwanted content into the brain. There was no first-class way to say "this source is managed elsewhere — leave it out of the cycle."

## What it does

Enabled-by-default; only the literal boolean `false` opts out, so existing sources are unaffected. Gated behind a shared `isAutopilotCycleEnabled()` helper (mirrors `isSourceFederated`) at all three per-source dispatch points:

- fan-out source selection (`selectSourcesForDispatch`) — surfaced as a new `skipped_cycle_disabled` bucket in the fan-out result + `--json` telemetry
- the freshness-sync loop
- the per-source atom auto-drain loop

It also teaches the three `gbrain doctor` per-source freshness checks (`cycle_freshness`, `sync_freshness`, `federation_health`) to skip excluded sources — otherwise an opt-out source would stand-FAIL forever (its `last_full_cycle_at` stops advancing), depress `brain_score` (which feeds back into autopilot's own `shouldFullCycle` gate), and emit misleading "run gbrain dream/sync" remediation.

Operator surface: `gbrain sources cycle|no-cycle <id>` (symmetric with `federate`/`unfederate`), plus a note in `docs/guides/multi-source-brains.md`.

## Tests

- `test/autopilot-fanout.test.ts`: `autopilot_cycle: false` excludes even a maximally-stale source; `true`/unset still participates.
- `test/doctor-cycle-freshness.test.ts`: an excluded stale source returns `ok` (no FAIL, not named, no remediation) and doesn't mask a real FAIL from an included source.
- `bun run typecheck` clean; fan-out + doctor suites green; `scripts/check-jsonb-pattern.sh` passes.

## Notes

- No `VERSION` / `package.json` / `CHANGELOG` bump — left for the maintainer fix-wave per `docs/RELEASING.md`.
- Possible follow-up: the brain-wide global maintenance phases (embed / orphans / symbol-edges) intentionally still cover an excluded source's already-imported content — the opt-out is about the per-source *refresh* cycle, not excising existing content from search / hygiene.

### DIFF
--- a/docs/guides/multi-source-brains.md
+++ b/docs/guides/multi-source-brains.md
@@ -109,6 +109,25 @@ behave exactly as before — every page appears in search.
 
 Flip later with `gbrain sources federate <id>` / `unfederate <id>`.
 
+## Autopilot cycle flag
+
+Every source row also stores `config.autopilot_cycle: boolean` in its JSONB
+config, controlling whether autopilot manages the source.
+
+| Value | Meaning |
+|-------|---------|
+| unset (default) / `true` | Source participates in the autopilot knowledge cycle: per-source sync + extract + full cycle. |
+| `false` | Source is excluded from autopilot — skipped at every per-source dispatch site and by the matching doctor freshness checks. |
+
+The use case is a source kept current by a dedicated external sync pipeline
+(say `my-code-repo`, re-synced by its own CI) rather than autopilot's
+per-source loop. Opting it out stops the two pipelines from racing the same
+source while leaving search, federation, and manual `gbrain sync <id>`
+untouched.
+
+Flip later with `gbrain sources no-cycle <id>` (opt out) / `cycle <id>`
+(re-enable, the default — clears the flag).
+
 ## Commands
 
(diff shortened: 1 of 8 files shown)


time-attack:
Reviewed and declined for now with the content-scope set. Thank you.

PR #2379, opened 2026-06-24
v0.42.53.0 fix(extract): pack-gate body-verb inference so packs that omit a verb don't get it

JiraiyaETH:
## Problem

GBrain infers relationship verbs (`founded`, `invested_in`, `advises`, `works_at`) from page **body prose** using deterministic regexes in `inferLinkType` (`src/core/link-extraction.ts`). Those regexes ran **without consulting the active schema pack**. The default `extractPageLinks` body paths call the legacy `inferLinkType` directly — not the additive pack-aware `inferLinkTypeFromPack` (which only ever *adds* pack verbs on top of the legacy matchers; it cannot *suppress* them).

Result: a brain whose active pack does **not** declare a verb (e.g. a pack with no `invested_in` in its `link_types`) still acquired `invested_in` edges whenever the regex misfired on narrative prose. The pack was already authoritative for frontmatter-derived edges and page types — but **not** for body-inferred verbs. This PR closes that gap.

## Fix

Body-verb inference is now **pack-gated**. A new pure helper `gateInferredVerb(verb, allowlist?)` in `link-extraction.ts` gates the four prose-guessed verbs against an optional `inferredVerbAllowlist` — the set of `link_type` names the active pack declares — threaded into `extractPageLinks`:

- verb in `{founded, invested_in, advises, works_at}` **and** the pack declares it → keep it;
- same verb **and** the pack omits it → degrade to `mentions`;
- **structural** verbs (`mentions`, `attended`, `image_of`) are **never** gated — they're set by a page's own type, not guessed from prose, so a pack that merely omits `attended` keeps its meeting↔attendee edges;
- **no allowlist** (`undefined` → no pack loaded) → return the verb unchanged, preserving the legacy no-pack path exactly.

Threaded at all three body-inference call sites, each loading the active pack's `link_type` names once:
- `put_page` auto-link (`src/core/operations.ts`, `runAutoLink`) — reuses the pack already loaded once per `put_page`;
- `gbrain extract links` (`extractLinksFromDB`) and `gbrain extract --stale` (`extractStaleFromDB`) in `src/commands/extract.ts` — via a shared `loadInferredVerbAllowlist(engine, sourceIdFilter)` that returns `undefined` on any pack-load failure (safe back-compat default).

## Backward-compat mechanism + evidence

**Mechanism chosen: simple pack-gating (the preferred non-breaking option).** I determined that the bundled default `gbrain-base` pack (`src/core/schema-pack/base/gbrain-base.yaml`) already declares **all four** gateable verbs in its `link_types` (`founded`, `invested_in`, `advises`, `works_at`, plus the structural `attended` / `image_of` / `mentions`). So:

- **Default installs are unaffected** — every inferred verb is in the allowlist, the gate is a pass-through.
- **Packs that omit a verb correctly suppress it** — which is the whole point of the fix.
- **No-pack / pack-load-failure path is preserved** — `undefined` allowlist → un-gated legacy behavior.

No new pack flag, no opt-in config, no migration — the gate is inert for the default pack and active only where a custom pack deliberately omits a verb.

## Test evidence

New focused test `test/link-extraction-pack-gated.test.ts` (generic placeholders only — `widget-co`, `a-founder`) proves:
- (a) a pack **without** `invested_in` suppresses a regex-inferred `invested_in` → `mentions` (both the markdown-ref and bare-slug call sites);
- (b) a pack **with** `invested_in` keeps it;
- (c) the no-pack path (no allowlist) is unchanged;
- (d) the real `gbrain-base` manifest declares all four verbs and gating against it keeps the inferred verb (default path unaffected);
- plus `gateInferredVerb` unit assertions, including that structural verbs are never gated even against an empty allowlist.

Results (per CLAUDE.md "never pipe tests through tail" — file output then tail, real exit captured):
- New suite: **11 pass / 0 fail**.
- `test/link-extraction*.test.ts` + `frontmatter-inference` + `link-inference-pack` + `schema-pack-page-to-link` + `regressions/gbrain-base-equivalence`: **200 pass / 0 fail**.
- `test/extract*.test.ts` + `operations-schema-pack`: **275 pass / 0 fail**.
- `bun run typecheck`: clean (EXIT 0).
- `bash scripts/run-verify-parallel.sh`: 30/30 checks green.


time-attack:
Thanks for the thorough analysis and the clean no-pack fallback — the engineering here is careful. I'm closing it on scope, though: I checked `gbrain-base.yaml` and it declares all four gateable verbs, so as your own write-up notes the gate is a pass-through for every default install. That makes this new plumbing across `operations.ts` and `extract.ts` (plus an extra pack load per run) surface that only activates for a custom pack that deliberately omits a verb, which isn't a gap I'm taking new API for. It also overlaps the same call sites as #2408, which fixes an inference bug that does affect every install.

PR #2380, opened 2026-06-24
fix(search): normalize `/` in FTS queries so slash-containing terms match

rwbaker:
## Problem

Keyword search silently returns **zero hits** for any query containing a `/`.

Postgres' default text-search parser classifies `foo/bar` as a single `file`-alias token. Under the `english` config that alias maps to the `simple` dictionary, so `websearch_to_tsquery('english', 'foo/bar')` produces the single un-stemmed lexeme `'foo/bar'`, which never matches indexed chunk text whose `to_tsvector('english', …)` split on the slash. Pasting a meeting/title fragment like `Author1/Author2 - Topic` hits this every time, while the same query with the slash removed matches fine.

Reproduce:
```sql
SELECT to_tsvector('english','NovaMind builds AI agents') @@ websearch_to_tsquery('english','NovaMind/AI');  -- f
SELECT to_tsvector('english','NovaMind builds AI agents') @@ websearch_to_tsquery('english','NovaMind AI');   -- t
```

## Fix

Add `normalizeKeywordQuery()` in `src/core/search/keyword.ts` (replace `/` with whitespace, collapse runs) and route the FTS-bound query params through it in both engines before `websearch_to_tsquery`:
- `PostgresEngine.searchKeyword` / `searchKeywordChunks`
- `PGLiteEngine.searchKeyword` / `searchKeywordChunks`

The slash-split tokens then stem and match through the english dictionary exactly as the manually-despaced query does. The semantic / vector path is untouched; quoted phrases (`"foo/bar"`) still degrade to phrase queries (`foo <-> bar`), which is at least as permissive as the current no-hits baseline.

## Tests

Adds a PGLite regression test (`test/pglite-engine.test.ts`) covering a leading `NovaMind/AI` query and a mid-query `AI NovaMind/agents` query; both now return the matching page where they previously returned nothing. Full `bun test test/pglite-engine.test.ts` passes (101 tests).

### DIFF
--- a/node_modules
+++ b/node_modules
@@ -1,0 +1,1 @@
+/paperclip/instances/default/gbrain/code/node_modules
--- a/src/core/pglite-engine.ts
+++ b/src/core/pglite-engine.ts
@@ -59,6 +59,7 @@ import {
   EmbeddingColumnNotRegisteredError,
 } from './search/embedding-column.ts';
 import { hasCJK, escapeLikePattern } from './cjk.ts';
+import { normalizeKeywordQuery } from './search/keyword.ts';
 
 type PGLiteDB = PGlite;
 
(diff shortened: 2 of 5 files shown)


time-attack:
Thanks — this is a real one. Confirmed on current master that both engines bind the raw query into `websearch_to_tsquery`, so a `/` collapses to a single file-alias lexeme and keyword search returns nothing; the fix is small, symmetric across both engines, and the PGLite regression test pins it. One thing must be dropped before this lands: the diff commits a `node_modules` symlink at the repo root pointing at an absolute machine-local path (`/paperclip/instances/default/gbrain/code/node_modules`). Remove that file and this is good to go.

PR #2381, opened 2026-06-24
fix(salience): exclude briefings/* from their own Brain Pulse

rwbaker:
## Problem

The daily briefing writes its output back into the vault, where it re-ingests as `briefings/<date>`. On the next run that freshly-touched page tops `getRecentSalience()` (the Brain Pulse), so the briefing perpetually surfaces **itself** as the most salient recent activity — pure self-reference with no real signal. Over time the Pulse becomes a list of past briefings instead of the underlying knowledge that changed.

## Fix

Exclude `briefings/%` slugs from `getRecentSalience()` in both engines (`PostgresEngine`, `PGLiteEngine`), *unless* the caller explicitly scopes the query to the `briefings/` prefix (in which case they asked for them). The suppression is a single `AND p.slug NOT LIKE 'briefings/%'` clause gated on the requested `slugPrefix`.

## Tests

Adds PGLite E2E coverage in `test/e2e/salience-pglite.test.ts`:
- default query hides `briefings/*` slugs
- explicit `slugPrefix='briefings/'` still returns them

`bun test test/e2e/salience-pglite.test.ts` passes (7 tests).

### DIFF
--- a/src/core/pglite-engine.ts
+++ b/src/core/pglite-engine.ts
@@ -5328,6 +5328,11 @@ export class PGLiteEngine implements BrainEngine {
       params.push(escaped);
       prefixCondition = `AND p.slug LIKE $${params.length} ESCAPE '\\'`;
     }
+    // Exclude briefing pages from their own Brain Pulse — see the matching
+    // block in postgres-engine.ts getRecentSalience() for the rationale.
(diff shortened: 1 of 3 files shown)


time-attack:
Duplicate of #1202. Same salience/briefings fix; keep #1202.

PR #2382, opened 2026-06-24
fix(anomalies): suppress cohorts without enough baseline history

rwbaker:
## Problem

`computeAnomaliesFromBuckets` flags a cohort whenever today's count clears the small-sample fallback (`count > mean + 1`). On a fresh import — or any cohort whose first real activity is *today* — the baseline window is all zero-fill, so **every brand-new cohort fires at once** and the Brain Pulse / `anomalies` report reads "everything is anomalous at 481σ." That Day-1 noise buries the genuine spikes the feature exists to surface.

## Fix

Add a `min_baseline_days` guard: a cohort must have at least that many *distinct active* baseline days (zero-fill days don't count) before it can be flagged.

- `computeAnomaliesFromBuckets(...)` gains a trailing `minBaselineDays = 0` param — **no behavior change** for direct callers.
- `PostgresEngine.findAnomalies` / `PGLiteEngine.findAnomalies` default the operation to **7**.
- The `find_anomalies` operation exposes `min_baseline_days` so callers can tune it or disable the guard (`0`).

## Tests

- `test/anomalies.test.ts` — unit coverage: suppress below threshold, fire at/above it, zero-fill days don't count toward active days, disabled at `0`.
- `test/e2e/anomalies-pglite.test.ts` — default `7` suppresses the Day-1 wedding cohort; `min_baseline_days: 0` reproduces the legacy firing behavior.

While updating the E2E fixture I also made its timestamps deterministic: it previously stamped pages via the server-local DB clock (`now()`) but derived the anomaly window from `new Date().toISOString()` (UTC), so the suite flaked whenever a run straddled a UTC midnight boundary. Pages are now anchored to the same UTC day the window uses.

`bun test test/anomalies.test.ts test/e2e/anomalies-pglite.test.ts` → 25 passing.

### DIFF
--- a/src/core/cycle/anomaly.ts
+++ b/src/core/cycle/anomaly.ts
@@ -50,29 +50,39 @@ export function meanStddev(samples: number[]): { mean: number; stddev: number } 
  *
  * For each cohort:
  *   1. Compute (mean, stddev) over the baseline daily counts.
- *   2. If stddev > 0:    anomalous when `today.count > mean + sigma*stddev`.
+ *   2. If the cohort has fewer than `minBaselineDays` distinct days with
+ *      activity in the baseline window, **suppress it**. A
+ *      brand-new cohort or one with one day of import noise would otherwise
+ *      produce a 481σ "everything is anomalous" report.
+ *   3. If stddev > 0:    anomalous when `today.count > mean + sigma*stddev`.
  *      sigma_observed   = (today.count - mean) / stddev.
- *   3. If stddev == 0:   small-sample fallback — anomalous when
+ *   4. If stddev == 0:   small-sample fallback — anomalous when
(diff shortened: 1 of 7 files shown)


time-attack:
Merging — the min-baseline guard is a correctness fix, not a tuning knob: on master a brand-new cohort has an all-zero-fill baseline and clears the `count > mean + 1` small-sample fallback, so on any fresh import every cohort fires at once and the report surfaces nothing usable. Defaulting the new parameter to 0 in the pure helper keeps direct callers unchanged, both engines move in lockstep, and `min_baseline_days: 0` preserves an escape hatch. The UTC-anchoring of the E2E fixture is a welcome extra — that midnight-straddle flake was real.

Issue #2081, opened 2026-06-11
extract links: globalBasename resolution exists but no CLI flag exposes it — cross-folder bare wikilinks underivable


victorfteha:
## Summary

`extract links` has working machinery for resolving bare wikilinks by basename across folders — `opts.globalBasename` (extract.ts ~307–313: "Ancestor walk misses + globalBasename on + basename hits → all matches") — but no CLI flag wires it, so it's unreachable in practice.

## Why it matters

Obsidian's default link style is bare `[[Page Title]]` regardless of folder. In a foldered vault (`10-people/`, `11-companies/`, `12-meetings/`, ...), most wikilinks point across folders, and the ancestor-walk-only default derives **zero** edges for them. Measured on a real vault: hundreds of dangling cross-folder refs that `extract links` couldn't resolve, all of which had unambiguous basename targets — the only path was scripting manual `add_link` calls.

## Ask

A `--global-basename` flag on `extract links` (opt-in, default off to preserve current behavior). Ambiguity policy can stay exactly as the existing code documents (return all matches / skip ambiguous).

time-attack:
Closing as fixed on current master. The premise ('no flag wires globalBasename, unreachable in practice') is false on current master: isGlobalBasenameEnabled (src/core/link-extraction.ts:1244-1254) resolves env GBRAIN_LINK_RESOLUTION_GLOBAL_BASENAME → DB config `link_resolution.global_basename` → default false, and every extract surface (fs, db, put_page auto-link) reads it (src/commands/extract.ts:1030,1148,1307,1383). `gbrain doctor` prints the exact enable command when the brain would benefit (link_resolution_opportunity check). A per-run --global-basename CLI flag would be redundant sugar and per-run toggling churns wikilink-resolved edge reconciliation; config/env is the intended surface.

If this still reproduces on the latest release, comment with details and we will reopen.

Issue #1964, opened 2026-06-08
extract links fails to resolve cross-directory wikilinks due to slug/path mismatch


harrylabsj:
## Version
gbrain 0.42.29.0

## Summary
`gbrain extract links` (both `--source fs` and `--source db`) processes all pages but creates 0 links, because wikilink targets in markdown files use original filenames (with spaces, mixed case) while gbrain stores pages with slugified paths.

## Steps to reproduce

1. Have a markdown page at `llm-wiki/entities/AI 3.0.md` (stored in gbrain as slug `llm-wiki/entities/ai-3.0`)
2. Have an MOC page `llm-wiki/_meta/books-by-domain-ai-technology.md` containing:
   ```
   - [[llm-wiki/entities/AI 3.0]] — 梅拉妮·米歇尔
   ```
3. Run `gbrain sync --all && gbrain extract links --source fs --dir /path/to/vault`
4. Check: `gbrain backlinks "llm-wiki/entities/ai-3.0"` → `[]`
5. Check: `gbrain graph "llm-wiki/entities/ai-3.0" --depth 1` → `"links": []`

## Observations

- Some wikilinks within `_meta/` resolve fine (e.g. `[[llm-wiki/_meta/books-by-author]]`), but links pointing to `entities/` do not
- The files definitely contain wikilinks (verified with grep), but gbrain's link graph remains empty
- `gbrain orphans --count` reports 1263/2198 pages as orphans (57%), which appears to be because the link extraction never populated the link graph for cross-directory references

## Expected behavior
Wikilinks written in Obsidian format `[[path/with spaces/and MixedCase.md]]` should resolve to gbrain's slugified page IDs (`path/with-spaces/and-mixedcase`).

time-attack:
Closing as fixed on current master (verified today against v0.42.70.0+, and the close was independently re-checked by an adversarial second pass before this comment). Evidence: Commit 78391e8b "fix(extract): resolve cross-directory wikilinks via sync-consistent slugification (#3161)" (on origin/master, verified via git merge-base --is-ancestor). link-extraction.ts now imports slugifyPath from sync.ts; test/link-extraction.test.ts contains tests named "#1964" using the issue's exact repro `[[llm-wiki/entities/AI 3.0]]`. Fixup 0a6b6970 further pins the #1964 assertions.

If this still reproduces for you on the latest release, comment and we will reopen.

Issue #1846, opened 2026-06-04
`extract links` bare-name wikilink resolver inconsistently drops edges on unambiguous targets


Walliiee:
## Summary

`gbrain extract links` (both `--source fs` and `--source db`, `gbrain 0.32.1.0`) inconsistently materializes edges from bare-name `[[wikilinks]]` even when the target slug is unambiguous and the source/target pages are both fully synced into the same source. Some bare-name targets resolve, others don't, with no obvious pattern based on slug shape, file location, or page age.

The same brain shows the resolver working for *some* bare-name links — e.g. `[[acme]]` reliably resolved to `wiki/companies/acme` — while structurally identical links to other pages were silently dropped. Manual `gbrain link` succeeds against the same slugs, so the pages exist and are reachable; the gap is in extract-time wikilink resolution.

## Repro shape

In a brain with these pages (all in the same source, all synced):

```
wiki/companies/acme.md
wiki/companies/widgetco.md
wiki/people/alice.md
wiki/people/bob.md
wiki/people/carol.md
wiki/meetings/alice-2026-02-20.md
```

Each page contains bare-name wikilinks pointing at the others — e.g. `acme.md` has `[[alice]]`, `[[bob]]`, `[[carol]]`, `[[widgetco]]`; person pages have `[[acme]]` and `[[widgetco]]`. Wikilinks live in normal body prose (both prose lines and `- list items`).

Run:

```bash
gbrain sync
gbrain extract links
# Expected: ~10–12 edges materialized across the wikilinks
# Actual:   ~4 edges; ~8 silently dropped
```

`gbrain backlinks <slug>` on the orphaned targets confirms the missing edges aren't there. A second `extract links` run is a no-op (idempotent), so re-running doesn't recover.

## Observed pattern

The resolver succeeded for `[[acme]]` and `[[widgetco]]` (company slugs) but consistently failed for `[[alice]]`, `[[bob]]`, `[[carol]]` (person slugs) in the same brain — including failing links *from* `acme.md`, where on the same line `[[carol]]` was dropped but `[[widgetco]]` later in the line was kept. So it's not a per-page issue; it's a per-target-slug issue.

A path-form wikilink in the same body (`[[meetings/alice-2026-02-20]]`) resolved correctly. Switching the failing bare-name links to `[[wiki/people/alice]]` is a workaround but breaks the established bare-name convention used elsewhere in the brain.

[...]

PR #2384, opened 2026-06-24
feat(extract): PARA-numbered dirs + alias/title/basename wikilink resolution

rwbaker:
## What

Improves `[[wikilink]]` → graph-edge resolution for Obsidian / PARA-organized vaults, where the FS extractor today resolves targets by path-equality only. On such vaults that yields a very low edge ratio (~5%, 45/812 links on a real PARA vault); with this change the same vault resolves **744/812 (≈16×)**.

Two additions, both derived purely from authored content (no heuristic guessing → no false-positive edges):

1. **PARA-numbered directories.** `DIR_PATTERN` now recognizes `\d+_word` prefixes (`10_projects`, `30_resources`, …) that PARA layouts use to force sidebar ordering. Extracted slugs are normalized via `slugifyPath` so PascalCase / spaced segments match the lowercased DB slug. Without this, PARA vaults extract **0** wikilinks despite hundreds in the source (the underlying complaint in #1493).
2. **Alias / title / basename fallback.** When exact path-equality misses, a per-run index resolves the target via the page's frontmatter `aliases:`, its H1 title, or (opt-in) a slug basename. **Basename stays gated behind the existing `link_resolution.global_basename` flag (#972)** so its opt-in semantics are preserved; alias/title are always-on.

## Schema change — the part that needs a maintainer call

`links.resolution_type` already exists with `CHECK (… IN ('qualified','unqualified'))`. To record *which* resolver pinned each edge (audit trail / recall grading), this PR **widens** that CHECK to also allow `'path','alias','title','basename'`:

- `src/schema.sql`, `src/core/pglite-schema.ts`, `src/core/schema-embedded.ts` updated in lockstep
- new idempotent migration `v120` (`links_resolution_type_widen`) drops + re-adds the constraint
- `addLinksBatch` in both engines threads the column through the `jsonb_to_recordset` INSERT

**This is the design decision I'd like your call on before you invest review time.** Three options, happy to refactor to whichever you prefer:

- **(a) widen the existing `resolution_type` CHECK** (this PR) — smallest surface, reuses the column;
- **(b) a separate `resolution_method` column** — keeps `resolution_type` strictly about source-pinning (qualified/unqualified) and puts resolver attribution in its own column;
- **(c) drop the attribution entirely** — keep just the PARA + alias/title resolution behavior, no schema change at all (the edges still get created; we just lose the audit trail).

If you'd rather not touch schema at all, (c) makes this a zero-migration, behavior-only PR.

## Relationship to existing PRs/issues

- Supersedes **#1188** (same feature, but it was branched on May master, now conflicting/`dirty`, and it also bundled unrelated source-id threading). This branch is rebuilt clean on current `master`, is mergeable, and drops the unrelated changes.
- Supersedes **#1186** (PARA dirs), which you closed in the backlog cleanup with an invitation to reopen if the issue is still live — it is: `DIR_PATTERN` on `master` still has no PARA support, so PARA vaults still extract 0 wikilinks.
- Addresses / relates to: #1493 (hardcoded `DIR_PATTERN` drops Obsidian wikilinks silently), #874 & #1964 (cross-directory slug/path mismatch), #1846 (bare-name resolver drops edges on unambiguous targets). Partial overlap with #2081 (basename exists but no CLI flag) and #2215 (PARA orphan overcount).

## Tests

- `test/link-extraction.test.ts` — PARA dir grammar
- `test/extract-wikilink-aliases.test.ts` — all four resolver paths + the `resolution_type` audit column

150 assertions; full `tsc --noEmit` clean.

### DIFF
--- a/src/commands/extract.ts
+++ b/src/commands/extract.ts
@@ -39,11 +39,12 @@ import {
   extractFrontmatterLinks, isGlobalBasenameEnabled, LINK_EXTRACTOR_VERSION_TS,
   WIKILINK_BASENAME_LINK_TYPE,
   buildBasenameIndex, queryBasenameIndex, stripCodeBlocks,
-  type UnresolvedFrontmatterRef, type LinkCandidate,
+  WikilinkAliasIndex, deriveTitleFromContent,
+  type UnresolvedFrontmatterRef, type LinkCandidate, type PageAliasFields,
 } from '../core/link-extraction.ts';
 import { createProgress } from '../core/progress.ts';
 import { getCliOptions, cliOptsToProgressOptions } from '../core/cli-options.ts';
-import { pathToSlug, pruneDir, isSyncable } from '../core/sync.ts';
+import { pathToSlug, pruneDir, isSyncable, slugifyPath } from '../core/sync.ts';
 // v0.41.18.0: withRetry + isRetryableConnError + WithRetryOpts moved to
 // src/core/retry.ts as the canonical primitive. Engine methods
 // (addLinksBatch/addTimelineEntriesBatch/upsertChunks) now self-retry via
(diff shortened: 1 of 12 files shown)


time-attack:
Closing after maintainer review — decision: not accepting this. A confirmed, still-unfixed gap stays unfixed and the work is thrown away.

PR #2385, opened 2026-06-24
fix(cycle): guard extract against empty-queue zero-walk + log symbol-resolve progress

rwbaker:
## What

Two independent robustness fixes to the dream / autopilot cycle, both isolated to `src/core/cycle.ts`:

### 1. Guard the extract phase against an empty-queue zero-walk

The extract phase treats a defined `changedSlugs` array as *incremental* — extract only those slugs. An **empty defined** array (`[]`) means sync reported zero changed pages. On a steady-state brain whose link graph is already built, walking nothing is the correct, fast outcome.

But when an out-of-band process drains git → DB **before** the cycle's own sync runs, sync sees `HEAD == HEAD` and returns `pagesAffected: []` even though links were never extracted. The incremental path then walks zero slugs, and **link coverage stays at zero indefinitely** after a brain rebuild. The same drained-queue path also nerfs downstream phases that key off the change set.

`resolveExtractSlugs()` (new, exported for testing) converts that degenerate `[]` to `undefined` (full walk) **only when** the source has pages but no links originate from it. Brains that already have links keep the fast empty-incremental path, so the common case pays nothing extra. The source id is threaded from the `cycleSourceId` already computed at cycle entry.

### 2. Per-batch progress logging in `resolve_symbol_edges`

`resolveSymbolEdgesIncremental` already exposes an `onProgress` callback, but the cycle phase ignored it — so a slow or wedged phase went dark for its whole duration with no way to tell how far it got. This wires `onProgress` to emit `chunks_walked` / `batches` / `edges_examined` / `resolved` plus per-batch and elapsed timing, leaving a visible last-walked checkpoint in the log.

## Why

Both are real production failure modes: (1) a brain that silently never rebuilds its link graph, and (2) an opaque multi-minute/hour phase with no observability when it stalls.

## Tests

- New `test/cycle-extract-zero-walk-guard.test.ts` (PGLite, in-memory) covers `resolveExtractSlugs`: `undefined` pass-through, non-empty pass-through, empty-with-pages-no-links → full walk, empty-no-pages stays empty, empty-once-links-exist keeps the fast path, and empty-without-sourceId untouched.
- Existing `test/extract-incremental.test.ts` (8 cases) still green.
- The logging change is observability-only (no behavior change).

### DIFF
--- a/src/core/cycle.ts
+++ b/src/core/cycle.ts
@@ -950,11 +950,55 @@ async function runPhaseSync(
   }
 }
 
+/**
+ * Decide the effective slug set for the cycle's extract phase.
+ *
+ * Extract runs in two modes:
+ *  - `changedSlugs === undefined` → full walk (first run / manual).
+ *  - `changedSlugs` defined → incremental: extract only those slugs.
+ *
+ * An *empty defined* array (`[]`) is the degenerate incremental case: sync ran
+ * but reported zero changed pages, so extract walks nothing. On a steady-state
+ * brain whose link graph is already built that is the correct, fast outcome.
+ *
+ * But if an out-of-band process drains git→DB before the cycle's own sync runs,
+ * sync sees HEAD == HEAD and returns `[]` even though links were never
+ * extracted — the incremental path then walks zero slugs and link coverage
+ * stays at zero indefinitely after a brain rebuild. Guard against that: when
+ * handed an empty defined queue, fall back to a full walk if the source has
+ * pages but no links originate from it. Brains that already have links keep the
+ * fast empty-incremental path (no extra work for the common case).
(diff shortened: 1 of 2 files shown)


time-attack:
Merging — confirmed on master that the extract phase treats any defined `changedSlugs` as incremental with no empty-array guard, so a brain whose queue was drained out-of-band never rebuilds its link graph at all. I appreciate that `resolveExtractSlugs` only runs its two COUNT queries in the degenerate empty case and only escalates when pages exist with zero originating links — the steady-state path pays nothing, which is why this is worth taking. The `onProgress` wiring rides along cleanly since the callback already existed and the change is observability-only.

Issue #2200, opened 2026-06-15
Federated read scope not honored by get_page or the semantic query cache (by-slug reads fail in federated sources; cross-client cache leak)


mlobo2012:
Hi @garrytan, two related bugs on current `master` (`543f9a71`, v0.41.21.0). They share one root cause: a client's `federated_read` source set reaches `search` and `query`, but not the final downstream boundary. One affects `get_page`, the other affects the semantic query cache. I found these while running a multi-source OAuth client and confirmed them by static inspection. Happy to send a PR if the proposed fixes look right to you.

## Bug A: `get_page` ignores `federated_read`

A client whose primary `source_id` is `A` and whose `federated_read` set is `['A', 'B']` can find a page that lives only in `B` through `search` and `query`, but `get_page` for that same slug returns not-found. The federated scope is applied to search and query, but not to the actual page fetch.

Evidence (master `543f9a71`):
- `sourceScopeOpts(ctx)` gives `ctx.auth.allowedSources` precedence as `{ sourceIds }` at `src/core/operations.ts:413-421`.
- `search` uses it at `src/core/operations.ts:1224-1228`. `query` builds `querySourceScope` and spreads it into the search call at `src/core/operations.ts:1353-1361` and `1401-1412`.
- `get_page` builds a scalar-only `sourceOpts` at `src/core/operations.ts:464-469`, then calls `ctx.engine.getPage(...sourceOpts)` at `:478` and `:484`. It only uses `sourceScopeOpts(ctx)` for fuzzy slug candidates at `:470-482`, never for the exact fetch.

Repro:
1. Register a client with `source_id='A'` and `federated_read=['A','B']`.
2. Create a page only in source `B`.
3. `search` or `query` returns the `B` page.
4. `get_page` for that slug returns not-found.

Impact: any agent that relies on `get_page` for deterministic by-slug reads cannot read its federated sources, even though it is authorized and can see those pages in search.

Proposed fix: extend `GetPageOpts` with `sourceIds?: string[]`, implement array precedence in `PostgresEngine.getPage` and `PGLiteEngine.getPage`, then have `get_page` use `sourceScopeOpts(ctx)` for both the exact and the fuzzy fetch. Resolve tags via the concrete page's source (`{ sourceId: page.source_id }`).

## Bug B: semantic query cache key under-specifies the federated source scope

[...]

PR #2387, opened 2026-06-24
feat(engine): opt-in Postgres RLS source-scope binding

harrisali0101:
## What this does

Adds opt-in [Postgres Row-Level Security](https://www.postgresql.org/docs/current/ddl-rowsecurity.html) scope binding to the engine's source-scoped read methods. When the environment variable `GBRAIN_RLS_SCOPE_BINDING=1` is set, every source-scoped read is wrapped in a transaction that sets a session variable `app.scopes` from the caller's allowed sources. A matching RLS policy on `pages` (and other source-scoped tables) then filters rows at the SQL layer — even when the application code path doesn't apply the source filter itself.

Default behavior is **unchanged**: with the env var unset, `withScopedReadTransaction` is effectively pass-through (one tx round-trip per read, no semantic change). Operators who don't want this see zero behavior difference on upgrade.

## Why

Defense in depth. Today the engine enforces source scope at the application layer — every read method has to remember to apply the filter. If any code path drifts (the comment in `sourceScopeOpts` calls out "drift between sites is the bug class"), the only safety net is the next code review. RLS gives the database itself a vote: a row's `source_id` not in the caller's grant set never reaches the network buffer.

This was caught when a federated OAuth client read past its grant due to a single op handler that dropped the federated array. The fix was already addressed at the app layer by #2200 (federated_read scope reaches by-slug reads in v0.42.46.0). This PR is the orthogonal defense-in-depth layer that catches the next instance of that bug class before it ships.

## How it works (the helper)

`withScopedReadTransaction(sourceIds?, sourceId?, callback)`:
1. Always opens a transaction. (Lets nested `SET LOCAL` calls in search methods scope cleanly; avoids `tx.begin` not being callable inside a postgres.js transaction handle.)
2. When `GBRAIN_RLS_SCOPE_BINDING=1`, emits `SELECT set_config('app.scopes', $1, true)` with `$1` derived from `sourceIds.join(',') || sourceId || '*'`. The `true` makes it transaction-local (equivalent to `SET LOCAL` but parameterised — `SET LOCAL` itself doesn't accept bound params, so `set_config()` is the right tool).
3. Calls back with the transaction handle.
4. When the env var is off: same wrap, no `set_config` emit. The transaction is the no-op overhead (~ms per read).

## Recommended RLS policy

For operators who turn this on:

```sql
ALTER TABLE pages ENABLE ROW LEVEL SECURITY;
CREATE POLICY pages_scope_filter ON pages
  USING (current_setting('app.scopes', true) = '*'
         OR source_id = ANY(string_to_array(current_setting('app.scopes', true), ',')));

-- For admin / autopilot / cycle connections that shouldn't be source-scoped:
ALTER ROLE <runtime-role> SET app.scopes = '*';
```

The `string_to_array` form handles a single source as well as the comma-separated federated set.

## Layered with #2200 (federated by-slug reads)

This PR sits on top of upstream HEAD which already includes v0.42.46.0 (federated `sourceIds[]` scope on by-slug secondary reads). The two layers compose cleanly:

- **App-layer (existing, v0.42.46.0)**: `getPage` / `getLinks` / `getBacklinks` filter rows with `AND source_id = ANY(${sourceIds}::text[])`.
- **SQL-layer (this PR, opt-in)**: same operations also wrap in `withScopedReadTransaction(sourceIds, sourceId, ...)` so the RLS policy enforces from below.

The three methods touched by both #2200 and this PR's commit `feat(postgres-engine): wrap remaining 16 source-scoped methods` keep both layers active.

## Commits

5 logical commits — left separate for review:

1. `feat(postgres-engine): opt-in RLS scope binding via SET LOCAL app.scopes` — adds the helper, defaulted off.
2. `fix(postgres-engine): use set_config() instead of SET LOCAL for parameterised scope binding` — small follow-up; `SET LOCAL` doesn't accept bound params.
3. `feat(postgres-engine): wrap remaining 16 source-scoped methods with withScopedReadTransaction` — the bulk of the wrap.
4. `fix(postgres-engine): flatten nested tx.begin in search methods + always-open wrap` — handles the postgres.js gotcha that transaction handles expose `.savepoint()` not `.begin()`.
5. `fix(postgres-engine): cast withScopedReadTransaction return to T` — typecheck fixup (`UnwrapPromiseArray<T>` → `T`).

Happy to squash on request.

## Compatibility

- **`GBRAIN_RLS_SCOPE_BINDING` unset / `0`**: helper is pass-through (one extra tx round-trip per read). No RLS policy required. **Default behavior for every existing operator.**
- **`GBRAIN_RLS_SCOPE_BINDING=1`** without an RLS policy on `pages`: still works — `set_config` is called but no policy enforces the value. Safe to enable in isolation before the policy is created.
- **`GBRAIN_RLS_SCOPE_BINDING=1`** + policy created: full defense-in-depth.

## Tests

`bun run verify` — all 30 checks pass (typecheck, privacy, jsonb, progress, test-isolation, wasm, admin-build, resolver, etc.) on the rebased branch against current upstream HEAD.

Manual verification on a Postgres 16 install with the policy applied: opt-in enabled, a federated OAuth client (`allowed_sources={a,b,c,d,e,f}`) confirmed to read across all granted sources via `query` / `get_page` / `list_pages`. A subject-scoped token (`allowed_sources={a}`) confirmed unable to read pages from `b` even when the operation handler is called with `source_id=b` — the RLS policy blocks the row return.

Happy to iterate on any of the above, or split into smaller PRs if you'd prefer (e.g. helper + getPage wrap first, then the remaining 15 wraps as a follow-up).

## Related work

Adjacent and complementary efforts in the same RLS defense-in-depth thread:

- **#2457** — `fix(security): views enforce caller RLS via security_invoker` (open). Flips gbrain's views from default `security_definer` to `security_invoker`, so views enforce the caller's RLS rather than the view owner's. Composes naturally with this PR: their fix ensures policies fire through views; this PR supplies the session-scoped value (`app.scopes`) those policies key off. Two layers, one posture.
- **#2454** — `tag-filtered query/search for profile-scoped retrieval` (open). Read-side scoping at the operations layer; this PR is the SQL-layer enforcement that makes that scoping load-bearing rather than advisory.
- **#2239** / **#2200** — merged app-layer fix (`v0.42.46.0`) where federated `sourceIds[]` reach by-slug reads. This PR is the orthogonal SQL-layer defense to the same threat class — drift in any single operation handler is caught at the database boundary instead of the next review.
- **#612** — `v0.26.8 feat(migration): v35 auto-RLS event trigger — new tables always secure`. Precedent for accepting opt-in RLS infrastructure that's a no-op for the unrestricted case and load-bearing for managed-Postgres deployments. Same shape as the env-var gate proposed here.
- **#1403** — `v0.41.3.0 fix(security/mcp): OAuth CORS lockdown + pre-register without DCR + validator surface`. Same defense-in-depth flavour at the MCP boundary; this PR extends the posture inward to the SQL row layer.

### DIFF
--- a/src/core/postgres-engine.ts
+++ b/src/core/postgres-engine.ts
@@ -156,6 +156,54 @@ export class PostgresEngine implements BrainEngine {
     return db.getConnection();
   }
 
+  // Source-scope binding for Postgres RLS — opt-in via env var.
+  //
+  // When `GBRAIN_RLS_SCOPE_BINDING` is set to `1` / `true`, source-scoped
+  // query methods (listPages, search*, getChunks, etc.) wrap their queries
+  // in a transaction that begins with
+  //   SET LOCAL app.scopes = '<csv-of-allowed-source-ids>'
+  // so Postgres RLS policies on source-scoped tables can filter rows by
+  // `current_setting('app.scopes', true)`. The expected policy shape:
+  //
+  //   USING (current_setting('app.scopes', true) = '*'
+  //          OR source_id = ANY(string_to_array(
+  //             current_setting('app.scopes', true), ',')))
+  //
+  // Recommended runtime-role default:
+  //   ALTER ROLE <runtime-role> SET app.scopes = '*';
+  // so admin / autopilot / cycle queries that don't pass scope info still
+  // see all rows. OAuth-scoped requests override the default per
+  // transaction with their allowed-source CSV.
+  //
+  // Default behavior (env var unset): the helper is a pass-through —
+  // returns `callback(this.sql)` with no transaction wrap and no SET
+  // LOCAL. Behaviorally equivalent to not having this helper at all,
+  // so the feature is safe to land disabled-by-default.
+  private get rlsScopeBindingEnabled(): boolean {
+    const v = process.env.GBRAIN_RLS_SCOPE_BINDING;
+    return v === '1' || v === 'true';
+  }
+
+  private async withScopedReadTransaction<T>(
+    sourceIds: string[] | undefined,
+    sourceId: string | undefined,
+    callback: (tx: ReturnType<typeof postgres>) => Promise<T>,
+  ): Promise<T> {
+    if (!this.rlsScopeBindingEnabled) {
+      return await callback(this.sql);
+    }
+    let scopesValue = '*';
+    if (sourceIds && sourceIds.length > 0) {
+      scopesValue = sourceIds.join(',');
+    } else if (sourceId) {
+      scopesValue = sourceId;
+    }
+    return await this.sql.begin(async (tx: any) => {
+      await tx`SET LOCAL app.scopes = ${scopesValue}`;
+      return await callback(tx as ReturnType<typeof postgres>);
+    });
+  }
+
   // Lifecycle
   async connect(config: EngineConfig & { poolSize?: number; parentConnectionManager?: ConnectionManager }): Promise<void> {
     this._savedConfig = config;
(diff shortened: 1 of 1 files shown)


harrisali0101:
Bump — this has been open since late June with no reviews.

Sits in the same RLS defense-in-depth thread as the merged app-layer fix #2239 and the newer view-level `security_invoker` PR #2457. All three are opt-in and additive with each other:
- #2239 (merged): app-layer federated `sourceIds[]` scope in by-slug reads
- #2457 (open): views enforce caller RLS via `security_invoker`
- **this PR**: opt-in `app.scopes` session-var binding + RLS policies keyed off it

Field-verified on our DIH pilot (single-VM WhatsApp deployment, 1,879 pages across 6 sources). Three-subject wire test 2026-06-29 confirms the enforcement: staff bearer `get_page(mesec)` returns `page_not_found` while super_admin gets the full page — end-to-end at the /mcp endpoint.

Happy to iterate on any concerns. Would love to know what's needed for review readiness. Thanks!

time-attack:
Superseded with credit by #2940 (just merged): your opt-in RLS source-scope binding landed with the security review's fixes applied — flag-off true pass-through (no per-read transaction overhead when disabled), behavioral tests for the set_config emission and precedence, operator docs in ENGINES.md, and your setseed determinism fix preserved. Authorship on the commit via Co-authored-by. Thank you — this is exactly the defense-in-depth layer a shared brain needed, and it passed a line-by-line security review with zero findings of concern.

PR #2392, opened 2026-06-24
v0.42.53.1 fix(pglite): drain in-flight operations before close (#2348)

### LINKED ISSUE #2348: [bug] PGLite engine becomes unusable with 58P01 after dream cycle writes
## Summary
After `gbrain dream` runs the sync phase and writes to PGLite, subsequent gbrain CLI calls (including `gbrain query`) fail with:
\`\`\`
error: relation "content_chunks" does not exist
  code: "58P01"
  file: "dfmgr.c"
  routine: "internal_load_library"
\`\`\`

The PGLite WASM library refuses to load. **Catastrophic** failure — brain is locked and only wipe + restore works.

## Reproduction
\`\`\`bash
# Before dream: gbrain query "anything" → works
gbrain dream  # sync phase writes
gbrain query "anything"  # FAILS with 58P01
\`\`\`

## Attempted fixes that DO NOT work
- \`pkill -f "gbrain serve"\` + restart server
- Delete \`~/.gbrain/brain.pglite/.gbrain-lock/\`
- Delete \`postmaster.pid\` inside pglite data dir

## Only working recovery
\`\`\`bash
mv ~/.gbrain/brain.pglite ~/.gbrain/brain.pglite.wrecked-$(date +%s)
gbrain init --pglite --embedding-model ollama:bge-m3 --embedding-dimensions 1024
gbrain import ~/.gbrain/pages  # need filesystem backup
\`\`\`

## Expected
- PGLite mutex should be released cleanly when CLI exits
- dream sync should not leave PGLite in a corrupted state
- At minimum, a \`gbrain recover\` or \`gbrain doctor --fix\` should be able to repair

## Actual
- Wipe + re-init is the **only** recovery path
- All embedded data (192 chunks × 1024d) is lost without filesystem backup
- **Dangerous for users who don't export before running dream**

## Environment
- gbrain v0.35.1.1
- PGLite v0.4.x (bundled)
- macOS 26.1, Apple Silicon

## Workaround
Always run \`gbrain export --dir ~/.gbrain/pages\` before \`gbrain dream\`. After 58P01, do wipe + init + import.

## Location
PGLite WASM module loading code, or \`src/core/pglite-engine.ts\`.

snvtac:
## Summary

Fixes the PGLite teardown race behind #2348: an aborted JS caller can return before the underlying PGLite WASM query actually settles, and `disconnect()` could then close the database handle while that operation was still running.

This keeps the existing PGLite lifecycle model intact: `_db` is still early-nulled, `db.close()` still runs before lock release, and `close()` remains outside `preservingProcessExitCode`.

## What changed

- Wrap the created PGLite handle so public async DB operations (`query`, `sql`, `exec`, `transaction`) are tracked while in flight.
- Drain tracked in-flight operations before calling `db.close()` during `disconnect()`.
- Preserve native PGLite getter receivers when proxying the handle, so getters backed by private fields (`ready`, `closed`) keep working.
- Add disconnect lifecycle regression coverage for:
  - public getter receiver safety on the tracked handle
  - aborted `executeRaw()` query drain before close
  - direct `exec` drain before close
  - transaction drain before close

If the underlying WASM operation never settles, this does not invent a new cancellation primitive; the existing CLI hard-deadline path remains the backstop.

## Verification

- `bun test --timeout 30000 test/pglite-engine-disconnect.serial.test.ts`
- `bun test --timeout 30000 test/pglite-engine.test.ts`
- `bun run typecheck`
- `git diff --check`

Fixes #2348

### DIFF
--- a/src/core/pglite-engine.ts
+++ b/src/core/pglite-engine.ts
@@ -225,6 +225,7 @@ export class PGLiteEngine implements BrainEngine {
   readonly kind = 'pglite' as const;
   private _db: PGLiteDB | null = null;
   private _lock: LockHandle | null = null;
+  private _inFlight = new Set<Promise<unknown>>();
   // #2034: captured at connect() so reconnect() can restore the same data dir
   // after a drop, matching PostgresEngine's _savedConfig contract.
   private _savedConfig: EngineConfig | null = null;
@@ -271,13 +272,14 @@ export class PGLiteEngine implements BrainEngine {
     // why the CLI's exit paths read gbrain's own verdict
     // (cli-force-exit.ts currentExitCode), never ambient process.exitCode.
     try {
-      this._db = await preservingProcessExitCode(() =>
+      const rawDb = await preservingProcessExitCode(() =>
         PGlite.create({
           dataDir,
           loadDataDir,
           extensions: { vector, pg_trgm },
         }),
       );
+      this._db = this.trackDb(rawDb);
     } catch (err) {
       // v0.13.1: any PGLite.create() failure becomes actionable. v0.41.8.0
       // (#1340): the previous error hint hardcoded the macOS 26.3 link, but
@@ -317,6 +319,7 @@ export class PGLiteEngine implements BrainEngine {
     this._lock = null;
     try {
       if (db) {
+        await this.drainInFlightOperations();
         // Deliberately NOT wrapped in preservingProcessExitCode: close's
         // status write (0) is long-standing baseline behavior that test-runner
         // processes depend on (wrapping it flipped bun test's own exit code —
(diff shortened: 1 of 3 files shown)


time-attack:
Thank you for the careful write-up and the lifecycle test coverage — but I'm closing this one: #2348 was already root-caused and fixed on master by #2400 (`058f448b`, never steal a live data-dir lock plus the corrupted-store recovery hint), and the issue has been closed since 16 July. What's left in this PR is hardening for a close race that wasn't reproduced, and the cost is high for that: a Proxy sits on every `query`/`sql`/`exec`/`transaction` call (already needing a getter special case to keep `ready`/`closed` working), and `drainInFlightOperations`' unbounded `while` loop can hang shutdown indefinitely if work keeps arriving. If you can reproduce a concrete close-during-query failure on current master, an issue with that repro would be the right next step.

Issue #2374, opened 2026-06-23
enrich fails on native Postgres: recordCompleted double-encodes op_checkpoints.completed_keys (jsonb string, not array) → CHECK violation


ralphvspanje-hub:
### Summary
On self-hosted **native PostgreSQL** (via `postgres.js`, direct pool), `gbrain enrich` — and any op routed through `recordCompleted` — fails with:

```
new row for relation "op_checkpoints" violates check constraint "op_checkpoints_completed_keys_array"
```

The enrich makes its LLM call, then the checkpoint write is rejected and the op aborts → `0 enriched`. The nightly `cycle.enrich_thin` phase hits the same wall.

### Root cause
`src/core/op-checkpoint.ts` → `recordCompleted` writes `completed_keys` as:

```
VALUES ($1, $2, $3::jsonb, now())   -- param: JSON.stringify(sorted)
```

On `postgres.js` against native Postgres, a JS string bound to a `$3::jsonb` param is **double-encoded** into a jsonb *string* (scalar), which fails `CHECK (jsonb_typeof(completed_keys) = 'array')`. The in-code comment asserts this form is "correct (not the double-encode trap)" — which holds on **PGLite**, but not on native Postgres/postgres.js (so PGLite-based testing doesn't catch it). The sibling `appendCompleted` already sidesteps this by binding `text[]`.

### Fix (verified locally)
Bind a `text[]` and convert with `to_jsonb`, mirroring `appendCompleted`:

```
VALUES ($1, $2, to_jsonb($3::text[]), now())   -- param: sorted   (string[], not JSON.stringify)
```

After this change, `enrich` writes its checkpoint and completes normally.

### Environment
- gbrain 0.42.52.0
- PostgreSQL 18 + pgvector, self-hosted (native), `postgres.js` direct pool
- Multi-agent (two `gbrain serve` processes share one brain → Postgres required; PGLite not viable per #2058)

### Secondary (possibly separate)
Capped runs (`--max-usd`) with **dateless** Anthropic ids (`claude-sonnet-4-6`, `claude-haiku-4-5`) throw `BudgetExhausted(reason: 'no_pricing')` because `ANTHROPIC_PRICING` is keyed by **dated** snapshots and the dateless id isn't resolved for pricing. Dated ids (`claude-haiku-4-5-20251001`) work. Suggest adding dateless aliases to the pricing map (or resolving alias → dated before the pricing lookup) so `--max-usd` works with the common dateless ids.

[...]

PR #2398, opened 2026-06-24
v0.42.54.0 fix(reliability): silent-failure wave — upload-raw persistence, PGLite setup hint, honest doctor, config merge guard (#2297 #2301 #2330 #2251)

garrytan:
## What this ships

Four "silent failure" bugs — each returned a green result while doing the wrong thing. Part of the post-#2375 cleanup (the jsonb dupe queue was bulk-closed separately; see below).

### Fixed
- **#2297 — `files upload-raw` silently no-op'd on small text/PDF.** The small-file branch printed `success` and returned without copying into the brain repo or inserting a `files` row. Both branches now persist via one shared writer. `storage_path` (globally UNIQUE) is source-namespaced so the same page slug in two sources can't clobber each other; a new `--source` flag strictly disambiguates a multi-source slug (scoped miss → null, no unscoped fallback).
- **#2301 — PGLite embedding setup dead-ended.** `config set embedding_model` (file-plane field) now points at `gbrain reinit-pglite --embedding-model X --embedding-dimensions N` instead of a manual recipe that omitted the required dimension flag.
- **#2330 — doctor counted un-scanned surfaces as healthy.** An un-runnable expected surface now emits an explicit failing check instead of a vacuous 100; `getHealth`/`getStats` agree on `page_count` (both exclude soft-deleted), and `pages_with_timeline`/`most_connected` exclude deleted so coverage shares one denominator.
- **#2251 — corrupted `sources.config` aborted the config merge.** The shape-repair merge called `jsonb_each()` on non-object array elements and threw, failing the whole UPDATE. Now skips non-object elements; PGLite brought to parity with the same self-healing merge.

## Tests
New `test/silent-failure-wave.test.ts` (8 tests) covers all four, including multi-source `storage_path` isolation and the strict `--source` behavior. Touched-area suites green (`doctor`, `files`, `list-all-sources`, `source-config-redact`, `doctor-brain-checks-score`). Typecheck clean. jsonb guard clean.

## Review
Plan-stage codex outside-voice + two diff-stage adversarial passes. Three rounds of P2 findings all resolved: source-collision on `storage_path`, half-exclusion of deleted pages in health, and the non-strict `--source` fallback.

## Queue hygiene (done out-of-band, no code)
Closed 16 resolved-by-#2375 jsonb dupes (PRs #2288 #2309 #2328 #2333 #2338 #2350 #2351 #2355 #2360; issues #2280 #2305 #2332 #2307 #2374 #2263 #2324) plus #1986, each with a superseded/fixed comment + contributor credit.

## Deferred to a separate incident PR
#2296 (process leak) and #2348 (PGLite 58P01 after `dream`) are not root-caused and ship separately after a reproduction spike — a guessed recovery fix for a corruption path should not ride a routine wave.


time-attack:
Merging — the #2297 lane is confirmed still live on master: the `upload-raw` small-file branch prints `success: true` and returns before copying into the brain repo or inserting the `files` row, which is silent data loss on the path users trust most, and the source-namespaced `storage_path` correctly closes the cross-source clobber under the global UNIQUE constraint. Two notes for the rebase: the branch is 12 versions behind so VERSION/package.json/CHANGELOG will conflict, and the #2251 config-merge lane is already resolved upstream (master's `source-config-sql.ts` guards non-object array elements), so that hunk can be dropped.

PR #2401, opened 2026-06-24
WS-954: add import opt-out for gitignored files

tangyuanjc:
## Summary

Ref: WS-954

`gbrain import` currently switches to `git ls-files --cached --others --exclude-standard` inside git worktrees, which is correct for normal repo imports but can silently drop generated import feeds that intentionally live under `.gitignore`d directories.

This adds explicit opt-outs for that import-only case:

- `gbrain import <dir> --all-files`
- `gbrain import <dir> --ignore-gitignore`
- `GBRAIN_IMPORT_IGNORE_GITIGNORE=1 gbrain import <dir>`

The opt-out is wired through `runImport` into `collectSyncableFiles(..., { ignoreGitignore: true })`; direct `collectSyncableFiles` callers, especially sync, keep the default gitignore policy unless they opt in programmatically.

## Verification

- `bun test test/sync-walker-symlink.test.ts`
- `bun test test/import-source-id.test.ts`
- `bun run typecheck`
- Smoke: disposable PGLite brain + git repo where `ignored/hidden.md` is ignored:
  - default import found 1 markdown file
  - `--all-files` import found 2 markdown files and imported `ignored/hidden`
  - `GBRAIN_IMPORT_IGNORE_GITIGNORE=1` import found 2 markdown files

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -2260,7 +2260,8 @@ ADMIN
   autopilot [--repo] [--interval N]  Self-maintaining brain daemon
   config [show|get|set] <key> [val]  Brain config
   storage status [--repo <path>]     Storage tier status and health
-        [--json]                     (git-tracked vs supabase-only)
+  storage vacuum [--json]            VACUUM (ANALYZE) + CHECKPOINT
+  storage backup [--json]            PGLite data-dir backup with retention caps
   serve                              MCP server (stdio)
   serve --http [--port N]            HTTP MCP server with OAuth 2.1
     --token-ttl N                    Access token TTL in seconds (default: 3600)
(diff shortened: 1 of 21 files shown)


time-attack:
Thanks for putting this together, but I'm closing it. The title and description cover a ~36-line change in `import.ts`, while the diff also ships a 313-line storage vacuum in `storage.ts`, a brand-new `local-sentence-transformers` provider recipe with gateway wiring, and edits across `capture.ts`, `schema.ts`, `embedding.ts` and both engines — that isn't reviewable as one unit, and the new third-party provider recipe is outside what I take in any case. On the nominal feature: the gitignore behavior in git worktrees is intentional, and three spellings of the same opt-out (two flags plus an env var) is more surface than the case warrants.

PR #2403, opened 2026-06-25
fix: DashScope embed batch-size cap + takes model from config

yiliwulu1999yunnan-boop:
## Summary

Two fixes for non-Anthropic provider setups:

### 1. DashScope embedding batch-size cap
DashScope's OpenAI-compatible `/embeddings` endpoint rejects requests with more than 10 inputs (`"batch size is invalid, it should not be larger than 10"`). This caused chunk-heavy pages (e.g. long transcripts) to fail embedding entirely.

**Fix**: Add `max_batch_inputs` to `EmbeddingTouchpoint` and enforce in `gateway.ts` — after splitting by token budget, further split each batch by input COUNT when the provider declares a cap. Set `max_batch_inputs=10` for the DashScope recipe.

### 2. Takes phases use config.chat_model
`propose_takes`, `grade_takes`, and `calibration_profile` were hardcoded to `claude-sonnet-4-6`, silently failing when no ANTHROPIC key was available.

**Fix**: Read model from `calibrationConfig.chat_model` and pass it through to the three takes phases. Users without Anthropic keys can now run takes on their configured chat model (e.g. deepseek-v4-pro). The hardcoded default in each phase file still applies as fallback when no model is explicitly passed.

### Bonus
- Change llama-server-reranker path from `/rerank` to `/reranks` (plural) for DashScope compatibility. llama-server accepts both forms.

### DIFF
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -1400,9 +1400,20 @@ export async function embed(texts: string[], opts?: EmbedOpts): Promise<Float32A
 
   // Pre-split is gated on max_batch_tokens. Recipes without it (e.g. OpenAI)
   // ride the fast path: one embedMany call, no recursion safety net.
-  const batches = maxBatchTokens
+  const tokenBatched = maxBatchTokens
     ? splitByTokenBudget(truncated, Math.floor(maxBatchTokens * effectiveSafetyFactor(recipe)), charsPerToken)
     : [truncated];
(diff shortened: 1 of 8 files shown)


time-attack:
DashScope support was reviewed and declined as a direction for now — closing the DashScope PR set together. (The takes-model config half of this was already covered by #2805's territory.) Thank you for the work.

PR #2405, opened 2026-06-25
feat(embedding): add max_batch_items to cap batch by item count

zijun-robotics:
DashScope text-embedding-v3/v4 limit each API call to 10 input items, but gbrain's token-based batching can pack 20+ small chunks into one request, causing HTTP 400.

Changes:
- Add max_batch_items to EmbeddingTouchpoint in types.ts
- Add provider_batch_items to AIGatewayConfig for runtime override
- Update build-gateway-config.ts to pass through provider_batch_items
- Update splitByTokenBudget to accept maxBatchItems; cap by tokens or items, whichever is hit first
- Set max_batch_items: 10 in DashScope recipe


time-attack:
DashScope support was reviewed and declined for now — closing the set together. The generic max-batch-items idea from this PR is being adopted separately with credit; thank you.

PR #2406, opened 2026-06-25
feat(links): resolve [[wikilink]] frontmatter values via global_basename

spiky02plateau:
## Problem

Bare-title `[[wikilink]]` values in **frontmatter link fields** never resolve to edges. Example from a PARA/Obsidian vault:

```yaml
sources:
  - "[[2025-12-25_mentor-extraction]]"
```

Walking `SlugResolver.resolve()` for that value:

1. **slug-direct getPage** — skipped: the value has no `/`.
2. **dirHint + slugify exact** — misses: `sources` maps to `dirHint: ['source','media']`, folders that don't exist in this brain (the real page is `…/raw/mentor-transcripts/2025-12-25_mentor-extraction`).
3. **dir-scoped fuzzy** — misses: scoped to the same non-existent dirHints, and the search string still carries the `[[ ]]`.

So it returns `null` and the edge is dropped. The body bare-wikilink path got basename-index resolution in #972 (`resolveBasenameMatches`, gated by `link_resolution.global_basename`), but the **frontmatter path was never wired to it**. On a real vault this silently drops the bulk of `sources:` / `related:` provenance edges (in my brain: ~1,400 `discussed_in` + ~1,600 `related_to` edges were missing).

## Fix

Extend `global_basename` resolution to frontmatter fields:

- `extractFrontmatterLinks` takes a `globalBasename` flag (threaded from `extractPageLinks`, where it's already read from config).
- On a `resolve()` miss, **unwrap** the `[[ ]]` and fall back to `resolver.resolveBasenameMatches` — the same index the body path uses.
- **Unique-match-only**: if the basename maps to >1 page (archive dupes, generic hubs like `_index`), it stays unresolved rather than guess a wrong edge.

Purely additive — frontmatter values that already resolved are unchanged.

## Scope / limitations

- Covers the **db-source extract** and **live `put_page`** paths (both use the real `makeResolver`, which implements `resolveBasenameMatches`).
- The `--source fs` extract uses an inline resolver without a basename index, so it **gracefully no-ops** there via the `typeof resolver.resolveBasenameMatches === 'function'` guard. Wiring an fs-side basename index could be a follow-up.

## Testing

3 new cases in `test/link-extraction.test.ts`:
- resolves a bare-title `[[wikilink]]` frontmatter value via the basename fallback when the flag is on;
- stays unresolved when the basename is ambiguous (>1 match);
- is gated off when `globalBasename` is false.

Full `link-extraction` suite green (130 pass, 0 fail).

Related: follow-up to #972 (global_basename for body wikilinks); complements #1983 (which handles bracket-unwrap + exact slug-path values, but not bare-title basename resolution).

### DIFF
--- a/src/core/link-extraction.ts
+++ b/src/core/link-extraction.ts
@@ -552,7 +552,7 @@ export async function extractPageLinks(
   // path needed `resolveBasenameMatches` on the real resolver.
   let fmUnresolved: UnresolvedFrontmatterRef[] = [];
   if (!opts.skipFrontmatter) {
-    const fm = await extractFrontmatterLinks(slug, pageType, frontmatter, resolver);
+    const fm = await extractFrontmatterLinks(slug, pageType, frontmatter, resolver, opts.globalBasename);
     candidates.push(...fm.candidates);
     fmUnresolved = fm.unresolved;
   }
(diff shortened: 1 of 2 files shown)


spiky02plateau:
This pairs with #2434 — together they make frontmatter link edges first-class on every write path: this PR resolves bracketed frontmatter values (`put_page` + manual extract), and #2434 keeps those edges fresh on the incremental cycle/sync path (today the cycle re-extracts body links only, so externally-edited frontmatter drifts stale). Both are default-off.

FWIW I ran the full local gate on both branches — `tsc --noEmit`, `bun run verify` (all checks), and the link/extract/cycle suites — all green. CI on the PRs looks parked in `action_required` (fork-PR workflow approval) rather than failing, if that's the holdup. Happy to rebase or adjust the shape of either if useful.

PR #2409, opened 2026-06-25
v0.44.2.0 fix(test): isolate GBRAIN_SOURCE_BOOST in resolveBoostMap unset test

JiraiyaETH:
## Summary

Test-hygiene fix, **no runtime change**. `test/sql-ranking.test.ts`'s `resolveBoostMap > "returns defaults when env is unset"` failed locally whenever `GBRAIN_SOURCE_BOOST` was exported in the developer's shell.

`resolveBoostMap(envValue = process.env.GBRAIN_SOURCE_BOOST)` reads the env var via its default parameter, so calling `resolveBoostMap(undefined)` falls through to the ambient env. The test asserted the *unset* contract without ensuring the env was actually unset — so it passed on CI (var absent) but failed for any developer with `GBRAIN_SOURCE_BOOST` set.

The test now saves, deletes, and restores `process.env.GBRAIN_SOURCE_BOOST` around the assertion, so it deterministically exercises the unset case in any environment.

## Test plan
- [x] `bun test test/sql-ranking.test.ts` with `GBRAIN_SOURCE_BOOST` exported — 52 pass / 0 fail (was 51 pass / 1 fail)

### DIFF
--- a/test/sql-ranking.test.ts
+++ b/test/sql-ranking.test.ts
@@ -209,7 +209,15 @@ describe('parseHardExcludesEnv', () => {
 
 describe('resolveBoostMap', () => {
   test('returns defaults when env is unset', () => {
-    expect(resolveBoostMap(undefined)).toEqual(DEFAULT_SOURCE_BOOSTS);
+    // Isolate from an ambient GBRAIN_SOURCE_BOOST in the dev's shell (#2409):
+    // resolveBoostMap(undefined) falls through to process.env via its default param.
+    const saved = process.env.GBRAIN_SOURCE_BOOST;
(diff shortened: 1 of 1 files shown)


time-attack:
Merging this. Confirmed the defect is still live on master: `resolveBoostMap(envValue = process.env.GBRAIN_SOURCE_BOOST)` means `resolveBoostMap(undefined)` reads the ambient env, so the 'returns defaults when env is unset' test was only truthful on machines that happen not to export `GBRAIN_SOURCE_BOOST`. The save/delete/restore in a `finally` makes the unset contract deterministic everywhere with zero runtime change. Thanks for the precise write-up of the default-parameter fallthrough.

Issue #2411, opened 2026-06-25
gbrain takes propose is documented as the promotion path for take_proposals but the CLI command is missing


615Works:
## Summary

`propose_takes` writes candidate rows to `take_proposals`, but the documented CLI path for promoting those rows into canonical takes does not exist. The source comment in `src/core/cycle/propose-takes.ts` says the operator accepts or rejects proposals with `gbrain takes propose --accept N`. The dispatcher in `src/commands/takes.ts` has cases for `search`, `add`, `update`, `supersede`, `resolve`, `scorecard`, `calibration`, `revisit`, and `extract`, but no `propose` case.

That leaves a producer phase that can spend model budget and populate `take_proposals`, while the documented consumer path is absent.

## Reproduction steps

1. Install gbrain 0.42.52.0 or inspect current `master`.
2. Read `src/core/cycle/propose-takes.ts`.
3. Note the documented operator path: `gbrain takes propose --accept N`.
4. Read `src/commands/takes.ts`.
5. Observe that `runTakes` has no `propose` case in its dispatcher.

## Expected behavior

Either `gbrain takes propose` should exist and support accepting or rejecting proposals, or the `propose_takes` documentation should point users to the actual supported promotion path.

## Actual behavior

The documented command is missing. A user can run the proposal producer but cannot use the documented consumer path to promote those proposals into canonical takes.

## Proposed fix / approach

Implement the documented `gbrain takes propose` subcommand, or update the `propose_takes` documentation to point at the real promotion path if `takes extract --from-pages` replaced it.

## Context

Observed on gbrain 0.42.52.0 and source-inspected on 2026-06-25. Relevant files:

1. `src/core/cycle/propose-takes.ts`
2. `src/commands/takes.ts`

Note: #2269 implies a TTY interactive `gbrain takes propose` CLI exists, and #1467 includes a review CLI as part of a broader feature request. Our source read against 0.42.52.0 did not find a `propose` case in `runTakes`. If the command was added after our inspection window, this issue is superseded by those. If the claim in #2269 is aspirational rather than describing an implemented CLI, this bug report and those issues are complementary.

time-attack:
## Severity corrected: p3 → **P1**

This issue was labelled p3 ("cosmetic, docs-only, or a rare configuration"). That was wrong. It was re-assessed by reproducing it against current master.

[...]

PR #2418, opened 2026-06-25
feat: add review-first take proposals

Mr-B-1:
## Summary
- Adds native review-first take proposal list/accept/reject flow.
- Wires MCP/operation allow-list support without auto-accepting proposed takes.

## Verification
- bun run typecheck
- bun test test/take-proposals.test.ts test/mcp-tool-defs.test.ts test/operations-allow-list.test.ts test/e2e/takes-postgres.test.ts
- git diff --check

### DIFF
--- a/src/commands/takes.ts
+++ b/src/commands/takes.ts
@@ -28,6 +28,11 @@ import {
   type ParsedTake,
 } from '../core/takes-fence.ts';
 import { withPageLock } from '../core/page-lock.ts';
+import {
+  acceptTakeProposal,
+  listTakeProposals,
+  rejectTakeProposal,
+} from '../core/take-proposals.ts';
 
 // --- Helpers ---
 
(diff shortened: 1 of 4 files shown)


devty:
Read this closely since it's the promotion path a lot depends on (#2411/#2269) — the core `acceptTakeProposal` logic is tight and the review-first principle is enforced throughout. Two blocking items, a few non-blocking notes, and a forward-compat note for a grading layer I'm planning on top.

## Blocking

**1. `pg_advisory_xact_lock` will throw on PGLite, rolling back the accept transaction.**
`src/core/take-proposals.ts` — the inner transaction calls `SELECT pg_advisory_xact_lock($1::bigint)` bare. PGLite doesn't implement it and throws, unwinding `engine.transaction`, so `takes propose accept` hard-fails on any PGLite install. The codebase already wraps advisory locks defensively (`operations.ts:1148-1152`); same fix here:
```ts
try {
  await tx.executeRaw('SELECT pg_advisory_xact_lock($1::bigint)', [pageId]);
} catch { /* PGLite single-process; no cross-process race to guard */ }
```

**2. An `accepted` proposal with `promoted_row_num = null` falls through to a confusing throw.**
The idempotency fast-path is `status === 'accepted' && promoted_row_num !== null`. If a prior accept partially committed (UPDATE landed, RETURNING didn't), the row is `accepted` with a null row_num and the next call hits `if (status !== 'pending') throw` — an opaque error instead of surfacing the partial state. Either early-return on `status === 'accepted'` (with a warning), or add a NOT NULL constraint so the half-state can't exist.

## Non-blocking

**3.** `rejectTakeProposal` accepts a `reason`, returns it, but never persists it (no `rejection_reason` column). The op description is transparent about this, which is good — but a schema migration to persist it (or a `migrate.ts` TODO) would keep it from reading like silent data loss to a future contributor.

**4.** `rejectTakeProposal` only blocks on `accepted`, so it will silently stamp a `superseded` proposal as `rejected`. If superseded→reject is valid re-triage, document it; otherwise guard it. No test covers the `superseded` path.

**5.** The markdown `existsSync(path)` check runs before `engine.transaction` (minor TOCTOU). Fine for current usage; worth noting for a hardened remote path.

## Nits

- CLI `since=(unset: no real claim-date)` is verbose for the terminal — `since=<none>` reads cleaner.
- A test asserting `takes_propose_list.scope === 'read'` and that the write ops require `proposal_id` would explicitly close the "wrong scope exposed over MCP" scenario.

## What's done well

`acceptTakeProposal` is careful: re-fetches the proposal *for update* inside the transaction, dedups against active takes by lowercased claim, writes markdown before the DB mirror, and rolls back the filesystem write on exception. The `since_date` fallback-suppression (dropping `source='fallback'` dates) is a good data-quality call. MCP ops are correctly `read`/`write`-scoped, the remote-trust model follows the established `ctx.sourceId` pattern, and `FakeProposalEngine` is thorough enough to catch real promotion bugs without a live DB.

## Forward-compat (not a blocker)

For a two-track grader on top of this, a promoted take needs `kind`, `weight`, `since_date`, and a path back to the proposal — all four survive (`kind`/`holder`/`weight` carried verbatim, `since_date` resolved with fallback-suppression, `source` stamped `proposal:${id}`). One thing to surface for the follow-up rather than block here: `kind` is carried from the proposal with no override at accept time. Since all current proposals are `kind='take'`, minting a gradeable `kind='bet'` will need either a kind override at accept (`--as bet`) or a post-accept update — neither exists today. That's the natural seam where "elect this opinion into a bet" would live.

sene1337:
Thanks for this PR. I reviewed it together with devty's comment and agree with the review-first shape: `propose_takes` → `take_proposals` → explicit accept/reject is the right quality gate, and the core `acceptTakeProposal` flow looks sound.

For Brad's brain we are taking the automatic facts→takes consolidate lane first, not auto-accepting proposals. I opened the missing take-embedding writer as #2483 because consolidate currently needs `gbrain embed --stale` to pick up newly promoted takes immediately.

I still want this proposal-accept path to land and become the curated lane alongside consolidate. The two items devty called out are the adoption blockers from my side too:

1. `pg_advisory_xact_lock` needs the defensive PGLite wrapper pattern from `operations.ts:1148-1152`; bare advisory lock calls make accept fail on the default PGLite install.
2. `accepted` with `promoted_row_num = null` should not fall through to the opaque non-pending throw. Either early-return with a warning on `status === 'accepted'`, or make the half-state impossible with a NOT NULL/constraint shape.

Once those are fixed, this gives the stack the right two-lane model:

- automatic lane: facts→takes consolidate for low-friction cold memory
- curated lane: proposals→review-first accept for human/agent-vetted takes

I do not want unreviewed proposal auto-accept; the review gate is the feature here.

time-attack:
Closing this one — it's a draft adding roughly 790 lines of brand-new take-proposal subsystem (new core module, new operations, new CLI subcommands, MCP wiring), which is exactly the kind of feature surface we're not taking right now. The code itself is clean and the tests are there, but 'well-built' isn't the bar for a new subsystem; the bar is a broken core promise, and takes already function without a proposal queue. Thanks for the effort you put into it.

PR #2420, opened 2026-06-25
feat: add native since-date repair

Mr-B-1:
## Summary
- Adds native takes repair-since-date command.
- Repairs missing take since_date from non-fallback page effective_date only, dry-run by default.

## Verification
- bun run typecheck
- bun test test/takes-since-date-repair.test.ts
- git diff --check

### DIFF
--- a/src/commands/takes.ts
+++ b/src/commands/takes.ts
@@ -28,6 +28,7 @@ import {
   type ParsedTake,
 } from '../core/takes-fence.ts';
 import { withPageLock } from '../core/page-lock.ts';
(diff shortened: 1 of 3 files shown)


time-attack:
Closing this one — it's a draft adding a new `takes repair-since-date` command and a supporting core module, which is new CLI surface for a backfill convenience rather than a repair of something broken. The dry-run-by-default design is the right instinct, but we're holding the line on added command surface. Thanks for taking the care to scope it to non-fallback effective dates only.

PR #2421, opened 2026-06-25
fix: ground take grading in hybrid evidence

Mr-B-1:
## Summary
- Grounds take grading with hybrid search evidence instead of placeholder evidence.
- Marks sparse or uncited evidence as unresolvable before paid judge calls.

## Verification
- bun run typecheck
- bun test test/grade-takes.test.ts test/grade-takes-ensemble.test.ts
- git diff --check

### DIFF
--- a/src/core/cycle.ts
+++ b/src/core/cycle.ts
@@ -2040,7 +2040,31 @@ export async function runCycle(
           checkAborted(opts.signal);
           progress.start('cycle.grade_takes');
           const { runPhaseGradeTakes } = await import('./cycle/grade-takes.ts');
-          const { result, duration_ms } = await timePhase(() => runPhaseGradeTakes(calibrationCtx, {}) as Promise<PhaseResult>);
+          const configValue = async (key: string): Promise<string | null> => {
+            const value = await engine.getConfig(key);
+            return value === undefined ? null : value;
+          };
+          const configBool = async (key: string, fallback: boolean): Promise<boolean> => {
+            const raw = await configValue(key);
+            if (raw === null) return fallback;
+            const normalized = raw.trim().toLowerCase();
+            if (['true', '1', 'yes', 'on'].includes(normalized)) return true;
+            if (['false', '0', 'no', 'off'].includes(normalized)) return false;
+            return fallback;
+          };
+          const configNumber = async (key: string, fallback: number): Promise<number> => {
+            const raw = await configValue(key);
+            if (raw === null) return fallback;
+            const n = Number(raw);
+            return Number.isFinite(n) ? n : fallback;
+          };
+          const gradeTakesConfig = {
+            autoResolve: await configBool('cycle.grade_takes.auto_resolve.enabled', false),
+            autoResolveThreshold: await configNumber('cycle.grade_takes.auto_resolve.threshold', 0.95),
+            minAgeMonths: await configNumber('cycle.grade_takes.min_age_months', 0),
+            takeLimit: await configNumber('cycle.grade_takes.take_limit', 50),
+          };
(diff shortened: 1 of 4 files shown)


devty:
Took a close read of this with an eye toward building on it — really glad to see the placeholder retriever finally getting grounded. Two blocking items (one is a compile error), a few non-blocking notes, then a forward-compat note since I'm planning a two-track grading layer on top of this.

## Blocking

**1. `defaultEvidenceRetriever` signature not updated to the widened `EvidenceRetrieverFn`.**
`src/core/cycle/grade-takes.ts:299` — the diff widens `EvidenceRetrieverFn` to `(take, scope, engine) => Promise<string>`, but `defaultEvidenceRetriever` still declares two params `(take, _scope)`. It's exported and used as an `evidenceRetriever` in `test/grade-takes.test.ts`, so this is a TS assignability error. Fix: add the unused third param `_engine: BrainEngine`.

**2. `minAgeMonths` default is incoherent across the two layers.**
`src/core/cycle.ts` passes `minAgeMonths: configNumber('cycle.grade_takes.min_age_months', 0)`, while the phase falls back to `?? 6` when the option is absent. Once cycle.ts supplies the value, the 6-month safety net silently becomes 0 — every brand-new take is immediately eligible on an unconfigured brain, inverting the age-gate intent. Pick one number and make both layers agree (config default `6` to match the phase, or document that `0` = "no minimum" intentionally).

## Non-blocking

**3.** `hybridEvidenceRetriever`'s `since` filter rides the engine's `afterDate`/`effective_date` SQL filter, so on a brain with sparse `effective_date` coverage pages get excluded before `formatGradeEvidenceFromResults` ever sees them — almost always returning "unavailable" with no operator-facing signal explaining why. A line in the function doc calling out that dependency would save a debugging session.

**4.** `evidenceHasCitedSnippets` (`:272`) keys off `^Source:`/`^Snippet:` with `/m`. Correct for the current marker, but if a judge ever echoes those tokens in its reasoning the guard could pass spuriously — worth a forward-compat note.

## Nits

- `test/grade-takes.test.ts` constructs the `UNAVAILABLE_EVIDENCE_MARKER` string inline rather than importing the constant — if the marker text changes the test silently passes against the old string. Import the constant (or expose via `__testing`).
- Consider naming the `insertGradeCache` helper's `applied` param `wasApplied` to avoid colliding with the outer `shouldApply` in readers' heads.

## What's done well

The pre-judge guard fires *after* the idempotency lookup and *before* any gateway call, so a retry never pays twice — exactly the right placement. `insertGradeCache` genuinely de-duplicates the SQL. `normalizeDateOnly` handling `YYYY-MM → YYYY-MM-15` is a thoughtful edge case, and the new `evidence_unavailable` counter gives operators visibility into how often the guard fires.

## Forward-compat (not a blocker)

I'm planning a follow-up two-track layer: per-kind age gates (`min_age_months.bet`/`.take`), grading `kind='take'` opinions as a separate "opinion track record" (hit-rate, not Brier), alongside the bet Brier track. This PR's design makes that clean — `hybridEvidenceRetriever` is a proper named export satisfying the `EvidenceRetrieverFn` seam, and `{KIND}` is already threaded into the judge prompt so the scorecard can read it from existing cache rows with no schema change. The one thing worth fixing here first is finding #2 (the `minAgeMonths` incoherence), since the follow-up splits that single scalar into two keys — better to settle the number once than migrate the confusion into two config paths.

time-attack:
Closing this. Master's grade-takes phase ships its placeholder evidence retriever deliberately — the header comment says so — so this is completing a feature rather than fixing a break, and it additionally introduces four new `cycle.grade_takes.*` config keys with an inline parsing helper block inside `cycle.ts`. Guarding against spending judge budget on placeholder evidence is a genuinely good instinct, but the whole change is draft-stage feature work on a surface we're not expanding. Thanks for digging into the grading path.

PR #2422, opened 2026-06-25
feat: attach concepts to extracted atoms

Mr-B-1:
## Summary
- Adds synthesize_concepts to the base schema phase list.
- Normalizes and writes concept labels produced during atom extraction.

## Verification
- bun run typecheck
- bun test test/cycle/extract-atoms-synthesize-concepts.test.ts test/schema-pack-manifest-v041_2.test.ts test/cycle-pack-gating.test.ts
- git diff --check

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -44,7 +44,7 @@ for (const op of operations) {
 }
 
 // CLI-only commands that bypass the operation layer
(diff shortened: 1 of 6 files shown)


time-attack:
Closing this one — it's a draft feature adding a `synthesize_concepts` phase to the shipped base schema pack plus a new concept normalization and write path through atom extraction and a backfill command. Changing the bundled base pack's phase list affects every install, and that's a big surface bet for something atoms work fine without today. Thanks for the tests around the synthesize path.

PR #2423, opened 2026-06-25
fix: classify enrich-thin blocked states

Mr-B-1:
## Summary
- Adds explicit enrich-thin result status classification.
- Distinguishes disabled, no sources, no candidates, evidence gaps, zero cap, blocked budget, and true exhaustion.

## Verification
- bun run typecheck
- bun test test/enrich-cycle-phase.test.ts test/e2e/dream-cycle-phase-order-pglite.test.ts
- git diff --check

### DIFF
--- a/src/core/cycle/enrich-thin.ts
+++ b/src/core/cycle/enrich-thin.ts
@@ -69,6 +69,55 @@ interface ResolvedConfig {
   model?: string;
 }
 
+export type EnrichThinResultStatus =
+  | 'completed'
+  | 'disabled'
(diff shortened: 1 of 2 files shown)


time-attack:
Closing this. It's a draft that adds an eleven-value status taxonomy and classifier to distinguish why enrich-thin did nothing — useful reporting granularity, but nothing on master is functionally broken without it, so it's observability surface expansion rather than a fix. Eleven distinct states is also a lot of new vocabulary to keep truthful over time. Thanks for the clear enumeration of the cases you found.

PR #2425, opened 2026-06-25
fix(search): honor source recency policy in queries

briancl2:
## Summary

- load a concrete source repo's `gbrain.yml recency:` map in the `query` operation and thread it into hybrid recency scoring
- include a stable recency-policy digest in `knobsHash` so query-cache rows do not survive ranking-policy changes
- fail closed on malformed YAML frontmatter during markdown import instead of storing raw frontmatter as searchable body text

## Validation

- `bun test test/query-recency-config.test.ts test/source-id-tx-regression.test.ts test/search.test.ts test/recency-decay.test.ts test/search-mode.test.ts test/search/knobs-hash-reranker.test.ts test/cross-modal-phase1.test.ts test/search-alias-resolved-boost.test.ts test/build-llms.test.ts`
- `bun run typecheck`
- `scripts/check-key-files-current-state.sh`
- `bun run build:llms`
- `bun run verify`

## Local note

`bun run ci:local:diff` was attempted but could not complete in this environment because `docker` is not installed on PATH; the non-Docker verify suite passed after repairing the local admin Rollup optional dependency install.


time-attack:
Thanks for the thorough work and the validation list, but I'm closing this one. It's titled as a fix yet lands four independent changes at once — a new per-source `gbrain.yml recency:` config surface, a knobs_hash version bump, a fail-closed frontmatter change in `import-file.ts`, and an unrelated new `cycle` CLI alias — which makes it unreviewable as a single unit, and the recency-policy piece is new configuration surface rather than a repair of broken behavior. The KEY_FILES additions also append duplicate `extension —` entries instead of updating the existing entries in place, which is the append-only pattern that doc is explicitly guarded against.

PR #2427, opened 2026-06-25
feat(facts): date conversation facts at claim time + owner-trusted local fact reads

jeanpierre121:
## What

Two fixes that make `find_trajectory` (and fact recall) actually usable for a
single-owner brain whose facts come from conversation extraction.

### 1. Date facts at claim time, not extraction time

`extract-conversation-facts` built fact rows without a `valid_from`, so
`insertFacts` stamped `now()`. `find_trajectory` orders entirely by `valid_from`,
so every entity's timeline collapsed onto the day it was extracted instead of
when the claim was actually made. Each segment already carries its own
timestamps; we now stamp each fact with the segment date (`segmentValidFrom`),
guarding the `1970` epoch fallback that `parseConversation` emits when a page
has no date at all (then `valid_from` stays unset and `insertFacts` keeps its
`now()` default).

### 2. Owner-trusted local fact reads

Fact rows are `private` | `world`. Remote callers are filtered to `world` only —
correct for a published or HTTP-served brain. But the stdio MCP server is an
unauthenticated **local pipe**: on a single-owner machine its caller is the
owner, yet it still defaults `remote: true`, so the owner's own agent gets an
empty `find_trajectory` over MCP even though the facts exist.

This adds `trustedFactReads` — a narrow, **read-only** trust elevation
decoupled from `remote`, so every other remote protection (file confinement,
source isolation, fence stripping, takes-holder scoping) stays fully in force.
The stdio server sets it only when the owner opts in via
`facts.trust_local_reads` (default off); the HTTP transport never sets it, so a
served brain stays `world`-only regardless.

## How

- `src/core/facts/reader-trust.ts` (new): `factsWorldOnly(ctx)` /
  `readableFactVisibilities(ctx)` — the single decision point for fact-read
  visibility. `world`-only iff `remote === true && !trustedFactReads`.
- All four fact-read visibility sites now route through it: `findTrajectory`
  (both engines, kept in lockstep), the recall meta-hook, and the
  `listFactsByEntity` op path. No more open-coded `remote === false ? … : …`.
- `OperationContext.trustedFactReads` + `TrajectoryOpts.trustedFactReads`;
  `find_trajectory` op threads `ctx.trustedFactReads` to the engine.
- `dispatch.ts` carries the flag; the **stdio** server reads
  `facts.trust_local_reads` (best-effort; a config blip falls back to the safe
  world-only default) and sets it. HTTP dispatch leaves it unset.
- `facts.trust_local_reads` registered as a config key (default off).
- `segmentValidFrom` extracted as a pure, exported helper and used in the
  extraction row build.

## Result

On a ~13.7K-page comms brain with ~58 private conversation facts:
- `find_trajectory(<entity>)` over a trusted reader returns the populated points
  list; an untrusted reader still gets `0`; local CLI unchanged. (Verified
  end-to-end against the live engine in all three modes.)
- Newly extracted facts carry the conversation date as `valid_from` instead of
  the extraction date.

## Tests

- `facts-reader-trust`: full truth table for both helpers.
- `engine-find-trajectory`: new case — `remote=true + trustedFactReads` returns
  private points while plain `remote=true` stays world-only.
- `extract-conversation-facts`: `segmentValidFrom` (real date / epoch fallback /
  unparseable).
- `bun run typecheck` clean; `bun run verify` 30/30; touched suites green.

Depends on #2357 (the Slack block-format parser) for facts to flow on a
collector-native brain.

---

### DIFF
--- a/src/commands/extract-conversation-facts.ts
+++ b/src/commands/extract-conversation-facts.ts
@@ -186,6 +186,19 @@ export interface ConversationSegment {
 }
 
 /**
+ * The `valid_from` to stamp on facts extracted from a segment: the date the
+ * claim was made (the segment's own timestamp), so find_trajectory orders on
+ * "when it happened" rather than "when extracted". Returns `undefined` for an
+ * unparseable date or the 1970 epoch fallback parseConversation emits when a
+ * page carries no date at all — letting insertFacts keep its now() default.
+ */
+export function segmentValidFrom(endIso: string): Date | undefined {
+  const d = new Date(endIso);
+  if (Number.isNaN(d.getTime()) || d.getUTCFullYear() <= 1971) return undefined;
+  return d;
+}
+
+/**
  * Core function opts. Strict — `sourceId` is always required (Eng-v2 A1).
  * Multi-source iteration is the caller's job.
  */
@@ -756,6 +769,10 @@ async function processPage(
     state.result.facts_extracted += extracted.length;
 
     if (!state.dryRun && extracted.length > 0) {
+      // Anchor each fact at WHEN IT WAS SAID — the segment's own date — not the
+      // extraction time (see segmentValidFrom).
+      const segValidFrom = segmentValidFrom(seg.endIso);
+
       // Eng-v2 C1 / E11: page-global row_num. Each fact in this batch gets
       // a unique row_num within (source_id, source_markdown_slug); the
       // accumulator increments across the segment loop.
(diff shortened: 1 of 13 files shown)


time-attack:
Thanks for the careful writeup and for keeping both engines in lockstep — the valid_from-at-claim-time observation is a genuinely good catch. I'm closing this because the second half introduces a new trust-elevation surface (trustedFactReads plus a new config key) that relaxes our fail-closed remote boundary, and that's feature surface we're not taking; the trust boundary stays as-is by design. If you want to resubmit just the segmentValidFrom dating change on its own, that's a much smaller and more reviewable proposition.

Issue #2339, opened 2026-06-21
sync: pin checkpoint write double-encodes jsonb scalar — every sync aborts, all sources go stale (0/100 doctor)


garrytan-agents:
## Summary

Multi-source brains stop syncing entirely. Every `gbrain sync` aborts at the first checkpoint write; no source can anchor, so all sources go stale and `gbrain doctor` drops to 0/100 (`cycle_freshness` FAIL across all sources).

## Symptom

```
[op-checkpoint] write failed (sync-target, <fp>): new row for relation "op_checkpoints" violates check constraint "op_checkpoints_completed_keys_array"
[sync] checkpoint target write failed (pool unavailable) — aborting before import; nothing drained, next run retries.
Sync PARTIAL at <commit>: imported 0 of N file(s), reason=checkpoint_unavailable.
```

The `(pool unavailable)` text is a **misleading wrapper** — the pool is healthy. The underlying error is SQLSTATE `23514` (CHECK violation) on `op_checkpoints.completed_keys`.

## Root cause

`recordCompleted()` in `src/core/op-checkpoint.ts` persists the sync-target pin as:

```ts
`INSERT INTO op_checkpoints (op, fingerprint, completed_keys, updated_at)
 VALUES ($1, $2, $3::jsonb, now()) ...`,
[key.op, key.fingerprint, JSON.stringify(sorted)]
```

Under `postgres.js` `.unsafe(sql, params)`, binding a JS **string** to a `$3::jsonb` param double-encodes it: the `text -> jsonb` cast wraps the already-JSON string into a jsonb **scalar string** instead of parsing it into a jsonb array. That fails the table CHECK:

```sql
CONSTRAINT op_checkpoints_completed_keys_array CHECK (jsonb_typeof(completed_keys) = 'array')
```

`sync.ts` writes this pin via `recordCompleted(ckpt.target, [pin])` **before** the first path flush (by design — so a crash resumes to the pinned target). Because that write throws on the first attempt, no source ever anchors; `appendCompleted` (the `text[]` path-flush) is never reached. The integrity constraint that was meant to protect the load now rejects every legitimate write.

## Reproduction (postgres.js `.unsafe`, against Postgres)

| binding | `jsonb_typeof(completed_keys)` |
|---|---|
| `$3::jsonb` + `JSON.stringify([...])` | ❌ throws 23514 (scalar) |
| `$3::text::jsonb` + `JSON.stringify([...])` | ✅ `array` |
[...]

PR #2428, opened 2026-06-25
fix(op-checkpoint): write completed_keys as a jsonb array, not a double-encoded scalar (#2328)

jeanpierre121:
## What

`recordCompleted` bound `JSON.stringify(sorted)` to a `$3::jsonb` param - the double-encode trap. postgres.js stringifies the string again, producing a jsonb **scalar** that the v119 `completed_keys` array CHECK rejects (`violates op_checkpoints_completed_keys_array`) on **every** write. It slipped the `check-jsonb-pattern` guard because that only matches the `${...}::jsonb` template form, not a positional param.

Impact: every `extract-conversation-facts` and `extract links --by-mention` run logged the failure and lost its resume checkpoint, so re-runs reprocessed already-completed work.

## How

Pass the JS array as a native `text[]` param and lift it to jsonb with `to_jsonb($3::text[])` - the same shape `appendCompleted` already relies on. No `JSON.stringify`, no double-encode.

## Verified

Against live Postgres: the write succeeds, `completed_keys` is `jsonb_typeof` **array**, and the key set round-trips. PGLite never reproduced the double-encode (it hid the bug), so the existing PGLite round-trip tests stayed green; added a `jsonb_typeof='array'` assertion to pin the shape. typecheck clean; verify 30/30.

---

### DIFF
--- a/src/core/op-checkpoint.ts
+++ b/src/core/op-checkpoint.ts
@@ -179,19 +179,25 @@ export async function recordCompleted(
   // REPLACE semantics (kept deliberately — #1794 V3). Callers like
   // extract-conversation-facts serialize a MUTABLE map through here and rely on
   // stale keys being REMOVED; an append would make them unremovable. The full
-  // set lands in the parent `completed_keys` JSONB column via a single UPSERT —
-  // exactly as before. JSON.stringify into `$3::jsonb` is correct (the text→jsonb
-  // cast yields a proper array; NOT the double-encode trap, which is the template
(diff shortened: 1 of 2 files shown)


time-attack:
You diagnosed this exactly right — it is the #2339 positional-param double-encode class, and the impact analysis on lost resume checkpoints is spot on. It's already fixed on master though: commit 814258dd (v0.42.53.0, #2375) binds through `$3::text::jsonb` in recordCompleted and added an AST guard for the positional form that the template-grep missed. Closing as already fixed; thank you for the sharp report.

Issue #2271, opened 2026-06-18
ollama recipe: validator rejects modern embed model dims AND models list is missing qwen3-embed-8b / snowflake-arctic-embed-l-v2


brettdavies:
**What happened?**

Two coupled gaps in the Ollama recipe block operators who want to run modern Ollama embedding models:

1. **The validator rejects every dim that isn't `default_dims`.** `gbrain init --embedding-model ollama:mxbai-embed-large --embedding-dimensions 1024` fails preflight with `Provider "ollama" model "mxbai-embed-large" does not support custom dimensions 1024 (this model only emits its default vector size).` Misleading: the local Ollama backend was perfectly happy to emit 1024d (mxbai's native size). The recipe declares `default_dims: 768` (nomic-embed-text) and gates every other model against it. Same hole for `snowflake/arctic-embed-l-v2` (1024d) and Qwen3-Embedding (0.6B/4B/8B → 1024/2560/4096 native). The `llama-server` and `litellm-proxy` recipes are in the same bucket: they declare `user_provided_models: true`, but Tier 3 of `isCustomDimValidForProvider` still rejects whatever dim the user picks.
2. **The recipe's models list is missing the modern embedding models.** Even with the validator relaxed, `qwen3-embed-8b` and `snowflake/arctic-embed-l-v2` are absent from the registered list, so the resolver doesn't recognize them as known-Ollama models. Operators today have to either set `models.tier.embed ollama:<exotic-model>` explicitly (bypassing the recognized-list check) or `ollama pull` and pray the validator doesn't refuse.

**What did you expect?**

The Ollama recipe accepts the modern embedding models at their native dims with no operator-side workaround:

- `gbrain init --embedding-model ollama:qwen3-embed-8b --embedding-dimensions 4096` works.
- `gbrain init --embedding-model ollama:snowflake-arctic-embed-l-v2 --embedding-dimensions 1024` works.
- `gbrain models | grep ollama` lists both as recognized Ollama embedding models.

**Steps to reproduce**

1. Pull a non-nomic embedding model: `ollama pull mxbai-embed-large` (or `qwen3-embed-8b`, or `snowflake/arctic-embed-l-v2`).
2. `gbrain init --embedding-model ollama:mxbai-embed-large --embedding-dimensions 1024` → preflight rejects with the misleading "does not support custom dimensions" message.
3. `gbrain models | grep qwen3-embed-8b` → no row (the model isn't in the recipe's known list).

[...]

PR #2429, opened 2026-06-26
v0.42.57.0 fix(embedding): accept configured user-provided dimensions

TheAngryPit:
## Summary

Centralizes the embedding-provider contract used by the gateway and schema dimension checks so user-provided OpenAI-compatible embedding models can declare their dimensions instead of being rejected as unknown models.

This specifically lets LiteLLM/llama-server style embedding aliases work when they are explicitly configured with dimensions.

## What changed

- Adds `src/core/ai/embedding-provider-contract.ts` as the shared contract for provider-owned and user-provided embedding touchpoints.
- Routes gateway embedding diagnostics through that contract instead of duplicating provider/model checks.
- Lets `resolveSchemaEmbeddingDim()` accept configured user-provided embedding models and reject unconfigured overrides with a useful diagnostic.
- Adds focused regression coverage for LiteLLM and llama-server embedding models.
- Updates `docs/architecture/KEY_FILES.md` for the new contract file.

## Verification

- `bun test --timeout 30000 test/litellm-user-provided-embedding.test.ts test/embedding-dim-check.test.ts test/embed-preflight.test.ts test/gateway-embed-model-override.test.ts`
- `bun run typecheck`
- `bun run check:doc-history`
- `git diff --check upstream/master...HEAD`

## Notes

This branch intentionally does not include a release/version bump or CHANGELOG entry. It is a small upstream contribution branch and can be folded into the maintainer's next ship/version wave.


time-attack:
Closing: current master already contains this fix. Master already accepts configured user-provided embedding dims via the #2271 trust_custom_dims passthrough tier (src/core/embedding-dim-check.ts isCustomDimValidForProvider, set on src/core/ai/recipes/litellm-proxy.ts:28, llama-server.ts:35, ollama.ts:29). The gateway guard the PR modifies no longer exists: src/core/ai/gateway.ts:779-796 (#1292/D6) removed user_provided_model_unset — with a comment stating it 'only ever false-positived for a fully-specified litellm:<model> with dims set, silently disabling vector search' — and replaced it with user_provided_dims_unset. The PR's diagnoseEmbedding change would reintroduce an _extendedModels registration requirement master deliberately rejected. mergeStateStatus DIRTY.

PR #2430, opened 2026-06-26
v0.42.58.0 fix(config): harden gateway config adapter boundary

TheAngryPit:
## Summary

Hardens the gateway configuration adapter boundary so command code stops rebuilding partial `GatewayConfig` objects by hand.

The branch keeps provider base URLs, file-plane API keys, embedding model selection, and multimodal model selection flowing through `buildGatewayConfig()` consistently.

## What changed

- Replaces scattered ad-hoc gateway config object construction with the shared `buildGatewayConfig()` adapter.
- Keeps provider base URL and API key precedence in one place instead of repeating partial mappings across command paths.
- Adds regression coverage proving `gbrain init` embedding diagnostics use the gateway config adapter, including file-plane API keys.
- Tightens affected tests around gateway config output instead of relying on duplicated expected object construction.
- Updates `docs/architecture/KEY_FILES.md` to describe the adapter boundary.

## Verification

- `bun test --timeout 30000 test/init-embed-check.test.ts test/ai/build-gateway-config.test.ts test/loadConfig-merge.test.ts test/cli-multimodal-integration.test.ts test/providers.test.ts`
- `bun run typecheck`
- `bun run check:doc-history`
- `git diff --check upstream/master...HEAD`

## Notes

This is independent of #2429, but it pairs well with it because both reduce duplicated embedding/gateway configuration logic.

This branch intentionally does not include a release/version bump or CHANGELOG entry. It is a small upstream contribution branch and can be folded into the maintainer's next ship/version wave.


time-attack:
Thanks for taking the time to trace the gateway config paths — routing everything through buildGatewayConfig is a reasonable instinct. I'm closing this because it's a refactor rather than a fix: nothing is currently broken by the hand-built partial configs, and a ~180-line reshuffle across six command files is churn we'd have to re-review against every in-flight gateway change. If you hit an actual precedence bug caused by one of these call sites, a targeted fix for that bug would be very welcome.

PR #2434, opened 2026-06-26
feat(extract): keep frontmatter links fresh on the incremental cycle

spiky02plateau:
## Problem

The autopilot/dream **cycle's extract phase** re-extracts only the slugs a sync changed (`extractForSlugs`), but it extracts **body links only**. A page's frontmatter link fields (`sources:` / `related:` etc.) are never re-extracted on the incremental path. So when frontmatter is edited externally (e.g. in Obsidian) and synced in, those edges silently **drift stale** relative to body edges.

Only two paths refresh frontmatter edges today: a `put_page` write, or a manual `gbrain extract links --include-frontmatter`. Plain `sync` + the cycle do not — so in a vault where frontmatter `sources:`/`related:` is curated heavily, the drift is continuous and operator-visible.

## Root cause

`extractLinksFromFile` already accepts and applies `{ includeFrontmatter }` per file. The incremental path just never threads the flag:

```ts
// extractForSlugs → src/commands/extract.ts
const links = await extractLinksFromFile(content, relPath, allSlugs, { globalBasename });
//                                                                     ^ includeFrontmatter dropped
```

`runExtractCore`'s `opts.slugs !== undefined` branch routes to `extractForSlugs` and never reaches the full-walk path, so the cycle never gets frontmatter extraction.

## Fix

Thread an `includeFrontmatter` flag from the cycle's extract dispatch through `ExtractOpts → extractForSlugs → extractLinksFromFile`. Because it's applied **per changed page inside the already-bounded slug set**, extraction stays incremental — no switch to a full DB scan. Cost = one extra frontmatter parse per *changed* page per cycle.

Gated behind a config key, **default off** so existing brains see zero behavior change on upgrade:

```
gbrain config set autopilot.incremental_extract_include_frontmatter true
```

## Relation to #2406

#2406 makes the *put_page* / *manual extract* paths resolve bracketed frontmatter values via `global_basename`. This PR closes the complementary gap on the *cycle/sync* path so externally-edited frontmatter stays fresh without a manual sweep. Edges land as `link_source='frontmatter'` (or `'frontmatter-basename'` once #2406's basename resolution is enabled). Together they make frontmatter edges first-class on every write path.

## Tests

Two regression cases added to `test/extract-incremental.test.ts` (PGLite, no DB connection):
- gate **off** → a frontmatter-only page yields `0 links` on the incremental path (body-only behavior preserved)
- `includeFrontmatter: true` → the `source:` frontmatter edge is extracted

`bun run verify` + the extract/link/cycle test surface pass locally.

## Note on the threading

`extractForSlugs` takes positional params; I added `includeFrontmatter` as a trailing optional (default `false`) to keep the change minimal and back-compat. Happy to refactor it to an opts object if you'd prefer that shape.

### DIFF
--- a/src/commands/extract.ts
+++ b/src/commands/extract.ts
@@ -536,6 +536,17 @@ export interface ExtractOpts {
    * paths: extractForSlugs, extractLinksFromDir, extractTimelineFromDir.
    */
   signal?: AbortSignal;
+  /**
+   * v0.42 — also extract frontmatter links on the incremental (slugs) path.
+   * `extractForSlugs` extracts BODY links only by default; set this true to also
+   * parse each changed page's frontmatter so `sources:`/`related:` edges stay fresh
+   * when YAML is edited externally and synced in. Applied PER changed page, so the
+   * incremental walk stays bounded (no switch to a full DB scan). Only honored on
+   * the incremental path (`slugs` defined); the full-walk path already covers
+   * frontmatter via its own dispatch. Gated upstream by the config key
+   * `autopilot.incremental_extract_include_frontmatter` (default off).
+   */
+  includeFrontmatter?: boolean;
 }
 
 /**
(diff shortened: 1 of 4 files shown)


time-attack:
Superseded with credit by #3087 (just merged) — your change landed rebased onto current master with authorship preserved via Co-authored-by. Thank you!

Issue #430, opened 2026-04-25
`put_page` silently corrupts mixed-case slugs: page row created with lowercased slug, but `upsertChunks` queries with the original case and throws `Page not found`


berncode:
### Summary

`put_page` fails when given a slug containing uppercase characters. The page row is created with the lowercased slug (via `validateSlug`), but the chunk-upsert step in the same transaction queries by the original mixed-case slug, returns 0 rows, and throws `Page not found: <originalSlug>`. The transaction rolls back. From the caller's perspective, a `put_page` call returns a `Page not found` error, which is misleading — the operation is a write, and the page row was created (and rolled back) inside the same transaction.

### Reproduction

```bash
curl -s -X POST http://localhost:8000/mcp \
  -H "Content-Type: application/json" \
  -H "Accept: application/json, text/event-stream" \
  --data '{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"name":"put_page","arguments":{"slug":"Test/Mixed-Case-Slug","content":"---\ntype: note\ntitle: t\n---\n\nbody"}}}'
```

**Expected:** page is created (and chunked) at the canonical lowercased slug.

**Actual:**

```
{"result":{"content":[{"type":"text","text":"Error: Page not found: Test/Mixed-Case-Slug"}],"isError":true},"jsonrpc":"2.0","id":1}
```

A subsequent `list_pages` shows no `test/mixed-case-slug` page either — the row was created and rolled back, not persisted.

### Root cause

Inside one `importFromContent` transaction, `putPage` and `upsertChunks` handle the slug differently:

- `src/core/utils.ts:8-13` — `validateSlug(slug)` returns `slug.toLowerCase()`.
- `src/core/postgres-engine.ts:131-132` — `putPage` calls `slug = validateSlug(slug)` before the INSERT.
- `src/core/postgres-engine.ts:393-399` — `upsertChunks` does **not** call `validateSlug`. It runs `SELECT id FROM pages WHERE slug = ${slug}` against the caller-supplied (mixed-case) slug.
- `src/core/import-file.ts:269-296` — `importFromContent` calls `tx.putPage(slug, ...)` and then `tx.upsertChunks(slug, chunks)` with the same `slug` variable, never lowercased at the importer level.

The same pattern exists in `src/core/pglite-engine.ts` (`putPage` validates, `upsertChunks` doesn't).

**Suggested fix**

Lowercase consistently. Three options, in order of strictness:

[...]

PR #2436, opened 2026-06-26
fix(import): canonicalize slug in importFromContent so mixed-case put_page with tags doesn't roll back

fxa3bah:
## Problem

`putPage` canonicalizes the slug (lowercases via `validateSlug`) before storing the row, but the tag/link/timeline reconcilers (`tx.addTag`, `addLink`, `addTimelineEntry`) look the page up by the slug **as passed**. So a `put_page` with a capitalized slug — e.g. `Projects/Team-Wiki/Quarterly-Roadmap` — stores the page under `projects/team-wiki/quarterly-roadmap`, then the tag step's existence check searches the original casing, finds nothing, and throws `addTag failed: page "…" not found`.

Because that throw happens inside the import transaction, the **entire write rolls back**: the page never persists under either casing. Any caller driving the HTTP MCP server (where slugs arrive verbatim) silently loses every page whose slug has a capital letter plus a frontmatter tag. The CLI capture path is unaffected because it already passes `slugifyPath()` output (lowercase).

## Fix

Canonicalize the slug once at the top of `importFromContent` — the single chokepoint shared by the MCP `put_page` op and CLI capture — so `putPage` and every reconciler agree on the lowercased slug. No-op for disk imports (already `slugifyPath()` output), idempotent with `putPage`'s own `validateSlug` call. Engine-agnostic, so PGLite and Postgres move together with no per-engine change.

## Test

`test/put-page-mixed-case-slug-tags.test.ts` (PGLite, hermetic) pins the regression: a capitalized slug + frontmatter tags now persists under the canonical lowercased slug with tags reconciled, plus a lowercase control case. Adjacent suites stay green: `put-page-provenance`, `put-page-namespace`, `slug-validation`, `import-file`, `reindex-preserve-tags`. `bun run typecheck` clean.

No version bump included — leaving release versioning to maintainer `/ship`.

### DIFF
--- a/src/core/import-file.ts
+++ b/src/core/import-file.ts
@@ -36,7 +36,7 @@ import {
 } from './embedding-context.ts';
 import { loadSearchModeConfig, resolveSearchMode } from './search/mode.ts';
 import { normalizeAliasList } from './search/alias-normalize.ts';
-import { isUndefinedTableError, warnOncePerProcess } from './utils.ts';
+import { isUndefinedTableError, warnOncePerProcess, validateSlug } from './utils.ts';
 import { computeCorpusGeneration } from './contextual-retrieval-service.ts';
 import { runGuardrails } from './guardrails.ts';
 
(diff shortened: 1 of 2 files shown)


justemu:
We hit this exact bug in production today (2026-07-08) with a 25k-page brain.

The Hermes MCP client was calling put_page with mixed-case slugs containing ASCII uppercase (e.g., HLX05-N, MRCT). validateSlug() lowercases them in putPage, but addTag in the same transaction uses the original casing — causing "page not found" and full transaction rollback.

Impact: 81 skipped writes per dream_phase1 cycle, intermittent gbrain unreachability (Hermes 3-strike rule).

This PR is the correct fix. Single chokepoint at importFromContent entry prevents mismatches across all downstream reconcilers (addTag, upsertChunks, addLink, etc.). #430 documented the same for upsertChunks.

We are applying this patch locally. Please merge + release when possible.

Error evidence from our logs:
```
2026-07-08 20:23:46 addTag failed: page "companies/..." (source=default) not found
2026-07-08 20:23:47 addTag failed: page "concepts/..." (source=default) not found
2026-07-08 20:24:10 MCP server gbrain is unreachable after 3 consecutive failures
```

PR #2439, opened 2026-06-26
Add local Ollama chat routing

lthibault:
## Problem

GBrain already has a local-first story for embeddings through Ollama, but higher-level reasoning paths still effectively require a hosted chat provider. That means users who choose Ollama for privacy, cost control, or offline/local development can embed and search locally, but cannot use the same local stack for query expansion, `think`, facts extraction, or dream synthesis.

Ollama exposes an OpenAI-compatible `/v1/chat/completions` API, so gbrain can route those reasoning calls locally without adding a new transport.

## Summary

- add Ollama expansion and chat touchpoints using the local OpenAI-compatible API
- mark Ollama chat as non-tool/non-subagent-loop capable while allowing local reasoning flows
- document local Ollama setup for embeddings, chat, dream, think, and facts extraction
- add recipe, gateway, and capability regression coverage

## Notes

This does not make Ollama a subagent/tool-loop provider. The recipe explicitly marks tool calling, prompt caching, and subagent-loop support as unsupported. Users who need gbrain's subagent/tool loop should still route that tier to a tool-capable provider.

## Testing

- `bun test test/ai/recipe-ollama.test.ts test/ai/gateway-chat.test.ts test/ai/capabilities.test.ts`
- `bun test test/ai/recipes-existing-regression.test.ts test/providers.test.ts test/init-env-detection.test.ts`
- `bun run typecheck`
- `bun src/cli.ts providers list`

### DIFF
--- a/docs/integrations/embedding-providers.md
+++ b/docs/integrations/embedding-providers.md
@@ -40,7 +40,7 @@ The resolved provider + dimensions get persisted to `~/.gbrain/config.json` atom
 | `deepseek` | (no embedding model — chat only) | — | — | — | — |
 | `groq` | (no embedding model — chat only) | — | — | — | — |
 
-**Note on local providers.** Ollama and llama-server have no required API key, so they don't show up in env-detection auto-pick. Pick them explicitly with `--embedding-model ollama:<model>` to avoid silently routing to a daemon that may not be running.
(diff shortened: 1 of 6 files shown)


time-attack:
Thanks for the clean recipe work and for being upfront about the tool-loop and prompt-cache limitations — that honesty is appreciated. We're closing this one on policy: new provider/vendor integrations are surface area we're not adding right now, and the existing gateway config already lets users point an OpenAI-compatible base_url at a local Ollama server for these touchpoints. Not a reflection on the quality of the patch.

Issue #2426, opened 2026-06-25
Write-through content stays DB-only and is silently deleted by `sync --full` (3 compounding durability bugs)


asmittyactivate:
## Summary

On a brain that uses the markdown write-through path (`sync.repo_path`), **write-through content never reaches git, and `sync --full` then silently deletes it.** Three compounding issues turn "DB-only content" (which is unavoidable given #1/#2) into permanent data loss. We hit this in production: a `sync --full` delete-reconcile soft-deleted ~2,780 genuine pages (years of daily digests, call transcripts, a long-form strategy doc, a real contact page) whose only fault was that their markdown had never been committed to git.

Versions: engine `0.42.52`, durability scaffolding installed via `gbrain sources harden` (post-commit hook + `scripts/brain-commit-push.sh` + 30-min pull cron).

## 1. Write-through is never committed/pushed

`put_page` / `capture` and the enrichment cycle write `.md` into the `sync.repo_path` clone (write-through), but nothing commits or pushes it:

- `gbrain sources harden` (verified via `--dry-run`) installs the post-commit hook, the `brain-commit-push.sh` helper, the durability RESOLVER rules, and the pull cron, and it pulls — but it does **not** itself commit pending write-through. Output on a dirty tree: `[pull] skipped: working tree dirty — pull skipped (in-progress edits preserved)`. No commit/push step.
- The post-commit hook is documented as "best-effort background auto-push," but it only fires **after a commit** — and write-through never makes one.
- Nothing invokes `brain-commit-push.sh` for enrichment write-through (modified `people/*.md`, `companies/*.md`, daily/email digests).

**Result:** write-through `.md` accumulates uncommitted in the clone indefinitely (we found 64 such files, mtimes days after their last commit) and never reaches the remote. There's a pull cron but no push trigger.

**Repro:** run an enrichment/dream cycle, then `git -C <clone> status --porcelain '*.md'` → modified pages, uncommitted, never pushed.

## 2. `brain-commit-push.sh` pulls before staging → fails on any modified file

The shipped helper does `git fetch` + `git pull --rebase` **before** `git add -- "$@"`. On any dirty tree (i.e. a modified/enriched page — exactly the write-through case) the rebase aborts:

```
error: cannot pull with rebase: You have unstaged changes.
error: please commit or stash them.
rebase conflict: manual attention needed   # exit 3
```

[...]

PR #2441, opened 2026-06-26
fix(durability): set rebase.autoStash during harden so pull --rebase tolerates uncommitted write-throughs

asmittyactivate:
## Problem

`gbrain capture` / `put_page` writes the DB row **and** the clone `.md`, but does **not** commit — durability is the writer's job via `scripts/brain-commit-push.sh`. In practice, concurrent captures and gbrain's own autopilot/dream-cycle passes leave **uncommitted tracked pages** in the brain clone's working tree.

The committed helper (`brain-commit-push.sh`) and the durability rules both start with `git pull --rebase`, which **hard-fails on any unstaged tracked change**:

```
error: cannot pull with rebase: You have unstaged changes.
error: please commit or stash them.
```

Once one uncommitted page is present, that first `git pull --rebase` fails, so **every subsequent commit-push wedges** — including fully-compliant collectors that *do* call the helper. Pages then stop reaching GitHub entirely.

I hit this live on a headless OpenClaw brain: **29 pages (including a real meeting transcript) were stuck in the DB + clone but never pushed for ~3h**, until a manual `git pull --rebase` + push cleared the backlog. It recurs whenever an uncommitted write-through accumulates.

## Fix

Set `rebase.autoStash` (repo-scoped, idempotent) during `hardenBrainRepo`, before the pull step — so the harden's own pull, the committed helper's pull, and the agent durability rules' `git pull --rebase` all **stash → rebase → reapply** instead of failing on a dirty tree.

- One statement, reuses the existing `gitConfigSet` helper.
- Repo-scoped local config; honors `dryRun`; non-fatal.
- Idempotent — re-running harden is a no-op on the config.

Verified RED→GREEN on the same git: with an uncommitted file, `git pull --rebase` errors as above; after `git config rebase.autoStash true`, the same pull succeeds.

I could not run the full gbrain test suite from my environment, so this is offered honest-blocked on CI/tests — happy to adjust placement (e.g. promote it to a reported harden `StepName`) if you'd prefer it surfaced as its own step.

### DIFF
--- a/src/core/brain-repo-durability.ts
+++ b/src/core/brain-repo-durability.ts
@@ -6,6 +6,8 @@
  * sessions edit a stale tree. The moment gbrain is given a PAT + a GitHub URL
  * for a brain repo, `hardenBrainRepo` makes durability work, idempotently:
  *
+ *   0. set `rebase.autoStash` so `git pull --rebase` tolerates uncommitted
+ *      write-throughs (otherwise the first unstaged page wedges every commit-push)
  *   1. pull current state (divergence-safe rebase; skip-on-dirty)
  *   2. repo-scoped credential wiring (reuse an existing helper if present)
  *   3. LOCAL untracked post-commit hook (best-effort background auto-push)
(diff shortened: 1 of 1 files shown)


time-attack:
This was a real and painful failure mode, and your write-up of it is excellent — 3h of unpushed pages is exactly the kind of thing we want reported. It's addressed on master already: #2426 (in commit 42375bde) reordered brain-commit-push.sh to stage and commit before pulling, and hardenBrainRepo's own pull now reports skipped_dirty instead of aborting, so a dirty tree no longer wedges the chain. We're also wary of repo-wide rebase.autoStash because a conflicting reapply can leave write-throughs parked in a stash. Closing as already fixed — thank you for the field report.

PR #2442, opened 2026-06-26
v0.42.27.0 fix(ai-gateway): fix silent no-op in Anthropic prompt caching; wire OpenAI prompt_cache_key

CoachRyanNguyen:
## What this fixes

Anthropic prompt caching in the chat gateway was a **silent no-op**. `chat()` set `cache_control` on the top-level `generateText` call options, but `@ai-sdk/anthropic` reads cache markers only from each system-message's / tool's own `providerOptions` — and `ai` core converts a bare `system:` string into a system message with **no** `providerOptions`. So every `cacheSystem:true` caller (`enrich`, page summaries, skillopt, the gateway tool loop) wrote **zero** cache breakpoints while the usage telemetry kept reading 0 cache tokens — nothing looked broken, but no caching happened.

Verified against the installed SDKs: `@ai-sdk/anthropic` 3.0.74 (`validator.getCacheControl(providerOptions, …)` is per-message/per-tool), `ai` 6.0.174 (`standardizePrompt` drops a string system's providerOptions). The SDK's own `SystemModelMessage` docs say this is the supported path for caching.

## Changes

`src/core/ai/gateway.ts`:
- **`buildChatCacheArgs()`** (new, pure, exported for tests) — places markers where the provider reads them: the system prompt as a `SystemModelMessage` carrying `providerOptions`, plus `cache_control` on the **last tool def** (Anthropic caches the whole tools→system prefix up to the marked block). Restores the documented `ChatOpts.cacheSystem` contract ("cache system prompt + last tool def"). Mirrors the already-correct native subagent path.
- **`openAIPromptCacheKey()` + `ChatOpts.cacheKey`** (new) — sets `providerOptions.openai.promptCacheKey` (`prompt_cache_key`) for **native-openai only**, derived from `sha256(system + sorted tool names)`. OpenAI caches prefixes automatically; the stable key keeps shared-prefix requests on the same engine to raise the hit rate. Tool order does not change the key; explicit `cacheKey` overrides.
- Removed the dead top-level `providerOptions.anthropic.cacheControl` (it was the no-op). Added imports: `type SystemModelMessage` from `ai`, `createHash` from `node:crypto`.

`test/gateway-prompt-cache.test.ts` (new) — 10 cases pinning marker placement + routing-key derivation. The load-bearing assertion (`typeof system).not.toBe('string')` when caching is on) reproduces the exact failure mode so it can't silently regress.

## Testing

- New suite: 10 pass / 0 fail
- Adjacent gateway tests (`gateway-model-messages`, `facts-meta-cache`, `think-gateway-adapter`): 28 pass / 0 fail
- `bun run typecheck`: 0 errors
- Version audit: VERSION / package.json / CHANGELOG all at `0.42.27.0`

## How to confirm the savings end-to-end

The unit tests pin placement; the definitive proof is runtime telemetry (already captured in `ChatResult.usage`): run a `cacheSystem:true` caller (e.g. `gbrain enrich`) twice with the same system prompt and check `cache_read_tokens` goes 0 → non-zero on the second call. For OpenAI, watch `providerMetadata.openai` cached-token counts climb across same-system calls.

## Notes

- Gemini and other providers are unchanged (implicit caching only — the right call).
- The full multi-agent pre-landing review army was not run; this change is small, SDK-source-verified, and unit-pinned.


time-attack:
Superseded with credit by #2933 (just merged) plus dependency behavior: the Anthropic cache_control half is already live on master — @ai-sdk/anthropic 3.0.74 forwards the gateway's providerOptions cacheControl as request-level cache_control (and #2771 covers the direct-SDK conversation path). Your OpenAI prompt_cache_key wiring was the genuine remnant and is now ported (stable key from system prompt + tool names, config override wins), with your authorship preserved via Co-authored-by. Thank you!

PR #2444, opened 2026-06-27
fix: handle CRLF skill frontmatter in health checks

Cjaewon0708:
## Summary

- Allow skill frontmatter parsing to handle Windows CRLF line endings.
- Allow block array parsing and check-resolvable trigger extraction to work with `\r\n`.
- Add regression coverage for CRLF frontmatter in `parseSkillFrontmatter` and `checkResolvable`.
- Add the missing `skill-optimizer` resolver route.

## Why

On Windows, `SKILL.md` files may use CRLF line endings. Some health/check-resolvable logic only matched LF-only frontmatter fences, which can cause valid skills or frontmatter trigger arrays to be missed during health checks.

## Verification

- `git diff --check origin/master..HEAD`
- `bun test test/check-resolvable.test.ts test/skill-brain-first.test.ts`
- `bun src/cli.ts check-resolvable --strict --skills-dir skills/`
- `bun run typecheck`

### DIFF
--- a/skills/RESOLVER.md
+++ b/skills/RESOLVER.md
@@ -60,6 +60,7 @@ This is the dispatcher. Skills are the implementation. **Read the skill file bef
 | "get more out of gbrain", "is my brain set up right", "weekly brain checkup", "advise me on my brain", "gbrain advisor" | `skills/gbrain-advisor/SKILL.md` |
 | Save or load reports | `skills/reports/SKILL.md` |
 | "Create a skill", "improve this skill" | `skills/skill-creator/SKILL.md` |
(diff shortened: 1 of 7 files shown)


time-attack:
You're right that parseSkillFrontmatter still only matches LF fences on master — that part of the diagnosis holds. I can't take this patch as written, though: the blockRe change swaps the double-escaped RegExp string for a plain template literal, so `\s` becomes a literal `s` and the pattern turns into `^<field>:s*…`, which breaks block-array parsing for LF files too. It also mixes in Windows path-casing logic in write-through.ts and a new resolver route that aren't CRLF fixes. A one-line `content.replace(/\r\n/g, '\n')` at the top of parseSkillFrontmatter, nothing else, would be a far safer submission.

PR #2449, opened 2026-06-27
[codex] Exclude resolved non-bets from calibration metrics

Vietduc88x:
## What changed

- Scope scorecard aggregates to `kind = 'bet'` in both Postgres and PGLite.
- Apply the same bets-only boundary to calibration-curve buckets.
- Add resolved fact, take, forecast, and hunch controls to the PGLite regression suite and Postgres/PGLite parity fixture.

## Root cause

`total_bets` filtered on `kind = 'bet'`, but the remaining scorecard aggregates and calibration curve selected resolved rows of every kind. Resolved non-bet claims could therefore alter `resolved`, outcome counts, partial/unresolvable rates, Brier score, and calibration buckets.

## Impact

Scorecards and calibration curves now reflect only forward bets, matching their API contract. Existing legitimate bet metrics are unchanged.

## Validation

- `bunx tsc --noEmit --pretty false`
- `bun test test/takes-engine.test.ts test/takes-resolution.test.ts test/take-forecast.test.ts` — 63 passed
- `DATABASE_URL=postgresql://... bun test --timeout 120000 test/e2e/takes-scorecard-parity.test.ts` against an isolated pgvector/Postgres 16 container — 10 passed
- `git diff --check`

The isolated test database was removed after verification.

### DIFF
--- a/src/core/pglite-engine.ts
+++ b/src/core/pglite-engine.ts
@@ -4511,6 +4511,8 @@ export class PGLiteEngine implements BrainEngine {
     // (correct|incorrect|partial) — NOT `resolved_quality IS NOT NULL` — so
     // historical comparisons against pre-v74 scorecards stay valid.
     // `unresolvable_count` is a sibling field counting the new 4th state.
+    // The whole aggregate is bets-only; resolved facts/takes/forecasts must
+    // never affect outcome counts, rates, Brier, or calibration.
     const res = await this.db.query(
       `SELECT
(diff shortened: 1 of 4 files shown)


time-attack:
Thanks for the thorough validation, including the Postgres/PGLite parity run — and your read of the aggregate scoping is accurate against current master. We're closing this because it's still in draft and the impact is metrics precision on a narrow takes/bets surface rather than a user-blocking or data-loss issue, which is the bar for taking draft work. No criticism of the analysis, which was solid.

PR #2452, opened 2026-06-27
fix(calibration,takes,cli): model resolution, source-scoped takes reads, BigInt-safe output, calibration CLI reachability

### LINKED ISSUE #2035: cli: calibration missing from CLI_ONLY set → "Unknown command" despite dispatch case existing
## Bug

`gbrain calibration` reports `Unknown command: calibration` even though the dispatch case exists:

```
$ gbrain calibration
Unknown command: calibration
Run gbrain --help for available commands.
```

Found on v0.37.11.0 (2026-05-22); **confirmed still present at v0.42.1.0 (`eefe8b57`)**: the dispatch case is at `src/cli.ts:1567` but `calibration` is missing from the `CLI_ONLY` set at `src/cli.ts:38`.

## Root cause

`src/cli.ts` has two parallel dispatch paths: commands in the `CLI_ONLY` set go to `handleCliOnly`; anything else falls through to the cliOps map and then the "Unknown command" error. `calibration` is in the `handleCliOnly` switch but not in `CLI_ONLY`, so it can never reach its own case.

## Fix

One word added to the `CLI_ONLY` set literal:

```diff
@@ src/cli.ts:38
-... 'status', 'skillopt']);
+... 'status', 'skillopt', 'calibration']);
```

## Impact

The entire `gbrain calibration` CLI family (print/regenerate the active profile) is inaccessible. The MCP `get_calibration_profile` op works, but the documented CLI command doesn't — blocks the v0.36.1.0 hindsight-calibration wave from CLI use.

## Suggested audit

A sweep of `case '...':` labels inside `handleCliOnly` against the `CLI_ONLY` set would catch any other top-level commands with the same dispatch gap.

### LINKED ISSUE #2450: CLI local-op output normalization crashes on BigInt (cli.ts JSON.stringify has no replacer)
## Summary

The local-engine CLI path normalizes every op return value with a bare
`JSON.parse(JSON.stringify(rawResult))` (`src/cli.ts:445`). When `rawResult`
contains a JS `bigint` — e.g. a `BIGSERIAL` primary key read back by the
Postgres engine — `JSON.stringify` throws `TypeError: Do not know how to
serialize a BigInt`, so the command crashes before any renderer runs.

The ENG-2 comment directly above that line documents the intended behavior:

> Date → ISO string; bigint → string (postgres.js shape); Buffer → object.

…but a plain `JSON.stringify` **cannot** turn a `bigint` into a string — it
throws. So the promised `bigint → string` parity is never actually achieved;
the op errors out instead.

## Repro

Any local op whose return shape includes a `bigint` column. The one we hit is
`get_calibration_profile` once a `calibration_profiles` row exists
(`id BIGSERIAL PRIMARY KEY`, `src/schema.sql:1227`):

```
gbrain call get_calibration_profile --holder <h>   # local-engine op path → cli.ts:445
```

With **no** profile row it returns `null` and serializes fine — the crash only
appears once data exists, which makes it read as intermittent.

(`CalibrationProfileRow.id` is typed `number` in `src/commands/calibration.ts`,
but `getLatestProfile`'s `SELECT id …` returns the `BIGSERIAL` as a JS bigint,
so the type annotation is also masking the hazard.)

## Root cause

`src/cli.ts:445`

```ts
const result = JSON.parse(JSON.stringify(rawResult));
```

No replacer → a bigint anywhere in the tree is unserializable → throw.

## Suggested fix

Add a replacer that mirrors the postgres.js wire shape the comment already
promises (postgres.js returns `int8` as a string):

```ts
const result = JSON.parse(
  JSON.stringify(rawResult, (_k, v) => (typeof v === 'bigint' ? v.toString() : v)),
);
```

This makes the local path match the routed path and removes the whole bigint
crash-class for every local op, not just calibration.

Note: the CLI-only `gbrain calibration --json` path has its own bare
`JSON.stringify(profile, …)` (`src/commands/calibration.ts`) with the same
hazard; it's currently unreachable due to #2035, but will surface the same
crash once that command is wired up, so it deserves the same treatment.

## Related (not duplicates)

- #2433 and #2256 are the same *family* (Date/timestamp serialization on the
  gateway tool-result path) but a different layer and root cause. This one is
  the CLI local-op output normalizer.

### LINKED ISSUE #2451: calibration_profile + voice-gate use bare model ids → parseModelId throws "missing a provider prefix"
## Summary

The `calibration_profile` cycle phase and its voice gate hardcode **bare** model
ids with no provider prefix. `gateway.chat()` runs the model string through
`parseModelId` (`src/core/ai/model-resolver.ts:41`), which throws
`Model id "<id>" is missing a provider prefix.` for any string lacking a
`provider:` / `provider/` prefix. So the moment the phase actually invokes the
LLM (holder has ≥5 resolved bets, budget allows, no test judge injected), it
crashes.

## Bare ids

- `src/core/cycle/calibration-profile.ts:231`
  ```ts
  const modelId = opts.model ?? 'claude-sonnet-4-6';
  ```
  This `modelId` feeds the budget estimate (`checkBudget`, :273), the
  pattern/bias generation via `gatewayChat`, and the stored `model_id` row
  (:378) — all off a bare id.

- `src/core/calibration/voice-gate.ts:162` (`defaultJudge`, the production judge
  per `voice-gate.ts:203` → `judge = opts.judge ?? defaultJudge`)
  ```ts
  model: 'claude-haiku-4-5',
  ```

## Repro

Run the `calibration_profile` phase for a holder with ≥5 resolved bets and a
non-zero budget, without injecting a test `voiceGateJudge` / `patternsGenerator`.
The first `gateway.chat()` call throws `missing a provider prefix`.

## Suggested fix

Resolve through the existing `TIER_DEFAULTS` (`src/core/model-config.ts:74`),
which already carry the provider prefix and are the single source of truth for
tier defaults:

- calibration-profile default → `TIER_DEFAULTS.reasoning`
  (`anthropic:claude-sonnet-4-6`)
- voice-gate judge default → `TIER_DEFAULTS.utility`
  (`anthropic:claude-haiku-4-5-20251001`)

That keeps the budget estimate, the gateway call, the voice gate, and the
stored `model_id` consistent on a single resolved `provider:model`.

## Related (not duplicates)

- #1857 is the same class (hardcoded judge model in `takes extract --from-pages`)
  but a different code path — and it at least carries the `anthropic:` prefix.
  These two are bare (no prefix) and crash on `parseModelId` instead of silently
  returning 0 claims.
- #1270 is hardcoded-model-rot in cross-modal-eval — adjacent, different site.

spinsirr:
## Summary

Four fixes around the `calibration_profile` phase, takes read scoping, and the
local CLI — each verified against `upstream/master` (v0.42.53.0). Issues were
filed first; this PR closes them.

**2026-07-12 update** — folded in the two gaps flagged in the #2450 comment
thread ([takes_search MCP path](https://github.com/garrytan/gbrain/issues/2450#issuecomment-4921442283),
[`gbrain call` exit](https://github.com/garrytan/gbrain/issues/2450#issuecomment-4954688069));
see the new section below.

### BigInt-safe local-op output normalization — Fixes #2450
`cli.ts` normalized every local-engine op result with a bare
`JSON.parse(JSON.stringify(rawResult))`. A bigint column (e.g.
`calibration_profiles.id BIGSERIAL`) threw `TypeError: Do not know how to
serialize a BigInt` and crashed the command — the ENG-2 comment promised
`bigint → string` but a plain `JSON.stringify` can't do that. Added a replacer;
covers every local op, `get_calibration_profile` included.

### #2450 comment thread: takes_search MCP path + `gbrain call` exit
Two sibling crash sites reported in the issue comments, same BigInt class:

- **`searchTakes` / `searchTakesVector`** returned raw driver rows cast
  straight to `TakeHit[]`; int8 columns (`take_id`, `page_id`) arrive as
  native BigInt from the pg driver, so MCP `takes_search` crashed the
  serializer the moment a query actually matched (no match → `[]` → reads as
  intermittent). `takes_list` was immune because it maps through
  `takeRowToTake`. Added `takeHitRowToHit` (mirrors `takeRowToTake`,
  `Number()` per the `TakeHit` numeric contract) and routed all four sites —
  Postgres + PGLite, keyword + vector — through it for engine parity.
- **`gbrain call`** bypasses cli.ts's op-output normalizer entirely — its own
  bare `JSON.stringify` at the command exit crashed on any op returning an
  int8 column (the calibration repro only survived because `getLatestProfile`
  now coerces at the source). Reused `bigintToStringReplacer` at that exit
  (static import is cycle-safe: cli.ts only dynamic-imports call.ts, and its
  main() sits behind `import.meta.main`).

Both fixes are pinned by mutation-verified regression tests: reverting either
one fails a test (e2e postgres searchTakes/searchTakesVector type +
serializability assertions; a `runCall` stub-engine wiring test). Without the
pins the whole suite stayed green under revert — PGLite returns int8 as plain
`number`, so only the real-driver e2e suite can produce the BigInt input, and
`searchTakesVector` previously had zero invocations repo-wide.

### calibration_profile + voice-gate model resolution — Fixes #2451
Both hardcoded bare model ids (`claude-sonnet-4-6`, `claude-haiku-4-5`).
`gateway.chat()` → `parseModelId` throws "missing a provider prefix" for a bare
id, crashing the phase on its first LLM call. Routed both through
`TIER_DEFAULTS` (reasoning / utility) so the budget estimate, gateway call,
voice gate, and stored `model_id` share one resolved `provider:model`.

### `gbrain calibration` reachability — Fixes #2035
`calibration` was missing from `CLI_ONLY`, so dispatch fell through to
"Unknown command" despite the `case 'calibration'` handler. Added it to the set.

### takes reads honor source / federated_read scope — Addresses #2200
`takes_list` / `takes_search` / `takes_scorecard` / `takes_calibration` applied
no source scope, while page reads and `get_calibration_profile` route through
`sourceScopeOpts` — so a `federated_read` client saw takes outside its granted
sources. `takes` has no `source_id` column, so it's scoped through the take's
`page.source_id` (JOIN for list/search, EXISTS for scorecard/curve). Threaded
`sourceScopeOpts(ctx)` into the four handlers and filtered in **both** the
Postgres and PGLite engines. The holder allow-list stays an orthogonal,
unchanged filter. (Partial fix for the broader #2200 sweep — the takes slice;
see the issue comment.)

## Tests

New / extended (all green under `bun test`):
- `test/cli-bigint-normalize.test.ts` — bigint replacer + `CLI_ONLY`
  reachability + `takeHitRowToHit` BigInt coercion + `runCall` output-exit
  wiring (stub engine, bigint-carrying op result)
- `test/e2e/takes-postgres.test.ts` — searchTakes hits assert numeric
  take_id/page_id + JSON-serializability on the real pg driver; new
  searchTakesVector test with the same pins (the file header already claimed
  vector coverage; now it exists)
- `test/calibration-profile.test.ts` — default model resolves to a
  provider-prefixed `model_id`; tier defaults parse via `parseModelId`
- `test/takes-source-scope.test.ts` — PGLite: list/search (JOIN) +
  scorecard/curve (EXISTS) source filtering, and source × holder-allow-list
  composition

Run locally:
- `bun test` (cli-bigint-normalize, takes-source-scope, calibration-profile,
  calibration-cli, voice-gate, takes-engine) → **105 pass, 0 fail**; after the
  2026-07-12 update, all 29 takes-related files → **357 pass, 0 fail**
- `test/e2e/takes-postgres.test.ts` against a real Postgres 17 + pgvector
  scratch DB → **23 pass, 0 fail** (2026-07-12 update)
- `bun run typecheck` → clean
- `bun run verify` (30-check pre-push gate, incl. `check:source-id-projection`,
  `check:operations-filter-bypass`, privacy, typecheck) → **30/30 green**
  (re-run after the update)

Skipped (with reasons):
- Full `bun run test` (3700+ unit shards) and the rest of `test/e2e/*` were
  not run locally. The targeted suite + the 30-check verify gate cover the
  changed code; the takes source-scope behavior is exercised on **real
  PGLite**, and the BigInt coercion on **real Postgres** (takes e2e file).

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -43,8 +43,18 @@ for (const op of operations) {
   }
 }
 
+/**
+ * JSON replacer: `bigint` → string, matching the postgres.js wire shape (int8
+ * comes back as a string on the routed path). Lets the local-engine output
+ * normalizer round-trip bigint columns (e.g. a `BIGSERIAL` `id`) instead of
+ * throwing `TypeError: Do not know how to serialize a BigInt`.
+ */
+export function bigintToStringReplacer(_key: string, value: unknown): unknown {
+  return typeof value === 'bigint' ? value.toString() : value;
+}
+
 // CLI-only commands that bypass the operation layer
-const CLI_ONLY = new Set(['init', 'reinit-pglite', 'upgrade', 'post-upgrade', 'check-update', 'integrations', 'publish', 'check-backlinks', 'lint', 'report', 'import', 'export', 'files', 'embed', 'serve', 'call', 'config', 'doctor', 'migrate', 'eval', 'sync', 'extract', 'extract-conversation-facts', 'enrich', 'features', 'autopilot', 'graph-query', 'jobs', 'agent', 'apply-migrations', 'skillpack-check', 'skillpack', 'resolvers', 'integrity', 'repair-jsonb', 'orphans', 'sources', 'mounts', 'dream', 'check-resolvable', 'routing-eval', 'skillify', 'smoke-test', 'providers', 'storage', 'repos', 'code-def', 'code-refs', 'reindex', 'reindex-code', 'reindex-frontmatter', 'code-callers', 'code-callees', 'frontmatter', 'auth', 'friction', 'claw-test', 'book-mirror', 'takes', 'think', 'salience', 'anomalies', 'transcripts', 'models', 'remote', 'recall', 'forget', 'edges-backfill', 'cache', 'ze-switch', 'founder', 'brainstorm', 'lsd', 'schema', 'capture', 'onboard', 'conversation-parser', 'status', 'connect', 'skillopt', 'quarantine', 'self-upgrade', 'advisor', 'watch']);
+export const CLI_ONLY = new Set(['init', 'reinit-pglite', 'upgrade', 'post-upgrade', 'check-update', 'integrations', 'publish', 'check-backlinks', 'lint', 'report', 'import', 'export', 'files', 'embed', 'serve', 'call', 'config', 'doctor', 'migrate', 'eval', 'sync', 'extract', 'extract-conversation-facts', 'enrich', 'features', 'autopilot', 'graph-query', 'jobs', 'agent', 'apply-migrations', 'skillpack-check', 'skillpack', 'resolvers', 'integrity', 'repair-jsonb', 'orphans', 'sources', 'mounts', 'dream', 'check-resolvable', 'routing-eval', 'skillify', 'smoke-test', 'providers', 'storage', 'repos', 'code-def', 'code-refs', 'reindex', 'reindex-code', 'reindex-frontmatter', 'code-callers', 'code-callees', 'frontmatter', 'auth', 'friction', 'claw-test', 'book-mirror', 'takes', 'think', 'salience', 'anomalies', 'calibration', 'transcripts', 'models', 'remote', 'recall', 'forget', 'edges-backfill', 'cache', 'ze-switch', 'founder', 'brainstorm', 'lsd', 'schema', 'capture', 'onboard', 'conversation-parser', 'status', 'connect', 'skillopt', 'quarantine', 'self-upgrade', 'advisor', 'watch']);
 // CLI-only commands whose handlers print their own --help text. These are
 // excluded from the generic short-circuit so detailed per-command and
 // per-subcommand usage stays reachable.
@@ -442,7 +452,7 @@ async function main() {
     // path's return value so renderers see the same shape they'd see on the
     // routed path. Date → ISO string; bigint → string (postgres.js shape);
     // Buffer → object. Microsecond-cost; eliminates a whole drift bug class.
-    const result = JSON.parse(JSON.stringify(rawResult));
+    const result = JSON.parse(JSON.stringify(rawResult, bigintToStringReplacer));
     const output = formatResult(op.name, result);
     if (output) process.stdout.write(output);
   } catch (e: unknown) {
--- a/src/commands/calibration.ts
+++ b/src/commands/calibration.ts
@@ -125,6 +125,7 @@ export interface RunCalibrationArgs {
   regenerate?: boolean;
   undoWave?: string;
   abReport?: boolean;
+  source?: string;
 }
 
 function parseArgs(args: string[]): { sub?: string; opts: RunCalibrationArgs } {
@@ -144,6 +145,7 @@ function parseArgs(args: string[]): { sub?: string; opts: RunCalibrationArgs } {
     else if (a === '--json') opts.json = true;
     else if (a === '--regenerate') opts.regenerate = true;
     else if (a === '--undo-wave') opts.undoWave = args[++i];
+    else if (a === '--source') opts.source = args[++i];
   }
   return { sub, opts };
 }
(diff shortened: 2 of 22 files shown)


time-attack:
Superseded by #2892, which just merged: your four fixes rebased onto current master with conflicts resolved (the think/gather slice was already on master via #2739, so it was dropped; everything else ported intact, 391 tests green). Full credit preserved via Co-authored-by. Thank you for the excellent work — the calibration CLI routing and takes source-scoping were both verified real and are now shipped.

PR #2455, opened 2026-06-27
feat(ai): add reranker touchpoint to LiteLLM proxy recipe

ozp:
## What

Adds a `reranker` touchpoint to the LiteLLM proxy recipe (`src/core/ai/recipes/litellm-proxy.ts`).

## Why

LiteLLM exposes `/v1/rerank`, which normalizes Cohere / Voyage / Jina / etc. to the **same wire shape `gateway.rerank()` already speaks** (the ZeroEntropy / llama.cpp contract):

```
{ model, query, documents, top_n }  →  { results: [{ index, relevance_score }] }
```

So any rerank model a user registers in their LiteLLM config becomes reachable via:

```
gbrain config set search.reranker.model litellm:<model-name>
```

with **no request/response adapter** — the same way embeddings already ride the proxy. Before this, `litellm:` could serve embeddings but not rerank, so users on a LiteLLM-centric stack had to add a separate ZeroEntropy/Voyage credential just for reranking.

## Shape

Mirrors the existing `llama-server-reranker` recipe (the other user-provided/proxy reranker):

- `models: []` — proxy-defined; user sets `search.reranker.model` explicitly.
- `path: '/v1/rerank'` — `base_url_default` is `http://localhost:4000` (no `/v1`), so concatenation yields `http://localhost:4000/v1/rerank` (single prefix, asserted in the test).
- `cost_per_1m_tokens_usd: 0` — the proxied backend bills; gbrain can't know the rate from here, and this keeps `--max-cost` callers from hard-failing at the recipe layer.

## Testing

- New `test/ai/recipe-litellm-reranker.test.ts` (touchpoint present, `/v1/rerank` path, base_url concatenation has no doubled `/v1/v1/`, `models: []`).
- `recipe-llama-server-reranker`, `recipes-existing-regression`, and `rerank` suites pass; `tsc --noEmit` clean.
- Verified end-to-end against a live LiteLLM proxy fronting Cohere `rerank-v4.0-fast`: `{model, query, documents, top_n}` → `{results:[{index, relevance_score}]}`, consumed without modification by `gateway.rerank()`.

### DIFF
--- a/src/core/ai/recipes/litellm-proxy.ts
+++ b/src/core/ai/recipes/litellm-proxy.ts
@@ -40,6 +40,30 @@ export const litellmProxy: Recipe = {
       // mismatched-dim responses pre-storage).
       supports_multimodal: true,
     },
+    // LiteLLM exposes a rerank endpoint at /v1/rerank that normalizes Cohere /
+    // Voyage / Jina / etc. to the same wire shape gbrain's gateway.rerank()
+    // already speaks (the ZeroEntropy/llama.cpp contract):
+    //   { model, query, documents, top_n } → { results: [{ index, relevance_score }] }
+    // So any rerank model the user registers in their LiteLLM config is reachable
+    // via `gbrain config set search.reranker.model litellm:<model-name>` with no
+    // request/response adapter — same as embeddings ride the proxy today.
+    reranker: {
+      models: [], // user-provided; whatever rerank models the proxy serves
+      // No canonical default — the proxy defines its own model ids. The user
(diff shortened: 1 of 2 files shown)


time-attack:
Nice, tight patch, and the note about the wire shape already matching gateway.rerank() is a helpful detail. We're closing it on policy though: new vendor/proxy integration touchpoints are surface area we're not taking on, since each one becomes ours to keep working as the recipe layer evolves. Thanks for the end-to-end verification write-up regardless.

PR #2456, opened 2026-06-27
fix(serve): lazy-connect stdio MCP engine

### LINKED ISSUE #2091: serve (stdio): engine connects eagerly before the MCP initialize handshake — slow PGLite boot / lock contention kills the host connection at its 30s timeout, and shutdown racing an in-flight connect leaks the lock dir

eloe:
Fixes #2091. Rebases/replaces the currently-conflicting #2092 against current `master`.

## Problem

`gbrain serve` on stdio received an eagerly-connected engine. For PGLite, that means the MCP `initialize` handshake waits on full WASM boot plus single-writer lock acquisition. Under contention the lock wait alone can reach the host's 30s connect timeout, so clients see an opaque MCP timeout instead of an actionable lock error.

There is also a shutdown race: if the host disconnects while the engine connect is still in flight, the old cleanup path could call `engine.disconnect()` before the lock was acquired, then the in-flight connect could complete and leave the lock behind.

## Fix

- Add `src/core/engine-provider.ts`, a lazy engine provider for stdio serve.
- Dispatch `gbrain serve` before `connectEngine()` in `cli.ts` and pass a lazy connector.
- Keep `--http` eager, because HTTP has no stdio initialize-handshake deadline.
- Have stdio `runServe` start connection in the background and pass the provider to the MCP server.
- Have `tools/call` await `provider.ensure()` and return a structured `engine_unavailable` tool error on connect failure; failed connects are retried on the next call.
- Have shutdown call `provider.disconnect()`, which awaits any in-flight connect before disconnecting the engine, closing the lock-leak race.
- Preserve Retrieval Reflex IPC by binding it only after the lazy engine becomes ready; includes current suppression + delivery logging behavior from master.

## Tests

- `bun test test/serve-lazy-engine-connect.test.ts test/serve-stdio-lifecycle.test.ts`
- `bun test test/e2e/serve-stdio-roundtrip.test.ts`
- `bun run typecheck`
- `bun run build:llms`

Note: `bun run verify` was attempted, but this checkout's parallel verify wrapper killed all 30 subchecks with rc=143 after ~13s even though the visible checks were clean up to termination. The targeted tests and typecheck above passed.


eloe:
Local carry test on top of current `master` (v0.42.53.0) passed the targeted suite:

- `bun test test/serve-lazy-engine-connect.test.ts test/serve-stdio-lifecycle.test.ts test/e2e/serve-stdio-roundtrip.test.ts` → 40 pass / 0 fail
- `bun run typecheck` → pass

Important operational finding: this PR fixes the stdio MCP initialize/connect-timeout class, but it does **not** make a long-running PGLite `gbrain serve` safe to keep enabled beside normal local CLI commands. After locally cherry-picking this PR and re-enabling a LaunchAgent running `MCP_STDIO=1 gbrain serve`, concurrent `gbrain status`, `gbrain search`, `gbrain doctor --json`, and `gbrain query` calls still timed out while the serve daemon held the embedded PGLite connection. Disabling the LaunchAgent restored those commands immediately.

So #2456 is still useful and correct for #2091, but durable Retrieval Reflex IPC on a PGLite brain appears to need an additional design change: either route local CLI/reflex queries through the already-running serve process, avoid holding the PGLite writer connection in idle serve, or otherwise make the single-connection ownership explicit.

eloe:
Updated this PR to cover the #2458 durable-PGLite-IPC coexistence failure too.

New behavior:

- `MCP_STDIO=1 gbrain serve` no longer warms/connects the PGLite engine just by starting.
- The Retrieval Reflex resolve socket binds immediately without requiring engine readiness.
- A resolve IPC request connects only for the request, then releases the engine again in durable stdio mode.
- Normal MCP tool calls still lazy-connect and retain the structured `engine_unavailable` behavior from the original #2456 fix.

Local validation on macOS + PGLite + LaunchAgent:

- started `com.gbrain.serve-ipc` with `MCP_STDIO=1`
- verified `.gbrain-resolve.sock` exists while serve is idle
- ran concurrent bounded CLI calls while serve was running:

```text
gbrain status                  OK (~0.57s)
gbrain search Fantasy15         OK (~1.44s)
gbrain doctor --json            OK (~3.83s)
gbrain query ...                OK (~2.72s)
```

Then after daily maintenance stopped/restarted the service, repeated smoke:

```text
gbrain status       OK (~0.53s)
gbrain search       OK (~0.60s)
gbrain doctor       OK (~0.78s)
```

Targeted tests:

```text
bun test test/serve-lazy-engine-connect.test.ts test/serve-stdio-lifecycle.test.ts test/e2e/serve-stdio-roundtrip.test.ts
41 pass / 0 fail

bun run typecheck
pass
```

This now addresses both classes:

- #2091: MCP initialize no longer waits on engine connect.
- #2458: durable PGLite Reflex IPC no longer monopolizes the DB while idle.

time-attack:
Closing as duplicate of #2092.

Competing PR for lazy-connect / MCP initialize-before-engine. Keep #2092.

PR #2457, opened 2026-06-27
v0.42.54.0 fix(security): views enforce caller RLS via security_invoker

jdeweedata:
## What

Migration **v120 `page_links_view_security_invoker`** defines GBrain's schema view `WITH (security_invoker = on)`, so it enforces the **querying role's** row-level security instead of the view owner's.

GBrain enables RLS on its core tables, but Postgres creates views SECURITY DEFINER by default — so a view over an RLS-protected table reads it as the view's owner and skips the caller's policy. This flips it to `security_invoker`, closing that gap. A managed-Postgres database linter flagged it (`0010_security_definer_view`).

## Why it's safe / scoped

- **No effect on gbrain itself** — it connects as the table-owning role, which bypasses RLS regardless. This only constrains untrusted roles on managed Postgres brains.
- **Embedded (PGLite) brains unaffected** — a single trusted connection with no separate roles; the option is a harmless no-op there.
- **Born-secure on fresh installs, auto-applied on upgrade.** Both `schema.sql` and `pglite-schema.ts` seed `version='1'`, so every fresh brain runs the full migration array including v120; existing brains pick it up on the next `initSchema()`. One migration covers all four cases (fresh/existing × Postgres/PGLite).
- View projection is unchanged (`id, from_page_id, to_page_id`).

## Test plan

- `bun run typecheck` — clean
- PGLite (PG18.3) accepts the option and stores the reloption (verified directly)
- Full migration chain applies `[120] ✓` on both engines (`test/migrate.test.ts`)
- **New** `test/migrations-v120.test.ts` — 5 pass / 0 fail: asserts the view ends up `security_invoker=on` after `initSchema()` with its column shape intact

## Version

`0.42.53.0 → 0.42.54.0` (VERSION, package.json, CHANGELOG all in sync). No CLAUDE.md/docs edits, so no `build:llms` change.


harrisali0101:
Composes naturally with #2387 — your view-level `security_invoker` switch ensures policies fire through views; #2387 adds the session-scoped value (`app.scopes`) those policies key off. Two layers, one defense-in-depth posture: views no longer launder the RLS check, and the policies themselves have a real per-request scope to compare against.

Linked from #2387's Related work for cross-visibility. Happy to align on whichever lands first.

time-attack:
Closing as superseded: the security_invoker fix for page_links this PR proposes already shipped on master as migration v120 in src/core/migrate.ts, via commit dde1132a (v0.42.55.0, #2399). Master's version also broadens the fix to search_path hardening on trigger functions. Thanks for flagging - it's covered.

PR #2460, opened 2026-06-27
fix(markdown): prefer body H1 for missing titles

### LINKED ISSUE #2446: Importer fabricates junk titles from slug when frontmatter lacks title: — should fall back to body H1 (markdown.ts:138)
# Importer fabricates junk titles from the slug when frontmatter has no `title:` — should fall back to the body H1

**Version:** gbrain `0.42.52.0` (Postgres engine, DB-only brain)
**Component:** `src/core/markdown.ts` — `parseMarkdown` title selection (~line 138) + `inferTitle` (~line 602)

## Problem

When an imported `.md` file has **no `title:` in frontmatter**, `parseMarkdown` derives the title from the **filename/slug** via `inferTitle`, ignoring the document's H1. For slugged imports this produces junk titles.

Concrete case: a contacts importer writes 3,439 person pages with slugs like `contact-20170928-5-john-defalco` and a correct H1 `# John DeFalco`, but no frontmatter `title:`. Result:

- stored `title` = `"Contact 20170928 5 John Defalco"` (slug humanized — note the date/id tokens leak in, and casing is lost: `Defalco`, not `DeFalco`)
- `compiled_truth` H1 = `# John DeFalco` (correct, available)

Because the importer rewrites `title` on **every** import (re-`putPage`), any downstream correction (e.g. a manual `UPDATE pages SET title`) is reverted on the next run. The only durable fix is at the importer.

## Root cause

`src/core/markdown.ts` ~line 138:

```ts
const title = coerceFrontmatterString(frontmatter.title).trim() || inferTitle(filePath);
```

`inferTitle` (~line 602) just humanizes the filename:

```ts
return filename.replace(/[-_]/g, ' ').replace(/\b\w/g, c => c.toUpperCase());
// contact-20170928-5-john-defalco → "Contact 20170928 5 John Defalco"
```

It never consults the body H1, which is already in scope (`body` at ~line 124, `const body = parsed?.content ?? content;`).

## Suggested fix

Prefer the body H1 between frontmatter and the slug fallback:

```diff
- const title = coerceFrontmatterString(frontmatter.title).trim() || inferTitle(filePath);
+ const h1 = /^#\s+(.+?)\s*$/m.exec(body)?.[1]?.trim();
+ const title = coerceFrontmatterString(frontmatter.title).trim() || h1 || inferTitle(filePath);
```

Precedence preserved: explicit frontmatter `title:` still wins (no regression for files that set it), the H1 is used when present, and `inferTitle(filePath)` remains the last resort for files with no H1. This yields correctly-cased `John DeFalco` and is stable across re-imports.

Verified locally: applying this and re-importing flipped 3,439/3,439 person titles from the junk form to the real H1 name (correct casing), with no regression to pages that carry a frontmatter title.

## Impact

Any slug-based import without frontmatter titles (contacts, calendar, etc.) gets junk titles, which also degrades anything keyed on `title` — notably the `extract links --by-mention` gazetteer, which buckets by the title's first token (junk titles collapse all entities into one useless bucket → 0 mention links).

dpersek:
## Summary
- Make `parseMarkdown` use the first body H1 as the title when frontmatter has no usable `title`
- Preserve explicit frontmatter `title` precedence
- Keep filename/path title inference as the final fallback

Fixes #2446

## Verification
- `bun test test/markdown.test.ts`
- `bun run verify`

Note: full `bun run test` was attempted locally but exceeded the tool timeout while still progressing; no test failure was observed.

### DIFF
--- a/src/core/markdown.ts
+++ b/src/core/markdown.ts
@@ -135,7 +135,9 @@ export function parseMarkdown(
   const type = coerceFrontmatterString(frontmatter.type) || (
     opts?.activePack ? inferTypeFromPack(filePath, opts.activePack) : inferType(filePath)
   );
-  const title = coerceFrontmatterString(frontmatter.title).trim() || inferTitle(filePath);
+  const title = coerceFrontmatterString(frontmatter.title).trim()
+    || inferBodyH1(compiled_truth)
+    || inferTitle(filePath);
   const tags = extractTags(frontmatter);
   const slug = coerceFrontmatterString(frontmatter.slug) || inferSlug(filePath);
 
@@ -597,6 +599,11 @@ function inferTypeWithPrefixes(
     }
   }
   return 'concept';
+}
+
+function inferBodyH1(body: string): string {
+  const h1Match = body.match(/^#\s+(.+)$/m);
+  return h1Match?.[1]?.trim() || '';
 }
 
 function inferTitle(filePath?: string): string {
--- a/test/markdown.test.ts
+++ b/test/markdown.test.ts
@@ -88,6 +88,31 @@ Content
     expect(parsed.slug).toBe('concepts/do-things-that-dont-scale');
   });
 
+  test('prefers body H1 over inferred filename when frontmatter title is missing', () => {
+    const md = `---
+type: person
+---
+# Alice Example
+
+Contact notes.
+`;
+    const parsed = parseMarkdown(md, 'contacts/contact-20170928-5-alice-example.md');
+    expect(parsed.title).toBe('Alice Example');
+  });
+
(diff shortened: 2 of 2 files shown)


time-attack:
Thanks @dpersek! 🙏

What this does: This PR fixes issue #2446 by making the title fallback chain: frontmatter title → body H1 → filename. It uses a simple regex to extract the first heading, with 2 test cases.

This overlaps with #2495, which we're carrying forward, so I'm closing this in its favor to keep the review queue focused. If yours covers a case #2495 misses, reopen with that detail and we'll reconcile them.

Issue #2465, opened 2026-06-28
Owner-identity fragmentation: brain owner spans self / brain / people-<owner> holder strings


devty:
The brain owner is represented by multiple holder strings — `self` (consolidate facts→takes writer), `brain` (propose_takes when the author asserts a claim), and `people/<owner>` + variants (extraction that names the owner). #2464 makes calibration / think / doctor default to the canonical `self`, but historical owner takes under `brain`/`people-<owner>` are not folded into the default profile.

Full unification is owner-identity entity-resolution — normalize the owner's holder strings, or union them at read time for the owner profile. Tracked separately from #2464; the `resolveOwnerHolder` comment (`src/core/owner-holder.ts`) points here.

Per-take attribution for *other* people (e.g. `people/george`) is unaffected and correct — this is purely about the owner appearing under several strings.

time-attack:
Closing after maintainer review — decision: not accepting this. The scattering stays; questions about the owner keep returning incomplete answers, and the problem grows as more data comes in.

Thank you for the contribution and the thought behind it; this was reviewed carefully rather than dismissed.

PR #2424, opened 2026-06-25
feat: integrate native GBrain review and evidence fixes

Mr-B-1:
## Summary
- Bundles the six native GBrain branches for production-style integration validation.
- Includes review-first take proposals, role-heading conversation parsing, since-date repair, hybrid evidence grading, atom concept labels, and enrich-thin blocked-state classification.
- Split draft PRs are also open for focused upstream review.

## Integration verification
- bun run typecheck
- bun test --timeout 30000 for the 14 focused files covering all six branches
- bun src/cli.ts eval conversation-parser test/fixtures/conversation-formats/all.jsonl --no-llm
- git diff --check
- integration diff scan for forbidden created_at / LOCAL PATCH markers

PR #2467, opened 2026-06-28
fix(calibration): resolve owner holder via config (default 'self'), fixes #2464

### LINKED ISSUE #2464: calibration_profile default holder 'garry' never matches the 'self' holder that consolidate writes → forecasting loop silently no-ops on every non-owner brain
## Summary

The takes/calibration pipeline writes and reads takes under **two different hardcoded holder values**, so the calibration loop can never see its own data on any brain that isn't the upstream owner's:

- The dream **`consolidate`** phase inserts takes with `holder: 'self'` — `src/core/cycle/phases/consolidate.ts:194` (`INSERT into takes(kind='fact', holder='self', …)`).
- The dream **`calibration_profile`** phase reads the scorecard with `holder ?? 'garry'` — `src/core/cycle/calibration-profile.ts:229`.

Both run in the *same cycle*. One writes `self`, the sibling reads `garry`, so `getScorecard({holder:'garry'})` returns `resolved: 0`, the phase takes its cold-brain branch, and you get:

```
✓ calibration_profile  holder=garry has only 0 resolved takes (need >=5 for a profile)
```

…even on a brain with many takes. `get_calibration_profile` then always returns `null`.

## Why this is a placeholder bug, not config

`'garry'` is hardcoded as a personal default across the calibration/forecasting surface:

- `src/core/cycle/calibration-profile.ts:98,229`
- `src/core/cycle/emotional-weight.ts:50` — `export const DEFAULT_USER_HOLDER = 'garry'`
- `src/core/think/index.ts:296`
- `src/commands/calibration.ts:161,243,248`
- `src/commands/doctor.ts:1206` — health check hardcodes `WHERE holder = 'garry'`, so the doctor surface is also blind on forks

A registered config key already exists for exactly this — `emotional_weight.user_holder` (`src/core/config.ts:884`), which `brainstorm/orchestrator.ts:626` honors (`config.emotional_weight?.user_holder ?? 'garry'`). The calibration phase, the `think` path, and `doctor` ignore it and hardcode the literal.

There's also a *third* convention: `extract-takes-from-pages.ts:128` defaults `holder ?? 'system'`. So three holder vocabularies (`self`, `garry`, `system`) coexist with no single source of truth.

## Reproduction

1. On any brain that isn't the upstream owner's, let `consolidate` run (writes `holder='self'` takes).
2. Run the `calibration_profile` phase (or `gbrain dream`).
3. Observe the `holder=garry has only 0 resolved takes` skip despite `gbrain takes-list` showing many takes with `holder='self'`.
4. `get_calibration_profile` → `null` forever.

## Expected

The phase that *reads* takes for calibration should query the same holder convention the pipeline *writes*. The profile should build once ≥5 resolved bets exist, regardless of whose brain it is.

## Proposed fix

Single source of truth for "the brain owner's holder," resolved in one place and consumed everywhere (calibration_profile, think, doctor, emotional-weight):

```
opts.holder ?? config.emotional_weight?.user_holder ?? brainIdentityHolder ?? 'self'
```

i.e. make `consolidate`'s `'self'` and the calibration default agree (either both `self`, or both resolved from `emotional_weight.user_holder` / brain identity), and route `doctor.ts:1206` + `think` through the same resolver instead of the bare literal.

## Related

Stacked behind #2079 (`takes list` slug footgun) and #2411 (`takes propose` promotion path missing) — all three sit between "proposals exist" and "a profile appears." Holder alignment is necessary even after those land. Model-resolution sibling fixes in #2451 / #2452. Verified against `master` v0.42.53.0.

devty:
## Summary
Closes #2464. The dream `consolidate` writer stamps owner takes `holder='self'`, but `calibration_profile` (and `think`, the calibration CLI/op, `doctor`, `emotional-weight`, brainstorm cold-start, and the serve-http admin calibration routes) defaulted to a hardcoded `'garry'` — so on any non-owner brain the nightly cycle (and the admin UI) builds/reads an empty profile forever.

This adds one resolver — `src/core/owner-holder.ts`:

```ts
export const DEFAULT_OWNER_HOLDER = 'self';
export function resolveOwnerHolder(opts: { override?: string | null; configValue?: string | null }): string {
  return opts.override ?? opts.configValue ?? DEFAULT_OWNER_HOLDER;
}
```

Precedence: explicit override > `emotional_weight.user_holder` config > `'self'`. Every default-resolution site is routed through it, fed by `engine.getConfig('emotional_weight.user_holder')`:

- `src/core/cycle/calibration-profile.ts` (nightly phase)
- `src/commands/calibration.ts` (CLI + `get_calibration_profile` op)
- `src/core/think/index.ts` (calibration block)
- `src/core/cycle/emotional-weight.ts` (`DEFAULT_USER_HOLDER` now delegates to `DEFAULT_OWNER_HOLDER`)
- `src/commands/doctor.ts` (`calibration_freshness`, parameterized `WHERE holder = $1`)
- `src/core/brainstorm/orchestrator.ts` (cold-start calibration context)
- `src/commands/serve-http.ts` (three admin calibration routes)
- `src/commands/takes.ts` (`takes resolve --by` default)
- `src/core/operations.ts` (op param description)

Per-take attribution and the `consolidate` writer are untouched — `people/<other>` takes (e.g. `people/george`) are unaffected.

## Notes
- Default `'self'` matches the `consolidate` writer + `docs/takes-vs-facts.md`. On the maintainer's own brain, owner takes may also live under `self`/`brain`/`people/<owner>`; the config key allows an override, and the broader owner-identity unification is tracked in #2465.
- Relationship to #2424 (draft): it changes the phase default to `'system'`, a *different* convention that matches `extract-takes-from-pages.ts` rather than the `consolidate` writer. This PR aligns the default with `consolidate` via a config resolver; if #2424 lands first the two reconcile to the resolver.
- Complementary to #2452 (model resolution / source-scoped takes reads); no overlap on these lines.

## Tests
- `test/owner-holder.test.ts` (resolver precedence: override > config > `'self'`)
- `test/calibration-profile.test.ts` (phase resolves to `self`; configured holder honored)
- `test/emotional-weight.test.ts` (`DEFAULT_USER_HOLDER === 'self'`)
- `bun run typecheck` clean; touched-area tests green (41/41). Full DB-backed suite deferred to CI.


devty:
Refreshed onto current master (v0.42.58.0) — merged cleanly, no conflicts.

Also fixed test debt this PR carried but that no CI run ever surfaced: the fix adds an `engine.getConfig('emotional_weight.user_holder')` read to `getCalibrationProfileOp` and `checkCalibrationFreshness`, but their mock engines only stubbed `executeRaw`. `getConfig` was `undefined`, so the MCP op threw and the doctor check swallowed the TypeError into a spurious `warn`. Both mocks now stub `getConfig() → null` (key unset → resolver falls back to the canonical `self`), and the three omitted-holder default tests updated `garry`→`self` to match the intended behavior.

Verified locally: typecheck clean; 148 calibration/holder/think/emotional-weight unit tests pass (0 fail). Ready for review.

time-attack:
Superseded with credit by #3077 (just merged) — your change landed rebased onto current master with authorship preserved via Co-authored-by. Thank you!

PR #2474, opened 2026-06-28
docs: add retrieval routing hints to search/query help text

### LINKED ISSUE #2416: Concept/landscape queries silently default to keyword search instead of hybrid (`query`) — no affordance steers agents to semantic retrieval
## Summary

gbrain exposes two retrieval commands as peers:

- `search <query>` — keyword search (tsvector)
- `query <question>` / `ask` — hybrid search (RRF + expansion, embeddings)

For an agent driving gbrain, nothing in the command surface or docs signals *which* to use for *which kind* of question. The result: agents reach for keyword search by default (it's the first-listed, simplest mental model), and only escalate to hybrid `query` when keyword "comes back thin." That fallback heuristic is the bug.

## The failure mode

A keyword pass for a fuzzy *concept* request ("all the X that do Y", "the landscape of Z", "anything matching <abstract category>") returns *some* results — so it never looks "thin," so the hybrid fallback never fires. But keyword recall only covers documents that happen to use the exact tokens the agent guessed. Anything described in synonym/outcome terms is silently omitted. The agent sees a populated result set and treats "got results" as "got the complete answer."

Concrete recall failure we hit: a record whose text contained a two-word spelling (`"off shore"`) was missed by a `%offshore%` keyword filter (one word). A single whitespace dropped a real, relevant hit. Hybrid/semantic retrieval surfaced it immediately. This generalizes: keyword cannot fail loudly, so completeness bugs are invisible at the call site.

## Why this is a gbrain-level issue (not just a caller mistake)

The tooling actively trains the wrong default:

1. **`search` is listed first** and reads as the primary verb; `query` reads as an advanced/optional alternative.
2. **No routing guidance** ships with either command. Neither help text says "use hybrid for concept/landscape/exhaustive-set questions; use keyword for known exact tokens / structured fields."
3. **The fallback framing is backwards.** The implicit rule agents inherit is "keyword first, semantic if thin." For concept questions the correct rule is the inverse: **semantic-first for recall completeness, keyword as a complementary precision/known-token pass.** A nonzero keyword count must NOT be treated as a completeness signal.

## Requested changes

1. **Help-text routing hints** on both `search` and `query`: one line each stating when each wins (keyword = exact known token / structured field / historical exact-match; hybrid = concept, synonym, landscape, "all/every/anything matching", recent-recall).
2. **A `--why`/decision note or a single recommended entry point** (e.g. make `query` the documented default for natural-language questions, and position `search` as the exact-token tool). Reduce the chance an agent picks keyword for a concept question.
3. **Docs:** add an explicit "choosing a retrieval mode" section, including the anti-pattern: *"a populated keyword result set is not proof of completeness."*
4. **Optional but ideal:** a heuristic in the query path that detects concept-shaped natural-language input (no exact identifiers, fuzzy quantifier like "all/every/anything/landscape") and either auto-runs hybrid or emits a hint suggesting `query` over `search`.

## Acceptance

- `search --help` and `query --help` each state when to use them and when not to.
- Docs contain a retrieval-mode decision guide + the "keyword count != completeness" anti-pattern.
- (If implemented) concept-shaped input to `search` surfaces a suggestion to use hybrid `query`.

## Impact

This is a silent-recall-loss class bug: the system returns plausible-but-incomplete answers for exactly the high-value question type (landscape/sector/exhaustive-set queries) and gives no signal that anything was missed.

mvanhorn:
## Summary
Adds retrieval routing hints to the `search` and `query` command help text so users pick the right tool: `search` for exact known tokens and structured-field lookups, `query` for concept, synonym, landscape, and recall-completeness questions. Each command's help now cross-references the other.

Adds a help-discoverability test asserting both hints render.

Fixes #2416

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -21,6 +21,17 @@ for (const op of operations) {
 // CLI-only commands that bypass the operation layer
 const CLI_ONLY = new Set(['init', 'upgrade', 'check-update', 'import', 'export', 'files', 'embed', 'serve', 'call', 'config', 'doctor']);
 
+const RETRIEVAL_ROUTING_HINTS: Record<string, string[]> = {
+  search: [
+    'Use for exact known tokens, structured fields, and historical exact-match lookups.',
+    'For concept, synonym, landscape, all matching/every matching/anything matching, or recall-completeness questions, use gbrain query.',
+  ],
+  query: [
+    'Use for concept, synonym, landscape, all matching/every matching/anything matching, and recall-completeness questions.',
+    'Use gbrain search for exact known tokens, structured fields, or historical exact-match lookups.',
+  ],
+};
+
 async function main() {
   const args = process.argv.slice(2);
   const command = args[0];
@@ -305,6 +316,14 @@ function printOpHelp(op: Operation) {
   const name = op.cliHints?.name || op.name;
   console.log(`Usage: gbrain ${name} ${positional} [options]\n`);
   console.log(op.description + '\n');
+  const routingHints = RETRIEVAL_ROUTING_HINTS[name];
+  if (routingHints) {
+    console.log('When to use which:');
+    for (const hint of routingHints) {
+      console.log(`  ${hint}`);
+    }
+    console.log('');
+  }
   const entries = Object.entries(op.params);
   if (entries.length > 0) {
     console.log('Options:');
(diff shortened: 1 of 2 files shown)


time-attack:
Closing: the help text this adds is now inaccurate — search has been cheap-hybrid by default since v0.41.34.0, so steering concept queries away from it would mislead users. Thanks for aiming at discoverability.

PR #2475, opened 2026-06-29
Fix native local embedding setup blockers

SolomonK9:
## Summary

- canonicalize explicit import slugs before chunk upsert so mixed-case CLI slugs don't create page/chunk casing mismatches
- allow llama-server embedding recipes with explicit user-provided model ids (for local GGUF embedding servers)
- add regression coverage for both paths

## Why

Local embedding providers should work through normal source/config/install paths, not runtime patches. A SPAS rehearsal exposed two setup blockers:

1. `importFromContent()` passed a mixed-case explicit slug through chunking after `putPage()` canonicalized the page row.
2. `diagnoseEmbedding()` rejected `llama-server:<model-id>` even when the model id was explicit, because the recipe declares `models: []` with `user_provided_models: true`.

## Tests

- `bun test test/gateway-embed-model-override.test.ts test/embed-input-type-wire.serial.test.ts`
- `bun test test/import-content-slug-canonicalization.test.ts --timeout 30000`
- `bun run verify`

### DIFF
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -704,12 +704,19 @@ export function diagnoseEmbedding(modelOverride?: string): EmbeddingDiagnosis {
     };
   }
 
-  // Openai-compat recipes with empty models list require a user-provided model.
+  // OpenAI-compatible recipes with an empty static model list are valid when
+  // the caller supplied a concrete provider:model id. Pre-fix, llama-server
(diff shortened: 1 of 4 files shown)


SolomonK9:
Closing: Solomon wants official latest repo/setup path only, no fork/PR branch path for local embedding setup. We will benchmark/configure without relying on forked or runtime-patched GBrain.

PR #2478, opened 2026-06-29
feat(config): auto_link_allow_remote flag for trusted MCP callers

onchito-walks:
## Problem

`auto_link` and `auto_timeline` are unconditionally skipped for remote (MCP) callers at `operations.ts:950`:

```typescript
if (ctx.remote !== false && !trustedWorkspace) {
  autoLinks = { skipped: "remote" };
  autoTimeline = { skipped: "remote" };
}
```

The security rationale (prompt-injected link planting → search poisoning via backlink boost) is valid for multi-tenant deployments. But for single-user personal brains where the MCP client is fully trusted, this creates pages with zero graph edges and zero timeline entries — floating untethered content that degrades brain integrity scores and breaks graph traversal.

## Impact on MCP-primary users

Users whose main write surface is MCP (e.g., Hermes agent → GBrain) end up with:
- Every `put_page` → `auto_links: skipped (remote)`, `auto_timeline: skipped (remote)`
- Pages have zero graph edges (no wikilinks resolved, no typed links)
- Pages have zero timeline entries (frontmatter dates not promoted)
- Brain integrity scores degrade as untethered pages accumulate
- `gbrain doctor` and `gbrain health` flag the damage but can not auto-repair without a full extract cycle

## Current workaround (external)

The community workaround for this gap is an explicit post-write metadata pass after every MCP `put_page`:

1. `add_timeline_entry` for each dated fact
2. `add_link` for each relationship
3. `add_tag` for frontmatter-derived tags

This is maintained as an agent-side skill/protocol rather than a code fix. For users running cron-based recovery, a 6-hour `auto_timeline` cron backfills from `created_at` timestamps as a safety net. But none of this is automatic at write time — it is reactive, not preventive.

## Fix

New config flag `auto_link_allow_remote` (default: `false`, preserving current behavior).

When set to `true`, auto-link and auto-timeline post-hooks fire even for remote (MCP) callers:

```bash
gbrain config set auto_link_allow_remote true
```

The gate becomes:

```typescript
if (ctx.remote !== false && !trustedWorkspace && !await isAutoLinkRemoteAllowed(ctx.engine)) {
```

`isAutoLinkRemoteAllowed()` follows the same pattern as `isAutoLinkEnabled()` / `isAutoTimelineEnabled()` in `link-extraction.ts`. Default-deny, opt-in only.

## Changes

- **`src/core/link-extraction.ts`**: Added `isAutoLinkRemoteAllowed()` function
- **`src/core/operations.ts`**: Added `isAutoLinkRemoteAllowed` to import; modified gate at L950
- **`src/core/config.ts`**: Registered `think.` config key prefix (drive-by fix for think.excerpt_length)

## Who this helps

- Single-user personal brains using MCP as primary write surface
- Hermes / OpenClaw / Claude Code users talking to local GBrain instances
- Anyone running `gbrain serve --http` with a trusted local MCP client

## Who is unaffected

- Multi-tenant deployments (default `false`)
- CLI users (`ctx.remote = false`, auto-link already fires)
- Autopilot daemon (`trustedWorkspace` already bypasses the gate)

## Testing

- [ ] Unit test: `isAutoLinkRemoteAllowed` returns false when config unset
- [ ] Unit test: `isAutoLinkRemoteAllowed` returns true when config = "true"
- [ ] Integration test: MCP `put_page` with `auto_link_allow_remote: true` produces auto-links/timeline

### DIFF
--- a/src/core/config.ts
+++ b/src/core/config.ts
@@ -944,6 +944,7 @@ export const KNOWN_CONFIG_KEY_PREFIXES: readonly string[] = [
   'mcp.',               // mcp.publish_skills, mcp.skills_dir (PR1 skill catalog)
   'autopilot.',         // autopilot.nightly_quality_probe.*, autopilot.auto_drain.* (#1685)
   'self_upgrade.',      // v0.42 self-upgrade (mode, quiet_hours, state)
+  'think.',             // think.excerpt_length, think.trajectory_enabled, etc.
 ];
 
 export function saveConfig(config: GBrainConfig): void {
--- a/src/core/link-extraction.ts
+++ b/src/core/link-extraction.ts
@@ -1201,6 +1201,26 @@ export async function isAutoTimelineEnabled(engine: BrainEngine): Promise<boolea
 }
 
 /**
+ * Read the auto_link_allow_remote config flag. Defaults to FALSE
+ * (remote callers are not trusted for auto-link/timeline).
+ *
+ * When TRUE: auto-link and auto-timeline post-hooks fire even for
+ * remote (MCP) callers. This is safe for single-user personal brains
+ * where the MCP client is trusted (e.g., Hermes agent talking to a
+ * local GBrain instance). Multi-tenant deployments should leave this
+ * FALSE — untrusted MCP clients can plant link-targeted pages to
+ * manipulate search ranking via the backlink boost.
+ *
+ * Same truthiness rules as the other config helpers.
+ */
+export async function isAutoLinkRemoteAllowed(engine: BrainEngine): Promise<boolean> {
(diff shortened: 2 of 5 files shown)


time-attack:
Thanks for the clear problem statement — the MCP-primary write pattern leaving pages without edges is a real ergonomics gap. I'm closing this because the flag's entire function is to switch off a trust boundary that exists on purpose: the remote skip guards against prompt-injected link planting boosting poisoned pages in search, and adding an opt-out doesn't remove that risk for whoever enables it. The explicit add_link / add_timeline_entry pass after an MCP write, or a local extract cycle, remains the supported path. The PR also registers a 'think.' config prefix beyond its stated scope, which is a separate change.

PR #2479, opened 2026-06-29
fix(think): surface page-vs-page conflicts on takes-empty brains; make retrieval breadth env-configurable

Adityaranjan-VeloxAI:
## Problem

`gbrain think`'s synthesis prompt only instructs the model to surface conflicts **between takes** ("If two takes contradict … surface BOTH"). On a brain whose knowledge lives in **pages** — takes table empty, the common personal-vault case — contradictory facts across pages (e.g. a price stated as `$2,000` in one doc and `$1,000` in another) are silently reconciled into a single answer instead of flagged. The conflict-detection feature is effectively dead when `takes = 0`.

## Changes (two commits)

**1. `fix(think)` — surface page-vs-page conflicts.** Broaden the Conflicts rule in `prompt.ts` to cover **pages and/or takes**: surface every fact where sources give divergent values, never reconcile a lower figure as a "discount" unless a source states it, enumerate conditional/variant values, suppress "sources agree" non-bullets, and tag each conflict `[HIGH]/[MEDIUM]/[LOW]` so the section is machine-parseable. No new behavior at default settings except that genuine page-vs-page conflicts now appear in the Conflicts section instead of being silently resolved.

**2. `feat(think)` — env-configurable retrieval breadth.** Retrieval breadth becomes tunable per-install via env vars, matching the existing `GBRAIN_*` knob convention (`GBRAIN_SYNC_*`, `GBRAIN_PACE_*`). **Defaults unchanged:**

| Env var | Default | Effect |
|---|---|---|
| `GBRAIN_THINK_GATHER_LIMIT` | 40 | pages gathered for synthesis |
| `GBRAIN_THINK_EXCERPT_LEN` | 600 | per-page excerpt chars (a too-narrow excerpt hides facts deep in a long page) |
| `GBRAIN_THINK_MAX_OUTPUT_TOKENS` | 4000 | synthesis output cap (conflict-dense answers truncate mid-JSON at 4000) |

## Why it matters

A drift/contradiction probe over a pages-only brain returned *zero* conflicts before this change despite real contradictions (stale prices, duplicated-then-diverged commitments); after it, the same probe surfaces them with severity. The env knobs let an install widen recall for that use case without forking.

## Notes for the maintainer

- I implemented the knobs as **env vars** to match the existing `GBRAIN_*` convention; happy to wire them as DB-plane `think.*` config keys instead if you'd prefer.
- Commits are split fix/feat so the bug fix can be cherry-picked independently of the knobs.
- Conventional-commit titles, no VERSION bump — left for your `/ship` flow.
- Verified by running `gbrain think` against a populated PGLite brain (compiles + runs; conflicts now surface and the Conflicts section is correctly omitted when there are none).

### DIFF
--- a/src/core/think/gather.ts
+++ b/src/core/think/gather.ts
@@ -24,7 +24,8 @@ export interface ThinkGatherOpts {
   question: string;
   /** Anchor entity slug. When set, the graph stream activates. */
   anchor?: string;
-  /** Soft cap on total results across all streams. Default 40. */
+  /** Soft cap on total results across all streams. Default 40, overridable via
+   *  GBRAIN_THINK_GATHER_LIMIT (widen for conflict/variant recall, e.g. drift probes). */
   gatherLimit?: number;
   /** Soft cap on take results. Default 30. */
   takesLimit?: number;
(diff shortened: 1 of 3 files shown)


time-attack:
Thanks for the detailed writeup — the observation that the Conflicts rule only mentions takes is a fair read of the prompt. I'm closing because both halves land outside what I'm merging: the prompt rewrite is a subjective change to synthesis behavior with no eval or regression test backing it, and the three GBRAIN_THINK_* knobs are new tuning surface for one install's breadth preference with defaults unchanged. Retrieval-quality changes to think need to be justified with measured results rather than prose, which is a bigger piece of work than this PR.

PR #2480, opened 2026-06-29
Fix takes search and think vector recall

sene1337:
## Summary

Fixes two take-retrieval gaps found while testing the rebuilt takes layer:

- `takes_search` could miss short terms and natural-language questions because it relied too narrowly on trigram similarity.
- `think` gathered keyword takes but skipped vector take recall because the local CLI and MCP operation paths did not pass an `embedQuestion` function into `runThink()`.

## Changes

- Adds full-text `websearch_to_tsquery` recall alongside pg_trgm for Postgres takes search.
- Adds all-term / any-term lexical fallbacks for natural-language questions.
- Preserves per-token holder allow-list filtering.
- Normalizes Postgres take IDs and scores into JSON-safe numbers.
- Wires `embedQuery` into both local `gbrain think` and MCP/operation `think`, while preserving graceful degradation if question embedding fails.
- Adds regression coverage for the lexical recall path.

## Verification

- `bun install --frozen-lockfile`
- `bun test test/takes-engine.test.ts test/think-pipeline.serial.test.ts test/takes-mcp-allowlist.serial.test.ts`
  - 68 pass / 0 fail

Local live smoke tests on the original checkout after the same fixes:

- `gbrain call takes_search '{"query":"Zeus","limit":3}'` returned Zeus Wallet takes.
- `gbrain call takes_search '{"query":"what does Brad believe about Bitcoin custody?","limit":3}'` returned takes.
- `gbrain think 'what does Brad believe about Bitcoin custody?' --json` reported `takesFromKeyword: 30` and `takesFromVector: 30`.
- `gbrain call think '{"question":"what does Brad believe about Bitcoin custody?","rounds":1}'` reported `takesFromKeyword: 30` and `takesFromVector: 30`.

### DIFF
--- a/src/commands/think.ts
+++ b/src/commands/think.ts
@@ -9,6 +9,7 @@ import type { BrainEngine } from '../core/engine.ts';
 import { runThink, persistSynthesis } from '../core/think/index.ts';
 import { loadConfig, isThinClient } from '../core/config.ts';
 import { callRemoteTool, unpackToolResult } from '../core/mcp-client.ts';
+import { embedQuery } from '../core/ai/gateway.ts';
 
 function flagValue(args: string[], name: string): string | undefined {
   const i = args.indexOf(name);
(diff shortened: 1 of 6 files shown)


darbsllim:
We are having a hard time getting takes & facts to work, more gbrain surgery this weekend.

time-attack:
Thanks — you're right that no runThink caller passes embedQuestion, so the takes vector lane never fires; I confirmed that on master. I'm closing anyway because nothing in the codebase ever writes takes.embedding either, so wiring embedQuery in today adds an embedding call to every think (CLI and MCP) while querying a column that is always NULL — it only pays off stacked on #2483, which I'm also not taking. The bundled takes_search SQL rewrite across both engines is recall tuning rather than a fix, and I'd need eval evidence before changing that ranking path.

PR #2483, opened 2026-06-29
Add stale take embedding writer

sene1337:
## Summary

- wires `gbrain embed --stale` to also scan active takes with `embedding IS NULL`
- implements `countStaleTakes`, `listStaleTakes({ sourceId, batchSize, afterTakeId })`, and `setTakeEmbedding` for both PGLite and Postgres
- writes take embeddings from the bare `claim`, stamps `embedded_at`, supports vector/halfvec Postgres columns, and keeps stale take work source-scoped
- loosens consolidate page resolution for older display-name-shaped `facts.entity_slug` values (exact slug first, then entity title / slugified paths) so eligible facts can actually promote to an entity page

## Why

The facts→takes consolidate lane can promote facts into `takes`, but new take rows were not picked up by `gbrain embed --stale`: the stale-take engine methods were effectively dead stubs and there was no `UPDATE takes SET embedding = ...` path. A consolidate run could therefore create invisible/retrieval-dead takes until someone manually re-embedded them out of band.

## Verification

- `bun test test/takes-engine.test.ts test/embed.serial.test.ts test/cycle-consolidate.test.ts` — pass (62/62)
- `bun test test/e2e/takes-postgres.test.ts` — pass/skip without `DATABASE_URL` (23 skipped)
- `scripts/check-jsonb-pattern.sh && scripts/check-test-isolation.sh && scripts/check-privacy.sh && scripts/check-test-real-names.sh` — pass
- `bun run build` — pass

Local note: `bun run verify` currently fails in this checkout on an unrelated OpenClaw SDK type mismatch (`buildMemorySystemPromptAddition` missing from `/Users/senemaro/node_modules/openclaw/dist/plugin-sdk/core`), not in the changed takes/consolidate code.

### DIFF
--- a/src/commands/embed.ts
+++ b/src/commands/embed.ts
@@ -128,6 +128,12 @@ export interface EmbedResult {
   total_chunks: number;
   /** Number of pages processed (whether or not they had stale chunks). */
   pages_processed: number;
+  /** Takes newly embedded in this run (separate from chunk count). */
+  embedded_takes?: number;
+  /** Takes that would be embedded in dry-run mode. */
+  would_embed_takes?: number;
+  /** Total stale takes considered by the stale-takes lane. */
+  total_takes?: number;
   /** True if this run was a dry-run. */
   dryRun: boolean;
   /**
@@ -644,7 +650,11 @@ async function embedAll(
     // D7: thread sourceId so `gbrain embed --stale --source X` actually scopes.
     // v0.41.18.0 (A13): thread batchSize/priority/catchUp into the stale path.
     // #1737: thread the external abort signal so the cycle embed phase bails.
-    return await embedAllStale(engine, sourceId, dryRun, result, onProgress, staleOpts, signature, signal);
+    await embedAllStale(engine, sourceId, dryRun, result, onProgress, staleOpts, signature, signal);
+    if (!isAborted(signal)) {
+      await embedAllStaleTakes(engine, sourceId, dryRun, result, staleOpts, signal);
+    }
+    return;
   }
 
   // --all path: pacer (no-op when off). E-1: lower the worker count to the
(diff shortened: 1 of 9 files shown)


darbsllim:
As part of our process to get takes & facts working with dreaming, we ran into a bug with the consolidate lane that this addresses - was verified with codex and Claude Code

sene1337:
Claude/default reviewed this in Brad's local closeout and approved the take-embedding writer, with one hardening request for the consolidate fuzzy resolver. I pushed that follow-up in `e80da103`.

Changes in the hardening commit:

- Adds per-run resolution-method counts to consolidate output: exact / title / suffix / fuzzy.
- Emits an audit line when the fuzzy rung is used, including `source_id`, `entity_slug`, chosen `page_slug`, score, and threshold.
- Includes sampled fuzzy resolutions in `details.fuzzy_resolution_samples` for reviewability.
- Makes the fuzzy threshold configurable via `cycle.consolidate.fuzzy_resolve_min_score`, defaulting to the prior `0.55` behavior.
- Adds tests that pin fuzzy audit output and prove the config threshold can disable loose fuzzy matches.

Verification run in the PR worktree after pushing:

- `bun test test/takes-engine.test.ts test/embed.serial.test.ts test/cycle-consolidate.test.ts` — 64 pass, 0 fail.
- `git diff --check` — pass.
- `bun run build` — pass.
- `scripts/check-jsonb-pattern.sh && scripts/check-test-isolation.sh && scripts/check-privacy.sh && scripts/check-test-real-names.sh` — pass.

The PR branch is now at `e80da1038e9c86d8a956cf1f8cef1a74dfee029a`.

time-attack:
Thanks for the thorough work — the diagnosis is accurate, takes really never get embedded anywhere on master today. I'm still closing this: activating that dormant lane is new feature surface (new engine methods on both backends, a new embed --stale scan, and a recurring per-take embedding bill) rather than a bug fix, and it arrives at 505 lines with an unrelated loosening of consolidate's entity_slug resolution folded in. Turning on a retrieval arm is a decision I'd want to make with eval evidence that it improves answers, not as a side effect of a plumbing PR. Appreciate the clean verification notes regardless.

PR #2487, opened 2026-06-29
fix(subagent): gateway tool-loop drops tool-results on resume → openai-compat dead-letter

ivandebot:
The gateway-native subagent loop (`runSubagentViaGateway` → `toolLoop`), used for non-Anthropic models, lacks the crash-recovery parity the legacy inline loop has:

1. It never persists the synthesized tool-result message (only the seed + assistant turns).
2. It has no resume reconciliation for a dangling assistant tool-call turn.

When a subagent job stalls (timeout / `max_stalled`) and is re-claimed, it reloads `[user, assistant(tool_calls), assistant(tool_calls), …]` with no tool-result messages between them. The first `chat()` then sends an assistant message whose `tool_calls` have no following tool results. Anthropic tolerates this; OpenAI-compatible providers (e.g. DeepSeek) reject it with `Tool results are missing for tool call X` → the job permanently dead-letters. Intermittent, since only stalled-then-resumed jobs hit it.

**Fix** (parity with the legacy loop):
- Add `ToolLoopOpts.onUserTurn`, fired before the in-memory push (write-before-side-effect). `runSubagentViaGateway` persists the tool-result user turn.
- Add replay reconciliation in `runSubagentViaGateway` mirroring the legacy reconciler: if the last persisted message is an assistant turn with dangling tool-call blocks, synthesize the tool-result user turn from `subagent_tool_executions` (complete → stored output, failed → stored error, pending/missing → re-dispatch idempotent tools, else throw), persist it, and feed it into `replayState`. Keyed by provider `tool_use_id`; `loadPriorToolsV2` now exposes it.

**Test**: `subagent-gateway-resume-reconciliation.test.ts` seeds the crash condition (assistant turn with parallel `tool_calls` + complete exec rows, no tool-result message). A validating stub transport throws the exact provider rejection on a dangling tool-call. Fails on baseline, passes with the fix, asserts reconciled results carry the real stored outputs.

Verified e2e against the DeepSeek API: without the fix a multi-turn synthesize subagent dead-letters 7/8 jobs; with it, 10/10 complete.

### DIFF
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -2908,6 +2908,14 @@ export interface ToolLoopOpts {
   ) => Promise<{ gbrainToolUseId: string }>;
   onToolCallComplete?: (gbrainToolUseId: string, output: unknown) => Promise<void>;
   onToolCallFailed?: (gbrainToolUseId: string, error: string) => Promise<void>;
+  /**
+   * Persist the synthesized tool-result user turn BEFORE it is appended to the
+   * in-memory conversation (write-before-side-effect, like onAssistantTurn). So
+   * a crash after the assistant turn but before the next LLM call reloads a
+   * well-formed [assistant(tool_calls), user(tool_results)] pair on replay —
+   * non-Anthropic providers reject a dangling assistant tool-call turn.
+   */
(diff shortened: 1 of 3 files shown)


time-attack:

PR #2488, opened 2026-06-29
fix(gateway): drop non-string text parts in toModelMessages (AI SDK v6 schema reject)

ivandebot:
A reasoning model (e.g. DeepSeek v4) can return a text content part whose `text` is `undefined`/`null`. The block type is `{type:'text', text:string}`, but the value arrives from the provider at runtime. On the next turn, `toModelMessages` emits it verbatim and AI SDK v6 rejects the request locally with `Invalid prompt: The messages do not match the ModelMessage[] schema` — failing the whole turn. (Confirmed: v6 rejects a text part with `undefined`/`null` text; empty-string text is fine.)

Intermittent, and in the subagent loop it *recovers on resume* because the replay path (`adaptContentBlocksToChatBlocks`) already drops non-string-text parts — so the rebuilt conversation is clean while the in-memory one was not.

**Fix**: `toModelMessages` (the documented `ChatMessage[]` → v6 `ModelMessage[]` converter) drops text parts whose `text` is not a string, mirroring the replay path. Covers every message source (live loop, replay, future). Empty content is valid for v6, so a turn left empty by the drop still passes.

**Tests** pin: non-string text dropped, valid siblings kept, empty-string kept.

### DIFF
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -2360,11 +2360,20 @@ export function toModelMessages(messages: ChatMessage[]): unknown[] {
     }
     return {
       role: m.role,
-      content: blocks.map((b) => {
-        if (b.type === 'text') return { type: 'text' as const, text: b.text };
-        if (b.type === 'tool-call') return { type: 'tool-call' as const, toolCallId: b.toolCallId, toolName: b.toolName, input: b.input };
-        return b;
-      }),
+      content: blocks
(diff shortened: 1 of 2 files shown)


time-attack:
This landed on master as part of the fix-wave A collector PR #2820 (commit 86962b24), which explicitly folds in this exact fix — `toModelMessages` now drops text blocks whose `text` isn't a string, matching the behavior proposed here. Closing as superseded by #2820; thanks for surfacing the bug and the clean repro.

PR #2492, opened 2026-06-30
fix(import): skip marked.lexer on fence-less pages to avoid transient-memory OOM on bulk import (#2437)

### LINKED ISSUE #2437: marked.lexer transient-memory amplification OOMs bulk import on fence-less pages
## Summary

`extractFencedChunks()` in `src/core/import-file.ts` runs `marked.lexer()` on **every** page body to pull out fenced code blocks. `marked.lexer` allocates transient memory proportional to the page size on every call, even for pages that contain no code fence at all. On large table- or document-heavy pages this transient spike is large enough that, during a bulk `import`/`sync` on an already-loaded brain, it can OOM the worker. The `try/catch` around the lexer cannot rescue this — an OOM is process death, not a throwable error.

## Reproduction

A ~2 MB markdown table (the shape produced when a spreadsheet/doc is converted to a wiki page) — **no code fence** — drives `marked.lexer` to ~110 MB of transient heap to produce a handful of tokens:

```js
// repro.mjs — run from repo root: node --max-old-space-size=80 repro.mjs
import { marked } from 'marked';
const rows = 12000, cols = 16;
const header = '| ' + Array.from({length: cols}, (_, i) => 'col' + i).join(' | ') + ' |';
const sep    = '| ' + Array.from({length: cols}, () => '---').join(' | ') + ' |';
const body   = Array.from({length: rows}, (_, r) =>
  '| ' + Array.from({length: cols}, (_, c) => 'v' + r + '_' + c).join(' | ') + ' |').join('\n');
const md = '# Overview\n\n' + header + '\n' + sep + '\n' + body + '\n';  // NO ``` fence
marked.lexer(md);                                                       // <-- spikes here
```

| heap cap | result |
|---|---|
| `--max-old-space-size=512` | OK, ~110 MB heap for 3 tokens (≈60× input amplification) |
| `--max-old-space-size=80`  | **FATAL — JS heap out of memory** inside `marked.lexer` |

In a fresh process with plenty of headroom no single page OOMs; the failure appears when free heap drops below the per-page transient need (e.g. the worker is already holding a large brain's chunk/embedding working set). That matches the observed field failure: it surfaced only once the brain was substantially full, not on a cold import.

## Why this is wasteful as well as risky

On a representative brain, **~99% of importable pages contain no code fence**, so `marked.lexer` is pure wasted work on them — it can never yield a `fenced_code` chunk. The largest no-fence pages are exactly the ones with the biggest transient spike (converted tables/docs).

## Proposed fix

Skip the lexer entirely when the body has no fence marker:

```ts
// in extractFencedChunks(), before marked.lexer(...)
if (!/(^|\n)[ \t]{0,3}(```|~~~)/.test(markdown)) return out;
```

- Zero correctness cost: a page with no ``` / `~~~` fence cannot produce a fenced chunk.
- Must match **both** ` ``` ` and `~~~`, or tilde-fenced code extraction regresses.

### Scope / limitation (intentionally honest)

This removes the transient-allocation surface for the common (fence-less) case, which is what causes the field OOM. It does **not** make `marked.lexer` safe for pages that *do* contain a fence — a structurally pathological fenced page (e.g. very deeply nested lists) can still OOM. A complete hardening would additionally cap input size / nesting before lexing. Filing the fence-less skip first because it is zero-risk and covers the observed incident; happy to follow up with the input cap.

## Environment

- `marked` ^18.0.0
- Surfaced during `gbrain import`/`sync` of a large brain under memory pressure.

javieraldape:
## Summary

Fixes #2437. `extractFencedChunks()` (`src/core/import-file.ts`) ran `marked.lexer()` on **every** page body to pull fenced code blocks. `marked.lexer` transiently allocates roughly 60× the input size, so a large (~2 MB) fence-less page spikes to ~110 MB and can OOM-kill bulk `sync`/`import`. Since ~99% of importable pages contain no code fence, ~99% of those lexer runs were pure waste.

## Fix

Add a cheap guard at the top of `extractFencedChunks()` that returns early when the body contains no fenced-code opener at all (a backtick or tilde fence at line start, per CommonMark), so `marked.lexer()` never runs on fence-less pages. About 3 lines; the happy path (pages that do contain fences) is unchanged.

## Test

`test/fence-extraction.test.ts` gains two cases, both exercised through the public `importFromContent` path:
- a large (~96 KB) fence-less prose body yields no fenced chunks — with a sanity guard asserting the body truly contains no fence, so the test can't pass on a body that happens to include one; and
- a body with a fenced code block still extracts its chunk, proving the guard doesn't break the happy path.

## Verification

- `bun run typecheck` → clean (`EXIT=0`)
- `bun test test/fence-extraction.test.ts` → all pass

Closes #2437.

### DIFF
--- a/src/core/import-file.ts
+++ b/src/core/import-file.ts
@@ -115,6 +115,9 @@ async function extractFencedChunks(
   startChunkIndex: number,
 ): Promise<ChunkInput[]> {
   const out: ChunkInput[] = [];
+  // #2437: marked.lexer transiently allocates ~60x the input; skip it entirely
+  // when there's no fence to find (the ~99% case) to avoid OOM on large bodies.
+  if (!/(^|\n)[ \t]{0,3}(```|~~~)/.test(markdown)) return out;
   let tokens: ReturnType<typeof marked.lexer>;
   try {
     tokens = marked.lexer(markdown);
--- a/test/fence-extraction.test.ts
+++ b/test/fence-extraction.test.ts
@@ -136,4 +136,34 @@ echo hi
     const fenceChunks = chunks.filter(c => c.chunk_source === 'fenced_code');
     expect(fenceChunks.length).toBe(0);
   });
+
+  // #2437: a large fence-less body must produce zero fenced_code chunks. The
+  // fence-opener guard short-circuits marked.lexer (the transient-OOM source)
+  // for these — proven here by the [] result on a body with no ``` / ~~~ at all.
+  test('large fence-less prose page produces zero fenced_code chunks', async () => {
+    // ~50KB of prose, no code fence anywhere (the ~99% case, and the shape
+    // that drove marked.lexer to a ~60x transient spike and OOM).
+    const para =
+      'This is ordinary wiki prose with no fenced code block of any kind. ' +
+      'It mentions code informally but never opens a fence. ';
+    const md = `# Large fence-less page\n\n${para.repeat(800)}\n`;
+    expect(md.length).toBeGreaterThan(40_000);
+    expect(md).not.toMatch(/```|~~~/); // sanity: truly fence-less
+
+    await importFromContent(engine, 'guides/fence-less-large', md, { noEmbed: true });
+    const chunks = await engine.getChunks('guides/fence-less-large');
+    const fenceChunks = chunks.filter(c => c.chunk_source === 'fenced_code');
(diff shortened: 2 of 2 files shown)


time-attack:
Duplicate of #2440. Same fence-less import OOM fix; keep #2440 (green+CLEAN).

PR #2494, opened 2026-06-30
fix(cli): BigInt-safe JSON serialization so local-op output normalization doesn't crash (#2450)

### LINKED ISSUE #2450: CLI local-op output normalization crashes on BigInt (cli.ts JSON.stringify has no replacer)

javieraldape:
## Summary

Fixes #2450. The CLI local-op output normalizer crashed on `BigInt`: `JSON.stringify` throws on a bigint (e.g. a `BIGSERIAL` primary key from the Postgres engine), aborting the command before any renderer ran.

## Root cause

`src/cli.ts` round-trips every local read/write op result through `JSON.parse(JSON.stringify(rawResult))` for normalization, with no replacer. A `BigInt` anywhere in the result makes `JSON.stringify` throw.

## Fix

Add a small local `bigintReplacer` (serializes `BigInt` via `.toString()`) and route the normalization round-trip and the `formatResult` default renderer through it.

`.toString()` (not `Number()`) is deliberate: the normalization path exists to match the renderer output of the routed path, and postgres.js returns `int8` as a **string** on the wire — so `.toString()` preserves that parity and avoids the precision loss `Number()` would introduce past 2^53. (Matches the existing `gateway.ts` convention for serializing bigint.)

**Scope kept tight:** the two sites that share the live crash risk are fixed (the normalization round-trip + `formatResult`'s default case). The `reindex --multimodal --json` site was left alone because its input is typed entirely as `number` (TS guarantees no bigint reaches it), and `gbrain calibration --json` is unreachable per #2035 — the issue itself flags it as later treatment.

## Test

6 new cases in `test/cli.test.ts` covering the normalization path with a `BigInt` (no throw + correct string serialization, including a past-`MAX_SAFE_INTEGER` precision case) plus the `formatResult` default. Verified to throw on the pre-fix code and pass after.

## Verification

- `bun run typecheck` → clean (`EXIT=0`)
- `bun test test/cli.test.ts` → all pass

Closes #2450.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -440,9 +440,10 @@ async function main() {
     }
     // ENG-2 (renderer parity by data shape): JSON-round-trip the local-engine
     // path's return value so renderers see the same shape they'd see on the
-    // routed path. Date → ISO string; bigint → string (postgres.js shape);
-    // Buffer → object. Microsecond-cost; eliminates a whole drift bug class.
-    const result = JSON.parse(JSON.stringify(rawResult));
+    // routed path. Date → ISO string; bigint → string (postgres.js shape, via
+    // bigintReplacer — a bare stringify THROWS on bigint, #2450); Buffer →
+    // object. Microsecond-cost; eliminates a whole drift bug class.
+    const result = normalizeLocalResult(rawResult);
     const output = formatResult(op.name, result);
     if (output) process.stdout.write(output);
   } catch (e: unknown) {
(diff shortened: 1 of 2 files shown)


time-attack:
Closing as duplicate of #2452.

Competing PR for BigInt-safe JSON / calibration-takes CLI fixes (#2450). Keep #2452.

Issue #2187, opened 2026-06-14
LiteLLM user-defined embedding models are not recognized as embedding-capable


thiagoasilva89:
When configuring GBrain with a LiteLLM Proxy embedding model using a user-defined model name, GBrain does not recognize the litellm provider as embedding-capable.

The LiteLLM provider is documented/hinted as supporting usage like:

gbrain init \
  --url "$GBRAIN_DATABASE_URL" \
  --embedding-model litellm:gbrain-embedding \
  --embedding-dimensions 4096

However, after initialization, gbrain providers list shows litellm with EMBED —, and embedding commands fail with:

Provider "litellm" requires a specific model name to be configured.
  Set one: gbrain config set embedding_model litellm:<model-name>
  Or run with --no-embed to import-only and embed later.

This happens even though embedding_model is already configured as litellm:gbrain-embedding.

Environment

* GBrain version: 0.42.42.0
* Engine: PostgreSQL
* pgvector: enabled
* LiteLLM Proxy: running locally on http://localhost:4000/v1
* Embedding backend: vLLM OpenAI-compatible endpoint
* Embedding model: qwen3-embedding-8b-nvfp4
* Embedding dimensions: 4096
* Chat model: Qwen model behind LiteLLM, exposed as gpt-4o-mini

LiteLLM model list:

model_list:
  - model_name: gpt-4o-mini
    litellm_params:
      model: openai/qwen3.6-35b
      api_base: http://192.168.1.12:30000/v1
      api_key: EMPTY
  - model_name: gbrain-embedding
    litellm_params:
      model: openai/qwen3-embedding-8b-nvfp4
      api_base: http://192.168.1.12:30001/v1
      api_key: EMPTY
      mode: embedding
      extra_body:
        encoding_format: float

The LiteLLM endpoint works correctly:

curl -s http://localhost:4000/v1/embeddings \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer EMPTY" \
  -d '{
    "model": "gbrain-embedding",
    "input": "test"
  }' | jq '.data[0].embedding | length'

Returns:

4096

Steps to reproduce

1. Run LiteLLM Proxy with a user-defined embedding model, for example:

model_list:
  - model_name: gbrain-embedding
    litellm_params:
      model: openai/qwen3-embedding-8b-nvfp4
      api_base: http://192.168.1.12:30001/v1
      api_key: EMPTY
      mode: embedding
      extra_body:
        encoding_format: float

[...]

Issue #2028, opened 2026-06-10
query embed deadline (~6s) silently degrades hybrid search to keyword-only — empty results for CJK content on slow embedding providers


JunhaoV5:
## Summary

v0.42.20.0 (#1775, Fix 3) bounds the query-time embed with a shared ~6s deadline (`QUERY_EMBED_TIMEOUT_MS`, default 6000ms; cache-lookup embed + inner embed share one budget). On timeout the catch falls back to **keyword-only with no surfaced warning** (src/core/search/hybrid.ts: `catch { /* Embedding failure is non-fatal, fall back to keyword-only */ }`).

## Effect

For brains whose embedding provider routinely takes >6s round-trip (e.g. OpenRouter reached through a proxy — common outside the US), **every** `query` op loses its vector arm:

- English queries limp along on Postgres FTS hits, so the degradation is invisible.
- CJK content returns `[]` outright — `websearch_to_tsquery('english', ...)` can't tokenize CJK, so keyword-only ≡ empty. rc=0, `mcp_request_log` status `success`, latency pinned at ~6.0s. Nothing tells the operator the vector arm is dead.

In our case the vector arm was silently dead for ~9 days; smoke tests asserting "query exits 0" passed throughout. Found only by noticing empty-result latencies clustering exactly at 6034±40ms.

## Asks (any subset helps)

1. **Surface the degradation**: set a `vector_enabled: false` / `embed_timeout: true` flag in HybridSearchMeta and emit one stderr warning per process, so operators can see keyword-only fallbacks happening.
2. **Document `GBRAIN_QUERY_EMBED_TIMEOUT_MS`** (it exists and works — raising it to 30000 fully restored our hybrid search — but appears in no docs / doctor output).
3. Consider a doctor check: time one live query-embed against the configured provider and warn when p50 exceeds the deadline.

Thanks — retrieval cathedral is excellent once both arms are actually running.

ChenyqThu:
Confirming on v0.42.59.0, with a **second trigger** for the same failure and what looks like a **broken safeguard already in the file**.

## Our embedding provider is fast. Expansion starves it anyway.

We embed with `text-embedding-3-large` straight from `api.openai.com` — 171–312ms measured. The vector arm still dies, because `expandFn` (an LLM call) runs *inside* the same absolute deadline, before the embed. Timing each stage separately, same query, 6 consecutive rounds:

| round | expansion | embed | keyword | budget left when embed starts (6s default) |
[...]

PR #2496, opened 2026-06-30
v0.42.54.0 fix(ai): embedding preflight accepts user-named models on litellm/llama-server (#1716)

javieraldape:
## What

The embedding preflight (`diagnoseEmbedding` in `src/core/ai/gateway.ts`) rejected **every** embedding model on the two bring-your-own-model recipes — **LiteLLM** and **llama.cpp's llama-server** — as `user_provided_model_unset`, even when the user named a concrete model like `litellm:my-model`. Those users could not embed at all.

Because `isAvailable('embedding')` delegates to `diagnoseEmbedding(...).ok`, the same false rejection **silently degraded hybrid search to keyword-only** on those brains — no error surfaced.

## Fix

One predicate. Only emit `user_provided_model_unset` when the model is **genuinely unset** (`parsed.modelId` empty). A named model now passes preflight and falls through to the (empty) auth-env check → `ok: true`. A bare provider with no model still errors as `unknown_provider` (the parser rejects it upstream, before the guard), which points the user at the `provider:model` form.

**Scope correction vs the original plan:** Ollama was **not** affected — it ships a non-empty built-in model list, so its models always resolved (`ollama:nomic-embed-text` was already `isAvailable === true`). The fix targets LiteLLM and llama-server, the only two recipes with `models: []` + `user_provided_models: true`.

## Tests

First-ever direct `diagnoseEmbedding` tests in `test/embed-preflight.test.ts`:
- `litellm:my-model`, `llama-server:bge-m3` → `ok: true` — **fail on the old code** (verified by reverting the predicate)
- bare `litellm` → `ok: false`, `reason: 'unknown_provider'`
- `ollama:nomic-embed-text` → `ok: true` (regression pin)
- `isAvailable('embedding')` → `true` for the named litellm/llama-server models (pins the keyword-only cascade)

Verification run locally:
- `bun run typecheck` → clean
- 82 pass / 0 fail across `embed-preflight`, `ai/gateway`, `gateway-embed-model-override`, `openai-compat-multimodal`, `recipe-llama-server`, `no-batch-cap-suppression`
- Reverting the one-line predicate flips exactly the 3 fix-pinning tests to fail — confirming they pin the bug, not just rubber-stamp the current code

## Issue references

Addresses #1716 and the #1812 / #2028 keyword-only-search cascade. The original wave also listed #2295, #2204, #1740, #1672, #1674, #2187 — I verified the **code path**, not each issue's text. Any issue specifically about **Ollama** is NOT addressed here (Ollama already worked), so I've deliberately avoided auto-close keywords; please confirm per-issue before closing.

---

## Reviewer notes (local verification)

I read `CONTRIBUTING.md` after opening this PR; gate status:

- **Full `bun run test`** — green except **2 pre-existing, environment-only failures** unrelated to this change: `brain-repo-durability.serial.test.ts` and `brain-durability-hook.serial.test.ts`. Both fail on `git push origin HEAD` against a throwaway temp repo (no reachable remote in a sandbox). **Verified they fail identically on a clean `origin/master` checkout** (23 pass / 2 fail, same `scaffolding commit/push failed` cause), so they're independent of this diff. This PR's own `embed-preflight` tests pass.
- **`bun run verify`** — release-relevant guards (privacy, jsonb, doc-history, no-pii-agent-voice, **test-isolation R1–R4**, typecheck) pass when run individually. The parallel wrapper's aggregate couldn't be confirmed on a host lacking a `timeout`/`gtimeout` binary (its fallback SIGTERMs passing checks); CI runs it correctly.
- **`gbrain eval replay`** — not run. `src/core/ai/gateway.ts` isn't in the retrieval trigger-path list, and this is a preflight *availability* fix (not ranking/fusion): it doesn't change scores or slug sets for queries that already worked — it only enables vector search for litellm/llama-server brains that were stuck keyword-only. Replay also needs a captured corpus a fresh checkout doesn't have.
- **Test isolation** — new tests drive the gateway via `configureGateway({env})` (no `process.env` mutation, no `mock.module`, no PGLite) → R1–R4 clean, parallel-shard safe.


time-attack:
Closing: current master already contains this fix. Master src/core/ai/gateway.ts:779-796 already removed the false-positive guard: comment reads 'This REPLACES the prior user_provided_model_unset guard, which was a structurally-unreachable... only ever false-positived for a fully-specified litellm:<model> with dims set, silently disabling vector search.' Master's replacement (user_provided_dims_unset, #1292/D6) lets named litellm/llama-server models pass preflight when embedding_dimensions is set, and is deliberately stricter (fail-closed on missing dims) than this PR's plain guard removal.

Issue #2337, opened 2026-06-21
init: refuse explicit --embedding-dimensions for user_provided_models (litellm, llama-server)


schsiung:
## Summary

`gbrain init` with `--embedding-model litellm:<model> --embedding-dimensions <N>` fails for any positive `N`, even though `litellm` (and `llama-server`) **require** an explicit dimension because `default_dims: 0`.

This blocks using OpenAI-compatible embedding backends (e.g. SiliconFlow `BAAI/bge-m3` at 1024d) through the documented `litellm` escape hatch.

## Reproduction

```bash
git clone https://github.com/garrytan/gbrain.git ~/gbrain && cd ~/gbrain
bun install && bun link

export LITELLM_BASE_URL=https://api.siliconflow.cn/v1
export LITELLM_API_KEY=<any-key>

gbrain init --pglite \
  --embedding-model litellm:BAAI/bge-m3 \
  --embedding-dimensions 1024
```

**Actual:**

```
Refusing to init: Provider "litellm" model "BAAI/bge-m3" does not support custom dimensions 1024
(this model only emits its default vector size). Either drop --embedding-dimensions or pick a
Matryoshka-aware model.
```

**Expected:** init succeeds with schema sized to 1024 (same as `llama-server` + explicit dims).

## Root cause

In `src/core/embedding-dim-check.ts`, `isCustomDimValidForProvider()` treats any `--embedding-dimensions` that differs from `default_dims` as a Matryoshka override. For `user_provided_models` recipes (`litellm`, `llama-server`), `default_dims` is intentionally `0` (sentinel), and `init.ts` **requires** `--embedding-dimensions` when using these providers.

So the flow is contradictory:

1. `init.ts` (lines ~265–270): refuses init without `--embedding-dimensions`
2. `embedding-dim-check.ts` (Tier 3): refuses init **with** `--embedding-dimensions` unless the model is Matryoshka-aware

## Proposed fix

Accept explicitly declared dimensions for `user_provided_models` recipes before the Matryoshka tier:

```typescript
// Tier 0: user-provided model recipes (litellm, llama-server). These ship
// default_dims=0 as a sentinel — init requires an explicit --embedding-dimensions
// because there's no canonical model. Trust the caller's declared native width.
if (recipe.touchpoints.embedding?.user_provided_models === true) {
  if (Number.isInteger(requestedDims) && requestedDims > 0) {
    return { valid: true, error: '' };
  }
}
```

[...]

Issue #2176, opened 2026-06-14
`litellm:` (and `llama-server:`) embedding setup is a catch-22: `--embedding-dimensions` is required yet rejected


SirMicksalot:
## Summary

Configuring an embedding model through the **`litellm`** proxy recipe is impossible: gbrain requires `--embedding-dimensions` for it, but then rejects any value you pass. The two errors are mutually exclusive, so there is no valid invocation. The **`llama-server`** recipe has the identical defect (both declare `default_dims: 0`).

Verified by source inspection of **v0.42.42.0** (installed via `bun install -g github:garrytan/gbrain`), macOS, PGLite engine.

## The catch-22

**Omit `--embedding-dimensions`:**
```
Embedding dimensions must be a positive integer; got 0.
```

**Pass `--embedding-dimensions 768` (or any N > 0):**
```
Provider "litellm" model "<model>" does not support custom dimensions 768
(this model only emits its default vector size).
Either drop --embedding-dimensions or pick a Matryoshka-aware model.
```

So: the first error says *add* `--embedding-dimensions`; the second says *drop* it. No setting satisfies both. Reproduced against every backend tried through the proxy (openai `text-embedding-3-small`/`-3-large`, `cohere/embed-v4.0`, fireworks `nomic`).

## Root cause (source-pinpointed)

1. `src/core/ai/recipes/litellm-proxy.ts:27` sets `default_dims: 0` with the comment *"user must declare --embedding-dimensions explicitly"*, and `setup_hint` (line 44) instructs the user to `pass --embedding-model litellm:<model> and --embedding-dimensions <N>`.

2. In `src/core/embedding-dim-check.ts`, `validateDimAgainstTouchpoint()`:
   - Line 369: `const dim = requestedDims ?? defaultDims;` → with no `--embedding-dimensions`, `dim = 0`.
   - Lines 371-374: `dim <= 0` → **"Embedding dimensions must be a positive integer; got 0."** (the *omit* half).
   - Line 386: `if (requestedDims !== undefined && requestedDims !== defaultDims)` → since `defaultDims = 0`, **any** passed `N > 0` enters custom-dim validation.

[...]

Issue #2170, opened 2026-06-13
Ollama provider hardcodes 768d, rejects 1024d models (bge-m3, mxbai-embed-large)


chvilalukas:
**What happened?**

The `ollama` embedding recipe pins a single `default_dims: 768` for the whole provider and provides no per-model dimension. As a result, two models that the docs themselves advertise as 1024-dimensional cannot be used at all:

- `ollama:bge-m3` — emits 1024d, but gbrain configures a 768d schema → embedding fails.
- `ollama:mxbai-embed-large` — `docs/integrations/embedding-providers.md` lists it as 1024d, but gbrain also configures 768d for it → same failure.

Two distinct dead ends:

1. **With `--embedding-dimensions 1024`** → `init` refuses outright:
   ```
   Refusing to init: Provider "ollama" model "bge-m3" does not support custom dimensions 1024
   (this model only emits its default vector size). Either drop --embedding-dimensions or pick a Matryoshka-aware model.
   ```
   (Same message for `ollama:mxbai-embed-large`.)

2. **Without `--embedding-dimensions`** → `init` silently sets a 768d schema, then `gbrain embed --stale` fails at runtime:
   ```
   Embedding dim mismatch: model bge-m3 returned 1024 but schema expects 768.
   ```

So there is no flag combination that lets a stock install use `ollama:bge-m3` (or `ollama:mxbai-embed-large`) end-to-end.

**Root cause (from source, v0.42.42.0)**

- `src/core/ai/recipes/ollama.ts` declares a single `touchpoints.embedding.default_dims: 768` with `models: ['nomic-embed-text', 'mxbai-embed-large', 'all-minilm']` and no `dims_options`. Every Ollama model inherits 768d regardless of its true output width.
- `src/core/ai/embedding-dim-check.ts` → `isCustomDimValidForProvider()` only allows a custom dimension when the provider is in a hardcoded Matryoshka allow-list (`voyage`, `zeroentropyai`, `openai`). For `ollama` it falls through to the Tier-3 rejection at ~L458 ("does not support custom dimensions"), so `--embedding-dimensions 1024` is always refused for Ollama models even when the model genuinely emits 1024d.

The Ollama OpenAI-compatible endpoint confirms the real width:
```
$ curl -s http://localhost:11434/v1/embeddings -H 'Content-Type: application/json' \
    -d '{"model":"bge-m3","input":"test"}' | jq '.data[0].embedding | length'
1024
```

**What did you expect?**

[...]

PR #2498, opened 2026-06-30
v0.42.54.0 feat(ai): local-provider custom embedding dims + broaden Ollama catalog (#2170 #2271 #2051)

javieraldape:
## What & why

Local embedding via Ollama couldn't be initialized with 1024-dim models (`bge-m3`, `mxbai-embed-large`) — the most common modern local-embedder setup. `gbrain init --embedding-model ollama:bge-m3 --embedding-dimensions 1024` exited 1 at preflight.

While verifying the root cause I found **the bug is bigger than the tickets**: it's not Ollama-specific. Running the pure resolver (`resolveSchemaEmbeddingDim`) confirms **all three local providers** are broken at init for any explicit dimension:

| input | before | after |
|---|---|---|
| `ollama:bge-m3 @ 1024` | ❌ rejected | ✅ |
| `ollama:mxbai-embed-large @ 1024` (already in the recipe!) | ❌ rejected | ✅ |
| `ollama:nomic-embed-text @ 768` (default) | ✅ | ✅ |
| `llama-server:my-model @ 1024` | ❌ rejected | ✅ |
| `litellm:some-model @ 1024` | ❌ rejected | ✅ |

## Root cause (verified)

`src/core/embedding-dim-check.ts:isCustomDimValidForProvider` validates a non-default requested dim through tiers: recipe `dims_options` → provider branches for **only** voyage/zeroentropy/openai → a blanket "does not support custom dimensions" reject. Every local provider falls through to that final reject.

A second correctness fact: for `openai-compatible` providers like Ollama, gbrain sends **no** `dimensions` param (`dims.ts:dimsProviderOptions`), so Ollama returns each model's **native** dim and cannot truncate. The schema's `vector(N)` column must therefore equal the model's true native dim exactly — which is why a per-model native-dim map is a *correctness* requirement, not a nicety.

## Changes (4 files, no migration)

- **`src/core/ai/dims.ts`** — `OLLAMA_NATIVE_DIMS` (verified against ollama.com/library: nomic 768, mxbai 1024, all-minilm 384, bge-m3 1024, snowflake-arctic-embed2 1024), `ollamaNativeDim()`, and `isLocalUserProvidedDims(recipe)` (ollama by id + llama-server/litellm via the existing `user_provided_models` flag).
- **`src/core/embedding-dim-check.ts`** — new local-provider branch before the final reject: trusts the explicit dim (the running backend is authoritative), but for *known* Ollama models rejects a contradicting dim with a paste-ready fix (since Ollama can't truncate).
- **`src/core/ai/recipes/ollama.ts`** — adds `bge-m3` + `snowflake-arctic-embed2`; `setup_hint` notes explicit dims + init-time sizing.
- **`src/commands/init.ts`** — a verbose Ollama model picked **without** `--embedding-dimensions` now infers its native dim instead of the nomic-anchored 768 default. Closes a silent wrong-width footgun the preflight alone can't catch (when the derived dim equals `default_dims`, the custom-dim check is bypassed).
- Docs: `docs/integrations/embedding-providers.md` Ollama section updated for the new models + dim behavior.

## Scope decision

The `isLocalUserProvidedDims` helper is the honest predicate, so it fixes llama-server + litellm at init in the same stroke. This is **partial progress on #2176 / #2337** (their dim-init portion) — not full closure (those may also cover probe/model-list/multimodal work).

## Testing

- `bun test test/embedding-dim-check.test.ts test/ai/recipe-ollama.test.ts test/ai/recipes-existing-regression.test.ts` — **57 pass / 0 fail** (7 new resolver cases + new recipe-shape test).
- Broader sweep `test/ai/` + `test/providers.test.ts` + `test/brain-score-recommendations.test.ts` — **360 pass / 0 fail**.
- E2E `test/e2e/init-fresh-pglite.test.ts` in isolation — **14 pass / 0 fail**.

**Acceptance met:** `ollama:bge-m3 --embedding-dimensions 1024` → builds a 1024d schema; `ollama:bge-m3` with no flag → now derives 1024 (not 768); `ollama:nomic-embed-text @ 1024` → correctly refused with the right value to use.

## Notes for reviewers

- **One flaky-looking result, disclosed:** running ~11 test files (incl. heavy E2E) in a single parallel `bun test` invocation, an unrelated E2E test (`bug-reporter no-op keys ... Levenshtein`) hit a `beforeEach/afterEach` **hook timeout** (5002ms, "killed 1 dangling process"). It is **not** caused by this change — it's in a setup/teardown hook, the test exercises config-key suggestions (not embeddings), the changed `init.ts` path only runs for `providerId === 'ollama'`, and the file **passes 14/0 in isolation**. It's an artifact of oversubscribing the machine; CI shards E2E so it won't manifest there.
- **Prepared without `/ship`:** the `/ship` and `/document-release` skills weren't available in the authoring session, so this PR was assembled by hand following the repo's version IRON RULES (VERSION + package.json + CHANGELOG trio audited to `0.42.54.0`, lockfile refreshed). The `/ship`-only steps (adversarial pre-landing review, coverage audit, full document-release doc sweep) were **not** run — a maintainer may want to run `/document-release` before merge. I did update the one directly-stale doc.
- **Branch vs workspace:** branch is `ollama-embedding-dims`; the Conductor workspace is `montevideo`. Not renamed per an explicit "do not rename" session instruction.


time-attack:
Closing: current master already contains this fix. Master fixed the same issues (#2170/#2271/#2051) via a trust_custom_dims passthrough tier in src/core/embedding-dim-check.ts (~line 462, tagged '#2271'), with the flag on all three local recipes (recipes/ollama.ts:29, llama-server.ts:35, litellm-proxy.ts:28), so explicit custom dims now pass for ollama/llama-server/litellm. Ollama catalog already broadened (qwen3-embed-8b, snowflake-arctic-embed-l-v2), and unlisted models like bge-m3 pass anyway since ollama is non-native tier (model-resolver.ts:126-128). The PR's residual delta (per-model native-dim map + init-time dim inference) is explicitly deferred on master: 'Per-model dims metadata is a tracked follow-up' (recipes/ollama.ts:20) — if wanted, that slice needs a fresh PR rebased on the trust_custom_dims design; this PR's approach (OLLAMA_NATIVE_DIMS branch replacing the reject) conflicts with master's mechanism.

Issue #2433, opened 2026-06-26
ModelMessage[] schema rejects Date in tool-result output.value (Postgres timestamps crash multi-tool loops)


dcarolan1:
Summary: Any multi-tool rollout or subagent loop that calls a brain tool returning Postgres timestamp columns (brain_list_pages, brain_query, etc.) crashes AI SDK v6 validation with AI_InvalidPromptError: The messages do not match the ModelMessage[] schema.
Repro:

Run gbrain skillopt <skill> (or any subagent loop) where a tool returns rows with updated_at/created_at.
The tool loop calls gateway.chat() to continue; toModelMessages() converts the tool-result message.
AI SDK v6 validates and throws the schema error the moment a tool result is included.

Root cause: toModelMessages() (src/core/ai/gateway.ts, the json branch of the tool-result mapping) emits { type: 'json', value: b.output }. node-postgres auto-deserializes timestamp columns into JavaScript Date instances. Zod's jsonValueSchema accepts only null | string | number | boolean | record | array — not Date — so the entire message tree is rejected. The failing path resolves to e.g. messages[N].content[1].output.value[0].updated_at: expected string, received Date.
Fix (one line):
ts// before
: { type: 'json', value: (b.output ?? null) }
// after
: { type: 'json', value: JSON.parse(JSON.stringify(b.output ?? null)) }
JSON.stringify invokes Date.prototype.toJSON() → ISO string; other non-JSON values get serialized too. Verified: rollout success rate went 1/5 → 5/5 on a real skill optimization.
src/core/ai/gateway.ts — the json branch of toModelMessages()'s tool-result mapping.

time-attack:
Closing as duplicate of the open fix PR #2491.

**Why:** This issue and PR #2491 describe the same problem; keeping one thread. Work continues on #2491.

_Triage: issue↔PR pair — closed the issue, left the fix PR open._

Issue #2115, opened 2026-06-12
subagent toolLoop: reasoning-model output breaks next-turn history — "Invalid prompt: messages do not match the ModelMessage[] schema" (deterministic on DeepSeek V4, intermittent on Gemini 3.5 Flash)


felix1904:
## Version
gbrain 0.42.37.0

## Summary
When `models.dream.synthesize` / `models.subagent` is a model that emits reasoning content, the multi-turn subagent loop (`src/core/minions/handlers/subagent.ts` → gateway toolLoop) intermittently or deterministically fails on the *next* call with the AI SDK validation error:

```
Invalid prompt: The messages do not match the ModelMessage[] schema.
```

Evidence points to reasoning/thinking parts of the assistant message being persisted into `subagent_messages` and rebuilt into the conversation history in a shape the AI SDK rejects on the subsequent turn/attempt.

## Observed (dream synthesize jobs, via OpenRouter)
- `deepseek/deepseek-v4-pro` (thinking on by default): **deterministic** — 3/3 attempts fail with the schema error; job dead-letters after burning the full transcript input 3×.
- `google/gemini-3.5-flash` (thinking mandatory): **intermittent** — same job: attempts 1–2 fail with the schema error, attempt 3 succeeds cleanly (5 turns, end_turn). Retry roulette, ~2/3 of attempt budget wasted.
- `anthropic:` models (no reasoning emitted by default): never reproduced.

## Why this is distinct from (but related to) #1886 / #2039
#1886 covers dropped tool-result messages and #2039 covers crash-mid-round dangling tool-call tails. This report adds the **reasoning-content dimension**: with reasoning models the corruption occurs *without* any crash — a clean turn that includes reasoning parts is enough to poison the rebuilt history. Fix likely overlaps: strip/normalize reasoning parts when persisting and rebuilding history (most providers require either omitting reasoning or echoing it in a provider-specific field — e.g. DeepSeek's thinking mode returns 400 unless `reasoning_content` is handled per their spec).

## Repro sketch
1. `gbrain config set models.dream.synthesize openrouter:deepseek/deepseek-v4-pro`
2. Queue a synthesize job on any ≥2000-char transcript that triggers tool use (search + put_page).
3. Watch attempts 2..n fail with the ModelMessage[] schema error.

time-attack:
[...]

PR #2499, opened 2026-06-30
fix(gateway): reliable tool-loop resume + Date-safe tool output on non-Anthropic providers (#2256 #2273 #2110 #2433)

### LINKED ISSUE #2256: Gateway subagent loop unreliable for multi-turn tool calls on non-Anthropic providers (resume loop + Date-in-output rejection)

javieraldape:
## Why

On non-Anthropic providers (Gemini / DeepSeek / openai-compat) the agentic subagent loop (`agent.use_gateway_loop=true`) is unusable across multi-turn tool conversations: after a crash+resume it rebuilds an unbalanced history ("Tool results are missing" / "messages do not match the ModelMessage[] schema"), and any tool whose output contains a Postgres `Date` crashes the loop. These are the core reasons non-Anthropic agent loops don't work today.

## Root causes & fixes

**A — tool-result turn never persisted.** `toolLoop` built the tool-result user message but did `void userMessageIdx` — pushing it in-memory only. Only assistant turns were persisted, so resume rebuilt assistant `tool_use` blocks with no following `tool_result` turn → unbalanced → next `chat()` rejects.
→ Added `onToolResultMessage` to `ToolLoopOpts`, awaited **before** the in-memory push (mirroring `onAssistantTurn`'s persist-before-proceed ordering), and wired it in `runSubagentViaGateway` to persist the `role:'user'` tool-result turn to `subagent_messages`.

**B — mid-execution crash leaves a dangling tool_use.** A worker killed mid-tool-round persists the assistant `tool_use` turn (it fires before dispatch) but not the tool-result wrapper — the outputs live only in `subagent_tool_executions`. The existing crash-replay tests **stub `chat()`** and so never validated that the rebuilt history is balanced; against a real provider it still rejected the dangling `tool_use`.
→ On resume, synthesize the tool-result user turn from the **settled** executions so the rebuilt history is balanced before the first `chat()`. Done **subagent-side** (reads `output`/`error` directly) so it is **v1- and v2-safe**: a loop-side re-dispatch would mis-key v1 legacy rows (`ordinal=NULL` misses the `ON CONFLICT` key) and re-execute a completed tool. A still-`pending` tool (crashed mid-`execute`) is surfaced as unrecoverable rather than re-run or fabricated.

**C — `Date` in tool output.** `toModelMessages` emitted a tool output object straight into a `{ type:'json', value }` AI-SDK part; a `Date` survived as a `Date` instance and failed AI-SDK v6 `JSONValue` validation.
→ Round-trip the `json` arm through `toJsonSafe` (`Date` → ISO string). Serialization fix at the SDK boundary — **not** a `::jsonb` DB cast.

## Tests

The transport-stubbed suite missed all three bugs, so the new tests are written to actually catch them (each was confirmed to fail without its fix):

- `test/ai/gateway-tool-loop.test.ts` — multi-turn run asserting `onToolResultMessage` fires once per tool round with contiguous interleaved `message_idx`, then rebuilds the conversation and asserts it is **balanced** (every assistant tool-call turn is followed by its matching tool-result turn).
- `test/e2e/subagent-crash-replay-multi-provider.test.ts` — captures the **actual messages** handed to `chat()` on resume and asserts they're balanced (`toModelMessages` projects cleanly, no dangling `tool_use`), the synthesized turn is persisted durably, and a still-pending non-idempotent trailing tool surfaces as unrecoverable.
- `test/gateway-model-messages.test.ts` — a `Date` (and nested `{ts: Date}`) tool output yields a JSON-safe value (no `Date` instance survives; ISO string present).

## Scope

No schema change — reuses `subagent_messages` + `persistMessage` (binds via `$N::text::jsonb`, CLAUDE.md-compliant). Engine parity unaffected. No version/CHANGELOG bump (left to the maintainer's release wave).

## Verification

- `bun run typecheck` — clean
- subagent-direct tests (7 files incl. crash-replay e2e): 89 pass / 0 fail
- gateway-direct tests (4 files): 43 pass / 0 fail
- negative check: both new e2e tests fail with the fix disabled (genuine guards)
- guards pass: `check:jsonb`, `check:gateway-routed`, `check:test-names`, `check:privacy`, `check:test-isolation`, `check:proposal-pii`

Fixes #2256, #2273, #2110, #2433. Refs #2115, #2202.

### DIFF
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -2324,6 +2324,18 @@ export interface ChatToolDef {
 }
 
 /**
+ * Round-trip a tool output through JSON so non-JSON values (e.g. a Postgres
+ * `Date` returned by a tool) become AI-SDK v6 `JSONValue`-safe — a `Date`
+ * serializes to its ISO string instead of surviving as a `Date` instance that
+ * the SDK's Zod `JSONValue` validation rejects. This is a serialization fix at
+ * the SDK boundary, NOT a `::jsonb` DB cast. BigInt/circular outputs would throw
+ * in `JSON.stringify`; those are already not LLM-serializable and out of scope.
+ */
+function toJsonSafe(v: unknown): unknown {
+  return JSON.parse(JSON.stringify(v ?? null));
+}
+
+/**
  * Convert gbrain's provider-neutral ChatMessage[] into AI SDK v6 ModelMessage[].
  *
  * The original code passed `opts.messages as any` straight to generateText,
@@ -2354,7 +2366,7 @@ export function toModelMessages(messages: ChatMessage[]): unknown[] {
               ? { type: 'error-text' as const, value: typeof b.output === 'string' ? b.output : JSON.stringify(b.output) }
               : (typeof b.output === 'string'
                 ? { type: 'text' as const, value: b.output }
-                : { type: 'json' as const, value: (b.output ?? null) as never }),
+                : { type: 'json' as const, value: toJsonSafe(b.output) as never }),
           })),
       };
     }
@@ -2891,6 +2903,8 @@ export interface ToolLoopOpts {
    *   2. onToolCallStart   — pending row persisted (D11 step 2)
    *   3. handler.execute   — side effect
    *   4. onToolCallComplete / onToolCallFailed (D11 step 4)
+   *   5. onToolResultMessage — tool-result user message persisted before the
+   *      in-memory push, so a crash mid-loop replays a balanced history.
    */
   onAssistantTurn?: (turnIdx: number, messageIdx: number, blocks: ChatBlock[], usage: ChatResult['usage'], model: string) => Promise<void>;
   /**
@@ -2908,6 +2922,15 @@ export interface ToolLoopOpts {
   ) => Promise<{ gbrainToolUseId: string }>;
   onToolCallComplete?: (gbrainToolUseId: string, output: unknown) => Promise<void>;
   onToolCallFailed?: (gbrainToolUseId: string, error: string) => Promise<void>;
+  /**
+   * Persist the tool-result user message that closes a tool round. Mirrors the
+   * onAssistantTurn ordering: fired (and awaited) BEFORE the in-memory push so a
+   * crash mid-loop resumes with a balanced history (every assistant tool_use turn
+   * is followed by its tool_result turn). Without this, only assistant turns are
+   * persisted and resume rebuilds an unbalanced conversation that the next chat()
+   * rejects ("Tool results are missing").
+   */
+  onToolResultMessage?: (turnIdx: number, messageIdx: number, blocks: ChatBlock[]) => Promise<void>;
 
   /** Optional per-call heartbeat for observability. */
   onHeartbeat?: (event: string, data: Record<string, unknown>) => void;
@@ -3131,9 +3154,12 @@ export async function toolLoop(opts: ToolLoopOpts): Promise<ToolLoopResult> {
 
     if (stopReason === 'aborted') break;
 
-    // Feed all tool results back as a single user message.
+    // Feed all tool results back as a single user message. Persist BEFORE the
+    // in-memory push (mirrors onAssistantTurn) so a crash mid-loop resumes with a
+    // balanced history. turnIdx here is still the just-finished assistant turn's
+    // index; message_idx is what governs replay ordering.
     const userMessageIdx = messageIdx++;
-    void userMessageIdx;
+    await opts.onToolResultMessage?.(turnIdx, userMessageIdx, toolResultBlocks);
     messages.push({ role: 'user', content: toolResultBlocks });
 
     turnIdx++;
--- a/src/core/minions/handlers/subagent.ts
+++ b/src/core/minions/handlers/subagent.ts
@@ -785,6 +785,63 @@ async function runSubagentViaGateway(args: GatewayRunArgs): Promise<SubagentResu
     content: adaptContentBlocksToChatBlocks(m.content_blocks),
   }));
 
+  // Resume rebalance (#2256). A worker killed mid-tool-round leaves the trailing
+  // assistant tool_use turn persisted (onAssistantTurn fires BEFORE dispatch) but
+  // its tool-result wrapper unwritten — the outputs live only in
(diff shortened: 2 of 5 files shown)


zsimovanforgeops:
Additional confirmation from a large Postgres-backed install on **v0.42.57.0**:

- The worker was restarted after the v0.42.57 merge, so this is not stale process code.
- Long `patterns` subagents still reach a transient first-attempt failure, then retries rebuild an unbalanced history and die with `Tool results are missing for tool calls ...`.
- An offline repro against persisted rows is deterministic: a fresh in-memory tool round validates, while the rehydrated sequence contains the assistant tool-call turn without its matching persisted tool-result turn.
- A separate minimal repro confirms a raw `Date` in persisted assistant content fails AI SDK message validation.
- The failure reproduces on the native Anthropic path, so it is provider-independent.

We have contained the production impact through the supported config seam (`dream.patterns.enabled=false`) rather than carrying another private source patch.

The repair shape in this PR—durable tool-result persistence, settled-execution reconciliation on resume, and Date-safe serialization at the SDK boundary—matches the observed failure boundaries. I can test this branch in an isolated clone against the preserved repros if that would help.

zsimovanforgeops:
Follow-up verification completed in an isolated clone:

1. PR branch `da85f8934f905976f39f1eedea49bfdd8f6ffe67`
   - `bun test test/ai/gateway-tool-loop.test.ts test/e2e/subagent-crash-replay-multi-provider.test.ts test/gateway-model-messages.test.ts`
   - **33 pass / 0 fail / 109 assertions**

2. Same PR merged locally with current upstream `v0.42.58.0` (`a25209bbb2bacf1b88e06fd5282b27f1bf4a3e7a`)
   - merge applied with **no conflicts**, including `src/core/ai/gateway.ts`
   - same targeted command: **33 pass / 0 fail / 109 assertions**
   - replay matrix passed for Anthropic, OpenAI, Google, OpenRouter, and DeepSeek; combined tree initialized through schema v122

The disposable merge was aborted after verification; no production fork or database was touched.

time-attack:
Closing as duplicate of #2257.

Competing PR for gateway tool-loop resume on non-Anthropic providers. Keep #2257.

PR #2500, opened 2026-06-30
fix(extract-timeline-from-meetings): bump pages.updated_at for touched entities

pablofavre12:
When projecting timeline entries from meetings onto entity pages (attendees and body mentions), the entity pages get new timeline_entries but their pages.updated_at is left behind. This causes get_health to report them as stale.

Bump updated_at on every touched entity page after the batch insert so the health metric stays honest.


time-attack:
Thank you for the careful work here — the observation that pages.updated_at lags behind timeline projection is legitimate. I'm closing this because the PR is much broader than its title: it also adds a new apply_timeline_baseline op that inserts synthetic 'Page created' entries specifically to raise timeline_coverage, and redefines the brain_score timeline denominator. Backfilling placeholder rows to lift a health metric moves the score without moving the underlying brain quality, which is the opposite of what these metrics are for. Also note VERSION/package.json/CHANGELOG bumps belong to the maintainer's ship flow, not contributor PRs.

PR #2505, opened 2026-06-30
Tune cycle freshness for nightly source dreams

sene1337:
## Summary

This PR makes `doctor.cycle_freshness` match an intentionally once-nightly dream cadence:

- scopes the check to operating sources via `doctor.cycle_freshness.source_allowlist` / `source_ignorelist`, with `DREAM_SOURCE_ALLOWLIST` as a runtime fallback
- adds DB-configured threshold keys: `doctor.cycle_freshness.warn_hours` and `doctor.cycle_freshness.fail_hours`
- keeps env vars as the highest-precedence override for emergency/operator use
- registers the new keys so `gbrain config set ...` works without `--force`

## Why

A once-nightly dream cron cannot stay green under the old 6h warn / 24h fail defaults, even when it is healthy. Operators can now set daily-with-grace thresholds once in DB config, e.g. warn at 30h and fail at 48h, and have both interactive and HTTP/remote doctor use the same cadence.

## Verification

- `bun test test/doctor-cycle-freshness.test.ts test/config-set.test.ts`
- `bun run typecheck`

### DIFF
--- a/src/commands/doctor.ts
+++ b/src/commands/doctor.ts
@@ -3742,6 +3742,105 @@ export async function checkPoolBudget(_engine: BrainEngine): Promise<Check> {
   }
 }
 
+type CycleFreshnessSourceScope = {
+  allowlist?: Set<string>;
+  ignorelist: Set<string>;
+  invalid: string[];
+  configured: boolean;
+};
+
+const CYCLE_FRESHNESS_SOURCE_ID_RE = /^[A-Za-z0-9_-]+$/;
+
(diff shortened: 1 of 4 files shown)


time-attack:
Thanks for the clean implementation and the precedence discipline (env over DB config is the right ordering). I'm closing this one on policy: it adds four new config keys plus an env fallback and a new source allow/ignore-list mechanism to doctor, and that's a meaningful expansion of configuration surface to accommodate one deployment's cron cadence rather than a fix for something broken. Per-install threshold tuning of a health check is not surface we're taking on.

PR #2506, opened 2026-06-30
Route calibration cycle models through reasoning tier

sene1337:
## Summary

Routes hindsight calibration-cycle model selection through the model-tier resolver instead of hardcoded/native Anthropic defaults:

- adds phase-specific config keys for `propose_takes`, `grade_takes`, and `calibration_profile`
- falls back to `models.tier.reasoning` when a phase-specific key is unset
- registers the new keys so `gbrain config set ...` works without `--force`
- updates docs/tests around the new routing behavior

## Why

Private-gateway or local-first operators can set one reasoning-tier model and have all calibration-cycle phases use it. Without this, the propose/grade/profile phases can drift back to native Anthropic defaults even when the rest of the brain is configured for private routing.

## Verification

- `bun test test/propose-takes.test.ts test/grade-takes.test.ts test/grade-takes-ensemble.test.ts test/calibration-profile.test.ts test/cycle-model-routing.test.ts test/config-set.test.ts` — 110 pass
- `bun run typecheck`


time-attack:
Thanks @sene1337! 🙏

What this does: This PR adds model-tier routing for the three calibration cycle phases (propose_takes, grade_takes, calibration_profile), allowing them to respect the brain's tier configuration instead of hardcoding claude-sonnet-4-6. It creates a…

This overlaps with #2805, which we're carrying forward, so I'm closing this in its favor to keep the review queue focused. If yours covers a case #2805 misses, reopen with that detail and we'll reconcile them.

PR #2509, opened 2026-06-30
v0.42.57.0 fix(cli): whoknows --explain renders per-result factor breakdown

SeanGearin:
## Summary
`gbrain whoknows "<topic>" --explain` was silently a no-op on the command line — it printed the same output with or without the flag, so the per-result ranking factor breakdown it advertises was unreachable from the CLI.

**Root cause:** `--explain` is a *global* flag (`src/core/cli-options.ts`) stripped from argv before op dispatch, so it never reached the `find_experts` op's `explain` param. The live human output for `gbrain whoknows` goes through `formatResult('find_experts')`, which had no case and fell through to the raw-JSON `default:` — so the readable breakdown was never rendered. (The ranking data was never lost: `find_experts` always returns each result's `factors` in its JSON, so agent / `--json` callers were unaffected.)

**Fix (presentation-only):** `formatResult` now reads the global flag via `getCliOptions().explain` and renders the expertise / recency / salience breakdown — the exact pattern `gbrain search --explain` and `gbrain query --explain` already use. Without `--explain`, output is unchanged (raw JSON), so scripted consumers are unaffected. The rendering logic is extracted into `formatWhoknowsResults()` so the legacy `runWhoknows` path and the live op-dispatch formatter stay in lockstep.

**Why not the "obvious" fix:** pushing `--explain` back into argv (so `parseOpArgs` sets the op param) would let `parseOpArgs` swallow the positional query in `gbrain search --explain "my query"` (an unknown flag consumes the next token). The formatter approach sidesteps that and keeps `--explain` position-independent for all three commands.

### Tool surface unchanged
The `find_experts` MCP op, its name, and its params (`topic` / `limit` / `explain`) are untouched — the only `operations.ts` change is a clarifying `//` code comment, which does not affect the generated tool schema. The customer-facing MCP tool count is unchanged.

## Test Coverage
- New regression test `test/whoknows-explain-format.test.ts` locks both halves of the wiring: the shared formatter (breakdown present with `explain`, absent without) and the live `formatResult('find_experts')` path under `getCliOptions().explain` on/off (the exact broken locus).
- The other half of the chain (`parseGlobalFlags --explain → cliOpts.explain`, position-independent) is already covered by `test/cli-options.test.ts`.
- Blast-radius suite green: 67/67 across `whoknows*`, `find-experts-op`, `cli`, `cli-args`, `cli-format-volunteer`, `cli-query-image`, and `e2e/whoknows`. Typecheck clean.

## Pre-Landing Review
No issues. CLI presentation only — no SQL, data-safety, auth, or LLM-trust-boundary surface. The `require('./commands/whoknows.ts')` inside `formatResult` matches the production-proven `search`/`query` `require` two cases above it; `find_experts` always returns an array (errors are caught upstream before `formatResult`).

## Test plan
- [x] Targeted blast-radius unit + e2e tests: 67/67 pass
- [ ] Full unit suite runs in CI (`.github/workflows/test.yml`) — it's a ~2h local run, delegated to CI for this CLI-only change


time-attack:
The diagnosis of the root cause is correct — `parseGlobalFlags` strips `--explain` at cli.ts:218 before dispatch, so `runWhoknows`'s own parse never sees it. But the fix targets the wrong path: on current master `gbrain whoknows` dispatches at cli.ts:1941 straight into `runWhoknows(engine, args)` and never reaches `formatResult('find_experts')`, so the new case is dead code and `parsed.explain` stays false after this change. Closing this one; the actual repair is reading `getCliOptions().explain` inside `runWhoknows` — a one-liner rather than a 203-line refactor plus a `require()` in an ESM module.

SeanGearin:
Confirmed - you're right, and thanks for tracing it to the actual dispatch site rather than taking the PR body's word for it. `whoknows` never reaches `formatResult('find_experts')`, so the new case was dead code and the tests were green against the wrong path. The `getCliOptions().explain` read inside `runWhoknows` is the right repair. Appreciate the close-out.

PR #2513, opened 2026-07-01
Fix cjk slug validate

loyur:


time-attack:
Closing during backlog triage: the diff removes several unrelated CI workflow files, and the slug validation change regresses mixed CJK+Latin slugs. A focused re-roll of just the CJK slug fix (with a test) would be very welcome!

PR #2516, opened 2026-07-01
fix(dream): resolve model via gateway tier resolver in propose_takes, grade_takes, calibration_profile

ervza:
## Summary

The three calibration/dreaming phases (`propose_takes`, `grade_takes`, `calibration_profile`) hardcoded `'claude-sonnet-4-6'` as the model fallback, bypassing the gateway model-tier resolver. On non-Anthropic stacks (ollama, openrouter, litellm, openai-compat) without `ANTHROPIC_API_KEY`, these phases would fail or silently fall back to an unreachable model.

This follows the same pattern `synthesize.ts` already uses: `resolveModel()` with a config key, tier, and fallback. The actual LLM calls already route through `gateway.chat()` — only the model resolution was missing.

## What changed

- **`propose-takes.ts`**: Replaced 3 hardcoded `opts.model ?? 'claude-sonnet-4-6'` references with `resolveModel(engine, { configKey: 'models.dream.propose_takes', tier: 'reasoning', fallback: 'claude-sonnet-4-6' })`
- **`grade-takes.ts`**: Replaced hardcoded `opts.model ?? 'claude-sonnet-4-6'` with `resolveModel()` and updated the judge call to use the resolved model
- **`calibration-profile.ts`**: Replaced hardcoded `opts.model ?? 'claude-sonnet-4-6'` with `resolveModel()`

## Config keys

Users can override per-phase:
```bash
gbrain config set models.dream.propose_takes ollama-cloud:kimi-k2.6
gbrain config set models.dream.grade_takes ollama-cloud:deepseek-v4-flash
gbrain config set models.dream.calibration_profile ollama-cloud:kimi-k2.6
```

## Related

- Follows the same pattern as `synthesize.ts` (which already uses `resolveModel`)
- Complements #2279 (brettdavies' fix for the patterns phase ANTHROPIC_API_KEY gate)
- Complements #2208 (brettdavies' litellm chat/expansion touchpoints)

### DIFF
--- a/src/core/cycle/calibration-profile.ts
+++ b/src/core/cycle/calibration-profile.ts
@@ -27,6 +27,7 @@
 
 import { BaseCyclePhase, type ScopedReadOpts, type BasePhaseOpts } from './base-phase.ts';
 import { chat as gatewayChat } from '../ai/gateway.ts';
+import { resolveModel } from '../model-config.ts';
 import { gateVoice, type VoiceGateGenerator, type VoiceGateJudge } from '../calibration/voice-gate.ts';
 import { patternStatementTemplate, type PatternStatementSlots } from '../calibration/templates.ts';
 // v0.41 T10 — domain widening. The aggregator module resolves the active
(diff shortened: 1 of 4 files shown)


time-attack:
Thanks for spotting the non-Anthropic-stack hazard here. Master has since moved on two of the three call sites through a different route — propose_takes now resolves via `getChatModel()` (#2125) and calibration_profile via `TIER_DEFAULTS.reasoning` (#2892) — so this branch would conflict with and partly revert the pattern that landed. Closing as already fixed; the only residual is the `judgeModelId` literal in grade-takes.ts, which is a one-line change against current master rather than this four-file diff.

PR #2517, opened 2026-07-01
fix: harden doctor, embedding, and source config handling

Carlo576:
## Summary
- harden doctor warning cleanup, oversized embedding handling, and adaptive embed batch behavior
- preserve local doctor health tuning and categorize onboard doctor checks
- handle non-object source config arrays safely in PGLite/Postgres source config paths

## Verification
- gbrain --version => 0.42.53.0 after rebasing onto upstream master
- bun test test/op-checkpoint.test.ts test/check-jsonb-params.test.ts
- bun test test/doctor-categories.test.ts test/config-set.test.ts test/e2e/onboard-full-flow.test.ts test/list-all-sources.test.ts test/e2e/list-all-sources-postgres.test.ts
- bun run check:jsonb
- gbrain sync --source wiki --no-pull --no-embed --yes --json
- gbrain sync --source gbrain-src --no-pull --no-embed --yes --json
- gbrain embed --stale --priority recent --batch-size 500
- gbrain doctor --fast

## Notes
This branch was rebased onto upstream v0.42.53.0, so the upstream #2339 op_checkpoints jsonb fix is included from master rather than carried as a local hotfix.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,8 @@ test/fixtures/pglite-snapshot.version
 
 # Private brain reports — never check these in (per CLAUDE.md privacy rule)
 reports/network-intelligence/
+
+# Auto-managed by gbrain (db_only directories)
+media/x/
+media/articles/
+meetings/transcripts/
--- a/src/commands/doctor.ts
+++ b/src/commands/doctor.ts
@@ -5161,16 +5161,20 @@ export async function buildChecks(
 
   // 4. pgvector extension
   progress.heartbeat('pgvector');
-  try {
-    const sql = db.getConnection();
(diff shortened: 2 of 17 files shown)


time-attack:
Thanks for the effort and for the detailed verification list. I'm closing this because it isn't reviewable as a unit: it bundles doctor check suppressions, a new brain-score config key, a new embed re-chunking env knob, onboard categorization, an AI recipe change and engine source-config handling across nine source files with no single defect driving it, and it also adds your local brain directories (media/x/, meetings/transcripts/) to the repo .gitignore. If any one of these is a real bug on current master, a standalone PR scoped to that one defect with its regression test is the way in.

PR #2518, opened 2026-07-01
fix(postgres): persist checkpoint completed_keys as jsonb array

TurgutKural:
Summary:
- Fixes native Postgres checkpoint writes in `recordCompleted()`.
- Binds checkpoint keys as `text[]` and converts server-side with `to_jsonb($3::text[])`.
- Avoids JSON-stringified arrays being persisted as JSONB strings through the postgres.js unsafe/direct path.

Problem:
- `completed_keys` has a CHECK requiring JSONB array shape.
- Passing `JSON.stringify(sorted)` into `$3::jsonb` can produce a JSONB string on native Postgres driver binding paths.
- This causes checkpoint writes to fail and sync operations to become partial with `checkpoint_unavailable`.

Fix:
- Replace `$3::jsonb` + `JSON.stringify(sorted)` with `to_jsonb($3::text[])` + sorted string array.
- Keeps deterministic sorting and replace/upsert semantics.

Tests:
- `bun test test/op-checkpoint.test.ts`
- `bun run typecheck`
- `bun run check:jsonb`
- Native Postgres integration coverage was not available in the existing targeted test suite; this adds parameter-shape coverage and JSONB array-shape assertions without private environment details.

Security:
- No credentials, local paths, or environment values included.

Related:
- None.

### DIFF
--- a/src/core/op-checkpoint.ts
+++ b/src/core/op-checkpoint.ts
@@ -180,25 +180,20 @@ export async function recordCompleted(
   // extract-conversation-facts serialize a MUTABLE map through here and rely on
   // stale keys being REMOVED; an append would make them unremovable. The full
   // set lands in the parent `completed_keys` JSONB column via a single UPSERT.
-  // #2339: bind through `$3::text::jsonb`, NOT `$3::jsonb`. Under postgres.js
-  // `.unsafe(sql, params)` (executeRawDirect's path) a JS string bound to a
-  // `$N::jsonb` param double-encodes — the text→jsonb cast wraps the already-JSON
-  // string into a jsonb *string scalar*, which fails the v119
-  // `op_checkpoints_completed_keys_array CHECK (jsonb_typeof = 'array')` and aborts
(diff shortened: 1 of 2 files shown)


alexandreroumieu-codeapprentice:
This matches the failure class I saw on a real Postgres/Supabase-backed brain: `completed_keys` must land as a JSONB array, not a JSONB string scalar. The `to_jsonb($3::text[])` direction here matches the local hotfix I validated.

Additional validation from a local 0.42.57 rebase:

```text
bun test test/op-checkpoint.test.ts test/op-checkpoint-mentions-fingerprint.test.ts test/sync-resumable-import.serial.test.ts
# included in a targeted run: 66 pass, 0 fail
bun run typecheck
# OK
```

This PR's extra storage-shape tests are stronger than my earlier minimal parameter-shape guard, so I would prefer this upstream path over a duplicate PR.

TurgutKural:
solved with PR #2375

PR #2519, opened 2026-07-01
v0.42.54.0 fix(cli): supersede active takes (#2078)

### LINKED ISSUE #2078: takes supersede looks up its target with active:false — superseding any ACTIVE take fails "Row #N not found"
## Summary

`gbrain takes supersede` can never find an **active** take: `cmdSupersede` looks up the target with `engine.listTakes({ page_id: pageId, active: false, limit: 500 })` (takes.ts:274), and the engine treats `active: false` as `WHERE active = false`. Superseding any active take — the normal case; superseding is *for* active takes — fails with "Row #N not found".

## Repro (v0.42.37.0)

```
gbrain takes add <slug> --claim "..." --weight 0.8 ...   # creates active take, row N
gbrain takes supersede <slug> --row N --claim "..."      # → "Row #N not found"
```

## Workaround (verified twice)

```sql
UPDATE takes SET active = false WHERE id = <target>;
```
then run the CLI supersede — `engine.supersedeTake` itself doesn't check `active`, the fence renders correctly, and the end state is byte-identical to what a fixed CLI would produce.

## Fix

`active: true` in the lookup (or omit the filter and match by row number among all takes on the page, erroring on already-superseded targets with a clearer message).

javieraldape:
## Summary

Fixes #2078 by making `gbrain takes supersede` load the active target take before inheriting metadata. The previous CLI lookup asked for inactive rows, so the normal active-take supersede path reported `Row #N not found` before reaching the engine.

## Diff

- `src/commands/takes.ts`: changes the supersede preflight lookup from `active: false` to `active: true`.
- `test/takes-cli.test.ts`: adds a focused CLI regression that proves supersede reads the active target row, inherits kind/holder/weight, updates the DB mirror call, and rewrites the markdown fence.

## Test Coverage

- `gbrain-gate.sh <worktree> test/takes-cli.test.ts` - GREEN
  - unit: `test/takes-cli.test.ts` - 1 pass, 0 fail, 7 assertions
  - e2e: diff selector returned the fail-closed all-E2E set, so the gate skipped unrelated local E2E; CI runs the full E2E suite

### DIFF
--- a/src/commands/takes.ts
+++ b/src/commands/takes.ts
@@ -271,7 +271,7 @@ async function cmdSupersede(engine: BrainEngine, args: string[]): Promise<void> 
     const pageId = await getPageId(engine, slug);
 
     // Read existing row to inherit kind/holder unless overridden
-    const existing = await engine.listTakes({ page_id: pageId, active: false, limit: 500 });
+    const existing = await engine.listTakes({ page_id: pageId, active: true, limit: 500 });
     const target = existing.find(t => t.row_num === rowNum);
     if (!target) {
       console.error(`Row #${rowNum} not found on ${slug}.`);
--- a/test/takes-cli.test.ts
+++ b/test/takes-cli.test.ts
@@ -1,0 +1,110 @@
+import { describe, expect, test } from 'bun:test';
+import { mkdtempSync, mkdirSync, readFileSync, rmSync, writeFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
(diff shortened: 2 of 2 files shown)


time-attack:
Thank you for chasing #2078 down to the one-line cause — your diagnosis was exactly right. It has since landed on master as #3275 (`fix(takes): query active row before superseding`), which flips the same preflight lookup from `active: false` to `active: true`, so this change is now a no-op against current master. Closing as already fixed.

PR #2523, opened 2026-07-02
Treat local-path source clone state as not-applicable

sene1337:
## Summary

Fix `sources_status` clone diagnostics for user-supplied local-path sources.

Sources registered with `--path` have `local_path` but no `remote_url`; they are not gbrain-managed clones. Before this change, `getSourceStatus` still called `validateRepoState(local_path, undefined)`, so a healthy local wiki repo with no `origin` remote could be reported as `clone_state: "corrupted"`. That is misleading for remote MCP callers and can trigger the wrong remediation path.

This changes `getSourceStatus` to run clone validation only for remote-managed sources (`local_path && remote_url`). Pure DB sources and local-path sources now report `clone_state: "not-applicable"`.

## Verification

- `bun test test/sources-ops.test.ts` — 39 pass
- `bun run src/cli.ts call sources_status '{"id":"citadel-portfolio"}'` from Brad's patched local tree returns `clone_state: "not-applicable"` for a real local-path source with `remote_url: null`
- Scripted local verification checked 24 `remote_url: null` sources; all returned `clone_state: "not-applicable"`
- Added regression coverage that remote-managed clone origin-probe failure still reports `clone_state: "corrupted"`

Note: `bun run typecheck` passes in Brad's local carry branch after the patch. On a clean upstream-master worktree, typecheck currently fails in unrelated OpenClaw SDK typings (`src/core/context-engine.ts`, `test/e2e/openclaw-plugin-load-real.test.ts`), so I did not treat that as caused by this two-file change.

### DIFF
--- a/src/core/sources-ops.ts
+++ b/src/core/sources-ops.ts
@@ -127,9 +127,10 @@ export interface SourceStatus {
   last_commit: string | null;
   archived: boolean;
   /**
-   * Discriminated union from validateRepoState. 'not-applicable' if the
-   * source has no local_path (pure DB source). Lets a remote MCP caller
-   * diagnose "is the clone OK?" without SSH access to the brain host.
+   * Discriminated union from validateRepoState for gbrain-managed remote
+   * clones. 'not-applicable' for pure DB and user-supplied local-path sources
+   * (remote_url absent): those paths are not auto-managed clones, so a missing
+   * origin remote must not be reported as a corrupted clone.
    */
   clone_state: RepoState | 'not-applicable';
 }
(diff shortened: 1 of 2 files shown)


time-attack:
Merging. I confirmed the bug is still live on master: getSourceStatus runs validateRepoState for any source with a local_path, and validateRepoState returns 'corrupted' as soon as `git remote get-url origin` fails — so every source registered with --path (local_path set, remote_url null) is misreported as corrupted, which is exactly the kind of wrong-remediation signal we don't want reaching remote MCP callers. Scoping clone validation to remote-managed sources is the correct fix and the regression test that keeps 'corrupted' for real origin-probe failures is the right guard. Thanks for the careful writeup and the 24-source local verification.

PR #2525, opened 2026-07-02
fix(doctor): brain_score orphan/timeline components use the orphans-audit linkable scope

pabloglzg:
## The contradiction

On a brain with a large raw archive, one `gbrain doctor` report reads:

```
[OK] orphan_ratio: Orphan ratio 19% (110/567 linkable pages)
[OK] brain_score: Brain score 76/100 (embed 35/35, links 25/25, timeline 1/15, orphans 5/15, dead-links 10/10)
```

`orphans 5/15` implies ~67% of pages are orphans; `orphan_ratio` says 19%. Same word, two denominators: the orphans audit (`src/commands/orphans.ts`) excludes archive/generated pages by design (`shouldExclude`), while `getHealth()`'s score counted every page. Timeline coverage has the same flaw — raw transcripts are never going to carry curated timeline entries, and each one diluted the component.

## What

- **`src/core/linkable-scope.ts`** (new): the exclusion constants + predicate move here from `commands/orphans.ts`, which now re-exports `shouldExclude` unchanged. The JS predicate and the SQL fragments (`LINKABLE_EXCLUDE_*`) derive from the same constants so the two evaluation paths cannot drift.
- **`getHealth()` in both engines** computes `orphan_pages` and the timeline component over linkable pages via a `linkable_pages` CTE; `BrainHealth.linkable_page_count` carries the denominator. `page_count`, embed, link-density, and dead-links are untouched. `linkable_page_count === 0` gets full marks for those two components (same vacuous-truth rule as the v0.37.10.0 empty-brain fix).
- **Three archive conventions the audit was missing**, now in the shared defaults:
  - leading `raw/` — `includes('/raw/')` requires a mid-path segment, so brains keeping their archive at the repo root (`raw/whatsapp/...`) had every archive page counted as an orphan;
  - `outputs/` alongside `output/`;
  - first-segment `daily` — the pages the calendar/email integrations write are dated logs that link out, not in.

## Engine parity + tests

- PGLite: 3 new cases in `test/brain-score-breakdown.test.ts` (archive pages excluded from the orphan component; timeline coverage over linkable only; an islanded *curated* page still counts). 11 pass.
- Postgres: the new CTE validated against a live database (`NOT LIKE ALL($arr)` param form) — same numbers as PGLite for the same fixture.
- `test/orphans-pure-fn.test.ts`: +3 cases for the new defaults (leading `raw/` incl. the `rawhide/` non-match guard, `daily`, `outputs/`). 15 pass.
- `tsc --noEmit` clean; `test/pglite-engine.test.ts`, `test/doctor-behavioral.test.ts`, `test/features.test.ts`, `test/doctor-brain-checks-score.test.ts`, `test/brain-score-recommendations.test.ts` all pass (151 tests across the affected suites).

## Effect

brain_score becomes actionable for archive-heavy brains: the no-orphans and timeline components now measure the curated graph — the thing a user can actually fix — instead of being dominated by pages that are excluded from the orphan audit precisely because nothing is expected to link to them.

### DIFF
--- a/src/commands/orphans.ts
+++ b/src/commands/orphans.ts
@@ -15,6 +15,7 @@
 import type { BrainEngine } from '../core/engine.ts';
 import { createProgress, startHeartbeat } from '../core/progress.ts';
 import { getCliOptions, cliOptsToProgressOptions } from '../core/cli-options.ts';
+import { shouldExcludeFromLinkableScope } from '../core/linkable-scope.ts';
 
 // --- Types ---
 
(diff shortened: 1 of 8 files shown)


peterostrander2:
CI monitor found the current Test failure on this PR. The failing gate is `bun run verify` -> `tsc --noEmit` because `test/advisor-ranking-eval.test.ts` mock `getHealth()` objects are missing the now-required `BrainHealth.linkable_page_count` field.

Verified local fix after merging current `origin/master` into the PR branch:
- Add `linkable_page_count: 500` to both `getHealth()` fixture objects in `test/advisor-ranking-eval.test.ts`.
- `bun test test/advisor-ranking-eval.test.ts` passes: 2 pass / 0 fail.
- `bun run verify` passes: 30 checks green.

I attempted to push the tested patch back to `pabloglzg/gbrain:brain-score-linkable-scope`, but GitHub rejected it for `peterostrander2` with 403 (`Permission to pabloglzg/gbrain.git denied`).

time-attack:
Superseded with credit by #3155 (just merged) — your change landed rebased onto current master with authorship preserved via Co-authored-by. Thank you!

PR #472, opened 2026-04-27
v0.42.59.0 fix(files): normalize bigint sizes before JSON serialization

vinsew:
## Problem

Postgres returns `files.size_bytes` (`BIGINT`) as a native JavaScript `BigInt`. Returning that value directly from `file_list` makes JSON serialization fail, and using it in CLI arithmetic can also throw. PGLite returns a normal number, so PGLite-only coverage does not expose the bug.

## Fix

Normalize `size_bytes` to `Number` at the operation boundary and in the CLI display path. The exact-integer ceiling is far above any practical attachment size.

## Coverage

- Unit regression supplies native `BigInt` rows, preserves `null`, and proves the result is JSON-serializable.
- Mechanical Postgres E2E now asserts the wire value is a number.
- Rebuilt as one focused commit on `v0.42.59.0` / `master` (`5008b287`).

## Verification

- `bun test test/files.test.ts`: 24 passed, 0 failed
- `bun run typecheck`: passed
- `bun run verify`: passed
- Full `bun test`: passed
- Real-Postgres E2E was not rerun locally because the Docker daemon is unavailable; the E2E assertion remains in the PR for CI.

PR #2528, opened 2026-07-02
fix(files): handle bigint size_bytes in list output

### LINKED ISSUE #2527: gbrain files list crashes when Postgres returns size_bytes as bigint
## Summary

`gbrain files list` can crash on Postgres-backed brains because `files.size_bytes` is returned as a JavaScript `bigint`, but the CLI divides it by the numeric literal `1024`.

This throws:

```text
Invalid mix of BigInt and other type in division
```

## Repro

On a Postgres-backed GBrain with rows in the `files` table:

```bash
gbrain files list
```

or:

```bash
gbrain files list <page-slug>
```

## Root cause

In `src/commands/files.ts`, `listFiles()` assumes `row.size_bytes` is a `number | null`:

```ts
const sizeBytes = row.size_bytes as number | null;
const size = sizeBytes ? `${Math.round(sizeBytes / 1024)}KB` : '?';
```

With Postgres, `size_bytes` may arrive as `bigint`. Dividing a `bigint` by the number literal `1024` throws at runtime.

## Suggested fix

Normalize the DB value before formatting:

```ts
const rawSizeBytes = row.size_bytes as number | bigint | string | null;
const sizeBytes = rawSizeBytes == null ? null : Number(rawSizeBytes);
const size = sizeBytes == null || !Number.isFinite(sizeBytes) ? '?' : `${Math.round(sizeBytes / 1024)}KB`;
```

This also handles `0` correctly as `0KB` instead of treating it as falsy and printing `?`.

## Local verification

After applying the above fix on a Postgres-backed host:

```bash
gbrain files list
```

returned:

```text
37 file(s):
...
```

and:

```bash
gbrain files list products/gbrain
```

returned cleanly:

```text
No files for page: products/gbrain
```

## Related

This is related in theme to #2450, but it is a separate crash site: this one is arithmetic in `files list`, not JSON serialization.

techtony2018:
## Summary

- normalize `files.size_bytes` before formatting `gbrain files list` output
- handle Postgres `bigint` values without mixing BigInt and number arithmetic
- add focused coverage for number, bigint, string, null, invalid, and zero sizes

Fixes #2527.

## Verification

```bash
bun test test/files-command.test.ts
bun run typecheck
```

### DIFF
--- a/src/commands/files.ts
+++ b/src/commands/files.ts
@@ -16,7 +16,7 @@ interface FileRecord {
   filename: string;
   storage_path: string;
   mime_type: string | null;
-  size_bytes: number;
+  size_bytes: number | bigint | string | null;
   content_hash: string;
   metadata: Record<string, unknown>;
   created_at: string;
@@ -40,6 +40,12 @@ function getMimeType(filePath: string): string | null {
 function fileHash(filePath: string): string {
   const content = readFileSync(filePath);
   return createHash('sha256').update(content).digest('hex');
+}
+
+export function formatFileSizeKb(rawSizeBytes: number | bigint | string | null): string {
+  if (rawSizeBytes == null) return '?';
+  const sizeBytes = Number(rawSizeBytes);
+  return Number.isFinite(sizeBytes) ? `${Math.round(sizeBytes / 1024)}KB` : '?';
 }
 
 export async function runFiles(engine: BrainEngine, args: string[]) {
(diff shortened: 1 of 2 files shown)


time-attack:
Closing as superseded: #472 merged with the same size_bytes normalization (op handler + CLI display) plus a BigInt regression test, and #2892 added the general CLI output replacer. Thanks — the fix shape here was right.

PR #2530, opened 2026-07-02
fix(sync,sources,autopilot,jobs): repo paths absolute at ingress; refuse stored relative anchors

panda850819:
## Problem

Repo paths are persisted verbatim at four ingress points and re-resolved against cwd at read time. Once a relative value (typically `.`) lands in storage, every later bare invocation resolves it against whatever directory the process happens to run from.

Real incident (2026-07-02): `sync.repo_path` held `.` (seeded by `gbrain import .`). A bare `gbrain sync` from an unrelated project directory imported that tree (140 md files) as the brain source and reconciled 2,496 real brain pages as "source file removed", dropping their links/timeline graph rows. Files survived in git; db-only rows didn't.

Ingress points that persisted raw input:

- `import.ts` — `setConfig('sync.repo_path', dir)` with the raw CLI arg (the original seeder).
- `sync.ts` — `--repo` parsed raw; `writeSyncAnchor` stores whatever string flows in; the resolve phase uses a stored anchor relative to cwd.
- `autopilot.ts` — `--repo`/config used raw; `--install` bakes the raw string into the generated wrapper script (`exec gbrain autopilot --repo '.'`), freezing the bug into launchd/cron.
- `sources-ops.ts` — `addSource` stores `--path` verbatim; a relative path also defeats the string-prefix overlap check.

## Fix

Invariant (new `src/core/repo-path.ts`): **a repo path is resolved to an absolute path exactly once, at arg ingress; storage never holds a relative path; a relative value read back from storage is a hard error, never silently resolved against cwd.**

- `resolveRepoArg(p)` — resolve user-typed `--repo`/`--path` against cwd at parse time (typing a relative path interactively stays legal; persisting it doesn't).
- `requireAbsoluteStoredPath(value, storageDesc)` — throw with remediation (`gbrain sync --repo <abs>` / `gbrain config set sync.repo_path <abs>`) when a stored anchor is relative. Protects installs that already carry legacy `.` rows.

Wired at: sync CLI parse + `performSyncInner` resolve phase (single choke point — covers bare sync, `--source`, `--all`, cycle.ts, jobs.ts, MCP op) + `writeSyncAnchor`; autopilot run + `--install` (shared `resolveAutopilotRepoPath`); `addSource`; `import.ts` anchor write.

## Stored legacy rows (second review round)

A resolve-phase guard alone is bypassable: `sync --all`, the jobs `sync` handler, and a dozen-plus `getConfig('sync.repo_path')` readers fetch the stored path themselves and pass it as `opts.repoPath` (= caller intent, resolved not refused). Layered fix instead of per-site patches:

- **Migration 120 `repo_path_anchors_absolute_only`** — clears legacy relative anchors (config DELETE, `sources.local_path` → NULL; POSIX + Windows drive/UNC count as absolute). Every consumer falls into its existing explicit "no repo path — use --repo / sources add --path" error; the next `gbrain sync --repo <abs>` re-seeds normalized.
- **`sync --all`** — sources with a relative `local_path` are excluded loudly before the cost estimator and fan-out (pre-migration window).
- **jobs `sync` handler** — `job.data.repoPath` is persisted payload; a relative value fails the job with remediation instead of resolving against the worker daemon's cwd.

## Clone dir (third review round)

A third round caught `sources add --url --clone-dir <relative>`: the URL branch overwrote the normalized path with raw `opts.cloneDir`. Now resolved like `--path` (`defaultCloneDir` is `$GBRAIN_HOME`-based, already absolute).

## Tests

`test/repo-path.test.ts` (13 tests, PGLite harness) + 1 in `test/sources-resync-recovery.test.ts`:

- helper behavior (resolve against cwd, absolute passthrough, refusal message names the storage key + remediation),
- `addSource` stores absolute for relative `--path`,
- regression shape: seeded `sync.repo_path='.'` → bare `performSync` throws at `resolve_repo`; seeded legacy relative `sources.local_path` → per-source sync throws naming the source; caller-supplied relative `repoPath` still resolves (not refused),
- migration 120: relative anchors cleared; POSIX / `C:\` / UNC absolutes survive,
- relative `--clone-dir` through the real (fake-git) clone flow stores an absolute row.

`bun run typecheck` clean; full unit suite: 9419 pass, 7 pre-existing environment-dependent failures reproduced identically on a clean master checkout (live `~/.gbrain` state, `OPENAI_API_KEY` present, global `tag.gpgsign=true` breaking lightweight-tag fixtures) — none related to this diff.

Deployed on the incident machine: migration applied cleanly, and re-seeding the incident state (`sources.local_path = '.'`) now refuses at `sync.resolve_repo` with the remediation message instead of importing the foreign cwd.

https://claude.ai/code/session_01AeWThyfkWsf2tHTD2NzJD8

### DIFF
--- a/src/commands/autopilot.ts
+++ b/src/commands/autopilot.ts
@@ -38,6 +38,24 @@ import { logSelfUpgrade } from '../core/audit/self-upgrade-audit.ts';
 import { detectInstallMethod } from './upgrade.ts';
 import { evaluateQuietHours } from '../core/minions/quiet-hours.ts';
 import { inspectLock } from '../core/db-lock.ts';
+import { resolveRepoArg, requireAbsoluteStoredPath } from '../core/repo-path.ts';
+
+/**
+ * Shared --repo resolution for the run loop and --install. A typed --repo is
+ * resolved against cwd at ingress; a config-sourced value must already be
+ * absolute (repo-path.ts invariant) — --install bakes it verbatim into the
+ * wrapper script, so a relative value here would freeze the cwd bug into
+ * launchd/cron.
+ */
+async function resolveAutopilotRepoPath(
+  engine: BrainEngine,
+  args: string[],
+): Promise<string | null> {
+  const arg = parseArg(args, '--repo');
+  if (arg) return resolveRepoArg(arg);
+  const stored = await engine.getConfig('sync.repo_path');
+  return stored ? requireAbsoluteStoredPath(stored, 'config sync.repo_path') : null;
+}
 
 /**
  * v0.37.7.0 #1162 — classify autopilot reconnect-loop errors.
@@ -331,7 +349,7 @@ export async function runAutopilot(engine: BrainEngine, args: string[]) {
     return;
   }
 
-  const repoPath = parseArg(args, '--repo') || await engine.getConfig('sync.repo_path');
+  const repoPath = await resolveAutopilotRepoPath(engine, args);
   const baseInterval = parseInt(parseArg(args, '--interval') || '300', 10);
   const jsonMode = args.includes('--json');
   const forceInline = args.includes('--inline');
@@ -1142,7 +1160,7 @@ exec '${safeGbrainPath}' autopilot --repo '${safeRepoPath}'
 }
 
 async function installDaemon(engine: BrainEngine, args: string[]) {
-  const repoPath = parseArg(args, '--repo') || await engine.getConfig('sync.repo_path');
+  const repoPath = await resolveAutopilotRepoPath(engine, args);
   if (!repoPath) {
     console.error('No repo path. Use --repo or run gbrain sync --repo first.');
     process.exit(1);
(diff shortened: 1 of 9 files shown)


time-attack:
Thanks — merging this. I confirmed the footgun is still live on master (`import.ts` persists the raw arg into `sync.repo_path`, and every downstream reader re-resolves it against its own cwd), and the failure mode is DB-row loss, not just a bad import, so it clears the bar for a change this size. I particularly like that stored relative anchors hard-error with remediation instead of being silently rewritten, and that the migration clears rather than guesses. One mechanical note: your migration is numbered 120 and master is now at 125, so it needs renumbering when you rebase.

garrytan:
Follow-up: this lands (rebased) in #3921 (v0.42.80.0) — the approval here was right and the merge never happened. The re-land keeps your design (absolute-at-ingress + read-time refusal + legacy-anchor migration) with the migration renumbered for the current slot and remediation hints parameterized per storage site. Credited in the release CHANGELOG — thanks @panda850819.

PR #2531, opened 2026-07-02
v0.42.54.0 fix(cli): honor search --json output (#2042)

### LINKED ISSUE #2042: search --json emits empty stdout when engine.disconnect() hangs and CLI force-exits
## Symptom
`gbrain search "<query>" --json` exits 0 with completely empty stdout, while the same query works in plain mode and via `gbrain query`.
stderr shows: `[cli] engine.disconnect() did not return within 10000ms — force-exiting`

## Environment
gbrain 0.42.37.0, Postgres engine via Supabase transaction pooler (6543), search.mode=balanced. Observed immediately after a bulk `extract links --by-mention --source db` write (~4.3K mention edges on a ~1.5K-page brain) — possibly the denser links table slows whatever disconnect waits on, widening the race.

## Suspected cause
When the disconnect watchdog force-exits, the process dies before the --json payload flushes to stdout. Plain-mode output (written incrementally) survives; the JSON payload (written late, one chunk) is lost. Exit 0 makes it silent — an agent consuming --json sees "no results" and trusts it.

## Suggested fix
Flush/drain stdout before the force-exit path (or write payload before initiating disconnect). Related lineage: v0.41.27.0 disconnect work (#1570) + the facts-queue drain-before-disconnect fix — same class, output side.

javieraldape:
## Summary

Fixes #2042 by making the shared `search` / `query` CLI formatter honor the CLI-local `--json` flag. The result payload is now emitted as a parseable JSON array before teardown, so agents consuming piped stdout no longer fall back to human text or an empty-looking result.

## Diff

- `src/cli.ts`: threads parsed CLI params into the shared operation formatter for both local and routed calls, treats `--json` as a CLI-only formatter flag, and renders `search` / `query` results as JSON when requested.
- `test/cli-args.test.ts`: pins `--json` parsing as a formatter flag that does not enter the public operation contract.
- `test/cli-format-search-json.test.ts`: covers non-empty `search --json` and empty `query --json` rendering as machine-readable arrays.
- `test/cli-search-dispatch.test.ts`: adds a real CLI subprocess regression for `gbrain search "<query>" --json` on an empty brain.

## Test Coverage

- `gbrain-gate.sh <worktree> test/cli-args.test.ts test/cli-format-search-json.test.ts test/cli-search-dispatch.test.ts` - GREEN
  - unit: 8 pass, 0 fail, 15 assertions across the three selected test files
  - e2e: selector returned the fail-closed all-E2E set, so the gate skipped unrelated local E2E; CI runs the full E2E suite

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -443,7 +443,7 @@ async function main() {
     // routed path. Date → ISO string; bigint → string (postgres.js shape);
     // Buffer → object. Microsecond-cost; eliminates a whole drift bug class.
     const result = JSON.parse(JSON.stringify(rawResult));
-    const output = formatResult(op.name, result);
+    const output = formatResult(op.name, result, params);
     if (output) process.stdout.write(output);
   } catch (e: unknown) {
     // v0.42.20.0 (codex D4): on error, set exitCode + return so the `finally`
@@ -526,7 +526,7 @@ async function runThinClientRouted(
       signal: sigintController.signal,
     });
     const result = unpackToolResult(raw);
-    const output = formatResult(op.name, result);
+    const output = formatResult(op.name, result, params);
     if (output) process.stdout.write(output);
   } catch (e: unknown) {
     if (e instanceof RemoteMcpError) {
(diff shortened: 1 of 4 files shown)


time-attack:
Thanks for the tests and the clean threading — but this targets #2042, which was closed on 2026-07-21, and the root cause there was the disconnect watchdog force-exiting before stdout flushed (fixed by the bounded-teardown work in #2084/#2141), not the formatter. What is left in this diff is a new `search`/`query` JSON output surface that master does not currently advertise, added by smuggling a CLI-local `json` key into the params object handed to operation handlers, which cuts against contract-first. Closing on both counts.

Issue #2484, opened 2026-06-29
extract_facts dream phase permanently jams on structurally-unfenceable hot-memory rows (guard predicate over-matches)


victorfteha:
**Version:** v0.42.37.0, Postgres engine. Behavior in the guard code is unchanged on master (v0.42.40).

## Summary

The `extract_facts` cycle phase has an empty-fence guard ([core/cycle/extract-facts.ts:107-125](https://github.com/garrytan/gbrain/blob/main/src/core/cycle/extract-facts.ts)) that refuses to run the destructive fence→DB reconcile while "legacy v0.31" rows exist:

```sql
SELECT COUNT(*) FROM facts WHERE row_num IS NULL AND entity_slug IS NOT NULL
```

`> 0` → `guardTriggered`, warn, **early-return for the whole phase** (no phantom-redirect pass, no reconcile). The warning tells the operator to `gbrain apply-migrations --yes` to finish v0_32_2.

The predicate assumes every `row_num IS NULL AND entity_slug IS NOT NULL` row is a *pre-v0.32.2* fact awaiting the one-time fence backfill. But the **inline facts pipeline keeps producing rows of exactly that shape after the migration has completed**, and they can never satisfy the predicate's exit condition — so the phase jams forever.

## Root cause

`runFactsPipeline` (`core/facts/backstop.ts`), used by the `extract_facts` MCP op (`core/operations.ts:~3669`) and other inline writers, is fence-first **only when the resolved `entity_slug` maps to a fenceable page**. When the entity ref resolves to a slug with no corresponding page at the source's `local_path` (free-form hint, non-canonical slug, or an entity that simply has no page yet), the pipeline documents a deliberate fallback to a **DB-only insert** with `row_num` left NULL (`backstop.ts:~394`, `~413`, `~476` — "legacy DB-only fallback for unparented / thin-client facts").

Those rows have `entity_slug IS NOT NULL` (the hint resolved to *some* string) but `row_num IS NULL` (never fenced) — i.e. they match the guard predicate. They are **not** pre-migration legacy rows; they're freshly written post-migration, and `apply-migrations` will not clear them:
- The v0_32_2 backfill ([commands/migrations/v0_32_2.ts](https://github.com/garrytan/gbrain/blob/main/src/commands/migrations/v0_32_2.ts) Phase B) only fences rows whose `entity_slug` maps to a writable page; an entity_slug with no page is structurally unfenceable.
[...]

PR #2532, opened 2026-07-02
Prevent MCP facts fallback from tripping extract guard

sene1337:
## Summary

- Prevent MCP-originated DB-only fallback facts from recreating the legacy `row_num IS NULL AND entity_slug IS NOT NULL` guard shape.
- Treat fallback-slugified entity guesses as unresolved so MCP extraction does not create phantom entity pages.
- Add regression coverage for both fallback-slugified MCP entities and thin-client/no-`local_path` MCP fallback.

## Why

When `extract_facts` runs through MCP and the markdown fence path is unavailable or blocked by the stub guard, the fallback direct DB insert could retain `entity_slug` with `row_num = NULL`. Those rows trip the deterministic extract-facts reconciliation guard and can halt future extraction. MCP fallback rows should remain queryable by recency/session/text, but not claim an entity fence row that was never written.

## Verification

- `bun test test/facts-backstop-integration.test.ts test/extract-facts-phase.test.ts test/fence-write.test.ts` — 35 pass
- `bun run typecheck` — pass

### DIFF
--- a/src/core/facts/backstop.ts
+++ b/src/core/facts/backstop.ts
@@ -119,6 +119,20 @@ function warnOnce(key: string, msg: string): void {
   // eslint-disable-next-line no-console
   console.warn(msg);
 }
+
+/**
+ * DB-only fallback is a degraded path. For MCP-originated facts, keep it
+ * from recreating the legacy guard shape (`row_num IS NULL AND
+ * entity_slug IS NOT NULL`) by dropping the entity link when the fact could
+ * not be written to a markdown fence. The fact remains queryable by recency
+ * / session / text, but it will not halt the deterministic extract_facts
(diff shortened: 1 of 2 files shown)


time-attack:
Thanks for the detailed reproduction and tests. The jam you describe was addressed on master by #2484, which tightened the empty-fence guard to only count rows whose `entity_slug` resolves to a live page — precisely so the inline writer's unfenceable `row_num IS NULL` rows stop gating the phase. Beyond that, dropping `entity_slug` for all MCP-originated fallback facts permanently discards the entity association on the main agent path to keep a guard quiet, which trades a real data-quality loss for a symptom; if a residual thin-client case remains, the fix belongs in the guard predicate or the fence path. Closing.

PR #2534, opened 2026-07-02
Prevent MCP facts fallback from tripping extract_facts guard

sene1337:
## Summary

Fixes a recurrence class where MCP-originated fact captures can fall through to the DB-only legacy path with `row_num IS NULL AND entity_slug IS NOT NULL`, which trips the deterministic `extract_facts` phase guard and can block future fence reconciliation.

## What changed

- Treat fallback-slugified MCP entity guesses as unresolved instead of creating/attaching synthetic entities.
- When MCP-originated facts (`mcp:extract_facts` / `mcp:put_page`) cannot be written to a markdown facts fence and must use DB-only fallback, drop the entity link so they do not recreate the legacy guard shape.
- Preserve recency/session/text recall for degraded fallback facts.
- Add integration regressions for:
  - fallback-slugified MCP entities with a configured `local_path`;
  - thin-client/no-`local_path` MCP fallback for an otherwise canonical entity.

## Local verification

Run on Brad's local checkout before opening PR:

- `bun test test/facts-backstop-integration.test.ts test/extract-facts-phase.test.ts test/fence-write.test.ts` → 35 pass, 0 fail.
- `bun run typecheck` → pass.
- `bun run build` → pass.
- `bun run verify` → 30 checks green.

Also verified on Brad's live GBrain DB after the companion local config fix:

- `default.local_path` no longer aliases `citadel-portfolio`.
- SQL guard check for `row_num IS NULL AND entity_slug IS NOT NULL` returns 0.
- MCP smoke inserted a temporary synthetic fact with `entity_slug: null`, then cleanup left guard rows at 0.

### DIFF
--- a/src/core/facts/backstop.ts
+++ b/src/core/facts/backstop.ts
@@ -119,6 +119,20 @@ function warnOnce(key: string, msg: string): void {
   // eslint-disable-next-line no-console
   console.warn(msg);
 }
+
+/**
+ * DB-only fallback is a degraded path. For MCP-originated facts, keep it
+ * from recreating the legacy guard shape (`row_num IS NULL AND
+ * entity_slug IS NOT NULL`) by dropping the entity link when the fact could
+ * not be written to a markdown fence. The fact remains queryable by recency
+ * / session / text, but it will not halt the deterministic extract_facts
+ * reconciliation cycle.
+ */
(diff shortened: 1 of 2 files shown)


time-attack:
Closing as duplicate of #2532.

Same fix: prevent MCP facts fallback from tripping the extract_facts guard. Keep #2532.

PR #2545, opened 2026-07-03
fix(gbrain): avoid harden hook push race

peterostrander2:
## Summary
- make hardenBrainRepo's internal scaffolding commit use --no-verify
- avoids the freshly installed post-commit hook racing harden's explicit synchronous push

## Verification
- bun test test/brain-durability-hook.serial.test.ts (5 consecutive runs)
- bun run verify (31 checks green)

CI monitor source: Test run 28610422792 failed in serial-tests on test/brain-durability-hook.serial.test.ts with git index.lock during note.md add.


time-attack:
Merging — thanks for chasing this down to the actual race rather than retrying the test. Confirmed on master that `commitScaffolding` commits without `--no-verify` and then pushes synchronously, so harden's own scaffolding commit triggers the post-commit hook it installed moments earlier; `--no-verify` on that one internal commit is exactly the right scope, and it leaves the committed helper's durability guarantee alone. Nice one-line fix with a real CI receipt behind it.

Issue #239, opened 2026-04-20
v0.13.1 schema bootstrap fails on pre-v11 brains: CREATE INDEX idx_links_source on column not yet added


jsclancy137:
Hit this on PGLite upgrading from schema v10 → v13. Reproduces 100% on any brain initialized before v0.13.

## Symptom

```
$ gbrain init
Setting up local brain with PGLite (no server needed)...
column "link_source" does not exist
```

## Cause

`PGLiteEngine.initSchema()` runs `PGLITE_SCHEMA_SQL` BEFORE `runMigrations(this)`. The schema SQL contains:

```sql
CREATE TABLE IF NOT EXISTS links (
  ...
  link_source    TEXT    CHECK (link_source IS NULL OR link_source IN ('markdown', 'frontmatter', 'manual')),
  ...
);

CREATE INDEX IF NOT EXISTS idx_links_source ON links(link_source);
```

On a fresh DB, the `CREATE TABLE IF NOT EXISTS` creates the table with `link_source`. Fine.

On a pre-v0.13 DB (schema v10), the table already exists WITHOUT `link_source`, so `CREATE TABLE IF NOT EXISTS` no-ops, leaving the table without the column. Then `CREATE INDEX IF NOT EXISTS idx_links_source ON links(link_source)` fails because the column doesn't exist yet.

The migration v11 (`links_provenance_columns`) that adds the column runs AFTER `initSchema()`, but `initSchema()` already crashed.

## Reproduce

1. Initialize a brain on gbrain ≤ 0.12.x
2. Upgrade to gbrain 0.13.0 or 0.13.1
3. `gbrain init` → fails

## Fix options

(a) Move `CREATE INDEX idx_links_source` (and any other v0.13-column-dependent indexes) out of `pglite-schema.ts` and into the migration v11 SQL.

(b) Run migrations BEFORE `PGLITE_SCHEMA_SQL` (re-orderable since the schema SQL is `IF NOT EXISTS` everywhere).

(c) Wrap the index creation in a `DO $$ BEGIN IF EXISTS (column) THEN CREATE INDEX ...; END IF; END $$;` guard.

I think (a) is cleanest — schema.ts should reflect the *fresh-install* shape; column additions and their indexes should live in their migration.

## Workaround

Wipe `~/.gbrain/brain.pglite` and re-import. Loses any in-DB-only state but rebuilds from the source vault.

## Env

- macOS 26.4.1 arm64
[...]

PR #2548, opened 2026-07-03
fix(schema): bootstrap timeline_entries.event_page_id (v121) forward-reference

chetan-guevara:
## Problem

`applyForwardReferenceBootstrap` (both engines) pre-creates columns that the embedded schema blob (`SCHEMA_SQL` / `PGLITE_SCHEMA_SQL`) forward-references but that older brains lack — it covers ~20 such columns. **It's missing the v121 entry.**

`SCHEMA_SQL` has:
```sql
CREATE INDEX IF NOT EXISTS idx_timeline_event_page  ON timeline_entries(event_page_id) WHERE event_page_id IS NOT NULL;
CREATE UNIQUE INDEX IF NOT EXISTS idx_timeline_event_dedup ON timeline_entries(event_page_id, date) WHERE event_page_id IS NOT NULL;
```
but `timeline_entries.event_page_id` is only added by **migration v121** (`timeline_entries_event_page_id`).

On a brain whose `timeline_entries` predates v121, `CREATE TABLE IF NOT EXISTS` is a no-op (the column is never added), so the schema-blob's `CREATE INDEX` throws **`column "event_page_id" does not exist`** during `initSchema()` — *before* `runMigrations()` can apply v121. Result: both `gbrain init --migrate-only` and `gbrain apply-migrations` wedge, the schema version stays behind (e.g. stuck at 120), and every query prints `Schema probe failed: column "event_page_id" does not exist`.

This is the exact wedge class the `schema-bootstrap-coverage` guard was built to prevent (incidents #239/#243/#266/#357/#366/#374/#375/#378/#395/#396). It slipped through because the static A2 check treats a column present in the current `CREATE TABLE` body as covered — which is true for *fresh* installs but not for *upgrade* brains where the `CREATE TABLE IF NOT EXISTS` no-ops — and `event_page_id` was never added to `REQUIRED_BOOTSTRAP_COVERAGE` nor stripped by the runtime tests.

### Observed live
A Postgres brain upgraded across the v121 boundary (image rebuild pulled a newer gbrain). `gbrain query` degraded to keyword-only with the schema-probe warning; `init --migrate-only` and `apply-migrations` both failed identically on `event_page_id`. Manually `ALTER TABLE timeline_entries ADD COLUMN IF NOT EXISTS event_page_id INTEGER` then `init --migrate-only` recovered it (v121+v122 applied cleanly) — i.e. exactly what the missing bootstrap should have done.

## Fix

- **`postgres-engine.ts` + `pglite-engine.ts`**: probe `timeline_entries.event_page_id` and, when the table exists without it, `ALTER TABLE timeline_entries ADD COLUMN IF NOT EXISTS event_page_id INTEGER` before the schema-blob replay. The guarded FK + partial indexes still land via the blob (`CREATE INDEX IF NOT EXISTS`) and the idempotent v121 migration — bootstrap only unblocks the replay, mirroring every other entry (e.g. v112 `links_extracted_at`).
- **`schema-bootstrap-coverage.test.ts`**: add the `REQUIRED_BOOTSTRAP_COVERAGE` entry and strip `event_page_id` + its indexes/FK in both runtime tests so the wedge is reproduced and guarded going forward.

## Verification (local, bun 1.3.10)

- `bun test test/schema-bootstrap-coverage.test.ts` → **9 pass / 0 fail** (migration chain applies v121→v122; bootstrap runs).
- **Negative check** — disabling the new bootstrap block reproduces the failure: the coverage assertion fails **and** `PGLITE_SCHEMA_SQL` replay throws `column "event_page_id" does not exist`. Restoring the block makes it green again. So the test genuinely guards this regression.
- `tsc --noEmit` clean for the edited files.

Additive + idempotent; no-op on fresh installs and on brains already at/after v121.

### DIFF
--- a/src/core/pglite-engine.ts
+++ b/src/core/pglite-engine.ts
@@ -498,7 +498,11 @@ export class PGLiteEngine implements BrainEngine {
         EXISTS (SELECT 1 FROM information_schema.columns
                 WHERE table_schema='public' AND table_name='pages' AND column_name='embedding_signature') AS pages_embedding_signature_exists,
         EXISTS (SELECT 1 FROM information_schema.columns
-                WHERE table_schema='public' AND table_name='pages' AND column_name='links_extracted_at') AS pages_links_extracted_at_exists
+                WHERE table_schema='public' AND table_name='pages' AND column_name='links_extracted_at') AS pages_links_extracted_at_exists,
+        EXISTS (SELECT 1 FROM information_schema.tables
+                WHERE table_schema='public' AND table_name='timeline_entries') AS timeline_entries_exists,
+        EXISTS (SELECT 1 FROM information_schema.columns
+                WHERE table_schema='public' AND table_name='timeline_entries' AND column_name='event_page_id') AS timeline_event_page_id_exists
     `);
     const probe = rows[0] as {
       pages_exists: boolean;
@@ -541,6 +545,8 @@ export class PGLiteEngine implements BrainEngine {
       pages_generation_exists: boolean;
       pages_embedding_signature_exists: boolean;
       pages_links_extracted_at_exists: boolean;
+      timeline_entries_exists: boolean;
+      timeline_event_page_id_exists: boolean;
     };
 
     const needsPagesBootstrap = probe.pages_exists && !probe.source_id_exists;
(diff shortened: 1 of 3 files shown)


alexandreroumieu-codeapprentice:
Confirmed this exact upgrade wedge on an existing Postgres/Supabase brain during a 0.42.53 -> 0.42.57 update.

Observed failure before the bootstrap fix:

```text
column "event_page_id" does not exist
```

The failure happened before migration v121 could repair the existing `timeline_entries` table, because schema replay attempted to create indexes referencing `timeline_entries(event_page_id)` first.

After adding an idempotent bootstrap for `timeline_entries.event_page_id` before the timeline event indexes, the same brain completed:

```text
gbrain init --migrate-only
# applied migrations 119 -> 122

gbrain sync --repo <brain-repo>
# OK / already up to date

gbrain embed --stale
# 0 stale found
```

Targeted local verification on the rebased 0.42.57 tree:

```text
bun test test/schema-bootstrap-coverage.test.ts test/op-checkpoint.test.ts test/op-checkpoint-mentions-fingerprint.test.ts test/sync-resumable-import.serial.test.ts test/schema-pack-base-v2-phases.test.ts test/doctor-extract-atoms-backlog.test.ts
# 66 pass, 0 fail
bun run typecheck
# OK
```

This supports the engine-level forward-reference bootstrap approach here. It is also why I would not open another duplicate PR for the same issue.

time-attack:
Closing during backlog triage: the timeline_entries.event_page_id forward-reference bootstrap fix already landed on master. Thanks for catching it!

PR #2549, opened 2026-07-03
v0.42.57.0 fix(cycle): stamp path-derived dream sources (#1869)

### LINKED ISSUE #1869: gbrain dream --dir <path> never stamps cycle_freshness (only --source does) → brain shows perpetually stale
## Summary
`gbrain dream --dir <path>` completes a cycle (exit 0, real work) but **never stamps per-source `cycle_freshness`**, so `gbrain doctor` reports the brain as perpetually stale even though it's being dreamed on schedule. `gbrain dream --source <id>` DOES stamp it.

## Environment
- Linux/WSL, PGLite engine, gbrain 0.42.x, single source `default`.

## Symptom
- A daily cron running `gbrain dream --dir ~/brain` succeeds every night (exit 0, propose_takes/consolidate/etc. all run).
- Yet `gbrain doctor` shows `[FAIL] cycle_freshness → Source 'default' last cycled NNh ago` indefinitely (observed 61h).

## Repro
1. `gbrain dream --dir ~/brain` → exit 0, "Dream cycle (partial)".
2. `gbrain doctor` → `cycle_freshness` still FAIL / unchanged.
3. `gbrain dream --source default` → exit 0.
4. `gbrain doctor` → `[OK] cycle_freshness: All sources cycled recently`. ✅

## Impact
Highly misleading: operators see "brain hasn't dreamed in N hours" and assume dreaming is broken, when it's running fine nightly — the `--dir` invocation just doesn't update the source's `last_cycled_at`. Cost us a long debugging detour.

## Suggested fix
Either stamp `cycle_freshness` on the `--dir` path too (resolve `--dir` to its source and update `last_cycled_at`), or have `--dir` print a clear notice that it does not stamp per-source freshness (and point to `--source`). Ideally the docs/`doctor` remediation text should mention this distinction.

javieraldape:
## Summary

Fixes #1869 by letting `gbrain dream --dir <path>` reuse the source id already derived from `sources.local_path` when it finishes a successful cycle. Path-scoped dream runs now stamp the same per-source freshness fields as explicit `--source` runs, so doctor no longer keeps reporting a stale cycle for that source.

## Diff

- `src/core/cycle.ts`: uses the resolved `cycleSourceId` for the DB cycle lock and for the best-effort `last_source_cycle_at` / `last_full_cycle_at` update.
- `src/commands/dream.ts`: updates the `--source` help text to describe the new `--dir` / configured-checkout derivation behavior.
- `test/cycle-last-full-cycle-at.test.ts`: covers no-`sourceId` cycle runs that should stamp when `brainDir` matches a source row, plus the still-skipped unmatched-dir case.
- `tsconfig.json` + `test/plugin-sdk-compat.d.ts`: adds a tiny test-only SDK compatibility shim so local typecheck remains deterministic when a host-installed plugin SDK is present.

## Test Coverage

- `gbrain-gate.sh <worktree> test/cycle-last-full-cycle-at.test.ts test/dream-cli-flags.test.ts` - GREEN
  - unit: 28 pass, 0 fail, 54 assertions across the two selected test files
  - e2e: selector returned the fail-closed all-E2E set, so the gate skipped unrelated local E2E; CI runs the full E2E suite

### DIFF
--- a/src/commands/dream.ts
+++ b/src/commands/dream.ts
@@ -319,9 +319,9 @@ Options:
 
   --source <id>       Scope the cycle to one source so doctor's
                       cycle_freshness check sees a fresh stamp on
-                      completion. Without this, gbrain dream's
-                      timestamp never lands and federated brains
-                      see "stale cycle" forever.
+                      completion. When omitted, gbrain derives the
+                      source from --dir / configured checkout when
+                      it matches sources.local_path.
   --source-id <id>    Alias for --source. Matches the v0.37.7.0+
                       naming used by import/extract/graph-query.
 
--- a/src/core/cycle.ts
+++ b/src/core/cycle.ts
@@ -1481,10 +1481,11 @@ export async function runCycle(
 
       let dbLock: LockHandle | null = null;
       try {
-        // v0.38: per-source lock ID when opts.sourceId is set; legacy
-        // `gbrain-cycle` otherwise (autopilot still passes nothing).
+        // v0.38: per-source lock ID when a source is known; legacy
+        // `gbrain-cycle` otherwise. `cycleSourceId` includes explicit
+        // opts.sourceId and the --dir/local_path-derived source.
         // cycleLockIdFor validates the sourceId via assertValidSourceId.
-        dbLock = await acquireDbCycleLock(engine, opts.sourceId);
+        dbLock = await acquireDbCycleLock(engine, cycleSourceId);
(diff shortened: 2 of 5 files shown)


time-attack:
Superseded with credit by #3178 (just merged) — your change landed rebased onto current master with authorship preserved via Co-authored-by. Thank you!

PR #2550, opened 2026-07-03
feat(ai): add OpenCode Zen provider

FilipHarald:
## Summary
- Adds a first-class `opencode` OpenAI-compatible provider recipe for OpenCode Zen.
- Maps `opencode_api_key` to `OPENCODE_API_KEY` in gateway config.
- Extends provider registry and regression coverage so `opencode:<model>` works for chat/subagent-capable models.

## Verification
- `bun test test/ai/gateway-chat.test.ts test/ai/recipes-existing-regression.test.ts test/ai/build-gateway-config.test.ts test/model-config.serial.test.ts`
- Live probe against OpenCode Zen: `bun src/cli.ts providers test --touchpoint chat --model opencode:deepseek-v4-flash`

### DIFF
--- a/src/core/ai/build-gateway-config.ts
+++ b/src/core/ai/build-gateway-config.ts
@@ -9,7 +9,7 @@
  * import it from `../../src/cli.ts`.
  *
  * The single ownership site for: (a) folding file-plane API keys
- * (openai/anthropic/zeroentropy) into the gateway env, and (b) threading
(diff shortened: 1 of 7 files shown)


FilipHarald:
There seems to be quite few recipies in repo. Maybe it is intentional to reduce noise? In that case I'm guessing this PR will just be closed :sweat_smile:

time-attack:
Reviewed and declined: micro-vendor recipes are out of scope; the openai-compat generic path covers this. Thank you.

PR #2566, opened 2026-07-03
fix(embed): allow user-provided-model recipes (llama-server, litellm) to pass dim validation and preflight

felixmw:
## Summary

Two bugs make local embedding providers (`llama-server`, `litellm`) completely unusable for embeddings:

- **`embedding-dim-check.ts`**: `validateDimAgainstTouchpoint` rejects any `--embedding-dimensions` value for user-provided-model recipes. `defaultDims` is `0` (sentinel for "user must supply"), so any provided value differs from `0` and triggers Matryoshka validation — which fails because `llama-server`/`litellm` aren't in any known provider allow-list. Fix: skip custom-dim validation when `defaultDims === 0`.

- **`gateway.ts`**: `diagnoseEmbedding` blanket-rejects all user-provided-model recipes with `user_provided_model_unset` even when a model name IS configured (e.g. `llama-server:bge-m3-mlx-fp16`). The check only tested `tp.models.length === 0` without verifying `parsed.modelId`. Fix: add `!parsed.modelId` guard.

Together these prevented `gbrain init --embedding-model llama-server:<model> --embedding-dimensions <N>` and `gbrain embed --stale` from succeeding with any local OpenAI-compatible embedding server.

## Test plan

- [x] Added 3 test cases to `test/embedding-dim-check.test.ts` covering llama-server and litellm dim resolution
- [x] Added 2 test cases to `test/embed-preflight.test.ts` covering llama-server and litellm preflight pass
- [x] All 43 tests pass (38 existing + 5 new)
- [ ] Verify `gbrain init --pglite --embedding-model llama-server:<model> --embedding-dimensions 1024` succeeds
- [ ] Verify `gbrain embed --stale` succeeds with a running llama-server endpoint

### DIFF
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -705,11 +705,16 @@ export function diagnoseEmbedding(modelOverride?: string): EmbeddingDiagnosis {
   }
 
   // Openai-compat recipes with empty models list require a user-provided model.
+  // Only reject when the user hasn't actually provided a model name — i.e. the
+  // config says `llama-server:` with no model id after the colon. When a model
+  // id IS present (e.g. `llama-server:bge-m3-mlx-fp16`), the user has fulfilled
+  // the requirement and we should let the embed pipeline proceed.
   const isUserProvided = (tp as any).user_provided_models === true;
   if (
     Array.isArray(tp.models) &&
     tp.models.length === 0 &&
-    (recipe.id === 'litellm' || isUserProvided)
+    (recipe.id === 'litellm' || isUserProvided) &&
+    !parsed.modelId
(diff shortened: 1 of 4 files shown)


time-attack:
Thank you for the careful write-up — the diagnosis was right, but both halves have since landed on master. Custom dims for llama-server/litellm/ollama now pass via the `trust_custom_dims` passthrough tier added in #2271, and the `user_provided_model_unset` guard in `gateway.ts` was removed for exactly the false-positive you describe. Closing as already fixed; please re-open an issue if `gbrain init --embedding-model llama-server:<model>` still fails on latest master.

PR #2582, opened 2026-07-04
feat(cycle): add dream.patterns.cooldown_hours spend cap (mirror synthesize)

spiky02plateau:
## What
Adds `dream.patterns.cooldown_hours`, a per-phase spend cap for the autopilot
`patterns` dream phase — a direct mirror of the existing
`dream.synthesize.cooldown_hours`. Default `0` (disabled) → no behavior change
unless opted in.

## Why
`patterns` and `synthesize` are both `'mixed'`-scope dream phases dispatched on
every per-source `autopilot-cycle` (~hourly per source). `synthesize` is gated by
a phase-level cooldown (`dream.synthesize.cooldown_hours`, default 12 → "~2
runs/day"); `patterns` has no such gate — it re-queries the last-`lookback_days`
reflection corpus and re-runs a Sonnet subagent from scratch on every cycle. On a
saturated pattern library that's ~19 runs/day re-analyzing a corpus that barely
changes between runs, for near-zero net-new pages — the LLM cost is paid every
cycle regardless. This is the same asymmetry the `synthesize` cooldown was added
to solve (v0.23), applied to its sibling phase.

## How
Mirrors synthesize: reads `dream.patterns.cooldown_hours` (default 0);
`checkCooldown()` against a new `dream.patterns.last_completion_ts`; skips with
reason `cooldown_active` before the subagent is queued; writes the completion
timestamp on a successful run only (a timeout/failed subagent does not start a
cooldown). Additive and opt-in — with default 0, checkCooldown returns
immediately and every existing path is unchanged.

## Tests
- unit (test/cycle-patterns.test.ts): asserts the config key, timestamp key,
  `cooldown_active` reason, and success-gated stamping.
- e2e (test/e2e/dream-patterns-pglite.test.ts, PGLite, no API key): cooldown
  active → skip with no LLM call; expired → passes the gate; default 0 → never
  blocks.

## Follow-ups (kept out of scope to stay minimal)
- A corpus-change gate (skip when the reflection set is unchanged since the last
  run, via a cursor/hash) — a stronger fix than a time cooldown but larger/riskier.
- A `--force` / `gbrain dream --no-cooldown` escape hatch (patterns has no ad-hoc
  `--input` invocation the way synthesize does).

Default chosen as 0 (not 12) to stay strictly non-breaking; happy to default it to
12 to match synthesize if you'd prefer.

### DIFF
--- a/skills/maintain/SKILL.md
+++ b/skills/maintain/SKILL.md
@@ -163,6 +163,11 @@ timestamp is stored in `dream.synthesize.last_completion_ts` and is written
 ONLY on successful runs (not on skipped/failed). Explicit `--input` /
 `--date` / `--from` / `--to` invocations bypass cooldown.
 
+**Patterns cooldown:** `dream.patterns.cooldown_hours` (default 0 = off) mirrors
+the synthesize cooldown for the `patterns` phase, which otherwise runs on every
+autopilot cycle. Set to 12 for ~2 pattern runs/day. Completion timestamp is
+`dream.patterns.last_completion_ts`, written only on a successful run.
+
 **`--dry-run` semantics:** runs the cheap Haiku significance filter (caches
 verdicts) but skips the Sonnet synthesis pass. NOT zero LLM calls.
 
--- a/src/core/cycle/patterns.ts
+++ b/src/core/cycle/patterns.ts
@@ -24,7 +24,7 @@ import type { BrainEngine } from '../engine.ts';
 import type { PhaseResult, PhaseError } from '../cycle.ts';
 import { MinionQueue } from '../minions/queue.ts';
 import { waitForCompletion, TimeoutError } from '../minions/wait-for-completion.ts';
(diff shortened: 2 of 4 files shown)


spiky02plateau:
Follow-up commit d48c361 after a production spend incident: failing `patterns` runs were bypassing the new cooldown because `last_completion_ts` was stamped only when the child subagent completed successfully.

The invariant is now: **budget spent ⇒ cooldown armed**. Once the patterns subagent is submitted, `dream.patterns.last_completion_ts` is stamped for `completed`, `failed`, `dead`, `cancelled`, and local wait `timeout` outcomes, and also from the outer catch if post-LLM bookkeeping throws.

Skip paths that do not spend LLM budget (`disabled`, `cooldown_active`, `insufficient_evidence`, `no_api_key`) and `dryRun` still do not stamp. Tests extended accordingly (32 pass locally).

time-attack:
Reviewed and declined for now with the cycle-knob set. Thank you.

PR #2583, opened 2026-07-04
feat(types): ship type declarations and typed exports for library consumers

meminiai:
### Problem

The package ships TypeScript source without declaration files, and the exports
map carries no "types" entries. A strict TypeScript consumer embedding gbrain
as a library cannot compile against it.

### Change

- `tsconfig.types.json` + `bun run build:types`: tsc emitDeclarationOnly into
  `dist-types/` (measured ~6s on an M-series laptop)
- `prepare` runs it, so both npm and git installs get declarations
- every entry in package exports gains a `types` condition next to its existing
  `default`; no runtime change of any kind
- `dist-types/` is gitignored

### Guards

- `check-exports-count.sh`: baseline stays 20 (keys unchanged)
- `test/public-exports.test.ts`: passes unchanged (36 tests)
- `bun run verify`: 31/31 green; `tsc --noEmit` clean

### Context

We run gbrain as an in-process library inside a production NestJS/Bun backend
(Memini, a multi-tenant consumer app) and have generated exactly these
declarations on postinstall since 0.42.56. This upstreams that step so every
embedder gets working types out of the box. If the `prepare` cost on
contributor installs is unwanted, happy to switch to generating at release
time instead.


time-attack:
Reviewed and declined for now: library-consumer support is not a committed audience yet. Thank you.

PR #2584, opened 2026-07-04
fix(budget): let deliberately-unpriced providers (openrouter:*) run under --max-cost

meminiai:
### Problem

TX2 hard-fails `reserve()` when a cost cap is set and the model is absent from
the native pricing tables. OpenRouter ids are deliberately unpriced there (the
provider bills at its own marked-up rates, so a table entry would lie), which
makes every `--max-cost` run routed through OpenRouter fail before its first
LLM call.

### Change

`CAP_EXEMPT_UNPRICED_PROVIDERS`: a sibling of the FREE_LOCAL_* sets, but NOT
zero-priced, because these are paid providers the tables cannot honestly
meter. Ids whose provider half is in the set fall through to the existing
warn-once path instead of the TX2 throw. Every other unknown model still
hard-fails.

### Tests

Two new cases in `test/core/budget/budget-tracker.test.ts`: openrouter:* under
--max-cost reserves via the warn-once path and writes a reserve_unpriced audit
row; an unknown non-exempt provider still throws no_pricing (regression
guard). 28/28 green; verify 31/31; typecheck clean.

### Context

Hit in production at Memini, where chat/embed traffic is OpenRouter-routed and
every capped brainstorm died at TX2. Carried as a vendored patch since 0.42.40.

### DIFF
--- a/src/core/budget/budget-tracker.ts
+++ b/src/core/budget/budget-tracker.ts
@@ -156,6 +156,21 @@ const FREE_LOCAL_EMBED_PROVIDERS: ReadonlySet<string> = new Set([
 ]);
 
 /**
+ * Provider id prefixes that are deliberately absent from the native pricing
+ * tables because the provider meters billing itself at its own rates —
+ * OpenRouter today (its markup != native pricing, so a table entry would
+ * lie). Unlike the FREE_LOCAL_* sets these are NOT free, so they never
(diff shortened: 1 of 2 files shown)


time-attack:
Closing after maintainer review — decision: not accepting this. The annoying blocker stays, but the spending cap stays honest. The better fix is probably adding real (even rough) prices for OpenRouter so the cap can count them.

PR #2585, opened 2026-07-04
feat(cycle): accept UUID source ids at the cycle lock via a slug-legal derived token

meminiai:
### Problem

`cycleLockIdFor` validates sourceId as a 1-32 lowercase slug. When gbrain is
embedded as a library in a multi-tenant service, sourceId is a tenant key: an
RFC 4122 UUID, 36 characters. Every per-source nightly cycle died at lock
acquisition (status='failed', reason='lock_acquisition_error') before its
first phase ran. Nobody dreamed that night.

### Change

A canonical lowercase UUID stripped of hyphens is exactly 32 lowercase hex
chars: slug-legal by construction, stable, and still recognizable next to the
raw id in logs. `cycleLockIdFor` derives the LOCK id that way; the raw
sourceId stays the data key everywhere else.

Deliberately narrow: every other non-slug shape still throws. The codex r2
P1-B defense-in-depth block (path traversal, whitespace, underscores,
uppercase, length, empty) is untouched and green; uppercase UUIDs also still
throw (canonical lowercase only).

### Tests

New describe block in `test/cycle-lock-per-source.test.ts`: derivation,
distinctness, non-canonical rejection. 44/44 green across the three files that
exercise cycleLockIdFor; verify 31/31; typecheck clean.

### Context

This is the patch we carry in production at Memini (a multi-tenant consumer
app built on gbrain). Full field notes coming as a write-up.

### DIFF
--- a/src/core/cycle.ts
+++ b/src/core/cycle.ts
@@ -508,6 +508,14 @@ export interface LockHandle {
 }
 
 /**
+ * Canonical (lowercase) RFC 4122 UUID. A 36-char UUID breaks the 32-char
+ * slug rule at the lock id, but hyphen-stripped it is slug-legal by
+ * construction (32 lowercase hex chars, alnum edges). Uppercase or
+ * otherwise non-canonical UUIDs deliberately stay invalid.
+ */
+const UUID_SOURCE_ID_RE = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/;
+
+/**
  * Compute the cycle lock ID for a given source.
  *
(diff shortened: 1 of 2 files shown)


time-attack:
Thanks for the clear problem statement and the narrow, well-tested patch. The catch is that `source_id` being a 1-32 lowercase slug is a repo-wide contract (`assertValidSourceId`), not a cycle-lock quirk, so a UUID tenant key is out of contract everywhere else too; special-casing the lock makes one call site accept an id the rest of the system still rejects. Since this exists to support a specific downstream multi-tenant deployment rather than a broken promise in gbrain itself, we are not taking it. Mapping tenant UUIDs to slug-legal source ids at your integration boundary should give you the same result today.

PR #2586, opened 2026-07-04
v0.42.57.0 fix(sync): pin PGLite code sync imports (#2189)

### LINKED ISSUE #2189: gbrain sync (code) fails on ~all files: "undefined is not an object (evaluating 'row.deleted_at')" on PGLite (v0.42.44.0)
Key points in the body: PGLite engine, gbrain 0.42.44.0 (HEAD 090bb532); 811/815 files fail with the identical row.deleted_at TypeError across all file types; the "fix your YAML / failed to parse" message misclassifies a runtime error; repro + impact + the ~/.gbrain/sync-failures.jsonl signature.

javieraldape:
## Summary

- Adds a first-sync PGLite regression for `strategy: 'code'` that imports a TypeScript file and asserts it lands as a code page with no per-file failures.
- Updates blocked sync copy to describe per-file failures as import failures instead of parse/frontmatter failures, so runtime import errors are not misdiagnosed.

Fixes #2189.

## Tests

- `gbrain-gate.sh <worktree> test/sync.test.ts`
  - unit: `test/sync.test.ts` - 67 pass, 0 fail, 186 expect calls

### DIFF
--- a/src/commands/sync.ts
+++ b/src/commands/sync.ts
@@ -196,7 +196,7 @@ export interface SyncResult {
   /** Pages re-embedded during this sync's auto-embed step. 0 if --no-embed or skipped. */
   embedded: number;
   pagesAffected: string[];
-  failedFiles?: number; // count of parse failures (Bug 9)
+  failedFiles?: number; // count of per-file import/sync failures (Bug 9)
   /**
    * v0.41.13.0 partial-sync fields (only set when status === 'partial').
    *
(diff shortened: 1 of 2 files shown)


time-attack:
Closing — the underlying crash from #2189 is already fixed on master by 44cae623 (`fix(pglite): guard putPage against zero-row RETURNING`, #1649), which added the missing empty-rows guard in `putPage`. What remains in this diff is a reword of the blocked-sync copy from "failed to parse" to "failed to import" plus one test, and we are not taking message-only churn as a standalone change. Thanks for chasing the misleading error text down.

PR #2587, opened 2026-07-04
fix schema pack resolution for compiled cli

FroeMic:
## Summary

Fix schema-pack resolution so the schema CLI uses the same pack locator as the active schema-pack loader.

This removes the CLI-only hardcoded `gbrain-base` path logic, exposes the shared bundled-pack registry/locator, and makes the mutation read-only guard derive from the same bundled-pack list.

## Root Cause

`gbrain schema show --as-filing-rules` can resolve the active pack through `loadActivePack`, but explicit schema CLI paths such as `schema show <pack>` used a separate `packPathByName()` helper that only knew about `gbrain-base` and user-installed packs.

In compiled CLI/runtime contexts, bundled pack YAML discovery also needed a repo-root fallback, because the YAML files are not embedded in the compiled binary.

## Validation

- `bun run typecheck`
- `bun test test/schema-cli.test.ts test/schema-pack-mutate.test.ts`
- `bun run verify`
- compiled CLI proof:
  - `bun build --compile --outfile /tmp/gbrain-schema-test src/cli.ts`
  - `cd /tmp && GBRAIN_REPO_PATH=/Users/michaelfrohlich/Repositories/gbrain /tmp/gbrain-schema-test schema show --as-filing-rules`


time-attack:
Closing — the hardcoded-pack-list half of this landed on master via #3110, which added a single bundled-pack registry (`src/core/schema-pack/bundled.ts`) that `schema.ts`, `mutate.ts` and `load-active.ts` all derive from, so the parallel `BUNDLED_SCHEMA_PACK_NAMES`/`locateSchemaPackFile` here would now collide with it. The remaining compiled-binary piece resolves pack YAML from `process.cwd()` and a `GBRAIN_REPO_PATH` env var, which is not a direction we want to take, and the PR also folds in per-command DB config reads and a version bump. Thank you for the clear root-cause write-up and the compiled-CLI repro.

Issue #2051, opened 2026-06-10
ollama:bge-m3 PGLite init/reinit resolves 768d or rejects native 1024d on v0.42.37.0


deyil:
## Bug

`ollama:bge-m3` cannot be used as a PGLite embedding model on `gbrain 0.42.37.0`.

It either:

1. resolves schema as `768d` even though `bge-m3` returns native `1024d`, or
2. rejects `--embedding-dimensions 1024` as unsupported "custom dimensions".

## Environment

- gbrain: `0.42.37.0`
- OS: macOS `15.7.7`
- Bun: `1.3.14`
- Ollama: `0.30.7`
- Engine: PGLite
- Current working model: `ollama:nomic-embed-text` / `768d`
- Target model: `ollama:bge-m3` / native `1024d`

## Repro

Current config:

```text
embedding_model: ollama:nomic-embed-text
embedding_dimensions: 768
```

### 1. Provider test confirms bge-m3 returns 1024d but suggests invalid migration command

```bash
gbrain providers test --model ollama:bge-m3
```

Output:

```text
Embedding dim mismatch: model bge-m3 returned 1024 but schema expects 768.
Fix: Run `gbrain migrate --embedding-model ollama:bge-m3 --embedding-dimensions 1024` or change models.
```

But the suggested command is invalid:

```bash
gbrain migrate --embedding-model ollama:bge-m3 --embedding-dimensions 1024
```

```text
Usage: gbrain migrate --to <supabase|pglite> [--url <url>] [--path <path>] [--force]
```

### 2. PGLite reinit rejects native bge-m3 1024d

```bash
gbrain reinit-pglite --embedding-model ollama:bge-m3 --embedding-dimensions 1024 --yes
```

Output:

```text
Refusing to init: Provider "ollama" model "bge-m3" does not support custom dimensions 1024 (this model only emits its default vector size).
```

But `1024` is the default/native output size for `bge-m3`, not a custom/truncated dimension.

### 3. Omitting dimensions also fails

```bash
gbrain reinit-pglite --embedding-model ollama:bge-m3 --yes
```

```text
--embedding-dimensions <N> is required.
```

### 4. `init --force` picks wrong 768d schema

```bash
gbrain init --force --pglite --embedding-model ollama:bge-m3
```

Output excerpt:

```text
Embedding: ollama:bge-m3 (768d)
Error: Embedding dim mismatch: model bge-m3 returned 1024 but schema expects 768.
```

## Expected

This should work:

```bash
[...]

PR #2592, opened 2026-07-04
fix(recipes): per-model embedding dims — unblocks ollama mxbai/all-minilm/bge-m3

museReed:
## Problem

The Ollama recipe declares a single `default_dims: 768` (nomic-embed-text's native width), but two of its three listed models emit different fixed sizes — `mxbai-embed-large` (1024) and `all-minilm` (384) — and the popular `bge-m3` (1024) isn't listed at all.

Configuring any non-768 ollama model hits a catch-22:

```
$ gbrain reinit-pglite --yes --embedding-model ollama:bge-m3 --embedding-dimensions 1024

$ gbrain reinit-pglite --yes --embedding-model ollama:bge-m3
--embedding-dimensions <N> is required.
```

Both directions are rejected, so these models are unusable. Additionally, the first-model auto-pick paths in `init.ts` read `tp.default_dims` directly and could provision a wrong-width `vector(768)` column for a 1024-dim model.

## Fix

- `EmbeddingTouchpoint.model_dims?: Record<string, number>` — optional per-model native dims
- `defaultDimsForModel(tp, modelId)` helper in `dims.ts`; wired through `resolveSchemaEmbeddingDim`, `resolveSchemaMultimodalDim`, the gateway's multimodal dim check, and the three `init.ts` dim-derivation paths
- Ollama recipe maps its models' true native dims and adds `bge-m3`

Models absent from the map fall back to `default_dims`, so recipes without `model_dims` behave exactly as before.

## Testing

- 4 new cases in `test/embedding-dim-check.test.ts` (explicit 1024 accepted; omitted flag resolves native 1024; mxbai resolves per-model 1024; unlisted model keeps 768 + still rejects custom dims)
- `bun test test/embedding-dim-check.test.ts` → 31 pass / 0 fail; `test/ai/gateway.test.ts` → 36 pass; `tsc --noEmit` clean
- Verified live: `gbrain init --pglite --embedding-model ollama:bge-m3 --embedding-dimensions 1024` now provisions a 1024-dim brain, imported 147 pages, hybrid search working

### DIFF
--- a/src/commands/init.ts
+++ b/src/commands/init.ts
@@ -10,6 +10,7 @@ import { saveConfig, loadConfig, loadConfigFileOnly, toEngineConfig, gbrainPath,
 import { createEngine } from '../core/engine-factory.ts';
 import { discoverOAuth, mintClientCredentialsToken, smokeTestMcp } from '../core/remote-mcp-probe.ts';
 import { runInitEmbedCheck } from '../core/init-embed-check.ts';
+import { defaultDimsForModel } from '../core/ai/dims.ts';
 
 export async function runInit(args: string[]) {
   // Help guard: cli.ts only routes --help to printOpHelp() for shared-op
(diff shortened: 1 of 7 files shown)


time-attack:
Closing — master already ships this: the ollama recipe carries a per-model `model_dims` map (including `bge-m3` at 1024) with `trust_custom_dims: true`, and resolution goes through `embeddingDimsForModel()` in `src/core/ai/model-resolver.ts`, landed under #2051/#2271. Your diagnosis of the catch-22 was exactly right and matches the fix that went in. Thank you for the detailed repro.

Issue #1970, opened 2026-06-08
Sync silently full-walks forever when last_commit bookmark is unreachable (orphaned by history rewrite) — no guard like the checkpoint-target reachability check


garrytan-agents:
## Summary
A source's `last_commit` bookmark can point at a commit that **no longer exists in the source repo** (orphaned by a history rewrite / force-push / branch cleanup). When this happens, `gbrain sync` silently degrades to a **full repo walk on every run, forever**, never advancing `last_sync_at`. On a large brain with a cross-region DB this never completes (see #1958), so the source goes **permanently stale with zero error surfaced**.

## Repro
1. Source `default` has `sources.last_commit = <X>`.
2. The source repo's history is rewritten so `<X>` is no longer reachable (e.g. a `master`→`main` consolidation that drops `<X>`; force-push; squash; branch delete).
3. Run `gbrain sync --source default`.

**Observed:** sync reports `Large sync (NNN,NNN files)` and walks the *entire* tree every run. `git diff <X> HEAD` fails internally because `<X>` is not a valid object. `last_commit` / `last_sync_at` never advance. No error, no warning — just eternal "in progress" that the cron `timeout` keeps killing.

**Expected:** sync detects the unreachable bookmark and either (a) reseeds to the nearest reachable ancestor by date and diffs forward, or (b) does a one-time full reconcile and then stamps the bookmark to HEAD — and in both cases emits a clear `[sync] bookmark <X> unreachable (history rewritten) — re-pinning` log line.

## Root cause
The incremental path resolves `lastCommit = readSyncAnchor(...)` and then `git diff lastCommit..HEAD`. There is handling for a *checkpoint target* that's no longer reachable (`merge-base --is-ancestor` → discard + re-pin to HEAD), but **no equivalent guard for the persisted `last_commit` bookmark itself.** An unreachable bookmark falls through to the full-walk fallback every time instead of being detected and re-pinned.

## Suggested fix
In the incremental sync entrypoint, before diffing:
```
if (lastCommit && !isAncestor(repoPath, lastCommit, headCommit)) {
  slog(`[sync] last_commit ${lastCommit.slice(0,8)} unreachable (history rewritten) — re-pinning to HEAD and reconciling via content_hash`);
  lastCommit = null;            // forces a content-hash reconcile, not an eternal blind re-walk
  // (content_hash already short-circuits unchanged files at ~10ms each)
}
```
[...]

PR #2593, opened 2026-07-04
v0.42.57.0 fix(sync): full-sync reconcile spares out-of-repo imports — ingest_log provenance guard + mass-delete breaker

plucasnz:
## What this fixes

A full sync could delete pages you imported from folders outside the synced repo. If you built a brain with `gbrain import ~/notes` (a folder that is not the repo `gbrain sync` watches) and later configured a `local_path`, the first full sync treated every one of those imported pages as "stale files that disappeared from the repo" and swept them. The reconcile was comparing imported pages against the wrong folder: `pages.source_path` is stored relative to the *import* dir, so out-of-repo imports can never appear in the repo's file enumeration, and the reconcile's three SAFETY conditions didn't cover that case.

Observed impact: 537 of 637 pages deleted on the first `gbrain sync` on a brain whose content came from out-of-repo `gbrain import` runs. Hand-authored (`put_page`) pages were never at risk; bulk-imported ones were.

## The fix

Two protections, plus the correctness fixes that make them sound.

**1. Import provenance guard.** New module-private `externallyImportedSlugs` in `src/commands/sync.ts` reads `ingest_log` `source_type='directory'` rows oldest→newest (last import per slug wins) and the full-sync reconcile excludes any slug whose most recent import root is outside the synced repo. A page later re-imported from inside the repo becomes reconcilable again. When `ingest_log` can't be read, the reconcile is skipped entirely (fail toward keeping pages), with the cause on stderr.

**2. Mass-delete circuit breaker.** Even with provenance missing (e.g. an import killed mid-run), the reconcile refuses an automatic sweep of ≥10 pages when that exceeds 20% of the source's file-backed pages — the signature of provenance loss, not a real cleanup. `GBRAIN_SYNC_RECONCILE_FORCE=1 gbrain sync --full` overrides deliberately.

**Correctness fixes** (found in adversarial + specialist review, without which the guard was a silent no-op on non-default sources):
- `runImport` and the incremental sync now thread `source_id` into `logIngest`, so provenance rows are scoped per source instead of all landing under `default`. The guard reads `source_id IN (sourceId, 'default')` so rows from older binaries stay visible; the legacy arm can only over-protect, never delete.
- `runImport` stores `source_ref` as an absolute path (`resolve(dir)`); the guard classifies any legacy non-absolute ref as external rather than resolving it against the sync process's cwd (which could misclassify an external import as repo-internal).
- Directory identity compared via the shared `realpathOrResolve` (`src/core/path-confine.ts`), memoized per reconcile.
- `SyncResult.reconcileSkipped?: 'provenance_unreadable' | 'mass_delete_breaker'` reports a fail-safe skip machine-readably (so cron/autopilot can tell "nothing stale" from "cleanup disabled").

## Were you affected on an earlier version?

Deletes are soft. `gbrain restore <slug>` brings a swept page back until the purge job removes it; re-running the original `gbrain import <dir>` also restores the content.

## Tests

8 new regression tests in `test/sync.test.ts` (`#1970` block): external-import sweep spared on first sync; last-import-wins re-import flip; fail-closed on unreadable ingest_log with recovery; legacy double-encoded / malformed / mixed-type provenance rows plus a deleted import dir; non-default sources including legacy `default`-filed rows; relative legacy `source_ref` treated as external; identical-content move keeps protection; breaker floor and trip + `GBRAIN_SYNC_RECONCILE_FORCE` override.

`bun test test/sync.test.ts` → 75 pass. `bun run typecheck` clean.

## Commits

- `a8607c0b` — full-sync reconcile spares out-of-repo imports (guard + mass-delete breaker + correctness fixes + tests)
- `91d69e1c` — docs: provenance guard + `GBRAIN_SYNC_RECONCILE_FORCE` knob (KEY_FILES + CLAUDE.md)
- `6d9bddef` — chore(ci): refresh `action-gh-release` SHA pin to current v2
- `8305008e` — v0.42.57.0 version bump + CHANGELOG

Deferred hardening (per-page `import_root` provenance, per-source import lock, remote `log_ingest` hardening, rename-follows-provenance) is filed in `TODOS.md`.

## Note on CI

This PR is from a fork (`plucasnz/gbrain`), so secret-dependent CI jobs won't receive base-repo secrets and may fail on this PR. Per `docs/RELEASING.md`, the branch can be moved into the base repo to run full CI before merge.


time-attack:
Closing — master now covers this scenario from two directions: the mass-delete safety valve (#2828, f15163f7) refuses a reconcile that would sweep a large fraction of a source's file-backed pages, and the never-committed guard (#2426, 42375bde) keeps and re-exports any stale page whose `source_path` was never in this repo's git history, which is exactly the out-of-repo `gbrain import` population you described. The 537-of-637 case trips both. Thank you for the very careful write-up — the provenance analysis is what makes reports like this actionable.

PR #2595, opened 2026-07-05
Strengthen think citation coverage diagnostics

jasonchen-lab:
## Summary

- add a top-ranked page citation contract to `gbrain think`
- expose page citation coverage diagnostics (`pageSlugsByRank`, `citedPageSlugs`, `topPageCited`)
- add a bounded native repair retry when a non-empty synthesis omits the rank-1 page citation
- cover the repair path with a regression test

## Why

`think` can retrieve the right page as rank 1, pass it into the prompt, and still return structured citations that omit that page. The previous behavior made it hard for callers and evals to distinguish retrieval failure from final citation selection failure.

This keeps the fix inside the native `think` owner rather than requiring downstream wrappers to patch citations.

## Verification

- `bun run typecheck`
- `bun test test/think-pipeline.serial.test.ts test/think-gateway-adapter.test.ts test/think-with-calibration.test.ts test/system-prompt.test.ts`
- local patched live probe against a public fixture set: 4/4 passed

### DIFF
--- a/src/core/think/index.ts
+++ b/src/core/think/index.ts
@@ -147,6 +147,12 @@ export interface ThinkResult {
     takesFromKeyword: number;
     takesFromVector: number;
     graphHits: number;
+    /** Page slugs in the same rank order used to render <pages>. */
+    pageSlugsByRank: string[];
+    /** Unique page slugs present in resolved structured or fallback citations. */
+    citedPageSlugs: string[];
+    /** False when a synthesis ran but the rank-1 page is absent from resolved citations. */
+    topPageCited: boolean;
   };
 }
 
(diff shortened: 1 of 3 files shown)


time-attack:
Closing — this is behavior and surface added to a working path rather than a fix for something broken: a prompt contract change, three new diagnostic fields on `ThinkResult`, and a repair retry that costs a second call on `think`, our most expensive op. Citation selection lagging retrieval rank is a quality preference we would want to settle with eval evidence before paying a per-query round-trip for it. The diagnostics idea is well-implemented and the write-up made the tradeoff easy to read — thank you.

PR #580, opened 2026-05-03
feat(search): make FTS language configurable via GBRAIN_FTS_LANGUAGE (1/3)

rafaelreis-r:
## Summary

Adds `GBRAIN_FTS_LANGUAGE` env var (default `'english'`) so non-English brains can opt into a Postgres text-search configuration that actually understands their content.

This is **PR 1 of 3** in a series. It fixes the *query side* only — the trigger functions that populate `search_vector` are addressed in the follow-up PRs:
- **PR 2:** schema migration v33 to recreate trigger functions with the configured language
- **PR 3:** `gbrain reindex-search-vector` CLI for changing language post-install

Each PR is mergeable on its own. The series can be reviewed and landed incrementally.

## Why

Searching a Portuguese (or Spanish, German, etc.) brain currently runs every query through the English stemmer and stop-word list. Result: Portuguese queries lose stemming entirely (`reuniões` doesn't match `reunião`) and miss obvious matches that vector search alone can't recover.

Today the workaround is a downstream patch that hardcodes a different language string into the engine. That patch breaks on every upstream merge. This PR makes the choice configurable so downstream forks for non-English deployments stop accumulating merge debt.

## What changed

- **New file:** `src/core/fts-language.ts` — `getFtsLanguage()` reads `GBRAIN_FTS_LANGUAGE`, validates against `/^[a-z][a-z0-9_]*$/`, falls back to `'english'` on missing/invalid, caches after first read.
- **`postgres-engine.ts`** + **`pglite-engine.ts`:** the four `websearch_to_tsquery('english', ...)` calls in `searchKeyword` and `searchKeywordChunks` now interpolate the helper's return value.
- **README.md:** new "Non-English brains (FTS language config)" subsection under Search.

## Validation strategy

Postgres `tsvector` / `tsquery` functions don't accept parameterized config names — the language **must** be a literal in the SQL string. So we interpolate, but only after the regex guard:

```ts
const VALID_CONFIG_NAME = /^[a-z][a-z0-9_]*$/;
```

A deliberately malicious `GBRAIN_FTS_LANGUAGE="english'; DROP TABLE pages; --"` falls back to `'english'` with a one-time warning. Verified by a unit test.

## Tests

14 unit tests in `test/fts-language.test.ts`:
- defaults (unset / empty / whitespace)
- valid configs (`pt_br`, `spanish`, `custom_lang_v2`)
- rejected forms (quotes, spaces, hyphens, leading digit, uppercase)
- caching behavior + reset for tests
[...]

PR #2596, opened 2026-07-05
fix(search): port the CJK keyword fallback to the Postgres engine (PGLite parity)

SilverNine:
## Summary

Ports the CJK keyword-search fallback from the PGLite engine to the Postgres engine. PGLite has routed CJK queries around `websearch_to_tsquery('english', ...)` since v0.32.7 — `hasCJK()` switches to ILIKE with occurrence-count ranking — but the Postgres engine's two FTS sites never got that branch, so the keyword arm of hybrid search quietly degrades for Korean/Chinese/Japanese brains running on Postgres.

## The gap in practice

The english parser does index space-delimited CJK tokens verbatim, so exact-word queries still match. What misses is everything inflected. Korean is agglutinative: `수급자` (beneficiary) shows up in text as `수급자의`, `수급자를`, `수급자가` — none of which the english stemmer relates to the bare form. On a private production Korean corpus (~74k chunks), a particle-suffixed two-word query returned 7 rows through english FTS and 152 through the ILIKE fallback; a common substring query, 21 vs 847. That is a 20-40x recall gap on the keyword arm, invisible unless you diff the arms directly.

## Change

`postgres-engine.ts` gains the `hasCJK(query)` branch at both FTS sites (the `searchKeyword` dedup path and the `searchKeywordChunks` chunk-grain path), delegating to a new private `_searchKeywordCJK`. The scoring expression, the escaped-pattern/raw-string two-binding discipline, the ILIKE `ESCAPE '\'` predicate, and the dedup vs chunk-grain split are carried over from PGLite unchanged. Where the two engines already diverge, the port sides with its host: `sql.begin` with `SET LOCAL statement_timeout`, named limit params, and the Postgres paths' own extra-filter set (`types`, `exclude_slugs`, its `COALESCE(updated_at, created_at)` date column). The ASCII path is untouched, and the stale PGLite docblock claiming the Postgres engine was intentionally left out is corrected.

This complements the open FTS-language work (#580/#581/#582): a configurable tsvector language fixes stemming for single-language brains, while this fallback covers CJK queries regardless of the configured language — the same layered design PGLite already ships.

## Validation

- 6 source-guardrail tests in `test/postgres-engine.test.ts`: routing at both call sites, ILIKE/ESCAPE discipline, the two-binding scoring contract, pool safety, and #861 source isolation on the fallback path.
- 3 Postgres-vs-PGLite parity e2e tests in `test/e2e/engine-parity.test.ts` with synthetic Korean fixtures, including a ranking assertion (3-occurrence chunk above 1-occurrence) and a LIKE-metachar query that must return nothing.
- `bun run verify` 31/31; touched-area suites 204 pass / 0 fail.

### DIFF
--- a/src/core/pglite-engine.ts
+++ b/src/core/pglite-engine.ts
@@ -1676,8 +1676,8 @@ export class PGLiteEngine implements BrainEngine {
    *     chunk_text are compared as-stored).
    *   - Empty-query guard returns no results without binding SQL.
    *
-   * Postgres engine is intentionally untouched (multi-tenant deployments
-   * can install pgroonga / zhparser when needed; out of scope here).
+   * Ported to postgres-engine.ts `_searchKeywordCJK` (Postgres parity) so
+   * both engines share the same CJK fallback semantics.
    */
   private async _searchKeywordCJK(
     query: string,
--- a/src/core/postgres-engine.ts
+++ b/src/core/postgres-engine.ts
@@ -63,6 +63,7 @@ import { ConnectionManager } from './connection-manager.ts';
 import { logConnectionEvent } from './connection-audit.ts';
 import { validateSlug, contentHash, rowToPage, rowToStalePage, rowToChunk, rowToSearchResult, parseEmbedding, tryParseEmbedding, takeRowToTake, takeHitRowToHit, isUndefinedTableError, warnOncePerProcess } from './utils.ts';
 import { resolveBoostMap, resolveHardExcludes } from './search/source-boost.ts';
+import { hasCJK, escapeLikePattern } from './cjk.ts';
 import { buildSourceFactorCase, buildHardExcludeClause, buildVisibilityClause, buildRecencyComponentSql, buildBestPerPagePoolCte } from './search/sql-ranking.ts';
 import { DEFAULT_EMBEDDING_MODEL, DEFAULT_EMBEDDING_DIMENSIONS } from './ai/defaults.ts';
 import { DELETE_BATCH_SIZE } from './engine-constants.ts';
@@ -1690,6 +1691,30 @@ export class PostgresEngine implements BrainEngine {
     const hardExcludePrefixes = resolveHardExcludes(opts?.exclude_slug_prefixes, opts?.include_slug_prefixes);
     const hardExcludeClause = buildHardExcludeClause('p.slug', hardExcludePrefixes);
 
+    // v0.26.5: visibility filter hides soft-deleted pages and pages from
+    // archived sources. Joined `sources s` lets the predicate compile to a
+    // column lookup. NOT bypassed by detail=high — soft-delete is a contract,
+    // not a temporal preference. (Hoisted above the CJK branch so the
+    // fallback can reuse it, mirroring the PGLite hoist.)
+    const visibilityClause = buildVisibilityClause('p', 's');
+
+    // v0.32.7 CJK branch (Postgres parity): `websearch_to_tsquery('english')`
+    // can't tokenize CJK — a Chinese/Japanese/Korean query compiles to an
+    // empty tsquery and the FTS path silently returns nothing. Switch to
+    // ILIKE on chunk_text with occurrence-count ranking as the ts_rank
+    // substitute when the query contains CJK characters, mirroring the
+    // PGLite engine's `_searchKeywordCJK`. ASCII path stays exactly the
+    // same below.
+    if (hasCJK(query)) {
+      return this._searchKeywordCJK(query, {
+        limit, offset, innerLimit, sourceFactorCase,
+        hardExcludeClause, visibilityClause,
+        detailFilter: detailLow ? `AND cc.chunk_source = 'compiled_truth'` : '',
+        opts,
+        dedup: true,
+      });
+    }
+
     const params: unknown[] = [query];
     let typeClause = '';
     if (type) {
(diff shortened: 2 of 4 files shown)


SilverNine:
Rebased onto current master (f72de97) — resolved conflicts with the configurable-FTS work (#2941) and adapted the fallback to the post-#2940 RLS scope-binding idiom:

- Kept `getFtsLanguage()` on both FTS sites; the CJK branch is unaffected (it exists to bypass the tokenizer entirely).
- `_searchKeywordCJK` now executes through `withScopedReadTransaction(..., { alwaysTransaction: true })` instead of a bare `sql.begin()`, so with `GBRAIN_RLS_SCOPE_BINDING` on, the fallback shares the same `set_config('app.scopes')` transaction as the FTS arm — the CJK path can never widen a caller's source scope. Guardrail test updated to pin this.
- Both new upstream guardrails pass (`postgres-engine.test.ts` 14/14), `tsc --noEmit` clean.

time-attack:
Merging. This is a genuine engine-parity defect — the CJK fallback has been PGLite-only since v0.32.7, so the keyword arm of hybrid search quietly under-recalls on Postgres, and engine parity is a hard invariant here. Thank you for keeping the ASCII path untouched, carrying over the escaped/raw two-binding discipline and the source-isolation clause, and backing it with both source guardrails and Postgres-vs-PGLite parity e2e tests.

seungsu-kr:
Two notes from running this patch on a production Korean corpus — one about the PR's status, one about a recall regression I hit with the phrase-matching approach.

## 1. This looks like it was closed without landing

The last comment here says "Merging", but the PR shows `merged: null` and the timeline has a `closed` event with no `merged` event. I also checked current `master` directly:

```
$ gh api repos/garrytan/gbrain/contents/src/core/postgres-engine.ts --jq .content | base64 -d | grep -c 'hasCJK\|_searchKeywordCJK'
0
```

`TODOS.md` still carries "v0.42 — Postgres CJK FTS via pgroonga / zhparser / ngram trigrams" as unchecked. So as far as I can tell the fix never landed. Possibly just a slip when closing out the queue.

## 2. Phrase-ILIKE is narrower than the tsquery it replaces

I cherry-picked `76cedaa2` onto a private Korean wiki brain (Postgres engine, ~21.6K chunks) and measured the keyword arm before/after. Single-token queries improved as expected, but **multi-token queries lost recall against the very FTS path the fallback exists to improve on.**

The cause is that `websearch_to_tsquery` joins terms with `&` and matches them **anywhere in the document**, while `ILIKE '%' || $1 || '%'` requires the entire query to appear as one **contiguous** substring. Raw chunk counts:

| query (gloss) | FTS | phrase-ILIKE | token-AND-ILIKE |
|---|---|---|---|
| 수명주기 규칙 (lifecycle rules) | 14 | 2 | 24 |
| 워크트리 정리 (worktree cleanup) | 29 | 2 | 47 |
| 세션 핸드오프 (session handoff) | 162 | 11 | 230 |
| 브랜치 삭제 (branch deletion) | 115 | 23 | 157 |
| 임베딩 모델 (embedding model) | 18 | 8 | 29 |
| 디렉토리 소유권 (directory ownership) | 0 | 0 | 2 |

Same shape through the real `searchKeyword` entry point (page-grain, after dedup, limit 50): 11 → 2 → 19, 18 → 1 → 30, 13 → 4 → 20.

I don't think this contradicts the 7 → 152 result in the PR description — that query was presumably a contiguous phrase in that corpus. It just doesn't generalize to queries whose terms are scattered across the chunk, which is most of mine.

### What worked

Split the query on whitespace and AND the per-token ILIKEs. That keeps tsquery's AND semantics while substring matching still absorbs the agglutinative suffixes this fallback exists for — `소유권` still matches `소유권은` / `소유권을`. Single-token queries are byte-identical to the current behavior (one clause).

```js
const tokens = qRaw.split(/\s+/).filter((t) => t.length > 0);
if (tokens.length === 0) return [];

for (const tok of tokens) {
  params.push(escapeLikePattern(tok));
  const likeParam = `$${params.length}`;
  params.push(tok);
  const rawParam = `$${params.length}`;
  ilikeClauses.push(`cc.chunk_text ILIKE '%' || ${likeParam} || '%' ESCAPE '\\'`);
  // …per-token occurrence + position terms, summed
}
const ilikePredicate = ilikeClauses.join('\n            AND ');
```

The escaped/raw two-binding discipline carries over unchanged — it just becomes two bindings **per token** instead of a fixed `$1`/`$2`.

## 3. Latent scoring bug in the ranking expression

Independent of the above, worth fixing either way:

```sql
1.0 / NULLIF(POSITION($2 IN cc.chunk_text), 0)::real
```

`ILIKE` is case-insensitive but `POSITION` is not. A Latin-cased token can satisfy the `WHERE` clause and still return `POSITION = 0` → `NULLIF` → `NULL`, which NULLs out the entire score expression. Under `ORDER BY score DESC` Postgres sorts `NULLS FIRST`, so those rows float to the **top** of the result set. Pure CJK queries never hit this; mixed CJK+ASCII ones do (e.g. a query mixing a Korean noun with a Latin identifier). `COALESCE(…, 0)` closes it.

---

Happy to open a PR with the token-AND change plus guardrail tests (I have it running locally with `tsc --noEmit` clean and the source-invariant tests updated to pin the two bindings by origin rather than by fixed `$1`/`$2` slot, plus new tests asserting the join is AND-never-OR and the COALESCE guard). But that only makes sense on top of this one landing first — happy to follow whatever order you prefer.

PR #2598, opened 2026-07-05
Make remote think read-scoped

colinagent:
## Summary
- Mark the `think` operation as OAuth/MCP `read` scoped.
- Keep `think.mutating = true` because trusted local CLI callers can still persist with `--save` / `--take`.
- Preserve the existing remote trust boundary: MCP/remote callers force `save` and `take` off before persistence, and tests now pin that read-scoped remote behavior.
- Update current-behavior docs for the thin-client and operations contract.

## Validation
- `bun test test/oauth.test.ts test/takes-mcp-allowlist.serial.test.ts test/build-llms.test.ts`
- `bun run typecheck`
- `bun run verify` passed 31/31 with a temporary local `gtimeout` shim; macOS lacks `timeout/gtimeout`, and the fallback runner records the sleep-cap rc=143 on this machine.
- `git diff --check`


time-attack:
Closing: think stays write-scoped deliberately. Scope checks are hierarchical (write satisfies read), so no legitimate token is blocked today — the only effect of this change would be granting read-only OAuth tokens the ability to trigger synthesis calls billed to the brain owner's API key, a policy loosening with no driving issue. It would also put local `gbrain think` under the CLI's 180s read-op wall-clock timeout, which multi-round synthesis can exceed. Happy to revisit with a concrete use case for read-token think access.

PR #2600, opened 2026-07-05
fix(ai): add chat touchpoint to litellm proxy recipe

colinagent:
## Summary

- add a `chat` touchpoint to the LiteLLM proxy recipe
- keep model IDs user-provided so the proxy/backend can validate them
- add focused recipe tests for chat support, embedding behavior, and auth resolution

## Why

LiteLLM proxy is an OpenAI-compatible proxy commonly used for chat models, but the recipe only declared an embedding touchpoint. That makes configured chat usage like `models.think = litellm:<model>` fail `assertTouchpoint("chat")` and fall back to no-LLM behavior even though the proxy can serve `/chat/completions`.

The existing capabilities error text already lists `litellm-proxy` among known chat providers, so this change aligns the recipe with that documented expectation.

## Tests

```sh
bun test test/ai/recipe-litellm-proxy.test.ts test/ai/recipes-existing-regression.test.ts
```

### DIFF
--- a/src/core/ai/recipes/litellm-proxy.ts
+++ b/src/core/ai/recipes/litellm-proxy.ts
@@ -20,6 +20,15 @@ export const litellmProxy: Recipe = {
     setup_url: 'https://docs.litellm.ai/docs/proxy/quick_start',
   },
   touchpoints: {
+    chat: {
+      // Models depend on the proxy's config; the openai-compat tier accepts
+      // user-provided IDs and lets the proxied backend validate them.
+      models: [],
+      supports_tools: true,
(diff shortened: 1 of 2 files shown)


time-attack:
Thanks for this — it looks like #2208 beat it to master (merged as 00523b84) and made the identical change: the LiteLLM proxy recipe now declares a `chat` touchpoint (`src/core/ai/recipes/litellm-proxy.ts`), plus an `expansion` touchpoint, so `assertTouchpoint("chat")` no longer fails for `models.think = litellm:<model>`. Closing this one as superseded — appreciate you catching the same gap.

Issue #2194, opened 2026-06-15
autopilot: multi-source Postgres fan-out produces continuous dead autopilot-cycle jobs + recurrent queue wedge


garrytan-agents:
## Summary

On a multi-source Postgres brain, autopilot's per-source fan-out generates a continuous stream of **dead `autopilot-cycle` jobs** (200+/24h in our deployment) and periodically **wedges the queue** (worker alive, 0 active, no completions). Root cause is a three-way interaction between `fanoutMax`, worker concurrency, and the fact that per-source cycle *locks* are isolated but the heavy cycle *phases* are still brain-wide.

## Environment

- `gbrain` autopilot + Minions supervisor, Postgres engine (PgBouncer txn-mode, port 6543).
- 5 federated sources, all local-path, `fanoutMax` resolves to **4** (Postgres default, `autopilot-fanout.ts:83`).
- Supervisor run with `--concurrency 3`.
- `--interval 300` (5-min tick).

## Observed

`jobs list` shows a steady cadence: every ~5-min tick dispatches 3–4 `autopilot-cycle` jobs (one per stale source). Per tick, typically **1–2 complete and 2–3 go DEAD after ~250–470s** with `max_attempts` exhausted, e.g.:

```
15305 autopilot-cycle dead       397.9s
15304 autopilot-cycle dead       399.9s
15303 autopilot-cycle completed  254.1s
15302 autopilot-cycle completed  169.0s
15301 autopilot-cycle dead       249.7s
```

`jobs stats` over 24h: `autopilot-cycle 309 total / ~91 done / 213 dead`. Doctor intermittently reports `wedged_queue: worker alive but not claiming work (1 waiting, 0 active, 19m since last completion)`; the documented remediation (`supervisor stop && start`) clears it, but it returns within the hour.

## Root cause (verified by reading source)

1. **Fan-out exceeds worker capacity.** `resolveFanoutMax()` returns 4 for Postgres (`autopilot-fanout.ts:83`), independent of the worker's `--concurrency`. With concurrency 3 and 4 fanned-out cycles, at least one cycle is queued behind the others every tick. `dispatchPerSource` deliberately omits `maxWaiting:1` (correct, to preserve per-source fan-out), so nothing back-pressures the surplus.

[...]

PR #2612, opened 2026-07-05
fix: scope per-source cycle embed and orphan phases

gregclarkgsd:
## Summary
- Threads `sourceId` into source-capable `autopilot-cycle` phases: `embed` and `orphans`.
- Preserves `sourceId` / `source_id` for single-phase Minion phase-wrapper jobs.
- Adds regression coverage proving `runCycle` forwards source scope into embed and orphan discovery.

## Why
Per-source `autopilot-cycle` jobs carried `source_id`, but some source-capable phases still ran avoidably whole-brain work. On larger Postgres/Supabase brains this can trigger `statement_timeout` during per-source maintenance jobs.

This patch keeps the fix narrow: no new config, no DB migrations, no global timeout increase.

## Changed files
- `src/core/cycle.ts`
- `src/commands/jobs.ts`
- `test/core/cycle.serial.test.ts`

## Verification
Run locally from `/Users/claudebot/gbrain-autopilot-timeout-fix`:

```bash
git diff --check
# passed
```

```bash
/Users/claudebot/.bun/bin/bun test test/core/cycle.serial.test.ts
# 30 pass, 0 fail, 80 expect() calls
```

```bash
/Users/claudebot/.bun/bin/bun test test/autopilot-fanout.test.ts test/core/cycle.serial.test.ts
# 58 pass, 0 fail, 140 expect() calls
```

```bash
/Users/claudebot/.bun/bin/bun test test/jobs-autopilot-cycle-braindir.serial.test.ts
# 1 pass, 0 fail, 7 expect() calls
```

## Deliberately not changed
- Did not raise or remove Postgres/Supabase `statement_timeout`.
- Did not decompose the remaining truly/global-ish phases (`purge`, `resolve_symbol_edges`, `grade_takes`, `calibration_profile`).
- Did not change queue retry semantics or introduce migrations.

### DIFF
--- a/src/commands/jobs.ts
+++ b/src/commands/jobs.ts
@@ -1921,10 +1921,17 @@ export async function registerBuiltinHandlers(
     const repoPath: string | null = typeof job.data.repoPath === 'string'
       ? job.data.repoPath
       : ((await engine.getConfig('sync.repo_path')) ?? null);
+    const sourceId =
+      typeof job.data.sourceId === 'string'
+        ? job.data.sourceId
+        : typeof job.data.source_id === 'string'
+          ? job.data.source_id
+          : undefined;
     const report = await runCycle(engine, {
       brainDir: repoPath,
       phases: [phase as any],
       signal: job.signal,
+      ...(sourceId ? { sourceId } : {}),
     });
     return { phase, status: report.status, report };
   };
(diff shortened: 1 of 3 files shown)


time-attack:
Closing: since the autopilot cycle split (#2194/#2227), per-source autopilot-cycle jobs already exclude the embed/orphans phases — they're global-scoped and run once per window in autopilot-global-maintenance — so the per-source whole-brain-embed path this targets no longer fires on master. As written, the change would also scope the *global* maintenance job's embed/orphans to the dir-resolved source (that job passes brainDir without a source_id), silently starving every other source of stale-embedding sweeps. Thanks for digging in — if the timeout reappears, a version that threads only the explicit opts.sourceId would be worth discussing.

PR #2617, opened 2026-07-05
fix(ai): promote DeepSeek reasoning_content when content is empty

spiky02plateau:
## Bug

DeepSeek-style reasoning models served through the openai-compatible `deepseek:` recipe (deepseek-reasoner, GLM reasoning variants, etc.) can return the substantive answer in `choices[].message.reasoning_content` while `content` is empty. The AI SDK adapter reads only `content`, so `gateway.chat()` yields `text: ''` and callers classify the response as empty.

Real-world impact: on one such model, **~54% of per-chunk synopsis calls** returned reasoning-only responses → `sanitizeSynopsis` → `kind:'empty'` → 1,030 pages collapsed to the title-only contextual-retrieval fallback, silently defeating the synopsis tier.

## Fix

A recipe-scoped inbound response rewriter, following the exact `voyageCompatFetch` / `zeroEntropyCompatFetch` precedent: when `content` is empty/whitespace and `reasoning_content` is a non-empty string, promote it into `content` before the SDK parses the body. Non-streaming JSON chat completions only; non-empty `content` always passes through untouched.

Plumbing: a new optional recipe-level `compat.fetch` hook (types.ts) wired in `applyOpenAICompatConfig`, so a recipe can install a static transport shim while preserving the normal `base_urls[recipe.id]` override chain and any `resolveOpenAICompatConfig` fetch taking precedence.

## Tests

New `test/deepseek-reasoning-content.test.ts` (fetch-stub style, drives the full `chat()` path):
- empty `content` + `reasoning_content` → text promoted
- non-empty `content` + `reasoning_content` → content unchanged, no duplication
- both empty → stays empty, no crash

https://claude.ai/code/session_01UTCLwsr3DCY3ukjeBz5Tv4

### DIFF
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -377,7 +377,11 @@ export function applyOpenAICompatConfig(
   cfg: AIGatewayConfig,
 ): { baseURL: string; fetch?: typeof fetch } {
   if (recipe.resolveOpenAICompatConfig) {
-    return recipe.resolveOpenAICompatConfig(cfg.env);
+    const compat = recipe.resolveOpenAICompatConfig(cfg.env);
+    return {
+      ...compat,
+      fetch: compat.fetch ?? recipe.compat?.fetch,
+    };
   }
   const baseURL = cfg.base_urls?.[recipe.id] ?? recipe.base_url_default;
   if (!baseURL) {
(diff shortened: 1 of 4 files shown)


time-attack:

PR #2618, opened 2026-07-05
v0.42.57.0 fix(think): persist --take rows (#2556)

### LINKED ISSUE #2556: think --take silently writes no take row (CLI and MCP)
`gbrain think "…" --anchor note/politica-sconti --take` (local CLI, v0.42.52.0 @ bb2e88c, chat model openai:gemma4:e4b) completes and prints the synthesis; the footer itself says `Takes: 0`, and `SELECT count(*) FROM takes` stays 0 afterwards. No error, no hint. The MCP variant with `take: true` behaves the same (when it doesn't hit client timeouts first).

Expected: a take row appended to the anchor page, or an explicit message saying what's missing.

Possibly the same holder-identity fragmentation family as #2464 / #2465 (take written under a holder that never matches?), or related to #2471 (takes never embedded — suggests the takes write path has low traffic). If takes require extra config (holder identity, calibration profile), `--take` should say so instead of silently writing nothing.

javieraldape:
## Summary

- Adds `persistThinkTake()` so a successful local `gbrain think --take --anchor <slug>` writes the synthesis answer as the next append-only take on the anchor page.
- Wires the same take persistence into the local `think` operation handler while preserving the remote trust boundary with an explicit `REMOTE_PERSISTED_BLOCKED` warning.
- Extends the CLI/operation response surface with `take_row` and `take_inserted`, and keeps empty/no-LLM synthesis from writing blank take rows.

Fixes #2556.

## Tests

- `bun test test/think-pipeline.serial.test.ts test/takes-mcp-allowlist.serial.test.ts`
  - 43 pass, 0 fail, 116 expect calls
- `gbrain-gate.sh <worktree> test/think-pipeline.serial.test.ts test/takes-mcp-allowlist.serial.test.ts`
  - unit: 43 pass, 0 fail, 116 expect calls

### DIFF
--- a/src/commands/think.ts
+++ b/src/commands/think.ts
@@ -6,7 +6,7 @@
  * degrades to gather-only output with a warning if missing.
  */
 import type { BrainEngine } from '../core/engine.ts';
-import { runThink, persistSynthesis } from '../core/think/index.ts';
+import { runThink, persistSynthesis, persistThinkTake } from '../core/think/index.ts';
 import { loadConfig, isThinClient } from '../core/config.ts';
 import { callRemoteTool, unpackToolResult } from '../core/mcp-client.ts';
 
@@ -92,6 +92,8 @@ prints what would have been the input (exit 0).
   let result: any;
   let savedSlug: string | undefined;
   let evidenceInserted = 0;
+  let takeRow: number | null = null;
+  let takeInserted = 0;
   const cfg = loadConfig();
   if (isThinClient(cfg)) {
     if (save || take) {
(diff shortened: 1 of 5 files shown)


time-attack:
Closing as superseded by #3164, which carries this exact `persistThinkTake` work forward — rebased onto master's `thinkSourceScopeOpts` split (#2739) and credited to you by name. The diagnosis and the trust-boundary handling here were right; they landed in the successor rather than being dropped. Thank you for the original fix.

garrytan:
Follow-up: the #2556 half of this (persistThinkTake — the take flag that nothing consumed) ships in #3912 (v0.42.79.0), built on your design with dual-plane fence+DB append added on top. Credited in the release CHANGELOG — thanks for the fix.

PR #2619, opened 2026-07-05
fix(think): honor source scope during gather

colinagent:
## Summary
- thread runThink sourceId/allowedSources into runGather
- apply the same source scope to hybrid page search, takes keyword search, takes vector search, and graph traversal
- make searchTakes/searchTakesVector honor sourceId/sourceIds in both Postgres and PGLite engines

## Why
MCP-bound think calls already receive the caller source scope and trajectory uses it, but gather dropped it. In multi-source brains this made think gather from the wrong source even when query with the same scope could find the evidence.

Observed in a multi-tenant Cloud brain:
- think("what did I discuss with the OpenBrain founder trial conversation") gathered 0 pages / 0 takes
- query("openbrain") on the same scoped brain returned the target record

The root cause was that runThink only applied source scope to trajectory. The gather phase, including takes search, did not consistently use it.

## Tests
- bun test test/think-pipeline.serial.test.ts
- bun test test/takes-mcp-allowlist.serial.test.ts
- bun run typecheck
- git diff --check

### DIFF
--- a/src/core/pglite-engine.ts
+++ b/src/core/pglite-engine.ts
@@ -4566,9 +4566,19 @@ export class PGLiteEngine implements BrainEngine {
 
   async searchTakes(
     query: string,
-    opts: { limit?: number; takesHoldersAllowList?: string[] } = {},
+    opts: SearchOpts & { takesHoldersAllowList?: string[] } = {},
   ): Promise<TakeHit[]> {
     const limit = clampSearchLimit(opts.limit, 30, 100);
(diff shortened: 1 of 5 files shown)


time-attack:
Closing as superseded by #2739 (merged in v0.42.59.0): master now enforces the caller's source scope across all of think's gather — hybrid page retrieval, takes keyword/vector retrieval, and graph traversal — with functionally identical engine changes. Thanks for the fix and the clear writeup!

PR #2620, opened 2026-07-05
fix(fences): stop pipe characters from dropping facts and takes

morluto:
## What & why

Facts and takes fences escaped literal pipe characters on write (`|` -> `\|`), but the shared row parser still split on every `|`. A row like `Founded Acme\|Co in 2017` was parsed as extra cells, then rejected as malformed. On later fence rewrites, only successfully parsed rows were rendered back, so a pipe-bearing fact or take could disappear from the markdown source of truth.

For users, this meant normal text such as `A|B testing`, `Acme|Co`, or `source: CRM | LinkedIn` could corrupt a facts/takes row. The visible effects were confusing: a newly added fact might not make it into the derived DB, search/trajectory/scorecard results could miss it, and a later unrelated append/update could silently remove the original row from the fence.

This fixes the shared fence parser so escaped pipes are treated as cell text, then unescaped back to literal `|` after splitting. Because facts and takes both use `parseRowCells`, the fix applies to both fences.

## Changes

- `parseRowCells` now splits only on unescaped pipe delimiters.
- Added `unescapeFenceCell` as the inverse of `escapeFenceCell`.
- Added facts regression coverage for:
  - `renderFactsTable -> parseFactsFence` preserving pipes in text cells.
  - `upsertFactRow` preserving an existing pipe-bearing row during an unrelated append.
- Added takes regression coverage for:
  - `renderTakesFence -> parseTakesFence` preserving pipes in text cells.
  - `upsertTakeRow` preserving an existing pipe-bearing row during an unrelated append.

## Verification

- [x] `bun test test/facts-fence.test.ts test/takes-fence.test.ts`
- [x] `bun test test/facts-fence-typed.test.ts test/takes-holder-validation.test.ts test/takes-holder-semantics.test.ts`
- [x] `bun run typecheck`
- [x] `bun run verify` - 31/31 checks green

## Documentation

No README or architecture-doc updates needed for this narrow parser fix. `CONTRIBUTING.md` says changelog/version entries are handled during the ship flow, so this PR leaves `CHANGELOG.md`, `VERSION`, and `package.json` unchanged.

### DIFF
--- a/src/core/fence-shared.ts
+++ b/src/core/fence-shared.ts
@@ -25,19 +25,31 @@
  *
  * Returns `null` when the line is not a table row (doesn't start with `|`
  * or has no second pipe). On a match, returns the cells with surrounding
- * whitespace trimmed, with the outer pipes already stripped.
- *
- * NOTE: does NOT unescape `\|` back to `|`. Round-trip-on-pipes is a
- * separate concern callers handle if their domain text legitimately
- * contains pipes (currently neither takes nor facts do at the LLM-extract
- * layer; if a hand-edit introduces one, escape-on-write at render time
- * protects the table shape).
+ * whitespace trimmed, with the outer pipes already stripped. Escaped pipes
+ * (`\|`) are treated as cell text and unescaped in the returned cells.
  */
 export function parseRowCells(line: string): string[] | null {
   const trimmed = line.trim();
   if (!trimmed.startsWith('|') || !trimmed.includes('|', 1)) return null;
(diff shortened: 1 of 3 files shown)


time-attack:
Closing as superseded by #2738 (merged in v0.42.59.0): the shared fence parser now splits only on unescaped pipes and round-trips render→parse symmetrically, covering both facts and takes. If you'd like, the takes-side regression tests from this PR would still be a welcome test-only follow-up. Thanks!

PR #2622, opened 2026-07-05
fix(gateway): stop misdiagnosing explicit litellm/user-provided embedding models as unset

colinagent:
## Summary

`diagnoseEmbedding`'s `user_provided_model_unset` guard fires on recipe-**static** properties (`models: []` + `user_provided_models: true`) without ever inspecting the user's model string. A fully explicit `litellm:zembed-1` is diagnosed as "model unset", `isAvailable('embedding')` returns `false`, and `hybridSearch` silently drops its vector arm on every litellm-proxied (or llama-server, or any user-provided-models recipe) deployment.

The failure mode is nasty to debug because it is **read-side only**: `embed-backfill` does not run this preflight, so vectors keep getting written — the corpus looks fully embedded, `gbrain doctor` embedding checks pass, but no query can ever retrieve through the vector arm. `gbrain search diagnose` shows:

```
"vector": { "note": "skipped — no embedding provider configured" }
```

despite `GBRAIN_EMBEDDING_MODEL=litellm:zembed-1` and a reachable proxy.

## Why removal (not a condition) is correct

A genuinely missing model part can never reach this guard:

- bare `litellm` → `parseModelId()` throws "missing a provider prefix"
- `litellm:` (empty model) → `parseModelId()` throws "empty provider or model"

Both surface earlier in `diagnoseEmbedding` as `unknown_provider` with the paste-ready `provider:model` format hint. Every string that reaches the guard therefore has a non-empty parsed model id — the guard could only ever misfire. Removed, with an explanatory comment at the site.

The `user_provided_model_unset` variant stays in `EmbeddingDiagnosis` so consumer switch arms (`init-embed-check.ts`, `embed-preflight.ts`) keep compiling; `diagnoseEmbedding` just no longer emits it.

## Production verification

Found and verified in a production deployment (OpenBrain Cloud) that fronts gbrain with an OpenAI-compatible internal gateway via `LITELLM_BASE_URL`:

- Before: `gbrain search diagnose "personel knowlege memori engin" --target projects/opagent` → vector `skipped — no embedding provider configured`, target absent from hybrid top-50.
- After (same env, recipe gate bypassed): vector arm runs, target present; a CJK long-question probe whose keyword tokens cannot match jumped to **vector rank 2**.

Related: this is the same litellm-recipe blind-spot family as #2600 (litellm chat touchpoint).

## Test plan

- [x] New `test/diagnose-embedding-user-provided.test.ts` (7 cases): explicit `litellm:<model>` ok with and without optional env; `llama-server:<model>` ok; `modelOverride` probe ok against a foreign global default; bare `litellm` and `litellm:` → `unknown_provider`; `missing_env` ordering preserved for recipes with required auth.
- [x] Existing adjacent suites green: `gateway-embed-model-override`, `embed-preflight`, `init-embed-check`, `embed.serial`, `sync-cost-gate.serial`, `cross-modal-hybrid-integration.serial`, `hybrid-meta.serial`, `hybrid-search-lite.serial`, `gateway-model-messages`, `think-gateway-adapter`.

### DIFF
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -704,21 +704,24 @@ export function diagnoseEmbedding(modelOverride?: string): EmbeddingDiagnosis {
     };
   }
 
-  // Openai-compat recipes with empty models list require a user-provided model.
-  const isUserProvided = (tp as any).user_provided_models === true;
-  if (
-    Array.isArray(tp.models) &&
-    tp.models.length === 0 &&
-    (recipe.id === 'litellm' || isUserProvided)
-  ) {
-    return {
-      ok: false,
-      reason: 'user_provided_model_unset',
-      model: modelStr,
-      provider: parsed.providerId,
-      recipeId: recipe.id,
-    };
-  }
+  // NOTE: there used to be a `user_provided_model_unset` guard here for
+  // openai-compat recipes with an empty models list (litellm, llama-server,
+  // ...). It fired on recipe-STATIC properties (`models: []` +
+  // `user_provided_models: true`) without ever inspecting the user's model
+  // string — so a fully explicit `litellm:zembed-1` was misdiagnosed as
+  // "model unset", isAvailable('embedding') returned false, and hybrid
+  // search silently dropped its vector arm on every litellm /
+  // user-provided-model deployment (writes were unaffected: embed-backfill
+  // does not run this preflight, which made the failure look like
+  // "vectors exist but never match").
+  //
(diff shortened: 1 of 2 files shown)


time-attack:
Closing as superseded by #2627 (merged in v0.42.58.0), which removed the structurally-unreachable user_provided_model_unset guard and replaced it with a dims-aware fail-closed check (user_provided_dims_unset). Merging this PR on top would regress that stricter guard. Thanks for the report and fix!

Issue #357, opened 2026-04-23
gbrain init wedges on pre-v0.13 brains: column "link_source" does not exist


SC-SoftwareHS:
## Summary

Running `gbrain init` against an existing `~/.gbrain/brain.pglite` that was created before v0.13 fails immediately with `column "link_source" does not exist`, and can never recover. Migration 11 never runs, so the column never gets added, and every subsequent `gbrain init` hits the same error. The only escape is `rm -rf ~/.gbrain/brain.pglite`, which loses all data.

## Reproduction

```bash
# On any install that created ~/.gbrain/brain.pglite before v0.13:
gbrain init
# →  Setting up local brain with PGLite (no server needed)...
# →  column "link_source" does not exist
```

Hit this today on a brain created in mid-April after upgrading to current master.

## Root cause

`initSchema()` in `src/core/pglite-engine.ts:86` runs the full schema SQL **before** `runMigrations()`:

```ts
async initSchema(): Promise<void> {
  await this.db.exec(PGLITE_SCHEMA_SQL);
  const { applied } = await runMigrations(this);
}
```

Both `src/core/pglite-schema.ts:82-83` and `src/schema.sql:83-84` have these lines inside the schema SQL:

```sql
CREATE INDEX IF NOT EXISTS idx_links_source ON links(link_source);
CREATE INDEX IF NOT EXISTS idx_links_origin ON links(origin_page_id);
```

On a **fresh** install, the preceding `CREATE TABLE IF NOT EXISTS links (...)` includes those columns, so the indexes work. On an **upgrade** install, the `links` table already exists without `link_source` / `origin_page_id`, so `CREATE TABLE IF NOT EXISTS` no-ops, and then `CREATE INDEX IF NOT EXISTS` throws because the column doesn't exist. `initSchema()` bails, migrations never run.

Migration 11 (`links_provenance_columns`) already creates both indexes correctly — **after** it adds the columns — at `src/core/migrate.ts:360-361`. The lines in the schema files are redundant duplicates that happened to succeed on fresh installs.

## Proposed fix

Remove the two premature `CREATE INDEX` lines from both schema files and leave a comment pointing at migration 11. Fresh installs are unaffected because migration 11's `ALTER TABLE ... IF NOT EXISTS` + `CREATE INDEX ... IF NOT EXISTS` is a no-op on fresh tables but correctly sequenced on existing ones.

### Patch

```diff
[...]

Issue #266, opened 2026-04-20
v0.14.x initSchema fails with "column link_source does not exist" when upgrading a PGLite brain from schema v10 (v0.12.x)


gopalpatel:
Hi Garry — running into a chicken-and-egg in `initSchema` when upgrading an existing PGLite brain from v0.12.3 (schema version 10) to master (v0.14.2).

## Reproducer

Starting from a brain at schema version 10 (the v0.12.3 baseline — `drop_timeline_search_trigger`):

```bash
cd ~/gbrain-tmp
git fetch origin && git reset --hard origin/master  # lands on v0.14.2
bun install && bun link
gbrain apply-migrations --yes
```

Output:

```
=== Applying migration v0.12.2: JSONB double-encode repair ===
column "link_source" does not exist
Migration v0.12.2 reported status=failed.
```

Same error from `gbrain init --migrate-only` directly.

## Root cause

`src/core/pglite-schema.ts:82-83` has:

```sql
CREATE INDEX IF NOT EXISTS idx_links_source ON links(link_source);
CREATE INDEX IF NOT EXISTS idx_links_origin ON links(origin_page_id);
```

These reference columns that are added by **migration v11 (`links_provenance_columns`)** in `src/core/migrate.ts:288`. But `PGLiteEngine.initSchema()` runs `PGLITE_SCHEMA_SQL` **before** it calls `runMigrations`:

```ts
// src/core/pglite-engine.ts:63-70
async initSchema(): Promise<void> {
  await this.db.exec(PGLITE_SCHEMA_SQL);   // <-- fails here on existing v10 brain
  const { applied } = await runMigrations(this);
  ...
}
```

`CREATE INDEX IF NOT EXISTS` guards by **index name**, not column existence — so for a brain already carrying `idx_links_from` / `idx_links_to` / etc. from its v10 state, the new `idx_links_source` doesn't exist yet and must be created, but the column it references also doesn't exist yet. SQL fails, `runMigrations` never runs, v11 never applies, user is stuck.

A fresh brain (`gbrain init --pglite`) isn't affected — the CREATE TABLE defines `link_source` in the initial schema, then the indexes create cleanly. Only users who went through the v0.11/v0.12 wave hit this.

## What I tried

1. `gbrain apply-migrations --yes` — fails at Phase A
2. `gbrain apply-migrations --force-retry 0.12.2` → re-run — same failure
3. `gbrain init --migrate-only` directly — same failure
[...]

PR #2623, opened 2026-07-06
fix(schema): bootstrap timeline_entries.event_page_id before schema-blob replay

colinagent:
## Summary

`initSchema` replays the embedded schema blob **before** running numbered migrations. The blob contains

```sql
CREATE INDEX IF NOT EXISTS idx_timeline_event_page ON timeline_entries(event_page_id) ...
CREATE UNIQUE INDEX IF NOT EXISTS idx_timeline_event_dedup ON timeline_entries(event_page_id, date) ...
```

but `event_page_id` is only added by migration **v121** (`timeline_entries_event_page_id`). On any pre-v121 brain the blob replay fails with `column "event_page_id" does not exist` — *before* v121 ever gets a chance to run. Both `gbrain init --migrate-only` and the automatic on-connect migration path wedge permanently, and every CLI command warns:

```
Schema probe failed: column "event_page_id" does not exist
Re-run: `gbrain apply-migrations --yes`
```

(`apply-migrations` also fails, for the same reason — its Phase A schema step hits the identical replay.)

This is the same bootstrap-or-deadlock forward-reference class as #239 / #266 / #357 / #375 (links / chunks / pages columns), which `applyForwardReferenceBootstrap` exists to prevent — v121's column just never got a bootstrap entry.

## Fix

Follows the established pattern exactly:

- `applyForwardReferenceBootstrap` (both engines) probes `timeline_entries.event_page_id` and adds the bare column before the blob replay. The FK constraint and both partial indexes still land through v121's guarded DO block / the blob's `CREATE INDEX IF NOT EXISTS`; v121 stays idempotent.
- New `REQUIRED_BOOTSTRAP_COVERAGE` entry in `schema-bootstrap-coverage.test.ts` (the CI gate this column slipped past).
- New rewind scenario in `bootstrap.test.ts`: drop the column/indexes/FK, run bootstrap, assert the column returns, then run **full `initSchema`** end-to-end to prove the wedge is gone.

## Production verification

Hit in a production Postgres brain at schema v119 running a v0.42.56 binary: migration to v122 was impossible until the column was added manually (`ALTER TABLE timeline_entries ADD COLUMN IF NOT EXISTS event_page_id INTEGER`), after which v120–v122 applied cleanly. With this patch the same upgrade completes without manual intervention.

## Test plan

- [x] `bun test test/bootstrap.test.ts test/schema-bootstrap-coverage.test.ts` — 16 pass (incl. new pre-v121 rewind + full initSchema replay)

### DIFF
--- a/src/core/pglite-engine.ts
+++ b/src/core/pglite-engine.ts
@@ -498,7 +498,11 @@ export class PGLiteEngine implements BrainEngine {
         EXISTS (SELECT 1 FROM information_schema.columns
                 WHERE table_schema='public' AND table_name='pages' AND column_name='embedding_signature') AS pages_embedding_signature_exists,
         EXISTS (SELECT 1 FROM information_schema.columns
-                WHERE table_schema='public' AND table_name='pages' AND column_name='links_extracted_at') AS pages_links_extracted_at_exists
+                WHERE table_schema='public' AND table_name='pages' AND column_name='links_extracted_at') AS pages_links_extracted_at_exists,
+        EXISTS (SELECT 1 FROM information_schema.tables
+                WHERE table_schema='public' AND table_name='timeline_entries') AS timeline_entries_exists,
+        EXISTS (SELECT 1 FROM information_schema.columns
+                WHERE table_schema='public' AND table_name='timeline_entries' AND column_name='event_page_id') AS timeline_event_page_id_exists
     `);
     const probe = rows[0] as {
       pages_exists: boolean;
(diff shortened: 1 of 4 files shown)


time-attack:
Closing as superseded by #2735 (merged in v0.42.59.0): both engines now probe and bootstrap timeline_entries.event_page_id in the startup bootstrap set, pinned by test/schema-bootstrap-coverage.test.ts. Already-wedged brains heal on the next command. Thanks!

PR #2628, opened 2026-07-06
perf(contextual-retrieval): bound per-chunk synopsis concurrency

spiky02plateau:
## Problem

The contextual-retrieval `per_chunk_synopsis` path generates one Haiku synopsis per non-code chunk **strictly sequentially** within a page. At 20–45 s per `gateway.chat()` call, a 120-chunk transcript page takes 60–90+ minutes of wall time and routinely outlives the job timeout — each death then burns `max_attempts × timeout_ms` of API spend before the page can ever land. (Production data point on the timeout side is in #2611.)

## Design

Replace the intra-page sequential loop with a bounded sliding worker pool using the repo's existing `runSlidingPool` helper. Completion order can vary, but synopses land in chunk order via index-addressed writes into a preallocated array. Code chunks continue to bypass synopsis generation (D20-T4), and the final embedding call remains one page-level `embedBatch` after all chunk synopsis work succeeds (D27 P2-2).

## Knob semantics

`GBRAIN_CONTEXTUAL_CHUNK_CONCURRENCY`, default `4`, clamped to `[1,16]`. **Value `1` restores the prior sequential behavior exactly.** The Minion handler resolves the env and passes it into the service; inline callers are unaffected.

The existing global synopsis rate-lease stays the cross-worker governor: each concurrent chunk task individually acquires/releases a lease. The lease id now travels from acquire to release as an opaque token instead of shared mutable handler state (which would have raced under concurrency), and lease waits are abort-responsive.

## Error/abort behavior

Unchanged classification: any chunk failure aborts queued work and routes through the same page-level fallback / transient / permanent paths as before (D27 P1-2), with the first failure by chunk index winning deterministically. Caller abort signals stop queued and in-flight work promptly, including while waiting on a lease.

Two dead imports (`extractFirstTwoSentences`, `logSynopsisFailure`) that were never called in this file are dropped.

## Tests

Added to `test/contextual-retrieval-service-pure.test.ts`: ordered parallel output identical to sequential, bounded max in-flight (fake gateway tracking concurrency), per-chunk lease acquire/release accounting, single chunk failure aborting queued work, prompt abort cancellation, fenced-code bypass, env default/clamp parsing. `bun test` 21/21 on the unit file, 5/5 on `test/e2e/contextual-retrieval-pglite.test.ts`, `bun run typecheck` clean, `scripts/check-test-isolation.sh` OK.

https://claude.ai/code/session_0183oqVjdjWzeuepzNM61bxM

### DIFF
--- a/src/core/contextual-retrieval-service.ts
+++ b/src/core/contextual-retrieval-service.ts
@@ -45,7 +45,6 @@ import { embedBatch } from './embedding.ts';
 import { resolveContextualRetrievalMode } from './contextual-retrieval-resolver.ts';
 import {
   buildContextualPrefix,
-  extractFirstTwoSentences,
   modeRequiresHaiku,
   modeRequiresWrapper,
   sanitizeTitle,
@@ -56,10 +55,8 @@ import {
   SYNOPSIS_PROMPT_VERSION,
   type GeneratePerChunkSynopsisResult,
 } from './page-summary.ts';
-import {
-  logSynopsisFailure,
-  type SynopsisFailureKind,
-} from './audit-synopsis.ts';
+import type { SynopsisFailureKind } from './audit-synopsis.ts';
+import { runSlidingPool } from './worker-pool.ts';
 import type { BrainEngine } from './engine.ts';
 import type { ChunkInput, CRMode, Page } from './types.ts';
 import type { SourceRow } from './sources-ops.ts';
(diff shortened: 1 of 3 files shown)


time-attack (approved):
Approving — not merging.

The concurrency-safe lease refactor is the load-bearing fix and it's at the right seam: `acquireSynopsisLease` now returns the lease and `releaseSynopsisLease(lease)` takes it back, replacing the single shared `currentLeaseId` that concurrent acquires would have clobbered. Behavior at `chunkConcurrency=1` is byte-identical to the old sequential loop.

Correctness verified:
- Order preserved via index-filled `wrappedTexts` array; embed input stays chunk-order regardless of completion order.
- Lowest-index failure chosen for classification, matching the sequential 'first failing chunk wins' semantics across all Phase1Result kinds.
- Lease lifecycle: `leaseAcquired` flag means acquire-throw never triggers a spurious release.
- Abort yields transient/timeout as before, with a redundant `poolResult.aborted` safety net.
- Reuses the CI-guarded sliding pool — no new dependency, no new abstraction.

Tests: 21 pass, hermetic via gateway transport stubs + fake engine (order, boundedness, lease pairing acquired==released, single-failure abort+fallback, fenced-code bypass, prompt abort).

Two non-blocking notes:
1. Peak per-page Haiku concurrency goes 1 → up to chunkConcurrency (default 4); the cross-worker rate lease still governs global concurrency but per-page lease demand rises. Intended tradeoff; worth a line in the PR body.
2. New env knob GBRAIN_CONTEXTUAL_CHUNK_CONCURRENCY isn't in CLAUDE.md's env-knob tables — /document-release territory.

PR #2629, opened 2026-07-06
fix(autopilot,eval): make the nightly quality probe enable path work end-to-end

p3ob7o:
## Summary

Enabling the nightly quality probe (v0.40.1.0 Track D / T6) has never worked. Six defects sat stacked on the same path — each found by running the probe for real after fixing the previous one. With this PR the probe runs nightly, reaches real pass/fail verdicts, and surfaces genuine retrieval signal. Verified end-to-end on a live brain.

1. **The enable hint writes a config plane nobody reads.** Doctor prints `gbrain config set autopilot.nightly_quality_probe.enabled true`, which writes the DB plane. The autopilot gate and doctor both read only the file plane, so following the hint did nothing. Fix: `resolveProbeEnabled` / `resolveProbeMaxUsd` apply the dual-plane rule `mcp.publish_skills` already uses (DB row wins, file plane falls back), shared by both call sites so they can't diverge. Doctor's old `Boolean()` read also disagreed with autopilot's `=== true` on string values; now it can't.

2. **The fixture resolves against the wrong repo.** The wiring passed `resolveRepoRoot = repoPath` — `sync.repo_path`, the user's brain repo — but the committed fixture lives in the gbrain package. Every real install error'd with fixture-not-found. The DI test harness passes `process.cwd()`, which is the gbrain repo in CI, so tests never saw it. Fix: resolve the package root from the module location, keep `repoPath` as fallback.

3. **Rate-limit skips flood the audit file and pin doctor at WARN.** The loop invokes the probe every cycle (~5-10 min); each within-24h skip wrote a `rate_limited` audit row — hundreds a day — and doctor counts any non-pass outcome as bad. Enabling the probe meant a permanent WARN by construction. Fix: skips are non-events. Only real runs write audit rows and gate the 24h window.

4. **Recipe ids passed to the raw SDK.** `eval-longmemeval` handed `resolveModel`'s recipe ids (`anthropic:claude-…`) straight to the raw Anthropic SDK, which 404s on the prefixed string. Every answer and extractor call failed; batches came back all-`upstream_error`. `think` avoids this by routing through the gateway; this eval keeps its raw SDK client for hermeticity, so it now strips the prefix via `splitProviderModelId`.

5. **A judge slot that can never run.** `DEFAULT_SLOTS` still pinned `openai:gpt-4o` after the OpenAI recipe dropped it. Slot A error'd "not listed for OpenAI chat" on every install, and the 3-slot panel couldn't reach its 2-model quorum without a Google key — verdicts were permanently `inconclusive`. Fix: bump to a listed model, plus a slots↔recipes consistency test so defaults can't drift again.

6. **A rubric the fixture can never pass.** The panel scored terse LongMemEval facts ("in widget-co") on the rich-response rubric (DEPTH / SOURCING ≥ 7) with neither the gold answer nor the haystack in view. A healthy brain failed every night. Fix: `eval-longmemeval` now emits the dataset's gold answer per row (additive — `evaluate_qa.py` ignores unknown fields), batch mode folds it into the judge task when present, and the probe adapter passes QA-shaped dimensions (CORRECTNESS + DIRECTNESS). No grounding dimension on purpose: judges who never see the haystack read correct extra detail as invented. A correct "before, with dates" answer scored 4/10 on such a dimension while I tested this.

## Verification

I ran the probe end-to-end on a live brain (Postgres engine, real keys) after each fix. Final state: 9/10 fixture questions pass. The one failure is real — a multi-session aggregation question needs three evidence sessions, retrieval returns two, the answer counts two. (`recall_hit` is true because the recall metric accepts partial overlap.) The probe now reports retrieval signal instead of plumbing artifacts.

That last point needs a maintainer call: with the fixture as-is, a healthy brain sits at 1/10 fail → nightly WARN. Either the aggregation case is a capability edge worth fixing (ranking/limit for count-questions), or the fixture should be calibrated so steady state is 10/10 and any fail means regression. I'm happy to file either as a follow-up.

Measured cost per real run: ~$1.50 (10 questions × 2-of-3 judge slots) against the $5 default cap — about $45/month nightly.

## Tests

- `test/nightly-probe-config-plane.test.ts` (new): dual-plane resolution for both flags.
- `test/cross-modal-default-slots.test.ts` (new): every default slot model must be listed in its recipe's chat touchpoint; slots must span three providers.
- `test/nightly-quality-probe.test.ts` + `test/autopilot-nightly-probe-wiring.test.ts`: updated to pin the new behavior (no rate-limit audit rows, package-root fixture resolution, dual-plane gate shape).
- Cross-modal + probe suites green (164 tests). `bun run typecheck` clean.

### DIFF
--- a/src/commands/autopilot.ts
+++ b/src/commands/autopilot.ts
@@ -1086,12 +1086,23 @@ export async function runAutopilot(engine: BrainEngine, args: string[]) {
       if (probeEnabled) {
         const { runLongMemEvalForProbe, runCrossModalBatchForProbe } = await import('../core/cycle/nightly-probe-adapters.ts');
         const { isAvailable } = await import('../core/ai/gateway.ts');
+        const { existsSync } = await import('node:fs');
+        const { fileURLToPath } = await import('node:url');
+        const { join } = await import('node:path');
         const maxUsd = resolveProbeMaxUsd(dbMaxUsd, cfg?.autopilot?.nightly_quality_probe?.max_usd);
+        // The committed fixture (test/fixtures/longmemeval-nightly.jsonl)
+        // lives in the gbrain PACKAGE, not the brain repo — repoPath is
+        // sync.repo_path (the user's brain), where the fixture never
+        // exists, so the probe error'd on every real install. Resolve the
+        // package root from the module location; keep repoPath as the
+        // fallback for setups that vendor the fixture into the brain repo.
+        const pkgRoot = fileURLToPath(new URL('../..', import.meta.url));
+        const fixtureAtPkgRoot = existsSync(join(pkgRoot, 'test', 'fixtures', 'longmemeval-nightly.jsonl'));
         await runNightlyQualityProbe({
           isEnabled: () => true, // already gated above; phase re-checks for defense-in-depth
           hasEmbeddingProvider: () => isAvailable('embedding'),
           resolveMaxUsd: () => maxUsd,
-          resolveRepoRoot: () => repoPath ?? gbrainHomePath('.'),
+          resolveRepoRoot: () => (fixtureAtPkgRoot ? pkgRoot : repoPath ?? gbrainHomePath('.')),
           runLongMemEval: runLongMemEvalForProbe,
           runCrossModalBatch: runCrossModalBatchForProbe,
           now: () => new Date(),
--- a/src/commands/autopilot.ts
+++ b/src/commands/autopilot.ts
@@ -1073,12 +1073,20 @@ export async function runAutopilot(engine: BrainEngine, args: string[]) {
     // loop. Probe runs even when cycleOk=false (probe may surface signal
     // explaining why the cycle is failing).
     try {
-      const probeEnabled = cfg?.autopilot?.nightly_quality_probe?.enabled === true;
+      const { resolveProbeEnabled, resolveProbeMaxUsd, runNightlyQualityProbe } = await import('../core/cycle/nightly-quality-probe.ts');
+      // Dual-plane read: `gbrain config set` (what the doctor enable hint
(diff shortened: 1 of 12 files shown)


p3ob7o:
CI triage for the `test (5)` failures: all 7 errors were `expected 1280 dimensions, not 1536` in `engine-find-trajectory.test.ts` — a file this PR doesn't touch.

Root cause is a latent shard-order flake, not the PR's changes. That test hardcodes `Float32Array(1536)` vectors but lets `initSchema` size its vector columns from process-global gateway state (`getEmbeddingDimensions()`, default 1280). Whether it passes depends on which files ran before it in the shard: any predecessor that leaves the gateway configured without dims — or a bare CI env — yields `vector(1280)`, and every insert dies. Adding test files reshuffles the weight-packed shards, so any PR that adds tests can trip it. `cosine-rescore-column.test.ts` hit the same class earlier and documents the fix in its own `beforeAll`.

Pushed two things: the same dim-pin fix applied to `engine-find-trajectory.test.ts` (`configureGateway(1536)` in `beforeAll`, `resetGateway` in `afterAll`), and a rebase onto current master (v0.42.57.0).

p3ob7o:
Pushed d309b0e8 — one more fix caught while dogfooding this branch post-rebase: doctor's `nightly_quality_probe_health` reported a days-old run as "Latest:" while its counts included newer runs. Mechanism: `readRecentQualityProbeEvents` walks the current week's audit file first, then the previous week's, so the array tail (which the check reports as Latest) was the OLDEST in-window event whenever last week's file had entries. The reader now returns events sorted oldest → newest; cross-week regression test included. Sister fix for the parser-probe check is on #2630.

time-attack:
Superseded with credit by #3094 (just merged) — your change landed rebased onto current master with authorship preserved via Co-authored-by. Thank you!

PR #2630, opened 2026-07-06
feat(autopilot): wire the nightly conversation-parser probe into the loop

p3ob7o:
## Summary

The v0.41.16.0 conversation-parser nightly probe shipped as a callable module with its scheduler wire-up "deferred to a follow-up (filed in TODOS.md)". The follow-up never landed and the TODO is gone. Today `runConversationParserNightlyProbe` has zero call sites: the enable flag does nothing, the D10 tokenmax default-on never fires, and doctor's `conversation_parser_probe_health` is a hardcoded "Skipped" stub that ignores config and mode.

This PR adds the missing wiring as autopilot step 4.6, mirroring the nightly quality probe (step 4.5):

- The flag reads dual-plane: the DB row (`gbrain config set autopilot.conversation_parser_probe.enabled true`) wins, `~/.gbrain/config.json` falls back (type added to `GBrainConfig.autopilot`). `search.mode=tokenmax` keeps the probe default-on per D10.
- Fixtures resolve from the gbrain package root (the committed `test/fixtures/conversation-formats/`), not the brain `repoPath`. Compiled binaries carry no source tree, so the wiring skips quietly with a once-per-process stderr note instead of writing failure rows that would flip doctor to WARN on every binary install.
- A new audit trail: `audit-parser-probe.ts`, built on the shared audit-writer primitive (`parser-probe-YYYY-Www.jsonl`, honors `GBRAIN_AUDIT_DIR`). `rate_limited` outcomes aren't logged — the loop ticks every few minutes and auditing each skip would flood the file. The 24h gate (`parserProbeRanWithin`) reads real outcomes only.
- The doctor stub becomes a real check (`computeConversationParserProbeHealthCheck`): enable hint when off and silent, OK when enabled with no events yet, WARN on any non-pass run in the 7-day window, all-pass summary otherwise.
- Failure posture matches 4.5: try/catch around the whole block, `logError('autopilot.parser_probe', …)`, never bumps `consecutiveErrors`.

## Notes for review

- Branched from master independently of #2629 (the quality-probe fix). The two touch adjacent blocks of `autopilot.ts` and merge cleanly in either order.
- The probe module itself is unchanged except its header comment (no longer "deferred").

## Tests

- `test/audit-parser-probe.serial.test.ts` (new): log/read round-trip, ISO-week prefix, window filtering, 24h-gate semantics (non-pass outcomes hold the window).
- `test/doctor-parser-probe-check.test.ts` (new): all four branches of the doctor check.
- `test/autopilot-parser-probe-wiring.test.ts` (new): source-shape pins for the dual-plane read, the tokenmax mode-gate, package-root fixture resolution, the flood guard, and the try/catch posture.
- `bun run typecheck` clean; quality-probe wiring tests untouched and green.

### DIFF
--- a/src/commands/autopilot.ts
+++ b/src/commands/autopilot.ts
@@ -527,6 +527,9 @@ export async function runAutopilot(engine: BrainEngine, args: string[]) {
   process.on('SIGINT',  () => { void shutdown('SIGINT'); });
 
   let consecutiveErrors = 0;
+  // Parser-probe fixture warning is once-per-process, not once-per-cycle
+  // (compiled-binary installs have no source tree; don't spam the log).
+  let parserProbeFixtureWarned = false;
   // v0.37.7.0 #1162 — counter for consecutive reconnect failures.
   // Reset on every successful health probe or reconnect. Threshold
   // controlled by GBRAIN_AUTOPILOT_MAX_RECONNECT_FAILS env (default 30).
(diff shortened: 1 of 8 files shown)


p3ob7o:
CI triage for the `test (5)` failures: all 7 errors were `expected 1280 dimensions, not 1536` in `engine-find-trajectory.test.ts` — a file this PR doesn't touch.

Root cause is a latent shard-order flake, not the PR's changes. That test hardcodes `Float32Array(1536)` vectors but lets `initSchema` size its vector columns from process-global gateway state (`getEmbeddingDimensions()`, default 1280). Whether it passes depends on which files ran before it in the shard: any predecessor that leaves the gateway configured without dims — or a bare CI env — yields `vector(1280)`, and every insert dies. Adding test files reshuffles the weight-packed shards, so any PR that adds tests can trip it. `cosine-rescore-column.test.ts` hit the same class earlier and documents the fix in its own `beforeAll`.

Pushed two things: the same dim-pin fix applied to `engine-find-trajectory.test.ts` (`configureGateway(1536)` in `beforeAll`, `resetGateway` in `afterAll`), and a rebase onto current master (v0.42.57.0).

p3ob7o:
Pushed d7b7bc83 — fixes a defect this PR inherited by mirroring the quality-probe doctor check: `readRecentParserProbeEvents` returned events in file order (current week first), so the check's "latest" was the oldest in-window run whenever last week's audit file had entries. Reader now sorts chronologically; cross-week regression test writes both week files directly (the writer routes by write time, so logging through it can't exercise the cross-file case). Sister fix for the upstream quality-probe check is on #2629.

time-attack:
Superseded with credit by #3094 (just merged) — your change landed rebased onto current master with authorship preserved via Co-authored-by. Thank you!

PR #2631, opened 2026-07-06
fix(providers): show user-provided chat recipes as chat-capable

### LINKED ISSUE #2602: providers list should show user-provided chat recipes as chat-capable
## Summary

`gbrain providers list` appears to render user-provided chat recipes as not chat-capable when the recipe intentionally has `touchpoints.chat.user_provided_models = true` and an empty `models` array.

## Observed

In a local DStack/DBrain integration, a LiteLLM-compatible recipe with a user-provided chat model can route chat successfully through the gateway, but `providers list` still displays the chat column as unavailable (`CHAT=-`).

## Expected

Provider capability display should treat `touchpoints.chat.user_provided_models === true` as chat-capable, even when `touchpoints.chat.models.length === 0`.

## Why it matters

The current display makes a working provider route look unsupported, which complicates post-upgrade/runtime audits and can send operators chasing a false provider configuration problem.

## Evidence boundary

This is from local runtime proof against `gbrain 0.42.56.0` at observed head `646179047a8e4ad9c462d83ce9a67a50ba076de8`. No secrets or provider payloads are included here.

hiro-nikaitou:
## Problem

`gbrain providers list` shows user-provided chat recipes (like LiteLLM) as not chat-capable because `hasChat` only checks `chat.models.length > 0`, ignoring the `user_provided_models` flag that proxy-style recipes use when they ship without a fixed model list.

Closes #2602.

## Changes

1. **ChatTouchpoint interface** (`src/core/ai/types.ts`) — added `user_provided_models?: true` field matching the existing `EmbeddingTouchpoint.user_provided_models`.
2. **formatRecipeTable** (`src/commands/providers.ts`) — updated `hasChat` to treat `user_provided_models === true` as chat-capable even when `models` is empty.
3. **litellm-proxy recipe** (`src/core/ai/recipes/litellm-proxy.ts`) — added a `chat` touchpoint with `user_provided_models: true` so `providers list` correctly shows CHAT=yes.

## Verification

Tested on **Windows 11 (Git Bash)**:
- `bun test test/providers.test.ts` — 10/10 pass
- `bun test test/ai/recipe-*.test.ts` — 58/58 pass
- `bun run typecheck` — clean

### DIFF
--- a/src/commands/providers.ts
+++ b/src/commands/providers.ts
@@ -73,7 +73,7 @@ export function formatRecipeTable(recipes: Recipe[], env: NodeJS.ProcessEnv = pr
   for (const r of recipes) {
     const hasEmbed = !!r.touchpoints.embedding && (r.touchpoints.embedding.models.length > 0);
     const hasExpand = !!r.touchpoints.expansion;
-    const hasChat = !!r.touchpoints.chat && r.touchpoints.chat.models.length > 0;
+    const hasChat = !!r.touchpoints.chat && (r.touchpoints.chat.models.length > 0 || r.touchpoints.chat.user_provided_models === true);
     const ready = envReady(r, env);
     const status = ready ? '✓ ready' : `✗ missing ${r.auth_env?.required?.[0] ?? 'setup'}`;
     rows.push(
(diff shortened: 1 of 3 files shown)


hiro-nikaitou:
No response after 2+ days. Closing to keep the queue clean.

PR #2632, opened 2026-07-06
fix(config): register Life Chronicle keys so the documented enable command works

p3ob7o:
## Summary

The v0.42.56.0 release notes tell users to enable Life Chronicle with `gbrain config set auto_chronicle true`. That command fails:

```
[config] Unknown config key "auto_chronicle".
[config] No similar known key. Run `gbrain config show` to see currently-set keys.
```

The key was never registered in `KNOWN_CONFIG_KEYS`, so the documented enable path dead-ends and the operator has to discover `--force` by reading source — the exact class the v0.42.42.0 spend-controls registration fixed for `spend.posture` and friends.

This registers `auto_chronicle` in `KNOWN_CONFIG_KEYS` and adds the `chronicle.` prefix to `KNOWN_CONFIG_KEY_PREFIXES` (covers `chronicle.tz` and future Life Chronicle knobs). A regression test pins both, so a future chronicle key that ships documented-but-unregistered fails CI.

Found while enabling the feature for real on a live brain (same session as #2629/#2630).

## Tests

- `test/config.test.ts`: new case asserting `auto_chronicle` is in `KNOWN_CONFIG_KEYS` and `chronicle.tz` matches a registered prefix. Full file green (23 tests); `bun run typecheck` clean.

### DIFF
--- a/src/core/config.ts
+++ b/src/core/config.ts
@@ -921,6 +921,12 @@ export const KNOWN_CONFIG_KEYS: readonly string[] = [
   // operator had to discover these by reading source. Registered so `config
   // set` accepts them directly. See docs/operations/spend-controls.md.
   'spend.posture',
+  // Life Chronicle (v0.42.56.0, #2390). The release notes' enable command is
+  // `gbrain config set auto_chronicle true`, but the key was never registered
+  // — so the documented command failed with "Unknown config key" and the
+  // operator had to discover --force by reading source. Same class as the
+  // spend-controls registration above.
+  'auto_chronicle',
   'sync.cost_gate_min_usd',
   'sync.federated_v2',
   'embed.backfill_cooldown_min',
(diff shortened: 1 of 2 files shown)


p3ob7o:
Added `takes.bootstrap_enabled` to the same registration + regression test — hit the identical unregistered-key failure while enabling the takes bootstrap on a live brain. Same class, one PR.

time-attack (approved):
Reviewed the diff vs master. Clean, minimal, root-caused fix: registers `auto_chronicle`, `takes.bootstrap_enabled`, and the `chronicle.` prefix in the `config set` known-key registry so the documented enable commands work without `--force`. Verified every registered key matches its actual `getConfig(...)` call site. Registry is suggestion-source only (not a runtime allow-list), so no behavior change beyond `config set` acceptance — effectively zero risk. Test coverage is proportionate. Approved (not merging).

PR #2637, opened 2026-07-06
fix(dream): stamp incremental extraction watermark

### LINKED ISSUE #2636: Dream incremental extract processes pages but never stamps links_extracted_at
## Summary

The default Dream cycle can successfully process changed pages in its incremental extract phase while leaving `pages.links_extracted_at` unchanged. Those pages remain permanently visible to `gbrain extract --stale` and `links_extraction_lag` until an operator runs a separate stale sweep.

This is a gap in the #1696 / #1755 maintenance-loop fix.

## Reproduction

Run Dream with both `sync` and `extract` enabled against a non-default source after changing pages.

Observed production example:

- Dream sync: `+6 added, ~6 modified`
- Dream extract: `0 link(s), 0 timeline entries (incremental: 12 slugs)`
- `pages_processed: 12`
- all 12 pages retained stale or NULL `links_extracted_at`

Across the affected brain, doctor reported 84 stale pages: 63 had NULL watermarks and 21 had `updated_at > links_extracted_at`. Forty-two of those pages already had outgoing links, confirming that extraction work and freshness accounting had diverged.

## Root cause

`runCycle` disables sync inline extraction when the separate extract phase is selected. It then calls:

`runPhaseExtract -> runExtractCore({ slugs }) -> extractForSlugs`

The incremental `extractForSlugs` path flushes link and timeline batches but never calls `stampExtracted` or `markPagesExtractedBatch`. It also does not receive the cycle source id, so source-correct stamping is impossible there.

When the next sync is up to date, Dream passes `slugs: []`, returns immediately, and never drains the accumulated stale backlog.

## Expected behavior

After both link and timeline batches successfully flush, incremental all-mode extraction should stamp every processed page using the cycle source id. Dry runs and links-only/timeline-only runs must not stamp the combined watermark.

## Workaround

`gbrain extract --stale --source-id <source>` clears the accumulated backlog. In the production example it converged from 84 stale pages to 0 and added 7 links plus 11 timeline entries.

## Proposed fix

- Thread `cycleSourceId` through `runPhaseExtract` and `runExtractCore`.
- Source-qualify incremental link/timeline batch rows.
- After both batches flush, stamp processed pages in non-dry-run `mode: all`.
- Add a non-default-source regression test asserting `countStalePagesForExtraction(...) === 0` after incremental extraction.

JavanC:
Closes #2636.

## Problem

When Dream runs `sync` and `extract` together, it disables sync inline extraction and sends `syncPagesAffected` through `runExtractCore({ slugs })`. The incremental `extractForSlugs` path flushes links and timeline rows but never stamps `links_extracted_at`.

The work succeeds while doctor and `extract --stale` continue reporting the same pages as stale. An up-to-date next sync passes an empty slug set, so the backlog is never drained by Dream.

## Fix

- Thread the resolved cycle source id into incremental extraction.
- Source-qualify incremental link and timeline batch rows.
- Track successfully processed pages.
- Stamp the combined extraction watermark only after both batches flush, only for non-dry-run `mode: all`.
- Add a non-default-source regression asserting the page is no longer stale.

## Validation

- `bun test test/extract-incremental.test.ts test/core/cycle.serial.test.ts test/cycle-abort.test.ts test/sync-inline-extract-stamps.serial.test.ts test/extract-stale.test.ts`
  - 63 pass, 0 fail
- `bun run typecheck`
  - pass
- `git diff --check`
  - pass

`bun run verify` on clean upstream/master remains affected by the existing macOS no-`timeout` fallback accounting bug: all 31 checks complete successfully but are recorded as rc=143. This branch does not modify the verify runner.

### DIFF
--- a/src/commands/extract.ts
+++ b/src/commands/extract.ts
@@ -516,6 +516,11 @@ export interface ExtractOpts {
    */
   slugs?: string[];
   /**
+   * Source owning `dir` when the incremental cycle path supplies `slugs`.
+   * Required for source-correct edge writes and extraction watermark stamps.
+   */
+  sourceId?: string;
+  /**
    * v0.41.15.0 (D9): in-process parallel file workers for the fs-walk
    * loops. Default 1. PGLite engines clamp to 1 (single-writer; though
    * extract is mostly CPU-bound, the DB batch flush still hits the
@@ -574,7 +579,17 @@ export async function runExtractCore(engine: BrainEngine, opts: ExtractOpts): Pr
       // Nothing changed — skip entirely.
       return result;
     }
-    const r = await extractForSlugs(engine, opts.dir, opts.slugs, opts.mode, dryRun, jsonMode, workers, opts.signal);
+    const r = await extractForSlugs(
+      engine,
+      opts.dir,
+      opts.slugs,
+      opts.mode,
+      dryRun,
+      jsonMode,
+      workers,
+      opts.signal,
+      opts.sourceId,
+    );
     result.links_created = r.links_created;
     result.timeline_entries_created = r.timeline_created;
     result.pages_processed = r.pages;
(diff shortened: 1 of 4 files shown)


time-attack:
Superseded with credit by #3115 (just merged) — your change landed rebased onto current master with authorship preserved via Co-authored-by. Thank you!

PR #2638, opened 2026-07-06
fix(takes): bootstrap runs progress through the corpus instead of rescanning the newest slice

p3ob7o:
## Summary

The takes bootstrap can't finish a corpus larger than one run. `extractTakesFromPages` selects pages by `updated_at DESC` + `LIMIT`, with no exclusion of pages that already hold takes, and the CLI clamps `--max-pages` to 1000. So on a brain with more than 1000 eligible pages, every re-run rescans the same most-recent slice: the older tail is unreachable, and each rescan re-spends Haiku budget producing upsert-identical rows.

Hit this live: a brain with 2,311 eligible pages ran the documented sweep, covered the newest 1,000, and a second run would have covered 0 new pages.

The fix: covered pages are skipped by default (`NOT EXISTS` against `takes.page_id`), so repeat runs progress through the corpus in slices — the natural bootstrap semantics. `--include-covered` restores the full rescan for refresh use. The usage text now documents both, plus the previously invisible 1000 clamp.

## Tests

`test/extract-takes-from-pages-progression.test.ts` (new, hermetic PGLite + stubbed chat transport): first run covers the corpus, second run scans 0 (progression), a page added later is picked up, and `--include-covered` rescans everything. Gateway state is pinned and reset per the isolation rules. `bun run typecheck` clean.

### DIFF
--- a/src/commands/takes.ts
+++ b/src/commands/takes.ts
@@ -591,7 +591,8 @@ async function cmdExtract(engine: BrainEngine, rest: string[]): Promise<void> {
   const sub = rest[0];
   if (sub !== '--from-pages') {
     process.stderr.write(
-      'Usage: gbrain takes extract --from-pages [--yes] [--dry-run] [--source-id <id>] [--max-pages N] [--holder <name>]\n',
+      'Usage: gbrain takes extract --from-pages [--yes] [--dry-run] [--source-id <id>] [--max-pages N (clamped to 1000)] [--include-covered] [--holder <name>]\n' +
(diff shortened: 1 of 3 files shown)


p3ob7o:
One known limitation of the NOT EXISTS approach, seen on the live run: pages that yield zero claims never get takes rows, so they stay 'uncovered' and are rescanned by every future bootstrap run (197 of 2,327 pages here). Bounded cost — bootstrap runs are rare and manual — but if you'd rather pin scanned-ness exactly, the alternative is a scanned marker (audit row or sentinel) instead of inferring coverage from takes existence. Happy to rework if you prefer that shape.

PR #2642, opened 2026-07-07
fix(embedding): per-model native dims for Ollama + explicit dims for …

### LINKED ISSUE #2170: Ollama provider hardcodes 768d, rejects 1024d models (bge-m3, mxbai-embed-large)

Willisbest:
…local providers (#2170)

GBrain pins one `default_dims` per provider, so Ollama models with different native widths can't be used, and user-provided-model local providers (llama-server, litellm) reject an explicit dimension:
- ollama:bge-m3 / ollama:mxbai-embed-large emit 1024d, but the Ollama recipe's single default_dims: 768 builds a 768d schema -> `gbrain embed` fails ("returned 1024 but schema expects 768"), and --embedding-dimensions 1024 is refused outright.
- llama-server / litellm (user_provided_models: true, default_dims: 0) hit the same Tier-3 rejection even with an explicit --embedding-dimensions.

Root cause: validateDimAgainstTouchpoint() (src/core/embedding-dim-check.ts) computes `dim = requestedDims ?? defaultDims` from the single provider-wide default_dims (no per-model notion), and isCustomDimValidForProvider() only accepts custom widths for the Matryoshka allow-list (voyage/zeroentropyai/openai). src/commands/init.ts also reads default_dims directly in three spots, bypassing the resolver.

Fix:
- Add optional model_dims?: Record<string, number> to EmbeddingTouchpoint (types.ts).
- Add effectiveDefaultDims(tp, modelId); the resolver uses the per-model width for both schema sizing and the explicit-dim comparison, so the native width is chosen with no flag and an explicit flag equal to it is accepted.
- Ollama recipe declares model_dims (nomic-embed-text 768, bge-m3 1024, mxbai-embed-large 1024, all-minilm 384) and adds bge-m3 to models; default_dims: 768 kept as fallback.
- isCustomDimValidForProvider() accepts any explicit dimension for user_provided_models local recipes (llama-server, litellm); the user's --embedding-dimensions is authoritative (positive-int + pgvector cap still enforced; a real mismatch surfaces at the vector(N) column).
- init.ts: the three direct default_dims reads now route through effectiveDefaultDims.

Tests (test/embedding-dim-check.test.ts): ollama bge-m3->1024, mxbai->1024, nomic->768, all-minilm->384; explicit 1024 accepted; a contradicting dim rejected; llama-server/ litellm accept explicit dims (incl. 896); no-dims rejected; a contract guard that model_dims keys are a subset of models. Red before, green after; tsc --noEmit clean.

Fixes #2170

### DIFF
--- a/src/commands/init.ts
+++ b/src/commands/init.ts
@@ -10,6 +10,7 @@ import { saveConfig, loadConfig, loadConfigFileOnly, toEngineConfig, gbrainPath,
 import { createEngine } from '../core/engine-factory.ts';
 import { discoverOAuth, mintClientCredentialsToken, smokeTestMcp } from '../core/remote-mcp-probe.ts';
 import { runInitEmbedCheck } from '../core/init-embed-check.ts';
+import { effectiveDefaultDims } from '../core/embedding-dim-check.ts';
 
 export async function runInit(args: string[]) {
   // Help guard: cli.ts only routes --help to printOpHelp() for shared-op
@@ -246,7 +247,7 @@ async function resolveAIOptions(opts: ResolveAIOptionsArgs): Promise<ResolvedAIO
       process.exit(1);
     }
     out.embedding_model = `${shorthand}:${firstModel}`;
-    out.embedding_dimensions = recipe.touchpoints.embedding!.default_dims;
+    out.embedding_dimensions = effectiveDefaultDims(recipe.touchpoints.embedding!, firstModel);
   }
 
   if (dimsArg !== null && !Number.isNaN(dimsArg) && dimsArg > 0) {
@@ -271,7 +272,11 @@ async function resolveAIOptions(opts: ResolveAIOptionsArgs): Promise<ResolvedAIO
       process.exit(1);
     }
     if (recipe?.touchpoints.embedding?.default_dims) {
-      out.embedding_dimensions = recipe.touchpoints.embedding.default_dims;
+      // #2170: resolve the SELECTED model's native width (e.g. ollama:bge-m3 →
+      // 1024), not the provider-wide default_dims, so the schema isn't sized
+      // wrong when no --embedding-dimensions was passed.
+      const modelId = out.embedding_model.slice(providerId.length + 1);
+      out.embedding_dimensions = effectiveDefaultDims(recipe.touchpoints.embedding, modelId);
     }
   }
 
@@ -436,7 +441,7 @@ async function resolveEmbeddingByEnv(out: ResolvedAIOptions, nonInteractive: boo
         await import('../core/ai/defaults.ts');
       const dims = fullModel === DEFAULT_EMBEDDING_MODEL
         ? DEFAULT_EMBEDDING_DIMENSIONS
-        : tp.default_dims;
+        : effectiveDefaultDims(tp, model);
       out.embedding_model = fullModel;
       out.embedding_dimensions = dims;
       console.error(
--- a/src/core/ai/recipes/ollama.ts
+++ b/src/core/ai/recipes/ollama.ts
@@ -13,8 +13,20 @@ export const ollama: Recipe = {
   },
   touchpoints: {
     embedding: {
-      models: ['nomic-embed-text', 'mxbai-embed-large', 'all-minilm'],
-      default_dims: 768, // nomic-embed-text native dim
+      models: ['nomic-embed-text', 'bge-m3', 'mxbai-embed-large', 'all-minilm'],
+      default_dims: 768, // nomic-embed-text native dim (fallback for models absent from model_dims)
+      // #2170: Ollama models emit DIFFERENT fixed native widths, but one
+      // provider-wide default_dims silently built a 768d schema for 1024d models
+      // (bge-m3, mxbai-embed-large) → `gbrain embed` then failed with a dim
+      // mismatch. Declare each model's real native width so the schema is sized
+      // correctly with no flag, and an explicit --embedding-dimensions matching
+      // the native width is accepted.
+      model_dims: {
+        'nomic-embed-text': 768,
+        'bge-m3': 1024,
+        'mxbai-embed-large': 1024,
+        'all-minilm': 384,
+      },
       cost_per_1m_tokens_usd: 0,
       price_last_verified: '2026-04-20',
       // Ollama's batch capacity depends on the locally loaded model + the
--- a/src/core/ai/types.ts
+++ b/src/core/ai/types.ts
@@ -28,6 +28,18 @@ export interface EmbeddingTouchpoint {
   models: string[];
   default_dims: number;
   dims_options?: number[]; // for Matryoshka-aware providers
+  /**
+   * #2170: per-model native embedding width, for providers whose models emit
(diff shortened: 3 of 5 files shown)


time-attack:
Thanks for the thorough root-cause trace on #2170 — it matched what the fix ended up being. Both pieces are already on master: the Ollama recipe now carries a per-model `model_dims` map (bge-m3 and mxbai at 1024, all-minilm at 384) resolved via `model-resolver.ts`, and llama-server/litellm accept explicit `--embedding-dimensions` through the `trust_custom_dims` passthrough tier in `embedding-dim-check.ts`. Closing as already fixed.

PR #680, opened 2026-05-06
v0.28.7 fix: adaptive embed batch sizing for Voyage token limits

garrytan-agents:
## What

Voyage tokenizer is 3-4x denser than OpenAI tiktoken, causing batches of 50+ texts to exceed the 120K token-per-batch limit even when DB token counts (from tiktoken) suggest they would fit.

This caused an infinite retry loop during secondary embedding backfill: the same oversized batch failed with `max allowed tokens per submitted batch is 120000`, then `WHERE embedding IS NULL ORDER BY id LIMIT 50` re-fetched the identical rows on the next iteration. 68K chunks (26% of corpus) were left un-embedded.

## Root Cause

The gateway `embed()` function passed all texts to `embedMany()` in a single call with no awareness of provider token limits. OpenAI is permissive here, but Voyage enforces a hard 120K token-per-batch limit — and its tokenizer produces ~3-4x more tokens per character than tiktoken on mixed content (conversations, code, markdown).

## Changes

1. **`EmbeddingTouchpoint.max_batch_tokens`** — new optional field so recipes can declare their provider token-per-batch limit
2. **Voyage recipe** — sets `max_batch_tokens: 120_000`
3. **Gateway `embed()`** — auto-splits into sub-batches using conservative 1:1 char-to-token estimate at 80% budget utilization
4. **`embedSubBatch()`** — on token-limit errors, recursively halves the batch and retries (down to single texts before giving up)
5. **`embedding.ts`** — reduces outer `BATCH_SIZE` from 100 to 50 as secondary guard
6. **Tests** — unit tests for batch splitting logic and error pattern matching

## Testing

- `bun test test/ai/adaptive-embed-batch.test.ts` — 9/9 pass
- `bun test test/ai/gateway.test.ts` — 20/20 pass (no regression)
- `npx tsc --noEmit` — clean
- Validated against real Voyage 4 backfill: old script hit 120K limit at 43K chunks; adaptive logic auto-splits those batches

PR #2643, opened 2026-07-07
fix(ai): enforce provider item-count batch caps (DashScope hard-limits 10 items/request)

YiconZiwei:
## Problem

DashScope's OpenAI-compatible `/embeddings` endpoint enforces a hard **item-count** cap: at most **10 texts per request**, regardless of token size. The error is:

```
<400> InternalError.Algo.InvalidParameter: Value error, batch size is invalid, it should not be larger than 10.: input.contents
```

Two gaps in the gateway made this unrecoverable:

1. **Pre-split is token-budget only.** `splitByTokenBudget()` never counts items, and `EmbeddingTouchpoint` has no item-count field — many short chunks easily pack >10 items into one request while staying far under the token budget.
2. **The halving safety net never engages.** `isTokenLimitError()`'s five patterns all require token-ish wording; the DashScope message contains none of it, so the whole `embed()` call fails — including sibling batches that would have succeeded.

Real-world effect: pages with many short chunks fail to embed, and repeatedly hand-tuning `max_batch_tokens` downward (8192 → 2000 → 800) is a probability game — there is no finite token budget that *guarantees* ≤10 items when chunk length has no lower bound.

## Fix (pure-additive, default-off)

- **types.ts** — add optional `EmbeddingTouchpoint.max_batch_items`. Unset ⇒ behavior is byte-identical for every existing recipe.
- **gateway.ts** — add `splitByItemLimit()`: a second subdivision pass after the token pre-split (no-op when the field is unset). Add `isBatchSizeError()` as a semantic sibling of `isTokenLimitError()`; item-count misses reuse the same recursive-halving recovery (halving reduces item counts too) but do **not** call `shrinkOnMiss()` — the miss has no token-budget cause, so tightening the token safety factor would pollute the adaptive state.
- **dashscope.ts** — declare `max_batch_items: 10`, restore `max_batch_tokens` to 8192 (it no longer has to double as an item-count proxy), add `text-embedding-v4` to the models list (registerExtendedModel already allows it at runtime; the static list was stale).

Shape precedent: this is a direct extension of the `max_batch_tokens` / `isTokenLimitError` mechanism (#680).

## Tests

- `splitByItemLimit` pure-helper describe: empty input, ≤limit passthrough, subdivision, independent per-batch splitting, 0/negative/undefined guard, order preservation.
- `isBatchSizeError` pure-helper describe: real DashScope message verbatim, generic variants, no false positives on token-limit/unrelated errors, non-Error throwables.
- Live-bug-shape regression: dashscope recipe + 47 short texts ⇒ transport receives exactly `[10, 10, 10, 10, 7]`, zero errors.
- Item-miss recovery: recipe *without* `max_batch_items` hitting an item-count error ⇒ halving engages, shrink state stays untouched.
- `recipe-dashscope.test.ts`: asserts `max_batch_items: 10` + v4 in models.

`bun test test/ai/`: **328 pass / 0 fail.**

Also verified against the real DashScope API (China endpoint, real key): 12 items direct ⇒ HTTP 400 with the message above; 15 items through the patched gateway ⇒ 15 vectors @1024, no 400.

### DIFF
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -1396,13 +1396,20 @@ export async function embed(texts: string[], opts?: EmbedOpts): Promise<Float32A
 
   const embedding = recipe.touchpoints?.embedding;
   const maxBatchTokens = embedding?.max_batch_tokens;
+  const maxBatchItems = embedding?.max_batch_items;
   const charsPerToken = embedding?.chars_per_token ?? DEFAULT_CHARS_PER_TOKEN;
 
   // Pre-split is gated on max_batch_tokens. Recipes without it (e.g. OpenAI)
   // ride the fast path: one embedMany call, no recursion safety net.
-  const batches = maxBatchTokens
+  let batches = maxBatchTokens
     ? splitByTokenBudget(truncated, Math.floor(maxBatchTokens * effectiveSafetyFactor(recipe)), charsPerToken)
     : [truncated];
+  // Providers with a hard ITEM-COUNT ceiling (DashScope: ≤10/request) need a
+  // second subdivision pass — token budget alone can still pack too many
+  // short texts into one request. No-op when max_batch_items is unset.
+  if (maxBatchItems) {
+    batches = splitByItemLimit(batches, maxBatchItems);
+  }
 
   const allEmbeddings: Float32Array[] = [];
   let _embedThrew = false;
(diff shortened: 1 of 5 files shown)


time-attack:
DashScope support was declined for now, so closing this with the set — BUT the generic item-count batch-cap design here was the best of the family and is being implemented in the provider-recipes work with Co-authored-by credit. Thank you.

PR #2648, opened 2026-07-07
v0.42.58.1 fix: pre-v121 PGLite brains crash migrating to v0.42.x

numanyilmaz:
## Summary

Fixes a crash that blocks any pre-v121 PGLite brain from upgrading to v0.42.x.

**The bug.** A brain built on an older schema has a `timeline_entries` table without the `event_page_id` column (added in migration v121). On upgrade, `initSchema()` replays `PGLITE_SCHEMA_SQL`, which builds `idx_timeline_event_page` and `idx_timeline_event_dedup` on that column. `CREATE TABLE IF NOT EXISTS timeline_entries` is a no-op against the already-present table, so the column never appears and the `CREATE INDEX` aborts with `column "event_page_id" does not exist`. That wedges every migration entry point: `gbrain init --migrate-only`, `gbrain apply-migrations`, and the post-upgrade chain.

**The fix.** `applyForwardReferenceBootstrap()` exists precisely to add forward-referenced columns before the schema replay, but the v121 column was never registered with it. This adds `timeline_entries.event_page_id` to the probe set and the ALTER block, matching every sibling forward-reference (a plain `INTEGER`; migration v121 still owns the foreign key and both indexes, and stays idempotent). The matching `REQUIRED_BOOTSTRAP_COVERAGE` entry is added so the coverage test guards it — a future forward-referenced column that skips the bootstrap now fails CI instead of a user's upgrade.

Verified end-to-end on a real brain: the jump from schema v24 to v122 went from a hard crash to 97 migrations applied in a single pass.

Commits:
- `fix:` pre-v121 PGLite brains crash migrating to v0.42.x (bootstrap coverage for `timeline_entries.event_page_id`)
- `docs:` note the column in the KEY_FILES bootstrap-coverage list

## Test Coverage

`test/schema-bootstrap-coverage.test.ts` gains the `timeline_entries.event_page_id` entry. That suite runs all 122 migrations against an in-memory PGLite brain and asserts the bootstrap covers every declared forward reference, so it exercises the new branch directly.

Directly-relevant suites: **270 pass, 0 fail** (`test/pglite-engine.test.ts`, `test/schema-bootstrap-coverage.test.ts`, `test/migrate.test.ts`).

## Pre-Landing Review

No issues found. The only DDL is an idempotent additive `ALTER TABLE ... ADD COLUMN IF NOT EXISTS event_page_id INTEGER`; probes are read-only `information_schema` queries with static literals (no injection surface); the guard fires only when the table exists but the column does not, so fresh installs and modern brains both no-op.

## Design Review

No frontend files changed — design review skipped.

## Eval Results

No prompt-related files changed — evals skipped.

## Plan Completion

No plan file detected.

## Documentation

Updated `docs/architecture/KEY_FILES.md`: `timeline_entries.event_page_id` added to the enumerated `applyForwardReferenceBootstrap` coverage list and noted as the pre-v121 wedge this closes.

## Test plan

- [x] Directly-relevant suites pass (270 tests across 3 files, 0 failures)
- [x] Bootstrap coverage test exercises the new column against an in-memory pre-v121 brain
- [x] Verified live: schema v24 → v122 applies 97 migrations with no crash


time-attack:
Closing during backlog triage: the timeline_entries.event_page_id forward-reference bootstrap fix already landed on master. Thanks for catching it!

PR #2649, opened 2026-07-07
lint: add DEFAULT_LINT_EXCLUDES and --exclude flag for mixed-content repos

ryangu00:
When a brain data repo also contains software-repo style content (generated docs, nested repos, test fixtures, project metadata), `gbrain lint` produces noisy findings on files that are not knowledge pages.

This PR adds:
- `DEFAULT_LINT_EXCLUDES` for common non-knowledge paths
- a repeatable `--exclude <glob>` flag as an escape hatch

**Repro**: run `gbrain lint` on a brain repo containing a `docs/` tree from a software project — hundreds of frontmatter/structure findings on files that were never meant to be pages. With this patch: `gbrain lint --exclude 'docs/**'`, or rely on the defaults.

No behavior change for repos without excluded paths.

### DIFF
--- a/src/commands/lint.ts
+++ b/src/commands/lint.ts
@@ -378,15 +378,39 @@ async function resolveLintContentSanity(
   };
 }
 
+// forge-patch: lint exclude for software-repo docs
+// Knowledge-repo lint should not scan vendored deps, nested software trees,
+// or repo metafiles (CHANGELOG etc.) — those produce thousands of false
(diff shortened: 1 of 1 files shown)


ryangu00:
Superseded by #3162 — closing in favor of the takeover. The opt-in `--exclude` design there is the right call for upstream: my hardcoded default list silently changed lint counts for existing repos and carried fork-specific conventions. Defaults like that belong in the caller, not the tool. Thanks @time-attack for the salvage and the credit.

PR #2650, opened 2026-07-07
pricing: resolve bare and provider-prefixed model ids via fallback lookup

ryangu00:
`pricingFor()` hard-fails when a caller passes a model id that is not a literal key of the pricing table — e.g. a bare `claude-haiku-4-5` (no version suffix) or a provider-prefixed id. After v0.42.x this crashed the contradiction-probe cost tracker on non-table judge ids.

This PR makes pricing resolution degrade gracefully:
1. exact match (unchanged)
2. provider/tail lookup (strip provider prefix, match tail)
3. versioned fallback (`id` → newest `id-*` row) with a `??` default

Cost estimates prefer a known equivalent price row over a hard failure.

**Repro**: call the eval/contradictions path with judge id `claude-haiku-4-5` on v0.42.56.0 — TypeError before, priced estimate after.

### DIFF
--- a/src/core/anthropic-pricing.ts
+++ b/src/core/anthropic-pricing.ts
@@ -1,37 +1,48 @@
 /**
- * Anthropic chat pricing — a bare-keyed VIEW of the canonical pricing table
- * (`src/core/model-pricing.ts`).
+ * v0.28: Anthropic model pricing constants for the dream-cycle budget meter.
  *
- * Kept as a distinct export because many callers look up by bare Claude id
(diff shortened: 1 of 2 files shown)


time-attack:
This PR's diff reverts anthropic-pricing.ts from a derived view of the canonical pricing table back to a hand-maintained duplicate — the exact drift pattern the later consolidation was built to prevent. The crash it targets also no longer reproduces: master's canonical table already carries the bare `claude-haiku-4-5` key, so `pricingFor` hits the exact-match branch directly. Closing as superseded by the pricing consolidation.

PR #2652, opened 2026-07-07
fix(mcp): treat stdio serve as a trusted local pipe (remote=false)

poudelprakash:
## Problem

Over the **stdio** MCP transport (Claude Desktop / Claude Code / any local pipe), `whoami` and **every write op** fail-closed:

```
{"error":"unknown_transport","message":"whoami called over a remote transport that did not thread ctx.auth. ..."}
```

`src/mcp/server.ts` dispatches stdio tool calls with `remote: true`, but a local pipe never threads `ctx.auth` (there's no per-token auth), so auth-gated ops throw. Reads work; writes and `whoami` are unusable over stdio.

## Fix

Treat stdio serve as the **trusted local pipe it is** — the same trust boundary as `gbrain call`, which already passes `remote: false` (`src/cli.ts`). This is exactly the whoami op's own prescribed remedy: *"populate ctx.auth OR set ctx.remote === false."*

- Extract the stdio dispatch context into an exported `stdioDispatchDefaults()` helper and set `remote: false`.
- **Private takes stay private:** `takesHoldersAllowList: ['world']` is an independent filter (`WHERE holder = ANY(...)`), not coupled to `remote`. Agent-facing stdio still sees only world-visible takes via `takes_list` / `takes_search` / `query`. (The old comment conflated the two — the allow-list, not `remote`, is what gates hunches.)

## Test

Adds `test/mcp-stdio-dispatch.test.ts` pinning the defaults (`remote === false`, allow-list preserved) — pure, no server/DB, matching `test/whoami.test.ts`'s op-contract style.

## Verification

```
bun test test/mcp-stdio-dispatch.test.ts test/whoami.test.ts   # 12 pass
bun run verify                                                  # 31 checks green
```

Manually confirmed against a freshly-spawned `gbrain serve`: `whoami` → `{transport:"local"}` and `put_page` writes succeed.

### DIFF
--- a/src/mcp/server.ts
+++ b/src/mcp/server.ts
@@ -15,6 +15,35 @@ import {
 } from '../core/context/resolve-ipc.ts';
 import { resolveEntitiesToPointers, logDeliveredReflexPointers } from '../core/context/retrieval-reflex.ts';
 
+/**
+ * Dispatch context defaults for the stdio MCP transport.
+ *
+ * stdio serve runs over a local pipe on the owner's own machine — the same
+ * trust boundary as `gbrain call` (which passes `remote: false` in src/cli.ts).
+ * Treating it as remote/untrusted (`remote: true`) meant `whoami` and every
+ * write op fail-closed with `unknown_transport`, because a local pipe never
+ * threads `ctx.auth` (there is no per-token auth). The whoami op's own error
+ * text prescribes the fix: "populate ctx.auth OR set ctx.remote === false".
+ *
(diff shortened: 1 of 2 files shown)


time-attack:
Closing after review: this inverts the repo's fail-closed trust invariant — stdio dispatch deliberately runs remote:true unless the operator explicitly opts in, and weakening that default is a security regression. The convenience goal is understood; the trust boundary has to stay fail-closed.

PR #2654, opened 2026-07-07
fix(lint): guard wrapping-fence strip behind code-fence-wrap predicate

hanksuper:
## Problem

`fixContent()` in `src/commands/lint.ts` ran the two wrapping-fence strip replaces **unconditionally** on every call. Because `runLintCore()` invokes `fixContent()` whenever *any* fixable issue is present (e.g. `frontmatter-nested-quotes`, `llm-preamble`), an unrelated fixable issue triggered a full rewrite and silently stripped a legit trailing ``` that closed an **interior** code block.

Result: code-heavy pages whose last block is a fenced snippet (e.g. technical docs) lose their closing fence, the code block runs to EOF, and the next lint pass sees an unbalanced fence.

## Root cause

Detection and fix were decoupled. The `code-fence-wrap` *rule* requires both an opening ```` ```markdown ````/```` ```md ```` fence AND a matching closing fence, but `fixContent()` applied the closing-fence strip unconditionally:

```ts
// before — runs on every fixContent() call, regardless of the rule firing
fixed = fixed.replace(/^```(?:markdown|md)\s*\n/, '');
fixed = fixed.replace(/\n```\s*$/, '');
```

## Fix

Guard the strip with the same predicate `lintContent()` uses for the `code-fence-wrap` rule, so it only fires when the page is genuinely wrapped:

```ts
const isMarkdownWrapped =
  content.match(/^```(?:markdown|md)\s*\n/m) && content.match(/\n```\s*$/m);
if (isMarkdownWrapped) {
  fixed = fixed.replace(/^```(?:markdown|md)\s*\n/, '');
  fixed = fixed.replace(/\n```\s*$/, '');
}
```

## Tests

Two regression tests added to `test/lint.test.ts`:
- non-wrapped page with an interior code block → returned byte-for-byte unchanged
- non-wrapped page with an LLM preamble (so `fixContent()` genuinely rewrites the body) → the interior closing fence survives

All 20 tests pass: `bun test test/lint.test.ts`.

### DIFF
--- a/src/commands/lint.ts
+++ b/src/commands/lint.ts
@@ -291,9 +291,19 @@ export function fixContent(content: string): string {
     fixed = fixed.replace(pattern, '');
   }
 
-  // Fix wrapping code fences
-  fixed = fixed.replace(/^```(?:markdown|md)\s*\n/, '');
-  fixed = fixed.replace(/\n```\s*$/, '');
+  // Fix wrapping code fences — guard the strip so it ONLY runs when the page
+  // is genuinely ```markdown/```md-wrapped (opening fence AND matching closing
+  // fence). Previously these replaces ran unconditionally on every fixContent()
+  // call, so any unrelated fixable issue (e.g. frontmatter-nested-quotes) would
+  // trigger a rewrite and silently strip a legit trailing ``` that closed an
+  // interior code block — corrupting code-heavy pages such as technical docs.
+  // The guard mirrors lintContent()'s code-fence-wrap predicate exactly.
+  const isMarkdownWrapped =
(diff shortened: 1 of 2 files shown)


time-attack:
Thanks @hanksuper! 🙏

What this does: This PR fixes a bug in fixContent() where fence-stripping was applied unconditionally on every call, corrupting interior code blocks. The fix adds a guard so the strip only runs when the document is actually wrapped in opening and closing…

This overlaps with #2741, which we're carrying forward, so I'm closing this in its favor to keep the review queue focused. If yours covers a case #2741 misses, reopen with that detail and we'll reconcile them.

PR #2655, opened 2026-07-07
feat(schema-pack): add slug_filter to retype mapping rules

YiconZiwei:
## Summary

`retype` mapping rules support an optional `path_filter` for disambiguating same-`from_type` rules by path. That filter matches against `pages.source_path`, which is only populated for pages synced from a git repo (`sync --repo`). Pages ingested via the `put_page` MCP tool — or any write path that doesn't go through `sync` — have `source_path = NULL`, so `path_filter` has no way to address this class of page at all, even though the intent (retype-by-slug-prefix) is otherwise perfectly expressible.

This PR adds `slug_filter`: an independent, optional `LIKE` filter on `pages.slug` (a column that's always populated, regardless of ingestion path). It composes with `path_filter` via `AND` when both are given, and is a no-op change for every existing pack that doesn't set it.

## Motivating case

We use `unify-types` to reclassify ~800 knowledge-base pages, all written through `put_page` (no `source_path`). A dry-run against 7 `retype` rules using `path_filter` returned `would_apply: 0` for every rule — not because there was nothing to retype (there were 109 matching pages by slug), but because the filter column itself was always `NULL` for this ingestion path. `slug_filter` closes that gap without touching `path_filter`'s existing semantics.

## Changes

- `manifest-v1.ts`: add optional `slug_filter` to `RetypeMappingRuleSchema`
- `retype.ts`: add `slug_filter` to the `RetypeRule` interface; `probeRule` and `applyRetypeRule` both apply `AND slug LIKE $N` when present, independent of `path_filter`
- `unify-types-handler.ts`: carry `rule.slug_filter` through the pack-mapping-rule → `RetypeRule` conversion (a field present only in the zod schema but dropped here would validate fine yet silently no-op at execution — this is the one non-obvious wiring point)
- `test/schema-pack-retype.test.ts`: three new tests — skips pages outside `slug_filter`, matches despite `NULL` `source_path` (the motivating case), and `path_filter` + `slug_filter` combined via `AND`

## Test plan

- [x] `bun test test/schema-pack*.test.ts test/operations-schema-pack.test.ts` — 320/320 pass (23 files, includes the dedicated `unify-types-handler` suite), 0 failures
- [x] `bun run typecheck` (`tsc --noEmit`) — clean
- [x] New tests specifically cover the `source_path IS NULL` case that motivated this change


time-attack:
This is a clean, well-tested diff and the gap you found is real — `path_filter` genuinely can't address pages ingested through `put_page`, since `source_path` is NULL for them. That said, we're not taking new feature surface on the schema-pack manifest right now; every optional field added to `RetypeMappingRuleSchema` is a contract we carry permanently for every pack author. For the ~800-page reclassification you describe, a scoped retype rule set or a one-off update against the affected slugs will get you there without the manifest change. Thank you for the detailed motivating case.

time-attack:
Reopening — closed in error. This PR was built/approved as part of the maintainer's explicit triage decisions (the doctor silent-failure batch, skills-integrity manifest, npm-shadow guard, DCR token TTL, and the retype slug_filter). The bulk strict-policy pass that swept it did not have that context. Merge queue for these resumes once CI's install step recovers.

time-attack:

Real gap, correct minimal fix: put_page-ingested pages have source_path=NULL so path_filter can never match them; slug_filter (LIKE on the always-populated slug column) is correctly added in all 3 src sites (zod schema, probe+apply SQL, handler pass-through) and the apply path is genuinely pinned (1 new test fails on full revert). But 2 of the 3 wiring points are provably untested: stubbing probeRule's slug_filter branch (dry-run would_apply — the motivating case's exact surface) leaves all 15 tests green, and deleting the unify-types-handler pass-through at line 155 — the line the PR body itself calls 'the one non-obvious wiring point [that] would silently no-op at execution' — leaves all 21 tests green. Two of three new tests pass with the entire feature reverted. Plus title lacks the version-first prefix and slug_filter is undocumented in docs/architecture/pack-upgrade-mechanism.md. Asks are cheap and concrete: (1) assert would_apply in a slug_filter dry-run, (2) one runUnifyTypes test with a pack retype rule carrying slug_filter, (3) retitle vX.Y.Z.W-first, (4) one doc line.

### Discrimination test

Reverted all three src files to origin/master while keeping the PR's test file: 1 of 3 new tests failed ('skips pages outside the slug_filter', total_applied 2 vs expected 1) — so the apply path is pinned and the PR passes the whole-diff discrimination test. Partial stubs then showed the other two wiring points are unpinned: (a) `if (false && slugFilter)` in probeRule → 15/15 green; (b) deleting the unify-types-handler.ts:155 pass-through → 21/21 green across both suites.

**Mergeability at review time:** {"mergeStateStatus":"CLEAN","mergeable":"MERGEABLE"}

### Blockers

**1.** Unpinned dry-run path: src/core/schema-pack/retype.ts:139-142 — changed probeRule's branch to `if (false && slugFilter)` and all 15 tests in test/schema-pack-retype.test.ts still pass; no test asserts would_apply for a slug_filter rule, so a probe/apply divergence (dry-run reports unfiltered count, apply updates filtered set) would ship undetected

**2.** Unpinned handler wiring: src/core/schema-pack/unify-types-handler.ts:155 — deleted the `slug_filter: rule.slug_filter` pass-through and all 21 tests (retype + unify-types-handler suites) still pass; a pack declaring slug_filter would validate then silently no-op through runUnifyTypes, the exact silent-failure class this PR exists to fix, per the PR body's own admission

**3.** Non-discriminating tests: 'matches slug_filter even when source_path is NULL' and 'combines path_filter AND slug_filter' both pass with the entire src diff reverted (verified); only 'skips pages outside the slug_filter' fails

**4.** PR title 'feat(schema-pack): add slug_filter to retype mapping rules' violates the version-first IRON RULE (must lead with vMAJOR.MINOR.PATCH.MICRO)

**5.** slug_filter undocumented: docs/architecture/pack-upgrade-mechanism.md:193 shows the retype mapping-rule example with path_filter only

### Not verified

- Full unit suite / bun run verify / ci:local not run (footprint rule); only typecheck + the 25 schema-pack-scoped test files
- Real Postgres not exercised (PGLite only) — low risk: diff adds parameterized `slug LIKE $N` predicates identical in shape to the existing path_filter ones, no jsonb/dialect surface
- End-to-end MCP put_page → unify-types flow with a real pack manifest not run; handler wiring verified by code read + typecheck only (and shown untested — that is blocker 2)

- `gh pr view 2655 --json mergeable,mergeStateStatus,... (twice; UNKNOWN then MERGEABLE/CLEAN)`
- `gh pr diff 2655 > /tmp/pr2655.diff (read every hunk)`
- `git fetch origin pull/2655/head:pr2655 && git checkout pr2655`
- `bun install`
- `bun test test/schema-pack-retype.test.ts (15/15)`
- `bun test test/schema-pack-unify-types-handler.test.ts (6/6)`
- `git checkout origin/master -- src/core/schema-pack/{retype,unify-types-handler,manifest-v1}.ts; bun test test/schema-pack-retype.test.ts (14 pass / 1 fail — discrimination)`
- `stub probeRule slugFilter branch to false; bun test test/schema-pack-retype.test.ts (15/15 green — unpinned)`
- `delete handler slug_filter pass-through; bun test both suites (21/21 green — unpinned)`
- `restore PR state; bun test both suites (21/21)`
- `bun test test/schema-pack test/operations-schema-pack.test.ts (368/368 across 25 files, EXIT=0)`
- `grep -rn path_filter src/ docs/ skills/ test/ (sibling-caller + docs coverage)`

</details>

---

PR #2659, opened 2026-07-07
Fix propose_takes proposal identity

sene1337:
## Summary
- split `propose_takes` page-scan idempotency into a new `take_proposal_scans` marker table
- add `take_proposals.claim_hash` per-claim identity so same-page multi-claim extraction persists every claim
- add v123 migration/backfill plus PGLite/Postgres forward-reference bootstrap coverage
- add PGLite and Postgres regression coverage for multi-claim inserts, zero-claim scan caching, and claim-hash backfill parity

## Verification
- `bun test test/propose-takes.test.ts test/propose-takes-pglite.test.ts test/schema-bootstrap-coverage.test.ts test/e2e/propose-takes-postgres.test.ts` ✅ 41 pass / 2 skipped (Postgres E2E skipped because no DATABASE_URL)
- `bun run build` ✅
- `git diff --check` ✅
- `bun run typecheck` ❌ existing unrelated OpenClaw SDK typing drift:
  - `src/core/context-engine.ts(90,44): Property 'buildMemorySystemPromptAddition' does not exist ...`
  - `test/e2e/openclaw-plugin-load-real.test.ts(281,9): Type 'unknown' is not assignable ...`
- `bun run test` full suite was started; first foreground attempt hit the 10-minute tool timeout while still running with no failure surfaced. A background rerun is in progress locally.

### DIFF
--- a/src/core/cycle/propose-takes.ts
+++ b/src/core/cycle/propose-takes.ts
@@ -6,10 +6,11 @@
  * `take_proposals` queue. User accepts/rejects via `gbrain takes propose`.
  *
  * Idempotency contract (D17 schema spec):
- *   The unique index on (source_id, page_slug, content_hash, prompt_version)
- *   means an unchanged page never re-spends LLM tokens. Bumping
- *   PROPOSE_TAKES_PROMPT_VERSION cleanly invalidates the cache so a tuned
- *   prompt re-runs proposals on every page.
+ *   `take_proposal_scans` is the page-scan cache keyed by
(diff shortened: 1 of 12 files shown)


time-attack:
Thank you for the careful writeup and the PGLite/Postgres parity coverage — the diagnosis was correct. Both halves have since landed on master, though: per-claim identity in ca4cf2a0 (#3297), which added `md5(claim_text)` to the `take_proposals` idempotency index, and zero-claim scan memoization in 3fcca330 (#2514/#3319). Rebasing would leave a competing migration against schema that's already settled, so closing this out.

PR #2668, opened 2026-07-08
v0.42.59.0 test(build): guard schema-embedded.ts freshness against schema.sql

BenjaminDSmithy:
## Why this matters

`src/core/schema-embedded.ts` is the database schema baked into every compiled
`gbrain` binary — generated from `src/schema.sql` (the single source of truth)
by `bun run build:schema`.

If a `schema.sql` edit lands without re-running the generator, the committed
blob goes stale and the binary ships an **old schema**. Every PGLite install
(the zero-config default) then bootstraps a schema that no longer matches the
source — columns/indexes/constraints the code expects can be missing, a silent,
hard-to-trace gap at the layer that has to be bulletproof.

The `llms-full.txt` docs bundle already has a regen-drift guard
(`test/build-llms.test.ts`). The far-more-load-bearing schema blob had none.
This closes that gap the same way.

*(The blob is in sync on `master` today — this is prevention, not a fix for
existing drift.)*

## What's added

- **Freshness guard, two lanes.** `test/build-schema-fresh.test.ts` (unit
  suite) and `scripts/check-schema-embedded-fresh.sh` (verify pool, wired into
  `run-verify-parallel.sh` + `package.json` as `check:schema-embedded`) each
  regenerate the blob via the **real** `scripts/build-schema.sh` into a temp
  file and assert the committed `src/core/schema-embedded.ts` is byte-identical.
  On drift, CI fails with "run `bun run build:schema`" — on both the fast
  `verify` pre-push gate and the unit suite.
- **`scripts/build-schema.sh` gains `SCHEMA_FILE` / `OUT_FILE` env overrides + a
  cd to the repo root**, so the guards drive THE generator into a temp file with
  zero duplication of the sed-escaping logic. The default `bun run build:schema`
  output is byte-for-byte unchanged.

## One bundled test fix

Adding a new test file re-balances CI's shard LPT bin-packing, which newly
co-locates `test/engine-find-trajectory.test.ts` with a 1280-d embedding
polluter in one shard. That polluter leaves the global gateway at a non-1536
dimension; find-trajectory's `beforeAll` runs `initSchema()` before the shared
legacy-embedding preload's `beforeEach` can restore 1536, so it built a
`vector(1280)` column and its 1536-d fixtures failed ("expected 1280
dimensions, not 1536").

Fix: pin the 1536-d default in find-trajectory's own `beforeAll` before
`initSchema()` — order-independent, so it holds regardless of which file the
packer places before it.

## Testing — every angle challenged

Guard behavior, verified on this branch against `master`:

| Scenario | Expected | Result |
|---|---|---|
| Clean tree | both guards pass | ✓ exit 0, "is fresh" |
| `schema.sql` edited, blob not regenerated | both guards FAIL, "run `bun run build:schema`" | ✓ check exit 1; test 2 fail |
| Generator fails (missing `schema.sql`) | labeled error, NOT a false "fresh" | ✓ exit 1, "schema generation failed" |
| Hostile `SCHEMA_FILE` env var | guard pins the canonical source, ignores it | ✓ exit 0 — env can't redirect the compare |
| Run from a subdirectory | cwd-independent | ✓ exit 0 |
| Temp-file cleanup | no leak | ✓ before == after |
| Two regenerations | byte-identical (deterministic) | ✓ |
| Backtick + `$` in `schema.sql` | escaped for template-literal safety | ✓ (dedicated `SCHEMA_FILE`-fixture test) |
| Verify wiring | check in dispatcher, `verify` delegates to it, CI runs `verify`, shell guard's own exit-0 path | ✓ (all asserted in the unit test) |

`test/build-schema-fresh.test.ts` → **10 pass**.

find-trajectory fix:

| Scenario | Result |
|---|---|
| Co-resident with a 1280-d polluter | ✓ 24 pass (was 17 pass / 7 fail) |
| Run alone | ✓ 18 pass |
| Order-independence | pin is unconditional in `beforeAll` |

A cross-model adversarial pass during development could not produce a false-pass
or false-fail; the one issue it surfaced (a `SCHEMA_FILE` env-pollution
false-pass) is closed by pinning the canonical source, as verified above.

## Notes

- Clean 9-file diff; **no `src/core/schema-embedded.ts` change** (already fresh
  on `master` — the guard is prevention).
- Pre-push gate per `CONTRIBUTING.md`: `bun run verify` (includes the new
  `check:schema-embedded`).
- Checked for duplicates before filing: no open PR/issue adds a freshness guard
  for the embedded schema blob.


time-attack:
Appreciate the care here, and the reasoning about a stale embedded blob shipping in every compiled binary is sound in principle. I regenerated the blob from `master`'s `schema.sql` through the real `scripts/build-schema.sh` and diffed it — byte-identical, which matches your own note that this is prevention rather than a fix. We're holding the line at test and CI work that repairs something actually broken, so this is a close; the bundled shard-balance fix and the now-stale version bump also make it larger than the problem warrants.

PR #2669, opened 2026-07-08
v0.42.58.0 fix(cycle): make dream lint audit-only (#2661)

### LINKED ISSUE #2661: dream lint phase mutates source tree with no no-fix/dry-run mode
## Bug / safety issue

`gbrain dream --phase lint` appears to run the lint core in fix mode and can write back to the source tree. For agent-managed workspaces this is a surprising side effect: most dream phases are DB-only/read-only, but `lint` can mutate markdown files.

## Expected behavior

One of:

- `dream --phase lint` defaults to read-only and reports proposed fixes, or
- expose an explicit `--fix` / `--no-fix` flag, with non-mutating behavior available for scheduled maintenance, or
- document prominently that `lint` is filesystem-mutating and should be excluded from read-only dream cycles.

## Why this matters

Users may schedule dream/maintenance cycles expecting brain/index updates, not source rewrites. Agents also need a safe mode they can run without risking unreviewed edits in a synced/git-managed workspace.

## Current workaround

Exclude `lint` from scheduled dream loops and run other phases only.

javieraldape:
## Summary

- Makes the cycle/dream lint phase audit-only by calling `runLintCore` with `fix:false`, so `gbrain dream --phase lint` reports fixable findings without rewriting tracked brain files.
- Updates cycle status accounting and docs for the new audit-only behavior; users who want edits still run `gbrain lint --fix` explicitly.
- Adds regression coverage at both the phase helper and `runDream --phase lint` surfaces to prove a fixable markdown page is left byte-for-byte unchanged.
- Normalizes duplicate separators in the operations allowlist guard so local verify compares the same relative paths across grep implementations.
- Adds a test-only type compatibility declaration for the dynamically imported plugin registration API, keeping repo typecheck stable without changing runtime imports.

Fixes #2661.

## Tests

- `bun run typecheck`
  - pass
- `bash scripts/gbrain-gate.sh <worktree> test/cycle-legacy-phases.test.ts test/dream.test.ts test/core/cycle.serial.test.ts`
  - verify: green
  - unit: 66 pass, 0 fail, 205 expect calls across 3 files


time-attack:
Merging this one. Unattended cycles rewriting a git-tracked brain with no way to opt out is a genuine safety bug, and it's still live on master (`src/core/cycle.ts` line 803 still passes `fix: true`) — the audit-only default with `gbrain lint --fix` as the explicit opt-in is the right shape. The regression tests that assert a fixable page stays byte-for-byte unchanged are what made this reviewable. One heads-up: the title carries a stale `v0.42.58.0` and the cycle status will now read `partial` on any brain with standing lint findings, which is honest but noisier than before — we'll sort the version on landing.

PR #2670, opened 2026-07-08
fix(gateway): use providerOptions[recipe.id] for thinking disable + add retry loop

### LINKED ISSUE #2577: chat: no providerOptions passthrough for openai-compatible providers — hybrid-reasoning models (DeepSeek V4) silently break strict-output ops
## Summary

`ChatOpts` has no way to pass provider-specific request-body params (e.g. DeepSeek's `thinking: {"type": "disabled"}`), even though the underlying `@ai-sdk/openai-compatible` layer already forwards unknown `providerOptions[<providerName>]` keys into the request body verbatim. For hybrid-reasoning models (DeepSeek V4 family) this makes strict-output ops silently unusable: thinking mode is the API default, the reasoning channel starves `content`, and e.g. `extract-conversation-facts` returns 0 facts across every segment with no error surfaced.

## Repro / evidence

Setup: Postgres brain, `chat_model = deepseek:deepseek-v4-flash` (openai-compatible recipe, `provider_base_urls.deepseek` pointing at an aggregator that serves DeepSeek V4).

1. DeepSeek V4 (`deepseek-v4-flash` / `deepseek-v4-pro`) defaults to **thinking mode ON** (verified against both api.deepseek.com and a Tencent-MaaS aggregator). `POST /chat/completions` with `"thinking": {"type": "disabled"}` cleanly disables it on both.
2. In thinking mode, strict-JSON extraction prompts return their tokens in `reasoning_content`; `content` is empty/truncated → `extractFactsFromTurn` parses nothing → `0 facts inserted across N segments` for every page. Same prompt with thinking disabled: 9 good facts on our test window.
3. gbrain cannot send that field today:
   - `ChatOpts` (src/core/ai/gateway.ts ~L2394) only carries `model / system / messages / tools / maxTokens / abortSignal / cacheSystem`.
   - Inside `chat()` the `providerOptions` object passed to `generateText` is local and only ever populated with `providerOptions.anthropic` for `cacheSystem`.
   - No config key feeds per-provider chat params.
4. The plumbing below is already done: `@ai-sdk/openai-compatible`'s `getArgs` spreads unknown keys from `providerOptions[this.providerOptionsName]` into the request body (only schema-known keys are filtered). So a passthrough would reach the wire with no SDK work.

## Suggested fix (small)

Either:
- add `providerOptions?: Record<string, Record<string, unknown>>` to `ChatOpts` and merge it into the local `providerOptions` in `chat()`, or
- a config surface like `models.chat_provider_options.<recipeId> = {...}` merged at the same spot (lets `facts.extraction_model = deepseek:deepseek-v4-flash` run non-thinking without any caller changes).

Workaround we run today: a tiny local proxy in front of the provider that injects `thinking: {"type":"disabled"}` for `deepseek-v4*` chat calls. Works, but it's a daemon nobody should need.

## Two sibling blind spots we hit on the same non-Anthropic path (can split into separate issues if preferred)

1. **BudgetTracker TX2 prices only via `ANTHROPIC_PRICING`** (src/core/budget/budget-tracker.ts `lookupPricing`): with `maxCostUsd` set (CLI default, minion jobs, cycle phases), any non-Claude chat model → `no_pricing` hard-throw → absorbed by callers like extract.ts → silent 0-output runs. DeepSeek/GLM/Kimi users can't run budget-capped ops at all. Suggestion: price chat via recipe `cost_per_1m_*` touchpoints (like embeddings already do) before falling back to hard-fail.
2. **`createAnthropic({ apiKey })` ignores `ANTHROPIC_BASE_URL`** (src/core/ai/gateway.ts native-anthropic branches): users on Anthropic-compatible proxies (common where api.anthropic.com is unreachable) get opaque `Invalid JSON response` on every native-anthropic call. One-liner: pass `baseURL` from env/config when present.

Happy to provide more repro detail or test a branch. Thanks for gbrain — the rest of the stack has been remarkably solid to align with.

justemu:
## Summary

Fixes #2577 — the thinking disable feature for openai-compatible providers was using the wrong key in `providerOptions`, causing DeepSeek V4's thinking mode to silently remain ON.

## Root Cause

`@ai-sdk/openai-compatible` reads `providerOptions[recipe.id]` (e.g. `providerOptions['deepseek']`), NOT `providerOptions.openaiCompatible`. This is because `createOpenAICompatible({ name })` sets `this.providerOptionsName = name`, where `name` is the recipe id.

Using `providerOptions.openaiCompatible` as a literal key silently drops the `thinking: { type: 'disabled' }` parameter before it reaches the wire → DeepSeek V4 defaults to thinking mode ON → reasoning tokens consume `max_tokens` quota → `content` is empty → AI SDK `generateText()` hangs until 300s timeout.

## Changes

1. **Fix providerOptions key**: `providerOptions.openaiCompatible` → `providerOptions[recipe.id]` so thinking disable reaches the API.
2. **Retry loop** for openai-compatible providers: 3 retries with exponential backoff (2s → 4s → 8s) and 30s per-attempt timeout, so transient failures do not fail the entire operation.

## Testing

- Verified `chat(deepseek:deepseek-v4-flash)` completes in ~835ms (was timing out at 300s before).
- Verified `propose_takes` cycle phase processes 100 pages and inserts 47 proposals with zero warnings (was failing all 100 extracts before).
- See issue #2577 for detailed diagnosis and timeline.

### DIFF
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -2735,6 +2735,26 @@ export async function chat(opts: ChatOpts): Promise<ChatResult> {
     providerOptions.anthropic = { cacheControl: { type: 'ephemeral' } };
   }
 
+  // Disable thinking/reasoning for all openai-compatible providers.
+  // DeepSeek V4 defaults to thinking mode ON, producing reasoning tokens
+  // that consume max_tokens quota and leave content empty, causing the AI SDK
+  // generateText() to hang until timeout (300s default).
+  //
+  // KEY: @ai-sdk/openai-compatible reads providerOptions[recipe.id], NOT
+  // providerOptions.openaiCompatible. The createOpenAICompatible({ name })
+  // field maps to this.providerOptionsName internally, so the key must match
+  // the recipe id (e.g. 'deepseek', 'openrouter', 'ollama').
+  // Using the literal 'openaiCompatible' key silently drops the parameter
+  // before it reaches the wire.
+  //
+  // Fixes: https://github.com/garrytan/gbrain/issues/2577
+  if (recipe.implementation === 'openai-compatible') {
+    providerOptions[recipe.id] = {
+      ...((providerOptions as any)[recipe.id] ?? {}),
+      thinking: { type: 'disabled' as const },
+    };
+  }
+
   let _budgetRecorded = false;
   const _recordBudget = (modelLabel: string, inputTokens: number, outputTokens: number): void => {
     if (!tracker || _budgetRecorded) return;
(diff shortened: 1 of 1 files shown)


time-attack:
Closing: current master already contains this fix. Its core fix (thinking-disable must go through providerOptions[recipe.id], not providerOptions.openaiCompatible) landed on master as commit 06f58c2 (#2857, 'fixes #2577'), which keys applyConfiguredChatProviderOptions on recipe.id (gateway.ts:2839-2841). The remaining delta is an unrequested blind 3x-retry/backoff loop inside gateway chat for openai-compat providers — scope creep that risks duplicate paid calls on non-idempotent failures; provider hang/tool-loop issues were separately addressed in fix-wave #2820.

PR #2672, opened 2026-07-08
fix(search): rerank docs >512 tokens & allow 1024-dim embedders

rgindrat-gif:
Two self-contained robustness fixes: (1) reranker fail-open on docs >512 tokens → sliding-window + max-pool; (2) Ollama recipe hardcoded to 768 dims → expose dims_options [768,1024] for multilingual embedders like bge-m3.


time-attack:
Thanks for looking at both of these. The Ollama half is already resolved on master — the recipe carries a per-model `model_dims` map (bge-m3 at 1024) plus `trust_custom_dims`, so nothing is pinned to 768 anymore. The reranker half is a new chunking-and-pooling strategy rather than a fix; chunks average around 400 tokens so the over-512 path is an edge case, and a change that multiplies reranker calls per document would need eval numbers and test coverage before we'd take it. Closing.

PR #2690, opened 2026-07-09
fix(db): chunk-row model label falls back to live gateway model, not build-time constant

### LINKED ISSUE #1717: Embedding: `gbrain embed`/`import` mislabels content_chunks.model as the gateway default (v0.41.x)

paul-0320:
## Problem

Closes #1717.

`content_chunks.model` records `DEFAULT_EMBEDDING_MODEL` (currently `'zeroentropyai:zembed-1'`) for every chunk of a brain configured with any other embedding provider. No `ChunkInput` builder sets `model` — `embed.ts` (3 call sites), `embed-stale.ts`, and `import-file.ts` all leave it undefined — so the `chunk.model || DEFAULT_EMBEDDING_MODEL` fallback in both engines' `_upsertChunksOnce` always wins, stamping provenance for a model that never produced the vectors.

Meanwhile the two *page-level* mechanisms resolved from the gateway are correct (`pages.embedding_signature` and the column DEFAULT baked by `initSchema`), so the row label silently contradicts them. Nothing reads the column for decisions today (stale detection is `embedding IS NULL`; model-swap detection is `pages.embedding_signature`), but any operator inspecting the DB — or any future feature that trusts the column — gets the wrong provider.

This is the recurrence of the v0.37 Lane A.7 fix: that pass replaced the hard-coded `'text-embedding-3-large'` literal with the `DEFAULT_EMBEDDING_MODEL` constant, but a build-time constant still isn't *this brain's* configured model.

## Relationship to #1803

#1803 fixes the same mislabel (#1717) at the `embed` command layer, so it covers `gbrain embed` only. This PR closes the class at the seam every path converges on — `upsertChunks` in both engines — so `put`, `import`, worker `embed-backfill`, and contextual re-index get the correct label too, including any future ChunkInput builder that forgets to set `model`.

## Change (2 files, same seam in both engines)

`_upsertChunksOnce` resolves the fallback from the live gateway (`getEmbeddingModel()`) at upsert time — the same pattern `initSchema` uses to bake the column DEFAULT. The constant remains only when the gateway is unconfigured (unit tests).

## Tests

`test/v0_37_gap_fill.serial.test.ts` Lane A.7 updated to pin both branches: configured gateway model wins; unconfigured gateway falls back to `DEFAULT_EMBEDDING_MODEL`. 13 tests / 0 fail; `tsc --noEmit` clean on top of current master.

### DIFF
--- a/src/core/pglite-engine.ts
+++ b/src/core/pglite-engine.ts
@@ -2104,6 +2104,15 @@ export class PGLiteEngine implements BrainEngine {
     const params: unknown[] = [];
     let paramIdx = 1;
 
+    // See postgres-engine.ts _upsertChunksOnce: model-less chunks fall back
+    // to the LIVE gateway model, not the build-time DEFAULT_EMBEDDING_MODEL
+    // constant, which mislabels provenance on non-ZE brains.
+    let fallbackModel: string = DEFAULT_EMBEDDING_MODEL;
+    try {
+      const gw = await import('./ai/gateway.ts');
+      fallbackModel = gw.getEmbeddingModel() || fallbackModel;
+    } catch { /* gateway unconfigured (unit tests) — keep the constant */ }
+
     for (const chunk of chunks) {
       const embeddingStr = chunk.embedding
         ? '[' + Array.from(chunk.embedding).join(',') + ']'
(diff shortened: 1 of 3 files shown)


time-attack:
Closing as duplicate of #1803.

Competing PR for chunk-row model label fallback. Keep #1803.

Issue #2673, opened 2026-07-08
v119 to v122 migration failure: column "event_page_id" does not exist


alexpunct:
## Summary

On an existing Postgres brain, `initSchema()` (and therefore `gbrain init --migrate-only`, `apply-migrations`, and any command that brings up the schema) crashes with:

```
column "event_page_id" does not exist
```

The schema stays stuck at v119 and migrations v120–v122 never apply. `gbrain sources status` also prints `Schema probe failed: column "event_page_id" does not exist` on every run.

## Root cause

`initSchema()` runs `applyForwardReferenceBootstrap(conn)` and then replays the embedded schema (`SCHEMA_SQL`) *before* running incremental migrations:

- `src/core/postgres-engine.ts:327` → `applyForwardReferenceBootstrap(conn)`
- `src/core/postgres-engine.ts:329` → `conn.unsafe(sqlText)` (SCHEMA_SQL replay)
- `src/core/postgres-engine.ts:332` → `runMigrations(this)`

The embedded schema now references `timeline_entries.event_page_id` (added by migration **v121** `timeline_entries_event_page_id`):

- `src/core/schema-embedded.ts:549` → `event_page_id INTEGER REFERENCES pages(id) ON DELETE CASCADE`
- `src/core/schema-embedded.ts:561-562` → `CREATE INDEX ... ON timeline_entries(event_page_id) ...`

On an existing brain `CREATE TABLE IF NOT EXISTS timeline_entries` is a no-op, so the column is **not** added, and the subsequent `CREATE INDEX ... (event_page_id)` in the SCHEMA_SQL replay throws. Because this happens before `runMigrations`, v121 (which would add the column) never gets a chance to run — the transaction rolls back and the version stays at 119.

`applyForwardReferenceBootstrap` exists precisely to add columns the embedded schema forward-references but old brains lack. Every other such reference has a bootstrap entry — e.g. v79 `pages.last_retrieved_at` at `src/core/postgres-engine.ts:802` (`ALTER TABLE pages ADD COLUMN IF NOT EXISTS last_retrieved_at ...`). The `event_page_id` case is missing its equivalent.

## Fix

Add an `event_page_id` bootstrap entry to `applyForwardReferenceBootstrap` (both the Postgres and PGLite mirrors), matching the v79 pattern:

```sql
ALTER TABLE timeline_entries ADD COLUMN IF NOT EXISTS event_page_id INTEGER;
```

[...]

PR #2691, opened 2026-07-09
fix(db): bootstrap timeline_entries.event_page_id before schema-blob index replay

### LINKED ISSUE #2626: Upgrade deadlock at schema v121: schema.sql eager-indexes timeline_entries.event_page_id before migration 121 adds the column
## Summary

On an **existing** Postgres brain upgrading across schema version **121** (`timeline_entries_event_page_id`, the Life Chronicle feature #2390), `initSchema()` deadlocks and every command that connects fails.

`schema.sql` unconditionally runs, on **every** `initSchema`:

```sql
CREATE INDEX IF NOT EXISTS idx_timeline_event_page  ON timeline_entries(event_page_id) WHERE event_page_id IS NOT NULL;
CREATE UNIQUE INDEX IF NOT EXISTS idx_timeline_event_dedup ON timeline_entries(event_page_id, date) WHERE event_page_id IS NOT NULL;
```

But on a pre-121 DB the `event_page_id` **column** doesn't exist yet — it's added by migration 121. Because `initSchema` runs `schema.sql` (the baseline) **before** `runMigrations`, the index DDL throws `column "event_page_id" does not exist`, the connection aborts, and migrations 116–122 never get to run. The column that would fix it is gated behind the very step that keeps crashing. `CREATE INDEX IF NOT EXISTS` still evaluates the column reference, so `IF NOT EXISTS` doesn't save it.

## Environment

- gbrain **v0.42.56.0** (master `64617904`)
- engine: `postgres` (remote)
- DB was at schema version **115**, target **122** (i.e. an upgrade that crosses the 121 boundary)

## Symptoms

- `gbrain init --migrate-only` → `column "event_page_id" does not exist`
- `gbrain health` / `gbrain embed --stale` / `gbrain dream` → `Schema probe failed: column "event_page_id" does not exist` (embed-stale even prints `Re-run: gbrain apply-migrations --yes`, which does not help — see below)
- `gbrain dream` **synthesize** and **propose_takes** phases fail with `write CONNECTION_CLOSED …:5432` (the pooled connection drops when `initSchema` aborts mid-phase). This silently half-breaks the nightly `dream` job — orphan/synthesis work stops while embed/sync still appear to run.

## Root cause (file references on master `64617904`)

- `src/schema.sql` **L557–558** — the two `event_page_id` indexes, created on every `initSchema`.
- `src/core/pglite-schema.ts` **L387–388** — same, for the pglite path.
- `src/schema.sql` **L545** — `event_page_id` is declared inside `CREATE TABLE timeline_entries (…)`, which is a **no-op on an existing table** (`CREATE TABLE IF NOT EXISTS`), so the column is *not* backfilled onto existing DBs by the baseline.
- `src/core/migrate.ts` **L5443+** (migration `version: 121`, `timeline_entries_event_page_id`) — the only place the column is actually added to existing DBs (`ALTER TABLE … ADD COLUMN IF NOT EXISTS event_page_id INTEGER;` at L5457, FK at L5467, indexes at L5473–5475).

Ordering: `initSchema` → `schema.sql` (eager index, **crashes here**) → `runMigrations` (never reached).

## Why fresh installs are unaffected (and CI stays green)

A fresh install's `CREATE TABLE timeline_entries (… event_page_id …)` creates the column up front, so the subsequent eager index DDL succeeds. Only DBs that already have a `timeline_entries` table and cross the 121 boundary hit the deadlock — a path a fresh-install test matrix won't exercise.

## `apply-migrations` is a red herring

The embed-stale hint suggests `gbrain apply-migrations --yes`, but that path runs named backfills (`v0.13.1`, `v0.29.1`, …) and does **not** fix the connect-time `schema.sql` index; it also surfaces a misleading `Applying migration v0.29.1 … status=failed` which is really the same missing column hit via `initSchema`.

## Workaround (what unblocked us — non-destructive)

Run once against the DB, then let the normal migrator finish:

```sql
ALTER TABLE timeline_entries ADD COLUMN IF NOT EXISTS event_page_id INTEGER;
```
```
gbrain init --migrate-only
```

With the column present, `schema.sql`'s index DDL becomes a no-op, `initSchema` stops aborting, and migrations 116–122 apply normally — migration 121 idempotently adds the FK + indexes and stamps the ledger. Schema reached 122; `health`/`dream` are clean.

## Suggested fix (approach is yours to choose)

The `event_page_id` index creation shouldn't live in the always-on `schema.sql` baseline **ahead of** the column that a later migration adds. Options:

1. **Move** the two `event_page_id` index statements out of `schema.sql` / `pglite-schema.ts` and rely solely on migration 121 (it already creates them `IF NOT EXISTS`) — assuming fresh installs run migrations after the baseline.
2. **Guard** them in `schema.sql` with a `DO $$ … IF column_exists … END $$;` block.
3. **Reorder** `initSchema` so `runMigrations` executes before the eager index DDL.

This class of bug will recur on any future migration that adds a column *and* eagerly indexes it in `schema.sql`, so a guard/convention (or a CI job that runs `init` against a snapshot of an older schema, not just a fresh install) would prevent the next one.

Happy to open a PR for whichever approach you prefer.

paul-0320:
## Problem

Fixes #2626, #2673.

Upgrading any pre-v121 brain wedges at `initSchema`: the schema blob's `CREATE TABLE IF NOT EXISTS timeline_entries` is a no-op on an existing table (it won't add the v121 column), so the blob's partial indexes `idx_timeline_event_page` + `idx_timeline_event_dedup` crash with `column "event_page_id" does not exist` **before** `runMigrations` can ever apply v121. The version stays stamped at the old number and every retry hits the same wall. Reproduced on a real v119 postgres brain upgrading to v0.42.57.0 (`gbrain init --migrate-only` dies with that one-line error).

Same wedge family as #239/#266/#357/#366/#374/#375/#378/#396: a column-with-index added to the schema blob without a corresponding `applyForwardReferenceBootstrap` entry.

## Relationship to #2548 / #2623

Both open PRs address the same wedge — offered here as an alternative in case either stalls; happy to close this if one of them lands. What this one adds:

- Bootstrap entry follows the exact house pattern in **both** engines (single-round-trip probe extension + guarded `ADD COLUMN IF NOT EXISTS` for the bare column only — FK and indexes land via the blob and v121, both idempotent).
- A red/green regression in `test/bootstrap.test.ts`: down-mutate a fresh brain to pre-v121 shape (drop indexes + FK + column, stamp version 119) and assert `initSchema` reaches `LATEST_VERSION` with the column restored. Verified the test fails on master without the fix and passes with it.
- Verified end-to-end against a real v119 → v122 postgres upgrade (migrations 120–122 apply, doctor reports latest).

## Tests

`bun test test/bootstrap.test.ts test/migration-v120.test.ts`: 9 tests / 0 fail; `tsc --noEmit` clean on top of current master.

### DIFF
--- a/src/core/pglite-engine.ts
+++ b/src/core/pglite-engine.ts
@@ -518,7 +518,11 @@ export class PGLiteEngine implements BrainEngine {
         EXISTS (SELECT 1 FROM information_schema.columns
                 WHERE table_schema='public' AND table_name='pages' AND column_name='embedding_signature') AS pages_embedding_signature_exists,
         EXISTS (SELECT 1 FROM information_schema.columns
-                WHERE table_schema='public' AND table_name='pages' AND column_name='links_extracted_at') AS pages_links_extracted_at_exists
+                WHERE table_schema='public' AND table_name='pages' AND column_name='links_extracted_at') AS pages_links_extracted_at_exists,
+        EXISTS (SELECT 1 FROM information_schema.tables
+                WHERE table_schema='public' AND table_name='timeline_entries') AS timeline_entries_exists,
+        EXISTS (SELECT 1 FROM information_schema.columns
+                WHERE table_schema='public' AND table_name='timeline_entries' AND column_name='event_page_id') AS timeline_event_page_id_exists
     `);
     const probe = rows[0] as {
       pages_exists: boolean;
@@ -561,6 +565,8 @@ export class PGLiteEngine implements BrainEngine {
       pages_generation_exists: boolean;
       pages_embedding_signature_exists: boolean;
       pages_links_extracted_at_exists: boolean;
+      timeline_entries_exists: boolean;
+      timeline_event_page_id_exists: boolean;
     };
 
     const needsPagesBootstrap = probe.pages_exists && !probe.source_id_exists;
@@ -637,6 +643,15 @@ export class PGLiteEngine implements BrainEngine {
     // it; pre-v112 brains crash without the column, so bootstrap adds it before
     // the CREATE INDEX runs. v112 runs later via runMigrations and is idempotent.
     const needsPagesLinksExtractedAt = probe.pages_exists && !probe.pages_links_extracted_at_exists;
+    // v0.42.x (v121, Life Chronicle #2390): idx_timeline_event_page +
+    // idx_timeline_event_dedup in PGLITE_SCHEMA_SQL reference event_page_id.
+    // `CREATE TABLE IF NOT EXISTS timeline_entries` is a no-op on pre-v121
+    // brains (won't add the column), so the blob's CREATE INDEX crashes with
+    // `column "event_page_id" does not exist` before v121 ever runs. Bootstrap
+    // adds the bare column; the FK + indexes land via v121 / the blob, both
+    // idempotent.
+    const needsTimelineEventPageId = probe.timeline_entries_exists
+      && !probe.timeline_event_page_id_exists;
 
     // Fresh installs (no tables yet) and modern brains both no-op.
     if (!needsPagesBootstrap && !needsLinksBootstrap && !needsChunksBootstrap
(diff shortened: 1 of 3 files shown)


time-attack:
Closing as superseded by #2735 (merged in v0.42.59.0), which ships the same event_page_id bootstrap in both engines with CI coverage. Thanks for the fix!

PR #2526, opened 2026-07-02
feat(embedding): Qwen3-Embedding Matryoshka dims on local openai-compat providers (ollama)

paul-0320:
## Problem

Local Matryoshka embedding is currently impossible. Every provider on the custom-dims allow-list (`voyage`, `zeroentropyai`, `openai` text-embedding-3, DashScope, Zhipu) is cloud-hosted, so:

1. `gbrain init --embedding-model ollama:qwen3-embedding:8b --embedding-dimensions 1536` is rejected at preflight with *"does not support custom dimensions"* (Tier 3 of `isCustomDimValidForProvider`), and
2. even past init, `dimsProviderOptions()` never emits the `dimensions` param for unlisted openai-compat models, so the server returns the model's native width (4096 for Qwen3-8B) and the first write explodes against the `vector(1536)` column.

For local-only / data-sovereignty setups that need to stay under pgvector's 2000-dim HNSW cap, there is no stock path — even though Qwen3-Embedding is MRL-capable and tops the MTEB multilingual leaderboard.

## Change (2 seams, mirrors the existing dashscope/zhipu pattern)

- **`src/core/ai/recipes/ollama.ts`** — list the `qwen3-embedding` models and declare `dims_options` so preflight accepts MRL steps (Tier 1).
- **`src/core/ai/dims.ts`** — emit `{ openaiCompatible: { dimensions } }` for `qwen3-embedding*` model ids. Qwen3 embeddings are symmetric, so no `input_type` is emitted.

## Verification

Ollama honors the OpenAI `dimensions` param for MRL models — measured on Ollama 0.31.1: `POST /v1/embeddings` and `POST /api/embed` with `qwen3-embedding:8b` + `dimensions: 1536` both return 1536-wide vectors.

End-to-end on a fresh PGLite brain (`init --embedding-model ollama:qwen3-embedding:8b-fp16 --embedding-dimensions 1536`): doctor reports `1536 dims, DB aligned`, 100% embed coverage, and semantic `query` retrieves a page sharing zero keywords with the query text.

## Tests

`test/ai/dims-qwen3-ollama.test.ts` pins both seams: dimensions emission (incl. quant-suffix tags), no `input_type` leakage, preflight acceptance at 1536, rejection outside `dims_options`, and the `nomic-embed-text` no-param regression. Existing dims/recipe/regression suites pass (`dims-openai`, `dims-zeroentropy`, `recipes-existing-regression`, `embedding-dim-check`: 80 tests, 0 fail).

[...]

PR #2692, opened 2026-07-09
feat(embedding): Qwen3-Embedding Matryoshka dims on local openai-compat providers (ollama)

paul-0320:
> Supersedes #2526 — same change, resubmitted on a clean single-commit branch rebased to current master (the old PR's head branch had accumulated unrelated commits).

## Problem

Local Matryoshka embedding is currently impossible. Every provider on the custom-dims allow-list (`voyage`, `zeroentropyai`, `openai` text-embedding-3, DashScope, Zhipu) is cloud-hosted, so:

1. `gbrain init --embedding-model ollama:qwen3-embedding:8b --embedding-dimensions 1536` is rejected at preflight with *"does not support custom dimensions"* (Tier 3 of `isCustomDimValidForProvider`), and
2. even past init, `dimsProviderOptions()` never emits the `dimensions` param for unlisted openai-compat models, so the server returns the model's native width (4096 for Qwen3-8B) and the first write explodes against the `vector(1536)` column.

For local-only / data-sovereignty setups that need to stay under pgvector's 2000-dim HNSW cap, there is no stock path — even though Qwen3-Embedding is MRL-capable and tops the MTEB multilingual leaderboard.

## Change (2 seams, mirrors the existing dashscope/zhipu pattern)

- **`src/core/ai/recipes/ollama.ts`** — list the `qwen3-embedding` models and declare `dims_options` so preflight accepts MRL steps (Tier 1).
- **`src/core/ai/dims.ts`** — emit `{ openaiCompatible: { dimensions } }` for `qwen3-embedding*` model ids. Qwen3 embeddings are symmetric, so no `input_type` is emitted.

## Verification

Ollama honors the OpenAI `dimensions` param for MRL models — measured on Ollama 0.31.1: `POST /v1/embeddings` and `POST /api/embed` with `qwen3-embedding:8b` + `dimensions: 1536` both return 1536-wide vectors.

End-to-end on a fresh PGLite brain (`init --embedding-model ollama:qwen3-embedding:8b-fp16 --embedding-dimensions 1536`): doctor reports `1536 dims, DB aligned`, 100% embed coverage, and semantic `query` retrieves a page sharing zero keywords with the query text. Since then also running in production on a postgres brain (v0.42.57.0 base) with the same doctor result.

## Tests

`test/ai/dims-qwen3-ollama.test.ts` pins both seams: dimensions emission (incl. quant-suffix tags), no `input_type` leakage, preflight acceptance at 1536, rejection outside `dims_options`, and the `nomic-embed-text` no-param regression. 7 tests / 0 fail on top of current master; existing dims/recipe/regression suites pass.

### DIFF
--- a/src/core/ai/dims.ts
+++ b/src/core/ai/dims.ts
@@ -177,6 +177,12 @@ export function dimsProviderOptions(
       // Anthropic has no embedding model.
       return undefined;
     case 'openai-compatible':
+      // Qwen3-Embedding via Ollama/llama-server — Matryoshka (MRL), accepts
+      // any output dim 32..4096 via the OpenAI-compat `dimensions` param
+      // (verified against Ollama /v1/embeddings). Symmetric — no input_type.
+      if (modelId.startsWith('qwen3-embedding')) {
+        return { openaiCompatible: { dimensions: dims } };
+      }
       // ZE zembed-1 — flexible Matryoshka dims + asymmetric input_type.
       // Lives BEFORE the generic openai-compatible fall-through to avoid
       // sending input_type to providers (Azure/DashScope/Zhipu) that
--- a/src/core/ai/recipes/ollama.ts
+++ b/src/core/ai/recipes/ollama.ts
@@ -24,9 +24,15 @@ export const ollama: Recipe = {
         'all-minilm',
         'qwen3-embed-8b',
         'snowflake-arctic-embed-l-v2',
+        'qwen3-embedding:8b',
+        'qwen3-embedding:4b',
+        'qwen3-embedding:0.6b',
       ],
       default_dims: 768, // nomic-embed-text native dim
       trust_custom_dims: true, // #2271: local models carry varied native dims
+      // Ollama honors the OpenAI `dimensions` param (MRL truncation) for
+      // Matryoshka models like qwen3-embedding; validated dims listed here.
+      dims_options: [256, 512, 768, 1024, 1536, 2048, 2560, 4096],
       cost_per_1m_tokens_usd: 0,
       price_last_verified: '2026-04-20',
       // Ollama's batch capacity depends on the locally loaded model + the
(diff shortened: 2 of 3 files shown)


stigrunar:
Additional real-runtime evidence for this path:

- Ollama: `0.31.2` on macOS/Apple Silicon
- Model: `qwen3-embedding:0.6b`
- Native `/v1/embeddings` response: 1024 dimensions
- Same request with `dimensions: 768`: 768 dimensions
- Server-returned 768d norm: `0.99999964`
- Cosine against independently computed `native.slice(0, 768)` + L2 renormalization: `0.99999984`
- Maximum absolute component difference: `6.99e-5`

This supports sending the dimension through the provider seam rather than doing a client-side projection. A reproducible composite branch with the provider seam, query preprocessing, and focused tests is at [`stigrunar/gbrain@8e91de7a`](https://github.com/stigrunar/gbrain/commit/8e91de7a649a28a1dc6ce25b55e8d4202d2ea0bc).

I also replayed a frozen 80-query / 2,000-document-chunk retrieval evaluation using the actual server-returned 768d vectors. It exactly reproduced the previous client-projection result:

- Recall@1: `0.8875`
- Recall@5 / Recall@10: `1.0`
- MRR: `0.92645833`
- changed target ranks: `0 / 80`
- maximum absolute rank delta: `0`

One adjacent model-specific retrieval detail: the query text used Qwen3's recommended instruction:

```text
Instruct: Given a web search query, retrieve relevant passages that answer the query
Query: <query>
```

Documents were unprefixed. I agree that no provider-specific `input_type` field should be emitted; the instruction is a text-preprocessing concern at the query/document boundary and can remain a focused follow-up rather than expanding this dimensions PR.

time-attack:
Thank you for the real-hardware verification against Ollama 0.31.1 — that measurement was the useful part, and the `dimensions`-emission seam it justified is already on master (`src/core/ai/dims.ts` matches `qwen3-embedding` and any `:tag` suffix). The remaining delta is the recipe-wide `dims_options` on the ollama touchpoint, and that one we can't take: Tier 1 of `isCustomDimValidForProvider` consumes `dims_options` before the `trust_custom_dims` passthrough, so it would start rejecting ollama users on any width outside the list — 384 for `all-minilm`, or any locally-pulled model. Closing, since the working path already exists.

PR #2699, opened 2026-07-09
fix(cycle): drain PGLite synth subagents inline

TheRealMrSystem:
## Summary
- Route `dream --phase synthesize` child subagent jobs on PGLite to a private per-run queue.
- Drain that queue inline before the synth parent starts `waitForCompletion`, because an embedded PGLite data-dir lock means a separate worker cannot reliably claim those children.
- Keep Postgres behaviour unchanged: non-PGLite runs still submit to the default worker queue.

## Why
On PGLite, the synth parent can enqueue subagent children and then wait forever because no separate worker can acquire the same embedded DB/data-dir lock to process them. This makes a valid synth run look hung even though the queue state is the real blocker.

## Test Plan
- `bun test test/e2e/dream-synthesize-pglite.test.ts`
- `bun run typecheck`
- `bun run verify`
- `git diff --check`

## Notes
This is deliberately narrow: it only covers the PGLite synth parent/child orchestration invariant and does not include the broader local synth-safety patch stack.

### DIFF
--- a/src/core/cycle/synthesize.ts
+++ b/src/core/cycle/synthesize.ts
@@ -28,6 +28,7 @@
 
 import type Anthropic from '@anthropic-ai/sdk';
 import { readFileSync, existsSync, writeFileSync, mkdirSync } from 'node:fs';
+import { randomUUID } from 'node:crypto';
 import { chat as gatewayChat, validateModelId, type ChatResult } from '../ai/gateway.ts';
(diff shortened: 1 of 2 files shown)


time-attack:
Superseded with credit by #3162 (just merged) — your change landed rebased onto current master with authorship preserved via Co-authored-by. Thank you!

PR #2700, opened 2026-07-09
fix: scope autopilot fan-out to source path

gregclarkgsd:
## Summary
- Make each per-source `autopilot-cycle` job use that source\u2019s registered `local_path`, rather than the launch daemon\u2019s shared repo path.
- Add a regression assertion for source-ID-to-repo-path pairing.

## Why
A daemon launched from the Jarvis workspace could enqueue unrelated source IDs with the workspace path, causing repeated full imports and statement timeouts.

## Verification
- `bun test test/autopilot-fanout.test.ts` \u2014 28 pass, 0 fail
- `bun run typecheck` \u2014 pass
- All 31 `verify` checks \u2014 passed when run sequentially

## Deliberately not changed
- No source registration/migration
- No scheduler re-enable
- No retry or reconciliation policy changes

### DIFF
--- a/src/commands/autopilot-fanout.ts
+++ b/src/commands/autopilot-fanout.ts
@@ -434,7 +434,9 @@ export async function dispatchPerSource(
       const job = await queue.add(
         'autopilot-cycle',
         {
-          repoPath: opts.repoPath,
+          // A fan-out job is scoped to this source. Carrying the daemon's
+          // launch path here can make an unrelated source import that repo.
+          repoPath: src.local_path,
           source_id: src.id,
           pull: !!remoteUrl,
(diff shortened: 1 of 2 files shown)


time-attack:
Thanks for chasing this down — the mixed-scope problem you describe was real, but it was already fixed on master at the handler level in bb2e88c4 (v0.42.52.0, #2194/#2227): `autopilot-cycle` re-reads the source's `local_path` and binds `effectiveBrainDir` to it, never to the payload `repoPath`, so a per-source job cannot import the daemon's launch tree today. Closing as already fixed rather than carrying a redundant payload edit.

PR #2701, opened 2026-07-09
feat(health): env-gated orphan-metric type exclusion

danwiggins:
## What

Adds an opt-in env var, `GBRAIN_HEALTH_ORPHAN_EXCLUDE_TYPES`, that drops the listed page `type`s from `getHealth().orphan_pages`. Unset/empty = count every type (the current behavior), so this is a pure no-op until an operator opts in.

## Why

`orphan_pages` counts "islanded" pages (no inbound AND no outbound link). Machine-generated chrome — code-artifact pointers, ingest/extract/self-test/freshness logs — is islanded by nature: nothing references it and it references nothing, so it inflates the orphan count without signalling a real knowledge-graph gap. On a brain that ingests a lot of such chrome the metric can read ~46% orphaned while the human-knowledge graph is healthy.

## How

- New `src/core/health-config.ts`: `resolveOrphanExcludeTypes()` + `orphanTypeExcludeClause()`, with strict token validation so the fragment is injection-safe (operator-set env only, but fail-closed regardless).
- Postgres engine: `cardinality()=0 OR NOT (type = ANY())` guard idiom (the array-param pattern already used elsewhere in the file).
- PGLite engine: validated inline `NOT IN` clause (engine parity).
- 7 unit tests incl. malformed-token rejection.

## Risk / testing

Default OFF, reversible (unset the env). No schema change, no migration. `bun run typecheck` clean; `bun test test/health-config.test.ts` 7/7.

### DIFF
--- a/src/core/health-config.ts
+++ b/src/core/health-config.ts
@@ -1,0 +1,55 @@
+/**
+ * Health-metric configuration knobs (env-gated, default OFF).
+ *
+ * `getHealth().orphan_pages` counts "islanded" pages — no inbound AND no
+ * outbound link. Machine-generated chrome (code-artifact pointers, ingest /
+ * extract / self-test / freshness logs) is islanded by nature: it references no
+ * entity and nothing references it, so it inflates the orphan count without
+ * signalling a real knowledge-graph gap. On a brain that ingests a lot of such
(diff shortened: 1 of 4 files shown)


time-attack:
Reviewed and declined for now with the content-scope set. Thank you.

PR #2702, opened 2026-07-09
feat(conversation-parser): wire the LLM parse fallback into the extraction orchestrator

danwiggins:
## What

Wires `runLlmFallback` — the conversation-parser LLM fallback shipped in v0.41.16.0 — into the extraction orchestrator, so `no_match` pages actually get parsed when an operator opts in.

## Why (this is dead code today)

`runLlmFallback` is defined but **never called** — the only reference outside its own file is a comment in `parse.ts`. So pages whose format no built-in regex pattern matches (phase `no_match`) yield zero segments and therefore zero conversation facts, and the documented opt-in (`config set conversation_parser.llm_fallback_enabled true`) is a no-op: the key was never registered in the config allowlist **and** the value was never read anywhere.

## How

- `extract-conversation-facts.ts` `processPage`: on `no_match` + the opt-in gate, call `runLlmFallback` (utility/Haiku tier) on the body. The prompt returns `[]` for non-chat content, so non-conversation pages stay skipped. Fail-open; content-hash cached in `llm-base` so re-runs of an already-parsed page are free.
- Thread the page's date context into `runLlmFallback` and inject it into the prompt, so time-only timestamps resolve to the real conversation date instead of the `1970-01-01` epoch default. Without this, the fallback's messages fall below the per-page segment/checkpoint watermark, so a backfill pass never advances and re-parses the same page every run.
- Register the `conversation_parser.` config prefix so `config set` accepts the documented flag without `--force`.
- Gate + model resolved once into `ExtractCoreState`; new `pages_llm_fallback` result counter.

## Risk / testing

Opt-in, default OFF (privacy of chat logs). No schema change. `bun run typecheck` clean; conversation + config test suites pass (189/189). Verified end-to-end on a real brain: previously-unparseable pages now extract facts, and re-runs correctly skip already-completed pages (checkpoint advances).

### DIFF
--- a/src/commands/extract-conversation-facts.ts
+++ b/src/commands/extract-conversation-facts.ts
@@ -246,6 +246,9 @@ export interface ExtractConversationFactsResult {
   pages_skipped: number;
   pages_skipped_too_large: number;
   pages_skipped_disappeared: number;
+  /** pages whose built-in regex parse missed but the opt-in LLM fallback
+   *  recovered messages from. Absent/0 when the fallback is off. */
+  pages_llm_fallback?: number;
   /**
    * v0.41.15.0 (D6): pages we attempted to claim but skipped because
    * another worker / parallel process held the advisory lock. The pages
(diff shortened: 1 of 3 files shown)


danwiggins:
Superseded by #3292, which rebases the fallback wiring onto current master and adds full-body overlap-safe chunking, strict output validation, cancellation/budget handling, focused integration coverage, and an operator/maintainer runbook.

PR #2709, opened 2026-07-10
fix(sync): allow file transport only for the trusted local durability origin

hunterbohm:
## Problem

`pullRepo` spreads `GIT_SSRF_FLAGS` unconditionally, so `protocol.file.allow=never` also applies to the brain's own local bare durability origin (`~/.gbrain/brain-origin.git`). On any install whose default brain source pulls from that local bare repo, every autopilot cycle fails `sync.git_pull` with:

```
fatal: transport 'file' not allowed
```

Effects observed on a real install: the durability pull never runs, and `autopilot.err` grows unbounded (183k lines / 6 MB of the same repeating failure). Manual `git pull` and SSH `ls-remote` both succeed, so the repo/remote/auth are fine — only the daemon's hardened pull is broken.

A local filesystem remote is not an SSRF vector (no DNS, no network, no redirects), so rejecting it buys no security on this path while breaking the product's own durability design.

## Fix

Strict stays the default — `pullRepo` behavior is unchanged for every existing caller.

- New `GIT_SSRF_FLAGS_LOCAL`: identical to `GIT_SSRF_FLAGS` except `protocol.file.allow=always`. Redirect + external-helper hardening are kept, and submodules stay disabled via `GIT_SSRF_SUBCOMMAND_FLAGS`, so `always` cannot open a recursive file surface.
- New gated opt-in `pullRepo(repoPath, { allowLocalFileOrigin })`: even when passed, the relaxed flags are used only after re-verifying (realpath containment) that the branch's configured pull remote is a local path physically inside `GBRAIN_HOME` (or an explicit `trustedOriginRoot`). A network remote, an scp-like `host:path`, or a local path outside the trusted root all keep full hardening.
- `sync.ts` passes the opt-in only for the `default` brain source, so federated https sources are double-gated and keep full SSRF hardening.
- `isLocalFileRemote()` exported for the transport classification (file:// URL or bare filesystem path; any URL scheme or scp-like syntax is remote).

## Tests

`bun test test/git-remote.test.ts` → 41 pass / 0 fail (37 existing + 4 new):

- strict default still rejects a file-transport origin,
- opt-in pulls when the origin is contained under the trusted root,
- opt-in still rejects when the trusted root doesn't contain the origin,
- network remotes are never relaxed regardless of the opt-in.

The new tests were mutation-checked: forcing the flags always-strict kills only the relax test; forcing always-relaxed kills the strict tests.

## Repro

1. Point a brain's default source origin at a local bare repo (`git remote set-url origin ~/.gbrain/brain-origin.git`).
2. Run the autopilot (or any `sync` with a pull).
3. Before: `sync.git_pull error … transport 'file' not allowed` every cycle. After: `sync.git_pull done`.

### DIFF
--- a/src/commands/sync.ts
+++ b/src/commands/sync.ts
@@ -1627,7 +1627,14 @@ async function performSyncInner(engine: BrainEngine, opts: SyncOpts): Promise<Sy
       // pullRepo's own 300s default). The catch below distinguishes
       // timeout (ETIMEDOUT / SIGTERM on err.cause) from ordinary pull
       // failure.
-      pullRepo(repoPath);
+      // Allow the file transport ONLY for the trusted local durability origin:
+      // the default brain source (its origin is the gbrain bare repo under
+      // GBRAIN_HOME). pullRepo re-verifies the origin is local AND contained
+      // under the trusted root, so federated https sources keep full SSRF
+      // hardening. Fixes the autopilot git_pull loop ("transport 'file' not allowed").
+      pullRepo(repoPath, {
+        allowLocalFileOrigin: !opts.sourceId || opts.sourceId === DEFAULT_SOURCE_ID,
+      });
       serr(`[gbrain phase] sync.git_pull done ${Date.now() - _t0}ms`);
     } catch (e: unknown) {
       const msg = e instanceof Error ? e.message : String(e);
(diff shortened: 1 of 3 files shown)


time-attack:
Thank you for the detailed repro and the containment check — the engineering here is careful. But strict `protocol.file.allow=never` on clone/pull is an intentional line (documented in docs/architecture/KEY_FILES.md), and self-hosted filesystem remotes already have a documented escape hatch in `GBRAIN_GIT_ALLOW_FILE_TRANSPORT=1` on the durability paths. Adding a second, parallel relaxed flag set to the SSRF module isn't surface we're taking, so I'm closing this.

PR #2716, opened 2026-07-10
fix(skillify): scaffold test file lands inside skillsDir, not repoRoot

cogpros:
## The bug

`gbrain skillify scaffold --skills-dir ~/skills` writes the skill files to `~/skills/<name>/` but drops the generated test at `~/test/<name>.test.ts` — one level above the skills dir. Every scaffold since needs a manual `mv ~/test/<name>.test.ts ~/skills/test/`.

## Root cause

`planScaffold` in `src/core/skillify/generator.ts` anchored `testPath` to `repoRoot`, which defaults to `dirname(skillsDir)`. The CLI (`src/commands/skillify.ts`) never passes `repoRoot`, so the default fires on every run and the test lands outside the skills tree.

## The fix

Anchor `testPath` to `join(skillsDir, 'test', ...)` directly. The `repoRoot` option stays in the interface for API compatibility (the resolver fallback still uses it). Also updated to match the corrected location:

- generated test import path (`../<name>/scripts/` instead of `../skills/<name>/`)
- CLI help text + success message
- the two path references in `skills/skillify/SKILL.md`

## Verification

- New regression test (`test file lands inside skillsDir when repoRoot is omitted`) fails on pre-fix code with `Expected: true, Received: false` on the `startsWith(skillsDir)` assertion, passes post-fix.
- All 20 skillify tests pass (52 assertions), `tsc --noEmit` clean, build clean — verified against current master (a25209b).


time-attack:
Thanks for the careful root-cause write-up. The trouble is the default is correct for the primary case: in this repo `skillsDir` is `<repo>/skills`, so `dirname(skillsDir)` puts the stub at `<repo>/test/<name>.test.ts` — where every one of our ~977 tests lives and where the shard runner picks it up. Anchoring to `<skillsDir>/test/` would move scaffolded tests out of the suite and leaves `repoRoot` as a dead option, so I'm closing this one.

PR #2719, opened 2026-07-10
v0.42.58.0 fix(skillopt): emit proposed.md in no-mutate mode (#2635)

### LINKED ISSUE #2635: skillopt --no-mutate: best.md is written but proposed.md is never emitted
## Observed

Running skillopt with --no-mutate (v0.42.56.0-era): training completes,
the promote gate accepts an edit, best.md is written with the winning
skill content — but proposed.md is never created.

## Expected

In --no-mutate mode, proposed.md is the contract artifact: the produced
proposal for human review, with the live skill left untouched. Automation
watching for proposed.md (our human-gate pipeline does exactly this) sees
nothing and can't distinguish "run produced no proposal" from "run
produced a proposal that only landed in best.md."

## Repro

1. skillopt run on any skill with --no-mutate, benchmark where the
   optimizer finds an accepted edit
2. Run completes clean (exit 0)
3. best.md exists with the edited skill; proposed.md absent

## Suspected location

The gate/emit path in bundled-skill-gate.ts — the no-mutate branch
appears to write the best-candidate file but skip the proposal-emit step.

Happy to provide full run logs if useful. (Related: we just opened a PR
on the gateway max_tokens default from the same production debugging arc.)

RerankerGuo:
## Summary

- Emit a physical `skillopt/proposed.md` for accepted `--no-mutate` runs.
- Keep `best.md` as the optimizer's current-best pointer and leave `SKILL.md` untouched.
- Return the real proposal path from the orchestrator and update the SkillOpt tutorial.

Fixes #2635

## Root cause

The no-mutate branch called `writeProposed`, but that helper only wrote `best.md`. The orchestrator then returned the `best.md` path while describing it as `proposed.md`, so automation waiting for the documented proposal artifact could not distinguish an accepted proposal from no proposal.

## Validation

- `bun test test/skillopt/version-store.test.ts`
- `bun test test/e2e/skillopt-loop.serial.test.ts`
- `bun test test/skillopt` (208 pass)
- `bun test test/build-llms.test.ts`
- `bun run typecheck`
- `bun run verify` (31 checks pass)
- Bun `1.3.14`

`bun run ci:local:diff` was not available in this environment because the Docker daemon and `gitleaks` are unavailable.

### DIFF
--- a/docs/tutorials/improving-skills-with-skillopt.md
+++ b/docs/tutorials/improving-skills-with-skillopt.md
@@ -233,13 +233,14 @@ keep it or `git checkout` to throw it away. Nothing is committed for you.
 
 **For a skill that ships with gbrain** (anything under the gbrain repo's own
 `skills/`): SkillOpt refuses to overwrite it by default and writes the winner to
-`skills/<name>/skillopt/best.md` instead, so an optimization pass can never
-silently mutate a skill other people depend on. Two ways to handle that:
+`skills/<name>/skillopt/proposed.md` instead (while keeping `best.md` as the
+optimizer's current-best pointer), so an optimization pass can never silently
+mutate a skill other people depend on. Two ways to handle that:
 
 ```bash
 # See the proposed improvement without touching SKILL.md (works for ANY skill):
 gbrain skillopt meeting-prep --split 1:1:1 --no-mutate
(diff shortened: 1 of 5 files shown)


time-attack:
Superseded with credit by #3182 (just merged): your change landed rebased onto current master, with authorship preserved via Co-authored-by. Thank you for the contribution!

PR #2720, opened 2026-07-10
fix(gbrain): source-scoping + facts-fence + calibration + migration-honesty correctness fixes

quassandra:
Batch of correctness + reliability fixes from a personal fork, offered upstream. Each commit is independently reviewable; all are test-covered. typecheck: 0 errors.

## Fixes
1. **`get_page`/`resolve_slugs` per-call source scoping** — a scoped `query` could surface a page from a non-default source a follow-up `get_page` on the same connection couldn't resolve. Adds a `source_id` param mirroring `query`. Also stops `schema sync` calling a prefix "dead" when pages exist under it.
2. **`list_pages` honors per-call `source_id`** — same gap; it silently scoped to the caller's ambient default. Routed through the grant-checked `resolveRequestedScope`; omitted → prior behavior (the #861 leak seal holds).
3. **`extract_facts` guard ignores structurally-unfenceable legacy rows** — the guard blocked the cycle on any `row_num IS NULL` fact, but v0.32.2 can never fence a row whose source has NULL `local_path`, so the phase wedged forever. Excludes NULL-local_path rows, mirroring the NULL-entity_slug exclusion.
4. **v0.32.2 reports unfenceable rows instead of silently self-certifying** — Phase B returned `complete` having fenced 0 of a non-empty backlog. Keeps `complete` (permanently unfenceable; `partial` would re-run forever) but reports the backlog in the detail + a `console.warn`.
5. **calibration: honor holder env + reject leaked scaffolding** — the phase was always invoked with `{}` (generated for the default holder regardless of `GBRAIN_CALIBRATION_HOLDER`); and a thin scorecard leaked clarifying-question chain-of-thought into `pattern_statements`. Reads the env now, and filters scaffolding shapes.
6. **`takes_list` holder fallback + BigInt-safe calibration serialization** — `get_calibration_profile` threw "cannot serialize BigInt"; `takes_list` returned `[]` for a valid holder.
7. **concept synthesis attributes to dominant atom source, not `default`** — stranded concept pages in the unbacked `default` source.
8. **chat: retry the provider call on the default-timeout abort** — the wall-clock backstop aborts a slow provider call and the SDK doesn't retry an abort, so a transient slow moment killed the call (timed out 10/21 pages in a dream run). Bounded retry (≤3, 0.75s→1.5s), fresh deadline per attempt, fires ONLY on default-timeout aborts — caller aborts never retried.
9. **voice-gate judge uses the configured chat model, not a hardcoded bare id** — `defaultJudge` hardcoded `model: 'claude-haiku-4-5'` (no provider prefix → rejected by model resolution; also anthropic-only, unreachable on a proxy-only setup). Latent because the calibration cold-brain branch short-circuits before the gate runs for <5-take holders. Drops the hardcoded model so `chat()` resolves `opts.model ?? getChatModel()`.

Note #5, #8, #9 form a chain: fixing the holder (#5) made the phase reach the LLM path for a real holder, which surfaced the abort-not-retried behavior (#8) and the hardcoded judge model (#9) — both previously masked by the 0-take short-circuit.

## Testing
typecheck 0; extract-facts-phase 14/0, phantom-redirect 38/0, calibration-profile 21/0, calibration-cli 21/0, voice-gate 24/0, migrations-v0_32_2 14/0, operations-descriptions 27/0, source-scope-resolver 16/0, http-transport 28/0, gateway-chat-timeout-retry 6/0.

Note: branch carries a VERSION bump to 0.42.57.0 from #1; adjust on merge as needed.


quassandra:
Closing — landing these fixes on our own fork's master instead of pursuing an upstream merge right now. The branch stays available if any of these are useful upstream later.

PR #2721, opened 2026-07-10
v0.42.59.0 fix(cycle): keep propose_takes dry runs read-only

2emr:
## Summary

- make `propose_takes` dry-run fail closed before budget accounting, model extraction, proposal inserts, receipt pages, and rollup updates
- route the exact `gbrain dream --phase propose_takes --dry-run` CLI path through the existing probe-only connection so pending schema migrations are not auto-applied before the phase
- skip GBrain cycle lock acquisition only for a propose-takes-only dry-run while preserving lock behavior for mixed phase selections
- keep PGLite's data-directory connection lock intact; it is a backend single-process safety lock, not the dream cycle lock
- cover cycle-context and direct phase dry-run entry points, normal-run persistence, live-lock behavior, and spawned CLI migration preflight

## Why

`gbrain dream --phase propose_takes --dry-run` could previously persist proposal state despite being a dry run. It also performed paid model work, appended budget audit entries, and acquired file/database cycle locks.

A second preflight write surface existed before `runDream()`: the normal CLI connection automatically applied pending schema migrations. The exact propose-takes dry-run now uses the existing `connectEngine({ probeOnly: true })` seam, so it can inspect a compatible brain without advancing `config.version` or bootstrapping GBrain schema.

The phase guard still runs before all phase-owned side effects. Other dry-run phases keep their existing connection and lock semantics because their preview contracts differ.

## Verification

- `bun test test/propose-takes.test.ts test/core/cycle.serial.test.ts test/cli-dream-engine-warn.test.ts test/dream-cli-flags.test.ts` — 85 pass
- spawned current-schema/Pending-version PGLite CLI — exit 0 and `config.version` unchanged
- spawned fresh PGLite CLI — no GBrain `pages` table created
- related PGLite dream E2E — 33 pass, 5 conditional skips
- official verify check list executed sequentially — 31/31 green
  - the parallel wrapper's existing macOS timeout fallback records terminated watchdog status `143`; the same checks were rerun one by one
- `bun run typecheck` — green
- `git diff --check` and diff secret/debug-marker scan — clean
- the broad ~957-file local unit runner was not completed within the local process window; focused suites above are green and remote CI remains the authoritative full-suite gate after maintainer approval

## Prior art

Supersedes the closed prototype in #1823. This version is based on current `master`, avoids unrelated cycle wiring changes, skips budget/model/persistence and migration side effects, and adds runCycle plus spawned-CLI regression coverage.

## Test plan

- [x] cycle-level `dryRun` skips budget checks, extractor calls, proposals, receipts, and rollups
- [x] direct phase `dryRun` has the same behavior
- [x] normal runs still write proposals, receipts, and rollups
- [x] propose-takes-only dry-run ignores and preserves an existing live GBrain cycle lock
- [x] mixed dry-run selections still honor the live cycle lock
- [x] exact CLI dry-run does not advance a pending schema version
- [x] exact CLI dry-run does not bootstrap GBrain schema on a fresh PGLite data directory
- [x] neighboring CLI shapes retain normal connection semantics


time-attack:
You're right that `propose_takes` ignores dry-run today — `runCycle` never threads `dryRun` into the phase, so a preview does paid extraction and writes. But this PR bundles that ~15-line guard with release-file bumps to 0.42.59.0 (master is already 0.42.66.1, and VERSION/CHANGELOG/package.json are owned by our ship flow), plus argv sniffing in cli.ts that misses `--phase=propose_takes` and multi-phase argv, plus a change to cycle-lock acquisition. That's more risk surface than the bug warrants, so I'm closing it as-is.

PR #2727, opened 2026-07-10
fix: give autopilot a shared wedge watchdog

herove:
## Summary

- share one `WorkerWedgeWatchdog` implementation between the standalone minion supervisor and `gbrain autopilot`
- teach autopilot to restart an alive-but-wedged child after sustained queue/claim stagnation
- keep restart attempts bounded and clear watchdog state during shutdown
- document the watchdog behavior in setup and queue-operations guidance

## Why

`gbrain autopilot` could keep its child process alive while claimable jobs stopped making progress. Because the process itself had not exited, the existing crash-restart loop did not intervene. This left dead or waiting work accumulating until an operator restarted the service.

The standalone supervisor already had the right queue-level signals. This change extracts that logic into a shared component and wires the same protection into the production autopilot path.

## Behavior

- checks queue progress every 60 seconds
- treats a child as wedged only after 15 minutes / 3 consecutive qualifying checks
- derives claimable handlers from the active worker configuration
- allows at most 3 watchdog restarts per 30-minute window
- sets `GBRAIN_SUPERVISED=1` for the child and cancels the timer on shutdown

## Validation

- `bun run typecheck`
- `bun test test/supervisor-wedge.test.ts test/autopilot-supervisor-wiring.test.ts test/child-worker-supervisor.test.ts` (33 tests)
- `bun run check:doc-history`
- `bun run check:admin-build`
- `GBRAIN_HOME=/tmp/gbrain-check-home scripts/check-skill-brain-first.sh`
- static guards: operations filter, gateway routing, worker-pool atomicity, key files, no-double-retry, and batch audit

The Docker/Postgres lane in `ci:local:diff` was not run locally because Docker is not installed on this machine; the hosted CI run is expected to cover it.

### DIFF
--- a/docs/guides/queue-operations-runbook.md
+++ b/docs/guides/queue-operations-runbook.md
@@ -23,16 +23,20 @@ container health), but its DB connection died (common behind a transaction
 pooler) and never came back, so it claims no jobs and finishes nothing. Jobs
 pile up with **0 active**. Liveness checks all pass; nothing crashes.
 
-As of v0.42.22.0 this self-heals — you usually won't have to do anything:
+The dead-pool path has self-healed since v0.42.22.0, so you usually won't have
+to do anything. Current managed-worker paths also share the same progress
+watchdog: standalone `gbrain jobs supervisor` and the worker embedded by
+`gbrain autopilot`:
 
 - **The worker exits on its own dead pool.** Under a supervisor, the worker's
   DB-liveness probe runs and self-exits (`db_dead`) after ~3 minutes; the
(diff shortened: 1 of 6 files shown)


time-attack:
Thanks for the thorough writeup and the bounded-restart design. This is still a draft and it's a behavior-adding refactor of the supervisor path (auto-restarting a live child on a stagnation heuristic) rather than a fix for something concretely broken, so it falls outside what we're taking right now. Closing — no reflection on the quality of the work.

PR #2730, opened 2026-07-10
fix(#1745): atomic module-singleton rebuild — no null window on reconnect; loud cycle stamp

### LINKED ISSUE #1745: "connect() has not been called" still reproduces on 0.41.28 — concurrent minion-worker path not covered by #1570 fix
## Follow-up to #1570 — "connect() has not been called" still reproduces on 0.41.28 (concurrency path the withRetry fix doesn't cover)

Thanks for the fix on #1570. We hit the same `No database connection: connect() has not been called` error on **gbrain 0.41.28.0** with Supabase Postgres, but via a path the `withRetry`-reconnect fix does not protect: a **concurrent** operation reading the module singleton while another caller is tearing it down.

### Environment
- gbrain 0.41.28.0, engine: Postgres
- Supabase, **direct** connection `db.<ref>.supabase.co:5432` (sslmode=require, IPv6) — not the pooler
- Symptom isolated to the `dream` cycle's `synthesize` phase + the minion worker; core retrieval (search/query/embed/sync/extract) unaffected.

### Symptom
Every dream/autopilot cycle, `synthesize` fails and `synth_pages=0`, with these repeating errors:
```
Promotion error: No database connection: connect() has not been called. ...
[extract.links_fs] connection blip, retrying (attempt 1/3): No database connection: connect() has not been called. ...
Dream cycle (partial): [InternalError/SYNTH_PHASE_FAIL] No database connection: connect() has not been called. ...
totals: ... synth_transcripts=0 synth_pages=0
```

### Root cause (the remaining path)
#1570's fix made the *retrying caller* reconnect. But the singleton is still **nulled out from under other in-flight operations**:

1. A transient blip triggers `PostgresEngine.reconnect()` (`src/core/postgres-engine.ts:~4515`), which does `try { await this.disconnect(); } catch {}` then reconnects.
2. In module-singleton mode, `this.disconnect()` routes to `db.disconnect()` (`src/core/postgres-engine.ts:~223`).
3. `db.disconnect()` (`src/core/db.ts:227`) executes `if (sql) { await sql.end(); sql = null; connectedUrl = null; }` — it **nulls the shared module singleton**.
4. The minion worker loop runs concurrently: `MinionWorker` → `this.queue.promoteDelayed()` (`src/core/minions/worker.ts:463`) → `engine.executeRaw` → `getConnection()` (`src/core/db.ts:~150`). It reads `sql` during the window where it is `null` and throws "connect() has not been called".
5. The `synthesize` phase submits its writer jobs through the same minion queue (`src/core/cycle/synthesize.ts` `engine.executeRaw`, surfaced as `SYNTH_PHASE_FAIL`), so the whole phase aborts.

So `withRetry` reconnects the *one* caller it wraps, but `db.disconnect()` setting `sql = null` breaks every *other* operation sharing the singleton during the disconnect→reconnect window. On Postgres this is newly exposed because minions are Postgres-only (`jobs work` is "Postgres only"); the path never ran on PGLite. We see the in-code instrumentation comment at `db.ts:228` referencing #1570 ("identify the caller that's nulling the module singleton mid-cycle") — that caller is `reconnect()`'s `disconnect()`, and the victims are the concurrent minion-queue ops.

### Isolation facts
- Not env/config: same process with `GBRAIN_DATABASE_URL` set runs `embed`, `orphans`, `purge`, `extract.timeline_fs` fine; only the minion-queue-backed path throws.
- Not connection instability per se: non-minion phases reuse the same connection without blips.
- Not dual-worker contention: reproduces with a single `gbrain jobs work`.

### Workaround we applied (stopgap, local source patch)
Gate the singleton release in `db.disconnect()` so a normal `reconnect()` no longer nulls the shared connection — postgres.js auto-reconnects dead sockets inside the pool, so keeping the singleton alive for the process lifetime is safe:
```ts
// db.ts disconnect(): only release on an explicit full shutdown
if (sql && process.env.GBRAIN_FORCE_DISCONNECT === '1') {
  await sql.end();
  sql = null;
  connectedUrl = null;
}
```
After this, the error disappears (0 occurrences across many cycles) and `synthesize` runs to completion instead of `SYNTH_PHASE_FAIL`.

### Suggested durable fix (your call)
Either (a) don't tear down the shared module singleton inside `reconnect()` — let postgres.js's pool self-heal; or (b) make `getConnection()` lazily reconnect (await an in-flight `connect()`) instead of throwing when `sql` is transiently null; or (c) guard concurrent queue ops against the disconnect window. Happy to test a patch against our Supabase setup.

### Reproduction
1. `gbrain init --supabase --url <conn> --embedding-model zeroentropyai:zembed-1 --embedding-dimensions 1280`
2. add a source, sync content, set a synthesize corpus dir
3. run `gbrain autopilot` (or `gbrain jobs work` + `gbrain dream --synthesize`)
4. observe `synth_pages=0` + the "connect() has not been called" errors during a connection blip.

MMurray01:
Fixes #1745 ("connect() has not been called" — concurrent minion-worker path not covered by the #1570 fix).

## Root cause
Two callers invoke `PostgresEngine.reconnect()` while a dream cycle is in flight: the minion supervisor's 3-strikes health loop and `batchRetry`'s reconnect hook.  For **module-style** engines (`poolSize` unset), `reconnect()` ran `disconnect()` + `connect()` on the **process-shared db.ts singleton**.  That opens a null window (every concurrent `getConnection()` throws the #1745 error) and `sql.end()` kills in-flight sibling queries.  The #1570 fix protected the three batch primitives after the fact, but plain phase queries (sync, synthesize bookkeeping) have no retry coverage, so the cycle's phases die.

A second defect then hides the damage: the `last_full_cycle_at` stamp ignores `updateSourceConfig`'s boolean return, so a zero-row UPDATE skips the stamp silently and the run exits 0 — `cycle_freshness` reads stale for weeks with no visible failure.  (Note: `reconnect()`'s old docstring claimed module mode no-ops; it never did — `connect()` saves config unconditionally, so the destructive path was live.)

## Fix
- **`db.rebuildConnection()`** (new): build + verify the NEW client first, swap the singleton reference, then drain the old pool in the background (`end({timeout: 5})`).  No null window, ever; in-flight queries get a graceful drain.  Fail-loud: if the fresh client cannot be built, the error propagates and the old pool is left untouched.
- **`PostgresEngine.reconnect()`**: module-style engines route through `rebuildConnection()`; instance pools keep the existing teardown+rebuild (engine-private, safe).
- **`cycle.ts` stamp**: checks the return value and warns loudly on a zero-row UPDATE instead of silently skipping.

## Proof
- `tsc --noEmit` clean; 45/45 nearest unit tests pass (cycle-last-full-cycle-at, autopilot-reconnect-classifier, retry-matcher, doctor-batch-retry).
- Live concurrency test against a real Supabase Postgres: 6 concurrent workers through 8 singleton rebuilds → **444 queries, 0 connect-class errors, 0 other errors** on the fixed path.  Negative control (old disconnect+connect sequence, same load) → **547 "connect() has not been called" errors + 83 in-flight query kills**, reproducing #1745 on demand.

## Repro pedigree
Independently reproduced on three seats (three machines, one shared brain) on 0.41.28 and 0.42.1.0; diagnostics in the #1745 thread.  Real-world impact on our fleet: a source's dream cycle silently dead for ~30 days while recall stayed healthy — the silent-stamp defect made it invisible.

### DIFF
--- a/src/core/cycle.ts
+++ b/src/core/cycle.ts
@@ -2127,9 +2127,16 @@ export async function runCycle(
   // the cost of missing a successful write (next cycle will redo work).
   if (opts.sourceId && engine && !dryRun && (status === 'ok' || status === 'clean' || status === 'partial')) {
     try {
-      await engine.updateSourceConfig(opts.sourceId, {
+      const stamped = await engine.updateSourceConfig(opts.sourceId, {
         last_full_cycle_at: new Date().toISOString(),
       });
+      // v0.42.2 (#1745 follow-on): updateSourceConfig returns false when the
+      // UPDATE matched zero rows (missing/renamed source, RLS-filtered row).
+      // Ignoring it produced a SILENTLY unstamped cycle: exit 0, freshness
+      // stale, no trace.  Loud or it did not happen.
+      if (!stamped) {
+        console.warn(`[cycle] last_full_cycle_at NOT stamped for source ${opts.sourceId}: UPDATE matched zero rows (source missing or not visible to this role). cycle_freshness will read stale despite this run.`);
+      }
     } catch (e) {
       // Best-effort; cycle already succeeded by the time we get here.
       console.warn(`[cycle] failed to write last_full_cycle_at for source ${opts.sourceId}: ${e instanceof Error ? e.message : String(e)}`);
--- a/src/core/db.ts
+++ b/src/core/db.ts
@@ -159,6 +159,47 @@ export function getConnection(): ReturnType<typeof postgres> {
   return sql;
 }
 
+/**
+ * v0.42.2 (#1745): build + verify a configured postgres.js client for the
+ * module singleton. Shared by connect() (cold start) and
+ * rebuildConnection() (atomic swap) so both paths carry identical
+ * pooler-detection, timeout, and NOTICE-silencing semantics.
+ */
+async function createModuleClient(url: string): Promise<ReturnType<typeof postgres>> {
+  const prepare = resolvePrepare(url);
+  const timeouts = resolveSessionTimeouts();
+  const opts: Record<string, unknown> = {
+    max: resolvePoolSize(),
+    idle_timeout: 20,
+    connect_timeout: 10,
+    types: {
+      // Register pgvector type
+      bigint: postgres.BigInt,
+    },
+    // Silence postgres NOTICE-level messages by default ("relation already
+    // exists, skipping" floods stdout under idempotent CREATE statements
+    // during migrations + initSchema, and breaks stdout-parsing callers like
+    // `gbrain jobs submit --json | ...`). Opt back in with GBRAIN_PG_NOTICES=1.
+    onnotice: process.env.GBRAIN_PG_NOTICES === '1' ? undefined : () => {},
+  };
+  if (Object.keys(timeouts).length > 0) {
+    opts.connection = timeouts;
+  }
+  if (typeof prepare === 'boolean') {
+    opts.prepare = prepare;
+    if (!prepare) {
+      console.warn(
+        '[gbrain] Prepared statements disabled (PgBouncer transaction-mode convention on port 6543). Override with GBRAIN_PREPARE=true if your pooler runs in session mode.',
+      );
+    }
+  }
+  const client = postgres(url, opts);
+  // Test connection
+  await client`SELECT 1`;
+  await setSessionDefaults(client);
+  return client;
+}
+
 export async function connect(config: EngineConfig): Promise<void> {
   if (sql) {
     // Warn if a different URL is passed — the old connection is still in use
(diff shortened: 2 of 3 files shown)


MMurray01:
Field certification from a second seat (the fleet whose outage motivated this): patch applied clean on 0.42.1.0, and `gbrain dream` completed all 14 phases end-to-end — the source's first full cycle traversal in 30 days.  `cycle_freshness` now stamps correctly.

Two observations from the certification run that may help reviewers:

1. **Stacked defect**: underneath the reconnect teardown we found a 10-day zombie `gbrain sync` process holding the cycle lock via a **session advisory lock**.  `doctor`'s stale_locks check is row/TTL-based and structurally cannot see session-held advisory locks, so every nightly cycle instant-skipped with `cycle_already_running` and nothing flagged why.  A session-advisory-lock check in doctor would close that blind spot (happy to follow up with a PR if useful).

2. **The loud-stamp change works as intended**: with the cycle now actually traversing, a previously-hidden failure surfaced immediately (`conversation_facts_backfill` throwing the connect error) — visible failure where there was silence before.

time-attack:
Closing after verified review: #1745's reconnect null-window was fixed on master via 46f944bb and the v0.42.20.0 reliability wave (ec5fed29) with a more complete design (idempotent db.connect(), ConnectionManager refresh, reap classification, pool-recovery audit), and this branch now conflicts (63 commits behind). One idea here is still valuable and NOT on master: the loud cycle-stamp (checking updateSourceConfig's boolean return at the cycle stamp site). A fresh, rebased PR with just that change would be very welcome — thanks for the thorough concurrency work!

PR #2732, opened 2026-07-10
Guarantee complete anchor evidence in think prompts

symmetric-matthew:
## Problem
The think prompt can reason from a retrieved page without a guaranteed complete copy of the explicitly requested anchor page. Truncation can also make whole-brain absence claims look justified when the anchor itself was incomplete.

## Fix
- guarantee an anchor-page evidence section for think requests
- XML-escape anchor attributes
- carry an explicit completeness signal into the prompt
- forbid absence conclusions when the anchor page was truncated
- add serial regression coverage for complete, oversized, and escaped anchors

## Verification
- anchor test: 11 passed
- related think suite: 136 passed
- `bunx tsc --noEmit`: passed

### DIFF
--- a/src/commands/think.ts
+++ b/src/commands/think.ts
@@ -165,7 +165,10 @@ prints what would have been the input (exit 0).
     console.log('');
   }
   console.log('---');
-  console.log(`Model: ${result.modelUsed} | Pages: ${result.pagesGathered} | Takes: ${result.takesGathered} | Graph: ${result.graphHits} | Citations: ${result.citations.length}`);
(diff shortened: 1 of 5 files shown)


time-attack:
Appreciate the clear problem statement and the serial regression coverage. This is new prompt + gather surface in the think pipeline (anchor evidence section, completeness signalling, new absence-conclusion rules) rather than a fix for something broken, and we're closing feature additions by default to keep that path small. Thanks for taking the time to write it up.

PR #2733, opened 2026-07-10
Unblock existing brains upgrading to timeline event schema

symmetric-matthew:
## Problem
On an existing pre-v121 brain, schema replay creates indexes that reference `timeline_entries.event_page_id` before numbered migration 121 can add the column. `gbrain init --migrate-only` therefore fails before the migration runner reaches v121.

## Fix
Add `event_page_id` to the Postgres and PGLite forward-reference bootstraps. Migration 121 remains responsible for the FK and indexes.

## Verification
- `test/schema-bootstrap-coverage.test.ts`: 9 passed
- `bunx tsc --noEmit`: passed
- live Postgres upgrade from schema 119 to 122 completed successfully


time-attack:
Closing as superseded by #2735 (merged in v0.42.59.0), which ships the same event_page_id bootstrap in both engines with CI coverage. Thanks for the fix!

PR #2748, opened 2026-07-11
feat(extract): per-source outcomes on extract receipts

Nazim22:
## What

Extract receipts now record **which sources a round actually processed and what each yielded**. An ingest-coverage audit ("is every staged transcript actually landing as atoms?") currently requires joining atom frontmatter back to staged content by hand — the receipt knows the answer but doesn't say it.

- New optional `sources` on `ExtractReceiptInput` — `[{ ref, hash?, rows, error? }]`. `rows: 0` is the load-bearing case: it proves a source WAS attempted and legitimately yielded nothing, distinguishing "processed, atom-less" from "never reached" without DB forensics.
- Capped at `MAX_RECEIPT_SOURCES` (200) with `sources_total` in frontmatter carrying the true count when truncated; first `BODY_SOURCE_ROWS` (50) render as a body table, rest live in frontmatter only.
- `extract-atoms` wires its per-source outcomes through (drain + page-discovery paths).
- Additive: extractors that don't track per-source state simply omit the field; receipts unchanged for them.

## Testing

- `test/extract-receipt-sources.test.ts` (new) — per-source outcomes + the zero-atom receipt gap; 13 targeted tests green (`extract-receipt-sources` + `extract-atoms-drain`).
- `bun run verify` — 31/31 checks green; `bun run typecheck` clean.
- Full unit suite run chunked on a constrained box (24G LXC — the stock 4-shard runner OOMs it); the only failures reproduce identically on clean `master` (environment-dependent: git-hook/worker tests against a live gbrain install), none in the extract/receipt area.

### DIFF
--- a/src/core/cycle/extract-atoms.ts
+++ b/src/core/cycle/extract-atoms.ts
@@ -49,7 +49,7 @@ import type { PhaseResult } from '../cycle.ts';
 import type { GBrainConfig } from '../config.ts';
 import type { ProgressReporter } from '../progress.ts';
 import { chat as gatewayChat } from '../ai/gateway.ts';
-import { writeReceipt } from '../extract/receipt-writer.ts';
+import { writeReceipt, type ExtractReceiptSourceOutcome } from '../extract/receipt-writer.ts';
 import { upsertExtractRollup } from '../extract/rollup-writer.ts';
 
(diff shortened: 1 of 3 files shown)


time-attack:
Reviewed and declined for now with the observability set. Thank you.

PR #2751, opened 2026-07-11
fix(sync): refresh freshness on no-op sync

JavanC:
## Summary

Successful no-op syncs currently return `up_to_date` without updating `sources.last_sync_at`. That makes daily Dream/autopilot cycles that correctly find no changes look stale to `doctor sync_freshness` after the freshness window elapses.

This adds a small `markSyncChecked` helper and calls it on the pure HEAD-equality no-op path. It refreshes `last_sync_at` for source-scoped syncs without advancing `last_commit` or touching `newest_content_at`, preserving the blocked/partial-sync safety invariants.

## Test plan

- `bun test test/sync-resumable-import.serial.test.ts`
- `bunx tsc --noEmit`

## Notes

Validated against a dogfood source where nightly Dream had been running daily with `syncStatus: up_to_date`, while `doctor` reported the source as stale because `last_sync_at` had not changed since the last content-bearing sync.

### DIFF
--- a/src/commands/sync.ts
+++ b/src/commands/sync.ts
@@ -1001,6 +1001,23 @@ async function writeSyncAnchor(
   await engine.setConfig(`sync.${which}`, value);
 }
 
+async function markSyncChecked(
+  engine: BrainEngine,
+  sourceId: string | undefined,
+): Promise<void> {
+  if (sourceId) {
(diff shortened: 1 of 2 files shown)


time-attack:
Closing as duplicate of #1197.

Same fix: refresh source sync freshness on no-op sync. Keep #1197.

PR #2752, opened 2026-07-11
fix(cycle): enforce extract-atoms drain deadline

### LINKED ISSUE #2750: fix(cycle): extract-atoms drain window does not bound lock/query/write lifetime
## Problem

An extract-atoms-drain job configured with window=120 ran for 282.5 seconds and had to be cancelled. Doctor still reports a 60-page backlog and a 50% atoms halt rate. The window currently does not bound lock acquisition, database reads, LLM work, writes, bookkeeping, and lock release as one operation.

## Root cause

Cancellation is not propagated through the full drain lifetime. Several database and bookkeeping paths can outlive the caller deadline, and partial progress is not durably recorded before returning.

## Acceptance

- One combined AbortSignal starts before lock acquisition and covers acquisition, discovery, chat, atomic writes, bookkeeping, and release.
- Postgres direct queries are cancellable; PGLite refuses deadline mode when it cannot guarantee cancellation.
- Lock cleanup is attempt-unique and cannot release another attempt.
- Multi-atom output writes in one atomic statement.
- Partial completed work writes receipt and rollup under a bounded grace period.
- External abort rejects after releasing the lock; internal window expiry returns a truthful partial result.
- Collision/idempotency and JSONB sanitation tests cover same-title origins, NUL, and lone surrogates.
- A live window=120 canary reaches a terminal state within window plus documented grace and leaves no active lock.

panda850819:
Closes #2750.

## What changed
- threads the drain deadline through lock acquisition, backlog count, chat/query work, writes, and lock release
- cancels external worker/query work when the window expires
- gives receipts/rollups and lock release a bounded bookkeeping grace period
- refuses the unsupported PGLite path where a real query cancel cannot be guaranteed
- preserves the existing single-statement/no-inline-retry invariant in `PostgresEngine.runUnsafe`
- adds regression coverage for hung chat, hung write, partial progress, atomicity, and deadline bookkeeping

## Verification
- 75/75 targeted tests across deadline, synthesize, direct-query, and connection-resilience suites
- `bun run typecheck`
- `bun run verify` (31/31)
- live Postgres canary: prior drain job ran 282.5s; patched runtime returned in 0.369s with backlog preserved

A full unit fan-out was sampled for 14 minutes and stopped because the repository runner was on pace to exceed its own 25-minute timeout. It surfaced the pre-existing single-statement static invariant; that was fixed in `29272f1d` and its owning test now passes. Other observed failures were unrelated parallel-test fixtures (git tag editor isolation and ambient-channel state).

### DIFF
--- a/src/commands/dream.ts
+++ b/src/commands/dream.ts
@@ -449,7 +449,6 @@ async function runDrain(
   engine: BrainEngine,
   opts: DreamArgs,
   resolvedSourceId: string | undefined,
-  brainDir: string | null,
 ): Promise<void> {
   const { LockUnavailableError } = await import('../core/db-lock.ts');
   const { countExtractAtomsBacklog } = await import('../core/cycle/extract-atoms.ts');
@@ -479,7 +478,6 @@ async function runDrain(
     result = await runExtractAtomsDrainForSource(engine, {
       sourceId: resolvedSourceId,
       windowSeconds: opts.windowSeconds,
-      brainDir: brainDir ?? undefined,
       onBatch: opts.json ? undefined : ({ batch, extracted, remaining }) => {
         process.stderr.write(`[drain] batch ${batch}: +${extracted} atom(s), ~${remaining ?? '?'} remaining\n`);
       },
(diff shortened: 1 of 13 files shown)


time-attack:
Thanks for the deep work here — the root cause you identified is real and still present on master (the drain only checks its deadline at loop boundaries, so a long batch blows straight through the window). I can't take this shape, though: `runExtractAtomsDrainForSource` now throws unconditionally when `supportsQueryCancellation === false`, which means PGLite — our zero-config default engine — goes from a working best-effort drain to a hard error, and the patch also quietly changes drain semantics (drops `brainDir`, forces `_transcripts: []`) beyond the deadline fix. At 915 lines across db-lock, the engine interface, both engines, and two writers, it's also past what we can responsibly review as one unit. Closing; a much smaller change that bounds the batch itself and degrades gracefully on PGLite instead of refusing would be a different conversation.

PR #2754, opened 2026-07-11
fix(pages): preserve content_created_at through re-ingest (effective_date precedence)

vishnujayvel:
## What

Adds a nullable `pages.content_created_at` column and makes `computeEffectiveDate` prefer it, so re-ingesting a page no longer silently rewrites its effective date to the row-insert time.

## Why

The pglite → Postgres re-ingest exposed a latent bug: `putPage()`'s INSERT never set `created_at` (it fell to the DB `now()` default) and `computeEffectiveDate` never read `frontmatter.created`, so re-ingested rows' `effective_date` fell back to the insert timestamp instead of the content's real date — every re-ingest silently rewrote history.

## What changed

- New nullable `pages.content_created_at` column (additive + idempotent migration) that never overwrites `created_at`.
- `computeEffectiveDate` precedence: `content_created_at` (explicit override) first, `frontmatter.created` last (a generic signal, below the dedicated `event_date`/`date`/`published`/filename fields), before falling back to `updated_at`/`created_at`.
- `putPage()` (postgres + pglite engines), `migrate-engine.ts`'s pglite→postgres copy, and `import-file.ts` thread the new field through with COALESCE-preserve semantics.

## Verification

- `bun run verify` — 31/31 checks green (privacy, jsonb, progress, test-isolation, wasm, admin-build, resolver, typecheck).
- `bun test test/effective-date.test.ts test/pglite-engine.test.ts` — **133 pass / 0 fail**, including the new `pages_content_created_at` migration test, 8 precedence-chain unit tests, and 3 `putPage` round-trip tests.


vishnujayvel:
Merged `master` into this branch to resolve the conflict (merge commit, no rebase). Only conflict was the migration version: #2941's `configurable_fts_language` landed as v123 on master, so this PR's `pages_content_created_at` migration is renumbered to v124 (keeping master's v123 avoids schema-version drift on brains that already applied it). The FTS migration test's "is the latest" assertion is pinned to v123 accordingly, with a comment. Affected suites green locally: 168/0 unit, `bun run verify` 31/31.

time-attack:
Thank you for the careful writeup — the precedence analysis is accurate and the tests are thorough. The reason I'm closing is scope: this adds a permanent `pages.content_created_at` column, a migration, a backfill script and COALESCE-preserve threading through both engines for a case that only affects pages with no date signal at all, where `effective_date` was already a synthetic fallback rather than real history. A new date column sitting next to `created_at` is durable schema surface we'd carry forever, and the behavior you actually want looks reachable by adding `frontmatter.created` to the existing precedence chain in `computeEffectiveDate`. Closing this one on surface-area grounds, not on correctness.

PR #2764, opened 2026-07-12
fix(cli): reuse PGLite engine for reindex-frontmatter

ssentamu:
## Summary
- Pass the CLI-owned `BrainEngine` into `reindexFrontmatterCli` rather than opening a second PGLite connection.
- Make `reindexFrontmatterCli` reuse a supplied engine and disconnect only an engine it owns.
- Add a regression test that proves the command operates on data through the caller-owned engine.

## Root cause
The CLI dispatcher already opens and locks PGLite before dispatching `reindex-frontmatter`. The command then opened a second engine against the same database directory, causing it to wait on its own PGLite lock until timeout.

## Validation
- `bun test --timeout 30000 test/reindex-frontmatter-connect.test.ts`
- Regression test fails on the unfixed `master` baseline and passes with this change.
- `bun x tsc --noEmit --pretty false`
- `bun run verify` (31 checks)
- Isolated PGLite `gbrain init --pglite --no-embed` E2E setup

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -1990,8 +1990,11 @@ async function handleCliOnly(command: string, args: string[]) {
         // v0.30.1: still works; canonical entrypoint is now `gbrain backfill
         // effective_date`. This command stays as a thin alias for back-compat.
         const { reindexFrontmatterCli } = await import('./commands/reindex-frontmatter.ts');
-        await reindexFrontmatterCli(args);
-        return; // reindexFrontmatterCli handles its own engine lifecycle
+        // This dispatcher already owns a connected engine. Passing it through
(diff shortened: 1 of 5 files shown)


time-attack:
Confirmed and merging. The dispatcher does connect its own engine before the `reindex-frontmatter` case, and `PGLiteEngine.connect()` takes an exclusive file lock, so the second engine this command opened deadlocked against our own process on the default engine — the command was flatly broken for PGLite users. The caller-owned-engine fix with ownership-aware disconnect is the right minimal shape and the regression test earns its keep. One note for next time: the `detectTini` change in `spawn-helpers.ts` is an unrelated fix riding along in this PR and would have been cleaner as its own.

PR #2777, opened 2026-07-13
v0.42.58.0 fix(cli): expose book-mirror detailed help (#2774)

### LINKED ISSUE #2774: bug(cli): book-mirror dispatcher is unreachable in v0.42.58.0
## Reproduction

On macOS with the current upstream HEAD (`a25209bbb2bacf1b88e06fd5282b27f1bf4a3e7a`, `gbrain 0.42.58.0`):

```bash
gbrain --version
# gbrain 0.42.58.0

gbrain book-mirror --help
# Usage: gbrain book-mirror
#
# gbrain book-mirror - run gbrain --help for the full command list.
```

The same result occurs when invoking the installed entrypoint directly with `bun run src/cli.ts book-mirror --help`.

## Expected

The Book Mirror help text from `src/commands/book-mirror.ts`, then normal execution via:

```bash
gbrain book-mirror --chapters-dir <path> --context-file <path> --slug <slug>
```

This matches the v0.25.1 changelog and the bundled `skills/book-mirror/SKILL.md`.

## Diagnosis

The implementation and downstream switch both exist:

- `src/commands/book-mirror.ts` exports the command implementation.
- `src/cli.ts` has `case 'book-mirror'` at ~1654.

But top-level dispatch checks `CLI_ONLY`, then `cliOps`/`cliAliases`, before reaching that later switch. `book-mirror` is not registered in `CLI_ONLY`, `cliOps`, or `cliAliases`, so it exits through the generic unknown-command path and the handler is unreachable.

`gbrain skillpack scaffold book-mirror --dry-run` copies the skill and shared conventions only (`pairedSourcesWritten: 0`), so scaffolding does not provide an alternate runnable command.

## Requested fix

Route `book-mirror` through the engine-backed command dispatcher before the shared-operation lookup, add a regression test for `gbrain book-mirror --help`, and release a tagged fix.

I am deliberately avoiding a local edit to the global install so that upgrades remain reproducible.

RerankerGuo:
## Summary

- route `book-mirror --help` and `book-mirror -h` to the command's detailed help before opening a brain
- keep normal `book-mirror` execution on the existing engine-backed dispatcher
- add a spawned-CLI regression test using an empty `GBRAIN_HOME`

## Root cause

`book-mirror` was already present in `CLI_ONLY`, so normal execution reached its handler. The per-command help short-circuit ran earlier, however, and emitted the generic one-line fallback because `book-mirror` was missing from `CLI_ONLY_SELF_HELP`. Simply bypassing that fallback would still try to connect a brain before the handler parsed `--help`.

The fix follows the existing `capture`/`sync`/`enrich` pattern: mark the command as self-helping and dispatch help before engine connection.

## Validation

- `bun test test/book-mirror.test.ts` (10 pass)
- `bun run typecheck`
- `bun run verify` (31 checks pass)

`ci:local:diff` is unavailable in this environment because the Docker daemon socket is inaccessible and `gitleaks` is not installed.

Fixes #2774

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -68,6 +68,8 @@ const CLI_ONLY_SELF_HELP = new Set([
   'capture',
   // v0.42 self-upgrade ships its own usage (flags + the agent-skill story).
   'self-upgrade',
+  // book-mirror ships detailed flag/cost help in its command module.
+  'book-mirror',
   // v0.43 (#2095): watch ships WATCH_HELP (flags + the stdin-turn protocol).
   'watch',
   // v0.37 fix wave (Lane D.4 + CDX2-12): sync's --no-embed flag was
@@ -1489,6 +1491,12 @@ async function handleCliOnly(command: string, args: string[]) {
   if (command === 'enrich' && (args.includes('--help') || args.includes('-h'))) {
     const { runEnrich } = await import('./commands/enrich.ts');
     await runEnrich(null as never, args);
+    return;
+  }
+
+  if (command === 'book-mirror' && (args.includes('--help') || args.includes('-h'))) {
+    const { runBookMirrorCmd } = await import('./commands/book-mirror.ts');
+    await runBookMirrorCmd(null as never, args);
     return;
   }
 
--- a/test/book-mirror.test.ts
+++ b/test/book-mirror.test.ts
@@ -6,16 +6,10 @@
  * the opt-in smoke test (test/e2e/skill-smoke-openclaw.test.ts when
  * EVALS=1 EVALS_TIER=skills is set).
  *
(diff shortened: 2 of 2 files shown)


time-attack:
Verified and merging. `book-mirror` really is in `CLI_ONLY` but missing from `CLI_ONLY_SELF_HELP`, so the generic stub fired and the command's own help block was unreachable — and as you note, the set entry alone would still have demanded a configured brain just to print usage. The pre-connect bypass matches the existing `eval cross-modal` / `longmemeval` / `dream` pattern in the same file, and the empty-`GBRAIN_HOME` spawn test is exactly the right proof. Thanks for the precise root-cause writeup.

PR #2800, opened 2026-07-13
fix(ai): tier-configured models reach the recipe allowlist — current Anthropic models, tier registration, honest probe labels

p3ob7o:
## Why

Configuring a current-generation model on a tier silently disables think and auto_think. On a live brain, `gbrain config set models.tier.deep anthropic:claude-opus-4-8` made every `think` run degrade to the gather-only stub — and `auto_think` produce nothing nightly — with the warning `NO_ANTHROPIC_API_KEY`, which sent us debugging env vars and keychains for an hour. The key was fine. Three stacked defects:

1. **The Anthropic recipe's allowlist stops at Opus 4.7.** `claude-opus-4-8` (GA since May), `claude-sonnet-5`, and `claude-fable-5` are all rejected as `unknown_model`.
2. **Tier-resolved models never join the extended set.** `assertTouchpoint`'s docstring says config-chosen models — `models.default` and `models.tier.*` included — bypass the native allowlist, but `reconfigureGatewayWithEngine` only registers chat/expansion/embedding/reranker. A model reachable *only* through a tier fails `probeChatModel`. (Sonnet 5 on the reasoning tier works today purely by side effect: `models.chat` resolves to it at startup.)
3. **think mislabels every probe failure as `NO_ANTHROPIC_API_KEY`.** An unusable model id and a missing key are different incidents with different fixes; the stub hid the real reason.

## What

- `src/core/ai/recipes/anthropic.ts` — chat models gain `claude-fable-5`, `claude-opus-4-8`, `claude-sonnet-5`; expansion gains `claude-sonnet-5`.
- `src/core/ai/gateway.ts` — `reconfigureGatewayWithEngine` now resolves all four tiers and registers the results as extended models, honoring the documented contract. Resolving per-tier also picks up `models.default` (it sits above tiers in the chain), pinned by a test.
- `src/core/think/index.ts` — on a null gateway client, re-probe (pure, no IO) and label honestly: `MODEL_NOT_USABLE:<reason>` with the probe detail and fix hint in the answer/gaps; `NO_ANTHROPIC_API_KEY` only when the key is actually the problem. The graceful no-throw degrade and the explicit-`--model` hard-error fork are unchanged.

## Tests

- `test/anthropic-model-ids.test.ts` — presence pins for the three new chat ids + Sonnet 5 expansion.
- `test/gateway-tier-extended-models.test.ts` (new) — a deliberately fictional `models.tier.deep` id validates after reconfigure while an unconfigured id stays rejected (so the allowlist still bites), plus the `models.default` path.
- `test/think-pipeline.serial.test.ts` — new test: unusable configured model yields `MODEL_NOT_USABLE:unknown_model`, not `NO_ANTHROPIC_API_KEY`, even keyless (validity beats the key check in `probeChatModel`); the existing non-explicit bogus-provider test updated from the old catch-all label to the honest one (its no-throw contract is unchanged).
- `test/think-gateway-adapter.test.ts` — client builds for all three new ids.

Touched suites (think pipeline/adapter/intent/trajectory/calibration, gateway model-messages/embed-override, anthropic ids/pricing, takes-mcp-allowlist): 165 pass / 0 fail. `bun run typecheck` clean.

Verified live on the affected brain: with the fix, the deep-tier probe passes and think synthesizes on Opus 4.8.

Related: #2799 (pricing rows for the same models).

### DIFF
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -502,6 +502,20 @@ export async function reconfigureGatewayWithEngine(engine: BrainEngine): Promise
   const expansionFull = newExpansion.includes(':') ? newExpansion : prefixWithProviderFrom(cfg.expansion_model ?? DEFAULT_EXPANSION_MODEL, newExpansion);
   const chatFull = newChat.includes(':') ? newChat : prefixWithProviderFrom(cfg.chat_model ?? DEFAULT_CHAT_MODEL, newChat);
 
+  // ALSO resolve the four tier models and register them as extended models.
+  // assertTouchpoint's contract (model-resolver.ts) says config-chosen models —
+  // `models.default` and `models.tier.*` included — bypass the native recipe
+  // allowlist, but pre-fix only chat/expansion/embedding/reranker were
+  // registered. A model reachable ONLY through a tier (e.g. `models.tier.deep`
+  // set to an Opus newer than the recipe list) failed `probeChatModel` at call
+  // time and silently degraded think/auto_think to the gather-only stub.
+  // Resolving per-tier also honors `models.default` (it sits above tiers in
+  // the resolveModel chain).
+  const tierModels: string[] = [];
+  for (const tier of ['utility', 'reasoning', 'deep', 'subagent'] as const) {
+    tierModels.push(await resolveModel(engine, { tier, fallback: TIER_DEFAULTS[tier] }));
+  }
+
   _config = { ...cfg, expansion_model: expansionFull, chat_model: chatFull };
   _modelCache.clear();
   _shrinkState.clear();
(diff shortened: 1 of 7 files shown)


p3ob7o:
CI triage for the `test (1)` failures: all failing assertions are `expected 1280 dimensions, not 1536` in the `checkHiddenBySearchPolicy` suite — doctor tests this PR doesn't touch (the diff is recipes/anthropic.ts, gateway tier registration, think labeling, and their tests).

This is the same latent shard-order flake documented on #2629/#2630: the test hardcodes vectors at one dimension but lets `initSchema` size the embedding column from whichever test file ran earlier in the shard (1280 = zembed, 1536 = openai). #2629's triage pinned the dims in `engine-find-trajectory.test.ts`; the same class evidently lives in the hidden-by-search-policy suite and moved shards with this PR's new test file changing the shard packing. Happy to pin the dims in that suite too if useful — either here or as a separate one-liner PR.

All other checks are green, including `serial-tests`, both LLM-skills tiers, and shards 2–10.

p3ob7o:
The dims pin for the `checkHiddenBySearchPolicy` suite is up as #2801 — same remedy as the two prior pins of this class. A re-run of `test (1)` here should go green independently, but #2801 removes the order-dependence for good.

PR #2802, opened 2026-07-13
fix(gateway,subagent): persist tool results per turn, drop null-text blocks on conversion

### LINKED ISSUE #2273: gateway-loop subagent path never persists tool-result user turns — every resume rebuilds an unbalanced conversation that next chat() rejects

### LINKED ISSUE #2609: gateway-native loop: tool results executed but never threaded back for OpenAI-compatible providers — subagent dies with 'Tool results are missing'
## Symptom

Dream-cycle synthesize subagents (`agent.use_gateway_loop=true`, provider `deepseek` via `provider_base_urls.deepseek` custom base URL) die on any transcript where the model uses tools:

```
[chat(deepseek:deepseek-v4-pro)] Tool results are missing for tool calls call_3f2f11a684ea42a692eab6b4, call_2307aad6f31f465cbe7e0bcc.
```

## Key evidence: the tools DID run

`subagent_tool_executions` shows both calls completed, with tool_use_ids exactly matching the ones the error claims are missing:

```
 job_id |          tool_use_id          |   tool_name    |  status
--------+-------------------------------+----------------+----------
  12512 | call_3f2f11a684ea42a692eab6b4 | brain_search   | complete
  12512 | call_2307aad6f31f465cbe7e0bcc | brain_get_page | complete
```

So execution + logging succeed, but the loop's history builder never appends the tool-result messages before the next model turn, and its own validator then rejects the history. Reproduced with both `deepseek-v4-flash` and `deepseek-v4-pro`, with 6 parallel calls and with 2. Injecting `parallel_tool_calls: false` at a proxy doesn't help (and a tool-less synthesize run on the same setup completes fine — job 12507).

Looks like the result-threading path only handles the Anthropic-loop message shape; OpenAI-compatible tool_call ids never get matched back. Sibling of the non-Anthropic blind spots reported in #2577.

Env: gbrain 0.42.56.0, Postgres engine, macOS (mini), model via OpenAI-compatible aggregator behind custom base URL.

justemu:
> **Final update (2026-07-13):** PR description rewritten after thorough investigation + 3 rounds of batch testing totaling ~60 test submissions. See [#2803](https://github.com/garrytan/gbrain/issues/2803) and [vercel/ai#17144](https://github.com/vercel/ai/issues/17144) for upstream bug reports.

---

## Summary

Fixes two distinct subagent bugs that together caused ~85-92% subagent job failure in nightly dream-cycle batch processing. Both bugs live in the AI SDK tool-loop and message-conversion code paths (`src/core/ai/gateway.ts`, `src/core/minions/handlers/subagent.ts`).

After this PR: **MissingToolResultsError → 0 cases. InvalidPromptError → retry recovery at ~97.5%**. Expected production success rate is now ~97.5% (with the remaining ~2.5% transient failures recoverable on retry).

---

### Root Cause

A dual interface mismatch between gbrain's `ChatBlock` (internal) and AI SDK v6's `ModelMessage` (SDK layer):

1. **`modelMessageSchema` (Zod validation, AI SDK core):** expects tool-call content parts to have field name **`args`**
2. **Wire converter (`@ai-sdk/openai-compatible` v2.x):** reads tool-call content parts using field name **`input`**

gbrain's `toModelMessages` function (the translation seam from `ChatBlock` → `ModelMessage`) was sending only `input`. This caused:

- **Any job ≥ 2 turns:** Zod rejection at `standardizePrompt()` because `args` was missing → 100% deterministic failure
- **Fresh 1-turn jobs:** passed through (no tool-call blocks in messages yet), so ~8-13% of jobs randomly succeeded

The Zod v4 `discriminatedUnion` lazy initialization in the AI SDK **also** has a mild race condition (~50% first-call failure), which compounded the problem but was not the primary cause.

### Fix (gateway.ts:2422)

Dual-field emission for tool-call blocks, satisfying both consumers:

```typescript
// BEFORE (missing args → Zod rejects):
{ type: 'tool-call', toolCallId, toolName, input: b.input }

// AFTER (both args for Zod + input for wire converter):
{ type: 'tool-call', toolCallId, toolName, args: b.input, input: b.input }
```

### Additional Defensive Changes (gateway.ts)

1. **User text-only content → string** (line 2408): `@ai-sdk/openai-compatible` rejects `{type:'text', text:'...'}` in user content arrays. When retry loads messages from DB (stored as block arrays), flatten to bare string.
2. **Null-text block filtering** (line 2411, inherited from prior InvalidPrompt fix for reasoning models)
3. **`bun update ai`** (6.0.174 → 6.0.224): picks up Zod v4 compatibility fixes

---

### Root Cause

gbrain's `toolLoop` (gateway.ts) calls `generateText()` with tool-call blocks in the assistant's response, then the model provider validates that every tool-call has a matching tool-result in the messages. The original code **did not persist** tool results between `chat()` calls — they were reconstructed from ephemeral in-memory state.

On subagent retry (which reloads conversation from DB), the reloaded messages contained assistant tool-call blocks but **no matching tool-result blocks**, causing `convertToLanguageModelPrompt` to throw `MissingToolResultsError`. This was **deterministic**: every multi-turn job that crashed and retried would die on the next turn.

### Fix (subagent.ts, gateway.ts)

1. **`onTurnComplete` callback** (subagent.ts): persists each toolLoop iteration's assistant (tool-calls) + matching user (tool-results) messages to `subagent_messages` table
2. **Message reconstruction** (subagent.ts `runSubagentViaGateway`): on retry, loads persisted tool-call/tool-result pairs from DB. `toolLoop` receives `replayState` with complete prior messages including tool results.
3. **`toModelMessages` tool-role conversion**: user messages with tool-result blocks are converted to `{role:'tool', content:[...]}` for AI SDK's three-role (user/assistant/tool) message schema

---

## Testing Results

Clean single-job test (20632, 3 turns with brain_search tools, dual-field + onTurnComplete fixes active):
```
Status:     completed
Attempts:   1/3 (started: 2 — attempt 1 InvalidPrompt, retry succeeded)
Turns:      3
Tokens:     41,729 in / 1,856 out
Output:     Full synthesis page with 7 sections, 10 cross-links to existing brain pages
MissingToolResults:  0
InvalidPrompt:      triggered on attempt 1, recovered on retry
```

---

## Related

- Closes #2273 (MissingToolResultsError)
- Closes #2609 (subagent crash-replay)
- Closes #2488 (InvalidPrompt from reasoning model null-text)
- References #2803 (remaining InvalidPrompt from AI SDK Zod race condition)
- References vercel/ai#17144 (upstream AI SDK bug report)

---

## Remaining Issue

~50% of first `generateText()` calls in a process hit a Zod v4 `discriminatedUnion` race in AI SDK v6's `modelMessageSchema`. This is an **AI SDK internal bug** — retry with freshTools + backoff recovers it. The fix for this requires waiting for an AI SDK v7 upgrade (breaking API changes) or an upstream Zod v4 patch. Current recovery rate: **~97.5%** on retry.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -8,7 +8,7 @@
         "@ai-sdk/anthropic": "^3.0.71",
         "@ai-sdk/google": "^3.0.64",
         "@ai-sdk/openai": "^3.0.53",
-        "@ai-sdk/openai-compatible": "^2.0.41",
+        "@ai-sdk/openai-compatible": "^2.0.59",
         "@anthropic-ai/sdk": "^0.30.0",
         "@aws-sdk/client-s3": "^3.1028.0",
         "@dqbd/tiktoken": "^1.0.22",
@@ -59,7 +59,7 @@
 
     "@ai-sdk/openai": ["@ai-sdk/openai@3.0.58", "", { "dependencies": { "@ai-sdk/provider": "3.0.10", "@ai-sdk/provider-utils": "4.0.26" }, "peerDependencies": { "zod": "^3.25.76 || ^4.1.8" } }, "sha512-2+5xGMROmrBboJuoOwqLL3b/o3i56+NRdxXDNVAiTyYjLiBj6KzembeuyuBT217be1X+zkEfAqD1H0irJlGIyw=="],
 
-    "@ai-sdk/openai-compatible": ["@ai-sdk/openai-compatible@2.0.45", "", { "dependencies": { "@ai-sdk/provider": "3.0.10", "@ai-sdk/provider-utils": "4.0.26" }, "peerDependencies": { "zod": "^3.25.76 || ^4.1.8" } }, "sha512-5YBvurNL7Oj7mT3srws4Rh4cQidoorfEGObAOb5jV40eld8IC7EkXWARZjnWYqgYzabUs6Sn6muiXfQVkgOyOQ=="],
+    "@ai-sdk/openai-compatible": ["@ai-sdk/openai-compatible@2.0.59", "", { "dependencies": { "@ai-sdk/provider": "3.0.14", "@ai-sdk/provider-utils": "4.0.38" }, "peerDependencies": { "zod": "^3.25.76 || ^4.1.8" } }, "sha512-CFsUizO+jL+jSlN13rW2nQE9EbWx+8PSFBdB3TtD0UGYOxtefCVa5hsrNo5NUOJJGX5f4xHiXE1i2m5nQ80QPg=="],
 
     "@ai-sdk/provider": ["@ai-sdk/provider@3.0.10", "", { "dependencies": { "json-schema": "^0.4.0" } }, "sha512-Q3BZ27qfpYqnCYGvE3vt+Qi6LGOF9R5Nmzn+9JoM1lCRsD9mYaIhfJLkSunN48nfGXJ6n+XNV0J/XVpqGQl7Dw=="],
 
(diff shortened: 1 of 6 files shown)


time-attack:
Closing as duplicate of #2274.

Competing PR for the same gateway/subagent tool-result persistence fix (also covered by consolidation stack). Keep #2274 (and/or #2820 if that supersession lands).

Issue #2138, opened 2026-06-12
propose_takes: `ON CONFLICT DO NOTHING` on the per-page unique key silently drops claim #2+ for multi-claim pages


Topture:
## Summary

`take_proposals` inserts use `ON CONFLICT DO NOTHING` against a unique key that is effectively **per page** (`source_id, page_slug, content_hash, prompt_version`). When the extractor returns multiple claims for one page, the first INSERT succeeds and every subsequent claim from that page is silently discarded — no error, no log, no proposal row.

## Impact

Multi-claim pages (often the richest ones — session summaries, decision logs) are systematically reduced to a single proposal. The data loss is invisible: the cycle reports success and the page is cached as done, so the dropped claims are never re-extracted.

## Repro

1. Ingest a page that clearly contains 2+ distinct take-worthy claims.
2. Run propose_takes.
3. `SELECT count(*) FROM take_proposals WHERE page_slug = ...` -> 1 row, regardless of how many claims the LLM returned (visible in the cycle's LLM output/logs).

## Suggested fix

Include a claim discriminator (claim hash or ordinal) in the unique key, or switch to an upsert keyed on (page, claim_hash). Observed on gbrain 0.42.40.0.

## Related

Filed alongside two companion issues from the same cost-runaway investigation of the propose_takes loop: the zero-claim cache miss and the cycle budget config disconnect.

sene1337:
Corroborating from another production Postgres brain on `gbrain 0.42.53.0` (`HEAD f2e24fa8`; fetched `origin/master 058f448b`, same code shape still present).

I hit the same silent multi-claim drop in the nightly `dream` / `propose_takes` lane. The useful diagnostic was comparing the phase's `details.proposals_inserted` in the JSON logs against rows that actually survived in `take_proposals` for the same `proposal_run_id`s.

Two consecutive nightly runs:

- 2026-07-05: 10 sources, 826 pages scanned, 376 cache misses, log reported 121 `proposals_inserted`; DB had 25 proposal rows for those runs.
- 2026-07-06: 10 sources, 832 pages scanned, 366 cache misses, log reported 101 `proposals_inserted`; DB had 12 proposal rows for those runs.
- Combined: 222 reported insert attempts vs 37 persisted proposal rows.

The downstream review/drain path then cleanly accepted/rejected all 37 persisted rows:

- 32 accepted / promoted into canonical `takes`
- 5 rejected
- 0 pending remaining

[...]

PR #2804, opened 2026-07-13
fix(takes): kill the propose_takes rescan loop — zero-claim scans cache, multi-claim pages keep every claim

p3ob7o:
## Why

`propose_takes` was ~90% of a live brain's autopilot LLM spend — ~1,400 extractor calls (~$15) per day, around the clock — and most of it bought nothing, twice over. Receipts from the audit streams and cycle reports:

- Back-to-back cycle runs six minutes apart: `cache_misses: 61`, then `cache_misses: 57` — the same pages re-extracted every run, ~62/hour steady including overnight.
- Over 3 days: 1,312 distinct (page, content_hash) scans produced exactly 1,312 rows — one row per scan, no matter how many claims a page yielded.

## The two defects

1. **Rescan loop.** Only proposal rows entered the idempotency cache. A page whose extraction yielded zero gradeable claims left no row for its (source_id, page_slug, content_hash, prompt_version), so the cache check missed on every subsequent cycle and the LLM call was re-spent on the same unchanged page, forever. The phase's own receipt pages fed back into the loop the same way (new page → scanned → zero claims → rescanned forever).

2. **Dropped claims.** The 4-column unique index collapsed a same-page multi-claim run to its FIRST claim: rows 2..N hit the conflict and `ON CONFLICT DO NOTHING` silently dropped them. The review queue has only ever held the first claim of each page version. The `proposals_inserted` counter also counted attempts, not landed rows (reported 163 on a run where at most ~60 could land).

## What

**Migration v123** (`take_proposals_empty_scan_sentinels_and_per_claim_rows`):
- status CHECK gains `'empty'`.
- The idempotency index gains `md5(claim_text)` — per-claim rows; the 4-column prefix still serves the per-scan cache lookup. Existing data is index-safe by construction: the old index guaranteed at most one row per tuple, so the widened index has no duplicates to trip on. The DROP+ADD constraint pair is idempotent as a unit.
- Fresh-install DDL updated in `src/schema.sql` (embedded schema regenerated) and `pglite-schema.ts`.

**Phase (`src/core/cycle/propose-takes.ts`):**
- Zero-claim scans write a `status='empty'` sentinel row (empty claim_text) so the cache check hits next cycle. Sentinels are excluded from the partial pending index — the review queue never sees them.
- Proposal inserts use the 5-column conflict target and `RETURNING id`; `proposals_inserted` now counts rows that actually landed.

## Tests

New `test/propose-takes-rescan.test.ts` (PGLite + injected extractor, hermetic): second run never re-extracts unchanged pages; third run is fully quiescent (zero extractor calls, zero cache misses — this is the assertion that fails on pre-fix code, where misses never converge); the 3-claim page stores 3 rows; the sentinel is invisible to the pending queue; a content change re-extracts exactly once and old sentinel + new pending rows coexist. The existing suite's mock engine now emulates `INSERT..RETURNING` for the honest counter. Ran with `test/propose-takes.test.ts`, `test/core/cycle.serial.test.ts`, `test/migrate.test.ts`, `test/schema-bootstrap-coverage.test.ts`: 227 pass / 0 fail. Typecheck clean.

## Expected impact

On the affected brain this drops autopilot steady-state spend from ~$17/day to roughly $2–4/day (probes + genuinely new content). One deployment note: code-before-migration produces a visible phase warn (the ON CONFLICT target needs the new index) rather than a silent misfire; the standard upgrade flow applies migrations first.

Related: #2638 fixes the same defect class (zero-claim pages never marked covered) in the separate takes-bootstrap classifier.

### DIFF
--- a/src/core/cycle/propose-takes.ts
+++ b/src/core/cycle/propose-takes.ts
@@ -5,11 +5,17 @@
  * a tuned LLM extractor, writes the extracted gradeable claims to the
  * `take_proposals` queue. User accepts/rejects via `gbrain takes propose`.
  *
- * Idempotency contract (D17 schema spec):
- *   The unique index on (source_id, page_slug, content_hash, prompt_version)
- *   means an unchanged page never re-spends LLM tokens. Bumping
- *   PROPOSE_TAKES_PROMPT_VERSION cleanly invalidates the cache so a tuned
- *   prompt re-runs proposals on every page.
+ * Idempotency contract (D17 schema spec; per-claim rows since migration
+ * v125, zero-claim sentinels since v126):
+ *   Every scan of a (source_id, page_slug, content_hash, prompt_version)
+ *   tuple leaves at least one row — one per extracted claim, or a single
+ *   status='empty' sentinel when extraction yields nothing — so an unchanged
+ *   page never re-spends LLM tokens. Pre-v126 only proposal rows were
+ *   written: a zero-claim page never entered the cache and was re-extracted
+ *   on EVERY cycle (observed live: ~60 such pages × every cycle ≈ 1,400
+ *   wasted extractor calls / ~$15 per day — ~90% of total autopilot spend).
+ *   Bumping PROPOSE_TAKES_PROMPT_VERSION cleanly invalidates the cache so a
+ *   tuned prompt re-runs proposals on every page.
  *
  * F2 fence dedup:
  *   The phase reads the page's existing `<!-- gbrain:takes:begin -->` fence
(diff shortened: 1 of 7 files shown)


p3ob7o:
CI triage: the `test (1)` failure is the known 1280/1536 dims shard-order flake in `checkHiddenBySearchPolicy` — same signature as on #2800, fixed by #2801 (green). Everything this PR touches passes (`test/propose-takes*`, `cycle.serial`, `migrate`, `schema-bootstrap-coverage` all in passing shards).

loweaxerium:
Independent confirmation from a live brain (Postgres, ~300 pages). Hit exactly this before finding your PR: a `propose_takes` run scanned 100 pages, the extractor proposed 90 claims, but only **23 rows** landed — one per page, `max(count)=1` across all pages. The 4-column unique index + `ON CONFLICT DO NOTHING` silently dropped claims 2..N of every multi-claim page, and `proposals_inserted += 1` (unconditional) reported 90 while the table held 23.

Cherry-picked your commit onto our fork — fix confirmed end-to-end: v123 applies clean, the `md5(claim_text)` index keeps every claim, the `status='empty'` sentinel kills the rescan loop, suite green (29 takes + 161 migrate incl. v123). The rescan-loop half was the bigger win for us — it was ~90% of nightly autopilot spend. Thanks for the thorough write-up.

p3ob7o:
Rebased onto current master (post-v0.42.64). One adaptation: master gained migrations v123 (configurable_fts_language) and v124 (page_search_vector_drop_compiled_truth) since this PR was filed, so this PR's migration is renumbered v123 → **v125**. No functional changes — the migration body, phase change, and tests are identical. Locally green: typecheck clean, 192 tests across the takes + migrate suites.

p3ob7o:
Rebased onto current master (post-v0.42.65) and **rescoped**: the per-claim md5(claim_text) index half of this PR landed upstream as migration v125 (`take_proposals_per_claim_idempotency`, #2138 class) — great to see it confirmed independently. This PR now carries only the remaining half: the zero-claim sentinel cache (status CHECK += 'empty' as migration **v126** + the sentinel write in the phase), i.e. the fix for the rescan loop that was ~90% of autopilot LLM spend on a live brain. The honest RETURNING-based counter now matches upstream's shipped implementation. Locally green: typecheck clean, 203 tests across the takes + migrate suites.

time-attack:
Thank you for this — the receipts-driven diagnosis was exactly right, and both defects you identified have since been fixed on master: the zero-claim sentinel/memoization landed via #3319 (commit 3fcca330) and the per-claim rows preserving multi-claim pages landed via #2574's sibling #3297 (commit ca4cf2a0). Master's `propose-takes.ts` now memoizes the empty case and uses the 5-column `ON CONFLICT (..., md5(claim_text))` target. Your migration slot 123 is also now taken by `configurable_fts_language`, so this can't apply cleanly. Closing as already fixed — the analysis genuinely helped.

PR #2805, opened 2026-07-13
feat(models): cycle phases and brainstorm honor the model-tier system

p3ob7o:
## Why

Four LLM call sites — `propose_takes`, `grade_takes`, `calibration_profile`, and the brainstorm orchestrator — hardcode `'claude-sonnet-4-6'`, but only as the **label**. The actual calls pass no model and ride the gateway's `chat_model`. On a tier-configured brain (`models.tier.reasoning` set) the two disagree: the calls run the tier model while budget metering prices — and `take_proposals.model_id` / `take_grade_cache.judge_model_id` / evidence signatures record — a model that never ran. And the sites ignore `models.default` / `models.tier.*` entirely, so the tier system's contract ("route every role through config") silently doesn't apply to the cycle's biggest spenders.

## What

- Each site resolves once through `resolveModel` with a new per-task config key (`models.propose_takes`, `models.grade_takes`, `models.calibration_profile`, `models.brainstorm`), falling through to `models.default` → `models.tier.reasoning` → the unchanged hardcoded fallback. **Stock installs resolve to exactly what ran before** (the reasoning-tier default is `claude-sonnet-4-6`).
- **One resolved string drives everything**: the FULL form goes to the chat call; the bare tail (`splitProviderModelId`) goes to budget labels, stored model ids, and grade-cache keys. The normalization matters: tier defaults resolve provider-prefixed, and without it every stock install's grade cache would have invalidated from a pure prefix change (`claude-sonnet-4-6` → `anthropic:claude-sonnet-4-6`). With it, stock caches are untouched — the existing cache-continuity tests pass unchanged and now pin that property. A genuinely different tier-configured judge still invalidates its cache, which is correct.
- `MODEL_CONTEXT_TOKENS` (synthesize) and `ANTHROPIC_OUTPUT_CAPS` (brainstorm judges) gain `claude-fable-5` / `claude-opus-4-8` / `claude-sonnet-5` rows.
- `gbrain models` lists the four new per-task keys, so the routing is visible.

## Tests

New tier-routing regression in `test/propose-takes.test.ts`: a `models.tier.reasoning` override reaches the extractor hint as the full string and the stored `model_id` as the bare tail. Mock engines gain `getConfig`. 130 tests across the 7 affected suites (propose/rescan/grade/ensemble/calibration/cycle-serial/brainstorm-timeout) pass; typecheck clean.

**Stacked on #2804** (both touch `propose-takes.ts`) — merge that first; this PR then reduces to its own commit.

### DIFF
--- a/src/commands/models.ts
+++ b/src/commands/models.ts
@@ -50,6 +50,10 @@ const PER_TASK_KEYS: Array<{ key: string; tier: ModelTier; description: string }
   { key: 'models.eval.contradictions_judge', tier: 'utility',  description: 'Contradiction probe judge (v0.34 temporal-aware)' },
   { key: 'models.expansion',                tier: 'utility',   description: 'Query expansion for hybrid search' },
   { key: 'models.chat',                     tier: 'reasoning', description: 'Default `gateway.chat()` model' },
+  { key: 'models.propose_takes',            tier: 'reasoning', description: 'propose_takes claim extractor' },
+  { key: 'models.grade_takes',              tier: 'reasoning', description: 'grade_takes verdict judge' },
+  { key: 'models.calibration_profile',      tier: 'reasoning', description: 'Calibration profile generator' },
+  { key: 'models.brainstorm',               tier: 'reasoning', description: '`gbrain brainstorm` orchestrator' },
 ];
 
 interface ModelEntry {
--- a/src/core/brainstorm/judges.ts
+++ b/src/core/brainstorm/judges.ts
@@ -61,7 +61,10 @@ export const MAX_OUTPUT_TOKENS_CEIL = 32_000;
  * (with a readable error) instead of the provider's opaque HTTP 400.
  */
 export const ANTHROPIC_OUTPUT_CAPS: Record<string, number> = {
+  'claude-fable-5': 64_000,
+  'claude-opus-4-8': 32_000,
   'claude-opus-4-7': 32_000,
+  'claude-sonnet-5': 64_000,
   'claude-sonnet-4-6': 64_000,
   'claude-haiku-4-5': 64_000,
   'claude-haiku-4-5-20251001': 64_000,
--- a/src/core/brainstorm/orchestrator.ts
+++ b/src/core/brainstorm/orchestrator.ts
@@ -32,6 +32,7 @@
  */
 
 import type { BrainEngine } from '../engine.ts';
+import { resolveAlias } from '../model-config.ts';
 import { chat as defaultChat, embedQuery, type ChatResult, type ChatOpts } from '../ai/gateway.ts';
 import { hybridSearch, hybridSearchCached } from '../search/hybrid.ts';
 import { fetchFar, type CloseRef, type FarPage } from './domain-bank.ts';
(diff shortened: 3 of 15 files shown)


p3ob7o:
CI triage: the `test (1)` failure is the known 1280/1536 dims shard-order flake in `checkHiddenBySearchPolicy` — same signature as on #2800 and #2804, fixed by #2801 (fully green, including this shard). Everything this PR touches passes: the propose/grade/calibration/brainstorm suites, `cycle.serial`, and the new tier-routing regression all sit in passing shards. Merging #2801 first de-flakes the re-runs here.

p3ob7o:
Rebased onto current master and restacked on #2804. Two adaptations to conventions master shipped in v0.42.62 while this PR was open:

1. **Full provider-prefixed ids everywhere for propose_takes and calibration_profile.** The original PR stored bare-tail model ids for label continuity; 0.42.62 moved these phases to full ids (and #2451 requires it for calibration — a bare id fed back into gateway.chat() throws). The tier-resolved full string now drives the chat call, the budget estimate, AND the stored model_id. The 0.42.62 gateway-chat-model and nested-provider-id regression tests pass unchanged.
2. **Unconfigured behavior preserved per phase.** propose_takes resolves explicit override > models.propose_takes > models.tier.reasoning > the gateway chat model (so brains without tier config keep exactly the 0.42.62 behavior); calibration_profile keeps its TIER_DEFAULTS.reasoning default.

grade_takes intentionally keeps the bare-tail judge_model_id: it keys the verdict idempotency cache and evidence signatures, so a prefix-only change must not re-judge every stock install (a genuinely different tier-configured judge still invalidates, which is correct).

Locally green: typecheck clean, 189 tests across the propose/grade/calibration/brainstorm/models suites — including master's three new model-convention tests.

Masashi-Ono0611:
Field data from a downstream fork ([Masashi-Ono0611/gbrain](https://github.com/Masashi-Ono0611/gbrain)) supporting this PR's premise, from a brain running with the tier system fully configured (`models.tier.reasoning`/`subagent` → `anthropic:claude-sonnet-5`, `models.tier.deep` → `anthropic:claude-opus-4-8`):

- Every route that resolves through `resolveModel` honored the tier config in our nightly dream cycles — `gbrain models` shows all per-task routes on the new generation.
- But that routing table is exactly what makes the remaining hardcodes invisible from the operator's seat: `grade_takes` (`judgeModelId = opts.model ?? 'claude-sonnet-4-6'`, grade-takes.ts:398 on current master) and the brainstorm orchestrator fallback (orchestrator.ts:507) don't appear in the routing display at all, so the console reads "fully migrated" while the cycle's grader can still run the previous-generation model. We only caught it by grepping the audit trail against the Anthropic billing console.
- Anecdotally the same visibility trap bit us on our own side first: a stale explicit `models.dream.synthesize` pin (set months ago, outranking the tier key by design) kept nightly synthesize on the old model unnoticed for weeks. Per-task config keys that *show up in `gbrain models`* — which this PR adds for the four sites — are what makes that class of drift visible and fixable.

Hope the #2804 restack lands soon — happy to re-verify on a tier-configured brain once it does.

time-attack:
The tier-config feature was declined for now with the cycle-knob set — but the budget-metering-records-wrong-model bug this PR also fixes is real and now tracked separately so it isn't lost. Thank you.

PR #2816, opened 2026-07-13
orphan_ratio: config-driven orphan_deny_prefixes to exclude record-type pages

stevotcp:
## Problem

On a record-heavy brain, `orphan_ratio` structurally FAILs and can't be fixed by linking. The "linkable" denominator counts every live page — including large auto-generated *record* areas (shopping orders, imported issue archives, email history, per-day dashboards) that never receive inbound links by design. On one brain that was ~1,500 of ~1,900 pages, pinning `orphan_ratio` near 90% (FAIL) no matter how well the curated graph was linked.

The built-in `DENY_PREFIXES` (`output/`, `dashboards/`, `scripts/`, `templates/`, `openclaw/config/`) covers common cases, but it can't know a given brain's record taxonomy — and hardcoding brain-specific prefixes into the shared tool isn't right.

## Change

Adds an optional `orphan_deny_prefixes: string[]` to `GBrainConfig`, read from the file plane in `findOrphans`/`shouldExclude` and applied on top of the built-in `DENY_PREFIXES`. Brains declare their own record areas without polluting the shared defaults:

```json
{ "orphan_deny_prefixes": ["personal/shopping/", "archive/issues/"] }
```

- **Backward-compatible**: optional param, empty = prior behavior. Existing `shouldExclude(slug)` calls are unchanged.
- Applied to **both** the numerator (orphan filter) and the denominator (`total_linkable`) so the ratio stays consistent.
- Mirrors the existing `GBRAIN_AUDIT_ARCHIVE_ROOTS` override pattern.
- Best-effort config read: missing/invalid config → no extra prefixes, never breaks a scan.

## Tests

`bun test test/orphans-pure-fn.test.ts test/orphans.test.ts` → **50 pass / 0 fail**.

## Possible follow-up

A cleaner long-term shape might be **page-type-driven** exclusion (let the schema pack mark record types like `shopping-order` / `email-history` as no-inbound-expected) rather than slug prefixes. Submitting the prefix version first since it's minimal and matches the existing prefix-based `DENY_PREFIXES` / `GBRAIN_AUDIT_ARCHIVE_ROOTS` design — happy to rework it type-driven if you'd prefer.

### DIFF
--- a/src/commands/orphans.ts
+++ b/src/commands/orphans.ts
@@ -15,6 +15,7 @@
 import type { BrainEngine } from '../core/engine.ts';
 import { createProgress, startHeartbeat } from '../core/progress.ts';
 import { getCliOptions, cliOptsToProgressOptions } from '../core/cli-options.ts';
+import { loadConfigFileOnly } from '../core/config.ts';
 
 // --- Types ---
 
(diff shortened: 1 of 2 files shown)


time-attack:
Thanks for the clear problem statement and for being upfront that the prefix approach isn't the shape you'd actually want. That last point is why I'm closing it: `orphan_ratio` isn't computing anything wrong here — the brain's page mix is unusual — and adding a permanent config key so one brain can declare its own slug taxonomy is exactly the kind of surface we're trying not to accumulate, especially when you've already identified page-type-driven exclusion as the better design. Closing on config-surface grounds rather than on quality; the implementation itself is clean.

PR #2817, opened 2026-07-13
fix: categorize timeline_coverage and register auto_chronicle

davidfragnoli-lang:
Summary:
- categorize `timeline_coverage` as a first-class brain/data completeness check instead of falling back to `meta`
- register `auto_chronicle` in `KNOWN_CONFIG_KEYS`
- add the `chronicle.` prefix so `chronicle.tz` is recognized without `--force`

Why:
- fixes the nightly `[doctor-categories] unknown check name 'timeline_coverage' — defaulting to 'meta'` warning
- makes the documented `gbrain config set auto_chronicle true` flow actually work

Verification:
- `bun test test/doctor-categories.test.ts`
- `bun run src/cli.ts config set auto_chronicle true`

Fragno tracking: FRA-1319

### DIFF
--- a/src/core/config.ts
+++ b/src/core/config.ts
@@ -926,6 +926,11 @@ export const KNOWN_CONFIG_KEYS: readonly string[] = [
   'embed.backfill_cooldown_min',
   'embed.backfill_max_usd_per_source_24h',
   'embed.backfill_max_usd',
+  // Life Chronicle (v0.42.56, #2390). Auto-emission of timeline events is OFF
+  // by default; the changelog's "To take advantage" step is `gbrain config set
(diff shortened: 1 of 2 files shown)


time-attack:
Thanks for catching these — they were real, and both have since landed on master. `timeline_coverage` is now categorized in `doctor-categories.ts` via #3075 (commit d6fe4863), and `auto_chronicle` plus the `chronicle.` prefix are registered in `KNOWN_CONFIG_KEYS` via #2632 (commit 285cf39f), so the nightly warning and the `gbrain config set auto_chronicle true` flow are both fixed. Closing as already fixed.

PR #2818, opened 2026-07-13
perf: batch loadConfigWithEngine DB reads, fix sources-list N+1

Kaynetiq:
## Summary
- `loadConfigWithEngine()` read 20 config keys via 20 serial `getConfig()` round trips on every cold `connectEngine()` call (measured ~150-1200ms/call on a first-connection pool, several seconds total). Adds optional `getConfigBatch()` to the `BrainEngine` interface, implements it on `PostgresEngine` (single `key = ANY($1::text[])` query, reusing the existing retry/reconnect wrapper), and has `loadConfigWithEngine()` prefer it — falling back to a parallel per-key fetch for engines that don't implement it (e.g. PGLite).
- Second N+1: `listSources()` in `sources-ops.ts` ran one `SELECT COUNT(*) FROM pages WHERE source_id = $1` per source instead of a single `GROUP BY`. Replaced with `countPagesBatch()`.
- Updated the stale `loadConfigWithEngine` doc comment — it said only the v0.27.1 multimodal flags participate in DB-merge; actually 20 keys across three feature groups (embeddings, content_sanity.*, dream.*) merge today.
- Added tests pinning `getConfigBatch()` usage: batch path preferred over per-key `getConfig`, fallback when absent, non-fatal on throw.

## Test plan
- [x] `bun test test/loadConfig-merge.test.ts` — 24/24 pass
- [x] `bun test test/sources-ops.test.ts test/sources-mcp.test.ts` — 59/59 pass
- [x] `bun tsc --noEmit -p .` — clean

### DIFF
--- a/src/core/config.ts
+++ b/src/core/config.ts
@@ -603,12 +603,17 @@ export function loadConfig(): GBrainConfig | null {
  * Precedence: env > file > DB > defaults. Env stays the operator escape hatch;
  * file is the durable per-machine config; DB is the user-mutable runtime knob.
  *
- * Today only the v0.27.1 multimodal flags participate in DB-merge. Existing
- * fields (embedding_model, etc.) keep their file/env-only loading because they
- * size the schema and must be stable across engine connect.
+ * DB-merge covers embedding/multimodal flags, embedding_columns +
+ * search_embedding_column, content_sanity.*, and dream.synthesize/patterns
+ * keys (see ALL_DB_KEYS below). Existing fields (embedding_model, etc.) keep
+ * their file/env-only loading because they size the schema and must be
+ * stable across engine connect.
  */
 export async function loadConfigWithEngine(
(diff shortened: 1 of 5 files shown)


time-attack:
Thanks — the measurement is convincing and the 20-serial-round-trips-per-cold-connect pattern is genuinely wasteful on a remote pooler. I'm closing it anyway on two grounds: this is a perf optimization rather than a correctness fix, and it adds `getConfigBatch()` to the `BrainEngine` interface while implementing it only on `PostgresEngine` (the diff never touches `pglite-engine.ts`). Our engine-parity invariant requires both engines to move in lockstep, so as written this permanently adds interface surface plus a dual batch/fallback code path that gives nothing to PGLite, our zero-config default. The `listSources()` N+1 is the more clearly-worth-it half if you ever want to revisit it standalone.

PR #2824, opened 2026-07-14
Add correlated query profiling and safe search setup optimizations

FroeMic:
## What changed

- Adds production-compatible correlated OpenTelemetry spans for GBrain query/search stages.
- Adds bounded connection-acquisition timing without logging URLs, credentials, query text, or user data.
- Batches search-mode configuration reads with per-key fallback.
- Reuses request-local mode/embedding configuration across cached misses and removes the redundant cache-config read.
- Traces actual serialization, keyword-only configuration, and hot-memory work with safe input/output summaries.
- Adds focused parity, call-count, profiling, redaction, and fail-open coverage.

## Validation

- Focused search/HTTP tests passed.
- `bun run verify` passed: 31/31 checks.
- `bun run typecheck` passed.
- `bun run build:all` passed.
- Disposable Monodrive persistent-HTTP profile pilot passed parity, ordering, limit, expansion-disabled, not-found, and source-isolation gates.
- Directional cold-latency improvement versus the prior candidate was approximately 8% for query and 17% for search; warm search remained approximately unchanged.


FroeMic:
Conflict resolution pushed in merge commit `5c131dc67dc88a9ddea88ae7a77283546a71122`, preserving GBrain version `0.42.62.0`. The PR is now mergeable. Post-resolution validation passed: `bun run verify` (31/31), `bun run typecheck`, and `bun run build:all`. The current FroeMic token cannot execute the upstream merge (`MergePullRequest` permission denied), so a garrytan/gbrain maintainer should merge this PR.

time-attack:
Thanks for the careful work here — the redaction discipline and fail-open handling are clearly thought through. We're closing this one: it adds seven new @opentelemetry runtime dependencies and ~1100 lines of new profiling surface on the hot search path, and tracing/observability is not a gap we're taking on right now. The batched config reads inside that change are a reasonable idea, but they'd need to arrive as a small standalone diff with no new dependencies to be considered.

PR #2832, opened 2026-07-14
feat(ai): add MiniMax M3 chat support

octo-patch:
Reason: add target provider/model to existing provider registry.

## Summary
- register `MiniMax-M3` and `MiniMax-M2.7` for chat while preserving the existing embedding model
- add a second compatible messages transport with exact global and China public base URLs and internal SDK path derivation
- document regional configuration and add registry, authentication, and request-capture coverage

## Checks
- locked dependency installation passed
- focused registry, transport, gateway, regression, and provider-list tests passed (61 tests)
- static type validation passed
- full repository verification passed (31 checks)

### DIFF
--- a/docs/ai-providers/minimax.md
+++ b/docs/ai-providers/minimax.md
@@ -1,0 +1,44 @@
+# MiniMax Chat Models
+
+GBrain exposes `MiniMax-M3` and `MiniMax-M2.7` through two compatible chat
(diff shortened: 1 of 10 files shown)


time-attack:
Reviewed and declined with the vendor-catalog set (also overlaps #2220). Thank you.

PR #2842, opened 2026-07-14
feat(ai): add first-class Amazon Bedrock recipe (via LiteLLM proxy)

davidtwu:
### What

Adds a dedicated **`bedrock`** provider recipe so gbrain can run entirely on
[Amazon Bedrock](https://aws.amazon.com/bedrock/) — Claude for synthesis, Cohere
Embed v4 for embeddings — through a local [LiteLLM](https://docs.litellm.ai)
proxy, with no code changes and no `--embedding-dimensions` flag:

```bash
gbrain init \
  --embedding-model bedrock:us.cohere.embed-v4:0 \
  --chat-model bedrock:us.anthropic.claude-opus-4-8
```

### Why a dedicated recipe (vs. the generic `litellm` one)

The generic `litellm` recipe is `user_provided_models` with `default_dims: 0`,
so it forces the user to pass `--embedding-dimensions N`. But
`isCustomDimValidForProvider` (`embedding-dim-check.ts`) then **rejects** any
custom dim for `litellm` because it isn't in a Matryoshka allow-list — so the
generic litellm embedding path can't be initialized at all.

The `bedrock` recipe closes that gap by declaring Cohere Embed v4 with an
explicit `default_dims: 1536` + Matryoshka `dims_options`, so `gbrain init`
needs no dimensions flag and custom dims pass validation via the recipe-declared
`dims_options` (Tier 1 of the checker).

### Design notes

- **Transport:** `tier: 'openai-compat'`, `implementation: 'openai-compatible'`.
  Bedrock's runtime API is AWS-SigV4-signed and speaks neither the OpenAI nor
  the Anthropic wire shape; LiteLLM normalizes both chat and embeddings to
  `/v1/chat/completions` and `/v1/embeddings`.
- **Model ids** are Bedrock **inference-profile** ids (the `us.`/`global.`
  prefix). Raw on-demand ids fail with a `ValidationException` on newer models.
- **Auth:** none on the gbrain side — AWS credentials live on the proxy's boto3
  chain. gbrain sends `Authorization: Bearer unauthenticated`, which a loopback
  LiteLLM with no `master_key` ignores (mirrors the ollama / llama-server
  no-auth pattern). `BEDROCK_PROXY_API_KEY` is honored if you set a master key.
- **Prompt caching advertised as `false`:** although these are Claude models,
  chat routes through the OpenAI-compatible provider, which drops the gateway's
  Anthropic-only `providerOptions.anthropic.cacheControl` markers. Caching would
  be a silent no-op, so the recipe declares `false` like the other openai-compat
  recipes — keeping the gateway's "only Anthropic claims prompt cache" invariant
  test true. (An earlier draft set this `true`; corrected here.)

### Files

| File | Change |
|------|--------|
| `src/core/ai/recipes/bedrock.ts` | new recipe |
| `src/core/ai/recipes/index.ts` | register `bedrock` in picker + `providers list` |
| `src/core/ai/recipes/litellm-proxy.ts` | doc/setup-hint touch-ups |
| `docs/guides/litellm-proxy.md` | new — full Bedrock setup walkthrough |
| `docs/integrations/embedding-providers.md` | Bedrock row in the matrix |
| `CHANGELOG.md` | `[Unreleased]` entry |
| `test/ai/recipe-bedrock.test.ts` | new — 7 recipe tests |

### Testing

- `bun test test/ai/` → **322 pass / 0 fail** (includes the 7 new recipe tests
  and the existing gateway prompt-cache invariant).
- `tsc --noEmit` → clean.
- `check:source-config-leak` + `check:no-pii-agent-voice` → clean.
- Verified live against a Bedrock account in **us-west-2**: Opus 4.8 chat +
  Cohere Embed v4 (1536-dim) both invoke through LiteLLM 1.91.0.


time-attack:
Reviewed and declined for now with the vendor-catalog set — Bedrock is the strongest case here and may be revisited; the LiteLLM proxy path reaches it today. Thank you.

PR #2847, opened 2026-07-14
fix(chunker): estimated-token hard cap — URL-dense/CJK-fallback chunks overflow strict embedding-server token limits

### LINKED ISSUE #2826: Chunker emits chunks past strict embedding-server token limits on URL-dense CJK docs (whitespace-word undercount) and un-subdividable AST nodes
**What happened?**

On URL/phone/email-dense CJK markdown (and large JSON code blocks), the recursive chunker emits 3–4K-char chunks that exceed strict per-request embedding-server token limits. With a local llama-server embedding backend started with `-ub 2048` (physical batch 2048), any single chunk past ~2,050 real tokens **crashes the server** (trace/BPT trap; the client sees EOF). The same chunk fails on every retry — a deterministic import/embed failure for affected pages.

Measured with the real Qwen3 embedding tokenizer on synthesized inputs (worst chunk per document):

| input | worst chunk (current chunker) |
|---|---|
| Korean rollup, 60 lines each ending in a long source URL | **2,889 tokens** (4,524 chars) |
| 7K-char minified JSON | **2,136 tokens** (4,629 chars) |
| 7K-char pretty-printed JSON | 1,967 tokens (4,287 chars — over the limit once the contextual-retrieval wrapper is added) |

**What did you expect?**

No emitted chunk should exceed a bounded token size, regardless of content shape — oversized chunks make embedding fail (or crash strict local servers) with no recovery path.

**Root cause (two spots)**

1. `countCJKAwareWords` (src/core/cjk.ts): when CJK char density falls below 0.30, word counting falls back to whitespace tokens — and a 150-char URL counts as **one word**. Documents that are semantically CJK but carry per-line URLs get pushed below the density threshold *by the URLs' own ASCII mass*, then sized at a fraction of their real bulk. `greedyMerge` packs up to 450 such "words" → 3–4K-char chunks. The existing `maxChars: 6000` cap (recursive.ts) was calibrated for OpenAI's 8,192-token limit assuming ≥1 char/token; URL-dense text tokenizes at ~1.6 chars/token, so the cap passes chunks worth up to ~3,700 real tokens.
2. Code chunker AST path (src/core/chunkers/code.ts): a node `splitLargeNode` cannot subdivide (giant single-statement function, huge literal — fewer than 2 body children) ships **whole at any size**.

**Steps to reproduce**

1. Synthesize a Korean markdown doc with 40+ lines, each ending in a long URL (a common shape for docs that cite one source link per line):

```ts
import { chunkText } from './src/core/chunkers/recursive.ts';

const lines = ['# rollup', ''];
for (let i = 0; i < 60; i++) {
  const hex32 = ((i * 2654435761) >>> 0).toString(16).padStart(8, '0').repeat(4);
  lines.push(`- **항목 ${i}**: 이 줄은 청커 동작 검증을 위한 의미 없는 한국어 예시 문장입니다 · 전화 000-0000-${1000 + i} · 이메일 user${i}@example.com · 링크: https://docs.example.com/pages/${hex32}?v=abcdef0123456789&ref=sample`);
}
const chunks = chunkText(lines.join('\n'));
console.log(Math.max(...chunks.map(c => c.text.length))); // 4,500+ chars
```

2. Tokenize the largest chunk with any Qwen3-family tokenizer (e.g. llama-server `POST /tokenize`) → 2,800+ tokens.
3. Embed through llama-server running with `-ub 2048` → server crashes (trace/BPT trap), client gets EOF. Retries fail identically.

The pure-chunker part (steps 1–2) reproduces with no DB or embedding backend at all.

**Environment**

- gbrain version: 0.42.59.0
- OS: macOS (Darwin 25.5, Apple Silicon)
- Bun version: 1.3.14
- Database: self-hosted Postgres (irrelevant — the defect is in the engine-independent text pipeline; reproduces with `chunkText()` alone)
- Embedding backend: ollama-managed llama-server, Qwen3-Embedding-8B, `-ub 2048`

**Fix**

PR attached (estimated-token hard cap + whitespace-word undercount floor + AST-path cap; conservative tokenizer-free estimator so the cap is safe against real tokenizers). After the fix, the same inputs measure 1,336 / 996 / 994 tokens worst-chunk — under the limit with margin, and the estimator over-estimates in every measured case.

paul-0320:
Fixes #2826

## Summary

URL/phone/email-dense CJK markdown and large JSON code blocks produce 3–4K-char chunks that overflow strict per-request embedding-server token limits (measured: a local llama-server with `-ub 2048` crashes past ~2,050 tokens per chunk — deterministic EOF for the affected pages; full reproduction and measurements in the linked issue).

Two defects let those chunks form, fixed with defense in depth:

1. **Whitespace-word undercount** — `countCJKAwareWords`' fallback (CJK density < 0.30) counts a 150-char URL as one word, and per-line URLs are exactly what pushes CJK docs below the density threshold. The chunker sizes such docs at a fraction of their bulk; the existing `maxChars: 6000` cap assumes ≥1 char/token and passes URL-dense chunks worth up to ~3,700 real tokens (~1.6 chars/token).
2. **Unbounded AST-path chunks** — a node `splitLargeNode` can't subdivide ships whole at any size.

## Changes

- **`src/core/cjk.ts`** — add `estimateEmbeddingTokens()`: tokenizer-free, per-char-class token estimate (CJK 1.0/char, other non-whitespace 0.75/char, whitespace 0.1/char). Weights are calibrated against measured ratios (URL-dense ~1.6 chars/token, base64-worst ~1.3) and deliberately **overestimate**, so any cap based on the estimate is safe against real tokenizers. Additive only — `countCJKAwareWords` semantics are untouched (search/`expansion.ts`, which imports it, is unaffected).
- **`src/core/chunkers/recursive.ts`** (`MARKDOWN_CHUNKER_VERSION` 3→4):
  - `countWords` floored at `ceil(nonWhitespaceChars / 6)` — long whitespace-less runs (URLs, minified blobs) count roughly per-character. Latin prose is unchanged (average English word < 6 chars, so the whitespace count still wins); pinned by tests.
  - `capByEstimatedTokens()` final pass with new `ChunkOptions.maxTokens` (default 1500 — headroom for the contextual-retrieval wrapper under a ~2,050-token server limit, and far below OpenAI's 8,192 so it is a pure safety win there too; normal 300-word chunks never approach it). Cut points prefer a newline, then whitespace, within a 300-char lookback so forced splits land at line boundaries instead of mid-URL/mid-line.
- **`src/core/chunkers/code.ts`** (`CHUNKER_VERSION` 4→5): `capCodeChunks()` applies the same cap to AST-path chunks (the `splitLargeNode`-returns-empty case previously shipped unbounded).

Version bumps route existing pages through the standard `gbrain reindex --markdown` sweep / `sources.chunker_version` sync gate.

## Measurements (real Qwen3 embedding tokenizer, worst chunk, same inputs)

| input | before | after |
|---|---|---|
| URL-dense Korean rollup (60 lines) | **2,889 tok** | 1,336 tok |
| 7K-char minified JSON | **2,136 tok** | 996 tok |
| 7K-char pretty JSON | 1,967 tok | 994 tok |

The estimator stayed above the real count in every failure-shape case (1,483–1,498 estimated vs 994–1,336 real). One boundary observation from a production-scale sweep (1,454 pages re-chunked, zero embed failures): on whitespace-heavy aligned-table content the estimate can undershoot the real count slightly (worst measured −2.4%: 1,277 est vs 1,308 real on a 4.2K-char chunk that is 72% spaces) — the 550-token headroom between the default cap (1,500) and the measured server limit (~2,050) absorbs this class comfortably.

## Tests

`test/chunkers/token-cap.test.ts` (new, 13 tests): reproduces both failure shapes (URL-dense CJK rollup, large JSON pretty/minified through both the prose path and the `fence.json` code-chunker path), and pins:

- the cap invariant (`estimateEmbeddingTokens(chunk) ≤ maxTokens` for every emitted chunk),
- content preservation (no lines dropped),
- newline-preferring cut placement and forward progress on whitespace-less input,
- **unchanged behavior for normal content** — Latin prose chunk boundaries and CJK-dense prose sizing.

Existing chunker suites updated only for the version-constant assertions (3→4, 4→5).

## Verification

- `bun run verify`: all 31 checks pass (typecheck, `check:test-isolation` R1–R4, privacy checks, etc.). Note: run sequentially per check — the parallel dispatcher is incompatible with my sandboxed shell; every check binary-passes individually.
- Unit suite (`bun test`): full run on this branch vs a master baseline in the same environment — **zero new failures from this change**; chunker/CJK/reindex/re-embed domain suites 100% green (12,515 tests ran; every failure in the delta vs baseline is a load-sensitive timing/subprocess test that passes when its file runs solo — verified; the ~630 remaining failures are environment-dependent on my machine, identical before and after this change).
- `bun run test:slow`: 41/41 pass.
- E2E: skipped locally (no `DATABASE_URL`), per the `test:full` script's documented skip path. The change is engine-agnostic — pure text pipeline in `src/core/chunkers/` + `src/core/cjk.ts`; no engine SQL, no persistence-layer changes (chunk rows flow through both engines' existing `upsertChunks` unchanged).
- **eval replay: not triggered** — the diff touches none of the CONTRIBUTING-listed retrieval trigger paths (`src/core/search/*`, `embedding.ts`, query/search op handlers, engine searchKeyword/searchVector). `cjk.ts` changes are additive exports only; `expansion.ts` behavior is bit-identical. Chunk boundaries change **only** for pathological content that previously produced un-embeddable chunks (i.e., content that currently has no vectors at all on strict backends), plus the >6000-char cap survivors now split at 1500 estimated tokens; normal prose boundaries are pinned unchanged by the test suite.

## Rebase note (2026-07-24) — relation to #1675

Rebased over current master, which now carries #1675's `capOversizedChunks` on the AST path. The two caps are complementary, not redundant, because they measure with different rulers:

- **#1675** splits structurally (recursive re-split, header rebuild, `opts.maxChunkTokens`) but **triggers on the generic `estimateTokens` (~3.5 chars/token)**, which undercounts CJK text (~1 char/token measured) and URL-dense runs (~1.6 chars/token) by 2–3.5×. An over-budget Korean chunk passes its trigger untouched — exactly the failure shape in the table above.
- **This PR's `capCodeChunks`** re-checks with the CJK-aware `estimateEmbeddingTokens` as a final pass, and is a no-op when every chunk is already under the estimated cap.

The AST path now composes them: `capCodeChunks(capOversizedChunks(...))` — #1675's structural split first, CJK-aware safety net last. The markdown path (`recursive.ts`) is untouched by #1675 and still relies solely on this PR's cap. All 73 chunker tests (including the 13 new ones) pass on the rebase; `check:test-names` and typecheck clean.

### DIFF
--- a/src/core/chunkers/code.ts
+++ b/src/core/chunkers/code.ts
@@ -18,8 +18,9 @@
  * at runtime.
  */
 
-import { chunkText as recursiveChunk } from './recursive.ts';
+import { chunkText as recursiveChunk, capByEstimatedTokens, DEFAULT_MAX_EST_TOKENS } from './recursive.ts';
 import { buildQualifiedName } from './qualified-names.ts';
+import { estimateEmbeddingTokens } from '../cjk.ts';
 
 // Embed the tree-sitter runtime + per-language grammars as files.
 // `with { type: 'file' }` returns a path (string) at runtime. Bun bundles
@@ -111,7 +112,15 @@ import G_ZIG from '../../assets/wasm/grammars/tree-sitter-zig.wasm' with { type:
 // chunks get the new columns populated. Without this, the v28 backfill
 // gives every existing chunk a search_vector but subsequent Layer 5 AST
 // work would silently no-op.
-export const CHUNKER_VERSION = 4;
+//
+// v5: estimated-token hard cap on AST-path chunks (capCodeChunks). A node
+// splitLargeNode can't subdivide (giant single-statement function, huge
+// literal) previously shipped WHOLE regardless of size and could overflow
+// strict per-request embedding-token limits (local llama-server crashes
+// past ~2,050 tokens, measured). Mirrors the markdown
+// chunker's v4 cap; fallback-path chunks are already capped inside
+// recursiveChunk.
+export const CHUNKER_VERSION = 5;
 
 // Lazy-loaded tree-sitter module (v0.22.x API: Parser is default export)
 let Parser: typeof import('web-tree-sitter') | null = null;
@@ -718,7 +727,13 @@ export async function chunkCodeTextFull(
     if (chunks.length === 0) {
       return { chunks: capOversizedChunks(fallbackChunks(source, filePath, language, opts), filePath, language, opts), edges: rawEdges };
     }
-    return { chunks: capOversizedChunks(mergeSmallSiblings(chunks, chunkTarget), filePath, language, opts), edges: rawEdges };
+    // Layered with #1675's capOversizedChunks: that pass splits structurally
+    // but triggers on the generic estimateTokens (~3.5 chars/token), which
+    // undercounts CJK (~1 char/token) and URL-dense text — an over-budget
+    // Korean chunk passes it untouched. capCodeChunks re-checks with the
+    // CJK-aware estimateEmbeddingTokens; it is a no-op when all chunks are
+    // already under the estimated cap.
+    return { chunks: capCodeChunks(capOversizedChunks(mergeSmallSiblings(chunks, chunkTarget), filePath, language, opts)), edges: rawEdges };
   } catch {
     return { chunks: fallbackChunks(source, filePath, language, opts), edges: [] };
   } finally {
@@ -799,6 +814,33 @@ function mergeSmallSiblings(chunks: CodeChunk[], chunkTarget: number): CodeChunk
     i = j;
   }
   return merged;
+}
+
+/**
+ * v5 final safety pass for AST-path chunks: split any chunk whose
+ * ESTIMATED embedding tokens (conservative per-char-class heuristic,
+ * cjk.ts) exceed DEFAULT_MAX_EST_TOKENS. Reaches chunks the AST logic
+ * can't subdivide — splitLargeNode returns [] for nodes with < 2 body
+ * children (giant single-statement functions, huge literals), which
+ * previously shipped whole at any size.
+ *
+ * Split pieces inherit the source chunk's metadata verbatim; start/end
+ * lines become approximate for pieces after the first. Acceptable —
+ * these chunks exist for embedding + retrieval, and the alternative was
+ * an embedding request the server rejects (or worse, crashes on).
+ */
+function capCodeChunks(chunks: CodeChunk[]): CodeChunk[] {
+  if (chunks.every((c) => estimateEmbeddingTokens(c.text) <= DEFAULT_MAX_EST_TOKENS)) {
+    return chunks;
+  }
+  const out: CodeChunk[] = [];
+  for (const c of chunks) {
+    const pieces = capByEstimatedTokens(c.text, DEFAULT_MAX_EST_TOKENS);
+    for (const piece of pieces) {
+      out.push({ ...c, text: piece, index: out.length, metadata: { ...c.metadata } });
+    }
+  }
+  return out;
 }
 
 function buildMergedChunk(group: CodeChunk[], index: number): CodeChunk {
--- a/src/core/chunkers/recursive.ts
+++ b/src/core/chunkers/recursive.ts
@@ -17,7 +17,13 @@
  * Lossless invariant: non-overlapping portions reassemble to original.
  */
 
-import { countCJKAwareWords, CJK_SENTENCE_DELIMITERS, CJK_CLAUSE_DELIMITERS } from '../cjk.ts';
+import {
+  countCJKAwareWords,
+  CJK_SENTENCE_DELIMITERS,
+  CJK_CLAUSE_DELIMITERS,
+  charEmbedTokenWeight,
+  estimateEmbeddingTokens,
+} from '../cjk.ts';
 
 /**
  * Markdown chunker version. Folded into the per-page chunker_version column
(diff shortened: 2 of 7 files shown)


stigrunar:
Additional production-shape validation for this content-preserving cap:

A streaming, read-only replay of `capByEstimatedTokens(..., 1500)` over an existing multi-source GBrain corpus found:

- input rows: `157,823`
- rows over the cap before splitting: `15,972`
- resulting pieces: `181,165`
- pieces over the cap afterward: `0`
- maximum afterward: `1,500` estimated tokens / `5,992` UTF-8 bytes

This is large enough that a gateway-only truncation workaround would silently embed incomplete content for a meaningful fraction of the corpus. Splitting in the versioned chunker is the correct layer.

The replay also found one lossless-invariant edge in the current PR diff. With:

```ts
const slice = text.slice(start, end).trim();
```

`15,882` rows did not reassemble byte-for-byte, although all non-whitespace content survived. Changing it to:

```ts
const slice = text.slice(start, end);
if (slice.length > 0) out.push(slice);
```

and pinning `expect(pieces.join('')).toBe(text)` reduced both exact and non-whitespace mismatches to `0` across all `157,823` rows. The exact fix and test are visible in [`stigrunar/gbrain@8e91de7a`](https://github.com/stigrunar/gbrain/commit/8e91de7a649a28a1dc6ce25b55e8d4202d2ea0bc). The focused combined Qwen/chunker gate remained `75 pass / 0 fail`, and the repo verify gate remained `31 / 31` green.

Independent local Ollama evidence is also consistent with the issue's physical-batch diagnosis: `qwen3-embedding:0.6b` passed 2,000 synthetic CJK characters, while 2,500 and 3,000-character inputs produced runner EOF. The chunk cap and the Ollama `min(num_ctx, num_batch)` fix remain useful defense in depth.

time-attack:
Thanks — the analysis is genuinely good, and you're right that the char cap can't bound token-dense content. We're closing it because of blast radius: bumping both chunker versions forces a full re-chunk and re-embed for every install, and a global 1500 estimated-token default meaningfully shrinks CJK and dense chunks for everyone, to accommodate a self-hosted embedding server's 2048-token batch setting (the default providers at 8192 can't overflow a 6000-char chunk). A provider-aware or config-driven cap that leaves the default chunking untouched would be the shape we could consider.

PR #2848, opened 2026-07-15
feat(guardrails): add ./guardrails to public exports

bimaben:
## What

Adds the `./guardrails` subpath to the package.json exports map, so external guardrail providers can import the seam the way `docs/guardrails.md` already documents:

```ts
import { registerGuardrailProvider } from 'gbrain/guardrails';
```

Today the doc shows a package import, but the subpath is missing from the exports map, so out-of-tree providers can only deep-import `src/` paths.

## Changes

- `package.json`: add `"./guardrails": "./src/core/guardrails.ts"` (exports surface 20 → 21)
- `scripts/check-exports-count.sh`: bump `EXPECTED_COUNT` to 21
- `test/public-exports.test.ts`: bump the count assertion and pin canary symbols (`registerGuardrailProvider`, `runGuardrails`, `hasGuardrails`)
- `docs/guardrails.md`: align the example import path with the real subpath form

No behavior change to the seam itself — additive public-API surface only, per the exports policy in CLAUDE.md.

## Motivation

We're building an out-of-tree, shadow-mode PHI/PII detection provider on the guardrail seams (the exact use case the seam doc describes). A clean package import keeps providers off gbrain's internal file layout.

## Testing

- `bun test test/guardrails.test.ts test/public-exports.test.ts` — 52 pass / 0 fail
- `bash scripts/check-exports-count.sh` — 21 entries, matches baseline
- `bun run verify` — all 31 checks green
- `tsc --noEmit` — clean

### DIFF
--- a/docs/guardrails.md
+++ b/docs/guardrails.md
@@ -52,7 +52,7 @@ import APIs.
 ## Writing a guardrail provider
 
 ```ts
-import { registerGuardrailProvider, type GuardrailInput } from 'gbrain/core/guardrails';
+import { registerGuardrailProvider, type GuardrailInput } from 'gbrain/guardrails';
 
 registerGuardrailProvider({
   id: 'my-firewall',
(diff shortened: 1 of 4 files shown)


time-attack:
Reviewed and declined for now with the library-consumer set. Thank you.

PR #2851, opened 2026-07-15
fix(probe): gate nightly quality probe on real run outcomes only

sanchalr:
Running the nightly probe under a 5-min autopilot tick, one rate-limit burst wedged the probe permanently: every blocked tick writes a rate_limited audit row, and the 24h gate counted those rows as activity - so the gate never reopened, and probe-health filled with ~288 identical rows/day.

- Gate now counts only real run outcomes (pass / fail / inconclusive / error / budget_exceeded).
- rate_limited is logged once per window, not per tick (latest-row guard).
- Equal-timestamp tie-break prefers the later row so the guard is stable at ms resolution.

Two regression tests: a skip-only history must not block the run; repeated blocked ticks write exactly one rate_limited row. `bun test test/nightly-quality-probe.test.ts` - 27 pass. Typecheck clean.

Hit this on a daily-driver install; the fix has been running there since mid-June.

@sanchalr 🫡

### DIFF
--- a/src/core/cycle/nightly-quality-probe.ts
+++ b/src/core/cycle/nightly-quality-probe.ts
@@ -23,7 +23,7 @@ import * as fs from 'node:fs';
 import * as path from 'node:path';
 import { tmpdir } from 'node:os';
 
-import { logQualityProbeEvent, readRecentQualityProbeEvents } from '../audit-quality-probe.ts';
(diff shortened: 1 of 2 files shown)


time-attack:
Thank you for the report and the reproduction — this was a real wedge. It has since been fixed on master by #3094 (commit b91350d7), which takes the simpler route of not writing an audit row at all on the 24h skip, so skip rows can no longer refresh their own gate and the probe-health flood is gone too. Closing as already fixed; please do re-open a report if you still see the probe wedged on current master.

PR #2853, opened 2026-07-15
fix(budget): price non-Anthropic chat models via canonical fallback

sanchalr:
With a budget cap set, any chat call routed to a non-Anthropic model failed before the provider was called: `lookupPricing` consulted only ANTHROPIC_PRICING, found nothing, and `reserve()` threw `no_pricing`. CANONICAL_PRICING already carries these ids - fall back to it for chat/rerank ids the Anthropic view omits.

Two regression tests: `openai:gpt-5.5` reserves under a cap; a genuinely-unknown id still throws. 44 pass across the budget/pricing suites. Typecheck clean.

@sanchalr 🫡


time-attack:
Closing as duplicate of #2127.

Same fix: price non-Anthropic chat models via the canonical pricing table. Keep #2127.

PR #2856, opened 2026-07-15
fix(schema-pack): merge extends chain + borrow_from into the resolved manifest (#1749)

### LINKED ISSUE #1749: Schema-pack extends/borrow_from never merges parent page_types — lens packs silently resolve to own-types-only
## Symptom

Activating any `extends`-based schema pack strips the brain's taxonomy to that pack's own (often tiny) declared types — the parent `gbrain-base` types are never inherited (v0.42.1.0):

```
gbrain-investor    (2 own + extends gbrain-base[25])   ->  Page types: 2    (expected 27)
gbrain-recommended (extends gbrain-base)               ->  Page types: 0
gbrain-everything  (extends investor + borrow lenses)  ->  Page types: 0
gbrain-base-v2     (extends: null, 15 inline)          ->  Page types: 15   OK
gbrain-base        (extends: null, 25 inline)          ->  Page types: 25   OK
```

Only `extends: null` packs work. Every lens/domain pack (`recommended`, `investor`, `creator`, `engineer`, `everything`) is structurally non-functional as a taxonomy: activating `investor` loses `person`/`company`/etc., and `gbrain schema explain person` returns *"not in active pack."* The packs *load* (identity hashes compute, phases declare) while their **types come back empty** — silently replacing the working base taxonomy.

## Root cause

`resolvePack` (`src/core/schema-pack/registry.ts`) walks the extends chain only to enforce the depth cap + build the cache snapshot, then constructs the resolved pack from the **child's own manifest verbatim**. The merge is an explicit, never-landed TODO:

> `registry.ts:278` — `// Full extends-merging (child-wins) is the v0.41+ T20 follow-up.`

The `while (cursor?.extends)` loop (`registry.ts:262`-`282`) loads each parent via `loadByName` but discards it — parents are used only to push names onto `chain`. No parent `page_types`/`link_types`/`takes_kinds`/`calibration_domains` are ever merged.
Permalink: https://github.com/garrytan/gbrain/blob/eefe8b5741c27e59bf65198d46e3dfe5bfa70ce9/src/core/schema-pack/registry.ts#L262-L282

1. **Implement the deferred merge in `resolvePack`.** Compose `page_types`, `link_types`, `takes_kinds`, `calibration_domains` base->child (child-wins on name), resolve `borrow_from` the same way, build `ResolvedPack` from the merged result. The core fix.
2. **`doctor` signal for an own-only active pack.** When the active pack declares `extends` but resolves to <= its own declared count, warn loudly — today a brain can silently run on a lens pack with zero inherited types and nothing surfaces it.
3. **Fail-loud on activation.** Reject/warn when a pack with `extends:`/`borrow_from:` resolves to a type set that excludes the parent's, so a broken lens pack can't be silently activated.

## Why this matters

The lens packs are a headline feature (`docs/architecture/lens-packs.md`; the `gbrain-everything` "YC-president pattern"). As shipped, activating any of them is *worse than nothing* — it silently swaps the working 15/25-type base taxonomy for an empty or 2-type one. This is the schema-layer instance of "imported != curated": the pack *looks* active (identity, phases) while the taxonomy it promises is absent, with no health signal.

## Repro

```
# investor = 2 own types + extends: gbrain-base (25). Set schema_pack: gbrain-investor in
# ~/.gbrain/config.json (it isn't in the `schema use` allow-list - see #1726), then:
$ gbrain schema active
Active pack: gbrain-investor v1.0.0
Page types: 2                       # expected 27
$ gbrain schema explain thesis      # own -> recognized
$ gbrain schema explain person      # from extends: gbrain-base
Type `person` not in active pack `gbrain-investor`.
```
```ts
// or directly, no CLI: resolvePack(gbrain-investor) -> resolved.manifest.page_types == the 2 own only
```

gbrain 0.42.1.0 (master HEAD `eefe8b5`), Bun 1.3.14, Supabase Postgres. Related: #1574, #1726.

coder8080:
## Problem

Every `extends`-based schema pack resolved to its **own declared types only** — the
parent taxonomy was never inherited. Activating any lens pack silently replaced the
working base taxonomy with an empty or 2-type one, while the pack still looked
healthy (identity hash computed, phases declared).

Fixes #1749.

## Root cause

`resolvePack` (`src/core/schema-pack/registry.ts`) walked the `extends` chain only to
enforce the depth cap and build the cache snapshot, then built `ResolvedPack` from the
child's manifest **verbatim** — each `loadByName(parent)` result was pushed onto
`chain` as a name and then discarded. `borrow_from` was never materialized at all. The
merge was an explicit, never-landed TODO (`registry.ts`, TODOS.md "T20 — extends-chain
merging").

### Code

- **New `src/core/schema-pack/merge.ts`** — pure, deterministic merge helper:
  `mergeByKey`, `mergeUnion`, `mergePageTypes`, `mergeInheritedManifest`. No disk,
  no engine.
- **`registry.ts:resolvePack`** — retains each ancestor manifest while walking the
  chain, materializes `borrow_from`, and composes them into `resolved.manifest` via
  `mergeInheritedManifest`. The alias graph + closure hash are now computed on the
  **merged** manifest, so a cross-pack alias cycle surfaces as `AliasCycleError` at
  resolve time.
- **Cache invalidation** — the cache entry now tracks the extends chain **plus**
  borrowed pack names, so editing a borrowed pack cascade-invalidates its borrowers.
- **`index.ts`** — re-exports the merge surface.

The cascade to consumers was transparent: every downstream reader already consumes
`resolved.manifest`, so no per-consumer changes were needed.

### Merge contract

Precedence, highest → lowest: **child → borrowed → nearest parent … → base**.

- **Six fields inherit, child-wins:** `page_types`, `link_types`, `frontmatter_links`,
  `enrichable_types`, `filing_rules`, `takes_kinds`.
- **`page_types` ordering** matters because `inferTypeFromPack` is
  first-`path_prefix`-match-wins in array order. An **override** of a base type keeps
  the base's declared position (base's curated prefix priority stays authoritative); a
  **genuinely new** type — from the child, a `borrow_from`, or a middle pack — is
  prepended nearest-first, so a more-derived type's prefix wins regardless of chain
  depth. This keeps a type's priority independent of whether its pack is active
  directly or sits mid-chain.
- **`takes_kinds` is UNION, not replace.** It carries a Zod default, so an omitted
  field is indistinguishable from an explicit one; replace-semantics would let a child
  that never mentions `takes_kinds` silently wipe the parent's. Documented consequence:
  a child cannot **narrow** `takes_kinds` below base ∪ parent.
- **`frontmatter_links`** key on the composite `(page_type, link_type)`, NUL-delimited —
  both are unconstrained strings, so a space-join would collide `{"a b","c"}` with
  `{"a","b c"}`.
- **`borrow_from` is selective, non-transitive, fail-closed:** only the named
  `types`/`link_types` are pulled, from the target's **own** declarations; omitting a
  category borrows none of it; a missing target pack throws `UnknownPackError`,
  matching the extends path.

### Deliberate deviation from the issue

The issue proposes merging `calibration_domains` too. This PR **excludes `phases` and
`calibration_domains`** from inheritance (they stay child-only, exactly as before):
they gate real cycle execution (`cycle.ts` `packDeclaresPhase`), and the manifest
contract is that each pack declares its own participation explicitly. Auto-inheriting
them would silently make a child run cycle phases it never requested — a behavior
change well beyond fixing the taxonomy bug. That is why `gbrain-everything` still
re-declares its phases and its 7 `calibration_domains` by hand.

Filed as a P3 follow-up in `TODOS.md`: if that redeclaration becomes painful, add an
explicit opt-in flag (e.g. `inherit_phases: true`) so a pack author opts in
consciously. Happy to fold it into this PR if you'd rather have it now.

Also not included (issue's proposed fixes 2 and 3): the `doctor` signal for an
own-only active pack, and fail-loud-on-activation. With the merge landed, the
silent-empty-taxonomy condition can no longer arise through this path; both remain
reasonable defence-in-depth and are easy to add here if you want them.

### Tests

`test/schema-pack-merge.test.ts` — 25 tests, pure (no disk, no DB):

- parent types visible in the child; override replaces without duplicating
- ordering: new type prepended wins `inferTypeFromPack`; override keeps base position;
  a new type from a **middle** pack is still prepended (no 2-vs-3-level asymmetry)
- the non-`page_types` keyed fields (`link_types`, `frontmatter_links` composite key,
  `enrichable_types`, `filing_rules`)
- `takes_kinds` union, including a child that omits it
- `phases` / `calibration_domains` staying child-only
- `borrow_from` via `resolvePack`: selective, category-omission, fail-closed on a
  missing pack
- merged alias cycle → `AliasCycleError`; borrowed-pack edit cascade-invalidates the
  borrower; dangling alias target benign
- idempotency; `extends: null` unchanged

### Docs

- `docs/architecture/schema-packs.md` — new "Merge contract (`extends` + `borrow_from`)"
  section; corrected the `extends: gbrain-base` comment in the authoring example.
- `docs/architecture/lens-packs.md` — merge contract summary + why `gbrain-everything`
  re-declares phases/calibration domains.
- `docs/architecture/KEY_FILES.md` — entry for the new `merge.ts`; `registry.ts` entry
  updated to the current truth (merge + borrow materialization + borrowed-pack cache
  tracking).
- `TODOS.md` — T20 marked done; P3 follow-up filed.

## Verification

Resolving the bundled packs, this branch vs its parent (`5008b28`):

| pack | `extends` | own | page_types before | after | link_types |
|---|---|---|---|---|---|
| `gbrain-investor` | `gbrain-base` | 2 | **2** | **29** | 0 → 12 |
| `gbrain-recommended` | `gbrain-base` | 0 | **0** | **27** | 0 → 12 |
| `gbrain-everything` | `gbrain-investor` + borrow | 0 | **0** | **30** | 0 → 12 |
| `gbrain-creator` | `gbrain-base` | 1 | **1** | **27** | 0 → 12 |
| `gbrain-engineer` | `gbrain-base` | 1 | **1** | **28** | 0 → 12 |
| `gbrain-base` | `null` | 27 | 27 | 27 | 12 → 12 |
| `gbrain-base-v2` | `null` | 17 | 17 | 17 | 14 → 14 |

`extends: null` packs are unchanged — the merge is a no-op for them. `gbrain-creator`
resolves to 27 rather than 28 because its one own type (`atom`) is an **override** of a
base type, not a new one; `gbrain-everything` is 27 base + 2 investor + `learning`
(borrowed, new) + `atom` (borrowed, override) = 30.

The issue's `schema explain` repro: `gbrain-investor` now sees `person` (inherited from
base) **and** `thesis` (own).

- `bun run verify` — 31/31 checks green
- `bun test test/schema-pack-*.test.ts test/schema-pack/ test/lens-pack-manifests.test.ts
  test/markdown*.test.ts test/operations-schema-pack.test.ts` — **452 pass / 0 fail**
  across 28 files
- E2E not run locally (no `DATABASE_URL` on my box). Both E2E tests touching this
  surface (`schema-cathedral`, `type-unification-full-flow`) drive `extends: null`
  packs, for which the merge is a no-op.

## Notes for the reviewer

- **Identity is still the child's own bytes.** `manifest_sha8` / `packIdentity` are
  computed on the unmerged child manifest (unchanged v0.38 contract), so a parent edit
  doesn't move the child's identity — staleness is handled by the existing
  `invalidatePackCache` + stat-TTL path, which this PR extends to cover borrowed packs.
  Happy to fold merged content into the identity if you'd prefer.
- **A `borrow_from` entry naming a type the target doesn't declare is silently ignored**
  (a missing *pack* throws). That matches the pre-existing selective-borrow contract,
  but a lint rule or a warn would be a cheap follow-up — say the word.
- `TODOS.md` uses "T20" for two unrelated items (extends-chain merging, and
  `gbrain schema downgrade` in `src/commands/schema.ts`). This PR closes the former.
- No `VERSION` / `package.json` / `CHANGELOG.md` changes — leaving the release bump to
  the fix-wave process.


time-attack:
Superseded with credit by #3181 (just merged) — your change landed rebased onto current master with authorship preserved via Co-authored-by. Thank you!

Issue #2018, opened 2026-06-09
put_page write-through mirrors unrouted pages into an arbitrary federated source's repo


WilliamCourterWelch:
**Version:** gbrain 0.40.6.0 (engine: postgres/Supabase, MCP stdio via `gbrain serve`)

**Summary:** When `put_page` (called over MCP) creates a page that has no project/source context, its markdown write-through mirror is written into the on-disk `local_path` of an unrelated federated source, rather than the source the page is actually assigned to.

**Observed behavior:**
- `gbrain serve` runs with cwd = `$HOME` (no `.gbrain-source` pin there).
- Creating cross-cutting pages (tagged `internal`, no project) via `put_page` returns `write_through: { written: true, path: "/Users/me/Projects/housefax/<slug>.md" }`.
- `housefax` is an unrelated registered federated source. Its `page_count` stays at 4 (the new pages are **not** assigned to it) — yet ~20 unrelated pages' `.md` mirrors accumulated in its repo dir, including auto-created subfolders (`office-hours/`, `internal/pitfalls/`).
- The pages themselves correctly live in the `default` source (`local_path: null`).

**Expected behavior:** write-through should either (a) mirror to the assigned source's `local_path`, or (b) skip the on-disk mirror entirely when the assigned source has `local_path: null` (as `default` does). It should **not** fall back to writing into some other federated source's working tree — that silently pollutes an unrelated git repo.

**Impact:** unrelated project repos accumulate stray untracked markdown files from every cross-cutting page write. Confusing, and easy to accidentally commit another topic's notes into the wrong repo.

**Workaround:** periodically sweep untracked `.md` files out of the affected source dir. There is no config knob (`gbrain config`, `put`, `export`, `sources`) to control the write-through target.

time-attack:
Closing as fixed on master (see 7c27fa12 / #2128). Please reopen if still reproducible on current HEAD.

PR #2867, opened 2026-07-16
write-through: honor GBRAIN_REPO_PATH env override + warn on skipped writes

waihonger:
## Problem

`sync.repo_path` is a single config row. On a multi-machine brain sharing one Postgres, it cannot be a valid filesystem path on every host — and `gbrain import` rewrites it as a side effect (`src/commands/import.ts`), so whichever machine imported last silently breaks write-through on every other machine: captures return `written:false` / `repo_not_found` with no signal, and the brain quietly degrades to DB-only rows.

Hit in production: a `gbrain import` on machine A repointed the global to an A-only path; a week of captures on machine B went DB-only before anyone noticed (the receipt field was the only trace).

## Fix

- `GBRAIN_REPO_PATH` env var takes precedence over the `sync.repo_path` global in `writePageThrough`, so each host can pin its own hub repo.
- The `no_repo_configured` / `repo_not_found` skips now warn through the existing logger instead of returning silently, so path drift is loud instead of invisible.

No behavior change when the env var is unset and the configured path resolves.

### DIFF
--- a/src/core/write-through.ts
+++ b/src/core/write-through.ts
@@ -67,11 +67,26 @@ export async function writePageThrough(
 ): Promise<WriteThroughResult> {
   const sourceId = opts.sourceId ?? 'default';
   try {
-    const repoPath = await engine.getConfig('sync.repo_path');
+    // GBRAIN_REPO_PATH env var overrides the `sync.repo_path` config global.
+    // When two machines share one Postgres brain, `sync.repo_path` is a single
+    // row that cannot be a valid filesystem path on both hosts — and
+    // `gbrain import` rewrites it as a side effect, so whichever machine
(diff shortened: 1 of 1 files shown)


time-attack:
Thanks for writing this up — the silent DB-only degradation you describe is a real and unpleasant failure mode. The write-through path has since been reworked on master (#2018): it resolves the assigned source's own local_path first and only falls back to sync.repo_path for sources without one, plus a guard against writing into a sibling source's tree, so this patch no longer applies cleanly and the multi-host case is better handled per-source than via a new global env override. Closing; a small standalone PR that only adds the warn-on-skip logging would be an easier ask.

PR #2868, opened 2026-07-16
fix(links): use the real resolver on extract --stale + resolve path-style/unslugged wikilink targets

Andredsouza1984:
Two related link-graph fixes on the extraction path, found running a production brain with `link_resolution.global_basename` enabled and an Obsidian-style corpus.

## 1. `extract --stale` drops bare `[[name]]` wikilinks

`extract --stale` used `nullResolver` and skipped the `{ skipFrontmatter, globalBasename }` opts, unlike `extractLinksFromDB`. With `link_resolution.global_basename` enabled, bare `[[name]]` wikilinks were silently dropped on stale re-extraction — frontmatter suppression swapped out basename resolution entirely. Fix: pass the real resolver (with `resolveBasenameMatches`) and the same opts, mirroring `extractLinksFromDB`.

## 2. Path-style and unslugged wikilink targets never resolve

- `normalizeBasename` / `queryBasenameIndex` now extract the path tail before normalizing, so path-style inputs like `[[ops/changes/2026-05-01-foo]]` match the tail-keyed basename index.
- `DIR_PATTERN` wikilink targets with spaces/uppercase (`[[ops/services/Some Agent]]`) are slugified to the kebab-case page slug before the existence check, so Obsidian-style titles produce edges instead of being silently dropped.
- `ops` added to the `DIR_PATTERN` entity-dir whitelist (operations wikis: services, changes, runbooks).

## Notes for review

- `LINK_EXTRACTOR_VERSION_TS` is deliberately **not** bumped here — bump it on merge if you want existing deployments to re-extract previously stamped pages with the fixed resolver.
- Tests: new cases in `test/link-extraction.test.ts` and `test/extract-stale.test.ts` pin all three behaviors (red on the old code). `bun run verify` green.

### DIFF
--- a/src/commands/extract.ts
+++ b/src/commands/extract.ts
@@ -1594,6 +1594,8 @@ async function extractStaleFromDB(
   },
 ): Promise<{ linksCreated: number; timelineCreated: number; pagesProcessed: number; staleRemaining: number }> {
   const { dryRun, jsonMode, includeFrontmatter, sourceIdFilter, catchUp } = opts;
+  // Read globalBasename once, mirroring extractLinksFromDB line 1315.
+  const globalBasename = await isGlobalBasenameEnabled(engine);
   const versionTs = LINK_EXTRACTOR_VERSION_TS;
 
   // Pre-flight count — cheap indexed COUNT. dry-run reports and returns.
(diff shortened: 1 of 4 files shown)


time-attack:
Thanks for this — your first finding was exactly right, and it has since been fixed on master by #2576, which always passes the real resolver plus { skipFrontmatter, globalBasename } into extractPageLinks on the --stale sweep. Closing on that basis; the remaining pieces we're declining separately, since 'ops' in DIR_PATTERN is specific to one vault layout and the path-tail/slugify broadening changes basename match semantics in a way that could create false edges (a risk your own notes flag). Appreciate the detailed writeup.

Issue #2607, opened 2026-07-05
sync --full git fast path bypasses pruneDir/isSyncable — re-imports ops/ and metafile pages that incremental sync excludes


jerome-marshall-ntx:
**Version:** 0.42.56.0 (also current master `6461790`) · **Engine:** Postgres 16 + pgvector · **Runtime:** Bun

## What happens

Incremental sync and full sync disagree about what belongs in the corpus, because the full-sync file enumerator bypasses the canonical exclusion gate.

**Incremental** sync filters every manifest entry through `isSyncable` (`src/commands/sync.ts:445-447` and `1849-1855`), whose `classifySync` (`src/core/sync.ts:341-360`) excludes:

- any path with a segment blocked by `pruneDir` — dot-dirs, `node_modules`, `vendor`, `dist`, `build`, `venv`, `.raw`, and `ops/` at any depth (`PRUNE_DIR_NAMES`, `src/core/sync.ts:242-259`);
- metafile basenames `schema.md` / `index.md` / `log.md` / `README.md` (`SYNC_SKIP_FILES`, `src/core/sync.ts:330`).

**Full sync** (`performFullSync` → `runImport` → `collectSyncableFiles`, `src/commands/sync.ts:2952`) never applies `isSyncable`. When the source is a git work tree it takes the `git ls-files` fast path, `gitListSyncableFiles` (`src/commands/import.ts:530-560`), which filters **only** by strategy/extension (`isCollectibleForWalker`) plus `.gitignore` — no `pruneDir`, no metafile exclusion. The non-git FS-walk fallback does call `pruneDir` at descent time (`import.ts:614`) but still imports the metafile basenames. And `runImport` itself applies no `isSyncable` check downstream.

Net effect on a git-backed source (the common case): `sync --full` imports — and *re-imports/resurrects previously soft-deleted* — everything incremental sync is designed to exclude: `ops/**`, `index.md`, `README.md`, `log.md`, `schema.md`, etc. A brain that was clean under months of incremental syncs silently regains excluded pages after any event that forces a full pass (first sync, `--full`, chunker-version bump, lost sync anchor).

## Observed

On a personal markdown vault synced as a source: pages under `ops/` that incremental sync had correctly excluded (and that had been soft-deleted) came back after a full re-import, with the incremental path then treating them as deletable again — the two modes cycle content in and out depending on which one ran last.

## What should happen

[...]

Issue #2404, opened 2026-06-25
ops/ hardcoded in PRUNE_DIR_NAMES: sync silently deletes ops/* pages and never imports ops/* files


rspinabellaagent:
# Bug: `ops/` directory is hardcoded in `PRUNE_DIR_NAMES`, so sync silently deletes any `ops/*` page and never imports `ops/*` files

**gbrain version:** 0.41.14.0
**Component:** `src/core/sync.ts` (and the delete path in `src/commands/sync.ts`)
**Severity:** High — silent loss of pages from the brain DB (markdown source files are unaffected, but the pages vanish and can never be re-synced).

## Summary

`PRUNE_DIR_NAMES` in `src/core/sync.ts` includes the literal string **`'ops'`** alongside
`'node_modules'` and `'.raw'`. `classifySync()` treats **any path containing an `ops`
segment** as `'pruned-dir'`, so:

- `isSyncable('ops/anything.md')` returns `false` → `collectSyncableFiles` / `runImport` /
  `performFullSync` **never treat `ops/*` files as import candidates** (even `gbrain sync --full`
  reports `imported=0` for them).
- A **modified** `ops/*` file falls into the `unsyncableModified` delete loop in
  `commands/sync.ts`. Its `unsyncableReason()` is `'pruned-dir'` (not `'metafile'`), so the
  delete-skip guard does **not** fire, and the page is **deleted** ("Deleted un-syncable page").

Net effect: **any page under a directory named `ops` is deleted on every sync and can never be
(re)imported.** This includes pages created via `gbrain put` with an `ops/...` slug — they are
silently destroyed by the next `gbrain sync` or `gbrain dream`.

Because `PRUNE_DIR_NAMES` matches a **single path segment basename at any depth**, the very
common directory name `ops/` (operations notes, runbooks, etc.) is unusable.

## Reproduction

```sh
# 1) A put-created ops page is deleted on the next sync:
echo -e "---\ntype: note\ntitle: t\ncreated: '2026-01-01'\n---\nhello" > /tmp/x.md
gbrain put ops/x < /tmp/x.md
gbrain list | grep ops/x          # present
gbrain sync                       # -> "Deleted un-syncable page: ops/x"
gbrain list | grep ops/x          # gone

# 2) A committed ops/*.md file is never imported:
mkdir -p <brain-repo>/ops && cp /tmp/x.md <brain-repo>/ops/x.md
[...]

PR #2870, opened 2026-07-16
fix(sync): never delete put_page pages under pruned dirs on incremental sync

Andredsouza1984:
Extends the v0.41.13 #1433 never-delete-on-metafile guard to the `pruned-dir` classification.

## The bug

A path under a pruned segment (`PRUNE_DIR_NAMES`: `ops/`, `node_modules/`, dot-dirs) is never sync-eligible, so a page at such a slug can only have come from `put_page` — whose write-through still lands a file in the brain repo. When anything edits that file (e.g. a merge-write appending source notes), the next incremental sync classifies it "modified but unsyncable" and **deletes the page from the index**. Sync can never re-create what it deletes there, so the page is orphaned permanently.

We tripped this in production: an `ops/design-history/*` page (created via `put_page`, later edited on disk) was deleted by the following incremental sync and had to be restored by hand.

## The fix

One line: the #1433 guard already skips deletion for the `metafile` classification; skip `pruned-dir` too.

## Tests

New regression test in `test/sync-metafile-skip.serial.test.ts` — red on the old code, green with the fix; the narrow-guard negative case (strategy-reason cleanup still deletes) still passes. `bun run verify` green.

### DIFF
--- a/src/commands/sync.ts
+++ b/src/commands/sync.ts
@@ -1885,7 +1885,13 @@ async function performSyncInner(engine: BrainEngine, opts: SyncOpts): Promise<Sy
   const pageOpts = opts.sourceId ? { sourceId: opts.sourceId } : undefined;
   for (const path of unsyncableModified) {
     // v0.41.13 #1433: never delete on metafile classification.
-    if (unsyncableReason(path, syncOpts) === 'metafile') continue;
+    // Same for pruned-dir. A path under a pruned segment (`ops/`,
+    // `node_modules/`, dot-dirs) was NEVER sync-eligible, so any page at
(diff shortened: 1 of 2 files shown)


time-attack:
Closing: current master already contains this fix. Master already contains the identical guard: src/commands/sync.ts:2313-2314 `const reason = unsyncableReason(path, syncOpts); if (reason === 'metafile' || reason === 'pruned-dir') continue;` with a '#2404 hardening' comment, landed in commit 42375bd 'fix(sync): stop the sync data-loss family — ops/ prune, DB-only write-through, full-sync gate drift (#2404, #2426, #2607) (#2938)'. The PR's one-line fix is byte-for-byte what master ships.

PR #2871, opened 2026-07-16
feat(sync): per-source keep_dirs — exempt named dot-dirs from the sync prune

mohanraj00:
## Problem

`pruneDir` unconditionally excludes dot-directories, so a repo's `.agents/*.md` / `.claude/skills/**` policy docs can never sync as pages. Worse, the two enumeration paths disagree about it:

- the **first-sync `git ls-files` fast path** filters by extension only, so tracked dot-dir markdown IS imported;
- the **incremental classifier** (`classifySync`) prunes those same paths, so the first time such a file is modified, sync **deletes the page** as "un-syncable".

Observed live: a sync deleted `.agents/agents`, `.agents/claude`, and two `.claude/skills/*/skill` pages that the initial sync had itself imported.

The existing `include` option can't express this — it's a restrictive allowlist (checked after the prune anyway), so `include: ['.agents/**']` would drop every *other* page as `include-glob-miss`.

## Fix

New per-source **`config.keep_dirs`** — directory segment names exempted from the prune:

- `classifySync` skips the `pruned-dir` rejection for listed segments (file segment never exempted; every other gate — strategy, metafile, include/exclude globs — still applies, and a pruned segment *below* a kept dir still prunes, so `.agents/.git/x.md` stays out).
- `collectSyncableFiles` FS-walk fallback descends into kept dirs (git fast path already enumerates them).
- `performSyncInner` hoists the setting from the source row at one choke point, so single-source CLI, `--all` fan-out, jobs handler, and watch all honor it without each caller re-reading config.
- New CLI: `gbrain sources keep-dirs <id> [dir ...] [--clear]` (mirrors `federate`'s config-flip shape). Entries are segment names, not globs — a `/` is refused loudly.

## Usage

```
gbrain sources keep-dirs my-repo .agents .claude
gbrain sync --source my-repo --full   # restore previously-deleted pages
```

## Testing

- New `keepDirs` block in `test/sync-strategy.test.ts` (kept-dir admitted, unlisted dot-dirs still pruned, metafile/strategy gates still apply, nested prune below a kept dir, file-named-like-dir not admitted) — 17 pass.
- `bun run verify` — all 30 checks green.
- Live-verified on a real source: pages restored by `--full`, and a subsequent incremental sync no longer deletes them on modification.

### DIFF
--- a/src/commands/import.ts
+++ b/src/commands/import.ts
@@ -45,7 +45,7 @@ export interface RunImportResult {
 export async function runImport(
   engine: BrainEngine,
   args: string[],
-  opts: { commit?: string; strategy?: SyncStrategy; sourceId?: string; managedBookmark?: boolean } = {},
+  opts: { commit?: string; strategy?: SyncStrategy; sourceId?: string; managedBookmark?: boolean; keepDirs?: string[] } = {},
 ): Promise<RunImportResult> {
   const noEmbed = args.includes('--no-embed');
   const fresh = args.includes('--fresh');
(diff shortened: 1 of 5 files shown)


time-attack:
The keep_dirs feature was declined for now with the content-scope set — but the sync-deletes-its-own-imports bug this PR also fixes is REAL and now tracked separately so it doesn't get lost (see the tracking issue referenced below). Thank you for finding it.

PR #2875, opened 2026-07-16
fix(search): preserve email citation metadata across result paths

### LINKED ISSUE #2962: Search paths drop trusted email citation provenance across hydration and cache
**What happened?**

Email-backed pages can carry a trusted `message_id`, optional `thread_id`, and exact subject in frontmatter, but search results do not preserve that provenance consistently.

The metadata is absent or dropped across several result-producing paths, including SQL projection, alias injection, relational recall/fanout, two-pass hydration, vector fusion/reranking, and semantic-cache serialization. A result that is otherwise correct therefore cannot be cited back to the exact email deterministically.

There is also a trust-boundary problem if projection accepts an internal `source_subject` field from raw frontmatter rather than deriving it from the allowlisted external `subject` field. Malformed top-level JSON-string frontmatter must not be double-decoded into trusted provenance.

**What did you expect?**

Every search path should preserve the same citation DTO for an email-backed result:

- `message_id` only when it is a nonblank trusted frontmatter string
- optional `thread_id`
- exact `source_subject`, derived from allowlisted `subject` for raw/frontmatter records
- internally projected `source_subject` retained during hydration and reranking

Raw `source_subject` must not be trusted, malformed/non-object frontmatter must fail closed, and cached DTOs from before this contract must be invalidated.

**Steps to reproduce**
1. Store an email-backed page with object-shaped frontmatter containing `message_id`, `thread_id`, and `subject`.
2. Retrieve it through direct hybrid search and note whether the result contains the three citation fields.
3. Retrieve the same page through an alias hop, relational recall/fanout, or two-pass hydration.
4. Store and replay the result through the semantic query cache.
5. Observe that one or more paths omit the provenance fields, preventing exact email citation.
6. As an adversarial variant, supply raw `source_subject` or top-level encoded JSON-string frontmatter and observe that it must not become trusted citation metadata.

Focused regressions covering these paths are included in #2875.

**Environment**
- gbrain version: `0.42.59.0`
- OS: macOS 26.5.2
- Bun version: `1.3.14`
- Database: reproduced and regression-tested with PGLite; PostgreSQL parity coverage included in the linked PR

```json
{
  "omitted": "The defect is deterministic and covered by isolated engine/search tests; raw doctor output is not attached because it may contain local source metadata."
}
```

**Impact**

Consumers cannot produce or repair exact email citations reliably even when the underlying page has sufficient provenance. This weakens evidence traceability and can cause different retrieval paths to return semantically equivalent results with different citation capabilities.

**Proposed fix**

Centralize trusted email citation projection, preserve the fields through every result transformation and cache DTO, enforce raw/internal field precedence, fail closed on malformed frontmatter, and bump the semantic-cache contract so older DTOs cannot survive.

**Related PR**

#2875

amtagrwl:
Closes #2962

## Summary

- project trusted Message-ID, optional thread ID, and exact email subject across PGLite/PostgreSQL search results
- preserve that provenance through alias injection, relational recall/fanout, two-pass hydration, vector fusion/reranking, and semantic-cache hits
- apply authoritative subject precedence: allowlisted raw `subject` is projected to internal `source_subject`; raw `source_subject` is not trusted
- fail closed on malformed/non-object frontmatter and preserve quarantine/source-visibility enforcement across indirect retrieval paths
- bump `KNOBS_HASH_VERSION` to 13 so cached result DTOs from before the citation contract cannot be reused

## Problem

Email-backed pages may contain enough provenance for an exact citation, but different search and hydration paths return different result DTOs. Metadata can disappear before the result reaches a consumer or be lost when the result is cached. That makes exact citations path-dependent and prevents deterministic citation repair.

## Safety contract

- `message_id` must be a nonblank trusted frontmatter value before exact email citation fields are published
- raw/frontmatter `source_subject` is ignored; only allowlisted `subject` may supply the raw exact subject
- internal SQL projection records may preserve `source_subject` during hydration/reranking
- malformed top-level JSON-string frontmatter is not double-decoded and fails closed
- source grants, page/source visibility, quarantine, deletion, and archive state are rechecked through aliases, relational paths, graph expansion, hydration, and cache lookup

## Verification

- `bun run verify` — 31/31 checks green
- `bun run typecheck` — green
- `bun run check:all` — green
- focused projection/cache/alias/relational/two-pass/reranker/security suites — green locally
- PGLite cache/source/visibility regressions — green locally
- prior exact-head PostgreSQL metadata/projection E2E — green on a superseded head; must be rerun on the final head

## Current status

**Not merge-ready.** Exact-head review of `a3d37a3e1e76139869eb34697f786709adbaa825` found remaining semantic-cache race/invalidation/replay-limit issues. They must be fixed and the final unchanged SHA must pass focused tests, serial tests, PostgreSQL parity, and exact-head review before merge.

Fork workflows for the current head are also awaiting repository-maintainer approval; no CI result is being claimed for this SHA.


time-attack:
Thank you for the depth here, and for being straightforward that the current head is not merge-ready. We're closing it: 1500+ lines across both engines, the query cache, two-pass hydration, relational recall and quarantine — with an acknowledged semantic-cache race outstanding — is not something we can review safely on the source-isolation and cache-keying path. If email citation provenance is worth pursuing, the way in is a much smaller diff scoped to one projection path with the cache contract left alone.

PR #2877, opened 2026-07-16
fix(extract): derive timeline entries from effective dates

ShintaroKawakami:
## Summary
Fix timeline backfill so DB-backed extraction creates a structured timeline entry from `pages.effective_date` when a page has no explicit markdown timeline row.

## Changes
- Add a deterministic fallback row using the effective date, page title, and date provenance.
- Apply the same fallback to `gbrain extract --stale` so initial and incremental extraction converge.
- Include `effective_date` fields in Postgres/PGLite `getPage` and stale-page projections.
- Add idempotency and precedence regression tests and update extraction documentation.

## Testing
- `bun test test/extract.test.ts test/extract-db.test.ts test/extract-stale.test.ts` — 51 passed.
- `bun run verify` — 31/31 checks passed.
- The full unit suite was started locally but stopped after approximately 10 minutes at 20% progress because this environment was projecting a 40+ minute runtime; no failures were observed before stopping.

## Checklist
- [x] Tests added/updated
- [x] Documentation updated
- [x] No breaking changes


time-attack:
Thanks for the clean implementation and tests. We're closing this one because it isn't a fix so much as a new behavior: it manufactures timeline rows that were never in the page markdown, for every dated page in a brain. The part that worries us most is reconciliation — if a page's effective_date or title later changes, the previously written fallback row isn't removed, so timelines would accumulate stale synthesized entries. Deriving that view at read time, rather than writing rows, would avoid the problem entirely.

PR #2883, opened 2026-07-16
feat(propose_takes): make extraction output-token budget configurable

realugbun:
## The problem

When gbrain runs its maintenance cycle against a **local, self-hosted model**, the `propose_takes` phase can silently produce nothing. It runs, spends time and GPU, and reports success, but inserts **zero** takes. There is no error to go on.

It shows up with some open models that do a lot of reasoning before answering (for example qwen3.6 in thinking mode). It is not every model, only the ones whose reasoning runs long enough to hit the output limit before they answer. People running gbrain locally for privacy, cost, or offline use tend to reach for exactly those models, so a core feature quietly stops working for them.

A second cost makes it worse. When the call returns nothing, the page is never recorded as processed, so there is no negative-result cache for it (see #2106). Every later cycle re-runs the same call on the same file and gets the same empty result, so these pages are re-processed on every run indefinitely, burning local inference each time.

## Why it happens

The extraction call asks the model for at most **2048 output tokens**. A heavy-reasoning model spends output tokens thinking before it emits its JSON answer, and that reasoning can exceed 2048 tokens on its own. The response is cut off (`finish_reason: length`) before any JSON is produced, so the parser sees empty output and the phase proposes nothing.

Measured on qwen3.6 (llama.cpp), same page and prompt, only the cap changed:

| output cap | `finish_reason` | tokens used | JSON returned | takes |
| --- | --- | --- | --- | --- |
| 2048 (current) | `length` | 2048 | none (truncated mid-reasoning) | **0** |
| 8192 | `stop` | 5404 | full array | **4** |

## The fix

Make the output-token ceiling configurable, and leave the default unchanged.

A local user with a heavy-reasoning model opts in:

```bash
gbrain config set propose_takes_max_output_tokens 8192
# or
GBRAIN_PROPOSE_TAKES_MAX_OUTPUT_TOKENS=8192 gbrain dream
```

Everyone else is unaffected. The default stays **2048**, so there is no change to behavior, latency, or cost for existing setups.

The key follows gbrain's existing `embedding_dimensions` convention: a typed `GBrainConfig` field with a matching `GBRAIN_*` override read in `loadConfig()`, registered in `KNOWN_CONFIG_KEYS`. The phase reads the value from `ctx.config` and passes it to the extractor.

Complementary to #2559: that change strips inline `<think>` tags so JSON that arrives can be parsed; this one ensures there is enough budget for the JSON to be emitted in the first place.

## Changes

- `src/core/config.ts`: new `propose_takes_max_output_tokens` field, `GBRAIN_PROPOSE_TAKES_MAX_OUTPUT_TOKENS` env mapping, and `KNOWN_CONFIG_KEYS` entry.
- `src/core/cycle/propose-takes.ts`: default of 2048, resolved from `ctx.config` and threaded to the extractor.

## Tests

New tests for this change:

- `defaultExtractor` uses the default (2048) when unset, and honors an explicit override.
- The phase reads `propose_takes_max_output_tokens` from config and passes it through to the extractor.
- `loadConfig()` reads the value from `GBRAIN_PROPOSE_TAKES_MAX_OUTPUT_TOKENS`.

### Full suite

Ran on a clean checkout of this branch (Bun 1.3.14): the `bun run test` parallel runner plus `bun run verify` (the pre-push gate). I had to fire up a 24 core 64GB RAM VM to get this suite to run!

- **`bun run verify`** — **31/31** checks green.
- **`bun run test`** — **14,175 passing, 2 failing**. Both failures are pre-existing and unrelated to this change (this PR touches only `src/core/config.ts` and `src/core/cycle/propose-takes.ts`):

| Failing test | Introduced by this PR? |
| --- | --- |
| `test/fuzz/filesystem-validators.test.ts` → ``validateUploadPath fuzz (fs-backed) > shaped traversal probes: explicit `..` patterns rejected`` | No — fails identically with this commit removed |
| `test/retrieval-reflex.test.ts` → `logDeliveredReflexPointers logs channel=reflex events through the drained sink` | No — fails identically with this commit removed |

To confirm they aren't introduced here, I checked out `master` (`5008b28`, v0.42.59.0 the commit this branch sits on) and re-ran each file: both fail the same way without this PR's commit present. They could be environment-specific, an fs-backed property/fuzz test and an async drained-sink logging assertion, rather than anything this change affects. The tests added by this PR pass.

### DIFF
--- a/src/core/config.ts
+++ b/src/core/config.ts
@@ -177,6 +177,14 @@ export interface GBrainConfig {
    * reflex knobs.
    */
   retrieval_reflex_window_turns?: number;
+  /**
+   * Max output tokens for the propose_takes extraction call (default 2048).
+   * Raise it for local models that do a lot of reasoning before answering (e.g.
+   * qwen3 in thinking mode): they can spend the whole default budget reasoning
+   * and get truncated before emitting any JSON, producing an empty extraction.
+   * File-plane / env (GBRAIN_PROPOSE_TAKES_MAX_OUTPUT_TOKENS).
+   */
+  propose_takes_max_output_tokens?: number;
   embedding_image_ocr?: boolean;
   embedding_image_ocr_model?: string;
 
@@ -552,6 +560,10 @@ export function loadConfig(): GBrainConfig | null {
       Number.isFinite(Number(process.env.GBRAIN_RETRIEVAL_REFLEX_WINDOW_TURNS))
       ? { retrieval_reflex_window_turns: Number(process.env.GBRAIN_RETRIEVAL_REFLEX_WINDOW_TURNS) }
       : {}),
+    ...(process.env.GBRAIN_PROPOSE_TAKES_MAX_OUTPUT_TOKENS &&
+      Number.isFinite(Number(process.env.GBRAIN_PROPOSE_TAKES_MAX_OUTPUT_TOKENS))
+      ? { propose_takes_max_output_tokens: Number(process.env.GBRAIN_PROPOSE_TAKES_MAX_OUTPUT_TOKENS) }
+      : {}),
     ...(process.env.GBRAIN_REMOTE_CLIENT_SECRET && fileConfig?.remote_mcp
       ? { remote_mcp: { ...fileConfig.remote_mcp, oauth_client_secret: process.env.GBRAIN_REMOTE_CLIENT_SECRET } }
       : {}),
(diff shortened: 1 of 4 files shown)


time-attack:
Thanks for the careful writeup and the measured qwen3.6 numbers — the diagnosis (reasoning tokens eating the 2048 cap before any JSON is emitted) is clearly correct. But we're not taking a new config key plus env override for this: the default is unchanged, so the failure still bites every affected user out of the box, and the knob only helps people who already know why their cycle proposed nothing. Closing as config/feature surface we're not adding; if this is ever addressed it will be by handling the truncated-response case directly rather than exposing a budget setting.

PR #2887, opened 2026-07-17
fix(facts): normalize null-like extractor entities; recover dash-flattened page slugs

sanchalr:
### Problem

Extractors sometimes emit placeholder entity strings (the literal "null", "none", "n/a") or page slugs with '/' flattened to '-' (companies-acme for companies/acme). The parser accepts any string entity and the resolver's slugify floor persists these as-is, creating fact rows attached to slugs no page backs.

### Fix

- Normalize null-like extractor entity strings to null before resolution.
- When a resolved slash-less slug has no live page in the source but exactly one live page's slug flattens to it, attach the fact to that page. Zero or multiple matches leave the slug unchanged.

Related: #2532 treats slugify-fallback guesses as unresolved and #2497 reworks the extract_facts guard predicate. This carries the two pieces neither touches.

### Testing

- `bun run typecheck`
- `bun test test/facts-backstop.test.ts test/facts-extract-silent-no-op.test.ts` - 20 passed, 0 failed

@sanchalr 🫡

### DIFF
--- a/src/core/facts/backstop.ts
+++ b/src/core/facts/backstop.ts
@@ -324,9 +324,35 @@ async function runPipelineWithBody(
     // D4: notability filter applied post-extraction, pre-insert.
     if (filter === 'high-only' && f.notability !== 'high') continue;
 
-    const resolvedSlug = f.entity_slug
+    let resolvedSlug = f.entity_slug
       ? await resolveEntitySlug(ctx.engine, ctx.sourceId, f.entity_slug)
       : null;
+
+    // Recover dash-flattened references to real pages. Extractors sometimes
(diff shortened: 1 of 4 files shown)


time-attack:
Thanks for digging into the fact-attachment path — the null/none/n-a placeholder observation is real. We're closing this one though: the substantive half is a heuristic that guesses which page a dash-flattened slug meant, and re-attaching facts based on a guess (plus two extra per-fact queries inside the backstop loop) is the opposite direction from treating unresolved slugs as unresolved. The remaining half is cosmetic data noise rather than a user-blocking bug, so it doesn't carry the PR on its own.

PR #2889, opened 2026-07-17
fix(embeddings): wire trust_custom_dims into the actual dimension passthrough

dnsco:
## Summary

`trust_custom_dims: true` is declared on the `ollama` and `llama-server` recipes and documented in `types.ts` as opting a local embedding model into custom-dimension truncation. But `dimsProviderOptions()` never actually consulted it — the flag only relaxed the schema-width validation in `embedding-dim-check.ts`. A brain could be configured with a truncated `embedding_dimensions`, but the gateway still requested the model's full native-size output from the provider, because any `openai-compatible` model outside the explicit ZE/Voyage/text-embedding-3/DashScope/Zhipu/MiniMax allowlist fell through to `return undefined`.

The mismatch only surfaces at first embed (or in `gbrain doctor`'s live probe) as an opaque dimension-mismatch error — well after `gbrain init --embedding-model ollama:... --embedding-dimensions N` reports success.

## Repro

```
$ gbrain init --pglite --embedding-model ollama:qwen3-embedding:8b --embedding-dimensions 2000
{"status":"success", ...}   # looks fine

$ gbrain doctor --json
"embedding_provider": "warn — Embedding provider probe failed: Embedding dim mismatch:
  model qwen3-embedding:8b returned 4096 but schema expects 2000."
```

Confirmed independently against a live Ollama instance that this isn't an Ollama limitation — both its native `/api/embed` and OpenAI-compat `/v1/embeddings` endpoints honor a `dimensions` field and return a genuinely truncated vector:

```
$ curl -s http://localhost:11434/v1/embeddings -d '{"model": "qwen3-embedding:8b", "input": "test", "dimensions": 2000}'
# returns a 2000-length embedding vector
```

gbrain just never sent the field for this provider.

## Fix

Thread `trust_custom_dims` through as a 5th `dimsProviderOptions` param, sourced from the resolved recipe at `gateway.ts`'s single `embed()` call site. Add a generic fallback branch *after* the existing per-model matches, so ZE/Voyage/text-3/DashScope/Zhipu/MiniMax keep their explicit (and in some cases dim-validated) handling unchanged, and only truly generic local-model recipes fall through to the new branch.

This covers `ollama` and `llama-server` uniformly instead of adding a new per-model allowlist entry every time a local embedding model ships (e.g. this would otherwise need `nomic-embed-text`, `mxbai-embed-large`, `qwen3-embedding`, `snowflake-arctic-embed-l-v2`, `all-minilm`, ... hardcoded individually).

## Test plan

- [x] New unit tests in `test/ai/dims-trust-custom.test.ts`: fallback fires only when `trustCustomDims=true`, doesn't affect the ZE branch's own dim validation, ignores `inputType` (symmetric assumption), doesn't leak into `native-openai`.
- [x] `bun test test/ai/dims-trust-custom.test.ts test/ai/dims-openai.test.ts test/ai/dims-zeroentropy.test.ts` — 36 pass, 0 fail.
- [x] `bun test test/ai/gateway.test.ts` — 36 pass, 0 fail (no regression from the new call-site param).
- [x] `bun run typecheck` — clean.

### DIFF
--- a/src/core/ai/dims.ts
+++ b/src/core/ai/dims.ts
@@ -108,12 +108,26 @@ export function isValidOpenAITextEmbedding3Dim(modelId: string, dims: number): b
  * `'document'`. Per-model filtering happens INSIDE the switch — the field
  * is NEVER emitted for providers that don't accept it (OpenAI text-3,
  * DashScope, Zhipu) so the request body stays clean for those endpoints.
+ *
+ * `trustCustomDims` — recipes that declare `trust_custom_dims: true`
+ * (ollama, llama-server) promise every listed embedding model accepts
+ * OpenAI-shaped `dimensions` truncation (Ollama's `/v1/embeddings` honors
+ * it; llama-server serves whatever model the user launched, so the user
+ * is asserting it does too). Before this flag was wired here, it only
+ * relaxed the schema-width validation in embedding-dim-check.ts — a brain
+ * could be configured for a truncated width, but the gateway silently
+ * requested the model's native (larger) output, producing a dim mismatch
+ * that only surfaced at first embed (or in `gbrain doctor`'s live probe).
+ * This is the generic fallback for any openai-compatible recipe that
+ * opts in, so it covers ollama, llama-server, and future local-model
+ * recipes uniformly instead of a per-model allowlist entry each time.
  */
 export function dimsProviderOptions(
   implementation: Implementation,
   modelId: string,
   dims: number,
   inputType?: 'query' | 'document',
+  trustCustomDims?: boolean,
 ): Record<string, any> | undefined {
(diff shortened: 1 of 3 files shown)


time-attack:
Thank you for the thorough repro and for actually curling Ollama to confirm it truncates — that part is convincing. The problem is scope: `trust_custom_dims` today means "don't validate the declared width", and this makes it also send `dimensions` on every request for ollama, llama-server and litellm-proxy alike, including brains where the declared width already equals the model's native size. llama-server serves whatever model the user launched and litellm forwards the field upstream, so setups that work today can start returning 400s — that's a bigger risk than the current loud doctor warning. Closing on that basis.

PR #2894, opened 2026-07-17
fix(cycle): abort propose_takes after 3 consecutive retryable transport failures

mdcruz88:
# fix(cycle): abort propose_takes after 3 consecutive transport failures

## Problem

gbrain cycle phases (including `propose_takes`) catch per-item LLM errors and
continue so one bad page cannot kill the whole run. That is correct for parse
failures and content rejection — but when the **egress path is fully dead**
(proxy packet-drop, DNS blackhole, provider 5xx storm), the same loop becomes
an infinite-retry / silent-success wedge:

- every page's extractor call fails
- each failure is swallowed into `warnings`
- the phase returns **status=ok** with zero proposals written
- process-level **rc=0**

**Evidence (2026-07-17):** under an nftables packet-drop drill against the
proxy port on our box, a fully blocked run exited rc=0 reporting the operator-
visible "Embedded 0 chunks" / zero-work success class. Cron and autopilot
treat that as healthy.

A second wedge exists one layer up: even when the phase correctly returns
`status: 'fail'`, the cycle rollup turns mixed fail+ok into overall
`'partial'`, and `gbrain dream` exits nonzero only on overall `'failed'`.
A normal multi-phase dream that trips the transport budget would still exit
rc=0.

## Fix

Abort `propose_takes` after a budget of **3 consecutive retryable transport
failures**. Phase result is **`status: 'fail'`** with
`details.transport_abort: true`. The cycle rollup treats that marker as
**cycle-fatal**: overall status becomes `'failed'` (not `'partial'`), so
`gbrain dream` exits nonzero — including multi-phase runs where another
phase succeeded. Ordinary per-item phase failures still roll to `'partial'`.

### What counts (retryable transport)

- Timeout / deadline (`TimeoutError`, `ETIMEDOUT`, deadline-exceeded messages)
- **Abort-caused-by-deadline/timeout only** — `AbortError` counts when its
  message, cause chain, or abort `reason` indicates a timeout/deadline
  (e.g. nested `TimeoutError`, "aborted due to timeout"). A **bare
  AbortError** (user cancellation, unrelated signal) does **not** count.
- Connection / DNS (`ECONNREFUSED`, `ENOTFOUND`, `EAI_AGAIN`, `ECONNRESET`, …)
- Provider **5xx** (`status` / `statusCode` 500–599)

### What does **not** count

- Bare `AbortError` without a timeout-shaped cause/reason/message
- Parse failures (extractor returns `[]` today; explicit parse-shaped throws
  also stay non-transport)
- Content rejection / safety filter
- Auth / config errors (401/403 and similar 4xx)
- Empty results

### Cause-chain classification (reviewer footgun)

Classification walks the error **cause chain** (`Error.cause`),
`AggregateError.errors`, and abort **`reason`** (AbortSignal / DOMException).
An SDK retry loop that exhausts and rethrows a wrapper whose only timeout
lives on `.cause` still increments the budget — message-string regex alone
is not sufficient.

### Counter rules

- Increment only on retryable transport failures
- **Reset to 0 after any successful LLM response**
- Non-transport errors neither increment nor reset
- On abort, the phase report preserves:
  - `transport_abort: true`
  - `abort_reason: 'consecutive_transport_failures'`
  - `consecutive_transport_failures`
  - `transport_failure_slugs` (affected page slugs)
  - `transport_failure_errors` (last error messages)

### Cycle rollup

- `deriveStatus` checks `details.transport_abort` via `phaseForcesCycleFailed`
- If any phase has the marker → overall `'failed'`
- Ordinary fail + ok (no marker) → still `'partial'`
- `dream.ts` exits `process.exit(1)` when `report.status === 'failed'`

## Scope

- Budget lives in `propose_takes` (the per-item catch-and-continue seam)
- Shared classifier: `src/core/cycle/transport-failure.ts` (reusable by
  sibling phases later; not retrofitted here — minimal upstream-shaped diff)
- Fatal rollup marker: `cycle.ts` `deriveStatus` + `phaseForcesCycleFailed`
- No cycle-level abort-signal wiring in this PR (consecutive budget is the
  direct fix and takes priority)

## Tests

`test/cycle/propose-takes-transport-abort.test.ts`:

1. 3 consecutive retryable failures → phase `fail` + budget fields
2. Interleaved success resets the counter
3. Auth / non-retryable errors do not increment
4. Wrapped / AggregateError timeout still increments (cause chain)
5. Last errors + page slugs present in the phase report
6. Bare AbortError does not trip budget (phase completes, no abort)
7. Multi-phase: transport_abort + sibling ok → overall `failed`; ordinary
   fail+ok stays `partial`; dream exit mapping is nonzero only on `failed`

## Files

- `src/core/cycle/transport-failure.ts` (new)
- `src/core/cycle/propose-takes.ts` (budget at extractor catch)
- `src/core/cycle.ts` (`phaseForcesCycleFailed` + `deriveStatus` fatal rollup)
- `test/cycle/propose-takes-transport-abort.test.ts` (new)

### DIFF
--- a/src/core/cycle.ts
+++ b/src/core/cycle.ts
@@ -2428,8 +2428,32 @@ function extractTotals(phases: PhaseResult[]): CycleReport['totals'] {
   return t;
 }
 
-function deriveStatus(phases: PhaseResult[], totals: CycleReport['totals']): CycleStatus {
+/**
+ * True when a phase result must escalate the OVERALL cycle status to
+ * 'failed' (not 'partial'). Used for transport-budget abort: ordinary
+ * per-item phase failures still roll to 'partial', but a dead egress
+ * path that trips the consecutive-transport budget is cycle-fatal so
+ * `gbrain dream` exits nonzero.
+ *
+ * Marker: `details.transport_abort === true` (set by propose_takes when
+ * CONSECUTIVE_TRANSPORT_FAILURE_BUDGET trips).
+ */
+export function phaseForcesCycleFailed(p: PhaseResult): boolean {
+  return p.details?.transport_abort === true;
+}
+
+/**
+ * Derive overall CycleReport.status from phase results + totals.
+ * Exported for unit tests of the transport-abort fatal rollup.
+ */
+export function deriveStatus(phases: PhaseResult[], totals: CycleReport['totals']): CycleStatus {
   if (phases.length === 0) return 'failed';
+  // Transport-budget abort is cycle-fatal. A multi-phase dream where
+  // propose_takes trips and another phase succeeds must NOT roll up to
+  // 'partial' (dream exits nonzero only on overall 'failed' — the
+  // silent-success wedge this patch kills). Ordinary fail+ok stays
+  // 'partial'.
+  if (phases.some(phaseForcesCycleFailed)) return 'failed';
   const anyFailed = phases.some(p => p.status === 'fail');
   const allFailed = phases.every(p => p.status === 'fail');
   const anyWarn = phases.some(p => p.status === 'warn');
--- a/src/core/cycle/propose-takes.ts
+++ b/src/core/cycle/propose-takes.ts
@@ -39,6 +39,10 @@
 
 import { randomUUID, createHash } from 'node:crypto';
 import { BaseCyclePhase, type ScopedReadOpts, type BasePhaseOpts } from './base-phase.ts';
+import {
+  CONSECUTIVE_TRANSPORT_FAILURE_BUDGET,
+  isRetryableTransportFailure,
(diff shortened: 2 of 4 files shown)


time-attack:
Appreciate the packet-drop drill and the precise write-up of the rc=0 silent-success wedge — that failure class is real. We're still closing this: it's 665 lines introducing a new heuristic error-classification module and a change to cycle rollup semantics, wired into propose_takes only, so every other phase keeps the exact same wedge. A per-phase consecutive-failure budget plus an error-shape sniffer is more permanent surface than we want for a problem that's cross-cutting by nature.

PR #2897, opened 2026-07-17
Fix host skill manifest conformance in doctor (SUP-3488)

levineam:
## Summary\n- derive host skill manifests when manifest metadata is intentionally absent\n- keep doctor conformance aligned with resolver and brain-first checks\n- add regression coverage for host-managed skill trees\n\n## Verification\n- bun test test/doctor.test.ts\n- bun run typecheck\n- git diff --check\n\nTracks SUP-3488.

### DIFF
--- a/src/commands/doctor.ts
+++ b/src/commands/doctor.ts
@@ -4345,7 +4345,7 @@ export async function buildChecks(
 
   // 2. Skill conformance (SKILL group — gated)
   if (scope === 'all' && skillsDir) {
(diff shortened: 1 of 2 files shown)


time-attack:
Thanks for the patch. Closing this one: it's still a draft, and the description is a single escaped-newline blob pointing at an internal tracker ID (SUP-3488) with no reproduction or user-visible symptom we can verify against master. Doctor host-manifest conformance nuance doesn't clear the bar we hold drafts to, which is crucial bug fixes only.

PR #2899, opened 2026-07-17
feat(sources): per-source trust-frontmatter-slug opt-in for SLUG_MISMATCH

MohabMohie:
## Problem

Federated sources whose frontmatter `slug:` declares a real external identity
(e.g. a Docusaurus docs site's custom routes and blog-post URLs) fail the
path-vs-frontmatter anti-spoof check in `importFromFile` and get skipped as
`SLUG_MISMATCH`. There was no way to tell gbrain "this source's frontmatter
slugs are authoritative" short of hand-patching the anti-spoof rejection.

Concretely: `shafthq.github.io` (a Docusaurus site registered as a gbrain
source) has 83/187 markdown files (all blog posts + custom-route docs pages)
whose `slug:` frontmatter is the page's real published URL, not a path match.
`gbrain frontmatter validate --fix` must never run there — it would rewrite
public routes. Full report: [ShaftHQ/SHAFT_ENGINE#3618](https://github.com/ShaftHQ/SHAFT_ENGINE/issues/3618).

## Fix

An explicit, per-source opt-in — same shape as the existing `federated` flag:

```
gbrain sources trust-frontmatter-slug <id>
gbrain sources untrust-frontmatter-slug <id>
```

Stored as `config.trust_frontmatter_slug` (JSONB, same as `federated`).
When set, `importFromFile`'s SLUG_MISMATCH branch honors the frontmatter slug
instead of rejecting it — threaded through both `sync.ts` call sites
(incremental + `--full`) and `import.ts`'s `runImport` (the actual per-file
path `performFullSync` runs through, so `--full` sync sees the flag too, not
just incremental).

Docusaurus-style root-relative slugs (`slug: /testing/web`) are normalized
(leading `/` stripped) before use — `validateSlug`'s leading-slash rejection
in `src/core/utils.ts` is a security guard against path-derived spoofing at
the storage chokepoint and stays strict for every other caller; normalization
happens only on this explicitly-trusted path, found by running the fix
against the real 189-file source (see Verification).

Default is unset/false. The anti-spoof rejection is byte-for-byte unchanged
for every existing source that doesn't opt in — a new regression test pins
this (`default (unset): mismatched frontmatter slug is still rejected`).

## Trust model

Same scoping as `--federated`: explicit, per-source, source-owner opt-in.
The doc comments and CLI success message both call out that this should
never be set on a user-writable/PR-mergeable source, since it's still
possible for a mismatched slug to overwrite another page within that same
source's namespace — the opt-in narrows the anti-spoof hijack surface to
sources whose owner has declared frontmatter slugs authoritative, it doesn't
remove the surface globally.

## Testing

- New unit tests in `test/import-file.test.ts` (default-off regression, opt-in
  honored, error-message hint, no-mismatch-no-op case, leading-slash
  normalization), `test/sources-load.test.ts` (`isFrontmatterSlugTrusted`),
  `test/sources.test.ts` (CLI subcommand pair, config preservation, not-found
  exit code).
- `bun test test/import-file.test.ts test/sources.test.ts test/sources-load.test.ts`
  — 61 pass, 0 fail.

## Verification against a real source

Applied `gbrain sources trust-frontmatter-slug shaft-userguide` and ran
`gbrain sync --source shaft-userguide --full` against the real
`shafthq.github.io` checkout (189 markdown files). First attempt (before the
leading-`/` normalization) surfaced the `validateSlug` rejection immediately
via a live "Invalid slug" error on the very first batch of Docusaurus files —
fixed and re-verified: pages previously rejected as `SLUG_MISMATCH` (e.g.
`docs/testing/web.mdx` → `slug: /testing/web`) now import cleanly under
`testing/web`, no `Invalid slug` errors, only the expected transient
embedding-timeout/content-dedup skips unrelated to this change.

### DIFF
--- a/src/commands/import.ts
+++ b/src/commands/import.ts
@@ -144,6 +144,19 @@ export async function runImport(
       if (nudge) process.stderr.write(nudge + '\n');
     }
   }
+
+  // Per-source SLUG_MISMATCH escape hatch (`gbrain sources trust-frontmatter-slug`).
+  // Load once, thread to every per-file importFile call below — same
+  // load-once-per-command shape as importActivePack above. This is the path
+  // `performFullSync` (sync.ts `--full`) actually runs per-file through, so
+  // it must see the flag too, not just the incremental sync.ts call sites.
+  let trustFrontmatterSlug = false;
+  if (sourceId) {
+    const { fetchSource, isFrontmatterSlugTrusted } = await import('../core/sources-load.ts');
+    const src = await fetchSource(engine, sourceId);
+    if (src) trustFrontmatterSlug = isFrontmatterSlugTrusted(src.config);
+  }
+
   const workersIdx = args.indexOf('--workers');
   const workersArg = workersIdx !== -1 ? args[workersIdx + 1] : null;
   // v0.22.13 (PR #490 Q2): shared parseWorkers helper rejects bad input
@@ -239,7 +252,7 @@ export async function runImport(
       // unreachable when the gate is off; defense-in-depth check anyway.
       const result = isImageFilePath(relativePath) && process.env.GBRAIN_EMBEDDING_MULTIMODAL === 'true'
         ? await importImageFile(eng, filePath, relativePath, { noEmbed, sourceId })
-        : await importFile(eng, filePath, relativePath, { noEmbed, sourceId, activePack: importActivePack });
+        : await importFile(eng, filePath, relativePath, { noEmbed, sourceId, activePack: importActivePack, trustFrontmatterSlug });
       const _fileMs = Date.now() - _fileT0;
       if (_fileMs > 5000) {
         console.error(`[gbrain phase] import.process_file slow ${_fileMs}ms ${relativePath}`);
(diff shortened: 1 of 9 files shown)


MohabMohie:
Post-merge-readiness note from verifying this against the real shaft-userguide source: \`gbrain doctor\`'s \`frontmatter_integrity\` check still reports the 84 SLUG_MISMATCH files as a WARN, since it re-scans frontmatter independently of the sync/import gate this PR fixes. That's expected — this PR only changes the import-time anti-spoof decision, not the standalone lint/frontmatter-validate path, deliberately (frontmatter validate --fix must never run against a source like this; it would rewrite public routes). Flagging in case a follow-up to make \`frontmatter_integrity\` source-aware of \`trust_frontmatter_slug\` (skip/downgrade the warning for sources that opted in) is worth a separate small PR — didn't want to widen this one's scope.

time-attack:
Thanks for the detailed problem statement — the Docusaurus custom-routes case is a legitimate friction point and you documented it well. We're closing it anyway: this adds two permanent CLI subcommands and a per-source flag whose whole purpose is to switch off the path-vs-frontmatter anti-spoof check, and a per-source opt-out of an integrity guard is not surface we want to carry. The motivating case is a single external site, which doesn't outweigh weakening that check for everyone who enables it.

PR #2070, opened 2026-06-10
fix(rerank): classify missing auth before fallback

maxpetrusenkoagent:
## Summary

Fixes #2059.

This makes missing reranker credentials fail open with the correct `auth` reason instead of being recorded as `unknown` while search silently falls back.

Changes:
- convert reranker `AIConfigError` from missing provider credentials into `RerankError(..., auth)` before any HTTP call
- keep search reranker fail-open behavior, but audit missing `ZEROENTROPY_API_KEY` as `auth`
- make `doctor` warn when recent reranker failures are repeatedly `unknown`, preserving visibility for historical #2059 rows

## Notes on overlap

Checked open PRs for #2059 / reranker auth / `ZEROENTROPY_API_KEY`. Adjacent provider-auth PRs exist, but none cover this ZE reranker failure classification and doctor-health behavior.

## Verification

Ran on Mac mini (`Mac.attlocal.net`, user `maxsmacmini`):

- `bun test test/ai/rerank.test.ts test/search/rerank.test.ts test/doctor.test.ts`
- Result: 121 pass, 0 fail, 292 expect calls

GitHub may report no check suite for this fork draft PR; if so, the focused mini test run above is the current verification evidence.

Issue #2059, opened 2026-06-10
Reranker auth failures (missing ZEROENTROPY_API_KEY) recorded as reason 'unknown' — doctor reranker_health reports ok while every rerank fails


na-dev-12:
Found during a deep audit of gbrain v0.42.26.0.

## Symptom
With the default `balanced` mode (reranker on, `zeroentropyai:zerank-2`) and no ZEROENTROPY_API_KEY provisioned, **every** reranked search fails open. 32 identical entries accumulated in `~/.gbrain/audit/rerank-failures-*.jsonl` over two weeks ("ZeroEntropy reranker requires ZEROENTROPY_API_KEY.") while `gbrain doctor` kept reporting healthy. Autocut is also a silent no-op in this state (it gates on `rerank_score`, which never gets stamped).

## Root cause
- The missing-key error is thrown by `defaultResolveAuth` (gateway.ts ~291-297) as an **AIConfigError**, not a RerankError, so `rerank.ts` (~86-87) records `reason: "unknown"`.
- `checkRerankerHealth` (doctor.ts ~1405) only matches `reason === 'auth'` (and payload/transient buckets at ~1414/1423-1425); "unknown" falls through to status **ok** (~1434-1438).
- The check also never runs under `doctor --fast` (it needs `engine.getConfig`, doctor.ts ~1391), even though the audit JSONL it could read requires no DB access.

## Suggested fix
1. Map missing-key AIConfigError to `reason: 'auth'` (wrap as RerankError in gateway.rerank, or classify in rerank.ts).
2. Make `checkRerankerHealth` warn on repeated `reason: 'unknown'` failures instead of returning ok.
3. Surface rerank-failure counts in `--fast` mode from the audit JSONL.

herove:
Confirmed on 0.42.56.0: audit found **9,402 consecutive rerank failures** (since 06-07) all recorded as reason `unknown` — root cause was a missing ZeroEntropy key (auth 401), silently fail-open the whole time while `doctor reranker_health` stayed green. We only discovered it during a manual search-quality audit. +1 for classifying auth errors distinctly and surfacing consecutive-failure streaks in doctor.

Masashi-Ono0611:
[...]

PR #2901, opened 2026-07-17
fix(doctor): run reranker_health in --fast and DB-down paths

Masashi-Ono0611:
## Summary

Part of #2059 (item 3 of its suggested fix, plus a mode-resolution correctness fix found while reproducing). Composes with #2070, which covers the other two halves (AIConfigError→auth classification + the unknown-failure threshold); the two touch different hunks and rebase cleanly in either order.

**Fix 1 — the check never ran in degraded states.** The `reranker_health` check reads only the rerank-failure audit JSONL (`~/.gbrain/audit/rerank-failures-*.jsonl`) — no DB access — yet it lived inside the DB-checks phase, so `doctor --fast` and DB-down runs silently skipped it. Those are exactly the degraded states where an operator most needs to see that every rerank has been failing open.

**Fix 2 — enablement was read from the wrong source of truth.** The check read the raw `search.reranker.enabled` key and treated unset as disabled, but enablement actually resolves through the mode-bundle chain (per-key override > `search.mode` bundle > balanced fallback) and the balanced default turns the reranker ON. A default brain with an empty audit was told "Reranker disabled — no failures expected" while hybridSearch was reranking (or failing open) on every query. Now resolved via `resolveSearchMode(await loadSearchModeConfig(engine)).reranker_enabled` — the same pair hybridSearch uses — so doctor's interpretation matches the live search path.

Live reproduction that motivated this: a long-running production brain accumulated **950 rerank-failure events across 9 ISO weeks** (949 of them the same missing-key config error, recorded as `unknown` per #2059) while `doctor` reported `reranker_health → OK (below threshold)` the whole time and `search --explain` confirmed no `rerank_score` was ever stamped.

Changes:
- `src/commands/doctor.ts` — widen `checkRerankerHealth` to accept a null engine (neutral zero-failure message instead of claiming "disabled" from an unread key), call it inside the `fastMode || !engine` early-return block (full engined runs keep the original DB-phase call site + heartbeat, so check ordering and progress events are byte-identical for normal runs), and resolve enablement through the mode-bundle chain. Config-read failures fail open to bundle defaults by `loadSearchModeConfig`'s documented contract — the audit file is the core signal, and "disabled" is only ever claimed from a successfully-read explicit override.
- `docs/architecture/KEY_FILES.md` — rerank-audit entry updated to current state (`bun run build:llms` run).

## Test Coverage

- `test/doctor-behavioral.test.ts` — reranker_health present **exactly once** across full, `--fast`, null-engine `--fast`, and DB-down (no `--fast`) paths; guards the two call sites against double-firing.
- `test/doctor-reranker-health.test.ts` (new) — leaf coverage: null-engine empty-audit → ok without claiming "disabled"; null-engine auth failures still warn; explicit `false` override → disabled; **default brain (no keys) → enabled via balanced bundle** (the fix-2 regression); `search.mode=conservative` → disabled; throwing `getConfig` → neutral-ok on empty audit, still warns on real failures. Env-mutating (GBRAIN_AUDIT_DIR), so kept out of the concurrent-eligible behavioral file.

Local gates: targeted doctor/rerank/llms test files pass (20 leaf + behavioral), `bun run typecheck` clean, `bun run verify` 31/31 green.


time-attack:
Merging this one. I confirmed both halves against master: `checkRerankerHealth` reads the raw `search.reranker.enabled` key and treats unset as disabled while the balanced bundle (the default) has the reranker on, and the check only ran inside the DB-checks phase despite being purely file-based — so `--fast` and DB-down runs skipped it. Resolving enablement through the same chain hybridSearch uses, keeping the full-run call site unchanged, and pinning exactly-once across all four paths is the right shape. Thanks for the clean scoping and the leaf tests.

Masashi-Ono0611:
Thanks for merging, and for independently re-verifying the balanced-bundle default rather than taking the diff at face value — appreciate it.

PR #2904, opened 2026-07-17
fix(ai): route openai:-prefixed models with custom baseURL through createOpenAICompatible

qq369658058:
## Root cause
GBrain routes any `openai:`-prefixed model through Vercel AI SDK's `createOpenAI`. Third-party **OpenAI-COMPATIBLE** endpoints (e.g. NVIDIA `integrate.api.nvidia.com`) reject `createOpenAI`'s chat-completion request shape and return **HTTP 404 ("Not Found")**. This silently disables:
- **chat / think** — `instantiateChat` uses `createOpenAI` → 404
- **expansion** — `expand()` uses `generateObject` (strict `json_schema` / structuredOutputs) → 404

Embeddings are unaffected (separate `/v1/embeddings` path). `createOpenAICompatible` works fine on these endpoints (HTTP 200).

## Reproduction
```bash
export OPENAI_API_KEY="nvapi-..."          # NVIDIA key
export OPENAI_BASE_URL="https://integrate.api.nvidia.com/v1"
gbrain config set models.tier.deep openai:meta/llama-3.1-8b-instruct
gbrain think "what is the approval module"
# before: "no LLM available" + stderr "[ai.gateway] ... Not Found"
# after:  real synthesis, Model: openai:meta/llama-3.1-8b-instruct
```

## Fix
**`src/core/ai/gateway.ts`**
1. `instantiateChat` — when `resolveNativeBaseUrl('openai')` returns a custom baseURL, build the model via `createOpenAICompatible` instead of `createOpenAI`. Plain `api.openai.com` (no baseURL) is unchanged.
2. `expand()` — replaced `generateObject` (which forces `response_format: json_schema`) with a direct `fetch` to `/chat/completions` using `response_format: {type:"json_object"}` + manual `JSON.parse`. `createOpenAICompatible` + `generateObject` still fails on NVIDIA ("response did not match schema"), so the fetch path is required for expansion.

**`src/core/ai/recipes/openai.ts`**
- List `meta/llama-3.1-8b-instruct` (verified working on NVIDIA) so `gbrain think --model openai:meta/llama-3.1-8b-instruct` is accepted by the allowlist.

## Verification (NVIDIA integrate.api.nvidia.com, llama-3.1-8b-instruct)
- `think` → returns real synthesis with citation, `Model: openai:meta/llama-3.1-8b-instruct` ✅
- `expandQuery()` isolation → returns 3 variants (original + 2 rewrites) ✅
- `query` (hybrid + expansion) → top results unchanged, expansion no longer degrades ✅

## Notes
- `meta/llama-3.3-70b-instruct` returns 504 on NVIDIA's free tier — use `llama-3.1-8b-instruct`.
- Alternative considered: always `createOpenAICompatible` for `openai:` prefix. Rejected because it changes behavior for real `api.openai.com` users; the `if (baseURL)` guard keeps that path identical.

### DIFF
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -2235,23 +2235,52 @@ export async function expand(query: string): Promise<string[]> {
 
   try {
     const { model, recipe, modelId } = await resolveExpansionProvider(getExpansionModel());
-    const result = await generateObject({
-      model,
-      schema: ExpansionSchema,
-      // v0.42.20.0 (codex P0) — expansion had NO abortSignal; same stalled-socket
-      // class as chat. Default the chat timeout.
-      abortSignal: withDefaultTimeout(undefined, AI_CHAT_TIMEOUT_MS),
-      prompt: [
-        'Rewrite the search query below into 3-4 different, related queries that would help find relevant documents.',
-        'Return ONLY the JSON object. Do NOT include the original query in the result.',
-        'Each rewrite should emphasize different aspects, synonyms, or framings.',
-        '',
-        `Query: ${query}`,
-      ].join('\n'),
+    // FIX (local patch, 2026-07-17): gbrain routes `openai:`-prefixed models
+    // through Vercel AI SDK's `createOpenAI`, but third-party OpenAI-COMPATIBLE
+    // endpoints (notably NVIDIA integrate.api.nvidia.com) reject `createOpenAI`'s
+    // chat-completion request shape with HTTP 404 ("Not Found"), silently
+    // disabling expansion. `createOpenAICompatible` works, but the cleanest
+    // cross-provider fix is to call the endpoint's chat/completions directly
+    // with `response_format: {type:"json_object"}` (universally supported) and
+    // parse the JSON ourselves. This bypasses the AI SDK provider layer entirely.
+    const cfg = requireConfig();
+    const baseURL = (process.env.OPENAI_BASE_URL ?? process.env.NVIDIA_BASE_URL ?? cfg.env.OPENAI_BASE_URL ?? cfg.env.NVIDIA_BASE_URL ?? cfg.base_urls?.[recipe.id] ?? recipe.base_url_default ?? '').replace(/\/$/, '');
+    const apiKey = process.env.OPENAI_API_KEY ?? process.env.NVIDIA_API_KEY ?? cfg.env.OPENAI_API_KEY ?? cfg.env.NVIDIA_API_KEY ?? '';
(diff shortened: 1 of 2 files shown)


qq369658058:
## Verification notes

Tested on two independent setups, both on `garrytan/gbrain` @ `0.42.61.0`:

1. **Clean upstream clone** — `git clone`, applied these 3 changes, `bun install`, ran against NVIDIA `integrate.api.nvidia.com` with `meta/llama-3.1-8b-instruct`:
   - `think` → real synthesis with citation, `Model: openai:meta/llama-3.1-8b-instruct` ✅
   - `expandQuery()` in isolation → 3 variants (original + 2 rewrites) ✅
   - `query` (hybrid + expansion) → top results unchanged, expansion no longer silently degrades ✅

2. **Local `bun install -g` install** — same three edits applied to the installed `node_modules`, identical results.

### Why expansion uses a direct `fetch` rather than `createOpenAICompatible` + `generateObject`
This matters for reviewers. On NVIDIA's endpoint:
- `createOpenAICompatible` + `generateText` (plain chat) → **200 OK** ✅
- `createOpenAICompatible` + `generateObject` (schema / `json_schema`) → **fails**: *"No object generated: response did not match schema"*
- bare `fetch` + `response_format: {type:"json_object"}` → **200 OK** ✅

So `instantiateChat` can use `createOpenAICompatible` (chat works), but `expand()` must use the `json_object` fetch path because NVIDIA does not honor the structured-output schema shape `generateObject` emits. That's why the two fixes differ.

### Allowlist note
Without `meta/llama-3.1-8b-instruct` in the recipe, `gbrain think --model openai:meta/llama-3.1-8b-instruct` is rejected by the allowlist ("not listed for OpenAI chat"), even though `models.tier.deep` config宽松放行. Adding it to the recipe lets both invocation styles work. Happy to move this to a separate `nvidia` recipe if you'd prefer not to widen the `openai` recipe.

time-attack:
Thanks for the detailed repro against the NVIDIA endpoint — the createOpenAI vs createOpenAICompatible distinction for third-party bases is a real observation. We can't take this as written: it rewrites `expand()` for every user (including plain api.openai.com) into a hand-rolled fetch with manual JSON parsing and NVIDIA-specific env sniffing, dropping the SDK's schema validation from a core search path, and it lists a NVIDIA-hosted Meta model in the `openai` recipe's allowlist. The comments also mark it as a local patch, and there are no tests. Closing as vendor-specific plumbing plus too much risk to a shared path.

Issue #1163, opened 2026-05-18
serve: terminal-launched processes accumulate across sessions, exhaust Supabase session pool over days


colin477:
## Summary

Each invocation of \`gbrain serve\` from an interactive terminal spawns a long-lived bun process that holds its own Postgres client pool. When the user closes the parent terminal (iTerm pane, tmux pane, VS Code integrated terminal, etc.) without explicitly \`Ctrl-C\`-ing or \`exit\`-ing first, **the serve process is not killed** — it orphans and stays alive indefinitely, still holding open Postgres connections.

Over days of normal operation (open new terminals, launch \`gbrain serve\`, close terminals later) the user's machine accumulates 4-6 orphaned serve processes. With Supabase's free-tier session pooler capped at 15 client connections, this saturates the pool and produces \`(EMAXCONNSESSION) max clients reached in session mode\` errors on legitimate MCP and CLI calls.

## Snapshot from my machine just now

```
$ ps aux | grep 'bun.*gbrain.*serve' | grep -v grep
  7541   Sun06AM   bun run /Users/me/Developer/gbrain/src/cli.ts serve
14092   Mon10:14AM bun run /Users/me/Developer/gbrain/src/cli.ts serve
41783   Sat04PM   bun run /Users/me/Developer/gbrain/src/cli.ts serve
59980   Fri08AM   bun run /Users/me/Developer/gbrain/src/cli.ts serve
70116   Sat05PM   bun run /Users/me/Developer/gbrain/src/cli.ts serve
```

Five serves spanning four days, only one of which (the Monday 10:14AM one) was attached to an active terminal. The other four are orphans from closed iTerm panes. Each was holding active TCP sessions toward \`aws-0-us-west-2.pooler.supabase.com:5432\`.

## Repro

1. Open terminal A, run \`gbrain serve &\` (or just \`gbrain serve\` in a tmux pane).
2. Note the serve PID.
3. Close terminal A's window without \`exit\` or \`Ctrl-C\` first (just close the iTerm tab / X-button the window).
4. \`ps aux | grep gbrain.*serve | grep <pid>\` — serve still running.
5. Repeat in 4 more terminals over a few days.
6. \`(EMAXCONNSESSION) max clients reached in session mode - max clients are limited to pool_size: 15\` shows up on subsequent legitimate calls.

## Diagnosis

[...]

Issue #446, opened 2026-04-26
Feature request: `gbrain serve` idle timeout (auto-exit after inactivity)


qixk:
Hi Garry,

I'm running gbrain 0.16.4 as an MCP server inside OpenClaw with multiple agents. Love the tool — the PGLite + vector search combo is exactly what I need.

**Problem I'm hitting:**

When `gbrain serve` is spawned as a stdio MCP server, each agent session forks a new `gbrain serve` process. The issue is that these processes never exit on their own — even after the parent session ends or the gateway restarts. Over the course of a day, I ended up with **41 zombie `gbrain serve` processes** all competing for the same PGLite database, which caused connection timeouts (30s) for new sessions.

**What I'd love to see:**

An idle timeout option for `gbrain serve`, similar to how Ollama handles model unloading with `OLLAMA_KEEP_ALIVE`:

```bash
# Exit automatically after 5 minutes of no MCP requests
gbrain serve --idle-timeout 300

# Or via config
gbrain config set serve.idleTimeout 300
```

The idea is simple:
1. `gbrain serve` starts → begins accepting MCP requests via stdio
2. After each request completes → reset an idle timer
3. If no requests come in for N seconds → gracefully exit (release PGLite locks, free memory)
4. The MCP host (OpenClaw/Claude Desktop/etc.) will fork a fresh instance when needed

This follows the same pattern as Ollama's resource management — the caller uses it, and when it's done, the resource gets released. If the caller forgets to clean up, the service cleans up after itself.

**Current workaround:**

I wrote a cron job that kills `gbrain serve` processes older than 30 minutes. Works but it's a band-aid.

**Environment:**
- gbrain 0.16.4, PGLite engine
- OpenClaw 2026.4.22, 14 agents, each can trigger MCP
- macOS, Apple M1 Max, 32GB RAM

Thanks for building gbrain — it's become core infrastructure for my AI team. Happy to test if you ship a beta with this feature.

PR #2905, opened 2026-07-17
feat(doctor): serve_process_accumulation check — surface leaked stdio serves

Masashi-Ono0611:
## Summary

Part of #1163 (visibility half). The `--stdio-idle-timeout` escape hatch shipped for #446, but it defaults OFF — so operators only learn it exists after leaked serves have already wedged their session pool. This adds the signpost: a `serve_process_accumulation` doctor check that counts live `gbrain serve` stdio processes and, at >=3 concurrent, warns with every PID + ELAPSED and the flag hint.

Live reproduction that motivated this: 5 concurrent stdio serves on one host (oldest up 1d9h, ~100MB RSS each, one held DB connection each), every client long gone — the clients were tailscale-ssh MCP sessions whose laptop slept, leaving the ssh half-open so stdin never delivered EOF. Discovered only by manually running `ps` while chasing an unrelated issue; doctor had nothing to say.

Design:
- **Detection** — `ps -axww -o pid=,uid=,etime=,command=` scan, scoped to the current uid (teammates' legitimate serves on a shared host are not accumulation). `-ww` matters: procps may clip COMMAND at 80 columns under redirection, truncating long bun-install paths before the `serve` token and silently undercounting. `serve --http` daemons are excluded — a long-lived HTTP server is the intended topology, not a leak. The match tolerates global flags (`gbrain --quiet serve`).
- **Fail-open** — Windows returns ok/not-applicable; any ps/parse error returns ok. Informational check; must never fail doctor on an exotic ps.
- **Bounded** — production scans are memoized for 30s so unrated remote `run_doctor` calls cannot fan out `ps` children.
- **Surfaces** — `buildChecks`' filesystem phase (runs under `--fast` and DB-down: pool exhaustion is exactly the state where doctor gets run degraded) and `doctorReportRemote` (MCP-heavy operators are the accumulation-prone ones). Registered under `ops` in `doctor-categories.ts`.

## Test Coverage

- `test/doctor-serve-accumulation.test.ts` (new) — pure parser against a live-shaped ps fixture from the real incident (stdio serves incl. global-flag and spaced-path forms matched; `--http`, `serve-http`, doctor, editors, wrapper scripts excluded; uid filter), ELAPSED parsing across all shapes, threshold + message logic (every PID listed, oldest called out, flag hint present), and the check wrapper through the `PsRunner` seam (warn path, fail-open path, module-level seam).
- `test/doctor-behavioral.test.ts` — exactly-once presence across full, `--fast`, and DB-down paths; ps stubbed via `_setServePsRunnerForTests` so real serves on the CI host cannot flake the orchestrator suite.
- `test/doctor-report-remote.serial.test.ts` — presence pinned on the remote surface; same stub for hermeticity.

Local gates: doctor test files 96+34 pass, `bun run typecheck` clean, `bun run verify` 31/31 green, live ok-path verified via `bun run src/cli.ts doctor --fast`.


Masashi-Ono0611:
Pushed one more commit — the check's remediation text was wrong in the case that actually shows up.

**What happened.** I've been running this check on my own install (borrowed onto a real brain) and it fired: three concurrent stdio serves, oldest alive 24h. The message told me to `launch with --stdio-idle-timeout <seconds>` — but my launcher already sets `--stdio-idle-timeout 14400` on all three. The one actionable sentence was the one that didn't apply.

**Why that case is the interesting one.** Reading `serve.ts`, the idle timer is re-armed on every stdin `data` chunk. So a process that outlives its window didn't ignore the flag — it kept *receiving data*. That points at the client holding the pipe open, which is a completely different investigation from "you forgot the flag". The message now says that instead.

The check already parsed each process's full command line for the PID/ELAPSED list, so this is a branch on data it had, not a new scan.

**One correctness trap, caught in review.** My first pass just grepped for the flag name. That is wrong twice over, and both spellings look enabled:

- `--stdio-idle-timeout=900` — `parseStdioIdleTimeout` does `args.indexOf('--stdio-idle-timeout')` and reads the *next argv slot*, so the equals form is not the flag at all.
- `--stdio-idle-timeout 0` — parses fine, but the timer is only installed when the value is `> 0`; zero is the documented off switch.

Claiming "the timer was armed and it still survived" for either would send an operator chasing a client that was never being timed. `hasArmedIdleTimeout` now mirrors `parseStdioIdleTimeout`'s accepted shape and requires a positive value, with a leading `(?:^|\s)` so a superstring like `--not-really--stdio-idle-timeout 900` can't match. It stays explicitly heuristic — `ps` flattens argv, so exact boundaries aren't recoverable.

**Tests:** 14 pass, exit 0; `typecheck` clean. New cases cover all-armed, mixed, the equals form, an explicit `0`, and the two superstring shapes. I also mutation-tested the branch (forcing the matcher to never match makes the new assertions fail, exit 1) rather than trusting a green run.

Happy to drop this back out if you'd rather keep the original PR minimal and land the wording separately.

time-attack:
Thanks for this, and the 5-concurrent-serves reproduction is a real operational annoyance. Closing anyway: this is new diagnostic feature surface rather than a fix — the `--stdio-idle-timeout` remedy already exists and works, and this adds a process-enumerating check plus a hardcoded concurrency threshold purely for discoverability. New doctor checks are a category we're deliberately not growing right now, independent of the quality of the patch.

Masashi-Ono0611:
Understood, thanks for the clear reasoning — makes sense to hold the line on new doctor-check surface generally, independent of this one's quality. The existing `--stdio-idle-timeout` remedy stands as the documented answer for anyone who hits the same pileup.

PR #2907, opened 2026-07-17
feat(recipes): add opencode-go recipe for OpenCode Go provider

barnaby-buttles:
## Summary

Add an openai-compat recipe for [OpenCode Go](https://opencode.ai/go/) — a $10/mo subscription providing curated open models via an OpenAI-compatible endpoint at `https://opencode.ai/zen/go/v1`.

Models included: `deepseek-v4-flash`, `deepseek-v4-pro`, `mimo-v2.5`, `mimo-v2.5-pro`, `glm-5.2`, `glm-5.1`, `kimi-k2.7-code`, `kimi-k2.6`, `qwen3.5-plus`. See the [OpenCode Go docs](https://opencode.ai/docs/go/) for current model list, pricing, and context limits.

## Changes

- **New:** `src/core/ai/recipes/opencode-go.ts` — recipe with chat touchpoint, `reasoning_content` compat fetch (same pattern as deepseek.ts)
- **Modified:** `src/core/ai/recipes/index.ts` — import + register
- **Modified:** `src/core/config.ts` — `GBrainConfig.opencode_go_api_key` field + `loadConfig` env merge
- **Modified:** `src/core/ai/build-gateway-config.ts` — env mapping for gateway
- **Modified:** `src/core/ai/capabilities.ts` — added opencode-go to known providers list
- **New:** `test/ai/recipe-opencode-go.test.ts` — 7 smoke tests (all pass)

## Usage

```bash
export OPENCODE_GO_API_KEY=sk-...
gbrain config set models.tier.reasoning opencode-go:deepseek-v4-flash
gbrain providers test --touchpoint chat --model opencode-go:deepseek-v4-flash
```

## Notes

- `agent.use_gateway_loop=true` is required for the subagent tool loop to dispatch on non-Anthropic models.
- Models served via the Anthropic Messages shape (`qwen3.6-plus`, `qwen3.7-plus`, `qwen3.7-max`, `minimax-*`) are NOT included — they would need a separate native-anthropic recipe.
- The `reasoning_content` compat fetch handles reasoning models that return chain-of-thought in a separate field (DeepSeek, Qwen, MiMo).

## Tests

```bash
bun test test/ai/recipe-opencode-go.test.ts
# 7 pass, 0 fail, 17 expect() calls
```

### DIFF
--- a/src/core/ai/build-gateway-config.ts
+++ b/src/core/ai/build-gateway-config.ts
@@ -38,6 +38,7 @@ export function buildGatewayConfig(c: GBrainConfig): AIGatewayConfig {
   // config.json) must reach the openrouter recipe's OPENROUTER_API_KEY.
   // process.env still wins via the later spread.
   if (c.openrouter_api_key) envFromConfig.OPENROUTER_API_KEY = c.openrouter_api_key;
+  if (c.opencode_go_api_key) envFromConfig.OPENCODE_GO_API_KEY = c.opencode_go_api_key;
 
   // v0.32 codex finding #4+#5 fix: thread local-server _BASE_URL env vars
   // into base_urls so the gateway hits the user's configured port. Without
--- a/src/core/ai/capabilities.ts
+++ b/src/core/ai/capabilities.ts
@@ -77,7 +77,7 @@ export function getProviderCapabilities(modelString: string): ProviderCapabiliti
   if (!chat) {
     throw new AIConfigError(
       `Provider "${recipe.id}" does not offer a chat touchpoint.`,
-      `Known providers with chat: openai, anthropic, google, openrouter, litellm-proxy, deepseek, groq, together, azure-openai, dashscope, minimax, zhipu, ollama, llama-server. Pick one for models.tier.subagent.`,
+      `Known providers with chat: openai, anthropic, google, openrouter, litellm-proxy, deepseek, groq, together, azure-openai, dashscope, minimax, zhipu, ollama, llama-server, opencode-go. Pick one for models.tier.subagent.`,
     );
   }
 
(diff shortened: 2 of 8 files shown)


time-attack:
Thanks for putting this together, but closing: wiring a third-party commercial subscription service into the gateway is a vendor integration, which is a category we don't accept regardless of implementation quality. It also adds nine externally-controlled model entries to `model-pricing.ts`, which is the repo's single canonical pricing table — that list would drift out of date the moment the provider changes its lineup, and we'd own the maintenance. No reflection on the patch itself.

DiscountDarcy:
Hi @time-attack Would you be able to elaborate on this 'category we don't accept' statement?

gbrain already has recipes for:

• openrouter.ts · Third-party aggregator of other providers' models
• groq.ts · Commercial inference service
• together.ts · Commercial inference service
• dashscope.ts · Alibaba's commercial API
• minimax.ts · Commercial service
• moonshot.ts · Commercial service
• nvidia.ts · Commercial service
• perplexity.ts · Commercial service
• voyage.ts · Commercial embedding service

OpenRouter's recipe also maps to the `same model-pricing.ts` and I'm not sure how you're any more insulated from their pricing than you would be from Opencode Go's.

You could certainly make the case that those options above are larger and/or simply more significant such that the burden of ownership in the codebase requires a bar to clear that is higher than what Opencode Go can clear, and/or that Gbrain is opinionated in that recipes are implicitly endorsements and a smaller aggregator doesn't make the cut. 

Having said that, Opencode Go is consistently one of the top recommendations for upstream providers for Hermes in every community source I've seen and it is listed in https://hermes-agent.nousresearch.com/docs/integrations/providers .

But 'we have better things to do than monitor every LLM aggregator out there' is a good argument, so perhaps a 3rd party plugin architecture might be the solution? You might not want to maintain this recipe, but I sure as heck do. :) 

Thanks,
DD (and barnaby)

PR #2910, opened 2026-07-17
fix(doctor): content-sanity audit scores new inflow, not chronic re-warns

Masashi-Ono0611:
## Summary

Part of #1893 (the recurring-FAIL half). `content_sanity_audit_recent` scored raw 7-day event volume, but the ingest gate intentionally re-logs every re-import — so a corpus with standing oversized pages (nightly transcript syncs) re-emits the same events every night and the check pins at warn/fail forever, drowning what the stream is actually good at: showing NEW inflow.

Live reproduction that motivated this: a personal brain with nightly conversation-transcript ingestion showed `2999 events ... warn=2999` at doctor — behind that number were 445 distinct pages, 434 of them the *same* chronic offenders re-warned every single night for 12+ days, and 11 genuinely new ones. The 2999 was noise; the 11 was the signal.

Design:
- **Chronic vs new signatures** — `summarizeContentSanityEvents` now splits distinct `(source_id, slug, event_type, reason_key)` signatures into chronic (events span >= 3 distinct UTC days inside the window — nightly re-imports cross this in 3 days, same-day retries never do) vs new. Timestamps normalize through `Date.toISOString()` so offset timestamps can't fabricate extra days; unparseable ones never count toward chronicity (fail-safe: unknown recency scores as new). Keys are JSON tuple encoded so an operator literal containing a delimiter can't merge two signatures into a false chronic one.
- **Scoring (extracted pure `computeContentSanityAuditCheck`, exported for tests)** — hard dispositions still always fail, chronic or not (content being blocked/hidden is never routine). New soft/flag pages warn. New warn-type pages warn at the same 10-unit threshold with the unit corrected from raw events to distinct pages, so one page re-imported 10 times in a day no longer trips it alone. `bypass_active` events always at least warn — the kill-switch is an incident-time escape hatch and chronicity must never bury a gate that is switched off.
- **Chronic-only streams score ok** — deliberately. Standing state is already owned, deduplicated, by the DB-backed checks (`oversized_pages` / `flagged_pages` / `quarantined_pages`); this check becomes the fresh-inflow sensor. The tradeoff this accepts: warn-tier pages between `bytes_warn` and `bytes_block` have no *scored* standing-state check once chronic — they remain visible in every doctor run via the message (`chronic=N` + top offenders named with event/day counts), which is exactly the "operator has acknowledged the pages are intentionally long" state #1893 asks for. If a scored standing check for the warn tier is wanted, that's a natural follow-up on the `oversized_pages` side.
- **Logging side untouched** — re-import re-logging is intentional ("the row count over time IS the signal"); this change finally *uses* the time dimension instead of being defeated by it.

Verified against the real 12-day incident corpus (copied JSONLs, window pinned to the last event): `2999 events -> 445 pages (chronic=434, new=11)`, top 3 offenders named with `7 events/7d` each, status computed from the 11 new pages rather than the 2999 raw events. A static acknowledged corpus like the one in #1893 scores ok; a genuinely new junk-source explosion still warns/fails.

## Test Coverage

- `test/audit/content-sanity-audit.test.ts` — chronic threshold boundaries (3 distinct days chronic; 2 days + same-day retries stay new), malformed-ts fail-safe, offset-timestamp UTC normalization + invalid calendar dates, signature granularity (event_type escalation counts as new; sorted-unique reason names; comma-containing literal can't collide with a multi-name set), per-source page distinctness, bypass counting, top_chronic ordering/cap.
- `test/doctor-content-sanity-check.test.ts` (new) — status matrix through the real summarizer: chronic-only ok + message names the standing-state checks, one new soft page warns over a chronic backdrop, threshold edge (9 new pages ok / 10 warn), 10 same-day re-imports of one page ok, chronic hard dispositions still fail, single AND chronic bypass warn, Top-chronic callout omitted when none.

Local gates: 29 targeted tests pass, `bun run typecheck` clean, `bun run verify` 31/31 green, `bun test test/doctor-behavioral.test.ts` green, `bun run build:llms` regenerated after the KEY_FILES.md update.


time-attack:
Thanks for the detailed live reproduction — the 434-chronic vs 11-new breakdown is a genuinely good observation about what the stream is telling you. I'm closing this one on surface-area grounds rather than on the analysis: it isn't a correctness bug (the check reports what it measures), and the fix redefines the metric by adding ~400 lines of new chronicity-classification heuristics — distinct-signature keying, a 3-distinct-UTC-day threshold, timestamp-normalization rules — to a health-score input. That's tunable heuristic surface on a reporting path, which is the kind of growth we're holding the line on right now.

Masashi-Ono0611:
Makes sense — appreciate you engaging with the 434-chronic-vs-11-new breakdown even while declining the heuristic surface it'd take to formalize as a metric. Thanks for the review.

Issue #1950, opened 2026-06-07
sync --source default wedged ~29min holding gbrain-sync lock; status reports 'idle' while proc live; no way to reap without pkill


garrytan-agents:
## Summary

A `sync --source default` process wedged for **~29 minutes** (1764s elapsed, 23.8% CPU, never exiting) while **holding the `gbrain-sync:default` lock**, blocking all dependent commit-hook syncs. It had to be killed by hand (`pkill -f "src/cli.ts sync --source default"`) and relaunched. The bookmark (`last_sync`) was stuck at `2026-06-04 03:39:13` the entire time and never advanced — the source showed `idle` in `sources status` while a sync process for it was still live, which is itself a state inconsistency.

## Observed

- `ps`: `579638  1764  23.8  bun run src/cli.ts sync --source default` — alive, burning CPU, not progressing.
- `sources status`: `default  3d  98%  idle  ...  last_sync 2026-06-04 03:39:13` — reports **idle** despite a running sync proc for that source.
- Commit hooks that call sync (brain post-commit) repeatedly hit: `gbrain sync in flight — waiting up to 90s` → timeout → `committing anyway (drift risk low)`. Several commits in a row had to override the wait.
- No lock file present under `/data/.gbrain/locks/` to inspect/clear manually — the lock appears to be DB/advisory-based, so a hung holder can't be cleared without killing the PID.

## Root cause (suspected)

This is the same source whose bookmark has been stuck since 2026-06-04 due to importer crashes on a few apple-notes files (`opts.title.toLowerCase is not a function`, `content-sanity.ts:379` — see #1939). The full-rewalk drain over ~204K files at ~2 files/sec (PgBouncer transaction-mode disables prepared statements) runs for a very long time, and **one of these long-running drains hung** rather than completing or erroring. Because the bookmark never advances past the crashing files without `--skip-failed`, normal syncs re-walk the whole corpus every time, maximizing the window in which a hang can occur while the lock is held.

## Asks

1. **Lock holder liveness / heartbeat + auto-reap.** If the lock holder hasn't advanced its bookmark or emitted progress in N minutes, it should self-abort and release the lock (or a watchdog should reap it). A single wedged sync should not be able to hold `gbrain-sync:<source>` indefinitely and starve commit hooks.
2. **`sources status` should not report `idle` while a sync process for that source is live.** Either reflect `running`/`stuck`, or detect the orphan. The idle-while-running inconsistency made diagnosis harder.
[...]

Issue #1737, opened 2026-06-01
Minions: long-running jobs (subagent/embed-backfill/autopilot-cycle) thrash on wall-clock timeout with broken attempt accounting, starving the queue


garrytan-agents:
## Summary
The shell-job lane is rock solid (38/38 done, ~11s avg). But the **long-running job lanes** — `subagent`, `embed-backfill`, `autopilot-cycle` — thrash and die without ever consuming an attempt, and pile up `waiting` behind them.

## Evidence (observed on a live worker: `gbrain jobs work --concurrency 3 --queue default --max-rss 16384`)

`jobs get` on dead jobs:
```
Job #11385: autopilot-cycle (DEAD after 0 attempts)
  Attempts: 0/2 (started: 3)
  Error: wall-clock timeout exceeded

Job #11383: embed-backfill (DEAD after 0 attempts)
  Attempts: 0/3 (started: 1)
  Error: wall-clock timeout exceeded
  Data: {"reason":"sync_all","sourceId":"straylight-brain","batchSize":500}
```

The smoking gun is **`Attempts: 0/2 (started: 3)`** — the job was *started* 3 times but the attempt counter never incremented. Pattern: worker claims the job (`started++`), the job runs long and exceeds the wall-clock, the lease is killed mid-flight, and the **attempt is never recorded as consumed**, so it gets reclaimed and re-run until it exhausts the `started` ceiling -> DEAD. Meanwhile a freshly-submitted `subagent` job (#11407) sat at `Attempts: 0/3 (started: 0)` — never claimed, because the thrashing long jobs monopolize the 3 concurrency slots.

## Impact
- The subagent (LLM) lane is effectively unusable for batch fan-out: jobs either never get claimed or die on wall-clock without making progress.
- Queue starvation: 5+ jobs stuck `waiting`, oldest 5 days old (`protected` #10506 from 2026-05-27).
- The doctor hint exists (`no subagent jobs completed — cap may be too tight; export GBRAIN_ANTHROPIC_MAX_INFLIGHT=64`) but the root issue is lease/attempt accounting, not just the inflight cap.

## Expected
- A job killed by wall-clock timeout should **increment attempts** (it consumed an attempt), not leave `attempts:0 / started:N`.
- `started >> attempts` should be impossible, or surfaced as a distinct "lease-lost / reclaim thrash" state instead of silently looping to DEAD.
[...]

PR #2911, opened 2026-07-17
fix(sync): thread AbortSignal through importFile — reap in-file wedges, not just detect them

Masashi-Ono0611:
## Summary

Part of #1950 — closes the follow-up TODO the stall watchdog shipped with ("thread a cancellation signal through importFile"). The watchdog composes its abort into `opts.signal`, but the drain only observed it BETWEEN files: a hang inside one `importFile` call — the documented incident shape, a stuck embed network request — ran to completion or the wall-clock hard deadline. The watchdog could detect the wedge but not reap it.

Design:
- **All three import entry points accept `signal?: AbortSignal`** (`importFromFile` / `importFromContent` / `importCodeFile`), observed via the canonical `throwIfAborted` (`abort-check.ts`, #1737) at the async phase boundaries the TODO named: entry/post-parse, pre-embed, pre-DB-write. Never inside the transaction — a file lands atomically or not at all, so cancellation can't produce partial page state.
- **The signal rides into the gateway call.** `embedBatch` already had an `abortSignal` passthrough to the embedding gateway; the import path just never used it. With it threaded, the in-flight embed HTTP request itself aborts — the wedge is reaped, not merely detected.
- **The signal, not the exception name, is the cancellation authority.** A provider-generated `AbortError` (SDK-internal timeout, our signal still live) is a real failure: the drain records it in the failure ledger like any other error, `importCodeFile`'s embed-failure swallow keeps its vectorless-landing policy for it (re-raising via `throwIfAborted` only when we were actually told to stop), and the rename loop only exits `partial('timeout')` on a genuine cancel. Without this, an SDK-shaped abort would make a file silently vanish from the sync — no checkpoint, no ledger row.
- **Abort vs checkpoint/ledger:** an abort escaping `importFile` in the drain is a clean return — no failure-ledger entry (which would auto-skip an innocent file after enough cancelled runs), and the path stays un-checkpointed so the resumed sync retries it. In the rename loop the rename itself has already landed and the path is not checkpointed; the resumed run re-imports it.
- **Unset signal → byte-identical behavior.** Every check is a no-op without a signal; MCP `put_page` and other `importFromContent` callers are untouched.

Deliberately out of scope (follow-ups): the rename loop runs before the stall watchdog is composed, so renames are covered by user `--timeout`/SIGINT but not the stall watchdog — extending the watchdog's progress-key semantics to the rename phase is a separate change to #1950's drain-scoped design. Full-sync (`performFullSync` → `runImport`) still takes no signal at all; that is the pre-existing `runImport` TODO and is unchanged here (the plumbing this PR adds to `import-file.ts` is the reusable half of that work).

## Test Coverage

- `test/import-file-abort-signal.test.ts` (new) — pre-aborted signal throws before any DB write on all three entry points (no `putPage` recorded); signal firing mid-flight (during `getPage`, after the post-parse check passed) is caught at the next boundary with no DB write; the code path's abort escapes the embed-failure swallow (exercising the catch's signal-authority re-raise) instead of half-landing vectorless; unset and live-but-unfired signals import exactly as before, and the vectorless-landing path survives for real embed failures.
- Honest gaps: no test drives a real in-flight gateway abort (would need a network-level fixture), and the drain-level checkpoint/ledger interaction is asserted by review rather than an e2e harness (none exists today for the stall watchdog either).

Local gates: 8 new tests + 90 adjacent import/sync tests pass, `bun run typecheck` clean, `bun run verify` 31/31 green, `bun run build:llms` regenerated after the KEY_FILES.md update.


time-attack:
Merging. Master's `import-file.ts` has no signal plumbing at all, so the stall watchdog can only detect an in-file wedge and never reap it — exactly the gap CLAUDE.md documents. Threading the existing composed signal into the existing `embedBatch({abortSignal})` passthrough is the minimal fix, and the two details that matter are both handled correctly: no abort check inside the transaction, and the signal rather than the exception name as cancellation authority so a provider-side AbortError still lands in the failure ledger. Thanks for the careful writeup on the ledger semantics.

Masashi-Ono0611:
Thanks for merging — glad the minimal scope (existing composed signal, existing `embedBatch({abortSignal})` passthrough) held up to review.

Issue #1422, opened 2026-05-25
gbrain dream silently fails all DB phases on PostgreSQL — connectEngine() swallowed by try/catch in cli.ts


foxhoundinc:
Version: 0.41.2.0
Schema: v94
Database: PostgreSQL 18.4 (local, pgbouncer-style connection)
OS: CachyOS Linux
Install path: ~/skynet/hermes/gbrain-latest/

Summary
gbrain dream runs successfully against a PostgreSQL brain but silently fails every DB-dependent phase. Only local file-scanning phases (lint, backlinks, extract) execute. All other phases — sync, extract_facts, patterns, consolidate, propose_takes, grade_takes, calibration_profile, embed, orphans, purge, extract_atoms, synthesize_concepts — return "No database connection."
gbrain doctor works correctly and connects to the same PostgreSQL instance without issue.

Root cause identified
In src/cli.ts lines 1007–1024, connectEngine() is wrapped in a try/catch that silently swallows failures:
typescriptlet eng: BrainEngine | null = null;
try { eng = await connectEngine(); }
catch { /* DB unavailable — lint + backlinks still run */ }
When connectEngine() throws, eng stays null and the dream cycle proceeds with no database. All DB-dependent phases fail silently with "No database connection: connect() has not been called."
gbrain doctor uses a different initialization path that succeeds. The dream path's connectEngine() → loadConfig() → createEngine() → connectWithRetry() chain is throwing somewhere but the error is swallowed.

Evidence DB is healthy
bashgbrain doctor
 3,125 pages, schema v94, 100% embedded, health 70/100
 PostgreSQL connected, all checks pass

Direct Bun test also confirmed connection:
import createEngine, connectWithRetry → returned full stats cleanly

Steps to reproduce

Fresh install of v0.41.2.0 against PostgreSQL (not PGLite)
Confirm gbrain doctor returns healthy
Run gbrain dream --json
Observe: lint, backlinks, extract complete (0 results — local file ops only)
Observe: all DB phases return "No database connection"
Brain score unchanged, atoms 0, concepts 0, cost $0

Additional context

GBRAIN_DATABASE_URL exported in gbrain wrapper — no effect
gbrain init --postgres run to ensure ~/.gbrain/config.json has correct URL — no effect
Stale gbrain_cycle_locks rows cleared manually between attempts — no effect
4 separate dream cycle attempts, all same result
gbrain jobs stats connects to DB fine
gbrain schema active connects to DB fine
[...]

PR #2912, opened 2026-07-17
fix(dream): honor the #2084 teardown contract — no direct process.exit in dream.ts

Masashi-Ono0611:
## Summary

Part of the command-module exit sweep the #2084 cross-model review filed (F2): `dream.ts` carried 23 direct `process.exit` sites, and every one of them preempts cli.ts's `finally { finishCliTeardown(...) }` — skipping the background-work drain, bounded disconnect, and stdout flush. The TODO names dream as the data-loss-prone module: a path exiting mid-cycle discards in-flight facts/search-cache writes, and `--json` through a slow pipe keeps the #1959 truncation exposure. Worst on the overnight cron, where nobody is watching.

Design:
- **A module-local sentinel, not a shared helper.** Every `process.exit(n)` becomes `bail(n)`, which throws `DreamExit(n)`. The public `runDream` is now a thin wrapper over `runDreamInner` that catches ONLY the sentinel, records the verdict via `setCliExitVerdict(n)`, and returns — cli.ts's `finally` runs `finishCliTeardown`, then the central seam (`flushThenExit(currentExitCode())`) exits with the same code the site asked for. Anything else propagates with its stack, unchanged.
- **Exit-code fidelity.** All existing codes are preserved: usage errors 2, failures 1, `EXIT_DRAIN_INCOMPLETE` 3. The intermediate catches in the module (`LockUnavailableError` in the drain, the resolver-user-error catch) rethrow non-matching errors, so the sentinel passes through untouched — including when `bail` fires from inside those catch blocks.
- **Tests stop mocking a global.** Exit-behavior tests move from spying on `process.exit` (and asserting a thrown `'EXIT'` string) to asserting `currentExitCode()` + resetting via the existing `_resetCliExitVerdictForTests` seam. Two new subprocess tests drive the real `cli.ts` dispatch end-to-end (hermetic `GBRAIN_HOME`, same unreachable-postgres trick as the #1422 test) and pin that usage errors still terminate the process with exit 2 / 1 — proving the code survives the teardown seam, not just the function contract.

Scope: dream.ts only — the module the TODO singles out for the in-flight-write consequence. The `doctor.ts` (~10 sites incl. the verdict exit), `ze-switch.ts` (~9), and cli.ts friction/claw-test verdict exits from the same TODO entry stay as follow-ups; each wants its own review per the TODO's framing.

Honest test-coverage note: the three `EXIT_DRAIN_INCOMPLETE` paths are not individually pinned at the CLI layer (they need a seeded extract-atoms backlog; the drain core is already dep-injected and covered by `test/extract-atoms-drain.test.ts`). They use the identical `bail` mechanism the pinned paths do.

## Test Coverage

- `test/dream.test.ts` — every exit-path test migrated to the verdict contract (`currentExitCode()`), including usage errors (2), engine-null/unknown-source/archived-source guards (1), unknown phase (1), and the both-null hard error; the clean/ok/partial no-exit test keeps its `process.exit` spy (restored properly) and now also asserts verdict 0.
- `test/dream-postgres.serial.test.ts` — both-null exit-1 assertions migrated the same way.
- `test/cli-dream-engine-warn.test.ts` — new `#2084` describe: two spawnSync tests through the real dispatch pin final process exit codes 2 (bad `--date`) and 1 (unknown `--phase`) with their stderr messages.

Local gates: dream suites 29/29 + engine-warn 4/4 pass, `bun run typecheck` clean, `bun run verify` 31/31 green, `bun run build:llms` regenerated after the KEY_FILES.md update.


time-attack:
Merging. Master still carries ~24 direct `process.exit` sites in `dream.ts`, every one of which preempts `finishCliTeardown` — the drain, bounded disconnect, and stdout flush — which is real data loss and `--json` truncation on the unattended overnight cron. The module-local `DreamExit` sentinel with a single catch in the `runDream` wrapper is the right scope (no new shared helper), exit-code fidelity is preserved, and moving the tests off `process.exit` spying onto verdict assertions is a genuine improvement. Thanks.

Masashi-Ono0611:
Thanks for merging, and for double-checking exit-code fidelity and the test migration off `process.exit` spying — appreciate the careful review.

PR #2916, opened 2026-07-17
v0.42.62.0: Add strict recovery snapshots

symmetric-matthew:
## Summary
- add a strict, deterministic, read-only gbrain recovery snapshot command for an explicit brain and source
- bind source-backed page, normalized, fact, take, link, timeline, tag, and retrieval identities without emitting private content or credentials
- document the disposable-target recovery sequence and correct the system-of-record guide
- require real Postgres/PGLite snapshot parity in the existing E2E workflow
- ship v0.42.62.0 documentation bundles and refresh the pinned release action

## Verification
- recovery unit tests: 5/5
- focused recovery/CLI calibration tests: 36/36
- typecheck: pass
- real Postgres/PGLite workflow dispatch: run 29595198362, all jobs green and parity step executed
- build:llms deterministic parity: pass
- key-files current-state guard: pass
- version audit: VERSION/package.json/CHANGELOG all 0.42.62.0

## Boundaries
This adds recovery evidence only. It does not migrate, rebuild, restore, cut over readers, delete a target, or claim production recovery/RTO completion.


time-attack:
Thanks for the thorough work here, including the green Postgres/PGLite parity run. Closing: at ~2,000 added lines across 29 files this is a new subsystem rather than a fix, and features of this size are outside what we're taking — unreviewable scope is its own reason. Separately, a contributor PR shouldn't be claiming a VERSION/package.json/CHANGELOG bump or editing the release workflow; version allocation runs through the maintainer /ship process, so those would need to come out regardless.

PR #2919, opened 2026-07-17
fix stdio MCP source scoping: resolve canonically instead of pinning 'default'

Cossackx:
## Summary

- stop pinning local stdio MCP calls to the seeded `default` source
- resolve the stdio source through the same canonical `resolveSourceId` ladder used by CLI operations
- preserve the stdio trust boundary: `remote: true` and world-scoped takes remain unchanged
- add a narrow resolver seam plus regression tests for brain-default routing, explicit override, and unavailable-resolver fallback

## Root cause

On a multi-source brain, the seeded `default` source can legitimately contain zero pages. The stdio MCP server hard-coded that source into every operation context, so source-scoped operations returned empty results even when the CLI correctly resolved the configured source.

## Verification

- `bun run typecheck` — pass
- `bun test --timeout 30000 test/mcp-stdio-source-resolution.test.ts` — 3 pass, 0 fail
- `bash scripts/check-source-id-projection.sh` — pass
- `bash scripts/check-source-config-leak.sh` — pass
- `bash scripts/check-source-scope-onboard.sh` — pass
- `bash scripts/check-privacy.sh` — pass
- `bash scripts/check-test-isolation.sh` — pass (981 non-serial files scanned)

## Environment limitation

The repository-wide `bun run verify` runner is not green on this Windows contributor host: Bun cannot dispatch the `.sh` package scripts directly, and the parallel unit harness later hangs in unrelated existing Windows/path lifecycle tests. This PR does not claim a full-suite pass; upstream Linux CI remains authoritative for the complete matrix.

### DIFF
--- a/src/mcp/server.ts
+++ b/src/mcp/server.ts
@@ -7,6 +7,7 @@ import { VERSION } from '../version.ts';
 import { buildToolDefs } from './tool-defs.ts';
 import { dispatchToolCall, validateParams, buildOperationContext } from './dispatch.ts';
 import { getBrainHotMemoryMeta } from '../core/facts/meta-hook.ts';
+import { resolveSourceId } from '../core/source-resolver.ts';
 import { loadConfig } from '../core/config.ts';
 import {
   resolveSocketPath,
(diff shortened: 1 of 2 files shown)


time-attack:
Merging. Confirmed on current master that `src/mcp/server.ts` still pins `sourceId` to `GBRAIN_SOURCE || 'default'` in both the tool-call handler and the reflex IPC path, which silently returns empty results on any multi-source brain — a real break of the source-routing contract. Reusing the canonical `resolveSourceId` ladder instead of new logic, and leaving `remote: true` / world-scoped takes alone, is exactly the right shape here. One thing to watch downstream: resolution now consults `process.cwd()` at startup, so an stdio server launched from inside a source checkout will path-match it; `GBRAIN_SOURCE` still wins as the explicit override.

PR #2925, opened 2026-07-17
fix(agent,subagent): honor the configured default source for subagent brain tools

### LINKED ISSUE #2922: Subagent tools ignore the configured default source: put_page always writes to seed 'default' (hardcoded in brain-allowlist.ts)
**What happened?**

Pages written by agent jobs (`gbrain agent run`) always land in the seed `default`
source, ignoring the brain default configured via `gbrain sources default <id>`.

Root cause (confirmed in current source, v0.42.59.0): `buildOpContext` in
`src/core/minions/tools/brain-allowlist.ts:220` hardcodes `sourceId: 'default'`.
The canonical resolution chain (`src/core/source-resolver.ts`, whose tier 5 reads
the `sources.default` config) is never consulted on the subagent path. The
`put_page` operation itself already honors `ctx.sourceId`
(`src/core/operations.ts:825-854`, since v0.31.8) — it faithfully writes to
`'default'` because that's what it receives. `gbrain agent run`
(`src/commands/agent.ts`) has no `--source` flag, and `SubagentHandlerData`
(`src/core/minions/types.ts:405-451`) has no `source_id` field, so there is no
channel to carry a resolved source even if the CLI resolved one.

Impact: on access-controlled multi-source brains the seed `default` is typically
unfederated, so everything an agent job writes is invisible to readers. We run a
periodic SQL sweep (`UPDATE pages SET source_id=... WHERE source_id='default'`)
in production as a workaround. Prompt-level "write to source X" instructions
don't (and shouldn't) work as the enforcement mechanism.

Proposed fix (happy to submit a PR): resolve the source at submit time in
`runAgentRun` via `resolveSourceId(...)` (optionally exposing `--source`),
persist it as `source_id?` on `SubagentHandlerData`, and have
`buildBrainTools`/`buildOpContext` use it instead of the literal — falling back
to `'default'` only when absent, so single-source brains see no change. Plus a
test pinning the behavior (`test/brain-allowlist.serial.test.ts` doesn't cover
`sourceId` today).

**What did you expect?**

Subagent writes honor the same source-resolution chain as every other write path:
with `sources.default` set to `corporate`, agent-job pages land in `corporate`
(or `gbrain agent run --source <id>` exists to scope them explicitly).

**Steps to reproduce**

1. Multi-source brain; run `gbrain sources default corporate` (confirmed: `Default source set to "corporate".`)
2. `gbrain agent run "Write a short page about X." --queue q` and wait for the job to complete
3. Check the pages table: the job's pages (`wiki/agents/<job-id>/...`) have `source_id = 'default'`, not `corporate`

**Environment**

- gbrain version: v0.26.5 deployed (behavior observed in production); root cause confirmed by inspection of current source v0.42.59.0
- OS: Linux (official Docker image, GKE)
- Bun version: as bundled in the image
- Database: self-hosted Postgres (GKE), `gbrain init --supabase` mode

```json
(not directly relevant — reproduces on any multi-source brain with sources.default set;
can provide redacted output on request)
```

jefersonlopes-pipefy:
Fixes #2922.

## Problem

Subagent tool dispatch hardcoded `sourceId: 'default'` in `buildOpContext`
(`src/core/minions/tools/brain-allowlist.ts`), so every `put_page` from a
`gbrain agent run` job landed in the seed `default` source — ignoring
`gbrain sources default <id>` entirely. The `put_page` operation itself already
honors `ctx.sourceId` (v0.31.8, `src/core/operations.ts`); the defect was the
literal at the source of the context. On access-controlled multi-source brains
where the seed `default` is unfederated, everything an agent job writes is
invisible to readers (full details in #2922).

## Fix

- **Resolve at submit time** (`src/commands/agent.ts`): `gbrain agent run` now
  resolves the write-target source through the canonical chain (`--source` flag
  [new] → `GBRAIN_SOURCE` env → `.gbrain-source` dotfile → `local_path` match →
  `sources.default` config → seed `'default'`) via the existing
  `resolveSourceId()` and persists it on the job payload
  (`SubagentHandlerData.source_id`), so restart/replay writes to the same source
  regardless of the worker's cwd/env. Fan-out children get the same stamp.
- **Thread it through** (`src/core/minions/handlers/subagent.ts` →
  `src/core/minions/tools/brain-allowlist.ts`): the handler passes
  `data.source_id` into `buildBrainTools`; `buildOpContext` uses it with a
  seed-`'default'` fallback.
- **Backward compatible**: legacy job rows and direct submitters (no
  `source_id` on the payload) keep the exact pre-fix behavior — the fallback is
  the same literal that was there before. Single-source brains see no change
  (the chain resolves to `'default'` anyway).

## Tests

- Two new cases in `test/brain-allowlist.serial.test.ts` pin both directions:
  `put_page` routed to `opts.sourceId`, and the legacy fallback when absent.
- `bun test test/brain-allowlist.serial.test.ts` → 20 pass / 0 fail.
- `bun test test/agent-cli.test.ts test/subagent-handler.test.ts` → 54 pass / 0 fail.
- `tsc --noEmit` clean.
- `bun run verify`: 25 pass / 6 fail — the 6 failures (`check:privacy`,
  `check:test-isolation`, `check:cli-exec`, `check:eval-glossary`,
  `check:skill-brain-first`, `check:resolver`) reproduce identically on a
  pristine `master` worktree on this machine (Windows checkout: CRLF/filemode
  artifacts), i.e. none are introduced by this diff.

## Notes

- The dream/cycle submit paths were intentionally left untouched to keep this
  PR minimal; they can adopt `source_id` stamping in a follow-up if desired.


time-attack:
Merging this one. I verified on current master that the payload plumbing already exists but `src/commands/agent.ts` never resolves a source, so CLI-submitted agent jobs really do still land everything in the seed `default` — a genuine write-routing bug on multi-source brains. Resolving at submit time via the existing `resolveSourceId()` chain (rather than at claim time, where the worker's cwd would win) is the right call, and the legacy fallback keeps single-source brains byte-identical. Thanks for the precise root-cause writeup in #2922 and for keeping the diff to the one thing.

PR #2944, opened 2026-07-17
feat(doctor): fts_language_drift check — surface the query/stored tokenizer split-brain

Masashi-Ono0611:
## Summary

Follow-up to the configurable FTS language (#2941, v0.42.62). The feature resolves the language on two independent sides: the query arm reads `GBRAIN_FTS_LANGUAGE` from the process env at first call (`getFtsLanguage()`, cached per process), while stored `search_vector` columns carry whatever language the trigger functions had at migration/reindex time. Nothing mechanically guards their agreement — `docs/guides/multi-language-fts.md` currently makes cross-context consistency a manual operator duty. A single context missing the env var (a launchd daemon, cron, autopilot, or MCP serve on a headless box) tokenizes queries under a different dictionary than the stored vectors, and the keyword recall arm silently degrades — no error anywhere, just quietly worse hybrid recall in exactly the contexts nobody watches.

Design:
- **New doctor check `fts_language_drift`** introspects the `update_page_search_vector` / `update_chunk_search_vector` bodies in `pg_proc` (scoped to the public schema) and parses their `to_tsvector('<lang>'` literals — the trigger bodies are ground truth for what the stored side tokenizes with: no stamp to trust, retroactive on any brain, and works on both engines (PGLite is full Postgres).
- **Warn states**: runtime-vs-stored mismatch (message carries both remediations — export the stored language in *every* context, or `reindex-search-vector --yes` to retokenize toward the runtime one); the two trigger functions disagreeing with each other (partial reindex); one function missing while its sibling exists (schema anomaly — both absent is a pre-FTS-language brain and reports not-applicable); multiple distinct languages inside one body (hand-edited/corrupted trigger).
- **Fail-open everywhere**: introspection errors and unparseable bodies report ok-with-note, never a false warn and never a doctor failure. Parsing tolerates `pg_catalog.` qualification, whitespace, and `::regconfig` casts, and collects every literal rather than trusting the first.
- Wired into both doctor surfaces (`buildChecks` + `doctorReportRemote`), registered under `ops` in `doctor-categories.ts`, pure decision logic extracted (`parseFtsLanguagesFromProsrc` + `computeFtsLanguageCheck`) for tests.

Honest limitation (documented in the wrapper's docstring): the trigger bodies tell you what NEW writes tokenize with, not a per-row guarantee — a reindex backfill interrupted mid-run leaves older rows under the previous language while the functions already carry the new one, and this check reads that as consistent. The reindex command is idempotent and re-running it converges; tracking per-table backfill completion would need a persisted stamp and belongs to the reindex command itself (natural follow-up if wanted).

## Test Coverage

`test/doctor-fts-language-drift.serial.test.ts` (serial: mutates `GBRAIN_FTS_LANGUAGE` + the module cache):
- Parser: literal extraction, custom config names (`pt_br`), `pg_catalog` qualification + whitespace + `::regconfig` cast, mixed-language bodies, never-guess empty result on parameterized/absent literals.
- Decision matrix: consistent → ok naming the language; runtime≠stored → warn carrying both remediations; function-vs-function disagreement → warn; missing sibling → warn; both absent → not-applicable ok; unparseable body → ok skip.
- Live e2e on a bootstrapped PGLite brain: default env resolves `english` on both sides → ok; flipping the env after bootstrap → the real introspection path reports the split-brain warn.

Local gates: 13 tests pass, doctor behavioral/remote/categories suites green, `bun run typecheck` clean, `bun run verify` 31/31 green, `bun run build:llms` regenerated after the KEY_FILES.md update.


time-attack:
Thanks for the careful follow-up to #2941 — the drift scenario you describe (a headless context missing the env var) is real and well explained. It's still net-new diagnostic surface rather than a fix for something broken, and parsing `to_tsvector('<lang>'` out of `pg_proc` bodies couples doctor to the exact text of our trigger definitions, which we'd have to keep in lockstep forever. Closing under the feature policy; the manual operator guidance in the multi-language FTS guide stays the answer for now.

Masashi-Ono0611:
Understood on the feature-policy line, and agreed that parsing `to_tsvector('<lang>'` out of `pg_proc` bodies would couple doctor to the trigger definitions' exact SQL text long-term. Thanks for engaging with the drift scenario itself.

PR #2945, opened 2026-07-17
feat(doctor): durability_push_health — surface the hardened-source git tier going silently dark

Masashi-Ono0611:
## Summary

Follow-up to the durability hardening line (#2426, carried forward by the #2938 sync data-loss wave). `gbrain sources harden` promises agent writes reach git: write-through commits the artifact path-limited, and the post-commit hook background-pushes with retry. Both tiers are deliberately best-effort at the write site — a git hiccup must never fail the write, the DB row and the on-disk file stay the durable sinks — which makes sustained failure silent **by design**: `commitWriteThroughFile` returns `false` into a void, the push helper appends `LOCAL-ONLY, NEEDS ATTENTION` to `~/.gbrain/brain-push.log`, and nothing ever surfaces either. For a source the operator explicitly hardened, that silence defeats the point: the git tier can be dark for weeks while every write looks fine.

Design — new doctor check `durability_push_health`:
- **Enumerate hardened sources** (`sources.local_path` + `isDurabilityHardened`, which is file-read-cheap), sorted for a deterministic probe set, capped at 5 repos with the unprobed source ids named in the message rather than silently dropped.
- **Per-repo git probes** (argv-array, 5s timeouts): commits ahead of `@{upstream}` (the push tier not reaching origin), upstream presence (a branch with no upstream means background pushes have nowhere to go), and a `status --porcelain` dirty count (informational — non-write-through edits are legitimate).
- **Push-log evidence**: a bounded 64KB tail of `brain-push.log` is reduced to "latest terminal state wins" — terminal `LOCAL-ONLY` / `lock-timeout` entries within 7 days count only when no later `[push] ok` follows, so a recovered repo stops warning immediately (the log carries no repo key; single-brain-repo is the overwhelming shape, noted in the docstring).
- **Warn** on local-only commits, missing upstream, or recent un-cleared push failures, carrying the recovery commands (`scripts/brain-commit-push.sh --push-only`, `gbrain sources pull <id>` + push, log path). **Fail-open throughout**: a probe error degrades that repo to an explicitly named "unverified" state — never folded into the pushed count (a repo doctor couldn't see must not read as a green light), and never a doctor failure.
- `pushLogPath()` is now exported from `brain-repo-durability.ts` as the single source for the log location. Pure decision + log-parsing functions (`computeDurabilityPushCheck`, `parsePushLogFailures`) extracted for tests; wired into both doctor surfaces; registered under `ops`.

Deliberately out of scope: only the checked-out branch is probed (the hardened flow pushes HEAD; enumerating all local branches is a different feature), and per-repo log keying would need the push template to start stamping a repo identifier — a natural follow-up if multi-repo hardening becomes common.

## Test Coverage

`test/doctor-durability-push-health.test.ts` — decision matrix (no hardened sources → not-applicable ok with the enable hint; all pushed → ok naming counts; ahead>0 / missing upstream / recent failures → warn with recovery text; probe-failure → ok but named unverified and excluded from the pushed count; dirty counts informational on both paths) and log-parser edges (terminal vs transient lines, later-ok clears earlier failures both ways, window expiry, unparseable lines, long-line truncation).

Local gates: 12 tests pass, doctor behavioral/remote/categories + brain-repo-durability suites green, `bun run typecheck` clean, `bun run verify` 31/31 green, `bun run build:llms` regenerated after the KEY_FILES.md update.


time-attack:
Thanks — the observation that a hardened source's git tier can go dark for weeks is a fair one, and the writeup is genuinely good. But the git tier is best-effort by design with the DB row and the on-disk file as the durable sinks, so nothing is actually lost here; this adds new diagnostic surface plus per-repo subprocess probes inside `doctor`. Closing under the feature policy, not on quality.

Masashi-Ono0611:
Good point that the DB row and on-disk file are the durable sinks and the git tier is best-effort by design — that reframes it usefully. Thanks for the review.

Issue #2781, opened 2026-07-13
bug(autopilot): interval-derived cycle timeout stamp overrides the #1737 handler anchor — 600s budgets at default interval dead-letter every substantive cycle at scale; patterns' 35-min worst case exceeds any interval-derived budget


alexputici:
**What happened?**

Deployment context for this batch of reports: see #2778 (same production operation, environment, and receipts discipline).

Autopilot stamps `timeout_ms = Math.max(baseInterval * 2 * 1000, 300_000)` on every autopilot-cycle job (`src/commands/autopilot.ts:641`, threaded through `autopilot-fanout.ts:399`/`:452`), and an explicit stamp always beats the handler default (`queue.ts:285` — `opts?.timeout_ms ?? defaultTimeoutMsFor(jobName)`; `handler-timeouts.ts:18-19` even documents "An explicit `opts.timeout_ms` always wins"). Consequence: the engine's own `THIRTY_MIN_MS` anchor for `autopilot-cycle` (`handler-timeouts.ts:26,37` — the #1737 fix, shipped v0.42.29.0) **never applies**. At the default 300s interval, cycles get a 600s budget.

On a ~7k-page brain this dead-letters every substantive cycle. Observed 2026-07-08: **38 dead cycles/24h**, cycle_freshness FAIL at 40h, extraction backlogs ballooning — the exact thrash #1737 was written to prevent.

Escalation, and the deeper shape of the bug: even 1800s budgets were insufficient — 30 more dead cycles, **every one aborted mid-`patterns`** — because the patterns phase submits its subagent with a hardcoded 30-min job timeout and waits up to 35 min (`src/core/cycle/patterns.ts:86,95`). One phase's worst case exceeds ANY interval-derived budget ≤ 35 min, so the six tail phases (consolidate → schema-suggest) starved for days. Real full cycles at a corrected budget measured 2502s (job #23745) and 2355s (#24261) — i.e. the native default, and even a "reasonable" 900s interval, are lethal at this scale.

**What did you expect?**

- Floor the stamped cycle timeout at the handler default (min 30 min), or add an `autopilot.cycle_timeout_ms` config knob decoupled from `--interval`.
- Sharper fix: budget `patterns` (and any phase that awaits a subagent) from **remaining cycle time**, not a hardcoded constant — a fixed 35-min phase worst-case inside an interval-derived budget can never be safe at every interval.

**Steps to reproduce**

[...]

PR #2960, opened 2026-07-18
fix(cycle): synthesize budgets its sequential child waits from remaining job time (#2781 follow-up, stacked on #2959)

Masashi-Ono0611:
## What

Part of #2781 — the synthesize half of the remaining-budget work, following the patterns-phase PR (#2959, which this branch is **stacked on**; the first two commits here are that PR).

Patterns needed one clamp because it runs one child and one wait. Synthesize has a second-order problem: it submits N children up-front, then waits for them **sequentially** — so a fixed per-wait timeout accumulates N × `subagent_wait_timeout_ms` past any parent budget. A single up-front clamp can't express that; the budget has to be recomputed as it drains.

- **Shared helpers move to `src/core/cycle/deadline-budget.ts`** (`CYCLE_DEADLINE_RESERVE_MS`, `MIN_SUBAGENT_START_BUDGET_MS`, `clampSubagentBudgets`, new `remainingChildBudgetMs`). patterns.ts already imports from synthesize.ts, so synthesize importing these back from patterns would create an import cycle. `MIN_PATTERNS_SUBAGENT_BUDGET_MS` is renamed `MIN_SUBAGENT_START_BUDGET_MS` — it now guards both phases.
- **Pre-fanout gate**: below the 2-min start minimum, skip honestly (`insufficient_cycle_budget`) before any Sonnet submit. Read after the verdict loop, which itself consumes wall-clock.
- **Per-child submit clamp**: each child's `timeout_ms` clamps to the remaining budget at ITS submit. Both the per-transcript and per-chunk guards stop at the start minimum (not merely ≤ 0 — a near-zero budget would submit guaranteed-kill children with second-scale timeouts). A partial fan-out is safe to resume: the per-chunk idempotency keys dedupe already-submitted work next cycle.
- **Per-iteration wait recompute**: every wait derives `min(config, remaining)` against the same absolute deadline. Once the budget is gone, the remaining children are settled without waiting: cancel-first (one write per child); a null return means the child was already terminal, so its real status is refetched — a completed sibling isn't misreported as timeout, and its pages still collect. The settle loop is deliberately sequential per-child rather than one bulk UPDATE: `cancelJob` carries per-job semantics (child_done emission, aggregator unblocking) a bulk write would skip, and each call is a millisecond-scale write with no waits.
- **Budget-truncated runs don't stamp `last_completion_ts`** — arming the default 12h cooldown over transcripts this run never gave a real chance would contradict "next cycle retries with a fresh budget".
- Wait-timeout path cancels the child, same rationale as patterns (the child's `timeout_ms` clock starts at its claim, so a queued child can outlive the clamp).

Direct callers (`gbrain dream`) pass no deadline and keep the configured behavior unchanged.

## Tests

`test/cycle-synthesize-deadline-budget.test.ts` (new): unit tests on `remainingChildBudgetMs` (null back-compat, reserve subtraction, non-positive on breach) + structural pins (cycle.ts threading, gate placement after the verdict loop, clamped-not-raw submit values, in-loop recompute anchored inside the wait loop body, cancel-first settle with real-status refetch, cooldown suppression, MIN-threshold guards at all three sites). The patterns test file follows the helpers to their new module. 314 pass / 0 fail across the touched-area sweep (all synthesize suites, patterns, minions queue, autopilot-cycle handler, cycle-abort); `bun run typecheck` clean.

Review: three-round external model review; round 1 found the ≤0-vs-minimum submit guard gap, the cooldown stamping contradiction, and the settle-loop reporting race; round 2 caught that a timeout on a deadline-CLAMPED wait must also count as budget truncation (the child never got its full configured wait). All fixed and re-verified.

Related: #2781, #2959 (stacked base — the patterns half), #2852 (dispatch floor, complementary), #1594 (hardcoded-subagent-timeout family).

### DIFF
--- a/src/commands/jobs.ts
+++ b/src/commands/jobs.ts
@@ -1964,6 +1964,7 @@ export async function registerBuiltinHandlers(
       brainDir: repoPath,
       phases: [phase as any],
       signal: job.signal,
+      deadlineAtMs: job.deadlineAtMs, // #2781: phases budget sub-work from remaining time
     });
     return { phase, status: report.status, report };
   };
--- a/src/commands/jobs.ts
+++ b/src/commands/jobs.ts
@@ -1780,6 +1780,7 @@ export async function registerBuiltinHandlers(
       brainDir: effectiveBrainDir,
       pull,
       signal: job.signal, // propagate abort so cycle bails on timeout/cancel
+      deadlineAtMs: job.deadlineAtMs, // #2781: phases budget sub-work from remaining time
       ...(sourceId ? { sourceId } : {}),
       ...(requestedPhases && requestedPhases.length > 0 ? { phases: requestedPhases as any } : {}),
       yieldBetweenPhases: async () => {
@@ -1817,6 +1818,7 @@ export async function registerBuiltinHandlers(
       brainDir: repoPath,
       pull: false, // brain-wide DB/maintenance work never git-pulls
       signal: job.signal,
+      deadlineAtMs: job.deadlineAtMs, // #2781: phases budget sub-work from remaining time
       phases,
       yieldBetweenPhases: async () => { await new Promise<void>((r) => setImmediate(r)); },
     });
--- a/src/core/cycle.ts
+++ b/src/core/cycle.ts
@@ -1695,6 +1695,7 @@ export async function runCycle(
           // #1586: scope synthesized writes to the cycle's resolved source
           // (explicit --source wins, else derived from the checkout dir).
           sourceId: cycleSourceId,
+          deadlineAtMs: opts.deadlineAtMs ?? null,
         }));
         result.duration_ms = duration_ms;
         phaseResults.push(result);
--- a/src/core/cycle.ts
+++ b/src/core/cycle.ts
@@ -479,6 +479,16 @@ export interface CycleOpts {
    * Validated via `assertValidSourceId` in `cycleLockIdFor` (defense-in-depth).
    */
   sourceId?: string;
+  /**
+   * Absolute wall-clock deadline (epoch ms) of the enclosing minion job,
+   * from `MinionJobContext.deadlineAtMs` (the claim-time `timeout_at`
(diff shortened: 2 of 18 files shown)


Masashi-Ono0611:
Closing as superseded — this branch's work has already been reconciled and absorbed via a maintainer "takeover" rebase, per this repo's established pattern for stale-but-valid contributor branches.

**Evidence:**

1. This PR's stacked base, #2959 ("budget the patterns subagent from remaining job time"), merged to master as commit `23e0541d9b9b44a33d07e6a998eeb945975abbe2` on 2026-07-20. Verified the three non-synthesize files this PR (#2960) also touches — `src/commands/jobs.ts`, `src/core/minions/types.ts` (`deadlineAtMs: number | null` field), `src/core/minions/worker.ts` (deadline-derived abort timer) — are already present on `upstream/master` with matching content, all originating from #2959.
2. The remaining synthesize-specific half of this PR (the part #2959 didn't cover) has an open, superseding PR: **#3028** ("fix(cycle): synthesize budgets its sequential child waits from remaining job time (#2781 follow-up, supersedes #2960)"), on branch `takeover/pr-2960-deadline-budget`. Its description states explicitly: *"Supersedes #2960 by @Masashi-Ono0611 (all design and code is theirs; this is a rebase/reconciliation, credited via Co-authored-by). Do not merge both."*
3. Diffed #3028 against this PR directly: #3028 touches exactly the 6 files this PR's incremental diff (on top of #2959) touches — `src/core/cycle.ts`, `src/core/cycle/deadline-budget.ts`, `src/core/cycle/patterns.ts`, `src/core/cycle/synthesize.ts`, `test/cycle-patterns-deadline-budget.test.ts`, `test/cycle-synthesize-deadline-budget.test.ts` — matching #3028's own claim of a byte-identical final tree on all six files.
4. #3028's CI is fully green (all checks: build/typecheck/tests/security scans SUCCESS).

Since #3028 is the sanctioned, up-to-date, conflict-free successor and its own description says not to merge both, closing this one in its favor rather than rebasing again.

PR #2968, opened 2026-07-18
Fix Bun SQL jsonb[] mis-encoding in addCodeEdges

zsimovanforgeops:
On the Postgres engine, addCodeEdges binds edge_metadata through `${metas}::jsonb[]` array binds. Bun SQL mis-encodes jsonb array binds: each element arrives as a double-encoded JSON string rather than an object. Every code edge inserted this way carries string metadata, so the resolver's `edge_metadata || jsonb_build_object(...)` UPDATE concatenates onto a string scalar and produces a jsonb array instead of setting `resolved_chunk_id`. The resolver logs `edges_resolved > 0` while the table keeps zero usable rows, and code_callers / code_callees / code_blast return nothing on Postgres-engine brains. PGLite is unaffected (its addCodeEdges already uses per-row placeholders).

To check whether a brain is affected:

    SELECT jsonb_typeof(edge_metadata), count(*)
    FROM code_edges_symbol GROUP BY 1;

Affected rows show `string` or `array`, never `object`.

The fix rewrites both inserts (code_edges_chunk, code_edges_symbol) to per-row `$n::text::jsonb` placeholders via sql.unsafe — the same shape executeRawJsonb and the PGLite engine already use.

Existing corrupted rows need a one-time repair before re-running the resolver:

    UPDATE code_edges_symbol SET edge_metadata = '{}'::jsonb
    WHERE jsonb_typeof(edge_metadata) <> 'object';

(and the same for code_edges_chunk). After that repair on our brain, a re-run resolved 219 edges that had been reported-but-lost on every previous pass.

Tests: code-edges + edge-extractor suites pass; tsc clean.

### DIFF
--- a/src/core/postgres-engine.ts
+++ b/src/core/postgres-engine.ts
@@ -5671,43 +5671,53 @@ export class PostgresEngine implements BrainEngine {
     const unresolved = edges.filter(e => e.to_chunk_id == null);
 
     if (resolved.length > 0) {
-      const fromIds = resolved.map(e => e.from_chunk_id);
-      const toIds = resolved.map(e => e.to_chunk_id as number);
-      const fromQual = resolved.map(e => e.from_symbol_qualified);
-      const toQual = resolved.map(e => e.to_symbol_qualified);
-      const edgeTypes = resolved.map(e => e.edge_type);
-      const metas = resolved.map(e => JSON.stringify(e.edge_metadata ?? {}));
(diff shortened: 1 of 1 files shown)


time-attack:
Superseded with credit by #3020 (merged): your $N::text::jsonb fix landed with a new Postgres-gated regression test proving the double-encode reproduces pre-fix. Authorship preserved via Co-authored-by. Thank you!

Issue #1493, opened 2026-05-26
extract links: hardcoded DIR_PATTERN whitelist drops all Obsidian/Notion-style wikilinks silently (created: 0 from N pages, no warning)


flamerged:
### Summary

`gbrain extract links` silently creates 0 edges from an Obsidian-style vault export because the hardcoded `DIR_PATTERN` regex in `src/core/link-extraction.ts` only matches gbrain's own slug conventions. Operators ingesting any non-canonical directory layout (Obsidian vaults, Notion exports, custom topic taxonomies) get 0% graph coverage with no error and no warning.

### Where

```ts
// src/core/link-extraction.ts
const DIR_PATTERN = '(?:people|companies|meetings|concepts|deal|civic|project|projects|source|media|yc|tech|finance|personal|openclaw|entities)';
```

…and the three regex variants below it all bake `DIR_PATTERN/` into the link-target requirement.

### Impact

For an Obsidian export with directories like `books/`, `documentation/`, `guides/`, `ideas/`, `nutrition/`, `travel/`, `writing-dump/`, `archive/`, none of the wikilinks resolve. The extractor reports `created: 0 from N pages` with no clue what's wrong. Operators have to grep the source to discover the hardcoded whitelist; the docs don't mention it.

Even when the directory DOES match (e.g. `projects/` is in the whitelist), Obsidian's *title-form* wikilinks (`[[Yomumichi Documentation]]`) fail because the regex requires `[[directory/slug-path]]`. Obsidian users never write that form — Obsidian resolves by title via its own fuzzy index, and exports preserve title-form.

### Repro

```bash
# Pages exist:
gbrain sources list
# → some-source  321 pages
gbrain get projects/project-list/yomumichi/yomumichi
# → full content with [[Yomumichi Documentation]], [[Project List.base]], [[Notes on X]]

gbrain extract links --source db
# Links: created 0 from 321 pages
gbrain get-links projects/project-list/yomumichi/yomumichi
# → []
```

### Proposed fix

1. **Make `DIR_PATTERN` configurable** via `gbrain config set link_extraction_dirs "books,documentation,..."`. Hardcoded whitelists in extraction layers are an anti-pattern; per [search results from older versions](https://www.marktechpost.com/2026/05/22/) some external write-ups already document this config key as if it exists, suggesting it was either planned or removed.
[...]

Issue #726, opened 2026-05-07
Link extractor only matches whitelisted top-level entity dirs — breaks for prefixed wikis


PascalWagner:
## Problem

\`gbrain extract links\` (both \`--source fs\` and \`--source db\`) hardcodes the entity-dir whitelist at the START of the link path: \`people/\`, \`companies/\`, \`meetings/\`, etc.

This silently produces 0 links for any wiki that nests entity dirs under a domain prefix. For example, a multi-domain personal wiki structured like:

\`\`\`
shared/people/reza-rastegar.md
shared/people/dusty-jolitz.md
gyc/companies/dlp-capital.md
gyc/companies/avenir-aviation.md
\`\`\`

…has zero entity links extracted, even when wiki pages explicitly reference each other via standard markdown links like \`[Reza](shared/people/reza-rastegar.md)\`.

The result: \`gbrain extract all\` reports success but \`gbrain doctor\` shows \`Entity link coverage 0%\`. Users have no idea why their entity graph is empty unless they read the regex source.

## Proposal

Add a config knob \`entity_dir_prefixes\` that defaults to \`[]\` (current behavior, fully backwards compatible). When non-empty, the extractor allows the listed prefixes before the entity dir name in both the \`ENTITY_REF_RE\` and the bare-slug regex.

\`\`\`json
{
  \"entity_dir_prefixes\": [\"shared\", \"gyc\", \"co-living\", \"pwh\"]
}
\`\`\`

After this, \`[Reza](shared/people/reza-rastegar.md)\` extracts \`shared/people/reza-rastegar\` as the canonical slug (matches what's stored in \`pages.slug\`).

## Current workaround

Local patch maintained at https://github.com/PascalWagner/pascal-wiki/blob/main/shared/ai-tooling/patches/gbrain-link-extraction-prefix-dirs.patch — happy to convert to a proper PR with the config knob design.

## Version

\`gbrain 0.28.3\`. Wiki has 2,141 pages, all matching the prefix pattern; extractor finds 0 links until patched.

rayers:
PR #632 (`make entity link directories configurable`) is the upstream answer here. The `extract.link_dirs` config it adds is more granular than your `entity_dir_prefixes` proposal but covers the same cases:

```bash
gbrain config set extract.link_dirs 'shared/people,shared/companies,gyc/companies,gyc/people,co-living/people,pwh/companies'
```

[...]

PR #2978, opened 2026-07-18
feat(links): resolve path-shaped wikilinks for any directory by exact slug match (#1493, #726)

gporterl:
## Summary

Path-shaped wikilinks (`[[any-dir/sub/name]]`) whose directory prefix is not in the hardcoded `DIR_PATTERN` whitelist are currently dropped silently by the DB-path extractor — no candidate, no unresolved report, `created: 0` with no clue why. This is the path-form half of #1493 and the core of #726: any corpus with its own directory taxonomy (`janus/`, `ops/`, `books/`, `guides/`, …) gets near-zero graph coverage.

This PR adds **any-directory exact-slug resolution** for path-shaped wikilinks, composing with (not duplicating) the #972 bare-name/basename machinery:

- Two new regex passes (`2a′` qualified / `2b′` unqualified) match path-shaped targets — ≥2 segments of `[a-z0-9][a-z0-9._-]*` — for any directory prefix. They run after the whitelist passes with those spans masked, so whitelisted-dir behavior stays byte-identical and nothing double-emits.
- Resolution is **exact slug match** (new optional `SlugResolver.slugExists`, served off the same source-scoped `getAllSlugs` snapshot the basename index already uses — still one engine call per run). Exact match can't false-positive the way basename matching can, so the gate (`link_resolution.any_dir_exact_path`) defaults **ON**, with an env override, documented alongside `global_basename`.
- A miss is **recorded, never silent**: it rides the existing unresolved-report machinery (`{field: 'wikilink', name: <slug>}`) into the `put_page` `auto_links` response and the extract summary. No ghost edges.
- Qualified `[[src:path]]` refs keep their source pin and go through cross-source endpoint resolution identically to whitelisted qualified refs.
- The #972 bare-name pass (`[[name]]`, `global_basename`, `wikilink_basename` edges) is untouched; tested side-by-side in one call.

Relationship to #2868: complementary, not competing. That PR improves basename tail-matching and Obsidian title-form targets and extends the whitelist for one deployment; this PR removes the whitelist dependency for exact-path targets entirely. Obsidian *title-form* links (`[[Some Page Title]]`) remain #972/#2868 territory.

`DIR_PATTERN` itself is untouched (markdown `[Name](dir/x.md)` links and the bare-slug prose pass remain whitelist-gated) — scope here is the wikilink symptom in #1493.

## Tests

- ~60 new tests: `test/link-extraction-any-dir.test.ts` (48 — regex passes, masking, resolution, pins, unresolved reporting, #972 coexistence, fetch discipline), `test/put-page-autolink-pin-dedup.test.ts` (new), plus additions to `extract-source-aware`, `extract-stale`, `extract-db`, and `config` suites.
- Targeted suites (17 files: both link-extraction files, extract fs/db/stale/source-aware/incremental, doctor, config, put-page, backlinks, global-basename PGLite e2e): 444/444 green; `tsc --noEmit` clean.
- Full `bun run test`: green except one pre-existing failure (`retrieval-reflex`) that fails identically on clean master.

## Real-corpus validation

Run against the motivating 43-page brain (directories `janus/`, `janus-ui/`, `projects/`, plus root-level pages): 76 extractable wikilinks; before → 29 path-shaped links to non-whitelisted dirs silently dropped; after → **all 29 become candidates, 0 unresolved, 0 false edges**; the 25 whitelist-dir links and 22 bare-name links behave exactly as before.

## Hardening (from two adversarial review rounds before submission)

- **Qualified source pins now real end-to-end**: `[[src:path]]` previously lost its qualifier at the candidate layer (for whitelisted dirs too); `LinkCandidate.targetSourceId` now threads the pin through resolution — pin miss = skip, never misdirected to a same-slug page in origin/default. Pins join the dedup key; same-source pin + unpinned duplicates re-collapse before writes so `auto_links.created` counts stay honest.
- **Scoped extraction keeps the documented local-first→default fallback**: `slugExists` resolves over scoped-source ∪ `default`, mirroring `resolveCandidateSources`, on both the DB and `--stale` paths (per-page source scoping on all-source sweeps). The basename index stays strictly source-scoped.
- **No silent drops on any path**: `--stale` and `--json` both surface unresolved refs (`unresolved_count` / `unresolved_refs`).
- `link_resolution.any_dir_exact_path` registered in `KNOWN_CONFIG_KEYS` so the documented disable works without `--force`.

## Notes for review

- `LINK_EXTRACTOR_VERSION_TS` **is bumped, as its own commit**: semantically the new pass warrants a one-time stale sweep on deployed brains. If you'd rather defer the bump to merge time (the #2868 convention), dropping that single commit is all it takes.
- Uppercase/spaced targets are not slugified here (documented + tested as-is); that belongs with the #2868 line of work.
- `[[own/slug]]` self-loops follow whitelist parity (emitted); the basename path drops self-loops — happy to add symmetry if preferred.

Fixes the path-form portion of #1493; addresses #726.


time-attack:
Thanks for the thorough work here — the masked-span composition with the #972 machinery shows you understood the existing extractor properly. At ~1.4k added lines changing link resolution for every corpus with a gate that defaults ON, this is more surface and more default-behavior change than I can responsibly review and support. Closing under the large-feature policy; the whitelist remains the supported way to declare a directory taxonomy.

PR #2980, opened 2026-07-19
fix(providers): reuse buildGatewayConfig for --model test override (#2863)

### LINKED ISSUE #2863: Bug Report: `gbrain providers test --model` ignores `provider_base_urls`, causing false "Incorrect API key" errors for non-default endpoints
# Bug Report: `gbrain providers test --model` ignores `provider_base_urls`, causing false "Incorrect API key" errors for non-default endpoints

**gbrain version:** 0.42.59.0
**OS:** macOS 26.5.2
**Install method:** `bun install -g gbrain`

## What happened?

When a user configures a custom DashScope endpoint via `provider_base_urls` in `~/.gbrain/config.json` (e.g., a China-region/华北 2 MaaS endpoint), the regular embedding path (`gbrain import`, `gbrain query`, `gbrain providers test` without `--model`) works correctly.

However, running:

```bash
gbrain providers test --touchpoint embedding --model dashscope:text-embedding-v3
```

fails with:

```text
[embed(dashscope:text-embedding-v3)] Incorrect API key provided.
```

even though the API key is valid and the endpoint works when tested directly with `curl`.

## What did you expect?

`gbrain providers test --model ...` should use the same `provider_base_urls` configuration as the brain's active embedding path, so the probe accurately reflects whether the configured provider is reachable.

## Steps to reproduce

1. Configure a non-default DashScope base URL and valid key:

   ```json
   // ~/.gbrain/config.json
   {
     "embedding_model": "dashscope:text-embedding-v3",
     "embedding_dimensions": 1024,
     "provider_base_urls": {
       "dashscope": "https://llm-xxxxxxxx.cn-beijing.maas.aliyuncs.com/compatible-mode/v1"
     }
   }
   ```

2. Set `DASHSCOPE_API_KEY` in `~/.zshrc` / `~/.gbrain/.env`.
3. Run `gbrain providers test --touchpoint embedding` → **passes** (uses config defaults).
4. Run `gbrain providers test --touchpoint embedding --model dashscope:text-embedding-v3` → **fails** with `Incorrect API key`.
5. Direct `curl` to the configured endpoint with the same key returns HTTP 200 and a valid embedding vector.

## Root cause

In `src/commands/providers.ts`, the `runTest` function overrides the gateway when `--model` is passed, but it only forwards `embedding_model` / `chat_model` and `env`. It does **not** forward `config.provider_base_urls`:

```typescript
// src/commands/providers.ts ~L183-195 (v0.42.59.0)
if (tpArg === 'embedding') {
  const dims = recipe?.touchpoints.embedding?.default_dims ?? 1536;
  configureGateway({
    embedding_model: modelArg,
    embedding_dimensions: dims,
    // base_urls is missing here!
    env: { ...process.env },
  });
} else {
  configureGateway({
    chat_model: modelArg,
    // base_urls is missing here too!
    env: { ...process.env },
  });
}
```

As a result, `configureGateway` falls back to the recipe's hard-coded default endpoint (`dashscope-intl.aliyuncs.com`), which rejects the user's regional API key.

## Proposed fix

Load `config.provider_base_urls` in `runTest` and pass it into `configureGateway` in both the embedding and chat override branches:

```typescript
let cfg: ReturnType<typeof loadConfig> | undefined;
if (modelArg) {
  // ... existing try/catch to warn about testing in isolation ...
  try {
    cfg = loadConfig();
    // ... warn messages ...
  } catch { /* no brain configured */ }

  if (tpArg === 'embedding') {
    const dims = recipe?.touchpoints.embedding?.default_dims ?? 1536;
    configureGateway({
      embedding_model: modelArg,
      embedding_dimensions: dims,
      base_urls: cfg?.provider_base_urls, // <-- add this
      env: { ...process.env },
    });
  } else {
    configureGateway({
      chat_model: modelArg,
      base_urls: cfg?.provider_base_urls, // <-- add this
      env: { ...process.env },
    });
  }
}
```

The `cfg` declaration needs to be lifted outside the `try` block so it is visible in the later `configureGateway` calls.

## Verification after local patch

After applying the above patch, both commands succeed against the configured 华北 2 endpoint:

```bash
$ gbrain providers test --touchpoint embedding
Probing embedding provider...
  ✓ 347ms, 1024 dims
All probes green.

$ gbrain providers test --touchpoint embedding --model dashscope:text-embedding-v3
Probing embedding provider...
  ✓ 221ms, 1024 dims
All probes green.
```

## Environment / doctor

- Engine: PGLite
- `gbrain doctor`: `embedding_width_consistency` OK, schema width 1024d matches gateway
- No ZeroEntropy / Anthropic keys configured
- `gbrain providers list`: `dashscope` shows `✓ ready`

## Suggested labels

`bug`, `cli`, `providers`

Masashi-Ono0611:
## What

`gbrain providers test --model <id>` ignored `config.provider_base_urls`,
causing false "Incorrect API key" errors when a user's brain is configured
against a non-default (e.g. regional DashScope) endpoint.

## Root cause

Two independent, drifted config resolvers:

- The production path (`src/cli.ts#connectEngine`,
  `src/core/init-embed-check.ts`) builds its `AIGatewayConfig` via
  `buildGatewayConfig()`, which folds `provider_base_urls`, env-sourced
  local-server base URLs, `provider_chat_options`, and file-plane API keys.
- The `--model` override branch in `runTest()`
  (`src/commands/providers.ts`) hand-rolled a second, narrower config
  object that only carried the overridden model + raw `env`, dropping
  `provider_base_urls` entirely. The probe then silently fell back to the
  recipe's hardcoded default endpoint and failed with a misleading auth
  error even though the key was valid for the configured endpoint.

Bare `gbrain providers test` (no `--model`) was never affected — it goes
through `configureFromEnv()`, which already forwards `base_urls` correctly.

## Fix

`cfg` (already loaded via `loadConfig()` inside `runTest` for the
isolation-warning message) is lifted out of its `try`/`catch` and
`buildGatewayConfig(cfg)` is spread into both `configureGateway()` calls
before overriding `embedding_model`/`chat_model`. The isolated `--model`
probe now resolves its endpoint exactly the way the brain's real
import/query path would, while still targeting exactly the model
requested. Falls back to bare `env` when no brain is configured yet (first
install), matching prior behavior for that case.

Confirmed `chat_fallback_chain` (also threaded through by
`buildGatewayConfig`) has no runtime retry effect — it's only consumed to
pre-register extended model ids — so spreading the full production config
does not mask an isolated model's own failures behind a silent fallback.

**Other diagnostic surfaces checked, unaffected (no fix needed):**
`providers list`/`env`/`explain` already call `configureFromEnv()` before
dispatch and don't accept `--model`, so they never hit the broken branch.

## Test plan

- Added `test/providers-test-model-base-url.test.ts`: drives
  `runProviders('test', ...)` end-to-end (loadConfig → configureGateway →
  gateway → AI SDK → fetch) against a mocked `fetch` + a temp
  `GBRAIN_HOME/.gbrain/config.json` with `provider_base_urls` set for the
  `dashscope` recipe (the exact provider named in the bug report), and
  asserts the outbound request hits the configured base URL rather than
  `dashscope-intl.aliyuncs.com`. A control case confirms bare
  `providers test` already worked correctly.
- Verified the new test is RED on the pre-fix code (`git stash` the src
  change, re-run — fails as expected) and GREEN after restoring the fix.
- `bun test test/providers-test-model-base-url.test.ts test/providers.test.ts test/ai/build-gateway-config.test.ts test/gateway-embed-model-override.test.ts` — 33 pass / 0 fail.
- Codex adversarial review (`gpt-5.6-sol`, high reasoning effort,
  `--base origin/master`) — ran independently, including a 20x repeated
  run of the new test for flakiness — no actionable findings.

Closes #2863

### DIFF
--- a/src/commands/providers.ts
+++ b/src/commands/providers.ts
@@ -7,6 +7,7 @@
 
 import { listRecipes, getRecipe } from '../core/ai/recipes/index.ts';
 import { configureGateway, embedOne, isAvailable as gwIsAvailable, chat as gwChat } from '../core/ai/gateway.ts';
+import { buildGatewayConfig } from '../core/ai/build-gateway-config.ts';
 import { probeOllama, probeLMStudio } from '../core/ai/probes.ts';
 import { loadConfig } from '../core/config.ts';
 import { AIConfigError, AITransientError } from '../core/ai/errors.ts';
@@ -164,8 +165,18 @@ async function runTest(args: string[]): Promise<void> {
     // the divergence at the top of the test so the recovery experience
     // doesn't repeat the bug-reporter's "providers test ✓ but import still
     // broken" trap.
+    //
+    // #2863: `cfg` is lifted out of the try block (not just used for the
+    // warning) so the configureGateway calls below can reuse it. Before this
+    // fix, the --model override only forwarded embedding_model/chat_model +
+    // env, dropping config.provider_base_urls entirely — a probe against a
+    // custom endpoint (e.g. a regional DashScope base URL) would silently
+    // fall back to the recipe's hardcoded default endpoint and fail with a
+    // misleading "Incorrect API key" error even though the key was valid for
+    // the configured endpoint.
+    let cfg: ReturnType<typeof loadConfig> | null = null;
     try {
-      const cfg = loadConfig();
+      cfg = loadConfig();
       const configuredModel = tpArg === 'embedding' ? cfg?.embedding_model : cfg?.chat_model;
       if (!configuredModel) {
         console.error(
@@ -181,17 +192,27 @@ async function runTest(args: string[]): Promise<void> {
       }
     } catch { /* loadConfig throws when no brain configured — first-time install path; the no-config branch above handles it. */ }
 
+    // Reuse the SAME resolver the production path uses (buildGatewayConfig —
+    // also used by cli.ts#connectEngine and init-embed-check.ts) so the probe
+    // sees the identical base_urls / provider_chat_options / folded API keys
+    // that a real `gbrain import`/`gbrain query` call would. Only the
+    // touchpoint's model (+ embedding dims) is overridden on top, so an
+    // isolated `--model` probe still targets exactly the requested model —
+    // it just resolves that model's endpoint the way the brain actually
+    // would. Falls back to bare env when no brain is configured yet (cfg is
+    // null on first-time install, matching the old behavior for that case).
+    const baseGatewayConfig = cfg ? buildGatewayConfig(cfg) : { env: { ...process.env } };
     if (tpArg === 'embedding') {
       const dims = recipe?.touchpoints.embedding?.default_dims ?? 1536;
       configureGateway({
+        ...baseGatewayConfig,
         embedding_model: modelArg,
         embedding_dimensions: dims,
-        env: { ...process.env },
       });
     } else {
       configureGateway({
+        ...baseGatewayConfig,
         chat_model: modelArg,
-        env: { ...process.env },
       });
     }
     void modelId; // intentionally unused but preserved for readability
--- a/test/providers-test-model-base-url.test.ts
+++ b/test/providers-test-model-base-url.test.ts
@@ -1,0 +1,107 @@
+/**
+ * #2863 regression — `gbrain providers test --model` must resolve
+ * `provider_base_urls` the same way the production embed/chat path does.
+ *
+ * Before the fix, the `--model` override branch in `runTest`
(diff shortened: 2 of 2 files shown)


time-attack:
Ported to #3029 (takeover branch from origin/master) — your commits/authorship are preserved; the takeover adds only the master-merge dedupe of the `buildGatewayConfig` import that master's f3e78fd2 introduced. Thanks for the fix!

time-attack:
Superseded with credit by #3029 (just merged): identical change, plus dropping a duplicate import that master picked up via #3000 after you branched. Your commits are preserved in the takeover branch. Thank you!

PR #475, opened 2026-04-27
fix: pass sourceId in cycle sync phase to prevent full reimport

garrytan:
## Problem

`cycle.ts:runPhaseSync()` calls `performSync(engine, { repoPath: brainDir, ... })` without passing `sourceId`. This means the sync phase always reads the **global** `config.sync.last_commit` key instead of the **per-source** `sources.last_commit` anchor.

The global anchor can drift out of git history when:
- A force push rewrites history and the old commit gets garbage-collected
- A rebase squashes commits that included the anchor
- The global key was written by an older gbrain version and never updated

When `git cat-file -t <anchor>` fails, sync concludes there was a force push and triggers a **full reimport of all files** — on a production brain with 78K+ files this takes 30+ minutes and blocks the entire autopilot cycle.

Meanwhile, the per-source anchor in `sources.last_commit` is updated on every successful sync and stays aligned with repo history. It was 668 commits behind HEAD but still valid. The system had the right answer in the right place — the cycle just wasn't reading it.

### Timeline

1. Global `config.sync.last_commit` pointed to commit `472f34d5`
2. That commit was garbage-collected from the repo (likely after a squash/rebase chain)
3. Every autopilot cycle attempted a full reimport of 78,797 files
4. The cycle appeared to "hang" — it wasn't stuck, just scanning every file from scratch
5. `sources.default.last_commit` had a valid ancestor (`00a62e50`) the whole time

## Error Log

The DB state showed the divergence:

## What We Tried

1. **Diagnosed the hang** — ran sync with `--verbose`, saw it scanning 78K files (not stuck, just slow)
2. **Checked both anchors** — found the dual-storage divergence between `config` and `sources` tables
3. **Verified the per-source anchor** — `git cat-file -t` succeeded, `merge-base --is-ancestor` confirmed it
4. **Manual fix** — updated both anchors to HEAD, sync returned `up_to_date` in <1s
5. **Deployed operational guard** — cron script that copies per-source anchor to global every 5 min (band-aid)
6. **This PR** — the proper fix: pass `sourceId` so cycle reads the right anchor

## Solution

**1 file changed: `src/core/cycle.ts`**

Added `resolveSourceForDir()` — queries the `sources` table for a row whose `local_path` matches the brain directory. Returns the source id when found, `undefined` otherwise.

[...]

PR #2983, opened 2026-07-19
feat(dream): --once for one-shot phase runs without toggling config gates (#2860)

### LINKED ISSUE #2860: Feature request: one-shot dream phase run (e.g. `gbrain dream --phase patterns --once`) without toggling the global dream.patterns.enabled flag
## Context

We run gbrain (0.42.59.0, bun global install, macOS) with dream driven by an external
nightly orchestrator rather than the autopilot's own cadence. Because the patterns
phase only runs when the global config `dream.patterns.enabled` is `true`, the only
way to run it once per night is:

1. `gbrain config set dream.patterns.enabled true`
2. `gbrain dream --phase patterns`
3. `gbrain config set dream.patterns.enabled false`

## Problem

That toggle window makes the global flag a cost hazard. The autopilot polls the same
flag, so if the orchestrator dies between steps 1 and 3 (crash, SIGKILL, machine
sleep, power loss), the flag stays `true` and the autopilot re-enqueues the patterns
phase on every cycle. We hit exactly this: a stuck-true flag ran patterns hourly for
~24h — 119 LLM jobs and roughly $400 of API spend before it was caught.

We've since wrapped the toggle in read-back verification, EXIT-trap resets, and an
external watchdog that force-resets the flag — but all of that is scaffolding around
the real gap: there's no way to run a dream phase once without mutating global state.

## Request

A per-invocation override that neither reads nor writes the global flag, e.g.:

    gbrain dream --phase patterns --once
    # or: --force / --ignore-config

Alternatively (or additionally):
- run-once semantics at the config level, e.g. `dream.patterns.enabled = "once"`
  that the autopilot consumes and auto-clears after one pass; or
- a TTL on the enable flag.

## Secondary suggestion

A guardrail in the autopilot for LLM-priced recurring phases would have contained
this class of failure independently: e.g. a warning (or hard stop) when a phase like
patterns runs more than N times per day, or a simple daily job/token budget.

Happy to provide logs or test a patch.

Masashi-Ono0611:
## What

Adds `gbrain dream --phase <name> --once`: a one-shot bypass of that
phase's own `dream.<phase>.enabled` / `cycle.<phase>.enabled` config gate,
for that invocation only. Never reads or writes config.

## Root cause (verified by reading the actual dispatch code, not assumed)

`--phase X` only controls which phase **function** `runCycle` calls in
`src/core/cycle.ts` — it does not bypass that phase's own internal
`.enabled` config read. Each of these phase modules reads its own enabled
flag and skips regardless of how the phase was selected:

- `patterns.ts` — `dream.patterns.enabled`
- `synthesize.ts` — `dream.synthesize.enabled`
- `conversation-facts-backfill.ts` — `cycle.conversation_facts_backfill.enabled`
- `enrich-thin.ts` — `cycle.enrich_thin.enabled`
- `skillopt/cycle-phase.ts` — `cycle.skillopt.enabled`

So with an external orchestrator running `gbrain dream --phase patterns`
on a cadence outside the autopilot, the only way to run patterns once was
the toggle dance: `config set dream.patterns.enabled true` → run →
`config set ... false`. A crash between steps left the flag stuck `true`,
and the autopilot (which polls the same flag) re-enqueued patterns every
cycle — 119 LLM jobs / ~$400 over 24h before it was caught (per the issue).

**`extract_atoms` / `synthesize_concepts` are a different mechanism** —
pack-declaration via `packDeclaresPhase`, not a config `.enabled` read —
and already have a working one-shot escape hatch: `--drain`. The existing
`extract_atoms_backlog` doctor warning already says `--phase extract_atoms
--drain --window 120` (verified directly in `src/commands/doctor.ts:3264`),
so **no doctor text needed updating** — this PR does not touch
`doctor.ts`.

## Design

`gbrain dream --phase <name> --once`. Rejected alternatives (per the
issue's own candidates):

- **Making explicit `--phase X` always bypass `.enabled`**: breaking
  change for existing crons/autopilot installs that rely on the disabled
  flag as a cheap no-op — an upgrade would silently start running LLM /
  write phases that were deliberately turned off.
- **A new subcommand**: adds a whole dispatch/help/arg surface that
  internally routes through the same override anyway — more surface for
  the same behavior.
- **Extending `--once` to also bypass `packDeclaresPhase` for
  `extract_atoms`/`synthesize_concepts`**: conflates two different gating
  mechanisms (config toggle vs. pack membership) under one flag;
  `extract_atoms` already has `--drain`, purpose-built for its
  batched/windowed execution model.

Two hard requirements, both to keep the flag from becoming its own
footgun:

1. **Requires an EXPLICIT `--phase <name>`.** Bare `--once` (full/default
   cycle) has no single phase to target, and force-enabling *every*
   currently-disabled phase in one run would recreate the exact
   unbounded-spend risk the flag exists to prevent. A phase **implied**
   by `--input` (→ synthesize) or `--drain` (→ extract_atoms) does not
   count either — `--drain` returns before the override is ever read, and
   `--input` already bypasses the synthesize gate on its own, so `--once`
   would silently do nothing in both cases. All three (bare, `--input
   ... --once`, `--drain --once`) reject with exit 2 and a usage hint;
   `--help` still short-circuits before that (see IRON RULE below).
2. **Typed as the named `CyclePhase`, not a boolean.** `CycleOpts.onceForPhase?:
   CyclePhase` — each gated phase's dispatch block only honors the
   override when `onceForPhase` matches *its own* phase name, so the
   bypass can never leak to a different phase even if a future
   programmatic caller passes a wider `phases` array than the CLI does.

Never reads or writes config — the phase still evaluates its `.enabled`
gate every call; `--once` only overrides the boolean *outcome* for that
one invocation, mirroring the existing `--unsafe-bypass-dream-guard` /
`--input` precedents in `synthesize.ts` (a stderr line at the bypass
point, no new config-touching code path).

## Test plan

- `test/dream-cli-flags.test.ts` — structural flag wiring (29 tests, incl.
  9 new for `--once`): flag declared, usage-error text, `phaseWasExplicit`
  captured before the `--input`/`--drain` defaulting blocks, threading to
  `runCycle`. **27/27 pass.**
- `test/dream.test.ts` — real `runDream` against a PGLite engine (incl. 6
  new for `--once`): bare `--once` exits 2, `--help --once` prints help
  and exits 0 (not the usage error), `--input <f> --once` exits 2,
  `--drain --once` exits 2, `--phase patterns --once` bypasses
  `dream.patterns.enabled=false` and falls through to
  `insufficient_evidence` (proving the bypass fired) **without writing
  the config key**, and the no-`--once` case still skips as `disabled`.
  **31/31 pass.**
- `test/e2e/dream-patterns-pglite.test.ts` — real `runPhasePatterns`
  against PGLite: `once: true` bypasses `disabled`, config read back
  unchanged after the call.
- `test/core/cycle.serial.test.ts` — `runCycle`-level non-leak proof:
  `onceForPhase: 'patterns'` bypasses patterns; `onceForPhase:
  'synthesize'` (mismatched) does **not** leak the bypass into a
  `phases: ['patterns']` run; no `onceForPhase` set is unchanged;
  config never written.
- Verified 6 of the first 9 new tests **fail** against the pre-fix source
  (`git stash` of source-only changes, confirmed via re-run), proving
  they're meaningful regressions rather than tautologies.

**Pre-existing flake, unrelated to this diff:** `runCycle — sourceId
resolution (regression #475) > sources table missing (very old brain)`
intermittently exceeds bun's default 5000ms test timeout when run
alongside the other 3 heavier PGLite test files concurrently (that test
spins up a *fresh* PGLiteEngine — full ~123-migration chain — then a full
`ALL_PHASES` `runCycle`, with no extended per-test timeout unlike this
file's `beforeAll`). Confirmed pre-existing: the diff to
`cycle.serial.test.ts` in this PR is a pure append after that test's
location in the file (byte-identical to `origin/master` at that test),
and the test passes reliably (100%, incl. isolated single-test and
whole-file runs) whenever run outside the 4-file-concurrent load.

## Codex review

Two rounds against `origin/master`, both addressed:

1. **[P2]** `--help` must short-circuit before the new `--once`
   usage-error exit(2), matching the repo's own pinned IRON RULE
   (`test/dream.test.ts`'s `--help --source whatever` case). Fixed by
   computing `wantsHelp` once in `parseArgs` and exempting the `--once`
   validation when set; added the matching `--help --once` pin test.
2. **[P3]** The `--once` validation checked the *derived* `phase` value,
   which `--input`/`--drain` populate implicitly before the check ran —
   so both silently slipped past the "explicit `--phase`" contract and
   `--once` became a true no-op for them. Fixed by capturing
   `phaseWasExplicit` before any implicit defaulting and validating
   against that instead; added the `--input ... --once` and `--drain
   --once` rejection pins.

Third round: clean (mechanical fix following Codex's own prescription).

Closes #2860

### DIFF
--- a/src/commands/dream.ts
+++ b/src/commands/dream.ts
@@ -76,6 +76,18 @@ interface DreamArgs {
   drain: boolean;
   /** Drain wallclock budget in seconds. Default 300 (5 min). */
   windowSeconds: number;
+  /**
+   * issue #2860 — `--once`. One-shot bypass of the named `--phase`'s own
+   * `dream.<phase>.enabled` / `cycle.<phase>.enabled` config gate, for this
+   * invocation only. Never reads or writes config — unlike the old
+   * "toggle enabled true, run, toggle back to false" workaround, a crash
+   * mid-run can't leave any global state stuck. Requires an explicit
+   * `--phase <name>`; bare `--once` is a usage error (there'd be no single
+   * phase to target). Applies only to phases with a config `.enabled` gate
+   * (patterns, synthesize, conversation_facts_backfill, enrich_thin,
+   * skillopt) — a no-op for phases that always run when named directly.
+   */
+  once: boolean;
 }
 
 const ISO_DATE_RE = /^\d{4}-\d{2}-\d{2}$/;
@@ -214,6 +226,20 @@ function parseArgs(args: string[]): DreamArgs {
     }
   }
 
+  // issue #2860: --once requires an explicit single --phase target. Bare
+  // `--once` (full/default cycle) has no single phase to bypass the gate
+  // for, and force-enabling EVERY currently-disabled phase at once would
+  // be exactly the kind of surprise-spend risk the flag exists to prevent.
+  const once = args.includes('--once');
+  if (once && !phase) {
+    console.error(
+      '--once requires --phase <name> (bypasses that one phase\'s ' +
+      'dream.<phase>.enabled / cycle.<phase>.enabled gate for this run only; ' +
+      'never touches config). Usage: gbrain dream --phase <name> --once',
+    );
+    process.exit(2);
+  }
+
   return {
     json: args.includes('--json'),
     dryRun: args.includes('--dry-run'),
@@ -229,6 +255,7 @@ function parseArgs(args: string[]): DreamArgs {
     source,
     drain,
     windowSeconds,
+    once,
   };
 }
 
@@ -310,6 +337,14 @@ Options:
                       "--dry-run" does NOT mean "zero LLM calls."
   --json              Emit the CycleReport as JSON (agent-readable)
   --phase <name>      Run a single phase: ${ALL_PHASES.join(' | ')}
+  --once              With --phase <name>: run that phase once even if its
+                      own dream.<phase>.enabled / cycle.<phase>.enabled
+                      config gate is false. Never reads or writes config —
+                      unlike toggling the flag on/off around the run, a
+                      crash mid-invocation can't leave it stuck. Applies to
+                      patterns, synthesize, conversation_facts_backfill,
+                      enrich_thin, skillopt; no-op on phases with no such
+                      gate. Requires --phase (bare --once is a usage error).
   --pull              git pull the brain repo before syncing (default: no pull)
   --dir <path>        Brain directory (default: configured brain). On a
                       postgres/remote brain with no local checkout, the
@@ -353,6 +388,7 @@ Examples:
   gbrain dream
   gbrain dream --dry-run --json
   gbrain dream --phase lint
+  gbrain dream --phase patterns --once   # run once, ignore dream.patterns.enabled=false
   gbrain dream --phase synthesize --input ~/transcripts/2026-04-25.txt
   gbrain dream --phase synthesize --from 2026-04-01 --to 2026-04-25
   0 2 * * * gbrain dream --json         # nightly via cron
@@ -594,6 +630,9 @@ export async function runDream(engine: BrainEngine | null, args: string[]): Prom
     synthFrom: opts.from ?? undefined,
     synthTo: opts.to ?? undefined,
     synthBypassDreamGuard: opts.bypassDreamGuard,
+    // issue #2860: opts.phase is guaranteed non-null here when opts.once is
+    // set (parseArgs enforces --once requires --phase).
+    onceForPhase: opts.once ? opts.phase! : undefined,
   });
 
   if (opts.json) {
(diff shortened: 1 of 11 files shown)


time-attack:
Ported to #3031 (takeover branch from origin/master) — your commits/authorship are preserved; the takeover resolves the three additive conflicts against landed #2959 (deadlineAtMs plumbing), keeping both features. Thanks for the feature!

time-attack:
Superseded with credit by #3031 (just merged): your --once flag landed reconciled with the deadline-budget fields #2959 added to the same interfaces. Authorship preserved. Thanks!

PR #2985, opened 2026-07-19
feat(eval): manual-review dismissal ledger for the contradiction probe

Masashi-Ono0611:
## Problem

`resolution_kind: 'manual_review'` exists, but a manual review's outcome has nowhere to live. When a human reviews a flagged pair and rules it a false positive (or an already-resolved temporal artifact), the same finding resurfaces on every read: the doctor `contradictions` check re-warns the same HIGHs daily, `review` re-lists them, MCP `find_contradictions` re-raises them to agents, and the synthesize prior block keeps asking dream to "reconcile" a contradiction a human already rejected.

Real-world trigger: my brain's probe flagged 2 HIGHs; manual review found both were false positives (two deliberately-separate apps on one device judged as "conflicting identity"; a historical measurement the flagged page itself already annotates as resolved). There was no way to record that outcome — the doctor warning is permanent.

## Design

**Identity — versioned, content-derived, judge-config-independent.** `pair_key = sha256("gbrain:contradiction-pair:v1\0" + kind + "\0" + sortedHash(textA) + "\0" + sortedHash(textB))` (order-independent, same sorting as the judge cache).

- model_id / prompt_version / truncation_policy are deliberately excluded: a human verdict about two statements outlives judge upgrades. To force re-evaluation after one, `undismiss` explicitly — auditable, unlike silent expiry.
- Content drift IS the expiry policy: either text changes → hashes change → the ledger row stops matching → the pair re-flags automatically. No TTL column.
- The identity is content-level, not occurrence-level (documented): the human judged the statements, not their location.

**Ledger — `eval_contradictions_dismissals` (migration v124 + PGLite schema parity).** `reason` required (`CHECK (reason ~ '[^[:space:]]' AND length(reason) <= 1000)`), optional `dismissed_by`. `undismiss` is soft-state (`undismissed_at`), so the ledger doubles as an audit trail; re-dismissing reactivates the row.

**Enforcement — one shared read-time projection, four surfaces.** `dismissals.ts` exposes `flattenRunFindings` (union of `contradictions` + runner-parked `dismissed`) and `projectContradictionFindings` (recomputes each finding's pair_key from kind + member texts). Applied at:

1. probe runner (new runs: dismissed pairs leave headline counts and the Wilson-CI numerator; full finding objects kept under `per_query[].dismissed` for audit),
2. doctor `contradictions` check,
3. `review` CLI,
4. MCP `find_contradictions` + the synthesize prior block.

Because keys are recomputed from finding content, dismissals/undismissals take effect **immediately** on all surfaces without re-running the probe, and **pre-ledger report rows project too** — existing HIGHs are dismissable on day one.

Wilson-CI semantics (deliberate): a dismissed pair leaves the **numerator only**; the query stays in the denominator because a human-verified false positive is exactly a clean evaluated sample. `verdict_breakdown` still tallies raw judge behavior.

**CLI.**

```
gbrain eval suspected-contradictions dismiss <pair_id> --reason "..."
gbrain eval suspected-contradictions undismiss <pair_id>
gbrain eval suspected-contradictions review [--include-dismissed]
```

`pair_id` (first 12 hex of pair_key) is printed next to findings in `review` and doctor HIGH lines; hex-prefix resolution rejects ambiguity instead of guessing. `dismiss` requires the prefix to match a finding in a run from the last 90 days (that's where the texts/hashes come from), so typos can't create dead ledger rows.

## Verification

- `test/eval-contradictions-dismissals.test.ts` (25 cases): identity order-invariance / kind separation / content-drift re-flag; projection matches pre-ledger rows by recomputed key; text-less shapes always surface; undismiss restores runner-parked findings via union projection; PGLite CRUD incl. soft-revoke audit row, reactivation, prefix ambiguity rejection, whitespace/overlong reason CHECKs; runner integration (headline counts + CI numerator + `per_query.dismissed` + pair_id stamping); CLI flag parsing.
- Full contradiction suite: 292/292 across 21 files (incl. MCP integrations + synthesize prior block). `tsc --noEmit` clean. `bun run verify` 31/31.
- Follow-up (out of scope, noted): skip judge calls (and their cost) for ledger-active pairs at probe time; event-sourced dismissal history if single-row audit proves insufficient.

https://claude.ai/code/session_01Hy4wZcx6ErDVBxYFerAsTQ


time-attack:
Thanks — the pair_key design (content-derived, judge-config-independent, drift-as-expiry) is genuinely thoughtful and you clearly understood the judge cache. It's still ~1.1k lines of new subsystem including a schema migration and parallel engine methods, added so an eval probe stops re-warning on two reviewed findings; that's a permanent storage contract for a suppression feature. Closing under the feature-surface and size policy rather than on quality.

Masashi-Ono0611:
Understood — a permanent storage contract (schema migration + parallel engine methods) is a heavy price for suppressing two reviewed findings. Thanks for taking the time to actually evaluate the `pair_key` design before closing on size.

PR #2988, opened 2026-07-19
fix(pages): default chunker_version to MARKDOWN_CHUNKER_VERSION on INSERT (#2807)

### LINKED ISSUE #2807: putPage defaults chunker_version to 1 on INSERT, causing perpetual doctor warning
`postgres-engine.ts:990` uses `COALESCE(${chunkerVersion}::smallint, 1)` in the INSERT. When callers don't supply `chunker_version` (which no MCP/subagent caller does — it's internal metadata), new pages land at version 1.

The `importFromFile` path correctly stamps `MARKDOWN_CHUNKER_VERSION` (currently 3), but dream subagents write through `putPage` directly — their pages get v1. Doctor then flags them as "older chunker_version" in the `contextual_retrieval_coverage` check forever.

The pages are actually chunked and embedded correctly at write time (embedding coverage stays 100%). Only the metadata column is wrong.

**Impact:** On an active brain with dream running twice daily, doctor fires `contextual_retrieval_coverage: warn` continuously. Running `reindex --markdown` clears it, but it recurs the next cycle.

**Fix (one line):**

```diff
- COALESCE(${chunkerVersion}::smallint, 1)
+ COALESCE(${chunkerVersion}::smallint, ${MARKDOWN_CHUNKER_VERSION})
```

Import `MARKDOWN_CHUNKER_VERSION` from `../core/chunkers/recursive.ts` (already imported elsewhere in the codebase).

The COALESCE-preserve on UPDATE is correct (don't regress pages already at a higher version). Only the INSERT default needs the bump.

JOJOCrazy123:
## Summary

`putPage`'s INSERT defaulted `chunker_version` to the literal `1`:

```ts
COALESCE(${chunkerVersion}::smallint, 1)
```

Callers that don't supply `chunker_version` (no MCP/subagent caller does — it's internal metadata) landed new pages at **v1**. Dream subagents write through `putPage` directly, so their pages got v1 and `doctor`'s `contextual_retrieval_coverage` check flagged them as "older chunker_version" **forever** — even though the pages were actually chunked and embedded with the current chunker (embedding coverage stays 100%). `reindex --markdown` clears it, but the next dream cycle brings it back.

## Fix

Default the INSERT to `MARKDOWN_CHUNKER_VERSION` (currently 3) on **both** engines (`postgres-engine.ts`, `pglite-engine.ts`), importing the constant from `core/chunkers/recursive.ts`. The `ON CONFLICT ... UPDATE` COALESCE-preserve is left untouched, so an explicitly supplied version is still honored and never lowered below the current version.

## Test plan

- [ ] `bun test test/chunker-version-insert-default.test.ts`
  - page written without `chunker_version` → lands at `MARKDOWN_CHUNKER_VERSION`, not 1
  - explicit `chunker_version` honored on INSERT
  - re-put without `chunker_version` does not lower an already-current version
- [ ] `bun test test/pglite-engine.test.ts` — existing putPage behavior unchanged.

> Note: I could not run the suite locally (this environment has no `bun` runtime or installed deps). The change is a two-line INSERT-default swap mirrored across both engines plus an engine-level regression test; CI will validate.

Closes #2807.

### DIFF
--- a/src/core/pglite-engine.ts
+++ b/src/core/pglite-engine.ts
@@ -23,6 +23,7 @@ import { runMigrations } from './migrate.ts';
 import { PGLITE_SCHEMA_SQL, getPGLiteSchema } from './pglite-schema.ts';
 import { DEFAULT_EMBEDDING_MODEL, DEFAULT_EMBEDDING_DIMENSIONS } from './ai/defaults.ts';
 import { DELETE_BATCH_SIZE } from './engine-constants.ts';
+import { MARKDOWN_CHUNKER_VERSION } from './chunkers/recursive.ts';
 import { acquireLock, releaseLock, type LockHandle } from './pglite-lock.ts';
 import { getFtsLanguage } from './fts-language.ts';
 import type {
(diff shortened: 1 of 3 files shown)


JOJOCrazy123:
Hi maintainers, this PR is ready for review, but the workflow is still awaiting maintainer approval. Could someone please approve the workflow and review the PR when convenient? Thank you!

PR #2990, opened 2026-07-19
fix: add GBrain autopilot cost guardrails

gidonco:
## Summary

- enforce a hard two-hour minimum between autopilot cycles across launchd, systemd, and cron installs
- make autonomous jobs single-attempt with two-hour-to-one-day failure cooldowns and zero SDK retries by default
- add an atomic, database-backed daily AI budget governor covering gateway chat, embeddings, expansion/OCR, and reranking
- charge crashed or expired reservations pessimistically so failures cannot restore budget headroom
- remove provider credentials from generated launchd configuration and make fatal exits stay stopped

## Why

The previous controls could amplify costs in several layers at once: low health shortened the autopilot interval, service managers relaunched failed processes, autonomous jobs retried, and the provider SDK also retried. Budget reservations were not fully atomic and expired reservations restored their estimated headroom, so crashes could weaken the cap.

This change treats cadence, retries, and spend accounting as independent safety boundaries. When `ai.daily_budget_usd` is configured, every priced gateway call reserves shared daily headroom before transport and fails closed if pricing or the ledger is unavailable.

## Impact

- generated autopilot installs use `--interval 7200 --min-interval 7200`
- launchd uses `KeepAlive=false`; systemd uses bounded failure-only restarts
- autonomous jobs use one attempt and provider SDK calls default to zero retries
- `ai.daily_budget_usd` is a registered configuration key; absent or zero preserves backward compatibility
- daily accounting is shared across GBrain processes and uses pessimistic settlement on provider or process failure

## Validation

- focused guardrail and gateway suite: 101 passed, 0 failed
- `bun run typecheck`
- `bun run verify`: 31/31 checks green
- `git diff --check`
- four parallel embedding-dimension failures from the broad suite passed in isolation (21/21)
- three remaining legacy fixture failures reproduce unchanged on the `origin/master` baseline (`hybrid-meta.serial` and `worker-registry.serial`)

### DIFF
--- a/docs/superpowers/plans/2026-07-18-gbrain-cost-guardrails.md
+++ b/docs/superpowers/plans/2026-07-18-gbrain-cost-guardrails.md
@@ -1,0 +1,131 @@
+# GBrain Cost Guardrails Implementation Plan
+
+> **For Codex:** REQUIRED SUB-SKILL: Use superpowers:executing-plans to implement this plan task-by-task.
+
+**Goal:** Keep GBrain autopilot to one maintenance cycle every two hours or less, stop automatic retry amplification, and enforce a persistent $1.50/day ceiling for gateway-routed paid AI calls.
+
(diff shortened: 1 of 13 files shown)


gidonco:
Closing because this change is intended for the contributor fork only. The replacement PR is gidonco/gbrain#1.

Issue #2367, opened 2026-06-23
normalizeBasename strips CJK (Korean) characters, breaking cross-folder wikilink resolution


Febrile-JoJoo:
## Summary

`normalizeBasename()` in `src/core/link-extraction.ts` uses the regex `/[^a-z0-9\s-]/g` which strips all non-ASCII characters including CJK (Korean, Japanese, Chinese). This causes `[[wikilink]]` resolution to fail for files with CJK filenames when the link crosses folder boundaries.

## Environment

- GBrain version: 0.42.52.0
- Engine: PostgreSQL
- OS: WSL (Ubuntu)
- Language: Korean filenames

## Steps to Reproduce

1. Create files with Korean filenames in different folders:
   ```
   wiki/pages/에버노트-루카텍-아카이브.md    (contains: [[루카텍 올핸즈 미팅]])
   03-Resources/루카텍/250814 목 - 루카텍 올핸즈 미팅.md
   ```

2. Run `gbrain extract links`

3. Expected: link created from `wiki/pages/에버노트-루카텍-아카이브` to `03-resources/루카텍/250814-목-루카텍-올핸즈-미팅`

4. Actual: 0 links created. The wikilink `[[루카텍 올핸즈 미팅]]` is not resolved.

## Root Cause

In `src/core/link-extraction.ts` line 836:

```typescript
export function normalizeBasename(s: string): string {
  return s.toLowerCase().replace(/[^a-z0-9\s-]/g, "").trim().replace(/\s+/g, "-");
}
```

This regex `[^a-z0-9\s-]` removes ALL characters except ASCII alphanumerics, spaces, and hyphens. For Korean text like "루카텍 올핸즈 미팅", this produces an empty string `""`, which never matches any slug.

Meanwhile, `slugifySegment()` in `src/core/sync.ts` correctly preserves CJK characters using `SLUGIFY_KEEP_RE`:

```typescript
const SLUGIFY_KEEP_RE = new RegExp(`[^a-z0-9.\\s_\\-${CJK_SLUG_CHARS}]`, "g");
```

So slugs ARE created with CJK characters preserved, but basename RESOLUTION strips them — creating a mismatch.

## Same-folder links work (ancestor walk)

Links within the same folder work because `resolveSlug()` uses ancestor path walking (joining `fileDir + relTarget`), which doesn't go through `normalizeBasename`. Only cross-folder `[[bare-name]]` wikilinks fall back to basename resolution.

## Proposed Fix

[...]

Issue #1638, opened 2026-05-30
CJK (Chinese/Japanese/Korean) entity names not extracted by mention scanner


alkalide:
## Problem

The mention scanner (`--by-mention`) cannot extract CJK entity names because:

1. `TOKEN_RE = /[a-zA-Z0-9]+/g` skips all CJK characters (line 118)
2. `MIN_NAME_LENGTH = 4` filters out 2-3 character CJK names (line 42)
3. `tokenizeTitle()` returns empty array for pure CJK titles, excluding them from gazetteer

This affects any brain containing Chinese, Japanese, or Korean named entities (people, companies, concepts).

## Reproduction

```bash
# Create a CJK entity page
echo '---
title: 纳瓦尔
type: person
---
Naval Ravikant, in Chinese context.' | gbrain put people/naval

# Scan for mentions — finds 0 links even though 纳瓦尔 appears in other pages
gbrain extract links --by-mention --source db
```

## Proposed Fix

PR #1637 adds CJK detection helpers, lowers min name length for CJK titles to 2 chars, handles pure CJK titles in the tokenizer, and adds a substring matching pass for CJK entities in `findMentionedEntities`.

## Verified

```
$ gbrain extract links --by-mention --source db
Mentions: created 27 links from 456 pages against gazetteer of 3 first-token buckets
```

The code comment at line 109-111 explicitly notes CJK was deferred: `"Non-ASCII (CJK, accented) is deliberately not tokenized in v1"` — this PR implements that deferred support.

time-attack:
Closing as duplicate of the open fix PR #1637.

**Why:** This issue and PR #1637 describe the same problem; keeping one thread. Work continues on #1637.

_Triage: issue↔PR pair — closed the issue, left the fix PR open._

Issue #648, opened 2026-05-05
fix(chunker): CJK-aware recursive prose chunker (current breaks on no-space CJK paragraphs, produces chunks > 8192 tokens)


chengzehsu:
## Bug

`src/core/chunkers/recursive.ts` `countWords()` returns `(text.match(/\S+/g) || []).length`. For CJK paragraphs that lack whitespace (Chinese / Japanese / Korean prose), this counts the entire paragraph as **1 "word"**.

With `chunkSize=300`, a 5000-character Chinese page returns `1 word ≤ 300`, so `chunkText` returns the entire content as ONE chunk. That single chunk then goes to OpenAI embeddings which hard-rejects input > 8192 tokens.

**Result in our brain**: 158 pages have chunks created with `embedding = NULL` — vector search misses them entirely, BM25 still works.

The same repo already has CJK-aware word counting in `src/core/search/expansion.ts` — proven by tests in `test/search.test.ts` *"CJK word count in expansion > CJK characters are counted individually"*. So the fix mirrors that existing logic.

## Why fixing `countWords` alone is INSUFFICIENT

A third-perspective review (Codex) flagged that just patching word-counting still leaves the chunker broken:

1. Even with CJK-aware word count, the `DELIMITERS` hierarchy is **ASCII-only** (`. `, `! `, `? `, `; `, `: `, `, `). Dense CJK prose has **none** of these — it has `。！？；：，、` and CJK newline variants. So the chunker correctly detects "too big" but has no boundary to split on.
2. The level-4 fallback `splitOnWhitespace` **also fails** on no-space CJK — same `\S+` regex, same problem.
3. **Critical**: 158 existing affected pages will NOT auto-recover. `importFromContent` short-circuits on unchanged `content_hash` (line 227), and `embed --stale` reuses existing chunks. Need a `PROSE_CHUNKER_VERSION` constant folded into the markdown content_hash so existing pages re-chunk on next sync.

## Proposed 4-piece fix

1. **`countWords()` mirror `expansion.ts:56` logic** — CJK chars count individually
2. **`DELIMITERS` add CJK punctuation** — `。`, `！`, `？`, `；`, `：`, `，`, `、` + newline variants
3. **Replace level-4 `splitOnWhitespace` fallback with script-aware fallback** — whitespace-words for spaced languages, character-/codepoint-slicing for CJK / no-space text
[...]

PR #2992, opened 2026-07-19
fix(extract): add CJK entity extraction — coverage filter, pure-digit filter, and CJK substring matching pass

justemu:
# PR: Add CJK entity extraction for Chinese/Japanese/Korean names

## Problem

The `by-mention` gazetteer tokenizer (`TOKEN_RE = /[a-zA-Z0-9]+/g`) is ASCII-only. This causes two related problems:

1. **Zero recall for CJK entities**: Pure CJK titles like "阿里巴巴" (Alibaba), "辉瑞制药" (Pfizer China), "騰訊" (Tencent HK) produce zero tokens from `tokenizeTitle()`. These entities are silently dropped from the gazetteer, making them invisible to the auto-link mention scanner.

2. **False-positive matches from mixed titles**: Mixed CJK/ASCII titles like "浪潮集团2026上半年工作总结会议" tokenize to `["2026"]` (4/17 = 24% character coverage). Any body text containing "2026" anywhere then incorrectly links to this page. Same for titles like "腾讯Tencent" or "2019-nCoV疫情分析".

**Severity**: In CJK-dominant brains (Chinese, Japanese, Korean user bases), entity-level auto-linking produces zero true positives and many false positives for CJK entities. The mention scanner is effectively broken for these users.

## Solution

Three changes to `src/core/by-mention.ts`:

### P2 Filters (gazetteer construction)

1. **Coverage-ratio filter** (line ~210): Drop entries where `tokens.join('').length / title.length < 0.5`. Mixed CJK/ASCII titles where the tokenizable (ASCII) portion is less than half the title are excluded. Eliminates the "浪潮集团2026..." false-positive class.

2. **Pure-digit single-token filter** (line ~215): Drop entries where `tokens.length === 1 && /^\d+$/.test(tokens[0])`. Prevents bare-year entries like "2026" or "2019" from matching every occurrence of that number in body text.

### P3 CJK Substring Matching (new matching pass)

3. **CJK substring matching** (line ~200-260, ~370-420):
   - Pure CJK titles bypass the ASCII tokenizer entirely and are stored with a `cjkSubstrings` field containing the full title string.
   - Mixed titles enter both the ASCII token path and get a `cjkSubstrings` field.
   - A new CJK substring pass runs after the ASCII maximal-munch loop. It walks each character offset in the body text, finding the longest CJK entry that matches as a substring at that offset (maximal-munch for CJK).
   - Same guards as the ASCII pass: self-link skip, cross-source isolation, first-mention dedup.
   - Gated by `CJK_MATCH_MODE: CjkMatchMode` (default `'substring'`, can be `'disabled'`).

### Design rationale

- **Separate pass, not unified tokenizer**: CJK has no word boundaries — a CJK tokenizer would require word segmentation (jieba, mecab, etc.) with language detection and dictionary dependencies. A substring pass is simpler, dependency-free, and covers all CJK languages uniformly.
- **No false-positive regression**: The P2 filters only drop entries (never add). The P3 pass finds only exact-title matches. Zero new false positives.
- **Configurable**: `CJK_MATCH_MODE` allows disabling for ASCII-only brains or future upgrades (Trie, Aho-Corasick).

## Testing

- Verified on a 36k-page production brain with ~2k CJK entity pages.
- Before: zero CJK entities found by `--by-mention --dry-run --type company`.
- After: `艾迪药业`, `腾盛博药`, `住友制药`, `AMCHEPRY（iPSC帕金森病再生医疗产品）` all correctly matched.
- False-positive rate: zero new. P2 filters eliminated existing false positives from mixed titles.
- Performance: ~5 pages/sec with dual pass (acceptable for the `--by-mention` full-brain scan which is already O(N*M) per page).

## Related

- Closes #1637 (feat: CJK entity extraction for Chinese/Japanese/Korean names)
- References #1638 (closed as duplicate)
- References #2367 (CJK characters stripped by normalizeBasename)
- References #648 (CJK-aware chunker — related but separate concern)

### DIFF
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -2736,6 +2736,20 @@ export async function chat(opts: ChatOpts): Promise<ChatResult> {
     providerOptions.anthropic = { cacheControl: { type: 'ephemeral' } };
   }
 
+  // Disable thinking/reasoning for all openai-compatible providers.
+  // DeepSeek v4-flash defaults to thinking mode, producing ~50% unnecessary
+  // output tokens for gbrain's non-reasoning use cases. Other providers
+  // (Ollama, OpenRouter, Groq, Together, llama-server) may add thinking in
+  // the future — the parameter is silently ignored by unrecognizing providers.
+  // Does not affect native providers (OpenAI o-series, Anthropic Claude,
+  // Google Gemini) which use different parameter names for reasoning control.
+  if (recipe.implementation === 'openai-compatible') {
+    providerOptions.openaiCompatible = {
+      ...((providerOptions as any).openaiCompatible ?? {}),
+      thinking: { type: 'disabled' as const },
+    };
+  }
+
   let _budgetRecorded = false;
   const _recordBudget = (modelLabel: string, inputTokens: number, outputTokens: number): void => {
     if (!tracker || _budgetRecorded) return;
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -709,7 +709,8 @@ export function diagnoseEmbedding(modelOverride?: string): EmbeddingDiagnosis {
   if (
     Array.isArray(tp.models) &&
     tp.models.length === 0 &&
-    (recipe.id === 'litellm' || isUserProvided)
+    (recipe.id === 'litellm' || isUserProvided) &&
+    !parsed.modelId
   ) {
     return {
       ok: false,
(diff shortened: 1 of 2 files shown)


time-attack:
Thanks for digging into CJK recall — the ASCII-only tokenizer gap is a fair observation. I can't take this as-is though: it bundles an unrelated AI-gateway change (thinking-mode disable plus an embedding-diagnosis tweak) with the by-mention work, ships no tests for a pass that silently changes gazetteer behavior for existing mixed-title pages, and leaves agent-run artifacts and mixed-language comments in the committed code. Substring matching without word boundaries also trades one false-positive class for another, which needs evidence before it can default on.

PR #2997, opened 2026-07-19
fix(takes): default takes extraction to the configured chat_model instead of hardcoded cloud Haiku

Nazim22:
## Problem

`extract-takes-from-pages.ts` hardcodes `anthropic:claude-haiku-4-5` as the takes-extraction model. On an OAuth/local-only install (no `ANTHROPIC_API_KEY`; chat routed through a local gateway model), every takes extraction dies with `llm_unavailable` — the takes layer silently never populates and the `takes_count` health check stays red, even though the brain has a perfectly good configured `chat_model`.

## Fix

Default the takes-extraction model to the configured `chat_model` (resolved through the existing model-config chain), keeping `anthropic:claude-haiku-4-5` as the final fallback when no chat model is configured. Explicit `opts.model` still wins. One file, 6 lines.

Verified live on a Postgres brain with an ollama-gateway chat model: 586 takes generated where extraction previously failed 100% of the time; `takes_count` health check now OK.

## Tests

No behavior change for cloud-key installs (fallback preserved). Suite: 304 takes-related tests pass; `bun run verify` all 31 checks green; typecheck clean.

### DIFF
--- a/src/core/extract-takes-from-pages.ts
+++ b/src/core/extract-takes-from-pages.ts
@@ -133,6 +133,11 @@ export async function extractTakesFromPages(
   const dryRun = opts.dryRun ?? false;
   const maxPages = opts.maxPages ?? 50;
   const holder = opts.holder ?? 'system';
+  // S302 (Pingu, local-only fix): default takes to the configured local
+  // chat_model instead of hardcoded cloud Haiku — this brain is OAuth/local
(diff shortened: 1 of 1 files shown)


Nazim22:
Closing in favor of #3021, which lands the same fix on the correct config plane (`getChatModel()` / file-plane gateway config instead of the DB plane). Thanks @time-attack for carrying it over the line.

PR #2998, opened 2026-07-19
feat(ops): reflex_pointers op — Retrieval Reflex over MCP transport

Nazim22:
## Problem

The Retrieval Reflex pointer layer (#1981) is only reachable through the context engine's `assemble()` path. Hosts that consume the brain via MCP ops — per-prompt hooks, non-context-engine agents, thin remote harnesses — have no way to get reflex pointers for a prompt, so the pointer layer is invisible to exactly the integration surface MCP exists for.

## Change

New contract-first `reflex_pointers` op in `src/core/operations.ts` (read scope): bridges the pointer layer to MCP/CLI transports by reusing `buildReflexAddition` through the host-resolver arm, bound to the op's already-open engine and source scope — so the gate, timeout, and heartbeat behave identically to the context-engine path. No new pipeline, no duplicated logic; one file, 35 lines.

Being contract-first, the op appears in CLI, MCP server, and tools-json automatically.

Running in production on our install since June (per-prompt recall hook consuming it over MCP-HTTP).

## Tests

Contract parity pinned by the existing generators: `test/parity.test.ts` 10 pass / 1297 assertions with the new op present. `bun run verify` all 31 checks green; typecheck clean. Happy to add a dedicated op test if you'd like one — the op is a thin binding over the already-tested `buildReflexAddition`.

### DIFF
--- a/src/core/operations.ts
+++ b/src/core/operations.ts
@@ -2633,6 +2633,40 @@ const get_raw_data: Operation = {
 
 // --- Resolution & Chunks ---
 
+const reflex_pointers: Operation = {
+  name: 'reflex_pointers',
+  description:
+    'Retrieval Reflex (#1981) over an op transport: extract salient entities from a turn of ' +
+    'user text and resolve them to compact page pointers. Same pipeline, gate, timeout and ' +
+    'heartbeat as the context-engine layer — for hosts (per-prompt hooks, non-context-engine ' +
+    'agents) that consume the brain over MCP instead of assemble().',
+  params: {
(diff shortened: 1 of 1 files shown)


time-attack:
Thanks — this is clean and small, and I understand the motivation for reaching the pointer layer from outside the context engine. Contract-first means a new op is permanent public surface across CLI, MCP, and tools-json, and we're deliberately not growing the op catalogue for integration convenience right now. Nothing is broken without it, so I'm closing on the feature-surface policy rather than on quality.

PR #3015, opened 2026-07-20
Add safe GBrain maintenance automation

jdewoski-cmd:
## Summary
- centralize orphan exclusion policy so health and orphan reporting use the same convention filters
- align health stale-page counting with the link extractor stale watermark
- add `gbrain maintain` with dry-run-by-default and `--safe` automation for stale extraction plus stale source-cycle remediation
- add regression coverage for convention exclusions, health orphan filtering, and maintain safety parsing

## Verification
- `bun test test/maintain.test.ts`
- `bun test test/orphans-pure-fn.test.ts`
- `bun test test/orphans.test.ts`
- `bun test test/maintain.test.ts test/orphans-pure-fn.test.ts test/orphans.test.ts` — 61 pass / 0 fail
- `bun run typecheck`
- `bun run check:progress`
- `gbrain maintain --dry-run --json`
- `gbrain maintain --safe` — no-op on current clean live brain: stale=0, cycle_freshness ok

## Notes
- `gbrain maintain --safe` intentionally does not mutate source frontmatter, apply schema-pack upgrades, or create semantic hub links. Those remain review/proposal work.
- Upstream push was denied for `jdewoski-cmd`, so this PR is from a fork branch.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -54,7 +54,7 @@ export function bigintToStringReplacer(_key: string, value: unknown): unknown {
 }
 
 // CLI-only commands that bypass the operation layer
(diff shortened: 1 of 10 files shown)


time-attack:
Thank you for the thorough work here — the tests and the dry-run-by-default posture are genuinely well considered. This is new user-facing command surface plus a new automation lane rather than a fix for something broken, and we are holding a hard line against adding feature surface right now, so I am closing it. If the orphan-exclusion and stale-counting alignment pieces are papering over an actual incorrect count you can demonstrate, a small standalone PR with just that correctness fix and its regression test would be a different conversation.

PR #3018, opened 2026-07-20
fix(extract): clear version-stale watermarks

aleks-bcap:
## Summary

`gbrain extract --stale` currently stamps every processed page with that page's `updated_at`. When a page predates `LINK_EXTRACTOR_VERSION_TS`, this leaves `links_extracted_at < LINK_EXTRACTOR_VERSION_TS` true forever: each sweep successfully reprocesses the page but can never clear the version-invalidation arm.

This change stamps each successfully processed page with the later of:

- its exact database-projected `updated_at_iso`, preserving the existing concurrent-edit and microsecond-precision guarantees; and
- `LINK_EXTRACTOR_VERSION_TS`, clearing version invalidation for older unchanged content.

## Coverage

- Adds a regression that backdates a page before the extractor version, runs `extract --stale`, and proves both stale predicates clear.
- Retains the existing microsecond-precision and concurrent-edit regressions.
- Updates the key-file contract to document the two-arm watermark invariant.

## Local verification

- `bun test test/extract-stale.test.ts` — 15 pass, 0 fail.
- `bun run verify` — 31 checks passed.
- `bun run test` — 7,292 tests passed and no assertion failure was reported; the four-shard runner ultimately exited nonzero after shard 1 was killed with rc=137 (resource exhaustion) during the 3,726-second run. The other serial tests completed successfully. This remains draft for upstream CI.
- Docker-backed real-Postgres E2E could not run on this host because access to `/var/run/docker.sock` is denied; upstream CI should supply that gate.


time-attack:
Thank you — your analysis of the version-staleness arm never clearing was exactly right, but master already carries this fix as of 6cf8d8d6 (#1791), which stamps the later of updated_at_iso and LINK_EXTRACTOR_VERSION_TS with the same concurrent-edit semantics you describe. Nothing to change on the current tree, so I am closing this. The draft status and the unrunnable Postgres e2e were also blockers, but the already-fixed status is the deciding one.

PR #3021, opened 2026-07-20
fix(takes): default takes extraction to the configured chat_model instead of hardcoded cloud Haiku (takeover of #2997)

time-attack:
## Summary

Takeover of #2997 by @Nazim22 (credited via Co-authored-by). The diagnosis and shape of the fix are the original author's; this branch swaps the config plane the model default is read from and adds the regression test.

**The bug (from #2997):** `extract-takes-from-pages.ts` hardcoded `anthropic:claude-haiku-4-5`. On OAuth/local-only installs (no `ANTHROPIC_API_KEY`; chat routed through a gateway model) every takes extraction died with `llm_unavailable` — the takes layer never populated and the `takes_count` health check stayed red despite a working configured chat model.

**Changed from the original PR:** #2997 read the default via `await engine.getConfig('chat_model')` — the brain's **DB** config plane. Model routing in this codebase lives on the **file-plane** gateway config (`src/core/model-config.ts` chain), read through `getChatModel()`; e.g. `src/commands/enrich.ts:425`'s `opts.model || getChatModel()`. Reading the DB plane would have introduced a second, divergent source of truth for the chat model. This branch uses the codebase idiom:

```ts
model: opts.model || getChatModel(),
```

Explicit `opts.model` still wins; unconfigured installs fall through to the gateway's `DEFAULT_CHAT_MODEL`.

**Added:** `test/extract-takes-model-resolution.test.ts` pins the plane, not just the fallback:
- gateway-configured `chat_model` is used when `opts.model` is unset
- a conflicting DB-plane `config.chat_model` row is **ignored**
- explicit `opts.model` wins

Verified the file-plane test fails against the pre-fix code (1 fail) and passes with the fix.

## Test evidence

- `bun test test/extract-takes-model-resolution.test.ts test/extract-takes-from-pages-progression.test.ts` — 6 pass / 0 fail
- Takes-adjacent suite (14 files incl. `test/e2e/takes-postgres.test.ts` on real Postgres): 190 pass / 1 batch-contention timeout flake that passes in isolation both with and without the fix (0 fail on re-run)
- `bun run typecheck` — clean

## Credit

Root-cause diagnosis, live verification on an ollama-gateway brain, and the resolution-order design by @Nazim22 in #2997.

### DIFF
--- a/src/core/extract-takes-from-pages.ts
+++ b/src/core/extract-takes-from-pages.ts
@@ -15,7 +15,7 @@
 
 import type { BrainEngine } from './engine.ts';
 import type { TakeBatchInput, TakeKind } from './engine.ts';
-import { chat, isAvailable } from './ai/gateway.ts';
+import { chat, getChatModel, isAvailable } from './ai/gateway.ts';
 
 export const ALLOWED_PAGE_TYPES = [
   'concept', 'atom', 'lore', 'briefing', 'writing', 'originals',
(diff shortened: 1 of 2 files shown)


Nazim22:
Thanks @time-attack — both for taking this over and for the generous credit. The file-plane correction is right: `getChatModel()` is the codebase idiom, and reading the DB plane (as my original patch did) would have introduced a second source of truth for the chat model. The regression test pinning the plane rather than just the fallback is a nice touch. +1 from me — I'll close #2997 in favor of this.

Issue #2928, opened 2026-07-17
`sources unfederate` / `config.federated=false` has no effect on unqualified `query`/`search`/`think` retrieval — an "isolated" source's pages still surface in cross-source results


skeptikoss:
**Version**: 0.42.59.0, SHA 5008b287 (no releases exist, installed via pinned master SHA: `github:garrytan/gbrain#5008b287e47bf791132eedfebf66bdef11e9398c`)

**Related**: this looks like the same root cause as #2561 ("`sources.config.federated` ignored by unqualified local CLI search/query"), but from the opposite direction — #2561 reports that a source with `federated: true` fails to *join* unqualified search; this report is that a source explicitly marked `federated: false` ("isolated") is *not excluded* from it. Worth triaging together since both point at the same dead code path.

## What happens

Marking a source isolated (`gbrain sources unfederate <id>`, or registering with `--no-federated`) sets `sources.config.federated = false` in the database and `gbrain sources list` correctly displays it as `isolated`. But the flag is never read anywhere on the retrieval path: an unqualified `gbrain query "…"` / `gbrain search "…"` / `gbrain think "…"` (no `--source`) still returns pages that live in the "isolated" source, mixed in with the resolved default source's results, ranked purely by relevance score.

## What I expected

A source with `config.federated = false` should never appear in an unqualified cross-source search — that is the entire purpose of the flag (and of the `isolated` label `sources list` prints for it).

## Where

- `src/core/sources-load.ts:44-45` — `LoadAllSourcesOpts.federatedOnly` is declared ("Only return sources with `config.federated === true`") but:
- `src/core/sources-load.ts:100-103` — the filter only runs `if (opts.federatedOnly)`, and grepping every call site of `loadAllSources` in the tree (`collect-uninstalled-brain-pack.ts`, `brain-resident-locate.ts`, `serve-http.ts`, `commands/sources.ts`, `commands/doctor.ts`, `commands/autopilot.ts`) shows **none of them ever pass `federatedOnly: true`**. The option is unused dead weight.
- `src/core/operations.ts:417-425` (`sourceScopeOpts`) and `:486-505` (`resolveRequestedScope`) are the two functions that decide the source scope for every read op (`query`, `search`, `think`, `get_page`, …). Both branch only on `ctx.auth?.allowedSources` (a remote OAuth grant) and `ctx.sourceId` (the CLI-resolved source). Neither reads `sources.config` at all, so `config.federated` never enters the decision.

```ts
[...]

Issue #2884, opened 2026-07-16
RFC: make source scope un-forgettable — required scope at the engine boundary, loud-refusal writes


spinsirr:
> Posting as RFC — feedback / redirection welcome. This is a design question, not a PR; if the framing is wrong, "here's how I think about it instead" is the most useful reply.

## TL;DR

Cross-source scope leaks keep recurring as point fixes because scope enforcement is **opt-in at every call site**: op handlers must remember to call `sourceScopeOpts` / `linkReadScopeOpts` / the per-call resolver, and every scope parameter on the `BrainEngine` interface is optional. This RFC proposes making the compiler enforce what convention currently doesn't:

1. **Resolve scope once at the dispatch boundary** and stamp it on `OperationContext` as a sealed, already-collapsed value (extends #2881's guard).
2. **Make `scope` a required parameter on source-scoped `BrainEngine` read methods**, with an explicit `Scope.ENTIRE_BRAIN` sentinel for trusted internal callers — forgetting becomes a compile error, migrated in per-method-family waves.
3. **Writes refuse instead of defaulting**: a write without resolved source attribution errors loudly instead of silently landing in `default`.
4. **RLS (#2387) as the SQL-layer backstop** for anything the app layer still misses.

## The evidence that this is a class, not a series of bugs

`operations.ts` itself says it: *"drift between sites is the bug class"* (comment on `sourceScopeOpts`). Instances so far:

**Read-side leaks** (remote caller sees out-of-grant data) — each fixed point-wise:

| Instance | Surface | Fix |
|---|---|---|
| #1336 | legacy tokens ignored `permissions.source_id` | #1999 |
| #1371 | explicit `source_id` bypass in query | #1999 |
| #1924 | `source_id: "__all__"` bypass (query, code_callers/callees) | #1999 |
| #1393 | `get_page` exact-slug path missed federated grant | #1999 |
| #2200 | by-slug reads (tags/links/backlinks/timeline) missed federated grant | #2239 |
| takes reads | `takes_search` MCP path unscoped | #2452 (open) |
| ingest log | `get_ingest_log` returned whole-brain log to any read-scoped client | #2881 (open) |

**Write-side misattribution** (writes silently land in `default` / lose provenance) — mostly still open: #929, #1434, #1522, #1586, #1587, #2143, #2229, #2706 (fix in #2718). Same shape on the CLI: #1712, #2289 (`__all__` silently falls back to `default`).

[...]

Issue #1464, opened 2026-05-26
Source-scoped dream cycles should update freshness and preserve attribution


shandutta:
## Problem

Doctor freshness can stay unhealthy even when autopilot is active if full-cycle freshness is not attributed to the specific source that completed work. A downstream patch aligns dream freshness with source-scoped cycles and hardens source attribution through synthesize/minion paths.

Related downstream changes touch `src/commands/dream.ts`, `src/commands/autopilot.ts`, `src/core/cycle.ts`, `src/core/cycle/synthesize.ts`, and minion/subagent plumbing.

## Expected behavior

- `gbrain dream --source <id>` should clearly run and record a source-attributed full cycle.
- Autopilot should treat stale per-source freshness as actionable full-cycle work.
- Synthesis/minion failures should retain enough source context for doctor/admin surfaces.
- Doctor should not report a stale source merely because global service liveness is healthy.

## Current downstream patch

A downstream patch adds source-scoped dream freshness handling and propagates `source_id` through cycle/minion plumbing so failures and absorb logs remain attributable.

time-attack:
Closing — this reads as an upstreaming request for a downstream branch rather than a reproducible defect report. There's no version, no command transcript, no observed-vs-expected output, and no cited file:line, so there's nothing here that can be confirmed or refuted against current master across the five files it touches. If doctor genuinely reports a stale source while a source-scoped `gbrain dream --source <id>` completed successfully, please file that single symptom with the exact commands, the doctor output, and the version — a narrow verifiable report is something we can act on.

PR #3025, opened 2026-07-20
feat: local federated read fan-out + deterministic dream routing (closes #2561)

### LINKED ISSUE #2561: sources.config.federated ignored by unqualified local CLI search/query — federated source invisible unless --source named (docs say it joins cross-source search)
## Summary

A source registered with `--federated` (`config.federated: true`) never appears in unqualified local CLI `gbrain query` / `gbrain search` results, contradicting `docs/guides/multi-source-brains.md`:

> | `true` | Source participates in unqualified `gbrain search "X"` results. |

Only the resolved source (for me: `default`) is ever searched.

## Environment

- gbrain 0.42.56.0 (local build at 21b137f0), Postgres engine
- Brain: `default` (federated, ~24k pages) + several isolated sources + the new test source below

## Repro

```bash
mkdir -p ~/gbrain-sources/recipes && cd ~/gbrain-sources/recipes
git init && echo "# Recipe: knowledge-base cleanup (dedup + thin-content quarantine) ..." > recipe.md
git add -A && git commit -m seed

gbrain sources add recipes --path ~/gbrain-sources/recipes --federated
# → "federated: true — appears in cross-source default search"
gbrain sync --source recipes    # 1 file imported, 1 chunk, 1 page embedded

cd $HOME
gbrain search "thin-content quarantine" --source recipes --limit 3
# → [1.1224] recipe ...  ✓ found, top score

gbrain search "thin-content quarantine" --limit 5
# → only default-source pages, best [0.8791]; the 1.1224 recipes hit is ABSENT
gbrain query "recipe knowledge base cleanup duplicates quarantine" --limit 5 --expand false
# → same: recipes page absent despite outscoring every returned hit within its source
```

Page row verified clean: `frontmatter = {"date": "2026-07-03"}` (no quarantine key), `sources.config->>'federated' = 'true'`, synced + embedded.

## Where I think it lives (code read, not bisected)

- `src/core/source-resolver.ts` resolves to exactly ONE `source_id` (flag → env → dotfile → cwd → brain default → `'default'`). From `$HOME` that's `default`.
- `OperationContext.sourceId` is required and scalar; `sourceScopeOpts()` emits `{sourceId}` for local CLI, and reads filter `WHERE source_id = $X`.
- The only code that produces a multi-source `sourceIds[]` scope is the OAuth `federated_read` grant path (#876/#2200). I could not find any path where local CLI unqualified search expands the resolved source with `sources.config.federated = true` rows.

So `config.federated` appears to affect doctor/sync reporting but not local CLI search scope — i.e. the documented "joins unqualified search" behavior seems unimplemented or regressed by the v0.34 source-scoping work (D4/D9).

Related but distinct: #1913 (config corruption flips federated→isolated — not my case, flag verified true), #2200 (OAuth federated grant scope, not local CLI).

## Expected

Unqualified local `gbrain search` / `gbrain query` = resolved source ∪ all `config.federated=true` sources, per the guide.

## Workaround

Import the content into the `default` source directly (`gbrain import <dir>`) — surfaces fine (top-1, 0.9047).

pedrojahara:
## Summary

- Local callers (CLI, `gbrain call`, stdio MCP) now honor `sources.config.federated` for default reads: when the active source is federated and no explicit source is named, `search`/`query`/`list_pages`/`get_page` and the other read ops span the union of federated, non-archived sources — the "Unified knowledge recall" scenario that `docs/guides/multi-source-brains.md` and the `sources` comment in `src/schema.sql` promise, but which no local read path implemented (only OAuth clients with a `federated_read` grant ever received the `sourceIds[]` path).
- New additive `OperationContext.sourceIds` (local federated READ scope): sits strictly below the authenticated grant in `sourceScopeOpts` and is honored only for unauthenticated callers (`!ctx.auth`), so a grant can neither be widened by the field nor widened through it; empty arrays never widen; write paths never read it. The set is resolved by a new `resolveLocalFederatedReadScope` helper in `sources-load.ts` (stays scalar for isolated/archived/missing actives, single-federated brains, and on lookup errors).
- `dream.home_source` (new optional config) pins the corpus-global dream phases (synthesize, patterns) to one source's cycle. Without it, the per-source cycle fan-out routes each run's output into whichever source's cycle first hits the expired cooldown — a lottery whose winner can be an ISOLATED source, hiding reflections from default recall (observed in production two nights in a row). Fail-closed on config-read errors, validated at `config set` time (source must exist and not be archived; warns when not federated), surfaced by a new `dream_home_source` doctor check; explicit `--input/--date/--from/--to` runs bypass with the same predicate as the cooldown.
- Patterns output now follows the cycle's source — the child subagent payload carries `source_id` and collection stamps the cycle source, completing what #2939 did for synthesize (patterns still hard-coded `'default'`). The synthesize/patterns call sites in `runCycle` also wrap `yieldDuringPhase` in `buildYieldDuringPhase(lock, ...)` (the same T3 closure the extract_atoms/synthesize_concepts sites use) so the cycle lock survives the up-to-35-minute subagent waits — closing the non-atomic cooldown race that let two cycles run the same phase in one night.
- Scope-policy adjustments for the new field: `think --save` persists to the active source (`persistSynthesis` accepts `{sourceId}`); `thinkSourceScopeOpts` keeps the active scalar alongside the local array so trajectory resolution stays anchored; trusted local link reads keep the scalar cross-source view; code traversal collapses a local multi-element scope to the active scalar instead of throwing; `chronicle_backfill` enumeration stays pinned to the scalar its enqueue tag uses.
- Docs: the multi-source guide documents the fan-out mechanics and the `dream.home_source` pin, and fixes the cross-source search example (`--source a,b` silently degrades to one literal, unregistered id); `KEY_FILES.md` entries updated to current behavior; `scripts/e2e-test-map.ts` registers the new e2e suite.

## Relationship to existing work

- Closes #2561; addresses the federated→visible direction of #2928.
- Supersedes the approach in #2014 (stdio-only, manual `GBRAIN_SOURCES` env list): this PR derives the set automatically from the `federated` flag across all three local transports; #2014's env var could still compose later as a manual override.
- Builds on #2939 (merged): reuses its `cycleSourceId` seam and extends the same source threading to patterns.
- Cross-refs #1464 (announced dream freshness/attribution patch; the file overlap here is narrow — `cycle.ts`, `synthesize.ts` — happy to rebase/sequence) and RFC #2884 (required scope at the engine boundary; the ctx field here is compatible glue until that lands).
- A natural follow-up (not in this PR): have the autopilot fan-out skip dispatching the corpus-global phases to non-home cycles entirely — the in-phase gate is correctness; the dispatch filter would be efficiency.

## Test Coverage

- 32 unit tests: the scope precedence ladder (auth grant > local array > scalar > none, empty-never-widens), resolver edges (isolated/archived/missing actives, single-federated brains, error paths, JSONB string-shaped `sources.config`), policy pins (link-read local scalar view, think dual-emit, code-intel collapse, dispatch passthrough), and the home-source gate (fail-closed on config errors, operator-target bypass proven by differential skip reasons).
- 15 e2e cases (PGLite in-memory, no DATABASE_URL, hermetic — `GBRAIN_HOME` points at an empty tempdir and provider keys are cleared so search stays keyword-only and deterministic): federated union spans peers and hides isolated + archived sources; an isolated active source keeps its scalar scope; explicit `--source` pins; writes under an expanded ctx land only in the active source; `auth.allowedSources` beats the local array; code traversal collapses for CLI and stdio ctx shapes; `chronicle_backfill` enumerates only the active source; the `runCycle`→patterns `sourceId` threading is pinned differentially; `__all__` still spans the whole brain for trusted local callers.
- `bun run verify` green (31 checks); full `bun run test:e2e` run against real Postgres with failures triaged — the remaining failure set is byte-identical on the base commit (pre-existing).

https://claude.ai/code/session_017DN2B4wiWaK1EvayiHbz7y


pedrojahara:
Pushed a 6th commit (`27bff8a7`): **patterns cooldown**. First production night with `dream.home_source` surfaced that the patterns phase has no per-item idempotency, so it re-ran an LLM child on every home-source cycle (~hourly under autopilot) producing near-duplicate pattern pages. New `dream.patterns.cooldown_hours` (default 12h, 0 disables) mirrors the synthesize cooldown via a shared `checkDreamCooldown` helper; stamp-on-success/partial, no stamp on fail-with-zero-writes (retry posture unchanged). 3 new unit tests; verify green.

time-attack:
Thank you for the detailed writeup — the gap between what the multi-source guide promises and what local reads actually do is described clearly. I am closing this: it is a large feature PR (1142 lines) that adds a new OperationContext field, a new config key, a new doctor check, and new fan-out behavior, and it modifies sourceScopeOpts, which is the boundary where a single missed read path becomes a cross-source data leak. A change of that size against that invariant is not something I can review to the confidence bar it needs, and it bundles at least three separable changes. Not a judgment on the code quality.

Issue #2646, opened 2026-07-07
extract_facts guard counts expired legacy rows — forget_fact can't unblock the phase, apply-migrations no-ops


herove:
## Summary

The `extract_facts` dream phase is permanently blocked by its legacy-row guard when pre-v0.32 fact rows exist, even after `apply-migrations` reports up-to-date — and `forget_fact` cannot clear it because the guard's SQL does not exclude expired rows.

## Environment

- gbrain 0.42.56.0, postgres engine (self-hosted), macOS (Apple Silicon)
- v0.32.2 migration: **applied** (per `gbrain apply-migrations --list`)

## Repro / observed behavior

1. Brain had 17 fact rows with `row_num IS NULL AND entity_slug IS NOT NULL` (written 2026-06-22 ~ 06-29 by a pre-fence-era writer, i.e. *after* v0.32.2 was applied — so the migration backfill never saw them).
2. Every `gbrain dream --phase extract_facts` run exits with:
   `extract_facts skipped: 17 legacy v0.31 facts pending fence backfill. Run gbrain apply-migrations --yes to complete v0_32_2 ...`
3. `gbrain apply-migrations --yes` → `All migrations up to date.` (no-op; guard count unchanged)
4. `forget_fact` on one of the rows → `{"expired": true, "path": "legacy_db"}` — but the guard count **stays at 17**, because the guard query is:

```sql
SELECT COUNT(*) FROM facts WHERE row_num IS NULL AND entity_slug IS NOT NULL
```

(`src/core/cycle/extract-facts.ts` ~L120) — no `expired_at IS NULL` filter. So the officially sanctioned removal path (`forget_fact`) can never unblock the phase.

## Impact

`extract_facts` (fence → DB reconcile) is silently dead for any brain in this state; downstream consolidate/takes see stale facts. We only noticed via a manual dream-log audit.

## Suggested fixes (either)

1. Guard excludes soft-removed rows: `AND expired_at IS NULL` (lets `forget_fact` drain the backlog), and/or
2. `apply-migrations` re-runs the v0_32_2 backfill when matching rows exist even if the migration is marked applied (rows can postdate the migration if an old writer is still around).

## Workaround we used

Backed up the 17 rows to JSON, deleted them directly in SQL — phase immediately recovered (`507 fact(s) reconciled across 4455 page(s)`).

time-attack:
[...]

PR #3039, opened 2026-07-21
fix(facts): transport fence-authored supersessions into DB columns (#3014)

### LINKED ISSUE #3014: recall --supersessions is structurally empty — fence-authored supersessions never reach superseded_by/expired_at
## Summary

`gbrain recall --supersessions` is **structurally empty** on any store whose supersessions come from `## Facts` fences (the normal write path). A fence row struck through with `context: superseded by #N` is understood by the fence parser, but the supersession is never transported into the `facts.superseded_by` / `facts.expired_at` columns that `listSupersessions` filters on. The strike survives only as prose in `facts.context`.

Observed on **v0.42.62.0** (Postgres engine), but the mechanism is engine-agnostic (`pglite-engine.ts` has the same `listSupersessions` shape).

## Evidence

On a real store (388 facts, 100% embedded), across every row:

```
$ psql gbrain -tc "select count(*) filter (where superseded_by is not null) as sb,
                          count(*) filter (where expired_at is not null)   as ea,
                          count(*) as tot from facts;"
  0 |  0 | 388
```

Concrete case — a deal page whose terminal transition minted a closure event and struck the prior commitment:

```
# facts-fence on the page (correct): row 1 struck, context "superseded by #3"
# in the DB:
$ psql gbrain -tc "select id, superseded_by, expired_at, context from facts where id = <commitment-row>;"
  <id> |  (null) | (null) | superseded by #3
```

The three facts exist, are active, and are recall-visible by entity slug — only `--supersessions` cannot see them, because its filter is unsatisfiable:

```sql
-- postgres-engine.ts:4307-4308 (listSupersessions)
AND expired_at   IS NOT NULL
AND superseded_by IS NOT NULL
```

With both columns NULL store-wide, the query returns nothing on any fence-authored store.

## Root cause

The fence parser understands supersession — `facts-fence.ts:28` maps `~~claim~~` + `context: superseded by #N` to `active=false, supersededBy=N` (parsed field at `:99`, `:150`). But the mapper that turns parsed fence rows into insert rows drops it:

- `extractFactsFromFenceText` (`extract-from-fence.ts:173`) reads `supersededBy` only to decide `valid_until`, and its own header comment defers the rest to a later phase: *"supersededBy rows preserve their existing validUntil if set; otherwise leave valid_until = null (the consolidator phase fills...)"* (`extract-from-fence.ts:25-26`). It emits no `superseded_by`, no `expired_at`, no `active`.
- The insert shape cannot carry them anyway: `interface NewFact` (`engine.ts:508`) has `valid_until` but no `superseded_by` / `expired_at` / `active` member.
- The only writers of `facts.superseded_by` are `insertFact` with `ctx.supersedeId` (the LLM-classifier path) and the ontology-dimension path — neither runs in the `extract-facts` fence-sync route that ingests these pages.

So the "consolidator fills it later" contract is never honored for fence-authored supersessions, and `--supersessions` is permanently empty.

### Secondary inconsistency (same query)

The ontology-dimension write path sets `superseded_by` + `valid_until` but **not** `expired_at`, so even rows written that way would fail `listSupersessions`'s `expired_at IS NOT NULL` conjunct. Worth reconciling in the same change: either `listSupersessions` should not require both columns, or every supersession writer should set both.

## Ask

Transport the fence parser's `supersededBy` / `active` through `extractFactsFromFenceText` and `NewFact` into the insert path, so a struck fence row lands with `superseded_by` (and an `expired_at`/`valid_until` consistent with what `listSupersessions` filters on). And reconcile the `expired_at IS NOT NULL AND superseded_by IS NOT NULL` requirement against the ontology path that sets only one of the two.

Happy to test a fix against the live store above and confirm `recall --supersessions` surfaces the closure.

andreineacsu:
## Summary

Fixes #3014: `gbrain recall --supersessions` is structurally empty on any fence-authored store. The `## Facts` fence parser understands `~~claim~~` + `superseded by #N`, but the data was dropped between the parser and the DB, and the query filtered on columns a fence-authored row could never satisfy. This PR transports fence supersessions all the way into `facts.superseded_by` / `facts.expired_at`, reconciles the query, and adds unit + engine + phase + real-Postgres e2e coverage.

## Root cause (the four drop points from #3014)

1. **Parser — OK.** `parseFactsFence` correctly maps `~~claim~~` + `context: superseded by #N` to `active=false, supersededBy=N` (`src/core/facts-fence.ts`).
2. **Mapper drops it.** `extractFactsFromFenceText` (`src/core/facts/extract-from-fence.ts`) read `supersededBy` only to (not) derive `valid_until`; it emitted no `superseded_by`, no `expired_at`, and deferred "the rest" to a consolidator phase that never honors it.
3. **Insert shape can't carry it.** `interface NewFact` (`src/core/engine.ts`) had `valid_until` but no `superseded_by`/`expired_at` members, and `insertFacts` never wrote them.
4. **Query requires both columns.** `listSupersessions` filtered `expired_at IS NOT NULL AND superseded_by IS NOT NULL`; no fence-authored row set either, and the ontology-dimension writer sets `superseded_by` + `valid_until` (not `expired_at`), so even those never appeared.

## The fix

- **Transport.** The mapper stamps `expired_at` on every inactive row (superseded / forgotten / inactive-unrecognized) — there is no DB derivation from `valid_until`, so it is set explicitly — and carries the page-local `superseded by #N` reference as a new `superseded_by_row` field. `NewFact` gains `expired_at`. Both engines' `insertFacts` insert `expired_at` and resolve `superseded_by_row` → `facts.superseded_by` in a **second pass inside the same transaction**, keyed on `(source_id, source_markdown_slug, row_num)` so a reference resolves against both just-inserted and pre-existing target rows. `superseded_by` is a real FK, so self / dangling / struck-target references resolve to NULL + a logged warning — never a guessed id.
- **Filter reconcile (chosen rule).** `listSupersessions` now filters on `superseded_by IS NOT NULL` alone, ordering / `since` via `COALESCE(expired_at, valid_until)`. The alternative — making every writer set both columns — was rejected: the ontology-dimension writer deliberately closes a superseded row via `valid_until` (not `expired_at`) so `getOntology(--asof)` time-travel still sees it, and `getOntology` filters `expired_at IS NULL` unconditionally. Forcing `expired_at` there fails the existing `test/chronicle-ontology.test.ts` "forward supersession + --asof valid-time travel" case. The relaxed query surfaces both fence-authored and ontology supersessions without touching any writer.
- **Idempotent heal, atomic.** The reconcile phase detects a struck row whose DB columns drifted from the fence (keyed off the *resolved* reference id, so a permanently-unresolvable reference does not churn every cycle) and re-heals it. The wipe runs as the first statement of `insertFacts`' own transaction (opt-in `deleteForPageFirst`), so a failing insert rolls the delete back — the page is never left emptied.
- **int4 guard.** The parser accepts any finite `#N`; an 11+ digit value would overflow the `row_num` (int4) comparison and abort the cycle. Both engines gate the resolution SELECT on `isInt4RowRef`, degrading an absurd `#N` to a dangling reference instead of an uncaught throw.
- **Layering.** The pure resolver + its types + the int4 guard live in `engine.ts`, so the engines resolve without importing upward from the fence-boundary module.

## Tests

- Unit / engine / phase (in-memory PGLite): mapper emit + `resolveSupersededByRow` self/dangling/struck + `isInt4RowRef` (`test/extract-from-fence.test.ts`, 30); engine-level resolution + atomic `deleteForPageFirst` rollback + cli-exclusion (`test/insert-facts-batch.test.ts`, 22); reconcile transport + no-churn idempotency + heal + int4 (`test/extract-facts-phase.test.ts`, 27); option-B ontology visibility + --asof (`test/chronicle-ontology.test.ts`, 11); fence-write warning seam (`test/fence-write.test.ts`, 15).
- E2E (real Postgres, `DATABASE_URL`-gated): `test/e2e/facts-fence-reconcile-postgres.test.ts` — struck-row transport, idempotent heal, dangling/int4 no-churn, `deleteForPageFirst` rollback, ontology visibility, and `{since}` COALESCE fallback. **8/8 on real Postgres.**
- `bunx tsc --noEmit` clean.

## Live-store evidence

Verified on a real fence-authored store (a synthetic staged-terminal fixture page, `deals/zz-fixture-epic57-staged-terminal`; no other store content shown).

- Before: `superseded_by | expired_at` counts = `0 | 0` (of ~388–487 total facts) — structurally empty, matching #3014's readout.
- After a targeted re-extract of the fixture page: `superseded_by | expired_at` = `1 | 1`. The fixture's struck row 1 (`superseded by #3`) healed to `superseded_by =` (row 3's id) + `expired_at` set.
- `gbrain recall --supersessions` now surfaces it (structure): `- id=<row1> expired=2026-07-21T00:00:00.000Z superseded_by=<row3>` for the synthetic fixture commitment.

## Notes for maintainers (deferred, not in this PR)

- No index on `facts.superseded_by`; the relaxed `listSupersessions` predicate is a filtered scan. Fine at current scale, worth an index at large-store scale.
- The second-pass resolution is per-row sequential (one SELECT + one UPDATE per struck row). Struck rows are rare per page, but it is batchable (single `WHERE row_num = ANY(...)` lookup + a `CASE` update) if a hot path emerges.
- Ontology-dimension supersessions now appear in `listSupersessions` while remaining active in `getOntology` (they close via `valid_until`, not `expired_at`). This is intentional per the `--asof` time-travel contract; flagging it since it slightly widens what `recall --supersessions` returns.

### DIFF
--- a/src/core/cycle/extract-facts.ts
+++ b/src/core/cycle/extract-facts.ts
@@ -54,6 +54,11 @@ interface ExistingPageFact {
   fact: string;
   source: string | null;
   row_num: number | string | null;
+  // v0.42 (#3014) — supersession columns, read so a struck row whose
+  // fence says "superseded" but whose DB columns are still NULL counts as
+  // drifted and re-heals through the wipe+reinsert fallback.
+  superseded_by: number | string | null;
+  expired_at: Date | string | null;
 }
 
 function factContentKey(fact: string, source: string | null | undefined): string {
@@ -85,7 +90,7 @@ async function listExistingFactsForPage(
   sourceId: string,
 ): Promise<ExistingPageFact[]> {
   return engine.executeRaw<ExistingPageFact>(
-    `SELECT fact, source, row_num
+    `SELECT fact, source, row_num, superseded_by, expired_at
        FROM facts
       WHERE source_id = $1
         AND source_markdown_slug = $2
@@ -313,18 +318,35 @@ export async function runExtractFacts(
       const desired = desiredByKey.get(factContentKey(f.fact, f.source));
       return desired !== undefined && Number(f.row_num) !== desired.row_num;
     });
+    // v0.42 (#3014) — a struck row whose fence says "superseded" (or
+    // otherwise inactive) but whose DB columns are still NULL has an
+    // identical content key + row_num, so the checks above miss it. Treat
+    // a mismatch between the fence-desired supersession/expiry state and
+    // the DB columns as drift so the wipe+reinsert fallback re-heals the
+    // row (transports superseded_by + expired_at that a pre-fix cycle
+    // dropped).
+    const hasSupersessionDrift = existing.some(f => {
+      const desired = desiredByKey.get(factContentKey(f.fact, f.source));
+      if (desired === undefined) return false;
+      const desiredSuperseded = desired.superseded_by_row !== undefined;
+      const dbSuperseded = f.superseded_by != null;
+      const desiredExpired = desired.expired_at != null;
+      const dbExpired = f.expired_at != null;
+      return desiredSuperseded !== dbSuperseded || desiredExpired !== dbExpired;
+    });
 
     if (
       existing.length === extracted.length &&
       !hasStaleExisting &&
       !hasDuplicateExisting &&
-      !hasRowNumDrift
+      !hasRowNumDrift &&
+      !hasSupersessionDrift
     ) {
       continue;
     }
 
     let toInsert = extracted.filter(f => !existingKeys.has(factContentKey(f.fact, f.source)));
-    if (hasStaleExisting || hasDuplicateExisting || hasRowNumDrift) {
+    if (hasStaleExisting || hasDuplicateExisting || hasRowNumDrift || hasSupersessionDrift) {
       // Fall back to the legacy page-level reconcile when old DB rows must
       // be removed. Same delete scoping as above: legacy
       // NULL-source_markdown_slug rows and `cli:`-origin conversation
@@ -368,6 +390,10 @@ export async function runExtractFacts(
 
     const inserted = await engine.insertFacts(toInsert, { source_id: sourceId }); // gbrain-allow-direct-insert: extract_facts cycle phase reconciles fence → DB
     result.factsInserted += inserted.inserted;
+    // v0.42 (#3014) — surface unresolvable `superseded by #N` references
+    // (self / dangling / chain) as cycle warnings; the row still inserts
+    // with superseded_by NULL + expired_at set (never a guessed FK).
+    for (const w of inserted.warnings) result.warnings.push(`${slug}: ${w}`);
   }
 
   // v0.42 Wave B3: receipt + rollup. extract_facts is deterministic
--- a/src/core/cycle/extract-facts.ts
+++ b/src/core/cycle/extract-facts.ts
@@ -39,8 +39,10 @@ import { upsertExtractRollup } from '../extract/rollup-writer.ts';
 import { parseFactsFence } from '../facts-fence.ts';
 import {
   extractFactsFromFenceText,
+  resolveSupersededByRow,
   FENCE_SOURCE_DEFAULT,
   type FenceExtractedFact,
+  type SupersedeTarget,
 } from '../facts/extract-from-fence.ts';
 import {
   runPhantomRedirectPass,
@@ -51,6 +53,10 @@ import { embed, isAvailable } from '../ai/gateway.ts';
 import { isAborted } from '../abort-check.ts';
 
 interface ExistingPageFact {
+  // v0.42 (#3014) — the row's own fact id. Read so the supersession-drift
+  // check can compare the DB's stored `superseded_by` (a fact id) against
+  // the fence reference re-resolved to the target row's current id.
+  id: number | string;
   fact: string;
   source: string | null;
   row_num: number | string | null;
@@ -90,7 +96,7 @@ async function listExistingFactsForPage(
   sourceId: string,
 ): Promise<ExistingPageFact[]> {
   return engine.executeRaw<ExistingPageFact>(
-    `SELECT fact, source, row_num, superseded_by, expired_at
+    `SELECT id, fact, source, row_num, superseded_by, expired_at
        FROM facts
       WHERE source_id = $1
         AND source_markdown_slug = $2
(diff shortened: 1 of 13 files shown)


time-attack:
Merging. I confirmed every drop point in #3014 is still live on master — the mapper discards supersededBy, NewFact has no expired_at, and both engines' listSupersessions require a column pair no fence-authored row can ever set — so supersession data users write in the fence is silently lost. Resolving superseded_by_row in a second pass inside the same transaction, keyed on (source_id, source_markdown_slug, row_num), with self/dangling/struck refs falling back to NULL plus a warning rather than a guessed id, is the right call given superseded_by is a real FK. The size is high but it is mostly test coverage and unavoidable dual-engine parity; thanks for the real-Postgres e2e.

andreineacsu:
Closing the loop on this PR for the record, with a state check run on 2026-08-24.

History, for future readers: on 2026-07-28 this PR was approved in detail by @time-attack ("Merging. I confirmed every drop point in #3014 is still live on master ...", https://github.com/garrytan/gbrain/pull/3039#issuecomment-5099302040, 02:38:21Z), and the PR closed one second later (02:38:22Z) with mergedAt null and no merge commit, so the approved code never reached master at that time. Code-level checks then confirmed the gap: no commit on origin/master referenced #3014, extract-from-fence.ts had no superseded_by mapping, and the superseded_by surface in src/core/engine.ts stayed at its pre-fix count. A re-land request posted here on 2026-07-29, citing those checks and offering to rebase this branch onto current master, is no longer visible on the thread.

Current state, verified against origin/master today: the fix has landed. PR #4475 (v0.46.28.0, merged 2026-08-22, "Fixes #3014") ships the design that was approved here. Commit 67e7e8a9 introduces superseded_by_row into src/core/facts/extract-from-fence.ts (mapped from supersededBy at line 248) and deleteForPageFirst across the engine surface. expired_at is now stamped on inactive rows, supersession references resolve in a second pass keyed on (source_id, source_markdown_slug, row_num), and self or dangling references fall back to NULL plus a warning via resolveSupersededByRow, exactly the semantics reviewed on this PR. The one conflict I had flagged earlier, #2646's preserveExpiredLegacy against this PR's deferred deleteForPageFirst, is resolved in the same change: deleteForPageFirst accepts a preserveExpiredLegacy option and the extract path passes both together. Issue #3014 is closed as completed.

Given that, I am withdrawing the earlier rebase offer, there is nothing left to re-land, and this comment is only so the next person reading a CLOSED PR with mergedAt null understands how the story ended. If it is easy on your side, a link from #4475 back to this PR would help future archaeology, since the merged implementation matches the design that was reviewed and approved here. Thanks for landing the fix.

PR #3049, opened 2026-07-21
fix(providers): route multimodal-only embedding models through embedMultimodal in providers test

lost9999:
## Problem

`gbrain providers test --model voyage:voyage-multimodal-3` probes via `embedOne()`, which POSTs `/embeddings`. Voyage rejects multimodal-only models on that endpoint with HTTP 400, so the probe reports a false failure for a model that works fine at ingest time — the gateway routes it through `/multimodalembeddings` (`gateway.ts embedMultimodal`). The voyage recipe already documents this asymmetry (`multimodal_models: ['voyage-multimodal-3']` — "embedMultimodal() would let voyage:voyage-3-large through ... HTTP 400"), but `providers test` never learned it.

## Fix

In the embedding probe path, detect multimodal-only models via the recipe declaration (`supports_multimodal` + `multimodal_models.includes(modelId)`) and probe `embedMultimodal()` with a text input — the same routing production uses. All other models keep the existing `embedOne()` path.

## Validation

- Before: `gbrain providers test --model voyage:voyage-multimodal-3` → ✗ 400 Bad Request
- After: `✓ 1214ms, 1024 dims (multimodal endpoint)` (live Voyage call)
- `bun test test/providers.test.ts test/ai/recipes-existing-regression.test.ts` → 27 pass / 0 fail

### DIFF
--- a/src/commands/providers.ts
+++ b/src/commands/providers.ts
@@ -6,7 +6,7 @@
  */
 
 import { listRecipes, getRecipe } from '../core/ai/recipes/index.ts';
-import { configureGateway, embedOne, isAvailable as gwIsAvailable, chat as gwChat } from '../core/ai/gateway.ts';
+import { configureGateway, embedOne, embedMultimodal, getEmbeddingModel, isAvailable as gwIsAvailable, chat as gwChat } from '../core/ai/gateway.ts';
 import { probeOllama, probeLMStudio } from '../core/ai/probes.ts';
 import { loadConfig } from '../core/config.ts';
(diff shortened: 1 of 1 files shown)


time-attack:
Merging. Confirmed against master: the embedding probe always goes through `embedOne()` while the voyage recipe already declares `voyage-multimodal-3` as multimodal-only, so `providers test` was reporting a false failure for a model that works fine at ingest time. Routing the probe through the same endpoint production uses is the right, minimal fix. One nit for a follow-up: the `else` branch wasn't re-indented, so the diff reads a little oddly.

PR #3053, opened 2026-07-21
fix(put_page): run auto-timeline for remote callers

KyawZinLatt:
## Summary

`put_page` skips **both** auto-link and auto-timeline for remote/untrusted MCP callers via one gate. The gate is correct for auto-link — its bare-slug regex matches slugs anywhere in page text, so a remote writer could plant `see meetings/board-q1` and forge a cross-page edge that the backlink boost then surfaces. But auto-timeline is bundled into the same gate unnecessarily, so a remote-first / company-brain deployment (the documented primary mode) gets **zero timeline coverage on `put_page`, permanently**.

This decouples the two: auto-link stays gated for remote callers; auto-timeline runs for them.

## Why auto-timeline is safe on remote

Timeline entries are keyed to the page's **own** slug (`batch.map` uses the closure `slug`), so a remote/untrusted writer can only add entries to the page it is already authoring — it cannot forge entries on other pages or influence cross-page ranking. The multi-tenant threat model that motivates the auto-link gate does not apply.

## Changes

- **Decouple** auto-timeline from the remote gate; auto-link stays gated. Local/trusted behavior unchanged.
- **Scope each timeline row to `ctx.sourceId`** (mirrors the `runAutoLink` call). Without this the batch defaults to `source_id='default'`, so a write under a non-default source would attach timeline rows to a same-slug page in the default source (cross-source forgery / silent drop). This makes the new path correct for multi-source brains.
- **Bound the newly remote-reachable path, untrusted callers only**: cap the batch at 500 entries (bulk-insert amplification) and drop out-of-range years (align with `add_timeline_entry`'s 1900–2199). Trusted/local writers keep the prior unbounded behavior so a legitimate large local import is not silently truncated.
- Preserve the paired `auto_links`/`auto_timeline` skip contract when a remote write never parses (oversized content, no `parsedPage`).

## Files Changed

- `src/core/operations.ts` — Modified (put_page auto-timeline block)
- `test/e2e/graph-quality.test.ts` — Modified (3 new e2e tests)

## Testing

Three e2e regressions added to `test/e2e/graph-quality.test.ts`:
1. timeline runs on `remote: true` while `auto_links` stays `{ skipped: 'remote' }`;
2. remote entries scope to the caller source — a `tenantB` write of a `default`-source slug never forges onto the default page;
3. a remote write that never parses (oversized content) still reports `auto_timeline: { skipped: 'remote' }`.

These live under `test/e2e/` and require the real-Postgres harness (`DATABASE_URL`, per CONTRIBUTING), so they run in CI rather than the unit shard. Verified locally by static syntax check + type-compatible trace; not executed against a DB in my environment.

## Notes / follow-ups

- Known residual (pre-existing, not introduced here): request-count rate limiting bounds calls but not aggregate timeline-entry volume across calls. Out of scope for this fix.
- Commits can be squash-merged; the second/third commits fold in security-review hardening.

## Related Issues

None.

### DIFF
--- a/src/core/operations.ts
+++ b/src/core/operations.ts
@@ -967,6 +967,12 @@ const put_page: Operation = {
       && ctx.allowedSlugPrefixes.length > 0;
     if (ctx.remote !== false && !trustedWorkspace) {
       autoLinks = { skipped: 'remote' };
+      // Default for remote callers; overwritten below by the timeline
+      // extraction when result.parsedPage is present. Kept here so a remote
+      // write that never parses (e.g. oversized content, no parsedPage) still
+      // reports auto_timeline: { skipped: 'remote' } instead of dropping the
+      // key, preserving the original paired-skip response contract.
+      autoTimeline = { skipped: 'remote' };
     }
     // Auto-timeline is decoupled from the remote gate: unlike auto-link's
     // cross-page bare-slug matching, timeline entries are keyed to the page's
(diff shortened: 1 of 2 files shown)


time-attack:
Thank you for the detailed threat-model analysis — it's genuinely well argued. I'm still closing: the remote gate on `put_page` is deliberate and this repo's standing invariant is that trust is fail-closed, so opening a new remote-reachable bulk-insert path is a surface expansion rather than a bug fix. The fact that the change needs its own source scoping, a 500-row cap, and year-range validation to be safe is itself a signal of new risk being introduced. Remote-first deployments can still populate timelines through `gbrain extract`.

PR #3060, opened 2026-07-21
fix(extract): link-aware Source — Summary delimiter in timeline bullets — stop shattering entries on dashes inside markdown links

### LINKED ISSUE #3059: sync re-extraction shatters timeline bullets whose text contains dashes inside markdown links — fragments accumulate in timeline_entries on every sync
## Describe the bug

`extractTimelineFromContent` (src/commands/extract.ts, Format 1) parses timeline bullets with:

```ts
```

Because the source group is non-greedy and the delimiter class includes a plain hyphen, the Source — Summary split lands on the **first dash after the pipe** — including hyphens inside markdown link **targets** and em-dashes inside link **labels**. Auto-generated backlink lines are the worst hit, since almost every slug is hyphenated:

```ts
const re = /^-\s+\*\*(\d{4}-\d{2}-\d{2})\*\*\s*\|\s*(.+?)\s*[—–-]\s*(.+)$/gm;
re.exec('- **2026-07-19** | Referenced in [Pedro](../people/pedro-franceschi.md)');
// source  = 'Referenced in [Pedro](../people/pedro'
// summary = 'franceschi.md)'
```

## Why it compounds

`sync` runs `extractTimelineForSlugs` over every page whose markdown changed. The corrupted (source, summary) tuple differs from any prior row, so `ON CONFLICT (page_id, date, summary, source) DO NOTHING` does not dedupe it — each sync of a touched page re-inserts fragment rows alongside whatever was there. On an active brain (agents appending timeline entries hourly), pages accumulate garbage timeline rows indefinitely. We measured 40 corrupted rows across 10 pages within days on a fresh brain; a manual DB cleanup was re-shattered as soon as the pages' markdown changed again.

Observed on 0.42.59.0 and 0.42.62.0; the pattern is unchanged at current master.

Note: `core/link-extraction.ts` `parseTimelineEntries` handles these lines fine — the defect is only in the `commands/extract.ts` copy that populates the `source` column.

## To reproduce

1. Put a timeline bullet whose text contains a markdown link with a hyphenated target on any page (any `Referenced in` backlink qualifies).
2. `gbrain extract --mode timeline` (or touch the page and let sync run).
3. `timeline_entries` gains a row with the link split across `source`/`summary`; repeat the sync after any markdown change and fragments accumulate.

## Expected behavior

Dashes inside `[label](target)` spans are content, not delimiters. Only a whitespace-surrounded dash outside link spans should split Source from Summary; a bullet with no such delimiter (the backlink shape) should be captured whole rather than shattered or dropped.

PR with a link-aware delimiter scan + regression tests incoming.

wright-io:
Fixes #3059.

## What

`extractTimelineFromContent`'s Format-1 parser split `Source — Summary` at the first `[—–-]` after the pipe, which lands inside markdown links: hyphenated link targets (`../people/pedro-franceschi.md`) and em-dash link labels (`[Deals — Q1 Review](...)`) shattered the entry into two fragments. Because `(page_id, date, summary, source)` uniqueness treats each fragment shape as a new tuple, every sync of a touched page re-inserted fragments — corruption that accumulates hourly on active brains (details + repro in #3059).

## How

- Match the bullet up to the pipe, then find the delimiter with `findDelimiterOutsideLinks`: only a whitespace-surrounded dash at bracket-depth 0 (outside every `[label](target)` span) splits Source from Summary.
- Delimiterless bullets — e.g. every auto-generated backlink line `- **date** | Referenced in [X](y.md)` — are captured whole (`source: 'markdown'`, full text as summary) instead of being shattered (before 0.42.x behavior) or silently dropped. This matches `core/link-extraction.ts`'s tolerance of the same lines and Format 2's use of the `'markdown'` source sentinel.

## Tests

4 new cases in `test/extract.test.ts` covering: hyphenated link target, spaced em-dash inside a link label, spaced dash *after* a link (still splits correctly), and delimiterless line kept whole. `bun test test/extract.test.ts`: 32 pass / 0 fail. Full unit suite: 6684 pass; the one failing test (`hardenBrainRepo > installs hook…`) fails identically on a clean master checkout in this environment (pre-existing, unrelated).

### DIFF
--- a/src/commands/extract.ts
+++ b/src/commands/extract.ts
@@ -469,15 +469,53 @@ export async function extractLinksFromFile(
 
 // --- Timeline extraction ---
 
+/**
+ * Index of the first dash (—, –, -) that can serve as the Source — Summary
+ * delimiter: it must have whitespace on both sides and sit outside every
+ * markdown-link span. Hyphens inside link targets
+ * (`../people/pedro-franceschi.md`) and dashes inside link labels
+ * (`[Deals — Q1 Review](...)`) are content, not delimiters — splitting on
+ * them shatters one entry into two fragments whose halves re-insert on
+ * every sync (the (page_id, date, summary, source) uniqueness sees each
+ * fragment shape as a new row). Returns -1 when the line has no delimiter.
+ */
+function findDelimiterOutsideLinks(text: string): number {
+  let depth = 0;
+  for (let i = 0; i < text.length; i++) {
+    const c = text[i];
+    if (c === '[' || c === '(') depth++;
+    else if (c === ']' || c === ')') { if (depth > 0) depth--; }
+    else if (
+      depth === 0 &&
+      (c === '—' || c === '–' || c === '-') &&
+      i > 0 && /\s/.test(text[i - 1]) &&
+      i + 1 < text.length && /\s/.test(text[i + 1])
+    ) {
+      return i;
+    }
+  }
+  return -1;
+}
+
 /** Extract timeline entries from markdown content */
 export function extractTimelineFromContent(content: string, slug: string): ExtractedTimelineEntry[] {
   const entries: ExtractedTimelineEntry[] = [];
 
   // Format 1: Bullet — - **YYYY-MM-DD** | Source — Summary
-  const bulletPattern = /^-\s+\*\*(\d{4}-\d{2}-\d{2})\*\*\s*\|\s*(.+?)\s*[—–-]\s*(.+)$/gm;
+  // The delimiter search is link-aware (see findDelimiterOutsideLinks); a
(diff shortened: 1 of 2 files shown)


time-attack:
Merging. Confirmed on master that the non-greedy source group plus a bare hyphen in the delimiter class splits inside link targets, and because the timeline uniqueness key includes summary+source those fragments re-insert on every sync — real accumulating corruption, exactly as #3059 describes. A useful bonus I verified while reviewing: the canonical enrich format documented at `skills/enrich/SKILL.md:241` carries no delimiter at all, so today it matches neither Format 1 nor the citation arm; your delimiterless branch recovers those entries too. Small, link-aware, and well covered by tests.

PR #3061, opened 2026-07-21
feat(search): retire pages from search via a frontmatter status marker

Walliiee:
## Problem

gbrain models supersession richly at the **facts / takes** layer (`auto-supersession`, `valid_until`, `forgotten`). But a raw markdown **page** has no author-facing way to say *"this is out of date — stop surfacing it."* The two existing page-level controls don't fit the common authoring case:

- `delete_page` (`deleted_at`) hides a page **entirely** and is an operator action.
- `quarantine` is reserved for **auto-detected junk** (the content-quality gate).

The everyday case is: *I wrote a newer page that supersedes an old one. Keep the old one reachable by slug for history, but don't let search return it as if it were current.* Today, moving the page to a `cold-storage/` folder does **nothing** — gbrain indexes and ranks it the same as any other page. So stale pages keep coming back as top hits.

## Approach — mirror the `quarantine` pattern

This adds a page **lifecycle** marker that is the exact sibling of `quarantine` (HIDES) / `content_flag` (WARNS):

- A page whose frontmatter `status` is a **hide value** — `superseded`, `deprecated`, `retired`, `obsolete` — is excluded from keyword/vector search via `lifecycleFilterFragment`, injected by `buildVisibilityClause` (the single filter all 6 search sites share).
- **Search-only.** A retired page is *not* deleted: it stays fetchable by slug (`get_page`), stays in `list`, and its links/backlinks resolve. Retirement means "not surfaced by search" — a one-line frontmatter edit instead of a file move.
- **No schema migration.** `status` is read straight from the existing `frontmatter` JSONB column (same property that made `quarantine` cheap). Works identically on Postgres and PGLite.
- **Backward compatible by construction.** Only the four explicit hide values retire a page. Every other `status` an author already uses — `active`, `draft`, `done`, `in-progress`, `review` — is inert and stays fully searchable. Pages with no `status` are unaffected.

`lifecycle.ts` is the single source of truth for the marker key + hide vocabulary + SQL fragment, exactly like `quarantine.ts`, so the search filter and the marker can never drift.

## Usage

```yaml
---
type: project
title: Adaptig platform integration (v1)
tags: [adaptig, platform]
status: superseded   # <- gbrain search now skips this page; get_page still returns it
---
```

## Not in this PR (deliberate, documented extension points)

- **WARN tier** (`status: stale` → stays searchable but rides a marker into `SearchResult`, mirroring `content_flag`). No SQL filter for it, by design.
- **`archived` as a hide value** — left out to avoid confusion with the source-level `archived` flag; one-line add in `LIFECYCLE_HIDE_VALUES` if wanted.
- **Rank-penalty (demote instead of hide)** — the fragment hides; a softer post-fusion penalty could be layered later.

## Tests

`test/lifecycle.test.ts` — 10 unit tests, no DB: predicate (`isRetired`, case/space-insensitive, non-string safety), `retiredStatus`, `filterOutRetired`, fragment shape + alias parameterization, and `buildVisibilityClause` wiring. All pass. The JSONB positional-param guard is not in scope (this reads a raw SQL fragment, adds no parameterized insert).

### DIFF
--- a/src/core/lifecycle.ts
+++ b/src/core/lifecycle.ts
@@ -1,0 +1,120 @@
+/**
+ * Page lifecycle marker: an author-set frontmatter signal that RETIRES a page
+ * from search without deleting it.
+ *
+ * Why this exists (the gap it closes):
+ *   gbrain already models supersession richly at the *facts / takes* layer
+ *   (auto-supersession, `valid_until`, `forgotten`). But a raw markdown *page*
+ *   had no author-facing way to say "this is out of date — stop surfacing it."
+ *   `deleted_at` (soft-delete) hides a page entirely and is an operator action;
+ *   `quarantine` is reserved for auto-detected junk. Neither fits the common
+ *   authoring case: "I wrote a newer page; keep the old one reachable by slug
+ *   for history, but don't let search return it as if it were current."
+ *
+ * The design mirrors `quarantine` (HIDES) vs `content_flag` (WARNS):
+ *   - A page whose frontmatter `status` is one of `LIFECYCLE_HIDE_VALUES`
+ *     (`superseded`, `deprecated`, `retired`, `obsolete`) is EXCLUDED from
+ *     keyword/vector search via `LIFECYCLE_FILTER_FRAGMENT`, injected by
+ *     `buildVisibilityClause` (`src/core/search/sql-ranking.ts`).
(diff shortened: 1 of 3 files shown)


time-attack:
Closing after maintainer review — decision: not accepting this. Nothing breaks; search behaves as it does today. The idea could return later as an opt-in feature with an escape hatch and docs.

PR #3067, opened 2026-07-21
v0.42.63.0 fix(write-through): case-insensitive-filesystem collision guard (#2831)

### LINKED ISSUE #2831: write-through.ts: no case-insensitive filesystem collision guard before atomic write (data-loss on Windows/macOS)
### Summary

`src/core/write-through.ts`'s atomic write path (`mkdirSync` → temp-write → `renameSync`) has no check for an existing file whose name differs from the target only in case. On a case-sensitive filesystem this is harmless. On Windows and default-configuration macOS, the filesystem collapses e.g. `docs/FOO.md` and `docs/foo.md` to the same inode — the write silently destroys whatever content was already at the differently-cased path. No error, no warning.

### Reproduction

On Windows (or macOS with the default case-insensitive APFS volume):

1. Have an existing file `docs/ops/EXISTING-DOC.md` **not** created via `put_page` — e.g. a human-authored file, or content checked out via `git`.
2. Call `put_page` with a slug that lowercases to the same path (`docs/ops/existing-doc`).
3. The write silently succeeds. The original file's content is gone.

We reproduced this live — it destroyed a session's own bug-report file mid-write while that file was being authored.

### Root cause

`write-through.ts`, the relevant lines (paraphrased from our fork, should match master):

```ts
if (sourceLocalPath) {
  if (!existsSync(sourceLocalPath) || !statSync(sourceLocalPath).isDirectory()) {
    return { written: false, skipped: 'repo_not_found' };
  }
  filePath = join(sourceLocalPath, `${slug}.md`);
  writeRoot = sourceLocalPath;
} else {
  ...
}
...
mkdirSync(dirname(filePath), { recursive: true });
const tmpPath = `${filePath}.tmp.${process.pid}.${randomBytes(4).toString('hex')}`;
writeFileSync(tmpPath, md, 'utf8');
renameSync(tmpPath, filePath);
```

There is no existence check against a differently-cased path anywhere before the atomic write. The file's own docstring comment block shows the author was clearly thinking about atomicity and cross-source leak guards (there's already a `source_repo_belongs_to_other_source` guard a few lines above, which refuses a write rather than risking cross-contamination) — but case-insensitive-filesystem collisions weren't covered by that same defensive pattern.

### Suggested fix

Before the `mkdirSync`/`writeFileSync` block, `readdirSync` the target directory and check for a sibling entry that matches the target basename case-insensitively but not exactly (an exact-case match is a normal in-place update, not a collision — should fall through unaffected). On a differently-cased match, refuse the write with a new `skipped` reason (e.g. `case_insensitive_collision`), matching the existing `WriteThroughResult.skipped` union and the `source_repo_belongs_to_other_source` guard's own refuse-rather-than-clobber shape.

Honest caveat on the fix shape: a `readdirSync`-then-write check has a narrow race window (another writer could create the colliding file between the check and the rename) — same risk class as the existing `source_repo_belongs_to_other_source` guard a few lines above it (a DB check followed by a later write, with nothing enforcing atomicity between them). Not a claimed lock, just closes "always silently clobbers" down to a narrow race.

### Verification

Fixed and verified in our fork (`exGeni/gbrain-local-patches`, branch `local-patches`, commit `02f1702a`) — verified via an isolated unit test of the detection logic (differently-cased collision → detected and refused; exact-case match → not flagged, normal update proceeds; brand-new file → not flagged). Happy to open a PR if useful.

Masashi-Ono0611:
## Summary

On Windows and default-configuration macOS (case-insensitive APFS), a path that differs from an existing one only in case collapses to the SAME inode. `writePageThrough`'s atomic write (`mkdirSync` → temp write → `renameSync`) had no check for this, so a human-authored (or otherwise non-gbrain) file sitting at a differently-cased path than the computed slug path was silently destroyed by the rename — no error, no warning. The reporter said this destroyed a real bug-report file mid-write.

## Fix

Before the write, walk every path segment from the write root down against what's actually on disk. An exact-case match at a segment means "keep descending into the same real entry" (normal in-place update); a case-insensitive-but-not-exact match at ANY segment — not just the final basename — means the write would land inside or over a differently-cased entry somewhere in the chain, so refuse with a new `case_insensitive_collision` skipped reason rather than clobber, matching the existing `source_repo_belongs_to_other_source` guard's refuse-rather-than-clobber shape in the same file.

Checking only the final basename (an earlier draft of this fix, caught by Codex review) misses the case where the mismatch is in a PARENT directory — e.g. an existing `Docs/ops/note.md` colliding with a target of `docs/ops/note.md`: the basename `note.md` matches exactly, but a case-insensitive filesystem resolves `docs/ops` to the SAME directory as `Docs/ops` before `readdirSync` ever sees the basename listing, so a basename-only check reports no collision and the rename proceeds to clobber. Walking every segment closes this gap; a regression test locks it in.

`formatSaveOutcome` (`gbrain brainstorm/lsd --save`) is updated to recognize the new skipped reason with an honest message — it previously fell through to the generic "see error above; `gbrain sync` will reconcile" branch, which is wrong here: no error was raised, and `gbrain sync` would hit the exact same refusal rather than reconcile anything.

**Honest caveat** (documented inline): this is a `readdirSync`-then-write check, not a lock — a narrow race window exists if another writer creates the colliding entry between the check and the rename. Same risk class as the existing leak guard a few lines above (a DB check followed by a later write with nothing enforcing atomicity between them); this closes "always silently clobbers" down to a narrow race, not a claimed guarantee.

## Testing

- New suite `test/write-through-case-collision.test.ts`: differently-cased basename collision (refused), exact-case in-place update (proceeds), brand-new file (proceeds), unrelated sibling (no false positive), and the parent-directory-case regression case above (refused).
- `formatSaveOutcome` test added covering the new skip reason's message.
- `bunx tsc --noEmit`: 0 errors.
- Reviewed with `codex review` (gpt-5.6-sol, high effort) across three rounds: round 1 caught the basename-only gap (P1) and the `formatSaveOutcome` fallthrough (P2), both fixed; round 3 came back clean.

Fixes #2831

---

A note on provenance: the original reporter (@exGeni) said in the issue that they'd already fixed this in their own fork and offered to open a PR themselves. Some time passed with no PR from them, so this is an independent implementation from scratch — I don't have access to their patch or fork, so this isn't a copy of their work, just my own read of the same bug. No interest in taking credit away from them here: if they'd rather submit their own version, or the maintainer prefers theirs, I'm happy to have this PR closed in favor of it. The goal is just getting the data-loss bug fixed.


Masashi-Ono0611:
Closing as absorbed: the underlying issue (#2831) was fixed on `master` by a different, already-merged implementation.

- Commit `c571bf82` — "fix(write-through): guard case-insensitive filesystem collisions before atomic write (#2831) (#3119)", merged 2026-07-23 via PR #3119 (https://github.com/garrytan/gbrain/pull/3119), explicitly "Fixes #2831".
- Current `src/core/write-through.ts` on master (around `writePageThrough`, after the `getPage`/`serializePageToMarkdown` step) now checks `dirname(filePath)` via `readdirSync` and refuses with `skipped: 'case_insensitive_collision'` when the target path exists on disk but no exactly-cased directory entry does — the same guard behavior this PR introduced, wired the same way (returned in `WriteThroughResult.skipped`, consumed by `formatSaveOutcome` in `src/commands/brainstorm.ts`).

One honest difference for the record: master's guard walks a single level (`dirname(filePath)` + `basename`), so it catches basename-level case collisions but not a case mismatch in an intermediate parent directory segment (e.g. existing `Docs/ops/x.md` vs target `docs/ops/x.md` where the basename matches exactly) — this PR's version walked every path segment from the write root down and had an explicit regression test for that parent-directory case. Not reopening this PR over it since the primary bug (#2831) is now closed by the merged approach and that's the maintainer's accepted shape; flagging only in case it's worth a narrow follow-up later.

Given the fix landed independently upstream, this PR is now redundant.

PR #3074, opened 2026-07-21
fix(chunker): estimated-token hard cap — URL-dense/CJK-fallback chunks overflow strict embedding limits

### LINKED ISSUE #2826: Chunker emits chunks past strict embedding-server token limits on URL-dense CJK docs (whitespace-word undercount) and un-subdividable AST nodes

time-attack:
Fixes #2826
Takeover of #2847 (rebased onto current master; original work by @paul-0320, credited via Co-authored-by — not closing that PR, just referencing).

## Summary

URL/phone/email-dense CJK markdown and large un-subdividable AST nodes produce 3–4K-char chunks that overflow strict per-request embedding-server token limits (measured: local llama-server with `-ub 2048` crashes past ~2,050 tokens per chunk). Two root causes:

1. **Whitespace-word undercount** — `countCJKAwareWords`' fallback (CJK density < 0.30) counts a 150-char URL as one word, and per-line URLs are exactly what pushes CJK docs below the density threshold. The existing `maxChars: 6000` cap assumes ≥1 char/token; URL soup tokenizes at ~1.6 chars/token.
2. **Unbounded AST-path chunks** — a node `splitLargeNode` can't subdivide ships whole at any size.

## Changes

- `src/core/cjk.ts` — additive `estimateEmbeddingTokens()`: tokenizer-free per-char-class estimate (CJK 1.0/char, other non-whitespace 0.75/char, whitespace 0.1/char), calibrated to deliberately overestimate so any cap based on it is safe against real tokenizers. `countCJKAwareWords` semantics untouched.
- `src/core/chunkers/recursive.ts` (`MARKDOWN_CHUNKER_VERSION` 3→4): `countWords` floored at `ceil(nonWhitespaceChars / 6)` (Latin prose unchanged — pinned by tests); `capByEstimatedTokens()` final pass with new `ChunkOptions.maxTokens` (default 1500), cut points preferring newline then whitespace within a 300-char lookback.
- `src/core/chunkers/code.ts` (`CHUNKER_VERSION` 4→5): `capCodeChunks()` applies the same cap to AST-path chunks.

Version bumps route existing pages through the standard reindex sweep.

## Tests

`test/chunkers/token-cap.test.ts` (new, 13 tests): reproduces both failure shapes (URL-dense CJK rollup; large JSON through prose and `fence.json` code paths) and pins the cap invariant, content preservation, cut placement, forward progress on whitespace-less input, and unchanged behavior for normal Latin/CJK prose. Existing chunker suites updated only for version-constant assertions.

## Verification

- `bun run typecheck` exit 0.
- `bun test test/chunkers/token-cap.test.ts test/chunkers/recursive.test.ts test/chunkers/code.test.ts test/chunker-version-gate.test.ts`: 73 pass / 0 fail.
- Pure text pipeline — no engine SQL, no persistence changes; engine parity not implicated.

### DIFF
--- a/src/core/chunkers/code.ts
+++ b/src/core/chunkers/code.ts
@@ -18,8 +18,9 @@
  * at runtime.
  */
 
-import { chunkText as recursiveChunk } from './recursive.ts';
+import { chunkText as recursiveChunk, capByEstimatedTokens, DEFAULT_MAX_EST_TOKENS } from './recursive.ts';
 import { buildQualifiedName } from './qualified-names.ts';
+import { estimateEmbeddingTokens } from '../cjk.ts';
 
 // Embed the tree-sitter runtime + per-language grammars as files.
 // `with { type: 'file' }` returns a path (string) at runtime. Bun bundles
@@ -111,7 +112,15 @@ import G_ZIG from '../../assets/wasm/grammars/tree-sitter-zig.wasm' with { type:
 // chunks get the new columns populated. Without this, the v28 backfill
 // gives every existing chunk a search_vector but subsequent Layer 5 AST
 // work would silently no-op.
-export const CHUNKER_VERSION = 4;
+//
+// v5: estimated-token hard cap on AST-path chunks (capCodeChunks). A node
+// splitLargeNode can't subdivide (giant single-statement function, huge
+// literal) previously shipped WHOLE regardless of size and could overflow
+// strict per-request embedding-token limits (local llama-server crashes
+// past ~2,050 tokens, measured). Mirrors the markdown
+// chunker's v4 cap; fallback-path chunks are already capped inside
+// recursiveChunk.
+export const CHUNKER_VERSION = 5;
 
 // Lazy-loaded tree-sitter module (v0.22.x API: Parser is default export)
 let Parser: typeof import('web-tree-sitter') | null = null;
@@ -708,7 +717,7 @@ export async function chunkCodeTextFull(
     if (chunks.length === 0) {
       return { chunks: fallbackChunks(source, filePath, language, opts), edges: rawEdges };
     }
-    return { chunks: mergeSmallSiblings(chunks, chunkTarget), edges: rawEdges };
+    return { chunks: capCodeChunks(mergeSmallSiblings(chunks, chunkTarget)), edges: rawEdges };
   } catch {
     return { chunks: fallbackChunks(source, filePath, language, opts), edges: [] };
   } finally {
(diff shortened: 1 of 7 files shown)


time-attack:
Thanks for the careful writeup and the takeover work — #2826 is a real failure mode and the reproduction is well documented. I'm closing this shape rather than the problem: `estimateEmbeddingTokens` weights non-CJK characters at 0.75 tokens/char, roughly 3x the real Latin-prose rate, so a default cap of 1500 estimated tokens shrinks ordinary English chunks from the current 6,000-char ceiling to about 2,000, and the two chunker-version bumps force a full re-chunk and re-embed of every page in every brain. That is a global retrieval-behavior change with no eval receipts, taken on to accommodate one embedding server's `-ub 2048` setting. A provider-aware or config-driven token limit that defaults to today's behavior would be a much smaller blast radius.

PR #3079, opened 2026-07-21
fix(extract): recognize plain-bullet timeline format (- YYYY-MM-DD — Summary)

time-attack:
Takeover of #1896 (rebased onto master; original by @ElliotDrel, credited as co-author).

## What

Adds a Format 4 arm to `extractTimelineFromContent` for the plain bullet `- YYYY-MM-DD — Summary` — the format gbrain's own enrich skill writes. Without it, every brain-authored timeline entry is invisible to extraction and timeline coverage stays at 0%.

## Why a takeover instead of merging #1896 as-is

Since #1896 was opened, #2524 landed an inline-citation arm as **Format 3** at the exact insertion point the PR targeted, so:

- The diff no longer applies and the "Format 3" name collides — the plain-bullet arm now lands as **Format 4**.
- Force-applying it would double-count: a plain bullet carrying a citation (`- 2026-05-31 — text [Source: User, 2026-05-31]`, the shape quality.md mandates) matches BOTH the plain-bullet pattern AND master's citation loop, whose skip regex only excludes bold Format-1 bullets. The citation loop's skip is extended to plain-bullet lines so that shape files exactly one entry (regression test included).

## Tests

- `bun test test/extract.test.ts` — 33 pass, 0 fail (5 new tests: plain-bullet extraction, single-entry for bullet+citation, no double-count of Format 1, no spurious entry from a date inside a summary, en dash/hyphen separators)
- `bun run typecheck` — clean

### DIFF
--- a/src/commands/extract.ts
+++ b/src/commands/extract.ts
@@ -509,8 +509,12 @@ export function extractTimelineFromContent(content: string, slug: string): Extra
   // DB-level uniqueness cannot collapse.
   const citationPattern = /\[Source:\s*([^\]]+?),\s*(\d{4}-\d{2}-\d{2})\s*\]/g;
   const bulletLinePattern = /^-\s+\*\*\d{4}-\d{2}-\d{2}\*\*\s*\|/;
+  // Lines captured by Format 4 (plain bullet) are skipped for the same
+  // reason: quality.md mandates a trailing [Source: ...] on those bullets,
+  // and re-extracting the citation would double-count the event.
+  const plainBulletLinePattern = /^-\s+\d{4}-\d{2}-\d{2}\s*[—–-]/;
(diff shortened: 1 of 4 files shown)


time-attack:
Thanks for doing the rebase and untangling the Format 3 collision — that analysis was useful. I'm closing this one because the premise doesn't hold on current master: `skills/enrich/SKILL.md` and `skills/_brain-filing-rules.md` both document the bold-pipe bullet (`- **YYYY-MM-DD** | Event ... [Source: ...]`), not the plain `- YYYY-MM-DD — Summary` shape, so adding a Format 4 arm is new format surface rather than a fix, and it widens matching to any bulleted date line. The genuine coverage gap for the documented enrich format is being addressed in #3060.

PR #3081, opened 2026-07-21
fix(jobs): prune --status filter + --older-than 0d as explicit no-age-floor

time-attack:
Takeover of #2282 (author: @brettdavies, credited via Co-authored-by).

Master's `queue.prune` already accepts a `status[]` param, but the CLI exposed neither a status filter nor any way below a 1-day floor. This salvages the CLI plumbing from #2282 and repairs the flaws found in adversarial verification:

- **`--status completed,failed,dead,cancelled`** — comma-separated terminal subset passed through to `queue.prune`; any non-terminal value fails fast. Parsing extracted into exported `parsePruneStatuses` (unit-tested, same pattern as `parseNiceFlag`).
- **`--older-than 0d` documented as what it does** — the original PR titled/described it as "same-day prune (preserves the audit trail)" while the code implemented *no age filter at all*. This lands it honestly: help text and the success line both say it removes the age floor and deletes ALL matching terminal jobs ("regardless of age"), so operators can't mistake it for a start-of-today cutoff.
- **Real tests** — the original body cited cases in `test/jobs.test.ts`, a file that does not exist. This PR adds queue-level status-filter + zero-age-floor tests to `test/minions.test.ts` and parser tests in `test/jobs-prune-flags.test.ts`.

Gates: `bun run typecheck` clean; `bun test test/minions.test.ts test/jobs-prune-flags.test.ts` → 187 pass / 0 fail.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -2387,7 +2387,7 @@ JOBS (Minions)
   jobs get <id>                       Job details + history
   jobs cancel <id>                    Cancel job
   jobs retry <id>                     Re-queue failed/dead job
-  jobs prune [--older-than 30d]       Clean old jobs
+  jobs prune [--older-than 30d] [--status s,..]  Clean old terminal jobs (0d = no age floor)
   jobs stats                          Job health dashboard
   jobs work [--queue Q]               Start worker daemon (Postgres only)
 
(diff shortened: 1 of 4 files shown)


time-attack:
Thanks for the salvage work and for catching that the original PR cited a test file that doesn't exist — that kind of verification is appreciated. Closing anyway: `jobs prune` on master parses `--older-than` correctly and rejects bad input, so nothing here is broken; `--status` and the `0d` no-age-floor semantics are new operator surface, and we're not taking new CLI flags right now. No reflection on the quality of the patch.

PR #3082, opened 2026-07-21
fix(cli): honor search --json output (#2042)

### LINKED ISSUE #2042: search --json emits empty stdout when engine.disconnect() hangs and CLI force-exits

time-attack:
Takeover of #2531 (author: @javieraldape), rebased onto master. Fixes #2042.

`gbrain search <q> --json` silently printed human text: search/query declare no `json` param, `parseOpArgs` dropped the undeclared trailing flag, and `formatResult`'s search/query case only emitted human text or `--explain` output.

- `parseOpArgs` treats `--json` as a CLI-local boolean formatter flag (deliberately NOT added to the operation contract; MCP `validateParams` ignores the extra key, so the thin-client routed path is unaffected).
- `formatResult(op, result, params)` emits the raw result array as parseable JSON for search/query when `params.json === true`.
- Rebase note: the local-path call site keeps master's `bigintToStringReplacer` argument; PR semantics carry over unchanged.

Tests: `test/cli-args.test.ts` (flag parsing), `test/cli-format-search-json.test.ts` (formatter, incl. empty array), `test/cli-search-dispatch.test.ts` (subprocess regression against the real cli.ts entrypoint). `bun run typecheck` exit 0; all three files pass.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -464,7 +464,7 @@ async function main() {
     // routed path. Date → ISO string; bigint → string (postgres.js shape);
     // Buffer → object. Microsecond-cost; eliminates a whole drift bug class.
     const result = JSON.parse(JSON.stringify(rawResult, bigintToStringReplacer));
-    const output = formatResult(op.name, result);
+    const output = formatResult(op.name, result, params);
     if (output) process.stdout.write(output);
   } catch (e: unknown) {
     // v0.42.20.0 (codex D4): on error, set exitCode + return so the `finally`
@@ -547,7 +547,7 @@ async function runThinClientRouted(
       signal: sigintController.signal,
     });
     const result = unpackToolResult(raw);
-    const output = formatResult(op.name, result);
+    const output = formatResult(op.name, result, params);
     if (output) process.stdout.write(output);
   } catch (e: unknown) {
     if (e instanceof RemoteMcpError) {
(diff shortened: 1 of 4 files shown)


time-attack:
Thanks for rebasing this and for the honest note about the CLI-local flag. Closing for two reasons: #2042 was already closed as completed — its real root cause was the JSON payload being lost when `engine.disconnect()` hung and the CLI force-exited, fixed by the #2084 exit contract in `src/core/cli-force-exit.ts` — and what remains here is new `--json` output surface for search/query that isn't in the documented help. The deliberate decision to add the flag outside `src/core/operations.ts` also cuts against the contract-first rule that CLI and MCP are both generated from one contract, so if this surface is ever wanted it should land in the contract.

Issue #1972, opened 2026-06-08
Compendium (0.42.34.0): stale dead-PID cycle locks never reaped, disconnect() 10s hang on PgBouncer txn-mode CLI, and incomplete #1737 cooperative-abort


garrytan-agents:
## Compendium: 3 production-observed defects on 0.42.34.0 (single deployment, 386K-page brain)

Filing as a consolidated report from a live, heavily-loaded deployment (386,190 pages, 2.6M typed edges, ~14K page-writes/day, Supabase split-pool + PgBouncer transaction-mode on `:6543`, multi-source federation: `default` + `straylight-brain`). All three are reproducible here and are observability/liveness gaps rather than data-corruption bugs. Version `0.42.34.0` @ `099d9a8f`.

Cross-links to adjacent open issues noted per-bug; none of these fully covers the specific failure below.

---

### BUG 1 — Stale `gbrain_cycle_locks` rows held by **dead PIDs** are never reaped except on-contention; a crashed sync/cycle silently blocks that source until something else happens to contend

**Observed:** After a worker recycle, `gbrain_cycle_locks` retained rows whose `holder_pid` was a **dead process**:

```
gbrain-sync:straylight-brain   pid=737217 (DEAD)  held=00:36:12
gbrain-cycle:straylight-brain  pid=746387 (DEAD)  held=00:06:59
```

`737217` exited >36 min ago — **past the 5-min `LOCK_TTL_MS`** (cycle.ts:17) and past the 30-min sync TTL — yet the row persists. `kill -0 737217` → no such process.

**Root cause:** lock reclaim is **lazy / on-contention only**. `tryAcquireDbLock` (cycle.ts:77) steals a row only when a *new* acquirer shows up AND (`age > TTL` OR `pid not alive on this host`) — see the comment at cycle.ts:90 (`...older than LOCK_TTL_MS OR the PID is no longer alive on this host`). There is **no background sweep**. So when a sync for source `S` crashes (worker killed, OOM, recycle), its `gbrain-sync:S` row strands. The **next** sync of `S` will eventually steal it — but until then `gbrain status` shows the source as "syncing," and any liveness/health probe keyed on lock presence reports a phantom in-flight cycle. With per-source locks, a low-traffic source can stay falsely "locked" for hours.

**Why TTL doesn't save you:** TTL expiry is also only evaluated *at acquire time*. A dead-pid lock with an expired TTL is still a live row in the table until contention; nothing deletes it proactively.

[...]

PR #3090, opened 2026-07-21
fix(cycle): stamp last_full_cycle_at for the resolved source; stale freshness forces autopilot fanout

### LINKED ISSUE #2060: autopilot targeted mode can leave cycle_freshness stale without dispatching a source cycle
## Summary

On a Postgres/federated-source brain maintained by `gbrain autopilot`, `cycle_freshness` can go stale while autopilot continues to dispatch and complete `sync` jobs. In this state, doctor reports the source as stale, but the autopilot decision loop stays in targeted remediation mode and never queues a per-source `autopilot-cycle`.

This is distinct from #1992, where a cycle runs but does not stamp `last_full_cycle_at`. Here, no `autopilot-cycle` job was dispatched for the stale source until a manual/native recovery was run.

## Observed

- gbrain version: 0.42.37.0
- One federated source
- `sync_freshness`: OK
- `cycle_freshness`: FAIL, source last cycled ~33h ago
- Job history over the stale window: repeated completed `sync` jobs, no `autopilot-cycle` jobs
- Brain score around 70 with a small targeted remediation plan, so the scheduler kept choosing targeted jobs

A one-shot native recovery (`gbrain dream --source <id>`) cleared `cycle_freshness`, which confirms the source and stamp path can work. The concern is that the autopilot scheduler does not appear to use stale source cycle freshness as an input when deciding whether to dispatch a full cycle.

## Expected

If a source is stale according to the same `last_full_cycle_at` signal used by doctor's `cycle_freshness` check, autopilot should dispatch a scoped `autopilot-cycle` for that source, subject to the existing full-cycle throttle/fanout limits.

## Possible fix direction

In `src/commands/autopilot.ts`, include stale per-source cycle freshness in the full-cycle dispatch decision. In other words, targeted sync/extract remediation should not indefinitely suppress a full source cycle when `last_full_cycle_at` is already beyond the freshness threshold.

A downstream test shape that caught this:

- score: 71
- remediable plan length: 2
- estimated plan seconds: 60
- minutes since last full dispatch: 120
- stale cycle source count: 1
- expected: dispatch full cycle

## Local mitigation

For now we are avoiding a local vendored code patch and using native manual recovery when doctor reports this failure: `gbrain dream --source <id>`, then rerun doctor.

time-attack:
Two fixes to the per-source cycle freshness loop (backlog unit c53):

- Takeover of #1993 (credit @100menotu001, co-authored on the commit): `runCycle`'s freshness stamp now keys off `cycleSourceId` (`opts.sourceId ?? resolveSourceForDir(brainDir)`) — the same id the cycle locked and scoped its phases to — instead of raw `opts.sourceId`. The autopilot's inline cycle passes `brainDir` with no explicit `sourceId`, so the stamp never fired and `cycle_freshness` stayed stale even while the autopilot cycled every interval. Rebased onto master's current stamp block: keeps the `!aborted` guard (#1972) and the `last_source_cycle_at` write (#2194). The two test cases from the original PR are carried over.
- Fixes #2060: autopilot's dispatch decision now consults per-source cycle staleness. New pure helper `countStaleSources` in `autopilot-fanout.ts` (reuses `isSourceStale` over `listAllSources({ localPathOnly: true })`); a stale source forces the fanout path and blocks the healthy-sleep gate, so a brain sitting at score 70–94 with a small targeted plan (≤3 steps, <300s) can no longer starve per-source cycle dispatch indefinitely. Fail-open to 0 on read errors; `dispatchPerSource`'s existing throttles (skipped_fresh / fanoutMax / failure cooldown) bound the resulting work.

Tests:
- `test/cycle-last-full-cycle-at.test.ts`: brainDir-resolves-source case (fails without the cycle.ts fix — verified) + brainDir-matches-no-source negative case.
- `test/autopilot-fanout.test.ts`: `countStaleSources` unit coverage.
- `test/autopilot-fanout-wiring.test.ts`: pins the `staleCycleSources` terms in `shouldFullCycle` / `shouldSleep`.

Gate: `bun run typecheck` exit 0; the three touched test files pass (42 pass / 0 fail).

### DIFF
--- a/src/commands/autopilot-fanout.ts
+++ b/src/commands/autopilot-fanout.ts
@@ -187,6 +187,18 @@ export function isSourceStale(src: SourceRow, now = Date.now(), floorMin = FULL_
 }
 
 /**
+ * #2060: count sources past the per-source cycle freshness floor. Consumed
+ * by autopilot's dispatch decision — a stale source forces the fanout path
+ * even when the doctor plan is small (score 70–94, plan ≤ 3, est < 300s),
+ * so targeted mode can't leave cycle_freshness stale indefinitely.
+ * dispatchPerSource's own throttles (skipped_fresh / fanoutMax / failure
+ * cooldown) bound the resulting work.
+ */
+export function countStaleSources(sources: SourceRow[], now = Date.now(), floorMin = FULL_CYCLE_FLOOR_MIN): number {
+  return sources.filter((s) => isSourceStale(s, now, floorMin)).length;
+}
+
+/**
  * Most recent SUCCESSFUL cycle for a source. Prefers `last_source_cycle_at`
  * (per-source phases, written by the split cycle) and falls back to the legacy
  * `last_full_cycle_at`, so this works before AND after the cycle split.
--- a/src/commands/autopilot.ts
+++ b/src/commands/autopilot.ts
@@ -901,13 +901,27 @@ export async function runAutopilot(engine: BrainEngine, args: string[]) {
         const FULL_CYCLE_FLOOR_MIN = 60;
         const minutesSinceLastFull = (Date.now() - lastFullCycleAt) / 60000;
 
+        // #2060: stale per-source cycle freshness is a dispatch input. Without
+        // it, a brain sitting at score 70–94 with a small targeted plan (≤3
+        // steps, <300s) stays in targeted mode indefinitely and no per-source
+        // cycle is ever dispatched — cycle_freshness never advances. A stale
+        // source forces the fanout path; dispatchPerSource's throttles
+        // (skipped_fresh / fanoutMax / failure cooldown) bound the work.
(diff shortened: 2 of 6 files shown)


time-attack:
Thanks for the careful writeup — I confirmed the root cause is genuine and still present on master: the freshness stamp block in `src/core/cycle.ts` keys off raw `opts.sourceId` rather than the resolved `cycleSourceId`, so the autopilot's inline cycle never stamps. I'm closing this particular branch because CI is red where master is fully green: serial-tests fails `runCycle — sourceId resolution (regression #475) > multiple rows with same local_path`, which is the exact resolver path this diff changes, so as submitted it trades one source-resolution bug for another. The bundled autopilot dispatch change (`countStaleSources` forcing fanout) is also a scheduler behavior change rather than a stamp fix, and folding it in makes the regression harder to isolate.

PR #3092, opened 2026-07-21
fix(migrate): bootstrap --to pglite --path destination as a standalone GBRAIN_HOME

### LINKED ISSUE #1271: gbrain migrate --to pglite: destination not self-bootstrapped + --url arg persists secret to dest config.json
Two adjacent surprises in `gbrain migrate` that came out of a backup/restore drill against a Supabase-backed brain on `gbrain 0.31.3`. Related to but distinct from #724 (which covers source-config rewrite). Filed separately because these are different code paths and have different fixes.

### Repro

1. Start with a Supabase-backed brain at `GBRAIN_HOME=~/.gbrain-prod`.
2. Run: `GBRAIN_HOME=~/.gbrain-prod gbrain migrate --to pglite --path /tmp/restore-test`.
3. Migration completes successfully, 36 pages copied, verify passes.
4. Try to use the destination as a brain: `GBRAIN_HOME=/tmp/restore-test gbrain query "..."`.

### Expected

The destination is a usable brain.

### Actual

```
No brain configured. Run: gbrain init
```

`/tmp/restore-test/` contains the PG data files (PG_VERSION, base/, global/, etc.) plus an empty `.gbrain/` directory. But `.gbrain/config.json` is absent, so `loadConfig()` can't initialize the engine.

### Mitigation (manual workaround)

After migrate, write the destination config by hand:

```bash
echo '{"engine": "pglite", "database_path": "/tmp/restore-test"}' \
  > /tmp/restore-test/.gbrain/config.json
```

### Proposed fix

`gbrain migrate --to pglite --path P` should write `P/.gbrain/config.json` as part of the migration finalization, with the canonical PGLite shape (`{engine, database_path}`). Makes the destination immediately usable without operator intervention.

---

### Repro

1. Start with a PGLite brain at `GBRAIN_HOME=/tmp/source-brain` containing data.
2. Ensure `GBRAIN_DATABASE_URL` is UNSET in env (otherwise the source-engine override at `loadConfig:134-141` forces source to postgres and the migrate equality-guard trips). Pass the target URL via `--url` argument:

   ```bash
   GBRAIN_HOME=/tmp/source-brain gbrain migrate --to supabase \
     --url postgresql://postgres.PROJ:PASSWORD@aws-...:5432/postgres
   ```

3. Migration completes; check destination config:

   ```bash
   cat /tmp/source-brain/.gbrain/config.json
   ```

### Expected

If `--url` is used as an ephemeral argument (operator's intent: pass the URL this one time, don't persist it), the destination config should EITHER (a) omit `database_url` entirely (rely on env for future invocations), or (b) reference the env var (e.g., `"database_url": "${GBRAIN_DATABASE_URL}"`).

### Actual

The full URL — including password — is written verbatim into `/tmp/source-brain/.gbrain/config.json`:

```json
{
  "engine": "postgres",
  "database_url": "postgresql://postgres.PROJ:PASSWORD@aws-...:5432/postgres"
}
```

Mode 0600 limits local exposure, but the URL is now at rest on disk in a file that may sync to version control, backup systems, or shared development environments depending on the operator's setup. The intent behind `--url` (pass-through CLI argument) does not telegraph "this will be persisted to disk indefinitely."

### Mitigation (manual workaround)

After every `gbrain migrate --to supabase --url X`, immediately strip the `database_url` field from the destination config:

```bash
jq 'del(.database_url)' /tmp/source-brain/.gbrain/config.json > /tmp/cfg.json \
  && mv /tmp/cfg.json /tmp/source-brain/.gbrain/config.json
```

Then rely on `GBRAIN_DATABASE_URL` env at runtime.

### Proposed fix

One of:

(a) Don't persist `--url` to config.json. The argument was passed ephemerally; treat it as a one-shot connection target, not a config override.

(b) Add `--no-persist-url` flag (default off for backward compat) that triggers behavior (a). Operators opting into the secure-by-default path can set it.

(c) Document the persistence behavior loudly in `--url`'s help text, and warn at runtime when `--url` is used: "`--url` will be persisted to `<dest>/.gbrain/config.json`. To avoid this, set `GBRAIN_DATABASE_URL` in env instead and omit `--url`."

## Context

Found during a Phase 1 backup/restore DR drill on a personal Supabase-backed gbrain. Same drill that prompted #724 from a different angle. Mitigations documented in my workspace runbook; filing here because future operators deploying gbrain on managed Postgres for clients will hit these and not have the workaround context.

## Version

`gbrain 0.31.3` (bun source clone, head at `7267462`). macOS 26.x, Bun 1.3.x. Source brain via Supabase pooled connection (`aws-1-us-east-1.pooler.supabase.com:5432`), destination PGLite under `/tmp/`.

time-attack:
Fixes #1271

**Finding 1 (fixed):** `gbrain migrate --to pglite --path P` completed successfully but left the destination unusable — `GBRAIN_HOME=P gbrain query ...` failed with `No brain configured` because nothing wrote `P/.gbrain/config.json`. The migration finalization now bootstraps the explicit-`--path` destination with its own `config.json` (`{engine: 'pglite', database_path: <abs path>}`, mode 0600) plus a `*` `.gitignore` (same worktree-safety pattern as `saveConfig`). It never clobbers an existing destination config and is best-effort — a completed migration never fails over it. Default-path migrations (no `--path`) are unchanged; the current home's config already covers those.

**Finding 2 (documented, by design):** persisting `--url` to config.json is required for the migrated brain to work without env — `gbrain init` for postgres persists `database_url` identically, and `saveConfig` already writes 0600 + a home `.gitignore`. Rather than change behavior, the persistence is now stated out loud: a one-line stderr note at migrate time when `--url` is used, and a note in the command's usage header.

**Test:** `test/migrate-engine-dest-bootstrap.test.ts` — asserts the destination config shape, 0600 mode, .gitignore, that `loadConfigFileOnly()` under `GBRAIN_HOME=<dest>` now resolves (the exact failure mode from the issue), and the no-clobber guard. Fails without the fix.

Gate: `bun run typecheck` exit 0; new + adjacent migrate-engine tests pass. (`test/migrate-stdout-clean.test.ts` hook-timeout is pre-existing on master, unrelated.)

### DIFF
--- a/src/commands/migrate-engine.ts
+++ b/src/commands/migrate-engine.ts
@@ -3,7 +3,11 @@
  *
  * Usage:
  *   gbrain migrate --to supabase [--url <connection_string>]
+ *     (--url is persisted to config.json, mode 0600, so the migrated brain
+ *      works without env — #1271)
  *   gbrain migrate --to pglite [--path <db_path>]
+ *     (an explicit --path destination is bootstrapped with its own
+ *      <path>/.gbrain/config.json so GBRAIN_HOME=<path> just works — #1271)
  *   gbrain migrate --to <engine> --force  (overwrite non-empty target)
  */
 
@@ -11,9 +15,9 @@ import { createEngine } from '../core/engine-factory.ts';
 import { loadConfig, saveConfig, toEngineConfig, gbrainPath, effectiveEnvDatabaseUrl, type GBrainConfig } from '../core/config.ts';
 import type { BrainEngine } from '../core/engine.ts';
 import type { EngineConfig } from '../core/types.ts';
-import { writeFileSync, readFileSync, existsSync, unlinkSync } from 'fs';
+import { writeFileSync, readFileSync, existsSync, unlinkSync, mkdirSync, chmodSync } from 'fs';
 import { createHash } from 'crypto';
-import { resolve } from 'path';
+import { resolve, join } from 'path';
 import { createProgress } from '../core/progress.ts';
 import { getCliOptions, cliOptsToProgressOptions } from '../core/cli-options.ts';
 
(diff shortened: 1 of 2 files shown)


time-attack:
Merging this one. I confirmed against current master that `finalizeMigration` only ever calls `saveConfig` for the current home, so nothing writes `<dest>/.gbrain/config.json` — the backup/restore drill in #1271 still reproduces today, and a migration that reports success but leaves an unusable destination is a broken promise worth fixing. The scope is right: best-effort, never clobbers an existing destination config, never fails a completed migration, and default-path migrations are untouched. The red `test (4)` check is the known shard-4 15-minute timeout cancellation, not a real failure.

Issue #2569, opened 2026-07-03
dream/synthesize: dream_generated provenance is render-time only — generated pages are unqueryable in DB, can leak during runs, and put_page writes erase it


TheRealMrSystem:
## What problem does this solve?

`dream_generated: true` is the identity surface for dream/synthesis output. It is used by the self-consumption guard (`isDreamOutput`, `src/core/cycle/transcript-discovery.ts`) and by generated-output demotion conventions in search/source handling. But on current `master` (`646179047a8e4ad9c462d83ce9a67a50ba076de8`), the stamp is not durable for generated child pages:

1. **DB/file split.** `renderPageToMarkdown()` in `src/core/cycle/synthesize.ts` stamps generated child pages via `frontmatterOverrides` only (`dream_generated: true`, `dream_cycle_date`). `serializePageToMarkdown()` then merges `{...page.frontmatter, ...frontmatterOverrides}` at render time (`src/core/markdown.ts`). That gets the marker into the rendered markdown file, but it does not persist the marker into the `pages.frontmatter` JSONB row for those pages. DB-level consumers therefore cannot reliably enumerate generated child pages via `frontmatter->>'dream_generated'`.

2. **Stamp decay on later write-through.** `put_page` write-through in `src/core/operations.ts` re-renders from the saved DB row with only ingestion/source overrides. If the DB row never had `dream_generated`, a later `put_page` write can silently erase the marker from the file.

3. **Reverse-write failures/skips are hard to observe as provenance loss.** The summary page persists its own generated marker into DB, but generated children rely on reverse-write rendering. If a child page is listed in a dream-cycle summary but the reverse-write marker is missing or later erased, downstream systems have no durable membership predicate.

4. **Agent/MCP-driven synthesis has no generated-page identity.** Pages produced by synthesis-driving agents through `put_page` can look like ordinary curated writes (`ingested_via: mcp:put_page`) rather than generated output, even when operationally they are part of a synthesis run.

Measured on one production brain corpus that motivated this report, under `wiki/originals/ideas/`:

- 156 generated/synthesis idea pages inspected
- 68 stamped / **88 unstamped** (`56%` unstamped)
- 67 pages are listed from dream-cycle summaries; **8 of those are unstamped**
- 80 unstamped pages were not listed from the summaries and look like ordinary `put_page` writes

That means generated pages can be impossible to enumerate reliably, and can be treated by retrieval/agents with the same authority as human-authored/curated pages.

## Code references checked against current master

[...]

Issue #1586, opened 2026-05-28
dream synthesize writes synthesized pages to 'default' source, ignoring the resolved brain source


chengdalong26:
## Summary

`gbrain dream --phase synthesize --dir <brain-dir>` writes the synthesized
reflection/idea pages to the `default` source, even when `<brain-dir>` is a
registered source (its `.gbrain-source` dotfile names the source id). Result:
persistent `multi_source_drift` warnings in `doctor`, and the named source's
`last_sync` never advances ("stale" forever).

## Root cause

In `src/core/cycle/synthesize.ts`, `collectChildPutPageSlugs()` (~L870)
hardcodes `source_id: 'default'` for all subagent-written pages. The comment at
L849-854 acknowledges this is a known gap:

> Subagent put_page tool schema doesn't expose source_id (subagents are scoped
> to a single source); default to 'default' for the current dream-cycle product
> behavior. Threading the source_id through reverseWriteRefs guarantees getPage
> targets the correct (source, slug) row...

So the `brain_put_page` write path never honors the resolved source from the
6-tier `resolveSourceId` chain (`--source` / `GBRAIN_SOURCE` / `.gbrain-source`
/ path-match / brain-default / 'default').

## Impact

- `doctor` → `multi_source_drift`: pages live at `default` but their intended
  source is the named brain source.
- `doctor` → `sync_freshness`: the named source is flagged stale forever,
  because dream writes land in `default` and never advance the source's sync.
- Workaround forces operators to `git commit && gbrain sync --source <id>` to
  re-home the pages, which leaves duplicate copies (one in `default`, one in the
  named source). Search dedupes by slug so retrieval is unaffected, but
  `page_count` / `chunk_count` inflate.

## Repro

1. `gbrain sources add mybrain --path ~/mybrain` (so `~/mybrain/.gbrain-source`
   = `mybrain`)
2. `gbrain dream --phase synthesize --dir ~/mybrain`
3. `gbrain doctor` → `multi_source_drift` lists the new reflection/idea pages as
   intended=`mybrain` but present at `default`.

## Suggested fix

Thread the resolved sourceId into the subagent put_page write path and through
`collectChildPutPageSlugs` / `reverseWriteRefs`, so synthesized pages are
attributed to the correct source at write time instead of the hardcoded
`'default'`.

## Env

gbrain 0.35.8.0, postgres engine, multi-source setup (multiple federated
sources).

xaviroblessarries:
[...]

PR #3096, opened 2026-07-21
feat(dream): orchestrator-owned transcript metadata + transcriptSource discovery field

### LINKED ISSUE #2285: dream transcript metadata: discovery exposes no source label, date inference fights with content metadata, orchestrator does not own deterministic frontmatter

time-attack:
Takeover of #2286 (credit: @brettdavies, co-authored on the commit). Fixes #2285.

The original branch was DIRTY against master and its rewritten `collectChildPutPageSlugs` hardcoded `source_id: 'default'`, regressing #1586's cycle-source scoping (a source-isolation bug for non-default cycles). This rebuild lands the same feature on top of current master with the `cycleSourceId` threading intact.

## What ships

- **`DiscoveredTranscript.transcriptSource`** — derived from the `<corpus>/<source>/<date>/<id>.md` layout during discovery (`deriveTranscriptSource`); null for ad-hoc inputs that don't match. Available to every downstream consumer.
- **Content-based date inference** — `inferredDate` now prefers the `| First message | <ISO> |` row in the transcript's `## Metadata` table (stable across rsync/Dropbox/Syncthing re-syncs that re-stamp mtime), falling back to the filename-regex date. #2286 promised this cascade in its body but never shipped the parser; it's implemented here, and the date-range filter now matches UUID-named archive transcripts on their content date.
- **Orchestrator-owned deterministic frontmatter** — each written page gets `transcript_id` (content-version suffix stripped), `transcript_hash`, `transcript_source`, `chunk` (i/n for chunked transcripts), and `date` stamped authoritatively. Subagents retain authority over type/title/tags/body; their drift on the deterministic fields is overwritten.

## How it differs from #2286's implementation

- Keeps master's #1586 source scoping: refs are stamped with the cycle's resolved source, never a hardcoded `'default'`.
- No second `engine.putPage` round-trip: the deterministic fields ride the existing #2569 `stampDreamProvenance` jsonb-merge funnel (raw object bound via `executeRawJsonb` — no stringify-into-jsonb), which already runs before `reverseWriteRefs`; the reverse-render re-reads the row, so DB and disk stay lockstep through one write path.
- `date` (position 2 in the `effective-date.ts` precedence chain) is stamped as plain `YYYY-MM-DD`; a redundant frontmatter `effective_date` key the chain never reads was dropped.

## Tests

- `test/cycle/transcript-discovery-metadata.test.ts` (new): `deriveTranscriptSource` layouts, `inferContentDate`, content-over-filename precedence, filename fallback, date-filter match on content date, `readSingleTranscript` parity.
- `test/cycle-synthesize-slug-collection.test.ts` (extended): ref↔job pairing keeps the cycle source (#1586 guard), full metadata stamp lands as queryable jsonb with subagent keys preserved, minimal stamp omits inapplicable fields.
- `bun run typecheck` clean; touched suites (`cycle-synthesize*`, `cycle-dream-output-root`, `synthesize-gateway-adapter`, e2e `dream-synthesize-pglite` + `dream-synthesize-chunking`, `v0_29-tool-surfaces`, `regression-strict-source-id`) all green.

### DIFF
--- a/src/core/cycle/synthesize.ts
+++ b/src/core/cycle/synthesize.ts
@@ -1139,7 +1139,8 @@ async function collectChildPutPageSlugs(
        FROM subagent_tool_executions
       WHERE job_id = ANY($1::int[])
         AND tool_name = 'brain_put_page'
-        AND status = 'complete'`,
+        AND status = 'complete'
+      ORDER BY id`,
     [childIds],
   );
   const rewritten = new Map<string, number>();
@@ -1149,7 +1150,10 @@ async function collectChildPutPageSlugs(
     const finalSlug = meta && meta.chunkTotal > 1
       ? rewriteChunkedSlug(r.slug, meta.hash6, meta.idx)
       : r.slug;
-    if (!rewritten.has(finalSlug)) rewritten.set(finalSlug, r.job_id);
+    // Last writer wins, in execution-row order (ORDER BY id): if two children
+    // collide on a final slug, the pages row holds the LAST put_page write, so
+    // the stamp must attribute that child's transcript — not an arbitrary one.
+    rewritten.set(finalSlug, r.job_id);
   }
   return [...rewritten.entries()]
     .sort(([a], [b]) => a.localeCompare(b))
--- a/src/core/cycle/synthesize.ts
+++ b/src/core/cycle/synthesize.ts
@@ -428,8 +428,13 @@ export async function runPhaseSynthesize(
 
     const queue = new MinionQueue(engine);
     const childIds: number[] = [];
-    /** Map child job_id → chunk metadata for D6 orchestrator-side slug rewrite. */
-    const chunkInfo = new Map<number, { idx: number; hash6: string }>();
+    /**
+     * Map child job_id → transcript metadata. Drives D6 orchestrator-side
+     * slug rewrite for chunked transcripts AND the deterministic frontmatter
+     * stampDreamProvenance merges into each written page. Populated for
+     * every child (single-chunk children carry chunkTotal=1).
+     */
+    const childMeta = new Map<number, ChildMeta>();
     /** Skip reasons for the cycle report (D5 cap hits, D8 legacy-key skips). */
     const skipReports: Array<{ filePath: string; reason: string }> = [];
 
@@ -513,9 +518,14 @@ export async function runPhaseSynthesize(
           { allowProtectedSubmit: true },
         );
         childIds.push(child.id);
-        if (isChunked) {
-          chunkInfo.set(child.id, { idx: i, hash6 });
-        }
+        childMeta.set(child.id, {
+          idx: i,
+          hash6,
+          chunkTotal: chunks.length,
+          transcriptSource: t.transcriptSource,
+          transcriptId: stripContentVersionSuffix(t.basename),
+          inferredDate: t.inferredDate,
+        });
       }
     }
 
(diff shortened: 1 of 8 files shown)


time-attack:
Thanks for doing the hard work here — rebuilding on top of master's #1586 source scoping instead of inheriting the hardcoded `source_id: 'default'` was the right call, and routing through the existing `stampDreamProvenance` funnel rather than a second `putPage` shows real care. That said, this is ~370 lines of net-new feature surface on the dream/cycle path (a new discovery field, a metadata-table date parser, and orchestrator-owned frontmatter stamping), and we're holding a firm line against adding feature surface that isn't fixing a broken core promise. Closing on that policy basis rather than on the quality of the work.

PR #3098, opened 2026-07-21
perf(config): batch + cache engine.getConfig to kill ~85 round-trips per query

time-attack:
Takeover of #1694 (by @Omerbahari) — rebased onto current master; nothing closed, just referenced.

## What

A single search fires ~85 serial `getConfig()` reads (`loadConfigWithEngine` x2 plus the mode/cache/intent/rerank/graph-signals resolvers). On a remote pooler each read is a round-trip; serial they dominate query latency and can push the op handler past `cli.ts`'s 10s disconnect force-exit, truncating stdout.

- First read batch-loads the whole `config` table into a process-lifetime Map (single-flight under concurrent cold reads).
- `setConfig`/`unsetConfig` write through, so a long-lived process never serves stale config it wrote itself.
- 30s TTL bounds multi-writer staleness; `GBRAIN_CONFIG_CACHE_TTL_MS=0` restores per-key reads (incident escape hatch).

## Why a takeover, not a rebase-merge of #1694

#1694 predates #1603/#1891: master's `getConfig`/`setConfig`/`unsetConfig` are now wrapped in `connRetry()` (retry + reconnect posture that fixed pooler-drop empty-stdout bugs), and the original diff would have replaced those with bare `this.sql` reads. Here both the batch load and the TTL=0 per-key fallback stay inside `connRetry()`; write-through happens after the retried statement succeeds.

## Deliberate scope

- PGLite stays uncached: in-process, zero round-trips, and unit-test fixtures seed the `config` table via raw SQL and expect fresh reads.
- E2E helpers pin `GBRAIN_CONFIG_CACHE_TTL_MS=0` since those suites also rewrite config via raw SQL mid-test.

## Tests

- New `test/postgres-engine-config-cache.test.ts`: single batch load for repeat reads, known-absent keys, write-through on set/unset, single-flight under concurrency, TTL=0 off switch, TTL expiry reload, and connRetry reconnect posture on the batch load.
- Updated `test/postgres-engine-config-reconnect.test.ts` fake rows to carry `key` (the cached path batch-loads `SELECT key, value FROM config`).
- `bun run typecheck` exit 0; touched test files pass (13 + 41 adjacent postgres-engine tests).

Co-authored-by: Omerbahari <Omerbahari@users.noreply.github.com>

### DIFF
--- a/src/core/postgres-engine.ts
+++ b/src/core/postgres-engine.ts
@@ -5565,30 +5565,78 @@ export class PostgresEngine implements BrainEngine {
     });
   }
 
+  /**
+   * perf (#1694 by @Omerbahari): process-lifetime config cache. A single
+   * search fires ~85 getConfig() reads (loadConfigWithEngine x2, plus
+   * mode/cache/intent/rerank/graph-signals resolvers). On a remote pooler
+   * each read is a round-trip; serial they dominate query latency and can
+   * push the op handler past cli.ts's 10s disconnect force-exit, truncating
+   * stdout. First read batch-loads the whole `config` table into this Map
+   * (inside the same connRetry posture as the per-key read — #1603/#1891);
+   * setConfig/unsetConfig write through. TTL bounds staleness for
+   * multi-writer processes; GBRAIN_CONFIG_CACHE_TTL_MS=0 disables.
+   * Only present keys are stored — Map.has() distinguishes known-absent.
+   */
+  private _configCache: Map<string, string> | null = null;
+  private _configCacheLoadedAt = 0;
+  private _configCacheLoad: Promise<void> | null = null;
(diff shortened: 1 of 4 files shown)


time-attack:
Appreciate the diligence in keeping both the batch load and the TTL=0 fallback inside `connRetry()` rather than reverting master's #1603/#1891 posture — that detail is what made the takeover worth reading. I'm still closing it: this is a performance optimization that buys speed by adding a process-lifetime cache with TTL-bounded staleness, write-through, and single-flight, and cache coherency is exactly the kind of surface area we don't take on without a correctness bug forcing it (the postgres-only caching also puts the two engines out of step). CI is also red here — serial-tests fails on this branch while master's latest run is green across every job.

PR #3100, opened 2026-07-21
feat(reference): reference-only entity flag, exempt from coverage metrics

### LINKED ISSUE #1899: Canon/reference figures (imported from books/articles) sit permanently red in entity-coverage metrics with no actionable fix

time-attack:
Fixes #1899
Takeover of #1900 (@ElliotDrel) — rebased onto master with a source-scoping repair; referenced, not closing that PR here.

## What

Canon/reference figures imported from books/articles (a book's author, a company an article discusses) are real `type: person`/`company` pages — searchable, enrichable, linkable — but have no dated history in the user's own life, so they sat permanently red in `timeline_coverage` / `entity_link_coverage` with no actionable remediation.

Opt-in `reference: true` frontmatter flag exempts a page from those entity-coverage metrics ONLY; nothing about retrieval changes.

- `src/core/reference-flag.ts`: single source of truth for the exclusion predicate `(frontmatter->>'reference') IS DISTINCT FROM 'true'`. Absent/false/non-true = normal counted entity (default).
- Predicate ANDed into BOTH numerator and denominator at every entity-coverage site: `getHealth()` entity_pages CTE (pglite + postgres engines — postgres inlines the literal since tagged-template interpolation is a bound param), `onboard/checks.ts` (timeline + link coverage), `onboard/init-nudge.ts`. `most_connected` (top-N display) intentionally NOT filtered.
- New CLI-only `gbrain reference <slug> [--unset] [--source <id>] [--brain <dir>]`: writes the flag to the markdown frontmatter (durable, survives re-ingest) AND the engine JSONB (immediate, no re-sync).
- Skills: `enrich`/`book-mirror`/`article-enrichment`/`maintain` conventions + canonical `skills/conventions/reference-entities.md`; KEY_FILES entry; llms bundles regenerated.

## Repair over #1900

The original PR's JSONB UPDATEs were keyed on bare `slug` — but slug uniqueness is `(source_id, slug)`, so marking a page would clobber same-named pages in every other source. The UPDATEs are now scoped to `(source_id, slug)`, with the source resolved from the brain dir via the standard `resolveSourceId` chain (`--source`/`GBRAIN_SOURCE` override). Also rebased the cli.ts `CLI_ONLY` + KEY_FILES hunks onto current master and fixed a template-literal syntax error in the postgres CTE comment.

## Tests

`test/reference-flag.test.ts`: predicate + frontmatter-edit units (from #1900), plus new PGLite e2e pinning (a) `getHealth` timeline-coverage exemption flips with the flag, and (b) marking a page leaves its same-slug twin in another source untouched — fails without the scoping fix.

Gates: `bun run typecheck` clean; `bun test test/reference-flag.test.ts test/build-llms.test.ts test/pglite-engine.test.ts test/e2e/onboard-full-flow.test.ts` all pass; jsonb pattern/param guards + KEY_FILES current-state guard clean.


time-attack:
Thanks for catching and repairing the real defect in the original #1900 — keying the JSONB updates on bare `slug` when uniqueness is `(source_id, slug)` would have clobbered same-named pages across sources, and your added test for that is a good catch. I'm closing this anyway because it's ~437 lines of new feature surface: a new frontmatter flag, a new `gbrain reference` CLI command, engine changes on both backends, and four skill docs, all to adjust how coverage metrics count canon entities. It's also currently conflicting with master. Red coverage numbers for reference figures are a reporting annoyance rather than a broken promise, so this isn't a feature we're taking.

PR #3102, opened 2026-07-21
fix(doctor): ClawHub workspace skills no longer hard-fail resolver_health

### LINKED ISSUE #1767: doctor resolver_health treats ClawHub workspace skills as required gbrain-routable skills
## Summary

`gbrain doctor` / `gbrain check-resolvable` appears to treat all skills under an OpenClaw workspace `skills/` directory as if they are part of the gbrain skill resolver surface. That creates false hard failures for ClawHub/OpenClaw workspace skills that are runtime integrations rather than gbrain memory-workflow skills.

In my case, two ClawHub-installed OpenClaw skills caused `resolver_health` to fail:

- `agentmail`
- `printing-press-library`

Both were installed under the OpenClaw workspace skill root and had ClawHub install metadata, but lacked gbrain-style `triggers:` frontmatter and `AGENTS.md` / `RESOLVER.md` routing rows.

The issue is not that these two particular skills need better metadata. The broader problem is that gbrain is assuming "installed workspace skill" means "must be routable by gbrain's resolver."

## Environment

- gbrain: `0.42.1.0`
- Engine: PGLite
- Host: macOS / OpenClaw workspace
- Workspace-local skills directory contained only two ClawHub skills:
  - `skills/agentmail`
  - `skills/printing-press-library`

Both had `.clawhub/origin.json` metadata, e.g. registry `https://clawhub.ai`, slug, installed version, installed timestamp.

## Repro shape

1. In an OpenClaw workspace, install one or more ClawHub skills into the workspace-local skill root.
2. Do not add gbrain-specific `triggers:` frontmatter or resolver rows for those skills.
3. Run:

```bash
gbrain doctor --json --fast
# or
gbrain check-resolvable --json
```

Observed resolver output included:

```json
{
  "type": "unreachable",
  "skill": "agentmail",
  "message": "Skill 'agentmail' is in manifest but has no trigger row in RESOLVER.md or AGENTS.md"
}
```

and:

```json
{
  "type": "mece_gap",
  "skill": "agentmail",
  "message": "Skill 'agentmail' has no triggers: field in its SKILL.md frontmatter"
}
```

Same pattern for `printing-press-library`.

Doctor moved from `unhealthy` to `warnings` only after manually adding:

- `triggers:` frontmatter to both ClawHub skills
- a small routing table in `AGENTS.md` pointing to those two skills

That workaround makes the checker happy, but it feels semantically wrong: an OpenClaw email integration should not necessarily become a gbrain "brain operations" route just because it exists in the workspace skill directory.

## Why this matters

GBrain's resolver-health model makes sense for gbrain's own skillpack. Those skills are operating procedures for memory work: query, capture, ingest, enrich, maintain, citation fixing, etc. For those, checking that every skill is reachable is useful.

But OpenClaw workspace skills are a broader category. Some are runtime capabilities or app integrations, not gbrain workflow skills. Examples from this case:

- `agentmail` is tied to OpenClaw config / SecretRefs / an agent email identity.
- `printing-press-library` is a tool-discovery/catalog skill.

Treating every ClawHub workspace skill as resolver-required means any ClawHub skill without gbrain-specific metadata can make `gbrain doctor` fail, even if gbrain itself is otherwise healthy.

This also pushes gbrain-specific frontmatter expectations into the broader ClawHub ecosystem, which seems backwards: ClawHub skills should not need to carry gbrain resolver metadata unless they explicitly participate in gbrain's skill router.

## Expected behavior

One of these would be less surprising:

1. `resolver_health` only audits skills in gbrain's configured skill roots / skillpack manifest, not arbitrary OpenClaw workspace skills.
2. Workspace/ClawHub skills are treated as external or optional unless explicitly opted into gbrain resolver checking.
3. Missing resolver metadata on ClawHub workspace skills is a warning, not a hard `resolver_health` failure.
4. `check-resolvable` distinguishes "installed tool/runtime skill exists" from "gbrain workflow skill must be routable."

## Related

This is related to #386, but a different failure mode. #386 is about overlay skill roots and avoiding local-edit conflicts. This issue is specifically about resolver-health scope: gbrain is auditing OpenClaw/ClawHub workspace skills as if they are required gbrain-routable skills.

## Suggested direction

My preference would be to narrow the resolver-health input set or add an explicit opt-in/opt-out boundary for workspace skills, e.g.:

- only skills installed by `gbrain skillpack ...` are strict resolver-health participants;
- ClawHub/OpenClaw workspace skills are ignored or soft-warned by default;
- users can explicitly opt a workspace skill into resolver checking when it is intended to act as a gbrain workflow skill.

time-attack:
Fixes #1767

`gbrain doctor` / `gbrain check-resolvable` treated every workspace skill dir with a `SKILL.md` as a required gbrain-routable skill, so ClawHub-installed OpenClaw integrations (e.g. an email skill) without gbrain `triggers:` frontmatter produced hard `unreachable` errors and `mece_gap` warnings.

**Fix (root cause, in the shared manifest loader `src/core/skill-manifest.ts`):** the derive path now skips skill dirs carrying `.clawhub/origin.json` — they are externally-managed runtime integrations, not gbrain resolver skills. This covers all three consumers at once (`checkResolvable`, `dry-fix` auto-fix walker, doctor's conformance/brain-first checks), and also prevents auto-fix from rewriting externally-managed skill files.

**Opt-in back to strict checking:** a ClawHub skill that declares `triggers:` in its SKILL.md frontmatter (the same surface that makes it routable) is included and fully checked; an explicit `manifest.json` listing is likewise honored verbatim.

Tests: new cases in `test/skill-manifest.test.ts` (skip / triggers-opt-in / explicit-manifest) and an end-to-end `checkResolvable` case in `test/check-resolvable.test.ts` reproducing the issue's OpenClaw derive-path scenario — all fail without the fix.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -206,7 +206,11 @@ jobs:
     needs: cache-check
     if: needs.cache-check.outputs.hit != 'true'
     runs-on: ubuntu-latest
-    timeout-minutes: 15
+    # 20 (was 15): shard 4 runs ~14.5 min on master (dream.test.ts ~29s/test
+    # dominates it) and hits the 15-min ceiling on slower runners, cancelling
+    # mid-run with 0 test failures. Rebalancing via
+    # scripts/mine-shard-weights.ts is the real fix; this stops the bleeding.
+    timeout-minutes: 20
     strategy:
       fail-fast: false
       matrix:
--- a/src/core/skill-manifest.ts
+++ b/src/core/skill-manifest.ts
@@ -23,6 +23,15 @@
  * hold conventions and shared rule files, not skills. Files like
  * `_brain-filing-rules.md` live at the root and are not considered
  * skills by either loader.
+ *
+ * ClawHub-installed workspace skills (#1767): a skill dir carrying
+ * `.clawhub/origin.json` is an externally-managed runtime integration
+ * (e.g. an email or catalog skill), not a gbrain-routable skill. The
+ * derive path SKIPS those so `gbrain doctor` resolver_health doesn't
+ * hard-fail on them — UNLESS the skill's SKILL.md frontmatter declares
+ * `triggers:`, which is the explicit opt-in to gbrain routing (and the
+ * same surface that makes it reachable). An explicit manifest.json that
+ * lists a ClawHub skill also keeps strict checking (verbatim path).
  */
 
 import { existsSync, readFileSync, readdirSync, statSync } from 'fs';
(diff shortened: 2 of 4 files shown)


time-attack:
Merging. I confirmed on master that the derive path in `src/core/skill-manifest.ts` has no ClawHub awareness at all — every subdir with a SKILL.md is treated as a required routable skill — so #1767 still reproduces and doctor hard-fails on externally-managed integrations the user never asked gbrain to route. Fixing it once in the shared loader instead of in each of the three consumers is the right shape, and keeping strict checks for anything that opts in via `triggers:` or an explicit manifest entry means this narrows the check rather than weakening it. The shard-4 timeout bump is a fair inclusion; it's the same 15-minute ceiling that cancelled shards on other branches in this batch.

PR #3103, opened 2026-07-21
fix(autopilot): translate positional subcommands (#1525 takeover of #1529)

### LINKED ISSUE #1525: UX trap: `gbrain autopilot status` starts the daemon instead of reporting status

time-attack:
## Items

- **Fixes #1525** — `gbrain autopilot status` started the daemon instead of reporting status: `runAutopilot` branched only on flag forms (`--status`/`--install`/`--uninstall`), so a positional `status` fell through to the default daemon-start path (lockfile + worker spawn + sync dispatch).
- **Takeover of #1529** — the community fix was correct but CONFLICTING against master (autopilot.ts moved since its May base). This PR rebases it onto current master; the code is @Oszkar's, credited via Co-authored-by.

## What changed

- `src/commands/autopilot.ts`: pure `translatePositionalSubcommands()` wired into `runAutopilot` after the `--help` short-circuit and before any side effect. Maps `status`/`install`/`uninstall` to their flag forms, drops `start` (default daemon launch), is value-flag aware (`--repo`, `--interval`), and fails loud (stderr + exit 2) on unknown positionals — including `stop`, deliberately not aliased (stopping a daemon is a new feature, not an alias). Help text documents the aliases.
- `test/autopilot-positional-subcommands.test.ts`: 22 tests including the exact ticket repro (`['status']` → `['--status']`), flag/positional interleaving, value-flag edge cases, and unknown/multiple-positional rejection.

## Verification

- `bun test test/autopilot-positional-subcommands.test.ts` — 22 pass / 0 fail
- `bun test test/autopilot-install.test.ts test/autopilot-lock-path.test.ts test/autopilot-lock.test.ts test/autopilot-resolve-cli.test.ts` — 18 pass / 0 fail

### DIFF
--- a/src/commands/autopilot.ts
+++ b/src/commands/autopilot.ts
@@ -151,6 +151,97 @@ export function shouldSpawnAutopilotWorker(args: string[]): boolean {
   return !args.includes('--no-worker');
 }
 
+/**
+ * #1525 — positional subcommand translation.
+ *
+ * Pre-fix, `gbrain autopilot status` silently fell through to "start daemon"
+ * because `runAutopilot()` only branched on flag forms (`--status`, etc.).
+ * `status` was treated as a stray positional and ignored.
+ *
+ * This translator maps known positional subcommands to their flag form so
+ * `autopilot status` is equivalent to `autopilot --status`, then rejects
+ * any unrecognized positional with a fail-loud error before any side
+ * effect (lockfile, daemon spawn, sync dispatch) runs.
+ *
+ * Scope decisions:
+ *   - Known aliases: `status` → `--status`, `install` → `--install`,
+ *     `uninstall` → `--uninstall`, `start` → (drop; default daemon launch).
+ *   - `stop` is intentionally NOT aliased here. Stopping a running daemon
+ *     is a new behavior (read PID from lock, SIGTERM, drain) that deserves
+ *     its own design and PR. Users typing `gbrain autopilot stop` today get
+ *     the unknown-positional error with the canonical alternatives.
+ *   - At most one positional allowed; multiple positionals fail loud.
+ */
+const AUTOPILOT_VALUE_FLAGS = new Set(['--repo', '--interval']);
+const AUTOPILOT_POSITIONAL_ALIASES: Record<string, string | null> = {
+  status: '--status',
+  install: '--install',
+  uninstall: '--uninstall',
+  start: null, // drop the positional; default behavior is daemon launch
+};
+
+export type PositionalTranslation =
+  | { ok: true; args: string[] }
+  | {
+      ok: false;
+      reason: 'unknown_subcommand' | 'multiple_subcommands';
+      message: string;
+    };
+
+export function translatePositionalSubcommands(args: string[]): PositionalTranslation {
+  const out: string[] = [];
+  let positionalSeen = false;
+  let i = 0;
+  while (i < args.length) {
+    const a = args[i];
+    if (AUTOPILOT_VALUE_FLAGS.has(a)) {
+      // Pass through the flag and its value untouched. If the value is
+      // missing at end-of-argv, fall through so the existing parseArg
+      // path can report the broken usage.
+      out.push(a);
+      if (i + 1 < args.length) {
+        out.push(args[i + 1]);
+        i += 2;
+      } else {
+        i += 1;
+      }
+      continue;
+    }
+    if (a.startsWith('-')) {
+      out.push(a);
+      i += 1;
+      continue;
+    }
+    // Positional subcommand.
+    if (positionalSeen) {
+      const known = Object.keys(AUTOPILOT_POSITIONAL_ALIASES).join(', ');
+      return {
+        ok: false,
+        reason: 'multiple_subcommands',
+        message: `Multiple subcommands given. Use only one of: ${known}.`,
+      };
+    }
+    positionalSeen = true;
+    if (a in AUTOPILOT_POSITIONAL_ALIASES) {
+      const alias = AUTOPILOT_POSITIONAL_ALIASES[a];
+      if (alias) out.push(alias);
+      i += 1;
+      continue;
+    }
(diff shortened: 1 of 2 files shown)


time-attack:
Merging. Confirmed on master that `runAutopilot` only inspects the flag forms, so a positional `status` silently reaches the daemon-start path and writes a lockfile and spawns a worker — a read-only command with side effects is a real bug worth fixing. I checked your value-flag set against every `parseArg` call site on master and it is complete, so no valid invocation gets rejected. Good call on leaving `stop` unaliased rather than smuggling in new daemon behavior.

PR #3104, opened 2026-07-21
feat(ai): add reranker touchpoint to LiteLLM proxy recipe

time-attack:
Takeover of #2455 (credit: @ozp, co-authored on the commit).

Adds a `reranker` touchpoint to the LiteLLM proxy recipe so any rerank model registered in the user's LiteLLM config is reachable via `gbrain config set search.reranker.model litellm:<model>` — LiteLLM normalizes Cohere/Voyage/Jina to the `{model, query, documents, top_n} → {results: [{index, relevance_score}]}` shape `gateway.rerank()` already speaks, no adapter needed.

Repairs vs the original PR:

- **Leaf path `/rerank`, not `/v1/rerank`.** Master's setup_hint (updated after #2455 branched) allows `LITELLM_BASE_URL` with or without the `/v1` suffix. The PR's pinned `/v1/rerank` doubled to `/v1/v1/rerank` → 404 on `/v1`-suffixed bases, and its test only covered the suffix-less default. LiteLLM serves both `/rerank` and `/v1/rerank`, so the leaf form (the same convention `llama-server-reranker` uses) is valid either way.
- **setup_hint appends instead of replaces.** The original replacement line was based on the pre-`/v1`-suffix hint and would have dropped that guidance.
- **`cost_per_1m_tokens_usd` stays `undefined`.** A proxy can front a paid provider, so pricing-unknown is the honest state — matching this recipe's embedding/chat touchpoints and budget-tracker's deliberate exclusion of litellm from the free-provider sets (the touchpoint cost field isn't consumed by rerank pricing anyway; `FREE_LOCAL_RERANK_PROVIDERS` is).

Test (`test/ai/recipe-litellm-reranker.test.ts`) pins the recipe shape and drives `gateway.rerank()`'s real URL builder through the stubbed transport for both base-URL forms; the `/v1`-suffixed case fails with the original PR's path.

Gates: `bun run typecheck` exit 0; new + sibling reranker/gateway tests pass (106 tests, 0 fail).

### DIFF
--- a/src/core/ai/recipes/litellm-proxy.ts
+++ b/src/core/ai/recipes/litellm-proxy.ts
@@ -56,6 +56,36 @@ export const litellmProxy: Recipe = {
       cost_per_1m_output_usd: undefined,
       price_last_verified: '2026-06-14',
     },
+    // LiteLLM normalizes Cohere / Voyage / Jina / etc. rerank backends to the
+    // same wire shape gbrain's gateway.rerank() already speaks (the
+    // ZeroEntropy/llama.cpp contract):
+    //   { model, query, documents, top_n } → { results: [{ index, relevance_score }] }
+    // So any rerank model the user registers in their LiteLLM config is
+    // reachable via `gbrain config set search.reranker.model litellm:<model>`
+    // with no request/response adapter — same as embeddings ride the proxy.
+    reranker: {
+      models: [], // user-provided; whatever rerank models the proxy serves
+      // No canonical default — the proxy defines its own model ids. The user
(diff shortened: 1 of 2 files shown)


time-attack:
Closing — this is new vendor-integration surface rather than a fix for something broken, and third-party proxy touchpoints are a category we are not expanding right now. The engineering is careful (the leaf `/rerank` path and the append-not-replace setup_hint are both the right calls over the original #2455), and none of that is in question. It is purely a scope decision about how much vendor surface the gateway carries. Thanks for the work.

PR #3105, opened 2026-07-21
fix(gateway): refresh eval cross-modal defaults + native chat allowlists

### LINKED ISSUE #1607: eval cross-modal defaults + native-provider chat allowlists are stale; native-Anthropic gateway 404s valid models; providers list reports unvalidated keys as ready
## Summary

A cluster of related staleness/validation issues surfaced while running `gbrain eval cross-modal` (v0.41.23.0). Net effect: the genuine multi-provider eval is hard to run because the shipped defaults and native-provider allowlists have drifted behind both the live provider APIs and gbrain's own model registry.

## 1. `eval cross-modal` default slot models are all dead

Defaults (`src/commands/eval-cross-modal.ts`): `openai:gpt-4o` / `anthropic:claude-opus-4-7` / `google:gemini-1.5-pro`.
- `gpt-4o` → rejected by gbrain's own OpenAI recipe allowlist (`Model "gpt-4o" is not listed for OpenAI chat`).
- `claude-opus-4-7` → **404 at the Anthropic API** (deprecated).
- `gemini-1.5-pro` → 404 at the Google API + needs a key.

So an out-of-the-box `eval cross-modal` returns `inconclusive: 0 of 3 models parseable`.

## 2. Native-provider chat allowlists are deprecated

- `recipes/openai.ts` chat: `['gpt-5.2', 'gpt-4o-mini']` (rejects `gpt-4o`).
- `recipes/google.ts` chat: `['gemini-2.0-flash-exp', 'gemini-2.0-flash', 'gemini-1.5-pro']` — **all three 404 at the live API** ("no longer available to new users"), while current models (`gemini-2.5-flash`, `gemini-flash-latest`) work at the API but are **rejected by the allowlist** ("not listed"). Net: native Google chat is unusable for new keys.
- The registry alias `gemini → google:gemini-3-pro` (from `gbrain models`) also disagrees with the recipe allowlist.

## 3. Native-Anthropic gateway 404s a valid key+model

`anthropic:claude-sonnet-4-6` via the gateway → `[chat(anthropic:claude-sonnet-4-6)] Not Found`. The same key + model via a direct `POST https://api.anthropic.com/v1/messages` (anthropic-version 2023-06-01) → **HTTP 200**. So the key and model are valid; gbrain's native-Anthropic chat path (via `@ai-sdk/anthropic`) is 404ing. Routing the same model through OpenRouter (`openrouter:anthropic/claude-sonnet-4.6`) works, which points at the native path specifically (SDK version / endpoint).

## 4. `gbrain providers list` reports unvalidated keys as "✓ ready"

A 4-character placeholder `OPENROUTER_API_KEY` showed `openrouter … ✓ ready`, then every call returned `400 Bad Request` (empty body). Readiness is apparently a non-empty-env-var check, not a format/auth check — misleading when debugging which providers actually work.

## Suggested fixes

- Refresh `eval cross-modal` default slots to current models (e.g. `openai:gpt-5.2` / a current `claude-*` / a current `gemini-*`).
- Refresh the native chat recipe allowlists to current models, or make the allowlist **advisory (warn-not-throw) for native providers too** (it's already non-enforced for the openai-compat tier), letting provider 404s surface naturally.
- Investigate the native-Anthropic gateway 404 (likely `@ai-sdk/anthropic` version or endpoint construction) — a valid key+model 404s only through the native path.
- Strengthen `providers list` readiness: validate key format (or do a cheap auth probe) rather than reporting any non-empty var as ready.

Filed from a real skillify-adoption effort (FoundryLabs/morty-v3 LABS-93) where the only way to get a multi-provider `eval cross-modal` running was to route everything through OpenRouter passthrough.

time-attack:
Fixes #1607 (points 1 and 2 — the halves adversarially confirmed on current master).

## What changed

- **`eval cross-modal` default slots were all dead out of the box** (`openai:gpt-4o` rejected by gbrain's own openai chat allowlist; `anthropic:claude-opus-4-7` and `google:gemini-1.5-pro` deprecated upstream). Defaults now resolve to models present in each recipe's current chat allowlist: `openai:gpt-5.2` / `anthropic:claude-opus-4-8` / `google:gemini-3-pro`. Help text updated to match.
- **Google chat allowlist refreshed**: adds `gemini-3-pro` (now agrees with the registry alias in `model-config.ts:DEFAULT_ALIASES`), `gemini-2.5-pro`, `gemini-2.5-flash`, `gemini-flash-latest`; drops ids no longer served to new keys (`gemini-1.5-pro`, `gemini-2.0-flash-exp`). `gemini-2.0-flash` kept for existing keys. Config-derived models still bypass the allowlist via `extendedModels`, so nobody's existing config breaks.
- **OpenAI chat allowlist** adds `gpt-5` (matches `DEFAULT_ALIASES.gpt`, which the allowlist was contradicting), `gpt-5.5`, and `gpt-4o` — all live API models already carried in canonical pricing.
- **Canonical pricing** gains `openai:gpt-5.2` and `google:gemini-3-pro` rows so the new defaults keep working cost estimates (single-table rule respected; consumers untouched).

## Anti-rot guard

New `test/model-defaults-in-allowlists.test.ts`: every hardcoded default model string (cross-modal `DEFAULT_SLOTS`, `DEFAULT_ALIASES`, `TIER_DEFAULTS`) must pass `assertTouchpoint` against its recipe's chat allowlist, and `DEFAULT_SLOTS` must be priced from canonical. Hardcoded defaults don't get the config `extendedModels` escape hatch, so this is exactly the drift class that produced the bug — the test fails on pre-fix master.

## Not addressed here (issue points 3 and 4)

- Native-Anthropic gateway 404 on a valid key+model: filed at v0.41.23, not reproducible from a code snapshot — needs a live re-test against current `@ai-sdk/anthropic` before any change.
- `providers list` reporting unvalidated keys as ready: a real ergonomics gap, but adding an auth probe is a product decision (cost/latency of probing on list), left open on #1607.

## Verification

- `bun run typecheck` exit 0
- `bun test` on the 8 touched/adjacent files: 154 pass / 0 fail (`model-defaults-in-allowlists`, `model-pricing`, `cross-modal-phase1`, `cross-modal-eval-cli`, `eval-cross-modal-batch`, `ai/gateway-chat`, `model-config.serial`, `anthropic-model-ids`)

### DIFF
--- a/src/commands/eval-cross-modal.ts
+++ b/src/commands/eval-cross-modal.ts
@@ -76,9 +76,9 @@ FLAGS:
                            dimensions (goal, depth, sourcing, specificity, useful).
   --cycles N               1-3. Default: 3 in TTY, 1 in non-TTY (T11). Each
                            cycle is 3 model calls; verdict aggregates over them.
-  --slot-a-model <id>      Override default 'openai:gpt-4o'.
-  --slot-b-model <id>      Override default 'anthropic:claude-opus-4-7'.
-  --slot-c-model <id>      Override default 'google:gemini-1.5-pro'.
+  --slot-a-model <id>      Override default 'openai:gpt-5.2'.
+  --slot-b-model <id>      Override default 'anthropic:claude-opus-4-8'.
+  --slot-c-model <id>      Override default 'google:gemini-3-pro'.
   --receipt-dir <path>     Default: gbrainPath('eval-receipts').
   --max-tokens N           Output token budget per call. Default: 4000.
   --json                   Emit final aggregate as JSON to stdout (progress to stderr).
--- a/src/core/ai/recipes/google.ts
+++ b/src/core/ai/recipes/google.ts
@@ -23,11 +23,20 @@ export const google: Recipe = {
       price_last_verified: '2026-04-20',
     },
     chat: {
-      models: ['gemini-2.0-flash-exp', 'gemini-2.0-flash', 'gemini-1.5-pro'],
+      // gemini-1.5-pro and gemini-2.0-flash-exp 404 at the live API for new
+      // keys (#1607); gemini-2.0-flash kept for existing keys. gemini-3-pro
+      // matches the registry alias in model-config.ts:DEFAULT_ALIASES.
+      models: [
+        'gemini-3-pro',
+        'gemini-2.5-pro',
+        'gemini-2.5-flash',
+        'gemini-flash-latest',
+        'gemini-2.0-flash',
+      ],
       supports_tools: true,
       supports_subagent_loop: true,
       supports_prompt_cache: false,
-      max_context_tokens: 1000000, // Gemini 1.5 Pro
+      max_context_tokens: 1000000, // Gemini 2.5/3 Pro
       cost_per_1m_input_usd: 0.30,
       cost_per_1m_output_usd: 1.20,
       price_last_verified: '2026-04-20',
--- a/src/core/ai/recipes/openai.ts
+++ b/src/core/ai/recipes/openai.ts
@@ -30,7 +30,9 @@ export const openai: Recipe = {
       price_last_verified: '2026-04-20',
     },
     chat: {
-      models: ['gpt-5.2', 'gpt-4o-mini'],
+      // gpt-5 matches the registry alias in model-config.ts:DEFAULT_ALIASES;
+      // gpt-4o/gpt-5.5 are live API models carried in model-pricing.ts (#1607).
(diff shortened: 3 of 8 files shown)


time-attack:
Closing. The bug this leads with is already fixed on master — slot A is `openai:gpt-5.2` and in the allowlist, and `test/cross-modal-default-slots.test.ts` already pins every default against its recipe touchpoint, which is the same anti-rot guard the PR adds. What is left is vendor model-catalog refresh across two allowlists plus new pricing rows, which we keep on our own cadence rather than taking as drive-by expansion. If a specific default is confirmed 404ing against a live key, a one-line default swap with the receipt would be an easy call.

PR #3106, opened 2026-07-21
fix(schema-pack): gbrain-base-v2 capability parity, full bundled-pack listing, calibration holder config

### LINKED ISSUE #1726: Observability: list_schema_packs hardcodes 2 of 7 bundled packs; no persistent config key for calibration holder
## Summary

Two small observability/config gaps surfaced while auditing a Railway+Supabase thin-client deployment on v0.42.1.0. Neither breaks the engine; both make the system harder to reason about. Filing together since both are one-line-ish fixes.

## Bug A — `list_schema_packs` MCP op hardcodes the bundled list

`src/core/operations.ts` (the `list_schema_packs` handler) has:

```ts
const bundled = ['gbrain-base', 'gbrain-recommended'];
```

But `src/core/schema-pack/base/` actually ships **7** packs:

```
gbrain-base.yaml  gbrain-base-v2.yaml  gbrain-recommended.yaml
gbrain-creator.yaml  gbrain-engineer.yaml  gbrain-investor.yaml  gbrain-everything.yaml
```

So `list_schema_packs` (and `get_active_schema_pack`'s sibling discovery) under-reports: an agent asking "what packs can I activate?" sees 2 of 7. The packs that declare the `extract_atoms` / `synthesize_concepts` phases (`gbrain-creator`, `gbrain-everything`) are invisible through the MCP surface, so a user can't discover that those nightly phases even exist without reading the repo. The loader itself reads the real directory; only the reporting list is frozen. Suggest deriving `bundled` from a directory read (same source the loader uses) instead of the literal array.

## Bug B — no persistent config key for the calibration holder (asymmetric with emotional_weight)

`src/core/config.ts` registers `emotional_weight.user_holder` (line ~688) so salience can be pointed at a non-default holder. There is **no** parallel `calibration.user_holder`. The calibration holder defaults to `'garry'` hardcoded in three places:

- `src/core/cycle/calibration-profile.ts:229` — `const holder = opts.holder ?? 'garry'`
- `src/commands/calibration.ts:161` and `:243` — same `?? 'garry'`

The only way to change it is the per-invocation `--holder` flag / MCP `holder` param. For a self-hosted brain owned by someone other than `garry`, the nightly `calibration_profile` phase always queries holder `garry`, finds 0 resolved takes, and skips — silently, forever — even though the user's takes live under their own holder slug. Setting `emotional_weight.user_holder` does NOT help (different subsystem). Suggest a `calibration.user_holder` config key that the cycle's `calibration_profile` phase reads, symmetric to `emotional_weight.user_holder`, so the nightly phase can target the brain owner without a flag.

Happy to send a PR for either if useful.

### LINKED ISSUE #2109: gbrain-base-v2 removes the meeting type but extract-timeline-from-meetings hardcodes type='meeting' — unify-types silently kills the feature
## Summary
The `gbrain-base-v2` schema pack removes the `meeting` page type entirely (it is not among the 15 declared types, nor an alias of any), but `extract-timeline-from-meetings` hardcodes the type filter:

```
src/core/extract-timeline-from-meetings.ts:66   WHERE type = 'meeting'
src/core/extract-timeline-from-meetings.ts:88   AND pf.type = 'meeting'
```

After running the v0.42 `unify-types` migration (gbrain-base → gbrain-base-v2), all `meeting` pages are catch-all retyped to `note`, so the extractor silently scans **0 meetings** — while `onboard --check` keeps recommending `extract-timeline-from-meetings` as the remedy for low `timeline_coverage`. The feature dies with no error.

## Repro (v0.42.26.0, Postgres engine)
1. Brain with N pages of `type: meeting` (granola-style meeting pages under `meetings/`).
2. `gbrain jobs submit unify-types --allow-protected --params '{"target_pack":"gbrain-base-v2"}'` — completes cleanly, meetings retyped to `note` (`legacy_type: meeting` preserved).
3. `gbrain jobs submit extract-timeline-from-meetings` → result `{"meetings_scanned":0,"entries_created":0}`.
4. `onboard --check` still recommends the job.

Before the migration the same job scanned all meetings; after restoring `type='meeting'` (via a local pack fork that re-adds the type) it scanned 256.

## Expected
Either (a) gbrain-base-v2 keeps/declares a `meeting` type (it declares `email` and `slack` as temporal types — meetings are the same class), or (b) type-dependent features like `extract-timeline-from-meetings` resolve types through the active pack (alias/prefix expansion) instead of hardcoding `'meeting'`.

## Workaround used
Forked the bundled pack to `~/.gbrain/schema-packs/<fork>/pack.yaml`, re-added `meeting` (primitive `temporal`, prefix `meetings/`), restored the retyped rows from `legacy_type`. Works, but diverges from the bundled pack. Related papercut: `gbrain schema fork gbrain-base-v2 <new>` fails with "Source pack not found" even though the pack is bundled and active (`schema list` doesn't show it either) — the manual copy was required.

### LINKED ISSUE #2117: `gbrain-base-v2` ships no `phases:` declaration and zero `link_types[].inference` regexes -- extract_atoms is silently pack-gated off and extract-ner returns `pack_unavailable`, making two of onboard's own recommendations inert
### Version

gbrain 0.42.40.0 (clone @ `ecd6ae87`), Postgres engine. Bundled pack `gbrain-base-v2@1.0.0`.

### Summary

Two gaps in the bundled `gbrain-base-v2` pack, found after migrating a brain from `gbrain-base@1.0.0` via `unify-types`:

1. **No `phases:` key.** Cycle phases like `extract_atoms` run only when the active pack declares them. `gbrain-base-v2` ships without any `phases:` declaration, so on a v2 brain `extract_atoms` is silently pack-gated off: autopilot cycles report ok, doctor's `extract_atoms_backlog` grows, and `extract_health` degrades -- there is no signal pointing at the pack.
2. **Zero `link_types[].inference` regexes.** `gbrain-base` v1 ships link-inference rules (founded, works_at, invested_in, attended). v2 declares the link types but no inference regexes at all, so `gbrain jobs submit extract-ner` returns `pack_unavailable` on a brain whose active pack is the current bundled default.

The sharp edge: after the v2 migration, `gbrain onboard` recommends running atom extraction and extract-ner -- both inert against the very pack the same onboard flow recommends installing.

### Repro

1. Migrate a brain to `gbrain-base-v2` via the onboard-recommended `unify-types`.
2. `gbrain jobs submit extract-atoms-drain --follow` -> the phase is pack-gated; backlog does not drain. `gbrain doctor` shows `extract_atoms_backlog` warn with the detail "active pack does not run extract_atoms".
3. `gbrain jobs submit extract-ner --follow` -> `pack_unavailable`.

### Observed

On our brain: `extract_atoms` had never run post-migration (backlog 463 pages), `extract-ner` hard-skipped. Doctor `extract_health` accumulated a high halt rate with no hint that the pack was the cause.

### Expected

The bundled successor pack should preserve the cycle-phase participation and inference capability of the pack it migrates away from, or the migration should warn loudly about the capability delta. Concretely: a `phases: [extract_atoms]` declaration and the v1 inference regexes (or successors) in `gbrain-base-v2`.

### Local mitigation

A local pack copy (`gbrain-base-v2-hss`) that re-adds `phases: [extract_atoms]` and ports the 4 v1 inference regexes. Verified: `extract_atoms_backlog` flipped warn -> OK with "active pack runs extract_atoms each cycle", extract-ner produced 160 links. Cost: the local copy must be manually re-synced on every upstream pack release, which is why we are filing rather than keeping the divergence quietly.

### Related

- #2109 (v2 drops the `meeting` type that extract-timeline-from-meetings hardcodes) is a third capability regression in the same pack -- same theme: v2 silently sheds capabilities v1 users depend on.
- #1750 is the same end state (a bundled pack shipping inert) for `gbrain-recommended`, via a YAML-parser bug rather than missing keys.

time-attack:
Backlog work unit c29: bundled-pack resolution + v2 pack capability gaps.

## Fixes

**Fixes #2109** — `extract-timeline-from-meetings` hardcoded `type='meeting'` in both SQL sites (meeting walk + attended-edge join). gbrain-base-v2's unify-types catch-all retypes meetings to `note` with `frontmatter.legacy_type='meeting'`, so migrated brains silently scanned 0 meetings while onboard kept recommending the job. Both sites now also match `frontmatter->>'legacy_type' = 'meeting'`; pre-unify `type='meeting'` brains behave exactly as before. New PGLite test fails without the fallback.

**Fixes #2117** — `gbrain-base-v2` shipped no `phases:` declaration and zero `link_types[].inference` regexes, so `extract_atoms` was silently pack-gated off and `extract-ner` returned `pack_unavailable` on the bundled default pack. The pack now declares `phases: [extract_atoms]` (v2 declares the `atom` type the phase writes) and ports v1's `founded`/`works_at`/`invested_in` inference regexes. v1's `attended` inference is page_type-bound to `meeting`, which v2 doesn't declare — porting it would fail the `link_types_undeclared_page_type` lint, so retyped meeting pages are covered by the #2109 legacy_type fallback instead (noted in the yaml).

**Fixes #1726** — both halves:
- (A) `list_schema_packs` hardcoded 2 of 7 bundled packs. The locator's registry is now the exported `BUNDLED_PACKS` const in `load-active.ts` and the op derives from it, so the MCP surface and the locator can't drift again.
- (B) New `calibration.user_holder` config key, symmetric with `emotional_weight.user_holder`. `resolveCalibrationHolder()` (explicit param → config → `'garry'`) is threaded through the nightly `calibration_profile` phase, the `gbrain calibration` CLI, and the `get_calibration_profile` op — a self-hosted brain owned by someone else no longer needs a per-invocation `--holder` flag for the nightly phase to find their takes. Fail-open: config lookup errors fall back to the legacy default.

## Not changed (deliberate)

- `src/core/brainstorm/orchestrator.ts` keeps its existing `emotional_weight.user_holder` fallback for calibration-profile lookup (typed config-slice plumbing; separate cleanup if wanted).
- `mutate.ts`'s `BUNDLED_PACK_NAMES` read-only guard (3 packs) untouched — different concern (mutation protection vs discovery), out of scope here.
- No `meeting` type re-added to v2 (product call); the legacy_type fallback fixes already-migrated brains with zero data migration.

## Tests

- `test/extract-timeline-legacy-type.test.ts` (new): retyped-meeting scan, v1 behavior preserved, plain notes not scanned.
- `test/schema-pack-base-v2-capabilities.test.ts` (new): phases + extract-ner `hasRegex` predicate + lint guard.
- `test/operations-schema-pack.test.ts`: bundled list equals the registry.
- `test/calibration-profile.test.ts`: config-driven holder + explicit-wins precedence.

`bun run typecheck` clean; 385 tests across the touched/neighboring suites (schema-cli, mutate, unify-types, onboard, lens-pack manifests, extract-atoms gating, calibration, e2e type-unification + onboard-full-flow) all pass.

### DIFF
--- a/src/commands/calibration.ts
+++ b/src/commands/calibration.ts
@@ -19,7 +19,7 @@
  */
 
 import type { BrainEngine } from '../core/engine.ts';
-import { runPhaseCalibrationProfile } from '../core/cycle/calibration-profile.ts';
+import { resolveCalibrationHolder, runPhaseCalibrationProfile } from '../core/cycle/calibration-profile.ts';
 import { sourceScopeOpts, type OperationContext } from '../core/operations.ts';
 import type { GBrainConfig } from '../core/config.ts';
 import { GBrainError } from '../core/types.ts';
@@ -167,7 +167,7 @@ export async function runCalibration(
   config: GBrainConfig,
 ): Promise<void> {
   const { opts } = parseArgs(args);
-  const holder = opts.holder ?? 'garry';
+  const holder = await resolveCalibrationHolder(engine, opts.holder);
   // Resolve --source / GBRAIN_SOURCE / .gbrain-source so the (now reachable, #2035)
   // calibration command targets the right source in a multi-source brain instead
   // of always reading `default`. No signal → 'default' (prior behavior).
@@ -253,14 +253,14 @@ export async function getCalibrationProfileOp(
   ctx: OperationContext,
   params: { holder?: string },
 ): Promise<CalibrationProfileRow | null> {
-  const holder = params.holder ?? 'garry';
-  if (typeof holder !== 'string' || holder.length === 0) {
+  if (params.holder !== undefined && (typeof params.holder !== 'string' || params.holder.length === 0)) {
     throw new GBrainError(
       'INVALID_HOLDER',
       'get_calibration_profile.holder must be a non-empty string',
-      'pass holder="<slug>" or omit to default to "garry"',
+      'pass holder="<slug>" or omit to default to the calibration.user_holder config (then "garry")',
     );
   }
+  const holder = await resolveCalibrationHolder(ctx.engine, params.holder);
   const scope = sourceScopeOpts(ctx);
   return getLatestProfile(ctx.engine, { holder, ...scope });
 }
--- a/src/core/config.ts
+++ b/src/core/config.ts
@@ -928,6 +928,10 @@ export const KNOWN_CONFIG_KEYS: readonly string[] = [
   // Emotional weight (v0.29)
   'emotional_weight.high_tags',
   'emotional_weight.user_holder',
+  // Calibration holder (#1726): persistent default for the nightly
+  // calibration_profile phase + `gbrain calibration`, symmetric with
+  // emotional_weight.user_holder. Falls back to 'garry' when unset.
+  'calibration.user_holder',
   // Cycle phase config
   'cycle.grade_takes.write_gstack_learnings',
   // Content sanity (v0.41)
--- a/src/core/cycle/calibration-profile.ts
+++ b/src/core/cycle/calibration-profile.ts
@@ -96,7 +96,7 @@ export type PatternStatementsGenerator = (input: {
 export type BiasTagsGenerator = (patterns: string[]) => Promise<string[]>;
 
 export interface CalibrationProfileOpts extends BasePhaseOpts {
-  /** Holder to generate the profile for. Default 'garry'. */
+  /** Holder to generate the profile for. Default: `calibration.user_holder` config, then 'garry'. */
   holder?: string;
   /** Inject the patterns generator (tests). */
   patternsGenerator?: PatternStatementsGenerator;
@@ -194,6 +194,26 @@ export function parseBiasTagsOutput(raw: string): string[] {
     .slice(0, 4);
 }
 
+/**
+ * #1726: resolve the calibration holder. Explicit param wins, then the
+ * persistent `calibration.user_holder` config key (symmetric with
+ * emotional_weight.user_holder), then the legacy 'garry' default. Fail-open:
+ * a missing config table / mock engine without getConfig falls through.
+ */
+export async function resolveCalibrationHolder(
+  engine: BrainEngine,
+  explicit?: string,
+): Promise<string> {
+  if (explicit) return explicit;
+  try {
+    const configured = await engine.getConfig('calibration.user_holder');
+    if (configured && configured.trim().length > 0) return configured.trim();
+  } catch {
+    // Config unavailable — use the legacy default.
+  }
+  return 'garry';
+}
+
 /** Pick the "loudest" pattern slot for the template fallback. */
 function pickFallbackSlots(scorecard: TakesScorecard): PatternStatementSlots {
   if (!scorecard || scorecard.resolved === 0) {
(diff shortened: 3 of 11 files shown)


time-attack:
Closing as-is. The bundled-pack listing half is already fixed on master — `list_schema_packs` derives from `BUNDLED_PACK_NAMES` in `schema-pack/bundled.ts` and returns all seven packs — and this PR would regress that by introducing a second `BUNDLED_PACKS` registry in `load-active.ts`, which is precisely the drift `bundled.ts` exists to prevent. The #2109 legacy_type fallback and the #2117 base-v2 phases gap are both genuinely still broken on master and I would look at either one as a focused standalone PR against the existing registry. Thanks for the thorough write-up.

PR #3107, opened 2026-07-21
fix(skillopt): emit proposed.md in no-mutate mode

### LINKED ISSUE #2635: skillopt --no-mutate: best.md is written but proposed.md is never emitted

time-attack:
Fixes #2635 (takeover of #2719 by @RerankerGuo, credited as co-author — not closing that PR, just referencing).

## What

`gbrain skillopt --no-mutate` (and the bundled-skill no-mutate path) promised a `skillopt/proposed.md` review artifact (documented in `help.ts` and the SkillOpt tutorial) but only ever wrote `best.md`; the orchestrator then returned the `best.md` path labeled as the proposal. Automation watching for `proposed.md` saw nothing.

## Fix

- `version-store.ts` `writeProposed` now writes both `best.md` (optimizer's current-best pointer) and `proposed.md` (the stable review artifact), returning the proposal path. Fixed in the shared helper so both accept-branch call sites in `orchestrator.ts` are covered.
- `orchestrator.ts` returns the real `proposedPath` instead of aliasing `bestPath`.
- Tutorial updated to describe the emitted artifact.

## Tests

- New `writeProposed` unit test in `test/skillopt/version-store.test.ts` (fails without the fix: returned path was `best.md` and `proposed.md` was never created).
- Extended `test/e2e/skillopt-loop.serial.test.ts` no-mutate case to pin both artifacts.
- `bun run typecheck` exit 0; `bun test test/skillopt/version-store.test.ts test/e2e/skillopt-loop.serial.test.ts` 23 pass / 0 fail.

### DIFF
--- a/docs/tutorials/improving-skills-with-skillopt.md
+++ b/docs/tutorials/improving-skills-with-skillopt.md
@@ -233,13 +233,14 @@ keep it or `git checkout` to throw it away. Nothing is committed for you.
 
 **For a skill that ships with gbrain** (anything under the gbrain repo's own
 `skills/`): SkillOpt refuses to overwrite it by default and writes the winner to
-`skills/<name>/skillopt/best.md` instead, so an optimization pass can never
-silently mutate a skill other people depend on. Two ways to handle that:
+`skills/<name>/skillopt/proposed.md` instead (while keeping `best.md` as the
+optimizer's current-best pointer), so an optimization pass can never silently
+mutate a skill other people depend on. Two ways to handle that:
 
 ```bash
 # See the proposed improvement without touching SKILL.md (works for ANY skill):
 gbrain skillopt meeting-prep --split 1:1:1 --no-mutate
-# → writes skills/meeting-prep/skillopt/best.md (the proposed rewrite), prints its path. Copy what you want.
+# → writes skills/meeting-prep/skillopt/proposed.md, updates best.md, and prints the proposal path.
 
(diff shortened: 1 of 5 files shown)


time-attack:
Closing: master already contains this exact fix via #3182 (takeover of #2719, merged earlier today) — the pile-repair track and the fix wave independently produced the same change, and review confirmed this branch now diffs to nothing meaningful against master. The skillopt proposed.md issue it targeted is resolved.

PR #1476, opened 2026-05-26
fix(cli): wire gbrain bench publish dispatcher (closes #1474)

Mr-B-1:
## Problem

The v0.41.1 wave shipped `src/commands/bench-publish.ts` (exporting `runBenchPublish`), documented `gbrain bench publish` in [`docs/eval-bench.md`](docs/eval-bench.md), and advertised the verb in [CHANGELOG.md v0.41.1.0](CHANGELOG.md), but `src/cli.ts` was never updated with the dispatcher case.

On master @ 32f8be96 (v0.41.14.0):

```bash
$ gbrain bench publish --from /tmp/captured.ndjson --to /tmp/baseline.ndjson
Unknown command: bench
Run gbrain --help for available commands.

$ grep "command === 'bench'" src/cli.ts
# (no matches)
```

The eval-loop documented in v0.41.1 CHANGELOG (`eval export → bench publish → eval gate`) is broken end-to-end on a fresh install.

## Fix

Two surgical edits to `src/cli.ts` mirroring the existing `case 'eval':` pattern:

1. Add `'bench'` to the `CLI_ONLY` set so the dispatcher reaches the case statement instead of rejecting at the cliOps lookup
2. Add a `case 'bench':` block that sub-dispatches `publish` to `runBenchPublish` (mirrors `case 'eval':` at cli.ts:1287-1290)

14 lines added, 1 deleted.

## Verification

Applied locally + tested on a v0.41.14.0 install:

```bash
$ gbrain bench publish
Error: --from FILE is required

gbrain bench publish — write a baseline file from captured queries
[full usage text follows...]

$ gbrain bench publish --from /tmp/captured.ndjson --to /tmp/baseline.ndjson --label test
Wrote /tmp/baseline.ndjson
  Label:    test
  Rows:     17
  [...]
$ gbrain eval gate --baseline /tmp/baseline.ndjson
[runs the regression gate to completion]
```

Closes #1474.

(Filed by Mr-B-1 — same v0.40.8 → v0.41.14 upgrade audit that produced sibling issues #1452-1455, #1475.)

PR #3108, opened 2026-07-21
fix(cli): wire gbrain bench publish dispatcher + honor DB-plane eval.capture

### LINKED ISSUE #1474: gbrain bench publish: CLI verb documented but dispatcher missing in v0.41.13.0 (Unknown command: bench)
**Title:** `gbrain bench publish` CLI verb documented + implemented but dispatcher wiring missing — `Unknown command: bench` on v0.41.13.0

**Version:** v0.41.13.0 (master @ `f56cc619`)

**Symptom:**

```bash
$ gbrain --version
gbrain 0.41.13.0

$ gbrain bench publish --from /tmp/captured.ndjson --to ~/.gbrain/baselines/test.baseline.ndjson --label "test"
Unknown command: bench
Run gbrain --help for available commands.

$ gbrain bench --help
Unknown command: bench
```

**Root cause:**

1. `docs/eval-bench.md` documents the user-facing invocation as `gbrain bench publish --from ... --to ...`
2. `CHANGELOG.md` v0.41.1.0 advertises **New CLI verbs:** `gbrain bench publish --from <captured.ndjson> --to <X.baseline.ndjson> [...]`
3. `src/commands/bench-publish.ts` implements `export async function runBenchPublish(args: string[]): Promise<void>` (line 269) and is fully functional
4. But `src/cli.ts` has NO dispatcher case for `command === 'bench'`. Verified via `grep -nE "command === '(bench|bench-publish)'" src/cli.ts` → 0 hits
5. Direct `bun run src/commands/bench-publish.ts --from ... --to ...` also fails silently because the file has no `if (import.meta.main)` main entry — it only exports the function

**Impact:**

The v0.41.1 "eval-loop" is broken end-to-end. Users can `gbrain eval export` and `gbrain eval gate`, but cannot create a baseline file with the documented command — leaving the regression gate unusable without a hand-edited NDJSON.

**Reproduction:**

```bash
git checkout master  # v0.41.13.0
bun install
gbrain bench publish --help    # → Unknown command: bench
bun run src/commands/bench-publish.ts --help   # → silent exit (no main entry)
grep -nE "command === 'bench'" src/cli.ts      # → no matches
```

**Suggested fix (likely ~15 lines in cli.ts):**

Add a `case 'bench':` block to the main command switch in `cli.ts` (alongside the existing `case 'eval':` at ~line 1280-1290) that:
1. Reads the sub-subcommand from `args[0]`
2. If `args[0] === 'publish'`: dynamic-imports `./commands/bench-publish.ts` and calls `runBenchPublish(args.slice(1))`
3. Otherwise: prints `Usage: gbrain bench publish ...` help

Optionally also add `if (import.meta.main) { runBenchPublish(process.argv.slice(2)) }` to `bench-publish.ts` so direct `bun run` works as a fallback.

**Workaround:** I wrote a shim at `scripts/run-bench-publish.ts` that imports `runBenchPublish` and invokes it with argv. Happy to PR the cli.ts fix if useful — would be a near-mechanical mirror of the existing `case 'eval':` pattern.

(Filed via gbrain upgrade audit by Mr-B-1 — sibling issues #1452-1455 from the same audit.)

### LINKED ISSUE #1475: gbrain config set eval.capture true persists but is not honored at runtime (capture silently no-ops without GBRAIN_CONTRIBUTOR_MODE=1)
**Title:** `gbrain config set eval.capture true` persists to disk but is not honored at runtime — capture only fires via `GBRAIN_CONTRIBUTOR_MODE=1`

**Version:** v0.41.14.0 (master @ `32f8be96`, Postgres engine)

**Symptom:**

```bash
$ gbrain config set eval.capture true
Set eval.capture = true

$ gbrain config get eval.capture
true

$ gbrain query "test capture" --limit 5 --detail low
# (query runs, returns results)

$ psql -c 'SELECT count(*) FROM eval_candidates;'
 count
-------
     0      # → no rows captured

$ psql -c 'SELECT count(*) FROM eval_capture_failures;'
 count
-------
     0      # → no failures logged either — capture not even attempted
```

Now with explicit env var:

```bash
$ GBRAIN_CONTRIBUTOR_MODE=1 gbrain query "test capture via env var" --limit 5 --detail low
# (same query, same brain, same flags)

$ psql -c 'SELECT count(*) FROM eval_candidates;'
 count
-------
     1      # → row captured
```

**Root cause (suspected):**

`src/core/eval-capture.ts:196-199` gate:

```typescript
export function isEvalCaptureEnabled(config: GBrainConfig | null | undefined): boolean {
  if (config?.eval?.capture === true) return true;
  if (config?.eval?.capture === false) return false;
  return process.env.GBRAIN_CONTRIBUTOR_MODE === '1';
}
```

`gbrain config set eval.capture true` writes the value to disk (verified — `gbrain config get eval.capture` returns `true`). But the runtime `config` object passed to `isEvalCaptureEnabled` does NOT contain `config.eval.capture`. So the dotted-key set lands in storage but the parsed `GBrainConfig` shape doesn't expose it under the expected nested path, and the gate falls through to the env-var fallback.

Likely a `loadConfig` / `parseConfig` issue that doesn't promote `eval.capture` dotted-key into `config.eval.capture` nested.

**Inspected:**
- `/root/.gbrain/config.json` contains only `{engine, embedding_model, embedding_dimensions, database_url, chat_model, chat_fallback_chain}` — no `eval` key visible there. The `config set eval.capture true` may be writing to a separate preferences file or a Postgres `gbrain_config` table that the runtime doesn't merge.

**Impact:**

Operators following the canonical eval-capture docs / CHANGELOG flow:
```
gbrain config set eval.capture true
# run queries
gbrain eval export --tool query > baseline.ndjson
```
…get an empty NDJSON and a silent failure. The eval-loop documented in v0.41.1 CHANGELOG (`gbrain bench publish` + `gbrain eval gate`) cannot be used without first discovering the env-var workaround.

**Workaround applied:**

Added systemd drop-in `Environment=GBRAIN_CONTRIBUTOR_MODE=1` to `gbrain-mcp.service` and `gbrain-worker.service`. Capture now fires for MCP traffic + CLI invocations that inherit the env. Tracked as local config in LOCAL_PATCHES.md.

**Suggested fix:**

Either (a) ensure `gbrain config set eval.capture true` writes a nested-shape entry that `loadConfig` materializes into `GBrainConfig.eval.capture`, or (b) document that `eval.capture` is env-var-only and reject `config set eval.capture` with a helpful error pointing at the env var.

(Filed by Mr-B-1 — sibling issues #1452-1455, #1474 from the same v0.40.8 → v0.41.14 upgrade audit.)

time-attack:
Two verified backlog items, one work unit (c136):

## Takeover of #1476 — bench dispatcher (closes #1474)

`gbrain bench publish` was documented (docs/eval-bench.md, CHANGELOG) but the cli.ts dispatcher case was never added: `bench` was absent from `CLI_ONLY` and `runBenchPublish` was imported nowhere but its test, so the command hit `Unknown command`.

#1476 had the right approach but was DIRTY against master — its diff rewrote the whole `CLI_ONLY` line from a stale v0.41.14 snapshot, which would have deleted ~15 commands master added since (calibration, onboard, connect, skillopt, quarantine, self-upgrade, advisor, watch, ...). This PR re-applies the fix against master's current set:

- Adds just the `'bench'` token to `CLI_ONLY`.
- Routes `bench` through a **no-DB bypass** before `connectEngine` (bench publish is pure file I/O — reads captured NDJSON, writes a baseline NDJSON), so the documented export → publish → gate loop works on machines without a brain.
- Adds `'bench'` to `CLI_ONLY_SELF_HELP` so `--help` reaches bench-publish's own detailed usage text.
- Unknown subcommand / bare `bench` → usage on stderr, exit 2.

Credit to @Mr-B-1 for the original diagnosis + fix (co-authored on the commit). Not closing #1476 here — just referencing.

## Fixes #1475 — `config set eval.capture true` silently no-oped

`gbrain config set eval.capture true` wrote the DB plane (key is allowlisted in `KNOWN_CONFIG_KEYS`, so `set` accepted it silently), but the capture gate `isEvalCaptureEnabled` reads `ctx.config` — the sync file-plane `loadConfig()` — and `loadConfigWithEngine` had no `eval.*` merge. The value was written and never read; capture only fired via `GBRAIN_CONTRIBUTOR_MODE=1`.

Fix, following the existing multimodal-flag pattern in the same code paths:

- `loadConfigWithEngine` sparse-merges `eval.capture` / `eval.scrub_pii` from the DB plane (file/env wins per key, DB fills gaps — same precedence as content_sanity/dream merges).
- `connectEngine` stashes the merged values on `GBRAIN_EVAL_CAPTURE` / `GBRAIN_EVAL_SCRUB_PII` (verbatim the `GBRAIN_EMBEDDING_MULTIMODAL` stash pattern), covering every ctx path (CLI ops, MCP stdio, HTTP) without making context-building async.
- The gates consult the stash when the file plane is silent. Explicit per-key setting beats the broad `CONTRIBUTOR_MODE` flag, matching how file-plane `false` already wins.

## Tests

- `test/bench-cli-dispatch.test.ts` (new): spawns the real CLI — `bench --help` reaches bench-publish help (fails on master with `Unknown command`), unknown subcommand exits 2, full publish roundtrip writes a baseline file. No DB required.
- `test/loadConfig-merge.test.ts`: 3 new tests for the `eval.*` DB-plane merge (repro, per-key file-wins precedence, no-keys leaves `cfg.eval` undefined).
- `test/eval-capture.test.ts`: 4 new tests for the env stash (repro, stash-false vs CONTRIBUTOR_MODE, file-plane precedence, scrub default).

All 5 non-bench tests fail without the src fix; all pass with it. `bun run typecheck` exits 0; all 4 touched test files pass (66/66).

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -54,7 +54,7 @@ export function bigintToStringReplacer(_key: string, value: unknown): unknown {
 }
 
 // CLI-only commands that bypass the operation layer
-export const CLI_ONLY = new Set(['init', 'reinit-pglite', 'upgrade', 'post-upgrade', 'check-update', 'integrations', 'publish', 'check-backlinks', 'lint', 'report', 'import', 'export', 'files', 'embed', 'serve', 'call', 'config', 'doctor', 'migrate', 'eval', 'sync', 'extract', 'extract-conversation-facts', 'enrich', 'features', 'autopilot', 'graph-query', 'jobs', 'agent', 'apply-migrations', 'skillpack-check', 'skillpack', 'resolvers', 'integrity', 'repair-jsonb', 'orphans', 'sources', 'mounts', 'dream', 'check-resolvable', 'routing-eval', 'skillify', 'smoke-test', 'providers', 'storage', 'repos', 'code-def', 'code-refs', 'reindex', 'reindex-code', 'reindex-frontmatter', 'code-callers', 'code-callees', 'reconcile-links', 'frontmatter', 'auth', 'friction', 'claw-test', 'book-mirror', 'takes', 'think', 'salience', 'anomalies', 'calibration', 'transcripts', 'models', 'remote', 'recall', 'forget', 'edges-backfill', 'cache', 'ze-switch', 'founder', 'brainstorm', 'lsd', 'schema', 'capture', 'onboard', 'conversation-parser', 'status', 'connect', 'skillopt', 'quarantine', 'self-upgrade', 'advisor', 'watch', 'reindex-search-vector']);
+export const CLI_ONLY = new Set(['init', 'reinit-pglite', 'upgrade', 'post-upgrade', 'check-update', 'integrations', 'publish', 'check-backlinks', 'lint', 'report', 'import', 'export', 'files', 'embed', 'serve', 'call', 'config', 'doctor', 'migrate', 'eval', 'bench', 'sync', 'extract', 'extract-conversation-facts', 'enrich', 'features', 'autopilot', 'graph-query', 'jobs', 'agent', 'apply-migrations', 'skillpack-check', 'skillpack', 'resolvers', 'integrity', 'repair-jsonb', 'orphans', 'sources', 'mounts', 'dream', 'check-resolvable', 'routing-eval', 'skillify', 'smoke-test', 'providers', 'storage', 'repos', 'code-def', 'code-refs', 'reindex', 'reindex-code', 'reindex-frontmatter', 'code-callers', 'code-callees', 'reconcile-links', 'frontmatter', 'auth', 'friction', 'claw-test', 'book-mirror', 'takes', 'think', 'salience', 'anomalies', 'calibration', 'transcripts', 'models', 'remote', 'recall', 'forget', 'edges-backfill', 'cache', 'ze-switch', 'founder', 'brainstorm', 'lsd', 'schema', 'capture', 'onboard', 'conversation-parser', 'status', 'connect', 'skillopt', 'quarantine', 'self-upgrade', 'advisor', 'watch', 'reindex-search-vector']);
 // CLI-only commands whose handlers print their own --help text. These are
 // excluded from the generic short-circuit so detailed per-command and
 // per-subcommand usage stays reachable.
@@ -104,6 +104,9 @@ const CLI_ONLY_SELF_HELP = new Set([
   // `gbrain connect --help` prints its own usage (flags + examples) from
   // runConnect; route around the generic one-line short-circuit.
   'connect',
+  // #1474: bench-publish ships its own detailed HELP (flags, exit codes,
+  // the export → publish → gate loop). Route around the generic stub.
+  'bench',
 ]);
 
 // v114 (#1941): alias -> operation lookup, kept separate from `cliOps` so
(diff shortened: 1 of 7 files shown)


time-attack:
Merging. Both defects reproduce on master — `runBenchPublish` has no dispatcher call site at all, and `eval.capture` is allowlisted for `config set` but never read back by the capture gate, so the write silently no-ops. I particularly appreciate that you re-applied the CLI_ONLY change as a single added token instead of replaying #1476's stale full-line rewrite, which would have deleted about fifteen commands master has added since. Thanks for the careful rebase.

PR #3112, opened 2026-07-21
fix(extract): path-style wikilinks fall back to tail-keyed basename lookup; add ops/ to DIR_PATTERN

### LINKED ISSUE #2576: extract --stale silently disables basename resolution (nullResolver) + DIR_PATTERN gaps for non-standard dirs
## Summary

`gbrain extract --stale` silently disables basename resolution by swapping in a `nullResolver` (no `resolveBasenameMatches`), diverging from the working `extractLinksFromDB` path. Additionally, `DIR_PATTERN` omits several directory prefixes (e.g. `ops/`), causing path-style qualified wikilinks like `[[ops/services/pointer-agent]]` to fall through to broken basename lookup instead of being recognized as qualified refs.

## Bug 1: `--stale` path uses `nullResolver` instead of real resolver

**File:** `src/commands/extract.ts`, `extractStaleFromDB` (~line 1618)

The stale extraction path sets:
```ts
const resolver = makeResolver(engine, { mode: 'batch' });
const nullResolver = { resolve: async () => null as string | null };
const activeResolver = includeFrontmatter ? resolver : nullResolver;
```

When `--include-frontmatter` is not passed (the default), `nullResolver` is used. It has no `resolveBasenameMatches` method, so the gate at `link-extraction.ts:485` (`!opts.globalBasename || typeof resolver.resolveBasenameMatches !== 'function'`) skips basename resolution entirely — regardless of whether `globalBasename` is enabled.

The non-stale path (`extractLinksFromDB`) always uses the real `resolver` and passes `{ skipFrontmatter: !includeFrontmatter, globalBasename }` to `extractPageLinks`. Frontmatter suppression is handled by the `skipFrontmatter` opt, not by swapping in a null resolver.

**Fix:** Pass the real `resolver` always; thread `{ skipFrontmatter, globalBasename }` opts (mirroring `extractLinksFromDB` ~line 1396).

## Bug 2: `DIR_PATTERN` omits `ops/` (and potentially other domain prefixes)

**File:** `src/core/link-extraction.ts` (~line 86)

```ts
const DIR_PATTERN = '(?:people|companies|meetings|concepts|deal|civic|project|projects|source|media|yc|tech|finance|personal|openclaw|entities)';
```

Brains with `ops/` directories (change records, runbooks, service notes) have their `[[ops/services/pointer-agent]]` wikilinks silently dropped during extraction because `ops` isn't in the whitelist. The refs never enter the qualified resolver's ancestor-walk + pg_trgm fuzzy path.

**Fix:** Add `ops` to `DIR_PATTERN`. Also: `normalizeBasename` strips slashes (`[^a-z0-9\s-]`), so path-style inputs like `ops/changes/2026-05-01-pointer-agent` normalize to `opschanges2026-05-01-pointer-…` — never matching the tail-keyed basename index. Extract the path tail before normalizing.

## Bug 3: `LINK_EXTRACTOR_VERSION_TS` needs periodic bumping

After running `extract --stale` once, all pages get stamped `links_extracted_at = now`. A subsequent `extract --stale` (after a fix to the extraction logic) finds 0 stale pages and no-ops. The version timestamp is designed to force re-extraction when the extractor logic changes, but it hasn't been bumped since `2026-05-31`.

## Reference implementation

We've validated a bundled fix for all three issues. Patch available here: https://github.com/drizzyenterprises/asset-007-hermes-runtime/pull/27

## Environment

- gbrain 0.42.53.0
- Postgres + pgvector
- 340 pages, 317 with un-extracted edges after initial bulk sync

time-attack:
Fixes #2576 (bug 2 — the part PR #2717 scoped out; bugs 1+3 are covered by #2717).

## What

- **`ops` added to `DIR_PATTERN`** (`src/core/link-extraction.ts`): `[[ops/services/pointer-agent]]` wikilinks and bare `ops/...` slug refs are now recognized as qualified entity references instead of falling through to the generic bare-wikilink pass.
- **Tail fallback in `queryBasenameIndex`** — the ONE shared basename matcher behind `makeResolver.resolveBasenameMatches`, the FS resolver, and the doctor `link_resolution_opportunity` check. Path-style refs from dirs outside the whitelist (e.g. `[[runbooks/2026-05-01-x]]`) used to be mangled by `normalizeBasename` (slashes stripped → garbage key that can never hit the tail-keyed index). They now fall back to the path tail, so qualified refs resolve by basename like everything else. Fixing it in the shared query fixes all three surfaces at once.
- **`LINK_EXTRACTOR_VERSION_TS` bumped** so `extract --stale` re-sweeps previously-stamped pages under the new extraction logic (the mechanism bug 3 describes).

## Tests

Two new tests in `test/link-extraction.test.ts`; both fail on master, pass with the fix:
- `recognizes ops/ qualified wikilinks (issue #2576 bug 2)`
- `resolveBasenameMatches: path-style ref falls back to the tail (issue #2576 bug 2)`

`bun run typecheck` exit 0; link-extraction test files (192 tests across 4 files) all pass.

## Design note

Went with extending the `DIR_PATTERN` constant rather than a `link_resolution.*` config-extensible whitelist — smallest correct diff; the config surface can land later if more domain prefixes show up.

### DIFF
--- a/src/commands/extract.ts
+++ b/src/commands/extract.ts
@@ -1743,7 +1743,15 @@ async function extractStaleFromDB(
       // `page.updated_at.toISOString()` — the JS Date is ms-truncated, so the
       // µs-precision DB updated_at stayed strictly greater and the page never
       // cleared on Postgres. Stamping the exact value makes them equal.
-      processedRefs.push({ slug: page.slug, source_id: page.source_id, extractedAt: page.updated_at_iso });
+      //
+      // Version-arm floor: a page last edited BEFORE LINK_EXTRACTOR_VERSION_TS
+      // would otherwise be stamped below the version watermark and stay
+      // permanently stale (`links_extracted_at < versionTs` re-fires every run).
+      // Stamp max(updated_at, versionTs) — versionTs is always a past release
+      // date, so a concurrent edit's now() still exceeds the stamp and D4 holds.
+      // Tie at ms precision picks updated_at_iso (its µs ≥ versionTs's .000000).
+      const stampTs = new Date(page.updated_at_iso) >= new Date(versionTs) ? page.updated_at_iso : versionTs;
+      processedRefs.push({ slug: page.slug, source_id: page.source_id, extractedAt: stampTs });
     }
 
     // Flush NON-swallowing (CDX-4): a throw here propagates out of the sweep so
--- a/src/core/link-extraction.ts
+++ b/src/core/link-extraction.ts
@@ -28,7 +28,7 @@ import { ensureWellFormed } from './text-safe.ts';
  * OR updated_at > links_extracted_at`. It is an ISO-8601 string (NOT a number) —
  * the column is TIMESTAMPTZ and the predicate binds it as `::timestamptz`.
  */
-export const LINK_EXTRACTOR_VERSION_TS = '2026-05-31T00:00:00Z';
+export const LINK_EXTRACTOR_VERSION_TS = '2026-07-21T00:00:00Z';
 
 // ─── Entity references ──────────────────────────────────────────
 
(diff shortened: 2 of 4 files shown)


time-attack:
Merging. I confirmed all three defects are live on current master: `DIR_PATTERN` still lacks `ops`, `queryBasenameIndex` still has no path-tail fallback, and the version-arm stamping floor is a real permanent-staleness bug. Fixing it inside the shared basename matcher rather than at each of the three call sites is the right call, and the diff stays small. Thanks.

PR #3113, opened 2026-07-21
fix(takes): kill propose_takes rescan loop, keep every claim, and route cycle/brainstorm models through config

time-attack:
Takeover of #2804 and Takeover of #2805 (stacked community PRs by @p3ob7o, rebased onto current master). References only — closes nothing.

## Rescan loop + dropped claims (takeover of #2804)

- **Zero-claim scans enter the idempotency cache.** Pre-fix only proposal rows were written, so a page whose extraction yielded no gradeable claims missed the `(source_id, page_slug, content_hash, prompt_version)` cache on EVERY cycle and re-spent the extractor LLM call forever. The phase now writes a `status='empty'` sentinel row (invisible to the review queue — the pending index is partial on `status='pending'`).
- **Multi-claim pages keep every claim.** The 4-column unique index made claims 2..N conflict with claim 1 and `ON CONFLICT DO NOTHING` silently dropped them. The idempotency index gains `md5(claim_text)`; the 4-column prefix still serves the per-scan cache lookup. `proposals_inserted` now counts rows that landed (`RETURNING id`), not attempts.
- **Migration renumbered 123 → 125** — the flaw named in review: master shipped v123 (`configurable_fts_language`) and v124 after the PR was cut, so the PR's duplicate v123 would never have run (`LATEST_VERSION = max(versions)`, runner applies `version > current`). Schema parity across `schema.sql`, `pglite-schema.ts`, `schema-embedded.ts`.

## Model-tier honoring (takeover of #2805, reworked for current master)

Master moved since the PR was cut (#2451 made stored model ids provider-prefixed because bare ids make `gateway.chat()` throw; propose-takes now labels via `getChatModel()`; #2997 established the `opts.model ?? getChatModel()` idiom). The PR's bare-tail-everywhere convention conflicted with those newer, deliberate semantics, so the resolution chain was reworked:

- `propose_takes`, `grade_takes`, `calibration_profile` resolve once via `resolveModel(models.<phase> > models.default > GBRAIN_MODEL > getChatModel())` — per-phase config keys are honored, and the gateway chat model (already tier-resolved by `reconfigureGatewayWithEngine`) stays the default. Stored ids remain full provider-prefixed strings, nested prefixes (`openrouter:anthropic/...`) stay intact, and master's #2451 tests pass unchanged.
- `grade_takes` fixes the still-live label-vs-actual mismatch (judge call rode the gateway `chat_model` while the cache key / budget label recorded a hardcoded `claude-sonnet-4-6`). The bare tail survives ONLY as the grade-cache key / stored `judge_model_id` for continuity with historical cache rows; the actual judge call gets the full resolved string.
- `brainstorm` resolves via `models.brainstorm` > reasoning tier (was hardcoded).
- `gbrain models` lists the four new per-task keys; output-cap and context-window entries added for newer Claude models (verified against the current model catalog; values are conservative and clamped by `MAX_OUTPUT_TOKENS_CEIL`).

## Tests

- `test/propose-takes-rescan.test.ts` (new, hermetic PGLite + injected extractor): second run cache-hits both the zero-claim and multi-claim pages, run 3 is fully quiescent, sentinel invisible to pending queue, 3-claim page stores 3 rows, content change re-extracts.
- Config-key coverage added in `test/propose-takes.test.ts` and `test/grade-takes.test.ts` (full string to the call, correct stored/cache label).
- `bun run typecheck` exit 0; 254 tests across the 6 touched files pass, plus 166 across brainstorm/synthesize/schema-bootstrap/migrate suites; jsonb guards clean.

References #2804 and #2805.

### DIFF
--- a/src/commands/models.ts
+++ b/src/commands/models.ts
@@ -50,6 +50,10 @@ const PER_TASK_KEYS: Array<{ key: string; tier: ModelTier; description: string }
   { key: 'models.eval.contradictions_judge', tier: 'utility',  description: 'Contradiction probe judge (v0.34 temporal-aware)' },
   { key: 'models.expansion',                tier: 'utility',   description: 'Query expansion for hybrid search' },
   { key: 'models.chat',                     tier: 'reasoning', description: 'Default `gateway.chat()` model' },
+  { key: 'models.propose_takes',            tier: 'reasoning', description: 'propose_takes claim extractor' },
+  { key: 'models.grade_takes',              tier: 'reasoning', description: 'grade_takes verdict judge' },
+  { key: 'models.calibration_profile',      tier: 'reasoning', description: 'Calibration profile generator' },
+  { key: 'models.brainstorm',               tier: 'reasoning', description: '`gbrain brainstorm` orchestrator' },
 ];
 
 interface ModelEntry {
--- a/src/core/brainstorm/judges.ts
+++ b/src/core/brainstorm/judges.ts
@@ -61,7 +61,10 @@ export const MAX_OUTPUT_TOKENS_CEIL = 32_000;
  * (with a readable error) instead of the provider's opaque HTTP 400.
  */
 export const ANTHROPIC_OUTPUT_CAPS: Record<string, number> = {
+  'claude-fable-5': 64_000,
+  'claude-opus-4-8': 32_000,
   'claude-opus-4-7': 32_000,
+  'claude-sonnet-5': 64_000,
   'claude-sonnet-4-6': 64_000,
   'claude-haiku-4-5': 64_000,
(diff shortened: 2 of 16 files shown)


time-attack:
Closing — the multi-claim data-loss half already shipped on master as migration v125 (`take_proposals_per_claim_idempotency`, #2138) with the same `md5(claim_text)` index, and `propose-takes.ts` already carries the matching ON CONFLICT clause. Because this PR also claims v125, its migration would collide with master's and never run, which is a hazard on its own. The zero-claim `status='empty'` sentinel is a real remaining cost bug and would be welcome as a small standalone PR, but not bundled with a stale migration and a new per-phase model-config surface.

PR #3116, opened 2026-07-21
fix(config): route remaining hand-rolled configureGateway callsites through buildGatewayConfig

time-attack:
Takeover of #2430 (credit: @TheAngryPit) — rebased onto master. Master already adopted the adapter for `providers.ts` (`configureFromEnv` + `runTest`) and `configureGatewayIfUninitialized`; this lands the remaining hand-rolled `configureGateway` callsites, which dropped file-plane API keys, env-provided provider base URLs, and `provider_chat_options`:

- **`src/commands/eval-takes-quality.ts`** — real bug: `configureGateway({ ...cfg, ...process.env } as any)` never populated the gateway's `env` field (the gateway never reads `process.env` at call time), so availability/diagnose checks dereferenced `undefined.env[k]` and threw, and file-plane keys were dropped.
- **`src/commands/eval-cross-modal.ts`** — `configureGatewayForCli` hand-rolled both branches; now one `buildGatewayConfig(loadConfig() ?? {})` call. Preserves master's `provider_chat_options` threading (the adapter carries it), which the original PR branch predated.
- **`src/commands/init.ts`** — all three `configureGateway` sites (merged-precedence helper, PGLite init, Postgres init) overlay the resolved model fields on `loadConfig()` and route through the adapter, so the init-time gateway sees the same keys/base URLs runtime will.
- **`src/commands/migrations/in-process.ts`** — `runMigrateOnlyCore` same conversion.

Tests:
- New `test/eval-takes-quality-gateway.test.ts` — fails (throws) on the pre-fix code path, passes with the fix; `resetGateway()` in `afterAll`.
- `LLAMA_SERVER_RERANKER_BASE_URL` added to the adapter passthrough sweep (was the one env base-URL var without coverage).
- `test/cli-multimodal-integration.test.ts` now imports the real adapter instead of maintaining a hand-copied mirror of it.

Docs: `docs/architecture/KEY_FILES.md` entries for `eval-cross-modal.ts` and `build-gateway-config.ts` updated to current state; `bun run build:llms` run (bundles unchanged — KEY_FILES is linked, not inlined).

Dropped from the original branch: the `providers.ts` hunk (already on master) and the API-key passthrough tests (master's adapter test already covers key folding + the empty-string clobber guard).

Verification: `bun run typecheck` exit 0; `bun test test/eval-takes-quality-gateway.test.ts test/ai/build-gateway-config.test.ts test/cli-multimodal-integration.test.ts test/init-embed-check.test.ts test/eval-cross-modal-batch.test.ts test/migration-in-process.serial.test.ts test/init-migrate-only.test.ts test/eval-takes-quality-cli.test.ts test/brain-score-recommendations.test.ts test/v0_37_fix_wave.serial.test.ts test/build-llms.test.ts` all pass.

No version bump / CHANGELOG — release bookkeeping happens at ship time.

Co-authored-by: TheAngryPit <TheAngryPit@users.noreply.github.com>


time-attack:
Thanks — merging this. I confirmed on master that `eval-takes-quality.ts` still spreads `process.env` into `configureGateway`, which leaves the gateway's `env` map undefined and makes the availability/diagnose path throw on `_config.env[k]`, so this is a live crash rather than a cleanup. Routing the remaining callsites through the adapter master already uses elsewhere is the right shape, and the new test fails on the pre-fix code path. Please rebase — the branch currently conflicts with master.

PR #3117, opened 2026-07-21
feat(facts): owner-trusted local fact reads via facts.trust_local_reads (takeover of #2427)

time-attack:
Takeover of #2427 (credit: @jeanpierre121, co-authored on the commit). Salvages the owner-trusted-reads feature with its trust regression repaired; references #2427 but closes nothing.

## What

Fact rows are `private` | `world`; remote callers are world-only. The stdio MCP server is a local, unauthenticated pipe — on a single-owner machine its caller IS the owner — yet it defaults `remote: true`, so the owner's own agent got an empty `find_trajectory` and world-only `recall` over MCP.

This adds `trustedFactReads`: a narrow, READ-ONLY trust elevation decoupled from `remote`, so every other remote protection (file confinement, source isolation, fence stripping, takes-holder scoping) stays fully in force. Opt-in via the new `facts.trust_local_reads` config key (default off). Only the stdio server sets it; HTTP dispatch never does, so a served brain stays world-only regardless.

- `src/core/facts/reader-trust.ts` (new): `factsWorldOnly` / `readableFactVisibilities` — single decision point for fact-read visibility.
- All four visibility sites route through it: `findTrajectory` (postgres + pglite engines, in lockstep), the recall op, and the facts meta-hook.
- `OperationContext.trustedFactReads` + `TrajectoryOpts.trustedFactReads`; `dispatch.ts` threads the flag; stdio `server.ts` reads the config best-effort (a config blip falls back to world-only).

## Repair vs #2427 as submitted

#2427's `factsWorldOnly` was **fail-open** (`remote === true && …`): an unset `remote` read private rows. It replaced two fail-closed sites (recall's visibility filter and the facts meta-hook, both `ctx.remote === false ? undefined : ['world']` on master), and its own test codified the regression (`factsWorldOnly({}) === false`), violating the CLAUDE.md trust invariant ("anything not strictly `false` is remote/untrusted").

Repaired: `factsWorldOnly` is now fail-closed (`t.remote !== false && t.trustedFactReads !== true`) and the truth-table test pins it. The two engine `findTrajectory` call sites keep the direct-engine default-local contract (unset `remote` at the engine layer = trusted in-process caller, matching master's `opts.remote === true` filter) by normalizing to explicit booleans before calling the helper.

## Dropped from #2427

The claim-time fact dating half (`segmentValidFrom`) already landed on master via #2958 (`seg.startIso` with the epoch guard) — not re-applied.

## Tests

- `test/facts-reader-trust.test.ts`: fail-closed truth table for both helpers (fails against #2427's version — `factsWorldOnly({})` must be `true`).
- `test/engine-find-trajectory.test.ts`: new case — `remote=true + trustedFactReads` returns private points while plain `remote=true` stays world-only.
- `bun run typecheck` exit 0; touched suites green (85 pass / 0 fail).

### DIFF
--- a/src/core/config.ts
+++ b/src/core/config.ts
@@ -908,6 +908,10 @@ export const KNOWN_CONFIG_KEYS: readonly string[] = [
   'facts.extraction_model',
   // #2113: output-token cap for the per-turn facts extractor (default 4000).
   'facts.extraction_max_tokens',
+  // Owner opt-in: let the local stdio MCP pipe read this owner's private facts
+  // (find_trajectory / recall). Default off; HTTP transport ignores it. See
+  // src/core/facts/reader-trust.ts.
+  'facts.trust_local_reads',
   // Dream cycle config
   'dream.synthesize.session_corpus_dir',
   'dream.synthesize.meeting_transcripts_dir',
--- a/src/core/engine.ts
+++ b/src/core/engine.ts
@@ -591,6 +591,8 @@ export interface TrajectoryOpts {
   sourceIds?: string[];
   /** When true, filters to visibility='world' only. Set by MCP layer from ctx.remote. */
   remote?: boolean;
+  /** Owner opt-in: read private facts despite `remote`. See facts/reader-trust.ts. */
+  trustedFactReads?: boolean;
   /** Metric filter. When set, only facts with this canonical metric label participate. */
   metric?: string;
   /**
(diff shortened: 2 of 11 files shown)


time-attack:
Thank you for the careful salvage work here — spotting that #2427's `factsWorldOnly` was fail-open and pinning a fail-closed truth table is genuinely good analysis. I'm closing it anyway: this adds a second trust axis (`trustedFactReads`) next to `remote` plus a new config key, and we keep trust decisions on one fail-closed marker by design. Private fact reads already work over the trusted local CLI, so this is a convenience gap rather than a broken promise, and new feature surface is not something we're taking right now.

Issue #3068, opened 2026-07-21
sync: failed internal `git pull` is downgraded to a warning but the sync anchor still advances — source reports `up_to_date` forever, `sync --full` cannot recover


loweaxerium:
## Summary

When the internal `git pull` of a source sync fails (visible only with verbose: `sync.git_pull error`), gbrain logs a warning, continues, reports the source `up_to_date` with 0 changes — and **advances the sync checkpoint anyway**. From that point the source is permanently stale: subsequent syncs keep reporting `up_to_date` even after the repo is manually pulled, and **`gbrain sync --all --full` does not recover either** (still 0 modified on the poisoned checkpoint). The only recovery that works is destroying and re-registering the source.

This is the same failure class as #574 (pull failure → warning → misleading "Already up to date", fixed for the detached-HEAD case in v0.30.3 via #635), but for any other pull-failure cause, and with the added twist that the checkpoint advancement makes it unrecoverable through any supported CLI path (#2707 confirms there is no anchor-reset CLI).

## Environment

- gbrain **0.42.59.0** (installed via bun)
- macOS (Darwin 25.5)
- Native local Postgres engine (migrated from PGLite)
- Embeddings: ollama `nomic-embed-text` (768d)
- Two federated sources: `default` → `~/notes` (a wiki repo), `teamwiki` → `~/teamwiki-mirror`
- `~/teamwiki-mirror` is a git clone whose **`origin` is a local filesystem path** (a checkout of a submodule of the first repo: `git clone ~/notes/team-wiki ~/teamwiki-mirror`)

Reproduced on 4 consecutive nights by a scheduled (launchd) `gbrain sync --all`.

## Steps to reproduce

1. Create a repo and a local clone of it:
   ```bash
   mkdir -p ~/upstream-repo && cd ~/upstream-repo && git init
   echo "# seed" > seed.md && git add -A && git commit -m seed
   git clone ~/upstream-repo ~/mirror-clone   # origin = local path
   ```
2. Register the clone as a federated source and sync once (imports fine):
   ```bash
   gbrain sources add mirror --path ~/mirror-clone --federated
   gbrain sync --source mirror
   ```
3. Commit new content to the upstream repo:
   ```bash
   cd ~/upstream-repo
   echo "zzuniquetoken99 new content" > new-page.md && git add -A && git commit -m "new page"
   ```
4. Sync the source again:
   ```bash
   gbrain sync --source mirror --verbose
   ```

## Actual behavior

[...]

PR #3121, opened 2026-07-21
fix(security): whoami over stdio, auto_link_allow_remote opt-in, sync file-transport for trusted durability origin (backlog part3-1)

time-attack:
Part 3 — Singleton P0s from the backlog fix wave. Three community-PR takeovers, each salvaged and repaired per its adversarial-review finding. References only — closes nothing.

## Takeover of #2376 — whoami over stdio returns identity instead of unknown_transport
Bug on master: `src/mcp/dispatch.ts` defaults `remote: true` and auth is only threaded from HTTP, so every `whoami` over the default stdio transport (Claude Desktop/Code) hits the unconditional `unknown_transport` throw.

Salvaged: the transport-tag threading (`DispatchOpts.transport` / `OperationContext.transport`, set to `'stdio'` by `src/mcp/server.ts` and `'http'` by both HTTP transports).

Repaired: the original PR returned `{transport: 'local', scopes: []}` for stdio — the exact shape `operations.ts` documents as the TRUSTED local-CLI marker, making untrusted stdio callers indistinguishable from the trusted surface. This version returns a distinct `{transport: 'stdio', scopes: []}`; `'local'` stays exclusive to `ctx.remote === false`, and an auth-less HTTP transport still fails closed.

Credit: @Facundo-Moreno-Dev

## Takeover of #2478 — auto_link_allow_remote flag for trusted MCP callers
Salvaged: `isAutoLinkRemoteAllowed` (default-deny, opt-in truthiness), the fail-closed gate at `put_page`, unchanged trust semantics (`config set` is CLI-only so a remote caller can't flip it for itself).

Repaired: stripped the unrelated, unadvertised `think/gather.ts` excerptLen 600→3000 default change and the `think.excerpt_length` config read that the diff smuggled in; registered `auto_link_allow_remote` in `KNOWN_CONFIG_KEYS` (instead of the smuggled `think.` prefix); wrote the promised unit tests (default-deny, opt-in values, fail-closed on garbage).

Credit: @onchito-walks

## Takeover of #2709 — file transport only for the trusted local durability origin
Bug on master: `GIT_SSRF_FLAGS` pins `protocol.file.allow=never` and sync's `pullRepo(gitContextRoot)` spreads it unconditionally, so a brain whose origin is the local bare durability repo under `GBRAIN_HOME` fails every `sync.git_pull` with `transport 'file' not allowed` (the `GBRAIN_GIT_ALLOW_FILE_TRANSPORT` hatch covers only the durability helpers).

Salvaged: the entire `git-remote.ts` design — strict default, `allowLocalFileOrigin` opt-in, realpath-containment gate (`isTrustedLocalOrigin`), network remotes never relaxed, mutation-checked argv tests.

Repaired: the PR was stale — its sync.ts hunk targeted a call site that no longer exists. Re-targeted at the current `pullRepo(gitContextRoot)` site; dropped the PR's sourceId-based gating (the containment check is the real guard, and durability-hardened non-default sources need the fix too).

Credit: @hunterbohm

## Skipped: #2745 (loss-evidence manifest foundation)
Not included. Beyond the migration-number collision (v123 is taken on master), the PR's `migrate.ts` hunk also changes `runMigrations` verify-retry semantics globally (best-effort → throw `MigrationDriftError` for ALL migrations), and the maintainer question on foundation-without-consumers scope is still open. That's a product decision, not a mechanical rebase.

## Tests
- `bun test test/whoami.test.ts test/link-extraction.test.ts test/git-remote.test.ts` — 188 pass / 0 fail
- Each fix carries regression tests that fail without it (stdio whoami shape + HTTP fail-closed preserved; auto_link_allow_remote default-deny/opt-in; pullRepo strict-default / trusted-root relax / outside-root strict / https strict)

### DIFF
--- a/src/commands/serve-http.ts
+++ b/src/commands/serve-http.ts
@@ -1768,6 +1768,7 @@ export async function runServeHttp(engine: BrainEngine, options: ServeHttpOption
       try {
         toolResult = await dispatchToolCall(engine, name, params as Record<string, unknown> | undefined, {
           remote: true,
+          transport: 'http',
           takesHoldersAllowList: tokenAllowList,
           sourceId: tokenSourceId,
           metaHook: getBrainHotMemoryMeta,
--- a/src/commands/sync.ts
+++ b/src/commands/sync.ts
@@ -1960,7 +1960,15 @@ async function performSyncInner(engine: BrainEngine, opts: SyncOpts): Promise<Sy
       // timeout (ETIMEDOUT / SIGTERM on err.cause) from ordinary pull
       // failure. Pull applies to the whole git repo (gitContextRoot), not
       // just the sync scope — git has no per-subdir pull.
-      pullRepo(gitContextRoot);
+      //
+      // #2709: allow the file transport ONLY for the trusted local durability
+      // origin (the gbrain-managed bare repo under GBRAIN_HOME). pullRepo
+      // re-verifies via realpath containment that the resolved origin is a
+      // local path under the trusted root before relaxing, so federated
+      // https sources and arbitrary local origins keep full SSRF hardening.
+      // Without this, a durability-hardened brain fails every sync.git_pull
+      // with "transport 'file' not allowed".
+      pullRepo(gitContextRoot, { allowLocalFileOrigin: true });
       serr(`[gbrain phase] sync.git_pull done ${Date.now() - _t0}ms`);
     } catch (e: unknown) {
       const msg = e instanceof Error ? e.message : String(e);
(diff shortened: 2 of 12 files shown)


time-attack:
Closing — the whoami half already landed on master via #1061, and with the exact shape you proposed: `whoami` returns `{transport: 'stdio', scopes: []}` when there's no auth and the stdio marker is set, `dispatch.ts` threads the transport tag, and `server.ts` sets it, with `'local'` still reserved for `ctx.remote === false`. Of what's left, `auto_link_allow_remote` is new opt-in config surface we aren't taking, and the git file-transport hunk sits on a sync path master has since reworked around #3068. Thanks for the adversarial repairs on the three upstream PRs — that analysis was the useful part and it's reflected in what shipped.

Issue #2038, opened 2026-06-10
migrate: renumbered migration recorded without executing (v102 timeline_entries_source_in_dedup) → all timeline writes fail


brandobuilds:
## Bug

Our brain reads `schema_version` 111, but `idx_timeline_dedup` was still the 3-column `(page_id, date, summary)` shape. Migration **v102 `timeline_entries_source_in_dedup`** — whose own comment says "originally v99, bumped to v102 after master merge" — never executed: the version counter passed it via the pre-renumber chain, so it was recorded as applied without its SQL ever running.

An artifact audit of v101–v109 against the live schema showed **only v102 missing** (link_kind column, migration_impact_log, atom-hash idx, slug_aliases, extract_rollup_7d, generation clock, embedding_signature, newest_content_at all present).

## Consequence at v0.42.x

Every `addTimelineEntry` / `addTimelineEntriesBatch` call fails:

```
there is no unique or exclusion constraint matching the ON CONFLICT specification
```

because both code paths infer on the 4-column index `(page_id, date, summary, source)` (`src/core/postgres-engine.ts`, both insert sites at `eefe8b57`). On our mesh this silently broke timeline writes on every node, including the hub.

## Recovery (what we did)

Ran the migration's own SQL by hand:

```sql
DROP INDEX IF EXISTS idx_timeline_dedup;
CREATE UNIQUE INDEX idx_timeline_dedup ON timeline_entries(page_id, date, summary, source);
```

## Suggested fixes

1. Key applied-migration tracking by stable **name**, not only the version integer — renumbering during merges then can't strand a migration as recorded-but-not-executed.
2. A doctor check that verifies migration **artifacts** (expected index/column/table per migration) rather than trusting the version counter — would have caught this immediately.
3. Defensively, `addTimelineEntry` could fall back to a plain insert-where-not-exists when the ON CONFLICT inference fails.

## Related observation

The `brain_score` SQL counts soft-deleted pages in every denominator (`SELECT count(*) FROM pages` with no `deleted_at IS NULL` filter), so a brain mid-purge-window is double-penalized on timeline/orphan/link-density components. Possibly intentional, but worth a look while in this code.

time-attack:
Closing as fixed on master (see 7c27fa12 / #2128). Please reopen if still reproducible on current HEAD.

Issue #895, opened 2026-05-12
Ranking results inverted - most relevant content gets lowest score


groobyming:
**What happened?**

When executing `gbrain query "Who is Zhang San"`, the ranking results are inverted:
- The least relevant result (concepts/memory-augmented-retrieval) appears at the top with score 1.0698
- The most relevant result (people/zhangsan) appears at the bottom with score 0.4922

```
$ gbrain query "Who is Zhang San"
[1.0698] concepts/memory-augmented-retrieval -- Memory Spa Method (stale)   ← Irrelevant
[1.0011] companies/goldfish-memory-tech -- Goldfish Memory Tech... (stale)
[0.9281] meetings/may-2026-meetup -- The symposium was held... (stale)
[0.5018] people/lisi -- Li Si, nicknamed "The Spender."... (stale)
[0.4922] people/zhangsan -- Zhang San, famously known as... (stale)   ← Most Relevant
```

**What did you expect?**

The most relevant result (people/zhangsan) should appear at the top with the highest score. The expected order should be:
1. people/zhangsan (highest score, most relevant)
2. people/lisi
3. companies/goldfish-memory-tech
4. meetings/may-2026-meetup
5. concepts/memory-augmented-retrieval (lowest score, least relevant)

**Steps to reproduce**
1. Import markdown files into GBrain brain (e.g., pages with people/zhangsan, people/lisi, companies/goldfish-memory-tech, concepts/memory-augmented-retrieval, meetings/may-2026-meetup)
2. Extract links: `gbrain extract links --source db`
3. Extract timeline: `gbrain extract timeline --source db`
4. Embed all: `gbrain embed --all`
5. Run query: `gbrain query "Who is Zhang San"`
6. Observe that people/zhangsan (the most relevant) is ranked last with lowest score

**Environment**
- gbrain version: gbrain 0.31.3
- OS: linux 
- Bun version: 1.3.13
- Database:  PGLite

**`gbrain doctor --json` output**
```json
{
    "schema_version": 2,
    "status": "warnings",
    "health_score": 85,
    "checks": [
        {
            "name": "resolver_health",
            "status": "warn",
            "message": "Could not find skills directory"
        },
        {
            "name": "connection",
            "status": "ok",
            "message": "Connected, 5 pages"
        },
        {
            "name": "pgvector",
            "status": "warn",
            "message": "Could not check pgvector extension"
        },
        {
            "name": "rls",
            "status": "ok",
[...]

PR #3122, opened 2026-07-21
fix(doctor,cli): NaN score prefix in query output + doctor check for put_page's missing ON CONFLICT arbiter (+#895 receipt)

### LINKED ISSUE #468: [NaN] score prefix in `gbrain query` output (v0.10.1, Postgres engine)
## Repro

Fresh install on macOS, `gbrain` v0.10.1, Postgres engine pointing at a Supabase project (Postgres 17.6, pgvector 0.8.0, pg_trgm 1.6).

Vault: 7 markdown pages, 10 chunks, all embedded with `text-embedding-3-large`.

```
$ gbrain query "folder conventions"
[NaN] claude -- # helm-vault — rulebook
[NaN] claude -- ## 1. Folder conventions
[NaN] index -- # helm-vault index
[NaN] log -- # log
[NaN] memory/memory -- # vault memory index
```

Every result line gets `[NaN]` as the score prefix. Ranking order itself looks correct (the `claude` chunks containing the matching content rank above unrelated pages), so this looks like a display / scoring-pipeline bug rather than a retrieval bug.

Reproduces with both literal-token and concept-style queries, and with `--no-expand`.

## Where it surfaces

`src/cli.ts:174`:

```ts
return results.map(r =>
  `[${r.score?.toFixed(4) || '?'}] ${r.slug} -- ${r.chunk_text?.slice(0, 100) || ''}${r.stale ? ' (stale)' : ''}`,
).join('\n') + '\n';
```

The formatter has an operator-precedence trap: when `r.score` is `NaN`, `NaN.toFixed(4)` returns the string `"NaN"`, which is truthy, so the `|| '?'` fallback never fires. Easy fix:

```ts
const s = Number.isFinite(r.score) ? r.score.toFixed(4) : '?';
return `[${s}] ${r.slug} -- ${r.chunk_text?.slice(0, 100) || ''}${r.stale ? ' (stale)' : ''}`;
```

That's the cosmetic patch. But the real question is **why the score is `NaN` to begin with** — that's almost certainly upstream in `src/core/search/` (RRF + multi-query expansion + dedup). Likely candidate: division by zero in the RRF normalizer when one of the source rankings is empty, or NaN propagation through the compiled-truth boost on a small corpus.

## Environment

- gbrain `0.10.1`
- bun `1.3.10`, node `22.22.0`
- macOS (Darwin arm64)
- Engine: Postgres on Supabase (transaction pooler, port 5432)
- 7 pages / 10 chunks / 100% embedded
- Schema version 4

## Doctor output

```
[OK] connection: Connected, 7 pages
[OK] pgvector: Extension installed
[OK] rls: RLS enabled on all tables
[OK] schema_version: Version 4 (latest: 4)
[OK] link_integrity: No dead links
```

Happy to test a fix against this Supabase install if useful.

### LINKED ISSUE #550: put_page fails with "no unique or exclusion constraint" — writes broken when pages_source_slug_key is missing, doctor doesn't catch it
**What happened?**

`gbrain put` and `mcp__gbrain__put_page` fail with a cryptic PostgreSQL error:

```
Error: there is no unique or exclusion constraint matching the ON CONFLICT specification
```

All write paths break — the CLI (`gbrain put <slug>`), the MCP server (`put_page` tool), and any cron/automation that writes pages. Reads continue to work normally, so the failure is silent until a write is attempted.

`gbrain doctor` does NOT catch this. The install scored 60/100 ("unhealthy") but gave no indication that writes were completely broken.

**What did you expect?**

`put_page` should succeed regardless of whether the `pages_source_slug_key` constraint was applied by schema.sql or by an external migration. Or at minimum, `gbrain doctor` should detect the missing constraint and flag it as a hard failure.

**Steps to reproduce**

1. Have a GBrain install connected to Supabase Postgres where the `pages` table was created or migrated by a process that uses a different unique constraint naming scheme (e.g., `UNIQUE(brain_id, slug)` instead of `UNIQUE(source_id, slug)`)
2. Run `echo '---\ntype: person\n---\n# Test\n' | gbrain put test-page`
3. Observe: `Error: there is no unique or exclusion constraint matching the ON CONFLICT specification`
4. Run `gbrain doctor --json` — writes are completely broken but doctor does not flag it

**Root cause**

`src/core/postgres-engine.ts` line 306 uses:
```sql
ON CONFLICT (source_id, slug) DO UPDATE SET ...
```

This requires a unique constraint on `(source_id, slug)`. The schema.sql (line 74) defines this as `CONSTRAINT pages_source_slug_key UNIQUE (source_id, slug)`. If that constraint was not applied during `gbrain init` — or was dropped and recreated with a different name by an external migration — the `ON CONFLICT` clause has nothing to match and PostgreSQL throws.

The same pattern exists in `src/core/pglite-engine.ts` line 258.

**Why doctor misses it**

`gbrain doctor` checks: connection, pgvector, RLS, schema version, embedding coverage, graph coverage, integrity samples, JSONB integrity, queue health. None of these verify that the unique constraints expected by the engine's upsert queries actually exist on the `pages` table.

**Fix suggestions**

1. **Short-term:** `gbrain doctor` should verify that `pages_source_slug_key UNIQUE (source_id, slug)` exists and flag its absence as a critical failure (not just a warning)
2. **Defensive:** `putPage` in `postgres-engine.ts` could fall back to a `SELECT`-then-`INSERT`/`UPDATE` pattern when the upsert fails, or run `CREATE UNIQUE INDEX IF NOT EXISTS` before the upsert
3. **Migration hardening:** `gbrain init` and `gbrain migrate` should assert the constraint exists after setup, not just trust that `CREATE TABLE IF NOT EXISTS` created it

**Environment**

- gbrain version: 0.22.6.1
- OS: macOS 15.3 (arm64)
- Bun version: 1.3.8
- Database: Supabase (managed Postgres + pgvector)

<details>
<summary>gbrain doctor --json output</summary>

```json
{"schema_version":2,"status":"unhealthy","health_score":60,"checks":[{"name":"resolver_health","status":"warn","message":"Could not find skills directory"},{"name":"connection","status":"ok","message":"Connected, 732 pages"},{"name":"pgvector","status":"ok","message":"Extension installed"},{"name":"rls","status":"fail","message":"3 table(s) WITHOUT Row Level Security: trigger_rules, trigger_fires, schema_migrations."},{"name":"schema_version","status":"ok","message":"Version 29 (latest: 29)"},{"name":"embeddings","status":"warn","message":"37% coverage, 586 missing."},{"name":"graph_coverage","status":"warn","message":"Entity link coverage 33%, timeline 93%."},{"name":"brain_score","status":"warn","message":"Brain score 66/100"},{"name":"integrity","status":"ok"},{"name":"jsonb_integrity","status":"ok"},{"name":"markdown_body_completeness","status":"ok"},{"name":"frontmatter_integrity","status":"ok"},{"name":"queue_health","status":"ok"}]}
```
</details>

time-attack:
Part 3 — Singleton P0s (backlog fix wave, unit part3-2).

## Fixes #468 — `[NaN]` score prefix in `gbrain query` output

`NaN.toFixed(4)` returns the truthy string `'NaN'`, so the formatter's `r.score?.toFixed(4) || '?'` fallback never fired and the CLI printed `[NaN] slug -- ...`. The formatter now gates on `Number.isFinite` and renders `[?]` for NaN/Infinity/missing scores. Test: `test/cli-format-query-score.test.ts`.

## Fixes #550 — silent write outage when the pages(source_id, slug) unique index is missing

`putPage` in both engines upserts via `ON CONFLICT (source_id, slug)`, which needs a non-partial unique index on exactly those columns as its arbiter. Migration v23 adds `pages_source_slug_key` by NAME only during migration runs — a brain stamped past v23 whose constraint was dropped/renamed by an external migration has every write failing with "no unique or exclusion constraint matching the ON CONFLICT specification", invisible to the version counter and to doctor.

New doctor check `pages_slug_unique_index` (wired into both the local `buildChecks` db phase and `doctorReportRemote`, mirroring the #2038 `timeline_dedup_index` shape check) probes `pg_indexes` for ANY conforming unique index — matched by columns, not name; partial indexes correctly rejected — and fails with a `gbrain apply-migrations --force-schema` hint. Test: `test/doctor-pages-slug-unique.test.ts` (includes proof that `putPage` really breaks in the detected state, and that a differently-named/ordered conforming index passes).

## #895 — ranking inversion: no longer reproduces; repro receipt pinned

Rebuilt the reported 5-page fixture (exact-name person page vs a backlink-magnet concept page, query "Who is Zhang San") on current master (PGLite). The inversion from v0.31.3 does not reproduce: the concept page gets its backlink boost (factor ~1.08, score 1.14) but title-match boost puts `people/zhangsan` first at 2.0, and all scores are finite. Pinned as a regression test so the inversion can't return: `test/ranking-exact-name-backlink-inversion.test.ts`. Referencing only — closing the issue is the maintainer's call with this receipt.

## Verification

- New test files + adjacent suites (`cli.test.ts`, `cli-format-volunteer.test.ts`, `doctor.test.ts`, `doctor-behavioral.test.ts`, `doctor-categories.test.ts`, `doctor-remote.serial.test.ts`, `doctor-report-remote.serial.test.ts`, `doctor-cli-smoke.serial.test.ts`) — all pass

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -888,7 +888,9 @@ export function formatResult(opName: string, result: unknown): string {
         return formatResultsExplain(results);
       }
       return results.map(r =>
-        `[${r.score?.toFixed(4) || '?'}] ${r.slug} -- ${r.chunk_text?.slice(0, 100) || ''}${r.stale ? ' (stale)' : ''}`,
+        // #468: NaN.toFixed(4) is the truthy string 'NaN', so `?.toFixed() || '?'`
+        // printed '[NaN]'. Gate on Number.isFinite instead.
+        `[${Number.isFinite(r.score) ? r.score.toFixed(4) : '?'}] ${r.slug} -- ${r.chunk_text?.slice(0, 100) || ''}${r.stale ? ' (stale)' : ''}`,
       ).join('\n') + '\n';
     }
     case 'get_tags': {
--- a/src/commands/doctor.ts
+++ b/src/commands/doctor.ts
@@ -528,6 +528,56 @@ export async function childTableOrphansCheck(engine: BrainEngine): Promise<Check
   };
 }
 
+/**
+ * #550 — pages(source_id, slug) unique-index presence.
+ *
+ * `putPage` in BOTH engines upserts via `ON CONFLICT (source_id, slug)`, which
+ * needs a non-partial unique index whose column set is exactly
+ * {source_id, slug} as its arbiter. Migration v23 adds `pages_source_slug_key`
+ * by NAME, so a brain whose constraint was dropped/renamed by an external
+ * migration is stamped past v23 with silently broken writes ("no unique or
+ * exclusion constraint matching the ON CONFLICT specification") that the
+ * version counter can't see. Match by COLUMNS, not name — any conforming
+ * unique index satisfies the arbiter (mirror of the #2038 timeline_dedup
+ * shape check).
+ */
+export async function checkPagesSlugUniqueIndex(engine: BrainEngine): Promise<Check> {
+  const name = 'pages_slug_unique_index';
+  try {
+    const rows = await engine.executeRaw<{ indexdef: string }>(
+      `SELECT indexdef FROM pg_indexes WHERE tablename = 'pages'`,
+    );
+    const ok = rows.some(r => {
+      const def = r.indexdef;
+      // Partial indexes can't arbitrate a bare ON CONFLICT (source_id, slug).
+      if (!/^CREATE UNIQUE INDEX/i.test(def) || /\bWHERE\b/i.test(def)) return false;
+      const open = def.lastIndexOf('(');
+      const close = def.lastIndexOf(')');
+      if (open < 0 || close < open) return false;
+      const cols = def
+        .slice(open + 1, close)
+        .split(',')
+        .map(c => c.trim().split(/\s+/)[0]) // drop DESC / opclass suffixes
+        .filter(Boolean)
+        .sort();
+      return cols.length === 2 && cols[0] === 'slug' && cols[1] === 'source_id';
+    });
+    if (ok) {
+      return { name, status: 'ok', message: 'pages has a unique index on (source_id, slug)' };
+    }
+    return {
+      name,
+      status: 'fail',
+      message:
+        'No unique index on pages(source_id, slug) — every put_page write fails with ' +
+        '"no unique or exclusion constraint matching the ON CONFLICT specification" (#550). ' +
+        'Run `gbrain apply-migrations --force-schema` to restore it.',
+    };
+  } catch {
+    return { name, status: 'warn', message: 'Could not check pages(source_id, slug) unique index' };
+  }
+}
+
 export async function doctorReportRemote(engine: BrainEngine): Promise<DoctorReport> {
   const checks: Check[] = [];
 
@@ -601,6 +651,9 @@ export async function doctorReportRemote(engine: BrainEngine): Promise<DoctorRep
   } catch {
     checks.push({ name: 'timeline_dedup_index', status: 'warn', message: 'Could not check idx_timeline_dedup shape' });
   }
+
+  // 2c. #550: pages(source_id, slug) unique index — putPage's ON CONFLICT arbiter.
+  checks.push(await checkPagesSlugUniqueIndex(engine));
 
   // v0.42.x — Life Chronicle (#2390): orphaned event projections. Reads already
   // hide projections whose event page is soft-deleted (read-time correctness);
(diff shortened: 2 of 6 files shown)


time-attack:
Merging. I verified the NaN path on master (`src/cli.ts:966`) — `NaN.toFixed(4)` is the truthy string `'NaN'`, so the `|| '?'` fallback genuinely never fires and users see `[NaN]`. The doctor check for a missing `pages(source_id, slug)` ON CONFLICT arbiter is squarely in the merge lane too: that failure mode is a total silent write outage invisible to the migration version counter, and matching by columns rather than index name is the correct probe. Appreciate the #895 no-repro receipt being pinned as a regression test instead of a speculative change.

Issue #2707, opened 2026-07-09
sources add accepts non-git dirs that can never sync; stale last_commit after re-init dies on raw 'git cat-file' fatal with no self-heal or reset CLI


amtagrwl:
## Bug (two related UX gaps around the git-repo requirement)

**(a) `sources add` accepts a non-git directory that can never sync.** Registering a plain directory succeeds; every subsequent sync then fails with:

```
Not a git repository: <local_path>. GBrain sync requires a git-initialized repo.
```

Nothing at registration time validates or hints at the requirement, so in an automated pipeline the source just goes silently stale (`last_sync_at` frozen) until an operator reads the sync logs. In our case an email-archive source sat unsyncable for 3 days while its markdown looked healthy from the earlier import.

**(b) After `git init` + commit in that directory, sync dies on a stale anchor with a raw git fatal.** `sources.last_commit` still held a SHA from the pre-git era, and `gbrain sync --source <id> --full` failed with:

```
fatal: git cat-file: could not get object info
```

No message names the stale anchor, and there is no CLI to reset it — we recovered by NULL-ing the `sources.last_commit` column in the database by hand, after which the full sync ran perfectly.

Worth noting: #1205 shows autopilot HAS a graceful path for this exact condition ("Sync anchor commit <sha> missing (force push?). Running full reimport."). CLI `sync --full` in this shape did not take that path — it surfaced the raw fatal instead.

**(c) Docs gap:** `git init` + an empty commit is NOT sufficient — the walker reads files through git objects, so untracked working-tree files are invisible. Files must actually be committed. This is discoverable only by experiment.

## Why this matters

The git requirement is reasonable, but its failure modes are all silent-or-cryptic: a source that never syncs, a fatal with no remediation hint, and a recovery that requires editing the database. Each one costs an operator a debugging session.

## Expected fix

- Validate (or at least warn) at `sources add` when `local_path` is not a git repo.
- When the anchor commit is missing from the repo, treat it as a first sync / full reimport (reusing the recovery that autopilot already has per #1205) instead of surfacing `git cat-file` fatals.
- Provide `gbrain sources reset-anchor <id>` (or document the supported way to clear `last_commit`).
- Document the "files must be committed, not just present" requirement.

Related: #1205 (anchor invalidation under autopilot), #753 (git-root discovery for subdir sources).

## Workaround

[...]

Issue #753, opened 2026-05-08
feat: sync should support subdir-of-git-repo as a source (atlas-style monorepo with N logical sources)


jeremyknows:
## Summary

For monorepo-style brains where a single git repo contains multiple logical sources (each in its own subdirectory), `gbrain sync` currently can't operate per-source — sync requires `local_path` to be a git root. This forces an architectural choice: either restructure the repo into N separate repos (operationally painful) OR fall back to `gbrain import` which is deliberately default-only (loses per-source isolation).

## Concrete repro / use case

I'm running gbrain on an "atlas" knowledge brain that has 1 git repo with 9 agent-specific subdirectories, each holding that agent's diary entries. Each agent should be its own source (per-source isolation for source-aware ranking). Plus a separate repo (`external-kb`) for VeeFriends KB content.

```bash
~/brain/                            # ← one git repo
├── agents/
│   ├── watson/memory/              # ← logical source: watson-diary
│   ├── librarian/memory/           # ← logical source: librarian-diary
│   ├── terminal/memory/            # ← logical source: terminal-diary
│   ├── augur/memory/               # ← logical source: augur-diary
│   └── ... (5 more)
└── shared/wiki/                    # ← logical source: wiki

~/projects/external-kb/           # ← separate git repo (works fine with sync today)
```

```bash
gbrain sources add watson-diary --path ~/brain/agents/watson/memory --no-federated
gbrain sync --source watson-diary --full
# Error: Not a git repository: ~/brain/agents/watson/memory.
#        GBrain sync requires a git-initialized repo.
```

`commands/sync.ts:374` enforces:
```js
if (!existsSync(join(repoPath, '.git'))) {
  throw new Error(`Not a git repository: ${repoPath}. GBrain sync requires a git-initialized repo.`);
}
```

For this case, the git repo is at `~/brain` (the parent), the source is at `~/brain/agents/watson/memory` (a subdir).

## Proposed fix

Add `--src-subpath <path>` flag (or rename `--repo` to `--root`, with `--source-path` semantics) so sync can operate on a subdirectory while reading git history from the repo root. Concretely:

```bash
gbrain sync --source watson-diary --root ~/atlas --src-subpath agents/watson/memory --full
```

Or, more elegant: when a source's `local_path` is registered, sync walks UP from `local_path` to find the nearest `.git` directory (longest-prefix git-root discovery), then uses THAT as the git context but only sync's files under `local_path`. No new flag needed; just smarter resolution.

[...]

PR #3123, opened 2026-07-21
feat(sources): persist + honor --include / --exclude globs across sync and lint (#2156)

### LINKED ISSUE #2156: sources: add --include / --exclude glob filters honored by sync AND lint

time-attack:
Fixes #2156
Takeover of #2157 (rebase + repair; original work by @brettdavies, credited via Co-authored-by — not closing that PR, just referencing)

## What

- `gbrain sources add --include <glob>... --exclude <glob>...` (repeatable) persist into `sources.config.include_globs` / `exclude_globs`; every subsequent `gbrain sync` of the source honors them.
- `gbrain sync --include <glob>` — allow-list counterpart to the existing #753/#774 `--exclude`. CLI flags merge (union) with the persisted config globs. Matched scope-relative in both the incremental and full-sync paths (`import.ts` gains the include filter). Exclusion stays conservative: never deletes previously-imported pages.
- `gbrain lint --include/--exclude`, plus auto-lift of the persisted source globs when the lint target matches a source's `local_path` — so a vault's `Templates/**` exclusion applies to lint without restating it.
- Migration **v125** (renumbered from the stale PR's v117/v120): `sources.config_fingerprint` caches a SHA-256 of the walk-affecting config fields (strategy + globs). Changing globs on an already-synced source now forces a full re-walk instead of "Already up to date" (git HEAD unchanged). NULL = never stamped; the first post-upgrade sync stamps quietly with no spurious re-walk.

## Repairs vs #2157

- Rebased onto current master; migration renumbered past master's v124; merged with the v0.42.52 `last_sync_at` heartbeat in the up-to-date gate and the #2707 `--force` flag on `sources add`.
- Globs are deliberately NOT threaded through `isSyncable`/`unsyncableReason` in the incremental path: the unsyncable-cleanup loop deletes pages for non-metafile classifications, which would have violated the documented conservative #1433 posture ("exclusion does NOT delete previously-imported pages"). Include instead mirrors master's existing scope-relative `excluded()` helper.
- CLI `--exclude`/`--include` merge with persisted config globs instead of being silently replaced by them (the PR's single-source path dropped CLI `--exclude`).
- Skipped the PR's whole-file reflow of `docs/guides/multi-source-brains.md`; added only the new filtering section.

## Tests

- `test/sync-source-globs.test.ts` — `parseGlobList` / `mergeGlobs` / `computeSourceConfigFingerprint` contracts.
- `test/sync-config-fingerprint.test.ts` — migration v125 column + persistence round-trip + drift detection on PGLite (bound via `$N::text::jsonb`, not stringify-into-`::jsonb`).
- `test/lint-source-globs.test.ts` — lint walker filter semantics (same matcher as sync).
- `test/sources.test.ts` — flag persistence, repeatability, no-`[]`-cruft guard.
- `test/sync-monorepo.test.ts` — new: `--include` on full + incremental sync, exclude-after-include precedence.
- `test/schema-bootstrap-coverage.test.ts` — `sources.config_fingerprint` exemption (same upgrade-path coverage as `chunker_version`).

`bun run typecheck` exit 0; 258 tests across the 7 touched files pass; JSONB guards (`check-jsonb-pattern.sh`, `check-jsonb-params.mjs`) clean; `bun run build:llms` regenerated (no bundle drift).

### DIFF
--- a/docs/guides/multi-source-brains.md
+++ b/docs/guides/multi-source-brains.md
@@ -115,6 +115,7 @@ Full subcommand reference:
 
 ```
 gbrain sources add <id> --path <p> [--name <n>] [--federated|--no-federated] [--force]
+                                  [--include <glob>...] [--exclude <glob>...]
                                Register a source. id: [a-z0-9](?:[a-z0-9-]{0,30}[a-z0-9])?
                                --path must be a git repo (or a subdirectory of one) — see
                                "The git requirement for --path sources" below. --force
@@ -130,6 +131,43 @@ gbrain sources detach          Remove .gbrain-source from CWD.
 gbrain sources federate <id>
 gbrain sources unfederate <id>
 ```
+
+## Filtering what gets synced (--include / --exclude)
+
+`--include` and `--exclude` on `gbrain sources add` accept repeatable glob
+patterns and are honored by every subsequent sync AND lint of the source.
+Common Obsidian vault setups need to exclude authoring scaffolding so it
+doesn't pollute search:
+
+```bash
+# Skip Templates/, Drafts/, and the smart-env sidecar; everything else syncs.
+gbrain sources add vault \
+  --path ~/Documents/vault --federated \
+  --exclude 'Templates/**' \
+  --exclude 'Drafts/**' \
+  --exclude '.smart-env/**'
+
+# Or: only sync the people/ and companies/ subtrees of a CRM vault.
+gbrain sources add crm \
+  --path ~/Documents/crm --no-federated \
+  --include 'people/**' \
+  --include 'companies/**'
+```
+
+Both persist into `sources.config.include_globs` / `exclude_globs` arrays.
+The filter runs `include` first, then `exclude`, so a path inside
+`people/**` is still rejected if it also matches `exclude_globs`. Globs use
+the same matcher as the rest of gbrain's sync classifier (`matchesAnyGlob`
+in `src/core/sync.ts`) and are matched against the source-root-relative
+path. Exclusion is conservative: it never deletes previously-imported pages.
+
+`gbrain sync --include <glob> --exclude <glob>` and
+`gbrain lint <dir> --include <glob> --exclude <glob>` take the same
+repeatable flags for one-off scope changes; for lint the persisted source
+globs are auto-applied when the lint target matches a source's `local_path`.
+Changing the persisted globs on an existing source triggers a full re-walk
+on the next sync (the source's config fingerprint invalidates the
+"already up to date" gate).
 
 ## The git requirement for --path sources
 
--- a/src/commands/import.ts
+++ b/src/commands/import.ts
@@ -54,6 +54,13 @@ export async function runImport(
     sourceId?: string;
     managedBookmark?: boolean;
     /**
+     * #2156: allow-list glob patterns — only dir-relative paths matching at
+     * least one pattern are imported. Applied BEFORE `exclude`. Threaded by
+     * performFullSync from `gbrain sync --include` / the source row's
+     * persisted `config.include_globs`.
+     */
+    include?: string[];
+    /**
      * #753/#774: glob patterns to exclude from the import (same semantics as
      * `isSyncable`'s `exclude` — matched against the dir-relative path).
      * Threaded by performFullSync for `gbrain sync --exclude`.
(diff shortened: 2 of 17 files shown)


time-attack:
Thanks for rebasing and repairing this — the notes on why globs deliberately don't thread through `isSyncable` (so the unsyncable-cleanup loop can't delete previously-imported pages) show you understood the #1433 posture properly. I'm still closing it: at ~1,500 lines with a new migration and a new `sources.config_fingerprint` column, this is a large feature rather than a fix, and master already carries the conservative `--exclude` path from #753/#774. Feature surface of this size is what we're deliberately not adding.

Issue #1641, opened 2026-05-30
Pooler→direct hostname swap in connection-manager.ts breaks IPv4-only networks during `gbrain init --url`


ayahya1983:
## Summary

`gbrain init --url <pooler-url>` against Supabase fails with `getaddrinfo ENOTFOUND` during the schema-migration step on IPv4-only networks, even when the caller explicitly provided a working Session pooler URL. The cause: `src/core/connection-manager.ts:147` derives a *direct* URL (`db.<project-ref>.supabase.co:5432`) from the pooler URL and uses it for the migration's direct-pool path. Supabase direct connections are **IPv6 only** unless the paid IPv4 add-on is enabled, so the lookup fails for every user behind an IPv4-only NAT.

The undocumented escape hatch `GBRAIN_DISABLE_DIRECT_POOL=1` (line 175) works around it, but it's not surfaced in `gbrain --help`, `gbrain init --help`, the README, or the warning emitted on direct-connection URLs. New users have to read source to find it.

## Environment

- macOS 26.5 (also reproduces on 26.3 per [#223](https://github.com/garrytan/gbrain/issues/223))
- Bun 1.3.13
- gbrain 0.41.29.0 (`bun install -g github:garrytan/gbrain`)
- Supabase project on `ap-northeast-2` (Seoul), free tier, no IPv4 add-on
- ISP is IPv4-only (\`bun --print 'await Bun.dns.lookup(\"db.<ref>.supabase.co\")'\` → \`DNSException: ENOTFOUND\`; same hostname's pooler resolves fine)

## Reproduction

1. Provision a fresh Supabase project on an IPv4-only network.
2. From the dashboard, switch the Connection-string panel to **Session pooler** and copy the URL (e.g. \`postgresql://postgres.<ref>:<pw>@aws-1-ap-northeast-2.pooler.supabase.com:5432/postgres\`).
3. Run:
   \`\`\`
   gbrain init --url \"<that exact URL>\" --force
   \`\`\`

### Expected

Init succeeds against the Session pooler. \`gbrain config show\` reports \`engine: postgres\`, \`database_url\` points at the pooler hostname, \`gbrain doctor\` is green.

### Actual

\`\`\`
Connecting to database...
pgvector extension not found. Attempting to create...
pgvector extension created successfully.
Running schema migration...
getaddrinfo ENOTFOUND
\`\`\`

[...]

PR #3125, opened 2026-07-21
fix(jobs/autopilot): worker liveness pre-flight, installer PATH, give-up exit codes, direct-pool config plane

### LINKED ISSUE #1868: autopilot --install: generated launcher assumes macOS/zsh and omits bun from PATH → systemd service crash-loops on Linux
## Summary
`gbrain autopilot --install` generates a launcher (`~/.gbrain/autopilot-run.sh`) that assumes **macOS/zsh** and never puts the `bun` bin dir on `PATH`. On Linux (incl. WSL) with a bun-linked gbrain install, the systemd **user** service then crash-loops: exec'ing the gbrain binary triggers its `#!/usr/bin/env bun` shebang, `bun` isn't on the service's `PATH`, and it dies instantly. systemd auto-restarts → infinite loop. **Net effect: autopilot never runs, so the brain never auto-cycles** (`cycle_freshness` stays perpetually stale), and it's easy to miss because manual `gbrain dream`/cron backstops mask it.

## Environment
- Linux (WSL2), bash login shell (no zsh, no `~/.zshenv`/`~/.zshrc`)
- gbrain installed via `bun link` (binary resolves to `~/.bun/bin/gbrain`, a `#!/usr/bin/env bun` script)
- gbrain `0.42.24.0`
- systemd **user** service `gbrain-autopilot.service`

## Symptom
- `systemctl --user status gbrain-autopilot.service` → `activating (auto-restart)`, never `active (running)`
- `~/.gbrain/autopilot.err` fills with (hundreds of KB):
  ```
  env: 'bun': No such file or directory
  env: use -[v]S to pass options in shebang lines
  ```
- `gbrain doctor` → `[FAIL] cycle_freshness → Source 'default' last cycled NNh ago`

## Root cause
The generated `~/.gbrain/autopilot-run.sh`:
```bash
#!/bin/bash
# Auto-generated by gbrain autopilot --install
# Sources shell profile for API keys, then runs autopilot.
# zshenv is the canonical place for env vars in zsh on macOS ...
[ -f ~/.zshenv ] && source ~/.zshenv 2>/dev/null
exec '/home/<user>/.bun/bin/gbrain' autopilot --repo '/home/<user>/brain'
```
On Linux: `~/.zshenv`/`~/.zshrc` don't exist; `~/.bashrc` (non-interactive) typically doesn't export `~/.bun/bin` onto `PATH`. systemd user services start with a minimal `PATH`. So when the kernel runs the `exec`'d gbrain script, its `#!/usr/bin/env bun` shebang can't find `bun` → `env: 'bun': No such file or directory`.

## Repro
1. Linux/WSL, bash. Install gbrain via `bun link`.
2. `gbrain autopilot --install`
3. `systemctl --user status gbrain-autopilot.service` → crash-looping; check `~/.gbrain/autopilot.err`.

## Suggested fix
In the generated launcher (and/or the systemd unit):
1. **Prepend the bun bin dir to `PATH`** — derive it from the resolved bun path (e.g. `dirname "$(command -v bun)"`) at install time, or `export PATH="$HOME/.bun/bin:$PATH"`.
2. **Don't hardcode zsh** — detect the shell/OS, or just source `~/.gbrain/keys.env` directly for API keys instead of relying on shell rc files (keys already live there).
3. Optionally invoke bun explicitly (`exec bun <gbrain-entry>`) so the shebang isn't load-bearing.

Manual workaround that resolved it: add `export PATH="$HOME/.bun/bin:$PATH"` (and `source ~/.gbrain/keys.env`) to `autopilot-run.sh`, then `systemctl --user restart gbrain-autopilot.service`.

## Related
- #135 (closed) — a different `autopilot --install` wrapper-generation bug (`process.execPath` returns bun). This one is distinct: macOS/zsh assumption + missing `bun` on `PATH` on Linux.

### LINKED ISSUE #2116: `onboard --auto` submits remediation jobs with no worker-liveness check and no inline fallback -- with jobsworkers down, every step burns its full timeout and aborts
### Version

gbrain 0.42.40.0 (clone @ `ecd6ae87`), Postgres engine, macOS, brain ~3.2k pages.

### Summary

`gbrain onboard --auto` executes its remediation plan by submitting each step to the job queue and polling for completion (`src/core/remediation/run.ts:271`: `waitForCompletion(queue, job.id, { pollMs, timeoutMs: (step.est_seconds + 60) * 1000 })`). It never checks whether a jobs worker is actually alive before submitting, and it has no inline-execution fallback. When the launchd jobsworkers are stopped (a deliberate quiesce during a maintenance window, or simply a machine where `gbrain jobsworker` was never installed), each step sits queued until its `est_seconds + 60` timeout expires and is marked aborted. From the operator's seat the command looks hung.

### Repro

1. Stop all jobs workers (e.g. `launchctl bootout gui/$UID/com.gbrain.jobsworker`), leaving the queue with no consumer.
2. Run `gbrain onboard --auto --max-usd 1`.
3. Watch the first remediation step submit, then poll silently until its timeout; subsequent steps repeat the pattern.

### Observed

Each step waits out its full per-step timeout with no progress and no diagnostic, then aborts. Observed live on 2026-06-11 during a pack-migration window with both jobsworkers deliberately booted out: the command appeared to hang, was killed by the operator, and the four planned actions were then run manually via `gbrain jobs submit <name> --follow` (the inline worker path), which executed them fine.

### Expected

Either of:

- A pre-flight worker-liveness check: if no worker heartbeat is visible, fail fast with "no jobs worker is running; start one or re-run with --inline", or
- An inline fallback (the machinery already exists; `jobs submit --follow` runs an inline worker in the caller's process).

Quiescing workers around maintenance is exactly when an operator runs `onboard` to verify state, so the combination is not exotic.

### Local mitigation

We run the planned remediation actions manually via `gbrain jobs submit <handler> --follow` whenever workers are quiesced, and treat `onboard --auto` as usable only while the launchd workers are up.

### Related

- #1306 is the PGLite-flavored cousin (dream synthesize hangs with no worker daemon); this one is the onboard/remediation path on Postgres, where workers exist but may be down.

### LINKED ISSUE #2234: autopilot: conversation_facts_backfill 'connect() has not been called' crash-loops, then clean-exits past Restart=on-failure
## Summary

On a multi-source Postgres brain (v0.42.37.0), the **`conversation_facts_backfill`** cycle phase throws `No database connection: connect() has not been called` for certain sources during the autopilot cycle. After repeated worker crashes the autopilot hits `max_crashes` and **exits with status 0**, so a systemd `Restart=on-failure` unit does **not** restart it — the daemon stays silently dead. In our case it went unnoticed for ~5 days (no syncs, no cycles, embed backlog re-accumulated).

Two distinct issues, the second is the more impactful:

### 1. `connect() has not been called` in `conversation_facts_backfill`

Same error class as the dream-cycle `connect()` bug fixed in **v0.42.21.0** (#1404/#1471/#1619), but it recurs in this phase's path. The phase is opt-in (`cycle.conversation_facts_backfill.enabled`, default false); when enabled it crashes the worker for some sources.

- Phase: `runPhaseConversationFactsBackfill` → `src/core/cycle/conversation-facts-backfill.ts` → `runExtractConversationFactsCore(engine, …)` (~line 214). The `engine` reaching the extract/write path appears not to be connected in the worker re-entry context — the same shape as the `--dir`-scoped extract phases the v0.42.21.0 module-singleton-ownership fix addressed.

### 2. Autopilot clean-exits after `max_crashes`, defeating `Restart=on-failure`

When a phase crash-loops to `max_crashes`, the autopilot daemon gives up and **exits 0**. systemd treats exit 0 as success, so `Restart=on-failure` never fires and the daemon stays down indefinitely with no signal. 

**Suggestion:** exit **non-zero** on the give-up path so `Restart=on-failure` recovers it (and/or document that operators should use `Restart=always`). A daemon that has decided it can't make progress should signal failure, not success.

## Environment

- gbrain 0.42.37.0, Postgres (Supabase pooler: transaction port 6543 for the read pool + session pool for the direct/DDL pool)
- Multi-source brain, `gbrain autopilot` under systemd `--user`

## Workarounds in place

- `Restart=always` on the autopilot unit (recovers regardless of exit code)
- Disabled the opt-in `conversation_facts_backfill` phase (returns to default)

Happy to provide more detail / test a fix.

### LINKED ISSUE #2731: GBRAIN_DISABLE_DIRECT_POOL is env-only (no config.json plane): daemons/headless jobs silently fall back to the failing direct pool
## Summary

The Supabase direct-pool kill-switch is read **only** from `process.env`:

```ts
// src/core/connection-manager.ts:175-176
return process.env.GBRAIN_DISABLE_DIRECT_POOL === '1' ||
  process.env.GBRAIN_DISABLE_DIRECT_POOL === 'true';
```

There is no `config.json` field for it. So any process that does not inherit the
env var (a `launchd`/`systemd` daemon, a cron job, a headless `claude -p`
subprocess) silently falls back to the direct connection, which then fails with
`ENOTFOUND`/`ECONNREFUSED` on Supabase Session-Pooler brains (the
`deriveDirectUrl` hostname-swap bug already tracked in #1641 / #1006).

The failure is **silent**: the daemon just never writes, and an interactive
shell "works" because the flag happens to be exported there, which masks the
root cause.

## Repro (real-world)

1. Brain on Supabase Postgres via the **session pooler** (host
   `aws-*.pooler.supabase.com`, port 6543).
2. Install `gbrain autopilot` as a `launchd` job (LaunchAgent plist, no `env`
   block; the wrapper sources no profile that exports the flag).
3. Every cycle fails: `deriveDirectUrl` synthesizes `db.<ref>.supabase.co:5432`,
   which does not resolve on an IPv4-only / no-IPv6 network, so `getaddrinfo
   ENOTFOUND`. Nothing is written; no actionable error surfaces to the operator.
4. Interactive `gbrain <cmd>` in a normal shell works fine (the flag is exported
   in the shell), so the bug looks environment-specific and is hard to spot.
5. Fix that unblocked it: manually add `export GBRAIN_DISABLE_DIRECT_POOL=1` to
   `~/.zshenv` so the non-interactive daemon subprocess inherits it.

## Why it matters

Env-only flags are invisible to exactly the processes that most need the
kill-switch: long-running background writers (autopilot, sync, cron ingest) and
headless agent jobs. Requiring every wrapper/plist to plumb the env var by hand
is fragile and fails silently when missed.

This is the same "no working config path" shape already reported for API keys in
#2119 / #2120: the pooler flag has the identical file-plane vs env gap.

## Proposed fix

Let the kill-switch also be read from `config.json` (file plane), for example
`disable_direct_pool: true`, with precedence env > config > default (mirrors the
existing `GBRAIN_DIRECT_DATABASE_URL` override). Then:

```
gbrain config set disable_direct_pool true   # file plane; persists for ALL processes
```

would make daemons and headless jobs pick it up with no shell-profile plumbing.
(Per #2119, ensure `config set` for this key writes the plane the runtime
actually reads.)

## Related

- #1641, #1006: the direct-pool hostname bug this flag works around
- #2119, #2120: config-plane inconsistency for keys (same root shape)

## Environment

- gbrain 0.42.58.0
- macOS (Darwin 25), IPv4-only network
- Supabase Postgres, session pooler (port 6543)

time-attack:
Backlog fix wave, unit c36-1 (jobs/autopilot workers, locks & installers).

## Fixes

Fixes #1868 — `autopilot --install`'s generated launcher assumed macOS/zsh and omitted bun from PATH, so a bun-shebang `gbrain` crash-looped under systemd's minimal PATH. The wrapper now bakes the install-time runtime dirs (gbrain's own dir, the running runtime's dir, `~/.bun/bin`) onto PATH *after* profile sourcing so they win even if a profile resets PATH. New pure `generateWrapperScript()` with tests (dedup, empty-safe — never emits a cwd-injecting empty PATH entry, quote-escaping).

Fixes #2116 — `onboard --auto` / `doctor --remediate` / MCP `run_onboard` submitted remediation jobs with no worker-liveness check: with workers quiesced, every step burned its full `(est_seconds + 60)` timeout and aborted silently. `runRemediation` now runs a pre-flight on the Postgres path (PGLite is inline and exempt): live if the local worker registry has an entry OR the DB shows an active job holding a live lock (busy remote worker). No signal → new `onNoWorker` hook + `no_worker` result field; CLI shells print a start-a-worker hint and exit 1. `GBRAIN_REMEDIATION_ASSUME_WORKER=1` covers the idle-remote-worker topology; probe failures fail open so a broken probe never blocks remediation.

Fixes #2234 — (part 2, the operationally impactful half per the reporter) autopilot's give-up paths (`max_crashes`, `cycle-failure-cap`) exited 0, so `Restart=on-failure` units never restarted a daemon that had decided it can't make progress — it stayed silently dead for days. `shutdown()` now exits via `shutdownExitCode(reason)`: 1 for give-up reasons, 0 for SIGTERM/SIGINT. Part 1 (the `connect() has not been called` in `conversation_facts_backfill` for some sources) is NOT addressed here: per-source errors in that phase are already caught and downgraded to `warn`, and no reproducible engine-ownership defect could be pinned without a repro — a speculative change there would be worse than none. Left open on the issue.

Fixes #2731 — `GBRAIN_DISABLE_DIRECT_POOL` was env-only, so daemons/launchd/cron that don't inherit the shell env silently fell back to the failing direct pool. New file-plane config key `pool.disable_direct` (file plane because this resolves before any DB connect), read by `resolveKillSwitch()` when the env var is unset. Env wins in both directions when set (`=0` overrides a config-plane `true`) — same env-above-config precedence as pace-mode. The deeper `deriveDirectUrl` hostname-swap root cause stays tracked in #1641/#1006.

## Tests

- `test/autopilot-install.test.ts` — wrapper PATH baking (order vs profile sourcing, dedup, empty-safe, quoting) + `shutdownExitCode` mapping
- `test/remediation-worker-liveness.test.ts` — new: registry-empty+no-locks → false, live-lock proxy → true, registered local worker → true, escape hatch, fail-open
- `test/connection-manager.serial.test.ts` — `resolveKillSwitch` config plane + env-wins-both-directions

`bun run typecheck` exit 0; all touched test files pass (72 + 13 e2e onboard flow + 34 adjacent config/worker-registry + 25 remediation/onboard).


time-attack:
Closing this one. The #1868 installer-PATH fix already landed on master and in a stronger form — `writeWrapperScript` now bakes in the dir derived from `process.execPath` (with a guard against `dirname('')` prepending cwd) and keeps `~/.bun/bin` only as a fallback, which covers Homebrew/nix/Docker installs that a hardcoded path misses. Merging this would pull a competing older implementation back in through a conflicting diff. The other three items are bundled together and each adds new hook/env/config surface, so they'd need to be separate, minimal PRs rather than a four-issue wave.

PR #3126, opened 2026-07-21
fix(gateway): stdio takes allow-list override, recipe-pricing drift guard, calendar lookahead, take-proposal review CLI

### LINKED ISSUE #1467: Take proposal workflow: configurable provider, multiple proposals, review CLI
## Feature request

The take-proposal pipeline would be easier to operate if it supported a complete review workflow rather than a single implicit proposal path. A downstream patch carries three related changes:

- make `propose-takes` provider/model configurable
- allow multiple take proposals per page in schema/code
- add a `takes` review CLI for inspecting and acting on proposals

## Motivation

Different deployments may want cheaper or provider-specific models for proposal generation, and pages can contain multiple independent candidate takes. Operators also need a CLI review surface before accepting/declining generated proposals.

## Expected behavior

- Proposal generation provider/model can be configured separately from default model paths.
- Multiple proposals can be stored per page without overwriting each other.
- CLI review commands expose pending proposals and allow operator review.

## Current downstream patch

Downstream changes touch `src/core/cycle.ts`, `src/core/cycle/propose-takes.ts`, schema files, `src/commands/takes.ts`, and `test/propose-takes.test.ts`.

### LINKED ISSUE #2657: stdio MCP: takesHoldersAllowList hardcoded to ['world'] — no operator override, so non-world takes are unreachable via MCP (v0.42.8, still on master)
`src/mcp/server.ts` hardcodes the takes visibility allow-list for the stdio transport:

```ts
return dispatchToolCall(engine, name, params, {
  remote: true,
  takesHoldersAllowList: ['world'],
  ...
```

There is no env var, config key, or CLI flag to widen it. Any take whose `holder` is not exactly `world` (e.g. `self`, or a named holder) is invisible to **every** MCP caller on stdio — `takes_list`, `takes_search`, and takes surfaced via `query`. The code comment suggests `gbrain call <op>` as the operator escape hatch, which works, but agents connected over stdio MCP (Claude Code/Desktop, the primary integration) can never see their operator's own private takes even though the pipe is local and single-user.

### Repro (v0.42.8.0, Postgres engine)

1. A takes row with `holder='self'`, `active=true`, on a live page (written via `mcp:put_page` fence).
2. `gbrain call takes_list '{}'` → returns the row (local trusted path, allow-list unset). ✅
3. Same brain over stdio MCP (`gbrain serve`): `takes_list {}` → `[]`; `takes_list {"holder":"self"}` → `[]`; `takes_search` → `[]`. ❌
4. Verified the row passes every SQL filter in `listTakes` (active, live page join); the only excluding predicate is `holder = ANY(['world'])`.

### Why it matters

For a single-user brain, `holder='self'` is the natural holder for the operator's own claims — the v0.28 default-deny (sensible for multi-tenant HTTP tokens) makes them permanently unreachable on the transport that local agents actually use. Downstream effect: an agent pipeline that checks "do I have a prior take on X?" via `takes_search` always gets an empty set and concludes NO PRIOR, which is worse than an error because it reads as a confident negative.

### Suggested fix

Any of:
- `GBRAIN_MCP_TAKES_HOLDERS=self,world` env override read by `src/mcp/server.ts` (stdio only, keep the `['world']` default);
- honor `permissions.takes_holders` from a config key for stdio like the http path does;
- a `serve --takes-holders` flag.

Related: #2529 covers the http-transport flavor (legacy bearer tokens ignore `permissions.takes_holders`); this issue is the stdio transport, where there is no token to carry permissions at all. The CLI-verb siblings (`gbrain takes list` printing "No takes on list" while `gbrain call takes_list` returns rows, and `gbrain call takes_search` dying on `JSON.stringify cannot serialize BigInt`) look like #2450.

### LINKED ISSUE #2689: Anthropic chat allowlist predates Claude 5: claude-sonnet-5 accepted by config but chat client build fails silently → 'no LLM available' empty synthesis
## Environment
gbrain 0.42.53 (`github:garrytan/gbrain#814258d`), PGLite, macOS 26.5.1, `anthropic_api_key` set in `~/.gbrain/config.json` (verified working).

## Repro
1. `gbrain config set models.think anthropic:claude-sonnet-5` — accepted, no warning. The model id resolves fine at config time.
2. `gbrain think "anything"` — synthesis comes back as "no LLM available" (and per #1698 the empty result is saved), with no error mentioning the model id.
3. `gbrain config set models.think anthropic:claude-sonnet-4-6` — everything works with the identical key/config.

So the registry/config layer and the chat-client build disagree about which Anthropic models exist: the id passes validation where it's cheap to reject, then fails where the failure is silent. Current Anthropic model ids that fail this way include `claude-sonnet-5` (and presumably the other Claude 5 family ids); `claude-sonnet-4-6` and other 4.x ids work.

## Asks
1. Refresh the Anthropic allowlist for the Claude 5 family — or better, stop allowlisting chat models for providers whose APIs already return a clean "unknown model" error, and surface that error instead (same stale-allowlist class as #2613/#2507/#1607 for Google).
2. When the chat client can't be built for the configured model, fail loudly at `config set` time or at `think` time with the model id in the message — the current path (accept at config, degrade to "no LLM available" at runtime) cost us a debugging session that ended in bisecting model ids. #1698 covers the saving-empty-synthesis half of this.

### LINKED ISSUE #2790: feat(recipes): calendar-to-brain lookahead window — recipe promises "knows who you're meeting tomorrow" but every prescribed window ends at today

The calendar-to-brain recipe's headline promise is meeting **prep**: "Your agent knows who you're meeting tomorrow, what you discussed last time, and what context matters" (`recipes/calendar-to-brain.md:37-39`), "When you have a meeting tomorrow, the agent pulls attendee dossiers automatically" (`:48`).

As shipped, that promise is structurally unfulfillable — every window the recipe prescribes ends at *today*:

- Backfill: `--start 2020-01-01 --end $(date +%Y-%m-%d)` (`:224`)
- Weekly cron: `--start $(date -v-7d +%Y-%m-%d) --end $(date +%Y-%m-%d)` (`:266-267`)
- No lookahead flag exists anywhere in the recipe or its prescribed collector design (repo-wide: no `--days-ahead`/lookahead in any calendar context).

Tomorrow's events are never fetched, so they can never be ingested — and the read side that WOULD surface them ("Upcoming within lookahead window", `src/core/context-engine.ts:436`) has nothing to read. The recipe's weekly-sync + daily-briefing cron pairing (`docs/guides/cron-schedule.md`) compounds it: even day-of events are up to a week stale.

Add a lookahead parameter to the recipe spec (e.g. `--days-ahead 7`, `timeMax = today + N`), so day files cover upcoming events; note in the cron guidance that the lookahead is what feeds the daily briefing. One parameter closes the gap between the recipe's promise and its mechanics.

**Alternatives considered**

Local collector fetches a lookahead window (deployed — the daily briefing cites tomorrow's meetings from it). Every operator who reads the recipe's promise will have to rediscover and rebuild this.

---

Environment: gbrain 0.42.57.0 (verified unchanged at 0.42.58.0 / a25209bb) · macOS 26.3.1 · Bun 1.3.13 · Supabase Postgres (PgBouncer transaction mode).

time-attack:
Backlog work unit c35-2-f1: provider recipes & model-id/allowlist rot (gateway component).

## Fixes #2657 — stdio MCP takes-holder allow-list operator override
`src/mcp/server.ts` hardcoded `takesHoldersAllowList: ['world']` with no escape hatch, so non-world takes were unreachable for stdio MCP agents. New `GBRAIN_MCP_TAKES_HOLDERS` env (comma-separated, trimmed) widens it — same env-escape-hatch pattern as `GBRAIN_SOURCE` two lines below. Fail-closed: unset / empty / whitespace-only values keep the `['world']` default; an operator can only opt IN to widening. Pinned by `test/mcp-stdio-takes-holders.test.ts`.

## Fixes #2689 — recipe allowlist ↔ pricing drift guard (rest already on master)
Both concrete halves of the issue landed on master since it was filed: the anthropic recipe already allowlists the Claude 5 family (`claude-sonnet-5`, `claude-fable-5`, `claude-opus-4-8`), and a config-set `models.think` that fails the probe no longer degrades silently — it's labeled `MODEL_NOT_USABLE:<reason>` with the model id in the answer/gaps (pinned by the "labels an unusable CONFIGURED model honestly" test in `test/think-pipeline.serial.test.ts`; that pinned graceful-degrade contract is deliberate, so this PR does not convert it to a hard throw). What was still missing is the guard the issue asked for against the rot class recurring: `test/recipe-pricing-drift.test.ts` now pins every model id the anthropic recipe accepts (chat + expansion + alias targets) to a `CANONICAL_PRICING` entry, so allowlist and pricing table can't silently diverge again.

## Fixes #2790 — calendar-to-brain lookahead window
The recipe promised "knows who you're meeting tomorrow" but prescribed backfill/cron windows ending at *today*, so upcoming events never reached the brain and the context engine's lookahead window had nothing to read. Docs-only fix: backfill and weekly-cron `--end` move to `$(date -v+7d +%Y-%m-%d)` (Linux variant noted), with a line explaining why the window runs ahead; `docs/guides/cron-schedule.md` example updated to match, noting the lookahead feeds the daily briefing's meeting prep.

## Fixes #1467 — take-proposal review CLI + configurable model
The `take_proposals` queue and multi-proposal support already landed, but the review path that `propose-takes.ts`'s own doc comment promises ("User accepts/rejects via `gbrain takes propose`") didn't exist — proposals were written and never actionable. This PR adds:
- `src/core/takes-proposals.ts` — `listTakeProposals` (source-scoped, holder allow-list fail-closed like `takes_list`) + `resolveTakeProposal` (accept promotes into the page's canonical takes fence under the per-page lock, mirrors to DB via `addTakesBatch`, records `promoted_row_num`; reject keeps the row as audit history; double-resolve fails loudly).
- Contract-first ops `takes_proposals_list` (read) and `takes_proposal_resolve` (write, `localOnly` — accept writes repo markdown, so remote/MCP callers are blocked fail-closed).
- CLI: `gbrain takes propose` lists pending; `--accept <id>` / `--reject <id>` resolve; `--status/--page/--limit/--json/--dir` supported.
- Configurable provider: the propose_takes phase model now resolves explicit `opts.model` > `models.propose_takes` config (alias-aware via `resolveModel`) > gateway default, and the *resolved* id reaches the extractor and the `model_id` audit column (previously only an explicit override did).
Pinned by `test/takes-proposals-review.test.ts` (real PGLite + temp brain dir) and a new model-routing case in `test/propose-takes.test.ts`.

## Verification
- New/touched tests: `test/mcp-stdio-takes-holders.test.ts`, `test/recipe-pricing-drift.test.ts`, `test/takes-proposals-review.test.ts`, `test/propose-takes.test.ts` — all pass
- Contract suites re-run green: mcp-tool-defs, operations-allow-list, operations-descriptions, operations-trust-boundary, trust-boundary-contract, operation-context-sourceid-required, schema-cli-contract, takes-command-source-scope, takes-mcp-allowlist
- CI guards: check-key-files-current-state, check-jsonb-pattern, check-jsonb-params — clean; `bun run build:llms` regenerated (llms-full.txt committed)


Masashi-Ono0611:
Independent confirmation from a downstream fork ([Masashi-Ono0611/gbrain](https://github.com/Masashi-Ono0611/gbrain)): hit this exact issue (#2657) today. Generated ~1176 takes via the extraction pipeline (default holder `system`/`brain`), but every MCP-bound `takes_list` / `think` call returned empty — the hardcoded `world`-only allow-list meant they were structurally invisible via the primary interface, even though `doctor` confirmed the rows existed. Took a while to root-cause since the CLI and `doctor` paths aren't holder-restricted, only the stdio MCP path is.

Backported this PR's `resolveStdioTakesHolders()` change locally (the `GBRAIN_MCP_TAKES_HOLDERS` opt-in is a nice touch — keeps the fail-closed default while giving operators an escape hatch) and confirmed it builds clean and `doctor` stays healthy. This closed the gap for us. Thanks for the fix.

Masashi-Ono0611:
Follow-up now that this is actually serving traffic on our brain, rather than just building.

**End-to-end verification through a live stdio MCP.** My earlier comment only confirmed the backport built and `doctor` stayed healthy — it did not exercise the MCP path. Driving a fresh `gbrain serve` over stdio with a JSON-RPC `tools/call` for `takes_list`:

| `GBRAIN_MCP_TAKES_HOLDERS` | result |
|---|---|
| `world,system,brain` | takes returned |
| unset (upstream default) | empty |

One trap worth flagging for anyone verifying this: our serve wrapper exports the variable itself, so passing a different value from outside is overwritten and **both arms look identical**. My first control run "passed" for the wrong reason. Drive `gbrain serve` directly to test it.

**Correction to my earlier comment.** I wrote that the takes were "default holder `system`/`brain`". Paginating all 1176 rows, **every one is `holder=system`; there are zero `brain` rows.** That sharpens the case for this PR rather than weakening it — on this brain the `world`-only default was hiding 100% of takes, not a private subset.

**One design observation.** The allow-list has no way to express "everything". The storage layer already supports it — `postgres-engine.ts` filters with `allowList IS NULL OR t.holder = ANY(allowList)`, so NULL means unfiltered — but `resolveStdioTakesHolders()` returns `['world']` for empty input and has no path that yields `undefined`. So an operator who wants all takes has to enumerate holders, and any holder introduced later (a `people/<slug>`, or `brain` once dream starts emitting takes) is silently filtered out again. That is the same failure mode as #2657, just deferred to whenever the holder set next changes — and it is quiet, which is what made the original hard to root-cause.

Would you take a follow-up mapping an explicit sentinel (`*` or `all`) to `undefined`? Happy to send it. The fail-closed posture would be preserved — the sentinel has to be typed deliberately, never the empty-string default. Not filing a separate issue since this is a refinement of #2657 and this PR is where the code lives.

time-attack:
Thanks for the thorough write-up, and for being straight that both concrete halves of #2689 already landed on master. Closing this: the bulk of the 648 lines is the new take-proposal review CLI with new entries in `operations.ts`, which is feature surface we're not taking, and it's bundled with a docs-only calendar change and a test-only drift guard. The `GBRAIN_MCP_TAKES_HOLDERS` override against the hardcoded `['world']` allow-list is the one piece with clear standalone merit, but it can't carry the other 600 lines with it.

Masashi-Ono0611:
That matches what I found independently on my fork (#2657) — appreciate you calling out the `GBRAIN_MCP_TAKES_HOLDERS` override as the one piece with standalone merit. I'll cut that alone into its own PR against master and drop the rest.

Masashi-Ono0611:
Sent the `GBRAIN_MCP_TAKES_HOLDERS` piece as its own PR: #3472. Everything else from this PR (take-proposal review CLI, calendar lookahead docs, pricing-drift guard) stays out of it.

Issue #2782, opened 2026-07-13
bug(cycle): patterns phase reports status:ok with child_outcome:timeout and writes nothing when no subagent-capable worker slot is free — silent no-op for days; doctor's fan-out check can't see cycle-spawned subagent capacity


alexputici:
**What happened?**

Deployment context for this batch of reports: see #2778 (same production operation, environment, and receipts discipline).

In the stock topology, the autopilot's embedded child worker is the only subagent-capable worker and runs at **concurrency 1**: `jobs work` defaults to `'1'` (`src/commands/jobs.ts:83` — `parseFlag(args, '--concurrency') ?? env.GBRAIN_WORKER_CONCURRENCY ?? '1'`), and autopilot spawns it with no `--concurrency` flag (`autopilot.ts:396`). A running autopilot-cycle therefore occupies the sole slot — **the cycle's own patterns subagent can never be claimed.**

The phase then hides the failure. In `src/core/cycle/patterns.ts`, a child `TimeoutError` sets `outcome = 'timeout'` without throwing (lines 99-102), and the phase returns `ok(...)` — which hardcodes `status: 'ok'` (patterns.ts:324-325) — with `child_outcome: 'timeout'`, `patterns_written: 0`. Every "completed" cycle's patterns phase reported ok while producing nothing, for days. Verified on jobs #23745/#23747; live-reproduced: subagent job #24262 sat WAITING through the entirety of cycle #24261.

Doctor cannot see it either: `computeAutopilotFanoutConcurrencyCheck` (`doctor.ts:1604-1637`) compares the autopilot per-source cycle fan-out against worker concurrency, reserving one slot (`effectiveSlots = max(1, concurrency - 1)`), but does not model that each RUNNING cycle additionally needs a free slot for its patterns subagent. At concurrency 1 with fan-out 1: `1 > max(1, 0)` is false → reports ok — the exact starvation case. (And when no supervisor audit event exists, the check skips entirely.)

**What did you expect?**

1. A subagent-awaiting phase should **fail loudly or run inline** when no free subagent-capable worker slot exists — never report `status: ok` on a timed-out child that wrote nothing. Silent no-op phases are the worst outage class: everything looks green while a whole judgment layer is dead.
2. The doctor fan-out check should account for **cycle-spawned subagent capacity** — including the autopilot's own embedded worker concurrency — not just cycle fan-out with a single reserved slot.

**Steps to reproduce**

[...]

PR #3127, opened 2026-07-21
fix(jobs/autopilot): recycled-PID lock takeover, extract inference takeover of #2408, worker-concurrency docs

### LINKED ISSUE #2503: autopilot lock can false-positive when stale PID is reused by unrelated process
## Summary

`gbrain status` / autopilot liveness can report autopilot as running when the lock file contains a stale PID that has been reused by an unrelated process.

In that state, the launchd/system supervisor can keep trying to start autopilot, but each new autopilot process exits with "Another autopilot instance is running" because the stale lock PID is alive. Maintenance then silently stops even though the status surface says autopilot is running.

## Observed environment

- `gbrain 0.42.52.0`
- macOS LaunchAgent install
- Postgres-backed brain
- `~/.gbrain/autopilot.lock` lockfile

## Observed failure mode

1. `gbrain status --json` reported:
   - `autopilot.installed: true`
   - `autopilot.lockfile_present: true`
   - `autopilot.running: true`
   - `autopilot.pid: 973`
2. `ps -p 973 -o command=` showed that PID `973` was not GBrain at all. It belonged to Apple `AirPlayUIAgent`.
3. `~/.gbrain/autopilot.err` repeated:
   - `Another autopilot instance is running (lock file holder is alive). Exiting.`
4. `launchctl print ...com.gbrain.autopilot` showed the service stuck in respawn/spawn-scheduled behavior, not a healthy long-running autopilot process.
5. No fresh `autopilot-cycle` jobs were being created; cycle freshness drifted for days.
6. Moving the stale lock aside and restarting the LaunchAgent immediately restored real autopilot operation.

## Expected behavior

Autopilot lock validation should not treat "PID exists" as sufficient proof that the lock holder is the active GBrain autopilot.

At minimum, status/startup should verify one or more of:

- process command/executable matches GBrain autopilot;
- PID start time is not older/newer than the lock acquisition metadata in an impossible way;
- lock includes a nonce/session/boot identity that a new process can validate;
- the supervising service state agrees with the pidfile;
- stale/unowned locks are broken automatically or surfaced as a hard warning.

## Actual behavior

A reused PID from an unrelated process was enough to make new autopilot processes exit and make `gbrain status` report autopilot as running.

## Impact

This is a bad silent-failure mode for a maintenance daemon:

- sync/capture may continue through other paths, so the system looks partly alive;
- embeddings/facts/cycles/backlinks can stop accumulating;
- `status` gives the operator false confidence;
- `doctor` eventually reports stale cycles, but the root cause is not obvious unless the operator manually checks the lock PID owner.

## Suggested fix

Harden the autopilot lock check and status liveness check so they validate process identity, not just PID existence. If the PID is alive but not a GBrain autopilot process, treat the lock as stale and either break it safely or report an explicit stale-lock/foreign-process diagnostic.

### LINKED ISSUE #2796: docs: worker-concurrency guidance is three-way inconsistent (guides say 4, snippets say 2, real defaults are 1/2); doctor cadence weekly vs daily across guides
**What happened?**

Two documentation inconsistencies a fresh operator cannot resolve without reading source (both bitten here during a from-scratch deployment-design exercise):

**1. Worker concurrency — three different answers, none of them the default.**
- Prose guides recommend `--concurrency 4`: `docs/guides/minions-deployment.md:33,69,72` and `docs/guides/queue-operations-runbook.md:101`.
- All three copy-paste deployment snippets use `--concurrency 2`: `docs/guides/minions-deployment-snippets/systemd.service:21`, `fly.toml.partial:13`, `Procfile:13`.
- The actual code defaults are **neither**: `jobs work` defaults to **1** (`src/commands/jobs.ts:83`) and `jobs supervisor` defaults to **2** (`jobs.ts:1179`).

This matters more than a typical docs nit because concurrency is load-bearing for the cycle-subagent starvation class (#2782): at the real `jobs work` default of 1, an autopilot cycle's own patterns subagent can never be claimed. An operator following any of the three documented values gets three different behaviors, and one following none of them (the actual default) gets the silent no-op.

**2. Doctor cadence — weekly vs daily.**
- `docs/guides/cron-schedule.md:26,44-45`: doctor is a **weekly** task (Mondays 6 AM cron).
- `docs/guides/operational-disciplines.md:54-56,109`: "DISCIPLINE 4: Daily Heartbeat Check" runs `gbrain doctor` **daily** ("the daily heartbeat catches these before they compound").

**What did you expect?**

State the actual defaults in exactly one place (per command — work vs supervisor differ), link every guide and snippet to it, reconcile the recommended values with a scale note (when 1 is fine, when 2/4 are needed and what they cost in DB connections), and pick one doctor cadence story (or state explicitly that heartbeat=daily, deep-doctor=weekly, if that's the intent).

**Environment**
- gbrain version: 0.42.57.0 (docs verified at 0.42.58.0 / a25209bb)

---

Related: #2782 (why the concurrency default deserves accurate documentation).

time-attack:
Backlog fix unit c36-2-f2 (jobs/autopilot workers, locks & installers).

## Items

- **Takeover of #2408** — subject-scope link-type inference + reserved-slug fuzzy guard (original author: @JiraiyaETH, credited via Co-authored-by). Salvages the PR intact and repairs the named blocker: `LINK_EXTRACTOR_VERSION_TS` is bumped to `2026-07-21T00:00:00Z` so previously-stamped pages re-extract under the new inference, and `test/extract-stale.test.ts` now derives its "recent" fixture date from the constant so future bumps can't silently break it. References #2408 only — nothing is closed here.
- **Fixes #2503** — autopilot lock false-positive on recycled PIDs. `decideLockAcquisition` now checks the holder's command line (`ps -p <pid> -o command=`, fail-open when `ps` can't judge) and takes over foreign holders with an explicit `foreign process holds stale lock` reason. `gbrain status` mirrors the check: `running=false` + `foreign_lock_holder: true` + a human-readable stale-lock hint instead of reporting a non-gbrain PID as a running autopilot.
- **Fixes #2796** — worker-concurrency guidance was three-way inconsistent (guides 4, snippets 2, real defaults 1/2). The deployment guide now states both real defaults once (`jobs work`=1, `jobs supervisor`=2) with a scale note (concurrency 1 starves cycle subagents per #2782; higher values cost DB connections), and all guide examples + CLI help align to the snippets' `--concurrency 2` baseline. Doctor cadence reconciled explicitly as heartbeat=daily / deep pass=weekly with cross-references. llms bundles regenerated.

## Tests

- `test/autopilot-lock.test.ts` — new: foreign-holder takeover, `pidLooksLikeGbrain` fail-open on dead PID + foreign detection on a live `sleep` child.
- `test/extract.test.ts`, `test/link-extraction.test.ts` — the #2408 suite (canonical FS-path inference, subject-scope guard, reserved-slug guard).
- Gate: `bun run typecheck` exit 0; touched test files (autopilot-lock, extract, link-extraction, extract-stale, build-llms, serve-stdio-lifecycle, status-sections, get-status-snapshot-op, jobs-nice-flag) all pass.

No VERSION/package.json/CHANGELOG changes — release bookkeeping happens at ship time.

### DIFF
--- a/docs/guides/cron-schedule.md
+++ b/docs/guides/cron-schedule.md
@@ -41,7 +41,9 @@ fixed. You wake up and the brain is smarter than when you went to sleep.
 # Calendar sync — Sundays at 10 AM
 0 10 * * 0 cd /path/to/calendar-sync && node calendar-sync.mjs --start $(date -v-7d +%Y-%m-%d) --end $(date +%Y-%m-%d)
 
-# Brain health — weekly Mondays at 6 AM
+# Brain health — weekly DEEP pass, Mondays at 6 AM (doctor + embed backfill).
+# This complements — does not replace — the DAILY `gbrain doctor` heartbeat
+# from operational-disciplines.md Discipline 4: heartbeat=daily, deep pass=weekly.
 0 6 * * 1 gbrain doctor --json >> /tmp/gbrain-health.log 2>&1 && gbrain embed --stale
 
 # Dream cycle — nightly at 2 AM
--- a/docs/guides/minions-deployment.md
+++ b/docs/guides/minions-deployment.md
@@ -26,11 +26,25 @@ with exponential backoff (1s → 60s cap), emits lifecycle events to an
 audit file, and drains gracefully on SIGTERM (35s worker-drain window
 before SIGKILL). Exit codes are documented so agents can branch on them.
 
+**Concurrency defaults (single source of truth):**
+
+| Command | Default | Flag / env |
+|---|---|---|
+| `gbrain jobs work` | **1** | `--concurrency N` / `GBRAIN_WORKER_CONCURRENCY` |
+| `gbrain jobs supervisor` | **2** | `--concurrency N` (passed through to the worker) |
+
+Every snippet in this guide (systemd, Fly, Procfile) pins `--concurrency 2`
+explicitly — the recommended baseline. Why not 1: at concurrency 1 a job that
+submits and waits on its own subagent job (e.g. an autopilot cycle) can never
+have that subagent claimed — the single slot is already occupied (#2782). Why
+not more: each slot holds a DB connection and can fan out an LLM subagent, so
+scale to 4 only when throughput demands it and your Postgres pool has headroom.
+
 **Typical commands:**
 
 ```bash
 # Start in the foreground (blocks; Ctrl-C to stop).
-gbrain jobs supervisor --concurrency 4
+gbrain jobs supervisor --concurrency 2
 
 # Start detached — returns {"event":"started","supervisor_pid":…} on stdout.
 gbrain jobs supervisor start --detach --json
@@ -66,10 +80,10 @@ isn't:
 ```bash
 # Full concurrency, low priority. Propagates to the spawned worker and its
 # children (shell jobs, subagents) via OS niceness inheritance.
-gbrain jobs supervisor --concurrency 4 --nice 10
+gbrain jobs supervisor --concurrency 2 --nice 10
 
 # Equivalent for a bare worker, or set it durably in the environment.
-GBRAIN_NICE=10 gbrain jobs work --concurrency 4
+GBRAIN_NICE=10 gbrain jobs work --concurrency 2
 ```
 
 `--nice` takes a POSIX value from `-20` (highest priority) to `19`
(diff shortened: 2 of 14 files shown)


time-attack:
Thank you for this, and the verification quality here is genuinely high — I confirmed the recycled-PID lock gap is still on master and the subject-scope attribution bug in #2408 is a real data-correctness problem. I'm closing on scope rather than substance: this bundles three unrelated p3 items, and the largest piece bumps `LINK_EXTRACTOR_VERSION_TS`, which forces a full re-extraction of every page in every existing brain on upgrade. A change with that kind of cost needs to stand alone with its own review and release note, not ride alongside a lock fix and a docs refresh.

PR #3128, opened 2026-07-21
fix(webhook): extract links for incremental push syncs

### LINKED ISSUE #2849: GitHub webhook sync can leave link extraction permanently stale
**What happened?**

`POST /webhooks/github` submits a `sync` Minion job without an explicit `noExtract` value. The standalone sync handler defaults missing `noExtract` to `true`:

```ts
const noExtract = job.data.noExtract !== false;
```

The webhook job therefore imports the pushed commit and advances the source bookmark without materializing link/timeline edges. On the following autopilot source cycle, sync is already `up_to_date`, so `pagesAffected` is empty and the incremental extract phase has no slugs to process. The affected page can remain stale indefinitely unless an operator runs extraction manually or another independent stale sweep happens to select it.

This was observed on a private installation; all names, identifiers, and content below are synthetic. A sanitized job sequence was:

```text
webhook sync: status=synced, pagesAffected=["notes/example"]
next source cycle sync: status=up_to_date, pagesAffected=[]
next source cycle extract: pages_processed=0
```

Current `master` (v0.42.59.0, commit `5008b287e47bf791132eedfebf66bdef11e9398c`) still has the same queue payload and defaulting behavior.

**What did you expect?**

A successful GitHub push webhook should leave link/timeline extraction at least as fresh as the imported source commit. For the current sync contract, the minimal fix is for the webhook caller to opt in explicitly with `noExtract: false`. That preserves the standalone handler's default for callers that intentionally submit a separate extract job.

For commits above the inline extraction threshold, the durable follow-up should be a source-scoped stale extraction job or equivalent native sweep tied to the consumed commit.

**Steps to reproduce**

1. Register a GitHub-backed federated source and configure `/webhooks/github`.
2. Add or edit a Markdown page so it introduces a new wikilink, then push the tracked branch.
3. Let the webhook-submitted `sync` job complete. Observe that the source bookmark advances while the page's graph/backlink extraction remains stale.
4. Let the next normal autopilot source cycle run. Sync reports no affected pages, incremental extraction processes zero pages, and the new backlink is still absent.
5. Run `gbrain extract --stale --source-id <source-id>` manually; the backlink appears, confirming ingestion succeeded and extraction was the missing step.

**Environment**
- Observed version: v0.42.57.0
- Confirmed affected source: v0.42.59.0 / current `master`
- OS: Linux x86_64 (behavior is platform-independent)
- Bun version: not material to this queue-data defaulting path

Not included because it can contain installation-specific metadata and the failure is isolated to the webhook queue payload plus sync-handler default. The sanitized job sequence and current source locations above reproduce the control-flow defect without private data.

time-attack:
Fixes #2849
Takeover of #2850 (salvages its noExtract:false opt-in + tests; credit to @patentsong via Co-authored-by)

## Problem

`POST /webhooks/github` and `gbrain sync trigger` submit `sync` minion jobs without `noExtract`, and the standalone sync handler defaults missing `noExtract` to `true`. A webhook push therefore imports the commit and advances the source bookmark **without** link/timeline extraction. The next autopilot cycle sees `up_to_date` (empty `pagesAffected`), and its extract phase only walks that cycle's changed slugs — so the pushed pages stay extraction-stale indefinitely until a manual `gbrain extract --stale`.

Even with the opt-in, syncs of >100 files hit the #1794 large-sync deferral, which skips inline extraction and (pre-fix) relied on a stale sweep that no standalone job path ever ran.

## Fix

- **serve-http.ts / sync.ts (`runSyncTrigger`)** — webhook + trigger payloads pass `noExtract: false`, opting normal incremental pushes into sync's inline extraction (from #2850).
- **sync.ts** — `performSync` surfaces `extractDeferred: true` when the >100-file branch skips inline extraction (only set when extraction was requested).
- **jobs.ts (sync handler)** — on `extractDeferred`, queues a source-scoped `extract` job with `{ stale: true }` (best-effort, idempotency-keyed, `maxWaiting: 1`; a duplicate sweep finds 0 stale and no-ops).
- **jobs.ts (extract handler) / extract.ts** — extract handler gains a `stale: true` mode backed by the same (now-exported) `extractStaleFromDB` core as `gbrain extract --stale`; DB-source, so it works on checkout-less brains.

## Tests

- `test/sync-large-deferral-extract.serial.test.ts` (new): real git repo, 101-file incremental sync through the actual sync job handler — asserts the deferral flag, the queued `extract{stale:true}` follow-up, and that running the follow-up recovers links + `links_extracted_at` watermark end-to-end. Fails without the fix (verified), passes with it. Also pins that sub-threshold syncs do NOT set the flag.
- `test/sources-webhook.test.ts` — webhook payload contract test (`noExtract: false` in the queued sync job), salvaged from #2850.
- `test/sync-trigger-cli.test.ts` — asserts the trigger job payload carries `noExtract: false`.

Gate: `bun run typecheck` exit 0; touched test files (6 files, 47 tests) all pass.

### DIFF
--- a/src/commands/extract.ts
+++ b/src/commands/extract.ts
@@ -1651,7 +1651,7 @@ async function extractTimelineFromDB(
  * make re-extraction idempotent). EVERY processed page is stamped, including
  * zero-link pages — they WERE processed.
  */
-async function extractStaleFromDB(
+export async function extractStaleFromDB(
   engine: BrainEngine,
   opts: {
     dryRun: boolean;
--- a/src/commands/jobs.ts
+++ b/src/commands/jobs.ts
@@ -1479,7 +1479,31 @@ export async function registerBuiltinHandlers(
       embedSkipReason = 'auto_embed_disabled';
     }
 
-    return { ...result, embed_job_id: embedJobId, embed_skip_reason: embedSkipReason };
+    // #2849: large-sync extract deferral follow-up. performSync skips inline
+    // link/timeline extraction when totalChanges > 100, leaving
+    // links_extracted_at unstamped. A standalone sync job (webhook push,
+    // sync trigger) has no autopilot extract phase behind it, so the pages
+    // would stay extraction-stale until a manual `gbrain extract --stale`.
+    // Queue a source-scoped stale sweep instead. Best-effort + idempotent:
+    // a duplicate sweep finds 0 stale pages and no-ops.
+    let extractJobId: number | null = null;
+    if (result.extractDeferred) {
+      try {
+        const { MinionQueue } = await import('../core/minions/queue.ts');
+        const queue = new MinionQueue(engine);
+        const followUp = await queue.add(
+          'extract',
+          { stale: true, ...(sourceId ? { sourceId } : {}) },
+          {
+            idempotency_key: `sync-extract-stale:${sourceId ?? 'default'}:${Math.floor(Date.now() / 30_000)}`,
+            maxWaiting: 1,
+          },
+        );
+        extractJobId = followUp.id;
+      } catch { /* best-effort: extract --stale sweeps it later */ }
+    }
+
+    return { ...result, embed_job_id: embedJobId, embed_skip_reason: embedSkipReason, extract_stale_job_id: extractJobId };
   });
 
   registerBuiltinJob(worker, engine, 'embed', async (job) => {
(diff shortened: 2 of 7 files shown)


time-attack:
Closing as already fixed — the `noExtract: false` change for both the webhook and trigger paths landed on master in d15e2ab8 (#3337), together with the same webhook and trigger contract tests carried here, so the #2849 gap is closed. The remaining piece is the follow-up `extract{stale:true}` job for the >100-file deferral branch, but master already routes that case to the resumable `extract --stale` watermark sweep in the autopilot cycle, which scans the whole source rather than just the current run's slugs. Thanks for the thorough writeup, and credit to @patentsong for the original.

PR #3130, opened 2026-07-21
fix(embed): per-request batch caps for Google/DashScope + parallel embed dispatch + import auto-workers

### LINKED ISSUE #970: Add max_batch_tokens for gemini-embedding-001 recipe
## Summary

The Google Gemini embedding recipe declares `gemini-embedding-001` but does not set `max_batch_tokens`, so startup/runtime warnings can complain that the embedding batch cap is missing.

## Evidence

Google's Gemini Embedding model page documents the `gemini-embedding-001` input token limit as 2,048 tokens:

https://ai.google.dev/gemini-api/docs/models/gemini-embedding-001

## Expected

The recipe should encode the documented cap so the gateway can pre-split batches and avoid the missing-`max_batch_tokens` warning.

## Local patch tested

I added:

```ts
max_batch_tokens: 2048,
```

to `src/core/ai/recipes/google.ts` under the embedding touchpoint.

Verification:

```bash
bun test test/recipes-contract.test.ts
# pass
```

### LINKED ISSUE #1199: dashscope: text-embedding-v3 rejects batches > 10 (`InvalidParameter`) — recipe has no `max_batch_count` field
## Symptom

When `gbrain import` (or any path that calls `embedBatch` with > 10 texts in one call) hits DashScope `text-embedding-v3`, the API returns:

```
[embed(dashscope:text-embedding-v3)] <400> InternalError.Algo.InvalidParameter:
  Value error, batch size is invalid, it should not be larger than 10.: input.contents
```

Individual files that produce > 10 chunks are silently skipped. On my 43-file `gbrain import`, **7 files (the long ones) failed for this reason** while shorter files imported fine.

## Root cause

DashScope's OpenAI-compatible endpoint enforces a per-request limit of **10 `input.contents`** (it's a hard count cap, not just a token cap).

`src/core/ai/recipes/dashscope.ts` only declares the token cap:

```ts
embedding: {
  ...
  max_batch_tokens: 8192,
  chars_per_token: 2,
}
```

There's no `max_batch_count` field, and grepping the gateway confirms it has no general per-recipe content-count cap mechanism:

```bash
$ grep -rn "max_batch_count\|maxBatchCount" src/core/ai/
# no results
```

Meanwhile `src/core/embedding.ts` (v0.36.3.0) ships with:

```ts
const BATCH_SIZE = 100;
```

So `embedBatch` sub-batches to 100 texts before handing off to the gateway, and the gateway has no way to know it should further split to 10 for DashScope.

## Reproduce

- v0.36.3.0 (also reproduces on v0.35.x — recipe unchanged)
- DashScope `text-embedding-v3` via OpenAI-compatible recipe
- `provider_base_urls.dashscope: https://dashscope.aliyuncs.com/compatible-mode/v1` (mainland-China key path)
- `embedding_dimensions: 1024`
- Import a directory where at least one file chunks into > 10 pieces (long markdown dumps, PDF-derived pages, etc.)

Excerpt from real `gbrain import`:

```
Warning: skipped daily/2026-05-18-2318.md: [embed(dashscope:text-embedding-v3)] <400> InternalError.Algo.InvalidParameter: Value error, batch size is invalid, it should not be larger than 10.: input.contents
Warning: skipped daily/2026-05-17.md:    (same error)
Warning: skipped daily/2026-05-17-1626.md: (same error)
Warning: skipped daily/2026-05-17-1625.md: (same error)
Warning: skipped daily/2026-03-23.md:    (same error)
[import.files] 43/43 (100%) done
  35 pages imported
  8 pages skipped (0 unchanged, 8 errors)
```

## Suggested fix

Two minimal changes:

1. **`src/core/ai/recipes/dashscope.ts`** — add `max_batch_count: 10`:
   ```ts
   embedding: {
     ...
     max_batch_tokens: 8192,
     max_batch_count: 10,  // DashScope OpenAI-compat API caps at 10 inputs/call
     chars_per_token: 2,
   }
   ```

2. **`src/core/ai/gateway.ts`** (or wherever sub-batching lives) — read `recipe.touchpoints.embedding.max_batch_count` and cap each outbound request to `min(currentBatch, maxBatchCount)`.

`embedding.ts` `BATCH_SIZE = 100` can stay as the outer paginator for progress callbacks — the actual cap belongs at the recipe / gateway layer where it's per-provider.

Worth checking other recipes too:
- Zhipu has its own per-call limits
- MiniMax is 32
- Voyage has limits at the high end too

## Workaround (what I'm running)

Local patch to `src/core/embedding.ts`:

```diff
-const BATCH_SIZE = 100;
+const BATCH_SIZE = 10;
```

Re-importing the failed 7 files completes cleanly. Blunt (slows OpenAI/Voyage paths to 10/batch too), but reliable until the per-recipe cap lands.

## Context

Running on a mainland-China JD Cloud Ubuntu ECS, brain is PGLite, embed is DashScope (no API egress to international endpoints), chat is Bedrock through a local LiteLLM proxy. Same environment as #1065 (which v0.36.3.0 resolved — thanks for that). Happy to send a PR for the recipe + gateway change if useful, but figured the issue first so you can decide whether `max_batch_count` is the right field name / where the gateway split should live.

### LINKED ISSUE #1207: `gbrain import` makes per-file OpenAI embedding calls (~2s/file); batching would give 10-50× speedup
## Observation

`gbrain import` against ~2000 staged markdown files runs at ~2.1 seconds per file. After 35 minutes the wall-clock had embedded 1000 of 1989 files:

```
~/.gbrain/import-checkpoint.json:
{
  "totalFiles": 1989,
  "processedIndex": 1000,
  "completedFiles": 1000,
  "timestamp": "2026-05-19T19:30:05.008Z"   ← 35 min after start
}
```

2100s ÷ 1000 files = **2.1 s/file**.

That number is unmistakable: it's a single OpenAI `text-embedding-3-large` (or similar) API round-trip per file. CPU work on the file itself is microseconds. The throughput is bottlenecked entirely on serial calls to OpenAI's `/v1/embeddings` endpoint.

## What gstack expected

gstack's `bin/gstack-gbrain-sync.ts:1063` cites an estimate:

> "Per ED2: ~25-35 min for ~11.7K transcripts = ~150ms/page synchronous"

150ms/page would be plausible if embedding calls were batched (per-batch round-trip cost amortized over many files). 2100ms/page is what you see when each file is a single API call.

## The fix

OpenAI's `/v1/embeddings` accepts **up to 2048 inputs per request** and a max payload of ~8M tokens. Batching the embedding step at the chunk level (or even at the file level, for short files) yields the kind of speedup that lets a one-time `gbrain import` of a personal archive finish in minutes instead of hours:

| Batch size | Embedding calls for 1989 files | Time @ ~2s/call |
|---|---:|---:|
| 1 (current) | 1989 | 35 min+ |
| 32 | 63 | ~2 min |
| 256 | 8 | ~16 s |

(Rough; assumes one chunk per file for the table — adjust for multi-chunk files.)

## Why this matters now

The personal-brain use case has gotten an order of magnitude bigger in the last month — e.g. one user (me) just landed ~166k Gmail markdown files in `~/brain/email/` after wiring the iCloud + Gmail LaunchAgents. At 2.1s/file, that's a **97-hour** initial embed at sequential rate. Multi-day. With batching at size 256, the same archive embeds in ~22 minutes.

The 35-minute SIGTERM wall in gstack (filed separately at garrytan/gstack#1611) is partially a downstream symptom of this — if gbrain batched, the wall would stop biting big-brain users.

## Where to look

I haven't poked at gbrain's embedder pipeline directly, but a quick grep for the OpenAI client call site in `src/embedding/` or wherever the per-chunk loop lives would be the place. Two paths:

1. **Tactical:** wrap the per-call site in a batcher that collects N items then flushes (back-pressure-safe via a bounded queue).
2. **Structural:** rewrite the import stream to fan inputs into a batched embedder with a configurable batch size (env: `GBRAIN_EMBEDDING_BATCH_SIZE`, default 256).

Either way, the same code path needs to respect OpenAI's token budget per request (~8M; conservative default 1M is plenty).

## Environment

- macOS 15.x, Apple Silicon
- gbrain v0.33.1.0
- gstack v1.40.0.0 (invoking `gbrain import` via `gstack-memory-ingest --bulk`)
- Engine: Supabase Session Pooler (postgres)
- OPENAI_API_KEY present in env; OpenAI API responding normally

## Verifying the diagnosis

If anyone wants to confirm this is single-call (not some other pathology), run with `OPENAI_LOG=debug` or wrap the embedder call in a logger that prints per-batch input count. A "1 input per call" log == this issue.

Happy to PR the batcher if it's a welcome change.

### LINKED ISSUE #1818: embedBatch() dispatches batches serially — give it bounded parallelism (the embed-stale.ts concurrency pool the cycle/extract path doesn't use)
**Type:** Performance / feature
**Version observed:** gbrain 0.42.10.0
**Severity:** Medium — the routine cycle/extract embed path runs embedding batches **serially**, so multi-batch embeds (long pages, fact extraction, contextual reindex) are gated on sequential API round-trips even though `embed-stale.ts` already proves bounded parallelism works elsewhere.

## TL;DR

`embedBatch()` in `src/core/embedding.ts` chunks `texts` into `BATCH_SIZE` slices and `await`s each slice **one at a time** (a serial `for` loop). For any caller with more than one batch worth of inputs, total time = sum of per-batch API latencies, when it could be `~max` of them with bounded concurrency. The infra to do this right already exists in `src/core/embed-stale.ts` (a `concurrency` worker-pool, default 20) — it's just not used by the general `embedBatch()` path that the cycle/extract callers go through. Request: give `embedBatch()` bounded parallelism (shared with / modeled on `embed-stale.ts`), configurable, default conservative.

## What's serial today

`src/core/embedding.ts` (~line 91):

```ts
export async function embedBatch(texts, options = {}) {
  if (!texts?.length) return [];
  // fast path: single batch
  if (texts.length <= BATCH_SIZE && !options.onBatchComplete) {
    return gatewayEmbed(texts, gwOpts);
  }
  const results = [];
  for (let i = 0; i < texts.length; i += BATCH_SIZE) {     // <-- serial
    const slice = texts.slice(i, i + BATCH_SIZE);
    const out = await gatewayEmbed(slice, gwOpts);          // <-- one round-trip at a time
    results.push(...out);
    options.onBatchComplete?.(results.length, texts.length);
  }
  return results;
}
```

Hot callers that hit the multi-batch path:
- `src/core/cycle/extract-facts.ts:238` — `const embeddings = await embed(texts)` in the routine cycle.
- contextual-retrieval / reindex paths that embed many chunks per page.

## Why this matters (measured context)

#1207 already documented the macro symptom: serial per-file embedding makes `gbrain import` run at ~2.1 s/file because each file is one API round-trip. That issue targets `import` batching. **This issue is the complementary, narrower gap:** even once you batch, `embedBatch()` dispatches the batches *sequentially*. On a provider that accepts 2048 inputs/call but where each call is ~300–2000 ms, a page/extract job that produces, say, 6 batches pays 6× the latency instead of ~1–2× with a concurrency of 3–4.

This is **not** "add batching" (that exists) and **not** the `embed-stale.ts` path (that's already concurrent). It's specifically: the general-purpose `embedBatch()` that the cycle/extract/contextual paths call is the one place still serial.

## Proposed implementation

1. **Bounded-parallel slice dispatch in `embedBatch()`.** Replace the serial `for` with a worker-pool of size `concurrency` (mirror `embed-stale.ts`'s `numWorkers = Math.min(concurrency, keys.length)` pattern). Preserve input order in `results` (index slices, don't rely on completion order). Keep `onBatchComplete` firing as each slice resolves.
2. **Config knob.** `EmbedBatchOptions.concurrency` (default e.g. 4 — conservative to respect provider rate limits), overridable via env `GBRAIN_EMBED_CONCURRENCY`. Single-batch fast path unchanged.
3. **Rate-limit / 429 safety.** Reuse the existing retry/backoff in `gatewayEmbed`; ensure the pool honors `abortSignal` at the same three sites `embed-stale.ts` documents (batch claim, retry sleep, HTTP body). On repeated 429s, the pool should back off effective concurrency rather than hammer.
4. **Share the primitive.** Ideally factor the bounded-concurrency map out of `embed-stale.ts` into a small shared helper so both paths use one implementation (DRY; one place to get backpressure right).

## Acceptance

- A multi-batch `embedBatch()` call with concurrency N completes in ~`ceil(batches/N)` API-latency units instead of `batches` units, order-preserving.
- `extract-facts` / contextual reindex throughput improves proportionally on multi-chunk pages, with no increase in 429 error rate at the default concurrency.
- `abortSignal` cancels in-flight and queued slices promptly.
- Single-batch callers see byte-identical behavior (fast path untouched).

## Related

- #1207 — `gbrain import` per-file embedding batching (the macro case; complementary, not duplicate).
- #1199 / #970 / #1080 — provider per-batch size limits; the concurrency pool must respect these existing `max_batch_*` recipe fields when sizing slices.
- `src/core/embed-stale.ts` — existing bounded-concurrency implementation to model on / extract from.

---
_Filed 2026-06-03 alongside the connection-pooler investigation. The DB round-trip is a minority of sync wall-clock; embedding parallelism is the higher-leverage speedup for embedding-bound jobs._

time-attack:
Backlog work unit c68 — four adversarially-verified embedding fixes.

- Fixes #970 — the `google` recipe now declares `max_batch_tokens` + `max_batch_count` (+ `chars_per_token`), so the gateway pre-splits and the missing-cap startup warning stops firing. The token budget is derived from Gemini's documented per-request limits (100 inputs × 2048 tokens/input), NOT the 2048 per-INPUT limit the issue proposed verbatim (that would over-split ~50×).
- Fixes #1199 — new optional `EmbeddingTouchpoint.max_batch_count`, enforced in `splitByTokenBudget` (a sub-batch flushes at N inputs even when the token budget still has room). `dashscope` sets 10 (the provider hard-caps embedding batches at 10 inputs per request). `isTokenLimitError` also matches DashScope's "batch size is invalid" message, so the recursive-halving safety net backstops any path that misses the pre-split.
- Fixes #1207 — `gbrain import` without `--workers` no longer hardcodes serial; the default routes through the shared `autoConcurrency` policy (PGLite → 1, >100 files on Postgres → 4). Explicit `--workers N` still wins, and the existing PGLite/no-database_url serial guard is untouched.
- Fixes #1818 — `embedBatch()` dispatches its 100-input sub-batches through a bounded worker pool (default 4, overridable via `EmbedBatchOptions.concurrency` or `GBRAIN_EMBED_BATCH_CONCURRENCY`; `concurrency: 1` restores serial). Results are index-addressed so output order always matches input order; the single-small-batch fast path is byte-identical; `abortSignal` still rejects the whole call.

Supporting change: `listRecipes()` now reads the exported `RECIPES` map instead of the private `ALL` array — one registry source of truth, and it lets tests inject a synthetic capless recipe so the startup-warning path stays covered now that every real recipe declares a cap.

Tests: extended `test/ai/adaptive-embed-batch.test.ts` (count-cap split semantics, DashScope error pattern, embed()-level ≤10/≤100 dispatch assertions, reworked D9-B warning test), `test/ai/recipe-dashscope.test.ts`, `test/ai/no-batch-cap-suppression.serial.test.ts`; new `test/embed-batch-concurrency.test.ts` (order preservation, bounded parallelism, env override, monotonic progress, failure propagation, fast path) and `test/import-default-workers.test.ts` (auto policy vs explicit flag). `bun run typecheck` exit 0; 559 tests across the touched + adjacent suites pass.

### DIFF
--- a/src/commands/import.ts
+++ b/src/commands/import.ts
@@ -170,10 +170,14 @@ export async function runImport(
   // v0.22.13 (PR #490 Q2): shared parseWorkers helper rejects bad input
   // (--workers 0, -3, "foo") with a loud error instead of silently falling
   // through to 1. Mirrors sync.ts's flag handling.
-  const { parseWorkers } = await import('../core/sync-concurrency.ts');
-  let workerCount: number;
+  const { parseWorkers, autoConcurrency } = await import('../core/sync-concurrency.ts');
+  // #1207: undefined (no --workers flag) defers to autoConcurrency below —
+  // the shared sync/import policy (PGLite → 1, >100 files → 4) — instead of
+  // hardcoding serial. Large Postgres imports stop paying one embedding
+  // round-trip per file in sequence.
+  let workerCount: number | undefined;
   try {
-    workerCount = parseWorkers(workersArg ?? undefined) ?? 1;
+    workerCount = parseWorkers(workersArg ?? undefined);
   } catch (e) {
     console.error(e instanceof Error ? e.message : String(e));
     process.exit(1);
@@ -252,8 +256,9 @@ export async function runImport(
   }
   const files = resumeFilter(allFiles, dir, completed);
 
-  // Determine actual worker count
-  const actualWorkers = workerCount > 1 ? workerCount : 1;
+  // Determine actual worker count. Explicit --workers wins; otherwise the
+  // shared autoConcurrency policy decides from engine kind + file count.
+  const actualWorkers = autoConcurrency(engine, files.length, workerCount);
   if (actualWorkers > 1) {
     console.log(`Using ${actualWorkers} parallel workers`);
   }
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -1513,12 +1513,21 @@ export async function embed(texts: string[], opts?: EmbedOpts): Promise<Float32A
 
   const embedding = recipe.touchpoints?.embedding;
   const maxBatchTokens = embedding?.max_batch_tokens;
+  const maxBatchCount = embedding?.max_batch_count;
   const charsPerToken = embedding?.chars_per_token ?? DEFAULT_CHARS_PER_TOKEN;
 
-  // Pre-split is gated on max_batch_tokens. Recipes without it (e.g. OpenAI)
-  // ride the fast path: one embedMany call, no recursion safety net.
-  const batches = maxBatchTokens
-    ? splitByTokenBudget(truncated, Math.floor(maxBatchTokens * effectiveSafetyFactor(recipe)), charsPerToken)
+  // Pre-split is gated on max_batch_tokens / max_batch_count. Recipes with
+  // neither (e.g. OpenAI) ride the fast path: one embedMany call, no
+  // recursion safety net.
+  const batches = (maxBatchTokens || maxBatchCount)
+    ? splitByTokenBudget(
+        truncated,
+        maxBatchTokens
+          ? Math.floor(maxBatchTokens * effectiveSafetyFactor(recipe))
+          : Number.MAX_SAFE_INTEGER,
+        charsPerToken,
+        maxBatchCount,
+      )
     : [truncated];
 
   const allEmbeddings: Float32Array[] = [];
@@ -1568,6 +1577,9 @@ export async function embed(texts: string[], opts?: EmbedOpts): Promise<Float32A
  *   responsible for applying any safety-factor shrink before passing in.
  * @param charsPerToken - Provider-specific character density. Defaults to
  *   `DEFAULT_CHARS_PER_TOKEN` (4) when omitted, matching OpenAI tiktoken.
+ * @param maxBatchCount - #1199: optional cap on INPUTS per sub-batch, for
+ *   providers that reject batches by count (DashScope: 10). When omitted,
+ *   only the token budget governs.
  *
  * @internal exported for tests; not part of the public gateway API.
  */
@@ -1575,15 +1587,17 @@ export function splitByTokenBudget(
   texts: string[],
   budgetTokens: number,
   charsPerToken: number = DEFAULT_CHARS_PER_TOKEN,
+  maxBatchCount?: number,
 ): string[][] {
   const ratio = charsPerToken > 0 ? charsPerToken : DEFAULT_CHARS_PER_TOKEN;
+  const maxCount = maxBatchCount !== undefined && maxBatchCount > 0 ? maxBatchCount : Infinity;
   const batches: string[][] = [];
   let current: string[] = [];
   let currentTokens = 0;
 
   for (const text of texts) {
     const estTokens = Math.ceil(text.length / ratio);
-    if (current.length > 0 && currentTokens + estTokens > budgetTokens) {
+    if (current.length > 0 && (currentTokens + estTokens > budgetTokens || current.length >= maxCount)) {
       batches.push(current);
       current = [];
       currentTokens = 0;
@@ -1609,7 +1623,11 @@ export function isTokenLimitError(err: unknown): boolean {
     /token.*limit.*exceeded/i.test(msg) ||
     // OpenAI embeddings: "Invalid 'input': maximum request size is 300000 tokens per request."
     /maximum request size.*tokens/i.test(msg) ||
-    /max.*tokens.*per.*request/i.test(msg)
+    /max.*tokens.*per.*request/i.test(msg) ||
+    // DashScope: "batch size is invalid, it should not be larger than 10." (#1199)
+    // Count-cap error, but recursive halving shrinks count too, so the same
+    // safety net converges.
+    /batch size is invalid/i.test(msg)
   );
 }
 
--- a/src/core/ai/recipes/dashscope.ts
+++ b/src/core/ai/recipes/dashscope.ts
@@ -31,6 +31,10 @@ export const dashscope: Recipe = {
       // path. Conservative declaration so the gateway pre-splits before
       // hitting whatever undocumented server-side limit exists.
       max_batch_tokens: 8192,
+      // #1199: DashScope hard-caps embeddings at 10 inputs per request
+      // ("batch size is invalid, it should not be larger than 10"). The
+      // token budget alone admits far more than 10 short chunks per batch.
+      max_batch_count: 10,
       // text-embedding-v3 mixes English + CJK heavily; the tokenizer is
       // closer to Voyage density than OpenAI tiktoken for CJK-dominant
       // content. Conservative chars_per_token=2 leaves headroom.
--- a/src/core/ai/recipes/google.ts
+++ b/src/core/ai/recipes/google.ts
@@ -16,6 +16,15 @@ export const google: Recipe = {
       dims_options: [768, 1536, 3072],
       cost_per_1m_tokens_usd: 0.15,
       price_last_verified: '2026-04-20',
+      // #970: Gemini's documented limits are per-INPUT (2048 tokens,
+      // silently truncated beyond) and per-REQUEST count (batchEmbedContents
+      // caps at 100 inputs). There is no separate per-request token cap, so
+      // the token budget is derived: 100 inputs × 2048 tokens. The count cap
+      // binds first for typical chunk sizes. Do NOT copy the 2048 per-input
+      // limit into max_batch_tokens — that would over-split 50×.
+      max_batch_tokens: 204_800,
+      chars_per_token: 4,
+      max_batch_count: 100,
     },
     expansion: {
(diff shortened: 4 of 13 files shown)


time-attack:
Merging, primarily for the two provider-recipe fixes: DashScope embedding is outright broken on master for any batch over 10 inputs, and the Google recipe has no caps at all. Nice judgment deriving the Gemini budget from the per-request limit rather than the per-input number the issue quoted — the literal reading would have over-split by ~50x. One process note: #1207 and #1818 are throughput changes to shared hot paths, not bug fixes, and bundling them under a `fix(embed)` title makes the risk harder to see; I read the concurrency diff carefully (index-addressed results, identical fast path, no dispatch after failure) and it holds up, but please split behavior changes like that out next time.

Issue #2825, opened 2026-07-14
query_cache ignores hard-exclude config — cached results leak GBRAIN_SEARCH_EXCLUDE'd slugs across processes


paul-0320:
**Version**: 0.42.59.0, postgres engine

**Summary**: `knobsHash` (`src/core/search/mode.ts`) does not include the resolved hard-exclude configuration (`GBRAIN_SEARCH_EXCLUDE` env / per-call `exclude_slug_prefixes`). Since the cache row id is `(query, source, knobsHash)`, two processes with *different* exclude configs share the same cache row — whichever runs first poisons the other.

**Repro**:
1. Have pages under a prefix you exclude, e.g. `GBRAIN_SEARCH_EXCLUDE="private/"` on your serve daemon.
2. From a shell **without** the env (e.g. a non-interactive script shell), run `gbrain query "..."` — results legitimately include `private/` pages, and the row is cached.
3. From a process **with** the env, run the same query — it returns the cached row, `private/` pages included. The hard exclude is silently bypassed.
4. The reverse also holds: an env-set query caches filtered results, so a deliberate opt-out query (`GBRAIN_SEARCH_EXCLUDE= gbrain query ...`) silently gets pre-filtered results.

**Impact**: hard excludes act as a privacy/scoping control (the docs position them as such), but the cache makes them nondeterministic across processes. In our deployment (serve daemon with env + occasional CLI shells without), we found **82 poisoned cache rows** in production before noticing.

**Suggested fix**: include the resolved hard-exclude list (sorted, comma-joined) in the `knobsHash` parts and bump `KNOBS_HASH_VERSION` (the code comment there already mandates the bump when adding a knob). Per-call `exclude_slug_prefixes`/`include_slug_prefixes` likely need the same treatment.

**Workaround we're using**: guarantee identical env for every process (moved the export to `~/.zshenv`) and purge poisoned rows: `DELETE FROM query_cache WHERE results::text LIKE '%<excluded-prefix>%'`.

PR #3131, opened 2026-07-21
fix(search): weak-top floor on autocut to stop cross-source result collapse

time-attack:
## Items

- Takeover of #1863 (fix_needed/P1) — original work by @rayers; rebased onto current master. Not closing #1863 here, just referencing.

## Problem

`applyAutocut` normalizes the rerank-score cliff test by the top score, so a weak top (~0.317 on a rare-term cross-source pool) rescales to 1.0 and a normal decay below it looks like a confident cliff — `--source __all__` collapses to 1 result while the same query single-source-scoped returns ~12.

## Fix

- `AutocutConfig.minTopScore` (default 0.5): below the floor autocut no-ops (recall preserved); at/above it, unchanged. `0` disables.
- Threaded like `autocut_jump`: mode bundle → `search.autocut_min_top_score` config → per-call → pick-chain → knobsHash.
- Gated on `rerankerEmitsNormalizedScores(<effective model>)` — honors a per-call `SearchOpts.reranker.model` override; raw-logit local rerankers (llama-server/Qwen3) get floor 0 since a [0,1] threshold is meaningless on an unbounded logit scale.
- `KNOBS_HASH_VERSION` 12→13, folding `acmts=` (4 decimals) so a floored write can't be served to a different-floor lookup.

## Rebase deltas vs #1863

The stale PR bumped the knobs-hash version to a value master has since taken for the #2825 hard-exclude fold. Rebased: v=13 is the acmts addition (append-only, after `hx=`); all five version-pin tests and the bump-history comments updated. Everything else salvaged as-is.

## Tests

- Pure-fn (`test/search/autocut.test.ts`): weak-top → no trim; strong-top → still trims; at-floor cuts; `minTopScore: 0` disables; partial-knobs default agreement.
- End-to-end (`test/search/autocut-integration.serial.test.ts`): weak-top no-trim, strong-top trim, per-call raw-logit model disables the floor.
- Plumbing (`test/search-mode.test.ts`): bundle defaults, precedence, config clamp [0,1], knobsHash separation, `SEARCH_MODE_CONFIG_KEYS`.
- `rerankerEmitsNormalizedScores` unit coverage (`test/search/rerank.test.ts`).

`bun run typecheck` exit 0; 205 tests across the 7 touched files pass.

### DIFF
--- a/src/commands/search.ts
+++ b/src/commands/search.ts
@@ -71,6 +71,7 @@ const KNOB_DESCRIPTIONS: Record<keyof ModeBundle, string> = {
   // v0.42.3.0 autocut
   autocut: 'Score-discontinuity result-sizing (cuts at the rerank-score cliff; no-op without a reranker)',
   autocut_jump: 'Autocut sensitivity: min normalized score gap that counts as a cliff (0..1, 0.20 default)',
+  autocut_min_top_score: 'Autocut weak-top floor: min top rerank score to trust the cliff (0..1, 0.50 default; 0 disables; below it autocut no-ops to preserve recall)',
   // v0.43 relational recall
   relationalRetrieval: 'Typed-edge relational recall arm (relational queries walk the graph; no-op otherwise)',
   relational_retrieval_depth: 'Max hops for relational traversal (1..3, 2 default)',
--- a/src/core/search/autocut.ts
+++ b/src/core/search/autocut.ts
@@ -33,6 +33,19 @@ export interface AutocutConfig {
   jumpRatio: number;
   /** Failsafe: never return fewer than this when candidates exist (≥1). */
   minKeep: number;
+  /**
+   * Weak-top floor: the minimum TOP rerank score for the cliff test to be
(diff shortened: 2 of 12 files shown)


time-attack:
Merging. I confirmed the bug is still live on master — `applyAutocut` normalizes purely by the top score, so a weak top rescales to 1.0 and normal decay looks like a cliff, which is exactly the cross-source collapse in #1863. The `minTopScore` floor is threaded through the existing mode/config/per-call/knobsHash chain rather than bolted on, the knobs-hash bump to v=13 correctly avoids the slot master already took for #2825, and gating on `rerankerEmitsNormalizedScores` is the right call for raw-logit local rerankers. Thanks also for rebasing @rayers' original work rather than discarding it.

garrytan:
Follow-up: this fix is re-landing in #3908 (v0.42.78.0), adapted across the three intervening cache-key versions with the bump history restored. The review here was right and the merge simply never happened — thanks @time-attack for the rebase and @rayers for the original fix in #1863; both are credited in the release CHANGELOG.

PR #3132, opened 2026-07-21
fix(cli): put --file, reject unknown op flags, list --offset pagination

### LINKED ISSUE #380: `gbrain put --file <path>` silently ignored; flag is `--content`, unknown flags should error
## Summary

`gbrain put <slug> --file <path>` is silently ignored — `--file` isn't a registered flag. The only documented flag is `--content`. When `--file` is passed, gbrain swallows it and falls back to reading stdin. Depending on whether stdin is connected, the call either:

- **fails opaquely** with `ENXIO: no such device or address, open '/dev/stdin'`, or
- **succeeds but writes garbage** — creates a page with only auto-derived `title`/`type` (no body, no tags, no chunks).

This bites anyone following a natural "write this file to that slug" mental model (including the `gbrain put <slug> [< file.md]` hint in `gbrain --help`, which looks like a `--file` flag at first glance).

## Repro

```bash
cat > /tmp/repro.md <<'EOF'
---
title: Repro Page
type: concept
tags: [bug-repro, gbrain, cli]
---

# Repro Page

Deliberately minimal body.
EOF

# Case 1 — --file, stdin closed: opaque ENXIO
gbrain put concepts/repro-a --file /tmp/repro.md < /dev/null
# ENXIO: no such device or address, open '/dev/stdin'

# Case 2 — --file + stdin content: --file is ignored, stdin wins
cat /tmp/repro.md | gbrain put concepts/repro-b --file /nonexistent/path
# -> succeeds, creates concepts/repro-b with the full content from the PIPED file,
#    proving --file path is ignored.

# Case 3 — --file pointing at real file, no stdin pipe: page created but empty
gbrain put concepts/repro-c --file /tmp/repro.md
# -> chunks: 0, body empty, only title/type from slug-derivation.

# Case 4 — the only working pattern (per --help):
gbrain put concepts/repro-d --content "$(cat /tmp/repro.md)"
# OR
cat /tmp/repro.md | gbrain put concepts/repro-e
```

## Expected behavior

At least one of:

1. `--file <path>` is a real, documented flag that reads the file and passes it through the same chunk/embed/auto-link pipeline as `--content`.
2. Unknown flags cause the CLI to exit with "unknown flag" rather than silently ignore.
3. The `gbrain --help` top-line hint `put <slug> [< file.md]` is clarified in the per-command help too, so newcomers don't invent `--file` from intuition.

Option 1 is the friendliest — `--file` is what most users want first. Option 2 prevents the silent-garbage case even without adding `--file`.

## Environment

- gbrain: v0.18.2
- Engine: postgres (Supabase), schema v24 (all migrations applied)
- Runtime: Bun 1.3.13 on Ubuntu 24.04 (WSL2)
- Shell: bash

## Observed side effect

During an install + upgrade workflow this evening we lost ~30 minutes of debugging time because an agent had been told in its prompt to "write pages via `gbrain put --file <path>`". Every call created an empty page with just title/type and the agent believed it had succeeded. Classic confabulation assist.

## Suggested fix (small)

Either add a `--file <path>` alias that calls the same internal put as `--content`, or change the CLI parser to reject unknown options. The latter is ~5 lines and prevents the opaque-failure class.

### LINKED ISSUE #2876: `gbrain list --limit` silently clamps at 100 with no pagination
## Summary

`gbrain list --limit N` silently returns at most 100 rows, even when `N > 100`. The help describes `--limit` only as "Max results" and does not disclose a cap. The command exposes no `--offset`, cursor, or other pagination mechanism, so a source/type with more than 100 pages cannot be fully enumerated through `gbrain list`.

This makes `list` unsafe as an existence check: a page can be directly `get`-able while absent from the returned 100-row window.

## Environment

- gbrain: 0.42.59.0
- installed/pinned commit: `5008b287e47bf791132eedfebf66bdef11e9398c`
- Bun: 1.3.9
- engine: PGLite
- OS: macOS
- corpus: >100 active `session-summary` pages

## Reproduction

```bash
gbrain list --type session-summary --limit 50  | wc -l
# 50

gbrain list --type session-summary --limit 500 | wc -l
# 100
```

A second check demonstrates that this is truncation, not a 100-page corpus:

```bash
gbrain list --type session-summary --sort updated_asc  --limit 500
# 100 rows

gbrain list --type session-summary --sort updated_desc --limit 500
# 100 rows
```

The two sort orders return different windows. A slug present only in the ascending window remains retrievable with `gbrain get <slug>`, while it is absent from the descending `list` output.

## Root cause at current HEAD

`src/core/operations.ts` clamps the public operation here:

```ts
limit: clampSearchLimit(p.limit as number | undefined, 50, 100),
```

The engine's `listPages` supports `offset`, but the `list_pages` operation/CLI does not expose it.

## Expected behavior

Any of these would make the contract honest:

1. Honor larger `--limit` values up to a documented safe maximum.
2. Expose pagination (`--offset` or cursor) and document the 100-row page size.
3. At minimum, reject `--limit > 100` with a clear error instead of silently truncating.

Related parser issue: #380 tracks unknown flags being silently ignored.

time-attack:
## Items

- **Fixes #380** — `gbrain put SLUG --file PATH` was silently swallowed by the permissive flag parser, creating empty pages (or hanging on stdin). `--file` now reads the op's stdin param from a file (5MB cap, mutually exclusive with `--content`/stdin in either order, loud error on unreadable path), and unknown flags on shared ops are a hard error before engine connect.
- **Takeover of #856** (not closing it — reference only) — salvaged the unknown-flag rejection + tests from @Kage18's PR, rebased onto current `parseOpArgs` (preserves `--no-<flag>` boolean negation and the `readFileSync(0)` stdin path that landed after the PR). Fixed the regression named in review: blanket rejection would have broken `--source` (read by makeContext on every op), `--brain`, `--dry-run`, and `--json` — these pass through via an explicit allowlist. Dropped the `put_page` hard-coding: `--file` is driven by the declared `cliHints.stdin`, so `volunteer-context` gets it too. Bonus: `--json` on shared ops now actually emits raw JSON (docs already promised it) on both the local-engine and thin-client routed paths.
- **Fixes #2876** — `gbrain list --limit` silently clamped at 100 with no way past it. `list_pages` now declares `offset` (threaded through the OFFSET support both engines already had on `PageFilters`; negative/non-finite values dropped) and the `limit` description discloses the 100-row cap.

## Tests

- `test/cli-args.test.ts` — pure `parseOpArgs`: `--file` fills `content`; allowlist flags (`--source`/`--json`/`--dry-run`) survive on ops that don't declare them.
- `test/cli.test.ts` — subprocess: unknown flag fails before DB connection; `--file`+`--content` conflict; missing `--file` path errors; `put --help` documents `--file`.
- `test/list-pages-offset.test.ts` — offset declared + threaded to `engine.listPages`; negative/non-finite/zero dropped; fractional floored.

`bun run typecheck` exit 0; touched test files (incl. R4 stdin regression pin, tool-defs, parity, pipe-truncation) all pass. Live-smoked on an isolated PGLite brain: init → `put --file` → `get`, `--json`, `--source` passthrough, `list --offset`, conflict/missing-file/unknown-flag error paths.

### DIFF
--- a/docs/tutorials/company-brain.md
+++ b/docs/tutorials/company-brain.md
@@ -223,14 +223,16 @@ export GBRAIN_REMOTE_CLIENT_ID=<Alice's client_id>
 export GBRAIN_REMOTE_CLIENT_SECRET=<Alice's client_secret>
 export GBRAIN_REMOTE_MCP_URL=https://brain.acme-co.com/mcp
 
-gbrain search "performance review" --remote
+gbrain search "performance review"
 ```
+
+(On a thin-client install every shared op routes through the remote MCP server automatically — no flag needed. The env vars select whose credentials the call uses.)
 
 Alice should see results only from `customers` and `shared`. The performance-review notes live in `internal`, which she's not scoped to read. She shouldn't see them.
 
 ```bash
 # Terminal 2, as Bob (export his credentials similarly)
-gbrain search "performance review" --remote
+gbrain search "performance review"
 ```
 
 Bob should see the performance-review notes from `internal`, plus anything related from `shared`. He shouldn't see anything that lives only in `customers`.
@@ -514,7 +516,7 @@ The first sync embeds every page, which takes time. Check `gbrain sources status
 
 ### "I see a page I shouldn't see"
 
-This shouldn't happen, but if you suspect it, run `gbrain search <query> --remote --json` as the constrained client and inspect the `source_id` field on every returned result. Every row should be in the client's `--federated-read` set. If one isn't, file an issue with the exact slug and source IDs.
+This shouldn't happen, but if you suspect it, run `gbrain search <query> --json` as the constrained client (thin-client install, with the client's `GBRAIN_REMOTE_*` env exported) and inspect the `source_id` field on every returned result. Every row should be in the client's `--federated-read` set. If one isn't, file an issue with the exact slug and source IDs.
 
 ### "The synthesized answer is wrong"
 
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -464,7 +464,11 @@ async function main() {
     // routed path. Date → ISO string; bigint → string (postgres.js shape);
     // Buffer → object. Microsecond-cost; eliminates a whole drift bug class.
     const result = JSON.parse(JSON.stringify(rawResult, bigintToStringReplacer));
-    const output = formatResult(op.name, result);
+    // #380 pass-through: `--json` (undeclared on most ops, promised by docs)
+    // emits the raw op result instead of the human formatter.
+    const output = params.json === true
+      ? JSON.stringify(result, null, 2) + '\n'
+      : formatResult(op.name, result);
     if (output) process.stdout.write(output);
   } catch (e: unknown) {
     // v0.42.20.0 (codex D4): on error, set exitCode + return so the `finally`
@@ -547,7 +551,10 @@ async function runThinClientRouted(
       signal: sigintController.signal,
     });
     const result = unpackToolResult(raw);
-    const output = formatResult(op.name, result);
+    // #380: same --json seam as the local-engine path (renderer parity).
+    const output = params.json === true
+      ? JSON.stringify(result, null, 2) + '\n'
+      : formatResult(op.name, result);
     if (output) process.stdout.write(output);
   } catch (e: unknown) {
     if (e instanceof RemoteMcpError) {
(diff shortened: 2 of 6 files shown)


time-attack:
Merging. Silently swallowing an unknown flag and writing an empty page is a data bug, and I confirmed both the `--file` gap and the permissive parser are still on master. I paid particular attention to the regression that blocked @Kage18's #856 — the passthrough allowlist covers `--source`/`--brain`/`--dry-run`/`--json`, and the global progress flags are already stripped by `parseGlobalFlags` before `parseOpArgs` runs, so blanket rejection is safe here. Note the `list --offset` item already landed on master, so that third of the diff is a no-op.

PR #3134, opened 2026-07-21
fix(skills): handle CRLF skill frontmatter in health checks

time-attack:
Takeover of #2444 (credit: @Cjaewon0708 — co-authored on the commit). Salvages the real CRLF fix, repairs its regression, and drops the undisclosed scope.

## What

- `src/core/skill-frontmatter.ts` + `src/core/check-resolvable.ts`: frontmatter fence and `triggers:` block parsing now accept Windows CRLF line endings (`/^---\r?\n…/`, `split(/\r?\n/)`). Previously CRLF `SKILL.md` files were silently invisible to health checks and MECE overlap detection.
- `skills/RESOLVER.md`: adds the missing `skill-optimizer` route (disclosed in the original PR); llms bundles regenerated.

## Repairs vs #2444

- The original rewrote `blockRe` with raw `\s` / `\t` escapes inside a template literal, which degrades `\s` to a literal `s` — `triggers: ` with trailing whitespace (previously matched) would silently stop parsing. This takeover keeps doubled escapes (`\\s*\\r?\\n((?:[ \\t]+-[ \\t]+[^\\r\\n]+\\r?\\n?)+)`) and adds a regression test pinning the trailing-whitespace case.
- Dropped ~100 lines of undisclosed `write-through.ts` Windows path-casing / rename-EPERM changes; that work should land as its own described PR.

## Tests

- `test/skill-brain-first.test.ts`: CRLF fence + block-array parse test; trailing-whitespace-after-colon regression guard (fails against the original #2444 pattern).
- `test/check-resolvable.test.ts`: CRLF skills now surface a `mece_overlap` that LF-only parsing missed.
- Both new tests fail without the src fix (verified via stash); `bun run typecheck` and `bun test test/build-llms.test.ts` green.

References #2444 (not auto-closing).

### DIFF
--- a/llms-full.txt
+++ b/llms-full.txt
@@ -1411,6 +1411,7 @@ This is the dispatcher. Skills are the implementation. **Read the skill file bef
 | "get more out of gbrain", "is my brain set up right", "weekly brain checkup", "advise me on my brain", "gbrain advisor" | `skills/gbrain-advisor/SKILL.md` |
 | Save or load reports | `skills/reports/SKILL.md` |
 | "Create a skill", "improve this skill" | `skills/skill-creator/SKILL.md` |
+| "optimize this skill", "tune the skill against the benchmark", "make the skill better", "run skillopt" | `skills/skill-optimizer/SKILL.md` |
 | "Skillify this", "is this a skill?", "make this proper" | `skills/skillify/SKILL.md` |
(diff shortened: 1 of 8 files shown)


time-attack:
Merging. Confirmed the LF-only fence and `blockRe` on master, so CRLF skill files are genuinely invisible to the health checks — silent invisibility is worse than a loud failure here. I appreciate two things specifically: catching that the original pattern degraded `\s` to a literal `s` inside the template literal (and pinning it with a regression test), and dropping the undisclosed `write-through.ts` scope instead of carrying it along. Credit to @Cjaewon0708 for the original diagnosis.

PR #3135, opened 2026-07-21
fix(sync): guard putPage 0-row RETURNING + reclassify sync failure copy

### LINKED ISSUE #2189: gbrain sync (code) fails on ~all files: "undefined is not an object (evaluating 'row.deleted_at')" on PGLite (v0.42.44.0)

time-attack:
Fixes #2189
Takeover of #2586 (credit: @javieraldape, co-authored on the commit)

## What

- **Root-cause guard (#2189):** `putPage`'s `INSERT … ON CONFLICT DO UPDATE … RETURNING` can yield 0 rows when brain-local DB state (e.g. a `BEFORE` trigger suppressing the write) prevents the upsert. Pre-fix, `rowToPage(rows[0])` crashed with the opaque `undefined is not an object (evaluating 'row.deleted_at')` that failed ~all files of the reporter's code sync. Both engines (`src/core/pglite-engine.ts` + `src/core/postgres-engine.ts`, parity invariant) now throw a descriptive error naming the slug + source_id, turning the anonymous TypeError into a diagnosable per-file failure that points at the suppressing trigger/rule.
- **Copy reclassification (salvaged from #2586):** the sync-blocked message said "failed to parse / fix the frontmatter", misclassifying runtime import errors as YAML problems. Reworded to "failed to import".
- **Tests:** #2586's first-code-sync PGLite smoke test, plus a new regression test (`test/putpage-returning-guard.test.ts`) that reproduces the exact 0-row RETURNING state via a suppressing trigger — it fails pre-fix with the reported crash signature and passes with the guard.

## Why the takeover

PR #2586's diff was copy-only + a smoke test that passes at the reporter's exact commit, so it could not catch #2189's crash; merging it as "Fixes #2189" would have closed a P1 unfixed. The copy + test are salvaged here; the missing behavioral guard is added at the root cause.

## Verification

- `bun test test/putpage-returning-guard.test.ts` — 2 pass (verified it fails pre-fix with the exact `row.deleted_at` signature)
- `bun test test/sync.test.ts` — 69 pass
- `bun run typecheck` — clean

### DIFF
--- a/src/commands/sync.ts
+++ b/src/commands/sync.ts
@@ -197,7 +197,7 @@ export interface SyncResult {
   /** Pages re-embedded during this sync's auto-embed step. 0 if --no-embed or skipped. */
   embedded: number;
   pagesAffected: string[];
-  failedFiles?: number; // count of parse failures (Bug 9)
+  failedFiles?: number; // count of per-file import/sync failures (Bug 9)
   /**
    * v0.41.13.0 partial-sync fields (only set when status === 'partial').
    *
@@ -3183,7 +3183,7 @@ async function performSyncInner(engine: BrainEngine, opts: SyncOpts): Promise<Sy
     await clearOpCheckpoint(engine, ckpt.target);
   };
 
-  // issue #1939 adversarial finding #1: a file that failed to parse (open ledger
+  // issue #1939 adversarial finding #1: a file that failed to import (open ledger
   // row) and is then deleted/renamed-away never re-enters failedFiles and never
   // imports, so its row would never clear and would age doctor to a permanent
   // FAIL. Treat removed paths as resolved so the ledger self-heals.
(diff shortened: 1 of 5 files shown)


time-attack:
Merging. The 0-row `RETURNING` path really is unguarded on both engines today, and an opaque `row.deleted_at` TypeError across every file of a sync is a bad failure mode for something that turns out to be a local DB trigger. Fixing it in both engines keeps the parity invariant, and the copy reclassification matters on its own — telling someone to fix their frontmatter when the actual failure is a runtime import error sends them down the wrong path entirely. Good catch that #2586 as-written would have closed a P1 unfixed; thanks for salvaging @javieraldape's test rather than dropping it.

PR #3136, opened 2026-07-21
fix(cycle): enforce the extract-atoms drain deadline end to end

### LINKED ISSUE #2750: fix(cycle): extract-atoms drain window does not bound lock/query/write lifetime

time-attack:
Fixes #2750
Takeover of #2752 (credit: @panda850819 — signal-threading design salvaged; three flaws repaired, see below). References #2752 only; closes nothing there.

## Problem

`runExtractAtomsDrain` checked `now() < deadline` only **between** batches. A single batch (sequential LLM calls per page), the lock acquire/release, and `countRemaining` were all unbounded by the window — observed as window=120s runs overrunning to 282.5s.

## Fix

The drain derives one real-time deadline signal (`AbortSignal.timeout(windowMs)` combined with the Minion `job.signal`) and threads it through everything the loop awaits:

- **Drain loop** (`extract-atoms-drain.ts`): hung count/batch/lock-acquire are cancelled at the deadline; after the window elapses the final backlog count is skipped (`remaining: null` instead of overrunning); an external worker abort rethrows after the lock is released so Minion records the abort.
- **Phase** (`extract-atoms.ts`): `abortSignal` forwarded to every gateway chat call, the discovery/count/idempotency queries, and every `putPage` write, plus a cooperative between-item check — the primary bound on PGLite, where query abort only abandons the waiter, so the **default engine keeps a working drain**. Billable usage is recorded before the post-chat abort check; receipt/rollup bookkeeping runs on a fresh 5s grace signal so committed atoms never lose their cost trail; deadline-truncated runs report `deadline_aborted: true` and don't count as completed rounds.
- **Lock** (`db-lock.ts`): `tryAcquireDbLock`/`withRefreshingLock` accept a deadline signal; release routes through the direct session pool (same rationale as refresh, #1794) on a fresh grace timeout; deadline-bound callers skip the unbounded same-host takeover path (honest busy result; TTL stays the backstop).
- **Engines** (both): `putPage` accepts an optional signal — Postgres cancels the in-flight statement, PGLite pre-checks (documented in the interface); `executeRawDirect` short-circuits an already-fired signal before pool routing and bounds a stalled direct-pool acquisition.

## What changed vs PR #2752

Salvaged: the deadline-signal architecture, the drain-loop semantics, the lock/release bounding, the bookkeeping grace window, and most tests.

Repaired:
1. **Conflict + regression with #2988**: the PR's atomic multi-atom bulk INSERT bypassed `putPage` and omitted `chunker_version`, re-introducing the doctor `contextual_retrieval_coverage` bug #2988 just fixed (and conflicting with its rewrite of the same hunk in both engines). This takeover keeps the `putPage` write path — a signal-capable `putPage` is equally cancellable, and every current/future `putPage` invariant (chunker_version default, dedup slugs, provenance COALESCE) applies automatically. The bulk-INSERT slug-collision/sanitization machinery became unnecessary and was dropped (master's deterministic `atomSlug` already covers collisions).
2. **PGLite hard refusal removed**: `runExtractAtomsDrainForSource` no longer throws on engines without real query cancellation — that would have broken `gbrain dream --drain` and the `extract-atoms-drain` Minion job on the default engine. PGLite gets cooperative between-item abort instead.
3. **Transcript-suppression scope change reverted**: the PR silently passed `_transcripts: []` and removed `brainDir` from the drain, changing what the drain processes. This takeover keeps master's transcript behavior byte-for-byte (pinned by a test).

## Tests

- `test/extract-atoms-drain.test.ts`: 6 new pure-loop tests (hung count aborted at deadline + lock released, external cancel rethrown after release, one combined signal into count/batch, window-vs-no_progress classification, hung lock acquisition bounded, wiring pins incl. `brainDir` retention) — all 6 fail against the old implementation.
- `test/cycle/extract-atoms-synthesize-concepts.test.ts`: 5 new phase tests (hung chat aborted, billable-usage-before-abort, partial progress still writes receipt + incomplete rollup, bookkeeping on fresh grace signal, hung atom write cancelled).
- `test/postgres-execute-raw-direct.test.ts`: pre-routing short-circuit + stalled direct-pool acquisition bounded.

Gates: `bun run typecheck` clean; touched test files 44/44 pass; JSONB guards (`check-jsonb-pattern.sh`, `check-jsonb-params.mjs`) clean; db-lock/dream/cycle/extract suites green.

### DIFF
--- a/src/commands/jobs.ts
+++ b/src/commands/jobs.ts
@@ -2059,6 +2059,8 @@ export async function registerBuiltinHandlers(
         sourceId,
         windowSeconds,
         brainDir: repoPath,
+        // #2750: worker cancel/timeout/lock-loss propagates into the drain.
+        abortSignal: job.signal,
       });
     } catch (e) {
       if (e instanceof LockUnavailableError) {
--- a/src/core/cycle/extract-atoms-drain.ts
+++ b/src/core/cycle/extract-atoms-drain.ts
@@ -23,6 +23,10 @@
  */
 
 import type { BrainEngine } from '../engine.ts';
+import { anySignal } from '../abort-check.ts';
+
+/** Fresh cleanup budget for the lock release after the window signal fires. */
+const LOCK_RELEASE_GRACE_MS = 5_000;
 
 export interface ExtractAtomsDrainDeps {
   /**
@@ -30,13 +34,13 @@ export interface ExtractAtomsDrainDeps {
    * via `withRefreshingLock`. MUST throw when the lock is held by another
    * process (e.g. `LockUnavailableError`) — the drain lets that propagate so
    * the caller can report `cycle_already_running` and exit, matching the
-   * routine cycle's skip contract.
+   * routine cycle's skip contract. The signal bounds lock acquisition too.
    */
-  withLock: <T>(work: () => Promise<T>) => Promise<T>;
-  /** Process one bounded batch (rediscovers eligibility). Returns counts. */
-  runBatch: () => Promise<{ extracted: number; skipped: number }>;
+  withLock: <T>(work: () => Promise<T>, signal: AbortSignal) => Promise<T>;
+  /** Process one batch. The signal fires at the drain wallclock deadline. */
+  runBatch: (signal: AbortSignal) => Promise<{ extracted: number; skipped: number }>;
   /** Count remaining eligible-but-unextracted pages, or null on query error. */
-  countRemaining: () => Promise<number | null>;
+  countRemaining: (signal: AbortSignal) => Promise<number | null>;
   /** Injectable clock. Production: Date.now. */
   now: () => number;
   /** Optional progress sink (one line per batch). */
@@ -48,6 +52,8 @@ export interface ExtractAtomsDrainOpts {
   windowMs: number;
   /** Hard cap on batches (belt-and-suspenders against a 0-progress loop). Default 1000. */
   maxBatches?: number;
+  /** External caller cancellation (worker timeout / shutdown). */
+  abortSignal?: AbortSignal;
 }
 
 export interface ExtractAtomsDrainResult {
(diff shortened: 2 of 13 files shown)


time-attack:
Merging. A drain window that only gets checked between batches isn't a window, and the 120s-becomes-282s overrun in #2750 reproduces on master — none of the lock acquire, the per-page LLM calls, or `countRemaining` were bound. Threading one derived deadline signal through everything the loop awaits is the right shape, and I specifically checked the details that usually get missed here: engine parity on the `putPage` signal, billable usage recorded before the post-chat abort check, and the cooperative between-item check that keeps PGLite bounded. Good test-to-source ratio.

PR #3137, opened 2026-07-21
fix(embed): label content_chunks.model with the model that produced the vector (#1717)

### LINKED ISSUE #1717: Embedding: `gbrain embed`/`import` mislabels content_chunks.model as the gateway default (v0.41.x)

time-attack:
Fixes #1717
Takeover of #1803 — salvages @harjothkhara's per-chunk labeling + preserve-on-untouched semantics, rebased onto current master (the original diff no longer applied after the pace-mode rework wrapped upsertChunks in `observed(pacer, ...)`), and extends the fix to the inline import/sync paths the PR didn't cover.

## What
`content_chunks.model` recorded the engines' hardcoded `DEFAULT_EMBEDDING_MODEL` instead of the gateway-configured model that actually produced each vector, because the embed and import paths built `ChunkInput[]` without a `model` field.

## Changes
- `src/core/embedding.ts`: new `resolveEmbeddingModelLabel()` — resolved `provider:model` gateway label, `undefined` when the gateway is unconfigured (callers then keep the chunk's existing model rather than mislabeling).
- `src/commands/embed.ts` (embedPage / embedAll / embedAllStale): stamp the label on (re)embedded chunks; chunks preserved from a prior embed keep their existing model so a mixed-model page isn't relabeled wholesale.
- `src/core/import-file.ts`: stamp the label on inline-embedded markdown chunks and re-embedded code chunks; incremental code imports that reuse an existing embedding now carry the existing model label forward instead of falling back to the default.

## Tests
- `test/embed.serial.test.ts`: re-embedded chunk gets the resolved model, untouched chunk keeps its original label (fails on master).
- `test/import-signature-stamp.serial.test.ts`: end-to-end PGLite import — `content_chunks.model` equals the configured gateway model, not the default (fails on master).

Both engines are unchanged (no SQL edits), so engine parity is unaffected; the existing `COALESCE(EXCLUDED.model, ...)` upsert semantics now receive correct per-chunk labels from every caller.

### DIFF
--- a/src/commands/embed.ts
+++ b/src/commands/embed.ts
@@ -1,5 +1,5 @@
 import type { BrainEngine } from '../core/engine.ts';
-import { embedBatch, currentEmbeddingSignature } from '../core/embedding.ts';
+import { embedBatch, currentEmbeddingSignature, resolveEmbeddingModelLabel } from '../core/embedding.ts';
 import type { ChunkInput } from '../core/types.ts';
 import { chunkText } from '../core/chunkers/recursive.ts';
 import { createProgress, type ProgressReporter } from '../core/progress.ts';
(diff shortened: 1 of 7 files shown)


time-attack:
Thanks for the careful writeup and the rebase work salvaging #1803. Closing as already fixed: master now resolves the runtime gateway model inside both engines' chunk upsert fallback (postgres-engine.ts:2436, pglite-engine.ts:2320), landed in e1919fab (reland of #2846 via #3343), so chunks no longer get stamped with the hardcoded DEFAULT_EMBEDDING_MODEL. The remaining delta here is only the mixed-model-page relabeling nuance, which isn't worth threading a new label resolver through three caller paths.

PR #3145, opened 2026-07-21
fix(embedding,sources,files): Ollama per-model dims, config-repair array guard, upload-raw persistence

### LINKED ISSUE #2170: Ollama provider hardcodes 768d, rejects 1024d models (bge-m3, mxbai-embed-large)

### LINKED ISSUE #2251: Dream cycle: updateSourceConfig array-coercion branch throws "jsonb_each on a non-object", permanently blocking last_full_cycle_at writes
## Summary

On the Postgres/Supabase engine, `updateSourceConfig` (`src/core/postgres-engine.ts`, ~L1312–1334) coerces a malformed `sources.config` back to an object before the `|| patch` merge, via a `CASE` statement. The **array** branch flattens with:

```sql
jsonb_array_elements(config) elem, jsonb_each(elem)
```

`jsonb_each` requires an object, so if the array contains any **non-object element**, the statement raises `cannot call jsonb_each on a non-object` and the **entire UPDATE fails**. The `string` and `else` branches are already guarded; the `array` branch is not.

Effect: the per-cycle `last_full_cycle_at` write in `cycle.ts` (~L2294–2302) fails every cycle for the affected source. The source then looks perpetually stale and is re-cycled indefinitely (wasted cycle/LLM budget). It is caught and logged as:

```
[cycle] failed to write last_full_cycle_at for source <id>: cannot call jsonb_each on a non-object
```

so it is non-fatal but silently wasteful.

## How the bad shape arises

An earlier build merged config via a naive `config || patch` concat. When `config` was a (double-encoded) JSON **string** rather than an object, `string || object` yields a JSONB **array**, and each subsequent cycle appended another `{"last_full_cycle_at": "..."}` element. Observed in the wild: configs that grew to ~86 KB — a leading double-encoded string element `"{\"federated\":true}"` followed by hundreds of timestamp objects, one per cycle. The current build's `CASE` coercion was added to repair this, but the array branch trips over the leading **string** element, so it can neither merge nor self-heal.

## Reproduce

1. Put a source's `config` into the bad shape:
   ```sql
   UPDATE sources SET config = '["x", {"last_full_cycle_at":"2026-01-01T00:00:00Z"}]'::jsonb WHERE id = '<id>';
   ```
2. Run a cycle that calls `updateSourceConfig` for that source.
3. Observe the `jsonb_each on a non-object` failure; `last_full_cycle_at` is never written.

## Version

gbrain 0.42.42.0, Postgres (Supabase) engine.

## Suggested fix

Guard the array branch to only `jsonb_each` **object** elements and fall back to `{}` otherwise, mirroring the existing string/else guards — e.g. add `WHERE jsonb_typeof(elem) = 'object'` to the `jsonb_array_elements` subquery and wrap in `COALESCE(..., '{}'::jsonb)`.

## Note for affected installs

`gbrain repair-jsonb` does **not** cover this — it targets only the v0.12 string-double-encode case on a fixed column list (`pages.frontmatter`, `raw_data.data`, `ingest_log.pages_updated`, `files.metadata`, `page_versions.frontmatter`), not `sources.config`. Affected rows need a manual data repair (reset `config` to a valid object, preserving the latest `last_full_cycle_at`).

### LINKED ISSUE #2297: files upload-raw silently no-ops for small text/PDF (returns success, persists nothing)

time-attack:
Backlog fix wave, work unit c37-1-f1 (embedding preflight & dims: local providers) + two verified P1s.

### 1. Per-model native dims for Ollama local embed models
Fixes #2170. Takeover of #2642 (rebased onto #2271's `trust_custom_dims`; credit to @Willisbest, co-author on the commit).

The Ollama recipe declared one provider-wide `default_dims: 768`, so the flagless path (`gbrain init --embedding-model ollama:bge-m3` or `ollama:mxbai-embed-large`) sized a 768d schema for 1024d models and `gbrain embed` failed at runtime. #2271 fixed only the explicit `--embedding-dimensions` path.

- `EmbeddingTouchpoint.model_dims` — per-model native width map (keys pinned as a subset of `models` by a contract test).
- `effectiveDefaultDims(tp, modelId)` in `embedding-dim-check.ts` — single source of truth; `validateDimAgainstTouchpoint` now takes the touchpoint and resolves the SELECTED model's width (NVIDIA natural dims keep precedence).
- `init.ts` routes all three default-dims sites (shorthand pick, verbose model without dims, env auto-detect) through it.
- Ollama recipe: adds `bge-m3` and declares native widths for all six listed models.

#2642's Tier-1.5 hunk (accept any explicit dim for `user_provided_models`) was dropped — fully superseded by master's `trust_custom_dims` — but its llama-server/litellm explicit-dims regression tests are kept.

### 2. updateSourceConfig array-branch repair no longer fatal on mixed shapes
Fixes #2251. The Postgres config-shape repair ran `jsonb_each(elem)` over every array element; a non-object element (the documented in-the-wild shape: double-encoded string + timestamp objects) raised `cannot call jsonb_each on a non-object`, failing the whole UPDATE — `last_full_cycle_at` never wrote and the dream cycle re-ran the source forever. The element is now guarded with a `CASE` (not a `WHERE` qual) so `jsonb_each` never sees a non-object regardless of qual-evaluation order. Postgres-only by design (the CASE repair exists only in postgres-engine). Expression semantics verified against Postgres 17.5 (PGLite) with the mixed `["str", {...}, 42, null]` shape; DATABASE_URL-gated e2e test added.

### 3. files upload-raw small-file branch actually persists
Fixes #2297. The `!needsCloud` branch printed `{success:true, storage:'git', path:<caller's input file>}` and returned without copying anything into the brain repo or inserting a files row. It now copies the file to the documented `<page-slug>/.raw/<filename>` sidecar (`unsorted/.raw/` without `--page`), records a files row (executeRawJsonb upsert on storage_path), reports the real destination, and fails loudly when no brain repo is configured.

### Tests
- `test/embedding-dim-check.test.ts` — per-model dim resolution (all six Ollama models, explicit-dim paths, trust_custom_dims passthrough), llama-server/litellm explicit dims, model_dims⊆models contract. New tests fail without the fix (bge-m3 resolved 768 before).
- `test/e2e/list-all-sources-postgres.test.ts` — mixed-array config repaired, patch merged, non-object elements skipped (DATABASE_URL-gated).
- `test/files-upload-raw.test.ts` — new: sidecar copy + files row + unsorted fallback + idempotent re-upload (fails without the fix: nothing persisted).

Gates: `bun run typecheck` exit 0; touched test files 149 pass / 0 fail; `check-jsonb-pattern.sh` + `check-jsonb-params.mjs` clean.

### DIFF
--- a/src/commands/files.ts
+++ b/src/commands/files.ts
@@ -196,13 +196,45 @@ async function uploadRaw(engine: BrainEngine, args: string[]) {
   const needsCloud = stat.size >= SIZE_THRESHOLD || isMedia;
 
   if (!needsCloud) {
-    // Small text/PDF files stay in git
+    // Small text/PDF files stay in git — copy into the brain repo's `.raw/`
+    // sidecar (the shape documented in skills/_brain-filing-rules.md) and
+    // record a files row. #2297: this branch used to print success and
+    // persist NOTHING (the reported path was the caller's input file), so
+    // provenance was silently lost.
+    const { getDefaultSourcePath } = await import('../core/source-resolver.ts');
+    const repoPath = await getDefaultSourcePath(engine);
+    if (!repoPath) {
+      console.error('No brain repo configured — cannot preserve the raw file in git.');
+      console.error('Set a source local_path (`gbrain sources add <id> --path <repo>`) or configure sync.repo_path.');
+      process.exit(1);
+    }
+    const content = readFileSync(filePath);
+    const hash = createHash('sha256').update(content).digest('hex');
+    const relPath = join(pageSlug ?? 'unsorted', '.raw', filename);
+    const destPath = join(repoPath, relPath);
+    mkdirSync(dirname(destPath), { recursive: true });
+    writeFileSync(destPath, content);
+
+    await executeRawJsonb(
+      engine,
+      `INSERT INTO files (page_slug, filename, storage_path, mime_type, size_bytes, content_hash, metadata)
+       VALUES ($1, $2, $3, $4, $5, $6, $7::jsonb)
+       ON CONFLICT (storage_path) DO UPDATE SET
+         content_hash = EXCLUDED.content_hash,
+         size_bytes = EXCLUDED.size_bytes,
+         mime_type = EXCLUDED.mime_type`,
+      [pageSlug, filename, relPath, mimeType, stat.size, 'sha256:' + hash],
+      [{ type: fileType, storage: 'git' }],
+    );
+
     console.log(JSON.stringify({
       success: true,
       storage: 'git',
-      path: filePath,
+      path: destPath,
+      storagePath: relPath,
       size: stat.size,
       size_human: humanSize(stat.size),
+      hash: `sha256:${hash}`,
     }));
     return;
   }
--- a/src/commands/init.ts
+++ b/src/commands/init.ts
@@ -10,6 +10,7 @@ import { saveConfig, loadConfig, loadConfigFileOnly, toEngineConfig, gbrainPath,
 import { createEngine } from '../core/engine-factory.ts';
 import { discoverOAuth, mintClientCredentialsToken, smokeTestMcp } from '../core/remote-mcp-probe.ts';
 import { runInitEmbedCheck } from '../core/init-embed-check.ts';
+import { effectiveDefaultDims } from '../core/embedding-dim-check.ts';
 
 export async function runInit(args: string[]) {
   // Help guard: cli.ts only routes --help to printOpHelp() for shared-op
@@ -246,7 +247,7 @@ async function resolveAIOptions(opts: ResolveAIOptionsArgs): Promise<ResolvedAIO
       process.exit(1);
     }
     out.embedding_model = `${shorthand}:${firstModel}`;
-    out.embedding_dimensions = recipe.touchpoints.embedding!.default_dims;
+    out.embedding_dimensions = effectiveDefaultDims(recipe.touchpoints.embedding!, firstModel);
   }
 
   if (dimsArg !== null && !Number.isNaN(dimsArg) && dimsArg > 0) {
@@ -271,7 +272,11 @@ async function resolveAIOptions(opts: ResolveAIOptionsArgs): Promise<ResolvedAIO
       process.exit(1);
     }
     if (recipe?.touchpoints.embedding?.default_dims) {
-      out.embedding_dimensions = recipe.touchpoints.embedding.default_dims;
+      // #2170: resolve the SELECTED model's native width (e.g. ollama:bge-m3 →
+      // 1024), not the provider-wide default_dims, so the schema isn't sized
+      // wrong when no --embedding-dimensions was passed.
+      const modelId = out.embedding_model.slice(providerId.length + 1);
+      out.embedding_dimensions = effectiveDefaultDims(recipe.touchpoints.embedding, modelId);
     }
   }
 
@@ -436,7 +441,7 @@ async function resolveEmbeddingByEnv(out: ResolvedAIOptions, nonInteractive: boo
         await import('../core/ai/defaults.ts');
       const dims = fullModel === DEFAULT_EMBEDDING_MODEL
         ? DEFAULT_EMBEDDING_DIMENSIONS
-        : tp.default_dims;
+        : effectiveDefaultDims(tp, model);
       out.embedding_model = fullModel;
       out.embedding_dimensions = dims;
       console.error(
--- a/src/core/ai/recipes/ollama.ts
+++ b/src/core/ai/recipes/ollama.ts
@@ -14,19 +14,29 @@ export const ollama: Recipe = {
   touchpoints: {
     embedding: {
       // #2271: modern local embed models added so assertTouchpoint accepts them.
-      // Each carries its own native dim (qwen3-embed-8b=4096, arctic-l-v2=1024);
-      // the recipe-wide default_dims below is only the nomic fallback, so users
-      // of the larger models pass --embedding-dimensions (allowed via
-      // trust_custom_dims). Per-model dims metadata is a tracked follow-up.
       models: [
         'nomic-embed-text',
+        'bge-m3',
         'mxbai-embed-large',
         'all-minilm',
         'qwen3-embed-8b',
         'snowflake-arctic-embed-l-v2',
       ],
-      default_dims: 768, // nomic-embed-text native dim
-      trust_custom_dims: true, // #2271: local models carry varied native dims
+      default_dims: 768, // nomic-embed-text native dim (fallback for models absent from model_dims)
+      // #2170: Ollama models emit DIFFERENT fixed native widths, but one
+      // provider-wide default_dims silently built a 768d schema for 1024d
+      // models (bge-m3, mxbai-embed-large) → `gbrain embed` then failed with a
+      // dim mismatch. Declare each model's real native width so the schema is
+      // sized correctly with no --embedding-dimensions flag.
+      model_dims: {
+        'nomic-embed-text': 768,
+        'bge-m3': 1024,
+        'mxbai-embed-large': 1024,
+        'all-minilm': 384,
+        'qwen3-embed-8b': 4096,
+        'snowflake-arctic-embed-l-v2': 1024,
+      },
+      trust_custom_dims: true, // #2271: locally-pulled model variants may carry non-listed dims
       cost_per_1m_tokens_usd: 0,
       price_last_verified: '2026-04-20',
       // Ollama's batch capacity depends on the locally loaded model + the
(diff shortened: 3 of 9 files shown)


time-attack:
Thanks for this — the diagnosis was right on all three, but the branch has been overtaken. Ollama per-model dims (#2170) are on master now via the `model_dims` map in recipes/ollama.ts, and the #2251 array-branch guard already ships as the `CASE WHEN jsonb_typeof(...) = 'object'` wrapper in source-config-sql.ts. The one item still live, the upload-raw small-file no-op (#2297), is carried in #3149 with the storage_path namespacing on top, so I'm closing this in favor of that one.

PR #3146, opened 2026-07-21
fix(facts): idempotent fence deposits + re-runnable fence-backfill command

### LINKED ISSUE #1867: Remote `extract_facts` deposits are never re-fenced — `row_num`-NULL backlog grows unboundedly; v0_32_2 is one-time and   un-rerunnable
### Environment
  - gbrain 0.42.23.0 (installed from github default-branch HEAD)
  - Engine: direct Postgres (localhost, no pooler/Supabase), schema v113
  - Personal single-source brain, `default` is repo-backed

  ### Summary
  Facts deposited via **remote MCP `extract_facts`** (`remote=true`) land in the DB with `row_num IS NULL` and are **never written into
  their entity page's `## Facts` fence**. The only code that fences DB→page is the **one-time `v0_32_2` Phase B** migration, which is
  registry-terminal and cannot be re-run. So the "legacy v0.31 facts pending fence backfill" counter is not a one-time historical backlog
  — it **grows with every remote deposit** (observed 587 → 707 over a couple of days), and the `extract_facts` cycle phase stays
  permanently guard-skipped.

  ### Why it happens (root cause)
  1. v0.32.2 contract: new facts write fence-first, then stamp the DB index. But the trust boundary suppresses markdown writes when
  `remote=true` (`src/mcp/server.ts` / `src/core/operations.ts`), so remote `extract_facts` facts are DB-only with `row_num IS NULL`.
  2. The cycle guard (`src/core/cycle/extract-facts.ts:112`) counts `SELECT COUNT(*) FROM facts WHERE row_num IS NULL AND entity_slug IS
  NOT NULL` and **refuses** the reconcile phase while any exist, pointing the operator to `gbrain apply-migrations --yes`.
  3. But `v0_32_2` is marked `complete` in `~/.gbrain/migrations/completed.jsonl`. `statusForVersion`
  (`src/commands/apply-migrations.ts:145`) returns `complete` if **any** entry is complete — checked **before** the `retry` branch — so
  `--force-retry` cannot regress it, `--force-orchestrator` only resets `wedged` migrations, and `--migration 0.32.2` still buckets it as
  `applied`. The runner only executes `[...partial, ...pending]` (`apply-migrations.ts:420`). There is no supported re-run path.
  4. No standalone fence-backfill command exists (`gbrain backfill` is unknown in this build); the only DB→fence logic is
  `__testing.phaseBFenceFacts` inside the migration.

  Net: remote-deposited facts accumulate as `row_num`-NULL forever, and the prescribed remedy is unreachable. Self-reinforcing: the guard
  message tells you to run a migration that won't run.

  ### Impact
  - `gbrain dream` `extract_facts` phase is permanently skipped (`legacyRowsPending > 0` guard).
  - Facts remain in the DB and stay queryable, but are **never mirrored into the markdown system-of-record**, so a rebuild purely from the
  repo (`gbrain rebuild`) would silently drop them. This breaks the v0.32.2 "DB is fully derivable from the repo" promise for any brain
  that deposits via remote MCP.
  - The backlog and the guard-skip are unbounded over time.

  ### Repro
  1. Postgres brain, v0_32_2 already applied (any normal upgrade).
  2. Deposit several facts via the remote MCP `extract_facts` op (`remote=true`).
  3. `SELECT count(*) FROM facts WHERE row_num IS NULL AND entity_slug IS NOT NULL;` → grows by each deposit.
  4. `gbrain apply-migrations --yes` → "All migrations up to date." `--migration 0.32.2 --dry-run` → "Already applied. Nothing to do." No
  way to fence them.

  ### Suggested fixes (any one)
  - A first-class, re-runnable command, e.g. `gbrain facts fence-backfill [--dry-run]`, exposing `phaseBFenceFacts` outside the migration
  registry (it's already idempotent on `row_num IS NULL`, atomic per-page, dirty-tree-guarded).
  - Have the cycle `extract_facts` phase fence orphan `row_num`-NULL rows itself (DB→fence) instead of only refusing, so the steady-state
  reconciler closes the loop.
  - Make remote `extract_facts` write its fence row on a subsequent trusted pass (e.g. the `*/15` conveyor or `gbrain extract all --source
  db`) so remote deposits don't perpetually re-arm the guard.

  ### Workaround (what I'm doing meanwhile)
  Parking the backlog (facts stay live/queryable/consolidating) and holding a deliberate out-of-band invocation of
  `__testing.phaseBFenceFacts` (backup → dry-run → apply → verify) as a pre-rebuild escape hatch.

### LINKED ISSUE #2044: Remote `extract_facts` fence-write is not idempotent — 2nd deposit to an already-fenced page throws `idx_facts_fence_key` duplicate and rolls back the whole batch
## Summary
On v0.42.23.0, remote (`remote=true`) MCP `extract_facts` now correctly fence-writes deposits (good — this is the behavior #1867 was about). But the fence-write path is **not idempotent across repeated writes to the same page**. The *first* remote `extract_facts` that fences a given entity/page succeeds; a *second* deposit routed to an already-fenced page throws:

```
```

and the **entire batch rolls back** — no facts land. A connected AI's deposit silently fails.

## Impact
This directly degrades the primary deposit path. Every MCP-connected AI (Claude app, Claude Code, Grok web/Build) uses `extract_facts visibility:world` as its default deposit. A second deposit in the same session to a page already fenced earlier loses all facts in that batch. The durable record survives only if the content also reached a committed page body by another route.

## Observed
- v0.42.23.0, direct Postgres (localhost:5432, no pooler).
- 2026-06-06: deposits ids 774–781 fenced fine on first write; a follow-up deposit to an already-fenced entity raised the duplicate-key error and rolled the batch back.

## Expected
A repeat fence-write to an already-fenced page should upsert / no-op on the existing fence key rather than insert-and-violate, so repeated same-page deposits within/across sessions are safe.

## Current workaround (standing, documented in our runbook)
Deposit once per page per session; route follow-up deposits to not-yet-fenced entities; or capture via the local CLI (`remote=false`). Durable record is the committed page body regardless.

## Relation
Sibling to #1867 (remote re-fencing). #1867 was about deposits *never* being fenced; this is about the now-active fence path colliding on *repeat* writes.

time-attack:
Fixes #2044
Fixes #1867

## #2044 — remote `extract_facts` fence-write is not idempotent

A second deposit to an already-fenced page threw `idx_facts_fence_key` and rolled back the whole `insertFacts` batch whenever the fence and the DB row_num keyspace diverged (e.g. legacy fence writes that landed at a path the current code no longer reads, leaving DB rows stamped against a fence that later restarted at row 1).

- `writeFactsToFence` now seeds the next row_num from `MAX(fence max, SELECT MAX(row_num) FROM facts WHERE source_id/source_markdown_slug)`, so a diverged page appends past the DB high-water mark instead of colliding.
- Belt-and-suspenders: `insertFacts` in BOTH engines (postgres + pglite, parity) adds `ON CONFLICT (source_id, source_markdown_slug, row_num) WHERE row_num IS NOT NULL DO NOTHING`, so a residual collision skips that one row instead of failing the batch. Returned ids stay input-ordered with skipped rows omitted; no caller indexes them positionally.

## #1867 — row_num-NULL backlog had no re-runnable remedy

The fence backfill lived only inside the one-shot `v0_32_2` migration, and the cycle `extract_facts` guard pointed at `gbrain apply-migrations --yes` — which never re-runs a completed migration, so the backlog (remote deposits that predate the fence backstop) was permanent.

- `phaseBFenceFacts` (already idempotent on `row_num IS NULL`) is exported and exposed as `gbrain facts fence-backfill [--dry-run]` (thin-client refused with a host hint, like the other FS-bound commands).
- The cycle guard message now points at the new command.
- The backfill (and its verify phase) route page paths through `resolvePageFilePath` so non-default sources fence into the same `.sources/<id>/` path the fence-write backstop and put_page write-through compute — the exact wrong-path divergence class behind #2044.

## Tests

- `test/fence-write.test.ts`: fence/DB divergence case — DB holds row_num 1..3, page has no fence; deposit lands at row 4 in both fence and DB (fails without the fix).
- `test/insert-facts-batch.test.ts`: collision now skips only the colliding row, rest of batch commits (updated from the old whole-batch-rollback contract).
- `test/facts-fence-backfill-command.test.ts` (new): backfill fences the backlog, re-run is a no-op, `--dry-run` writes nothing.
- `test/extract-facts-phase.test.ts`: guard hint asserts the new command name.

`bun run typecheck` clean; touched test files green (62 pass across the 5 direct files; one pre-existing cross-file embedding-dim contamination failure in `facts-backstop-integration` reproduces identically on pristine master).

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -54,7 +54,7 @@ export function bigintToStringReplacer(_key: string, value: unknown): unknown {
 }
 
 // CLI-only commands that bypass the operation layer
-export const CLI_ONLY = new Set(['init', 'reinit-pglite', 'upgrade', 'post-upgrade', 'check-update', 'integrations', 'publish', 'check-backlinks', 'lint', 'report', 'import', 'export', 'files', 'embed', 'serve', 'call', 'config', 'doctor', 'migrate', 'eval', 'sync', 'extract', 'extract-conversation-facts', 'enrich', 'features', 'autopilot', 'graph-query', 'jobs', 'agent', 'apply-migrations', 'skillpack-check', 'skillpack', 'resolvers', 'integrity', 'repair-jsonb', 'orphans', 'sources', 'mounts', 'dream', 'check-resolvable', 'routing-eval', 'skillify', 'smoke-test', 'providers', 'storage', 'repos', 'code-def', 'code-refs', 'reindex', 'reindex-code', 'reindex-frontmatter', 'code-callers', 'code-callees', 'reconcile-links', 'frontmatter', 'auth', 'friction', 'claw-test', 'book-mirror', 'takes', 'think', 'salience', 'anomalies', 'calibration', 'transcripts', 'models', 'remote', 'recall', 'forget', 'edges-backfill', 'cache', 'ze-switch', 'founder', 'brainstorm', 'lsd', 'schema', 'capture', 'onboard', 'conversation-parser', 'status', 'connect', 'skillopt', 'quarantine', 'self-upgrade', 'advisor', 'watch', 'reindex-search-vector']);
+export const CLI_ONLY = new Set(['init', 'reinit-pglite', 'upgrade', 'post-upgrade', 'check-update', 'integrations', 'publish', 'check-backlinks', 'lint', 'report', 'import', 'export', 'files', 'embed', 'serve', 'call', 'config', 'doctor', 'migrate', 'eval', 'sync', 'extract', 'extract-conversation-facts', 'enrich', 'features', 'autopilot', 'graph-query', 'jobs', 'agent', 'apply-migrations', 'skillpack-check', 'skillpack', 'resolvers', 'integrity', 'repair-jsonb', 'orphans', 'sources', 'mounts', 'dream', 'check-resolvable', 'routing-eval', 'skillify', 'smoke-test', 'providers', 'storage', 'repos', 'code-def', 'code-refs', 'reindex', 'reindex-code', 'reindex-frontmatter', 'code-callers', 'code-callees', 'reconcile-links', 'frontmatter', 'auth', 'friction', 'claw-test', 'book-mirror', 'takes', 'think', 'salience', 'anomalies', 'calibration', 'transcripts', 'models', 'remote', 'recall', 'forget', 'edges-backfill', 'facts', 'cache', 'ze-switch', 'founder', 'brainstorm', 'lsd', 'schema', 'capture', 'onboard', 'conversation-parser', 'status', 'connect', 'skillopt', 'quarantine', 'self-upgrade', 'advisor', 'watch', 'reindex-search-vector']);
 // CLI-only commands whose handlers print their own --help text. These are
 // excluded from the generic short-circuit so detailed per-command and
 // per-subcommand usage stays reachable.
@@ -991,6 +991,8 @@ const THIN_CLIENT_REFUSED_COMMANDS = new Set([
   // hint pointing at the routable MCP tools; per-subcommand splits are
   // a v0.31.x follow-up TODO.
   'takes', 'sources',
+  // #1867: fence-backfill edits local .md fences + stamps the local DB.
+  'facts',
   // v0.32 thin-client routing audit (Codex round 2 findings #2, #4):
   // - `pages` purge-deleted is admin+localOnly (operations.ts:856-864)
   // - `files` list / file_url MCP ops are localOnly (operations.ts:1769-1879)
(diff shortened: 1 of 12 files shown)


time-attack:
Merging. Confirmed on master that insertFacts has no per-row conflict skip in either engine, so one diverged row_num still rolls back the whole deposit batch — that's fact loss, not just noise. Exposing the already-idempotent phaseBFenceFacts as `facts fence-backfill` is justified given the migration can never re-run and the backlog is otherwise permanent. Good call routing the backfill through resolvePageFilePath so non-default sources fence to the same path the write-through computes.

PR #3147, opened 2026-07-21
fix(gateway): providers key checks read the folded gateway env; autopilot wrapper sources ~/.gbrain/env + boot warning

### LINKED ISSUE #2608: autopilot --install wrapper sources ~/.zshrc for API keys — worker env silently lost, every LLM job no-ops with no warning
**Version:** 0.42.56.0 (also current master `6461790`) · **Engine:** Postgres 16 + pgvector · **Runtime:** Bun · **Host:** remote Linux VM, systemd user unit · **Chat:** corporate OpenAI-compatible endpoint (key in a dedicated env file)

## What happens

`gbrain autopilot --install` generates `~/.gbrain/autopilot-run.sh` whose only mechanism for loading API keys is sourcing shell rc files (`writeWrapperScript`, `src/commands/autopilot.ts:1118-1141`):

```bash
[ -f ~/.zshenv ] && source ~/.zshenv 2>/dev/null
exec '<gbrain>' autopilot --repo '<repo>'
```

If `~/.zshrc` exists, the `||` chain stops there — `~/.bashrc` is never tried, and there is no fallback to any gbrain-owned env file. On a host where the LLM provider env (`OPENROUTER_API_KEY`, `OPENROUTER_BASE_URL`, `NODE_EXTRA_CA_CERTS`, …) lives in a dedicated env file rather than in zsh rc files, the autopilot worker starts with **no provider environment at all**.

The failure is then completely silent. Every LLM-dependent job no-ops instead of erroring:

- the chronicle judge checks `isAvailable('chat')` and returns `{ events: [] }` when false (`src/core/chronicle/extract-events.ts:172-173`), which becomes an ordinary `status: 'no_events'` result;
- dream/enrich phases likewise produce nothing.

Observed symptom: autopilot runs green for days — cycles complete, sync/embed work (they don't need chat) — while chronicle and dream extract zero events and spend zero tokens. Nothing in `autopilot.err`, job output, or doctor points at "the worker has no chat credentials".

## What should happen

Either or both:

1. **Source the env the daemon actually needs**, not the user's interactive shell config: a gbrain-owned env file (e.g. `~/.gbrain/keys.env` or a configured `env_file`) sourced unconditionally after the rc-file attempts. #1868 suggests the same for the Linux `PATH`/bun variant of this problem — the rc-file-sourcing approach is the shared root cause.
2. **Fail loudly when chat is unavailable in a daemon context**: if `isAvailable('chat')` is false during an autopilot cycle that includes LLM phases, log a prominent warning (and/or surface it in `doctor`/`status`) instead of letting every judge return empty results indefinitely.

## Where (master `6461790`)

- `src/commands/autopilot.ts:1118-1141` (`writeWrapperScript` — rc-file sourcing, no env-file fallback)
- `src/core/chronicle/extract-events.ts:172-173` (`isAvailable('chat')` → silent `{ events: [] }`)

## Repro

1. Put provider keys in a dedicated env file only (none in `~/.zshrc`); ensure `~/.zshrc` exists.
2. `gbrain autopilot --install` on a systemd host; start the unit.
3. Watch chronicle/dream phases: all pages return `no_events`, token spend 0, no errors anywhere.

## Workaround

Hand-patch `autopilot-run.sh` to `source ~/.gbrain/<env-file>` — but `autopilot --install` regenerates the wrapper, so the patch must be reapplied after every reinstall/upgrade.

## Related papercut

After an unclean stop, a stale `~/.gbrain/autopilot.lock` makes the systemd unit flap ("Another autopilot instance is running… Exiting" in a restart loop) until the lock is removed manually. #2503 covers the adjacent stale-PID-reuse case; a daemon-managed lock probably shouldn't require manual cleanup after any crash.

### LINKED ISSUE #2728: providers command bypasses canonical gateway config
## Summary

`gbrain providers` builds gateway configuration separately from the rest of the CLI instead of using `buildGatewayConfig()`. On v0.42.58.0 this means provider diagnostics can ignore file-plane API keys and can behave differently from the real configured gateway path. The explicit `providers test --model` override also drops configured provider base URLs.

This is especially visible when a launcher injects an empty provider environment variable: `buildGatewayConfig()` correctly filters the empty value and preserves the configured key, while the providers command does not use that ownership seam.

## Affected version

- `v0.42.58.0`
- commit `a25209bbb2bacf1b88e06fd5282b27f1bf4a3e7a`

## Reproduction

1. Put a valid Anthropic key and native base URL in the GBrain config plane.
2. Run from an environment where `ANTHROPIC_API_KEY` is unset or injected as an empty string.
3. Run:

```bash
gbrain providers test --touchpoint chat --model anthropic:claude-sonnet-4-6
```

`src/commands/providers.ts` currently constructs gateway config manually:

- `configureFromEnv()` copies model/base URL fields plus raw `process.env`, but does not map file-plane API keys.
- The `--model` test override passes only the model/dimensions plus raw `process.env`, dropping the configured key and configured base URL.

The result can disagree with the real CLI path that uses `buildGatewayConfig()`.

## Expected behavior

Provider listing/testing should use the same gateway configuration ownership seam as normal CLI execution. `providers test --model` should override only the requested model and embedding dimensions; credentials, base URLs, fallback config, and empty-env handling should continue to come from `buildGatewayConfig()`.

## Proposed fix

- Route `configureFromEnv()` through `buildGatewayConfig(loadConfig() ?? { engine: 'pglite' })` so the command still works before `gbrain init`.
- Add a pure `buildProviderTestGatewayConfig()` helper that starts from the canonical config and changes only the tested model/dimensions.
- Add a regression test using the repository's `withEnv()` helper to prove an empty `ANTHROPIC_API_KEY` does not clobber the configured key and that configured base URLs survive the override.

## Verification

On a clean branch based directly on v0.42.58.0:

```text
bun test test/providers.test.ts test/ai/build-gateway-config.test.ts
21 pass, 0 fail
```

The repository verification suite is 30/31 on the untouched baseline and on the patch. The one pre-existing failure is unrelated: `test/e2e/openclaw-plugin-load-real.test.ts:281` resolves an undeclared parent `openclaw/plugin-sdk` type that lacks `registerContextEngine`.

An apply-ready two-file patch/commit is available.

time-attack:
## Items

- Fixes #2728 — `providers env` and `providers explain` key detection read bare `process.env`, so an API key living only in `~/.gbrain/config.json` (file plane) was reported as missing even though the real gateway path resolved it fine. New `gatewayEnv()` helper (loadConfig → buildGatewayConfig → env) is now the single env source for `list`/`env`/`explain` key checks. (The issue's primary claim — `configureFromEnv` bypassing `buildGatewayConfig` — and the `test --model` override were already fixed on master by #2863; this closes the residual surfaces.)
- Fixes #2608 — the `autopilot --install` wrapper sourced only `~/.zshenv`/`~/.zshrc`/`~/.bashrc`; non-interactive daemon shells never see zshrc-only exports, so worker env was silently lost and every LLM phase no-op'd with no signal. The wrapper now also sources the gbrain-owned `~/.gbrain/env` (after the profiles, so it wins), and autopilot logs one boot-time stderr warning when `isAvailable('chat')` is false so the no-op failure mode is visible in the daemon log.

## Tests

- `test/providers-gateway-env.test.ts` — file-plane-only OPENAI_API_KEY folds into `gatewayEnv()` (fails on the pre-fix bare-process.env behavior).
- `test/autopilot-install.test.ts` — wrapper sources `~/.gbrain/env` after the shell profiles; `chatBootWarning` warns when chat is unavailable, silent otherwise.

`bun run typecheck` exit 0; touched test files 21 pass / 0 fail.

### DIFF
--- a/src/commands/autopilot.ts
+++ b/src/commands/autopilot.ts
@@ -365,7 +365,11 @@ export function chatBootWarning(chatAvailable: boolean): string | null {
   if (chatAvailable) return null;
   return (
     '[autopilot] WARNING: no chat provider available — LLM phases (extract, dream, enrich) will no-op. ' +
-    'Set a key via `gbrain config set anthropic_api_key <key>` or export it in ~/.gbrain/env (sourced by the daemon wrapper).'
+    // NOT `gbrain config set anthropic_api_key`: that writes the DB plane,
+    // which the gateway never reads (isAvailable('chat') stays false) — the
+    // split-brain class build-gateway-config.ts documents. File plane or env
+    // are the two paths that actually resolve.
+    'Export ANTHROPIC_API_KEY in ~/.gbrain/env (sourced by the daemon wrapper) or set "anthropic_api_key" in ~/.gbrain/config.json.'
   );
 }
 
--- a/src/commands/autopilot.ts
+++ b/src/commands/autopilot.ts
@@ -356,6 +356,19 @@ async function attemptAutopilotSelfUpgrade(
   }
 }
 
+/**
+ * #2608 — pure function (test seam, same pattern as generateLaunchdPlist):
+ * the boot-time warning emitted when no chat provider is available, so the
+ * silent no-op of every LLM phase is visible in the daemon log.
+ */
+export function chatBootWarning(chatAvailable: boolean): string | null {
+  if (chatAvailable) return null;
+  return (
+    '[autopilot] WARNING: no chat provider available — LLM phases (extract, dream, enrich) will no-op. ' +
+    'Set a key via `gbrain config set anthropic_api_key <key>` or export it in ~/.gbrain/env (sourced by the daemon wrapper).'
+  );
+}
+
 export async function runAutopilot(engine: BrainEngine, args: string[]) {
   if (args.includes('--help') || args.includes('-h')) {
     console.log(
@@ -415,6 +428,18 @@ export async function runAutopilot(engine: BrainEngine, args: string[]) {
   } catch { /* best-effort */ }
 
   console.log(`Autopilot starting. Repo: ${repoPath}, interval: ${baseInterval}s`);
+
+  // #2608: LLM phases (extract-events, dream synthesis, enrich) gate on
+  // isAvailable('chat') and silently no-op when no chat provider resolves —
+  // the classic symptom of a daemon shell that never sourced the API keys.
+  // One loud boot-time stderr line makes that failure mode visible in the
+  // daemon log instead of manifesting as "autopilot runs but nothing gets
+  // extracted".
+  try {
+    const { isAvailable } = await import('../core/ai/gateway.ts');
+    const warn = chatBootWarning(isAvailable('chat'));
+    if (warn) console.error(warn);
+  } catch { /* diagnostic only — never blocks the loop */ }
 
   // Mode resolution: Minions dispatch when the user has opted in AND the
   // worker daemon can actually run (Postgres only; PGLite's exclusive file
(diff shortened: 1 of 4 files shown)


time-attack:
Merging. The inconsistency is plain on master: providers.ts:148 already reads the folded gateway env for the status table while the key checks at 286/295/312 read bare process.env, so file-plane keys get reported missing while the gateway resolves them fine. Routing all three surfaces through one `gatewayEnv()` helper is the right shape. The autopilot boot warning when `isAvailable('chat')` is false is the more valuable half — a daemon that silently no-ops every LLM phase is the worst failure mode we ship.

Issue #1503, opened 2026-05-26
Cycle extract phase still writes links/timeline to `default` source on federated brains (follow-up to #1204)


seaneoliver:
# Cycle extract phase still writes links/timeline to `default` source on federated brains (follow-up to #1204)

## Summary

#1204 fixed the **CLI** `gbrain extract` source-awareness by adding `--source-id` (v0.37.7.0). But the **automated dream/autopilot cycle** still hits the same root cause on a federated brain: every cycle logs `Links: created 0 from N pages` / `Timeline: created 0 entries from N pages` and persists nothing, because the cycle's extract phase never threads a `sourceId` and falls back to the literal `'default'` source.

Two interacting defects, both still present on **v0.41.2.0**:

1. **`runPhaseExtract` doesn't thread `sourceId`.** It calls `runExtractCore({ mode, dir, slugs })` with no source. `ExtractOpts` has no `sourceId` field, so the full-walk extractors (`extractLinksFromDir`/`extractTimelineFromDir`) never set `from_source_id/to_source_id`, and `addLinksBatch` maps missing → literal `'default'`, then INNER-JOINs `pages ON (slug, source_id='default')`. On a brain whose content lives in a non-`default` source, every row is dropped by the JOIN → 0 inserted, no error. (Same mechanism as #1204 Bug 1, just reached via the cycle instead of the CLI.)

2. **`noExtract: willRunExtractPhase` disables the *working* path.** `runPhaseSync` resolves the correct `sourceId` (`resolveSourceForDir`) and `performSync` would run the source-aware inline hooks (`extractLinksForSlugs`/`extractTimelineForSlugs`, which #1204 confirmed are correct). But sync passes `noExtract: willRunExtractPhase` to avoid duplicate work — so when the cycle's extract phase is scheduled (the default, `ALL_PHASES`), sync **skips its source-correct inline extraction** and defers to the broken full-walk phase. Net: the cycle turns off the good path and keeps the broken one.

## Environment

- gbrain **v0.41.2.0**, Supabase Postgres engine
- Federated brain, 3 sources: `default` (0–10 pages, isolated), `sos` (~32.7k pages, repo-backed, the live vault), `msft` (~2.8k pages)
- Symptom in `~/.gbrain/autopilot.log`, every cycle:
  ```
  Links: created 0 from 13671 pages
  Timeline: created 0 entries from 13671 pages
  ```

## Repro

[...]

PR #840, opened 2026-05-10
fix(import-file): skip empty code files so reindex-code stops failing on them

nightlaro:
## Summary

`importCodeFile` had no guard for zero-byte content. An empty file got a page row inserted with NULL `compiled_truth`, then every subsequent `gbrain reindex-code` pass failed on that slug with `missing compiled_truth` (the early-return at `src/commands/reindex-code.ts:193-196`).

Empty files are legitimate in real repos — stub/placeholder files committed during refactors (e.g., a Kotlin class temporarily emptied while the API surface is being reworked). Skipping them at import is the same shape as the existing `MAX_FILE_SIZE` branch directly above the new code, so the diff is one symmetric `if`.

## Repro

Before this fix on an Android KMP repo with three empty tracked code files:

```
$ gbrain reindex-code --source gstack-code-bile-... --yes
reindex-code: 0 reindexed, 1781 skipped, 3 failed (1784 total code pages).

3 failure(s):
  games-head-to-head-...-playerandchallengerprofile-kt: missing compiled_truth
  games-head-to-head-...-profilewithscoreandname-kt: missing compiled_truth
  ios-jeopardycoresample-logstore-swift: missing compiled_truth
```

After this fix + one-time `gbrain purge-deleted --older-than-hours 0` to clear the stuck rows:

```
$ gbrain reindex-code --source gstack-code-bile-... --yes
reindex-code: 0 reindexed, 1781 skipped, 0 failed (1781 total code pages).
```

## Test

`test/import-empty-code-file.test.ts` adds two cases against PGLite:

- Empty content → `status: 'skipped'`, `error: 'empty code file'`, and `getPage(slug)` returns `null`. The short-circuit must happen before any page row is written, otherwise reindex-code still trips on NULL `compiled_truth`.
- Non-empty content still imports normally.

Follows the canonical PGLite test pattern from `CONTRIBUTING.md` (single `beforeAll(connect+initSchema)` / `afterAll(disconnect)`, no `process.env` mutation, no `mock.module`).

## Test plan

- [x] `bun test test/import-empty-code-file.test.ts` — 2/2 pass
- [x] `bun test test/incremental-chunking.test.ts test/reindex-code.test.ts test/sync-failures.test.ts test/code-def-refs.test.ts` — 59/59 pass (no regression in `importCodeFile` callers)
- [x] `bun run typecheck` — clean
- [x] `bun run check:wasm` / `check:admin-build` / `check:cli-exec` — OK
[...]

PR #3148, opened 2026-07-21
fix(core): empty code files, multi-source cycle sync, root-level entity refs (#840 #1079 #1101)

time-attack:
Part 4-1 backlog fix wave: three singleton fix-needed community PRs, salvaged and rebased onto master.

## Takeover of #840 — skip empty code files in importCodeFile
The original guard was correct but no longer applied (master inserted the `runGuardrails` seam) and a bare skip left a file that BECAME empty ghosting its stale page/chunks forever. Rebased the guard ahead of the guardrail seam and added stale-page deletion (chunks cascade), mirroring the markdown importer's empty-content branch. `reindex-code` no longer fails every pass on 0-byte files.

## Takeover of #1079 — sync all registered sources in the autopilot cycle
`runPhaseSync` synced only the single brainDir-matched source. It now enumerates every registered, non-archived, `syncEnabled` source with a local checkout, while:
- preserving the `SyncLockBusyError` per-source skip (#1794) the original rewrite would have clobbered,
- keeping the legacy brainDir fallback when no registered source covers it,
- respecting explicit `--source` scoping (#1503) so a scoped cycle stays single-source,
- keeping the single-target phase-result shape byte-compatible (aggregate summary only appears with 2+ targets).

## Takeover of #1101 — root-level file references in ENTITY_REF_RE
The original bare-filename alternative was greedy and emitted slugs WITH the `.md` extension, plus bogus refs for anchors (`#section`) and non-markdown assets (`chart.png`), and broke dir-whitelist semantics. The rebased branch requires a mandatory `.md` suffix, excludes `#`/`/`/`:` from the class, strips `.md` in `extractEntityRefs`, and reports `dir: ''` for root-level refs (matching the generic-wikilink pass). `LINK_EXTRACTOR_VERSION_TS` bumped so previously-stamped pages re-extract the new refs (one fixture date in `test/extract-stale.test.ts` moved with it).

## Tests
- `test/import-empty-code-file.test.ts` (new): skip, became-empty deletion, non-empty unaffected.
- `test/core/cycle.serial.test.ts`: multi-source enumeration (archived/disabled/checkout-less excluded), legacy fallback, explicit-source scoping.
- `test/link-extraction.test.ts`: root-level `.md` refs, `./` prefix, anchors/assets/URLs rejected, dir-prefixed refs unchanged.
- `bun run typecheck` clean; all touched + neighboring suites green (link/extract/cycle/sync/backlinks/reindex batches, 458 tests, 0 fail).

Credits: @nightlaro (#840), @samvilian (#1079), @essexcanning (#1101).

### DIFF
--- a/src/core/cycle.ts
+++ b/src/core/cycle.ts
@@ -978,10 +978,17 @@ async function runPhaseSync(
           sourceId: t.sourceId,
           dryRun,
           noPull: !pull,
-          noEmbed: true,                       // embed is a separate phase
-          noExtract: willRunExtractPhase,      // dedupe ONLY when cycle's extract phase will also run.
-                                               // If extract isn't scheduled (e.g. `gbrain dream --phase sync`),
-                                               // sync's inline extract still runs to preserve prior behavior.
+          noEmbed: true,                       // embed is a separate phase (runEmbedCore --stale is brain-wide,
+                                               // so it covers every target here)
+          // Dedupe inline extract ONLY for the target the cycle's extract
+          // phase will actually cover: extract walks brainDir under
+          // cycleSourceId (runPhaseExtract), so slugs synced from OTHER
+          // checkouts are invisible to it — disabling their inline extract
+          // would silently drop link/timeline extraction for every
+          // non-brainDir source. If extract isn't scheduled at all (e.g.
+          // `gbrain dream --phase sync`), inline extract runs everywhere
+          // to preserve prior behavior.
+          noExtract: willRunExtractPhase && t.repoPath === brainDir,
         });
         synced++;
         added += result.added;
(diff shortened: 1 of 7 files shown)


time-attack:
Merging. Spot-checked all three against master: importCodeFile has a size ceiling but no empty guard, and runPhaseSync still resolves a single brainDir-matched source, so both bugs are live. The rebase quality is what sells it — preserving the #1794 per-source lock skip and #1503 scoping instead of clobbering them, and keeping the single-target result shape byte-compatible. Thanks to @nightlaro, @samvilian and @essexcanning for the originals; worth flagging in the release notes that the LINK_EXTRACTOR_VERSION_TS bump forces a re-extract pass.

Issue #2301, opened 2026-06-19
Bug: PGLite init silently falls back to `--no-embedding`, then blocks all recovery paths


zunami:
### Summary

`gbrain init --pglite --embedding-model <anything>` unconditionally forces `--no-embedding: deferred setup`, regardless of the supplied embedding model. After that, every documented recovery path (`config set`, `init --force`, `mv + re-init`) leads to the same dead end. The user is caught in an infinite loop of "configure with config set" → "file-plane field, can't change" → "wipe and re-init" → same `--no-embedding`.

### Environment

| | |
|---|---|
| **Project** | [garrytan/gbrain](https://github.com/garrytan/gbrain) |
| **Version** | 0.42.51.0 |
| **Engine** | PGLite |
| **OS** | Debian trixie/sid (Proxmox LXC) |
| **Runtime** | Bun (container) |
| **Embedding Infrastructure** | Ollama 0.5.x on separate host (192.168.1.245:11434), model `nomic-embed-text:latest` (274 MB, 768d) — verified reachable and functional via `curl` |
| **gbrain endpoint** | `http://192.168.1.236:8889` |

### Steps to Reproduce

1. Fresh PGLite init with explicit embedding model:
   ```bash
   gbrain init --pglite --embedding-model ollama:nomic-embed-text
   ```
   → Output: `--no-embedding: deferred setup — configure with gbrain config set embedding_model <id> before import`

2. Follow the prompt:
   ```bash
   gbrain config set embedding_model ollama:nomic-embed-text
   ```
   → Output: `embedding_model is a file-plane field that sizes the schema. Setting it in the DB has no effect on the embed pipeline (silent no-op). To switch embedding models/dimensions on PGLite, wipe and re-init: mv ... gbrain init --pglite --embedding-model ... gbrain sync`

3. Follow the second prompt — wipe and re-init with same parameters:
   ```bash
   mv /root/.gbrain/brain.pglite /root/.gbrain/brain.pglite.bak
   gbrain init --pglite --embedding-model ollama:nomic-embed-text
   ```
   → Same result: `--no-embedding: deferred setup`

4. Try `--force`:
   ```bash
   gbrain init --force --pglite --embedding-model ollama:nomic-embed-text
   ```
   → Same: `--no-embedding: deferred setup`

5. Try to embed after any of the above:
   ```bash
   gbrain embed --stale
   ```
   → `This brain was initialized with --no-embedding (deferred setup). Configure an embedding provider before running embed / import`

[...]

PR #3149, opened 2026-07-21
fix(reliability): revive soft-deleted slugs on put, degraded-search reason, silent-failure wave takeover

### LINKED ISSUE #2028: query embed deadline (~6s) silently degrades hybrid search to keyword-only — empty results for CJK content on slow embedding providers

### LINKED ISSUE #2330: Diagnostics surface: health reporting is structurally dishonest (omitted checks score as PASS)
# gBrain diagnostics surface — health reporting is structurally dishonest

**Version:** gbrain 0.42.43.0 · engine: postgres (Supabase) · OS: Windows · schema pack: gbrain-nick-v1
**Date:** 2026-06-20
**Validated:** MCP surface reproduced live (`run_doctor`, `get_health`, `get_stats`, `list_skills`, `takes_scorecard`). CLI figures (70/15) are operator-reported, not independently reproduced here — the MCP half is what proves the divergence and it reproduced exactly.

---

## TL;DR

Three doctor surfaces disagree on the same brain at the same moment. Root cause is **not** "MCP omits checks" — it's that **omitted/unscored checks silently count as PASS**. Live proof: `run_doctor` returned `skill` category **100/100** while `list_skills` could not find the skills directory at all. A perfect score on a set the server cannot read.

Fix the default-pass behavior and the headline divergence (Bug 1) and the unfixable-FAIL (Bug 2) both surface honestly on their own.

A second, previously-unreported divergence exists **inside the MCP layer itself**: `get_health` and `get_stats` disagree on page count (1385 vs 915) and coverage.

---

## Severity (re-ordered from original report)

| # | Issue | Sev | Note |
|---|---|---|---|
| 0 | Omitted check scores as PASS (the generator) | **High** | Was buried inside original Bug 1. This is the actual defect. |
| 1 | Bundled skills read-only → unfixable hard FAIL | **High** | Only fix that clears the FAIL. Confirmed harder than reported. |
| 2 | MCP `run_doctor` check-set divergence | **High** | Falls out once #0 is fixed. |
| 3 | `get_health` vs `get_stats` divergence (MCP-vs-MCP) | **Med** | New. Not in original report. |
| 4 | `takes_count` advertises a dead config key | **Med** | Cheap fix. |
| 5 | `reranker_health` warns on a disabled path | **Low** | Reconcile warn vs disable state. |
| 6 | `config show` omits `search.*` / `takes.*` keys | **Low** | Operator-reported; not re-tested. |

---

## Bug 0 — Omitted/unscored checks silently score as PASS (root cause)

**Severity:** High — this is the mechanism behind the entire health-surface divergence.

**Live evidence:**
- `run_doctor` → `health_score: 90`, `status: "warnings"`, `category_scores.skill: 100`.
- `list_skills` (same brain, same session) → `storage_error: "No skills directory found on the server host."`

The resolver can't locate the skills tree, yet the skill category reports a perfect 100. The doctor is scoring an absent/unreadable check set as fully healthy rather than `unknown` or FAIL. Every check the MCP surface drops (`resolver_health`, `type_proliferation`, `pack_upgrade_available`, `cycle_freshness`, `effective_date_health`, `skill_brain_first`, `content_sanity_audit_recent`, entity/timeline coverage) is invisibly contributing a pass.

**Expected:** An omitted or unscored check resolves to status `unknown` and **never** to `ok`. Any FAIL-level check anywhere in the full registry caps overall status at `unhealthy`.

**Verify:** With the skills dir absent, `category_scores.skill` must not be 100. With `resolver_health` FAILing, top-line `status` must be `unhealthy`, not `warnings`.

---

## Bug 1 — MCP `run_doctor` reports a reduced check set as full health

**Severity:** High.

**Live evidence:** MCP `run_doctor` returned 32 checks scoring 90/100 "warnings". It omits `resolver_health` and the seven other checks the CLI runs. The two surfaced `top_issues` are `brain_score` (67/100) and `reranker_health` — both warns, no FAIL ever shown.

**Impact:** Any operator/agent auditing via MCP (Cowork, Claude Code) sees 90/healthy and never learns the skill resolver is failing.

**Expected:** MCP `run_doctor` runs the same registry as the CLI, OR explicitly declares itself a subset **and** bubbles any FAIL-level check from the full set.

**Verify:** MCP `run_doctor` check set == CLI `doctor` check set (or a declared subset that still surfaces FAILs).

---

## Bug 2 — Bundled skills are read-only; the documented `--fix` cannot act, and the resolver can't even see them

**Severity:** High — the headline FAIL is unfixable by the documented command, and its root is a shipping defect.

**Original repro:**
```
gbrain check-resolvable --fix
→ refused: skills dir resolved via install-path fallback (read-only).
```

**Live evidence (worse than reported):** MCP `list_skills` doesn't return skills-with-missing-triggers — it errors outright:
```
storage_error: No skills directory found on the server host.
```
So from the MCP host the skills directory isn't merely read-only, it's unresolved. `resolver_health`'s 66 `MECE_GAP` warnings are each a **bundled gbrain skill** (`query`, `maintain`, `enrich`, `capture`, `skill-creator`, `skillpack-harvest`, …) missing a `triggers:` array, plus `skill-optimizer` UNREACHABLE. These ship inside the install tree (`C:\Users\nkrum\gbrain\skills`), which `--fix` correctly refuses to mutate. Separately, `skill_conformance` reports 51/51 skills pass — the skills are valid; only resolver routing metadata is missing.

**Impact:** Operator cannot clear the top FAIL. Pointing `GBRAIN_SKILLS_DIR` at a writable copy and patching ~50 vendor `SKILL.md` files gets reverted on the next `gbrain upgrade`.

**Expected:** Bundled skills ship with the `triggers:` frontmatter the resolver requires (fix upstream), OR `resolver_health` soft-grades the read-only install tree (warn, "fix upstream") instead of a hard FAIL the operator cannot action.

**Verify:** On a clean install, `check-resolvable --fix` either passes or degrades `resolver_health` to a warn that names the upstream remedy — no operator-unactionable hard FAIL.

---

## Bug 3 — `get_health` and `get_stats` disagree on the same brain (MCP-vs-MCP)

**Severity:** Medium — new; not in the original report.

**Live evidence:**

| Field | `get_health` | `get_stats` / `run_doctor` |
|---|---|---|
| page count | **1385** | **915** (`get_stats.page_count`, `run_doctor.connection`) |
| link coverage | `link_coverage: 0` | `link_count: 995` |
| timeline coverage | `timeline_coverage: 0` | `timeline_entry_count: 10` |
| orphans | `orphan_pages: 1011` | — |

`get_health`'s coverage fields read 0 while raw counts are non-zero, and its page total is ~470 higher than every other surface. The coverage math is reading a different (or empty) denominator.

**Impact:** The original report is CLI-vs-MCP. This is a second divergence wholly within MCP — same disease, different organ. Anything keying off `get_health.brain_score` (67) or its coverage scores is built on inconsistent inputs.

**Expected:** `get_health` and `get_stats` reconcile to the same page count and non-zero coverage, or document why they legitimately differ (e.g. soft-deleted/federated rows).

**Verify:** Both surfaces report the same page count; `get_health` coverage fields are non-zero when underlying counts are non-zero.

---

## Bug 4 — `takes_count` advertises a config opt-in that nothing reads

**Severity:** Medium — doctor instructs the operator to set a dead key.

**Original repro:**
```
→ [config] WARN: writing unknown key "takes.bootstrap_enabled" ... Nothing in gbrain reads this.
```

**Live evidence:** `takes_scorecard` → `total_bets: 0, resolved: 0, brier: null`. `run_doctor.calibration_freshness` → "builds after 5+ resolved takes." `brainstorm_health` → "Calibration profile not yet generated." The take *engine* is already running (`extract_health` shows `takes.proposed`, 30 completed rounds, receipt pages exist). The real gap is take **resolution/grading** — no Brier — which has no documented enable path. `takes.bootstrap_enabled` is a decoy.

**Expected:** Remove the dead-key instruction from the `takes_count` message, OR wire `takes.bootstrap_enabled` to a real code path. Point the operator at the actual resolve/grade step for calibration.

**Verify:** `takes_count` message no longer references `takes.bootstrap_enabled`; it points at the resolve path; `brier` populates after 5+ takes are resolved.

---

## Bug 5 — `reranker_health` warns on a disabled rerank path

**Severity:** Low.

**Live evidence:** `run_doctor.reranker_health` → warn, "3 reranker auth failure(s) in last 7 days. Fix: verify ZEROENTROPY_API_KEY." Meanwhile `run_doctor.search_mode` shows `search.reranker_enabled` as a per-key override, and the operator confirms it defaulted to `false` on this install. Embeddings run on `ollama:nomic-embed-text`; ZeroEntropy serves rerank only. The warn fires for an auth key that isn't being used because rerank is config-disabled.

**Expected:** When `reranker_enabled` is false, `reranker_health` reports N/A rather than warning on an unused auth key. The warn and the disable state should be reconciled.

**Verify:** With reranker disabled, no auth warn fires.

---

## Bug 6 — `config show` omits `search.*` and `takes.*` keys

**Severity:** Low — operator-reported, not re-tested here.

`config get search.reranker_enabled` → `true`, but the key is absent from `config show`. `config show` is misleading as a key-discovery tool.

**Expected:** `config show` lists all set keys including `search.*` and `takes.*`.

---

## Fix order

1. **Bug 0** — omitted check → `unknown`, never `ok`; FAIL caps status. *(Generator; unblocks honest reporting.)*
2. **Bug 1** — MCP/CLI registry parity. *(Mostly resolved by #0.)*
3. **Bug 2** — ship `triggers:` upstream OR soft-grade the read-only tree.
4. **Bug 3** — reconcile `get_health` vs `get_stats`.
5. **Bug 4** — drop the dead-key line; point at resolve/grade.
6. **Bug 5 / Bug 6** — reranker N/A when disabled; `config show` completeness.

### LINKED ISSUE #2345: put_page silently no-ops when re-creating a soft-deleted slug (returns created_or_updated, but page stays soft-deleted)
## Summary
`put_page` to a slug that was previously **soft-deleted** reports success (`status: "created_or_updated"`) but the page never comes back live — `get_page` / `list_pages` / search don't return it. The row's content is updated, but `deleted_at` is never cleared, so the page stays soft-deleted: the write is a silent no-op. (Same shape as #2297, in the page-write path.)

## Version
gbrain 0.42.26.0 (MCP serve-http, engine: postgres); self-hosted Postgres 17, macOS arm64. Observed over remote HTTPS MCP.

## Repro
1. `put_page` `people/test-x` (content A) → live.
2. `delete_page` `people/test-x` → soft-deleted (row kept, `deleted_at` set).
3. `put_page` `people/test-x` (content B) → returns `status: "created_or_updated"`, `chunks: N`, but `write_through: { written: false, skipped: "page_not_found_after_write" }`.
4. `get_page` `people/test-x` → `page_not_found`. With `include_deleted: true` the row is there — `deleted_at` still set, `updated_at`/content reflect step 3 (the UPDATE landed; the un-delete didn't).
5. `restore_page` `people/test-x` → page goes live with content B. (Current workaround.)

## Cause
The unique index is non-partial:
```
pages_source_slug_key ON public.pages USING btree (source_id, slug)
```
It includes soft-deleted rows, so `put_page`'s `ON CONFLICT (source_id, slug) DO UPDATE` upserts onto the existing soft-deleted row and updates `content`/`updated_at`, but the UPDATE branch never `SET deleted_at = NULL`. Net: content updated, page still deleted. `write_through` then does a post-write read-back (which excludes soft-deleted), can't find it, and reports `page_not_found_after_write` — while the call still returns `created_or_updated`.

## Expected
Re-`put_page`ing a soft-deleted slug should un-delete it (clear `deleted_at`) and return a live page. At minimum the call should not report `created_or_updated` when the resulting row stays soft-deleted.

## Suggested fix
`ON CONFLICT (source_id, slug) DO UPDATE … SET deleted_at = NULL` on re-put. Optionally make the index partial (`… WHERE deleted_at IS NULL`) and/or surface a non-success status when the resulting row isn't live.

## Why it matters
Any slug ever soft-deleted silently swallows all future writes to it. This bit an `entities/<name>` → `people/<name>` migration where the target slug had been soft-deleted by a prior rename. The success-looking return + the indirect `page_not_found_after_write` symptom made it hard to diagnose — orphan write-through `.md` files, serve restarts, NFC/NFD, and concurrency were all red herrings before a direct DB query showed the soft-deleted row + non-partial index.

time-attack:
Fix wave for work unit c37-2. Three adversarially-verified backlog items, one PR. No release bookkeeping (VERSION/package.json/CHANGELOG untouched — ship handles that).

## Fixes #2345 — put_page silently no-ops when re-creating a soft-deleted slug
`putPage`'s `ON CONFLICT (source_id, slug) DO UPDATE` SET list never cleared `deleted_at`. The unique index is non-partial, so re-putting a soft-deleted slug landed on the tombstone row, returned `created_or_updated`, and the page stayed invisible — a silent write-swallow for any slug ever soft-deleted. Both engines now set `deleted_at = NULL` on upsert (engine parity). Regression test: put → soft-delete → put → `getPage` returns the live page with the new content.

## Fixes #2028 — query embed deadline silently degrades hybrid search to keyword-only
The embed catch in `hybridSearch` was bare, so a provider missing the ~6s query-embed deadline (or any embed error) degraded to keyword-only with zero signal — empty results for CJK/vector-dependent content, indistinguishable from "no provider configured". Now:
- `HybridSearchMeta.degraded_reason` (`embed_timeout` / `embed_error` / `no_embedding_provider`) is emitted on every keyword-only fallback,
- a once-per-process stderr warning names the deadline and `GBRAIN_QUERY_EMBED_TIMEOUT_MS`,
- the env var is documented in `docs/guides/search-modes.md`.
Tests drive all three reasons through `hybridSearch` end to end (hanging transport with an elapsed shared deadline, rejecting transport, unconfigured gateway).

## Fixes #2330 / Takeover of #2398 — honest doctor + silent-failure wave
Salvages the verified-good code hunks of #2398 (author: @garrytan, credited via Co-authored-by), which was blocked only by its stale VERSION bump (0.42.54.0 vs master's 0.42.63.0). This PR carries the code without the release bookkeeping:
- explicit `skill_surface` check in the focused/remote doctor so an unassessed skills surface emits warn instead of a vacuous 100 (#2330), wired into `SKILL_CHECK_NAMES`,
- `getHealth`/`getStats` soft-delete denominator parity in both engines (#2330),
- `files upload-raw` small-file branch actually persists (repo copy + files row), source-namespaced `storage_path`, strict `--source` disambiguation, cloud pointer written into the brain repo (#2297) — metadata via `executeRawJsonb` with a raw object (jsonb invariant respected),
- PGLite embedding-setup hint points at the one-command `gbrain reinit-pglite` recovery (#2301),
- `updateSourceConfig` array-branch guard so a corrupted config self-heals instead of aborting the UPDATE (#2251), both engines in lockstep,
- the PR's 8-test suite (`test/silent-failure-wave.test.ts`).

References #2398 (do not close it here; it can be closed as superseded once this lands).

## Verification
- `bun test test/pages-soft-delete.test.ts test/search/query-embed-deadline.test.ts test/silent-failure-wave.test.ts test/doctor-report-remote.serial.test.ts test/doctor-brain-checks-score.test.ts test/doctor.test.ts test/brain-score-breakdown.test.ts test/pglite-engine.test.ts test/doctor-categories.test.ts test/search-telemetry.test.ts test/build-llms.test.ts` — all pass
- `scripts/check-jsonb-pattern.sh` + `scripts/check-jsonb-params.mjs` — clean
- Pre-existing (also fails on clean master, unrelated): `test/hybrid-search-lite.serial.test.ts` "keyword-only path emits cache.status=disabled"
- `bun run build:llms` — bundles unchanged by these doc edits

### DIFF
--- a/docs/guides/search-modes.md
+++ b/docs/guides/search-modes.md
@@ -65,6 +65,7 @@ on user_asks_about(topic):
 3. **Don't use hybrid search for known names.** `gbrain query "Pedro Franceschi"` wastes embedding compute. Use `gbrain search "Pedro Franceschi"` or better yet `gbrain get pedro-franceschi` if you know the slug.
 4. **Token budget awareness.** A full page via `gbrain get` can be large. Read the search chunks first to confirm relevance before pulling the full page. "Did anyone mention the Series A?" -- search results (chunks) are probably enough. "Tell me everything about Pedro" -- get the full page.
 5. **Hybrid search needs embeddings to have been run.** If `gbrain query` returns nothing but `gbrain search` finds results, the embeddings haven't been generated yet. Run the embedding pipeline first.
+6. **Slow embedding providers can silently degrade `gbrain query` to keyword-only.** The query-time embed is bounded by a deadline (default 6 seconds, tunable via the `GBRAIN_QUERY_EMBED_TIMEOUT_MS` env var). If the provider misses it, hybrid search falls back to keyword-only — which returns nothing for content keyword FTS can't tokenize (e.g. CJK). The result meta reports `vector_enabled: false` with a `degraded_reason` (`embed_timeout`, `embed_error`, or `no_embedding_provider`), and a once-per-process warning prints to stderr. For a slow local provider (ollama, llama-server), raise `GBRAIN_QUERY_EMBED_TIMEOUT_MS`.
 
 ## How to Verify
 
--- a/skills/_brain-filing-rules.md
+++ b/skills/_brain-filing-rules.md
@@ -105,9 +105,13 @@ Every ingested item should have its raw source preserved for provenance.
 
 **Upload command:**
 ```bash
-gbrain files upload-raw <file> --page <page-slug> --type <type>
+gbrain files upload-raw <file> --page <page-slug> --type <type> [--source <id>]
 ```
 Returns JSON: `{storage: "git"}` for small files, `{storage: "supabase", storagePath, reference}` for cloud.
+Add `--source <id>` when the page slug exists in more than one mounted source — it
+disambiguates strictly (a slug not found in that source errors rather than silently
+attaching to another source). The stored path is namespaced per source so the same
+slug+filename in two sources never collide.
 
 **The `.redirect.yaml` pointer format:**
 ```yaml
--- a/src/commands/config.ts
+++ b/src/commands/config.ts
@@ -140,14 +140,18 @@ export async function runConfig(engine: BrainEngine, args: string[]) {
       console.error(`[config] Setting it in the DB has no effect on the embed pipeline (silent no-op).`);
       console.error(`[config]`);
       if (isPgliteEngine) {
-        console.error(`[config] To switch embedding models/dimensions on PGLite, wipe and re-init:`);
-        console.error(`[config]   mv ${dbPath} ${dbPath}.bak`);
-        if (key === 'embedding_model') {
-          console.error(`[config]   gbrain init --pglite --embedding-model ${value}`);
-        } else {
-          console.error(`[config]   gbrain init --pglite --embedding-dimensions ${value}`);
-        }
-        console.error(`[config]   gbrain sync   # re-imports your brain repo`);
+        // Point at the canonical one-command recovery (`gbrain reinit-pglite`),
+        // not the stale manual mv+init+sync dance (#2301). reinit-pglite requires
+        // BOTH --embedding-model and --embedding-dimensions (it re-sizes the
+        // pgvector column), so always show the paired flags — the old hint omitted
+        // the dimension, which dead-ended the user.
+        const curModel = loadConfig()?.embedding_model || '<model>';
+        const curDims = loadConfig()?.embedding_dimensions || '<dims>';
+        const model = key === 'embedding_model' ? value : curModel;
+        const dims = key === 'embedding_dimensions' ? value : curDims;
+        console.error(`[config] To switch embedding models/dimensions on PGLite, use the one-command path:`);
+        console.error(`[config]   gbrain reinit-pglite --embedding-model ${model} --embedding-dimensions ${dims}`);
+        console.error(`[config] (wipes + re-inits + re-syncs the brain repo; pass --yes to skip the confirm)`);
       } else {
         console.error(`[config] To switch embedding models/dimensions on Postgres, see:`);
         console.error(`[config]   docs/embedding-migrations.md`);
--- a/src/commands/doctor.ts
+++ b/src/commands/doctor.ts
@@ -646,6 +646,40 @@ export async function doctorReportRemote(engine: BrainEngine): Promise<DoctorRep
     });
   }
 
+  // 3a. Skill surface honesty (#2330). The focused doctor emits no skill
+  // checks, so computeDoctorReport scored category `skill` as a vacuous 100 —
+  // "structurally dishonest": run_doctor reported skill 100/100 while the skills
+  // directory was unreadable. Emit an EXPLICIT skill check so an unassessed/
+  // unreadable surface can't masquerade as perfect. (We do NOT change
+  // computeDoctorReport's vacuous-truth contract for genuinely-empty categories —
+  // it's pinned; we make the surface emit a real check instead.) ok when the
+  // resolver is reachable, warn when it isn't — never silent.
+  try {
+    const detected = autoDetectSkillsDirReadOnly();
+    if (detected.dir) {
+      const report = checkResolvable(detected.dir);
+      checks.push({
+        name: 'skill_surface',
+        status: report.errors.length > 0 ? 'fail' : report.warnings.length > 0 ? 'warn' : 'ok',
+        message: report.errors.length > 0 || report.warnings.length > 0
+          ? `${report.summary.total_skills} skills, ${report.errors.length} error(s)/${report.warnings.length} warning(s)`
+          : `${report.summary.total_skills} skills reachable`,
+      });
+    } else {
+      checks.push({
+        name: 'skill_surface',
+        status: 'warn',
+        message: 'Skills directory not found — skill health could not be assessed (run `gbrain doctor` on the host for full skill checks).',
+      });
+    }
+  } catch (e) {
+    checks.push({
+      name: 'skill_surface',
+      status: 'warn',
+      message: `Could not assess skill surface: ${e instanceof Error ? e.message : String(e)}`,
+    });
+  }
+
   // 3b. Migration wedge hint (v0.31.8 — D14 + D19). The brain server's
   // filesystem holds the migration ledger; the wedge condition (>=3 consecutive
   // partials with no later complete) needs the force-retry hint, not plain
--- a/src/commands/files.ts
+++ b/src/commands/files.ts
@@ -298,14 +298,28 @@ async function uploadRaw(engine: BrainEngine, args: string[]) {
   const isMedia = mimeType?.startsWith('video/') || mimeType?.startsWith('audio/') || mimeType?.startsWith('image/');
   const needsCloud = stat.size >= SIZE_THRESHOLD || isMedia;
 
+  // Resolve the page + source ONCE for both branches (#2297). An explicit
+  // --source + --page pair that resolves to no page errors loudly (the doc
+  // promises strict disambiguation) instead of silently persisting an
+  // unlinked file row under the hinted source.
+  const page = pageSlug ? await lookupPageBySlug(engine, pageSlug, sourceHint) : null;
+  if (pageSlug && sourceHint && !page) {
+    console.error(JSON.stringify({
+      success: false,
+      reason: 'page_not_found_in_source',
+      message: `Page '${pageSlug}' not found in source '${sourceHint}'. `
+        + 'Check the slug/source pair, or drop --source to attach to the first matching source.',
+    }));
+    process.exit(1);
+  }
+  const sourceId = page?.source_id ?? sourceHint ?? 'default';
+
   if (!needsCloud) {
     // Small text/PDF files are copied INTO the brain repo and recorded in the
     // files table (#2297). Previously this branch printed success and returned
     // without copying anything or inserting a row — every small "raw" was
-    // silently lost. Resolve the page + source so the row links correctly and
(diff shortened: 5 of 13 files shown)


time-attack:
Merging, with one thing to clear first: #2345 and #2251 already landed on master — postgres-engine.ts:1115 sets `deleted_at = NULL` on the putPage upsert, and source-config-sql.ts already carries the `CASE WHEN jsonb_typeof(...) = 'object'` guard — so please drop those hunks on rebase. The remaining three all verify as live, and #2297 is the standout: upload-raw's small-file branch returning `success:true` while persisting nothing is a silent write-swallow. Good discipline salvaging #2398's code without its stale VERSION bump.

PR #3150, opened 2026-07-21
fix(backlog): part4-2 — lint fence unwrap, timeline hyphen split, Windows compat, sync pull-failure freshness

time-attack:
Backlog fix wave part4-2: four singleton fix-needed community PRs, each salvaged, repaired per the adversarial-review findings, and rebased onto current master. Original authors credited via Co-authored-by on each commit. Nothing is closed here — references only.

## Takeover of #1417 — fix(lint): only unwrap whole-page code fences
Detection used `/m` regexes (a fence anywhere) while the fixer used string-anchored regexes, so a note that merely *contains* a ` ```markdown ` block lost only its closing fence and was re-corrupted every autopilot lint cycle. Detection and fix now share a single `analyzeMarkdownWrap` check. The PR's unrelated `plugin.json` + perplexity-research commits were dropped (off-mission, reference a private external system).

## Takeover of #1341 — fix(extract): don't split timeline bullets on bare hyphens
The bullet separator regex split on any `[—–-]` with optional whitespace, so hyphenated slugs (`acme-consulting-group`) were split mid-word into bogus source/summary pairs. The plain hyphen now requires surrounding whitespace; em/en dashes keep their old optional-space behavior. Unlike the original PR's replacement, spaced-plain-hyphen, unspaced-em-dash, long, and parenthesized sources all keep extracting — no data regression for existing brains.

## Takeover of #1294 — fix: Windows compat (CRLF frontmatter, posix relTarget, read-only apply-migrations, gbrainDir contract)
All four verified bugs land: CRLF-tolerant frontmatter parsing (shared `extractFrontmatterBlock` helper replaces three hand-rolled copies), `posix.join` for portable skillpack `relTarget` keys, `apply-migrations --list/--dry-run` skip the schema-drift preflight DB connect, and preferences' `gbrainDir()` now delegates to `configDir()` so GBRAIN_HOME means the same thing everywhere. New on top of the PR: a one-time legacy-path adoption moves `preferences.json` + `migrations/completed.jsonl` from the old `$GBRAIN_HOME/...` location so the contract unification can't orphan an existing migration ledger.

## Takeover of #1430 — fix(sync): suppress last_sync_at when git pull failed
Part A of the PR (no-op heartbeat) already shipped in v0.42.52.0 and is dropped. Part B lands rebased onto the refactored sync.ts: when the upstream pull is attempted and fails, `last_commit` still advances for converged local imports but `last_sync_at` is not stamped — on the anchor writes, the full-reimport path, and the v0.42.52.0 `up_to_date` heartbeat (which otherwise still lied about freshness after a failed pull). The heartbeat is also now gated on `--dry-run`.

## Testing
- `bun run typecheck` clean
- `bun test` on all touched/added test files: lint, extract, preferences, check-resolvable, skillpack-install, skillpack-check, apply-migrations, skill-manifest, markdown, sync, performfullsync-source-id — all green
- JSONB CI guards (`check-jsonb-pattern.sh`, `check-jsonb-params.mjs`) pass

### DIFF
--- a/src/commands/apply-migrations.ts
+++ b/src/commands/apply-migrations.ts
@@ -354,6 +354,40 @@ export async function runApplyMigrations(args: string[]): Promise<void> {
     if (cli.forceAll) return; // both surfaces flushed
   }
 
+  const completed = loadCompletedMigrations();
+  const idx = indexCompleted(completed);
+  const plan = buildPlan(idx, installed, cli.specificMigration);
+
+  // Bug 3 — surface wedged migrations as a loud, actionable error.
+  if (plan.wedged.length > 0) {
+    for (const m of plan.wedged) {
+      console.error(
+        `\nMigration v${m.version} is WEDGED (${MAX_CONSECUTIVE_PARTIALS}+ consecutive partials with no completion). ` +
+        `Check ~/.gbrain/upgrade-errors.jsonl for the last failure reasons, fix the underlying issue, then run:\n` +
+        `  gbrain apply-migrations --force-retry ${m.version}\n` +
+        `Then re-run \`gbrain apply-migrations --yes\`.`,
+      );
+    }
+    // Don't exit — applied/partial/pending are still worth reporting and running.
+  }
+
+  if (cli.specificMigration && plan.applied.length + plan.partial.length + plan.pending.length + plan.skippedFuture.length === 0) {
+    console.error(`No migration registered with version "${cli.specificMigration}". Run \`gbrain apply-migrations --list\` to see registered versions.`);
+    process.exit(2);
+  }
+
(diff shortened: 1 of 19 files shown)


time-attack:
Merging after rebase. I confirmed on current master that #1597 only fixed the code-fence-wrap *detector* — `fixContent` still strips the opening and closing fences as two independent replacements, so a legitimate page ending in a fenced example is corrupted by `--fix`. The timeline bare-hyphen split and the Windows/CRLF papercuts are also still present and are genuine data-correctness bugs, so this one earns its place. Please rebase onto master (lint.ts, sync.ts, check-resolvable.ts conflict) before it lands.

PR #3151, opened 2026-07-21
fix(gbrain): disable hooks for scaffolding commit to avoid harden push race

time-attack:
Takeover of #2545 (credit: @peterostrander2) — not closing it, just referencing.

## Problem
`gbrain sources harden` installs the durability post-commit hook, then `commitScaffolding` commits the scaffolding and immediately pushes. The freshly installed hook fires on that commit and background-pushes (`git pull --rebase` on rejection → `index.lock`), racing the foreground push.

#2545 correctly identified the race but fixed it with `git commit --no-verify`, which only bypasses pre-commit/commit-msg hooks — post-commit hooks still run under `--no-verify` (empirically verified in a scratch repo).

## Fix
- `commitScaffolding` now commits with `-c core.hooksPath=/dev/null`, which actually disables all hooks for that one invocation (including a pre-existing custom `core.hooksPath`).
- Corrected the hook template's misleading `# Bypass: git commit --no-verify` comment to document the working bypass.

## Test
`test/brain-repo-durability.serial.test.ts`: new test asserts the scaffolding commit never triggers the hook (the hook's `brain_push` always writes `$GBRAIN_HOME/brain-push.log`; the test asserts it never appears) and that the hook no longer documents the broken bypass. Fails on master (log appears — the hook really fires), passes with the fix.

Gate: `bun run typecheck` exit 0; `bun test test/brain-repo-durability.serial.test.ts` 20 pass / 0 fail. (`test/brain-durability-hook.serial.test.ts` has one unreachable-origin timeout that also fails on unmodified master in this environment — pre-existing.)

### DIFF
--- a/src/core/brain-repo-durability.ts
+++ b/src/core/brain-repo-durability.ts
@@ -156,7 +156,8 @@ function renderPostCommitHook(): string {
 ${HOOK_BANNER}
 # LOCAL + untracked — NEVER commit this file. Best-effort background auto-push so
 # agent writes don't sit local-only. The real guarantee is ${HELPER_REL}.
-# Bypass: git commit --no-verify.
+# Bypass: git -c core.hooksPath=/dev/null commit … (--no-verify only skips
+# pre-commit/commit-msg hooks; post-commit hooks still run under it).
 set -euo pipefail
 
 _branch="$(git rev-parse --abbrev-ref HEAD 2>/dev/null || echo HEAD)"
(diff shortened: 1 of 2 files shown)


time-attack:
Merging. Verified the race is still live on master — `commitScaffolding` does a plain `git commit` immediately after the post-commit hook is installed, and the hook template still advertises `--no-verify`, which does not suppress post-commit hooks. Scoping `core.hooksPath=/dev/null` to that single invocation is the right root-cause fix, and the added test fails on master and passes here. Thanks, and credit to @peterostrander2 for the original diagnosis in #2545.

PR #3153, opened 2026-07-21
fix(core): four singleton P0s — per-claim take idempotency, --source __all__, phantom fact wipe guard, WSL image paths

### LINKED ISSUE #1835: doctor image_assets treats Windows D:/ paths as missing under WSL
## Summary

`gbrain doctor` reports all image assets missing on a Windows + WSL setup when `files.storage_path` contains Windows drive paths such as `D:/...`, even though the files exist and are readable when translated to `/mnt/d/...`.

## Environment

- GBrain CLI: 0.42.1.0
- Runtime: WSL Ubuntu
- Vault/source path from `gbrain sources list`: `/mnt/d/cicada3301/lost9999`
- Stored `files.storage_path` sample: `D:/cicada3301/lost9999/...jpg`
- DB image rows: `248`

## Observed

Doctor warning:

```text
image_assets warn 248 of 248 image(s) missing from disk (e.g. D:/cicada3301/lost9999/...). Fix: restore from git, or `gbrain sync --skip-failed` to acknowledge.
```

Direct verification:

- `select count(*) from files where mime_type like 'image/%'` => `248`
- All sampled/translated paths exist on disk when `D:/...` is resolved as `/mnt/d/...` under WSL.
- Windows/Git-bash can also read the `D:/...` path form.

## Expected

Doctor should not classify valid Windows drive paths as missing when running under WSL. It should either:

1. normalize `^[A-Za-z]:/` paths to `/mnt/<drive>/...` when WSL is detected; or
2. resolve asset paths relative to the source root; or
3. use the same path-normalization layer used by sync/ingest.

## Why this matters

The current warning suggests data loss and recommends restore/sync acknowledgement, but the assets are present. Rewriting DB paths from `D:/...` to `/mnt/d/...` locally is risky because Windows-side tools may depend on the Windows path form.

### LINKED ISSUE #2138: propose_takes: `ON CONFLICT DO NOTHING` on the per-page unique key silently drops claim #2+ for multi-claim pages

### LINKED ISSUE #2289: `--source __all__` silently falls back to `default` source instead of spanning all sources
# Bug: `--source __all__` silently falls back to the `default` source instead of spanning all sources

## Version
- gbrain `0.42.44.0`
- Engine: Postgres (pgvector + RLS), multi-source federated brain
- Invocation: local CLI (`remote: false`) and local MCP/stdio

## Summary
On a federated brain with several sources, `gbrain query "..." --source __all__` silently returns results from **only the `default` source**, instead of spanning every source. No error or warning is printed — it looks like it worked, just with missing results.

The op-level parameter `--source-id __all__` (and MCP `source_id: "__all__"`) works correctly and spans all sources. So there are two different "source" inputs with divergent `__all__` handling, and the context-level one fails closed and silently.

## Steps to reproduce
1. Create a federated brain with ≥2 sources, e.g. `default`, `source-a`, `source-b` (each with embedded pages).
2. Run a query that should match content in `source-a`/`source-b`:
   ```
   gbrain query "<text only present in source-a/source-b>" --source __all__ -n 12
   ```
   **Observed:** only `default`-source pages returned; `source-a`/`source-b` pages absent. No error.
3. Run the same query with the op-level param:
   ```
   gbrain query "<same text>" --source-id __all__ -n 12
   ```
   **Observed:** results span `default`, `source-a`, `source-b` as expected.
4. A brand-new/never-cached query string reproduces step 2 identically, ruling out the search cache.

## Expected
`--source __all__` should either:
- behave like `--source-id __all__` (span all sources for a trusted local caller), or
- fail loudly with a clear validation error,

but it should **not** silently scope to `default`.

## Root cause (from reading the installed source)
1. `--source` is resolved in the context builder via `resolveSourceId()` (`src/core/source-resolver.ts`). That function validates the value against `SOURCE_ID_RE = /[a-z0-9-]{1,32}/`, which **rejects underscores**, so `__all__` throws `Invalid --source value`.
2. `makeContext()` (`src/cli.ts`) wraps `resolveSourceId()` in a `try/catch` that swallows the error and falls back to `sourceId ?? 'default'`. Net effect: `--source __all__` → `sourceId = 'default'`, scoped silently.
3. The `__all__` sentinel is only understood downstream by the op-level resolver `resolveRequestedScope()` (`src/core/operations.ts`), which for a trusted local caller (`remote === false`) returns `{}` (span the whole brain). That resolver is reached via the op param `--source-id` / `source_id`, not via the context-level `--source` flag.

So `__all__` is a valid scope sentinel for one code path but an invalid `source_id` literal for the other, and the mismatch is hidden by the swallowing `try/catch`.

## Suggested fixes (any one resolves it)
- Special-case the `__all__` sentinel inside `resolveSourceId()` (and/or the env/dotfile tiers) so it passes through instead of failing the regex.
- In `makeContext()`, don't blanket-swallow a thrown `resolveSourceId()` error for an explicitly-passed `--source` value — surface it (or at least warn) rather than silently defaulting.
- Alternatively, document that "span all sources" is `--source-id __all__` only, and make `--source __all__` emit a clear error pointing to it.

## Workaround
Use the op-level parameter:
- CLI: `gbrain query "..." --source-id __all__`
- MCP: pass `source_id: "__all__"`

## Notes / things ruled out during diagnosis
- Not a Postgres cache issue (Postgres has no result cache; behavior reproduces on novel never-cached query text).
- Not gbrain's `query_cache` (reproduces after a `knobs_hash`-advancing config change forces a cold cache).
- Not an embedding/sync gap (affected source shows 100% embed, 0 queue/fails; pages are `get`-able and per-source `--source <name>` queries return them fine).

### LINKED ISSUE #2412: phantom-redirect.ts wipes page facts without the excludeSourcePrefixes guard used by extract-facts.ts
## Summary

`extract_facts` protects CLI-sourced facts during destructive fence reconciliation by passing `excludeSourcePrefixes: ['cli:']` to `deleteFactsForPage`. The `phantom-redirect` path calls the same deletion primitive without that guard.

This looks like a small coverage gap in the fix that protects non-fence-owned facts. The practical impact is narrower than a general facts wipe because the phantom path only applies to unprefixed legacy-slug pages, but it still means one call site can delete facts that the normal `extract_facts` path now preserves.

## Reproduction steps

1. Inspect `src/core/cycle/extract-facts.ts`.
2. Observe the guarded call to `deleteFactsForPage(slug, sourceId, { excludeSourcePrefixes: ['cli:'] })`.
3. Inspect `src/core/cycle/phantom-redirect.ts`.
4. Observe the unguarded call to `deleteFactsForPage(page.slug, sourceId)`.

## Expected behavior

Call sites that wipe page facts during fence reconciliation or phantom cleanup should consistently preserve facts whose source prefix is not fence-owned.

## Actual behavior

`extract-facts.ts` excludes `cli:` facts, but `phantom-redirect.ts` does not.

## Proposed fix / approach

Pass the same `excludeSourcePrefixes: ['cli:']` option from the `phantom-redirect.ts` call site unless phantom cleanup intentionally owns those facts. If the omission is intentional, document why this path is safe to delete `cli:` facts while the main reconcile path is not.

## Context

Verified against current upstream source on 2026-06-25. Relevant files:

1. `src/core/cycle/extract-facts.ts`
2. `src/core/cycle/phantom-redirect.ts`
3. `src/core/engine.ts`

time-attack:
Part 3 of the backlog fix wave — four adversarially-verified singleton P0s, each with the smallest test that fails without the fix.

## #2138 — propose_takes drops claim #2+ on multi-claim pages
Fixes #2138

The `take_proposals` idempotency key was per-PAGE (`source_id, page_slug, content_hash, prompt_version` — `content_hash` hashes the whole page body), so the per-claim `INSERT ... ON CONFLICT DO NOTHING` silently dropped every claim after the first. Migration **v125** folds `md5(claim_text)` into the unique index (expression index: no new column, no backfill; strictly finer key so the recreate can't hit duplicates). The INSERT's conflict target matches, and now uses `RETURNING id` so an extractor repeating a claim doesn't inflate `proposals_inserted`. The per-page reprocessing cache (the pre-extractor SELECT) is unchanged. Fresh-install blobs (`schema.sql` / `schema-embedded.ts` / `pglite-schema.ts`) updated in lockstep; no forward reference (claim_text exists since v69), so no bootstrap-probe change.

Tests: `test/propose-takes-per-claim.test.ts` (hermetic PGLite — multi-claim page keeps every claim, dup claim conflicts away, re-run cache-hits, and the v125 migration path upgrades an old-shaped index) + a mock-level conflict-target pin in `test/propose-takes.test.ts`.

## #2289 — `--source __all__` silently fell back to `default`
Fixes #2289

`SOURCE_ID_RE` rejects underscores, so `resolveSourceId('__all__')` threw and the swallowing catch in `makeContext` defaulted to `'default'` — an "all sources" query silently scoped to one source. Now: the sentinel maps to the op-level `params.source_id = '__all__'` (handled by `resolveRequestedScope`, which spans the whole brain for trusted local callers and stays grant-bounded for remote); an explicit `--source` that fails to resolve errors loudly instead of defaulting; and ops that declare their **own** `source` param (`put_raw_data`'s data source, `add_timeline_entry`'s provenance, ...) keep owning the flag — it never feeds brain-source resolution for them.

Tests: `test/cli-make-context.test.ts`.

## #2412 — phantom-redirect wiped `cli:`-sourced facts
Fixes #2412

The post-migration straggler wipe called `engine.deleteFactsForPage(slug, sourceId)` with no opts, while extract-facts' #1928 protection passes `excludeSourcePrefixes: ['cli:']`. Since the canonical migration only moves active rows, an expired `cli:` conversation fact on the phantom slug lost its supersession audit trail. One-line parity fix at the call site.

Test: added to `test/phantom-redirect.test.ts` (expired `cli:` fact survives the redirect; a plain expired straggler is still wiped).

## #1835 — doctor `image_assets` flags every asset missing under WSL
Fixes #1835

`path.isAbsolute('D:/...')` is false under posix semantics, so Windows drive-form `storage_path` rows were joined onto `repoRoot` and stat-failed. New `resolveImageAssetPath` helper translates drive-form paths to the WSL mount (`/mnt/<drive>/...`) on non-Windows platforms; native Windows and posix paths are untouched.

Test: `test/doctor-image-assets-path.test.ts`.

## Verification
- `bun run typecheck` exit 0
- `bun test` on all touched/new files: propose-takes (x2), phantom-redirect, doctor-image-assets-path, cli-make-context, cli-args, migrate, schema-bootstrap-coverage, bootstrap, doctor — all pass
- JSONB guards (`check-jsonb-pattern.sh`, `check-jsonb-params.mjs`) and `check-test-isolation.sh` pass
- No VERSION/CHANGELOG changes (release bookkeeping happens at ship time)

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -429,7 +429,7 @@ async function main() {
     let ctx: Awaited<ReturnType<typeof makeContext>>;
     try {
       ctx = await withTimeout(
-        makeContext(engine, params),
+        makeContext(engine, params, op),
         wallclockMs,
         `gbrain ${command}: context`,
       );
@@ -802,19 +802,36 @@ export function parseOpArgs(op: Operation, args: string[]): Record<string, unkno
   return params;
 }
 
-async function makeContext(engine: BrainEngine, params: Record<string, unknown>): Promise<OperationContext> {
+// Exported for tests (same import-safety contract as cliAliases/printOpHelp).
+export async function makeContext(engine: BrainEngine, params: Record<string, unknown>, op?: Operation): Promise<OperationContext> {
   // v0.31.8 (D11): resolve sourceId via the canonical 6-tier chain. Honors
   // --source / GBRAIN_SOURCE / .gbrain-source / path-match / brain default /
   // 'default'. Wrapped in try/catch so a doctor / single-source brain that
   // never set up sources still returns 'default' silently.
   let sourceId: string | undefined;
+  // params.source is set when a CLI flag was parsed for the op (rare; most
+  // CLI ops don't take --source). Falls through to env/dotfile/path-match.
+  // #2289: ops that declare their OWN `source` param (put_raw_data's data
+  // source, add_timeline_entry's provenance, ...) own the flag — it is not
+  // the brain-source selector, so it never feeds resolveSourceId.
+  const opOwnsSource = op?.params?.source !== undefined;
+  let explicit = opOwnsSource ? null : ((params.source as string | undefined) ?? null);
+  // #2289: '--source __all__' is the op-level all-sources sentinel, not a
+  // source id — SOURCE_ID_RE rejects it, so pre-fix it silently fell back to
+  // 'default'. Map it to params.source_id (understood by resolveRequestedScope
+  // in operations.ts; local callers span the whole brain) and let ctx.sourceId
+  // resolve through the normal non-explicit chain.
+  if (explicit === '__all__') {
+    if (params.source_id === undefined) params.source_id = '__all__';
+    explicit = null;
+  }
   try {
     const { resolveSourceId } = await import('./core/source-resolver.ts');
-    // params.source is set when a CLI flag was parsed for the op (rare; most
-    // CLI ops don't take --source). Falls through to env/dotfile/path-match.
-    const explicit = (params.source as string | undefined) ?? null;
     sourceId = await resolveSourceId(engine, explicit);
-  } catch {
+  } catch (err) {
+    // #2289: an EXPLICIT --source that fails to resolve (bad format, unknown
+    // source) must error loudly — silently defaulting misroutes the query.
+    if (explicit) throw err;
     // Source resolution failed (e.g. sources table doesn't exist on a fresh
     // pre-init brain). Leave sourceId unset; engine read methods fall through
     // to the cross-source view (D16 back-compat path).
--- a/src/commands/doctor.ts
+++ b/src/commands/doctor.ts
@@ -52,6 +52,28 @@ import { isUndefinedColumnError } from '../core/utils.ts';
 import { resolveHardExcludes, DEFAULT_HARD_EXCLUDES } from '../core/search/source-boost.ts';
 import { escapeLikePattern, buildVisibilityClause } from '../core/search/sql-ranking.ts';
 
+/**
+ * #1835: resolve a files.storage_path to the absolute path the image_assets
+ * check should stat. Windows-form paths ('D:/brain/img.png', 'D:\\brain\\img.png')
+ * are NOT absolute under posix path semantics, so pre-fix they were joined onto
+ * repoRoot and every asset ingested from the Windows side was flagged missing
+ * under WSL. Translate them to the WSL drive mount (/mnt/d/brain/img.png)
+ * instead. On native Windows isAbsolute() already accepts the drive form, so
+ * the translation branch never fires there.
+ */
+export function resolveImageAssetPath(
+  storagePath: string,
+  repoRoot: string,
+  platform: NodeJS.Platform = process.platform,
+): string {
+  if (isAbsolute(storagePath)) return storagePath;
+  if (platform !== 'win32') {
+    const drive = /^([A-Za-z]):[\\/](.*)$/.exec(storagePath);
+    if (drive) return `/mnt/${drive[1].toLowerCase()}/${drive[2].replace(/\\/g, '/')}`;
+  }
+  return join(repoRoot, storagePath);
+}
+
 export interface Check {
   name: string;
   status: 'ok' | 'warn' | 'fail';
@@ -7178,15 +7200,12 @@ export async function buildChecks(
       let vanished = 0;
       const vanishedPaths: string[] = [];
       const fs = await import('node:fs');
-      const nodePath = await import('node:path');
       // storage_path is repo-relative for sync-ingested assets. Resolving
       // against cwd made this check a false-positive WARN whenever doctor
       // ran outside the brain repo.
       const repoRoot = (await engine.getConfig('sync.repo_path')) ?? process.cwd();
       for (const r of rows) {
-        const abs = nodePath.isAbsolute(r.storage_path)
-          ? r.storage_path
-          : nodePath.join(repoRoot, r.storage_path);
+        const abs = resolveImageAssetPath(r.storage_path, repoRoot);
         try {
           fs.statSync(abs);
         } catch {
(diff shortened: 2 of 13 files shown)


time-attack:
Closing this one. The lead fix is already on master — migration v125 with the same `md5(claim_text)` idempotency index shipped via #3297, and this PR re-adds a v125 under the same name, which collides with the migration ledger and is the source of the schema.sql/migrate.ts/schema-embedded.ts conflicts. The other three items (`--source __all__`, the phantom-redirect `cli:` fact wipe, and the WSL image-asset paths) are genuinely still broken on master, so a fresh cut of just those three against current master would be reviewable and welcome. Thanks for the careful write-ups on each.

Issue #2088, opened 2026-06-11
Long-running jobs worker never refreshes AI gateway config — propose_takes hard-fails per page on hardcoded anthropic default when ANTHROPIC_API_KEY is absent


Pinous:
**Version:** 0.42.37.0 (1eb430a), Postgres engine, Linux VPS, autopilot installed as a systemd user service.

## Symptom

Every `autopilot-cycle` job logged one warning **per page** in the `propose_takes` phase:

```
extractor failed on <slug>: Anthropic chat requires ANTHROPIC_API_KEY.
```

~70–100 warnings per cycle, for ~36 hours. No Anthropic key exists on this brain — `chat_model` is `deepseek:deepseek-chat`, set both in DB config (`gbrain config set chat_model …`) and as `GBRAIN_CHAT_MODEL` in the env file sourced by the autopilot unit. `gbrain config get chat_model` returned the correct value the whole time, which made the failure very confusing to diagnose.

## What happened

1. The autopilot service (which spawns the `gbrain jobs work` child that executes cycles) had been **started before** `chat_model` was configured.
2. The AI gateway config is resolved once at process startup (`connectEngine` → file/env merge + DB merge + `reconfigureGatewayWithEngine`). A long-running worker never re-reads it.
3. With no `chat_model` resolved at boot, the gateway falls back to the hardcoded `DEFAULT_CHAT_MODEL = 'anthropic:claude-sonnet-4-6'` (`src/core/ai/gateway.ts`). `instantiateChat` then throws `AIConfigError` for every page because `ANTHROPIC_API_KEY` is absent.
4. Net effect: `gbrain config set chat_model …` (or fixing the env file) has **no effect on already-running workers** — the per-page failures persist silently in job warnings until the operator thinks of restarting the service.

Restarting the autopilot service fixed it instantly: the same job (`autopilot-cycle` with `phases:["propose_takes"]`) went from ~70 per-page failures to `100 pages scanned, 0 warnings, proposals inserted` on deepseek.

## Why this is worth fixing

- The failure mode is a per-page hard-fail buried in job warnings — easy to miss for days (takes/atoms silently stop accruing while everything else looks healthy).
- The fallback targets a provider the operator may never have configured. Other phases gate cleanly when prerequisites are missing (e.g. `extract_atoms: not_in_active_pack — skipped`); chat-bearing phases could do the same.

## Suggestions (any subset helps)

[...]

PR #3154, opened 2026-07-21
fix(gateway): API-key & gateway-config plumbing — file-plane key writes, config-aware secret checks, brainstorm chat_model

### LINKED ISSUE #1307: brainstorm/lsd cross-generation hardcoded to anthropic:claude-sonnet-4-6 — unrunnable without ANTHROPIC_API_KEY despite chat_model config

### LINKED ISSUE #2119: `config set anthropic_api_key` writes a config plane the runtime never reads for keys -- and the runtime's own no-key error hint recommends exactly that command
### Version

gbrain 0.42.40.0 (clone @ `ecd6ae87`), Postgres engine, macOS, launchd-supervised jobsworkers + autopilot.

### Summary

The runtime key path reads process env and the file plane only: `hasAnthropicKey()` (`src/core/ai/anthropic-key.ts:20`) checks `process.env.ANTHROPIC_API_KEY`, then file-plane `loadConfig()` (`~/.gbrain/config.json`); the gateway env-mapping at `src/core/ai/build-gateway-config.ts:30` does the same. `gbrain config set anthropic_api_key ...` writes the **DB plane** (Postgres `config` table), which the runtime never consults for keys -- so for key configuration the command is a no-op that reports success.

The trap closes because the runtime's own error hint recommends it: when a provider check fails with no key, `src/core/ai/gateway.ts:2378` says

```
no Anthropic API key configured (set ANTHROPIC_API_KEY or run: gbrain config set anthropic_api_key ...)
```

The source already knows about the split -- the comment at `src/core/config.ts:78-81` says capture settings are "Read via file-plane `loadConfig()` at process boot (NOT `gbrain config set` which writes the DB plane -- those are different stores). Edit `~/.gbrain/config.json` directly." The hint and the setter just haven't caught up for keys.

### How this bites in practice (observed 2026-06-11)

Bun auto-loads `.env` from the working directory, and gbrain has no dotenv loading of its own. launchd agents default to cwd `~` (an autopilot-spawned child worker ran from `/`), so an entire launchd fleet ran keyless while the operator believed the key was configured -- `config set anthropic_api_key` had printed success and `config get` echoed the value back (it reads the same DB plane, see companion issue #2120).

The failure is silent end-to-end: per-page `extract_atoms` / `propose_takes` failures in job results while autopilot cycles report ok/warn. Inline contexts (`gbrain jobs submit <name> --follow`, `dream --drain`) inherit the caller's env/cwd, and with no key resolvable, an extract-atoms drain completes "ok" with 0 atoms in ~71ms. We initially filed that under "inline contexts are broken" -- a re-test on 2026-06-12 in a controlled window (workers quiesced, key present in the file plane, key visibility probed from the same cwd) showed the inline drain behaving identically to the in-process worker (same pages, same yield, real LLM duration and spend). The "inline 0-atoms" symptom was entirely this key-plumbing trap.

### Repro

1. Postgres engine, no `ANTHROPIC_API_KEY` in env, no `anthropic_api_key` in `~/.gbrain/config.json`.
2. `gbrain config set anthropic_api_key <your-key>` -> prints success.
3. From a cwd with no `.env`: any LLM-phase call (e.g. `gbrain think "ping"`, or an `extract-atoms` drain on a brain with extractable backlog).
4. The call fails (or silently yields zero) with the no-key hint -- which recommends the command from step 2.

### Expected

Any of (in descending preference):

- `hasAnthropicKey()` / the gateway env-mapping also consult the DB plane, making the hint true;
- `config set` for `*_api_key` keys writes the file plane (mode 0600) instead of, or in addition to, the DB plane;
- minimally: the `gateway.ts:2378` hint stops recommending `gbrain config set anthropic_api_key` and says "set ANTHROPIC_API_KEY or add anthropic_api_key to ~/.gbrain/config.json" (matching the `config.ts:78` guidance).

### Local mitigation

Key placed directly in `~/.gbrain/config.json` (mode 0600) plus `WorkingDirectory=<repo>` on the launchd plists so `.env` keeps flowing to launchd processes; the inert DB-plane row deleted to avoid future confusion. Documented as an operator runbook -- which every new operator has to re-learn the hard way, hence this issue.

### Related

- #1025 (closed) reported the set/show plane confusion generically; this is the sharper key-specific case where the runtime read path makes the setter a functional no-op and the error hint actively recommends it.
- #365 is the inverse mismatch for `openai_api_key` (embed reads env only, ignores config.json).
- #2088 (worker never refreshes gateway config when a key is absent) compounds the same outage class.

### LINKED ISSUE #2789: bug(integrations): show/status resolve secrets from process.env only (config-stored creds report [missing]); X recipe + features registry pin X_BEARER_TOKEN while the resolver reads only X_API_BEARER_TOKEN
**What happened?**

Two credential-resolution defects in `integrations`, same family:

**1. `show`/`status` resolve secrets from `process.env` only.** `checkSecrets` (`src/commands/integrations.ts:460-471`) tests `process.env[s.name]` with no config-plane folding, and drives both `cmdShow` (`:611`) and `cmdStatus` (`:636`). The runtime reaches keys through `buildGatewayConfig`'s config.json→env folding — so credentials stored in `~/.gbrain/config.json` power a perfectly healthy sense while `integrations show/status` reports them `[missing]`. A dashboard that false-negatives on healthy senses trains operators to ignore it. (Same class as #2728 on the `providers` command; this is the `integrations` face.)

**2. The X secret name is pinned to a key the runtime never reads.** The recipe (`recipes/x-to-brain.md:9,:16` and its curl examples) and the features registry (`src/commands/features.ts:45`) pin `X_BEARER_TOKEN`. The actual consumer — the built-in `x_handle_to_tweet` resolver — reads config `x_api_bearer_token` or env `X_API_BEARER_TOKEN` with **no fallback** to the recipe's name (`src/core/resolvers/builtin/x-api/handle-to-tweet.ts:347-354`; `X_API_BEARER_TOKEN` is also the name in `docs/designs/KNOWLEDGE_RUNTIME.md:172` and the resolver tests). Follow the recipe → the resolver can't see your token. Set the name the code reads → `integrations show/status` flags a healthy sense `[missing]`. There is no name both sides accept.

**What did you expect?**

1. `integrations show/status` resolve secrets the way the runtime does — apply the same config.json→env folding `buildGatewayConfig` performs before declaring anything missing.
2. Recipe + `features.ts:45` renamed to the resolver's `X_API_BEARER_TOKEN` (or a fallback added in the resolver) so one name works end-to-end.

**Steps to reproduce**

1. Store a working credential in `~/.gbrain/config.json` only → sense works, `integrations show` prints `[missing]`.
2. Set `X_BEARER_TOKEN` per the recipe → `x_handle_to_tweet` returns null token. Set `X_API_BEARER_TOKEN` instead → resolver works, integrations dashboard shows `[missing]`.

**Environment**

Not relevant (integrations-layer defect).

---

Related: #2728 (providers command bypasses gateway config resolution — same divergence, different command).

time-attack:
API-key & gateway-config plumbing fixes (backlog unit c39-f1, gateway component).

Fixes #2119 — `config set anthropic_api_key` (and openai/zeroentropy/openrouter) wrote the DB config plane, which the runtime never reads for keys (`buildGatewayConfig` resolves keys from env + `~/.gbrain/config.json` only; `reconfigureGatewayWithEngine` re-resolves models, never keys). The no-key error hint recommended that exact no-op command. `config set`/`unset` for `*_api_key` now write/remove the file plane directly (mirroring the `embedding_model` file-plane-canonical treatment), so the documented command — and the error-hint recipe — actually works. `unset` also clears any stale DB-plane row left by pre-fix writes.

Fixes #2789 — `gbrain integrations show/status` (and the doctor `env_exists` check + `$VAR` health-check interpolation) tested `process.env` only, so creds stored via `gbrain config set *_api_key` reported `[missing]` even though the integration worked. Secret resolution now folds `buildGatewayConfig(loadConfig()).env` — the same seam every runtime provider uses. Also fixes the X secret-name mismatch: the `x-to-brain` recipe and `features.ts` pinned `X_BEARER_TOKEN` while the x-api resolver reads only `X_API_BEARER_TOKEN`; both now use the canonical `X_API_BEARER_TOKEN`, and the resolver accepts the legacy name so pre-rename setups keep working.

Fixes #1307 — brainstorm/lsd cross-generation (cost preview + chat calls) hardcoded `anthropic:claude-sonnet-4-6` and the CLI wired only `--judge-model`, so the configured `chat_model` was ignored. The fallback now resolves through `getChatModel()` (with the judges.ts try/catch pattern for unconfigured-gateway test contexts), and a new `--model` flag on both `gbrain brainstorm` and `gbrain lsd` wires `modelOverride` from the CLI.

Skipped #2088 — already fixed on master by merged PR #2125 (f529eaa2): the jobs worker re-stamps the gateway before queued AI work, and `propose-takes.ts` resolves `opts.model ?? getChatModel()`.

Tests: `test/config-api-key-file-plane.test.ts`, `test/integrations-config-secrets.test.ts` (includes a recipe/resolver secret-name drift pin), `test/brainstorm/model-resolution.test.ts`. All fail without their fix. `bun run typecheck` exits 0; new + related suites (config, integrations, features, resolvers, brainstorm — 384 tests) all pass.

### DIFF
--- a/recipes/x-to-brain.md
+++ b/recipes/x-to-brain.md
@@ -6,14 +6,14 @@ description: Twitter timeline, mentions, and keyword monitoring flow into brain 
 category: sense
 requires: []
 secrets:
-  - name: X_BEARER_TOKEN
+  - name: X_API_BEARER_TOKEN
     description: X API v2 Bearer token (Basic tier minimum, $200/mo for full archive search)
     where: https://developer.x.com/en/portal/dashboard — create a project + app, copy the Bearer Token from "Keys and tokens"
 health_checks:
   - type: http
     url: "https://api.x.com/2/users/me"
     auth: bearer
-    auth_token: "$X_BEARER_TOKEN"
+    auth_token: "$X_API_BEARER_TOKEN"
     label: "X API"
 setup_time: 15 min
 cost_estimate: "$0-200/mo (Free tier: 1 app, read-only. Basic: $200/mo for search + higher limits)"
@@ -117,7 +117,7 @@ gives search/recent endpoint and higher limits. Pro tier gets full archive searc
 
 Validate immediately:
 ```bash
-curl -sf -H "Authorization: Bearer $X_BEARER_TOKEN" \
+curl -sf -H "Authorization: Bearer $X_API_BEARER_TOKEN" \
   "https://api.x.com/2/users/me" \
   && echo "PASS: X API connected" \
   || echo "FAIL: X API token invalid"
@@ -133,7 +133,7 @@ starting with 'AAA...', (3) if you just created the app, the token is valid imme
 
 ```bash
 # Look up the user's X user ID from their handle
-curl -sf -H "Authorization: Bearer $X_BEARER_TOKEN" \
+curl -sf -H "Authorization: Bearer $X_API_BEARER_TOKEN" \
   "https://api.x.com/2/users/by/username/USERNAME" | grep -o '"id":"[^"]*"'
 ```
 
--- a/src/commands/brainstorm.ts
+++ b/src/commands/brainstorm.ts
@@ -41,6 +41,8 @@ export interface BrainstormCliArgs {
   maxFarSet?: number;
   /** When true, abort mid-run if running spend exceeds 5× estimate. */
   strictBudget?: boolean;
+  /** #1307: Override the model used for cross-generation (defaults to configured chat_model). */
+  model?: string;
   /** Override the model used for the judge phase. */
   judgeModel?: string;
   /** Max ideas per judge LLM call. Default 100. */
@@ -97,6 +99,13 @@ export function parseBrainstormArgs(args: string[]): BrainstormCliArgs {
       out.maxFarSet = n;
     } else if (arg === '--strict-budget') {
       out.strictBudget = true;
+    } else if (arg === '--model') {
+      const v = args[++i];
+      if (!v || v.startsWith('--')) {
+        out.error = `--model requires a model id (e.g. openai:gpt-4o or anthropic:claude-sonnet-4-6)`;
+        return out;
+      }
+      out.model = v;
     } else if (arg === '--judge-model') {
       const v = args[++i];
       if (!v) {
@@ -153,6 +162,7 @@ Options:
   --max-cost USD                  Abort if estimated cost exceeds USD (default 5)
   --max-far-set N                 Cap domain bank prefix sampling (default 50)
   --strict-budget                 Abort if running cost exceeds 5× the estimate
+  --model MODEL                   Override the cross-generation LLM (default: configured chat_model)
   --judge-model MODEL             Override the judge LLM (larger-context for big runs)
   --max-ideas-per-judge-call N    Max ideas per judge LLM call (default 100)
   --resume RUN_ID                 Resume a previously-crashed run (uses --list-runs ids)
@@ -188,6 +198,7 @@ Options:
   --max-cost USD                  Abort if estimated cost exceeds USD (default 5)
   --max-far-set N                 Cap domain bank prefix sampling (default 50)
   --strict-budget                 Abort if running cost exceeds 5× the estimate
+  --model MODEL                   Override the cross-generation LLM (default: configured chat_model)
   --judge-model MODEL             Override the judge LLM (larger-context for big runs)
   --max-ideas-per-judge-call N    Max ideas per judge LLM call (default 100)
   --resume RUN_ID                 Resume a previously-crashed run (uses --list-runs ids)
@@ -274,6 +285,7 @@ async function runBrainstormCli(
       maxCostUsd: parsed.maxCost,
       maxFarSet: parsed.maxFarSet,
       strictBudget: parsed.strictBudget,
+      modelOverride: parsed.model,
       judgeModel: parsed.judgeModel,
       maxIdeasPerJudgeCall: parsed.maxIdeasPerJudgeCall,
       resumeRunId: parsed.resume,
--- a/src/commands/config.ts
+++ b/src/commands/config.ts
@@ -35,6 +35,19 @@ export function redactConfigValue(key: string, value: string): string {
   return value;
 }
 
+/**
+ * #2119: provider API keys the runtime only ever reads from the env/file
+ * plane (see src/core/ai/build-gateway-config.ts). `config set` for these
+ * writes ~/.gbrain/config.json instead of the DB plane, which nothing reads
+ * for keys. Exported for tests.
+ */
+export const API_KEY_FILE_PLANE_KEYS = [
+  'openai_api_key',
+  'anthropic_api_key',
+  'zeroentropy_api_key',
+  'openrouter_api_key',
+];
+
 export async function runConfig(engine: BrainEngine, args: string[]) {
   const action = args[0];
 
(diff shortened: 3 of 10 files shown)


time-attack:
Merging the API-key plumbing. Verified on master that `config set *_api_key` still writes only the DB plane the gateway never reads for keys, and that `integrations show/status` still probes `process.env` alone — both make a documented command silently lie, which is squarely in the fix lane. One note before rebase: the #1307 brainstorm item is already on master via `resolveBrainstormChatModel`, so please drop that commit (it is the orchestrator.ts conflict) and keep the `--model` flag only if you want it as a separate change.

PR #3155, opened 2026-07-21
fix(doctor): brain_score orphan/timeline components use the orphans-audit linkable scope

time-attack:
Takeover of #2525 (author: @pabloglzg) — rebased onto current master; not closing the original, just referencing it.

## The contradiction

On an archive-heavy brain one `gbrain doctor` report could read "orphan ratio 19% (linkable pages)" in one check while brain_score's `orphans` component implied ~70% — same word, two denominators. The orphans audit (`src/commands/orphans.ts`) excludes archive/generated pages by design; `getHealth()` counted every page. Timeline coverage had the same dilution.

## What

- **`src/core/linkable-scope.ts`** (new): the exclusion constants + predicate move here from `commands/orphans.ts` (which re-exports `shouldExclude` unchanged). The JS predicate and the SQL fragments derive from the same constants so the two paths cannot drift.
- **`getHealth()` in BOTH engines** computes `orphan_pages` and the timeline component over a `linkable_pages` CTE; `BrainHealth.linkable_page_count` carries the denominator. `linkable_page_count === 0` gets full marks for those two components (same vacuous-truth rule as the empty-brain fix).
- **Rebase deltas vs #2525:** master's first-segment exclusions added since the original branch (`raw`, `atoms`, `skills`) are folded into the shared scope, so the move loses none of the audit's current behavior; the advisor eval fixture health mocks gained the new required field.
- New scope defaults from the original PR retained: leading `raw/`, `outputs/`, `daily`, `extracts`, `inbox`, root `readme`/`index`/`schema`/`log`, `/readme` suffix.

## Tests

- `test/brain-score-breakdown.test.ts`: +3 (archive pages excluded from the orphan component; timeline coverage over linkable only; an islanded curated page still counts).
- `test/orphans-pure-fn.test.ts`: +7 for the new scope defaults (incl. the `rawhide/` non-match guard).
- `bun run typecheck` clean; all suites touching `getHealth`/`shouldExclude`/`findOrphans` pass (218 tests across 13 files).

Co-authored-by: pabloglzg


time-attack:
Review + conflict resolution pushed (696491f5). Master landed #3023 (`src/core/orphan-policy.ts`) after this branch was cut, which already centralized the exclusion policy AND filtered the orphan **numerator** in both engines' `getHealth` — but left the **denominators** (no-orphans, timeline coverage) whole-brain, so the dilution this PR targets was still real.

Resolution keeps master's `orphan-policy.ts` as the single source of truth and drops this branch's `src/core/linkable-scope.ts`:

- `linkable-scope.ts` deleted — it was a second competing "single definition", and its SQL projection (a) bypassed #3023's per-brain config overrides (`orphans.exclude_prefixes`/`exclude_slugs`), reintroducing numerator/denominator drift, and (b) had a LIKE wildcard bug: `%/_index` treats `_` as a single-char wildcard, so the SQL scope excluded slugs like `foo/xindex` that the JS predicate kept.
- Both engines now fetch per-page `islanded`/`has_timeline` flags and compute `linkable_page_count`, `orphan_pages`, and the timeline numerator in TS through the same `shouldExcludeFromOrphanReporting(slug, overrides)` call — one predicate, overrides included, engine parity by construction.
- This PR's new scope defaults folded into `orphan-policy.ts`: `outputs/` deny prefix, `inbox` first segment, root `readme`/`index`/`schema`/`log`, `/readme` suffix. (`raw/` leading segment, `daily`, `extracts/` were already covered by #3023.)
- `linkablePageCount === 0` → full marks for both components (vacuous-truth rule), as in the original branch.
- Tests kept: +3 in brain-score-breakdown (the timeline-dilution one fails on master without this fix), +7 in orphans-pure-fn; advisor/recommendation fixtures carry the new required `linkable_page_count`.

Gates: `bun run typecheck` clean; 304 tests across the orphan/doctor/engine/advisor suites pass.

PR #3156, opened 2026-07-21
fix(backlog): part4-5 — bundled-pack registry, per-tool timeouts, non-Anthropic budget pricing, facts model gating

time-attack:
Part 4 fix wave — singleton fix-needed PRs (P1). Four community PRs salvaged, repaired, and rebased onto master. References only — closes nothing.

## Takeover of #2017 — fix(schema): list all bundled packs
The schema.ts hunks were already on master; the still-valid delta was `BUNDLED_PACK_NAMES` (3 packs) drifting from load-active.ts's locator (7 packs), so `gbrain schema show/use gbrain-creator|investor|engineer|everything` missed. Went one step further than the original PR: the registry now lives ONCE in `mutate.ts` and the locator consumes it, so the two can't drift again. New test asserts every registry entry ships a real YAML in `base/`.
Co-authored-by: @arisgysel-design

## Takeover of #2086 — fix: bound subagent tool stalls
Kept the per-tool timeout + idempotent retry (`executeToolWithTimeoutAndRetry`) wired into both `toolLoop` and the legacy Anthropic subagent path, so a wedged pooler/socket settles as a failed tool result instead of squatting a worker slot until the job-level timeout. Repairs: `GBRAIN_SUBAGENT_TOOL_MAX_ATTEMPTS` is parsed as a count (not through `resolveAiTimeoutMs`), the `as any` heartbeat casts are replaced by a typed `tool_retry` event, and the `queue.ts` retryJob hunk is dropped (master already resets started_at/attempts).
Co-authored-by: @JiraiyaETH

## Takeover of #2127 — fix(budget): price non-Anthropic chat models
Kept the `canonicalLookup` fall-through in `lookupPricing` (pre-fix, every non-Anthropic chat model hit `BudgetExhausted(no_pricing)` before the provider call under a cost cap). Dropped the $0 `CANONICAL_PRICING` entries for `openrouter:` ids — those intentionally MISS by the canonical table's contract, and a $0 shared entry would zero cost accounting for every user routing those models. A regression test pins the openrouter fail-closed behavior.
Co-authored-by: @troyhoffman-oss

## Takeover of #2233 — fix(facts): extraction gating and provider config
Salvaged the two real fixes: (1) `extractFactsFromTurn` now probes `isAvailable('chat', <effective facts model>)` instead of the global chat model, ending the silent no-op when facts route through a different provider; (2) `loadConfigWithEngine` merges DB-plane `provider_base_urls.*` keys (previously written by `gbrain config set` but never read back). Dropped: the openrouter_api_key plumbing (already on master) and the hardcoded personal-name normalization + regex turn pre-filter + kind normalizer (privacy-rule violation; fork-specific heuristics).
Co-authored-by: @sene1337

## Verification
- `bun run typecheck` exit 0
- 229 tests green across the touched suites (schema-pack, schema-cli, budget-tracker, subagent tool timeout, loadConfig-merge, facts-extract, model/anthropic pricing drift guards, subagent handler/audit + gateway-path e2e)

### DIFF
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -80,6 +80,20 @@ const AI_CHAT_TIMEOUT_MS = resolveAiTimeoutMs('GBRAIN_AI_CHAT_TIMEOUT_MS', 300_0
 const AI_EMBED_TIMEOUT_MS = resolveAiTimeoutMs('GBRAIN_AI_EMBED_TIMEOUT_MS', 60_000);
 /** multimodal per request. */
 const AI_MULTIMODAL_TIMEOUT_MS = resolveAiTimeoutMs('GBRAIN_AI_MULTIMODAL_TIMEOUT_MS', 60_000);
+/** Per tool execution inside the tool loop. A hung DB/pooler/client call must
+ * settle as a failed tool result instead of squatting a worker slot until the
+ * job's wall-clock timeout expires. */
+const AI_TOOL_TIMEOUT_MS = resolveAiTimeoutMs('GBRAIN_SUBAGENT_TOOL_TIMEOUT_MS', 60_000);
+/** Max attempts for idempotent transient/timeout tool failures. A COUNT, not a
+ * timeout — parsed by its own resolver, not resolveAiTimeoutMs. */
+const AI_TOOL_MAX_ATTEMPTS = resolveAiCount('GBRAIN_SUBAGENT_TOOL_MAX_ATTEMPTS', 2);
+
+function resolveAiCount(envVar: string, fallback: number): number {
+  const raw = process.env[envVar];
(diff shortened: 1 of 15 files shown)


time-attack:
Closing. Two of the four items already landed on master — the bundled-pack registry now lives in `src/core/schema-pack/bundled.ts` as the single source consumed by both `mutate.ts` and `load-active.ts` (this PR would move it back into mutate.ts, a regression), and the `provider_base_urls.*` DB-plane merge is already in config.ts. The budget `canonicalLookup` fall-through and the facts model-gating probe are still real one-liner-class fixes; a small standalone PR for just those two against current master would be an easy review. Thanks for the salvage work and the credits to the original authors.

PR #3157, opened 2026-07-21
fix(lint): only remove paired whole-document Markdown fences

time-attack:
Takeover of #2741 (credit: @codxt, co-authored on the commit).

**Root cause (still on master):** `fixContent` stripped the opening ```` ```markdown ```` fence and a closing fence at EOF as two independent replacements, so any legit document that merely ends in a fenced code example lost its closing fence under `lint --fix`.

**Flaw in #2741:** the paired regex tail `[^\S\r\n]*$` required the closing fence at absolute end-of-string. A wrapped file with a normal trailing newline (the on-disk norm — lint reads raw `readFileSync` content) no longer matched, so detection kept reporting `code-fence-wrap` as fixable while `--fix` never fixed it.

**This PR:** one shared `DOC_FENCE_WRAPPER` regex (paired fences, `\s*$` tail tolerating trailing newlines) used by both detection and `fixContent`, so report and fix can never disagree. Detection also stops false-flagging documents that only end in a legit fenced example.

**Tests:** keeps #2741's two preservation cases (fail on master), adds the trailing-newline unwrap case (fails on #2741's regex) and a detection-consistency case.

- `bun test test/lint.test.ts`: 22 pass / 0 fail
- `bun run typecheck`: clean

### DIFF
--- a/src/commands/lint.ts
+++ b/src/commands/lint.ts
@@ -68,6 +68,12 @@ const LLM_PREAMBLES = [
   /^Absolutely\.?\s*Here[^.\n]*\.?\s*\n*/gim,
 ];
 
+// A whole-document Markdown wrapper: an opening ```markdown fence at the very
+// start AND a closing fence at the very end (trailing newline allowed). Both
+// fences must be present — a closing fence at EOF may belong to a legitimate
+// final code example. Shared by detection + fix so they can never disagree.
+const DOC_FENCE_WRAPPER = /^```(?:markdown|md)[^\S\r\n]*\r?\n([\s\S]*?)\r?\n```\s*$/;
+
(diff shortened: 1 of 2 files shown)


time-attack:
Closing as a duplicate of #3150, which fixes the same `fixContent` unpaired-fence bug. I went with #3150's version because its `analyzeMarkdownWrap` skips YAML frontmatter and refuses to unwrap when a bare fence appears mid-body, whereas the single paired regex here would unwrap — and thereby damage — a document that merely opens with a ```markdown example and closes with a fenced one. Your trailing-newline case is covered there too. Thanks, and credit to @codxt for #2741.

PR #1988, opened 2026-06-08
feat(ai): support OpenRouter prompt caching

tmchow:
## Summary

- mark OpenRouter OpenAI chat routes and Anthropic Claude routes as prompt-cache capable on a per-model-family basis
- add an OpenRouter fetch shim that rewrites the OpenAI-compatible adapter's private prompt-cache sentinel into top-level `cache_control` for Anthropic Claude routes only
- preserve OpenAI route behavior as automatic/no-extra-request-mutation and keep non-cacheable OpenRouter families degraded
- surface OpenAI-compatible cached-token usage from AI SDK structured usage when available
- fix an existing test-isolation lint violation in `test/db-lock-heartbeat-takeover.test.ts` by using the repo's `withEnv()` helper

Fixes #1987.

## Why

OpenRouter supports prompt caching for compatible routed providers, but not uniformly across the full catalog:

- OpenAI routes: caching is automatic when the model/provider supports it
- Anthropic Claude routes: OpenRouter requires explicit `cache_control`
- Other routed families should not be blanket-marked cache-capable

This keeps the recipe capability honest while removing the `degraded:no_caching` warning for the OpenRouter model families that do support caching.

## Tests

- `bun test test/ai/recipe-openrouter.test.ts test/ai/capabilities.test.ts`
- `scripts/check-test-isolation.sh`
- `bun test test/db-lock-heartbeat-takeover.test.ts`
- `bun run typecheck`
- `bun run test` exceeded the local 10-minute command timeout while the full sharded suite was still running; no failure from this change was observed before timeout.

PR #1618, opened 2026-05-29
feat(zhipu): add chat/expansion support + fix expand() for openai-compat providers

punksterlabs:
## Summary

- **Add chat + expansion touchpoints to the Zhipu recipe** — the recipe only declared embedding touchpoints, so `isAvailable('chat')` and `isAvailable('expansion')` returned false even with `ZHIPUAI_API_KEY` set. Adds GLM-4.7, GLM-4.7-flash, GLM-5.1, GLM-5-turbo as chat/expansion models.

- **Fix silent expansion failure for all openai-compatible providers** — `generateObject()` uses `response_format: json_schema` which most openai-compatible endpoints don't support. The call fails silently and `expand()` returns only the original query. For openai-compat recipes, fall back to `generateText()` + manual JSON parse. Native providers (Anthropic, OpenAI, Google) keep the existing `generateObject()` path unchanged.

This affects every openai-compatible provider with expansion enabled: Zhipu/GLM, DeepSeek, Groq, Together, Ollama, and any future recipe using the `openai-compatible` implementation.

## Test plan

- [x] Confirmed `gbrain providers list` shows zhipu with EMBED/EXPAND/CHAT all `yes` + `ready`
- [x] Confirmed `expand()` returns 3-4 expanded queries via GLM-4.7 through Z.ai endpoint
- [x] Confirmed `isAvailable('expansion')` and `isAvailable('chat')` return true with `ZHIPUAI_API_KEY` set
- [x] Confirmed native providers (Anthropic/OpenAI) still use the `generateObject()` path (no behavioral change)
- [ ] Verify other openai-compat providers (DeepSeek, Groq) also benefit from the generateText fallback

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3158, opened 2026-07-21
fix(ai/gateway): expansion structured-output fallback, zhipu chat recipe, OpenRouter Anthropic prompt caching

### LINKED ISSUE #1987: Support prompt caching for OpenRouter Anthropic routes
## Summary

OpenRouter now documents prompt caching support for Anthropic/Claude routes, including explicit `cache_control` breakpoints and sticky routing/session behavior. GBrain currently treats the OpenRouter recipe as not prompt-cache-capable:

```ts
// src/core/ai/recipes/openrouter.ts
supports_prompt_cache: false
```

This means `openrouter:anthropic/...` models are classified as uncached/degraded in GBrain even though the upstream OpenRouter route may support caching.

## Why this matters

For high-volume GBrain loops (`dream`, extraction, subagents), prompt caching is a major cost control. PR #1761 improves Anthropic-direct call sites, but users who route Claude through OpenRouter still get doctor warnings / degraded capability classification.

Current behavior observed locally:

```text
models.tier.subagent = openrouter:...
provider does not support prompt caching
The subagent loop runs hot
```

## Relevant docs

OpenRouter prompt caching docs:
https://openrouter.ai/docs/guides/best-practices/prompt-caching

Key details:

- OpenRouter supports prompt caching on supported providers/models.
- Anthropic-style explicit caching uses `cache_control`.
- OpenRouter supports sticky routing via `session_id` / `x-session-id`, useful for multi-turn agent workflows.
- Cache usage can be inspected through `prompt_tokens_details.cached_tokens` / `cache_write_tokens`.

Anthropic prompt caching docs:
https://platform.claude.com/docs/en/build-with-claude/prompt-caching

## Related GBrain issues/PRs

- #1761 — Anthropic-direct prompt caching call-site opt-ins
- #1157 — configurable Anthropic-compatible provider list
- #621 — provider/helper-friendly dream/subagent calls
- #1487 / #1739 / #1886 — gateway/non-Anthropic tool-loop issues

## Design question

Should GBrain support prompt caching for OpenRouter Anthropic routes narrowly, e.g.:

```text
openrouter:anthropic/claude-*
```

or should this be part of a broader provider capability override system?

Possible narrow implementation:

1. Detect OpenRouter Anthropic model IDs:
   - `openrouter:anthropic/claude-*`
2. Mark those as prompt-cache-capable while leaving other OpenRouter models uncached.
3. Pass Anthropic-style `cache_control` through OpenRouter’s OpenAI-compatible request shape when `cacheSystem` is set.
4. Optionally set OpenRouter `session_id` / `x-session-id` for sticky routing in multi-turn loops.
5. Add tests that verify request shape without requiring a live cache hit.

## Expected behavior

When configured with:

```bash
gbrain config set models.tier.subagent openrouter:anthropic/claude-sonnet-4.6
```

GBrain should not classify the model as `degraded:no_caching` if the route supports prompt caching.

## Non-goals

- Do not mark all OpenRouter models as prompt-cache-capable.
- Do not rely on live billing/cached-token assertions in normal CI.
- Do not conflate OpenRouter’s OpenAI-compatible transport with Anthropic-compatible direct APIs.

### LINKED ISSUE #2372: query expansion on OpenAI-compatible providers: silent degradation + no path to strict structured outputs
**What happened?**

Query expansion through an OpenAI-compatible provider has two related problems, and the existing reports each cover one half:

1. **Silent degradation.** `expand()` calls `generateObject()` with a Zod schema, which makes the AI SDK request a `json_schema` response_format. Most OpenAI-compatible endpoints don't support that, so `@ai-sdk/openai-compatible` drops the schema, logs an `AI SDK Warning ("responseFormat" is not supported)`, and falls back to a schemaless `json_object` request. When the model then names the JSON key something other than `queries` (it often picks `rewrites`, the most prompt-salient noun), schema validation fails and `expand()` silently returns `[query]` with no expansion. This is the failure described in #1156.

2. **No path to strict structured outputs.** Some OpenAI-compatible backends *do* honor a strict `json_schema` response_format. For those, the current code can't ask for it (the provider model is built without `supportsStructuredOutputs`), so they are stuck on the lossy schemaless path even though they could give strict, validated output.

Between them, #1158 addresses (1) at the prompt level (naming the key), and #1618 addresses (1) at the call level (skip `generateObject` for OpenAI-compatible providers, use `generateText` + manual parse). Neither addresses (2), and there is no single place that decides per-recipe which path a backend should take.

**What did you expect?**

- Expansion should never silently no-op on an OpenAI-compatible backend.
- A backend that supports strict `json_schema` should be able to opt in and get strict, schema-validated output instead of the lossy path.
- Native providers (Anthropic, OpenAI, Google) should keep their existing `generateObject` behavior.

**Steps to reproduce**

1. Configure `models.expansion` to an OpenAI-compatible provider (`litellm:*`, `deepseek:*`, `groq:*`, a local proxy, etc.).
2. Run `gbrain query "<anything>"` with expansion enabled (the default).
3. Observe the `responseFormat is not supported` warning on stderr, and (depending on the backend's key choice) expansion silently collapsing to the single original query.

**Environment**

- gbrain `master`; reproduces on Linux and macOS, Postgres and PGLite.
- Any OpenAI-compatible expansion provider.

**Proposed fix**

Decide the expansion path per recipe and unify the existing fixes:

- Add `ChatTouchpoint.supports_structured_outputs` (default `false`) and thread it into `createOpenAICompatible`'s `supportsStructuredOutputs`.
- Route `expand()` three ways: native providers use `generateObject` unchanged; OpenAI-compatible recipes that opt into structured outputs request a strict `json_schema` and fall back to the text path if it is rejected at call time; every other OpenAI-compatible recipe skips the `json_schema` attempt and parses the model's text directly (no warning, no silent no-op).
- Keep the explicit `queries`-key constraint in the prompt (from #1158) and a tolerant JSON decode (fences / prose) so the schemaless path is reliable.

Related: #1156, #1158, #1618. A PR implementing this follows, and cherry-picks the worthwhile pieces of #1158 and #1618 so their authorship is preserved.

time-attack:
Backlog work unit c38-f1 (gateway component): three commits, one per item.

## Items

- **Fixes #2372** — query expansion on OpenAI-compatible providers no longer silently degrades. `expand()` routes openai-compat backends through `generateText` + tolerant parse (prompt now pins the `queries` key); a new opt-in `ChatTouchpoint.supports_structured_outputs` threads `supportsStructuredOutputs` into `createOpenAICompatible` and picks strict `generateObject` (with call-time text fallback) for backends that declare it. `parseLlmJson` extracted to leaf `src/core/llm-json.ts` with a re-export preserving existing importers.
- **Takeover of #2373** (author: @brettdavies, credited via Co-authored-by) — that PR is exactly this fix but CONFLICTING against master after #2857/#2981 drift; rebased mechanically here, no semantic changes.
- **Takeover of #1618** (author: @punksterlabs, credited via Co-authored-by) — reduced to its still-missing half: the zhipu recipe gains chat + expansion touchpoints (GLM via openai-compat). Its `expand()` fix half is superseded by the #2372 commit above.
- **Fixes #1987** — prompt caching for OpenRouter Anthropic routes (rework of #1988, author: @tmchow, credited via Co-authored-by). `supports_prompt_cache` may now be function-valued per model id; OpenRouter scopes it to `anthropic/claude-*`, so those routes stop classifying `degraded:no_caching`. The post-#2981 system-block breakpoint additionally rides `providerOptions.openaiCompatible` on the compat wire, and a new chat-scoped recipe fetch shim (`compat.chatFetch`) lifts the marker into OpenRouter's documented content-part `cache_control` shape before the request leaves the process. `chatFetch` deliberately does not displace the embedding path's asymmetric `input_type` shim the way `compat.fetch` would.

References (not auto-closed): #2373, #1618, #1988.

## Tests

- `test/ai/gateway-chat.test.ts` — structured-output gating + schemaless recovery seam (fails without the expand() rework).
- `test/ai/recipe-zhipu.test.ts` — chat/expansion touchpoint shape (fails without the recipe addition).
- `test/ai/gateway-cache-breakpoint.test.ts` — openrouter claude route plants both cache markers on the system block; non-claude route stays uncached (fails without #1987).
- `test/ai/recipe-openrouter.test.ts` — capability scoping, `liftMessageCacheControl`, fetch shim wire rewrite, chatFetch installation.
- `test/ai/capabilities.test.ts` — function-valued capability classification.

`bun run typecheck` exit 0; full `test/ai/` suite (412 tests, 39 files) + `test/conversation-parser/` green.

### DIFF
--- a/src/core/ai/capabilities.ts
+++ b/src/core/ai/capabilities.ts
@@ -90,7 +90,10 @@ export function getProviderCapabilities(modelString: string): ProviderCapabiliti
 
   return {
     supportsToolCalling: chat.supports_tools === true,
-    supportsPromptCaching: chat.supports_prompt_cache === true,
+    // #1987: may be per-model-id (OpenRouter caches anthropic/claude-* routes).
+    supportsPromptCaching: typeof chat.supports_prompt_cache === 'function'
+      ? chat.supports_prompt_cache(parsed.modelId)
+      : chat.supports_prompt_cache === true,
     // No recipe exposes parallel-tools-specifically yet; gate on supports_tools.
     // Subsequent waves can split this into its own recipe field if a provider
     // ever supports tools without parallel dispatch.
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -48,6 +48,7 @@ import type {
 } from './types.ts';
 import { resolveRecipe, assertTouchpoint, parseModelId } from './model-resolver.ts';
 import { resolveModel, TIER_DEFAULTS } from '../model-config.ts';
+import { parseLlmJson } from '../llm-json.ts';
 import type { BrainEngine } from '../engine.ts';
 import { dimsProviderOptions } from './dims.ts';
 import { hasAnthropicKey } from './anthropic-key.ts';
@@ -405,6 +406,20 @@ export function applyOpenAICompatConfig(
     );
   }
   return { baseURL, fetch: recipe.compat?.fetch };
+}
+
+/**
+ * Whether an openai-compatible recipe's backend honors OpenAI structured
+ * outputs. Threaded into `createOpenAICompatible`'s `supportsStructuredOutputs`
+ * at the chat + expansion build sites, and consulted by `expand()` to pick the
+ * strict `generateObject` path over the schemaless text path. Single source of
+ * truth read from the chat touchpoint: the backend serves both chat and
+ * expansion, so the capability is declared once.
+ *
+ * @internal exported for tests.
+ */
+export function recipeSupportsStructuredOutputs(recipe: Recipe): boolean {
+  return recipe.touchpoints.chat?.supports_structured_outputs === true;
 }
 
 /**
@@ -2264,6 +2279,7 @@ function instantiateExpansion(recipe: Recipe, modelId: string, cfg: AIGatewayCon
         baseURL: compat.baseURL,
         ...(compat.fetch ? { fetch: compat.fetch } : {}),
         ...auth,
+        supportsStructuredOutputs: recipeSupportsStructuredOutputs(recipe),
       }).languageModel(modelId);
     }
   }
@@ -2272,6 +2288,20 @@ function instantiateExpansion(recipe: Recipe, modelId: string, cfg: AIGatewayCon
 const ExpansionSchema = z.object({
   queries: z.array(z.string()).min(1).max(5),
 });
+
+/**
+ * Recover expansion queries from a schemaless model response. Used by the
+ * openai-compatible expansion paths: a tolerant JSON decode plus schema
+ * validation pulls the `queries` array out of the model's text (the prompt
+ * pins it to a bare JSON object). Returns null when the text carries no valid
+ * `{ queries: string[] }` object.
+ *
+ * @internal exported for tests.
+ */
+export function parseExpansionResponse(text: string): string[] | null {
+  const parsed = ExpansionSchema.safeParse(parseLlmJson<unknown>(text));
+  return parsed.success ? parsed.data.queries : null;
+}
 
 /**
  * Expand a search query into up to 4 related queries.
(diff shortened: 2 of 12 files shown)


time-attack:
Closing — everything in this PR is already on master. The expansion structured-output gating plus the extracted `src/core/llm-json.ts` and the OpenRouter Anthropic prompt-cache scoping landed in #3152, and the zhipu chat touchpoint landed in #3084. I checked each seam directly (`recipeSupportsStructuredOutputs`, `openrouterSupportsPromptCache`, `zhipu.touchpoints.chat`) and found no remaining gap. Thanks for the work, and credit to @brettdavies, @punksterlabs and @tmchow for the originals.

PR #3159, opened 2026-07-21
fix(extract): derive timeline entries from real-provenance effective dates

time-attack:
Takeover of #2877 (author: @ShintaroKawakami — credited via Co-authored-by on the commit).

## What
DB timeline extraction (`gbrain extract timeline --source db` and `gbrain extract --stale`) emits one deterministic fallback timeline row (`date=effective_date`, `summary=page.title`, `source=page.effective_date:<provenance>`) for pages that have no explicit markdown timeline row.

## What was wrong with #2877 as submitted
1. **Conflicting with master** — the `getPage` / `listStalePagesForExtraction` hunks in `postgres-engine.ts` predated the RLS scoped-read-transaction refactor. Rebased; the column additions now live inside the `withScopedReadTransaction` projections.
2. **Unfiltered on provenance** — `computeEffectiveDate` falls back to `updated_at`/`created_at` with source `'fallback'`, so nearly every imported page carries a non-null `effective_date`. The PR as written would have created one mtime-dated noise timeline row per timeline-less page brain-wide on the next extract run. Fixed: `extractTimelineEntriesForPage` only emits the fallback row for real date provenance (`event_date`/`date`/`published`/`filename`); `'fallback'` and null provenance are skipped, with regression tests at both the unit and DB (PGLite) level.

## Testing
- `bun test test/extract.test.ts test/extract-db.test.ts test/extract-stale.test.ts` — 61 pass, 0 fail
- `bun run typecheck` — clean
- `scripts/check-key-files-current-state.sh` — ok; `bun run build:llms` — bundles unchanged

References #2877 (not auto-closing; maintainer can close it in favor of this one).


time-attack:
Closing. This is a new capability rather than a fix — nothing on master is broken; it synthesizes a timeline row for every page that has real date provenance but no markdown timeline entry, which on the first `extract --stale` after upgrade writes machine-derived rows across an entire brain and mixes them into curated human timelines. The provenance filter you added over #2877 is good defensive work, but it is tuning a new surface, and new timeline-generation surface is not something we are taking right now. Thanks for the careful rebase and for crediting @ShintaroKawakami.

PR #3160, opened 2026-07-21
fix(page-lock): never fast-steal a fresh lock on a dead-looking PID

### LINKED ISSUE #2840: page-lock: `isPidAlive()` silently steals a live lock across PID namespaces (containerized deploys)
**Version:** verified on `master` @ `5008b287` (**0.42.59.0**). `src/core/page-lock.ts` has received no commits in the range I checked, so this is not a regression.
**Deployment:** Kubernetes — `oven/bun` image, Postgres (CNPG) engine, `GBRAIN_HOME` on a shared volume, with `gbrain serve --http` and `gbrain jobs work` running as **separate containers**.

## Summary

`withPageLock` decides staleness partly by PID liveness. A PID is only meaningful inside the namespace that produced it, so as soon as two writers live in different PID namespaces — two containers, or two pods sharing `GBRAIN_HOME` — the check is wrong **in both directions**.

`src/core/page-lock.ts:56`:

```ts
function isPidAlive(pid: number): boolean {
  if (pid <= 0) return false;
  if (pid === process.pid) return true;
  try { process.kill(pid, 0); return true; }
  catch (e) { return (e as NodeJS.ErrnoException).code !== 'ESRCH'; }
}
```

The lock file holds `{pid}\n{iso-timestamp}`; staleness is `mtime > LOCK_TTL_MS` **OR** dead PID.

- **Live lock stolen.** The holder's PID doesn't exist in the reader's namespace → `ESRCH` → judged stale → the reader takes a lock someone else holds. Two concurrent read-modify-writes on the same `<slug>.md`; last writer wins, no error surfaced.
- **Foreign PID mistaken for self.** PIDs collide across namespaces. If the reader's own `process.pid` equals the recorded one, `pid === process.pid` returns `true` and an unrelated process in another namespace is treated as "a concurrent caller in this process".

The 5-minute mtime TTL bounds how long a stale lock lingers, but does not remove the race.

## Why this isn't exotic

It bites any containerized deploy where more than one process writes facts or takes — e.g. `gbrain serve` (which exposes `extract_facts` / `forget_fact` over MCP) next to `gbrain jobs work`. That's a documented way to run a worker beside a server.

## Relation to existing issues

This is the **file-lock** sibling of a pattern already recognised elsewhere in the codebase:

- **#2308** — `gbrain_cycle_locks` lease takeover is same-host only (`holder_host` / `holder_pid`). Same root cause — process identity isn't portable across containers — but that's the **db-lock**; this one is the FS lock in `page-lock.ts` and is not covered there.
- **#2348** — the `.gbrain-lock` PID-1 trap is the exact mirror: there a PID-1 one-off sees *itself* and refuses to steal a dead lock (fails loudly); here a container sees a foreign PID and steals a live one (corrupts quietly).
- **#1972 / #2015** (closed) — stale dead-PID *cycle* locks never reaped: opposite direction (not reaped, vs. wrongly stolen).
- **#2443** — `serve` as PID 1 in containers; adjacent process-identity assumption.

## Suggested directions

1. **Namespace-stable holder identity** instead of a bare PID — boot-id + PID, or a UUID minted per process and heartbeated via mtime. (Note #2308 shows `holder_host` alone is not enough, for the same reason.)
2. **Drop PID-liveness here and rely on mtime + refresh**: the holder refreshes while working, TTL expiry becomes the only staleness signal. Simpler and namespace-agnostic; costs a bounded wait after a crash.
3. **At minimum**, document that `GBRAIN_HOME` must not be shared across PID namespaces, and refuse to steal when the recorded holder can't be verified.

I'd lean toward (2) as the smallest sound change.

> See also: #2839 — the sync/import path never takes this lock at all, so fixing this alone still leaves the importer outside it.

time-attack:
Fixes #2840

## Problem

`src/core/page-lock.ts` classified a lock as stale when the holder's PID looked dead (`process.kill(pid, 0)` → ESRCH), stealing it INSTANTLY even with a fresh mtime. Across PID namespaces (two containers sharing GBRAIN_HOME — a documented serve + jobs-work topology) a live holder's PID reads as ESRCH in the reader's namespace, so a live lock was silently stolen and the atomic read-modify-write the lock protects (fence writes, takes, forget) could race and corrupt page markdown.

## Fix

- Staleness is now **mtime-TTL only** (5 min). Holders already have `refresh()`; a crashed holder ages out via the TTL. This matches the #2348 pglite-lock philosophy: never steal a possibly-live holder. The PID-liveness fast-steal is dropped entirely (it cannot be made namespace-safe without a boot-id handshake, and the TTL backstop already bounds recovery).
- Ownership for `refresh()`/`release()` moves from PID comparison (same cross-namespace false-self problem: container A pid 7 vs container B pid 7) to a per-acquire random token written as the lock file's third line, so a stale handle can never clobber or unlink a new owner's live lock.

## Tests

`test/page-lock.test.ts`:
- New regression test: a fresh lock whose PID reads as dead (the cross-namespace shape) is NOT stolen; the same lock past the TTL IS reclaimable. Fails on master.
- Foreign-release test updated to token-based ownership.

Gates: `bun run typecheck` exit 0; `bun test test/page-lock.test.ts test/fence-write.test.ts test/takes-fence.test.ts test/privacy-strip-and-forget.test.ts test/takes-command-source-scope.test.ts` all pass.

### DIFF
--- a/src/core/facts/fence-write.ts
+++ b/src/core/facts/fence-write.ts
@@ -11,8 +11,8 @@
  *
  * Concurrency: reuses the v0.28 page-lock primitive
  * (`src/core/page-lock.ts`), an FS-level lockfile under
- * `~/.gbrain/page-locks/<sha256-of-slug>.lock` with PID-liveness +
- * 5-minute TTL. Multi-process safe — two `gbrain` invocations writing
+ * `~/.gbrain/page-locks/<sha256-of-slug>.lock` with a 5-minute
+ * mtime TTL. Multi-process safe — two `gbrain` invocations writing
  * to the same entity page serialize through the same kernel-visible
  * lockfile. 5-second timeout per the plan's "5s retry" failure mode.
  *
--- a/src/core/page-lock.ts
+++ b/src/core/page-lock.ts
@@ -9,8 +9,13 @@
  * Lock file path: `~/.gbrain/page-locks/<sha256-of-slug>.lock`. SHA-256
  * keeps filenames safe regardless of slug content (slashes, unicode, etc.).
  *
- * File contents: `{pid}\n{iso-timestamp}`. Staleness = mtime older than
- * `LOCK_TTL_MS` (5 min) OR the PID is no longer alive on this host.
+ * File contents: `{pid}\n{iso-timestamp}\n{owner-token}`. Staleness = mtime
+ * older than `LOCK_TTL_MS` (5 min) — nothing else. PID-liveness
+ * (`process.kill(pid, 0)`) is deliberately NOT consulted (#2840): across PID
+ * namespaces (two containers sharing GBRAIN_HOME) a live holder's PID reads
+ * as ESRCH, so a liveness fast-steal silently steals a live lock. Same
+ * philosophy as the #2348 pglite-lock: never steal a possibly-live holder.
+ * Holders `refresh()` to stay fresh; a crashed holder ages out via the TTL.
  *
  * Usage:
  *
@@ -24,7 +29,7 @@
 
 import { existsSync, mkdirSync, readFileSync, statSync, unlinkSync, writeFileSync } from 'node:fs';
 import { join } from 'node:path';
-import { createHash } from 'node:crypto';
+import { createHash, randomUUID } from 'node:crypto';
 import { gbrainPath } from './config.ts';
 
 const LOCK_TTL_MS = 5 * 60 * 1000; // 5 minutes — matches eng-review fold spec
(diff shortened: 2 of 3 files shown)


time-attack:
Merging. I confirmed on current master that page-lock still treats an ESRCH PID as authorization to steal a fresh lock, and that refresh() rewrites the lockfile without checking ownership — both let two writers hold the same page. Making staleness mtime-TTL-only with a UUID owner token matches the existing pglite-lock (#2348) rule that liveness can protect a holder but never authorize a steal, and the diff stays small with a test pinning the new invariant. Thanks for keeping it to the one primitive.

PR #3163, opened 2026-07-21
fix(put): reject empty content and stacked frontmatter instead of writing corrupt pages

### LINKED ISSUE #2743: put_page silently accepts multi-frontmatter content (YAML double-put corruption class)
## Class

`gbrain put` (both the CLI form and `mcp__gbrain__put_page`) accepts input containing more than one leading `---...---` YAML frontmatter block and stores it as-is. At read time the second block silently shadows the first, corrupting page state (missing `type`, wrong `title`, dropped `links[]`).

## Concrete failure surfaced 2026-07-10

An enrich script emitted `links:\n` for the empty-list case. The next round's regex saw `links: null` and inserted a second `---...---` frontmatter block on top of the file. The put succeeded. The page now had two frontmatter blocks stacked. Downstream reads silently used the second block. Fanout audit was blind to it until forensic YAML parsing.

Root cause is that `gbrain put` does no YAML shape validation on the leading region.

## Proposed fix

At `gbrain put` ingest time:

1. Count leading `---...---` blocks in the input (walk from position 0 skipping blank lines; stop at first non-frontmatter character).
2. If `> 1`:
   - **Strict mode (default):** reject the put with a structured error identifying \"YAML double-put corruption\" and pointing to this class doc.
   - **Auto-canonicalize mode (opt-in `--merge-frontmatter`):** merge the blocks with last-wins semantics for scalar keys and array-append for `links[]`, `tags[]`, then proceed. Log the merge to the ingest log so it is auditable.

The count is well-defined and cheap: single-pass, no YAML parser needed for detection. Merging is more work but the code path already knows how to serialize `gbrain page frontmatter` so the reverse (parse + merge two YAML blocks) is a small addition.

## Scope

- Applies to both the CLI `gbrain put` path and the MCP `put_page` tool.
- Does NOT apply to `---` horizontal rules inside the body — the detector must be leading-only (stop at first non-frontmatter character).
- Body-side edits should be untouched.

## Workspace-side inoculation (already landed)

We shipped a workspace-side PreToolUse hook (\`multi-frontmatter-guard\`) that catches this class before it reaches `gbrain put` from the paths we control (MCP put_page + `gbrain put <slug> < file` Bash redirects). See michaeladair44/zeus-openclaw-workspace PR #73. The hook is a stopgap; the canonical fix belongs here in `gbrain put`.

## Test cases upstream should carry

- empty input → allow (0 blocks)
- body-only, no frontmatter → allow (0 blocks)
- single well-formed frontmatter → allow (1 block)
- stacked double-put (the 2026-07-10 class) → reject in strict, merge in opt-in
- pathological triple stacked → same
- single frontmatter + body `---` horizontal rule → allow (rules are not frontmatter)
- BOM-prefixed frontmatter → allow

## Not a PR

Filing as an ISSUE per the workspace-side review contract; PR authorship on `garrytan/gbrain` is gated on Mike. Happy to sketch the upstream code shape in comments if useful.

## Related

- Meeting-ingest backfill incident report: `status/2026-07-10-meeting-backfill-complete` substrate finding #2 (in the reporter's brain repo).
- Workspace-side inoculation: michaeladair44/zeus-openclaw-workspace#73.

### LINKED ISSUE #2822: CLI `put`: empty non-TTY stdin silently writes an empty page (0 chunks, invisible to embed --stale) — stray file-path positional silently discarded
## Version / environment

- gbrain 0.42.59.0, postgres engine, macOS (bun install)
- Reproduced identically on 0.42.57 — not a recent regression.
- We run the `qwen3-mrl-support` fork, but its patches touch only `dims.ts` / the ollama recipe / engine chunk labels — every code path cited below is upstream-unmodified.

## Summary

`gbrain put` is the CLI generated from the `put_page` op via `cliHints: { name: 'put', positional: ['slug'], stdin: 'content' }` (src/core/operations.ts:1111). In a **non-TTY shell** (agent harnesses, cron, CI — increasingly the common way this CLI gets driven), calling it the "obvious wrong" way:

```
gbrain put ./note.md --slug notes/foo
```

exits 0 and writes a page with an **empty body**: 0 `content_chunks`, no embeddings, invisible to semantic search. Two silent behaviors compound:

1. **The file path is silently discarded.** `./note.md` is consumed as the `slug` positional, then `--slug notes/foo` overwrites it (`parseOpArgs`, src/cli.ts:739ff). No warning that a positional was dropped.
2. **Empty stdin becomes the body.** With `!process.stdin.isTTY`, `parseOpArgs` reads fd 0 (src/cli.ts:771–778) — `/dev/null` yields `""`. The required-param check only catches `undefined` (src/cli.ts:344), so `content: ""` passes validation. `parseMarkdown("")` → zero chunks → the empty-chunks branch in `importFromContent`'s transaction even runs `tx.deleteChunks`.

In an interactive TTY this fails loudly (stdin isn't read, `content` stays undefined → required-param error). The trap is **exclusive to non-TTY invocation** — exactly where agents run the CLI, and where nobody is watching stderr.

## Why this is nasty to diagnose

- The page exists and `get` works; it's just missing from semantic search. `gbrain embed --stale` reports 0 (stale = chunks missing embeddings; a zero-chunk page isn't in the pool), so no background sweep ever repairs it.
- A follow-up "fix" attempt (`gbrain put ./note.md --slug notes/foo --type decision`) *also* sends `content=""` → `content_hash` identical → `status: skipped` (src/core/import-file.ts:562). To the operator it looks like **updates are being ignored** on top of **embeddings being broken** — two phantom bugs from one silent empty read.
- We initially attributed this to the v0.41 content-sanity gate (`embed_skip`) and spent a full investigation ruling that out.

## Repro (0.42.59.0)

```
$ gbrain put notes/repro-a </dev/null
→ status "created_or_updated", chunks 0; DB: compiled_truth length 0, 0 content_chunks rows

$ printf -- '---\ntype: note\n---\n\nreal body text\n' | gbrain put notes/repro-b
→ status "created_or_updated", chunks 1, embedded inline (correct)

$ gbrain embed --stale --dry-run
→ "[dry-run] Would embed 0 chunks (0 stale found)"   # repro-a is invisible forever
```

## Suggested fixes (any subset)

1. When a `cliHints.stdin`-sourced **required** param resolves to empty/whitespace-only in a non-TTY shell, error out (or at minimum warn loudly) instead of writing an empty page.
2. When a positional value is about to be overwritten by an explicit flag **and** the discarded value matches an existing file path, warn: `did you mean 'gbrain put <slug> < file.md' or 'gbrain capture --file PATH --slug SLUG'?`
3. Include a reason on zero-chunk results (e.g. `chunks: 0, chunk_skip_reason: "empty_body"`) so this is diagnosable from the response instead of from the DB.

Related: #2743 (another silent-acceptance corruption class in `put_page`).

time-attack:
Two adversarially-verified backlog items on the put_page ingest path.

## Fixes #2822 — CLI `put`: empty non-TTY stdin silently wrote an empty page

An empty pipe (cron with no input, broken shell substitution) produced a 0-chunk page: invisible to search and to `embed --stale`, so the mistake never surfaced.

- `parseOpArgs` now leaves the stdin-fed param **unset** when stdin is empty/whitespace, so the existing required-param check rejects the call. The error names the real problem (`required "content" is missing — stdin was empty or not piped`) instead of only printing the positional usage line.
- A flag that overwrites a positional value (`gbrain put CONTENT --slug foo` — CONTENT was parsed as the slug, then silently discarded) now warns to stderr, and vice versa.
- `importFromContent` throws on empty/whitespace content as the shared backstop — covers MCP `put_page` (which only required presence, so `""` passed), capture, and every other caller. Thrown, not returned, matching the ContentSanityBlockError flow so every wrapper site surfaces the message.

## Fixes #2743 — put_page silently accepted stacked frontmatter (double-put corruption)

gray-matter parses the first `---…---` block and silently leaves a second one in the body, poisoning every subsequent round-trip.

- `parseMarkdown(validate: true)` grows a `MULTI_FRONTMATTER` finding: first non-empty line after the close fence opens another fenced block containing YAML-shaped `key:` lines. Lone horizontal rules, hrule pairs around plain prose, and the `---` + `## Timeline` sentinel are NOT flagged.
- `gbrain lint` maps it to the stable rule name `frontmatter-multi`.
- `importFromContent` rejects `MULTI_FRONTMATTER` before any DB write; the other validation codes keep their lint-only semantics.

## Tests

- `test/cli-args.test.ts`: empty/whitespace stdin `put` exits 1 via real subprocess (isolated HOME); overwrite warnings both directions; no warning when values agree.
- `test/markdown-validation.test.ts`: MULTI_FRONTMATTER positive (adjacent + blank-line-separated) and negative (body hrules, prose between rules, timeline sentinel) cases.
- `test/import-file.test.ts`: empty and stacked-frontmatter content throw before any `putPage` call; body hrules still import.

Gates: `bun run typecheck` exit 0; all three touched test files green (67 pass / 0 fail). Two hook-timeout flakes seen in unrelated PGLite suites reproduce identically with this diff stashed (parallel cold-start contention, pre-existing).

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -365,6 +365,12 @@ async function main() {
     if (def.required && params[key] === undefined) {
       if (queryHasAlt && key === 'query') continue;
       const cliName = op.cliHints?.name || op.name;
+      // #2822: when the missing param is the op's stdin-fed one, the usage
+      // line alone is misleading (the positionals may all be present — the
+      // pipe was just empty). Name the real problem.
+      if (op.cliHints?.stdin === key) {
+        console.error(`Error: required "${key}" is missing — stdin was empty or not piped. Pipe content on stdin or pass --${key.replace(/_/g, '-')}.`);
+      }
       const positional = op.cliHints?.positional || [];
       const usage = positional.map(p => `<${p}>`).join(' ');
       console.error(`Usage: gbrain ${cliName} ${usage}`);
@@ -761,6 +767,10 @@ export function parseOpArgs(op: Operation, args: string[]): Record<string, unkno
   const params: Record<string, unknown> = {};
   const positional = op.cliHints?.positional || [];
   let posIdx = 0;
+  // #2822: track which params came from positionals so a later flag that
+  // silently discards one (`gbrain put CONTENT --slug foo` — CONTENT was
+  // parsed as the slug) gets a stderr warning instead of vanishing.
+  const positionallySet = new Set<string>();
 
   for (let i = 0; i < args.length; i++) {
     const arg = args[i];
@@ -778,13 +788,20 @@ export function parseOpArgs(op: Operation, args: string[]): Record<string, unkno
       if (paramDef?.type === 'boolean') {
         params[key] = true;
       } else if (i + 1 < args.length) {
+        if (positionallySet.has(key) && params[key] !== args[i + 1]) {
+          console.error(`Warning: ${arg} overrides the positional <${key}> value ${JSON.stringify(params[key])}.`);
+        }
         params[key] = args[++i];
         if (paramDef?.type === 'number') params[key] = Number(params[key]);
       }
     } else if (posIdx < positional.length) {
       const key = positional[posIdx++];
       const paramDef = op.params[key];
+      if (params[key] !== undefined && params[key] !== (paramDef?.type === 'number' ? Number(arg) : arg)) {
+        console.error(`Warning: positional <${key}> overrides the earlier --${key.replace(/_/g, '-')} value ${JSON.stringify(params[key])}.`);
+      }
       params[key] = paramDef?.type === 'number' ? Number(arg) : arg;
+      positionallySet.add(key);
     }
   }
 
@@ -796,7 +813,11 @@ export function parseOpArgs(op: Operation, args: string[]): Record<string, unkno
       console.error(`Error: stdin content exceeds ${MAX_STDIN} bytes. Split into smaller inputs.`);
       process.exit(1);
     }
-    params[op.cliHints.stdin] = stdinContent;
+    // #2822: empty/whitespace-only stdin (cron with no input, broken pipe)
+    // stays UNSET so the required-param check rejects the call instead of
+    // silently writing an empty page (0 chunks, invisible to search and
+    // embed --stale).
+    if (stdinContent.trim().length > 0) params[op.cliHints.stdin] = stdinContent;
   }
 
   return params;
--- a/src/commands/lint.ts
+++ b/src/commands/lint.ts
@@ -48,6 +48,7 @@ const FRONTMATTER_RULE_NAMES: Record<ParseValidationCode, string> = {
   NESTED_QUOTES: 'frontmatter-nested-quotes',
   NON_STRING_FIELD: 'frontmatter-non-string-field',
   EMPTY_FRONTMATTER: 'frontmatter-empty',
+  MULTI_FRONTMATTER: 'frontmatter-multi',
 };
 
 /** Codes whose lint findings are fixable by `gbrain frontmatter validate --fix`. */
--- a/src/core/import-file.ts
+++ b/src/core/import-file.ts
@@ -301,6 +301,17 @@ export async function importFromContent(
   // silently fabricated a duplicate at (default, slug) — causing later
   // bare-slug subqueries (getTags, deleteChunks, etc.) to crash with 21000.
   const sourceId = opts.sourceId;
+  // #2822: reject empty/whitespace-only content before any work happens. An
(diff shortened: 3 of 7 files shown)


time-attack:
Closing this one. The stdin guard in cli.ts is a genuinely good catch, but it arrives bundled with three unrelated hardening changes, and the importFromContent throw is the part I can't take: every other rejection in that function returns {status:'skipped', error} and importFile routes file syncs through it, so a zero-byte placeholder note in someone's vault would start throwing mid-sync. The new MULTI_FRONTMATTER heuristic is also a new validation rule rather than a fix to something broken. Thanks for the writeup — the repro detail was clear and made this quick to evaluate.

PR #3164, opened 2026-07-21
fix(takes): persist think --take rows + review-first take proposals (#2556, #2411)

### LINKED ISSUE #2411: gbrain takes propose is documented as the promotion path for take_proposals but the CLI command is missing

### LINKED ISSUE #2556: think --take silently writes no take row (CLI and MCP)

time-attack:
Fix-wave unit c40-2-f1: takes lifecycle (propose / persist).

## Items

- **Fixes #2556** — `think --take` silently wrote no take row (CLI and MCP). `RunThinkOpts.take` was declared but never consumed; no persistence path existed. Now `persistThinkTake` appends the synthesis answer as the next take row on the anchor page, refuses empty/no-LLM synthesis (`TAKE_EMPTY_NOT_PERSISTED`), and the CLI exits non-zero with an explicit message when no row is written. Remote callers stay blocked at the trust boundary and now get an explicit `REMOTE_PERSISTED_BLOCKED` warning.
- **Fixes #2411** — `gbrain takes propose` was documented as the only queue→canonical promotion path (D17) but the CLI command didn't exist. Adds `takes propose list/accept/reject` (txn + `FOR UPDATE` + advisory lock, markdown rollback on DB failure, idempotent accept/reject, dedup guard) plus `takes_propose_list/accept/reject` ops.
- **Takeover of #2618** (credit @javieraldape) — rebased onto master's `thinkSourceScopeOpts` split (#2739); logic otherwise landed as reviewed.
- **Takeover of #2418** (credit @Mr-B-1) — rebased onto master's source-resolver-threaded takes dispatcher, plus the review repairs below.

## Repairs over the original PRs

- **Source-isolation invariant:** the three proposal op handlers hand-rolled scoping as `ctx.remote ? ctx.sourceId : undefined` — fail-open when `ctx.remote` is `undefined`. They now route through `sourceScopeOpts(ctx)` (federated array > scalar > nothing), and `listTakeProposals` / `acceptTakeProposal` / `rejectTakeProposal` enforce `sourceIds[]` scope.
- **Trust boundary:** `takes_propose_accept` is `localOnly` — it writes the source markdown file under `sync.repo_path` (same class as `sync`), and D17 makes operator opt-in the only promotion path. Both write ops carry `mutating: true`.
- Corrected the stale D17 doc comment in `propose-takes.ts` to the shipped syntax.

## Tests

- `test/take-proposals.test.ts`: full accept/reject/list coverage from #2418, plus new tests for federated-scope filtering, out-of-scope accept/reject refusal, and the fail-closed `ctx.remote === undefined` handler path (fails without the fix).
- `test/think-pipeline.serial.test.ts`: `persistThinkTake` happy path + empty-synthesis refusal on real PGLite.
- `test/takes-mcp-allowlist.serial.test.ts`: asserts the explicit `REMOTE_PERSISTED_BLOCKED` warning.

Gate: `bun run typecheck` exit 0; touched test files green (`take-proposals` 10 pass, `think-pipeline.serial` 31 pass, `takes-mcp-allowlist.serial` 13 pass, `takes-command-source-scope` 3 pass, `parity`/`mcp-tool-defs`/`cli-args` 20 pass).

### DIFF
--- a/src/commands/takes.ts
+++ b/src/commands/takes.ts
@@ -29,6 +29,11 @@ import {
 } from '../core/takes-fence.ts';
 import { withPageLock } from '../core/page-lock.ts';
 import { resolveSourceId } from '../core/source-resolver.ts';
+import {
+  acceptTakeProposal,
+  listTakeProposals,
+  rejectTakeProposal,
+} from '../core/take-proposals.ts';
 
 // --- Helpers ---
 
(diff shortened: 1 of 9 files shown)


time-attack:
Closing. The #2556 half — a declared take flag that nothing consumed — is a legitimate finding, but it's carrying about a thousand lines of net-new feature with it: a new take-proposal module, three new MCP ops, and a new CLI subcommand tree. New feature surface is a default no here regardless of how well it's built, and this also overlaps heavily with #3167. The source-isolation repairs you made over the original PRs (routing through sourceScopeOpts instead of the hand-rolled ctx.remote check) were the right instinct and I appreciate you calling them out explicitly.

PR #3166, opened 2026-07-21
fix(takes): supersede targets active rows; reserved list subcommand; serve --http holder allow-list; derive-phase privacy excludes

### LINKED ISSUE #2078: takes supersede looks up its target with active:false — superseding any ACTIVE take fails "Row #N not found"

### LINKED ISSUE #2079: `gbrain takes list` parses "list" as a page slug and prints "No takes on list."
## Summary

`gbrain takes list` prints `No takes on list.` even when the brain holds many takes. Root cause: the takes CLI has no `list` subcommand — `cmdList` treats `args[0]` as a **page slug** (takes.ts:110–115), so the word `list` is looked up as a page named "list", which has no takes.

```
$ gbrain takes list
No takes on list.          # ← "list" was parsed as a slug

$ gbrain call takes_list '{}'
[ ...all takes returned fine... ]
```

It's a natural thing to type (every other noun-style CLI surface supports `list`), and the failure mode reads like "the takes table is empty" — an agent or user reasonably concludes there are no takes and moves on.

## Suggested fix (either works)

- Support a global list: `gbrain takes list` (or bare `gbrain takes`) → all active takes, i.e. CLI parity with the `takes_list` operation; or
- Reserve `list` as a word and error with the usage string instead of silently slug-ifying it.

## Related papercut (same discovery session)

`gbrain eval --help` (and `gbrain eval suspected-contradictions --help`) prints the generic "run gbrain --help for the full command list" stub rather than the eval subcommand's actual usage — the only way to learn the `run/trend/review` sub-subcommands and the `--queries-file | --query | --from-capture` requirement is to read `src/commands/eval-suspected-contradictions.ts`. Happy to split this into its own issue if preferred.

### LINKED ISSUE #2529: serve --http ignores permissions.takes_holders for legacy bearer tokens — remote MCP callers always see only holder='world' takes (v0.42.52.0)
### Summary

`gbrain auth permissions <token> set-takes-holders ...` has no effect when the brain is served via `gbrain serve --http`. Remote MCP callers authenticated with a legacy bearer token always get the fail-closed default allow-list `['world']`, regardless of what `access_tokens.permissions.takes_holders` says. Since takes written by the owner/agent typically carry `holder=brain` or `holder=people/<slug>`, every takes surface over HTTP MCP (`takes_list`, `takes_search`, `takes_scorecard`, `takes_calibration`) returns empty/zero for them, while the same ops via local CLI (`gbrain call takes_list ...`) return the rows.

Version: 0.42.52.0, PGLite engine, macOS (arm64), Bun 1.3.11.

### Repro

```bash
gbrain auth create my-token                      # default takes_holders=["world"]
gbrain auth permissions my-token set-takes-holders world,brain
gbrain takes add notes/decision-ledger --claim "test claim" --kind take --who brain --weight 0.75
gbrain serve --http --port 3131 &

curl -s -X POST http://127.0.0.1:3131/mcp \
  -H "Authorization: Bearer <token>" -H "Content-Type: application/json" \
  -H "Accept: application/json, text/event-stream" \
  -d '{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"name":"takes_list","arguments":{}}}'
# → "[]"   (expected: the holder=brain row)

gbrain call takes_list '{}'   # local CLI, serve stopped
# → returns the row
```

I verified the permissions row itself is stored correctly (proper jsonb object, `{"takes_holders":["world","brain"]}`), and that `listTakes` returns the rows when handed that allow-list directly — the drop happens in auth threading.

### Root cause

`GBrainOAuthProvider.verifyAccessToken` (`src/core/oauth-provider.ts`), legacy `access_tokens` branch: it parses `permissions` and reads `permissions.source_id` for the source grant, but never reads `permissions.takes_holders`, and the returned AuthInfo carries no `takesHoldersAllowList` field.

The consumer in `src/commands/serve-http.ts` (POST /mcp handler) then does:

```ts
const tokenAllowList = (authInfo as AuthInfo & { takesHoldersAllowList?: string[] }).takesHoldersAllowList
  ?? ['world'];
```

so the fallback `['world']` always wins for legacy bearer tokens. (The OAuth-client branch of `verifyAccessToken` doesn't thread it either.)

Note the older transport (`src/mcp/http-transport.ts`) does this correctly — its `verifyAccessToken` returns `takesHoldersAllowList` read from `permissions.takes_holders` and threads it into `dispatchToolCall`. The regression is that the OAuth-provider path used by `serve --http` never got the same plumbing, so the documented `--takes-holders` / `set-takes-holders` surface is a silent no-op there.

### Suggested fix

In the legacy branch of `GBrainOAuthProvider.verifyAccessToken`, mirror the source-grant handling:

```ts
const holdersRaw = permissions && typeof permissions === 'object'
  ? (permissions as Record<string, unknown>).takes_holders
  : undefined;
const takesHoldersAllowList = Array.isArray(holdersRaw)
  ? (holdersRaw as unknown[]).filter((h): h is string => typeof h === 'string')
  : undefined;
return {
  ...,
  ...(takesHoldersAllowList ? { takesHoldersAllowList } : {}),
} as CoreAuthInfo as SdkAuthInfo;
```

I applied exactly this locally and the curl repro above returns the `holder=brain` rows; tokens without the key keep the fail-closed `['world']` default. The OAuth-client branch probably wants an equivalent (e.g. from `oauth_clients` metadata) but I only exercised the legacy-token path.

Happy to turn this into a PR if useful.

### LINKED ISSUE #2780: bug(privacy): derive phases (chronicle/facts/atoms/takes) re-materialize excluded-prefix pages as searchable pages outside the prefix — no containment mechanism gates derivation
**What happened?**

chronicle_backfill / chronicle_extract swept a page under the operator's `GBRAIN_SEARCH_EXCLUDE` prefix (sensitive third-party HR content under a zero-propagation hold) and wrote **8 derived `life/events/*` pages OUTSIDE the excluded prefix** — searchable once embedded, defeating the exclusion. Observed 2026-07-08, job #23157, `events_written=8`. Each derived page carries `event.depth: <excluded-page-path>` frontmatter (`extract-events.ts:150-153`), which is what proves the leak path.

Verified at HEAD (a25209bb): `GBRAIN_SEARCH_EXCLUDE` is consumed exclusively in the search-ranking SQL (`source-boost.ts` `resolveHardExcludes` → `postgres-engine.ts:1568/1717/1838`). None of the derive phases apply ANY exclusion when selecting source pages:

- **chronicle** — `operations.ts:5293-5294`, `listPages` by type, no prefix filter
- **facts** — `extract-facts.ts:176-185`, `getAllSlugs()` = every page in the brain
- **atoms** — `extract-atoms.ts:166-195`, raw SELECT with a `dream_generated` guard but no prefix exclusion
- **takes** — `extract-takes.ts:197-199`, `listAllPageRefs()` = all live pages

Derived output lands under `life/events/` and `atoms/` — outside any excluded prefix, embedded and searchable.

To pre-empt a fair objection: `GBRAIN_SEARCH_EXCLUDE` is documented as a search-ranking knob (source-boost.ts:12-13), so the derive phases are not violating its stated contract. The defect is that **no containment mechanism exists that derivation honors**: content an operator has excluded from search gets re-materialized by the engine as new, searchable pages outside the excluded prefix. Cleanup is not a fix — nothing prevents recurrence on the next derive pass, so this is an ongoing privacy-risk class.

**What did you expect?**

Derive phases (chronicle, facts, atoms, takes) to honor a containment prefix. Either:
- skip pages under excluded prefixes by default (extend `GBRAIN_SEARCH_EXCLUDE` to gate derivation), or
- file derived output under the source page's prefix (so exclusion follows the content), or
- add a dedicated containment/privacy prefix that both search AND derivation respect.

**Steps to reproduce**

1. `GBRAIN_SEARCH_EXCLUDE=private/`; create a meeting-type page under `private/`.
2. Run `chronicle_backfill` (or trigger the put_page backstop on that page).
3. Observe `life/events/*` pages derived from the private page, outside `private/`, carrying `event.depth: private/…` — searchable once embedded.

**Environment**

Not relevant (no doctor check covers this; that absence is part of the point).

---

Local mitigation: derived pages enumerated via the `event.depth` marker and queued for operator-approved deletion after each pass — a treadmill, not a fix.

time-attack:
Four adversarially verified backlog fixes in the takes lifecycle / gateway component (work unit c40-1-f1).

- Fixes #2078 — `takes supersede` looked up its target with `active: false` (src/commands/takes.ts), so superseding any ACTIVE take failed "Row #N not found". The lookup now targets active rows, and an already-superseded row gets a distinct "already superseded" error instead of "not found".
- Fixes #2079 — `gbrain takes list` parsed "list" as a page slug and printed "No takes on list." `list` is now a reserved subcommand: bare `takes list` lists all active takes brain-wide (CLI parity with the `takes_list` op, page slug shown per row), `takes list <slug>` scopes to one page.
- Fixes #2529 — `serve --http` ignored `permissions.takes_holders` for legacy bearer tokens: `GBrainOAuthProvider.verifyAccessToken` never read the grant, so the transport's `['world']` fallback always won and `gbrain auth set-takes-holders` was a silent no-op on that path. The legacy branch now parses the grant (string[] validated) and threads `takesHoldersAllowList` on `AuthInfo`, mirroring `src/mcp/http-transport.ts`. Default stays `['world']` (fail-closed) when unset; OAuth clients are unchanged.
- Fixes #2780 — derive phases (chronicle event extraction, extract_facts, extract_atoms, extract_takes fs+db) re-materialized excluded-prefix pages as searchable rows/pages outside the excluded prefix. All four now skip source pages under `GBRAIN_SEARCH_EXCLUDE` prefixes via one shared helper pair (`resolveDeriveExcludes` + `isHardExcludedSlug` in `src/core/search/source-boost.ts`). Scoping note: the gate deliberately covers the env-configured privacy prefixes only — the `DEFAULT_HARD_EXCLUDES` noise prefixes (`test/`, `attachments/`, `.raw/`) remain derivable because existing suites pin extraction from `test/` pages and those defaults are a ranking concern, not a privacy boundary.

Tests: new `test/takes-cli-list-supersede.test.ts` + `test/derive-hard-excludes.test.ts`; extended `test/oauth.test.ts` (legacy takes_holders grant + fail-closed default) and `test/extract-atoms-page-discovery.test.ts` (excluded page never atomized). Verified the new tests fail on unfixed master (4 failures) and pass with the fix. `bun run typecheck` exit 0; 267 tests across the 17 touched/adjacent files pass.

### DIFF
--- a/src/commands/takes.ts
+++ b/src/commands/takes.ts
@@ -128,11 +128,9 @@ function writeBody(path: string, body: string): void {
 // --- Subcommands ---
 
 async function cmdList(engine: BrainEngine, args: string[]): Promise<void> {
-  const slug = args[0];
-  if (!slug) {
-    console.error('Usage: gbrain takes <slug> [--json]');
-    process.exit(1);
-  }
+  // #2079: `gbrain takes list` (no slug) lists all active takes brain-wide —
+  // CLI parity with the takes_list op. A leading flag also means "no slug".
+  const slug = args[0] && !args[0].startsWith('--') ? args[0] : undefined;
   const json = flagPresent(args, '--json');
   const holder = flagValue(args, '--who');
   const kind = flagValue(args, '--kind') as string | undefined;
@@ -153,16 +151,17 @@ async function cmdList(engine: BrainEngine, args: string[]): Promise<void> {
   }
 
   if (takes.length === 0) {
-    console.log(`No takes on ${slug}.`);
+    console.log(slug ? `No takes on ${slug}.` : 'No takes in brain.');
     return;
   }
-  console.log(`# Takes on ${slug}\n`);
+  console.log(slug ? `# Takes on ${slug}\n` : '# All takes\n');
   for (const t of takes) {
     const tag = t.active ? '' : ' [superseded]';
     const w = Number(t.weight).toFixed(2);
     const since = t.since_date ?? '';
     const src = t.source ? ` — ${t.source}` : '';
-    console.log(`#${t.row_num} [${t.kind} • ${t.holder} • w=${w}${since ? ` • ${since}` : ''}]${tag}\n  ${t.claim}${src}\n`);
+    const ref = slug ? `#${t.row_num}` : `${t.page_slug}#${t.row_num}`;
+    console.log(`${ref} [${t.kind} • ${t.holder} • w=${w}${since ? ` • ${since}` : ''}]${tag}\n  ${t.claim}${src}\n`);
   }
 }
 
@@ -290,11 +289,17 @@ async function cmdSupersede(engine: BrainEngine, args: string[], sourceId?: stri
   await withPageLock(slug, async () => {
     const pageId = await getPageId(engine, slug, sourceId);
 
-    // Read existing row to inherit kind/holder unless overridden
-    const existing = await engine.listTakes({ page_id: pageId, active: false, limit: 500 });
+    // Read existing row to inherit kind/holder unless overridden.
+    // #2078: the supersede target is an ACTIVE take — active:false could
+    // only ever find already-superseded rows, so every real supersede failed.
+    const existing = await engine.listTakes({ page_id: pageId, active: true, limit: 500 });
     const target = existing.find(t => t.row_num === rowNum);
     if (!target) {
-      console.error(`Row #${rowNum} not found on ${slug}.`);
+      const superseded = await engine.listTakes({ page_id: pageId, active: false, limit: 500 });
+      const stale = superseded.find(t => t.row_num === rowNum);
+      console.error(stale
+        ? `Row #${rowNum} on ${slug} is already superseded.`
+        : `Row #${rowNum} not found on ${slug}.`);
       process.exit(1);
     }
     const kind = ensureKind(flagValue(args, '--kind') ?? target.kind);
@@ -554,6 +559,7 @@ export async function runTakes(engine: BrainEngine, args: string[]): Promise<voi
 Subcommands:
   takes <slug> [--json] [--who h] [--kind k] [--sort weight|since_date|created_at] [--expired]
                                           List takes for a page
+  takes list [<slug>] [same flags]        List all active takes (or one page's)
   takes search "<query>" [--limit N] [--json]
                                           Keyword search across all takes
   takes add <slug> --claim "..." --kind <fact|take|bet|hunch> --who <holder>
@@ -583,6 +589,9 @@ Common flags:
   const rest = args.slice(1);
 
   switch (sub) {
+    // #2079: reserved word — never a page slug. Bare `takes list` lists all
+    // active takes; `takes list <slug>` behaves like `takes <slug>`.
+    case 'list':        return cmdList(engine, rest);
     case 'search':      return cmdSearch(engine, rest);
     case 'add':         return cmdAdd(engine, rest, await resolveTakesSourceId(engine));
     case 'update':      return cmdUpdate(engine, rest, await resolveTakesSourceId(engine));
--- a/src/core/chronicle/extract-events.ts
+++ b/src/core/chronicle/extract-events.ts
@@ -10,6 +10,7 @@
 // configured it returns zero events (auto-emit is a no-op, never an error).
 import type { BrainEngine } from '../engine.ts';
 import { computeContentHash } from '../ingestion/types.ts';
+import { isHardExcludedSlug, resolveDeriveExcludes } from '../search/source-boost.ts';
 
 export interface ChronicleEventProposal {
   when: string;            // ISO datetime or YYYY-MM-DD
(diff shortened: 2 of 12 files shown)


Masashi-Ono0611:
Independent confirmation from a downstream fork ([Masashi-Ono0611/gbrain](https://github.com/Masashi-Ono0611/gbrain)): hit exactly the `takes list` reserved-word bug (#2079) this morning after running `takes extract --from-pages`. `gbrain takes list` silently parsed "list" as a page slug and reported no takes, even though ~1176 rows existed (confirmed via `doctor`'s `takes_count`).

Backported the relevant hunk of this PR locally and re-verified: `gbrain takes-list --limit 5` now returns real data. Thanks for the fix — hope this helps prioritize the merge.

time-attack:
Merging. I verified two of these directly against master: verifyAccessToken genuinely never reads permissions.takes_holders, so the legacy bearer path always lands on the ['world'] fallback and set-takes-holders is a silent no-op — an operator thinks that restriction is enforced when it isn't. The derive-phase exclude gate is likewise absent on master, and re-materializing excluded-prefix pages outside their prefix is a privacy boundary leak. Heads up that the #2078 supersede lookup already landed on master since you branched; your hunk converges on the same active:true value so it merges clean, but that item is now a no-op.

PR #3167, opened 2026-07-21
feat(takes): review-first take proposal list/accept/reject over MCP + CLI

### LINKED ISSUE #2269: MCP: expose takes_propose_accept / takes_propose_reject so Hermes/OpenClaw can review proposals natively in chat

`gbrain takes propose` is a TTY-only interactive CLI. There is no way for an agent platform (Hermes, OpenClaw) to surface the proposal review queue natively in a chat window. The MCP operation surface exposes read-only takes tools (`takes_list`, `takes_search`, `takes_scorecard`) but has no `takes_propose_accept` or `takes_propose_reject` operation.

This is my default state for running gbrain as a server-side brain backing a chat agent, id be interested to know if others would get value from this, I'll happily build it out of people would use it:

- The agent (Hermes/OpenClaw) runs on a server with no TTY
- `gbrain takes propose` exits 1 with no output when stdin is not a terminal
- `propose_takes` runs every autopilot cycle, building up a queue the user can never drain without SSHing into the server
- Proposals accumulate silently with no notification and no chat-native way to process them

The source comment in `cycle/propose-takes.ts` says:

> *"Operator opt-in via `gbrain takes propose --accept N` is the only path from queue to canonical fence (D17)."*

The intent is right — silent auto-acceptance should stay rejected. But the only blessed review path being a TTY CLI makes the takes proposal workflow a dead end for a primary deployment target.

## What does the solution look like?

Add three MCP operations to the agent-facing surface:

typescript
// List pending proposals with queue-specific filter
takes_propose_list(limit?: number, offset?: number): ProposalRow[]

// Accept — promotes to canonical takes fence on the page
takes_propose_accept(proposal_id: string): { slug: string, row_num: number }

// Reject — marks acted_at, sets status='rejected', never re-proposed
takes_propose_reject(proposal_id: string, reason?: string): { ok: true }

These map directly to the existing `take_proposals` table columns (`id`, `acted_at`, `acted_by`, `status`, `promoted_row_num`) — no schema changes needed. The promotion logic already exists in the CLI path; it just needs to be exposed as an operation callable over MCP.

**Trust boundary:** `takes_propose_accept` and `takes_propose_reject` should be classified as control-plane operations (same tier as `put_page`) — gated by write scope, not read scope. This aligns with the three-plane cut proposed in RFC #976.

The resulting chat-native UX:

Agent: You have 12 pending takes proposals. Here are the first 5:

  [take|0.75] ai-coding-coach: "Loop 3 is the hardest and most
  valuable coaching loop; without it you have a feedback-free product."

  Accept / Reject / Skip?

User: accept

Agent: ✅ Promoted to takes fence on research/ai-coding-coach-strategic-analysis #row4
       Next proposal: ...

The TTY CLI stays as an alternative path — this is purely additive.

## Alternatives considered

**Auto-accept above a confidence threshold** — e.g. automatically promote proposals with `weight >= 0.8`. Rejected: defeats the purpose of human review, and the source code explicitly documents that silent acceptance is the wrong default (D17).

**Pipe mode for the existing CLI** (`gbrain takes propose --pipe` outputting JSON to stdout) — would unblock scripted workflows but still doesn't give the agent a first-class review surface. A shell wrapper could work as a short-term shim but it's the wrong abstraction; the agent should be calling an operation, not parsing CLI output.

**Agent polls and batch-accepts via direct DB write** — possible today as a workaround (the schema is straightforward) but bypasses the fence promotion logic in `takes.ts`, risks markdown/DB divergence.

## References

- RFC #976 — Memory as a coherent operation surface (three-plane cut of `operations.ts`)
- `src/core/cycle/propose-takes.ts` — D17 schema spec, idempotency contract
- `src/commands/takes.ts` — existing CLI `propose` subcommand (TTY path)
- README v0.36.1.0 — `propose_takes` / `grade_takes` / `calibration_profile` phases described as PROTECTED

time-attack:
Fixes #2269
Takeover of #2418 (credit: @Mr-B-1 — co-authored on the commit; salvages the operation trio, promotion transaction, and CLI surface from that PR)

## What

The `take_proposals` queue built by the `propose_takes` cycle had no non-TTY review path: no MCP operation and no CLI subcommand could list, accept, or reject proposals, so server-side agent deployments accumulated an undrainable queue. This adds:

- `takes_propose_list` (scope: `read`), `takes_propose_accept` / `takes_propose_reject` (scope: `write`) in the shared operation registry — generated into both the CLI and MCP surfaces, so agents can review proposals natively in chat.
- `gbrain takes propose list|accept|reject` CLI subcommands.
- Accept promotes atomically: appends to the page's markdown takes fence, mirrors to the `takes` table in the same transaction (advisory-locked, duplicate-guarded), stamps the proposal, and restores the markdown if the DB mirror fails. Idempotent re-accept/re-reject.
- D17 posture preserved: nothing auto-accepts; accept is an explicit reviewed action.

## Repairs over the original PR

- **Source isolation**: handlers route through `sourceScopeOpts(ctx)` (federated grant array > scalar source > nothing) instead of `ctx.remote ? ctx.sourceId : undefined`, which ignored federated `allowedSources` grants and was fail-open when `remote` was unset.
- **Holder privacy (D4)**: all three ops thread `ctx.takesHoldersAllowList`, so an allow-listed MCP token can neither list nor act on proposals for holders outside its grant (empty list matches nothing).
- **Containment guard**: the accept path refuses any `page_slug` that would resolve a markdown write outside the brain directory (the op is remote-callable).
- **Rebased** onto master's takes.ts source-scope refactor; invalid `status` values now error instead of silently returning nothing.

## Tests

`test/take-proposals.test.ts` — real in-memory PGLite engine (no SQL fake): scope isolation (scalar/federated/holder allow-list, plus a dispatch-level regression test that fails against the original PR's handler), promotion + markdown mirror + since-date threading, rollback on injected DB failure, duplicate guard, reject lifecycle, and operation-registry scope/schema assertions.

- `bun test test/take-proposals.test.ts` — 14 pass / 0 fail
- Neighboring suites (`mcp-tool-defs`, `operations-allow-list`, `operations-trust-boundary`, `operations-descriptions`, `takes-command-source-scope`, `takes-source-scope`, `takes-mcp-allowlist.serial`, `takes-fence`, `operation-context-sourceid-required`) — 116 pass / 0 fail
- JSONB CI guards clean

### DIFF
--- a/src/commands/takes.ts
+++ b/src/commands/takes.ts
@@ -29,6 +29,11 @@ import {
 } from '../core/takes-fence.ts';
 import { withPageLock } from '../core/page-lock.ts';
 import { resolveSourceId } from '../core/source-resolver.ts';
+import {
+  acceptTakeProposal,
+  listTakeProposals,
+  rejectTakeProposal,
+} from '../core/take-proposals.ts';
 
 // --- Helpers ---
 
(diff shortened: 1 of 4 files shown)


time-attack:
Closing. This is net-new feature surface — a new review queue, new core module, and new MCP tools — and new features are a default no on this repo regardless of implementation quality. It also duplicates the #2411 work carried inside #3164, so the two overlap directly. The transaction discipline (FOR UPDATE plus the advisory lock and markdown rollback) was well thought through; that care just isn't enough on its own to open a new surface here.

PR #3168, opened 2026-07-21
fix: backlog part4-12 — four verified singleton PR takeovers (#2754, #2764, #2817, #2851)

time-attack:
Four adversarially-verified fix-needed community PRs, salvaged and rebased onto master. One commit per item, each crediting the original author. References only — closes nothing.

## Takeover of #2754 — preserve content_created_at through re-ingest
The design was sound (nullable `pages.content_created_at`, COALESCE-preserve in both engines, all three schema sources updated) but its migration claimed version 124, which master had since taken (`page_search_vector_drop_compiled_truth`, #2704). Renumbered to **v125**, dropped the already-superseded fts-language test hunk (master removed the "is the latest migration" assertion), and kept master's `MARKDOWN_CHUNKER_VERSION` default in the putPage conflict resolution. `schema-embedded.ts` regenerated via `bun run build:schema`. Bootstrap-coverage guard passes (column covered by the schema blob).

## Takeover of #2764 — reuse PGLite engine for reindex-frontmatter
The engine-reuse fix is intact: the CLI dispatcher passes its connected engine into `reindexFrontmatterCli`, with an `ownsEngine`-gated disconnect for standalone callers, plus the regression test. The undisclosed, unrelated tini-probe changes (`spawn-helpers.ts` + supervisor-tini tests) were dropped — they belong in their own PR.

## Takeover of #2817 — categorize timeline_coverage
Kept only the one-line `timeline_coverage` addition to `BRAIN_CHECK_NAMES` (the check is emitted by `onboard/checks.ts` and was falling through to the warn-once 'meta' path). The `auto_chronicle` config half already landed on master via another PR, so it's dropped to avoid a duplicate entry. Added a categorizeCheck regression test.

## Takeover of #2851 — gate nightly quality probe on real outcomes
Kept the run-outcome filter + latest-row log-once guard (fixes the livelock where per-tick `rate_limited` rows kept the 24h gate closed forever) with both regression tests. Fixed the symmetric regression the original introduced: `no_embedding_key` now counts as a completed attempt, so a keyless brain warns + audit-logs once per 24h window instead of once per 5-minute autopilot tick. Added a regression test for that path.

## Verification
- `bun run typecheck` exits 0
- `bun test` on all touched files (doctor-categories, nightly-quality-probe, reindex-frontmatter-connect, effective-date, pglite-engine, migrate-extensions, fts-language-migration.serial, schema-bootstrap-coverage, bootstrap, migrate): 365 pass / 0 fail
- jsonb guards (`check-jsonb-pattern.sh`, `check-jsonb-params.mjs`) clean

### DIFF
--- a/scripts/backfill-content-created-at.ts
+++ b/scripts/backfill-content-created-at.ts
@@ -75,7 +75,10 @@ export async function backfillContentCreatedAt(
   const slugPrefix = opts.slugPrefix?.replace(/[\\%_]/g, (c) => '\\' + c) ?? null;
 
   while (true) {
-    const slugFilter = slugPrefix ? `AND slug LIKE $2 ESCAPE '\\\\'` : '';
+    // NOTE: ESCAPE takes a single character. In this TS source that's `'\\'`
+    // (one backslash at runtime) — the doubled `'\\\\'` form sends a 2-char
+    // escape string and Postgres rejects it with "invalid escape string".
+    const slugFilter = slugPrefix ? `AND slug LIKE $2 ESCAPE '\\'` : '';
     const params: unknown[] = [lastId];
     if (slugPrefix) params.push(slugPrefix + '%');
 
--- a/scripts/backfill-content-created-at.ts
+++ b/scripts/backfill-content-created-at.ts
@@ -1,0 +1,204 @@
+#!/usr/bin/env bun
+/**
+ * tasks-41o — one-off, idempotent backfill for `pages.content_created_at`
+ * (migration v125).
+ *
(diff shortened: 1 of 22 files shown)


time-attack:
Closing this as a bundle. The individual salvage work looks careful — dropping the undisclosed tini-probe changes from #2764 and catching the symmetric no_embedding_key regression in #2851 are both good calls. But a new nullable column with a v125 migration touching schema.sql, both engines, and the bootstrap probe set can't ride alongside three unrelated fixes; that migration number also goes stale the moment master takes another version, which is exactly the renumbering problem you hit with #2754 in the first place. Schema changes need to stand alone.

PR #3169, opened 2026-07-21
fix(doctor,think): file-plane pack resolution, page-vs-page conflicts, cycle-freshness tuning

time-attack:
Backlog fix wave part4-8-f1: three community PRs salvaged to their correct core, flaws repaired, rebased onto master. References only — closes nothing.

## Takeover of #2469 — doctor type_proliferation reads file-plane config (credit: @onchito-walks)

`checkTypeProliferation` and `checkPackUpgradeAvailable` passed `cfg: null` to `loadActivePack`, so a schema pack selected via `gbrain schema use` (tier-6 file-plane `~/.gbrain/config.json`) never resolved — both checks silently compared against the default pack. They now pass `loadConfig()`, keeping master's tier-4 `dbConfig` read, and `schema_pack` is registered in `KNOWN_CONFIG_KEYS`. The original PR's ~22 unrelated fork files (docs/workarounds cruft, doctor policy downgrades, model-pricing edits, gateway env stuffing, longmemeval changes) were dropped.

Test: `test/onboard-pack-upgrade-checks.test.ts` seeds a 2-type pack referenced only from a temp-home `config.json`; fails on master (check falls back to the default pack), passes with the fix. The file now pins `GBRAIN_HOME` to a temp dir so the host's real config can't leak into the suite.

## Takeover of #2479 — page-vs-page conflicts on takes-empty brains (credit: @Adityaranjan-VeloxAI)

The think synthesis prompt limited the Conflicts rule to take-vs-take contradictions, so pages-only brains surfaced zero conflicts. The rule now covers conflicting values across pages and/or takes with `[HIGH]/[MEDIUM]/[LOW]` severity tags and a VARIANTS enumeration rule. The original PR's example hardcoded contributor-brain slugs; replaced with generic placeholders (`finance/acme-pricing` vs `outreach/acme-scripts`) per the repo privacy rule, and the new test pins that.

Retrieval breadth knobs land as env vars (`GBRAIN_THINK_GATHER_LIMIT`, `GBRAIN_THINK_EXCERPT_LEN`, `GBRAIN_THINK_MAX_OUTPUT_TOKENS`), all defaults unchanged. The maintainer's env-vs-DB-config question from the original PR is left open deliberately — env-only matches the incident-escape-hatch pattern (pace mode, sync knobs) and DB keys can be layered under them later without breaking anything.

Test: `test/think-env-knobs.test.ts` (prompt rule + placeholder pinning + all three knobs).

## Takeover of #2505 — cycle-freshness tuning for nightly-cadence brains (credit: @sene1337)

`checkCycleFreshness` had env-only thresholds (warn 6h / fail 24h) — too tight for brains cycled by a nightly cron, and no way to scope the check to intentionally-dreamed sources. Adds DB-plane keys `doctor.cycle_freshness.warn_hours` / `fail_hours` / `source_allowlist` / `source_ignorelist` with env > config > default precedence (pace-mode pattern). Invalid scope entries warn instead of silently hiding sources; scoped results carry structured `details`.

The original PR's fallback to `DREAM_SOURCE_ALLOWLIST` — a fork-private cron env var with zero hits upstream — was dropped.

Tests: `test/doctor-cycle-freshness.test.ts` (+6 cases), `test/config-set.test.ts` key registration.

## Skipped: #2517 (needs maintainer coordination)

The salvageable core of #2517 is the `getHealth` deleted_at filtering, which directly collides with open PR #2525's `getHealth`/brain_score rewrite — landing it here would create a three-way merge conflict on the same SQL. The rest of the bundle is fork policy (subagent_capability downgrade naming a private fork's routing), PG16-only `IS JSON` syntax on the supported PG15/Supabase path, and speculative rechunk/pricing changes. Recommend resolving #2525 first, then rebasing the deleted_at filters as a scoped follow-up.

## Verification

- `bun test test/doctor-cycle-freshness.test.ts test/think-env-knobs.test.ts test/config-set.test.ts test/onboard-pack-upgrade-checks.test.ts test/think-max-output-tokens.test.ts` — 56 pass / 0 fail
- All new tests confirmed to fail without the src changes (6 fail + 1 import error on stashed src)


time-attack:
Closing. Two of the three items here are new configuration surface rather than fixes — three new think env knobs plus a prompt rewrite, and four new doctor.cycle_freshness keys with allowlist/ignorelist scoping — and your own writeup leaves the maintainer's env-versus-DB-config question open, so that surface isn't settled yet. The #2469 file-plane pack resolution bug is the one real fix in here and it's a few lines lost inside a 495-line bundle. Dropping the original PRs' unrelated fork cruft and scrubbing the hardcoded contributor slugs per the privacy rule were both the right moves.

PR #3170, opened 2026-07-21
fix(embed): dynamic embedding column write target — upsertChunks + stale scans (#1262)

### LINKED ISSUE #1262: PostgresEngine.upsertChunks ignores embedding_columns registry — writes 2560-dim embeddings into legacy vector(1536), fails with dimension mismatch

time-attack:
Fixes #1262
Takeover of #1263 (rebased onto current master; credit to @DmitryBMsk, co-authored on the commit)

## Problem

`upsertChunks` in BOTH engines hardcoded the legacy `embedding` column — the INSERT column list, the `::vector` cast, and every ON CONFLICT clause. The `embedding_columns` registry had read-side consumers only (PR #1164), so a brain configured with a registered alternate column (e.g. `halfvec(2560)`) could *search* the right column but failed every `put_page`/`import`/`sync`/`embed` write with a dimension mismatch.

## Fix (write-side symmetric to the read-side resolver)

- **`resolveWriteColumn(cfg)` / `resolveWriteColumnForEngine(engine)`** (`src/core/search/embedding-column.ts`): a user-declared registry entry whose `provider` matches the current embedding model resolves to a descriptor; no registry / no match ⇒ `undefined` (legacy `embedding`, no behavior change). Builtins are never consulted, so the multimodal builtin can't capture text writes.
- **`upsertChunks` accepts a caller-resolved `embeddingColumn`** in `postgres-engine.ts` AND `pglite-engine.ts` (lockstep): target column, cast (`::vector` / `::halfvec(N)`), and the v0.40.3.0 D24 ON CONFLICT race-fix CASE arms all follow the column. Engines stay config-free — the boundary resolves, same contract as the read side.
- **Stale scans follow the write column**: `countStaleChunks`, `listStaleChunks` (both cursor arms), and the shared `buildStaleChunkWhere` accept the descriptor. Without this an alt-column brain re-selects (and re-pays for) already-embedded chunks on every `embed --stale` run, forever.
- **Boundary threading**: `runEmbedCore` (embedPage / embedAll / embedAllStale), `embedStaleForSource` + the `embed-backfill` Minion handler, `importFromContent` / `importCodeFile` / `withImportTransaction`, and the contextual-retrieval re-embed transaction resolve once and pass the descriptor to every write.
- **`preflightDimMismatch`** skips the legacy-column dim comparison when a non-default write column resolved — it would otherwise hard-block `gbrain embed` on alt-column brains even after the write fix.

## What changed vs #1263

The original PR was correct in approach but targeted v0.37.6.0; the patched region was rewritten since (D24 ON CONFLICT race fix, batchRetry wrapper, signature-stale machinery, embed-backfill/keyset re-entry, updated_desc cursor arm). This re-threads the same design through the current code: D24 CASE semantics preserved per-column, both `listStaleChunks` cursor orders covered, `embed-stale.ts` + the Minion handler included, and the preflight dim-check made column-aware.

## Tests

- `test/search/embedding-column.serial.test.ts` — `resolveWriteColumn` unit coverage (no registry, cfg-model match, gateway-model match, no-match fallback, builtin-never-matches, malformed entry throws).
- `test/e2e/embedding-column-pglite.test.ts` — write lands in the alternate halfvec column (legacy stays NULL), text-unchanged re-upsert preserves the alt-column vector (D24 follows the column), stale count/list contrast, and an `embed --stale --dry-run` integration proving boundary resolution + convergence (reports 1-chunk perpetual backlog without the fix).
- `test/e2e/embedding-column-postgres.test.ts` — DATABASE_URL-gated twins for real pgvector.

All three fail without the src changes (verified by stashing src/): 3 fail → 0 fail.

Gates: `bun run typecheck` exit 0; touched test batch (17 files, 195 pass / 0 fail) + import/sync sanity batch (7 files, 70 pass / 0 fail); JSONB pattern + params guards clean.

### DIFF
--- a/src/commands/embed.ts
+++ b/src/commands/embed.ts
@@ -1,6 +1,7 @@
 import type { BrainEngine } from '../core/engine.ts';
 import { embedBatch, currentEmbeddingSignature } from '../core/embedding.ts';
-import type { ChunkInput } from '../core/types.ts';
+import type { ChunkInput, ResolvedColumn } from '../core/types.ts';
+import { resolveWriteColumnForEngine } from '../core/search/embedding-column.ts';
 import { chunkText } from '../core/chunkers/recursive.ts';
 import { createProgress, type ProgressReporter } from '../core/progress.ts';
 import { getCliOptions, cliOptsToProgressOptions } from '../core/cli-options.ts';
@@ -183,8 +184,13 @@ export class EmbeddingDimMismatchError extends Error {
  * fresh-install bug class at the very first invocation instead of letting
  * the worker pool hammer N pages with raw 22000 errors.
  */
-async function preflightDimMismatch(engine: BrainEngine, dryRun: boolean): Promise<void> {
+async function preflightDimMismatch(engine: BrainEngine, dryRun: boolean, embeddingColumn?: ResolvedColumn): Promise<void> {
   if (dryRun) return; // dry-run never embeds, no risk
+  // #1262: an alt-column brain writes to `embeddingColumn`, not the legacy
+  // `embedding` column — the legacy column's dims are irrelevant, and the
+  // registry entry (validated at resolve time) pins the target's dims. Only
+  // the legacy default path needs the schema-vs-gateway dim comparison.
+  if (embeddingColumn && embeddingColumn.name !== 'embedding') return;
   const { readContentChunksEmbeddingDim, embeddingMismatchMessage } = await import('../core/embedding-dim-check.ts');
   const { getEmbeddingDimensions, getEmbeddingModel } = await import('../core/ai/gateway.ts');
   let existing;
@@ -238,7 +244,12 @@ export async function runEmbedCore(engine: BrainEngine, opts: EmbedOpts): Promis
   // v0.37.11.0 (Lane D.2): pre-flight dim-mismatch check. Catches the headline
   // fresh-install bug class before the worker pool spends 20 parallel calls
   // hitting raw Postgres dimension errors.
-  await preflightDimMismatch(engine, !!opts.dryRun);
+  // #1262: resolve the write-side embedding column ONCE at the boundary
+  // (merged config + gateway model) and thread the descriptor through every
+  // upsertChunks / stale-scan below. undefined => legacy `embedding` column.
+  const embeddingColumn = await resolveWriteColumnForEngine(engine);
+
+  await preflightDimMismatch(engine, !!opts.dryRun, embeddingColumn);
 
   const result: EmbedResult = {
     embedded: 0,
@@ -253,7 +264,7 @@ export async function runEmbedCore(engine: BrainEngine, opts: EmbedOpts): Promis
     for (const s of opts.slugs) {
       if (isAborted(opts.signal)) break; // #1737: stop the per-slug loop on abort
       try {
-        await embedPage(engine, s, !!opts.dryRun, result, opts.sourceId, opts.signal);
+        await embedPage(engine, s, !!opts.dryRun, result, opts.sourceId, opts.signal, embeddingColumn);
       } catch (e: unknown) {
         serr(`  Error embedding ${s}: ${e instanceof Error ? e.message : e}`);
       }
@@ -347,7 +358,7 @@ export async function runEmbedCore(engine: BrainEngine, opts: EmbedOpts): Promis
         catchUp: opts.catchUp,
         pacer,
         paceMaxConcurrency,
-      }, opts.signal);
+      }, opts.signal, embeddingColumn);
     } finally {
       // E1: surface pacing telemetry (human + structured) when pacing was on.
       const snap = pacer.snapshot();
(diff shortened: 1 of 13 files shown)


time-attack:
Merging this one. I confirmed on current master that `_upsertChunksOnce` still hardcodes the legacy `embedding` column in both engines while the `embedding_columns` registry only has read-side resolvers, so an alt-column brain can search but never write — and the stale scans re-charge for already-embedded chunks on every run. The write-side resolver mirroring the read side, the lockstep engine changes, and the parity tests are the right shape. GitHub currently marks the branch as conflicting, so please rebase onto master before it goes in.

PR #3171, opened 2026-07-21
fix(takes): wire the takes.embedding writer so vector takes search works

### LINKED ISSUE #2089: takes.embedding has no writer — vector takes search (think takes_vec, takes search) is structurally dead; DDL also hardcodes vector(1536) regardless of configured embedder dims
## Summary

`takes.embedding` is **never written by any code path** — INSERTs into `takes` omit the column (postgres-engine.ts:4018 and :4393; pglite-engine.ts:3927) and no `UPDATE takes SET embedding` exists anywhere in the tree. The read side is fully built and shipping: `searchTakesByVector`, `getTakeEmbeddings` (postgres-engine.ts:4312), `countStaleTakes` (:4327), `listStaleTakes` (:4335), and the engine interface ("v0.28 stale-takes row", engine.ts:322/1429). `listStaleTakes`/`countStaleTakes` additionally have **zero callers** outside the engine layer — `embed --stale` covers pages/chunks only, and the dream cycle doesn't embed takes either, so there is no recovery knob. Verified identical on v0.42.37.0 and master (ecd6ae87, v0.42.40.0).

Consequences:

- `gbrain takes search` (vector-only) returns empty forever, on every install.
- `think`'s `takes_vec` gather arm (src/core/think/gather.ts:7) never contributes. With `takes_kw` inheriting #2077's AND-fragility for natural-language questions, `think` effectively runs **takes-blind** — observed `Takes: 0` in think's own stats line with 21 active takes in the DB, several directly relevant to the question asked.
- The synthesis then picks up take *summaries* from ordinary pages instead and can misstate claim text/scope (observed twice in one run: a take's meaning inverted, a narrow claim broadened) — the takes table is exactly the layer that was supposed to prevent this.

## Second defect: hardcoded dims

The takes DDL is static `vector(1536)` (schema-embedded.ts:289) while init's dim-substitution gives `content_chunks` the configured embedder dims — a ZE-configured live DB shows `content_chunks.embedding vector(1280)` and `takes.embedding vector(1536)` side by side. The init.ts:1017 comment acknowledges the static-substitution gap for Postgres. Net: even when a writer ships, any non-1536 embedder will fail the insert or the `embedding <=> query_vec` comparison until the column dims align with the configured embedder.

## Repro (v0.42.37.0, Postgres engine, any embedder)

1. `gbrain takes add` a few takes; run sync + a full dream + `gbrain embed --stale`.
2. `SELECT count(*) FROM takes WHERE active AND embedding IS NULL;` → equals the total active count, indefinitely.
3. `gbrain takes search "<distinctive term from a claim>"` → empty. `gbrain think "<question relevant to a take>"` → `Takes: 0`.

## Ask

1. Add the takes embed pass (claim text → embedding) — natural homes: the dream cycle's embed phase and/or `embed --stale`; the `listStaleTakes` machinery is already there waiting for a caller.
2. Create the takes column with the configured embedder dims (the same substitution `content_chunks` gets), and/or migrate existing `vector(1536)` columns on upgrade.

Cross-refs: #2077 (keyword-arm AND-fragility — currently the only live takes-retrieval arm), #2078/#2079 (takes CLI family).

time-attack:
## What

Fixes #2089

`takes.embedding` had a full read side — `searchTakesVector`, think's `takes_vec` gather arm, `getTakeEmbeddings`, plus `countStaleTakes`/`listStaleTakes` stale tracking in both engines — but **no writer anywhere in `src/`**. The vector takes arm was structurally dead on every install.

## Changes

- **`BrainEngine.updateTakeEmbeddings`** (postgres-engine + pglite-engine in lockstep): writes `embedding` + `embedded_at` for active takes. `updated_at` deliberately untouched — embeddings are derived data, not a content change.
- **`embedStaleTakes`** (`src/commands/embed.ts`): claim-text backfill wired into the `--stale` path, so `gbrain embed --stale`, the cycle/dream embed phase (`runEmbedCore({stale:true})`), and sync auto-embed all populate takes embeddings. Log-and-skip failure semantics (stops on first failure, remaining takes retry next run); dry-run counts into `would_embed`; `EmbedResult.takes_embedded` reports the count.
- **Migration v125 `takes_embedding_dim_align`**: the takes DDL hardcoded `VECTOR(1536)` while content_chunks/facts resolve dims from config, so non-1536 embedder brains could never take writes even with a writer. Because the column was writer-less it is all-NULL on every install, so a guarded DROP/ADD retype to the configured dims is lossless — and works on PGLite, which can't `ALTER COLUMN TYPE vector(N)`. No-op when dims already match; refuses (with a stderr note) if the column somehow holds data. All takes access uses explicit column lists, so the column-order change is safe.
- **Test** `test/embed-stale-takes.test.ts`: fails without the fix — backfill flips `countStaleTakes` 2→0, `searchTakesVector` goes from dead to returning ranked hits, dry-run counts without writing, second run is a no-op, and v125 retypes an empty 1536 column to config dims with a working write path at the new dims.
- `takes.embedding` added to `COLUMN_EXEMPTIONS` in the bootstrap-coverage guard (the takes table is migration-created; v125's ADD COLUMN is a retype, not a new column).

## Not covered (deliberate)

The `embed-backfill` minion handler calls `embedStaleForSource` directly (per-source chunk pipeline) and does not run the takes pass; takes get picked up by the next cycle embed phase or any `embed --stale`. Takes stale-listing is brain-wide (existing `listStaleTakes` shape), not source-scoped — a write-side backfill, not a read-side scope concern.

## Verification

- `bun run typecheck` exit 0
- `bun test test/embed-stale-takes.test.ts test/embed.serial.test.ts test/embed-stale.serial.test.ts test/schema-bootstrap-coverage.test.ts test/takes-engine.test.ts` → 76 pass / 0 fail
- `bun test test/migrate.test.ts` → 161 pass / 0 fail
- jsonb guards (`check-jsonb-pattern.sh`, `check-jsonb-params.mjs`) clean

### DIFF
--- a/src/commands/embed.ts
+++ b/src/commands/embed.ts
@@ -131,6 +131,13 @@ export interface EmbedResult {
   /** True if this run was a dry-run. */
   dryRun: boolean;
   /**
+   * #2089: take claims newly embedded in this run (the `--stale` path also
+   * backfills `takes.embedding` so the vector takes arm — think takes_vec —
+   * has data to search). Present only when > 0. In dryRun mode stale takes
+   * are counted into `would_embed` instead.
+   */
+  takes_embedded?: number;
+  /**
    * E1 (paced-backfill): end-of-run pacing telemetry. Present ONLY when pacing
    * was active (enabled bundle). The number the operator could not get from an
    * external wrapper ("zero pauses" ≠ "queue safe").
@@ -644,7 +651,17 @@ async function embedAll(
     // D7: thread sourceId so `gbrain embed --stale --source X` actually scopes.
     // v0.41.18.0 (A13): thread batchSize/priority/catchUp into the stale path.
     // #1737: thread the external abort signal so the cycle embed phase bails.
-    return await embedAllStale(engine, sourceId, dryRun, result, onProgress, staleOpts, signature, signal);
+    await embedAllStale(engine, sourceId, dryRun, result, onProgress, staleOpts, signature, signal);
+    // #2089: backfill takes.embedding — before this pass nothing ever wrote
+    // it, so the vector takes arm (searchTakesVector / think takes_vec) was
+    // structurally dead. Runs on every --stale caller (CLI, cycle embed
+    // phase, sync auto-embed). listStaleTakes is brain-wide (not
+    // source-scoped) — embedding another source's takes is a harmless
+    // idempotent write, not a read-side leak.
+    if (!isAborted(signal)) {
+      await embedStaleTakes(engine, dryRun, result, signal);
+    }
+    return;
   }
 
   // --all path: pacer (no-op when off). E-1: lower the worker count to the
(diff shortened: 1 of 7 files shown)


time-attack:
Merging. Master really does have the whole read side — `searchTakesVector`, `getTakeEmbeddings`, `countStaleTakes` — with no writer anywhere in `src/`, so the vector takes arm has been dead on every install; wiring the writer is a correctness fix, not new surface. The lockstep engine methods and the fail-forward log-and-skip semantics look right. One blocker: master already ships migration version 125, so please renumber `takes_embedding_dim_align` when you rebase.

Issue #1626, opened 2026-05-29
`--source __all__` cross-source query/search intermittently returns "No results" (non-deterministic); single-source is reliable


mrudoy:
## Summary

`gbrain query "<q>" --source __all__` (and `--source-id __all__`, and the MCP `query`/`search` tools with `source_id: "__all__"`) **intermittently returns "No results" for queries that single-source scoping returns reliably.** The same `__all__` query is **non-deterministic run-to-run** — it returned a full result set on one run and "No results." on a later identical invocation in the same session.

Crucially, **single-source queries are 100% reliable and deterministic** for the exact same query strings. So the corpus is present and embedded; only the cross-source (`__all__`) fan-out/merge path is flaky.

This also **contradicts an assumption stated in #1484**, which says *"the same query against `--source __all__` does work (cross-source pipeline runs correctly), so the routing logic is fine."* In this brain, `__all__` is **not** a reliable workaround.

## Environment

- gbrain **0.41.14.0**
- **bun 1.3.14** (so this is NOT #1610 — that truncation bug is `bun < 1.1.27`; this brain is also 100% embedded)
- Engine: postgres (Supabase pooler, port 6543, PgBouncer transaction mode — prepared statements disabled)
- Embedding: `zeroentropyai:zembed-1` (2560d)
- Multi-source brain: `wiki` (33 pages), `outputs` (10), `interview` (13), plus an empty `default` (0 pages). 56 pages total, 478 chunks, **100% embedded** per `gbrain stats`.

## Repro

`__all__` — flaky / non-deterministic:

```
$ gbrain query "acquisitions 2026 pipeline" --source __all__ --limit 2
# run 1: ~13 good results across wiki/interview/outputs
# run 2 (identical command, same session): No results.

$ gbrain query "data warehouse retention pipeline" --source __all__ --limit 3
No results.

$ gbrain search "Comulate" --source __all__
No results.
```

Single-source — 100% reliable for the *same* query strings:

```
$ gbrain query "data warehouse retention pipeline" --source wiki --limit 3
[0.86] themes/ai-tooling-roadmap ...
[0.86] operations/systems-and-data ...

$ gbrain query "data warehouse retention pipeline" --source interview --limit 3
[0.89] union-bay-board-meeting-...-transcript ...
[0.55] union-bay-ai-strategy-meeting-...-transcript ...

$ gbrain search "Comulate" --source wiki
[0.37] log ...
[0.33] themes/ai-tooling-roadmap ...
[...]

PR #3172, opened 2026-07-21
fix(ai,search,cli): embed-shim single body read, zero-hit source hint, vector-arm failure telemetry

### LINKED ISSUE #1484: `gbrain query` silently defaults to source='default' instead of cross-source — invisible miss when content lives in named sources
## Summary

When `gbrain query` / `gbrain search` is run without `--source` and without
`GBRAIN_SOURCE` set, the CLI scopes to the source named literally `default`
(which holds whatever the user captured with bare `gbrain put` — typically a
small set of notes/learnings) and silently ignores every other source in
the brain.

This makes the brain feel "broken at search" for any user whose data lives
in named sources (federated dirs, code repos, m365, imessage, calendar, etc.) —
queries return zero or near-zero hits because the bulk of the corpus is invisible.

## Repro

Brain with multiple sources:

```
$ gbrain sources list
SOURCES
  default               federated       3394 pages  never synced
  browser               federated      69793 pages
  calendar              federated       6410 pages
  contacts              federated        763 pages
  imessage              federated     169396 pages
  m365                  federated     179947 pages
  gstack-code-*         federated       (~5K total)
```

Query for content that exists only in m365:

```
$ gbrain query "Garry Tan" --limit 10
[0.78] transcripts/claude-code/.../...  (in 'default')
[0.64] transcripts/claude-code/.../...  (in 'default')
[0.64] transcripts/claude-code/.../...  (in 'default')
(nothing from m365 — but m365 has thousands of matching pages)
```

Now scope explicitly:

```
$ gbrain query "Garry Tan" --source m365 --limit 10
[0.40] mail/2026/05/20260524t213050z-...  ← hit
[0.40] mail/2026/05/20260522t125202z-...  ← hit
[0.40] mail/2026/05/20260517t102916z-...  ← hit
(5 hits, all relevant)
```

The same query against `--source __all__` does work (cross-source pipeline runs
correctly), so the routing logic is fine — it's just that the silent default
is `default` instead of `__all__`.

## Expected behavior (suggestion)

For interactive queries, `OperationContext.sourceId` should default to `__all__`
(cross-source) when no explicit scope is set. The source named `default` is a
catch-all bucket; treating it as the fallback search scope means users with
named sources get a surprising empty-result experience.

Alternatives if `__all__`-by-default feels too broad:

1. Add a global config setting `query.default_source_scope` (values: `__all__`,
   `default`, or a named source) that users can flip once.
2. Warn / hint in CLI output when a bare query returns 0 hits but the brain has
   sources beyond `default`:
   `  Hint: this brain has 7 sources; you searched only 'default'. Try --source __all__.`
3. Make the help text for `--source-id` explicit that "no flag = source named 'default', not cross-source" (current wording is ambiguous).

## Environment

- gbrain v0.41.0.0
- Postgres engine
- Brain: 440K+ pages across 7+ named federated sources

## Related

- #1167 (closed in v0.37.7.0) — fixed `--source` on the WRITE path; this issue
  is the analogous polish on the READ path's default fallback.
- Companion CLAUDE.md note re-confirmed on v0.37.9.0 that `code-def`/`code-refs`/
  `code-callers`/`code-callees` are still source-flag no-ops on the read path —
  separate sub-issue worth tracking.

## Workaround

Until this lands, users can `export GBRAIN_SOURCE=__all__` in their shell
init. Explicit `--source <name>` still overrides for narrow scopes.

### LINKED ISSUE #1610: ZeroEntropy/Voyage embeddings fail with "Invalid JSON response" for multi-chunk pages on bun < 1.1.27 (Response.clone() truncates large bodies)

time-attack:
Backlog work unit c44 — three verified fixes.

## Fixes #1610 — ZeroEntropy/Voyage embeddings fail with "Invalid JSON response" for multi-chunk pages on bun < 1.1.27

`voyageCompatFetch` and `zeroEntropyCompatFetch` did `await resp.clone().json()`. On bun < 1.1.27, `Response.clone()` truncates large bodies (oven-sh/bun#6348): the parse threw and the catch fell back to the ORIGINAL response — whose wire shape (ZE `{results: ...}`, Voyage base64 embeddings) the AI SDK's openai-compatible Zod schema rejects — so multi-chunk pages failed with "Invalid JSON response".

Both shims now read the body ONCE (`resp.text()` + `JSON.parse`) and rebuild the Response on every JSON return path (modified, unmodified, and parse-failure fallback), deleting the now-stale `Content-Length`/`Content-Encoding` headers instead of copying `resp.headers` verbatim. Layer 1/Layer 2 OOM caps and their ordering are preserved (structural pins updated to the new body-read shape). Shims exported as `__*ForTests` seams; new behavioral test simulates the truncating `clone()`.

## Fixes #1484 — bare `gbrain query` silently defaults to source='default' (invisible miss)

A bare query/search that returns zero results on a brain with >1 registered source now prints a stderr hint naming the source that was actually searched and how to widen scope (`--source-id __all__` / `--source-id <id>`). Fires only when the caller did not scope explicitly (`--source`, `--source-id`, `--all-sources`); best-effort (any lookup failure is silent); stdout stays clean. Changing the default scope to `__all__` remains a separate maintainer policy call, per the triage note.

## Refs #1626 — `--source __all__` cross-source query intermittently returns "No results" (diagnostics)

Unreproduced (single reporter, PgBouncer transaction-mode, v0.41.14.0), but the plausible mechanism is real on master: `hybridSearch`'s text-vector arm swallowed failures with a bare `catch {}`. The arm only runs when the embedding provider probed available, so a throw there (embed timeout, transient pooler error on the `searchVector` fan-out) is a genuine failure — pre-fix it silently collapsed the run to keyword-only/"No results" with zero diagnostics. It now logs the swallowed reason via `warnOncePerProcess` while keeping the fail-open keyword fallback, so the next occurrence is diagnosable. Not claiming closure of the underlying non-determinism.

## Skipped

- #1469 (identical results regardless of input, vector path): needs live reproduction with the reporter's ollama setup on a version ~40 releases behind master; cannot verify a fix from code alone. The #1626 telemetry change covers the code-side actionable part of its fix sketch (surfacing swallowed vector-arm errors).

## Tests

- `test/ai/compat-fetch-body-read.serial.test.ts` (new): truncating-clone simulation for both shims + stale Content-Length pin + non-JSON fallback readability.
- `test/cli-source-scope-hint.test.ts` (new): hint fires/doesn't-fire matrix.
- `test/hybrid-vector-arm-warn.serial.test.ts` (new): embed-transport throw → warning logged + keyword fallback preserved.
- Updated structural pins in `test/voyage-response-cap.test.ts` / `test/ai/zeroentropy-compat-fetch.test.ts` to the new single-read shape.

`bun run typecheck` exit 0; all touched + neighboring test files green (79 + 23 + 23 tests, 0 fail).

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -466,6 +466,11 @@ async function main() {
     const result = JSON.parse(JSON.stringify(rawResult, bigintToStringReplacer));
     const output = formatResult(op.name, result);
     if (output) process.stdout.write(output);
+    // #1484 — invisible-miss hint: a bare query/search that hit zero results
+    // on a multi-source brain tells the user (stderr) which source it
+    // actually searched and how to widen the scope.
+    const hint = await sourceScopeHint(op.name, params, ctx.sourceId, engine, result);
+    if (hint) console.error(hint);
   } catch (e: unknown) {
     // v0.42.20.0 (codex D4): on error, set exitCode + return so the `finally`
     // STILL runs (drains every background-work sink + disconnects). A bare
@@ -835,6 +840,44 @@ async function makeContext(engine: BrainEngine, params: Record<string, unknown>)
     // every transport.
     sourceId: sourceId ?? 'default',
   };
+}
+
+/**
+ * #1484 — a bare `gbrain query`/`search` silently scopes to the resolved
+ * source (usually 'default'); on a multi-source brain a zero-hit run looks
+ * identical to "the brain doesn't know this" even when the answer lives in
+ * another source. Returns a stderr hint when (a) the op is query/search,
+ * (b) it returned zero results, (c) the caller did NOT scope explicitly
+ * (--source / --source-id / --all-sources), and (d) the brain has >1
+ * registered source. Best-effort: any lookup failure returns null.
+ *
+ * Exported for tests (same import-safety contract as formatResult).
+ */
+export async function sourceScopeHint(
+  opName: string,
+  params: Record<string, unknown>,
+  sourceId: string,
+  engine: BrainEngine,
+  result: unknown,
+): Promise<string | null> {
+  if (opName !== 'query' && opName !== 'search') return null;
+  if (!Array.isArray(result) || result.length > 0) return null;
+  // Explicit scoping (flag tier) = user intent; don't second-guess it.
+  if (params.source || params.source_id || params.all_sources) return null;
+  if (sourceId === '__all__') return null;
+  try {
+    const rows = await engine.executeRaw<{ n: number }>(
+      `SELECT count(*)::int AS n FROM sources`,
+    );
+    const n = Number(rows[0]?.n ?? 0);
+    if (n <= 1) return null;
+    return (
+      `Hint: this brain has ${n} sources; you searched only "${sourceId}". ` +
+      `Retry with --source-id __all__ (all sources) or --source-id <id>.`
+    );
+  } catch {
+    return null; // hint is best-effort; never fail the query over it
+  }
 }
 
 // Exported for tests (same import-safety contract as cliAliases/printOpHelp).
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -1000,9 +1000,24 @@ const voyageCompatFetch = (async (input: RequestInfo | URL, init?: RequestInit) 
   // Voyage diverges from OpenAI in two places that break the parser:
   //   - `embedding` is a base64 string (SDK schema expects `number[]`)
   //   - `usage` lacks `prompt_tokens` (SDK schema requires it when usage present)
+  //
+  // #1610: read the body ONCE via text() and JSON.parse it. The pre-fix
+  // `await resp.clone().json()` truncated large bodies on bun < 1.1.27
+  // (oven-sh/bun#6348) — the parse threw, the catch fell back to the raw
+  // response, and multi-chunk pages died with "Invalid JSON response".
+  // Every JSON return path below rebuilds the Response so a stale
+  // Content-Length/Content-Encoding header from the original can't lie
+  // about the rewritten body.
+  const bodyText = await resp.text();
+  const rebuild = (body: string) => {
+    const headers = new Headers(resp.headers);
+    headers.delete('content-length');
+    headers.delete('content-encoding');
+    return new Response(body, { status: resp.status, statusText: resp.statusText, headers });
+  };
   try {
-    const json: any = await resp.clone().json();
-    if (!json || typeof json !== 'object') return resp;
+    const json: any = JSON.parse(bodyText);
+    if (!json || typeof json !== 'object') return rebuild(bodyText);
     let modified = false;
     if (Array.isArray(json.data)) {
       for (const item of json.data) {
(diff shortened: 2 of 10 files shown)


time-attack:
Merging. The clone-truncation bug is still live on master — three shim sites still do `await resp.clone().json()` — and the single-read + rebuilt-Response fix keeps the existing OOM caps and their ordering intact, which was the part I checked hardest. The zero-hit source hint is stderr-only with stdout clean, and turning the vector arm's bare `catch {}` into a warn-once keeps the fail-open behavior while making the next occurrence diagnosable. Good call not claiming closure on #1626.

PR #3173, opened 2026-07-21
fix(remediation): gate sync/extract recs on real extraction lag, refresh at D7 recheck

time-attack:
Takeover of #2363 (do not close — referenced for credit; original author @DarkNightForge, co-authored on the commit).

## What

The `sync.repo` / `extract.all` remediation recommendations gated on `health.stale_pages` — a proxy ("pages whose `updated_at` predates their newest timeline entry") that stopped meaning anything after migration v10 dropped the trigger behind it. They now gate on the honest counter, `engine.countStalePagesForExtraction` — the SAME staleness `gbrain extract --stale` and doctor's `links_extraction_lag` check use — so the recommendation only fires when running the pipeline will actually reduce it.

## Repairs on top of #2363

- **Frozen-gate flaw (the reason for the takeover):** `runRemediation` loads `RecommendationContext` once (run.ts) and the D7 per-step recheck reused it — moving the gate into that frozen `ctx.extractionLagPages` meant a completed sync/extract step could never clear the gate, and the pipeline re-fired on every recheck until `maxJobs`. The D7 recheck now refreshes the gate alongside `getHealth` via a shared `countExtractionLag()` helper in `src/core/remediation/context.ts`. Regression-pinned by `test/remediation-run-d7-refresh.serial.test.ts` (serial per R2: uses `mock.module` for the minion queue; fails against the un-repaired PR).
- **Autopilot wiring:** `src/commands/autopilot.ts` builds its `RecommendationContext` by hand; without wiring it would silently never fire `sync.repo`/`extract.all` again. It now populates `extractionLagPages` from the same helper.

Salvaged intact from #2363: the `RecommendationContext.extractionLagPages` field + docs, the gate/severity/rationale changes in `brain-score-recommendations.ts`, the `BrainHealth.stale_pages` legacy annotation, the `loadRecommendationContext` wiring, and both test files (updated + new).

## Tests

- `bun run typecheck` → exit 0
- `bun test test/remediation-run-d7-refresh.serial.test.ts test/remediation-context-extraction-lag.test.ts test/brain-score-recommendations.test.ts` → 32 pass / 0 fail
- `bun test test/e2e/onboard-full-flow.test.ts test/remediation-step.test.ts` + autopilot wiring tests → all pass

### DIFF
--- a/src/commands/autopilot.ts
+++ b/src/commands/autopilot.ts
@@ -686,6 +686,7 @@ export async function runAutopilot(engine: BrainEngine, args: string[]) {
       try {
         const { MinionQueue } = await import('../core/minions/queue.ts');
         const { computeRecommendations, embeddingProviderConfigured, HOSTED_EMBED_KEY_CONFIG } = await import('../core/brain-score-recommendations.ts');
+        const { countExtractionLag } = await import('../core/remediation/context.ts');
         const queue = new MinionQueue(engine);
         const slotMs = Math.floor(Date.now() / (baseInterval * 1000)) * baseInterval * 1000;
         const slot = new Date(slotMs).toISOString();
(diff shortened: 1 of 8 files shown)


time-attack:
Merging. On master the recommendations still gate on `health.stale_pages`, which has been a dead proxy since v10 dropped its trigger, so sync/extract fire against a counter they cannot move. Catching the frozen-context flaw in the original #2363 — where the D7 recheck reused a stale gate and re-fired to maxJobs — is the part that makes this a takeover worth taking, and wiring autopilot's hand-built context keeps the recs from silently going dark. Credit to the original author is properly carried on the commit.

PR #3174, opened 2026-07-21
fix(facts,serve,jobs): fence into canonical source_path, reap PID-1 orphans, wait out dead supervisor locks, merge fence on write-through

### LINKED ISSUE #2308: jobs supervisor cannot restart after container recreation (gbrain_cycle_locks lease takeover is same-host only)
# `jobs supervisor` cannot restart after container recreation — `gbrain_cycle_locks` lease takeover is same-host only

> **Related:** #2307 — found together while standing up push-webhook sync in a containerized deploy.

**Version:** gbrain 0.42.44.0 · engine: Postgres (external) · deployment: Docker container (Coolify)

## Summary
The supervisor's singleton lease (`src/core/minions/supervisor.ts:528`):
```ts
this.dbLock = await tryAcquireDbLock(this.engine, this.supervisorLockId(), SUPERVISOR_LOCK_TTL_MIN); // = 5
```
is stored in `gbrain_cycle_locks` with `holder_host` / `holder_pid`. Auto-takeover (`classifyHolderLiveness` in `src/core/db-lock.ts`) only steals a holder that is **same-host** *and* has a provably-dead PID:
```ts
if (holderHost !== localHost) return 'cross_host';   // never stolen
```
In containerized deployments, **each container recreation (redeploy/restart) gets a new hostname**, so the previous (now-dead) supervisor's lease is classified `cross_host` and is never taken over. The new supervisor logs:
```
Supervisor already running for queue 'default' on this database
(another supervisor holds the queue lock, regardless of pidfile path). Exiting.
```
and exits with `LOCK_HELD`. Because supervisor start is one-shot at boot, **no worker runs** until the 5-minute TTL lapses — and even then nothing retries.

## Repro
1. In a container, start `gbrain jobs supervisor start --detach`.
2. Recreate the container (new hostname) — e.g. a Coolify/`docker compose` redeploy.
3. On boot, run `gbrain jobs supervisor start --detach` again → exits `LOCK_HELD`.
4. `gbrain jobs supervisor status` → "not running".
5. The lease row for `supervisorLockId('default')` in `gbrain_cycle_locks` is still within `ttl_expires_at` (holder is the old, dead container's host/pid).

## Root
`classifyHolderLiveness` (`db-lock.ts`) returns `cross_host` whenever `holderHost !== os.hostname()`, with a comment that remote `process.kill` is meaningless. That's correct for genuine multi-host setups, but **containers on the same physical host legitimately change hostname across recreations**, so a dead holder is misclassified as a live remote one.

## Impact
Silent worker outage after **every** container redeploy/restart where the hostname isn't pinned. Easy to miss: `serve`/MCP keep answering and the HTTP webhook still returns `202` + enqueues jobs — but the queue never drains, so (e.g.) push-driven sync silently stops.

## Workarounds
- Pin the container hostname so the dead holder is same-host (still subject to the 60 s grace + `DEFAULT_STEAL_GRACE_SECONDS` heartbeat grace).
- Run a bare `gbrain jobs work` loop instead of the supervisor (it takes no singleton lease).
- Clear the stale lease row on boot.

## Suggested fix
- Add a `--steal` / `--force` to `jobs supervisor start`, **or**
- Allow TTL-based takeover regardless of host once the lease is well past `ttl_expires_at` (a row whose TTL has clearly lapsed is dead no matter the host), **and/or**
- Document hostname-pinning as a requirement for containerized supervisor deployments.

### LINKED ISSUE #2443: bug: `gbrain serve` as PID 1 doesn't reap orphaned subprocesses → zombie `git` accumulation → cgroup pids.max exhaustion → ~3–4 day crash-loop
## Summary

On a containerized deployment, `gbrain serve` runs as **PID 1** and never reaps orphaned child processes. Subprocesses it (and its autopilot worker) spawn — chiefly `git` — get reparented to PID 1 when their parent is force-evicted, and because a Bun process running as PID 1 installs no `SIGCHLD` reaper, they accumulate as **zombies**. Over ~3–4 days they hit the cgroup `pids.max` ceiling, `fork()` starts returning `EAGAIN`, and the deployment crash-loops. Restart resets the zombie count, so it recurs on a very regular cadence.

This is a liveness/process-hygiene gap (sibling to the lock-reaping issues in #1972), not data corruption.

## Environment

- gbrain `0.42.52.0` (also present structurally in `0.42.53.0`)
- Runtime: `oven/bun:1.3`, `gbrain serve --http` is the container entrypoint (PID 1)
- Single-source ("legacy single-source") brain, Postgres/Supabase backend
- Host: Railway, cgroup v2, `pids.max = 1000`, memory limit 24 GB

## Symptom

Deployment "crashed" on a near-fixed cadence (~3.6 days apart across four occurrences). **Not** OOM:

```
# /sys/fs/cgroup/memory.events
oom_kill 0
# memory.peak ≈ 656 MB of a 24 GB limit
```

At the crash boundary the logs show:

```
error: cannot fork() for merge-base: Resource temporarily unavailable
error: cannot fork() for fetch: Resource temporarily unavailable
/app/entrypoint.sh: fork: retry: Resource temporarily unavailable   (x4)
```

`fork(): Resource temporarily unavailable` is `EAGAIN` — process/thread-slot exhaustion.

## Root cause

Inside the running container, ~13h after a restart:

```
pids.current = 463 / pids.max = 1000
```

Enumerating `/proc/[0-9]*/stat` (the stripped container `ps` does not surface them):

- **336 processes in state `Z` (defunct), all `comm = git`, all `ppid = 1`.**
- PID 1 is `bun /usr/local/bin/gbrain serve --http ...`.

So `git` children are reparenting to PID 1 (`gbrain serve`) and never being `wait()`ed. A process running as PID 1 must reap orphaned descendants; Bun/gbrain installs no such reaper, so every reparented child becomes a permanent zombie holding a PID slot until the cgroup ceiling is hit.

### How the git children get orphaned

Two contributing spawn paths:

1. **Autopilot worker force-eviction.** We repeatedly saw:
   ```
   [cycle] phase 'propose_takes' ran 33s, exceeding the 30s worker force-evict deadline ... (#1972)
   ```
   When a worker that has spawned a `git` child is force-evicted/aborted, the `git` child is orphaned and reparents to PID 1.

2. **A failing per-cycle `git pull` (dominant source).** The autopilot's `cycle.sync` runs `git pull` in the source clone every interval. In our case it failed every cycle:
   ```
   fatal: could not read Username for 'https://github.com': No such device or address
   ```
   Each failed invocation still spawned a `git`/`git-remote-https` process that exited and zombied under PID 1. At a 60s interval that's a steady ~1 leaked zombie/cycle, which matches the observed accumulation rate and the ~3.6-day time-to-ceiling.

Either way, the **invariant being violated is PID-1 child reaping**, independent of why a given child is orphaned.

## Impact

Hard crash-loop on a regular cadence on any container deployment where `gbrain serve` is PID 1 and any spawned subprocess is orphaned (force-evict, failing git, etc.). It presents as a mysterious "crashes every few days" with no OOM and no obvious error until you read the `fork: EAGAIN` lines.

## Suggested fix

Any one of:

1. **Run an init as PID 1** that reaps zombies — e.g. `ENTRYPOINT ["/usr/bin/tini","--"]` (or Docker/runtime `--init`). This is what we did and it resolved it (PID 1 became tini, zombie count → 0, stable). Worth documenting as the recommended container pattern in the deploy docs at minimum.
2. **Have `gbrain serve` install a `SIGCHLD` reaper** when it detects it is PID 1 (subreaper / `waitpid` loop), so it's robust without an external init.
3. **Ensure spawned `git` subprocesses are always `await`ed/reaped at the call site** even when the owning worker is aborted/force-evicted (addresses the orphaning at the source; complements #1972's abort work).

(1) is the pragmatic immediate guard; (2)/(3) make it correct without relying on the operator's container config.

## Workaround for others hitting this

Add `tini` (or `--init`) as PID 1, and fix any failing per-cycle `git` (e.g. stale/absent credentials) so it isn't leaking a process every interval.

### LINKED ISSUE #2722: fence-write derives the file path from slug, never pages.source_path — fact writes to spaced-filename pages fail on idx_facts_fence_key and resurrect deleted dup-slug stubs
## Summary

`writeFactsToFence` derives its target file as `join(target.localPath, `${target.slug}.md`)` (fence-write.ts:169). `FenceTarget` carries only `{ sourceId, localPath, slug }` — the page's real `pages.source_path` never reaches the write path. For any page whose file basename differs from its slug (the common case for human-titled pages: `10-people/Victor Braca.md` owns slug `10-people/victor-braca`), the fence write targets a file that doesn't exist.

Observed on v0.42.53.0.

## Mechanism (source-confirmed)

1. `existsSync(filePath)` misses the canonical spaced-name file, so the write **stub-creates a new file at the hyphen path** (the stub-creation guard at ~line 183 only refuses slugs with no directory prefix — a fully-qualified `10-people/…` slug passes).
2. `upsertFactRow` computes the next `row_num` from the **empty new fence** → `rowNum=1`.
3. The DB already holds rows for `(source_id, source_markdown_slug, row_num=1…N)` from the canonical page's history → the insert collides with `idx_facts_fence_key` and fails.
4. But the file rename (step 5, fence-write.ts:249) lands **before** the DB stamp (step 6) — so every failed attempt leaves a resurrected orphan file behind, containing only the new facts.

## Consequences

- **Fact writes to these pages fail permanently** — every `extract_facts` targeting such an entity errors on the fence key, forever.
- **Dup-slug dedupes silently undo themselves**: deleting a hyphen-slug stub file leaves the DB page alive; the next fact write re-creates the stub at the hyphen path. We watched 3 of 11 deduped stubs resurrect within days, each carrying a divergent fence.
- The canonical page's fence and the DB drift apart with no error surfaced to the user (the failure is only visible in write-failure logs).

## Repro

1. Create `10-people/Jane Doe.md` (spaced filename), sync — page gets slug `10-people/jane-doe`.
2. Run `extract_facts` on any transcript mentioning Jane so a fact resolves to her.
3. Result: a new `10-people/jane-doe.md` stub appears; if the canonical page already had fence rows, the insert dies on `idx_facts_fence_key`, and the stub remains.

## Ask

Resolve the write path from `pages.source_path` when the page row exists (thread it through `FenceTarget`, or look it up inside `writeFactsToFence`); fall back to the slug-derived path only for genuinely new pages. Secondarily, consider deferring the file rename until after the DB insert succeeds (or unlinking the stub on insert failure) so a failed write can't strand an orphan file.

Related: the legacy `localPath === null` fallback can also mint DB-only rows with `row_num = NULL`, which the dream-phase fence reconcile then can't match — same family of fence/DB identity drift.

### LINKED ISSUE #2839: sync/import and fact-writers use disjoint lock domains — page-lock cannot deliver the atomicity its docstring promises
**Version:** verified on `master` @ `5008b287` (**0.42.59.0**).
**Not container-specific:** reproduces on a single host, single process tree, no Docker.

## Summary

`src/core/page-lock.ts` states its purpose as:

> per-page file lock for atomic markdown read-modify-write … so two parallel `gbrain takes add` calls + a `takes seed --refresh` running in autopilot can't race on the same `<slug>.md` file.

But the sync/import path — the code most likely to rewrite `<slug>.md` — never takes that lock. It uses a **different, unrelated** lock domain, so the two writers do not exclude each other.

| Path | lock taken |
|---|---|
| `src/core/facts/fence-write.ts:172` | `withPageLock(slug)` — **FS lock**, per page |
| `src/core/facts/forget.ts:140` | `withPageLock(slug)` — FS lock |
| `src/commands/takes.ts` (×4) | `withPageLock(slug)` — FS lock |
| **`src/commands/sync.ts` (import)** | **`syncLockId` from `src/core/db-lock.ts` — DB lock, per source** |

`withPageLock` appears **0 times** in `sync.ts`. The DB lock serializes syncs against other syncs; it says nothing about per-page FS writes. The two domains are disjoint: **neither writer excludes the other on the same file.**

Consequence: an import rewriting `<slug>.md` can interleave with a `fence-write` read-modify-write on that same file. The fact deposit is silently lost, or the import clobbers a just-written fact. No error is surfaced. The page lock only serializes fact writers *among themselves* — it offers no protection against the importer.

## Repro sketch (single host, no containers)

1. Register a source whose repo contains `<slug>.md`.
2. Start `gbrain autopilot` — it pulls and syncs by default (`src/commands/autopilot.ts:962-965`: *"Autopilot daemon path: pulls by default"*), so the trigger ships in the product.
3. Concurrently call `extract_facts` (MCP) targeting `<slug>`.
4. Race window = the import's write of `<slug>.md` vs. the fenced read-modify-write.

The combination is not contrived: `autopilot` is the recommended daemon, and `extract_facts` is exposed over MCP — a caller has no way to know a sync is mid-flight.

## Suggested directions

1. **Have the import take `withPageLock(slug)`** for each page it rewrites. Cost is lock churn on large imports — could be limited to pages that already exist on disk (a first import has no concurrent fenced reader by definition).
2. **Unify the domains** — express the page lock in `gbrain_cycle_locks` so sync's lock and fact writes are comparable. Note this inherits the same-host limitation of #2308 until that is fixed.
3. **Or make the exclusion explicit**: have `sync` refuse to start while a fence write is in flight (or vice versa) and document that facts must not be written during a sync.

Direction 1 looks like the smallest change that matches the stated intent.

## Relation to existing issues

- **#1950** — `sync --source default wedged ~29min holding gbrain-sync lock`: confirms sync's lock is the DB/sync-scoped one, not the page lock.
- **#1841** — concurrent `capture` during `serve` corrupts writes: same family (concurrent writers to one brain), but PGLite/engine-level rather than the FS page lock.
- **#2722** — `fence-write` derives the file path from slug, never `pages.source_path`: adjacent, same writer.

> See also: #2840 — where the page lock *is* taken, its PID-based staleness check is unsound across containers. The two are independent: fixing only this one extends a lock that is still unsound in containerized deploys.

time-attack:
Backlog fix unit c43-f1 — four adversarially-verified items.

## Fixes #2722 — fence-write derives the file path from slug, never `pages.source_path`

`writeFactsToFence` computed the target file purely from the slug, so a page imported from a spaced filename (`10 People/Jane Doe.md`, slug `10-people/jane-doe`) got a parallel hyphen-path stub created next to the canonical page, forking it. Now:

- New `resolveFenceFilePath` consults `pages.source_path` when the page row exists (markdown files only, `isWriteTargetContained` guard against hostile rows), falling back to the slug-derived path for genuinely new pages.
- `forgetFactInFence` shares the same resolver — its bare `join(localPath, slug + '.md')` had the same bug and silently degraded spaced-filename forgets to the DB-only path (which doesn't survive rebuild).
- Secondary: a freshly stub-created page file is unlinked when `engine.insertFacts` throws, so a transient DB failure can't leave a phantom entity page. Pre-existing pages keep their fence (system of record).

## Fixes #2443 — `gbrain serve` as PID 1 doesn't reap orphaned subprocesses

The existing no-op SIGCHLD handler only makes the runtime reap its OWN children; orphaned grandchildren re-parented onto PID 1 were never waited on, accumulating zombie `git` processes until the cgroup pids limit starved fork.

- `installPid1OrphanReaper` (wired at CLI boot; no-op unless PID 1 on Linux): a periodic /proc sweep collects Z-state children and reaps only pids seen in **two consecutive sweeps**, via `bun:ffi` libc `waitpid(pid, WNOHANG)`. The two-sweep confirmation avoids racing the runtime's own per-child reaping (a blanket `waitpid(-1)` on SIGCHLD can steal exit statuses from `Bun.spawn`/`child_process` and hang callers).
- `docs/mcp/DEPLOY.md` gains a container section recommending `docker run --init` / tini as the primary entrypoint; the in-process reaper is the backstop for images without one. (llms bundles regenerated.)

## Fixes #2308 — jobs supervisor cannot restart after container recreation

Supervisor start was one-shot on the queue-scoped DB lock: a recreated container (new hostname) exited `LOCK_HELD` inside the old holder's TTL window, and nothing retried — the queue stayed dead until an operator intervened. Cross-host takeover is TTL-only by design, so the fix is to wait it out:

- `acquireSupervisorDbLockWithWait` polls the acquire while the holder is provably dead, bounded by TTL + steal grace + margin. Liveness is decided by observation, not a single snapshot: a same-host live PID fails fast immediately (interactive duplicate-start keeps its instant exit 2), and a heartbeat that ADVANCES between polls proves a live cross-host holder (fail fast). A heartbeat that never advances is a dead holder — polling continues until the upsert's host-agnostic TTL takeover fires.
- Emits `db_lock_wait` supervisor events while waiting so operators see progress.

## Fixes #2839 — write-through clobbers on-disk facts fence; disjoint lock domains

`writePageThrough` (put_page / capture / brainstorm disk write-through) atomically rewrote `<slug>.md` rendered from the DB row with no page lock. Fence writers deposit facts into the FILE first and the DB `pages.body` only catches up on the next sync — so a put_page could silently drop just-deposited fence rows from disk.

- The read-merge-rename now runs under the same `withPageLock(slug)` the fence writers hold (lock timeout surfaces via the existing `error` field; `writePageThrough` still never throws).
- New `mergeDiskFactsFence` merges the on-disk fence into the rendered body before writing: union by `row_num`, the disk row wins collisions (the fence file is system of record for facts).
- Note: the issue named sync/import as the clobbering writer; verification showed sync only READS page files — the real unlocked writer is the write-through, which is what this fixes.

## Tests

- `test/fence-write.test.ts`: source_path routing (spaced filename, no hyphen fork), slug fallback, hostile-source_path containment, stub cleanup on DB failure, pre-existing page kept on DB failure.
- `test/zombie-reap.test.ts`: /proc stat parsing (comm with spaces/parens), two-sweep confirmation semantics, PID-1/Linux gating, interval wiring.
- `test/supervisor-db-lock-wait.test.ts` (new): dead cross-host holder waited out, same-host live PID fails fast, advancing heartbeat fails fast, maxWait bound, vanished-row polling.
- `test/write-through.test.ts`: on-disk fence rows survive a write-through; `mergeDiskFactsFence` union/collision/append/no-op cases.

Gates: `bun run typecheck` clean; 181 tests across the 13 touched/related files pass; jsonb guards clean; `bun run build:llms` regenerated and freshness test passes.

### DIFF
--- a/docs/mcp/DEPLOY.md
+++ b/docs/mcp/DEPLOY.md
@@ -258,6 +258,24 @@ the user owns the machine.
 See [ALTERNATIVES.md](ALTERNATIVES.md) for a comparison of ngrok, Tailscale
 Funnel, and cloud hosts (Fly.io, Railway).
 
+### Running in a container (PID 1)
+
+When `gbrain serve` (or `gbrain jobs supervisor start`) is the container
+entrypoint, it runs as PID 1 and inherits init duties: every orphaned
+grandchild process the kernel re-parents onto it must be reaped, or zombies
+accumulate until the container's pid limit is exhausted. Prefer a real init
+as the entrypoint:
+
+```bash
+docker run --init ...            # Docker's bundled tini
+# or in a Dockerfile:
+ENTRYPOINT ["tini", "--", "gbrain", "serve", "--http"]
+```
+
+gbrain also ships an in-process backstop: when it detects it is PID 1 on
+Linux it periodically reaps orphaned zombie children. An explicit init like
+tini is still the recommended setup.
+
 ## Troubleshooting
 
 **"missing_auth" error**
--- a/llms-full.txt
+++ b/llms-full.txt
@@ -3902,6 +3902,24 @@ the user owns the machine.
 See [ALTERNATIVES.md](ALTERNATIVES.md) for a comparison of ngrok, Tailscale
 Funnel, and cloud hosts (Fly.io, Railway).
 
+### Running in a container (PID 1)
+
+When `gbrain serve` (or `gbrain jobs supervisor start`) is the container
+entrypoint, it runs as PID 1 and inherits init duties: every orphaned
+grandchild process the kernel re-parents onto it must be reaped, or zombies
+accumulate until the container's pid limit is exhausted. Prefer a real init
+as the entrypoint:
+
+```bash
+docker run --init ...            # Docker's bundled tini
+# or in a Dockerfile:
+ENTRYPOINT ["tini", "--", "gbrain", "serve", "--http"]
+```
+
+gbrain also ships an in-process backstop: when it detects it is PID 1 on
+Linux it periodically reaps orphaned zombie children. An explicit init like
+tini is still the recommended setup.
+
 ## Troubleshooting
 
 **"missing_auth" error**
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -1,7 +1,13 @@
 #!/usr/bin/env bun
 
-import { installSigchldHandler } from './core/zombie-reap.ts';
+import { installSigchldHandler, installPid1OrphanReaper } from './core/zombie-reap.ts';
 installSigchldHandler();
+// #2443: when gbrain IS the container init (PID 1 — `gbrain serve` /
+// `gbrain jobs supervisor start` as the entrypoint), re-parented orphan
+// grandchildren (e.g. git spawned by a dead worker) become zombies nobody
+// waits on and eventually exhaust the cgroup pid budget. No-op unless
+// PID 1 on Linux.
+installPid1OrphanReaper();
 // v0.41.6.0 D5: cleanup registry + signal handlers for SIGTERM/SIGHUP/SIGPIPE/
 // uncaughtException. NOT SIGINT (the existing AbortController path at :254
 // owns SIGINT). Installed at module load so locks acquired during boot
--- a/src/core/facts/fence-write.ts
+++ b/src/core/facts/fence-write.ts
@@ -33,11 +33,12 @@
  * sees the constraint.
  */
 
-import { existsSync, mkdirSync, readFileSync, writeFileSync, renameSync, appendFileSync } from 'node:fs';
-import { dirname } from 'node:path';
+import { existsSync, mkdirSync, readFileSync, writeFileSync, renameSync, appendFileSync, unlinkSync } from 'node:fs';
+import { dirname, join } from 'node:path';
 
 import type { BrainEngine, NewFact, FactVisibility } from '../engine.ts';
 import { resolvePageFilePath } from '../markdown.ts';
+import { isWriteTargetContained } from '../path-confine.ts';
 import { withPageLock } from '../page-lock.ts';
 import { gbrainPath } from '../config.ts';
 import { upsertFactRow, parseFactsFence } from '../facts-fence.ts';
@@ -91,6 +92,38 @@ export interface FenceWriteResult {
 }
 
 const FAILURE_LOG_PATH = (): string => gbrainPath('facts.write_failures.jsonl');
+
+/**
+ * #2722: compute the on-disk file to fence into. When the page row exists
+ * with an import-time `pages.source_path`, that path (relative to the
+ * source's local_path — the same root `scanOneSource` walked) is the
+ * canonical file, and it can differ from the slug-derived name (slugs are
+ * normalized: `10 People/Jane Doe.md` → slug `10-people/jane-doe`). Only
+ * honored for markdown files that stay inside the source root (the DB row
+ * is data, not trusted path input). Everything else falls back to the
+ * slug-derived path, which is also the stub-create location for genuinely
+ * new pages.
+ *
+ * Shared with `forgetFactInFence` so the forget path edits the SAME file
+ * the fence write deposited into.
+ */
+export async function resolveFenceFilePath(engine: BrainEngine, target: FenceTarget): Promise<string> {
+  const root = target.localPath as string;
+  try {
+    const rows = await engine.executeRaw<{ source_path: string | null }>(
+      `SELECT source_path FROM pages WHERE source_id = $1 AND slug = $2 LIMIT 1`,
+      [target.sourceId, target.slug],
+    );
+    const sourcePath = rows[0]?.source_path ?? null;
+    if (sourcePath && sourcePath.toLowerCase().endsWith('.md')) {
+      const candidate = join(root, sourcePath);
+      if (isWriteTargetContained(candidate, root)) return candidate;
+    }
+  } catch {
+    // Best-effort lookup — fall back to the slug-derived path.
+  }
+  return resolvePageFilePath(root, target.slug, target.sourceId);
+}
 
 function recordWriteFailure(slug: string, sourceId: string, warnings: string[], filePath: string): void {
   // Best-effort JSONL append — never throws back into the caller. The
@@ -172,7 +205,14 @@ export async function writeFactsToFence(
   // the put_page write-through and dream-cycle reverse-render compute. The
   // bare join wrote main-source fences to the repo ROOT (the default source's
   // tree), polluting ~/brain with stray root-level fence files.
-  const filePath = resolvePageFilePath(target.localPath, target.slug, target.sourceId);
+  //
+  // #2722: when the page row already exists with an import-time
+  // `pages.source_path`, fence into THAT file. Slugs are normalized (spaces →
+  // hyphens, lowercase), so a page imported from `10 People/Jane Doe.md` has
+  // slug `10-people/jane-doe` — the slug-derived path would stub-create a
+  // parallel `10-people/jane-doe.md` next to the canonical page and fork it.
+  // The slug-derived path stays the fallback for genuinely new pages.
+  const filePath = await resolveFenceFilePath(engine, target);
   const tmpPath = `${filePath}.tmp`;
 
   return withPageLock(
@@ -180,6 +220,7 @@ export async function writeFactsToFence(
     async () => {
       // 1. Read existing body or stub-create.
       let body: string;
+      let stubCreated = false;
       if (existsSync(filePath)) {
         body = readFileSync(filePath, 'utf-8');
       } else {
@@ -215,6 +256,7 @@ export async function writeFactsToFence(
         // Stub-create the parent directory if it doesn't exist.
         mkdirSync(dirname(filePath), { recursive: true });
         body = stubEntityPage(target.slug);
+        stubCreated = true;
       }
 
       // 2. Upsert each fact onto the fence in input order. row_num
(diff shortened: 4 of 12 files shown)


time-attack:
Merging, mainly on the fence-write fix: master still derives the fence path purely from the slug, so a page imported from a spaced filename forks into a parallel hyphen-path stub, and `forgetFactInFence` had the same bug — that is data integrity, and catching the shared root cause in both callers is the right shape. The supervisor lock-wait is also a genuine availability fix; a recreated container currently exits LOCK_HELD and the queue stays dead until someone intervenes. My one reservation is the bun:ffi PID-1 reaper, since `docker run --init` already solves that and it adds FFI surface — it is gated tightly enough to keep, but I would rather see that half live or die on its own next time than ride along in a four-item bundle.

PR #3175, opened 2026-07-21
fix(search): preserve email citation metadata across result paths

### LINKED ISSUE #2962: Search paths drop trusted email citation provenance across hydration and cache

time-attack:
Takeover of #2875 (fix_needed/P1, backlog part4-13). Also carries the base projection from #2873 (same author, subsumed by this diff) — referencing both, closing nothing.

## What this is

PR #2875 had real value (email citation provenance preserved across every search result path) but was self-declared not merge-ready: it bundled VERSION/package.json/CHANGELOG bumps, overlapped #2873, and its head had an unreviewed hot-path defect plus a PostgreSQL E2E run that was only green on a superseded head. This takeover rebases the full delta onto current master, drops all release bookkeeping, fixes the hot-path defect, and re-verifies everything on this exact head.

## Salvaged (author: @amtagrwl, credited via Co-authored-by)

- Project trusted `message_id` / `thread_id` / Message-ID-gated `source_subject` through keyword, chunk-keyword, CJK, and vector result paths in BOTH engines (postgres + pglite, in lockstep)
- Preserve the citation DTO through alias injection, relational recall/fanout, two-pass hydration, vector fusion/reranking, and semantic-cache hits
- Trust boundary: raw `source_subject` is never trusted; only allowlisted `subject` may supply it, and only when a nonblank Message-ID proves the page is an email; malformed/non-object frontmatter fails closed (no double-decode of JSON-string frontmatter)
- Source visibility / quarantine / deletion rechecked across indirect retrieval paths (alias hop, relational hydrate, two-pass expansion, graph walk, and a lookup-time visibility gate on cached rows)
- Typed JSON cache scope keys (`["scalar",id]` / `["set",...]` / `["all"]`) — injective encoding, forged legacy `__set__:` keys can no longer collide
- Store-side race gate: writeback is skipped when the page-generation clock advanced during the producing search, so a mid-search page write can't be certified fresh
- `KNOBS_HASH_VERSION` 12 → 13 so pre-projection cached DTOs miss instead of replaying the old result shape (one-time cold-miss, refills within TTL)

## Fixed on top of the original head

- **Cache-scope fail-open (hot-path crash class):** `cacheScopeKey`'s forged-id rejection (`assertValidSourceId` throw) was evaluated inline at the cache lookup/store call sites inside `hybridSearchCached`, outside any catch — an invalid scope id rejected the entire search instead of skipping the cache, violating the module's own "cache must never break the search hot path" contract. The key is now computed once up front, fail-open: invalid scope → semantic cache skipped, search unaffected, nothing written under a forged scope. Pinned by `test/search/hybrid-cache-scope-failopen.serial.test.ts` (verified failing on the original head, passing here).
- Full `HybridSearchOpts`/`SearchOpts` field audit on this head: every result-shaping knob is either folded into the knobs hash, in the `isSemanticCacheRequestSafe` unsafe list, covered by a skip condition, or result-neutral — no cache cross-serve gaps found.

## Not included (deliberately)

- VERSION / package.json / CHANGELOG bumps — release bookkeeping happens at ship time per the /ship-owns-versioning flow.

## Verification (this exact head)

- `bun run typecheck` — clean
- 379 touched unit tests across 15 files — green
- 3 serial files (one bun process each, matching CI) — green
- PGLite e2e: cache-gate, source-isolation — green
- **Real PostgreSQL** (pgvector/pgvector:pg16): engine-parity 26 pass, source-routing 9 pass — green
- Repo guards: jsonb-pattern, jsonb-params AST scan, key-files-current-state, test-isolation, progress-to-stdout — all clean

Takeover of #2875. Takeover of #2873 (base projection). Closes #2962.


time-attack:
Merging despite the size — only about 460 of the added lines are in `src/`, the rest is e2e parity, source-isolation, and cache coverage, so it reviews as a correctness fix rather than a feature drop. The catch that earns the takeover is the cache-scope fail-open: computing `cacheScopeKey` inline at the lookup/store sites inside `hybridSearchCached` put a throwing forged-id check on the search hot path, which breaks the module's own "cache never breaks search" contract. The trust boundary on `source_subject` (Message-ID gated, malformed frontmatter fails closed) and the KNOBS_HASH_VERSION bump so pre-projection DTOs miss rather than replay are both the right calls. Thanks for dropping the VERSION/CHANGELOG bookkeeping from the original and crediting the author.

PR #3176, opened 2026-07-21
fix(admin): source scoping + redirect URIs in admin OAuth registration

### LINKED ISSUE #1036: admin: dashboard at :3131/admin does not expose redirect_uris field for OAuth client registration

### LINKED ISSUE #1490: admin: "Register Agent" form does not expose source_id / federated_read — multi-source clients silently broken

time-attack:
Takeover of #1558 (credit: @flamerged — source changes salvaged intact, rebased onto current master with admin/dist + src/admin-embedded.ts regenerated so the asset hashes/manifest match).

Fixes #1490
Fixes #1036

## What

- **#1490 — source scoping in admin OAuth registration.** `/admin/api/register-client` accepts `source_id` + `federated_read` (snake and camel case), validated fail-closed against the registered source list via the new `normalizeAdminClientSourceScope` (`src/core/admin-source-scope.ts`) before reaching `registerClientManual`. Unknown or malformed source ids are a structured 400 — never a silently-hardcoded `'default'`. A new `requireAdmin`-gated `GET /admin/api/sources` feeds the Register Agent form, which now has a Write Source select + Read Sources checkboxes; the agents table and drawer show each client's scope.
- **#1036 — redirect URIs in the Register Agent form.** New Redirect URIs textarea (one per line). When present, the POST carries `redirectUris` plus the `authorization_code` + `refresh_token` grants, mirroring the CLI's `--redirect-uri` grant-inference convention (`src/commands/auth.ts`). Backend plumbing already existed; this half was UI-only.

## Why a takeover

#1558 was correct but DIRTY against master: it deleted `admin/dist/assets/index-DqP-zmqH.js` while master now ships a newer asset hash. This branch rebuilds the admin SPA on current master (`bun run build:admin`), so `scripts/check-admin-embedded.sh` passes.

## Tests

- `test/admin-source-scope.test.ts` — normalization matrix (defaults, comma-separated, empty fallback, malformed + unregistered ids rejected).
- `test/oauth.test.ts` — `registerClientManual` persists `source_id`/`federated_read` into `AuthInfo` end-to-end through token exchange.
- `test/fix-wave-structural.test.ts` — pins for the endpoint body fields, `normalizeAdminClientSourceScope` call, `/admin/api/sources` gate, and the form's `source_id`/`federated_read`/`redirectUris` POST shape.

`bun run typecheck` exit 0; the three test files above: 128 pass / 0 fail; `scripts/check-admin-embedded.sh` + `scripts/check-admin-scope-drift.sh` both green. (Admin-workspace `tsc` has 3 pre-existing errors on master at `Agents.tsx` `SETUP_SNIPPETS` — untouched here.)


time-attack:
Merging. Master's `/admin/api/register-client` has no source-scope handling at all, so agents registered through the admin UI are silently pinned to `default` — source scoping is a must-never-violate invariant here, and fail-closed validation against the registered source list with a structured 400 is exactly the right treatment. The redirect-URIs field is UI-only wiring for plumbing that already existed, so it is fine riding along. Rebuilding admin/dist on current master is what makes this land cleanly where the original #1558 could not; credit to the original author is carried.

PR #3177, opened 2026-07-21
fix(search,embed): normalize / in FTS queries; CPU-safe defaults for local embedding endpoints

### LINKED ISSUE #2552: Cloud-tuned embedding defaults (concurrency 20, no_batch_cap, char-based batch estimate) silently wedge CPU-only Ollama boxes
Three compounding defaults assume fast cloud APIs and degrade **silently** on a CPU-only Ollama box (4-core ARM, bge-m3):

(a) `GBRAIN_EMBED_CONCURRENCY` defaults to 20 (`embed.ts:264`) — 20 parallel pages against a single Ollama slot multiply latency ×20 and blow past the Bun fetch ~300s timeout;
(b) the `ollama` recipe sets `no_batch_cap: true` — a whole page goes out in one request (cf. #2383 for the same question on the google recipe);
(c) batch size is estimated in chars at ~4 chars/token, but code-dense pages run ~2 chars/token — real batches land at ~2× the cap.

Net effect: small pages embed, big pages never do, and no error surfaces anywhere — the box looks healthy while the backfill starves (it cost us ~14h of hunting on a 54-book corpus @ bb2e88c, v0.42.52.0).

Suggestion: CPU-safe defaults when the embedding recipe is ollama (concurrency 1, batch cap ~8000 chars), or at least a doctor warning when concurrency > 1 with a local endpoint. Related: #1946.

time-attack:
Two verified backlog fixes (work unit c48):

## Takeover of #2380 — fix(search): normalize `/` in FTS queries

Postgres' default text-search parser classifies `foo/bar` as a single `file`-alias token mapped to the `simple` dictionary, so `websearch_to_tsquery('english', 'foo/bar')` produces one un-stemmed lexeme that never matches indexed text. Any slash-containing query (e.g. a pasted `Author1/Author2 - Topic` title) bypassed FTS AND semantics: the primary pass returned zero hits and only the OR fallback surfaced (over-broad) results.

- `normalizeKeywordQuery()` in `src/core/search/keyword.ts` replaces `/` with whitespace before parse; all three FTS arms (`searchKeyword`, `searchKeywordChunks`, `searchTitles`) route through it in BOTH engines.
- Beyond the original PR: the title arm was missed there (master's `searchTitles` still bound the raw query), and the stray repo-root `node_modules` symlink in the original diff is dropped.
- Regression tests are AND-vs-OR distinguishing (master's OR fallback would mask a plain "returns results" assertion): both fail pre-fix, pass post-fix.

Credit: original diagnosis and fix by @rwbaker (co-authored on the commit). Referencing only — not closing #2380 here.

## Fixes #2552 — fix(embed): cloud-tuned defaults silently wedge CPU-only Ollama boxes

Three compounding cloud-tuned defaults starved local-endpoint backfills with no surfaced error:

- The `ollama` recipe now declares a conservative static batch cap (`max_batch_tokens: 4096`, `chars_per_token: 2` — ~8K chars/request, matching the code-dense ~2 chars/token reality) instead of `no_batch_cap: true`. Ollama never returns a recognizable token-limit error, so the recursive-halving safety net can't fire; the pre-split cap is the only guard.
- Bulk embed auto-caps worker fan-out at `LOCAL_EMBED_CONCURRENCY_CAP` (2) when the embedding endpoint is a local inference server (`ollama` / `llama-server` recipes, or an explicit localhost base URL) and `GBRAIN_EMBED_CONCURRENCY` is unset. An explicit env value always wins; the cap logs one stderr line when it engages.
- `gbrain doctor` grows an `embed_concurrency` check (OPS category) that warns when an explicit `GBRAIN_EMBED_CONCURRENCY > 2` fans out against a local endpoint.

## Tests

- `test/pglite-engine.test.ts`: slash-query AND-semantics regression tests for `searchKeyword` + `searchTitles` (verified failing without the engine fix).
- `test/embed-local-endpoint.serial.test.ts` (new): `isLocalEmbeddingEndpoint`, `resolveEmbedConcurrency` (auto-cap / env-wins / pace-only-lowers), `computeEmbedConcurrencyCheck`.
- `test/ai/no-batch-cap-suppression.serial.test.ts`: updated — ollama now asserts the static cap; litellm/llama-server keep `no_batch_cap`.
- `bun run typecheck` clean; touched suites (engine, search, doctor, embed, providers, gateway batch) all pass.

### DIFF
--- a/src/commands/doctor.ts
+++ b/src/commands/doctor.ts
@@ -820,6 +820,10 @@ export async function doctorReportRemote(engine: BrainEngine): Promise<DoctorRep
   // v0.42.x (#1794, 4A): pool-budget nudge when GBRAIN_MAX_CONNECTIONS is set.
   checks.push(await checkPoolBudget(engine));
 
+  // #2552: warn when an explicit embed-concurrency override fans out against
+  // a local single-slot embedding endpoint (silent backfill starvation).
+  checks.push(await checkEmbedConcurrency());
+
   // v0.42.7 (#1696): link-extraction lag. Strictly SQL (single indexed COUNT),
   // safe on the thin-client/remote path — remote operators on checkout-less
   // Postgres brains are exactly who can't otherwise see the extraction backlog.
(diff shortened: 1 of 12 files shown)


time-attack:
Review correction pushed (5882d526): the normalize-only approach regressed literal slash-form matching. Postgres emits the joined file-alias lexeme on the index side too, so a query rewritten to split words could no longer find documents containing the literal `foo/bar` (file paths, URLs, pasted titles) — a class the pre-PR code matched exactly, and one the OR fallback can't rescue. Replaced `normalizeKeywordQuery` with `buildWebsearchQueryExpr` in `sql-ranking.ts` (shared by both engines): slash queries now match `(websearch_to_tsquery(translate($1,'/',' ')) || websearch_to_tsquery($1))` — split prose AND literal forms; slash-free queries keep byte-identical SQL. Added a literal-slash regression test (verified failing under normalize-only). Existing AND-vs-OR tests still pass; typecheck clean. The embed/#2552 half reviewed clean as-is.

time-attack:
Merging. Both halves check out on master: there is no keyword-query normalization on the FTS path, so a pasted slash-containing title collapses the primary pass to zero hits and only the over-broad OR fallback survives, and the `ollama` recipe still declares `no_batch_cap: true` with no error Ollama can surface for the halving net to catch. Catching the `searchTitles` arm the original #2380 missed, and making the regression tests AND-vs-OR distinguishing so the OR fallback cannot mask them, is the detail that makes this trustworthy. Keeping explicit `GBRAIN_EMBED_CONCURRENCY` authoritative over the local cap is the right default.

PR #3178, opened 2026-07-21
fix(cycle): stamp path-derived dream sources; close engine on autopilot shutdown

### LINKED ISSUE #1869: gbrain dream --dir <path> never stamps cycle_freshness (only --source does) → brain shows perpetually stale

### LINKED ISSUE #1872: systemctl stop of autopilot (SIGTERM) corrupts the PGLite engine — no graceful shutdown
## Summary
Stopping the autopilot service with `systemctl --user stop gbrain-autopilot` (SIGTERM) shuts PGLite down **uncleanly and corrupts the engine**: every subsequent `gbrain` invocation then fails with `PGLite failed to initialize its WASM runtime / Aborted()`. Recovery required a full `reinit-pglite` + resync.

## Environment
- Linux/WSL, PGLite engine, gbrain 0.42.x, autopilot run as a systemd **user** service.

## Repro
1. Run `gbrain autopilot` on a PGLite brain (e.g. via the systemd service) so it holds the DB and cycles.
2. `systemctl --user stop gbrain-autopilot` (SIGTERM) while it's alive.
3. `gbrain doctor` / `gbrain search` → `PGLite failed to initialize its WASM runtime. Aborted()`, persistently.
4. Verified NOT a stale lock: no `gbrain`/`bun` process alive (`pgrep`, `lsof +D <datadir>` empty), removing `postmaster.pid` doesn't help. The data dir itself is corrupted.
5. Recovery: `gbrain reinit-pglite --embedding-model <p:m> --embedding-dimensions <N> --yes` (backs up to `*.bak`) then `gbrain sync --repo <brain>`.

## Root cause (suspected)
The autopilot process doesn't trap SIGTERM to close the PGLite WASM Postgres cleanly, so a normal `systemctl stop` kills it mid-write → corrupted data directory.

## Impact
A routine `systemctl stop` corrupts the brain engine. Content survives (markdown is source of truth) but the index must be rebuilt. Combined with PGLite being single-connection (autopilot also blocks all queries while running), this makes **autopilot effectively unusable on PGLite**.

## Suggested fix
- Trap SIGTERM/SIGINT in autopilot and perform a clean PGLite shutdown (close the connection / checkpoint) before exit.
- Set a sane `TimeoutStopSec` / `KillSignal` in the generated systemd unit and document graceful-stop.
- Consider refusing/​warning when `autopilot` is run against a PGLite engine (recommend Postgres for long-running autopilot).

time-attack:
Two verified backlog fixes (work unit c45).

## Fixes #1869 — `gbrain dream --dir <path>` never stamps cycle freshness

`runCycle`'s stamp gate reads `opts.sourceId`, and dream only set it from `--source` — so path-scoped runs never wrote `last_source_cycle_at` / `last_full_cycle_at` and doctor's `cycle_freshness` stayed perpetually stale.

Fixed at the command level: `dream.ts` derives the source id from the resolved brain dir via `resolveSourceForDir` (now exported from `cycle.ts`) and passes it as `opts.sourceId`. `runCycle`'s stamp/lock semantics are untouched, so the `autopilot-global-maintenance` handler (GLOBAL_PHASES with a brainDir and no sourceId) cannot falsely stamp per-source freshness — the #2194-class poisoning that a runCycle-wide gate change would reintroduce. A derived match on an archived source is skipped, mirroring the explicit `--source` archived guard.

**Takeover of #2549** (not closed here): salvages its intent and help-text improvement while moving the fix out of `runCycle`. Credit: @javieraldape (co-authored on the commit).

Tests: `test/dream-dir-source-stamp.test.ts` (stamp lands with `--dir`; archived source not stamped — fails without the fix) + updated the superseded back-compat case in `test/dream.test.ts` to pin cross-source isolation instead.

## Fixes #1872 — `systemctl stop` of autopilot corrupts the PGLite engine

Autopilot's shutdown drained the worker child and hard-exited without ever closing the engine; on PGLite the cycle runs INLINE in the autopilot process, so SIGTERM mid-write killed WASM Postgres with the WAL dirty. Additionally, process-cleanup's SIGTERM handler (installed at `cli.ts` module load) exits within its 3s cleanup deadline, so the fix must ride both paths:

- `shutdown()` (SIGINT + internal stops) now aborts the in-flight inline cycle via an `AbortController` threaded into `runCycle`, drains it briefly, and awaits `engine.disconnect()` before `process.exit(0)`.
- The same `closeEngine` is registered in the process-cleanup registry (`autopilot-engine-close`) so the SIGTERM path closes the engine too.

`disconnect()` is idempotent (snapshots + nulls the handle), so both paths firing is safe.

Tests: `test/autopilot-shutdown-engine-close.test.ts` — static-shape wiring guards, following the established `autopilot-supervisor-wiring.test.ts` precedent (the shutdown path is deep inside the daemon loop and ends in `process.exit`).

## Verification

- `bun run typecheck` → exit 0
- New/touched test files (`dream-dir-source-stamp`, `autopilot-shutdown-engine-close`, `dream`, `dream-cli-flags`, `cycle-last-full-cycle-at`, `cycle-extract-source`, `autopilot-*-wiring`, `autopilot-global-maintenance`, `cli-dream-engine-warn`, `cycle-dream-output-root`) → all pass
- Both new tests confirmed to FAIL on master without the fixes (5 failures pre-fix)
- JSONB guards (`check-jsonb-pattern.sh`, `check-jsonb-params.mjs`) → clean

### DIFF
--- a/src/commands/autopilot.ts
+++ b/src/commands/autopilot.ts
@@ -38,6 +38,7 @@ import { logSelfUpgrade } from '../core/audit/self-upgrade-audit.ts';
 import { detectInstallMethod } from './upgrade.ts';
 import { evaluateQuietHours } from '../core/minions/quiet-hours.ts';
 import { inspectLock } from '../core/db-lock.ts';
+import { registerCleanup } from '../core/process-cleanup.ts';
 
 /**
  * v0.37.7.0 #1162 — classify autopilot reconnect-loop errors.
@@ -433,6 +434,37 @@ export async function runAutopilot(engine: BrainEngine, args: string[]) {
   let stopping = false;
   let childSupervisor: ChildWorkerSupervisor | null = null;
 
+  // #1872: graceful engine shutdown. On PGLite the cycle steps run INLINE in
+  // this process, so a hard `process.exit` mid-write (systemctl stop →
+  // SIGTERM) kills WASM Postgres with the WAL dirty and can corrupt the
+  // brain. Two exit paths must both close the engine:
+  //   - autopilot's own shutdown() below (owns SIGINT + internal stops like
+  //     max_crashes / cycle-failure-cap), and
+  //   - process-cleanup's SIGTERM handler (installed at cli.ts module load;
+  //     it runs the cleanup registry with a 3s deadline and then exits) —
+  //     which is why closeEngine is ALSO registered there.
+  // closeEngine aborts the in-flight inline cycle (runCycle checks the
+  // signal between phases and threads it into phase sub-work), gives it a
+  // short bounded window to wind down, then disconnects. PGLite's
+  // disconnect() drains the pending query and checkpoints before closing;
+  // a second call is a no-op (disconnect snapshots + nulls the handle), so
+  // both paths firing is safe.
+  const shutdownAbort = new AbortController();
+  let inflightInlineCycle: Promise<unknown> | null = null;
+  const closeEngine = async () => {
+    shutdownAbort.abort(new Error('autopilot shutdown'));
+    if (inflightInlineCycle) {
+      // ponytail: 2s cap keeps us inside process-cleanup's 3s deadline; a
+      // between-phase abort resolves instantly, a mid-phase one may not.
+      await Promise.race([
+        inflightInlineCycle.catch(() => { /* cycle errors already logged by the loop */ }),
+        new Promise((r) => setTimeout(r, 2_000)),
+      ]);
+    }
+    try { await engine.disconnect(); } catch { /* best-effort */ }
+  };
+  const deregisterEngineClose = registerCleanup('autopilot-engine-close', closeEngine);
+
   if (spawnManagedWorker) {
     const cliPath = resolveGbrainCliPath();
     // Cgroup-aware auto-sized RSS watchdog cap (issue #1678). The old flat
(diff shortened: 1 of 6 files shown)


felipe-silva-azos:
We're very interested in this fix. We've hit the PGLite corruption described in #1872 on WSL/systemd, and it has kept us from safely enabling autopilot again.

Do you have an expected timeline for merging and releasing this change? We're also happy to help test it on WSL with systemd and PGLite, or provide any additional reproduction details or logs that would be useful.

PR #3183, opened 2026-07-21
fix(schema-pack): takeover of #1799 — reject YAML block scalars loudly instead of silently truncating (#1750)

### LINKED ISSUE #1750: parseYamlMini silently drops every key after a block scalar — gbrain-recommended ships with 0 page_types

time-attack:
Supersedes #1799 (fork branch by @harjothkhara; co-author credit preserved on the commit).

## Why a takeover

#1799 was fleet-verified but its refreshed CI run failed. Diagnosis: the failure was **not** the PR — `test (4)` shard hit the workflow's 15-minute job timeout mid-run (`##[error]The operation was canceled`) while tests were passing, and the `test-status` aggregator turned the cancellation into a red check. The PR's touched files (schema-pack loader/yaml/test) aren't in that shard's failing path. Content re-applied cleanly onto current master; re-verified locally.

## What it does (unchanged from #1799, fixes #1750)

- `parseYamlMini` (`src/core/schema-pack/loader.ts`) documented block scalars (`|`, `>`) as unsupported but silently mis-parsed them: `description: |` parsed as the literal `"|"` and the indented body tripped the `indent > baseIndent` break, dropping every following top-level key with no error. `parseScalar` now throws a clear, actionable error on a bare block-scalar indicator (including chomping/indent forms `|-`, `>+`, `|2`).
- Inlines `gbrain-recommended.yaml`'s `description` so the bundled pack loads its 12 `page_types` (it previously resolved to 0, silently).
- Quoted `"|"` still parses as the string `"|"`; real values like `> 50% growth` are unaffected (guard matches a lone indicator only).

## Verification on current master

- `bun test test/schema-pack-loader.test.ts test/schema-pack-mutate.test.ts test/commands/schema-packpath.test.ts` — 79 pass / 0 fail

### DIFF
--- a/src/core/schema-pack/base/gbrain-recommended.yaml
+++ b/src/core/schema-pack/base/gbrain-recommended.yaml
@@ -26,12 +26,7 @@ name: gbrain-recommended
 version: 1.0.0
 gbrain_min_version: 0.39.0
 extends: gbrain-base
-description: |
-  The recommended starter pack for an operational personal brain.
-  Based on docs/GBRAIN_RECOMMENDED_SCHEMA.md — Karpathy's LLM wiki
-  pattern extended into a full operational knowledge base. Use this
-  when you want the documented behavior immediately instead of
-  authoring your own pack from scratch.
+description: "The recommended starter pack for an operational personal brain. Based on docs/GBRAIN_RECOMMENDED_SCHEMA.md — Karpathy's LLM wiki pattern extended into a full operational knowledge base. Use this when you want the documented behavior immediately instead of authoring your own pack from scratch."
 
 # Pack-specific page_types (extend gbrain-base's seed types).
 # Each maps a path prefix to a primitive + opt-in flags. Most types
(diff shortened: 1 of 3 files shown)


time-attack:
Closing as already fixed. Master's ea6cb025 (#3110) solved #1750 by actually supporting block scalars — `parseBlockScalar()` at src/core/schema-pack/loader.ts:162, wired in for both `|` and `>` at lines 254/256/288/290 — so `gbrain-recommended.yaml`'s `description: |` now parses and the pack loads its full page_types. Since this PR instead makes the loader throw on a block-scalar indicator, applying it would reject the bundled pack that master now handles correctly. Thanks to you and @harjothkhara for chasing this down; the diagnosis in #1799 was right even though the landed fix took the other branch.

PR #3184, opened 2026-07-21
feat(links): resolve path-shaped wikilinks for any directory by exact slug match (takeover of #2978)

time-attack:
Supersedes #2978 (fork PR by @gporterl, skipped in the weekend queue and left behind by later merges). This is a rebase-takeover onto current master; the original author is credited via Co-authored-by.

## Rebase notes

- One conflict in `docs/architecture/KEY_FILES.md`: kept master's newer `extract-conversation-facts` entry, merged the PR's link-extraction additions (the PR's entry was a pure superset of master's).
- Regenerated `llms.txt` / `llms-full.txt` with `bun run build:llms` (the PR's bundle hunk was against stale master).
- No code changes beyond the original diff.

## Summary (from #2978)

Path-shaped wikilinks (`[[any-dir/sub/name]]`) whose directory prefix is not in the hardcoded `DIR_PATTERN` whitelist are currently dropped silently by the DB-path extractor — no candidate, no unresolved report. This adds any-directory exact-slug resolution for path-shaped wikilinks, composing with (not duplicating) the #972 bare-name/basename machinery:

- Two new regex passes (`2a'` qualified / `2b'` unqualified) match path-shaped targets for any directory prefix, run after the whitelist passes with those spans masked (whitelisted-dir behavior stays byte-identical, nothing double-emits).
- Resolution is exact slug match (new optional `SlugResolver.slugExists`, served off the same source-scoped `getAllSlugs` snapshot the basename index uses — still one engine call per run). Gate `link_resolution.any_dir_exact_path` defaults ON, env override documented alongside `global_basename`, registered in `KNOWN_CONFIG_KEYS`.
- A miss is recorded, never silent: rides the existing unresolved-report machinery into the `put_page` `auto_links` response and the extract summary. No ghost edges.
- Qualified `[[src:path]]` refs keep their source pin end-to-end (`LinkCandidate.targetSourceId`); pin miss = skip, pins join the dedup key.
- Scoped extraction keeps the documented local-first→default fallback: `slugExists` resolves over scoped-source ∪ `default`, mirroring `resolveCandidateSources`. Basename index stays strictly source-scoped.
- `LINK_EXTRACTOR_VERSION_TS` is bumped (one-time stale sweep on deployed brains).

Fixes the path-form portion of #1493; addresses #726.

## Verification on this rebase

- `bun run typecheck` exit 0
- Touched suites green: `link-extraction-any-dir` (new), `put-page-autolink-pin-dedup` (new), `link-extraction`, `config`, `extract-db`, `extract-source-aware`, `extract-stale`, `extract-fs`, `build-llms`, `backlinks`, `doctor` — 379 pass / 0 fail
- CI guards clean: `check-jsonb-pattern.sh`, `check-jsonb-params.mjs`, `check-key-files-current-state.sh`


time-attack:
Repaired the semantic collision with #2866 (merged to master after this branch's last rebase): the any-dir exact-path passes were consuming path-shaped wikilink spans before the generic pass where #2866's basename-suffix resolution lives, failing three of its unit tests plus the test-isolation lint.

Fix (d3fc1e7): the two features now compose — exact slugExists match wins; on a miss (or when no verifier is available) the #2866 basename/suffix path is the fallback before anything lands in the unresolved report; gating `any_dir_exact_path` off restores post-#2866 legacy behavior instead of a silent drop. Two #2866 e2e assertions that pinned the basename mechanism for an exactly-existing slug now assert the user-visible guarantee (edge created, sibling excluded). Env-override test moved to withEnv().

Local gates: typecheck 0, `bun run verify` 31/31, all 15 link-extraction-related test files green (369 pass / 0 fail).

time-attack:
Closing — thank you for the careful rebase and for preserving @gporterl's credit; the engineering here is clearly solid. It's a 1500-line feature across 15 files that adds two new extraction passes, a new resolver API, and a new config key defaulting ON, and the `LINK_EXTRACTOR_VERSION_TS` bump forces a one-time stale re-extraction sweep on every deployed brain. That's more link-resolution surface and more upgrade cost than we want to take on, so we're passing on it rather than letting it sit open.

PR #3186, opened 2026-07-21
fix(ai): Ollama local-provider compatibility fixes (takeover of #1854)

time-attack:
Takeover of #1854 by @starm2010 (fork branch, CONFLICTING with master). Supersedes #1854.

## What landed (rebased onto current master)

1. **`FREE_LOCAL_CHAT_PROVIDERS`** in `src/core/budget/budget-tracker.ts` — local chat providers (`ollama`, `llama-server`) return zero pricing, so `--max-cost`-bounded brainstorm/lsd no longer hard-fails on a `lookupPricing` miss. Sibling to the existing `FREE_LOCAL_EMBED_PROVIDERS` / `FREE_LOCAL_RERANK_PROVIDERS` sets.
2. **Numeric idea-id coercion** in `validateIdeaShape` (`src/core/brainstorm/judges.ts`) — models like Kimi K2 return numeric `id` fields; `5` is coerced to `"Idea 05"` before the string type check.
3. **Ollama recipe chat + expansion touchpoints** (`src/core/ai/recipes/ollama.ts`) — `llama3.3` / `qwen3` / `deepseek-r1`, `supports_tools: true`, zero cost, so the recipe system can route reasoning calls to Ollama.
4. **Graceful "no LLM available" message** (`src/core/think/index.ts`) now mentions `chat_model` / `expansion_model` config, not just `anthropic_api_key`.

## What was dropped as superseded

The original PR scoped the `NO_ANTHROPIC_API_KEY` warning to `providerId === 'anthropic'`. Master's #1698 probe-based warning already does this more robustly: `probeChatModel` returns reason `'unavailable'` only for anthropic-without-key, and every other failure is labeled `MODEL_NOT_USABLE:<reason>`. Resolved the conflict by keeping master's version and adding a comment documenting the invariant; the now-unused `splitProviderModelId` import was dropped.

## Testing

- `bun run typecheck` exit 0
- `bun test` on `budget-tracker`, `brainstorm/`, `think-gateway-adapter`, `think-pipeline.serial`, `think-ab`, `model-pricing`, `gateway-model-messages`, `gateway-tier-extended-models`: 195 pass / 0 fail

Co-authored-by credit to @starm2010 is in the commit.

### DIFF
--- a/src/core/ai/recipes/ollama.ts
+++ b/src/core/ai/recipes/ollama.ts
@@ -33,6 +33,20 @@ export const ollama: Recipe = {
       // OLLAMA_NUM_PARALLEL config; no static cap to declare. v0.32 (#779).
       no_batch_cap: true,
     },
+    chat: {
+      models: ['llama3.3', 'qwen3', 'deepseek-r1'],
+      supports_tools: true,
+      supports_subagent_loop: false,
+      supports_prompt_cache: false,
+      cost_per_1m_input_usd: 0,
+      cost_per_1m_output_usd: 0,
+      price_last_verified: '2026-06-02',
+    },
+    expansion: {
+      models: ['llama3.3', 'qwen3', 'deepseek-r1'],
+      cost_per_1m_tokens_usd: 0,
+      price_last_verified: '2026-06-02',
+    },
   },
   setup_hint: 'Install Ollama from https://ollama.ai, then `ollama pull nomic-embed-text` and `ollama serve`.',
 };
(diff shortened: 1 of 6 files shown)


time-attack:
Closing — thank you, and to @starm2010 for the original. This isn't a fix to existing behavior: it adds a chat and expansion touchpoint to the Ollama recipe and rewrites the test that deliberately asserts ollama is embedding-only, so it's a provider-surface expansion rather than a repair. Hardcoded third-party model lists (`llama3.3`, `qwen3`, `deepseek-r1`) are the kind of surface that goes stale and needs maintaining forever, and separately the Kimi id coercion in judges.ts mutates the caller's object in place while guessing at an `Idea NN` format. We're keeping the provider matrix narrow on purpose.

PR #3189, opened 2026-07-22
v0.42.66.1 fix(chronicle): mirror-source backfill guard hardening (#2786)

### LINKED ISSUE #2786: bug(chronicle): backfill scoped to a source holding same-slug mirrors silently derives sole-copy db-only life/events pages under that source
**What happened?**

During cleanup of a legacy artifacts source that carried same-slug mirror copies of default-source pages, we found **19 `life/events/*` pages derived from meeting pages (whose canonical home is `default`) sitting under the LEGACY source** — and those wrong-source rows were the ONLY copies: db-only, no file backing, one `sources remove` away from silent data loss. Observed 2026-07-08 on v0.42.57.

Root cause, verified at HEAD (a25209bb) — and correcting our own initial diagnosis, so a maintainer doesn't have to: there is **no cross-source stamping bug inside the extractor.** `runChronicleExtract` (`src/core/chronicle/extract-events.ts:99-164`) reads the source page and writes every derived event strictly under `opts.sourceId` — the scope of the enqueue (`chronicle_backfill`: `operations.ts:5303`; put_page backstop: `chronicle/backstop.ts:33`), and page selection is source-scoped (`sourceScopeOpts`, `operations.ts:424`). The failure needs a precondition: **same-slug copies existing under two sources.** Then a backfill scoped to the mirror source derives a full, legitimate-looking set of `life/events/*` under that source — db-only, indistinguishable from canonical derivations, with no warning that the slugs it processed are mirrors of another source's canonical pages.

So the defect class is: **chronicle derivation has no awareness of the mirrored-slug condition that `multi_source_drift` already knows how to detect**, and its output placement (db-only, under whatever source was scoped) makes the resulting fork invisible until someone audits or deletes a source.

**What did you expect?**

Any of:
1. Derived pages carry provenance of the origin page's source (not just `event.depth` path), so a wrong-source fork is detectable and repairable;
2. `chronicle_backfill` warns (or skips) when the scoped source's page is a same-slug mirror of another source's page — the `multi_source_drift` detection logic already identifies exactly this condition;
3. At minimum, docs state that chronicle output is enqueue-source-scoped and that mirrored slugs will fork derived history under whichever source is backfilled.

**Steps to reproduce**

1. Same slug under source A (canonical, file-backed) and source B (mirror copy).
2. `chronicle_backfill` scoped to B.
3. `life/events/*` derived under B, db-only — sole copies if A was never backfilled; removing B destroys them.

**Environment**

`multi_source_drift` flags the mirror condition itself (that's how we found the mirrors) but nothing flags the derived-page fork.

---

Local remediation (one-time): chronicle-backfill re-derived the events at `default` from the canonical meeting pages, then the legacy source was removed.

Masashi-Ono0611:
## Summary

Fixes #2786 — a backfill scoped to a source holding same-slug mirrors of `default`-source pages could derive `life/events/*` pages that exist ONLY under that mirror source: db-only, no file backing, one `sources remove` away from silent, unrecoverable data loss.

Adds a mirror guard to `runChronicleExtract`: before deriving under a non-`default`, non-durable (no `local_path`, or archived) source, it checks whether a byte-identical, same-slug copy already exists under a durable/canonical source (`default` or an active file-backed source). If so, derivation is refused (`status: 'skipped', reason: 'mirrored_source_slug'`) instead of forking a sole, removable copy of the derived history under the wrong source.

Also closes a read-side information-leak vector this guard's own result field opened up, and centralizes a `statement_timeout` GUC-leak fix in a shared connection primitive. Details below.

## What's in this PR

- **Mirror guard** (`multi-source-drift.ts`, `chronicle/extract-events.ts`): single-slug DB lookup (`findMirrorSources`) restricted to durable/canonical sources so two non-canonical DB-only mirrors don't mutually block each other, and same-slug-different-content across sources (the supported multi-tenant case) is never blocked. Date comparison (`content_hash` + `effective_date`) treats fallback/NULL `effective_date_source` (legacy pre-v0.29.1 rows, or pages where no real date could be inferred) as "no real date" and skips the date check rather than wrongly gating on it.
- **Migration v125** (`pages_chronicle_mirror_idx`): the mirror lookup runs once per page on the Life Chronicle hot path; without a `(slug, content_hash)` partial index it becomes an unindexed sequential scan repeated per candidate page. `CREATE/DROP INDEX CONCURRENTLY` run as separate top-level statements (a `DO` block can't run `CONCURRENTLY` DDL) on a `withReservedConnection` session.
- **Read-side redaction**: the guard's own result field can reveal that a hidden/foreign source's copy exists, to a caller who wasn't the one who submitted the job. `get_job`/`list_jobs`/`cancel_job`/`retry_job`/`replay_job` now redact `data`/`result` for job kinds in a new, deliberately narrow `CROSS_SOURCE_SENSITIVE_JOB_NAMES` set (currently just `chronicle_extract`) for any `ctx.remote !== false` caller. This is a separate set from `PROTECTED_JOB_NAMES` (which gates *submission*, mostly for cost control on jobs like `subagent`) — reusing that set for read-redaction would have broken the supported remote `submit_agent` → `get_job` result-readback workflow.
- **`chronicle_extract`** registered in `PROTECTED_JOB_NAMES` so it can't be submitted directly over MCP to probe for mirrors.
- **`withReservedConnection` (postgres-engine.ts)**: now resets `statement_timeout` before releasing a session back to the pool. On dual-pool Postgres this primitive routes through a shared, reused direct/DDL pool; several existing sql-declared migrations already shorten this GUC on their reserved session without resetting it, which used to be relatively harmless (regular pool) but becomes a real problem once shared with other long-running DDL callers (`vector-index.ts` rebuilds, `backfill-base.ts` writes). Centralized fix, not scoped to this PR's own migration.

## Testing

- `bunx tsc --noEmit` — clean.
- `bun test` on the 15 affected files — 383 pass, 0 fail.
- `bun run build:llms` regenerated after `KEY_FILES.md` edits.

## Note on provenance

This issue was reported by @alexputici with an exceptionally thorough root-cause writeup — root-caused down to exact line numbers and the precondition needed to trigger it. No PR was open against #2786 at the time of this submission. Offered in a collaborative spirit — if @alexputici (or anyone else) already has a fix in flight, happy to defer, or to have this reviewed alongside theirs.


time-attack:
Merging. I verified #2786 is still unguarded on master — there is no `findMirrorSources` in multi-source-drift.ts and no mirror awareness in the chronicle extractor — and the issue carries a real production observation of sole-copy db-only derived pages, which is the data-loss class we always fix. The migration is correctly renumbered to v126 against master's v125, and roughly half the diff is tests. Please retitle to the current release version before the final commit, since v0.42.64.0 has already shipped.

Masashi-Ono0611:
Thanks for merging, and for verifying `findMirrorSources` is genuinely absent on master before taking it. Retitled to `v0.42.66.1` (current master) as requested.

PR #3199, opened 2026-07-22
fix(sync): make full reconcile convergence truthful

caterpillarC15:
## Problem

A full sync currently advances the source bookmark through the failure gate before deletion reconciliation runs. If the mass-delete safety valve refuses the deletion, or if one or more deletes fail, the command can still return first_sync with an advanced bookmark. The next ordinary sync then reports up to date even though the database did not converge to the Git tree.

This was reproduced against real Postgres with 21 committed pages and 11 committed files removed: the safety warning fired, all 21 pages remained, the bookmark advanced to the deletion commit, and the next sync reported up to date.

## Change

- Reconcile before advancing the bookmark.
- Return structured blocked_by_reconcile results for mass-delete refusal, deletion failure, and convergence mismatch.
- Keep the bookmark unchanged on every blocked path so a retry is honest.
- Verify expected deletions are absent before reporting success.
- Preserve never-committed DB-only pages.
- Thread the explicit source ID through the sync_brain full-sync path.

This is a general sync correctness fix. GraceDB consumes the structured result in immutable execution receipts, but no GraceDB-specific behavior is embedded here.

## Evidence

- Focused suite: 28 passed, 0 failed, 77 expectations.
- Real-Postgres actual-handler test: 1 passed, 0 failed, 10 expectations; proves refusal, unchanged bookmark, explicit retry, deletion, and advance.
- 30 of 31 local verify checks pass on macOS. The only remaining check is the pre-existing src// path-normalization failure on official master, isolated in PR #3198; running that fixed scanner against this branch passes.
- git diff --check passes.

### DIFF
--- a/src/commands/sync.ts
+++ b/src/commands/sync.ts
@@ -177,8 +177,8 @@ function resolveSyncYieldEvery(): number {
  * extraction-lag nudge. Fires on every non-error completion — crucially
  * `first_sync` (a fresh / --full import is the BIGGEST un-extracted backlog) and
  * `up_to_date` (a no-op sync over a brain with a pre-existing backlog still
- * warrants the nudge). Excludes `dry_run` (preview) / `blocked_by_failures` /
- * `partial` (inconsistent state). Pure so D5's contract is unit-testable
+ * warrants the nudge). Excludes `dry_run` (preview), both blocked statuses,
+ * and `partial` (inconsistent state). Pure so D5's contract is unit-testable
  * without driving the CLI.
  */
 export function shouldNudgeAfterSync(status: SyncResult['status']): boolean {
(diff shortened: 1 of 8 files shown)


time-attack:
Merging — this is a genuine correctness bug: the bookmark advances through the failure gate before delete reconciliation, so a refused mass-delete or a failed delete leaves the DB diverged while the next sync happily reports `up_to_date`. Verifying convergence from database state instead of affected-row counts is the right call. Two things to watch post-merge: the legacy no-sourceId path now reconciles `DEFAULT_SOURCE_ID` where it previously skipped reconciliation entirely, and a legitimate bulk deletion now hard-blocks the bookmark until `GBRAIN_ALLOW_MASS_RECONCILE=1` is set — both are the honest behavior, but they are behavior changes worth a CHANGELOG line.

PR #3200, opened 2026-07-22
feat(mcp): export governed downstream dispatch and rate limits

caterpillarC15:
## Context

Downstream hosts can import operation definitions, but the native dispatcher is internal and always resolves from the complete registry with ambient process config. A multi-tenant host that needs a fixed capability subset must otherwise call handlers directly or recreate validation, context construction, error envelopes, and tool schema behavior. Both options drift from gbrain.

## Change

- Add a public governed dispatcher built from exact Operation objects supplied at construction time.
- Freeze the capability ceiling; a request-time resolver may narrow access but cannot add or swap a handler.
- Require explicit source authority and explicit config/auth context.
- Preserve native parameter validation and OperationError envelopes.
- Scrub unexpected exception and metadata-hook details from remote responses.
- Export the existing native MCP rate limiter instead of requiring downstream copies.
- Document the embedding contract and lock it with package-path tests.

The first consumer is GraceDB, which retains authentication, routing, policy, and audit while delegating native operation execution to gbrain. The API is intentionally generic and contains no GraceDB types.

## Evidence

- Focused package-path and dispatcher suite: 47 passed, 0 failed, 132 expectations.
- Covers immutable capability ceilings, no widening or handler swapping, explicit source authority, frozen config/auth, native validation, error preservation, secret scrubbing, duplicate rejection, and public exports.
- git diff --check passes.


time-attack:
Closing after maintainer review — decision: not accepting this. No security posture change; gbrain keeps its rule that permission checks always run inside gbrain itself. The requesting project must integrate another way.

PR #3201, opened 2026-07-22
feat(api): export the operation checkpoint contract

caterpillarC15:
## Context

gbrain already owns durable operation checkpoints, stable fingerprints, append-only completed-path recording, and resume filtering. These primitives are internal, so downstream ingestion orchestrators must either duplicate checkpoint state or reach through an unsupported relative path.

## Change

- Add a narrow gbrain/operation-checkpoints public barrel.
- Export only the generic fingerprint, load, record, append, clear, and resume-filter primitives.
- Add package-path contract tests and documentation.
- Keep storage, migration, and JSONB behavior owned by the existing native implementation.

GraceDB uses this seam to bind versioned corpus ingestion attempts to native durable checkpoints instead of creating a second checkpoint engine. No GraceDB-specific identifiers or schema are added here.

## Evidence

- Public export suite: 41 passed, 0 failed, 108 expectations.
- Real-Postgres JSONB parity: 3 passed, 0 failed, 8 expectations; completed keys remain a JSON array and replacement semantics round-trip.
- git diff --check passes.


time-attack:
Closing after maintainer review — decision: not accepting this. The outside project has to wait or copy the code; gbrain keeps freedom to change its internals. Nothing breaks.

PR #3203, opened 2026-07-22
fix(budget): make paid MCP spend atomic and fail closed

caterpillarC15:
## Problem

The MCP image-spend path performs a read/check, makes the paid provider call, then records spend best-effort. Concurrent requests can observe the same headroom and exceed the client cap. A successful paid response can also escape without durable accounting when the post-call write fails.

The existing reserve helper documents a transaction-scoped advisory lock, but its sweep, read, and insert currently execute as separate operations. Settlement similarly updates the reservation and inserts the spend mirror separately.

## Change

- Run same-client sweep, committed/pending read, cap decision, and reservation insert in one transaction under a Postgres advisory lock.
- Validate reservation inputs and fail closed on malformed accounting reads.
- Settle reservation state and spend-log attribution atomically.
- Record late paid outcomes even after hold expiry and distinguish settled from missing reservations.
- Reserve search_by_image spend immediately before the paid call.
- Settle a documented pessimistic amount on ambiguous provider failure.
- Block successful responses when accounting settlement fails.

This PR changes correctness only. It does not add a new package export; the public AI/budget seam is a separate dependent request.

## Evidence

- Focused meter and image-operation suite: 30 passed, 0 failed, 64 expectations.
- Real-Postgres concurrency: 20 simultaneous 10-cent attempts under a 100-cent cap produce exactly 10 committed reservations and 10 BudgetExceeded rejections; 1 test, 13 expectations.
- Transaction rollback, idempotent settlement, late settlement, token attribution, fail-closed reads, and pessimistic error charging are covered.
- git diff --check passes.

### DIFF
--- a/src/core/minions/budget-meter.ts
+++ b/src/core/minions/budget-meter.ts
@@ -1,11 +1,11 @@
 /**
- * v0.38 Slice 2 — budget meter for the subagent tool loop.
+ * Durable reserve-then-settle meter for paid OAuth/MCP operations.
  *
- * Reserve-then-settle pattern (D3) prevents the "concurrent agents bust the
+ * Reserve-then-settle pattern (D3) prevents the "concurrent requests bust the
  * cap" race that the pre-v82 best-effort post-call recording allowed. Two
- * agents from the same OAuth client both pre-flight pass at $2 of $5,
- * both spend $2, total spend = $4 of $5 → fine. But raise the per-agent
- * estimate to $3 and both agents see "$5 cap - $2 spent = $3 headroom, ok"
+ * calls from the same OAuth client both pre-flight pass at $2 of $5,
+ * both spend $2, total spend = $4 of $5 → fine. But raise the per-call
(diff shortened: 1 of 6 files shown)


caterpillarC15:
Downstream consumer context: GraceDB v3 keeps gbrain as the tenant-brain dependency and treats gbrain BudgetLedger as the authoritative spend ledger; it must not recreate reservation/accounting logic in the governance database. The atomic paid-MCP behavior in this PR is therefore carried temporarily in the maintained GraceDB integration candidate a610ee00e7e17e74e538f698a1c00b194668926f, built on official d9eb027bdd75928bf39276769af5cbbc4e375e43. The combined candidate passed gbrain verify (31/31) and the GraceDB v3 gate (115 DB-free and 185 real-Postgres tests), including concurrent spend-cap coverage. Consumer integration is reviewable in caterpillarC15/gracedb#2. If this lands upstream, GraceDB can drop the corresponding fork patch on its next audited rebase.

PR #3204, opened 2026-07-22
fix(budget): record actual resolver spend before cap error

caterpillarC15:
## Problem

BudgetLedger.commit currently clamps an already-incurred provider charge to remaining headroom and throws from inside the transaction. Throwing rolls back the reservation and ledger updates, so the paid call can remain pending and the ledger deliberately records less than the provider actually charged.

A cap can prevent the next call; it cannot retroactively reduce a completed provider charge.

## Change

- Record actualUsd in full and finalize the reservation in the transaction.
- Return a cap_exceeded error value from the transaction and throw it only after commit.
- Include other live reservations when calculating the overage.
- Keep negative and non-finite input rejection unchanged.

This is a narrow accounting-correctness PR and does not add package exports.

## Evidence

- Enrichment/BudgetLedger suite: 24 passed, 0 failed, 51 expectations.
- Regression asserts the full actual charge is committed, held amount is released, reservation is finalized, and cap_exceeded is reported after durable accounting.
- Parallel reservation, rollback, idempotency, scope isolation, and TTL cleanup remain green.
- git diff --check passes.

### DIFF
--- a/src/core/enrichment/budget.ts
+++ b/src/core/enrichment/budget.ts
@@ -58,7 +58,11 @@ export interface BudgetStateRow {
 // Errors
 // ---------------------------------------------------------------------------
 
-export type BudgetErrorCode = 'reservation_not_found' | 'already_finalized' | 'invalid_input';
+export type BudgetErrorCode =
+  | 'reservation_not_found'
+  | 'already_finalized'
+  | 'invalid_input'
+  | 'cap_exceeded';
 
 export class BudgetError extends Error {
   constructor(public code: BudgetErrorCode, message: string, public reservationId?: string) {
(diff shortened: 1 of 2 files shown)


caterpillarC15:
Downstream consumer context: GraceDB v3 keeps gbrain as the tenant-brain dependency and relies on gbrain BudgetLedger as the authoritative record of actual provider spend; the GraceDB governance database must not duplicate or reinterpret that ledger. The commit-before-cap-error behavior in this PR is therefore carried temporarily in the maintained GraceDB integration candidate a610ee00e7e17e74e538f698a1c00b194668926f, built on official d9eb027bdd75928bf39276769af5cbbc4e375e43. The combined candidate passed gbrain verify (31/31) and the GraceDB v3 gate (115 DB-free and 185 real-Postgres tests), including spend exhaustion and retry coverage. Consumer integration is reviewable in caterpillarC15/gracedb#2. If this lands upstream, GraceDB can drop the corresponding fork patch on its next audited rebase.

PR #3208, opened 2026-07-22
fix(cli): route init --help to its own usage text

alexey-metaengage:
## Summary

`gbrain init --help` prints the generic one-line CLI-only short-circuit (`printCliOnlyHelp`) instead of `init`'s own usage. In `src/cli.ts`, `init` is in `CLI_ONLY` but not in `CLI_ONLY_SELF_HELP` — and the dispatcher fires the generic short-circuit for any `CLI_ONLY` command that isn't in the self-help set, so `init`'s detailed usage is unreachable.

## The fix

- Add `'init'` to `CLI_ONLY_SELF_HELP` so `init --help` routes to its own usage text (same treatment the other self-documenting CLI-only commands already get).

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -59,6 +59,7 @@ export const CLI_ONLY = new Set(['init', 'reinit-pglite', 'upgrade', 'post-upgra
 // excluded from the generic short-circuit so detailed per-command and
 // per-subcommand usage stays reachable.
 const CLI_ONLY_SELF_HELP = new Set([
+  'init',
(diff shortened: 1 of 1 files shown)


time-attack:
Merging — I confirmed this on current master: `init` is in `CLI_ONLY` but not `CLI_ONLY_SELF_HELP`, so the dispatcher's generic short-circuit at src/cli.ts:305 fires and the `printInitHelp()` guard already sitting in src/commands/init.ts is unreachable dead code. One-line fix that makes existing intent work. Thanks for the precise diagnosis.

PR #3212, opened 2026-07-22
fix(migrate): tolerate non-superuser roles lacking BYPASSRLS

alexey-metaengage:
## Summary

The RLS-backfill migrations v24 / v29 / v31 / v35 (`src/core/migrate.ts`) `RAISE EXCEPTION` when the current role lacks `BYPASSRLS`, aborting the migration. On managed Postgres (e.g. Cloud SQL) the application role is a non-superuser that cannot hold `BYPASSRLS` but still **owns** the tables. Since gbrain ships no policies and no `FORCE ROW LEVEL SECURITY`, the owner is exempt from RLS and `ENABLE ROW LEVEL SECURITY` is a harmless no-op. The hard abort permanently wedges the migration chain on such instances — the runner never advances past the failing version.

## The fix

- In the four `IF NOT has_bypass` guards, change `RAISE EXCEPTION` to `RAISE WARNING` and fall through to the `ENABLE ROW LEVEL SECURITY` statements. A `BYPASSRLS` role behaves exactly as before; a non-superuser owner now warns and proceeds (safe on a single-owner database) instead of wedging.

### DIFF
--- a/src/core/migrate.ts
+++ b/src/core/migrate.ts
@@ -906,13 +906,14 @@ export const MIGRATIONS: Migration[] = [
       BEGIN
         SELECT EXISTS (SELECT 1 FROM pg_roles pr WHERE pg_has_role(current_user, pr.oid, 'USAGE') AND (pr.rolbypassrls OR pr.rolsuper)) INTO has_bypass; -- #1385: superuser + inherited-role BYPASSRLS, not just the role's own rolbypassrls
         IF NOT has_bypass THEN
-          -- Fail the migration loudly instead of WARNING + version-bump.
-          -- The runner unconditionally records schema_version on success,
-          -- so a silent WARNING here would permanently lock the backfill out
-          -- on future runs even after switching to a bypass role. Raising
-          -- aborts the transaction, leaves schema_version at the prior value,
-          -- and lets the next invocation retry after the role is fixed.
-          RAISE EXCEPTION 'v24 rls_backfill_missing_tables: role % does not have BYPASSRLS privilege — cannot enable RLS safely. Re-run as postgres (or another BYPASSRLS role). The migration will retry automatically on the next initSchema call.', current_user;
+          -- A non-superuser table OWNER (e.g. the app role on managed
+          -- Postgres) can never hold BYPASSRLS, but it OWNS these tables and
+          -- gbrain sets no FORCE ROW LEVEL SECURITY and no policies, so the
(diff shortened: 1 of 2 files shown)


alexey-metaengage:
Thanks for the review in #3283 — both defects you flagged were real, and I've confirmed and repaired them here. Rebased onto `master` @ 3fafb69b.

1. **Didn't typecheck.** Correct. Genericizing the comment for upstream introduced markdown backticks around a role name *inside* the `MIGRATIONS` template literal, which terminated the string early: `bun run typecheck` exited 2 with ~40 syntax errors. Reworded without backticks. (No CI checks report on this branch, so nothing caught it — my miss for not running typecheck before pushing.)

2. **Stale test pins.** Correct. `test/migrate.test.ts` asserted the old contract in three places (v24: `RAISE EXCEPTION` present + `RAISE WARNING` absent; v31: `RAISE EXCEPTION` present). They now pin the new contract, and I took your point about pinning *warn-and-proceed, never warn-and-skip* — the v24 test also asserts the `ALTER … ENABLE ROW LEVEL SECURITY` statements sit after the guard's `END IF`, so a future refactor can't move them inside the branch and silently reintroduce the original hazard.

Also folded your fail-closed reasoning into the SQL comment, since it's the part that makes this safe and wasn't written down: the migration stays fail-closed via **ownership** rather than BYPASSRLS — `ENABLE ROW LEVEL SECURITY` requires ownership, so a non-owner role aborts the block with `schema_version` unbumped.

Verified on the rebased branch:
- `bun test test/migrate.test.ts -t BYPASSRLS` — 4 pass / 0 fail

**On overlap:** #3283 supersedes this PR and I'm glad to have either land — no objection to closing this in favor of the wave, and thanks for the `Co-authored-by`. I've repaired it mainly so the fix isn't blocked on the 3-PR bundle if any part of that wave needs more work. Your call which route is easier to merge.

time-attack:
Thanks for the careful writeup — the owner-exemption reasoning is correct as far as it goes, and the diff is clean. We're going to keep the fail-closed posture here, though: the safety argument depends on the migrating role, the table owner, and the runtime role all being the same principal, which gbrain doesn't enforce. On the very common managed-Postgres split (migrations as the owning role, app connecting as a separate limited role) warn-and-proceed enables RLS with no policies and the app role then reads zero rows everywhere — a silent total-retrieval failure that's far worse than a loud, actionable migration abort. Closing as a security-posture relaxation we're not taking.

Issue #1777, opened 2026-06-02
archive/ content is ingested but silently hard-excluded from all search by default (in-repo should mean findable)


garrytan-agents:
### Summary

A markdown page that lives in the brain repo under `archive/` cannot be found by `query`/`search`/`ask` unless the caller explicitly opts the prefix back in. There is no signal at call time that an entire subtree was withheld from the result set. For a large brain this produces a class of "the page exists in the repo but the agent swears it doesn't exist" failures — the agent queries, gets nothing, and confidently concludes the content isn't there.

The guiding principle this violates: **if it's committed to the brain repo, it should be embedded, graphed, and findable by default.** Findability should be opt-out, not opt-in.

### Root cause

Two independent gates, only the second of which is the problem:

1. **Ingest gate** — `isSyncable` / `pruneDir` in `src/core/sync.ts`. Prune set is small (`node_modules`, `.raw`, `ops`). `archive/` is **not** pruned here, so archive pages are eligible for embedding + graphing on sync. (Good.)

2. **Search gate** — `DEFAULT_HARD_EXCLUDES` in `src/core/search/source-boost.ts`:
   ```ts
   export const DEFAULT_HARD_EXCLUDES: string[] = [
     'test/',
     'archive/',
     'attachments/',
     '.raw/',
   ];
   ```
   `resolveHardExcludes()` unions this into every search and subtracts matching slug prefixes from results unless a caller passes `include_slug_prefixes: ['archive/']`. So `archive/` pages are embedded but **invisible to the default search path**, with no diagnostic emitted.

`test/`, `attachments/`, and `.raw/` are legitimately noise. `archive/` is not categorically noise — it routinely holds high-signal historical content (imported conversation exports, prior-system logs, older notes) that users absolutely expect to retrieve.

### Impact

- Whole high-value subtrees are unreachable via the primary retrieval surface; only raw `grep` over the working tree finds them.
- The exclusion is invisible: an empty result set is indistinguishable from "withheld by a hardcoded prefix," so callers (human or agent) wrongly conclude the content does not exist.
- Scales with brain size — the bigger and older the brain, the more lands in `archive/` and disappears.

### Repro

1. Have any page at `archive/.../something.md` that is synced/embedded.
2. `gbrain search "<an exact phrase from that page>"` → no results.
3. `gbrain query "<topic of that page>"` → no results.
4. The page is present on disk and (verify) present in `pages`; only the search gate hides it.

### Proposed fix

[...]

PR #3213, opened 2026-07-22
fix(search): fold resolved source-boost map into query-cache knobs hash

SVZTeam:
## Problem

The semantic query cache keys on mode knobs, embedding column/provider, schema pack, and (since #2825) the resolved hard-exclude list — but **not the source-boost map**. Boosts only apply at DB-query build time (cache miss), so after any ranking-policy change (a `GBRAIN_SOURCE_BOOST` tune, or an upgrade that ships new `DEFAULT_SOURCE_BOOSTS` — e.g. the #1777 `archive/` demote needed a manual version bump precisely because of this gap) lookups keep being served rows ranked under the **old** policy for up to `cache.ttl_seconds`.

In practice this makes boost tuning appear nondeterministic: identical queries return different rankings depending on which policy wrote the cache row. We hit this operating gbrain against a 240K-page multi-source company corpus where the default boost prefixes (`originals/`, `concepts/`, `writing/`) don't exist and `GBRAIN_SOURCE_BOOST` carries the entire ranking policy — every tune was invisibly masked by stale cache rows until the TTL expired.

## Fix (same pattern as #2825's `hardExcludes` fold)

- `KnobsHashContext` gains `sourceBoostMap?: Record<string, number>` — canonicalized (entries sorted by prefix) so map insertion order is irrelevant; `undefined` falls back to the literal `'default'` for legacy callers.
- `hybridSearchCached` threads `resolveBoostMap()` into the cache key alongside `resolveHardExcludes()`.
- `KNOBS_HASH_VERSION` **12 → 13** (one-time global cold-miss on upgrade; refills within `cache.ttl_seconds`, same as prior bumps).
- A future boost-policy default change no longer needs a manual version bump — the hash sees it directly.

## Tests

- Version-pin tests updated (12 → 13) in `test/search-mode.test.ts` + `test/search/knobs-hash-reranker.test.ts`.
- New v=13 participation suite: different maps → different hashes; same map in different insertion order → same hash; single-prefix factor change → different hash; undefined fallback stable and distinct from an explicit resolved map.
- `bun test test/search-mode.test.ts test/search/knobs-hash-reranker.test.ts test/query-cache-knobs-hash.serial.test.ts` → 105 pass / 0 fail. `tsc --noEmit` clean.

### DIFF
--- a/src/core/search/hybrid.ts
+++ b/src/core/search/hybrid.ts
@@ -15,7 +15,7 @@ import type { SearchResult, SearchOpts, HybridSearchMeta } from '../types.ts';
 import { embed, embedQuery } from '../embedding.ts';
 import { registerBackgroundWorkDrainer } from '../background-work.ts';
 import { resolveEmbeddingColumn, isCacheSafe } from './embedding-column.ts';
-import { resolveHardExcludes } from './source-boost.ts';
+import { resolveHardExcludes, resolveBoostMap } from './source-boost.ts';
 import {
   resolveAdaptiveReturn,
   applyAdaptiveReturn,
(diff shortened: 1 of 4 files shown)


time-attack:
Verified against master: the knobs hash folds hardExcludes but not the resolved boost map, so a ranking-policy change really is masked by stale cache rows for up to the cache TTL. This is the same contamination class as #2825 and it's fixed the same way — sorted canonical hash input, version pin bumped, tests updated. Merging. Note the same fix also appears bundled in #3283; this focused version is the one we're taking.

PR #3230, opened 2026-07-22
fix(doctor): checkSubagentCapability honors explicit models.subagent before tier resolution (#2112 residual)

Masashi-Ono0611:
## Summary

Fix-wave A (#2820) superseded most of #2112 but missed its `doctor.ts` hunk: `checkSubagentCapability` never consulted the explicit `models.subagent` override — it only read `models.tier.subagent` and `models.default`. Implemented per the maintainer's TODOS.md entry ("P2 — cherry-pick #2112's uncovered doctor.ts hunk").

- Resolution order is now `models.subagent` (explicit) > `models.tier.subagent` > `models.default`, matching the runtime resolver's own precedence (which already honors `models.subagent` above the tier value) — the doctor check was silently out of sync with what actually runs.
- The `ok`-path message now names the explicit model when it's the source of resolution.
- The `degraded:no_caching` fix hint is hardcoded to recommend the scoped `models.subagent` override (not the triggering `${source}` label) — otherwise a warning that originated from `models.default` would tell the operator to rewrite the global default model for every workload just to fix a subagent-only cost concern. Caught in codex review round 1 (P2), fixed + regression-tested in round 2.
- Removes the now-completed P2 entry from `TODOS.md`.

## Test plan

New `test/doctor-subagent-capability.test.ts` (4 cases, PGLite-hermetic):
- explicit `models.subagent` (bad model) wins over `models.tier.subagent` (good model) → warn attributed to `models.subagent`.
- `models.subagent` unset, `models.tier.subagent` (bad) → warn attributed to `models.tier.subagent` (unchanged fallback order).
- explicit `models.subagent` (good) → ok, message names the explicit model.
- `degraded:no_caching` reached via `models.default` → fix hint recommends `models.subagent`, never `models.default`.

Verified red-before-green: stashed the `src/commands/doctor.ts` change (keeping the new test), confirmed 2 of 4 cases fail against pre-fix code, then restored the fix and confirmed all 4 pass.

- [x] `bun test test/doctor-subagent-capability.test.ts` — 4 pass
- [x] `bun test test/doctor.test.ts test/doctor-embedding-env-override.test.ts test/doctor-subagent-health.test.ts` — no regressions (100 pass across the 4 files combined)
- [x] Two rounds of `codex exec` (gpt-5.6-sol, high reasoning) review: round 1 flagged a P2 (the `${source}` caching-hint issue above), fixed and re-tested; round 2 — no remaining findings.

### DIFF
--- a/TODOS.md
+++ b/TODOS.md
@@ -2,10 +2,6 @@
 
 ## community fix-wave follow-ups (filed v0.42.60.0)
 
-- [ ] **P2 — cherry-pick #2112's uncovered doctor.ts hunk.** Fix-wave A (#2820) superseded
-  most of #2112 but not its `checkSubagentCapability` fix (check explicit `models.subagent`
-  before `models.tier.subagent`). Refile or cherry-pick; the rest of that PR is covered.
-
 ## v0.42.59.0 follow-ups (five-fix rollup #2735–#2739)
 
 Filed as follow-ups from v0.42.59.0 (bootstrap probe for
--- a/src/commands/doctor.ts
+++ b/src/commands/doctor.ts
@@ -3046,7 +3046,11 @@ export async function checkSubagentCapability(engine: BrainEngine): Promise<Chec
             `${source} is "${resolved}" — provider does not support prompt caching. ` +
             `The subagent loop runs hot (cost scales linearly with conversation length). ` +
             `For lower cost on long loops, use an Anthropic model: ` +
-            `\`gbrain config set ${source} anthropic:claude-sonnet-4-6\`.`,
+            // Always recommend the scoped `models.subagent` override here (not
+            // `${source}`) — when source is `models.default`, `${source}`
+            // would recommend rewriting the global default model for every
+            // workload instead of just the subagent loop.
+            `\`gbrain config set models.subagent anthropic:claude-sonnet-4-6\`.`,
         };
       }
       return null;
--- a/src/commands/doctor.ts
+++ b/src/commands/doctor.ts
@@ -3011,6 +3011,7 @@ async function checkEmbeddingEnvOverride(engine: BrainEngine): Promise<Check> {
 export async function checkSubagentCapability(engine: BrainEngine): Promise<Check> {
   try {
     const { classifyCapabilities } = await import('../core/ai/capabilities.ts');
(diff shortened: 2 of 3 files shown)


Masashi-Ono0611:
Superseded by master — and the remaining delta would now be a regression. Flagging rather than rebasing.

I went to resolve this branch's conflict so it wouldn't cost you a takeover rebase, and found master already does the load-bearing part:

```
// src/commands/doctor.ts (master)
let resolvedSource: string | null = null;
if (modelsSubagent) {          resolvedSource = 'models.subagent';       ... }
else if (modelsDefault) {      resolvedSource = 'models.default';        ... }
else if (tierSubagent) {       resolvedSource = 'models.tier.subagent';  ... }
```

That covers the three things this PR was for: `models.subagent` is consulted, the resolved source is named in the message, and the `degraded:no_caching` hint is hardcoded to the scoped `models.tier.subagent` key rather than the triggering `${source}`.

**Where my branch is now wrong.** It orders `models.tier.subagent` ahead of `models.default`, on the stated grounds that this "matches the runtime resolver's own precedence". Checking `src/core/model-config.ts` on master, the runtime is the other way around:

```
 *   2. New-key config (models.subagent)
 *   4. Global default (models.default)
 *   5. Tier override (models.tier.subagent)
```

and `resolveModel` returns at step 4 before ever reading the tier key. So **master's doctor order matches the runtime and mine no longer does.** Merging my side would put the check back out of sync with what actually runs — the same failure mode #2112 was filed about in the first place.

I'd rather not hand you a conflict resolution that reintroduces that. Happy to close this; leaving it open for a moment in case you'd prefer the branch kept for reference.

One thing I did *not* verify: whether `models.default` beating `models.tier.subagent` at runtime is itself intended. It reads a little surprising that a general default outranks a subagent-specific tier override, but that's a separate question from this PR and I'm not going to relitigate it here on a guess.

Masashi-Ono0611:
Closing per the analysis above: master already covers what this PR was for, and my remaining delta (tier before default) contradicts `resolveModel`'s actual order, so merging it would put doctor back out of sync with the runtime.

Leaving the branch on my fork in case the ordering question is ever worth revisiting on its own.

PR #3234, opened 2026-07-22
fix(cycle+migrations): extract_facts legacy-row guard counts only active rows + v0_32_2 drift-repair lane (#2646)

Masashi-Ono0611:
## Summary

Addresses the report in #2646 — and full credit to @herove for a precise diagnosis: the `extract_facts` legacy-row guard counted every `row_num IS NULL AND entity_slug IS NOT NULL` row regardless of `expired_at`, so a brain whose legacy rows postdate the v0_32_2 backfill (written by a lingering pre-fence writer) was permanently blocked. `apply-migrations` no-ops (migration already marked applied) and `forget_fact` — the officially sanctioned removal path — only soft-expires, so the guard count could never drain. The issue's SQL-level analysis was exactly right.

We hit the same state on our own production brain (78 legacy rows, all active): `extract_facts skipped: 78 legacy v0.31 facts pending fence backfill` on every nightly dream, with no supported way out.

This PR implements **both** suggested fixes from the issue — they are two halves of one consistency repair: (1) alone cannot drain an active backlog without hand-running `forget_fact` per row, and (2) alone would have re-fenced forgotten (soft-expired) facts, resurrecting them in markdown.

## Part 1 — guard counts only active rows (suggested fix 1)

- The guard COUNT in `src/core/cycle/extract-facts.ts` now carries `AND expired_at IS NULL`, so `forget_fact` visibly drains the pending count toward release.
- **Reconcile preserves soft-expired legacy rows** (codex round-1 P2): with the guard now releasing while expired legacy rows exist, an expired legacy row whose `source_markdown_slug` matches a live page would have entered the reconcile comparison, read as perpetually stale, and been hard-deleted by the wipe (destroying the forget record). Both sides now exclude the `row_num IS NULL AND expired_at IS NOT NULL` shape: `listExistingFactsForPage` filters it, and the wipe passes a new `preserveExpiredLegacy` option to `deleteFactsForPage` (implemented in both engines per the parity rule).
- **Pinned semantics for claim overlap** (codex round-2): if the fence still carries the same `(claim, source)` as an expired legacy row, the reconcile inserts it as a fresh active fence-owned row. Deliberate, documented in code + pinned by a test: legacy DB-only forgets are documented NOT to survive rebuild (`forget.ts` header), and suppressing the insert would create silent fence↔DB divergence — the exact failure mode the guard exists to prevent. Codex ACCEPTed this disposition in round 3.

## Part 2 — active-only backfill + targeted drift-repair lane (suggested fix 2)

The completed-migration ledger marker is **never touched or reset**: "the migration ran" (ledger) and "the DB currently satisfies the post-condition" (drift) are treated as different statements.

- **Repair lane in `apply-migrations`**: when v0.32.2 is ledger-`complete` AND active legacy rows exist, the v0_32_2 orchestrator re-runs as a repair after the normal chain. `--list` / `--dry-run` surface the drift (`Would REPAIR`); a repair that leaves rows behind reports the remainder and resumes on the next run from the drift check.
- **One backfill-target predicate everywhere** (`row_num IS NULL AND entity_slug IS NOT NULL AND expired_at IS NULL`): dry-run counts, the phase B fetch, drift detection, and the phase C remainder report cannot disagree. Soft-expired legacy rows are never fenced back into markdown.
- **Race-safe stamps**: DB stamps carry `row_num IS NULL AND expired_at IS NULL` + `RETURNING id`; a `forget_fact` winning the race between fetch and stamp leaves the row expired-legacy, and the just-committed unbacked fence row is removed again (second atomic rewrite, counted as `raced_forgets`) so the forgotten claim cannot be resurrected by the next cycle (codex P1, tested via an engine proxy injecting the race).
- **Duplicate legacy rows**: fence row_nums are consumed as a once-only queue — the previous `find()`-based lookup assigned the same row_num to every DB row sharing a `(claim, source)` key, violating the partial UNIQUE index on the second stamp and leaving the group permanently half-migrated. True duplicates are soft-expired (never hard-deleted), reported via `expired_duplicates`. Fence row_nums already claimed by a stamped DB row are excluded from the queue (mid-stamp interruption resume).
- **Touched-pages verification**: phase C verifies the pages phase B rewrote this run (+ reports the remaining active-legacy count); pre-existing drift on unrelated pages belongs to doctor and no longer turns legitimate runs partial.
- **Strict, unit-tested repair verdict**: failed/partial status or an unverifiable remainder exit 1 (ledger untouched, progress preserved). `complete` + remaining > 0 stays exit 0 with a loud warning — deliberate deviation from a codex suggestion, because rows can legitimately remain (`skipped_no_local_path` on thin-client sources) and exiting 1 would perma-fail every postinstall `apply-migrations` for those users.
- **PGLite lock-release handoffs awaited**: both `detectV0_32_2Drift` and the orchestrator's `finalizeResult` now await their engine disconnect before the runner opens the next engine (recount), closing a false-"could not be verified" race.

## Testing

- `test/extract-facts-phase.test.ts`: 5 new guard/reconcile regression tests (guard drain, mixed active+expired counting, hybrid-shape preservation incl. under a stale-row wipe, fence-canonical overlap pinning).
- `test/migrations-v0_32_2.test.ts`: active-only predicate (expired rows never fenced; dry-run parity), duplicate handling (one stamped / extras soft-expired / no UNIQUE violation; different-source same-claim not deduped), interrupted-run resume (rename-committed-no-stamps; mid-stamp with duplicate claims), raced-forget cleanup via engine-proxy race injection, drift detection, touched-pages verify scoping.
- `test/apply-migrations-repair-lane.serial.test.ts`: in-process runner lifecycle — detect → repair → recount with the ledger byte-identical throughout, next-run up-to-date exit 0, `--dry-run` prints `Would REPAIR` with zero side effects, no-drift path emits no repair messaging (process.exit stubbed to a sentinel, GBRAIN_HOME sandboxed).
- `test/apply-migrations.test.ts`: pure repair-verdict helper unit tests; the #1062 exit-code source-pattern test updated for the new block shape (invariant unchanged: exit 0).
- **Revert-check performed**: with the `src/` changes reverted, the new tests fail (guard tests, preserve tests, mid-stamp queue test red; migration tests fail at import against the pre-change module); with the fix restored, all pass.
- Affected suites: 87 pass / 0 fail (`extract-facts-phase`, `insert-facts-batch`, `facts-reconcile-protect-cli`, `migrations-v0_32_2`, `apply-migrations`, `apply-migrations-repair-lane.serial`), plus the wider facts/fence suites green. `bun run typecheck` clean.
- Known test limitation (codex round-3 P3, accepted): the lifecycle test runs on the shared test-engine override, which intentionally skips orchestrator-owned disconnects — so the awaited-disconnect fix itself is not regression-guarded by it (a config-backed dual-engine harness would be needed; noted rather than built to keep the PR bounded).

## Multi-model review

Codex (gpt-5.6-sol, high reasoning), 3 rounds on each half: guard half — round-1 P2 fixed, round-2 P2 resolved as pinned deliberate semantics, round 3 ACCEPT with no findings. Repair half — round-1 P1 (raced forget) + P2s fixed, round-2 ACCEPTs + new P2 (disconnect handoff) fixed, round-3 P2 ACCEPT; remaining P3s addressed (console-output assertions) or documented above.

### DIFF
--- a/src/commands/apply-migrations.ts
+++ b/src/commands/apply-migrations.ts
@@ -261,6 +261,56 @@ function printDryRun(plan: Plan, installed: string): void {
   }
 }
 
+/**
+ * #2646: verdict for a finished v0.32.2 drift-repair run. Pure so the
+ * exit-code semantics are unit-testable (codex P2/P3).
+ *
+ * - status failed/partial, or an unverifiable remainder (null after a
+ *   run that just held the DB) → treated as failure (exit 1); the
+ *   ledger stays untouched and progress is preserved for a re-run.
+ * - status complete + remaining 0 → success.
+ * - status complete + remaining > 0 → success WITH a loud warning, not
+ *   an error: rows can legitimately stay behind (skipped_no_local_path
+ *   on thin-client sources). Failing here would make every subsequent
+ *   `apply-migrations --yes` — including postinstall — exit 1 forever
+ *   for those users.
+ */
+function evaluateV0_32_2RepairOutcome(
+  status: 'complete' | 'partial' | 'failed',
+  remaining: number | null,
+): { failed: boolean; message: string } {
+  if (status === 'failed') {
+    return { failed: true, message: 'v0.32.2 drift repair reported status=failed. Fix the reported phase and re-run.' };
+  }
+  if (status === 'partial') {
+    return {
+      failed: true,
+      message: 'v0.32.2 drift repair finished as PARTIAL. Progress is preserved; ' +
+        'resolve the reported phase detail and re-run `gbrain apply-migrations --yes`.',
+    };
+  }
+  if (remaining === null) {
+    return {
+      failed: true,
+      message: 'v0.32.2 drift repair ran, but the remaining backlog could not be verified ' +
+        '(brain unreachable). Re-run `gbrain apply-migrations --yes` to confirm.',
+    };
+  }
+  if (remaining > 0) {
+    return {
+      failed: false,
+      message: `Drift repair finished with ${remaining} active legacy row(s) still pending ` +
+        `(see the fence_facts detail — e.g. skipped_no_local_path). Progress is ` +
+        `preserved; re-run \`gbrain apply-migrations --yes\` after resolving.`,
+    };
+  }
+  return {
+    failed: false,
+    message: 'Drift repair complete: no active legacy rows remain. ' +
+      'The extract_facts guard will release on the next cycle.',
+  };
+}
+
 function orchestratorOptsFrom(cli: ApplyMigrationsArgs): OrchestratorOpts {
   return {
     yes: cli.yes || cli.nonInteractive,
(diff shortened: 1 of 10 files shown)


time-attack:
Thank you for the depth here, and the diagnosis of #2646 is right — master's guard really does count soft-expired legacy rows, so forget_fact can never drain it. We're closing this as scoped, though: the fix for the reported bug is one predicate plus the reconcile-preservation guard, while the other ~900 lines add a new drift-repair lane, a re-runnable migration orchestrator path, a new engine option in both engines, and a race-accounting layer. That's a migration subsystem rather than a bug fix, and at this size it isn't reviewable to the standard a destructive facts path needs.

Masashi-Ono0611:
Understood — a migration subsystem is a fair characterization of the extra ~900 lines, and I agree that's a different bar than a destructive-path bug fix needs. Thanks for confirming the core diagnosis is right; I'll leave a minimal version (the one predicate + the reconcile-preservation guard) as a possible follow-up rather than resubmitting the full lane.

Masashi-Ono0611:
Sent the narrowed cut as its own PR: #3474 — the guard predicate + the reconcile-preservation guard only, per your scoping. The drift-repair lane / orchestrator path / race-accounting are dropped.

Issue #2682, opened 2026-07-09
Image OCR import fail-opens: OCR API errors create filename stubs with errors=0 and no repair path


amtagrwl:
## Summary

When image OCR fails during import (observed with Anthropic API returning insufficient-credit errors), GBrain imports the image as a filename/stub chunk, reports `errors=0`, and advances normally. The brain looks healthy at the sync/import layer even though OCR text is absent for those image assets.

## Observed

Environment: macOS, GBrain v0.42.56.0, Postgres engine, multimodal/image OCR enabled.

- Anthropic OCR call failed with an API error.
- Image import still completed with `errors=0`.
- `content_chunks.chunk_source='image_asset'` rows were created, but `chunk_text` was only a filename/short stub.
- Passive counters showed the damage only after digging: many OCR attempts failed while importer success/error reporting stayed green.
- The nightly fidelity check had to add a custom heuristic: recent image chunks with very short `chunk_text` indicate OCR failure.

## Why this is bad

This is a silent data-quality failure. Operators see a successful import/sync, but the searchable text layer is missing. Because sync has already advanced, there is no obvious built-in re-OCR/retry path for the damaged image chunks.

## Expected

Any OCR failure should be surfaced as one or more of:

1. Per-file import error/non-zero import `errors` count.
2. A durable failure ledger row with enough metadata to retry OCR later.
3. `gbrain doctor` warning when image chunks have filename/stub text despite OCR being enabled.
4. A repair command such as `gbrain ocr --stale` / `gbrain images repair-ocr` that re-runs OCR for image chunks whose text is missing or suspiciously short.

## Suggested detection

A cheap doctor check could flag:

```sql
SELECT count(*) FILTER (WHERE length(chunk_text) < 120), count(*)
FROM content_chunks
WHERE chunk_source = 'image_asset'
  AND created_at > now() - interval '24 hours';
```

If OCR is enabled and most recent image chunks are short stubs, the import path is not healthy even if sync reports zero errors.

time-attack:
[...]

Issue #3066, opened 2026-07-21
test: add CI guard — files calling configureGateway must resetGateway in afterAll (cross-file leak class, ~70 files)


time-attack:
Master shard 6 broke (twice, deterministic) because `test/ai/adaptive-embed-batch.test.ts` left the global gateway configured with a remote provider + fake key past its file boundary; the next embed-triggering file made a live HTTP call. Fixed at the leaker in #3065.

An audit shows ~70 test files call `configureGateway()` and none (except `test/ai/gateway.test.ts`) reset it in an `afterAll`. The legacy-embedding preload only re-applies defaults when the gateway slot is EMPTY, so any of these files can poison downstream files whenever shard composition shifts — i.e., every time a new test file is added. This has now cost multiple master hard-stops.

Proposal: a `scripts/check-*`-style guard (matching the jsonb/progress guard pattern) failing CI when a test file calls `configureGateway(` or `__setEmbedTransportForTests(` without a file-level `afterAll` that calls `resetGateway()`. Alternative considered: a global per-file reset in the preload — rejected for now because bun preload afterAll semantics are per-process, not per-file, and a per-test reset would break files that configure once in beforeAll.

time-attack:
Thanks — the actual leak was fixed at its source in #3065, and what remains here is a proposal for new CI tooling rather than a defect in shipped behavior. We're keeping the issue queue to confirmed user-facing defects, so I'm closing this proposal; the pragmatic mitigation in the meantime is adding `afterAll(resetGateway)` opportunistically when touching one of the ~78 files that call `configureGateway()`. Happy to look at a PR if you ever want to land the guard directly.

PR #3238, opened 2026-07-22
fix(import): surface image-OCR failures instead of importing silent filename stubs (#2682)

Masashi-Ono0611:
## Summary

Addresses the report in #2682: when image OCR fails during import (observed with an Anthropic API insufficient-credit error), the image still imports as a filename/stub chunk and the run reports `errors=0` — a silent data-quality failure. Sync advances normally, so operators only discover it by manually inspecting `chunk_text` lengths, which is exactly what the reporter (thanks, @amtagrwl, for the detailed diagnosis and the doctor-check SQL) had to build a custom heuristic for.

Scope note: of the reporter's 4 expected-behavior items, this PR implements **1** (per-file import warning) and **3** (a `gbrain doctor` check). Items **2** (durable failure ledger) and **4** (`gbrain ocr --stale` repair command) are **not implemented** — flagged as follow-ups below.

## What changed

- **`maybeOcr` (`src/core/import-file.ts`)** now returns a structured outcome (`{ text, failed, reason, message }`) instead of a bare string. `importImageFile` threads `ocrFailed` / `ocrError` onto `ImportResult` when the OCR call itself fails (API error, or no expansion model configured) — as opposed to OCR being opted out, or genuinely returning no text for an image with nothing to extract (see "Judgment calls" below).
- **`runImport` (`src/commands/import.ts`)** counts these into a new `ocrWarnings` field on `RunImportResult`, surfaced in the progress tick line, the final human/`--json` summary, and `sync.ts`'s full-sync debug log.
- **New `gbrain doctor` check `image_ocr_stub_ratio`** (`src/commands/doctor.ts`) looks at `content_chunks` written in the last 24h (`chunk_source = 'image_asset'`, `chunk_text` length < 120), warning when >50% are stubs (vacuous-sample guard at <5 total chunks in the window; gated on `GBRAIN_EMBEDDING_IMAGE_OCR=true`, same env var `maybeOcr` reads).

## Judgment call: warning, not a hard `errors`/`failures` entry

The pre-existing code comment on `maybeOcr` already states OCR is designed to "never fail the import." Hard-gating on OCR failure (adding to `errors`/`failures`, which blocks `sync.last_commit` advancement and checkpoint-clearing) would wedge every subsequent sync — of the *entire* tree, not just images — until the OCR provider is fixed, with no escape hatch other than `--skip-failed`. Since there's no repair/re-OCR command yet (item 4, out of scope here), that would strand operators. I kept OCR failure as a **new, separate `ocrWarnings` counter** — visible, but non-blocking — mirroring the existing `flagged`/`quarantined` precedent for other content-quality issues that stay searchable but get flagged rather than hard-failed.

## Judgment call: empty OCR text is NOT treated as a failure

`generateOcrText`'s own system prompt instructs the model: "If there is no text, return an empty string." An empty result is therefore a legitimate, expected outcome for images with no visible text (e.g. a plain photo), not a failure signal — only an actual thrown error (API error, or no expansion model configured) sets `ocrFailed`. Treating all-empty results as failures would produce false-positive warnings on any photo-heavy import. (Raised and confirmed during multi-round review — see below.)

## Known limitation surfaced during review (now called out in both warning messages)

`importImageFile`'s `content_hash` shortcut skips OCR entirely for byte-unchanged files. That means once an image has landed as a stub, fixing the OCR provider and re-running `gbrain import`/`gbrain sync` will **not** retroactively retry OCR on it (the file is still recognized as unchanged and skipped before OCR is attempted) — `touch`-ing the file doesn't help either, since mtime isn't part of the hash. The only way to get real OCR text on an existing stub today is to delete + re-add the page, or otherwise change the image's bytes. Both new user-facing messages say this explicitly rather than implying a plain re-run fixes it.

## Also confirmed, left out of scope

Incremental sync's rename/add/modify paths call `importFile` (the markdown importer) directly for all files, not `importImageFile` — so `ocrWarnings` only surfaces via `gbrain import` / full sync (`performFullSync`) today, not ordinary incremental `gbrain sync`. This is a pre-existing architectural gap (incremental sync doesn't route images through `importImageFile` at all) unrelated to this diff; flagging as a follow-up rather than expanding scope here.

## Follow-ups (not implemented here)

- Durable failure ledger for OCR failures (reporter's item 2).
- `gbrain ocr --stale` / repair command to re-run OCR for stub chunks (reporter's item 4) — would also be the natural fix for the content_hash-shortcut limitation above.
- Wiring incremental sync's per-file image handling through `importImageFile` (found during review, pre-existing, unrelated to this diff).

## Testing

- `test/import-image-ocr-failure.serial.test.ts` (new, `mock.module`-stubbed `src/core/embedding.ts` + `src/core/ai/gateway.ts` so OCR success/failure/no-key can be forced deterministically without a real API key): `importImageFile` outcome threading (success / API failure / no expansion model / opted-out) + `runImport` aggregation into `ocrWarnings` vs `errors`.
- `test/doctor-image-ocr-stub-ratio.test.ts` (new): the `image_ocr_stub_ratio` check across opt-out / vacuous-sample (<5) / >50% warn / <=50% ok / 24h-window-exclusion cases.
- Revert-check: stashed the `src/` changes and confirmed both new test files fail against the pre-fix code (4/6 and 6/6 tests red respectively), then restored and confirmed green — the tests exercise the actual bug, not a tautology.
- `bun run typecheck` and `bun run verify` (31 checks: privacy, jsonb pattern, progress-to-stdout, test-isolation, wasm-embedded, admin-build, resolver drift, etc.) both green.
- Reviewed over 2 rounds by Codex (`gpt-5.6-sol`, high reasoning effort) — no P1s; addressed 2 of 3 P2s (misleading re-run guidance, reworded twice after a follow-up catch that `touch` doesn't actually help either) and pushed back on the third (empty-OCR-text-as-failure) with reasoning confirmed correct by the reviewer.


Masashi-Ono0611:
CI note: the `test (9)` failures here (`expected 1280 dimensions, not 1536` inside the eval-trajectory CLI tests) are unrelated to this diff — they're the cross-file `configureGateway` leak class tracked in #3066. This PR adds two test files, which reshuffles shard composition; on the merge ref (which now includes the Qwen3/Matryoshka dims work from master), a 1280-dim gateway config leaks from an earlier file in shard 9 into the eval-trajectory suite. I have a systematic fix for #3066 in flight (guard script + a sweep adding `afterAll(resetGateway)` to every configuring test file); I'll retrigger this PR once that lands rather than burning re-runs against a deterministic leak.

Masashi-Ono0611:
Follow-up to the CI note above: the systemic fix is now up as #3244 — the guard identified the exact leakers behind this PR's test(9) failure (`asymmetric-encoding-contract.test.ts` and `doctor-ze-checks.test.ts`, neither resetting the gateway they configure). I'll retrigger this PR once #3244 lands on master.

time-attack:
Thanks for the thorough investigation, and for the honest scope notes. Closing this one: master already surfaces OCR failure through the ocr_attempted/ocr_succeeded/ocr_failed_* counters and the existing `ocr_health` doctor check, so the new `image_ocr_stub_ratio` check largely re-derives a signal doctor already reports. As the PR notes, it also doesn't repair the stub rows (the content_hash shortcut skips re-OCR) and doesn't cover incremental sync — so the net effect is an additional overlapping doctor surface rather than a fix, which isn't surface we're taking on an opt-in path.

Masashi-Ono0611:
Understood — I hadn't connected the existing `ocr_health` check to the same signal, and you're right that this doesn't repair the stub rows anyway. Thanks for the review.

Issue #2570, opened 2026-07-03
Add review-first lifecycle for dream/synthesis-generated pages


TheRealMrSystem:
## What problem does this solve?

Dream/synthesis can write pages autonomously, but generated pages currently enter the normal brain corpus without a native review/adoption boundary. Unlike proposed takes — which have a review-first path before promotion into canonical knowledge — generated pages have no first-class way to distinguish:

- generated but pending
- adopted
- dismissed
- superseded

This creates two failure modes:

1. **Inert generated knowledge.** Useful generated pages exist, but remain linked only from dream summaries or sibling generated pages and are never operationally adopted.
2. **Over-trusted generated knowledge.** Generated pages can be served by retrieval/agents with the same apparent authority as curated human-authored pages.

This depends on durable generated-page provenance first: see #2569. Once GBrain can reliably enumerate generated pages, it can expose their review/adoption state without turning synthesis into manual admin homework.

## Evidence from one production corpus

Measured under `wiki/originals/ideas/`:

- 156 generated/synthesis idea pages inspected
- 32 pages (`21%`) have zero inbound links
- 40 pages (`26%`) are linked only from a dream-cycle summary
- 12 pages (`8%`) are linked only from sibling generated pages in the generated cluster
- 72 pages (`46%`) have some non-generated external inbound link

So **84 / 156 pages (`54%`) have no meaningful adoption signal**.

Existing orphan detection can only see the first group. The summary-only and generated-cluster-only pages are technically linked, but those links are provenance artifacts, not adoption. Connectivity is not the same thing as review/adoption state.

## Related upstream context

- #2418 — review-first take proposals pattern
- #2269 — MCP accept/reject for take proposals
- #160 — trust-gate discussion for untrusted enrichment writes
- #2163 — generated/synthesized pages unreachable by retrieval
- #2415 — generated-page filing/path concerns
- #2283 — dream synthesize backfill/frontmatter/operator-visibility cluster
- #1511 — prior no-write dream review lane was withdrawn/fork-carried, not rejected

This proposal is the pages half of the same review-first doctrine already emerging for takes: generated/proposed knowledge should not silently become stronger canonical knowledge without an explicit adoption boundary.

## Proposed v1: read-only report

Add a read-only command, for example:

```bash
gbrain generated list --since 30d --source-id default --json
```

or similar naming preferred by the maintainer.

v1 should do **zero writes**. It should be a query/report surface only.

### Inputs/signals

Deterministic signals only:

[...]

PR #3239, opened 2026-07-22
feat(generated): read-only v1 report for dream-generated page adoption state (#2570)

Masashi-Ono0611:
Related to #2570 (v1: the zero-write, read-only report — no lifecycle writes).

Implemented against the five-point alignment contract proposed in https://github.com/garrytan/gbrain/issues/2570#issuecomment-5051240641. Each point is adjustable on this PR if you'd like any of them changed:

1. **Scope** — strictly read-only: `gbrain generated list [--json] [--since <spec>] [--source-id <id>] [--limit N]` + a `list_generated_pages` MCP read op. Zero writes; repeated runs are idempotent.
2. **Marker condition** — a page counts as generated when it carries the durable #2569 DB provenance: `dream_generated: true` (canonical jsonb boolean) AND a valid `dream_cycle_date`, excluding `extract_receipt` marker reuse. `--since` compares against `dream_cycle_date` (inclusive UTC date), not `updated_at`.
3. **#2569 best-effort caveat** — the report claims "generated pages with a durable marker", not a complete enumeration. A cheap coverage probe (`coverage.summary_linked_unstamped`) counts live pages a dream-cycle summary links to that lack the marker, so under-enumeration is surfaced instead of silent.
4. **Link classification** — fixed precedence, most-adopted wins: `external > generated_cluster > summary_only > none`, evaluated per DISTINCT inbound origin page (JOIN by page_id, never slug). Self-links and soft-deleted origins/targets are excluded; duplicate edges from one origin count once.
5. **`mentions` edges** — `link_source='mentions'` never promotes to `external`. The JSON carries both the raw inbound edge count and the classification-eligible counts so nothing is hidden. Eligibility of every built-in link_source ('markdown' / 'frontmatter' / 'wikilink-resolved' / 'manual' / NULL-legacy) is documented in the module header.

### Implementation notes

- **Contract-first**: `src/core/operations.ts` gets the read-scope `list_generated_pages` op (source-scoped via `sourceScopeOpts`, so federated grants see the union of allowed sources); the CLI is a CLI_ONLY dispatcher (`src/commands/generated.ts`), same split as `orphans` / `find_orphans`. Thin clients get a refusal hint pointing at the MCP op.
- **Pure classifier**: classification is a pure function (DB rows in → bucket out) in `src/core/generated-report.ts`, unit-tested without an engine.
- **Engine parity**: one raw SQL path through `engine.executeRaw` positional params on both engines. A `targets` CTE pushes the marker/receipt/date-shape/`--since`/source prefilters below the links join so large brains never materialize out-of-report edges; the TS predicate re-validates every row (SQL is a superset filter).
- **Source scope** applies to the target side only; inbound origins count from ANY source, mirroring `findOrphans`' documented deliberate definition — a page in source X linked from source Y is genuinely adopted, and candidates never expose origin slugs, only aggregate counts.
- **JSON contract**: stable `schema_version`, complete bucket counts always included, deterministic candidate ordering (least-adopted first, newest cycle first, then source/slug), capped list with `candidates_truncated` + `total_candidates`.

Example output shape:

```json
{
  "schema_version": 1,
  "report": "generated_pages",
  "window": { "since": "2026-06-23" },
  "source_scope": null,
  "summary": {
    "generated_pages_scanned": 156,
    "buckets": { "external": 72, "generated_cluster": 12, "summary_only": 40, "none": 32 }
  },
  "coverage": { "summary_linked_unstamped": 3, "note": "…" },
  "candidates": [
    {
      "slug": "wiki/originals/ideas/example",
      "source_id": "default",
      "title": "Example generated page",
      "type": "note",
      "dream_cycle_date": "2026-07-01",
      "bucket": "summary_only",
      "inbound": {
        "raw_edges": 3, "self_link_edges": 0, "eligible_edges": 1,
        "eligible_origins": 1, "mentions_only_origins": 2,
        "origin_kinds": { "external": 0, "generated": 0, "summary": 1 }
      }
    }
  ],
  "candidates_truncated": true,
  "total_candidates": 84
}
```

### Testing

- `test/generated-report.test.ts` — 31 pure-classifier unit tests (all four buckets, precedence, mentions exclusion incl. mixed-edge origins, self-links, distinct-origin dedup, marker predicate incl. receipt exclusion, `--since` parsing).
- `test/e2e/generated-report-pglite.test.ts` — 5 e2e tests over the full query path (buckets, soft-delete handling, noncanonical string marker → coverage not scan, cross-source adoption under a scoped report, `--since` window incl. boundary, scalar + federated source scoping, deterministic sort/cap, idempotent re-run).
- `test/e2e/generated-report-postgres.test.ts` — DATABASE_URL-gated Postgres parity suite (same gating pattern as the other postgres e2e suites).
- `bun run typecheck` clean; jsonb-pattern/params, key-files, and progress-to-stdout CI guard scripts pass locally.

Doctor check (`unreviewed_generated_pages`) and any v2 lifecycle verbs are intentionally left out of v1 per the issue's scoping; happy to follow up either behind maintainer preference.


time-attack:
@Masashi-Ono0611 needs better name other than that looks good

Masashi-Ono0611:
Thanks for the review — renamed `gbrain generated list` to `gbrain adoption list` (also updated the file/CLI wiring, thin-client refusal hint, and docs). "adoption" felt like the better fit since that's the actual concept this report tracks (adopted / summary-only / generated-cluster / none), and it leaves room for the v2 lifecycle verbs the issue mentions to live under the same namespace later. Happy to go with something else if you had a different name in mind — just say the word.

On a personal note: I use gbrain every day and burn a fair amount of tokens on it myself, so it's genuinely nice to be able to give a little back. Thanks for building and maintaining this.

time-attack:
Reviewed and declined for now with the observability set. Thank you for building to the #2570 spec — that diligence is noted.

Masashi-Ono0611:
Fair call — and thanks for saying so explicitly rather than closing it silently.

Two things, briefly.

Design steer is welcome any time, and earlier than this. Looking back, I posted the five-point alignment check on #2570 and then started building about half an hour later without waiting for a signal — that was mine to wait for, not yours to catch. Bug fixes I'll keep just sending; for anything that adds product surface I'd rather hear "not this shape" or "not now" before you spend a review on it.

Nothing to revisit today, but if the observability set comes back around I'd be glad to pick this up again — the branch is still on my fork.

PR #3244, opened 2026-07-22
test(ci): guard — configureGateway() callers must resetGateway() in afterAll (#3066)

Masashi-Ono0611:
Addresses the report in #3066.

## Problem

Bun's parallel test runner loads multiple test files into one process per shard. `test/ai/adaptive-embed-batch.test.ts` left the global AI gateway configured with a remote provider + fake key past its own file boundary; the next embed-triggering file in the same shard made a live HTTP call and hard-stopped master shard 6 (twice, deterministically). That specific leaker was fixed in #3065, but the audit in #3066 found this is a class, not a one-off: a large number of test files call `configureGateway()`/`__setEmbedTransportForTests()` without resetting the gateway in a file-level `afterAll`, so whatever a file last configures leaks into whichever file runs next in its shard process — and shard composition shifts every time a test file is added, removed, or reordered.

**This is not hypothetical.** While this PR was in progress, PR #3238's test(9) failed with `expected 1280 dimensions, not 1536` in the `eval-trajectory` CLI test group — a live instance of exactly this leak class, caused by an unrelated PR's diff shifting shard composition so a 1280-dim-configuring file (fixed here) ran ahead of a file expecting the ambient 1536-dim default.

## Fix

**New guard** (`scripts/check-gateway-reset.mjs`, wired in as `scripts/check-gateway-reset.sh` / `check:gateway-reset`): fails CI when a test file calls a gateway test-seam mutator (`configureGateway(`, `__setEmbedTransportForTests(`, `__setChatTransportForTests(`, `__setGenerateTextTransportForTests(`, `__setRerankTransportForTests(` — `resetGateway()` clears all five in one call, so any one left unreset leaks the same way) without a file-wide `afterAll(...)`/`afterEach(...)` that actually clears it.

`check-test-isolation.sh`'s R3/R4 rules already guard the identical "state leaks across a shard-process file boundary" shape for `PGLiteEngine`, using a simple whole-file co-occurrence heuristic (`afterAll(` present + `.disconnect(` present, anywhere in the file, over raw source). Three rounds of self-review (two of them via Codex, `gpt-5.6-sol` at high reasoning effort) against that starting point:

- **Co-occurrence isn't enough**: several files have both an `afterAll(` — for unrelated cleanup like `engine.disconnect()` — **and** a `resetGateway(` call elsewhere (typically in `beforeEach`, which resets state for the *next* test but does nothing once the file's *last* test finishes) — exactly the leak shape, invisible to co-occurrence. So the scanner extracts the balanced-paren span of every candidate teardown-hook call and requires a valid reset *inside* that span.
- **Raw-source regex isn't lexically safe**: a commented-out `// afterAll(() => resetGateway())`, or a string literal containing the text `resetGateway()`, would otherwise be accepted. Fixed by building one same-length lexical mask of the whole file (every comment/string/template/regex literal blanked to spaces — regex-literal detection via the standard division-vs-regex "previous significant token" heuristic, with the token lookback surviving intervening whitespace so `return /foo/` is still recognized as a regex, not division) and running every regex + the span walk against that mask, never raw source.
- **"File-wide" has to be enforced from real structure, not indentation convention**: a hook nested inside one of *several sibling* `describe()`s only tears down that describe's own tests — but a hook nested inside the file's *sole* top-level `describe()` (an extremely common convention, wrapping the whole suite) is just as file-wide as a literal top-level hook. The scanner counts top-level `describe(`s and accepts a one-level-deeper hook only when there's exactly one, rather than assuming either extreme.
- **The reset has to match what was dirtied**: `configureGateway()` sets `_config`, which only `resetGateway()` clears — a bare `__setChatTransportForTests(null)` doesn't touch it, so it can't satisfy a file that also called `configureGateway()`. A file using *only* the individual setters (no `configureGateway()`) may satisfy each one via `resetGateway()` or that *same* setter called with `null` (a fully sufficient, equivalent reset for that one seam).

Scope mirrors `check-test-isolation.sh`: `test/**/*.test.ts`, excluding `*.serial.test.ts` (never shares a shard process with another file) and `test/e2e/**` (confirmed by reading `scripts/run-e2e.sh`: it explicitly spins up a fresh bun process per E2E file, so the leak's precondition — two files sharing a process — is structurally absent there).

**Sweep**: 31 test files (re-audited against the current tree, after #3065's leaker fix and #3231's unrelated shard-timeout change) were missing the reset. Each now resets via `afterAll(() => resetGateway())`, either as a new top-level hook or folded into an existing `afterAll` (`engine.disconnect()`, etc.). Four files (`test/ai/diagnose-embedding-dims.test.ts`, `test/facts-extract-silent-no-op.test.ts`, `test/facts-extract-truncation.test.ts`, `test/search/title-retrieval-arm.test.ts`) already carried a documented "restore the legacy 1536-d pin on exit" pattern for a related dimension-poisoning bug — `resetGateway()` was inserted as the *first* statement of their existing `afterAll` (ahead of the async disconnect and the explicit restore), so the restore's intended final state is unchanged. Widening the trigger to all five setters surfaced 6 more matching files, all already safe via the `afterEach`/setter-null patterns above — no changes needed to those six. One more file, `test/extract-conversation-facts.test.ts`, surfaced once the scanner started enforcing genuine file-wide scope: it has 7 sibling top-level `describe()`s and its only reset was nested in just one of them — safe today only because no later `describe` happens to re-touch the gateway, which is fragile rather than structurally guaranteed. Added a real file-wide `afterAll(() => resetGateway())` rather than leave that fragility in place.

## Verification

- `bun scripts/check-gateway-reset.mjs`: clean (0 violations) on the current tree.
- Guard self-tests (14 fixture files, not committed, covering every distinct behavior the design depends on): a bare violation; the `beforeEach`-only false negative; a commented-out `afterAll`; `resetGateway()` mentioned only in a string; a keyword-prefixed regex literal next to a real reset (must still pass); a *double*-keyword-prefixed regex (`return typeof /resetGateway()/`, must still fail); a file wrapped in one top-level `describe` with its reset one level deep (must pass); a file with several sibling top-level `describe`s and its reset in only one (must fail); a nested (2-levels-deep) `describe`'s reset (must fail); a dead, never-invoked top-level helper function containing a real reset alongside one real `describe` (must fail — the reset isn't lexically inside the describe); a sole top-level `describe` with its reset, plus a *sibling* top-level `test(` outside it (must fail — the hook never covers that sibling); `configureGateway()` paired with only a different setter's null-reset (must fail); the `gateway-reset-guard-ok` opt-out as a real comment (must pass) vs. only inside a string (must fail). All 14 resolve correctly.
- `bun run verify`: 32/32 checks green (includes the new `check:gateway-reset`), re-confirmed after each review round and after rebasing onto master three times during review.
- `bun run typecheck`: clean.
- `bun test` across every `test/*.test.ts` file this PR touches (32 files) + the canonical `test/ai/gateway.test.ts`: 371 pass / 0 fail.
- Full `bun run test` (sharded suite): see CI run on this PR.


Masashi-Ono0611:
CI note: the `osv-scan` failure here is not introduced by this diff — the scan is path-gated on `package.json`/`bun.lock` and this PR's one-line scripts addition triggered it, surfacing three pre-existing lockfile advisories (fast-uri 3.1.3 GHSA-v2hh-gcrm-f6hx High, @hono/node-server 1.19.14 GHSA-frvp-7c67-39w9, body-parser 2.2.2 GHSA-v422-hmwv-36x6). Any PR touching the manifests — and next Monday's scheduled scan — will hit the same three. I'm preparing a separate dependency-bump PR to clear them; the 22 code checks here are all green.

Masashi-Ono0611:
Rebased onto upstream/master (was 205 commits behind, now conflict-free).

One conflict: `test/propose-takes.test.ts` — master had independently added `withEnv`/`emptyHome` imports (unrelated env-isolation test) on the same import line this PR touches; resolved by merging both import additions (took both sides, no logic conflict).

Master also picked up a new test file since this PR was opened — `test/facts-migration-dim.test.ts` — which the check-gateway-reset guard correctly flagged: it has two sibling top-level `describe()` blocks, and its `resetGateway()` call lives only inside the second describe's own `afterAll` (depth 1), so it isn't file-wide per the guard's scope rule. Added a module-scope `afterAll(() => resetGateway())` (same pattern already used in `test/extract-conversation-facts.test.ts` for the identical multi-describe shape).

Verified post-rebase: `bun run check:gateway-reset` clean, `bun run check:test-isolation` clean (1042 files), `bun test` on all 33 touched files → 357 pass / 0 fail, `bun run typecheck` clean.

time-attack:
Verified the leak class is still live on master — gateway-chat.test.ts and friends reset only in beforeEach, so the last test's configuration escapes the file boundary and lands on whichever file shares the shard process. Given this already hard-stopped a master shard with a live HTTP call, the 31-file afterAll sweep is worth taking. One note: the ~500-line bespoke scanner is a lot of tooling surface for a lint; it's in on the strength of the sweep and the existing check-*.mjs precedent, and #3288's smaller R5 rule covers overlapping ground, so please keep the guard boring from here.

Masashi-Ono0611:
Thanks for merging, and noted on keeping the guard boring from here — makes sense given #3288's R5 rule already covers overlapping ground. Appreciate you verifying the leak class is still live on master before taking the scanner.

PR #3246, opened 2026-07-22
fix(deps): clear OSV advisories in bun.lock (fast-uri, body-parser, @hono/node-server)

Masashi-Ono0611:
## Summary

`osv-scan` on #3244 (path-gated on `package.json` changes) surfaced 3 known vulnerabilities already present in `bun.lock` — the weekly scheduled `osv-scan` (Mondays 06:30 UTC) would otherwise fail on the whole repo next week regardless of #3244's own contents. This PR clears them via `package.json` `overrides` + a `bun install` lockfile refresh, following the same-major-bump pattern from #2927.

| advisory | severity | package | before | after |
|---|---|---|---|---|
| [GHSA-v2hh-gcrm-f6hx](https://github.com/advisories/GHSA-v2hh-gcrm-f6hx) | High 7.5 | `fast-uri` | 3.1.3 | 3.1.4 |
| [GHSA-frvp-7c67-39w9](https://github.com/advisories/GHSA-frvp-7c67-39w9) | Medium 5.9 | `@hono/node-server` | 1.19.14 | 2.0.11 |
| [GHSA-v422-hmwv-36x6](https://github.com/advisories/GHSA-v422-hmwv-36x6) | Low 3.7 | `body-parser` | 2.2.2 | 2.3.0 |

## `@hono/node-server` next-major note (needs a maintainer call)

Unlike the other two, this fix ships **only on the 2.x line** — there is no 1.x patch. This deviates from the same-major-only precedent in #2927, so flagging explicitly:

- Upstream's [v2.0.0 release notes](https://github.com/honojs/node-server/releases/tag/v2.0.0) state the bump is perf-only: *"the public API stays the same"*.
- gbrain's own source never imports `hono` or `@hono/node-server` directly. The only consumer is `@modelcontextprotocol/sdk`'s `streamableHttp.js`, and it only imports `getRequestListener` — a Node↔WebStandard request adapter — **not** `serveStatic`, which is the actual vulnerable code path (the advisory is a Windows-only path-traversal in `serve-static` via an encoded backslash, `%5C`). So the vulnerable surface isn't reachable through gbrain's usage regardless of version.
- The full `serve-http`/`oauth`/`connect` suite (253 tests — the closest local proxy for the MCP HTTP transport this package feeds) stays green after the bump, as does `typecheck`.

If a maintainer would rather hold `@hono/node-server` back and accept the Medium finding as a documented false-positive (given the unreachable code path), say so and I'll drop that hunk — the High and Low fixes stand on their own either way.

## Verification

- `git diff bun.lock` confirmed the lockfile diff is solely the resolution consequence of the three `overrides` changes (resolved versions + the new nested transitive deps that `body-parser@2.3.0` itself requires, e.g. `content-type@2.0.0`, `type-is@2.1.0`) — no unrelated dependency drift.
- `osv-scanner scan --lockfile bun.lock`: **3 vulnerabilities → 0**.
- `bun run typecheck`: clean.
- `test/serve-http-bootstrap-token.test.ts`, `serve-http-trust-proxy.test.ts`, `serve-http-cors.test.ts`, `serve-http-health.test.ts`, `oauth.test.ts`, `oauth-authorize-scope-default.test.ts`, `connect.test.ts`: **253/253 pass**.
- `test/voyage-response-cap.test.ts`, `test/ai/voyage-code-3-recipe.test.ts`: **23/23 pass**.
- Full local `bun run test` (4-shard parallel + serial phase) could not be used for verification this session — my local machine had two duplicate full-suite runs contending for memory at once (unrelated to this diff), which produced PGLite WASM "Out of memory" errors and shard-level timeouts across all 4 shards. After cleaning that up, I re-ran the one specific test that had failed under that load (`migrations-v0_19_0.test.ts`) solo — it passed 12/12, confirming the failures were machine-load artifacts, not a regression from this change. Deferring full-suite confirmation to this repo's CI shard environment.
- Reviewed with `codex exec` (`gpt-5.6-sol`, high reasoning effort) against the full diff — no P1/P2 issues raised; it independently re-derived the `getRequestListener`-only usage via its own search of `node_modules`.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -51,8 +51,9 @@
     "@electric-sql/pglite",
   ],
   "overrides": {
-    "@hono/node-server": "^1.19.13",
-    "fast-uri": "^3.1.2",
+    "@hono/node-server": "^2.0.5",
+    "body-parser": "^2.3.0",
+    "fast-uri": "^3.1.4",
     "fast-xml-builder": "^1.1.7",
     "fast-xml-parser": "^5.7.0",
     "form-data": "^4.0.6",
@@ -162,7 +163,7 @@
 
     "@electric-sql/pglite": ["@electric-sql/pglite@0.4.3", "", {}, "sha512-ichuWTgtd4mOM1G4SpyGJa5trT03lWbMypDV0fUXUCXg5hiHqVAz/bZyV68NqmkLB7WcYmj1RMJVSp8HV/v/ZQ=="],
 
-    "@hono/node-server": ["@hono/node-server@1.19.14", "", { "peerDependencies": { "hono": "^4" } }, "sha512-GwtvgtXxnWsucXvbQXkRgqksiH2Qed37H9xHZocE5sA3N8O8O8/8FA3uclQXxXVzc9XBZuEOMK7+r02FmSpHtw=="],
+    "@hono/node-server": ["@hono/node-server@2.0.11", "", { "peerDependencies": { "hono": "^4" } }, "sha512-bjD221KPLoJTWUwso1J6fGKiTXEUFedG/s0visavY4zakFPkeGURMRNly+FhBHs7T8Dz4qHaZIMX9ZoJHSJtKA=="],
 
     "@jsquash/avif": ["@jsquash/avif@2.1.1", "", { "dependencies": { "wasm-feature-detect": "^1.2.11" } }, "sha512-LMRxd0fMgfCLtobDh0/sFYJMMiRJTNYSEEWvRDKXlAeZ08t3gI5V+1thIT0XjXJ+SVG7Zug9B0XPyx0Ti5VRNA=="],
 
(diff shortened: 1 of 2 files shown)


Masashi-Ono0611:
Superseded — the release changelog for v0.42.65.0 (#3346) confirms this exact fix landed independently: "`package.json`/`bun.lock`: raised the existing security-floor overrides so three OSV-flagged transitive packages resolve to patched versions (none on gbrain's own runtime path; MCP/OAuth unit tests pass against the resolved versions)" — matching this PR's three targets (fast-uri 3.1.4, @hono/node-server 2.0.11, body-parser 2.3.0) and the same unreachable-path reasoning for the `@hono/node-server` major bump. Verified master's `package.json` overrides now carry the identical version floors. Closing without merge — no action needed, thanks for picking this up independently.

Issue #3221, opened 2026-07-22
One-token doctor probes falsely fail reasoning models


aaronkhawkins:
The chat reachability probe uses `maxTokens: 1`. Direct reasoning models can consume that budget without emitting usable text and return `length`; the new gateway validation then reports a configuration failure even though the provider/model is reachable.

Suggested fix: use a sufficient diagnostic output budget, or treat a valid length-exhausted response as proof of transport/model reachability while clearly reporting the generation limitation.

- Severity: P2
- Confidence: 100
- Review run: `20260722-072545-ff448d19`
- Finding ID: `src/commands/models.ts:521-one-token-doctor-probe`

time-attack:
Partly accurate but the failure it predicts can't happen on master. The `maxTokens: 1` detail is right (`src/commands/models.ts:517`), however the claimed consequence depends on "the new gateway validation" — and no such validation exists: `chat()` (`src/core/ai/gateway.ts:3442-3459`) returns normally for empty text and for `finishReason: 'length'`, and `probeModel` only reports a failure when `chat()` throws (`models.ts:524-527`). So a length-exhausted reasoning model is currently reported reachable, not as a configuration failure. Closing since there is no defect in current behavior.

Issue #3217, opened 2026-07-22
Zero-token empty completions still succeed


aaronkhawkins:
`validateChatResult()` accepts a contentless completion whenever `output_tokens <= 0`. Providers commonly normalize omitted usage to zero, so an empty non-refusal response can still be treated as success and silently propagate no result.

Suggested fix: validate usable content independently of reported usage; preserve refusal/content-filter exceptions and classify length exhaustion distinctly. Add zero-usage and omitted-usage regression tests.

- Severity: P1
- Confidence: 100
- Review run: `20260722-072545-ff448d19`
- Finding ID: `src/core/ai/gateway.ts:2490-zero-token-empty-completion`

time-attack:
This doesn't reproduce against master: `validateChatResult` does not exist anywhere in the repository (zero hits repo-wide), and `src/core/ai/gateway.ts:2490` is inside the query-expansion prompt builder, not a completion validator. `chat()` (`gateway.ts:3442-3459`) returns text/blocks/stopReason/usage with no usage-gated content validation at all, so there is no `output_tokens <= 0` acceptance path to correct. Closing as not applicable to the current tree — happy to look again if you can point at the symbol on a specific commit.

PR #3249, opened 2026-07-23
fix(gateway): reject contentless chat completions independently of reported usage (#3217)

Masashi-Ono0611:
## Summary

Related to #3217 ("Zero-token empty completions still succeed").

`gateway.chat()` returned `{ text: '', blocks: [] }` for a contentless non-refusal completion, so an empty provider response propagated "no result" to every caller as if it were a valid answer. Providers commonly normalize omitted usage to zero output tokens, so any gate keyed on reported usage would wave these responses through as free successes — the guard added here validates usable **content** (non-whitespace text or tool-call blocks) independently of reported usage, after block normalization in `chat()`.

## Behavior

| Completion shape | Before | After |
|---|---|---|
| Contentless, `finishReason: stop`/`other` (zero or omitted usage) | success with `text: ''` | `AITransientError` (retryable provider blip) |
| Contentless, `finishReason: length` | success with `text: ''` | `AIConfigError` with an actionable fix recipe — the output budget was exhausted before any text was emitted (thinking models spend it on internal reasoning first), which is deterministic for the same call, so a retry is futile |
| Refusal / content-filter stop | returned with its stop reason | unchanged — callers (`toolLoop`) branch on these stop reasons |
| Tool-call-only completion (no text) | success | unchanged — a tool call is usable content |
| Non-empty text with zero-reported usage | success | unchanged — zero usage alone is never treated as failure |

## Budget accounting

The guard throws from inside the `try` block whose catch path records budget via `_extractUsageFromError`. The thrown error carries the usage fields the provider actually reported, so the record charges **actual** tokens instead of the pessimistic `maxOutputTokens` ceiling. When the provider omitted usage entirely, nothing is attached and the pessimistic fallback stays in force (attaching a synthesized `{0, 0}` would record the failed call as free — caught in multi-model review).

## Tests

`test/ai/gateway-empty-completion.test.ts` (new, via the existing `__setGenerateTextTransportForTests` seam, hermetic):

- zero-normalized usage and fully omitted usage both throw (proves the check is usage-independent)
- whitespace-only text is not usable content
- contentless `length` stop is classified distinctly (`AIConfigError` + `fix`)
- refusal / content-filter completions still return, not throw
- non-empty text with zero usage and tool-call-only completions still succeed
- budget: the throw records actual usage (equal spend to a successful control call with identical usage), and omitted usage keeps the pessimistic fallback (spend > 0)

Verified red → green: all guard tests fail on master, pass with the fix. `bun run typecheck` clean; targeted gateway/budget suites (74 tests across 7 files) pass. Full suite left to CI per contribution guidance.

## Scope note

Adjacent to #3221 (one-token doctor probes falsely fail reasoning models): that issue concerns the probe caller in `src/commands/models.ts` (`maxTokens: 1`) and is addressed separately. With this guard in place, a probe whose reasoning model burns the whole output budget surfaces as `AIConfigError` ("output budget exhausted...") rather than an empty success — the probe-side fix for #3221 can treat that error distinctly as proof of reachability.


Masashi-Ono0611:
CI note: the test (9) failure (`expected 1280 dimensions, not 1536` in computeAllSourceMetrics / eval-trajectory tests) is the pre-existing gateway config leak class tracked in #3066 — identical signature to the test (9) failure on #3238, and none of the failing tests are touched by this PR. #3244 adds the afterAll resetGateway guard for the leaking suites; I'll retrigger this PR once that lands on master.

time-attack:
Confirmed on master that a contentless completion returns a success-shaped empty result, which every chat() caller reads as a valid answer — in the facts path that quietly becomes 'no facts', which is the worst way for a memory system to fail. Gating on usable content rather than reported usage is the right choice, and fixing it at the single gateway choke point beats guards in 25 callers. I also checked generateOcrText (whose prompt legitimately returns empty) — it bypasses chat(), so it isn't caught by the new guard. Merging.

Masashi-Ono0611:
Thanks for merging, and for checking `generateOcrText`'s legitimate-empty case wasn't accidentally caught by the new guard — good catch that it bypasses `chat()` entirely so it's unaffected.

PR #3250, opened 2026-07-23
fix(models): doctor chat probe token budget — stop falsely failing reasoning models

Masashi-Ono0611:
## Summary

Addresses the report in #3221: the `gbrain models doctor` chat/expansion reachability probe capped completions at `maxTokens: 1`. Direct reasoning models spend output budget on internal reasoning before emitting any text, so the probe either came back length-exhausted with zero usable text or was rejected outright by providers that require the cap to exceed the model's minimum reasoning spend — and a perfectly reachable model was reported as a configuration failure.

## Changes

- **Sufficient diagnostic budget**: the probe now caps output at a shared `PROBE_MAX_OUTPUT_TOKENS` (64) instead of 1. Providers bill actual tokens generated, not the cap, so the probe stays minimal-cost for non-reasoning models (they answer the `.` prompt in a handful of tokens and stop).
- **Length-exhaustion counts as reachability**: a valid length-exhausted response with no text is treated as proof of transport + auth + model routing, with the generation limitation surfaced in the probe message (human output and `--json`) instead of a bare `reachable`.
- **Extended-thinking budgets respected**: when the user configured `provider_chat_options.<provider>.thinking.budgetTokens`, the probe widens its cap above that budget via a new read-only gateway helper `getConfiguredThinkingBudget` (`@internal`; mirrors `chat()`'s recipe-alias resolution so model-scoped budgets are found through stale aliases too). Anthropic rejects `max_tokens <= thinking.budgetTokens`, so a fixed small cap would recreate the false failure for exactly these models.
- **Docs swept**: stale one-token probe references updated in the `models.ts` header, the doctor help text, `docs/INSTALL.md`, `docs/architecture/KEY_FILES.md`, and `skills/conventions/model-routing.md`. (The 1-token `gateway.embed` reference in `KEY_FILES.md` is the separate embed probe and was intentionally left.)

## Tests

New `test/models-doctor-probe-token-budget.test.ts` (behavioral, via the gateway `__setChatTransportForTests` / `__setGenerateTextTransportForTests` seams — no network, no real API calls). Verified red on the old 1-token behavior and green on the fix. Pins:

1. The probe requests the shared budget constant (floor >= 64, not 1).
2. Length-exhausted empty completion → `ok` with the limitation surfaced.
3. Length-exhausted completion WITH text and normal completions keep the plain `reachable` message.
4. Provider errors (401 etc.) still classify as failures.
5. The cap widens above a configured `thinking.budgetTokens` (provider-scoped, and model-scoped reached through a recipe alias).
6. The non-json render path prints ok-probe caveat messages (source-text pin, same convention as `test/models-doctor-embed.test.ts`).

`bun run typecheck` clean; adjacent suites (`models-doctor-embed`, `models-doctor-reranker`, `ai/gateway-chat`) pass.

## Notes

- Adjacent to #3217 (zero-token empty completions still succeed): this PR deliberately keeps the probe's success criterion as "the provider answered", but the length-exhausted caveat message now makes the "reachable but generation not exercised" state explicit, which should compose with any stricter empty-completion validation from that work.
- Reviewed over 3 rounds with a second model (codex); all P1/P2 findings addressed (human-output caveat rendering, budget floor pin, stale doc sweep, thinking-budget conflict, alias resolution).


time-attack:
Reproduced the premise on master: the probe really does send maxTokens: 1, so `gbrain models doctor` falsely fails reasoning models — a health check that lies about a working configuration is worth fixing. The 64-token cap plus widening above a configured thinking budget is the right minimal shape, and since providers bill generated tokens rather than the cap, the probe stays cheap. Merging this focused version; the same fix appears bundled in #3283 and #3286, which we'll handle separately.

Masashi-Ono0611:
Following up on the close here — the closing comment said "Merging this focused version", but the PR shows closed without a merge commit (`mergedAt: null`), and current master (`52306ed`) still sends `maxTokens: 1` in the models-doctor chat probe (`src/commands/models.ts`), so the probe still falsely fails reasoning models. #3283 and #3286 (mentioned as carrying the same fix) are also closed without a merge commit.

Did the focused version land somewhere I'm missing, or did the merge fall through? Happy to rebase and resubmit the minimal version if that's useful — no rush either way.

Masashi-Ono0611:
Following up on this batch of PRs — you closed them as "merging" on 2026-07-28, but none appear to have landed on master yet (checked independently just now, 16 days later):

- #3249 (gateway: reject contentless chat completions)
- #3378 (think: unwrap markdown-fenced JSON before parsing)
- #2901 (doctor: run reranker_health in --fast and DB-less mode)
- #2912 (dream: honor the #2084 teardown contract)
- #3189 (chronicle: mirror-source backfill)
- #3250 (models: doctor chat probe token budget)
- #2911 (sync: thread AbortSignal through importFile)

For what it's worth, I've been running all seven of these as local patches on top of master in my own production brain since they were opened, and they've held up fine in real day-to-day use — so this isn't just theoretical, it's verified working.

No rush if these are just queued behind other work — flagging in case they slipped through the cracks. Happy to rebase/reopen any of them if that's easier on your end than cherry-picking.

Issue #3224, opened 2026-07-22
'backfill' missing from CLI_ONLY set — command fully implemented but returns 'Unknown command'


WilliamCourterWelch:
**Version:** 0.42.59.0

## Repro

```
$ gbrain backfill --help
Unknown command: backfill
Run gbrain --help for available commands.
```

## Cause

The implementation is all there:

- `src/cli.ts:1996` — `case 'backfill': {`
- `src/cli.ts:1999-2000` — imports and calls `runBackfillCommand` from `./commands/backfill.ts`
- `src/commands/backfill.ts` — handles `kind | list | --help`

But `backfill` is absent from the `CLI_ONLY` set at `src/cli.ts:47`, so dispatch rejects it before the `case` is ever reached. (`edges-backfill` **is** in the set, which is probably why this went unnoticed.)

## Fix

Add `'backfill'` to the `CLI_ONLY` set at `src/cli.ts:47`.

## Suggestion

A startup assertion that every `case` in the command switch has a matching `CLI_ONLY` entry would catch this class of bug — this is a one-word omission that silently disabled a whole command.

PR #3251, opened 2026-07-23
fix(cli): register backfill in CLI_ONLY + guard that every switch case is dispatchable (#3224)

Masashi-Ono0611:
## Summary

`backfill` was fully implemented (`runBackfillCommand`, `kind|list|--help`
in `src/commands/backfill.ts`) but missing from the `CLI_ONLY` set in
`src/cli.ts`, so dispatch rejected the whole command with `Unknown
command: backfill` before ever reaching the handler. `edges-backfill` was
registered, which is why the omission went unnoticed. Thanks
@WilliamCourterWelch for the clear repro and root-cause writeup.

Fixing the one-word omission alone would have introduced a worse, silent
bug: `runBackfillCommand` opens its own PGLite engine end-to-end
(`createEngine` + `connect` + `disconnect`) and takes no engine argument
at all, but the old `case 'backfill':` sat behind `handleCliOnly`'s
shared `const engine = await connectEngine()`. Dispatching there would
open a **second** PGLite connection to the same data dir while the first
is still held — PGLite's single-writer file lock has no same-process
reentrancy check, so every real (non `--help`) run would hang for the
full 30s lock timeout and then fail.

## Changes

- `backfill` added to `CLI_ONLY` (the reported fix).
- `backfill` dispatch moved to an unconditional pre-`connectEngine()`
  branch (same spot as `schema`/`init`/`auth`/`remote`, mirroring the
  existing `--help` pattern already used for `sync`/`capture`/`enrich`)
  instead of leaving it behind the shared connect — avoids the
  double-connect deadlock and makes `--help` reachable from a fresh
  tmpdir with no configured brain.
- The now-unreachable `case 'backfill':` removed from the switch,
  replaced with an explanatory comment.
- `backfill` added to `CLI_ONLY_SELF_HELP` (own detailed help, not the
  generic one-line stub) and to `THIN_CLIENT_REFUSED_COMMANDS` +
  `THIN_CLIENT_REFUSE_HINTS` (it mutates the local engine directly with
  no MCP equivalent — same class as `config`).

## Tests (`test/cli-backfill-dispatch.test.ts`, new)

- Runtime dispatch: `backfill --help` / `-h` reach the real help text
  (not "Unknown command", not the generic stub); `backfill` with no
  brain configured fails on config, not on dispatch.
- A real end-to-end run (`gbrain init --pglite --no-embedding` then
  `gbrain backfill effective_date --dry-run`) against a configured
  PGLite brain, with an explicit `spawnSync` timeout well under the 30s
  lock timeout — proves the double-connect deadlock is actually gone,
  not just the `--help` path.
- The issue's own suggested class-guard: a static structural test
  asserting every `case` in `handleCliOnly`'s switch has a matching
  `CLI_ONLY` entry (scoped to the switch body with comments stripped, so
  prose merely mentioning a case label can't be misread as a real
  dispatch site), with a documented allowlist for 4 pre-existing,
  unrelated dead-code case labels (`search`, `pages`, `notability-eval`,
  `whoknows`) that are out of scope here.
- `test/thin-client-routing-audit.test.ts` extended with a conformance
  test for the new `THIN_CLIENT_REFUSED_COMMANDS` entry.

Verified each guard actually detects the regression it targets by
reverting the corresponding fix locally and confirming the test goes
red (CLI_ONLY removal → "Unknown command" + guard-list mismatch;
pre-connectEngine branch removal → the real 30s PGLite lock timeout
reproduces; THIN_CLIENT_REFUSED_COMMANDS entry removal → conformance
test fails), then restoring and confirming green again.

Addresses the report in #3224.

## Test plan

- [x] `bun test test/cli-backfill-dispatch.test.ts test/thin-client-routing-audit.test.ts` — 31/31 pass
- [x] Related targeted suites (`cli-exit-verdict-pin`, `cli.test`, `cli-help-discoverability`, `cli-dispatch-thin-client`, `backfill-base`, `backfill-concurrency-clamp.serial`) — 95/95 pass, no regressions
- [x] Revert-check: each new guard/test goes red when its corresponding fix is reverted, and green again when restored


Masashi-Ono0611:
CI note: the test (9) failure (`expected 1280 dimensions, not 1536` in eval-trajectory CLI tests) is the pre-existing gateway config leak class tracked in #3066 — same signature as on #3238 and #3249, and none of the failing tests are touched by this PR (which only changes CLI dispatch wiring). #3244 adds the afterAll resetGateway guard for the leaking suites; I'll retrigger once it lands on master.

time-attack:
Closing — both halves are already on master, and the branch can no longer be rebased into something coherent.

Two merged PRs overtook this one:
- **#3529** added `backfill` to `CLI_ONLY`, closing issue #3224.
- **#3558** root-caused the double-connect deadlock you predicted, by passing the shared engine through as `runBackfillCommand(engine, args)`.

I ran **your own e2e deadlock test against current master** and it passes — no PGLite lock hang. So the failure mode you were guarding against is genuinely gone.

The mechanical problem is that this branch is CONFLICTING/DIRTY and its premise no longer holds. The PR states `runBackfillCommand` takes no engine argument; on master it takes `(engine, args)`. The pre-`connectEngine` dispatch hunk here calls it with one argument, so it would not typecheck after a rebase. Resolving that is a rewrite, not a merge.

**One piece here is genuinely worth keeping, and I would like it back as its own PR:** the class guard asserting every switch case is dispatchable. I verified it discriminates — injecting an unregistered case turns it red and names the case. That is a real guard against a whole bug class (finished code that was never wired up, which has happened at least four times in this repo), and it is independent of everything above. Filed on its own against current master it would land quickly.

Closing the branch, not the idea.

Issue #3056, opened 2026-07-21
sync rename path: failed `updateSlug` is swallowed by an empty catch, leaving a duplicate row with no diagnostic


johannesmanske:
### Summary

In the rename path of `sync`, a failed `updateSlug` is caught and discarded with an empty block. The run then continues into `importFile`, which creates a new row for the new path. The old row is left behind, live and orphaned. Nothing is logged, nothing is counted, and the page total does not change, so the result looks like a successful rename.

We hit this twice on a 3-source brain and lost time to it both times.

### The code

`src/commands/sync.ts` on `origin/master` @ `3cc34c92`, lines 2725 to 2734:

```ts
const oldSlug = opts.sourceId
  ? (fromSlugByPath.get(from) ?? resolveSlugForPath(from))
  : await resolveSlugByPathOrSourcePath(engine, from, undefined);
// The new path doesn't yet have a row, so resolve from path only.
const newSlug = resolveSlugForPath(to);
try {
  await engine.updateSlug(oldSlug, newSlug, renameOpts);
} catch {
  // Slug doesn't exist or collision, treat as add
}
```

The comment states the intent, "treat as add". What it does not say is that the pre-existing row is not removed or reconciled, so "treat as add" means "leave a duplicate".

### What we observed

Renaming a file whose stored slug did not match its path-derived slug produced two live rows for one document. The old row carried **0 `content_chunks`**, so it was invisible to vector and keyword search while still occupying its slug. The new row indexed normally. Page counts before and after were identical, which is why a count-based check does not see this.

We could not determine **why** `updateSlug` matched no rows. That is the substance of this report rather than an aside: the error that would have explained it is discarded by the empty `catch`. `resolveSlugsByPaths` had a `source_path` available for the affected pages, so the obvious explanation does not hold, and we have no further signal to go on.

### Why we think it is worth a log line

The two neighbouring loops in the same function were deliberately hardened in the opposite direction. The comment directly below this block reads:

```
// a malformed renamed file is recorded to failedFiles
// and skipped, NOT thrown uncaught. ... A `skipped` result carrying an
// error is also captured so the failure is recorded rather than silently dropped.
```

So the adds and deletes paths already treat "record the failure" as the standard. The `updateSlug` call a few lines above is the one place in the same loop where a failure produces no record at all.

[...]

PR #3252, opened 2026-07-23
fix(sync): surface and reconcile failed updateSlug in the rename path instead of leaving a duplicate row (#3056)

Masashi-Ono0611:
Addresses the report in #3056. Thanks to @johannesmanske for the precise field report and the code archaeology — the observation that the two neighbouring loops were hardened while this one call site produced no record at all is what shaped this patch.

## The problem

In the sync rename loop, `updateSlug` failure was invisible twice over:

1. A **thrown** failure (destination slug occupied, invalid new slug) was swallowed by an empty `catch`.
2. A **zero-row UPDATE** (old slug has no row in the scoped source) never threw at all — `updateSlug` returned `void`, so there was nothing to observe.

Both shapes fell through to `importFile`, which created/updated the row at the new path while the old row stayed behind live: 0 chunks after the next embed pass, slug occupied, page count unchanged — exactly the duplicate described in the report, invisible to every existing signal.

## What this PR does

Following the report's "suggested changes, in order of value":

**1. Surface (log + count).** `updateSlug` now returns the number of rows moved (both engines: `result.count` on postgres.js, `affectedRows` on PGLite — both established patterns in each engine). The rename loop warns on either failure shape with `oldSlug -> newSlug`, the `from -> to` paths, and the error message — the self-diagnosing line the reporter asked for. Fallbacks are counted in a new optional `SyncResult.renameFallbacks` field, a run-end stderr summary line, and the ingest-log summary. Existing `updateSlug` callers (`migrate.ts`) ignore the return value unchanged.

**2. Reconcile the stale row.** When the rename fell back and the destination row demonstrably materialized, the row that still represents the OLD path is located **positively via `source_path = from`** (never by the resolved-slug guess, which can name an unrelated row after a collision) and deleted, scoped to the source. This is the same delete-and-add posture the existing rename-to-unsyncable path (F-C in `test/sync.test.ts`) already takes for a rename whose cheap path can't apply — a rename is semantically old-path-gone + new-path-present, and git has already reported the old path gone. Ordering is loss-safe: the reconcile runs only **after** a successful import, so a failed import never widens into losing the old row too. If no row carries `source_path = from` (the report's unexplained divergent case), nothing is deleted and the warn is the record.

**3. Identity-slug renames resolve instead of blocking.** When the import skips against the very row that represents the old file (`source_path = from`) and the destination slug is free — e.g. a frontmatter-pinned `slug:` whose path-derived form is degenerate, the exact divergence class the report names — the row keeps its identity slug and only its `source_path` moves to the new path (`updateSlug` gained an optional `sourcePath` refresh for file-move renames; pure slug migrations omit it).

**4. A rename that never happened no longer looks like one that did.** If the fallback import did not land at the destination (identity dedup against the old row while a foreign row occupies the destination slug), or the reconcile delete itself fails, the path is recorded in `failedFiles` and skips the checkpoint — the existing failure gate holds `last_commit` and the next run retries the same rename diff. This reuses the sync's established failure posture (ledger, `--skip-failed`, bounded auto-skip) rather than inventing a new one.

Cleaning up duplicates from past incidents is intentionally out of scope, per the issue.

## Safety posture of the reconcile (why deleting is OK here)

- The delete target is identified by `source_path = from` only — a positive claim that the row represents the renamed-away file. No source_path match → no delete.
- It fires only after the destination row provably exists with the file's content (`imported`, or an errorless skip that names the destination slug).
- `deletePage` is the same scoped single-row primitive the sync delete loop already uses for removed files; a rename is git telling us the old path was removed.
- Failure of the delete blocks the bookmark (recorded failure) instead of being checkpointed past.

## Tests (`test/sync-rename-fallback.serial.test.ts`, +2 engine-contract cases)

All written red-first against the pre-fix code and revert-checked (4 fail on the original code):

- engine contract: `updateSlug` returns 1 on a real move, 0 on the silent no-op.
- collision: destination occupied → fallback surfaced (`renameFallbacks`, stderr) + stale old row reconciled; 1 live row after.
- divergent stored slug with no `source_path`: surfaced (warn names both slugs), nothing deleted — honest about what can't be safely located.
- dedup-skip against the old row: the only copy of the content survives and the run **blocks** rather than advancing past a rename that never happened.
- unrelated manual row at the guessed slug: survives (reconcile never deletes by slug guess).
- injected reconcile-delete failure: run blocks, bookmark holds, next run retries to convergence **and closes the failure-ledger row**.
- pinned-slug emoji filename rename: converges (identity slug kept, `source_path` follows the file) instead of blocking.
- happy path: clean `git mv` keeps `page_id`, reports zero fallbacks, and now also refreshes `source_path` (an unchanged-file rename skips reimport, so previously the row kept the old path).

The test file runs under a per-test temp `HOME` (serial-quarantined per `scripts/check-test-isolation.sh`) so blocked runs never touch the operator's real `~/.gbrain/sync-failures.jsonl`.

## Verification

- `bun run typecheck` clean.
- Targeted suites green (313 tests across the touched area: the new file + `sync-rename-batch`, `sync`, `source-id-tx-regression`, `sync-delete-batch`, `sync-reconcile-mass-delete`, `migrate`, `sync-failure-ledger`).
- `scripts/check-test-isolation.sh` green.
- Full suite deferred to CI (per contributor guidance to keep local runs targeted).
- Three rounds of external model review (Codex, high effort); all P1/P2 findings fixed and re-verified — including two data-loss hazards in the first reconcile cut that the final design closes (dedup-skip deleting the only copy; deleting by slug guess).

### DIFF
--- a/src/commands/sync.ts
+++ b/src/commands/sync.ts
@@ -2887,7 +2887,12 @@ async function performSyncInner(engine: BrainEngine, opts: SyncOpts): Promise<Sy
       // NAV-1 TOCTOU: refuse a destination that realpath-resolves outside the
       // repo (committed symlink pointing out).
       const filePath = join(gitContextRoot, to);
-      let importOk = false;
+      // #3056: "destination materialized" = the new path's row demonstrably
+      // exists with the file's content. A bare errorless `skipped` is NOT
+      // enough — identity dedup can skip against the OLD row (result.slug ===
+      // oldSlug), in which case nothing landed at newSlug and reconciling
+      // would delete the only copy.
+      let destMaterialized = false;
       if (existsSync(filePath) && isPathSafe(filePath, gitContextRoot)) {
         try {
           const result = await importFile(engine, filePath, to, { noEmbed, sourceId: opts.sourceId, activePack: syncActivePack });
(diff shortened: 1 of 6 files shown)


johannesmanske:
We hit exactly this class in production: frontmatter-pinned atom slugs whose path-derived form was degenerate (trailing hyphen). We independently arrived at the same core diagnosis while healing it manually: the rename never registers as a slug change because the stripped derivation equals both the old and new path, so updateSlug has nothing to do and the old row stays live. Locating the stale row by source_path rather than the resolved-slug guess is exactly the safe reconcile, and identity-slug-resolve is the right call for the pinned-degenerate case. Fix design matches our findings. Thanks for building it.

Masashi-Ono0611:
Thanks for the independent corroboration — that's genuinely useful. Good to know the degenerate-derivation case (frontmatter-pinned slug whose path-derived form collapses to the same value on both sides) shows up in production beyond my own brain, and that locating the stale row by source_path rather than by the resolved-slug guess matched what you landed on while healing it by hand.

If you have a case where the pinned slug is degenerate in a way identity-slug-resolve does not cover, I'd be glad to hear it — happy to widen the reconcile if there's a shape I missed.

time-attack:
Thank you for this — the diagnosis is correct and well evidenced: master really does swallow both failure shapes (`updateSlug` returns void, so a zero-row UPDATE is silent, and the empty `catch` eats the throw). I'm closing on scope rather than on the finding. The reconcile half introduces a new `deletePage` call plus three-outcome branching and revised checkpoint semantics inside the sync rename loop, which is the highest-blast-radius path in the system, and your own writeup notes the first cut of it carried two data-loss hazards. The surfacing half — returning the moved-row count from both engines, warning with `oldSlug -> newSlug` and both paths, and counting it in `SyncResult` — is about twenty lines and stands entirely on its own; that shape is what I can review with confidence. Cleanup of the stale row is better left to the existing `integrity` / `orphans` tooling than to an inline delete during ingest.

Masashi-Ono0611:
Thanks for confirming the diagnosis and for pointing at the surfacing-only cut as the reviewable shape — that's a better scope than what I sent. I'll send that piece on its own (moved-row count returned from both engines, the `oldSlug -> newSlug` warning, counting it in `SyncResult`) and leave the reconcile/delete half to the `integrity`/`orphans` tooling as you suggested.

Masashi-Ono0611:
Sent the surfacing-only cut as its own PR: #3473. Scoped to exactly the piece you flagged as reviewable (moved-row count + warn + `SyncResult` counter) — no `deletePage`, no reconcile branching.

PR #3254, opened 2026-07-23
fix(schema-pack): parseYamlMini fails loud on block scalars instead of silently truncating (takeover of #1799)

### LINKED ISSUE #1750: parseYamlMini silently drops every key after a block scalar — gbrain-recommended ships with 0 page_types

time-attack:
Supersedes #1799 (fork branch had gone stale vs master; diff re-applied 3-way onto current master, all context clean). Fixes #1750. Credit: @harjothkhara (Co-authored-by preserved on the commit).

## Problem

`parseYamlMini` (`src/core/schema-pack/loader.ts`) documents block scalars (`|`, `>`) as unsupported, but silently mis-parsed them instead of rejecting: a `description: |` line was read as the literal string `"|"`, then the indented block body tripped the `indent > baseIndent` break in `parseMapping`, silently dropping `page_types` and every later top-level key with no error.

This bit the bundled **gbrain-recommended** pack — it uses `description: |`, so `gbrain schema use gbrain-recommended` activated a pack with 0 of its 12 declared page_types.

## Fix

1. `parseScalar` now throws a clear, actionable error when a bare block-scalar indicator (`|`, `>`, incl. chomping/indent forms `|-`, `>+`, `|2`) reaches it — enforcing the parser's documented contract.
2. Inline `gbrain-recommended.yaml`'s description so the shipped pack loads its 12 page_types.

Round-trip safe: `emitYamlScalar` (`mutate.ts`) already emits multi-line strings as JSON-quoted single lines, never block scalars. No false positives: the guard only matches a lone indicator (`^[|>][0-9+-]*$`); real values like `> 50% growth` and quoted `"|"` are unaffected.

## Verification (against current master)

- `bun run typecheck` — 0 errors
- `bun test test/schema-pack-loader.test.ts` — 41 pass / 0 fail
- full schema-pack suite (24 files) — 335 pass / 0 fail

### DIFF
--- a/src/core/schema-pack/base/gbrain-recommended.yaml
+++ b/src/core/schema-pack/base/gbrain-recommended.yaml
@@ -26,12 +26,7 @@ name: gbrain-recommended
 version: 1.0.0
 gbrain_min_version: 0.39.0
 extends: gbrain-base
-description: |
-  The recommended starter pack for an operational personal brain.
-  Based on docs/GBRAIN_RECOMMENDED_SCHEMA.md — Karpathy's LLM wiki
-  pattern extended into a full operational knowledge base. Use this
-  when you want the documented behavior immediately instead of
-  authoring your own pack from scratch.
+description: "The recommended starter pack for an operational personal brain. Based on docs/GBRAIN_RECOMMENDED_SCHEMA.md — Karpathy's LLM wiki pattern extended into a full operational knowledge base. Use this when you want the documented behavior immediately instead of authoring your own pack from scratch."
 
 # Pack-specific page_types (extend gbrain-base's seed types).
 # Each maps a path prefix to a primitive + opt-in flags. Most types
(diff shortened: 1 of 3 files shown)


time-attack:
Closing — this one got overtaken. Master now implements block scalars properly rather than rejecting them: `parseBlockScalar` landed in `src/core/schema-pack/loader.ts` via #3110 (ea6cb025), handling both `|` and `>` from `parseMapping`. `gbrain-recommended.yaml` still uses `description: |` and parses fine now, so the dropped-`page_types` symptom in #1750 is resolved. Merging the fail-loud guard at this point would reject syntax the parser has since learned to read. Genuine thanks to you and @harjothkhara — the diagnosis in #1799 was accurate, it just landed as support rather than rejection.

PR #3256, opened 2026-07-23
fix(onboard): gate extract-ner / timeline recs on whether the action can actually help (takeover of #2362)

time-attack:
Supersedes #2362 (fork branch went stale against fast-moving master; diff re-applied 3-way onto current master, applies cleanly). Full credit to @DarkNightForge — Co-authored-by preserved on the commit.

## Problem

`gbrain onboard --check --explain` surfaces remediations that can never clear, because they're emitted purely on a coverage threshold with no check that the recommended job is capable of moving the metric:

- **`extract-ner`** is recommended whenever `entity_link_coverage` is low — even when the active schema pack declares no `link_types[].inference.regex` rules. `extractNerLinks` then early-returns `pack_unavailable`, creates 0 links, and the rec re-fires forever.
- **`extract-timeline-from-meetings`** is recommended whenever `timeline_coverage` is low — even with no dated meeting pages. The handler skips meetings without `effective_date`, so it creates 0 entries and the rec persists.

## Fix

- **`packSupportsNerInference(pack)`** extracted into `src/core/schema-pack/best-effort.ts` — the exact capability `extractNerLinks` already checks. The handler and the onboard recommender now gate on one definition and can't drift.
- **Timeline gate** — `checkTimelineCoverage` only emits the rec when a dated meeting page exists, matching the handler's own precondition.
- When a gate withholds the rec, the check stays WARN (coverage really is low) and the message explains why there's no auto-fix. Merged coverage branches preserve the original severity/est_seconds tiers via ternaries.

`checks.ts` is intentionally brain-wide (`sourcescope:file-brain-wide` header) — the new COUNT query follows the file's existing pattern.

## Testing

- `test/onboard-phantom-rec-gates.test.ts` (new): pins `packSupportsNerInference` (null / no link_types / inference-without-regex / regex present) plus PGLite engine tests for both gates — withheld-with-reason when capability/content is absent, emitted when a dated meeting exists.
- `bun run typecheck` exit 0.
- `bun test test/onboard-phantom-rec-gates.test.ts test/schema-pack-best-effort.test.ts test/onboard-pack-upgrade-checks.test.ts` — 19 pass / 0 fail.

Versioning/CHANGELOG left to the maintainer's ship.

### DIFF
--- a/src/core/extract-ner.ts
+++ b/src/core/extract-ner.ts
@@ -19,7 +19,7 @@ import type { BrainEngine } from './engine.ts';
 import type { LinkBatchInput } from './engine.ts';
 import { buildGazetteer, findMentionedEntities, type Gazetteer } from './by-mention.ts';
 import { inferLinkTypeFromPack } from './schema-pack/link-inference.ts';
-import { loadActivePackBestEffort } from './schema-pack/best-effort.ts';
+import { loadActivePackBestEffort, packSupportsNerInference } from './schema-pack/best-effort.ts';
 
 export interface ExtractNerOpts {
   /** When true: enumerate but don't write. */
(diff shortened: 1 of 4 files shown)


time-attack:
Merging. Master really does emit both recommendations off a bare coverage threshold while the handlers structurally refuse to do anything — `extractNerLinks` early-returns `pack_unavailable` with no `inference.regex` rules, and the timeline job skips undated meetings — so these recs can never clear. Sharing one `packSupportsNerInference` predicate between the recommender and the handler is exactly the right shape; a duplicated capability check would have drifted within a release. Keeping the check at WARN with a reason instead of silently dropping it is the right ergonomics. Credit noted for @DarkNightForge.

PR #3258, opened 2026-07-23
fix(ai): Ollama compatibility fixes for local-first usage (takeover of #1854)

time-attack:
Supersedes #1854 (fork branch stalled against fast-moving master). Credit to @starm2010 — authorship preserved via Co-authored-by.

## What landed (4 of the original 5 fixes)

1. **`--max-cost` no longer hard-fails on local chat providers**: new `FREE_LOCAL_CHAT_PROVIDERS` set (`ollama`, `llama-server`) in `src/core/budget/budget-tracker.ts` returns zero pricing, sibling to the existing `FREE_LOCAL_EMBED_PROVIDERS` / `FREE_LOCAL_RERANK_PROVIDERS`.
2. **Numeric idea IDs no longer crash brainstorm judges**: `validateIdeaShape` coerces numeric `id` (`5` → `"Idea 05"`) before the string type check (some models, e.g. Kimi K2 via Ollama, emit numeric ids).
3. **Clearer "no LLM available" message**: `buildGracefulMessage` in `src/core/think/index.ts` now mentions `chat_model`, `expansion_model`, or `anthropic_api_key` — that sentinel can fire for any provider, not just Anthropic.
4. **Ollama recipe gains chat + expansion touchpoints** (`llama3.3`, `qwen3`, `deepseek-r1`; `supports_tools: true`, zero cost) so the recipe system can route reasoning calls to Ollama.

## What was dropped as already superseded

The original fix #3 (scope the `NO_ANTHROPIC_API_KEY` warning to `providerId === 'anthropic'`) is obsolete on current master: the #1698 probe work in `runThink` already distinguishes `MODEL_NOT_USABLE` from `NO_ANTHROPIC_API_KEY`, and `probeChatModel` only returns `'unavailable'` for anthropic-without-key — so a non-Anthropic provider can no longer produce a false `NO_ANTHROPIC_API_KEY`.

## Testing

- `bun run typecheck` exit 0
- 217 tests pass across the 16 test files touching the changed areas (budget-tracker, brainstorm, think gateway adapter, recipes regression, model-pricing drift guard, providers)

Related: #1307, #1779

### DIFF
--- a/src/core/ai/recipes/ollama.ts
+++ b/src/core/ai/recipes/ollama.ts
@@ -33,6 +33,20 @@ export const ollama: Recipe = {
       // OLLAMA_NUM_PARALLEL config; no static cap to declare. v0.32 (#779).
       no_batch_cap: true,
     },
+    chat: {
+      models: ['llama3.3', 'qwen3', 'deepseek-r1'],
+      supports_tools: true,
+      supports_subagent_loop: false,
+      supports_prompt_cache: false,
+      cost_per_1m_input_usd: 0,
+      cost_per_1m_output_usd: 0,
+      price_last_verified: '2026-06-02',
+    },
+    expansion: {
+      models: ['llama3.3', 'qwen3', 'deepseek-r1'],
+      cost_per_1m_tokens_usd: 0,
+      price_last_verified: '2026-06-02',
+    },
   },
   setup_hint: 'Install Ollama from https://ollama.ai, then `ollama pull nomic-embed-text` and `ollama serve`.',
 };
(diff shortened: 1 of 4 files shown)


time-attack:
Thanks for picking this up and for carefully dropping the fix that master already superseded — that part of the triage was well done. I'm closing it as submitted: it bundles four unrelated changes, and the recipe hunk that adds a hardcoded `llama3.3`/`qwen3`/`deepseek-r1` chat catalog with a `price_last_verified` date on zero-cost local inference is model-list surface I'd have to keep chasing, not a bug fix. It also lands four behavior changes with no regression test, which this repo pins as a rule. The `FREE_LOCAL_CHAT_PROVIDERS` gate in `budget-tracker.ts` is the real bug here and is about five lines — that hunk on its own, with a test, is the reviewable shape. Credit to @starm2010 for the original report.

PR #3259, opened 2026-07-23
fix(test): size test vectors from the live schema column — shard-composition dim flake (eval-trajectory + source-health)

paul-0320:
## Summary

`test/eval-trajectory.test.ts` and `test/source-health.test.ts` hardcode 1536-wide vectors, but the width their schema actually gets is decided at `initSchema()` time by the **module-global AI gateway**: a neighboring test file in the same shard process that touched the gateway leaves it on the 1280d default, and the next `initSchema()` sizes `facts.embedding` / `content_chunks.embedding` at 1280. The hardcoded 1536 insert then dies with `expected 1280 dimensions, not 1536` (halfvec `CheckExpectedDim`).

Shard composition decides who the neighbors are, and it reshuffles whenever **any** PR adds or removes a test file — so unrelated PRs go red while master stays green. Same failure class as the reporter-count flake fixed in #3231: an absolute expectation on process-global state, triggered by shard composition.

Observed on #2692 `test (9)` ([run](https://github.com/garrytan/gbrain/actions/runs/29972645493/job/89097780848)): 3 eval-trajectory tests + 1 source-health test, identical error, deterministic across in-run retries. The PR touches no progress/facts/source-health code — it merely adds a test file.

## Minimal repro

```
bun test test/gateway-embed-model-override.test.ts test/eval-trajectory.test.ts
# → 3 fail: "expected 1280 dimensions, not 1536"
```

`test/init-embed-check.test.ts` as the neighbor reproduces it identically.

## Change

Read the live column width after `initSchema()` — the idiom `test/embedding-signature-stale.test.ts` already uses — and build vectors at that width. Test-only, no runtime code touched.

Note: whether the gateway's reset path should restore "unconfigured" rather than the 1280 default is a separable design question; this PR just stops two test files from inheriting neighbor state they don't own.

## Verification

- Repro above: 3 fail → 0 fail after this change.
- Full CI-shard-order combo (`gateway-embed-model-override` + `init-embed-check` + `source-health` + `eval-trajectory` in one process): 52 pass / 0 fail.
- Standalone: eval-trajectory 7/7, source-health 28/28 — unchanged.


Masashi-Ono0611:
Independent corroboration: hit the same failure on 4 unrelated PRs of mine (#3238, #3249, #3251, #3382) — the identical 3 eval-trajectory failures on `test (9)` with `expected 1280 dimensions, not 1536`, while master stays green (#3249 also hit the source-health test you mention). None of them touch trajectory/facts/source-health code; they merely add test files, which matches your shard-composition diagnosis exactly. Reading the live column width as this PR does would fix all four.

time-attack:
Closing — the defect this fixes no longer exists on master, verified rather than assumed.

You correctly diagnosed a real shard-composition flake: hardcoded 1536-d vectors in `test/eval-trajectory.test.ts` and `test/source-health.test.ts` broke whenever a neighbouring test left the module-global gateway on the 1280 default. I reproduced that bug and confirmed your fix works on your branch.

But **#3557** (merged 2026-07-29, after you filed this) fixed the root cause at the shared seam instead: `resetGateway()` now restores the preload's 1536 baseline across all 93 reset call sites, with a guard test. Verified by ancestor check that #3557 is in `origin/master` and not in this branch — and all three of your reproduction combinations now pass on clean master (3c61e255).

That is the better fix and it is the one this repo should have: sizing two test files from the live schema helps those two files, while repairing `resetGateway` helps every current and future caller. Your diagnosis was right and it pointed at the same root cause; someone just got to the shared seam first.

Nothing lost by closing — there is no residual hardening value once the global reset is correct. Thanks for the diagnosis; it was accurate.

PR #3261, opened 2026-07-23
fix(mcp): let the search tool scope to a single source (+ search --json)

darw007d:
## Problem

The MCP `search` tool has no per-call source override. Its params lack a `source_id` field and its handler uses plain `sourceScopeOpts(ctx)`, so it can only see the source baked into the `OperationContext` (CLI `--source` / `GBRAIN_SOURCE` / per-token grant). `query`, `search_by_image`, and the code-intel ops (`code_callers`/`code_callees`/`code_blast`/`code_flow`) already support per-call scoping via `resolveRequestedScope`.

A multi-source consumer wiring a second repo into one shared brain (reported by **morzu117/swairm**) hit this: there was no way to scope a single `search` **tool call** to one source — only the CLI's whole-invocation `--source` flag worked.

## Fix

- **`search` op:** add an optional `source_id` param and resolve it through the shared `resolveRequestedScope(ctx, source_id)` — the same trust+grant resolver the other source-scoped reads use. Remote callers now get `permission_denied` for an out-of-grant source instead of a silent cross-source result; `'__all__'` spans every source for trusted local callers and only granted sources for remote callers. Backward-compatible: omitting `source_id` keeps the exact prior behavior (falls back to `ctx.sourceId`).
- **`search --json`:** CLI-only raw-JSON output (mirrors the existing `mode` "local callers only" convention) so scripts/agents piping `gbrain search` don't have to parse the human one-line-per-result format. `params` is now threaded through both `formatResult` call sites (the flag was otherwise inert).

## Tests

- `test/mcp-search-source-scope.test.ts` drives `dispatchToolCall` — the real stdio + HTTP MCP path — across default ctx scope, explicit `source_id` override, remote grant allow/deny (`permission_denied`), and `'__all__'`. Runs against in-memory PGLite, keyword-only (no embedding provider needed).
- `test/cli-format-search-json.test.ts` covers the `--json` formatter, including the empty-set edge (`[]`, not `No results.`) and confirms the default path still renders the human format.

30/30 green across the touched surface (`bun test` on the two new files + `cli-format-volunteer` + `source-scope-resolver`).

---
Contributed from the [swarph](https://swarph.ai) project, which runs gbrain as its semantic-memory layer. Thanks for gbrain! 🧠

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -466,7 +466,7 @@ async function main() {
     // routed path. Date → ISO string; bigint → string (postgres.js shape);
     // Buffer → object. Microsecond-cost; eliminates a whole drift bug class.
     const result = JSON.parse(JSON.stringify(rawResult, bigintToStringReplacer));
-    const output = formatResult(op.name, result);
+    const output = formatResult(op.name, result, params);
     if (output) process.stdout.write(output);
   } catch (e: unknown) {
     // v0.42.20.0 (codex D4): on error, set exitCode + return so the `finally`
@@ -549,7 +549,7 @@ async function runThinClientRouted(
       signal: sigintController.signal,
     });
     const result = unpackToolResult(raw);
-    const output = formatResult(op.name, result);
+    const output = formatResult(op.name, result, params);
     if (output) process.stdout.write(output);
   } catch (e: unknown) {
     if (e instanceof RemoteMcpError) {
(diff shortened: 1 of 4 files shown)


time-attack:
Closing — this is new tool surface rather than a fix for something broken. The per-call scoping you want already exists on master: the `query` op takes source_id and resolves it through the same resolveRequestedScope trust+grant path, and it is the documented full-control variant of the cheaper `search` tool, so a multi-source consumer can scope a single call today. Adding a second parameterized entry point plus a new CLI flag widens the contract we have to keep correct forever for a capability that is already reachable. The work is clean and the tests are well-targeted — it just is not surface we are taking on; thanks for building on gbrain.

darw007d:
Verified and true, ok for close, thank you team.

PR #3271, opened 2026-07-23
v0.42.65.0 fix(mcp): HTTP transport releases PGLite lock on SIGTERM/SIGINT/SIGHUP

jasonm-max:
## Summary

- `serve-http.ts` never installed signal handling, so stopping `gbrain serve --http` (service manager restart, plain `kill`, container stop) skipped `engine.disconnect()` entirely and left the PGLite write lock behind — the next `sources`/`sync` call would hang until `--break-lock`. The stdio transport got this exact fix in v0.31.3 (#801); the HTTP transport was added afterward and never got the equivalent.
- New `installHttpLifecycle()` mirrors the stdio path's pattern: idempotent SIGTERM/SIGINT/SIGHUP handlers, a bounded cleanup deadline so a wedged disconnect still exits, `server.close()` to stop accepting new connections. Exported standalone so it's unit-testable without booting express or a real HTTP listener.
- Also fixes a real race in `logDeliveredReflexPointers`: it deferred registering its fire-and-forget telemetry write through a dynamic `import()`, so a fast-exiting caller could return before the write was tracked for drain-on-exit. Switched to a static import + synchronous registration.
- Bundled: fixes for 5 dev-machine-only test-isolation gaps found while verifying this branch (all reproduce on any developer machine with real local state — a live `ANTHROPIC_API_KEY`, a real `~/.openclaw` workspace, a real `~/.gbrain` config — but pass in a clean CI sandbox). Full detail in the CHANGELOG entry and commit messages.
- Two more pre-existing, unrelated test flakes were found and documented in `TODOS.md` (not fixed here — both are shard-parallelism-only races tied to shared module-level state, out of scope for this branch, real root cause identified for whoever picks them up).

## Test plan

- [x] New `test/serve-http-lifecycle.test.ts` (6 cases: normal shutdown, all 3 signals, idempotency, wedged-disconnect deadline, disconnect()-rejection) — all pass
- [x] Full unit suite: 13940 pass, 2 pre-existing unrelated flakes (documented in TODOS.md, don't reproduce standalone)
- [x] Regression-checked the test-isolation fixes against 45 related test files (449 pass, 0 fail)


time-attack:
Merging. Confirmed on master that serve-http.ts installs no signal handlers at all, so the PGLite write lock survives any SIGTERM/container stop and the next sources/sync hangs until --break-lock — a real, reproducible wedge, and the stdio path has had this fix since v0.31.3. The extracted installHttpLifecycle() with a bounded cleanup deadline is the right shape. Note master is now at 0.42.66.1, so the v0.42.65.0 bump in this branch is stale and needs re-allocating at ship time.

MrZoidberg:
@time-attack why it's closed instead of merge?

PR #3274, opened 2026-07-23
fix(cli): federate unqualified reads and skip cold migrations

mattchronicle:
## Summary

- make trusted local, unqualified read operations span all `federated=true` sources while preserving explicit/env/dotfile/local-path source pins
- use `probeOnly` connections for read operations so cold reads do not run migration/schema startup
- stop routing free-text `gbrain search <query>` into the `search modes|stats|tune` dashboard dispatcher
- add regression coverage for federation, isolated-source exclusion, explicit pins, write scoping, and cold-read routing

## Why

The documented multi-source contract says unqualified reads span federated sources, but the local CLI always populated scalar `ctx.sourceId` from the write resolver, narrowing reads to `default`. Cold reads also paid the migration probe path and `sources list` could hit its 10-second timeout before dispatch.

## Validation

- `bun run typecheck`
- `bun test test/cli-federated-read-context.test.ts`
- `bun test test/cli-cold-read-routing.test.ts`
- `bun run build`
- production smoke: `sources list` fell from timeout (~12s, exit 124) to ~4.3s, and unqualified search returned results from both default and federated meeting sources

`ci:local:diff` reached the Docker runner but the repository script failed before tests with `no such service:` after starting its four Postgres shards; targeted tests and typecheck pass.


time-attack:
Closing on size and blast radius rather than intent. 2088 lines across 65 files and 23 commits, changing what every unqualified local read is scoped to, is not reviewable as one unit — and source isolation is the invariant where a review miss is a cross-source data leak, so this is exactly the change class that needs to arrive in small verifiable pieces. Master has also moved underneath it (the search op already federates via federatedSearchScope per #2561), and your own notes say the full ci:local:diff run never completed. If the cold-read migration-probe fix and the `gbrain search <query>` dispatcher misroute are the parts causing real pain, each of those as its own small PR would be very welcome.

PR #3281, opened 2026-07-23
fix(backlog): frontmatter YAML-comment autofix, all-provider-failed drain retry, --background per-source fan-out

### LINKED ISSUE #3218: Provider failures complete durable atom jobs
When every atom-extraction provider call in a batch fails, per-item errors are converted into a warning result. The drain adapter discards that status and returns `ok/no_progress`, so the Minion worker marks the durable job complete while the backlog remains instead of applying its retry policy.

Suggested fix: propagate an all-provider-failed drain outcome to the Minion handler so the durable job retries. Preserve partial-success behavior when at least one item succeeds.

- Severity: P1
- Confidence: 100
- Finding ID: `src/core/cycle/extract-atoms.ts:575-provider-failure-durable-retry`

### LINKED ISSUE #3225: frontmatter validate --fix corrupts valid frontmatter containing YAML comments (misreads them as markdown headings)
**Version:** 0.42.59.0

## Summary

`frontmatter validate --fix` treats YAML comments inside the frontmatter fence as markdown headings and rewrites **valid** files into broken ones.

## Repro

A file with this (fully valid) shape:

```markdown
---
# a YAML comment inside the frontmatter block
title: "Real Title"
pubDate: 2026-06-29
---
Body...
```

`gbrain frontmatter validate <repo> --fix` reports:

```
[MISSING_CLOSE]:2 Heading at line 2 found inside frontmatter zone (closing --- comes after)
fixed: Inserted closing --- before heading at line 2
```

and rewrites the file to:

```markdown
---
---

# a YAML comment inside the frontmatter block
title: "Real Title"
pubDate: 2026-06-29
---
Body...
```

— an empty frontmatter block, with the **real** frontmatter (`title`, `pubDate`, …) pushed into the document body.

## Impact

Lines starting with `#` inside a frontmatter fence are **YAML comments**, not markdown headings — the original file parses cleanly with any YAML parser. In our case the affected files were Astro content-collection blog posts on a production site: the "fix" stripped `title`/`description`/`pubDate`/`faqs` out of the frontmatter, which would have failed the content-collection schema and broken the live blog had we committed it. Caught only by diffing before commit; restored from git.

(The `--fix` backups under `~/.gbrain/backups/frontmatter/` worked as designed — credit where due.)

## Fix

Before classifying a `#`-prefixed line inside the frontmatter zone as a markdown heading, attempt a real YAML parse of the zone up to the next `---`. If it parses, it's frontmatter with comments — leave it alone. The heading heuristic should be a last resort after YAML parsing fails, not a pre-empt.

### LINKED ISSUE #3227: extract-conversation-facts --background submits a Minion job that dies instantly without --source-id (doctor prescribes exactly that invocation)
## Summary

`gbrain doctor` prescribes this exact fix for a conversation-facts backlog:

```
[WARN] conversation_facts_backlog: 59 eligible pages without extraction.
Fix: gbrain extract-conversation-facts --background --max-cost-usd 5
```

Running that command submits a Minion job that dies immediately:

```
$ gbrain extract-conversation-facts --background --max-cost-usd 5
job_id=18011
$ psql ... "select status, error_text from minion_jobs where id=18011"
dead | extract-conversation-facts Minion job requires data.sourceId
```

The foreground path iterates ALL sources when `--source-id` is omitted (documented in `--help`: "Multi-source: when --source-id is omitted, the command iterates ALL"), but the `--background` Minion handler hard-requires `data.sourceId` and the CLI does not synthesize per-source jobs. So the doctor-prescribed invocation can never work, and a user who follows it concludes the backlog "runs in background" while nothing ever drains.

## Environment

- gbrain 0.42.64.0, engine postgres
- Job result: `status=dead`, `error_text="extract-conversation-facts Minion job requires data.sourceId"`

## Expected

Either the CLI `--background` path fans out one Minion job per eligible source (splitting `--max-cost-usd` across them), or the Minion handler accepts a missing `sourceId` and performs the same all-sources iteration as the foreground path, or the doctor recommendation includes an explicit `--source-id`.

## Related

While reproducing, we also hit the phantom-backlog side of this: pages whose parse yields 0 segments (single-message or assistant-only conversation pages) are skipped WITHOUT a terminal audit row, so `conversation_facts_backlog` counts them forever even though there is nothing to extract; the number can never reach 0. If preferred I can split that into its own issue.

## Workaround

Per-source submission works fine:

```
for SID in <conversation sources>; do
  gbrain extract-conversation-facts --background --source-id "$SID" --max-cost-usd 1
done
```

time-attack:
Backlog fix wave nwf3 — three verified issues.

## Fixes #3225 — autoFixFrontmatter corrupts closed frontmatter containing a YAML comment
Takeover of #3247 (thanks @Masashi-Ono0611, co-authored). The MISSING_CLOSE pass in `autoFixFrontmatter` broke its scan at the first heading-shaped line, so a `#` YAML comment before the real closing `---` was misdetected as a missing close and a spurious `---` split valid frontmatter in two whenever any other fixable error (e.g. SLUG_MISMATCH) triggered the auto-fixer. Now the full zone is scanned for the closer FIRST; the heading heuristic only applies when the close is genuinely absent — mirroring #2153's validator fix in the auto-fixer. 4 regression tests including the issue's exact repro.

## Fixes #3218 — all-provider-failed atom drains complete durable jobs over an untouched backlog
Takeover of #3248 (thanks @Masashi-Ono0611, co-authored). When every `chat()` call in a drain batch threw, `runBatch` mapped only `atoms_extracted`/`duplicates_skipped` (both 0), the loop stopped `no_progress`, and the Minion handler marked the job complete. Now `providerFailure` is derived from the phase result (failures > 0 and zero items processed), surfaced as `status: 'provider_failure'`, and the handler throws so the queue's retry/backoff policy applies; autopilot submits with `max_attempts: 3` instead of 1. Partial-success behavior unchanged.

## Fixes #3227 — doctor-prescribed `extract-conversation-facts --background` always produced an instantly-dead job
The Minion handler is single-source fail-closed (`data.sourceId` required), but the CLI `--background` path submitted one job with `sourceId: undefined` when `--source-id` was omitted — exactly what `gbrain doctor` prescribes. The CLI now mirrors the foreground multi-source loop: it enumerates eligible sources and submits one job per source (job_id is per-call by design), printing every job_id. Explicit `--source-id`, PGLite (inline degrade), and parse errors fall through unchanged. Per-job `--max-cost-usd` matches the documented foreground semantics (per-source budget cap).

## Verification
- `bun test test/brain-writer.test.ts test/extract-atoms-drain.test.ts test/cycle/extract-atoms-synthesize-concepts.test.ts test/autopilot-auto-drain-wiring.test.ts test/extract-conversation-facts.test.ts test/extract-conversation-facts-workers.test.ts` — 123 pass / 0 fail

### DIFF
--- a/src/commands/autopilot.ts
+++ b/src/commands/autopilot.ts
@@ -825,7 +825,10 @@ export async function runAutopilot(engine: BrainEngine, args: string[]) {
                         {
                           queue: 'default',
                           idempotency_key: idemKey,
-                          max_attempts: 1,
+                          // issue #3218: the handler now throws on an
+                          // all-provider-failed batch, so give the queue's
+                          // backoff a chance (was 1 — dead-lettered instantly).
+                          max_attempts: 3,
                           timeout_ms: timeoutMs,
                         },
                         { allowProtectedSubmit: true },
--- a/src/commands/extract-conversation-facts.ts
+++ b/src/commands/extract-conversation-facts.ts
@@ -1318,7 +1318,7 @@ in the facts table (source='${TERMINAL_AUDIT_SOURCE}'). gbrain doctor's
 conversation_facts_backlog check counts pages without this row.
 `;
 
-function buildJobParams(args: string[]): Record<string, unknown> {
+export function buildJobParams(args: string[]): Record<string, unknown> {
   const parsed = parseArgs(args);
   return {
     sourceId: parsed.sourceId,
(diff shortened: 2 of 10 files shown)


time-attack:
Closing as superseded: two of the three fixes here are already on master — autoFixFrontmatter now scans the whole zone for the closing fence before applying the heading heuristic (#3225), and the drain already derives providerFailure, reports status 'provider_failure', and has jobs.ts throw so the retry policy applies (#3218). Only the extract-conversation-facts --background fan-out over eligible sources is still outstanding, and merging a bundle that is two-thirds stale just buys conflicts. If you want that last one, please re-open it as a standalone PR against current master and it will get a clean read. Thanks for the careful write-up.

Issue #1434, opened 2026-05-25
`gbrain sync` without `--source` flag silently fails edits to existing pages


infiniteGameExp:
_Filed via brain-janitor from `Learning and Strategy/Wiki/reference/gbrain-known-bugs.md` (bug #7)._

- **Filed**: 2026-05-24; root cause corrected 2026-05-25 (pending upstream filing — not yet posted)
- **Symptom**: `gbrain sync` after editing an existing wiki page fails for that page with `Warning: skipped <path>: createVersion failed: page "<slug>" (source=default) not found`. The page IS in the index (verifiable via direct `psql` query), but the version-create flow can't find it under the source it's looking in. Sync reports the file as "skipped" rather than "failed"; without `--skip-failed`, sync is blocked entirely. With `--skip-failed`, the failure is acknowledged but the on-disk edit never reaches gbrain's index.
- **Root cause** (corrected 2026-05-25 via direct database inspection): pages were originally ingested under `source_id='studiovault'` (the source registered with `local_path` pointing at the vault). Subsequent `gbrain sync` invocations without `--source studiovault` default to `source_id='default'`. The page lookup is keyed on `(source_id, slug)` — a `default`-scoped sync trying to update a `studiovault`-source page finds no row and reports "not found." The error message is misleading: the page exists in `pages`, just not under the source the sync queried. **This is a config/UX bug, not a logic bug.** The earlier hypothesis (missing join target in `page_versions`) was wrong — direct SQL confirmed page rows exist in `pages`, and `page_versions` legitimately has 0 rows only because the source-mismatched sync has never successfully written a version row.
- **India's encounter**: Hit on 2026-05-24 syncing edited [[phase-6-online-arbitrage-sourcing]] and this very [[gbrain-known-bugs]] page. Both failed identically. The `--skip-failed` bulk acknowledgement reported `532 pre-existing failure(s)`. Initial framing assumed all 532 were createVersion failures from this bug — wrong. Inspection of `~/.gbrain/sync-failures.jsonl` on 2026-05-25 revealed only **5** of those were createVersion failures; **514** were `expected 1536 dimensions, not 1280` errors from the 2026-05-24 PGLite → Postgres migration (bug #5), and the remaining 13 were embedding-key-missing errors. The "532 silent edit failures" narrative in the earlier draft of this entry was not supported by the failure log.
[...]

PR #3282, opened 2026-07-23
fix(engine,resolver,migrate): put-after-soft-delete resurrection, sole-source routing guard, migrate failure accounting

### LINKED ISSUE #3070: sole_non_default tier silently hijacks writes when 'default' holds an established corpus
## Environment

- gbrain 0.42.62.0 (binary from releases, Ubuntu 24.04, Postgres 16); the code path is unchanged on current main (0.42.64.0).
- Multi-source brain: `default` holds a ~1050-page curated KB; one additional source (`tg-*`, Telegram-topic sync) registered with a `local_path`; no `sources.default` config set.

## What happened

A full KB re-ingest ran `gbrain put <slug>` for 439 pages without `--source` (the CLI help for `put` doesn't document a source flag). Every page was silently filed into the `tg-*` source instead of `default`, so source-scoped retrieval over `default` kept serving the stale pages. No warning was printed in our pipeline (stderr was captured away, and `GBRAIN_NO_SOLE_NON_DEFAULT_NUDGE` aside, a nudge is easy to miss in batch runs).

Root cause is tier 5.5 (`pickSoleNonDefaultSource`, v0.41.13 / #1434): with no brain default set and exactly one non-default source registered, EVERY command without `--source` auto-routes there — even when `default` is an established, actively-used corpus. The tier's stated charter was rescuing brains whose `default` held 0 pages ('532 silent edit failures'), but it fires far beyond that.

## Repro

1. Brain with pages in `default` (any N > 0).
2. Register one extra source with a local_path: `gbrain sources add side --path /tmp/side`.
3. `echo hi | gbrain put test:page` → row lands in `source_id='side'`, not `default`.

## Suggested fix

Fire the tier only when `default` is actually empty (its original charter). Patch we're running with, tests included (established corpus → `seed_default`; empty default → tier preserved):

```diff
diff --git a/src/core/source-resolver.ts b/src/core/source-resolver.ts
index 8b9f3ba..c42d587 100644
--- a/src/core/source-resolver.ts
+++ b/src/core/source-resolver.ts
@@ -185,8 +185,24 @@ async function pickSoleNonDefaultSource(engine: BrainEngine): Promise<string | n
       `SELECT id FROM sources WHERE local_path IS NOT NULL AND id != 'default'`,
     );
   }
-  if (rows.length === 1) return rows[0].id;
-  return null;
+  if (rows.length !== 1) return null;
+  // Guard: the tier exists to rescue brains whose 'default' source holds
+  // 0 pages (#1434's "532 silent edit failures"). When 'default' is an
+  // ESTABLISHED corpus, auto-routing hijacks every bare `put`/`capture`
+  // into the sole side-source — observed in the wild: a 1045-page default
+  // KB re-ingest silently filed 439 pages into a Telegram-sync source.
+  // Fire only when default is actually empty.
+  try {
+    const defaultPages = await engine.executeRaw<{ one: number }>(
+      `SELECT 1 AS one FROM pages WHERE source_id = 'default' AND deleted_at IS NULL LIMIT 1`,
+    );
+    if (defaultPages.length > 0) return null;
+  } catch {
+    // pages.deleted_at predates this guard on every supported schema; a
+    // failure here means an exotic/legacy brain — keep the pre-guard
+    // behavior rather than breaking resolution outright.
+  }
+  return rows[0].id;
 }
 
 /**
```

Happy to turn this into a PR if you want it.

### LINKED ISSUE #3071: putPage to a soft-deleted (source_id, slug) updates content but leaves the page dead
## Environment

- gbrain 0.42.62.0 (Postgres 16); same code on current main (0.42.64.0). Applies to both engines (postgres-engine.ts and pglite-engine.ts share the upsert shape).

## What happened

Live sequence on a KB slug:

1. `gbrain delete <slug>` (soft delete — `deleted_at` set),
2. `gbrain put <slug>` with fresh content.

The upsert hits `ON CONFLICT (source_id, slug) DO UPDATE`, which rewrites `compiled_truth` / `content_hash` / `updated_at = now()` — but never touches `deleted_at`. Result: a row whose `updated_at` is seconds newer than its `deleted_at`, invisible to every `deleted_at IS NULL` read (get_page, query, sync dedup index). The write looks successful, the fresh content is dead, and retrieval keeps serving whatever stale sibling still matches.

## Repro

```
echo one | gbrain put t:x
gbrain delete t:x
echo two | gbrain put t:x
gbrain get t:x        # -> not found, while pages has the row with content 'two'
```

## Suggested fix

`put` is by definition a write of live content — the upsert should resurrect the row. Patch we're running with (both engines; contract test added next to the soft-delete suite — it fails pre-fix):

```diff
diff --git a/src/core/pglite-engine.ts b/src/core/pglite-engine.ts
index 54efd08..695be15 100644
--- a/src/core/pglite-engine.ts
+++ b/src/core/pglite-engine.ts
@@ -1046,6 +1046,10 @@ export class PGLiteEngine implements BrainEngine {
          frontmatter = EXCLUDED.frontmatter,
          content_hash = EXCLUDED.content_hash,
          updated_at = now(),
+         -- A put IS a write of live content: resurrect a soft-deleted row.
+         -- Mirrors postgres-engine.ts (pre-fix the row stayed invisible to
+         -- every deleted_at IS NULL read while looking freshly updated).
+         deleted_at = NULL,
          effective_date        = COALESCE(EXCLUDED.effective_date,        pages.effective_date),
          effective_date_source = COALESCE(EXCLUDED.effective_date_source, pages.effective_date_source),
          import_filename       = COALESCE(EXCLUDED.import_filename,       pages.import_filename),
diff --git a/src/core/postgres-engine.ts b/src/core/postgres-engine.ts
index b42f2a9..8e89da9 100644
--- a/src/core/postgres-engine.ts
+++ b/src/core/postgres-engine.ts
@@ -1108,6 +1108,11 @@ export class PostgresEngine implements BrainEngine {
         frontmatter = EXCLUDED.frontmatter,
         content_hash = EXCLUDED.content_hash,
         updated_at = now(),
+        -- A put IS a write of live content: resurrect a soft-deleted row.
+        -- Pre-fix, putting to a soft-deleted (source_id, slug) updated the
+        -- content but left deleted_at set, so the page stayed invisible to
+        -- every deleted_at IS NULL read while looking freshly updated.
+        deleted_at = NULL,
         effective_date        = COALESCE(EXCLUDED.effective_date,        pages.effective_date),
         effective_date_source = COALESCE(EXCLUDED.effective_date_source, pages.effective_date_source),
         import_filename       = COALESCE(EXCLUDED.import_filename,       pages.import_filename),
```

Happy to turn this into a PR if you want it.

### LINKED ISSUE #3194: migrate silently drops pages with NULL optional columns (reports N/N success)
### Environment
- gbrain 0.42.62.0
- Source engine: `pglite` (WASM Postgres)
- Target engine: native `postgres` 18.4 + `pgvector` 0.8.5
- Command: `gbrain migrate --to supabase --url postgresql://gbrain:***@127.0.0.1:5433/gbrain`

### Symptom
`gbrain migrate` printed the page-copy progress counter reaching `136/136 pages copied`, but **10 pages were silently absent from the target database** after the run. The migration counter said success; the data was incomplete.

The failure only surfaced later when the link-copy phase aborted:
```
addLink failed: page "people/hayden-bohlen" (source=default) or "projects/hermes/overview" (source=default) not found
```
The target was missing `projects/hermes/*` and `hermes-*` pages (and two `test-*` artifacts).

### Root cause
The dropped pages each contain a column value that is `NULL` in PGLite but rejected by the Postgres `putPage` write path. Running `putPage` directly on the target produced:
```
UNDEFINED_VALUE: Undefined values are not allowed
    at putPage (src/core/postgres-engine.ts:2249:7)
```
PGLite tolerates the NULL; the Postgres/Postgres.js binding throws `UNDEFINED_VALUE` (it does not accept JS `undefined`/untyped nulls for the bound parameter). The **per-page copy loop in `migrate-engine.ts` caught the error per-page, skipped the page, and continued** — so the progress counter still advanced to the full count and the run "succeeded" without those pages.

This means any page with a nullable column that happens to be NULL on the source is silently lost on migration. For a brain with many pages, that's a quiet data-loss bug.

### Reproduction
1. Have a `pglite` brain with ≥1 page where an optional column (`effective_date`, `import_filename`, `links_extracted_at`, `source_path`, etc.) is `NULL`.
2. `gbrain migrate --to supabase --url postgresql://...`
3. Observe `N/N pages copied` with no error.
4. `SELECT count(*) FROM pages` on target < source; missing pages are exactly those with NULL optional columns.

### Detect / verify
After migration, diff page counts:
```sql
-- source (pglite) vs target (postgres)
SELECT count(*) FROM pages;  -- source
SELECT count(*) FROM pages;  -- target  (will be lower)
```
Missing slugs can be found by intersecting source slugs with target slugs.

### Workaround used (data recovered)
For each missing slug, copy the row directly with explicit NULL handling:
```ts
const row = await src.executeRaw('SELECT * FROM pages WHERE slug=$1', [slug]);
await tgt.executeRaw(
  `INSERT INTO pages (source_id, slug, type, ...) VALUES ($1,$2,$3,...)`,
  [row.source_id, row.slug, row.type, /* ...explicitly pass each col, nulls allowed... */]);
// then content_chunks, page_versions, and remap links by slug
```
After manual recovery: 146 source pages → 144 target pages (2 intentional `test-*` exclusions), 326 links remapped 1:1.

### Suggested fix
In `migrate-engine.ts` page-copy loop: (a) do not treat a per-page write error as "success" — track and report skipped/failed page slugs in the final summary, and ideally fail the migration if any page is not copied; (b) ensure `putPage`/the insert path coerces `null` → SQL `NULL` for nullable columns instead of letting Postgres.js throw `UNDEFINED_VALUE`.

### Severity
Data loss on migration (silent). Would be catastrophic for users who trust the "N/N copied" counter and delete their old brain.

/cc @garrytan

time-attack:
Backlog fix wave nwf6 — three adversarially verified issues.

Fixes #3071 — `putPage`'s `ON CONFLICT (source_id, slug) DO UPDATE` never cleared `deleted_at`, so put → soft-delete → put left the new content invisible to every `deleted_at IS NULL` read (`get_page`, `query`, `list`). Now `deleted_at = NULL` on update in BOTH engines (engine-parity invariant). Regression test in `test/pages-soft-delete.test.ts`.

Fixes #3070 — the `sole_non_default` resolver tier (#1434) auto-routed bare `put`/`capture` into the side source even when `default` held an established corpus. `pickSoleNonDefaultSource` now probes for live pages in `default` and falls through to the `default` terminal when any exist; fails open on probe error so legacy schemas keep prior behavior. Tests: established default → resolves `default`; empty default → tier still fires; probe failure → fails open.

Fixes #3194 — `gbrain migrate` could report success while pages were missing:
- per-page copy failures are collected (and kept OUT of the resume manifest so a re-run retries them) instead of being invisible or aborting the run
- the link phase collects per-link failures instead of aborting mid-phase on the first `addLink` error
- per-source live page counts are verified source-vs-target before the config switch; any failure or mismatch prints a failed-pages summary, leaves config + manifest intact for resume, and exits non-zero
- soft-deleted pages are documented as deliberately excluded (live counts are the contract), with a note printed when any are left behind

Tests: `bun test test/pages-soft-delete.test.ts test/source-resolver-sole-non-default.test.ts test/migrate-engine-copy-failures.test.ts` plus 12 adjacent suites (358 tests) — all green; `bun run typecheck` exit 0. No VERSION/CHANGELOG changes (release bookkeeping happens at ship time).

### DIFF
--- a/src/commands/migrate-engine.ts
+++ b/src/commands/migrate-engine.ts
@@ -16,6 +16,7 @@ import { createHash } from 'crypto';
 import { resolve } from 'path';
 import { createProgress } from '../core/progress.ts';
 import { getCliOptions, cliOptsToProgressOptions } from '../core/cli-options.ts';
+import { setCliExitVerdict } from '../core/cli-force-exit.ts';
 
 interface MigrateOpts {
   targetEngine: 'postgres' | 'pglite';
@@ -441,7 +442,11 @@ export async function runMigrateEngine(sourceEngine: BrainEngine, args: string[]
     }
     console.error(`\nMigration INCOMPLETE: ${migrated} pages transferred, ${failedPages.length} failed.`);
     console.error('Config NOT switched; resume manifest kept. Fix the errors above and re-run the same migrate command to retry.');
-    process.exitCode = 1;
+    // #3194: route through the owned verdict channel — the one-shot exit seam
+    // (cli.ts flushThenExit(currentExitCode())) ignores raw process.exitCode
+    // writes (PGLite WASM can scribble over it), so a bare assignment here
+    // would let an INCOMPLETE migration exit 0.
+    setCliExitVerdict(1);
     await targetEngine.disconnect();
     return;
   }
--- a/src/commands/migrate-engine.ts
+++ b/src/commands/migrate-engine.ts
@@ -10,7 +10,7 @@
 import { createEngine } from '../core/engine-factory.ts';
 import { loadConfig, saveConfig, toEngineConfig, gbrainPath, effectiveEnvDatabaseUrl, type GBrainConfig } from '../core/config.ts';
 import type { BrainEngine } from '../core/engine.ts';
-import type { EngineConfig } from '../core/types.ts';
+import type { EngineConfig, Page } from '../core/types.ts';
 import { writeFileSync, readFileSync, existsSync, unlinkSync } from 'fs';
 import { createHash } from 'crypto';
 import { resolve } from 'path';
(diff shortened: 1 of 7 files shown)


time-attack:
Closing as superseded: the put-after-soft-delete resurrection is already fixed in both engines on master (`deleted_at = NULL` is in the ON CONFLICT DO UPDATE in pglite-engine.ts and postgres-engine.ts), and migrate-engine.ts already collects per-page copy failures, keeps them out of the resume manifest, and verifies live counts before switching config. That leaves only the sole_non_default probe change, which alters write-routing semantics and should be reviewed on its own rather than inside a two-thirds-stale bundle. Please re-file that one standalone against current master.

PR #3283, opened 2026-07-23
fix(search,migrate,models): backlog fix wave nwf1 — knobs-hash source-boost fold, BYPASSRLS warn-and-proceed, doctor probe token budget

time-attack:
Three adversarially verified community PRs repaired, rebased onto master, and bundled as one fix wave.

## Takeover of #3213 — fold resolved source-boost map into query-cache knobs hash
The PR was correct (KNOBS_HASH_VERSION 12→13, canonicalized `sb=` part mirroring #2825's `hx=` pattern) but left two version pins at 12, which would have failed CI. Repaired: `test/cross-modal-phase1.test.ts` and `test/search-alias-resolved-boost.test.ts` pins updated to 13 with the bump rationale.

## Takeover of #3212 — tolerate non-superuser roles lacking BYPASSRLS
The v24/v29/v31/v35 RLS-backfill migrations `RAISE EXCEPTION` when the role lacks BYPASSRLS, permanently wedging the upgrade chain on managed Postgres (the app role owns the tables but can never hold BYPASSRLS; fresh installs are unaffected). Now warn-and-PROCEED — still fail-closed against the original "silent version bump without RLS" hazard, since `ENABLE ROW LEVEL SECURITY` requires ownership and a non-owner role aborts the DO block with `schema_version` unbumped. Repaired over the original PR: (1) its SQL comment contained backticks inside the TS template literal, so it did not typecheck; (2) the three pinned guards in `test/migrate.test.ts` asserting the old RAISE EXCEPTION contract now pin the new contract, including that the ALTERs sit OUTSIDE the guard branch (warn-and-proceed, never warn-and-skip).

## Takeover of #3250 — models doctor probe token budget (#3221)
`maxTokens: 1` falsely fails reasoning models. `PROBE_MAX_OUTPUT_TOKENS=64`, widened above any configured extended-thinking budget; a length-exhausted empty-text response counts as reachable with the limitation surfaced. Repaired: composes with the gateway contentless-length guard proposed in #3249 (#3217) — when `chat()` throws the `output budget exhausted` `AIConfigError` before the probe sees a `ChatResult`, `probeModel`'s catch recognizes it and reports reachable-with-caveat (the throw itself proves transport + auth + model routing); every other `AIConfigError` still classifies as a failure. Two new tests pin the composition.

References (not closed by this PR): #3213, #3212, #3250, #3221, #3217, #3249.

## Verification
- `bun test` on all six touched test files — 314 pass / 0 fail (plus 438 pass across adjacent models/gateway/llms suites)
- `scripts/check-key-files-current-state.sh` — ok; `bun run build:llms` regenerated (no bundle drift)

Original authors credited via Co-authored-by on each commit: @SVZTeam, @alexey-metaengage, @Masashi-Ono0611.


time-attack:
Merging. I re-verified all three against current master: the BYPASSRLS RAISE EXCEPTION is still in migrate.ts (v24/v29 both), models.ts still probes at maxTokens: 1, and KNOBS_HASH_VERSION is still 12 with no source-boost part in the hash — so all three defects are live, and the upgrade-chain wedge on managed Postgres is the user-blocking one. Thanks also for repairing the version pins and the backtick-in-template-literal typecheck break from the original PRs rather than merging them broken. Please confirm the rebase is clean before landing, since master has moved several releases since this was opened.

Issue #3055, opened 2026-07-21
`gbrain list` silently caps at 100 rows regardless of `--limit` (two downstream failures observed)


johannesmanske:
### Summary

`gbrain list` accepts `--limit N` but never returns more than 100 rows. There is no warning and no truncation marker. Anything built on `list` as a counting or enumeration primitive is silently wrong once a brain passes 100 pages of the queried kind.

We hit this twice in production on a 550-page brain, in two unrelated places. Both failures were silent for days.

### Reproduction

```
$ gbrain list --type atom --source egora --limit 100   | wc -l
100
$ gbrain list --type atom --source egora --limit 5000  | wc -l
100
$ gbrain list --type atom --source egora --limit 100 --updated-after 2026-01-01 | wc -l
100
```

Actual count for that source and type at the time: 124.

### Mechanism

The clamp is explicit, in `src/core/operations.ts` (line 1417 on `origin/master` @ `3cc34c92`, line 1370 in the 0.42.53.0 tree):

```ts
limit: clampSearchLimit(p.limit as number | undefined, 50, 100),
```

The third argument is a hard maximum. The engine below it honours whatever it is handed, `pglite-engine.ts:1171` reads `filters?.limit || 100`, where 100 is only a default. So the ceiling is imposed at the operation layer, and the `list_pages` param description one line above it reads:

```ts
limit: { type: 'number', description: 'Max results (default 50)' },
```

It documents the default and not the ceiling, which is why `--limit 5000` looks like it should work.

### Failure 1: a loop that could never terminate correctly

Our nightly job runs `dream --phase extract_atoms` in a loop until the atom count stops rising, because the phase's internal budget cap (`DEFAULT_BUDGET_USD`) makes it skip the remaining corpus once reached. The loop used `gbrain list --type atom --limit 5000 | grep -c` as its counter.

Once the source passed 100 atoms, the counter was pinned at 100. Before and after were always equal, so the loop broke after the first pass, every night. From the log:

```
[cycle.extract_atoms] 1 2 atoms / 39 skipped
extract_atoms egora Pass 1: Atome 100 -> 100
```

**The same log line reports two new atoms and an unchanged count.** The loop had been non-functional for an unknown number of nights before we noticed.

### Failure 2: soft-deleted pages became unenumerable

`--include-deleted` appeared unreliable: the same query form returned 8 soft-deleted pages in one session and 0 in another for a page we had just soft-deleted and verified via `get`.

[...]

Issue #2828, opened 2026-07-14
sync --full reconcile deletes every subdirectory page on Windows (source_path separator mismatch) — total source data loss


1alessio:
## Summary

On Windows, `gbrain sync --full` deletes **every page whose `source_path` contains a directory separator** when those pages were written with forward-slash paths (e.g. by capture/write-through operations). In our case the full-sync reconcile hard-deleted 1,215 of 1,224 pages in a capture-origin source. Root-level pages (`readme`, `schema`, etc. — no separator in the path) survive, which is the diagnostic fingerprint.

Version: 0.42.59.0, Postgres engine, Windows 11 host (Postgres in WSL2, CLI via bun on Windows).

## Root cause

`src/commands/sync.ts`, full-sync reconcile (~line 3104 in 0.42.59):

```ts
const current = new Set(
  collectSyncableFiles(repoPath, { strategy: opts.strategy ?? 'markdown' })
    .map(abs => relative(repoPath, abs)),
);
```

On Windows, Node's `relative()` returns **backslash** paths (`processes\2026\07\file.md`). But `pages.source_path` can be stored with **forward slashes** — in our brain, every page created by the capture op had `processes/2026/07/file.md`. The membership test

```ts
&& !current.has(r.source_path)
```

then fails for every subdirectory page, so the reconcile classifies the entire corpus as "backing file removed" and deletes it, while the import phase in the same run happily reports those very files as `skipped (unchanged)` (its comparison normalizes differently).

The comment above the reconcile explicitly worries about this class ("otherwise every page looks stale and the reconcile would wrongly delete live pages") — the separator dimension is the missed case.

## Reproduction

1. Windows host, Postgres engine.
2. Create pages in a source via a write path that stores forward-slash `source_path` values under subdirectories (capture ops do this), with matching files in the repo working tree.
3. `gbrain sync --source <id> --repo <path> --full --no-pull --yes`
4. Observe: `Reconciled N stale page(s) whose source file was removed.` where N ≈ every subdirectory page; import phase in the same run reported the same files as unchanged.

## Suggested fix

Normalize separators on **both** sides of the membership test (e.g. `.replace(/\\/g, '/')`), the same way other path-handling in the codebase does. A belt-and-braces addition: refuse to reconcile-delete more than some fraction (e.g. >50%) of a source's pages in one run without an explicit flag — a mass delete of an entire live source is almost always a comparison bug, and this failure mode is total data loss for users without backups.

## Impact

[...]

PR #3284, opened 2026-07-23
fix(sync): separator-agnostic path containment — unblock sync on Windows

### LINKED ISSUE #3057: Windows: sync fully blocked isPathSafe rejects every file (hardcoded '/' vs realpathSync backslashes)
**What happened?**

On Windows, `gbrain sync` is fully blocked since the per-file symlink-escape guard (`isPathSafe`, the #774/#2942 NAV-1 TOCTOU hardening) was added: **every** file in the repo — plain regular files, no symlinks or junctions anywhere (verified with `lstat`) — is rejected with `path resolves outside git repo (symlink escape)`, code `SYMLINK_NOT_ALLOWED`, one open ledger row per file in `~/.gbrain/sync-failures.jsonl`. The bookmark never advances, so the brain is permanently frozen at the last pre-upgrade commit.

Cause: `src/commands/sync.ts` `isPathSafe` (~line 1116) does `real.startsWith(rootReal + '/')`, but `realpathSync` returns native **backslash** separators on Windows (`C:\Users\me\vault\wiki\page.md` vs `C:\Users\me\vault/`) — the check can never match. Same bug class as #2828 (source_path separator mismatch, fixed by #2836 via separator normalization). Not present on 0.42.57.0 (the guard doesn't exist there yet).

**What did you expect?**

`isPathSafe` returns `true` for a regular in-repo file; the guard only rejects a committed symlink whose target realpath-resolves outside the repo. Separator-correct fix (same `path.relative` idiom the same file already uses for the scope-entry check ~line 1890):

```js
import { relative, isAbsolute, sep } from 'path';
function isPathSafe(filePath, gitRoot) {
  try {
    const rel = relative(realpathSync(gitRoot), realpathSync(filePath));
    return rel === '' || (!rel.startsWith('..' + sep) && rel !== '..' && !isAbsolute(rel));
  } catch { return false; }
}
```

Minimal alternative: `return real === rootReal || real.startsWith(rootReal + sep) || real.startsWith(rootReal + '/');` — this two-line local edit is the workaround we're running (sync healthy since), but it's wiped by every reinstall. A regression test on `windows-latest` (assert `isPathSafe` is `true` for a regular in-repo file) would have caught it.

**Steps to reproduce**
1. Windows machine, gbrain 0.42.63.0, any git-repo source with regular markdown files (no symlinks).
2. Commit any file change, then run `gbrain sync --repo <repo> --no-pull`.
3. Result: `Sync BLOCKED at <sha>: N file(s) failed to parse`, every file recorded as `{"error":"path resolves outside git repo (symlink escape)","code":"SYMLINK_NOT_ALLOWED"}`. Standalone repro of the predicate:
   ```js
   const { realpathSync } = require('fs');
   const real = realpathSync('C:/Users/me/vault/wiki/page.md');   // backslashes
   const rootReal = realpathSync('C:/Users/me/vault');
   real.startsWith(rootReal + '/');                               // false — always
   ```

**Environment**
- gbrain version: 0.42.63.0
- OS: Windows 11 Pro (NTFS local path, no symlinks/junctions in the repo)
- Bun version: 1.3.14
- Database: PGLite (local, no server)

Note: captured with `--fast` and **after** applying the two-line local workaround (with vanilla 0.42.63.0 the sync is fully blocked and the ledger holds one `SYMLINK_NOT_ALLOWED` row per file; those rows cleared once the patched sync succeeded).

```json
{"schema_version":2,"status":"warnings","health_score":80,"brain_checks_score":95,"category_scores":{"brain":95,"skill":90,"ops":95,"meta":100},"checks":[{"name":"resolver_health","status":"warn","message":"Could not find skills directory","category":"skill"},{"name":"retrieval_reflex_health","status":"warn","message":"pglite — serve IPC socket not present; enabled but no observed activity and no visible resolve path (host capability may still supply it; policy skill carries otherwise) — policy skill not installed; run `gbrain integrations install retrieval-reflex --target <host-repo>`","category":"skill"},{"name":"sync_failures","status":"warn","message":"5 unresolved sync failure(s) [UNKNOWN=5]. <head> (git HEAD verification failed: spawnSync git ETIMEDOUT); wiki/claude/handoffs/done/gbrain-sync-ps1-2026-07-20.md ([embed(ollama:nomic-embed-text)] Failed after 3 attempts. La); wiki/claude/handoffs/done/handoff-ob-humanizerplus-2026-07-20.md ([embed(ollama:nomic-embed-text)] Failed after 3 attempts. La), and 2 more.","category":"brain"},{"name":"nightly_quality_probe_health","status":"ok","message":"disabled (opt-in).","category":"brain"},{"name":"progressive_batch_audit_health","status":"ok","message":"No progressive-batch operations in the last 7 days","category":"ops"},{"name":"conversation_parser_probe_health","status":"ok","message":"Skipped (nightly probe is opt-in)","category":"brain"},{"name":"home_dir_in_worktree","status":"ok","message":"gbrain home is outside any enclosing git worktree.","category":"ops"},{"name":"connection","status":"warn","message":"Skipping DB checks (--fast mode, URL present from config-file-path)","category":"ops"}]}
```

time-attack:
Backlog fix wave nwf9.

## Fixes #3057 — Windows sync fully blocked by hardcoded '/' in isPathSafe

`isPathSafe` in `src/commands/sync.ts` did `real.startsWith(rootReal + '/')`; `realpathSync` returns backslash separators on Windows, so the predicate was false for every in-repo file — each one recorded `SYMLINK_NOT_ALLOWED`, freezing the sync bookmark. Same separator class as #2828/#2836.

Root cause is the shared idiom: `isPathContained` in `src/core/path-confine.ts` carried the same hardcoded `'/'`. Fix lands there once:

- New pure `isResolvedContained(child, parent, pathMod)` using the `path.relative` idiom (matching `isWriteTargetContained`), separator-agnostic and cross-drive-safe; `pathMod` is injectable so POSIX CI pins the win32 semantics with `path.win32`.
- `isPathContained` delegates to it (all existing callers — sources-ops clone confinement, repo-root skills-dir detection — get the fix for free).
- `sync.ts`'s `isPathSafe` and the NAV-1/NAV-2 scope-entry guard (same bug: every `--src-subpath` sync threw on Windows) both delegate to the shared helper.

Tests: new `isResolvedContained` win32 + posix cases in `test/path-confine.test.ts`; all existing containment/symlink-escape tests unchanged and green.

## References #3056 — verified, fix already open as PR #3252 (not duplicated here)

Adversarial verification of the in-flight community fix: applied #3252 onto current master locally — typecheck clean, its 9-test suite passes, `updateSlug` signature change lands in BOTH engines in lockstep, failures are recorded (stderr warn + `failedFiles` + `renameFallbacks`, checkpoint blocked on unresolved fallbacks) rather than re-swallowed, and the legitimate "old row genuinely absent → add" path is preserved. Recommend merging #3252 as-is; nothing for it is included in this PR.

## Skipped #3055 — already fixed on master

`list_pages` now honors explicit limits for trusted local callers (`ctx.remote === false` → effectively uncapped), warns on remote clamp, and the param description documents the remote 100 cap — landed in commit 70ffe4a2 (#2591). Issue #3055 can be closed.

### DIFF
--- a/src/commands/sync.ts
+++ b/src/commands/sync.ts
@@ -1,6 +1,7 @@
 import { existsSync, readFileSync, writeFileSync, statSync, realpathSync } from 'fs';
 import { execFileSync } from 'child_process';
 import { join, relative } from 'path';
+import { isPathContained, isResolvedContained } from '../core/path-confine.ts';
 import type { BrainEngine } from '../core/engine.ts';
 import { DELETE_BATCH_SIZE } from '../core/engine-constants.ts';
 import { importFile } from '../core/import-file.ts';
@@ -1112,15 +1113,14 @@ function createSyncBaselineCommit(repoPath: string): void {
  * #774 NAV-1 TOCTOU: true only if filePath realpath-resolves inside gitRoot.
  * Guards symlink escape at the per-file level (a committed symlink whose
  * target lives outside the repo), not just at scope entry.
+ *
+ * #3057: delegates to the shared separator-agnostic helper — the previous
+ * inline `startsWith(rootReal + '/')` was always false on Windows
+ * (realpathSync returns backslashes), recording every in-repo file as
+ * SYMLINK_NOT_ALLOWED and freezing the sync bookmark.
  */
 function isPathSafe(filePath: string, gitRoot: string): boolean {
-  try {
-    const real = realpathSync(filePath);
-    const rootReal = realpathSync(gitRoot);
-    return real === rootReal || real.startsWith(rootReal + '/');
-  } catch {
-    return false;
-  }
+  return isPathContained(filePath, gitRoot);
 }
 
 function hasOriginRemote(repoPath: string): boolean {
@@ -1887,7 +1887,8 @@ async function performSyncInner(engine: BrainEngine, opts: SyncOpts): Promise<Sy
   // NAV-1/NAV-2 scope-entry guard: the realpath-resolved scope must live
   // inside the realpath-resolved git root. Catches `--src-subpath ../escape`
   // AND a symlinked subdir pointing outside the repo, before any git op runs.
-  if (syncScopeRoot !== gitContextRoot && !syncScopeRoot.startsWith(gitContextRoot + '/')) {
+  // #3057: separator-agnostic (both sides already realpath'd above).
+  if (!isResolvedContained(syncScopeRoot, gitContextRoot)) {
     throw new Error(
       `Sync scope ${syncScopeRoot} resolves outside git repo ${gitContextRoot}. ` +
       `Refusing to sync: possible path traversal via --src-subpath.`,
--- a/src/core/path-confine.ts
+++ b/src/core/path-confine.ts
@@ -20,11 +20,34 @@
  */
 
 import { realpathSync, existsSync, type Stats } from 'fs';
+import * as nodePath from 'path';
 import { resolve as resolvePath, relative, isAbsolute, dirname, basename, join } from 'path';
 
 /**
+ * Pure containment predicate over ALREADY-resolved paths: true iff `child`
+ * IS `parent` or lives under it. Separator-agnostic via `path.relative`.
+ *
+ * #3057: the previous `startsWith(parent + '/')` form is false for EVERY
+ * path on Windows — `realpathSync` returns backslash separators there — so
+ * every containment check built on it failed closed and blocked sync
+ * entirely (same separator class as #2828/#2836). `pathMod` is injectable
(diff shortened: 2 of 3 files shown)


time-attack:
Closing — this landed on master already as 5f84fb88 (#3415), which introduced `isWithinRoot()` and routed both `isPathSafe` and the `syncScopeRoot` scope-entry guard through the same separator-agnostic `path.relative` predicate, with regression tests. The only remaining delta here is a refactor of `isPathContained`, which is a no-op on the platforms we ship. Thanks for chasing this one down — the diagnosis was correct.

PR #3285, opened 2026-07-23
fix(facts,cycle,doctor): fence stub-dup on basename≠slug, quality-probe self-mute, wedged-queue misdiagnosis

### LINKED ISSUE #3063: doctor wedged_queue reports "worker alive but not claiming work" without checking whether a worker exists — misdirects the operator when the worker is simply absent
## Summary

`wedged_queue` diagnoses a specific failure — a live worker whose DB pool has gone bad —
but its query cannot distinguish that from *no worker running at all*. Both produce the
same row shape in `minion_jobs`, and the check asserts the former in its message and its
remedy.

## Mechanism

`computeWedgedQueueCheck` (`src/commands/doctor.ts:1606`) reads only `minion_jobs`:

```sql
SELECT queue,
  count(*) FILTER (WHERE status = 'active' AND lock_until > now()) AS active_healthy,
  count(*) FILTER (WHERE status = 'waiting')                       AS waiting,
  EXTRACT(EPOCH FROM (now() - max(updated_at) FILTER (WHERE status = 'completed'))) / 60
    AS mins_since_completion
FROM minion_jobs GROUP BY queue
```

and flags when `active_healthy === 0 && waiting > 0 && mins > threshold`. Nothing in that
predicate observes worker liveness — no supervisor PID file, no heartbeat, no
`readSupervisorConcurrency`. It then reports:

> Wedged queue(s) — **worker alive but not claiming work**: 'default' (40 waiting, 0 active,
> 1934m since last completion). Restart the worker so it **rebuilds a fresh DB pool**:
> `gbrain jobs supervisor stop && gbrain jobs supervisor start`

Both bolded claims are inferences the query cannot support.

## Observed

A brain where the autopilot had shut *itself* down three days earlier:

```
[reconnect] ERROR: undefined is not an object (evaluating 'config.database_url')
5 consecutive cycle failures. Stopping autopilot.
Autopilot stopping (cycle-failure-cap).
```

It exited `0/SUCCESS`, so the systemd unit sat `inactive (dead)` (a clean exit, and
`Restart=always` had already been satisfied). `gbrain jobs supervisor status` correctly
reported **"Supervisor: not running"**, and no `jobs work` process existed. Meanwhile
`doctor` insisted the worker was alive and blamed the DB pool.

Note the adjacent check gets this right in the same run:

```
[OK] autopilot_fanout_concurrency: No supervisor observed — skipping fan-out/concurrency check
```

`computeAutopilotFanoutConcurrencyCheck` calls `readSupervisorConcurrency('default')` and
returns null when no supervisor has been observed. So the liveness signal is already
available in the same file; `wedged_queue` just doesn't consult it.

## Impact

The remedy happens to start a worker, so an operator following it recovers — but for the
wrong stated reason, and the misdiagnosis costs real time on the two questions it provokes
("which worker? why would its pool be bad?"). It also hides the genuine failure: a
supervisor that died and did not come back is a different incident from a wedged pool, with
a different root cause and a different durable fix.

## Suggested fix

Branch on worker liveness before choosing the message:

- **no worker observed** → `status: 'fail'`, message along the lines of "queue 'default' has
  N waiting jobs and no worker running", remedy `gbrain jobs supervisor start` (plus a
  pointer to the autopilot unit if one is installed).
- **worker observed, not claiming** → keep the current wedged-pool message and remedy.

Reuse `readSupervisorConcurrency` / the supervisor PID-file check that
`gbrain jobs supervisor status` already relies on.

Failing that, at minimum soften the message so it states the observation
("no live-locked active jobs, N waiting, last completion Xm ago") rather than the
unverified conclusion.

## Environment

gbrain 0.42.63.0, postgres engine, autopilot installed as a systemd user unit.

### LINKED ISSUE #3064: nightly_quality_probe self-mutes after one run: rate_limited skip events re-arm the 24h window
## Summary

On an autopilot ticking every `--interval` (default 300s), `nightly_quality_probe` runs exactly once and then never fires again. The probe self-mutes because `rate_limited` skip events re-arm the same 24h window they are gated by.

Observed on 0.42.63 (from-source build).

## Mechanism

1. `runNightlyQualityProbe` logs a `rate_limited` audit event on every skip (`src/.../nightly-quality-probe.ts`, around line 109).
2. `shouldRunNightly(now, readRecentQualityProbeEvents(2, now))` returns "rate limited" when ANY recent event has `ts >= now - 24h`. There is no outcome filter, so a `rate_limited` skip event counts toward the window exactly like a real run.
3. Autopilot ticks every `--interval` (default 300s, `src/.../autopilot.ts` around line 387).

So after the first real run at T0, each ~5-minute tick sees a `<24h`-old event, skips, and logs another `rate_limited` event with `ts = now`. The newest event is always ~5 minutes old, so the 24h window never clears. Result: one real run, then roughly 288 `rate_limited` events per day, indefinitely. The recurring nightly signal the probe is meant to produce never recurs.

## Impact

An operator who enables `nightly_quality_probe` gets a single run and then permanent silence, with no error surfaced (the skips are logged as `rate_limited`, which reads as expected throttling rather than a stuck gate). The `~/.gbrain/audit/quality-probe-*.jsonl` file fills with skip events while no new verdict is ever recorded.

## Suggested fix

Either:
- Exclude `rate_limited` / skip outcomes from the window check in `shouldRunNightly`, so only real run outcomes arm the 24h window; or
- Do not write an audit event on a `rate_limited` skip (only real runs get an event), which has the same effect.

Either change lets the window clear ~24h after the last real run, restoring the intended daily cadence.

### LINKED ISSUE #3069: writeFactsToFence stub-creates a duplicate page (and collides on idx_facts_fence_key) when the entity file's basename differs from its slug
`writeFactsToFence` (src/core/facts/fence-write.ts) locates the entity page with:

```ts
const filePath = join(target.localPath, `${target.slug}.md`);
```

For any page whose on-disk basename is not exactly its slug — e.g. a vault file `10-people/Joey Fteha.md` imported with slug `10-people/joey-fteha` — `existsSync` misses the real file and the function falls into the stub-create branch. Consequences, all observed live on v0.42.59:

1. A duplicate-slug stub `.md` is written next to the real page (the dup-slug stub class adjacent to #2723), because the stub write happens BEFORE the DB insert.
2. The stub's fence is empty, so `upsertFactRow` assigns row_num starting at 1; `engine.insertFacts` then dies on `idx_facts_fence_key` because the REAL page's rows already occupy those row_nums. The whole batch rolls back — but the stub file stays on disk.
3. The stub gets `type: concept` even for people pages, because `stubEntityPage`'s prefix map matches `people/` but real vaults use prefixes like `10-people/`.

The engine already knows the true filename: `pages.import_filename` holds the imported basename. Suggested fix (running locally): when the slug-derived path is absent, resolve via `import_filename` in the slug's directory before concluding the page doesn't exist:

```ts
if (!existsSync(filePath)) {
  const rows = await engine.executeRaw(
    `SELECT import_filename FROM pages WHERE slug = $1 AND source_id = $2 AND deleted_at IS NULL LIMIT 1`,
    [target.slug, target.sourceId],
  );
  const importName = rows[0]?.import_filename;
  if (importName) {
    const dir = target.slug.includes('/') ? target.slug.slice(0, target.slug.lastIndexOf('/')) : '';
    const candidate = join(target.localPath, dir, importName.endsWith('.md') ? importName : `${importName}.md`);
    if (existsSync(candidate)) filePath = candidate;
  }
}
```

**Repro:** import a page named `First Last.md` under a prefixed dir, let facts land on it once (fence row 1 exists), then call the `extract_facts` op with a hint resolving to that slug. Observe the stub + the unique-constraint error:

```
ERROR:  duplicate key value violates unique constraint "idx_facts_fence_key"
DETAIL:  Key (source_id, source_markdown_slug, row_num)=(default, 10-people/joey-fteha, 1) already exists.
```

Happy to open a PR with the patch above if the approach looks right.

time-attack:
Backlog fix wave, work unit nwf7. Three independently verified fixes, each with a regression test that fails without the fix.

## Fixes #3069 — writeFactsToFence stub-creates a duplicate page when the file's basename differs from its slug

`writeFactsToFence` located the entity file via the slug-derived path only, so a page like `10-people/First Last.md` (slug `10-people/first-last`) missed `existsSync`, bypassed the stub guard, and stub-created a duplicate page. The stub's empty fence restarted `row_num` at 1, colliding with the real page's DB rows on `idx_facts_fence_key` (whole batch rolled back, stub file stranded on disk).

Fix: before concluding the page doesn't exist, resolve the real on-disk file from the page row — `source_path` first, then `import_filename` in the slug's directory. When a page row exists but no backing file can be located, refuse to stub-create (an empty-fence stub would both collide on row_num and get imported over the real page content by a later sync) and route the facts to the legacy DB-only path via the existing `stubGuardBlocked` return so nothing is dropped. Genuinely-new slugs (no page row) keep the pre-existing stub-create + unprefixed-slug guard behavior.

Tests: `test/fence-write.test.ts` — basename≠slug resolves to the real file and continues the row_num sequence (asserts row_num 2, no duplicate stub); page-row-without-file returns `stubGuardBlocked` and creates nothing.

## Fixes #3064 — nightly quality probe runs once, then self-mutes forever

`shouldRunNightly` counted ANY recent audit event toward the 24h window, and the skip path logged its own `rate_limited` audit row. Autopilot ticks every ~5 min, so each skip re-armed the window: exactly one real run, then permanent self-mute at ~288 skip rows/day — which also permanently tripped doctor's `nightly_quality_probe_health` warn (rate_limited counts as non-PASS).

Fix: the gate now filters `rate_limited` rows before deciding (defends against historical skip rows already in pre-fix brains' audit files), and the skip no longer logs an audit event.

Tests: `test/nightly-quality-probe.test.ts` — a real run 25h ago plus recent historical skip rows must run the probe; the skip path writes no audit row.

## Fixes #3063 — doctor blames a wedged DB pool when the worker simply isn't running

`computeWedgedQueueCheck` unconditionally emitted "worker alive but not claiming work … rebuilds a fresh DB pool" without observing worker liveness, sending operators down the wrong path when the supervisor was cleanly dead.

Fix: the check now inspects the supervisor's singleton DB lock (`gbrain_cycle_locks`, the same proven-alive signal `resolveEffectiveFanoutMax` trusts) per wedged queue. No live holder → fail with "waiting jobs and no worker running" + `gbrain jobs supervisor start` remedy; live holder → the existing wedged-pool message. Lock-table read errors fall back to the conservative wedged-pool message, and the check still only reads `.kind` + `.executeRaw` off the engine.

Tests: `test/doctor-wedged-queue.test.ts` — no live lock → no-worker message; live lock row → wedged-pool message. All pre-existing cases unchanged.

## Verification

- `bun run typecheck` exit 0
- `bun test test/fence-write.test.ts test/nightly-quality-probe.test.ts test/doctor-wedged-queue.test.ts` — 51 pass / 0 fail
- Adjacent suites (`test/facts-backstop.test.ts`, `test/stub-guard-audit.test.ts`, `test/doctor.test.ts`) — 107 pass / 0 fail
- Each new test verified to fail with the fix reverted

### DIFF
--- a/src/commands/doctor.ts
+++ b/src/commands/doctor.ts
@@ -1623,7 +1623,7 @@ export async function computeWedgedQueueCheck(engine: BrainEngine): Promise<Chec
        FROM minion_jobs
        GROUP BY queue`,
     );
-    const wedged: string[] = [];
+    const wedged: Array<{ queue: string; label: string }> = [];
     for (const r of rows) {
       const activeHealthy = Number(r.active_healthy ?? 0);
       const waiting = Number(r.waiting ?? 0);
@@ -1631,21 +1631,59 @@ export async function computeWedgedQueueCheck(engine: BrainEngine): Promise<Chec
       // Conservative: only flag stale-after-progress (non-null mins past
       // threshold). The null-completions case is the supervisor's job.
       if (activeHealthy === 0 && waiting > 0 && mins !== null && mins > thresholdMin) {
-        wedged.push(`'${r.queue}' (${waiting} waiting, 0 active, ${Math.round(mins)}m since last completion)`);
+        wedged.push({
+          queue: r.queue,
+          label: `'${r.queue}' (${waiting} waiting, 0 active, ${Math.round(mins)}m since last completion)`,
+        });
       }
     }
     if (wedged.length === 0) {
       return { name: 'wedged_queue', status: 'ok', message: 'No wedged queues' };
     }
-    return {
-      name: 'wedged_queue',
-      status: 'fail',
-      message:
-        `Wedged queue(s) — worker alive but not claiming work: ${wedged.join('; ')}. ` +
+    // #3063: split the diagnosis on worker liveness. The wedged-pool message
+    // ("worker alive but not claiming work → restart to rebuild the DB pool")
+    // only applies when a supervisor actually holds a live singleton lock.
+    // With no live holder the worker is simply not running — telling the
+    // operator to blame the DB pool sends them down the wrong path.
+    const alive: string[] = [];
+    const noWorker: string[] = [];
+    for (const w of wedged) {
+      let hasLiveWorker = false;
+      try {
+        const { supervisorLockId } = await import('../core/minions/supervisor.ts');
+        const lockRows = await engine.executeRaw<{ live: boolean }>(
+          `SELECT (ttl_expires_at > now()) AS live FROM gbrain_cycle_locks WHERE id = $1`,
+          [supervisorLockId(w.queue)],
+        );
+        hasLiveWorker = lockRows.length > 0 && lockRows[0].live === true;
+      } catch {
+        // Lock table unreadable → can't prove absence; keep the
+        // conservative wedged-pool message.
+        hasLiveWorker = true;
+      }
+      (hasLiveWorker ? alive : noWorker).push(w.label);
+    }
+    const parts: string[] = [];
+    if (noWorker.length > 0) {
+      parts.push(
+        `Queue(s) with waiting jobs and no worker running: ${noWorker.join('; ')}. ` +
+        `Start the worker: \`gbrain jobs supervisor start\` ` +
+        `(or restart the autopilot unit if installed).`,
+      );
+    }
+    if (alive.length > 0) {
+      parts.push(
+        `Wedged queue(s) — worker alive but not claiming work: ${alive.join('; ')}. ` +
         `Restart the worker so it rebuilds a fresh DB pool: ` +
         `\`gbrain jobs supervisor stop && gbrain jobs supervisor start\`, ` +
         `then \`gbrain jobs retry <id>\` on any dead-lettered jobs.`,
-      details: { wedged_queues: wedged.length, threshold_minutes: thresholdMin },
+      );
+    }
+    return {
+      name: 'wedged_queue',
+      status: 'fail',
+      message: parts.join(' '),
+      details: { wedged_queues: wedged.length, no_worker_queues: noWorker.length, threshold_minutes: thresholdMin },
     };
   } catch (e) {
     // Pre-migration brains / transient errors: advisory check stays ok.
--- a/src/core/cycle/nightly-quality-probe.ts
+++ b/src/core/cycle/nightly-quality-probe.ts
@@ -101,21 +101,20 @@ export async function runNightlyQualityProbe(deps: NightlyProbeDeps): Promise<Ni
     return { outcome: 'disabled', exit_code: 0, detail: 'feature flag off' };
   }
 
-  // 24h rate limit — skip + audit "rate_limited".
+  // 24h rate limit — skip quietly (no audit row).
+  // #3064: gate only on events from ATTEMPTED runs. The skip path used to log
+  // its own 'rate_limited' audit row, and autopilot ticks every ~5 min — each
+  // skip re-armed the 24h window, so the probe ran exactly once and then
+  // self-muted forever. Filter defends against historical skip rows already
+  // sitting in the audit files of pre-fix brains.
   const now = deps.now();
   const recent = readRecentQualityProbeEvents(2, now); // 2-day window is enough for 24h check
-  const decision = shouldRunNightly(now, recent);
+  const decision = shouldRunNightly(now, recent.filter((e) => e.outcome !== 'rate_limited'));
   if (!decision.run) {
-    logQualityProbeEvent({
-      outcome: 'rate_limited',
-      exit_code: 0,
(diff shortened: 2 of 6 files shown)


time-attack:
Closing this bundle. The quality-probe self-mute is already resolved on master — the skip path no longer writes a `rate_limited` audit row — so that hunk is stale and the branch will not apply cleanly. The wedged-queue change is wording only, but it buys that wording with a dynamic import and a per-queue lock query inside a loop. That leaves the fence-write basename-vs-slug fix as the one live defect, and it is not reviewable while braided together with two unrelated changes. Thanks for the careful write-up on each item.

PR #3286, opened 2026-07-23
fix(models): doctor probes per-task routes + probe token budget; honor file-plane chat_model

### LINKED ISSUE #3206: chat_model is dead config in any engine-backed process: resolveModel's tier default outranks it, so a self-hosted chat provider silently never runs
## Summary

Setting `chat_model` (in `~/.gbrain/config.json`, in the `config` table, or both)
has **no effect** on `getChatModel()` in any process that has an engine — which
is every real usage: the CLI, the MCP server, and the minion worker. It is
silently replaced by `TIER_DEFAULTS.reasoning` (`anthropic:claude-sonnet-4-6`).

For a brain configured against a self-hosted OpenAI-compatible endpoint and no
Anthropic key, the consequence is that every chat-gated feature reports
"unavailable" and does nothing, with **no error and no log line**.

## Mechanism

`resolveModel` (`src/core/model-config.ts:133`) ends:

```ts
  // 7. Tier default (v0.31.12 — when no override beats us, the tier's
  //    canonical model wins over caller-supplied fallback)
  if (opts.tier && TIER_DEFAULTS[opts.tier]) {
    return await resolveAlias(engine, TIER_DEFAULTS[opts.tier]);
  }

  // 8. Hardcoded fallback (caller-supplied)
  return await resolveAlias(engine, opts.fallback);
```

Step 7 deliberately outranks step 8, and the comment describes step 8 as a
"hardcoded last-resort fallback". That is a reasonable contract **only if the
fallback really is hardcoded**.

It isn't. `reconfigureGatewayWithEngine` (`src/core/ai/gateway.ts:511`) passes the
user's configured value as that fallback:

```ts
const newChat = await resolveModel(engine, {
  configKey: 'models.chat',
  tier: 'reasoning',
  fallback: cfg.chat_model ?? DEFAULT_CHAT_MODEL,   // <-- user config, ranked BELOW the tier default
});
...
_config = { ...cfg, expansion_model: expansionFull, chat_model: chatFull };
```

So user config enters the chain at the lowest priority and can never win. Unless
`models.chat`, `models.default`, `models.tier.reasoning` or `GBRAIN_MODEL` is
set, `chat_model` is overwritten on every startup.

`cli.ts:2231` calls `reconfigureGatewayWithEngine(engine)` right after
`connectEngine()`, so this happens on essentially every command.

`expansion_model` is passed the same way (`tier: 'utility'`) and has the same
problem.

## Why it is invisible

The failure lands somewhere other than the setting that caused it. In facts
extraction:

1. `getFactsExtractionModel` uses `configKey: 'facts.extraction_model'` — step 2,
   which **does** outrank the tier default. So the extraction model resolves
   correctly to the self-hosted model.
2. But `extractFactsFromTurn` guards on `isAvailable('chat')`, which uses
   `getChatModel()` — the *overwritten* value.
3. That resolves to the Anthropic recipe, whose `auth_env.required` is
   `['ANTHROPIC_API_KEY']`, absent → `isAvailable('chat')` is false.
4. `extractFactsFromTurn` returns `[]` at its first guard.

The guard checks a **different model than the one the call would have used**. The
job then completes with `{"inserted":0,"duplicate":0,"superseded":0}` — identical
to a page that genuinely had no facts — in well under a second.

## Reproduction / evidence

Brain: postgres engine, self-hosted OpenAI-compatible chat via the `litellm`
recipe, no `ANTHROPIC_API_KEY`. Both config planes set:

```
config.json : chat_model = litellm:qwen2.5:7b-instruct-q4_K_M
config table: chat_model = litellm:qwen2.5:7b-instruct-q4_K_M
              facts.extraction_model = litellm:qwen2.5:7b-instruct-q4_K_M
```

Temporary instrumentation at the top of `extractFactsFromTurn`, running a real
`facts-absorb` job through `gbrain jobs work`:

```
[DBG-EXTRACT] textLen=8614 chatAvail=false chatModel=anthropic:claude-sonnet-4-6
[DBG-EXTRACT] BAILING: isAvailable(chat) false
→ job completed, {"inserted":0,...}, 0s
```

Note `chatModel` is Anthropic despite both planes saying otherwise.

Adding **`models.chat`** (step 2, which outranks the tier default) and changing
nothing else:

```
[DBG-EXTRACT] textLen=8614 chatAvail=true chatModel=litellm:qwen2.5:7b-instruct-q4_K_M
→ job completed, {"inserted":7,"fact_ids":[5,6,7,8,9,10,11]}, 128s
```

Same page, same code, same credentials. Only the config key differs.

Corroborating: calling `configureGateway(buildGatewayConfig(loadConfig()))` in a
bare process with **no engine** yields `chat_model = litellm:...` and
`isAvailable('chat') === true` — because `reconfigureGatewayWithEngine` never
runs. Engine-free code paths see the user's config; engine-backed ones do not.

## Suggested fixes

1. **Distinguish "user configured this" from "caller's hardcoded default".** Add
   something like `userFallback` that ranks *above* the tier default, and have
   `reconfigureGatewayWithEngine` pass `cfg.chat_model` there — keeping
   `DEFAULT_CHAT_MODEL` at step 8 where the existing comment's reasoning holds.
2. **Failing that, don't silently discard it** — if `cfg.chat_model` is set and
   the tier default overrides it, warn once naming both values and the key to set
   (`models.chat`). Today nothing indicates the setting was ignored.
3. **Make the availability guard check the model it will actually call.**
   `extractFactsFromTurn` should gate on the resolved extraction model, not on
   `getChatModel()`. Even with the above fixed, these can diverge whenever
   `facts.extraction_model` differs from `chat_model`, and the failure mode is
   the same silent zero.
4. **Document the precedence.** `models.chat` beating `chat_model` is very
   surprising, and `gbrain config set chat_model` reports success either way.

Related: #3062 (the same silent-zero surfaces because the unauthenticated-chat
path returns `[]` with no `skipped` reason and no log).

## Environment

gbrain 0.42.63.0, postgres engine, `litellm` recipe against a self-hosted
endpoint, no Anthropic credentials.

### LINKED ISSUE #3219: Doctor skips configured background model routes
`gbrain models doctor` probes only the global chat and expansion models. The newly configurable `models.dream.extract_atoms` and `models.dream.synthesize_concepts` routes can point at a distinct unavailable provider while doctor still reports green.

Suggested fix: resolve and probe each distinct background route, deduplicate identical provider/model combinations, and retain route-key attribution in JSON diagnostics.

- Severity: P1
- Confidence: 100
- Finding ID: `src/commands/models.ts:604-doctor-background-routes`

### LINKED ISSUE #3221: One-token doctor probes falsely fail reasoning models

time-attack:
Backlog fix wave nwf4 — three adversarially verified, unfixed problems in the model-routing/doctor surface.

## Fixes #3219 — doctor skips configured background model routes

`gbrain models doctor` probed only the global chat + expansion models, so a per-task background route (`models.dream.synthesize`, `models.think`, `facts.extraction_model`, ...) pointing at a distinct unreachable provider still reported green. (The issue's route names `models.dream.extract_atoms` / `models.dream.synthesize_concepts` don't exist on master; the fix uses the actual `PER_TASK_KEYS`.)

- New `resolvePerTaskProbePlan(engine, alreadyProbed)` resolves every `PER_TASK_KEYS` route through the same `resolveModel` path `buildReport` uses, dedups against the already-probed chat/expansion models and across routes, and probes each distinct resolved model exactly once.
- Route-key attribution is retained in `ProbeResult.touchpoint` (comma-joined keys, e.g. `models.auto_think,models.think`), and `--skip=<provider>` applies to route probes too.
- Test: `test/models-doctor-per-task-routes.test.ts`.

## Fixes #3221 — one-token doctor probe falsely fails reasoning models (takeover of #3250)

Takeover of open PR #3250 by @Masashi-Ono0611 (credited via Co-authored-by; applied cleanly onto master, nothing to close there — just referencing). The probe's `maxTokens: 1` is either exhausted by reasoning burn (length stop, empty text) or rejected outright by providers with output-token minimums.

- Shared `PROBE_MAX_OUTPUT_TOKENS = 64` budget, widened above any configured extended-thinking budget (`getConfiguredThinkingBudget`, alias-aware); providers bill generated tokens, so non-reasoning models stay ~1-token cost.
- A valid length-exhausted empty completion classifies as reachable, with the generation limitation surfaced in human and `--json` output.
- Added on top of the takeover: a catch-branch classification so the probe stays coherent with #3249's gateway-side contentless-completion rejection if that lands first — the gateway-thrown "output budget exhausted before any content" error is the same reachable-but-reasoning-burned class, not a probe failure.
- Test: `test/models-doctor-probe-token-budget.test.ts` (from #3250, extended with the #3249-coherence pin).

## Fixes #3206 — configured chat_model silently replaced by the tier default

`reconfigureGatewayWithEngine` passed the file-plane `chat_model` / `expansion_model` as `resolveModel`'s bottom-rung `fallback`, which loses to `TIER_DEFAULTS` (step 7) — so explicit user config was silently overwritten on every engine-backed startup, and on non-Anthropic brains every chat-gated feature reported unavailable with no error.

- New `userFallback` slot on `ResolveModelOpts`, ranked between the env var and the tier default: caller-attested explicit user config beats `TIER_DEFAULTS`, while `models.*` config keys and `GBRAIN_MODEL` still win. `reconfigureGatewayWithEngine` passes it only when the gateway value differs from the hardcoded default, so hardcoded defaults stay at the bottom of the chain.
- Silent-zero half: the facts extraction gate (`extractFactsFromTurn`) and the classifier gate (`classifyAgainstCandidates`) now check `isAvailable('chat', <model actually used>)` instead of the global chat default — `facts.extraction_model` on a different provider than `chat_model` no longer extracts zero facts silently.
- Tests: `test/facts-model-gate.test.ts` (new), plus precedence pins in `test/model-config.serial.test.ts` and `test/gateway-tier-extended-models.test.ts`.

## Verification

- `bun run typecheck` exits 0.
- All touched/new test files pass (53 tests across the 6 model/doctor files; 81 across the 10 related facts/doctor suites). Each fix's pinning tests were verified to FAIL with the fix reverted.
- Docs touched (`docs/INSTALL.md`, `docs/architecture/KEY_FILES.md`, `skills/conventions/model-routing.md`) → `bun run build:llms` run; bundles fresh (`test/build-llms.test.ts` passes).
- No VERSION / package.json / CHANGELOG changes (release bookkeeping happens at ship time).


time-attack:
Closing. The `chat_model` being outranked by `TIER_DEFAULTS` is a real defect, but it arrives here inside 706 lines that also add a new per-task doctor probe fan-out (new diagnostic surface) and insert a new rung into `resolveModel`'s precedence chain — a behavior change in the most load-bearing path in the model layer, bundled with unrelated work and a takeover of #3250. We are not taking that combined surface in one reviewable unit. The analysis of each item is genuinely good; the packaging is what closes it.

PR #3287, opened 2026-07-23
fix(backlog): facts-absorb wall clock, contentless-completion guard, pack-driven link extraction (nwf5)

### LINKED ISSUE #3190: Schema-pack link verbs and frontmatter_links cannot produce typed edges via extract links — entity-dir whitelist and hardcoded frontmatter map make custom packs decorative for link extraction
## Summary

A custom schema pack can declare `link_types[].inference.regex` and
`frontmatter_links:` mappings that validate cleanly and display correctly in
`gbrain schema show` — but `gbrain extract links` (both `--source db` and the
filesystem path) can never create a single typed edge from them for a brain
whose content doesn't follow gbrain's built-in directory taxonomy. Three
independent gates each kill it; all three are hardcoded and none is
pack-configurable. The result is silent: `extract links` walks every page and
reports `created 0` with no hint about why.

## Environment

- gbrain 0.42.58.0 (`bun install -g github:garrytan/gbrain`, commit `a25209b`);
  code paths re-verified identical on master @ 0.42.64.0
- Postgres 17 + pgvector 0.8.5 engine; Bun 1.3.14; WSL2 Ubuntu
- Custom pack (5 page types with `path_prefixes: [wiki/people/, …]`; 6 link
  verbs with regex-only inference; `frontmatter_links` mapping
  `parents`→`parent_of`, `children`→`child_of`). Pack is active, `schema
  validate` and `schema lint` clean.

## Minimal repro

Corpus: an Obsidian-style wiki whose pages live under nested dirs
(`wiki/people/<slug>.md`) and cross-link with standard relative markdown links:

```markdown
- parent of [Harriet Emeline Pettit](pettit-harriet-emeline-1828.md) _(S022, unverified)_
- member of the [Pettit](../lines/pettit.md) line
```

with frontmatter:

```yaml
parents: [pettit-james-b-1777, felt-lucy-w-1777]
children: [barker-helen-elizabeth-1825, pettit-harriet-emeline-1828]
```

Pack link verb: `{name: parent_of, inference: {regex: '[Pp]arent of \['}}` and
`frontmatter_links: [{page_type: person, fields: [parents], link_type: parent_of}]`.

Run: `gbrain extract links --source db` → walks all pages, `created 0`.

## Root cause (traced by executing the installed functions directly)

Trace harness: called `extractPageLinks` / `extractFrontmatterLinks` /
`inferLinkTypeFromPack` from the installed package against the real page
content and synthetic controls differing only in link-target shape.

**Gate 1 — `ENTITY_REF_RE` / `DIR_PATTERN` whitelist**
(`src/core/link-extraction.ts` ~L86): markdown-link targets must start (after
optional `../`) with one of
`people|companies|meetings|concepts|deal|civic|project|projects|source|media|yc|tech|finance|personal|openclaw|entities`.
Same-directory links (`[N](slug.md)`) and non-whitelisted dirs
(`[N](../lines/x.md)`) yield **candidates = 0**, so pack verb regexes are
never consulted. Trace: real page containing `parent of [` → 0 candidates;
identical sentence with target `people/slug` → 1 candidate. The whitelist is a
`const`; `path_prefixes` declared by the active pack's page types do not feed
it, and no config key extends it.

**Gate 2 — pack inference never runs on the plain extract path**:
`inferLinkTypeFromPack` is called only from `src/core/extract-ner.ts`
(`--ner`). The plain fs/db path persists `candidate.linkType` from the
hardcoded `inferLinkType` — so even a whitelisted-shape link in a sentence
matching a pack verb regex is stored as `mentions`. Trace: whitelisted-shape
candidate came back `linkType: "mentions"` while `inferLinkTypeFromPack` on
the same context returns `parent_of`. (NER-based typing is not a substitute
where entities share exact names — genealogy corpora have identical-name
ancestors, and the gazetteer's maximal-munch tie-break mis-targets them.)

**Gate 3 — `frontmatter_links` has no extraction-time consumer**:
`extractFrontmatterLinks` iterates only the hardcoded `FRONTMATTER_LINK_MAP`
(fields `company`, `investors`, `attendees`, `sources`, …). The pack's
`frontmatter_links` array is parsed, validated, and read by exactly one
consumer — the schema-graph visualization op (`operations.ts` ~L4608). Trace:
real frontmatter with `parents`/`children` → 0 candidates; the only resolver
calls came from the built-in `sources` mapping.

(Minor fourth gate for completeness: harvested slugs are relative to the
whitelist dir (`people/…`) and don't match nested stored slugs
(`wiki/people/…`), so even surviving candidates would drop at the persist
join.)

## Why this matters

The pack authoring surface (`schema add-link-type`, `frontmatter_links`)
reasonably implies that a lint-clean active pack will drive extraction. It
doesn't, and nothing warns. For any KB not born in gbrain's canonical
directory layout, `extract links` silently produces zero typed edges while
`schema show` displays a perfectly-registered set of verbs — a several-day
debugging trap (config layer is provably correct, engine never consults it).

## Proposal (either/both would resolve it)

1. **Feed the entity-dir whitelist from the active pack**: union
   `DIR_PATTERN` with the active pack's `page_types[].path_prefixes` (and/or
   add a `link_extraction.entity_dirs` config key). Also normalize harvested
   slugs against the matching page-type prefix so nested layouts
   (`wiki/people/…`) resolve.
2. **Wire `frontmatter_links` into `extractFrontmatterLinks`**: append
   pack-declared mappings to `FRONTMATTER_LINK_MAP` at pack load (fields →
   link_type, page_type filter; values that are already page slugs/ids should
   skip name-resolution). This makes the existing manifest field real.
3. Cheap interim: have `extract links` emit a one-line notice when a pack
   declares link verbs / frontmatter_links but the run produced 0 typed
   candidates — pointing at the whitelist — so the failure isn't silent.

## Workaround (for anyone else landing here)

Bypass extraction: derive edges from frontmatter directly and batch-insert via
`engine.addLinksBatch` with a custom kebab `link_source` provenance (survives
reconciliation; idempotent reload = delete-by-provenance + re-insert). Happy
to share the ~150-line loader if useful.

### LINKED ISSUE #3207: facts-absorb missing from HANDLER_DEFAULT_TIMEOUT_MS: LLM extraction is wall-clock killed at the tight null-default, exactly the failure that file exists to prevent
## Summary

`facts-absorb` performs a single LLM extraction call per page, but it is absent
from `HANDLER_DEFAULT_TIMEOUT_MS` (`src/core/minions/handler-timeouts.ts:33`). It
therefore inherits the tight null-default wall-clock budget
(`lockDuration * 2 * GREATEST(max_stalled,1)`, `queue.ts:779`) and is
dead-lettered mid-generation on any page that takes longer than that.

This is precisely the scenario that file was written to fix. From its own
docblock:

> Short jobs (shell, lint, backlinks) want the tight default wall-clock […]. Long
> jobs do not: **a 30-min LLM loop or a 10-15 min embed backfill submitted WITHOUT
> an explicit `timeout_ms` would inherit that short null-default and get
> wall-clock-killed mid-progress** — one half of #1737's thrash.

`chronicle_extract` is in the map with a 10-minute budget and is described as
"one page = one LLM extraction call + a few writes". `facts-absorb` is the same
shape and was missed.

## Observed

Brain on a self-hosted OpenAI-compatible chat endpoint (CPU inference, so
generation is slow but entirely healthy).

Two `facts-absorb` runs on an 8.6k-char page completed normally at **122s** and
**128s**, inserting 4 and 7 facts.

A third on a comparable page:

```
id    | status | timeout_ms | attempts_made | error_text
16947 | dead   | (null)     | 1             | wall-clock timeout exceeded
elapsed: 181s
```

`timeout_ms` is NULL on the row (no explicit value, no column default), so the
null-default applied and killed a job that was working correctly. Nothing was
inserted; the page's facts are simply lost until something resubmits.

## Why this is worse than a slow job

The margin is razor-thin — successful runs at 122s and 128s against a cut-off
around 180s — so the same page succeeds or dies depending on model load. And the
outcome is not obviously a timeout to anyone looking at facts coverage: the job
is dead-lettered, `facts` simply does not grow, and `doctor`'s `facts_health`
reports `[OK] 0 active, 0 today`. It reads as "this page had no facts."

The exposure grows with provider latency. On a fast hosted API 180s is ample; on
a self-hosted or heavily-loaded endpoint it is routinely breached. Self-hosting is
a supported configuration, so the default should not assume hosted-API latency.

## Second, related cap

`backstop.ts:194` hardcodes `timeout_ms: 180_000` when it submits the durable job
itself:

```ts
{
  queue: 'default',
  idempotency_key: `facts-absorb:${ctx.sourceId}:${parsedPage.slug}:${contentHash}`,
  max_attempts: 3,
  timeout_ms: 180_000,
}
```

So even the intended submission path caps at 180s, and an explicit `timeout_ms`
outranks any default added to the map. Both need to move together, or adding
`facts-absorb` to `HANDLER_DEFAULT_TIMEOUT_MS` will have no effect on the jobs the
backstop actually creates.

Note also that `max_attempts: 3` does not help: each retry re-runs a generation
that will breach the same ceiling, so a slow page burns three full budgets and
still dies.

## Suggested fix

1. Add `'facts-absorb'` to `HANDLER_DEFAULT_TIMEOUT_MS` — `TEN_MIN_MS` matches
   `chronicle_extract`, which is the same one-page-one-LLM-call shape.
2. Raise or remove the hardcoded `180_000` in `backstop.ts` so it does not
   override that default; if a hard cap is wanted, derive it rather than
   hardcoding a hosted-API-latency assumption.
3. Consider making the dead-letter reason visible where facts coverage is
   reported — a page whose facts were lost to a wall-clock kill currently looks
   identical to one that genuinely had none.

## Environment

gbrain 0.42.63.0, postgres engine, self-hosted chat via the `litellm` recipe,
`facts.extraction_max_tokens` at its 4000 default.

Related: #3206 (`chat_model` overridden by the tier default) and #3062 (chat
unavailability yields a silent zero) — all three produce the same
indistinguishable "0 facts, no error" outcome from different causes.

### LINKED ISSUE #3217: Zero-token empty completions still succeed

time-attack:
Backlog fix wave nwf5 — three verified items, one commit each.

## Fixes #3207 — facts-absorb wall-clock budget
- `'facts-absorb': TEN_MIN_MS` added to `HANDLER_DEFAULT_TIMEOUT_MS` (same one-page-one-LLM-call shape as `chronicle_extract`).
- The durable submit in `src/core/facts/backstop.ts` no longer hardcodes `timeout_ms: 180_000` — the map's contract ("an explicit `opts.timeout_ms` always wins") would have neutralized a map-only fix, so the submit now inherits the stamp from `MinionQueue.add`.
- Test: `test/facts-backstop.test.ts` pins both the map value and the stamped `minion_jobs.timeout_ms` on a real short-lived-CLI durable submit.

## Fixes #3217 — contentless chat completions reported as success (takeover of #3249)
Applied #3249's fix verbatim (it was correct and complete); credit to @Masashi-Ono0611 via co-author trailer. `chat()` now rejects completions with no usable content (no non-whitespace text, no tool-call blocks), gated on CONTENT and never on reported usage. Refusal/content_filter stops stay returned, not thrown; a contentless `length` stop is an `AIConfigError` with an actionable fix; the throw carries the provider's real usage so budget accounting charges actual tokens (pessimistic fallback preserved when usage was omitted). Test file `test/ai/gateway-empty-completion.test.ts` included from #3249. Not closing #3249 here — just referencing.

## Fixes #3190 — custom schema packs were decorative for plain link extraction
All three gates opened, in the canonical layer (`src/core/link-extraction.ts`) so the db extract path, `extract --stale`, and `put_page` auto-link behave identically:
1. **Entity-dir whitelist**: `extractEntityRefs` / the bare-slug pass union `DIR_PATTERN` with the active pack's `page_types[].path_prefixes` (slug-safe charset only, longest-first so nested prefixes win).
2. **Verb inference**: `extractPageLinks` consults pack `link_types[].inference` (page-type-bound + regex under a per-page ReDoS budget) before the legacy matchers — the exact caller wrap `schema-pack/link-inference.ts` documents.
3. **frontmatter_links**: pack mappings extend `FRONTMATTER_LINK_MAP` at extraction time (outgoing direction, pack entity dirs as resolution hints). The hardcoded map wins on any `(pageType, field)` it already covers, so a pack restating a built-in field cannot emit reversed or duplicate edges.

Guard: the codegen'd base packs (`gbrain-base`, `gbrain-base-v2`) are excluded via `packLinkExtractionView` — they are generated FROM the in-code tables, and consulting their simplified sketch regexes (or their direction-less frontmatter entries) would regress default-pack behavior. `LINK_EXTRACTOR_VERSION_TS` bumped so `extract --stale` re-extracts previously-stamped pages.

Not covered (follow-up material): the filesystem walk path (`extractLinksFromFile`) keeps its own dir-based inference; the issue's repro (`extract links --source db`) and put_page are fully covered.

## Verification
- `bun run typecheck` exit 0
- `bun test` on touched + adjacent files: `link-extraction-pack`, `facts-backstop*`, `ai/gateway*`, `link-extraction`, `link-inference-pack`, `extract-db`, `extract-stale`, `extract`, `extract-fs`, `extract-incremental`, `extract-source-aware`, `reconcile-links.serial`, `doctor-links-extraction-lag`, `regressions/gbrain-base-equivalence`, `minions/*` — all pass (750+ tests, 0 fail)

### DIFF
--- a/src/commands/extract.ts
+++ b/src/commands/extract.ts
@@ -43,6 +43,8 @@ import {
 } from '../core/link-extraction.ts';
 import { createProgress } from '../core/progress.ts';
 import { getCliOptions, cliOptsToProgressOptions } from '../core/cli-options.ts';
+import { loadActivePackBestEffort } from '../core/schema-pack/best-effort.ts';
+import { packLinkExtractionView } from '../core/schema-pack/link-inference.ts';
 import { pathToSlug, pruneDir, isSyncable } from '../core/sync.ts';
 // v0.41.18.0: withRetry + isRetryableConnError + WithRetryOpts moved to
 // src/core/retry.ts as the canonical primitive. Engine methods
@@ -1381,6 +1383,9 @@ async function extractLinksFromDB(
   // Issue #972: opt-in global-basename wikilink resolution. Read once
   // per extract run; threaded into each extractPageLinks call.
   const globalBasename = await isGlobalBasenameEnabled(engine);
+  // #3190: active custom pack (null for the codegen'd base packs) so pack
+  // path_prefixes / link verbs / frontmatter_links drive extraction.
+  const pack = packLinkExtractionView(await loadActivePackBestEffort({ engine } as never));
   // v0.32.8: listAllPageRefs enumerates (slug, source_id) so we can thread
   // sourceId to getPage AND build a cross-source resolution map for link
   // disambiguation. Pre-fix used getAllSlugs() which collapsed
@@ -1461,7 +1466,7 @@ async function extractLinksFromDB(
     // basename lookup; off by default for back-compat.
     const extracted = await extractPageLinks(
       slug, fullContent, page.frontmatter, page.type, resolver,
-      { skipFrontmatter: !includeFrontmatter, globalBasename },
+      { skipFrontmatter: !includeFrontmatter, globalBasename, pack },
     );
     unresolved.push(...extracted.unresolved);
 
@@ -1687,6 +1692,9 @@ export async function extractStaleFromDB(
   const resolver = makeResolver(engine, { mode: 'batch' });
   const nullResolver = { resolve: async () => null as string | null };
   const activeResolver = includeFrontmatter ? resolver : nullResolver;
+  // #3190: same pack threading as extractLinksFromDB so `extract --stale`
+  // produces the same edges as a manual `extract links --source db`.
+  const pack = packLinkExtractionView(await loadActivePackBestEffort({ engine } as never));
   const allRefs = await engine.listAllPageRefs();
   const allSlugs = new Set<string>();
   const slugToSources = new Map<string, string[]>();
@@ -1719,6 +1727,7 @@ export async function extractStaleFromDB(
       const fullContent = page.compiled_truth + '\n' + page.timeline;
       const extracted = await extractPageLinks(
         page.slug, fullContent, page.frontmatter, page.type, activeResolver,
+        { pack },
       );
       for (const c of extracted.candidates) {
         const r = resolveCandidateSources(c, page.slug, page.source_id, allSlugs, slugToSources);
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -3207,14 +3207,66 @@ export async function chat(opts: ChatOpts): Promise<ChatResult> {
     const providerMetadata = (result as any).providerMetadata as Record<string, any> | undefined;
     const anthropicCache = providerMetadata?.anthropic ?? {};
 
-    const inTok = Number(usage.inputTokens ?? usage.promptTokens ?? 0);
-    const outTok = Number(usage.outputTokens ?? usage.completionTokens ?? 0);
+    const rawInTok = usage.inputTokens ?? usage.promptTokens;
+    const rawOutTok = usage.outputTokens ?? usage.completionTokens;
+    const inTok = Number(rawInTok ?? 0);
+    const outTok = Number(rawOutTok ?? 0);
+
+    const stopReason = mapStopReason((result as any).finishReason, providerMetadata);
+    const text = blocks
+      .filter(b => b.type === 'text')
+      .map(b => (b as { type: 'text'; text: string }).text)
+      .join('');
+
+    // A completion with no usable content must not report success (#3217).
+    // The check is on the CONTENT (text / tool-call blocks), never on reported
+    // usage — providers commonly normalize omitted usage to zero, so a
+    // usage-based gate would wave empty responses through as "free" successes
+    // that silently propagate no result to every chat() caller. Refusal and
+    // content-filter stops stay non-throwing: they are meaningful terminal
+    // signals callers branch on (toolLoop surfaces each as its own stop
+    // reason). A length stop with no content is classified distinctly — the
+    // output budget was exhausted before any text was emitted (thinking
+    // models spend it on internal reasoning first), which is deterministic
+    // for the same call, so it gets a config error with an actionable fix
+    // instead of a futile retry.
+    if (
+      text.trim().length === 0 &&
+      !blocks.some(b => b.type === 'tool-call') &&
+      stopReason !== 'refusal' &&
+      stopReason !== 'content_filter'
+    ) {
+      const label = `chat(${recipe.id}:${modelId})`;
+      const emptyErr =
+        stopReason === 'length'
+          ? new AIConfigError(
+              `${label}: output budget exhausted before any content was emitted (maxOutputTokens=${maxOutputTokens})`,
+              'Raise maxTokens for this call (or the configured max output tokens). Thinking models spend output budget on internal reasoning before emitting text.',
+            )
+          : new AITransientError(
+              `${label}: provider returned an empty completion (stopReason=${stopReason}, output_tokens=${outTok}) — no text and no tool calls`,
+            );
+      // Carry the real usage on the error so the catch-path budget record
+      // charges actual tokens instead of the pessimistic ceiling — but only
+      // the fields the provider actually reported. When usage was omitted
+      // entirely, attach nothing so _extractUsageFromError keeps its
+      // pessimistic fallback (attaching a synthesized {0,0} would record the
+      // spend as free).
+      const reportedUsage: Record<string, number> = {};
+      if (typeof rawInTok === 'number' && Number.isFinite(rawInTok)) reportedUsage.input_tokens = rawInTok;
+      if (typeof rawOutTok === 'number' && Number.isFinite(rawOutTok)) reportedUsage.output_tokens = rawOutTok;
+      if (Object.keys(reportedUsage).length > 0) {
+        (emptyErr as { usage?: unknown }).usage = reportedUsage;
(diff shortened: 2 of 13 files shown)


time-attack:
Closing. The pack-driven link-extraction work is the bulk of this and it is feature work — three new extraction gates, a base-pack exclusion view, and a `LINK_EXTRACTOR_VERSION_TS` bump that forces repo-wide re-extraction — not a fix. It is bundled with a gateway change that makes contentless completions throw for every `chat()` caller (taken over from #3249) and a one-line timeout entry. Three unrelated changes at 773 lines is not something we can review as a unit. Thank you for the thorough per-item verification notes.

PR #3288, opened 2026-07-23
fix(facts,extract,test): chat-unavailable diagnostics, link-aware timeline delimiter, R5 gateway-reset guard

### LINKED ISSUE #3059: sync re-extraction shatters timeline bullets whose text contains dashes inside markdown links — fragments accumulate in timeline_entries on every sync

### LINKED ISSUE #3062: facts extraction and query expansion silently no-op when the chat gateway is unauthenticated — jobs report success, zero diagnostics
## Summary

When no chat provider credential is present, `extractFactsFromTurn` returns `[]` on its
very first guard and every downstream surface reports **success**. A `facts-absorb`
minion job completes with `{"fact_ids": [], "inserted": 0, "duplicate": 0, "superseded": 0}` —
byte-identical to the legitimate "this page genuinely had no extractable facts" result.
Nothing is logged, no `skipped` reason is set, and `doctor`'s `facts_health` check reports
`[OK]`.

Observed on a brain where 40 queued `facts-absorb` jobs all completed against substantial
documents — a 142k-char changelog, a 23k-char `CLAUDE.md`, several reference docs — and
inserted zero facts. The feature had never worked on that host, and every health surface
said it was fine.

## Mechanism

`src/core/facts/extract.ts`:

```ts
if (!isAvailable('chat')) {
  // No chat gateway → no extraction. Caller still inserts facts via direct
  // `gbrain take add` paths.
  return [];
}
```

`isAvailable` (`src/core/ai/gateway.ts:834`) resolves the chat model, looks up the recipe,
and returns false when the recipe's `auth_env.required` keys are absent:

```ts
const required = recipe.auth_env?.required ?? [];
if (required.length === 0) return true;
return required.every(k => !!_config!.env[k]);
```

With the default chat model on the anthropic recipe, `auth_env.required` is
`['ANTHROPIC_API_KEY']`. Unset → `isAvailable('chat')` is false → silent `[]`.

`runFactsBackstop` then returns `inserted: 0` with **no** `skipped` field. The enum already
carries `'extraction_disabled'` and `` `eligibility_failed:${string}` `` — but neither is
emitted here, because this path is upstream of both.

## Why this is worse than a missing feature

The result is indistinguishable from a correct no-op, at every layer:

| Surface | What it shows | What is true |
|---|---|---|
| `minion_jobs.status` | `completed` | never called the model |
| job `result` | `inserted: 0` | same shape as a genuine empty extraction |
| `FactsBackstopResult.skipped` | absent | should say "no chat gateway" |
| `doctor` → `facts_health` | `[OK] 0 active, 0 today, 0 this week` | feature has never run |
| stderr / `ingest_log` | nothing | — |

The extractor already demonstrates the right instinct one branch below: the `stopReason ===
'length'` path writes an explicit `[facts-extract] WARN:` to stderr naming the cap and the
remedy. The unauthenticated path deserves at least that.

## Same root cause hits query expansion

`src/core/ai/gateway.ts:2268`:

```ts
if (!isAvailable('expansion')) return [query];
```

The `tokenmax` search mode's headline knob is `expansion` (LLM multi-query). On an
unauthenticated brain it silently degrades to single-query while the mode still reports
`[OK] Mode: tokenmax` in `doctor` and still pays tokenmax's larger `searchLimit` and token
budget. The operator is charged for the mode and gets none of its distinguishing behaviour.

## Suggested fix

1. Add a distinct skip reason rather than reusing the kill-switch one — e.g.
   `skipped: 'chat_unavailable'` on `FactsBackstopResult`, so the job result records *why*
   zero facts came back.
2. Log once per process (not per page) at the guard, naming the missing env key and the
   recipe's `setup_url`, matching the existing truncation-warning style.
3. Make the health surfaces honest:
   - `facts_health` should distinguish "no facts because nothing to extract" from "no facts
     because there is no chat gateway".
   - `search_mode` should flag that `expansion` is configured-on but inert.
4. Consider declining to enqueue `facts-absorb` at all when the gateway is unavailable —
   currently a sync run queues work that is guaranteed to do nothing (40 jobs/day on the
   brain above).

## Environment

- gbrain 0.42.63.0, postgres engine
- self-hosted embeddings (ollama recipe, no auth required — embeddings work fine)
- no chat provider credential configured, which is the supported state for a
  sovereignty-focused install: embeddings are self-hosted specifically to avoid a
  third-party dependency. The bug is the silence, not the absence of a key.

### LINKED ISSUE #3066: test: add CI guard — files calling configureGateway must resetGateway in afterAll (cross-file leak class, ~70 files)

time-attack:
Backlog fix wave nwf8 — three verified items.

## Fixes #3062 — facts extraction + query expansion silently no-op when the chat gateway is unauthenticated

- `runFactsBackstop` now records `skipped: 'chat_unavailable'` (both queue and inline modes) instead of returning a result byte-identical to a genuine empty extraction, and declines to enqueue `facts-absorb` jobs that are guaranteed to no-op.
- New `unavailableReason()` / `warnUnavailableOnce()` in the gateway name the configured model, the missing `auth_env.required` keys, and the recipe's `setup_url`. A once-per-process stderr warning fires at the `extractFactsFromTurn` guard and at `expand()`'s silent single-query degrade (tokenmax's headline knob). The memo is cleared by `resetGateway()`.
- `doctor` `facts_health` now distinguishes "0 active facts because the chat gateway is unreachable" (warn, with the actionable reason) from a genuinely empty brain (ok).
- Regression tests in `test/facts-extract-silent-no-op.test.ts` (the file that owns this bug class).

## Fixes #3059 — timeline bullets shatter on dashes inside markdown links (takeover of #3060)

Salvages #3060 by @wright-io (credited via Co-authored-by): the Format-1 timeline parser split `Source — Summary` at the first dash after the pipe, including hyphens inside link targets and em-dashes inside link labels; each shattered shape upserts as a new `timeline_entries` row on every sync. The delimiter scan is now bracket-depth-aware and whitespace-anchored (`findDelimiterOutsideLinks`); delimiterless bullets are kept whole under the existing `'markdown'` source sentinel instead of dropped. Changes from the original PR: test fixtures renamed to the repo's generic placeholders per the privacy rule (`alice-example`, `companies/acme-example`), including the code-comment example. 4 regression tests.

## Fixes #3066 — CI guard: files calling configureGateway must resetGateway in a teardown hook

`scripts/check-test-isolation.sh` gains rule R5: any non-serial unit test file calling `configureGateway(` or `__setEmbedTransportForTests(` must call `resetGateway(` and have an `afterAll(`/`afterEach(` hook. Two hardening details beyond the loose R4 shape: the greps run on comment-stripped lines (a prose mention of `resetGateway()` hid the `cycle-consolidate` leaker), and the trigger requires exact call syntax so test-name prose ("works WITHOUT configureGateway (reads registry)") doesn't fire it. The 11 current violators are fixed in the same PR with `resetGateway()` in their `afterAll`; 5 fixture cases in `test/scripts/check-test-isolation.test.ts` pin the rule.

## Verification

- `bun run typecheck` exit 0
- `bun test test/extract.test.ts test/facts-extract-silent-no-op.test.ts test/scripts/check-test-isolation.test.ts test/facts-backstop-gating.test.ts` — 67 pass / 0 fail
- All 11 R5-fixed test files run together: 140 pass / 0 fail
- `scripts/check-test-isolation.sh` and `scripts/check-test-real-names.sh` both clean
- The 5 failures in a combined `test/facts-*` run (vector-dimension leak) reproduce identically on clean master with the same file combination — pre-existing, unrelated

### DIFF
--- a/scripts/check-test-isolation.sh
+++ b/scripts/check-test-isolation.sh
@@ -18,6 +18,12 @@
 #  R4: any file that creates `new PGLiteEngine(` must call `.disconnect(`
 #      inside an `afterAll(` block. Without disconnect, engines leak across
 #      file boundaries within a shard process.
+#  R5: any file that calls `configureGateway(` or
+#      `__setEmbedTransportForTests(` must also call `resetGateway()` and
+#      have an `afterAll(`/`afterEach(` hook. The gateway is module-global;
+#      a configured remote provider + fake key left past the file boundary
+#      makes the next embed-triggering file in the shard fire a live HTTP
+#      call (issue #3066; master shard 6 broke twice this way).
 #
 # Scope:
 #  - Recursively scans `test/**/*.test.ts`.
@@ -130,6 +136,22 @@ while IFS= read -r f; do
     if ! grep -qE 'afterAll[[:space:]]*\(' "$f" 2>/dev/null \
        || ! grep -qE '\.disconnect[[:space:]]*\(' "$f" 2>/dev/null; then
       emit_violation "$f" "R4" "creates PGLiteEngine but missing afterAll(() => engine.disconnect()); engine leaks across files in the shard process" ""
+    fi
+  fi
+
+  # R5: gateway configuration requires resetGateway + an afterAll/afterEach
+  # hook. Same loose two-grep shape as R4 — `resetGateway(` present plus at
+  # least one afterAll(/afterEach( — but on comment-stripped lines: a prose
+  # mention like "a test that calls resetGateway()" must not satisfy the
+  # rule (that exact false pass hid the cycle-consolidate leaker).
+  # No [[:space:]]* before the paren on the trigger side: prose in a test
+  # name ("works WITHOUT configureGateway (reads registry...)") must not
+  # trigger the rule; real call sites are always `configureGateway(`.
+  r5_code=$(grep -vE '^[[:space:]]*(//|\*)' "$f" 2>/dev/null || true)
+  if printf '%s\n' "$r5_code" | grep -qE 'configureGateway\(|__setEmbedTransportForTests\('; then
+    if ! printf '%s\n' "$r5_code" | grep -qE 'resetGateway\(' \
+       || ! printf '%s\n' "$r5_code" | grep -qE 'afterAll[[:space:]]*\(|afterEach[[:space:]]*\('; then
+      emit_violation "$f" "R5" "configures the AI gateway but never calls resetGateway() in afterAll/afterEach; gateway state (provider, fake keys, transports) leaks across files in the shard process" ""
     fi
   fi
 done <<EOF
--- a/src/commands/doctor.ts
+++ b/src/commands/doctor.ts
@@ -7090,11 +7090,16 @@ export async function buildChecks(
     );
     if (factsExists[0]?.exists) {
       const health = await engine.getFactsHealth('default');
-      const status: 'ok' | 'warn' = health.total_active >= 0 ? 'ok' : 'warn';
       const top = health.top_entities
         .slice(0, 3)
         .map(t => `${t.entity_slug}:${t.count}`)
         .join(', ') || '—';
+      // #3062: "0 active facts" used to read [OK] even when the chat
+      // gateway had never been reachable — indistinguishable from a brain
+      // with genuinely nothing to extract. Distinguish the two.
+      const { unavailableReason } = await import('../core/ai/gateway.ts');
+      const chatDown = health.total_active === 0 ? unavailableReason('chat') : null;
+      const status: 'ok' | 'warn' = chatDown ? 'warn' : health.total_active >= 0 ? 'ok' : 'warn';
       checks.push({
         name: 'facts_health',
         status,
@@ -7102,7 +7107,8 @@ export async function buildChecks(
           `facts_health(default): ${health.total_active} active, ` +
           `${health.total_today} today, ${health.total_week} this week, ` +
           `${health.total_consolidated} consolidated, ` +
-          `top entities ${top}`,
+          `top entities ${top}` +
+          (chatDown ? ` — extraction has no chat gateway: ${chatDown}` : ''),
       });
     } else {
       checks.push({
--- a/src/commands/extract.ts
+++ b/src/commands/extract.ts
@@ -469,15 +469,53 @@ export async function extractLinksFromFile(
 
 // --- Timeline extraction ---
 
+/**
+ * Index of the first dash (—, –, -) that can serve as the Source — Summary
+ * delimiter: it must have whitespace on both sides and sit outside every
+ * markdown-link span. Hyphens inside link targets
+ * (`../people/alice-example.md`) and dashes inside link labels
+ * (`[Deals — Q1 Review](...)`) are content, not delimiters — splitting on
+ * them shatters one entry into two fragments whose halves re-insert on
+ * every sync (the (page_id, date, summary, source) uniqueness sees each
+ * fragment shape as a new row). Returns -1 when the line has no delimiter.
+ */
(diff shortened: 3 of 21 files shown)


time-attack:
Closing. Rule R5 plus the 11 teardown fixes is defensible test-hygiene work on its own, but it ships here alongside new gateway diagnostics surface (once-per-process stderr warnings, a new `facts_health` doctor branch) and a takeover of #3060's timeline parser — three unrelated changes touching 21 files. We are not merging bundles of that shape regardless of how sound the individual items are. The R5 comment-stripping detail was a sharp catch.

PR #3291, opened 2026-07-23
v0.42.59.0 fix(jobs): honor prune safety flags (#2712)

### LINKED ISSUE #2712: jobs prune silently ignores --dry-run and deletes rows
## Bug

`gbrain jobs prune --dry-run` executes the prune and deletes rows while the operator believes they are previewing. Observed on v0.42.53.0 (Postgres engine): the command printed `Pruned 8732 jobs older than 30 days.` and the rows were really gone (row count verified against a pre-command `pg_dump`).

## Root cause

`src/commands/jobs.ts` `case 'prune'` parses only `--older-than`:

```ts
case 'prune': {
  const olderThanStr = parseFlag(args, '--older-than') ?? '30d';
  ...
  const count = await queue.prune({ olderThan: new Date(Date.now() - days * 86400000) });
  console.log(`Pruned ${count} jobs older than ${days} days.`);
}
```

Unknown flags are silently dropped, so `--dry-run` (which `jobs submit` DOES support, teaching users to expect it) falls through and the destructive default runs.

## Expected

Either honor `--dry-run` (print the would-be count, delete nothing) or reject unknown flags on destructive subcommands with a hard error. Silent-ignore of a safety flag on a delete path is the worst of both.

## Repro

1. Any brain with job history older than 30 days.
2. `gbrain jobs prune --dry-run`
3. Rows are deleted; output claims a completed prune.

javieraldape:
Fixes #2712.

## Summary

- Make `gbrain jobs prune` parse its own supported flags strictly before touching queue state.
- Add side-effect-free handling for `--help` and `--dry-run`; dry-run now counts rows with the same terminal-status/age predicate without deleting them.
- Fail fast on unknown prune flags so mistyped safety flags cannot silently fall through to the destructive path.
- Add focused regression coverage for dry-run, help, and unknown-flag behavior.

## Tests

- `bun test test/jobs-prune-cli.test.ts` (3 pass)
- `bun test test/minions.test.ts` (176 pass)
- Local gate: GREEN; verify passed with the known platform-only operations-filter guard tolerated, focused units passed (179 pass), and diff-selected E2E resolved to the fail-closed all-set so it was skipped locally per gate policy.

### DIFF
--- a/src/commands/jobs.ts
+++ b/src/commands/jobs.ts
@@ -22,6 +22,21 @@ function hasFlag(args: string[], flag: string): boolean {
   return args.includes(flag);
 }
 
+/** Parse `--max-waiting N` from CLI args. Returns undefined if absent.
+ *  Throws on malformed input (caller should surface the error and exit).
+ *  Clamps to [1, 100] to match the queue-layer clamp in MinionQueue.add.
+ *  Exported for unit tests; the CLI handler at `jobs submit` wraps this
+ *  with process.exit(1) on throw so operators see 'must be positive integer'. */
+export function parseMaxWaitingFlag(args: string[]): number | undefined {
+  const raw = parseFlag(args, '--max-waiting');
+  if (raw === undefined) return undefined;
+  const parsed = parseInt(raw, 10);
+  if (!Number.isFinite(parsed) || parsed < 1) {
+    throw new Error('--max-waiting must be a positive integer (will be clamped to [1, 100])');
+  }
+  return Math.max(1, Math.min(100, parsed));
+}
+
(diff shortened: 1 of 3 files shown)


time-attack:
Merging. Silently ignoring a safety flag on a destructive delete path is exactly the class of bug we take without hesitation, and I confirmed it is still live on master — `case 'prune'` parses only `--older-than` and drops everything else on the floor. The fix is well-scoped: a read-only `countPrunable()` next to the existing delete, hard rejection of unknown flags on the destructive subcommand, and tests that prove the row survives a `--dry-run`. Thank you for the clean repro and the tight diff.

Issue #1890, opened 2026-06-05
conversation_parser.llm_fallback_enabled / llm_polish are dead config (never read; doctor recommends the no-op)


ElliotDrel:
## Summary

`conversation_parser.llm_fallback_enabled` (and its sibling `conversation_parser.llm_polish`) is **dead config**: setting it has no effect because runtime never reads it, the key isn't registered, and the LLM functions that would consume it are never called. Worse, `gbrain doctor` actively *recommends* setting it, so users enable a no-op and assume LLM fallback is now active.

Observed on **gbrain 0.42.24.0** (HEAD `935fc2a7`). `gbrain config set conversation_parser.llm_fallback_enabled true --force` succeeds but the flag changes nothing.

## Root cause (verified against source)

1. **Key not registered.** `conversation_parser.llm_fallback_enabled` (and `conversation_parser.llm_polish`) do not appear anywhere in `src/core/config.ts`. So `config set` writes it as an unknown key (hence `--force` is required), and no validated config surface exposes it.

2. **Live parser is sync + regex-only.** `parseConversation()` (`src/core/conversation-parser/parse.ts:436`) returns `ParseResult` synchronously and only scores/selects regex pattern candidates. There is no branch that consults the flag or falls back to an LLM when zero messages match.

3. **The fallback function is dead code.** `runLlmFallback()` (`src/core/conversation-parser/llm-fallback.ts:68`) is exported but has **zero callers** in `src/` (grep: the only hit is its own definition). The live paths — `conversation-parser scan` and the conversation-fact extractor — call the regex parser directly and never invoke it.

4. **Same for polish.** `runLlmPolish()` (`src/core/conversation-parser/llm-polish.ts:88`) is likewise exported but never wired into any live parse/scan/extract path, and `llm_polish` is also absent from `config.ts`.

5. **Doctor recommends the no-op.** `src/commands/doctor.ts:4529`, in the `conversation_format_coverage` warn branch, tells users to run `gbrain config set conversation_parser.llm_fallback_enabled true` — i.e. it points users straight at the dead flag, implying LLM fallback is an available remedy when it does nothing.

## Impact

- Users hit the `conversation_format_coverage` warning, follow doctor's advice, set the flag, and silently get no behavior change — the parser still returns `no_match` on unrecognized formats.
- The scaffolding (`runLlmFallback` / `runLlmPolish` / the LLM base helpers) gives the impression the feature exists when it's unreachable.

## Suggested fix (one of)

[...]

PR #3292, opened 2026-07-23
v0.42.64.3 feat(conversation-parser): wire the opt-in LLM fallback

### LINKED ISSUE #2247: conversation-parser LLM fallback is unwired: `runLlmFallback` has no callers, but `doctor` recommends enabling a no-op `conversation_parser.llm_fallback_enabled` flag
## Summary

The conversation-parser **LLM fallback** is implemented and unit-tested but **never wired into the parse path** — yet `gbrain doctor` actively tells users to enable it via a config key that nothing reads. The advice is a no-op.

Verified on `master` @ **v0.42.47.0**.

## Evidence

- `src/core/conversation-parser/llm-fallback.ts` exports `runLlmFallback(...)`, but it has **zero production callers** — the only references in the repo are its definition and `test/conversation-parser/llm-fallback.test.ts`:
  ```
  $ git grep -n runLlmFallback -- 'src/**'
  src/core/conversation-parser/llm-fallback.ts:68:export async function runLlmFallback(
  ```
  No module imports `llm-fallback` except the test.
- `src/core/conversation-parser/parse.ts:430` confirms it: `// LLM fallback (T4; not yet wired).` The `noFallback` option in `parseConversation` gates the **regex** full-body re-score fallback, not an LLM path.
- `src/commands/doctor.ts:4768` (the `conversation_format_coverage` check) recommends:
  > Enable LLM fallback (opt-in): `gbrain config set conversation_parser.llm_fallback_enabled true`
- But `conversation_parser.llm_fallback_enabled` is **read by nothing** — the only references are doc comments (`llm-fallback.ts:6`, `llm-base.ts:11`). `gbrain config set` rejects it as unknown and, with `--force`, warns: *"Nothing in gbrain reads this."*

## Impact

A user whose `conversation_format_coverage` is flagged follows the doctor's advice, sets the flag, and nothing changes — there's no wired code path that consumes it. (Separately, the metric measures regex coverage with `noFallback:true`, so an LLM fallback wouldn't move it even if wired — but that's a second-order point.)

## Repro

1. `gbrain doctor` on a brain with >10% unmatched conversation pages → warns, recommends the flag.
2. `gbrain config set conversation_parser.llm_fallback_enabled true` → `[config] ... Unknown config key`; with `--force` → `WARN: ... Nothing in gbrain reads this.`
3. `git grep -n runLlmFallback -- 'src/**'` → only the definition.

## Suggested fix (one of)

- **Wire it:** call `runLlmFallback` from `parseConversation`'s no-match path, gated by `conversation_parser.llm_fallback_enabled`, and register the config key in the schema.
- **Or stop recommending it:** until T4 lands, drop the "enable LLM fallback" hint from the `conversation_format_coverage` doctor message so it doesn't point users at a no-op flag.

danwiggins:
## Summary

- wire the existing conversation-parser LLM fallback into conversation fact extraction behind the exact, default-off `conversation_parser.llm_fallback_enabled=true` privacy gate
- preserve deterministic parsing as the first and preferred path; dry runs never call a provider
- parse the full unmatched body through cached 100-line windows with bounded overlap, one-to-one exact-first deduplication, and complete-body preference at boundaries
- make page date metadata part of the prompt and cache identity, validate strict zoned RFC3339 output, canonicalize UTC timestamps, and stable-sort before segmentation
- reject truncated, refused, filtered, tool-call, and otherwise non-terminal model results before parsing or caching
- propagate caller cancellation and hard budget stops while ordinary provider failures remain fail-open; surface post-call budget overages and `pages_llm_fallback`
- add an operator/maintainer runbook covering privacy, dispatch, provider data, caching, timestamps, failure modes, observability, and test contracts

## Privacy boundary

This remains opt-in. Unmatched transcript text reaches the configured utility model only when the exact config key is set to `true`. Deterministic matches, disabled/default configuration, and dry runs do not use the fallback provider. Transcript content is explicitly treated as untrusted data in the prompt.

## Verification

- `bun test test/conversation-parser test/extract-conversation-facts.test.ts test/extract-conversation-facts-workers.test.ts test/config-set.test.ts` (211 pass)
- `bun x tsc --noEmit`
- `bun run build` (1,729 modules, compiled executable)
- `bun run build:llms`
- `bun run check:conversation-parser` (17/17 fixtures, recall 1.000)
- fixture privacy guard
- KEY_FILES current-state guard
- independent coverage and adversarial reviews: no remaining blockers

No schema migrations.

Fixes #2247
Addresses #1890
Supersedes #2702


time-attack:
Superseded with credit by #3371 (just merged) — your change landed rebased onto current master with authorship preserved via Co-authored-by. Thank you!

PR #3311, opened 2026-07-23
reland: fix(frontmatter): derive validate slug from brain root, not absolute path (#2340)

### LINKED ISSUE #565: frontmatter validate --fix derives wrong path-prefixed slug when invoked via CLI vs daemon
Running gbrain frontmatter validate <path> --fix --dry-run --json produces an "expected" slug that includes the absolute filesystem path prefix, while the same SLUG_MISMATCH error surfaced from the live sync daemon shows the brain-root-relative slug.
CLI invocation output:
does not match path-derived slug "root/projects/virgil-sean/context/priorities/active/<file>"
Daemon (journalctl -u gbrain-proxy.service) output for the same file:
does not match path-derived slug "context/priorities/active/<file>"
Environment:

gbrain v0.22.9 (source at /root/gbrain/)
Brain root managed by the daemon: /root/projects/virgil-sean/
Daemon invocation: mcp-proxy --host 127.0.0.1 --port 8787 -- /root/.bun/bin/gbrain serve
CLI invocation tried: both gbrain frontmatter validate /root/projects/virgil-sean/context --fix --dry-run --json and gbrain frontmatter validate context --fix --dry-run --json from inside /root/projects/virgil-sean/. Both produce the same absolute-path-prefixed expected slug.

Impact: Running --fix without --dry-run would write ~142 files with slugs like root/projects/virgil-sean/context/..., tying them to one machine's literal filesystem path and breaking every other consumer of the brain (Worker, MCP clients, downstream sync targets).
Question: Is there a flag or env var that tells the CLI which brain-root to derive against, or is this a CLI/daemon parity bug? If the latter, happy to test a patch.
Context: Hit this trying to repair 281 SLUG_MISMATCH files from the v0.22.4 frontmatter-guard rollout. Sync is unblocked via --skip-failed; the broken files remain acknowledged-as-failed pending the right --fix invocation.

time-attack:
Relands #2340 (squash commit 1a9ab6a9), which was reverted in 47d7e95b as part of a batch auto-revert when the batch went red on master CI — not because of this change.

## Why this is innocent
- Clean cherry-pick of the original squash commit onto post-revert master (no conflicts).
- `bun run typecheck` — exit 0.
- `bun test` on the new regression test plus all frontmatter-adjacent suites (`frontmatter-validate-slug-565`, `frontmatter-cli`, `frontmatter-install-hook`, `frontmatter-inference`, `frontmatter-non-string-fields`, `doctor-frontmatter-partial`, `lint-frontmatter`) — 87 pass / 0 fail.

## Original change (unchanged)
Single-file `frontmatter validate` derived the expected slug from the absolute path (`relative(resolve(target), file) || file`), yielding bogus `root/<abs>` slugs and a false SLUG_MISMATCH — which the install-hook pre-commit hook hits on every commit. Fix walks up to the brain root (nearest `.git`) and uses `relative(brainRoot, file) || basename(file)`, matching runAudit/runGenerate. Fixes #565.

Credit: original author @alessioalionco (authorship preserved via cherry-pick).

### DIFF
--- a/src/commands/frontmatter.ts
+++ b/src/commands/frontmatter.ts
@@ -17,7 +17,7 @@
 
 import { readFileSync, writeFileSync, existsSync, lstatSync, readdirSync } from 'fs';
 import { setCliExitVerdict } from '../core/cli-force-exit.ts';
-import { join, relative, resolve } from 'path';
+import { join, relative, resolve, basename, dirname } from 'path';
 import type { BrainEngine } from '../core/engine.ts';
 import { loadConfig, toEngineConfig } from '../core/config.ts';
 import { createEngine } from '../core/engine-factory.ts';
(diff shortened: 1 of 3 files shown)


time-attack:
Rebased onto current master (clean cherry-pick, no conflicts) and fixed the CI failure from the last run.

The test (1) shard failure (`findCandidateDuplicates > embedding cosine ordering` — `expected 1280 dimensions, not 1536`) was the known shard-order 1280/1536 dims flake, same class as #2801: `facts-engine.test.ts` hardcodes 1536-dim vectors but let `initSchema` size vector columns from process-global gateway state. Adding `test/frontmatter-validate-slug-565.test.ts` reshuffled the weight-packed shards and exposed it. Fixed with the established pattern: `configureGateway(1536)` in `beforeAll` before `initSchema`, `resetGateway()` in `afterAll`.

Gates re-run locally on the rebased branch: `bun run typecheck` exit 0; `bun test` on facts-engine + all frontmatter-adjacent suites — 101 pass / 0 fail.

PR #3348, opened 2026-07-24
feat: per-call source_id for get_page/list_pages, plus list_pages offset

boundless-forest:
## Summary

Adds optional `source_id` narrowing and deterministic `offset` pagination to the page-read operations, complementing the whoami source grant introspection merged in #3332.

### Changes

**`get_page` — per-call source narrowing**
- New optional `source_id` parameter, routed through `federatedSearchScope` for grant-aware resolution.
- Exact and fuzzy reads respect the scope; cross-source lookups return `page_not_found`.

**`list_pages` — source narrowing + offset pagination**
- New optional `source_id` parameter (same grant semantics as `get_page`).
- New optional `offset` parameter (nonnegative safe integer, clamped in handler).
- Enables deterministic 100+1 overflow probing for remote inventory clients.

### Contract tests

10 new tests in `test/page-read-pagination-contract.test.ts`:
- Exact match in requested source returns the page
- Exact match in wrong source returns page_not_found
- Unqualified read sees federated cross-source pages
- Source-scoped listing (both sources)
- Offset (skip first, skip 2+limit 1, beyond total → empty, default 0)
- Metadata-only output (no body/prose leak)

### Verification

```
bun run typecheck: pass
bun run build: pass
82 relevant tests: all pass
```

### DIFF
--- a/src/core/operations.ts
+++ b/src/core/operations.ts
@@ -737,11 +737,15 @@ export interface Operation {
 
 const get_page: Operation = {
   name: 'get_page',
-  description: 'Read a page by slug (supports optional fuzzy matching). Soft-deleted pages are hidden by default; pass include_deleted: true to surface them with deleted_at populated (see v0.26.5 recovery window).',
+  description: 'Read a page by slug (supports optional fuzzy matching and per-call source narrowing). Soft-deleted pages are hidden by default; pass include_deleted: true to surface them with deleted_at populated (see v0.26.5 recovery window).',
   params: {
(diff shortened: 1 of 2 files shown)


time-attack:
Thank you for the careful work here — the grant-aware routing through `federatedSearchScope` and the ten contract tests are well constructed. That said, this widens the parameter contract on two of the most central read operations to add narrowing and a pagination axis that nothing today is blocked on, and we are holding a hard line against new operation surface that isn't closing a real gap. Closing as feature surface we're not taking; no reflection on the quality of the patch.

Issue #919, opened 2026-05-12
native-openai recipe ignores provider_base_urls for expansion/embedding/chat


sheplergeraldo302-crypto:
## Bug Description

The `native-openai` recipe implementation in `src/core/ai/gateway.ts` does **not** read `provider_base_urls[\"openai\"]` when instantiating models for expansion, embedding, or chat. This causes `createOpenAI()` to always use the default `api.openai.com` endpoint, ignoring any custom base URL configured by the user.

## Affected Code

Three locations in `src/core/ai/gateway.ts`:

**Line ~677 (embedding):**
```typescript
const client = createOpenAI({ apiKey });
// Missing: baseURL: cfg.base_urls?.["openai"]
```

**Line ~1129 (expansion):**
```typescript
return createOpenAI({ apiKey }).languageModel(modelId);
// Missing: baseURL: cfg.base_urls?.["openai"]
```

**Line ~1332 (chat):**
```typescript
return createOpenAI({ apiKey }).languageModel(modelId);
// Missing: baseURL: cfg.base_urls?.["openai"]
```

In contrast, the `openai-compatible` implementation correctly reads the base URL via `applyOpenAICompatConfig(recipe, cfg)`.

## Impact

Users who want to use OpenAI-compatible endpoints (e.g., DashScope coding endpoint, Azure OpenAI, LiteLLM) **with the `openai:` model prefix** cannot do so. The custom base URL is silently ignored, and requests go to `api.openai.com`, causing timeouts or auth failures.

## Reproduction Steps

1. Configure `provider_base_urls: { openai: "https://coding.dashscope.aliyuncs.com/v1" }`
2. Set `expansion_model: "openai:qwen3.6-plus"` and `OPENAI_API_KEY`
3. Run `gbrain query "test"`

**Expected:** Request goes to `coding.dashscope.aliyuncs.com`
**Actual:** Request goes to `api.openai.com` → timeout (401 or 403 if key is rejected)

## Suggested Fix

Add `baseURL: cfg.base_urls?.["openai"]` to all three `createOpenAI()` calls. This makes `native-openai` consistent with how `openai-compatible` handles custom endpoints, while preserving the default `api.openai.com` when no override is set.

## Environment

- gbrain version: 0.32.0
- OS: macOS 26.2
- Engine: PostgreSQL + pgvector

haocn-ops:
This looks like the right bug boundary: the configured provider base URL needs to be applied at every client construction site, not just in one chat path.

[...]

PR #3351, opened 2026-07-24
feat(gateway): resolveNativeBaseUrl falls back to the config plane

### LINKED ISSUE #3350: resolveNativeBaseUrl reads only the env var, not provider_base_urls (config-plane half of #919)
**What happened?**

`resolveNativeBaseUrl` in `src/core/ai/gateway.ts` reads only the `*_BASE_URL`
env var (`ANTHROPIC_BASE_URL` / `OPENAI_BASE_URL`). It does not read
`provider_base_urls.<provider>` from `config.json`. So for the native
`anthropic:` / `openai:` recipes, a base URL set only in config.json is silently
ignored and requests go to the default `api.anthropic.com` / `api.openai.com`.

This is the config-plane half of #919. That issue was closed as completed by
adding `resolveNativeBaseUrl`, but the fix wired the env var, not the
`provider_base_urls` config field the original report (and its repro) was about,
so the config.json path is still unaddressed for native providers.

It bites hardest in contexts with no exported env var: a `gbrain serve` systemd
unit, launchd, or an MCP server that never sourced the shell that set
`ANTHROPIC_BASE_URL`. `config.json` looks complete, but the native provider
still can't reach the configured endpoint.

**What did you expect?**

`provider_base_urls.<provider>` in config.json to be used as a fallback when the
env var is absent, the same way `buildGatewayConfig` already folds
`anthropic_api_key` / `openai_api_key` / `voyage_api_key` from config into the
gateway env (env wins when present, config is the fallback for env-less
contexts).

**Steps to reproduce**
1. Set `provider_base_urls: { "anthropic": "https://<compatible-endpoint>/v1" }` in `~/.gbrain/config.json`, and `anthropic_api_key`. Do not export `ANTHROPIC_BASE_URL`.
2. Set a native model, e.g. `chat_model: "anthropic:<model>"`.
3. Run a chat/expansion call (or `gbrain models doctor`).

Expected: request targets the configured endpoint. Actual: request targets `api.anthropic.com` (env var absent, config plane ignored).

**Environment**
- gbrain version: 0.42.65.0
- OS: aarch64 Ubuntu 24.04
- Bun version: 1.3.14

```json
(not illustrative here: the failure is a silent base-URL fallthrough, not a doctor-reported fault. Repro above is the reliable signal.)
```

oliver-mee:
## Summary

`resolveNativeBaseUrl` only read the `*_BASE_URL` env var, so a native
`anthropic:` / `openai:` provider could not pick up a base URL set only in
`config.json` (`provider_base_urls.<provider>`). Env-less contexts (a `gbrain
serve` systemd unit, launchd, an MCP server that never sourced the shell) could
not reach a configured compatible endpoint even though config.json looked
complete.

This falls back to `cfg.base_urls[provider]` when the env var is absent, keeping
the existing `/v1` normalisation for both sources. Env still wins when present,
mirroring the config-as-fallback rationale `buildGatewayConfig` already applies
to API keys. It completes the config-plane path #919 described (that issue was
closed by wiring the env var, not `provider_base_urls`).

## Linked issue

Fixes #3350

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Breaking change
- [ ] Documentation
- [ ] CI, tooling, or config

## How to test

1. `bun test test/ai/resolve-native-base-url.test.ts` (11 pass). The suite now includes the config-plane fallback cases the fix adds: falls back to `cfg.base_urls[provider]` when the env var is unset, a real env value wins over config, an empty env var falls through to config, and the openai path, all with `/v1` normalisation. The pre-existing env-only cases are unchanged.
2. Manually: set `provider_base_urls.anthropic` in config.json with no `ANTHROPIC_BASE_URL` exported, use `chat_model: "anthropic:<model>"`, and confirm the request targets the configured endpoint.

## Checklist

- [x] I ran it and verified end to end. `bun run verify` (31/31 checks green), the unit test (11 pass), and `check-privacy.sh`. Did not run the DB e2e suite (needs DATABASE_URL); this is a pure config-resolution function covered by unit tests.

## Model used

Claude Opus 4.8 (1M context) via Claude Code, used to sanitise the comment, write the PR/issue, and run the test. The one-line logic change is the author's own patch.

---

### DIFF
--- a/node_modules
+++ b/node_modules
@@ -1,0 +1,1 @@
+/home/ubuntu/gbrain-fork/node_modules
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -445,7 +445,16 @@ export function resolveNativeBaseUrl(
   cfg: AIGatewayConfig,
 ): string | undefined {
   const envKey = provider === 'anthropic' ? 'ANTHROPIC_BASE_URL' : 'OPENAI_BASE_URL';
-  const raw = cfg.env[envKey];
+  // Fall back to the config plane (provider_base_urls.<provider>, routed into
+  // cfg.base_urls by buildGatewayConfig) when the *_BASE_URL env var is absent or
+  // empty, so a native provider can target an OpenAI/Anthropic-compatible endpoint
+  // from config.json alone, with no dependence on an exported env var. This is the
+  // same fallback rationale buildGatewayConfig already applies to API keys: env
+  // wins when it carries a real value, but daemon/launchd/MCP contexts that never
+  // sourced the shell (or that inject an empty ANTHROPIC_BASE_URL) still resolve
+  // from config.json.
+  const envRaw = cfg.env[envKey];
+  const raw = envRaw && envRaw.trim() ? envRaw : cfg.base_urls?.[provider];
   if (!raw || !raw.trim()) return undefined;
   const trimmed = raw.trim().replace(/\/+$/, '');
   return /\/v1$/.test(trimmed) ? trimmed : `${trimmed}/v1`;
--- a/test/ai/resolve-native-base-url.test.ts
+++ b/test/ai/resolve-native-base-url.test.ts
@@ -7,7 +7,7 @@
  * makes the SDK POST <base>/messages → 404. resolveNativeBaseUrl normalizes a configured
  * base URL to carry /v1, and returns undefined when unset so the SDK default is preserved.
  *
- * Pure function over cfg.env — no process.env mutation, so no withEnv() needed.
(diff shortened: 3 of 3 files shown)


time-attack:
Thanks for the writeup and for extending the unit suite — the env-empty-falls-through case is a nice detail. Two blockers though: the diff commits a `node_modules` symlink pointing at `/home/ubuntu/gbrain-fork/node_modules`, which would break every other checkout, and `cfg.base_urls` is keyed by recipe id everywhere else in `gateway.ts`, so indexing it by bare provider name mixes two keyspaces on one map. The documented `*_BASE_URL` env path already reaches these providers, so we're closing rather than taking a config-plane duplicate route.

Issue #2662, opened 2026-07-07
voyage_api_key in config.json is not folded into gateway env; multimodal requires manual VOYAGE_API_KEY export


amtagrwl:
## Bug

`~/.gbrain/config.json` can contain `voyage_api_key`, but the AI gateway does not fold it into `VOYAGE_API_KEY`. As a result, multimodal/image embedding with Voyage fails unless the caller manually exports `VOYAGE_API_KEY` in the process environment.

## Evidence from v0.42.56.0

- `src/core/ai/build-gateway-config.ts` folds only:
  - `openai_api_key` → `OPENAI_API_KEY`
  - `anthropic_api_key` → `ANTHROPIC_API_KEY`
  - `zeroentropy_api_key` → `ZEROENTROPY_API_KEY`
- It does **not** fold:
  - `voyage_api_key` → `VOYAGE_API_KEY`
- `src/core/brain-score-recommendations.ts` explicitly excludes `VOYAGE_API_KEY` from `HOSTED_EMBED_KEY_CONFIG`, with a comment saying this is because `buildGatewayConfig` does not thread it today.
- `gbrain providers list` reports Voyage missing when `voyage_api_key` is present only in config; with `VOYAGE_API_KEY` exported from the same config value, Voyage reports ready.

## Why this matters

Image-bearing sources need multimodal embeddings/OCR during sync. For launchd/daemon/MCP contexts, config-file keys are the durable mechanism; requiring an extra process-env export makes Voyage behave differently from OpenAI/Anthropic/ZeroEntropy and leads to silent or confusing image-pipeline failures.

## Expected fix

Map config-plane `voyage_api_key` into the gateway env just like ZeroEntropy:

```ts
if (c.voyage_api_key) envFromConfig.VOYAGE_API_KEY = c.voyage_api_key;
```

Then update the recommendation/doctor key map and tests so config-keyed Voyage is treated as usable by the same runtime that executes embeds.

## Workaround

Every wrapper that can import/sync image-bearing sources must currently do something like:

```bash
export VOYAGE_API_KEY="$(python3 -c 'import json,os;print(json.load(open(os.path.expanduser("~/.gbrain/config.json"))).get("voyage_api_key",""))')"
```

time-attack:
[...]

PR #3353, opened 2026-07-24
feat(config): fold llama_server_api_key into gateway env

### LINKED ISSUE #3352: config.json llama_server_api_key is never folded into the gateway env
**What happened?**

`config.json`'s `llama_server_api_key` is accepted at the file plane (it is a
valid `GBrainConfig` field) but `buildGatewayConfig` never folds it into the
gateway env. So the `llama-server` recipe's optional `LLAMA_SERVER_API_KEY` auth
env is only reachable via a raw process env var. A user who sets the key only in
`config.json`, pointed at an auth-gated llama-server embed proxy (a bearer-gated
reverse proxy in front of the embed endpoint), gets a 401: the config-plane key
never reaches the recipe.

This is the same class as #2662 (voyage_api_key), which was fixed by adding the
fold. `openai_api_key`, `anthropic_api_key`, `zeroentropy_api_key`,
`openrouter_api_key`, and `voyage_api_key` are all folded in `buildGatewayConfig`;
`llama_server_api_key` is the one sibling still missing.

It bites hardest in env-less contexts (a `gbrain serve` systemd unit, launchd, an
MCP server) that never sourced the shell that would export `LLAMA_SERVER_API_KEY`.

**What did you expect?**

`llama_server_api_key` folded into the gateway env alongside the other
`*_api_key` folds, with process env still winning when present.

**Steps to reproduce**
1. Set `llama_server_api_key` in `~/.gbrain/config.json`, no `LLAMA_SERVER_API_KEY` exported.
2. Point `provider_base_urls["llama-server"]` at an auth-gated embed proxy.
3. Run an embed in a daemon/MCP context.

Expected: the bearer reaches the proxy. Actual: 401, the config-plane key was dropped.

**Environment**
- gbrain version: 0.42.65.0
- OS: aarch64 Ubuntu 24.04
- Bun version: 1.3.14

```json
(not illustrative: silent env-fold gap, not a doctor-reported fault. Repro above is the signal.)
```

oliver-mee:
## Summary

`config.json`'s `llama_server_api_key` was accepted at the file plane but never
folded into the gateway env, so the `llama-server` recipe's optional
`LLAMA_SERVER_API_KEY` was only reachable via a raw process env var. A
config-plane key aimed at an auth-gated llama-server embed proxy then 401s in
env-less contexts (a `gbrain serve` systemd unit, launchd, MCP) despite
config.json looking complete.

This folds it into the gateway env alongside the existing
openai/anthropic/zeroentropy/openrouter/voyage folds. Process env still wins via
the later spread; this is the file-plane fallback. Same class as #2662 (voyage).

## Linked issue

Fixes #3352

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Breaking change
- [ ] Documentation
- [ ] CI, tooling, or config

## How to test

1. `bun test test/ai/build-gateway-config.test.ts` (17 pass; includes two new tests: llama_server_api_key folds into `LLAMA_SERVER_API_KEY`, and a real env value wins over the config-plane fallback, mirroring the voyage tests).
2. Manually: set `llama_server_api_key` in config.json with no `LLAMA_SERVER_API_KEY` exported and confirm the llama-server recipe receives it.

## Checklist

- [x] I ran it and verified end to end. `bun run verify` (31/31 checks green) and the unit tests (17 pass). Did not run e2e (needs DATABASE_URL); the change is a config-env fold covered by unit tests.

## Model used

Claude Opus 4.8 (1M context) via Claude Code, used to sanitise the comment, add the two mirroring tests, and write the PR/issue. The fold itself is the author's patch.

---

### DIFF
--- a/src/core/ai/build-gateway-config.ts
+++ b/src/core/ai/build-gateway-config.ts
@@ -44,6 +44,12 @@ export function buildGatewayConfig(c: GBrainConfig): AIGatewayConfig {
   // multimodal/image embeds despite config.json looking complete. process.env
   // still wins via the later spread.
   if (c.voyage_api_key) envFromConfig.VOYAGE_API_KEY = c.voyage_api_key;
+  // Bearer for an auth-gated llama-server embed proxy (e.g. a reverse proxy
+  // fronting the embed endpoint). The recipe declares LLAMA_SERVER_API_KEY as an
+  // optional auth env; without this fold the only channel was a raw env var, so a
+  // config-plane URL pointed at an auth-gated endpoint 401s. Mirrors the
+  // openai/anthropic/zeroentropy folds above.
+  if (c.llama_server_api_key) envFromConfig.LLAMA_SERVER_API_KEY = c.llama_server_api_key;
 
   // v0.32 codex finding #4+#5 fix: thread local-server _BASE_URL env vars
   // into base_urls so the gateway hits the user's configured port. Without
--- a/src/core/config.ts
+++ b/src/core/config.ts
@@ -63,6 +63,14 @@ export interface GBrainConfig {
    * config.json file-plane route is wired through today.
    */
   voyage_api_key?: string;
+  /**
+   * llama-server (llama.cpp) API key. Needed once the embed endpoint is
+   * fronted by an auth-gated reverse proxy (e.g. a bearer-gated cloudflared
+   * URL) rather than a local no-auth port. Folds into the gateway env as
+   * LLAMA_SERVER_API_KEY (see buildGatewayConfig), which the llama-server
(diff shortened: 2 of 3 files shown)


time-attack:
Thanks for the patch and for mirroring the voyage tests. One correction on the premise: `llama_server_api_key` isn't accepted at the file plane on master today — the only reference anywhere in `src/` is the recipe's optional env declaration, so this adds a new config key and its registration rather than wiring up a dead one. Since `LLAMA_SERVER_API_KEY` already reaches the recipe via the environment, we're closing this as new config surface we're not taking.

PR #3358, opened 2026-07-24
feat(ai): configurable output-language policy for curation phases (#3357)

### LINKED ISSUE #3357: Feature: configurable output-language policy for dream curation phases (non-English captures drift to English)
**Summary**

The dream curation phases (`extract_facts`, `propose_takes`, `synthesize`, …) have no output-language policy, so on non-English captures the model tends to emit **English** claim text — losing the source-language framing. Requesting a config-level knob to pin curation output to the source language (or an explicit language).

**Current behavior**

`EXTRACTOR_SYSTEM` (`src/core/facts/extract.ts`) says *"Capture user statements verbatim where possible. Do not paraphrase tone."* — but there's no explicit output-language rule. With a current Claude model, a Korean-language capture yields **English** fact claims (anonymized example: a Korean capture produces `"<Person> is targeting a partial exit …"` in English rather than the Korean original). Same for `propose_takes` and `synthesize` prose.

**Why it matters**

For bilingual brains (e.g. Korean/English), the source-language wording carries nuance, tone, and searchability that a silent translation drops. Proper nouns survive "verbatim," but the surrounding claim text doesn't.

**Constraints today**

- Phase prompts are hardcoded TS constants with no injection hook.
- Schema packs (`schema-pack/base/*.yaml`) are schema-only (page_types/path_prefixes) — no per-phase prompt/guidance field.
- So there's no supported way to add a language directive short of patching source.

**Proposed**

A DB-plane config key (like the `models.tier.*` pattern): `dream.curation.output_language: "source" | "<bcp47>"` (default `"source"` = match the detected capture language; an explicit code forces a language), optionally per-phase. Implementation could append a short directive to the phase system prompts at build time — prompts stay hardcoded, policy is config-driven. Mirrors the "config layer should drive/surface behavior" theme in #2689.

**Related:** #2689.

Happy to send a PR if you're open to the approach — wanted to align on the config shape first.

howardpark:
## Summary

Implements the configurable output-language policy requested in #3357.

Curation phases (`extract_facts`, `propose_takes`, `synthesize_concepts`) carried no output-language rule beyond *"capture verbatim where possible"*, so on non-English captures the model drifts to English — losing the source-language framing of a bilingual brain.

## Change

New DB-plane config `dream.curation.output_language`:
- `"source"` (default) — match the input's detected dominant language
- `"<lang>"` — force a language (e.g. `ko`, `en`, `ja`)
- `"off"` / `"none"` — prior behavior (no directive)

`src/core/ai/curation-language.ts` (new) holds the config read + a dependency-free ratio-based dominant-language detector + the directive builder. Each phase appends the directive to its system prompt — prompts stay hardcoded, only the policy is config-driven. Registered in `KNOWN_CONFIG_KEYS`.

**Backward-compatible:** default `"source"` leaves English captures unchanged; only non-English output changes (the fix). Fail-open — a missing/unavailable config never breaks a phase.

## Validation

`test/curation-language.test.ts` — 9 cases (detection ko/en/stray/empty; directive source/off/forced/null-engine/throwing-engine). Empirically, a Korean-language capture that previously produced English fact claims now produces Korean, with proper nouns preserved verbatim.

## Notes

- Left `VERSION`/`CHANGELOG` to your `/ship` flow to avoid conflicts.
- `synthesize`'s subagent-prose path is a follow-up (gated by default via `session_corpus_dir`); happy to extend it in this PR if you'd prefer.

Closes #3357.

### DIFF
--- a/src/core/ai/curation-language.ts
+++ b/src/core/ai/curation-language.ts
@@ -1,0 +1,73 @@
+/**
+ * Configurable output-language policy for dream curation phases (issue #3357).
+ *
+ * gbrain's phase prompts (EXTRACTOR_SYSTEM, propose_takes, synthesize, …) carry
+ * no output-language rule beyond "capture verbatim where possible", so on
+ * non-English captures the model tends to emit English claim text — losing the
+ * source-language framing of a bilingual brain.
+ *
+ * This appends a short directive to the phase system prompt, driven by the
+ * `dream.curation.output_language` config (DB plane, like `models.tier.*`):
+ *   - unset / "source" (default): match the detected dominant language of the input.
+ *   - "off" / "none":             no directive (prior native behavior).
+ *   - a language code ("ko","en","ja","zh",…): force that language.
+ *
+ * Prompts stay hardcoded; only the policy is config-driven — no fork of the
+ * prompt text, and default behavior is unchanged for English users.
+ */
+
+const LANG_NAMES: Record<string, string> = {
+  ko: 'Korean',
+  en: 'English',
+  ja: 'Japanese',
+  zh: 'Chinese',
+};
+
+/**
+ * Cheap, dependency-free dominant-language detection by script. Hangul-heavy
+ * text → 'ko', otherwise 'en'. Intentionally conservative (a small amount of
+ * Korean in mostly-Latin text still reads as 'ko', matching how a bilingual
+ * user writes: Korean prose with Latin brand/entity names).
+ */
+export function detectDominantLanguage(text: string): string {
+  let hangul = 0;
+  let latin = 0;
+  for (const ch of text) {
+    const c = ch.codePointAt(0) ?? 0;
+    if ((c >= 0xac00 && c <= 0xd7a3) || (c >= 0x3130 && c <= 0x318f)) hangul++;
+    else if ((c >= 0x41 && c <= 0x5a) || (c >= 0x61 && c <= 0x7a)) latin++;
+  }
+  // Ratio-based so it works for short and long inputs alike (a small absolute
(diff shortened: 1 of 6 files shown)


time-attack:
Thank you for the thorough implementation and for the nine-case test file — the fail-open design and the backward-compatible `source` default show real care. This is still new feature surface though: a new config key, a hand-rolled language detector we'd maintain indefinitely, and prompt coupling across three phases with a fourth already noted as an unfinished follow-up. We're closing feature PRs by default to keep the surface area down, so this one is a close — nothing to do with the quality of the work.

howardpark:
@time-attack — unrelated to the close, and nothing here needs reopening.

When I opened this PR I accidentally included content in a test fixture that should never have been published (a third party's personal data — **not** a security issue in gbrain, and no risk to the project or its users). I've since rewritten the branch and deleted my fork, so it no longer resolves under my account, but `refs/pull/3358/head` on this repo still resolves to `ac01739`, so the superseded file is still served by the API and the diff view.

PR refs are read-only, so neither of us can remove it — only GitHub Support can, and their documented process asks for the request to come from the repository side. Would you be willing to send them a short note? Everything they need:

- Repository: `garrytan/gbrain`
- Affected PR: #3358 (closed, unmerged)
- Refs to purge: `refs/pull/3358/head`, plus `refs/pull/3358/merge` if cached
- Commit: `ac01739a0d822e1e74280bd4ab63396eabe61519`
- Ask: dereference/delete the affected PR ref, garbage-collect, clear cached views — no change to any branch or to the repo's history

I already wrote to privacy@github.com as the PR author and got the standard autoreply, which is why I'm asking here. Happy to handle any follow-up they send; I just can't originate it. Thanks either way.

Elialex31:
.

PR #3360, opened 2026-07-24
fix(transcripts): surface unreadable corpus dir via onWarn + doctor check

### LINKED ISSUE #3359: listRecentTranscripts silently skips an unreadable/missing corpus dir (looks identical to empty)
**What happened?**

`listRecentTranscripts` (`src/core/transcripts.ts`) skips a missing or unreadable
corpus dir silently (the catch just continues). Combined with an unset
`dream.synthesize.session_corpus_dir`, three distinct failures collapse into one
opaque result: an unset config, a wrong/misspelled path, and a genuinely empty
dir all report identically as "no recent transcripts". So a misconfigured corpus
is indistinguishable from an empty one, and there is no signal to the operator
that the dir they configured is unreadable or does not exist. This is a real
footgun for the dream/synthesize cycle, where a silent no-op looks like "nothing
to do".

**What did you expect?**

An unreadable or non-existent configured corpus dir to be surfaced (a warning on
the path that reads it, plus a `doctor` check), while a dir that exists and is
readable but empty stays quiet. The MCP op should keep its current default-silent
behaviour so nothing downstream changes by default.

**Steps to reproduce**
1. Set `dream.synthesize.session_corpus_dir` to a path that does not exist (or is not readable).
2. Run the transcript/dream path (or `gbrain doctor`).

Expected: a clear "corpus dir unreadable/missing" signal. Actual: "no recent transcripts", identical to an empty corpus.

**Environment**
- gbrain version: 0.42.65.0
- OS: aarch64 Ubuntu 24.04
- Bun version: 1.3.14

oliver-mee:
## Summary

`listRecentTranscripts` silently skipped a missing or unreadable corpus dir, so a
misconfigured `dream.synthesize.session_corpus_dir` was indistinguishable from an
empty one: unset config, wrong path, and empty dir all reported as "no recent
transcripts".

This adds an opt-in `onWarn` callback on the read path and a
`transcript_corpus_health` check in `doctor`, so a broken corpus dir is visible.
It is opt-in and default-silent: the `get_recent_transcripts` MCP op is unchanged,
a genuinely empty dir stays `[OK]`, and an unreadable/missing configured dir
surfaces `[WARN]` in doctor. A broken dir must not suppress an otherwise readable
corpus (pinned by a test).

## Linked issue

Fixes #3359

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Breaking change
- [ ] Documentation
- [ ] CI, tooling, or config

## How to test

1. `bun test test/transcripts-onwarn.test.ts` (5 pass: non-existent dir skipped silently by default, onWarn fires when opted in, a broken dir does not suppress a readable corpus, doctor WARN vs OK).
2. Manually: point `session_corpus_dir` at a non-existent path and run `gbrain doctor`; the corpus health check reports it.

## Checklist

- [x] I ran it and verified end to end. `bun run verify` (31/31 green) and the new test (5 pass). Did not run e2e (needs DATABASE_URL); the change is filesystem/doctor logic covered by unit tests.

## Model used

Claude Opus 4.8 (1M context) via Claude Code, used to sanitise and write the PR/issue. The fix and its test are the author's patch.

---

### DIFF
--- a/node_modules
+++ b/node_modules
@@ -1,0 +1,1 @@
+/home/ubuntu/gbrain-fork/node_modules
--- a/src/commands/doctor.ts
+++ b/src/commands/doctor.ts
@@ -5394,6 +5394,58 @@ export async function buildChecks(
     // Best-effort filesystem-hygiene check; never block doctor.
   }
 
+  // 3f. transcript_corpus_health.
+  //
+  // `listRecentTranscripts` skips an unreadable corpus dir and returns [], and
+  // its comment used to tell users to run `gbrain doctor` — but no such check
+  // existed, so a typo'd path was indistinguishable from an empty corpus and
+  // from no config at all. This is that check; the comment is now true.
+  //
+  // Unconfigured is OK, not a warning: the dream corpus is opt-in and most
+  // brains never set it. Only a configured-but-unreadable dir is a problem —
+  // the user asked for something that silently isn't happening.
+  if (engine !== null) try {
+    const corpusKeys: Array<[string, string]> = [
+      ['dream.synthesize.session_corpus_dir', 'session'],
+      ['dream.synthesize.meeting_transcripts_dir', 'meeting'],
+    ];
+    const configured: Array<{ label: string; dir: string }> = [];
+    for (const [key, label] of corpusKeys) {
+      const dir = await engine.getConfig(key);
+      if (dir) configured.push({ label, dir });
+    }
+
+    if (configured.length === 0) {
+      checks.push({
+        name: 'transcript_corpus_health',
+        status: 'ok',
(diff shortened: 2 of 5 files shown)


time-attack:
Thanks for spotting the ambiguity between an unset, wrong, and empty `session_corpus_dir` — that is a real diagnosability gap. But the patch is additive observability rather than a behavior fix (the read path stays default-silent by design), which puts it in the new-surface lane we're holding the line on, and the diff also commits a `node_modules` symlink pointing at `/home/ubuntu/gbrain-fork/node_modules` that would break other checkouts. Closing on both counts; the underlying observation was a good one.

Issue #2758, opened 2026-07-12
skillpack scaffold omits skills referenced by the stock resolver


raymeboltd:
## Summary

At commit `a25209bbb2bacf1b88e06fd5282b27f1bf4a3e7a` (currently `upstream/HEAD`), the stock resolver routes users to `frontmatter-guard` and `smoke-test`, but `openclaw.plugin.json` does not include either directory in its `skills` bundle list. A consumer that scaffolds the bundle can therefore receive resolver routes to skill files that were not installed.

## Evidence

- `skills/RESOLVER.md` routes frontmatter validation to `skills/frontmatter-guard/SKILL.md` and restart health checks to `skills/smoke-test/SKILL.md`.
- `skills/manifest.json` declares both skills.
- `openclaw.plugin.json#skills` includes neither `skills/frontmatter-guard` nor `skills/smoke-test`.

## Minimal reproduction

1. Start with an empty workspace.
2. Run `gbrain skillpack scaffold --all --workspace <empty-workspace> --trust`.
3. Confirm that `<empty-workspace>/skills/frontmatter-guard/SKILL.md` and `<empty-workspace>/skills/smoke-test/SKILL.md` are absent.
4. Read the stock `skills/RESOLVER.md` and route a matching request. Both target paths are missing from the scaffolded workspace.

## Expected

Every skill referenced by the stock resolver should either be included by the canonical `--all` scaffold bundle, or the resolver/scaffold command should emit an explicit supported exception with a canonical retrieval path.

## Actual

The resolver references two absent skill directories without an installation-time warning or supported workspace-local fallback.

## Suggested fix

Add the two directories to the bundled skill list, or make the bundle/resolver contract explicit and machine-checkable (including a warning or canonical external reference for intentionally excluded skills).

## Impact

Downstream routing can select a missing implementation, making an otherwise valid stock intent fail at runtime.

xaviroblessarries:
Fresh confirmation on current `master` (`c44cdb52b1ced1a7726c3e2be099cbba6ed25ff4`): `skills/manifest.json` declares 52 skills, but `openclaw.plugin.json` still bundles only 38 (`version: 0.32.3.0`). This now affects 14 manifest skills, not only the two originally reported. `INSTALL_FOR_AGENTS.md` simultaneously advertises 43. Suggested acceptance gate: derive or validate the scaffold list against `skills/manifest.json`, with an explicit machine-readable exclusion list for intentionally host-only or dangerous skills.

[...]

PR #3363, opened 2026-07-24
fix(skillpack): export manifest-declared shared bundle files in scaffold

### LINKED ISSUE #2759: skillpack scaffold does not export the stock resolver declared by skills/manifest.json
## Summary

At commit `a25209bbb2bacf1b88e06fd5282b27f1bf4a3e7a` (currently `upstream/HEAD`), `skills/manifest.json` declares `"resolver": "RESOLVER.md"`, but `gbrain skillpack scaffold --all` does not export `skills/RESOLVER.md` to the target workspace. The bundle also has no documented canonical export for that resolver.

## Minimal reproduction

1. Start with an empty workspace.
2. Run `gbrain skillpack scaffold --all --workspace <empty-workspace> --trust`.
3. Check for `<empty-workspace>/skills/RESOLVER.md`. It is absent.
4. Compare with the source manifest, which declares `RESOLVER.md` as the resolver.

## Expected

A scaffolded stock skill system should have a supported, portable routing surface. The command should either:

- export `skills/RESOLVER.md` (and any required manifest) as managed scaffold output; or
- provide a documented, machine-readable canonical resolver export/lookup mechanism that downstream hosts can use without copying source files manually.

## Actual

The skill files are scaffolded, but the stock dispatcher that defines their trigger routing is absent. Consumers must invent a host-local fallback or reference a separate checkout, neither of which is a portable bundle contract.

## Impact

Host integrations cannot reproduce stock routing from the canonical scaffold alone, and local workarounds can drift from the pinned bundle.

oliver-mee:
## Summary

`skillpack scaffold` exports only the files in `openclaw.plugin.json#shared_deps`.
That list omitted several files the bundle's own manifest and skills reference,
so a scaffolded consumer received a bundle pointing at files that were never
installed:

- `RESOLVER.md`, declared as the resolver in `skills/manifest.json` but never exported (#2759)
- `manifest.json` itself (skill-creator's MECE check reads it)
- `migrations/` (maintain and skillpack-check reference `skills/migrations/*`)
- `_friction-protocol.md` (referenced by several skills)

This adds all four to `shared_deps` so `enumerateScaffoldEntries` carries them
into a scaffolded workspace.

## Linked issue

Fixes #2759

(Sibling #2758, which is about the `skills` bundle list omitting resolver-routed
skills like `frontmatter-guard`/`smoke-test`, is a separate field and out of
scope here.)

## Type of change

- [x] Bug fix (non-breaking)
- [ ] Breaking change
- [ ] Documentation
- [ ] CI, tooling, or config

## How to test

1. `bun test test/scaffold-shared-bundle-files.test.ts` (2 pass; reads the real `openclaw.plugin.json` + `skills/manifest.json` and asserts the declared resolver and the referenced shared files are all in `shared_deps`).
2. Manually: `gbrain skillpack scaffold --all --workspace <empty> --trust`, then confirm `<empty>/skills/RESOLVER.md`, `manifest.json`, `migrations/`, and `_friction-protocol.md` are present. Before this change `RESOLVER.md` (and the others) were absent.

## Checklist

- [x] I ran it and verified end to end. `bun run verify` (31/31 green), the new unit test (2 pass), and a real `scaffold --all` into a scratch workspace (all four files now export). Did not run e2e (needs DATABASE_URL).

## Model used

Claude Opus 4.8 (1M context) via Claude Code, used to extend the shared_deps list to also cover RESOLVER.md (#2759), add the test, and write the PR. The original shared_deps fix is the author's patch.

---

### DIFF
--- a/openclaw.plugin.json
+++ b/openclaw.plugin.json
@@ -75,7 +75,11 @@
     "skills/_AGENT_README.md",
     "skills/_brain-filing-rules.md",
     "skills/_brain-filing-rules.json",
-    "skills/_output-rules.md"
+    "skills/_output-rules.md",
+    "skills/RESOLVER.md",
+    "skills/manifest.json",
+    "skills/_friction-protocol.md",
+    "skills/migrations"
   ],
   "excluded_from_install": [
     "skills/setup",
--- a/test/scaffold-shared-bundle-files.test.ts
+++ b/test/scaffold-shared-bundle-files.test.ts
@@ -1,0 +1,33 @@
+/**
+ * #2759: skillpack scaffold exports only the files listed in
+ * openclaw.plugin.json#shared_deps. skills/manifest.json declares a resolver
+ * (RESOLVER.md), and several skills reference sibling bundle files
+ * (manifest.json, migrations/, _friction-protocol.md), but none of those were
+ * in shared_deps, so a scaffolded consumer received a bundle whose own manifest
+ * and skills pointed at files that were never installed.
+ *
+ * These assertions read the REAL bundle config (not a fixture) so the shared_deps
+ * list cannot silently drop a manifest-declared or skill-referenced file again.
+ */
+import { describe, test, expect } from 'bun:test';
+import { readFileSync } from 'fs';
+import { join } from 'path';
+import { fileURLToPath } from 'url';
+
+const root = join(fileURLToPath(new URL('.', import.meta.url)), '..');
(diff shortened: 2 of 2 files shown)


time-attack:
Merging — confirmed on master that skills/manifest.json declares RESOLVER.md as its resolver while shared_deps omits it, so scaffolded workspaces really do ship a bundle pointing at files that were never exported. The four-line list fix plus a test that reads the real manifest (rather than a fixture) is exactly the right size for this. Thanks for scoping out the sibling #2758 concern instead of bundling it.

PR #3365, opened 2026-07-24
fix: circuit-break Gemini embedding billing failures

mattchronicle:
## Summary
- add official paid-tier pricing for google:gemini-embedding-001 so capped backfill jobs can enforce budgets
- classify AIConfigError 429s as provider-wide and skip rate-limit retries
- probe one page before worker fan-out so billing/config failures issue one request instead of one per worker

## Why
Google AI Studio monthly spending-cap responses use HTTP 429. The generic rate-limit path retried each concurrent page five times, turning a permanent billing failure into minutes of request amplification. The budget tracker also rejected Gemini backfills before calling the provider because the embedding pricing table lacked the configured model.

## Verification
- 72 targeted pricing/budget/embed tests pass
- 9 embed-stale integration tests pass
- TypeScript typecheck passes
- production proof: capped-provider backfill exits after one request in 14s, with no retries or worker fan-out

### DIFF
--- a/src/commands/embed.ts
+++ b/src/commands/embed.ts
@@ -19,6 +19,7 @@ import {
 } from '../core/pace-mode.ts';
 import { tryAcquireDbLock, type DbLockHandle } from '../core/db-lock.ts';
 import { embedBackfillLockId } from '../core/embed-backfill-lock.ts';
+import { AIConfigError } from '../core/ai/errors.ts';
 
 export interface EmbedOpts {
   /** Embed ALL pages (every chunk). */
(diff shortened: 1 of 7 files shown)


time-attack:
Thanks for the detailed writeup and the production measurement. We're closing this because the central claim doesn't hold against master: normalizeAIError in src/core/ai/errors.ts classifies 4xx-except-429 as AIConfigError and lets 429 fall through to AITransientError, so a Google spend-cap 429 never becomes an AIConfigError and the new breaker can't trip on the exact case described. What's left is a serial probe added ahead of worker fan-out in two hot backfill paths plus a pool branch that silently drops queued keys, which is more risk than we'll take on an unverified premise — and the three concerns here (pricing entry, retry classification, concurrency) really want to be separate.

PR #3369, opened 2026-07-24
docs(skills): require browser fallback for dynamic research

MarvinDontPanic:
## Why
Dynamic-site research sometimes stops after direct HTTP or API access fails even though the requested fields remain visible in a rendered browser.

## Change
Add two data-research failure modes: use browser/manual click-through before conceding, and report technical limitations only where they affect field confidence.

## Validation
Documentation-only two-line skill guidance change; diff reviewed against current upstream skill.


MarvinDontPanic:
Closing immediately at the repository owner request; this should not have been published.

Issue #2539, opened 2026-07-02
onboard 'chunks invisible to vector search' counts intentional embed_skip NULLs as critical backlog


victorfteha:
## Summary

`gbrain onboard --check` reports chunks with NULL embeddings as `[critical] N chunks invisible to vector search` and recommends an embed run — but the count includes chunks belonging to pages stamped `embed_skip: true`, whose embeddings are **intentionally** NULL (the documented privacy mechanism). On an install using `embed_skip` for sensitive subtrees, the check cries critical forever, and the number never goes down.

## Repro (v0.42.53.0)

1. Stamp `embed_skip: true` on some pages; full-reimport so the guard applies (chunks stored, embeddings NULL).
2. `gbrain onboard --check` → `[critical] (auto) embed — N chunks invisible to vector search`, where N includes all the embed_skip chunks.
3. Verify via SQL: join `content_chunks` (embedding IS NULL) to `pages` — the rows are on `embed_skip` pages.

## Why it matters

- A permanent `[critical]` finding trains users to ignore onboard output — and it masks *real* embed backlogs (e.g. post-upgrade re-chunk backlogs) behind a number that never reaches zero.
- The remediation is marked `(auto)` — an auto-apply path would attempt to embed content the user explicitly opted out of embedding (egress to the embedding provider).

## Suggested fix

Exclude chunks whose page frontmatter has `embed_skip` from the invisible-chunk count (same key-existence test the embed path uses), or report them as a separate informational line ("N chunks embed-skipped by policy"). The auto-remediation must never embed `embed_skip` pages.

time-attack:
Thanks for the precise report — the diagnosis was exactly right, and this is fixed on master. Commit a82a83db (#3634, 'health metric that cannot converge') changed the missing_embeddings health count in both engines to use the same predicate as `gbrain embed --stale`: `embedding IS NULL` AND the page does NOT carry `embed_skip` in frontmatter. So embed-skipped chunks no longer count toward the 'invisible to vector search' finding, the number converges to zero, and the (auto) remediation routes through the stale-chunk path which already excluded embed_skip pages — it never embeds opted-out content. Available since v0.42.69; if you still see embed_skip chunks in the count after upgrading, please reopen with `gbrain onboard --check` output plus the SQL join from your repro.

PR #3376, opened 2026-07-24
fix(onboard): stop counting policy-skipped embed_skip chunks as a critical embed backlog (#2539)

Masashi-Ono0611:
## Defect

`gbrain onboard --check` reports chunks with NULL embeddings as
`[critical] N chunks invisible to vector search` and queues an `(auto)`
`gbrain embed --stale` remediation — but `N` included chunks belonging to
pages stamped `embed_skip: true`, whose NULL embedding is the *intentional*
privacy mechanism (`src/core/embed-skip.ts`). On an install using
`embed_skip` for sensitive subtrees, the check cried critical forever and
the number never reached zero, training users to ignore onboard output.

(#2539)

## Root cause

`BrainHealth.missing_embeddings` (`getHealth()` in both
`postgres-engine.ts` and `pglite-engine.ts`) is a raw
`content_chunks WHERE embedded_at IS NULL` count with no join to
`pages`/frontmatter — it has no way to tell a policy-skip apart from a
real backlog.

The actual embed job (`gbrain embed --stale`, dispatched via
`countStaleChunks()` / `listStaleChunks()`) already excludes `embed_skip`
pages via the JSONB key-existence test
`COALESCE(p.frontmatter, '{}'::jsonb) ? 'embed_skip'`
(`src/core/postgres-engine.ts:2586`, `:2687`, `:2697`; same predicate in
`pglite-engine.ts`). So the count onboard showed and the work the queued
job would actually do had silently diverged.

## Fix

- Added `BrainHealth.embed_skip_missing_embeddings` (optional, both
  engines populate it), computed with the **exact same** key-existence
  predicate the embed path already uses — same string, same semantics, no
  second test to drift out of sync.
- `computeRecommendations()` (`src/core/brain-score-recommendations.ts`)
  now subtracts `embed_skip_missing_embeddings` from `missing_embeddings`
  before deciding whether to emit the `embed.stale` recommendation and
  what count/cost/time estimate to show:
  - All-skipped brain → **no critical entry at all** (nothing the embed
    job could act on).
  - Genuine backlog on non-skipped pages → still `[critical]`, unchanged
    severity, count reflects only the genuine subset.
  - Mixed case → critical entry for the genuine subset, with the skipped
    count folded into the rationale as an informational aside, e.g.
    `"10 chunks invisible to vector search (100 more skipped by embed_skip
    policy, not counted)"` — this reuses the existing "why:" rationale
    line onboard already renders, rather than adding a new output surface.

## Auto-remediation / egress safety (the sharper half of the issue)

Verified the `(auto)` remediation **cannot** embed `embed_skip` content,
today, unmodified by this PR:

- The `embed.stale` recommendation queues job `embed` with
  `params: { stale: true, sourceId }`.
- That job is registered in `src/commands/jobs.ts:1485` and calls
  `runEmbedCore(engine, { stale: true, ... })` (`src/commands/embed.ts`).
- The stale path resolves work via `engine.countStaleChunks()` /
  `engine.listStaleChunks()`, both of which already filter out
  `embed_skip` pages with the same JSONB key-existence predicate
  (`postgres-engine.ts:2577-2610`, `:2658-2699`; `pglite-engine.ts` has
  the equivalent `NOT (... ? 'embed_skip')` guard in every branch of
  `listStaleChunks`).

So no redundant guard was added — the egress protection already existed
independently of the (buggy) count; this PR only fixes the count/severity
that onboard *displays*, matching it to what the job actually does.

## What I verified

```
bun test test/brain-score-recommendations.test.ts test/pglite-engine.test.ts \
  test/embed-skip.test.ts test/sum-stale-chunk-chars.test.ts \
  test/advisor-core.test.ts test/advisor-ranking-eval.test.ts \
  test/features.test.ts test/doctor-timeline-metric-labels-2298.test.ts \
  test/onboard-pack-upgrade-checks.test.ts test/e2e/onboard-full-flow.test.ts \
  test/doctor-brain-checks-score.test.ts test/brain-score-breakdown.test.ts
# 246 pass, 0 fail, 511 expect() calls, 12 files

bun run typecheck   # tsc --noEmit — clean
bun run verify       # 31 parallel checks — all green (incl. check:jsonb)
```

New/extended tests (fail before the fix, pass after):
- `test/brain-score-recommendations.test.ts` — 4 new cases: all-skipped
  chunks produce no critical entry; genuine backlog on normal pages still
  fires critical with the plain rationale; mixed case excludes the
  skipped count from severity/cost/time and folds it into the rationale;
  omitted `embed_skip_missing_embeddings` defaults to 0 (back-compat for
  older callers/mocks).
- `test/pglite-engine.test.ts` — real-engine integration test: seeds an
  `embed_skip` page with a NULL-embedding chunk and asserts
  `getHealth().embed_skip_missing_embeddings` grows by exactly that
  chunk while `missing_embeddings` also grows by 1 (both counted, only
  one flagged as policy).

## Multi-model review (Codex, `model_reasoning_effort=high`)

No Critical findings. Two Warnings, both about code paths **not touched
by this diff** (confirmed by grepping the diff — neither `embed_coverage`,
`classifyChecks`, nor `maxReachableScore` appear in it) — documenting them
here rather than expanding this fix's surface:

1. **`embed_coverage` / `embed_coverage_score` / `maxReachableScore` still
   count policy-skipped chunks against the brain-health score**, and
   `classifyChecks()` still labels the `missing_embeddings` check
   `remediable` even when everything remaining is policy-skipped — so
   `--target-score` can claim an unreachable ceiling is reachable. This
   is pre-existing: `embed_coverage` was already computed over all chunks
   with no `embed_skip` awareness before this PR, and
   `classifyChecks`/`maxReachableScore` never received per-check health
   detail to begin with. This PR only fixes the onboard critical
   count/rationale (the literal scope of #2539); fixing the score
   denominator has a much larger blast radius (`embed_coverage` also
   feeds `gbrain doctor`'s separate warn-level check, `features.ts`'s
   pitch copy, `migrate-engine.ts`, and the advisor's
   `collect-usage-shape.ts`) and deserves its own issue/PR.
2. **Whole-brain counts driving a source-scoped job**: `getHealth()`
   takes no arguments and has always been whole-brain for every field
   (`stale_pages`, `dead_links`, the original `missing_embeddings`, and
   now `embed_skip_missing_embeddings`) — `computeRecommendations()` has
   never scoped health data to `ctx.sourceId` even though it stamps
   `sourceId` onto the emitted job. Pre-existing architectural gap,
   identical in kind for every `BrainHealth` field; not introduced or
   worsened by this PR. Threading `sourceId` through `getHealth()` across
   both engines is a separate, materially larger change.

## Out of scope (deliberately)

- The two Codex warnings above (separate follow-up).
- `gbrain doctor`'s own `warn`-level "embeddings" check
  (`src/commands/doctor.ts:5767-5769`) and `features.ts`'s pitch copy
  still use the raw (non-skip-aware) `missing_embeddings` — they're
  `warn`, not a permanent `[critical]`, and outside the issue's stated
  repro (`gbrain onboard --check`).

### DIFF
--- a/src/core/brain-score-recommendations.ts
+++ b/src/core/brain-score-recommendations.ts
@@ -226,13 +226,27 @@ export function computeRecommendations(
 
   // ---------------------------------------------------------------------
   // embed.stale — missing embeddings. Critical: invisible to vector search
+  //
+  // #2539: chunks on `embed_skip` pages are excluded from the critical
+  // count — their NULL embedding is the intentional privacy mechanism
+  // (embed-skip.ts), not a backlog. Pre-fix this fired [critical] forever
+  // on any install using embed_skip, training users to ignore onboard
+  // output. `gbrain embed --stale` (the job this recommendation queues)
+  // already excludes embed_skip pages via the same key-existence test
+  // (countStaleChunks/listStaleChunks), so subtracting them here just
+  // makes the count match what the job will actually do.
   // ---------------------------------------------------------------------
-  if (health.missing_embeddings > 0 && ctx.embeddingProviderConfigured !== false) {
+  const skippedByPolicy = Math.min(
+    health.embed_skip_missing_embeddings ?? 0,
+    health.missing_embeddings,
+  );
+  const genuineMissingEmbeddings = health.missing_embeddings - skippedByPolicy;
+  if (genuineMissingEmbeddings > 0 && ctx.embeddingProviderConfigured !== false) {
     const params = { stale: true, sourceId: ctx.sourceId };
     const embedModel = ctx.embeddingModel ?? 'openai:text-embedding-3-large';
     const embedDims = ctx.embeddingDimensions ?? 3072;
     // Rough char estimate per chunk ~ 1.5k chars (chunker target).
-    const estChars = health.missing_embeddings * 1500;
+    const estChars = genuineMissingEmbeddings * 1500;
     let est_usd_cost = 0;
     try {
       const priceLookup = lookupEmbeddingPrice(embedModel);
@@ -242,17 +256,21 @@ export function computeRecommendations(
     } catch {
       /* unknown model — leave at 0, surface as warning elsewhere */
     }
+    const rationale = `${genuineMissingEmbeddings} chunk${genuineMissingEmbeddings === 1 ? '' : 's'} invisible to vector search` +
+      (skippedByPolicy > 0
+        ? ` (${skippedByPolicy} more skipped by embed_skip policy, not counted)`
+        : '');
     out.push({
       id: 'embed.stale',
       job: 'embed',
       params,
       idempotency_key: idemKey(source, 'embed', { ...params, embedModel, embedDims }),
       severity: 'critical',
-      est_seconds: Math.min(3600, 5 + health.missing_embeddings * 0.05),
+      est_seconds: Math.min(3600, 5 + genuineMissingEmbeddings * 0.05),
       est_usd_cost,
       // sync should run first so embed sees fresh pages.
       depends_on: ctx.repoPath && health.stale_pages > 0 ? ['sync.repo'] : [],
-      rationale: `${health.missing_embeddings} chunk${health.missing_embeddings === 1 ? '' : 's'} invisible to vector search`,
+      rationale,
       status: 'remediable',
     });
   }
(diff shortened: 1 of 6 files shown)


time-attack:
Merging — the divergence is real: getHealth()'s missing_embeddings is a raw NULL-embedding count with no pages join, while the embed job it queues already filters embed_skip pages, so the critical count could never reach zero on a brain using embed_skip. Reusing the exact same JSONB key-existence predicate in both engines (rather than writing a second test that can drift) and folding the skipped count into the existing rationale line is the right call. Particularly appreciated that you traced the auto-remediation path and confirmed no extra guard was needed instead of adding a defensive one.

Masashi-Ono0611:
Thanks for merging, and for confirming the auto-remediation path directly rather than assuming a defensive guard was needed on top.

Issue #1978, opened 2026-06-08
feat: raw-source persistence guarantee for synthesis/import pipelines


adriankae:
## Feature description

Add a first-class raw-source persistence guarantee for synthesis/import pipelines.

Desired invariant:

> Any synthesized/compiled page that is derived from a source document/transcript/API response must either reference a persistent raw trace or explicitly declare a reviewed exemption.

Examples of acceptable raw traces:

- `raw_data` row attached to the page/source
- source markdown/transcript/archive file path
- durable source URI plus ingestion receipt
- storage object/file record
- explicit `raw_trace_exempt: true` with reason

## Motivation

GBrain should never silently replace read source text with a synthesized summary. Synthesis is valuable, but it must be additive: raw material persists, then curated truth is written on top.

This matters for:

- rebuildability
- auditability
- provenance
- avoiding LLM over-compression as data loss
- allowing later agents/humans to inspect the original source
- preventing dream/synthesis/import jobs from becoming lossy transformations

## Proposed solution

Add an upstream-native guard/check that can run in doctor/onboard or as part of synthesis/import write paths:

1. For generated/synthesized pages, require one of:
   - `raw_trace`
   - `raw_source`
   - `source_uri`
   - attached `raw_data`
   - storage/file receipt
   - explicit exemption marker with reason
2. Surface violations in `gbrain doctor` or a dedicated integrity check.
3. Prefer fail-closed or warn-hard for synthesis pipelines that would write without raw provenance.
4. Keep synthesis pages additive; never allow a synthesis output to be the only persisted representation of read source text unless explicitly exempted.

## Non-goals

- Do not require every hand-authored note to have a raw trace.
- Do not store secrets or sensitive source payloads unsafely.
- Do not mandate one storage backend; raw trace can point to existing GBrain raw_data, markdown, storage, or source receipts.

## Current operational workaround

For one deployment, this is being enforced operationally via a silent hygiene guard and an agent rule: "raw first, synthesis second." That should remain a workaround, not a local patch.

PR #3377, opened 2026-07-24
v0.42.65.0 fix(backfill): stamp raw-trace exemption on pre-existing dream-cycle index pages (#1978)

Masashi-Ono0611:
Follow-up to #1978. Complements #3300.

## The gap

#3300 added the warn-only `raw_provenance` check in `doctor`: a synthesized page must carry a raw trace (`raw_trace` / `raw_source` / `source_uri` frontmatter, a `raw_data` row, or `synthesis_evidence` rows) or an explicit `raw_trace_exempt` marker.

In the same wave `writeSummaryPage` in `src/core/cycle/synthesize.ts` began stamping the per-cycle index page `dream-cycle-summaries/<date>` with `raw_trace_exempt: true` + `raw_trace_exempt_reason` — the index is deterministic, it has no source document of its own, and the raw traces live on the pages it lists.

**That stamping is prospective only. There is no backfill.** Index pages written before that code landed carry no stamp, so on any brain that has them the check warns permanently with no supported way to clear it.

This PR adds `gbrain backfill dream_cycle_index_provenance`, which applies the writer's own stamp to exactly those rows. It is not a new design decision — it applies semantics this repo already shipped to rows that predate the code.

## What dogfooding showed

This started as a live warn on a real 1609-page brain (v0.42.65.0, schema v125, PGLite). Measured directly in Postgres with the `raw_provenance` predicate:

| probe | value |
|---|---|
| pages matching the `raw_provenance` predicate | **183** |
| ... `dream-cycle-summaries/<ISO date>` | **49**, created 2026-05-24 .. 2026-07-23 |
| ... `wiki/*` | 79, created 2026-07-03 .. 2026-07-23 |
| ... `extracts/*` | 55, created 2026-05-29 .. 2026-07-23 |
| rows in `synthesis_evidence` | **0** |
| of the 49 index pages, `dream_cycle_date` == slug tail | 49 |

Being precise about the state you would see if you probed that brain right now: **it reports OK, because I remediated it out-of-band a few hours before opening this PR** — I stamped the frontmatter directly in the DB (content untouched, so no re-chunk or re-embed), keeping a backup of all 183 rows first. So the 183 is a real measurement with a retained artifact, not a number I can re-derive from that brain on demand. I would rather state that plainly than either present an unreproducible live count or drop the evidence entirely.

Two observations from that data that cut in different directions:

- **`synthesis_evidence` is empty on a brain that has been running dream cycles for two months.** That branch of the check is theoretical in practice, so for these pages the exemption stamp is the only thing that can satisfy it.
- **My out-of-band fix for the other 134 pages is deliberately *not* what this PR proposes.** For `wiki/*` and `extracts/*` I stamped a blanket exemption to quiet my own brain, which silences the check rather than satisfying it. That is a defensible local stopgap and a bad upstream default, which is exactly why this PR scopes itself to the index pages and leaves those two namespaces to you (see below).

The `dream_cycle_date == slug tail` row above is load-bearing for the predicate: it confirms the extra guard excludes no real index page.

## Why the backfill registry and not a schema migration

A migration auto-applies on upgrade and would reach every brain without operator action, which is genuinely attractive here. I went the other way because of what this repo already does with each vehicle:

- `MIGRATIONS` in `src/core/migrate.ts` carries schema/DDL and trigger work. The only `UPDATE pages` in the whole array is a no-op `SET id = id` touch to re-fire a trigger.
- Historical **row-data** repairs go through the backfill lane — `effective_date`, `emotional_weight`, `modality` — which brings keyset pagination, adaptive batch halving, a reserved-connection write path and `--dry-run`.
- `modality` is the exact precedent for this shape: `doctor` warns and the message says ``Fix: `gbrain backfill modality` ``. This PR mirrors that, and updates the `raw_provenance` message to point at the new lane the same way.

Happy to convert it to a migration if you would rather it auto-apply — it is a small change either way.

## Deliberately out of scope: `wiki/*` and `extracts/*`

Other pre-existing synthesized pages want a `raw_source` naming the transcript they were derived from. That mapping (child job -> source path) is gone once the cycle has run, so it cannot be reconstructed retroactively. Blanket-stamping an exemption on them would silence the check rather than satisfy it, and would destroy the signal for the pages that genuinely lack provenance. **I am leaving that call to you** — the options are roughly: accept a permanent warn for pre-existing content pages, add a distinct lower-confidence marker, or scope the check to pages created after the stamping code landed.

## Implementation notes

- **No drift by construction.** The predicate and the exemption literals move into `src/core/raw-provenance.ts`; the check, the writer and the backfill all import from there.
- **Predicate.** The shared untraced-page predicate, plus the index slug shape (`^dream-cycle-summaries/[0-9]{4}-[0-9]{2}-[0-9]{2}$`), plus `dream_generated = 'true'`, plus `dream_cycle_date = right(slug, 10)` — the writer always composes the slug from that date, so a mismatch means the page did not come from `writeSummaryPage`.
- **Idempotent.** The stamp it writes is one of the keys the shared predicate excludes, so a second run matches nothing.
- **Content untouched.** Only `frontmatter` moves — `compiled_truth` and `content_hash` are unchanged, so nothing re-chunks or re-embeds. The existing generation/cache-invalidation triggers fire, as they would for any frontmatter write.
- Two optional, backward-compatible knobs on `BackfillSpec`:
  - `columnUpdateSql` — per-column RHS builder. Used as `frontmatter = COALESCE(frontmatter,'{}'::jsonb) || $N::jsonb`, which both binds a raw object at an explicit `::jsonb` position (never `JSON.stringify` there — the #2339 shape) and merges against the *current* row instead of a JS-side snapshot.
  - `resumable` — defaults to true, so existing backfills are unaffected. This one sets it false: its predicate is self-clearing, so a checkpoint buys nothing and costs correctness (see the known limitation below).

**Known limitation.** The backfill repairs the DB row, not the historical markdown file the dream cycle dual-wrote. If such a file is later edited and re-imported, the import replaces frontmatter wholesale and the exemption is lost again. `resumable: false` is what makes that recoverable: a plain re-run re-stamps the row rather than skipping it below a watermark. Persisting the stamp into the canonical markdown is a larger change and I left it out.

## Testing

- New `test/backfill-dream-cycle-index-provenance.test.ts` against real PGLite: stamps the unstamped index page; leaves alone an already-stamped page, a hand-authored page in the namespace, a date-shaped impostor whose `dream_cycle_date` disagrees with its slug, a page with a `raw_data` row, a soft-deleted page, and `wiki/*` / `extracts/*` synthesis pages; asserts the merge preserves existing keys and leaves `compiled_truth` / `content_hash` untouched; asserts re-running is a no-op and `--dry-run` writes nothing.
- `test/backfill-base.test.ts`: pins the exact generated `UPDATE` SQL and parameter order for `columnUpdateSql`, and that `resumable: false` neither reads nor writes a checkpoint.
- `test/e2e/postgres-jsonb.test.ts`: `DATABASE_URL`-gated case asserting `jsonb_typeof(frontmatter) = 'object'` after the merge, since PGLite cannot surface the double-encode class.
- `bun run verify` 31/31 green; `bun run typecheck` clean; affected suites green. I could not run the `DATABASE_URL`-gated e2e case locally (no Postgres available in this environment) — CI covers it.

### DIFF
--- a/src/commands/backfill.ts
+++ b/src/commands/backfill.ts
@@ -79,11 +79,17 @@ Usage:
   gbrain backfill <kind> [flags]      Run a registered backfill.
   gbrain backfill list                 Show registered backfills + checkpoints.
 
-Backfills (v0.30.1):
+Backfills (\`gbrain backfill list\` is the authoritative registry):
   effective_date     Compute effective_date for pages imported pre-v0.29.1.
   emotional_weight   Recompute emotional_weight for pages with stale stamp.
   embedding_voyage   Declared-only in v0.30.1 (multi-column embedding lands
                      in v0.30.2 alongside the schema migration).
+  modality           Tag image-asset chunks whose modality was missed by
+                     pre-v0.27.1 ingest.
+  dream_cycle_index_provenance
+                     Stamp raw_trace_exempt on dream-cycle index pages that
+                     predate synthesize.ts stamping it (clears the doctor
+                     raw_provenance warn for those pages, #1978).
 
 Flags:
   --batch-size N     Initial batch size before adaptive halving (default 1000).
--- a/src/commands/doctor.ts
+++ b/src/commands/doctor.ts
@@ -22,6 +22,7 @@ import { loadCompletedMigrations } from '../core/preferences.ts';
 import { compareVersions } from './migrations/index.ts';
 import { createProgress, startHeartbeat, type ProgressReporter } from '../core/progress.ts';
 import { categorizeCheck, type CheckCategory } from '../core/doctor-categories.ts';
+import { untracedSynthesizedPagesPredicate } from '../core/raw-provenance.ts';
 import { rankIssues, type RankedIssue } from '../core/doctor-cause-rank.ts';
 import { getCliOptions, cliOptsToProgressOptions } from '../core/cli-options.ts';
 import type { DbUrlSource } from '../core/config.ts';
@@ -548,12 +549,10 @@ export async function childTableOrphansCheck(engine: BrainEngine): Promise<Check
  * childTableOrphansCheck so tests can target it directly.
  */
 export async function rawProvenanceCheck(engine: BrainEngine): Promise<Check> {
-  const where = `
-        p.deleted_at IS NULL
-    AND (COALESCE(p.frontmatter->>'dream_generated', '') = 'true' OR p.type = 'synthesis')
-    AND NOT (COALESCE(p.frontmatter, '{}'::jsonb) ?| ARRAY['raw_trace', 'raw_source', 'source_uri', 'raw_trace_exempt'])
-    AND NOT EXISTS (SELECT 1 FROM raw_data rd WHERE rd.page_id = p.id)
-    AND NOT EXISTS (SELECT 1 FROM synthesis_evidence se WHERE se.synthesis_page_id = p.id)`;
+  // Predicate lives in core/raw-provenance.ts so the
+  // `dream_cycle_index_provenance` backfill (which exists to clear exactly
+  // these rows) cannot drift from the check it satisfies.
+  const where = untracedSynthesizedPagesPredicate('p');
   try {
     const rows = await engine.executeRaw<{ n: string | number }>(
       `SELECT COUNT(*)::int AS n FROM pages p WHERE ${where}`,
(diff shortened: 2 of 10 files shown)


time-attack:
Thanks for the unusually honest writeup — flagging that you remediated your own brain out-of-band before filing, and that the wiki/* and extracts/* majority is deliberately out of scope, is the kind of candour that makes a PR easy to evaluate. We're closing it because it adds a permanent new CLI surface (~570 lines) to clear a warn-only check where nothing is actually broken or lost, and it covers 49 of the 183 affected pages. The backfill-registry-over-migration reasoning was sound; the issue is that the whole lane isn't warranted here.

Masashi-Ono0611:
Understood — 570 lines of new CLI surface for a warn-only check with no actual loss is a fair line to hold, especially at 49/183 coverage. Appreciate you calling the candour on the out-of-band remediation useful rather than penalizing it.

Issue #2364, opened 2026-06-22
think: synthesisOk=false because model's markdown-fenced JSON isn't unwrapped before parsing


changyoutaxiang:
## Summary

Every `think` call comes back with `synthesisOk: false` and `warnings: ["LLM_OUTPUT_NOT_JSON", "CITATIONS_REGEX_FALLBACK"]`, even when the underlying answer is fully correct.

Root cause: the synthesis model wraps its JSON response in a Markdown code fence (` ```json … ``` `), and the parser appears to run `JSON.parse` on the fenced string directly, which fails. The regex fallback then recovers citations, but:

1. `synthesisOk` is reported `false` on a perfectly good answer — a misleading failure signal for any caller that gates on it.
2. The top-level `answer` field ships the raw fenced blob — a ` ```json\n{"answer":"…","citations":[…]}\n``` ` string — so the real prose ends up double-wrapped (a JSON string, inside a fence, inside the `answer` field). Consumers must strip the fence and re-parse to get clean Markdown.

## Environment

- Version: `0.42.51.0`
- Engine: `pglite`
- Transport: MCP (remote, oauth)
- Synthesis model reported by `think`: `openrouter:anthropic/claude-sonnet-4.6`

## Repro

Any `think(question=…)` against a brain with relevant pages. Observed 2/2 on two separate entity-centric questions (with and without `anchor`).

## Actual (both runs)

```
"warnings": ["LLM_OUTPUT_NOT_JSON", "CITATIONS_REGEX_FALLBACK"],
"synthesisOk": false,
"answer": "```json\n{\n  \"answer\": \"## …actual markdown…\",\n  \"citations\": [...],\n  \"gaps\": [...]\n}\n```"
```

## Expected

- Strip a leading/trailing Markdown code fence (```` ```json ```` … ```` ``` ````) before `JSON.parse`, so the structured parse succeeds.
- `synthesisOk: true` when the answer + citations are recovered.
- `answer` carries clean Markdown prose, not a re-fenced JSON envelope.

## Why it matters

The fence-wrapping is reliable model behavior (at least for `claude-sonnet-4.6`), so this isn't an occasional flake — it's the steady state. `synthesisOk` is effectively always `false` and every consumer pays an unwrap tax. A one-line fence-strip before parse fixes both symptoms.

_(Observed as of `0.42.51.0`; `0.42.52` shipped with an orchestrator rewrite — please confirm whether this is already addressed there.)_

time-attack:
[...]

PR #3378, opened 2026-07-24
fix(think): unwrap markdown-fenced JSON before parsing synthesis output (#2364)

Masashi-Ono0611:
## Defect

`think` reports `synthesisOk: false` with warnings `LLM_OUTPUT_NOT_JSON` + `CITATIONS_REGEX_FALLBACK` on perfectly good answers, and the top-level `answer` field ships the raw Markdown-fenced JSON blob instead of clean prose. Reported in (#2364).

## Root cause

The synthesis model routinely wraps its structured JSON response in a Markdown code fence (```` ```json … ``` ````) despite the system prompt's "no prose outside JSON" instruction. `tryParseJSON()` in `src/core/think/index.ts` only stripped a fence that **anchored the entire response** (`^```...```$`). Any lead-in sentence before the fence (common with reasoning-style models) defeated that anchored strip, and the whole-text brace-scan fallback would then latch onto a stray `{`/`}` inside that preamble instead of the real JSON — producing exactly the reported symptom even though the JSON was present and well-formed inside its fence.

## Fix

`tryParseJSON()` now builds up to three candidate substrings and prefers whichever one actually looks like the expected `{answer, citations, gaps}` shape:

1. The full trimmed text (handles the common no-fence case).
2. The fenced region, located by **first opening marker → last closing marker** (not anchored to the very start/end, so a lead-in sentence doesn't defeat it; not non-greedy, so a nested fence inside the JSON's own `answer` markdown doesn't truncate it early).
3. The first `{...}` block found anywhere in the full text (the original fallback, for prose-wrapped JSON with no fence at all).

Every candidate that parses as an object is considered before returning — a candidate with a string `answer` field always wins over one that doesn't, rather than short-circuiting on whichever candidate happened to parse first. This ordering came out of three rounds of bounded Codex review against the draft diff, each of which surfaced a real edge case (see Verification below). The fence-tag match was also widened to be case-insensitive with an optional space (```` ```JSON ````, ```` ``` json ````).

The regex citation fallback (`CITATIONS_REGEX_FALLBACK` path in `cite-render.ts`) is untouched — it's a real fallback for genuinely non-JSON output and still fires correctly when there's no JSON to recover.

## Verified

- `bun test test/think-pipeline.serial.test.ts` — 35 pass, 0 fail, 92 `expect()` calls. Includes a new test that reproduces the original bug (a preamble sentence with a stray brace before a real fence) and was confirmed to fail against the pre-fix implementation before the fix was applied, plus tests for the anchored-fence case, and three regression-guard tests added in response to Codex review findings (embedded fence inside `answer`, embedded fence + trailing prose, and an embedded fence whose content is itself coincidentally valid JSON).
- `bun run typecheck` — clean (`tsc --noEmit`, no errors).
- `git diff --stat upstream/master` — 2 files changed (`src/core/think/index.ts`, `test/think-pipeline.serial.test.ts`), no unintended files, no secrets.
- Bounded `codex exec` review (4 rounds, `model_reasoning_effort=high`, diff-scoped) — each round's Critical/Warning findings were addressed in the diff before the next round; the final round's remaining two findings are noted below as deliberate scope boundaries.

## Deliberately out of scope

Two findings from the final Codex review round were not chased further, to stay within "robust but not speculative":

- The whole-text brace-scan fallback still can't disambiguate multiple independent brace regions in already-noisy text (e.g. valid JSON followed by unrelated prose that itself happens to contain a balanced `{...}`). This is a pre-existing limitation of the brace-scan fallback (it existed in the original code too, since the anchored strip never touched this case either) — not introduced by this change.
- The candidate-preference logic doesn't try to detect an earlier illustrative "example" JSON object the model might include unprompted before the real response. Disambiguating that would need a materially different extraction strategy, and is an increasingly unrealistic scenario given the system prompt's explicit "no prose outside JSON" instruction — there's no evidence models exhibit this pattern for this schema.

Both would need a more general "extract the right one of several plausible JSON objects from noisy text" solution, which felt disproportionate to the reported bug.

### DIFF
--- a/src/core/think/index.ts
+++ b/src/core/think/index.ts
@@ -174,19 +174,58 @@ function inferIntent(question: string, anchor?: string): string {
   return 'general';
 }
 
+/**
+ * Parse the synthesis model's response into the `{answer, citations, gaps}`
+ * shape, tolerating a Markdown code fence around (or a lead-in sentence
+ * before) the JSON — models routinely add one despite the system prompt
+ * asking for "no prose outside JSON" (#2364).
+ *
+ * Three candidate substrings are tried, in this fixed order, and — this is
+ * the important part — EVERY candidate that parses as an object is
+ * considered before returning: a candidate whose parsed shape actually has
+ * a string `answer` field always wins over one that doesn't, regardless of
+ * which candidate was tried first. Short-circuiting on the first candidate
+ * that merely parses (rather than the first one that looks like a real
+ * response) is what caused earlier regressions here — e.g. a fenced code
+ * snippet embedded INSIDE a valid `answer` value can itself look like a
+ * wrapping fence, or even parse to a valid-but-wrong object like `{}`.
+ *   1. The full trimmed text (handles the common no-fence case, and any
+ *      fence-looking sequence that's actually just part of a string value).
+ *   2. The Markdown-fenced region, if a fence is present: from the first
+ *      opening fence marker to the LAST closing one (not anchored to the
+ *      very start/end, so a lead-in sentence before the fence doesn't
+ *      defeat it; not non-greedy, so a nested fence inside the JSON's own
+ *      `answer` markdown doesn't truncate it early).
+ *   3. The first `{...}` block found anywhere in the full text — the
+ *      original fallback, for prose-wrapped JSON with no fence at all.
+ */
 function tryParseJSON(text: string): unknown {
-  // The model may wrap JSON in code fences. Strip if present.
-  const stripped = text.trim().replace(/^```(?:json)?\s*\n?/, '').replace(/```\s*$/, '');
-  try {
-    return JSON.parse(stripped);
-  } catch {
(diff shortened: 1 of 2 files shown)


time-attack:
Merging — confirmed the anchored strip on master is exactly as described, and that a preamble sentence lets the brace-scan fallback latch onto the wrong region, which explains both the false synthesisOk:false and the raw blob leaking into `answer`. Keeping the whole change inside tryParseJSON and preferring the candidate that actually has a string `answer` field (rather than short-circuiting on first parse) handles the nested-fence cases properly. Thanks for writing a test that was verified to fail against the pre-fix implementation, and for naming the brace-scan ambiguity as a pre-existing limitation instead of expanding scope to chase it.

Masashi-Ono0611:
Thanks for merging, and for verifying the nested-fence cases against the preamble-sentence failure mode directly rather than taking the description on faith.

PR #3396, opened 2026-07-25
v0.42.66.0 fix(onboard): honor file-plane schema pack in checks (#2538)

### LINKED ISSUE #2538: onboard pack_upgrade_available resolves active pack with cfg:null — file-plane schema_pack invisible, false unify-types recommendation
## Summary

`checkPackUpgradeAvailable` (onboard) resolves the active schema pack with `cfg: null`, which makes a file-plane `schema_pack` set in `~/.gbrain/config.json` invisible to the check. With no DB-side `schema_pack` config row, resolution falls through to the default (`gbrain-base`), so the check recommends the `unify-types` migration to installs that are **already on a custom or v2-based pack**. The same blind spot makes `gbrain doctor`'s pack checks report the wrong active pack identity.

## Repro (v0.42.53.0)

1. Activate a custom pack the file-plane way: `gbrain schema use my-pack` → `~/.gbrain/config.json` gets `"schema_pack": "my-pack"`; the DB `config` table has no `schema_pack` row.
2. `gbrain schema active` → correctly reports `my-pack` (it reads the file config).
3. `gbrain onboard --check` → emits `unify-types — Pack upgrade gbrain-base → gbrain-base-v2`, even though the install is not on `gbrain-base`.

## Root cause

`src/core/onboard/checks.ts` (`checkPackUpgradeAvailable`):

```ts
const active = await loadActivePack({ cfg: null, remote: false, dbConfig })
```

The comment says resolution "falls through to file-plane/env/default resolution when unset" — but it can't: `buildResolutionInput` (`src/core/schema-pack/load-active.ts:281-295`) reads the file-plane tier only from `input.cfg?.schema_pack`, so `cfg: null` guarantees `homeConfig` is undefined. With `dbConfig` also unset, resolution lands on the default `gbrain-base`, and `findPackSuccessors('gbrain-base', …)` finds `gbrain-base-v2` via its `migration_from`.

## Why it matters

- The emitted remediation is the `unify-types` Minion job. If a user on a custom fork follows it with `target_pack: gbrain-base-v2`, the run applies v2's `mapping_rules` **including the `*unknown*` catch-all**, flattening every custom page type to `note`, and then flips the active pack to `gbrain-base-v2` — silently deactivating their fork.
- Doctor's `pack_upgrade_available` / `type_proliferation` checks show the same wrong pack identity, which erodes trust in doctor output (`gbrain schema active` and doctor disagree permanently).

## Suggested fix

Load the file-plane config and pass it through (e.g. `loadActivePack({ cfg: loadConfigFileOnly(), … })`), or extend `buildResolutionInput` to fall back to `loadConfigFileOnly()` when `cfg` is null. A regression test: file-plane-activated pack + no DB config row → `pack_upgrade_available` reports "current".

javieraldape:
Fixes #2538.

## Summary

- Resolve the active schema pack for onboard upgrade/proliferation checks with the file-plane config included.
- Preserve DB config precedence so a post-unify `schema_pack` row still wins over file config.
- Add a regression where config.json selects `gbrain-base-v2` and the DB has no `schema_pack` row, proving the upgrade check reports the selected pack as current instead of recommending base-to-v2.

## Tests

- `bun test test/onboard-pack-upgrade-checks.test.ts` (9 pass)
- Local gate: GREEN; verify passed, focused units passed (9 pass), and diff-selected E2E resolved to the fail-closed all-set so it was skipped locally per gate policy.


time-attack:
Repair pushed to this branch (maintainer edit):

- **Merged origin/master** (branch was ~20 commits behind; clean merge, no conflicts).
- **Hermeticity fix for the pre-existing tests**: routing `checkPackUpgradeAvailable`/`checkTypeProliferation` through `loadConfigFileOnly()` made this file's two pre-existing test groups read the developer's real `~/.gbrain/config.json` — they failed on any machine whose config sets `schema_pack`. They're now sandboxed with `withEnv({ GBRAIN_HOME: emptyHome(), GBRAIN_SCHEMA_PACK: undefined }, ...)`, the same idiom this PR's new test already used.

Verified: `bun run typecheck` exit 0; `bun test test/onboard-pack-upgrade-checks.test.ts` 9/9 pass both with a clean env and with `GBRAIN_HOME` pointed at a config setting `schema_pack=gbrain-base-v2` (pre-fix that dirty run failed 2/9).

Credit to @javieraldape for the underlying fix.

Issue #3392, opened 2026-07-25
Job system records cache tokens but ModelPricing has no cache rates, and there is no actual-spend report — downstream cost tooling is forced to duplicate (and drift from) the canonical table


Masashi-Ono0611:
## Summary (corrected 2026-07-26, resolved via PR #3399)

gbrain records cache token counters but can't price them, and until PR #3399 landed had no surface that reports actual spend across the LLM traffic that matters. Two concrete gaps, plus a scope caveat found mid-implementation that changed the shape of the fix:

**Gap 1 — `ModelPricing` has no prompt-cache rates.** The `subagent` job handler records cache token counters (per-message `tokens_cache_read` / `tokens_cache_create` in `subagent_messages`, plus a per-job `tokens_cache_read` rollup — `src/core/minions/transcript.ts:29-30`, `schema.sql:895`), and for a healthy brain cache traffic *dominates* the token mix (our nightly dream cycles run at cacheR/in ratios in the tens of thousands, for the calls that do go through this path). But `ModelPricing` (`src/core/model-pricing.ts:39-44`) carried only `input`/`output`, so none of that recorded cache traffic could be priced in-tree. Anthropic prices cache-read at 0.1× input and 5m cache-write at 1.25× input, so ignoring it isn't a rounding error — on a cache-heavy workload the input/output-only view is off by a large factor in either direction depending on which side you drop.

**Gap 2 — no actual-spend report, for any scope.** All the pricing consumers before this issue were *pre-flight estimators* (`estimateMaxCostUsd` is documented as an upper bound for budget caps; budget-tracker / budget-meter / batch-projection are the same family). There was no op or command that joins recorded token counters × the canonical table into "here is what the last night / week actually cost."

**Gap 3 (found while implementing #1/#2, corrected the original framing of this issue) — the job system only sees a small slice of gbrain's actual LLM traffic.** `subagent_messages` is written from exactly one call site: the `subagent` minion-job handler (`src/core/minions/handlers/subagent.ts:1424`). Verified by grep which cycle phases and features actually dispatch through that path vs. calling `gateway.chat()` directly:

| Reaches `subagent_messages` | Calls `gateway.chat()` directly (untracked by DB *or* audit JSONL *or* BudgetTracker) |
|---|---|
[...]

Issue #2823, opened 2026-07-14
Test suite writes content-sanity audit events into the operator's real ~/.gbrain/audit (GBRAIN_AUDIT_DIR never set in tests) — pollutes doctor's content_sanity_audit_recent


paul-0320:
## Version / environment

- gbrain 0.42.59.0, macOS (fork `qwen3-mrl-support`; the paths cited below are upstream-unmodified)

## Summary

`test/import-file.test.ts` exercises `importFromContent` with a mock engine — the DB is fully isolated — but the content-sanity gate's audit logger (`logContentSanityAssessment`) defaults to the **real** `~/.gbrain/audit/content-sanity-YYYY-Www.jsonl`. A `GBRAIN_AUDIT_DIR` override exists (src/core/audit/content-sanity-audit.ts) but the test harness never sets it. Running the suite on a machine that also hosts an operational brain appends fixture events to the operator's live audit trail, and `doctor`'s `content_sanity_audit_recent` then reports them as production signal.

## Observed (production machine)

`doctor` reported `soft_block: 4` for source `default` in `content_sanity_audit_recent`. All four events in the JSONL were the same test fixture:

```
{"event_type":"soft_block","slug":"borderline-slug","source_id":"default","bytes":4900000,...}
```

- Timestamps match our test-suite runs exactly (3× on 2026-07-09, 1× on 2026-07-13 during an upgrade's full-suite run).
- `pages` has no `borderline-slug` row — mock engine kept the DB clean while the JSONL got polluted.
- The fixture is written from test/import-file.test.ts (~line 385: `importFromContent(engine, 'borderline-slug', content, { noEmbed: true })` with a 4.9MB body crossing the soft-block threshold).

Cost of the pollution in practice: the spurious `soft_block: 4` coincidentally matched the count of an unrelated ingestion problem we were investigating, and steered that investigation toward the content-sanity gate as the presumed cause. Doctor checks lose their value when tests can write into their evidence.

## Suggested fix

Set `GBRAIN_AUDIT_DIR` to a per-run temp directory in the shared test bootstrap (one env line), or have the audit logger honor an existing test-environment signal. Either closes the leak for every audit-emitting test, not just this fixture.

Related: #1893 (noise on the same `content_sanity_audit_recent` surface from missing JSONL rotation — different cause, same diagnostic surface).

time-attack:
[...]

PR #3399, opened 2026-07-26
feat(usage): provider-boundary lifecycle ledger + get_usage coverage-contract report

Masashi-Ono0611:
Rebuilt from scratch on current master. Related to #3392 / #3425.

## Why the rebuild

The first cut of this PR recorded usage inside `gateway.chat()` only. Reconciling against my Anthropic Console invoice showed it captured **48% of real spend**: the default subagent path (`agent.use_gateway_loop` unset) calls the Anthropic SDK directly and never reaches the gateway. Two more defects compounded it — the report priced Sonnet 5 at exactly 1.5× the billed (introductory) rate, and failed calls were logged with the budget-tracker's worst-case *ceilings* presented as observations. Details and measurements are in the Draft-revert comment above.

Patching those one by one would have left the shape that caused them. The rebuild changes the shape.

## Design

**Record where the money leaves the building.** Every chat API call crosses exactly one provider boundary exactly once, so the ledger (`chat_usage_log`, migration v126) is written at the boundaries themselves:

- `gateway.chat()`'s dispatch (all gateway-routed traffic, including the `claude-cli` adapter and the gateway-loop subagent mode), and
- the legacy Anthropic-direct `client.create()` loop in `handlers/subagent.ts` — the path v1 missed.

Whichever high-level loop invoked it, each call lands one row; the two modes of the subagent can never double-count.

**A lifecycle ledger, not a success callback.** The attempt row is opened *before* the provider call (`started`/`pending`) and finalized in the failure path too. `request_status` (what happened to the call) and `usage_status` (what we know about billed tokens) are separate columns, because "failed" does not imply "unbilled":

- provider returned usage → `final`
- error object carried real usage → `partial` (a billed lower bound)
- terminated with nothing → `unknown`, tokens **NULL — never 0, never an estimate**
- process died mid-call → the `started` row survives and is reported as an orphan at query time

**Per-row rate snapshots.** Costs are computed at write time from `model-pricing.ts` and frozen into the row (`rate_snapshot`/`rate_source`/`rate_version`), so later price-table edits can't rewrite history. Anthropic cache pricing uses the request's own TTL — a request carries one TTL for all its breakpoints, so the boundary passes it explicitly (`cache_write_ttl`); 5m prices at 1.25× input, 1h at 2×. Anything unverifiable — model not in the table (`claude-cli:*` subscription traffic, deliberately), cache tokens on a provider whose cache semantics aren't verified, unknown TTL — stores `cost_usd = NULL` and is counted as *unpriced*, not guessed.

**`get_usage` returns a coverage contract, not just totals.** The op reports token ground truth plus:

- `known_cost_lower_bound_usd` — always safe to quote
- `complete_calculated_cost_usd` — non-null **only** when the window has no gaps
- `coverage.status` — `complete_observed` / `partial`. The name is deliberate: completeness is judged over rows the ledger observed; a process that lost both ledger writes for an attempt leaves nothing any reader can count, and the status must not claim otherwise
- `coverage.gaps` — typed: `usage_unknown`, `usage_partial`, `orphaned_attempt`, `in_flight`, `pricing_missing`, `recorder_failures_this_process`
- `coverage.out_of_scope` — the declared unmetered spend paths (embeddings, rerank, transcription, the offline eval harness), sourced from the registry below
- `basis: published_rate_snapshot` — these are list-rate estimates; promotional/negotiated pricing, credits and tax are explicitly out of scope. The one number this report will never produce is a false "total".

**A tripwire against the next silent path.** `src/core/ai/provider-call-registry.ts` classifies every file that can spend provider money (metered / unmetered-with-reason), and `test/provider-call-registry.test.ts` scans `src/` for runtime SDK imports, provider-endpoint literals and claude-CLI subprocess spawns — an unclassified hit fails CI, forcing the metered/unmetered decision into review. It already earned its keep during development: it flagged `claude-cli-language-model.ts` (subprocess dispatch, no SDK import — invisible to my hand inventory). It is a tripwire, not a guarantee: it detects new files and new endpoint literals, not new call expressions inside already-registered files. Confining SDK construction to a single adapter layer would close that residual gap, but that's an architecture call I'd rather leave to you than smuggle into a contributor patch.

**House conventions kept**: telemetry is best-effort and can never fail or slow the user's call (same contract as `minion-spend.ts`); write failures are counted and surfaced as a coverage gap. Pending writes drain through the `background-work.ts` registry (5th sink) so PGLite disconnects stay safe (#1762 class). Phase attribution (`think`, `dream.*`, `facts.extract`, `brainstorm`/`lsd`) rides an AsyncLocalStorage label; subagent rows fall back to their minion job name at query time.

## What this deliberately does not do

- No estimation of promo pricing — Console reconciliation stays a human step against token ground truth.
- No metering of embeddings/rerank/transcription — declared out of scope in the registry and visible as such in every `get_usage` response, rather than silently absent.
- No streaming instrumentation — no streaming chat path exists in the codebase today; if one appears, the tripwire forces classification and the started/finalized model already fits it.

## Review

Two independent adversarial review rounds on the full diff, plus a design consultation before any code was written.

**Round 1 (Grok Build CLI)** — verdict fix-first; the second commit closes what it found: the `$N::text::jsonb` house form for `rate_snapshot` (the #2339 double-encode PGLite hides), the success path being able to record "final, 0 tokens, $0" when an SDK route omits usage, the TTL string failing OPEN to the 5m rate, abort classification diverging between the two boundaries, and the untested started-INSERT recovery path.

**Round 2 (Codex, gpt-5.6-sol/high, on the post-round-1 diff)** — verdict fix-first; the third commit closes: slash-spelled non-Anthropic models receiving Anthropic cache multipliers (provider now comes from `splitProviderModelId`, not punctuation), a transiently-failed terminal UPDATE leaving the row `started` forever (bounded retry + the finished latch resets on failure), subagent error finalization queued behind `releaseLease`, fabricated cache zeros when a subagent response carried no usage object, unbounded fire-and-forget writes (256-cap with dropped attempts counted), and two structural test gaps — production wiring through `reconfigureGatewayWithEngine` with no manual seam, and gateway-loop exclusivity (one `gateway.chat` row, zero legacy rows).

Findings answered with evidence rather than code: the process-global engine mirrors the gateway's own one-active-engine-per-process model rather than adding a new constraint; the `node_modules` deletion is the intentional counterpart of #3462/#3463; SDK-internal retries can't be per-attempt metered at this boundary, so the limitation is stated permanently in the coverage notes (sums remain valid lower bounds).

## Verification

- `bun run typecheck` clean; new tests: 27 (lifecycle/pricing/sanitize) + 9 (get_usage contract) + 6 (registry tripwire) + 7 (gateway boundary, real dispatch path via the transport seam) + 3 (subagent boundary, real handler via the fake-client seam, including "pricing actually lands").
- **Mutation-tested, six ways** — each mutation applied, detection confirmed, then reverted and re-verified green: recorder neutralized (6 failures), wrong 1h cache multiplier (2), gap classification removed (1), gateway boundary unwired (5), subagent boundary unwired (3), final→partial downgrade invariant removed (2).
- Local full-suite runs kept hitting the 600s per-shard wallclock cap on my (currently loaded) machine — exit 124, not test failures; the v126 migration replay passes in the shard logs. CI's matrix on this PR is the authoritative full-suite verdict, so I'm not claiming a local one.
- Live verification against my own brain + Anthropic Console follows as a comment once this has soaked a night on my install.


Masashi-Ono0611:
Front-loaded this onto my macmini install (tier-configured brain, real production `subagent_messages` data — not just the CI fixtures) and ran it live:

```
$ gbrain usage --since 2026-07-19T02:31:13Z
{
  "schema_version": 1,
  "since": "2026-07-19T02:31:13.000Z",
  "until": "2026-07-26T02:31:14.193Z",
  "by_model": [
    { "model": "anthropic:claude-sonnet-4-6", "tokens_input": 139, "tokens_output": 89805,
      "tokens_cache_read": 6603282, "tokens_cache_create": 2404945, "cost_usd": 12.347020350000001 },
    { "model": "anthropic:claude-sonnet-5", "tokens_input": 70, "tokens_output": 133613,
      "tokens_cache_read": 5451663, "tokens_cache_create": 1587077, "cost_usd": 9.591442650000001 }
  ],
  "total_cost_usd": 21.938463000000002,
  "unpriced_models": []
}
```

Numbers are consistent with the PR's documented scope (subagent/job-tracked spend only) — lower than my full Anthropic Console total for the same window, as expected since direct gateway callers outside `subagent_messages` aren't counted here. `gbrain doctor --fast` stayed healthy after applying. No issues found running against real data.

Masashi-Ono0611:
**Correction to my comment above** — converting this back to draft while I address a scope issue found after posting it.

The `$21.94` figure I reported is misleading as stated. I traced which call sites actually write to `subagent_messages` (the only data source `get_usage` reads) and found it's a much narrower slice of gbrain's real spend than the PR description implies:

- **Reaches `subagent_messages`**: the `patterns` cycle phase (the only cycle phase that dispatches via `queue.add('subagent', ...)` — `src/core/cycle/patterns.ts:180`), plus ad-hoc `gbrain agent run` / MCP `submit_agent` calls.
- **Calls `gateway.chat()` directly and is invisible to this report** (and to the audit JSONL, and to BudgetTracker): `synthesize`, `propose_takes`, `grade_takes`, `drift`, `calibration_profile`, `extract_atoms`, `synthesize_concepts`, `auto_think` (i.e. `models.tier.deep`'s main consumer), `brainstorm`/`lsd`, facts extraction, `think`/`query` CLI.

On our brain, `synthesize` alone is usually the single biggest nightly cost line — so `$21.94` for 7 days is a significant undercount of actual spend, not a representative total. I've updated the parent issue (#3392) with the full breakdown and a corrected proposal — either scope this report explicitly narrower (rename/redocument as covering only patterns+agent-run traffic) or instrument the shared `gateway.chat()` seam so every caller is captured. Converting to draft until that's resolved; will re-open for review once it's not misleading by omission.

Masashi-Ono0611:
Reconciled `gbrain usage`'s output against the Anthropic console's actual per-model invoice for the same day and found one instrumentation gap this PR's own coverage doesn't close yet: `expand()` (backs `gbrain query --expand` and search-time query expansion) and `generateOcrText()` (image OCR) call the AI SDK directly (`generateObject` / `generateText`) rather than going through `chat()`, so their spend stayed invisible to `chat_usage_log` even with this PR's `withChatPhase` instrumentation applied everywhere else (dream cycle phases, `think`, `brainstorm`/`lsd`, facts extraction).

Pushed a follow-up commit to this same branch (rather than opening a separate PR — it's a gap in this PR's own stated scope, not an independent change) that:
- Instruments `expand()`'s success and failure paths the same way `chat()`'s production call site does (real usage on success; pessimistic pre-call estimate on failure, same convention as `extractUsageFromError`'s doc comment describes).
- Instruments `generateOcrText()`'s success path only — its failure path intentionally stays un-recorded, since `importImageFile`'s `ocr_failed_other` routing depends on this function throwing on failure, and swallowing the error to log usage would change that contract.
- Adds `test/ai/gateway-expand-usage.test.ts` (3 tests: success, failure, and the no-engine-on-hand negative case), mirroring `gateway-chat-usage.test.ts`'s existing structure and test-transport-seam pattern (added a `_generateObjectTransport` seam analogous to the existing `_generateTextTransport` one).

Full suite (485 tests / 47 files) green, `bun run typecheck` clean, `bun run check:gateway-routed` OK. (Couldn't run `ci:local:diff`'s e2e phase locally — no Docker daemon on this machine — but nothing in this diff touches anything that phase would exercise beyond what the unit suite already covers.)

Masashi-Ono0611:
Follow-up to the comment above: also live-verified this against the real Anthropic API on my running install (not just the mocked-transport unit tests) — ran `gbrain query "..." --expand` once, which exercises `expand()`'s real success path end-to-end. `chat_usage_log` recorded it correctly:

```
model=anthropic:claude-haiku-4-5-20251001, tokens_in=275, tokens_out=42, succeeded=true
```

and `gbrain usage` picked it up as `$0.000485` for that call. Real spend, real row, confirms the instrumentation works outside the test harness too, not just against the stubbed transport.

Masashi-Ono0611:
The `test (6)` CI failure on the previous commit (5986f831) traced to a pre-existing bug unrelated to this PR's own diff, but exposed by it — recording it here in case it's useful context for review.

Root cause: adding `test/ai/gateway-expand-usage.test.ts` shifted `scripts/sharding.ts`'s weight-based bin-packing (new file → corpus-median fallback weight → different partition) enough that `test/minions-shell.test.ts` and `test/audit/audit-dir-preload.test.ts` landed in the same shard (6/10) for the first time, with `minions-shell.test.ts` running first. Its `'shell-audit: write'` describe block overrides `GBRAIN_AUDIT_DIR` per-test via `beforeEach`, but its `afterAll` unconditionally `delete`d the var instead of restoring whatever the shared test bootstrap (`test/helpers/audit-dir-preload.ts`) had already set. Since `bun test` runs an entire shard in one process, that delete leaked forward into `audit-dir-preload.test.ts`, which then saw `GBRAIN_AUDIT_DIR` as `undefined` and failed its own #2823 regression checks.

Fixed in `5986f831..91fefd2a` by mirroring the capture/restore pattern `test/gbrain-home-isolation.test.ts` already uses for `GBRAIN_HOME`: capture the preload's value once at module load, restore that value (not delete) in `afterAll`.

Verified by reproducing the exact failing shard locally (`scripts/test-shard.sh --dry-run-list 6 10`'s full file list run through `bun test` with both preloads): 1434 pass / 3 fail before the fix (matching the CI failure exactly), 1437 pass / 0 fail after. `bun run typecheck` clean.

Masashi-Ono0611:
Two follow-up checks before considering this done, since CI-green alone doesn't prove the original coverage goal is met:

**1. Resolved the `gbrain usage` total-vs-invoice gap mentioned upthread.** Queried `chat_usage_log` directly: `MIN(occurred_at) = 2026-07-26 23:01:20 JST` (14:01 UTC). The table didn't exist for the first ~14 hours of that UTC day (migration/patch landed partway through), so a `--since 2026-07-26T00:00:00Z` query was only ever going to capture the back half of the day — comparing that partial window against a full-day invoice total was apples-to-oranges, not an instrumentation accuracy problem. No further action needed here; a clean full-day reconciliation is possible once a day fully inside `chat_usage_log`'s coverage window closes out.

**2. Swept the whole `src/` tree for other direct-AI-SDK blind spots** (same shape as the `expand()`/OCR gap): `gateway.ts` is the only file that calls `generateText`/`generateObject`/`streamText` directly — every other file goes through gateway's own wrapped functions. Within `gateway.ts`, grepped every remaining raw call site: the only one left is `generateOcrText()`, which this branch already instruments (see the earlier commits). No other gaps found.

Masashi-Ono0611:
Ran an independent Codex review (gpt-5.6-sol, high reasoning effort, bounded diff-only pass) over the full PR diff before considering this ready. It found 4 warnings; fixed 3 in `ee56a5d4`, flagging the 4th here rather than fixing it in this pass.

**Fixed:**
1. `generateOcrText()` — failed OCR calls were completely unrecorded. My original comment claimed logging required swallowing the error; that was wrong. Now: catch, log `succeeded:false` with a pessimistic estimate, then re-throw — `importImageFile`'s `ocr_failed_other` routing still gets its throw, the failure is now also logged.
2. `expand()` — the try block kept running result-processing (dedup/filter) *after* the `succeeded:true` row was already recorded. A throw in that step would fall into the catch and log a second row for the same already-recorded call. Fixed by recording success only after the result is fully processed.
3. `expand()`'s catch block was logging failures under the raw/unresolved `getExpansionModel()` string even when `resolveExpansionProvider()` had already succeeded and the real `recipe:modelId` was known (a generation-only failure, not a resolution failure). Fixed by hoisting the resolved identity so the catch block can use it.
4. (coverage gap, not a bug) `generateOcrText()` called `generateText` directly instead of the existing `_generateTextTransport` seam `chat()` already uses, making it untestable without a real SDK call. Routed through the same seam.

**Not fixed — flagging for maintainer visibility:** Codex also flagged that `resetGateway()` clears `_pendingChatUsageWrites` without draining it first, which could in principle untrack a live insert before the registered shutdown drain sees it. This is in the original PR content (not something I touched), and `resetGateway()` is test-only today (not on the production CLI-exit path, which goes through `awaitPendingChatUsageWrites`/the background-work drain registry instead) — so real-world risk looks low, but a proper fix would need to make `resetGateway()` async or add an explicit pre-clear drain, which is a wider-blast-radius change (many test files call it synchronously in `afterEach`) than I want to make as a drive-by in this follow-up. Leaving this to your judgment on whether it's worth a dedicated pass.

Added `test/ai/gateway-ocr-usage.test.ts` (3 tests) and 2 more cases to `gateway-expand-usage.test.ts` (nonzero cache-create, and the double-record regression itself). Full verification: typecheck clean, 16/16 across the three gbrain#3392 test files, 533/533 across `test/ai/` + related gateway tests, CI shard 6 reproduced locally at 1418/1418, `check:gateway-routed` OK.

Masashi-Ono0611:
Filed the resetGateway()/_pendingChatUsageWrites draining gap as its own issue rather than folding it into this PR, given the ~67-file blast radius a proper async fix would touch: #3425

Masashi-Ono0611:
Moving this back to Draft. I validated it against a real Anthropic Console bill and found two problems, neither fixed yet. Posting the measurements so the state is visible rather than leaving a misleading PR open.

## 1. The default subagent path never reaches `gateway.chat()`, so this op cannot count it

`handlers/subagent.ts` routes through `gateway.toolLoop()` only when `agent.use_gateway_loop` is `"true"` or `"1"`. With the flag unset — the default, and the case on my install (`gbrain config get agent.use_gateway_loop` returns empty) — it takes the legacy Anthropic-direct branch and calls `client.create()`, bypassing `chat()` entirely.

That makes this claim in the `get_usage` JSDoc wrong as written:

> calls dispatched off the minion job queue … lose their `phase` attribution … but they do NOT lose cost accuracy; the row is still written

What I observed:

- **Nightly dream run, 2026-07-27.** The subagent audit JSONL records 31 completed LLM turns totalling 3,482,343 cache-read and 884,407 cache-write tokens. `chat_usage_log` gained no rows carrying cache counters in that window.
- **Isolated repro.** Submitted one `subagent` job to a dedicated queue and drained it with a freshly spawned `gbrain jobs work --queue probe`. The job completed successfully; no `chat_usage_log` row appeared.
- **Whole table.** No row has a non-zero cache counter, so the cache-rate pricing this PR adds has not been exercised against stored data.

I have not yet run the A/B with `agent.use_gateway_loop=true`, so I am not claiming the gateway-loop path is broken — only that the default path is uncovered and the doc oversells the coverage.

## 2. A static price table cannot back an "actual spend" report during a promotional window

`CANONICAL_PRICING` deliberately holds standard rates, which is correct for its stated purpose. But Anthropic is currently billing Sonnet 5 at introductory $2/$10 per MTok through 2026-08-31 (with the usual 0.1× read / 1.25× 5-minute write multipliers ⇒ $0.20 / $2.50). Those are exactly 1.5× below the table's entries in every category, so anything this op reports for Sonnet 5 during the promo is inflated by 1.5×.

Pricing all tokens I observed for one day (1,183,814 input / 122,974 output / 3,482,343 cache-read / 884,407 cache-write):

| | cost |
|---|---|
| at introductory rates | $6.5049 |
| at `CANONICAL_PRICING` rates | $9.7573 (= 1.5×) |
| Console billed total, Sonnet 5, same day | $6.50 |

Caveat on that last row: the $6.5049 figure includes 7 failed calls whose token counts are `_extractUsageFromError` ceilings rather than measured usage ($0.7121 of the total). Measured-only is $5.7927, and this op excludes `succeeded=false` rows anyway. So the Console line should be read as *consistent with* the introductory rates, not as a reconciliation — I did not control for billing timezone, credits, or 1-hour cache writes. The 1.5× table-vs-introductory ratio is exact and independent of that caveat.

Swapping the entry to introductory numbers would just invert the bug after 2026-08-31. A correct actual-spend report needs effective-dated rates, or the applied unit prices (or computed cost) snapshotted on each row at write time.

## Before I ask for another look

- Integration test covering the default (legacy) subagent mode, not just the gateway-loop path.
- A/B with `agent.use_gateway_loop=true` to establish what the gateway path actually records.
- Make accounting-write failures observable instead of silently swallowed.
- Effective-dated or per-row snapshotted pricing.

If you would rather not carry this, say so and I will close it with these findings instead of leaving it in Draft.

time-attack:
Closing this one. It is a draft that adds ~1750 lines of new feature surface — a cache-rate pricing model plus a new actual-spend reporting op — and we hold a default-close bar on new features unless they fix a broken core promise or a user-blocking gap. Spend visibility is genuinely useful, but it is not something we are taking on as new surface area right now. Thank you for the careful work on the pricing derivation; the canonical-table discipline in it was clearly considered.

PR #3400, opened 2026-07-26
fix: preserve source scope and recover deferred embeddings

justinr1234:
## Summary

- preserve explicit/ambient `source_id` through `sync_brain` and source-scoped imports
- prevent cross-source same-slug reads from versioning the wrong page
- report when large incremental syncs defer inline embeddings
- enqueue one durable, source-scoped embed backfill after automatic deferral
- enqueue a backfill when an otherwise up-to-date source still has missing embeddings

## Why

A federated `sync_brain` could ingest into `default` or read a same-slug page from another source. Large syncs could also defer embeddings without leaving durable recovery work, making newly indexed pages invisible to vector retrieval.

## Verification

- `bun test test/sync-operation-source.serial.test.ts test/sync.test.ts test/source-id-routing.test.ts` — 80 passed
- `bun run verify` — 31/31 checks passed
- `git diff --check` — passed
- live source-scoped `sync_brain` advanced `devdocs-ops` and retrieval returned the new provider records from that source

## Safety

- backfill jobs are source-scoped and deduplicated by the existing submit helper
- explicit `--no-embed` remains respected and does not enqueue automatic work
- `.codex/` local configuration is excluded

### DIFF
--- a/scripts/check-wasm-embedded.sh
+++ b/scripts/check-wasm-embedded.sh
@@ -27,6 +27,15 @@ trap 'rm -f "$OUT_BIN"' EXIT
 # chunker + WASM path resolution, not unrelated CLI wiring.
 bun build --compile --outfile "$OUT_BIN" scripts/chunker-smoketest.ts >/dev/null 2>&1
 
+# Some hardened macOS hosts kill Bun 1.3.12 compiled binaries because that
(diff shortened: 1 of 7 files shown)


justinr1234:
Closing per project direction: these fixes are moving to the DevDocsorg-owned gbrain fork, which will be our authoritative runtime rather than an upstream contribution.

PR #3416, opened 2026-07-26
docs: correct Postgres schema templating comment

jack60612:
## What changed

Correct the `initPostgres()` comment describing PostgreSQL embedding-dimension handling.

## Why

The comment still said that PostgreSQL executed the static `vector(1536)` schema and did not support alternate dimensions. Current behavior is different: `PostgresEngine.initSchema()` resolves the configured model and dimensions and calls `getPostgresSchema()`, which templates the static schema source before execution.

The stale explanation can lead operators to add unnecessary schema-migration workarounds for fresh ZeroEntropy or Voyage installations.

## Impact

Documentation-only change. Runtime behavior is unchanged.

## Validation

- Traced `initPostgres()` through `PostgresEngine.initSchema()` to `getPostgresSchema()`.
- Confirmed existing unit coverage asserts the default PostgreSQL schema becomes `vector(1280)` and explicit dimension overrides are preserved.
- `git diff --check`

### DIFF
--- a/src/commands/init.ts
+++ b/src/commands/init.ts
@@ -1108,12 +1108,12 @@ async function initPostgres(opts: {
 
   // v0.37.10.0 T6 (D11) + v0.37.11.0 Lane B.2: ALWAYS configure gateway BEFORE
   // initSchema. Same preflight contract as PGLite. Refuse to call initSchema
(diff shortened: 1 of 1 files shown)


jack60612:
hi from startup school :)

PR #3420, opened 2026-07-26
fix(sources): recover corrupted config shapes

### LINKED ISSUE #3401: localFederatedSourceIds reads config->>'federated' with no shape coercion — a non-object sources.config silently collapses every stdio-MCP read to `default`
### Environment
gbrain 0.42.66.0 · Postgres engine · local **stdio** MCP (`gbrain serve`) · 6 sources · Linux

### What happens
The federated-scope resolver queries the raw JSONB with `->>`:

```ts
// src/core/source-resolver.ts:409-419
rows = await engine.executeRaw<{ id: string }>(
  `SELECT id FROM sources WHERE config->>'federated' = 'true' AND archived = false ORDER BY id`,
);
…
const ids = [sourceId, ...rows.map((r) => r.id).filter((id) => id !== sourceId)];
return ids.length > 1 ? ids : undefined;
```

`config->>'federated'` yields `NULL` on any row where `config` is a **JSON string scalar** or an **array** — both documented historical corruptions of this exact column (see the `#2829` note at `src/commands/doctor.ts:586-594`, and #2251 for how arrays arise). Those rows don't match, `ids.length` falls to 1, the function returns `undefined`, and `federatedSearchScope` (`src/core/operations.ts:594-609`) never widens. **Every stdio MCP read collapses to the single `default` source.**

The failure is invisible from every user-facing surface, because the **read** path coerces and the **resolver** path does not:

```ts
// src/core/sources-load.ts:101-109 — parseSourceConfig unwraps nested JSON strings
export function parseSourceConfig(config: unknown): Record<string, unknown> {
  const { value, layers } = unwrapConfigLayers(config);
```

So `gbrain sources list` / `sources_list` cheerfully report `federated: true` for a source the resolver has already excluded.

### What I expected
Either the resolver tolerates the same shapes `parseSourceConfig` tolerates, or the mismatch is loud.

### Minimal repro
Zero setup — the resolver's exact query against three shapes, all nominally `federated: true`:

```sql
WITH sources(id, config, archived) AS (VALUES
  ('default',  '{"federated":true}'::jsonb,            false),
  ('source-a', '"{\"federated\":true}"'::jsonb,        false),   -- string scalar
  ('source-b', '["{}", {"federated":true}]'::jsonb,    false)    -- array (per #2251)
)
SELECT id FROM sources WHERE config->>'federated' = 'true' AND archived = false ORDER BY id;
```
```
 id
---------
 default
```

Only `default` matches → `ids.length === 1` → `undefined` → no widening.

### Why it matters
This is a **silent, total** loss of cross-source retrieval on the stdio transport — agents get one source's worth of context and nothing anywhere says so. It is silent in both directions: the triggering corruption is only visible via `gbrain doctor`, and `sources list` actively contradicts the real behaviour. It survives indefinitely, since nothing on the read path rewrites the row.

Observed on a 6-source brain: MCP reads served only `default` for an extended period while `sources list` showed `federated: true` for every source. Repairing the config shape (`gbrain sources federate <id>`, which routes through `normalizeSourceConfig`) restored cross-source retrieval immediately — the widening logic itself was never at fault.

### Suggested fix
Reuse the coercion the engine already has. `postgres-engine.ts:1439-1455` normalizes every bad shape inline via `CASE jsonb_typeof(config)`; the resolver can do the cheap half:

```sql
SELECT id FROM sources
WHERE (CASE WHEN jsonb_typeof(config) = 'string'
            THEN NULLIF(config #>> '{}', '')::jsonb
            ELSE config END) ->> 'federated' = 'true'
  AND archived = false
ORDER BY id
```

Belt-and-braces: have `sources list` surface the *effective* federation state, so the two paths can't silently disagree again.

### LINKED ISSUE #3402: sources archive/restore use a raw `config || '{"federated": …}'` merge — on a string-shaped config this produces a 2-element array, not a merge
### Environment
gbrain 0.42.66.0 · Postgres engine · Linux

### What happens
Both archive writers merge with a bare JSONB `||`:

```sql
-- src/core/destructive-guard.ts:193 (softDeleteSource)
config = COALESCE(config, '{}'::jsonb) || '{"federated": false}'::jsonb
-- src/core/destructive-guard.ts:235 (restoreSource)
config = COALESCE(config, '{}'::jsonb) || $1::jsonb
```

Postgres `||` between a JSONB **string** and a JSONB **object** is *concatenation*, not merge — it yields an array:

```sql
SELECT ('"{}"'::jsonb || '{"federated": false}'::jsonb) AS v,
       jsonb_typeof('"{}"'::jsonb || '{"federated": false}'::jsonb) AS t;
--          v                          |   t
-- ["{}", {"federated": false}]        | array
```

This hazard is documented in this repo, one file over:

```ts
// src/core/postgres-engine.ts:1429-1431
// `||` between JSONB object + JSONB string yields a
// JSONB ARRAY (concat semantics for non-matching types), which
// wipes every existing config key.
```

`updateSourceConfig` was hardened against exactly this with a `CASE jsonb_typeof(config)` coercion (`postgres-engine.ts:1439-1455`). `softDeleteSource` / `restoreSource` were missed and still use the raw form.

### What I expected
Archiving a source sets `federated: false` and leaves the rest of `config` intact, whatever shape the row started in.

### Minimal repro
```bash
# given a source whose config is already string-shaped (the #2829 / #2251 shape,
# detectable via `gbrain doctor` → source_config_shape warn)
gbrain sources archive source-a
# then: SELECT jsonb_typeof(config) FROM sources WHERE id='source-a';  → 'array'
```

### Why it matters
It converts a *recoverable* corruption into an *unrecoverable* one. `normalizeSourceConfig` (`src/core/sources-load.ts:84-96`) unwraps string layers but returns `{}` for an array — so the next config write silently discards `remote_url`, `federated`, and every other key. `doctor`'s suggested repair SQL is also a no-op on arrays (filed separately). And per #2251, arrays are the shape that makes `updateSourceConfig`'s own array branch throw `jsonb_each on a non-object` when any element isn't an object, which can wedge the dream cycle's `last_full_cycle_at` for that source.

Practical consequence: `sources archive`/`restore` is the intuitive way an operator would try to "re-stamp" a source to fix a config warning, and it makes things strictly worse.

### Suggested fix
Route both writers through the existing normalizer, as every other `sources` config writer already does (`src/commands/sources.ts:715, 902, 958, 982, 1007, 1023`) — or inline the same guard used at `postgres-engine.ts:1439`:

```sql
config = (CASE WHEN jsonb_typeof(config) = 'object' THEN config
               WHEN jsonb_typeof(config) = 'string' AND (config #>> '{}') IS JSON
                 THEN (config #>> '{}')::jsonb
               ELSE '{}'::jsonb END) || '{"federated": false}'::jsonb
```

### LINKED ISSUE #3403: doctor's source_config_shape remediation SQL is a no-op on the array shape it detects
### Environment
gbrain 0.42.66.0 · Postgres engine · Linux

### What happens
The check detects any non-object config…

```ts
// src/commands/doctor.ts:601
`SELECT id, jsonb_typeof(config) AS typ FROM sources WHERE jsonb_typeof(config) <> 'object'`
```

…and prints the same repair for all of them:

```ts
// src/commands/doctor.ts:618-621
`UPDATE sources SET config = (config #>> '{}')::jsonb ` +
`WHERE jsonb_typeof(config) <> 'object';`
```

`#>> '{}'` extracts the value as text and re-casts. For a **string** that unwraps one layer correctly; for an **array** the text is the array literal, which re-parses as the same array:

```sql
SELECT ('["a","b"]'::jsonb #>> '{}')::jsonb AS v,
       jsonb_typeof(('["a","b"]'::jsonb #>> '{}')::jsonb) AS t;
--       v        |   t
-- ["a", "b"]     | array     ← unchanged

SELECT jsonb_typeof(('"{\"federated\":true}"'::jsonb #>> '{}')::jsonb);
-- object     ← the string branch works
```

Array is not a hypothetical shape here — it is the documented outcome of the naive `||` merge (#2251, and the comment at `postgres-engine.ts:1429-1431`).

### What I expected
Either the printed SQL repairs both shapes it detects, or it is scoped to the shape it actually repairs.

### Minimal repro
Run the two `SELECT`s above — no brain state needed. Or: `gbrain doctor` on a brain with an array-shaped `sources.config`, run the suggested `UPDATE`, re-run `gbrain doctor` → same warning, unchanged.

### Why it matters
An operator follows the tool's own instructions, sees `UPDATE 1`, and reasonably concludes it is fixed. It isn't, and the check still warns — which reads as a false positive in `doctor` rather than a failed repair. The array shape is also the more damaging one (`normalizeSourceConfig` returns `{}` for arrays, dropping every key), so it is the case that most needs a working repair.

### Suggested fix
Either scope the suggestion honestly:

```ts
`WHERE jsonb_typeof(config) = 'string';`
```

or emit the flattening repair the engine already implements at `postgres-engine.ts:1447-1453`:

```sql
UPDATE sources SET config = COALESCE(
  (SELECT jsonb_object_agg(kv.key, kv.value)
     FROM jsonb_array_elements(config) elem, jsonb_each(elem) kv), '{}'::jsonb)
WHERE jsonb_typeof(config) = 'array';
```

Guard the array branch against non-object elements — that is the #2251 throw.

arisgysel-design:
## Summary

- make federated source resolution use the same tolerant config parser as user-facing source reads
- normalize nested-string and array-shaped config atomically before config writes and archive/restore
- make the doctor remediation repair arrays instead of reporting a successful no-op
- share the SQL coercion across PGLite, Postgres, archive/restore, and doctor

## Verification

- `bun run typecheck`
- 80 focused tests across source loading, federation, archive/restore, engine config writes, and doctor
- `bun run verify`: 30/31 local checks passed; `check:wasm` was killed with exit 137 on this Mac after its compile phase, matching the local runner limitation seen on the prior PR

Fixes #3401
Fixes #3402
Fixes #3403


time-attack:
Pushed a repair commit (9d7f10b4) to this branch (maintainer-edit):

- **Fix:** `restoreSource` bound the federated patch (a JS string) to a bare `$1::jsonb` — the #2339 double-encode class on the Postgres engine. postgres.js binds it as a jsonb string scalar, so `coerced_object || string-scalar` evaluates as array-concat and restore could re-corrupt the exact shape this PR repairs. PGLite masks it. Now binds through `$1::text::jsonb` per the repo JSONB rule.
- **Coverage:** new DATABASE_URL-gated Postgres regression `test/e2e/restore-source-config-jsonb-postgres.test.ts` — seeds a string-scalar config, runs archive → restore, asserts `jsonb_typeof(config) = 'object'` + federated flag + key preservation. Verified red on the bare bind (config became a jsonb array) and green on the fix against a real pgvector container; skips cleanly without DATABASE_URL.

Gates: typecheck clean, all 5 touched unit test files pass (80/80), both jsonb guard scripts clean.

PR #3424, opened 2026-07-27
fix(synthesize): dedupe successful transcripts after corpus moves

zsimovanforgeops:
Synthesis idempotency keys currently include the absolute transcript path. Moving `dream.synthesize.session_corpus_dir` therefore gives unchanged successful transcripts new keys and can trigger another paid synthesis run.

Operators can spot the condition with:

```sql
SELECT id, idempotency_key, status, result
FROM minion_jobs
WHERE name = 'subagent'
  AND idempotency_key LIKE 'dream:synth:%'
ORDER BY id DESC;
```

Duplicate work appears as successful rows with the same filename and trailing content hash under different corpus roots.

The legacy completion check now loads successful `end_turn` rows once per source and matches the full filename plus content hash. New producer keys use a disjoint namespace containing the percent-encoded source, complete filename, and hash; chunk indices keep their existing suffix. Existing jobs and pages remain valid, so no repair or schema migration is required.

The moved-root regression fails on master with one new child and passes with zero. The existing single- and multi-chunk key-shape tests pin the path-independent namespace.

Tests: `bun run verify`; `bun run build`; `bun test test/e2e/dream-synthesize-chunking.test.ts`; `bun test test/e2e/multi-source-bug-class.test.ts`.


time-attack:
Maintainer repair pushed (2cac1791, after merging master which now carries #3414):

1. **Legacy CHUNKED completions now dedupe.** The scan only matched keys ending `:<hash16>` (legacy single-chunk). Transcripts synthesized under the pre-v2 chunked family `dream:synth:<path>:<hash16>:c<i>of<n>` would re-run as a full paid v2 synthesis post-upgrade. `findLegacyCompletion` now also matches the chunked family and counts a transcript as done only when the full chunk set `c0..c(n-1)` completed; partial sets fall through to a fresh v2 run.

2. **Reverted the `result->>'stop_reason' = 'end_turn'` filter to plain `status = 'completed'`.** Two reasons: (a) it changed suppression semantics — legacy completed jobs with any other stop_reason went from suppressed (queue-level idempotency blocks completed re-submission regardless of stop_reason, pinned in test/minions.test.ts) to re-running at cost, and that tightening wasn't documented as intended; (b) the naive `->>` read misses double-encoded jsonb `result` rows (the same class the file already COALESCE-defends for `input`), which would have re-run those too. Dropping the filter restores the old cost-safe behavior and moots the double-encoding read. If semantic-success filtering is actually wanted here, it needs the `COALESCE(result->>'stop_reason', (result #>> '{}')::jsonb->>'stop_reason')` form and a stated rationale.

New tests: legacy chunked full-set suppression + partial-set resubmission; double-encoded result row still recognized. Gates: typecheck clean, touched e2e files green, jsonb pattern/params guards clean.

PR #3426, opened 2026-07-27
feat(sync): --missing-path skip — classify absent-local_path sources in --all instead of failing

Lazydayz137:
## Summary

`sync --all` treats a source whose `local_path` does not exist on this machine as a hard per-source failure and exits 1 — on every run. `sources.local_path` is machine-specific state in a brain-wide table, so any brain whose sources were registered from more than one machine — or a sanctioned setup mid-migration (topologies.md Topology 2, or the system-of-record git flow before every repo is cloned to a new box) — reports phantom failures forever.

Observed on a real fleet: **12 phantom ✗ per hourly cron tick, rc=1 with zero actionable signal**, which trains operators to ignore the exit code — the worst property a cron can have.

## Fix

- New flag on `sync --all`: `--missing-path <fail|skip>`, **default `fail`** (behavior unchanged — on a single-machine brain a missing path usually means an unmounted volume or deleted checkout; staying loud is correct, skip is explicit opt-in).
- `skip` classifies those sources honestly: `⊘` line in the human aggregate, status `skipped_missing_path` (+ `local_path`) in the `--json` envelope, new `skipped_count`; excluded from `error_count` and the exit-1 gate.
- Pure helpers `parseMissingPathMode` + `partitionMissingPathSources` exported for tests (sync-all-parallel style: no DB, no fs — classification driven by an injected `pathExists`).
- Passing the flag without `--all` warns instead of silently no-oping.
- Docs updated in the same PR: `sync --help`, `docs/TESTING.md` inventory, `docs/architecture/KEY_FILES.md` sync entry. CHANGELOG/VERSION deliberately untouched per the release process.

## Tests

- `bun test test/sync-all-missing-path.test.ts test/sync-all-parallel.test.ts` → **26 pass / 0 fail**
- `bunx tsc --noEmit` → clean
- `bun run verify` → **31/31 checks green**
- Note: `test/sync-break-lock-all.test.ts` shows a pre-existing `beforeEach` 5s hook timeout in my environment that reproduces identically on master — unrelated to this change.


time-attack:
Repaired the conflict vs master (maintainer push, `maintainerCanModify`):

- Merged `origin/master` in twice (a second merge picked up #3414, which landed mid-repair).
- `src/cli.ts`: kept both the `--missing-path` help lines and master's `autopilot --install` note (`sync --install-cron` was removed on master).
- `docs/architecture/KEY_FILES.md`: master's current `sync.ts` entry with this PR's `--missing-path` clause and the `skipped_count` envelope field folded in (current-state prose only).
- `src/commands/sync.ts` auto-merged cleanly — the `runnableSources` partition still feeds parallel-eligibility, the worker-cap math, and the `--json` envelope; master's #3068 `pull_failed` exit gate coexists with the skip classification.

Gates: `bun run typecheck` exit 0; full `test/sync*` suite 462/463 pass (the one failure is a beforeEach migration-setup timeout under suite load — passes in isolation); `test/sync-all-missing-path.test.ts` and `test/build-llms.test.ts` green; `bun run build:llms` produced no diff.

PR #3432, opened 2026-07-27
Support client-local source checkout bindings

AllocatorXy:
## What changed

- Add a public process-local `GBRAIN_SOURCE` + absolute `GBRAIN_SOURCE_PATH` binding for one resolved source.
- Keep shared source identity, remote identity, bookmarks, and legacy `sources.local_path` behavior unchanged when no binding is supplied.
- Use the resolved source consistently for sync operations and brainstorm/LSD saves.
- Keep explicit sync roots above client bindings during DB-only re-export, and reject queued sync when a process-local checkout cannot follow the job.

## Why

Clients that share one Postgres database and one Git brain remote can clone that remote at different absolute paths. A shared `sources.local_path` cannot safely represent those per-client checkouts.

## User impact

Each client can bind its own checkout without writing that path back to shared Postgres. Inline sync, source status/pull/hardening, and write-through use the matching local checkout while preserving existing fallback behavior.

## Validation

- 132 focused source, write-through, operation, trigger, and save tests
- 20 focused durability tests
- 10 focused sync cost/routing tests
- 2 focused DB-only reconciliation tests
- TypeScript typecheck and test-isolation guard
- Downstream single phased Codex product E2E with two clients at different paths, one shared Postgres, one Git remote, correction, synchronization, and fresh-Postgres recovery

The full upstream suite was not run; validation was intentionally scoped to the changed public seam and downstream product gate.

### DIFF
--- a/docs/guides/multi-source-brains.md
+++ b/docs/guides/multi-source-brains.md
@@ -95,6 +95,27 @@ So inside `~/.gstack/plans/` on a brain that pinned `gstack` to
 the `gstack` source. Outside any registered directory with no env/dotfile
 set, it writes to the default.
 
+## Client-local checkout binding
+
+Clients that share one database and Git remote can keep the same source
+identity while cloning it to different local paths. Set both variables for
+the process running a single-source operation:
+
+```bash
+export GBRAIN_SOURCE=wiki
+export GBRAIN_SOURCE_PATH=/absolute/path/to/this-client/brain-wiki
+```
+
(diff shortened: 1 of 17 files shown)


time-attack:
Reviewed and declined for now: the client-local checkout binding introduces a public env contract covering only part of the local_path consumer surface; the engineering was solid but the tenancy answer stays brains/sources for now. Thank you — the fail-closed design here was genuinely well done.

PR #3433, opened 2026-07-27
fix(admin): deduplicate concurrent login submissions

HammerTech-Z:
## Summary

- Add a single-flight guard to the Admin `LoginPage` submission path.
- Concurrent submissions trigger only one `/admin/login` request.
- A new login submission is allowed after the prior request succeeds or fails.

## Validation

- `bun test test/admin-login-submit-once.test.ts`
- `bun run typecheck`
- `git diff --check`

This pull request is intentionally opened as a draft.

### DIFF
--- a/admin/src/lib/login-submit-once.ts
+++ b/admin/src/lib/login-submit-once.ts
@@ -1,0 +1,24 @@
+type AsyncOperation<Args extends unknown[], Result> = (...args: Args) => Promise<Result>;
+
+export function createLoginSubmitOnce<Args extends unknown[], Result>(
+  operation: AsyncOperation<Args, Result>,
+): AsyncOperation<Args, Result> {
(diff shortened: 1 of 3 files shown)


time-attack:
Closing after verification: the double-submit path this guards can't occur — Login.tsx sets loading=true synchronously in the submit handler and the sole submit button renders disabled (React 19 discrete events flush before the next event dispatches; HTML implicit submission won't fire on a disabled default button; api.login has exactly one call site). Server-side, a duplicate POST /admin/login is also harmless — it just mints a second session cookie. Appreciate the defensive instinct, but this would be speculative hardening on a race with no reproducible path and no consequence. Note if you revisit: the coalescing wrapper ignores arguments, so a second submit with a corrected token would receive the first request's result.

PR #3280, opened 2026-07-23
fix(extract,cli,pricing): link-aware timeline delimiter, dispatchable backfill, zerank-2 rerank pricing (nwf2)

time-attack:
Backlog fix wave nwf2 — three verified items, one commit each.

## fix(extract): link-aware Source — Summary delimiter in timeline bullets
Takeover of #3060 (code by @wright-io, credited via Co-authored-by — not closing, just referencing). The old bullet regex split on any dash, so a bare hyphen inside a markdown link target shattered one timeline entry into two fragments that re-inserted on every sync. Delimiter search is now link-aware (whitespace-flanked dash outside bracket/paren spans); delimiterless bullets are kept whole under the 'markdown' source label. Repaired from the original PR: real-person test fixtures renamed to privacy-rule placeholders (alice-example / acme-example), title format corrected.

## fix(cli): make `backfill` dispatchable
Fixes #3224 (fix approach by @Masashi-Ono0611 in #3251, credited via Co-authored-by). `backfill` was missing from CLI_ONLY, so dispatch hit 'Unknown command' before the implemented `case 'backfill'` could run. Registered in CLI_ONLY + CLI_ONLY_SELF_HELP; dispatched before handleCliOnly's shared `connectEngine()` because `runBackfillCommand` manages its own engine end-to-end (avoids a PGLite single-writer double-connect); refused on thin clients with a hint; plus a class-guard test that every switch case is CLI_ONLY-registered.

## fix(pricing): price `zeroentropyai:zerank-2` reranks
Fixes #3223 (fix approach by @Masashi-Ono0611 in #3233, credited via Co-authored-by). `lookupPricing`'s rerank branch only consulted ANTHROPIC_PRICING and the free-local set, so the tokenmax-default zerank-2 reranker returned null pricing and every `--max-cost` rerank reserve TX2 hard-failed. Adds the $0.025/1M entry to EMBEDDING_PRICING and makes the rerank branch fall back to `lookupEmbeddingPrice` before the free-local check.

## Verification
- `bun run typecheck` exit 0
- `bun test test/extract.test.ts test/core/budget/budget-tracker.test.ts test/thin-client-routing-audit.test.ts` — 85 pass, 0 fail
- `bun test test/cli-backfill-dispatch.test.ts` — 5 pass, 0 fail (includes real PGLite init + dry-run backfill, no lock hang)
[...]

PR #3441, opened 2026-07-27
fix(extract): timeline bullet parser shatters entries on dashes inside markdown links

wright-io:
## Problem

`extractTimelineFromContent`'s Format-1 regex uses a non-greedy source group with a `[—–-]` delimiter class, so the Source→Summary split lands on the **first** dash after the pipe — including hyphens inside link targets (`../people/pedro-franceschi.md`) and spaced dashes inside link labels (`[Deals — Q1 Review](...)`).

Each affected bullet produces a corrupted `(source, summary)` pair. Because the `(page_id, date, summary, source)` uniqueness constraint sees every fragment shape as a new tuple, re-extraction on each sync re-inserts fragments **alongside** prior rows, accumulating garbage on any page whose markdown changed.

## Fix

Match the bullet up to the pipe, then find the delimiter with a **link-aware scan** — only a whitespace-surrounded dash that sits *outside* every `[label](target)` span splits Source from Summary. Delimiterless bullets (e.g. auto-generated backlink lines `- **date** | Referenced in [X](y.md)`) are kept whole as the summary with source `markdown`, instead of being dropped or shattered.

## Tests

Adds cases to `test/extract.test.ts` covering hyphens-in-targets, spaced-dashes-in-labels, and delimiterless backlink lines.

```
bun test test/extract.test.ts
33 pass · 0 fail · 72 expect() calls
```

Verified against current `master`.

https://claude.ai/code/session_019PQjMs8qgu4hWHDCLUU1eP

### DIFF
--- a/src/commands/extract.ts
+++ b/src/commands/extract.ts
@@ -472,15 +472,53 @@ export async function extractLinksFromFile(
 
 // --- Timeline extraction ---
 
+/**
+ * Index of the first dash (—, –, -) that can serve as the Source — Summary
+ * delimiter: it must have whitespace on both sides and sit outside every
+ * markdown-link span. Hyphens inside link targets
+ * (`../people/pedro-franceschi.md`) and dashes inside link labels
+ * (`[Deals — Q1 Review](...)`) are content, not delimiters — splitting on
+ * them shatters one entry into two fragments whose halves re-insert on
+ * every sync (the (page_id, date, summary, source) uniqueness sees each
+ * fragment shape as a new row). Returns -1 when the line has no delimiter.
+ */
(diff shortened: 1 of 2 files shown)


time-attack:
Closing as a duplicate of your own earlier #3060, which carries a byte-identical fix and is still open and mergeable. The bug is real — the non-greedy source group splits on hyphens inside link targets and shatters entries that then re-insert on every sync — and the link-aware delimiter scan is the right shape, so please keep #3060 as the live one. Note #3280 and #3288 also bundle this same delimiter change; consolidating on the smallest standalone PR will help it land.

Issue #160, opened 2026-04-16
extractAndEnrich writes pages from untrusted text without a gate


garagon:
## Summary

`extractAndEnrich` in `src/core/enrichment-service.ts` runs a regex over arbitrary text, pulls out anything that looks like a capitalized name, and unconditionally calls `engine.putPage()` on the owner's brain under `people/{slug}` or `companies/{slug}`. There is no confirmation step, no allow-list, no review queue, and the source of the text is whatever skill invoked the call — ingest, meeting-ingestion, idea-ingest, any of the recipes that forward raw email / transcript / pasted content.

The practical shape:

1. A meeting transcript, pasted article, or email comes in via one of the ingest skills.
2. The skill calls `extractAndEnrich(engine, text, sourceSlug)` as part of enrichment.
3. `extractEntities` greps for `\b[A-Z][a-z]+(?:\s+[A-Z][a-z]+){1,3}\b`, accepts any hit, and feeds each to `enrichEntity`.
4. If the entity page doesn't exist, `enrichEntity` creates it with `generateStubContent(name, type, context)` — where `context` is a 50-char window of the raw input text — and writes it to the brain as an authoritative page.

So the attacker-shaped case looks like: a hostile email (or a spoofed meeting transcript, or a poisoned CSV attached to a briefing) carrying the sentences

> *I had lunch with Paul Graham today. He mentioned that Ycombinator Inc. is pivoting to crypto.*

ends up creating `people/paul-graham` and `companies/ycombinator-inc` pages in the owner's brain, with the sentence copied in as the summary. The owner never typed that. Agents that query the brain later ("what does gbrain say about Paul Graham?") quote it back as ground truth.

This is a data-injection / authz gap, not an RCE. It's latent today because the current in-tree caller set is small, but the skill-driven ingest paths this wires into (meeting-ingestion, idea-ingest, media-ingest) all take external content by design.

## Why it matters

The brain's value is that it's trusted. Everything downstream — search, enrichment chains, agent responses — assumes pages in `people/` and `companies/` reflect the owner's observations. Once untrusted ingest gets unbounded write access to those namespaces, the trust model collapses silently: the owner can't tell which entries came from their own notes and which came from a sentence that happened to sit in an incoming email.

Secondary effects:

[...]

PR #3457, opened 2026-07-27
feat(doctor): silent-failure check batch — content-hash duplicates, undeclared DB-only pages, heartbeat staleness, db_only collector collision (#2250 #2784 #2787 #2788)

### LINKED ISSUE #2250: Feature: gbrain doctor should detect content-hash duplicate pages (bare-slug vs path-prefixed from wrong import root)
## Summary

`gbrain import` creates silent content-hash duplicates when accidentally run from the wrong root directory. If the import is rooted one level too deep (inside the `brain/` subdirectory instead of its parent), the path prefix is dropped from every imported slug. The result is both `people/brian-swem` (correct, path-prefixed) and `brian-swem` (bare slug, wrong) coexisting in the database with identical content.

`dream --phase purge` does not remove these because purge targets file-backed orphans whose source files are gone (confirmed in CHANGELOG v0.42.35.0). Bare-slug duplicates have valid source paths and are not orphans; they just have the wrong slug form. `gbrain doctor` currently provides no health check that surfaces this condition.

Adding a content-hash duplicate detection check to `gbrain doctor` would surface the problem immediately rather than requiring users to discover it through query anomalies (two results for the same entity, inflated page counts, wikilink resolution ambiguity).

## Reproduction steps

1. Set up a brain with correctly path-prefixed pages via `gbrain import /path/to/brain-root`. Pages get slugs like `people/brian-swem`, `projects/my-project`.
2. Accidentally run `gbrain import /path/to/brain-root/brain` (one level too deep; the `brain/` prefix is dropped from all slugs).
3. Observe: the database now contains both `people/brian-swem` and `brian-swem` with identical content. Same for all other page types.
4. Run `dream --phase purge`. Bare-slug duplicates are **not** removed.
5. Run `gbrain doctor`. No warning is surfaced.

## Expected behavior

`gbrain doctor` detects and reports content-hash duplicate pairs where one slug is bare and the other is path-prefixed:

```
⚠  2 content-hash duplicate pairs detected (same content, differing slug forms):
   brian-swem  ←→  people/brian-swem
   my-project  ←→  projects/my-project
   To resolve: gbrain pages delete <bare-slug> for each, then gbrain pages purge-deleted --older-than 0
```

## Actual behavior

`gbrain doctor` completes without any duplicate warning. Bare-slug pages persist indefinitely, producing:

1. Two results for the same entity in semantic queries
2. Inflated page counts in doctor health metrics
3. Wikilink ambiguity when both slug forms exist in the database

## Proposed fix / approach

A SQL health check in `gbrain doctor` that groups pages by `content_hash` and flags groups where the same hash appears under both a bare slug (no `/` separator) and a path-prefixed slug (contains `/`). Report as a structured warning with the resolution command.

Optionally, a `gbrain dedup --dry-run` / `--apply` subcommand for operational cleanup would pair well with this, but the doctor check alone addresses the visibility gap.

## Context

1. gbrain version: 0.42.36.0
2. Node: v24.14.1, Linux / WSL2
3. The wrong-root import mistake is easy to make when the `brain/` subdirectory and its parent have similar names in a shell prompt. The correct import command targets the parent of `brain/`.
4. `dream --phase purge` scope is intentionally limited to file-backed orphans (CHANGELOG v0.42.35.0). This is not a request to expand purge, just a request to make `gbrain doctor` visible about the duplicate condition.
5. Manual resolution once duplicates exist: `gbrain pages delete <bare-slug>` for each, then `gbrain pages purge-deleted --older-than 0`.

### LINKED ISSUE #2784: feat(doctor): warn on DB pages with no backing file outside declared db_only paths; ship derive-phase output prefixes (life/events, atoms, extracts, dream-cycle-summaries) as db_only defaults

A full DB-vs-file sweep of 6,913 pages on this brain found **1,805 DB-only pages outside any declared db_only path** — all of it engine derive-phase output: `life/events/` 1,323, `atoms/` 433, `extracts/` 33, `dream-cycle-summaries/` 10. These are re-derivable machine output by upstream design (we ratified keeping them DB-only), but they sit **undeclared**: not git-backed, not in `storage.db_only` (which has no built-in defaults — `storage-config.ts:157,221-226` reads only what `gbrain.yml` declares, defaulting to `[]`), and therefore invisible to any file-lane backup/recovery reasoning. An operator auditing "what would survive a DB loss" gets a silently wrong answer.

The same sweep also caught 4 genuinely-missing pages (a local worker defect on our side) — a doctor check would have surfaced those in a day instead of via a user-facing miss weeks later.

Verified at HEAD (a25209bb): no doctor check compares DB rows against backing files — the check registry (`doctor-categories.ts`) has nothing in this class, and doctor's checks are deliberately "Strictly SQL — no filesystem/git access" (`doctor.ts:3090`), so this needs a new check that is allowed to stat the repo.

1. **Doctor check:** a DB page outside every declared db_only path with no backing file in the source repo → warn (with a count + sample slugs, like existing checks).
2. **Ship the engine's own derive-phase output prefixes as declared-db_only defaults** (`life/events/`, `atoms/`, `extracts/`, `dream-cycle-summaries/`) so the declared list matches what the engine actually does, and the check above stays quiet on healthy brains.

Together these make the file-lane story auditable: everything is either file-backed, declared db_only, or flagged.

**Alternatives considered**

A hand-run read-only DB-vs-file sweep script (deployed locally; found the numbers above). Works, but it's exactly the kind of operator-side scaffolding a doctor check should replace.

---

Related: #2426 (db-only durability), #344 / #1856 (DB-only page/timeline → file drift family).

Environment: gbrain 0.42.57.0 (verified unchanged at 0.42.58.0 / a25209bb) · macOS 26.3.1 · Bun 1.3.13 · Supabase Postgres (PgBouncer transaction mode).

### LINKED ISSUE #2787: feat(integrations): heartbeat_max_age health-check type — doctor passes senses that have been dead for weeks (all shipped check types are point-in-time)

Recipe `health_checks` support only point-in-time types — `http`, `env_exists`, `command`, `any_of` (`src/commands/integrations.ts:108`, dispatch switch `:173-292`), plus a deprecated embedded-recipe string type. None is staleness-aware, and `gbrain integrations doctor` (`cmdDoctor`, `integrations.ts:687-708`) runs ONLY health_checks — it never reads an integration's heartbeat file (heartbeat age is read by `getStatus`/`cmdStatus`/`cmdList`/`cmdStats`, but never inside a health-check type and never in doctor).

Live consequence: a calendar sense that had been **16 days stale passed `integrations doctor` cleanly**. Its only health check (`recipes/calendar-to-brain.md:21-30`) is an `any_of` of `http GET <gateway>/health` OR `env_exists GOOGLE_CLIENT_ID` — both can stay green forever while zero data flows. This is the sense silent-death mode: the collector runs green, the checks pass, and the brain quietly stops learning. It took this deployment's senses down for weeks in June before local tooling existed.

(`cmdStatus` does print a hardcoded "no events in N days" note at `integrations.ts:676`, but it is not a declarable check type and doctor never surfaces it — the ask is specifically a *declarable staleness type* that `doctor` enforces.)

A `heartbeat_max_age` health-check type: the recipe declares its expected cadence (e.g. `max_age: 48h`), and `integrations doctor` reads the integration's heartbeat file age and fails past the threshold. Recipes that install crons already know their cadence — the declaration is one line.

**Alternatives considered**

A local sense-health script enforcing "silence > 2× cadence = BREACH" plus a daily alert cron (deployed here; it caught every subsequent stall). It works, but every operator rebuilding this same shim is exactly what a native check type should prevent.

---

Environment: gbrain 0.42.57.0 (verified unchanged at 0.42.58.0 / a25209bb) · macOS 26.3.1 · Bun 1.3.13 · Supabase Postgres (PgBouncer transaction mode).

### LINKED ISSUE #2788: feat(sync/doctor): warn when a collector's output dir is declared db_only — auto-gitignore makes sync AND import silently skip its files (7-week green-collector outage)

When an operator declares a collector's output directory in `storage.db_only`, the ingestion path for that collector silently dies — with everything reporting green:

- `manageGitignore()` runs on every successful sync and auto-appends each db_only dir to `.gitignore` (`src/commands/sync.ts:4513`, `:4598-4610`).
- Git-walking sync diffs via `git diff --name-status` (`sync.ts:875,893,1810`), so gitignored files never appear in the diff and never reach the DB.
- At HEAD, `gbrain import` ALSO honors `.gitignore` inside a work tree (`import.ts:582-592`, `git ls-files --exclude-standard`) — so even a manual re-import skips the files. The trap has no floor.

The calendar-to-brain recipe prescribes writing day FILES to exactly such a directory (`recipes/calendar-to-brain.md:50-66` → `daily/calendar/{YYYY}/{date}.md`) and relies on `gbrain sync` to ingest them (Step 7, `:262-273`). On this deployment, `daily/calendar/` had been declared db_only in a storage cleanup — and **no calendar page reached the DB for seven weeks** (2026-05-19 → 2026-07-08) while the collector ran green the whole time.

No warning exists anywhere for this combination: the only db_only diagnostics at HEAD is the tier-overlap warning (`storage-config.ts:262`), and `integrations.ts` has zero awareness of storage tiers.

A doctor check or sync-time warning when a configured recipe/collector's output path falls inside a db_only prefix — "collector X writes to `daily/calendar/` which is db_only: files there will never be ingested by sync or import." Recipes could declare their output dir in frontmatter so the check has ground truth.

**Alternatives considered**

Local fix (deployed): the calendar handler imports day files via `importFromFile` (hash-skip idempotent) on every run, bypassing the git walk. Works, and stays useful — but the silent-death class needs the native warning.

---

Related: #2784 (declared-db_only auditing — same storage-tier blind-spot family).

Environment: gbrain 0.42.57.0 (verified unchanged at 0.42.58.0 / a25209bb) · macOS 26.3.1 · Bun 1.3.13 · Supabase Postgres (PgBouncer transaction mode).

time-attack:
Maintainer-approved silent-failure check batch ("approve after thorough testing"). Four warn-level doctor / integrations-doctor checks, each backed by a production receipt in its issue. The common thread: every one of these failure modes reported green at every existing surface while ingestion or data integrity silently degraded.

### 1. `content_hash_duplicates` (#2250 — reported by @615Works)
`gbrain import` run from a root one level too deep drops the path prefix from every slug, leaving `people/alice-example` and `alice-example` coexisting with identical content. Purge never removes them (they aren't file-backed orphans) and doctor never surfaced them. The check is a **single GROUP BY** over `(source_id, content_hash)` with FILTER aggregates — no N² hash comparison — flagging only hash groups that hold BOTH a bare and a path-prefixed slug (the wrong-root signature). Warn carries sample pairs plus the remediation the issue asked for: `gbrain pages delete <bare-slug>` … then `gbrain pages purge-deleted --older-than 0`. Grouping is per-source; soft-deleted and hash-less pages excluded.

### 2. `undeclared_db_only_pages` (#2784 — reported by @alexputici)
For each source with a local repo on this host: markdown pages with no backing file that sit outside every declared db_only path → warn with count + sample slugs + the gbrain.yml fix. The issue's derive-phase prefixes ship as defaults (`DERIVE_PHASE_DB_ONLY_DEFAULTS`: `life/events/`, `atoms/`, `extracts/`, `dream-cycle-summaries/`) treated as implicitly declared, so the check stays quiet on healthy brains.

**One deliberate deviation from the issue text:** the defaults are NOT merged into `loadStorageConfig` itself. A global merge would flow into `manageGitignore` and auto-gitignore those dirs — which is exactly the #2788 silent-skip trap for any brain that file-backs them. The defaults live at the check layer; explicitly declared dirs alone drive `.gitignore`.

This is the one check deliberately allowed to stat the repo (the registry's SQL-only stance is what made this class invisible).

### 3. `heartbeat_max_age` (#2787 — reported by @alexputici)
New declarable health-check type for recipe `health_checks`. The recipe states its cadence (`max_age: 48h`); `gbrain integrations doctor` reads the heartbeat file and **fails** past the threshold — including the "no heartbeat data at all" case. All previously shipped types are point-in-time and pass a sense that has been dead for weeks (the 16-day receipt in the issue). Not embedded-gated: it reads only the local heartbeat file (no exec, no network), so user recipes can declare it too. `calendar-to-brain` now declares `max_age: 48h`. Works inside `any_of`.

### 4. `db_only_collector_collision` (#2788 — reported by @alexputici)
Recipes now declare their collector output dirs in frontmatter (`output_paths`, ground truth per the issue's suggestion; calendar-to-brain declares `daily/calendar/`). The doctor check warns when a **configured** collector's output path falls inside a declared db_only prefix — auto-gitignore means the git-walking sync AND `gbrain import` silently skip every file it writes (the 7-week outage receipt). The same warning also fires at config-write time inside sync's `manageGitignore`, right where the `.gitignore` entry is managed; it never blocks the gitignore housekeeping.

All three `gbrain doctor` checks are registered in `doctor-categories.ts` (`content_hash_duplicates`/`undeclared_db_only_pages` → brain, `db_only_collector_collision` → ops), stream progress heartbeats, and degrade to warn-with-reason on any internal error.

## Testing (the maintainer's condition)

Every check has positive AND negative units, plus an e2e that constructs the real failure condition:

| Suite | Result |
|---|---|
| `test/doctor-silent-death-checks.test.ts` (PGLite units, pos+neg per check incl. cross-source negative, soft-delete negative, sibling-prefix negative, derive-default positives) | 21 pass / 0 fail |
| `test/integrations-heartbeat-max-age.test.ts` (parse table, fresh / 16-day-stale / no-data / invalid / any_of / untrusted-recipe, shipped-recipe frontmatter pin) | 12 pass / 0 fail |
| `test/storage-sync.test.ts` (+2 manageGitignore collision-warn cases: warn text + non-blocking) | 16 pass / 0 fail |
| `test/e2e/doctor-silent-death-parity.test.ts` — **real wrong-root import** (same files imported through `importFromFile` from the correct root and from one level too deep) fires the check with the exact remediation text; **engine parity** seeds identical state on PGLite + real Postgres (pgvector/pg16 via docker-compose.ci.yml) and asserts byte-identical verdicts for all three checks, plus a clean-engine negative on both | 4 pass / 0 fail (Postgres half executed with a live DATABASE_URL, not skipped) |
| `bun run typecheck` | clean |
| Guards: check-jsonb-pattern, check-jsonb-params, check-progress-to-stdout, check-test-isolation, check-key-files-current-state, `bun test test/build-llms.test.ts` | all pass |
| `test/doctor-categories.test.ts` drift guard + `test/doctor.test.ts` + `test/doctor-behavioral.test.ts` + `test/integrations.test.ts` | pass |
| Full unit suite (`bun run test`) + serial suite (`bun run test:serial`) | in flight at PR-open time (started clean; targeted suites above all green). Final numbers will be posted as a PR comment — note master carries a small set of known-failing unit files unrelated to this branch. |
| `bun build --compile` binary smoke | builds, `--version` runs |

Docs: `docs/architecture/KEY_FILES.md` (current-state entries for doctor / integrations / storage-config), `docs/storage-tiering.md`, `docs/integrations/README.md`; llms bundles regenerated (`bun run build:llms`).

Closes #2250, closes #2784, closes #2787, closes #2788


time-attack:
CI triage — `test (8)` failure is inherited from master, not this branch:

- Failing test: `test/ai/recipes-existing-regression.test.ts` › "all 3 touchpoints produce identical auth for the same recipe + env" — throws `AIConfigError: Azure OpenAI (Entra/keyless): could not get an access token via 'az account get-access-token'`.
- Cause: master commit 2c758e23 (#2354) added an Entra/keyless fallback to `azure-openai.resolveAuth` that shells out to `az` when no API key is in the auth env; the regression test enumerates every openai-compatible recipe and hits that branch on runners without `az` login.
- Proof it's branch-neutral: master's own Test run on b30f0aa7 (run 30315251140, zero commits from this branch) fails the identical test in the identical shard. This branch's shard 8 is otherwise green: 1347 pass / 1 fail, and every test added by this PR passes in CI (doctor-silent-death-checks, integrations-heartbeat-max-age, storage-sync).

time-attack:
Final test receipts (maintainer condition: "approve after thorough testing"):

**GitHub CI (authoritative)** — run 30315453628: 18 pass / 1 fail / 1 pending at comment time — the pending shard, test (4), contains none of this PR's test files (this PR's suites ran in shards 5, 7, and 10, all green). The only red is the master-inherited test (8).

- All tests added by this PR pass in CI: `doctor-silent-death-checks` (21), `integrations-heartbeat-max-age` (12), `storage-sync` (+2 collision-warn cases), `doctor-silent-death-parity` e2e.
- The single red check, `test (8)`, is the master-inherited azure-openai Entra regression (#2354 / 2c758e23) — proof in the comment above; shard is otherwise 1347 pass / 1 fail.

**Local (this branch, 10-core arm64)**
- `test/e2e/doctor-silent-death-parity.test.ts` against **real Postgres** (pgvector/pg16, docker-compose.ci.yml): 4 pass / 0 fail — byte-identical check verdicts on PGLite vs Postgres, plus the real wrong-root `importFromFile` reproduction of #2250 firing `content_hash_duplicates` with the exact remediation text.
- `bun run typecheck` clean; `bun run verify` 31/31 green; jsonb/progress/test-isolation/key-files/build-llms guards all pass.
- Local serial suite: 3 failing files (`facts-mcp-allowlist`, `hybrid-meta`, `hybrid-search-lite`) — all environment-only (API keys present in the local shell); the same 22 tests pass with keys stripped, and CI `serial-tests` is green.

time-attack:
CI final: test (4) landed green — 19/20 checks pass. The only red remains the master-inherited azure-openai Entra regression in test (8) (#2354 / 2c758e23; proof upthread).

time-attack:
Thanks for the detailed writeup and the production receipts behind each check — the wrong-root import signature and the collector/gitignore collision are both real diagnoses. I'm closing this as feature surface rather than a fix: all four checks are detection-only, so the underlying failure modes survive the merge unchanged, and this adds ~1130 lines across four unrelated issues in a single PR. If one of these is genuinely load-bearing, the version worth taking is the root-cause fix for that one failure mode on its own — e.g. stopping `manageGitignore` from ignoring a configured collector's output path in the first place, rather than warning about it afterward.

time-attack:
Reopening — closed in error. This PR was built/approved as part of the maintainer's explicit triage decisions (the doctor silent-failure batch, skills-integrity manifest, npm-shadow guard, DCR token TTL, and the retype slug_filter). The bulk strict-policy pass that swept it did not have that context. Merge queue for these resumes once CI's install step recovers.

time-attack:

The aim is right — all four checks target real, receipted silent-failure classes, the tests are genuinely discriminating (stubbing the check breaks them), typecheck is clean, and both a fresh and a populated brain show zero false positives. But four concrete defects block it: (1) the PR is CONFLICTING with master (src/commands/doctor.ts — master's #160 unverified_extractions check landed in the same region); (2) the content_hash_duplicates remediation string is broken twice and the tests PIN the broken text — `gbrain pages delete <bare-slug>` hard-fails with "Unknown command: pages" (real command is `gbrain delete <slug>`), and `purge-deleted --older-than 0` is silently ignored by parseOpArgs (real flag is `--older-than-hours`; proved live: wrong flag purges 0, right flag purges 1) — the printed fix exits green while doing nothing, the exact silent-failure class this PR exists to kill; (3) the flagship check misses the canonical #2250 scenario for plain markdown: a real wrong-root import of `people/alice-example.md` with no frontmatter created both twin slugs in my live brain yet the check reported OK, because import's content_hash includes the path-inferred `type` ('person' vs 'note', import-file.ts:572-580) — the e2e admits the dodge at test/e2e/doctor-silent-death-parity.test.ts:56-58 by requiring explicit `type:` frontmatter; (4) title lacks the version-first prefix and branch VERSION 0.42.66.1 is below master's 0.42.68.1 with no CHANGELOG entry, so the CI version-gate rejects as-is. Fixes needed: rebase onto master, correct the remediation string in src/commands/doctor.ts:3642 to `gbrain delete <slug>` + `gbrain purge-deleted --older-than-hours 0` (and update the three test assertions pinning the old text), either widen the duplicate detection to a type-insensitive comparison or document/warn that untyped-markdown wrong-root imports are not caught, and run /ship for the version/title mechanics.

### Discrimination test

Stubbed checkContentHashDuplicates to always return status 'ok' and re-ran test/doctor-silent-death-checks.test.ts: 2 tests failed (19 pass), then restored the file. The tests bind to real behavior — they run the actual exported check functions against a real PGLite engine, not substring greps.

**Mergeability at review time:** {"mergeStateStatus":"DIRTY","mergeable":"CONFLICTING"} — confirmed by hand: `git merge origin/master --no-commit` conflicts in src/commands/doctor.ts (conflict markers at lines 3632/3821/3864), aborted after inspection

### Blockers

**1.** CONFLICTING with master: trial merge conflicts in src/commands/doctor.ts (~lines 3632-3864) against master's #160 check; gh reports mergeable=CONFLICTING, mergeStateStatus=DIRTY

**2.** Remediation string broken and test-pinned (src/commands/doctor.ts:3642): 'gbrain pages delete' -> 'Unknown command: pages' (exit 1); 'gbrain pages purge-deleted --older-than 0' -> no 'pages' group, and '--older-than' is silently dropped by parseOpArgs (src/cli.ts:792 maps it to older_than; handler reads older_than_hours, default 72) so it exits 0 and purges nothing — proved live (wrong flag: count 0; '--older-than-hours 0': count 1). test/doctor-silent-death-checks.test.ts:87-88 and test/e2e/doctor-silent-death-parity.test.ts:97-98 assert the broken text

**3.** content_hash_duplicates false-negatives on the canonical #2250 case: live wrong-root import of a plain '# Alice' markdown file created both 'alice-example' and 'people/alice-example' in the DB but the check returned OK, because content_hash includes path-inferred type (import-file.ts:572-580; person vs note). Only files with explicit 'type:' frontmatter are caught — the e2e comment (doctor-silent-death-parity.test.ts:56-58) acknowledges this constraint but the PR body and check message do not

**4.** Version/process gate: PR title does not lead with vMAJOR.MINOR.PATCH.MICRO (IRON RULE); branch VERSION 0.42.66.1 < master 0.42.68.1; no CHANGELOG entry for this change — CI version-gate fails as-is

### Not verified

- Postgres half of test/e2e/doctor-silent-death-parity.test.ts (requires DATABASE_URL + docker; skipped for execution footprint — PGLite half of the same assertions ran and passed)
- Full unit suite / bun run verify (footprint rule); targeted suites all green: doctor-silent-death-checks + integrations-heartbeat-max-age + storage-sync (49 pass), doctor-categories + integrations (121 pass), typecheck clean
- PR-body claim that the full suite was 'in flight at PR-open time' — no completion comment checked
- manageGitignore recipe-scan cost on every sync (loads all recipes + reads heartbeat files; readHeartbeat can rewrite the heartbeat file as a pre-existing prune behavior) — noted, not measured

- `gh pr view 3457 --json ... (twice; mergeable UNKNOWN then CONFLICTING/DIRTY)`
- `git fetch origin pull/3457/head:pr-3457 && git checkout pr-3457`
- `git merge origin/master --no-commit --no-ff (conflict in src/commands/doctor.ts; aborted)`
- `bun install`
- `bun test test/doctor-silent-death-checks.test.ts test/integrations-heartbeat-max-age.test.ts test/storage-sync.test.ts (49 pass / 0 fail)`
- `bun run typecheck (clean, exit 0)`
- `GBRAIN_HOME=/tmp/pr3457-brain gbrain init --embedding-model openai:text-embedding-3-small`
- `gbrain doctor on fresh brain (all 3 new checks OK)`
- `gbrain import from correct root, then from one level too deep (both pages created); gbrain doctor -> content_hash_duplicates OK (false negative proved)`
- `gbrain pages delete alice-example (exit 1: Unknown command: pages); gbrain delete alice-example (soft-deleted)`
- `gbrain purge-deleted --older-than 0 (exit 0, count 0) vs gbrain purge-deleted --older-than-hours 0 (count 1)`
- `gbrain integrations doctor (configured-only gating confirmed live)`
- `discrimination test: stubbed checkContentHashDuplicates to 'ok', re-ran unit file (2 fail), restored`
- `bun test test/doctor-categories.test.ts test/integrations.test.ts (121 pass)`
- `cleanup: rm -rf /tmp/pr3457-brain /tmp/pr3457-repo`

</details>

---

time-attack:

PR #3458, opened 2026-07-27
feat(extract): quarantine lane for auto-extracted entities from untrusted input (#160)

### LINKED ISSUE #160: extractAndEnrich writes pages from untrusted text without a gate

time-attack:
## What

Implements the quarantine-lane decision on #160 for the `extractAndEnrich` entity-extraction write path. Extracted-entity stubs are only written as authoritative `people/` / `companies/` pages when the caller is the trusted local CLI **and** passes an explicit opt-in flag; every other caller's stubs land in a review queue and rank as ordinary content until the owner promotes them.

## Mechanism

- **Markers, not a new table.** `src/core/extraction-review.ts` (sibling of `quarantine.ts` / `embed-skip.ts`, frontmatter-JSONB-key pattern — no schema migration). Stubs from untrusted input carry `provenance: auto-extracted` + `status: unverified`; the pair is required, so user pages with their own `status`/`provenance` frontmatter never match. `unverifiedExtractionFragment()` is the single SQL source of truth for every consumer (search guard, engine method, list op, doctor check).
- **Fail-closed trust.** `enrichEntity` / `enrichEntities` / `extractAndEnrich` take `EnrichmentTrustOptions`; only an explicit `trusted: true` writes direct. The op layer grants that solely for `ctx.remote === false` + `--trusted-extraction` — mirroring the `OperationContext.remote` invariant (anything not strictly `false` is untrusted, including unset).
- **Retrieval: no entity authority for unverified stubs.** They skip the compiled-truth fusion boost (`stampUnverifiedExtractions` runs pre-fusion on all three hybrid paths and the keyword-only opt-out; both `rrfFusion` variants honor the flag) and the `people/`/`companies/` namespace source-boost (guard inside `buildSourceFactorCase`, shared by both engines' search SQL). Search results from such pages carry `unverified: true` so agents can label the provenance. New engine method `getUnverifiedExtractionPageIds` in BOTH engines (parity via the shared fragment).
- **Review queue (contract-first ops + generated CLI).**
  - `extract_entities` — the sanctioned surface for extraction (`gbrain extract-entities --text ... --source-slug ... [--trusted-extraction]`).
  - `extraction_pending` — read-scoped, source-scoped list of stubs awaiting review.
  - `extraction_review` — batch `promote` (status → `verified`; `provenance` kept as audit trail) / `reject` (soft-delete, recoverable). Owner-only: `localOnly` + a handler-level `ctx.remote === false` gate, so injected content cannot self-promote through the review path (note: a remote caller already holding generic put_page write scope can rewrite the plain frontmatter markers — that adds no privilege over authoring an unmarked page outright, and is documented as known scope in extraction-review.ts).
- **Doctor nudge.** `unverified_extractions` warns when stubs older than 7 days await review, with the exact review commands.

## Tests

- `test/extraction-review.test.ts` (PGLite, hermetic): fail-closed trust matrix (remote `true` / `false` / **unset**), marker predicates, fusion-boost skip, source-boost guard, review queue (batch promote/reject, per-slug statuses, owner-only refusal), doctor check, and an end-to-end hostile-transcript case proving fabricated entities land in the review queue and rank below an owner-authored page of equal lexical relevance, with `unverified: true` in result metadata.
- `test/e2e/extraction-review-postgres.test.ts` (live Postgres, DATABASE_URL-gated): same SQL-touching pieces verified against pgvector/pg16 — markers, engine method, source-boost ratio in `searchKeyword`, pending/promote/reject ops.
- `test/sql-ranking.test.ts` expectations updated to current state.
- CLI smoke on a scratch brain: extract → pending → promote → doctor, plus the `--trusted-extraction` direct path.

## Notes

- No migration needed (frontmatter-key pattern; current `MIGRATIONS` max untouched).
- No VERSION bump in this PR — leaving the version allocation to the ship flow per repo convention; happy to fold one in if preferred.
- The vector arm's re-rank CTE projects the guard predicate as a boolean column (`unverified_stub` in `hnsw_candidates`, both engines), so the namespace factor is neutralized there too — pinned by a factor-1.0 test on both engines. (An earlier revision wrongly claimed the fusion guard covered this arm; adversarial review caught it and the SQL-level guard now closes it.)

Closes #160


time-attack:
Review-round fixes pushed (d43ed5d8), addressing the adversarial verification findings:

- **Blocker closed — vector-arm source boost.** `hnsw_candidates` in BOTH engines now projects the unverified predicate as `unverified_stub`; `buildSourceFactorCase` takes the guard column for the bare-slug re-rank form, so unverified stubs get factor 1.0 inside the pre-LIMIT `scored` CTE (no rank advantage, no candidate-pool eviction of legitimate pages). Pinned by identical-basis-embedding tests on PGLite and live Postgres (score ratio = 1.2 for verified vs unverified). The prior comment/doc claim that fusion covered this arm was wrong and is corrected.
- **extract_entities hardening.** 200k-char input cap (loud `invalid_params`), 200-entity cap surfaced as `truncated` + `entities_found` in the result; same default cap in the library `extractAndEnrich`. Verified `OperationContext` carries no abort-signal field, so the caps are the bound; flood test (300 name-shaped tokens → 200 processed) included.
- **Promote no longer round-trips putPage.** Targeted JSONB-merge `UPDATE` flips `status` only — `page_kind` / `content_hash` can't be reset by the upsert. `extraction_pending` now applies `buildVisibilityClause` (archived sources excluded).
- **Wording (chose: document, not namespace).** The markers stay plain frontmatter keys — a remote write-scope caller who could strip them can author an unmarked `people/` page via put_page anyway, so namespacing + strip-listing adds no privilege. The op description and module header now state that assumption plainly, and the CREATE-only scope (the UPDATE-path timeline/edge write is the separately-tracked slug-collision finding) is documented in the module header.

Receipts: extraction-review suite 22/22 (PGLite), live-PG e2e 4/4 (pgvector/pg16), full `test/search/` 340/340 incl. searchvector-maxpool + title-retrieval-arm + rrf-source-key, sql-ranking, doctor 89/89, ops/cli/thin-client suites green, typecheck clean, JSONB + key-files guards clean.

Issue #1985, opened 2026-06-08
`gbrain upgrade` (bun-link) prints "Auto-upgrade failed" after a successful schema migration — likely child-subprocess pool exhaustion in the v0.12.0 orchestrator on a Supabase session pool


stevotcp:
### Summary

On a `bun-link` install upgrading a long-lived brain (schema 54 → 115), `gbrain upgrade` ran the full schema migration successfully, then aborted the post-upgrade orchestrator chain at the **v0.12.0 Knowledge-Graph orchestrator** and printed:

```
=== v0.12.0 — Knowledge Graph auto-wire ===

Config key not found: auto_link
Auto-upgrade failed. Try manually:
  cd <repo> && git pull && bun install
```

…even though the schema fully migrated to 115 and `gbrain doctor` reports a healthy brain afterward. The remediation it suggests (`git pull && bun install`) is the exact thing that just ran, so it reads as a confusing no-op loop.

### Environment

- Upgrade path: **bun-link** source clone (global `gbrain` is `bun link`ed to the repo).
- From **v0.33.0** (schema 54) → **v0.42.36.0** (schema 115). 60 migrations.
- Engine: **Postgres on a Supabase session pooler**, `pool_size: 15` (port 5432). `GBRAIN_DISABLE_DIRECT_POOL` was set, so all traffic — including child subprocesses — went through the session pool.
- Brain: ~782 markdown pages, ~1023 chunks. No code pages.

### What I verified in the source (so these are NOT the cause)

- `phaseBConfigCheck` (`src/commands/migrations/v0_12_0.ts:71-76`) **catches** the non-zero exit of `gbrain config get auto_link` and correctly defaults to enabled. So the `Config key not found: auto_link` line is **harmless child stderr**, not the failure — but it's alarming to users and should be suppressed or routed.
- `apply-migrations` (`src/commands/apply-migrations.ts` ~438) only hard-fails (`failed = true; break`) when an orchestrator returns `status === 'failed'`. `orchestrator()` only returns `'failed'` via the `phaseASchema` early-return; every other phase failure yields `'partial'`, which does NOT exit non-zero.
- `gbrain extract links --source db` (phase C) still works in v0.42 (verified via `--dry-run`), so stale CLI syntax is not the cause.
[...]

PR #3463, opened 2026-07-28
fix(repo): untrack node_modules symlink, guard against tracked symlinks

### LINKED ISSUE #3462: Committed node_modules symlink → /tmp/fleet/repo/node_modules breaks bun install for every consumer (regression in faf5cdba / #2357)
`faf5cdba` (PR #2357) added a **git-tracked symlink** at the repo root:

```
$ git ls-tree origin/master node_modules
120000 blob 96daf4519285f3d0d70c3532889d42ddc14cd088	node_modules
$ git cat-file -p 96daf4519285f3d0d70c3532889d42ddc14cd088
/tmp/fleet/repo/node_modules
```

That's a build-sandbox path. On any other machine it's a dangling symlink, so `bun install` refuses to run:

```
bun install v1.3.13
ENOENT: could not open the "node_modules" directory
```

Repro:

```bash
git clone https://github.com/garrytan/gbrain && cd gbrain && bun install
```

Hit via `gbrain upgrade` (0.35.0.0 → 0.42.66.1), which does `git pull` then `bun install` — the pull succeeds, the install dies, leaving the CLI on new source with old/absent deps. Workaround is `rm node_modules && bun install`, but the tracked symlink comes back on every pull.

Two side notes:

- `gbrain upgrade`'s failure hint prints `cd ~/gbrain && git pull && bun install`, i.e. a hardcoded `$HOME/gbrain`. That path doesn't exist for a bun-global install, where the clone lives at `~/.bun/install/global/node_modules/gbrain`. Worth deriving the hint from the actual install dir.
- Minor hardening: a tracked symlink into world-writable `/tmp` means that if `/tmp/fleet/repo/node_modules` happens to exist (or is creatable by another user on a shared box), module resolution silently follows it. Not calling it a vulnerability given it's a build artifact, but it's an argument for a `.gitignore` entry plus a CI guard against committed symlinks.

Fix: `git rm --cached node_modules`, add to `.gitignore`.

eoin-obrien:
Fixes #3462.

## Why `.gitignore` didn't already stop this

Worth calling out, because the issue's suggested fix ("add to `.gitignore`") is a no-op on its own: **`.gitignore:1` already was `node_modules/`.**

A pattern with a trailing slash matches **directories only**. A symlink named `node_modules` is not a directory to git, so it was never ignored and `git add -A` swept it in. Line 18's `admin/node_modules/` has the same hole.

So the fix is three parts, not two.

## Changes

**1. Untrack the symlink** — `git rm --cached node_modules`.

**2. Drop the trailing slash** from both `node_modules` patterns so they match any entry type, not just directories.

**3. Add `scripts/check-no-tracked-symlinks.sh`**, wired into `bun run verify` and `check:all`.

The `.gitignore` change alone is not a backstop — `git add -f` bypasses `.gitignore` entirely, which is one plausible way the original symlink got staged. The guard fails on any tracked mode-`120000` entry and prints the offending path plus its target. The repo has zero legitimate tracked symlinks today, so it starts with an empty allowlist and no grandfathering.

This also covers the hardening point raised in the issue: a tracked symlink into world-writable `/tmp` means module resolution silently follows it if that path becomes creatable on a shared box. A guard that refuses tracked symlinks outright removes the class rather than this one instance.

## Test coverage

`test/no-tracked-symlinks-guard.test.ts` (5 tests) — builds a throwaway git repo containing the exact symlink shape from `faf5cdba`, asserts the guard exits 1 and names the offender, asserts it stays green on this repo, asserts it's present in the verify dispatcher's `CHECKS[]`, and asserts no `node_modules` pattern in `.gitignore` regains a trailing slash.

Testing the failure path matters more than the happy path here: a guard that silently never fires looks identical to a clean repo.

## Verification

- `bun run verify` → 32/32 green (was 31 checks; +1 new)
- `bun run check:all` → clean
- `bun test test/scripts/run-verify-parallel.test.ts test/privacy-script-wired.test.ts` → 13 pass (the two tests that pin the check list)
- `bun test test/no-tracked-symlinks-guard.test.ts` → 5 pass
- Confirmed `bun install` succeeds on a tree with the symlink removed

No `VERSION` / `CHANGELOG.md` changes — leaving those to the maintainer per `docs/RELEASING.md`. The `package.json` diff is scripts-only (no dependency change), though it will trigger the OSV-Scanner workflow.

## Follow-ups, deliberately not in this PR

The issue raises the upgrade hint printing `cd ~/gbrain && git pull && bun install`. Reading `src/commands/upgrade.ts:39`, that line interpolates a resolved path rather than a literal:

```ts
console.error(`  cd ${linkInfo.repoRoot} && git pull && bun install`);
```

`repoRoot` comes from `detectBunLink()` (`upgrade.ts:762`), which walks up from `process.argv[1]` for a `.git/config` referencing `garrytan/gbrain` — so on my machine the `~/gbrain` in that message was my actual clone path. Possible the reporter's install has a shape I'm not seeing; flagging what I read in case it saves someone a repro, not to close the question.

Either way there's a real diagnosability problem nearby: the `catch` at `upgrade.ts:37` discards the underlying error and prints a generic "Auto-upgrade failed", so the user gets pointed at `bun install` — the exact command that just failed. That's most of why this needed digging to diagnose at all. #1985 reports the same generic message from an unrelated root cause, so that looks like the better home for a fix than this PR.


dialthewolff:
Independent local verification on macOS (Bun 1.3.14) passed:

- `bun install`
- `bun test test/no-tracked-symlinks-guard.test.ts` — 5/5
- `bun test test/scripts/run-verify-parallel.test.ts test/privacy-script-wired.test.ts` — 13/13
- `bun run verify` — 32/32

The current `action_required` runs are blocked on maintainer approval for this external-contributor PR, not a reproduced test failure.

dialthewolff:
@garrytan Maintainer action needed to unblock #3462: please approve/run Actions on this external-contributor PR. I independently verified `bun install`, the targeted guards, and `bun run verify` locally; all passed. Once hosted checks are green, this is ready to merge.

Issue #94, opened 2026-04-13
Plans for a Claude Code-native mode (no API keys, like gstack)?


jiongxuan:
## TL;DR

Are there any plans to support a **Claude Code-native** operating mode for GBrain — where the brain runs entirely on top of Claude Code's built-in model access (skills, slash commands, sub-agents) instead of requiring users to bring their own `OPENAI_API_KEY` / `ANTHROPIC_API_KEY`?

The model here is [`gstack`](https://github.com/anthropics/claude-code) style projects: pure markdown skills + slash commands, zero API keys, zero billing setup, the user just installs Claude Code and it works.

## Why

Today's install flow (from README) requires:

```
export OPENAI_API_KEY=sk-...          # required for vector search
export ANTHROPIC_API_KEY=sk-ant-...   # optional
```

For people who already pay for Claude Code (Pro / Max / Team), bringing a second set of API keys is real friction:

- Two billing relationships instead of one
- Have to provision and rotate keys
- Onboarding stalls at "go get an OpenAI key" — the opposite of "~30 minutes to a fully working brain"
- Many Claude Code users don't have an OpenAI account at all

GBrain's own ethos docs (`docs/ethos/THIN_HARNESS_FAT_SKILLS.md`, `MARKDOWN_SKILLS_AS_RECIPES.md`) already argue that markdown skills *are* the product. A Claude Code-native mode would be the logical end state of that philosophy.

## What it might look like

- **Embeddings**: optional. Default to keyword/trigram search via PGLite (already supported as a fallback). Skip the vector path entirely unless the user opts in. Or: use a local embedding model bundled via WASM.
- **Query expansion / reranking / dream cycle**: done by Claude Code itself via skills + sub-agents, not by API calls from the CLI process.
- **Ingestion summarization, entity extraction, back-linking**: same — invoked as Claude Code skills, not via `ANTHROPIC_API_KEY`.
- **`gbrain init` mode flag**: e.g. `gbrain init --mode claude-code` that picks the keyless defaults and wires the skills accordingly.

Net effect: a user with only Claude Code installed can clone the repo, run `gbrain init`, and have a working brain with no third-party keys.

## Questions

1. Is this on the roadmap, or has it been considered and rejected? If rejected, what was the blocker (search quality without embeddings? something else)?
2. Would a community PR exploring a `--mode claude-code` path be welcome?
3. Are there constraints from the contract-first architecture (`src/core/operations.ts`) that would make this awkward?

[...]

PR #3464, opened 2026-07-28
feat(init): --mode claude-code — keyless mode for Claude Code subscri…

mnemonik-dev:
…bers (#94)

A Claude Code subscriber gets a working brain with zero provider API keys: PGLite engine, keyword + graph + title search (hybrid search's existing no-embedding-provider path), chat via the claude-cli recipe's OAuth-session subprocess.

- init: new --mode claude-code flag. Implies --pglite, writes claude_code_mode: true alongside embedding_disabled: true, defaults chat_model to claude-cli:claude-sonnet-4-6, seeds search.mode=conservative (the one bundle with no reranker / LLM expansion spend), and best-effort-checks the claude binary on PATH.
- import: a claude-code brain imports without vectors (implicit --no-embed) instead of refusing with the D9 deferred-setup error.
- embed: refusal message tailored to keyless mode with the vector upgrade recipe instead of the deferred-setup hint.
- advisor: embedding_disabled is not a setup smell when the brain is keyless by design.
- fix: explicit --embedding-model/--model now clears a persisted deferred/keyless sentinel — the documented upgrade path (gbrain init --force --embedding-model ...) previously lost to the sentinel and was silently ignored; a resolved model also removes stale sentinels from config.json.
- fix: the post-init subagent-Anthropic caveat no longer fires for claude-cli:* chat models (the recipe drives the subagent loop through the OAuth session; no ANTHROPIC_API_KEY involved).

Verified end-to-end with all provider keys unset: init -> import ->
search (keyword arm) -> embed refusal -> re-init upgrade path.

Claude-Session: https://claude.ai/code/session_01WQ7za53ZWxWqL2XAfSgTux

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,17 @@
 
 All notable changes to GBrain will be documented in this file.
 
+## [Unreleased]
+
+### Added
+
+- `gbrain init --mode claude-code` (#94): Claude Code-native keyless mode. No provider API keys required — PGLite engine, keyword + graph + title search via hybrid search's existing no-embedding-provider path, chat through the `claude-cli` recipe's OAuth session. Writes a `claude_code_mode: true` sentinel alongside `embedding_disabled: true`; `gbrain import` on such a brain proceeds without vectors (implicit `--no-embed`) instead of refusing, `search.mode` defaults to `conservative` (the one bundle with no reranker or LLM-expansion spend), and doctor/advisor treat the missing embedding provider as intentional rather than an unfinished setup.
+
+### Fixed
+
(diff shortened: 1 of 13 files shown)


time-attack:
Closing under our no-new-feature-surface policy: this is the '--mode claude-code' proposal from issue #94, which was also closed for the same reason. A whole keyless operating mode layered over another tool's model access is more surface than we're willing to own. Partial path already exists today: PGLite is the zero-config default, keyword/trigram search works with no embeddings, and the embedding provider is pluggable to local endpoints. Thanks for the effort — 486 lines is real work, and the write-up was clear.

PR #3469, opened 2026-07-28
fix: reserve autopilot worker capacity

### LINKED ISSUE #2050: autopilot: drain-worker runs at concurrency=1, self-deadlocking any cycle phase that spawns a subagent (patterns, synthesize)
**Engine:** Postgres (self-hosted Supabase)
**Version:** 0.42.38.0

## Summary

`gbrain autopilot` spawns its drain-worker as a bare `gbrain jobs work` (no `--concurrency`), which defaults to **concurrency 1** (`src/commands/jobs.ts:82` → `?? '1'`; `src/core/minions/worker.ts:200` → `?? 1`). The `patterns` and `synthesize` cycle phases submit a **child `subagent` job to the same queue and then block on it** via `waitForCompletion` (`src/core/cycle/patterns.ts:88,94` — 35-min timeout; `src/core/cycle/synthesize.ts:484`).

At concurrency 1 the parent `autopilot-cycle` handler occupies the only in-flight slot — `worker.ts:514` claims a new job only while `inFlight.size < concurrency`, and the parent stays in `inFlight` for the whole `await waitForCompletion(...)`. So the child subagent can never be claimed by the same (only) worker → **self-deadlock**. The cycle then trips the 10-min per-job timeout; because the timed-out handler isn't actually cancelled (cf. #212), it becomes a zombie that keeps running. Repeated every interval, zombies accumulate and the queue wedges (hundreds of `waiting` jobs, 0 completions, dead-holder lock churn).

`cycle.ts` actually documents the assumption ("synthesize/patterns renew the cycle-lock TTL during long waits") — i.e. the design expects a *separate* worker slot to process the subagent while the parent waits. Autopilot's bare concurrency-1 worker doesn't provide one.

## Why this is under-reported

- The **documented** worker is `gbrain jobs supervisor --concurrency 4` (`docs/guides/minions-deployment.md`), which has spare slots → never deadlocks. Autopilot's built-in concurrency-1 worker is the outlier.
- `patterns` only fires with ≥ `dream.patterns.min_evidence` (default 3) reflection pages; `synthesize` only with new transcripts. Many brains skip both, so the subagent is never submitted and the deadlock never triggers.
- The population that hits it = {runs the `autopilot` daemon, not the supervisor} ∩ {has reflections / pending transcripts}.

## Reproduction

1. Postgres engine, `gbrain autopilot` running (default path → bare concurrency-1 worker).
2. Ensure a subagent-spawning phase actually runs (≥3 reflection pages for `patterns`, or pending transcripts for `synthesize`).
3. A cycle reaches `[cycle.patterns] start` and never logs `done`; the submitted `subagent` job stays `waiting`; the cycle force-evicts at the 10-min timeout; zombie handlers accumulate; the queue wedges.

Isolation that confirms the cause:
- `gbrain dream` (runs the cycle inline, **no** separate worker) hangs at `[cycle.patterns]` until killed — the queued subagent has nothing to run it.
- Running `gbrain jobs work --concurrency 4` alongside, the *same* cycle reaches `[cycle.patterns] done` and the full cycle completes.

## Workaround

`export GBRAIN_WORKER_CONCURRENCY=4` in the environment that `autopilot-run.sh` sources — the autopilot-spawned worker inherits it (`autopilot.ts:398` passes `...process.env`). Verified: a patterns cycle completes instead of hanging. (Or disable the phases: `gbrain config set dream.patterns.enabled false`, likewise `dream.synthesize.enabled false`.)

## Suggested fix (any one)

- `gbrain autopilot` launches its worker at `--concurrency ≥2` by default — it submits subagent-spawning cycles, so a concurrency-1 worker can never satisfy them.
- Run the `patterns`/`synthesize` subagent **inline** within the phase (this also fixes the PGLite case in #1306, where a 2nd worker process can't exist at all).
- Don't count a parent that's blocked on a child against the in-flight cap (release/yield the slot during `waitForCompletion`).
- A dedicated queue + worker for child subagents.

Independently: the per-job timeout should actually cancel the handler (#212) so a deadlocked cycle dies cleanly rather than becoming a zombie.

## Related

- #1306 — PGLite: `dream --phase synthesize` hangs indefinitely (no worker daemon to process the queued subagent). Same shape; file-lock variant.
- #1753 — Supabase synthesize writes 0 pages; children fail/timeout.
- #631 — clarify autopilot vs `jobs supervisor` production deployment shape.
- #1737 — long-running subagent/cycle jobs starve the queue.
- #212 — cycle handler ignores AbortSignal, leaving zombie work after per-job timeout (the amplifier here).
- #1934 — subagent `AI_MissingToolResultsError` on gateway-loop replay. Surfaces *after* the deadlock is removed: with a free slot the subagent runs but currently fails this way, so patterns/synthesize complete with 0 pages until that lands.

dialthewolff:
Closes #2050.

## Summary
- size the managed worker for every per-source parent, global maintenance, and one child
- clamp per-tick fanout if config rises after startup
- apply the same safe capacity to externally managed workers and doctor guidance

## Verification
- `bun test test/autopilot-fanout.test.ts test/autopilot-supervisor-wiring.test.ts test/autopilot-fanout-clamp.serial.test.ts test/doctor-autopilot-fanout-concurrency.serial.test.ts` (44 pass)
- `bun run verify` (31/31)
- independent Codex blocker review: PASS

### DIFF
--- a/src/commands/autopilot-fanout.ts
+++ b/src/commands/autopilot-fanout.ts
@@ -92,6 +92,15 @@ export interface FanoutResult {
  * PGLite produces no parallelism, only queue pressure. The override is
  * still allowed (operator opt-in) but documented as ineffective on PGLite.
  */
+/** Worker capacity = every possible waiting parent plus one child slot. */
+export function resolveManagedWorkerConcurrency(fanoutMax: number): number {
+  return fanoutMax + 2; // per-source parents + global-maintenance parent + child
+}
+
+export function clampFanoutToManagedWorker(fanoutMax: number, workerConcurrency: number): number {
+  return Math.max(0, Math.min(fanoutMax, workerConcurrency - 2));
+}
+
 export async function resolveFanoutMax(engine: BrainEngine): Promise<number> {
   const override = await engine.getConfig('autopilot.fanout_max_per_tick');
   if (override) {
@@ -132,8 +141,8 @@ export async function readSupervisorConcurrency(queue = 'default'): Promise<numb
  * Resolve fanoutMax CLAMPED to the worker's effective concurrency (#2194 fix #1).
  *
  * Fanning out more cycles than the worker can run guarantees waiters that then
- * race the stalled-sweeper. Clamp to `max(1, concurrency - 1)` — reserving ≥1
- * slot for targeted sync/embed jobs that share the `default` queue.
+ * race the stalled-sweeper. Reserve one global-maintenance parent and one child
+ * slot; low-concurrency workers may therefore run zero per-source parents.
  *
  * codex #9 / D5: the clamp is BEHAVIOR-changing, so it trusts only a
  * proven-alive supervisor (live DB-lock holder, `ttl_expires_at`-gated). With
@@ -152,7 +161,7 @@ export async function resolveEffectiveFanoutMax(engine: BrainEngine, queue = 'de
     if (!snap || !isLockHolderLive(snap, SUPERVISOR_LOCK_TTL_MIN)) return base; // no live holder → unknown → no clamp
     const concurrency = await readSupervisorConcurrency(queue);
     if (concurrency === null) return base;
-    return Math.max(1, Math.min(base, concurrency - 1));
+    return clampFanoutToManagedWorker(base, concurrency);
   } catch {
     return base;
   }
--- a/src/commands/autopilot.ts
+++ b/src/commands/autopilot.ts
@@ -434,6 +434,7 @@ export async function runAutopilot(engine: BrainEngine, args: string[]) {
 
   let stopping = false;
   let childSupervisor: ChildWorkerSupervisor | null = null;
+  let autopilotWorkerConcurrency: number | null = null;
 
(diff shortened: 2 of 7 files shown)


dialthewolff:
@garrytan Maintainer action needed: please approve/run Actions on this external-contributor PR. Local targeted tests passed 44/44, `bun run verify` passed 31/31, and independent Codex blocker review returned PASS.

time-attack:
Thanks for looking at the fan-out/worker capacity interaction. I don't think this lands safely as written: clampFanoutToManagedWorker can return 0, so a low-concurrency worker dispatches zero per-source parents and autopilot silently does nothing — the previous Math.max(1, ...) floor was deliberate. Relatedly, the doctor change now recommends `gbrain config set autopilot.fanout_max_per_tick 0`, but resolveFanoutMax rejects any override below 1, so that advice can't take effect. Auto-deriving worker concurrency from fan-out also raises DB pool pressure, which is the contention problem Pace Mode exists to manage. Happy to revisit with a floor of 1 preserved and the concurrency derivation dropped.

PR #3470, opened 2026-07-28
fix: close federated read scope gaps

dialthewolff:
## Summary
- thread federated source grants through `get_chunks`, `get_raw_data`, and `get_versions` on both engines
- scope `code_def` and `code_refs` using the existing code-intel resolver
- make explicit remote `source_id` fail closed against either the federated grant or scalar source binding
- reconcile the completed P1 in `TODOS.md`

## Verification
- `bun test test/source-scope-resolver.test.ts test/code-def-refs.test.ts test/get-page-federated-scope.test.ts` (45 pass)
- `bun run verify` (31/31)
- independent Codex security review: PASS


dialthewolff:
@garrytan Maintainer action needed: please approve/run Actions on this external-contributor PR. Local source-isolation tests passed 45/45, `bun run verify` passed 31/31, and independent Codex security review returned PASS.

time-attack:
Reviewed carefully. **The core fix is real and worth landing — but not with the `resolveRequestedScope` hunk attached.** Please split.

## Confirmed: the read scope gap is real

Three sibling handlers in `src/core/operations.ts` each hard-drop the federated grant — `get_versions`, `get_raw_data`, and `get_chunks` all do `const sourceOpts = ctx.sourceId ? { sourceId: ctx.sourceId } : {}` and never consult `ctx.auth.allowedSources`. Combined with `src/commands/serve-http.ts:1882` (`const tokenSourceId = authInfo.sourceId ?? 'default'`) and `pglite-engine.ts:2418` (`opts?.sourceId ?? 'default'`), a client granted `allowedSources: ['beta']` arrives with `ctx.sourceId = 'default'` and reads the default source's content. And per the comment at `serve-http.ts:1879`, v60 backfilled `oauth_clients.source_id` NULL → `'default'`, so that's the *default* posture, not an edge case.

The engine side of your fix is good: parity is respected (both engines, symmetric `ANY($N::text[])`), the JSONB rule is respected (`text[]`, no `JSON.stringify`), and the fragment composition matches existing precedent at `postgres-engine.ts:3735`. Your new engine tests genuinely fail without the fix.

## Why it can't merge as-is

The `resolveRequestedScope` rewrite makes scalar `ctx.sourceId` a hard ceiling for every remote caller. I probed master vs this branch on four cases — master passes 4/4, this branch fails 3/4:

- `ctx = {remote:true, sourceId:'default', localFederatedSourceIds:['default','wiki']}` with `source_id:'wiki'` → master returns `{sourceId:'wiki'}`; this branch throws `permission_denied: source 'wiki' is outside your granted sources` — while the *unqualified* read on that same ctx still returns `{sourceIds:['default','wiki']}`. So you may read wiki implicitly but not name it.
- Plain stdio MCP (`src/mcp/server.ts:67` sets `remote:true` + `sourceId` from `GBRAIN_SOURCE`, no auth) with `source_id:'gstack'` → throws. That removes `source_id` from `query`, `get_page`, `list_pages`, `search`, `resolve_slugs`, `search_by_image`, and the `code_*` ops for every Claude Code / Claude Desktop user — i.e. one of the two organizational axes CLAUDE.md is built around.

This branch also deletes the master test that pinned the current behavior (`test/source-scope-resolver.test.ts`, "remote with no federated grant array can pass an explicit source_id (scalar-floor model)", from `1eb430a2` / #1999) and replaces it with the inverted assertion. When a test says a behavior is deliberate, inverting it is a design change that needs its own discussion, not a line in a bugfix PR.

**There is a narrower real hole nearby:** a scalar-*granted* token (`hasSourceGrant=true`) can name a sibling source. The right fix threads the existing `hasSourceGrant` concept from `src/mcp/http-transport.ts:242` into ctx so the ceiling applies only to operator-granted scalars. Worth a separate PR.

## Also scope creep

Flipping `code_def`/`code_refs` from brain-wide to source-scoped: the TODO you're closing explicitly says *"confirm whether brain-wide is intentional before scoping."* This decided it unilaterally, and via `resolveCodeIntelScope` a stdio-MCP agent can no longer span sources at all (`all_sources` collapses back to `ctx.sourceId` for remote).

## Ask

Drop the `resolveRequestedScope` hunk and its test edit, and drop the `code_def`/`code_refs` change. The engine + three-handler grant threading lands on its own and I'd take it. File the scalar-ceiling question and the code-intel scoping question separately.

For transparency on my own verification: I ran `bun test test/source-scope-resolver.test.ts test/code-def-refs.test.ts test/get-page-federated-scope.test.ts test/local-federated-search-scope.test.ts` → 82 pass / 0 fail. The Postgres e2e parity tests did **not** run here (`DATABASE_URL` unset), so the `ANY(...::text[])` arms are unverified against real Postgres.

dialthewolff:
Superseded by #3550, containing only the maintainer-requested engine + three-handler federated-read fix. The resolver, code-intel, and TODOS changes were removed.

PR #3472, opened 2026-07-28
fix(mcp): GBRAIN_MCP_TAKES_HOLDERS stdio takes-holder allow-list override (#2657)

Masashi-Ono0611:
Narrower re-send of the `GBRAIN_MCP_TAKES_HOLDERS` piece from #3126, which you closed as superseded but flagged as "the one piece with standalone merit" — the rest of that PR (take-proposal review CLI, calendar lookahead docs, pricing-drift guard) is intentionally left out here.

**Problem (#2657).** The stdio MCP transport hardcodes `takesHoldersAllowList: ['world']` with no escape hatch. Non-`world` takes are structurally unreachable via `takes_list` / `takes_search` / `query` for stdio MCP agents, even though `doctor` and the CLI can see them fine — confirmed independently on my own fork ([Masashi-Ono0611/gbrain](https://github.com/Masashi-Ono0611/gbrain)) when ~1176 extracted takes under the default `system`/`brain` holder were invisible via the primary MCP interface.

**Fix.** `resolveStdioTakesHolders()` reads `GBRAIN_MCP_TAKES_HOLDERS` (comma-separated, same escape-hatch pattern as `GBRAIN_SOURCE`) as an opt-in widening. Unset / empty / whitespace-only values keep the fail-closed `['world']` default — it can only widen, never silently shrink or flip open.

Tests cover: unset → default, empty/whitespace/garbage → default (never an empty allow-list), comma-split with trimming, and a single non-`world` holder being honored.

`bun run typecheck` clean; full test file green under both a clean env and an ambient `GBRAIN_MCP_TAKES_HOLDERS` (the one test that reads the parameter default is env-isolated so it can't pick up whatever the running shell happens to have set).

### DIFF
--- a/src/mcp/server.ts
+++ b/src/mcp/server.ts
@@ -15,6 +15,21 @@ import {
 } from '../core/context/resolve-ipc.ts';
 import { resolveEntitiesToPointers, logDeliveredReflexPointers } from '../core/context/retrieval-reflex.ts';
 
+/**
+ * #2657: stdio takes-holder allow-list. Fail-closed default ['world'] (the
+ * v0.28 contract — agent-facing callers never see private hunches unless the
+ * operator opts in). GBRAIN_MCP_TAKES_HOLDERS is the env escape hatch (same
+ * pattern as GBRAIN_SOURCE): comma-separated holder list, e.g.
+ * `GBRAIN_MCP_TAKES_HOLDERS=world,brain,people/alice-example`.
+ * Empty / whitespace-only values fall back to the default (never widen).
+ */
+export function resolveStdioTakesHolders(
+  raw: string | undefined = process.env.GBRAIN_MCP_TAKES_HOLDERS,
+): string[] {
+  const parsed = (raw ?? '').split(',').map((s) => s.trim()).filter(Boolean);
+  return parsed.length > 0 ? parsed : ['world'];
+}
+
 export async function startMcpServer(engine: BrainEngine) {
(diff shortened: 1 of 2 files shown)


Masashi-Ono0611:
FYI on the CI red here: `verify` / `test (*)` are failing at `bun install` with `ENOENT: could not open the "node_modules" directory` — that's the tracked `node_modules` symlink (`node_modules -> /tmp/fleet/repo/node_modules`) currently on `master`, which #3463 already has open. Confirmed this isn't from this diff: `master`'s own CI is red right now for the same reason. Not attempting to fix it here since #3463 already covers it.

time-attack:
Thanks for the clean implementation — the fail-closed default and the empty/whitespace handling are both done right. I'm still going to pass on it, though. The hardcoded ['world'] allow-list on stdio is a deliberate boundary rather than an oversight, and the documented way for an operator to read everything is already there: `gbrain call <op>`, which runs trusted with remote=false. Adding an env var that widens what an untrusted agent-facing transport can see gives us a second, weaker path to a capability that already has a trusted one.

Masashi-Ono0611:
Understood — "a second, weaker path to a capability that already has a trusted one" is a fair way to put it, and I hadn't weighed `gbrain call <op>` as the sanctioned read-everything path properly. I'll carry this as a fork-local patch on my own install rather than re-proposing. Thanks for engaging with the implementation details even while passing on the direction.

PR #3473, opened 2026-07-28
fix(sync): surface a failed updateSlug in the rename path instead of silently swallowing it (#3056)

Masashi-Ono0611:
Surfacing-only re-send of #3252, scoped to exactly what you flagged as reviewable: "returning the moved-row count from both engines, warning with `oldSlug -> newSlug` and both paths, and counting it in `SyncResult`." The reconcile half (locating and deleting the stale old row via a new `deletePage` call + three-outcome branching in the sync rename loop) is intentionally **not** included — that's the highest-blast-radius path in the system, per your review, and stays out of scope here.

**What changed.** `updateSlug` returns the number of rows moved (0 = no-op) in both `pglite-engine.ts` and `postgres-engine.ts`. The sync rename loop observes that return value — plus the throw case (destination collision) — to warn with both slugs and both paths, and counts it in `SyncResult.renameFallbacks`. Behavior is otherwise **unchanged**: a fallback still resolves as "treat as add," nothing is deleted or reconciled.

**Being upfront about what this does and doesn't fix.** In the collision case (destination slug pre-occupied), `updateSlug` throws, so `importFile` at the new path upserts the pre-existing destination row in place — silently overwriting whatever content was there before, while the orphaned old-slug row is left behind untouched. That data-loss shape is pre-existing (not introduced by this diff) and this PR does not change it — it only makes the fact that a fallback happened *visible*, via the warn + counter, instead of silent. Fixing the overwrite itself is exactly the reconcile work you scoped out of #3252.

**Testing.** `test/sync-rename-fallback.serial.test.ts`: engine contract (`updateSlug` returns 1 / 0), collision fallback (surfaced, old row left in place), zero-row fallback (divergent stored slug, surfaced), happy path (clean rename, `renameFallbacks` 0, page_id preserved), and a partial-result test proving `renameFallbacks` survives a mid-run abort into the returned partial `SyncResult` (deterministic — wraps `updateSlug` to abort after the first call rather than racing a checkpoint poll).

That last test exists because my first pass had a real bug here: `buildPartialResult` didn't carry the field at all, so a timeout/checkpoint-abort mid-rename-loop would silently drop the count — caught by an independent Codex review pass, along with a test-isolation bug (my test changed `process.env.HOME`, which doesn't actually redirect Bun's `os.homedir()` — needed `GBRAIN_HOME`) and two lines that overshot your stated scope (an aggregate warning + an ingest-summary suffix), both removed.

`bun run typecheck` clean. Full new test file green (6/6, repeated 5x with no flakes). Ran the broader `sync` / `pglite-engine` / `source-id-tx-regression` / `gbrain-home-isolation` / `sync-resumable-import` suites (225 tests) with no regressions. Postgres-engine parity for the new return-value contract is untested — no Docker on this machine — the implementation mirrors the established `result.count` pattern already used elsewhere in that file, but flagging the gap rather than claiming coverage I don't have.

### DIFF
--- a/src/commands/sync.ts
+++ b/src/commands/sync.ts
@@ -199,6 +199,16 @@ export interface SyncResult {
   pagesAffected: string[];
   failedFiles?: number; // count of parse failures (Bug 9)
   /**
+   * #3056: renames whose cheap path (updateSlug) did not move a row — either
+   * the UPDATE matched nothing (old slug absent in the scoped source) or it
+   * threw (destination slug occupied, invalid slug). Each one fell back to
+   * add semantics, same as before this counter existed; the difference is
+   * it's now surfaced (counted here + a per-file warn naming both slugs)
+   * instead of silently swallowed. 0 / absent = every rename took the cheap
+   * page_id-preserving path.
+   */
+  renameFallbacks?: number;
+  /**
    * v0.41.13.0 partial-sync fields (only set when status === 'partial').
    *
    * D-V3-1 (honest scope): --timeout aborts ONLY in pre-bookmark phases
@@ -1794,6 +1804,8 @@ function buildPartialResult(opts: {
   renamed: number;
   reason: 'timeout' | 'pull_timeout' | 'pull_failed' | 'stall_timeout' | 'checkpoint_unavailable';
   bankedFiles?: number;
+  /** #3056: renames that fell back to add semantics before this abort. */
+  renameFallbacks?: number;
 }): SyncResult {
   return {
     status: 'partial',
@@ -1809,6 +1821,7 @@ function buildPartialResult(opts: {
     filesImported: opts.filesImported,
     reason: opts.reason,
     bankedFiles: opts.bankedFiles,
+    ...(opts.renameFallbacks ? { renameFallbacks: opts.renameFallbacks } : {}),
   };
 }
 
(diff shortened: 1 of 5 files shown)


Masashi-Ono0611:
FYI on the CI red here: `verify` / `test (*)` are failing at `bun install` with `ENOENT: could not open the "node_modules" directory` — that's the tracked `node_modules` symlink (`node_modules -> /tmp/fleet/repo/node_modules`) currently on `master`, which #3463 already has open. Confirmed this isn't from this diff: `master`'s own CI is red right now for the same reason. Not attempting to fix it here since #3463 already covers it.

time-attack:
You've found a real defect — the bare `catch {}` in the rename path swallows genuine DB errors as readily as slug collisions, and a zero-row UPDATE doesn't throw at all, so sync reports success either way. My hesitation is with the shape of the fix rather than the diagnosis: after this change the stale row is still left live alongside the newly created one, so the actual integrity problem survives and we've added a SyncResult field plus an engine signature change to describe it. I'd very much like a PR that reconciles the duplicate — that's the fix worth the surface area, and it should be a much smaller diff.

Masashi-Ono0611:
That works for me — reconcile PR incoming. Shape based on your reviews of both #3252 and this one: locate the stale row positively by `source_path = from` (never the slug guess), delete it only after the destination demonstrably materialized (so a dedup-skip against the old row can't widen into deleting the only copy), surface a reconcile failure through `failedFiles` so the existing failure gate retries it — and none of the `SyncResult` reporting surface from this PR.

Masashi-Ono0611:
Reconcile PR is up: #3479.

PR #3474, opened 2026-07-28
fix(cycle): extract_facts guard counts only active legacy rows; reconcile preserves forget records (#2646)

Masashi-Ono0611:
Narrowed re-send of #3234, scoped to exactly what you named reviewable when closing it: "the fix for the reported bug is one predicate plus the reconcile-preservation guard." The drift-repair lane, the re-runnable migration orchestrator path, and the race-accounting layer are all **excluded** (original ~1050 lines → ~345, over half of which is tests).

**The bug (#2646).** master's empty-fence guard counts soft-expired legacy rows (`row_num IS NULL`, `expired_at` set). `forget_fact` — the sanctioned removal path — soft-expires legacy rows rather than deleting them, so the backlog can never drain: `apply-migrations` no-ops (already marked applied), the guard stays triggered forever, and `extract_facts` jams.

**The fix, in two parts:**
1. **The predicate**: the guard's legacy COUNT adds `AND f.expired_at IS NULL`, so each `forget_fact` visibly drains the pending counter toward release.
2. **The reconcile-preservation guard**: once the guard stops counting expired rows, the destructive reconcile can newly run on pages that carry them — so it must not destroy the forget records. `listExistingFactsForPage` excludes soft-expired legacy rows (never fence-owned: they must neither read as perpetually stale nor mask a fence row from insertion), and the two wipe call sites pass `preserveExpiredLegacy: true` so `deleteFactsForPage` keeps them. That option is the smallest seam I could find for the guard — ~6 lines in each engine, mirrored exactly. If you'd rather not grow `deleteFactsForPage`'s options even by this much, say the word and I'll rework it, but every alternative I considered (inline SQL in the phase, read-and-reinsert) seemed worse.

**Semantics pinned by test, not hidden**: if the fence still carries the same claim as an expired legacy row, the reconcile reinserts it as a fresh active fence-owned row — legacy DB-only forgets are documented non-durable (forget.ts header), and suppressing the insert would create the silent fence↔DB divergence the guard exists to prevent. The expired row survives alongside as the audit record of the earlier forget. There's a dedicated test asserting this exact coexistence stays idempotent across runs.

**Testing.** 5 new cases in `test/extract-facts-phase.test.ts` (guard drains via forget_fact; hybrid expired row survives reconcile untouched; survives even a stale-row wipe on the same page; fence-canonical reinsertion is idempotent; mixed active+expired rows count only the active one). Plus 3 Postgres cases in `test/e2e/facts-fence-reconcile-postgres.test.ts` pinning the postgres.js tagged-fragment SQL in both `deleteFactsForPage` branches and the returned delete count — being upfront: those three are DATABASE_URL-gated and I could not run them locally (no Docker on this machine); they load and skip cleanly, the SQL columns are verified against the v45 DDL in `migrate.ts`, but their first real execution will be this PR's CI. The PGLite half (25/25) and the adjacent facts suites (93 pass total across 10 files) are green locally, `bun run typecheck` clean.

An independent Codex review pass over the narrowed diff confirmed engine parity and no leakage from the excluded migration-lane code, and caught one doc-comment contradiction (the engine.ts JSDoc implied preservation prevents revival; reworded — preservation protects the audit record, fence canonicality independently permits reactivation).


Masashi-Ono0611:
FYI on the CI red: same known cause as the rest of today's PRs — the tracked `node_modules` symlink on `master` fails every `bun install` with `ENOENT: could not open the "node_modules" directory` (#3463 has the fix open). Verified in this run's `verify` job log; not from this diff.

time-attack:
Merging. Verified the full three-link chain on master rather than relying on the writeup.

1. The guard COUNT at `src/core/cycle/extract-facts.ts:198-209` has no `expired_at IS NULL`, and `legacyCount > 0` returns before the per-page loop (`:212-219`) — so the entire fence→DB reconcile is dead for the whole run.
2. `forget_fact` can't clear it: `src/core/facts/forget.ts:103-112` sets `canFence = row.row_num !== null`, false for a legacy row, so it falls to `expireFact` which only stamps `expired_at`. `row_num` stays NULL and the row stays counted.
3. `apply-migrations` genuinely no-ops — `src/commands/apply-migrations.ts:199` classifies a ledger-complete v0.32.2 as applied, and Phase B won't re-run without `--force-retry`.

So one soft-expired row with a live page yields `factsInserted=0` for every page in the brain, every cycle, indefinitely.

Confirmed the narrowing held: the drift-repair lane, the orchestrator re-run path, and the race-accounting from #3234 all stayed out. The `preserveExpiredLegacy` half isn't scope creep — once the guard stops counting those rows the reconcile reaches them, and `listExistingFactsForPage` has no expiry filter either, so they'd read as stale and get wiped. It's the necessary consequence.

Engine parity checked line by line: the predicate is identical in both engines and appended in both branches. `idx_facts_fence_key` is partial on `row_num IS NOT NULL`, so the preserved expired row can't collide with the fresh fence row. Tests: 25 pass with the fix; reverting `src/` to master while keeping the tests gives 20 pass / 5 fail, so they pin behavior rather than scaffolding. Adjacent 47-file sweep is 531/9, byte-identical to the master baseline.

**Two things I'm accepting knowingly, for the record:** this doesn't restore an automated drain (active legacy rows still gate, and `apply-migrations` still no-ops — the excluded lane), so #2646 is only partly resolved and the drain stays manual. And claim revival on a still-present fence row is pre-existing, not new — master does the same on any page where the guard didn't fire, except it also destroys the audit row, which this preserves.

Your new Postgres E2E file is `DATABASE_URL`-gated and reads correct, but its first real execution will be CI — flagging that since you noted you couldn't run it.

PR #3475, opened 2026-07-28
feat(config): embedding_max_chunk_tokens — opt-in estimated-token chunk cap for strict embedding backends

paul-0320:
Refs #2826. This is the config-driven reshape of #2847 / #3074, built to the shape named in their closing review: an opt-in token limit that "defaults to today's behavior."

## Why

URL/phone/email-dense CJK markdown and large JSON code blocks produce 3–4K-char chunks that overflow strict per-request embedding-server token limits (measured: a local llama-server with `-ub 2048` EOFs deterministically past ~2,050 tokens per chunk; Ollama's runner fails the same way past its physical batch size — full reproduction in #2826). The word pipeline's whitespace fallback counts a 150-char URL as one word, so the existing `maxChars: 6000` cap never sees the real token mass. Default providers at 8,192 tokens can't hit this; self-hosted backends can, deterministically.

## What changed vs the closed PRs (blast-radius receipt)

| #2847/#3074 objection | this PR |
|---|---|
| Chunker version bumps force a full re-chunk + re-embed of every install | **No version bumps** (markdown stays v3, code stays v4). Nobody re-chunks unless they opt in and run the sweep themselves |
| Global 1500 default shrinks everyone's CJK/dense chunks; ~3× ASCII overestimate shrinks ordinary English chunks | **No default.** The cap exists only when `embedding_max_chunk_tokens` is set. Unset = **byte-identical chunking**, pinned by tests (including a pin that the URL-dense failure shape still emits over-cap chunks by default — proof the default path takes no cap) |
| Global retrieval-behavior change with no eval receipts | Default behavior unchanged ⇒ nothing to eval. Opted-in installs change chunks only for (a) content that today produces un-embeddable chunks on their backend (no vectors at all), and (b) whatever their chosen cap conservatively splits — an explicit, documented trade-off (see below) |

## Changes

- **`src/core/config.ts`** — new key `embedding_max_chunk_tokens` (positive int). All three planes: `config.json`, `GBRAIN_EMBEDDING_MAX_CHUNK_TOKENS` env, DB plane via `gbrain config set` (merged in `loadConfigWithEngine`, env/file wins, invalid/non-positive values ignored). Registered in `KNOWN_CONFIG_KEYS`.
- **`src/core/cjk.ts`** — `estimateEmbeddingTokens()`: tokenizer-free per-char-class estimate (CJK 1.0/char, other non-whitespace 0.75/char, whitespace 0.1/char). Calibrated against measured ratios (URL-dense ~1.6 chars/token, base64-worst ~1.3) and deliberately **overestimates**, so a cap based on it holds against real tokenizers. Additive exports only — `countCJKAwareWords` and `search/expansion.ts` are untouched.
- **`src/core/chunkers/recursive.ts`** — `ChunkOptions.maxEmbedTokens` (opt-in): when set, `capByEstimatedTokens()` runs as a final pass after `capByChars`, guaranteeing `estimateEmbeddingTokens(chunk) ≤ cap` for every emitted chunk. Cut points prefer a newline, then whitespace, within a 300-char lookback, so forced splits land at line boundaries instead of mid-URL. Pieces reassemble **byte-for-byte** (`pieces.join('') === input`). When unset: the option resolves to `undefined` and the added code paths are skipped entirely.
- **`src/core/chunkers/code.ts`** — `CodeChunkOptions.maxEmbedTokens` (opt-in): same final pass on AST-path chunks (the `splitLargeNode`-returns-empty case ships whole at any size). Composes with #1675's `capOversizedChunks`: that cap triggers on the generic `estimateTokens` (~3.5 chars/token), which undercounts CJK ~3.5× and URL-dense text ~2×, so an over-budget Korean chunk passes it untouched; `capCodeChunks` re-checks with the CJK-aware estimate and is a no-op when everything is already under cap. Threaded into the fallback path via `ChunkOptions.maxEmbedTokens`.
- **`src/core/import-file.ts`** — the page-ingest path (`importFromContent`: prose, timeline, and fenced-code chunks) resolves the key from the same effective config as the content-sanity gate (no additional config round-trips beyond the existing lift) and threads it into both chunkers.

**Scope note**: the standalone code-file import path (`importCodeFile`) is not auto-wired — the measured failures are all markdown-page shaped, and wiring it means either per-file DB config reads in bulk code sync or a plane asymmetry. `CodeChunkOptions.maxEmbedTokens` is exposed, so wiring it (resolved once per sync run) is a clean follow-up if wanted.

## Opt-in workflow

```
gbrain config set embedding_max_chunk_tokens 1500   # headroom for the contextual-retrieval prefix under a ~2,050-token limit
gbrain reindex --markdown                            # re-chunk existing pages; new ingests apply immediately
```

## The conservatism trade-off, stated plainly

The 0.75/char ASCII weight overstates real Latin tokenization ~3× **by design** — the cap must hold against any tokenizer without shipping one. Consequence for opted-in installs: a 1500 cap splits merge-inflated (~450-word) English chunks that a real tokenizer would count at ~600 tokens. This is pinned by a dedicated test so it's a documented property, not a surprise, and it's the user's dial: pick the cap for your backend's real limit; don't set it at all on 8K-token providers.

## Measurements (worst chunk, real Qwen3-family tokenizer via llama-server `/tokenize`, re-measured on THIS branch's code)

| input | cap unset (= today) — real tokens | cap 1500 — real tokens (estimate) |
|---|---|---|
| URL-dense Korean rollup (60 lines) | **2,462** (past the ~2,050 server limit) | 1,208 (est 1,481) |
| 7K-char minified JSON | 2,004 | 1,040 (est 1,499) |
| 7K-char pretty JSON | 1,836 | 1,050 (est 1,499) |

The estimator stayed above the real count in every case (1,481–1,499 estimated vs 1,040–1,208 real), so the guarantee holds with margin. A production-scale replay of the cap logic over 157,823 chunk rows (posted on #2847's review by @stigrunar) split 15,972 over-cap rows into 181,165 pieces with zero over-cap survivors — and surfaced a lossless-invariant bug in the original diff (`trim()` on forced-split pieces dropped boundary whitespace on 15,882 rows). That fix is folded in here, with `pieces.join('') === input` pinned in the unit tests.

## Tests

`test/chunkers/token-cap.test.ts` (17 tests):

- **default behavior pins**: URL-dense failure shape still emits over-cap chunks when the option is unset (the cap is opt-in, not a silent default); unset vs `{}` produce identical output,
- opt-in cap invariant for both failure shapes (URL-dense CJK rollup; large JSON pretty/minified through the prose path and `fence.json` through the code-chunker path),
- content preservation + byte-for-byte reassembly (regression for the #2847 trim bug),
- newline-preferring cut placement, forward progress on whitespace-less input,
- normal-content safety: Latin/CJK prose chunks identical with and without a clearing cap; the tight-cap conservative split pinned explicitly,
- `estimateEmbeddingTokens` weight sanity.

`test/loadConfig-merge.test.ts` (+4): DB-plane merge, file/env precedence, unset stays undefined, invalid values ignored.

## Verification

- `bun run verify`: **31/31 green**.
- Unit suite (project shard runner, `bun run test`): run on this branch vs an `upstream/master` (ddd66e1d) baseline on the same machine — **zero failures unique to this branch**. Branch: 10,526 pass across 3 completed shards, 6 failures (1 sink-drain flake + 5 serial), every one of which also fails on the baseline; baseline additionally fails 3 audit-isolation flakes. My environment has no embedding provider configured, which wedges one shard on the branch run and two on the baseline run (pre-existing, load-sensitive — the same class documented on #2847); the changed surface is fully covered by the directly-exercised domain suites (chunkers / cjk / config / import-file, 100% green above).
- `bun run test:slow`: 41/41.
- E2E: skipped locally (no `DATABASE_URL`), per the `test:full` documented skip path. Engine-agnostic change — pure text pipeline + config plumbing; chunk rows flow through both engines' existing `upsertChunks` unchanged, and the config merge is exercised against the engine interface with fake-engine tests.
- Eval replay: not triggered — no CONTRIBUTING-listed retrieval trigger paths touched, and default chunk boundaries are pinned unchanged by the test suite.

### DIFF
--- a/src/core/chunkers/code.ts
+++ b/src/core/chunkers/code.ts
@@ -18,8 +18,9 @@
  * at runtime.
  */
 
-import { chunkText as recursiveChunk } from './recursive.ts';
+import { chunkText as recursiveChunk, capByEstimatedTokens } from './recursive.ts';
 import { buildQualifiedName } from './qualified-names.ts';
+import { estimateEmbeddingTokens } from '../cjk.ts';
 
 // Embed the tree-sitter runtime + per-language grammars as files.
 // `with { type: 'file' }` returns a path (string) at runtime. Bun bundles
@@ -177,6 +178,18 @@ export interface CodeChunkOptions {
    * smallest common embedder context (e.g. nomic-embed-text, 2048).
    */
   maxChunkTokens?: number;
+  /**
+   * Opt-in hard cap on any emitted chunk's ESTIMATED embedding tokens, for
+   * embedding backends with strict per-request token limits. Unlike
+   * `maxChunkTokens` (structural re-split triggered by the generic
+   * estimateTokens, ~3.5 chars/token — an underestimate for CJK and
+   * URL-dense text), this is measured with the conservative CJK-aware
+   * estimateEmbeddingTokens (cjk.ts) and enforced as a final pass, so it
+   * GUARANTEES the bound. UNSET (default): no final cap — behavior is
+   * identical to previous releases. Wired to the
+   * `embedding_max_chunk_tokens` config key by the import path.
+   */
+  maxEmbedTokens?: number;
 }
 
 /**
@@ -718,7 +731,14 @@ export async function chunkCodeTextFull(
     if (chunks.length === 0) {
       return { chunks: capOversizedChunks(fallbackChunks(source, filePath, language, opts), filePath, language, opts), edges: rawEdges };
     }
-    return { chunks: capOversizedChunks(mergeSmallSiblings(chunks, chunkTarget), filePath, language, opts), edges: rawEdges };
+    // Layer both caps when maxEmbedTokens is set. capOversizedChunks (#1675)
+    // splits by the generic estimateTokens (~3.5 chars/token), which
+    // undercounts CJK (~1 char/token) and URL-dense text — a Korean chunk 3x
+    // over the real limit passes its trigger untouched. capCodeChunks
+    // re-checks with the CJK-aware estimateEmbeddingTokens and is a no-op
+    // when every chunk is already under the cap.
+    const astChunks = capOversizedChunks(mergeSmallSiblings(chunks, chunkTarget), filePath, language, opts);
+    return { chunks: opts.maxEmbedTokens ? capCodeChunks(astChunks, opts.maxEmbedTokens) : astChunks, edges: rawEdges };
   } catch {
     return { chunks: fallbackChunks(source, filePath, language, opts), edges: [] };
   } finally {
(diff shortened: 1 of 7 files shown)


time-attack:
Careful work, and the underlying observation is correct: the generic ~3.5 chars/token estimate does undercount CJK and URL-dense text. But this adds a second, parallel cap alongside the existing maxChunkTokens mechanism, plus a new config key and a CJK module, defaulting to off — so it's a lot of permanent surface area for no behavior change out of the box. If the existing estimator is wrong for CJK, the fix I'd take is making capOversizedChunks use the CJK-aware estimate, which should be a handful of lines and improves the default path for everyone instead of gating it behind a knob nobody will find.

Issue #3037, opened 2026-07-21
Chunker caps by characters (maxChars=6000) but embedders limit by tokens — CJK/dense content permanently unembeddable, silently, exit 0


johnno100:
## Summary

The recursive chunker caps chunks by **character** count (`maxChars`, default 6000), but embedding models limit by **token** count. For CJK, or dense tabular/URL-heavy content, 6000 characters exceeds `nomic-embed-text`'s 2048-token context, so those chunks can never be embedded. `gbrain embed --stale` fails on them on every run and **exits 0**, so the failure is silent.

Observed on a 4,360-page corpus: **258 chunks across 28 pages permanently unembeddable** (98.42% coverage, stuck).

- Version: `gbrain 0.42.62.0`
- Embedding: `ollama:nomic-embed-text` (768d), Ollama OpenAI-compatible endpoint
- Engine: Postgres + pgvector

## Root cause

`src/core/chunkers/recursive.ts:75`

```ts
const maxChars = opts?.maxChars || 6000;   // "sliding-window safety belt"
```

The cap is in characters. The constraint it is protecting is in tokens. The two only correlate for ASCII prose (~4 chars/token). CJK tokenizes at roughly 1–2 tokens per character, so a 6000-char chunk can be 3–6k tokens — far past the 2048 context.

## Reproduction

A real chunk from the corpus: **6001 characters, 942 of them CJK** (a bilingual EN/中文 page).

```bash
# native endpoint
curl -s http://OLLAMA:11434/api/embed -H 'Content-Type: application/json' \
  -d @chunk.json    # {"model":"nomic-embed-text","input":"<6001 chars>","truncate":false}
# {"error":"the input length exceeds the context length"}

# OpenAI-compatible endpoint — same result, HTTP 400
curl -s http://OLLAMA:11434/v1/embeddings -H 'Content-Type: application/json' \
  -d @chunk_v1.json
# {"error":{"message":"the input length exceeds the context length",
#           "type":"invalid_request_error"}}
```

Controls that isolate it to tokens-vs-chars:

| input | result |
|---|---|
| the real 6001-char chunk (942 CJK) | **HTTP 400** |
| same chunk truncated to 4000 chars | HTTP 200 |
| synthetic **6000-char pure ASCII** (~1500 tok) | HTTP 200 |
| batch of 14 × 6000-char ASCII in one request | HTTP 200 |

So it is not batch size, not a transient, and not embedder capacity — it is purely a function of the token density of the chunk text.

[...]

PR #3477, opened 2026-07-28
fix(chunker): CJK-aware oversize measurement + close capOversizedChunks' fallback-path gaps

paul-0320:
Refs #2826. The shape named in #3475's closing review: `capOversizedChunks` measures with a CJK-aware estimate — no new config key, no new module (reuses `cjk.ts`'s existing exports), no parallel cap.

## Measurement first — where the current estimator actually diverges

Counted against the Qwen3-Embedding tokenizer (the strict-backend family from #2826), same inputs through the current `estimateTokens`:

| content | cl100k (current) | Qwen3 (real embedder) | divergence |
|---|---|---|---|
| English prose (1,000c) | 175 | 175 | identical |
| JSON (1,000c) | 389 | 389 | identical |
| Korean prose (500c) | 433 | 356 | cl100k already conservative |
| **URL-dense Korean mix (2,389c)** | **1,039** | **1,509** | **−31% undercount** |

So with tiktoken loaded, cl100k is already right for pure-ASCII source and already conservative for CJK-dominant text — the gap is **mixed CJK+ASCII chunks**, exactly the #2826 failure shape. (The heuristic fallback, when tiktoken doesn't load, undercounts CJK ~2.5×.)

## Change 1 — `estimateEmbedTokens()` at capOversizedChunks' measurement points

For chunks containing CJK: `max(estimateTokens, ceil(cjk·1.0 + otherNonWs·0.75 + ws·0.1))`. Consequences, all pinned by tests:

- **ASCII-only chunks short-circuit to `estimateTokens` verbatim** — bit-identical measurement, so cap decisions and boundaries are unchanged for English/JSON source.
- **CJK-dominant text unchanged too** — cl100k already exceeds the weighted form there, so `max()` returns today's value (measured: Korean prose 1,302 cl100k vs 1,162 weighted).
- Only mixed-script chunks estimate higher. Margins vs the real Qwen3 count on the divergence shapes: +15% (URL-dense Korean, 2.4K), +22% (same at 12K), +16% (JSON with Korean values) — above the real count everywhere, i.e. the cap now actually holds on strict backends.

The hard-split slice budget becomes 1 char/token for CJK-bearing pieces (the weighted estimate can reach 1 token/char); ASCII pieces keep the 3.5 chars/token slice exactly as today.

## Change 2 — the safety net now covers every fallback emission

While testing this I found `capOversizedChunks`' stated contract ("guarantee no emitted chunk exceeds the budget") has holes independent of the estimator: only the empty-AST branch routes its fallback through the cap. The no-language, parse-timeout, no-semantic-nodes, and parse-throw branches ship word-counted chunks unchecked — and **every JSON/YAML fence takes the no-semantic-nodes branch deterministically** (their node types aren't in `TOP_LEVEL_TYPES`), so a 14K-char JSON fence emits ~2,700-token chunks past the 2,000 default cap today, in pure ASCII, with tiktoken loaded. Hoisting the cap into `fallbackChunks` covers all five paths with one wrap (and un-nests the empty-AST call site).

## Tests

`test/chunkers/cap-oversized-cjk.test.ts` (6): ASCII bit-identity of the estimator; max-composition invariant; mixed-script lift; oversized JSON-with-Korean fence re-splits under the default cap with content preserved; whitespace-less mixed blob makes progress and stays under cap; ASCII chunks measured by the identical estimator. Existing chunker suites untouched and green.

## Verification

- `bun run verify`: 31/31 green. `bun run test:slow`: 41/41.
- Chunker/CJK/import-file domain suites: 139 pass / 0 fail. Typecheck clean.
- E2E: skipped locally (no `DATABASE_URL`), per the documented `test:full` skip path — engine-agnostic text-pipeline change; no persistence or retrieval trigger paths touched, so no eval replay per CONTRIBUTING.
- No `CHUNKER_VERSION` bump included: output changes only for chunks that are over-budget on strict backends (currently un-embeddable there) and for the previously-uncapped fallback emissions. If you'd rather force existing brains to converge on next sync, I'm happy to add the 4→5 bump.


time-attack:
Merging — verified independently, and this is a good fix. Notes so the follow-up is clear.

**Confirmed on master.** `src/core/chunkers/code.ts:719` is the only call site wrapping `fallbackChunks(...)` in `capOversizedChunks`; lines 579 (no language), 609 (parse timeout), 621 (no semantic nodes), and 723 (parse throw) all return it raw. JSON/YAML fences hit `:621` deterministically since `TOP_LEVEL_TYPES` (`code.ts:283`) has no JSON/YAML entry, and this is a live path — `src/core/import-file.ts:158` runs `chunkCodeText` per markdown fence and `FENCE_TAG_TO_PSEUDO_PATH` includes `json`/`yaml`/`yml`.

Reproduced at defaults (14,362-char ASCII json fence, cap 2000): 2 chunks over cap, max 2,299 tokens. Your "~2,700" was directionally right; I measure 2,299.

Independently verified the test discriminates behavior rather than just the new export: applying the test to master hard-fails on `estimateEmbedTokens` not being exported, so I stubbed only the estimator into master's `code.ts` and reran — **5 pass / 1 fail, `Expected: <= 2060, Received: 2680`**. The fallback-cap hoist is the load-bearing change. Regression sweep across 13 chunker/import/reindex files: 152 pass / 0 fail, and the recursive-chunker output is byte-identical to master.

**Three things this does NOT do — flagging so nobody assumes otherwise:**

1. **#2826 stays open.** The cap still doesn't hold: an ASCII JSON fence still emits 2,070- and 2,095-token chunks, over both the 2,000 cap and #2826's 2,048 strict-backend limit. Cause: `splitToTokenBudget` re-invokes `recursiveChunk` with the same 300-word/50-overlap params that produced the oversize piece, and the hard-split fallback assumes 3.5 chars/token for ASCII while URL-dense JSON runs ~2.6. Deriving the budget from the measurement — `floor(piece.length * cap / estimateEmbedTokens(piece))` — is roughly a one-liner and would close it properly.
2. **#2826's headline markdown reproduction is untouched** — `recursive.ts` still yields a 4,192-char / 2,182-token chunk at defaults for URL-dense Korean, identical before and after.
3. **`estimateEmbedTokens` is in the wrong home.** `code.ts` imports `recursive.ts`, so `recursive.ts` can never reuse it without a cycle. It belongs in `src/core/cjk.ts`, which would also let `capByChars` use it — that's the fix that actually closes #3037 and #2826.

No `CHUNKER_VERSION` bump, so existing over-cap fenced chunks persist until a reindex. Leaving it that way deliberately — consistent with not forcing a re-chunk sweep on every install.

Follow-up welcome for (1) and (3) together.

PR #3478, opened 2026-07-28
fix: harden multi-source maintenance and Chronicle backfill

frxiaobei:
## Summary

- restore the self-contained `backfill` CLI dispatch and keep `retrieval-upgrade`
- prevent non-federated link extraction from falling back across sources
- resolve doctor image assets against each source root and deduplicate logical assets by content hash
- skip pages already projected into Chronicle during bulk backfill
- retry a confirmed truncated Chronicle judge as two disjoint segments, preserving the all-or-nothing write barrier

## Why

These issues surfaced during a real multi-source maintenance pass:

- `gbrain backfill modality` was implemented but unreachable from the CLI dispatcher
- link extraction could recreate cross-source edges for an isolated source
- image doctor checks used the global sync root and double-counted duplicate asset rows
- Chronicle backfill repeatedly enqueued pages that already had event projections
- one dense conversation still hit the provider output cap after raising the configured judge cap

The Chronicle retry keeps the normal one-call path. It splits only after `stopReason=length`, requires both halves to return usable output, and writes nothing if either half truncates, fails to parse, is refused, or is filtered.

## Validation

- `70 pass, 0 fail` across:
  - `test/cli.test.ts`
  - `test/extract-stale.test.ts`
  - `test/doctor-image-assets.test.ts`
  - `test/chronicle-backfill.test.ts`
  - `test/chronicle-extract.test.ts`
- `bun run verify`: `32/32` checks green with an isolated `GBRAIN_HOME`
- production-like validation: a previously `judge_truncated` dense page completed through the segmented retry, with the existing write barrier and idempotent event path intact

Draft while the full local test suite and upstream CI finish.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -55,13 +55,14 @@ export function bigintToStringReplacer(_key: string, value: unknown): unknown {
 }
 
 // CLI-only commands that bypass the operation layer
-export const CLI_ONLY = new Set(['init', 'reinit-pglite', 'upgrade', 'post-upgrade', 'check-update', 'integrations', 'publish', 'check-backlinks', 'lint', 'report', 'import', 'export', 'files', 'embed', 'serve', 'call', 'config', 'doctor', 'migrate', 'eval', 'sync', 'extract', 'extract-conversation-facts', 'enrich', 'features', 'autopilot', 'graph-query', 'jobs', 'agent', 'apply-migrations', 'skillpack-check', 'skillpack', 'resolvers', 'integrity', 'repair-jsonb', 'orphans', 'maintain', 'sources', 'mounts', 'dream', 'check-resolvable', 'routing-eval', 'skillify', 'smoke-test', 'providers', 'storage', 'repos', 'code-def', 'code-refs', 'reindex', 'reindex-code', 'reindex-frontmatter', 'code-callers', 'code-callees', 'reconcile-links', 'frontmatter', 'auth', 'friction', 'claw-test', 'book-mirror', 'takes', 'think', 'salience', 'anomalies', 'calibration', 'transcripts', 'models', 'remote', 'recall', 'forget', 'edges-backfill', 'cache', 'ze-switch', 'retrieval-upgrade', 'founder', 'brainstorm', 'lsd', 'schema', 'capture', 'onboard', 'conversation-parser', 'status', 'connect', 'skillopt', 'quarantine', 'self-upgrade', 'advisor', 'watch', 'reindex-search-vector']);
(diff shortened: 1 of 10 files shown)


time-attack:
Reviewed. Three of the bugs in here are real and confirmed against master — but this is five unrelated fixes plus a self-declared `WIP:` commit, so it can't be reviewed or landed as one unit. Please split.

**Confirmed real:**
1. `backfill` is missing from `CLI_ONLY` (`src/cli.ts:58`) while its dispatch sits at `src/cli.ts:2168` inside the switch `handleCliOnly` only reaches when `CLI_ONLY.has(command)` (gate at `src/cli.ts:327`). So `gbrain backfill modality` is unreachable dead code, with no test covering it. **This one-liner is shippable today** — send it alone and it goes straight in. (Note #3251 also touches this area; check for overlap.)
2. `src/core/operations.ts:5665` enqueues `chronicle_extract` with `sourceId: ctx.sourceId ?? 'default'` instead of `page.source_id` — a cross-source misroute on any federated scan. Real source-isolation bug.
3. The doctor image check resolving every `storage_path` against the single global `sync.repo_path`.

**One thing to fix regardless of how you split it:** `chronicle_backfill`'s new `covered` query is a full unscoped `pages ⋈ timeline_entries` scan with no `LIMIT` and no `sourceScopeOpts`, and it runs on every invocation including the empty case. Per CLAUDE.md every read-side path routes through `sourceScopeOpts(ctx)` — a missed thread there is a cross-source leak, not just a slow query.

Keep the segmented Chronicle judge retry out of the split entirely; that's the risky part and shouldn't ride along with one-line dispatch fixes.

Also FYI: no upstream CI has run on this branch (`statusCheckRollup` is empty — fork workflows await maintainer approval), so the fork-side green runs in the body don't tell us about this base.

time-attack:
Status update: reviewed in the 2026-08-18 full-backlog triage — still wanted, but not yet mergeable: Five unrelated fixes in one draft; reviewer asked to split; several bugs still real on master but branch is conflicting and untouched since 2026-07-28. If you update it, it will be re-reviewed promptly; alternatively the fix may be adopted with credit in an upcoming fix wave. Thanks for the contribution and the patience.

time-attack:
Re-trialed against the current integration branch (post v0.46.23-25 waves): Conflicts: src/cli.ts, src/commands/{doctor,extract}.ts, src/core/operations.ts, test/{chronicle-backfill,doctor-image-assets}.test.ts. Backfill dispatch piece superseded; 4 fixes still novel but need rebase onto ops/doctor peels. — a rebase along those lines puts this back in the active merge queue; re-trials run nightly, so a fresh push gets picked up automatically. Thank you!

garrytan:

PR #3479, opened 2026-07-28
fix(sync): reconcile the stale old row when a rename falls back to add (#3056)

Masashi-Ono0611:
The reconcile PR you asked for on #3473. Src diff is ~100 lines (engines + sync loop); the rest is tests.

**What it does.** When the cheap rename didn't move a row — `updateSlug` now returns the moved-row count in both engines, since a zero-row UPDATE never throws and the count is the only way to see it — and the destination demonstrably materialized, the stale old row is deleted. The reporting surface from #3473 (`SyncResult.renameFallbacks`) is gone; the engine signature change stays because the fix itself needs it.

**The two safety rails, from your #3252 review:**
- *Destination-materialized check*: reconcile runs only after `imported`, or an errorless skip **at the new slug**. An identity-dedup skip against the old row means nothing landed at the destination — deleting the old row there would destroy the only copy, so it doesn't.
- *Positive location only*: the stale row is found by `source_path = from` via `resolveSlugsByPaths`, never by the old-slug guess (after a collision, a path-derived fallback slug can name an unrelated, e.g. manually curated, row). No match → nothing is deleted.

**Failure semantics** (this is the part that grew beyond my first cut, and why): a failed reconcile delete records a `<rename:…>` **sentinel** in the failure ledger rather than an ordinary path failure, and the rename is not checkpointed. An ordinary path row would be chronic-skipped by the auto-skip valve after N attempts — advancing the bookmark and making a multi-run delete outage a *permanent* duplicate, which is the exact bug being fixed. The sentinel hard-blocks the gate (same machinery as `<head>`), only for runs where the delete actually failed (prior-run ledger rows don't block), and clears through the ordinary success path when the rename converges. The sentinel-blocked message is split so `<head>`'s force-push text doesn't fire for a rename failure. This failure mode was caught by an independent Codex review pass over the first version of this diff — I'd initially recorded it as an ordinary path failure.

**Known limitation, stated rather than hidden**: code-strategy imports don't populate `source_path`, so a collision fallback on a code rename reconciles nothing (safe direction — the old row is left rather than guessed at). Also flagged by the same review pass.

**Testing.** 7 tests in `test/sync-rename-reconcile.serial.test.ts`: engine contract (1/0), collision → old row deleted + destination carries the renamed content + page count converges to 1, dedup-skip → only copy survives, slug-guess → unrelated manual row survives, happy path keeps `page_id`, and an injected delete failure → `blocked_by_failures` with an open `<rename:…>` sentinel → next run converges and clears the sentinel. Ran 3x with no flakes. Broader regression (sync / pglite-engine / source-id-tx / sync-resumable-import / gbrain-home-isolation / sync-failure-ledger, 248 tests) green, `bun run typecheck` clean.

---

**Review follow-up (f601e91f).** Both blockers fixed, the three requested tests added, and four of the five smaller findings taken:

- The blocked-run message now prints the failing sentinel rows verbatim and documents the operator exit: `gbrain delete <stale-slug>`. Since that command is a **soft** delete, the reconcile and the orphan probe consider **active rows only** (`deleted_at IS NULL`, all rows — not a one-row resolve, so a soft-deleted row can't mask a live duplicate sharing the path) — that's what makes the remedy actually unwedge a permanently-failing DELETE. The remedy is scoped honestly: it covers environments where UPDATE works but the DELETE doesn't (DELETE-denying RLS, FK RESTRICT), not a fully read-only database.
- An open `<rename:…>` sentinel whose OLD path no longer has any active row self-clears — wired through the incremental gate, the full-sync gate (including post-purge, so `--force` converges in one run), and both `up_to_date` early returns (your exact probe shape). The OLD path is read back from the sentinel's own error text; the slug/path slots are JSON-encoded so the format is self-delimiting for any path bytes, and parsing is fail-closed — a misread leaves the row open, never clears it.
- The reconcile now removes **every** active row carrying the old path (your `resolveSlugsByPaths` ORDER BY finding, taken at the call site — the engine method is left as-is because the deletes loop needs to see soft-deleted rows so its slug-guess fallback never fires for them). `RENAME_RECONCILE` classification is envelope-first so a wrapped cause can't steal it, the full-sync sentinel block no longer misreports a rename failure as a history change, and a head+rename dual fire appends the rename detail instead of losing it.

**Scope gaps, stated:** (1) markdown pages predating the v0.32.7 `source_path` column have NULL there, so both the reconcile and the orphan probe silently no-op on those rows in older brains — same safe direction as the code-strategy gap above (the old row is left, never guessed at). (2) A sync that keeps exiting on a partial path (persistent pull failure / timeout) doesn't run the orphan sweep — the first healthy run does, and that failure mode is loudly surfaced on its own.


time-attack:
Strong work — **the destructive core is sound and I tried hard to break it.** I couldn't. Two availability regressions need fixing first, then this is a merge.

## What I verified on the data-loss question

I went looking for an input where the reconcile deletes the only copy, across eleven scenarios: destination-occupied collision, identity-dedup skip against the old row, slug-guess delete of an unrelated curated row, destination pre-existing with identical content, case-only rename, same-slug-different-source, two files to one destination, rename chains and swaps, concurrent syncs, crash between import and delete, and a hypothetical where `result.count` comes back undefined on postgres.js. **None of them loses data.**

Two structural properties do more work than the two rails you named: `importFromFile` always passes `sourcePath: relativePath` and `putPage` does `source_path = COALESCE(EXCLUDED.source_path, pages.source_path)`, so the row the import just wrote can never be found by `resolveSlugsByPaths([from])`. The `staleSlug !== newSlug` guard then catches the residual case. Worth noting the reconcile is strictly *more* conservative than the delete loop already on master at `sync.ts:2755`, which deletes from a path-derived guess.

Engine parity, JSONB, and multi-source scoping all check out — `resolveSlugsByPaths` is source-scoped in both engines with no slug fallback, and `deletePage` defaults `sourceId` identically in both.

Tests: 7 pass on your branch; reverting the 4 src files to master gives 3 pass / 4 fail, so they pin behavior. Regression sweep 208 + 108 pass / 0 fail. Typecheck clean.

## The two blockers, both reproduced

**1. A permanent reconcile failure wedges sync with no operator escape.** Five consecutive `gbrain sync --skip-failed` runs against an always-throwing `deletePage`:

```
run1..run5 (--skip-failed) = blocked_by_failures  [["<rename:people/dana.md>","open",1..5]]
```

That's structural: `decideGateAction` hard-blocks on any sentinel *even with* `--skip-failed`, and `acknowledgeFailures` explicitly skips sentinels (`sync-failure-ledger.ts:515`). Realistic triggers are RLS denying DELETE on a Supabase brain, an FK RESTRICT, or a read-only DB. The blocked message says only "The next 'gbrain sync' retries the reconcile from the same diff" — it names no remedy. **Please put the remedy in the message** (`gbrain delete <stale-slug>`, after which the reconcile no-ops and the sentinel clears). Turning a cosmetic duplicate row into a total sync outage needs a documented exit.

**2. An orphaned sentinel becomes a permanent, unfixable `doctor` FAIL.** After a force-push that invalidates the pinned target, the destination stops recurring in the diff and the row is never cleared:

```
run1 (--skip-failed) = synced      [["<rename:people/dana.md>","open"]]
run2 (--skip-failed) = up_to_date  [["<rename:people/dana.md>","open"]]
```

`decideSyncFailureSeverity` escalates `sync_failures` to FAIL once the oldest open row passes `failHours`, and no CLI can clear a sentinel — hand-editing `~/.gbrain/sync-failures.json` is the only fix. Master already treats this class as a bug for `<head>` and added an unconditional clear (`sync.ts:3350`). This needs the analogue: clear any open `<rename:X>` whose stale `source_path` row no longer resolves.

**In fairness to you:** I first thought this wedged every future sync. It doesn't — the banked `ckpt.target` pin makes the blocked run replay the same diff, so the normal retry path converges (my first probe hit `synced` on run 2). Only pin invalidation orphans the row.

## Smaller things, not blockers

- The full-sync gate (`sync.ts:3691`) wasn't updated: a rename sentinel there prints "repository history changed during sync", and `fullSucceeded` filters `isSkippablePath`, so `gbrain sync --force` — the usual operator reset — is blocked with a misleading message.
- If `<head>` and `<rename:…>` both fire, the head branch wins and the rename is invisible.
- `formatCodeBreakdown` doesn't classify the sentinel: blocked output is literally `UNKNOWN: 1` with no path (visible in your own test output).
- `resolveSlugsByPaths` has no `ORDER BY`, so with two rows sharing `source_path = from` only one arbitrary duplicate is removed — and the rename is then checkpointed, so the second never retries.
- **Undisclosed scope gap:** you disclosed that code pages never populate `source_path`, but markdown pages predating the v0.32.7 `source_path` column also have NULL, so this silently no-ops on older brains. Worth saying in the body.

Fix the two blockers and add tests for the two wedge scenarios plus the dedup-skip-at-`newSlug` case (safe, but it's the one path where `destMaterialized` is true without a write happening), and I'll take it.

time-attack:
Maintainer decision: **merging as-is.** Recording what that accepts, so it's findable if it ever bites.

**What's good, and I tested it hard.** I tried eleven scenarios to make this destroy data — destination-occupied collision, identity-dedup skip against the old row, slug-guess delete of an unrelated curated page, destination pre-existing with identical content, case-only rename, same-slug-different-source, two files renaming onto one destination, rename chains and swaps, concurrent syncs, crash between import and delete, and `result.count` coming back undefined on postgres.js. **None of them loses data.** Two structural properties do the heavy lifting: `importFromFile` always passes `sourcePath`, and `putPage` does `source_path = COALESCE(EXCLUDED.source_path, pages.source_path)`, so the row the import just wrote can never be found by `resolveSlugsByPaths([from])`; the `staleSlug !== newSlug` guard covers the residual. The reconcile is strictly more conservative than the delete loop already on master at `sync.ts:2755`.

Engine parity, JSONB, and multi-source scoping all check out. Tests: 7 pass; reverting the 4 src files to master gives 3 pass / 4 fail, so they pin behavior rather than scaffolding.

**Accepted risks, knowingly:**

1. **A permanently failing reconcile can wedge sync with no operator escape.** Verified across five consecutive `--skip-failed` runs: `blocked_by_failures` every time. `decideGateAction` hard-blocks on any sentinel even with `--skip-failed`, and `acknowledgeFailures` explicitly skips sentinels (`sync-failure-ledger.ts:515`). Realistic triggers are RLS denying DELETE on a Supabase brain, an FK RESTRICT, or a read-only DB. The blocked message doesn't name the remedy — which is `gbrain delete <stale-slug>`, after which the reconcile no-ops and the sentinel clears.
2. **An orphaned sentinel becomes a permanent doctor FAIL.** After a force-push invalidates the pinned target, the destination stops recurring in the diff and the row is never cleared. `decideSyncFailureSeverity` escalates to FAIL past `failHours`, and no CLI can clear a sentinel — hand-editing `~/.gbrain/sync-failures.json` is the only fix. Master already treats this class as a bug for `<head>` (unconditional clear at `sync.ts:3350`); the `<rename:X>` analogue doesn't exist yet.

Both are follow-ups worth filing rather than blockers, given the maintainer's call. Smaller residuals: the full-sync gate (`sync.ts:3691`) prints "repository history changed during sync" for a rename sentinel and `--force` can't clear it; `<head>` wins if both fire; `formatCodeBreakdown` renders the sentinel as a bare `UNKNOWN: 1`; `resolveSlugsByPaths` has no `ORDER BY` so with two rows sharing `source_path` only one arbitrary duplicate is removed; and the fix silently no-ops on brains predating the v0.32.7 `source_path` column (not just on code pages, which was the disclosed part).

Thanks for the iterations on this one — the two safety rails you added after the earlier review are what made it safe enough to take.

Masashi-Ono0611:
Thank you for the careful review rounds and for taking this — the sentinel + gate design ended up much stronger because of them, and I'm really glad the safety rails earned the merge. The two blockers you tracked in #3570 were actually in flight as review fixes when the merge landed, so I've rebased them onto current master as #3583 (CI green, 22/22). Happy to adjust anything there.

Issue #3417, opened 2026-07-26
slugifySegment strips Hebrew (and other non-CJK scripts) → mass slug collisions, silent last-writer-wins data loss, permanent re-import churn


amitmai:
## Summary

`slugifySegment()` strips Hebrew characters entirely (they're not in `SLUGIFY_KEEP_RE`), so every Hebrew-titled file collapses to a slug built only from its Latin/digit remnants — usually just the directory prefix or a bare number. On a Hebrew-heavy vault this causes **mass slug collisions → silent last-writer-wins data loss**, plus **permanent re-import churn**.

This is the same bug class CJK had before v0.32.7 added the CJK ranges to the slug grammar; Hebrew (and other non-CJK, non-Latin scripts: Arabic, Cyrillic, Greek, …) never got the equivalent fix.

## Observed on

- gbrain 0.42.58.0, PGLite engine, `gbrain import <dir>` over a personal vault of 7,149 markdown files (mixed Hebrew/English)

## Impact measured

- **59 collision groups covering 697 files.** Worst group: 426 Hebrew-titled notes in one folder (`notes/פגישה עם דני.md`, `notes/רשימת קניות.md`, …) all slugify to the identical slug `notes` — one page slot for 426 documents.
- **Silent data loss:** one slug = one page, so only the last-imported file of each group is retrievable; the other ~638 documents are invisible to search while `import` reports `0 errors`.
- **Permanent churn:** every subsequent import re-imports exactly those 697 files (each group ping-pongs over its shared slot), so imports never converge to a no-op — on this vault every pass re-embeds ~700 pages (~160 s + wasted embedding work per run, forever).

## Repro

```bash
mkdir -p /tmp/brain-repro/notes
echo "shopping" > "/tmp/brain-repro/notes/רשימת קניות.md"
echo "meeting"  > "/tmp/brain-repro/notes/פגישה עם דני.md"
gbrain import /tmp/brain-repro   # both files → slug "notes"; second overwrites first
gbrain import /tmp/brain-repro   # reports 1 imported again (ping-pong), not 0
```

## Root cause

`src/core/sync.ts`:

```ts
const SLUGIFY_KEEP_RE = new RegExp(`[^a-z0-9.\\s_\\-${CJK_SLUG_CHARS}]`, 'g');
```

Hebrew (U+0590–U+05FF) isn't in the keep class, so `.replace(SLUGIFY_KEEP_RE, '')` deletes the entire title.

## Suggested fix (with one caveat)

[...]

PR #3481, opened 2026-07-28
v0.42.68.0 feat(ai): default to openai:text-embedding-3-small@1280 ahead of the ZeroEntropy sunset (#3390)

### LINKED ISSUE #3390: ZeroEntropy (the default embedding provider) sunsets 2026-09-04 — default-config brains have no forward migration path

ZeroEntropy [announced on 2026-07-24](https://zeroentropy.dev/articles/zeroentropy-is-joining-notion) that it has been acquired by Notion. All ZeroEntropy products — including the hosted `/models/embed` and `/models/rerank` endpoints — **sunset on 2026-09-04**. Weights are now Apache-2.0, so self-hosting is possible, but the hosted API goes away.

That is gbrain's shipped default:

- `src/core/ai/defaults.ts` — `DEFAULT_EMBEDDING_MODEL = 'zeroentropyai:zembed-1'`, `DEFAULT_EMBEDDING_DIMENSIONS = 1280`
- `README.md` — ZeroEntropy listed as the default embedding provider, `zerank-2` as the default reranker in `balanced` / `tokenmax`
- `INSTALL_FOR_AGENTS.md` — tells installing agents to obtain a `ZEROENTROPY_API_KEY` as the default stack
- `docs/tutorials/personal-brain.md` — recommends it as "about 2× faster than OpenAI and about 2.6× cheaper"
- v0.36.2.0 shipped the switch onto ZeroEntropy as the default

I couldn't find an existing issue or commit tracking this, so filing it — the date is ~6 weeks out.

**What breaks, precisely:** it's worse than "new pages stop embedding." The query-embedding path uses the same endpoint, so on 2026-09-04 any brain still on `zeroentropyai:*` loses **semantic retrieval entirely** — existing vectors stay in the DB but become unqueryable, because you can't embed the query to search against them. Hybrid search silently falls back to the lexical arm.

**Why a config change isn't enough: 1280 is only reachable from the sunsetting provider.**

I enumerated `default_dims` / `dims_options` across all 16 recipes in `src/core/ai/recipes/` at `3fafb69`. `zeroentropyai` is the only one that offers 1280:

| recipe | dims |
|---|---|
| openai / azure-openai | 256, 512, 768, 1024, 1536, 3072 |
| google | 768, 1536, 3072 |
| openrouter | 512, 768, 1024, 1536 |
| zhipu | 256, 512, 1024, 2048 |
| nvidia | 1024, 2048, 4096 |
| dashscope | 64 … 1024 |
| voyage / mistral / perplexity | 1024 default |
| minimax | 1536 default |
| ollama / llama-server / litellm-proxy | explicit `--embedding-dimensions` required |
| **zeroentropyai** | **2560 default; 1280/640/320/…** |

So a brain on the shipped default (`zembed-1` @ 1280) cannot do a same-dimension swap to any hosted alternative. **Every default-config brain needs a schema dimension transition**, and `embedding_dimensions` is deliberately rejected by `config set` as a schema-sizing field (`src/core/embed-preflight.ts`).

**And the existing lever can't reach these brains.** gbrain already implements this migration — `src/commands/ze-switch.ts` plus `src/core/retrieval-upgrade-planner.ts` / `-prompt.ts`, including `runSchemaTransition()`, which does the non-destructive dimension change on a populated brain. That's the hard part, already written and shipped. But the only entry point is the *reverse* of the v0.36 switch, and it requires a stored snapshot (`retrieval-upgrade-planner.ts:495-503`):

```ts
export async function undoRetrievalUpgrade(engine: BrainEngine): Promise<...> {
  const snapshotStr = await engine.getConfig(KEY_PREVIOUS_SNAPSHOT);
  if (!snapshotStr) {
    return { status: 'no_snapshot' };
  }
```

A brain that *migrated onto* ZeroEntropy has `ze_switch_previous_snapshot` and can `--undo`. A brain **initialized on the ZeroEntropy default has no snapshot** and gets `no_snapshot` — leaving only `gbrain init --force` (destructive) or hand-written SQL. Mine is in that category: 0 rows matching `key LIKE 'ze_switch%'`, on a 23k-page / 65k-chunk Postgres brain.

The command the codebase already tells users to run doesn't exist. Both of these point at it:

- `README.md:312` — "Run `gbrain doctor`. It will print the exact `gbrain config set ...` or `gbrain retrieval-upgrade` command to repair the mismatch."
- `src/commands/doctor.ts:5915` — `//   - non-empty brain → \`gbrain retrieval-upgrade --to … --reindex\``

But `retrieval-upgrade` is not in `CLI_ONLY` (`src/cli.ts:58`) and has no dispatch branch; grepping `src/` for the string literal `'retrieval-upgrade'` returns nothing. So the documented repair path for a non-empty brain with a dimension mismatch is currently unreachable.

Since `runSchemaTransition()` and the planner already exist, the ask is small — a generalized forward version of `ze-switch`, not keyed to a stored snapshot:

```
gbrain retrieval-upgrade --to <provider:model> --dimensions <N> [--reindex] [--dry-run] [--json]
```

That would also make good on what README and doctor already promise. Worth pairing with a default flip in `defaults.ts` before 2026-09-04 so new installs don't land on a dead provider.

**Alternatives considered**

- **Self-host `zembed-1`** via `llama-server` / `litellm-proxy` / `ollama`, which accept explicit dims — this is the one path that preserves existing vectors at 1280, and the weights are now Apache-2.0. Real option, but it means standing up GPU inference, a large cost and ops step for personal-brain users who chose ZeroEntropy precisely because it was cheap and hosted.
- **`gbrain init --force --embedding-model … --embedding-dimensions …`** — what `doctor.ts` recommends for an *empty* brain. Destructive on a populated one.
- **Manual `ALTER TABLE content_chunks ALTER COLUMN embedding TYPE vector(N)` + HNSW rebuild + full re-embed** — what I'd otherwise have to do by hand, duplicating `runSchemaTransition()` badly and without the planner's cost preview or resume support.
- **Wait for a hosted drop-in at 1280** — none exists in the recipe set today.

Happy to send a PR for the `retrieval-upgrade` command if that's useful — say the word and I'll follow the contributing guide.

time-attack:
Closes #3390.

## Why

ZeroEntropy's hosted API shuts down **2026-09-04** (~5 weeks out). `zeroentropyai:zembed-1` has been `DEFAULT_EMBEDDING_MODEL` since v0.36.2.0, so every brain that never picked a model explicitly was going to lose semantic retrieval on that date — and not just for new writes. Query embedding runs through the same `/models/embed` endpoint, so existing vectors become unqueryable too.

The maintainer decision: default to **`openai:text-embedding-3-small` at 1280 dimensions**.

Per the new **Default-provider policy** added to `CLAUDE.md` in this PR: a gbrain DEFAULT embedding or reranking model must be either open-weight, or from the vendor with the longest proven model-lifetime record. Novel/startup providers may ship as opt-in recipes, never as the default. The v0.36 `zembed-1` default stranded every default-config brain when ZeroEntropy was acquired and gave ~6 weeks notice; the policy exists so that cannot repeat.

## The no-schema-change property

**1280, not 1536, is load-bearing.** OpenAI's `text-embedding-3-*` family is Matryoshka, and `isValidOpenAITextEmbedding3Dim` (`src/core/ai/dims.ts:87`) accepts **any integer width from 1 up to the model's native size** (1536 for `-small`) — not a discrete step list the way ZeroEntropy and Voyage have.

Consequence: a brain created under the previous 1280-wide ZeroEntropy default keeps its existing `vector(1280)` column **and its HNSW index**. `applyEmbeddingMigration` only calls `runSchemaTransition` when `col.dims !== plan.to_dims`, so migrating at the same width rebuilds the **vectors only** — no dimension transition, no `ALTER`, no index rebuild, no `reinit-pglite`.

Verified by probing `resolveSchemaEmbeddingDim` directly rather than taking the claim on faith, which is how the first of the two blockers below surfaced.

## Migration for existing brains

`gbrain migrate embeddings` (shipped in #3459) is the forward path:

```bash
gbrain migrate embeddings --to openai:text-embedding-3-small --dim 1280 --dry-run
gbrain migrate embeddings --to openai:text-embedding-3-small --dim 1280
```

Resumable — a killed run re-runs the same command. `--dim` matters: without it the target resolves to the openai recipe's `default_dims` of 1536 and forces a needless dimension transition. The upgrade banner computes `--dim` from the brain's actual current width, so users get a correct paste-ready command rather than a generic one.

Staying on `zembed-1` self-hosted (weights are Apache-2.0, via llama-server or Ollama) also works and preserves existing vectors outright — no migration at all.

## Two blockers found and fixed

The default swap did not work as a one-line constant change. Both of these would have shipped a default that the codebase itself rejects:

1. **The openai recipe's `dims_options` omitted 1280.** That list is Tier 1 in `isCustomDimValidForProvider` and wins over the Tier-2 `isValidOpenAITextEmbedding3Dim` range check, so `resolveSchemaEmbeddingDim({ embedding_model: 'openai:text-embedding-3-small', embedding_dimensions: 1280 })` returned `{ ok: false, error: 'rejects custom dimensions 1280' }`. `gbrain init` would have refused its own shipped default. Pinned by a new regression test.

2. **The recipe listed `text-embedding-3-large` first.** `init`'s `resolveEmbeddingByEnv` picks `touchpoints.embedding.models[0]` and only adopts `DEFAULT_EMBEDDING_DIMENSIONS` when that equals the canonical default. A fresh `OPENAI_API_KEY`-only install would have landed on `3-large@1536` and the declared default would have been unreachable in practice. Reordered so "the declared default" and "what a fresh install actually gets" are the same thing. Also pinned.

## Upgrade banner

The `ze_sunset_notice_shown` scaffold from #3459 now:
- names the concrete target (`openai:text-embedding-3-small`) instead of `<provider:model>`,
- passes `--dim <current width>` and explains that the same width reuses the existing column and HNSW index,
- states the **2026-09-04** date,
- reads the **DB config plane** as well as the file plane. This one matters: with the default no longer a ZE model, `cfgSchema.embedding_model ?? DEFAULT_EMBEDDING_MODEL` would silently stop detecting brains created under the ZE default that never wrote `embedding_model` to `~/.gbrain/config.json`. Those brains do carry the DB-plane row seeded by `initSchema`, so the banner still fires for them.

## Deliberately unchanged

The **ZeroEntropy reranker default** (`tokenmax` mode) is untouched — no replacement has been chosen, and that is a separate product decision. Its sunset is called out in the banner, the README, and `docs/ai-providers/zeroentropy.md`. The `zerank` weights are Apache-2.0, so the existing `llama-server-reranker` recipe is a working local path.

## Also in here

Fixes a **pre-existing** ambient-environment flake in `test/e2e/fresh-install-pglite.test.ts`: it cleared a hardcoded pair of provider env keys, so it failed on any machine with a third provider key set (reproduced on unmodified `origin/master` — `Multiple embedding providers env-ready: minimax, zeroentropyai`). It now clears every non-OpenAI embedding provider key enumerated from the recipe registry, so the file no longer passes or fails based on the developer's shell.

## Docs updated to current truth

`CLAUDE.md` (new policy section), `README.md`, `INSTALL_FOR_AGENTS.md`, `docs/integrations/embedding-providers.md`, `docs/guides/embedding-migration.md`, `docs/embedding-migrations.md`, `docs/ai-providers/zeroentropy.md` (deprecation banner), both tutorials, `gbrain init`'s no-provider hint, and `gbrain advisor`'s missing-embedding-key smell (was telling users to set `zeroentropy_api_key`).

## Receipts

All redirected to files with real exit codes — no piping through `tail`.

| Gate | Result |
|---|---|
| `bun run typecheck` | exit 0 |
| `bun test` — touched unit files (8 files) | 173 pass / 0 fail |
| `bun test` — ZE-adjacent sweep (14 files) | 281 pass / 0 fail |
| `bun test test/e2e/fresh-install-pglite.test.ts` | 2 pass / 0 fail |
| `bun test test/e2e/init-fresh-pglite.test.ts` | 14 pass / 0 fail |
| `bun run build:llms` + `bun test test/build-llms.test.ts` | exit 0; 12 pass / 0 fail |
| `scripts/check-key-files-current-state.sh` | exit 0 (`reference docs history-free; CLAUDE.md within cap`) |
| `scripts/check-no-tracked-symlinks.sh` | exit 0 (no tracked symlinks) |
| Version trio audit | `VERSION` / `package.json` / `CHANGELOG.md` all `0.42.68.0` |

`bun.lock` needed no change — it does not pin the root package version.

## New tests

`test/ai/dims-openai.test.ts` gains a `default embedding config (v0.42.68.0 #3390)` block, all assertions derived from the live constants rather than hardcoded literals:

- `DEFAULT_EMBEDDING_MODEL` / `DEFAULT_EMBEDDING_DIMENSIONS` are `openai:text-embedding-3-small` / `1280`
- the default width is a valid Matryoshka width for the default model (the no-schema-change property)
- `dimsProviderOptions` passes 1280 through to the wire
- `resolveSchemaEmbeddingDim` **accepts** the shipped default config (fails if 1280 leaves `dims_options`)
- `init` env-detection lands on the default model when `OPENAI_API_KEY` is the only key (fails if the recipe model order regresses)

Updated tests that pinned the ZE default: `test/v0_37_fix_wave.serial.test.ts`, `test/ai/gateway.test.ts`, `test/ai/schema-templating.test.ts`, `test/e2e/fresh-install-pglite.test.ts`, `test/e2e/init-fresh-pglite.test.ts`, `test/e2e/embedding-column-pglite.test.ts` (its provenance guard now asserts against the live constant so it survives future default swaps).

**Not merged** — awaiting maintainer review.


time-attack:
**HOLD — do not merge as-is.** The multilingual eval built in #3486 contradicts this PR's core choice.

Measured on real Wikipedia corpora (80 articles × 4 languages, same topics so script is the only variable), vector-arm-only, nDCG@10:

| model | en | he | zh | ja |
|---|---|---|---|---|
| `3-small@1280` (this PR's default) | 0.929 | **0.625** | 0.902 | 0.833 |
| `bge-m3@1024` (open weights, local) | 0.915 | **0.908** | 0.950 | 0.888 |

`text-embedding-3-small` collapses 30 points on Hebrew and lands **below the repo's own 0.65 ship-it threshold** (0.349 on author-written Hebrew questions; 0.596 on the in-script robustness group). A 0.6B open-weight model running on a laptop beats it by 21–29 points, surviving a factor-60 Bonferroni correction. gbrain has confirmed Hebrew users (#3417) and CJK users (#1637).

What this PR got RIGHT and should keep: **1280 vs 1536 is within noise on all 12 comparisons** — the truncation is free, so 1280 is the correct width, saving 17% of vector storage. The two validation bugs this PR found (1280 missing from `dims_options`, recipe model order making the declared default unreachable) are real and worth keeping regardless of which model wins.

Also note: this PR's own new CLAUDE.md policy prefers open-weight defaults, and `bge-m3` satisfies it where `3-small` only satisfies the vendor-longevity clause.

Recommend: keep the plumbing fixes, the banner, the policy, and the 1280 width; revisit the model choice against #3486's data before merging.

time-attack:
## The eval is done, and it overturns this PR's model choice

I ran the multilingual matrix this PR was waiting on: 8 candidates × 6 language slices × 3 reps, ~125k per-query records. **`openai:text-embedding-3-small` is the worst model tested on Hebrew** — the exact language that motivated the request.

Vector nDCG@10, mean of 3 reps:

| Model | en | **he** | ar | ja | ru | zh | store |
|---|---|---|---|---|---|---|---|
| `gemini-embedding-001@768` | 0.945 | **0.964** | 0.987 | 0.953 | 0.971 | 0.980 | **3.6MB** |
| `gemini-embedding-001@1536` | 0.947 | 0.965 | 0.987 | 0.955 | 0.971 | 0.980 | 7.2MB |
| `gemini-embedding-2@1536` | 0.954 | 0.957 | 0.992 | 0.968 | 0.972 | 0.981 | 7.2MB |
| `openai:text-embedding-3-large@1536` | 0.940 | 0.812 | 0.940 | 0.925 | 0.945 | 0.942 | 7.2MB |
| `bge-m3@1024` (open weights) | 0.888 | 0.901 | 0.954 | 0.893 | 0.919 | 0.949 | 4.8MB |
| `qwen3-embedding:0.6b@1024` (open weights) | 0.879 | 0.829 | 0.899 | 0.862 | 0.906 | 0.906 | 4.8MB |
| **`3-small@1280`** ← this PR | 0.907 | **0.645** | 0.864 | 0.833 | 0.884 | 0.905 | 6.0MB |
| `3-small@1536` | 0.906 | 0.645 | 0.867 | 0.837 | 0.890 | 0.906 | 7.2MB |

Hebrew 0.645 → 0.964 is +0.32 nDCG@10, roughly +50% relative, `p_adj < 0.0001` across 72 Bonferroni-corrected comparisons. Every language improves significantly.

**What this PR got right and I'm keeping:** 1280 vs 1536 is free (0.645 vs 0.645 on he) — the no-schema-change property holds, and the same is true for Gemini at 768 vs 1536 (0.964 vs 0.965). So a narrower default costs nothing but storage savings.

## But the policy this PR adds argues against Gemini

This PR added the **Default-provider policy** to `CLAUDE.md`: *a default must be open-weight, or from the vendor with the longest proven model-lifetime record.*

Google is not obviously that vendor. **They just retired `gemini-1.5-pro`** — I filed #3510 today because it's still the hardcoded default at four sites and the recipe allowlist masks it from the guard test. That is precisely the vendor-sunset failure mode the policy exists to prevent, and it happened while this PR was open.

So the honest framing is a genuine trade-off, not a winner:

| Option | Hebrew | Policy | Sunset risk |
|---|---|---|---|
| `gemini-embedding-001@768` | **0.964** | fails the open-weight test; Google's record is now questionable | real — see #3510 |
| `bge-m3@1024` | 0.901 | **passes** — open weights, cannot be sunset | **none** |
| `3-small@1280` (this PR) | 0.645 | passes | low |

`bge-m3` recovers most of the Hebrew gap (0.645 → 0.901) with **zero** sunset risk, and it's the only option that satisfies the policy this PR wrote. It costs throughput: 7.2 docs/s vs Gemini's 61, so a 10k-page backfill is ~23 min instead of ~3.

## Three caveats I won't bury

1. **Gemini was not pre-registered.** The harness's pre-registered pairs covered only the OpenAI/Ollama set. I computed the Gemini bootstrap myself from the saved per-query records using the harness's own method (paired on qid, 10k resamples, per-query mean across reps) — so these are **post-hoc** comparisons. Weaker evidence than pre-registered, even at these effect sizes. `docs/eval/SEARCH_MODE_METHODOLOGY.md` cares about this distinction.
2. **Voyage was never tested** — `VOYAGE_API_KEY` isn't set. That arm is a hole in the matrix, not a finding.
3. **Two hard prerequisites before Gemini could ship as a default**, both filed: `GEMINI_API_KEY` is never read anywhere (only `GOOGLE_GENERATIVE_AI_API_KEY`, and neither is folded into gateway env — #3500), and `google:` has **no entry in `embedding-pricing.ts`**, so cost tracking and budget gates would be blind to it.

## What I'm doing with this PR

Holding it rather than force-landing either choice. Changing the default embedding model affects every fresh install and every brain with `embedding_model` unset, and I hold `push`/`triage` but not `admin` on this repo — that's an owner call, same as #3492.

@garrytan the decision I'd like from you: **`bge-m3` (policy-compliant, no sunset risk, 0.901) or `gemini-embedding-001@768` (best quality, 0.964, but needs #3500 first and Google just sunset a model)?** I'll rewrite this PR to whichever you pick. If you want the raw data, the receipt + all per-query records are in the eval JSONL and I can attach the post-hoc significance table.

**On urgency:** ZeroEntropy shuts down 2026-09-04, 38 days out. But this PR only helps fresh installs and brains with `embedding_model` unset — brains that have `zeroentropyai:zembed-1` written into config (including everything migrated by the `ze_switch` flow) need `gbrain migrate embeddings` regardless of what the default says. That migration outreach is the part with a real deadline; the default change is a ~1-day job.

time-attack:
Superseded by #3611. Maintainer decision on the embedding default landed on `ollama:bge-m3` @1024 (open-weight — satisfies the Default-provider policy this PR introduced with no vendor sunset risk at all), with `openai:text-embedding-3-small` as the loud hosted fallback when Ollama isn't available. #3611 carries this PR's policy section, the `gbrain upgrade` ZE-banner DB-plane fix, and the doc sweep forward, re-targeted — thank you @time-attack, the groundwork here shaped the final change. Closing in favor of #3611.

PR #3482, opened 2026-07-28
fix(test): restore extended local E2E reliability

cybernaut6404:
## Summary
- repair seven extended E2E contracts exposed by the complete local gate
- reset shared source state between database-backed E2E files
- make Postgres/PGLite representative-path selection deterministic
- preserve the guarded test-database reset opt-in inside Docker E2E runs

## Evidence
- targeted seven-file set: 77 passed, 0 failed
- contamination sequence: 32 passed, 0 failed
- engine parity: 26 passed, 148 assertions, 0 failed
- complete default one-worker Docker gate: 164 files, 1,113 tests, 0 failed
- current upstream verification: 32/32 checks passed

Follow-up to closed #3467, split as requested. This PR contains no private brain page slugs, local evidence notes, node_modules cleanup, or unrelated artifact-sync changes.

### DIFF
--- a/scripts/ci-local.sh
+++ b/scripts/ci-local.sh
@@ -262,12 +262,14 @@ printf '%s\\n' 1 2 3 4 | xargs -P4 -I{} sh -c '
   echo \"[shard \${shard}] e2e phase (SHARD=\${shard}/4, DATABASE_URL=postgres-\${shard})\" >> \$log
   if [ -s /tmp/e2e-selected.txt ]; then
     SHARD=\${shard}/4 \\
+    GBRAIN_TEST_DB=1 \\
     DATABASE_URL=postgresql://postgres:postgres@postgres-\${shard}:5432/gbrain_test \\
     GBRAIN_PGBOUNCER_URL=postgresql://postgres:postgres@pgbouncer:5432/gbrain_pgbouncer \\
     GBRAIN_PGBOUNCER_DIRECT_URL=postgresql://postgres:postgres@postgres-1:5432/gbrain_test \\
     xargs -a /tmp/e2e-selected.txt bash scripts/run-e2e.sh >> \$log 2>&1
   else
     SHARD=\${shard}/4 \\
+    GBRAIN_TEST_DB=1 \\
     DATABASE_URL=postgresql://postgres:postgres@postgres-\${shard}:5432/gbrain_test \\
     GBRAIN_PGBOUNCER_URL=postgresql://postgres:postgres@pgbouncer:5432/gbrain_pgbouncer \\
     GBRAIN_PGBOUNCER_DIRECT_URL=postgresql://postgres:postgres@postgres-1:5432/gbrain_test \\
--- a/scripts/run-e2e.sh
+++ b/scripts/run-e2e.sh
@@ -80,6 +80,9 @@ mkdir -p "$E2E_TMP_HOME/.gbrain"
 for _e2e_var in $(env | grep -oE '^(CONDUCTOR_|MCP_|OPENCLAW_|GBRAIN_)[A-Za-z0-9_]*' | sort -u); do
   case "$_e2e_var" in
     GBRAIN_HOME) ;;  # required for HOME isolation (set above) — keep
+    # Narrow test-only safety opt-in. Schema drift still requires a test-shaped
+    # database name before it will reset a non-local Docker service host.
+    GBRAIN_TEST_DB) ;;
     *) unset "$_e2e_var" || true ;;
   esac
 done
--- a/src/core/cross-modal-eval/aggregate.ts
+++ b/src/core/cross-modal-eval/aggregate.ts
@@ -107,8 +107,11 @@ export function aggregate(input: AggregateInput): AggregateResult {
     const mean = scores.reduce((a, b) => a + b, 0) / scores.length;
     const min = Math.min(...scores);
     const roll: DimensionRoll = { mean: round1(mean), min, scores };
-    if (roll.mean < PASS_MEAN_THRESHOLD) roll.failReason = 'mean_below_7';
-    else if (roll.min < PASS_FLOOR_THRESHOLD) roll.failReason = 'min_below_5';
+    // The floor is the stronger failure: when a score breaches both rules,
+    // report the individual-model collapse rather than hiding it behind the
+    // aggregate mean.
+    if (roll.min < PASS_FLOOR_THRESHOLD) roll.failReason = 'min_below_5';
+    else if (roll.mean < PASS_MEAN_THRESHOLD) roll.failReason = 'mean_below_7';
     dimensions[dim] = roll;
   }
 
--- a/src/core/cross-modal-eval/runner.ts
+++ b/src/core/cross-modal-eval/runner.ts
@@ -79,6 +79,8 @@ export interface RunEvalOpts {
   abortSignal?: AbortSignal;
   /** Stderr progress callback (cycle 1/3, slot A done, etc.). */
   onProgress?: (event: ProgressEvent) => void;
+  /** Test seam: alternative chat function; production defaults to the gateway. */
+  _chat?: typeof gwChat;
 }
 
 export type ProgressEvent =
@@ -142,6 +144,7 @@ export async function runEval(opts: RunEvalOpts): Promise<RunEvalResult> {
       abortSignal: opts.abortSignal,
       cycle,
       onProgress: opts.onProgress,
+      chat: opts._chat ?? gwChat,
     });
 
     const agg = aggregate({ slots: slotResults });
@@ -199,6 +202,7 @@ interface OneCycleOpts {
   abortSignal?: AbortSignal;
   cycle: number;
   onProgress?: (event: ProgressEvent) => void;
+  chat: typeof gwChat;
 }
 
 async function runOneCycle(opts: OneCycleOpts): Promise<SlotResult[]> {
@@ -226,7 +230,7 @@ async function callSlot(
     const messages: ChatMessage[] = [
       { role: 'user', content: prompt },
     ];
-    const result = await gwChat({
+    const result = await opts.chat({
       model: slot.model,
       system: SYSTEM_PROMPT,
       messages,
--- a/src/core/pglite-engine.ts
+++ b/src/core/pglite-engine.ts
@@ -3310,7 +3310,8 @@ export class PGLiteEngine implements BrainEngine {
              array_agg(DISTINCT n.last_link_type)
                FILTER (WHERE n.last_link_type IS NOT NULL) AS via_link_types,
              (array_agg(array_to_string(n.path, chr(9))
-               ORDER BY n.depth ASC, array_length(n.path, 1) ASC))[1] AS path_str,
+               ORDER BY n.depth ASC, array_length(n.path, 1) ASC,
+                        array_to_string(n.path, chr(9)) ASC))[1] AS path_str,
              (SELECT cc.id FROM content_chunks cc
                WHERE cc.page_id = n.id ORDER BY cc.chunk_index ASC LIMIT 1) AS canonical_chunk_id
       FROM walk n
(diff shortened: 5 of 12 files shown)


time-attack:
Nearly there — **11 of 12 files are justified and I want them.** One hunk needs to come out first, and then this goes in.

## Please drop the `src/core/cross-modal-eval/aggregate.ts` hunk

You flip master's `:110-111` so `min_below_5` is checked before `mean_below_7`. Master's test does genuinely fail (`Expected "min_below_5", Received "mean_below_7"` at `test/e2e/cross-modal-eval.test.ts:153`) — but the cause is the **stale model string**, not the ordering. `src/core/cross-modal-eval/runner.ts:52` is already `openai:gpt-5.2` on master while the test's stub keyed on `openai:gpt-4o`, so slot A fell through the stub and dragged the mean under 7 as well as the min under 5.

I tested this directly: in your branch I reverted `aggregate.ts` to master's ordering and kept every other change → **12 pass / 0 fail**. So the reorder isn't required by any test. It also changes user-visible output (`aggregate.ts:157` branches on `'mean_below_7'` to build the verdict message), has no coverage of the both-rules-breached case it changes, and leaves the parallel implementation at `src/core/takes-quality-eval/aggregate.ts:136-137` on the old ordering — introducing drift between two tables that were previously identical. It also collides with #3491, which is a dedicated `aggregate.ts` PR.

## What I verified and want to keep

**The two engine edits are a genuine fix, not test-bending** — which is what I went in suspicious of. `pglite-engine.ts:3310-3312` / `postgres-engine.ts:3459-3461` add `array_to_string(n.path, chr(9)) ASC` as a final tiebreaker in `relationalFanout`. `(depth, path_length)` is not a total order, so the representative path was plan-dependent — real product nondeterminism in `gbrain graph` and relational answers, fixed at the root, in both engines in lockstep. `test/relational-{ab,fanout,intent,recall}.test.ts` → 30 pass / 0 fail.

**The `_chat` DI seam is right.** `mock.module` is process-global in Bun and leaks into every file sharing the shard — that's the actual flakiness source — and the identical `_chat?: typeof gatewayChat` seam already exists in `src/core/cycle/extract-atoms.ts`. Established pattern, not invented.

**Test-weakening audit: clean.** I grepped every added line for `skip`/`skipIf`/`.todo`/`timeout`/`retry`. The only hit is the slug string `'notes/skip'`. No timeouts raised, nothing disabled. Both changed assertions check out against production: notes really are extractable (`gbrain-base.yaml:217-224` + `unionExtractableTypes`), and `facts.embedding` really is HALFVEC on pgvector ≥ 0.7 (`migrate.ts:2317`), so both old assertions were stale rather than protective.

Master fails cross-modal Q2, cross-modal Q3, and dream-cycle phase order (14 pass / 3 fail); your branch fixes all three (17 / 0). Those are real pre-existing breakages and nobody else has fixed them.

## Two merge-order heads-ups

- #3501 touches `runner.ts` and `test/e2e/cross-modal-eval.test.ts` and will change slot models again, re-staling the model strings you just fixed.
- Dropping the `aggregate.ts` hunk removes your conflict with #3491 entirely.

One nit for whenever you next touch it: `test/e2e/serve-http-oauth.test.ts:578` — house style pairs `sql.array()` with an explicit `::text[]` (see `test/e2e/embed-stale-pagination.test.ts:59,86`); the bare `'{read}'` works but the cast form matches the rest of the repo.

Drop that one hunk and ping me — the rest is good work on files that had genuinely rotted, and the reason they rotted is that PR CI has no Postgres E2E job, so none of `test/e2e/*` is validated by green checks.

time-attack:

Six of the seven repairs are genuine and verified (the two DB-free e2e files fail 3 tests on master and pass 9/9 on the PR branch; the Postgres-side test edits all match current production behavior structurally: gbrain-base declares note extractable, facts.embedding is halfvec on pgvector>=0.7, helpers re-seed matches schema.sql verbatim, GBRAIN_TEST_DB only relaxes the localhost requirement behind the test-shaped-dbname hard floor, and the engine ORDER BY tie-break lands identically in both engines). The blocker is src/core/cross-modal-eval/aggregate.ts:113-114: the failReason priority swap is an unforced production behavior change — reverting it leaves every test in the PR green (verified by discrimination run), no repaired test needs it, and it de-synchronizes the sibling takes-quality-eval/aggregate.ts:136-137 which keeps the old mean-first ordering, so the two eval surfaces now report different failReasons for identical both-breach scores. Drop that hunk, or keep it with a both-breach discriminating fixture (e.g. [8,8,4] -> assert min_below_5) and mirror the sibling. Title lacks the version-first prefix and there is no VERSION/CHANGELOG bump, but recently merged fix(test) PRs (#3566) show the maintainer batches those, so flagged as convention notes only.

### Discrimination test

Ran twice, both directions. (1) Old-behavior stub-back: on the PR branch, reverted src/core/cross-modal-eval/aggregate.ts to the base commit and re-ran test/e2e/cross-modal-eval.test.ts + test/cross-modal-eval-aggregate.test.ts — 12/12 still passed, proving the aggregate reordering hunk has no discriminating test (blocker). (2) Fail-on-master: checked out origin/master and ran test/e2e/cross-modal-eval.test.ts + test/e2e/dream-cycle-phase-order-pglite.test.ts — 3 tests fail (Q2 min-floor, Q3 inconclusive, dream-cycle phase order) exactly as the PR claims; same files pass 9/9 on the PR branch, so the test-repair hunks DO discriminate.

**Mergeability at review time:** mergeable=MERGEABLE, mergeStateStatus=CLEAN (initial poll returned UNKNOWN/UNKNOWN; re-poll settled CLEAN). Base fd8be831 is 32 commits behind master but master-side changes to the shared files (run-e2e.sh, both engines) do not overlap the PR hunks.

### Blockers

**1.** src/core/cross-modal-eval/aggregate.ts:113-114 — failReason ordering swap is a production behavior change with zero discriminating coverage: with the old ordering stubbed back in, all 12 cross-modal tests (e2e + unit) still pass; the Q2 fixtures ([9,8,4], mean exactly 7.0) never exercise the both-breach path. Either drop the hunk or add a both-breach fixture (scores [8,8,4]: mean 6.7<7 AND min 4<5, assert min_below_5).

**2.** src/core/takes-quality-eval/aggregate.ts:136-137 — sibling aggregator keeps the OLD mean-first ordering; if the cross-modal reordering is kept, the two eval surfaces report different failReasons for identical both-breach inputs. Mirror the change (with its test) or drop the cross-modal hunk.

### Not verified

- Postgres/Docker-backed runs of test/e2e/extract-atoms-discovery-sql.test.ts, phantom-redirect.test.ts, serve-http-oauth.test.ts, engine-parity.test.ts, schema-drift.test.ts — skipped for execution-footprint reasons; verified structurally instead (pack default, halfvec column creation, scopes TEXT[], guard logic)
- PR body's Docker-gate numbers (164 files / 1,113 tests / 0 failed; 77-test seven-file set; 26-test engine parity) — only the two DB-free files (9 tests) reproduced locally
- The original failure mode of sql.array(['read']) in serve-http-oauth.test.ts — the '{read}' literal replacement is a valid TEXT[] insert, but I could not cheaply reproduce why sql.array broke
- Whether the engine tie-break's flake actually manifests on the engine-parity seed data (requires real Postgres to observe cross-engine array_agg ordering divergence)

- `gh pr view 3482 --json title,body,mergeable,mergeStateStatus,files,...`
- `gh pr diff 3482 > /tmp/pr3482.diff`
- `gh pr checks 3482 (all pass)`
- `git fetch origin pull/3482/head:pr-3482 && git checkout pr-3482 (base fd8be831, 32 behind master)`
- `git diff fd8be831..origin/master -- <touched files> (overlap check: disjoint hunks)`
- `bun install`
- `bun test test/e2e/cross-modal-eval.test.ts test/e2e/dream-cycle-phase-order-pglite.test.ts > /tmp/pr3482_e2e_pr.txt (PR branch: 9 pass / 0 fail)`
- `git checkout origin/master && bun test <same two files> > /tmp/pr3482_e2e_master.txt (master: 6 pass / 3 fail — Q2, Q3, dream-cycle)`
- `git checkout pr-3482 && git checkout fd8be831 -- src/core/cross-modal-eval/aggregate.ts && bun test test/e2e/cross-modal-eval.test.ts test/cross-modal-eval-aggregate.test.ts (12 pass with OLD ordering — discrimination failure) && git checkout pr-3482 -- src/core/cross-modal-eval/aggregate.ts`
- `bun run typecheck > /tmp/pr3482_tc.txt (EXIT=0)`
- `grep verification: GBRAIN_TEST_DB guard (schema-drift.test.ts:103-107), min_below_5/mean_below_7 consumers (found unsynced sibling takes-quality-eval/aggregate.ts:136), gbrain-base.yaml note extractable:true + registry.ts:119 default, schema.sql:65 default-source seed, embedding-dim-check.ts halfvec, oauth scopes TEXT[], helpers.ts TRUNCATE CASCADE`
- `gh pr list --state merged --limit 8 + gh pr view 3566 --json files (merged fix(test) PRs carry no version bump/title prefix)`

</details>

---

time-attack:

The fixes are real and none landed in the v0.42.69.0 wave: reproduced test/e2e/dream-cycle-phase-order-pglite.test.ts failing on current master (532e5916) with "Export named 'willEmbedSynchronously' not found" (src/core/embedding.ts:210 exists but the test's mock.module omits it); extract-atoms expectations are stale because note is extractable:true (src/core/schema-pack/base/gbrain-base.yaml:223, unioned at src/core/cycle/extract-atoms.ts:116-127); phantom-redirect's pg_typeof != 'vector' assertion false-fails on halfvec facts.embedding (src/core/migrate.ts:2291); GBRAIN_TEST_DB=1 is the documented CI opt-in (test/e2e/schema-drift.test.ts:104) with the db-name hard floor preserved. Engine path_str tie-break lands identically in both engines (parity kept), the _chat seam mirrors the existing extract-atoms.ts seam pattern, and with the conflict resolved I got 17/17 tests passing plus a clean typecheck on the merged tree. But the PR cannot merge verbatim: its one conflicting file needs a semantic edit, so it must be rebased first.

**Conflict state vs current master:** Single conflicting file: test/e2e/cross-modal-eval.test.ts, and the conflict is semantic, not contextual. Master's #3531 (335e4703) retired gemini-1.5-pro — DEFAULT_SLOTS slot C is now deepseek:deepseek-v4-pro (src/core/cross-modal-eval/runner.ts:52-58) and gpt-4o was renamed gpt-5.2. The PR (based pre-#3531) still keys its chat stubs on 'google:gemini-1.5-pro'; taking the PR side verbatim makes all 4 cross-modal e2e tests fail with "mock: no scores configured for model deepseek:deepseek-v4-pro". The fix still applies post-wave and the resolution is mechanical: keep the PR's _chat-seam restructure, rename gemini-1.5-pro to deepseek:deepseek-v4-pro in the stub keys/order array (verified: 17/17 pass, typecheck clean). All 11 other files auto-merge cleanly.

**Risk if merged as-is:** Low after rebase. Production deltas: (a) aggregate.ts failReason precedence swap changes only which reason is reported when both mean and floor breach — verdict unchanged, no existing test pins mean-first on a both-breach fixture; note the sibling table src/core/takes-quality-eval/aggregate.ts:136 keeps mean-first (minor cross-eval inconsistency worth a follow-up, not a blocker); (b) runner.ts _chat defaults to gwChat; (c) deterministic ORDER BY tie-break in both engines. The real hazard is merging with the wrong conflict resolution (PR's gemini stub keys), which turns the local e2e gate red on 4 cross-modal tests.

time-attack:
Superseded: every E2E repair here now lives on master — engine tiebreak (#4170), dream-cycle mock (447f8195), halfvec assertion, oauth bind, note-extractable, source-state reset; the DB-free files pass 9/9. The aggregate reorder was review-rejected; the _chat seam and GBRAIN_TEST_DB opt-in were superseded by the mock-quarantine lint and #4128's guard rework.

PR #3483, opened 2026-07-28
fix(index): preserve code files containing NUL bytes

cybernaut6404:
## Summary

- sanitize U+0000 to a visible `\0` representation at the code storage boundary
- use the same stored representation for chunking and call-site offset mapping
- retain the raw source in the content hash so unchanged-file detection stays source-faithful

## Why

A real `gbrain sync` of this repository rejected five valid TypeScript files because PostgreSQL text columns cannot store U+0000. The files were silently absent from semantic code search.

## Evidence

- focused import suite: 29 passed, 0 failed
- current upstream verification: 32/32 checks passed
- operational retry: all five previously rejected files imported, 59 chunks created, 0 errors
- pinned source increased to the complete 2,298 pages

Follow-up to closed #3467, submitted separately as requested. No private brain content or local sync artifacts are included.

### DIFF
--- a/src/core/import-file.ts
+++ b/src/core/import-file.ts
@@ -1161,6 +1161,10 @@ export async function importCodeFile(
   const title = `${relativePath} (${lang})`;
   const sourceId = opts.sourceId;
   const txOpts = sourceId ? { sourceId } : undefined;
+  // PostgreSQL text columns reject U+0000 even though source files may
+  // legitimately contain it inside string/regex fixtures. Preserve a visible,
+  // searchable representation instead of dropping the entire code page.
+  const storageContent = content.replaceAll('\0', '\\0');
 
   const byteLength = Buffer.byteLength(content, 'utf-8');
   if (byteLength > MAX_FILE_SIZE) {
@@ -1202,7 +1206,7 @@ export async function importCodeFile(
   // from the chunker (nested methods carry ['ClassName'] etc.) so the
   // chunk-grain FTS trigger picks up scope for ranking and downstream
   // Layer 5 edge resolution can use scope-qualified identity.
-  const { chunks: codeChunks, edges: extractedEdges } = await chunkCodeTextFull(content, relativePath);
+  const { chunks: codeChunks, edges: extractedEdges } = await chunkCodeTextFull(storageContent, relativePath);
   const chunks: ChunkInput[] = codeChunks.map((c, i) => ({
     chunk_index: i,
     chunk_text: c.text,
@@ -1270,7 +1274,7 @@ export async function importCodeFile(
       type: 'code' as string,
       page_kind: 'code',
       title,
-      compiled_truth: content,
+      compiled_truth: storageContent,
       timeline: '',
       frontmatter: { language: lang, file: relativePath },
       content_hash: hash,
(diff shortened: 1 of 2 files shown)


time-attack:
Real bug, correct fix, and the offset analysis holds up — **two small amendments and this goes in.**

## Verified, including on a real engine

Master genuinely fails: my PGLite probe gives `error: invalid byte sequence for encoding "UTF8": 0x00`, page absent, 0 chunks. Your branch imports it. And the repo is its own witness — **6 checked-in `.ts` files contain a raw NUL** (`src/core/skill-catalog.ts`, `src/core/sync-failure-ledger.ts`, `src/core/chunkers/symbol-resolver.ts` ×3, `src/core/eval-capture-graph.ts`, `test/skill-catalog-security.test.ts`, `evals/embedding-default/harness-runner.ts`), used deliberately as map-key separators. Since `importFile` throws, gbrain cannot index its own source tree.

**Offset consistency: verified consistent.** The +1-char-per-NUL shift applies uniformly — tree-sitter parses the sanitized text so `callSiteByteOffset` is in sanitized space, `findChunkForOffset` walks the same string, and `compiled_truth` stores what was chunked. Line/col are untouched since the substitution adds no newline, and `code_edges` stores chunk ids plus qualified symbol names with **no** byte or column offsets, so nothing in code-intel is offset-anchored. Nothing anchors stored offsets back to a file seek either. (Aside: `findChunkForOffset` walks char indices against a value named `byteOffset` — a pre-existing non-ASCII mismatch at `edge-extractor.ts:564`. Your change shifts char index and UTF-8 byte offset by +1 identically, so it neither fixes nor worsens it.)

Tests: 29 pass on your branch; reverting `import-file.ts` to master gives 28 pass / 1 fail on exactly your new test.

## Amendment 1 — hash the sanitized text, not the raw (one identifier)

The stated design ("retain the raw source in the content hash") causes measurable churn. `reindex-code` re-imports from the **stored** `compiled_truth` (`src/commands/reindex-code.ts:299`), so it writes `hash(sanitized)`, and the next sync computes `hash(raw)` from disk. Measured on your branch:

```
REINDEX-from-compiled_truth     = {"status":"imported"}   <- expected "skipped"
RE-SYNC-from-disk-after-reindex = {"status":"imported"}   <- expected "skipped"
RE-SYNC-again                   = {"status":"skipped"}
```

That's two full re-chunks and re-embeds per `reindex --code` → `sync` cycle, forever, for every NUL-containing file. Hashing `storageContent` fixes it: sanitization is idempotent (`\0` contains no NUL), so both entry points converge and unchanged-detection still works — I confirmed sync-only detection is fine either way.

## Amendment 2 — keep the loud failure for mis-decoded files

`importFromFile` does `readFileSync(filePath, 'utf-8')` with no BOM or encoding detection (the live bug that closed PR #2393 targeted). A UTF-16 file decodes to NUL-per-character. Pre-PR that failed loudly; post-PR it **silently imports as `e\0x\0p\0o\0r\0t\0…`, gets chunked, embedded, and served by semantic code search.** That's a worse outcome than the crash it replaces.

A density guard keeps both properties: bail (or flag) when NULs exceed ~1% of characters or a UTF-16 BOM is present, while still fixing the sparse intentional NULs you're targeting. Related: `byteLength` is still measured on raw `content`, so a NUL-dense file can cross `MAX_FILE_SIZE` after sanitization.

## Two notes, not blockers

- **Scope:** markdown/text pages with a NUL still hard-fail — `importFromContent` writes `compiled_truth: parsed.compiled_truth` unsanitized. Narrow scope is defensible since NULs realistically show up in code, but the engine `putPage` boundary is where a complete fix would live.
- **The test is mock-only.** It uses `mockEngine`, so it pins the sanitized shape but never proves a real engine accepts the row — and the actual defect is a Postgres constraint a mock can't express. A ~15-line PGLite test would pin it properly; I wrote and ran one to verify, happy for you to take that approach.

Both amendments are small. Send them and I'll merge.

PR #3484, opened 2026-07-28
feat(search): pin versioned retrieval contracts

bluk1020:
## Summary

Persisted vectors can currently be searched under a different runtime retrieval configuration without an explicit compatibility boundary. Matching dimensions are not enough: provider/model identity, query-vs-document encoding, storage representation, distance metric, fusion behavior, reranking, contextual text policy, and post-processing all affect retrieval semantics.

This introduces an opt-in, versioned Search Contract that can be pinned to a brain. Once pinned, ordinary runtime entrypoints fail closed when the active configuration is incompatible, while doctor, config, contract inspection, and explicit embedding-migration commands remain available for recovery.

The contract fingerprint is also part of semantic query-cache identity, so retrieval-defining changes invalidate stale cache entries. Doctor reports contract, schema, and embedding-provenance drift without making explicit repair workflows unreachable.

Existing installations remain unchanged until an operator runs `gbrain search contract pin`. The pin survives search-mode reset and can be inspected with `gbrain search contract show` or validated with `gbrain search contract check`. Future embedding-space changes are documented as reviewed migrations rather than dimension-only configuration edits.

## Validation

- `bun test test/search-contract.test.ts test/search-mode.test.ts test/doctor-search-mode.test.ts test/commands-search.test.ts test/cli-search-dispatch.test.ts test/doctor-categories.test.ts test/doctor.test.ts test/search.test.ts test/hybrid-search-lite.serial.test.ts` — 270 passed, 0 failed
- `bun run verify` — 32/32 checks passed
- `gbrain eval export --since 7d` — zero captured rows, so retrieval replay was not available

## New concepts

**Retrieval-space contract**

A retrieval-space contract is a serialized compatibility boundary for persisted vectors and the behavior that consumes them. It treats semantic identity as more than vector width: two 2,560-dimensional vectors may still be incompatible when they come from different models or use different query/document encoding semantics.

The contract is pinned separately from tunable search settings. Operators can tune within the compatible space, but changing a field that redefines that space produces an explicit drift failure and a cache-fingerprint change instead of silently mixing results.

This pattern is useful when embeddings outlive individual process configurations or deployments. It is unnecessary for stateless retrieval systems that never persist vectors or semantic caches.

### DIFF
--- a/docs/architecture/KEY_FILES.md
+++ b/docs/architecture/KEY_FILES.md
@@ -8,6 +8,8 @@ lives in `CHANGELOG.md` + `git log` / `git blame`, NOT here. Do not append
 per-release `**vX.Y.Z:**` narration — CI enforces this
 (`scripts/check-key-files-current-state.sh`).
 
+- `docs/architecture/SEARCH_CONTRACT.md` + `src/core/search/search-contract.ts` — versioned retrieval compatibility pin. `gbrain search contract pin|check` records provider/model, dimensions, pgvector column/representation, query/document semantics, cosine metric, contextual-retrieval profile, reranker policy, and autocut. Ordinary runtime startup fails closed on pinned drift; doctor/config/migration recovery remain reachable. The contract fingerprint participates in query-cache identity.
+
 - `docs/operations/conversation-parser-llm-fallback.md` — operator and maintainer contract for the default-off LLM parse fallback: exact config key, deterministic-first dispatch boundary, sampled data surface, untrusted-content prompt handling, page-date/cache-key coupling, timestamp validation, cache/checkpoint behavior, observability, limitations, and focused test commands.
 
 - `src/commands/serve-http.ts` confidential revoke extension — a pre-router `/revoke` handler validates the RFC 7009 body, verifies hash-only secrets for both `client_secret_post` and `client_secret_basic`, rejects mixed authentication, preserves the SDK path for public clients, and separates opaque client-auth failures from retryable/backend failures. OAuth metadata advertises both confidential methods. Pinned by `test/e2e/serve-http-oauth.test.ts`.
--- a/docs/architecture/SEARCH_CONTRACT.md
+++ b/docs/architecture/SEARCH_CONTRACT.md
@@ -1,0 +1,73 @@
+# Search Contract v1
+
+GBrain Search Contract v1 is an opt-in compatibility pin for the embedding
+space and retrieval profile that produce persisted vectors and cached results.
+It does not select a new model and it does not move inference into PostgreSQL.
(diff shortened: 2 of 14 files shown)


time-attack:
Closing — thanks for the thorough writeup, and credit where it's due: you bumped `KNOBS_HASH_VERSION` (13→14) correctly, and CI-wise the llms/docs side is clean. But this doesn't clear the gatekeeping bar, for two reasons.

**1. The hazard is already guarded at four layers.** Dimension mismatch is loud, not silent (`src/core/embedding-dim-check.ts`, `src/core/ai/dims.ts`). Cache contamination is already covered — `knobsHash` in `src/core/search/mode.ts` folds in every field the new contract carries (`prov`, `col`, `rr/rrm/rri/rro/rrt`, `ac/acj`, `cr/crd`), and `isCacheSafe()` already skips both cache read and writeback on model/dim disagreement. Provenance already exists via v108 `pages.embedding_signature`, and drift self-heals through `embed --stale` on every `runEmbedCore` caller. Doctor already has `embedding_width_consistency`, `embedding_env_override`, `embedding_provider`, `ze_embedding_health`.

**2. The fail-closed gate is net-negative.** The pin records knobs that are *designed to be tuned*. `gbrain search tune --apply` writes `search.mode` (`src/commands/search.ts:405,429`) — on a pinned brain that single write makes every subsequent `query`/`sync`/`import`/`embed`/`serve` throw at `connectEngine()`. `reranker.timeout_ms` isn't even result-affecting, yet a timeout tweak fails the brain closed. Same on upgrade: the pin stores resolved bundle values, so any future release that nudges `balanced.reranker_top_n_in` bricks every pinned brain right after `gbrain upgrade`, MCP `serve` included, with no migration path. And the recovery path the doc promises is blocked — `src/cli.ts:118` exempts only `search contract`, so `search modes --reset` (the documented way to clear `search.*`) still throws.

Also worth noting: `checkSearchContractV1Health` inlines a stale predicate instead of calling `engine.countStaleChunks({ includeNullSignature: true })` — `src/core/embed-skip.ts` exists specifically to prevent that duplication. And there's no CI signal on this branch at all for a 683-line change to the search hot path.

**The genuine gap you found, and it is real:** no doctor surface reports stale/absent `embedding_signature` counts, so a same-dimension provider swap degrades recall silently until the next stale sweep. That's worth fixing — as roughly 20 lines in `doctor.ts` calling the existing `countStaleChunks`, warning with `gbrain embed --stale`. No new file, no new CLI subcommand, no new config key, no startup gate, no cache cold-miss. If you want to send that, it'd be reviewed in the fix lane.

PR #3487, opened 2026-07-28
fix(cycle): isolate global transcript discovery

### LINKED ISSUE #2666: dream: patterns/synthesize write output files into other registered sources' work trees (multi-source brain)
## Symptom

On a multi-source brain, dream phase outputs (patterns / synthesize product files) sometimes land in the work tree of a *different* registered source instead of the host/default source.

Observed on 0.42.56.0 (three sources: default wiki + a gstack checkout + a read-only KB clone):
- After a patterns run, 3 stray output files appeared inside the gstack source work tree and 1 inside the KB clone work tree — none belong to those sources.
- This recurs on every patterns run (verified twice on consecutive runs after clean-up).
- Earlier in June the same mechanism left ~74 stray pages that later synced into the wrong source; because one of those work trees prunes untracked dirs, strays can also trigger un-syncable page deletions downstream.

## Impact

Wrong-source pages pollute cross-source search and require manual surgery (we've now cleaned up twice, with backups). Combined with source-side pruning, it can escalate from cosmetic to data-loss-adjacent.

## Suggested fix

Dream writers should resolve output paths against the host/default source root explicitly, rather than whatever work tree the phase's cwd/traversal happens to touch.

frxiaobei:
## Summary

- restrict configured transcript corpus discovery to the `default` source
- keep non-default source DB page extraction unchanged
- preserve explicit `_transcripts` injection for focused callers and tests

Configured transcript corpus paths are brain-global. Discovering them for every federated source duplicated atom extraction and model spend.

Closes #2666

## Validation

- `bun test test/extract-atoms-page-discovery.test.ts test/extract-atoms-drain.test.ts` - 39 pass, 0 fail
- `bun run typecheck` - pass
- `GBRAIN_HOME=/private/tmp/gbrain-pr3487-final-verify bun run verify` - 54/54 checks green
- `git diff --check upstream/master...HEAD` - clean
- compared `ci:local:diff` failures with clean `upstream/master` at the same SHA; no branch-specific failures


time-attack:
Reviewed. One genuine bug, wrapped in an architectural change that needs to be decided separately. Please split.

**The real bug.** `src/core/cycle/extract-atoms.ts:423` runs `discoverTranscripts()` over the brain-global corpus dirs (`dream.synthesize.session_corpus_dir`, `meeting_transcripts_dir`) for *every* source, then `putPage(..., {sourceId})`. Since master fans `NON_GLOBAL_PHASES` out to every source (`autopilot-fanout.ts`), each federated source writes its own copy of the same transcript-derived atoms — and pays Haiku again for each. That fires in production. The `sourceId === 'default'` gate on transcript auto-discovery is ~6 lines and lands on its own; send that.

**What needs to be separate:**
- The new hardcoded `NON_DEFAULT_FANOUT_PHASES` allowlist is a fanout-policy decision, not a bug fix.
- Claim-time phase re-validation is framed as defense against "hand-crafted queue payloads," but the minion queue is a local trusted table — anyone who can INSERT there already owns the DB. That's not a threat model that justifies the surface.
- Switching enumeration from `listAllSources({localPathOnly:true})` to all-sources-plus-NULL-path-default → `sync.repo_path` is a behavior change affecting existing installs, and needs its own justification and tests.

No CI has run upstream on this branch yet (fork workflows await approval).

time-attack:
**Merging this closes #2666** (severity **p0**).

#2666 was one of 58 issues carrying a `p3` label. All 58 were re-triaged by reproducing them against current master, and **the label was wrong on 56 of them** — the corrected distribution is 4 p0, 40 p1, 12 p2, and 2 genuinely p3. #2666 came out at **p0**, and this PR is what fixes it.

Two things follow from that:

- Add `Closes #2666` to the PR description if you want the link wired up — a closing keyword only fires from the body or a commit message, not from a comment like this one.

The reproduction and evidence are on #2666.

time-attack:
Status update: reviewed in the 2026-08-18 full-backlog triage — still wanted, but not yet mergeable: Genuine extract_atoms transcript leak persists, but the contested phase-allowlist must be split per review and the #2666 write class was fixed on master. If you update it, it will be re-reviewed promptly; alternatively the fix may be adopted with credit in an upcoming fix wave. Thanks for the contribution and the patience.

time-attack:

PR #3488, opened 2026-07-28
fix(search): preserve homogeneous results during type diversity

frxiaobei:
## Summary

- skip the page-type diversity ratio when every candidate has the same type
- preserve all distinct same-type results instead of dropping them without adding diversity
- document the behavior in the architecture key-file map

This fixes the nightly-7-shaped failure where three relevant `note` sessions were reduced to two by the 60% type cap even though no alternate page type existed.

## Tests

- `bun test test/dedup.test.ts` (15 pass)
- `bun run typecheck`
- equivalent fork CI: [Test](https://github.com/herove/gbrain/actions/runs/30341347752), [E2E Tests](https://github.com/herove/gbrain/actions/runs/30341354962), and [Semgrep](https://github.com/herove/gbrain/actions/runs/30341358619) all passed

Upstream fork-triggered checks currently require maintainer approval. This PR is independent and can land on its own.


time-attack:
This one looks finished — could you mark it ready for review?

Verified the bug independently: `enforceTypeDiversity` (`src/core/search/dedup.ts:129`) with `MAX_TYPE_RATIO = 0.6` (`dedup.ts:23`) means 3 same-type results give `maxPerType = ceil(3*0.6) = 2`, so one relevant result is silently dropped and no diversity is gained in exchange. Layers 1/2/4 don't catch it.

It's a single guard line at the right place — the shared layer rather than a caller — with one regression test, no scope creep, and the `KEY_FILES.md` edit obeys the current-state rule. Of the batch you sent, this is the one that's ready to go; it just needs CI to actually run, which requires undrafting (fork workflows are gated on maintainer approval, so `statusCheckRollup` is empty right now).

time-attack:
Status update: reviewed in the 2026-08-18 full-backlog triage. It is queued but currently deferred: Now CONFLICTING/DIRTY on master (GitHub mergeable=CONFLICTING); needs rebase before it can queue. Address that and it re-enters the merge queue. Thanks for the contribution.

time-attack:

PR #3489, opened 2026-07-28
test(schema): exercise v121 bootstrap coverage

cybernaut6404:
## Summary

- strip the v121 `timeline_entries.event_page_id` column and dependent indexes from both synthetic pre-migration fixtures
- assert the column is genuinely absent before exercising the forward-reference bootstrap
- ensure both isolated bootstrap recovery and full PGLite schema replay cover this migration

## Root cause

The coverage manifest included the v121 column, but the synthetic old-brain fixture never removed it. The post-bootstrap assertion therefore passed without exercising recovery.

## Verification

- failing reproduction before fixture repair: 8 passed, 1 failed
- targeted regression: 9 passed, 0 failed
- `git diff --check`

### DIFF
--- a/test/schema-bootstrap-coverage.test.ts
+++ b/test/schema-bootstrap-coverage.test.ts
@@ -256,6 +256,11 @@ test('applyForwardReferenceBootstrap covers every forward reference declared in 
       ALTER TABLE pages DROP COLUMN IF EXISTS generation;
       ALTER TABLE pages DROP COLUMN IF EXISTS contextual_retrieval_mode;
       ALTER TABLE pages DROP COLUMN IF EXISTS corpus_generation;
+
+      DROP INDEX IF EXISTS idx_timeline_event_dedup;
+      DROP INDEX IF EXISTS idx_timeline_event_page;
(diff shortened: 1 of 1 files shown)


time-attack:
Merging. Verified independently rather than taking the body's word for it.

The guard really was vacuous: `test/schema-bootstrap-coverage.test.ts:173` declares `{kind:'column', table:'timeline_entries', column:'event_page_id'}` in `REQUIRED_BOOTSTRAP_COVERAGE`, and `src/core/pglite-engine.ts` has the whole recovery branch (`:546` probe, `:667-668` `needsTimelineEventPageId`, `:933` the ALTER) — but the synthetic fixture never stripped the column, so the flag was always `false` and the post-bootstrap assertion passed without the bootstrap doing anything.

Reproduced your claimed failure exactly: removing only the first fixture-strip block while keeping the new assertion gives `8 pass / 1 fail` at `:273`, and the unmodified PR gives `9 pass / 0 fail`. Matches the body.

The explicit pre-bootstrap `expect(...).toHaveLength(0)` is the part that makes this durable — it's what stops the fixture silently re-rotting. Good catch.

PR #3490, opened 2026-07-28
fix(cycle): make nightly quality probes budget-safe and retry-safe

herove:
## Summary

- derive the nightly question count from the configured cross-modal judge budget
- remove the adapter's `--yes` cap bypass and pass the selected `--limit` explicitly
- short-circuit before LongMemEval when the budget cannot afford one judged question
- rotate from the latest compatible audit cursor instead of the calendar day
- advance the cursor only after a complete, internally consistent summary
- audit attempted question IDs on errors or partial results without consuming the cursor, so the same batch retries
- treat exit 1 without a summary as an ambiguous runtime error rather than guessing `budget_exceeded`

At current defaults, the estimator selects one question for `$0.20` and all 10 committed nightly questions for the default `$5` cap.

## Tests

- related nightly/cross-modal tests: 73 pass
- LongMemEval + hybrid slow/serial tests: 33 pass
- `bun run typecheck`
- equivalent fork CI: [Test](https://github.com/herove/gbrain/actions/runs/30341373367), [E2E Tests](https://github.com/herove/gbrain/actions/runs/30341376848), and [Semgrep](https://github.com/herove/gbrain/actions/runs/30341380279) all passed

Upstream fork-triggered checks currently require maintainer approval. This PR is independent of the search-dedup and aggregate-normalization fixes.


time-attack:
Closing — there's a real bug in here, but it's one line inside ~880, and the framing doesn't hold up.

**The real bug, and it's worth fixing:** `src/core/cycle/nightly-probe-adapters.ts:107` passes `--yes` alongside `--max-usd`, and `--yes` is precisely what bypasses the pre-flight cap. So the nightly probe's `max_usd` is decorative today — a genuine spend-safety defect, fixed by deleting one array element. Send that alone and it goes in the fix lane.

**Why the rest doesn't:** budget-derived batch sizing, a durable fixture-rotation cursor recovered from ISO-week audit files, summary-denominator validation, and audit schema extensions are ~700 lines of new feature on a probe that is default-OFF and opt-in.

**One correction on the body.** It says the change rotates "from the latest compatible audit cursor *instead of the calendar day*." Master has no rotation of any kind — `grep -nE 'getUTCDay|getDay|rotat' src/core/cycle/nightly-quality-probe.ts` comes back empty; it runs the whole 10-row fixture. Please don't describe added behavior as replacing behavior that isn't there; it makes review slower for everyone.

**Undeclared conflict.** This is described as independent, but it collides semantically with #3498 — both rewrite the same region around `PROBE_QA_DIMENSIONS` / `runCrossModalBatchForProbe`, and both make contradictory edits to the same `docs/eval-bench.md` and `KEY_FILES.md` lines. Separately, #3501 swaps cross-modal slot C off the retired Gemini model, which invalidates this PR's `estimateCost(DEFAULT_SLOTS, …)` selector and the hard-coded "$2.00 for 10 questions" in its docs.

Also note no upstream CI has ever run on this branch — `statusCheckRollup` is empty because fork workflows await maintainer approval. The fork-side runs linked in the body don't tell us anything about this base.

PR #3491, opened 2026-07-28
fix(eval): normalize cross-modal judge dimension names

frxiaobei:
## Summary

- trim and lowercase judge dimension keys before cross-model aggregation
- match each model's scores by the same normalized key
- prevent casing differences such as `CORRECTNESS` vs `correctness` from becoming separate one-model dimensions

## Tests

- `bun test test/cross-modal-eval-aggregate.test.ts` (9 pass)
- `bun run typecheck`
- equivalent fork CI: [Test](https://github.com/herove/gbrain/actions/runs/30341362524), [E2E Tests](https://github.com/herove/gbrain/actions/runs/30341366105), and [Semgrep](https://github.com/herove/gbrain/actions/runs/30341369673) all passed

Upstream fork-triggered checks currently require maintainer approval. This PR is independent and can land on its own.


time-attack:
Reviewed — the failure class is real, but this fixes the symptom rather than the cause. Keeping it open for a smaller, more complete fix.

**Confirmed real.** `src/core/cross-modal-eval/aggregate.ts:93` does a raw `allDimNames.add(dim)`, and judge key names are **model-chosen** — `buildPrompt` (`runner.ts:259-286`) only shows `"dim_1_name"` as a placeholder and never tells the judge which keys to emit. Divergent keys split into one-model dimensions, and since the pass criterion is `every dim mean >= 7`, a 9/5 pair fails the whole panel instead of passing at mean 7. That's a genuine eval-correctness bug.

**Why `trim().toLowerCase()` isn't enough.** It covers case and whitespace, but `Goal Achievement` vs `GOAL_ACHIEVEMENT` vs `goal_achievement` still split three ways — separator differences survive normalization. The root-cause fix is to pin the exact dimension keys in the prompt so the judge can't invent variants (optionally plus stripping non-alphanumerics as a belt), and that's *smaller* than the current per-dimension `.find()` loop.

One side effect to note either way: aggregate output and `describeFailure` text now render dimensions lowercase, where committed receipts carry the model's original casing.

time-attack:
Status update: reviewed in the 2026-08-18 full-backlog triage — still wanted, but not yet mergeable: Real dim-splitting bug still on master, but reviewer wants the root cause (pin judge keys in the prompt) and the draft is conflicting on KEY_FILES.md. If you update it, it will be re-reviewed promptly; alternatively the fix may be adopted with credit in an upcoming fix wave. Thanks for the contribution and the patience.

time-attack:
Re-trialed against the current integration branch (post v0.46.23-25 waves): Conflict only in docs/architecture/KEY_FILES.md: PR's stale aggregate/runner doc lines vs wave's #3531 slot rewrite. src+test merge clean; not superseded (wave still lacks dim normalization). Trivial doc fix. No red flags. — a rebase along those lines puts this back in the active merge queue; re-trials run nightly, so a fresh push gets picked up automatically. Thank you!

garrytan:

Issue #3439, opened 2026-07-27
RerankerTouchpoint's wire-shape is hardcoded to ZeroEntropy's top_n/results — no adapter fields for other rerank providers


mrudoy:
## `RerankerTouchpoint`'s wire-shape is hardcoded to ZeroEntropy's `top_n`/`results` — no adapter fields for other rerank providers

**Context:** ZeroEntropy (the only reranker recipe gbrain ships today, `zeroentropyai`) shuts down 2026-09-04 (`zeroentropy.dev` migration notice; see #3390). Every gbrain brain using the reranker needs a replacement provider before then. I migrated `gbrain-gb`'s reranker to Voyage AI (`rerank-2.5`) as a real-world test of what that migration requires.

**What I found:** `gateway.ts`'s `rerank()` hardcodes two things that are NOT universal across rerank APIs:

1. **Request body key name** — always sends `top_n` (`gateway.ts` ~line 3282, pre-patch):
   ```ts
   ...(input.topN !== undefined ? { top_n: input.topN } : {}),
   ```
   Voyage's `/rerank` endpoint requires `top_k`, not `top_n`. Sending `top_n` is silently ignored by Voyage's API (not an error — it just doesn't cap results), which is a worse failure mode than a loud rejection.

2. **Response array key name** — always reads `json.results` (`gateway.ts` ~line 3354, pre-patch):
   ```ts
   if (!json || !Array.isArray(json.results)) {
     throw new RerankError('rerank: malformed response (no results array)', 'unknown');
   }
   const mapped = json.results.map(...)
   ```
   Voyage returns the array under `json.data`, not `json.results`. Every Voyage rerank call fails with `"rerank: malformed response (no results array)"` even though the HTTP call succeeded (200) and returned well-formed data — just under a different key.

The `path` field is *already* recipe-pluggable (`tp.path ?? '/models/rerank'`), which is why this is a partial gap, not a total one — Voyage's request correctly reaches `/rerank` (verified live against `api.voyageai.com`, confirming `/models/rerank` 404s and `/rerank` 200s), it just can't parse the reply or set a result cap.

Inner item shape (`{index, relevance_score}`) is identical between ZE and Voyage, so no adapter is needed there.

**What I did locally (patch attached, `gbrain-gb`-only, lost on reinstall):**
- Added two optional fields to `RerankerTouchpoint` in `types.ts`: `top_n_param?: string` (default `'top_n'`) and `response_key?: string` (default `'results'`).
[...]

Issue #3421, opened 2026-07-26
Reranker fail-open is silent — HybridSearchMeta has no rerank_status, so callers can't distinguish a reranked result set from an un-reranked fallback


mrudoy:
**What happened?**

`core/search/rerank.ts`'s `applyReranker()` fails open on every error class — logs to the rerank-audit JSONL and returns the original RRF order. The posture is right; a flaky reranker must never break search. The problem is that it is **invisible to the caller at query time**.

`applyReranker` returns `SearchResult[]`. Success and failure are the same type, the same length, and the same shape. The only in-band difference is that `rerank_score` is left undefined on every result — easy to miss, undocumented as a failure signal, and indistinguishable from "reranker disabled for this mode."

`HybridSearchMeta` — the struct whose whole job is telling a caller *what actually ran* — has a field for the vector arm (`vector_enabled`) and one for expansion (`expansion_applied`), but **nothing for the reranker**.

So a query can silently return materially worse ordering and nothing downstream knows. Any consumer doing a before/after comparison (migration eval, regression harness, A/B of a config change) will attribute that movement to whatever it was actually testing.

**What did you expect?**

The same treatment the vector arm already gets: a field in `HybridSearchMeta` plus one stderr line, so a fail-open is observable in-band.

This is exactly the ask in **#2028** ("set a `vector_enabled: false` / `embed_timeout: true` flag in HybridSearchMeta and emit one stderr warning") applied to the reranker arm instead of the embed arm. `vector_enabled` exists; the reranker has no equivalent.

Related but distinct: **#3386** and **#2059** both concern how reranker failures are *classified in the audit log and counted by doctor*. Those are out-of-band, after-the-fact surfaces. This issue is that the failure never reaches the caller **during the query** at all. #3386's own framing — "the reranker had been failing open on every single search and the operator had no signal proportional to that" — is the same underlying pain from the other end.

**Steps to reproduce**

1. Force a failure: `gbrain config set search.reranker.timeout_ms 1`
2. Run a search through `hybridSearch` with an `onMeta` callback:

```js
let meta = null;
const results = await hybridSearch(engine, 'some query', {
  mode: 'tokenmax', limit: 50, autocut: false,
  onMeta: (m) => { meta = m; },
});

console.log(results.length);                                     // 50 — looks fine
console.log(meta.vector_enabled, meta.expansion_applied);        // true false
console.log(results.filter(r => r.rerank_score != null).length); // 0
[...]

PR #3492, opened 2026-07-28
v0.42.69.0 feat(ai): default reranker to cohere:rerank-v3.5 before the ZeroEntropy sunset

time-attack:
## Why now

ZeroEntropy's hosted API — including `/v1/models/rerank` — stops serving **2026-09-04** (#3390). gbrain's default reranker pointed at it.

## The fix that matters most is not the provider swap (#3421)

There was **no `isAvailable('reranker')` check before `applyReranker`**, even though the embedding arm has had one for releases (`hybrid.ts` ~1157). `balanced` is `DEFAULT_SEARCH_MODE` and ships `reranker_enabled: true`, so *today* — not just post-sunset — every query on the default mode from an install with no reranker key issues a doomed HTTP request and burns the full 5s `reranker_timeout_ms` before `applyReranker` fails open. Silently, on every search.

`hybrid.ts` now pre-gates on the **resolved per-call** reranker model, the same reason the embedding arm probes the resolved column's provider rather than the global default. The `rerankerFn` test seam is exempt (same shape as `isAvailable`'s existing chat-transport-stub exemption). This one guard makes the sunset a non-event for anyone without a reranker key.

## Why Cohere, natively

- Cohere **created the commercial rerank-API category** and has shipped v2 → v3 → v3.5 → v4 while **keeping v3.5 served after v4 launched** — a demonstrated non-abandonment record, which is exactly what a default-retrieval-path provider needs.
- It is the **most widely integrated reranker** in every RAG framework, which is why its dialect became the de-facto standard (see below).
- **ZeroEntropy's own migration guide names it as the target.**
- It satisfies the Default-provider policy #3481 adds to CLAUDE.md (open-weight OR longest proven model-lifetime record) on the second limb.

**Deliberately NOT via OpenRouter**, even though that path needed zero code: routing the *default* retrieval path through a startup proxy adds a single point of failure. OpenRouter reranking stays the opt-in it already is.

## Key finding: this was cheap because no adapter was needed — verified

gbrain's so-called "ZeroEntropy wire shape" **IS** the industry-standard Cohere dialect. Verified independently rather than taken on trust:

- `docs.cohere.com/reference/rerank` → `POST https://api.cohere.com/v2/rerank`, request `{model, query, documents[], top_n?}`, response `results[{index, relevance_score}]`, scores normalized to `[0,1]`.
- That is byte-identical to what `gateway.rerank()` already sends and parses. `dashscope-rerank.ts`'s own header comment already says so out loud ("Wire shape matches ZeroEntropy: request `{model, query, documents, top_n?}`, response `{results: [{index, relevance_score}]}`").
- So **#3439 does not block this** — Voyage (`top_k` / `data[]`) is the sole outlier, exactly as the issue describes.

`test/ai/recipe-cohere.test.ts` pins the roundtrip end to end (URL with no `/v2/v2` doubling, `top_n` not `top_k`, Bearer from `COHERE_API_KEY`, `results[]` → `RerankResult[]`, allowlist behavior), so if that stops being true, tests fail instead of production searches.

## What's in the diff

| Area | Change |
|---|---|
| **Pre-gate (#3421)** | `isAvailable('reranker', resolvedModel)` before `applyReranker`; `rerankerFn` exempt |
| **New recipe** | `src/core/ai/recipes/cohere.ts` — rerank-only (precedent: `dashscope-rerank.ts`), `api.cohere.com/v2` + `/rerank`, `rerank-v3.5` / `rerank-v4.0-fast` / `rerank-v4.0-pro`, registered in `recipes/index.ts` |
| **Stop persisting a dead model** | `retrieval-upgrade-planner.ts`'s `ZE_TARGET_RERANKER_MODEL` was writing `zeroentropyai:zerank-2` into **user config** at two sites (~411, ~478). Now points at the live default |
| **Default swap** | `DEFAULT_RERANKER_MODEL` + all three `mode.ts` bundles + `doctor.ts`/`models.ts` hints that hardcoded one provider's env var |
| **Key threading** | `cohere_api_key`: file plane → `loadConfig` env merge → `buildGatewayConfig` → recipe. The in-code comments at `build-gateway-config.ts:32-46` document this exact bug shipping **twice** (ZE v0.37, Voyage #2662) — key accepted at the file plane, never reaching the gateway, silent failure in daemon/launchd/MCP |
| **Pricing** | Cohere rerank keys added to `embedding-pricing.ts` (the table `budget-tracker`'s rerank lookup falls back to) so `--max-cost` callers don't TX2 hard-fail **on the default path**. Plus the OpenRouter pseudo-rate correction |
| **Docs** | Reranker provider doc now leads with self-hosting (vLLM recommended); ZE provider doc gets a deprecation banner |
| **Test hygiene** | `test/e2e/zeroentropy-live.test.ts` skips on/after 2026-09-04 (`GBRAIN_ZE_SUNSET_OVERRIDE=1` to force) |

## Pricing honesty — read this before trusting the numbers

Cohere bills rerank **per search** (one query + up to 100 docs), not per token. I did **not** add a `cost_per_search_usd` field: the budget tracker carries one `$/1M-token` unit end to end, so a per-search unit is a tracker change, not a recipe change — and the task explicitly allowed pricing conservatively with a comment instead. So both `cohere.ts` and `embedding-pricing.ts` carry **deliberately conservative pseudo-rates** with the derivation spelled out:

> `balanced` sends `top_n_in=25` chunks ≈ 10K estimated tokens (tracker estimates chars/4), so `0.20/1M` ≈ **$0.002/search** — the higher of the two rates third-party trackers report. It over-estimates on `tokenmax` (50 docs, same per-search bill), which is the safe direction for `--max-cost`.

**Could not verify the real Cohere per-search rate.** `cohere.com/pricing` renders the rerank per-search row client-side — WebFetch got the Model Vault infrastructure table ($5.00/hr, $3,250/mo Medium) and the FAQ's definition of a search unit, but no per-search figure. Third-party trackers **disagree**: one says $2.00/1K searches, another $0.001/search ($1.00/1K), and `cloudprice.net` renders it as "$0.0020 per 1M tokens" which is a unit error. The Microsoft Marketplace listing returns 403. **A human with a Cohere account should confirm the rate and tighten the pseudo-rate if it's off.**

**OpenRouter's pseudo-rate was worse than 4x low.** OpenRouter publishes **$0.001/search** for `cohere/rerank-v3.5` (verified on `openrouter.ai/cohere/rerank-v3.5`). The recipe's `cost_per_1m_tokens_usd: 0.001` estimated **$0.00001/search** at gbrain's `balanced` shape — ~100x low. The comment defending it ("~$0.00025 at ~4K chars/search") was also arithmetically wrong on its own terms (4K chars → 1K tokens → $0.000001). Corrected to `0.10`, which puts a 10K-token search at exactly $0.001, with the derivation in the comment.

## Self-host fallback — and a correction to the llama.cpp premise

**vLLM is the recommendation.** `docs.vllm.ai` labels its endpoint the **"Cohere Rerank API"** (`/rerank`, `/v1/rerank`, `/v2/rerank`), so it works through this PR's recipe with **no code change** — just repoint `provider_base_urls.cohere`.

**Correction on llama.cpp #16407.** The brief described it as "open since 2025-10-03." It is **closed as completed 2025-10-09** (`gh api repos/ggml-org/llama.cpp/issues/16407` → `state: closed`, `state_reason: completed`). The bug was real and severe — near-zero uncorrelated scores (~`1e-28`) for Qwen3-Reranker 0.6B/4B/8B, plus wrong/non-matching scores for BGE, mxbai, and Jina — but it is fixed upstream. Writing "open" into the doc would have been false, so the doc instead warns about what still bites users: **the failure mode is silent** (a 200 with garbage numbers defeats both gbrain's fail-open contract and `gbrain models doctor`), so **pin a build after 2025-10-09** and run the included `curl` sanity check after any llama.cpp upgrade.

**ZE weights licensing:** `huggingface.co/zeroentropy/zerank-2` now reads `apache-2.0` (verified 2026-07-28), so self-hosting is viable post-sunset. The page carries no relicense date, so the docs say "previously CC-BY-NC-4.0, relicense is recent, third-party pages still say non-commercial — check upstream, not the mirrors" rather than asserting the 2026-07-24 date I couldn't confirm.

## Scope notes (found, deliberately not fixed here)

- **`gbrain config set cohere_api_key` does nothing useful.** It writes the DB plane, which `loadConfig` deliberately does not merge for any `*_api_key` field — a pre-existing gap shared by `zeroentropy_api_key` / `voyage_api_key` and documented in `config.ts`'s own comments. Rather than expand scope into a secrets-in-DB change, the recipe `setup_hint`, the provider doc, and the CHANGELOG all now point at `COHERE_API_KEY` or `~/.gbrain/config.json` and say so explicitly. Worth a follow-up issue.
- **The whole `ze-switch` / retrieval-upgrade-prompt flow still advertises upgrading to a dying embedding provider.** Out of scope here; this PR only stops it persisting a dead *reranker* string and de-brands the one benchmark line that named `zerank-2` while the planner installs Cohere.
- **ZE *embedding* migration is untouched and remains the urgent, non-automatic half of the sunset** (queries embed through the same endpoint that produced your stored vectors). `gbrain upgrade`'s existing banner already covers it; the ZE doc's new banner repeats it.

## Receipts

| Gate | Result |
|---|---|
| `bun run typecheck` | **exit 0** |
| `test/ai/recipe-cohere.test.ts` (new) | 12 pass / 0 fail |
| `test/search/hybrid-reranker-integration.serial.test.ts` (+2 pre-gate tests) | 9 pass / 0 fail |
| Touched-test sweep (18 files: search-mode, models-doctor-reranker, balanced-reranker-default, knobs-hash-reranker, ai/rerank, budget-tracker, embedding-migration, retrieval-upgrade-planner, doctor, openrouter-reranker-recipe, recipes-existing-regression, loadConfig-merge, zeroentropy-recipe, rerank-audit, search/rerank, e2e/zeroentropy-live, build-llms, …) | **412 pass / 0 fail** |
| `bun test test/search/ test/ai/` | 820 pass / 0 fail (after fixing `gateway.test.ts`, which used `cohere` as its stand-in for an *unregistered* provider) |
| `scripts/check-no-tracked-symlinks.sh` | OK (no tracked symlinks) |
| `scripts/check-key-files-current-state.sh` | OK |
| `bun run build:llms` + `test/build-llms.test.ts` | regenerated, test passes |
| VERSION / package.json / CHANGELOG audit | all three read `0.42.69.0` |
| **CI on this PR** | **all 24 checks green** — all 10 `test` shards, `serial-tests`, `verify`, `semgrep`, `gitleaks`, `osv-scan`, `JSONB parity`, Tier 1 + Tier 2 |

**Local-only failures, verified not mine** — these three fail identically at `HEAD~1` (= `origin/master`) on my machine and **pass in CI on this PR**, so they are local-environment, not in-branch:
- `test/retrieval-reflex.test.ts` — `logDeliveredReflexPointers logs channel=reflex events through the drained sink`
- `test/mcp-eval-capture.test.ts` — `captures MCP query call with remote=true`, `captures search call with tool_name="search"`

Also pre-existing and unrelated: running `hybrid-reranker-integration.serial` and `autocut-integration.serial` **in the same bun process** fails on embedding-dim bleed (the `#1527` hazard the file's own comment describes). Both pass individually, which is how the shard runner invokes `.serial.` files; reproduced identically at `HEAD~1`.

`node_modules` was installed with `bun install` in the worktree — never copied or symlinked.


time-attack:
Closing — the maintainer's decision is to leave the default reranker as-is.

Recording the consequence plainly so it isn't a surprise later: `zeroentropyai:zerank-2` remains the default, and ZeroEntropy's hosted API shuts down **2026-09-04**. After that date, every brain on the default reranker will fail its rerank call.

The good news is that this degrades softly rather than breaking search: a failed rerank falls back to the RRF fusion order, so results still come back — just without the cross-encoder re-sort. That's a quality regression, not an outage, which is presumably why it's an acceptable trade.

Two follow-ups this decision makes more important, not less:

1. **#3548** (open) warns brains pinned to a shutting-down provider. Its author noted that the warning currently only fires for a **config-set** `search.reranker.model`, not for brains on the shipped default — which, after this decision, is now the population that matters most. That gap should be closed before September.
2. Anyone who wants reranking to keep working still has options: self-host `zerank-2` (its weights are Apache-2.0, so llama-server or Ollama both work and preserve current behavior exactly), or set `search.reranker.model` to another provider explicitly.

The verification work here isn't wasted — the live Cohere test caught a real defect (one of the three models we'd have advertised is unreachable on a free account), and that finding is documented in `docs/ai-providers/llama-server-reranker.md` regardless of what the default is.

The trial Cohere key I created for testing has been deleted at the maintainer's instruction.

PR #3494, opened 2026-07-28
fix(extract): preserve facts on parse warnings

cybernaut6404:
## Summary

- treat warning-bearing fact-fence parses as non-authoritative for reconciliation
- preserve the affected page's existing indexed facts instead of interpreting skipped malformed rows as deletions
- continue reconciling clean pages in the same cycle

## Root cause

`parseFactsFence` deliberately skips malformed rows and returns the rows it can recover alongside warnings. `runExtractFacts` surfaced those warnings but still reconciled against the partial row set, so skipped rows looked removed and triggered page-level wipe/reinsert.

## Verification

- failing reproduction before fix: 20 passed, 1 failed; malformed row caused an unexpected insert after destructive reconcile
- targeted regression after fix: 21 passed, 0 failed
- `git diff --check`

### DIFF
--- a/src/core/cycle/extract-facts.ts
+++ b/src/core/cycle/extract-facts.ts
@@ -17,11 +17,12 @@
  *      DB rows need cleanup (#1781 — the unconditional wipe-and-reinsert
  *      made every cycle non-idempotent, re-appending duplicate rows).
  *
- * After the phase, the DB index for every affected page matches the
- * fence's canonical (claim, source) row set (modulo embeddings +
- * runtime-derived fields). Pages with no fence wipe DB rows for that
- * page coordinate only; legacy NULL-source_markdown_slug rows survive
- * because deleteFactsForPage targets source_markdown_slug = slug only.
+ * After the phase, the DB index for every cleanly parsed affected page
+ * matches the fence's canonical (claim, source) row set (modulo embeddings
+ * + runtime-derived fields). Warning-bearing parses are non-authoritative
+ * and preserve that page's existing index. Pages with no fence wipe DB rows
+ * for that page coordinate only; legacy NULL-source_markdown_slug rows
+ * survive because deleteFactsForPage targets source_markdown_slug = slug only.
  *
  * Empty-fence guard (Codex R2-#7; #2484): the phase refuses to do its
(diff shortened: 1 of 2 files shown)


time-attack:
Merging. Verified the defect and the fix independently.

`src/core/cycle/extract-facts.ts:301-306` on master pushes `parsed.warnings` and then falls through with the *partial* row set. `parseFactsFence` `continue`s past every malformed row (11 distinct warn+continue sites in `src/core/facts-fence.ts:170-297`), so at `:346` the skipped row's DB row reads as absent-from-desired, `:368` wipes the page, and `:372` reinserts only what parsed. The previously-indexed fact whose fence row got typo'd is deleted from the index while the fence still asserts it.

Reproduced: your test against master `src/` fails with `factsInserted` received 2, expected 1 — master wipes and reinserts.

Two things I checked because they'd have been blockers:
- The freeze is **page-local**, not run-global (`continue`, not `return`), and your test pins that a clean sibling page still reconciles in the same cycle. Confirmed.
- It's **visible**, not silent — `src/core/cycle.ts:1166` sets phase `status: 'warn'` whenever warnings exist.

**Stating the cost plainly so it's on the record:** a page with one malformed row now indexes no new facts from that fence until a human fixes the markdown, where master indexed the good ones. That's trading silent deletion for visible stalling, which is the right side of the trade on a destructive path — but it is a behavior change. The one case where this is strictly worse than master is `FACTS_ROW_NUM_COLLISION` / `row before header`, where master self-corrected via wipe+reinsert and this now freezes pending a human edit. Low likelihood given `renderFactsTable` round-trips rowNums, but real.

**Non-blocking follow-up worth sending:** the `continue` sits before `if (parsed.facts.length > 0) result.pagesWithFacts += 1`, so warning pages silently drop out of `pagesWithFacts` and out of `--dry-run` counts, and there's no `pagesSkippedMalformed` counter — the result object gives no explicit "I froze N pages" signal, only prose warnings. A counter would make this properly observable.

Also confirmed this and #3474 don't conflict — merged both directions with exit 0, and the combined suite is 26 pass / 0 fail with typecheck clean.

Issue #3466, opened 2026-07-28
inferTypeByDir asserts works_at from people/->companies/ adjacency with no evidence (19k false employment claims on a non-startup brain)


ethanbeard:
## Summary

`inferTypeByDir` (`src/commands/extract.ts:347`) types **every** markdown link from a `people/` page to a `companies/` page as `works_at`, with no evidence required. It cannot do otherwise: it receives `dirname(target)` — the string `"companies"` — so it has no way to know *which* company it is typing, or whether the page said anything about employment.

On a brain whose `people/` pages aren't startup bios, this writes false facts at scale. On ours (an art brain) it produced **19,497** claims of the form:

- `Chris Ofili works_at Hayward Gallery` — a gallery that holds his work in its collection
- `Robert Rauschenberg works_at Artforum` — a magazine that reviewed him
- `Marina Abramović works_at Le Centre d'Art`, `Laura Aguilar works_at LACMA`, `Rose B. Simpson works_at MASS MoCA` …

Every one is an artist typed as an employee of an institution that merely exhibits or collects them.

It also mislabels genuine employees. A person page that links a company they *visited once* gets `works_at` for it too, because the rule never looks past the directory pair.

## The asymmetry

The `founded` branch immediately above is evidence-gated — it requires the frontmatter to actually say so:

```js
if (from === 'people' && to === 'companies') {
  if (Array.isArray(frontmatter?.founded)) return 'founded';   // requires evidence
  return 'works_at';                                           // fires when there is none
}
```

`works_at` is the **fallback when that check fails** — so the specific factual claim is asserted precisely when no evidence exists for anything. That ordering seems inverted: the unevidenced case should get the weakest verb, not a specific one.

## Reproduction

Minimal, and already in your suite (`test/extract.test.ts:100`):

```js
const content = 'See [Brex](../companies/brex.md).';
const allSlugs = new Set(['people/pedro', 'companies/brex']);
const links = await extractLinksFromFile(content, 'people/pedro.md', allSlugs);
expect(links[0].link_type).toBe('works_at');   // passes today
```

The content contains no employment language of any kind. The existing test therefore encodes the behavior rather than a contract, which is likely why this has gone unnoticed.

[...]

PR #3495, opened 2026-07-28
fix(extract): inferTypeByDir no longer asserts works_at without evidence (#3466)

RerankerGuo:
## Summary

This PR fixes issue #3466: the directory-based link-type inference function `inferTypeByDir` defaulted to `works_at` for every people→companies adjacency, producing **19k+ false employment claims** on non-startup brains where people pages simply linked to companies without any employment relationship.

## Root Cause

`inferTypeByDir` (in `src/commands/extract.ts`) had no evidence threshold — it asserted `works_at` purely from the directory path pattern (people/ → companies/), unlike the canonical `inferLinkType` in `link-extraction.ts` which only emits a typed edge after a regex match on explicit context text.

## Changes

1. **Default people→companies links to `mentions`** instead of `works_at` — matching the evidence-based policy of `inferLinkType`
2. **Only infer `works_at`** when frontmatter carries explicit employment signals: `role`, `title`, `position`, `company`, `employer`, `employee` fields
3. **Keep `founded` inference** (via `frontmatter.founded` array) as before
4. **Export `inferTypeByDir`** for unit testability
5. **Add comprehensive test suite** (`test/infer-type-by-dir.test.ts`) with 14 test cases covering:
   - Default behavior (mentions without evidence)
   - Founded inference (with founded array)
   - Works_at inference (with employment signals)
   - Signal precedence (founded > employment signals)
   - All other adjacency types (deals, meetings, unrecognized)

## Impact

Prevents mass knowledge graph corruption on non-startup brains where people pages reference companies in non-employment contexts (e.g., investment, advisory, competition). The fix aligns fs-source inference with the evidence-based policy already used by db-source `inferLinkType`.

## Testing

- [x] Added 14 unit tests for `inferTypeByDir` covering all adjacency types and signal combinations
- [x] Regression test for the specific issue #3466 scenario (people→companies without evidence → mentions)
- [x] All existing adjacency types preserved (deals, meetings, unrecognized)

## Checklist

- [x] Backward compatible — only changes default behavior, no API breakage

### DIFF
--- a/src/commands/extract.ts
+++ b/src/commands/extract.ts
@@ -343,13 +343,36 @@ export function resolveSlugAll(
  * v0.13: aligned type names with link-extraction.ts (was: 'mention' →
  * 'mentions', 'attendee' → 'attended'). Diverged historically; the v0_13_0
  * migration normalizes any legacy rows on existing brains.
+ *
+ * v0.42.70: fix #3466 — the people→companies adjacency no longer
+ * asserts `works_at` without evidence. Previously every link from a
+ * people/ page to a companies/ page defaulted to `works_at`, producing
+ * 19k+ false employment claims on non-startup brains. The function now
+ * only returns `works_at` when the frontmatter carries explicit signals
+ * (role / title / position / company / employer / employee fields),
+ * defaulting to `mentions` otherwise — matching the evidence-based
+ * policy of `inferLinkType` in link-extraction.ts, which never
+ * asserts a typed edge without a regex match.
  */
-function inferTypeByDir(fromDir: string, toDir: string, frontmatter?: Record<string, unknown>): string {
+export function inferTypeByDir(fromDir: string, toDir: string, frontmatter?: Record<string, unknown>): string {
   const from = fromDir.split('/')[0];
   const to = toDir.split('/')[0];
   if (from === 'people' && to === 'companies') {
     if (Array.isArray(frontmatter?.founded)) return 'founded';
-    return 'works_at';
+    // issue #3466: only infer `works_at` when frontmatter carries
+    // explicit employment signals (role, title, position, company,
+    // employer, employee). Otherwise default to `mentions`.
+    if (frontmatter) {
+      const hasEmploymentSignal =
+        frontmatter.role !== undefined
+        || frontmatter.title !== undefined
+        || frontmatter.position !== undefined
+        || frontmatter.company !== undefined
+        || frontmatter.employer !== undefined
+        || frontmatter.employee !== undefined;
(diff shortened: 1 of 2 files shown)


time-attack:
Closing. #3466 is one of the best-evidenced issues in the tracker (19,497 rows), and your root-cause read is right — but this fix breaks a test it didn't update, and the evidence gate reintroduces the bug it's meant to fix.

## It breaks an existing test

`bun test test/extract.test.ts` on this branch: **28 pass / 1 fail**, EXIT=1 — `✗ extractLinksFromFile > infers link type from directory structure`. The issue text names that exact test (`test/extract.test.ts:100`). CI would be red.

## The `hasEmploymentSignal` gate reintroduces the reported bug

Ran it on your branch — a person page with `company: acme` linking both Acme and a gallery they merely exhibited at:

```
C (company: acme): [ "companies/acme=works_at", "companies/hayward-gallery=works_at" ]
```

Both get `works_at`. The signal is page-level, so it licenses `works_at` for *every* company that page links, which is exactly the failure mode you're fixing. Worse, `company:` **already** produces the correct target-specific edge through `FRONTMATTER_LINK_MAP` (`src/core/link-extraction.ts:797` — `{ fields: ['company','companies'], type: 'works_at' }`), so the clause adds nothing but false positives.

`role:` has the identical shape and is very likely on the reporter's art brain (`role: Artist`), so **most of the 19,497 false claims would survive this fix.**

## The `title` clause is dead code

`parseMarkdown` hoists `title` and `type` out of frontmatter — observed `parsed.frontmatter keys: ["role"]` for a page carrying all three. So at the only call site that clause can never fire, yet the PR ships `returns works_at when frontmatter has title` as a passing intent test and exports the function, making that non-behavior public surface.

## It fixes the leak but not the flood, without saying so

`addLinksBatch` is additive (`ON CONFLICT DO NOTHING`, `src/commands/extract.ts:1681`), and `LINK_EXTRACTOR_VERSION_TS` (`src/core/link-extraction.ts:34`) isn't bumped — whose own doc comment at `:19-23` says to bump it "whenever the shape of … inference changes meaningfully" so stamped pages re-extract. So already-stamped pages never re-extract and every existing false `works_at` edge persists forever. No backfill, no watermark bump, no mention in the body. A user reading this PR would reasonably think their 19k bad edges get cleaned up.

## What I'd take

The reporter's own suggestion: a bare `return 'mentions'` for people→companies. That's smaller, correct, and doesn't invent an evidence heuristic that leaks. Plus updating `test/extract.test.ts:100` and bumping `LINK_EXTRACTOR_VERSION_TS` so stamped pages actually re-extract.

PR #3496, opened 2026-07-28
fix(chunker): CJK-dense content gets token-aware char cap (issue #3037)

RerankerGuo:
## Summary

This PR fixes issue #3037: the markdown chunker capped chunks by **characters** (default 6000) while the embedder limited by **tokens**. For English text (~3.5 chars/token) this was safe (~1714 tokens), but for CJK text (~1 char/token) a 6000-char chunk was ~6000 tokens — dangerously close to the 8192-token embedder limit. CJK-dense content was **silently rejected** (exit 0, never embedded), causing permanent retrieval gaps.

## Root Cause

The `capByChars()` function in `src/core/chunkers/recursive.ts` used a fixed `maxChars` with no awareness of the different char-per-token ratios between languages. The chunker's word-counting was already CJK-aware (via `countCJKAwareWords`), but the safety-belt character cap was not.

## Changes

1. **Added `computeEffectiveMaxChars()`** function that detects CJK density and scales the char cap proportionally:
   - CJK density < 0.30: keep original `maxChars` (Latin-dominant, no behavior change)
   - CJK density ≥ 0.30: scale by `CJK_DENSITY_THRESHOLD / density`
   - At pure CJK density (1.0): 6000 × 0.30 ≈ **1800 chars** ≈ 1800 tokens, well under the 8192-token limit

2. **Applied the effective cap** to BOTH code paths in `chunkText()`:
   - Single-chunk fast path (word count ≤ chunkSize)
   - Multi-chunk sliding window path

3. **Bumped `MARKDOWN_CHUNKER_VERSION`** 3→4 to force post-upgrade re-chunk of existing pages with CJK-dense content

4. **Added 3 new test cases**:
   - Pure CJK text: chunks capped at ~1800 chars (vs old 6000)
   - High-density CJK: properly produces multiple chunks
   - English text: keeps original 6000-char cap unchanged

## Testing

- [x] Updated `MARKDOWN_CHUNKER_VERSION` test to expect version 4
- [x] Added 3 new test cases for CJK token-aware cap behavior
- [x] Existing English chunking behavior is unchanged (verified by regression test)

## Impact

**Prevents silent embedding failures** for CJK-dense content. The fix:
- Ensures CJK chunks stay well under the embedder's token limit
- Maintains backward compatibility for English-dominant text
- Requires re-chunking of CJK-heavy pages after upgrade (signaled by chunker_version bump)

## Checklist

- [x] Backward compatible — English text behavior unchanged

### DIFF
--- a/src/core/chunkers/recursive.ts
+++ b/src/core/chunkers/recursive.ts
@@ -14,10 +14,17 @@
  * guarantees no chunk overflows OpenAI's 8192-token embedding limit even
  * on pathological CJK / whitespace-less text.
  *
+ * v0.42.70 (issue #3037): CJK-dense content gets a token-aware cap.
+ * The embedder limits by TOKENS not characters; for CJK text 1 char ≈ 1
+ * token, so a 6000-char CJK chunk can be ~6000 tokens and get silently
+ * rejected (exit 0, never embedded). The chunker now detects CJK density
+ * and applies a tighter char budget (3000 chars for CJK-dominant text)
+ * so chunks stay under the typical 8192-token limit.
+ *
  * Lossless invariant: non-overlapping portions reassemble to original.
  */
 
-import { countCJKAwareWords, CJK_SENTENCE_DELIMITERS, CJK_CLAUSE_DELIMITERS } from '../cjk.ts';
+import { countCJKAwareWords, CJK_SENTENCE_DELIMITERS, CJK_CLAUSE_DELIMITERS, CJK_DENSITY_THRESHOLD, hasCJK } from '../cjk.ts';
 
 /**
  * Markdown chunker version. Folded into the per-page chunker_version column
@@ -34,7 +41,7 @@ import { countCJKAwareWords, CJK_SENTENCE_DELIMITERS, CJK_CLAUSE_DELIMITERS } fr
  * post-upgrade reembed sweep. See
  * `src/core/contextual-retrieval-service.ts`.
  */
-export const MARKDOWN_CHUNKER_VERSION = 3;
+export const MARKDOWN_CHUNKER_VERSION = 4;
 
 const DELIMITERS: string[][] = [
   ['\n\n'],                          // L0: paragraphs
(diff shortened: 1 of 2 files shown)


time-attack:
Closing. The bug is real, but this fix cannot fire on the input that was reported — it's gated in the wrong direction.

**The measurement.** #3037's reproduction is a 6,001-char chunk with 942 CJK chars. That's a CJK density of ~942/5000 ≈ **0.19**, which is *below* `CJK_DENSITY_THRESHOLD = 0.30` (`src/core/cjk.ts:41`), so `computeEffectiveMaxChars` returns `maxChars` unchanged (`recursive.ts:154`). I reproduced it end-to-end on a 6,001-char / 954-CJK bilingual doc:

```
master:  maxChars=6000  maxTok=2207
this PR: maxChars=6000  maxTok=2207   (identical)
```

Same for #2826's URL-dense Korean case (density 0.205) — identical on both.

**And the gated path is unreachable in production anyway.** All four `chunkText` call sites pass no options (`src/core/import-file.ts:673,678`, `src/commands/embed.ts:561,566`), so `chunkSize=300`. For density ≥ 0.30, `countCJKAwareWords` (`cjk.ts:57-68`) counts *characters*, so chunks land far under 6,000: pure CJK 20K chars → 67 chunks, max **300** chars; density-0.60 Korean+URLs 13K → 24 chunks, max **797** chars. So `maxChars` only binds when density < 0.30 — the regime this PR excludes. Your tests only pass because they pass `chunkSize: 100000`, which no caller uses; that's the tell.

**Meanwhile it isn't free.** `MARKDOWN_CHUNKER_VERSION` 3→4 drives the post-upgrade sweep (`src/core/post-upgrade-reembed.ts:52`, `reindex.ts:120,139`, `doctor.ts:1076`) — a full re-chunk and re-embed, with real API spend, for every install, in exchange for no change on real content. #2847/#3074 were closed over exactly this.

**Two smaller corrections.** The "silently rejected" framing isn't accurate — the error *is* printed at `src/commands/embed.ts:790`. What's actually silent is the exit code 0 and the per-page blast radius (#3037's problems 2 and 3), and this PR changes neither. And `effectiveMaxChars = maxChars * (0.30 / density)` reuses `CJK_DENSITY_THRESHOLD` — a word-counting *strategy switch* — as though it were a chars-per-token ratio; those aren't the same dimension. Real measured cl100k ratios: zh 0.83, ko 0.95, en 0.22. Density is also computed over the whole document, so a 25%-CJK page with one pure-CJK section gets nothing.

Minor: `recursive.ts:27` imports `hasCJK` unused; `recursive.ts:149` re-inlines the CJK ranges instead of using `cjk.ts`'s `CJK_SLUG_CHARS`; the docstring describes the reciprocal of what the code does.

**What would fix #3037.** Token-aware capping inside `capByChars` itself, using a weighted estimator moved into `src/core/cjk.ts` — no density gate, no version bump, applies in the density<0.30 regime where the bug actually lives. #3477 just landed a weighted estimator for the code chunker; lifting it to `cjk.ts` and calling it from `capByChars` is the natural next step and I'd take that.

PR #3497, opened 2026-07-28
fix(think): preserve federated source scope

cybernaut6404:
## Summary

- apply one source-precedence rule across think gather, trajectory entity resolution, trajectory reads, and citation persistence
- resolve federated trajectory candidates across the granted source array instead of collapsing to `default`
- persist synthesis evidence against the granted duplicate-slug page and pass operation scope into the save path

## Root cause

The federated source array reached gather and `findTrajectory`, but entity resolution used `opts.sourceId ?? 'default'`. Citation persistence accepted no source scope and queried duplicate slugs without `source_id`. These gaps could omit a granted trajectory or attach saved evidence to an arbitrary same-slug page.

## Verification

- failing reproduction before fix: 42 passed, 2 failed
- targeted source-scope suites after fix: 44 passed, 0 failed
- `git diff --check`

### DIFF
--- a/src/core/operations.ts
+++ b/src/core/operations.ts
@@ -2055,7 +2055,7 @@ const think: Operation = {
     let savedSlug: string | undefined;
     let evidenceInserted = 0;
     if (safeSave) {
-      const persisted = await persistSynthesis(ctx.engine, result);
+      const persisted = await persistSynthesis(ctx.engine, result, thinkScope);
       savedSlug = persisted.slug;
       evidenceInserted = persisted.evidenceInserted;
       for (const w of persisted.warnings) result.warnings.push(w);
--- a/src/core/think/index.ts
+++ b/src/core/think/index.ts
@@ -114,6 +114,24 @@ export interface RunThinkOpts {
   remote?: boolean;
 }
 
+type ThinkSourceScope = Pick<RunThinkOpts, 'sourceId' | 'allowedSources'>;
+
+function sourceIdsForThink(scope: ThinkSourceScope): string[] {
+  const allowed = scope.allowedSources?.filter(Boolean);
+  if (allowed && allowed.length > 0) return allowed;
+  return [scope.sourceId ?? 'default'];
+}
+
+function engineSourceScope(scope: ThinkSourceScope): {
+  sourceId?: string;
+  sourceIds?: string[];
+} {
+  const sourceIds = sourceIdsForThink(scope);
+  return scope.allowedSources?.length
+    ? { sourceIds }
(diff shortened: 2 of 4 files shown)


time-attack:
Closing — right diagnosis, but the fix introduces two regressions on the *primary local path*, both of which I reproduced. Please send a fresh one; the underlying bugs are real and worth fixing.

## The bugs you found are real

- `src/core/think/index.ts:213-218` — `SELECT id FROM pages WHERE slug = $1 LIMIT 1` with no `source_id`, so with duplicate slugs across sources, synthesis evidence attaches to an arbitrary source's page.
- `src/core/think/index.ts` ~:364 — `const sourceIdScalar = opts.sourceId ?? 'default'` ignores `opts.allowedSources`, so a federated caller's granted non-default entity never resolves.

Both confirmed on master. Neither is fixed elsewhere.

## What breaks

The problem is `engineSourceScope({})` → `sourceIdsForThink({})` returning `['default']`, where master's `...(opts.sourceId !== undefined ? ... : {})` produced `{}` — and `gather.ts:109-113` treats `{}` as *unscoped*. Since `src/commands/think.ts:143` calls `runThink` with no `sourceId` and no `allowedSources`, every plain `gbrain think` takes that path.

Two probes, same fixture, master vs this branch:

1. **`gbrain think` gather narrows from brain-wide to `default`.** Seeded a page + take only in source `work`, asked `"widgetco"`. Master: the evidence reaches the prompt. This branch: `only-in-work` is absent.
2. **`gbrain think --save` silently drops citation evidence.** `persistSynthesis(engine, result, scope = {})` → `persistCitations` gains `AND source_id = 'default'`; callers `src/commands/think.ts:148` and `src/core/cycle/auto-think.ts:166` pass no scope. Citation to a page existing only in `work` → master `evidenceInserted: 1`, this branch `0`.

That's the same "collapse to `default`" failure your own root-cause section indicts, moved from the federated path to the unscoped path. Your two new tests are real — they'd fail without the fix — but they only exercise `allowedSources: ['work']`, which is why "44 passed, 0 failed" is true and still misses this.

**The fix:** keep empty scope meaning *unscoped* — skip the `source_id` predicate when `sourceIds` is empty, rather than synthesizing `['default']`. With that change and a test for the no-scope local path, I'd take it.

One more thing worth knowing: #3470 (same wave) over-narrows the unscoped caller in a different place. Landing both would have cost multi-source users `source_id` over MCP *and* brain-wide `think` in one release, with neither PR's tests catching it, because both suites only assert the federated-array path. Worth adding an unscoped case to the shared scope-resolver suite so this class stops slipping through.

PR #3498, opened 2026-07-28
fix(eval): use QA rubric for gold-answer batches

herove:
## Summary

`gbrain eval cross-modal --batch` accepts the JSONL emitted by LongMemEval, but its implicit rubric was the generic agent-response rubric (`DEPTH`, `SOURCING`, and so on). That can fail a terse, factually correct hypothesis even when the row includes the benchmark's gold answer.

This change:

- adds one shared LongMemEval QA rubric (`CORRECTNESS` / `DIRECTNESS`);
- auto-selects it only when every row actually scored after `--limit` has a non-empty gold `answer`;
- preserves the generic rubric for no-gold and mixed batches;
- preserves an explicit `--dimensions` override;
- keeps the nightly probe's existing explicit QA dimensions through a back-compatible alias;
- documents and tests the selection matrix and gold-answer task construction.

Single-task cross-modal behavior, aggregate thresholds, and batch exit precedence are unchanged.

## Why

A real nightly fixture exposed the mismatch: a hypothesis equivalent to the gold answer could still fail on `DEPTH` or `SOURCING`. Those dimensions are appropriate for rich agent output, but not for closed-answer benchmark hypotheses.

The judge sees the question, gold answer, and hypothesis—not the source haystack—so the QA rubric deliberately avoids an unscorable grounding dimension.

## Selection matrix

| Batch input | Implicit rubric |
|---|---|
| Every scored row has non-empty `answer` | QA (`CORRECTNESS`, `DIRECTNESS`) |
| No scored row has `answer` | Generic five dimensions |
| Mixed gold / no-gold scored rows | Generic five dimensions |
| Explicit `--dimensions` | Caller-provided dimensions |

## Validation

- `bun test test/eval-cross-modal-batch.test.ts test/cycle/nightly-probe-adapters.test.ts` — 31 passed
- `GBRAIN_HOME=/tmp/gbrain-verify-lme-qa bun run verify` — 32/32 checks passed
- `bun run build:llms` — generated reference bundles remain unchanged

This is a Draft PR. It has not been merged or deployed.


time-attack:
Closing — the automated path this claims to fix already does the right thing on master.

`src/core/cycle/nightly-probe-adapters.ts:91` already defines `PROBE_QA_DIMENSIONS` (CORRECTNESS/DIRECTNESS) and threads it through `--dimensions`, and `src/commands/eval-cross-modal.ts:711-713` already folds the gold answer into the judge task. So for the nightly probe — the consumer that actually matters — there's no defect here. The real delta is auto-selecting the QA rubric for a human running `--batch` without `--dimensions`, plus a `.trim()` on `answer` and hoisting a constant.

**The design is the objection, not the size.** Making the rubric — and therefore the exit code — depend on whether the input data happens to carry gold answers means two batches silently stop being comparable. An explicit `--rubric qa` flag is both smaller and honest about what it changes. Implicit mode-switching on data shape is the kind of thing that produces a confusing eval receipt six months later.

Also: this collides with #3490 despite both bodies claiming independence — you both rewrite the region around `PROBE_QA_DIMENSIONS` / `runCrossModalBatchForProbe` and make contradictory edits to the same `docs/eval-bench.md` and `KEY_FILES.md` lines.

If you want the `.trim()` on `answer` as a standalone one-liner, that's fine to send.

PR #3302, opened 2026-07-23
reland: feat(recipes): add reranker touchpoint to OpenRouter (#2164)

time-attack:
Relands #2164 (squash commit 1a449bf5), which was auto-reverted in e20a6a53 when its merge BATCH went red on master CI — the revert swept the whole batch, not this change specifically.

Evidence #2164 is innocent:
- Recipe-only change: adds a reranker touchpoint to `src/core/ai/recipes/openrouter.ts` + one hermetic unit test. No gateway, search-layer, DB, or env changes.
- Cherry-picked cleanly onto current master (no conflicts).
- `bun run typecheck` — exit 0.
- `bun test` on every test file touching the diff or the reranker path (`openrouter-reranker-recipe`, `ai/recipe-openrouter`, `ai/gateway-cache-breakpoint`, `balanced-reranker-default`, `models-doctor-reranker`, `rerank-audit`) — 57 pass / 0 fail.

Original author: @Hippityy.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3499, opened 2026-07-28
fix(ai): enable OpenRouter query expansion

cybernaut6404:
## Summary
- declare OpenRouter's missing expansion touchpoint using cheap routed chat models
- prove arbitrary OpenRouter provider/model IDs resolve for expansion
- close the now-complete provider-agnostic and LiteLLM P2 backlog entries

## Evidence
- failing regression before fix: OpenRouter expansion touchpoint was undefined
- targeted AI tests: 104 passed, 0 failed
- focused final tests: 53 passed, 0 failed
- full local verify: 32 passed, 0 failed

No live provider request or credential was used.

### DIFF
--- a/TODOS.md
+++ b/TODOS.md
@@ -62,17 +62,20 @@ Deferred from the provider-agnostic plumbing wave (#1249/#1250/#1292/#2271/#2209
 Plan + review trail at `~/.claude/plans/system-instruction-you-are-working-keen-newell.md`.
 The eng-review + Codex outside-voice narrowed the wave to these deferrals:
 
-- [ ] **P2 — Capability-aware query expansion on OpenAI-compat providers (#2372).**
+- [x] **P2 — Capability-aware query expansion on OpenAI-compat providers (#2372).**
   Expansion only runs for recipes that declare an `expansion` touchpoint, and only the
   native providers (anthropic/openai/google) do. To make expansion work on
   litellm/openrouter/groq/together/deepseek you must ADD expansion touchpoints to those
   chat-capable recipes AND add a `generateObject`→`generateText` capability fallback for
   backends without strict structured outputs. Feature-shaped; overlaps the general
   OpenAI-compat proxy story (`docs/designs/COMMUNITY_IDEAS.md`). Community PR #2373 is a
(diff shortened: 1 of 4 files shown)


time-attack:
Merging. Verified rather than taken on faith, including the TODOS.md claims.

**Not a duplicate of #3302** — that reland added the *reranker* touchpoint (on master at `src/core/ai/recipes/openrouter.ts:208`); this adds *expansion*. Disjoint.

**Defect confirmed.** `isAvailable()` hard-requires the touchpoint (`src/core/ai/gateway.ts:909-910`: `if (!touchpointConfig) return false;`) and `expand()` bails at `:2467` (`if (!isAvailable('expansion')) return [query]`). Master's openrouter declared embedding/chat/reranker only, so expansion was a silent no-op. Swept all 24 recipes: openrouter was the **last** chat-capable openai-compat recipe missing it — deepseek `:89`, groq `:23`, together `:22`, litellm-proxy `:44`, mistral `:64`, moonshot `:22` all have it.

**No new code needed on the execution path** — openrouter is `openai-compatible` so it lands in the generic branch at `:2427-2437`, and since it declares no `supports_structured_outputs` it takes the `viaText` `generateText` + tolerant-parse path at `:2530-2533`. A 3-line recipe addition and nothing else, which is why this is an easy yes.

**TODOS.md flips check out.** #2373 (structured-output opt-in + capability-aware expansion fallback) merged 2026-07-27; #2208 (litellm chat + expansion touchpoints) merged 2026-07-17 and `litellm-proxy.ts:44` does declare expansion. Both defensible for their stated scope.

One caveat for the record: `nvidia`, `minimax`, and `zhipu` declare `chat` without `expansion`, so "expansion on OpenAI-compat providers" isn't universally true — but the TODO enumerated litellm/openrouter/groq/together/deepseek and that list is now complete, so the flip is fair.

Tests: `bun test test/ai/` → **468 pass / 0 fail** across 42 files. Your new assertion fails on master by construction.

Nits for next time, not blockers: the PR title should lead with the version per the repo's IRON RULE, `price_last_verified` with no `cost_per_1m_tokens_usd` is a date verifying nothing, and there's no CHANGELOG entry.

Issue #3500, opened 2026-07-28
Bug: dashscope_api_key from config is never folded into gateway env (missing mapping in build-gateway-config.ts, same class as #2662 Voyage fix)


NoahEleven:
## Description

In `src/core/ai/build-gateway-config.ts`, config-level API keys are folded into the gateway env so recipes can find them:

```ts
if (c.openai_api_key) envFromConfig.OPENAI_API_KEY = c.openai_api_key;
if (c.anthropic_api_key) envFromConfig.ANTHROPIC_API_KEY = c.anthropic_api_key;
if (c.zeroentropy_api_key) envFromConfig.ZEROENTROPY_API_KEY = c.zeroentropy_api_key;
if (c.openrouter_api_key) envFromConfig.OPENROUTER_API_KEY = c.openrouter_api_key;
if (c.voyage_api_key) envFromConfig.VOYAGE_API_KEY = c.voyage_api_key; // #2662
// ❌ missing: dashscope_api_key → DASHSCOPE_API_KEY
```

The DashScope recipe exists (`recipes/dashscope.ts`, embedding + rerank) and reads `DASHSCOPE_API_KEY`, but there is **no `dashscope_api_key` fold** — the exact same seam that was already fixed for Voyage in #2662 and for OpenRouter/ZeroEntropy.

## Impact

- `config.json` / `config set dashscope_api_key X` silently never reaches the DashScope recipe; embedding auth fails unless the user happens to have `DASHSCOPE_API_KEY` exported in the process env.
- Worse: if a **stale/invalid** `DASHSCOPE_API_KEY` is lingering in the OS env (very common for users who used the DashScope SDK before), it silently wins over the correct key in config, producing `Incorrect API key provided` that is very hard to trace back. I hit exactly this on Windows 11 and it took a long debugging session to find the missing fold.

## Repro

1. Put a valid `dashscope_api_key` in config, set `embedding_model: "dashscope:text-embedding-v3"`.
2. Make sure `DASHSCOPE_API_KEY` is absent (or set to an invalid value) in the process env.
3. Import any document → embedding call fails auth.

## Suggested fix

```ts
// build-gateway-config.ts, next to the voyage fold
if (c.dashscope_api_key) envFromConfig.DASHSCOPE_API_KEY = c.dashscope_api_key;
```

(plus the `dashscope_api_key` field in the config schema if not present yet.)

## Side suggestion

[...]

PR #3501, opened 2026-07-28
fix(eval): replace retired Gemini cross-modal slot

frxiaobei:
## Summary

- replace the retired `google:gemini-1.5-pro` cross-modal slot C default with `deepseek:deepseek-v4-pro`
- keep the default panel on three distinct providers: OpenAI, Anthropic, and DeepSeek
- update CLI help, Skillify guidance, architecture docs, and the mocked runner E2E
- pin the exact default panel plus recipe and canonical-pricing coverage

## Why

Google shut down Gemini 1.5 Pro on 2025-09-29. The old default therefore failed
slot C even after an operator supplied `GOOGLE_GENERATIVE_AI_API_KEY`.

DeepSeek v4 Pro is already registered by the DeepSeek v4 recipe migration and
has a canonical pricing entry, so the replacement keeps the independent
three-provider judge design without pointing at a dead model.

Google lifecycle reference:
https://ai.google.dev/gemini-api/docs/changelog#september_29_2025

## Validation

- `bun test test/cross-modal-default-slots.test.ts test/model-pricing.test.ts test/eval-cross-modal-batch.test.ts` (43 pass)
- `bun test test/e2e/cross-modal-eval.test.ts` (4 pass)
- `bun run typecheck`
- `HOME=/tmp/gbrain-crossmodal-verify-home bun run verify` (32/32)


time-attack:
Reviewed. The bug is real and the replacement model is a valid choice — but this patches one caller of a shared dead default, so the sibling surfaces stay broken. Please sweep them all.

**Confirmed.** `src/core/cross-modal-eval/runner.ts:54` still defaults slot C to `google:gemini-1.5-pro`, which Google has shut down. `deepseek:deepseek-v4-pro` is a sound swap — it's in the recipe chat touchpoint (`src/core/ai/recipes/deepseek.ts:95`) and in canonical pricing (`src/core/model-pricing.ts:101`), so the three-distinct-provider contract holds.

**Still pointing at the retired model after this PR:**
- `src/core/cycle/grade-takes.ts:222`
- `src/core/takes-quality-eval/runner.ts:39`
- `src/core/takes-quality-eval/pricing.ts:44`
- `src/core/ai/recipes/google.ts:26`

That last one is why the existing `cross-modal-default-slots.test.ts` guard passes today — the recipe still advertises the dead model as a valid chat target. Drop it from the recipe list and every dead default surfaces at once, which is the check you actually want.

Per CLAUDE.md the lazy fix is the root-cause fix: one sweep across all five sites is a smaller diff than five separate PRs, and it leaves no sibling caller broken.

Heads-up on ordering: this changes slot C's cost, which invalidates the `estimateCost(DEFAULT_SLOTS, …)` budget selector and the hard-coded "$2.00 for 10 questions" figure in #3490 (now closed).

time-attack:
Thanks — you correctly identified the dead slot-C default and picked the right replacement. This landed on master in 335e4703 (via the #3500/#3510 takeover, PR #3531), which uses your exact swap and credits this PR in the code comment. That commit also swept the sibling sites still pointing at the retired model (google recipe, takes-quality-eval runner/pricing, grade-takes), so the whole panel is consistent now. Closing as superseded — the fix you proposed is shipped and pinned by test/cross-modal-default-slots.test.ts and test/default-model-panels.test.ts.

PR #3503, opened 2026-07-28
v0.42.66.2 fix(cli): reject unknown operation flags (#380)

### LINKED ISSUE #380: `gbrain put --file <path>` silently ignored; flag is `--content`, unknown flags should error

javieraldape:
Fixes #380.

## Summary

- Reject unknown long flags in the shared operation CLI parser instead of forwarding or silently ignoring them.
- Reject extra short-flag-looking tokens once an operation has no positional slot left, covering cases like `gbrain list -n 500`.
- Preserve the existing global `--source <id>` routing path and dash-leading positional text for commands that accept free-form input.

## Tests

- `bun test test/cli-args.test.ts`
- `bun test test/cli-args.test.ts test/thin-client-source-scope.test.ts`
- Local gate: GREEN with `GBRAIN_VERIFY_TIMEOUT=300`; verify passed, focused units passed (17 pass), and diff-selected E2E resolved to the fail-closed all-set so it was skipped locally per gate policy.


time-attack:
**The parser fix is correct and I want it — but it can't land without a doc/skill sweep in the same PR, because 16 of the repo's own documented invocations start hard-erroring, including 3 shipped skills.**

## The bug is worse than "ignored", and your fix is in the right place

Master's `parseOpArgs` (`src/cli.ts:776-805`): an unknown flag with `paramDef === undefined` falls into `else if (i + 1 < args.length) { params[key] = args[++i]; }` — it **swallows the next token**. So `gbrain put slug --file /tmp/x.md` loses both the flag and the path, then falls through to the stdin read at `:808`, producing #380's ENXIO / empty-page confabulation. Fixing it in the one shared parser every op routes through is exactly right.

**What I verified is safe:** global flags are stripped by `parseGlobalFlags` *before* `parseOpArgs` (`src/cli.ts:222` → `src/core/cli-options.ts:54`), so `--quiet`, `--progress-json`, `--progress-interval[=]`, `--timeout[=]`, `--explain` are all fine. `--help`/`-h` short-circuits at `:305`. Your `--source` carve-out works (`{"query":"X","source":"wiki"}`). `--no-expand` still resolves to `expand:false`. Error UX is clean — `e.message`, exit 1, no stack trace. Tests: all 17 non-serial `test/cli-*.test.ts` plus `test/thin-client-source-scope.test.ts` → **167 pass / 0 fail**.

## What breaks

I ran every `gbrain <shared-op> … --flag` invocation in the repo's `docs/`, `skills/`, `README.md`, `CLAUDE.md`, and `TODOS.md` through your parser (excluding CLI_ONLY commands and `search modes|stats|tune|diagnose`, which never reach `parseOpArgs`). 16 newly throw:

```
THROW  query --json                    Unknown flag: --json
THROW  search --type/--sort            Unknown flag: --type
THROW  get --raw                       Unknown flag: --raw
THROW  query --brain                   Unknown flag: --brain
THROW  query -- <text>                 Unknown flag: --
```

The load-bearing ones are in **shipped skills that agents execute verbatim**:

- `skills/briefing/SKILL.md` — `gbrain search --type person --sort updated --limit 10`. `search` declares only `query, limit, offset, mode`. Today it silently returns unfiltered results; after this PR **briefing hard-fails**.
- `skills/brain-pdf/SKILL.md` — `gbrain get "$SLUG" --raw`.
- `skills/concept-synthesis/SKILL.md` and `skills/migrations/v0.29.1.md` (×3) — `gbrain query … --json`.
- `docs/ai-providers/llama-server-reranker.md`, `docs/tutorials/company-brain.md` (×2), `docs/architecture/brains-and-sources.md`, `docs/migrations/v0.41.2-markdown-greenfield.md`.

**And you surfaced a genuine #380-class bug in the process:** `--brain` is documented in CLAUDE.md as one of gbrain's two organizational axes, but it's **parsed nowhere** — `makeContext` reads only `params.source`. So `gbrain query "X" --brain media-team` silently queries the *wrong brain* today. Erroring is strictly better than that, but it's a documented flag going silent-wrong → fatal, which needs a decision rather than a side effect.

Also: `gbrain query -- text` now throws, because `arg.slice(2)` is `''` with no paramDef. The POSIX end-of-options terminator should keep working.

## To land it

Same PR, four additions:
1. Sweep those 16 sites — either implement the flags or fix the copy. `--json` on `query`/`search` is the one real users will hit; it's plausible that should just *work*.
2. Decide `--brain`: wire the routing or remove it from the docs.
3. Tolerate `--`.
4. `BREAKING CHANGE` note in CHANGELOG. A "did you mean … / see `--help`" hint in the error would be a nice touch.

The parser change itself I'd take as-is. Labeling this for a maintainer call since the `--brain` decision isn't mine to make unilaterally.

Separate incidental finding, pre-existing and not yours: `parseOpArgs`'s stdin read at `src/cli.ts:808` blocks forever on `readFileSync(0)` for stdin-ops in a non-TTY with no piped input. Filing that on its own.

time-attack:
Update: the blocker on this PR is resolved.

You'll recall this was held because rejecting unknown flags would make 16 documented invocations hard-fail, and the worst of those was `--brain` — advertised in CLAUDE.md as one of gbrain's two organizational axes but parsed nowhere, so `gbrain query "X" --brain media-team` silently searched the wrong brain.

The maintainer's decision was to **wire it up**, and that's now done in #3576: `--brain` is parsed by `parseGlobalFlags` and stripped before `parseOpArgs` — exactly like `--source` — so it no longer reaches your unknown-flag check at all. Unknown brain ids now error instead of silently falling back, matching the behavior #3524 established for `--source`.

That shrinks your blast radius. What still needs a sweep before this can land:

- `--json` on `query`/`search` — the most common one users will hit; appears in `skills/concept-synthesis/SKILL.md`, `skills/migrations/v0.29.1.md` (×3), `docs/ai-providers/llama-server-reranker.md`, `docs/migrations/v0.41.2-markdown-greenfield.md`
- `--type` / `--sort` on `search` — `skills/briefing/SKILL.md` runs `gbrain search --type person --sort updated --limit 10`, which would make briefing hard-fail
- `--raw` on `get` — `skills/brain-pdf/SKILL.md`
- `--remote` — `docs/tutorials/company-brain.md` (×2)
- `gbrain query -- <text>` — the POSIX end-of-options terminator currently throws, since `arg.slice(2)` is `''` with no paramDef

For each: either implement the flag or fix the copy. `--json` on the output-formatting ops is the one I'd guess should just work.

Also still wanted: a `BREAKING CHANGE` note in the CHANGELOG, and ideally a "did you mean … / see `--help`" hint in the error.

Worth saying again: the parser change itself is correct and I want it. 167 tests pass across all 17 non-serial `cli-*` files. It's the doc/skill drift around it that needs to land in the same PR so nobody's briefing breaks on upgrade.

time-attack:

The aim is right (#380: unknown flags silently swallowed, valued flags eat the next token) and the parser rejection + '--' end-of-options handling is genuinely tested. But the PR is CONFLICTING with master, has no VERSION bump (title claims v0.42.66.2, below master's 0.42.67.0; CHANGELOG entry filed under [Unreleased]), and the bundled --type feature violates two CLAUDE.md invariants: (1) it is a result-affecting search knob threaded into hybridSearchCached without a KNOBS_HASH_VERSION bump or hash component, so a --type-filtered lookup is served an unfiltered cached row (and vice versa); (2) engine parity is broken — Postgres applies singular type in searchKeyword/searchKeywordChunks/searchVector but PGLite (the default engine) only applies it in searchTitles, making --type a near-no-op on PGLite. The feature half has zero behavioral test coverage (all 8 new tests still pass with every type-threading site removed). The strict parser also hard-fails two documented invocations the PR's doc sweep missed, including 'gbrain query "X" --brain media-team' in the exact doc paragraph the PR edited to recommend passing --brain. Finally, 'get --raw' is accepted but provably identical to the default output — reintroducing the silently-ignored-flag class #380 complains about.

### Discrimination test

Surgically removed the four `type:` threading sites (operations.ts search/query/image paths + hybrid.ts searchOpts) while keeping the param declarations, then re-ran the PR's test file: 8/8 tests still green — the tests pin only parseOpArgs, never the filter behavior. The parser-rejection tests themselves DO discriminate (they assert toThrow; master's parser silently swallows), so the parser half is real. Separately executed the PR's parser against documented invocations: `query --brain` and `search --reranker` throw; `--source`, `--mode`, `--since`, `list --sort`, `--json`, `get --raw`, and global flags (--quiet/--timeout/--explain, stripped pre-dispatch) all still parse.

**Mergeability at review time:** mergeable=CONFLICTING, mergeStateStatus=DIRTY

### Blockers

**1.** Does not merge: gh reports mergeable=CONFLICTING / mergeStateStatus=DIRTY against master

**2.** Version gate: VERSION and package.json both 0.42.67.0 (equal to master, not strictly greater); CHANGELOG entry under [Unreleased] not a version header; PR title claims v0.42.66.2 which is below master's 0.42.67.0

**3.** Cache contamination: new `type` knob flows into hybridSearchCached on search (src/core/operations.ts:1654) and query (:1834) ops but src/core/search/mode.ts knobsHash (KNOBS_HASH_VERSION=14, unchanged) has no type component and skipCache (src/core/search/hybrid.ts:1833) does not exclude typed calls — `gbrain query "founders"` then `gbrain query "founders" --type person` (same embedding/source/hash) returns the cached UNFILTERED rows; violates the explicit CLAUDE.md KNOBS_HASH invariant

**4.** Engine parity broken: postgres-engine.ts applies singular `type` in searchKeyword (1659), searchKeywordChunks (1982), searchVector (2106); pglite-engine.ts applies it only in searchTitles (1740) — searchKeyword (1578) and searchVector (2055) ignore opts.type, so --type is a near-no-op on the default PGLite engine

**5.** Discrimination test failure (feature half): with all four `type:` threading sites removed from operations.ts + hybrid.ts, all 8 new tests in test/cli-args.test.ts still pass — zero behavioral coverage of the filter

**6.** Strict parser breaks documented invocations the PR's own sweep missed: docs/architecture/brains-and-sources.md:130 `gbrain query "X" --brain media-team` now throws 'Unknown flag: --brain' — and the PR's OWN edit to that paragraph (lines 131-133) recommends passing --brain; also docs/ai-providers/llama-server-reranker.md:149 `gbrain search ... --reranker ...` now hard-fails (verified by executing parseOpArgs on the PR branch)

**7.** src/cli.ts:1190 `get --raw` is accepted but provably a no-op (renderCliResult raw branch returns the identical formatResult as the default) — a silently-ignored flag, the exact #380 bug class

### Not verified

- Cache-contamination repro against a live DB (reasoned from code: knobsHash components, skipCache conditions, and the hit path's lack of post-filtering; not executed end-to-end against PGLite/Postgres per execution-footprint rules)
- PR body's 'local gate GREEN with GBRAIN_VERIFY_TIMEOUT=300' claim (did not run bun run verify — prohibited; and the branch is CONFLICTING with master anyway)
- Full doc-surface coverage: the invocation scanner skips templated lines (${...}, $(...), <placeholders>) and multi-line heredocs, so more breakages than the 2 found may exist
- test/e2e/engine-parity.test.ts and test/build-llms.test.ts not run (footprint; parity gap confirmed by direct code inspection instead)

- `gh pr view 3503 --json number,title,body,mergeable,mergeStateStatus,headRefName,files`
- `gh pr diff 3503 > /tmp/pr3503.diff (read every hunk)`
- `gh pr view 3503 --json mergeable,mergeStateStatus (re-poll: CONFLICTING/DIRTY)`
- `git fetch origin pull/3503/head:pr-3503 && git checkout pr-3503`
- `cat VERSION; node -e require('./package.json').version; grep '^## \[' CHANGELOG.md`
- `bun install`
- `bun probe3503.ts — parseOpArgs execution probe of 8 documented invocations on the PR branch`
- `bun run typecheck > /tmp/pr3503_tc.txt (EXIT=0)`
- `bun test test/cli-args.test.ts (8 pass) and test/thin-client-source-scope.test.ts (13 pass), output to files`
- `discrimination: sed-removed all `type:` threading sites, re-ran test/cli-args.test.ts (8/8 still pass), then git checkout restored`
- `bun scan3503.ts — scanned 38 shipped shared-op invocations in skills/ + docs/ through the PR parser (CLI_ONLY commands excluded): 2 hard failures`
- `grep audits: knobsHash components in src/core/search/mode.ts, skipCache/cacheScopeKey in hybrid.ts, opts.type handling in both engines' searchKeyword/searchVector/searchTitles, parseOpArgs callers, parseGlobalFlags stripping, brain-resolver importers (test-only), llms-full.txt staleness for edited docs`

</details>

---

time-attack:

The core fix is real and unsuperseded: origin/master src/cli.ts:791-823 still silently swallows unknown flags plus their next token (issue #380, labeled verified-real; no wave commit touches it), and the PR's 6 new tests fail against master's parser and pass with the fix. But the PR threads a `type` page-filter into hybridSearchCached in both the search and query op handlers (src/core/operations.ts) without folding it into the cache key — knobsHash (src/core/search/mode.ts:823-930, v=15 has no type component), cacheScopeKey (hybrid.ts:2056), and skipCache (hybrid.ts:1834) all ignore it — so within the default 3600s TTL a `gbrain search "founders" --type person` write is served to an unfiltered `gbrain search "founders"` lookup and vice versa, exactly the [CDX-4] contamination class this repo has bumped KNOBS_HASH_VERSION for four times. Additionally the PR's own doc edit (docs/architecture/brains-and-sources.md:132-133) recommends `--brain media-team`, which the new parser hard-rejects on `query` (brain is not an op param and not special-cased), docs/ai-providers/llama-server-reranker.md:149's `gbrain search ... --reranker` example now hard-fails, the `get --raw` special case in renderCliResult (src/cli.ts) is byte-identical to the fallthrough (a silently-accepted no-op flag — the exact class the PR eliminates), and `--dry-run` (read into ctx.dryRun at cli.ts:1054 by put/delete/tag/link ops that don't declare dry_run) now hard-errors instead of being special-cased like `source`. Fixes needed: fold type into knobsHash (v=16) or add it to skipCache; sweep the four remaining doc invocations; implement or drop `get --raw`; special-case dry_run.

**Conflict state vs current master:** Only CHANGELOG.md conflicts with current master (both added top entries) — trivial context drift, resolved by keeping both. src/cli.ts, src/core/operations.ts, src/core/search/hybrid.ts, and test/cli-args.test.ts all auto-merge cleanly; the wave's cli.ts changes (#3652 init --help, #3650 sync --timeout, #3546 stdin) are in the dispatch region, orthogonal to parseOpArgs. Merged tree: typecheck green, PR's focused tests 21/21 pass, test/search/ 341/341 pass, build-llms freshness passes; 6 serial-test failures reproduce identically on the master baseline (pre-existing). The fix fully applies post-v0.42.69.0 — nothing in the wave addressed #380.

**Risk if merged as-is:** Two behavioral risks if merged as-is: (1) query-cache cross-contamination between --type-filtered and unfiltered searches (silent wrong results, the repo's documented must-fix class); (2) hard-error regressions on documented invocations the PR missed — `query --brain` (recommended by the PR's own doc edit), `search --reranker` (llama-server-reranker.md), bare `purge-deleted --older-than` (op param is --older-than-hours), `graph --type` (param is --link-type), and `--dry-run` on put/delete/tag/link ops — all silently ignored today, fatal after this PR. CI itself stays green (all failures observed are pre-existing on master).

javieraldape:
Updated this branch to address the review feedback in commit `4b06ccb8`.

What changed:
- Folded the page type filter into the search/query cache key by adding `pageType` to `KnobsHashContext` and bumping `KNOBS_HASH_VERSION` to 16. This prevents filtered and unfiltered query-cache rows from colliding within the TTL window.
- Threaded `SearchOpts.type` into the `hybridSearchCached` `knobsHash` context.
- Kept the stricter unknown-flag parser while preserving documented/shared CLI behavior:
  - `--dry-run` is accepted as a shared CLI flag for mutating ops that read `ctx.dryRun`.
  - `purge-deleted --older-than` maps to `older_than_hours` and accepts `h`/`d` duration suffixes.
  - `graph --type` remains accepted as a compatibility alias for `--link-type`.
- Implemented `get --raw` as body-only markdown output instead of a no-op alias for the default renderer.
- Swept the stale docs examples for `search --reranker`, bare `purge-deleted --older-than`, and `graph --type`; regenerated the skills manifest after the skill doc edit.

Validation run locally:
- `bun test test/cli-args.test.ts` — 11 pass
- `bun test test/search/knobs-hash-reranker.test.ts test/search-mode.test.ts test/search-alias-resolved-boost.test.ts test/cross-modal-phase1.test.ts` — 141 pass
- `bun test test/build-llms.test.ts` — 12 pass
- `bun run typecheck` — pass

CI has started on the pushed head `4b06ccb8`; current PR status is still `UNSTABLE` only because checks are in progress.

garrytan:
Thanks — superseded by the CLI-wide strict flag registry that shipped in v0.42.76.0 (#3902): unknown op flags now fail loudly everywhere, with the docs/skills sweep handled registry-wide. Closing as superseded.

Issue #1261, opened 2026-05-21
recipes: add canonical m365-contacts-to-brain.md (Microsoft Graph contacts → people/ enrichment)


rayers:
## Summary

`recipes/` ships Google-side `email-to-brain.md` and `calendar-to-brain.md` with `requires: [credential-gateway]`. The Google Contacts analog is requested in #1099. There is **no canonical Microsoft 365 recipe family yet** (no `m365-mail-to-brain.md`, no `m365-calendar-to-brain.md`, no `m365-contacts-to-brain.md`).

This issue scopes the **contacts** half specifically, to parallel #1099. I'm filing rather than waiting because contacts is the lowest-friction first M365 recipe to bring upstream: smaller surface than mail, no licensing-tier complications like Teams channels.

## Why this matters

Same shape as #1099: calendar attendees and mail senders are the bulk of `people/<slug>` pages. A user's M365 contacts directory is the ground-truth name/email/phone/organization record for those entities. Without a canonical recipe, M365-first users either skip contacts entirely or maintain divergent custom collectors.

Concrete data point: I built a local `m365-ingest.py` (Azure AD device-code, Microsoft Graph `/me/messages` + `/me/chats/getAllMessages`) that pulls mail + chats into 320K m365-mail + 13K m365-chat brain pages. **Today I also built a signature extractor over those mail pages** that yields ~50% hit rate, ~470 unique (name, email, title?, company?, phone?) tuples per 1000-email sample — JSONL output at `~/.gbrain/m365-contact-signatures.jsonl`. Both prove the data surface works; what's missing is the canonical recipe so M365-first users don't have to write any of this themselves.

## Proposed shape

Mirror the Google recipes:

```yaml
---
id: m365-contacts-to-brain
name: M365-Contacts-to-Brain
version: 0.1.0
description: Microsoft 365 contacts become canonical people/ pages, enriching brain entities with ground-truth name/email/phone/org data.
category: sense
requires: [credential-gateway]
secrets:
  - name: AZURE_TENANT_ID
    description: Azure AD tenant ID (Option A — direct Microsoft Graph)
    where: Azure AD app registration overview page
  - name: AZURE_CLIENT_ID
    description: Azure AD app client ID (Option A)
    where: Azure AD app registration overview page
  - name: CLAWVISOR_URL
    description: ClawVisor gateway URL (Option B — recommended once ClawVisor adds M365 actions)
    where: https://clawvisor.com — agent settings
[...]

Issue #1099, opened 2026-05-17
recipes: add canonical contacts-to-brain.md (Google Contacts → people/ enrichment)


RyeMcKenzie81:
## Summary

`recipes/` ships `email-to-brain.md` and `calendar-to-brain.md`, both with `requires: [credential-gateway]`. `credential-gateway.md` already instructs users to activate Google Contacts alongside Gmail + Calendar:

> 3. Activate the services you need:
>    - **Gmail** (for email-to-brain)
>    - **Google Calendar** (for calendar-to-brain)
>    - **Google Contacts** (for enrichment)

And ClawVisor's catalog exposes `google.contacts` actions: `list_contacts`, `get_contact`, `search_contacts`. But there is no canonical `recipes/contacts-to-brain.md` to consume that surface.

## Why this matters

Calendar attendees and email senders are the bulk of `people/<slug>` brain pages. A user's Google Contacts is the ground-truth name + email + phone + organization directory for those entities. Linking contact records into the brain closes the "who is this person" loop automatically and feeds enrichment per [`skills/enrich/SKILL.md`](https://github.com/garrytan/gbrain/blob/master/skills/enrich/SKILL.md).

Without a canonical recipe, users either:
1. Skip contacts (lose enrichment data)
2. Build a custom collector (drift from the deterministic-collector pattern in [`docs/guides/deterministic-collectors.md`](https://github.com/garrytan/gbrain/blob/master/docs/guides/deterministic-collectors.md))

## Proposed shape

Mirror the structure of `email-to-brain.md` / `calendar-to-brain.md`:

```yaml
---
id: contacts-to-brain
name: Contacts-to-Brain
version: 0.1.0
description: Google Contacts become canonical people/ pages, enriching brain entities with ground-truth name/email/phone/org data.
category: sense
requires: [credential-gateway]
secrets:
  - CLAWVISOR_URL / CLAWVISOR_AGENT_TOKEN  (Option A)
  - GOOGLE_CLIENT_ID / GOOGLE_CLIENT_SECRET  (Option B)
health_checks: any_of (same pattern as siblings)
setup_time: 15 min
cost_estimate: "$0"
---
```

Workflow:
1. **Deterministic collector** — paginated `list_contacts` → JSON records → markdown staging files
2. **Step 4 enrichment** — agent merges contacts into existing `people/<slug>` pages (Notability Gate per `_brain-filing-rules.md`); creates new pages for notable contacts not already in brain
[...]

PR #3505, opened 2026-07-28
feat(recipes): add canonical contacts-to-brain recipe (#1099)

rayers:
Closes the gap in #1099. Sending it as the self-contained PR you invited there.

Credit where it is due: the shape is @RyeMcKenzie81's from the issue, including the frontmatter sketch and the four-part workflow. I implemented it because I had just hit the same gap from the Microsoft Graph side (#1261) and had the failure modes fresh.

## What this adds

`recipes/contacts-to-brain.md`, the missing sibling to `email-to-brain` and `calendar-to-brain`. `credential-gateway.md` already instructs users to activate Google Contacts "for enrichment" and lists the `contacts.readonly` scope, but nothing consumed it.

It follows the deterministic-collector split from `docs/guides/deterministic-collectors.md`: the collector owns slug derivation, alias normalization, pagination, `syncToken` state, and stable output; the agent owns the Notability Gate and merging into `people/<slug>` without clobbering prose.

## Two design decisions worth reviewing

**1. The alias map is the primary deliverable, not a side effect.** The recipe emits `aliases.json` mapping every email and phone to one slug, and it makes identity reconciliation Step 6a, explicitly BEFORE page creation in 6b.

The reasoning: contacts are low-volume and high-value precisely because they resolve ambiguity elsewhere. Without the map, `email-to-brain` sees `a.example@acme-example.com` and `alice@acme-example.com` as two strangers, `calendar-to-brain` files "A. Example" apart from "Alice Example," and you get three thin pages for one person. Every one of them then scores worse on search. Ordering reconciliation after enrichment means creating duplicates and merging them by hand, which is the mistake I made on my own contacts ingest.

The recipe also says to run it before backfilling email or calendar when the user has the choice, for the same reason.

**2. `otherContacts` feeds the alias map but never creates pages.** Google auto-collects an entry for everyone you have ever emailed. Those are excellent alias material and terrible brain entities, so the recipe harvests their addresses and files zero pages from them. It is a separate endpoint from `connections` with a narrower allowed `personFields` set, which the troubleshooting section calls out since "alias count equals contact count" is the symptom of getting it wrong.

## Production gotchas documented

These are the ones that bite silently rather than loudly:

- **`syncToken` expiry returns HTTP 410, not an empty result.** A collector that treats 410 as generic failure stops updating while the heartbeat still looks healthy. Also, the API rejects `syncToken` combined with a sort order.
- **Unstable output turns a free sync into a recurring embedding bill.** The People API does not guarantee array order between calls, so unsorted arrays or loose frontmatter key order rewrite every file weekly, which re-imports and re-embeds unchanged contacts forever.
- **`organizations[0]` files people at companies they left.** Prefer the entry with no `endDate`.
- **Deleted contacts must not delete pages.** Removing a phone number from an address book says nothing about whether the brain should forget the person, and the page usually holds history the contact record never had.
- **`personFields` is required**, not defaulted.

## Also changed

Registered the recipe in the four existing index sites so it is discoverable: `docs/GBRAIN_SKILLPACK.md`, `docs/integrations/README.md`, `docs/guides/cron-schedule.md`, and `RECIPE_META` in `src/commands/features.ts`. Regenerated `llms-full.txt`, which inlines `cron-schedule.md`.

The `features.ts` line is the only non-docs change, and it is one entry. It makes `gbrain features` offer the recipe like the other seven. Happy to drop it if you would rather keep this PR docs-only. I bumped the count in `test/features.test.ts`'s title from 7 to 8 since that assertion is a count-in-name only (the body just checks `runFeatures` is defined).

No new secrets, no new dependencies, no runtime behavior beyond that one offer entry. Recipes are discovered by directory scan in `getRecipeDirs`, so nothing else needed wiring.

## Verification

- `bun run typecheck` clean
- `bun test test/features.test.ts test/build-llms.test.ts` 21 pass / 0 fail (build-llms fails without the regenerated bundle, confirmed both directions)
- `bun test test/integrations.test.ts` 111 pass / 0 fail
- `scripts/check-key-files-current-state.sh` ok
- Recipe parses through the real `parseRecipe`: `id`, `name`, `category: sense`, and the `any_of` health-check block all resolve, matching the sibling recipes' shape
- Branched off `fd8be831`, no fork-local patches in the diff

Examples use the `alice-example` / `acme-example` placeholders per the privacy rule. No real contacts appear anywhere in the recipe.

### DIFF
--- a/docs/GBRAIN_SKILLPACK.md
+++ b/docs/GBRAIN_SKILLPACK.md
@@ -91,6 +91,7 @@ Wiring up your life.
 | [Email-to-Brain](../recipes/email-to-brain.md) | Gmail messages flow into entity pages via deterministic collector |
 | [X-to-Brain](../recipes/x-to-brain.md) | Twitter monitoring with deletion detection + engagement velocity |
 | [Calendar-to-Brain](../recipes/calendar-to-brain.md) | Google Calendar events become searchable daily brain pages |
+| [Contacts-to-Brain](../recipes/contacts-to-brain.md) | Google Contacts become canonical people/ pages plus an email-to-slug alias map |
 | [Meeting Sync](../recipes/meeting-sync.md) | Circleback transcripts auto-import with attendee propagation |
 
 ## Administration
--- a/docs/guides/cron-schedule.md
+++ b/docs/guides/cron-schedule.md
@@ -22,6 +22,7 @@ fixed. You wake up and the brain is smarter than when you went to sleep.
 | Every 30 min | X/Twitter collection | Create/update media pages, entity extraction | [x-to-brain](../../recipes/x-to-brain.md) |
 | 3x/day (weekdays) | Meeting sync | Full ingestion + attendee propagation | [meeting-sync](../../recipes/meeting-sync.md) |
 | Weekly | Calendar sync | Daily files + attendee enrichment | [calendar-to-brain](../../recipes/calendar-to-brain.md) |
+| 2x/week | Contacts sync | Staging files + alias map + people/ enrichment (syncToken expires in 7 days) | [contacts-to-brain](../../recipes/contacts-to-brain.md) |
 | Daily AM | Morning briefing | Search calendar attendees, deal status, active threads | [briefing skill](../../skills/briefing/SKILL.md) |
 | Weekly | Brain maintenance | `gbrain doctor`, embed stale, orphan detection | [maintain skill](../../skills/maintain/SKILL.md) |
 | Nightly | Dream cycle | Entity sweep, enrich thin spots, fix citations | See below |
--- a/docs/integrations/README.md
+++ b/docs/integrations/README.md
@@ -34,6 +34,7 @@ These are integration recipes your agent can set up for you. Run
 | [email-to-brain](../../recipes/email-to-brain.md) | Sense | credential-gateway | Gmail messages flow into entity pages via deterministic collector | 20 min |
 | [x-to-brain](../../recipes/x-to-brain.md) | Sense | — | Twitter timeline, mentions, keyword monitoring with deletion detection | 15 min |
 | [calendar-to-brain](../../recipes/calendar-to-brain.md) | Sense | credential-gateway | Google Calendar events become searchable daily brain pages | 20 min |
+| [contacts-to-brain](../../recipes/contacts-to-brain.md) | Sense | credential-gateway | Google Contacts become canonical people/ pages plus an email-to-slug alias map | 15 min |
 | [meeting-sync](../../recipes/meeting-sync.md) | Sense | — | Circleback meeting transcripts auto-import with attendee propagation | 15 min |
 
 ### Manual Integration Guides
--- a/llms-full.txt
+++ b/llms-full.txt
@@ -2894,6 +2894,7 @@ fixed. You wake up and the brain is smarter than when you went to sleep.
 | Every 30 min | X/Twitter collection | Create/update media pages, entity extraction | [x-to-brain](../../recipes/x-to-brain.md) |
 | 3x/day (weekdays) | Meeting sync | Full ingestion + attendee propagation | [meeting-sync](../../recipes/meeting-sync.md) |
 | Weekly | Calendar sync | Daily files + attendee enrichment | [calendar-to-brain](../../recipes/calendar-to-brain.md) |
+| 2x/week | Contacts sync | Staging files + alias map + people/ enrichment (syncToken expires in 7 days) | [contacts-to-brain](../../recipes/contacts-to-brain.md) |
 | Daily AM | Morning briefing | Search calendar attendees, deal status, active threads | [briefing skill](../../skills/briefing/SKILL.md) |
 | Weekly | Brain maintenance | `gbrain doctor`, embed stale, orphan detection | [maintain skill](../../skills/maintain/SKILL.md) |
 | Nightly | Dream cycle | Entity sweep, enrich thin spots, fix citations | See below |
--- a/recipes/contacts-to-brain.md
+++ b/recipes/contacts-to-brain.md
@@ -1,0 +1,744 @@
+---
+id: contacts-to-brain
+name: Contacts-to-Brain
+version: 0.1.0
+description: Google Contacts become canonical people/ pages, enriching brain entities with ground-truth name, email, phone, and organization data.
+category: sense
+requires: [credential-gateway]
+secrets:
+  - name: CLAWVISOR_URL
+    description: ClawVisor gateway URL (Option A, recommended, handles OAuth for you)
+    where: https://clawvisor.com, create an agent, activate Google Contacts service
+  - name: CLAWVISOR_AGENT_TOKEN
+    description: ClawVisor agent token (Option A)
+    where: https://clawvisor.com, agent settings, copy the agent token
+  - name: GOOGLE_CLIENT_ID
+    description: Google OAuth2 client ID (Option B, direct API access, you manage tokens)
+    where: https://console.cloud.google.com/apis/credentials, create OAuth 2.0 Client ID
+  - name: GOOGLE_CLIENT_SECRET
+    description: Google OAuth2 client secret (Option B)
+    where: https://console.cloud.google.com/apis/credentials, same page as client ID
+health_checks:
+  - type: any_of
+    label: "Auth provider"
+    checks:
+      - type: http
+        url: "$CLAWVISOR_URL/health"
+        label: "ClawVisor"
+      - type: env_exists
+        name: GOOGLE_CLIENT_ID
+        label: "Google OAuth"
+setup_time: 15 min
+cost_estimate: "$0 (both options are free)"
+---
+
+# Contacts-to-Brain: Your Address Book Becomes the Identity Layer
+
+Calendar attendees and email senders are the bulk of `people/<slug>` pages. Your
+contacts are the ground-truth directory for those entities: real names, every
+email address a person uses, phone numbers, and current organization. This recipe
+turns that directory into brain pages, and produces an email-to-slug map the agent
+uses to decide which pages are the same person.
+
+## IMPORTANT: Instructions for the Agent
+
+**You are the installer.** Follow these steps precisely.
+
+**Why this matters:** Contacts are low-volume and high-value. A few hundred
+records resolve thousands of ambiguous references. Without them, `email-to-brain`
+sees `a.example@acme-example.com` and `alice@acme-example.com` as two strangers,
+and `calendar-to-brain` files "A. Example" separately from "Alice Example." You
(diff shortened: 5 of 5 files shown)


time-attack:
Holding this for a maintainer decision rather than applying the default close-features policy, because it was explicitly invited: #1099 was closed with "if you or another contributor want to send `recipes/contacts-to-brain.md` as a self-contained PR following the deterministic-collector pattern, it will be reviewed on its merits." You did exactly that, disclosed the trade-offs, and honored the privacy rule throughout (`alice-example` / `acme-example`, no real contacts). Auto-closing invited work isn't right.

Review notes so the decision is easy:

- **`llms-full.txt` is genuinely generated, not hand-edited** — verified the added line is byte-identical to the `cron-schedule.md` line at the correct offset, which is what `bun run build:llms` produces. Claim checks out.
- **`GOOGLE_CONTACTS_API_KEY` in `src/commands/features.ts` is invented** — it appears nowhere else in the repo, and your own recipe frontmatter declares `CLAWVISOR_URL`/`CLAWVISOR_AGENT_TOKEN`/`GOOGLE_CLIENT_ID`/`GOOGLE_CLIENT_SECRET` instead. Since `features.ts:149` gates on `secrets.every(s => process.env[s])`, the recipe reads as permanently unconfigured however the user actually sets it up. (`GOOGLE_CALENDAR_API_KEY` has the same flaw on master, so this is consistent rather than novel — but it means that line adds noise, not function.)
- **The test change is decorative** — `covers all 7 recipes` → `all 8` while the body only asserts `runFeatures` is defined. You disclosed this; noting it so nobody reads it as coverage.
- **ClawVisor is established precedent, not a new vendor** — `calendar-to-brain`, `email-to-brain`, and `credential-gateway` already do the same on master, and your Option B is direct OAuth.

**The trim that would make this an easy yes:** docs-only, 3 files — the recipe plus the 3 index rows. Drop `src/commands/features.ts` and `test/features.test.ts` (and consequently `llms-full.txt`). That removes all runtime surface and the decorative assertion, and you already volunteered it in the body. Separately, the four production gotchas you documented (410-not-empty, stable output or a recurring embedding bill, `organizations[0]` past-employer, deleted-contacts-don't-delete-pages) apply to *every* deterministic collector — those are worth ~15 lines in `docs/guides/deterministic-collectors.md` regardless of what happens to this PR.

rayers:
Pushed. Took the trim, and then a lot more, because I put this through an eight-round adversarial review loop (Codex plus independent reviewers, looping until two consecutive rounds found nothing defendable) before touching it. That surfaced defects well past the packaging question, including four wrong Google People API claims that I had authored with confidence. Thank you for holding it instead of merging on the strength of a well-written gotchas section.

## Your review notes, all applied

- **`GOOGLE_CONTACTS_API_KEY` dropped**, along with `test/features.test.ts`. You were right that the gate at `features.ts:149` (`secrets.every(s => process.env[s])`) meant a fully configured install would read as unconfigured forever. Now docs-only.
- **Test change dropped**, so nothing pretends to be coverage.
- `llms-full.txt` verified generated, per your check.

## One correction to your review

The parenthetical "(and consequently `llms-full.txt`)" does not hold. The bundle line comes from `docs/guides/cron-schedule.md`, which `scripts/llms-config.ts` inlines; it has no relationship to `features.ts`. I tested both directions: reverting `llms-full.txt` while keeping the cron-schedule row fails `test/build-llms.test.ts`. So the diff is 5 files, not 3, and the fifth is generated. Still zero runtime surface.

## What the review loop found, all fixed

The API contract was wrong in four places. Each verified against `developers.google.com`, not memory:

| Claim I shipped | Actual |
|---|---|
| `otherContacts.list` uses `personFields`, scope `contacts.readonly` | **`readMask`**, scope **`contacts.other.readonly`** |
| One field mask for both endpoints | `otherContacts` under the default `READ_SOURCE_TYPE_CONTACT` allows only `emailAddresses, metadata, names, phoneNumbers, photos`; my mask had five illegal fields |
| API rejects `syncToken` + sort order | No such rule. The real constraint is that *all* params must match the originating call |
| (nothing) | **Sync tokens expire 7 days after full sync**, and my cron was exactly 7 days, so every "incremental" run would have taken the full-sync fallback and paid its quota |

On the expired-token gotcha I managed to be wrong twice. My original keyed on HTTP 410; a reviewer asserted 400; the reference actually pins no status at all, only `ErrorInfo.reason = EXPIRED_SYNC_TOKEN`, while Google's own contacts guide sample keys on 410. It now matches the reason and explains why keying on either bare status is unsafe. That is the specific claim you were going to have me generalize into `docs/guides/deterministic-collectors.md`, which would have propagated a wrong contract to every collector, so I have not sent that follow-up.

Design defects, also fixed:

- **The gate gated nothing.** Old Step 5 imported all of staging before the Notability Gate ran, so the "plumber" was already chunked and embedded. Staging now lives in `sources/contacts/` per `_brain-filing-rules.md:48`, is gitignored, and is **never imported**. Only `people/` pages that pass the gate reach the brain.
- **PII into a git repo**, with both escape hatches silently broken: gitignoring it made `gbrain import` enumerate zero files via `git ls-files --cached --others --exclude-standard`, and not gitignoring it committed the address book forever. There is now a Step 0 that resolves and validates `BRAIN_DIR` (via `gbrain config get sync.repo_path`, matching `skills/book-mirror/SKILL.md:86`), appends the ignore rule newline-safely and idempotently, and **asserts `git check-ignore` before the collector runs**.
- **The alias map deleted itself weekly.** `build_alias_map` started from `{}` while the cron passed only changed contacts, so each run dropped every prior mapping. Now load-merge-write, sorted keys, with a never-overwrite-the-incumbent rule and a deterministic `resourceName` tie-break for shared addresses.
- **Auto-collected contacts no longer claim slug pins** (different resourceName namespace, and they cannot return an organization to disambiguate a collision), and unresolved ones are queued as `resolved: false` for the agent instead of guessing.
- `+tag` stripping is now gated on Google domains, since elsewhere `alice+ops@` is a distinct mailbox and folding it merges two real people. Phone normalization is E.164 rather than "last 10 digits," which collided `+44 20 7946 0958` with `+1 207 946 0958`.
- Also corrected: I had claimed `gbrain sync` ignores uncommitted edits. It does not, `--others` covers untracked-not-ignored, pinned by `test/sync.test.ts`. Step 6 now gives an honest reason to commit and stages `companies/` too, since the back-link Iron Law applies.
- Dropped the false claim that other recipes consume `aliases.json`. `grep` finds no consumer; it is documented as an agent-read artifact.
- Four of the six original tests could not fail. There are now 10 that can, including the deletion-containment and PII-containment cases that had no coverage.

## Verification

`bun run typecheck` clean, `test/build-llms.test.ts` + `test/features.test.ts` + `test/integrations.test.ts` 132 pass / 0 fail, all 12 bash blocks pass `bash -n`, the recipe parses through the real exported `parseRecipe`, and every gbrain claim is cited to source in this repo. Rebased onto current master (`6920744d`).

Happy to keep iterating, or to close this if you would rather not carry another recipe. Either is a fine outcome from my side; I would rather it be correct than merged.

time-attack:
Closing this one. The maintainer's call, and I want to be straight with you about the reasoning rather than hide behind a policy line.

You did nothing wrong here. You were invited to send this, you followed the deterministic-collector pattern you were pointed at, you disclosed the trade-offs honestly in the body, and you respected the privacy rule throughout (`alice-example` / `acme-example`, no real contacts anywhere). I explicitly held this PR back from an automated close pass for exactly that reason, and flagged to the maintainer that auto-closing invited work is a real cost.

The decision is still no. gbrain's bar for new surface area is deliberately high, and a recipe adds a permanent maintenance obligation to a third-party API (Google People) that no CI exercises — when Google changes `syncToken` expiry or `personFields` requiredness, the repo ships a wrong recipe silently.

Two things from your work that shouldn't be lost, and I'd take either as a small PR:

1. **The four production gotchas you documented** — 410-not-empty on an expired sync token, stable output or you pay a recurring embedding bill, `organizations[0]` being a past employer, and deleted contacts not deleting pages. Those apply to **every** deterministic collector, not just contacts. They'd be ~15 lines in `docs/guides/deterministic-collectors.md` and would improve the calendar and email recipes too.
2. **`GOOGLE_CONTACTS_API_KEY` in `src/commands/features.ts` is a bug you surfaced** — it appears nowhere else in the repo, and `features.ts:149` gates on `secrets.every(s => process.env[s])`, so that recipe reads as permanently unconfigured however a user sets it up. `GOOGLE_CALENDAR_API_KEY` has the identical flaw. Worth a one-line fix on its own.

Thanks for the care you put into this, and sorry the answer is no after you were invited to send it.

rayers:
Fair call on the close, and the reasoning lands. A recipe pinned to a third-party API that no CI exercises is exactly how a repo ends up shipping something wrong silently, and this PR demonstrated that on four separate facts before it was fixed. No complaints.

One thing I need to flag before anyone writes the follow-up, because it is my error and I would rather not have it outlive the PR.

**Gotcha 1 in your list should not be generalized as "410-not-empty."** That was my original claim and it is not supported. The reference for both `people.connections.list` and `otherContacts.list` documents this condition only as an error carrying a `google.rpc.ErrorInfo` with reason `EXPIRED_SYNC_TOKEN`, and pins no HTTP status to it. Google's own contacts guide sample keys on `410`; field reports show `400`. So both of the confident versions are wrong, including the `400` one an earlier reviewer talked me into mid-cycle.

The generalizable lesson is actually better than the one I wrote: **match on the documented error reason, not on a status code you inferred.** A collector hard-coded to either number can silently stop syncing while its heartbeat still reports `ok`, and one hard-coded to bare `400` additionally swallows genuine `INVALID_ARGUMENT` bugs. That is the version worth 15 lines in `docs/guides/deterministic-collectors.md`, and it generalizes past Google.

The other three you listed hold up as written: stable output or you pay a recurring embedding bill, `organizations[0]` is a past employer, and deleted upstream records must not delete brain pages. Also worth adding, since it bit this recipe and is not collector-specific: **sync tokens expire on a clock** (7 days for People), so a cron at exactly the expiry interval turns every "incremental" run into a full sync plus its quota cost.

Happy to send that as a small PR whenever you want it, with the reason-based framing. Not sending it unprompted, since you have had enough of my recipe surface for one week.

On your second item, a small correction so the credit lands on the right bug: `GOOGLE_CONTACTS_API_KEY` was **my** invention in this PR and I removed it in the trim, so master never carried it. The live one is `GOOGLE_CALENDAR_API_KEY` at `features.ts:44`, which appears nowhere else in the repo and makes `calendar-to-brain` read as permanently unconfigured however a user actually sets it up. Sending that one-line fix separately.

Thanks for holding this back from the auto-close pass and for saying why. Reviewing an invited PR on its merits and then declining it with reasons is more work than either auto-merging or auto-closing, and it is the reason the four wrong API facts got found at all.

PR #3506, opened 2026-07-28
v0.42.67.0 fix(build): force LF for shell scripts and route package.json checks through bash

daragao3:
## What this fixes

`bun run test`, `bun run verify`, `bun run ci:local` and `bun run test:e2e` all dispatch through bash, and on Windows that dispatch was broken in two independent places. Neither failure was loud: the commands returned a result, so a run looked finished when almost none of the checks had inspected anything.

Measured on this base (`origin/master` at the time of writing), on Windows:

| | pass | fail | `command not found: scripts/…` |
|---|---|---|---|
| before | 1 | 31 | 29 |
| after | **25** | **7** | **0** |

None of the 7 residual failures are caused by this change. All 7 fail on the pristine baseline too.

**On attribution, precisely:** the measured delta above is the shebang-dispatch fix. All 29 of the baseline's errors were `command not found: scripts/…`; none were `$'\r'`. That is because the bash that happened to be on `PATH` for this measurement is the Cygwin bash shipped with Git for Windows, which *tolerates* CR. The line-ending defect is real but a CR-tolerant shell cannot exhibit it, so it is verified structurally instead (see below) rather than by that number. This cuts both ways and is worth stating plainly: on a machine with a CR-tolerant bash, `.gitattributes` buys nothing today. It matters for contributors whose bash is strict (WSL, and any Linux/macOS consumer of these scripts), and it prevents the failure from reappearing.

## The two defects

**1. Line endings.** Every tracked `*.sh` was checked out with CRLF. The committed blobs were always clean LF; Git for Windows installs with a system-level `core.autocrlf=true` that rewrites them on checkout. A strict bash then died at `scripts/run-unit-parallel.sh` line 23 with `$'\r': command not found`.

A root `.gitattributes` pinning `*.sh text eol=lf` overrides `autocrlf` regardless of the contributor's git config.

Structural verification, since a CR-tolerant shell cannot demonstrate this one behaviourally:

- Before the attribute, 59 of 59 tracked `.sh` files carried CR bytes in the working tree. After it, 0 of 59 do.
- `git add --renormalize .` is a no-op against the change, confirming the index was always correct and only the working tree was wrong. Zero stored content moves.
- `core.autocrlf=true` is still set (system level, `C:/Program Files/Git/etc/gitconfig`); the attribute wins over it, which is the point.

**2. Shebang dispatch.** 33 `package.json` scripts invoked `scripts/foo.sh` directly. Bun cannot exec a `.sh` via its shebang on Windows, so those commands failed the moment they were called. This was invisible until the line endings were fixed. They now go through `bash`, matching the 11 scripts that already did.

Every tracked `.sh` file carries a bash shebang (52 `#!/usr/bin/env bash`, 7 `#!/bin/bash`), so the treatment is uniform. The five `scripts/*.ts` entries still run under bun and are untouched.

## Residual failures (all pre-existing)

- `check:privacy`, `check:test-names`, `check:test-isolation` exceed `run-verify-parallel.sh`'s 120s cap on Windows. They now run for the first time, so this is newly observable rather than newly broken.
- `typecheck` also hits the 120s cap inside the harness. Standalone `bun run typecheck` exits 0.
- `check:wasm` cannot create its `node_modules` symlink on Windows without developer mode.
- `check:skill-brain-first` and `check:resolver` are pre-existing content/state issues. `check:resolver` is not a shell script at all (`bun src/cli.ts check-resolvable`), so this change cannot affect it.

Both classes are filed as follow-ups in `TODOS.md`.

## Platform impact

No behavior change on Linux or macOS. No stored file content moves.

## Verification

```bash
bash -n scripts/run-unit-parallel.sh   # exits 0; previously $'\r': command not found
bun run verify                          # pass=25 fail=7, zero CRLF-class errors
bun run typecheck                       # exits 0
```

## Files

- `.gitattributes` (new) — pins `*.sh text eol=lf`
- `package.json` — 33 check commands routed through `bash`; version bump
- `VERSION`, `CHANGELOG.md`, `TODOS.md` — release bookkeeping for v0.42.67.0


time-attack:
Merging, and merging promptly because of the version claim — details below.

**Version-trio audit, run against both sides:**

```
PR #3506                      origin/master
VERSION:      0.42.67.0       VERSION:      0.42.66.1
package.json: 0.42.67.0       package.json: 0.42.66.1
CHANGELOG:    ## [0.42.67.0]  CHANGELOG:    ## [0.42.66.1]
```

All three agree, `0.42.67.0` beats master, and `git merge-base --is-ancestor origin/master HEAD` confirms master is already an ancestor — so no rebase needed today. I also checked every other open PR: nothing else claims `0.42.67.0` (#3503 → 0.42.66.2, #3481 → 0.42.68.0, #3492 → 0.42.69.0), so the number is free. That's why this goes in now rather than sitting — the moment anything else lands on that number this would need a full re-bump per the merge-conflict recovery procedure.

**`.gitattributes` verified structurally, against the index rather than the working tree.** All 59 tracked `.sh` files contain zero CR bytes in the index, so `*.sh text eol=lf` triggers no renormalization and no stored content moves. All 59 carry a bash shebang — zero non-bash — so the blanket `bash` prefix is uniformly correct. On macOS/Linux the attribute is a no-op.

**The failure mode I actually went looking for was a `bash` prefix silently turning a check into a no-op.** It doesn't: `bun run verify` reports **32 checks, 32 pass** on both this branch and master — identical check count, which is the number that matters. Seven checks invoked individually (`check:newlines`, `check:jsonb`, `check:doc-history`, `check:privacy`, `check:test-isolation`, `check:cli-exec`, `check:source-config-leak`) all exit 0. `bun test test/build-llms.test.ts` → 12 pass; no regeneration needed since `docs/TESTING.md` and `CHANGELOG.md` are both `includeInFull: false` in `scripts/llms-config.ts`.

I can't verify the Windows 1→25 numbers. Credit for being explicit in the body that the `.gitattributes` half is verified structurally rather than behaviorally, because Git-for-Windows' Cygwin bash tolerates CR — that's the right way to report a limit instead of overclaiming, and it's what let me confirm the structural part independently.

Note there's no CI on this branch (fork, no checks reported), which is why I ran the gate on both sides locally.

PR #3511, opened 2026-07-28
v0.42.67.1 fix(engines): load chronicle ontology helpers statically to stop the Windows test-runner crash

daragao3:
## What this fixes

A dynamic `await import('./chronicle/ontology.ts')` issued from `mergeOntologyFact` while the PGLite WASM instance is live can crash the Bun test runner on Windows.

Bun dies before printing its summary, so a single crash discards the pass/fail totals for **every file in that invocation**, not just the file that crashed. A large run can finish reporting no totals at all, which makes a healthy branch and a broken one look identical and leaves nothing to gate a merge on.

The import is hoisted to a static top-level import in both engines, keeping them in lockstep per the engine-parity rule in `CLAUDE.md`.

No cycle risk: `chronicle/ontology.ts` imports only `ingestion/types.ts`, which reaches the engine solely through erased `import type`. Behaviour is unchanged, the same functions run, bound earlier.

## Verification

Controlled A/B on the same 12 files, at `origin/master` `fd8be831`:

```bash
bun test test/chronicle-*.test.ts test/eval-chronicle.test.ts \
  --timeout=60000 --max-concurrency=1
```

| | before | after |
|---|---|---|
| outcome | `panic(main thread): Segmentation fault` | run completed |
| exit code | 3 | 1 |
| totals printed | none, all 12 files discarded | `Ran 64 tests across 12 files`, 63 pass |
| files reaching migration end | 9, died on the 10th | 10 |
| peak RSS | 2.78GB | n/a |

The one remaining failure is environmental, not a regression: `chronicle-advisor.test.ts` hit a 62109ms hook timeout against the 60000ms limit under batch contention, and passes solo in 29.05s (3 pass, exit 0).

`postgres-engine.ts` is not loaded by PGLite tests, so it was covered separately by the 7 non-e2e files that import it, including the import-boundary guard: **77 pass, 0 fail**.

`bun run typecheck` (`tsc --noEmit`): **exit 0**.

## Scope and limits

This removes one trigger, not all of them. Please do not read it as making the full suite pass on Windows.

- Other dynamic imports remain on the engine path (`ai/gateway.ts`, `retry.ts`, `retry-matcher.ts`, `search/recency-decay.ts`, and two audit modules). Most sit on error or cold paths, so they are lower risk than the `mergeOntologyFact` one, which ran on every ontology merge.
- Large runs on Windows can still crash for other reasons.
- The signature observed in these runs is exit 3 with a segfault. Exit 127 has been reported for the same class of crash on this platform but was not observed here.

Both limits are filed as follow-ups in `TODOS.md`.

Worth recording for anyone chasing this later: a single-file or five-file run passes green **even with the trigger present**, so a green small repro is not evidence that a trigger is gone. It took 12 database-backed files in one invocation to crash. Also, `git log -S` will not find a dynamic-to-static import change, because the pickaxe counts occurrences of the string and the count does not change. Use `git log -G`.

## Changes

**Fix**
- `src/core/pglite-engine.ts`, `src/core/postgres-engine.ts`: `valueHash`, `normalizeDimension`, and `isNovelDimension` move from an `await import()` inside `mergeOntologyFact` to a static top-level import, in both engines.

**Release**
- `VERSION`, `package.json`: 0.42.66.1 to 0.42.67.1.
- `CHANGELOG.md`: 0.42.67.1 entry.
- `TODOS.md`: two follow-ups, auditing the remaining dynamic imports on the engine path, and the repro-scale/pickaxe notes above.

Note on the version: this branch originally claimed 0.42.67.0, but #3506 landed that version upstream while this was in flight. The branch merged master and re-bumped to 0.42.67.1, so the first commit here carries a now-superseded v0.42.67.0 subject.


time-attack:
Heads-up on a queue collision: **#3511 and #3536 both self-allocate `0.42.67.1`.** Both are MERGE verdicts on review, but the CI version-gate requires `VERSION` strictly greater than master's, so whichever lands second will fail and need a re-bump.

Merge-order note for the wider queue: the self-allocated ladder is #3511=.67.1, #3536=.67.1 (collision), #3518=.67.3, #3527=.70.0, #3543=.71.0. Landing #3527 (.70.0) first would strand every .67.x PR, so the .67.x set needs to go first, renumbered sequentially.

No action needed from you unless you land second — I'll handle the renumber if it comes to that.

daragao3:
Thank you for the original engine import hardening. The broader replacement in #3596 subsumes the ontology helper hoist, reconciles the remaining engine-live imports, and adds an AST-backed policy guard with focused regression coverage. It also uses the corrected non-causal framing: later testing did not demonstrate dynamic imports as the deterministic cause of the Windows test-runner crash. Superseding this PR with #3596.

Issue #1633, opened 2026-05-30
gbrain sync --source <id> spins indefinitely (busy loop, SIGTERM ignored) → orphaned processes pile up under scheduler


Simona-digital-ops:
## Summary

`gbrain sync --source <id>` can enter an infinite CPU-spinning loop that never resolves, and the process ignores SIGTERM (requires SIGKILL to terminate). When the sync is triggered from a scheduler that has its own session timeout, the parent exits and leaves the sync process orphaned (PPID 1), still burning CPU indefinitely.

## Environment

- gbrain v0.41.11.1 (bun-linked from source)
- macOS 26.3 arm64 (Mac mini M4 16 GB)
- Engine: Postgres (Supabase)
- Source: `briefings` → `/Users/simona/.openclaw/workspace/memory` (local git repo, ~119 pages, markdown strategy)
- Triggered by: OpenClaw cron → isolated agent session → `exec gbrain sync --source briefings --no-pull --no-embed`

## What was observed

On 2026-05-29 ~20:00 CT:

```
ps -axo pid,ppid,etime,%cpu,rss,args | grep gbrain
```

```
13 concurrent processes, all identical:
  bun /Users/simona/.bun/bin/gbrain sync --source briefings --no-pull --no-embed

  PPID: 1 (orphaned — parent had exited)
  Oldest elapsed time: 1-02:56:11 (>24 hours)
  CPU per process: ~60–72%
  RSS per process: ~250 MB
  Combined load average: ~10
  Free memory: 121 MB (of 16 GB); 8.3 GB in compressor (thrashing)
```

After `kill -9` on all 13: load dropped from ~10 → ~5, free memory jumped 121 MB → 9 GB.

### Bug 1 — Individual run hangs in a busy loop (the underlying bug)

A single `gbrain sync --source briefings --no-pull --no-embed` run does not complete. It is not blocked-idle — it pegs ~65% CPU continuously, indicating a busy loop or retry storm rather than I/O wait. The oldest instance had been spinning for >24 hours.

**Suspected location:** The `briefings` source syncs a directory of markdown files (daily notes, memory files). The most likely candidates for the spin, based on the sync code path:

1. `extractLinksForSlugs` / `extractTimelineForSlugs` — iterating over `pagesAffected` with expensive per-file operations
2. `runFactsBackstop` loop — the per-slug loop that calls `queue.add(...)` for each affected page; if queue submission is retrying on a transient error without backoff this would spin
[...]

PR #3517, opened 2026-07-28
v0.42.69.0 fix(build): force LF for Markdown too (frontmatter parsers anchor on LF)

daragao3:
Follow-up to #3506, which pinned `*.sh` to LF. Same root cause, different failure mode, so it gets its own line rather than riding along.

## The problem

`core.autocrlf=true` is the Git for Windows default. It checks `.md` out as CRLF while the stored blob stays LF. The YAML frontmatter fence then carries a trailing carriage return, which an LF-anchored `/^---\n/` (or a `startsWith("---\n")`) does not match. A well-formed document parses as having no frontmatter.

Nothing raises an error. The field just comes back empty, and the caller quietly does the wrong thing. Four visible symptoms trace back to this one cause:

- skill descriptions coming back blank
- a fixer inserting its banner above the frontmatter instead of below it
- resolver trigger extraction dropping entries
- a generated-doc freshness check reporting every line as drifted when nothing drifted

## The change

One line in the existing root `.gitattributes`, plus a comment explaining the failure mode for the next person:

```
*.md text eol=lf
```

## Design notes

**Explicit `text`, not `text=auto`.** Matches the `*.sh` line directly above it. `text=auto` defers to git's binary heuristic; the explicit form is unconditional. A blanket `*` would also pull every tracked file into the eventual renormalize instead of the 349 `.md` files.

**Zero index churn.** All 349 tracked `.md` blobs are already `i/lf`:

```
$ git ls-files --eol -- '*.md' | awk '{print $1}' | sort | uniq -c
    349 i/lf
```

So there is nothing to renormalize and no stored content moves.

**Committing this changes nothing on disk.** It only affects future checkouts. Refreshing an existing working copy is a separate, deliberate step, documented in the CHANGELOG's "To take advantage of" block.

**The parsers stay CR-tolerant.** This attribute fixes this repository's own files. It cannot fix the product behavior, because gbrain also parses Markdown from directories it does not own (`skillsDir` is a runtime parameter). The `\r?`-tolerant readers earn their keep independently.

## Diff

Two commits: the change plus a doc sync.

| Area | File | Change |
|---|---|---|
| Build config | `.gitattributes` | `*.md text eol=lf` plus explanatory comment |
| Release | `VERSION` | `0.42.67.0` to `0.42.69.0` |
| Release | `package.json` | version field to match |
| Docs | `CHANGELOG.md` | `[0.42.69.0]` entry with the self-repair block |
| Docs | `CONTRIBUTING.md` | Windows section covers `*.md`, not just `*.sh`, and gains a verification line for the one-time refresh |
| Docs | `docs/TESTING.md` | line-endings paragraph updated to current state |

Version skips `0.42.68.x`, which is claimed by unrelated in-flight work.

## Verification

- `bun run typecheck` exits 0, clean.
- `bun test test/build-llms.test.ts` passes 12/12.
- `bun run build:llms` regenerates byte-identical bundles. `CONTRIBUTING.md` and `docs/TESTING.md` are linked rather than inlined, so neither bundle moves.

One note for anyone reproducing on Windows: `test/build-llms.test.ts` compares the committed bundles against fresh generator output byte for byte, so it fails on a CRLF working copy and passes once the generator has rewritten the files with LF. The committed blobs are correct either way, and Ubuntu CI never sees it. `llms-full.txt` is a `.txt`, so this PR's `*.md` rule does not cover it. Worth a separate look rather than widening this change.
- `bun install` leaves `bun.lock` unmodified; this lockfile format does not pin the root version.
- `git ls-files --eol` confirms the 349/349 `i/lf` figure quoted above.

The full unit suite does not complete on the Windows box this was prepared on, for a documented pre-existing reason unrelated to this branch. The diff contains no source files, so typecheck plus the bundle freshness test are the gates that apply.

No behavior change on Linux or macOS.


time-attack:

The `.gitattributes` one-liner itself is safe and churn-free (349/349 tracked `.md` are already `i/lf w/lf`; `git check-attr` confirms the rule applies; nothing is rewritten in place; worktree stays clean). But the PR is a symptom patch whose stated design rationale is demonstrably false, it ships zero tests and zero guard, and its user-facing CHANGELOG overclaims. The PR body asserts "The parsers stay CR-tolerant on their own merits" as the reason a checkout-only pin is sufficient scope. I executed that claim against the code: 5 of the 6 LF-anchored frontmatter fence sites do NOT normalize CR. `parseSkillFrontmatter` (src/core/skill-frontmatter.ts:85, 6 callers including `doctor`, `skillify-check`, `dry-fix`) returns `null` on CRLF. Worse, `splitFrontmatter` (src/core/skillopt/apply-edits.ts:99) returns `bodyStart=0` on CRLF, which silently disables the file's own stated D5 safety invariant ("Forbid frontmatter mutation") — I got a confirmed frontmatter-mutating write that is correctly rejected on LF. Both sites read `skillsDir`, a runtime parameter pointing at directories gbrain does not own, so `.gitattributes` can never reach them. The root-cause fix is one line (`.replace(/\r\n/g,'\n')`) already established as the house pattern at check-resolvable.ts:231. Separately, the verification command shipped in both CONTRIBUTING.md and the CHANGELOG self-repair block gives a verified false all-clear on mixed-ending files. Version discipline is clean (title leads with v0.42.69.0; VERSION/package.json/CHANGELOG all agree at 0.42.69.0; master is 0.42.67.0 so strictly greater; bun.lock correctly untouched). No engine/JSONB/source-isolation/trust/knobs-hash/migration surface is touched — 6 files, zero source. Branch is 25 commits behind master.

### Discrimination test

EXECUTED and conclusive. The PR adds zero test files, so I stubbed the behavior back by replacing .gitattributes with `git show origin/master:.gitattributes` (removing all 13 added lines) and re-ran every gate the PR body cites as its verification: `bun run typecheck` -> exit 0 (identical to PR branch); `bun test test/build-llms.test.ts` -> 12 pass / 0 fail / 98 expect() calls (identical); `bash scripts/check-key-files-current-state.sh` -> exit 0, "ok (reference docs history-free; CLAUDE.md within cap)" (identical). Then restored .gitattributes and confirmed `git status --porcelain` empty. Additionally `grep -rn "gitattributes" test/ scripts/ package.json .github/` returns zero matches, so no check anywhere in the repo would fail if the line were deleted tomorrow. The PR's own gates prove nothing about the PR.

### Blockers

**1.** No test, no guard, and the discrimination test passes trivially. Reverting the entire `*.md text eol=lf` addition (.gitattributes:17-29) to master's version leaves every gate the PR cites identical: `bun run typecheck` exit 0, `bun test test/build-llms.test.ts` 12 pass/0 fail, `bash scripts/check-key-files-current-state.sh` exit 0. `grep -rn gitattributes test/ scripts/ package.json .github/` returns ZERO hits — neither this pin nor #3506's `*.sh` pin has any guard, and CI is Ubuntu-only so it can never observe the failure. CLAUDE.md: "CI is the enforcement, not this prose."

**2.** Symptom patch, and the PR's load-bearing claim is false. PR body: "The parsers stay CR-tolerant on their own merits." Verified false at 5 of 6 LF-anchored sites (only check-resolvable.ts:232 normalizes). Executed probe on src/core/skill-frontmatter.ts:85: input `---\nname: my-skill\ndescription: does a thing\ntriggers:...\n---\n` -> `{name:'my-skill', triggers:['alpha','beta']}`; the same document with CRLF -> `parseSkillFrontmatter(...) === null`. Six callers (filing-audit.ts:27, skill-trigger-index.ts:32, dry-fix.ts:31, commands/doctor.ts:10, commands/skillify-check.ts:36) all silently see "no frontmatter". Same LF-only regex unfixed at skill-manifest.ts:50 (parseSkillName, reads `skillsDir` -> the PR's claimed symptom #1), skill-brain-first.ts:167, skillpack.ts:225, enrich/thin.ts:210.

**3.** CONFIRMED safety-invariant bypass that the PR names as a symptom and does not fix. src/core/skillopt/apply-edits.ts:99 `splitFrontmatter` matches `/^---\n[\s\S]*?\n---\n/`; on a CRLF SKILL.md it returns `{body: <whole text>, bodyStart: 0}`, putting the frontmatter inside the mutable slice and disabling D5 (apply-edits.ts:4-7: "Forbid frontmatter mutation... Any edit whose anchor would resolve into the frontmatter is rejected"). Executed: `applyEdit(text, {op:'replace', target:'does a thing', replacement:'PWNED'})` on the LF document -> `rejected / target_not_found` (correct); on the identical CRLF document -> `applied`, output `"---\r\nname: my-skill\r\ndescription: PWNED\r\n---\r\n..."`. version-store.ts atomically writes that result. This IS the PR's claimed symptom #2 ("a fixer inserting its banner above the frontmatter instead of below it"), and .gitattributes cannot reach it because skillopt operates on the install path / skillsDir.

**4.** Shipped verification step gives a verified false all-clear. CONTRIBUTING.md:34 and the CHANGELOG "To take advantage" block (CHANGELOG.md, step 2) both prescribe `git ls-files --eol -- '*.md' | grep -c 'w/crlf'` and state "`0` means it worked." Executed: a `.md` with mixed endings (`---\r\nname: x\n---\nbody\r\n`) reports `i/none w/mixed attr/text eol=lf`, NOT `w/crlf`. The command prints `0`, the contributor is told the refresh succeeded, and the silent-frontmatter bug it exists to detect persists. Needs `grep -cE 'w/(crlf|mixed)'`. (Also: `grep -c` exits 1 on a zero count, so the line is unusable in any `&&` chain.)

**5.** CHANGELOG.md overclaims in the user-facing headline: "the settings block at the top of a Markdown file is now read correctly. Until this release it was often ignored." False on two counts. (a) With `text eol=lf` active, git normalizes a CRLF worktree file for comparison, so an existing Windows checkout keeps CRLF on disk AND reports `git status` clean — nothing is read correctly until the contributor runs the manual destructive refresh, which nothing prompts or detects. (b) Two of the four enumerated symptoms stay live for any Markdown outside this repo (blockers 2 and 3). The fourth enumerated symptom (generated-doc freshness) is also not covered: `git check-attr text eol -- llms-full.txt llms.txt` returns `text: unspecified / eol: unspecified` for both.

### Not verified

- Full unit suite (bun run test / verify / ci:local) — deliberately skipped per the execution-footprint constraint. Justified here: the diff contains zero source files, so typecheck plus the bundle-freshness test are the only gates that can be affected.
- marked@18 CRLF+NBSP super-linear lexer bug (#1633) — NOT reproduced. Installed version is marked@18.0.6 (package.json range ^18.0.2). I probed CRLF+NBSP shapes at 39, 69, 99, 129, 189, 212, 812 and 3212 bytes through marked.lexer(); every one completed in <=1ms with no super-linear growth. I do not have the original 50-byte payload, so this is a no-repro on my shapes, not a clearance. Note the direction of interaction is favorable either way: pinning repo .md to LF reduces local exposure, and the product entry point (src/core/import-file.ts:131) gates lexer entry behind a pre-check regex that already accounts for \r.
- Real Windows checkout behavior — no Windows box available. Conclusions about Windows are derived from git attribute semantics (verified locally via `git check-attr` and `git ls-files --eol` against real CRLF and mixed-ending files I created) rather than from an actual Git for Windows clone.
- Branch is 25 commits behind origin/master and was reviewed as-is. The VERSION / package.json / CHANGELOG.md merge conflict on rebase is near-certain; the post-merge 3-line trio audit has not been performed and must be before landing.

- `gh pr view 3517`
- `gh pr diff 3517 > /tmp/pr3517.diff`
- `gh pr checkout 3517`
- `git fetch origin master && git show origin/master:VERSION && git show origin/master:package.json`
- `git merge-base --is-ancestor origin/master HEAD ; git rev-list --count HEAD..origin/master`
- `git check-attr text eol -- README.md CLAUDE.md docs/TESTING.md llms-full.txt llms.txt`
- `git ls-files --eol -- '*.md' | awk '{print $1, $2}' | sort | uniq -c`
- `git ls-files --eol | grep -E 'i/(crlf|mixed)'`
- `grep -rn 'gitattributes' test/ scripts/ package.json .github/`
- `grep -rn "^---\\n|'---\\n'" src/ scripts/  (enumerate all LF-only frontmatter fence sites)`
- `grep -rn 'skill-frontmatter' src/  (caller census: 6 callers)`
- `bun run typecheck > /tmp/tc_pr.txt   [exit 0, 0 TS errors]`
- `bun test test/build-llms.test.ts > /tmp/llms_pr.txt   [12 pass / 0 fail]`
- `git show origin/master:.gitattributes > .gitattributes   (discrimination stub)`
- `bun test test/build-llms.test.ts > /tmp/llms_stub.txt   [12 pass / 0 fail — IDENTICAL]`
- `bun run typecheck > /tmp/tc_stub.txt   [exit 0 — IDENTICAL]`
- `bash scripts/check-key-files-current-state.sh   [exit 0 — IDENTICAL]`
- `bun run /tmp/crlf_probe.ts   (parseSkillFrontmatter + extractTriggers, LF vs CRLF)`
- `bun run /tmp/d5_probe3.ts   (splitFrontmatter bodyStart + applyEdit D5 bypass, LF vs CRLF)`
- `bun run /tmp/marked_probe.ts   (marked@18.0.6 lexer timing, CRLF+NBSP shapes)`
- `printf CRLF and mixed .md fixtures + git add -N + git ls-files --eol -- 'zz_*'   (w/crlf vs w/mixed reporting)`
- `grep -n '0.42.6|"version"' bun.lock`
- `git show origin/master:CHANGELOG.md | grep 'reset --hard'   (destructive block is pre-existing, not new here)`
- `cp /tmp/ga.bak .gitattributes ; rm probe files ; rm node_modules symlink ; git status --porcelain   [clean]`

</details>

---

time-attack:

PR #3518, opened 2026-07-28
v0.42.67.3 fix(test-runner): derive the per-shard cap and add a stall watchdog

daragao3:
## Why

`scripts/run-unit-parallel.sh` gives each shard a hardcoded wallclock cap. A constant is really a guess about two moving quantities, files in the suite and per-file cost on the host, so it goes stale on its own. It has gone stale twice: `900s` was sized for the 8-shard/~80-file split and false-killed the 4-shard/159-file split, and its `1500s` replacement was calibrated on Apple Silicon and false-kills healthy shards on Windows.

`1500 / 253 files = 5.93s`, so the constant was always a per-file budget wearing a wallclock costume. This makes that explicit.

## What changed

**Derived cap.** `SHARD_TIMEOUT` is now `files-per-shard x per-file budget`, floor `600s`. Per-file budget is `6s` on Unix and `30s` on Windows (`uname -s` matching `MINGW*`/`MSYS*`/`CYGWIN*`, falling back to `$OS`). Behaviour-preserving on Unix: `253 x 6 = 1518s` against the `1500s` it replaces. Adding test files widens the cap with no constant to remember.

**Stall watchdog.** A cap generous enough to be safe is a poor hang detector, so a second guard keys on forward progress instead of elapsed time. It polls each shard's log size and kills a shard whose log has not grown for `GBRAIN_TEST_SHARD_STALL_SECONDS` (default `600`), reporting `STALLED` separately from `WEDGED`. Same instrument gbrain already uses for sync's stall abort. The two paths print different remedies because the fix differs, and the banner now states how the cap was derived.

All three knobs (`GBRAIN_TEST_SECONDS_PER_FILE`, `GBRAIN_TEST_SHARD_TIMEOUT`, `GBRAIN_TEST_SHARD_STALL_SECONDS`) reject non-numeric input with exit 2 rather than deriving a `0s` cap that would kill every shard instantly.

## Interaction with #2864 / #3340

This branch was authored before #3340 landed and conflicted with it in the shard-spawn block. Resolved keeping **both** behaviours, and the result generalises that fix rather than reverting any of it:

- The shard now runs in the background under **both** the `timeout(1)` and no-timeout paths, so the watchdog can attach to it. #3340 only needed the fallback path backgrounded.
- `rc` is still captured immediately after the shard's own `wait`, before any teardown can overwrite `$?` with a killed watchdog's `143`.
- The children-first reaping #3340 introduced (`pkill -P` then `kill` then `wait`) is now a loop over both helpers, guarded by `[ -n "$helper" ] || continue` since `cap_pid` is empty on the `timeout(1)` path.

The regression suite #3340 added is retained unchanged, now guarded by a symlink-capability probe (see below).

## Tests

`test/scripts/run-unit-parallel.test.ts`: **17 pass / 4 skip / 0 fail**, exit 0. `bun run typecheck` exit 0. `scripts/check-progress-to-stdout.sh` OK.

New coverage for the derived cap (floor, scaling with file count, explicit override, non-numeric rejection, banner naming its own derivation) and for the watchdog (banner, disable, rejection, plus a real module-top-level hang that must be killed on the progress signal and not the cap).

Also fixes three pre-existing Windows failures in the runner's own test:

- an MSYS `/tmp` vs Windows-path assertion
- an EBUSY tmpdir teardown race
- the #3340 fallback suite's `beforeAll` builds a curated PATH from `symlinkSync`, which throws `EPERM` on Windows without `SeCreateSymbolicLinkPrivilege`. That threw hook did not just report a spurious failure, it aborted mid-hook and perturbed an unrelated suite in the same file. It is now behind a `describe.skipIf` symlink probe, so it still runs everywhere it can. Verified against pristine `master` on the same host, which reproduces the identical `EPERM` failure.

And a heartbeat rendering bug where a no-match `grep -c` printed two lines and broke the arithmetic building the progress line.

## Known limitation

`WINDOWS_SECONDS_PER_FILE=30` was chosen for headroom, **not measured end to end**. Nobody has run the unit suite to completion on Windows to find the real per-file cost, so the Windows cap is a deliberately generous guess and the stall watchdog, not the cap, is what catches a real hang there. Calibrating it needs a suite that completes on Windows, which is separate blocked work. Filed in `TODOS.md` rather than guessed at again here. The Unix side is anchored at `1518s` against the `1500s` constant it replaces.

## Files

| Area | File |
|---|---|
| runner | `scripts/run-unit-parallel.sh` |
| tests | `test/scripts/run-unit-parallel.test.ts` |
| docs | `docs/TESTING.md` |
| release | `VERSION`, `package.json`, `CHANGELOG.md`, `TODOS.md` |


time-attack:

PR #3518 replaces the hardcoded per-shard wallclock cap in scripts/run-unit-parallel.sh with files-per-shard x per-file budget (6s Unix / 30s Windows, floor 600s) and adds a log-growth stall watchdog that SIGKILLs a shard silent for 600s. The engineering is careful, the discrimination test is strong (11 of 19 tests in the file genuinely fail against master's script; the stall case hangs 748s and gets STALLED=""), typecheck/verify/build-llms are green, the version trio is consistent, and I additionally executed the WEDGED path on BOTH branches (rc=124 under timeout(1), rc=143 + sentinel on the fallback) plus orphan reaping, all of which the PR body admits it did not test. Two real problems remain: the new derivation cuts the cap 47% (1500s -> 792s) on the explicitly-supported SHARDS=8 path, which the script's own recorded measurements say needs ~1030-1270s, so that configuration false-kills; and the new default-on stall watchdog's justification ("10x bun's 60s per-test cap") is wrong, with a measured healthy-shard quiet gap of 161s on a loaded box (3.7x margin) against a mechanism I proved by execution will SIGKILL a healthy passing test. Both are one-line tuning fixes in a PR whose entire purpose is to stop false-killing healthy shards.

### Discrimination test

EXECUTED AND PASSED. Stubbed master's scripts/run-unit-parallel.sh in over the PR's version (`git show origin/master:scripts/run-unit-parallel.sh > scripts/run-unit-parallel.sh`) and re-ran the PR's own test file: 8 pass / 11 fail, exit 1, 763s total. Every new case fails on master — all 7 derived-cap cases, all 4 watchdog cases. The load-bearing one ('kills a shard that stops producing output') ran 748,834ms before bun's 240s test timeout could report, and asserted `Expected to contain: "STALLED" / Received: ""` — on master the hung shard rides the 9000s pinned cap with no watchdog and the wrapper never emits STALLED at all. Restored the PR version (`git status` clean) and re-ran: 19 pass / 0 fail, exit 0, 38.7s. This is a genuine discriminating test, not a both-ways pass.

### Blockers

**1.** scripts/run-unit-parallel.sh:158-160 (derived cap) — SHARDS=8 regression, cap drops 1500s -> 792s. EXECUTED: `SHARDS=8 bash scripts/run-unit-parallel.sh --dry-run` prints `timeout=792s (132 files/shard × 6s)`; master prints `timeout=1500s`. The 6 s/file budget is calibrated at N=4, but the script's own retained comment records 8-shard runs at 620-770s for ~80 files/shard = 7.8-9.6 s/file, because 8 concurrent PGLite WASM inits contend (that is why the constant was widened to 1500 in the first place). At today's 132 files/shard that predicts 1030-1270s of real work against a 792s cap => all eight shards WEDGED. Failing input: `SHARDS=8 bun run test` on any host. Fix: scale the per-file budget with N, or floor the derived value at the 1500s it replaces.

**2.** scripts/run-unit-parallel.sh:240-263 + :296 (stall watchdog, default 600s ON) — a new default-on SIGKILL path whose stated safety argument does not hold. The comment claims 600s is "10× bun's 60s per-test cap, so it clears any legitimate quiet period", but log-quiet intervals are NOT bounded by the per-test cap: the unit set declares per-test timeouts up to 180s (test/cli-pipe-truncation.test.ts:46) and 240s (this PR's own new test, test/scripts/run-unit-parallel.test.ts:456), and a spawnSync/execFileSync child's output never reaches the shard log at all. EXECUTED proof of the mechanism: a fixture with one healthy, PASSING test that sleeps 45s (under bun's 60s cap) run with GBRAIN_TEST_SHARD_STALL_SECONDS=20 produced `shard 1/4: STALLED after 20s with no output (rc=137)` and WRAPPER_EXIT=1 — the watchdog SIGKILLs healthy work, it does not merely detect hangs. MEASURED margin: a real unit shard (SHARD=1/4, 51 files completed) on a loaded box held a 161s maximum quiet gap => 3.7x headroom, not 10x. On Windows, where this PR itself states per-file cost is several times higher, 600s is the least-tested surface in the change. Failing input: any host under 3-4x load running a >200s-quiet unit test, or any Windows host. Fix: raise the default (e.g. 1800s) or ship the watchdog opt-in until a real quiet-gap distribution is measured.

**3.** scripts/run-unit-parallel.sh:531-537 (WEDGED remedy text) — low severity, but the block unconditionally prints "the cap is derived from the file count; ... raise the per-file budget: GBRAIN_TEST_SECONDS_PER_FILE=N" even when TIMEOUT_ORIGIN is `env GBRAIN_TEST_SHARD_TIMEOUT`, where that knob has zero effect. EXECUTED: with GBRAIN_TEST_SHARD_TIMEOUT=20 the failure log reads `--- shard 1: WEDGED after 20s (env GBRAIN_TEST_SHARD_TIMEOUT) ---` immediately followed by the derived-cap advice. Misleading remedy in exactly the diagnostic block the PR added to make WEDGED actionable.

### Not verified

- Full 4-shard `bun run test` to completion on this host. Two attempts were lost: the first background wrapper died at ~62s with exit 144 (harness artifact), and my own `SHARDS=8 --dry-run` probe wiped the in-flight run's shard logs (the `rm -f $LOG_DIR/shard-*` at scripts/run-unit-parallel.sh:188 runs BEFORE the --dry-run exit at :212 — pre-existing on master, but it now also blinds the new watchdog into a false STALLED after 600s if anyone runs a second invocation in the same worktree). Two other agents were running the unit suite concurrently on this box throughout, so a clean full-suite wallclock number was not obtainable. Substituted a single-shard probe (51/264 files) for the quiet-gap measurement.
- Windows behavior end to end. WINDOWS_SECONDS_PER_FILE=30 and the 600s stall default on MSYS2/Git Bash/Cygwin were not executed — no Windows host available. The PR itself files this as a P2 TODO and admits the 30s figure is unmeasured; I could not narrow it either.
- GNU-grep behavior of the loosened heartbeat regex. shard_pglite_init_count changed the literal `→` to `.` (scripts/run-unit-parallel.sh:375). Verified matching under both LC_ALL=C and en_US.UTF-8 with macOS BSD grep; GNU grep under LC_ALL=C treats `.` as one byte and would not match the 3-byte `→`. Heartbeat cosmetics only, not correctness, but unverified on Linux.
- Actual CI LPT rebalance cost of the file's larger runtime. scripts/test-weights.json still lists test/scripts/run-unit-parallel.test.ts at 427ms against a measured ~38.7s; all 10 CI test shards are green on this PR so the imbalance is currently absorbed, but I did not measure the per-shard delta.
- `stat` dependency added by shard_quiet_seconds is absent from the fallback suite's curated PATH tool list (test/scripts/run-unit-parallel.test.ts:267), so the heartbeat quiet note silently reads 0 there. Degrades gracefully; I did not construct a case where it matters.
- The claim in the CHANGELOG/PR body of '253 files/shard' and '1518s'. Actual on this checkout is 264 files/shard and 1584s (verified via --dry-run). Cosmetic drift in the release copy, not re-derived against their host.

- `gh pr view 3518 — exit 0`
- `gh pr diff 3518 > /tmp/pr3518.diff — exit 0 (827 lines, every hunk read)`
- `gh pr checkout 3518 — exit 0 (branch claude/busy-keller-74b5d3)`
- `bun install — exit 0`
- `bash scripts/run-unit-parallel.sh --dry-run — exit 0, banner `timeout=1584s (264 files/shard × 6s) | stall=600s``
- `SHARDS=8 bash scripts/run-unit-parallel.sh --dry-run — exit 0, banner `timeout=792s (132 files/shard × 6s)` (master: 1500s) — BLOCKER 1`
- `bun test test/scripts/run-unit-parallel.test.ts — exit 0, 19 pass / 0 fail, 38.7s (PR body claims 17 pass / 4 skip = 21 tests; minor count mismatch)`
- `git show origin/master:scripts/run-unit-parallel.sh > scripts/run-unit-parallel.sh; bun test test/scripts/run-unit-parallel.test.ts — exit 1, 8 pass / 11 fail, 763s — DISCRIMINATION TEST PASSED`
- `cp /tmp/pr_version.sh scripts/run-unit-parallel.sh (restored; git status clean) — exit 0`
- `bun run verify — exit 0, 32/32 checks green, 58s`
- `bun test test/build-llms.test.ts — exit 0, 12 pass`
- `false-kill demo: healthy 45s-quiet PASSING test + GBRAIN_TEST_SHARD_STALL_SECONDS=20 — WRAPPER_EXIT=1, `shard 1/4: STALLED after 20s with no output (rc=137)` — BLOCKER 2 mechanism proven`
- `WEDGED both branches: GBRAIN_TEST_SHARD_TIMEOUT=20, stall off — timeout(1) path rc=124 WEDGED; no-timeout fallback rc=143 WEDGED + sentinel. Both exit 1`
- `orphan check, fallback path, GBRAIN_TEST_SHARD_TIMEOUT=7777 — WRAPPER_EXIT=0, `shard 1/2: pass=1 fail=0 skip=0 rc=0`, zero `sleep 7777` orphans after teardown`
- `single real shard quiet-gap probe (SHARD=1/4, 51/264 files) — max log-quiet gap 161s on a loaded box (3.7x margin vs the 600s default, not the claimed 10x)`
- `version trio audit (VERSION / package.json / CHANGELOG) + `git show origin/master:VERSION` — all 0.42.67.3 vs master 0.42.67.0`
- `gh pr view 3518 --json mergeable,mergeStateStatus,statusCheckRollup — MERGEABLE / CLEAN, 24 checks SUCCESS (Heavy tests SKIPPED)`
- `grep -rnoE '\}, *[0-9_]{6,}\)' test (excluding slow/serial/e2e) — max declared per-test timeout in the unit set is 240s; 180s at test/cli-pipe-truncation.test.ts:46`

</details>

---

time-attack:

The core value is real and NOT on master: origin/master scripts/run-unit-parallel.sh:82 still hardcodes SHARD_TIMEOUT=1500 (the constant that false-kills healthy Windows shards), has no stall watchdog, and still carries the heartbeat double-output bug (`grep -cE ... || echo 0` at line 222) this PR fixes. The wave's merged #3631 (e3806cf4, TERM->KILL escalation) is complementary, not a supersession, so it is not ALREADY_FIXED — but a trial merge shows the PR cannot land as-is: it must be rebased to preserve #3631's behavior, which its spawn-block rewrite currently reverts.

**Conflict state vs current master:** Trial merge conflicts in 5 files. VERSION/package.json/CHANGELOG are trivial staleness (PR claims 0.42.67.3; master is 0.42.69.0 — CI version-gate fails without a re-bump). scripts/run-unit-parallel.sh and test/scripts/run-unit-parallel.test.ts are SEMANTIC conflicts with merged #3631: the PR's timeout(1) invocation drops `--signal=TERM --kill-after="${SHARD_KILL_AFTER}s"`, its fallback reverts `sleep "$SHARD_KILL_AFTER"` to `sleep 5`, and it renames pid->run_pid — and #3631's source-contract tests on master pin those exact strings (describe 'timeout escalation contract'), so any resolution taking the PR's side turns master red. The fix still applies post-wave and the reconciliation is well-defined: background the timeout(1) path WITH --kill-after retained, keep the PR's WEDGED-sentinel-before-kill + derived cap + watchdog, keep master's rc-137->WEDGED line (auto-merges cleanly; STALLED is checked before WEDGED so diagnosis stays correct), and update #3631's contract-test strings for the pid->run_pid rename.

**Risk if merged as-is:** If merged with a naive take-PR-side resolution: (1) reverts #3631's TERM->KILL escalation so TERM-resistant bun shards can again outlive the cap; (2) immediately fails #3631's two source-contract tests in test/scripts/run-unit-parallel.test.ts (string pins on --kill-after and sleep "$SHARD_KILL_AFTER"); (3) stale VERSION 0.42.67.3 < master 0.42.69.0 fails the CI version-gate. After a correct rebase, residual risk is low — dev-tooling only, no src/ changes, and the new tests are hermetic --dry-run banner probes plus one real-hang watchdog test.

Issue #1955, opened 2026-06-08
PGLite WASM crash on macOS 26.3 — resolved by clearing old data


itstimwhite:
## Resolved

After clearing  and running Setting up local brain with PGLite (no server needed)...
  Embedding: ollama:nomic-embed-text (768d)
  Expansion: ollama:llama3.2
  Chat: deepseek:deepseek-chat

Brain ready at /Users/timwhite/.gbrain/brain.pglite
275 pages. Engine: PGLite (local Postgres).

Existing brain detected. To wire up the v0.10.3 knowledge graph:
  gbrain extract links --source db        (typed link backfill)
  gbrain extract timeline --source db     (structured timeline backfill)
  gbrain stats                            (verify links > 0)

When you outgrow local: gbrain migrate --to supabase

--- GBrain Mod Status ---
Skills: 49 loaded
GStack: found (claude)
Resolver: skills/RESOLVER.md
Soul audit: run `gbrain soul-audit` to customize agent identity, GBrain works on macOS 26.3 (Darwin 25.3.0, arm64).

## Root Cause

The WASM crash was caused by corrupted PGLite data from a previous macOS version, not by the macOS 26.3 WASM runtime itself.

## Fix

Setting up local brain with PGLite (no server needed)...
  Embedding: voyage:voyage-4-large (1024d)

═══════════════════════════════════════════════════════════════
[gbrain] search mode tentatively set to: conservative
[gbrain] Reason: No OpenAI key configured — semantic cache still works, but no LLM expansion possible.
═══════════════════════════════════════════════════════════════

[AGENT] Before continuing, SHOW this matrix to your operator and
[AGENT] ask which mode they want. The default above preserves the
[AGENT] v0.31.x retrieval shape. Run `gbrain config set search.mode
[AGENT] <conservative|balanced|tokenmax>` if the operator picks
[AGENT] differently. See INSTALL_FOR_AGENTS.md Step 3.5 for the
[AGENT] full protocol.

Per-query cost @ 10K queries/mo (search payload only, no cache savings):

                    Haiku 4.5     Sonnet 4.6    Opus 4.7
                    ($1/M input)  ($3/M input)  ($5/M input)
   conservative     $40/mo        $120/mo       $200/mo
   balanced         $100/mo       $300/mo       $500/mo
   tokenmax         $200/mo       $600/mo       $1,000/mo

   (scales linearly: ×10 for 100K queries/mo, ÷10 for 1K)
[...]

Issue #1954, opened 2026-06-08
PGLite WASM crash on macOS 26.3 — fresh init works briefly then crashes


itstimwhite:
## Summary

Fresh Setting up local brain with PGLite (no server needed)...
  Embedding: ollama:nomic-embed-text (768d)
  Expansion: ollama:llama3.2
  Chat: deepseek:deepseek-chat succeeds but subsequent commands crash with WASM runtime error.

## Environment
- macOS 26.3 (Darwin 25.3.0, arm64)
- gbrain 0.42.8.0
- Bun 1.3.11, 1.3.14 (tried multiple versions)
- PGLite pinned at 0.4.3

## Steps to Reproduce
1. 
2. Setting up local brain with PGLite (no server needed)...
  Embedding: ollama:nomic-embed-text (768d)
  Expansion: ollama:llama3.2
  Chat: deepseek:deepseek-chat

═══════════════════════════════════════════════════════════════
[gbrain] search mode tentatively set to: conservative
[gbrain] Reason: No OpenAI key configured — semantic cache still works, but no LLM expansion possible.
═══════════════════════════════════════════════════════════════

[AGENT] Before continuing, SHOW this matrix to your operator and
[AGENT] ask which mode they want. The default above preserves the
[AGENT] v0.31.x retrieval shape. Run `gbrain config set search.mode
[AGENT] <conservative|balanced|tokenmax>` if the operator picks
[AGENT] differently. See INSTALL_FOR_AGENTS.md Step 3.5 for the
[AGENT] full protocol.

Per-query cost @ 10K queries/mo (search payload only, no cache savings):

                    Haiku 4.5     Sonnet 4.6    Opus 4.7
                    ($1/M input)  ($3/M input)  ($5/M input)
   conservative     $40/mo        $120/mo       $200/mo
   balanced         $100/mo       $300/mo       $500/mo
   tokenmax         $200/mo       $600/mo       $1,000/mo

   (scales linearly: ×10 for 100K queries/mo, ÷10 for 1K)
   25x corner-to-corner spread. Natural diagonal pairings span ~4x.

To change later: gbrain config set search.mode <mode>
To see what is running: gbrain search modes

Brain ready at /Users/timwhite/.gbrain/brain.pglite
0 pages. Engine: PGLite (local Postgres).
Next: gbrain import <dir>

When you outgrow local: gbrain migrate --to supabase

--- GBrain Mod Status ---
Skills: 49 loaded
GStack: found (claude)
[...]

PR #3519, opened 2026-07-28
fix(pglite): stop blaming macOS for a bare Aborted(); name WAL damage (#2674)

### LINKED ISSUE #223: PGLite WASM crash on macOS 26.3 with Bun 1.3.11

### LINKED ISSUE #2674: Corrupted-store PGLite init abort is mislabeled as the macOS 26.3 WASM bug — banner masks the real cause
## Summary
When a PGLite brain store is damaged (heap corruption from a crash-mid-write), every new process aborts during PGLite init with `Aborted(). Build with -sASSERTIONS for more info.` — and gbrain prints the canned macOS 26.3 WASM bug hint (#223) for it. The hint sent us chasing a phantom OS bug for a corrupted-store problem.

## Evidence it was not the OS/WASM bug (macOS 27.0 beta, Bun 1.4.x, gbrain 0.42.44.0, PGLite 0.4.3)
- Fresh standalone PGLite 0.4.3 **with vector + pg_trgm** created and queried fine on the same machine while gbrain commands aborted.
- `gbrain doctor` (filesystem checks) reported 95 unresolved sync failures with `dead heap-only tuple (0, 136) is not linked to from any HOT` — classic Postgres heap corruption, likely from a crash during concurrent access to the single-writer store.
- Quarantining the store and running `gbrain init --pglite` fresh: all commands work immediately; full re-sync + re-embed restored parity.

## Ask
1. Distinguish init-abort causes where possible (corrupted-store read vs runtime WASM incompatibility), or soften the banner to "most common causes include ... OR a damaged store — try `gbrain init` into a scratch GBRAIN_HOME to isolate".
2. Consider a `gbrain doctor` probe that attempts a scratch-store init to separate "environment can't run PGLite" from "this store is damaged".

Happy to provide more detail from our forensics.

time-attack:
Fixes #2674. Also fixes the actionable half of #223, #1954 and #1955 (all closed as duplicates of the same misdiagnosis).

## Diagnosed against a real broken brain

I hit this live on my own install, so this is measured rather than reasoned from the issue text. On **macOS 26.4 / Bun 1.3.14 / PGLite 0.4.3**:

| Test | Result | What it rules out |
|---|---|---|
Fresh scratch store: create → write → close → reopen | **works** | the runtime is not broken on this OS |
A **copy** of the damaged store | **fails identically** | lock contention, concurrent access |
`rm postmaster.pid` on the copy | still fails | the thing everyone tries first |
PGLite with `debug: 1` | **reveals the real cause** | the OS theory |

```
pglite: found DB, resuming
LOG:   database system was interrupted; last known up at 2026-07-11 00:12:20
LOG:   invalid resource manager ID in checkpoint record
PANIC: could not locate a valid checkpoint record at 0/37ADDF8
Aborted()
```

So it's **WAL/checkpoint damage** — Postgres started fine and PANICked during recovery. Nothing to do with macOS. And the `last known up at` line means the brain had been silently unreadable for **17 days** while the error blamed the OS.

## The detail that shapes the fix

**PGLite reports only `Aborted()` to JS and prints the PANIC to its own stderr.** `classifyPgliteInitError` never sees the checkpoint text, so no regex on it can help on the default path — and from `Aborted()` alone, a damaged store and a broken runtime are genuinely indistinguishable. Guessing "macOS" there is exactly what caused the misdiagnosis. #2891 platform-gated the banner for non-darwin but left it on macOS, which is where every reporter is.

## Three changes

**1. `abort-ambiguous` for a bare `Aborted()`.** Rather than guessing, the hint teaches the 5-second discriminator this issue asked for:

```
GBRAIN_HOME=$(mktemp -d) gbrain init --pglite --no-embedding
```

Succeeds → your store is damaged. Fails too → the runtime. Identical text on darwin and elsewhere, because the string carries no platform signal — the old code branched on `process.platform` here, which was guessing dressed as detection.

**2. `wal-corrupt` for the recovery-PANIC signatures**, so any caller that *does* surface stderr gets a precise verdict. Its hint rules out the OS bug and locks explicitly, says plainly that **markdown is unaffected** because the DB holds derived data (chunks, embeddings, links, facts) that a re-sync rebuilds, points at `reinit-pglite`, and names the two things that cannot work (`postmaster.pid`, re-running migrations — both verified no-ops).

**3. The header is verdict-dependent.** `"PGLite failed to initialize its WASM runtime."` was asserted for *every* verdict, including a damaged store where the runtime started fine. That header was itself part of the misdiagnosis. Store-fault verdicts now say `"PGLite could not open your brain."`

## Verification

`bun run verify` 32/32 · `bun run typecheck` clean · 65 pass across the classifier + structural + engine-factory suites.

**The tests discriminate: 50 pass here, 6 fail on master.** Failures on master are behavioral (wrong verdict, wrong hint, platform-branching), not missing-symbol.

I updated the pre-existing `all verdicts produce the canonical header line` test rather than deleting it — its real intent (every verdict has a header and carries the original error) is preserved as two assertions, while the uniformity claim it encoded is the bug.

## Scope

**This does not close #223.** That umbrella also covers a genuinely distinct compiled-binary `$bunfs` mode already handled by the `bunfs` branch. This fixes the *diagnosis*, not a crash — worth being precise about, since #223's own thread shows reporters on macOS 15.5, which is itself evidence the OS version was never the variable.

`gbrain doctor` still has no scratch-store init probe (this issue's ask #2). That's a separate, larger change; the hint above gives the user the same discriminator by hand in the meantime.

### DIFF
--- a/src/core/pglite-engine.ts
+++ b/src/core/pglite-engine.ts
@@ -164,7 +164,17 @@ export function computeSnapshotSchemaHash(
  * errors). Match the literal `$$bunfs` marker OR ENOENT+pglite.data
  * co-occurrence.
  */
-export type PgliteInitFailure = 'bunfs' | 'macos-26-3' | 'corrupt' | 'unknown';
+export type PgliteInitFailure =
+  | 'bunfs'
+  | 'macos-26-3'
+  | 'corrupt'
+  // #2674: Postgres PANICked during WAL recovery — a damaged store, not a
+  // broken runtime. Recoverable diagnosis; distinct from `corrupt` (catalog).
+  | 'wal-corrupt'
+  // #2674: a bare `Aborted()` with no further signal. Deliberately NOT guessed
+  // at — the hint teaches the 5-second isolation test instead.
+  | 'abort-ambiguous'
+  | 'unknown';
 
 // #2674: non-Error rejections (Emscripten aborts can throw plain objects)
 // used to stringify as "[object Object]" — prefer .message when present.
@@ -182,9 +192,33 @@ export function classifyPgliteInitError(message: string): PgliteInitFailure {
   if (/58P01|internal_load_library|type "?vector"? does not exist|relation "?content_chunks"? does not exist/i.test(message)) {
     return 'corrupt';
   }
-  if (/abort.*runtime|macos.*26\.3|wasm.*runtime/i.test(message)) {
+  // #2674: WAL / checkpoint damage is a DIFFERENT corruption shape from #2348's
+  // catalog damage, and it is what the reporters in #223 actually hit. Postgres
+  // PANICs during recovery rather than failing a catalog lookup:
+  //
+  //   LOG:  database system was interrupted; last known up at <ts>
+  //   LOG:  invalid resource manager ID in checkpoint record
+  //   PANIC: could not locate a valid checkpoint record at 0/37ADDF8
+  //
+  // Matched here so that any caller which DOES surface PGLite's stderr gets a
+  // precise verdict. Note the default path does not: PGLite reports only
+  // `Aborted()` to JS and prints the PANIC to its own stderr, which is exactly
+  // why `abort-ambiguous` below exists.
+  if (
+    /could not locate a valid checkpoint record|invalid resource manager ID|database system was interrupted|could not access status of transaction|dead heap-only tuple|invalid checkpoint record/i.test(
+      message,
+    )
+  ) {
+    return 'wal-corrupt';
+  }
+  if (/macos.*26\.3|wasm.*runtime/i.test(message)) {
     return 'macos-26-3';
   }
+  // A bare Emscripten abort with nothing else to go on. Do NOT guess a cause:
+  // measured on macOS 26.4 / Bun 1.3.14 / PGLite 0.4.3, a store with WAL damage
+  // and a healthy runtime produce the SAME string here. Blaming the OS sent
+  // #223, #1954, #1955 and #2674 chasing a phantom platform bug for weeks.
+  if (/^(RuntimeError:\s*)?Aborted\(\)/i.test(message.trim())) return 'abort-ambiguous';
   return 'unknown';
 }
 
@@ -195,7 +229,14 @@ export function buildPgliteInitErrorMessage(
   // monkey-patching process.platform.
   platform: NodeJS.Platform = process.platform,
 ): string {
-  const header = 'PGLite failed to initialize its WASM runtime.';
+  // #2674: only assert "WASM runtime" when we actually believe the runtime is
+  // at fault. For a damaged store the runtime started fine and then PANICked
+  // during recovery, so the old unconditional header was itself part of the
+  // misdiagnosis.
+  const header =
+    verdict === 'wal-corrupt' || verdict === 'corrupt' || verdict === 'abort-ambiguous'
+      ? 'PGLite could not open your brain.'
+      : 'PGLite failed to initialize its WASM runtime.';
   let hint: string;
   switch (verdict) {
     case 'bunfs':
@@ -222,6 +263,37 @@ export function buildPgliteInitErrorMessage(
         '       gbrain reinit-pglite --embedding-model <id> --embedding-dimensions <N>\n' +
         '       (wipes + re-inits + re-syncs; DB-only state is re-derived).\n' +
         '  Deleting .gbrain-lock/ or postmaster.pid does NOT fix this.';
+      break;
+    case 'wal-corrupt':
+      hint =
+        '  Postgres started but PANICked replaying the write-ahead log, so the\n' +
+        '  store is damaged — this is NOT the macOS WASM bug and NOT a lock.\n' +
+        '  The "last known up at" timestamp above is when the brain was last\n' +
+        '  healthy; anything written after it is not recoverable in place.\n' +
+        '  Your markdown is unaffected — the DB holds derived data (chunks,\n' +
+        '  embeddings, links, facts), all of which a re-sync rebuilds. Recover:\n' +
+        '    1. Restore a backup of the brain.pglite directory if you have one, OR\n' +
+        '    2. gbrain reinit-pglite --embedding-model <id> --embedding-dimensions <N>\n' +
+        '       (wipes + re-inits + re-syncs from your sources).\n' +
+        '  Deleting postmaster.pid does NOT fix this; neither does re-running\n' +
+        '  migrations. pg_resetwal would, but PGLite does not ship it.';
+      break;
+    case 'abort-ambiguous':
+      hint =
+        '  PGLite aborted without saying why. Two causes produce this exact\n' +
+        '  message, so run the 5-second test that tells them apart:\n' +
+        '\n' +
+        '      GBRAIN_HOME=$(mktemp -d) gbrain init --pglite --no-embedding\n' +
+        '\n' +
+        '  - That SUCCEEDS  -> the runtime is fine and YOUR STORE is damaged.\n' +
+        '    Recover with `gbrain reinit-pglite` (your markdown is untouched;\n' +
+        '    the DB holds derived data a re-sync rebuilds), or restore a backup.\n' +
+        '  - That FAILS TOO -> the runtime itself cannot start here. Report it\n' +
+        '    with your OS + Bun version; see #223 for prior reports.\n' +
+        '\n' +
+        '  `gbrain doctor` runs the same check. Note a damaged store and a\n' +
+        '  broken runtime are indistinguishable from this string alone, so\n' +
+        '  do not assume the macOS 26.3 bug without running the test above.';
       break;
     case 'unknown':
     default:
--- a/test/pglite-init-classifier.test.ts
+++ b/test/pglite-init-classifier.test.ts
@@ -119,11 +119,100 @@ describe('buildPgliteInitErrorMessage — hint routing', () => {
     expect(msg).not.toContain('issues/223');
   });
 
-  test('all verdicts produce the canonical header line', () => {
-    for (const v of ['bunfs', 'macos-26-3', 'corrupt', 'unknown'] as const) {
-      const msg = buildPgliteInitErrorMessage(v, original);
-      expect(msg.startsWith('PGLite failed to initialize its WASM runtime.')).toBe(true);
+  test('runtime-fault verdicts claim the WASM runtime; store-fault verdicts do not', () => {
+    // #2674: the header used to assert "failed to initialize its WASM runtime"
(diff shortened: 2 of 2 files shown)


time-attack:

The core change is real and correct: a bare `Aborted()` (the exact string quoted in #2674) no longer gets the macOS-26.3 hint, and the "failed to initialize its WASM runtime" header stops being asserted for store-fault verdicts. I ran it: 71 pass on the branch across the classifier/structural/engine-factory/lock suites, typecheck clean, `bun run verify` 32/32 exit 0, and the discrimination test genuinely fails on master (6 behavioral failures, no missing-symbol crashes). But the PR's entire payload is user-facing diagnostic text, and three lines of that text are wrong: it tells users `gbrain doctor` runs the scratch-init check (doctor has no such probe — the PR body admits it), it cites `gbrain reinit-pglite` without its two required flags (pasting it exits 1), and it replaces master's non-darwin "lock contention" cause with a confident "two causes" claim that the repo's own pglite-lock.ts contradicts. All fixes are a few lines of string edits.

### Discrimination test

PASSES — executed, not inferred. I copied `git show origin/master:src/core/pglite-engine.ts` over the branch source and re-ran the PR's test file: 22 pass / 6 fail, exit 1 (/tmp/discrim.txt). All 6 are behavioral, not missing-symbol: `classifyPgliteInitError('Aborted(). Build with -sASSERTIONS for more info.')` returned "unknown" instead of "abort-ambiguous"; the PANIC signature returned "unknown" instead of "wal-corrupt"; the header assertion for corrupt/wal-corrupt/abort-ambiguous got "PGLite failed to initialize its WASM runtime."; and the darwin-vs-linux equality test showed master's platform branch emitting the macOS-26.3 hint on darwin for a bare Aborted() — i.e. the reported bug reproduced under test. Restoring the PR file returns 50/50 pass. The PR's claim of "50 pass here, 6 fail on master" is accurate.

### Blockers

**1.** src/core/pglite-engine.ts:294 — the abort-ambiguous hint ships the sentence "`gbrain doctor` runs the same check." and it is false. `grep -n 'mkdtemp|mktemp|scratch|PGlite' src/commands/doctor.ts src/commands/doctor-remote.ts src/core/onboard/checks.ts` returns ZERO hits; the PR's own body says "gbrain doctor still has no scratch-store init probe (this issue's ask #2)". Worse, in this exact failure state doctor cannot open the brain at all (it routes through createEngine → connect → the same abort), so a user who trusts the line runs doctor, gets the same abort, and concludes the runtime is broken — reproducing the misdiagnosis this PR exists to kill. Fix: delete the sentence.

**2.** src/core/pglite-engine.ts:288 — the hint tells the user `gbrain reinit-pglite` bare. src/commands/reinit-pglite.ts:213-219 hard-fails on missing flags: running `gbrain reinit-pglite` prints "--embedding-model <provider:model> is required." and exits 1. The sibling hints get this right (the pre-existing `corrupt` hint and this PR's own `wal-corrupt` hint both spell out `--embedding-model <id> --embedding-dimensions <N>`), so the one paste-ready command in the new hint is the one that doesn't run.

**3.** src/core/pglite-engine.ts:280-296 — abort-ambiguous asserts "Two causes produce this exact message" and its discriminator says a successful scratch init means "YOUR STORE is damaged." The repo documents a third cause in two places: src/core/pglite-lock.ts:6 ("when two processes try to connect, PGLite throws `Aborted()` because it can't handle concurrent access") and src/core/pglite-engine.ts:310 ("Acquire file lock to prevent concurrent PGLite access (crashes with Aborted())"). Measured: `buildPgliteInitErrorMessage('abort-ambiguous','x','linux').includes('lock') === false`, whereas master's non-darwin branch named "another gbrain process holding the database (lock contention)" — so on non-darwin this PR DELETES a correct cause and routes it to "your store is damaged → reinit". Reachable whenever the gbrain lock is bypassed (non-gbrain process on the data dir, two path spellings/symlink to the same dir, pre-lock gbrain version, PID-reuse steal). Bounded (reinit renames to .bak, TTY-confirmed, prints a rollback line), but it is a diagnosis regression shipped by an accuracy fix. Fix: keep lock contention as a named third possibility, or stop claiming the list is exhaustive.

**4.** src/core/pglite-engine.ts:206-210 — `database system was interrupted` is in the wal-corrupt regex, and by itself it is a BENIGN log line that precedes every successful crash recovery, not evidence of damage. Measured: the 3-line stderr `"LOG: database system was interrupted; last known up at …\nLOG: … automatic recovery in progress\nAborted()"` (no PANIC anywhere) classifies as `wal-corrupt` → header "PGLite could not open your brain" + "the store is damaged" + wipe-and-reinit. It cannot fire today because the only production caller (line 353) passes `err.message`, never PGLite's stderr — i.e. the whole `wal-corrupt` verdict is currently dead code — but the PR's stated reason for those regexes is a future caller that DOES surface stderr, which is precisely when this misfires. Fix: require the PANIC/`invalid resource manager ID`/`invalid checkpoint record` signatures and drop the bare `database system was interrupted` alternative.

### Not verified

- Did not reproduce a real WAL-corrupted PGLite store end-to-end, so the PR's field claim that PGLite surfaces only `Aborted()` to JS while printing the PANIC to its own stderr is taken from the code path (single caller passes err.message) rather than from a live corrupted store. It is consistent with everything I could read and with issue #2674's verbatim error text.
- Did not run the full unit suite or e2e (DATABASE_URL-gated) — the diff contains no SQL, no schema, and no engine method, and I grepped repo-wide for the only externally-observable string it changes (`failed to initialize its WASM runtime`): the sole assertion outside src/ is the classifier test this PR updates.
- No CHANGELOG.md / VERSION / package.json entry in the PR (diff is 2 files) and docs/architecture/KEY_FILES.md:34 still documents the verdict union as `'bunfs' | 'macos-26-3' | 'unknown'` (already stale for `corrupt`, now two verdicts further behind). check-key-files-current-state.sh passes since it only guards release-clause markers and size, so this is maintainer-side /ship + /document-release work, not a CI failure.
- PR title `fix(pglite): stop blaming macOS…` violates the CLAUDE.md IRON RULE that every PR title starts with `vMAJOR.MINOR.PATCH.MICRO`; trivially fixable with gh pr edit --title.

- `gh pr view 3519 — exit 0`
- `gh pr diff 3519 > /tmp/pr3519.diff — exit 0`
- `gh pr checkout 3519 (branch fix/pglite-abort-diagnosis-2674 @ 8ab8178a) — exit 0`
- `bun install — exit 0`
- `bun test test/pglite-init-classifier.test.ts test/fix-wave-structural.test.ts — exit 0 (50 pass / 0 fail)`
- `bun test test/engine-factory.test.ts test/pglite-init-classifier.test.ts test/fix-wave-structural.test.ts test/pglite-lock.test.ts — exit 0 (71 pass / 0 fail)`
- `bun run typecheck — exit 0 (tsc --noEmit clean)`
- `bun run verify — exit 0 (pass=32 fail=0, all checks green)`
- `DISCRIMINATION: cp master's src/core/pglite-engine.ts over the branch; bun test test/pglite-init-classifier.test.ts — exit 1 (22 pass / 6 fail, all behavioral); restored branch file afterwards (git diff --stat empty)`
- `bun /tmp/probe.ts (classifier probe over 9 realistic abort/recovery strings) — exit 0`
- `grep -n 'mkdtemp|mktemp|scratch|PGlite' src/commands/doctor.ts src/commands/doctor-remote.ts src/core/onboard/checks.ts — 0 hits (no scratch-init probe)`
- `grep -c 'classifyPgliteInitError|InitFailure' src/core/postgres-engine.ts — 0`
- `read src/commands/reinit-pglite.ts parseArgs (lines 198-232) — --embedding-model and --embedding-dimensions are hard-required`

</details>

---

time-attack:
Review finding (one-sentence fix): the classifier work is good — the `abort-ambiguous` regex is start-anchored inside the `PGlite.create` catch, so it can't misattribute unrelated aborts, and there's no conflict with the v0.42.57.0 lock work. The blocker is ironic: while fixing false user-facing messages, the new `abort-ambiguous` hint asserts '`gbrain doctor` runs the same check' — but no scratch-store init probe exists in `src/commands/doctor.ts` (your own PR body concedes this). Drop or soften that sentence (e.g. 'run `gbrain doctor` for broader diagnostics') and this merges. Worth noting #3569 is adding exactly that probe; if it lands first, the sentence becomes true and this is moot.

time-attack:
Thanks for this — the diagnosis here was right, and measured rather than guessed: the bare `Aborted()` was WAL/checkpoint damage in the store, not an OS WASM bug, and the old error copy was blaming the wrong layer. That analysis matched what we independently confirmed. Master has since shipped v0.42.75.0 (#3901), which lands the same reclassification (the `macos-26-3` verdict is gone; the real production `Aborted()` message now routes to `wasm-abort`) and goes one step further than a better error message: it ports pg_resetwal into a bounded in-place auto-repair with restore-on-failure, plus a manual `gbrain pglite-repair` command and doctor integration — so the store is fixed rather than the user being taught to triage it. That supersedes this PR's approach, and the branch now conflicts with the reworked classifier. Closing as superseded by #3901; genuinely appreciate the end-to-end reproduction work here — it's the kind of evidence that moved this class of bug from folklore to fixed. Your companion #3569 (doctor-side) will be evaluated against the new doctor pglite_data_dir check separately.

PR #3523, opened 2026-07-28
fix(doctor): stop reporting Windows-drive image paths as missing under WSL (#1835)

### LINKED ISSUE #1835: doctor image_assets treats Windows D:/ paths as missing under WSL

time-attack:
Fixes #1835.

## Problem

`files.storage_path` rows written by a Windows gbrain install carry Windows drive paths (`D:/cicada3301/lost9999/img.jpg`, `D:\...`). On POSIX, `path.isAbsolute()` is false for those, so the `image_assets` doctor check joined them onto the repo root and statted a path that can never exist (`/mnt/d/.../D:/cicada3301/...`). Result: a false-positive WARN advising "restore from git" — implying data loss that never happened.

## Fix

New `src/commands/doctor-asset-paths.ts` (`resolveAssetPath` + `parseWslAutomountRoot` + `detectWslMountRoot`), consumed by the `image_assets` block:

- **WSL** (linux + `microsoft` in `/proc/version`): translate `D:/x` → `<automount root>/d/x`, then stat that. The automount root is read from `/etc/wsl.conf` `[automount] root` (configurable per WSL docs), defaulting to `/mnt`.
- **macOS / plain Linux**: a Windows drive path is unresolvable on this host — the check now SKIPS it and reports `(N Windows-drive path(s) skipped — not resolvable on this platform)` rather than inventing a never-exists path. Skipped paths are excluded from the checked denominator; genuinely vanished assets still WARN.
- **win32**: drive paths are absolute; statted as-is (unchanged).
- Non-drive absolute/relative paths keep their existing behavior.

## Coordination with #3478 (open, touches the same block)

#3478 rewrites this block into `resolveImageAssetPath(storagePath, sourceLocalPath, fallbackRepoRoot)` with **no drive translation**. This PR deliberately puts the translation in its own module so whichever lands second has a one-line integration: `resolveImageAssetPath` should call `resolveAssetPath` (or apply the same foreign-skip) instead of bare `isAbsolute`/`join`. Happy to rebase onto #3478 if it merges first.

## Discrimination proof (test fails on master behaviorally)

`test/doctor-image-assets-wsl.test.ts` drives the master-existing `buildChecks` seam (imports nothing added by this PR). Against unmodified master:

```
✗ D:/ path is skipped with a note, not reported missing   — expect "ok", received "warn"
✗ backslash D:\ path is also skipped                       — expect "ok", received "warn"
✗ drive path skip does not mask a genuinely vanished asset — expect "warn" for the real missing file only
✓ present relative asset still resolves against repoRoot (regression guard)
```

With the fix: 4/4 pass. Pure translation-logic tests (`test/doctor-asset-paths.test.ts`, 13 tests, injected platform + mount root so they run identically anywhere) pass.

## Verification

- `bun test test/doctor-asset-paths.test.ts test/doctor-image-assets-wsl.test.ts` → 17 pass / 0 fail
- `bun run typecheck` → clean
- `bun run verify` → 32 pass / 0 fail

**UNVERIFIED-ON-PLATFORM:** no real Windows/WSL host was available; the WSL translation path and `/proc/version` / `/etc/wsl.conf` detection were not executed on WSL. The unit tests pin the intended mapping via injected platform/mount-root parameters, and the behavioral tests cover the macOS/plain-Linux skip path that this environment can execute.


time-attack:

The core fix is real and its tests genuinely discriminate — I reverted ONLY the doctor.ts hunk (keeping the new module + both test files) and 3 of the 4 behavioral tests failed (14 pass / 3 fail); with the hunk restored, 17/17 pass. typecheck clean. No engine/JSONB/migration/knobs-hash/source-scope/dry-run surface is touched (jsonb guards run clean, doctor is read-only stat-only). The flagged traversal danger does NOT exist as a new class: I probed 10 adversarial inputs and `D:/../../../etc/passwd` does normalize to `/etc/passwd`, but (a) master already statted POSIX-absolute `storage_path` as-is and already normalized `..` out of `join(repoRoot, …)`, so escape parity is unchanged; (b) the resolved path is only `statSync`'d, never read; (c) the WARN prints `r.storage_path` (the original), never the resolved abs — so no new oracle or read. Native POSIX behavior is unchanged (probe + the PR's own regression guard). What blocks merge is three one-line items, not the design: PR title violates the version-first IRON RULE; test fixtures hardcode the reporter's private vault directory/handle (`cicada3301/lost9999`) against the privacy rule; and the new drive-detection regex silently mis-handles a repo-relative path whose first segment is `<letter>:` (legal on POSIX), which master resolved correctly. Recommended hardening that also closes the mirror-image gap: stat the repo-relative resolution FIRST, then fall back to drive translation.

### Discrimination test

Executed. `git diff HEAD~1 HEAD -- src/commands/doctor.ts | git apply -R` (stubs the pre-PR `isAbsolute ? path : join(repoRoot, path)` block back in, keeps the new module and BOTH new test files), then `bun test test/doctor-asset-paths.test.ts test/doctor-image-assets-wsl.test.ts` → 14 pass / 3 fail. The 3 failures are exactly the behavioral claims: "D:/ path is skipped with a note, not reported missing", "backslash D:\\ path is also skipped", "drive path skip does not mask a genuinely vanished asset". The 4th behavioral test ("present relative asset still resolves against repoRoot") passes both ways and is declared as a regression guard, which is legitimate. Re-applying the hunk → 17 pass / 0 fail. The 13 pure-unit tests target a file that does not exist on master, so they cannot discriminate by construction — that is disclosed, and the behavioral file carries the proof. Verdict: tests are NOT tautological.

### Blockers

**1.** src/commands/doctor-asset-paths.ts:32 — WINDOWS_DRIVE_RE `/^([A-Za-z]):[\\/](.*)$/` also matches a legitimate repo-RELATIVE path whose first segment is a single letter plus colon (legal filename on Linux/macOS). Failing input: a brain repo containing directory `C:` with `storage_path='C:/x.png'` and the file present on disk — master resolved `<repoRoot>/C:/x.png` and reported ok; this PR reports it foreign and drops it from the checked denominator on macOS/plain Linux (silently unverified), and on WSL mistranslates it to `/mnt/c/x.png` → false `missing from disk … restore from git`. Verified by direct probe of resolveAssetPath. Fix: stat the repo-relative resolution first and only drive-translate when that misses.

**2.** test/doctor-asset-paths.test.ts:192 and test/doctor-image-assets-wsl.test.ts:70 — fixtures hardcode `D:/cicada3301/lost9999/img.jpg`, i.e. the issue reporter's private vault directory plus their handle, copied out of issue #1835. CLAUDE.md privacy rule bans real people/paths in checked-in artifacts (contributor CREDIT in CHANGELOG is allowed; a private filesystem path in test code is not). Replace with `D:/example/assets/img.jpg`.

**3.** PR title `fix(doctor): stop reporting Windows-drive image paths as missing under WSL (#1835)` violates the version-first IRON RULE — must lead with `vMAJOR.MINOR.PATCH.MICRO` (e.g. `v0.42.68.0 fix(doctor): …`). Fix with `gh pr edit 3523 --title`.

### Not verified

- The WSL branch itself was never EXECUTED on a real WSL host: /proc/version 'microsoft' detection and /etc/wsl.conf reading in computeWslMountRoot() are covered only by injected-platform unit tests (the author discloses this as UNVERIFIED-ON-PLATFORM). I did not Docker-bind-mount a fake /proc/version to execute the production composition, per the execution-footprint limit. Reviewed by inspection: composition is correct (production call passes no opts → process.platform + detectWslMountRoot, memoized).
- Mirror-image gap NOT fixed and not a regression: doctor run on native win32 against rows written under WSL (`/mnt/d/…`) hits `win32.isAbsolute` → true → stats `C:\mnt\d\…` → the identical false 'restore from git' WARN. Same bug class, same reporter's dual-boot setup. Pre-existing on master, so a gap rather than a defect; the stat-repo-relative-first fix above would cover it.
- Collision with open PR #3478, which rewrites the same image_assets block into resolveImageAssetPath(storagePath, sourceLocalPath, fallbackRepoRoot) and is already CONFLICTING with master. Whoever lands second needs manual integration; note that #3478's source-root resolution alone does NOT fix #1835 (join(sourceRoot, 'D:/…') still misses).
- Cosmetic: when every row is foreign the ok-message reads '0 image(s) all present on disk (248 Windows-drive path(s) skipped …)'. Honest but odd; not tested.
- Did not run the full unit suite, bun run verify, or e2e (execution-footprint limit). No Postgres-only behavior in this diff, so no Docker run was warranted.
- Branch is behind master (does not contain origin/master); I reviewed it as-is and did not rebase.

- `gh pr view 3523`
- `gh pr diff 3523 > /tmp/pr3523.diff`
- `gh issue view 1835`
- `git fetch origin pull/3523/head:pr3523-rev && git checkout pr3523-rev`
- `ln -s <main>/node_modules ./node_modules  (deps missing in worktree; removed after)`
- `bun test test/doctor-asset-paths.test.ts test/doctor-image-assets-wsl.test.ts  → 17 pass / 0 fail`
- `git diff HEAD~1 HEAD -- src/commands/doctor.ts > /tmp/doctorhunk.diff && git apply -R /tmp/doctorhunk.diff && bun test <same two files>  → 14 pass / 3 fail (discrimination)`
- `git apply /tmp/doctorhunk.diff  (restore PR state)`
- `bash scripts/check-key-files-current-state.sh  → EXIT=0`
- `bash scripts/check-jsonb-pattern.sh  → EXIT=0 (also ran check-jsonb-params: clean)`
- `bun run trav3523.ts  (10-case adversarial resolveAssetPath probe: D:/../../../etc/passwd, D:/../../etc/shadow, d:/, /etc/passwd, ../../../etc/passwd, assets/ok.png, \\\\server\\share, D:img.png, //wsl$/…, darwin C:/…)`
- `grep -rn storage_path src --include=*.ts | grep -E 'isAbsolute|join\(|statSync|readFile'  (only the doctor site resolves storage_path to disk — no sibling caller left broken)`
- `grep -rli 'lost9999|cicada3301' . --exclude-dir=node_modules`
- `gh pr view 3478 --json state,mergeable + gh pr diff 3478 (overlap check)`
- `bash scripts/check-engine-dynamic-import.sh → 127, script does not exist on this branch (landed on master after the branch point)`

</details>

---

time-attack:

PR #3527, opened 2026-07-28
v0.42.70.0 fix(windows): resolve filesystem paths with fileURLToPath, not URL.pathname

daragao3:
## What

On Windows, deriving a filesystem path from `URL.pathname` yields `/C:/Users/...` — a URL path with a leading slash that no Win32 API accepts. This lands in 24 places across the tree. It is an exact identity transform on POSIX, so Linux CI cannot catch any of it.

**Production impact:** `src/commands/schema.ts::packPathByName` used `dirname(new URL(import.meta.url).pathname)`, so every name in `BUNDLED_PACK_NAMES` resolved to `null` on Windows. `gbrain schema use gbrain-base` and `gbrain schema active` had nothing to load.

**Contributor impact:** 23 test call sites passed the bad path as a `Bun.spawn` `cwd:`. libuv fails the chdir and reports:

```
ENOENT: no such file or directory, uv_spawn 'bun'
```

That error names `argv[0]`, not the missing directory, so it reads as "bun is not installed" while bun resolves fine via PATHEXT. Interpolated into a script argument the same expression gives `Module not found "/C:/.../src/cli.ts"`. On a full run it also surfaced as an unhandled error between tests and aborted the run.

## Commits

| SHA | What |
|---|---|
| `eaea337d` | Fix all 24 sites with `fileURLToPath()`; add `test/helpers/repo-root.ts` |
| `f98f0e3f` | Make the `resolveBunGlobalRoot` `BUN_INSTALL` assertion separator-agnostic |
| `00ad1f4c` | Add `scripts/check-url-pathname-fs.sh` CI guard, wire into `verify` |
| `f862d642` | v0.42.70.0 version bump + CHANGELOG |
| `6f6d170c` | Doc sync (CONTRIBUTING, TESTING, README, llms bundle) |

## Changes by area

**Code (1 file)**
- `src/commands/schema.ts` — `packPathByName` resolves the bundled-pack directory with `fileURLToPath()`. The only production call site.

**Tests (11 files)**
- New `test/helpers/repo-root.ts` exports `REPO_ROOT` (absolute, native separators, no trailing separator) and `repoPath(...segments)`, replacing 23 hand-rolled expressions across ten test files. `test/helpers/repo-root.test.ts` pins its invariants.
- `test/upgrade.serial.test.ts` builds its `BUN_INSTALL` expectation with `join()` instead of a hardcoded POSIX literal. `resolveBunGlobalRoot` returns a native path because the value is used as an `execFileSync` cwd and printed as a paste-ready recovery hint, so the test was wrong, not the implementation. `posix.join` returns the old string byte for byte, so this is a no-op off Windows.
- `test/integrations.test.ts` — one `getRecipeDirs` assertion normalizes separators before `endsWith('/recipes')`.

**CI (3 files)**
- New `scripts/check-url-pathname-fs.sh`. Flags three shapes, all requiring a *file* URL: the single expression (including formatter-split calls and a leading-dot continuation), a `file:` literal or a `pathToFileURL()` result, and the two-step form where a variable holds the URL and is read later.
- Non-file URLs stay unflagged. Parsing a connection string, routing an HTTP request, and assigning `u.pathname = '/db'` are correct usage — `test/e2e/helpers.ts` and `test/e2e/self-upgrade-binary-swap.test.ts` both do this and stay clean.
- Comments are stripped so the JSDoc documenting the rule does not self-trip; `url-pathname-guard-ok` opts a line out.
- `git grep` prefilters candidates, so a full-tree scan reads 17 files rather than 2124 (2m6s to 23s on Windows, inside verify's 120s per-check cap).
- Wired into `package.json` (`check:url-pathname`, `check:all`) and `scripts/run-verify-parallel.sh`. Invoked through `bash`, matching the convention #3506 established.
- `test/check-url-pathname-fs.test.ts` validates 6 known-bad shapes (all flagged, exit 1) and 7 known-good ones taken from real call sites (all clean, exit 0).

**Docs (6 files)**
- `CLAUDE.md` gains the rule as a cross-cutting invariant, alongside the JSONB and engine-parity entries.
- `docs/architecture/KEY_FILES.md` documents the check and the helper.
- `CONTRIBUTING.md` carries the rule in the Windows section, next to the existing bash-dispatch and CRLF notes.
- `docs/TESTING.md` gains a "Filesystem paths in tests" section and registers the two new test files. It also corrects two pre-existing counts: it listed `check:all` at 22 and `verify` at "~30" against an actual 24 and 33.
- `README.md` gains a troubleshooting entry for the user-visible symptom.
- `llms-full.txt` regenerated via `bun run build:llms`.

## Testing

Run on Windows 11, since that is the only platform where the bug is observable.

| Check | Result |
|---|---|
| `bun run typecheck` | EXIT=0 |
| `repo-root` + `check-url-pathname-fs` + `schema-packpath` + `upgrade.serial` | 29 pass / 0 fail |
| `bun test test/build-llms.test.ts` | 12 pass / 0 fail |
| 7-file affected slice (`commands/`, `cli`, `check-update`, `cli-exit-verdict-pin`, `book-mirror`, `integrations`) | 186 pass / 1 fail |
| `bash scripts/check-url-pathname-fs.sh` | exit 0, 17 candidate files scanned |

The 1 failure is `integrations.test.ts > executeHealthCheck > command returns ok for exit 0`. It spawns `argv: ['true']`, a POSIX-only binary with no Windows equivalent. The test is byte-identical to `origin/master` and untouched by this branch; it passes on Linux.

## Compatibility

No behavior change on macOS or Linux. `fileURLToPath()` and `URL.pathname` return the same string there, and `posix.join` reproduces the old assertion literal exactly. The only platform where anything moves is Windows, where it moves from broken to working.

## Note on the version bump

This PR carries a VERSION / package.json / CHANGELOG bump to 0.42.70.0 and a version-first title. If that does not fit the fix-wave process in `docs/RELEASING.md`, please renumber or drop the bump commit (`f862d642`) and take the rest on their own — the code commits do not depend on it. 0.42.70.0 was chosen to clear every version currently claimed by an open PR (0.42.67.1, 0.42.67.3, 0.42.68.0, 0.42.69.0).


time-attack:

A first reviewer concluded this was mergeable. Two independent refuters then re-ran it with different lenses; 1 of 2 killed the recommendation. Their findings:

### Refutation 1

MERGE refuted; NEEDS_CHANGES. The production fix (src/commands/schema.ts::packPathByName -> fileURLToPath) and the 23 test-site conversions are correct, typecheck is clean, and the 140 touched tests pass. But the PR's new CI guard — which the PR itself installs in CLAUDE.md as a cross-cutting invariant and which KEY_FILES.md calls "the only backstop" because Linux CI structurally cannot catch this bug class — fails open, and I reproduced it three ways.

In the embedded awk strip(), the block-comment /* scan runs BEFORE the // line-comment truncation. So a line comment containing /* (e.g. `// (via the TS registry, not skills/migrations/*.md)`) sets inblock=1 and every subsequent line in that file is blanked until a */ appears, which never comes. Proofs: (1) a two-line fixture whose second line is the canonical banned expression -> guard prints "ok", exits 0; (2) run against current origin/master (c6dc0adf), the PR's guard reports only 23 of the 24 real violations, silently missing test/upgrade.serial.test.ts:208, poisoned by the //-comment at line 201 of that same file — the guard would have exited 0 with that site still broken; (3) moving the // truncation above the /* loop takes it to 24/24 on master with zero new false positives, and all 7 of the PR's good-case fixtures stay clean. Latent scope: 20 tracked source files already carry a poisoning line, including src/commands/upgrade.ts:696, src/commands/sync.ts:1109, src/commands/doctor.ts:6325, src/commands/serve-http.ts:1694. The guard's own self-test (6 bad / 7 good) writes each shape to its own one-or-two-line file, so comment-state carry-over is untested.

Discrimination test independently performed: stubbing `new URL('../..', import.meta.url).pathname` back into test/helpers/repo-root.ts leaves test/helpers/repo-root.test.ts at 4 pass / 0 fail on POSIX (schema-packpath likewise). The author discloses this, but it means the shell guard is the sole Linux-side regression protection — and it is the thing that is broken.

Invariants checked and clean: no engine files touched (parity N/A), no JSONB casts, no sourceScopeOpts/source-isolation paths, no ctx.remote, no knobs_hash, no MIGRATIONS DDL, no dry-run paths, PR title is version-first. Guard produces zero false positives on the real tree (the 6 legitimate .pathname reads in src/ stay clean). packPathByName is genuinely the only production call site (grepped every .pathname in src/), so no sibling caller is left broken. Merge into origin/master conflicts only in CLAUDE.md, llms-full.txt, package.json (mechanical). check-key-files-current-state.sh ok; build-llms freshness 12/12.

Nits: VERSION loses its trailing newline (0.42.67.0\n -> 0.42.70.0 with no LF); check-trailing-newline.sh does not cover extensionless files. This is also the only check:* script depending on git/git grep, though its non-git find fallback is fail-safe.

Unverified for execution-footprint reasons: bun run verify, bun run test, full suite, e2e/Docker, 10-shard CI binning, and any Windows execution (the target platform) — Windows behavior rests on code reading plus the author's report.

Fix is small and local: replace strip()'s two-phase comment handling with one left-to-right scan taking whichever of // or /* comes first (a naive reorder has a rarer inverse hole on `/* see // note */`), and add a fixture pairing a /*-bearing line comment with a violation on the next line. Everything else can land as-is.

**Defect the first reviewer missed:** scripts/check-url-pathname-fs.sh: the awk strip() scans for block-comment `/*` before truncating `//` line comments, so any line comment containing `/*` (e.g. a glob like `skills/migrations/*.md`) sets inblock=1 and blanks every subsequent line in that file until a `*/` appears — which never comes. The guard then silently reports clean. Reproduced: on current origin/master the guard finds 23 of the 24 real violations, missing test/upgrade.serial.test.ts:208 (poisoned by the `//` comment at line 201 of the same file); a 2-line fixture with `// glob note: skills/migrations/*.md` followed by `const REPO = new URL("..", import.meta.url).pathname;` exits 0. 20 tracked source files already contain such a poisoning line, including src/commands/upgrade.ts:696, src/commands/sync.ts:1109, src/commands/doctor.ts:6325, src/commands/serve-http.ts:1694. Moving the `//` truncation above the `/*` loop yields 24/24 on master with no new false positives and all 7 good-case fixtures still clean. Not covered by test/check-url-pathname-fs.test.ts, and not the caveat the script header discloses (that one is line-scoped, this is file-scoped).

---

time-attack:

Real, unfixed, root-cause bug fix squarely in the MERGE lane: origin/master src/commands/schema.ts:372 still uses dirname(new URL(import.meta.url).pathname), which yields /C:/... on Windows so every BUNDLED_PACK_NAMES entry resolves to null (gbrain schema use/active dead on Windows), plus 20+ test call sites with the same pattern; the v0.42.69.0 wave did NOT touch this (its WSL fix was doctor-only, and no master commit since base 6920744d touches schema.ts or adds fileURLToPath). Fix is correct (fileURLToPath is identity on POSIX), consolidates 23 hand-rolled expressions into test/helpers/repo-root.ts, and adds a CI guard wired into verify/check:all. Verified in a /tmp worktree merge: typecheck EXIT=0 and all 201 tests across the 11 touched test files pass on the merged tree.

**Conflict state vs current master:** Six conflicting files, all trivial: VERSION/package.json/CHANGELOG.md/CLAUDE.md/llms-full.txt (version-metadata + invariant-block unions; PR's 0.42.70.0 still clears master's 0.42.69.0; llms-full.txt regenerates clean via bun run build:llms) and test/integrations.test.ts (one import-union hunk). package.json check:all needs the union of master's check-engine-dynamic-import.sh and the PR's check-url-pathname-fs.sh — verified both resolve cleanly. The fix fully applies post-wave; nothing on master supersedes it.

**Risk if merged as-is:** Guaranteed red CI if merged with only textual conflict resolution: the v0.42.69.0 wave added two NEW files carrying the flagged pattern that git auto-merges silently — test/embed-exit-code-3037.serial.test.ts:22 and test/reindex-frontmatter-pglite-spawn.serial.test.ts:27 — and the PR's new scripts/check-url-pathname-fs.sh (in verify + check:all) exits 1 on them (reproduced in worktree). Required merge-time fix is one line each (replace with `import { REPO_ROOT as REPO } from './helpers/repo-root.ts';`); after that the guard exits 0, typecheck passes, and both files' tests pass (162/162). Secondary: keep the 0.42.70.0 bump only if it still clears the version queue at merge time (author explicitly offered to renumber/drop commit f862d642).

time-attack:

click once rebased.

PR #3530, opened 2026-07-28
fix(doctor,migrate): detect and repair a missing pages(source_id, slug) unique index (#550)

### LINKED ISSUE #550: put_page fails with "no unique or exclusion constraint" — writes broken when pages_source_slug_key is missing, doctor doesn't catch it

time-attack:
Fixes #550.

## The bug — p0, all writes dead, no recovery path

`put_page` upserts with `ON CONFLICT (source_id, slug)` on both engines (`src/core/postgres-engine.ts`, `src/core/pglite-engine.ts`). When the arbiter unique index on `pages(source_id, slug)` is missing, **every write** fails with:

```
there is no unique or exclusion constraint matching the ON CONFLICT specification
```

while reads keep working, so nothing surfaces it. Reproduced on a real PGLite engine migrated to head:

```
BASELINE putPage ok -> {"id":1,...}
unique CONSTRAINTS on pages: [{"conname":"pages_source_slug_key"}]
after dropping it:      []
AFTER-DROP putPage THREW -> there is no unique or exclusion constraint matching the ON CONFLICT specification
```

**Both recovery paths were no-ops:** `initSchema()` is additive-only and does not restore the constraint, and `apply-migrations --force-schema` just re-runs `runMigrations` from the current config version — nothing is pending at head, so a wedged brain stayed wedged forever.

Root cause: migration v23 (Postgres) / v21 (PGLite) guarded the `ADD CONSTRAINT pages_source_slug_key` on `pg_constraint.conname` — matched by **name**, not by columns. Any brain stamped >= v23 without that DDL actually executing (fresh-init blob stamped to head, externally-created `pages`, a later drop/rename) never gets it back.

## The fix

1. **Shared column-shape probe** (`src/core/pages-unique-repair.ts`): matches what `ON CONFLICT` inference actually accepts — a **non-partial, non-expression unique index** (constraint-backed or bare `CREATE UNIQUE INDEX`) whose key columns are exactly `{source_id, slug}`, in either order. Partial indexes are correctly rejected (inference refuses them without a matching `WHERE`). Skips cleanly (not FAIL) when the `pages` table or `source_id` column doesn't exist yet — a fresh/pre-v21 brain is not broken.
2. **`pages_unique_index` doctor check**: registered in `BRAIN_CHECK_NAMES`, wired into BOTH `buildChecks()` and `doctorReportRemote()`. FAILs loudly with the literal repair SQL (below). It deliberately does not point at `--force-schema`.
3. **Column-based migration guards**: v23 (Postgres) and v21 (PGLite) now use the shared shape-guarded DDL. Idempotent: a satisfying index under **any** name is a no-op (no duplicate constraint); if the canonical name is taken by a wrong-shaped relation, the constraint is added under a fallback name instead of failing (the name is not load-bearing — inference is by shape).
4. **Every-pass self-heal**: `runMigrations` runs the repair on every pass, including the no-pending path — the same always-run slot as the #2038 `idx_timeline_dedup` heal (same drift class). Since `initSchema()` ends in `runMigrations()` on both engines, a wedged brain heals on its next init/migrate pass. The repair never auto-deletes duplicate pages: if unconstrained duplicates snuck in, `ADD CONSTRAINT` fails naming the duplicated key and that cleanup stays with the operator.

## Operator repair (immediate, any version)

```sql
ALTER TABLE pages ADD CONSTRAINT pages_source_slug_key UNIQUE (source_id, slug);
```

`gbrain doctor` now detects the condition and prints exactly this.

## Discrimination proof (tests fail on master, pass here)

Test file + support module copied onto an unmodified `origin/master` checkout (6920744d):

```
✗ #550 ... > reproduction: dropping the constraint kills put_page; a plain migrate pass heals it
    error: there is no unique or exclusion constraint matching the ON CONFLICT specification
    (routine: "infer_arbiter_indexes" — put_page still dead after runMigrations)
✗ #550 ... > doctor check FAILS loudly with the literal repair SQL (and never mentions --force-schema)
    error: doctor.ts does not export pagesUniqueIndexCheck (#550 check missing)
✗ #550 ... > missing pages table is OK/skip, not FAIL — a fresh brain is not broken
    error: doctor.ts does not export pagesUniqueIndexCheck (#550 check missing)
 4 pass / 3 fail
```

On this branch: **7 pass / 0 fail** (`bun test test/pages-unique-repair.test.ts`).

## Verification

- `bun run typecheck` — clean
- `bun run verify` — 32/32 green
- `bun test test/pages-unique-repair.test.ts` — 7/7
- Neighbors: `test/migrate.test.ts`, `test/doctor.test.ts`, `test/doctor-categories.test.ts`, `test/doctor-report-remote.serial.test.ts`, `test/timeline-dedup-repair.test.ts`, `test/schema-bootstrap-coverage.test.ts`, `test/bootstrap.test.ts`, `test/multi-source-integration.test.ts` — all green
- Engine parity: no engine methods added; the probe/repair go through `engine.executeRaw` and behave identically on both engines (PGLite is real Postgres). The v21 PGLite and v23 Postgres guards carry the identical shared DDL.


time-attack:
Closing at the maintainer's decision. Recording what that leaves unfixed, so it's findable later.

**#550 stays broken and stays unrecoverable.** A brain that loses the unique index on `pages(source_id, slug)` has **every write fail** — `put_page` errors with `there is no unique or exclusion constraint matching the ON CONFLICT specification`. Reads keep working, so nothing surfaces it. I verified during review that the user cannot self-recover: `initSchema()` is a no-op at schema head, and `apply-migrations --force-schema` only re-runs from the current version, so it does nothing either. There is no doctor check that detects it. A wedged brain stays wedged.

Incidence is unknown but non-zero — it needs a brain that reached ≥v23 without the DDL actually executing (a fresh-init blob stamped to head, an externally-created `pages` table, or a later drop/rename). The v23 guard matches the constraint **by name** rather than by columns, which is what allows it.

**Two smaller things that could be revisited independently**, if the objection was to the automatic repair rather than to the work:

1. **Detection only.** The doctor check on its own — probing `pg_index` for a non-partial unique index on those columns *by shape*, failing with the literal `ALTER TABLE pages ADD CONSTRAINT pages_source_slug_key UNIQUE (source_id, slug);`. No write to the migration path, no auto-repair, just an accurate diagnosis and a paste-ready command. That's the part with essentially no blast radius.
2. **The v23 guard fix alone** — matching by columns instead of by name, so re-running migrations can repair a brain that already has the problem.

Not reopening either; just noting they're separable if the concern was the every-install write rather than the diagnosis.

Thanks to the author for the care taken here — refusing to auto-delete duplicate pages and naming the blocking duplicate instead was the right instinct, and the by-shape probe is the correct way to detect this.

Issue #3461, opened 2026-07-28
insertChunks stamps DEFAULT_EMBEDDING_MODEL on chunk provenance: getEmbeddingModel() throws, so the || guard is dead code


mariokarras:
## Summary

`insertChunks` stamps the compile-time `DEFAULT_EMBEDDING_MODEL` onto `content_chunks.model` whenever the AI gateway is unconfigured — corrupting embedding provenance for chunks whose vectors came from the config-resolved provider.

The guard intended to prevent this is dead code, because **`getEmbeddingModel()` throws rather than returning falsy**.

## The code

`src/core/postgres-engine.ts` (~line 2452, `insertChunks`):

```ts
let resolvedModel: string = DEFAULT_EMBEDDING_MODEL;   // 'zeroentropyai:zembed-1'
try {
  const gw = await import('./ai/gateway.ts');
  resolvedModel = gw.getEmbeddingModel() || resolvedModel;
} catch {
  // Gateway unconfigured (unit tests / pre-connect): keep the default.
}
```

The `|| resolvedModel` fallback presumes `getEmbeddingModel()` returns a falsy value when unconfigured. It does not — `src/core/ai/gateway.ts:697` calls `requireConfig()`, which throws:

```
AIConfigError: AI gateway is not configured. Call configureGateway() during engine connect.
```

So in any unconfigured context the `||` never evaluates. Control goes to the `catch`, which leaves `resolvedModel` at the compile-time constant. The chunk is written with a provenance label naming a provider that did not produce its vector.

The comment directly above this block describes the exact bug it is meant to fix:

> *"the old `chunk.model || DEFAULT_EMBEDDING_MODEL` fallback stamped the hardcoded default (e.g. `zeroentropyai:zembed-1`) onto rows whose vectors were produced by a different, config-resolved model — corrupting the provenance that signature-drift staleness + dim-migration logic trust."*

That fix replaced the `chunk.model ||` term but kept the hardcoded constant underneath, so the failure mode survives on the throw path.

## Reproduction

```ts
const gw = await import('./src/core/ai/gateway.ts');
gw.getEmbeddingModel();
// -> throws AIConfigError, rather than returning undefined
```

[...]

PR #3532, opened 2026-07-28
fix(embed): stop one bad chunk darkening a page, and exit non-zero on failures (#3037)

time-attack:
Fixes the two user-impactful halves of #3037.

## What this PR fixes

**Sub-problem 2 — one oversized chunk darkened its ENTIRE page.** All three embed paths (`embedPage`, `embedAll --all`, `embedAllStale`) batched a whole page's chunks in one `embedBatch` call inside a catch that only logged to stderr; when the batch threw, `upsertChunks` never ran and every sibling chunk stayed NULL (~8.6x blast radius from a single bad chunk). The existing `embed-skip` guard doesn't cover this (it's a page-byte threshold; chunk-level quarantine was explicitly deferred).

All three paths now route through a shared `embedPageTexts` helper:
- **Batch first** — the cheap, common path is unchanged (one call per page).
- **On a PERMANENT request-shaped failure** (non-429, non-`AITransientError`, non-401/403), retry **once per chunk** so one bad chunk costs one chunk. Failed chunks stay `embedding IS NULL`, so the next `--stale` pass picks them up; a partially-failed page is never signature-stamped as current provenance.
- **Cost bounding — no fan-out on transient failures.** A sustained 429 rethrows after `embedBatchWithBackoff`'s existing backoff (fanning out N single-chunk calls would hammer the same limiter N-fold); `AITransientError` (5xx/network, per the gateway's `normalizeAIError` classification) rethrows because the batch *content* isn't the problem; auth errors rethrow because nothing is chunk-specific. The per-chunk pass runs at most once per page per run, no recursion, and re-spends roughly the same tokens the failed batch would have. A fresh 429 arising *during* a fan-out still gets normal backoff (each single-chunk call goes through `embedBatchWithBackoff`).

**Sub-problem 3 — silent, exit 0.** `EmbedResult` gains additive `failures` (chunk count) and `failure_samples` (capped at 10 `slug: message` entries) fields, recorded at every catch site. `src/cli.ts`'s `embed` case **mirrors the `import` precedent fifteen lines above it** (`importResult.errors > 0 → setCliExitVerdict(1)`, whose comment describes this exact bug class) instead of discarding the result — so cron/CI/health gates no longer read a total no-op as success. Exit code is 1, matching the `import` convention. `runEmbed` also prints a one-line stderr failure summary, and the `--catch-up` remaining-stale warning now reads the same counter.

`--json` / structured consumers: both new `EmbedResult` fields are **additive** (always present, `0` / `[]` on clean runs); no existing field changed. No new progress phases; all failure output goes to stderr (progress contract preserved).

## What this PR deliberately does NOT fix

**Sub-problem 1 of #3037 — the chunker char-vs-token cap — is out of scope here.** A prior attempt (#3496) was closed for tackling it wrongly, and #3477 already landed a partial fix on the code-chunker path. This PR only fixes the blast radius and the exit-code silence.

## Discrimination proof (tests fail on unmodified master, behaviorally)

Both new test files were copied verbatim onto a clean `origin/master` checkout (`6920744d`) and run:

- `test/embed-partial-failure-3037.serial.test.ts` — **0 pass / 5 fail on master**: `upsertChunks` never ran for the poisoned page (`Expected length: 1, Received: 0` — the blast radius itself), `result.embedded` 0 instead of 2, `result.failures` `undefined`. **5/5 pass with this fix.**
- `test/embed-exit-code-3037.serial.test.ts` (real spawned CLI, tmpdir PGLite brain, mock OpenAI-compatible embedding server) — **fails on master**: stderr shows `Error embedding note: ... mock provider exploded` yet the process exits **0** (`Expected: not 0`). **Passes with this fix**: exit non-zero while the provider fails, then exit 0 on the same brain once the provider is healthy (failed chunks converge via `--stale`).

## Verification

- `bun run verify` — 32/32 green (includes the jsonb pattern guards, progress-to-stdout guard, key-files current-state guard).
- All existing embed/consumer suites green: `embed.serial`, `embed-stale.serial`, `embed-skip`, `embed-helper-migration`, `handlers-embed-backfill`, `embed-multimodal-batching`, `embedding-signature-stale`, `core/cycle.serial`, `migrate-embeddings-*`, `sync-deleted-slug-embed`, `v0_37_gap_fill.serial`, `fix-wave-structural`, `cli-exit-verdict-pin`, `build-llms` (216 tests across the targeted set).
- `docs/architecture/KEY_FILES.md` embed entry updated to current behavior; `bun run build:llms` re-run (bundles unchanged byte-for-byte).

No changes to VERSION / package.json / CHANGELOG.md (per instructions — release bookkeeping belongs to /ship).


time-attack:
Rebased onto master via merge commit — resolved the semantic collision with #3538 (six hunks in `src/commands/embed.ts`). Composed behavior:

1. **Wrapped text everywhere, including the fan-out.** `wrapChunkTextsForStoredMode(page, chunks)` is computed once per page and those wrapped strings feed `embedPageTexts` — so the per-chunk isolation retry re-sends the WRAPPED text, never raw `chunk_text`. An isolation retry cannot strip contextual prefixes. `fenced_code` stays unwrapped (the exemption lives inside the wrapper).
2. **Restamp only on a genuinely FULL re-embed.** `restampIfDemotedToTitleTier` now sits behind `failed === 0` in all three paths (composed with #3538's `stale.length === existing.length` gate on the stale path, and inside the existing `toEmbed.length === chunks.length` gate on the per-slug path). If isolation left chunks NULL, the page was not fully re-embedded — no signature stamp, no CR restamp, so `contextual_retrieval_mode` cannot start lying again (the #3461 bug).

New composition pin in `test/embed-partial-failure-3037.serial.test.ts`: a `per_chunk_synopsis` page with one bad chunk — asserts every embed call (failed batch + each single-chunk retry) starts with the `<context>` prefix, siblings embed, and neither `setPageEmbeddingSignature` nor `updatePageContextualRetrievalState` fires.

Post-merge: `bun run typecheck` clean, `bun run verify` 32/32, embed suites (mine + #3538's updated `embed.serial` / `embed-stale.serial` + `embedding-context`) 113/113, spawn exit-code test green.

time-attack:

The two halves the PR claims are real and correctly built: per-chunk failure isolation genuinely salvages a poisoned page's siblings, partial success is durable (the upsert's embedding CASE preserves an existing vector when EXCLUDED.embedded_at IS NULL, and a failed stale chunk correctly stays NULL for the next pass), signature/CR restamping is gated on failed===0 in all three touched paths, and `gbrain embed --stale` now exits 1 instead of 0. Discrimination executed both directions. But the fix landed on the CLI copy of the loop and NOT on src/core/embed-stale.ts:embedStaleForSource, the extracted "single source of truth" used by the embed-backfill minion — the unattended path for webhook syncs, decoupled `sync --all` and fresh-source onboarding — which I reproduced still going 0/3 chunks dark on one bad chunk while returning done:true / status:'success'. The PR body and the KEY_FILES.md entry it adds both assert "all three embed paths" are covered, so it ships a doc that is wrong on landing. Separately, switching embedPage/embedAll to embedBatchWithBackoff silently drops the AI SDK's 2 transient retries (measured 3 requests on master vs 1 on the PR against a 500-ing provider).

### Discrimination test

PASSES — executed both directions. (1) Blast-radius half: stubbed master behavior back into embedPageTexts (return the plain embedBatchWithBackoff result, no isolation) → test/embed-partial-failure-3037.serial.test.ts went 3 pass / 3 fail (the 3 blast-radius tests fail: upsertCalls length 0 vs 1, embedded 0 vs 2); 6/6 pass with the fix. The 2 cost-bounding tests and the total-failure test pass either way by construction (they assert absence of fan-out / the new failures field), so they are guard tests, not discriminators. (2) Exit-code half: removed the `if (embedResult && embedResult.failures > 0) setCliExitVerdict(1)` guard from src/cli.ts → test/embed-exit-code-3037.serial.test.ts fails ("Expected: not 0"); passes with the guard. Independently reproduced end-to-end: `gbrain embed --stale` against a 500-ing mock provider exits 0 on origin/master and 1 on this branch.

### Blockers

**1.** src/core/embed-stale.ts:200 — embedStaleForSource (used by src/core/minions/handlers/embed-backfill.ts:169) still batches a whole page in one embedFn call inside a log-only catch. Reproduced on this branch: 3-chunk page, embedFn throws a 400-shaped error when the batch contains 'BAD' → 0/3 chunks embedded, result.done===true, no failures field on EmbedStaleResult, handler returns status:'success'. Identical input on the fixed CLI --stale path yields 2/3 embedded + failures:1. This is the exact #3037 blast radius + silence, on the path that runs unattended and is itself the terminal recovery path (nothing self-heals it). Fix is small: export embedPageTexts (embed-stale.ts already imports embedBatchWithBackoff + restampIfDemotedToTitleTier from commands/embed.ts), route the call site through it, add a failures counter to EmbedStaleResult, reflect it in EmbedBackfillResult.

**2.** docs/architecture/KEY_FILES.md:10 — the added entry states 'all three page paths embed via embedPageTexts'; the PR body states 'All three embed paths now route through a shared embedPageTexts helper'. There are four page-batch embed paths and the fourth is untouched, so a checked-in reference doc is factually wrong at merge time (CLAUDE.md: KEY_FILES describes CURRENT behavior only).

**3.** src/commands/embed.ts:681 and :867 — embedPage and embedAll moved from embedBatch(texts, …) to embedBatchWithBackoff, which passes maxRetries:0 and only retries 429. Measured with a mock provider returning HTTP 500 for `gbrain embed note`: origin/master issues 3 embedding requests (SDK default maxRetries=2, node_modules/ai/dist/index.mjs:2680), this branch issues 1. embedPage is also sync's post-import inline embed step. Undisclosed reliability regression with no test; either restore SDK retries for non-429 or make the wrapper retry AITransientError.

### Not verified

- Live Postgres engine parity — no DATABASE_URL available, so the PGLite-only runs cannot surface a postgres-specific upsert difference. Low risk: the diff touches no engine file and both engines' _upsertChunksOnce embedding CASE were read and match.
- Full unit suite / bun run ci:local not run end to end (only the targeted embed/cycle/llms suites plus the whole serial lane). A failure in an unrelated shard would not have been caught.
- Real-provider fan-out cost for a genuine oversized-chunk 400 (OpenAI/Voyage) — the token-limit path in gateway.ts:embedSubBatch already binary-splits before throwing, so actual request count per poisoned page is gateway-split + N singles rather than the 'roughly the same tokens' the PR doc claims. Reasoned from code, not measured against a live provider.
- Whether embed-backfill jobs are actually scheduled in the reviewer's default install (submitEmbedBackfill gating) — the handler and its call path were read, but no live minion run was executed. This affects the practical severity, not the existence, of blocker 1.

- `gh pr view 3532 — exit 0`
- `gh pr diff 3532 > /tmp/pr3532.diff — exit 0`
- `git fetch origin pull/3532/head:pr3532 && git checkout pr3532 — exit 0 (branch is BEHIND origin/master c6dc0adf; merge-base check exit 1)`
- `bun install — exit 0`
- `bun run verify — exit 0 (32/32 checks green, 24s)`
- `bun test test/embed-partial-failure-3037.serial.test.ts — exit 0 (6 pass / 0 fail)`
- `bun test test/embed-exit-code-3037.serial.test.ts — exit 0 (1 pass / 0 fail)`
- `GBRAIN_DISCRIM_STUB=1 bun test test/embed-partial-failure-3037.serial.test.ts (master behavior stubbed) — exit 1 (3 pass / 3 fail)`
- `bun test test/embed-exit-code-3037.serial.test.ts with cli.ts guard removed — exit 1 (Expected: not 0)`
- `bun test embed.serial embed-stale.serial embed-skip cli-exit-verdict-pin embedding-migration embedding-signature-stale — exit 0 (96 pass / 0 fail)`
- `bun test embed-multimodal-batching embed-helper-migration embed-input-type-wire.serial embedding-context embed-backfill-submit handlers-embed-backfill test/cycle/ admin-embed-spawn.serial — exit 0 (155 pass / 0 fail)`
- `bun test test/sync-deleted-slug-embed.test.ts — exit 0 (1 pass)`
- `bun test test/build-llms.test.ts — exit 0 (12 pass)`
- `bash scripts/run-serial-tests.sh — exit 1 (3 files failed: facts-mcp-allowlist, hybrid-meta, hybrid-search-lite)`
- `same 3 serial files on origin/master worktree — exit 1 (5 fail) => PRE-EXISTING, local env has API keys so 'missing key' assertions fail`
- `scratch test test/zz-reviewer-sibling-3037.serial.test.ts against embedStaleForSource — exit 1: embedded=0 chunksProcessed=3 done=true lit_chunks=0/3 (sibling path unfixed; file deleted after)`
- `bun /tmp/retrycount.mjs (spawned CLI vs 500-ing mock, PR branch) — exit 0: EMBED_REQUESTS_FOR_ONE_PAGE=1`
- `bun /tmp/retrycount.mjs (origin/master worktree) — exit 0: EMBED_REQUESTS_FOR_ONE_PAGE=3`
- `bun /tmp/rc2.mjs both branches — master: SLUG requests=3 exit=1, STALE requests=1 exit=0; PR: SLUG requests=1 exit=1, STALE requests=1 exit=1`

</details>

---

time-attack:

Issue #3242, opened 2026-07-22
Pages ingested into a non-default federated source are invisible to get_page/search/list_pages (CLI and MCP) — only resolve_slugs sees them


jeffpoulton:
**Version:** 0.42.59.0, Postgres engine

## Setup

- Second source registered alongside an established `default`:
  `sources add --id knowledge --path /home/.../knowledge/knowledge --federated true`
  (path points at a **subdirectory** of a git repo, so `clone_state: "no-git"` — the `.git` is one level up; sync of the files themselves works fine)
- Ingested via CLI: `gbrain sync --source knowledge`

## Observed

Ingestion succeeds, retrieval cannot see any of it:

| Surface | Result |
|---|---|
| ingest log | `Imported 7 pages, 0 skipped, 40 chunks`, correct slugs (`knowledge/glossary`, …) |
| `sources status knowledge` | `page_count: 7`, correct `last_commit` |
| `resolve_slugs("knowledge/")` (MCP) | **returns the new slugs** — rows exist |
| `get_page("knowledge/glossary")` (MCP) | `page_not_found` |
| `gbrain get knowledge/glossary` (CLI, on the brain host) | nothing |
| `search` (MCP, query the glossary should dominate) | only `default`-source hits |
| `list_pages` (MCP) | never lists them |

So pages ingested into a non-default source are visible to slug resolution but invisible to `get_page`, `search`, and `list_pages` — on **both** the CLI and MCP paths — despite `federated: true`. Caller is a legacy token with `read`/`write`/`admin` (no source-scoped OAuth grant involved).

## Attribution inconsistency (possibly the same root)

The ingest-log row for this sync reports `source_id: "default"` with `source_ref` pointing at the knowledge directory, while `sources_status` credits the same 7 pages to `knowledge`. Retrieval filtering hard-wired to `default` would explain both the misattribution and the invisibility.

## Related

Looks like the inverse of #2928 (`federated: false` sources still surface in cross-source retrieval) — together they suggest the `federated` flag isn't consulted by retrieval at all, just defaulted differently per path. Possibly same neighborhood as #2929/#2930 (source auto-routing) and #2922 (hardcoded `default` writes).

## Expected

A `federated: true` source's pages appear in unqualified `get_page`/`search`/`list_pages`/`query` results, and the ingest log attributes a `--source` sync to that source.

PR #3533, opened 2026-07-28
fix(sources): honor unfederate for unqualified reads (#2928)

### LINKED ISSUE #2928: `sources unfederate` / `config.federated=false` has no effect on unqualified `query`/`search`/`think` retrieval — an "isolated" source's pages still surface in cross-source results

time-attack:
Fixes #2928.

## Root cause — confirmed, but NOT where the issue points

The issue names `sourceScopeOpts` / `resolveRequestedScope` and the dead `federatedOnly` loader flag. Independent trace disagrees, and I verified that disagreement: those functions never see `config.federated` because the flag is consumed upstream. The leak is the interaction of two upstream pieces:

1. `pickSoleNonDefaultSource` (tier 5.5, the #1434 sole-source convenience) anchors an unqualified call on the single non-default source — regardless of `config.federated`.
2. `localFederatedSourceIds` (`src/core/source-resolver.ts`) then widens that anchor into `[resolvedSource, ...federated others]` — the resolved source is always included, and the seeded `default` is federated. So in the issue's exact repro (brain = `default` + one explicitly-unfederated source, unqualified `gbrain query` from an unrelated cwd), the read scope becomes `["isolated-src", "default"]`: the isolated source's pages mixed into a cross-source result set — precisely the reporter's observation.

Migration v62's `oauth_clients.federated_read` backfill is not implicated — that governs the remote OAuth channel (`ctx.auth.allowedSources`); this leak is on the trusted local unqualified path.

## Fix — READ path only

`--no-federated` / `sources unfederate` means "don't mix me into cross-source READS", not "don't route my writes here". So the fix lives entirely in `localFederatedSourceIds`, which feeds `federatedSearchScope` exclusively (search/query/get_page/list_pages/resolve_slugs — reads): when the **resolved** source is explicitly `config.federated === false`, the widening set is suppressed and the scalar scope stands. The isolated source's unqualified reads stay in-source; its pages are never mixed with `default`'s and vice versa. Same seam serves CLI, stdio MCP, and HTTP MCP transports.

Deliberately NOT changed:

- **Tier 5.5 write routing is untouched.** Unqualified `sync`/`import` on a single-vault brain still lands in the vault even when it is `--no-federated` (#1434; `test/sync-sole-non-default-routing.test.ts` passes unmodified). An earlier revision of this branch excluded isolated sources from tier 5.5 itself; CI correctly rejected that — it silently redirected writes to `default`. Reverted in the second commit.
- **UNSET `federated` keeps pre-existing widening.** Only an explicit opt-out changes behavior; a fresh `sources add` (no `federated` key) behaves exactly as before.
- **Empty local scope stays UNSCOPED.** `sourceScopeOpts`/`federatedSearchScope` return `{}` for a local caller with no source in scope — never collapsed to `'default'` (the over-narrowing class that closed #3470/#3497; pinned by dedicated tests).
- **Explicit tiers still reach an isolated source** (`--source`, env, dotfile) — naming it IS the qualified case, per the flag's own contract.

Semantics note for the maintainer: with a sole isolated vault, an unqualified read is scoped to that vault alone (the read analog of #1434) rather than the vault being invisible. Isolation = "never mixed into cross-source results", not "invisible to the brain's own unqualified reads". If the intended contract is full invisibility, reads and writes would have to resolve different anchors from the same unqualified context (a `gbrain import` immediately invisible to `gbrain query`) — flagging that as a design decision rather than landing it silently.

## Discrimination proof (behavioral, on unmodified master)

`test/unfederate-read-scope-2928.test.ts` imports only master-existing exports. With `src/core/source-resolver.ts` checked out from `origin/master` (`git checkout origin/master -- src/core/source-resolver.ts`), 3 tests fail behaviorally and the rest pass:

```
✗ explicitly isolated resolved source gets NO federated widening set
    expect(localFed).toBeUndefined()   Received: ["isolated-src", "default"]
✗ full unqualified read chain stays scalar: no default pages mixed in
    expected { sourceId: "isolated-src" }   Received: { sourceIds: ["isolated-src", "default"] }
✗ isolated brain_default anchor is not widened either (same seam)
    Received: ["isolated-src", "default"]
```

With the fix: 11/11 pass, and `test/sync-sole-non-default-routing.test.ts` passes **unmodified** alongside.

Coverage highlights:
- isolated anchor is never widened (sole_non_default AND brain_default tiers)
- full read chain (`resolveSourceWithTier` → `localFederatedSourceIds` → `federatedSearchScope`) yields scalar scope
- tier 5.5 still anchors the isolated sole vault (write-routing pin)
- UNSET-federated sole source keeps auto-routing AND keeps widening (#1434 + #3242 guards)
- explicit `--source` still reaches an isolated source
- multi-source widening set already excludes isolated non-anchors (pin)
- **unscoped local path**: `sourceScopeOpts`/`federatedSearchScope` with no sourceId and no grant return `{}` — never `'default'` (#3470/#3497 regression class)
- empty `allowedSources: []` does not widen; federated grant > scalar precedence pin

## Verification

- `bun test test/unfederate-read-scope-2928.test.ts test/sync-sole-non-default-routing.test.ts test/source-resolver*.test.ts test/source-scope-resolver.test.ts` → 88 pass / 0 fail
- `bun test test/local-federated-search-scope.test.ts` (pre-existing #3242 widening suite) → 19 pass / 0 fail
- `bun run typecheck` → clean
- `bun run verify` → 32 pass / 0 fail

### DIFF
--- a/src/core/source-resolver.ts
+++ b/src/core/source-resolver.ts
@@ -200,32 +200,27 @@ export function resolveSourceIdEngineFree(
  * doesn't auto-resolve. Shared by `resolveSourceId` and `resolveSourceWithTier`
  * so the heuristic can't drift between the two entry points.
  *
- * #2928: also refuses to auto-pick a source EXPLICITLY marked isolated
- * (`gbrain sources unfederate <id>` / `--no-federated` → config.federated =
- * false). Auto-routing there made an unqualified `query`/`search`/`think`
- * surface the isolated source's pages — the exact thing unfederate promises
- * to prevent ("only searched when explicitly named"). UNSET federated stays
- * eligible: a fresh `sources add` writes no federated key, and the #1434
- * sole-source convenience must keep working for it. Narrowing-only by
- * design — an isolated sole source falls through to 'default'; the tier
- * never widens to some OTHER source it wouldn't have picked before.
+ * NOTE (#2928): this tier deliberately does NOT consult config.federated —
+ * `--no-federated` governs READ mixing, not write routing, and unqualified
+ * `sync`/`import` on a single-vault brain must keep landing in the vault
+ * (#1434, pinned by test/sync-sole-non-default-routing.test.ts). The
+ * unfederate read fix lives in `localFederatedSourceIds` below.
  */
 async function pickSoleNonDefaultSource(engine: BrainEngine): Promise<string | null> {
   // archived column was added in v34 (v0.26.5). Older brains may not have
   // it — fall back to the un-archived query in that case via try/catch.
-  let rows: Array<{ id: string; config: unknown }>;
+  let rows: Array<{ id: string }>;
   try {
-    rows = await engine.executeRaw<{ id: string; config: unknown }>(
-      `SELECT id, config FROM sources WHERE local_path IS NOT NULL AND id != 'default' AND archived = false`,
+    rows = await engine.executeRaw<{ id: string }>(
+      `SELECT id FROM sources WHERE local_path IS NOT NULL AND id != 'default' AND archived = false`,
     );
   } catch {
-    rows = await engine.executeRaw<{ id: string; config: unknown }>(
-      `SELECT id, config FROM sources WHERE local_path IS NOT NULL AND id != 'default'`,
+    rows = await engine.executeRaw<{ id: string }>(
+      `SELECT id FROM sources WHERE local_path IS NOT NULL AND id != 'default'`,
     );
   }
-  if (rows.length !== 1) return null;
-  if (parseSourceConfig(rows[0].config).federated === false) return null;
-  return rows[0].id;
+  if (rows.length === 1) return rows[0].id;
+  return null;
 }
 
 /**
@@ -404,7 +399,10 @@ export async function resolveSourceWithTier(
  *
  *   - explicit tiers (`flag` / `env` / `dotfile`): the user named a source;
  *     scalar scope stands (that IS the qualified case);
- *   - no other federated source exists: keep the scalar fast path unchanged.
+ *   - no other federated source exists: keep the scalar fast path unchanged;
+ *   - #2928: the resolved source is explicitly isolated (config.federated =
+ *     false): it must not be mixed into a cross-source read in EITHER
+ *     direction, so the scalar scope stands.
  *
  * Archived sources are excluded (same rationale as pickSoleNonDefaultSource);
  * the archived column is v34+, so fall back to the un-archived query on older
@@ -426,6 +424,16 @@ export async function localFederatedSourceIds(
     rows = await engine.executeRaw<{ id: string; config: unknown }>(
       `SELECT id, config FROM sources ORDER BY id`,
     );
+  }
+  // #2928: an EXPLICITLY isolated anchor (`sources unfederate` /
+  // `--no-federated` → config.federated = false) opted out of cross-source
+  // read mixing — never widen it into the federated set (which would drag
+  // other sources' pages into its unqualified reads and vice versa). Scalar
+  // scope stands. UNSET federated keeps the pre-#2928 widening behavior;
+  // write routing (tier 5.5 above) is deliberately untouched.
+  const resolvedRow = rows.find((row) => row.id === sourceId);
+  if (resolvedRow && parseSourceConfig(resolvedRow.config).federated === false) {
+    return undefined;
   }
   const ids = [
     sourceId,
--- a/src/core/source-resolver.ts
+++ b/src/core/source-resolver.ts
@@ -16,7 +16,7 @@
 import { readFileSync, lstatSync, type Stats } from 'fs';
 import { join, dirname, resolve } from 'path';
 import type { BrainEngine } from './engine.ts';
-import { isSourceFederated } from './sources-load.ts';
+import { isSourceFederated, parseSourceConfig } from './sources-load.ts';
 import { SOURCE_ID_RE, isValidSourceId } from './source-id.ts';
 import { isTrustedDotfile, realpathOrResolve } from './path-confine.ts';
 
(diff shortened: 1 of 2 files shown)


time-attack:

The fix is real, root-cause, and read-path-only as claimed: tier 5.5 anchors the sole non-default source and localFederatedSourceIds widened it into [isolated-src, default] (seeded default is federated:true — verified in pglite-engine.ts:706/migrate.ts:835); the early return at source-resolver.ts:434-437 suppresses widening only for an explicit federated:false anchor. Discrimination test passed: stubbing the hunk back makes exactly 3/11 tests fail behaviorally. Write routing verified untouched (pickSoleNonDefaultSource unchanged; sync-sole-non-default-routing.test.ts green unmodified); the output feeds only federatedSearchScope (reads, remote===false only) and all three transports (cli.ts:894, mcp/server.ts:46, http-transport.ts:400) route through the one fixed function. Empty scope stays {} (pinned). Query-cache is safe without a knobs bump: cacheScopeKey keys federated sets as __set__:sorted-ids, distinct from scalar. BUT the PR does not merge: CONFLICTING/DIRTY against master (#3524 __all__ sentinel touched the same import block in source-resolver.ts). I resolved the 2-line import-union conflict locally and the merged tree is fully green (typecheck clean, 108/108 across both PRs' suites incl. all-sources-sentinel.test.ts), so composition is proven — author needs to rebase, bump the VERSION trio (branch at 0.42.67.0 vs master 0.42.68.1, no CHANGELOG entry — CI version-gate fails), and prefix the PR title with vX.Y.Z.W per the IRON RULE. After those three mechanical fixes this is a MERGE.

### Discrimination test

Removed the PR's early-return hunk (source-resolver.ts:434-437) from the checked-out PR branch and re-ran test/unfederate-read-scope-2928.test.ts: 3 tests failed behaviorally (isolated anchor widened to ['isolated-src','default'], full read chain returned sourceIds instead of scalar, brain_default anchor widened) — exit 1. Restored the fix: 11/11 pass, exit 0. Matches the PR body's claim exactly; the tests discriminate.

**Mergeability at review time:** {"mergeStateStatus":"DIRTY","mergeable":"CONFLICTING"}

### Blockers

**1.** mergeable: CONFLICTING / mergeStateStatus: DIRTY — import-line conflict in src/core/source-resolver.ts:19-21 vs master's #3524 (__all__ sentinel). Trivial union resolution (keep parseSourceConfig import + ALL_SOURCES import); verified green post-merge locally, but the PR as pushed does not merge.

**2.** VERSION gate: branch VERSION/package.json/CHANGELOG all at 0.42.67.0 (== its base), master is 0.42.68.1 — not strictly greater, and no CHANGELOG entry for the fix. CI version-gate will reject.

**3.** PR title 'fix(sources): honor unfederate for unqualified reads (#2928)' lacks the leading vMAJOR.MINOR.PATCH.MICRO (IRON RULE).

### Not verified

- bun run verify / full unit suite / ci:local — forbidden by execution-footprint rules; PR's claimed '32 pass verify' and '88 pass' aggregate not reproduced (my subset runs: 11+40+56 all green)
- Real-Postgres e2e parity — PGLite only; the diff touches no engine SQL writes so risk is low
- End-to-end compiled-binary CLI run (gbrain query on a live brain) — tested at the resolver/scope-helper seam, which grep confirms is the exact wiring in cli.ts:888-915

- `gh pr view 3533 --json mergeable,mergeStateStatus (twice; UNKNOWN then CONFLICTING/DIRTY)`
- `git fetch origin master pull/3533/head:pr-3533 && git checkout pr-3533`
- `git diff origin/master...pr-3533 (--stat and full; 2 files, +195/-2)`
- `git merge-tree --write-tree origin/master pr-3533 (conflict in src/core/source-resolver.ts)`
- `git show b3b43d0f --stat (the conflicting master commit, #3524)`
- `bun install`
- `bun test test/unfederate-read-scope-2928.test.ts (11 pass)`
- `stub: removed the early-return hunk, re-ran same file (3 fail, exit 1), restored via git checkout`
- `bun test test/sync-sole-non-default-routing.test.ts test/local-federated-search-scope.test.ts test/source-scope-resolver.test.ts (40 pass)`
- `bun test test/source-resolver*.test.ts (56 pass)`
- `bun run typecheck (clean, PR branch)`
- `git merge origin/master → resolved 2-line import conflict → bun run typecheck clean + 108/108 across 8 suites incl. test/all-sources-sentinel.test.ts → git merge --abort`
- `grep callers of localFederatedSourceIds (cli.ts, mcp/server.ts, mcp/http-transport.ts — all route through the fixed function)`
- `grep cacheScopeKey in src/core/search/hybrid.ts + query-cache.ts (federated scope keyed __set__:sorted, no cache contamination)`
- `grep unfederate in src/commands/sources.ts (command exists, line 1346)`
- `cat VERSION / package.json version / CHANGELOG top entry on branch (0.42.67.0) vs git show origin/master:VERSION (0.42.68.1)`

</details>

---

time-attack:

PR #3536, opened 2026-07-28
v0.42.67.1 fix(minions): settle the spawn loop when a child never launches

daragao3:
## The bug

`ChildWorkerSupervisor.spawnOnce()` resolves its promise only from inside `child.on('exit')`.

A spawn that fails to **launch** — ENOENT/EACCES on `cliPath`, or a target the OS refuses to execute — emits `'error'` and `'close'` but **never** `'exit'`. So the promise never settles and `run()` awaits it forever.

The comment above that handler on master says:

> Node fires 'error' first, then 'exit' with code=null. We log the error; the 'exit' handler increments crashCount as usual so the restart loop bounds permanent misconfigs via max_crashes.

That has never been true for a launch failure, which is why the documented crash-bounding never actually worked. On master the only `resolve()` paths are the `isStopping` early-return, the *synchronous* spawn throw, and `'exit'`. There is no `'close'` handler at all.

## Why this is not platform-specific

`buildSpawnInvocation()` returns `{ cmd: cliPath }` directly whenever tini is absent, and `detectTini()` resolves tini via `which tini` — not present on most hosts. So a bad `cliPath` *is* the spawn command, and the failure is a plain ENOENT on every platform, Linux included.

Blast radius:

- `supervisor.ts` does `await this.childSupervisor.run()` in `runSuperviseLoop()`. A bad `cliPath` wedges the `gbrain jobs supervisor` daemon permanently — no respawn, no crash count, no give-up, one log line and then silence. The process stays alive, so it looks healthy while processing nothing.
- `autopilot.ts` uses `void childSupervisor.run()`, so autopilot itself does not block — but its worker silently never respawns while the loop keeps running. Quieter, same root cause.

It was invisible in CI because nothing in the suite pointed `cliPath` at an unlaunchable target.

## The fix

`settleSpawnFailure()` settles from `'error'` when `child.pid === undefined` (no pid means the OS never created a process, so no `'exit'` is coming), plus a `'close'` fallback for platform-dependent EACCES shapes that do get a pid.

- A settle-once guard keeps the exit path and the failure path from double-counting.
- The `'close'` fallback is gated on a `spawnErrored` flag, because `'close'`/`'exit'` ordering is not guaranteed and an ungated handler could pre-empt a healthy run's classified exit.
- The failure counts as a crash, so it pays the existing exponential backoff and is bounded by `hardStopMaxCrashes` like any other permanent misconfig.
- It emits `worker_exited` with `code: null`, `likelyCause: 'spawn_failed'`. That cause is deliberately **not** added to `CLEAN_EXIT_CAUSES` in `supervisor-audit.ts`, so the existing denylist semantics already count it as a crash. It lands in the `legacy` bucket, which that file's own docblock describes as the intended catch-all signal for a new upstream `likely_cause` — so no change is needed on the audit side.

## Tests

Two new regression tests point `cliPath` at a nonexistent path. That is ENOENT everywhere, so this class is guarded on CI rather than only where it was found.

The file's internal safety net changes from an event-count to a wall clock. A counter that only advances when the supervisor emits is structurally blind to a failure mode whose symptom is *"stops emitting"* — which is why this bug hung the whole `bun test` process instead of failing one test. Worker fixtures move off `#!/bin/sh` to a platform shell one-liner, or a `.mjs` run via `process.execPath`.

Verification against this branch:

- `bun test test/child-worker-supervisor.test.ts` → 15 pass / 1 fail, twice. **A different test failed each run** (`restartCurrentChild …`, then `code=0 after stable 5min+ run …`), both pre-existing wall-clock-sensitive tests, on a contended Windows dev box. Both new spawn-failure tests passed in both runs. These two are worth a look on CI hardware.
- Negative control: with the fix reverted, both new tests fail with a diagnostic and the process **terminates** instead of hanging.
- `bun run typecheck` → exit 0.

## Notes for the maintainer

- Versioned `0.42.67.1` (VERSION + package.json + CHANGELOG) to satisfy the version-first convention. Renumber freely if this collides with an in-flight wave.
- The full unit suite was not run to completion on this Windows box; the targeted file, its neighbours, and typecheck were.


time-attack:
Heads-up on a queue collision: **#3511 and #3536 both self-allocate `0.42.67.1`.** Both are MERGE verdicts on review, but the CI version-gate requires `VERSION` strictly greater than master's, so whichever lands second will fail and need a re-bump.

Merge-order note for the wider queue: the self-allocated ladder is #3511=.67.1, #3536=.67.1 (collision), #3518=.67.3, #3527=.70.0, #3543=.71.0. Landing #3527 (.70.0) first would strand every .67.x PR, so the .67.x set needs to go first, renumbered sequentially.

No action needed from you unless you land second — I'll handle the renumber if it comes to that.

time-attack:

Issue #1575, opened 2026-05-27
`unify-types` CLI applies by default — handler interface says "Default false (dry-run)" but worker registration flips it to true


matthewod11-stack:
## Summary

Two facts about `unify-types` disagree:

1. **Handler interface** in [`src/core/schema-pack/unify-types-handler.ts:41-50`](src/core/schema-pack/unify-types-handler.ts#L41-L50) documents `apply?: boolean` as:
   > `/** Apply mutations. Default false (dry-run). */`

2. **Worker registration** in [`src/commands/jobs.ts:1644`](src/commands/jobs.ts#L1644) explicitly does:
   ```ts
   apply: data.apply ?? true,            // worker invocation defaults to apply
   ```

The worker default wins for the canonical operator invocation, i.e., **the CLI invocation applies destructively by default**:

```
gbrain jobs submit unify-types --allow-protected --params '{"target_pack":"X"}'
```

There's also no `--dry-run` (or `--apply`) ergonomic flag on `gbrain jobs submit`. The only way to opt into dry-run is to pass it inside `--params`:

```
gbrain jobs submit unify-types --allow-protected --params '{"target_pack":"X","apply":false}'
```

…which isn't discoverable from `gbrain jobs submit --help`.

## Impact

First-time operators reading the handler interface (or any internal issue template that quotes its doc-comment) reasonably expect dry-run on first invocation — that's what the function-level documentation says is the default. The actual behavior is a one-shot destructive migration.

In a recent morty-v3 pack-activation session (2026-05-27, LABS-102), a planned dry-run checkpoint between preview and apply was lost because the CLI applied on first invocation. The outcome matched the operator's confirmed scope, but the operator-checkpoint was gone. The destructive default was clearly intentional (the inline comment "worker invocation defaults to apply" is explicit), but the disagreement between handler-level documentation and worker behavior is friction.

### A. CLI ergonomic only (smallest patch)

Add a `--dry-run` shortcut to `gbrain jobs submit` for `unify-types` (or all PROTECTED handlers) that injects `apply: false` into params. Default behavior unchanged; the safer mode becomes discoverable via `--help`.

### B. Realign default to match handler doc (safest)

Flip [`jobs.ts:1644`](src/commands/jobs.ts#L1644) to `apply: data.apply ?? false`, and add a `--apply` shortcut for the operator-friendly destructive mode. Aligns with the handler interface doc-comment and biases the system toward "show me what you'd do, I'll confirm."

[...]

PR #3543, opened 2026-07-28
v0.42.71.0 feat(sync): make dry-run previews immutable and exact-target applies fail closed

mamedov:
## Summary

This makes `gbrain sync --dry-run` a genuinely zero-write preview backed by one immutable `SyncPlan`, then lets automation apply that exact target with explicit commit, bookmark, clean-worktree, and full-plan commitments.

- Return one schema-1 JSON document on stdout for success and refusal paths.
- Bind `plan_digest` to the target bytes, operation identities (including rename sources), corpus projection, strategy state, and active schema-pack identity.
- Require every paired real apply to supply the reviewed digest, rebuild the plan under the same source lock, and refuse changed Git, database, source, or pack state before sync-domain mutation.
- Bind exact-target file bytes, timestamps, delete slugs, checkpoint provenance, and final bookmarks to the captured commit.
- Reuse the reviewed active-pack/no-pack snapshot through full import and fail closed when a planned rename no longer owns exactly one live source row.
- Coordinate sync, archive/restore, purge, and managed re-clone operations through the canonical per-source lifecycle lock.
- Prove lock ownership with `(id, pid, host)`, surface lost refresh/release ownership, recheck archived state under purge lock, protect `default`, and atomically replace attach markers.

Basecamp: https://3.basecamp.com/4058523/buckets/34718435/card_tables/cards/10135180500

## Compatibility decisions

- Ordinary unpinned sync remains available for interactive use; strict exact-target behavior is opt-in through the new preconditions.
- Paired applies require `--no-pull`, the source-scoped refreshing lock, and the preview's lowercase 64-character `plan_digest`; they use a private materialization of the captured Git commit and defer embedding/extraction from the deterministic receipt.
- Ordinary failure auto-skip behavior remains compatible, while paired applies refuse `--skip-failed` and never auto-skip an incomplete target.
- Ordinary incremental rename fallback remains compatible; paired rename collision or vanished-source cases block without add/upsert fallback or bookmark advance.
- `--all --json` retains its aggregate envelope; cost-gate evidence is nested in that one document and human diagnostics stay on stderr.
- Gitignored content is excluded from exact-target mode because it cannot be proven from the target commit.
- No database schema migration is required.

## Verification

- Focused final matrix: 320 passed, 0 failed, 1,134 assertions across 14 sync-plan, exact-target, resumable-import, full-sync, lifecycle, and lock files.
- Final explicit no-pack snapshot regression: 1 passed, 0 failed, 10 assertions.
- `bun run typecheck` passed after the final fix.
- Documentation-history, trailing-newline, progress/stdout, privacy, proposal-PII, test-name, source-config-leak, and `git diff --check` checks passed.
- Generated `llms-full.txt` is current.
- Independent final adversarial audit found no remaining P0/P1 issue in the reviewed commitment/rename scope.
- Verified commit: `6436f3a6f2b236d420cb3d158c93ce2da6454b2e`.
- The full suite was intentionally not run in this review; upstream PR CI remains the broad-run authority.

## Release status

No release, package publication, installation, merge, or deployment was performed as part of this work.


mamedov:
@codex review

mamedov:
@codex review

mamedov:
The latest-head GitHub Actions runs are currently marked `action_required` because this PR comes from a fork. A maintainer will need to approve the workflows before CI can report. Local focused verification on `6436f3a6f2b236d420cb3d158c93ce2da6454b2e` is green (320 tests, 0 failures, 1,134 assertions; final no-pack regression 1/0/10; typecheck and repository guards passed). No release or deployment was performed.

time-attack:
**You found a real and serious bug. Please split it out — I can't take it in this shape.**

## The bug is real, verified on master, and worse than the title suggests

Two distinct dry-run violations in `src/commands/sync.ts`:

1. **`--dry-run` deletes pages.** The un-syncable-page loop calls `await engine.deletePage(slug, pageOpts)` (~L2397) *before* the `if (opts.dryRun)` early-return (~L2407). A tracked page whose file becomes unsyncable under the selected strategy is **destroyed by a preview**.
2. **`--dry-run` mutates the working tree.** `pullRepo(gitContextRoot)` (~L2017) is gated only on `!opts.noPull && !detachedHead && originRemotePresent`. The deliberate design (see the comment at ~L295) covers the *fetch* for cost estimates — a pull is well beyond that.

This is the third instance of this class in two days: #2712 (`jobs prune --dry-run` deleted rows) and #1575 (`unify-types` applied by default, contradicting its own docs) both landed yesterday. Yours is the most severe of the three, because `sync` is the command people run most and `--dry-run` is exactly what a cautious user reaches for first.

## Why I can't merge it as-is

The diff is **10,947 lines across 30 files**, +3,668 in `sync.ts` alone — currently the most-churned file in the repo. `mergeable` is already `CONFLICTING` against master.

The ~20-line dry-run fix is buried inside an entire new apply protocol: plan digests, `--expected-target` / `--expected-bookmark` / `--expected-plan-digest` / `--require-clean`, private commit materialization, `(id,pid,host)` lock ownership, and lifecycle-lock coordination across archive/restore/purge/re-clone, plus engine interface changes.

I can't verify that surface to merge confidence against the engine-parity, JSONB, and source-isolation invariants — and it collides with sync work that landed yesterday (#3479 rename reconcile, #3522 slug regexes) and with the checkpoint/lock tuning subsystem.

## What I'd merge today

A standalone PR that does only:

- move the un-syncable-delete loop **below** the dry-run early-return
- gate `pullRepo` on `!opts.dryRun` — or, if pull-on-dry-run is genuinely intended for bookmark accuracy, document it at the call site and leave it
- a regression test proving a `--dry-run` against a brain with an un-syncable tracked page leaves the page present

That lands immediately. Given the severity I'd rather have those 20 lines this hour than the full protocol next week.

The apply protocol deserves its own proposal and its own review — it's a substantial design change, and bundling it behind a data-loss fix means the fix waits on the design discussion. Same reasoning applied to #3470 when it bundled a genuine cross-source read leak with an unrelated scope-ceiling change; the leak fix is landing separately as #3550.

Thanks for finding this — the delete-before-dry-run-return ordering is subtle and I'd not have caught it from the outside.

mamedov:
Thanks for the clear split. I opened #3624 with only the two requested dry-run fixes and focused regressions: `--dry-run` no longer runs `pullRepo`, and tracked pages that become unsyncable are left intact.

The apply protocol is intentionally excluded. I’m keeping this PR open as the separate design/proposal thread.

time-attack:
Closing this in favour of #3624, with thanks — and I want to be precise that this is about reviewability, not about the work being wrong.

**Why:** this PR changes 8,930 lines across 30 files in the sync path. That is the same code where #3583 was just found, by execution, to hard-delete every active row whose `source_path` matches a renamed file rather than the one row intended — and where a dry-run sweep sits above the dry-run guard. Two confirmed destructive defects have now come out of this area in a week.

I hold every PR to the same bar right now: reproduce the bug, confirm the fix, and confirm the PR's own test fails without it. Across 32 PRs reviewed that way, 1 was mergeable. I cannot honestly apply that bar to 9,000 lines of page-deletion logic, and I do not think a review that claims to have done so would be truthful.

**#3624 is the narrow split that was asked for**, and it reviewed clean: 144 lines, `--dry-run` skips `pullRepo` and returns before out-of-strategy cleanup, with tests proving a strategy-changing preview preserves both the indexed page and the sync bookmark. You deliberately left the apply protocol out, which was the right call. That one is going in.

**What would help most:** the apply protocol as its own small PR. Specifically the exact-target fail-closed behaviour — that is the valuable half and it deserves its own review rather than being carried along with 8,800 other lines. Same for anything else here you consider load-bearing; one concern per PR and I can actually verify each one.

One thing from this PR that should not get lost regardless: the dry-run problem is now at **six instances across four commands**, tracked as class issue #3594. The durable fix is almost certainly a dry-run engine wrapper whose write methods throw, so a destructive call above a guard fails loudly instead of deleting silently. Guard placement is invisible at review time, which is why this keeps happening to careful people. If you wanted to take that on, it would prevent the entire class rather than one more instance of it.

Not closed because the work is unwelcome. Closed because I would be pretending to have reviewed it.

PR #3544, opened 2026-07-28
fix(search): fold detail into the query-cache key (#3515)

### LINKED ISSUE #3515: query_cache does not key on detail: a detail=low result set can be served to a detail=medium lookup
Found while fixing #3430; independent of it and not fixed there.

`knobsHash` (`src/core/search/mode.ts`) is the **only** thing that varies the query-cache key — the lookup filter is `WHERE qc.source_id = $2 AND qc.knobs_hash = $4 AND qc.embedding <=> $1::vector < $3` (`src/core/search/query-cache.ts:158-170`). `detail` is not in the hash's fixed-order key list.

But `detail` is result-affecting by design: it gates dedup, chunk-source filtering, and (until #3430) the compiled_truth boost. So two queries with near-identical embeddings and different `detail` values fall on the same cache row, and whichever ran first decides what the other gets, for the whole TTL (3600s default).

Concretely: `query "X" --detail low` populates the cache with a compiled-truth-only result set; a subsequent `query "X"` at default `medium` inside the TTL gets that narrow set back instead of the full one. Silent, and it looks like a relevance problem rather than a cache problem.

This is the same contamination class as the `[CDX-4]` cross-mode fix that introduced `knobs_hash`, and the same class as the v=2→3 (`floor_ratio`, `col`/`prov`) and v=9→10 (`relationalRetrieval`) additions — a knob that changes results has to be in the key.

**Minimal fix:** append `det=${detail ?? 'medium'}` to the `knobsHash` parts list and bump `KNOBS_HASH_VERSION` (append-only, per the comment in that function). Note `detail` is a per-call `SearchOpts` value rather than a resolved mode knob, so it arrives via the `KnobsHashContext` path the way `col`/`prov` already do, not through `ResolvedSearchKnobs`.

One-time cache-miss spike on upgrade, same as every prior version bump.

time-attack:
Fixes #3515.

## The bug

`knobsHash` (`src/core/search/mode.ts`) is the only thing that varies the query-cache key, and `detail` was not in it — yet `detail` gates dedup, chunk-source filtering, and the compiled_truth boost. So `query "X" --detail low` populated the cache with a compiled-truth-only result set, and a later `query "X"` at default `medium` inside the TTL (3600s) got that narrow set back. Silent, and it looks like a relevance problem rather than a cache problem. Same contamination class as `[CDX-4]`, the v=2→3 (`floor_ratio`, `col`/`prov`) and v=9→10 (`relationalRetrieval`) additions.

## The fix

- Append `det=${ctx?.detail ?? 'medium'}` to the `knobsHash` parts list (append-only, per the list's own convention) and bump `KNOBS_HASH_VERSION`.
- `detail` is a per-call `SearchOpts` value, not a mode knob, so it rides `KnobsHashContext` the way `col=`/`prov=` do — `ResolvedSearchKnobs` stays pure.
- `hybridSearchCached` threads the **effective** level: `opts.detail ?? autoDetectDetail(query)`, matching what bare `hybridSearch` resolves. This closes the residual leak where an auto-detected `high` query would otherwise share a `det=medium` row with a genuinely-medium query of similar embedding.
- Undefined falls back to `'medium'` (the documented default), so legacy callers that don't thread detail share the default-detail rows.

One-time global cache cold-miss on upgrade; refills within `cache.ttl_seconds` — same as every prior bump.

## Coordination with #3514 (open, bumps 13→14)

PR #3514 (fix for #3430) already claims v=14 with no new hash parts. Per the repo's established sequencing convention (the v=4/v=5 D8 note in `mode.ts`), **this PR is based on origin/master and bumps 13→15, leaving v=14 to #3514.** All five `KNOBS_HASH_VERSION` pin sites are updated here (`test/search-mode.test.ts` ×2, `test/search/knobs-hash-reranker.test.ts`, `test/cross-modal-phase1.test.ts`, `test/search-alias-resolved-boost.test.ts`) — the same five files #3514 touches, so whichever lands second will hit an explicit (trivial) merge conflict on the pins rather than a silent drift. If #3514 is closed unmerged, v=14 is simply skipped, which is harmless.

## Tests

- `test/search-mode.test.ts`: low/medium/high/unset produce distinct hashes; unset ≡ medium.
- `test/query-cache-knobs-hash.serial.test.ts`: real store/lookup on PGLite — a `detail=low` write is NOT served to a default (medium) lookup; the low caller still hits its own row (mirrors the #2825 hard-exclude isolation tests).

**Fail-on-master proof** (new tests copied onto an unmodified master checkout — hashes collide and the medium lookup HITS the low row):

```
✗ knobsHash determinism + cross-mode separation (CDX-4) > #3515: detail set vs unset produces DIFFERENT hashes
✗ detail cache isolation (#3515) > detail=low write is NOT served to a default (medium) lookup [22.79ms]
✗ detail cache isolation (#3515) > undefined detail keys like the documented medium default
```

With the fix: 153 pass / 0 fail across the five affected files. `bun run typecheck` clean; `bun run verify` 32/32 green.

### DIFF
--- a/src/core/search/hybrid.ts
+++ b/src/core/search/hybrid.ts
@@ -1777,6 +1777,13 @@ export async function hybridSearchCached(
     // resolves) into the cache key so a row written under one exclude
     // policy can't be served to a lookup under another.
     hardExcludes: resolveHardExcludes(opts?.exclude_slug_prefixes, opts?.include_slug_prefixes),
+    // #3515 — fold the EFFECTIVE detail level into the cache key. detail
+    // gates dedup, chunk-source filtering, and the compiled_truth boost, so
+    // a `--detail low` write (compiled-truth-only result set) must never be
+    // served to a default `medium` lookup. Resolve auto-detect the same way
+    // bare hybridSearch does (opts.detail ?? autoDetectDetail(query)) so an
+    // auto-detected `high` query keys like an explicit `high` one.
+    detail: opts?.detail ?? autoDetectDetail(query),
   });
 
   // Cache decision: opts.useCache (explicit) wins over global config; global
--- a/src/core/search/mode.ts
+++ b/src/core/search/mode.ts
@@ -766,7 +766,17 @@ export function attributeKnob<K extends keyof ModeBundle>(
 // written between the #3391 stale-fix (which changes which chunks count as
 // current) and the operator's migration run. Same one-time global cold-miss
 // pattern as the bumps above.
-export const KNOBS_HASH_VERSION = 13;
+//
+// bump 13→15 (#3515): `detail` folds into the key via ctx.detail (det=).
+// detail is result-affecting by design — it gates dedup, chunk-source
+// filtering, and the compiled_truth boost — but was absent from the key, so
+// a `--detail low` write (compiled-truth-only result set) was served to a
+// default `medium` lookup for the whole TTL. Same contamination class as
+// [CDX-4], floor_ratio (v=3), and relationalRetrieval (v=10). v=14 is
+// claimed by in-flight #3514 (compiled_truth boost scope, #3430), so this
+// lands as v=15 per the D8 sequencing convention (see the v=4/v=5 note
+// above). Same one-time global cold-miss pattern as the bumps above.
+export const KNOBS_HASH_VERSION = 15;
 
 /**
  * v0.36 (D8 / CDX-2) — second-arg context for the cache key. The
@@ -805,6 +815,17 @@ export interface KnobsHashContext {
    * 'none' for legacy callers that don't thread excludes.
    */
   hardExcludes?: string[];
+  /**
+   * v=15 (#3515): the EFFECTIVE detail level for this call — per-call
+   * SearchOpts.detail, or the auto-detected level when the caller didn't
+   * specify (hybridSearchCached threads `opts.detail ?? autoDetectDetail(query)`,
+   * matching what bare hybridSearch resolves). detail gates dedup,
+   * chunk-source filtering, and the compiled_truth boost, so a detail=low
+   * write must never be served to a detail=medium lookup. Lives in ctx (not
+   * ResolvedSearchKnobs) because it's per-call, not a mode knob — same path
+   * as col=/prov=. Undefined falls back to 'medium' (the documented default).
+   */
+  detail?: 'low' | 'medium' | 'high';
 }
 
 export function knobsHash(
@@ -898,6 +919,11 @@ export function knobsHash(
     // across processes. Sorted copy so ['a/','b/'] and ['b/','a/'] hash
     // identically; undefined falls back to 'none' for legacy callers.
     `hx=${ctx?.hardExcludes ? [...ctx.hardExcludes].sort().join(',') : 'none'}`,
+    // v=15 addition (#3515, append-only): effective detail level. detail
+    // gates dedup, chunk-source filtering, and the compiled_truth boost, so
+    // a low write (compiled-truth-only set) must never be served to a
+    // medium/high lookup. Undefined falls back to 'medium' (the default).
+    `det=${ctx?.detail ?? 'medium'}`,
   ];
   const h = createHash('sha256');
   h.update(parts.join('|'));
--- a/test/cross-modal-phase1.test.ts
+++ b/test/cross-modal-phase1.test.ts
@@ -136,7 +136,7 @@ describe('D2 — knobsHash differs across cross-modal knob values', () => {
     return resolveSearchMode({ mode: 'balanced' });
   }
 
-  test('KNOBS_HASH_VERSION is 13 (cross-modal still appended; 12→13 embedding-provider migration #3390)', () => {
+  test('KNOBS_HASH_VERSION is 15 (cross-modal still appended; 13→15 detail fold #3515)', () => {
     // v0.35 ladder: 1→2 reranker, 2→3 floor_ratio. v0.36 piggybacks on v=3
     // with 7 cross-modal knobs + column/provider context. v0.40.4 (salem) +
     // v0.39 T21 (master) bump to v=4 for graph_signals + schema-pack fields.
@@ -146,7 +146,8 @@ describe('D2 — knobsHash differs across cross-modal knob values', () => {
     // v0.43: 9→10 relational recall arm. #1400: 10→11 query-side input_type
     // finally reaches asymmetric providers — pre-fix rows were keyed on
     // document-side query vectors. #2825: 11→12 hard-exclude fold (hx=).
-    expect(KNOBS_HASH_VERSION).toBe(13);
+    // #3515: 13→15 detail fold (det=); v=14 claimed by in-flight #3514.
+    expect(KNOBS_HASH_VERSION).toBe(15);
   });
 
   test('flipping unified_multimodal changes the hash', () => {
--- a/test/query-cache-knobs-hash.serial.test.ts
+++ b/test/query-cache-knobs-hash.serial.test.ts
@@ -277,3 +277,38 @@ describe('hard-exclude cache isolation (#2825)', () => {
     expect((await cache.lookup(emb, { knobsHash: envExcludeHash })).hit).toBe(true);
   });
 });
+
+describe('detail cache isolation (#3515)', () => {
+  // Hashes computed the way hybridSearchCached does: same resolved mode, ctx
+  // carrying the effective detail level. A row written by a `--detail low`
+  // call (compiled-truth-only result set) must not be served to a default
+  // `medium` lookup, and vice versa.
+  const lowHash = knobsHash(resolveSearchMode({ mode: 'balanced' }), { detail: 'low' });
+  const mediumHash = knobsHash(resolveSearchMode({ mode: 'balanced' }), { detail: 'medium' });
+  const unsetHash = knobsHash(resolveSearchMode({ mode: 'balanced' }));
+
+  test('detail=low write is NOT served to a default (medium) lookup', async () => {
+    const cache = new SemanticQueryCache(engine);
+    const emb = makeEmbedding(8);
+
+    // Simulate `query "X" --detail low` populating the cache with the
+    // narrow compiled-truth-only result set.
+    await cache.store('what is the deploy process', emb, makeResults('narrow', 2), {
+      vector_enabled: true, detail_resolved: 'low', expansion_applied: false,
+    }, { knobsHash: lowHash });
+
+    // Default-detail lookup inside the TTL → MISS (falls through to a
+    // fresh, full search) instead of the narrow set.
+    expect((await cache.lookup(emb, { knobsHash: mediumHash })).hit).toBe(false);
+
+    // The low-detail caller still hits its own row.
+    const original = await cache.lookup(emb, { knobsHash: lowHash });
+    expect(original.hit).toBe(true);
+    expect(original.results?.length).toBe(2);
+  });
+
+  test('undefined detail keys like the documented medium default', () => {
+    expect(unsetHash).toBe(mediumHash);
+    expect(unsetHash).not.toBe(lowHash);
+  });
+});
--- a/test/search-alias-resolved-boost.test.ts
+++ b/test/search-alias-resolved-boost.test.ts
@@ -89,7 +89,7 @@ describe('alias_resolved boost stage', () => {
 });
 
 describe('KNOBS_HASH_VERSION', () => {
-  it('is 13 (12→13 embedding-provider migration invalidates rows written against the prior embedding space, #3390)', () => {
-    expect(KNOBS_HASH_VERSION).toBe(13);
+  it('is 15 (13→15 detail fold makes detail-contaminated rows unreachable, #3515; v=14 claimed by in-flight #3514)', () => {
+    expect(KNOBS_HASH_VERSION).toBe(15);
   });
 });
--- a/test/search-mode.test.ts
+++ b/test/search-mode.test.ts
@@ -413,7 +413,24 @@ describe('knobsHash determinism + cross-mode separation (CDX-4)', () => {
     // #3390/#3391: bumped 12→13 for the embedding-provider migration wave —
     // legacy callers hash prov=default before AND after a provider swap, so
     // pre-migration cache rows must become unreachable on upgrade.
-    expect(KNOBS_HASH_VERSION).toBe(13);
+    // #3515: bumped 13→15 to fold the effective detail level (det=) — a
+    // detail=low write must not be served to a detail=medium lookup. v=14
+    // is claimed by in-flight #3514 (#3430 compiled_truth boost scope).
+    expect(KNOBS_HASH_VERSION).toBe(15);
+  });
+
+  test('#3515: detail set vs unset produces DIFFERENT hashes (cache contamination prevention)', () => {
+    const knobs = resolveSearchMode({ mode: 'balanced' });
+    const low = knobsHash(knobs, { detail: 'low' });
+    const medium = knobsHash(knobs, { detail: 'medium' });
+    const high = knobsHash(knobs, { detail: 'high' });
+    const unset = knobsHash(knobs);
+    expect(low).not.toBe(medium);
+    expect(medium).not.toBe(high);
+    expect(low).not.toBe(high);
+    // Undefined falls back to 'medium' — the documented default — so legacy
+    // callers that don't thread detail share the default-detail rows.
+    expect(unset).toBe(medium);
   });
 
   test('T1 (codex): floor_ratio set vs unset produces DIFFERENT hashes (cache contamination prevention)', () => {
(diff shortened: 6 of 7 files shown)


time-attack:

A first reviewer concluded this was mergeable. Two independent refuters then re-ran it with different lenses; 2 of 2 killed the recommendation. Their findings:

### Refutation 1

I refute MERGE, but NOT on correctness — the fix itself is real and I verified it harder than the original reviewer did. I refute on (1) the load-bearing half of the diff having zero test coverage (confirmed by my own deletion test), and (2) the PR being literally unmergeable.

WHAT I CONFIRMED IN THE PR'S FAVOR (independently, not on their word)
- I wrote my own end-to-end probe (test/zz-r24-detail-probe.serial.test.ts, ~135 lines, harness copied from test/search-telemetry-cache-wiring.serial.test.ts's mocked-embedQuery seam; deleted after use) driving real hybridSearchCached on PGLite. It goes further than theirs: besides the explicit `--detail low` → default-lookup case, it adds an AUTO-DETECT case with no `--detail` flag anywhere ("who is alice…" → entity → low writes the row; "when did alice last meet…" → temporal → high must not be served it, same mocked embedding so cosine 1.0). On plain master (c6dc0adf, v=14, post-#3514): 2 of 3 FAIL — #3515 reproduces on BOTH the explicit and the flagless auto-detect path. With the PR merged: 3/3 pass.
- No correctness regression is structurally possible from the hash side. Appending a parts entry is a strict refinement of the cache key: every post-PR hit implies the same (query, embedding, knobs, det) pair would also have hit pre-PR and been served the same rows. So the change can only convert hits into misses, never produce a new wrong answer. The escalation residual they flagged (hybrid.ts auto-escalates on `opts?.detail === 'low'`, so an explicit-low run that returns 0 stores its escalated high set under `det=low`) is strictly narrower than master, where every detail level shared one row. Confirmed analytically, agrees with them.
- undefined ≡ 'medium' is behaviorally sound, verified by reading all three consumers: `shouldBoostCompiledTruth` is `detail === 'low'` (hybrid.ts:74, post-#3514), both engines gate on `opts?.detail === 'low'`, and `buildSourceFactorCase` (sql-ranking.ts:66-72) only special-cases 'high' after trim/lowercase. undefined and 'medium' are identical in all of them.
- Invariant sweep clean: no engine SQL, no JSONB cast, no MIGRATIONS entry, no sourceScopeOpts/source-isolation surface, no ctx.remote, no dry-run path. Hash stays 16 hex, so no schema change. knobsHash has exactly one caller (hybrid.ts:1797) used for BOTH cache.lookup and cache.store, so the thread is at the shared choke point.
- Pins: all five exact pins updated, plus master's sixth tolerant pin (test/search-compiled-truth-boost-scope.test.ts:74, `toBeGreaterThanOrEqual(14)`) is satisfied. On my resolved merge: `bun run typecheck` EXIT=0; the six affected files 160 pass / 0 fail.
- Discrimination A (theirs, re-run by me): delete the `det=` parts entry from mode.ts, keep v=15 → 3 fail / 80 pass (search-mode "#3515 detail set vs unset" + both new serial tests). The hash tests are not vacuous.

GROUND 1 — THE WIRING LINE IS 100% OF THE USER-VISIBLE FIX AND 0% TESTED (my own deletion test)
Delete the single line `detail: opts?.detail ?? autoDetectDetail(query)` from src/core/search/hybrid.ts (line 1812 on the merged state) and leave everything else: `ctx.detail` is then undefined for every caller, `det=medium` for everyone, zero segregation, #3515 fully restored. Result: the PR's six affected test files stay **160 pass / 0 fail** (/tmp/r24_discB.txt), and my probe fails IDENTICALLY to plain master. So half the diff is dead code that no PR test can tell apart from the fix.
The only tests anywhere in the repo that flip are two intent assertions in test/hybrid-search-lite.serial.test.ts ("temporal query classifies as temporal", "event query classifies as event"). Those are not coverage: they already FAIL on plain master when that file is batched with the other three hybridSearchCached files (4 fail on master vs 2 with the PR — the PR incidentally fixes them), and they PASS on master in isolation. That file deletes OPENAI_API_KEY to force keyword-only, so the signal is gateway-env leakage across files, i.e. exactly the kind of thing a reviewer dismisses as a flake.
The reviewer found this themselves, wrote a working probe, deleted it, and downgraded it to "non-blocking… matches established repo convention." That defense does not survive checking: the immediately preceding rung of the same version ladder, #3514 (85286a55, the PR this one explicitly sequences behind), shipped a dedicated 112-line test file whose assertions cover 100% of its production change. This PR ships tests covering ~50% of its change, and the uncovered 50% is the entire user-visible effect. Per the stated gate — "stub the old behavior back in and re-run the PR's own tests; if they still pass, the tests prove nothing" — that is NEEDS_CHANGES, and the fix is one small file the reviewer has already written.

GROUND 2 — THE PR CANNOT BE MERGED AS-IS
`gh pr view 3544 --json mergeable` → CONFLICTING / DIRTY. #3514 merged 2026-07-29 taking v=14; this branch is based at v=13 (merge base b3b43d0f). I reproduced the merge onto current master: conflicts in src/core/search/mode.ts + test/cross-modal-phase1.test.ts + test/search-alias-resolved-boost.test.ts + test/search-mode.test.ts + test/search/knobs-hash-reranker.test.ts. I resolved them keeping v=15 and it is green, so this is friction not breakage — but "MERGE" is not an executable verdict on a CONFLICTING PR, and every conflict sits precisely on the cache-version invariant lines, which is the worst possible place to hand a hand-resolution to whoever clicks merge.

GROUND 3 — SHIPPED COMMENTS ARE FALSE ON LANDING
mode.ts and all four pin test files say "bump 13→15" and "v=14 is claimed by **in-flight** #3514". #3514 is merged. The next person bumping the version reads a comment asserting 14 was never taken. CLAUDE.md explicitly forbids this narration rot in reference material. Free to fix inside the conflict resolution that is required anyway.

GROUND 4 — PR TITLE VIOLATES AN IRON RULE
`fix(search): fold detail into the query-cache key (#3515)` — no `vMAJOR.MINOR.PATCH.MICRO` prefix. One `gh pr edit --title`.

SCOPE NOTE (not a defect, but the reviewer's framing overstates it)
"Not a symptom patch… the one-line thread covers every cached-search path" is true only for the `detail` axis. The `query` op (operations.ts:1830-1850) also threads `language`, `symbol_kind`, `types`, `since`, `until`, `salience`, `recency` into hybridSearchCached, and NONE of them appear in the knobsHash parts list I read end to end (mode.ts:816-919). `query "X" --since 2026-01-01` still writes a date-filtered result set into the row an unfiltered `query "X"` reads back — the identical contamination class as #3515, and with `since`/`until`/`types` it is a real filter, not just a re-rank. Pre-existing and out of scope for #3515, but it deserves a follow-up issue rather than the claim that this axis was the whole bug.

VERDICT: NEEDS_CHANGES — (a) keep the behavioral probe of the wiring line (already written, twice, by two reviewers), (b) merge master and correct 13→15 → 14→15 plus drop "in-flight #3514", (c) add the version prefix to the title. The underlying fix is correct and I would approve it the moment the load-bearing line has a test that fails when it is removed.

UNVERIFIED (footprint rule, honestly declared): full unit suite / `bun run verify` / `ci:local` — deliberately not run; I ran `bun run typecheck` plus the six affected files, the five hybridSearchCached test files, and my probe. Real Postgres — no Docker; the diff touches no engine code so PGLite-only is sufficient. CI shard behavior after the required master merge — verified locally only. Cache hit-rate cost of keying on the auto-detected level (paraphrases that classify into different intents no longer share rows) — correct by design, unmeasured.

**Defect the first reviewer missed:** Untested load-bearing wiring: src/core/search/hybrid.ts's `detail: opts?.detail ?? autoDetectDetail(query)` (line 1812 on the merged state) is the entire user-visible fix — without it `ctx.detail` is undefined for every caller, `knobsHash` emits `det=medium` unconditionally, and #3515 is fully restored — yet deleting it leaves the PR's six affected test files at 160 pass / 0 fail (independently reproduced), and my own end-to-end probe against hybridSearchCached then fails identically to plain master on BOTH the explicit `--detail low` path and the flagless auto-detect path. The only repo tests that flip are two gateway-env-leakage-dependent intent assertions in test/hybrid-search-lite.serial.test.ts that already fail on master when batched and pass on master in isolation, so they are not usable coverage. A future refactor can delete the fix and CI stays green. The immediately preceding rung of the same version ladder, #3514 (85286a55), shipped a dedicated test file covering 100% of its production change, so "repo convention" does not excuse the gap. Secondary: the PR is CONFLICTING/DIRTY against master (v=14 landed via #3514; branch based at v=13) with all five conflicts on the cache-version invariant lines, and its shipped comments in mode.ts plus four pin test files still say "13→15" and "v=14 is claimed by in-flight #3514", which is false the moment it lands.

### Refutation 2

REFUTED — NEEDS_CHANGES, but narrowly: the fix itself is correct and I verified it harder than the prior reviewer did. What fails is mergeability plus the discrimination test as literally specified.

WHAT I CONFIRMED IS GOOD (independent, not taken on their word):
- End-to-end correctness, both directions AND the auto-detect half. My probe (result-set-level assertions vs cache-off baselines, deleted afterward) passes 4/4 on the PR merged with current master; 2 of those 4 fail with the wiring line removed.
- Single-seam coverage is genuine: `knobsHash(` has exactly one caller in src/ (src/core/search/hybrid.ts:1797) and `cacheKnobsHash` feeds BOTH `cache.lookup` (:1881) and `cache.store` (:2017). The inner `hybridSearch(engine, query, {...opts})` (:1955) receives the same query string and opts, and resolves detail identically at hybrid.ts:967 — so the hashed value equals the value that shapes the stored result set. Not a symptom patch.
- Direction-safety: identical effective-detail string ⇒ identical behavior, different string ⇒ separate row. The fold can therefore only cost extra misses, never false sharing. `undefined ≡ 'medium'` holds in every detail-sensitive site I grepped: `shouldBoostCompiledTruth` (`=== 'low'`, hybrid.ts:74), both engines' compiled_truth filters (`opts?.detail === 'low'`, pglite-engine.ts:1602/1744/1991/2080 and the postgres mirror), and `buildSourceFactorCase`'s `'high'` bypass (sql-ranking.ts:66-72). Not a scope collapse.
- Pin-site census on CURRENT master (#3514 merged 2026-07-29, master at v=14): 5 exact pins (test/search-mode.test.ts:419,585; test/search/knobs-hash-reranker.test.ts:70; test/cross-modal-phase1.test.ts:150; test/search-alias-resolved-boost.test.ts:93) + 1 tolerant pin added by #3514 (test/search-compiled-truth-boost-scope.test.ts:74, `>=14`). All 5 at 15 in the PR, tolerant one satisfied.
- Diff is exactly the 7 files gh shows (`git diff --stat b3b43d0f 7ff32a87`), no hidden files. No engine method, SQL, JSONB cast, migration, `sourceScopeOpts`, `ctx.remote`, or dry-run surface touched; hash stays 16 hex (no schema change); no new test file, so no shard re-binning.
- Runs: `bun run typecheck` EXIT=0 on PR head; `bun test` the 5 pin files + new serial = EXIT=0, 153 pass/0 fail. Merged origin/master locally (conflicts in mode.ts + the 4 pin files; hybrid.ts auto-merged cleanly and both #3514's `shouldBoostCompiledTruth` and this PR's wiring line survive), resolved keeping v=15 (highest wins): typecheck EXIT=0, 6 affected files 160 pass/0 fail, and test/search-telemetry-cache-wiring.serial.test.ts (4 pass) + test/hybrid-cached-hit-budget-meta.serial.test.ts (1 pass) green in isolation. Discrimination A reproduced too: delete the `det=` parts entry, keep v=15 → 3 fail / 80 pass, so the hash tests are not vacuous.

WHY I STILL REFUTE THE MERGE VOTE:
1. (Primary, CONFIRMED — see new_defect) Stub the old production behavior back in and the PR's own tests are 100% green. The two new tests are misleading by name — they claim "#3515 detail cache isolation" while mirroring the key computation rather than exercising it. The behavior-delivering line is a one-line thread that a future refactor (moving the autoDetect call, re-deriving detail inside bare hybridSearch) breaks silently in CI. The repo's own history is a list of exactly this class re-shipping. "Prior bumps only test the pure hash" is a weak defense: those bumps (`hx=`, `col=`/`prov=`) threaded values already produced by `resolvedForCache`/`resolveHardExcludes`, whereas `det=` introduces a NEW resolution expression that exists nowhere else and duplicates hybrid.ts:967 by hand. The wiring test costs ~60 lines on a harness that already exists.
2. (CONFIRMED) `gh pr view` reports `mergeable: CONFLICTING`. It is not mergeable as-is; the author should merge master and re-push so CI runs the shards on the real merged tree (I verified the resolution locally but did not run shards).
3. (CONFIRMED) Title `fix(search): fold detail into the query-cache key (#3515)` violates the version-first IRON RULE — no `vMAJOR.MINOR.PATCH.MICRO` prefix. Fix with `gh pr edit 3544 --title`. Wave siblings doing the same doesn't make it compliant.
4. (Housekeeping, load-bearing for the next bumper) The body, the mode.ts comment block, and 3 test messages all say "v=14 claimed by in-flight #3514". #3514 is MERGED and bumped to 14 WITHOUT adding a comment to mode.ts's ladder, so after resolution the ladder documents 13→15 with no 14 entry. Rewrite to past tense and record what 14 was, or the next bumper re-derives it from git.
5. (PLAUSIBLE, pre-existing but should be a comment/TODO on the PR) hybrid.ts:1608 escalates on `opts?.detail === 'low'`, not on the effective level: an explicit-low call returning 0 recurses at `detail:'high'`, and that FULL set is stored under `det=low`. A later embedding-similar low lookup can be served results a low search would not produce, so "det=low rows hold low-detail sets" is not an invariant. Strictly narrower than master (one shared row for all levels), hence not a regression — but the PR's narrative overstates the guarantee.

Blocking cost here is one round trip on a well-specified, mostly mechanical ask (merge master, add the wiring test, fix title/wording). Merging as-is means a permanently unguarded load-bearing line in the exact subsystem this repo has re-broken three times, on a branch GitHub currently refuses to merge.

UNVERIFIED (footprint rule, honest): full unit suite / `bun run verify` / `ci:local` not run; CI shard behavior on the merged tree not run; real Postgres not exercised (no Docker — diff touches no engine code, PGLite-only is adequate here); the permanent (not one-time) cache hit-rate cost of keying on the AUTO-DETECTED level — paraphrases that classify differently can no longer share a row, correct by design but unmeasured, and the PR body only advertises the one-time cold miss; and src/core/operations.ts:1771's unvalidated `(p.detail as 'low'|'medium'|'high')` cast, which now lets an untrusted MCP caller mint arbitrary `det=<string>` cache rows (I reasoned it is behaviorally safe — extra rows only, no false sharing, and remote callers can already mint rows via arbitrary query text — but I did not test it).

**Defect the first reviewer missed:** The PR's entire test suite passes with the fix removed. I deleted the one production line that delivers the user-visible fix — `detail: opts?.detail ?? autoDetectDetail(query)` (src/core/search/hybrid.ts:1786 on the PR head; :1812 after merging master) — left the `det=` hash part and v=15 in place, and re-ran the PR's own five pin files plus its new serial file: EXIT=0, 153 pass / 0 fail. That includes both brand-new tests in the `describe('detail cache isolation (#3515)')` block, which compute `lowHash`/`mediumHash` themselves and drive `SemanticQueryCache` directly, never touching the production cache key. So the PR's suite would be fully green on a variant of this PR that fixes nothing about #3515: the tests pin the pure hash function, and the only caller that feeds `ctx.detail` (src/core/search/hybrid.ts:1797 `knobsHash(...)`, the single `knobsHash(` call site in src/) is unguarded. My own probe proves the seam is cheap and available: ~60 lines reusing the existing mocked-`embedQuery` PGLite harness verbatim from test/search-telemetry-cache-wiring.serial.test.ts, asserting on served RESULT SETS (not just `cache.status`) against cache-off baselines, gives 4 pass on the PR and 2 fail with that line deleted — one on the explicit `--detail low` write → default lookup path (the #3515 bug reproduced: `cache.status` 'hit' and the narrow set served), and one on the auto-detected-level path (`'who is Quokka Truth'`, auto-low, poisoning a general query with an identical mocked embedding), which the prior reviewer's probe did not cover.

---

time-attack:

Genuine cache-contamination bug still on master: detail gates the compiled_truth boost (src/core/search/hybrid.ts:74-75, post-#3514 boost only at detail=low), the high-detail boost skip (hybrid.ts:1475), and low->high auto-escalation (hybrid.ts:1609), yet cacheKnobsHash at hybrid.ts:1798 omits it — a --detail low write is served to a medium lookup for the TTL. Verified fail-without-fix (2 new tests fail on master src, pass with fix; 153/153 green + typecheck clean on a resolved merge). Cannot merge as-is: the PR claims KNOBS_HASH_VERSION=15 but master's #3677 (b4a9c768, FTS fold) already took v=15 (#3514 took v=14), so it needs a rebase to v=16 with the 5 test pins updated.

**Conflict state vs current master:** Conflicts in src/core/search/mode.ts (version comment ladder + knobsHash parts list, where both #3677's fts= and this PR's det= append at the tail) and 5 test pin files; hybrid.ts merges clean. Semantic but mechanical: resolve as v=16, keep both fts= and det= parts, re-pin the 5 tests to 16, reword the ladder comment (v=14 is #3514, v=15 is #3677). I validated exactly this resolution: typecheck clean, 153/153 pass. Trap: the KNOBS_HASH_VERSION = 15 line itself auto-merges (both sides wrote 15), so a naive both-blocks resolution ships a changed parts list without a bump — functionally safe (det= still changes the hash string) but violates the pin convention.

**Risk if merged as-is:** If resolved lazily at v=15 the stale-row invalidation still works (parts change forces new hashes) but the pin convention breaks and future bumps double-book v=16; the 5 test pins fail CI unless updated consistently. Post-merge: one-time global query-cache cold-miss, same as every prior bump. No engine-parity/jsonb/source-scoping surface touched.

time-attack:
Landed in the v0.45.10.0 community fix-wave 2 (#4085) as 5bd2c51053, rebased/repaired against current master during wave assembly.

Issue #2943, opened 2026-07-17
flaky: brain-durability post-commit hook tests fail intermittently under CI parallelism (3x in one day, unrelated diffs)


time-attack:
The serial-tests job intermittently fails on `test/brain-durability-hook.serial.test.ts` timing-sensitive cases — observed 3 times today on completely unrelated diffs:

- "the hook works even with the committed helper deleted (self-contained)" — on a docs-only PR (#2798 run) and on #2932 (facts extraction)
- "a direct commit auto-pushes in the background" — on #2940 (postgres-engine only)

Each passed on rerun, and the fix-wave E agent independently confirmed the "LOCAL-ONLY line" case reproduces as flaky on unmodified master under load (3/3 both ways). The tests assert on background git operations with fixed timing windows; under a loaded CI runner the background push/hook hasn't completed when the assertion fires.

Suggested fix: poll-with-deadline instead of fixed sleeps around the background-push assertions (the repo already has this pattern in other serial tests), or gate the timing-sensitive cases behind a retry helper. Low priority for correctness, high value for CI signal quality — every flake costs a full rerun cycle and erodes trust in red.

time-attack:
4th occurrence: the LOCAL-ONLY case failed master's push run for 8b325041 (run 29764977261), passing on rerun. Also logging a SECOND flaky case observed in the same push: E2E 'JSONB roundtrip — v0.12.1 reliability wave' failed at its 5000ms cap (run 29764978464) and passed on rerun — likely the same loaded-runner timing class; consider a poll-with-deadline or a larger cap there too.

time-attack:
3 more hits on 2026-07-24, all the same case (`post-commit hook (D9 local, D7 self-contained) > logs a clear LOCAL-ONLY line when origin is unreachable`), all on unrelated diffs:

- master run 30118513811 (batch tip 540b86ff)
- PR #3338 run 30115666230
- PR #3341 run 30115833276

All cleared on rerun. Hit rate is high enough now that this is costing a rerun on roughly every other merge-train batch — bumping priority on an actual fix (likely needs the origin-unreachable simulation to not race the hook's network timeout under shard parallelism).

time-attack:
[...]

PR #3546, opened 2026-07-28
fix(cli): stop parseOpArgs hanging on a non-TTY stdin with no input (#3513)

### LINKED ISSUE #3513: parseOpArgs stdin read blocks forever in a non-TTY with no piped input
Surfaced while reviewing #3503 (not caused by it — this is on master).

`parseOpArgs` reads stdin for stdin-capable ops at `src/cli.ts:808` via `readFileSync(0)`. In a non-TTY with **no piped input** — a CI step, a cron job, an agent harness that inherits a non-TTY stdin without writing to it — that call never returns. It hung a review scan for 5 minutes before I killed it.

Distinct from #380: that one is about flags being swallowed. This is the blocking read those swallowed flags fall through *into*, and it reproduces on its own whenever a stdin-op is invoked with neither an inline value nor piped input.

Minimal fix: guard the read on stdin actually being readable — check `process.stdin.isTTY` is falsy **and** that data is available (or apply a short deadline and treat a timeout as empty), rather than assuming non-TTY implies piped content. Failing fast with the existing missing-content error is the right outcome.

time-attack:
Fixes #3513.

## The bug

`parseOpArgs` read stdin for stdin-capable ops via `readFileSync(0)`, assuming non-TTY implies piped content. In a non-TTY with **no piped input** — a CI step, a cron job, an agent harness that inherits a non-TTY stdin without writing to it — that call never returns (it hung a review scan for 5 minutes).

## The fix

The stdin fill moves out of `parseOpArgs` into an async `applyStdinParam` + `readStdinBounded` (both exported), called by the op dispatch immediately after arg parsing:

| stdin kind | behavior |
|---|---|
| TTY | skipped (as before) |
| regular file / `/dev/null` (fstat: not pipe/socket) | `readFileSync(0)` — returns without blocking; `gbrain put x < /dev/null` still yields `''` (pre-fix parity) |
| FIFO/socket | stream-read with a deadline on the **first byte only** (default 5000ms; `GBRAIN_STDIN_TIMEOUT_MS` overrides). Once any byte arrives the deadline lifts and the read drains to EOF — slow producers keep working. Empty-but-closed pipe → `''`. Never-written pipe → timeout → param stays unset → the existing required-param usage error fails fast (exit 1) |

Cases deliberately NOT broken (each pinned by a test): `echo foo | gbrain put x`, heredocs, `< /dev/null`, `: | gbrain put x`, and interactive TTYs.

## Tests (`test/cli-stdin-hang.test.ts`)

- **Load-bearing repro**: spawns the real CLI (`gbrain put <slug>`) with a held-open, never-written pipe. Exits ~1s with `Usage: gbrain put` on the fix; hangs until the 20s observation window kills it on master.
- Parity: data pipe, `/dev/null`, empty closed pipe (no hang, clean exit).
- Subprocess driver pinning content preservation through `applyStdinParam` (verbatim pipe content, `''` for readable-empty, unset on timeout).

**Fail-on-master proof** (test file copied onto an unmodified master checkout):

```
✗ #3513 — stdin-capable op with a non-TTY, never-written stdin > exits fast with the usage error instead of blocking forever [20009.12ms]   ← hung until killed
✗ … applyStdinParam content preservation (4 driver tests; helper doesn't exist on master)
 3 pass (the parity cases master already handled)
 5 fail
```

With the fix: 8 pass / 0 fail in ~2s. `bun run typecheck` clean; `bun run verify` 32/32 green.


time-attack:
Rebasing onto current master: this branch's CI ran at 20:51:30Z and the `post-commit hook > logs a clear LOCAL-ONLY line` failure is the #2943 flake, which #3537 fixed at 20:58:10Z — seven minutes later. Not attributable to this diff (which touches `src/cli.ts` only). Re-running against the fix.

PR #3548, opened 2026-07-28
fix(doctor,upgrade): warn brains pinned to a provider shutting down 2026-09-04

time-attack:
## Why

The hosted embedding + rerank API that the shipped defaults resolve to shuts down on **2026-09-04**. Query embedding goes through the same endpoint as ingestion, so on that date affected brains don't just stop indexing new content — **existing vectors become unqueryable** and semantic retrieval dies entirely (verified: the query path resolves the same `provider:model` recipe/endpoint as the ingest path).

`gbrain migrate embeddings` (#3459) is the fix, but the only discovery surface was a one-shot `gbrain upgrade` banner with a generic command. Brains with `zeroentropyai:zembed-1` written into config — including every brain the `ze_switch` flow deliberately migrated onto it — need the migration run regardless of how the default-change question (#3481/#3492, on hold) lands.

## What

- **New `provider_sunset` doctor check** — fires on every `gbrain doctor` run until the brain is off the provider (warn before the date, **fail** after, when retrieval is already down). The paste-ready `gbrain migrate embeddings --to <provider:model> --dim <N> --dry-run` hint fills `--dim` from the brain's **actual** `vector(N)` column width (config can drift; a wrong `--dim` forces a needless dimension transition + index rebuild). The self-host escape hatch — zembed-1 weights are Apache-2.0; llama-server/Ollama keep every existing vector with **no migration at all** — gets equal billing. Categorized `ops` per `doctor-categories.ts`.
- **Upgrade banner** — same actual-width `--dim`, states plainly that existing vectors become unqueryable (users assume only new content is affected), points at the recurring doctor check. Date now lives once (`ZEROENTROPY_SUNSET_DATE`, `src/core/ai/defaults.ts`).
- **`EmbedResult.lock_skipped` + accurate pause message** — found during end-to-end verification: a migration killed with SIGKILL leaves its single-flight embed lock behind (60-min TTL), and the immediate re-run skipped the re-embed while still printing "re-run the same command to resume" — a hint the re-run could not honor. `migrate embeddings` now names the lock as the blocker and when it clears; `--json` carries `lock_skipped: true`.
- **Docs** — migration guide (discovery surfaces, `--dim` guidance, hard-kill lock caveat), provider matrix + provider doc shutdown notes, KEY_FILES. `bun run build:llms` re-run: no bundle drift (edited docs aren't inlined).

## End-to-end verification of the migration path (scratch PGLite brain)

Simulated the affected starting state (config pinned to `zeroentropyai:zembed-1` @1280d, 330 pages / 348 chunks, correctly-shaped vectors + signature stamps written directly — no hosted key available), then migrated to a local OpenAI-compatible mock provider:

- same-dim provider swap: completed; vectors replaced; both config planes updated; signatures restamped; completion marker written; post-migration search returns sensible results (query embeds through the new provider)
- dimension change (1280→768→1280): schema transition ran; completed
- `kill -9` mid-run: state marker + NULL-embedding cursor held; re-run resumed **exactly** the 270 remaining chunks with zero duplicate re-embedding (verified via mock-server request accounting) — after the stale lock cleared (the `lock_skipped` finding above)

Not exercised (no hosted keys): real zembed-1 embeds, real hosted target providers, the Postgres engine path (PGLite only; `test/e2e/migrate-embeddings-postgres.test.ts` covers that in CI).

## Tests

`test/provider-sunset-doctor.serial.test.ts` — drives the master-existing `buildChecks` / `runEmbedCore` seams, so on unmodified master it fails **behaviorally** (doctor never surfaces the sunset; lock-skip invisible on the result): 5 fail / 1 pass on master, 6/6 here. Affected suites green (migrate-embeddings flow/boundary, embedding-migration, embed.serial, doctor-behavioral, doctor-categories, upgrade suites, build-llms). `bun run typecheck` + `bun run verify` clean.

## Out of scope (deliberate)

`DEFAULT_EMBEDDING_MODEL` / `DEFAULT_RERANKER_MODEL` unchanged — #3481/#3492 are owner decisions on hold. Every surface here is correct whichever way those land: a default change only rescues fresh installs; any brain with existing vectors in the sunsetting space needs the migration (or the self-host path) either way.


time-attack:

Adds a recurring `provider_sunset` doctor check + a richer `gbrain upgrade` banner for brains pinned to `zeroentropyai:*` before its 2026-09-04 hosted-API shutdown, plus a genuinely useful `EmbedResult.lock_skipped` flag so `migrate embeddings` stops printing a resume hint a re-run cannot honor. Everything passes today (6/6 own test, typecheck 0, verify 32/32, 155 affected-suite tests 0 fail) and the discrimination test is honest — 5 fail / 1 pass with the pre-fix src stubbed back in, with the first failure behavioral through `buildChecks`. Two defects block it: the date escalation to `fail` is unconditional on evidence, so on 2026-09-04 every brain on the shipped defaults (which this PR deliberately does not change) reports `unhealthy` and `gbrain doctor` exits 1 — I reproduced an existing CI test breaking that way — and the reranker branch reads the wrong plane, so it reports `ok` for exactly the post-migration brains it exists to protect.

### Discrimination test

CONFIRMED — it discriminates. From the PR head (fd9522f7) I checked out the parent's version of all six touched src files (`git checkout 6920744d -- src/commands/{doctor,embed,migrate-embeddings,upgrade}.ts src/core/ai/defaults.ts src/core/doctor-categories.ts`) while keeping the new test, and ran it: EXIT=1, 5 fail / 1 pass — exactly the PR's claim. The first failure is behavioral, not a missing export: `buildChecks surfaces the sunset ...` fails because master's doctor never emits a `provider_sunset` check, and `a held per-source lock makes runEmbedCore report lock_skipped` fails on `result.lock_skipped` being undefined while `embedded` is still 0. On the PR: 6 pass / 0 fail. (Tests 2-4 fail on the missing `checkProviderSunset` export, which is the weaker form, but test 1 carries the behavioral pin.)

### Blockers

**1.** src/commands/doctor.ts:2381,2411 — `provider_sunset` escalates to status `fail` on `Date.now() >= sunset` + a config STRING alone, with no evidence the brain ever used the provider. `DEFAULT_EMBEDDING_MODEL` stays `zeroentropyai:zembed-1` (the PR puts that change out of scope, #3481 on hold), so after 2026-09-04 every default-config brain fails. EXECUTED PROOF: set `ZEROENTROPY_SUNSET_DATE='2026-01-01'` (the only date input) and ran `bun test test/doctor-cli-smoke.serial.test.ts` — baseline 1 pass EXIT=0, post-date 1 fail EXIT=1, `doctor --json` returned `status:"unhealthy", health_score:55`, exit 1, with `provider_sunset` the SOLE `fail` in 90 checks. The brain in that test is a fresh `gbrain init --migrate-only` with 0 pages, no ZEROENTROPY_API_KEY (`embedding_provider: "Skipped (no provider credentials)"`, `embeddings: warn "No embeddings yet"`), yet the top-ranked issue reads "semantic retrieval is offline (queries can no longer be embedded against your existing vectors)" — factually wrong for a brain with zero vectors. Impact: (a) test/doctor-cli-smoke.serial.test.ts breaks in CI on 2026-09-04 with no code change, 37 days out; (b) every fresh install and every doctor-as-CI-gate exits 1 from that date. Gate the `fail` on real exposure (credentials present AND non-zero embedded chunks in that space) or keep it at `warn`.

**2.** src/commands/doctor.ts:2368,2373 + src/commands/upgrade.ts:477 — reranker exposure is read from the wrong plane, producing a false `ok`. The check uses `getRerankerModel()` (gateway plane, `undefined` unless explicitly configured — gateway.ts:809-811 deliberately does NOT fall back) and the banner uses `engine.getConfig('search.reranker.model')` (also unset by default), but the model search actually reranks with comes from the mode bundle: EXECUTED `resolveSearchMode({})` → `reranker_enabled: true, reranker_model: 'zeroentropyai:zerank-2'`, and hybrid.ts:1624 passes that as `opts.model` which rerank() prefers over everything. Failing input: a brain that followed this PR's own advice and ran `gbrain migrate embeddings --to openai:...` (which by design does not touch the reranker, docs/guides/embedding-migration.md:125) and never set `search.reranker.model` → `provider_sunset` returns `ok`, message "No configured provider has an announced shutdown", while its search reranks through zerank-2 and silently degrades to unreranked ordering on 2026-09-04. Asymmetric with the embedding branch two lines above, which correctly falls back to the shipped default. Root cause is duplicated affected-detection in two call sites — resolve the effective reranker once (resolveSearchMode/MODE_BUNDLES) and share it.

**3.** test/provider-sunset-doctor.serial.test.ts:65 — `const FLAGGED = ['warn','fail']` makes every status assertion date-tolerant, so no test anywhere pins the post-sunset behavior. That is precisely why blocker 1 is invisible: the suite is green on both sides of the date flip while the product hard-fails on one side. An injectable clock (or reading the constant and asserting both branches) is ~5 lines and would have caught it.

### Not verified

- Postgres-engine doctor path — no DATABASE_URL available here, so `test/e2e/mechanical.test.ts` ('gbrain doctor exits 0 on healthy DB') was not executed. Read-only inspection says it is NOT exposed to blocker 1: it pins `--embedding-model openai:text-embedding-3-large` at init (mechanical.test.ts:1341-1352). The PGLite subprocess smoke test I DID execute has no such pin, which is why it breaks.
- Real hosted-provider behavior on/after 2026-09-04 (no ZeroEntropy key, and the endpoint is still live) — the sunset date itself is taken from the PR's claim; I verified only what the code does on either side of that constant.
- The PR's own end-to-end migration narrative (330-page scratch brain, kill -9 resume, mock-provider request accounting) — not reproduced; I exercised the committed test seams instead.
- `base_urls.zeroentropyai` self-host override: gateway.ts:417/2006/2147 do honor a per-recipe base_url override, so a brain keeping `zeroentropyai:zembed-1` pointed at a local llama-server would be falsely flagged (and hard-failed after the date). I did not build that config end to end. Low severity — docs/guides/embedding-migration.md:132 steers self-hosters to a different recipe id — but the check could cheaply exclude an overridden base_url.

- `gh pr view 3548 / gh pr diff 3548 — exit 0`
- `git fetch origin pull/3548/head:pr3548 && git checkout pr3548 — exit 0 (head fd9522f7, base 6920744d, 25 commits behind origin/master, gh reports MERGEABLE/CLEAN)`
- `bun install — exit 0`
- `bun test test/provider-sunset-doctor.serial.test.ts — exit 0 (6 pass / 0 fail)`
- `bun run verify — exit 0 (pass=32 fail=0)`
- `bun test test/doctor.test.ts test/doctor-behavioral.test.ts test/doctor-categories.test.ts test/embedding-migration.test.ts test/build-llms.test.ts — exit 0 (155 tests, 0 fail)`
- `DISCRIMINATION: git checkout 6920744d -- <6 touched src files>; bun test test/provider-sunset-doctor.serial.test.ts — exit 1 (5 fail / 1 pass)`
- `bun test test/doctor-cli-smoke.serial.test.ts (unmodified sunset date) — exit 0 (1 pass)`
- `POST-DATE SIM: ZEROENTROPY_SUNSET_DATE='2026-01-01'; bun test test/doctor-cli-smoke.serial.test.ts — exit 1 (1 fail, doctor exitCode 1, provider_sunset sole 'fail')`
- `bun run rr-probe.ts (resolveSearchMode({})) — exit 0 → reranker_enabled: true, reranker_model: zeroentropyai:zerank-2`
- `git checkout src/core/ai/defaults.ts; git status --short — clean, worktree restored to fd9522f7`

</details>

---

time-attack:
Review finding (blocking): the recurring-check and `lock_skipped` fixes are solid maintenance-lane work, and the 2026-09-04 date is real (master already ships the upgrade banner). The blocker is severity escalation: the new check returns `status:'fail'` once the date passes, and because the check falls back to `DEFAULT_EMBEDDING_MODEL` when the gateway is unconfigured, every stock-default brain — including ones that never held a ZeroEntropy key or a single affected vector — will report `unhealthy` and `gbrain doctor` will exit nonzero in ~33 days, with no suppression key (the banner has `ze_sunset_notice_shown`; this has nothing).

Three changes would unblock it: (a) gate the check on evidence the brain is actually affected (a ZE key present, or vectors embedded by the affected provider) rather than the resolved default; (b) keep severity at `warn` unless affected vectors exist; (c) add a suppression config so CI/cron keyed on doctor's exit code has an off switch. Also worth freezing the clock in the tests — they flip behavior on the sunset date as written.

time-attack:
Landed in the v0.45.10.0 community fix-wave 2 (#4085) as fd9bb12b42, rebased/repaired against current master during wave assembly.

PR #2568, opened 2026-07-03
feat(import): seed a brain from AI chat exports (ChatGPT/Claude) via memvelope envelopes

SeanGearin:
closes #2567

what this adds: a reference importer that seeds a gbrain from any AI chat export. flow: export -> memvelope envelope (free in-browser converter at memvelope.com, or any conforming tool) -> this script -> markdown pages your sync ingests. one command from "years of chat history" to a populated brain.

why: every new brain starts empty, and the cold-start problem isn't building the brain - it's filling it. this makes "bring your whole history" an onboarding path instead of a blank page.

what lands: `scripts/envelope-to-gbrain.mjs` (zero deps, deterministic, no network), a small sample envelope as a test fixture, a self-contained test, and a docs section. one page per conversation - date + conversation-id filenames (collision-proof), provenance frontmatter (origin / source / conversation id), message-id citations, `type: conversation`.

receipts: live-verified against upstream v0.42.56.0 - sample envelope -> 1 page; a real 662MB Claude export -> 353 conversations = 353 distinct pages, keyword-searchable, provenance + citations intact. the envelope side is conformance-tested upstream (golden fixtures, byte-for-byte determinism, timezone matrix).

posture: no behavior change to existing paths, no new deps. memvelope is MIT and vendor-neutral - gbrain was the biggest influence on it and is credited in the spec. happy to adapt naming and placement to your conventions, and to maintain this going forward.

PR #3549, opened 2026-07-28
feat(import): standalone importer seeding a brain from envelope-v0 chat-history files

### LINKED ISSUE #2567: import: seeding a brain from AI chat exports (ChatGPT/Claude) via memvelope envelopes
every new gbrain starts empty - and the densest personal corpus most people have is years of AI chat history locked in per-vendor export formats. we run gbrain in production and hit this on day one: the cold-start problem isn't building the brain, it's filling it.

we built an open MIT standard for the export side (memvelope - one portable JSON envelope any chat export converts into; gbrain was the biggest influence on how we think about portable memory, credited in the spec) and a reference importer that maps an envelope into gbrain's markdown-brain-repo model: one page per conversation, provenance frontmatter, message-id citations, `type: conversation`.

live-verified against upstream v0.42.56.0: a real 662MB Claude export -> 353 conversations -> 353 distinct pages, keyword-searchable, provenance and citations intact. deterministic, zero deps, no network.

PR follows with the importer adapted to this repo's conventions plus a self-contained test. no behavior change to any existing path. happy to maintain it.

SeanGearin:
closes #2567

Resubmission of #2568 per the closure guidance: script standalone, endorsement stripped to a spec link.

**What this adds:** `scripts/envelope-to-gbrain.mjs` - a zero-dependency, deterministic, no-network importer that turns an envelope-v0 file (format spec: github.com/memvelope/memvelope) into one Markdown page per conversation, written into a directory `gbrain sync` ingests. Solves cold-start: a new brain can ingest years of existing ChatGPT/Claude history instead of starting empty.

**What lands:** the script, a self-contained test (8 cases), and a small fixture envelope. No docs guide - usage, output layout, edge cases, and verification steps are in the script header. No behavior change to existing paths, no new deps.

**Receipts:** rebased and re-verified on current master - `bun test test/envelope-to-gbrain.test.ts` 8/8; fixture run writes 1 page. Previously live-verified end-to-end: a real 662MB Claude export -> 353 conversations = 353 distinct pages, searchable after sync, provenance and message-id citations intact.

Happy to adjust naming/placement to repo conventions and to maintain this going forward.

### DIFF
--- a/scripts/envelope-to-gbrain.mjs
+++ b/scripts/envelope-to-gbrain.mjs
@@ -1,0 +1,103 @@
+#!/usr/bin/env node
+/**
+ * Import an envelope-v0 file (a JSON serialization of AI chat history; format
+ * spec: github.com/memvelope/memvelope) into a brain repo as one Markdown page
+ * per conversation, which `gbrain sync` ingests.
+ *
+ * Usage:
+ *   node scripts/envelope-to-gbrain.mjs <envelope.mve.json> [outDir]
+ *
+ * Zero dependencies. Deterministic. No network. It does NOT call gbrain — it
+ * only writes Markdown files.
+ *
+ * Output layout:
+ *   - One page per conversation, filename = date + conversation id (shared
+ *     titles cannot collide; the id is the natural key). A duplicate id
+ *     overwrites its own filename and warns on stderr; stdout reports DISTINCT
+ *     files written, not write calls.
+ *   - Frontmatter: `type: conversation` (keeps pages eligible for
+ *     conversation-facts extraction and chronicle behavior after sync), the
+ *     source provider, the conversation id, and `origin: memvelope/envelope-v0`.
+ *   - Page `date` is the first 10 chars of the conversation's ISO-8601
+ *     `created_at`. Body keeps message-id citations beside each speaker turn.
+ *
+ * Memory: the whole envelope is held in memory (no streaming); envelopes are
+ * far smaller than the vendor exports they serialize.
+ *
+ * Verify:
+ *   node scripts/envelope-to-gbrain.mjs test/fixtures/memvelope/sample.mve.json /tmp/out
+ *     -> expect "wrote 1 markdown page(s)"
+ *   bun test test/envelope-to-gbrain.test.ts
+ *
+ * STATUS: live-verified against gbrain v0.42.56.0 on 2026-07-03: the sample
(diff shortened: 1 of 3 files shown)


time-attack:
Taking it. You did what was asked on #2568 — neutral spec link, standalone script, zero dependencies, no network, no change to any existing path — and the maintainer's call is to accept it.

Two small things worth a follow-up PR whenever convenient. Neither blocks the merge:

1. **Unquoted YAML frontmatter value.** `source: ${env.meta?.source_provider}` is interpolated raw, so a provider string containing a newline could inject additional frontmatter keys. Low stakes given the input is a local file the user chose, but `JSON.stringify` on that interpolation closes it for free — and it's the same shape as an injection bug in a less benign context.
2. **`undefined` leaks as a literal.** `memvelope_conversation_id: ${JSON.stringify(c.id)}` emits the four-character string `undefined` when `c.id` is absent, rather than omitting the key.

Both are one-liners. Happy to take them together in a small follow-up.

For the record on why this went to the maintainer rather than being auto-closed: gbrain's bar for new surface is deliberately high and third-party-format importers carry a permanent maintenance obligation to a spec someone else controls. But you were explicitly invited to resubmit, and closing invited work is a cost that belongs to the maintainer to spend, not to a triage pass. That's now happened twice in two days, so it's worth the project settling a standing rule for invited resubmissions.

SeanGearin:
Appreciate you taking it, and genuinely appreciate the reasoning you put in the record about invited resubmissions - that was more than the PR needed, and I noticed. I'd support a standing rule whichever way it lands.

Both findings are right and I'm taking both, so please don't spend your own time on them.

On the second one, the reason it's worth fixing rather than tidying: `undefined` asserts a value instead of reporting absence. Every id-less conversation claims the same id, so anything grouping or deduping on that key merges unrelated pages. A missing key can be filled in later; a false one has already propagated.

Root cause on both is the same shape of miss. `title:` directly above was already quoted and `source:` wasn't, and the existing missing-id test asserted the fallback filename without ever looking at the frontmatter that same envelope produced. The follow-up puts the filename and the frontmatter behind one `hasId` predicate so they can't disagree again, and the injection case parses the output with js-yaml rather than substring-matching it, since a substring check passes for the wrong reason when the hostile text sits inside a quoted value.

PR shortly. Happy to keep taking review notes at this level of detail - it has made the script better both times, and I'd like to keep contributing here.

PR #3552, opened 2026-07-28
fix(dream): honor a configured 0 in synthesize + auto_think config resolution (stop coercing to the default)

brettdavies:
## Summary

**The bug.** Two dream config resolvers coerce a configured value through `parseInt(str, 10) || <default>` (and `parseFloat(str) || <default>`), which discards a legitimately-configured `0`: `parseInt("0", 10)` is `0`, and `0 || 12` evaluates to the default, so a value the operator explicitly set to `0` is silently replaced. This is the config-plane "set X, no effect" class the doctor already tracks, not a new knob. It appears in two places:

- `src/core/cycle/synthesize.ts` (`loadSynthConfig`): `dream.synthesize.cooldown_hours` (`|| 12`) and `dream.synthesize.min_chars` (`|| 2000`). `checkCooldown` treats `hours <= 0` as "no cooldown", so `cooldown_hours = 0` is the intended way to disable the cooldown, e.g. a backfill catch-up where every queued transcript should run now. Today that intent is dropped and synthesize keeps skipping `cooldown_active` for 12 hours after each run; `min_chars = 0` (accept any length) is likewise overridden back to 2000.
- `src/core/cycle/auto-think.ts` (`loadConfig`): `dream.auto_think.cooldown_days` (`|| 30`) and `dream.auto_think.budget` (`|| 2.0`). A configured `budget = 0` (spend nothing / disable) silently becomes `$2`, so the resolver can authorize spend against an explicit zero.

**The fix.** Route each value through the existing `getNumberConfig` helper (which parses via `Number(raw)` and returns a parsed `0` unchanged), wrapped in `Math.max(0, ...)` to preserve the original negative clamp:

```ts
const cooldownHours = Math.max(0, await getNumberConfig(engine, 'dream.synthesize.cooldown_hours', 12));
```

`synthesize.ts` already uses `getNumberConfig` for its subagent timeouts; `auto-think.ts` gains a local copy mirroring `synthesize.ts` / `patterns.ts`. A present-but-unparseable string still falls back to the default; an absent key still defaults; a configured `0` now resolves to `0`. Behavior is unchanged for every non-zero value. `dream.auto_think.max_per_cycle` is intentionally left alone: its `Math.max(1, ...)` floor already makes `0` invalid there, so no configured value is lost.

**One defect, split into two commits for reviewability.** The same coercion lives in two dream config resolvers and is fixed with the same shape in each. It is structured as two focused, file-scoped commits so the diff reviews cleanly file-by-file, but it is a single fix and merges as one PR:

- Commit 1: `src/core/cycle/synthesize.ts` (`cooldown_hours` + `min_chars`).
- Commit 2: `src/core/cycle/auto-think.ts` (`cooldown_days` + `budget`).

## Tests

- `test/cycle-synthesize-config-zero.test.ts` (commit 1, 6 pass): `"0" -> 0` for `cooldown_hours` and `min_chars`; absent keys keep the defaults; unparseable falls back; positive values round-trip; a negative value clamps to 0. Each loader is exercised through a stub `engine.getConfig` (no PGLite), pinned via a `__testing` export.
- `test/cycle-auto-think-config-zero.test.ts` (commit 2, 6 pass): the same matrix for `budget` and `cooldown_days` (budget accepts fractions).
- The existing `test/auto-think-phase.test.ts` suite still passes (15/15). `bun run typecheck` clean.


time-attack:

Issue #1250, opened 2026-05-21
instantiateChat native-anthropic: 404 when env injects ANTHROPIC_BASE_URL without /v1 suffix


ahnkwangwook-oss:
# `instantiateChat` native-anthropic: 404 when env injects `ANTHROPIC_BASE_URL=https://api.anthropic.com` (no /v1)

## Summary

`createAnthropic({ apiKey })` in `src/core/ai/gateway.ts` does NOT pass
`baseURL`. The Vercel AI SDK's `@ai-sdk/anthropic` provider then reads
`process.env.ANTHROPIC_BASE_URL` verbatim and POSTs to `${baseURL}/messages`.

When the host environment injects `ANTHROPIC_BASE_URL=https://api.anthropic.com`
(the bare host, no `/v1` suffix), every chat call hits
`https://api.anthropic.com/messages` and returns **HTTP 404 Not Found**.

The Anthropic API requires the `/v1` segment: `https://api.anthropic.com/v1/messages`.

## Repro

1. Set `process.env.ANTHROPIC_BASE_URL=https://api.anthropic.com` (bare host)
2. Set `process.env.ANTHROPIC_API_KEY=<valid-key>` (or ~/.gbrain/config.json)
3. Run `gbrain think "hi"`
4. Observe: `LLM_OUTPUT_NOT_JSON` warning, "no LLM available" answer
5. Add debug print in catch block → `AI_APICallError`, `status: 404`, `Not Found`

`curl https://api.anthropic.com/v1/messages` with the same key works (HTTP 200).

## Affected hosts

Claude Code (Claude.app's MCP subprocess tree) injects this env var. Possibly
others. Claude Code's own SDK normalizes internally; the Vercel SDK does not.

## Fix

`src/core/ai/gateway.ts` `instantiateChat` `case 'native-anthropic'`:

```ts
const apiKey = cfg.env.ANTHROPIC_API_KEY;
if (!apiKey) throw new AIConfigError(`Anthropic chat requires ANTHROPIC_API_KEY.`, recipe.setup_hint);
const envBaseUrl = cfg.env.ANTHROPIC_BASE_URL;
const baseURL = envBaseUrl
  ? (/\/v\d+\/?$/.test(envBaseUrl) ? envBaseUrl : `${envBaseUrl.replace(/\/$/, '')}/v1`)
  : undefined;
return createAnthropic(baseURL ? { apiKey, baseURL } : { apiKey }).languageModel(modelId);
```

[...]

PR #3553, opened 2026-07-28
fix(models): hint the /v1 base-URL suffix when a doctor chat probe 401s on an openai-compatible proxy

brettdavies:
## Summary

**The bug.** `gbrain models doctor` fires a 1-token `gateway.chat()` probe per configured surface and reports the provider's error verbatim. Against an OpenAI-shaped proxy whose base URL omits the `/v1` prefix (codex-proxy, some Azure-OpenAI mirrors, a LiteLLM proxy fronting an OpenAI-route backend), the AI SDK appends `/chat/completions` to the bare base and the proxy answers 401 Unauthorized, not 404. `classifyError` maps that to `status: 'auth'`, so the doctor points the operator at the bearer token when the real problem is the URL shape. The same 401-not-404 ambiguity that traps the runtime also traps the diagnostic built to catch it: `curl $BASE/models` returns 200 with the same bearer. The `setup_hint` documents the `/v1` convention (landed in #2627), but `models doctor` still gives no signal that the base URL, not the key, is the issue.

**The fix.** On an `auth`-classified probe failure, resolve the model's base URL and, when the model routes through an openai-compatible recipe (proxy tier, not native anthropic/openai/google) AND the base URL does not already end in `/v1`, attach a corrective `fix` line to the probe result:

```text
fix: If the API key is correct, the base URL may be missing the /v1 suffix.
     OpenAI-shaped proxies serve /v1/chat/completions and 401 on the bare path.
     Confirm with: curl <base>/models returns 200 with the same bearer, then append /v1.
```

`ProbeResult.fix` is the existing carrier (the human renderer already prints it), so nothing new is added to the output shape. The decision is a pure, unit-tested helper `openAiCompatV1Hint(modelStr, baseURL)`; the probe-side base-URL resolution reuses the same `loadConfig()` + `buildGatewayConfig()` the CLI already builds the gateway from, so env-derived base URLs like `LITELLM_BASE_URL` match the live path. The whole hint block is wrapped fail-open: any resolution error yields no hint and never breaks the probe. Native providers and base URLs already ending in `/v1` are untouched, and there is no second network call.

This is a conditional diagnostic hint, not a definitive verdict: the wording says "may be" and gives the `/models` self-check, which keeps it honest when a 401 is a genuine bad key. A follow-up could make it definitive by adding a live `GET <base>/models` cross-probe (200 there while chat 401s proves the path shape), but that adds network behavior; this change stays contained to the hint.

## Tests

- `test/models-doctor-v1-hint.test.ts` (6 pass): an openai-compat proxy base URL without `/v1` yields a hint containing `/v1`; a URL already ending in `/v1` or `/v1/` yields none; native anthropic/openai yield none; a missing base URL yields none. Pure unit tests against the real synchronous recipe registry (no network, no `mock.module`).
- The existing `test/models-doctor-embed.test.ts` and `test/models-doctor-reranker.test.ts` suites still pass (8/8). `bun run typecheck` clean.

### DIFF
--- a/src/commands/models.ts
+++ b/src/commands/models.ts
@@ -34,6 +34,7 @@ import {
   resolveModel,
   type ModelTier,
 } from '../core/model-config.ts';
+import { resolveRecipe } from '../core/ai/model-resolver.ts';
 
 const TIERS: ModelTier[] = ['utility', 'reasoning', 'deep', 'subagent'];
 
@@ -184,6 +185,44 @@ function classifyError(err: unknown): { status: ProbeStatus; message: string } {
     return { status: 'network', message: msg };
   }
   return { status: 'unknown', message: msg };
+}
+
+const OPENAI_COMPAT_V1_HINT =
+  'If the API key is correct, the base URL may be missing the /v1 suffix. ' +
+  'OpenAI-shaped proxies (codex-proxy, Azure-OpenAI mirrors, LiteLLM fronting an OpenAI route) ' +
+  'serve /v1/chat/completions and 401 on the bare path. ' +
+  'Confirm with: `curl <base>/models` returns 200 with the same bearer, then append /v1 to the base URL.';
+
+/**
+ * Fix-hint for the openai-compatible-proxy `/v1`-suffix trap.
+ *
+ * An OpenAI-shaped proxy whose base URL omits `/v1` (codex-proxy, some
+ * Azure-OpenAI mirrors, a LiteLLM proxy fronting an OpenAI-route backend)
+ * serves `/v1/chat/completions` and returns 401 on the bare `/chat/completions`
+ * the AI SDK appends to the base. `classifyError` reads that 401 as `auth` and
+ * points the operator at the bearer token, when the real fix is the URL shape.
+ *
+ * Returns the corrective hint only when the model routes through an
+ * openai-compatible recipe (proxy tier, not native anthropic/openai/google),
+ * `baseURL` is set, and `baseURL` does not already end in `/v1` (optionally with
+ * a trailing slash). Pure: recipe resolution is synchronous and does no
+ * network/engine work; any resolution failure returns undefined.
+ *
+ * @internal exported for tests.
+ */
+export function openAiCompatV1Hint(
+  modelStr: string,
+  baseURL: string | undefined | null,
+): string | undefined {
+  if (!baseURL || !baseURL.trim()) return undefined;
+  if (/\/v1\/?$/.test(baseURL.trim())) return undefined;
+  try {
+    const { recipe } = resolveRecipe(modelStr);
+    if (recipe.tier !== 'openai-compat') return undefined;
+    return OPENAI_COMPAT_V1_HINT;
+  } catch {
+    return undefined;
+  }
 }
 
 /**
@@ -523,7 +562,29 @@ async function probeModel(modelStr: string, touchpoint: 'chat' | 'expansion'): P
     }
   } catch (err) {
     const { status, message } = classifyError(err);
-    return { model: modelStr, touchpoint, status, message, elapsed_ms: Date.now() - start };
+    const result: ProbeResult = { model: modelStr, touchpoint, status, message, elapsed_ms: Date.now() - start };
+    // An openai-compatible proxy whose base URL omits `/v1` returns 401 (not
+    // 404) on the bare `/chat/completions` path, which classifyError reads as
+    // `auth`. Attach the URL-shape hint so the operator doesn't chase the
+    // bearer token. Fail open: any error resolving the base URL yields no hint
+    // and never breaks the probe.
+    if (status === 'auth') {
+      try {
+        const { loadConfig } = await import('../core/config.ts');
+        const { buildGatewayConfig } = await import('../core/ai/build-gateway-config.ts');
+        const fileCfg = loadConfig();
+        if (fileCfg) {
+          const cfg = buildGatewayConfig(fileCfg);
+          const { recipe } = resolveRecipe(modelStr);
+          const baseURL = cfg.base_urls?.[recipe.id] ?? recipe.base_url_default;
+          const hint = openAiCompatV1Hint(modelStr, baseURL);
+          if (hint) result.fix = hint;
+        }
+      } catch {
+        // fail open — no hint
+      }
+    }
+    return result;
   }
 }
 
--- a/test/models-doctor-v1-hint.test.ts
+++ b/test/models-doctor-v1-hint.test.ts
@@ -1,0 +1,40 @@
+import { describe, test, expect } from 'bun:test';
+import { openAiCompatV1Hint } from '../src/commands/models.ts';
+
+/**
+ * `gbrain models doctor` — the openai-compatible-proxy `/v1`-suffix hint.
+ *
+ * `openAiCompatV1Hint` is pure: it resolves the model's recipe synchronously
+ * (no network, no engine) to decide whether the provider is an openai-compatible
+ * proxy, then inspects the passed base URL. These cases pin the four branches
+ * without any transport stub.
+ */
+describe('openAiCompatV1Hint', () => {
+  test('openai-compat proxy without /v1 suffix returns a /v1 hint', () => {
+    const hint = openAiCompatV1Hint('litellm:gpt-4o', 'http://localhost:4000');
+    expect(hint).toBeDefined();
+    expect(hint).toContain('/v1');
+  });
+
+  test('base URL already ending in /v1 returns undefined', () => {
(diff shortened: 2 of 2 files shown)


time-attack:

The idea is sound (a 401 from an openai-shaped proxy on a bare base path is genuinely misdiagnosed as an auth failure), and the mechanics are safe: fail-open try/catch, no new output shape, no key material, no engine/JSONB/migration/source-scope/dry-run surface (2 files). But four things block it. (1) The hint's gate is "openai-compat tier AND base URL doesn't end in /v1", which fires on `zhipu`, whose OWN shipped default base URL is `https://open.bigmodel.cn/api/paas/v4` — so a genuine bad ZHIPUAI_API_KEY now tells the operator to append `/v1` to a `/v4` route. That is exactly the dead-end-hint failure mode. Verified by executing the helper against every openai-compat recipe's own default. (2) Discrimination test FAILED: I deleted the entire probeModel hint block (the whole behavior change) and all 6 of the PR's new tests still passed — the test file imports only the new pure helper, so the actual wiring has zero coverage. (3) Symptom patch: `classifyError`→`auth` is consumed at three probe sites; only `probeModel` is patched. `probeEmbeddingReachability` (models.ts:531) and `probeRerankerReachability` (models.ts:489) hit the identical URL-shape trap on the same proxies — and litellm/ollama/llama-server are documented in this repo primarily as EMBEDDING providers. (4) It re-implements half of the existing single-ownership seam `applyOpenAICompatConfig` (gateway.ts:409), dropping the `recipe.resolveOpenAICompatConfig` override — so the PR's advertised Azure-OpenAI case resolves a base URL the live gateway never uses. Plus the PR title omits the mandatory vMAJOR.MINOR.PATCH.MICRO prefix.

### Discrimination test

EXECUTED and FAILED. Stubbed master's behavior back in by replacing the whole added block in src/commands/models.ts probeModel (`if (status === 'auth') { ...loadConfig/buildGatewayConfig/resolveRecipe/result.fix... }`) with a bare `return result;`, leaving only the new exported pure helper. `bun test test/models-doctor-v1-hint.test.ts` → EXIT=0, 6 pass / 0 fail (identical to the unmodified branch: EXIT=0, 6 pass). All 6 new tests survive full removal of the behavior change, because they exercise only `openAiCompatV1Hint(modelStr, baseURL)` in isolation. Only the helper's existence (a new export, absent on master) is pinned; the fix itself is untested. models.ts was restored from backup afterward — `git status --short` clean.

### Blockers

**1.** src/commands/models.ts:216-231 (openAiCompatV1Hint) — false-positive hint on a genuine auth failure. The only gate is `recipe.tier === 'openai-compat'` plus `!/\/v1\/?$/`. `zhipu` is tier openai-compat with `base_url_default: 'https://open.bigmodel.cn/api/paas/v4'` (src/core/ai/recipes/zhipu.ts:23) and a first-class `chat` touchpoint (supports_subagent_loop: true). Failing input: chat_model=`zhipu:glm-4.6`, ZHIPUAI_API_KEY genuinely wrong → provider 401 → classifyError 'auth' → doctor prints "the base URL may be missing the /v1 suffix ... then append /v1 to the base URL". Appending /v1 to the /v4 route is a 404 dead end while the real problem (bad key) is now de-emphasized. EXECUTED: enumerated all 19 openai-compat recipes against their OWN shipped defaults; `zhipu -> https://open.bigmodel.cn/api/paas/v4` and `litellm -> http://localhost:4000` are the two that get the hint. litellm is the intended case; zhipu is a straight false positive. Suggested gate: skip when `baseURL === recipe.base_url_default` (a recipe's own default is correct by construction) unless the recipe opts in, or require an operator-supplied URL (`cfg.base_urls[recipe.id]` present) / a per-recipe `v1_suffix_trap` flag. Note this generalizes: any future non-/v1 openai-compat route (/v1beta, /compatible-mode/v2) inherits the wrong advice.

**2.** test/models-doctor-v1-hint.test.ts:1-40 — the new tests prove nothing about the fix. DISCRIMINATION TEST EXECUTED: I replaced the entire `if (status === 'auth') { ... }` block in probeModel (src/commands/models.ts:571-586) with a bare `return result;` (i.e. master's behavior) and re-ran `bun test test/models-doctor-v1-hint.test.ts` → EXIT=0, 6 pass / 0 fail. The file imports only the pure helper `openAiCompatV1Hint`; the three load-bearing pieces of the actual change — the `status === 'auth'` gate, the loadConfig/buildGatewayConfig base-URL resolution, and `result.fix = hint` — have zero coverage. Nothing in this PR would catch a regression that silently stops attaching the hint.

**3.** src/commands/models.ts:571 — symptom patch; two sibling callers of the same root cause left broken. `classifyError` is consumed at three probe sites: probeRerankerReachability (line 489/490), probeEmbeddingReachability (line 531/532), probeModel (line 564, the only one patched). Failing input: embedding_model=`litellm:text-embedding-3-large` with LITELLM_BASE_URL=`http://localhost:4000` (bare) and the proxy serving only /v1/embeddings → 401 → `embedding_reachability` probe reports `auth` with NO hint, the exact bug the PR exists to fix. This matters more than the chat path here: docs/integrations/embedding-providers.md documents litellm/ollama/llama-server primarily as embedding providers, and litellm's own setup_hint is embedding-focused. Root-cause fix is one hint helper applied at the shared classifyError consumption point — a smaller diff than the chat-only patch.

**4.** src/commands/models.ts:579 — duplicates half of the existing single-ownership seam. `applyOpenAICompatConfig(recipe, cfg)` (src/core/ai/gateway.ts:409, already exported and used by test/ai/recipe-azure-openai.test.ts and test/ai/deepseek-reasoning-content.test.ts) owns BOTH `cfg.base_urls?.[recipe.id] ?? recipe.base_url_default` AND the `recipe.resolveOpenAICompatConfig` override. The PR re-implements only the first half. Failing input: azure-openai (tier openai-compat, no base_url_default, live URL templated to `${AZURE_OPENAI_ENDPOINT}/openai/deployments/${AZURE_OPENAI_DEPLOYMENT}` at recipes/azure-openai.ts:124-141) → the PR reads `base_urls['azure-openai']`, a value the live gateway IGNORES. Result: the PR's explicitly-advertised "Azure-OpenAI mirrors" case yields no hint at all; and if an operator sets provider_base_urls['azure-openai'], the hint reasons about a URL that is never used. Call applyOpenAICompatConfig — it's already inside the fail-open try/catch, and it throws AIConfigError on a missing URL which the catch swallows.

### Not verified

- Full unit suite, `bun run verify`, `bun run ci:local`, and e2e — deliberately skipped for execution footprint per instructions. Ran only the PR's own test file, the two adjacent doctor suites, and typecheck.
- No live network reproduction: I did not stand up a bare-path openai-shaped proxy to confirm it really answers 401 (not 404 or 405) on POST /chat/completions. The PR's premise is plausible and matches the pre-existing #1250 native-base-URL fix rationale in gateway.ts:427-444, but the 401 claim itself is unverified here.
- Whether an operator in practice sets provider_base_urls['azure-openai'] (the path where the duplicated resolution produces a hint about an unused URL) — code-verified as reachable, not observed in the wild.
- Did not run `gbrain models doctor` end to end against a real misconfigured proxy, so the rendered single-line ~380-char `fix:` output was read from the renderer (models.ts:711), not observed. Note the PR body shows the hint pre-wrapped over 3 lines; the code emits one unwrapped line.

- `gh pr view 3553`
- `gh pr diff 3553 > /tmp/pr3553.diff`
- `gh pr checkout 3553 (branch fix/models-doctor-v1-path-probe, head 7d2e237d)`
- `ln -s <main-workspace>/node_modules <worktree>/node_modules  # worktree had no deps installed`
- `bun test test/models-doctor-v1-hint.test.ts  -> EXIT=0, 6 pass`
- `bun test test/zz-scratch-fp.test.ts  # scratch: enumerate every openai-compat recipe's own base_url_default through openAiCompatV1Hint -> EXIT=1, HINT fires on litellm + zhipu (scratch file deleted, tree clean)`
- `# discrimination: replaced probeModel's auth-hint block with `return result;` then re-ran`
- `bun test test/models-doctor-v1-hint.test.ts  -> EXIT=0, 6 pass (old behavior, tests still green)`
- `cp /tmp/models.bak.ts src/commands/models.ts  # restore; git status --short clean`
- `bun run typecheck  -> EXIT=0 (tsc --noEmit clean)`
- `bun test test/models-doctor-embed.test.ts test/models-doctor-reranker.test.ts  -> EXIT=0, 8 pass`
- `grep -n 'tier:' src/core/ai/recipes/*.ts ; grep -rn 'base_url_default' src/core/ai/recipes/*.ts`
- `grep -rn 'resolveOpenAICompatConfig|applyOpenAICompatConfig' src/ test/`
- `read src/core/ai/build-gateway-config.ts, src/core/config.ts:551 (loadConfig), src/commands/models.ts:444-590 + 660-720`

</details>

---

time-attack:

Issue #1801, opened 2026-06-03
Supervisor never restarts an alive-but-wedged worker: dead worker DB pool + no-op no_recent_completions warn = silent 15h processing halt


garrytan-agents:
**Version:** gbrain 0.42.10.0 (Postgres engine, transaction-mode pooler port 6543, `prepare:false`; minion `jobs supervisor` + 1 child `jobs work`, concurrency 3, `--max-rss 16384`)
**Severity:** High — brain processing silently halted for ~15h; no crash, no alert, no self-heal.

## TL;DR

A child worker's DB pool died (`write CONNECTION_ENDED` on every background-interval tick) and **never recovered**, but the worker process **stayed alive the entire time**. Because the supervisor's restart logic only fires on child *exit*, an alive-but-wedged worker is invisible to it. The supervisor's own health check (separate connection) kept succeeding, so it emitted `health_warn reason=no_recent_completions` **once a minute for 914 consecutive minutes** — and took **zero corrective action**. Jobs piled to 57 waiting / 0 active, and every long-running `autopilot-cycle` dead-lettered with `max stalled count exceeded`. Manual fix: kill the supervisor tree so the respawn path rebuilds a fresh worker.

This is **distinct from** the existing connection bugs:
- #1720 / #1745 = crash-**loop**: worker exits `code=1` and respawns repeatedly (in-process reconnect missing).
- #1678 = RSS watchdog SIGTERM-kills the worker mid-cycle.
- **This one = no crash at all.** The worker never exits, so no respawn path is ever entered. The dead pool just sits there.

## Observed (sanitized)

`/tmp/gbrain-worker.log`, the same lines every ~minute for 15h:

```
[supervisor 05:01:11] health_warn reason=no_recent_completions waiting_count=57 minutes_since_completion=913 queue=default
Stall detection error: write CONNECTION_ENDED aws-1-...pooler.supabase.com:6543
Timeout detection error: write CONNECTION_ENDED aws-1-...pooler.supabase.com:6543
Wall-clock timeout detection error: write CONNECTION_ENDED aws-1-...pooler.supabase.com:6543
```

[...]

PR #3555, opened 2026-07-28
fix(minions): reconnect the stall timer, stop failJob masking job errors, unify retry matchers (#1720)

### LINKED ISSUE #1720: autopilot worker crash-loops on pooler CONNECTION_CLOSED: background queue loops have no in-process reconnect

time-attack:
Fixes the remaining gaps in #1720. Earlier PRs already covered part of the report: #2025 gave `promoteDelayed`/`claim` in-process reconnect (#1491) and #1824 added the supervised db-liveness probe + `db_dead` exit (#1801). What still crash-looped noisily:

## What changed

**1. Stall-detection interval reconnects (`src/core/minions/worker.ts`)**
The stall/timeout/wall-clock sweep interval was the last background loop with no reconnect — a dead pool sprayed `Stall detection error: write CONNECTION_CLOSED ...` every tick until the ~3-minute `db_dead` exit (the reporter's 20k-line log spray). On a retryable connection error it now routes through the same `reconnectAfterConnectionError` used by the #1491 fix, **at most once per tick** shared across the three sweeps, so a fully-dead pool costs one rebuild attempt per `stalledInterval`, not three.

**2. `failJob` no longer masks the original job error (`worker.ts`)**
When recording a failure threw on the same dead pool, the rejection escaped to `launchJob`'s catch and the ORIGINAL job error was never logged anywhere — the reporter's debuggability complaint. Now: the original error is logged FIRST (it survives no matter what), then the pool reconnects and the recording is retried once. If the retry also fails, the row is left for the stall detector (the established D8a recovery path — the lock has stopped renewing, so `handleStalled` requeues it cleanly on a live pool). Non-retryable recording errors still propagate.

**3. One canonical retry matcher (`src/core/db.ts`, `src/core/retry-matcher.ts`)**
The startup connect matcher (`isRetryableDbConnectError`, a private 5-pattern list) had drifted from the runtime matcher (`retry-matcher.ts`) — it lacked `connection.*closed` / `CONNECTION_CLOSED`, so a pooler close hitting `connectWithRetry` was treated as permanent. The private list is deleted; `isRetryableDbConnectError` is now an alias of `isRetryableConnError` (a strict superset of the old five patterns — existing negative tests in `test/minions.test.ts` still pass). A drift-guard test pins agreement over a corpus of error shapes so the lists can't silently diverge again. `retry-matcher.ts` also gains a `code === 'CONNECTION_CLOSED'` check (mirror of the existing `CONNECTION_ENDED` check) for wrappers that rethrow with the code but a rewritten message.

**Deliberately NOT included — reporter's proposal 3.** Repairing a dead-but-non-null module singleton inside `db.connect()` conflicts with the #1471 create-vs-join contract: the create decision is deliberately atomic (no `await` between the null-check and assignment), and a liveness probe requires an `await`; teardown of a shared singleton is owner-only, so a borrower-initiated repair violates ownership; and a per-call `SELECT 1` probe adds connection pressure to every engine join. `engine.reconnect()` is the established repair path and every worker background loop now routes through it. That redesign needs its own review.

## Evidence

`test/worker-conn-resilience-1720.test.ts` — 5 tests that inject the postgres.js error shapes (`write CONNECTION_CLOSED host:port`, `code: 'CONNECTION_CLOSED'`). Discrimination proof (same test file, `src/` stashed to pristine master `6136e139972a`):

- unmodified master: **0 pass / 5 fail**, all at the behavioral assertions (reconnects=0, failJob never retried, matchers disagree on `CONNECTION_CLOSED`)
- with the fix: **5 pass / 0 fail**

Gates: `bun run typecheck` clean; `bun run verify` 32/32 green; all test files touching the worker/queue/matcher pass (327 tests across 10 files: worker-promote-reconnect, worker-supervised-db-probe, minions, worker-lock-renewal{,-shape}, minions-lease-full-retry, minions-quiet-hours, retry-matcher, postgres-engine-getter-selfheal, core/retry).

UNVERIFIED-ON-ENVIRONMENT: a real Supavisor session-pooler socket close was not reproduced — the tests inject the documented error shapes. The reconnect mechanism itself is the one already proven in production by #2025.

## Connection-pressure reasoning

Worst case (pool permanently dead) this adds: ≤1 reconnect attempt per stall tick (default 30s) + ≤1 per failed job's recording. `PostgresEngine.reconnect()` is single-flighted (`_reconnecting` latch) and build-then-swap, and the supervised probe still exits `db_dead` after ~3 minutes, bounding the window. That is strictly less pooler pressure than the status quo (a crash-looping worker respawning 547 times, each respawn opening a fresh pool). The matcher unification only widens *startup* retry classification (3 attempts, exponential backoff, unchanged).

Fixes #1720.


time-attack:

Issue #3554, opened 2026-07-28
resetGateway() leaves the gateway unconfigured, wiping the 1536 test preload — 93 files can poison 122 others depending on shard bin-packing


time-attack:
Found while investigating a CI failure on #3545 that was **not caused by that PR**. This is a latent, shard-composition-dependent test-isolation bug, and it is almost certainly the source of several failures dismissed as "environmental" over time.

## Mechanism

`bunfig.toml` preloads `test/helpers/legacy-embedding-preload.ts`, which pins the gateway ONCE at process start:

```
embedding_model: 'openai:text-embedding-3-large'
embedding_dimensions: 1536
```

`resetGateway()` (`src/core/ai/gateway.ts`) sets `_config = null`. It does not restore the preload's pin — it leaves the gateway **unconfigured**:

```
BEFORE: openai:text-embedding-3-large 1536
AFTER THREW: AI gateway is not configured. Call configureGateway() during engine connect.
```

So the next file that connects an engine reconfigures from the SHIPPED default (`zeroentropyai:zembed-1` @1280, `src/core/ai/defaults.ts`), not 1536. A PGLite store created at 1536 then rejects the vectors:

```
error: expected 1280 dimensions, not 1536
routine: "CheckExpectedDim"
```

## Blast radius

- **93 test files call `resetGateway()`** (usually in `afterAll` — correct hygiene for their own purposes).
- **122 test files reference 1536.**

Whether a resetting file lands before a 1536-dependent file in the same shard process is decided by weight-based LPT bin-packing (`scripts/sharding.ts` + `scripts/test-weights.json`). So **adding or renaming any test file can reshuffle which pairs collide** — the failure appears to move around at random and looks like flakiness.

## Why it is invisible locally

Reproduced on #3545: three `eval-trajectory` tests failed in CI shard 9, twice in a row. The same file passes 7/7 locally in isolation, and 44/44 when run with four of its shard siblings. The trigger was that #3545 adds one new test file (`test/docs-cli-commands.test.ts`), which re-binned the shards and placed `eval-trajectory` after `test/extract-takes-model-resolution.test.ts` — a file whose `afterAll` calls `resetGateway()`.

Execution order in that shard, from the CI log:

```
test/extract-takes-model-resolution.test.ts   <- afterAll: resetGateway()
... 15 files ...
test/eval-trajectory.test.ts                  <- creates a 1536 store, gets 1280
```

## Suggested fix

[...]

PR #3545, opened 2026-07-28
fix(docs): remove references to the nonexistent gbrain install command (#3502)

time-attack:
Fixes #3502.

## The reported bug

`docs/tutorials/personal-brain.md` Step 6 instructed `gbrain install`, which fails with `Unknown command: install` (the managed-install model was retired in v0.36.0.0). Step 6 now documents the current flow — `gbrain init --supabase` in the **brain repo**, `gbrain skillpack scaffold --all` in the **agent workspace** — states which repo each command runs in, how it feeds the Step 7 Supabase setup, and the `--pglite` / `migrate --to supabase` alternative.

## The full sweep

Per the issue's class, every `gbrain <verb>` invocation in README/docs/skills was checked against the live CLI surface (`CLI_ONLY` + operation `cliHints` names + aliases). Dead references found and fixed:

| dead | live replacement | sites |
|---|---|---|
| `gbrain install` | `init` + `skillpack scaffold` | personal-brain.md, MARKDOWN_SKILLS_AS_RECIPES.md |
| `gbrain put-page` / `put_page` / `get_page` | `put` / `get` | multi-source-brains.md, book-mirror, perplexity-research, eiirp |
| `gbrain add_link` | `link` | 5 guides |
| `gbrain add_timeline_entry --entry` | `timeline-add <slug> <date> <summary>` (the `--entry` flag never existed) | 6 guides |
| `gbrain get_timeline` / `list_pages` | `timeline` / `list` | operational-disciplines.md |
| `gbrain get_links` / `put_raw_data` / `get_raw_data` (MCP-only ops, no CLI name) | `gbrain call <op> '<json>'` | 6 guides |
| `gbrain find_trajectory` | `find-trajectory` | 2 tutorials |
| `gbrain file upload` | `gbrain files upload` | mcp/DEPLOY.md |
| `gbrain pages restore` | `gbrain restore` | schema-unify, 2 architecture docs |
| `gbrain rebuild --confirm-destructive` | never shipped — replaced with the real recovery sequence (`psql` derived-table wipe + `sync` + `extract all`; `reinit-pglite` on PGLite) | system-of-record.md |
| `gbrain extract-facts` | the `extract_facts` cycle (op, no CLI name) | what-schemas-unlock.md |
| `gbrain transcription`, `gbrain research init`, `gbrain cron`, `gbrain plugin list` | stale forward-notes reworded to current truth | voice-note-ingest, data-research, cron-via-minions, plugin-authors |

[...]

PR #3556, opened 2026-07-28
fix(engine): exclude soft-deleted pages from getHealth counts (#1305)

### LINKED ISSUE #1305: getHealth()/brain_score counts soft-deleted pages while getStats() excludes them
## Summary

`getHealth()` (which computes `brain_score`) counts soft-deleted pages in its denominator and graph subqueries, while `getStats()` explicitly excludes them. So `gbrain delete` (soft delete) shrinks `gbrain stats` page_count but does **not** move `brain_score` until the autopilot purge TTL hard-removes the rows — surprising when using soft-delete to clean up a brain.

## Location — `src/core/postgres-engine.ts`

`getStats()` (~line 3625) filters correctly:
```sql
(SELECT count(*) FROM pages WHERE deleted_at IS NULL) as page_count
```
with the comment *"exclude soft-deleted from page_count. Same posture as the search filter and getPage default — soft-deleted is hidden everywhere the user looks."*

`getHealth()` (~line 3660+) does **not** filter:
```sql
(SELECT count(*) FROM pages) as page_count,
...
(SELECT count(*) FROM pages p
 WHERE NOT EXISTS (SELECT 1 FROM links l WHERE l.to_page_id = p.id)
   AND NOT EXISTS (SELECT 1 FROM links l WHERE l.from_page_id = p.id)) as orphan_pages,
(SELECT count(DISTINCT page_id) FROM timeline_entries) as pages_with_timeline,
```
and the `entity_pages` CTE (`SELECT id, slug FROM pages WHERE type IN ('person','company')`) likewise includes soft-deleted pages.

Since `brain_score` components are coverage ratios (`min(numerator/page_count, 1) * weight`), soft-deleted pages inflate `page_count` / `orphan_pages` and depress every graph component (links / timeline / orphans).

## Suggested fix

Add `deleted_at IS NULL` to `getHealth()`'s `page_count`, `orphan_pages`, `pages_with_timeline` (via a join to `pages`), and the `entity_pages` CTE — matching `getStats()`'s posture. Leave the `content_chunks`-based `embed_coverage` as-is (chunks intentionally stay raw until the autopilot purge per the `getStats` comment).

## Impact / repro

On a brain where ~800 stale pages were soft-deleted, `brain_score` stayed flat while `gbrain stats` page_count had already dropped; a hard `DELETE FROM pages` was required to move the score. Observed on v0.37.9.0.

time-attack:
Fixes #1305.

## Confirmed reproduction (unmodified master, PGLite)

50 pages → soft-delete 40:

```json
{
  "stats_page_count": 10,
  "health_page_count": 50,
  "orphan_pages": 50,
  "brain_score_before_delete": 10,
  "brain_score_after_delete": 10
}
```

`getStats()` filters `deleted_at IS NULL` (v0.26.5); `getHealth()` never did, so `brain_score` is inert to soft-deletes. PR #2235 fixed `pages_by_type` and `sources list/status` but explicitly did not close this.

## Fix

Every **page-scoped** count in `getHealth` now excludes soft-deleted rows, with identical SQL shapes in both engines:

- `page_count` subquery → `WHERE deleted_at IS NULL`
- `entity_pages` CTE (feeds `link_coverage` / `timeline_coverage` denominators) → `AND deleted_at IS NULL`
- `most_connected` top-5 → `AND p.deleted_at IS NULL`
- the islanded/timeline per-page scan (feeds `orphan_pages`, `no_orphans_score`, `timeline_coverage_score`) → `WHERE p.deleted_at IS NULL`

## The boundary decision (which counts still include soft-deleted rows)

- **Chunk/link storage counts stay raw**: `embed_coverage`, `missing_embeddings`, `link_count`, `dead_links`. This matches `getStats`' documented posture — chunks/links occupy storage until the autopilot purge phase runs. `dead_links` semantics unchanged: a link to a soft-deleted page is not "dead" (the row still exists; purge is what kills it).
- **Destructive-removal counts untouched** — those deliberately count all rows (#2235 kept that on purpose).
- `stale_pages` was already filtered via `buildStalePagesWhere` (`deleted_at IS NULL` is its first condition) — no change needed.

## Discrimination proof

`test/health-soft-delete.test.ts`, run against **unmodified origin/master** (`6136e139`):

```
3 fail / 1 pass
  page_count:     Expected 4,  Received 10
  orphan_pages:   Expected 0,  Received 8
  link_coverage:  Expected 1,  Received 0.5
```

(the 1 pass is the boundary test — raw chunk counting is unchanged behavior)

With the fix: `4 pass / 0 fail`. All failures are behavioral assertions, not import/export shape.

## Verification

- `bun run typecheck` — clean
- `bun run verify` — 32 pass / 0 fail, all checks green
- getHealth consumers swept: `brain-score-breakdown`, `pages-soft-delete`, `advisor-core`, `advisor-op-gate`, `brain-score-recommendations`, `doctor-brain-checks-score`, `doctor-timeline-metric-labels-2298`, `features`, `pglite-engine` — 216 tests, 0 fail

No VERSION/CHANGELOG changes — fix-only PR.

### DIFF
--- a/src/core/pglite-engine.ts
+++ b/src/core/pglite-engine.ts
@@ -5332,12 +5332,16 @@ export class PGLiteEngine implements BrainEngine {
     // pages_with_timeline) and v0.10.3 graph layer (link_coverage, timeline_coverage,
     // most_connected). Both coexist: master's brain_score is the composite
     // dashboard, v0.10.3 metrics give entity-page-level granularity.
+    // #1305: every page-scoped count here excludes soft-deleted rows — same
+    // posture as getStats — so brain_score moves when the user deletes pages.
+    // Chunk/link counts stay raw (storage until the purge phase), matching
+    // getStats, and destructive-removal counts elsewhere deliberately stay raw.
     const { rows: [h] } = await this.db.query(`
       WITH entity_pages AS (
-        SELECT id, slug FROM pages WHERE type IN ('entity', 'person', 'company')
+        SELECT id, slug FROM pages WHERE type IN ('entity', 'person', 'company') AND deleted_at IS NULL
       )
       SELECT
-        (SELECT count(*) FROM pages) as page_count,
+        (SELECT count(*) FROM pages WHERE deleted_at IS NULL) as page_count,
         (SELECT count(*) FROM content_chunks WHERE embedded_at IS NOT NULL)::float /
           GREATEST((SELECT count(*) FROM content_chunks), 1)::float as embed_coverage,
         0 as stale_pages,
@@ -5362,7 +5366,7 @@ export class PGLiteEngine implements BrainEngine {
       SELECT p.slug,
              (SELECT count(*) FROM links l WHERE l.from_page_id = p.id OR l.to_page_id = p.id)::int as link_count
       FROM pages p
-      WHERE p.type IN ('entity', 'person', 'company')
+      WHERE p.type IN ('entity', 'person', 'company') AND p.deleted_at IS NULL
       ORDER BY link_count DESC
       LIMIT 5
     `);
@@ -5381,6 +5385,7 @@ export class PGLiteEngine implements BrainEngine {
               AND NOT EXISTS (SELECT 1 FROM links l WHERE l.from_page_id = p.id)) as islanded,
              EXISTS (SELECT 1 FROM timeline_entries te WHERE te.page_id = p.id) as has_timeline
       FROM pages p
+      WHERE p.deleted_at IS NULL
     `);
 
     const r = h as Record<string, unknown>;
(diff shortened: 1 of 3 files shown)


time-attack:
The `test (9)` failure is **#3554, not this PR**. Same three `eval-trajectory` tests, same signature (`expected 1280 dimensions, not 1536`, `routine: "CheckExpectedDim"`) — three occurrences in the shard log.

Mechanism: `resetGateway()` leaves the gateway unconfigured, wiping the 1536 pin from `test/helpers/legacy-embedding-preload.ts`; the next file to connect an engine reconfigures from the shipped default (`zembed-1` @1280) and a 1536 store rejects it. 93 files call `resetGateway()`, 122 reference 1536, and which pairs collide is decided by weight-based shard bin-packing — so **adding any test file reshuffles the mines**. This PR adds one.

Fix is up as #3557 (verified: the ordered pair `extract-takes-model-resolution` → `eval-trajectory` fails 3 on master, passes 9/9 on that branch). Holding this PR until #3557 lands, then rebasing — it should go green untouched. Nothing to change here.

This is the second PR today caught by that mine (#3545 was the first), which is why the merge train is paused rather than pushing through it.

PR #3558, opened 2026-07-28
v0.42.68.1 fix(reindex-frontmatter): reuse the connected engine instead of self-deadlocking on the PGLite lock (#1963)

### LINKED ISSUE #1963: reindex-frontmatter always times out on PGLite after upgrade to 0.42.32.0
After upgrading gbrain from 0.42.26.0 to 0.42.32.0, `gbrain reindex-frontmatter` consistently fails with: "GBrain: Timed out waiting for PGLite lock."

Steps to reproduce:
1. `gbrain init --pglite --no-embedding`
2. `gbrain import <dir> --no-embed`
3. `gbrain config set search.mode conservative`
4. `gbrain reindex-frontmatter`

Expected: frontmatter reindex completes.
Actual: always times out on PGLite lock.

Other write commands work fine (`gbrain config set`, `gbrain list`, `gbrain extract links`). Only `reindex-frontmatter` blocks.
Environment: PGLite, no embeddings, 76 pages, schema migrated successfully via upgrade.

time-attack:
## Summary

`gbrain reindex-frontmatter` was 100% broken on PGLite: it always exited 1 after exactly 30s with `Timed out waiting for PGLite data-dir lock. Process <pid> has held it since ... (command: ... reindex-frontmatter ...)` — where the named lock holder is **the command's own process**.

The issue's `statement_timeout` theory is a red herring (the backfill itself completes in milliseconds). The real cause is a **same-process double-connect**:

1. `src/cli.ts` dispatch runs `connectEngine()` for every engine-bound command — on PGLite this takes the data-dir lock.
2. `reindexFrontmatterCli` then built and connected a **second** engine on the same data dir.
3. `acquireLock` never reaps a live PID (by design, #2348), and the holder is this very process — so the second connect spins the full 30s timeout and throws.

## What changed

- `reindexFrontmatterCli(engine, args)` and `runBackfillCommand(engine, args)` now take the **already-connected engine** from cli.ts's dispatch switch instead of building their own. cli.ts's shared `finally` (`finishCliTeardown`) owns teardown.
- **`gbrain backfill <kind>` had the identical double-connect** (it is the documented canonical entrypoint that `reindex-frontmatter` aliases) and is fixed the same way. A sweep of `createEngine` users in `src/commands/` found no other command dispatched from the engine-connected switch that self-connects — the rest (init, upgrade, apply-migrations, frontmatter, integrity, lint, cache, schema, migrations) dispatch pre-engine or intentionally manage a second engine (engine transfer).
- `docs/architecture/KEY_FILES.md` entries updated to the new truth, including an explicit note that the fix is engine reuse at the dispatch layer, NOT lock reentrancy.

## Deliberately not changed

`src/core/pglite-lock.ts` is untouched. A same-PID special case (fail-fast or reap) was prototyped and rejected: `test/pglite-lock.test.ts` pins the #2348 corruption-fix semantics — a live-PID holder is NEVER stolen, PID-reuse included, and those tests deliberately simulate live holders with the test process's own PID. Softening the lock fights that pinned invariant; the root cause is the second connect.

## Verification

Regression test `test/reindex-frontmatter-pglite-spawn.serial.test.ts` spawns the real CLI against a fresh PGLite home (`init --migrate-only` → `import` → `reindex-frontmatter --yes --json` → `backfill effective_date --dry-run`).

- **Unmodified master:** test FAILS behaviorally — `reindex-frontmatter` exits 1 after ~30s with `Timed out waiting for PGLite data-dir lock. Process 37716 has held it since ... (command: .../src/cli.ts reindex-frontmatter --yes --json)` (the process blocking it is itself).
- **With this fix:** the same unchanged test passes in ~5s; both commands exit 0.
- `bun run typecheck` clean; `bun run verify` 32/32 green; neighboring suites (`pglite-lock`, `backfill-base`, `backfill-concurrency-clamp`, `cycle-pglite-lock-ordering`, `reindex-frontmatter-connect`) all pass (42/42).

No engine files touched (parity n/a), no JSONB changes, no VERSION/CHANGELOG edits.

Fixes #1963


time-attack:
## Adversarial review — `MERGE`, and now actually mergeable

This is the **only PR of 32** in this review round to survive two independent refuters. It was the one with a real bug, a root-cause fix, and a test that fails when the fix is removed:

> `gbrain reindex-frontmatter` and `gbrain backfill <kind>` were 100% dead on PGLite — `cli.ts:1803` takes the data-dir lock, the command modules built a second engine on the same dir, `acquireLock` never reaps a live PID → 30s timeout, exit 1. Reproduced on the parent commit at 33.2s, with the error naming the process itself as the holder; passes in 5.0s post-fix.

The sibling census was done properly too: every other `createEngine` caller in `src/commands/` dispatches before `connectEngine`, so these were the only two, and `pglite-lock.ts` was correctly left alone. Softening the lock would have been the wrong fix and this didn't do it.

### What I changed before it could land

Review also found it **did not merge** — `mergeable: CONFLICTING`. The green CI rollup was from 2026-07-28 against a stale base, so it proved nothing about the merged tree. Three other PRs in this batch had the same problem and it was caught only because a reviewer thought to ask `gh`; no checklist includes it.

- **Merged master.** One conflict, in `docs/architecture/KEY_FILES.md`, and it was semantic rather than textual: master added the engine dynamic-import clauses to the `pglite-engine` and `postgres-engine` entries, while this branch added the #1963 same-process-reentrancy note to the `pglite-lock` entry. Both are current-state truth, so I took master's two engine entries and this branch's lock entry rather than picking a side. A directional `--ours`/`--theirs` here would have silently dropped one of them.
- **Bumped the version.** `VERSION` equalled master's `0.42.67.0`, which the CI version-gate rejects outright — it requires strictly greater. Now `0.42.68.1` with a matching CHANGELOG entry (`0.42.68.0` is claimed by two other open PRs).
- **Retitled** to lead with the version per the IRON RULE, since the merge commit inherits it.

Merged tree: typecheck exit 0, `pglite-lock` + `reindex-frontmatter` 16 pass, llms bundle fresh. Now `MERGEABLE`/`CLEAN`.

### One thing a maintainer should know before clicking merge

**The Postgres path was never executed** (no Docker in the review environment). The bug is PGLite-only, so that is where all the verification went — but the *change* is ownership of connection teardown, and Postgres is where teardown has consequences: an un-ended pool, an advisory lock or `gbrain_cycle_locks` row not released, a process that doesn't exit. The reasoning that `createEngine` without `poolSize` wraps the `db.ts` module singleton (so reuse is behaviour-identical) is sound and probably right, but it is inference, not a run.

Related gap of the same shape: `reindex-frontmatter`'s interactive-cancel path calls `process.exit(1)`, which runs no `finally` at all.

Worth one run against real Postgres before merge, or an explicit decision to accept the inference.

PR #3559, opened 2026-07-28
fix(autopilot): stop the drain-worker self-deadlocking at concurrency=1 (#2050)

### LINKED ISSUE #2050: autopilot: drain-worker runs at concurrency=1, self-deadlocking any cycle phase that spawns a subagent (patterns, synthesize)

time-attack:
Fixes #2050.

## Confirmed evidence

- `src/commands/autopilot.ts:480` spawns the drain worker as `['jobs', 'work', '--max-rss', N]` — no `--concurrency`; `resolveWorkerConcurrency` (`src/commands/jobs.ts:126`) resolves `flag ?? env ?? '1'` → 1 with an empty env; `worker.ts:524` only claims while `inFlight.size < concurrency`.
- The existing mitigation is PGLite-only: `runPgliteSubagentsInline` early-returned unless `engine.kind === 'pglite'`, and the child queue was `'default'` on Postgres (`patterns.ts:192`, `synthesize.ts:552`). PR #2128 listed the Postgres side as a deferred follow-up.
- So on Postgres (the environment the issue reports): the `patterns`/`synthesize` parent job holds the single worker slot, submits its `subagent` child to `'default'`, and blocks in `waitForCompletion` — the child is never claimed. The phase burns the full `subagent_wait_timeout_ms` window (35 min default), cancels its own child, and repeats every cycle.

## Why not just raise the concurrency number

No fixed number is structurally safe. Autopilot fans out up to `fanoutMax` (default 4) `autopilot-cycle` parents per tick, all submitted **before any child exists** — so with N slots and ≥N queued parents, every slot fills with a blocking parent and the deadlock reappears (claim order is `priority ASC, created_at ASC`, and parents always predate their children). The `resolveEffectiveFanoutMax` clamp only fires for a live *supervisor* lock holder, which autopilot's bare `jobs work` spawn is not. Bumping the number would also spend real DB-pool slots — exactly the pressure the pace-mode subsystem (`db-pacer.ts` / `pace-mode.ts`) exists to avoid, because pool saturation starves lock renewals (`lock-renewal-failed` → dead jobs).

## Fix — the child must not need a worker slot at all

The smallest structural change: extend the already-shipped inline drain to both engines.

1. **`runPgliteSubagentsInline` → `runSubagentsInline`** (`src/core/cycle/synthesize.ts`): the kind gate is removed; the claim → run → complete/fail drain now runs on Postgres too, from the slot the parent already holds.
2. **Private per-run child queue on both engines** (`patterns.ts`, `synthesize.ts`; previously `'default'` on Postgres): the inline drain can never claim unrelated `'default'` jobs, and a `'default'`-queue worker can never race the parent for its own child.
3. **Child lock heartbeat** (new, required for Postgres correctness): the drain renews the child's 30s claim lock at `lockMs/3` (worker cadence parity) while the handler runs. Without it, a concurrent worker's `handleStalled()` sweep — which scans **all** queues, not just its own — would requeue any >30s child mid-run and stall-churn it to dead. A `false` renew (row cancelled/reclaimed) aborts the handler; renew errors are swallowed (best-effort, never an unhandledRejection).

## Pool / lock-renewal safety

- **Zero added concurrency**: the child's work replaces the parent's idle 5s `waitForCompletion` polling *in the same worker slot*. No new claim slots, no new engine pools, no change to how many jobs run at once — so this cannot add PgBouncer pressure or starve the supervisor's/worker's lock renewals.
- The heartbeat is one tiny `UPDATE` per 10s on the direct session pool — the same channel (`executeRawDirect`) the worker's own renewals already use, chosen precisely because the transaction pooler can't reap it mid-hold.
- The spawn args are deliberately untouched: adding an explicit `--concurrency` flag would out-rank `GBRAIN_WORKER_CONCURRENCY` and break that operator escape hatch (env still flows to the child via the cloned env). Raising worker concurrency remains a pure throughput knob, no longer a correctness one; the doctor `autopilot_fanout_concurrency` advisory stays valid for throughput.
- Throughput note: children now run serially within their parent's slot. Under the doctor-recommended config (fanout = concurrency − 1) that is equal-or-better than before — all parents previously shared a single free slot for all their children; now each parent has its own lane.

## Discrimination proof

`test/autopilot-drain-deadlock-2050.test.ts` runs the **real** `runPhasePatterns` and the real drain against a PGLite engine masked as `kind='postgres'` (only the `kind` property is intercepted — the exact branch production Postgres takes; no `.kind` reads remain anywhere else in the drain path).

Against **unmodified origin/master** (`6136e139`) — behavioral failures, not import shape:

```
✗ child is drained inline on a private queue instead of waiting out the parent
    Expected: "dead"      Received: "timeout"     (self-deadlock burned the wait window)
✗ inline drain runs children on a non-pglite engine and heartbeats their lock
    Expected: "completed" Received: "waiting"     (child never claimed)
```

With the fix: 2 pass / 0 fail. The second test also pins the heartbeat: the stub child outlives a shortened claim lock and runs the same `handleStalled()` sweep a concurrent worker fires — it must not touch the running child.

## Verification

- `bun run typecheck` — clean
- `bun run verify` — 32 pass / 0 fail, all checks green
- `test/cycle-patterns-child-outcome.test.ts` (2), `test/e2e/dream-synthesize-pglite.test.ts` (15), `test/cycle-patterns.test.ts` — all pass
- `test/cycle-enabled-phase-completeness.test.ts` has 1 failure in my environment that reproduces identically on unmodified master (embed-provider gating, environmental) — not from this change

No VERSION/CHANGELOG changes — fix-only PR.

### DIFF
--- a/src/core/cycle/patterns.ts
+++ b/src/core/cycle/patterns.ts
@@ -30,14 +30,13 @@ import { serializeMarkdown } from '../markdown.ts';
 import type { Page, PageType } from '../types.ts';
 // #2415: allow-list + output-root resolution shared with the synthesize
 // phase — both phases must agree on the configured namespace.
-// runPgliteSubagentsInline is shared too: PGLite has no separate Minions
-// worker process (the embedded data-dir holds an exclusive file lock), so a
-// job submitted via queue.add() sits in 'waiting' forever unless something
-// drives the claim -> run -> complete loop inline. synthesize.ts already
-// does this for its own children; patterns.ts previously submitted and
-// waited without ever draining, so every real (non-dry-run) invocation on a
-// PGLite brain hung until subagentWaitTimeoutMs (default 35 min).
-import { loadAllowedSlugPrefixes, loadOutputRoot, runPgliteSubagentsInline } from './synthesize.ts';
+// runSubagentsInline is shared too: a job submitted via queue.add() sits in
+// 'waiting' forever unless something drives the claim -> run -> complete
+// loop — on PGLite because no separate worker can open the embedded
+// data-dir, on Postgres because the parent phase itself occupies a worker
+// slot and can deadlock a fully-occupied worker (#2050). synthesize.ts
+// drains its own children the same way.
+import { loadAllowedSlugPrefixes, loadOutputRoot, runSubagentsInline } from './synthesize.ts';
 import { probeChatModel } from '../ai/gateway.ts';
 import { normalizeModelId } from '../model-id.ts';
 
@@ -185,13 +184,12 @@ export async function runPhasePatterns(
     }
 
     const queue = new MinionQueue(engine);
-    // PGLite children drain inline (no separate worker can open the embedded
-    // data-dir), so give this job a private per-run queue: the inline drain
-    // must never claim unrelated 'default'-queue jobs a Postgres worker owns.
-    // Mirrors synthesize.ts's childQueueName derivation exactly.
-    const childQueueName = engine.kind === 'pglite'
-      ? `dream-inline-${Date.now()}-${randomUUID().slice(0, 8)}`
-      : 'default';
+    // #2050: children drain inline on BOTH engines (see runSubagentsInline),
+    // so give this job a private per-run queue: the inline drain must never
+    // claim unrelated 'default'-queue jobs, and a 'default'-queue worker must
+    // never claim a child this parent is about to run itself. Mirrors
+    // synthesize.ts's childQueueName derivation exactly.
+    const childQueueName = `dream-inline-${Date.now()}-${randomUUID().slice(0, 8)}`;
     const data: SubagentHandlerData = {
       prompt: buildPatternsPrompt(reflections, config.minEvidence, config.sourceSlugPrefix, config.outputSlugPrefix),
       model: config.model,
@@ -207,12 +205,11 @@ export async function runPhasePatterns(
       allowProtectedSubmit: true,
     });
 
-    // PGLite cannot run a separate Minions worker because the embedded DB
-    // holds an exclusive file lock. Drain this phase's private child queue
-    // inline so the parent observes the terminal state instead of polling
-    // waitForCompletion until subagentWaitTimeoutMs expires. No-op on
-    // Postgres (a real worker process claims the job there).
-    await runPgliteSubagentsInline(engine, queue, childQueueName, opts.yieldDuringPhase);
+    // Drain this phase's private child queue inline so the parent observes
+    // the terminal state instead of polling waitForCompletion until
+    // subagentWaitTimeoutMs expires. Runs on BOTH engines — on Postgres the
+    // parent job otherwise deadlocks a fully-occupied worker (#2050).
+    await runSubagentsInline(engine, queue, childQueueName, opts.yieldDuringPhase);
 
     let outcome: string;
     try {
--- a/src/core/cycle/synthesize.ts
+++ b/src/core/cycle/synthesize.ts
@@ -265,27 +265,41 @@ export interface SynthesizePhaseOpts {
   once?: boolean;
 }
 
-const INLINE_PGLITE_LOCK_MS = 30_000;
+const INLINE_LOCK_MS = 30_000;
 
 /**
- * PGLite cannot be served by a separate Minions worker process: the embedded
- * data-dir holds an exclusive file lock, so subagent children enqueued by the
- * synth parent would sit in 'waiting' until waitForCompletion times out.
- * Drive the same claim → run → complete/fail loop a worker would perform,
- * inline, against this phase's private child queue.
+ * Drain this phase's private child queue inline: drive the same claim → run →
+ * complete/fail loop a worker would perform, from the parent's own slot.
+ *
+ * Why inline on BOTH engines:
+ *   - PGLite: no separate Minions worker can run at all (the embedded
+ *     data-dir holds an exclusive file lock), so children would sit in
+ *     'waiting' until waitForCompletion times out.
+ *   - Postgres (#2050): the parent phase itself runs as a job inside a
+ *     `jobs work` process. A worker whose slots are all occupied by such
+ *     parents (autopilot spawns its drain worker at the default
+ *     concurrency=1) can never claim the child the parent is blocking on —
+ *     a structural self-deadlock. Running children inline means a child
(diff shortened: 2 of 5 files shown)


time-attack:
Pulled this out of the fix wave at the last step — the deadlock fix itself reviewed clean and I verified it discriminates (reverting the two src files makes your test fail with `Expected "dead"/Received "timeout"`). The blocker is an interaction with #3555, which is landing in the same wave.

#3555 added `reconnectAfterConnectionError` to `MinionWorker` at five sites: the housekeeping sweep (`worker.ts:308`), `promoteDelayed` (:534), `claim` (:560), and `failJob` (:1159). `runSubagentsInline` calls all four of those same queue operations as bare awaits with no recovery — and `reconnectAfterConnectionError` is a private method, so it isn't reachable from `synthesize.ts`.

That was harmless while the drain was PGLite-only: an embedded DB doesn't get pooler reaps. Extending it to Postgres puts it in exactly the environment whose routine pooler reaps are #3555's premise. The failure mode isn't a retry, it's a permanent orphan — a transient `CONNECTION_CLOSED` during the sweep or `failJob` throws out of the phase, and the child is stranded in a per-run randomly-named `dream-inline-*` queue that no worker will ever claim. On `failJob` specifically, the DB error also masks the original handler failure, which is the exact bug #3555 fixed one layer up.

What would unblock it: route the drain's `promoteDelayed`/`handleStalled`/`handleTimeouts`/`handleWallClockTimeouts`/`claim`/`failJob`/`completeJob` calls through the same retry-and-reconnect path #3555 established — most likely by extracting `reconnectAfterConnectionError` into a shared helper both the worker and the inline drain use, rather than duplicating it. A test that injects a retryable conn error mid-drain and asserts the child still reaches a terminal state would pin it.

Happy to take it as soon as that's addressed; #2050 is a real deadlock and worth fixing. Found by an independent cross-PR review pass over the composed wave, not by anything wrong in your diff in isolation.

time-attack:
Landed in the v0.45.10.0 community fix-wave 2 (#4085) as 9b9bd8b241, rebased/repaired against current master during wave assembly.

Issue #2541, opened 2026-07-02
Local custom embedders can't pass dim validation; wikilink extractor ignores schema-pack path prefixes (4 related bugs, v0.42.56.0)


0-sax-0:
---

**Version:** 0.42.56.0 (bun install from master, 2026-07-02)
**Area:** embedding-dim validation for local/user-provided models

Three connected facets, all in the same corner (`src/core/embedding-dim-check.ts` +
recipes). Net effect: a fully-local embedder that isn't one of the three registered Ollama
models cannot be brought up through any documented path.

## Facet 1 — llama-server: mandatory explicit dims are always rejected

`recipes/llama-server.ts` declares `default_dims: 0 // forces explicit
--embedding-dimensions`, but declares no `dims_options`. In
`validateDimAgainstTouchpoint`, any `requestedDims !== defaultDims` (i.e. anything ≠ 0)
routes into `isCustomDimValidForProvider`, which has no Tier-1 `dims_options` for this
recipe and no Tier-2 allow-list, so it falls to the Tier-3 rejection:

```
Refusing to init: Provider "llama-server" model "<any>" does not support custom
dimensions <any> (this model only emits its default vector size).
```

The recipe's own required flag can never validate. Repro:
`gbrain init --pglite --embedding-model llama-server:anything --embedding-dimensions 1024`

Suggested fix: give llama-server (and other `user_provided_models` recipes) either a
permissive `dims_options` or a `user_provided_models`-aware bypass in
`isCustomDimValidForProvider` — the operator asserting dims is the recipe's design intent.

## Facet 2 — Ollama: any model outside the 3-model list is unusable

`recipes/ollama.ts` registers `['nomic-embed-text','mxbai-embed-large','all-minilm']` at
`default_dims: 768`. For any other Ollama model (e.g. `qwen3-embedding:0.6b`, native
1024d, in the Ollama library), explicit `--embedding-dimensions 1024` is rejected by the
same Tier-3 fallthrough — even though the requested dims are the model's NATIVE size.
Note `mxbai-embed-large` is natively 1024d, so the single 768 default is also wrong for
one of the three registered models.

## Facet 3 — `providers test --model` self-defeats for these recipes

`src/commands/providers.ts` (isolation-test path) configures the gateway with
`recipe.touchpoints.embedding.default_dims ?? 1536` — for llama-server that's `0`, so the
probe always reports "Embedding provider not configured or not ready". Separately,
[...]

PR #3560, opened 2026-07-28
fix(extract): stop silently dropping links from non-whitelisted directories (#2576)

time-attack:
Fixes bug 2 of #2576 (bugs 1 and 3 shipped in #2717, whose description explicitly left bug 2 to its own PR).

## Confirmed on master (executed matrix)

For a non-whitelisted directory (`ops/`), 5 of 6 reference shapes were DROPPED, with identical `people/` controls resolving in all 6:

| shape | `global_basename` off | on |
|---|---|---|
| `[Pointer](../ops/services/pointer-agent.md)` | DROPPED | DROPPED |
| bare prose `ops/services/pointer-agent` | DROPPED | DROPPED |
| `[[ops/services/pointer-agent]]` | DROPPED | `wikilink_basename` only (untyped, via #2866's path-tail rescue) |

## Root cause and fix shape

`DIR_PATTERN` gated candidate **emission**, but candidate **persistence** never depended on it — every persist path already validates page existence (`resolveCandidateSources` in `extract.ts`, `put_page`'s `allSlugs` filter, `addLinksBatch`'s INNER JOINs on `(slug, source_id)`). The whitelist could therefore only subtract true edges; a user with a custom schema pack hit it for every directory they invented, silently.

Fix: emit candidates for **any dir-shaped path** and let the existing existence checks decide.

- `ENTITY_REF_RE` (markdown links) and the bare-slug prose regex match `ANY_DIR_SEGMENT` instead of the whitelist.
- Slash-shaped wikilinks outside the whitelist get a direct verb-typed candidate in pass 2c — the same treatment pass 2b gives whitelisted dirs, and it works with the flag off. The flag-gated basename suffix rescue (`[[notes/x]]` → `vault/notes/x`, #2866) is preserved for non-exact matches and no longer double-emits the exact literal.
- Deliberately NOT a pack-derived prefix registry: the extraction core is pure (no DB), and deriving prefixes from the active pack would still drop links to real pages in undeclared dirs. Existence-at-persist is strictly more accurate than any prefix list and needs no registry to maintain. (Also resolves the #1547 community idea; noted in `docs/designs/COMMUNITY_IDEAS.md`.)
- Self-loop guard on the bare-slug pass (latent on master for whitelisted dirs).

## Observability

A silently dropped edge with no counter is what let this hide. The DB extract paths now count candidates skipped for missing endpoints and report them in the human summary, the `--json` payload (`skipped_missing_target`), and `extractStaleFromDB`'s return (`skippedMissingTarget`).

## Re-extraction

`LINK_EXTRACTOR_VERSION_TS` bumped to `2026-07-28T00:00:00Z` — pages stamped by pre-fix sweeps re-flag as stale, so existing brains recover the missing edges on the next `extract --stale` sweep instead of keeping them missing forever.

## Proof

New `test/link-extraction-dir-whitelist-2576.test.ts` on **unmodified master**: 7 fail / 3 pass (the 3 passers are regression pins: suffix rescue, self-loop guard, bare-`[[name]]` flag gating). On this branch: 10/10 pass.

Existing pins updated where they asserted the old drop/demotion behavior (`link-extraction.test.ts`, `e2e/global-basename-pglite.test.ts` — exact-path wikilinks now produce the typed edge instead of a `wikilink_basename` demotion).

`bun run typecheck` clean; `bun run verify` 32/32 green; extract/doctor/backlinks/put_page suites green (3 unrelated hook-timeout flakes in a 50-file parallel batch pass standalone). No engine files touched; no JSONB changes. Docs: `KEY_FILES.md` entries updated to current behavior + `bun run build:llms` regenerated (freshness test green).


time-attack:
The `test (9)` failure is #3554, not this PR — same three `eval-trajectory` tests, same `expected 1280 dimensions, not 1536` signature (3 occurrences in the shard log). Your CI ran before #3557 landed.

That's now the third PR caught by it (#3545 and #3556 were the others, both green after a rebase). Rebasing this one onto the fixed master.

time-attack:

The diagnosis is correct and the fix is at the root cause: DIR_PATTERN gated candidate *emission* while every persist path already gated on page existence, so the whitelist could only subtract true edges. The discrimination test is real (7 fail / 3 pass on master source, 10/10 on branch, both executed). But I reproduced two defects the PR ships: (1) the provenance flip for exact-path wikilinks (wikilink_basename/wikilink-resolved -> mentions/markdown) combined with the forced LINK_EXTRACTOR_VERSION_TS re-sweep and insert-only extract paths leaves DUPLICATE edges in the graph after upgrade — executed, getLinks returns 2 rows for the same pair, with no migration or cleanup in the PR; (2) the headline observability counter is ~93% noise because ANY_DIR_SEGMENT on the bare-prose regex matches ordinary English (`and/or`, `read/write`, `8/10`, `try/catch`) and URL tails, so it can never surface the genuinely dropped edge it was added to surface.

### Discrimination test

PASSES — executed. I checked out pr3560, then `git checkout origin/master -- src/core/link-extraction.ts src/commands/extract.ts` (stubbing the old behavior back in with the new test file intact) and ran `bun test test/link-extraction-dir-whitelist-2576.test.ts`: EXIT=1, 3 pass / 7 fail, with the 7 failures being exactly the ops/ and startups/ candidate-emission cases plus the VERSION_TS bump assertion (`Received: undefined` / `Received: []` / `Received: "wikilink_basename"`). Restoring the branch files gave 10 pass / 0 fail. The 3 passers are honestly labelled in the file as regression pins (suffix rescue, self-loop guard, bare-[[name]] flag gating). The test genuinely discriminates.

### Blockers

**1.** BLOCKER — src/core/link-extraction.ts:506-523 + src/core/link-extraction.ts:143 (VERSION_TS bump): post-upgrade DUPLICATE graph edges, executed and confirmed. The links unique key is (from_page_id, to_page_id, link_type, link_source, origin_page_id) [src/schema.sql:493], and `extract --stale` / `extract links --source db` are insert-only (addLinksBatch ON CONFLICT DO NOTHING — no reconciliation; only put_page reconciles). Pre-fix, an exact-path wikilink `[[notes/struktura]]` to an existing page wrote (wikilink_basename, wikilink-resolved). Post-fix, the new pass-2c direct candidate writes (mentions, markdown) — a DIFFERENT unique key — and the LINK_EXTRACTOR_VERSION_TS bump to 2026-07-28 *guarantees* every stamped page is re-swept. FAILING INPUT (executed in PGLite): pages `notes/struktura` + `concepts/x` with body `See [[notes/struktura]].`, `link_resolution.global_basename=true`, seed the pre-fix edge via addLink('concepts/x','notes/struktura','wikilink (basename match)','wikilink_basename','wikilink-resolved'), then `runExtract(engine, ['links','--source','db'])`. Result: getLinks('concepts/x') = [[notes/struktura, wikilink_basename, wikilink-resolved], [notes/struktura, mentions, markdown]] — 2 edges for one reference. The stale wikilink_basename row is never removed (no migration, no reconcile step in the diff), and per src/schema.sql:470-475 `mentions` is filtered out of backlink-count for ranking while `wikilink_basename` is not, so graph traversal and graph-query double-count the pair permanently. Affects every brain with global_basename enabled that uses slash-shaped wikilinks. Fix: either keep the wikilink_basename provenance for the exact literal (don't flip), or ship a migration/reconcile that drops the superseded wikilink_basename/wikilink-resolved row whose to_slug equals the literal written path.

**2.** NEEDS_CHANGES — src/core/link-extraction.ts:177 (ANY_DIR_SEGMENT) applied to the bare-slug prose regex at src/core/link-extraction.ts:589: the observability counter added by this same PR is defeated by the widening. `\b([a-z0-9][a-z0-9_-]*\/[a-z0-9][a-z0-9/-]*[a-z0-9])\b` matches ordinary prose and URL tails. MEASURED (executed) over this repo's own docs/ tree, 120 markdown pages: master extractor 9 candidates -> branch extractor 536 candidates (59x), top hits `com/garrytan/gbrain`, `10/10`, `before/after`, `person/company`, `5/10`, `application/json`, `try/catch`, `read/write`, `8/10`, `1/2/3`, `symlinks/hardlinks/devices`. EXECUTED end-to-end on a seeded PGLite brain with one realistic doc page: `Links: created 1 from 2 pages (db source)` followed by `Skipped 13 candidate(s) whose target page doesn't exist` — 1 real edge, 13 noise skips from a single page. The PR body's stated rationale is "A silently dropped edge with no counter is what let this hide", but at 13 noise skips per prose page (and ~527 for 120 docs pages) a genuinely dropped edge is invisible in the counter, and the line prints on essentially every run of every brain. Minimal fix: only count skips for candidates from explicit-intent shapes (markdown link / wikilink), not the bare-prose pass; and/or tighten the bare-prose pass (require >=2 path separators, or a `-`/`_` in the tail segment) so `and/or` and `8/10` stop being link candidates.

### Not verified

- The DB-path relative-link mis-resolution class I found by code read but did NOT execute: extractEntityRefs strips `../` and keeps the literal path with no ancestor walk (the FS path's resolveSlug DOES walk ancestors). Post-fix, a page at `wiki/notes/index` writing `[Ref](api/reference.md)` for a sibling now produces an edge to a top-level `api/reference` page if one exists — a WRONG edge where master produced none. This class already exists on master for whitelisted dirs (`[X](people/alice.md)` from a nested page), so the PR widens rather than introduces it. I did not build the 3-page PGLite fixture to confirm.
- The severity of the cross-source amplification via resolveCandidateSources' `toSourceId = 'default'` fallback (src/commands/extract.ts:131) on a real federated brain. The code path is pre-existing and unchanged, but the set of tokens that can reach it grows ~59x. I did not build a two-source fixture to measure whether this produces observable cross-source edges in practice, nor whether graph reads scope them out.
- Full unit suite (~4000 tests across all shards) — not run. I ran the 9 test files that cover the changed code plus the two files the PR edited. The PR body's claim of '3 unrelated hook-timeout flakes in a 50-file parallel batch' is unverified by me.
- Postgres-engine behavior — no DATABASE_URL available, so all DB probes ran on PGLite only. Since the diff touches no SQL and no engine file, and the unique constraint text is identical in schema.sql, I judge the duplication finding to hold on Postgres, but I did not execute it there.
- E2E lifecycle / bun run ci:local — not run (time).

- `gh pr view 3560 — exit 0`
- `gh pr diff 3560 > /tmp/pr3560.diff — exit 0 (496 lines, every hunk read)`
- `git fetch origin pull/3560/head:pr3560 && git checkout pr3560 — exit 0 (base = 913d2d7f, 1 fix commit + 1 master merge)`
- `bun install --frozen-lockfile — exit 0`
- `bun run verify — exit 0 (32/32 checks green, 26s)`
- `DISCRIMINATION: git checkout origin/master -- src/core/link-extraction.ts src/commands/extract.ts && bun test test/link-extraction-dir-whitelist-2576.test.ts — exit 1 (3 pass / 7 fail)`
- `git checkout pr3560 -- <same 2 files> && bun test test/link-extraction-dir-whitelist-2576.test.ts — exit 0 (10 pass / 0 fail)`
- `bun test test/link-extraction.test.ts test/link-extraction-code-refs.test.ts test/backlinks.test.ts test/extract-fs.test.ts test/extract-stale.test.ts test/link-extraction-dir-whitelist-2576.test.ts — exit 0 (245 pass / 0 fail)`
- `bun test test/e2e/global-basename-pglite.test.ts test/doctor-links-extraction-lag.test.ts — exit 0 (19 pass / 0 fail)`
- `bun test test/build-llms.test.ts — exit 0 (12 pass / 0 fail)`
- `PROBE scan_probe.ts over docs/ with BRANCH extractor — exit 0 (120 pages -> 536 candidates, 462 distinct)`
- `PROBE scan_probe.ts over docs/ with MASTER extractor — exit 0 (120 pages -> 9 candidates, 9 distinct)`
- `PROBE test/probe-upgrade-dup.test.ts (PGLite, global_basename on, seeded pre-fix edge, runExtract links --source db) — exit 1: 2 edges for one reference (BLOCKER 1 confirmed)`
- `PROBE test/probe-counter-noise.test.ts (PGLite, one realistic prose page) — exit 0: printed 'Links: created 1' + 'Skipped 13 candidate(s)' (BLOCKER 2 confirmed)`
- `grep -nE "node:fs|readFile|process.cwd" src/core/link-extraction.ts — exit 0 (no filesystem surface; traversal cleared)`
- `All 3 probe files deleted; git status --short clean at end`

</details>

---

time-attack:
**Merging this closes #2541** (severity **p1**).

#2541 was one of 58 issues carrying a `p3` label. All 58 were re-triaged by reproducing them against current master, and **the label was wrong on 56 of them** — the corrected distribution is 4 p0, 40 p1, 12 p2, and 2 genuinely p3. #2541 came out at **p1**, and this PR is what fixes it.

Two things follow from that:

- Add `Closes #2541` to the PR description if you want the link wired up — a closing keyword only fires from the body or a commit message, not from a comment like this one.

The reproduction and evidence are on #2541.

time-attack:
Merged into master via the v0.42.70.0 community fix wave (#3710) with your commit attributed via Co-Authored-By and CHANGELOG credit. One composite-review addition on top of your fix: with the DIR_PATTERN gate removed, the bare-path scanner could read inside `[[...]]` spans and emit a spurious edge to a wikilink's parent page — wikilink spans are now masked (equal-length blanks) before that pass, with a discriminating test. Your ungating itself is unchanged. Master CI is green. Thank you!

Issue #1696, opened 2026-05-31
extract is not wired into the default maintenance loop — brains silently accumulate 0% link/timeline coverage


garrytan:
## Symptom (real-world, 280K-page brain)

On a production brain of ~280K pages, `pages.edges_backfilled_at` was **NULL on every single page** and the link table was **99.7% untyped `mentions`** (1,042,187 of 1,045,858 links). Only a few hundred typed edges existed total:

```
mentions: 1042187
attended: 887
related_to: 774
founded: 251
works_at: 221
invested_in: 70
advises: 39
```

The brain had been ingesting for months. `extract all` had effectively **never run at scale**, so the graph could find entity *names* in text but had almost no semantic relationships (`founded`, `works_at`, `invested_in`, `advises`) to reason over. `gbrain think`, graph traversal, and link-based retrieval were all running on a graph that was structurally a bag of mentions.

After manually running `extract all` once: typed edges jumped to founded 7,232 / invested_in 3,016 / works_at 2,417 / advises 1,430 — a **+12,500 typed-edge** gain from a pass that had simply never been triggered.

## Root cause

`autopilot`'s cycle is `sync → extract → embed`, but when autopilot isn't installed/running (the common case for CLI-first or externally-cron'd deployments), nothing ever calls `extract`. `sync` and `embed` are the steps people wire into their own cron because they're the obvious 'ingest' and 'make searchable' steps. `extract` is the silent third leg, and there is no warning, health-check failure, or `doctor` signal when it's been skipped.

## Proposed fixes (any subset)

1. **`doctor` check for extraction lag.** Add a health signal: `% of non-tweet/non-digest pages with edges_backfilled_at older than EDGE_EXTRACTOR_VERSION_TS or NULL`. Fail (or warn loudly) above a threshold (e.g. >20%). This is the cheapest high-leverage fix — it makes the gap *visible* instead of silent.
2. **Fold `extract` into `sync` by default** with an opt-out flag (`sync --no-extract`), so the canonical 'pull new content' command also types its edges. `sync` already knows which pages are new/changed, so this is incremental, not a full re-scan.
3. **Emit a one-line stderr nudge** at the end of `sync` when extraction coverage is low: `[sync] N pages have no extracted edges — run 'gbrain extract all' or enable autopilot`.
4. **`extract --stale`** mirroring `embed --stale`: only process pages whose `edges_backfilled_at` is NULL or older than the extractor version timestamp. Makes the incremental case trivial to cron.

[...]

PR #3561, opened 2026-07-28
fix(sync,webhook): consume deferred link extraction above the size gate (#2849)

### LINKED ISSUE #2849: GitHub webhook sync can leave link extraction permanently stale

time-attack:
Closes #2849. PR #2850 fixed the sub-gate case (both push triggers now submit `noExtract: false`) and its own description said #2849 "should stay open until a consumed-commit-bound, source-scoped extraction follow-up covers `totalChanges > 100`." This is that follow-up.

## Confirmed on master

Above the gate, `performSync` only prints a "deferring link/timeline extraction" hint and leaves `links_extracted_at` unstamped (`src/commands/sync.ts:3451-3457`). The autopilot cycle's extract phase is **not** a stale sweep — it's `runExtractCore({mode:'all', slugs: syncPagesAffected})` (`src/core/cycle.ts`), so the next cycle's `up_to_date` sync hands it an empty slug list and it reports `pages_processed=0`. Reproduced: a 101-file sync on master queues nothing and the pages stay stale forever; only a manual `gbrain extract --stale` recovers.

## Fix shape

The size gate stays (its whole point is keeping a huge synchronous extract out of the webhook request path). What changes is that the deferred work is now **durably queued and consumed**:

- **Queued at the point of deferral** — `performSync`'s defer branch submits an `extract` Minion job `{stale: true, sourceId?, deferred_commit}`:
  - **commit-bound**: idempotency key `extract-stale:<sourceId|default>:<headCommit>`, so webhook redeliveries / re-syncs of the same commit coalesce onto one job (`maxWaiting: 1` backstops pileup);
  - explicit `timeout_ms` (35 min) covering the sweep's internal 30-min budget — the tight null-default wall-clock would otherwise kill it mid-sweep (#1737 class);
  - best-effort: if submission fails, behavior degrades to the pre-fix hint (pages stay stale and doctor-visible, never mis-stamped).
- **Consumed by the existing `extract` handler** — `{stale: true}` routes through the existing `extractStaleFromDB` (checkout-less, resumable, idempotent — the exact sweep the hint told operators to run), no new job kind invented. When the internal budget leaves `staleRemaining > 0` with forward progress, the handler chains a continuation job so very large backlogs converge.
- **Trust boundary**: the webhook enqueues nothing new — it still submits only the `sync` job for its HMAC-verified registered source. The deferred extract job is created by the trusted sync path and scoped to that sync's `sourceId`, so an untrusted push can only ever trigger a sweep of its own source. The unscoped shape occurs only for local CLI syncs of a default brain (matching what the CLI hint already tells that operator to run). No hand-rolled source filtering: scoping rides `extractStaleFromDB`'s existing `sourceIdFilter`.

The webhook path is exactly the path where a jobs worker is guaranteed (the sync job itself is consumed by one), so the queued follow-up always runs there.

## Proof

New `test/sync-deferred-extract-queue.serial.test.ts` on **unmodified master**: 2 fail / 3 pass — no job exists after a 101-file sync, and the handler ignores `{stale: true}` (it would dir-walk instead). The 3 passers are negative pins: sub-gate syncs queue nothing (inline extract still owns them, #1696 stamp contract asserted), `--no-extract` suppresses the job, and re-syncing the same commit doesn't pile up jobs. On this branch: 5/5 pass, including end-to-end handler consumption — links materialize (`notes/page-7 → people/alice`) and the watermark is stamped with `staleRemaining=0`.

`bun run typecheck` clean; `bun run verify` 32/32 green; sync/handlers/webhook/extract-stale suites green (117/117). No engine files touched; no JSONB casts (job payloads go through `MinionQueue.add`'s existing binding). `KEY_FILES.md` entries updated + `bun run build:llms` regenerated (freshness test green).


time-attack:

The PR does what it claims: above `performSync`'s `totalChanges > 100` gate it now submits a commit-bound, source-scoped `extract` minion job with `{stale: true}`, and the `extract` handler routes that through the existing `extractStaleFromDB` watermark sweep. Typecheck, `bun run verify` (32/32) and the related sync/webhook/extract-stale suites are green, and the discrimination test is real — both hunks fail independently when reverted. But the submission passes `maxWaiting: 1`, and `MinionQueue.add`'s backpressure coalesce is keyed only on (name, queue, data.sourceId) with a NULL sourceId matching EVERY waiting `extract` row. I reproduced it: with one unrelated waiting `extract` job present, the fix silently no-ops, `links_extracted_at` stays NULL, and the operator log falsely reports "queued stale-sweep job #1 ... a running jobs worker will consume it". One-line fix, so NEEDS_CHANGES not CLOSE.

### Discrimination test

PASSES — and I verified both hunks separately. (1) Kept the new test, reverted src/commands/sync.ts + src/commands/jobs.ts to the merge base 6136e139: `bun test test/sync-deferred-extract-queue.serial.test.ts` → EXIT=1, 3 pass / 2 fail; test 1 fails on `expect(jobs.length).toBe(1)` Expected 1 / Received 0, test 5 fails with `TypeError: undefined is not an object (evaluating 'job.id')` because no job row exists. (2) Restored PR sync.ts and reverted ONLY jobs.ts: `bun test ... -t "handler consumes"` → EXIT=1, 0 pass / 1 fail, `expect(result.stale).toBe(true)` Expected true / Received undefined — so the handler hunk is independently pinned, not just carried by the sync hunk. (3) On PR HEAD, 5 pass / 0 fail, EXIT=0. The three "negative pin" tests do pass both ways by design (sub-gate, --no-extract, no-pileup); test 2's `toBeLessThanOrEqual(1)` is weak (it would also pass if nothing were ever queued) but tests 1 and 5 carry the discrimination.

### Blockers

**1.** src/commands/sync.ts:3478 — `maxWaiting: 1` silently swallows the stale-sweep submission whenever ANY unrelated `extract` job is already waiting, because MinionQueue.add's coalesce filter (src/core/minions/queue.ts:206) is `name=$1 AND queue=$2 AND status='waiting' AND ($3::text IS NULL OR data->>'sourceId' IS NOT DISTINCT FROM $3)` — with the unscoped default-brain payload bpSourceId is NULL, so the sourceId predicate short-circuits and every waiting `extract` row counts. EXECUTED REPRO (scratch test, PGLite, PR HEAD f66bc9f1): queue.add('extract', {mode:'links', dir:<empty dir>}) → then a 101-file unscoped sync. Result: `total extract jobs = 1, stale jobs = 0`; running the coalesced job through the registered handler returns `{links_created:0, timeline_entries_created:0, pages_processed:0}` and `SELECT links_extracted_at FROM pages WHERE slug='notes/page-7'` is still NULL — i.e. #2849's exact symptom persists. This is not hypothetical: src/core/brain-score-recommendations.ts:288 emits an `extract.all` recommendation with `{mode:'all', dir}` (no sourceId) and src/core/remediation/run.ts:256 submits it with `maxWaiting: 1`, so any concurrent `gbrain doctor --remediate` / advisor apply opens the window. The idempotency key already gives commit-level dedup; `maxWaiting: 1` adds nothing here except this false-positive coalesce.

**2.** src/commands/sync.ts:3487 — `queuedJobId = job.id` is trusted without checking that the returned row is the job just submitted. `queue.add` returns a pre-existing row on both the coalesce path (above) and the idempotency fast-path (queue.ts:146, which returns even a `completed` row for the same key). EXECUTED: the log printed `Large sync: deferring link/timeline extraction — queued stale-sweep job #1 (source: default); a running jobs worker will consume it.` while job #1 was the unrelated `{mode:'links'}` job and no `{stale:true}` job existed. The operator (and any log-scraping monitor) is told the deferral was banked when it was dropped. Guard with e.g. `job.idempotency_key === key && job.data?.stale === true` before claiming it, and fall back to the hint-only wording otherwise.

**3.** test/sync-deferred-extract-queue.serial.test.ts — no test covers a non-empty `extract` queue, which is the only state in which the new code silently fails. The suite's `beforeEach` does `DELETE FROM minion_jobs`, so every case runs against a clean queue. Add the coalesce case (a waiting unrelated `extract` job must NOT suppress the stale submission).

### Not verified

- The full unit suite (~4000 tests / 10 shards) — I ran only the new file, build-llms, and four directly related suites. A cross-suite interaction with the new `extract` handler branch is possible though unlikely (the branch is gated on `job.data.stale === true`).
- Postgres-engine behavior: no DATABASE_URL, so `bun run test:e2e` / engine-parity ran nowhere. The coalesce repro was on PGLite only. The offending filter is plain SQL with pg_advisory_xact_lock and behaves identically on both engines, but I did not execute it against real Postgres.
- A live end-to-end webhook → jobs-worker run (HTTP POST with a real HMAC + `gbrain jobs work` consuming the chained job). I verified the webhook code path by reading serve-http.ts:2340-2405 and exercised the handler directly, not through a running worker.
- Continuation-chain convergence under a real 30-min budget hit: `STALE_TIME_BUDGET_MS` (src/commands/extract.ts:85, GBRAIN_EXTRACT_TIME_BUDGET_MS-overridable) never fired in a 101-page fixture, so the `staleRemaining > 0 && pagesProcessed > 0` chaining branch is untested by the PR and unexecuted by me. Related unexecuted concerns worth a second look: the hardcoded `timeout_ms: 35 * 60 * 1000` does not derive from that env-configurable budget (raise the env above 35min and you re-create the #1737 kill the comment says it avoids), and the handler never passes `job.signal` into extractStaleFromDB (which accepts none), so a timed-out or cancelled sweep keeps running after the job is dead-lettered.
- PR title does not lead with the version, contrary to CLAUDE.md's PR-title IRON RULE (`v0.42.x fix(sync,webhook): ...`). Process item for whoever lands it; I did not edit the PR.

- `gh pr view 3561 → 0`
- `gh pr diff 3561 → 0`
- `git fetch origin master pull/3561/head:pr3561 && git checkout pr3561 → 0 (HEAD f66bc9f1, base 6136e139, 13 commits behind master, mergeStateStatus CLEAN)`
- `bun install → 0 (node_modules was empty in the worktree; first typecheck attempt exited 127 'tsc: command not found')`
- `bun run typecheck → 0`
- `bun run verify → 0 (pass=32 fail=0, 27s)`
- `bun test test/sync-deferred-extract-queue.serial.test.ts (PR HEAD) → 0 (5 pass / 0 fail, 58.70s)`
- `git checkout 6136e139 -- src/commands/sync.ts src/commands/jobs.ts; bun test test/sync-deferred-extract-queue.serial.test.ts → 1 (3 pass / 2 fail, 75.71s)`
- `git checkout pr3561 -- src/commands/sync.ts (base jobs.ts only); bun test test/sync-deferred-extract-queue.serial.test.ts -t 'handler consumes' → 1 (0 pass / 1 fail, result.stale undefined)`
- `bun test test/build-llms.test.ts → 0`
- `bun test test/extract-stale.test.ts test/sources-webhook.test.ts test/sync-inline-extract-stamps.serial.test.ts → 0 (31 pass across 3 files)`
- `bun test test/zz-scratch-coalesce.serial.test.ts (reviewer scratch, since deleted) → 1 — reproduced the maxWaiting coalesce defect: 'total extract jobs = 1, stale jobs = 0', handler result {pages_processed:0}, links_extracted_at NULL, log claimed 'queued stale-sweep job #1'`

</details>

---

time-attack:
Review finding (blocking, one-token fix): the bug is real — the size gate only prints a hint, never stamps `links_extracted_at`, and the cycle's extract phase is slug-scoped, so a >100-file sync stays stale forever. The queued-job fix is the right shape. The blocker: the idempotency key binds to `headCommit`, but `MinionQueue.add` returns an existing **completed** row for a reused key (only `dead`/`cancelled` free the slot). So a checkpoint-resumed second size-gated sync toward the same HEAD gets the already-finished job handed back, and the newly imported pages stay stale — the reported bug survives on exactly the resume path. Key on the persisted pin instead of `headCommit`.

Secondary: `maxWaiting: 1` counts all waiting `extract` jobs for the source, so a hand-submitted `gbrain jobs submit extract` silently swallows the stale sweep while the log claims 'queued'; and the dedup test asserts `toBeLessThanOrEqual(1)`, which passes on master and pins nothing — tighten it to assert the actual coalescing behavior.

time-attack:
Landed in the v0.45.10.0 community fix-wave 2 (#4085) as bd4c976a85, rebased/repaired against current master during wave assembly.

PR #3563, opened 2026-07-29
feat(by-mention): Unicode-aware word tokenizer — Vietnamese/diacritic name extraction

cfeddersen:
## Problem

The by-mention entity scanner tokenizes with an ASCII-only regex:

```js
const TOKEN_RE = /[a-zA-Z0-9]+/g;   // src/core/by-mention.ts
```

Both `tokenizeForScan` (body text) and `tokenizeTitle` (gazetteer) share it, and neither folds diacritics. Every non-ASCII, non-CJK letter is dropped and the surrounding ASCII fragments become separate tokens, so Latin-script names with diacritics shred. Vietnamese is the motivating case:

```
"Nguyễn Văn Đức" -> nguy | n | v | n | c
"Đà Nẵng"        -> n | ng
```

The gazetteer then keys on garbage fragments and matching runs on 1–2 char tokens like `n`/`ng`. Two failure modes result: **false-positive explosion** (short fragments collide with unrelated text) and **diacritic collapse** (`Ngọc` == `Ngoc` == `Ngộc` -> `ng`+`c`, merging distinct entities). Net: `extract links --by-mention`, backlinks, entity timelines, and find-experts are all degraded for Vietnamese (and any diacritic Latin script).

This is **not** the problem `feat: CJK entity extraction` (#1637) solved. CJK needed char-level tokenization + min-length-2 because it has no word boundaries. Vietnamese is space-delimited; it only needs the tokenizer to stop discarding non-ASCII letters. Search is unaffected — it uses embeddings + the already-Unicode-aware `tokenizeTitle` in `search/title-match.ts`; this PR is specifically the entity-link graph.

## Fix

`TOKEN_RE` becomes a Unicode word-run tokenizer, with CJK excluded so it keeps flowing through the existing per-character path unchanged:

```js
const TOKEN_RE = new RegExp(
  `(?![${CJK_CHARS}])[\\p{L}0-9]` +
  `(?:(?![${CJK_CHARS}])[\\p{L}\\p{M}0-9])*`,
  'gu',
);
```

Four boundaries are deliberate, and each one was a real regression caught in review of this PR's first version (which used `/[[\p{L}\p{M}\p{N}]--[CJK]]+/gv`):

- **The lead must be a letter or ASCII digit**, so a token can never consist of combining marks alone. U+FE0F (VARIATION SELECTOR-16, category `Mn`) rides on most emoji, so a mark-only token would hijack the gazetteer key of every emoji-prefixed entity title — `❤️ Health Notes` keying on a bare U+FE0F instead of `health`, losing its plain-text links and collapsing all emoji-prefixed titles into one mutually-confusable bucket.
- **Combining marks are allowed after the lead.** NFD Vietnamese is base letter + mark, so excluding `\p{M}` would re-fragment the exact names this tokenizer exists to keep whole.
- **Digits are ASCII-only**, exactly as `/[a-zA-Z0-9]+/` was. `\p{N}` additionally mints tokens for `¹` `½` `１` (Nl/No/non-ASCII Nd), and `findMentionedEntities` requires an entry's tokens to be **strictly adjacent** in the body — so a superscript between the words of `Acme Corp` silently breaks a match that used to work.
- **Plain `u` flag, not `v`.** The CJK exclusion is a negative lookahead over the shared class, the same construction `src/core/think/gather.ts:203` already uses. No set-subtraction syntax, no es2024 target requirement.

Tokens are additionally **NFC-normalized**, so an NFD body matches an NFC gazetteer title and vice versa — without it the fix is inert for exactly the input it targets. Normalization is applied **per token, never over the whole text**: `Mention.offset` is contracted to index into the original body, which `extract-ner.ts` slices context windows from to infer link verbs, and normalizing up front would silently shift every offset.

The CJK class is `CJK_SLUG_CHARS` (`src/core/cjk.ts`), the repo-wide single source of truth since #3522, used verbatim. `TOKEN_RE`, `hasCJK()`, `cjkCharCount()` and both per-character walkers now derive from that one import, **replacing four inline copies of the ranges with none**.

That carries one deliberate behaviour change, called out because it is not free: this module's own copy of the ranges also covered Han Extension A (U+3400–4DBF), which `cjk.ts` scopes out repo-wide. Ext-A is therefore no longer char-level in by-mention — it tokenizes as a word run, and an Ext-A-only entity title becomes a single sub-4-character token that falls below `MIN_NAME_LENGTH` instead of qualifying under `MIN_CJK_NAME_LENGTH`. Search, chunking and slug grammar already ignore Ext-A, so this makes by-mention consistent with them instead of being the lone subsystem that disagreed. In-scope CJK is untouched. Happy to restore Ext-A here if that scope call is wrong.

Result:

```
"Nguyễn Văn Đức" -> nguyễn | văn | đức
"Đà Nẵng"        -> đà | nẵng
"纳瓦尔"          -> (excluded; char-level path emits 纳|瓦|尔 as before)
"㐀㐁"            -> 㐀㐁   (Ext-A; word run now, was char-level — see above)
"Acme Corp"      -> acme | corp   (unchanged)
"web3 and h2o"   -> web3 | and | h2o   (ASCII digits unchanged)
```

Diacritics stay significant (`Hồng` does not match `Hong`), matching the multilingual intent.

## Evidence (old vs new, on a 1,251-page Vietnamese corpus)

Re-measured against the **final** tokenizer in this PR, not the first version. Read-only: page bodies are read from a real brain and run through `findMentionedEntities` with a one-entity gazetteer built by each tokenizer. No links are written. **These are fresh-scan counts on a brain with no pre-existing `link_source='mentions'` rows** — see Known limitation below.

| Entity | OLD tokens | NEW tokens | OLD pages | NEW pages | substring presence |
|---|---|---|---|---|---|
| Đà Nẵng | `[n\|ng]` | `[đà\|nẵng]` | 799 | **431** | 433 |
| Nguyễn | `[nguy\|n]` | `[nguyễn]` | 260 | **97** | 97 |
| Unicity | `[unicity]` | `[unicity]` | 178 | **178** | 179 |
| Zalo | `[zalo]` | `[zalo]` | 244 | **244** | 251 |

OLD carried +85% false positives on `Đà Nẵng` and +168% on `Nguyễn`. Both ASCII controls are identical old vs new — the change is a no-op for ASCII, as intended.

The last column is crude case-insensitive substring presence, which is an **upper bound**, not ground truth — it counts occurrences the scanner is designed to skip. Every one of the 10 pages where substring hits and the tokenizer does not was classified, and all 10 are intentional:

- **9 are mentions that occur only inside code blocks**, which `findMentionedEntities` removes via `stripCodeBlocks` before scanning. Pre-existing behaviour, unrelated to this PR (e.g. `` `Đà Nẵng` in first 100 words `` inside backticks; a `contact:\n  zalo:` YAML fence).
- **1 is a compound word** — `AskTinaZalo` tokenizes as `asktinazalo`, which correctly does not match the entity `Zalo`.

There are no unexplained differences. `Nguyễn` matches substring presence exactly because it happens to have no code-block-only occurrences in this corpus.

(The earlier revision of this description quoted 820/440/271/187/258 against a 1,264-page corpus. Those were measured on 2026-07-29 with the first tokenizer; the corpus has since changed size. The numbers above supersede them.)

## Tests

`test/by-mention.test.ts`: 34 -> 46 tests. `bun test` on the 9 suites that touch `by-mention` or `cjk` -> 119 pass / 0 fail. `tsc --noEmit` clean. `check:privacy`, `check-test-real-names`, `check-test-isolation` -> exit 0. Fictional Vietnamese names only.

**Discrimination.** The fixture gazetteer builder now calls the production `tokenizeTitle` instead of re-declaring the tokenizer — a duplicated copy makes every test here non-discriminating, because reverting the source would leave the fixture on the new tokenizer and title/body would keep agreeing. Every Vietnamese case now asserts **tokenization**, not just the resolved mention: a mention-only assertion passes even under the old tokenizer, which fragmented title and body symmetrically (a 5-fragment gazetteer entry still matched a 5-fragment body run) while `Mention.name` is copied from the untouched `title` column.

Reverting `TOKEN_RE` to `/[a-zA-Z0-9]+/g` in the source **fails 9 of these tests** (8 Vietnamese + 1 `buildGazetteer`). In the first version of this PR it failed 1 of 8.

Three further tests guard the tokenizer's own boundaries — mark-only tokens, non-ASCII-numeric adjacency, and the Ext-A scope change. The first two pass on master by construction: they guard *this change*, not the original bug.

One reviewer case is deliberately **not** restored: a combining mark bound directly to a word (`Acme̶ Corp`, U+0336 inside the run) still does not match. A mark inside a word genuinely alters the word, and the only fix — NFKD + strip `\p{M}`, as `think/gather.ts` does — would destroy the diacritic significance this PR exists to establish. It is asserted as expected behaviour rather than left silent.

## Known limitation — existing `link_source='mentions'` rows are not cleaned up

`engine.addLinksBatch` is additive. The checkpoint's `gazetteerHash` does self-invalidate when the gazetteer keys change, so a re-scan genuinely happens — but the correct links are **added alongside** the false positives the old tokenizer already wrote. A brain that has already run `extract links --by-mention` therefore does not reach the numbers above by upgrading; those are fresh-scan figures.

This is deliberately **not** fixed here, because the correct cleanup is a larger design question than a tokenizer change:

- `link_source='mentions'` is a **shared namespace with two producers**. `extract-ner.ts` deliberately writes `link_source='mentions'` + `link_kind='typed_ner'` rather than minting new provenance. Deleting by `link_source` alone would wipe every NER-derived typed link; deleting only `link_kind != 'typed_ner'` would leave NER rows that were derived from the same bad gazetteer. Both producers need a coordinated purge + re-run.
- There is no bulk primitive to build on — the engine exposes only per-pair `removeLink(from, to, linkType?, linkSource?)`. A cleanup needs a new `BrainEngine` method implemented on both `postgres-engine.ts` and `pglite-engine.ts`, a new CLI flag, and a run of the `DATABASE_URL`-gated engine-parity suite.

Filed as #3674 so it can be reviewed as the write-path change it is.

## Notes

- `MIN_NAME_LENGTH = 4` still filters short single-token names the same across languages — left as-is; a language-aware min-length is a separate policy call.
- NFD Hangul remains unmatched against NFC Hangul titles. This is unchanged from master (which emitted no tokens at all for conjoining Jamo) and consistent with `cjk.ts`'s documented scope, which lists compatibility Jamo as out of scope.
- Han Ext-A is no longer treated as CJK by this module — see the Fix section. Deliberate, tested, and reversible if maintainers want by-mention to keep its wider class.
- `tokenizeForScan` and `tokenizeTitle` are now exported, solely so tests can assert on tokenization.
- Follow-up, analogous to the CJK commit's Chinese verb regexes: Vietnamese link-type inference (`sáng lập`->founded, `đầu tư`->invested_in, `cố vấn`->advises, `làm việc tại`->works_at) + date format `ngày D tháng M năm YYYY`. Not in this PR.

### DIFF
--- a/src/core/by-mention.ts
+++ b/src/core/by-mention.ts
@@ -107,7 +107,9 @@ export interface FindMentionsOpts {
 /**
  * Token-only tokenizer. Returns `[token, offset]` pairs.
  *
- * ASCII: each `[a-zA-Z0-9]+` run is a single token, lowercased.
+ * Word runs: each contiguous run of Unicode letters/marks/numbers (excluding
+ *   CJK) is a single token, lowercased. Covers ASCII and diacritic Latin
+ *   scripts like Vietnamese ("Nguyễn" → one token, not ["nguy","n"]).
  * CJK: each CJK character (Chinese/Japanese/Korean) is an individual
  *   token, lowercased. This allows the normal maximal-munch scan path
  *   to reach CJK gazetteer entries without a separate substring pass.
@@ -116,7 +118,14 @@ export interface FindMentionsOpts {
  * run) — single-word "Acme" lookup succeeds at offset 0; the trailing 's'
  * is harmless noise.
  */
-const TOKEN_RE = /[a-zA-Z0-9]+/g;
+// Word-run tokenizer. Matches contiguous Unicode letters/marks/numbers but
+// EXCLUDES CJK (Han, Hiragana, Katakana, Hangul), which keep going through the
+// per-character path in the walkers below (the excluded ranges mirror isCJK()
+// / cjkCharCount()). The `v` flag enables set subtraction. Latin-script
+// languages with diacritics (Vietnamese, etc.) now tokenize as whole words
+// instead of fragmenting on every accented character — e.g. "Nguyễn" is one
+// token, not ["nguy", "n"], and "Đà Nẵng" is ["đà", "nẵng"], not ["n", "ng"].
+const TOKEN_RE = /[[\p{L}\p{M}\p{N}]--[㐀-䶿一-鿿぀-ゟ゠-ヿ가-힯]]+/gv;
 
 interface ScannedToken {
   text: string;       // lowercase
@@ -198,8 +207,9 @@ function cjkCharCount(s: string): number {
 /**
  * Tokenize a page title for gazetteer insertion.
  *
- * ASCII titles: standard `[a-zA-Z0-9]+` tokenization, lowercased.
- * CJK titles (no ASCII content): split into individual characters —
+ * Word-run titles: Unicode letters/marks/numbers (excluding CJK) tokenization,
+ *   lowercased — ASCII plus diacritic Latin scripts (Vietnamese, etc.).
+ * CJK titles (no word-run content): split into individual characters —
  *   e.g. "纳瓦尔" → ["纳","瓦","尔"]. This allows normal multi-token
  *   maximal-munch matching to work with character-level CJK tokens
  *   produced by `tokenizeForScan`.
--- a/src/core/by-mention.ts
+++ b/src/core/by-mention.ts
@@ -109,21 +109,27 @@ export interface FindMentionsOpts {
  * The CJK character set this module treats as char-level, declared ONCE.
  *
  * `CJK_SLUG_CHARS` (src/core/cjk.ts) is the repo-wide single source of truth
- * — Han U+4E00–9FFF, Hiragana, Katakana, Hangul syllables. This module has
- * additionally always treated Han Extension A (U+3400–4DBF) as CJK in its
- * walkers, while cjk.ts scopes Ext-A out deliberately (see its header), so
- * Ext-A is named and appended here rather than widening the shared constant.
+ * — Han U+4E00–9FFF, Hiragana, Katakana, Hangul syllables — and this module
+ * now uses it verbatim.
+ *
+ * Note the deliberate behaviour change: the walkers here used to carry their
+ * own copy of the ranges that also covered Han Extension A (U+3400–4DBF),
+ * which cjk.ts scopes out repo-wide (see its header). Aligning on the shared
+ * constant means Ext-A characters are no longer treated as CJK by
+ * by-mention: they tokenize as word runs and, being a single sub-4-character
+ * token, an Ext-A-only entity title now falls below MIN_NAME_LENGTH instead
+ * of qualifying under MIN_CJK_NAME_LENGTH. Search, chunking and slug grammar
+ * already ignore Ext-A, so this makes by-mention consistent with them rather
+ * than being the one subsystem that disagrees.
  *
  * Everything below — TOKEN_RE, hasCJK(), cjkCharCount() and the two
- * per-character walkers — derives from this one declaration. There are no
- * other copies of the ranges in this file.
+ * per-character walkers — derives from this one import. There are no copies
+ * of the ranges in this file.
  */
-const HAN_EXT_A_CHARS = '㐀-䶿';
-const CJK_CHARS = `${CJK_SLUG_CHARS}${HAN_EXT_A_CHARS}`;
-const CJK_CHAR_RE = new RegExp(`^[${CJK_CHARS}]$`, 'u');
+const CJK_CHAR_RE = new RegExp(`^[${CJK_SLUG_CHARS}]$`, 'u');
 
 /**
- * Conservative code-point bounds for CJK_CHARS, derived from the range
+ * Conservative code-point bounds for CJK_SLUG_CHARS, derived from the range
  * string itself (strip the `-` separators and the remaining characters are
  * exactly the range endpoints) so they can never drift from it. Used only
  * as a cheap pre-filter — Latin/Vietnamese text short-circuits before the
@@ -132,7 +138,7 @@ const CJK_CHAR_RE = new RegExp(`^[${CJK_CHARS}]$`, 'u');
 const CJK_BOUNDS = ((): { min: number; max: number } => {
   let min = 0x10ffff;
   let max = 0;
-  for (const ch of CJK_CHARS.replace(/-/g, '')) {
+  for (const ch of CJK_SLUG_CHARS.replace(/-/g, '')) {
     const cp = ch.codePointAt(0)!;
     if (cp < min) min = cp;
     if (cp > max) max = cp;
@@ -172,12 +178,12 @@ function isCJKChar(ch: string): boolean {
  *    ADJACENT in the body — so a superscript between the words of
  *    "Acme Corp" would silently break a match that used to work.
  *  - Plain `u` flag, not `v`: the CJK exclusion is a negative lookahead
- *    over CJK_CHARS, the same construction src/core/think/gather.ts
+ *    over CJK_SLUG_CHARS, the same construction src/core/think/gather.ts
  *    already uses. No es2024 target requirement, no set-subtraction syntax.
  */
 const TOKEN_RE = new RegExp(
-  `(?![${CJK_CHARS}])[\\p{L}0-9]` +
-  `(?:(?![${CJK_CHARS}])[\\p{L}\\p{M}0-9])*`,
+  `(?![${CJK_SLUG_CHARS}])[\\p{L}0-9]` +
+  `(?:(?![${CJK_SLUG_CHARS}])[\\p{L}\\p{M}0-9])*`,
   'gu',
 );
 
--- a/src/core/by-mention.ts
+++ b/src/core/by-mention.ts
@@ -27,6 +27,7 @@
  */
 
 import type { BrainEngine } from './engine.ts';
+import { CJK_SLUG_CHARS } from './cjk.ts';
 import { stripCodeBlocks } from './link-extraction.ts';
 
 /** D2: hardcoded entity types for v1. Pack-aware extension is TODO-1. */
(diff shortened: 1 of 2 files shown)


time-attack:

The PR broadens by-mention's tokenizer from /[a-zA-Z0-9]+/g to a v-flag Unicode class minus CJK, which does fix Vietnamese name fragmentation. But including \p{M} and \p{N} in the class lets a token consist of only combining marks or only digits, and I reproduced two resulting regressions against master with the real findMentionedEntities: emoji-prefixed entity titles (VS16 is \p{Mn}) get their gazetteer key hijacked by a bare U+FE0F, losing plain-text links and cross-matching unrelated emoji; and newly-tokenizable glyphs between the words of an ASCII entity name break the matcher's strict token adjacency (5/5 cases matched on master, 0 on the PR). The discrimination test also fails: 7 of the 8 new tests pass on master's ASCII regex, because the old tokenizer fragments title and body symmetrically. Separately, the branch is 31 commits stale and predates merged #3522, so it hardcodes a 5th/6th copy of the CJK range instead of reusing CJK_SLUG_CHARS from src/core/cjk.ts, which src/core/think/gather.ts already composes an equivalent CJK-excluding Unicode tokenizer from without needing the v flag.

### Discrimination test

FAILS. Executed: restored /[a-zA-Z0-9]+/g in both src/core/by-mention.ts:128 and the test's duplicated copy at test/by-mention.test.ts:62 (true master state), then `bun test test/by-mention.test.ts -t "VN"` -> 7 pass / 1 fail (exit 1). Only the buildGazetteer token-content assertion fails on master; all 7 pure-fn VN tests pass without the fix. Restoring the PR source: 8 pass / 0 fail. Root cause of the hole is that master's ASCII tokenizer fragments the gazetteer title and the body symmetrically, so a 5-fragment entry still matches a 5-fragment body run, and the asserted `name` comes from the untouched `title` column rather than from tokens.

### Blockers

**1.** src/core/by-mention.ts:128 — \p{M} permits a mark-only token. VS16 (U+FE0F, category Mn) is carried by most common emoji, so an entity page titled '❤️ Health Notes' tokenizes to ["️","health","notes"] and the gazetteer key becomes a bare U+FE0F instead of 'health'. EXECUTED against production findMentionedEntities: master keys=["health"] and links body 'Plain health notes, no emoji.' -> 1 mention; PR keys=["️"] -> 0 mentions (link lost). Simultaneous false positive: PR links 'Sprint ⚠️ health notes were fine.' and 'See ⭐️ health notes below.' -> 1 mention each to companies/health-notes. All emoji-prefixed entity titles also collapse into a single shared U+FE0F bucket, making them mutually confusable. Fix: require a base letter/number to lead the run, e.g. [[\p{L}\p{N}]--CJK][[\p{L}\p{M}\p{N}]--CJK]* .

**2.** src/core/by-mention.ts:128 — \p{N} (Nd/Nl/No) and \p{M} mint tokens where master emitted none, and findMentionedEntities (src/core/by-mention.ts:366-399) requires gazetteer tokens to be STRICTLY ADJACENT in the body token array, so any such glyph landing between the words of a multi-word entity name breaks the match. EXECUTED with gazetteer title 'Acme Corp': bodies 'We met Acme¹ Corp today.' (U+00B9 No), 'Acme ½ Corp' (No), 'Acme １ Corp' (fullwidth Nd), 'Acme̶ Corp' (stray Mn), 'Acme ❤️ Corp' (VS16 Mn) each produce 1 mention on master and 0 on the PR — 5/5 new false negatives on pure-ASCII entity names. The PR's evidence table used only single-token controls (Unicity, Zalo), which structurally cannot detect this.

**3.** test/by-mention.test.ts:397-492 — DISCRIMINATION FAILURE: 7 of the 8 new tests pass on master. EXECUTED `bun test test/by-mention.test.ts -t "VN"` after restoring /[a-zA-Z0-9]+/g in BOTH src/core/by-mention.ts:128 and the test's duplicated copy at test/by-mention.test.ts:62 (true master state): 7 pass / 1 fail. Only 'VN person title enters gazetteer keyed on first diacritic-preserving token' (the buildGazetteer token-content assertion) discriminates. The test named 'regression: no diacritic fragmentation' passes WITH the fragmenting tokenizer, because master fragments title and body symmetrically (gazetteer ['nguy','n','v','n','c'] still matches the identical 5-fragment body run) and `name` is read from the untouched `title` column. Every assertion must check tokenization, not just the resolved mention.

**4.** src/core/by-mention.ts:128 — no NFC/NFKC normalization, so the fix is inert for the exact input it targets. EXECUTED: NFC gazetteer title 'Nguyễn Văn' vs NFD body -> 0 mentions, and NFD title vs NFC body -> 0 mentions, identical before and after the PR. src/core/search/title-match.ts:38 already does .normalize('NFKC') and src/core/think/gather.ts:207 does .normalize('NFKD'), making by-mention the only tokenizer in the repo that skips normalization. Same gap re-appears for NFD Hangul, which the new regex tokenizes as one Jamo run while NFC Hangul goes char-level.

**5.** src/core/by-mention.ts:128 + test/by-mention.test.ts:62 — ignores the established single source of truth and the branch is stale. Branch is 31 commits behind origin/master and does NOT contain merged #3522 (#3417), which made src/core/cjk.ts (CJK_SLUG_CHARS) the single CJK character class; src/core/think/gather.ts:203 already composes a CJK-excluding Unicode tokenizer from it with a plain `u`-flag negative lookahead, no `v` flag required. This PR hardcodes a 5th copy of the range in production and a 6th in the test, and its copy includes U+3400–4DBF, which src/core/cjk.ts:14-16 documents as deliberately out of scope. The duplicated test copy is the direct cause of the 7/8 non-discriminating tests.

**6.** src/commands/extract.ts:1941 — no cleanup for the corruption the PR documents. engine.addLinksBatch is additive; mentionsFingerprint's gazetteerHash (src/commands/extract.ts:1899-1902) does self-invalidate because the gazetteer keys change, so a re-scan happens, but the ~820 false-positive link_source='mentions' rows the PR's own evidence table reports remain in existing brains. Correct links get added alongside the garbage rather than replacing it, so no upgraded Vietnamese brain reaches the 440 figure the PR claims.

### Not verified

- Full `bun test` suite did not complete inside the 600s limit. Interim run showed 27 failures across 24 files (test/migrations-v99, fence-write, cycle-patterns-deadline-budget, eval-contradictions-cache, minions-lease-full-retry, retrieval-reflex, eval-replay-gate, supervisor-wedge, extract-takes-model-resolution, chunk-grain-fts, think-pipeline.serial, mcp-eval-capture, facts-context-injection.serial, dream-postgres.serial, eval-chronicle, enrichment, extract-conversation-facts, facts-backstop-gating, takes-mcp-allowlist.serial, facts-doctor-shape, search-image-column, extract-source-aware, storage-pglite, doctor-extract-atoms-backlog). All are timeout-shaped (5-11s durations, mostly '(unnamed)') and none touch by-mention, the tokenizer, or link extraction — the 7 directly related test files pass 98/98. I did NOT re-run these 24 files on origin/master, so 'pre-existing' is inference, not verification.
- `bun run verify` reported 3/32 failing (check:privacy, check:test-isolation, typecheck) all with rc=143 SIGTERM at the 120s parallel timeout. Each passed standalone (exit 0), so I treat verify as effectively 32/32 — but I did not fix the harness timeout or re-run the full parallel gate to green.
- Did not run the DATABASE_URL-gated e2e / engine-parity suite (no Postgres DATABASE_URL in this worktree). Not applicable to this diff (no engine or SQL changes), but unexecuted.
- Did not reproduce the PR's 1,264-page Vietnamese corpus numbers (820->440 Đà Nẵng, 271->97 Nguyễn). No access to that corpus; the claimed false-positive reduction is unverified, though the tokenizer-level mechanism is confirmed by probe.ts.
- Did not test whether the branch merges cleanly onto current master (31 commits behind, and #3522 rewrote src/core/cjk.ts's PAGE_SLUG_SEG/SLUG_WORD_CHARS). No textual conflict expected in by-mention.ts, but the rebase was not executed.
- Real-world frequency of emoji-prefixed titles specifically on LINKABLE_ENTITY_TYPES pages (person/company/organization/entity) is a judgement call, not measured. The mechanism and both failure directions are executed and confirmed; only the incidence rate is estimated.

- `gh pr view 3563 — exit 0`
- `gh pr diff 3563 > /tmp/pr3563.diff — exit 0 (180 lines, every hunk read)`
- `gh pr checkout 3563 — exit 0 (branch feat/vietnamese-diacritic-tokenizer @ 7f9e9afd)`
- `bun install — exit 0 (283 packages; worktree had no node_modules)`
- `bun run typecheck (tsc --noEmit) — exit 0`
- `bun test test/by-mention.test.ts test/extract-by-mention.test.ts — exit 0 (54 pass / 0 fail)`
- `DISCRIMINATION: perl -0pi -e reverted TOKEN_RE to /[a-zA-Z0-9]+/g in src/core/by-mention.ts:128 AND test/by-mention.test.ts:62; bun test test/by-mention.test.ts — exit 1 (41 pass / 1 fail)`
- `DISCRIMINATION: bun test test/by-mention.test.ts -t "VN" on reverted state — exit 1 (7 pass / 1 fail)`
- `bun test test/by-mention.test.ts -t "VN" on restored PR state — exit 0 (8 pass / 0 fail)`
- `bun run probe.ts (24-case tokenizer matrix: ASCII, VN NFC/NFD, Hebrew, Arabic+harakat, Persian ZWNJ, Cyrillic, Greek, Thai, Devanagari, Han, Han ExtA, Han ExtB SMP, Hiragana/Katakana, Hangul NFC/NFD, halfwidth katakana, fullwidth Latin, emoji ZWJ, keycap, mixed VN+CJK, combining-only lead, Roman numeral/fraction, VS16) — exit 0`
- `bun run probe2.ts against production findMentionedEntities, PR state then reverted state (9 cases) — exit 0 both; 5 adjacency regressions confirmed`
- `bun run probe3.ts against production findMentionedEntities, PR state then reverted state (VS16 gazetteer-key hijack) — exit 0 both; false negative + false positive confirmed`
- `bun run verify — exit 1 (29/32 pass; the 3 'failures' are rc=143 SIGTERM local 120s-timeout artifacts)`
- `bun run check:privacy standalone — exit 0`
- `bash scripts/check-test-isolation.sh standalone — exit 0 (1060 files scanned)`
- `bash scripts/check-test-real-names.sh standalone — exit 0`
- `bun test test/extract-by-mention.test.ts test/extract-by-mention-resume.test.ts test/op-checkpoint-mentions-fingerprint.test.ts test/backlink-count-mention-filter.test.ts test/by-mention.test.ts test/cjk.test.ts test/title-match.test.ts — exit 0 (98 pass / 0 fail)`
- `bun test (full suite) — did not finish inside 600s; interim 27 failures across 24 files, none in by-mention/tokenizer paths`
- `git rev-list --count HEAD..origin/master — 31; git log HEAD --grep=3417 — empty (#3522 absent from branch)`
- `node -e new RegExp('[[\\p{L}]--[a]]','v') — exit 0 (v-flag OK on node v24.16.0)`

</details>

---

time-attack:

Issue #2011, opened 2026-06-09
bug: extract --stale aborts on 'invalid input syntax for type json' — excerpt() slices surrogate pairs


garrytan-agents:
## Problem

`gbrain extract all --source db --stale` aborts hard with `invalid input syntax for type json` and commits zero links/timeline entries. On a 192K-page brain it died at ~1,550/61,300 pages every run; the stale watermark re-hits the same point so extraction stays wedged.

## Root cause

`excerpt()` in `src/core/link-extraction.ts` slices the link-context window by raw UTF-16 char index:

```ts
function excerpt(s, idx, width) {
  const half = Math.floor(width / 2);
  const start = Math.max(0, idx - half);
  const end = Math.min(s.length, idx + half);
  return s.slice(start, end).replace(/\s+/g, ' ').trim();
}
```

When a window boundary lands inside a surrogate pair (any non-BMP char — emoji 🏆/🌌, math alphanumerics, etc.), the slice keeps a lone surrogate (e.g. `U+D83C` with no `U+DC00`–`U+DFFF` partner). That flows into the link `context` field; the batch is serialized to JSONB for `jsonb_to_recordset(($1::jsonb)->'rows')` (the #1861 path) and Postgres rejects the whole batch on `::jsonb` — killing the entire run, not just the row.

Confirmed by instrumenting `executeRawJsonb`: the failing batch had `row[18].context` with a lone `U+D83C` sliced out of an emoji in a person page's profile text.

## Fix

Strip unpaired surrogate halves from the excerpt before returning:

```ts
return s.slice(start, end)
  .replace(/[\uD800-\uDBFF](?![\uDC00-\uDFFF])/g, '')
  .replace(/(?<![\uD800-\uDBFF])[\uDC00-\uDFFF]/g, '')
  .replace(/\s+/g, ' ').trim();
```

Verified: after patching, `extract --stale` ran cleanly through all 59,902 stale pages -> 13,735 timeline entries + 76 links committed (was dying at 1,550).

## Defense in depth (optional)

The batch-insert primitives already strip NUL (`buildTakeRows`); they should also strip lone surrogates centrally via a `sanitizeForJsonb()` over batch row string fields, so any future slicer can't re-trigger this jsonb class.

## Environment
- gbrain 0.42.37.0, Postgres (Supabase) engine, 192K-page brain. Reproduced every run until the excerpt() fix.

PR #3564, opened 2026-07-29
fix(chunker): measured hard-split budgets + token-aware capByChars — the follow-up invited in #3477's merge review

### LINKED ISSUE #3037: Chunker caps by characters (maxChars=6000) but embedders limit by tokens — CJK/dense content permanently unembeddable, silently, exit 0

paul-0320:
The follow-up invited in #3477's merge review — items (1) and (3), together. Refs #2826, #3037.

## What changed

**(1) The hard-split budget is measured, not guessed — and the rebuilt header is budgeted.** `splitToTokenBudget` now derives its slice size from each piece's own density — `floor(length × cap / estimateEmbedTokens(piece))` — and re-checks every slice (density varies inside a piece), recursing with a re-derived budget until each fits. The old fixed 3.5 chars/token assumption runs ~2.6 on URL-dense JSON, which is how 2,070–2,379-token slices escaped a 2,000 cap. `capOversizedChunks` additionally reserves the re-added structured header's tokens from the body budget: with the budget fix alone, body-capped pieces re-emerged with the `[JSON] fence.json:…` header pushing totals ~11–60 tokens back over. The pinned contract is now strict — emitted chunk, header included, ≤ cap, no slack.

**(3) The estimators moved below both chunkers; `capByChars` now bounds tokens too.** One deviation from the review's suggestion, forced by your own guard: cjk.ts is a `check:fuzz-purity` bundle target, and tiktoken's loader pulls `node:fs` into the bundle — `bun run verify` fails with the estimators there. They land in `src/core/chunkers/token-estimate.ts` instead: same position in the import graph (below both chunkers, reusing cjk.ts's char classes), no cycle, and code.ts re-exports both so every existing import site is unchanged (pinned by test).

With the estimator reachable, `capByChars` enforces `DEFAULT_MAX_CHUNK_TOKENS` (2000, shared with `capOversizedChunks`) alongside `maxChars`: when text over-runs the token budget the window shrinks to `floor(length × budget / estimate)`, every slice is re-checked for local density, and the overlap formula keys off the effective window. ASCII prose is untouched — 6000 chars measure ~1.5–1.7k cl100k tokens, under budget, so the window stays `maxChars`.

## Measured — the three failure shapes, before → after (estimate / chunks over 2,000)

| shape | before | after |
|---|---|---|
| URL-dense ASCII json fence, 14.4K chars (the review's leak) | max 2,379 · 2 over | **max 2,000 · 0 over** |
| URL-dense Korean markdown at defaults (#2826's markdown repro) | max 4,295 · 3 of 3 over | **max 2,000 · 0 over** |
| low-density bilingual table (#3037's shape — CJK below the density threshold, 6,000-char windows) | max 4,344 · 2 over | **max 2,000 · 0 over** |

Real-tokenizer spot-check (Qwen3-Embedding family, the strict backend from #2826): the worst emitted chunk after the fix measures **1,720** (Korean markdown) and **1,478** (bilingual table) real tokens — under the 2,048 strict context with margin. One honest boundary: the worst pure-ASCII fence chunk measures 2,318 real Qwen3 tokens against its 2,000 cl100k estimate — that is the ASCII short-circuit accepted in #3477 (cl100k as the ASCII base), unchanged here. Strict-backend operators can lower `maxChunkTokens`; making the ASCII base provider-aware is exactly the #2847/#3074 territory this PR deliberately stays out of.

## Scope notes

- Closes the **CJK half** of #3037 (char cap guarding a token limit). The pure-ASCII dense case is **not** closed: the worst ASCII fence chunk still measures 2,318 real Qwen3 tokens against its 2,000 cl100k estimate, so it stays unembeddable on a 2,048-context backend. That is the ASCII short-circuit accepted in #3477, and making the ASCII base provider-aware stays #2847/#3074 territory. (Corrected from "closes #3037's core problem" after the pre-merge review read the self-disclosed 2,318 figure back to me — it was an overclaim.) #3037's problems 2 (one failed chunk darkening the whole page's embeds) and 3 (`embed --stale` exiting 0 on permanent failures) are separate machinery, untouched here.
- No `CHUNKER_VERSION`/`MARKDOWN_CHUNKER_VERSION` bump: output changes only for chunks that are over-budget today (unembeddable on the affected backends), consistent with the #3477 merge note. Happy to add bumps if you'd rather force convergence on next sync.

## Tests

`test/chunkers/embed-estimate-home.test.ts` (6): re-export identity (import sites unchanged); strict headers-included cap on the ASCII fence; the #2826 markdown repro under budget at defaults; the #3037 bilingual shape under budget; mixed-density slice re-check; ASCII prose passthrough (single chunk, verbatim). All three cap tests fail on master.

## Verification

- `bun run verify`: 32/32 (including `check:fuzz-purity` — the reason for the token-estimate.ts home). `bun run test:slow`: 41/41.
- Chunker + adjacent domain suites (import-file, version gate, sync-cost-preview, token-budget, diarize payload-fitter): 168 pass / 0 fail. Typecheck clean.
- E2E: skipped locally (no `DATABASE_URL`), per the documented `test:full` skip path — text-pipeline change only, no persistence or retrieval trigger paths touched.

### DIFF
--- a/src/core/chunkers/code.ts
+++ b/src/core/chunkers/code.ts
@@ -20,7 +20,13 @@
 
 import { chunkText as recursiveChunk } from './recursive.ts';
 import { buildQualifiedName } from './qualified-names.ts';
-import { CJK_SLUG_CHARS, CJK_RANGES_REGEX } from '../cjk.ts';
+import { estimateTokens, estimateEmbedTokens, DEFAULT_MAX_CHUNK_TOKENS } from './token-estimate.ts';
+
+// Both estimators moved to token-estimate.ts (#3477 follow-up) so
+// recursive.ts can share them without an import cycle. Re-exported here:
+// commands/sync.ts, commands/reindex-code.ts, and tests import them from
+// this module.
+export { estimateTokens, estimateEmbedTokens } from './token-estimate.ts';
 
 // Embed the tree-sitter runtime + per-language grammars as files.
 // `with { type: 'file' }` returns a path (string) at runtime. Bun bundles
@@ -559,7 +565,6 @@ export function parseWithTimeout(
 }
 
 const DEFAULT_CHUNKER_TIMEOUT_MS = 30_000;
-const DEFAULT_MAX_CHUNK_TOKENS = 2000;
 
 function resolveChunkerTimeoutMs(): number {
   const raw = process.env.GBRAIN_CHUNKER_TIMEOUT_MS;
@@ -851,9 +856,14 @@ function capOversizedChunks(
       continue;
     }
     // Strip the structured header ("[Lang] path:N-M symbol\n\n") so the splitter
-    // works on the raw body; buildChunk re-adds a header to each piece.
-    const body = c.text.replace(/^\[[^\]]+\] [^\n]+\n\n/, '');
-    for (const piece of splitToTokenBudget(body, cap, opts)) {
+    // works on the raw body; buildChunk re-adds a header to each piece. The
+    // re-added header costs tokens too — budget for it, or every piece split
+    // to exactly `cap` re-emerges a header's-worth over it (measured: a 2,000
+    // cap emitted 2,011-token fence chunks when the body alone was capped).
+    const headerMatch = c.text.match(/^\[[^\]]+\] [^\n]+\n\n/);
+    const body = headerMatch ? c.text.slice(headerMatch[0].length) : c.text;
+    const bodyCap = Math.max(1, cap - (headerMatch ? estimateEmbedTokens(headerMatch[0]) : 0));
+    for (const piece of splitToTokenBudget(body, bodyCap, opts)) {
       if (!piece.trim()) continue;
       out.push(buildChunk({
         body: piece,
@@ -880,42 +890,27 @@ function splitToTokenBudget(text: string, cap: number, opts: CodeChunkOptions): 
     chunkSize: opts.fallbackChunkSizeWords ?? 300,
     chunkOverlap: opts.fallbackOverlapWords ?? 50,
   }).map((p) => p.text);
-  for (const piece of pieces) {
-    if (estimateEmbedTokens(piece) <= cap) {
-      out.push(piece);
-      continue;
+  // Hard-split budget is derived from each piece's own measured density
+  // (chars per estimated token) scaled to the cap, not a fixed chars-per-
+  // token guess: the previous 3.5 chars/token ASCII assumption undercuts
+  // URL-dense JSON (~2.6 chars/token measured), leaving 2,070–2,095-token
+  // slices past a 2,000 cap. Slices are re-measured and re-derived (density
+  // varies within a piece), so the cap holds by construction.
+  const hardSplit = (p: string): void => {
+    const est = estimateEmbedTokens(p);
+    if (est <= cap) {
+      out.push(p);
+      return;
     }
-    // Hard-split slice size. Pure-ASCII pieces: ~3.5 chars/token is a
-    // conservative cl100k estimate for source text. CJK-containing pieces:
-    // the weighted estimate can reach 1 token/char, so budget 1 char/token
-    // to keep every slice under cap by construction.
-    const charBudget = Math.max(1, Math.floor(cap * (CJK_RANGES_REGEX.test(piece) ? 1 : 3.5)));
-    for (let i = 0; i < piece.length; i += charBudget) out.push(piece.slice(i, i + charBudget));
-  }
+    const charBudget = Math.max(1, Math.floor((p.length * cap) / est));
+    if (charBudget >= p.length) {
+      out.push(p); // 1-char floor on a tiny cap — nothing left to split
+      return;
+    }
+    for (let i = 0; i < p.length; i += charBudget) hardSplit(p.slice(i, i + charBudget));
+  };
+  for (const piece of pieces) hardSplit(piece);
   return out;
-}
-
-const CJK_CHARS_G = new RegExp(`[${CJK_SLUG_CHARS}]`, 'g');
-
-/**
- * Embedding-safe token estimate for the oversize cap. estimateTokens
- * (cl100k) matches embedding-family tokenizers closely on pure-ASCII source
- * (measured identical on English prose and JSON vs Qwen3-Embedding), but
- * UNDERCOUNTS mixed CJK+ASCII chunks — measured −31% on URL-dense Korean
- * text vs the Qwen3 embedding tokenizer, which is exactly the shape that
- * overflows strict embedding backends (#2826). For chunks containing CJK,
- * take the max of cl100k and a per-char-class overestimate (CJK 1.0/char,
- * other non-whitespace 0.75/char, whitespace 0.1/char). CJK-DOMINANT text
- * is unaffected too: cl100k already counts it above the weighted form, so
- * max() returns the same value as today. Only mixed-script chunks — the
- * measured divergence class — estimate higher.
- */
-export function estimateEmbedTokens(text: string): number {
-  const cjk = (text.match(CJK_CHARS_G) || []).length;
-  if (cjk === 0) return estimateTokens(text);
-  const ws = (text.match(/\s/g) || []).length;
-  const weighted = Math.ceil(cjk + (text.length - cjk - ws) * 0.75 + ws * 0.1);
-  return Math.max(estimateTokens(text), weighted);
 }
 
 // ---------- Internals ----------
@@ -1243,54 +1238,6 @@ function normalizeSymbolType(type: string): string {
 
 function sanitize(name: string): string {
   return name.replace(/[\n\r\t]+/g, ' ').replace(/\s+/g, ' ').trim();
-}
-
-// v0.19.0 (Layer 5): accurate token count via @dqbd/tiktoken cl100k_base,
-// the same encoder text-embedding-3-large uses. The old len/4 heuristic was
-// 2-3x off for code. Lazy-init so dev and compiled-binary both only pay
-// the init cost once. Falls back to the heuristic if the encoder fails
-// to load (vanishingly unlikely but keeps the chunker available).
-let tiktokenEncoder: { encode: (s: string) => Uint32Array; free: () => void } | null = null;
-let tiktokenInitialized = false;
-
-// v0.20.0 Cathedral II Layer 8 (D1) — exported so commands/sync.ts can
-// estimate embed cost before a --all sync blows a surprise OpenAI bill.
-// Same cl100k_base tokenizer the embedding path actually uses, so cost
-// estimates match actual billing within tokenizer noise.
-export function estimateTokens(text: string): number {
-  if (!text) return 0;
-  if (!tiktokenInitialized) {
-    try {
-      // eslint-disable-next-line @typescript-eslint/no-var-requires
-      const m = require('@dqbd/tiktoken');
-      tiktokenEncoder = m.get_encoding('cl100k_base');
-    } catch {
-      tiktokenEncoder = null;
-    }
-    tiktokenInitialized = true;
-  }
-  if (tiktokenEncoder) {
-    try {
-      return tiktokenEncoder.encode(text).length;
-    } catch {
-      // Code legitimately contains tiktoken special-token strings (e.g. CLIP/GPT
-      // tokenizers embed the literal "<|endoftext|>"). The default encode() uses
-      // disallowed_special='all' and THROWS on those, crashing reindex-code on
-      // valid source files. For a token COUNT we don't need special-token
-      // semantics: re-encode treating them as ordinary text (never throws),
-      // heuristic only if even that fails.
-      try {
-        return (
-          tiktokenEncoder as unknown as {
-            encode: (s: string, allowed: string[], disallowed: string[]) => Uint32Array;
-          }
-        ).encode(text, [], []).length;
-      } catch {
-        return Math.max(1, Math.ceil(text.length / 4));
-      }
-    }
-  }
-  return Math.max(1, Math.ceil(text.length / 4));
 }
 
 // v0.20.0 Cathedral II Layer 4: display name derived from the language
--- a/src/core/chunkers/code.ts
+++ b/src/core/chunkers/code.ts
@@ -20,7 +20,8 @@
 
 import { chunkText as recursiveChunk } from './recursive.ts';
 import { buildQualifiedName } from './qualified-names.ts';
-import { estimateTokens, estimateEmbedTokens, DEFAULT_MAX_CHUNK_TOKENS } from './token-estimate.ts';
+import { estimateTokens, estimateEmbedTokens, estimateEmbedTokensCeiling, DEFAULT_MAX_CHUNK_TOKENS } from './token-estimate.ts';
+import { safeSplitIndex } from '../text-safe.ts';
 
 // Both estimators moved to token-estimate.ts (#3477 follow-up) so
 // recursive.ts can share them without an import cycle. Re-exported here:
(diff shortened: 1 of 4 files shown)


time-attack:

The PR fixes a real and severe leak — on master `scripts/test-weights.json` chunks at maxEst 2059 and a minified ASCII blob at maxEst 6011 against a 2000 cap, and the PR drives both to 0 over-cap. Its own tests genuinely discriminate (3/6 fail with master's behavior stubbed back in). But it also introduces three new defects I reproduced on files in this repo: it emits over-cap chunks on `src/core/migrate.ts` where master emitted none (header-reservation arithmetic is invalid because `estimateEmbedTokens` is super-additive for mixed-script text), it splits astral surrogate pairs at the new arbitrary-parity offsets (34/35 code chunks and 7/8 markdown chunks fail a UTF-8 roundtrip vs 0 on master) while the repo's own `text-safe.ts:safeSplitIndex` exists for exactly this, and it emits ~38-char remainder runt chunks (5 of 14 on `scripts/test-weights.json`). CI is 22/22 green and caught none of it.

### Discrimination test

PASSED — the PR's tests are genuinely discriminating. I stubbed master's behavior back into the PR branch (restored master's char-only `capByChars`, master's unbudgeted `body`/`bodyCap = cap`, and master's fixed 3.5/1.0 chars-per-token `splitToTokenBudget` loop) while keeping token-estimate.ts and the re-exports so the imports still resolved, then ran `bun test test/chunkers/embed-estimate-home.test.ts`: EXIT=1, 3 pass / 3 fail — ASCII json fence received 2494 (expected ≤2000), Korean markdown 3807, bilingual table 4344. Restoring the PR code: EXIT=0, 6/6 pass. Caveat: the 4th cap test ('mixed-density input … every slice re-checked') PASSED against the stub, so that one test proves nothing and should be tightened or dropped.

### Blockers

**1.** src/core/chunkers/code.ts:865 — `bodyCap = cap - estimateEmbedTokens(headerMatch[0])` is invalid arithmetic: `estimateEmbedTokens` is SUPER-additive across the header/body join for mixed-script chunks. An ASCII header estimates via cl100k (59-char header = 17 tokens) but contributes ~0.75/char (~40 tokens) to the whole chunk's estimate once the body contains any CJK, so the reservation under-counts ~2.3x. EXECUTED: `chunkCodeText(readFileSync('src/core/migrate.ts'), 'src/core/migrate.ts')` — PR emits 2 chunks at est 2006 and 2023 against the 2000 cap (headerEst=17, bodyEst=1964, whole=2006); master emits 0 over-cap, maxEst 1439. This is a REGRESSION on a real repo file and it falsifies the PR body's headline claim ("The pinned contract is now strict — emitted chunk, header included, ≤ cap, no slack"). The PR's own tests miss it because the ASCII-fence test has no CJK (header cost == cl100k, additive) and the CJK tests go through capByChars, which has no header.

**2.** src/core/chunkers/recursive.ts:150 and src/core/chunkers/code.ts:910 — raw `.slice()` at the new derived, arbitrary-parity window offsets splits astral surrogate pairs, emitting chunk_text that is not well-formed UTF-16. EXECUTED, markdown: `chunkText('\u{20000}\u{20001}\u{20002}'.repeat(1500))` → PR window 1333 (odd), 8 chunks, 7 lone surrogates, 7/8 fail `Buffer.from(s,'utf8').toString('utf8') === s`; master window 6000/stride 5500 (even), 2 chunks, 0 lone. EXECUTED, code: a minified JSON one-liner of astral Han + emoji → PR 35 chunks / 34 corrupted; master 5 chunks / 0 corrupted. The repo already ships `src/core/text-safe.ts:safeSplitIndex` for precisely this ('the tier-3 hard-split fallback in the dream-cycle chunker … refuses to orphan a high surrogate'), and that file's docstring names the prior production incident this reintroduces: '#2011 — extract --stale died at ~1,550 pages because excerpt() raw-sliced a window boundary through an emoji' and 'a lone surrogate is rejected by Postgres inside a ::jsonb cast and aborts the WHOLE batch'. Worse, capByChars' own docstring (recursive.ts:135-137) still claims 'BMP-only safe (does not split astral surrogate pairs in practice)' — that guarantee rested on maxChars=6000/stride=5500 both being even and is now false. Fix: route both split loops through `safeSplitIndex`.

**3.** src/core/chunkers/code.ts:910 — `for (let i = 0; i < p.length; i += charBudget) hardSplit(p.slice(i, i + charBudget))` leaves `p.length mod charBudget` as a standalone runt chunk at every recursion level, and each runt is re-headered by buildChunk into its own embedding row. EXECUTED: `scripts/test-weights.json` — master 7 chunks / 0 runts; PR 14 chunks / 5 runts (~38 chars each). A pure-ASCII minified blob: master 7 clean chunks; PR 39 chunks of which 19 are 38-40 chars (lengths `2020,38,2018,40,2018,40,…`). Half the emitted chunks are near-empty fragments that get embedded and can match queries. One-line fix: even out the split (`Math.ceil(p.length / Math.ceil(p.length / charBudget))`).

**4.** test/chunkers/recursive.test.ts:180 'maxChars hard cap fires on whitespace-less CJK at chunkSize boundary' — PRE-EXISTING test breaks on a cold run. EXECUTED twice: `bun test test/chunkers/` → 10647ms, and `bun test test/chunkers/recursive.test.ts` → 6746ms, both exceeding bun's 5000ms per-test limit (EXIT=1); master runs the entire file in 427ms, 24/24. Passes once bun's WASM cache is warm, so it is a latent CI flake, not a hard fail. Root cause measured: `capByChars` now calls `estimateEmbedTokens(text)` unconditionally on the FULL chunk before any length check, and tiktoken is superlinear on CJK — `estimateTokens` on 2K CJK chars = 11ms vs 20K CJK chars = 1289–3702ms. On master that path was an O(1) length compare (0ms). A cheap `if (text.length <= maxChars && text.length < someCharFloor) return [text]` style pre-gate, or capping the estimator input, removes it.

### Not verified

- Full `bun test` unit suite — started but stopped before completion (~6150 lines in). The 10 failures visible at that point were all wall-clock timeouts in PGLite migrate-regression and concurrency tests on a machine under load from my own parallel probes, none in chunker or import paths, and the targeted 10-file chunker+adjacent run was 133/133. I cannot state a full-suite pass/fail count.
- E2E suite (`bun run test:e2e`) — not run, no DATABASE_URL in this worktree. Means the engine-parity e2e and the DATABASE_URL-gated JSONB parity tests were not exercised; the diff touches no engine or SQL code, so low risk, but the lone-surrogate-into-Postgres consequence of blocker 2 is therefore reasoned from text-safe.ts's documented #2011 incident rather than observed against a live Postgres.
- The PR's real-tokenizer numbers (Qwen3-Embedding 1720/1478/2318 tokens) — no Qwen3 tokenizer available locally. I could not confirm or refute them. If the self-disclosed 2318-real-vs-2000-estimated figure for the worst ASCII fence chunk is right, then #3037's ASCII-dense case (nomic-embed-text, 2048 ctx) remains unembeddable after this PR, so 'Closes #3037's core problem' overclaims — it closes the CJK class only.
- Whether the recursive.test.ts cold-run timeout would fire in the project's actual CI sharding (scripts/test-shard.sh). It did not fire on the PR's 22 green checks. I reproduced it twice locally on a cold cache; I did not reproduce it under the CI shard runner.

- `gh pr view 3564 — exit 0`
- `gh pr diff 3564 — exit 0 (465 lines, every hunk read)`
- `gh pr checkout 3564 — exit 0 (branch fix/embed-estimate-home-split-budget; 13 commits behind origin/master, mergeStateStatus CLEAN)`
- `gh pr view 3564 --json statusCheckRollup — 22/22 SUCCESS (green CI caught none of the blockers)`
- `bun install — exit 0`
- `bun run verify — exit 0 (32/32 pass, includes check:fuzz-purity)`
- `bun test test/chunkers/ — exit 1 (66 pass / 1 fail: recursive.test.ts CJK cap test timed out at 10647ms)`
- `bun test test/chunkers/recursive.test.ts — exit 1 (23 pass / 1 fail, 6746ms timeout) then exit 0 on warm re-runs`
- `bun test test/chunkers/recursive.test.ts with origin/master chunkers restored — exit 0 (24/24, 427ms)`
- `bun test test/chunkers/embed-estimate-home.test.ts — exit 0 (6/6 on PR)`
- `bun test test/chunkers/embed-estimate-home.test.ts with old behavior stubbed in — exit 1 (3 pass / 3 fail: 2494, 3807, 4344 vs cap 2000) [DISCRIMINATION TEST]`
- `bun test test/chunkers/ test/import-file.test.ts test/privacy-strip-and-forget.test.ts test/page-search-vector-overflow.test.ts test/chunker-version-insert-default.test.ts test/reindex.test.ts test/upgrade-reembed-prompt.test.ts — exit 0 (133/133 across 10 files)`
- `custom probes (written+run+deleted in the worktree): astral-surrogate UTF-8 roundtrip on both chunkers, PR vs master; runt-chunk census on src/core/operations.ts, src/core/migrate.ts, scripts/test-weights.json, PR vs master; estimateTokens superlinearity on 2K/20K CJK; chunkText wall-clock on 18K/120K/180K CJK + 182K EN pages, PR vs master — all exit 0`
- `bun test (full unit suite) — STOPPED at ~6150 lines / 10 failures, all timing-shaped (PGLite migrate v8/v9/v80/v81/v89 regression timers at 6-9s, hardenBrainRepo 8.4s, per-chunk synopsis concurrency 10.6s) on a loaded box; none chunker-related. Reported as UNVERIFIED.`

</details>

---

paul-0320:
**All four reproduced, all four fixed.** Each was reproduced against this branch's parent (`6136e139`) first, so the regression attribution is measured rather than assumed. Rebased onto current master; `bun run verify` 33/33.

| | parent | before this push | after |
|---|---|---|---|
| **1** `src/core/migrate.ts` over cap | 0 (maxEst 1,434) | **2** (2,006 · 2,023) | **0** (maxEst 1,412) |
| **2** astral round-trip failures | 0 / 0 | **7 of 8** md · **18 of 37** code | **0 / 0** |
| **3** runts on `scripts/test-weights.json` | 0 of 7 | **5 of 14** (50–86 ch) | **0 of 9** (min 584) |
| **4** `recursive.test.ts` | 427 ms, 24/24 | 6.7 s, **EXIT 1** | **365 ms, 24/24** |

**(1)** You were exactly right about super-additivity. New `estimateEmbedTokensCeiling` applies the weighted form unconditionally — a true upper bound for any body, since the weighted form is additive per char class. Measured marginal header cost on migrate.ts: 42 tokens actual vs 17 reserved.

**(2)** Both split loops now route through `text-safe.ts:safeSplitIndex`. The `capByChars` docstring's "BMP-only safe" claim is corrected — you were right that it rested on 6000/5500 both being even.

**(3)** Your one-line fix, with one extension it needed: `safeSplitIndex` backs a cut off by up to two units, and with a fixed width that drift accumulates into a fresh tail runt (all-astral blob: 4-unit chunks trailing 724-unit ones). The width is re-derived from the remaining length each step. I did **not** apply the same evening to `capByChars` — that loop overlaps rather than partitions, and shrinking its stride to land the last window flush against the end collapses successive windows into near-duplicates (measured: two 6,047-char chunks differing by 47).

**(4)** The window is now derived from a bounded-prefix density probe. Correctness is unchanged: every emitted slice is still measured exactly, and the one path that emits without a re-check confirms exactly first. Cold-run reproduction under `scripts/test-shard.sh` — still not done; this is fixed from the measured cause (2K chars 11 ms → 20K 1,138 ms), not from a reproduced CI failure.

**The 4th cap test is dropped, not tightened.** You were right that it proved nothing, and the reason is worse than it looked: its dense run was whitespace-less CJK, which `countCJKAwareWords` scores per character, so the word pipeline had already cut it to ≤300-char pieces before the belt was ever consulted (parent: 20 dense chunks, max 300 chars). The belt's discriminating shape is the low-density one, already covered by the #3037 bilingual-table test.

**On the `Closes #3037` overclaim** — agreed, and I'll reword the PR body. The ASCII-dense case stands exactly as you read it: the CJK class is closed, the pure-ASCII fence is not, and making the ASCII base provider-aware is #2847/#3074 territory.

**Still not verified:** full unit suite (chunker + 6 adjacent domain suites: 138/138 across 10 files); E2E (no `DATABASE_URL`); the Qwen3 real-token figures (no local tokenizer).

time-attack:
Eval-replay gate (CONTRIBUTING.md retrieval requirement) — **PASS**. Because this PR moves chunk boundaries, the gate brain was **fresh-seeded with this PR's chunker** (init + import + embed all from the merged worktree) rather than copied: mean Jaccard@k **1.000**, top-1 stability **0.9167**, latency inside the master band — identical answer quality to the master-chunked baseline across 3 runs, zero regressed rows. Known remaining caveat from review (disclosed in your PR body): mixed-boundary corpora on already-embedded brains until a chunker-version bump forces re-chunking — worth a follow-up, not a gate failure. Queued for the next fix wave.

time-attack:

PR #3567, opened 2026-07-29
fix(entities): resolveEntitySlug fallback keeps the path separator (#3447)

### LINKED ISSUE #3447: resolveEntitySlug fallback corrupts well-formed slugs: slugify() strips the path separator (people/alice-example → people-alice-example)
## Summary

`resolveEntitySlug`'s slugify fallback **corrupts already-well-formed slugs**. When an extracted entity is a valid slug (`people/alice-example`) but no page exists at it yet, the fallback runs it through `slugify()`, whose `[^a-z0-9]+ → '-'` rule rewrites the path separator. The result — `people-alice-example` — is written to `facts.entity_slug` and can never resolve, because no page will ever have that slug.

It is also self-perpetuating: the corrupted slug never matches a page, so every subsequent extraction of the same entity re-mints it.

## Call chain

| Step | Location | Behavior |
|---|---|---|
| 1 | `src/core/facts/extract.ts:185-186` | Prompt instructs the model: *entity = a canonical slug (e.g. `"people/alice-example"`)… else a display name* |
| 2 | `src/core/facts/extract.ts:363` | Model's `entity` → `entity_slug`, unvalidated |
| 3 | `src/core/facts/backstop.ts:373-374` | `resolveEntitySlug(engine, sourceId, f.entity_slug)` |
| 4 | `src/core/entities/resolve.ts:53-56` | `looksLikeSlug('people/alice-example')` → true → `tryExactSlug` → no page → null |
| 5 | `src/core/entities/resolve.ts:65` | `isBareName` → **false** (line 87 rejects any input containing `/`) |
| 6 | `src/core/entities/resolve.ts:72` | `tryFuzzyMatch` → no hit |
| 7 | **`src/core/entities/resolve.ts:77`** | `return slugify(trimmed)` → **`people/alice-example` → `people-alice-example`** |

The corrupting line is `src/core/entities/resolve.ts:399`, inside `slugify()`:

```ts
.replace(/[^a-z0-9]+/g, '-')   // '/' is not [a-z0-9] → becomes '-'
```

## Root cause

The docstring at `src/core/entities/resolve.ts:36-37` states the assumption:

> *"Fall back to a deterministic slugify: lowercase-no-spaces with hyphen-collapse. NOT prefixed with a directory — caller decides whether to prefix `people/`, `companies/`, etc."*

The fallback was designed for **display names** ("Alice"). It also receives **already-slug-shaped input**, and for that input slugify is destructive rather than neutral. A slug that merely fails to resolve is turned into one that *cannot*.

## Why the stub guard doesn't catch it

`src/core/facts/backstop.ts:504-522` refuses to spawn a phantom unprefixed page, but explicitly routes the fact to the legacy DB-only path so it isn't dropped — *"the slug stays attached but no markdown file is created."* The malformed slug is preserved by design, so nothing surfaces in `gbrain doctor`.

## Impact

On a ~560-page single-source brain, **60 of 259 live facts (23%)** carried hyphen-flattened entity slugs. Combined with display-name fallbacks that produce unprefixed slugs, only **32% of live facts resolved to an existing page**. `facts_health` rated the brain `ok` throughout — and printed a flattened slug in its own status message.

Because `resolveEntitySlug` is also reused by signal-detector (per the module docstring), the corruption is not confined to the facts pipeline.

## Suggested fix

Pass slug-shaped input through untouched, so it stays fixable by creating the page it names:

```ts
// 4. Fallback: deterministic slugify.
if (trimmed.includes('/')) return trimmed;
return slugify(trimmed);
```

The same guard is needed in `resolveEntitySlugWithSource` (the `fallback_slugify` return), or the two resolvers disagree on the fallback.

Verified locally: after applying the guard and restarting the daemons, newly extracted facts emit `companies/<name>` where the previous build emitted `companies-<name>`. Brain-wide fact resolution went from 32% → 91% after the patch plus a backfill of the entities that had no page.

## Environment

- gbrain 0.42.66.0
- Engine: Postgres (Supabase) + pgvector
- Schema pack: `gbrain-base-v2@1.0.0`, schema version 125

time-attack:
## Summary

`resolveEntitySlug` / `resolveEntitySlugWithSource`'s slugify fallback corrupted slug-shaped input: `slugify()`'s `[^a-z0-9]+ → '-'` rule rewrote the path separator, so a well-formed but not-yet-created slug (`people/alice-example`) came back as `people-alice-example` — a slug no page can ever have. Self-perpetuating: the flattened slug never matches a page, so every re-extraction re-mints it (reporter measured 23% of live facts corrupted this way).

## Fix

One shared `fallbackSlugify()` used by both resolvers (root cause, single site): input containing `/` is slugified **per segment** — identity for already-well-formed slugs, normalizing for messy path-shaped input (`People/Alice Example` → `people/alice-example`). Display names (no `/`) keep the existing plain-slugify behavior.

## Discrimination proof

`test/entity-resolve-slug-fallback.test.ts` on unmodified master: **3 fail / 1 pass** — `companies/zeta-widgets-nonexistent` came back as `companies-zeta-widgets-nonexistent` (behavioral, not a missing export). With the fix: 4/4 pass. Neighboring suites (`entity-resolve`, `facts-backstop`, `facts-canonicality`, `think-entity-extract`, trajectory tests) all green.

`bun run typecheck` clean; `bun run verify` 32/32 green.

Closes #3447


time-attack:

The core fix is genuinely right: root cause in one shared helper (`fallbackSlugify`) used by BOTH resolvers, per-segment slugify (strictly better than the issue's suggested `return trimmed` — it also normalizes `People/Zeta Nonexistent`), no path-traversal exposure (`..` segments collapse to '' and are filtered; `/etc/passwd` → `etc/passwd`), and the discrimination test is real (3 of 4 new tests FAIL with the old behavior stubbed back in; the 4th is a deliberate display-name control). No engine/SQL/JSONB/source-scope/knobs-hash surface is touched. What blocks it is an undisclosed, verified second-order effect: keeping the separator flips the facts fence stub-guard from blocked → allowed for the fallback arm, so unresolved entities now create brand-new directories + stub .md files in the user's brain repo from LLM-controlled strings, with no prefix allowlist — and those pages are permanently exempt from the phantom-redirect cleanup (`page.slug.includes('/')` → `not_phantom`). Secondarily, there is no repair path for the already-flattened facts the issue reports (23% of live facts), and fact dedup buckets hard on `entity_slug`, so each corrupted fact gets re-inserted under the corrected slug — leaving an orphan/duplicate pair per fact.

### Discrimination test

EXECUTED. Replaced the body of `fallbackSlugify` with `return slugify(trimmed);` (old master behavior) and re-ran the PR's own test file: `bun test test/entity-resolve-slug-fallback.test.ts` → EXIT=1, 1 pass / 3 fail, with real behavioral diffs (Expected "people/zeta-nonexistent", Received "people-zeta-nonexistent") — not import/export errors. Restored the fix → 4 pass / 0 fail. The one test that passes on old behavior is the deliberate display-name control ('Zeta Nonexistent Q. Persson' → 'zeta-nonexistent-q-persson'), which is legitimate as a no-regression pin. So the tests do discriminate; the PR body's "3 fail / 1 pass on master" claim is accurate.

### Blockers

**1.** src/core/entities/resolve.ts:88-93 — stub-guard flip creates unreclaimable brain-repo litter. Before the fix the fallback arm ALWAYS produced a slash-free slug, so src/core/facts/fence-write.ts:203 (`if (!target.slug.includes('/'))`) blocked stub-creation and the fact went DB-only with no file. After the fix any slash-containing LLM `entity` value passes the guard and mkdir+writes a stub page. EXECUTED PROBE (test/zz-probe-fence.test.ts, 2 pass, since removed): slug `https-acme-example-com-blog` → `stubGuardBlocked: true`, no file; slug `https/acme-example-com/blog` → guard passes and `<brainDir>/https/acme-example-com/blog.md` is created. Verified fallback outputs for realistic LLM entity strings: 'https://acme.example.com/blog' → 'https/acme-example-com/blog', 'CEO/Founder' → 'ceo/founder', 'Q1/Q2 2026' → 'q1/q2-2026', 'C:/Windows' → 'c/windows', '~/secrets/key' → 'secrets/key' — each now mints a new top-level directory in ~/brain. src/core/facts/fence-write.ts:124-136 maps unknown prefixes to type 'concept' so nothing rejects them, and src/core/cycle/phantom-redirect.ts:389 returns `not_phantom` for any slug containing '/', so the existing self-healing pass will never reclaim them. Lazy fix: preserve the separator only when segment 0 is a known page directory (people/companies/deals/topics/meetings/notes/projects), else flatten as before — the reporter's verified win was `companies/<name>`, which that allowlist keeps.

**2.** src/core/entities/resolve.ts:88 (data-quality, no repair path) — issue #3447 reports 60/259 live facts already carry flattened slugs; the PR fixes new writes only. Dedup is a hard bucket filter (`WHERE source_id=$1 AND entity_slug=$2`, src/core/pglite-engine.ts:4419-4426), so on the next extraction each previously-corrupted fact is NOT seen as a duplicate and is re-inserted under `people/alice-example` while the orphaned `people-alice-example` row stays forever. Failing input: brain with an existing fact at entity_slug='companies-acme'; re-extract the same claim → second row at 'companies/acme', neither superseding the other. Needs a doctor check or one-shot repair mapping `<known-dir>-<tail>` → `<known-dir>/<tail>` (the issue reporter had to hand-backfill), or an explicit note that upgraders will see duplicates.

**3.** PR title violates the IRON RULE in CLAUDE.md: it must lead with vMAJOR.MINOR.PATCH.MICRO. Current title is `fix(entities): resolveEntitySlug fallback keeps the path separator (#3447)`. Fix with `gh pr edit 3567 --title "v0.42.66.1 fix(entities): ..."`. No VERSION/package.json/CHANGELOG bump in the diff either (only 2 files) — fine if /ship does the bump at release, flagged so it is not forgotten.

### Not verified

- Full unit suite, `bun run verify`, `bun run test:e2e`, and any Postgres-engine parity e2e — skipped per the execution-footprint rule (no engine or SQL changes in the diff)
- The end-to-end facts pipeline with a live LLM extractor: I did not observe a real model emitting a slash-containing junk entity. The stub-guard flip itself is executed and confirmed; the frequency of junk-prefixed entities in production is inferred from the extractor prompt (src/core/facts/extract.ts:185-186 explicitly asks for slug-shaped output and the value is stored unvalidated at extract.ts:363)
- The duplicate-fact consequence (blocker 2) is established by reading findCandidateDuplicates' hard `entity_slug = $2` filter, not by executing a before/after re-extraction against a brain seeded with flattened slugs
- signal-detector's reuse of resolveEntitySlug (named in the module docstring and in the issue) — I did not trace whether it also writes files from the fallback slug

- `gh pr view 3567`
- `gh pr diff 3567 > /tmp/pr3567.diff`
- `gh issue view 3447`
- `git fetch origin pull/3567/head:pr3567 && git checkout pr3567`
- `ln -s <main-worktree>/node_modules ./node_modules (worktree had none)`
- `bun test test/entity-resolve-slug-fallback.test.ts  → EXIT=0, 4 pass`
- `edit fallbackSlugify to old behavior; bun test test/entity-resolve-slug-fallback.test.ts  → EXIT=1, 1 pass / 3 fail; git checkout src/core/entities/resolve.ts`
- `bun test test/entity-resolve.test.ts test/facts-canonicality.test.ts test/facts-backstop.test.ts  → EXIT=0, 50 pass`
- `bun test test/think-trajectory-injection.test.ts test/relational-recall.test.ts test/doctor.test.ts  → EXIT=0, 111 pass`
- `bun /tmp/fb.ts (adversarial fallbackSlugify input probe: traversal, URLs, absolute paths, multi-slash)`
- `bun test test/zz-probe-fence.test.ts (temporary probe of the fence stub-guard flip)  → EXIT=0, 2 pass; file removed afterward`
- `grep -rn 'resolveEntitySlug' src/ test/ (all callers)`
- `grep -rn 'slugify(' src/ (sibling mint sites)`
- `grep -rn "includes('/')" src/core src/commands (downstream slash assumptions)`

</details>

---

time-attack:

PR #3569, opened 2026-07-29
fix(doctor): distinguish a damaged PGLite store from a broken runtime (#2674)

time-attack:
## What

The diagnostic half of #2674 — the `doctor` probe the reporter explicitly asked for. PR #3519 (open) fixed the messaging half; this PR adds the check that turns the ambiguous failure into a verdict.

**The load-bearing constraint:** PGLite reports only `Aborted()` to JS and prints the real PANIC (`could not locate a valid checkpoint record …`) to its own stderr. From the JS-visible string, a damaged store and a broken WASM runtime are byte-identical — which is how a store with WAL damage spent 17 days blamed on the macOS 26.3 bug. The scratch-store probe is the only thing that can tell them apart.

New `pglite_scratch_probe` check: initialize a throwaway store in a temp dir, write and read back a row, report:

- **scratch works** → the runtime is healthy; **your store is damaged**. Recovery: restore a backup or `gbrain reinit-pglite` — markdown is unaffected, the DB holds derived data a re-sync rebuilds.
- **scratch fails too** → the runtime cannot start on this machine; report OS + Bun versions on #223.

Verified end to end against a synthetically damaged store:

```
[FAIL] pglite_scratch_probe → A scratch PGLite store initialized, wrote and read back
fine on this machine (1.6s), so the runtime is healthy and YOUR STORE is damaged —
NOT the macOS WASM bug (#223), not a lock. Your markdown is unaffected: ...
Recover: restore a backup of the store directory, or run `gbrain reinit-pglite` ...
```

## Design decisions

- **Cost gate.** A PGLite cold start is 5–20s on loaded machines, so the probe never runs on a routine healthy `gbrain doctor`. It fires only when (a) PGLite init actually failed (engine=null, config engine is pglite, not `--fast`) — the moment the diagnosis is needed and its cost is trivially justified — or (b) on explicit `--probe-pglite`. Both gates are pinned by tests.
- **Cross-surface parity.** Wired into BOTH `buildChecks()` (with a new `doctor.pglite_probe` heartbeat phase so agents see life during the cold start — documented in `docs/progress-events.md`) and `doctorReportRemote()` (on the PGLite connection-failure short-circuit only, so the remote surface pays the cold start only when already broken). Registered in `BRAIN_CHECK_NAMES`; the categories drift guard passes.
- **Cleanup.** The scratch dir comes from `mkdtemp` under `os.tmpdir()` and is removed in a `finally` — success, failure, or guard refusal. Tests assert no leaked `gbrain-pglite-probe-*` dirs on both paths.
- **Never touch the real store.** Beyond the structural guarantee (mkdtemp under tmpdir), the probe refuses to run if the scratch path overlaps the configured store path in either direction, before any PGLite work. A guard refusal surfaces as a `warn` ("probe could not run"), never as a diagnosis.
- **Stderr capture: not done, deliberately.** PGLite 0.4.3 exposes no print/printErr hook on `PGliteOptions` (checked the shipped types: only `debug`, which still writes to the process's own stderr). Intercepting would mean monkey-patching `process.stderr.write` around an async WASM init while the doctor's own progress reporter writes to stderr — exactly the hack to avoid. The probe's ok/fail outcome carries the diagnosis instead; the JS-visible error is still fed to `classifyPgliteInitError` and surfaced in `details.verdict`, so once #3519's `wal-corrupt` verdict lands, any caller that does see richer text gets the precise classification for free.
- **Coordination with #3519.** Based on master, not on #3519's branch. The only shared file is `src/core/pglite-engine.ts`; this PR adds a standalone function between `preservingProcessExitCode` and the engine class — outside every hunk #3519 touches — so the two merge cleanly in either order, and none of #3519's messaging changes are duplicated. The probe intentionally uses `PGlite.create` directly rather than `PGLiteEngine.connect`, both for raw errors (connect wraps failures in `buildPgliteInitErrorMessage`, whose hint text would pollute re-classification) and to stay independent of #3519's message changes.

## Proof: tests fail on master behaviorally

New `test/doctor-pglite-scratch-probe.test.ts` imports only master-existing symbols at top level (`buildChecks`, `doctorReportRemote`). Against an unmodified master checkout (git archive of `origin/master` + the test file):

```
✗ PGLite init failure (engine=null, pglite config) → probe runs; ...
    expect(check).toBeDefined()  Received: undefined
✗ on demand: --probe-pglite with no init failure reports the runtime as healthy (ok)
    expect(check).toBeDefined()  Received: undefined
✗ a dead PGLite connection on the remote surface triggers the probe
    expect(check).toBeDefined()  Received: undefined
3 pass, 5 fail
```

The check is genuinely absent from both surfaces on master — not a missing-export failure. With this branch: 8 pass, 0 fail.

## Verification

- `bun run typecheck` clean
- `bun run verify` 32/32 green
- Related suites green: doctor-categories drift guard, pglite-init-classifier, doctor, doctor-behavioral, doctor-embedding-env-override, doctor-cli-smoke, doctor-scope-filter, sync-failures (220 tests)
- llms bundle freshness test green (no CLAUDE.md edits)

Closes the diagnostic half of #2674.

### DIFF
--- a/docs/progress-events.md
+++ b/docs/progress-events.md
@@ -134,6 +134,9 @@ sync that calls import emits `sync.import.<file>`, not `import.<file>`.
 Stable phase names shipped in v0.15.2:
 
 - `doctor.db_checks` (umbrella for all DB-side doctor checks)
+- `doctor.pglite_probe` (the #2674 scratch-store probe; only when PGLite init
+  failed or `--probe-pglite` was passed — a cold start can take 5–20s, so the
+  heartbeat is the only sign of life)
 - `orphans.scan`
 - `embed.pages`
 - `extract.links_fs`, `extract.timeline_fs`, `extract.links_db`, `extract.timeline_db`
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -2475,7 +2475,7 @@ SETUP
   migrate embeddings --to <p:model>  Re-embed onto another embedding provider
   upgrade                            Self-update
   check-update [--json]              Check for new versions
-  doctor [--json] [--fast]            Health check (resolver, skills, pgvector, RLS, embeddings)
+  doctor [--json] [--fast] [--probe-pglite]  Health check (resolver, skills, pgvector, RLS, embeddings; --probe-pglite runs the scratch-store probe)
   integrations [subcommand]          Manage integration recipes (senses + reflexes)
 
 PAGES
--- a/src/commands/doctor.ts
+++ b/src/commands/doctor.ts
@@ -627,6 +627,93 @@ export async function checkSourceConfigShape(engine: BrainEngine): Promise<Check
   }
 }
 
+/**
+ * #2674 — pglite_scratch_probe: distinguish a damaged PGLite store from a
+ * broken WASM runtime.
+ *
+ * PGLite reports only `Aborted()` to JS (the PANIC goes to its own stderr),
+ * so when init fails, the error string cannot say WHICH of the two it is.
+ * The probe initializes a throwaway store in a temp dir, round-trips a row,
+ * and reads the outcome:
+ *
+ *   - scratch works, real init failed → the runtime is fine; YOUR STORE is
+ *     damaged. Recovery: restore a backup or `gbrain reinit-pglite` (markdown
+ *     is unaffected — the DB holds derived data a re-sync rebuilds).
+ *   - scratch fails too → the runtime cannot start on this machine; report
+ *     OS + Bun versions on #223.
+ *
+ * COST GATE: a PGLite cold start is 5–20s on loaded machines, so this never
+ * runs on a routine `gbrain doctor`. It runs only when (a) the real PGLite
+ * engine actually failed to open (engine=null, not --fast, configured engine
+ * is pglite), or (b) the operator asks with `--probe-pglite`.
+ *
+ * `probeFn` is a test seam so message routing can be pinned without paying
+ * four real cold starts.
+ */
+export async function checkPgliteScratchProbe(opts: {
+  realInitFailed: boolean;
+  realStorePath?: string;
+  probeFn?: () => Promise<import('../core/pglite-engine.ts').PgliteScratchProbeResult>;
+}): Promise<Check> {
+  const name = 'pglite_scratch_probe';
+  try {
+    const probe =
+      opts.probeFn ??
+      (async () => {
+        const { probePgliteScratchStore } = await import('../core/pglite-engine.ts');
+        return probePgliteScratchStore(opts.realStorePath);
+      });
+    const r = await probe();
+    const secs = (r.duration_ms / 1000).toFixed(1);
+    if (r.ok) {
+      if (opts.realInitFailed) {
+        return {
+          name,
+          status: 'fail',
+          message:
+            `A scratch PGLite store initialized, wrote and read back fine on this machine (${secs}s), ` +
+            `so the runtime is healthy and YOUR STORE is damaged — NOT the macOS WASM bug (#223), not a lock. ` +
+            `Your markdown is unaffected: the DB holds derived data (chunks, embeddings, links, facts) that a re-sync rebuilds. ` +
+            `Recover: restore a backup of the store directory, or run \`gbrain reinit-pglite\` (wipes + re-inits + re-syncs).`,
+          details: { scratch_ok: true, duration_ms: r.duration_ms },
+        };
+      }
+      return {
+        name,
+        status: 'ok',
+        message: `PGLite runtime healthy: scratch store round-trip in ${secs}s.`,
+        details: { scratch_ok: true, duration_ms: r.duration_ms },
+      };
+    }
+    const errLine = (r.error ?? 'unknown error').split('\n')[0];
+    if (opts.realInitFailed) {
+      return {
+        name,
+        status: 'fail',
+        message:
+          `A fresh scratch PGLite store ALSO failed to start (${secs}s), so the WASM runtime cannot run ` +
+          `on this machine — your store is not necessarily damaged. Report your OS and Bun versions on ` +
+          `https://github.com/garrytan/gbrain/issues/223. Scratch error: ${errLine}`,
+        details: { scratch_ok: false, duration_ms: r.duration_ms, error: r.error, verdict: r.verdict },
+      };
+    }
+    return {
+      name,
+      status: 'warn',
+      message:
+        `Your real store opened, but a fresh scratch PGLite store failed to initialize (${secs}s) — ` +
+        `new stores can't be created on this machine. Report your OS and Bun versions on ` +
+        `https://github.com/garrytan/gbrain/issues/223. Scratch error: ${errLine}`,
+      details: { scratch_ok: false, duration_ms: r.duration_ms, error: r.error, verdict: r.verdict },
+    };
+  } catch (e) {
+    // Includes the never-touch-the-real-store guard refusal. The probe not
+    // running is a diagnostic gap, not a diagnosis — warn, don't fail.
+    const msg = e instanceof Error ? e.message : String(e);
+    return { name, status: 'warn', message: `scratch probe could not run: ${msg}` };
+  }
+}
+
 export async function doctorReportRemote(engine: BrainEngine): Promise<DoctorReport> {
   const checks: Check[] = [];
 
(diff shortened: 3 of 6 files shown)


time-attack:

The probe machinery works and is genuinely pinned (3 of 8 tests fail when the wiring is stubbed out — not the #3563 pattern), cleanup leaks nothing, typecheck is clean, and the category drift guard passes. But the diagnosis it prints is unsound and the remediation it prints does not run. The gate is `!engine && !fastMode && config.engine==='pglite'` (src/commands/doctor.ts:5793), and `engine === null` in `gbrain doctor` means "connectEngine threw for ANY reason, or the whole DB walk threw" (src/cli.ts:1479-1487, empty catch). I built a real, healthy PGLite store, round-tripped a row through it, handed its lock to a live `gbrain serve` holder, and ran the real CLI: doctor printed `[FAIL] pglite_scratch_probe → ... YOUR STORE is damaged — NOT the macOS WASM bug (#223), not a lock ... run \`gbrain reinit-pglite\``. The store was provably fine, the cause provably WAS a lock, the accurate LiveServeLockError text was swallowed, and the recommended recovery command exits 1 ("--embedding-model <provider:model> is required.") — the exact false-remediation defect round 1 failed #3519 for. Its full form wipes the store. This is the "false 'your store is damaged' panics users into deleting a good brain" scenario, reproduced end to end. Root-cause fix: thread the real connect error into buildChecks and only blame the store when that error is actually a PGLite abort/corrupt classification (the probe currently never looks at the real store at all), and print reinit-pglite with its required flags as src/core/pglite-engine.ts:222 already does.

### Discrimination test

RAN IT. Stubbed the old behavior back in by disabling only the two new wiring sites (`if (false && (probeRequested || pgliteInitFailed))` at src/commands/doctor.ts:5794 and `if (false && engine.kind === 'pglite')` at :738) while leaving the new exported functions importable, then re-ran the PR's own file: `bun test test/doctor-pglite-scratch-probe.test.ts` → EXIT=1, 5 pass / 3 fail. The 3 failures are exactly the three behavioral tests (buildChecks init-failure, --probe-pglite on demand, doctorReportRemote parity) — so the user-visible wiring IS genuinely pinned. The 5 that still pass are the 3 negative cost-gate tests (which pass vacuously on master, since a check that does not exist is trivially absent) plus the 2 probe-internals tests that exercise the still-present new functions. Restored the file (`git checkout src/commands/doctor.ts`, worktree clean). Verdict on this axis: PASSES the discrimination test — unlike #3563, the tests are not proving nothing. The defects are in what the tests assert, not whether they discriminate.

### Blockers

**1.** src/commands/doctor.ts:5793 + :675 — FALSE 'your store is damaged' verdict for every engine=null cause other than store damage. EXECUTED repro: healthy PGLite store at /tmp/pr3569-repro/brain.pglite (CREATE/INSERT/SELECT round-trip verified), lock dir planted with a LIVE `gbrain serve` holder (PID alive), config engine=pglite, then real CLI `bun src/cli.ts doctor --scope=brain` → exit 1 with `[FAIL] pglite_scratch_probe → ... so the runtime is healthy and YOUR STORE is damaged — NOT the macOS WASM bug (#223), not a lock. ... Recover: restore a backup ... or run `gbrain reinit-pglite` (wipes + re-inits + re-syncs).` The store is undamaged and the cause IS a lock; PGLiteEngine.connect acquires the lock (src/core/pglite-engine.ts:381) BEFORE PGlite.create (:409), so LiveServeLockError/lock-timeout never touches the store. cli.ts:1482-1484's catch discards that accurate message, so the only thing the user sees is the wipe advice. Same false verdict fires when the doctor DB walk itself throws mid-run (cli.ts comment at :1473 says both failure kinds fall back to runDoctor(null)) and on an embedding-dim-mismatch connect refusal.

**2.** src/commands/doctor.ts:677 (and the docstring at :640) — the remediation command as printed does not run. EXECUTED: `GBRAIN_HOME=/tmp/probe-home-x bun src/cli.ts reinit-pglite` → EXIT=1, stderr `--embedding-model <provider:model> is required.` (src/commands/reinit-pglite.ts:212-218 requires --embedding-model AND --embedding-dimensions, parsed before anything else). Identical class to round 1's #3519 rejection. The repo's existing corrupt-store hint already gets this right — src/core/pglite-engine.ts:222 prints `gbrain reinit-pglite --embedding-model <id> --embedding-dimensions <N>` — so this PR ships a regression against text already in the tree.

**3.** src/commands/doctor.ts:738 — same unsound inference on the remote surface. `doctorReportRemote` fires the probe whenever `engine.getStats()` throws and kind==='pglite'; getStats throws for missing schema/unrun migrations/transient WASM state too, and every such case now returns the 'YOUR STORE is damaged … reinit-pglite' fail to a thin-client/MCP caller (run_doctor op, scope=admin, localOnly=false — src/core/operations.ts:2690-2700). A remote admin also pays a fresh WASM PGLite instantiation per call on that path.

**4.** test/doctor-pglite-scratch-probe.test.ts:75 — the new test enshrines the false positive as the contract. `makePgliteHome()` writes a config whose `database_path` (`<tmp>/.gbrain/brain.pglite`) is NEVER created, passes engine=null, and then asserts the check says `YOUR STORE is damaged` and mentions `reinit-pglite`. Nothing is damaged in that fixture; the test locks in blaming a store that does not exist. Same at :141 for the remote surface, where the stub engine's failure mode is a thrown Error, not store damage.

**5.** PR title violates the version-first IRON RULE: `fix(doctor): distinguish a damaged PGLite store from a broken runtime (#2674)` has no leading vMAJOR.MINOR.PATCH.MICRO. Fix with `gh pr edit 3569 --title "vX.Y.Z.W fix(doctor): …"`.

### Not verified

- Did NOT run bun run verify / bun run test / ci:local or the full suite (execution-footprint rule). Ran only typecheck, the PR's own test file, and test/doctor-categories.test.ts.
- The mid-walk-throw lane of the same false positive (a check inside the DB walk throwing → cli.ts:1483 catch → runDoctor(null) → probe blames the store) is reasoned from the code and its own comment at src/cli.ts:1473-1477, not executed. The lock lane, which I did execute, is the same defect through the same catch, so the conclusion does not depend on it.
- Did NOT verify the message's claim that 'the DB holds derived data … that a re-sync rebuilds' is true for every DB-resident table (transcripts/facts/salience/calibration). The same wording already exists in src/core/pglite-engine.ts's corrupt hint, so if it is wrong it is pre-existing, not introduced here.
- No Postgres/Docker run — nothing in the diff touches Postgres.
- Did not test the merge with #3519 (the PR claims non-overlapping hunks in pglite-engine.ts); reviewed against master only.

- `gh pr view 3569`
- `gh pr diff 3569 > /tmp/pr3569.diff`
- `git fetch origin pull/3569/head:pr3569-local && git checkout pr3569-local`
- `bun install`
- `mkdir -p /tmp/probe-home-x && GBRAIN_HOME=/tmp/probe-home-x bun src/cli.ts reinit-pglite  → EXIT=1 '--embedding-model <provider:model> is required.'`
- `STORE=/tmp/pr3569-repro/brain.pglite bun repro-mkstore.ts  → 'store built + readable rows= [{"n":1}]' (real healthy PGLite store with vector+pg_trgm)`
- `bun repro-lockcase.ts  → plants a live `gbrain serve` lock on the healthy store, runs `bun src/cli.ts doctor --scope=brain` with GBRAIN_HOME set → exit 1, '[FAIL] pglite_scratch_probe → … YOUR STORE is damaged … not a lock … run `gbrain reinit-pglite`'`
- `bun test test/doctor-pglite-scratch-probe.test.ts  → EXIT=0, 8 pass / 0 fail (on branch)`
- `python3 (stub the two wiring sites to `if (false && …)`) ; bun test test/doctor-pglite-scratch-probe.test.ts  → EXIT=1, 5 pass / 3 fail ; git checkout src/commands/doctor.ts`
- `bun test test/doctor-categories.test.ts  → EXIT=0, 10 pass / 0 fail`
- `GBRAIN_HOME=/tmp/pr3569-repro/home bun src/cli.ts doctor --fast --probe-pglite --scope=brain  → '[OK] pglite_scratch_probe: PGLite runtime healthy: scratch store round-trip in 0.9s.'`
- `ls /tmp | grep -c gbrain-pglite-probe-  → 0 (no leaked scratch dirs)`
- `git diff --stat origin/master...HEAD`
- `grep -n 'acquireLock|PGlite.create|async connect' src/core/pglite-engine.ts ; sed -n 1440,1500p src/cli.ts ; sed -n 190,270p src/commands/reinit-pglite.ts ; sed -n 100,250p src/core/pglite-lock.ts`

</details>

---

time-attack:
Review finding (blocking, small fix): the substance is merge-lane — issue #2674 asked for exactly this scratch-store probe and master only has the classifier half. But the branch is based on a pre-guard master: it adds three unmarked runtime dynamic imports on an engine-live path (`src/core/pglite-engine.ts` ~310-312: `node:fs/promises`, `node:os`, `node:path`), and `scripts/check-engine-dynamic-import.sh` — now wired into `check:all`/`verify` — exits 1 on your head while master exits 0. Your 'verify 32/32 green' was measured before that guard landed.

The fix is also the smaller diff: make those three static top-level imports (they're node builtins; there's no lazy-load justification). Two more asks: rebase onto current master so CI runs against the real guard set, and consider a timeout on the probe — `doctorReportRemote` now pays a 5-20s WASM cold start on an agent-facing surface whenever a pglite connection is dead. Happy to merge once the guard passes.

time-attack:
Landed in the v0.45.10.0 community fix-wave 2 (#4085) as 2ae5d60b98, rebased/repaired against current master during wave assembly.

PR #3571, opened 2026-07-29
fix(sync): surface swallowed updateSlug failures in the rename path (#3056)

### LINKED ISSUE #3056: sync rename path: failed `updateSlug` is swallowed by an empty catch, leaving a duplicate row with no diagnostic

time-attack:
## Summary

`sync`'s rename loop caught `engine.updateSlug` errors with an empty `catch` and silently fell back to `importFile` — creating a new row for the new path while the old row stayed behind, live and orphaned. Nothing logged, nothing counted, page totals unchanged: the failed rename read as a clean rename. The reporter lost time to this twice and could not determine WHY updateSlug missed, because the error that would have explained it was discarded.

## Fix

Log the caught error to stderr with old slug, new slug, and the error message (`[sync] rename: updateSlug "a" → "b" failed (...); treating as add — the previous row may remain at "a"`). This matches the record-the-failure standard the neighboring adds/deletes loops in the same function already follow (their comments explicitly call it out). Fallback behavior is unchanged — this is the diagnostic the issue's suggestion #1 asked for.

## Discrimination proof

`test/sync-rename-updateslug-diagnostic.test.ts` drives a real git rename (R100) through `performSync` where the destination slug is already occupied — `updateSlug` hits the unique `(source_id, slug)` violation, the exact divergence class from the report. On unmodified master: **fails behaviorally** (no stderr line naming the slugs — pure silence). With the fix: passes.

Neighboring suites green (`sync-rename-batch`, `sync.test` — 73 pass). `bun run typecheck` clean; `bun run verify` 32/32 green.

Closes #3056

### DIFF
--- a/src/commands/sync.ts
+++ b/src/commands/sync.ts
@@ -2876,8 +2876,16 @@ async function performSyncInner(engine: BrainEngine, opts: SyncOpts): Promise<Sy
       const newSlug = resolveSlugForPath(to);
       try {
         await engine.updateSlug(oldSlug, newSlug, renameOpts);
-      } catch {
-        // Slug doesn't exist or collision, treat as add
+      } catch (e) {
+        // Slug doesn't exist or collision — fall back to add. #3056: the old
+        // row (if any) is NOT reconciled by this fallback, so a silent miss
+        // here reads as a clean rename while a duplicate row may remain.
+        // Surface the error so the incident is self-diagnosing.
+        console.error(
+          `[sync] rename: updateSlug "${oldSlug}" → "${newSlug}" failed `
+          + `(${e instanceof Error ? e.message : String(e)}); treating as add — `
+          + `the previous row may remain at "${oldSlug}"`,
+        );
       }
       // Reimport at new path (picks up content changes). Wrapped to match the
       // deletes/adds loops: a malformed renamed file is recorded to failedFiles
--- a/test/sync-rename-updateslug-diagnostic.test.ts
+++ b/test/sync-rename-updateslug-diagnostic.test.ts
@@ -1,0 +1,82 @@
+/**
+ * #3056 — sync's rename path: a failed engine.updateSlug was swallowed by an
+ * empty catch. The run then falls back to importFile (creating a NEW row for
+ * the new path) while the OLD row stays behind, live and orphaned — nothing
+ * logged, nothing counted, page total unchanged, so it reads as a clean
+ * rename.
+ *
+ * Behavioral pin: when updateSlug throws (here: rename destination slug
+ * already occupied by another row → unique (source_id, slug) violation), the
+ * sync run emits a diagnostic on stderr naming the old slug, the new slug,
+ * and the error, instead of silence.
+ */
+import { describe, test, expect, beforeAll, afterAll, afterEach } from 'bun:test';
+import { mkdtempSync, writeFileSync, rmSync, mkdirSync } from 'fs';
+import { execSync } from 'child_process';
+import { tmpdir } from 'os';
+import { join } from 'path';
+import { PGLiteEngine } from '../src/core/pglite-engine.ts';
+
+let engine: PGLiteEngine;
+let repoPath: string;
+
+beforeAll(async () => {
+  engine = new PGLiteEngine();
+  await engine.connect({});
+  await engine.initSchema();
+});
+
+afterAll(async () => {
+  await engine.disconnect();
+});
+
+afterEach(() => {
+  if (repoPath) rmSync(repoPath, { recursive: true, force: true });
+});
+
+describe('sync rename updateSlug failure diagnostic (#3056)', () => {
+  test('a swallowed updateSlug failure is reported to stderr, not silent', async () => {
+    const { performSync } = await import('../src/commands/sync.ts');
+
+    repoPath = mkdtempSync(join(tmpdir(), 'gbrain-rename-diag-'));
+    execSync('git init', { cwd: repoPath, stdio: 'pipe' });
+    execSync('git config user.email "test@test.com"', { cwd: repoPath, stdio: 'pipe' });
+    execSync('git config user.name "Test"', { cwd: repoPath, stdio: 'pipe' });
+    mkdirSync(join(repoPath, 'topics'), { recursive: true });
+    writeFileSync(join(repoPath, 'topics/old-name.md'), [
+      '---',
+      'type: concept',
+      'title: Old Name',
+      '---',
+      '',
+      'Stable body so git detects the move as a rename (R100).',
+    ].join('\n'));
+    execSync('git add -A && git commit -m initial', { cwd: repoPath, stdio: 'pipe' });
+
+    const first = await performSync(engine, { repoPath, noPull: true, noEmbed: true });
+    expect(first.added).toBe(1);
+
+    // Occupy the rename DESTINATION slug so updateSlug hits the unique
+    // (source_id, slug) constraint — the divergence class #3056 describes.
+    await engine.executeRaw(
+      `INSERT INTO pages (source_id, slug, source_path, type, title, compiled_truth, timeline, frontmatter)
+       VALUES ('default', 'topics/new-name', 'somewhere-else.md', 'note', 'Squatter', 'body', '', '{}'::jsonb)`,
+    );
+
+    execSync('git mv topics/old-name.md topics/new-name.md', { cwd: repoPath, stdio: 'pipe' });
+    execSync('git commit -m rename', { cwd: repoPath, stdio: 'pipe' });
+
+    const errLines: string[] = [];
+    const origError = console.error;
+    console.error = (...args: unknown[]) => { errLines.push(args.map(String).join(' ')); };
+    try {
+      await performSync(engine, { repoPath, noPull: true, noEmbed: true });
+    } finally {
+      console.error = origError;
+    }
+
+    const diagnostic = errLines.find(l =>
+      l.includes('updateSlug') && l.includes('topics/old-name') && l.includes('topics/new-name'));
+    expect(diagnostic).toBeDefined();
+  }, 60_000);
+});


time-attack:

The PR is one `console.error` in sync's rename catch plus a test. The test genuinely discriminates (verified: fails on unmodified master, passes with the line), and there is no hard-failure regression — the log does not throw, and the catch only fires on a real `updateSlug` throw (a zero-row UPDATE never enters it), so no per-file spam on the common benign path. But the PR is built on a base that PREDATES the already-merged #3479 (e72d93fd), which fixed the same issue #3056 substantively: master now observes the zero-row/throw cases and reconciles the stale row. Consequences: (1) the PR is CONFLICTING/DIRTY — the `catch {}` hunk context no longer exists; (2) the new operator-facing message asserts "the previous row may remain at <oldSlug>" while master's very next block deletes exactly that row and prints "removed stale row <oldSlug>" — I reproduced BOTH contradictory lines in one run; (3) it introduces the only raw `console.error` in performSyncInner's ~1600 lines (52 sibling calls use `serr`), which I empirically confirmed drops the per-source `[id]` prefix, so under `sync --all --parallel` the one line an operator needs attributed is the one line without a source tag; (4) the PR title has no vMAJOR.MINOR.PATCH.MICRO prefix. Residual value is real (master still discards the underlying error text — "duplicate key value violates unique constraint pages_source_slug_key" appears only with this patch), so this is a rebase + serr + reword, not a close.

### Discrimination test

EXECUTED. Copied test/sync-rename-updateslug-diagnostic.test.ts from the PR onto unmodified master (HEAD c6dc0adf) and ran it: FAILS — `expect(diagnostic).toBeDefined()` received undefined at test:80, 0 pass / 1 fail. Then hand-resolved the PR's console.error into master's post-#3479 catch and re-ran: 1 pass / 0 fail, `bun run typecheck` EXIT=0. So the test does discriminate for its own assertion (no existing line contains "updateSlug"). Separately I probed the underlying mechanism with a standalone script: engine.updateSlug('topics/old-name','topics/new-name') against an occupied destination genuinely throws `duplicate key value violates unique constraint "pages_source_slug_key"`, i.e. the test exercises the real (source_id, slug) collision it claims. Caveat: the assertion only requires a line containing "updateSlug" + both slugs — a message that omitted the error text (the PR's actual value) would still pass. Also verified on master that the substantive #3056 fix already runs: end DB state after the fallback is a single row at topics/new-name with the old row deleted, and master already prints "rename reconciled: removed stale row ..." — so the test's failure on master reflects the missing error-text line, not a missing fix.

### Blockers

**1.** src/commands/sync.ts (PR hunk @2876) — PR does not merge: `gh pr view` reports mergeable=CONFLICTING / mergeStateStatus=DIRTY. PR base is a8a3b6df, which predates e72d93fd (#3479, already on master) that rewrote this exact catch into `renameApplied = (await engine.updateSlug(...)) > 0` plus a reconcile block. The diff's context lines no longer exist. Failing input: `git merge origin/master` on fix/sync-rename-swallow-3056 conflicts in the rename loop.

**2.** src/commands/sync.ts:2884-2888 (post-rebase) — the new message contradicts the code 40 lines below it. Reproduced with a real git R100 rename into an occupied destination slug: the run now emits `[sync] rename: updateSlug "topics/old-name" → "topics/new-name" failed (duplicate key value violates unique constraint "pages_source_slug_key"); treating as add — the previous row may remain at "topics/old-name"` immediately followed by `[sync] rename reconciled: removed stale row topics/old-name (... fell back to add).` The row demonstrably does NOT remain (verified in DB: the old row is gone). Two contradictory stderr lines per rename failure is a worse diagnostic than one.

**3.** src/commands/sync.ts:2888 — uses `console.error` instead of `serr`. It is the ONLY raw console call inside performSyncInner (verified by awk over lines 1815-3400); every sibling diagnostic, including the neighboring reconcile messages, uses `serr`. src/core/console-prefix.ts's docstring names sync.ts as covered and says to call serr instead of console.error. Verified empirically: inside `withSourcePrefix('wiki', ...)`, serr emits `[wiki] serr line` while console.error emits an unprefixed `console.error line`. Failing input: `gbrain sync --all --parallel 4` with a rename collision in one source — the failure line interleaves with three other sources' output carrying no source attribution.

**4.** PR metadata — title is `fix(sync): surface swallowed updateSlug failures in the rename path (#3056)`. CLAUDE.md IRON RULE: every PR title MUST lead with vMAJOR.MINOR.PATCH.MICRO. Fix with `gh pr edit 3571 --title "vX.Y.Z.W fix(sync): ..."`.

**5.** src/commands/sync.ts:2856 — incomplete root cause. The same rename block has a second silent swallow 30 lines above: `try { m = await engine.resolveSlugsByPaths(batch, {sourceId: sid}) } catch { m = new Map(); }`, which then falls back to a path-derived `resolveSlugForPath(p)` guess for oldSlug — precisely the "path-derived fallback slug could name an unrelated row" hazard the reconcile comment below warns about. The PR's own stated standard ("record the failure, like the neighboring adds/deletes loops") is not applied to it, so a DB hiccup in slug resolution stays as silent as the bug being fixed.

### Not verified

- Did NOT run `bun run verify`, `bun run test`, `bun run ci:local`, or the full unit suite (execution-footprint rule). Only the PR's own test file plus `bun run typecheck` were run.
- No real-Postgres run: postgres-engine's updateSlug collision path was not exercised (no engine code changed, and the PGLite path reproduced the constraint violation, so a Docker Postgres was not warranted).
- Neighboring suites the PR claims green (sync-rename-batch, sync.test) were not re-run against the rebased resolution.
- OBSERVATION, not attributable to this PR: in my reproduction the already-merged #3479 reconcile deleted the ORIGINAL content row (id=1) and kept the pre-existing squatter row (id=2), overwriting the squatter's own content/source_path. The renamed file's content survives, but a manually curated page occupying the destination slug is silently destroyed. That may be the same class round 1 flagged on #3583; I did not investigate it further since it is master behavior, not a change in #3571.

- `gh pr view 3571`
- `gh pr diff 3571 > /tmp/pr3571.diff`
- `gh pr view 3571 --json headRefName,baseRefName,mergeable,mergeStateStatus`
- `git fetch origin pull/3571/head:pr3571`
- `git merge-base --is-ancestor e72d93fd a8a3b6df  # PR base PREDATES the #3479 reconcile fix`
- `git merge-base --is-ancestor e72d93fd origin/master  # master HAS it`
- `git log --oneline --all --grep=3056`
- `sed -n 2820,2960p src/commands/sync.ts`
- `grep -rn serr src/commands/sync.ts | head -5; sed -n 85,120p src/core/console-prefix.ts`
- `awk 'NR>=1815 && NR<=3400 && /console\.(error|log)\(/' src/commands/sync.ts  # only hit is the PR's new line`
- `git show pr3571:test/sync-rename-updateslug-diagnostic.test.ts > test/sync-rename-updateslug-diagnostic.test.ts`
- `bun test test/sync-rename-updateslug-diagnostic.test.ts > /tmp/t_master.txt 2>&1  # EXIT=1, 0 pass / 1 fail on unmodified master`
- `bun test test/sync-rename-updateslug-diagnostic.test.ts > /tmp/t_merged.txt 2>&1  # EXIT=0, 1 pass with the fix hand-merged onto master`
- `bun run typecheck > /tmp/tc.txt 2>&1  # EXIT=0`
- `bun run diag.ts  # probe: updateSlug THREW duplicate key ... pages_source_slug_key; full sync emits both the new line and master's 'rename reconciled: removed stale row'; end state = old row deleted`
- `bun run diag2.ts  # prefix check: ['[wiki] serr line','console.error line']`
- `git checkout -- src/commands/sync.ts && rm -f test/sync-rename-updateslug-diagnostic.test.ts diag.ts diag2.ts node_modules  # worktree restored clean`

</details>

---

time-attack:
Closing as already fixed. Master commit e72d93fd (PR #3479 — merged 27 seconds before this was opened, and co-authored by you, so the race is understandable) fixes #3056 at the root: `updateSlug` returns the moved-row count in both engines, and the sync rename path now reconciles the stale old row with positive source_path lookup, delete, stderr diagnostics on every fall-through, and a `<rename:…>` failure sentinel. This PR edits the exact catch block that commit rewrote, so post-v0.42.69.0 the diff no longer applies. Nothing further needed — the fix is live. Thanks!

PR #3572, opened 2026-07-29
fix(cycle): log swallowed lock.release() failures (#1470)

### LINKED ISSUE #1470: runCycle silently swallows lock.release() errors, stranding rows in gbrain_cycle_locks

time-attack:
## Summary

`runCycle`'s `finally` block swallowed `lock.release()` errors with an empty catch — and the composed PGLite file+DB lock handle had a second identical swallow one level deeper (`try { await dbLock.release(); } catch { /* fall through */ }`), so on PGLite the outer catch never even saw the error. When release throws (e.g. `CONNECTION_ENDED` after the pool was ended out from under the cycle), the row in `gbrain_cycle_locks` persists with zero operator-visible signal, and the next `gbrain dream` within the TTL skips with a phantom `cycle_already_running`.

## Fix

Both swallow sites keep release best-effort (a release failure never fails the cycle) but now emit a one-line stderr diagnostic naming the error and the stranded-lock consequence. No behavior change on the success path.

## Discrimination proof

`test/cycle-lock-release-diagnostic.serial.test.ts` simulates the pool-ended failure (the release `DELETE FROM gbrain_cycle_locks` rejects with `CONNECTION_ENDED (simulated)`; everything else passes through) and runs a real `runCycle`. On unmodified master: **fails behaviorally** — the cycle completes, the stranded lock row is confirmed present, and stderr contains nothing about it. With the fix: the diagnostic line appears; cycle status and the best-effort contract are unchanged (asserted).

Neighboring suites green (`cycle-lock-per-source`, `cycle-pglite-lock-ordering.serial`, `cycle-abort`). `bun run typecheck` clean; `bun run verify` 32/32 green.

Closes #1470


time-attack:

Logging-only change, and I confirmed it IS logging-only: both new catch bodies contain nothing but a `console.error`, the composed PGLite handle still calls `pgliteFileLock.release()` after the DB-release failure, and `report.status` is unaffected (the PR's own test asserts it, and 28 neighbouring lock/abort tests pass). stdout stays clean (stderr only), matching the existing precedent at src/core/cycle.ts:730. No engine/JSONB/source-scope/knobs-hash/migration/dry-run surface is touched. Two problems keep it off master as-is. (1) The discrimination test fails on half the diff: reverting ONLY the outer `finally` hunk (src/core/cycle.ts:2490-2500) back to master's `catch { /* best-effort */ }` leaves the new test GREEN (verified, exit 0) — and that unpinned hunk is precisely the site issue #1470 reports (postgres.js `CONNECTION_ENDED`; on Postgres `lock = dbLock` with no file lock, so the outer catch is the only diagnostic). The test drives PGLite-with-file-lock, so it pins only the inner composed hunk (reverting that one DOES fail the test, exit 1). (2) Partial root cause: three sibling swallows of the same failure class stay silent — src/core/cycle.ts:1642 and :1669 (`pgliteFileLock.release()`, a stranded file lock is arguably worse post-#2348 since a live PID is never reaped) and src/core/minions/supervisor.ts:656 (`lock.release()` on the same `gbrain_cycle_locks` table, same wait-out-the-TTL consequence). Patching 2 of 4+ call sites is arbitrary; the smaller fix is one shared best-effort-release helper (or log-and-rethrow inside both `DbLockHandle.release()` implementations in db-lock.ts, keeping engine parity) so every caller inherits the diagnostic.

### Discrimination test

Executed, three-way. (a) PR as-is: `bun test test/cycle-lock-release-diagnostic.serial.test.ts` → 1 pass / 0 fail, EXIT=0. (b) Reverted ONLY the outer finally hunk (cycle.ts:2490-2500) to master's empty catch, PR's inner hunk kept → still 1 pass / 0 fail, EXIT=0 — the test proves nothing about the hunk the issue is actually about. (c) Restored outer, reverted ONLY the inner PGLite composed hunk (cycle.ts:1697) → EXIT=1, fails at `expect(diagnostic).toBeDefined()`. Conclusion: the test pins exactly one of the two source hunks, and it is the secondary (PGLite) one. Source file restored via `git checkout src/core/cycle.ts`; worktree clean.

### Blockers

**1.** src/core/cycle.ts:2490 — the outer `finally` hunk is unpinned by any test. Stubbing master's `try { await lock.release(); } catch { /* best-effort */ }` back in and re-running test/cycle-lock-release-diagnostic.serial.test.ts gives 1 pass / 0 fail (EXIT=0). This is the exact path issue #1470 describes (Postgres engine: `lock = dbLock`, no file lock, outer catch is the only site), so the PR's headline fix ships with zero regression protection while the secondary PGLite hunk gets the pin.

**2.** src/core/cycle.ts:1642 and src/core/cycle.ts:1669 and src/core/minions/supervisor.ts:656 — same swallowed-release class left silent. supervisor.ts:656 releases the same `gbrain_cycle_locks` row via the same DbLockHandle; a throwing release there strands the singleton lock and the next supervisor start waits out the TTL with no signal, i.e. the identical operator-facing symptom the PR claims to fix. Failing input: any release that throws (pool ended / CONNECTION_ENDED) reached via the supervisor shutdown path or the two early-return file-lock releases → still zero stderr output.

**3.** PR title `fix(cycle): log swallowed lock.release() failures (#1470)` violates the CLAUDE.md IRON RULE that every PR title leads with `vMAJOR.MINOR.PATCH.MICRO`. Fix with `gh pr edit 3572 --title "vX.Y.Z.W fix(cycle): ..."`.

### Not verified

- Full unit suite, `bun run verify`, and E2E were NOT run (execution-footprint rule). Specifically not verified: whether any e2e test asserts exact stderr content for `gbrain dream` and would now see the extra line. Risk is low — the new output only fires on a release failure, never on the success path.
- Real-Postgres behavior of the outer catch (the issue's actual CONNECTION_ENDED path) was not exercised; no Docker Postgres was started. The outer hunk is 1 line of console.error and reads correctly, but nothing — in the PR or in my run — executes it.
- Whether the added console.error should honor `--quiet` (getCliOptions) was not tested; it does not, but neither does the pre-existing sibling at cycle.ts:730, so this is consistent-with-precedent rather than a regression.
- Not investigated further (pre-existing, not introduced here): DbLockHandle.release() calls `deregister()` from process-cleanup BEFORE the DELETE, so a throwing release also loses the process-exit cleanup safety net for that row.

- `gh pr view 3572`
- `gh pr diff 3572 > /tmp/pr3572.diff`
- `gh pr view 3572 --json headRefName,baseRefName,mergeable,mergeStateStatus,statusCheckRollup`
- `gh issue view 1470`
- `git fetch origin pull/3572/head:pr3572 && git checkout pr3572`
- `grep -rn 'gbrain_cycle_locks' src/`
- `grep -n 'release()' src/core/cycle.ts`
- `grep -rn 'catch { /\* best.effort' src/`
- `sed -n '180,320p' src/core/db-lock.ts`
- `sed -n '1590,1760p' src/core/cycle.ts`
- `sed -n '640,660p' src/core/minions/supervisor.ts`
- `grep -n 'serial' scripts/test-shard.sh`
- `bun install --frozen-lockfile (worktree had no node_modules)`
- `bun run typecheck > /tmp/tc3572.txt (EXIT=0)`
- `bun test test/cycle-lock-release-diagnostic.serial.test.ts (PR as-is: 1 pass, EXIT=0)`
- `bun test test/cycle-lock-release-diagnostic.serial.test.ts (outer hunk reverted to master: 1 pass, EXIT=0)`
- `bun test test/cycle-lock-release-diagnostic.serial.test.ts (inner hunk reverted to master: 1 fail, EXIT=1)`
- `bun test test/cycle-lock-per-source.test.ts test/cycle-pglite-lock-ordering.serial.test.ts test/cycle-abort.test.ts (28 pass, EXIT=0)`
- `git checkout src/core/cycle.ts && git status --porcelain (clean)`

</details>

---

time-attack:
**Merging this closes #1470** (severity **p2**).

#1470 was one of 58 issues carrying a `p3` label. All 58 were re-triaged by reproducing them against current master, and **the label was wrong on 56 of them** — the corrected distribution is 4 p0, 40 p1, 12 p2, and 2 genuinely p3. #1470 came out at **p2**, and this PR is what fixes it.

Two things follow from that:

- Add `Closes #1470` to the PR description if you want the link wired up — a closing keyword only fires from the body or a commit message, not from a comment like this one.

The reproduction and evidence are on #1470.

time-attack:

Issue #3521, opened 2026-07-28
binary-self-update downloads assets from the empty releases API, so notified upgrades cannot be applied


time-attack:
Surfaced while fixing #486 (see PR #3520) and deliberately left out of that PR's scope.

#3520 fixes upgrade **notification** by resolving the latest version from `raw.githubusercontent.com/.../master/VERSION`, because `api.github.com/repos/garrytan/gbrain/releases/latest` returns 404 — the repo has **zero** published releases:

```
gh api repos/garrytan/gbrain/releases --jq length  -> 0
gh release list                                    -> (empty)
only tag in the repo                               -> eval-run-v0.35.1.0-baseline
```

But `src/core/binary-self-update.ts` still downloads upgrade **assets** from that same empty releases API. So for binary installs the flow is now: correctly notice an upgrade exists → fail to apply it, because there is no release asset to fetch.

That's an improvement over the previous state (silence), but it's a dead end the user can now see.

**Two possible resolutions, and this is a maintainer call rather than a code question:**

1. **Start publishing GitHub releases** on the existing release train. `/ship` already bumps the version trio and writes a CHANGELOG entry per release, so a release + built binaries would be an additive CI step. This is the option that makes `binary-self-update.ts` correct as written.
2. **Point the binary path elsewhere** (build-from-source instructions, or a different artifact host) and make the notification say so explicitly for binary installs, instead of promising a self-update that cannot happen.

Until one of those lands, binary-install users get an accurate "upgrade available" notice with no working apply path. Worth being deliberate about which, because option 1 also fixes the `releases/latest` lookup that #486 had to route around in the first place.

Note the notification fix in #3520 is install-method agnostic, so this only affects the apply step.

PR #3573, opened 2026-07-29
feat(ci): publish GitHub releases so binary self-update can work (#3521)

time-attack:
Implements the maintainer decision on #3521: start publishing GitHub releases, so the binary self-update path (`src/core/binary-self-update.ts`) has assets to consume.

## The gap

`release.yml` already built and attested exactly the two assets the self-updater requests — but it only triggered on `v*` tag pushes, and in 283 version bumps no such tag was ever pushed. Result: zero releases, `releases/latest` a permanent 404, and binary installs that (post-#3520) correctly *notice* an upgrade but can never *apply* it (`no_asset`).

## What the self-updater actually expects (verified, not assumed)

- `GET https://api.github.com/repos/garrytan/gbrain/releases/latest` → `tag_name` + `assets[].browser_download_url`
- Raw (non-archived) binaries named exactly `gbrain-darwin-arm64` and `gbrain-linux-x64` (`expectedAssetName()`); every other platform deliberately degrades to notify-only
- No version comparison at swap time — it installs whatever `releases/latest` carries. This is why release cadence matters (below).

## Design

- **Trigger:** push to `master` touching `VERSION` (+ `workflow_dispatch` for the manual first run / repair). The workflow reads `VERSION` — the documented single source of truth — and never re-derives the version. No tag push in the loop; the release mints tag `v<VERSION>` at the pushed commit (`target_commitish: github.sha`).
- **Cadence — every bump, deliberately:** `check-update` resolves "latest" from `VERSION` on master, while self-update downloads from `releases/latest`. Any release that lags `VERSION` tells binary users an upgrade exists that self-update then can't deliver (or worse, swaps in an older binary). Selective releases reintroduce exactly the mismatch #3520 fixed. ~18 releases/week is list noise, but the releases page is cheap and the alternative is a correctness bug. CI cost is the real price (2 runners × test+verify+build per bump) — acceptable on a public repo; trim the in-workflow `bun test`/`verify` later if it hurts.
- **Idempotency + self-repair:** a `version` job skips build+release when a release for `v<VERSION>` already exists **with all expected assets**. Mere tag-exists or a half-published release (assets missing) does NOT short-circuit — softprops updates the existing release in place on re-run. Re-running the pipeline is always a no-op or a repair, never a duplicate or a failure. Racing bumps queue via a `release` concurrency group; a skipped intermediate version is fine since `latest` is what the updater reads.
- **Release notes:** `scripts/changelog-entry.sh` extracts the `## [X.Y.Z.W] - date` entry from `CHANGELOG.md` (exact-string header match, so 3-segment historical versions and 4-segment versions can't cross-match); falls back to a CHANGELOG link if the entry is missing rather than failing the release.
- **Permissions:** top-level `contents: read`; `contents: write` scoped to the release job only; build keeps `id-token`/`attestations` for the existing provenance attestation.
- **Guard rails:** `fail_on_unmatched_files: true` (a missing binary fails loudly instead of publishing an assetless release the updater then chokes on). Historical 3-segment tags are never rewritten.

## Drift guard (the ask)

`test/release-workflow.test.ts` pins the contract both ways: the workflow's build-matrix `artifact:` set must equal the set `expectedAssetName()` can return, every asset must appear in the release `files:` list and the idempotency completeness check, the tag must derive from `VERSION`, and `contents: write` must stay job-scoped. Plus fixture tests for the changelog extractor (top entry, historical entry, missing version, prefix non-match).

## Also

- Updated the now-stale "gbrain publishes NO GitHub releases" comment in `src/commands/check-update.ts` (check-update deliberately keeps reading `VERSION`, not `releases/latest`).
- `docs/RELEASING.md` gains a "GitHub releases" section; llms bundles regenerated (no byte change).

## Not done here (on purpose)

- **No release published from this PR** — the workflow only fires on master. First release: maintainer merges, then either lands the next VERSION bump or runs the `Release` workflow via `workflow_dispatch` on master.
- Unverifiable without publishing: the end-to-end softprops create/update behavior against the live repo, and a real binary swap from a published asset (unit + e2e swap tests cover the client side with injected fetchers).

Gates: `bun run typecheck` clean, `bun run verify` 32/32 green, new + related tests 56/56 pass, `actionlint` clean on the workflow.


time-attack:

The PR fixes a genuinely real bug — I confirmed `git ls-remote --tags origin` returns exactly ONE tag (`eval-run-v0.35.1.0-baseline`, zero `v*` tags) and `releases/latest` is a live 404, so 283 VERSION bumps really did publish nothing and binary self-update really can never apply. The design is sound (fork PRs cannot trigger it, all action SHAs pinned and verified against upstream tags, `contents: write` correctly job-scoped, idempotency fails open toward re-publish), and its test is a real discriminator: 6/10 tests fail when master's behavior is stubbed back in. But it ships one expression-injection sink into the single job that publishes the binaries users auto-download and execute: `v="${{ needs.version.outputs.version }}"` puts VERSION-file content straight into the release job's shell. I demonstrated arbitrary command execution locally with a space-free payload, exit code 0 (silent). Two-line fix; semgrep and actionlint both passed and neither catches it.

### Discrimination test

EXECUTED and it PASSES. I copied master's `release.yml` over the PR's and deleted `scripts/changelog-entry.sh`, then re-ran `bun test test/release-workflow.test.ts --timeout=60000`: exit 1, 4 pass / 6 fail — the idempotency asset-string check, tag-derives-from-VERSION, `fail_on_unmatched_files: true`, top-level `contents: read`, and both positive changelog-extraction tests all fail. Restored the PR files: exit 0, 10/10 pass. Caveat: 2 of the 10 tests ("exits non-zero for a version with no entry", "prefix does not false-match") pass vacuously when the script is absent (bash exits 127, which is also ≠ 0), so they do not discriminate — but the four positive assertions and the four workflow-shape assertions genuinely do.

### Blockers

**1.** `.github/workflows/release.yml:112` — GitHub Actions expression injection into the release job's shell, in the one job holding `contents: write`. The step is `v="${{ needs.version.outputs.version }}"`, and that output is the raw content of the `VERSION` file at the pushed master commit. `tr -d '[:space:]'` in the version job blocks GITHUB_OUTPUT injection (no newlines survive) but strips no shell metacharacters, and `${IFS}` removes the need for spaces. Verified by executing a local simulation of the generated step: because GitHub substitutes the expression textually *before* bash parses the line, a VERSION file whose contents close the quote and append a second command runs that command. Confirmed arbitrary execution with the step still exiting 0 and `echo "$v"` still printing a plausible version string, so nothing in the log looks wrong. (Payload withheld here; it is the standard Actions expression-injection shape and `${IFS}` removes the need for whitespace.) Failure input/outcome: a commit lands on master whose ONLY diff is the VERSION file — the highest-volume, least-scrutinized change shape in this repo (283 of them, and /ship generates them mechanically) — and it gets arbitrary code execution in the job authorized to create and overwrite the release assets that every binary install downloads and executes via `binary-self-update.ts`. That is the malicious-release-artifact outcome. Fix: `env:\n  V: ${{ needs.version.outputs.version }}` then `"$V"`, plus validate the format at the source in the version job (`case "$version" in *[!0-9.]*) exit 1;; esac`, which also enforces the CLAUDE.md 4-segment numeric invariant so `tag_name:`/`name:` cannot carry junk either).

**2.** `.github/workflows/release.yml:26` — `workflow_dispatch: {}` is not ref-gated. Failure input: a write-access holder dispatches Release from a feature branch; the version job reads that branch's VERSION and the release job publishes with `target_commitish: ${{ github.sha }}`, pointing `releases/latest` at a binary compiled from unreviewed branch code that every binary install then self-installs. docs/RELEASING.md says "via workflow_dispatch on master" but nothing enforces it. Fix: `if: github.event_name != 'workflow_dispatch' || github.ref == 'refs/heads/master'` on the version job.

### Not verified

- End-to-end release publish: softprops create-then-update-in-place, `target_commitish` behavior, `fail_on_unmatched_files` firing, and `releases/latest` actually resolving. Cannot be executed without publishing a real release to the 30k-star repo — the PR states this openly. The idempotency/self-repair claim is therefore reasoning, not evidence.
- `concurrency: group: release` queue-collapse behavior (3+ rapid VERSION bumps: GitHub cancels the intermediate PENDING run). Reasoned about, not observed. The PR's conclusion looks right — the surviving run reads VERSION at its own newer commit, so `releases/latest == VERSION` still holds — but I could not execute it.
- Full local `bun test` did not finish inside my window (>20 min; still enumerating at cutoff). Signal so far: 1 failure, `(unnamed) [64806ms] a beforeEach/afterEach hook timed out` — the known local slow-hook class from #3566, outside this diff's blast radius (the only runtime source change in the PR is a doc comment). Relying on CI's 10 green shards + serial-tests + Tier 1/2 for whole-suite coverage.
- Whether GitHub's `paths: [VERSION]` push filter reliably fires for the specific merge shape /ship uses (squash vs merge commit). Read the semantics; did not push to master to observe. If it misfires, the workflow silently never runs and the bug is unfixed — worth one `workflow_dispatch` smoke run before trusting it.

- `git fetch origin pull/3573/head:pr3573 && git checkout pr3573 — exit 0`
- `gh pr view 3573 / gh pr diff 3573 — exit 0 (328-line diff, read every hunk)`
- `git merge origin/master --no-edit — exit 0 (clean, no conflicts)`
- `bun install — exit 0`
- `bun run typecheck (tsc --noEmit) — exit 0, clean`
- `bun run verify — exit 0, `pass=32 fail=0 | all checks green``
- `bun test test/release-workflow.test.ts --timeout=60000 — exit 0, 10 pass / 0 fail`
- `DISCRIMINATION: master release.yml stubbed in + changelog-entry.sh removed, bun test test/release-workflow.test.ts — exit 1, 4 pass / 6 FAIL`
- `bun test test/binary-self-update.test.ts test/check-update-refresh.serial.test.ts test/release-workflow.test.ts — exit 0, 38 pass / 0 fail`
- `bun test test/build-llms.test.ts --timeout=60000 — exit 0, 12 pass (llms bundle freshness; RELEASING.md is inlined at scripts/llms-config.ts:251)`
- `actionlint .github/workflows/release.yml — exit 0, no findings (does NOT flag the injection)`
- `bash scripts/check-bun-test-timeout.sh — exit 0 on merged tree`
- `bash /tmp/injsim/sim.sh (injection PoC) — arbitrary command executed, step exit 0`
- `git ls-remote --tags origin — exit 0, 1 tag total, ZERO v* tags`
- `gh api repos/garrytan/gbrain/releases/latest — exit 1, HTTP 404 (premise confirmed)`
- `gh api repos/actions/checkout/git/ref/tags/v4 + repos/softprops/action-gh-release/git/ref/tags/v2 — exit 0, both match the pinned SHAs`
- `bash scripts/changelog-entry.sh 0.42.66.0 / 0.41.16.0 | wc -c — exit 0, 8357 / 13365 bytes`
- `gh pr view 3573 --json mergeable,mergeStateStatus,statusCheckRollup — CLEAN/MERGEABLE, 23/23 checks SUCCESS incl. semgrep + actionlint`

</details>

---

time-attack:
Merged into master via #3717 (v0.42.71.0) with your commit attributed via Co-Authored-By and CHANGELOG credit — plus one hardening from review (env-bound the release-notes interpolation) and one post-launch correction (the build job now gates on a binary smoke test instead of re-running the full suite serially, which blocked the first run exactly as the review predicted). The Releases page is live with organized notes and binaries, self-update resolves against real assets, and today's two fix waves are backfilled. Thanks — this closed both #3521 and #3716.

Issue #892, opened 2026-05-11
v0.32 security: 'gbrain config set openai_api_key <key>' echoes full key to stderr


topcoder1:
## Bug

`gbrain config set openai_api_key sk-...` confirms by printing the **full key** in plaintext to stderr:

```
Set openai_api_key = sk-svcacct-MGoTY7Gw3zMRXJb7n_7Y...full-key-visible...
```

The same applies to any sensitive key set via `gbrain config set`.

## Why this matters

- Keys end up in shell history (if user ran the command directly)
- Keys end up in CI logs / agent transcripts (if a tool wrapper called it)
- Keys end up in tmux/screen scrollback
- Keys end up in any log capture that sucks in stderr

In my session, an AI agent helping the user set up gbrain leaked their full OpenAI service-account key to the conversation transcript via this confirmation message. The user now has to rotate the key.

## Suggested fix

Redact the value in confirmation output. Mask all but first 8 + last 4 chars, or just print `Set openai_api_key = <REDACTED>`. Same treatment for any key whose key-name matches `*_api_key`, `*_token`, `*_secret`, `*password*`, `*pat*`.

For comparison: `claude mcp add` and the gstack `gstack-gbrain-supabase-provision` helper both correctly redact their secret args.

## Repro

```bash
gbrain config set openai_api_key sk-test-1234567890abcdef
# Outputs: Set openai_api_key = sk-test-1234567890abcdef
```

## Environment

gbrain 0.32.0

PR #3575, opened 2026-07-29
fix(config): config show renders object values as JSON, not [object Object] (#575)

### LINKED ISSUE #575: bug: `gbrain config show` prints `[object Object]` for `provider_base_urls` field
## Summary

When `provider_base_urls` is set in `~/.gbrain/config.json`, `gbrain config show` prints the literal string `[object Object]` for that field instead of the JSON content. Missing `JSON.stringify()` somewhere in the config-show formatter.

## Repro

```bash
# config.json contains:
# {
#   "provider_base_urls": {
#     "ollama": "http://localhost:11434"
#   }
# }

gbrain config show
```

Output (snippet):
```
provider_base_urls: [object Object]
```

## Expected

```
provider_base_urls: { "ollama": "http://localhost:11434" }
```

or pretty-printed JSON.

## Fix

One-line change — wrap the field's render with `JSON.stringify(value, null, 2)` (or your existing object-render helper). Happy to send a PR if pointed at the relevant printer file.

## Environment

- gbrain v0.27.0

time-attack:
## Summary

`gbrain config show` interpolated non-string config values straight into a template literal, so object-valued fields (e.g. `provider_base_urls`) printed the literal string `[object Object]`.

## Fix

Objects render via `JSON.stringify`; objects under a sensitive key (per the existing `isSensitiveConfigKey` allowlist) redact to `***` exactly like their string counterparts — the fix does not open a leak path for structured secrets. Scalars unchanged.

## Discrimination proof

`test/config-show-object-values.test.ts` on unmodified master: **fails behaviorally** — captured output shows `provider_base_urls: [object Object]`. With the fix: the JSON value renders (`http://localhost:11434` visible), `[object Object]` absent, and an object under a sensitive key (`some_nested_secret`) does not leak its contents (asserted).

Config suites green (`config`, `config-set`, `config-unset`, `config-get-plane` — 64 pass). `bun run typecheck` clean; `bun run verify` 32/32 green.

Closes #575


time-attack:

The bug (#575) is real and the fix direction is right, but the implementation converts a no-leak render (`[object Object]`) into a full plaintext secret dump for every object-valued config key whose TOP-LEVEL name doesn't match the sensitivity regex. Empirically confirmed leaks: `storage.accessKeyId`/`secretAccessKey`/`serviceRoleKey`, `remote_mcp.oauth_client_secret`, `provider_chat_options.*.apiKey`, and any nested `postgresql://user:pass@` (the string-path `redactUrl` never runs on the object branch). All three are declared fields of `GBrainConfig`/`StorageConfig`, not hypotheticals. This is a regression on the exact surface v0.36.x #892 hardened ("API keys leak via shell history + scrollback"), and the repo already established the counter-pattern for this class in `src/core/source-config-redact.ts` + `scripts/check-source-config-leak.sh`. The PR's own test suite is green (60 pass) and typecheck is clean, so CI cannot see this. The PR's `[object Object]` assertion DOES discriminate against master (good), but its security assertion is vacuous: it picks key `some_nested_secret`, which matches `isSensitiveConfigKey` and is therefore the one nested case already covered — it passes on master too. Fix is ~5 lines: recurse leaves through the existing `redactConfigValue` before `JSON.stringify`. Also: PR title omits the mandatory leading vMAJOR.MINOR.PATCH.MICRO.

### Discrimination test

EXECUTED, both directions. (1) PR's own test vs master behavior: checked out `origin/master -- src/commands/config.ts` (restoring the one-line `const display = typeof v === 'string' ? redactConfigValue(k, v) : v;`) and ran `bun test test/config-show-object-values.test.ts` → EXIT=1, 0 pass / 1 fail, output shows `provider_base_urls: [object Object]`. So the core assertion is NOT vacuous — the test genuinely pins the fix. However its second (security) assertion passes on master too, so 1 of the 2 assertions proves nothing. (2) Reverse discrimination on the regression: wrote test/zz-review-leak.test.ts (deleted after) asserting no nested secret appears in `config show` output for `storage`, `remote_mcp`, `provider_chat_options`. On the PR branch: EXIT=1, 0 pass / 3 fail — all three secrets printed verbatim. With master's config.ts stubbed in: EXIT=0, 3 pass / 0 fail, all three rendering `[object Object]`. This proves the leak is introduced by this PR and does not pre-exist.

### Blockers

**1.** src/commands/config.ts:52-58 — new secret exposure. `isSensitiveConfigKey(k)` is checked only on the top-level key, so object values under a non-sensitive name are JSON.stringify'd in full. Failing input: ~/.gbrain/config.json = {"engine":"pglite","storage":{"backend":"s3","bucket":"b","accessKeyId":"AKIAIOSFODNN7EXAMPLE","secretAccessKey":"wJalrXUtnFEMI-SUPER-SECRET","serviceRoleKey":"eyJSERVICE-ROLE-JWT"}}. `gbrain config show` on master prints `storage: [object Object]`; on this branch it prints `storage: {"backend":"s3","bucket":"b","accessKeyId":"AKIAIOSFODNN7EXAMPLE","secretAccessKey":"wJalrXUtnFEMI-SUPER-SECRET","serviceRoleKey":"eyJSERVICE-ROLE-JWT"}`. `storage` is a declared GBrainConfig field (src/core/config.ts:106) and StorageConfig declares those exact credential fields (src/core/storage.ts:23-27).

**2.** src/commands/config.ts:52-58 — same class, second confirmed carrier: `remote_mcp` (src/core/config.ts:342-347) declares `oauth_client_secret`. Failing input: {"engine":"pglite","remote_mcp":{"issuer_url":"https://i","mcp_url":"https://m","oauth_client_id":"cid","oauth_client_secret":"OAUTH-CLIENT-SECRET-VALUE"}} → printed in full; `isSensitiveConfigKey('remote_mcp')` is false (asserted in probe).

**3.** src/commands/config.ts:52-58 — nested postgresql:// password bypasses redactUrl. The object branch never routes through `redactConfigValue`, so `redactUrl` only protects top-level strings. Failing input: {"database_url":"postgresql://u:TOPLEVELPASS@h/db","provider_chat_options":{"openai":{"apiKey":"sk-nested-provider-key","dsn":"postgresql://u:NESTEDPASS@h/db"}}} → output is `database_url: postgresql://u:***@h/db` (correctly redacted) but `provider_chat_options: {"openai":{"apiKey":"sk-nested-provider-key","dsn":"postgresql://u:NESTEDPASS@h/db"}}` (both secrets in the clear).

**4.** test/config-show-object-values.test.ts:47-49 — the security assertion is non-discriminating by construction. It asserts an object under key `some_nested_secret` doesn't leak `sk-super-secret`; that key MATCHES /(^|[._-])(secret)([._-]|$)/ so it hits the '***' branch, and the same assertion also passes on unmodified master (`[object Object]` contains no secret). No test covers a nested secret under a non-sensitive key — i.e. the only realistic shape. Re-running this file with master's `const display = typeof v === 'string' ? redactConfigValue(k, v) : v;` stubbed back in: only the `[object Object]` assertion fails; the redaction assertion passes.

### Not verified

- Full unit suite / `bun run verify` / `bun run ci:local` — deliberately skipped per execution-footprint rules. The PR claims 'verify 32/32 green'; I confirmed typecheck clean + the 4 config suites (60 pass) only.
- No real-Postgres path exercised (Docker not needed — this is a pure file-plane config printer under PGLite).
- Whether any downstream doc/skill instructs users to paste `gbrain config show` output into a bug report (which would raise severity further): grepped docs/, README.md, skills/ for 'config show' with zero hits, so no in-repo instruction found; external/support-channel practice not checked.
- Did not check whether an equivalent leak exists in any non-repo surface (e.g. a hosted admin UI) that renders loadConfig() output.

- `gh pr view 3575`
- `gh pr diff 3575`
- `gh issue view 575`
- `git fetch origin pull/3575/head:pr3575-review-16 && git checkout pr3575-review-16`
- `ln -s <workspace>/node_modules ./node_modules  (worktree had none; removed afterwards)`
- `bun test test/zz-review-leak.test.ts  (custom leak probe, PR branch) -> EXIT=1, 0 pass / 3 fail`
- `git checkout origin/master -- src/commands/config.ts`
- `bun test test/zz-review-leak.test.ts  (master behavior) -> EXIT=0, 3 pass / 0 fail`
- `bun test test/config-show-object-values.test.ts  (master behavior) -> EXIT=1, 0 pass / 1 fail`
- `git checkout pr3575-review-16 -- src/commands/config.ts`
- `bun run typecheck -> EXIT=0 clean`
- `bun test test/config.test.ts test/config-set.test.ts test/config-unset.test.ts test/config-show-object-values.test.ts -> EXIT=0, 60 pass / 0 fail`
- `grep -rn 'redactConfigValue|isSensitiveConfigKey' src/  (only 2 call sites: show + set)`
- `grep -rln 'redact|scrubSecret|maskSecret' src/  (found src/core/source-config-redact.ts prior-art)`
- `rm test/zz-review-leak.test.ts; git status --porcelain -> clean`

</details>

---

time-attack:
**Merging this closes #575** (severity **p2**).

#575 was one of 58 issues carrying a `p3` label. All 58 were re-triaged by reproducing them against current master, and **the label was wrong on 56 of them** — the corrected distribution is 4 p0, 40 p1, 12 p2, and 2 genuinely p3. #575 came out at **p2**, and this PR is what fixes it.

Two things follow from that:

- Add `Closes #575` to the PR description if you want the link wired up — a closing keyword only fires from the body or a commit message, not from a comment like this one.

The reproduction and evidence are on #575.

PR #3576, opened 2026-07-29
fix(cli): make --brain actually route to the named brain

time-attack:
## The bug

CLAUDE.md and `docs/architecture/brains-and-sources.md:130` promise:

```
gbrain query "X" --brain media-team   → runs against the team's DB
```

`src/core/brain-resolver.ts` has implemented the full 6-tier chain since v0.19 — but **nothing ever called it from the CLI dispatch path**. `--brain <id>` fell through per-command parsing as an unknown flag and the command silently ran against the **host** brain. Same for `GBRAIN_BRAIN_ID` and the `.gbrain-mount` dotfile. Silent-wrong-results: the user cannot tell it happened.

## The fix

**Precedence chain implemented** (the documented one, via the existing `resolveBrainId`):

```
--brain <id>  >  GBRAIN_BRAIN_ID  >  .gbrain-mount dotfile  >
longest-prefix mount path  >  (brains.default, reserved)  >  'host'
```

- `parseGlobalFlags` (`src/core/cli-options.ts`) strips `--brain <id>` / `--brain=<id>` before per-command parsing, exactly like `--source` — so it cannot collide with #3503's unknown-flag rejection. Exact-match only: skillopt's `--brain-wide-max-cost-usd` and any `--brain-*` flag pass through untouched. A missing or malformed value **throws at parse time**.
- `connectEngine` (`src/cli.ts`) is the single choke point every local command routes through (shared ops, CLI-only commands, the search dashboard path) — brain resolution lands there once. Non-host ids route through `BrainRegistry.getBrain`, which lazily connects the mount's own engine (brain = a different database, not a row filter).

**Fail-closed, not fall-back** (mirror of #3524's explicit-source decision):
- Unknown/disabled brain id → `UnknownBrainError`, non-zero exit, names the available ids. Pre-fix behavior (silent host fallback) is exactly the bug.
- Invalid id shape → parse-time throw.

**Trust boundary:** brain selection is derived ONLY from the trusted local CLI boundary (global flag / process env / operator-owned dotfiles) at engine-connect time, before any op runs. It is never read from op params: `makeContext` stamps `ctx.brainId` from module state, no operation declares a `brain`/`brain_id` param, and `buildOperationContext` ignores caller-supplied ones — all pinned by tests. A remote MCP caller has no path to name a brain; the server keeps serving whatever brain the operator started it on. Thin-client installs refuse an explicit `--brain` loudly (the remote server is one brain); ambient tiers are ignored there, matching the source axis's behavior.

**Access control:** mounts carry their own DB credentials in `~/.gbrain/mounts.json`; the CLI connects with those and ONLY those. Enforcement of what the subscriber may do is the mounted database's own grants (a read-only role rejects writes at the DB) — gbrain does not re-implement the publisher's policy client-side. Additionally, no auto-migrations ever run against a mount (`connectMountEngine` skips the migration runner — schema is the publisher's job, same decision as `BrainRegistry.initMountBrain`), and the AI gateway stays configured from the HOST config, so a mount's DB-plane model config can never redirect the caller's API spend.

**Read vs write scope:** both route. Brain = which database, for every command — a write with `--brain team-a` lands in team-a, which is what "ingest data that belongs to a specific team" (skills/conventions/brain-routing.md) requires. Write *permission* is the mount credential's job (above). `childGlobalFlags` propagates `--brain=<id>` into child `gbrain` subprocesses so a routed parent can't spawn host-targeting children (env/dotfile tiers self-propagate via inherited env + cwd).

## Discrimination proof (unmodified origin/master @ e72d93fd)

Master tree materialized via `git archive origin/master` into a clean dir, new test files copied in, run behaviorally (no missing-export failures — every failure is an assertion on real CLI output):

```
bun test test/brain-flag-routing.serial.test.ts        →  4 pass, 3 fail
✗ --brain team-a → the mount database, not host
    Expected to contain: "team-page"
    Received: "host-page\tnote\t2026-07-29\thost-page\n"   ← the silent wrong answer
✗ --brain <unknown> hard-errors — never a silent host fallback
    expect(r.exitCode).not.toBe(0)   — master exits 0 with host data
✗ GBRAIN_BRAIN_ID=team-a env tier is wired through the same seam
    Received: "host-page..."

bun test test/cli-options.test.ts -t brain             →  4 pass, 8 fail
    Expected: "media-team"  Received: undefined  (flag never parsed)
```

With the fix: 7/7 and 39/39 pass. The passing subset on master is the fail-closed pins (untrusted caller cannot cross brains; `--brain-*` pass-through; control spawn) — those guard the new surface.

## Gates

- `bun run typecheck` — clean
- `bun run verify` — 32/32 green (one flaky-under-parallel-load `check:wasm` broken-pipe on first run; passes standalone and on re-run, also present-shape on master)
- `bun run build:llms` — bundle fresh (`test/build-llms.test.ts` green)
- Adjacent suites green: brain-resolver, brain-registry, mounts-cli, all-sources-sentinel, cli-dispatch-thin-client, cli-args, thin-client-upgrade-prompt, doctor-orphan-ratio, extract-by-mention(+resume), cli-search-dispatch
- No VERSION / package.json / CHANGELOG changes (per instruction)

## Interaction with #3503 (unknown flags become hard errors)

Once this lands, `--brain` is a recognized global flag stripped by `parseGlobalFlags` before `parseOpArgs` runs — so `gbrain query "X" --brain media-team` drops out of #3503's set of documented invocations that would newly throw. #3503's blast radius shrinks accordingly, and the two changes compose: an unknown flag errors there, an unknown *brain id* errors here.


time-attack:

The core claim is true and I verified it by execution: `resolveBrainId` was dead code, and this PR wires it into `connectEngine` so all six tiers now route (I independently exercised flag, env, `.gbrain-mount` dotfile, mount-path-prefix, host default, and unknown-brain error). Discrimination test reproduces cleanly on an unmodified master tree (3 fail / 4 pass), typecheck exit 0, `bun run verify` 32/32, the new serial file plus 9 adjacent suites all green, llms bundle regenerates with no diff. The blocker is incompleteness of the same bug class the PR exists to kill: `--brain` is stripped globally in `parseGlobalFlags` but honored only inside `connectEngine`, and ~9 CLI-only commands build their own engine from the HOST config — so `gbrain auth create tok --brain team-a` exits 0, prints "Token created", and writes the credential into the HOST brain (executed and confirmed; team-a's `access_tokens` stays empty), while the KEY_FILES.md entry this PR ships documents the flag as routing with no qualification.

### Discrimination test

PASSES. I materialized unmodified origin/master (e72d93fd) via `git archive` into /tmp/master3576, copied ONLY the new test file in, and ran it: 4 pass / 3 fail. The three failures are behavioral assertions on real CLI stdout, not missing-export errors: `--brain team-a → the mount database` got "host-page\tnote\t...\thost-page\n" (the silent wrong answer); `--brain <unknown> hard-errors` got exit 0; `GBRAIN_BRAIN_ID=team-a` got the host page. Same file on the PR branch: 7 pass / 0 fail. The 4 tests that pass both ways are the fail-closed pins (no op declares a brain param, makeContext/buildOperationContext ignore params.brain, `--brain-*` pass-through) — correctly guards, not fix-detectors, and the PR body says so.

### Blockers

**1.** src/cli.ts:2402 + src/core/cli-options.ts:160 — `--brain` is stripped for EVERY command but honored only by commands that acquire their engine via `connectEngine`. Executed failing input: `gbrain auth create team-scoped-token --brain team-a` against a fake home with host brain + one enabled PGLite mount `team-a` → exit 0, stdout "Token created for \"team-scoped-token\"", HOST `access_tokens` = [{name:'team-scoped-token'}], TEAM `access_tokens` = []. The credential lands in the wrong database with no warning. Affected early-return commands that build their own engine from `loadConfig()` (host) instead of `connectEngine`: `auth` (src/cli.ts:1342 → src/commands/auth.ts:45), `integrity` (src/cli.ts:1352), `cache` (src/cli.ts:1389), `schema` (src/cli.ts:1274), `lint` (src/cli.ts:1372), `frontmatter` (src/cli.ts:1367), `upgrade` (src/cli.ts:1311), `repair-jsonb` (src/cli.ts:1440), `apply-migrations` (src/cli.ts:1431). Second executed instance: `gbrain integrity check --brain team-a` prints "Scanned 3 pages" (the host's 3) while `gbrain stats --brain team-a` correctly prints "Pages: 1". NOT a regression — master is equally silent — but the PR's own stated principle is loud-beats-silent (it applies exactly that rule to thin clients at src/cli.ts:402) and the KEY_FILES.md entry it adds (docs/architecture/KEY_FILES.md:10) tells readers the flag routes, unqualified. Minimal ask: one shared guard in `handleCliOnly` refusing an EXPLICIT `--brain` for that command set (ambient tiers must stay permissive or `cd ~/mount-clone && gbrain lint` breaks for existing users), plus a one-line caveat in the KEY_FILES.md entry.

**2.** src/commands/migrations/v0_12_2.ts:51 — the PR's new `childGlobalFlags()` brain propagation (src/core/cli-options.ts:182) splices `--brain=<id>` into `gbrain repair-jsonb`, which is in the bypass set above and ignores it, while the sibling call at src/commands/migrations/v0_12_0.ts:98 (`gbrain extract links`) does honor it. So a single migration orchestrator emits two child commands that disagree about which database they target, and the repair child silently mutates the host. Self-inconsistent with the feature the PR adds.

### Not verified

- Postgres-backed mounts: every probe used PGLite on both sides (no DATABASE_URL available here). The per-instance `poolSize: 5` path in BrainRegistry.initMountBrain and the read-only-role-rejects-writes access-control story the PR body leans on are UNVERIFIED by execution.
- Embedding-dimension mismatch across brains: a mount whose DB was embedded at a different dimension/provider than the host gateway (`connectMountEngine` configures the gateway from HOST config only). Untested — a `query --brain <mount>` there could error or silently return nothing.
- Source-axis interaction: `makeContext` still stamps `sourceId` resolved from the host cwd/env, so `--brain team-a` inherits the host's source scope. My probe happened to match on `default`. A mount whose content lives under a different source_id was not tested and would return silent empty results.
- Daemon paths under an ambient brain signal (`gbrain serve` / `jobs work` / `autopilot` / `dream` started from inside a mount clone, or with GBRAIN_BRAIN_ID exported): these all route through connectEngine at src/cli.ts:1819 so they WILL target the mount with no migrations applied. Not executed; flagged as a behavioral consequence of activating tiers 3/4, which docs/architecture/brains-and-sources.md:194-197 does document as intended.
- `bun test` full unit suite and `bun run test:e2e` were not run (no DATABASE_URL; full suite exceeds the time budget). I ran typecheck, verify, the new file, and 14 adjacent files instead.

- `gh pr view 3576 — exit 0`
- `gh pr diff 3576 — exit 0`
- `git fetch origin pull/3576/head:pr3576 && git checkout pr3576 — exit 0`
- `bun install — exit 0`
- `bun test test/brain-flag-routing.serial.test.ts (PR branch) — exit 0, 7 pass / 0 fail`
- `git archive e72d93fd | tar -x -C /tmp/master3576 (+ copy new test) then bun test test/brain-flag-routing.serial.test.ts — exit 1, 4 pass / 3 fail (DISCRIMINATION CONFIRMED)`
- `bun run verify — exit 0, pass=32 fail=0`
- `bun test test/cli-options.test.ts test/brain-resolver.test.ts test/brain-registry.test.ts test/mounts-cli.test.ts test/cli-args.test.ts test/thin-client-upgrade-prompt.test.ts test/doctor-orphan-ratio.test.ts test/build-llms.test.ts — exit 0, 156 pass / 0 fail`
- `bun test test/all-sources-sentinel.test.ts test/cli-dispatch-thin-client.test.ts test/cli-search-dispatch.test.ts test/extract-by-mention.test.ts test/extract-by-mention-resume.test.ts — exit 0, 51 pass / 0 fail`
- `bun test (own probe) test/zz-tiers-probe.test.ts — exit 0, 6 pass: proved tiers 3 (.gbrain-mount) and 4 (mount path prefix) route, --brain host beats dotfile, unknown dotfile brain exits 1, malformed GBRAIN_BRAIN_ID exits 1`
- `bun test (own probe) test/zz-bypass-probe.test.ts — exit 0: `integrity check --brain team-a` reported the HOST's 3 pages while `stats --brain team-a` reported 1`
- `bun test (own probe) test/zz-auth-probe.test.ts on PR branch — exit 0: `auth create --brain team-a` wrote the token to the HOST access_tokens table, team-a empty`
- `bun test (same probe) on /tmp/master3576 — exit 0: identical host-write, so the auth misroute is pre-existing, not a regression`
- `bun run build:llms + git status --porcelain — exit 0, no diff (bundle fresh)`
- `rm probe test files; git status --porcelain — clean, worktree left as checked out`

</details>

---

time-attack:

PR #3582, opened 2026-07-29
test(build-llms): compare bundle content, not line endings

daragao3:
## Summary

Fix the `build-llms` freshness test on Windows by comparing normalized bundle content rather than checkout-specific CRLF bytes.

With `core.autocrlf=true`, Git checks the committed `.txt` bundles out as CRLF while the generator emits LF scaffolding. The previous byte comparison therefore failed on a clean Windows checkout with a zero-line diff, even though `git diff --quiet -- llms.txt llms-full.txt` reported no content drift.

The test now normalizes `\r\n` to `\n` on both sides of both Case 4 assertions. Lone `\r` bytes remain significant, so the change neutralizes only the checkout artifact. This is intentionally an in-test fix rather than a `.gitattributes` pin, because it also repairs existing clones without requiring a forced re-checkout.

## Tests

- `bun test test/build-llms.test.ts --timeout=120000` (`12 pass`, `0 fail`) with the checked-out bundles still containing CR bytes
- Negative control: appending a sentinel to both committed bundles still fails Case 4 and prints the sentinel
- Negative control: appending a sentinel to inlined `docs/guides/push-context.md` while leaving the bundles stale still fails Case 4 and prints the sentinel
- `bun run typecheck`

## Scope

- One test-only file
- No generated-bundle changes
- No version or changelog changes; release numbering remains with the maintainer fix-wave process

### DIFF
--- a/test/build-llms.test.ts
+++ b/test/build-llms.test.ts
@@ -7,6 +7,17 @@ import { SECTIONS, FULL_SIZE_BUDGET } from "../scripts/llms-config";
 
 const repoRoot = join(import.meta.dir, "..");
 
+// Compare bundle CONTENT, not bytes. On Windows `core.autocrlf=true` (the
+// Git-for-Windows installer default) checks the committed bundles out as CRLF,
+// while the generator emits its own scaffolding as LF — so a byte compare fails
+// on a checkout with zero real drift. `.gitattributes` can't fix this here: it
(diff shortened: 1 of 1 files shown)


daragao3:
Closing at the contributor's request; the fix will remain on the contributor branch for local use.

Issue #3570, opened 2026-07-29
Rename sentinel from #3479 can wedge sync with no operator escape, and orphans into a permanent doctor FAIL


time-attack:
Follow-up to #3479, which the maintainer merged knowingly accepting these. Filing so they're tracked rather than buried in a PR comment.

#3479 added a `<rename:X>` sync-failure sentinel. It is data-loss safe — I tested eleven scenarios trying to make it destroy a page and could not. But it introduces two availability problems, both reproduced.

## 1. A permanently failing reconcile wedges sync with no operator escape

Five consecutive `gbrain sync --skip-failed` runs against an always-throwing `deletePage`:

```
run1..run5 (--skip-failed) = blocked_by_failures  [["<rename:people/dana.md>","open",1..5]]
```

Structural, not incidental: `decideGateAction` hard-blocks on any sentinel **even with** `--skip-failed`, and `acknowledgeFailures` explicitly skips sentinels (`src/core/sync-failure-ledger.ts:515`). Realistic triggers: RLS denying DELETE on a Supabase brain, an FK RESTRICT, a read-only DB.

The blocked message says only *"The next 'gbrain sync' retries the reconcile from the same diff"* — it names no remedy. **The remedy is `gbrain delete <stale-slug>`**, after which the reconcile no-ops and the sentinel clears. That belongs in the message.

## 2. An orphaned sentinel is a permanent, unfixable doctor FAIL

After a force-push invalidates the pinned target, the destination stops recurring in the diff and the row is never cleared:

```
run1 (--skip-failed) = synced      [["<rename:people/dana.md>","open"]]
run2 (--skip-failed) = up_to_date  [["<rename:people/dana.md>","open"]]
```

`decideSyncFailureSeverity` escalates `sync_failures` to FAIL once the oldest open row passes `failHours`, and **no CLI can clear a sentinel** — hand-editing `~/.gbrain/sync-failures.json` is the only fix.

Master already treats this exact class as a bug for `<head>` and added an unconditional clear (`src/commands/sync.ts:3350`, `...(headVerificationSucceeded ? ['<head>'] : [])`). The `<rename:X>` analogue is missing: clear any open `<rename:X>` whose stale `source_path` row no longer resolves.

## Smaller residuals from the same review

[...]

PR #598, opened 2026-05-03
fix(import): allow frontmatter slug for un-sluggable root paths (CJK filenames)

vinsew:
## Summary

`slugifyPath()` strips non-ASCII characters before validation. A markdown file at the brain repo root with a Chinese-only (or any non-Latin) filename — `小米.md`, `品牌圣经.md`, `创意原型库.md` — sluggifies to the empty string, which fails downstream `Invalid slug: ""` validation. Sync silently rejects the file every cycle and the failure accumulates in `~/.gbrain/sync-failures.jsonl`.

The previous code preferred the (empty) path-derived slug over the frontmatter slug, so users adding an explicit `slug:` field to the frontmatter saw their file STILL rejected, this time with a `SLUG_MISMATCH` error — a frustrating second failure mode for the same root cause.

## Fix

When `slugifyPath(relativePath)` returns empty AND the file's frontmatter has a `slug:` field, use the frontmatter slug as the authority. The original anti-spoofing rule (path slug exists AND frontmatter slug differs → reject) is preserved unchanged: a normal file at `wiki/people/elon.md` cannot impersonate a different page via frontmatter.

The exception only fires when there is no usable path-derived slug to begin with. Empty-on-both-sides yields a friendlier error message that points at the fix (add a frontmatter `slug:` or move the file).

## Impact

In the field this affected 23 files in a 1386-page brain on the maintainer's deployment — all root-level Chinese-named files. Without this fix, brain sync had a 70-row backlog of `Invalid slug: ""` failures that re-appeared every 15-minute live-sync cycle.

## Test plan

- [x] `bun run typecheck` clean
- [x] New case in `test/import-file.test.ts` asserting `小米.md` with `slug: companies/xiaomi` imports correctly
- [x] Existing attack-defense case (`notes/random.md` trying to claim `people/elon`) unchanged and still rejected
- [x] All 33 import-file tests pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
<a href="https://app.blacksmith.sh/garrytan/codesmith/gbrain/pr/598"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need.</sup>

[...]

PR #3583, opened 2026-07-29
fix(sync): rename-sentinel operator exit + orphan self-heal (#3570)

Masashi-Ono0611:
## Summary

Follow-up to #3479, closing #3570, revised end to end in response to the adversarial review below. The branch does three things:

1. **The rename reconcile deletes only provably-stale rows** (review blocker 1). The reviewed head deleted every active row whose `source_path` matched the rename's old path — a set that routinely contains a LIVE page, because `updateSlug` never rewrites `source_path` and an unchanged-content re-import writes nothing. A candidate is now deleted only when no tracked file derives to its CURRENT slug, with derivation mirroring import end to end: ordinary paths through `resolveSlugForPath` (import's anti-spoof check makes the path the slug authority), frontmatter-fallback files (#598) through the same parse + `validateSlug` chokepoint import runs, and content read the way import reads it (frontmatter inference, working tree AND index blob, `.gitattributes` filters consulted at the anchor commit's epochs — a filtered or unreadable state spares rather than deletes). A cheap rename also repairs the row's `source_path` when it lands, source-scoped exactly the way `updateSlug` scopes the move. That repair is best-effort — its own SQL error is swallowed so it can never fail the rename — and a row whose repair fails keeps the pre-existing full-sync-purge exposure.

   Liveness is a three-way verdict (live / stale / unknown), and **unknown spares**. An unprovable spare otherwise proceeds normally — that shape is usually a live row whose slug simply could not be read (a content filter, a shallow clone, an unmerged index, an over-size file), where nothing is pending. The one exception is when the rename ALREADY has an open `<rename:…>` sentinel: convergence has been denied before, and clearing a non-convergence marker requires proof of convergence, which "unprovable" is not. That case records a reconcile failure, so the sentinel survives and the bookmark freezes until the unreadable file is fixed — and the failure deliberately names no row AS STALE (`?`), because the operator remedy is `gbrain delete <slug>` and an unprovable row may be live. The cause text still lists which candidates were unprovable, so the operator is not left without a lead.

2. **The ordinary `--dry-run` path no longer mutates page, source, checkpoint, or failure-ledger state, and rename sentinels self-heal safely** (review blocker 2). One exception I did not touch: explicitly combining `--skip-failed` still acknowledges pre-existing failures, because `runSync` acks before the sync without consulting `dryRun`. The quiet-run sweep is dry-run guarded, and four more pre-existing preview writes are fixed the same way: the `last_sync_at` heartbeat bump, the unreachable-pin checkpoint clear, the un-syncable-modified page hard-delete, and the unconditional `writeChunkerVersion` after the forced re-chunk (that one fired on a preview because `performFullSync` returns early on a dry run and control resumed past it — it also acknowledged the version on a BLOCKED run, losing the retry signal). The orphan self-heal got a probe/clear race fix (two consecutive positive probes, then a post-clear verify that RESTORES any sentinel whose old path re-acquired an active row), and the sweep sits outside the failure gate, so a sentinel the SWEEP clears cannot be lost to a throwing `advance()` — the gate still clears `succeededPaths` before calling it, which this branch does not change. The sentinel parser now also reads the PRE-#3479 unquoted format where it is unambiguous, so a brain left with an ORPHANED legacy sentinel by #3056 before upgrading heals on its first run instead of keeping a permanent doctor FAIL — with more than one delimiter occurrence still failing closed.

3. **The named coverage holes have tests, and the array bind is exercised on real PostgreSQL** (review blocker 3 + the review's "not verified" list). All five holes are covered in the PGLite serial file; the PostgreSQL file adds two tests for the driver-level bind, not an engine-parametrized copy of the five. All five holes: the `totalChanges==0` sweep site (dry-run/real pair), `performFullSync`'s pre-gate probe, the post-purge sweep (which the pre-gate probe *cannot* cover — at that point the stale row is still active, so the sentinel is legitimately open; control executed by disabling that single call), the widened multi-row delete against a live row (including the review's PROBE B shape), and byte-identical dry-run ledger purity. Plus the upgrade simulation the review listed as not run, and `test/e2e/sync-reconcile-postgres.test.ts` — the reconcile's `= ANY($1::text[])` binding through postgres.js `unsafe()` executed against real PostgreSQL 17.10 + pgvector (DATABASE_URL-gated, skips cleanly without it).

**One open question I would rather put to you than decide myself.** Sparing on `unknown` means a duplicate that master WOULD have deleted can survive, and across several renames while the index stays incomplete those accumulate (measured: three chained renames leave 4 active rows here vs 1 at the merge base). That is the direct consequence of "never delete without proof" — master reaches 1 row by deleting rows it cannot prove stale, which is blocker 1. I have deliberately NOT added a persistent-warning state or a bookmark stall to chase it, because both are new machinery you did not ask for. If you would rather have the aggressive cleanup back with a narrower proof, say so and I will do that instead.

**Scope note.** Earlier revisions of this branch also redesigned the full-sync purge, and separately built a three-rail proof for the cheap move's SOURCE selection (derived slug ∩ active path claim ∩ anchor-blob content match). Both are removed. Master proves nothing on either path — its cheap move takes whatever the batch resolve returns — so every shape those mechanisms closed reproduces at the merge base, and neither was required by the blocker, which is delete-side. Worse, an executed merge-base control showed the move-side proof — specifically its sole-claimant widening, which is what admits a non-derived row — made the branch *worse* than master on a reachable shape: a signal twin (another tracked file's live row, sole claimant of the path, identical title/compiled_truth/timeline) passed the content check and was moved and overwritten, where master keeps it. The purge here is byte-for-byte master's plus the one behavior blocker 3 named (the post-purge sentinel sweep), and the cheap-move selection is master's on the source-scoped path — the bare no-`sourceId` path is default-source scoped (see Notes), and an empty old slug skips an `updateSlug` call that would match no row either way. The purge redesign is preserved on `backup/3583-full-head-fa815d0b` in case it is worth a separate PR later.

## Testing

<details>
<summary>Executed verification (commands and controls)</summary>

- `bun test test/sync-rename-reconcile.serial.test.ts` — 56 pass / 0 fail (master has 7 tests in this file).
- `bun run typecheck` clean; `bun run verify` clean; `bun run build:llms` produces no bundle change and `test/build-llms.test.ts` passes.
- Discrimination, executed:
  - **Blocker 1 probe** (live row with stale bookkeeping, occupied-destination rename) — full three-point control: PASSES on this head, FAILS on the reviewed head 6c6b9d83 (the live row is destroyed), PASSES at merge base 913d2d7f. The destruction was introduced by the reviewed revision, not inherited.
  - **Blocker 2 probe** (quiet-repo `--dry-run` with an open sentinel, ledger compared byte-for-byte) — PASSES on this head, FAILS on 6c6b9d83 (the preview clears the sentinel). The committed test cannot execute at 913d2d7f because it calls `renameSentinelPath`, which does not exist there; rewriting the probe against the literal sentinel path would give a merge-base reading, and I did not do that — so I claim only the two points I actually measured.
  - **Signal-twin shape**: this head matches merge base exactly — the twin row survives with its id, and the destination is a separate new row.
  - **Unprovable retry vs an open sentinel** (`R4_UNKNOWN_CLEARS_A_PRIOR_SENTINEL`): the sentinel survives, the run does not report `synced`, the anchor does not advance, nothing is deleted, and the sentinel does NOT name any spared row as the stale one. Two controls executed: disabling the new condition drops the sentinel count 1 → 0; restoring the old `staleSlug` assignment makes the ledger read `stale row "people/alpha"` — a row that same run had just refused to delete.
  - Legacy-format upgrade simulation: a byte-for-byte pre-#3479 ledger row heals on the first run after upgrading; control executed by removing the legacy branch — the sentinel stays open (the permanent-FAIL symptom). An ambiguous legacy row stays open in both directions.
  - Post-purge sweep control: disabling that single call leaves the sentinel open while the purge still removes the row.
- `test/e2e/sync-reconcile-postgres.test.ts`: 2 pass against PostgreSQL 17.10 + pgvector; 2 skip with no DATABASE_URL (verified in that direction too — `scripts/test-shard.sh` excludes `test/e2e/*` by construction).
- `parseRenameReconcileFrom` round-tripped over all 65,536 single-code-unit slugs: 0 mismatches. The check discriminates — emitting the path raw instead of JSON-encoded fails all 65,536.
- Whole-battery note: running all 39 `test/sync*.ts` files in ONE bun invocation produces 3 failures — writer-lock reentrance, the failed-git-pull anchor test, and a delete-batch timing test; all three pass in isolation, and CI never runs this exact 39-file one-process shape — non-serial files are partitioned across shards while the `.serial.test.ts` files run in the dedicated serial-tests job. So this is contention in an ad-hoc mode, not a branch failure.

</details>

## Notes

- The blocked-run message prints the failing sentinel rows verbatim and the operator exit (`gbrain delete <stale-slug>`); since that command is a soft delete, the reconcile considers ACTIVE rows only, so a row the operator already removed is nothing left to reconcile rather than a reason to re-fail. A sentinel reading `stale row ?` names nothing deliberately — that run proved no row stale — and the message says so rather than pointing `gbrain delete` at a page that may be live.
- The bare no-`sourceId` `source_path` resolve is now scoped to the default source, matching every write that path performs; an unscoped lookup could return another source's row sharing the path, and the rename loop consumes that value directly as its cheap-move target.
- Two limits of the liveness proof that the code records and I did not try to close: an import-time content filter supplied by unversioned attributes and later removed, and force-pushed or otherwise unreachable import history — in either case a live historical frontmatter slug can be invisible while the index still reports itself complete. The sentinel clear/probe/verify sequence likewise narrows but does not eliminate the ledger/DB race: a writer landing after the post-clear verification is outside the one-shot.
- One deliberate non-change: `--dry-run` without `--no-pull` still runs the internal `git pull` — it mutates the operator's repo, not the brain, and looks like longstanding intended behavior with its own knob; raised rather than decided here.


time-attack:

PR #3583 makes the #3056 rename-reconcile soft-delete aware, widens the reconcile from "delete the one arbitrary row resolveSlugsByPaths returned" to "delete EVERY active row whose source_path == the rename's old path", and adds an orphaned-`<rename:…>`-sentinel self-heal at four new call sites. Its own tests are real (6 of them fail with src reverted, confirmed by execution), typecheck and `bun run verify` are clean, and 189 sync tests across 16 files pass. But the widened delete is a data-loss regression I reproduced: because `updateSlug` never rewrites `source_path` and an unchanged-content re-import is a no-write skip, a LIVE page routinely carries the OLD path of a previous cheap rename — so a later rename of a recreated file at that path hard-deletes the live page, and no incremental sync ever heals it. Separately, the new sweep sits ABOVE the `--dry-run` return, so `gbrain sync --dry-run` now silently clears open sentinel rows from the failure ledger; both behaviors pass on the parent commit and fail on this PR.

### Discrimination test

EXECUTED and PASSED. `git checkout HEAD~1 -- src/commands/sync.ts src/core/sync-failure-ledger.ts src/core/sync.ts` then `bun test test/sync-rename-reconcile.serial.test.ts test/sync-failures.test.ts` → exit 1, 68 pass / 6 fail. The 6 failures are exactly the PR's new tests: blocker-1 operator exit, blocker-2 orphan self-heal, both format round-trip tests, the non-unique source_path test, and the RENAME_RECONCILE classification pin. Restoring src returns 74 pass / 0 fail. Two of the failures (blocker 1, blocker 2) import nothing new, so they fail behaviorally, not on a missing export. The PR's own claim of 5/6 is conservative — it is 6. My counter-probes discriminate in the opposite direction: PROBE B and PROBE C pass on the parent commit and fail on this PR.

### Blockers

**1.** DATA LOSS (regression, CONFIRMED by execution) — src/commands/sync.ts:3067 (`for (const s of staleSlugs) { await engine.deletePage(s, renameOpts) }`): the reconcile now hard-deletes EVERY active row whose `source_path` equals the rename's old path. That set routinely contains a LIVE page, because `updateSlug` (src/core/pglite-engine.ts:5483 / postgres-engine equivalent) never rewrites `source_path` and the follow-up `importFile` is an unchanged-content skip that writes nothing — so after any ordinary cheap rename the surviving row keeps the OLD path. Reproduced end to end (test/zz-probe-3583.serial.test.ts, PROBE A + PROBE B): (1) sync people/alpha.md; (2) `git mv people/alpha.md people/beta.md` + sync → cheap rename, DB row is `{slug: people/beta, source_path: people/alpha.md}` (PROBE_A_ROWS); (3) create a new unrelated people/alpha.md + sync → two active rows share `source_path=people/alpha.md`; (4) `git mv people/alpha.md people/gamma.md` into an occupied destination slug (the #3056 fallback this feature exists for) + sync → status `synced`, and the DB is left with ONLY `people/gamma`. people/beta.md still exists on disk, was never in the diff, and the very next sync reports `up_to_date` with the page still gone (`PROBE_B_HEAL_STATUS up_to_date beta_alive_after_resync false`) — permanent silent loss of the page, its chunks and its links until a `--full` sync. Control: the identical probe PASSES with src reverted to the parent commit (master deletes only the single resolved row). The blocked-run remedy text is pointed the same way: it tells the operator to `gbrain delete <stale-slug>` naming whichever row the loop was on, which can be a live page.

**2.** DRY-RUN IS NOT READ-ONLY (regression, CONFIRMED by execution) — src/commands/sync.ts:2394: `await sweepOrphanedRenameSentinels(...)` is inserted in the git-HEAD-equality `up_to_date` early return, which sits ABOVE the `if (opts.dryRun)` guard at src/commands/sync.ts:2545. `sweepOrphanedRenameSentinels` calls `clearFailures()`, which rewrites ~/.gbrain/sync-failures.jsonl. Reproduced (PROBE C): plant one open `<rename:people/dana.md>` sentinel, run `performSync(..., { dryRun: true })` on a quiet repo → status `up_to_date` and the ledger row is GONE (`PROBE_C_OPEN_AFTER 0`); the same probe passes on the parent commit (row survives, `PROBE_C_OPEN_AFTER 1`). A `--dry-run` preview must not delete the operator's only wedge signal. Fix is one line: move the sweep below the dry-run return (the sibling site at 2599 is already correctly placed after it, and performFullSync's dry-run return at 3834 precedes its probe at 3913, so only this one site is wrong).

**3.** TEST-COVERAGE GAP — three of the four new self-heal call sites are unexercised: src/commands/sync.ts:2599 (totalChanges==0 sweep), :3913 (performFullSync pre-gate orphan probe) and :4116 (post-purge sweep) have no test; no test covers the widened multi-row delete against a live row (the PR's own non-unique-source_path test only plants rows whose files are all gone), and no test asserts dry-run purity. The one new multi-row test asserts both rows SHOULD die, which is exactly the assumption Blocker 1 falsifies.

### Not verified

- Postgres engine behavior: everything ran on PGLite. The new `executeRaw('… = ANY($1::text[]) …', [batch, sourceId])` was not exercised against real Postgres (no DATABASE_URL available), so array-parameter binding through postgres.js `unsafe()` is inferred from 8 existing precedent call sites, not executed.
- The two performFullSync self-heal sites (src/commands/sync.ts:3913, :4116) were never executed — no test in the repo drives `--full` through the rename sentinel path, and I did not build one.
- Legacy-format sentinels: I did not run an upgrade simulation, but the PR's own test pins that the pre-#3479 unquoted error text parses to undefined, so anyone already wedged by #3056 before upgrading keeps the permanent doctor FAIL this PR claims to fix. Incomplete fix, not verified in a live upgrade.
- Full unit suite was not run (only 16 sync-related files + the PR's two); a re-bin/shard effect is not expected since no test file was added.

- `gh pr view 3583 / gh pr diff 3583 — exit 0`
- `gh pr checkout 3583 (head fix/3570-rename-sentinel-operator-exit @ 6c6b9d83) — exit 0`
- `bun install — exit 0`
- `bun run typecheck (tsc --noEmit) — exit 0`
- `bun run verify (32 parallel checks) — exit 0, pass=32 fail=0`
- `bun test test/sync-rename-reconcile.serial.test.ts test/sync-failures.test.ts — exit 0, 74 pass / 0 fail`
- `git checkout HEAD~1 -- src/{commands/sync.ts,core/sync-failure-ledger.ts,core/sync.ts} && bun test (same two files) — exit 1, 68 pass / 6 fail (discrimination test)`
- `bun test test/zz-probe-3583.serial.test.ts (my probes, PR src) — exit 1, PROBE B + PROBE C fail (live page deleted; dry-run mutated ledger)`
- `bun test test/zz-probe-3583.serial.test.ts (same probes, parent-commit src) — exit 0, 3 pass / 0 fail (control: both behaviors are regressions)`
- `bun test test/sync-rename-batch.test.ts test/sync-failure-ledger.serial.test.ts test/sync-classifier-widening.test.ts test/sync-reconcile-db-only.serial.test.ts test/sync-reconcile-mass-delete.test.ts test/sync-delete-batch.test.ts test/sync.test.ts test/performfullsync-source-id.test.ts — exit 0, 143 pass`
- `bun test test/sync-deleted-slug-embed.test.ts test/sync-no-embed-sentinel.test.ts test/sync-resumable-import.serial.test.ts test/sync-metafile-skip.serial.test.ts test/sync-ops-pages.serial.test.ts test/sync-pull-failed-anchor.serial.test.ts test/doctor-sync-consolidation.test.ts test/sync-monorepo.test.ts — exit 0, 46 pass`

</details>

---

Masashi-Ono0611:
Thanks for the review — the executed probes and the merge-base controls made this much easier to work against, and I adopted the same method throughout.

All three blockers and all four of your "not verified" items are addressed. One of the unverified ones turned out to be a real defect rather than merely unchecked, and I would rather call that out than let it read as a clean sweep.

| Item | Status | Control |
|---|---|---|
| 1. Live-row deletion | Addressed — a row is deleted only on positive proof of staleness (no tracked file derives to its CURRENT slug), with derivation mirroring import end to end; anything unprovable is spared | Your PROBE B shape: passes here, fails on 6c6b9d83 (live row destroyed), passes at merge base — so the destruction came from that revision, not from master |
| 2. `--dry-run` mutates the ledger | Addressed for the ordinary preview path — the quiet-run sweep is dry-run guarded, and four other preview writes are fixed too (including an unconditional `writeChunkerVersion` that fired on previews and on blocked runs). Not touched: `--dry-run --skip-failed` still acknowledges pre-existing failures | Quiet-repo probe comparing the ledger byte-for-byte: passes here, fails on 6c6b9d83. The committed test cannot execute at 913d2d7f because it calls `renameSentinelPath`, which does not exist there; I did not rewrite the probe against the literal sentinel path, so I claim only the two points I measured |
| 3. Coverage gap (5 holes) | Addressed — all five have tests; this file has 7 tests on master and 56 now | The post-purge sweep needed its own shape (the pre-gate probe cannot cover it); control = disabling that one call leaves the sentinel open while the purge still removes the row |
| Postgres / array binding through `unsafe()` | Addressed — new `test/e2e/sync-reconcile-postgres.test.ts` | 2 pass against real PostgreSQL 17.10 + pgvector; 2 skip without `DATABASE_URL`, verified in that direction too |
| The two `performFullSync` self-heal sites | Addressed — both exercised | Included above |
| Legacy sentinels / no upgrade simulation | **This was a real defect.** The self-heal parsed only the JSON-quoted format, so a brain wedged by #3056 *before* upgrading could never heal — the permanent doctor FAIL this self-heal exists to end, denied to exactly the operators who need it. The parser now also reads the pre-#3479 format where it is unambiguous; more than one delimiter occurrence still returns undefined and the row stays open | A byte-for-byte legacy ledger row heals on the first run after upgrading; control = removing the legacy branch, and the sentinel stays open. An ambiguous row stays open in both directions |
| Full unit suite | Addressed — CI green across all ten shards plus serial-tests, verify, and the JSONB parity guard | — |

**One correction that changed the shape of this PR.** I initially worked to an absolute safety criterion rather than the regression criterion your review uses, and built two redesigns you never asked for — the full-sync purge, and a proof for the cheap move's source selection. Merge-base controls killed both: every shape they closed is pre-existing, and the move-side proof — through the sole-claimant widening it needed — was measurably worse than master on a reachable shape. Both are removed; the purge is master's plus the one behavior blocker 3 named. I have put the mechanisms and the controls that refuted them in a follow-up comment, so you do not re-derive them while re-implementing.

**One question I would rather put to you than decide myself.** Sparing on an unprovable liveness verdict means a duplicate master would have deleted can survive, and across chained renames while the index stays incomplete those accumulate (measured: three renames leave 4 active rows here vs 1 at merge base). That is the direct consequence of "never delete without proof" — master gets to 1 by deleting rows it cannot prove stale, which is blocker 1. I did add one narrow guard: if the rename already has an open `<rename:…>` sentinel, an unprovable run no longer retires it and the bookmark freezes, because clearing a non-convergence marker needs proof of convergence. That failure names no stale row, since the remedy it prints is `gbrain delete <slug>` and an unprovable row may be live. Beyond that I deliberately stopped rather than invent a persistent-warning state or a broader bookmark stall. If you would rather have the aggressive cleanup back behind a narrower proof, say the word and I will do that instead.

The branch is eight commits and the PR description now covers the whole diff rather than the first commit, with the command list and the control results.


One thing I got wrong and want to correct rather than leave standing: CI does have `DATABASE_URL` — `e2e.yml` provisions Postgres and hard-fails if it is missing. The real gap is that no workflow invokes the new `test/e2e/sync-reconcile-postgres.test.ts`, so it runs in `DATABASE_URL`-backed local and `test:full` runs (that script skips the whole E2E phase without one) but never on push. Adding it to the jsonb-parity job's file list is a one-liner if you want it wired in; I left CI config alone rather than widen this PR.

Please re-review the current head.

Masashi-Ono0611:
FYI, since you re-implement external PRs from the diff rather than merging them: the most useful thing I can add is what is deliberately NOT in this diff, and the executed evidence against it. Four mechanisms lived in this branch at some point. Each was removed after a control showed it was out of scope, or actively worse than master. If any of them looks attractive while you are re-implementing, this is the case against it.

| Removed mechanism | What it was meant to fix | The control that killed it |
|---|---|---|
| Full-sync purge redesign: watermark capture, liveness-proof partition, dedup spare, soft removal instead of `deletePages` | The purge hard-deletes rows whose bookkeeping drifted | Every shape it addressed reproduces at 913d2d7f and is strictly WORSE there — master hard-deletes the accepted `put_page` update on the FIRST full sync, while the branch deferred then removed recoverably. Pre-existing behavior, not this PR's regression |
| Cheap-move three-rail proof: move only when the path derives a slug AND an active row at that slug claims the path AND the row holds the anchor blob's content signals | The cheap move takes whatever the batch resolve returns, so an unrelated row can be moved and overwritten | Too strict on its own: a row written by an older slug algorithm sits at a non-derived slug and fails the derived-slug rail, so the cheap move is skipped and the rename is banked as complete with the legacy row stranded; a later edit is then skipped against that row by stable-ID dedup, and a full sync ages the row out. Executed: that head lost the edit and then the row; the merge base kept both. The widening below was the attempt to recover this, and it is what actually broke |
| Sole-claimant widening: accept a non-derived row as the move source when it is the path's only active claimant | The three-rail proof was too strict for a row written by an older slug algorithm — the cheap move was skipped, the rename was banked as complete with the legacy row stranded, a later edit was skipped against that row by stable-ID dedup, and a full sync aged the row out | A signal twin — another tracked file's live row, sole claimant of the path, identical title/compiled_truth/timeline but its own frontmatter — is admitted by the widening and then passes the content check. Executed: that head moved twin `id=1` to the destination and dropped `people/twin`; merge base kept `id=1` and revived the real row as `id=2`. Worse than master on a reachable shape |
| `reconcileDeferred`: withhold the success path when a candidate was spared as unprovable, so the sentinel stays open and the next run retries | An unprovable spare was clearing the sentinel and banking an unfinished repair | It cannot work where it sits. `applySyncFailureGate` advances `last_commit` and clears both checkpoint keys regardless, so the next run returns `up_to_date` and the retry never happens. Executed: sentinel open, duplicate present, anchor advanced, retry never fires — the permanent-wedge shape #3570 exists to end |

One more that never shipped, from a local experiment rather than a commit: making EVERY unprovable verdict a reconcile failure. It failed nine existing tests covering legitimate shapes — a configured content filter, a shallow clone, an unmerged index, an over-size file — where the spared row IS the file's own live row and nothing is pending. That is why the final rule is narrow: an unprovable spare proceeds, unless the rename already carries an open sentinel.

**Method, in case it is useful.** Each tabled removal used the same two-direction check your review used: run the probe at the then-current head containing that mechanism, then run the identical probe with `src/` checked out at 913d2d7f, and record both. (This head no longer contains them, so the probes do not reproduce against it.) For those experiments, a shape that reproduced at the merge base was treated as pre-existing and went to the residual list rather than into the PR — unless closing it was needed to finish one of your named blockers, which is why a few explicitly pre-existing defects ARE fixed here. I started this response without that discipline — chasing an absolute "no reachable data loss" criterion — and spent about ten rounds building the purge redesign above before the first merge-base control showed the shapes I was chasing were already there and worse. The three-rail proof came from the same mistake one layer down. Both are gone; the branch is 8 commits and +3,396/-42, down from 35 commits and +4,534/-81 at its worst.

**Residual — surfaced along the way, NOT addressed here.** Evidence varies per item and each says so: some are executed merge-base controls, one is inspection only. None is a regression this branch introduces. I am listing them rather than filing issues so you can decide what is worth tracking.

- `sync.ts:3448` ("SP-5: both old and new slugs use `resolveSlugForPath`") is inaccurate on both sides — `oldSlug` can come from the DB batch resolve. Byte-identical to `913d2d7f:2819`, so it predates this branch.
- The liveness index enumerates the anchor tree only for fallback-regime paths (`sync.ts:1990`); ordinary paths come from `git ls-files` alone, so a committed live file that is staged for deletion drops out of the index and a row carrying its slug can be judged stale. Reachable by inspection — I have not executed this one, and master deletes in that shape unconditionally, so it is not a regression. Worth flagging only because the obvious fix is wrong: enumerating the anchor tree's ordinary paths too would register the rename's OLD path every time, so the stale row would look live forever and the reconcile would stop working entirely.
- The cheap move on master selects from the path resolve without an identity proof that the selected row is the renamed file's row; the decoy shapes above are reachable there today. I am explicitly NOT proposing the fix again, only noting the shapes exist.
- The content-signal comparison, if you ever do reach for it, is a necessary condition and not an identity proof: a byte-identical twin satisfies it. That is the trap the second row of the table describes.
- A rename that spares an unprovable row leaves a duplicate master would have deleted, and chained renames accumulate them (measured: three renames, 4 active rows here vs 1 at merge base). This one is a live tradeoff rather than a residual — it is the open question in the PR description, and I would rather you pick the direction.

Happy to turn any of these into issues, or to drop this comment entirely if it is noise — it is here because re-deriving a refuted mechanism is expensive and I already paid for these.

Masashi-Ono0611:
My human part:
I found that gbrain doctor goes fail because of the sentinel stuck.
On my personal GBrain env, this patch is working well. I believe this issue finally resolved🙇

Screenshot:
Screenshot 2026-08-02 at 1 51 34 PM
Screenshot 2026-08-02 at 1 51 45 PM

Masashi-Ono0611:
FYI — an update on how this has held up, not asking for anything:

This branch has stayed applied cleanly through 5 separate `gbrain upgrade`s since my last comment (v0.42.71.0 → v0.42.72.1 ×3 → v0.42.74.0, spanning 2026-08-02 to 2026-08-09), rebasing with zero conflicts and zero `.rej` files each time. Across that span the doctor FAIL from #3570 (the sentinel stuck) hasn't come back once — still clean today against current master.

No new evidence beyond "still working," just didn't want that data to be invisible while this sits in the queue.

Masashi-Ono0611:
FYI, not asking for anything specific — this has now been open 3 weeks. Since my last update (2026-08-09), it went from cleanly rebasing across 5 upgrades to now showing `CONFLICTING` against current master (v0.46.21.0), which is expected drift at this range rather than a regression in the fix itself. Still running clean on my own env with zero issues since the original fix. Happy to rebase whenever it's useful for review, or let me know if this should take a different shape.

time-attack:
Status update: reviewed in the 2026-08-18 full-backlog triage. It is queued but currently deferred: CONFLICTING/DIRTY on master; needs rebase, then security scan (security_flag) before re-queue. Address that and it re-enters the merge queue. Thanks for the contribution.

Masashi-Ono0611:
Rebased — this is ready to re-enter the queue.

- Head is now `f900f9f0`, rebased onto current `master` (`649ffe5f` is an ancestor of the head).
- GitHub reports `MERGEABLE` / `CLEAN`; the conflict is gone.
- CI at that head: 27 pass, 12 skipped, 0 failing.

One note for the `security_flag` scan you mentioned, so it is not a surprise when you run it: the tip commit (`f900f9f0`) suppresses a semgrep finding in `src/core/sync.ts`. It is a suppression added **after** the real fix, not instead of it — the preceding commit (`19e91583`) adds an explicit path-containment assertion in `fallbackSlugsForFile`, and the suppression exists because semgrep still flags the now-guarded join. Worth reading those two commits together rather than the tip alone.

No action needed from me otherwise — happy to rebase again if it goes stale while queued.

time-attack:
Re-trialed against current master: still conflicting (sync surface moved again in v0.46.23.0). The rename-sentinel + orphan self-heal work remains wanted — a rebase over master puts it back in review.

Masashi-Ono0611:
Rebased over master at 08b43cf4 — mergeable. The only red job is the pre-existing test/e2e/sync-delegation-under-serve.serial.test.ts Pin 4 flake (also failing on unmodified master), not caused by this change. Ready for re-trial.

Masashi-Ono0611:
Rebased over master at 680c33b3 (v0.46.25.0, #4368) — mergeable, CI in progress.

The 8-15 conflict was mostly mechanical: `updateSlug` target selection was
already carrying the three-rail proof from the prior review round, so it
resolved cleanly against #4368's parallel sync-safety work. Full local
verification before push: typecheck clean, module-size ratchet green,
324 sync-related tests pass (rename-reconcile, failures, batch, ledger,
classifier-widening, reconcile-db-only, mass-delete, delete-batch, resumable
import, monorepo, and more).

Ready for re-trial whenever it's convenient.

Masashi-Ono0611:
Rebased onto current master (v0.46.28.0) — the branch had been sitting on a v0.46.25.0 base. Conflicts: the sync.ts import block (union with master's image-sync imports), `module-size-limits.tsv` (carried master's megawave ceilings; sync.ts raised 4512→5442 for this branch's rename-reconcile machinery, note updated in the row), and `cli-flag-registry.generated.ts` (regenerated from the rebased tree rather than hand-merged — 100 commands / 4759 flag entries). Both sync suites pass (118/118 across sync-rename-reconcile.serial + sync-failures), typecheck + check:module-size clean. Issue #3570 remains OPEN and was not part of the v0.46.28.0 megawave, so the fix itself is unchanged.

time-attack:
Re-trial vs current master: two conflicts. (1) src/commands/sync.ts sentinel-clear line — master's #4496 added the `!importErrored` condition there; combine it with your renameSentinelPath change rather than replacing it. (2) scripts/module-size-limits.tsv sync.ts row — recount after merge. Also flagging: the PR silently drops a hunk during rebase (see the trial notes) — please double-check your sentinel deletion still covers the error path. Nightly re-trials will pick this up automatically once pushed.

Masashi-Ono0611:
Rebased/merged onto current master and resolved both flagged conflicts:

1. succeededPaths.push/markCompleted now use renameSentinelPath(to) and !importErrored, combining your #4496 condition with the sentinel-path change rather than replacing it.
2. scripts/module-size-limits.tsv's sync.ts ceiling recounted post-merge (5462 -> 5468, see below).

On the "hunk silently dropped" flag: I re-checked the delete-side error path (staleSlug/reconcileFailed/sentinel recording around the reconcile try/catch) against current master — nothing is missing there.

While double-checking the merge I found a real gap in the !importErrored condition itself: importFile can return status: 'skipped' with .error set (e.g. a frontmatter/path slug mismatch) without setting importErrored, so a rename whose destination fails that way still got checkpointed as done — the exact bug class this condition exists to prevent. Fixed (importErrored = true on that branch) with a red/green discrimination test (test/sync-rename-reconcile.serial.test.ts): without the fix, a second sync with unchanged broken content reports synced instead of blocked_by_failures. Ceiling bumped again for the +6 lines (5462 -> 5468).

Separately: this branch's new test/e2e/sync-reconcile-postgres.test.ts isn't wired into e2e.yml's tier1 job (only reachable via the nightly coverage-full-e2e glob) — filed as #4568 rather than guessing at a workflow-file fix without a local Postgres/Docker environment to verify it.

Masashi-Ono0611:
Closing — the rename-sentinel operator-exit safety fix from #3570 has been absorbed by an independent implementation (`scripts/module-size-limits.tsv` on upstream/master, `sync.ts` row: "wave-k: #3570 rename-sentinel operator-exit safety (cross-source-scoped resolve, tracked-file liveness index, orphaned-sentinel self-heal)"), consistent with #3570 having been closed as COMPLETED directly.

Comparing wave-k's implementation against this branch's final state, a few review-driven fixes made on top of the original design during this PR's lifecycle aren't present there:

- restoration of the `#3942`-guarded resolver (`resolveSlugsForRemovedPaths`/`resolveRemovedPathSlug`) for both the batched and no-sourceId rename lanes — wave-k still uses the unguarded `engine.resolveSlugsByPaths` + raw `resolveSlugForPath`/`resolveSlugByPathOrSourcePath` fallback this PR replaced (a Codex-review-driven fix restoring foreign-origin protection)
- setting `importErrored` on skipped+error rename destinations (a maintainer-lens-review-driven fix)
- a related sentinel-value comparison correction tied to the guard fix
- the regression test covering these (`test/sync-rename-reconcile.serial.test.ts`: "#3583 review: anchor-only liveness can bank a fallback rename duplicate")

I'll follow up with a narrower PR against current master reapplying just this residual delta. Closing this one to avoid confusion with the wave-k landing.

PR #3584, opened 2026-07-29
feat(embed): apply Qwen3's query-side Instruct template — close the asymmetric-template gap

paul-0320:
## Problem — the asymmetric-template gap

Qwen3-Embedding is an asymmetric instruction-tuned family: the **query side** is trained on an `Instruct: {task}\nQuery:{query}` template, while the document side is embedded raw. The [model card](https://huggingface.co/Qwen/Qwen3-Embedding-8B) is explicit that omitting the query-side instruction costs **1–5% retrieval quality**.

gbrain already knows which embeds are query-side — `embedQuery()` threads `inputType: 'query'` into every search embed — but that signal only ever reaches the **wire level**: `dimsProviderOptions` emits `input_type` for providers whose APIs take the field (ZE, Voyage, NVIDIA). For qwen3-embedding on the openai-compatible path the field doesn't exist, so the signal is silently dropped ([dims.ts#L314](https://github.com/garrytan/gbrain/blob/913d2d7f/src/core/ai/dims.ts#L314) is dim-only) and **every query is embedded document-side** — the exact degradation the model card warns about, on a model the recipe explicitly supports for Matryoshka dims.

## What this does

One seam in `gateway.embed()`, immediately before the MAX_CHARS cap ([gateway.ts#L1681](https://github.com/garrytan/gbrain/blob/913d2d7f/src/core/ai/gateway.ts#L1681)):

- When `inputType === 'query'` **and** the model id matches the qwen3-embedding family (the same match as dims.ts' Matryoshka branch), prepend the model card's template with its stock English task sentence.
- `GBRAIN_QUERY_INSTRUCT` overrides the task sentence — retrieval quality is corpus/language-sensitive (data below); an **empty string disables** the template entirely (operational kill switch).
- **Cache safety**: the template changes what `embedQuery()` produces, so `KNOBS_HASH_VERSION` bumps 14→15 (same one-time cold-miss class as the 10→11 `input_type` bump), and the effective sentence folds into the key as an append-only `qi=` part via `ctx.queryInstruct` — processes running different sentences can never cross-serve rows (the same per-process contamination pattern as #2825's `hardExcludes`).

**No re-embedding, no migration.** Document-side inputs and every stored artifact are untouched — raw document vectors are already this family's correct usage. The template rides the embedding input only, the same input-only invariant as the contextual-retrieval wrapper (`embedding-context.ts`).

Out of scope, deliberately: `nomic-embed-text` has the same class of gap (its model card mandates `search_query:` / `search_document:` prefixes) but the **document side** is affected too, which means an existing-corpus re-embed/migration story — a separate change. Hosted Qwen3 derivatives with a server-side instruct parameter (DashScope `text-embedding-v4`) have different model ids and are untouched.

## Measurements

**Vector-layer isolation** (Ollama `/v1/embeddings` + exact cosine scan, 3.1K-page production corpus, 38 Korean conversational queries, sign test two-sided): template vs none improved **14/1, p = 0.001** (top-1 17→24, MRR 0.573→0.715). English-only and identifier-heavy query batches sat at ceiling and were unaffected. A falsification control (unrelated-sentence prefix, same format) was random (p = 0.75) and pushed 3 queries out of top-50 — the instruction's *content* is what works, not the format.

**End-to-end** (production pipeline: hybrid + multi-query expansion + reranker + autocut, 62 queries, cache purged per condition): the biggest effect is **candidate-set recall** — misses (answer absent from top-20) dropped **13→9**, with 6 previously-undetected answers recovered (3 straight to rank 1); top-5 48→52. Top-1 dipped 36→33 (rank-boundary shuffling near the autocut line). Per-question data is in the eval attachment below.

**Eval replay characterization** (62 queries, off → template):

| metric | value |
|---|---|
| Jaccard@10 | 0.672 |
| top-1 stability | 45/62 (72.6%) |
| latency Δ | ≈ 0 (6.22s → 6.14s avg) |

**Language sensitivity** (why the override matters): on our Korean-heavy corpus a Korean task sentence measured significantly ahead (vector-layer 14/1 p=0.001 vs the English default's not-significant 13/6 p=0.167; E2E misses 13→9 for both). We run the Korean sentence in production via `GBRAIN_QUERY_INSTRUCT`; the in-tree default stays the model card's stock English sentence.

This change has been running in our production brain (Ollama qwen3-embedding:8b-fp16 @ 1536d MRL, ~22K chunks) since 2026-07-29.

## Verification

```
bun test test/embed-query-instruct-prefix.serial.test.ts   # 8 tests — outbound body via real-SDK mocked-fetch
bun test test/search/knobs-hash-query-instruct.test.ts     # qi= participation
bunx tsc --noEmit
```

The serial suite pins: template on the query side (+ Matryoshka `dimensions` untouched), raw document side for the same model, non-qwen3 models untouched (including a lookalike id), override + kill switch, template survival under MAX_CHARS truncation, and `effectiveQueryInstruct()` (the cache-key feed) mirroring every decision. Existing KNOBS_HASH_VERSION pins updated (4 files).


time-attack:

The engineering is sound and I could not break it — the flagged asymmetric-template danger is not real, and the strongest failure hypothesis I could construct (semantic-cache false hits from the shared prefix) measured clean in the opposite direction on the real model. Verified empirically against qwen3-embedding:0.6b on local Ollama rather than by argument: documents stay raw in BOTH conditions (the seam only fires on inputType==='query'), and query->document separation IMPROVES (mean true-match 0.7059->0.6921, mean unrelated 0.1993->0.0988, separation 0.5066->0.5933). Raw docs + templated queries IS this family's trained usage, so "no re-embed, no migration" holds and existing brains are not degraded. The discrimination test passes: stubbing applyQueryInstruct back to the old no-op fails 4/11 tests in the new serial suite, and deleting the qi= knobsHash part fails 2/5 in the new knobs test — the tests are not vacuous. Cache-key fold verified working on the default column (I initially suspected embeddingModel==='' defeated it; that '' is normalizeEngineColumn, engine-layer, not in the knobsHash path). All 5 KNOBS_HASH_VERSION pin sites updated, none missed. isQwen3EmbeddingModel is character-identical to dims.ts:314. Budget reserve computes totalChars post-prefix so added chars are billed. No engine/JSONB/migration/source-isolation/trust/dry-run surface touched. NEEDS_CHANGES is for two mechanical pre-merge items only — the IRON-RULE PR title and an undocumented operator kill switch — NOT for any code defect. No code change is required.

### Discrimination test

EXECUTED, both halves, and the tests discriminate. (1) Stubbed applyQueryInstruct's body back to the pre-PR no-op (`return texts;`) in src/core/ai/gateway.ts, re-ran test/embed-query-instruct-prefix.serial.test.ts: EXIT=1, 7 pass / 4 fail — the 4 failures are exactly the template-application assertions; the surviving 7 are the negative-scope tests (non-qwen3 stays raw) plus the effectiveQueryInstruct() unit tests, which are a separate function I did not stub. (2) Deleted the `qi=${ctx?.queryInstruct ?? 'none'}` line from the knobsHash parts array in src/core/search/mode.ts, re-ran test/search/knobs-hash-query-instruct.test.ts: EXIT=1, 3 pass / 2 fail — the two cross-serve-prevention assertions fail; the 3 survivors are the determinism/legacy-stability/fixture-sanity tests, which are correctly insensitive to the new part. Both stubs reverted from backups; worktree confirmed clean via git status --porcelain. Contrast with round 1's #3563 (7 of 8 new tests passed on master): this PR's tests fail when the behavior is removed.

### Blockers

**1.** PR title 'feat(embed): apply Qwen3's query-side Instruct template — close the asymmetric-template gap' has no vMAJOR.MINOR.PATCH.MICRO prefix, violating the CLAUDE.md IRON RULE ('Every PR title MUST start with the version'). The merge commit subject inherits the title, so this lands version-less in git log and breaks the version-first convention the PR list/merge commit/changelog all read from. Fix: gh pr edit 3584 --title "vX.Y.Z.W feat(embed): ..."

**2.** GBRAIN_QUERY_INSTRUCT (src/core/ai/gateway.ts:1710) is a new user-facing operator knob whose empty-string form is the ONLY kill switch for this feature, and it is documented in exactly zero shipped artifacts: no docs/, no README.md, no docs/architecture/KEY_FILES.md, no CHANGELOG.md. Grep confirms the string appears only in gateway.ts, mode.ts comments, and the two new test files. Every comparable env knob is documented (GBRAIN_SEARCH_EXCLUDE and GBRAIN_SOURCE_BOOST in KEY_FILES.md:72, the six GBRAIN_SYNC_* in CLAUDE.md). Failure scenario: an operator hits the 1-5% regression case or a bad interaction in production, needs to disable the template, and has no discoverable way to learn the switch exists — the incident escape hatch is unreachable at incident time.

### Not verified

- Absolute-magnitude generalization of my two measurements: I used qwen3-embedding:0.6b (1024d, the tag already pulled locally) rather than the 8b-fp16 @ 1536d MRL the author runs. Same family, same training template, so the DIRECTION of both results (separation improves; query-query similarity falls) is trustworthy, but the exact deltas will differ at 8b. I did not pull 8b — multi-GB download plus inference load on a shared box.
- The PR's headline retrieval numbers (14/1 sign test p=0.001, E2E misses 13->9, Jaccard@10 0.672, top-1 stability 45/62) are unreproduced. They need the author's 3.1K-page Korean corpus and 38/62-query eval sets, neither of which is in the diff. My independent measurement corroborates the DIRECTION on 5 hand-built query/doc pairs; it does not confirm their magnitudes.
- One of my 5 measured query/doc pairs moved against the trend: 'recipe for sourdough starter' true-match cosine dropped 0.6612 -> 0.5383 (-0.12). Separation still improved for it because the unrelated-doc score fell further, but this confirms the template does NOT uniformly preserve absolute query->document cosine. I grepped and found no absolute-cosine gate on the query side today (see invariants), so nothing consumes that number now — but any FUTURE absolute similarity floor would be calibrated against the wrong distribution.
- Empty-query behavior change: pre-PR embedQuery('') sent input:[''] to the provider; post-PR it sends the bare instruction, so an empty query would retrieve pages nearest the instruction sentence instead of failing/erroring. I did not trace whether hybrid.ts's upstream guards make an empty query reachable at this seam, so I cannot say it is live. Cosmetic sibling: resolveQueryInstruct() returns the env value untrimmed when non-empty, so GBRAIN_QUERY_INSTRUCT=' foo ' embeds and hashes the padding.
- gbrain eval --strategy vector (src/core/search/eval.ts:237) embeds the query via embed() = DOCUMENT side, so that diagnostic arm now measures un-templated queries while production templates them. This is a pre-existing divergence identical to the v=11 input_type gap, widened here. It is not the hybrid production path, and I did not run the eval suite (footprint).
- Model-id coverage is narrower than the model family: only exact 'qwen3-embedding' or 'qwen3-embedding:*' match, case-sensitive. Users on 'Qwen/Qwen3-Embedding-8B' (HF TEI / vLLM / DeepInfra / Together / OpenRouter) or 'dengcao/Qwen3-Embedding-8B:Q8_0' (a widely-pulled community Ollama tag) get no template. DEFENSIBLE — it is character-for-character the existing Matryoshka match in dims.ts:314, so those users already get no Matryoshka dims either, and widening should widen both in lockstep as a separate change. I verified the consistency claim but did not survey real-world id distribution.
- GBRAIN_QUERY_INSTRUCT is read from process.env directly rather than the gateway's injected `env` config plane (gateway.ts has only 5 process.env reads total). The author flags this as a deliberate C3 exception; I found no CI guard script enforcing that rule either way. Implication I did not chase: in a thin-client/remote-MCP deployment where gateway env arrives via config rather than process.env, the knob may be unsettable through the config plane.
- Skipped for execution-footprint reasons per instructions: bun run verify, bun run test, bun run ci:local, the full unit suite, all e2e/DATABASE_URL parity tests, and Docker/real-Postgres. I ran typecheck, the 2 new test files, and the 6 related existing files (155 tests, all green). No JSONB/engine/progress-to-stdout check-*.sh script is relevant to this diff, so none were run.
- VERSION / package.json / CHANGELOG.md are not bumped in this diff and the 3-line version-consistency audit was not run — /ship owns that at land time, so this is expected for a contributor PR, not a defect.

- `gh pr view 3584`
- `gh pr diff 3584`
- `git fetch origin pull/3584/head:pr3584 && git checkout pr3584`
- `grep -n qwen3 -ri src/core/ai/dims.ts   # confirms dims.ts:314 match is character-identical to isQwen3EmbeddingModel`
- `grep -rn 'embedQuery' src/   # enumerated every query-side caller`
- `grep -rn 'inputType' src/   # found the two probe sites (init-embed-check.ts:92, models.ts:484), both dim-only`
- `grep -rn 'knobsHash(' src/   # exactly ONE call site (hybrid.ts:1797), so the fold has no sibling gap`
- `grep -rn 'KNOBS_HASH_VERSION' test/ src/ scripts/ docs/   # all 5 pin sites updated; compiled-truth-boost-scope uses >=14`
- `sed -n 372,450p src/core/search/embedding-column.ts   # exported resolveEmbeddingColumn vs normalizeEngineColumn:529`
- `sed -n 25,56p src/core/ai/model-resolver.ts   # parseModelId('') throws`
- `sed -n 100,200p src/core/search/query-cache.ts   # cosine-on-query-embedding lookup at fixed 0.92 threshold`
- `sed -n 1755,1800p src/core/ai/gateway.ts   # tracker.reserve() uses post-prefix truncated array`
- `bunx tsc --noEmit   # EXIT=0`
- `bun test test/embed-query-instruct-prefix.serial.test.ts test/search/knobs-hash-query-instruct.test.ts   # EXIT=0, 16 pass`
- `bun test test/embed-input-type-wire.serial.test.ts test/search/knobs-hash-reranker.test.ts test/search-mode.test.ts test/cross-modal-phase1.test.ts test/search-alias-resolved-boost.test.ts test/search-compiled-truth-boost-scope.test.ts   # EXIT=0, 155 pass / 0 fail`
- `DISCRIMINATION 1: stubbed applyQueryInstruct -> no-op, bun test test/embed-query-instruct-prefix.serial.test.ts   # EXIT=1, 7 pass / 4 fail`
- `DISCRIMINATION 2: deleted qi= from knobsHash parts, bun test test/search/knobs-hash-query-instruct.test.ts   # EXIT=1, 3 pass / 2 fail`
- `ad-hoc bun test proof: resolveEmbeddingColumn(undefined, {embedding_model:'ollama:qwen3-embedding:8b-fp16'}) -> embeddingModel:'ollama:qwen3-embedding:8b-fp16' (NOT ''); descriptor fast-path -> undefined -> falls back to global; effectiveQueryInstruct('') -> undefined (unreachable path)`
- `curl -s http://127.0.0.1:11434/api/tags   # qwen3-embedding:0.6b already pulled locally`
- `bun scratch-cachecheck.mjs   # 12 distinct queries, 66 pairs, raw vs template: mean 0.3555->0.0981, max 0.5612->0.2551, pairs>=0.92: 0->0`
- `bun scratch-qd.mjs   # 5 query/doc pairs + 1 unrelated doc: true-match 0.7059->0.6921, unrelated 0.1993->0.0988, separation 0.5066->0.5933`
- `rm scratch-*.mjs test/proof3584.test.ts node_modules(symlink); git status --porcelain -> clean`

</details>

---

paul-0320:
Both blockers addressed — one pushed, one I need a number from you.

## Blocker 2 — undocumented kill switch: fixed in d9de611

`GBRAIN_QUERY_INSTRUCT` is now documented in `docs/architecture/KEY_FILES.md`, inside the `src/core/ai/gateway.ts` entry that owns the seam — the same placement as `GBRAIN_SOURCE_BOOST` / `GBRAIN_SEARCH_EXCLUDE` in the `source-boost.ts` entry you cited. It covers what makes the template fire (`inputType === 'query'` AND the qwen3-embedding family match), the task-sentence override and why it is corpus/language-sensitive, **the empty-string disable as the operator kill switch** (with a copy-pasteable one-shot A/B form), and the `effectiveQueryInstruct()` → `knobsHash` `qi=` cache-key fold.

`bash scripts/check-key-files-current-state.sh` → ok (reference docs history-free). `bash scripts/check-trailing-newline.sh` → ok. Docs-only, no behavior change, no new exports.

I deliberately left `CHANGELOG.md` alone — your own note says the VERSION/CHANGELOG bump is `/ship`'s at land time, so a contributor entry there would just conflict.

## Blocker 1 — version-first PR title: I can't close this one alone

The version in the title has to be the version this actually lands as, and that number is yours. The same review says: *"VERSION / package.json / CHANGELOG.md are not bumped in this diff ... /ship owns that at land time, so this is expected for a contributor PR, not a defect."* I'd be guessing (`v0.42.68.0`, off the current `VERSION` = 0.42.67.0), and a wrong number baked into the merge-commit subject is worse than an absent one.

Two data points for calibration rather than argument: none of my merged PRs here carried the prefix (#3477, #2866, #2865, #2864, #2717), and neither do the recent merges I checked (#3610, #3608, #3600, #3599, #3598). The rule reads as enforced on the `/ship` path rather than on contributor PRs.

Name the number and I'll set it immediately:

```
gh pr edit 3584 --title "vX.Y.Z.W feat(embed): apply Qwen3's query-side Instruct template — close the asymmetric-template gap"
```

Or set it yourself at land time if that's simpler.

## On the unreproduced headline numbers

Noted, and I can't close that gap: the 3.1K-page corpus and both eval sets are private company data and can't ship in the diff. Your independent 0.6b measurement (separation 0.5066 → 0.5933, documents raw in both conditions) is the right thing to rely on — treat my magnitudes as unverifiable and the direction as the claim.

## Two "Not verified" items I'm confirming rather than fixing here

- `resolveQueryInstruct()` returning the env value untrimmed is real — `GBRAIN_QUERY_INSTRUCT=' foo '` embeds and hashes the padding. Trimming it would change the hashed sentence, and I'd rather not touch the cache key twice in one PR. Happy to take it as a follow-up.
- `gbrain eval --strategy vector` embedding the query document-side is real and pre-existing — identical in kind to the v=11 `input_type` gap, widened here, not introduced. It's a diagnostic arm, not the hybrid production path. Also follow-up material.

Both are deliberate scope calls, not oversights. Say the word if you'd rather have either in this PR.

## Landing-order dependency

#3617 and #3621 also bump `KNOBS_HASH_VERSION` 14→15. Whichever two land second need a rebase (16 / 17); it's declared in all three PR bodies and I'll rebase whichever you take later.

time-attack:

Merge-lane despite the feat label: gbrain's existing inputType:'query' signal (threaded by embedQuery, gateway.ts:2060) is silently dead for qwen3-embedding on the openai-compat path, so every query embeds document-side against the model card's mandate — a correctness gap in a family gbrain explicitly supports (dims.ts:314 Matryoshka branch; the wave merged more qwen3/local fixes). Not superseded: current master gateway.ts has zero queryInstruct/instruct handling (grep clean); the wave's qwen3 items (#3699 dims suppression) are orthogonal. Verified in a /tmp worktree: core files (gateway.ts, hybrid.ts, embedding.ts) auto-merge clean; after resolving conflicts as 15→16 all 164 tests pass, typecheck clean, and neutering applyQueryInstruct makes 4 of the new serial tests fail (tests pin real outbound-body behavior). But it cannot land as-is — see conflict assessment.

**Conflict state vs current master:** Semantic version collision, not context drift: PR bumps KNOBS_HASH_VERSION 14→15, but master's #3677 (same author, merged 2026-07-31) already consumed v=15 for the fts= fold. 5 files conflict (src/core/search/mode.ts at the version const + the appended-parts list, plus 4 test files pinning the version: cross-modal-phase1, search-mode, search-alias-resolved-boost, search/knobs-hash-reranker). Correct resolution is mechanical and verified: bump to 16, keep BOTH the fts= and qi= parts, merge the pin-test comment lines (both sides edited the same test() declarations). The fix itself still applies fully post-v0.42.69.0 — gateway/hybrid/embedding hunks merged clean and behave as designed at v=16.

**Risk if merged as-is:** One-time global query-cache cold miss from the version bump (the documented pattern, same as #3677; refills within cache TTL). Behavior change is scoped to qwen3-embedding model ids on the query side only, with a GBRAIN_QUERY_INSTRUCT kill switch; no re-embedding, no migration, no engine/jsonb/source-scoping surface touched. Only landing hazard: a sloppy conflict resolution that keeps v=15 or drops one of the fts=/qi= parts — resolver must produce v=16 with both parts (even the sloppy case isn't cache-unsafe since the appended qi= part changes every hash, but test pins would disagree).

paul-0320:
Rebased onto current master (`2e554500`) — this had gone conflicting since #3677 merged and took `KNOBS_HASH_VERSION` 15.

Applied:
- `KNOBS_HASH_VERSION` now **18**. 15 is the merged #3677 (`fts=`), 16 is #3621 (`acm=`), 17 is #3617 (`kof=`) — this claims the next free number.
- `qi=` appended after `fts=` in the `knobsHash` array; both parts kept.
- Bump-rationale comments merged; the four version pins and their titles updated to 18.
- The docs commit for `GBRAIN_QUERY_INSTRUCT` (`docs/architecture/KEY_FILES.md`) carried through the rebase unchanged.

Verification on the rebased head: `bun run verify` 34/34 · `bunx tsc --noEmit` clean · 475/475 across the patch-adjacent suites plus this PR's own `test/embed-query-instruct-prefix.serial.test.ts` and `test/search/knobs-hash-query-instruct.test.ts`.

Coordination: #3621, #3617 and this PR were all conflicting on the same seam, so I rebased the three in a staircase — **#3621 → 16, #3617 → 17, this → 18** — so that taking any one of them doesn't force a renumber on the other two. Out-of-order merges just skip a number, which is harmless for a cache-busting counter.

Still open from your earlier review: the version-prefix question on the title. I couldn't self-assign a landing version, so the title is still unprefixed — same as the recent merges on master. If you'd like a specific number in it, tell me which and I'll amend.

paul-0320:
Rebased onto current master (`1ec6a6e8`, v0.45.9.0). One docs-only conflict — the `gateway.ts` entry in KEY_FILES.md had grown upstream in v0.44.1.0 — resolved by keeping upstream's text and re-appending the `GBRAIN_QUERY_INSTRUCT` paragraph. `KNOBS_HASH_VERSION` claim unchanged at 18 (`qi=`; 17 = #3617, 16 = #3621). Verify 38/38; instruct-prefix + knobs-hash suites 46/46. Both review blockers remain addressed — the only open item is the title version prefix, which is yours to assign. Merge-ready whenever a wave picks this lane up.

paul-0320:
CI note: the `test (9)` failure is not this PR's — it is a pre-existing cross-file state leak on master, now root-caused and fixed in #4061. `test/sweep.test.ts`'s serve harness drives `runServe()`, which flips console-prefix's module-global stdout→stderr redirect (#3844) and never resets it; this PR adds one test file, which reshuffled shard 9 and paired the leak with `test/sync-all-parallel.test.ts` for the first time. Two-line repro on unmodified master `1ec6a6e8`: `bun test test/sweep.test.ts test/sync-all-parallel.test.ts` → same failure, same assertion. The failing test passes 17/17 in isolation on this head, and sibling #3617 (same base, pushed minutes earlier) passes the same shard. With #4061's one-line reset applied, the full shard-9 file list in CI order goes 1302/0 locally.

time-attack:
Status update: reviewed in the 2026-08-18 full-backlog triage — still wanted, but not yet mergeable: Real gap (queries embed doc-side on qwen3), reviewer-vetted and actively rebased, but conflicting again on the single-writer KNOBS_HASH_VERSION. If you update it, it will be re-reviewed promptly; alternatively the fix may be adopted with credit in an upcoming fix wave. Thanks for the contribution and the patience.

paul-0320:
Rebased onto current master (`649ffe5f`, v0.46.21.0). Two things beyond the mechanical renumber, both forced by what landed upstream while this was open:

**1. `KNOBS_HASH_VERSION` 18 → 20** (18 went to `acm=`/#1863; 19 is the sibling #3617). `qi=` stays append-only after `det=`, all upstream parts kept; comment + four pin tests updated.

**2. New commit `refactor(embed): move the query-instruct seam out of gateway.ts`** — `gateway.ts` (4428) and `hybrid.ts` (2500) now sit *at* their module-size ceilings (`scripts/check-module-size.sh` runs in `verify`), so the ~80-line block you reviewed in place cannot land in `gateway.ts` any more without a large TSV bump. It moves verbatim into `src/core/ai/query-instruct.ts`; `gateway.ts` grows by the import line only and the call site is unchanged in effect (`applyQueryInstruct(texts, modelId, opts?.inputType)` right before the `MAX_CHARS` cap). Two `+1` TSV edits, both noted in the row (`gateway.ts` 4428→4429 for the import, `hybrid.ts` 2500→2501 for the `qi=` knobs-ctx line).
   - `isQwen3EmbeddingModel` now lives in `dims.ts` (exported) and is shared by the Matryoshka `dimensions` branch and the template. #4123 widened dims' match to the cased hub form (`Qwen/Qwen3-Embedding-8B`), so a private copy here would have silently broken this PR's own invariant ("same family match as dims") — your "widen both in lockstep" note. Consequence: hub-form ids get the template exactly where they get Matryoshka dims. The lookalike negative test moves to an id outside the family (`qwen3-embed-custom`); one positive assertion pins the lockstep.
   - `effectiveQueryInstruct(modelString?)` (cache-key feed, global-model default) lives in `embedding.ts` next to the other gateway-wrapping accessors — `hybrid.ts` already imported it from there. `query-instruct.ts` stays gateway-free (`queryInstructFor(modelString)`, pure, never throws).
   - KEY_FILES: the `GBRAIN_QUERY_INSTRUCT` paragraph moves to a `query-instruct.ts` entry; the kill switch is documented exactly as before.

Verification on the rebased head: `bun run verify` 54/54 · `bunx tsc --noEmit` clean · 1068/1068 across this PR's two suites + `test/ai/` + `embed-input-type-wire` + the knobs-hash/search-mode/cross-modal/alias-boost suites. Discrimination re-run after the move: stubbing `applyQueryInstruct` back to a pass-through fails 4 of 11 in `embed-query-instruct-prefix.serial` (same 4 as your round), restored → 11/11.

Staircase: #3617 → 19, **this → 20**, #3621 → 21. Still open from the first review: the title version prefix — the number is yours to assign.

time-attack:
Superseded: the autocut floor landed in v0.46.23.0 via the sibling #3621 adoption (ack= cache-key token, knobs-hash v19) — your test file is byte-identical to what shipped, which says the approach was right. Thank you for the contribution; credited lineage is in the CHANGELOG.

paul-0320:
This looks like a mix-up between two sibling PRs: the closing note describes **#3621** (autocut floor, `ack=` token, knobs-hash v19) — which was indeed adopted in v0.46.23.0 and correctly closed. This PR is the **query-side Instruct template** for qwen3-embedding (`qi=`, `GBRAIN_QUERY_INSTRUCT`), and none of it is on master: `src/core/ai/query-instruct.ts` doesn't exist and `GBRAIN_QUERY_INSTRUCT` has zero hits in the repo. The 08-18 triage assessment on this thread ("real gap, reviewer-vetted") still stands. Could you reopen? I'll rebase onto current master immediately after — holding off on pushing until then so the branch stays reopenable.

PR #3585, opened 2026-07-29
fix(import): checkpoint on a time interval and before preserve, not only every 100 files

Kyzcreig:
## Problem

`runImport` saves its resume checkpoint only when `completed.size % 100 === 0`. There is no time-based save, and no save before the clear/preserve decision.

On a corpus whose files are individually expensive, the next 100-boundary can be **hours** of work away — so a run killed before reaching it discards every file completed since the last boundary, and the import can never converge under a supervisor that bounds runtime.

## Measurements

From a 2,207-file markdown corpus on the Postgres engine:

- Chunking cost tracks **paragraph count**, not bytes. Reference-style pages (dictionaries, textbooks — tens of thousands of short entries) cost roughly **10x** a prose page of the same size.
- One 21k-paragraph file took **~11 minutes on its own**.
- At that rate `completed` advances about **one file per ten minutes**, putting the next 100-boundary **~16 hours** out.
- Under a supervisor that kills long runs, the import made **zero durable progress** across repeated attempts: the checkpoint sat at 1600 completed while the process burned ~30 minutes of real work each pass.

## Changes

**1. Save on a time interval as well as the count boundary.** Save when EITHER `completed.size % 100 === 0` OR `CHECKPOINT_MAX_INTERVAL_MS` (120s) has elapsed since the last save and at least one new file completed. Progress becomes durable independently of per-file cost.

**2. Save once more immediately BEFORE the existing clear/preserve block**, gated on `errors > 0` so it matches that block's preserve condition exactly. A run that ends between triggers otherwise leaves its tail unsaved.

Ordering matters here and is covered by a test: the final save runs *before* `clearCheckpoint()`, so a **clean run still ends with no checkpoint file**. It only makes the error path's already-preserved checkpoint complete.

## Tests

Both in `test/import-resume.test.ts`:

- **`interrupted run preserves its tail below the 100-file boundary`** — three good files plus one that raises, asserting the preserved checkpoint carries all three. **Fails without the fix.**
- **`clean completion still leaves no checkpoint`** — pins the ordering so the final save cannot resurrect a checkpoint on a clean run.

### Note for reviewers

A `SLUG_MISMATCH` fixture does **not** exercise this path — it is a soft `failures` entry that leaves `errors` at 0, so the checkpoint is cleared rather than preserved. The test uses a file that raises during import instead. (This cost me a debugging cycle; flagging it so it doesn't cost one in review.)

## Verification

| Gate | Result |
|---|---|
| `bun test` (import suites) | 45 pass / 0 fail |
| `bun run typecheck` | clean |
| `bun run check:all` | clean |
| `scripts/check-test-isolation.sh` | OK |

Reverting the source change makes the new test fail, and restoring it makes it pass — the test genuinely gates the fix.

## Scope

+95 / −2 across `src/commands/import.ts` and `test/import-resume.test.ts`. No schema, config, or API surface change.

Related but distinct: #1794 / #1980 fixed the same *class* of non-convergence on the **sync** path (the `last_commit` anchor, `src/core/op-checkpoint.ts`); that work never touched `src/commands/import.ts`, which still has the count-only trigger.

### DIFF
--- a/src/commands/import.ts
+++ b/src/commands/import.ts
@@ -268,6 +268,12 @@ export async function runImport(
   let skipped = 0;
   let errors = 0;
   let processed = 0;
+  // Time-based checkpoint floor (see the save site below). Chunking cost scales
+  // with paragraph count, not bytes, so a single reference-style file can take
+  // many minutes; a count-only trigger leaves that work undurable.
+  const CHECKPOINT_MAX_INTERVAL_MS = 120_000;
+  let lastCheckpointMs = Date.now();
+  let lastCheckpointSize = completed.size;
   let chunksCreated = 0;
   const importedSlugs: string[] = [];
   const errorCounts: Record<string, number> = {};
@@ -343,7 +349,17 @@ export async function runImport(
     // Save checkpoint every 100 SUCCESSFUL adds (not every 100 processed).
     // Failed files never enter `completed`, so a flaky file can't push the
     // checkpoint past it — the next run will retry it.
-    if (completed.size > 0 && completed.size % 100 === 0) {
+    // ...and ALSO save on a time interval. On a corpus with an expensive tail
+    // `completed` can advance ~1 file per several minutes, so the next
+    // 100-boundary may be hours away; any kill before it discards every file
+    // since the last boundary and the run can never converge.
+    const nowMs = Date.now();
+    const dueByCount = completed.size > 0 && completed.size % 100 === 0;
+    const dueByTime = completed.size > lastCheckpointSize
+      && nowMs - lastCheckpointMs >= CHECKPOINT_MAX_INTERVAL_MS;
+    if (dueByCount || dueByTime) {
+      lastCheckpointMs = nowMs;
+      lastCheckpointSize = completed.size;
       const cpDir = gbrainPath();
       if (!existsSync(cpDir)) {
         try { const { mkdirSync } = await import('fs'); mkdirSync(cpDir, { recursive: true }); }
@@ -427,6 +443,30 @@ export async function runImport(
     if (count > 5) {
       console.error(`  ${count} files failed: ${err.slice(0, 100)}`);
     }
+  }
+
+  // Final checkpoint save BEFORE the clear/preserve decision below. The
+  // periodic triggers above are gated on a 100-file boundary or an interval,
+  // so a run that ends between them would otherwise leave its tail unsaved.
+  // This must run before clearCheckpoint() so a clean run still ends with no
+  // checkpoint file — it only makes the ERROR path's preserved checkpoint
+  // complete.
+  if (errors > 0 && completed.size > lastCheckpointSize) {
+    try {
+      const cpDir = gbrainPath();
+      if (!existsSync(cpDir)) {
+        const { mkdirSync } = await import('fs');
+        mkdirSync(cpDir, { recursive: true });
+      }
+      saveCheckpoint(checkpointPath, {
+        schema_version: 1,
+        owner: 'gbrain',
+        kind: 'import',
+        dir,
+        completedPaths: Array.from(completed),
+        timestamp: new Date().toISOString(),
+      });
+    } catch { /* non-fatal: the next run simply redoes the tail */ }
   }
 
   // Clear checkpoint on clean completion. On error, the path-based checkpoint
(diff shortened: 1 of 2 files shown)


time-attack:

The code is genuinely good and I verified it end-to-end: runImport's checkpoint really did save only on completed.size % 100 === 0 with no final save before the clear/preserve decision (src/commands/import.ts), so a sub-boundary run killed or ending in errors lost its whole tail. The fix adds a 120s time-based trigger plus an error-gated final save before clearCheckpoint(), placed in the one shared function all three callers (cli.ts:1814, jobs.ts:1682, sync.ts:3721) route through. The discrimination test is real: with master's import.ts stubbed back in, the new tail-preservation test fails (6 pass/1 fail); with the PR source it passes (7/7). saveCheckpoint is atomic (tmp+rename) and self-swallowing, so the new periodic call can't crash a run. Typecheck clean, sibling suite (import-checkpoint, 24 tests) clean, no SQL/engine/scope/trust/knobs surface touched, no dry-run path, no new env vars. What blocks merge is pure ship mechanics: the title has no version prefix (IRON RULE: vMAJOR.MINOR.PATCH.MICRO first) and the branch carries no VERSION/package.json/CHANGELOG bump — branch VERSION is 0.42.67.0 while master is already 0.42.68.1, so the CI version-gate rejects the merge as-is. A maintainer takeover via /ship (assign next version slot, retitle version-first, add CHANGELOG entry) lands this with zero code changes needed. Cosmetic nit only: the first test comment says the fixture 'exceeds the content-sanity block threshold' but the actual fixture is a chmod-0o000 unreadable file (the second comment is correct).

### Discrimination test

Stubbed master's src/commands/import.ts back in (git checkout origin/master -- src/commands/import.ts) while keeping the PR's tests, then re-ran bun test test/import-resume.test.ts: 'interrupted run preserves its tail below the 100-file boundary' FAILED (6 pass / 1 fail, exit 1). Restored the PR source: 7 pass / 0 fail. The new test genuinely gates the fix. The second new test ('clean completion still leaves no checkpoint') passes on both, as the PR body itself states — it is an ordering regression guard, not the discriminator.

**Mergeability at review time:** mergeable=MERGEABLE, mergeStateStatus=UNSTABLE (fork PR, 'no checks reported on the branch' — no CI has run; no merge conflict)

### Blockers

**1.** PR title lacks the version-first prefix required by the repo IRON RULE (must be 'vX.Y.Z.W fix(import): ...'; currently 'fix(import): checkpoint on a time interval...')

**2.** No VERSION/package.json/CHANGELOG bump: branch VERSION is 0.42.67.0 but master is already 0.42.68.1, so the CI version-gate rejects the merge until a strictly-greater version lands on the branch (or a maintainer takes it over via /ship)

### Not verified

- The PR body's corpus measurements (2,207-file corpus, 21k-paragraph file at ~11 min, ~16h to next boundary) — not reproducible without the reporter's corpus; the mechanism they describe (count-only trigger, no final save) is verified directly from master's source instead
- PR body's claimed 'bun test (import suites) 45 pass' — I ran the two directly-relevant files (7 + 24 tests), not every import-adjacent suite, per execution-footprint limits
- Full CI rollup — no checks have run on this fork branch (mergeStateStatus UNSTABLE); typecheck and touched-file tests run locally instead
- chmod-0o000 test fixture behaves as intended on non-root runners (verified locally on macOS); would misbehave only on a root-run CI, which GitHub Actions is not

- `gh pr view 3585 --json number,title,mergeable,mergeStateStatus,baseRefName,headRefName,author,body`
- `gh pr view 3585 --json mergeable,mergeStateStatus (re-poll: MERGEABLE/UNSTABLE)`
- `gh pr diff 3585 > /tmp/pr3585.diff (read every hunk, 140 lines)`
- `gh pr checks 3585 (no checks reported)`
- `git fetch origin pull/3585/head:pr-3585 && git checkout pr-3585`
- `git show origin/master:VERSION (0.42.68.1) vs branch VERSION (0.42.67.0)`
- `bun install (worktree shipped without node_modules)`
- `bun test test/import-resume.test.ts (PR branch: 7 pass / 0 fail)`
- `git checkout origin/master -- src/commands/import.ts && bun test test/import-resume.test.ts (reverted: 6 pass / 1 fail — new tail test fails, discrimination confirmed)`
- `git checkout pr-3585 -- src/commands/import.ts (restore)`
- `bun run typecheck (EXIT=0, clean)`
- `bun test test/import-checkpoint.test.ts (24 pass / 0 fail)`
- `grep for all runImport callers (cli.ts, jobs.ts, sync.ts — all route through the fixed function)`
- `read src/core/import-checkpoint.ts saveCheckpoint (atomic tmp+rename, error-swallowing)`
- `grep -n dry src/commands/import.ts (no dry-run path exists)`

</details>

---

time-attack:

PR #3587, opened 2026-07-29
feat(ai): codex-cli recipe — GPT chat via Codex CLI OAuth subscription

runsmlee:
# feat(ai): `codex-cli` recipe — GPT chat via Codex CLI OAuth subscription

## What

The GPT sibling of the `claude-cli` recipe (#334 lineage). ChatGPT Plus/Pro subscribers can route `gateway.chat` / `gateway.toolLoop` through the local `codex` CLI's OAuth session instead of paying per-token `OPENAI_API_KEY` charges:

```bash
gbrain config set chat_model codex-cli:gpt-5.6-terra
# or as a fallback behind claude-cli:
gbrain config set chat_fallback_chain '["codex-cli:gpt-5.6-terra"]'
```

`codex-cli:gpt-5.6-terra` (subscription) sits alongside `openai:gpt-5.6` (API key + per-token billing) the same way `claude-cli:*` sits alongside `anthropic:*`.

## Why

Same operator need that motivated claude-cli in #334, on the GPT side: Minions subagent dispatch and chat fallback against an existing subscription. Mixed-provider fallback chains (`claude-cli:* → codex-cli:*`) currently have no subscription path for the GPT leg — this closes that.

## How (mirrors `claude-cli-language-model.ts`)

- **Transport**: `codex exec -` with the prompt on **stdin** (argv has a hard size ceiling; subagent prompts routinely exceed it). Final message read from `-o <file>` — stdout is progress logs and is discarded.
- **Tool use**: the same `<use_tools>[{id,name,input}]</use_tools>` prompt protocol claude-cli teaches; parallel calls round-trip into `LanguageModelV2` `tool-call` parts. Protocol helpers are mirrored, not extracted, to keep this PR additive-only — a shared `cli-tool-protocol` module is a natural follow-up now that two providers prove the shape (happy to do it in this PR if you prefer).
- **Agent isolation**: `--ignore-user-config` (no user MCP servers — including gbrain's own MCP → recursion + PGLite single-writer lock contention), `--sandbox read-only`, dedicated tmpdir cwd, `--skip-git-repo-check`. Codex has no `--system-prompt` flag, so system messages render as a leading `## System` stdin section.
- **Billing guard**: `OPENAI_API_KEY` / `OPENAI_BASE_URL` scrubbed from the child env so an inherited key never silently flips billing to the API (mirrors claude-cli's `ANTHROPIC_*` scrub).
- **Usage tokens**: honest-`undefined` — the `-o` channel carries no token accounting, and fabricating numbers would poison the budget ledger. Subscription billing makes the ledger nominal for this recipe anyway (per the recipe comment convention).

## Touched

| File | Change |
|---|---|
| `src/core/ai/recipes/codex-cli.ts` | new recipe (chat-only, `auth_env.required: []`) |
| `src/core/ai/providers/codex-cli-language-model.ts` | new `LanguageModelV2` subprocess provider |
| `src/core/ai/types.ts` | `Implementation` union + `'codex-cli'` |
| `src/core/ai/recipes/index.ts` | registration |
| `src/core/ai/gateway.ts` | 3 switch branches (embedding error / expansion / chat), mirroring claude-cli |
| `test/codex-cli-recipe.test.ts` | 19-test suite |

## Testing

- `test/codex-cli-recipe.test.ts` — 19 tests mirroring `test/claude-cli-recipe.test.ts`'s shell-stub strategy (no Codex install or subscription needed in CI): registration, text round-trip, tool-call extraction (single / parallel / fenced / malformed / unterminated / declined), argv isolation flags + stdin marker + `## System` section, `OPENAI_*` env scrub, abort SIGTERM, non-zero exit, empty `-o` file, missing binary, `doStream` not-supported, cwd-reachability hint.
- `bun test test/ai/` — 472 pass, 0 fail (no regressions).
- `bunx tsc --noEmit` — clean (the `Implementation` union extension exercises every exhaustive switch).
- Live smoke: `doGenerate` against real Codex CLI 0.145.0 / `gpt-5.6-terra` — system + user prompt → exact text back, `finishReason: 'stop'`.

## Notes for review

- Model list ships with `gpt-5.6-terra` / `gpt-5.6-sol` (verified against `codex exec -m` on 2026-07-29); aliases `terra` / `sol`. Happy to adjust to whatever id set you prefer to pin.
- Cost figures are the nominal API-tier numbers for the budget ledger, `price_last_verified: 2026-07-29`, same convention as claude-cli.
- A `grok-cli` recipe following the same shape is a natural sibling; kept out of this PR to keep review scope tight.

---

## Update — addressing @kweiner's review (9152437)

Thanks for testing this against a real self-hosted install; the snap finding is real and reproduced as described.

**Confirmed diagnosis, different fix.** The scratch cwd is passed as both `-C` and the parent of the `-o` output file, so a sandbox-confined child has to be able to open it — under snap it cannot, and the CLI reports a bare `os error 2` that names neither the path nor the cause.

I went with your option 3+ rather than option 1: the failure now names the cwd and the `TMPDIR` escape hatch, and the host requirement is documented. Reasoning for not moving the default to `$HOME`:

- `CODEX_CWD` has no cleanup path (only `outFile` is removed), and it is per-pid — under `/tmp` the OS reclaims it, under `$HOME` it would accumulate one empty directory per process, forever, on every host.
- It would trade a one-line env export on a minority of hosts for home-directory clutter on all of them.
- A dotdir default (`~/.gbrain/...`) would not fix snap anyway: the `home` interface permits only **non-hidden** paths under `$HOME` — which is also why your `TMPDIR=/home/ken/some-dir` repro worked. Worth flagging in case it saves someone the same hour.

Happy to switch to a `$HOME` default (with a fixed path + cleanup, not per-pid) if you'd rather it work with no configuration on snap — your call.

**Also folded in:** your other two findings, as `docs/ai-providers/codex-cli.md` — minimum CLI 0.145.0 (with the `gpt-5.3-codex`-under-ChatGPT-auth error as the symptom of an older snap), the `TMPDIR` workaround, and `ca-certificates` for minimal Docker bases whose image ships no OS trust store.

**Corrected:** the test count in this description said 20; it was 18. Now 19 with a test covering the hint (and that ordinary CLI failures stay clean). `bun test test/ai/` still 472 pass / 0 fail, `bunx tsc --noEmit` clean.

### DIFF
--- a/docs/ai-providers/codex-cli.md
+++ b/docs/ai-providers/codex-cli.md
@@ -1,0 +1,97 @@
+# codex-cli — GPT chat via the Codex CLI OAuth subscription
+
+The `codex-cli` recipe routes `gateway.chat` / `gateway.toolLoop` through the
+local `codex` CLI's ChatGPT OAuth session, so a ChatGPT Plus/Pro subscription
+covers the calls instead of per-token `OPENAI_API_KEY` billing. It is the GPT
+sibling of the `claude-cli` recipe.
+
+```bash
+gbrain config set chat_model codex-cli:gpt-5.6-terra
+# or behind claude-cli in a fallback chain:
+gbrain config set chat_fallback_chain '["codex-cli:gpt-5.6-terra"]'
+```
+
+`codex-cli:gpt-5.6-terra` (subscription) sits alongside `openai:gpt-5.6`
+(API key, per-token billing) the same way `claude-cli:*` sits alongside
+`anthropic:*`. Token usage is not reported on the CLI's output channel, so
+usage fields are undefined and the budget ledger treats these calls as
+nominal.
+
+`GBRAIN_CODEX_CLI_BIN` overrides the binary path if `codex` is not on `PATH`.
+
+## Host requirements
+
+### CLI version
+
+Needs a `codex` new enough to have `--ignore-user-config` and the
+`gpt-5.6-terra` / `gpt-5.6-sol` model ids — 0.145.0 or later. Older builds
+fail in confusing ways: the Ubuntu snap's stable channel lagged at 0.114.0,
+whose default `gpt-5.3-codex` is not available under ChatGPT-account auth at
+all (`{"detail":"The 'gpt-5.3-codex' model is not supported when using Codex
+with a ChatGPT account."}`). `sudo snap refresh codex` if you are on the snap.
+
+### Sandbox-confined installs (snap) — set TMPDIR
+
+Each call spawns `codex exec` from a dedicated scratch directory, passed as
+both `-C` (so `AGENTS.md` auto-discovery finds no local files) and the parent
+of the `-o` output file. The child must therefore be able to open that
+directory.
+
+The default location is under `os.tmpdir()`, which a sandbox-confined `codex`
+cannot reach: snap's `home` interface permits only **non-hidden** paths under
+`$HOME`, so nothing in `/tmp` is accessible. The failure surfaces as a bare
+errno that names neither the path nor the cause:
+
+```
+codex-cli exited 1: Error: No such file or directory (os error 2)
+```
+
+Point `TMPDIR` at a non-hidden directory inside `$HOME`:
+
+```bash
+export TMPDIR="$HOME/gbrain-tmp"
+```
+
+A hidden directory (`~/.gbrain/tmp`) does **not** work under snap
+confinement — the `home` interface excludes dotfiles. This is also why the
+adapter cannot simply default to a dotdir in `$HOME`.
+
+### Minimal Docker images — install ca-certificates
+
+`codex` is a separate Rust process that trusts the OS certificate store,
+unlike Bun's `fetch` which bundles its own roots. Base images that ship no
+trust store at all (e.g. `oven/bun:1`) make every `chatgpt.com` /
+`api.openai.com` call fail with:
+
+```
+invalid peer certificate: UnknownIssuer
+```
+
+Add the trust store to the image:
+
+```dockerfile
+RUN apt-get update && apt-get install -y ca-certificates && rm -rf /var/lib/apt/lists/*
+```
+
+(`apk add --no-cache ca-certificates` on Alpine/musl.)
+
+## Isolation notes
+
+The subprocess is deliberately kept close to a raw LLM rather than a full
+Codex agent:
+
+- `--ignore-user-config` stops `~/.codex/config.toml` from loading, so user
+  MCP servers stay out — including gbrain's own MCP, which would otherwise
+  recurse and contend for the PGLite single-writer lock. Auth state
+  (`auth.json`) still loads.
+- `--sandbox read-only` pins the agent sandbox down for defense in depth.
+- `--skip-git-repo-check` skips the repo probe in the scratch cwd.
+- `OPENAI_API_KEY` / `OPENAI_BASE_URL` are scrubbed from the child env, so an
+  inherited key cannot silently flip billing back to per-token API usage.
+- Codex has no `--system-prompt` flag, so system messages render as a leading
+  `## System` section of the stdin prompt. The prompt goes over stdin because
+  argv has a hard size ceiling that subagent prompts routinely exceed.
+
+Tool use rides the same prompt-instructed
+`<use_tools>[{id,name,input}]</use_tools>` protocol the `claude-cli` provider
+teaches.
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -1531,6 +1531,10 @@ function instantiateEmbedding(recipe: Recipe, modelId: string, cfg: AIGatewayCon
       throw new AIConfigError(
         `claude-cli has no embedding model. Use openai or google for embeddings.`,
       );
+    case 'codex-cli':
+      throw new AIConfigError(
+        `codex-cli has no embedding model. Use openai or google for embeddings.`,
+      );
     case 'openai-compatible': {
       // D12=A: unified auth via Recipe.resolveAuth (or default).
       const auth = applyResolveAuth(recipe, cfg, 'embedding');
@@ -2484,6 +2488,12 @@ function instantiateExpansion(recipe: Recipe, modelId: string, cfg: AIGatewayCon
       const { ClaudeCliLanguageModel } = require('./providers/claude-cli-language-model.ts');
       return new ClaudeCliLanguageModel(modelId);
     }
+    case 'codex-cli': {
+      // Same subprocess pattern as claude-cli: the Codex CLI owns auth
+      // (ChatGPT OAuth session); chat-only recipe, no expansion touchpoint.
+      const { CodexCliLanguageModel } = require('./providers/codex-cli-language-model.ts');
+      return new CodexCliLanguageModel(modelId);
+    }
     case 'openai-compatible': {
       // D12=A: unified auth via Recipe.resolveAuth (or default).
       const auth = applyResolveAuth(recipe, cfg, 'expansion');
@@ -3045,6 +3055,14 @@ function instantiateChat(recipe: Recipe, modelId: string, cfg: AIGatewayConfig):
       // openai-compatible path below. No env-var switch, no global flag.
       const { ClaudeCliLanguageModel } = require('./providers/claude-cli-language-model.ts');
       return new ClaudeCliLanguageModel(modelId);
+    }
+    case 'codex-cli': {
+      // The GPT sibling of claude-cli: `codex exec` owns auth (ChatGPT
+      // OAuth session managed by `codex login`). `codex-cli:gpt-5.6-terra`
+      // lands here; `openai:gpt-5.6` continues through native-openai with
+      // per-token API billing. No env-var switch, no global flag.
+      const { CodexCliLanguageModel } = require('./providers/codex-cli-language-model.ts');
+      return new CodexCliLanguageModel(modelId);
     }
     case 'openai-compatible': {
       // D12=A: unified auth via Recipe.resolveAuth (or default).
--- a/src/core/ai/providers/codex-cli-language-model.ts
+++ b/src/core/ai/providers/codex-cli-language-model.ts
@@ -1,0 +1,435 @@
+/**
+ * ai-sdk LanguageModelV2 implementation that dispatches via the `codex exec`
+ * CLI subprocess. Used by the `codex-cli` recipe to route gateway.toolLoop /
+ * gateway.chat calls through the Codex CLI's ChatGPT OAuth session instead of
+ * the OpenAI SDK + OPENAI_API_KEY.
+ *
+ * Per-call routing is the contract: the gateway resolves the model string
+ * to this recipe based on the `codex-cli:` prefix, instantiates one of
+ * these objects per modelId, and dispatches doGenerate. Sibling subagent
+ * jobs with `claude-cli:...` or `litellm:...` continue routing through their
+ * own providers in the same worker; no env-var switch, no global state.
+ *
+ * Tool use is supported via prompt-instructed JSON emission — the SAME
+ * `<use_tools>[{id,name,input}, ...]</use_tools>` protocol the claude-cli
+ * provider teaches (see claude-cli-language-model.ts). The protocol helpers
+ * are mirrored here rather than extracted so this change stays additive-only;
+ * extraction into a shared cli-tool-protocol module is a natural follow-up
+ * once two providers prove the shape.
+ *
+ * Context isolation:
+ *   - Spawned from a dedicated tmpdir (`-C`) so AGENTS.md auto-discovery has
+ *     no local files to find; `--skip-git-repo-check` skips the repo probe.
+ *   - `--ignore-user-config` stops ~/.codex/config.toml from loading — user
+ *     MCP servers (including gbrain's own MCP → recursion + PGLite lock
+ *     contention), custom model defaults, and instruction overrides all stay
+ *     out of the subprocess. Auth state (auth.json) still loads.
+ *   - `--sandbox read-only` pins the agent sandbox down for defense in
+ *     depth; with the tool-use protocol the model answers in text and has
+ *     nothing to execute anyway.
+ *   - Codex has no `--system-prompt` flag, so system messages are rendered
+ *     as a leading `## System` section of the stdin prompt.
+ *
+ * Output channel: `-o <file>` writes the agent's final message verbatim;
+ * stdout carries progress logs and is discarded. Token usage is not exposed
+ * on this channel, so usage fields are undefined (the budget ledger treats
+ * subscription-billed calls as nominal anyway — see the recipe comment).
+ *
+ * doStream is not yet implemented; the model declares no streaming. Callers
+ * (gateway.toolLoop primarily) use doGenerate.
+ */
+import { spawn } from 'node:child_process';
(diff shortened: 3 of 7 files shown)


kweiner:
Tested this against a real self-hosted install (Docker Compose + Postgres, `codex` installed via the Ubuntu snap, ChatGPT Plus subscription). Test suite passes clean (18/18), and I confirmed a full live round trip works: `doGenerate` against `gpt-5.6-terra` returns a correctly-shaped `LanguageModelV2` result billed to the subscription with no `OPENAI_API_KEY` present.

One real bug found along the way, worth fixing before merge:

**`CODEX_CWD` (line 58) breaks on snap-installed `codex`.**

```ts
const CODEX_CWD = join(tmpdir(), `gbrain-codex-cli-cwd-${process.pid}`);
```

`os.tmpdir()` resolves to `/tmp` on Linux, and snap's confinement won't let a snap-packaged binary access arbitrary paths under `/tmp` — only specific allowed locations (home dir, the snap's own data dirs, etc.). Every invocation fails with a fairly opaque:

```
ERROR: {"detail": ...}   // when it gets further
codex-cli exited 1: Error: No such file or directory (os error 2)
```

Confirmed the root cause by pointing `TMPDIR` at a directory under `$HOME` instead — the exact same code path then works end-to-end. Repro:

```bash
# fails (default os.tmpdir() under /tmp) on snap-installed codex
bun live-test.ts

# works
TMPDIR=/home/ken/some-dir bun live-test.ts
```

Snap is a common install path for `codex` on Ubuntu/Debian hosts (it's what `snap install codex` gives you, separate from the npm package), so this will hit other self-hosters running gbrain outside a Mac/npm-global setup. Suggest either:
- Default `CODEX_CWD` under `os.homedir()` instead of `os.tmpdir()`, or
- Fall back to a home-based dir when the tmpdir spawn fails, or
- At minimum, document the `TMPDIR` workaround for snap installs.

Happy to send a small follow-up patch for this if useful — wanted to flag it here first since it's a one-line root cause but affects anyone testing this on a fresh snap install the way I did.

Separately: also had to `sudo snap refresh codex` to 0.146.0 — the `stable` channel snap (0.114.0) predates both `--ignore-user-config` and the `gpt-5.6-terra`/`gpt-5.6-sol` model ids (`gpt-5.3-codex`, the older default, isn't available under ChatGPT-account auth at all — `{"detail":"The 'gpt-5.3-codex' model is not supported when using Codex with a ChatGPT account."}`). Not a PR bug, just worth knowing if anyone else hits confusing errors on an older CLI.

One more, from deploying this for real in Docker (`oven/bun:1` base image): that image ships no CA trust store at all. Bun's own `fetch` doesn't care (bundles its own root CA list), but the `codex` binary is a separate Rust process that relies on the OS-level store, and fails with `invalid peer certificate: UnknownIssuer` against every `chatgpt.com`/`api.openai.com` call without one. Not this PR's bug either (base-image gap, not gbrain's), but worth a line in the setup docs since minimal Docker images are a common way people self-host gbrain — `apt-get install -y ca-certificates` (or the Alpine/musl equivalent) fixes it. Full round trip through the actual Minions job queue works cleanly once that's in place: submitted a real `subagent` job via `gbrain jobs submit subagent --follow`, executed inline through `codex-cli:gpt-5.6-terra`, returned in 4.1s.

runsmlee:
@kweiner — thank you, this is exactly the kind of report that's hard to get without someone running it on real hardware. Docker Compose + Postgres + snap + a live subscription is four environment variables away from my setup, and all three of your findings were things I could not have hit locally.

Fixed in 9152437.

### The snap bug

Your diagnosis is right, and the mechanism is slightly worse than "snap can't read `/tmp`": that directory is handed to the child as **both** `-C` and the parent of the `-o` output file, so a confined `codex` needs read *and* write access to it. Hence the `os error 2` — and it fires before anything useful reaches stderr.

I went with your option 3, plus teeth. Rather than move the default, the failure now explains itself:

```
codex-cli exited 1: Error: No such file or directory (os error 2)
--- hint ---
codex ran with cwd /tmp/gbrain-codex-cli-cwd-4831. Sandbox-confined installs
(e.g. the Ubuntu snap) cannot open arbitrary paths under /tmp; set TMPDIR to a
non-hidden directory inside $HOME and retry.
```

Reasoning for not defaulting to `$HOME` (option 1), since you offered a patch and I'd rather not have you write one I'd argue against:

- `CODEX_CWD` is per-pid and has **no cleanup path** — only `outFile` is removed. Under `/tmp` the OS reclaims it; under `$HOME` it would leave one empty directory per process, forever, on every host. Moving the default means also making it a fixed path and cleaning it up — a bigger change than the one-line version looks.
- It spends home-directory clutter on every operator to save a minority of hosts a single env export.

One detail worth flagging, because it cost me a second look and would have bitten an `$HOME` patch: **snap's `home` interface excludes hidden paths.** The AppArmor rule is effectively `owner @{HOME}/[^.]*`, which is why snaps keep state in `~/snap/<name>/` instead of `~/.config`. So the "obvious" fix of `~/.gbrain/codex-cwd` would have reproduced your exact bug — and it's also why your `TMPDIR=/home/ken/some-dir` repro worked while a dotdir wouldn't have.

If you'd still rather it work with **zero configuration** on snap, I'm happy to switch — fixed path under `$HOME` plus removal on exit, not per-pid. Your call as the person who actually has the failing host; say the word and I'll push it.

### Your other two findings

Both folded into a new `docs/ai-providers/codex-cli.md`, since neither is a code bug but both will hit the next self-hoster:

- **Minimum CLI 0.145.0**, with your `gpt-5.3-codex`-under-ChatGPT-auth error documented as the symptom of an older snap. That error message is genuinely confusing on first contact and now has a page it can be found from.
- **`ca-certificates` on minimal Docker bases.** Good catch on the root cause — Bun's `fetch` bundling its own roots is exactly what makes this one look like a gbrain bug instead of a base-image gap. Documented for both apt and apk.

### And you were right about the test count

18, not 20 — the description was wrong, not your run. Corrected, and it's 19 now: added one covering the hint (and asserting ordinary CLI failures stay clean, so the regex can't start swallowing unrelated errors). `bun test test/ai/` 472 pass / 0 fail, `bunx tsc --noEmit` clean.

Thanks again — particularly for confirming the full Minions job-queue round trip. That path I could test locally, but not on someone else's install.

time-attack:
Maintainer decision: **accepted in principle**, subject to review and to the conditions below. I had been leaning toward closing this as an off-mission vendor integration; that call has been overruled, and I think correctly — gbrain does ship a provider layer, and this is a provider recipe.

It still has to clear the same review bar as everything else (of 32 PRs reviewed adversarially, 1 was mergeable), plus two requirements specific to what this PR does:

**1. Fully opt-in, and inert when the binary is absent.** The recipe must have zero effect on any install that does not have the CLI present and has not explicitly configured it. Specifically: no change to provider resolution order, no new probe on a startup or doctor path that shells out to a binary that may not exist, and no warning emitted to users who never asked for this. A user who has never heard of this vendor should not be able to tell the recipe exists. I will verify by running the full CLI surface on a machine without the binary installed.

**2. Terms of service.** This routes chat through a consumer subscription login rather than an API key. Please confirm in the PR body that programmatic use via that OAuth flow is permitted by the vendor's current terms. I am not asking you to be a lawyer — just to state which terms you checked and when. If it is genuinely ambiguous, say so and we will decide with that on the table rather than discovering it later. gbrain cannot ship an integration whose supported path is one the vendor prohibits.

Beyond those, the ordinary bar applies, and two of these are where most PRs in this batch failed:

- **Its tests must fail without the fix.** Stub the old behavior back in and paste the result. 11 of the last 32 PRs had tests that passed with the production change reverted — one had 7 of 8 doing so. If the test only asserts the recipe file's contents, it proves nothing about routing.
- **No new `GBRAIN_*` env knob undocumented.** If this adds one, it goes in `CLAUDE.md` or `docs/`.
- **PR title must lead with the version** (`vMAJOR.MINOR.PATCH.MICRO`). 13 of 32 failed this; see #3698.
- **`VERSION` strictly greater than master's** — equal fails the CI version-gate outright, and several open PRs are in that state without knowing it.
- **No secret or token material in logs or error paths**, which matters more than usual for an OAuth-based flow.

1,169 lines is large for a recipe. If any of it is refactoring the provider layer rather than adding this recipe, splitting that out would get the recipe landed considerably faster — one concern per PR is reviewable, and I can verify each part.

Reviewing it properly next; no action needed from you until then unless you want to address the ToS point ahead of time.

time-attack:

Not a vendor plug: claude-cli (#334) is shipped precedent, gbrain ships a provider layer, and the maintainer already posted "accepted in principle" on the thread with explicit conditions. I checked out the branch and verified the implementation is sound: end-to-end wiring works (chat({model:'codex-cli:terra'}) resolves the alias to gpt-5.6-terra and round-trips text through a stub subprocess), typecheck clean, the PR's 19 tests + test/ai (472) + all registry-dependent suites (148) pass, tests bind to real behavior (removing the OPENAI_* env scrub fails 1 test; crippling tool-call extraction fails 4), all six codex flags plus -o and stdin-prompt verified against real codex-cli 0.146.0 on this machine, opt-in/inert-when-absent confirmed in code (doctor and init auto-pick both skip zero-required-env recipes; no import-time side effects; per-call prefix routing only), GBRAIN_CODEX_CLI_BIN documented in docs/ai-providers/codex-cli.md, documented config keys (chat_model, chat_fallback_chain) exist in the settable-key list, sibling-caller sweep clean (all 3 implementation switches covered; no other implementation=== sites apply), pricing follows the existing recipe-touchpoint convention (canonical table untouched, drift guard doesn't cover recipes — claude-cli identical). Three of the maintainer's own stated conditions remain unmet: (1) no ToS statement in the PR body, (2) the gateway routing hunk is unpinned by any test — reverting src/core/ai/gateway.ts to master leaves all 19 tests green, (3) title lacks the leading version and branch VERSION (0.42.67.0) is behind master (0.42.68.1). All three are small, concrete author fixes; merge after they land.

### Discrimination test

Three probes, branch restored between each: (1) reverted the entire gateway.ts hunk to origin/master — the PR's 19 tests STILL PASS (routing unpinned, blocker #2), while my external wiring probe (configureGateway + chat with codex-cli:terra + stub binary) fails with 'AIConfigError: Unknown implementation: codex-cli' and passes on the branch, proving the hunk is load-bearing; (2) deleted the OPENAI_API_KEY/OPENAI_BASE_URL scrub from the provider — 18 pass / 1 fail (env-scrub test discriminates); (3) crippled extractToolCalls to never find the <use_tools> block — 15 pass / 4 fail (tool-call tests discriminate). Tests import the real src module and spawn a real stub subprocess — not a re-declared copy.

**Mergeability at review time:** mergeable: MERGEABLE, mergeStateStatus: UNSTABLE (no CI checks reported on the fork branch — external contributor, workflows not run)

### Blockers

**1.** Maintainer condition 2 unmet: PR body contains no ToS statement — author must state which OpenAI/ChatGPT terms were checked and when for programmatic use via the Codex CLI OAuth flow (grep of the PR body for terms/tos/policy: zero hits)

**2.** Gateway routing unpinned: with src/core/ai/gateway.ts reverted to master (old behavior stubbed back in), all 19 new tests still pass — the maintainer explicitly required tests that fail without the fix. Add one gateway-level test: configureGateway({env:{}}) + chat({model:'codex-cli:terra', messages:[...]}) with GBRAIN_CODEX_CLI_BIN pointed at a stub, asserting text round-trip and model === 'codex-cli:gpt-5.6-terra' (this exact probe fails on reverted gateway with AIConfigError: Unknown implementation: codex-cli, passes on the branch)

**3.** Title/version conditions the maintainer restated for this PR: title is 'feat(ai): codex-cli recipe — ...' (no leading vMAJOR.MINOR.PATCH.MICRO) and branch VERSION is 0.42.67.0 vs master 0.42.68.1 (observed practice is that community PRs land without bumps and releases bump separately, but the maintainer put both in writing on this PR — title fix is one gh pr edit; version handling is the maintainer's call at merge)

### Not verified

- Live subscription round trip and validity of gpt-5.6-terra / gpt-5.6-sol model ids — local codex is API-key-authed, a live call would bill the user's key and not exercise the OAuth path; kweiner independently confirmed a live round trip on a real ChatGPT Plus + snap install per the PR thread
- Whether OpenAI's current consumer terms actually permit this routing — I did not fetch vendor terms; the maintainer's condition only requires the author to document what they checked (they haven't — blocker #1)
- Snap-confinement failure mode and the TMPDIR workaround (kweiner-reported, author-addressed in 9152437; macOS host here, not reproducible)
- Full CLI-surface run on a binary-less machine (maintainer said they will verify this themselves; approximated by code inspection + the fact that no test or probe shelled out to the real binary)

- `gh pr view 3587 --json ... (body, files, mergeable, reviews, comments)`
- `git fetch origin master pull/3587/head:pr-3587 && git checkout pr-3587 (merge-base 913d2d7f, 8 commits behind master)`
- `bun install (worktree)`
- `bun run typecheck > /tmp/pr3587_tsc.txt (EXIT=0)`
- `bun test test/codex-cli-recipe.test.ts (19 pass / 0 fail)`
- `bun run ./pr3587_wiring.ts — end-to-end chat() probe with stub codex binary (WIRING OK: alias terra→gpt-5.6-terra, text round-trip, usage zeros not NaN)`
- `mutation probe 1: removed OPENAI_* env scrub → 18 pass / 1 fail`
- `mutation probe 2: crippled extractToolCalls → 15 pass / 4 fail`
- `mutation probe 3: git checkout origin/master -- src/core/ai/gateway.ts → 19/19 tests still pass; wiring probe fails with 'Unknown implementation: codex-cli'`
- `bun test test/ai/ (472 pass / 0 fail — matches PR body claim)`
- `bun test test/providers.test.ts test/init-provider-picker.test.ts test/integrations.test.ts test/claude-cli-recipe.test.ts (148 pass / 0 fail)`
- `bun run check:privacy, check:gateway-routed, check:doc-history, check:test-names, check:test-isolation, check:fixture-privacy (all EXIT=0)`
- `codex --version (0.146.0) + codex exec --help — verified --ignore-user-config, --sandbox read-only, --skip-git-repo-check, -C, -m, -o/--output-last-message, stdin prompt all exist`
- `codex login status (API-key auth locally — skipped live smoke to avoid billing the user's key; wouldn't test the OAuth path anyway)`
- `grep sweeps: claude-cli sibling call sites (7 files, all mirrored), implementation=== special cases (none apply), chat_model/chat_fallback_chain in config settable-key list, doctor/init zero-required-env skip logic, model-pricing drift-guard scope`

</details>

---

time-attack:
Closing. This is net-new provider surface whose only value over the existing `openai` recipe is brokering a ChatGPT subscription's OAuth session through a `codex exec -` subprocess, and that's the part we won't take on: OpenAI's subscription terms bind that session to their own interfaces, so proxying it into a third-party memory daemon is a vendor-relationship risk gbrain shouldn't carry. An API key covers the functional need today.

Two secondary problems even setting the mission aside: the recipe hardcodes `cost_per_1m_input_usd` for model ids that appear nowhere in `CANONICAL_PRICING` (`src/core/model-pricing.ts`), so the ledger bills a hand-copied number the drift guard in `test/model-pricing.test.ts` can't see — that's exactly the duplication the one-canonical-table invariant exists to prevent; and usage tokens come back `undefined`, so every call on this path is invisible to the budget ledger. The dependency on an external binary's private auth state and undocumented flags is fragile in a way we'd be signing up to maintain.

The craft is fine — it follows the `claude-cli` recipe shape properly. It's the mission fit that fails. Thanks for the work.

PR #3588, opened 2026-07-29
fix(features): RECIPE_META secret names never matched the recipes, so 4 recipes read as permanently unconfigured

rayers:
The `features.ts` bug you flagged on #3505, with two corrections to the framing and a slightly wider blast radius than either of us thought.

**First correction, so credit lands on the right bug:** `GOOGLE_CONTACTS_API_KEY` was *my* invention in #3505 and I removed it in the trim, so master never carried it. Nothing to fix there.

**Second: it is 4 of 7 entries, not just calendar.** Every one of these named an env var that appears nowhere else in the repo:

| Recipe | RECIPE_META said | Recipe frontmatter actually says |
|---|---|---|
| `calendar-to-brain` | `GOOGLE_CALENDAR_API_KEY` | `CLAWVISOR_AGENT_TOKEN` / `GOOGLE_CLIENT_ID` |
| `email-to-brain` | `GMAIL_APP_PASSWORD` | `CLAWVISOR_AGENT_TOKEN` / `GOOGLE_CLIENT_ID` |
| `meeting-sync` | `CIRCLEBACK_API_KEY` | `CIRCLEBACK_TOKEN` |
| `credential-gateway` | `OAUTH_CLIENT_SECRET` | `CLAWVISOR_AGENT_TOKEN` / `GOOGLE_CLIENT_ID` |

So `gbrain features` has been nagging "4 integration recipes available but not configured" at users who configured all four correctly.

## Why it is not a one-line fix

Renaming alone does not fix it. The gate is `!r.secrets.every(s => process.env[s])`, and three of these recipes authenticate through an `any_of` health check (ClawVisor **or** direct OAuth). No set of names can satisfy `every`: a ClawVisor user never sets `GOOGLE_CLIENT_ID`, and an OAuth user never sets `CLAWVISOR_AGENT_TOKEN`. Whichever names you pick, one path stays permanently unconfigured.

So the predicate moves to `some`: configured once one path is present. That matches what the recipes document via `any_of`. The tradeoff is a half-configured recipe (client ID set, secret missing) now reads as configured, which I think is clearly better than the current permanent false positive, but say the word if you would rather I model `any_of` groups properly instead of flattening to `some`.

## Guard

`test/features-recipe-secrets.test.ts` asserts every `RECIPE_META` secret name appears in its recipe file, every id has a recipe on disk, and pins the any-of semantics with the exact ClawVisor-only case that was misreported. I verified it fails cleanly when `GOOGLE_CALENDAR_API_KEY` is restored, rather than only passing on the fixed tree.

This required exporting `RECIPE_META`. Happy to keep it private and assert through `runFeatures` output instead if you prefer a narrower surface, though that would test the recommendation string rather than the data.

`bun run typecheck` clean, 16 pass / 0 fail across the new file plus `test/features.test.ts`. Branched off current master.

Note the prior `test/features.test.ts` "covers all 7 recipes" case only asserts `runFeatures` is defined, which is why none of this was caught. I left it alone rather than widen this PR.


time-attack:

The bug is real and I reproduced it end-to-end: with all 10 env vars that recipes/*.md actually document set, master's `gbrain features --json` still nags about exactly email-to-brain, calendar-to-brain, meeting-sync and credential-gateway; on the PR branch it nags about none. Each single auth path (ClawVisor token only / GOOGLE_CLIENT_ID only / CIRCLEBACK_TOKEN) also correctly clears the right recipes, and nothing but recipe display names reaches the output — a sentinel secret value never appeared in any of 7 executed cases. typecheck clean. What fails review is the guard, not the fix: the risky half of the change (`every` -> `some`) is completely unpinned. I restored `every` on the PR branch and all 16 tests still passed, because the test file's `configured-detection semantics` block declares its own local `isUnconfigured` helper instead of exercising features.ts. The name half IS pinned (reverting one name fails 3 tests), but the substring match is weak enough that I made it green with a knowingly-wrong mapping. Separately, the PR's own root-cause argument applies verbatim to `getStatus` in integrations.ts, which I left untouched and confirmed still reports the same three recipes as AVAILABLE for a ClawVisor-only user — so after this merge `gbrain features` and `gbrain integrations list` disagree about the same install.

### Discrimination test

EXECUTED, three ways. (A) Behavioural half: reverted src/commands/features.ts:158 to master's `!r.secrets.every(s => process.env[s])` while keeping the PR's corrected secret names, ran `bun test test/features-recipe-secrets.test.ts test/features.test.ts` -> EXIT=0, 16 pass / 0 fail. The predicate change is provably untested. (B) Name half: reverted only the calendar-to-brain entry to `secrets: ['GOOGLE_CALENDAR_API_KEY']` -> EXIT=1, 4 pass / 3 fail (drift guard plus both semantics cases). That half genuinely discriminates. (C) Whole-file master stub: `git checkout origin/master -- src/commands/features.ts` -> the test file cannot even load ("Export named 'RECIPE_META' not found"), a degenerate result, which is why I ran the two halves separately. Guard-strength probe on top: RECIPE_META ngrok-tunnel entry set to the wrong-but-mentioned `TWILIO_AUTH_TOKEN` -> 7 pass / 0 fail.

### Blockers

**1.** src/commands/features.ts:158 — the behavioural half of the fix is not tested. test/features-recipe-secrets.test.ts:60-61 defines its own `const isUnconfigured = (secrets, env) => !secrets.some(s => env[s])` and asserts against that local copy, never against features.ts. Executed: I changed line 158 back to `!r.secrets.every(...)` (master's predicate, PR's names) and ran `bun test test/features-recipe-secrets.test.ts test/features.test.ts` -> EXIT=0, 16 pass / 0 fail. The predicate flip is the part that can regress silently and the part with the disclosed false-positive tradeoff; it needs a test that imports the real code path (export `scanFeatures`, or drive `runFeatures(engine, ['--json'])` with a stub engine and assert the `no-integrations` recommendation is absent when only CLAWVISOR_AGENT_TOKEN is set).

**2.** test/features-recipe-secrets.test.ts:34 — the drift guard is `body.includes(secret)` over the whole recipe markdown, not a check against the frontmatter `secrets:` list, so any env name mentioned anywhere in the file passes. Executed counterexample: set RECIPE_META's ngrok-tunnel entry to `secrets: ['TWILIO_AUTH_TOKEN']` (that name appears only in ngrok-tunnel.md's prose at line 144, never in its frontmatter) -> `bun test test/features-recipe-secrets.test.ts` EXIT=0, 7 pass / 0 fail. The guard catches this specific drift instance but would not catch the same class next time. Parse the `secrets:\n  - name: X` block and assert membership.

**3.** src/commands/integrations.ts:475-478 (`getStatus` -> `checkSecrets`, 'All required secrets must be set to be "configured"') — the sibling caller has the identical all-of bug the PR body diagnoses, and the PR now makes the two surfaces contradict each other. Executed via real `runIntegrations(['list'])` with CLAWVISOR_URL + CLAWVISOR_AGENT_TOKEN + CIRCLEBACK_TOKEN + X_BEARER_TOKEN + TWILIO_AUTH_TOKEN + NGROK_AUTHTOKEN set and GOOGLE_CLIENT_ID/SECRET unset (Option A, the recipe's own recommended path): credential-gateway = AVAILABLE, calendar-to-brain = AVAILABLE, email-to-brain = AVAILABLE. Same run with GOOGLE_CLIENT_ID/SECRET added: all three flip to CONFIGURED. So the ClawVisor user is now told 'configured' by `gbrain features` and 'AVAILABLE' (i.e. not configured) by `gbrain integrations list`. integrations.ts already parses the `any_of` health-check DSL (line 282), so it has the information to honour it in `getStatus`.

### Not verified

- Full unit suite / `bun run verify` / `bun run ci:local` — deliberately skipped per the execution-footprint rule. I ran only test/features-recipe-secrets.test.ts and test/features.test.ts (16 pass) plus `bun run typecheck` (EXIT=0).
- No Postgres container was started; nothing in the diff touches SQL or an engine, so real Postgres could not have changed any conclusion.
- Whether the disclosed `some` tradeoff (GOOGLE_CLIENT_ID set but GOOGLE_CLIENT_SECRET missing now reads as configured) causes a downstream failure at recipe-run time — I confirmed the reporting behaviour but did not execute a half-configured OAuth recipe. Note that if RECIPE_META ever gains a genuinely all-of recipe (e.g. TWILIO_ACCOUNT_SID + TWILIO_AUTH_TOKEN), `some` under-reports silently and the new drift guard would not notice; the shape has no per-entry any-of/all-of marker.
- recipes/agent-voice.md, restart-sweep.md and retrieval-reflex.md are absent from RECIPE_META (7 of 10 recipe files). Pre-existing on master and unchanged by this PR; I did not investigate whether that omission is intentional.
- I symlinked node_modules from the main workspace into the worktree to run anything at all, then removed the symlink. Dependency resolution therefore matched the main checkout rather than a fresh `bun install`.

- `gh pr view 3588`
- `gh pr diff 3588`
- `git fetch origin master && gh pr checkout 3588`
- `ls recipes/ ; grep -rn 'GOOGLE_CALENDAR_API_KEY|GMAIL_APP_PASSWORD|CIRCLEBACK_API_KEY|OAUTH_CLIENT_SECRET' . (only the new test file mentions them)`
- `grep -rn 'CLAWVISOR_AGENT_TOKEN|GOOGLE_CLIENT_ID|GOOGLE_CLIENT_SECRET|CIRCLEBACK|any_of' recipes/`
- `grep -rn 'RECIPE_META' src test`
- `grep -rln 'health_checks|any_of' src`
- `ln -s <main-workspace>/node_modules node_modules   # later rm'd`
- `bun test test/features-recipe-secrets.test.ts test/features.test.ts   # EXIT=0, 16 pass`
- `# discrimination A: line 158 some -> every, then: bun test test/features-recipe-secrets.test.ts test/features.test.ts   # EXIT=0, 16 pass (STILL GREEN)`
- `# discrimination B: calendar entry -> ['GOOGLE_CALENDAR_API_KEY'], then: bun test test/features-recipe-secrets.test.ts   # EXIT=1, 4 pass / 3 fail`
- `# discrimination C: git checkout origin/master -- src/commands/features.ts, then bun test ...   # EXIT=1, module-load error (no export)`
- `# guard-strength probe: ngrok-tunnel entry -> ['TWILIO_AUTH_TOKEN'], then bun test test/features-recipe-secrets.test.ts   # EXIT=0, 7 pass`
- `bun run harness-3588.ts   # real runFeatures(--json) over 7 env permutations, PR branch: all 4 recipes clear, sentinel never logged`
- `bun run harness-3588.ts   # same harness against a master-behaviour stub (old names + every): nags Email/Calendar/Meeting Sync/Credential Gateway with every documented env var set`
- `bun run harness-3588b.ts   # real runIntegrations(['list']) with all secrets set -> CONFIGURED; ClawVisor-only -> credential-gateway/calendar-to-brain/email-to-brain AVAILABLE`
- `bun run typecheck   # EXIT=0`
- `rm -f harness-3588.ts harness-3588b.ts node_modules ; git status --short   # clean, worktree at PR HEAD`

</details>

---

rayers:
Pushed 190cef5f addressing all three blockers. Each fix carries a discrimination check.

**Blocker 1 — predicate untested.** Exported `scanFeatures` and rewrote the `configured-detection semantics` block to drive the real code path with a stub engine, asserting on the `no-integrations` recommendation. The old block asserted against a local `isUnconfigured` copy. Discrimination: reverting `features.ts:158` from `some` back to `every` now fails 2 tests (`bun test test/features-recipe-secrets.test.ts` -> 5 pass / 2 fail).

**Blocker 2 — weak drift guard.** Replaced `body.includes(secret)` with a parser that reads the frontmatter `secrets:` block and asserts membership. Discrimination: pointing a RECIPE_META entry at `NGROK_DOMAIN` (present 5x in ngrok-tunnel.md prose, absent from its `secrets:` frontmatter) passed the old guard and fails the new one.

**Blocker 3 — sibling surface contradiction.** `getStatus` in `integrations.ts` now honors the recipe's own `any_of` health check instead of requiring every flat secret. Each `any_of` group needs one branch satisfied by env (env_exists -> var present; http -> its `$VAR` refs present, a sync proxy for the network check). Recipes with no `any_of` (x-to-brain, twilio, meeting-sync, ngrok) keep the exact all-secrets rule, so their behavior is unchanged. Now a ClawVisor-only user reads `configured` on both `gbrain features` and `gbrain integrations list`. Pinned by the new `test/integrations-getstatus-anyof.test.ts`; reverting getStatus fails the two any-of cases while the non-any-of cases stay green.

typecheck clean. Affected suites: 151 pass / 0 fail across test/integrations*.test.ts and test/features*.test.ts.

time-attack:

Genuine bug fix still needed on master: src/commands/features.ts (origin/master ~line 42-50) still declares GMAIL_APP_PASSWORD, GOOGLE_CALENDAR_API_KEY, CIRCLEBACK_API_KEY, OAUTH_CLIENT_SECRET — none appear in the recipes' frontmatter (calendar/email/credential-gateway declare CLAWVISOR_*/GOOGLE_CLIENT_*; meeting-sync declares CIRCLEBACK_TOKEN) — and the `!r.secrets.every(...)` gate at features.ts:150 plus getStatus's all-of rule at integrations.ts:499 make any_of-auth recipes permanently 'unconfigured'. Verified in a /tmp worktree: PR tests fail without the fix, and 141 pass after a two-word rename. But it cannot merge as-is (see conflict_assessment).

**Conflict state vs current master:** One textual conflict + one carried-over test bug + one semantic drift, all from post-base master commits (not the wave itself). (1) features.ts RECIPE_META x-to-brain line: master's #3649 (addf0311) renamed X_BEARER_TOKEN → X_API_BEARER_TOKEN; trivial resolution keeping master's name. (2) Both new test files hardcode the stale X_BEARER_TOKEN — verified 2 test failures post-merge even after resolving (1); sed rename to X_API_BEARER_TOKEN → 141/141 pass. (3) Semantic: #3648 (7376c026) made integrations.ts resolve secrets through secretEnv() (config.json→env folding; checkSecrets at :483 and the env_exists runner at :272 both use it), but the PR's new branchSatisfiedByEnv/checkEnvRefs read raw process.env — so a ClawVisor token stored only in ~/.gbrain/config.json would show [set] in `integrations show` yet 'available' in getStatus, reintroducing the #2789/#3648 inconsistency on the exact recipes this PR fixes. Two-line fix: use secretEnv() in branchSatisfiedByEnv. The v0.42.69.0 wave itself did not touch these predicates — fix still fully applies.

**Risk if merged as-is:** If merged after only resolving the textual conflict, the PR's own two X-token tests fail and turn master red (verified in worktree). The `some`/any_of relaxation means a half-configured recipe (e.g. CLAWVISOR_URL set, token missing) reads as configured — acknowledged tradeoff in the PR body, false-negative direction, non-breaking. Exporting RECIPE_META/scanFeatures/getStatus widens surface slightly; harmless.

rayers:
Rebased onto master (v0.42.71.0) and addressed all three items from your conflict assessment.

1. **Textual conflict (features.ts RECIPE_META x-to-brain).** Merged master, kept its `X_API_BEARER_TOKEN` rename (#3649) alongside this PR's corrected any_of names for email/calendar/credential-gateway.
2. **Carried-over test bug.** Both new test files renamed `X_BEARER_TOKEN` -> `X_API_BEARER_TOKEN`.
3. **Semantic drift vs #3648.** `branchSatisfiedByEnv`/`authConfigured` now resolve through `secretEnv()` instead of raw `process.env`, so a secret stored only in `~/.gbrain/config.json` counts. This is the same source `checkSecrets` (:483) and the env_exists health-check runner (:312) read, so `gbrain integrations show` and `getStatus` no longer disagree on a config-stored ClawVisor token.

Post-merge state: no conflict markers, version trio agrees at 0.42.71.0, `bun run verify` green (34/34), and 172 pass / 0 fail across the features + integrations suites. Discrimination pins survive the merge: reverting the `some` predicate fails 2 features tests, and forcing getStatus back to all-of fails 2 getStatus tests.

time-attack:
Status update: this PR was reviewed in the 2026-08-18 full-backlog triage and confirmed valuable — it sits in the verified merge queue (position 14 of 53, risk: low). CLEAN green. Pair with 3900 (both edit commands/features.ts) — merge first. No action needed from you right now — thank you for the contribution and the patience.

time-attack:

Issue #2606, opened 2026-07-05
Chronicle judge: output truncated at maxTokens=1500 is silently recorded as no_events (parse failure indistinguishable from empty page)


jerome-marshall-ntx:
**Version:** 0.42.56.0 (also current master `6461790`) · **Engine:** Postgres 16 + pgvector · **Runtime:** Bun · **Chat:** OpenAI-compatible endpoint

## What happens

The chronicle judge caps its completion at `maxTokens: 1500` (`src/core/chronicle/extract-events.ts:186`) and then treats any unparseable output as "no events":

- `parseJudgeJson` returns `[]` on any parse failure — missing `]`, truncated JSON, anything (`extract-events.ts:200-214`);
- `runChronicleExtract` maps an empty proposals array to `status: 'no_events'` (`extract-events.ts:129-130`);
- the only `stopReason` values handled are `refusal` / `content_filter` (`extract-events.ts:188`) — a `max_tokens` / length stop is not detected.

So when a page is event-dense enough that the JSON array exceeds ~1500 output tokens, the response is cut off mid-array, `lastIndexOf(']')`/`JSON.parse` fails, and the page is recorded as legitimately having no events. Nothing in logs, job output, or the result distinguishes this from a genuinely empty page.

## Evidence

Running chronicle extraction across ~61 daily/meeting-shaped pages: 49 extracted fine, but 11 pages persistently returned `no_events` across retries — and they were consistently the larger ones (3-4K chars of dense bullet lists, i.e. many events), while smaller pages extracted normally. That pattern matches the JSON array being truncated at the token cap and silently swallowed, rather than the model finding nothing.

## What should happen

Any of (ideally all):

1. Detect truncation: when `stopReason` indicates the token cap was hit, do not feed the text to `parseJudgeJson` as if complete — retry with a higher cap, or fail the page explicitly.
2. Distinguish `parse_failed` from `no_events` in `ChronicleExtractResult` (there is already a `reason` field and a `malformed_proposal` skip path — a parse failure after non-empty model text deserves the same visibility).
3. Make the judge `maxTokens` configurable (or scale it with input size), so event-dense pages have headroom.

## Where

- `src/core/chronicle/extract-events.ts:186` (`maxTokens: 1500`)
- `src/core/chronicle/extract-events.ts:188` (only refusal/content_filter checked)
- `src/core/chronicle/extract-events.ts:200-214` (`parseJudgeJson` → `[]` on failure)
[...]

PR #3595, opened 2026-07-29
fix(jobs): chronicle_extract must re-resolve models from the engine (#3387)

### LINKED ISSUE #3387: chronicle_extract silently returns no_events on PGLite (gateway not configured in inline worker + maxTokens too low for reasoning models)
**What happened?**

`chronicle_extract` jobs silently return `{ status: "no_events", events_written: 0 }` on PGLite. Two separate root causes were found:

1. **AI gateway never configured in the inline worker path.** The `--follow` inline worker in `src/commands/jobs.ts` (around line 373-386) creates a `MinionWorker` and registers handlers, but never calls `configureGateway(buildGatewayConfig(loadConfig()))`. The gateway singleton (`_config` in `src/core/ai/gateway.ts`) stays `null`, so `isAvailable('chat')` returns `false`. The chronicle judge (`src/core/chronicle/extract-events.ts:172-173`) checks `isAvailable('chat')` and returns `{ events: [] }` immediately without calling the LLM. The job completes in ~0.2s with no error.

2. **Chronicle judge maxTokens too low for reasoning models.** After fixing issue #1, the LLM IS called (16.4s completion) but returns an empty response with `STOP REASON: length`. The judge hardcodes `maxTokens: 1500` (`src/core/chronicle/extract-events.ts:186`). Reasoning models (e.g. DeepSeek `deepseek-v4-pro`, `deepseek-v4-flash`) consume all tokens on internal thinking before producing output. 1500 and 4000 both produce empty responses; 8000 produces correct output.

On the latest master, a `GATEWAY_REFRESH_JOB_NAMES` set was added (around line 36 of `src/commands/jobs.ts`) that wraps certain job handlers with `refreshGatewayForJob(engine)`. However, `chronicle_extract` is **not** in that set, so even the latest version has issue #1 for chronicle specifically. Issue #2 (`maxTokens: 1500`) is unchanged on master.

The Postgres daemon path (`gbrain jobs work`) goes through `connectEngine()` which calls `configureGateway`, so the gateway singleton is set in the daemon process. Issue #1 is specific to the PGLite inline `--follow` path. Issue #2 affects both PGLite and Postgres.

**What did you expect?**

`chronicle_extract` should call the configured chat model to segment the meeting page into timeline events. With a valid `chat_model` and API key configured, `isAvailable('chat')` should return `true` and the judge should invoke the LLM with enough tokens to produce a complete JSON array.

After local patching (gateway config fix + `maxTokens: 8000`), the chronicle judge successfully extracted 11 events from a Portuguese meeting note in 43 seconds. Events were correctly typed (meeting, decision, commitment) with correct attendees and dates.

**Steps to reproduce**

1. Initialize a PGLite brain with a non-Anthropic chat model configured:
   ```bash
   gbrain init --pglite --no-embedding
   gbrain config set chat_model deepseek:deepseek-v4-pro
   gbrain config set auto_chronicle true --force
   gbrain config set chronicle.tz America/Sao_Paulo --force
   gbrain apply-migrations --yes
   ```

2. Create a meeting page (chronicle-eligible: `type: meeting`, ≥80 chars body):
   ```bash
   printf '%s' "---
   date: 2026-07-09
   attendees:
     - people/test-person
   ---

   # Test Meeting

   ## Summary
   A meeting to discuss the test project. Decided to proceed with phase 1.
   " | gbrain put meetings/2026-07-09-test
   ```

3. Run chronicle-backfill (enqueues the job):
   ```bash
   gbrain chronicle-backfill
   # Output: { "scanned": 1, "eligible": 1, "enqueued": 1, ... }
   ```

4. Execute the queued job inline (PGLite only supports `--follow`):
   ```bash
   gbrain jobs submit chronicle_extract --params '{"slug":"meetings/2026-07-09-test","sourceId":"default"}' --follow
   ```

5. Observe the output (issue #1 — gateway not configured):
   ```
   Job #N completed in 0.2s
   Result: {"slug":"meetings/2026-07-09-test","status":"no_events","events_written":0}
   ```

6. After applying the gateway config fix, re-run step 4. Observe the output (issue #2 — maxTokens too low):
   ```
   Job #N completed in 16.4s
   Result: {"slug":"meetings/2026-07-09-test","status":"no_events","events_written":0}
   ```
   The 16.4s completion (vs 0.2s) confirms the LLM was called but returned empty due to `STOP REASON: length`.

7. After applying both fixes (gateway config + `maxTokens: 8000`), re-run step 4:
   ```
   Job #N completed in 43.0s
   Result: {"slug":"meetings/2026-07-09-test","status":"extracted","events_written":11}
   ```

**Environment**

- gbrain version: 0.42.57.0 (confirmed both issues still present on master as of v0.42.65.0)
- OS: Linux (Debian-based Docker container)
- Bun version: 1.3.14
- Database: PGLite (local embedded)

```json
[WARN] subagent_capability: chat_model is "deepseek:deepseek-v4-pro" (non-Anthropic) and ANTHROPIC_API_KEY is not set. Subagent features (gbrain dream, gbrain agent run, gbrain autopilot) will fail at job submission unless agent.use_gateway_loop=true. Chat alone (gbrain think) still works. Either set ANTHROPIC_API_KEY or enable: `gbrain config set agent.use_gateway_loop true`.
```

Note: `gbrain think --model deepseek:deepseek-v4-pro` works correctly (the gateway is configured via `connectEngine` in that path). Issue #1 is specific to the `jobs submit --follow` inline worker path.

Suggested fixes:
- Issue #1: call `configureGateway(buildGatewayConfig(loadConfig()))` in the `--follow` path before `registerBuiltinHandlers`, OR add `chronicle_extract` to `GATEWAY_REFRESH_JOB_NAMES` on master.
- Issue #2: increase `maxTokens` to 8000 (or make it configurable via `gbrain config set chronicle.max_tokens`). Non-reasoning models would simply not use the extra tokens; reasoning models need them.

time-attack:
Fixes #3387 (part 1).

## The bug

`registerBuiltinJob` (`src/commands/jobs.ts`) wraps a handler with `refreshGatewayForJob(engine)` **only** when its name is in `GATEWAY_REFRESH_JOB_NAMES`:

```ts
if (!GATEWAY_REFRESH_JOB_NAMES.has(name)) { worker.register(name, handler); return; }
worker.register(name, async (job) => { await refreshGatewayForJob(engine); return await handler(job); });
```

That refresh calls `reconfigureGatewayWithEngine` (`src/core/ai/gateway.ts:532`), which resolves `models.chat` **from the engine** — `resolveModel(engine, { configKey: 'models.chat', tier: 'reasoning', fallback: cfg.chat_model … })` — falling back to the file/env plane.

`chronicle_extract` was absent from the set. So it saw only connect-time file/env config, a model set with `gbrain config set` never reached the gateway, and `extract-events.ts:196`'s `if (!isAvailable('chat')) return { events: [] }` returned a silent `no_events`.

## Why it survived triage

Two independent reviews reached opposite verdicts. One **ran the reporter's repro live** — real key, `jobs submit chronicle_extract --follow` — got `{"status":"extracted","events_written":1}`, and concluded not-a-bug. The other traced it statically and said it was real.

Both were correct about different config planes: the live run's model came from an environment variable, which the connect-time config already carries, so the DB-plane path was never exercised. A passing live test genuinely could not see this.

## Tests

Pin set membership, because that **is** the mechanism, plus a floor listing every chat-calling job so the next one dropped from this hand-maintained list fails loudly rather than silently ignoring config.

**3 pass with the fix, 2 fail on master** (both membership assertions). `bun run typecheck` clean, `bun run verify" green.

## Scope

Part 1 only. Part 2 (the token cap) was largely fixed by #2606 — default raised 1500 → 4000, configurable via `chronicle.judge_max_tokens`, truncation surfaced instead of parsed as empty. Open PR #3478 adds split-and-retry but touches neither `jobs.ts` nor this set, so it does not fix part 1.

**The wider hazard**: this set is hand-maintained, so any job added without an entry silently ignores DB-plane model config. The maintainer chose the one-line fix over a full sweep for now; worth revisiting if another instance appears.

### DIFF
--- a/src/commands/jobs.ts
+++ b/src/commands/jobs.ts
@@ -47,6 +47,15 @@ const GATEWAY_REFRESH_JOB_NAMES = new Set([
   'embed-backfill',
   'extract-takes-from-pages',
   'embed-catch-up',
+  // #3387: chronicle_extract calls the chat model, so it must re-resolve
+  // models from the ENGINE (the DB config plane) before running. Without
+  // this entry the job sees only the connect-time file/env config, so a
+  // model set via `gbrain config set` is silently ignored and
+  // extract-events.ts's `if (!isAvailable('chat')) return { events: [] }`
+  // returns a silent `no_events`. The bug is invisible when the model comes
+  // from an env var — which is why a live repro can pass while the reported
+  // (DB-plane) configuration still fails.
+  'chronicle_extract',
 ]);
 
 function registerBuiltinJob(
--- a/test/chronicle-extract-gateway-refresh.test.ts
+++ b/test/chronicle-extract-gateway-refresh.test.ts
@@ -1,0 +1,75 @@
+/**
+ * #3387: `chronicle_extract` must re-resolve models from the ENGINE before it runs.
+ *
+ * `registerBuiltinJob` wraps a handler with `refreshGatewayForJob(engine)` only
+ * when its name is in `GATEWAY_REFRESH_JOB_NAMES`. That refresh calls
+ * `reconfigureGatewayWithEngine`, which resolves `models.chat` from the DB
+ * config plane (`resolveModel(engine, { configKey: 'models.chat', … })`),
+ * falling back to the file/env plane.
+ *
+ * Without the entry the job sees only the connect-time file/env config, so a
+ * chat model set with `gbrain config set` is silently ignored and
+ * `extract-events.ts` returns a silent `no_events`.
+ *
+ * WHY THE TEST IS SHAPED THIS WAY: the bug is invisible when the model comes
+ * from an environment variable, because the connect-time config already carries
+ * it. A reviewer ran the reporter's repro live, it passed, and they concluded
+ * not-a-bug for exactly that reason. Set membership IS the mechanism, so that
+ * is what this pins.
+ */
+import { describe, test, expect } from 'bun:test';
+import { readFileSync } from 'node:fs';
+import { resolve } from 'node:path';
+
+const JOBS_SRC = readFileSync(resolve(import.meta.dir, '../src/commands/jobs.ts'), 'utf-8');
+
+/**
+ * Members of the GATEWAY_REFRESH_JOB_NAMES set literal.
+ * Scans only quoted entries on their own line, so prose in comments (which may
+ * legitimately contain quoted identifiers) cannot be mistaken for a member.
+ */
+function refreshSetMembers(): string[] {
+  const start = JOBS_SRC.indexOf('const GATEWAY_REFRESH_JOB_NAMES = new Set([');
+  if (start === -1) throw new Error('GATEWAY_REFRESH_JOB_NAMES not found — declaration moved?');
+  const end = JOBS_SRC.indexOf(']);', start);
+  if (end === -1) throw new Error('GATEWAY_REFRESH_JOB_NAMES has no terminator');
+  return JOBS_SRC.slice(start, end)
+    .split('\n')
+    .map((line) => line.trim())
+    .filter((line) => !line.startsWith('//'))
+    .map((line) => /^'([^']+)',$/.exec(line))
+    .filter((m): m is RegExpExecArray => m !== null)
+    .map((m) => m[1]);
+}
+
+describe('#3387: chronicle_extract re-resolves models from the engine', () => {
+  test('chronicle_extract is in GATEWAY_REFRESH_JOB_NAMES', () => {
+    // Fails on master: 13 members, this is not one of them.
+    expect(refreshSetMembers()).toContain('chronicle_extract');
+  });
+
+  test('the refresh wrapper is what consults the DB config plane', () => {
+    // Pins the mechanism, so a refactor that drops the engine-aware refresh
+    // shows up here rather than as a mystery `no_events`.
+    expect(JOBS_SRC).toContain('GATEWAY_REFRESH_JOB_NAMES.has(name)');
+    expect(JOBS_SRC).toContain('refreshGatewayForJob(engine)');
+  });
+
+  test('every chat-calling job stays in the set', () => {
+    // Regression floor: each of these invokes a chat/expansion model and would
+    // silently ignore DB-plane model config if dropped from the set.
+    const members = refreshSetMembers();
+    const required = [
+      'chronicle_extract',
+      'extract_facts',
+      'extract-conversation-facts',
+      'synthesize',
+      'patterns',
+      'consolidate',
+      'extract-takes-from-pages',
+      'enrich',
+    ];
+    const missing = required.filter((n) => !members.includes(n));
+    expect(missing).toEqual([]);
+  });
+});


time-attack:

The fix is inert. `chronicle_extract` is registered at src/commands/jobs.ts:1611 via bare `worker.register(...)`, NOT via `registerBuiltinJob(...)` — and `registerBuiltinJob` is the only reader of `GATEWAY_REFRESH_JOB_NAMES` (grep confirms exactly one consumer at jobs.ts:67). All 13 pre-existing set members go through `registerBuiltinJob`; chronicle_extract is the sole exception. Adding the name to the set therefore changes nothing at runtime: the handler is still never wrapped with `refreshGatewayForJob(engine)`. I proved this with an executed runtime probe, and proved the real fix is a one-word change on line 1611.

The PR's own 3 tests all pass on the PR branch while the claimed bug remains fully unfixed, because every assertion is a source-text grep (`readFileSync` on jobs.ts + `toContain('chronicle_extract')` / `toContain('GATEWAY_REFRESH_JOB_NAMES.has(name)')`). The 2 that fail on master fail tautologically — they assert the presence of the string the diff adds. The "regression floor" test has the same defect: it can never catch a job that is in the set but not routed through the wrapper, which is precisely the bug in front of it.

Secondary: the shipped comment and PR body misstate the mechanism. `connectEngine()` already calls `reconfigureGatewayWithEngine(engine)` at src/cli.ts:2435 on every connect, so connect-time config ALREADY includes the DB plane. The real (narrower) gap is config changed after the worker process started — which is exactly what `refreshGatewayForJob`'s own docstring at jobs.ts:25-30 says, and not what the new comment claims. Shipping the wrong mechanism as an 8-line code comment misdirects the next debugger.

The prompt's specific checks came back clean: the set has exactly 14 members after the diff, and the member parser does NOT sweep quoted strings out of comments (it filters `//` lines and requires `^'([^']+)',$`, so the `('chat')` inside the new comment is correctly excluded — I verified `members` does not contain `chat`). Typecheck is clean. Those were the right things to check; they just are not where the defect is.

To make this mergeable: change jobs.ts:1611 to `registerBuiltinJob(worker, engine, 'chronicle_extract', async (job) => {`, replace the grep tests with the runtime probe (or equivalent), correct the comment to say "config changed after worker start", and retitle with a leading vMAJOR.MINOR.PATCH.MICRO.

### Discrimination test

EXECUTED, three ways.

(1) Behavioral probe on the PR branch (the test the PR should have written). test/zz-disc3595.test.ts: import `registerBuiltinHandlers` from src/commands/jobs.ts, pass a fake worker that captures handlers by name plus an engine Proxy whose every method sets `engineTouched = true`, then invoke the captured 'chronicle_extract' handler with `{data:{}}`. Result on the PR branch: `expect(msg).toContain('requires data.slug')` PASSES (we reached the inner handler, so nothing ran before it) and `expect(engineTouched).toBe(true)` FAILS — Expected true, Received false. EXIT=1. The PR's stated fix does not happen at runtime.

(2) Harness validity control. Same probe against 'enrich' (in the set, registered via registerBuiltinJob) → engineTouched === true, PASSES. So the probe genuinely detects the refresh wrapper; the chronicle_extract failure is the code, not the harness. (The second control, 'lint', reported true because the lint handler itself queries the engine rather than throwing early — not a harness flaw, since chronicle_extract provably throws before any engine call, which the 'requires data.slug' assertion pins.)

(3) Real fix confirms the diagnosis. Changed jobs.ts:1611 `worker.register('chronicle_extract', …)` → `registerBuiltinJob(worker, engine, 'chronicle_extract', …)` (nothing else). Probe → 1 pass / 0 fail, EXIT=0. Reverted; `git status --short` clean.

(4) Old behavior stubbed back into the PR's own tests. Deleted only the `'chronicle_extract',` line from the set, keeping the new test file: `bun test test/chronicle-extract-gateway-refresh.test.ts` → 1 pass / 2 fail, matching the PR's "2 fail on master" claim. But with the line restored → 3 pass / 0 fail while probe (1) still fails. The tests discriminate the presence of a string, not the presence of the fix. Restored via `git checkout src/commands/jobs.ts`.

### Blockers

**1.** src/commands/jobs.ts:1611 — INERT FIX. `chronicle_extract` is registered with bare `worker.register('chronicle_extract', …)`, bypassing `registerBuiltinJob` (the only reader of GATEWAY_REFRESH_JOB_NAMES, jobs.ts:67). Executed proof: registerBuiltinHandlers() with a recording engine Proxy, then invoke the registered 'chronicle_extract' handler with `{data:{}}` → it throws 'chronicle_extract job requires data.slug' (inner handler reached) with engineTouched === false (no gateway refresh ran). Identical on master and PR branch. Positive control: same probe on 'enrich' (in the set via registerBuiltinJob) → engineTouched === true, so the harness detects the wrapper. Applying the real fix (line 1611 → `registerBuiltinJob(worker, engine, 'chronicle_extract', …)`) flips the probe to pass, EXIT=0.

**2.** test/chronicle-extract-gateway-refresh.test.ts:46,54,63 — TESTS PROVE NOTHING. All three tests only readFileSync jobs.ts and assert substrings. `bun test test/chronicle-extract-gateway-refresh.test.ts` → 3 pass / 0 fail on the PR branch while the handler is provably still unwrapped. The 2 failing on master fail tautologically (they assert the string the diff adds). The 'every chat-calling job stays in the set' floor test is structurally incapable of catching the very defect present in its own subject.

**3.** src/commands/jobs.ts:50-57 — WRONG MECHANISM SHIPPED AS A COMMENT. Claims the job 'sees only the connect-time file/env config' and that a model set with `gbrain config set` is 'silently ignored'. But src/cli.ts:2435 calls `reconfigureGatewayWithEngine(engine)` inside `connectEngine()` on every connect, which resolves models.chat via `resolveModel(engine, …)` off the DB plane. So DB-plane config set before `jobs work` starts IS already picked up. The genuine gap is only config mutated mid-worker-lifetime — which is what jobs.ts:25-30 already documents. The PR body repeats the same wrong framing, including the two-reviewers-disagreed narrative built on it.

**4.** src/commands/jobs.ts:36 — SYMPTOM, NOT ROOT CAUSE, and siblings left broken. Bare-`worker.register` jobs that call chat/LLM models and are absent from the set: `facts-absorb` (1727), `extract-ner` (2147), `extract-timeline-from-meetings` (2174), `unify-types` (2208), `skillopt` (2237). Each carries the same mid-lifetime stale-model exposure. The PR body acknowledges the hand-maintained-list hazard and defers it, but the 'floor' test it offers as mitigation greps source rather than inspecting registration, so it cannot flag any of them.

**5.** PR title — 'fix(jobs): chronicle_extract must re-resolve models from the engine (#3387)' has no leading version token. CLAUDE.md IRON RULE requires `vMAJOR.MINOR.PATCH.MICRO <type>(<scope>): <summary>`. Fix with `gh pr edit 3595 --title "vX.Y.Z.W fix(jobs): …"`.

### Not verified

- Full unit suite, `bun run verify`, `bun run test`, `bun run ci:local` — deliberately skipped per the execution-footprint constraint. Ran only `bun run typecheck` plus the PR's own test file and my three scratch probes.
- Whether issue #3387's originally reported symptom is fully explained by the mid-lifetime staleness gap. Establishing that needs a real Postgres brain plus a live API key and a `gbrain config set` between worker start and job dispatch — no container spun up. Immaterial to the verdict: whatever the true scope, this diff does not change runtime behavior at all.
- Whether the 5 other bare-`worker.register` chat-calling jobs (facts-absorb, extract-ner, extract-timeline-from-meetings, unify-types, skillopt) each definitively reach a gateway chat call — inferred from their names and handler imports, not traced call-by-call. Directional finding, not a merge blocker for this PR.
- The PR body's cross-references to #2606 and #3478 (token-cap scope claims) — not independently checked; orthogonal to part 1.

- `gh pr view 3595`
- `gh pr diff 3595 > /tmp/pr3595.diff`
- `git fetch origin pull/3595/head:pr3595-local && git checkout pr3595-local`
- `grep -n 'GATEWAY_REFRESH_JOB_NAMES' -A 60 src/commands/jobs.ts`
- `grep -rn 'chronicle_extract' src/`
- `grep -rn 'GATEWAY_REFRESH_JOB_NAMES|refreshGatewayForJob' src/ test/`
- `grep -n "registerBuiltinJob(worker, engine, '" src/commands/jobs.ts   # 13 sites`
- `grep -n "worker.register('" src/commands/jobs.ts   # 29 sites, incl. chronicle_extract:1611`
- `grep -rn 'reconfigureGatewayWithEngine' src/ | grep -v 'ai/gateway.ts'   # cli.ts:2435, jobs.ts:33`
- `sed -n '515,560p' src/core/ai/gateway.ts`
- `sed -n '2415,2445p' src/cli.ts`
- `ln -s <main-workspace>/node_modules ./node_modules   # worktree had none; removed afterward`
- `bun test test/zz-disc3595.test.ts   # runtime probe: EXIT=1, engineTouched=false on PR branch`
- `bun test test/zz-control3595.test.ts   # harness control: 'enrich' -> engineTouched=true`
- `bun test test/zz-disc3595.test.ts   # after changing jobs.ts:1611 to registerBuiltinJob: EXIT=0, 1 pass`
- `bun test test/chronicle-extract-gateway-refresh.test.ts   # PR branch: EXIT=0, 3 pass`
- `bun test test/chronicle-extract-gateway-refresh.test.ts   # set entry removed: EXIT=1, 1 pass / 2 fail`
- `bun test test/zz-count3595.test.ts   # COUNT=14, members does not contain 'chat'`
- `bun run typecheck   # EXIT=0`
- `git checkout src/commands/jobs.ts && rm test/zz-*.test.ts node_modules && git status --short   # clean`

</details>

---

time-attack:
Closing this. It does not work, and I verified that against my own PR.

Adversarial review found it **inert**: `chronicle_extract` is registered with a bare `worker.register`, which bypasses the only reader of `GATEWAY_REFRESH_JOB_NAMES`. Adding the name to that set therefore changes nothing at runtime. The handler stays unwrapped.

Worse, all three tests I wrote passed anyway, because they `readFileSync` `src/commands/jobs.ts` and substring-match it. They assert that the source contains a string — not that the behavior happens. So the suite was green while the fix provably did nothing. That is the exact failure mode I have been rejecting other people's PRs for this week, and I filed #3665 about it before noticing I had done it here.

The underlying bug in #3387 is real and unfixed: `chronicle_extract` does not re-resolve models from the engine, so it ignores the DB config plane. Anyone picking it up should note two things this attempt got wrong:

1. **The wiring, not the list, is the fix.** Route the registration through whatever consumes `GATEWAY_REFRESH_JOB_NAMES`, or make the refresh unconditional at the handler boundary. Adding to the set alone is a no-op.
2. **Test the behavior.** Submit the job with a model configured on the DB plane only and assert the resolved model — do not assert on the contents of `jobs.ts`. The repo's own precedent for this is #3224 (spawn the CLI, assert exit code and output).

I am moving to review and triage only rather than authoring fixes, so I am not going to iterate on this. Closing rather than leaving a broken PR open to imply the issue is handled. #3387 stays open and accurately reflects that nothing is fixed.

PR #3611, opened 2026-07-30
feat(ai): default embedder → ollama:bge-m3 @1024 with loud hosted fallback (#3390, supersedes #3481)

time-attack:
## What

The default embedder moves off the sunsetting ZeroEntropy hosted API (dies **2026-09-04**) to **`ollama:bge-m3` at its native 1024 dimensions** — open-weight, local, free, and the strongest open-weight multilingual retriever in the 8-candidate × 6-language eval that drove the decision (vector nDCG@10 ≥ 0.89 on every slice, 0.901 on Hebrew where `text-embedding-3-small` scores 0.645).

Because the default has no hosted vendor, it ships with a **declared hosted fallback**: `openai:text-embedding-3-small` @ **1024**.

This supersedes #3481 (which proposed `openai:text-embedding-3-small` @ 1280 as the default). Its Default-provider policy section, the `gbrain upgrade` ZE-banner DB-plane fix, and the doc sweep are carried forward here, re-targeted.

## How detection works (and what it costs)

- `gbrain init` probes the local Ollama daemon **once** — a single `GET {OLLAMA_BASE_URL|localhost:11434}/api/tags` with a 1.5s timeout, fail-open (`src/core/ai/ollama-detect.ts`). No daemon → instant ECONNREFUSED; worst case 1.5s, once per init.
- The resolved choice persists to `config.json`. **Embed calls never probe.**
- Ollama up + `bge-m3` pulled → the default wins, over every env key (it needs no key and costs nothing).
- Ollama unreachable (or model not pulled) + `OPENAI_API_KEY` set → init lands on the fallback **loudly**: the notice names the reason (`not reachable` vs `running but bge-m3 is not pulled`), the multilingual quality cost, and the paste-ready way back (`ollama pull bge-m3` + `gbrain migrate embeddings --to ollama:bge-m3 --dim 1024`).
- A `embedding_default_fallback` config marker makes **`gbrain doctor` re-probe on every run** (new `embedding_default_fallback` check), so "installed Ollama later" surfaces the switch automatically. The marker is cleared by any re-init that resolves off the fallback, and the doctor check ignores stale markers.
- No Ollama and no OpenAI key → existing behavior unchanged (single other key auto-picks; multi-key picker; non-TTY fail-loud hint now leads with `ollama pull bge-m3`).

## Why the fallback is `text-embedding-3-small`, plainly

It's the key most users already have, it's cheap ($0.02/MTok), and it's from the vendor with the longest hosted-embedding lifetime record — satisfying the Default-provider policy for the hosted slot. **It is also the worst multilingual performer among the evaluated candidates (nDCG@10 0.645 on Hebrew vs bge-m3's 0.901).** That is exactly why the fallback is loud, never silent: the init notice states the trade-off, the docs state it, and doctor keeps offering the way back.

## Dims decisions

- **Default: 1024** — bge-m3's native width. The Matryoshka free-truncation measured for `text-embedding-3-*` and Gemini was NOT verified for bge-m3, so no rounding in either direction.
- **Fallback: 1024, not the model's native 1536** — `text-embedding-3-*` is Matryoshka (`isValidOpenAITextEmbedding3Dim` accepts any width ≤ native; 1024 is already in the recipe's `dims_options`), so pinning the fallback at bge-m3's width makes the later fallback→default migration a **vector-only rebuild**: same `vector(1024)` column, same HNSW index, no dimension transition.

## Existing brains: untouched

**This changes the default for fresh installs only. Existing brains keep their configured model. No migration re-embeds anyone.** Brains still on ZeroEntropy get the existing one-shot `gbrain upgrade` banner, which now (a) also reads the DB config plane so default-config ZE brains that never wrote `embedding_model` to `config.json` are still detected (carried from #3481), and (b) prints both concrete targets: the bge-m3 default (width change 1280→1024, dimension transition) and the same-width hosted option (`--dim <current>`, vectors-only).

## Throughput expectation (documented)

Local bge-m3 embeds ~8× slower than hosted APIs (~7 docs/s vs ~60). For a 10K-page brain that's **~23 minutes for the first full embed instead of ~3 — one time**; incremental syncs are unaffected. In `docs/integrations/embedding-providers.md` and the company-brain tutorial.

## Version trio

Deliberately **no** VERSION / package.json / CHANGELOG edits. #3481 self-allocated `0.42.68.0`, which master has since passed — any number picked at PR time can go stale the same way. Recommendation: let `/ship` allocate at merge (its Step 12 queue-aware allocator handles collisions) and prepend the allocated version to this PR title per the title rule.

## Tests + proof

New: `test/init-ollama-default.test.ts` (14 tests — constants, probe fail-open, resolver happy/fallback/model-missing/other-provider paths with the probe stubbed via `__setOllamaProbeForTests`, doctor re-check warn/informational/stale/skip). Updated: `test/e2e/init-fresh-pglite.test.ts` (subprocess CLI: fallback path + fail-loud hint), `test/e2e/fresh-install-pglite.test.ts` (bare-init happy path end-to-end against a stubbed local tags endpoint), plus the default-pinning sweeps in `gateway.test.ts`, `schema-templating.test.ts`, `v0_37_fix_wave.serial.test.ts`.

**Master discrimination (behavioral, not missing-export):** the updated subprocess e2e tests copied onto an unmodified `origin/master` checkout fail on assertions, with all sibling tests passing (12 pass / 2 fail):

- fallback test — master resolves `openai:text-embedding-3-large (1536d)` (`Detected OPENAI_API_KEY env var. Using openai:text-embedding-3-large (1536d)`) where the test expects the loud 3-small@1024 fallback + marker;
- fail-loud hint test — master's hint prints the ZeroEntropy line and lacks `ollama pull bge-m3`.

Same files pass on this branch. `bun run typecheck` clean; `bun run verify` green (32/32 checks).


time-attack:

The feature itself works: I ran a real fresh `gbrain init --pglite` on a machine with Ollama + bge-m3 pulled and got a working 1024d brain end to end (import → semantic query, top hit 0.8897), and a dead-Ollama + real-OPENAI_API_KEY init landed loudly on openai:text-embedding-3-small@1024 and also worked end to end. The discrimination test passes for real (the updated e2e assertions fail behaviorally on master, pass here). But two blockers survive: (1) the PR breaks an existing CI guard — test/doctor-categories.test.ts fails because the new `embedding_default_fallback` check name is never registered in src/core/doctor-categories.ts (10 pass on master, 1 fail here, plus a stderr nag on every real doctor run); (2) the "Existing brains: untouched" claim is empirically false — a brain whose ~/.gbrain/config.json has no `embedding_model` (the exact v0.36–v0.42 population the PR's own upgrade.ts hunk adds DB-plane detection for) silently flips to ollama:bge-m3@1024 against its vector(1280) column and every import fails, while the PR's own post-upgrade banner simultaneously tells that user the brain "embeds with zeroentropyai:zembed-1". `bun run typecheck` and `bun run verify` are both green; verify does not run unit tests, which is why the red guard was missed.

### Discrimination test

PASSES — verified by execution, not inference. I copied the PR's updated test/e2e/init-fresh-pglite.test.ts onto an unmodified origin/master checkout (c6dc0adf) and ran it: 7 pass / 3 fail, with two genuine behavioral assertion failures — master printed `Detected OPENAI_API_KEY env var. Using openai:text-embedding-3-large (1536d)` where the test expects `default embedder is ollama:bge-m3` + `Falling back to the hosted openai:text-embedding-3-small`, and master's no-key hint lacks `ollama pull bge-m3` (it still lists ZEROENTROPY_API_KEY). The third failure was an unrelated beforeEach hook timeout. The same file is 14 pass / 0 fail on the PR branch. These are behavior assertions, not missing-export import errors. Separately, I flipped ONLY the two default constants back to the master values on the PR branch to isolate cause for the availability regression (ok:true → ok:false) and to clear test/mcp-eval-capture.test.ts of PR blame (it fails either way, and fails in setup on master, so it is pre-existing on this machine).

### Blockers

**1.** CI-RED: src/commands/doctor.ts:78 adds check name 'embedding_default_fallback' to buildChecks but it is never added to BRAIN/SKILL/OPS/META_CHECK_NAMES in src/core/doctor-categories.ts. Executed: `bun test test/doctor-categories.test.ts` → 9 pass / 1 fail on this branch ("These check names appear in doctor check emitters but are not categorized ... embedding_default_fallback"), 10 pass / 0 fail on origin/master. Also user-visible: every real `gbrain doctor` run now prints `[doctor-categories] unknown check name 'embedding_default_fallback' — defaulting to 'meta'.` (reproduced twice against live brains).

**2.** DATA-PATH REGRESSION: existing brains with no `embedding_model` in ~/.gbrain/config.json break completely. src/core/ai/gateway.ts:472 resolves `config.embedding_model ?? DEFAULT_EMBEDDING_MODEL` and src/core/config.ts:loadConfigWithEngine never merges the DB-plane `embedding_model`, so flipping src/core/ai/defaults.ts:385 silently re-points those brains at ollama:bge-m3@1024. Executed on a real PGLite brain created on master (ZE default, column vector(1280), DB config row `zeroentropyai:zembed-1`, embedding_model removed from config.json to model the pre-v0.37.10 population the PR itself names): master `gbrain doctor` → `[OK] embedding_width_consistency: Schema width (1280d) matches gateway embedding_dimensions`; PR-branch `gbrain import /tmp/oldrepo` → exit 1, `Warning: skipped test-page.md: expected 1280 dimensions, not 1024`, 0 pages imported. PR-branch doctor's remediation is destructive and points at the NEW default (`gbrain reinit-pglite --embedding-model ollama:bge-m3 ...` / "drops every embedding in your brain"), and PR-branch `gbrain post-upgrade` on the same brain prints "This brain embeds with zeroentropyai:zembed-1" — the banner reads the DB plane, the pipeline does not. No pin-migration or DB-plane merge ships with the flip.

**3.** FAIL-OPEN AVAILABILITY PREDICATE (root cause not fixed at the shared choke point): the ollama recipe declares `auth_env.required: []`, so with the new default `diagnoseEmbedding()`/`isAvailable('embedding')` report success on a machine with zero keys and no Ollama daemon. Executed: `configureGateway({env:{}})` then diagnoseEmbedding() → `{ok:true, model:'ollama:bge-m3', isAvailable:true}` on this branch; flipping ONLY DEFAULT_EMBEDDING_MODEL/DIMENSIONS back to the master values in src/core/ai/defaults.ts gives `{ok:false, reason:'missing_env', isAvailable:false}`. The new reachability probe went into init (src/commands/init.ts:537) and one doctor check, not into the shared predicate that 9 call sites use (src/core/operations.ts:937 put_page noEmbed gate, src/core/cycle/extract-facts.ts:443, src/core/cycle/synthesize-concepts.ts:241, src/commands/autopilot.ts:1163, src/commands/search-diagnose.ts:63, src/commands/doctor.ts:6057). Consequence executed on the live default brain with OLLAMA_BASE_URL pointed at a dead port: `gbrain import` → exit 1 with a bare transport error (`[embed(ollama:bge-m3)] Failed after 3 attempts. Last error: Cannot connect to API...`), no `--no-embed` guidance, no 'start ollama serve' hint — whereas the missing-hosted-key path still prints the designed actionable message with the --no-embed escape.

**4.** GOLDEN-PATH FALSE ALARM: on a machine that IS correctly set up (Ollama running, bge-m3 pulled), `gbrain init --pglite --non-interactive` prints `Heads up: an embedding key is set but a test embed failed (timeout)`. Cause: src/core/init-embed-check.ts:90 uses a 5000ms live-probe budget, but a cold bge-m3 load takes longer — measured on this machine 5.09s cold vs 0.18s warm (`curl /api/embeddings`). Every first install of the new default embedder will show a scary failure notice; the embed path itself is fine (GBRAIN_AI_EMBED_TIMEOUT_MS default 60s).

**5.** UNDISCLOSED INIT-SEMANTICS CHANGE: src/commands/init.ts:561 places the fallback branch BEFORE the ready.length checks, so any install with OPENAI_API_KEY plus other embedding keys now auto-resolves to text-embedding-3-small@1024 instead of hitting the deliberate D2/D3 disambiguation gate. Executed with OPENAI+VOYAGE+MINIMAX keys and a dead Ollama, non-TTY: master → exit 1 `Multiple embedding providers env-ready: openai, voyage, minimax`; this branch → exit 0, `Falling back to the hosted openai:text-embedding-3-small (1024d)`. The PR body and docs/integrations/embedding-providers.md both still claim "multiple keys fire an interactive picker", and no D2/D3 test covers the bypass — so a Voyage user silently gets the PR's own self-declared weakest multilingual model.

**6.** CLAUDE.md doc defect: the diff replaces the `## Eval discipline (v0.32.3)` heading with `## Default-provider policy` instead of inserting above it, so the whole eval-discipline block (metric glossary, methodology, .gbrain-evals records) is now orphaned under the new heading in the always-loaded file. Verified on the branch: `grep '^## ' CLAUDE.md` shows no Eval-discipline section, and CLAUDE.md:262-279 has that text sitting under Default-provider policy.

**7.** Process (repo IRON RULES): PR title is `feat(ai): default embedder → ...` with no leading version — CLAUDE.md requires `vMAJOR.MINOR.PATCH.MICRO <type>(<scope>): ...`. No CHANGELOG entry for a change that alters the default embedder for every fresh install (the PR deliberately defers VERSION/package.json/CHANGELOG to /ship, which is defensible, but the CHANGELOG text still has to exist before merge). scripts/test-weights.json is not updated for the new test file (falls back to corpus median — non-blocking).

### Not verified

- Postgres/Supabase engine path — only PGLite was exercised locally (no DATABASE_URL available in this environment), so the initPostgres fallback-marker hunk and the width-mismatch behavior on a real vector(1280) Postgres column are inferred from the identical PGLite code path, not executed.
- Full unit suite / all CI shards — shard 1/6 was terminated early by my branch switches. I ran ~1050 targeted tests across the embedding/search/doctor/init/upgrade/e2e-pglite surface instead; other shards could hide additional stale-default assertions (several test comments still say 'DEFAULT_EMBEDDING_DIMENSIONS is 1280' in search-types-filter, consolidate-valid-until, query-cache, find-experts, operations-find-trajectory — those files pass, the comments are just stale).
- The exact size of the affected existing-brain population. I demonstrated the failure on a faithfully reconstructed brain (v0.36-era shape: 1280 column + DB-plane ZE row + no config.json embedding_model), and the PR's own upgrade.ts comment asserts such brains exist, but I could not enumerate real installs.
- bge-m3 multilingual eval numbers (nDCG@10 0.901 Hebrew vs 0.645 for 3-small) — quoted from the PR; no eval was rerun here.
- The claimed ~8x throughput gap (~7 vs ~60 docs/s) — not benchmarked; only single-call latency was measured (5.09s cold / 0.18s warm on this machine).

- `gh pr view 3611 → 0`
- `gh pr diff 3611 → 0`
- `git fetch origin pull/3611/head:pr3611-local && git checkout pr3611-local → 0`
- `bun install → 0`
- `bun run typecheck → 0`
- `bun run verify → 0 (32/32 checks)`
- `bun test test/init-ollama-default.test.ts → 0 (14 pass)`
- `bun test test/e2e/init-fresh-pglite.test.ts (PR branch) → 0 (14 pass)`
- `bun test test/e2e/init-fresh-pglite.test.ts (PR test file on origin/master src) → 1 (7 pass / 3 fail: 2 behavioral + 1 hook timeout)`
- `bun test test/doctor-categories.test.ts (PR) → 1 (9 pass / 1 fail)`
- `bun test test/doctor-categories.test.ts (master) → 0 (10 pass)`
- `bun test test/ai/ test/embedding-pricing test/embedding-dim-check{,-facts} test/doctor-ze-checks test/ze-switch-* test/embedding-migration test/retrieval-upgrade-planner test/doctor-embedding-env-override test/search-mode test/query-cache → 0 (710 pass / 0 fail across 54 files)`
- `bun test test/doctor*.test.ts subset + advisor-* + init-env-detection + init-embed-check + init-provider-picker + upgrade-* (15 files) → 1 (227 pass / 1 fail: only doctor-categories)`
- `bun test test/e2e/fresh-install-pglite.test.ts test/v0_37_fix_wave.serial.test.ts test/v0_37_gap_fill.serial.test.ts → 0 (33 pass)`
- `bun test test/build-llms.test.ts → 0 (12 pass)`
- `bun test test/config.test.ts test/config-set.test.ts → 0 (50 pass)`
- `bun test test/e2e/embedding-column-pglite.test.ts → 1 on BOTH PR and master (pre-existing, not PR-caused)`
- `bun test test/mcp-eval-capture.test.ts → 1 on PR, 1 with old default flipped back, 1 (setup timeout) on master → pre-existing/env, not PR-caused`
- `bash scripts/test-shard.sh 1 6 → terminated early by my branch switches; only failures observed were the pre-existing mcp-eval-capture pair`
- `GBRAIN_HOME=/tmp/oldbrain bun src/cli.ts init --pglite --non-interactive (master src, ZE-only env) → 0`
- `GBRAIN_HOME=/tmp/oldbrain bun src/cli.ts import /tmp/oldrepo (PR src, embedding_model stripped from config.json) → 1 ("expected 1280 dimensions, not 1024", 0 imported)`
- `GBRAIN_HOME=/tmp/oldbrain bun src/cli.ts doctor → 0 on both; master: width OK / PR: 3 WARNs incl. destructive remediation`
- `GBRAIN_HOME=/tmp/oldbrain bun src/cli.ts post-upgrade (PR) → 0 (DB-plane ZE detection + Option A/B banner works)`
- `GBRAIN_HOME=/tmp/newbrain bun src/cli.ts init --pglite --non-interactive (real local Ollama) → 0 (ollama:bge-m3/1024, but prints 'test embed failed (timeout)')`
- `GBRAIN_HOME=/tmp/newbrain bun src/cli.ts import /tmp/newrepo → 0 (2 pages, 2 chunks); query 'what color are the widgets' → 0 (0.8897 correct top hit)`
- `GBRAIN_HOME=/tmp/newbrain OLLAMA_BASE_URL=dead bun src/cli.ts import → 1 (bare ECONNREFUSED, no --no-embed hint)`
- `GBRAIN_HOME=/tmp/fbbrain OLLAMA_BASE_URL=dead bun src/cli.ts init --pglite (real OPENAI key) → 0 (loud fallback + embedding_default_fallback marker); import → 0; query → 0; doctor → 0 (fallback nag fires); migrate embeddings --to ollama:bge-m3 --dim 1024 --dry-run → 0 (non-destructive, as advertised)`
- `multi-key non-TTY init (OPENAI+VOYAGE+MINIMAX, dead Ollama): PR → 0 (auto-fallback), master → 1 (disambiguation gate)`
- `curl -m60 localhost:11434/api/embeddings bge-m3 → 5.09s cold, 0.18s warm`
- `bun probe_avail.ts (configureGateway({env:{}}) + diagnoseEmbedding): PR → {ok:true,model:'ollama:bge-m3'}; DEFAULT_* flipped to master values → {ok:false,reason:'missing_env'}`

</details>

---

RoamingQuack:
How about letting people define and run their own embedding models instead of vendor lock-in all over again, fellas? 

I'd love to use Nemotron3-embed-8b @ 4096-d, hosted locally. pgvector maxes at 2000-d, or 4000-d halfvec. its also time to upgrade to pgvectorscale or allow other vector stores.

time-attack:

Not superseded: current master src/core/ai/defaults.ts:20-21 still defaults to zeroentropyai:zembed-1@1280, which dies 2026-09-04 (p1/verified-real issue #3390) — the merged #3459 (v0.42.67.0) shipped only the migration path, not the default flip, so this is fix-lane work, not an off-mission feature. But the PR's own 2026-07-30 adversarial review returned NEEDS_CHANGES and zero commits have landed since (single commit b9eaebc3, 2026-07-29). I independently confirmed the two hard blockers: (1) CI-red — the new 'embedding_default_fallback' doctor check (src/commands/doctor.ts) is absent from src/core/doctor-categories.ts on the PR head (grep returns nothing), so test/doctor-categories.test.ts fails; (2) data regression — master gateway.ts:472/773/844 resolve config.embedding_model ?? DEFAULT_EMBEDDING_MODEL and the PR ships no pin migration, so existing 1280d brains lacking a config.json embedding_model silently re-point to bge-m3@1024 and every import fails ('expected 1280 dimensions, not 1024', reviewer-executed), contradicting the PR's 'existing brains untouched' claim. Also confirmed: the init hosted-fallback tier is inserted before the ready.length multi-key disambiguation gate (silent bypass of D2/D3), and the CLAUDE.md diff replaces the '## Eval discipline (v0.32.3)' heading with '## Default-provider policy', orphaning that section.

**Conflict state vs current master:** Test-merge onto 532e5916 yields exactly one content conflict: src/core/embedding-pricing.ts — the wave refreshed the pricing table (verified-2026-07-28 comments, voyage-4-nano no-$0-entry policy) at the same lines where the PR inserts 'ollama:bge-m3': {pricePerMTok: 0}. Trivially resolvable, though the $0 entry should be reconciled with the wave's FREE_LOCAL_CHAT_PROVIDERS approach in src/core/budget/budget-tracker.ts (#3541) and the voyage-4-nano precedent against $0 rows for open-weight models. The fix itself still applies post-wave — nothing on master flips the default — and the wave actually helps it: src/core/ai/recipes/ollama.ts:38 now carries the bge-m3 native-width 1024 map and #3699 suppresses native-width dims params. Remaining post-merge chores: CLAUDE.md/llms-full.txt auto-merged cleanly but llms must be regenerated (test/build-llms.test.ts freshness gate) and the stale eval numbers/heading fixed.

**Risk if merged as-is:** Merging as-is turns master red: test/doctor-categories.test.ts fails (uncategorized check name) and test/build-llms.test.ts will likely fail on the merged-but-unregenerated llms bundle. Worse than CI: existing pre-v0.37.10 brains (no config.json embedding_model pin) break on import against their vector(1280) columns, with doctor offering a destructive reinit as remediation — a user-facing data-path regression.

time-attack:
Superseded: v0.46.3.0 (#4160) flipped new installs off ZeroEntropy to voyage:voyage-4@1024 (defaults.ts:47) with sunset migration + playbook, resolving #3390. The ollama:bge-m3 design wasn't adopted; NEEDS_CHANGES blockers were never addressed and the branch conflicts with master.

PR #3613, opened 2026-07-30
fix(search): raise hnsw.ef_search to match the vector candidate request

paul-0320:
## Problem — the vector search asks for 100+ candidates and can only ever receive 40

`searchVector`'s inner CTE requests `offset + max(limit*5, 100)` candidates ([postgres-engine.ts#L2156](https://github.com/garrytan/gbrain/blob/945fed61/src/core/postgres-engine.ts#L2156), [pglite-engine.ts#L2106](https://github.com/garrytan/gbrain/blob/945fed61/src/core/pglite-engine.ts#L2106)) so the source-boost re-rank and the per-page collapse (T1) operate on a wide pool. But a pgvector HNSW index scan returns **at most `hnsw.ef_search` rows** — the GUC sizes the scan's candidate list, so it caps the row count before the CTE's `LIMIT` is even applied. pgvector's default is **40**, and nothing in the codebase sets it (`grep -rn "ef_search" src/` on `945fed61` → 0 hits).

So on any HNSW-indexed corpus, every candidate request past 40 is silently truncated. Everything downstream that was sized around the innerLimit budget — per-page collapse, RRF fusion, `top_n_in`-windowed rerankers — runs on roughly a third of the pool it was designed for. The engine pays for a wide net and receives a narrow one, with no error and no signal.

Measured on our production brain (Postgres 18.4, ~22K chunks, 1536d, `EXPLAIN ANALYZE` on the exact inner-CTE shape, 125-candidate request):

| ef_search | chunk candidates received | → unique pages | exec time |
|---|---|---|---|
| 40 (default) | **35** | 24 | 64.7 ms |
| 200 | **125** (full request) | 81 | 86.5 ms |

3.6× the candidates for +22 ms — while a local reranker stage spends 2,000–7,000 ms per query on the (starved) pool it receives.

## What this does

One shared helper, applied in both engines' `searchVector`:

- `hnswEfSearchFor(candidateLimit)` (new, `vector-index.ts`): the inner CTE's candidate limit, floored at pgvector's default (40) and capped at the GUC max (1000).
- **Postgres**: the search already runs in a transaction that does `SET LOCAL statement_timeout` ([postgres-engine.ts#L2302](https://github.com/garrytan/gbrain/blob/945fed61/src/core/postgres-engine.ts#L2302)) — add a transaction-scoped `set_config('hnsw.ef_search', …, true)` beside it.
- **PGLite**: the vector query was a bare autocommit `db.query`; it gains a `db.transaction` wrap so the same `SET LOCAL` semantics apply to the engine's single session without leaking.

No config surface: the value derives from what the query already asks for — the engine should receive the candidates it requests. Non-HNSW plans (seq scan, ivfflat, corpora where the index was skipped for dimension limits) ignore the GUC entirely.

## Regression tests

`test/vector-ef-search.test.ts` (unit — helper formula + PGLite) and `test/e2e/vector-ef-search-postgres.test.ts` (gated on `DATABASE_URL` like the other e2e files), sharing a fixture (`test/helpers/ef-search-fixture.ts`):

- 150 pages × 1 chunk, seeded-PRNG unit vectors, `searchVector` with `limit=100`. **Before the fix both engines return exactly 40 rows; after, 100.**
- The HNSW-ordered plan is forced with `enable_seqscan=off` + `enable_sort=off` (the technique pgvector's own regression suite uses) — on a 150-row corpus the planner otherwise picks a seq scan or another index + explicit Sort, which bypasses the cap and would let the test pass without the fix. `enable_sort` matters because the joined query can satisfy "no seq scan" via a different index and still sort explicitly.
- Postgres forcing goes through `ALTER DATABASE … SET` + a fresh `poolSize` engine (a session-level `SET` would land on one pooled connection while the search transaction may run on another; without `poolSize` the engine reuses the module-singleton pool whose connections predate the `ALTER`). `afterAll` RESETs both GUCs.
- Fixture geometry note (documented in the fixture): structured embeddings (basis vectors, smooth angle gradients) build degenerate HNSW graphs whose greedy search plateaus far below `ef_search` (a basis-vector corpus stalled at ~34 rows even with `ef_search=500`), which would mask the fix. Seeded-random geometry keeps the graph connected so the only cap left is the GUC under test.

## Measurements — end-to-end

**Eval replay characterization** (production pipeline: hybrid + multi-query expansion + reranker + autocut, 62 queries, `query_cache` purged per condition, ef 40 → 200):

| metric | value |
|---|---|
| Jaccard@10 | 0.841 |
| top-1 stability | 57/62 (91.9%) |
| latency Δ | ≈ 0 (6.10 s → 6.18 s avg; worst case 7.51 s → 7.03 s) |

Answer-level effects are boundary churn, slightly net-positive: answers found in top-20 went 53 → 54 (2 recovered — one straight to rank 1 — vs 1 dropped at the autocut boundary), mean answer rank 1.83 → 1.72, and no answer moved ≥3 ranks between arms. That is the expected shape: the fix widens the pool the re-rank stages see; it does not reorder healthy results.

This change has been running in our production brain (as the equivalent `ALTER DATABASE … SET hnsw.ef_search = 200`) since 2026-07-30: no latency regression (±0.2 s per query on a 6-question spot set), rank-stable results.

## Verification

```
bun test test/vector-ef-search.test.ts                              # 4 pass (helper + PGLite regression)
DATABASE_URL=… bun test test/e2e/vector-ef-search-postgres.test.ts  # 1 pass
bun run verify                                                      # 32/33 — the one failure is typecheck, pre-existing on the base commit (see note)
bash scripts/run-unit-parallel.sh                                   # 7529 pass / 9 fail — the same 9 fail on clean 945fed61 (env-coupled, unrelated files)
DATABASE_URL=… bash scripts/run-e2e.sh                              # all green except 2 pre-existing cross-modal-eval.test.ts failures (also fail on clean 945fed61)
```

Note on the base: `945fed61`'s own CI shows `verify: failure` / `test-status: failure` — `tsc --noEmit` currently fails master in `test/admin-sse-handshake.test.ts` + `test/serve-http-lifecycle.test.ts`. This branch adds no new typecheck errors (verified by diffing `tsc` output against the clean base commit), and the regression-test pair was verified fail-before/pass-after on both engines by reverting `src/` to the base and re-running.


time-attack:

The bug is real and the fix is correct: pgvector's hnsw.ef_search default (40) silently caps searchVector's candidate pool below the offset+max(limit*5,100) it requests; I reproduced the exact 40-row cap on both PGLite and real Postgres 17+pgvector by stubbing the fix out, and both regression tests discriminate perfectly (40 without fix, 100 with). Engine parity, timeout backstop (8s SET LOCAL + ef capped at 1000), source isolation, and merge cleanliness all check out. But the change alters which rows hybrid search returns (author's own Jaccard@10=0.841) without bumping KNOBS_HASH_VERSION, so stale pre-fix cache rows get served for up to an hour post-upgrade — repo precedent (13→14 for the #3430 fix) requires the bump plus its 4 test pin sites. Also missing: version-first PR title and the VERSION/CHANGELOG bump. One-line fix + pin updates away from MERGE.

### Discrimination test

Reverted only the two engine files to base 945fed61 (kept the new helper so imports resolve) and re-ran the PR's own tests: PGLite regression failed with Expected:100 Received:40, and the Postgres e2e against Docker pgvector:pg17 failed identically (40 rows). Restoring the PR's engine changes made both pass at 100 rows. The tests discriminate perfectly and empirically confirm the 40-row ef_search cap on both engines.

### Blockers

**1.** Missing KNOBS_HASH_VERSION bump (src/core/search/mode.ts:769, 14→15) + its 4 test pin sites (test/search-mode.test.ts:419,585; test/cross-modal-phase1.test.ts:139-150; test/search-alias-resolved-boost.test.ts:93; test/search-compiled-truth-boost-scope.test.ts:74). The fix changes hybrid results for identical knobs (PR's own Jaccard@10=0.841), so post-upgrade query_cache lookups serve pre-fix truncated-pool rows for up to ttl=3600s. Repo precedent: 13→14 was bumped for the #3430 result-affecting fix, same class.

**2.** PR title violates the version-first IRON RULE: must be 'vX.Y.Z.W fix(search): raise hnsw.ef_search …', currently starts with 'fix(search):'.

**3.** No VERSION/package.json/CHANGELOG bump: branch trio agrees at 0.42.67.0 but master is 0.42.68.1; CI version-gate requires VERSION strictly greater than master's.

### Not verified

- PR body's production EXPLAIN ANALYZE latency numbers (64.7→86.5ms @ 22K chunks) and the 62-query eval-replay metrics (Jaccard@10 0.841, top-1 stability) — require the author's production brain + API keys; the core 40-row-cap claim was independently reproduced instead
- Full unit suite / bun run verify not run (shared-machine footprint rules); PR claims same 9 pre-existing failures as base
- Typecheck failures confirmed pre-existing only by file scope (both failing files untouched by this PR and disclosed in the PR body); did not re-run tsc on the bare base commit

- `gh pr view 3613 --json mergeable,mergeStateStatus → MERGEABLE/CLEAN`
- `git fetch origin pull/3613/head:pr-3613 && git checkout pr-3613; bun install`
- `bun test test/vector-ef-search.test.ts → 4 pass (fix present)`
- `git checkout 945fed61 -- src/core/{pglite,postgres}-engine.ts; bun test test/vector-ef-search.test.ts → FAIL Expected:100 Received:40 (discrimination test, PGLite)`
- `docker run pgvector/pgvector:pg17; DATABASE_URL=… bun test test/e2e/vector-ef-search-postgres.test.ts → 1 pass (fix present)`
- `revert postgres-engine.ts to base; same e2e → FAIL Expected:100 Received:40 (discrimination test, real Postgres); restored; docker rm -f`
- `bun run typecheck → EXIT=2, 3 errors all in test/serve-http-lifecycle.test.ts + test/admin-sse-handshake.test.ts (files untouched by PR)`
- `git diff 945fed61 origin/master -- <3 src files + mode.ts> → empty (no master drift in touched files)`
- `grep audits: all '<=>' HNSW scan sites (takes search LIMIT≤100 sibling gap; facts k≤20 ok; query_cache LIMIT 1 ok), KNOBS_HASH_VERSION pin sites, VERSION trio vs origin/master`

</details>

---

time-attack:
Eval-replay gate (CONTRIBUTING.md retrieval requirement) — **PASS**. Replayed the committed hermetic baseline (`gbrain-evals/baselines/v0.41-launch.baseline.ndjson`, 12 rows — all that exists publicly) on a PGLite brain seeded from the qrels fixtures, PR merged onto v0.42.69.0 master: mean Jaccard@k **1.000** (baseline 1.000), top-1 stability **0.9167** (baseline 0.9167 — the one flip is a pre-existing deterministic drift from the v0.41 capture, identical on master), mean latency ~5.2ms (master band 5.3–7.8ms), 3 runs, zero regressed rows.

Two honest limits: the corpus is keyword-only, and at 24 chunks it sits below pgvector's ef_search default of 40, so the recall *improvement* this PR exists for is unobservable here — what the gate proves is the risky part: the new PGLite transaction wrapper + Postgres `set_config` in `searchVector` execute with zero result drift, confirmed by a supplementary hybrid vector-arm hand-query diff (master vs PR: identical on 3 queries). Queued for the next fix wave.

time-attack:

PR #3616, opened 2026-07-30
fix(search): classify first-person "what do I know about X" as an entity query

### LINKED ISSUE #3615: classifyQuery treats "what do I know about X" as a general query, unlike the you/we phrasings
`classifyQuery` matches `what do you know` and `what do we know` in `ENTITY_PATTERNS`, but not `what do I know`. The first-person phrasing of the same question falls through to `intent: 'general'`.

Observed on current master (c6dc0adf), `src/core/search/query-intent.ts`:

| query | intent | suggestedDetail |
|---|---|---|
| `what do I know about widget-co` | `general` | `undefined` |
| `what do you know about widget-co` | `entity` | `low` |
| `what do we know about widget-co` | `entity` | `low` |

`autoDetectDetail` consumes `suggestedDetail`, so three interchangeable phrasings of one question produce two different retrieval plans.

Looks like a gap in the alternation rather than a deliberate exclusion - the surrounding patterns (`overview`, `background`, `profile`) are all person-agnostic. Happy to send the one-line patch plus a test if you agree it should match.

SeanGearin:
Fixes #3615.

`ENTITY_PATTERNS` matched `what do you know` and `what do we know` but not `what do I know`, so the first-person phrasing fell through to `intent: 'general'` with `suggestedDetail: undefined`, while the you/we phrasings returned `entity` / `low`. `autoDetectDetail` feeds `suggestedDetail` into the retrieval plan, so three interchangeable phrasings produced two different plans.

**Before and after** on `c6dc0adf`:

| query | before | after |
|---|---|---|
| `what do I know about widget-co` | `general` / `undefined` | `entity` / `low` |
| `what do you know about widget-co` | `entity` / `low` | unchanged |
| `what do we know about widget-co` | `entity` / `low` | unchanged |

**The change** is one character in the alternation: `(you|we)` becomes `(i|you|we)`. `query-intent.ts` has no imports and no I/O, so nothing outside the pattern list is reachable from this diff. I deliberately did not add the pattern to `CANONICAL_PATTERNS` - that would force both recency and salience off for hybrid phrasings like "prep me for acme - what do we know", which the module docblock says should keep them on, and no test covers it.

**Receipts.** The added test fails on unpatched master with `Expected: "entity", Received: "general"` and passes after. `bun test test/query-intent.test.ts test/query-intent-legacy.test.ts test/cross-modal-phase1.test.ts` is 102 pass / 0 fail. `tsc --noEmit` and `check:test-names` both clean. I did not run the full suite to completion - it provisions Postgres and applies 120 migrations, and this module touches neither.

**One note in the interest of not wasting your time:** of the four assertions in the new test, only `intent` and `suggestedDetail` change behavior. `suggestedRecency` and `suggestedSalience` already pass on unpatched master - they pin existing behavior rather than prove the fix. Happy to trim them if you would rather the test assert only what moved.

### DIFF
--- a/src/core/search/query-intent.ts
+++ b/src/core/search/query-intent.ts
@@ -96,7 +96,7 @@ const ENTITY_PATTERNS = [
   /\boverview\b/i,
   /\bbackground\b/i,
   /\bprofile\b/i,
-  /\bwhat\s+do\s+(you|we)\s+know\b/i,
+  /\bwhat\s+do\s+(i|you|we)\s+know\b/i,
 ];
 
 const FULL_CONTEXT_PATTERNS = [
--- a/test/query-intent.test.ts
+++ b/test/query-intent.test.ts
@@ -33,6 +33,14 @@ describe('classifyQuery — entity / canonical queries → both axes off', () =>
     expect(r.suggestedSalience).toBe('off');
   });
 
+  test('"what do I know about widget-co" → entity, same as you/we phrasings', () => {
+    const r = classifyQuery('what do I know about widget-co');
+    expect(r.intent).toBe('entity');
+    expect(r.suggestedDetail).toBe('low');
+    expect(r.suggestedRecency).toBe('off');
+    expect(r.suggestedSalience).toBe('off');
+  });
+
   test('"history of X" → both off (canonical)', () => {
     const r = classifyQuery('history of acme corp');
     expect(r.suggestedRecency).toBe('off');


time-attack:
**Merging this closes #3615.**

Independent triage of #3615 reproduced the reported bug against current master and then confirmed this PR is what fixes it. This PR's first-person entity-query classification is exactly the case that issue describes.

Worth adding `Closes #3615` to the PR description — a closing keyword only auto-closes from the body or a commit message, not from a comment like this one, so right now the link is visible but not wired up.

SeanGearin:
Thanks for the repro on #3615 - good to have it confirmed against master rather than taking my word for it.

On the closing keyword: it is already in the body - the PR opens with `Fixes #3615`, and GitHub has it registered (`closingIssuesReferences` returns #3615, base is `master`). So it should auto-close on merge with no edit needed.

PR #3617, opened 2026-07-30
feat(search): make the keyword-arm AND→OR fallback a mode knob (search.keywordOrFallback)

paul-0320:
## Problem — on unstemmed corpora the AND→OR fallback floods, and ts_rank can't clean it up

The hybrid keyword arm opts into the engine's zero-recall AND→OR retry unconditionally ([hybrid.ts — `orFallback: true`](https://github.com/garrytan/gbrain/blob/945fed61/src/core/search/hybrid.ts#L1003)). For stemmed languages that's the right default: one bad token no longer zeroes keyword recall (fix/title-retrieval-arm, D2).

But `websearch_to_tsquery('english', …)` can't stem languages outside its config. For CJK/agglutinative text the document side is indexed as surface forms (particles attached), so on conversational queries **strict AND collapses on nearly every query** — tokens like `조달청에` ("to the procurement agency") never co-occur — and the OR retry fires constantly. Then two things compound:

- The OR bag contains conversational filler ("그래서"/"so", not in the English stopword list), and common fillers match huge slices of the corpus.
- `ts_rank` has **no IDF term** — nothing demotes a common-token hit relative to a rare-token hit.

Measured on our production brain (Postgres 18.4, ~22K chunks, Korean-heavy):

- A representative conversational query: strict AND → **0 rows**; OR retry → **5,577 rows (25% of the corpus)**, with one filler token alone contributing 5,337.
- Rank inversion: a filler-only document scored `ts_rank` **36% higher** than the true answer for the same OR query.

So on these corpora the relaxed rows blend into RRF as noise on nearly every query — the keyword arm's contribution goes negative, not zero. The right operator remedy is to turn the fallback off there; today that requires a fork patch.

## What this does

Exposes the existing engine flag as a mode knob, following the `intentWeighting` pattern end to end:

- `ModeBundle.keywordOrFallback` — **default `true` in every bundle** (the previous hardcoded behavior; nothing changes unless an operator opts out).
- Per-key config override **`search.keywordOrFallback`** (same parse and precedence as `search.intentWeighting`), registered in `SEARCH_MODE_CONFIG_KEYS` and shown in the `gbrain search modes` dashboard.
- [hybrid.ts](https://github.com/garrytan/gbrain/blob/945fed61/src/core/search/hybrid.ts#L1003) threads the resolved knob into the keyword arm's engine opts. Direct `searchKeyword` consumers (countMentions, link-extraction, eval) are untouched — they never set the flag and keep the strict-AND contract.
- **Cache safety**: `kof=` joins `knobsHash` (append-only) and `KNOBS_HASH_VERSION` bumps 14→15 — a fallback-on row must never serve a fallback-off lookup (the zero-strict-recall result sets are disjoint). Heads-up: #3584 (open) also bumps 14→15; whichever lands second rebases to 16 — happy to do that on this side.

## Measurements

**Eval replay characterization** (production pipeline: hybrid + reranker + autocut, 62 Korean-heavy queries, `query_cache` purged per condition, fallback on → off):

| metric | value |
|---|---|
| Jaccard@10 | 0.756 |
| top-1 stability | 57/62 (91.9%) |
| latency Δ | **−0.51 s** (6.22 s → 5.71 s avg — fewer flood rows for the reranker to score) |

The fallback fires broadly on this corpus — **43/62 queries' result lists changed** when it was turned off — yet recall is a wash: answers found in top-20 went 50 → 50, with one answer recovered from under the flood (a conversational query whose answer had been crowded out) traded against one genuine fallback rescue lost (a rare-token query — exactly the case the fallback exists for, and why the default stays on). Mean answer rank 1.60 → 1.74 (churn in both directions, no systematic shift); mean returned-list length 6.6 → 6.2. This is the intended shape for an operator knob: on corpora where the fallback mostly injects noise, turning it off is quality-neutral and strictly cheaper; elsewhere the default keeps the D2 rescue.

## Verification

```
bun test test/search-mode.test.ts                    # knob resolution, config parsing, kof= hash participation
bun test test/search/title-retrieval-arm.test.ts     # end-to-end pair through hybridSearch (PGLite): default rescues, config=false suppresses
bun run verify                                       # 32/33 — the one failure is typecheck, pre-existing on the base commit
```

The new hybrid-level pair starves the title arm on purpose (page title shares no token with the query) so the OR retry is the only rescue path — the knob's effect is observable end-to-end, not just at the resolver. Existing `KNOBS_HASH_VERSION` pins updated (4 files).

Note on the base: `945fed61`'s own CI shows `verify: failure` (typecheck errors in `test/admin-sse-handshake.test.ts` + `test/serve-http-lifecycle.test.ts`); this branch adds no new typecheck errors.


time-attack:
**Approved on review — but #3677 just landed and took `KNOBS_HASH_VERSION = 15`, so this needs the rebump you already agreed to.**

I test-merged this against current master and it conflicts on exactly five files:

```
src/core/search/mode.ts
test/cross-modal-phase1.test.ts
test/search-alias-resolved-boost.test.ts
test/search-mode.test.ts
test/search/knobs-hash-reranker.test.ts
```

All five are the version constant and its pin sites — no conflict in your actual feature code. The constant is single-writer, which is why the cluster has to land one at a time.

**What it needs:** rebase onto master, take **16**, and update all five pins. Two notes from watching this exact thing go wrong before:

- **Update every pin site, and run those five files specifically.** A previous knobs-hash bump missed three pins and only failed on CI shards 3, 8 and 10 — a targeted subset run looked green.
- **The conflict in `mode.ts` is textual, not semantic.** Master's v15 folds in the FTS configuration name; yours appends the keyword-fallback knob. Both parts belong in the final version — take master's addition *and* yours, then set the constant to 16. Resolving directionally with `--ours`/`--theirs` will silently drop one of them, and the tests will still pass because each part is hashed independently.

Everything else about this PR cleared review: default behavior is unchanged for existing users, both new tests discriminate (reverting the change fails them), and the live CLI behavior was verified. The eval numbers in the description come from your private corpus so I did not treat them as load-bearing, but nothing rests on them.

I am doing review and triage rather than authoring changes at the moment, so I would rather you take the rebase than have me rewrite your PR — you know which parts are load-bearing. Ping me when it is pushed and I will re-check and merge.

For sequencing context: #3621 takes 17 behind you, and #3613 needs to *add* a bump as part of its own fix, since it changes results without fencing the cache.

time-attack:

The gap is real on current master: src/core/search/hybrid.ts:1004 still hardcodes orFallback: true and no keywordOrFallback knob exists anywhere on origin/master, so CJK/unstemmed-corpus operators still have no way to suppress the measured OR-retry flood short of a fork patch. Though labeled "feat", this is an operator remedy for a retrieval-quality degradation, mirrors the established intentWeighting knob pattern, keeps the default unchanged, fences the cache via knobsHash, and reviewer time-attack already approved the substance with an explicit "rebase to 16 and I will merge". It cannot land as-is: #3677 (merged as b4a9c768) took KNOBS_HASH_VERSION=15 (src/core/search/mode.ts:782), which this PR also claims.

**Conflict state vs current master:** Test-merge against origin/master (532e5916) conflicts on exactly 5 files: src/core/search/mode.ts plus 4 test files that pin KNOBS_HASH_VERSION (test/cross-modal-phase1.test.ts, test/search-alias-resolved-boost.test.ts, test/search-mode.test.ts, test/search/knobs-hash-reranker.test.ts). The feature code (hybrid.ts, commands/search.ts, title-retrieval-arm test) merges clean. The mode.ts conflict is a single-writer version-constant collision, semantic in one respect: master's v=15 added the fts= hash part (#3677) on the same lines where this PR adds kof= and its 14->15 comment. Correct resolution keeps BOTH parts, sets the constant to 16, and updates all 5 pin sites; a directional --ours/--theirs resolution would silently drop one cache-fencing part while tests still pass. The fix fully applies post-v0.42.69.0 — nothing in the wave gates the OR fallback.

**Risk if merged as-is:** If the conflict is resolved directionally, either the fts= or kof= knobsHash part gets silently dropped (each is hashed independently, so per-part tests stay green), reopening a query_cache contamination hole. Also sequencing: #3621 is slated to take v=17 behind this PR's v=16 — landing order matters for the single-writer constant. Bumping to 16 causes the usual one-time global cache cold-miss (accepted pattern). Otherwise none: default behavior is unchanged for all existing users.

paul-0320:
Rebased onto current master (`2e554500`) — this was conflicting since #3677 merged and took `KNOBS_HASH_VERSION` 15.

Applied:
- `KNOBS_HASH_VERSION` now **17**. 15 is the merged #3677 (`fts=`), 16 is the sibling #3621 (`acm=`), so this PR claims the next free number rather than colliding.
- `kof=` appended after `fts=` in the `knobsHash` array — both parts kept.
- Bump-rationale comments merged; the four version pins and their titles updated to 17.

Verification on the rebased head: `bun run verify` 34/34 · `bunx tsc --noEmit` clean · patch-adjacent suites 464/464 (`test/search-mode`, `test/search/`, `test/cross-modal-phase1`, `test/search-alias-resolved-boost`), including the `keywordOrFallback` bundle/config/hash tests this PR adds.

Coordination: #3621, this PR, and #3584 were all conflicting on the same seam, so I rebased the three in a staircase — **#3621 → 16, this → 17, #3584 → 18**. Merging them in any order works without a second renumber; an out-of-order merge just skips a number, which is harmless for a cache-busting counter. Happy to renumber to contiguous values if you'd prefer them to match your merge order.

paul-0320:
Rebased onto current master (`1ec6a6e8`, v0.45.9.0) — no conflicts. `KNOBS_HASH_VERSION` claim unchanged at 17 (`kof=`; 16 = #3621, 18 = #3584). Verify 38/38; search-mode + knobs-hash adjacent suites 183/183. Review state unchanged since the explicit approval upthread — merge-ready whenever a wave picks this lane up.

time-attack:
Status update: reviewed in the 2026-08-18 full-backlog triage — still wanted, but not yet mergeable: Reviewer-approved operator knob for CJK/unstemmed OR-flood, still absent on master, but needs another mechanical KNOBS_HASH_VERSION renumber + rebase. If you update it, it will be re-reviewed promptly; alternatively the fix may be adopted with credit in an upcoming fix wave. Thanks for the contribution and the patience.

paul-0320:
Rebased onto current master (`649ffe5f`, v0.46.21.0) — this had gone conflicting again once 16 (`det=`), 17 (degradation stamp) and 18 (`acm=`, #1863) landed on the single-writer constant.

Applied:
- `KNOBS_HASH_VERSION` **18 → 19**; `kof=` appended after `det=`, every upstream part kept (`fts=`, `det=`, `acm=` all present). Bump-rationale comment and the four pin tests updated to 19 (both `describe` blocks in `search-mode.test.ts` kept — upstream's v0.46.15 knobs block and this PR's `keywordOrFallback` block).
- The dashboard entry follows upstream's split of `KNOB_DESCRIPTIONS` / `buildModesReport` into `src/core/search/modes-report.ts` (same two lines, new home); `formatModesText` in `commands/search.ts` unchanged from before.
- `hybrid.ts` sits at its module-size ceiling (`scripts/check-module-size.sh`, in `verify`), so the wiring is now a one-line change plus a tightened comment — net zero lines, no TSV edit. Rationale lives on `ModeBundle.keywordOrFallback` as before.

Verification on the rebased head: `bun run verify` 54/54 · `bunx tsc --noEmit` clean · `search-mode` + `test/search/` + `cross-modal-phase1` + `alias-resolved-boost` 484/484 (incl. the two PGLite e2e cases in `title-retrieval-arm`).

Staircase (siblings rebased the same way today): **this → 19, #3584 → 20, #3621 → 21** — merge in any order, an out-of-order merge just skips a number. Default behavior still unchanged; the review approval upthread stands unaltered in substance.

time-attack:
Re-trialed against the current integration branch (post v0.46.23-25 waves): Merge vs wave@2cdf825e conflicts in 5 files: src/core/search/mode.ts + 4 tests (search-mode, cross-modal-phase1, search-alias-resolved-boost, search/knobs-hash-reranker). Not superseded (knob absent on branch). No red flags. — a rebase along those lines puts this back in the active merge queue; re-trials run nightly, so a fresh push gets picked up automatically. Thank you!

paul-0320:
Rebased onto current master (`4e4677b1b`, post-v0.46.28.0). Conflicts were the five files your re-trial named — all the same shape (the version-pin seam). Applied:
- `KNOBS_HASH_VERSION` 19 → **24** (19 `ack=` through 23 `xp=` were claimed upstream while this was open); `kof=` stays append-only after `xp=`. The four version-pin tests updated accordingly.
- `hybrid.ts` net-zero preserved (the file sits exactly at its 2845-line ratchet ceiling).

Verification on the rebased head: `bun run verify` 54/54 · `bunx tsc --noEmit` clean · the five patch-adjacent test files 178/178.

time-attack:
Landed in v0.46.29.0 (release PR #4567, merged as 492b5528) — your change shipped with Co-authored-by credit on the landing commit and is in the release notes. Thank you for the contribution! Closing since the content is now on master.

Issue #3614, opened 2026-07-30
fix(progress): the shared progress reporter silently disables Ctrl-C for bulk commands


time-attack:
Any command that creates a progress reporter has Ctrl-C silently disabled, unless that command also installs its own exiting SIGINT handler. The first Ctrl-C is swallowed; the process keeps running.

## Cause

`src/core/progress.ts:65-84` installs a SIGINT listener to abort live reporters:

```ts
process.once('SIGINT', () => onSignal('SIGINT'));
```

`onSignal` aborts the reporters and returns. It never exits, and it never restores the default disposition. In Node/Bun, installing *any* SIGINT listener suppresses the default terminate behavior — so this handler alone converts Ctrl-C into a no-op. The inline comment says `once()` is used "so we don't block user handlers or double-fire", but `once()` does not avoid suppressing the default disposition; it only avoids firing twice.

`src/core/process-cleanup.ts` does not cover this: it deliberately handles SIGTERM/SIGHUP/SIGPIPE (each with an explicit `process.exit`) and explicitly does **not** listen to SIGINT (see the comment at `process-cleanup.ts:118-124`).

So the affected set is: commands that start a progress reporter **and** do not install their own exiting SIGINT handler. `sync`, `watch`, `autopilot`, `serve`, `recall`, `agent`, and the minion worker each install their own and are unaffected. The bulk commands that rely solely on the shared reporter are affected — `embed --stale` (a long backfill, the worst case for an ignored Ctrl-C), `doctor`, `extract`, `import`, `export`, `orphans`, `lint`, `check-backlinks`, `repair-jsonb`, `integrity auto`.

## Repro

```ts
// _sig_target.ts
if (process.argv[2] === 'progress') {
  const { createProgress } = await import('./src/core/progress.ts');
  createProgress({ mode: 'human' }).start('repro.phase', 100);
}
process.stderr.write('READY\n');
setInterval(() => {}, 1000);
```

Spawn it from a driver that gives the child default signal dispositions (a plain `cmd &` in a non-interactive shell sets SIGINT to `SIG_IGN` for the child and will mask this), wait for `READY`, then send SIGINT:

```
control    -> EXITED code=130      # no reporter: dies on the first SIGINT
progress   -> STILL ALIVE after 3s # reporter installed: Ctrl-C swallowed
```

## Suggested fix

[...]

PR #3618, opened 2026-07-30
v0.42.68.2 fix(serve,import): stop the SIGINT orphan; quote imported frontmatter

time-attack:
Post-hoc fixes for defects in three PRs that were merged **without review** yesterday. That was my mistake; this is the cleanup. Each defect below was reproduced by execution before it was fixed, and each fix ships a check that fails without it.

## 1. `serve --http` ignored the first Ctrl-C and left an orphan holding the write lock

From #3599. The port was released, so the server looked stopped — but the process stayed alive, never disconnected from the brain, and kept the PGLite write lock. The next command needing a write could block on a lock held by a process the operator believed was closed. With a dashboard tab open, shutdown never completed at all.

Two causes, both fixed:

- `close()` stops the listener and then waits for every open connection to drain, and nothing severed them. Now live sockets are tracked and destroyed once `close()` has stopped accepting. Worth knowing: **Bun 1.3.x ships `closeAllConnections()` and `closeIdleConnections()` as no-op stubs**, so manual tracking is the only portable teardown — the obvious fix silently does nothing.
- The `serve` path skips both `finishCliTeardown` and the force-exit seam by design, so nothing disconnected the engine or exited. Now it disconnects (checkpoint / pool drain, so the store isn't left needing recovery) raced against the existing `CLEANUP_DEADLINE_MS`, then exits.

Measured with a spawned process, master vs this branch:

| | master as shipped | this branch |
|---|---|---|
| SIGINT, idle | **hung, orphan** | exit 0 in 23ms |
| SIGINT, connection open | **hung, orphan** | exit 0 in 57ms |
| SIGTERM, idle | exit 143, 85ms | exit 143, 39ms |
| SIGTERM, connection open | exit 143, 8ms | exit 143, 7ms |

`SIGTERM` was never affected — `process-cleanup.ts` bounds it — so supervisors and container runtimes saw nothing wrong. Which is why this went unnoticed.

Two tests added: socket-severing, and settle-exactly-once for the first-signal-wins property. Both are change-detectors and I want to be explicit that they are: a fake whose `close()` always succeeds on the next microtask **cannot** observe the real hang, which is exactly why the original green CI missed it. The durable regression guard would be a spawned-process e2e; noted as a follow-up rather than smuggled in here.

## 2. Conversation import quoted one frontmatter value and not its neighbour

From #3600. One interpolated value was emitted unquoted while the value beside it was quoted — under a comment asserting that every value was quoted. A source value carrying unusual characters could reshape the frontmatter of the page being written, and in the worst case the page lost its provenance fields with no error reported.

Fixed by quoting it consistently; an absent date still emits a real YAML null rather than text. The regression test is parameterized over both failure shapes and discriminates properly: **12 pass with the fix, 3 fail without it.**

Per the repo's responsible-disclosure rule I've kept the specifics out of this description and the CHANGELOG. Operator guidance (re-import and re-sync, and how to spot an affected page) is in the CHANGELOG's "To take advantage" block.

## 3. Reverting my own bad fix from #3610

When #3598 and #3599 landed together they broke master's typecheck — each had faked private `Pick<>` aliases, a semantic merge failure that per-PR CI cannot catch. I cleared it the wrong way: exported three private types from `serve-http.ts` and cast the test fakes through `as unknown as T`.

That double cast disables type checking entirely, so the annotations verified nothing, and one of the three exported types (`SignalSource`) had **no consumers at all**. I widened a production API and deleted the safety it was supposed to provide, to get green faster.

The real problem was the contract: typing the parameters as `Pick<HttpServer, …>` / `Pick<Response, …>` demanded `this`-returning Node/Express fidelity the functions never use — every return value is discarded. Narrowed to minimal structural contracts returning `unknown`, kept private, all four casts deleted, and the now-unused `HttpServer` import removed. The real types are still validated where it matters: the call site passes an actual `http.Server`.

Proof the constraint binds now — each of these fails typecheck, and each compiled clean under #3610:

- dropping `write()` from the SSE fake → `TS2345`
- `listening` as a string → `TS2345` on both call sites
- removing `close()`'s callback parameter → `TS2345`

## Gates

`bun run typecheck` exit 0, 0 errors. `bun run check:wasm` exit 0. 93 tests pass across the touched suites (serve-http-lifecycle, admin-sse-handshake, admin-sse-eventsource, envelope-to-gbrain, files, process-cleanup, serve-stdio-lifecycle, build-llms).

Honest caveat on `bun run verify`: it was **33/33 green** on this source, but a later run on the committed state showed 4 failures — three of them `TIMED OUT after 120s` (`typecheck`, `check:privacy`, `check:test-isolation`) and the fourth (`check:wasm`) passing when run alone. Machine load average was 61–77 from an unrelated parallel workload at the time. I'm calling that environmental rather than claiming a clean run I didn't get; CI is the arbiter.

## Not in this PR

#3614 — a separate pre-existing root cause found while auditing this. The shared progress reporter installs a SIGINT listener that aborts reporters and never exits, which suppresses the default disposition and so silently disables Ctrl-C for bulk commands that install no handler of their own (`embed --stale`, `doctor`, `extract`, and others). It is *not* the cause of the `serve` bug — `serve` has its own handler — but it is why the pre-`#3599` baseline there also failed to exit on the first Ctrl-C, which made this regression look narrower than it was. Repro and suggested fix are in the issue.


time-attack:
## Update — adversarial review of this PR found four defects; all four fixed

I sent this PR through the same gauntlet I've been applying to everyone else's: two independent reviewers, told the author was me and instructed not to soften anything, each with a different lens and told to default to "refuted" when uncertain.

**One returned `refuted: false`** and independently reproduced the fix (SIGINT exit 0 in ~9–17ms, SIGTERM 143, no orphan), and killed the concern I'd flagged as most likely to be real: I worried that quoting `date` would break a consumer expecting a date type. It's the opposite — the *unquoted* form parsed to a JS `Date`, and `deriveDateContext` reads only `typeof === 'string'`, so it was silently skipping the field. Quoting makes that path work for the first time.

**The other returned `refuted: true` with four findings. All four were real. I verified each myself before fixing.**

### 1. The remediation block gave a false all-clear — this was the blocking one

I told operators to confirm the orphan was gone with `gbrain doctor`. Run against a live holder:

```
pgrep -fl "gbrain serve"  ->  (FOUND NOTHING)
gbrain doctor             ->  exit 0
                              Overall health score: 95/100. All checks OK (some warnings).
```

`doctor` has no check for the PGLite filesystem write lock (`stale_locks` inspects DB rows with `ttl_expires_at`, a different thing), and `pgrep -f "gbrain serve"` misses a server started from a source checkout. So anyone following my block verbatim would see green and conclude they were fine.

This is the same defect class I flagged in #3519 in this same review round — remediation text naming a command that does not do what the text claims. Writing it myself two days later is the more useful data point.

Replaced with the brain's own lock file, which records the holding pid and command line. Also verified the stale case: once the recorded pid is dead, gbrain reclaims the lock automatically (`capture` exit 0), so the block now says the leftover file is harmless rather than sending people chasing it. Every command in the block was executed against a live holder before being written down.

### 2. Path traversal via the same untrusted value this PR hardens

I quoted `created_at` in the frontmatter under a comment calling the envelope third-party and hostile — then eleven lines above, the same raw value prefixes the filename, and `join(outDir, name)` resolves `../`. `convId` went through `slug()`; `date` did not.

```
created_at: "../victim/p"
  ->  wrote 1 markdown page(s) to .../outdir   (exit 0)
      outdir:  (empty)
      victim:  -c-evil.md          <- outside outDir
```

I reasoned about the attacker-controlled field and stopped at the frontmatter. `date` now takes the same `slug()` the id already used. Regression test added; it fails without the fix.

### 3. `TrackedSocket` was decorative — the exact defect this PR exists to remove

The PR body says `as unknown as T` "disables type checking entirely." I then introduced a functionally identical unchecked assertion in *production* code. Proof: adding a member no socket could ever have still typechecked clean.

My first attempt to fix it — narrowing the listener signature to `on(event: 'connection', listener: (socket: TrackedSocket) => void)` — **also failed**, because method parameters are bivariant in TypeScript, so annotating the callback matches my own declaration whatever a real socket does. Now bound to `net.Socket` via `Pick`, which is right here specifically because no typechecked double has to satisfy it (fakes arrive through `emit`, which is untyped). A key that isn't a real Socket method now fails to compile.

### 4. My settle-exactly-once test was theater

It passed with the implementation's `settled` guard deleted. Two reasons, both mine: a native Promise settles once regardless, so asserting resolve-count could never test anything; and my own `if (server.listenerCount(event) > 0)` guard meant the second signal was never delivered, because the first emit had already consumed the `once` listener.

What the guard actually protects is `finish()`'s side effects — deregistering the shared-cleanup entry. The test now drives the real double-finish path (close callback reports failure while the server also emits `close`, so `onError` and `onClose` both reach `finish`) and asserts the entry is deregistered once. Verified: with the guard removed it fails, `2` instead of `1`.

I filed #3665 about tests that pass on master roughly an hour before writing this one. Same trap.

### 5. Also addressed: the load-bearing half had no test

The reviewer showed that reverting `serve-http.ts` entirely and keeping only the `serve.ts` teardown still fixes all four signal scenarios — so the teardown is what actually stops the orphan, and it had no test and was unreachable from one. Extracted as `finishHttpServe` using the `exit`/`log` seams `ServeOptions` already provided, with tests for disconnect-before-exit ordering, a throwing disconnect, and a disconnect outliving the deadline.

**Still not covered:** the single line wiring `finishHttpServe` into `runServe`. The teardown logic is tested; the call site is verified only by the spawned-process matrix, which I ran by hand. A spawned-process e2e is the durable guard and remains a follow-up.

### Gates after the fixes

`bun run verify` **33/33, all checks green, exit 0** — which also resolves the caveat in my original description: those 4 earlier failures were load, not code (three were literally `TIMED OUT after 120s`). `bun run typecheck` exit 0, 0 errors. 55 tests pass across the touched suites.

Signal matrix re-run after the refactor, since I'd changed the load-bearing path:

| | before | after refactor |
|---|---|---|
| SIGINT idle | exit 0, 23ms | exit 0, 15ms |
| SIGINT + connection | exit 0, 57ms | exit 0, 11ms |
| SIGTERM idle | 143, 39ms | 143, 10ms |
| SIGTERM + connection | 143, 7ms | 143, 11ms |

Note this branch and #3633 both claim `v0.42.68.0`; whichever lands second will need a re-bump.

time-attack:

PR #3621, opened 2026-07-30
feat(search): wire search.autocut_min_keep — the documented autocut floor config is silently dead

paul-0320:
## Problem — `search.autocut_min_keep` is a silently dead config key

`autocut.ts` ships a config resolver that reads `search.autocut_min_keep` ([`autocutFromConfig()`](https://github.com/garrytan/gbrain/blob/c6dc0adf/src/core/search/autocut.ts#L67)), and `applyAutocut` implements the floor correctly (the cliff scan starts at `minKeep - 1`; the pure-fn tests pin it). But the resolver has **zero call sites** — the live resolution plane for autocut is the mode bundle (`resolvedMode.autocut` / `autocut_jump`), and the search path hardcodes the floor:

```ts
// hybrid.ts — the only applyAutocut call site
{ enabled: true, jumpRatio: resolvedMode.autocut_jump, minKeep: 1 },
```

So `gbrain config set search.autocut_min_keep 6` stores fine, shows nowhere, and changes nothing — the operator gets no error and no effect. Its two sibling knobs (`search.autocut`, `search.autocut_jump`) resolve normally, which makes the gap easy to misread as "my floor value is wrong" rather than "the key is unwired".

Why an operator reaches for this knob: autocut's `jumpRatio` was calibrated against `zerank-2` score curves. On other rerankers (e.g. a local Qwen3-Reranker via the llama-server recipe) the score curve can decay smoothly with no dramatic cliff, and "the largest gap" degenerates into noise — we measured a cut decided by a **1.5% difference** between two nearly identical gaps, trimming a 25-deep list to 2 and dropping a correct answer that sat at rank 3. A floor is the surgical remedy when the consumer is an LLM that reads the whole returned list ("deep but visible" beats "trimmed away"), without giving up autocut entirely.

## What this does

Wires the knob through the live plane, following the `autocut_jump` pattern end to end:

- `ModeBundle.autocut_min_keep` — **default `1` in every bundle** (the previous hardcoded behavior; nothing changes unless an operator opts in).
- Per-key config override **`search.autocut_min_keep`** (integer ≥ 1; junk/out-of-range falls through to the bundle — same validation the stranded resolver in `autocut.ts` uses, so semantics stay consistent), registered in `SEARCH_MODE_CONFIG_KEYS` and shown in the `gbrain search modes` dashboard.
- `hybrid.ts` passes `minKeep: resolvedMode.autocut_min_keep` at the single `applyAutocut` call site (replacing the hardcoded `1`).
- Per-call `SearchPerCallOpts.autocut_min_keep` participates in the same pick chain (per-call > config > bundle), mirroring `autocut_jump`.
- **Cache safety**: `acm=` joins `knobsHash` (append-only) and `KNOBS_HASH_VERSION` bumps 14→15 — the floor changes how many rows survive the cut, so a minKeep=1 write (trimmed to the cliff) must never serve a raised-floor lookup. Same contamination class as `ac=`/`acj=`. Heads-up: #3584 and #3617 (both open) also bump 14→15; whichever lands later rebases to the next number — happy to do that on this side.

`applyAutocut` itself is untouched — the floor logic and its unit tests already existed; this PR only connects the config surface that was documented but never wired.

## Measurements

**Eval replay characterization** (production pipeline: hybrid + reranker + autocut forced on per-call, 62 Korean-heavy queries, floor 1 → floor 6, per-question ABAB interleave so corpus drift lands symmetrically; cross-condition cache contamination is structurally impossible — the two conditions hash to different `acm=` keys, which is the point of the bump):

| metric | value |
|---|---|
| Jaccard@10 | 0.594 — entirely a list-length effect: autocut is trim-only, so the kept prefix is identical by construction and the floor only appends more of the ranked tail |
| top-1 stability | **62/62** (structural: the top item always survives a trim) |
| latency Δ | **≈ 0** (6.20 s → 6.14 s avg — autocut runs after reranking, so the floor changes no upstream work) |

Behavior on this corpus (mean returned-list length 7.5 → 20.3 of limit 25): with a floor of 6, the cliff scan usually finds no qualifying gap past the floor on this reranker's smoothly-decaying curves, so it declines to cut at all rather than cutting at rank 6. That is the intended never-empty semantics of `minKeep` (the floor suppresses cut points above it; it does not force a cut at the floor), but operators should know the compression trade-off is coarse on smooth curves. Answers found in top-25 went 50 → 51 with zero losses (structural: a trim-only stage can only hide answers, so raising the floor can only recover them) — the recovered answer was one the reranker had scored at 1.4% of the top score, i.e. exactly the "deep but present" case the floor exists for. One honest caveat on that recovery count: these eval queries are paraphrase-style conversational questions authored to avoid the answer documents' vocabulary, which over-represents reranker-underscored answers relative to keyword-style production queries — read the found@k delta as the favorable-case bound, not a typical yield.

## Verification

```
bun test test/search-mode.test.ts                        # knob resolution, config parsing, acm= hash participation
bun test test/search/autocut-integration.serial.test.ts  # end-to-end through hybridSearch (PGLite): default floor unchanged, config floor moves the cut, junk ignored
bun run verify                                           # 33/33
bun run test:full                                        # 7,505 pass / 7 fail, all env-dependent (see note)
```

Note on the 7 failures: they are environment-dependent on this dev box (a live `~/.gbrain` config leaks into env-sensitive tests — e.g. the hybrid-meta tests that delete `OPENAI_API_KEY` still resolve a local embedding provider, and the repo-durability/worker-registry tests touch the real gbrain home). Every failure inspected reproduces identically on the pristine base commit `c6dc0adf`; none are in patch-adjacent suites (search-mode, knobs-hash, autocut unit/eval/integration, query-cache — all green). CI in a clean environment should be authoritative.

The new hybrid-level cases drive the real pipeline (PGLite keyword arm → dedup → reranker test seam → autocut) with a two-cliff score curve, so the knob's effect is observable end-to-end: default floor cuts at the biggest cliff (pre-fix behavior, unchanged), config floor 3 moves the cut to the second cliff, a floor above the pool declines to cut, and junk config values fall through to the default. Existing `KNOBS_HASH_VERSION` pins updated (4 files).


time-attack:

Technical core verified sound and earned it. (1) Dead-key claim TRUE: autocutFromConfig has zero src call sites (only its own unit test at test/search/autocut.test.ts), and the single applyAutocut caller (src/core/search/hybrid.ts:1679) hardcoded minKeep: 1 — `gbrain config set search.autocut_min_keep 6` stored and did nothing pre-PR. (2) Discrimination test passed on BOTH planes: stubbing minKeep: 1 back into hybrid.ts fails exactly the 2 floor-wiring integration tests (default-behavior pins correctly still pass); removing the acm= part from knobsHash fails the floors-1-vs-6 hash test. (3) Floor bounds verified structurally in applyAutocut: trim-only (kept ⊆ results), Math.max(1, minKeep), cut index ≥ minKeep, floor > pool → empty scan loop → no cut, and returnPool.slice(offset, offset+limit) runs AFTER autocut — can never return more than requested nor fewer than 1. (4) KNOBS_HASH_VERSION 14→15 is correct: master is still at 14, competing bumps #3584/#3617 are still OPEN; all 5 pin sites accounted for (4 updated, the 5th uses toBeGreaterThanOrEqual(14)); acm= is append-only; bare and cached resolvers read the same config plane so cache write/lookup keys agree. (5) Executed the operator surfaces live in an isolated GBRAIN_HOME: config set works, `search modes` shows the knob description, --json config_keys includes it, `search modes --reset` clears it. (6) typecheck clean; 183/183 tests pass across all 6 touched files. No engine/SQL/JSONB/source-isolation/trust surface touched; no privacy violations. What blocks merge is purely release mechanics: title lacks the version prefix (IRON RULE), VERSION trio is 0.42.67.0 (= base, < master's 0.42.68.1, no CHANGELOG entry), and docs/architecture/KEY_FILES.md:125 still lists only search.autocut + search.autocut_jump as config keys. All three are maintainer-mechanical; a takeover-ship via /ship would clear them in one pass. Conditional hazard: if #3584 or #3617 lands first and claims v=15, this PR must rebump to 16 (author already offered).

### Discrimination test

Stubbed the old behavior back in two ways and re-ran the PR's own tests. (a) Reverted hybrid.ts:1679 to the pre-PR hardcoded `{ enabled: true, jumpRatio: resolvedMode.autocut_jump, minKeep: 1 }` → test/search/autocut-integration.serial.test.ts went 9 pass / 2 fail, failing exactly 'config floor 3 cuts at the SECOND cliff' and 'floor above the candidate pool returns full set' while the default-floor and junk-config pins correctly stayed green (default behavior is unchanged by design). (b) Removed the `acm=` line from knobsHash in mode.ts → test/search-mode.test.ts went 74 pass / 1 fail, failing 'knobsHash includes acm= — floors 1 vs 6 differ'. Both reverts restored afterward; the tests genuinely pin the wiring and the cache-key participation.

**Mergeability at review time:** {"mergeStateStatus":"CLEAN","mergeable":"MERGEABLE"} (first poll returned UNKNOWN; re-polled to CLEAN/MERGEABLE)

### Blockers

**1.** PR title violates the version-first IRON RULE: 'feat(search): wire search.autocut_min_keep ...' has no vMAJOR.MINOR.PATCH.MICRO prefix. Fix: gh pr edit 3621 --title 'vX.Y.Z.W feat(search): ...' with the ship-allocated version.

**2.** VERSION trio not advanced: branch VERSION/package.json/CHANGELOG all read 0.42.67.0 (inherited from base c6dc0adf) while master is 0.42.68.1 — not strictly greater, and no CHANGELOG entry describes the change. Equal-or-less fails the gate.

**3.** Doc drift: docs/architecture/KEY_FILES.md:125 autocut entry says 'Config: search.autocut, search.autocut_jump' — now stale; must add search.autocut_min_keep (reference docs describe CURRENT behavior per repo rule).

**4.** Conditional: KNOBS_HASH_VERSION 15 is contested by open PRs #3584 and #3617 (both also bump 14→15). Fine if this lands first (master verified still at 14); whichever lands second must rebump to 16 and update the 4 pin sites.

### Not verified

- PR body's eval-replay measurements (Jaccard@10 0.594, 62/62 top-1 stability, latency ≈0, found@25 50→51) — require the author's private Korean-heavy corpus + local reranker; not reproducible here. The structural claims backing them (trim-only ⇒ top-1 always survives; floor only appends ranked tail) were verified by reading applyAutocut and hold by construction.
- Full suite (bun run verify / test:full) not run per shared-machine footprint restrictions; coverage here = typecheck + the 6 touched test files + PR CI (green incl. serial-tests, JSONB parity).
- Real-Postgres path not exercised (no SQL changes; PGLite integration test is the relevant seam).
- Merge-commit state against current master not built locally (GitHub reports MERGEABLE/CLEAN; master's mode.ts confirmed textually unchanged at KNOBS_HASH_VERSION=14).

- `gh pr view 3621 --json ... (metadata + mergeable, re-polled)`
- `gh pr diff 3621 > /tmp/pr3621.diff (read every hunk, 423 lines)`
- `gh pr view 3584 / 3617 --json state (both OPEN — v=15 not yet claimed)`
- `git show origin/master:VERSION (0.42.68.1) + grep KNOBS_HASH_VERSION on master mode.ts (= 14)`
- `gh pr checkout 3621 into isolated worktree`
- `grep -rn autocutFromConfig src/ test/ (zero src call sites — dead-key claim confirmed)`
- `grep -rn applyAutocut src/ (single caller in hybrid.ts — sibling-caller check)`
- `Read src/core/search/autocut.ts + hybrid.ts call-site context (floor bounds + post-autocut limit slice)`
- `bun run typecheck > file (EXIT=0)`
- `bun test <6 touched files> > file (183 pass / 0 fail)`
- `DISCRIMINATION: stub minKeep:1 into hybrid.ts, bun test autocut-integration (EXIT=1, 2 fail — the wiring tests); restore`
- `DISCRIMINATION: remove acm= from knobsHash, bun test search-mode (EXIT=1, 1 fail — the hash test); restore`
- `grep KNOBS_HASH_VERSION repo-wide (5th pin site uses >=14, no stale toBe(14))`
- `inspected both perCall construction sites in hybrid.ts (bare + cached wrapper — same config plane, keys agree)`
- `GBRAIN_HOME=/tmp/pr3621-home: gbrain init --pglite --no-embedding; gbrain config set search.autocut_min_keep 6 (works); gbrain search modes (+--json: knob description + config_keys present); gbrain search modes --reset (clears the new key)`
- `grep autocut docs/ (search-modes.md silent on autocut — no drift; KEY_FILES.md:125 stale)`
- `gh pr checks 3621 (all green incl. serial-tests + JSONB parity)`
- `cat VERSION / package.json version / CHANGELOG head on branch (all 0.42.67.0)`

</details>

---

time-attack:

Bug still exists on current master: src/core/search/autocut.ts:67-83 ships autocutFromConfig() reading search.autocut_min_keep (unit-tested in test/search/autocut.test.ts:201) but has zero production call sites, and hybrid.ts:1680 hardcodes minKeep: 1 at the only applyAutocut call site — so the config key stores silently and does nothing (wires documented-but-dead behavior = bug-fix lane). The v0.42.69.0 wave did NOT fix this; nothing on master touches it and siblings #3617/#3584 (also open) touch different knobs. I test-merged the PR onto origin/master in a throwaway worktree: after resolving the KNOBS_HASH_VERSION collision (renumber 15→16, keep both fts= and acm= parts), typecheck is clean, all 187 patch-adjacent tests pass, and the 2 new integration tests fail when the hardcoded minKeep: 1 is restored — the fix is real and root-cause. NEEDS_FIX only because the required rebase is semantic, not context drift: the author must renumber the hash bump and merge the comment hunks (they pre-committed to this in the PR body).

**Conflict state vs current master:** 5 files conflict: src/core/search/mode.ts (2 hunks) + 4 test files with KNOBS_HASH_VERSION pins. Semantic collision: merged #3677 (fts= fold, in the wave) claimed v=15; this PR also claims 14→15 for acm=. Required resolution is mechanical and verified working: bump to KNOBS_HASH_VERSION = 16, keep BOTH the fts= and acm= parts in the knobsHash array (acm= appended after fts= at mode.ts ~line 950), merge both bump-rationale comments, and update the 4 test pins (test/cross-modal-phase1.test.ts, test/search-alias-resolved-boost.test.ts, test/search-mode.test.ts x2, test/search/knobs-hash-reranker.test.ts) to expect 16. src/core/search/hybrid.ts and src/commands/search.ts auto-merge cleanly. The fix fully applies post-v0.42.69.0. Note: even a botched merge keeping v=15 would not contaminate the cache (the appended acm= part changes every hash string, forcing cold misses), but the renumber is the repo's discipline and the test pins enforce it.

**Risk if merged as-is:** After the rebase: one-time global query-cache cold-miss from the KNOBS_HASH_VERSION bump (documented, intended pattern — master just did the same in #3677). Behavior default is unchanged (autocut_min_keep: 1 in all three bundles = the previous hardcoded value; only opt-in config changes anything). KNOB_DESCRIPTIONS is Record<keyof ModeBundle, string> so typecheck forces the dashboard entry (present). No engine files, no jsonb, no source-scoping, no pricing tables touched. Coordination: #3617 and #3584 (both open, same author) also plan hash bumps — whichever lands next must renumber to 17.

paul-0320:
Rebased onto current master (`2e554500`) with exactly the resolution you specified — thank you for test-merging it rather than just describing the conflict.

Applied:
- `KNOBS_HASH_VERSION` renumbered **15 → 16**; the merged #3677 keeps 15.
- Both parts kept in the `knobsHash` array, `acm=` appended after `fts=`.
- Both bump-rationale comments merged, with the 15→16 note stating why the number moved.
- The four version pins updated to 16 (`test/cross-modal-phase1.test.ts`, `test/search-alias-resolved-boost.test.ts`, `test/search-mode.test.ts` ×2, `test/search/knobs-hash-reranker.test.ts`), and their titles amended so they name both bumps instead of only the FTS one.

`src/core/search/hybrid.ts` and `src/commands/search.ts` auto-merged, as you found.

Verification on the rebased head:
- `bun run verify` 34/34 · `bunx tsc --noEmit` clean
- Patch-adjacent suites 469/469 (`test/search-mode`, `test/search/`, `test/cross-modal-phase1`, `test/search-alias-resolved-boost`)
- Discriminating check re-run: restoring the hardcoded `minKeep: 1` at the `applyAutocut` call site fails exactly the 2 new integration tests (9 pass / 2 fail); with the wiring in place, 11/11.

One coordination note. #3617 and #3584 were both conflicting on the same seam, so rather than leave them to collide with this one I rebased all three in a staircase: **#3621 → 16, #3617 → 17, #3584 → 18**. Whichever of the three you take first, the other two are already numbered above it and will not need a second renumber. If you merge them out of order the intermediate number is simply skipped, which is harmless for a cache-busting counter — but say the word if you'd rather they were contiguous and I'll renumber to match your merge order.

time-attack:
Status update: reviewed in the 2026-08-18 full-backlog triage — still wanted, but not yet mergeable: Wires a documented-but-dead config key (verified still dead on master); reviewer-specified resolution applied, but conflicts again on KNOBS_HASH_VERSION. If you update it, it will be re-reviewed promptly; alternatively the fix may be adopted with credit in an upcoming fix wave. Thanks for the contribution and the patience.

paul-0320:
Rebased onto current master (`649ffe5f`, v0.46.21.0). This one needed more than a renumber — v0.46.15 (#1863) landed `autocut_min_top` through the very same seams:

- **`KNOBS_HASH_VERSION` 18 → 21** (16 `det=`, 17 degradation stamp, 18 `acm=`; 19/20 are the siblings #3617/#3584). Comment + four pin tests updated.
- **Hash part renamed `acm=` → `amk=`**: upstream's `acm=` is now the weak-top floor (`autocut_min_top`), so this PR's minKeep part takes `amk=`, appended after `det=`; both parts present, all upstream parts kept. Same rename for the `loadOverridesFromConfig` local (`amk`) next to upstream's `acm`.
- **Call site**: `applyAutocut(..., { enabled, jumpRatio, minKeep: resolvedMode.autocut_min_keep, minTopScore: resolvedMode.autocut_min_top })` — both floors thread from the resolved mode; the two are orthogonal (min_top gates *whether* autocut runs, min_keep floors *how far* it may trim). `hybrid.ts` sits at its module-size ceiling, so the change is net zero lines (comment tightened, no TSV edit).
- `ModeBundle` / bundles / overrides / per-call / `pick` / `SEARCH_MODE_CONFIG_KEYS`: min_keep lands next to min_top in every list. Dashboard description follows upstream's move of `KNOB_DESCRIPTIONS` to `modes-report.ts` (still type-forced by `Record<keyof ModeBundle, …>`). KEY_FILES autocut entry names `search.autocut_min_keep` alongside `search.autocut_min_top`.
- The dead-key claim still holds on master: `autocutFromConfig()` reads `autocut_min_keep` (and now min_top) but has no production caller, and the `applyAutocut` call site hardcodes `minKeep: 1` — this PR is what wires it.

Verification on the rebased head: `bun run verify` 54/54 · `bunx tsc --noEmit` clean · `search-mode` + `test/search/` + `cross-modal-phase1` + `alias-resolved-boost` 489/489. Discrimination re-run: restoring the hardcoded `minKeep: 1` fails exactly the 2 floor-wiring integration tests (9/2); dropping the `amk=` part fails the floors-1-vs-6 hash test (79/1) — both restored, 11/11 and 80/80.

Staircase: #3617 → 19, #3584 → 20, **this → 21**. Title version prefix / VERSION trio remain the maintainer's at land time, as before.

time-attack:

Issue #3594, opened 2026-07-29
Dry-run is unsafe in at least 3 commands: destructive call precedes the guard, or the default contradicts the docs


time-attack:
Three independent instances of the same defect class surfaced in two days. Filing the class, because fixing them one at a time will not stop the fourth.

| Where | What the "preview" actually did |
|---|---|
| `jobs prune --dry-run` (#2712, fixed in #3525) | Parsed only `--older-than`; `--dry-run` was silently dropped and the `DELETE` ran |
| `unify-types` (#1575, fixed in #3574) | `apply: data.apply ?? true` while the handler's own docstring said *"Default false (dry-run)"* |
| `gbrain sync --dry-run` (#3543, open) | Calls `engine.deletePage()` **before** the `if (opts.dryRun)` early-return, and runs `pullRepo` — so it deletes indexed pages and mutates the working tree |

The third is the worst: `sync` is the most-run command, and `--dry-run` is precisely what a cautious user reaches for before a risky operation.

**The shared shape** is one of two mistakes:

1. The destructive call sits *above* the dry-run early-return, so the guard is real but arrives too late.
2. The default is `?? true` / action-by-default while the docs, the flag name, or the handler comment promise the opposite.

Neither is caught by tests, because a dry-run test that asserts the *output* still passes while the side effect happens underneath. #2712's test only caught it because it asserted the row still existed afterwards.

**Suggested guard — cheap and mechanical:**

- A lint or unit check that, for every handler accepting a `dry_run` / `dryRun` / `apply` parameter, the documented default matches the code default. #1575 would have failed this immediately.
- A convention that dry-run returns as early as possible — ideally before any engine handle is used — so ordering mistakes are structurally impossible rather than review-dependent.
- For the highest-risk commands (`sync`, `prune`, `unify-types`, `orphans`, `purge`, anything calling `deletePage`/`DELETE`), a test that runs the command with the dry-run flag against a seeded brain and asserts row counts are **byte-identical** before and after. That is the assertion that catches all three of these; asserting on printed output catches none of them.

Worth auditing the remaining commands that accept a dry-run flag rather than waiting for a fourth report — `gbrain orphans`, `sources purge`, `migrate`, and the backfill kinds all have preview modes I have not verified.

time-attack:
## Fourth instance, found in review of an open PR

Adding a data point that strengthens the case for fixing this as a class rather than one command at a time.

[...]

PR #3624, opened 2026-07-30
v0.42.78.0 fix(sync): prevent dry-run pull and out-of-strategy page deletion

mamedov:
## Summary

This is the narrow dry-run fix requested by @time-attack in #3543.

- Skip `pullRepo` when `gbrain sync --dry-run` is set.
- Return the incremental dry-run result before out-of-strategy page cleanup.
- Prove that a strategy-changing preview preserves the indexed page and sync bookmark.
- Prove that an ordinary sync still performs the same cleanup.
- Prove that a dry run with an `origin` remote never enters the pull phase.

The apply protocol from #3543 is intentionally not included.

## Why

On `master`, dry-run could mutate both the Git working tree and the brain. A pull could update the source checkout, and a modified tracked page that fell outside the selected strategy could be deleted before the dry-run early return.

## Verification

- `bun install --frozen-lockfile` — no changes.
- `bun run verify` — 33/33 checks passed.
- `bun test test/sync.test.ts` — 72 passed, 0 failed, 211 assertions.
- `bun test test/sync-pull-failed-anchor.serial.test.ts` — 3 passed, 0 failed, 22 assertions.
- `bun run test:slow` — 41 passed, 0 failed.
- `git diff --check origin/master...HEAD` — clean.

The broader local suites are not green on the current baseline:

- Unit: 14,813 passed, 3 failed, 18 skipped. The failures are in unchanged retrieval-reflex, doctor/config-isolation, and worker-registry tests; the doctor case passes with an empty `GBRAIN_HOME`.
- E2E: 1,091 passed, 14 failed across 10 unchanged files. The only sync-named failure reproduces byte-for-byte on exact `origin/master`: its lock test reaches embedding credential preflight before lock acquisition because it omits `--no-embed`; with a non-secret placeholder key it passes 7/7. The other failures were isolated to current-master test drift, local pgvector/Bun differences, or shared-environment contamination.

## Scope

The implementation changes only `src/commands/sync.ts` and `test/sync.test.ts`. Most production diff churn is moving the existing dry-run return above the cleanup loop. There are no schema, configuration, or apply-protocol changes.


time-attack:
Confirming this is the narrow split that was asked for in #3543, and that the apply protocol was correctly left out. Thank you for scoping it tightly.

**One ordering hazard worth knowing about before this lands.**

#3583 (`fix(sync): rename-sentinel operator exit + orphan self-heal`) touches this same file and, per an adversarial review of it, introduces a *new* instance of exactly the class this PR closes:

> `src/commands/sync.ts:2394` — `await sweepOrphanedRenameSentinels(...)` is inserted in the git-HEAD-equality `up_to_date` early return, which sits **above** the dry-run guard.

So the two PRs are in direct tension: this one removes destructive work from the dry-run path in `sync.ts`, and that one adds some back on a different early-return path. If #3583 merges after this, the class regresses in the same file, and the regression will not be obvious in the diff because it is a *placement* problem, not a logic one.

Suggested order: **land this first**, then require #3583 to be rebased on it and to demonstrate dry-run purity on the `up_to_date` path specifically. #3583 also has a confirmed data-loss blocker (its reconcile hard-deletes every active row whose `source_path` matches the rename's old path, not the intended one), so it is not close to landing regardless.

Tracking the pattern in #3594 — this is now the fifth instance across four commands, which is why I think the durable fix is a dry-run engine wrapper whose write methods throw, rather than continuing to fix guard placement one call site at a time. That would have made both this bug and #3583's uncommittable.

No changes requested here. Flagging the sequencing only.

time-attack:

The fix is real and root-cause (single pullRepo site gated on !opts.dryRun; dry-run return moved above the out-of-strategy cleanup at src/commands/sync.ts:2398), and its tests genuinely discriminate: stubbing master's sync.ts back in fails 2 of the 3 new tests (the third is a deliberate real-sync behavior pin). Typecheck + test/sync.test.ts (72/72) + test/sync-pull-failed-anchor.serial.test.ts (3/3) pass on the branch. Title, trio (0.42.78.0 > master 0.42.68.1), CHANGELOG conventions, and the 'gbrain upgrade' remediation command all check out; #3583's ordering hazard is defused on its side (its sweeps are now !opts.dryRun-guarded). The sole blocker: the PR is CONFLICTING/DIRTY against master — reproduced locally, conflicts are exactly VERSION, package.json, CHANGELOG.md (master shipped 0.42.68.1 after branch); sync.ts and tests auto-merge cleanly. Resolve the trio per CLAUDE.md's recovery procedure (keep 0.42.78.0, keep both changelog entries), push, then this is a MERGE.

### Discrimination test

Checked out pr3624, ran `git checkout origin/master -- src/commands/sync.ts` (old behavior stubbed back in, PR tests kept), then `bun test test/sync.test.ts`: 70 pass / 2 fail — 'strategy-changing dry-run preserves previously indexed out-of-strategy pages' and 'dry-run does not attempt git pull when origin exists' both fail on master's code and pass with the fix. The third new test ('strategy-changing real sync deletes previously indexed out-of-strategy pages') passes on both by design — it pins that non-dry-run cleanup behavior is unchanged. Tests discriminate; fix is load-bearing.

**Mergeability at review time:** {"mergeStateStatus":"DIRTY","mergeable":"CONFLICTING"}

### Blockers

**1.** mergeable=CONFLICTING / mergeStateStatus=DIRTY: local test-merge of origin/master confirms conflicts in VERSION, package.json, CHANGELOG.md (master moved to 0.42.68.1 after this PR branched). src/commands/sync.ts and test/sync.test.ts merge cleanly. Needs the standard trio resolution (keep 0.42.78.0 on all three, keep both changelog entries) and a re-push before it can merge.

### Not verified

- PR body's full-suite claims (bun run verify 33/33, unit 14813 pass/3 fail, e2e 1091 pass/14 fail) — skipped per execution-footprint rules; I ran typecheck + the two sync test files only
- Real-Postgres behavior — tests ran on the PGLite path; no engine SQL changed so risk is low
- bun.lock freshness — bun install in the worktree left the tree clean (only sync.ts modified during the discrimination step), consistent with the PR's frozen-lockfile claim, but I did not run --frozen-lockfile explicitly

- `gh pr view 3624 --json number,title,body,mergeable,mergeStateStatus,headRefName,baseRefName,state`
- `gh pr diff 3624 > /tmp/pr3624.diff`
- `gh pr view 3624 --json mergeable,mergeStateStatus (re-check: CONFLICTING/DIRTY)`
- `git fetch origin master; git show origin/master:VERSION (0.42.68.1)`
- `gh pr view 3583 --json state,mergedAt,title (OPEN, not merged)`
- `git fetch origin pull/3624/head:pr3624 && git checkout pr3624`
- `git merge --no-commit --no-ff origin/master (conflicts: CHANGELOG.md, VERSION, package.json only) then git merge --abort`
- `bun install (EXIT=0)`
- `bun run typecheck > /tmp/pr3624_typecheck.txt (EXIT=0)`
- `bun test test/sync.test.ts > /tmp/pr3624_tests.txt (EXIT=0, 72 pass/0 fail)`
- `git checkout origin/master -- src/commands/sync.ts && bun test test/sync.test.ts > /tmp/pr3624_discrim.txt (EXIT=1, 70 pass/2 fail — discrimination test)`
- `git checkout pr3624 -- src/commands/sync.ts (restore)`
- `gh pr diff 3583 > /tmp/pr3583.diff; grep for dry-run guards in its sweeps`
- `bun test test/sync-pull-failed-anchor.serial.test.ts > /tmp/pr3624_pull_anchor.txt (EXIT=0, 3 pass)`
- `grep -rn pullRepo src/ (single gated call site)`
- `grep src/cli.ts for 'upgrade' (command exists at cli.ts:1295)`

</details>

---

time-attack:

PR #3631, opened 2026-07-30
fix(test): kill timeout-resistant unit shards

Vyacheslav-Zakharov:
## Summary
- add a configurable 30-second TERM-to-KILL grace period to each parallel unit shard
- use GNU `timeout --kill-after` so TERM-resistant Bun descendants cannot outlive the shard cap
- classify both timeout exit 124 and forced-KILL exit 137 as wedged
- keep the no-`timeout` fallback on the same configurable grace period

## Verification
- `bun test test/scripts/run-unit-parallel.test.ts` — 10 pass, 0 fail
- `bunx tsc --noEmit --pretty false --incremental false`
- `bash -n scripts/run-unit-parallel.sh`
- live process-group smoke: TERM-resistant parent and child were both gone after `timeout --kill-after=2s` returned 137

## Risk
Low. The change only affects shards that exceed their wall-clock cap; normal test execution is unchanged.

PR #3633, opened 2026-07-30
v0.42.68.0 perf(test): bound PGLite unit-suite memory

Vyacheslav-Zakharov:
## Summary

Bounds Bun/PGLite unit-suite memory without converting hundreds of tests or switching the PGLite production path to another engine.

- Recycles each logical shard through sequential Bun batches instead of keeping one WASM heap alive for the entire shard.
- Builds one pre-migrated PGLite snapshot with the same runtime-rendered test schema used by `initSchema()`, including the 1536-dimensional test profile.
- Validates snapshots against schema/migration semantics, handler source, migration flags, exact PGLite package version, extension identity, and tar checksum.
- Falls back observably to cold initialization for missing, stale, incompatible, or corrupt snapshots.
- Keeps migration/bootstrap contracts on the cold path and resets snapshot state across reconnects.
- Uses conservative defaults measured on a 12 GiB host: 3 shards, 10 normal files/process, 2 cold files/process, 1500s shard cap.
- Improves heartbeat reporting for recycled batches and aggregates multiple Bun summary blocks correctly.

Related: #3631 adds TERM-to-KILL escalation for timeout-resistant shard descendants. It is intentionally not part of this PR's diff.

## Measurements

Representative 41-file bounded benchmark:

| Files/process | Wall time | Max RSS |
|---:|---:|---:|
| 5 | 34.49s | 1,149,440 KiB |
| 10 | 32.97s | 1,036,796 KiB |
| 20 | 31.13s | 1,229,484 KiB |
| 40 | 31.04s | 1,783,492 KiB |

`10` kept wall time within 6.2% of `40` while using about 42% less peak RSS.

A bounded broad run with the final `3 × 10` defaults completed the runner in 15:24 with:

- systemd cgroup memory peak: **3.1 GiB**
- swap peak: **0 B**
- OOM / OOM-kill: **none**
- observed result: 9,434 passed, 8 failed

The eight failures reproduce when their files are run separately, including on the cold path where applicable; this PR does not modify those test files or their production targets. The broad measurement was run while #3631 was stacked locally. This commit was then rebased directly onto `master` and the focused checks below were rerun.

For comparison, the prior `4 × 10` bounded run reached about 4.42 GiB and started swapping, so it was stopped and the default fan-out was reduced to 3.

## Verification

- `bun test test/pglite-snapshot-metadata.test.ts test/scripts/run-unit-shard.test.ts test/scripts/run-unit-parallel.test.ts` — **24 passed, 0 failed**
- bounded `test/pglite-snapshot.serial.test.ts` + `test/source-health.test.ts` — passed
- `bunx tsc --noEmit --pretty false --incremental false` — passed
- Bash syntax checks for the modified runners/build scripts — passed
- `bun run verify` — **33/33 checks passed**
- snapshot mismatch/corruption fallback smoke — passed

## Risks and safeguards

- Snapshot loading is test-only and opt-in through `GBRAIN_PGLITE_SNAPSHOT`.
- Metadata format changes invalidate old snapshots instead of trusting them.
- Exact PGLite package-version mismatch and checksum mismatch force cold initialization.
- Migration/bootstrap tests bypass the snapshot; tests that explicitly clear the snapshot environment are detected and isolated into cold batches.
- A failed batch stops the rest of that shard and logs how many batches were not run, preserving fail-fast behavior without presenting a partial shard as complete.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,6 +45,7 @@ AGENTS.local.md
 
 # Tier 3 PGLite snapshot fixture (built on demand by build:pglite-snapshot)
 test/fixtures/pglite-snapshot.tar
+test/fixtures/pglite-snapshot.metadata.json
 test/fixtures/pglite-snapshot.version
 
 # Private brain reports — never check these in (per CLAUDE.md privacy rule)
--- a/docs/TESTING.md
+++ b/docs/TESTING.md
@@ -11,13 +11,23 @@ Seven test command tiers, each with a clear scope:
 
 | Command | What it runs | Wallclock | When to use |
 |---|---|---|---|
-| `bun run test` | Parallel unit-test fast loop. 8-shard fan-out via `scripts/run-unit-parallel.sh`, then a serial pass over `*.serial.test.ts`. Excludes `*.slow.test.ts` and `test/e2e/*`. No pre-checks, no typecheck. | ~85s on a Mac dev box (3650+ tests) | Inner edit loop. Default. |
+| `bun run test` | Parallel unit-test fast loop. 3-shard fan-out via `scripts/run-unit-parallel.sh`; each shard recycles Bun after bounded file batches, then a serial pass over `*.serial.test.ts`. Excludes `*.slow.test.ts` and `test/e2e/*`. Builds and validates a pre-migrated PGLite snapshot before fan-out. No pre-checks, no typecheck. | Host-dependent | Inner edit loop. Default. |
 | `bun run verify` | CI's authoritative pre-test gate set, fanned out in parallel by `scripts/run-verify-parallel.sh`: the full `check:*` battery (~30 checks — privacy, jsonb, progress, source-id, test-isolation, wasm, …) plus `bun run typecheck`. The `CHECKS` array in that script is the single source of truth — CI literally calls `bun run verify` in a dedicated job. | ~16s (parallel; typecheck dominates) | Before pushing; before `/ship`. |
 | `bun run test:full` | `verify && bun run test && bun run test:slow && [smart e2e]`. The local equivalent of "everything CI runs." Smart e2e: runs e2e only when `DATABASE_URL` is set; else loud skip notice to stderr. | ~3-5min depending on slow + e2e | Pre-merge sanity, before opening a PR. |
 | `bun run test:slow` | Just the `*.slow.test.ts` set (intentional cold-path correctness checks). | seconds-to-minutes | When touching slow-path code. |
 | `bun run test:serial` | Just the `*.serial.test.ts` set (cross-file-contention quarantine; one bun process per file for true module-registry isolation). | ~1s per quarantined file | Debugging a specific quarantined file. |
 | `bun run test:e2e` | Real Postgres E2E. Requires Docker + `DATABASE_URL`. Sequential. | ~5-10min | Pre-ship; nightly. |
 | `bun run check:all` | The historical pre-check scripts (22, chained sequentially in package.json). Overlaps `verify` heavily but is NOT a superset — `verify`'s `CHECKS` array in `scripts/run-verify-parallel.sh` (~30 entries incl. typecheck) is the authoritative gate; `check:all` keeps a few local-only extras (trailing-newline, exports-count, no-legacy-getconnection). | ~10s | Local-only sweep for the extras. |
+
+#### Unit-runner memory controls
+
+The fast-loop runner limits retained PGLite/WASM memory in two complementary ways:
+
+- `GBRAIN_TEST_BATCH_SIZE` (default `10`) limits files per Bun process. The process exits after each batch so its WASM high-water memory returns to the OS. `--batch-size N` overrides it locally. Cold migration/bootstrap coverage uses the stricter `GBRAIN_TEST_COLD_BATCH_SIZE` (default `2`).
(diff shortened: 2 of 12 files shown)


time-attack:

Substance is solid and unusually well-tested — bounded Bun batch recycling plus a rigorously validated PGLite snapshot fast-restore, fail-open to cold init at every branch, with tests that provably pin the new behavior (both discrimination stubs made them fail). The sole blocker is mechanical: the version trio was never bumped — VERSION/package.json/CHANGELOG all sit at base's 0.42.67.0 while master is 0.42.68.1, so the CI version-gate rejects it, and the title's claimed v0.42.68.0 exists in no file and is already stale/colliding. Re-bump to a fresh version (5-file sync + bun.lock + title), rebase, and this merges.

### Discrimination test

Two-sided: (A) stubbed master's run-unit-shard.sh/run-serial-tests.sh back in and re-ran test/scripts/run-unit-shard.test.ts — the new tests fail (ERROR: unknown arg: --batch-size; wrong exit codes), so they genuinely pin the batching behavior; (B) sabotaged isSnapshotMetadataCompatible to always return true and re-ran test/pglite-snapshot.serial.test.ts — 'cold-initializes when metadata is stale' fails (exit 1), proving connect() actually routes through the validator. Both restored; worktree diff clean afterward. Failing-test propagation is pinned by the fake-bun test (batch exits 7 → shard exits 7, remaining batches reported unrun).

**Mergeability at review time:** MERGEABLE / mergeStateStatus=UNSTABLE (no conflicts; rollup pending — the version-gate will fail it as-is)

### Blockers

**1.** Version trio not bumped: branch VERSION, package.json, and CHANGELOG top entry are all 0.42.67.0 (inherited from base) while master is 0.42.68.1 — CI version-gate requires strictly greater. Title's v0.42.68.0 appears in no file, is below master, and collides with other open PRs. Fix: allocate a fresh version, update VERSION + package.json + CHANGELOG entry, run bun install for bun.lock, retitle the PR version-first.

### Not verified

- Author's broad-suite claim (9,434 pass / 8 fail, 3.1 GiB cgroup peak, 0 swap) — reproducing requires the full unit suite, forbidden by execution-footprint rules; the 8 failures being pre-existing is the author's claim only
- 3-shard x 1500s shard-timeout headroom on slow Mac hosts: shard count drops 4→3 (+33% files/shard) plus per-batch bun startup overhead against an unchanged 1500s cap; snapshot restore likely more than compensates (author's full runner: 15:24 total; source-health went from multi-second cold replay to 1.51s here) but I did not run a full shard locally
- GBRAIN_PGLITE_SNAPSHOT=off path in run-unit-parallel.sh read-verified only, not executed
- Tests that reconfigure embedding dims AFTER engine.connect() would silently keep the snapshot's 1536 schema (pre-existing _snapshotLoaded semantics, but this PR widens snapshot exposure to the whole unit loop); classifier + marker grep + connect-time rendered-schema matching mitigate, and the author's broad run is the evidence the suite is clean — not reproduced here

- `gh pr view 3633 --json mergeable,mergeStateStatus,files,body,title`
- `git fetch origin pull/3633/head:pr3633 && git checkout pr3633 (worktree)`
- `3-line VERSION/package.json/CHANGELOG audit on branch vs origin/master:VERSION`
- `bun install`
- `bun test test/pglite-snapshot-metadata.test.ts test/scripts/run-unit-shard.test.ts test/scripts/run-unit-parallel.test.ts (24 pass / 0 fail)`
- `bun test test/pglite-snapshot.serial.test.ts (3 pass; builds real snapshot via spawned builder)`
- `discrimination A: cp master scripts in, re-run run-unit-shard.test.ts (fails: unknown arg --batch-size), restore`
- `discrimination B: stub isSnapshotMetadataCompatible => true, re-run serial test (1 fail: cold-initializes when metadata is stale), restore`
- `bash scripts/check-engine-dynamic-import.sh (ok, 3 files)`
- `bash scripts/run-unit-parallel.sh --dry-run; bash scripts/run-unit-shard.sh --dry-run-list`
- `bun run build:pglite-snapshot (exit 0; artifacts gitignored, git status clean)`
- `GBRAIN_PGLITE_SNAPSHOT=test/fixtures/pglite-snapshot.tar bun test test/source-health.test.ts (28 pass, 1.51s, 0 cold Schema-version lines)`
- `greps: stale .version consumers (none), sibling callers of computeSnapshotSchemaHash/createSnapshotMetadata (all updated), delete-env marker files, CI workflow GBRAIN_PGLITE_SNAPSHOT usage (none — CI unaffected)`

</details>

---

time-attack:
Review finding (rebase + two changes requested): the memory problem is real (one Bun process retains every PGLite/WASM heap for ~159 files x 4 shards) and the batching + snapshot-metadata tightening are genuinely good — the metadata binding closes a real hole where a 1280d snapshot could load into a 1536d profile.

Blockers: (1) `run_group` aborts on the first failed batch, so a failing run executes strictly FEWER test files than master's single-invocation shard — that breaks failure-triage (the ship workflow classifies the full failure list; your own PR body reports 8 failures it couldn't fully triage for exactly this reason). Run all batches, aggregate exit codes. (2) Drop the hardcoded `N=3`/`SHARD_TIMEOUT=1500` hunk — that's one 12GiB box's tuning as a repo default, and it collides with #3518's derived cap. Note the ordering we're planning: #3631 (TERM→KILL, merged in the current wave) first, then this rebased on top with the shard-count hunk dropped. Also note your memory fix is local-loop only — CI's 10 shards run `scripts/test-shard.sh`, which this doesn't touch; worth saying so in the body.

time-attack:

click once rebased.

PR #3634, opened 2026-07-30
fix: health metric that cannot converge, and facts lost to a row_num collision

georgell-ceo:
Two independent bugs found while debugging why `gbrain doctor --remediate` never converged on a ~47k-page brain. Both are still present on `master`. Each commit stands alone.

## 1. `missing_embeddings` disagrees with the embedder

`getHealth()` counts `content_chunks.embedded_at IS NULL` with no join to `pages`. The code that actually resolves that number — `buildStaleChunkWhere`, used by `countStaleChunks` and `embed --stale` — predicates on `embedding IS NULL` and excludes `embed_skip` pages.

Two divergences:

- `embedded_at` can be non-NULL while `embedding` is NULL. That is precisely why the stale-chunk path keys on the embedding column; the CONSISTENCY note on the `upsertChunks` upsert spells it out. Health was keying on the other one.
- Chunks on `embed_skip` pages were counted as missing, but they are excluded from embedding by design, so the number could never reach zero.

**Where it bites.** `computeRecommendations` emits an `embed.stale` step whenever `health.missing_embeddings > 0`. `embed --stale` then reports zero work, the metric does not move, and the recheck re-plans the identical step every pass until `maxJobs`.

On the brain this was found on, all 354 chunks reported as "missing" were on `embed_skip` pages. `embed --stale` correctly reported 0. `--remediate` reported the target as unreachable and burned wall-clock re-running a step that could not help.

Both engines updated. No behaviour change where the two predicates already agreed.

## 2. Fence `row_num` is derived from the markdown, but enforced in the database

`row_num` uniqueness is enforced by `idx_facts_fence_key` on `(source_id, source_markdown_slug, row_num)`. But `writeFactsToFence` derives the value via `upsertFactRow`, which reads the facts fence out of the markdown file and **falls back to 1 when the file has no fence at all**.

Any write path that rewrites a page without preserving its fence — the `put_page` write-through, `sync`, the dream-cycle reverse-render — drops the counter below what the database already holds. The next absorb re-issues a taken `row_num`, and the batch dies with:

```
```

The facts are dropped. Worse, because the file never regains its fence, the page fails **permanently** rather than transiently — every subsequent absorb on it hits the same collision.

Measured on a live brain: **11 of 14 fence-managed pages** were already in this state. One had 24 facts in the database and none in the file, so its counter would restart at 1 and collide on the first row every time.

**Fix.** Seed the counter from `max(fileMax, dbMax) + 1`. The file stays the human-readable mirror; the database stays authoritative about which numbers have been issued. The lookup is a hint rather than a hard dependency — it degrades to the previous file-only behaviour on failure (pre-v51 brains without the fence columns, transient DB errors), so a fence write can never become impossible just because the hint was unavailable.

## Tests

Adds two cases to `test/fence-write.test.ts`:

- **the regression** — writes two facts, strips the fence from the file the way a non-fence-aware writer would, writes a third. Fails with the duplicate-key error without the fix; passes with it. Verified by stashing the source change and re-running: 15 pass / 1 fail, and the failure is exactly this case.
- **the degraded path** — an engine whose `MAX(row_num)` lookup throws still completes the fence write.

Typecheck clean. Ran the facts, fence, health, brain-score, remediation, embed and doctor suites: all green.

## Notes

- I have not touched `stale_pages` or the `sync.repo` rationale that calls those pages "stale on disk". That metric was redefined recently and the wording looked like a separate conversation rather than something to bundle here.
- Happy to split this into two PRs, adjust the comment density to taste, or rework either fix if you would rather solve it a different way.

Environment note: the full `bun test` run OOMs on my machine (macOS, the WASM
issue in #223) — but it does so identically on an unmodified `master` checkout,
273 out-of-memory errors there too, so it is not related to this change. The
suites listed above were run in smaller batches, where they pass.

### DIFF
--- a/src/core/facts/fence-write.ts
+++ b/src/core/facts/fence-write.ts
@@ -219,10 +219,51 @@ export async function writeFactsToFence(
 
       // 2. Upsert each fact onto the fence in input order. row_num
       //    monotonically increases (max-existing + 1 per call, append-only).
+      //
+      //    Seed the counter from the DB as well as the fence file. Uniqueness
+      //    is enforced by idx_facts_fence_key on
+      //    (source_id, source_markdown_slug, row_num) in Postgres, but
+      //    upsertFactRow derives the next value from the fence in the markdown
+      //    alone — and falls back to 1 when the file has no fence at all. Any
+      //    write path that rewrites a page without preserving its facts fence
+      //    (put_page write-through, sync, dream-cycle reverse-render) therefore
+      //    resets the counter below what the DB already holds, and the next
+      //    absorb re-issues a row_num that is already taken. That surfaces as
+      //    "duplicate key value violates unique constraint idx_facts_fence_key"
+      //    and the whole batch of facts is dropped.
+      //
+      //    Symptom in the wild: a page whose fence had been rewritten away had
+      //    24 facts in the DB and none in the file, so every subsequent absorb
+      //    on it failed permanently. Taking the max of both sources keeps the
+      //    file as the readable mirror while the DB stays authoritative about
+      //    which row_nums have been issued.
+      //
+      //    Degrades to the previous file-only behaviour if the lookup fails
+      //    (pre-v51 brain without the fence columns, or a transient DB error):
+      //    a fence write must not become impossible just because the counter
+      //    hint is unavailable.
+      let dbMaxRowNum = 0;
+      try {
+        const rows = await engine.executeRaw<{ max_row_num: number | null }>(
+          `SELECT MAX(row_num) AS max_row_num FROM facts
+            WHERE source_id = $1 AND source_markdown_slug = $2`,
+          [target.sourceId, target.slug],
+        );
+        dbMaxRowNum = Number(rows[0]?.max_row_num ?? 0);
+      } catch {
+        dbMaxRowNum = 0;
+      }
+      const { facts: existingFenceFacts } = parseFactsFence(body);
+      const fileMaxRowNum = existingFenceFacts.length > 0
+        ? Math.max(...existingFenceFacts.map(f => f.rowNum))
+        : 0;
+      let nextRowNum = Math.max(fileMaxRowNum, dbMaxRowNum) + 1;
+
       const assignedRowNums: number[] = [];
       for (const f of facts) {
         const validFromStr = (f.validFrom ?? new Date()).toISOString().slice(0, 10);
         const { body: updated, rowNum } = upsertFactRow(body, {
+          rowNum:      nextRowNum++,
           claim:       f.fact,
           kind:        (f.kind ?? 'fact') as 'fact' | 'event' | 'preference' | 'commitment' | 'belief',
           confidence:  f.confidence ?? 1.0,
--- a/src/core/pglite-engine.ts
+++ b/src/core/pglite-engine.ts
@@ -2407,7 +2407,7 @@ export class PGLiteEngine implements BrainEngine {
     }
 
     // CONSISTENCY: when chunk_text changes and no new embedding is supplied, BOTH embedding AND
-    // embedded_at must reset to NULL so `embed --stale` correctly picks up the row for re-embedding.
+    // embedded_at must reset to NULL so 'embed --stale' correctly picks up the row for re-embedding.
     // See postgres-engine.ts upsertChunks for the full rationale — pglite mirrors it for parity.
     //
     // v0.40.3.0 D24 NULL→non-NULL race fix mirrors postgres-engine.ts. Two writers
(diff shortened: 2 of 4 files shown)


time-attack:

Both fixes are real and correct. Bug 2 (row_num collision) confirmed: insertFacts is a plain INSERT (no ON CONFLICT) in one transaction against the v51 partial unique index, so a fence-less rewrite makes every later absorb on that page fail permanently and drop the whole batch; the max(fileMax, dbMax)+1 seed is the root-cause fix, preserves existing fact identity (upsertFactRow is append-only, never renumbers), respects the (source_id, source_markdown_slug) composite key, and degrades safely on pre-v51 brains. Bug 1 (health predicate) confirmed via my own discrimination test: old code counts embed_skip chunks and misses embedded_at-lies chunks (reported 3 vs resolver's 1 in my scenario), PR code agrees with countStaleChunks exactly, so doctor --remediate's embed.stale step converges. Engine parity intact, typecheck clean, 87 targeted tests pass, JSONB guards clean, MERGEABLE. Blocked only by mechanical gate items: no version bump (branch trio 0.42.67.0 is behind master 0.42.68.1 — CI version-gate rejects), title violates the version-first IRON RULE, and the health half ships no committed regression test. Substance is merge-worthy; land via /ship (which fixes items 1-2) after asking the author (or a takeover commit) to add a health-predicate test mirroring my ad-hoc one.

### Discrimination test

Row_num half: reverted only src/core/facts/fence-write.ts to base c6dc0adf and re-ran bun test test/fence-write.test.ts → 15 pass / 1 fail, the failure being exactly the new regression case with 'duplicate key value violates unique constraint "idx_facts_fence_key" Key (default, people/carol, 1) already exists' (matches PR body's claim verbatim); restored fix → 16/16. Health half (no PR test exists): wrote an ad-hoc PGLite test with 3 NULL-embedding chunks on an embed_skip page plus 1 chunk with embedded_at set but embedding NULL; reverted only the pglite getHealth hunk → missing_embeddings=3 vs countStaleChunks=1 (test fails, proving the old metric can never converge); PR code → 1=1 (passes). Both halves discriminate.

**Mergeability at review time:** mergeable: MERGEABLE, mergeStateStatus: UNSTABLE (no CI checks reported on the fork branch at all — nothing was verified by CI; all evidence above is from local execution)

### Blockers

**1.** VERSION trio on branch is 0.42.67.0 (all three agree) but NOT strictly greater than master's 0.42.68.1 — the CI version-gate rejects the merge as-is; no CHANGELOG entry for the fix

**2.** PR title 'fix: health metric that cannot converge...' does not lead with vMAJOR.MINOR.PATCH.MICRO (IRON RULE); fix with gh pr edit 3634 --title 'vX.Y.Z.W fix(health,facts): ...'

**3.** The health-predicate half (src/core/pglite-engine.ts:5380, src/core/postgres-engine.ts:5490) has no committed regression test — my ad-hoc discrimination test (embed_skip page + embedded_at-non-NULL/embedding-NULL chunk, assert getHealth().missing_embeddings === countStaleChunks()) fails on old code and passes on PR code; it should be added to the repo so the predicate can't silently drift back

### Not verified

- Real-Postgres execution of both hunks — verified by inspection only (SQL shape identical to countStaleChunks/insertFacts which already run on Postgres; jsonb_exists is a core function equivalent to the ? operator); no Docker spun up per footprint rules, and the DATABASE_URL-gated engine-parity e2e was not run
- Full embed/doctor/remediation suites — ran the directly-touched and adjacent test files only, per footprint rules; PR author claims those suites green in batches
- PR body field measurements (11 of 14 fence-managed pages broken, 354 embed_skip chunks, 47k-page brain) — anecdotal from the author's live brain, not reproducible here; the mechanism behind them is what I verified instead
- Residual non-convergence note: embed_coverage (still embedded_at-based over ALL chunks) and doctor's 90%-coverage warn threshold can still nag 'Run: gbrain embed --stale' on skip-heavy brains — pre-existing, deliberately out of the PR's scope, flagged for a follow-up TODO

- `gh pr view 3634 --json mergeable,mergeStateStatus,... (MERGEABLE/UNSTABLE)`
- `gh pr checks 3634 (no checks reported)`
- `git fetch origin pull/3634/head:pr-3634 && git checkout pr-3634 (base c6dc0adf, 2 commits)`
- `bun test test/fence-write.test.ts (16 pass / 0 fail on PR code)`
- `git checkout c6dc0adf -- src/core/facts/fence-write.ts && bun test test/fence-write.test.ts (15 pass / 1 fail, duplicate-key on idx_facts_fence_key) then restored`
- `bun test /tmp/pr3634_health_check.test.ts on PR code (missing_embeddings=1 == countStaleChunks=1, pass) and with pglite getHealth reverted (missing_embeddings=3 != 1, fail) then restored`
- `bun test test/facts-fence.test.ts test/health-soft-delete.test.ts test/brain-score-recommendations.test.ts (71 pass / 0 fail)`
- `bash scripts/check-jsonb-pattern.sh (exit 0) and node scripts/check-jsonb-params.mjs (exit 0)`
- `cat VERSION + node -e package.json version + grep CHANGELOG (all 0.42.67.0; origin/master:VERSION is 0.42.68.1)`

</details>

---

time-attack:

PR #3641, opened 2026-07-30
fix(test): make typecheck hermetic to ambient parent openclaw SDK types (#2729)

### LINKED ISSUE #2729: Typecheck depends on ambient parent openclaw SDK types
## Summary

`bunx tsc --noEmit` can fail on a clean v0.42.58.0 checkout depending on an undeclared `openclaw` package found in a parent `node_modules` directory.

The real plugin-load E2E test dynamically imports `openclaw/plugin-sdk`, but TypeScript still resolves and type-checks that bare module. The `@ts-ignore` is attached to the import line; the actual error occurs on the following property-access line.

## Affected version

- `v0.42.58.0`
- commit `a25209bbb2bacf1b88e06fd5282b27f1bf4a3e7a`

## Reproduction

On a clean checkout where a parent directory contains an OpenClaw SDK version without a typed `registerContextEngine` export:

```bash
bun install --frozen-lockfile
bunx tsc --noEmit
```

Observed:

```text
test/e2e/openclaw-plugin-load-real.test.ts(281,37): error TS2339:
Property 'registerContextEngine' does not exist on type
'typeof import(".../node_modules/openclaw/dist/plugin-sdk/index")'.
```

`package.json` declares OpenClaw compatibility metadata but does not declare `openclaw` as a dependency or devDependency. Bun/TypeScript therefore walks upward and compiles against ambient parent state.

Relevant code:

```ts
// @ts-ignore — bare specifier resolution depends on node_modules.
const sdk = await import('openclaw/plugin-sdk');
registerContextEngine = sdk.registerContextEngine;
```

The ignore does not cover the failing property access.

## Expected behavior

A clean checkout's typecheck should not depend on whatever OpenClaw SDK happens to be installed in a parent directory. The runtime compatibility test may still dynamically probe the installed CLI/SDK, but compile-time verification should remain hermetic.

## Possible fix

Treat the dynamically imported SDK as an unknown runtime object and validate the optional property structurally, for example with an explicit `unknown` cast and a local interface/type guard. This keeps the runtime test's intent without binding repository typecheck to an undeclared ambient package version.

## Baseline proof

The failure was reproduced after stashing all local changes and running `bunx tsc --noEmit` against untouched v0.42.58.0. Restoring the unrelated provider patch produced the identical single failure.

Masashi-Ono0611:
## Root cause

`test/e2e/openclaw-plugin-load-real.test.ts` dynamically imports `openclaw/plugin-sdk`, which is deliberately not a declared dependency — the Tier 2 test probes whatever install is present at runtime. But TypeScript still resolves and type-checks that bare specifier through upward `node_modules` resolution, so `bunx tsc --noEmit` on a clean checkout depends on ambient parent state:

- If no ancestor `node_modules` provides `openclaw`: the `@ts-ignore` on the import line suppresses TS2307 and typecheck passes.
- If an ancestor `node_modules` provides an `openclaw` whose typed SDK surface lacks `registerContextEngine`: TS2339 fires on the property-access line **after** the import — outside the `@ts-ignore`'s reach — and typecheck fails:

```text
test/e2e/openclaw-plugin-load-real.test.ts(281,37): error TS2339:
Property 'registerContextEngine' does not exist on type
'typeof import(".../node_modules/openclaw/dist/plugin-sdk/index")'.
```

Repo typecheck outcomes should not depend on what unrelated software is installed in ancestor directories.

## Fix

Cast the awaited import to a local structural interface declaring the one member the test uses (`registerContextEngine`, optional):

```ts
interface OpenclawPluginSdk {
  registerContextEngine?: (id: string, factory: () => unknown) => void;
}
// @ts-ignore — bare specifier; whether this resolves (TS2307 or not)
// depends on ambient node_modules, so @ts-expect-error would flip.
const sdk = (await import('openclaw/plugin-sdk')) as unknown as OpenclawPluginSdk;
```

- TypeScript never consults the ambient module's types for the property access, so typecheck is hermetic by construction — identical output whether or not any ancestor `node_modules` provides `openclaw`, at any version. It's also immune to future line drift, unlike moving/adding suppression comments.
- The `@ts-ignore` stays on the import statement and stays `@ts-ignore` (not `@ts-expect-error`): whether TS2307 fires there is itself ambient-dependent, so `@ts-expect-error` would error in exactly the environments where the import resolves.
- Runtime behavior is unchanged: `as` casts erase at compile time, and the export's presence/shape is still verified at runtime (`if (!registerContextEngine)` fallback chain + `expect(typeof registerContextEngine).toBe('function')`).
- The second, absolute-path fallback import (`await import(sdkPath)`) is untouched: its specifier is a runtime variable, so TS types it `any` and it has no hermeticity problem.
- No dependency changes — `openclaw` remains intentionally undeclared; declaring it just to obtain types would change the dependency graph for a types-only concern.

## Verification

- `bunx tsc --noEmit` — exit 0, zero diagnostics.
- `bun test test/e2e/openclaw-plugin-load-real.test.ts` — runs, 6 skip / 0 fail (openclaw CLI not installed in this environment; Tier 2 gating by design). The change is purely type-level, so the runtime path is unaffected.
- Ambient condition on the verification machine: no ancestor directory of the checkout has a `node_modules` at all, so the failing case cannot be reproduced locally — the pass here covers the "openclaw unresolvable" branch. The "openclaw resolvable" branch is covered by construction: with the structural cast, the property access never touches the ambient module's types, so its presence/absence/version cannot change typecheck output. (Confirmed independently by a second reviewer pass that verified `openclaw/plugin-sdk` is unresolvable and typecheck is clean.)

Fixes #2729


time-attack:
**Merging this closes #2729** (severity **p2**).

#2729 was one of 58 issues carrying a `p3` label. All 58 were re-triaged by reproducing them against current master, and **the label was wrong on 56 of them** — the corrected distribution is 4 p0, 40 p1, 12 p2, and 2 genuinely p3. #2729 came out at **p2**, and this PR is what fixes it.

Two things follow from that:

- Add `Closes #2729` to the PR description if you want the link wired up — a closing keyword only fires from the body or a commit message, not from a comment like this one.

The reproduction and evidence are on #2729.

time-attack:

The fix is real and correctly built: casting the dynamic `openclaw/plugin-sdk` import to a local structural interface makes the test file's typecheck hermetic, verified by a live repro (hostile ambient openclaw makes master fail with the exact TS2339 at 281,37; PR branch stays clean). But the PR patches one of two vulnerable call sites: src/core/context-engine.ts:88-90 uses the identical @ts-ignore-import + unguarded-property-access pattern on `openclaw/plugin-sdk/core`, and I reproduced the same class of ambient-dependent TS2339 failure there ON THE PR BRANCH (context-engine.ts(89,40) and (90,44)) with an ambient openclaw whose typed plugin-sdk/core lacks the two accessed exports. The surviving site is in src/, worse than the test file, and 'Fixes #2729' would close the issue while its stated expectation ('compile-time verification should remain hermetic') remains falsifiable. Required change: apply the same local-interface cast (~4 lines) to the context-engine.ts import. With that added, this is a MERGE.

### Discrimination test

Planted a fake node_modules/openclaw with a typed ./plugin-sdk export lacking registerContextEngine (the exact ambient condition from #2729). Old code (master's file stubbed back in): tsc exit 2, TS2339 at openclaw-plugin-load-real.test.ts(281,37) — identical to the issue report. PR fix, same hostile ambient: tsc exit 0. No ambient: exit 0 both ways. The test discriminates; the fix is load-bearing. Then extended the fake with a typed ./plugin-sdk/core lacking the accessed exports: PR branch fails tsc in src/core/context-engine.ts — the sibling site is provably still broken.

### Blockers

**1.** src/core/context-engine.ts:88-90 — same non-hermetic pattern the PR claims to eliminate. Failing input: any ancestor/local node_modules containing an openclaw whose package.json exports a typed './plugin-sdk/core' lacking delegateCompactionToRuntime or buildMemorySystemPromptAddition. Reproduced on the PR branch: `bunx tsc --noEmit` exits 2 with TS2339 at context-engine.ts(89,40) and (90,44). Fix: cast `await import('openclaw/plugin-sdk/core')` through a local structural interface exactly as this PR does for the test file.

### Not verified

- The 'ambient openclaw resolvable WITH compatible typed registerContextEngine' branch was not separately executed — covered by construction (as-unknown-as cast never consults ambient types; the incompatible-ambient run already exercises the stronger case)
- Full unit suite / bun run verify not run (footprint rule); typecheck + the touched test file were run
- CI rollup status on the PR head not re-inspected beyond mergeStateStatus CLEAN
- Real openclaw CLI installed end-to-end (Tier 2 path executes rather than skips) — no openclaw install on this machine; PR is type-only so runtime path is unchanged by construction

- `gh pr view 3641 --json mergeable,mergeStateStatus (MERGEABLE/CLEAN)`
- `gh issue view 2729`
- `gh pr diff 3641`
- `git fetch origin pull/3641/head:pr-3641 && git checkout pr-3641`
- `bun install (283 packages; worktree shipped without node_modules)`
- `bunx tsc --noEmit on PR branch, no ambient openclaw -> exit 0`
- `planted fake node_modules/openclaw (typed ./plugin-sdk without registerContextEngine); bunx tsc --noEmit on PR branch -> exit 0 (hermetic)`
- `git checkout origin/master -- test/e2e/openclaw-plugin-load-real.test.ts; bunx tsc --noEmit -> exit 2, TS2339 at (281,37) (old behavior fails; discrimination confirmed)`
- `extended fake with typed ./plugin-sdk/core lacking accessed exports; bunx tsc --noEmit on PR branch -> exit 2, TS2339 in src/core/context-engine.ts(89,40)+(90,44) (sibling still broken)`
- `rm -rf fake openclaw; bunx tsc --noEmit -> exit 0`
- `bun test test/e2e/openclaw-plugin-load-real.test.ts -> 6 skip / 0 fail, exit 0 (matches PR body)`
- `grep for all openclaw import sites across src/ and test/`
- `gh pr list --state merged --limit 8 (title/version convention check)`

</details>

---

PR #3647, opened 2026-07-30
fix(agent): resolve the brain source at submit time instead of hardcoding the seed default

### LINKED ISSUE #2922: Subagent tools ignore the configured default source: put_page always writes to seed 'default' (hardcoded in brain-allowlist.ts)

Masashi-Ono0611:
## Problem

`gbrain agent run` never consults the source-resolution chain, so every page an agent job writes lands in the seed `default` source — even on brains with `gbrain sources default <id>` configured. On access-controlled multi-source brains the seed `default` is typically unfederated, making agent-written pages invisible to readers.

Per the rescoping comment on #2922: the downstream plumbing already exists on master (#1586) — `SubagentHandlerData.source_id` is threaded by the subagent handler into `buildBrainTools`/`buildOpContext`, which validates it and stamps `OperationContext.sourceId` (falling back to `'default'` only when absent). The remaining gap is the entry point: `src/commands/agent.ts` contained zero `sourceId` references, so nothing ever populated the field the handler was ready to consume. (#2925 addressed the wider original scope but was closed after #1586 landed its payload-plumbing half; this PR is the narrowed remainder.)

That comment flags the new CLI surface as a maintainer call — hence Draft. Happy to drop or rename the flag if you'd rather land only the resolver threading.

## Fix

- `runAgentRun` resolves the source once at submit time via the existing `resolveSourceId` chain (`src/core/source-resolver.ts`: explicit flag → `GBRAIN_SOURCE` → `.gbrain-source` → local_path match → `sources.default` → sole non-default → `'default'`) and stamps it on `SubagentHandlerData.source_id` for the single-job path and both fan-out paths (single-entry short-circuit + N-entry children). Submit time rather than claim time, so the worker daemon's cwd/env can't win over the submitter's intent.
- New `--source <id>` flag on `gbrain agent run`, matching the flag convention used by `sync`, `dream`, etc. An unknown explicit source errors loudly (via the resolver's `assertSourceExists`) rather than silently falling back — consistent with #3524. An empty `--source ""` is a usage error instead of silently degrading to the lower tiers.
- The `__all__` sentinel is rejected at submit time with a clear message: subagent writes must target exactly one source, and `validateSourceId` at tool-registry build time would reject it at claim anyway — failing at submit is strictly better.
- Archived-source guard mirroring `dream.ts`: submitting agent jobs against an archived source errors with the restore hint instead of writing pages that stay invisible until restore.
- Resolver user errors (unknown source / malformed id) surface as a clean stderr line + exit, mirroring `dream.ts:isResolverUserError`; genuine bugs still propagate with a stack trace.

Single-source brains see no behavior change: with no signal at any tier the resolver returns `'default'`, which is exactly what `buildOpContext` used before.

## Verification

- `bun test test/agent-cli.test.ts`: 42 pass, 0 fail — includes 7 new tests: `--source` flag parsing (+ missing-value usage error), explicit `--source` landing on `SubagentHandlerData.source_id`, `sources.default` (tier 5) honored without a flag, `GBRAIN_SOURCE` (tier 2) honored, no-signal fallback to `'default'` preserved, and fan-out children all carrying the resolved `source_id`. Env mutation goes through the `withEnv` helper.
- `bun test test/brain-allowlist.serial.test.ts test/subagent-handler.test.ts test/source-resolver-silent-fallback.test.ts test/code-scoped-source-resolve.test.ts test/subagent-aggregator.test.ts test/subagent-v1-v2-shim.test.ts`: 97 pass, 0 fail.
- `tsc --noEmit`: passed.
- `scripts/check-test-isolation.sh`: OK (1072 non-serial unit files scanned).

Fixes #2922

### DIFF
--- a/src/commands/agent.ts
+++ b/src/commands/agent.ts
@@ -18,6 +18,7 @@ import type { BrainEngine } from '../core/engine.ts';
 import { MinionQueue } from '../core/minions/queue.ts';
 import { waitForCompletion, TimeoutError } from '../core/minions/wait-for-completion.ts';
 import type { MinionJobInput, SubagentHandlerData, AggregatorHandlerData } from '../core/minions/types.ts';
+import { resolveSourceId, ALL_SOURCES } from '../core/source-resolver.ts';
 import { runAgentLogs } from './agent-logs.ts';
 
 // ── arg parsing helpers ────────────────────────────────────
@@ -72,6 +73,10 @@ SUBMITTING
     --max-turns <n>              Max assistant turns (default 20)
     --tools a,b,c                Subset of registered tool names (comma list)
     --timeout-ms <n>             Per-job wall-clock timeout
+    --source <id>                Brain source the subagent's writes are scoped to.
+                                  Default: the standard resolution chain (GBRAIN_SOURCE,
+                                  .gbrain-source, sources.default, ...) — see
+                                  \`gbrain sources current\`
     --fanout-manifest <path>     JSON array of {prompt, input_vars?} — one child each
     --follow                     Tail status until terminal (default on TTY)
     --detach                     Submit + print job id, exit immediately
@@ -116,6 +121,7 @@ interface RunFlags {
   maxTurns?: number;
   tools?: string[];
   timeoutMs?: number;
+  source?: string;
   fanoutManifest?: string;
   follow: boolean;
   detach: boolean;
@@ -181,6 +187,7 @@ function parseRunFlags(args: string[]): { flags: RunFlags; rest: string[] } {
       case '--max-turns':       flags.maxTurns = parseIntFlagValue(requireFlagValue(args, ++i, a), a); break;
       case '--tools':           flags.tools = requireFlagValue(args, ++i, a).split(',').map(s => s.trim()).filter(Boolean); break;
       case '--timeout-ms':      flags.timeoutMs = parseIntFlagValue(requireFlagValue(args, ++i, a), a); break;
+      case '--source':          flags.source = requireFlagValue(args, ++i, a); break;
       case '--fanout-manifest': flags.fanoutManifest = requireFlagValue(args, ++i, a); break;
       case '--follow':          flags.follow = true; break;
       case '--no-follow':       flags.follow = false; break;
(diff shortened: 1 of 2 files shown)


Masashi-Ono0611:
This directly implements the gap you scoped out in #2922:

> Fixing it means either adding a `--source` flag to `gbrain agent run` or threading the existing source resolver into that submit path — i.e. new CLI surface, which is a maintainer call rather than a drive-by fix.

I went with the `--source` flag + resolver threading at submit time. It also follows the behavior contract you called out in that same comment:

> Whatever lands here should match that behavior [#3524's error-not-fallback] rather than reintroducing a silent fallback.

An unknown explicit `--source` value errors via `assertSourceExists` instead of silently falling back to `default`, matching #3524.

Marked Ready since the underlying fix is complete and CI is green — happy to drop or rename the flag if you'd rather land only the resolver threading, per the original note in the PR description.

Masashi-Ono0611:
My human part:
While my GBrain is still single-source, i found a risk where GBrain could’t read pages written by agent run for multi-source GBrain users. i think i also will go to multi-source very soon, so would love to fix this issue.

Screenshot:
Screenshot 2026-08-02 at 1 03 46 PM

time-attack:
Merged in the v0.45.10.0 community fix-wave 2 (#4085). Your fix landed on master as 0cfedd026d, credited to you via Co-Authored-By, and is called out in the v0.45.10.0 CHANGELOG entry. Thank you for the contribution!

PR #3653, opened 2026-07-30
feat(synth): skip re-synthesizing concepts with unchanged membership

alexey-metaengage:
## What

`synthesize_concepts` stamps a per-concept membership signature (`frontmatter.member_sig` = hash of a version tag + sorted `(atom slug, content_hash)` pairs) and skips the LLM call + page write when a concept's atom membership is unchanged since the last run.

## Why

Every nightly run currently re-synthesizes every concept regardless of whether its atoms changed, spending LLM budget on identical output. Only concepts whose membership actually changed (add / edit / soft-delete / re-tag / tier crossing) need to refresh.

## Behavior

- Sorted `(slug, content_hash)` pairs make adds, soft-deletes, re-tagging, AND atom-body edits observable.
- A `SYNTH_VERSION` tag folded into the hash forces a full re-synthesis when the prompt/model changes.
- Fail-soft: any signature-prefetch query error falls back to an empty map, which behaves exactly like today (re-synthesize everything).
- New `concepts_skipped_unchanged` count in the phase result + summary line.

Rides the existing `importFromContent` concept write path; no schema change.

### DIFF
--- a/src/core/cycle/synthesize-concepts.ts
+++ b/src/core/cycle/synthesize-concepts.ts
@@ -30,7 +30,12 @@ import { chat as gatewayChat, isAvailable } from '../ai/gateway.ts';
 // source-boost's 1.3× 'concepts/' weighting can actually reach them.
 import { importFromContent } from '../import-file.ts';
 import { serializeMarkdown } from '../markdown.ts';
+import { createHash } from 'crypto';
 
+// Bump to force a full re-synthesis pass (e.g. when SYNTH_PROMPT or the model
+// changes). Folded into the per-concept membership signature so a prompt
+// change invalidates every stored signature at once.
+const SYNTH_VERSION = 'v0.41';
 const DEFAULT_BUDGET_USD = 1.5;
 const TIER_T1_MIN = 10;
 const TIER_T2_MIN = 5;
@@ -50,13 +55,16 @@ export interface SynthesizeConceptsOpts {
   /** Test seam: alternative chat function. */
   _chat?: typeof gatewayChat;
   /** Test seam: skip DB query; cluster these atoms directly. */
-  _atoms?: Array<{ slug: string; concept_refs: string[]; body: string; title: string }>;
+  _atoms?: Array<{ slug: string; concept_refs: string[]; body: string; title: string; content_hash?: string }>;
 }
 
 interface AtomGroup {
   conceptSlug: string;
   atomTitles: string[];
   atomBodies: string[];
+  /** (slug, content_hash) of every atom in the group — drives the membership
+   * signature used to skip re-synthesis when nothing changed. */
+  members: Array<{ slug: string; hash: string }>;
   tier: 'T1' | 'T2' | 'T3' | 'T4';
 }
 
@@ -81,9 +89,10 @@ export async function runPhaseSynthesizeConcepts(
         slug: string;
         title: string;
         compiled_truth: string;
+        content_hash: string | null;
         frontmatter: { concepts?: string[]; imported_from?: string };
       }>(
-        `SELECT slug, title, compiled_truth, frontmatter
+        `SELECT slug, title, compiled_truth, content_hash, frontmatter
            FROM pages
           WHERE type = 'atom'
             AND deleted_at IS NULL
@@ -96,6 +105,7 @@ export async function runPhaseSynthesizeConcepts(
           title: r.title,
           body: r.compiled_truth,
           concept_refs: r.frontmatter!.concepts!,
+          content_hash: r.content_hash ?? undefined,
         }));
     } catch {
       // No atoms table or query failed — phase no-ops cleanly.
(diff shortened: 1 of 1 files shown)


time-attack:

The feature is real and correctly wired: member_sig round-trips through importFromContent → pages.frontmatter → the prefetch query; a second run with identical membership makes zero LLM calls, and any hash change/atom add re-synthesizes (verified with a 5-test PGLite harness at test/cycle/pr3653-discrimination.test.ts; 3 of the core tests fail when master's file is stubbed back in, so they discriminate). Typecheck clean, existing 30 synthesize-concepts tests pass, MERGEABLE/CLEAN, no engine/schema/JSONB/env-knob issues. BUT it ships a proven regression: src/core/cycle/synthesize-concepts.ts:274 stamps member_sig unconditionally, including when the narrative came from the LLM-failure fallback (catch at :253-259) or the budget-exhausted fallback (:224-225). Pre-PR, a transient gateway outage self-healed on the next nightly run; post-PR the template stub is permanent until atom membership changes or SYNTH_VERSION is manually bumped — my test proves run 2 makes 0 chat calls after a failed run 1 (expected 1). Fix: omit member_sig (or write a poison value) on fallback-written T1/T2 pages. Also: the PR ships zero tests despite extending the _atoms seam for exactly this, the title lacks the mandatory version-first prefix, and there is no VERSION/package.json/CHANGELOG bump (branch trio sits at 0.42.67.0, below master's 0.42.68.1). Aim is right — nightly re-synthesis waste is real and the fail-soft prefetch design is careful — so fix and resubmit, not close.

### Discrimination test

Wrote test/cycle/pr3653-discrimination.test.ts (5 tests, real PGLite engine + counting chat stub): unchanged membership skips with 0 LLM calls; changed content_hash re-synthesizes; added atom re-synthesizes; transient-LLM-failure retry; dry-run writes nothing. On PR branch: 4 pass, 1 fail (the failure-retry test — the proven regression). With master's synthesize-concepts.ts stubbed back in: 3 fail (skip fields absent), proving the tests discriminate and the feature is actually wired end-to-end, not decorative.

### Blockers

**1.** src/core/cycle/synthesize-concepts.ts:274 — member_sig stamped on LLM-failure fallback writes (catch at :253) and budget-exhausted fallback (:224); failing input: gateway down for one nightly run → concept permanently keeps the deterministic template narrative; proven by test (run 2 chat calls: expected 1, received 0 on PR branch; passes on master)

**2.** Zero tests in the PR: 64-line behavior change with the _atoms seam extended for testability but no test asserting skip/re-synthesis; a skip-path test would have caught the fallback-stickiness regression

**3.** PR title 'feat(synth): skip re-synthesizing concepts with unchanged membership' violates the version-first IRON RULE (must be 'vX.Y.Z.W feat(synth): ...')

**4.** No VERSION/package.json/CHANGELOG bump: branch trio is 0.42.67.0 across all three, below master's 0.42.68.1, and the feature has no CHANGELOG entry

### Not verified

- Real-Postgres roundtrip of the sig prefetch/write (only PGLite exercised; no engine SQL changes so risk is low, and footprint rules ruled out Docker for this)
- The PR body's implied cost-savings magnitude at scale (code read confirms pre-PR unconditional nightly re-synthesis, but no quantitative reproduction)
- Whether takes-resolution/enrichment paths can mutate compiled_truth without content_hash beyond revertToVersion (revertToVersion confirmed by code read to update compiled_truth+frontmatter without content_hash — a pre-existing wart the sig proxy inherits; flagged as a note, not a blocker)

- `gh pr view 3653 --json ... (mergeable: MERGEABLE, mergeStateStatus: CLEAN on retry; first call UNKNOWN)`
- `git fetch origin pull/3653/head:pr-3653 && git checkout pr-3653`
- `git diff origin/master...pr-3653 (single file: src/core/cycle/synthesize-concepts.ts +64/-5)`
- `git show pr-3653:VERSION (0.42.67.0) vs origin/master:VERSION (0.42.68.1)`
- `bun test test/cycle/pr3653-discrimination.test.ts (PR branch: 4 pass initially; then 4 pass/1 fail after adding the failure-retry test — Expected 1, Received 0)`
- `git checkout origin/master -- src/core/cycle/synthesize-concepts.ts && bun test test/cycle/pr3653-discrimination.test.ts (1 pass, 3 fail — discriminates), then restored PR version`
- `bun test test/cycle/extract-atoms-synthesize-concepts.test.ts test/cycle/synthesize-concepts-progress.test.ts (30 pass, 0 fail)`
- `grep member_sig across src/ + test/ (no other references, no tests in PR)`
- `code-read: import-file.ts sourceId default 'default'; progress.ts tick(n=1) accepts 0; pglite-engine.ts revertToVersion updates compiled_truth without content_hash`

</details>

---

time-attack:
Closing. This is a cost optimization rather than a bug fix — nightly re-synthesis of unchanged concepts is working as designed — and the implementation has two problems that outweigh the saved spend: the signature keys on `pages.content_hash`, which is nullable and coalesced to `''` here, so on brains where atom pages carry no content_hash an atom-body edit becomes invisible and the concept goes permanently stale; and the `member_sig` prefetch is unscoped by source while the write path pins `source_id: 'default'`, so a same-slug concept in another source can shadow-match and suppress a legitimate re-synthesis. There are also no tests for the new skip branch despite two existing test files for this phase. Silent staleness is a worse failure than redundant work. Thanks for the contribution.

PR #3655, opened 2026-07-30
feat(migrate): opt-in RLS enable without BYPASSRLS

alexey-metaengage:
## Context

Follows #3212 / #3283 (both closed). The position there was to keep the fail-closed `RAISE EXCEPTION` for the RLS-enable migrations (v24/v29/v31/v35) when the migrating role lacks BYPASSRLS, because on a managed-Postgres split (migrator role != runtime app role) enabling RLS without BYPASSRLS can lock the app out of its own data. That default is correct and this PR keeps it unchanged.

## What this adds

An **opt-in** escape hatch for the single-principal case: when the migrating role also OWNS the tables and is the runtime role (e.g. a managed Cloud SQL `postgres` that cannot hold BYPASSRLS but owns every table), the owner is exempt from RLS as long as no `FORCE ROW LEVEL SECURITY` / policies apply, so enabling RLS is a harmless no-op - yet the hard failure permanently wedges the migration chain.

Setting `GBRAIN_RLS_ALLOW_NONBYPASS=1` downgrades the abort to a `RAISE WARNING` and proceeds with the `ALTER`. **Unset (the default) is unchanged fail-closed behavior.**

## Why it's safe

- Default preserved: with the env var unset, all four migrations still emit `RAISE EXCEPTION`. The existing structural test (`test/migrate.test.ts` - "fails loudly on non-BYPASSRLS roles instead of silently bumping version") stays green.
- One env var, read at module load, flips all four migrations together.
- Precedent already in-tree: v32 already warns-and-skips OAuth-table RLS for the same non-BYPASSRLS case.

## Implementation

A single `rlsBypassGuard(label)` helper returns the WARNING-and-proceed statement when the env var is set, else the existing `RAISE EXCEPTION`. The four DO-blocks call it in place of the hardcoded exception. No runner or connection changes.

### DIFF
--- a/src/core/migrate.ts
+++ b/src/core/migrate.ts
@@ -154,6 +154,33 @@ async function dropInvalidConcurrentIndex(
   return isInvalid;
 }
 
+/**
+ * Opt-in escape hatch for the RLS-enable migrations (v24/v29/v31/v35).
+ *
+ * Enabling RLS from a session that lacks BYPASSRLS would lock a NON-owning role
+ * out of its own data, so these migrations fail closed by default: a role
+ * without BYPASSRLS gets a RAISE EXCEPTION and the migration aborts (the runner
+ * leaves schema_version unchanged, so it retries cleanly once the role is
+ * fixed). That default is correct for the common managed-Postgres split where
+ * the migrating role and the runtime app role differ.
+ *
+ * On a single-principal deployment the tradeoff inverts: when the migrating
+ * role also OWNS the tables and is the runtime role (e.g. a managed Cloud SQL
+ * `postgres` that cannot hold BYPASSRLS but owns every table), the owner is
+ * exempt from RLS as long as no FORCE ROW LEVEL SECURITY / policies apply, so
+ * enabling RLS is a harmless no-op and the hard failure only wedges the
+ * migration chain. Set GBRAIN_RLS_ALLOW_NONBYPASS=1 there to downgrade the
+ * abort to a WARNING and PROCEED with the ALTER. Unset (the default) is
+ * unchanged fail-closed behavior. Read at module load, so one env var flips all
+ * four migrations together. Compare v32, which already warns-and-skips
+ * OAuth-table RLS for the same non-BYPASSRLS case.
+ */
+function rlsBypassGuard(label: string): string {
+  return process.env.GBRAIN_RLS_ALLOW_NONBYPASS === '1'
+    ? `RAISE WARNING '${label}: role % lacks BYPASSRLS — enabling RLS anyway (GBRAIN_RLS_ALLOW_NONBYPASS opt-in; safe only when this role owns the tables and no FORCE ROW LEVEL SECURITY / policies apply).', current_user;`
+    : `RAISE EXCEPTION '${label}: role % does not have BYPASSRLS privilege — cannot enable RLS safely. Re-run as postgres (or another BYPASSRLS role), or set GBRAIN_RLS_ALLOW_NONBYPASS=1 if this role owns the tables. The migration will retry automatically on the next initSchema call.', current_user;`;
+}
+
 // Migrations are embedded here, not loaded from files.
 // Add new migrations at the end. Never modify existing ones.
 // Exported for tests that structurally assert migration contents (e.g., "v9 must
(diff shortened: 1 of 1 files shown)


time-attack:

The mechanism is sound and I verified the dangerous claim end-to-end on real Postgres 16 using the PR's own generated SQL: default stays fail-closed (RAISE EXCEPTION, no RLS applied, clean retry), the opt-in is strictly gated on GBRAIN_RLS_ALLOW_NONBYPASS === '1', and after the opt-in enables RLS the non-BYPASSRLS OWNER retains full access to its own data (no FORCE ROW LEVEL SECURITY anywhere in src/; live SELECT returned the row) while a policy-less non-owner is denied — exactly the intended anon posture. All four fail-closed sites are covered with no missed siblings (schema.sql/schema-embedded.ts already warn-and-skip; v32 precedent confirmed at migrate.ts:1585). But it ships with zero committed tests for the new path (reverting the diff leaves every test green — discrimination-test fail), the suite breaks on any machine that exports the knob (4 pre-existing structural tests in test/migrate.test.ts hard-assert the fail-closed variant), the new GBRAIN_* env knob is undocumented (docs/guides/rls-and-you.md is the obvious home and covers this exact scenario), the success NOTICE on the new path falsely logs 'role % has BYPASSRLS' (migrate.ts:981, 1239, 1494), and I verified the opt-in still wedges vanilla-PG non-superusers at v35's CREATE EVENT TRIGGER ('Must be superuser') — a boundary the PR body/docs never state. Plus process gates: no VERSION/package.json/CHANGELOG bump (branch 0.42.67.0 vs master 0.42.68.1, strictly-greater gate fails) and the PR title lacks the version-first prefix (IRON RULE).

### Discrimination test

The PR adds no tests, so the formal discrimination test fails by construction: reverting the diff leaves all 162 pre-existing tests in test/migrate.test.ts green (they assert the default fail-closed variant, which both old and new code emit with the env var unset). I separately proved the change IS live-wired: running the same structural suite with GBRAIN_RLS_ALLOW_NONBYPASS=1 flips 4 tests to failing because MIGRATIONS content genuinely changes to RAISE WARNING at module load — i.e., the code works but nothing committed pins the new behavior.

**Mergeability at review time:** MERGEABLE / CLEAN (gh pr view --json mergeable,mergeStateStatus; first query returned UNKNOWN, re-query after 5s returned MERGEABLE/CLEAN)

### Blockers

**1.** No test exercises the opt-in path: PR adds zero tests, so stubbing the old always-EXCEPTION behavior back keeps all 162 tests in test/migrate.test.ts green. Fix: a spawned-subprocess test (env var must be set before module load) asserting the RAISE WARNING variant appears in v24/v29/v31/v35 and RAISE EXCEPTION does not.

**2.** Suite breaks when the knob is exported: GBRAIN_RLS_ALLOW_NONBYPASS=1 bun test test/migrate.test.ts fails 4 pre-existing structural tests (test/migrate.test.ts:453-454 'fails loudly on non-BYPASSRLS roles', :1342-1346 v31 'gates RLS and fails loudly', v35 :545+ 'backfill is gated', one more). The structural tests should delete the env var in their preamble (they assert the DEFAULT posture).

**3.** New GBRAIN_* env knob undocumented — required by repo convention. Add to docs/guides/rls-and-you.md (which already documents the single-principal/managed-Postgres scenario) and the src/core/migrate.ts entry in docs/architecture/KEY_FILES.md.

**4.** Misleading log line on the exact path this PR creates: after the opt-in WARNING fires, the trailing success NOTICE still prints 'role % has BYPASSRLS' — false. src/core/migrate.ts:981 (v24), :1239 (v29), :1494 (v31). Reword to not assert BYPASSRLS.

**5.** Undocumented boundary (verified live on PG16): with the opt-in set, v35 still aborts for vanilla-Postgres non-superusers at CREATE EVENT TRIGGER ('Must be superuser to create an event trigger'). The escape hatch fully unwedges only roles that can create event triggers (Cloud SQL/RDS/Supabase master roles). Docs/PR body must state this or a vanilla-PG user sets the flag and wedges again with a different error.

**6.** Process gates: no VERSION/package.json/CHANGELOG bump — branch sits at 0.42.67.0 while master is 0.42.68.1 (CI version-gate requires strictly greater); PR title 'feat(migrate): opt-in RLS enable without BYPASSRLS' violates the version-first IRON RULE.

### Not verified

- Behavior on actual managed platforms (Cloud SQL/RDS/Supabase master roles) — only vanilla Docker postgres:16 tested; the claim that those roles can create event triggers is from platform docs, not reproduced
- v29/v31 opt-in variants not individually executed against Postgres (identical guard shape; v24 and v35 executed live)
- Full unit suite / e2e / bun run verify not run (shared-machine footprint rules); only typecheck + test/migrate.test.ts both env states
- Whether the CI version-gate blocks this PR at merge time (no version-gate check appeared in the PR's check rollup; the strictly-greater rule is documented for pushes)

- `gh pr view 3655 --json number,title,body,mergeable,mergeStateStatus,files (+ re-query for mergeable)`
- `gh pr diff 3655 (read every hunk)`
- `gh pr checks 3655 (all green incl. verify, serial-tests, 10 test shards)`
- `git fetch origin pull/3655/head:pr-3655 && git checkout pr-3655 (in isolated worktree)`
- `cat VERSION (0.42.67.0) vs git show origin/master:VERSION (0.42.68.1)`
- `grep FORCE ROW LEVEL SECURITY / GBRAIN_RLS_ALLOW_NONBYPASS / RAISE EXCEPTION.*BYPASSRLS across src/ docs/ CLAUDE.md (sibling-caller + docs check)`
- `Read migrate.ts v24/v29/v31/v32/v35 blocks + runner (no checksums), test/migrate.test.ts assertions, docs/guides/rls-and-you.md, src/schema.sql:1387-1440`
- `bun run typecheck > file (EXIT=0)`
- `bun test test/migrate.test.ts > file (EXIT=0, 162 pass)`
- `GBRAIN_RLS_ALLOW_NONBYPASS=1 bun test test/migrate.test.ts > file (EXIT=1, 4 fail — proves module-load wiring, exposes env-sensitivity)`
- `docker run postgres:16; created NOSUPERUSER NOBYPASSRLS owner role 'app' + peer role + appdb`
- `bun -e extract of the PR's actual MIGRATIONS v24 SQL under both env states (EXCEPTION vs WARNING variants confirmed)`
- `psql as app: default v24 SQL → ERROR + relrowsecurity=f (fail-closed, no partial apply)`
- `psql as app: opt-in v24 SQL → WARNING + proceed, relrowsecurity=t on all tables, owner SELECT returns row (no lockout), peer SELECT returns 0 rows (deny-by-default)`
- `psql as app: opt-in v35 SQL → passes bypass guard but aborts at CREATE EVENT TRIGGER 'Must be superuser' (documented boundary finding)`
- `docker rm -f pr3655-pg (teardown)`

</details>

---

time-attack:
Closing. This is the third pass at the same security-posture relaxation — #3283 was closed on the grounds that on a managed-Postgres split (owning migrator role + separate limited app role) proceeding enables RLS with no policies and the app then reads zero rows everywhere. `GBRAIN_RLS_ALLOW_NONBYPASS=1` doesn't change that: gbrain still enforces nothing, the operator just promises single-principal ownership, and the silent total-retrieval-failure mode is one env var away. The current fail-closed `RAISE EXCEPTION` is the intended design. Thanks for the effort, but we're not taking this one.

PR #3659, opened 2026-07-30
v0.42.67.1 fix(facts): preserve remote facts fence writes (#2044)

### LINKED ISSUE #2044: Remote `extract_facts` fence-write is not idempotent — 2nd deposit to an already-fenced page throws `idx_facts_fence_key` duplicate and rolls back the whole batch

javieraldape:
## Summary

Fixes #2044.

This patch makes remote Facts fence writes idempotent across the privacy boundary:

- Preserves an existing Facts fence when an untrusted remote round-trip submits an empty or missing fence for a page that still has canonical fence rows.
- Makes duplicate `(source_id, source_markdown_slug, row_num)` fact inserts no-op instead of rolling back the surrounding batch.
- Applies the insert behavior consistently to Postgres and PGLite.

## Tests

- `bun test --timeout 300000 test/insert-facts-batch.test.ts test/e2e/system-of-record-invariant.test.ts`
- `DATABASE_URL=postgresql://postgres:postgres@localhost:5434/gbrain_test bun test --timeout 300000 test/e2e/schema-drift.test.ts test/e2e/engine-parity.test.ts -t "stale-page extraction methods|schema drift"`
- `bash scripts/check-privacy.sh && bash scripts/check-jsonb-pattern.sh && bash scripts/check-source-id-projection.sh`

## Local Gate Note

A full local `bun run verify` did not complete because `tsc --noEmit` timed out without diagnostics. The same timeout reproduced on a clean `upstream/master` worktree, so it does not appear to be introduced by this diff.


time-attack:
**Merging this closes #1867** (severity **p2**).

#1867 was one of 58 issues carrying a `p3` label. All 58 were re-triaged by reproducing them against current master, and **the label was wrong on 56 of them** — the corrected distribution is 4 p0, 40 p1, 12 p2, and 2 genuinely p3. #1867 came out at **p2**, and this PR is what fixes it.

Two things follow from that:

- Add `Closes #1867` to the PR description if you want the link wired up — a closing keyword only fires from the body or a commit message, not from a comment like this one.

The reproduction and evidence are on #1867.

time-attack:

The core fix is real and correctly aimed at #2044: ON CONFLICT DO NOTHING on the v51 fence key (both engines, predicate matches the partial index) stops a duplicate deposit from rolling back the whole batch, and the import-file guard stops a remote round-trip from truncating a facts fence the remote never saw. Discrimination test passed: with the three src files stubbed back to merge-base, exactly the two new/changed tests fail. Trust boundary holds: put_page threads ctx.remote !== false fail-closed into opts.remote, the guard only re-injects server-side existing content (unbalanced-fence crafting is blocked by the warnings gate), put_page's response carries no body so the preserved private fence never leaks to the untrusted caller, and remote get_page still strips. Hash is computed after re-injection so repeat puts converge to idempotent skips. But four concrete problems block merge as-is: (1) CONFIRMED by a live probe — the preservation guard requires the incoming fence to have ZERO facts (src/core/import-file.ts:583), so the documented remote round-trip of a mixed world+private page (the majority shape, since deposits default to visibility:world) still silently destroys the private rows and recreates the exact fence/DB drift that feeds #2044's collisions, which the ON CONFLICT change now converts into silent skips (markdown row N = new fact, DB row N = stale private fact until the next reconcile); (2) the insertFacts API contract in src/core/engine.ts:1775-1787 still promises atomic rollback on constraint violation AND ids returned in input-order for fence-row→DB-id correlation — both now false (ids shorten on skips); the identical stale comment sits directly above the new ON CONFLICT in both engines. No current caller zips positionally (all 6 call sites checked), but the documented contract now invites misalignment; (3) the postgres-engine.ts hunk ships untested: engine-parity.test.ts has zero insertFacts coverage and the author's listed real-Postgres runs (-t "stale-page extraction methods|schema drift") never execute the new SQL through postgres.js — CLAUDE.md's parity invariant requires a pin for a new SQL shape; (4) version gate: no VERSION/package.json/CHANGELOG bump in the diff, PR branch VERSION 0.42.67.0 and title version v0.42.67.1 are both below master's 0.42.68.1, failing the strictly-greater CI version-gate. All are fixable: widen the guard to preserve stripped non-world rows (merge by row_num, refusing remote edits to non-world rows), rewrite the three stale contract comments, add an insertFacts conflict-skip case to engine-parity.test.ts, rebase and re-bump the trio with a fresh title.

### Discrimination test

Reverted src/core/import-file.ts, src/core/pglite-engine.ts, src/core/postgres-engine.ts to merge-base c6dc0adf while keeping the PR's test files, then ran bun test --timeout 300000 on both test files: exactly the two behavior-pinning tests failed ('v51 partial UNIQUE index collision skips duplicate rows without rolling back the batch' and 'remote put_page round-trip preserves an existing private-only facts fence'), 16 others passed. Restored PR src: all 18 pass. The tests genuinely discriminate the new behavior — for the PGLite and import-file changes; the postgres-engine hunk has no test that exercises it (blocker 3).

**Mergeability at review time:** mergeable: MERGEABLE, mergeStateStatus: CLEAN

### Blockers

**1.** CONFIRMED: mixed-visibility pages still lose private facts on the documented remote round-trip — src/core/import-file.ts:583 guard fires only when incomingFacts.facts.length === 0; probe with one world + one private row showed PRIVATE row destroyed, only WORLD row survives (this is the majority page shape and the same drift creator behind #2044's collisions)

**2.** Stale API contract: src/core/engine.ts:1775-1787 still documents batch rollback on constraint violation and ids-in-input-order fence-row→DB-id correlation — both false after ON CONFLICT DO NOTHING (ids array shortens on skips); same wrong comment retained directly above the new SQL in pglite-engine.ts (~4259) and postgres-engine.ts (~4440)

**3.** postgres-engine.ts ON CONFLICT hunk is untested: test/e2e/engine-parity.test.ts has zero insertFacts coverage and the PR's listed Postgres test filter never executes this path — violates the CLAUDE.md engine-parity pin requirement for new SQL shapes

**4.** Version gate: no VERSION/package.json/CHANGELOG bump; branch VERSION 0.42.67.0 and title v0.42.67.1 are below master's 0.42.68.1 — fails the strictly-greater CI version-gate; needs rebase + trio re-bump + retitle

### Not verified

- postgres-engine.ts ON CONFLICT behavior against real Postgres via postgres.js — no test exists for it (blocker 3) and I did not spin up Docker per footprint rules (the pg container on :5434 belongs to another agent's review); PGLite (real Postgres WASM) verified the identical statement, so syntax risk is low but driver-path behavior is untested
- Author's claimed DATABASE_URL runs of schema-drift/engine-parity subsets (taken at face value; their filter provably does not reach insertFacts)
- Full unit suite / bun run verify (skipped per footprint rules; only the 10 test files adjacent to the diff were run — 160 tests total, all pass)

- `gh pr view 3659 --json number,title,body,mergeable,mergeStateStatus,headRefName,author,files`
- `gh pr view 3659 --json mergeable,mergeStateStatus (re-poll: MERGEABLE/CLEAN)`
- `gh issue view 2044`
- `gh pr diff 3659 (read every hunk)`
- `git fetch origin master pull/3659/head:pr-3659 && git checkout pr-3659`
- `cat VERSION (0.42.67.0) vs git show origin/master:VERSION (0.42.68.1)`
- `bun install`
- `bun test test/insert-facts-batch.test.ts test/e2e/system-of-record-invariant.test.ts (1 fail: pre-existing rebuild test exceeds bun's 5s default timeout)`
- `bun test --timeout 300000 test/e2e/system-of-record-invariant.test.ts (5 pass — timeout flake, not a regression)`
- `git checkout c6dc0adf -- src/core/{import-file,pglite-engine,postgres-engine}.ts && bun test --timeout 300000 <PR test files> (discrimination: exactly the 2 new tests fail) && git checkout pr-3659 -- <same>`
- `bun run typecheck (EXIT=0, contradicts author's timeout note)`
- `bash scripts/check-jsonb-pattern.sh; node scripts/check-jsonb-params.mjs; bash scripts/check-privacy.sh (all EXIT=0)`
- `grep sweep of all 6 insertFacts call sites (extract-facts.ts, fence-write.ts, backstop.ts, extract-conversation-facts.ts x3, longmemeval/extract.ts) — none zips ids positionally`
- `grep insertFacts test/e2e/engine-parity.test.ts (0 hits)`
- `bun test --timeout 300000 test/embedding-dim-check-facts.test.ts test/fence-write.test.ts test/facts-reconcile-protect-cli.test.ts test/facts-fence.test.ts (74 pass)`
- `bun test --timeout 300000 test/import-file.test.ts test/import-file-content-sanity.test.ts test/put-page-empty-guard.test.ts test/put-page-provenance.test.ts (68 pass)`
- `scratch probe test: mixed world+private fence remote round-trip → PRIVATE row destroyed (then deleted the scratch file; worktree clean)`
- `read put_page handler end-to-end in src/core/operations.ts:853-1225 (trust threading + response payload)`

</details>

---

Issue #361, opened 2026-04-23
Autopilot minion_jobs UPDATE can leave pgbouncer backend in ClientRead forever, holding RowExclusiveLock for days


orendi84:
## Summary

On Supabase Micro with the pooler (transaction mode, port 5432), an autopilot cycle's `UPDATE minion_jobs SET lock_until = now() + ($1::double precision * interval '1 millisecond'), updated_at = now()` occasionally leaves a backend session stuck in `state='active'`, `wait_event='ClientRead'`, holding a `RowExclusiveLock` on `minion_jobs` indefinitely. Seen twice on my instance: first at 22h age, then again at 23h51m age, blocking any subsequent `ALTER TABLE minion_jobs` (e.g. migration 15).

## Observed state (from `pg_stat_activity`)

```
pid=213932 state=active wait=Client/ClientRead age=23:51:40
query=UPDATE minion_jobs SET lock_until = now() + ($1::double precision * interval '1 millisecond'), updated_at = now()
```

Held `RowExclusiveLock` on `minion_jobs`. Not naturally released; had to `pg_terminate_backend()` manually.

## Suspected pattern

1. Autopilot cycle (or a worker) issues a lock-refresh UPDATE on `minion_jobs`.
2. The postgres.js client dies (launchctl reload, SIGKILL, crash) before COMMIT/ROLLBACK.
3. Pgbouncer keeps the upstream Postgres backend alive but the client side is gone, so the backend sits in ClientRead waiting for the next command that never arrives.
4. On Supabase Micro there's no default `idle_in_transaction_session_timeout` for pooler-held sessions, so these never time out on their own.
5. Autovacuum can't clean them up because the session holds active locks, not idle.

## Impact

- Blocks schema migrations on the same table (reproduced: migration 15's ALTER hung until I terminated the backend).
- Burns a connection slot on Micro (default 10 total from the pooler, smaller budget than it looks).
- Dead weight that accumulates over time if autopilot restarts enough.

## Suggested mitigations (any one helps; ideally #1)

1. **Set session-level timeouts in postgres.js on connect.** `idle_in_transaction_session_timeout = '2min'` + `statement_timeout = '5min'` as the first statements after establishing a connection. Postgres.js exposes this via the `connection.application_name` / `connection` hook. This ensures an abandoned in-flight transaction gets killed on the server side even when the client vanishes.
[...]

PR #3666, opened 2026-07-30
fix(postgres): cancel timed-out health checks

MikeHathaway:
## Summary

- Cancel the live `SELECT 1` when the `/health` deadline expires, instead of returning 503 while the losing query remains active or queued.
- Preserve the existing liveness-only response body, 3-second timeout, error envelope, and timer cleanup.
- Keep the change scoped to the health-cancellation fix requested by the maintainer.

## Review feedback addressed

- [x] Rebased onto the current `master` integration wave.
- [x] Removed the bundled `max_pipeline=1` default and all related tests/docs.
- [x] Removed the connection tracer and its dependency on postgres.js's undocumented `onexecute` hook.
- [x] Retained only the explicitly requested health-cancel change.

## Root cause and fix

`probeLiveness` previously raced `SELECT 1` against a timer. When the timer won, the HTTP request returned 503, but the database query continued running or waiting for a pool slot. Repeated orchestrator probes could therefore compound pool saturation.

The probe now executes through `BrainEngine.executeRaw(..., { signal })` and aborts the signal when the deadline fires. Postgres forwards that abort to postgres.js query cancellation, including queries waiting in the client queue. Timeout classification uses `controller.signal.aborted` as its single source of truth, so synchronous cancellation rejection cannot change the public response from the expected timeout 503.

PGLite has no external connection pool or kernel-level cancellation; it retains the engine's documented best-effort behavior of returning promptly and discarding the eventual query result.

## Verification

- `bun run verify`: 54/54 checks passed
- Focused and adjacent unit tests: 26 passed, 77 assertions
- Synchronous-abort race stress: 1,000/1,000 timeout classifications passed
- Disposable PostgreSQL, active query: cancelled; sole pool slot reusable in 3 ms
- Disposable PostgreSQL, client-queued query: cancelled; sole pool slot reusable in 2 ms
- `test/e2e/serve-http-oauth.test.ts`: 38 passed, 178 assertions
- Version consistency: `VERSION`, `package.json`, and the top `CHANGELOG.md` entry remain `0.46.25.0`

GitHub Actions require maintainer approval because this PR originates from a fork.

Related: #361


time-attack:

Real bug, root-cause fix, quantified end-to-end on real PgBouncer 1.25.2 transaction mode. Before: every timed-out /health probe leaves its SELECT 1 alive in the postgres.js pool — 20 timed-out probes = 20 zombie queries still queued (measured with the PR's own trace instrumentation). After: all 20 cancelled with SQLSTATE 57014, 0 zombies, and the pool slot is immediately reusable. The PR's claims about postgres.js internals all check out against the vendored 3.4.9 source: unsafe() spreads options so the undocumented onexecute hook IS honored and its `return true` correctly preserves the pipelining gate (connection.js:177); max_pipeline=0 would short-circuit sql.begin's internal onexecute and break transactions (index.js:242,299), so the =1 floor is right — I verified 10/10 concurrent sql.begin transactions commit through the pooler at max_pipeline=1; cancel() on a queued query removes it from the global queue and rejects immediately (index.js:350). Sibling-caller audit clean: all 3 postgres() pool-creation sites (db.connect, connection-manager readPool, postgres-engine connect) got resolveMaxPipeline; the ddl pool is correctly excluded (hardcodes prepare:true, session-mode direct URL); runUnsafe's only two callers (executeRaw/executeRawDirect) both updated; doctor's resolvePrepare use is URL-inspection only. Engine parity holds — PGLite executeRaw already had opts.signal (v0.41.18.0, documented best-effort gap, no regression). Trace is opt-in (default off), and the emitted JSONL verified free of SQL/params/URLs; the API structurally cannot accept them. Both new GBRAIN_* knobs documented in docs/guides/minions-deployment-snippets/gbrain.env.example. Throughput cost of max_pipeline=1 on prepare=false pools: ~14% on localhost (58ms vs 51ms per 400 concurrent mixed queries), overridable via GBRAIN_MAX_PIPELINE and ?max_pipeline=. PR's "3 tests / 207 assertions" e2e claim reproduced exactly. Merge-time metadata for the captain: title lacks the version-first prefix and the branch carries no VERSION/CHANGELOG bump (trio internally consistent at 0.42.67.0 = master) — consistent with the repo's batch-release flow (cf. v0.42.66.0 chore(release) batching 54 fixes); retitle with the allocated version at merge (gh pr edit 3666 --title "vX.Y.Z.W fix(postgres): ...").

### Discrimination test

Stubbed the old behavior back into probeLiveness (engine.executeRaw('SELECT 1') with no AbortSignal, no controller.abort()) and re-ran test/serve-http-health.test.ts: exactly the new "timeout path: aborts the live query before returning 503" test fails (querySignal?.aborted / abortObserved assertions), 7/8 pass. Same stub re-run against real PgBouncer flips the quantified outcome from 20/20 cancelled to 20/20 zombies — the e2e queued-cancellation test would fail too. Tests discriminate.

**Mergeability at review time:** {"mergeStateStatus":"UNSTABLE","mergeable":"MERGEABLE"} (UNSTABLE = fork PR awaiting maintainer CI approval; branch is 1 commit behind master, typecheck clean on branch)

### Refutations

**Refuter 1:** The MERGE recommendation is REFUTED by a confirmed data-loss-grade defect in the PR's max_pipeline=1 default that the first reviewer's verification was structurally unable to catch. Their claims that I re-verified all held: typecheck clean (EXIT=0), the 3 new test files pass (16/16), the discrimination test on the health fix is genuine (stubbing the old no-abort probeLiveness back fails exactly the new abort test, 7/8 pass), PGLite parity holds (pglite-engine.ts executeRaw has accepted opts.signal since v0.41.18.0), the trace API is opt-in and structurally cannot receive SQL/params/URLs, and both GBRAIN_* knobs are documented. But their headline transaction check — '10/10 concurrent sql.begin transactions commit at max_pipeline=1' — only exercised BEGIN as the ACTIVE statement on a connection (sent.length=0, gate 0<1 passes). postgres.js's pipeline gate (node_modules/postgres/src/connection.js:176-177) is `sent.length < max_pipeline && (!q.options.onexecute || q.options.onexecute(connection))` — a short-circuiting &&. When the pool is saturated and a sql.begin BEGIN is dispatched onto a busy connection (index.js handler falls through to `busy` at line 339), the BEGIN is pushed to `sent` (length 1), `1 < 1` is false, and onexecute — the ONLY mechanism by which sql.begin captures its reserved connection (index.js:242,299) — is silently skipped. I reproduced this on real Postgres 16 (Docker) twice: (a) driver-level with the PR's option, and (b) end-to-end through gbrain's own PostgresEngine.connect with ONLY the PR's defaults (?prepare=false URL, no explicit max_pipeline — pool options confirmed {prepare:false, max_pipeline:1}). Under one in-flight query, sql.begin on the engine pool throws `TypeError: undefined is not an object (evaluating 'connection.onclose = reject')`; the BEGIN had already been written to the socket, leaving an orphaned open transaction on the pooled connection (measured 1 backend 'idle in transaction'); a subsequent INSERT on the pool reported success but was INVISIBLE to a second connection — acknowledged writes silently swallowed into the orphaned transaction, rolled back when gbrain's own idle_in_transaction_session_timeout kills the backend. The control run with the pre-PR pipeline default passed cleanly (begin OK, insert visible, 0 orphans), so the regression is introduced solely by this PR. Blast radius: gbrain calls sql.begin on the affected pool in 8+ places (postgres-engine.ts:264 withScopedReadTransaction alwaysTransaction — the SEARCH paths; 1007, 4324, 4350, 4453, 5190, 5699; db.ts:377), the default applies to every prepare=false deployment including port-6543 Supabase auto-detect, and pool saturation is the routine state during bulk sync/embed (the shipped env example suggests GBRAIN_POOL_SIZE=2) — i.e., the PR's defense-in-depth breaks transactions and silently loses writes precisely in the regime it claims to defend, and through a transaction pooler it pins pooler backends idle-in-transaction, a worse version of the #361 disease. Secondary findings: the max_pipeline wiring at connection-manager.ts:319 and db.ts:317 is pinned by no test at all (only the postgres-engine site is asserted, and only in the env-gated e2e); title lacks the version-first prefix and branch VERSION (0.42.67.0) is below master's (0.42.68.1) with no CHANGELOG entry — arguably covered by the repo's batch-release flow, but failing the stated gates. Correct disposition is NEEDS_CHANGES, not CLOSE: the health-probe AbortSignal fix is real, root-caused, and properly discriminated, and the trace layer is clean — but the max_pipeline=1 default must be dropped or reworked (any finite bound N breaks sql.begin whenever BEGIN is the Nth pipelined statement; N=1 makes that routine). Unverified for footprint reasons (moot given the confirmed defect): full unit suite, the PgBouncer e2e rerun, and the throughput numbers. Cleanup done: Docker container removed, scratch repro files deleted, worktree restored clean.

*Missed by the first reviewer:* max_pipeline=1 default (src/core/db.ts:334 resolveMaxPipeline, wired at postgres-engine.ts:317, connection-manager.ts:319, db.ts:317) breaks sql.begin under pool saturation: postgres.js's gate `sent.length < max_pipeline && (!q.options.onexecute || q.options.onexecute(connection))` (postgres/src/connection.js:176-177) short-circuits before invoking sql.begin's onexecute whenever the BEGIN is pipelined behind an active query, so begin() never captures its connection. Reproduced on real Postgres 16 through gbrain's own engine with only PR defaults (?prepare=false, poolSize=1, one in-flight pg_sleep): sql.begin throws `TypeError: undefined is not an object (evaluating 'connection.onclose = reject')` (postgres/src/index.js:244), the already-sent BEGIN orphans an open transaction on the pooled connection (1 backend 'idle in transaction'), and a subsequent INSERT on the pool reports success while remaining invisible to other connections — silent data loss when the orphaned transaction is later rolled back by gbrain's own idle_in_transaction_session_timeout. Affects every sql.begin site on the main pool (postgres-engine.ts:264 [search paths], 1007, 4324, 4350, 4453, 5190, 5699; db.ts:377) for all prepare=false deployments (incl. Supabase 6543 auto-detect). Control with pre-PR pipeline default: begin OK, 0 orphans. Fix: drop the =1 default (keep the discriminated health-abort fix), or bound pipelining only on pools that never run sql.begin.

**Refuter 2:** MERGE is refuted; verdict should be NEEDS_CHANGES. The health-probe AbortSignal half of the PR is genuinely good (I confirmed the reviewer's discrimination logic and the wiring: old probeLiveness already routed through engine.executeRaw via sqlQueryForEngine, so the change is equivalent-plus-cancellation, and the trace API is opt-in/off by default with unsafe options spread verified at index.js:119-127). But the second half — defaulting max_pipeline=1 on every prepare=false pool — is a demonstrated hard correctness regression on the exact deployment shape the PR targets: transaction-pooler (Supabase :6543 auto-detect) deployments under load. I reproduced it twice against real Postgres 16: once at the driver level isolating the single knob (default OK / =1 UNSAFE_TRANSACTION), and once end-to-end through PostgresEngine.connect + engine.transaction() on the PR branch (FAILED with the PR default, OK with GBRAIN_MAX_PIPELINE=100 as control). The regression converts "bounded queuing under saturation" into deterministic hard failures of search (all four arms use sql.begin via alwaysTransaction:true) and fact writes at trivially low concurrency (in-flight >= poolSize; docs recommend poolSize 2). The first reviewer ran only the happy path for transactions (concurrent begins on an idle pool) and the PR's own tests structurally cannot catch it (poolSize:1 disables postgres.js's UNSAFE_TRANSACTION guard; no test mixes begin with a saturated pool). Fix path: drop the =1 default (keep it URL/env opt-in) or make the bound transaction-safe; the abort fix can ship alone. Secondary, already-known gate issues: title lacks the version-first prefix and VERSION (0.42.67.0) equals master's, though repo history shows a batch-release flow. Unverified (footprint rules): full unit suite, the PR's pgbouncer e2e suite (defect is client-side dispatch, no pooler needed), and whether the failed BEGIN additionally leaks an idle-in-transaction connection back into the open pool (connection.js:583 onopen path suggests it may — worth checking during the fix). Docker container torn down; scratch files removed; worktree clean.

*Missed by the first reviewer:* The max_pipeline=1 default breaks ALL sql.begin transactions under pool saturation with a hard UNSAFE_TRANSACTION error. Mechanism (vendored postgres.js 3.4.9): connection.js:172-177 returns `write(...) && ... && sent.length < max_pipeline && (!q.options.onexecute || q.options.onexecute(connection))` — the max_pipeline clause short-circuits BEFORE onexecute. sql.begin (index.js:242,299) relies on that onexecute call as the ONLY mechanism that sets connection.reserved. When in-flight queries >= pool size, the BEGIN is dispatched onto a busy connection as the 2nd pipelined statement (sent.length=1), so with max_pipeline=1 the gate short-circuits, onexecute never runs, the already-written BEGIN executes server-side, and CommandComplete (connection.js:605: `result.command === 'BEGIN' && max !== 1 && !connection.reserved`) rejects the transaction with UNSAFE_TRANSACTION. Empirically confirmed on real Postgres 16 (Docker): (1) driver-level A/B — max:2, prepare:false, both connections saturated with pg_sleep, then sql.begin: driver default OK, max_pipeline=1 FAILED code=UNSAFE_TRANSACTION; (2) end-to-end through the PR's own code — PostgresEngine.connect('...?prepare=false', poolSize:2) resolves max_pipeline=1, two executeRaw pg_sleep blockers, then engine.transaction() FAILED code=UNSAFE_TRANSACTION; control with GBRAIN_MAX_PIPELINE=100 on the same scenario: OK. Blast radius: prepare=false is AUTO-DETECTED for port 6543 (src/core/db.ts AUTO_DETECT_PORTS) — every Supabase transaction-pooler deployment. The capped pool runs sql.begin in all four search arms (withScopedReadTransaction alwaysTransaction:true — postgres-engine.ts:1847,1993,2131,2314 → searchKeyword/searchTitles/searchKeywordChunks/searchVector), insertFact/insertFacts (4324,4350,4453), supersedeTake (5190), setPageAliases (5699), and public engine.transaction() (1005). With the repo-recommended GBRAIN_POOL_SIZE=2 (in the very env example this PR edits), 3 concurrent queries where one is a search or fact-write now hard-fails where it previously queued and completed — and pool saturation is the normal state here (pace-mode exists because backfills saturate the pool). The first reviewer's 10/10 sql.begin check ran begins on an idle pool (each got a fresh connection, sent.length=0, onexecute fires); the failure requires mixed load. The PR's own e2e uses poolSize:1, where max=1 disables the connection.js:605 check entirely.

### Not verified

- Full unit suite / bun run verify / test:slow not run (shared-machine footprint rules); PR claims 33/33 verify and full-suite results — only the changed-surface + adjacent files were run here (74 tests total, all pass)
- Supavisor-specific behavior: only standard PgBouncer 1.25.2 tested; the PR itself states the #1033 defect did not reproduce on standard PgBouncer, so the max_pipeline=1 default is verified as bounded-and-live defense in depth, not as a Supavisor repro fix
- Throughput impact on high-latency (non-localhost) links: the ~14% localhost delta likely grows with RTT; not measured over a WAN
- PR body's '519 lifecycle events' trace run and full-E2E-failure triage (8 pre-existing on master) taken on faith; my smaller-scale trace runs were consistent with the claim

- `gh pr view 3666 --json mergeable,mergeStateStatus (UNKNOWN first, then MERGEABLE/UNSTABLE)`
- `gh pr diff 3666 (all 991 lines read)`
- `git fetch origin pull/3666/head:pr-3666 && git checkout pr-3666; bun install`
- `bun test test/serve-http-health.test.ts test/resolve-max-pipeline.serial.test.ts test/connection-trace.serial.test.ts (16 pass / 0 fail)`
- `discrimination stub of probeLiveness + bun test test/serve-http-health.test.ts (EXIT=1, 7 pass / 1 fail: the abort test)`
- `read node_modules/postgres/src/{index,connection}.js 3.4.9: verified unsafe options spread, onexecute gate at connection.js:177, max_pipeline semantics, cancel() queued-query removal at index.js:350`
- `docker: postgres:16-alpine + edoburu/pgbouncer 1.25.2 pool_mode=transaction on :5434/:6543`
- `GBRAIN_PGBOUNCER_URL=... GBRAIN_PGBOUNCER_DIRECT_URL=... bun test test/e2e/pgbouncer-pipeline-liveness.test.ts (3 pass, 207 assertions — matches PR claim)`
- `quantify-leak.ts with GBRAIN_CONNECTION_TRACE=1: PR branch {probes_503:20, cancelled_query_error:20, zombies:0, codes:[57014]}; old-behavior stub {probes_503:20, cancelled:0, zombies:20}`
- `tx-and-throughput.ts: 10/10 concurrent sql.begin commits at max_pipeline=1 through pgbouncer; 400 concurrent queries avg 58ms (pipeline=1) vs 51ms (pipeline=100)`
- `bun test test/connection-manager.serial.test.ts test/resolve-prepare.test.ts test/postgres-engine.test.ts (55 pass / 0 fail)`
- `grep audits: probeLiveness callers, resolvePrepare pool sites, runUnsafe callers, postgres() creation sites; gh issue view 361 (real, on-point: pgbouncer backend stuck in ClientRead holding RowExclusiveLock)`
- `spot-checked trace JSONL output for SQL/URL/credential leakage (clean)`
- `docker rm -f + network rm (torn down), scratch files removed, worktree clean`

</details>

---

time-attack:
Review finding (split requested): the health-check half is a clean merge-lane fix — `probeLiveness` races `SELECT 1` against a timer with no cancellation, and the signal-cancel plumbing it needs already exists in both engines. We'd take that ~20-line change today.

The blocker is the bundling: (a) a new global default `max_pipeline=1` for every `prepare=false` pool — this serializes statements per socket on exactly the pooler path gbrain's bulk work uses (embed/sync/lock-renewal, often at `GBRAIN_POOL_SIZE=2`), a plausible throughput regression traded for a driver defect your own PR body says did not reproduce; and (b) a ~128-line connection-trace subsystem hanging off postgres.js's undocumented `onexecute` internals, where a driver upgrade can silently stop pipelining since the hook must return true.

Please re-cut as: the health-cancel fix alone (merges fast); `max_pipeline` as env/URL opt-in with NO default change; the tracer as its own PR if at all.

time-attack:
Status update: reviewed in the 2026-08-18 full-backlog triage — still wanted, but not yet mergeable: Health-cancel half explicitly wanted by reviewer ('we'd take that ~20-line change today'), but the requested split from max_pipeline=1 default is unanswered. If you update it, it will be re-reviewed promptly; alternatively the fix may be adopted with credit in an upcoming fix wave. Thanks for the contribution and the patience.

time-attack:
Re-trialed against the current integration branch (post v0.46.23-25 waves): Conflicts vs wave: serve-http.ts, connection-manager.ts, db.ts, postgres-engine.ts (7 hunks; wave rewrote probeLiveness engine->SqlQuery). Not superseded: health-cancel still absent on wave; split undone. No red flags — a rebase along those lines puts this back in the active merge queue; re-trials run nightly, so a fresh push gets picked up automatically. Thank you!

MikeHathaway:
> Re-trialed against the current integration branch (post v0.46.23-25 waves): Conflicts vs wave: serve-http.ts, connection-manager.ts, db.ts, postgres-engine.ts (7 hunks; wave rewrote probeLiveness engine->SqlQuery). Not superseded: health-cancel still absent on wave; split undone. No red flags — a rebase along those lines puts this back in the active merge queue; re-trials run nightly, so a fresh push gets picked up automatically. Thank you!

I just updated the pr accordingly

garrytan:

Issue #1712, opened 2026-05-31
CLI `query --source __all__` returns "No results": sentinel rejected by source-id regex, falls back to empty `default` source


AIScottK:
See trace below. Reproduced on 0.42.1.0.

## Summary
On a multi-source (federated) PGLite brain, `gbrain query "<q>" --source __all__` always returns **"No results"**, even though the same query against a specific source returns good hits. The documented cross-source sentinel `__all__` is silently downgraded to the empty `default` source.

## Version
gbrain 0.42.1.0 (git clone + `bun link`), engine: pglite, schema v111, embedding `openai:text-embedding-3-small`, 3 federated sources (none of them `default`).

## Reproduce
```
$ gbrain query "Who is Jeff Woodrow?" --source __all__ --no-expand
No results.

$ gbrain query "Who is Jeff Woodrow?" --source scott-brain --no-expand
[0.8765] people/jeff-woodrow -- # Jeff Woodrow
...
```

## Root cause
`--source __all__` is rejected by the source-id regex; the resulting throw is swallowed and the context falls back to the empty `default` source:

1. `src/core/source-resolver.ts:78-79` — `if (!SOURCE_ID_RE.test(explicit)) throw new Error(...Must match [a-z0-9-]{1,32}...)`. `__all__` contains underscores, so the regex rejects it.
2. `src/cli.ts` `makeContext` — `sourceId = await resolveSourceId(engine, explicit)` throws; `catch { sourceId = undefined }` swallows it (the comment intends the `undefined` → cross-source back-compat path), but the return then does `sourceId: sourceId ?? "default"`, overriding `undefined` back to `"default"`.
3. The query is scoped to `WHERE source_id = "default"`, which has 0 pages → "No results".

The operation-layer cross-source branch in `src/core/operations.ts` (`sourceIdParam === "__all__" ? {} : { sourceId }`) is never reached for CLI callers because `makeContext` corrupts the value first.

## Suggested fix
Special-case the sentinel before the regex check so it threads through to the operation handler, and stop `makeContext` from overriding the cross-source intent with `"default"`:
```ts
const explicit = (params.source as string | undefined) ?? null;
if (explicit === "__all__") {
  sourceId = "__all__"; // or a dedicated cross-source sentinel the op handler honors
} else {
  sourceId = await resolveSourceId(engine, explicit);
}
```

## Workaround
Pass an explicit single `--source <id>` and route per query at the app layer.

PR #3675, opened 2026-07-31
fix(export): thread sourceId filter via --source

alexey-metaengage:
Rebased onto fresh master and reworked against the review. Retitled from `feat` to `fix` per the review's own read: `PageFilters.sourceId` is already honored by both engines and `export` simply never threads it, so a multi-source export flattens by slug today.

**Stacked on #3755.** The cross-source leak found in review (`getTags` / `getRawData` unscoped in the export loop) is a master bug, present with or without this flag - `getTags` falls back to `source_id = 'default'` and `getRawData` applies no source predicate at all. It is fixed separately in #3755 with its own discriminating tests; this branch sits on top and no longer carries it. Please take #3755 first.

## Changes since the reviewed version

**1. The flag now resolves through the canonical chain (#1712 loud-fail).** `resolveSourceWithTier(engine, explicitSource)` replaces the bare `args.indexOf` read, so an unknown or malformed id is a hard error instead of a successful zero-page export, and the `__all__` sentinel spans every source rather than being passed through as a source id that matches nothing.

**2. Resolution stays EXPLICIT-ONLY - the resolver runs only when the flag was actually passed.** The review asked for `GBRAIN_SOURCE` / dotfile / path-match to be honored too. I've deliberately not done that, matching two existing commands that made the same call for the same reason:

- `src/commands/doctor.ts:4721` - *"EXPLICIT-ONLY by design - a raw flag parse, NOT resolveSourceWithTier. The tier resolver would pick a default source when `--source` is absent and silently scope a bare `gbrain doctor` to one source."*
- `src/commands/orphans.ts:226` - same reasoning, same wording.

Export has the identical hazard with worse consequences: its output is an archive, so an ambient dotfile narrowing a bare `gbrain export` drops pages from a backup silently, and the loss only surfaces at restore time. Happy to switch if you'd rather export match `sync`/`capture` instead - it's a one-line change - but the default-safety argument reads stronger here than anywhere else in the CLI.

**3. `--source` is documented** in the export help block (`src/cli.ts:2622`).

## Tests

Six tests in `test/storage-export.test.ts`. Discrimination verified by stubbing the resolver call back to the raw `args.indexOf` parse - three fail:

```
(fail) an unknown source id is a hard error, not a silent empty export
(fail) a malformed source id is rejected before any page query
(fail) --source __all__ spans every source rather than matching a literal id
```

The scoping tests (`--source` exports one source; bare export still exports all) discriminate against removing the `filters.sourceId` wire, as before.

One test does NOT discriminate and is a regression guard rather than a proof: *"an absent `--source` does not consult GBRAIN_SOURCE"* passes both before and after, because the raw parse also ignored the env var. It's there to fail if someone later moves the resolver call out of the `if (explicitSource)` branch.

13 pass / 0 fail in that file, `tsc --noEmit` clean.

## Notes

- `test/cli.test.ts` shows 8 pass / 10 fail here, identical on untouched master - pre-existing and environmental, not from this branch.
- `test/storage-export.test.ts` needs `--timeout 120000` locally; its `beforeAll` runs 120 PGLite migrations and blows the default 5s hook budget on Windows. Also reproduces on master.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -2619,6 +2619,7 @@ IMPORT/EXPORT
   export [--dir ./out/]              Export to markdown
   export --restore-only [--repo <p>] Restore missing supabase-only files
         [--type T] [--slug-prefix S] With optional filters
+        [--source ID]                Scope to one source (default: every source)
 
 FILES
   files list [slug]                  List stored files
--- a/src/commands/export.ts
+++ b/src/commands/export.ts
@@ -108,7 +108,11 @@ export async function runExport(engine: BrainEngine, args: string[]) {
   let exported = 0;
 
   for (const page of pages) {
-    const tags = await engine.getTags(page.slug);
+    // Slugs are unique per source, not brain-wide, so both sidecar reads are
+    // pinned to the page's own source. Unscoped, `getTags` falls back to
+    // `source_id = 'default'` and stamps the default source's tags onto a
+    // same-slug page from another source (dropping its real ones).
+    const tags = await engine.getTags(page.slug, { sourceId: page.source_id });
     const md = serializeMarkdown(
       page.frontmatter,
       page.compiled_truth,
@@ -120,8 +124,13 @@ export async function runExport(engine: BrainEngine, args: string[]) {
     mkdirSync(dirname(filePath), { recursive: true });
     writeFileSync(filePath, md);
 
-    // Export raw data as sidecar JSON
-    const rawData = await engine.getRawData(page.slug);
+    // Export raw data as sidecar JSON. Unscoped, this matches the slug in
+    // EVERY source and the loop below merges the rows into one sidecar keyed
+    // by `rd.source`, so another source's raw data silently overwrites this
+    // page's own on a key collision.
+    const rawData = await engine.getRawData(page.slug, undefined, {
+      sourceId: page.source_id,
+    });
     if (rawData.length > 0) {
       const slugParts = page.slug.split('/');
       const rawDir = join(outDir, ...slugParts.slice(0, -1), '.raw');
--- a/src/commands/export.ts
+++ b/src/commands/export.ts
@@ -5,7 +5,11 @@ import { serializeMarkdown } from '../core/markdown.ts';
 import { createProgress } from '../core/progress.ts';
 import { getCliOptions, cliOptsToProgressOptions } from '../core/cli-options.ts';
 import { loadStorageConfig, isDbOnly } from '../core/storage-config.ts';
-import { getDefaultSourcePath } from '../core/source-resolver.ts';
+import {
+  ALL_SOURCES,
+  getDefaultSourcePath,
+  resolveSourceWithTier,
+} from '../core/source-resolver.ts';
 import type { PageType } from '../core/types.ts';
 
 export async function runExport(engine: BrainEngine, args: string[]) {
@@ -21,7 +25,25 @@ export async function runExport(engine: BrainEngine, args: string[]) {
   const slugPrefixIdx = args.indexOf('--slug-prefix');
   const slugPrefix = slugPrefixIdx !== -1 ? args[slugPrefixIdx + 1] : undefined;
 
+  const sourceIdx = args.indexOf('--source');
+  const explicitSource = sourceIdx !== -1 ? args[sourceIdx + 1] : undefined;
+
   const restoreOnly = args.includes('--restore-only');
+
+  // EXPLICIT-ONLY, matching `doctor` and `orphans`: the flag is resolved
+  // through the canonical chain but ONLY when it was actually passed. Calling
+  // the resolver unconditionally would let GBRAIN_SOURCE or a `.gbrain-source`
+  // dotfile silently narrow a bare `gbrain export` to one source, which for a
+  // command whose output is a backup means silent data loss at restore time.
+  // Resolving the explicit value still buys the #1712 loud-fail on an unknown
+  // or malformed id, so a typo can't quietly export zero pages.
+  let sourceFilter: string | undefined;
+  if (explicitSource) {
+    const { source_id } = await resolveSourceWithTier(engine, explicitSource);
+    // `__all__` means "span every source" — the no-filter default, not a
+    // source literally named `__all__`.
+    sourceFilter = source_id === ALL_SOURCES ? undefined : source_id;
+  }
 
   // Resolution chain (D5): explicit --repo → typed sources.getDefault() →
   // hard-error for restore-only paths (never fall through to cwd).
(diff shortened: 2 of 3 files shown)


time-attack:

Right aim, real tests (discrimination test holds: stubbing the filters.sourceId wire fails the new scoping test), but the scoped export leaks cross-source data in the very loop the PR touches. Live repro: with slug notes/shared in both src-x and default (secret raw data + tag on default), `gbrain export --source src-x` writes default's raw data into src-x's sidecar JSON, stamps default's tag into the markdown, and drops src-x's own tag — because src/commands/export.ts:120 getTags(page.slug) unscoped hardcodes source_id='default' and :133 getRawData(page.slug) unscoped does a cross-source read. Fix is one line each using page.source_id (already on every Page row): getTags(page.slug, { sourceId: page.source_id }) and getRawData(page.slug, undefined, { sourceId: page.source_id }); add a colliding-slug test case (current tests use disjoint slugs so they cannot detect it). Also: no VERSION/package.json/CHANGELOG bump (VERSION would equal master's 0.42.67.0 — fails CI version-gate) and PR title lacks the vX.Y.Z.W prefix (IRON RULE). Minor: `--source <typo>` silently exports 0 pages exit 0, the #1712 bug class — validate the source id exists.

### Discrimination test

Stubbed the old behavior back (removed `if (sourceFilter) filters.sourceId = sourceFilter;` from src/commands/export.ts) and re-ran bun test test/storage-export.test.ts: 6 pass / 1 fail — 'export --source > exports only the named source' fails as it should (y-only.md gets written). Tests discriminate; restored PR code afterward (7/7 pass).

### Blockers

**1.** src/commands/export.ts:120 — getTags(page.slug) not source-scoped: scoped export writes the DEFAULT source's tags and drops the scoped source's own tags when the slug exists in both (getTags unscoped defaults to source_id='default'); reproduced live

**2.** src/commands/export.ts:133 — getRawData(page.slug) not source-scoped: `export --source src-x` writes the default source's raw_data secret into src-x's .raw sidecar JSON (cross-source data leak); reproduced live

**3.** No VERSION/package.json/CHANGELOG bump: merged as-is VERSION equals master's 0.42.67.0, which fails the CI version-gate (needs maintainer takeover via /ship or a bump)

**4.** PR title 'feat(export): add --source flag...' lacks the leading vMAJOR.MINOR.PATCH.MICRO version (IRON RULE)

### Not verified

- Full unit suite / bun run verify not run (footprint rule) — only the touched test file + typecheck
- Real Postgres (Docker) not exercised; leak repro ran on PGLite, but both engines' getTags/getRawData share the same unscoped-default semantics per source inspection
- Thin-client behavior of `gbrain export --source` not exercised (export is CLI_ONLY; dispatch inspection only)

- `gh pr view 3675 --json number,title,body,mergeable,mergeStateStatus,files (MERGEABLE/CLEAN)`
- `gh pr view 3675 --json statusCheckRollup (all SUCCESS)`
- `git show origin/master:VERSION (0.42.67.0)`
- `git fetch origin pull/3675/head:pr-3675 && git checkout pr-3675`
- `bun test test/storage-export.test.ts (7 pass / 0 fail)`
- `discrimination test: removed the filters.sourceId line, bun test test/storage-export.test.ts (6 pass / 1 fail — the scoping test), restored`
- `bun /tmp/pr3675_leak_repro.ts — live repro: raw sidecar leaks default-source secret=true, md carries default-only tag=true, md carries src-x's own tag=false`
- `grep of src/core/pglite-engine.ts getTags/getRawData, src/core/postgres-engine.ts listPages (parity confirmed for PageFilters.sourceId), src/cli.ts dispatch (CLI_ONLY export gets raw args, flag wired in real path)`

</details>

---

time-attack:
Review finding (blocking, but the gap is real): `PageFilters.sourceId` is already honored by both engines and `export` never threads it, so multi-source export genuinely flattens by slug — this is closer to a bug fix than the `feat` label suggests. The blocker is the flag plumbing: `args.indexOf('--source')` bypasses the canonical resolver chain (`resolveSourceWithTier`), so the new flag ignores `GBRAIN_SOURCE`, the `.gbrain-source` dotfile, and path-match — and skips the #1712 loud-fail on an invalid id. A typo'd `--source projct-a` silently exports 0 pages, which is the exact silent-failure class this repo hunts.

To unblock: route through the same source-resolution path the rest of the CLI uses (or at minimum `validateSourceId` + loud-fail on no-match), add the `--help` line, and a doc note beside the sibling `--type`/`--slug-prefix` flags.

time-attack:
Automated triage (2026-08-01): adds new product surface (new export flag), which is reserved for the maintainer under triage policy. Queued for maintainer product decision; no automated test or merge action taken.

alexey-metaengage:
Thanks for the review - both findings were worth acting on, and one of them turned out to be bigger than this PR. Rebased onto fresh master, retitled `feat` -> `fix` per your own read that `PageFilters.sourceId` is already honored by both engines and `export` simply never threads it.

**The cross-source leak is a master bug, not something this branch introduced.** Your repro is real, but the two calls it lands in are unscoped on master today, with or without `--source`:

- `src/commands/export.ts:111` - `getTags(page.slug)` with no opts, so both engines fall back to `source_id = 'default'` (`postgres-engine.ts:3786`, `pglite-engine.ts:3646`). Any non-default page exports the default source's tags and loses its own.
- `src/commands/export.ts:124` - `getRawData(page.slug)` with no opts hits the four-branch `else` at `postgres-engine.ts:4193`, which applies **no source predicate at all**. Every source's rows come back and the loop collapses them into one sidecar keyed by `rd.source`.

Split out to **#3755** with its own discriminating tests - reverting it fails them with the defect right there in the output (`title: other title` carrying `tags: - tag-default`). This branch is stacked on top and no longer carries that diff. Worth taking #3755 first; it stands alone as a fix on a multi-source brain even if `--source` never lands.

**On the flag plumbing: adopted the loud-fail, declined the ambient tiers.** `resolveSourceWithTier` now resolves the flag, so a typo'd id is a hard error rather than a successful zero-page export - that was a fair catch. It also surfaced something neither of us flagged: `--source __all__` was being passed through as a literal source id and matching nothing. Fixed.

But the resolver runs **only when the flag was actually passed**, so `GBRAIN_SOURCE`, the dotfile, and path-match still don't narrow a bare `gbrain export`. That's matching two commands that made the same call for the same stated reason:

- `doctor.ts:4721` - *"EXPLICIT-ONLY by design - a raw flag parse, NOT resolveSourceWithTier. The tier resolver would pick a default source when `--source` is absent and silently scope a bare `gbrain doctor` to one source."*
- `orphans.ts:226` - same reasoning, same wording.

Export has the same hazard with a worse blast radius: its output is an archive, so an ambient dotfile silently drops pages from a backup and you find out at restore time. If you'd rather it match `sync`/`capture`, it's a one-line change and I'll make it - just flagging that the default-safety argument reads stronger here than anywhere else in the CLI.

Also documented `--source` in the export help block, which the first pass missed.

One disclosure on the tests: *"an absent `--source` does not consult GBRAIN_SOURCE"* does **not** discriminate - it passed before this change too, since the raw parse also ignored the env var. It's a regression guard against someone later hoisting the resolver out of the `if` branch, not evidence of a fix. The other three guard tests do discriminate; stubbing the resolver call back to `args.indexOf` fails exactly those three.

time-attack:
Thanks — this is genuinely careful work: resolveSourceWithTier for loud-fail, the __all__ sentinel, the doctor/orphans explicit-only precedent argued honestly, and discriminating tests. The quality isn't the issue. We're holding a hard line on new CLI/product surface right now, so a new export flag is a default no regardless of how well it's built — nothing to do with the craft here. Two concrete paths forward: (1) the sidecar-read scoping this PR stacks on is a real fix and already lives in #3755 — that's the fix-lane piece, and it stands on its own; (2) if you're hitting the export-scoping gap operationally, the flag itself can re-enter as a follow-up rebased on post-#3755 master with that evidence attached — an additive default-unchanged flag with real field evidence is the one feature shape that has landed here. The explicit-only vs full-chain resolution question you flagged is the maintainer call we'd settle at that point (your backup-command silent-narrowing argument is a good one). Closing this one as-is; please do keep #3755 moving.

PR #3678, opened 2026-07-31
v0.42.68.0 fix(contextual): route synopsis through configured models

### LINKED ISSUE #3597: contextual retrieval per_chunk_synopsis unreachable without ANTHROPIC_API_KEY — hardcoded anthropic: model at the import-file.ts call site bypasses claude-cli (#3310)
Verified live on master @ `1057bf4` (VERSION `0.42.67.0`).

Since #3310 landed the `claude-cli` recipe, gbrain has a first-class path for users on a Claude subscription with **no `ANTHROPIC_API_KEY`**. That path works for chat — but contextual retrieval's `per_chunk_synopsis` tier structurally cannot reach it, so those users are still pinned to the `title` tier.

Filing as an issue rather than reopening #1358 — that PR is from another contributor and I can't reopen it. The close comment there said *"If you believe the underlying issue is still live on the latest master, reopen with a quick note and we'll take another look."* It is still live; evidence below.

## Why the default alone isn't the bug

`contextual-retrieval-service.ts` already accepts an override, so a config-driven default looks like it would be enough:

```ts
// src/core/contextual-retrieval-service.ts:73
const DEFAULT_HAIKU_MODEL = 'anthropic:claude-haiku-4-5-20251001';
// :302
haikuModel: args.haikuModel ?? DEFAULT_HAIKU_MODEL,
// :315
const haikuModel = args.haikuModel ?? DEFAULT_HAIKU_MODEL;
```

But the **call site overrides it with a literal**, so no default — config-driven or otherwise — can ever take effect on the import path:

```ts
// src/core/import-file.ts:754
haikuModel: 'anthropic:claude-haiku-4-5-20251001',
```

That line is the actual blocker. Two more hardcodes in the same path:

```ts
// src/core/page-summary.ts:45
const DEFAULT_SYNOPSIS_MODEL = 'anthropic:claude-haiku-4-5-20251001';
// :150
model: args.model ?? DEFAULT_SYNOPSIS_MODEL,

// src/core/minions/handlers/contextual-reindex-per-chunk.ts:52
const RATE_LEASE_KEY = 'anthropic:utility:contextual-synopsis';
```

`contextual-reindex-per-chunk.ts` passes no `haikuModel` at all, and its rate-lease key is provider-qualified, so even a fixed call site would keep leasing under an `anthropic:` bucket.

There is no `GBRAIN_CONTEXTUAL_SYNOPSIS_MODEL` or equivalent config key anywhere on master.

## Effect

On a brain with `embedding_model = zeroentropyai:zembed-1` and no `ANTHROPIC_API_KEY`:

- `models.default = claude-cli:claude-sonnet-4-6` → chat works. `gbrain think` returns normally against the CLI's OAuth session.
- `search.contextual_retrieval = per_chunk_synopsis` → cannot work, because the import path calls `anthropic:` directly and the gateway throws `Anthropic chat requires ANTHROPIC_API_KEY` (`gateway.ts:3036`).

So the recipe added in #3310 for exactly these users is unreachable from the one retrieval tier that would benefit most. Affected operators silently run the degraded `title` tier.

## Suggested fix

1. Thread the model through the call site — `import-file.ts:754` should pass a resolved/configured model (or pass nothing and let the service default apply) rather than a literal.
2. Resolve the default through the existing `models.*` tier system (`models.tier.utility` is the natural home) instead of a module-level constant, in both `contextual-retrieval-service.ts` and `page-summary.ts`.
3. Derive the rate-lease key from the resolved provider so leasing doesn't stay pinned to `anthropic:`.

#1358 proposed an env var for (2) only; on current master that alone would not fix it, because of (1).

## Related — same bug class

- #1307 — brainstorm/lsd hardcoded to `anthropic:claude-sonnet-4-6`, unrunnable without `ANTHROPIC_API_KEY` despite `chat_model` config (open)
- #3593 — `propose_takes` / `grade_takes` bypass the models tier system, `models.default` has no effect (open)
- #1358 — earlier attempt at the synopsis-model override (closed unmerged)

Happy to send a PR if the suggested shape looks right.

cvillarroel2:
## Linked Issue

Closes #3597

## Description

Contextual-retrieval synopsis generation previously carried a hardcoded Anthropic Haiku model through the worker, service, and corpus-generation paths. Installations using another configured chat provider could therefore fall back to title-only embeddings even though their selected model was available through the gateway.

This change:

- resolves a dedicated `models.contextual_synopsis` route through the standard model hierarchy;
- preserves `contextual_retrieval.haiku_model`, `haikuModel`, and `modeRequiresHaiku` as deprecated compatibility surfaces;
- registers explicitly configured native models only for the chat touchpoint before synopsis generation;
- threads the resolved model through each page job once;
- reports the route and its source in `gbrain models`;
- keeps title-only corpus generations independent of the unused synopsis model while retaining model-sensitive invalidation for `per_chunk_synopsis`;
- isolates cross-worker synopsis leases by resolved model and adds `GBRAIN_CONTEXTUAL_SYNOPSIS_RPM`, with the Haiku-named setting retained as a compatibility alias;
- updates release metadata and the architecture reference for v0.42.68.0.

No schema migration is required.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Refactor (no functional changes)
- [ ] Documentation update

## Breaking Changes

N/A. Legacy configuration and service arguments remain supported with deprecation warnings.

## Test Coverage

- [x] I added/updated unit tests
- [ ] I added/updated integration tests
- [x] I tested manually (describe below)
- [ ] No tests needed (explain why)

Manual verification:

- focused contextual-model, config, gateway, embedding, and doctor tests passed: 95 tests, 0 failures;
- `bun run typecheck` passed;
- `bun run build` passed;
- `bun run verify` passed all 33 checks;
- `git diff --check` passed;
- independent review findings were addressed: title-generation identity was normalized, model-specific lease isolation was added, and the separate Codex proxy was removed from this PR;
- the broad 4-shard unit run exposed two unrelated `init-migrate-only` failures under concurrent load; that unchanged test file passed standalone (6 tests, 0 failures).

## Checklist

- [x] I have performed a self-review of my code
- [x] I have added tests that prove my fix/feature works
- [x] New and existing tests pass locally
- [x] I have updated documentation if needed


time-attack:
**Merging this closes #3597.**

Independent triage of #3597 reproduced the reported bug against current master and then confirmed this PR is what fixes it. Routing synopsis through configured models removes the hardcoded provider that issue reports as blocking the feature without that vendor's key.

Worth adding `Closes #3597` to the PR description — a closing keyword only auto-closes from the body or a commit message, not from a comment like this one, so right now the link is visible but not wired up.

time-attack:

Root-cause fix for a real bug (#3597: per_chunk_synopsis hardcoded to Anthropic Haiku, unreachable for claude-cli installs) with genuinely discriminating tests — stubbing each of the three seams (service model threading, handler resolution, chat-touchpoint registration) back to old behavior failed 1/4/3 tests respectively. VERSION trio agrees (0.42.68.0 > master 0.42.67.0), title version-first, typecheck + llms freshness + 85/85 touched and 19/19 adjacent tests pass, CHANGELOG commands execute correctly on a real PGLite brain, and default installs keep the byte-identical Haiku model string so there is no silent cost escalation or corpus_generation churn. BUT the per-touchpoint narrowing of the gateway _extendedModels registry ships a reproduced regression: `gbrain models doctor` probes the configured expansion model via gateway.chat() (src/commands/models.ts:542), which now consults only the chat-touchpoint extended set (src/core/ai/gateway.ts:3032). A valid expansion config like google:gemini-2.0-flash-lite (listed in the google recipe's expansion touchpoint but not its chat allowlist) passes the gate on master (ok:true, live probe succeeds) but throws unknown_model on the PR before any HTTP, and doctor process.exit(1)s (models.ts:693) — cron/CI-wired doctors hard-fail on a working config after upgrade. Fix: probe expansion reachability with the expansion extended set (or via the expansion path), or dual-register expansion models for chat. Minor: the 'deprecated' contextual_retrieval.haiku_model key never existed on master (git grep at c6dc0adf is empty) — the compat narrative is phantom and the key isn't in KNOWN_CONFIG_KEYS (setting it requires --force); either drop it or document it honestly.

### Discrimination test

Three stub-backs of old behavior, each re-running the PR's own tests: (A) service hardcodes 'anthropic:claude-haiku-4-5-20251001' instead of synopsisModel → 'threads a provider-neutral synopsis model to gateway chat byte-for-byte' FAILS; (B) handler stops passing synopsisModel to the service → 4 tests FAIL (resolve-once seam + all 3 registers-native-route tests); (C) registerConfigSelectedChatModel neutered to no-op → 3 registers-native-route tests FAIL (chat call rejected at touchpoint gate). All stubs reverted; tree clean against PR head. Tests are discriminating. One weak spot: the 'inline import contextual synopsis containment' test is readFileSync+substring on src/core/import-file.ts (non-load-bearing).

**Mergeability at review time:** MERGEABLE / UNSTABLE (gh pr checks: no checks reported on the branch)

### Blockers

**1.** src/core/ai/gateway.ts:3032 + src/commands/models.ts:542,693 — per-touchpoint _extendedModels narrowing breaks `gbrain models doctor` for valid native-recipe expansion configs: with `models.expansion`/expansion_model = google:gemini-2.0-flash-lite (natively listed for expansion, not for chat), the doctor's chat() probe now throws AIConfigError unknown_model at config level (master: passes gate and probes live), and doctor exits 1 on a working config. Reproduced empirically on both branches with the same script.

### Not verified

- Full unit suite / bun run verify / ci:local not run (execution-footprint rules); PR body's 'verify passed all 33 checks' and '95 focused tests' claims not reproduced beyond the 104 tests I ran
- No live provider probes (no API spend); the doctor regression was proven at the config-gate level (validateModelId mirrors resolveChatProvider's assertTouchpoint), not with a live gemini-2.0-flash-lite HTTP call
- e2e engine-parity not run (diff contains no engine/SQL changes)
- Rolling-upgrade lease-key transition (old anthropic:utility:contextual-synopsis leases vs new contextual-synopsis:<model> keys can transiently double concurrency for ~60s TTL) reasoned about, not executed

- `gh pr view 3678 --json mergeable,mergeStateStatus,... (MERGEABLE/UNSTABLE)`
- `gh pr checks 3678 (no checks reported on branch)`
- `git fetch origin master pull/3678/head:pr-3678; git checkout pr-3678; git merge-base --is-ancestor origin/master HEAD (up to date with master c6dc0adf)`
- `bun install`
- `VERSION trio audit (0.42.68.0 on all three; master VERSION 0.42.67.0)`
- `bun run typecheck (clean)`
- `bun test test/build-llms.test.ts (12 pass)`
- `bun test test/contextual-synopsis-model.serial.test.ts test/contextual-retrieval-service-pure.test.ts test/embedding-context.test.ts test/config-set.test.ts (85 pass, 0 fail)`
- `bun test test/gateway-tier-extended-models.test.ts test/models-doctor-embed.test.ts test/models-doctor-reranker.test.ts test/ai/gateway-probe-chat-model.test.ts (19 pass)`
- `3x discrimination stub-backs + targeted test reruns (1/4/3 failures as expected), reverted`
- `PR-vs-master probe script: configureGateway(expansion_model=google:gemini-2.0-flash-lite) + validateModelId(model,'chat') → master ok:true, PR unknown_model`
- `bun src/cli.ts init --pglite --no-embedding --skip-embed-check (temp GBRAIN_HOME)`
- `bun src/cli.ts config set models.contextual_synopsis openai:gpt-4o-mini (ok)`
- `bun src/cli.ts models (route reported, source 'config: models.contextual_synopsis')`
- `bun src/cli.ts config set contextual_retrieval.haiku_model ... (rejected without --force; key not in KNOWN_CONFIG_KEYS)`
- `git grep contextual_retrieval.haiku_model c6dc0adf -- src/ (empty: key never existed on master)`
- `gh issue view 3597 (scope alignment confirmed)`

</details>

---

time-attack:

PR #3691, opened 2026-07-31
fix(cycle): stop extract_atoms silently skipping all work on unpriced models

awilhite:
`extract_atoms` extracts nothing and reports success whenever the configured model is not in the pricing maps. Every Groq, local llama-server, OpenRouter and OpenCode Go user hits this; no Anthropic user does.

## Cause

`BudgetTracker.reserve()` hard-fails with `BudgetExhausted(reason:'no_pricing')` when a model is absent from the pricing maps **and** a cap is set. With no cap it warns once and proceeds.

`extract_atoms` always set a cap (`DEFAULT_BUDGET_USD = 0.3`), so on an unpriced model the first `reserve()` throws, the catch latches `budgetExhausted`, and every remaining item is skipped. Skips are not failures, so `failures[]` stays empty and the phase returns `status: 'ok'`.

## Observed

v0.42.67.0, Postgres 17, schema v125. `doctor` reported `1506 page(s) pending` indefinitely.

Instrumenting discovery showed it working correctly — 50 rows returned with the full 18-type allowlist resolved from the active pack. Then:

```
work=50 budgetCap=5 model=litellm:gemma4-12b
item kind=page spent=0 exhausted=false
item kind=page spent=0 exhausted=true
item kind=page spent=0 exhausted=true   (x48)
```

`exhausted` latches after the first item while `spent` is still 0 against a $5 cap. Dropping the cap for unpriced models produced atoms on the first run.

## Changes

- Export `isModelPriceable()` from `budget-tracker.ts`; skip the **default** cap when it returns false. A cap cannot be enforced on a model that cannot be priced, so the real choice is between silently skipping the workload and running with a logged warning. This mirrors the existing `FREE_LOCAL_RERANK_PROVIDERS` escape hatch that keeps the rerank path from tripping the same hard-fail.
- Return `warn` instead of `ok` when every work item was skipped and nothing was extracted.

A user-specified cap on a priceable model is unchanged.

## Follow-up worth considering

This PR does not distinguish a **user-specified** cap from a **defaulted** one. Arguably the more precise fix is to keep the `no_pricing` hard-fail when the operator explicitly asked for a cap (the safety property it was written for) and skip it only for defaults. That needs a `BudgetTracker` API change to carry that intent, so I kept this PR narrow. Happy to extend it if you'd prefer that shape.

## Tests

- 98 pass / 0 fail across the 11 existing atom + budget test files
- New `test/extract-atoms-unpriced-model.test.ts` — 3 tests covering the predicate

## Related

- #2117 — `gbrain-base-v2` ships no `phases:` declaration, gating `extract_atoms` off before it can reach this. Both must be fixed for atoms to flow on a derived pack.
- #1531 — same visible symptom (0 atoms, clean exit) but a different cause; that one is a Postgres discovery failure. Discovery is confirmed healthy here.

### DIFF
--- a/src/core/budget/budget-tracker.ts
+++ b/src/core/budget/budget-tracker.ts
@@ -223,6 +223,17 @@ function lookupPricing(modelId: string, kind: BudgetKind): ModelPricing | null {
   return null;
 }
 
+/**
+ * True when the budget tracker can price this model, i.e. when setting a cost
+ * cap is meaningful. Callers that apply a *default* cap (rather than one the
+ * user asked for) should skip the cap when this returns false — otherwise
+ * `reserve()` hard-fails with BudgetExhausted(reason:'no_pricing') and the
+ * caller silently does no work.
+ */
+export function isModelPriceable(modelId: string, kind: BudgetKind): boolean {
+  return lookupPricing(modelId, kind) !== null;
+}
+
 function costForUsage(modelId: string, inputTokens: number, outputTokens: number, kind: BudgetKind): number | null {
   const p = lookupPricing(modelId, kind);
   if (!p) return null;
--- a/src/core/cycle/extract-atoms.ts
+++ b/src/core/cycle/extract-atoms.ts
@@ -52,7 +52,7 @@ import type { PhaseResult } from '../cycle.ts';
 import type { GBrainConfig } from '../config.ts';
 import type { ProgressReporter } from '../progress.ts';
 import { chat as gatewayChat, withBudgetTracker } from '../ai/gateway.ts';
-import { BudgetExhausted, BudgetTracker } from '../budget/budget-tracker.ts';
+import { BudgetExhausted, BudgetTracker, isModelPriceable } from '../budget/budget-tracker.ts';
 import { writeReceipt } from '../extract/receipt-writer.ts';
 import { upsertExtractRollup } from '../extract/rollup-writer.ts';
 import { createHash } from 'crypto';
@@ -549,8 +549,21 @@ export async function runPhaseExtractAtoms(
   } catch {
     // Keep safe defaults: Haiku + $0.30.
   }
+  // A cost cap is only meaningful for a model the tracker can price.
+  // BudgetTracker.reserve() hard-fails with BudgetExhausted(reason:'no_pricing')
+  // when the model is absent from the pricing maps AND a cap is set; with no cap
+  // it warns once and proceeds. Because this phase always set a cap, every
+  // non-Anthropic model tripped that hard-fail on the first item, latched
+  // `budgetExhausted`, and skipped the entire workload while reporting ok.
+  const priceable = isModelPriceable(extractModel, 'chat');
+  if (!priceable) {
+    console.error(
+      `[extract_atoms] model "${extractModel}" is not in the pricing maps; ` +
+        `running without a cost gate (a cap cannot be enforced on an unpriced model).`,
+    );
+  }
   const budgetTracker = new BudgetTracker({
-    maxCostUsd: budgetCap,
+    maxCostUsd: priceable ? budgetCap : undefined,
     label: 'cycle.extract_atoms',
   });
 
@@ -726,7 +739,16 @@ export async function runPhaseExtractAtoms(
 
   return {
     phase: 'extract_atoms',
-    status: failures.length > 0 ? 'warn' : 'ok',
+    // A phase that skipped every work item and produced nothing did not
+    // succeed, even though skips are not failures and leave failures[] empty.
+    // Reporting 'ok' there hides a total no-op behind a green status.
+    status:
+      failures.length > 0 ||
+      (work.length > 0 &&
+        totalAtomsExtracted === 0 &&
+        transcriptsSkipped + pagesSkipped === work.length)
+        ? 'warn'
+        : 'ok',
     duration_ms: 0,
     summary:
       `extract_atoms: ${totalAtomsExtracted} atoms from ` +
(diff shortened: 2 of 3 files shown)


time-attack:

Real bug, correctly diagnosed (verified by execution: cap+unpriced model -> BudgetExhausted(no_pricing) on item 1, latched, all work skipped, status 'ok'), and the fix mechanism works (cap unset -> warn-once, proceeds). But the fix drops the cost cap for ALL unpriced models including a user-configured cycle.extract_atoms.budget_usd, creating an unbounded-spend path through LiteLLM-fronted paid providers — the exact case budget-tracker.ts:148 deliberately keeps fail-closed ('a LiteLLM proxy can front a paid provider'). Verified record() accrues $0 for unpriced models, so no ceiling of any kind remains. The PR's deferral excuse ('needs a BudgetTracker API change') is false: configuredBudget is in scope 10 lines above the change; a 2-line userSetCap guard makes it precise, and the PR's own all-skipped->'warn' change then surfaces the user-capped case honestly. The new tests also fail the discrimination test: with extract-atoms.ts reverted wholesale to master, all 3 pass (they pin only the new pure predicate, not the wiring or status change), despite four existing test files already harnessing runPhaseExtractAtoms with stubs. Plus procedural failures: title not version-first, VERSION trio unchanged at 0.42.67.0 == master (fails the strictly-greater CI gate). The all-skipped->'warn' status change itself is correct (skip counters only increment on budget skips).

### Discrimination test

Reverted src/core/cycle/extract-atoms.ts to origin/master (fix fully unwired, only the exported predicate remaining) and re-ran the PR's new test file: 3 pass / 0 fail, exit 0. The tests prove nothing about the actual fix — they pin only the new pure predicate isModelPriceable, not the cap-skip wiring nor the all-skipped->'warn' status change.

### Blockers

**1.** src/core/cycle/extract-atoms.ts:566 — `maxCostUsd: priceable ? budgetCap : undefined` silently voids a USER-CONFIGURED cap. Failing input: models.dream.extract_atoms=litellm:<paid-fronted-model> + cycle.extract_atoms.budget_usd=5 + large backlog -> entire backlog runs with no cost ceiling (verified: record() accrues $0 for unpriced models; no runtime cap either). Fix: set userSetCap where configuredBudget is parsed (same function, ~line 546) and use `priceable || userSetCap ? budgetCap : undefined`; no BudgetTracker API change needed, contrary to the PR body.

**2.** test/extract-atoms-unpriced-model.test.ts — fails the discrimination test: reverting src/core/cycle/extract-atoms.ts wholesale to master leaves all 3 new tests green (they test only the new isModelPriceable predicate). Need a test that fails on master, e.g. runPhaseExtractAtoms with a stub engine/gateway and an unpriced model asserting work is processed (and/or all-skipped -> status 'warn'); test/cycle/extract-atoms-progress.test.ts shows the harness pattern.

**3.** Procedural: PR title must lead with vMAJOR.MINOR.PATCH.MICRO; VERSION/package.json/CHANGELOG untouched at 0.42.67.0 == master's 0.42.67.0 — equal fails the CI version-gate (strictly-greater required).

### Not verified

- Did not run the full test suite or bun run verify (footprint rules); the PR body's '98 pass across 11 existing atom + budget test files' claim was not reproduced — I ran only the PR's new test file plus typecheck
- Did not run a live end-to-end cycle against a real Postgres brain with an actual unpriced provider (reporter's litellm:gemma4-12b repro); mechanism verified at the BudgetTracker unit level plus code-path reading instead
- mergeStateStatus UNSTABLE: no CI checks reported on the fork branch, so the green-CI question is moot but CI behavior on this head was not observed

- `gh pr view 3691 --json number,title,body,mergeable,mergeStateStatus,... (mergeable: MERGEABLE, mergeStateStatus: UNSTABLE)`
- `gh pr diff 3691 (read every hunk)`
- `gh pr checks 3691 (no checks reported on fork branch)`
- `git fetch origin pull/3691/head:pr-3691 && git checkout pr-3691`
- `VERSION trio audit: VERSION=0.42.67.0, package.json=0.42.67.0, CHANGELOG top=0.42.67.0, origin/master VERSION=0.42.67.0 (equal, not strictly greater)`
- `bun run typecheck > /tmp/pr3691_typecheck.txt (EXIT=0)`
- `bun test test/extract-atoms-unpriced-model.test.ts > /tmp/pr3691_newtest.txt (EXIT=0, 3 pass)`
- `git checkout origin/master -- src/core/cycle/extract-atoms.ts; bun test test/extract-atoms-unpriced-model.test.ts > /tmp/pr3691_discrim.txt (EXIT=0, 3 pass — discrimination test FAILED)`
- `bun run tmp-verify-3691.ts — behavioral proof: capped+unpriced reserve() throws BudgetExhausted(reason:'no_pricing'); uncapped+unpriced proceeds with warn-once; record() of 5M+5M tokens on unpriced model accrues $0 (totalSpent=0, cap=undefined)`
- `grep -rn 'new BudgetTracker' src/ (11 sites; sibling default-cap sites: enrich.ts:515 ($5 default), extract-conversation-facts.ts:1422 ($5 default), conversation-facts-backfill.ts:203 ($5 default))`
- `grep -rln runPhaseExtractAtoms test/ (4 existing harness files — behavioral test was feasible)`
- `Read src/core/budget/budget-tracker.ts (TX2 contract, lookupPricing miss policy incl. deliberate litellm exclusion at ~line 148), src/core/cycle/extract-atoms.ts lines 500-790 (cap resolution, work loop skip semantics, PhaseResult details)`

</details>

---

time-attack:
Merged into master via the v0.42.69.0 community fix wave (#3709) with your commit attributed via Co-Authored-By and CHANGELOG credit. One reconciliation: #3541 in the same wave prices ollama/llama-server at $0, so your isModelPriceable test's llama-server example inverted — the wave swaps in groq (the paid-but-unpriced case your own description cites) and adds the positive assertion that free local providers keep their cap enforced. The two fixes compose well: local models keep enforceable caps, genuinely unpriced providers now skip loudly. Master CI is green. Thank you!

PR #3693, opened 2026-07-31
fix(chunker): make Dart code intelligence work at all

JohannOosthuizen:
## Summary

`code-def`, `code-refs`, `code-callers` and `code-callees` return `count: 0`
for every Dart symbol in an indexed Flutter repo. Three independent causes,
each fatal on its own.

Dart is registered in `LANGUAGE_MANIFEST` and `.dart` is correctly detected, so
everything upstream reports success and nothing warns.

### 1. The bundled grammar cannot load

`tree-sitter-dart.wasm` is built at ABI 15; the pinned `web-tree-sitter` 0.22.6
accepts 13–14. `setLanguage` throws, so no `.dart` file is ever parsed and each
one falls back to a single whole-file chunk with a null `symbol_name` — the
column `code-def` queries.

Loading each bundled grammar through gbrain's own web-tree-sitter:

```
FAIL  dart: Incompatible language version 15. Compatibility range 13 through 14.
OK    kotlin
OK    python
OK    typescript
OK    go
OK    rust
```

Rebuilt from `UserNobody14/tree-sitter-dart@be07cf7`, matching the recipe
already documented for the SQL grammar at `code.ts:80`:

```bash
npx tree-sitter-cli@0.26.3 generate --abi 14
npx tree-sitter-cli@0.26.3 build --wasm
```

The grammar's checked-in `src/parser.c` is `#define LANGUAGE_VERSION 15`; after
`generate --abi 14` it is 14, and the resulting wasm loads and parses under
web-tree-sitter 0.22.6.

### 2. `SYMBOL_NODE_TYPES` has no `dart` key

So no Dart node qualified as a symbol even once the grammar loaded. Node names
are taken from the grammar's own `src/node-types.json` rather than inferred.
`static_final_declaration_list` is what covers top-level `final x = ...`, which
is how Riverpod providers and most Dart module-level constants are declared.

### 3. Dart emits a function as two sibling nodes

The signature, then a separate `function_body` — where every other bundled
grammar nests the body inside the declaration node. Unpaired, a top-level
function chunk is one line, so `mergeSmallSiblings()` folds it into an unnamed
`merged` chunk and the symbol name is lost. This one only becomes visible after
fixing 1 and 2: classes and enums resolve while top-level functions stay
missing.

### Also: `extractSymbolName` needed a Dart arm

Dart signatures carry the return type as a `type_identifier` child *before* the
`identifier` that names the thing, so the generic `endsWith('identifier')` scan
returns `Future` for `Future<T?> _approve(...)`. And `method_signature` /
`static_final_declaration_list` hold no identifier at all — the name is one
level down. The new arm runs *after* the existing `name`-field check, so
grammars that share a node name are unaffected (TS has both `function_signature`
and `method_signature`, each with a `name` field).

## Testing

Measured on a 328-file Flutter app: named chunks **0 → 1972 of 1991 (99.0%)**,
0 parse errors — 1135 classes, 388 functions, 296 top-level finals, 41 enums.
Symbol lookup that previously returned `count: 0` now resolves to the correct
type and line range, verified against a live index.

Since (3) and the `extractSymbolName` arm touch shared code, I ran the chunker
over 626 non-Dart files (TypeScript, Kotlin, Java, Python) before and after:
**byte-identical chunk output**, so the shared-code changes are inert for other
languages.

`test/code-intel`, `code-def-refs`, `code-edges`, `chunker-*` and
`reindex-code`: 109 pass, 0 fail.

## One thing worth considering beyond Dart

The failure mode is indistinguishable from a true negative:

```json
{"symbol": "Foo", "count": 0, "status": "ready", "ready": true, "results": []}
```

`status: "ready"` asserts the index is built and the answer trustworthy. An
agent asking "who calls this before I change it", getting `count: 0`, will
conclude the symbol has no callers and refactor on that basis.

Not part of this PR, but worth a follow-up: surfacing `Language.load` failures
instead of degrading to whole-file chunks, and having the code-intel commands
report an unsupported or unparsed language rather than `count: 0,
status: "ready"`. That would have made this self-diagnosing, and it would catch
the next grammar that drifts past the ABI pin.


time-attack:

The fix is real and verified end-to-end: master's ABI-15 Dart wasm fails to load under pinned web-tree-sitter 0.22.6 (reproduced: 'Incompatible language version 15'), degrading every .dart file to one unnamed whole-file chunk (reproduced: 0 named symbols); the PR's rebuilt ABI-14 wasm loads, and the three code fixes (TOP_LEVEL_TYPES dart set, signature+function_body sibling pairing, Dart extractSymbolName arm) produce correctly-named chunks for classes/mixins/extensions/enums/functions/top-level finals. Malformed-Dart robustness verified across 14 adversarial inputs with no crash or hang (the one slow case, a 200K-char single line, costs the same for TypeScript — pre-existing, language-independent). Non-Dart chunk+edge output is byte-identical between master and PR across TS/JS/Kotlin/Python/Go/Rust fixtures. But the PR ships ZERO tests and fails the discrimination test outright: with master's broken chunker and wasm stubbed back in, all 161 tests in the 12 suites the PR body cites still pass. The repo's own check-wasm-embedded.sh calls grammar-load regression 'the #1 silent-failure mode' yet guards only TypeScript — exactly how the ABI-15 wasm shipped unnoticed, and exactly how the next web-tree-sitter bump regresses Dart back to count:0 silently. Needs: (1) a Dart chunking fixture test asserting named symbols that fails on master, ideally plus an all-bundled-grammars Language.load ABI test; (2) version-first PR title per IRON RULE; (3) provenance comment on the G_DART import (grammar commit + --abi 14 recipe) matching the SQL precedent at code.ts:80.

### Discrimination test

Checked out master's src/core/chunkers/code.ts + old tree-sitter-dart.wasm on the PR branch and re-ran the exact test files the PR body cites (test/chunkers/, test/chunker-timeout.test.ts, test/code-def-refs.test.ts, test/code-edges.test.ts, test/code-intel/, test/language-manifest.test.ts): 161 pass / 0 fail — identical to the PR-branch run. The suite cannot detect the bug this PR fixes; the PR adds no tests. FAILED.

**Mergeability at review time:** mergeable=MERGEABLE, mergeStateStatus=UNSTABLE (no CI checks reported on the fork branch — green CI is entirely absent, not stale)

### Blockers

**1.** No test fails without the fix: stubbing master's code.ts + old wasm back in leaves all 161 tests across the PR-cited suites green (test/chunkers/, chunker-timeout, code-def-refs, code-edges, code-intel/, language-manifest) — the change is unpinned and the ABI-drift failure class that caused this bug remains unguarded for all grammars except TypeScript (scripts/check-wasm-embedded.sh)

**2.** PR title 'fix(chunker): make Dart code intelligence work at all' violates the version-first IRON RULE — must lead with vMAJOR.MINOR.PATCH.MICRO at merge

**3.** src/core/chunkers/code.ts:46 — G_DART import lacks the provenance comment (source repo+commit, tree-sitter-cli@0.26.3 --abi 14 recipe) that the SQL grammar precedent at code.ts:80-84 establishes for checked-in wasm blobs (minor)

### Not verified

- Byte-for-byte reproducibility of the rebuilt tree-sitter-dart.wasm from UserNobody14/tree-sitter-dart@be07cf7 with tree-sitter-cli@0.26.3 --abi 14 (needs network + emscripten toolchain); I verified the artifact loads at ABI 14 with 527 node types and behaves correctly, but did not rebuild it
- PR body's 328-file Flutter app numbers (1972/1991 named, 99.0%) — directionally consistent with my fixtures but not reproduced at that scale
- Full test suite / bun run verify / scripts/check-wasm-embedded.sh (compiles a binary) — skipped per execution-footprint rules
- Live code-def/code-refs/code-callers CLI round-trip against an indexed repo — verified at the chunker layer only

- `gh pr view 3693 --json title,body,mergeable,mergeStateStatus,files (MERGEABLE / UNSTABLE — no CI checks reported on fork branch)`
- `git fetch origin pull/3693/head:pr-3693 && git checkout pr-3693; bun install`
- `bun .verify-3693/abi-check.ts — old wasm: 'Incompatible language version 15. Compatibility range 13 through 14.'; new wasm: loads, parses, confirms function_signature+function_body sibling shape; lang.version=14, 527 node types`
- `bun .verify-3693/chunk-dart2.ts / chunk-dart3.ts on PR branch — named chunks for class/enum/mixin/extension/function/static-final (e.g. counterProvider, _approve, main, ApprovalService, Loggable, StringX)`
- `same fixture with master code.ts+wasm — 1 unnamed 'module' chunk, 0 named (bug reproduced)`
- `gtimeout-capped malformed-Dart suite, 14 cases (garbage bytes, truncated class, unbalanced braces, 5000-deep nesting, 3000 tiny functions, signature-at-EOF, nested signature bomb, unicode soup) — all OK, no crash/hang`
- `long-line probe: dart vs typescript at 100K chars (10.9s vs 10.1s) and 200K chars (39.1s vs 38.7s) — slowness is language-independent and pre-existing`
- `regression diff: chunkCodeTextFull over TS/JS/Kotlin/Python/Go/Rust fixtures under master vs PR chunker — byte-identical (diff exit 0)`
- `bun run typecheck — clean (exit 0)`
- `bun test test/chunkers/ test/chunker-timeout.test.ts test/code-def-refs.test.ts test/code-edges.test.ts test/code-intel/ test/language-manifest.test.ts — 161 pass / 0 fail on PR branch AND with old behavior stubbed back (discrimination test failed)`
- `git show c44cdb52 --stat — recently merged community fix #3341 shipped no VERSION bump (maintainer bumps at release; convention) but did ship a 155-line test file`

</details>

---

time-attack:
Review finding (two blockers, otherwise merge-lane): verified the bug is real — master advertises Dart in `LANGUAGE_MANIFEST` and `.dart` detection yet produces zero symbols (module:null chunks), while your branch's grammar + code produce proper class/enum/function/mixin symbols, and Python output is byte-identical before/after. Good work.

Two things block it: (1) **zero tests** — no Dart fixture in `test/chunkers/code.test.ts`, so nothing fails without the fix and nothing guards the next ABI drift; please add a fixture asserting the symbol names/types you demonstrated. (2) The **1.1MB contributor-built `tree-sitter-dart.wasm`** — a maintainer can't trust an opaque binary from a PR blob, especially since `Language.load` errors are swallowed and a bad grammar silently degrades every .dart file back to whole-file chunks. Please include the exact rebuild recipe (source repo + commit + `--abi 14` command) in the PR body and ideally a checked-in `scripts/` note so the maintainer can rebuild and diff the hash locally. Cosmetic follow-up, non-blocking: `normalizeSymbolType` has no Dart arms, so types surface as 'static final declaration list' — same as Kotlin's existing quirk.

time-attack:
Status update: reviewed in the 2026-08-18 full-backlog triage — still wanted, but not yet mergeable: Real bug (Dart ABI-15 wasm can't load under web-tree-sitter 0.22.6) but review blockers unanswered: zero tests and an unverifiable contributor-built wasm binary. If you update it, it will be re-reviewed promptly; alternatively the fix may be adopted with credit in an upcoming fix wave. Thanks for the contribution and the patience.

time-attack:
Superseded: v0.46.23.0 adopted a competing Dart grammar fix (#3977 by @Jey2311) with the pinned build script, and this branch now conflicts on both files. One residual you found remains real (top-level final bindings unchunked) — please file it as a focused issue and we will fast-track it. Thank you!

PR #3704, opened 2026-07-31
fix(sync): reject NUL bytes before code import

exGeni:
## Summary

- reject NUL-bearing code content before guardrails, chunking, or database writes
- keep sync fail-closed using the existing `NULL_BYTES` failure-ledger classification
- cover a NUL beyond the capture command's 8 KB sniff window

## Reproduction

`gbrain sync --strategy auto --no-embed` attempted to index tracked `.ts` and `.kt` binary fixtures. PostgreSQL rejected each write with `invalid byte sequence for encoding "UTF8": 0x00`, leaving the source checkpoint blocked and the error classified as `UNKNOWN`.

## Verification

- regression before fix: expected `skipped`, received `imported`
- `bun test test/import-file.test.ts`: 29 pass, 0 fail
- `git diff --check`

The broader sync-failure suite was also run; its writable-ledger cases fail under the read-only home sandbox (`EROFS`), while all pure classification checks pass.

### DIFF
--- a/src/core/import-file.ts
+++ b/src/core/import-file.ts
@@ -1157,6 +1157,9 @@ export async function importCodeFile(
   opts: { noEmbed?: boolean; force?: boolean; sourceId?: string } = {},
 ): Promise<ImportResult> {
   const slug = slugifyCodePath(relativePath);
+  if (content.indexOf('\0') !== -1) {
+    return { slug, status: 'skipped', chunks: 0, error: 'Content contains null bytes (likely binary corruption)' };
+  }
   const lang = detectCodeLanguage(relativePath) || 'unknown';
   const title = `${relativePath} (${lang})`;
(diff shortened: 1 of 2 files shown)


time-attack:
Closing — master already fixes this, and the better way. `fix(index): preserve code files containing NUL bytes (#3483)` escapes `\0` to `\\0` in `src/core/import-file.ts` and keeps the page indexed. This PR's early `return { status: 'skipped' }` goes back to dropping the whole file, which directly contradicts the merged test `test/import-file.test.ts` ("stores code containing NUL bytes without dropping the page") — the two tests assert opposite behavior, so merging would turn master red. Legitimate source files (regex and string fixtures) do contain U+0000, so skipping them is data loss rather than a fix. Thanks for the report either way; the underlying concern was real.

Issue #3674, opened 2026-07-31
extract links --by-mention: no cleanup path for stale/incorrect link_source='mentions' rows


cfeddersen:
Split out of #3563 (review follow-up), where it was raised as a blocker but is a write-path design change rather than a tokenizer fix.

## Problem

`engine.addLinksBatch` is additive, and `extract links --by-mention` has no purge or rebuild mode. When the gazetteer changes such that previously-written links become wrong — the tokenizer fix in #3563 is the motivating case, but renaming or deleting an entity page does it too — the op-checkpoint's `gazetteerHash` correctly self-invalidates and a full re-scan runs, but the **correct links are added alongside the stale ones**. Nothing ever removes a `link_source='mentions'` row that the current gazetteer would no longer produce.

Concretely, in the corpus that motivated #3563 the old ASCII tokenizer wrote ~820 `mentions` rows for one entity where the correct count is ~440. After upgrading and re-scanning, that brain has ~440 correct rows *plus* the ~380 false positives, not ~440 total.

## Why it is not a one-line DELETE

**`link_source='mentions'` is a shared namespace with two producers.** `src/core/extract-ner.ts` deliberately writes `link_source='mentions'` with `link_kind='typed_ner'` rather than minting new provenance (see the comment at the top of that file — the choice was made to avoid breaking existing `mentions` consumers). So:

- deleting by `link_source='mentions'` also wipes every NER-derived typed link;
- deleting only `link_kind != 'typed_ner'` leaves NER rows that were derived from the *same* stale gazetteer, and are wrong for the same reason.

Both producers need a coordinated purge + re-run, which is a decision about how the two passes compose, not a detail of either one.

**There is no bulk primitive.** The engine exposes per-pair `removeLink(from, to, linkType?, linkSource?, opts?)` only. A cleanup needs:

- a new method on the `BrainEngine` interface, implemented in both `postgres-engine.ts` and `pglite-engine.ts`;
- a CLI surface (`extract links --by-mention --rebuild`? a `doctor` repair?) and a decision about whether it is opt-in or automatic on gazetteer-hash change;
- a run of the `DATABASE_URL`-gated engine-parity suite, since it is a destructive write path on both backends.

## Possible shapes

[...]

PR #3711, opened 2026-08-01
feat(doctor): stale_mentions — surface by-mention links the gazetteer no longer produces

cfeddersen:
Read-only first step for #3674, which you triaged p2 and endorsed this shape for:

> Low-risk first step is its option (3): a read-only doctor check that counts `link_source='mentions'` rows whose (from_page, to_page) pair the current gazetteer scan would not produce […] making the problem visible with no write path.

## Problem

`extract links --by-mention` writes through `addLinksBatch`, which is purely additive, and `put_page` reconciliation deliberately excludes `'mentions'` (`operations.ts:1329-1333`). A mentions row therefore outlives its own justification — rewrite a body so it stops naming the entity, delete or retype the entity page, or change the tokenizer, and the row stays. Re-running the scan does not help: it adds today's correct links *alongside* the old ones. Nothing surfaces the drift, so it compounds silently and feeds relational retrieval and graph walks.

## What this adds

`scanStaleMentions(engine, { limit })` in `src/core/by-mention.ts` rebuilds the gazetteer, re-derives mentions for the pages that carry mentions rows, and reports rows whose target the current scan no longer produces. A `stale_mentions` doctor check formats it.

```
stale_mentions  warn  3 of 12 by-mention link(s) no longer follow from the current
                      gazetteer (2 plain, 1 typed_ner). Re-running the scan will NOT
                      remove them — the write path is additive. Remove individually
                      with: gbrain unlink <from> <to> --link-type mentions
                      --link-source mentions. e.g. notes/meeting -> companies/acme
```

## Design choices worth reviewing

- **Strictly read-only, and tested as such.** No repair path, no `RemediationStep`. The real fix is a per-page scoped delete-then-insert needing a new bulk engine method on *both* backends, coordinated with extract-ner's `typed_ner` rows and the resume checkpoint. Shipping that as a side effect of a doctor check would be a destructive write nobody asked to review. #3674 stays open for it.
- **Staleness is target-based, not kind-based**, so it covers `link_kind='typed_ner'`: extract-ner derives those from the same mention set, so a target this scan no longer yields cannot have a live verb-typed edge either. The converse does *not* hold — a still-produced target says nothing about whether the stored verb is still right — so `typed_ner` counts are reported separately rather than folded into one number.
- **Bounded at 500 pages** in slug order, and the message states what was *not* covered (`sampled N of M pages carrying mentions links`) rather than implying a full sweep.
- **Empty gazetteer is called out separately.** It makes every mentions row unreproducible, which is a different diagnosis from ordinary drift; without saying so, an operator reads "everything is stale" and starts deleting real links.
- **Never `fail`, only `warn`** — wrong data with a per-pair `gbrain unlink` workaround is the p2 bar.
- **Early-returns after one COUNT** when the brain has no mentions rows, so brains that never ran by-mention pay almost nothing.
- Registered in `doctor-categories.ts` (BRAIN) so the categorization drift guard stays green.

## Tests

+12 in `test/doctor-stale-mentions.test.ts`: your exact reproduction from the issue (body rewritten, row survives), soft-deleted entity, `typed_ner` split, legacy NULL `link_kind`, read-only + idempotency, `limit` vs unbounded total, cross-source guard, plus doctor-surface coverage that the check is emitted, warns with the additive-write-path wording and the workaround, and is categorized.

## Verified / not verified

- `tsc --noEmit` — exit 0.
- All **47** doctor test files + `by-mention`, `cjk`, `e2e/orphan-reduction` — **416 pass / 4 fail**.
- The 4 failures are load-dependent timeouts in `doctor.test.ts` (`graph_signals_coverage`, `link_resolution_opportunity`, in-progress-sync-lock) and `doctor-orphan-ratio.test.ts` (`auto_chronicle` volume). All pass when those files run alone. **`origin/master` produces the identical 4** — 404 pass / 4 fail, same test names, same durations (5.3/5.3/7.0/7.2s vs 5.8/5.3/7.1/7.4s). Pre-existing, verified by running the baseline rather than assumed.
- `check:privacy`, `check-test-real-names`, `check-test-isolation` — exit 0.
- NOT run: the full `bun test` suite (it does not complete in a reasonable window on this machine), and the `DATABASE_URL`-gated engine-parity suite. Neither is implicated — this adds no engine method and no write path; the two new queries are plain reads through `executeRaw`.

### DIFF
--- a/src/commands/doctor.ts
+++ b/src/commands/doctor.ts
@@ -6511,6 +6511,79 @@ export async function buildChecks(
     checks.push({ name: 'orphan_ratio', status: 'warn', message: 'Could not check orphan ratio' });
   }
 
+  // 9c. stale_mentions — auto-linked mentions the current gazetteer no
+  // longer produces (issue #3674).
+  //
+  // `extract links --by-mention` writes through addLinksBatch, which is
+  // purely additive, and put_page reconciliation deliberately excludes
+  // 'mentions'. So a link survives its own justification: rewrite a body so
+  // it stops naming the entity, delete or retype the entity page, or change
+  // the tokenizer, and the row stays. Re-running the scan does not help —
+  // it adds today's correct links ALONGSIDE the old ones.
+  //
+  // READ-ONLY by design. There is no repair path yet (a scoped
+  // delete-then-insert needs a new bulk engine method on both backends,
+  // coordinated with extract-ner's link_kind='typed_ner' rows and the
+  // resume checkpoint), so this check makes the drift visible and stops
+  // there rather than shipping a destructive write nobody reviewed.
+  // No RemediationStep for the same reason.
+  //
+  // Never 'fail': the data is wrong but a per-pair `gbrain unlink`
+  // workaround exists, which is the p2 bar.
+  progress.heartbeat('stale_mentions');
+  const staleMentionsHb = startHeartbeat(progress, 're-deriving by-mention links…');
+  try {
+    const { scanStaleMentions } = await import('../core/by-mention.ts');
+    const res = await scanStaleMentions(engine);
+    // Say what was NOT covered rather than implying the whole brain was.
+    const coverage = res.pagesScanned < res.totalPagesWithMentions
+      ? ` (sampled ${res.pagesScanned} of ${res.totalPagesWithMentions} pages carrying mentions links)`
+      : '';
+
+    if (res.totalPagesWithMentions === 0) {
+      checks.push({
+        name: 'stale_mentions',
+        status: 'ok',
+        message: 'No by-mention links in this brain.',
+      });
+    } else if (res.staleLinks === 0) {
+      checks.push({
+        name: 'stale_mentions',
+        status: 'ok',
+        message: `Re-derived ${res.linksScanned} by-mention link(s) across ${res.pagesScanned} page(s)${coverage}; all still follow from the current gazetteer.`,
+      });
+    } else {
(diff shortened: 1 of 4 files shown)


time-attack:
Automated triage (2026-08-01): adds new product surface (new doctor check), which is reserved for the maintainer under triage policy. Queued for maintainer product decision; no automated test or merge action taken.

time-attack:
Status update: reviewed in the 2026-08-18 full-backlog triage. It is queued but currently deferred: No CI ever ran at head AND new doctor check awaits maintainer product call. Address that and it re-enters the merge queue. Thanks for the contribution.

time-attack:
Trial-merge is clean; tsc fails on master skew: test/doctor-stale-mentions.test.ts:28 CliOptions literal now requires the 'brain' field (1-line fix). Separately the new doctor check still awaits a maintainer product decision, so this stays open either way — but the type fix makes it landable the moment that call is made.

garrytan:
Thanks @cfeddersen — this read-only-first shape was exactly the right opening move for #3674, and it landed. `src/commands/doctor/checks/stale-mentions.ts` and `scanStaleMentions` in `src/core/by-mention.ts` shipped with the v0.46.28.0 megawave (#4475), with source comments crediting "(#3674, lands PR #3711)" by name. Master then went a step further than this PR proposed: `src/commands/extract.ts` now carries a repair path for the drift the check reports. Closing as absorbed with credit already in the source — thank you for making the problem visible before making it writable.

PR #3715, opened 2026-08-01
fix(extract): timeline bullet separator must be a spaced em/en-dash, not a bare hyphen

andreineacsu:
The Format-1 bullet parser splits source/summary on `\s*[—–-]\s*`, a class that includes a bare hyphen-minus, with a non-greedy source group. On any bullet lacking the canonical `Source — ` prefix (and on canonical ones with an earlier in-word hyphen) it matches the first hyphen anywhere (`scale-incumbent`, `Viorica-Ioana`, `follow-up`), mis-filing the summary prefix as the source. The mangled (source, summary) tuple then diverges from cleanly-written rows, defeating the `(page_id, date, summary, source)` dedup, so re-syncs accumulate duplicates.

Fix: the separator is now a spaced em/en-dash only; a bullet with no such separator has no source label (the whole text is the summary). Regression tests included.

We have run this patch locally since mid-July (it is the carry we re-apply after every upgrade); this PR upstreams it so the carry can retire.

### DIFF
--- a/src/commands/extract.ts
+++ b/src/commands/extract.ts
@@ -476,11 +476,26 @@ export async function extractLinksFromFile(
 export function extractTimelineFromContent(content: string, slug: string): ExtractedTimelineEntry[] {
   const entries: ExtractedTimelineEntry[] = [];
 
-  // Format 1: Bullet — - **YYYY-MM-DD** | Source — Summary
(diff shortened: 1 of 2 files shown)


garrytan:
Thanks — master now ships whitespace-bounded, link-aware dash matching for timeline bullets (735dec83b via #3280), which covers the in-word-hyphen mis-split this PR fixed. Closing as superseded.

PR #3718, opened 2026-08-01
fix(cycle): honour models.dream.synthesize in synthesize_concepts

awilhite:
The narrative `chat()` call in `synthesize_concepts` passes no `model:`, so it falls through to the gateway default (`models.chat`). The routing table advertises `models.dream.synthesize -> tier.reasoning`, but that key is never read on this path — configuring it does nothing.

## Why it is quiet

When the inherited model rejects the calls, each one throws, `narrative` falls back to `deterministicNarrative(group)`, and the phase still reports the same concept count. The only signal is the trailing `(N LLM-failed → template fallback)`. Concepts exist, so nothing looks broken — they are just stubs instead of narratives.

## Observed

v0.42.67.0, Postgres, 1,858-atom corpus, 555 concepts (T1=54 T2=90 T3=411).

With `models.chat` on a free-tier provider whose daily token allowance is smaller than one synthesis pass (144 narrative calls, each carrying 5 atom bodies + titles):

```
run 1: 136 LLM-failed → template fallback
run 2: 131 LLM-failed → template fallback     (different day, cap reset)
```

Same corpus pointed at a provider with headroom:

```
run 3:   0 failures, 344s   (vs 887s — the failing calls were burning retry/backoff)
```

Consistent across runs because the daily cap resets but the workload does not.

## Fix

Resolve the task model with the same helper every other phase uses — cf. `auto-think.ts`:

```ts
const synthModel = await resolveModel(engine, {
  configKey: 'models.dream.synthesize',
  tier: 'reasoning',
  fallback: 'sonnet',
});
```

and pass it on the `chat()` call. Restores the documented routing. No behaviour change for anyone who has not set `models.dream.synthesize` — the tier + fallback path resolves as before.

## Why this matters beyond the config bug

`propose_takes` and `synthesize_concepts` are different work — strict-JSON extraction across thousands of pages versus generative synthesis over a fixed ~144 clusters — and they suit different models. The tier system already encodes that. Sharing `models.chat` between them forces one choice for both; this restores the split the tiers were designed for.

## Tests

- 73 pass / 0 fail across `cycle-synthesize`, `cycle-synthesize-config-zero`, `synth-enabled-default`, `model-config.serial`

## Related

Same silent-degradation shape as #3691 (extract_atoms skipping all work while reporting ok). Both are phases that continue after their real work fails, with the failure visible only in a suffix or not at all.

### DIFF
--- a/src/core/cycle/synthesize-concepts.ts
+++ b/src/core/cycle/synthesize-concepts.ts
@@ -19,6 +19,7 @@
 //   5. Write concept-typed pages.
 
 import type { BrainEngine } from '../engine.ts';
+import { resolveModel } from '../model-config.ts';
 import type { PhaseResult } from '../cycle.ts';
 import type { ProgressReporter } from '../progress.ts';
 import { writeReceipt } from '../extract/receipt-writer.ts';
@@ -175,6 +176,14 @@ export async function runPhaseSynthesizeConcepts(
     }
   }
 
+  // Honour the documented per-task routing. Without an explicit model this
+  // call inherits models.chat, so models.dream.synthesize (advertised in the
+  // routing table as tier.reasoning) had no effect on this path.
+  const synthModel = await resolveModel(engine, {
+    configKey: 'models.dream.synthesize',
+    tier: 'reasoning',
+    fallback: 'sonnet',
+  });
   for (const group of atomGroups) {
     tierCounts[group.tier]++;
     let narrative: string;
@@ -184,6 +193,7 @@ export async function runPhaseSynthesizeConcepts(
       } else {
         try {
           const result = await chat({
+            model: synthModel,
             system: SYNTH_PROMPT,
             messages: [
               {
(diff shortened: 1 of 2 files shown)


time-attack:
Triage note (automated): routed to maintainer review. Risk-tier scoring: first-time contributor (0 prior merged PRs) with a non-trivial production behavior change (new model resolution wired into the synthesize_concepts chat call), and no test accompanies the production change. Head CI is green; the underlying report (models.dream.synthesize never read on this path) looks plausible from source inspection — resolveModel at src/core/model-config.ts:133 is the documented precedence chain and the narrative chat() call indeed passes no model. Queued rather than auto-merged per triage policy.

time-attack:
Status update: this PR was reviewed in the 2026-08-18 full-backlog triage and confirmed valuable — it sits in the verified merge queue (position 13 of 53, risk: low). CLEAN green. Wires resolveModel into synthesize_concepts; hermetic pin test. No action needed from you right now — thank you for the contribution and the patience.

time-attack:

Issue #3661, opened 2026-07-30
config set silently ignores an unknown flag and writes anyway (--dry-run persists the mutation)


exGeni:
## Summary

`gbrain config set` accepts a flag it does not implement, ignores it silently, and performs the write anyway. A caller passing `--dry-run` — a flag several other gbrain commands do implement — gets a persisted mutation instead of a preview or an error.

## Reproduction

```
$ gbrain config set models.tier.subagent 'claude-cli:probe-invalid' --dry-run
Set models.tier.subagent = claude-cli:probe-invalid

$ gbrain config get models.tier.subagent
claude-cli:probe-invalid
[config] source: db plane
```

The value was written to the DB config plane. `--dry-run` produced no warning, no error, and no change in behaviour.

## Why this shape matters here

`--dry-run` is implemented across the CLI — `sync`, `import`, `extract`, `frontmatter validate`, `quarantine scan`, `pages purge-deleted` all honour it. An operator who has learned that habit will reasonably assume it applies to `config set`. Model-routing keys are exactly where a preview is worth having: `models.default` is consulted ahead of `models.tier.*`, so a wrong value there silently overrides every tier.

I hit this while probing whether a provider string was valid, and persisted a bogus routing value into a live config as a result. It was noticed and reverted within a minute, but nothing in the output suggested anything had happened.

## Suggested fix

Reject unknown flags on `config set` rather than dropping them. Either exit non-zero with "unknown flag", or implement `--dry-run` to match the rest of the CLI.

## Scope of the claim

This report asks for unknown-flag rejection. It does **not** argue that `config set` should validate provider or model names against the recipe registry — arbitrary values are legitimate for many config keys, and that is a separate design question.

## Environment

gbrain 0.42.67.0, commit c6dc0adf26a2d20df1147d2ec87c8922ca86d410, Linux, Postgres engine.

time-attack:
## Verified — reproduced against current master

**Severity: P1**  ·  confidence: high

[...]

PR #3720, opened 2026-08-01
fix(cli): reject unknown flags on config set instead of writing anyway (#3661)

Masashi-Ono0611:
Addresses #3661.

## What this changes

`gbrain config set <key> <value> --dry-run` dropped the unimplemented flag, printed the usual `Set <key> = <value>` confirmation, and persisted the mutation to the DB config plane — no warning, no error. `--dry-run` is honored by `sync`, `import`, `extract`, `frontmatter validate`, `quarantine scan` and `pages purge-deleted`, so an operator carrying that habit over got a live write where they expected a preview. Model-routing keys are the worst place for that, since `models.default` is consulted ahead of `models.tier.*`.

`config set` now refuses any flag it does not implement, **before** running any validation or the write:

```
$ gbrain config set models.tier.subagent 'claude-cli:probe-invalid' --dry-run
[config] unknown flag: --dry-run
[config] `gbrain config set` accepts: --force, --coverage-override, --yes.
[config] Nothing was written.
$ echo $?
1
```

The check reads only the argument tail after `<key> <value>`, so a value that happens to start with `-` is never mistaken for a flag. Globals (`--quiet`, `--progress-json`, `--brain`, a parseable `--timeout`, …) are stripped by `parseGlobalFlags` before dispatch and are unaffected.

## Discriminating power (test fails without the fix)

`test/config-set-unknown-flag.serial.test.ts` spawns the real CLI against a throwaway PGLite brain in a temp `GBRAIN_HOME` and asserts on behavior — exit code plus what `config get` reports afterwards. The write-didn't-happen assertion is the load-bearing one: an implementation that errored *after* `engine.setConfig` would still pass an exit-code-only test.

**Without the fix** (`src/commands/config.ts` reverted to master, test file unchanged):

```
$ bun test test/config-set-unknown-flag.serial.test.ts --timeout=240000
114 |     const set = await runCli(['config', 'set', KEY, 'claude-cli:probe-invalid', '--dry-run']);
115 |     expect(set.exitCode).not.toBe(0);
                                   ^
error: expect(received).not.toBe(expected)

Expected: not 0

(fail) config set rejects unknown flags instead of writing anyway > the reported case: --dry-run is refused and nothing is persisted [1309.62ms]
130 |     const set = await runCli(['config', 'set', KEY, 'anthropic:claude-opus-4-1', '--bogus']);
131 |     expect(set.exitCode).not.toBe(0);
                                   ^
error: expect(received).not.toBe(expected)

Expected: not 0

(fail) config set rejects unknown flags instead of writing anyway > any unimplemented flag is refused, not just --dry-run [1275.67ms]

 4 pass
 2 fail
 23 expect() calls
Ran 6 tests across 1 file. [11.75s]
EXIT=1
```

**With the fix:**

```
$ bun test test/config-set-unknown-flag.serial.test.ts --timeout=240000
 6 pass
 0 fail
 30 expect() calls
Ran 6 tests across 1 file. [13.96s]
EXIT=0
```

The 4 tests that pass on both trees are the regression pins: the no-flag control still writes, and each of the three implemented flags (`--force`, `--coverage-override`, `--yes`) still writes.

## Scope

Per the issue's "Scope of the claim" section, this implements **unknown-flag rejection only**:

- **Not** implementing `--dry-run` for `config set` — the issue offers it as an alternative, and it is a separate feature.
- **Not** validating provider or model names against the recipe registry — the issue explicitly rules that out as a separate design question, and arbitrary values are legitimate for many config keys.
- **No change to the shared/global flag parser** (`src/core/cli-options.ts`) or to any other command. The gate lives inside the `config set` branch of `runConfig`.

## Verification

- `bun test test/config-set-unknown-flag.serial.test.ts` — 6 pass
- `bun test test/config-set.test.ts test/config.test.ts test/config-get-plane.test.ts test/config-unset.test.ts test/config-show-object-values.test.ts` — 65 pass
- `bun run verify` — 34/34 checks green (includes `tsc --noEmit`)
- `bun build --compile --outfile bin/gbrain src/cli.ts` — clean


Masashi-Ono0611:
My human part:
When i often call config set, i found config set didn’t have —dry-run while other ones had. So, i made this that the message would show unknown flag: --dry-run.

Screenshot:
Screenshot 2026-08-02 at 0 36 53 PM

time-attack:
Automated triage: two independent adversarial reviews split on this one. Reviewer A demonstrated empirically that `config set <key> --dry-run <value>` (flag before value) still exits 0 on this branch and persists the literal string `--dry-run` as the config value — the gate only inspects the argument tail after `<key> <value>`, so the silent-write class from #3661 remains reachable with a natural argument order. Reviewer B considered that pre-existing/out-of-scope. A split verdict routes to maintainer review rather than auto-merge. The fix as written is otherwise sound (bug reproduced on master, hermetic regression test, allowlist complete).

Masashi-Ono0611:
Following up on the split verdict: the hole Reviewer A demonstrated is closed in `f5dc925a`, pushed after that triage comment.

The first commit only inspected the argument tail after `<key> <value>`, so a natural flag-before-value order (`config set <key> --dry-run <value>`) still exited 0 and persisted the literal `--dry-run` as the value. `f5dc925a` moves the unknown-flag rejection ahead of value binding, so a `--flag` in any position is refused before anything is written. The allowlist is unchanged; only the scan position moved.

Live check, with this branch's change running on a real brain (postgres engine, ~2.7k pages) on top of current master `a948dfd`:

```
$ gbrain config set self_upgrade.mode --dry-run off
[config] unknown flag: --dry-run
[config] `gbrain config set` accepts: --force, --coverage-override, --yes.
[config] Nothing was written.
$ echo $?
1

$ gbrain config get self_upgrade.mode
off
[config] source: db plane
```

So the reachable-by-natural-argument-order case now exits non-zero and leaves the stored value untouched, and the ordinary `config set <key> <value>` path is unaffected (exit 0, written and read back).

CI on `f5dc925a` is green: 22 pass, 1 skipping.

Masashi-Ono0611:
Rebased onto current master (was 7 behind). No conflicts;
`src/core/cli-flag-registry.generated.ts` auto-merged and a fresh
`bun run build:flag-registry` produced no diff.

**On the red `test (8)`: it is not this branch.** The failing assertion is

```
(fail) startCycleLockRefresher (Tier-1 #1 + D5.11) >
       aborts the controller with LockStolenError when a fenced refresh returns false
error: expect(received).toBeInstanceOf(expected)
Expected constructor: [class LockStolenError extends Error]
Received value: undefined
```

That is `test/db-lock-fencing.test.ts`, master's cycle-lock fencing test from
#4127 — nothing this PR touches (`config set` unknown-flag validation). It is
the defect filed as #4140: the runtime intermittently delivers
`signal.aborted === true` with `signal.reason === undefined`, so the
`instanceof` assertion sees `undefined`.

Reproduced locally on this branch's merged tree, three consecutive runs of that
one file: **fail / fail / pass**. So it reproduces here at roughly the rate #4140
documents, and it reproduces because the branch inherits master's test, not
because of anything in the diff.

#4141 is the fix (keys the steal decision on the `aborted` flag rather than on
the reason surviving, and makes this exact assertion capture the reason through
an abort listener). Once that lands, this shard should stop flapping. Happy to
re-run CI here after it merges.

The rest of this PR's own surface is green: `bun run typecheck` clean,
`test/cli-flag-validation.test.ts` 25 pass / 0 fail on the merged tree.

time-attack:
Status update: this PR was reviewed in the 2026-08-18 full-backlog triage and confirmed valuable — it sits in the verified merge queue (position 10 of 53, risk: low). CLEAN green, fixes #3661. Pair with 4100 (both edit commands/config.ts) — merge first. No action needed from you right now — thank you for the contribution and the patience.

time-attack:

Issue #3686, opened 2026-07-31
`gbrain eval|storage|reindex --help` print a one-line stub that hides 15 eval subcommands, reindex's --multimodal mode, and eval's own shipped help text


andreineacsu:
## Summary

Three CLI_ONLY commands short-circuit `--help` to a generic one-line stub instead of their real usage. For `eval` this is worst: a complete help block ships in the binary and `--help` is the only way you cannot see it.

This is the untreated residue of #578 (`gbrain sync --help`). That fix added `sync` to `CLI_ONLY_SELF_HELP`; `eval`, `storage`, and `reindex` were left behind.

## What happens

```
$ gbrain storage --help
Usage: gbrain storage

gbrain storage - run gbrain --help for the full command list.

$ gbrain eval --help
Usage: gbrain eval

gbrain eval - run gbrain --help for the full command list.

$ gbrain reindex --help
Usage: gbrain reindex

gbrain reindex - run gbrain --help for the full command list.
```

All exit 0. (`gbrain capture --help` is fine — it prints its full block, because `capture` is already in `CLI_ONLY_SELF_HELP`.)

## Expected

Each prints its real usage: subcommands for `eval` and `storage`, target flags for `reindex`.

## Why this hides a real surface

**`eval` — 14 subcommands in `src/commands/eval.ts`** (`brainstorm`, `code-retrieval`, `compare`, `conversation-parser`, `cross-modal`, `export`, `gate`, `prune`, `replay`, `retrieval-quality`, `run-all`, `suspected-contradictions`, `trajectory`, `whoknows`), plus `takes-quality` routed at `src/cli.ts:1886` — 15 total, none discoverable.

`eval` already parses its own help flag and has a renderer, both unreachable:

- `src/commands/eval.ts:198` — `case '--help': case '-h': opts.help = true; break;`
- `src/commands/eval.ts:117-120` — `if (opts.help) { printHelp(); return; }`

The dispatcher returns before `runEvalCommand` is ever called. You can only reach `printHelp()` by triggering the missing-required-arg error path:

```
$ gbrain eval
Error: --qrels <path|json> is required

gbrain eval — measure and compare retrieval quality

USAGE
  gbrain eval --qrels <path>
  gbrain eval --qrels <path> --config-a <path> --config-b <path>

OPTIONS
  --qrels <path|json>         Path to qrels JSON file (required)
  --config-a <path|json>      Config for strategy A (default: hybrid with defaults)
[...]

PR #3721, opened 2026-08-01
fix(sources): gate harden --dry-run pull step on dryRun flag

### LINKED ISSUE #3692: sources harden --dry-run performs a live git fetch + pull --rebase before any dry-run guard
## What happens

`gbrain sources harden <id> --dry-run` is not read-only. `hardenBrainRepo` runs step 1
unconditionally before any `dryRun` check, and step 1 is a live network + working-tree operation.

`src/core/brain-repo-durability.ts:673-676` calls `divergenceSafePull(...)` before the
`dryRun` guards that gate steps 2-7. `divergenceSafePull` (`src/core/git-remote.ts:489-531`)
runs `git fetch` followed by `git pull --rebase` against the user's brain working tree.

So a user reaching for `--dry-run` to see what hardening WOULD do gets a real fetch and a real
rebase-pull of their brain repository. On a repo with local commits not yet pushed, that is a
history-touching operation performed by a flag whose entire purpose is to avoid performing
operations.

## Expected

`--dry-run` performs no network calls and no working-tree mutations. Either move the pull inside
the same `dryRun` guard the other six steps use, or report it as "would pull" in the plan.

## Where it is

- `src/core/brain-repo-durability.ts:668-712` -- the orchestrator; step 1 at :673-676 precedes
  every `dryRun` check
- `src/core/git-remote.ts:489-531` -- `divergenceSafePull`, the fetch + `pull --rebase`

## Notes

Two smaller things found in the same read, mentioned in case they are one fix:

- `gbrain sources harden --help` and `gbrain sources --help` both print only the generic
  `gbrain sources - run gbrain --help for the full command list.` stub, so the command's flags
  (including `--dry-run`) are undiscoverable from its own help. This is the same
  CLI-self-help class as the report filed as #3686.
- The credential step stores an https PAT. On a source whose remote is an SSH URL that store is
  inert, and the harden run still reports the step as done.

## Version

gbrain 0.42.67.0. Established by reading the dispatch (`src/commands/sources.ts:1365-1367` ->
`src/commands/sources-harden.ts:53` `runHarden`) and the orchestrator source. I did NOT execute
`harden --dry-run` against a live brain -- precisely because of the behavior reported here.

gregario:
Fixes #3692.

`hardenBrainRepo` runs `divergenceSafePull(repoPath, branch)` at `src/core/brain-repo-durability.ts:677` before any `dryRun` check, so `gbrain sources harden <id> --dry-run` performs a live `git fetch` + `git pull --rebase` against the user's brain working tree. On a repo with local unpushed commits, that rewrites history under a flag whose entire purpose is to avoid performing operations. Steps 2-7 all gate on `dryRun`; step 1 was missed.

**Before and after** on `f07e9dff`:

| scenario | before | after |
|---|---|---|
| `harden <id> --dry-run` with origin ahead | pull executes, HEAD advances, `upstream.md` materializes | pull skipped, HEAD unchanged, report says "would fetch + pull" |
| `harden <id> --dry-run` with local unpushed commits + origin ahead | local commit rebased to new SHA | HEAD unchanged, no history rewrite |

**The fix** adds an `else if (dryRun)` branch at line 676 that reports the pull step as `skipped` with a descriptive detail string, matching the step-7 verify pattern at line 709-710. The existing detached-HEAD guard (line 673) is untouched. I deliberately did not address the two side-notes in the issue (the `--help` stub, filed separately as #3686; and the SSH-remote/PAT inertness) — those are independent concerns.

**Receipts.** The added test fails on unpatched master with `Expected: "<sha-before>", Received: "<sha-after>"` (HEAD advanced) and passes after. `bun test test/brain-repo-durability.serial.test.ts -t "dry-run"` is 2 pass / 0 fail. `tsc --noEmit` clean. `bun run verify` green (34/34 checks). The pre-existing failure in "installs hook" is environment-dependent (push-probe can't push with `http.followRedirects=false` in a bare-local test setup) and fails identically on unpatched master — confirmed not introduced by this change.

Title carries no version prefix: the landing version is /ship's at land time, not something a contributor can know.

### DIFF
--- a/src/core/brain-repo-durability.ts
+++ b/src/core/brain-repo-durability.ts
@@ -890,6 +890,8 @@ export async function hardenBrainRepo(opts: HardenOpts): Promise<DurabilityRepor
   // Refuse on detached HEAD — pushing to a wrong ref is worse than not pushing.
   if (currentBranch(repoPath) === 'HEAD') {
     push('pull', { status: 'needs_attention', detail: 'detached HEAD — checkout a branch before hardening' });
+  } else if (dryRun) {
+    push('pull', { status: 'skipped', detail: `dry-run — would fetch + pull --rebase origin/${branch}` });
   } else {
     // 1. pull current state
     try { push('pull', pullDetail(divergenceSafePull(repoPath, branch))); }
--- a/src/core/brain-repo-durability.ts
+++ b/src/core/brain-repo-durability.ts
@@ -672,6 +672,8 @@ export async function hardenBrainRepo(opts: HardenOpts): Promise<DurabilityRepor
   // Refuse on detached HEAD — pushing to a wrong ref is worse than not pushing.
   if (currentBranch(repoPath) === 'HEAD') {
     push('pull', { status: 'needs_attention', detail: 'detached HEAD — checkout a branch before hardening' });
+  } else if (dryRun) {
+    push('pull', { status: 'skipped', detail: `dry-run — would fetch + pull --rebase origin/${branch}` });
   } else {
     // 1. pull current state
     try { push('pull', pullDetail(divergenceSafePull(repoPath, branch))); }
--- a/test/brain-repo-durability.serial.test.ts
+++ b/test/brain-repo-durability.serial.test.ts
@@ -172,6 +172,27 @@ describe('hardenBrainRepo', () => {
     expect(existsSync(join(work, 'scripts', 'brain-commit-push.sh'))).toBe(false);
   });
 
+  test('dry-run does not fetch or pull from origin', async () => {
+    const secondClone = mkdtempSync(join(root, 'pusher-'));
+    execFileSync('git', ['-c', 'protocol.file.allow=always', 'clone', '-q', bare, secondClone], { stdio: 'ignore' });
+    execFileSync('git', ['-C', secondClone, 'config', 'user.email', 't@t.t'], { stdio: 'ignore' });
+    execFileSync('git', ['-C', secondClone, 'config', 'user.name', 'tester'], { stdio: 'ignore' });
+    writeFileSync(join(secondClone, 'upstream.md'), 'new upstream content\n');
+    execFileSync('git', ['-C', secondClone, 'add', 'upstream.md'], { stdio: 'ignore' });
+    execFileSync('git', ['-C', secondClone, 'commit', '-qm', 'advance origin'], { stdio: 'ignore' });
+    execFileSync('git', ['-c', 'protocol.file.allow=always', '-C', secondClone, 'push', '-q', 'origin', 'main'], { stdio: 'ignore' });
+
+    const headBefore = git(work, 'rev-parse', 'HEAD');
+    const trackingBefore = git(work, 'rev-parse', 'refs/remotes/origin/main');
+    const report = await harden({ dryRun: true });
+
+    expect(git(work, 'rev-parse', 'HEAD')).toBe(headBefore);
+    expect(existsSync(join(work, 'upstream.md'))).toBe(false);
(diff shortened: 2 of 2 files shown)


Masashi-Ono0611:
I had independently written the same fix and have closed mine (#3722) in favour of this — yours is smaller and its test lives in the existing file, which is the right shape.

Four assertions my version made that this test does not, offered in case any are worth folding in. They need no fixtures beyond the bare remote you already build:

- `git rev-parse refs/remotes/origin/main` unchanged. Asserting HEAD alone still passes if `fetch` ran and only `pull` was skipped; the tracking ref catches that.
- `FETCH_HEAD` absent. More direct evidence that no network call happened at all, since a bare `git fetch` writes it.
- reflog unchanged. Catches a pull that resolved to a no-op fast-forward on a repo whose HEAD already matched.
- the non-dry-run path still pulls. A regression pin, so a later refactor cannot satisfy the dry-run test by disabling the pull outright.

The first two are the ones I would actually add; the last two are belt-and-braces. Entirely your call, and none of it blocks this PR.

Two notes from working the same issue:

- I can independently confirm the `installs hook` failure you mention is pre-existing — it fails identically for me on unpatched master.
- While reading this function I found a second instance of the same class in step 4: `installHelper()` runs `chmodSync(helperPath, 0o755)` above its `dryRun` guard, on the branch taken when the helper is already current, so `--dry-run` still changes a file mode. It survives this fix, so I filed it as #3736 rather than widening your scope.

gregario:
Thanks for the kind words and the detailed response. The two stronger assertions are a good call — pushed them in 23ce585:

- `refs/remotes/origin/main` unchanged (catches a sneaky fetch that skips pull)
- `FETCH_HEAD` absent (direct proof no network call happened)

Left out the other two — agree they're belt-and-braces.

Good catch on the `chmodSync` in `installHelper()` too — glad it has its own issue now (#3736).

gregario:
Heya - caught the new requirements for contributions. I have the above pending and a few other small fixes I did yesterday, I can repro the issues and put screenshots if you'd like? 

Happy to help!

Masashi-Ono0611:
@gregario — you don't need the screenshots. Those contribution requirements were withdrawn two days
after you asked, in #3805 (`revert(ci): remove the PR gate and withdraw the v0.42.72.1 contribution
requirements`, merged 2026-08-04), and the current `CONTRIBUTING.md` on master no longer carries the
section. So the thing your question was about no longer exists — sorry you've been holding for an
answer to it.

One other thing you may not have seen, since the PR has been quiet: it now shows as `CONFLICTING`
against master. Nothing to do with the change itself, just drift — master has moved a lot since
08-02.

To be clear about what I am and am not saying: I'm another contributor, not a maintainer. I can
tell you the requirement is gone because that's checkable, but I can't speak to review or merge
intent here.

Disclosure of my interest, since it's the reason I noticed: I closed my own duplicate of this fix
(#3722) in favour of yours back on 08-01, and I've been carrying your change as a fork-local patch
since then, so this PR has been on my watch list.

gregario:
Thanks for the nudge. I’ve rebased this onto master and kept the no-fetch/no-pull assertions. The separate chmod dry-run fix has landed, so this is now just the pull fix.

time-attack:
Status update: reviewed in the 2026-08-18 full-backlog triage — still wanted, but not yet mergeable: Verified data-safety bug (dry-run runs live fetch+pull --rebase on the brain repo) with a correct 2-line fix + tests, but CI verify fails on a stale test manifest. If you update it, it will be re-reviewed promptly; alternatively the fix may be adopted with credit in an upcoming fix wave. Thanks for the contribution and the patience.

time-attack:

PR #3722, opened 2026-08-01
fix(durability): gate the harden pull step on dryRun (#3692)

Masashi-Ono0611:
Addresses #3692.

## What was broken

`hardenBrainRepo` ran step 1 — `divergenceSafePull(repoPath, branch)`, which is a live `git fetch` followed by `git pull --rebase` — before any dry-run check, while steps 2-7 all took `dryRun`. So `gbrain sources harden <id> --dry-run` reached the network and could rewrite local history: with an unpushed local commit and an advanced origin, the "preview" rebased the commit onto a new SHA.

## What changed

Step 1 now takes the same gate as every other step and reports what it would have done, mirroring the step-7 verify pattern. One conditional; no signature changes; the detached-HEAD refusal in front of it is untouched.

```
  } else if (dryRun) {
    push('pull', { status: 'skipped', detail: `dry-run — would fetch + pull --rebase origin/${branch}` });
  } else {
```

Observed CLI output on the same scratch repo, before and after:

```
before:  + pull   advanced 857c43b→da474c0     (upstream.md materialized, HEAD moved)
after:   · pull   dry-run — would fetch + pull --rebase origin/main
```

## The test fails without the fix

`test/harden-dry-run-pull-guard-3692.serial.test.ts` spawns the real CLI (no source-text matching) against a throw-away `GBRAIN_HOME` brain, with a scratch bare remote advanced behind the registered source's back. The load-bearing assertions are git-state observations that only hold if no network call happened: the remote-tracking ref, the absence of `FETCH_HEAD`, HEAD, the reflog, and the fact that the origin-only file never materialized. A second case pins that the pull still runs without `--dry-run`.

Removing the guard (`git apply -R` of the source hunk only, test untouched) and re-running:

```
$ bun test test/harden-dry-run-pull-guard-3692.serial.test.ts
EXIT_WITHOUT_FIX=1
bun test v1.3.8 (b64edcb4)

test/harden-dry-run-pull-guard-3692.serial.test.ts:
133 |     expect(git(dry.work, 'rev-parse', 'refs/remotes/origin/main')).toBe(trackingBefore);
                                                                        ^

Expected: "418aaa95c7a590a3feeba01e0837366167f7333b"
Received: "052368b86c7b40792c7a9f5ff1aa517eea28df43"

      at <anonymous> (test/harden-dry-run-pull-guard-3692.serial.test.ts:134:68)
(fail) gbrain sources harden --dry-run (#3692) > does not fetch, pull, or move the repo — and says what it WOULD do [868.05ms]

 1 pass
 1 fail
 7 expect() calls
Ran 2 tests across 1 file. [6.43s]
```

Restoring the guard and re-running the identical command:

```
$ bun test test/harden-dry-run-pull-guard-3692.serial.test.ts
EXIT_WITH_FIX=0
bun test v1.3.8 (b64edcb4)

 2 pass
 0 fail
 12 expect() calls
Ran 2 tests across 1 file. [5.81s]
```

The regression case passes in both states by design — it guards the non-dry-run path, which this change does not touch.

## Checks

```
bun run typecheck                       EXIT=0
bun run verify                          EXIT=0   (34/34 checks green)
bun run build                           EXIT=0
bun test test/brain-repo-durability.serial.test.ts \
         test/git-remote-durable.serial.test.ts \
         test/brain-durability-hook.serial.test.ts \
         test/durability-cron.test.ts   EXIT=0   (42 pass / 0 fail)
```

Test hygiene: `.serial` matches the other PGLite-spawning suites; hooks and tests carry explicit 60s budgets per #3566; global and system git config are neutralized on both the fixture side and inside the spawned CLI so an operator's `core.hooksPath` / `url.*.insteadOf` cannot perturb the run; the CLI subprocess has a timeout and surfaces `error`/`signal`, and runs from the scratch root so no ancestor `.gbrain-mount` can reroute it.

## Scope

Deliberately limited to the pull step. The two adjacent items noted in the issue — the generic `sources harden --help` stub (attributed to #3686) and the inert https PAT wired onto SSH-remote sources — are separate problems and are not touched here.

The doc touch is the `brain-repo-durability.ts` entry in `docs/architecture/KEY_FILES.md`, updated to current state: the pull step is now described as gated, and the new test file is listed.


Masashi-Ono0611:
Closing in favour of #3721, which was opened two minutes before this one and is the better PR by this repo's own bar.

It is the same fix — both add the `else if (dryRun)` branch with the detail string the issue thread specified. The difference is size: #3721 is +25 across two files and puts its test in the existing `test/brain-repo-durability.serial.test.ts`; this one is +163 across three because I added a new test file and touched `KEY_FILES.md`. RELEASING.md says the smaller diff wins when the fix is the same, and that is the right call here.

I have left the few assertions my test made that this one does not as a comment on #3721, to take or leave. No action needed here.

Thanks @gregario.

Issue #3602, opened 2026-07-29
Feature request: C# symbol-level chunks and call edges (code_def/code_callers empty for C# codebases)


bry-uips:
Distinct from #3601, and understood to be a deliberate scope line —
`edge-extractor.ts` documents call-edge support for TypeScript, TSX, JavaScript,
Python, Go, Rust, and Java, with "other languages flow through with zero edges."

Even with #3601 fixed, C# yields one chunk per namespace and no edges, so
`code_def` / `code_callers` / `code_flow` / `code_blast` stay empty for C#
codebases. Java nests essentially identically to C# and is already configured,
so the delta is small:

1. `BODY_NODE_TYPES` += `declaration_list` — C#'s namespace/class body node, so
   `collectImmediateNestedChildren` can descend (it currently only enters
   `*_body` / the listed set).
2. `NESTED_EMIT_CONFIG.c_sharp` — parents: `namespace_declaration`,
   `file_scoped_namespace_declaration`, `class_declaration`,
   `interface_declaration`, `struct_declaration`, `record_declaration`;
   children: `method_declaration`, `constructor_declaration`,
   `property_declaration`.
3. `CALL_CONFIG.c_sharp` = `{ callNodeTypes: new Set(['invocation_expression']),
   calleeFieldName: 'function' }`.

Measured on our three repos (274 `.cs` files), with #3601 also fixed:

| | before | after |
|---|---:|---:|
| chunks from a 559-line file | 2 | 48 |
| named symbols (whole brain) | 14 | 2323 |
| C# `calls` edges | 0 | 4245 |

Qualified names come out as expected, e.g.
`UIPS.Dynamo.DynamoRepository.SaveAsync`.

`bun test test/chunkers/{code,recursive}.test.ts` and
`test/{edge-extractor,code-edges,code-def-refs,code-graph-readiness,search-lang-symbol-kind}.test.ts`
— **105 pass, 0 fail.** Worth noting item 1 touches a set shared by every
language, which is why those suites were run rather than only the C# path.

C# stays outside `RECEIVER_RESOLUTION_LANGS`, so edges are bare-token — matching
the existing Java/Go/Rust behaviour rather than claiming more than is tested.

Happy to open a PR with the above if the scope is welcome.

---

## Our patch

`deploy/patches/0001-csharp-symbols-and-call-edges.patch` (13 lines) implements
both, and applies cleanly at `5008b287`. It is applied at Docker build time until
this lands upstream.

PR #3723, opened 2026-08-01
fix(chunkers): index C# file-scoped namespace body instead of silently dropping it

### LINKED ISSUE #3601: C# files using a file-scoped namespace are indexed as empty (whole body silently dropped)
**Version:** 0.42.59.0 (`5008b287e47bf791132eedfebf66bdef11e9398c`)

### Summary

For any C# file using the C# 10 file-scoped namespace form (`namespace Foo;`),
the entire file body is silently dropped from the index. Only the leading `using`
directives are chunked. No warning, no sync failure — the page is created and
looks healthy.

### Cause

`TOP_LEVEL_TYPES.c_sharp` in `src/core/chunkers/code.ts` includes
`namespace_declaration` but not `file_scoped_namespace_declaration`. Since
`chunkCodeTextFull` selects semantic nodes with

```js
root.namedChildren.filter((n) => topLevelTypes.has(n.type))
```

…and in the file-scoped form every declaration in the file is a *child* of the
unmatched `file_scoped_namespace_declaration` node, nothing below it is ever
visited. The block-scoped form happens to work because `namespace_declaration`
*is* listed and its node spans the whole body.

This is very likely just an oversight from before C# 10 — the block-scoped form
is handled, the file-scoped one was missed.

### Reproduction

`Repro.cs`:

```csharp
using System;
namespace Demo;

public class OrderService
{
    public decimal ComputeRefund(decimal amount) => amount * 0.9m;
    public void Ship(string id) => Console.WriteLine(id);
}
```

```js
const { chunks } = await chunkCodeTextFull(source, 'Repro.cs');
// chunks cover only the `using` line; OrderService is absent.
```

Real-world: a 559-line file in our codebase produced **2 chunks covering lines
1-12**. The interface and class — lines 14-560 — were not in the brain at all.

Add `'file_scoped_namespace_declaration'` to `TOP_LEVEL_TYPES.c_sharp` and the
body is recovered.

### Impact

This degrades ordinary semantic search, not just the code-graph surfaces. In our
corpus **234 of 289 `.cs` files (81%)** use the file-scoped form, so the great
majority of our C# was absent from an index that reported itself healthy —
`doctor` is all-green and the pages exist with plausible chunk counts. We only
found it while investigating why `code_def` returned nothing.

Suggested extra: since a `.cs` file that yields zero named symbols is almost
always a parse/config gap rather than a genuinely empty file, a debug-level
warning would have surfaced this immediately.

---

## Our patch

`deploy/patches/0001-csharp-symbols-and-call-edges.patch` (13 lines) implements
both, and applies cleanly at `5008b287`. It is applied at Docker build time until
this lands upstream.

gregario:
Fixes #3601.

`TOP_LEVEL_TYPES.c_sharp` at `src/core/chunkers/code.ts:324-328` lists `namespace_declaration` but not `file_scoped_namespace_declaration`. In a C# 10 file using the file-scoped form (`namespace Demo;`), the AST root's `namedChildren` are `[using_directive, file_scoped_namespace_declaration]`. Since `file_scoped_namespace_declaration` isn't in the set, the filter at `code.ts:623-624` (`root.namedChildren.filter(n => topLevelTypes.has(n.type))`) only matches `using_directive` — the entire class body inside the namespace node is never visited. The page is created and appears healthy (doctor is green, chunks exist), but contains only `using` lines.

**Before and after** on `f07e9dff`:

| scenario | before | after |
|---|---|---|
| `chunkCodeText` on file-scoped `namespace Demo;` with `OrderService` class | 1 chunk: `[C#] OrderService.cs:1-1 using directive System` — class body absent | chunks include `OrderService` and `ComputeRefund` |
| `chunkCodeText` on block-scoped `namespace Demo { ... }` | class body indexed (already worked) | unchanged |

**The fix** adds `'file_scoped_namespace_declaration'` to the `TOP_LEVEL_TYPES.c_sharp` set — one string in a `new Set([...])` initializer. This makes the file-scoped form produce the same namespace-body chunk the block-scoped `namespace_declaration` already produces. C# has no `NESTED_EMIT_CONFIG` entry, so no nested-emission interaction exists. I deliberately did not address the issue's secondary asks (debug warning for zero-symbol .cs files, call-edge extraction) — those overlap with feature request #3602 and are a distinct scope.

**Receipts.** The added test fails on unpatched master with `Expected to contain: "ComputeRefund", Received: "[C#] OrderService.cs:1-1 using directive System\n\nusing System;"` and passes after. `bun test test/chunkers/code.test.ts` is 33 pass / 0 fail. `tsc --noEmit` clean. `bun run verify` green (34/34 checks).

### DIFF
--- a/src/core/chunkers/code.ts
+++ b/src/core/chunkers/code.ts
@@ -324,7 +324,7 @@ const TOP_LEVEL_TYPES: Partial<Record<SupportedCodeLanguage, Set<string>>> = {
   c_sharp: new Set([
     'method_declaration', 'class_declaration', 'interface_declaration',
     'struct_declaration', 'enum_declaration', 'namespace_declaration',
-    'using_directive', 'property_declaration',
+    'file_scoped_namespace_declaration', 'using_directive', 'property_declaration',
   ]),
   cpp: new Set([
     'function_definition', 'class_specifier', 'struct_specifier',
--- a/test/chunkers/code.test.ts
+++ b/test/chunkers/code.test.ts
@@ -424,3 +424,22 @@ describe('chunkCodeText — empty input', () => {
     expect(await chunkCodeText('   \n  ', 'foo.ts')).toEqual([]);
   });
 });
+
+describe('chunkCodeText — C# file-scoped namespace (#3601)', () => {
+  test('indexes the class body inside a file-scoped namespace', async () => {
+    const source = `using System;
+namespace Demo;
+
+public class OrderService
+{
+    public decimal ComputeRefund(decimal amount) => amount * 0.9m;
+}
+`;
+
+    const result = await chunkCodeText(source, 'OrderService.cs');
+    const text = result.map(chunk => chunk.text).join('\n');
+
+    expect(text).toContain('OrderService');
+    expect(text).toContain('ComputeRefund');
+  });
+});
--- a/test/chunkers/code.test.ts
+++ b/test/chunkers/code.test.ts
@@ -424,3 +424,45 @@ describe('chunkCodeText — empty input', () => {
     expect(await chunkCodeText('   \n  ', 'foo.ts')).toEqual([]);
   });
 });
+
+describe('chunkCodeText — C# file-scoped namespace (#3601)', () => {
+  const FILE_SCOPED = `using System;
+namespace Demo;
+
+public class OrderService
+{
+    public decimal ComputeRefund(decimal amount) => amount * 0.9m;
+    public void Ship(string id) => Console.WriteLine(id);
+}
+`;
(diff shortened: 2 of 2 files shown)


time-attack:
Status update: this PR was reviewed in the 2026-08-18 full-backlog triage and confirmed valuable — it sits in the verified merge queue (position 1 of 53, risk: low). P1 CLEAN green. C# bug verified live (code.ts:324). Merge before 4095 (chunkers/code.ts overlap). No action needed from you right now — thank you for the contribution and the patience.

time-attack:

PR #3725, opened 2026-08-01
fix(skills): advertise usable tools on stdio transport

### LINKED ISSUE #3635: skill catalog reports every tool as unavailable over stdio MCP, and tells the agent not to call them
**Version:** 0.42.67.0 (checkout at c6dc0adf, clean)
**Transport:** stdio MCP, registered exactly as `docs/mcp/CLAUDE_CODE.md:11` prescribes — `claude mcp add gbrain -- gbrain serve`

## Symptom

On the stdio transport, `list_skills` returns every skill with `usable_tools: []`,
every declared tool in `unavailable_tools`, and `instructions.available_brain_tools: []`.
The same payload then instructs the agent (`src/core/operations-descriptions.ts:208,224`):

> "Only call tools in this skill's `usable_tools`; tools in `unavailable_tools` are not callable by you on this server."
> "Do not invent tools — only the tools in `usable_tools` are callable by you."

An agent that obeys the catalog concludes it cannot use the brain at all.

## It is not true

Same server, same session, same minute:

- `whoami` → `{"transport":"stdio","scopes":[]}`
- `search`, `get_health`, `list_skills` → all succeed
- `put_page` — a `write`-scope op the catalog listed as unavailable — **succeeded and wrote through to disk**:
  `{"status":"created_or_updated","chunks":1,"write_through":{"written":true,"path":"..."}}`

For contrast, the CLI path on the same skill reports the truth:
`gbrain skill brain-ops` → `usable_tools: [search, query, get_page, put_page, add_link, add_timeline_entry, get_backlinks, sync_brain]`, `unavailable_tools: []`.

## Mechanism

`src/core/skill-catalog.ts:295-299`

```ts
function opCallableByCaller(op: Operation, ctx: OperationContext): boolean {
  if (ctx.remote === false) return true;   // local CLI
  if (op.localOnly) return false;
  return hasScope(ctx.auth?.scopes ?? [], op.scope ?? 'read');
}
```

The stdio pipe is deliberately `remote: true` (`src/core/operations.ts:339-344`) and carries
no auth, so `ctx.auth?.scopes ?? []` is empty and `hasScope([], 'read')` is false for every op.

The catalog is the **only** place that consults scopes on this transport. `hasScope` has three
call sites in the whole repo: OAuth grant issuance, request-time enforcement in
`src/commands/serve-http.ts:1893` (HTTP only), and `skill-catalog.ts:298`. There is no
scope enforcement in the stdio dispatch path, which is why every op executes regardless.

So the advertisement applies an HTTP authorization model to a transport that has no
authorization, and reports the exact inverse of reality.

## Suggested fix

Treat the auth-less stdio transport as trusted for the purposes of this advertisement —
it already is for dispatch. Something like:

```ts
if (ctx.remote === false || ctx.transport === 'stdio') return true;
```

(`OperationContext` already carries the `'stdio'` transport marker per the note at
`src/core/operations.ts:339-344`.)

## Why it matters

This is upstream's own zero-config path — `docs/mcp/DEPLOY.md:18-25` and
`docs/mcp/CLAUDE_CODE.md:11-15`, "No server, no tunnel, no token needed". Any agent that
follows the catalog's instructions on that path silently loses the entire brain surface.

gregario:
Fixes #3635.

`opCallableByCaller()` at `src/core/skill-catalog.ts:295-299` requires `hasScope(ctx.auth?.scopes ?? [], op.scope)` for any remote caller. The stdio MCP pipe is `remote: true` with no auth (`ctx.auth` is undefined), so `ctx.auth?.scopes ?? []` resolves to `[]` and `hasScope([], 'read')` is false — every op fails the check. Meanwhile `src/mcp/dispatch.ts:237` enforces no scopes on stdio, so every op actually executes. The result: `list_skills` reports all tools in `unavailable_tools` with `usable_tools: []`, and the instruction text tells the agent not to call them, while in reality the tools all work.

**Before and after** on `f07e9df`:

| scenario | before | after |
|---|---|---|
| `list_skills` over stdio | `usable_tools: []`, all tools in `unavailable_tools` | `usable_tools` contains all non-localOnly ops (search, put_page, etc.) |
| `localOnly` ops over stdio | Listed unavailable (catalog) but execute (dispatch) | Listed unavailable — catalog agrees with what dispatch *should* enforce |
| HTTP-remote with OAuth scopes | Scope-gated split (read → search/query, admin → all) | Unchanged |

**The fix** adds one line after the `op.localOnly` check: `if (ctx.transport === 'stdio') return true;`. Placed AFTER `localOnly` so those ops remain unadvertised — matching the intended posture that stdio is trusted but not fully local. This is advertisement honesty, not a trust grant: dispatch already runs everything on stdio.

Deliberately left unchanged: the deeper inconsistency where stdio dispatch doesn't enforce `localOnly` (verified in the issue — `purge_deleted_pages` executes over stdio). That's a separate product decision for the maintainer and a different issue.

**Receipts.** The added test fails on unpatched master with `Expected to contain: "search" / Received: []` and passes after. `bun test test/skill-catalog-transports.test.ts` is 10 pass / 0 fail. `bun run verify` is 34/34 clean.

### DIFF
--- a/src/core/skill-catalog.ts
+++ b/src/core/skill-catalog.ts
@@ -295,6 +295,7 @@ export function confineManifestPath(skillsDir: string, entry: ManifestEntry): st
 function opCallableByCaller(op: Operation, ctx: OperationContext): boolean {
   if (ctx.remote === false) return true; // local CLI — OS is the trust boundary
   if (op.localOnly) return false; // not reachable over a remote transport
+  if (ctx.transport === 'stdio') return true; // auth-less local pipe — dispatch enforces no scopes
   return hasScope(ctx.auth?.scopes ?? [], op.scope ?? 'read');
 }
 
--- a/test/skill-catalog-transports.test.ts
+++ b/test/skill-catalog-transports.test.ts
@@ -54,7 +54,7 @@ function unpack(res: { content: { text: string }[]; isError?: boolean }): {
 async function call(
   name: string,
   params: Record<string, unknown>,
-  opts: { remote: boolean; auth?: AuthInfo },
+  opts: { remote: boolean; auth?: AuthInfo; transport?: 'stdio' },
 ) {
   return unpack(await dispatchToolCall(engine, name, params, { sourceId: 'default', ...opts }));
 }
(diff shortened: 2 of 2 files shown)


time-attack:
Merged in the v0.45.8.0 community fix-wave (#4050). Your fix landed on master as 9b6db85d39, credited to you via Co-Authored-By, and is called out in the v0.45.8.0 CHANGELOG entry. Thank you for the contribution!

PR #3727, opened 2026-08-01
fix(budgeting): add Sonnet 5, Fable 5, and Opus 4.8 to the synthesize context map and brainstorm output caps

p3ob7o:
## What

The canonical chat-pricing table knows the current Claude generation (Sonnet 5 and Fable 5 landed in v0.42.61 via #2799), but the two phase-local budgeting maps don't — so current-generation models silently fall to conservative unknown-model defaults:

- **synthesize's `MODEL_CONTEXT_TOKENS`**: a 1M-context Sonnet 5 / Fable 5 / Opus 4.8 gets the unknown-model prompt budget instead of its real window, truncating synthesis input harder than necessary.
- **brainstorm's `ANTHROPIC_OUTPUT_CAPS`**: a 64K-output Sonnet 5 / Fable 5 judge falls to `MAX_OUTPUT_TOKENS_CEIL` (32K), halving the output budget on exactly the models most likely to be configured today.

## Change

Data-only rows in both maps, values per Anthropic's published limits: 1M context for all three; 64K output for Sonnet 5 and Fable 5, 32K for Opus 4.8. No behavior change for any model already in the maps.

## Tests

Cap tests pin the judge formula resolving each new entry — including the provider-prefixed form — instead of the fallback ceiling (the observable difference: a 500-idea batch on Sonnet 5 caps at 64K with the entry vs 32K without). 106 brainstorm-suite tests green; typecheck clean.

These rows ran in production on a live Postgres brain for the past three weeks (they shipped inside the now-closed #2805; this extracts just the budgeting-map slice, which is independent of the declined tier-config feature).

### DIFF
--- a/src/core/brainstorm/judges.ts
+++ b/src/core/brainstorm/judges.ts
@@ -61,6 +61,9 @@ export const MAX_OUTPUT_TOKENS_CEIL = 32_000;
  * (with a readable error) instead of the provider's opaque HTTP 400.
  */
 export const ANTHROPIC_OUTPUT_CAPS: Record<string, number> = {
+  'claude-fable-5': 64_000,
+  'claude-sonnet-5': 64_000,
+  'claude-opus-4-8': 32_000,
   'claude-opus-4-7': 32_000,
   'claude-sonnet-4-6': 64_000,
   'claude-haiku-4-5': 64_000,
--- a/src/core/brainstorm/judges.ts
+++ b/src/core/brainstorm/judges.ts
@@ -62,6 +62,7 @@ export const MAX_OUTPUT_TOKENS_CEIL = 32_000;
  */
 export const ANTHROPIC_OUTPUT_CAPS: Record<string, number> = {
   'claude-fable-5': 64_000,
+  'claude-opus-5': 32_000,
   'claude-sonnet-5': 64_000,
   'claude-opus-4-8': 32_000,
   'claude-opus-4-7': 32_000,
--- a/src/core/cycle/synthesize.ts
+++ b/src/core/cycle/synthesize.ts
@@ -62,6 +62,9 @@ const SUMMARY_SLUG_RE = new RegExp(`^${PAGE_SLUG_SEG}(\\/${PAGE_SLUG_SEG})*$`, '
  * resolver returns for known Anthropic aliases.
  */
 const MODEL_CONTEXT_TOKENS: Record<string, number> = {
+  'claude-fable-5': 1_000_000,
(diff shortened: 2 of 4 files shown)


time-attack:
Automated triage note (gbrain-pr-triage): tests pass and the change is purely additive, but adversarial review raised a concrete completeness gap, so this is queued for maintainer review rather than auto-merged:

1. `claude-opus-5` is shipped in this repo's own chat recipe (`src/core/ai/recipes/anthropic.ts`) and pricing table (`src/core/model-pricing.ts`), but is absent from both maps this PR extends — an Opus 5 user still falls to the 32K output fallback in `judges.ts` and the unknown-model prompt budget in `synthesize.ts`, which is exactly the bug class this PR closes for the other three models.
2. The new `MODEL_CONTEXT_TOKENS` rows in `synthesize.ts` are keyed on bare model ids, and that half of the change has no test pinning that a provider-prefixed id (`anthropic:claude-sonnet-5`) resolves to the new entries rather than the fallback.

Neither point makes the current diff wrong — it may just want one more row and a synthesize-side test.

p3ob7o:
Both triage points addressed in 95d925a3:

1. **`claude-opus-5` added to both maps** — 1M context in `MODEL_CONTEXT_TOKENS`, 32K output cap in `ANTHROPIC_OUTPUT_CAPS` (matching the existing Opus-family rows), with cap tests including the provider-prefixed form.

2. **The requested synthesize-side test uncovered a real gap, now fixed.** `computeChunkCharBudget` did a raw keyed lookup while `judges.ts` strips the provider prefix first — and `resolveModel` returns provider-prefixed strings whenever `TIER_DEFAULTS`/config values carry a prefix (the current tier defaults all do). So every tier-resolved brain silently fell to the 180K unknown-model fallback: coincidentally harmless on 200K models (180K fallback == 200K × 0.9 headroom), but a 5× prompt-budget cut on the 1M-context models this PR adds. The lookup now strips the prefix via `splitProviderModelId`, mirroring judges.ts, and `computeChunkCharBudget` is exported for tests. The new suite pins: all four current-gen models resolve their real budget, prefixed == bare, unknown models still fall back, config override still wins.

111 tests green across the affected suites; typecheck clean.

time-attack:
Merged in the v0.45.8.0 community fix-wave (#4050). Your fix landed on master as cb07cfda8d, credited to you via Co-Authored-By, and is called out in the v0.45.8.0 CHANGELOG entry. Thank you for the contribution!

PR #3735, opened 2026-08-01
fix(sync,import): global sync.* anchors only move for the brain repo they describe (#2114)

### LINKED ISSUE #2114: gbrain import silently overwrites sync.repo_path — dream full-reimport fallback clobbers the configured brain repo
## Version
gbrain 0.42.37.0

## Summary
`src/commands/import.ts:463` runs `await engine.setConfig('sync.repo_path', dir)` as an unconditional side effect of every `gbrain import <dir>`. Any import from a directory that is *not* the canonical brain repo silently repoints write-through and sync at that directory.

## How we got bitten (real incident, 2026-06-12 03:30 nightly dream)
1. `sync.repo_path` correctly pointed at our brain repo (git repo A).
2. The dream cycle's sync phase ran against a different `--dir` (a transcript staging dir, git repo B). Its incremental anchor (`sync.last_commit`) referenced a commit that doesn't exist in repo B → anchor miss → **full reimport fallback**.
3. The full reimport path calls `gbrain import`-equivalent logic → import.ts:463 silently rewrote `sync.repo_path` to repo B.
4. From then on, every `put_page` write-through landed in the wrong directory; the canonical repo silently stopped receiving md mirrors (DB-only divergence). Nothing logged the config change.

## Expected
- `import` should not mutate `sync.repo_path` implicitly. Suggest: only set it when unset (first-run convenience), or behind an explicit flag (`--set-repo-path`), and always log the old → new value when it does change.
- The dream sync full-reimport fallback in particular should never repoint the configured repo.

smdesai27:
## Summary

Fixes #2114 — the global sync anchors (`sync.repo_path` / `sync.last_commit`) silently move to any directory a git-repo import or an unresolved-source sync touches.

The globals describe THE brain repo — the default source's working tree. That identity is already established in the codebase: `source-resolver.ts` documents `sync.repo_path` as the legacy pre-v0.18 default-source key, and migration v16 seeds the `default` source row FROM it. Two paths violated it:

1. **`runImport`'s checkpoint block** wrote `sync.repo_path` + `sync.last_run` unconditionally (and `sync.last_commit` on failure-free imports) for every git-directory import — the direct `gbrain import` surface from #2114.
2. **`writeSyncAnchor`'s legacy no-sourceId branch** did the same for sync-driven writes. It is reached whenever a caller cannot resolve a source for the dir: `gbrain dream --dir <unregistered-dir>` (whose anchor-miss forces the full-reimport fallback — the exact incident in #2114), the minion `sync` job with an unmatched `repoPath`, and MCP `sync_brain`. #1939's `managedBookmark` stops `runImport`'s duplicate write on that path, but `performFullSync` then performs the same global write itself via `writeSyncAnchor`.

Either way: `put_page` write-through silently repoints at the foreign directory (`write-through.ts` falls back to `sync.repo_path` when the source row has no `local_path`), the incremental anchor is poisoned, and nothing logs the change.

## The fix — one ownership predicate, two enforcement points

`ownsGlobalSyncAnchor` (new, `core/sync.ts`) is consulted by BOTH `runImport`'s checkpoint block and `writeSyncAnchor`'s legacy branch, so the two layers cannot drift:

- only the default source (or the legacy no-sourceId path, which is default-by-definition) may move the globals;
- the dir must match the brain repo's configured identity — the **global key when set, else the default source row's `local_path` when set**. Modern sync writes its anchor to the sources row and leaves the global unset, so an unset global alone must not green-light a takeover;
- only a truly fresh brain (neither identity present) bootstraps — the documented first-run flow (`gbrain init` + `gbrain import <dir>`) is unchanged;
- both sides are canonicalized via `realpathSync.native`, so symlinked (`/tmp` vs `/private/tmp`) and case-variant (APFS `/users` vs `/Users`) spellings of the same repo never false-refuse (plain `realpathSync` does not fix case);
- refusals are loud: a stderr notice names the paste-ready `gbrain config set sync.repo_path "<dir>"` for an intentional repoint. **Imports are never blocked — only the bookmark writes are guarded.**

With this change, every write site for the two global anchors sits behind the guard (`grep -rn "setConfig('sync\." src/`). `sync.last_run` remains unguarded: it is write-only telemetry with zero readers in src/.

## Behavior after this change

| Scenario | Before | After |
|---|---|---|
| Fresh brain, first `gbrain import <dir>` | bootstraps globals | unchanged — bootstraps  |
| Re-import of the SAME repo, any path spelling | advances bookmark | unchanged — advances  |
| Import of a DIFFERENT git dir (default source) | **repoints `repo_path` + `last_run` silently; `last_commit` too on clean imports** | moves nothing; loud notice |
| Global unset but `sources.default.local_path` configured, foreign import | **bootstrap arm claims the globals** | refused — the source row is a brain-repo identity too |
| Import with `--source-id <non-default>` (or #1434 sole-source auto-route) | clobbers the globals | never touches them (deliberately quiet — routine multi-source imports shouldn't nag) |
| Sync full-reimport fallback / `dream --dir` on an unregistered dir | **legacy branch clobbers globals** | `writeSyncAnchor` legacy branch refuses + notices |
| `managedBookmark: true` callers | skip block | unchanged |

Deliberately NOT changed, to keep this diff minimal (observed while working here, flagging for follow-up):

- `recordFailures` just above the guard uses `opts.sourceId ?? 'default'` while source resolution can land elsewhere (`--source-id` / auto-route) — so those failures file under `default`. Pre-existing on master; left byte-identical here rather than shipping an untested behavioral change inside a bugfix.
- A machine-readable bookmark outcome on `RunImportResult` (and the minion `import` job result) would surface refusals to programmatic callers; today the notice is stderr-only. Cheap follow-up if wanted.

## Tests

`test/import-repo-path-guard.test.ts` — 7 hermetic PGLite cases: bootstrap; foreign-repo refusal (import still succeeds, notice fires, all three keys held); same-repo advance; **non-canonical configured spelling** (seeds the config with the alternate spelling so the realpath comparison is load-bearing — a `resolve()`-only mutant fails this); **unset-global + default-source-row identity** (the false-accept hole); non-default-source no-op (both unset and configured globals); **`writeSyncAnchor` legacy-branch guard** (foreign refusal for both keys, threaded `repoDir`, sourceId branch untouched).

Hermeticity per repo conventions: `GBRAIN_HOME` via `withEnv` (runImport's checkpoint never touches the real `~/.gbrain` — the pattern from `test/import-resume.test.ts`), `GBRAIN_SOURCE` cleared, git fixtures run with `GIT_CONFIG_GLOBAL=/dev/null` so user hooks/templates can't fire, fixture filenames are repo-unique to avoid cross-source slug collisions.

Ran locally: new file 7/7 (3 consecutive runs, no flake), `bun run verify` 34/34, 284/284 across the 13-file import/sync neighborhood, plus a CLI-level replay of the issue's repro (config → import foreign dir → config unchanged, notice printed, import succeeded, exit 0). Full-suite note: 5 serial failures in `autocut`/`hybrid-meta`/`brain-durability` reproduce identically on clean master in my environment (pre-existing, untouched files); the 4-way parallel shard fan-out wedges on my 16 GB machine — CI is authoritative there.

## Notes for the maintainer

- VERSION / CHANGELOG deliberately untouched, and the PR title carries no version token — per the community-PR wave process, happy to have this re-landed in a collector branch with attribution.
- The issue's "Expected" section floats an explicit `--set-repo-path` flag; this PR uses the stderr notice + existing `gbrain config set` instead (no new flag surface). Easy to add on top if preferred.
- The non-default-source arm is deliberately silent. Consequence for the #1434 sole-non-default topology: a bare import there no longer populates the globals, so raw `getConfig('sync.repo_path')` readers (`takes.ts`, `remediation/context.ts`, `notability-eval`, autopilot's start check) see "no repo configured" where they previously saw the last-imported dir. Readers that go through `getDefaultSourcePath` are unaffected (it checks `sources.local_path` first). Migrating the raw readers to that helper looks like the right follow-up; flagging rather than expanding this diff.
- Credit to the 2026-07-31 triage comment on #2114 for the master-state repro and for confirming no other PR covers this path.

### DIFF
--- a/src/commands/import.ts
+++ b/src/commands/import.ts
@@ -534,8 +534,9 @@ export async function runImport(
 
   // Import → sync continuity: write sync checkpoint if this is a git repo.
   // Bug 9 — gate last_commit on "no failures" so import doesn't silently
-  // stomp on the sync bookmark when parsing broke. We still write
-  // last_run + repo_path either way (those are progress indicators).
+  // stomp on the sync bookmark when parsing broke. last_run + repo_path are
+  // written alongside it, but ONLY when this import owns the globals (#2114
+  // guard below) — a foreign directory must not repoint the brain repo.
   let gitHead: string | null = null;
   try {
     if (existsSync(join(dir, '.git'))) {
@@ -557,17 +558,37 @@ export async function runImport(
       const { recordFailures } = await import('../core/sync.ts');
       recordFailures(opts.sourceId ?? 'default', failures, gitHead);
     }
-    if (failures.length === 0) {
-      await engine.setConfig('sync.last_commit', gitHead);
-    } else {
+
+    // #2114 guard: the global sync.* keys describe THE brain repo (the
+    // default source's working tree). Pre-fix this block rewrote them on
+    // every git-repo import, silently repointing put_page write-through
+    // and poisoning the incremental sync anchor. Ownership + the bootstrap
+    // rule live in ownsGlobalSyncAnchor (shared with writeSyncAnchor's
+    // legacy branch in sync.ts, so the two layers cannot drift).
+    const { ownsGlobalSyncAnchor } = await import('../core/sync.ts');
+    const { owns, configured } = await ownsGlobalSyncAnchor(engine, sourceId, dir);
+
+    if (owns) {
+      if (failures.length === 0) {
+        await engine.setConfig('sync.last_commit', gitHead);
+      } else {
+        console.error(
+          `\nImport completed with ${failures.length} failure(s). ` +
+          `sync.last_commit NOT advanced — re-run 'gbrain sync' to retry, or ` +
+          `'gbrain sync --skip-failed' to acknowledge and move past them.`,
+        );
+      }
+      await engine.setConfig('sync.last_run', new Date().toISOString());
+      await engine.setConfig('sync.repo_path', dir);
+    } else if ((sourceId ?? 'default') === 'default') {
       console.error(
-        `\nImport completed with ${failures.length} failure(s). ` +
-        `sync.last_commit NOT advanced — re-run 'gbrain sync' to retry, or ` +
-        `'gbrain sync --skip-failed' to acknowledge and move past them.`,
+        `\n[import] sync.repo_path stays at ${configured ?? '(unset)'} — NOT repointing to "${dir}". ` +
+        `Sync bookmarks were not advanced. If this directory IS your brain repo, run: ` +
+        `gbrain config set sync.repo_path "${dir}"`,
       );
     }
-    await engine.setConfig('sync.last_run', new Date().toISOString());
-    await engine.setConfig('sync.repo_path', dir);
+    // Non-default sources: deliberately silent no-op — the globals are not
+    // this import's to move (its sync anchors live on the `sources` row).
   }
 
   return { imported, skipped, errors, chunksCreated, failures };
--- a/src/commands/sync.ts
+++ b/src/commands/sync.ts
@@ -19,6 +19,7 @@ import {
   isSkippablePath,
   resolveAutoSkipThreshold,
   DEFAULT_SOURCE_ID,
+  ownsGlobalSyncAnchor,
 } from '../core/sync.ts';
 import {
   computeSyncDelta,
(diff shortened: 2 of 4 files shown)


time-attack:
Merged in the v0.45.8.0 community fix-wave (#4050). Your fix landed on master as 636628fdb2, credited to you via Co-Authored-By, and is called out in the v0.45.8.0 CHANGELOG entry. Thank you for the contribution!

Issue #3738, opened 2026-08-01
integrity: the subcommand comment documents `gbrain integrity --dry-run`, which the CLI rejects


Masashi-Ono0611:
## What happens

The subcommand list in `src/commands/integrity.ts` documents a form the CLI does not accept.

`src/commands/integrity.ts:12-16` on `master` (`f07e9dff`):

```
 * Subcommands:
 *   gbrain integrity check              Read-only report to stdout
 *   gbrain integrity auto               Three-bucket repair with confidence
 *   gbrain integrity --dry-run          Same as auto, no writes
```

The third line is not a real invocation. `runIntegrity` dispatches on `sub` (`:210` `check`, `:214` `auto`) and `--dry-run` is parsed at `:414`, inside the `auto` path. Passing it as the subcommand exits 1:

```
$ gbrain integrity --dry-run
Unknown subcommand: --dry-run
Usage: gbrain integrity <subcommand> [options]
```

The runtime help text is correct — it lists `--dry-run` indented under `auto [options]`. Only the source comment is wrong, and the two disagree with each other.

## Why it is worth a line

Small, but it costs a round trip in exactly the situation where a user is being careful. I hit it while trying to preview an `integrity auto` before letting it write to a live brain: I read the comment, ran the documented form, and got an error instead of a preview. Someone less inclined to re-read the runtime help could reasonably conclude the dry-run path is broken rather than that the comment is stale.

It is also the kind of drift that only gets worse — the comment block is the first thing a reader sees when opening the file, so it seeds the wrong mental model for anyone modifying the command later.

## Two ways to close it, and I do not want to pick for you

**(a) Correct the comment** — one line, `gbrain integrity auto --dry-run`. Makes the comment match the implementation and the runtime help. This is the change I opened a PR for, since it is the minimal one and does not decide anything on your behalf.

**(b) Accept `integrity --dry-run` as an alias for `auto --dry-run`** — makes the implementation match what the comment already promises. Defensible if the comment reflects the intended surface, but it widens the CLI's accepted forms, which is your call and not something I would do unprompted.

The PR takes (a). If you would rather have (b) — or would rather drop the line entirely — say so and I will redo it; the comment fix is trivial to throw away.

## Note on testing

[...]

PR #3739, opened 2026-08-01
docs(integrity): correct the --dry-run line in the subcommand comment (#3738)

Masashi-Ono0611:
Addresses #3738.

The subcommand comment at the top of `src/commands/integrity.ts` listed a form the CLI rejects:

```
 *   gbrain integrity --dry-run          Same as auto, no writes
```

`runIntegrity` dispatches on `sub` (`:210` `check`, `:214` `auto`) and parses `--dry-run` at `:414`, inside the `auto` path. Running the documented form exits 1:

```
$ gbrain integrity --dry-run
Unknown subcommand: --dry-run
Usage: gbrain integrity <subcommand> [options]
```

The runtime help was already right — it shows `--dry-run` indented under `auto [options]`. Only the comment disagreed, and it is the first thing a reader sees when opening the file.

## The change

One line. `gbrain integrity --dry-run` → `gbrain integrity auto --dry-run`, keeping the column alignment of the surrounding block.

## No test, and why

This is a comment-only change, so there is no test that fails without it. I am flagging that explicitly rather than letting it read as an oversight — the repo asks contributors for a test that goes red without the fix, and I cannot produce one here.

The alternative direction *would* be testable: make `integrity --dry-run` an accepted alias for `auto --dry-run`, then spawn the CLI and assert exit code 0 plus an unchanged brain. I did not do that, because it widens the CLI's accepted surface and that is a design call that belongs to you rather than to a drive-by doc fix. #3738 lays out both options; say the word and I will redo this as the alias version, or drop the line entirely if that is what you would rather have.

## Verification

- Reproduced the error output above against the installed binary at v0.42.71.0, and re-read the comment from `git show upstream/master:src/commands/integrity.ts` to confirm it is still present on `f07e9dff`.

### DIFF
--- a/src/commands/integrity.ts
+++ b/src/commands/integrity.ts
@@ -12,7 +12,7 @@
  * Subcommands:
  *   gbrain integrity check              Read-only report to stdout
  *   gbrain integrity auto               Three-bucket repair with confidence
- *   gbrain integrity --dry-run          Same as auto, no writes
+ *   gbrain integrity auto --dry-run     Same as auto, no writes
  *
  * Three-bucket confidence (contract with x_handle_to_tweet resolver):
  *   >= 0.8 → auto-repair through BrainWriter transaction


Masashi-Ono0611:
My human part:
While my daily(nightly) GBrain status check, I just found a different btw comments and help codes about integrity function. I would love to fix and make it cleaner. 

Screenshot:
Screenshot 2026-08-02 at 0 46 09 PM

*same as in https://github.com/garrytan/gbrain/pull/3739#issue-5039002108

PR #3743, opened 2026-08-01
fix(context): support international home locations

### LINKED ISSUE #3740: bug: gbrain-context injects hard-coded SF home context for international users
## Summary

`gbrain-context` accepts an international `currentLocation.timezone`, but still treats San Francisco / US Pacific as the user's fixed home. For a user whose actual home and current timezone are both `Europe/Zurich`, the injected context contains the correct local time plus an incorrect `Home (SF)` / Pacific-time line on every turn.

This makes the plugin unsafe for non-SF users even when `memory/heartbeat-state.json` is populated exactly as documented.

## Reference

- GBrain: `f07e9dff116897273a90fb7ad22970caa96e2d1d` (`master` at reproduction time)
- OpenClaw: `2026.7.1-2`
- Bun: `1.3.12`
- macOS arm64

## Reproduction

Create the documented operational state:

```json
{
  "currentLocation": {
    "city": "Basel",
    "country": "Switzerland",
    "timezone": "Europe/Zurich",
    "source": "user-confirmed"
  }
}
```

Load `gbrain-context` and assemble a turn.

## Actual behavior

The generated block includes correct Zurich time/location, then adds an unrelated SF home clock:

```text
- Time: ... (Europe/Zurich)
- Location: Basel (source: user-confirmed)
- Home (SF): ... PT
```

The behavior comes from hard-coded user-specific defaults and state naming:

- `DEFAULT_TZ = 'US/Pacific'` and `DEFAULT_HOME = 'San Francisco'`
- `garryAwake` / `garryAwokeAt`
- every non-Pacific current timezone computes a Pacific `homeTime`
- formatter always labels that value `Home (SF)`

Permalinks:

- https://github.com/garrytan/gbrain/blob/f07e9dff116897273a90fb7ad22970caa96e2d1d/src/core/context-engine.ts#L240-L256
- https://github.com/garrytan/gbrain/blob/f07e9dff116897273a90fb7ad22970caa96e2d1d/src/core/context-engine.ts#L513-L528
- https://github.com/garrytan/gbrain/blob/f07e9dff116897273a90fb7ad22970caa96e2d1d/src/core/context-engine.ts#L593-L597

## Expected behavior

The context engine should not inject an unconfigured home city/timezone. International users should be able to configure current and home location/timezone, or omit the home line entirely when no home is configured.

## Suggested repository-level direction

- Make home city/timezone configurable with backwards-compatible SF defaults only where explicitly intended.
- Replace user-specific state keys (`garryAwake`) with generic keys while preserving migration compatibility.
- Emit `Home (...)` only when an explicit home differs from the current location.
- Add a non-SF test where both current and home timezone are `Europe/Zurich` and assert that no SF/PT context appears.

No local patch was applied; the plugin was left disabled to avoid injecting false context.

arisgysel-design:
## Summary

- replace the hard-coded San Francisco / US Pacific home assumptions with optional `homeLocation` state
- use generic `userAwake` / `userAwokeAt` keys while retaining reads of the legacy wake keys
- refuse to guess a city or timezone when neither current nor home location is configured
- emit a home clock only when an explicit home timezone differs from the current timezone
- document the international heartbeat-state shape and add Basel/Europe-Zurich regressions

Closes #3740.

## Verification

- `env -u DATABASE_URL -u GBRAIN_DATABASE_URL -u GBRAIN_DIRECT_DATABASE_URL bun test test/context-engine.test.ts test/e2e/openclaw-context-engine-plugin.test.ts` — 29 pass, 0 fail
- `env -u DATABASE_URL -u GBRAIN_DATABASE_URL -u GBRAIN_DIRECT_DATABASE_URL bun run typecheck` — pass
- `bun run verify` — 33/34 checks pass; the existing macOS `check:wasm` process is terminated with exit 137
- full parallel unit run was attempted; unrelated `notability-eval` cases hit their existing 60-second timeout, so the run was stopped after the relevant context-engine tests had passed

### DIFF
--- a/docs/guides/quiet-hours.md
+++ b/docs/guides/quiet-hours.md
@@ -67,12 +67,23 @@ the agent's operational state:
 
 ```json
 {
+  "userAwake": true,
   "currentLocation": {
-    "timezone": "US/Pacific",
-    "city": "San Francisco"
+    "timezone": "Europe/Zurich",
+    "city": "Basel",
+    "source": "user-confirmed"
+  },
+  "homeLocation": {
+    "timezone": "Europe/Zurich",
+    "city": "Basel"
   }
 }
 ```
+
+`homeLocation` is optional. The context engine shows a separate home clock only
+when an explicit home timezone differs from the current timezone. It never
+guesses a home city or timezone. Existing `garryAwake` state remains readable
+for migration compatibility; new producers should write `userAwake`.
 
 **Update the timezone when:**
 - Calendar shows the user flying somewhere (check for airline/hotel events)
--- a/llms-full.txt
+++ b/llms-full.txt
@@ -3169,12 +3169,23 @@ the agent's operational state:
 
 ```json
 {
+  "userAwake": true,
   "currentLocation": {
-    "timezone": "US/Pacific",
-    "city": "San Francisco"
+    "timezone": "Europe/Zurich",
+    "city": "Basel",
+    "source": "user-confirmed"
+  },
+  "homeLocation": {
+    "timezone": "Europe/Zurich",
+    "city": "Basel"
   }
 }
 ```
+
+`homeLocation` is optional. The context engine shows a separate home clock only
+when an explicit home timezone differs from the current timezone. It never
+guesses a home city or timezone. Existing `garryAwake` state remains readable
+for migration compatibility; new producers should write `userAwake`.
 
 **Update the timezone when:**
 - Calendar shows the user flying somewhere (check for airline/hotel events)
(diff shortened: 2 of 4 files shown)


time-attack:
Automated triage: this change adds new product surface (new `homeLocation`/`userAwake` heartbeat-state schema keys and a behavior change for the unconfigured-location default). Queued for maintainer product decision — no autonomous action will be taken.

time-attack:
Status update: this PR was reviewed in the 2026-08-18 full-backlog triage and confirmed valuable — it sits in the verified merge queue (position 52 of 53, risk: medium). CLEAN green but introduces new homeLocation/userAwake schema keys — maintainer product call before merge. No action needed from you right now — thank you for the contribution and the patience.

time-attack:

PR #3752, opened 2026-08-02
fix(autopilot): generated wrapper carries Homebrew bin paths

harshsahrawat-commits:
The launchd wrapper that `gbrain autopilot --install` generates execs gbrain (a `#!/usr/bin/env bun` script) with only `~/.bun/bin` prepended to PATH. On machines where bun is Homebrew-installed (`/opt/homebrew/bin/bun`), launchd's bare PATH can't find it and the service crash-loops with exit 127 / `env: bun: No such file or directory`.

One-line fix in the wrapper template: prepend `/opt/homebrew/bin:/usr/local/bin` alongside `~/.bun/bin`. Hit in production on a Homebrew-bun Mac 2026-08-01; the patched wrapper has run the daemon cleanly since.

### DIFF
--- a/src/commands/autopilot.ts
+++ b/src/commands/autopilot.ts
@@ -1329,7 +1329,7 @@ source ~/.zshrc 2>/dev/null || source ~/.bashrc 2>/dev/null || true
 # lockfile that blocks every subsequent tick. Prepending ~/.bun/bin here
 # keeps the wrapper self-contained regardless of which init file the OS
 # loaded.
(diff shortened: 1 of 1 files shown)


harshsahrawat-commits:
Closing as obsolete: f0a28eb2 (#3397) now derives the bun runtime dir for the generated wrapper from process.execPath, which covers Homebrew installs more generally than the hardcoded paths proposed here. The init fixes remain in #3753, which has been rebased over current master.

PR #3755, opened 2026-08-02
fix(export): scope tag + raw-data sidecar reads to the page's own source

alexey-metaengage:
## The bug

`gbrain export` reads two sidecars per page keyed by **slug alone**, but slugs are unique per `(source_id, slug)`, not brain-wide. On any multi-source brain, a slug present in more than one source exports the wrong metadata:

- `src/commands/export.ts:111` - `engine.getTags(page.slug)` with no opts. Both engines fall back to `source_id = 'default'` (`postgres-engine.ts:3786`, `pglite-engine.ts:3646`), so a page owned by another source is exported carrying the **default source's tags**, with its own tags dropped.
- `src/commands/export.ts:124` - `engine.getRawData(page.slug)` with no opts. This one applies **no source predicate at all** (the four-branch `else` at `postgres-engine.ts:4193`), so every source's raw data for that slug comes back and the loop below collapses the rows into one sidecar keyed by `rd.source` - a same-named raw source in another brain source silently overwrites this page's own.

Both engines already accept the scoping opts, and `Page.source_id` has been required and populated by `rowToPage` since v0.32.8. The export loop just never threaded it.

## The fix

Pass `{ sourceId: page.source_id }` at both call sites. No new flags, no behavior change on single-source brains (where `page.source_id` is `'default'` and the reads resolve identically).

## Tests

Two tests in `test/storage-export.test.ts`, fixtures on the only shape where the bug can appear: the same slug in two sources.

Discrimination verified - reverting `src/commands/export.ts` to master fails both, with the defect visible in the output:

```
Expected to contain: "tag-other"
Received: "---\ntype: note\ntitle: other title\ntags:\n  - tag-default\n---\n\nbody\n"
```

A page owned by source `other`, exported with the default source's tag and missing its own. The raw-data test fails on the extra `feed-default` key in `other`'s sidecar.

With the fix: 7 pass / 0 fail in that file, `tsc --noEmit` clean.

Note on the fixture: both pages share a slug, so they collide on one output path and only the last write survives. The test bumps the default page's `updated_at` to pin the `updated_desc` export order so the **non-default** page is the survivor, and asserts on its title first - otherwise the assertions would stop discriminating if that ordering ever changed, rather than failing.

## Unrelated observation

`test/storage-export.test.ts` times out its `beforeEach` under the default 5s budget on Windows - the `beforeAll` runs 120 PGLite migrations. Reproduces on untouched master; needs `--timeout 120000` locally. Not touched here.

### DIFF
--- a/src/commands/export.ts
+++ b/src/commands/export.ts
@@ -108,7 +108,11 @@ export async function runExport(engine: BrainEngine, args: string[]) {
   let exported = 0;
 
   for (const page of pages) {
-    const tags = await engine.getTags(page.slug);
+    // Slugs are unique per source, not brain-wide, so both sidecar reads are
+    // pinned to the page's own source. Unscoped, `getTags` falls back to
+    // `source_id = 'default'` and stamps the default source's tags onto a
+    // same-slug page from another source (dropping its real ones).
+    const tags = await engine.getTags(page.slug, { sourceId: page.source_id });
     const md = serializeMarkdown(
       page.frontmatter,
       page.compiled_truth,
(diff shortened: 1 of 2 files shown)


alexey-metaengage:
Split out of #3675 after review found this leak there. It reproduces on master with no flag involved, so it stands alone: on a multi-source brain, any page whose slug also exists in `default` exports the default source's tags instead of its own, and its raw-data sidecar can be overwritten by another source's rows.

#3675 is now rebased on top of this and no longer carries the fix.

time-attack:
Merged in the v0.45.10.0 community fix-wave 2 (#4085). Your fix landed on master as 0c485415a5, credited to you via Co-Authored-By, and is called out in the v0.45.10.0 CHANGELOG entry. Thank you for the contribution!

Issue #3736, opened 2026-08-01
sources harden --dry-run: installHelper chmods the helper above the dryRun guard (5th instance of the #3594 class)


Masashi-Ono0611:
## What happens

`installHelper()` runs `chmodSync(helperPath, 0o755)` above the `dryRun` guard, on the path taken when the helper is already up to date. So `gbrain sources harden <id> --dry-run` changes a file mode on disk.

`src/core/brain-repo-durability.ts:393-405` on current `master` (`f07e9dff`):

```ts
function installHelper(repoPath: string, dryRun: boolean): { status: StepStatus; detail: string } {
  const helperPath = join(repoPath, HELPER_REL);
  const script = renderCommitPushHelper();
  if (existsSync(helperPath) && readFileSync(helperPath, 'utf-8') === script) {
    // Ensure exec bit even when content is current.
    try { chmodSync(helperPath, 0o755); } catch { /* */ }
    return { status: 'ok', detail: `${HELPER_REL} already current` };
  }
  if (dryRun) return { status: 'fixed', detail: `would write ${HELPER_REL} (dry-run)` };
  mkdirSync(dirname(helperPath), { recursive: true });
  writeFileSync(helperPath, script); chmodSync(helperPath, 0o755);
  return { status: 'fixed', detail: `wrote ${HELPER_REL}` };
}
```

The `dryRun` check at the fifth line only gates the *write* branch. The early-return branch above it has already mutated the mode.

## Why I am filing it separately

This is a fifth instance of the class in #3594 — "the destructive call sits above the dry-run early-return, so the guard is real but arrives too late" — and #3594 explicitly asks for the class rather than the instances, on the grounds that fixing them one at a time will not stop the next one.

It is also a *second* instance inside the same command. #3692 is the `divergenceSafePull` at step 1 of `hardenBrainRepo`; this one is step 4, a different function, and it survives the step-1 fix. I found it while working on #3692 and confirmed it is not covered by that change.

## Severity, honestly stated

Low. The mode is set to `0o755`, which is what the file already has in the ordinary case, so most runs of this path change nothing observable. It is not data loss and I would not argue it blocks anything.

What makes it worth a line in the ledger is the contract rather than the blast radius: `--dry-run` is the flag people reach for when they are not sure, and "no persistent state changes" is the whole promise. A dry-run that touches file modes on a repository the user has not decided to modify yet is a smaller version of the same surprise as #3543.

[...]

PR #3759, opened 2026-08-02
fix(sources): stop harden --dry-run from chmodding an already-current helper (#3736)

Masashi-Ono0611:
Addresses #3736.

## What this changes

`installHelper()` in `src/core/brain-repo-durability.ts` ran `chmodSync(helperPath, 0o755)` on the "already current" early-return path — three lines *above* the `dryRun` check — so `gbrain sources harden <id> --dry-run` mutated the helper's file mode even though the guard exists. This is the same class as #3594 (destructive call sits above the dry-run early-return) and a second, independent instance inside `hardenBrainRepo` alongside #3692 (which gated the step-1 `pull`, not this step-4 chmod).

```diff
   if (existsSync(helperPath) && readFileSync(helperPath, 'utf-8') === script) {
-    // Ensure exec bit even when content is current.
-    try { chmodSync(helperPath, 0o755); } catch { /* */ }
+    // Ensure exec bit even when content is current — but not in dry-run: a
+    // preview must not mutate permissions (#3736).
+    if (!dryRun) { try { chmodSync(helperPath, 0o755); } catch { /* */ } }
     return { status: 'ok', detail: `${HELPER_REL} already current` };
   }
```

The non-dry-run write path later in the same function is untouched — real hardening still writes and chmods exactly as before.

## Discriminating power (test fails without the fix)

`test/brain-repo-durability.serial.test.ts` adds two tests under `describe('hardenBrainRepo', ...)`, following the file's existing style (real git fixtures, no mocks):

- **`dry-run does not chmod an already-current helper (#3736)`** — hardens for real, manually drifts the helper to `0o644`, re-hardens with `dryRun: true`, asserts the mode is still `0o644` (untouched). This is the load-bearing assertion — it fails on unpatched `master` because the early-return path chmods back to `0o755` regardless of `dryRun`.
- **`non-dry-run restores the exec bit on an already-current helper`** — control: same drift, then a real (non-dry-run) re-harden, asserts the exec bit comes back. Confirms the fix didn't silently break real hardening.

## Verification

- `bun test test/brain-repo-durability.serial.test.ts` — 21 pass, 0 fail, 49 `expect()` calls
- `bun run typecheck` — clean
- Diff: 2 files changed, +19/-2 (1 commit)

### DIFF
--- a/src/core/brain-repo-durability.ts
+++ b/src/core/brain-repo-durability.ts
@@ -394,8 +394,9 @@ function installHelper(repoPath: string, dryRun: boolean): { status: StepStatus;
   const helperPath = join(repoPath, HELPER_REL);
   const script = renderCommitPushHelper();
   if (existsSync(helperPath) && readFileSync(helperPath, 'utf-8') === script) {
-    // Ensure exec bit even when content is current.
-    try { chmodSync(helperPath, 0o755); } catch { /* */ }
+    // Ensure exec bit even when content is current — but not in dry-run: a
+    // preview must not mutate permissions (#3736).
+    if (!dryRun) { try { chmodSync(helperPath, 0o755); } catch { /* */ } }
     return { status: 'ok', detail: `${HELPER_REL} already current` };
   }
   if (dryRun) return { status: 'fixed', detail: `would write ${HELPER_REL} (dry-run)` };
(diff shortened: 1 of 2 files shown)


Masashi-Ono0611:
My human part:
When I switched 755 to 644 mode and called --dry-run, GBrain silently backed 755 mode and told already current. It breaks the concept of --dry-run.
so, i fixed the dry-run chmod bug in sources harden — filed as #3736.

Screenshot:
Image

PR #3531, opened 2026-07-28
fix(ai): fold dashscope + google keys into gateway env, drop the retired Gemini default (#3500, #3510)

time-attack:
Fixes #3500 and #3510 — two provider-config correctness bugs, one change since Part B's guard mechanism depends on Part A's recipe hygiene story staying coherent.

## Part A — #3500: config-plane provider keys never reached the gateway

`buildGatewayConfig` folded openai/anthropic/zeroentropy/openrouter/voyage (+ azure endpoint fields) into the gateway env, but **dashscope was missing** while two live recipes (`dashscope`, `dashscope-rerank`) require `DASHSCOPE_API_KEY`. The fix is two-part as the field didn't exist either:

- `GBrainConfig` gains `dashscope_api_key` and `google_api_key` file-plane fields (`src/core/config.ts`), both registered in `KNOWN_CONFIG_KEYS` so `gbrain config set` accepts them without `--force`. Redaction is pattern-based (`*_api_key`) so both auto-redact.
- `buildGatewayConfig` folds them into the gateway env: `DASHSCOPE_API_KEY` and `GOOGLE_GENERATIVE_AI_API_KEY` (`src/core/ai/build-gateway-config.ts`).
- **Sibling gap closed:** no Google/Gemini key was ever foldable — `gateway.ts` reads `cfg.env.GOOGLE_GENERATIVE_AI_API_KEY` at three sites and the only configuration route was exporting that exact env name. This honors the explicit deferral note in `src/core/brain-score-recommendations.ts` ("when a future change threads `google_api_key` into buildGatewayConfig, re-add the matching entry here"): `HOSTED_EMBED_KEY_CONFIG` gains `GOOGLE_GENERATIVE_AI_API_KEY → google_api_key` and `DASHSCOPE_API_KEY → dashscope_api_key` in the same change, so doctor/autopilot count a config-plane key as configured only now that the gateway can actually use it.
- **`GEMINI_API_KEY` accepted as an alias** for `GOOGLE_GENERATIVE_AI_API_KEY` (the name Google's own docs/SDKs use). Precedence: env `GOOGLE_GENERATIVE_AI_API_KEY` > env `GEMINI_API_KEY` > config `google_api_key`.

**Deliberately unchanged:** process.env spreading after `envFromConfig` (stale OS env wins over config) is the documented, intentional precedence for every provider — not touched.

[...]

Issue #3510, opened 2026-07-28
Retired google:gemini-1.5-pro is still the hardcoded default at 4 sites; the recipe allowlist masks it from the guard test


time-attack:
`google:gemini-1.5-pro` has been retired by Google, but it is still the hardcoded default at four call sites on master. Surfaced while reviewing #3501, which fixes only the fifth (`cross-modal-eval/runner.ts`).

**Still pointing at the retired model:**

| File | Line | What breaks |
|---|---|---|
| `src/core/cycle/grade-takes.ts` | 222 | take grading falls back to a model that no longer answers |
| `src/core/takes-quality-eval/runner.ts` | 39 | takes-quality eval default slot |
| `src/core/takes-quality-eval/pricing.ts` | 44 | derived pricing view carries a dead entry |
| `src/core/ai/recipes/google.ts` | 26 | the recipe still advertises it as a valid chat target |

**Why the existing guard doesn't catch it.** `test/cross-modal-default-slots.test.ts` validates default slots against the recipe's advertised model list. Because `recipes/google.ts:26` still lists the retired model, the guard passes. Removing it from the recipe list makes every dead default fail at once — which is the check we actually want, and is the reason to fix `google.ts` first rather than last.

**Minimal fix:** one sweep across all five sites (the four above plus the one #3501 handles), driven by dropping the model from the recipe allowlist so the test surfaces the rest. Per the root-cause rule in CLAUDE.md this is a smaller diff than five separate patches and leaves no sibling caller broken.

Note this is the same class as the `deepseek:deepseek-v4-pro` substitution proposed in #3501 — that model is present in both the recipe chat touchpoint (`recipes/deepseek.ts:95`) and canonical pricing (`model-pricing.ts:101`), so it is a valid replacement target for the slots that need a third distinct provider. Slots that just need *a* working model can use whatever the tier default already resolves to.

time-attack:
## Fixed by #3531 — verified, not assumed

Merged as `335e4703`. I checked every remaining reference on master rather than trusting the PR description, because a grep that still finds hits is exactly how a half-done sweep gets closed:

```
src/core/ai/recipes/google.ts:26          // comment explaining the retirement
src/core/cross-modal-eval/runner.ts:54    // comment explaining the retirement
src/core/model-pricing.ts:87              // comment explaining the retirement
src/core/model-pricing.ts:90              'google:gemini-1.5-pro': { input: 1.25, output: 5.00 }
[...]

PR #3760, opened 2026-08-02
fix(recipes): drop the retired gemini-2.0 family from the Google recipe

alexey-metaengage:
## The bug

Every chat and expansion model the Google recipe declares is dead. Verified against the live API on 2026-08-02 with a real `generateContent` call per id:

| model | recipe slot | result |
|---|---|---|
| `gemini-2.0-flash-exp` | chat | **HTTP 404** |
| `gemini-2.0-flash` | chat, expansion | **HTTP 404** |
| `gemini-2.0-flash-lite` | expansion | **HTTP 404** |
| `gemini-2.5-flash` | - | live |
| `gemini-2.5-flash-lite` | - | live |

`gemini-embedding-001` is unaffected and still live.

This is the same failure #3510 / #3531 fixed for `gemini-1.5-pro`, one generation later: `gemini-2.0-flash` was put in as the replacement and has since been retired itself. A brain configured on Google gets a 404 on every `think`, and `DEFAULT_MODEL_PANEL` in the takes-quality eval names a model that cannot answer.

Note `ListModels` still advertises the whole `gemini-2.0` family, so the endpoint list is not a reliable signal here - only a real call is.

## The fix

Follows #3531's shape exactly.

- **Recipe** (`src/core/ai/recipes/google.ts`): chat and expansion both move to `['gemini-2.5-flash', 'gemini-2.5-flash-lite']`. Chat output price corrected 1.20 -> 2.50 and expansion 0.10 -> 0.30 to match 2.5-flash. The existing "deliberately NOT listed" comment is extended to name the 2.0 family, keeping the rule that a dead model never goes back in the list to quiet a failing default.
- **Pricing** (`src/core/model-pricing.ts`): the retired 2.0 ids **stay**, matching how 1.5-pro and `deepseek-chat` are handled, so historical usage and audit rows still price. The live 2.5 ids are added.
- **Hardcoded defaults**: `DEFAULT_MODEL_PANEL` (`takes-quality-eval/runner.ts`), the takes-quality allowlist (`takes-quality-eval/pricing.ts`), and the ensemble-judge doc comment (`cycle/grade-takes.ts`) move to `gemini-2.5-flash`.
- **Docs**: `docs/eval-takes-quality.md` panel examples.

## Tests

`test/default-model-panels.test.ts` already guards defaults against the recipe lists, but its own header notes it "only works if recipes list LIVE models" - and nothing was guarding *that*. Added the missing half: a test asserting the Google recipe offers none of the four known-retired ids, plus one asserting every chat model it does offer is priced.

Discrimination verified - reverting `recipes/google.ts` and `takes-quality-eval/runner.ts` to master fails three tests:

```
(fail) google recipe lists no retired model > neither the chat nor the expansion touchpoint offers a dead model
(fail) google recipe lists no retired model > every model the google recipe offers is priced
(fail) chat touchpoint - model resolver + aliases > assertTouchpoint accepts chat for chat-capable native + openai-compat providers
```

With the change: 79 pass / 0 fail across the five affected files plus the takes-quality runner suite. `tsc --noEmit` clean.

## Known gap, recorded rather than hidden

Model **liveness** is live-verified per the table above. Model **prices** are not - `gemini-2.5-flash` at $0.30/$2.50 and `gemini-2.5-flash-lite` at $0.10/$0.40 come from Google's published pricing, not an API call, since there's no endpoint to check them against. Same class of gap #3531 disclosed, in the opposite direction: that PR verified prices but not retirement.

`gemini-2.5-pro` and `gemini-3.5-flash` also answered live and are deliberately **not** added - substituting dead models one-for-one is the fix here, and adding capability is a separate call for you to make.

### DIFF
--- a/docs/eval-takes-quality.md
+++ b/docs/eval-takes-quality.md
@@ -31,7 +31,7 @@ receipt file from disk and re-renders it. The other modes need the brain.
 | `--budget-usd N` | unset | Abort before next call's projected cost would exceed cap. Models without a `pricing.ts` entry fail loud (codex #4). |
 | `--source db|fs` | `db` | `fs` is reserved for v0.33+. |
 | `--slug-prefix P` | unset | Filter takes to pages whose slug starts with P. |
-| `--models a,b,c` | `openai:gpt-5.2,anthropic:claude-opus-4-7,google:gemini-2.0-flash` | Comma-separated panel. |
+| `--models a,b,c` | `openai:gpt-5.2,anthropic:claude-opus-4-7,google:gemini-2.5-flash` | Comma-separated panel. |
 | `--json` | off | Emit the full receipt to stdout. |
 
 ## Receipt JSON shape (`schema_version: 1`)
@@ -50,7 +50,7 @@ receipt file from disk and re-renders it. The other modes need the brain.
   },
   "prompt_sha8": "abcd1234",
   "models_sha8": "abcd1234",
-  "models": ["openai:gpt-5.2", "anthropic:claude-opus-4-7", "google:gemini-2.0-flash"],
+  "models": ["openai:gpt-5.2", "anthropic:claude-opus-4-7", "google:gemini-2.5-flash"],
   "cycles_run": 3,
   "successes_per_cycle": [3, 3, 2],
   "verdict": "pass",
--- a/src/core/ai/recipes/google.ts
+++ b/src/core/ai/recipes/google.ts
@@ -29,22 +29,24 @@ export const google: Recipe = {
       safety_factor: 0.8,
     },
     expansion: {
-      models: ['gemini-2.0-flash', 'gemini-2.0-flash-lite'],
-      cost_per_1m_tokens_usd: 0.10,
-      price_last_verified: '2026-04-20',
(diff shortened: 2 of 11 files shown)


time-attack:
Status update: reviewed in the 2026-08-18 full-backlog triage — still wanted, but not yet mergeable: Master's Google recipe still lists gemini-2.0 models the PR shows 404ing (same class as #3510/#3531); valuable but CONFLICTING with churned pricing/eval files. If you update it, it will be re-reviewed promptly; alternatively the fix may be adopted with credit in an upcoming fix wave. Thanks for the contribution and the patience.

time-attack:
Re-trialed against the current integration branch (post v0.46.23-25 waves): Conflict: src/core/ai/recipes/google.ts, 1 hunk — wave #4298 per-model googleSupportsPromptCache vs PR supports_prompt_cache:false. Not superseded (wave still lists gemini-2.0 defaults). Trivial rebase; keep wave fn. No red flags. — a rebase along those lines puts this back in the active merge queue; re-trials run nightly, so a fresh push gets picked up automatically. Thank you!

garrytan:

PR #3761, opened 2026-08-02
fix(think): name the resolved provider's credential in the no-LLM message

alexey-metaengage:
## The bug

When `think` degrades gracefully because a provider credential is missing, it tells the user to set an **Anthropic** key regardless of which provider the model actually resolves to:

```ts
// src/core/think/index.ts:836
content: [{ type: 'text', text: '(no LLM available - set anthropic_api_key via gbrain config or ANTHROPIC_API_KEY env)' }],
```

That string is surfaced verbatim as the user's answer. On a brain configured with `models.default = google:gemini-2.5-flash` and no Google key, the answer to every question becomes "set ANTHROPIC_API_KEY" - advice that cannot fix the run and sends operators to check a credential that isn't involved. The same hardcoding sits on the no-client path at `:491` ("set ANTHROPIC_API_KEY or pass `client`").

The information needed was already in hand: both sites have the resolved model string, and every recipe already declares `auth_env.required` plus a `setup_url`.

## The fix

One helper, `missingCredentialHint(modelStr)`, resolves the model to its recipe and names that provider's env var and key page. Both sites call it. A new provider needs no change here - the advice comes from its recipe.

Unresolvable ids fall back to the previous Anthropic wording, since Anthropic is still the default provider for a bare model id and there's nothing better to say. The sentinel path exists to degrade gracefully, so it must not throw.

**Deliberately unchanged:** the `NO_ANTHROPIC_API_KEY` warning code. It's a machine-readable contract that consumers key on, and its name being provider-specific is a separate call. Only the human-facing prose moves.

## Tests

Four tests in `test/think-gateway-adapter.test.ts`. Discrimination verified - stubbing the helper back to a constant Anthropic string fails two:

```
(fail) the sentinel names the resolved provider credential, not Anthropic
(fail) each provider gets its own env var and key page
```

The other two are fallback guards (an Anthropic model still gets Anthropic wording; an unknown provider falls back rather than throwing) and pass either way by design - they exist to catch a future change that breaks the default path, not to prove this one.

20 pass / 0 fail in that file, 31 pass / 0 fail in `think-pipeline.serial.test.ts`, `tsc --noEmit` clean.

## Context

Found while running a brain on Google. Companion to #3760, which retires the dead `gemini-2.0` models that produced the failure in the first place - a 404 from Google surfaced to the user as "set your Anthropic key". The two are independent and can land in either order.

### DIFF
--- a/src/core/think/index.ts
+++ b/src/core/think/index.ts
@@ -26,6 +26,7 @@ import { resolveCitations, type ParsedCitation } from './cite-render.ts';
 import { resolveOwnerHolder } from '../owner-holder.ts';
 import { resolveModel } from '../model-config.ts';
 import { chat as gatewayChat, probeChatModel, type ChatResult } from '../ai/gateway.ts';
+import { resolveRecipe } from '../ai/model-resolver.ts';
 import { AIConfigError } from '../ai/errors.ts';
 import { normalizeModelId } from '../model-id.ts';
 import { hasAnthropicKey } from '../ai/anthropic-key.ts';
@@ -488,7 +489,7 @@ export async function runThink(
         question: opts.question,
         answer: modelProblem
           ? `(model "${modelUsed}" not usable — ${detail}${fix})`
-          : '(no LLM available — set ANTHROPIC_API_KEY or pass `client`)',
+          : `(no LLM available — ${missingCredentialHint(modelUsed)}, or pass \`client\`)`,
         citations: [],
         gaps: [
           modelProblem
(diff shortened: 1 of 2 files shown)


time-attack:
Thanks — this was a real bug and a well-built fix: the failing-first tests, the recipe-driven hint, and the deliberate scoping around the warning code were all right. We verified it end-to-end against three adopter profiles and it passed cleanly. It was queued for the next fix-wave, but #4014 (v0.44.1.0) landed first and absorbed this territory during its review: think's graceful sentinel now surfaces the thrown AIConfigError directly, and the gateway's missing-key errors already name the resolved provider's env var and setup page — so a Google- or OpenAI-configured brain with a missing key now gets the correct credential advice, plus provider 4xxs (model_not_found) are no longer mislabeled as key problems, which a recipe-side re-derivation couldn't distinguish. The branch now conflicts with master on both hunks. Closing as superseded — if you spot a remaining path where the wrong credential is named on current master, please reopen with a repro and we'll take it in the next wave.

Issue #2163, opened 2026-06-13
synthesize_concepts writes concept pages that are never chunked/embedded → unreachable by retrieval despite source-boost 1.3×


thomaskong119:
## Summary

`synthesize_concepts` writes `concepts/` pages via `putPage`, but those pages end up with **0 `content_chunks` and a NULL embedding** — so they are invisible to `query` (vector retrieval) and effectively unreachable. This is despite `src/core/search/source-boost.ts` assigning `'concepts/': 1.3`, i.e. the retrieval layer is explicitly designed to rank concept pages *above* bulk content. In practice that boost is dead — there are no concept chunks to boost.

The CHANGELOG's own worked example shows the intended behavior (`concepts/widget-pattern` → "returned (rank 1)"), but a deployment whose concept pages are never embedded can never reproduce it.

## Environment

Observed on a v0.42.x deployment. Verified the gap still exists on `main` at **v0.42.42**: `src/core/cycle/synthesize-concepts.ts` still only calls `engine.putPage('concepts/${title}', …)` with no chunk/embed step, and the header comment still reads `// … dedup-by-embedding-similarity ship in v0.42+.`

## Evidence (deployment with ~1,700 concept pages)

- `SELECT count(*) FROM content_chunks c JOIN pages p ON p.id = c.page_id WHERE p.slug LIKE 'concepts/%'` → **0**
- Every `concepts/` page has `embedding_signature IS NULL`.
- A 20-question retrieval probe (questions chosen to match concept topics) returned a concept page in top-5 in **0/20** cases.
- FTS `search` also fails to surface them even on near-verbatim title queries: the pages *do* have `search_vector` populated, but being short they are out-ranked by longer pages and fall below the result limit.
- **Control:** `distilled/` and `dream-cycle-summaries/` pages (also dream-generated) *do* have chunks. So the `dream_generated` anti-loop guard is not the cause — it is specific to the `synthesize_concepts` (and `extract_atoms`) write path.

## Expected vs actual

- **Expected:** concept pages — the curated "distilled framework" output the schema treats as first-class and `source-boost.ts` prioritizes at 1.3× — should be in the retrieval surface (chunked + embedded) so `query`/`search` can return them.
- **Actual:** `synthesize_concepts` writes them but they are never chunked/embedded, so they are unreachable via the primary retrieval paths.

## Possible directions (maintainers to weigh)

[...]

PR #3762, opened 2026-08-02
fix(cycle): extract_atoms writes atoms outside the chunk/embed pipeline, making every atom invisible to search

awilhite:
extract_atoms persists each atom with engine.putPage, which is a bare
INSERT INTO pages ... ON CONFLICT. It never chunks and never embeds, so
atom pages never reach content_chunks and cannot be returned by search.

This is the same defect #2163 fixed for concept pages in
synthesize-concepts.ts, which serializes to markdown and imports through
the canonical pipeline. The fix was never applied one layer down.

On my brain this left 3,942 atom pages with 0 rows in content_chunks.
Atoms carry source_quote, lesson and source_slug, so the richest and
most grounded layer was the one layer retrieval could not see. It fails
silently: rows exist, counts look right, the cycle reports ok, and
searches return the concept summary of the atom instead, which looks
like a reasonable answer.

Route atoms through serializeMarkdown -> importFromContent, keeping
sourceId threaded (v0.41.2.1 D9 #1). type: 'atom' rides in frontmatter,
which parseMarkdown honours as an explicit override ahead of path
inference, so the page type survives the round-trip.

Verified on a live brain: 72 newly extracted atoms produced 72 chunks
and 72 embeddings (previously 0), all 4,014 atom pages still have
type='atom', source_quote/lesson/source_slug are preserved, and atoms
are now returned by gbrain search.

Tests cover the round-trip invariants this introduces, since atom text
is model-authored and now passes through YAML: page type survival
(including against a filePath that would infer otherwise), grounding
frontmatter, titles containing colons, quotes containing quotes and
hashes, bodies containing --- rules, and numeric-looking titles. Two
further tests pin that optional fields must use a conditional spread,
because js-yaml throws on undefined where the old row-write tolerated
it — atom_type is safe to assign unconditionally only because
parseAtomsResponse drops atoms missing it.

---

Same shape as the fix in #2163, applied where it was missed. Sibling file `synthesize-concepts.ts` already routes concept pages through `importFromContent` for exactly this reason; `extract-atoms.ts` still used the bare row writer.

`bun run typecheck` clean. 156 tests pass across the extract-atoms, cycle-synthesize and doctor-backlog suites, including 10 new ones.

Note for existing installs: this fixes writes going forward. Atoms already on disk stay unchunked until re-imported (`forceRechunk`), so a backfill is needed to make historical atoms searchable.

### DIFF
--- a/src/core/cycle/extract-atoms.ts
+++ b/src/core/cycle/extract-atoms.ts
@@ -6,14 +6,16 @@
 //      pages already extracted by content hash — see "Idempotency" below).
 //   2. Dedup by content_hash; transcripts win on collision.
 //   3. Per work-item, ask Haiku for 1-3 atoms.
-//   4. Write each atom via engine.putPage(slug, page, {sourceId})
-//      with sourceId threaded so federated brains route correctly.
+//   4. Write each atom via importFromContent(slug, markdown, {sourceId})
+//      with sourceId threaded so federated brains route correctly. The
+//      canonical import path (not engine.putPage) is what chunks and embeds
+//      the page — see the write site below and #2163.
 //
 // Idempotency (per-atom, via deterministic slug):
 //   Each atom's slug is `atoms/<source-date>/<stem>-<title-hash>` — built from
 //   the SOURCE date (the transcript's own date / the page slug), NOT the run
 //   date, plus a 6-char hash of the title. Re-extracting the same atom resolves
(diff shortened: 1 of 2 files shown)


time-attack:
Merged in the v0.45.10.0 community fix-wave 2 (#4085). Your fix landed on master as ca260baaaa, credited to you via Co-Authored-By, and is called out in the v0.45.10.0 CHANGELOG entry. Thank you for the contribution!

PR #3767, opened 2026-08-02
fix(import): scope pre-write existence check and read-back to the write namespace

garrytan-agents:
## Bug

`importFromContent` / `importFile` / `importImageFile` default their writes (`putPage`/`createVersion`) to source `default`, but the pre-write existence check ran **unscoped** (`sourceId ? { sourceId } : undefined`), which matches a same-slug row from **any** source (`LIMIT 1`).

On a multi-source brain where another source owns the slug and no default-source row exists:

1. `existing` comes back truthy (foreign-source row),
2. `tx.createVersion(slug)` scoped to `default` finds 0 rows,
3. the import throws `createVersion failed: page "<slug>" (source=default) not found` and the whole operation aborts.

In practice this crashed `synthesize_concepts` in **every** dream cycle (concept pages are written to the default namespace) as soon as one synthesized concept slug collided with a row owned by another source — `last_full_cycle_at` never stamped, so `cycle_freshness` degraded permanently on all sources.

The post-write read-back guard (`verifyPageReadable`) had the same unscoped `getPage`: it can match another source's row and fail with a false "stale content_hash" desync error.

## Fix

Existence check (markdown + code + image paths) and the read-back guard now use `{ sourceId: sourceId ?? 'default' }` so the check and the write always agree on the namespace. Explicit-`sourceId` callers are unchanged.

## Tests

`test/import-existence-source-scope.test.ts` — source-aware mock engine keyed on `(source_id, slug)` whose `createVersion` mirrors the real engine's failure mode:

- crash repro: foreign-source slug + default-namespace write now imports as a NEW page (no snapshot),
- existence check is scoped to `default` when no `sourceId` is given,
- existing default-source page still gets its version snapshot,
- explicit-`sourceId` path unchanged (scoped check + scoped snapshot).

`bun test` on the import suite: 74 pass / 0 fail (import-file, content-sanity, image, dedup, new file).

Verified live on a 4-source deployment: previously every `dream --source <s>` died at `synthesize_concepts` with the error above; with this patch cycles complete and stamp `last_full_cycle_at`.

### DIFF
--- a/src/core/import-file.ts
+++ b/src/core/import-file.ts
@@ -569,7 +569,14 @@ export async function importFromContent(
   // #1035: fetch the existing page BEFORE the hash compute so (a) the type
   // preservation below participates in the hash (a no-op re-put stays a
   // hash-match skip) and (b) the hash short-circuit below reuses this row.
-  const existing = await engine.getPage(slug, sourceId ? { sourceId } : undefined);
+  // Existence check must target the same namespace the writes below target.
+  // putPage/createVersion default sourceId to 'default', but an unscoped
+  // getPage matches a same-slug row from ANY source (LIMIT 1). When another
+  // source owns the slug and no default-source row exists, `existing` is
+  // truthy while tx.createVersion(slug) scoped to default finds 0 rows and
+  // throws, aborting the whole import (crashes synthesize_concepts in every
+  // dream cycle on multi-source brains).
+  const existing = await engine.getPage(slug, { sourceId: sourceId ?? 'default' });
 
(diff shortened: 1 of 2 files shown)


time-attack:
Superseded by v0.46.11.0 (#4219, commit 5ef85ac9): master scopes the existence check (import-file.ts:606), read-back (:1061), code path (:1348) and image path (:1848) to the write's source, adds CI guard scripts/check-getpage-scoped-write.mjs + engine tiebreak. Thanks — the fix class is fully landed.

PR #3777, opened 2026-08-03
AGE-87: expose vector fallback and dedup public results

splemoine8:
## Summary

- add explicit vector fallback metadata for unavailable providers, embedding errors/timeouts, empty vector results, and successful vector retrieval
- treat successful-but-empty vector retrieval as keyword fallback
- deduplicate public results to one result per `(source_id, slug)` on all three public return paths while preserving internal chunk recall
- add focused regression coverage

Authority/approval ranking changes are intentionally out of scope pending the local 12-query evaluation.

## Verification

- `npx tsc --noEmit`
- `npx bun test test/dedup.test.ts test/hybrid-meta.serial.test.ts` (25 passed, 0 failed)
- `git diff --check`

AGE-87

### DIFF
--- a/src/core/search/hybrid.ts
+++ b/src/core/search/hybrid.ts
@@ -886,6 +886,11 @@ export async function embedQueryBounded(
   }
 }
 
+function vectorFailureReason(err: unknown): HybridSearchMeta['vector_fallback_reason'] {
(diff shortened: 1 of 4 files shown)


splemoine8:
Closing at the request of the project owner; this change was not requested for the official repository.

PR #3780, opened 2026-08-03
fix(cli): route `takes --help` to its own usage block instead of the generic stub

ghizi:
Repro on a fresh checkout:

```
$ gbrain takes --help
Usage: gbrain takes

gbrain takes - run gbrain --help for the full command list.
```

`takes` is in `CLI_ONLY` but missing from `CLI_ONLY_SELF_HELP`, so the generic short-circuit in `main()` fires before `runTakes` can print its own help. The detailed usage block in `src/commands/takes.ts` is unreachable from the CLI, which makes `add` / `update` / `supersede` / `resolve` / `scorecard` / `calibration` / `search` undiscoverable without reading the source.

This is the same holdout pattern already fixed for `capture` (v0.39.3.0 WARN-5), `sync` (v0.37 CDX2-12), and `schema` (v0.40.6.0) — and it needs the same two parts those did:

1. Add `takes` to `CLI_ONLY_SELF_HELP` so the generic stub stays out of the way.
2. Add a pre-engine-bind `--help` branch in `handleCliOnly`. The dispatch case still binds an engine, so part 1 alone turns the stub into `No brain configured. Run: gbrain init` on a fresh tmpdir. `runTakes` returns from its help path before touching the engine, so it takes `null` exactly like `runEnrich` / `runSync`.

## Tests

Three tests added to `test/cli-help-discoverability.test.ts`, next to the existing `capture` / `sync` cases and using the same `runCli` subprocess harness (`GBRAIN_HOME` pointed at a nonexistent dir, so the no-brain case is covered).

All three **fail on the parent commit** and pass here — verified by stashing the `src/cli.ts` change and re-running (10 pass / 3 fail → 13 pass / 0 fail).

- `bun test test/cli.test.ts test/cli-args.test.ts test/cli-dispatch-thin-client.test.ts test/takes-command-source-scope.test.ts test/takes-mcp-allowlist.serial.test.ts` — 54 pass / 0 fail

No behavior change outside the `--help` path.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -121,6 +121,12 @@ const CLI_ONLY_SELF_HELP = new Set([
   // --help` print the migration flags from runMigrateEmbeddings. `migrate`
   // (engine transfer) keeps its own dispatch too.
   'migrate', 'retrieval-upgrade',
+  // `gbrain takes --help` printed the generic one-line stub, so the nine
+  // subcommands (add/update/supersede/resolve/scorecard/calibration/revisit/
+  // extract/search) were undiscoverable from the CLI — the detailed usage
+  // block in runTakes (src/commands/takes.ts) was unreachable. Same holdout
+  // pattern as `capture`, `sync`, and `schema` above.
+  'takes',
 ]);
 
 // v114 (#1941): alias -> operation lookup, kept separate from `cliOps` so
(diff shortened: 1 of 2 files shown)


time-attack:
Status update: reviewed in the 2026-08-18 full-backlog triage — still wanted, but not yet mergeable: Bug still present on master ('takes' in CLI_ONLY but not CLI_ONLY_SELF_HELP); matches the capture/sync/schema precedent, but the branch conflicts with churned cli.ts. If you update it, it will be re-reviewed promptly; alternatively the fix may be adopted with credit in an upcoming fix wave. Thanks for the contribution and the patience.

time-attack:
Re-trialed against the current integration branch (post v0.46.23-25 waves): Conflicts: src/cli.ts (CLI_ONLY_SELF_HELP tail grew: jobs/dream/sources etc) + test/cli-help-discoverability.test.ts (EOF append clash). Not superseded — 'takes' still absent from SELF_HELP on wave. Trivial union rebase. — a rebase along those lines puts this back in the active merge queue; re-trials run nightly, so a fresh push gets picked up automatically. Thank you!

garrytan:

PR #3781, opened 2026-08-03
fix(takes): resolve the page before writing markdown in `takes add`

ghizi:
`cmdAdd` writes the `.md` file first and resolves the page second:

```ts
writeBody(path, nextBody);

// Mirror to DB. Page may not be in DB yet if not synced — caller must run sync first.
const pageId = await getPageId(engine, slug, sourceId);
```

`getPageId` exits 1 when the slug is not in the brain. So adding a take against an unsynced or misspelled slug leaves a markdown file on disk carrying a take with no DB row. The command prints `Page not found in brain: <slug>` and exits 1, so the user reasonably believes nothing was recorded — but the fence is real, and it is:

- **invisible to `scorecard` / `calibration`**, which read the DB, yet
- **still counted for row numbering**, so a later successful `takes add` on that page starts past a row the DB never saw.

It is easy to hit on a Postgres brain whose markdown tree is partial or absent, because the read paths (`takes <slug>`, `scorecard`) query the DB directly and work fine — the mismatch only surfaces on the write path, and it surfaces as silent data loss rather than an error.

## Why cmdAdd specifically

`cmdUpdate`, `cmdSupersede`, and `cmdResolve` all already resolve the page before touching markdown. `cmdAdd` was the only outlier. The fix hoists the `getPageId` call to the top of the `withPageLock` callback so all four agree.

No behavior change on the success path: same file contents, same `addTakesBatch` payload, same row numbering. The only difference is that the failure path now leaves no on-disk residue.

## Tests

Two tests added to `test/takes-command-source-scope.test.ts`, reusing that file's existing mock-engine + tmpdir harness:

- **missing page leaves no orphaned `.md`** — asserts exit 1, the expected stderr, no file on disk, and no `addTakesBatch` call. **Fails on the parent commit** (the file exists there), passes here.
- **existing page still writes both markdown and DB** — guards the happy path so the reorder cannot silently break it. Passes on both.

`process.exit` / `console.error` are stubbed with `spyOn(...).mockRestore()` in a `finally`, matching the pattern already used in `test/config-get-plane.test.ts`.

- all 15 `takes-*` / `extract-takes-*` suites — 182 pass / 0 fail

### DIFF
--- a/src/commands/takes.ts
+++ b/src/commands/takes.ts
@@ -210,6 +210,13 @@ async function cmdAdd(engine: BrainEngine, args: string[], sourceId?: string): P
   const brainDir = await resolveBrainDir(engine, dirArg ?? null);
 
   await withPageLock(slug, async () => {
+    // Resolve the page BEFORE touching the markdown. getPageId exits 1 when the
+    // page isn't in the brain; doing this after writeBody left a .md file
+    // carrying a take with no DB row — invisible to scorecard/calibration but
+    // present on disk, so a later `takes add` would number the next row past a
+    // take the DB never saw. update/supersede/resolve already resolve first.
+    const pageId = await getPageId(engine, slug, sourceId);
+
     const path = pageFilePath(brainDir, slug);
     const body = readBodyOrEmpty(path);
     const { body: nextBody, rowNum } = upsertTakeRow(body, {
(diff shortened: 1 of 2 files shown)


time-attack:
Merged in the v0.45.8.0 community fix-wave (#4050). Your fix landed on master as 8db07e4a79, credited to you via Co-Authored-By, and is called out in the v0.45.8.0 CHANGELOG entry. Thank you for the contribution!

PR #3782, opened 2026-08-03
fix(write-through): honor the page's recorded source_path instead of re-deriving <slug>.md

JonMcCutchen:
## The bug

`writePageThrough()` rebuilds the disk target from the slug (`<slug>.md`) and ignores `pages.source_path`, which already records the file the row was imported from.

Whenever the on-disk filename isn't the slug — the normal case for a human-authored Obsidian vault, where `Library/People/Steve Jobs.md` slugifies to `library/people/steve-jobs` — every `put_page` / `capture` mints a **second** file beside the original:

```
Library/People/Steve Jobs.md   <- authored, now stale
Library/People/steve-jobs.md   <- twin written by write-through
```

Two artifacts for one row, with the newer content in whichever one the caller wasn't looking at.

### It's a sync-correctness bug, not just clutter

`gbrain sync` keys delete-reconcile on `source_path` (`collectMissingSourcePaths`), so the twin is **invisible to reconcile**. Removing the *original* file makes the page look absent from disk and deletes it — even though a file for that page is still sitting right there. Same data-loss family as #2426 / #2607.

### Reproduction

```bash
# a vault file whose name isn't its slug
mkdir -p "$REPO/Library/People" && printf -- '---\ntitle: Steve Jobs\ntype: person\n---\n\n# Body\n' \
  > "$REPO/Library/People/Steve Jobs.md"
gbrain import "$REPO"                      # slug=library/people/steve-jobs, source_path set
gbrain put library/people/steve-jobs < updated.md
ls "$REPO/Library/People/"
# before: 'Steve Jobs.md' AND 'steve-jobs.md'   <- twin
# after:  'Steve Jobs.md'                        <- updated in place
```

## The fix

1. **Prefer `source_path` when set**, behind a structural sanitizer that rejects absolute paths, `..` traversal, NUL bytes, and non-`.md` values. The existing `isWriteTargetContained()` check still runs after the join, so this only *narrows* what's accepted.

2. **Fall back to a contained `file://` `source_uri` when `source_path` is NULL.** `gbrain capture --file` records the absolute input path as `source_uri` but never sets `source_path`, so capturing a file the user authored inside the brain repo would mint a twin next to the very file it had just read. URIs resolving outside the page root are ignored.

`resolvePageFilePath()` remains the fallback for genuinely file-less pages (inbox captures, `brainstorm --save`), so their behavior is unchanged.

## Tests

Six cases added to `test/write-through.test.ts`:

- Title-Case `source_path` honored, no twin created
- NULL `source_path` still falls back to the slug-derived path
- `source_uri` recovery when `source_path` is NULL
- `source_uri` pointing outside the repo is ignored
- a traversing `source_path` is refused rather than joined

`bun run verify` — 31/31 green. Write-through, put-page-write-through, write-through-commit, storage-export and three sync suites all pass (60 tests).

## One note for reviewers

The new tests assert "no twin" via a directory walk rather than `existsSync()`. When a slug differs from the authored name only by **case**, a case-insensitive filesystem (macOS/Windows default) folds the two paths, so `existsSync()` reports a twin that doesn't exist. Only enumerating real directory entries is truthful on every filesystem. This is adjacent to the collision guard from #2831, which correctly doesn't fire here — `Steve Jobs.md` and `steve-jobs.md` are genuinely different names, not a case fold.

### DIFF
--- a/src/core/write-through.ts
+++ b/src/core/write-through.ts
@@ -22,7 +22,7 @@
  */
 
 import { existsSync, statSync, mkdirSync, writeFileSync, renameSync, unlinkSync, readdirSync } from 'fs';
-import { basename, dirname, join } from 'path';
+import { basename, dirname, isAbsolute, join, relative, resolve } from 'path';
 import { randomBytes } from 'crypto';
 import type { BrainEngine } from './engine.ts';
 import { serializePageToMarkdown, resolvePageFilePath } from './markdown.ts';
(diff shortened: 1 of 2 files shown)


time-attack:
Merged in the v0.45.10.0 community fix-wave 2 (#4085). Your fix landed on master as f8b0ececcb, credited to you via Co-Authored-By, and is called out in the v0.45.10.0 CHANGELOG entry. Thank you for the contribution!

PR #3784, opened 2026-08-03
fix(pglite-lock): Windows PID-reuse guard for stale-lock reaper

zhangyimin870220:
## Root cause

On Windows, PIDs are recycled aggressively. When a gbrain process holding the
PGLite data lock is force-killed (OOM, manual `taskkill`, crash), its PID can be
reassigned to an unrelated live process within seconds. The stale-lock reaper in
`src/core/pglite-lock.c` used a bare `process.kill(pid, 0)` to test liveness.
After PID reuse that call returns `true` (the PID *is* alive — just owned by
someone else), so the reaper never treats the lock as stale and the next gbrain
invocation hangs until its 30s `acquire` timeout. The existing comment in the
live-holder branch already acknowledges this ("or its PID was reused").

This was the cause of a real incident: two force-killed `bun run src/cli.ts`
sessions leaked ~5.5 GB and froze the host; after killing them the stale lock
then blocked every subsequent run for 30s.

## Fix

On `win32`, `isProcessAlive` now additionally verifies the live PID's image
name via `tasklist` against a known gbrain host allowlist (`bun.exe` /
`node.exe`). A recycled PID owned by an unrelated process is correctly treated
as stale and reaped. The POSIX path is unchanged (`process.kill(pid, 0)`).

## Verification

- Reproduced the failure by planting a stale lock claiming a live non-gbrain PID
  (python.exe): before the fix this hangs 30s; after the fix it is reaped and
  `gbrain stats` exits 0 in ~13s with no lingering bun process.
- `bun --check` passes on the patched module.

## Notes

- The one-shot CLI process-hang issue from the same incident (PGlite WASM
  handle keeping the event loop alive) was already resolved upstream (bounded
  teardown + explicit exit), so this commit touches only the lock layer.
- Lock reaping only ever deletes the `.gbrain-lock` directory; data files under
  `base/` are untouched.

### DIFF
--- a/src/core/pglite-lock.ts
+++ b/src/core/pglite-lock.ts
@@ -16,6 +16,7 @@
 
 import { mkdirSync, existsSync, readFileSync, writeFileSync, rmSync, statSync } from 'fs';
 import { join } from 'path';
+import { execSync } from 'child_process';
 import { parseGlobalFlags } from './cli-options.ts';
 
 const LOCK_DIR_NAME = '.gbrain-lock';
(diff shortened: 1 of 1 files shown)


time-attack:
Thanks -- the win32 PID-recycling failure mode is real, the incident repro is convincing, and tasklist image verification is the right primitive. Closing this branch as-is for two reasons rooted in how this seam is maintained. First, the error posture is inverted relative to master: v0.42.75.0 hardened isProcessAlive to fail-closed -- uncertainty must read ALIVE, because a false 'dead' reaps a live single-writer's lock, a corruption class this file has been burned by before (#2348). As written, `catch { return false }`, `pid <= 0 -> false`, and an allowlist miss all reap on uncertainty, and LOCK_OWNER_IMAGES omits the compiled binary (gbrain ships via `bun build --compile`, so a live compiled-binary holder would be classified stale and stolen). Second, the branch conflicts with that hardening and carries no discriminating test. The salvageable cut is small and squarely fix-lane: the same ~35-line win32 hunk, resubmitted fail-closed -- return 'dead' ONLY when tasklist succeeds and affirmatively shows a foreign image (include the compiled binary in the allowlist, or compare against an image name recorded in the lock file at acquire), with every error/timeout/malformed-pid path reading ALIVE -- rebased onto current master, plus a mocked win32 test covering foreign-image -> dead, tasklist-failure -> alive, compiled-binary -> alive, with a fails-on-unpatched-master receipt. Send that alone and it rides the next fix wave. Closing the branch, not the idea.

PR #3787, opened 2026-08-03
fix(models): replace the /v1 doctor hint with a /models-probe base-URL classifier

brettdavies:

I run gbrain behind an OpenAI-shaped proxy (litellm / codex-proxy). When I point the base URL at the proxy without the `/v1` suffix, every `gbrain models doctor` probe comes back 401, and the doctor tells me to check the API key. The key is fine; the base URL is just missing `/v1`. The static hint that shipped in v0.42.69.0 (#3553) is supposed to catch this, but it misfires: it tells a bad zhipu key to append `/v1` onto a `/v4` route, it only looks at the chat probe, and it never actually checks that `/v1` is where the route lives. I wanted the doctor to point at the URL when the URL is the problem, and at the key when the key is.

Screenshot 2026-08-03 at 2 35 50 PM

Text from the screenshot:
```text
LITELLM_BASE_URL=http://127.0.0.1:4599 LITELLM_API_KEY=demo-key bun run src/cli.ts models doctor --skip=zeroentropyai

[models] tier.subagent resolved to "litellm:gpt-4o" via "models.default" — provider does not support prompt caching. The loop will run hot (cost scales linearly with conversation length). For lower cost on long loops, set models.tier.subagent to an Anthropic model.
Model reachability probe:
  ✔ embedding_config  zeroentropyai:zembed-1                             ok (0ms)
  ✔ reranker_config   (none)                                             ok (7ms)
  ✘ chat              litellm:gpt-4o                                     auth (28ms)
      [chat(litellm:gpt-4o)] Unauthorized
      fix: The API key authenticates at http://127.0.0.1:4599/v1 (its /models requires auth and accepted the key), but the configured http://127.0.0.1:4599 does not serve that route, so the base URL is missing its version prefix — append /v1 (base URL http://127.0.0.1:4599/v1).
  ✘ expansion         litellm:gpt-4o                                     auth (1ms)
      [chat(litellm:gpt-4o)] Unauthorized
      fix: The API key authenticates at http://127.0.0.1:4599/v1 (its /models requires auth and accepted the key), but the configured http://127.0.0.1:4599 does not serve that route, so the base URL is missing its version prefix — append /v1 (base URL http://127.0.0.1:4599/v1).

Summary: 2/4 reachable.
```

**What changed**

v0.42.69.0 (#3553) shipped `models doctor`'s static `/v1` base-URL hint, which guesses "append /v1" from the base-URL string alone. Its review raised four issues that shipped unaddressed: the gate false-fired on a recipe's own versioned default (zhipu `/v4`); only the pure helper was tested, not the wiring; only the chat probe was patched while the embedding and reranker probes hit the same trap; and the base-URL resolution hand-rolled half of `applyOpenAICompatConfig`.

This replaces the helper with a probe-based classifier in `src/core/ai/base-url-probe.ts`. On an `auth` (401/403) or `model_not_found` (404) probe failure against an OpenAI-compatible recipe, it GETs `/models` at the configured base URL and each canonical version off the URL's version-root, with the same bearer, and reads the status codes into one verdict: append/switch (verified via an unauthenticated control-probe), URL-and-credentials, key-or-model, key-only, wrong-endpoint, or silent when inconclusive. A version is recommended only when its `/models` answers 200, so the guess is gone.

- All four #3553 review issues closed:
  - the version is discovered, not assumed;
  - the wiring is tested end-to-end at all three probe sites (reverting any site's hint call fails a test);
  - resolution goes through `applyOpenAICompatConfig` / `applyResolveAuth` / a shared `authToHeaders` / `requireConfig`.
- Provider-agnostic (reads the observed `/models` status, no per-provider table):
  - authenticated `/models` (litellm, ollama),
  - public `/models` (OpenRouter, which 200s on a bad key, so "check the key"),
  - and no-`/models` providers (Voyage embeddings, the default ZeroEntropy reranker, which never get a false "fix your URL").
- Fail-open, `redirect: 'manual'` (a 3xx never forwards the bearer), memoized per doctor run.

**How it was tested**

`test/models-doctor-v1-hint.test.ts` (30 cases): every verdict in both trigger contexts, the control-probe confident-vs-ambiguous split, the inconclusive and silent cases, the per-run memo, the gate exclusions, and end-to-end discrimination at all three probe sites. Every case injects the transport and a header-aware stub `fetch`, so it runs with no network. The adjacent doctor and gateway suites still pass; `bun run typecheck` is clean.

### DIFF
--- a/src/commands/models.ts
+++ b/src/commands/models.ts
@@ -34,7 +34,8 @@ import {
   resolveModel,
   type ModelTier,
 } from '../core/model-config.ts';
-import { resolveRecipe } from '../core/ai/model-resolver.ts';
+import { maybeAttachVersionSuffixHint } from '../core/ai/base-url-probe.ts';
+import type { AIGatewayConfig } from '../core/ai/types.ts';
 
 const TIERS: ModelTier[] = ['utility', 'reasoning', 'deep', 'subagent'];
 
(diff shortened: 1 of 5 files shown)


time-attack:
Merged in the v0.45.8.0 community fix-wave (#4050). Your fix landed on master as b0b9af042f, credited to you via Co-Authored-By, and is called out in the v0.45.8.0 CHANGELOG entry. Thank you for the contribution!

Issue #3657, opened 2026-07-30
Default reranker zeroentropyai:zerank-2 is being sunset 2026-09-04 — new signups already disabled


ethanbeard:
## Summary

`DEFAULT_RERANKER_MODEL` is `zeroentropyai:zerank-2`. ZeroEntropy has been acquired by Notion and is shutting down:

> "All ZeroEntropy products will remain fully supported until **September 4th, 2026** with no degradation, after which they will be sunset. **We have disabled new signups**, and encourage existing customers to migrate away from ZeroEntropy products."
> — [ZeroEntropy Joins Notion](https://www.zeroentropy.dev/articles/zeroentropy-is-joining-notion)

The signup half is what makes this urgent rather than a September problem. **A new user who enables reranking today has no reachable provider and no way to obtain one** — ZeroEntropy is the only hosted provider in the recipe registry that declares a `reranker` touchpoint, and its dashboard no longer accepts new accounts. The failure is a silent fail-open: search returns un-reranked results and logs a warning.

## Where it's pinned

- `src/core/ai/gateway.ts` — `const DEFAULT_RERANKER_MODEL = 'zeroentropyai:zerank-2'`
- `src/core/search/mode.ts` — `reranker_model: 'zeroentropyai:zerank-2'` hardcoded in three mode bundles (~lines 297, 347, 402), so setting `search.reranker.model` isn't obviously sufficient depending on which bundle is canonical

Of the recipes that declare a reranker touchpoint (`zeroentropyai`, `dashscope-rerank`, `llama-server-reranker`, `openrouter`), only ZeroEntropy is a general-purpose hosted option; the others are region-specific, self-hosted, or a gateway.

## Why this went unnoticed for a while (possibly worth a separate fix)

`gbrain models doctor` could not have told anyone. `probeModel()` in `src/commands/models.ts` calls `chat({ ..., maxTokens: 1 })`, and OpenAI rejects `max_output_tokens` below 16:

```
Invalid 'max_output_tokens': integer below minimum value. Expected a value >= 16, but got 1 instead.
```

So the probe failed on **every** OpenAI chat/expansion model regardless of reachability, and a genuinely dead provider was indistinguishable from the noise. On our install this masked ~960 reranker auth failures over 7 days. Changing `maxTokens: 1` → `16` makes the check meaningful again (4/6 → 6/6 here, with the one real failure correctly isolated).

## Proposed fix

[...]

PR #3790, opened 2026-08-03
fix(models): probe chat with max_output_tokens=16, not 1 (#3657)

ethanbeard:
Fixes the (a) piece of #3657 — the sub-bug triage called out as a safe one-liner.

## The bug

`probeModel()` in `src/commands/models.ts` sent `maxTokens: 1`. OpenAI rejects
`max_output_tokens` below 16:

```
Invalid 'max_output_tokens': integer below minimum value.
Expected a value >= 16, but got 1 instead.
```

So `gbrain models doctor` reported **every** OpenAI-family chat and expansion
model as unreachable whether or not it actually was — which is the single thing
the probe exists to distinguish. A dead model and a healthy one produced the
same verdict, and the sunset/upgrade guidance downstream keys off it.

16 is the documented floor. The probe still costs at most 16 output tokens.

## The test fails without the fix

Per the contributing bar, here is the same test on unpatched `master`
(`maxTokens: 1`) — it fails, and the mock's own record shows why:

```
--- max-output-tokens seen by the mock: [1,1] ---
206 |       expect(chat!.status).toBe('ok');
                                 ^

Expected: "ok"
Received: "unknown"

(fail) models doctor chat probe respects the provider max-output-tokens floor
       (#3657) > chat + expansion probe as reachable against a provider that
       enforces the >= 16 floor [3213.66ms]

 0 pass
 1 fail
```

With the one-line fix applied, same test, same environment:

```
 1 pass
 0 fail
 11 expect() calls
Ran 1 test across 1 file. [3.46s]
```

## How the test is built

It asserts behavior, not source text. It spawns the real CLI
(`bun run src/cli.ts models doctor --json`) against a tmpdir PGLite brain and
parses the machine-readable report, then asserts the process exit code.

The openai provider is pointed at a local mock via `OPENAI_BASE_URL`, so the
SDK emits its genuine wire shape — the mock receives
`POST /v1/responses {"model":"gpt-4o-mini",...,"max_output_tokens":1}` — and
**no API key or network is involved**. The mock enforces the floor exactly as
OpenAI documents it: reject when the cap is present and `< 16`, with OpenAI's
verbatim error body; otherwise answer normally. That is what makes the test
red on `maxTokens: 1` and green on 16, rather than it merely restating the
constant.

Two guards against the test passing for the wrong reason:
- it asserts the mock actually received a request, so a change that stopped
  probing entirely would fail rather than pass;
- it asserts every observed cap is `>= 16`, not that it equals any particular
  number.

`--skip=zeroentropyai --skip=llama-server` keeps unrelated providers out of the
summary, so the exit code is a clean function of the thing under test.

## Checks

- all 18 `test/*model*.test.ts` files — 149 pass / 0 fail

## Not included

The other two pieces of #3657 are deliberately out of scope here: the
`checkProviderSunset` reranker-plane feedback on #3548, and the default-provider
swap, which is the product decision deferred by closing #3492/#3481. Happy to
follow up on the Voyage reranker adapter (`top_k` / `data[]`) separately if the
provider decision lands that way — that one is running in production on this
end and can be offered with the same red/green treatment.

### DIFF
--- a/src/commands/models.ts
+++ b/src/commands/models.ts
@@ -589,7 +589,13 @@ async function probeModel(modelStr: string, touchpoint: 'chat' | 'expansion'): P
       await chat({
         model: modelStr,
         messages: [{ role: 'user', content: '.' }],
-        maxTokens: 1,
+        // OpenAI rejects max_output_tokens below 16 ("Invalid
+        // 'max_output_tokens': integer below minimum value. Expected a value
+        // >= 16, but got 1 instead."), so a probe of 1 fails for EVERY
+        // OpenAI-family chat model regardless of whether it is reachable —
+        // which is precisely what this check exists to tell apart. 16 is the
+        // documented floor; the probe still costs at most 16 output tokens.
+        maxTokens: 16,
         abortSignal: controller.signal,
       });
       return { model: modelStr, touchpoint, status: 'ok', message: 'reachable', elapsed_ms: Date.now() - start };
--- a/test/models-probe-max-output-tokens.serial.test.ts
+++ b/test/models-probe-max-output-tokens.serial.test.ts
@@ -1,0 +1,226 @@
+/**
+ * #3657(a) — `gbrain models doctor` must probe chat models with a
+ * max-output-tokens value the provider will actually accept.
+ *
+ * Pre-fix, probeModel() sent `maxTokens: 1`. OpenAI rejects
+ * max_output_tokens below 16 with a 400 ("Invalid 'max_output_tokens':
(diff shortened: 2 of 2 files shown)


time-attack:
Status update: this PR was reviewed in the 2026-08-18 full-backlog triage and confirmed valuable — it sits in the verified merge queue (position 12 of 53, risk: low). CLEAN green. One-line probe floor (#3657) + failing-first regression test. No action needed from you right now — thank you for the contribution and the patience.

time-attack:

Issue #1522, opened 2026-05-27
IngestionEvent provenance fields (source_id, source_kind, source_uri) dropped by ingest_capture handler before page write


TTJC24:
# IngestionEvent provenance fields (source_id, source_kind, source_uri) dropped by ingest_capture handler before page write

## Summary

`ingest_capture` validates `IngestionEvent` payloads that include `source_id`, `source_kind`, and `source_uri`, but the handler does not thread those fields into `importFromContent` when it writes the page. The result is that the documented public ingestion path can accept a valid event, write a page, and silently lose the event's source isolation/provenance fields.

## Reproduction

- gbrain version: `0.41.14.0`
- Engine: PGLite
- Setup: local install, ZeroEntropy embeddings
- Steps:
  1. Submit an `IngestionEvent` through the public `ingest_capture` handler with `source_id`, `source_kind`, and `source_uri` all populated.
  2. Let the handler validate the event and write the page.
  3. Inspect the resulting page row.

The symptom is that validation passes and the page is written, but the page row is not written under the submitted `source_id`, and the `source_kind` / `source_uri` provenance fields are not persisted from the event.

## Expected vs Actual

Expected: the page has `source_id`, `source_kind`, and `source_uri` matching the submitted `IngestionEvent`.

Actual: those provenance fields are not carried through by the `ingest_capture` write path.

## Evidence

We verified the contract works end-to-end when `source_id`, `source_kind`, and `source_uri` are threaded manually via direct `importFromContent` calls. Play B documents this in detail at `https://github.com/TTJC24/gbrain-eval` (private repo, can share relevant excerpts if useful). Specifically:

- Bypassing `ingest_capture` and calling `importFromContent` directly with the provenance fields threaded: `source_id` / `source_kind` / `source_uri` all persist correctly, isolation tests pass, idempotency works.
- Using `ingest_capture` as documented: validation passes, page is written, but provenance fields are dropped before the page write.

The direct write shape used in Play B was:

```ts
await importFromContent(engine, slug, event.content, {
  sourceId: SOURCE_ID,
  noEmbed: true,
  source_kind: event.source_kind,
  source_uri: event.source_uri,
  ingested_via: 'ingestion-source',
});
```

With that direct threading, a dedicated `m365-calendar` source preserved provenance, citations, idempotency, and explicit source isolation across 21 imported calendar event pages.

## Code location

In `src/core/ingestion/types.ts`, `IngestionEvent` declares the fields as required provenance:

[...]

PR #3791, opened 2026-08-03
fix(ingest): attribute webhook captures to the client's trusted source

igbymyboy:

I was inspecting my gbrain database and noticed that the `source_id` value I'm passing in when I call the `/ingest` endpoint was being dropped and that everything had a `source_id` of `default`. This is a problem b/c it opens the possibility of slug collision and also is lossy at best and misleading at worst. This PR addresses this disconnect and ensures the client's registered source (gbrain auth register-client --source ... / rescope-client) flows all the way through.

screenshot-3791

---

## What changed

`POST /ingest` wrote every capture under source `default`, even when the authenticated OAuth client was registered with a source.

The webhook read `X-Gbrain-Source-Id` into the ingestion event, but the handler refused a caller-supplied id at the write — correctly, since a caller must not choose its own write partition. So the refusal was the only behavior: nothing ever routed, and the header only ever came back in the response. Because page identity is `(source_id, slug)` and the caller controls the slug via `X-Gbrain-Slug`, every integration shared one namespace and could overwrite another's pages.

The trusted alternative already existed and was unused: `authInfo.sourceId`, populated from `oauth_clients.source_id`, which the MCP write path already uses at `serve-http.ts:1868`. `/ingest` goes through the same `requireBearerAuth` verifier, so the value was one field access away.

**`src/commands/serve-http.ts`**
- `X-Gbrain-Source-Id` no longer drives routing. `event.source_id` becomes the emitter identity (`webhook-<clientId>`), which is what that field is documented to mean.
- The write source resolves server-side as `authInfo.sourceId ?? 'default'` and travels on the job as `job.data.sourceId`, mirroring how `slug` already flows.
- The effective source joins the queue idempotency key, so a client rescoped from one source to another lands a new capture instead of being deduped against its old job.
- The `202` gains an additive `write_source_id` naming the routed destination. The existing `source_id` field is unchanged for back-compat.

**`src/core/minions/handlers/ingest-capture.ts`**
- Trusted `job.data.sourceId` takes precedence, validated against a live (non-archived) `sources` row. `archived` is a soft delete with a recovery window, so a row-existence check alone would happily write into a source that was just retired.
- The #1522 daemon path (`event.source_id` with `!untrusted_payload`) is unchanged and stays silent. An unregistered emitter id is normal for `inbox-folder` / `file-watcher`, so reporting it would mean a per-capture warning in steady state.
- A source FK violation retries once under the default source, closing the check-then-write race. The match requires the error to name the source constraint (the discrimination `maybeRewriteSourceFkError` already uses), so an unrelated FK failure propagates instead of being silently rewritten as "source unavailable."
- `IngestCaptureResult` gains an optional `source_fallback` describing any redirect and its reason (`not_registered` / `archived` / `fk_violation`).

**`src/core/ingestion/types.ts`**
- The `metadata` doc comment promised persistence into page frontmatter under `ingestion_metadata`. That is unimplemented — the only reference in the tree was the comment itself. It now says so rather than promising behavior that does not exist.

### Upgrade note

The per-request header is not the mechanism any more. A client that was sending `X-Gbrain-Source-Id` needs its source set once on the client instead:

```
gbrain auth register-client my-app --grant-types client_credentials --scopes "read write" --source my-app-notes
# or, for an existing client:
gbrain auth rescope-client <client_id> --source my-app-notes
```

`--source` is pre-existing on both commands; this PR adds no CLI surface. It connects a wire that was already present on both ends.

### Compatibility

No path becomes more restrictive than current behavior.

| Situation | Write source | Same as before? |
|---|---|---|
| Client registered with `--source X` (live) | `X` | new — the fix |
| Client registered without a source | `default` | yes |
| Legacy bearer token | `default`, or its `permissions.source_id` grant if set | yes |
| Source unregistered, archived, or lost to an FK race | `default` | yes |
| Daemon / non-webhook emitters | unchanged | yes |
| `ingest_capture` jobs queued before upgrade | unchanged | yes |

No schema change, no migration. `job.data.sourceId` and `source_fallback` are additive and optional, so in-flight jobs and existing consumers are unaffected. Rollback is reverting the PR.

Composes with the slug-prefix fence added in v0.42.72.0: that gate returns `403` for slug-bound clients before this routing runs, so bound clients stay blocked and only unbound clients reach source attribution.

## How it was tested

| Layer | Coverage | Added |
|---|---|---|
| Unit (`test/ingestion/ingest-capture.test.ts`) | trusted source routes; unregistered → default + `not_registered`; archived → default + `archived`; FK violation → retry under default + `fk_violation`; unrelated-constraint 23503 propagates; non-FK error propagates; `sourceId: 'default'` production shape; daemon path silent for both registered and unregistered emitter ids | +10 |
| E2E (`test/e2e/serve-http-ingest-webhook.test.ts`) | scoped client → its source; scoped client + spoofed `X-Gbrain-Source-Id` → still its source; unscoped → default; archived-source client → default | +4 |
| Integration (`test/e2e/ingestion-roundtrip.test.ts`) | daemon roundtrip asserts `source_id` plus the provenance columns | +1 |

`bun run typecheck` clean. Targeted suites: 251 pass / 0 fail. Full sharded suite (`bun run test`): 10417 pass / 10 fail.

Those 10 are pre-existing and unrelated to this diff. Each was verified against a clean `master` worktree and reproduces identically there: autocut ×3 and hybrid-meta ×2 (9 pass / 5 fail on master, byte-identical), ambient-channel reflex ×1 and op-layer capture ×2 (identical on master). The remaining two (post-commit hook, skillify-check) pass in isolation and depend on git-origin reachability and file recency. No failure occurs in any file this branch touches.

The Postgres HTTP e2e cases are `DATABASE_URL`-gated and skip without Docker locally; they run in CI.


time-attack:
Status update: reviewed in the 2026-08-18 full-backlog triage — still wanted, but not yet mergeable: Valid, unlanded fix routing webhook captures to the client's registered source, but the branch is CONFLICTING with master and needs a rebase. If you update it, it will be re-reviewed promptly; alternatively the fix may be adopted with credit in an upcoming fix wave. Thanks for the contribution and the patience.

igbymyboy:
Rebased onto master. No behavior change from the rebase itself. Three things worth calling out:

- **`src/core/import-file.ts` is gone from the diff.** This branch carried the unscoped-check/scoped-write fix for `importFromContent`; v0.46.11.0 (`5ef85ac9`) landed the same fix in the wider sweep of that bug class, so I dropped mine and took master's.
- **`docs/architecture/KEY_FILES.md`** — the `serve-http.ts` entry was rewritten upstream since this branch was cut. I took the new entry whole and re-applied this PR's two edits on top: the corrected fence rationale, and the routing description.
- **Two new guards.** `scripts/module-size-limits.tsv` raises the `serve-http.ts` ceiling 3194 → 3212, with a note naming this PR. And `test/e2e/serve-http-ingest-webhook.test.ts` now calls `assertSafeE2eDatabaseUrl()` — it connects a `PostgresEngine` to the ambient URL and cleans up its own fixtures, which the new coverage gate flags.

time-attack:

PR #3803, opened 2026-08-04
v0.43.0.0: Persist per-source sync strategy

mamedov:

While aligning gstack’s explicit /sync-gbrain path with gbrain sync --strategy auto, I discovered that scheduled and autopilot jobs configured independently could still fall back to the legacy Markdown default for the same source. Storing the sync strategy on the source ensures a single, consistent policy across all sync entry points—manual, scheduled, watch, retry, cycle, Minions, autopilot, and local MCP—without implicitly enabling multimodal processing.

Disposable smoke test: a source stores the code strategy and uses it during a real sync

The screenshot is from a disposable local source and shows the persisted `code` strategy, its `stored` provenance, and a real code-file import. It contains no private repository data or credentials.

**What changed**

- Add canonical `markdown`, `code`, and `auto` source strategy values plus explicit provenance (`override`, `stored`, `default`, or `invalid_fallback`).
- Accept `sources add --strategy` and add `sources set-sync-strategy <id> <markdown|code|auto|default>` without replacing the source or dropping unrelated configuration.
- Resolve the effective strategy inside the source writer lock for every sync caller; an invocation override remains one-off and never mutates stored policy.
- Keep existing sources on the historical Markdown default, safely fall back on malformed legacy values, and emit a bounded repair command. No migration or historical backfill is required.
- Reject `sync --all --strategy`, where one global override would defeat per-source policy.
- Keep multimodal enrichment as a separate opt-in gate.
- Show the effective strategy and provenance in human and JSON source output, and use the same resolver for cost previews.
- Document the policy, command surface, architecture path, and test contracts.

Basecamp tracking card: https://3.basecamp.com/4058523/buckets/34718435/card_tables/cards/10157599174

**How it was tested**

- `bun run verify` — 34/34 checks passed on final HEAD.
- Focused strategy suite — 132 tests, 335 assertions, 0 failures.
- Focused Postgres strategy contracts — 12 tests, 0 failures.
- Slow suite — 41 tests, 1,604 assertions, 0 failures.
- Unit matrix — 15,400 passed, 20 skipped; the two remaining failures reproduce on clean `origin/master` (the existing retrieval-reflex async race and worker-registry live-PID round trip).
- Full E2E was compared with clean `origin/master`; all observed failures reproduce on master. The one apparent feature-only `engine-parity` result also reproduces standalone on clean master (27 passed, 1 failed).
- Disposable live smoke — stored `code` policy resolved as `stored` and imported a real code file.

## Documentation

- `docs/guides/multi-source-brains.md` provides the command reference, setup examples, precedence rules, recovery path, and the independent multimodal boundary.
- `docs/architecture/KEY_FILES.md` records the shared resolver, lock-time policy snapshot, callers, and test contracts.
- `docs/TESTING.md` maps the focused CLI, resolver, Postgres, cost-preview, and sync integration suites.
- `CHANGELOG.md` explains the user-visible behavior, exact commands, compatibility default, and no-backfill posture.

Coverage: the shipped surface has reference, how-to, and explanation coverage; no stale architecture diagram or documentation debt was found.


time-attack:

time-attack:
Thanks — this is one of the strongest external PRs in the queue: the inconsistency you found is real (single-source sync ignores stored config.strategy while --all honors it), the resolver-inside-the-writer-lock design is right, and the test coverage including the Postgres parity contract is exactly what we ask for. The blocker is shape, not craft. New product surface (the set-sync-strategy subcommand, the --strategy flag on sources add, the changed sync --json envelope, the sources list schema_version wrapper, the cost-gate path-precedence swap) defaults to no here regardless of implementation quality, and bundling several of those in one PR — plus a self-allocated 0.43.0.0 version bump, which happens at ship time — is what keeps this out of the fix wave. There is a clean cut that goes in immediately: a small PR that makes the single-source sync path read stored config.strategy exactly like the --all path already does (no new CLI, no output changes, no version files), with a test that fails on unpatched master. Send that alone and it lands in the fix lane. If you'd like the full per-source strategy surface considered as a feature, split it into its own follow-up so it can get a product decision on its own merits.

PR #3804, opened 2026-08-04
fix(links): resolve relative markdown links against the linking page's directory

alexey-metaengage:
## Problem

`extractPageLinks` resolves relative markdown links (`[x](../concepts/x.md)`) as if they were root-relative, ignoring the directory the linking page actually lives in. The markdown-link regex matches `(?:\.\.\/)*` and **discards** the `../` run, so `[x](../concepts/x.md)` always yields the target `concepts/x` no matter where the linking page sits.

For the canonical flat layout (entity dirs at the git root, pages one level down) this happens to be correct — `../` unwinds exactly to the root. But when content is nested under a prefix — e.g. a source ingested with `--src-subpath <dir>`, whose slugs are `<dir>/concepts/x` — every cross-directory link misses: the extractor emits `concepts/x`, which doesn't exist; only `<dir>/concepts/x` does. `resolveCandidateSources` then drops the candidate (target not in the slug set), so `extract`/sync produce **zero** cross-dir edges and the graph shows only `contains` edges.

## Fix

Preserve the `../` depth on the ref (`EntityRef.upLevels`) and, in `extractPageLinks` (which already knows the linking page's slug), resolve the target against that page's own directory — standard filesystem-relative semantics.

- Backward-compatible: a page one level below the root lands exactly at the root, so results are identical to the old `../`-stripping behavior. It only diverges — correctly — when the page is nested deeper than the `../` count unwinds. An over-long `../` run clamps at the root. Absolute refs (engine-slug `people/x`, `[[wikilinks]]`) are unchanged.
- `upLevels` is only set for relative refs, so the `EntityRef` shape for absolute refs is untouched (no churn for equality consumers).
- Bumps `LINK_EXTRACTOR_VERSION_TS` so already-stamped pages re-sweep and pick up the newly-resolvable edges.

## Tests

New `test/link-extraction-relative-path.test.ts` covers single/multi `../`, over-long clamp, absolute-ref passthrough, and `upLevels` capture, plus regression pins that the flat layout (incl. the #2576 cases and verb inference) is unchanged. Updated the one existing `extractEntityRefs` case that now carries `upLevels: 1`.

All link-extraction suites pass. `bun run typecheck` clean. (One unrelated `backlinks.test.ts` dedupe case fails identically on `master` — pre-existing, different code path, no `../`.)

### DIFF
--- a/src/core/link-extraction.ts
+++ b/src/core/link-extraction.ts
@@ -43,7 +43,7 @@ import { slugifyPath } from './sync.ts';
 // PRE-wave code after this date reads as fresh and won't re-extract until
 // the page is next edited; no fixed watermark can cover code that keeps
 // running past it.
-export const LINK_EXTRACTOR_VERSION_TS = '2026-08-01T00:00:00Z';
+export const LINK_EXTRACTOR_VERSION_TS = '2026-08-04T00:00:00Z';
 
 // ─── Entity references ──────────────────────────────────────────
 
@@ -75,6 +75,16 @@ export interface EntityRef {
    * LinkCandidate per resolved match.
    */
   needsResolution?: boolean;
+  /**
+   * Number of leading `../` segments on the original markdown link (0 when
+   * absent or for engine-slug refs). The match regex strips the `../` run, so
+   * this preserves the relative depth for callers that know the linking page's
+   * slug (`extractPageLinks`) to resolve the target against the page's own
+   * directory. Without it, `[x](../concepts/x.md)` from a nested page resolves
+   * as the root-relative `concepts/x` instead of `<page-dir-parent>/concepts/x`
+   * — the reason subtree-scoped / nested brains got zero cross-dir edges.
+   */
+  upLevels?: number;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


time-attack:
Automated triage ran the full gate on this PR. Results:

**Passed:** PR CI all green at head fa39eb41; targeted link-extraction suites (199 tests) green; sibling-pinning sweep (public-exports, extract, extract-stale, title-retrieval-arm, sync-inline-extract-stamps, link-inference-pack — 115 tests) green; `bun run verify` 34/34; before/after consumer harness on 3 pinned public adopters confirmed the claimed fix (baseline: `extract links --source db` creates 0 cross-dir edges for subtree-scoped content; candidate: edge present).

**Blocked on adversarial review** — two independent reviewers converged on the same concrete regression:

For a page **≥2 directories deep** with an *under-counted* `../` run — e.g. `meetings/2026/standup` containing `[x](../people/alice.md)`, the wiki-root-relative authoring style the FS path's `resolveSlug` ancestor walk deliberately supports — master emits the working target `people/alice`, while this PR emits the nonexistent `meetings/people/alice`, which is silently dropped. Because `LINK_EXTRACTOR_VERSION_TS` is bumped, the re-sweep + put_page reconciliation then **actively deletes those previously-working edges** on existing brains. The DB extraction path gets strict resolution with no ancestor/root fallback, so the two extraction paths now disagree, and no test pins this case.

Suggested direction: fall back to ancestor-walk (or root) resolution when the strictly-resolved slug doesn't exist in the slug set, mirroring the FS path's semantics — then the nested-prefix fix and the existing root-relative authoring style can both hold.

Queued for maintainer decision rather than auto-merged.

time-attack:

PR #3805, opened 2026-08-04
v0.42.73.1 revert(ci): remove the PR gate and withdraw the v0.42.72.1 contribution requirements

time-attack:
Removes the PR gate that v0.42.73.0 added, withdraws the v0.42.72.1 contribution policy it enforced, and says plainly why.

## Why

The gate needs two things this repository does not grant: an `ANTHROPIC_API_KEY` Actions secret for its usefulness verdict, and read-write workflow permissions to post a comment or set a label. Both are admin-only. Without them it can only skip.

It also caused a real incident before that was understood. On its first live runs a read-only token turned every comment and label call into a 403, the code treated that as a crash, and the check went red on an outside contributor's PR (`feat/ingest-source-attribution`) four times, with no comment explaining why. v0.42.73.0 fixed the crash — that PR now passes — but a check that runs on every pull request and can never reach a verdict does not earn a place in the repo.

## What this removes

- **The gate** (commit 1): `.github/workflows/pr-gate.yml`, `scripts/pr-gate.mjs`, `scripts/pr-gate.d.mts`, `test/pr-gate-workflow.test.ts`. No remaining references anywhere (grepped across json/sh/mjs/ts/yml/md). The code is preserved in git history at v0.42.73.0.
- **The policy** (commit 2, maintainer follow-up): reverts 6d1232d5 (v0.42.72.1). The human-written intent paragraph and gbrain-in-use screenshot are no longer required; `CONTRIBUTING.md`, both issue templates, and the pull-request template return to their pre-2026-08-02 state. The v0.42.72.1 CHANGELOG entry stays as historical record; VERSION/package.json keep moving forward.

## What does not change

Issues and PRs are reviewed on their content by maintainers, as they were before 2026-08-02. Everything else v0.42.73.0 shipped (the five contributed correctness fixes, `slug_filter`, and the four dependency pins that cleared six CVEs) is unaffected.

## If the gate is ever restored

Give the mechanical half — the version-first title rule, the ten red flags, none of which need an API call — a `$GITHUB_STEP_SUMMARY` output path instead of a PR comment. That requires no token permission and no secret, so it works on a repository like this one.

## Verification

- `bun run typecheck` clean
- `bun run verify` 34/34 green (re-run after commit 2: 34/34 green)
- `bun run build:llms` regenerated; `test/build-llms.test.ts` green (`CONTRIBUTING.md` and the templates are not part of the llms bundle)
- Version trio: `VERSION` = `package.json` = CHANGELOG top = `0.42.73.1`


time-attack:
Pushed 698ba272: this PR now also reverts 6d1232d5 (the v0.42.72.1 contribution-guidelines change), per maintainer decision — removing the gate while keeping the policy it enforced left a requirement with no enforcement path. CONTRIBUTING.md, both issue templates, and the PR template are back to their pre-2026-08-02 state; the CHANGELOG 0.42.73.1 entry and the PR body are updated to match. `bun run verify` 34/34 green after the revert.

PR #3807, opened 2026-08-04
feat(export): standalone exporter writing conversation pages back to envelope-v0 (closes the #3549 round trip)

SeanGearin:
I wrote the envelope-v0 importer that merged as #3549 - scripts/envelope-to-gbrain.mjs - to facilitate bringing outside chat history/context into gbrain. Today, we are unable to export the history built using gbrain out using the same format. gbrain can read envelope-v0 but nothing writes it. Once a brain is seeded from an envelope file, that file is the only copy of that shape that exists. If I lose it I can't regenerate it from the brain it built.

This PR completes the circle - scripts/gbrain-to-envelope.mjs. It is built with the same specs as the importer: standalone, zero dependencies, deterministic, no network, and doesn't call gbrain. The exporter reads conversation pages and writes envelope-v0 back out so a brain that was seeded from a file can produce that file again. In this case, the cohesiveness of the record is important as users compound context over time and take their brain to other model, tools, platforms, and so on.

## What changes

The export half of the round trip: `scripts/gbrain-to-envelope.mjs`, a standalone script that reads gbrain conversation pages back into an envelope-v0 file. It is the counterpart of `scripts/envelope-to-gbrain.mjs` (#3549, made parser-legible in #3788) — that script reads the format, this one writes it. Zero dependencies, deterministic, no network, never calls gbrain; point it at a `gbrain export --dir` tree, a brain repo, or the importer's output.

**1. Identity from the frontmatter record, not from prose.** On pages the importer writes since #3788, per-message identity — the id and the full RFC 3339 timestamp — lives in a frontmatter `messages:` array, and body turn headers carry only a speaker and a minute-resolution clock. The exporter reads it back the same way: `messages[i]` belongs to the i-th body turn, and a header-shaped line is a turn boundary only when it carries the clock derived from the next recorded `ts` (the importer's own derivation, kept in lockstep). So message text cannot split a message by looking like a header, an id cannot break the line that carries it, and one trailing space no longer silently absorbs a turn. Pages without the key keep the exact legacy read. A recorded page that cannot be read honestly is refused loudly — unreadable record, non-string id, empty record, or a body anchoring a different number of turns than the record holds — skipped with both numbers on stderr, never joined wrong. Fenced code blocks are never scanned for headers or sentinels, with a staged demotion so a fence can never swallow the turns after it.

**2. One envelope per provider — a mixed directory no longer falsifies `source:`.** An envelope names ONE `meta.source_provider`. The previous revision collapsed a mixed-provider page set to the first provider in sorted path order (warned once), and re-importing that envelope stamped the winner onto every losing page: the identity check matched the conversation ids, read a legitimate refresh, and overwrote `source:` at exit 0 — with nothing on disk retaining the truth, since envelope-v0 has no per-conversation provider field. Merging several vendors' history into one directory is the format's mainstream case, so this fired on the pitch, not on an edge.

Now the set fans out — one conforming envelope per provider, named for it (`out.chatgpt.mve.json`, `out.claude.mve.json`), each listed on stdout with its own counts:

| | before | after |
|---|---|---|
| mixed chatgpt+claude directory | 1 envelope, one provider chosen for everyone | 2 envelopes, each true |
| the circle back into the same directory | claude pages come back `source: "chatgpt"`, exit 0, silent | every id-carrying page byte-identical |

A single-provider directory is unchanged: one file, the same name and words as before. Pages with no `source:` key are their own group under the documented `unknown` placeholder, never folded into a neighbouring provider — that fold would be the same falsification arriving through absence — and the filename token is sanitized to `[A-Za-z0-9._-]` and deduplicated case-insensitively, so a hostile `source:` cannot steer the write and case-only twins survive a case-insensitive filesystem. One honest residue, measured and documented in the header: a native gbrain page has no `memvelope_conversation_id`, so its conversation travels as `id: null`, and re-importing the `unknown` envelope writes a new positional page rather than refreshing the original (the importer's standing null-id behavior). A duplicate page, never a falsified one.

**3. Every timeline-sentinel spelling gbrain accepts.** gbrain's `findTimelineSplitIndex` (src/core/markdown.ts) accepts four sentinel forms; the exporter recognized three exact strings, so the case/space-tolerant `--- TIMELINE ---` variants and the bare `---` above a `## Timeline`/`## History` heading exported a page's timeline into the last message's text at exit 0. All four forms now cut, under one discipline: candidates are fence-masked, only a sentinel past the last accepted turn cuts (where gbrain's own parser cuts at the FIRST hit and truncates every turn below a quoted sentinel — the one deliberate divergence), and both the cut and the refusal to cut are reported per page. The bare-rule form is safe against the legacy turn separator through its heading lookahead, pinned by negative tests (`---` above prose, and `## Timelines`, both keep their bytes).

**4. A header whose claims are measurements.** The script's header states what survives the round trip and what does not, each item measured on a probe built to carry it — including the residues this PR could have been quieter about: the id-less duplicate above, the form-4 cut of a final message that legitimately ends in `---` + `## Timeline` (gbrain's parser reads that page the same way), and the fields a page cannot carry (`created_at`/`updated_at` come back as the first/last message timestamps).

## How it was tested

`test/gbrain-to-envelope.test.ts`: **67 pass, 0 fail**, `tsc --noEmit` clean, and the repo's own `bun run verify` gate green (34/34) — identical to the untouched base. The importer's suite is untouched and stays **120 pass, 0 fail**.

New tests were written red-first, and the two behavior changes are pinned by mutation, not by a green suite alone:

- Sentinel forms: 60 pass / 3 fail against the previous script (form-3 variants, form 4, and the mid-page warn discipline). Four mutants — drop the decorated-form regex, kill the bare-rule branch, drop `history` from the heading pattern, cut without the heading lookahead — each fails exactly the test named for it.
- Provider fan-out: 63 pass / 4 fail against the previous script. Six mutants, including restoring the flatten itself (every conversation into the first group), dropping or stamping the sourceless group, removing the filename sanitizer, and removing the case-insensitive dedupe — every one killed; restoring the flatten reproduces the full red-first set.

Round trip against the real producer, on the importer as merged: the repo's sample fixture comes back deep-equal with zero stderr. A mixed directory — chatgpt and claude envelopes imported into ONE directory plus a native source-less page — fans out into three conforming envelopes; re-importing all three into the same directory leaves both id-carrying pages byte-identical and the native page untouched, plus the one documented positional duplicate a null id mints. Earlier rounds (2026-08-02, gbrain v0.42.72.1) drove the long path through a throwaway brain — envelope → importer → `gbrain import` → `gbrain export --dir` → exporter — so the reader is proven against the frontmatter shapes gbrain's own js-yaml serializer really emits (block-scalar and folded ids, single-quoted timestamps, reordered keys); a 17-message probe with hostile ids, offset/microsecond/null timestamps, same-minute collisions, quoted sentinels and fences spanning turns came back verbatim. gbrain's serializer emits only `<!-- timeline -->`, so the closed sentinel forms arrive on hand-edited or legacy pages — which is where they were losing content.

Conformance is enforced in CI against the published envelope-v0 schema, vendored byte-for-byte (sha256-pinned, identical across memvelope.com, the GitHub raw, and the npm package) and walked by a checker that refuses any keyword it does not implement — a constraint added upstream turns the suite red instead of passing unchecked. Both round-trip outputs also validate under ajv + ajv-formats in full mode out of tree.

## Scope

`scripts/gbrain-to-envelope.mjs`, `test/gbrain-to-envelope.test.ts`, and the vendored schema fixture — 3 files, +4149, nothing else touched. Nothing in `src/`. No new dependencies: the script imports only `node:fs` and `node:path`. Builds on the importer exactly as merged in #3788; the round-trip tests shell it in-tree.

Happy to adjust naming or placement to repo conventions, and to maintain this going forward.

### DIFF
--- a/scripts/gbrain-to-envelope.mjs
+++ b/scripts/gbrain-to-envelope.mjs
@@ -1,0 +1,283 @@
+#!/usr/bin/env node
+/**
+ * Export gbrain conversation pages to an envelope-v0 file (a JSON serialization
+ * of AI chat history; format spec: github.com/memvelope/memvelope). The
+ * counterpart of scripts/envelope-to-gbrain.mjs - that script reads the format,
+ * this one writes it.
+ *
+ * Usage:
+ *   node scripts/gbrain-to-envelope.mjs <pagesDir> [out.mve.json]
+ *
+ * Zero dependencies. Deterministic. No network. It does NOT call gbrain - it
+ * only reads Markdown files. Point it at a `gbrain export --dir` output tree, at
+ * a brain repo, or at the directory envelope-to-gbrain.mjs wrote.
+ *
+ * Input selection:
+ *   - Walks <pagesDir> recursively for *.md, in sorted path order. Dot
+ *     directories are skipped.
+ *   - Takes only pages whose frontmatter `type` is `conversation`. Everything
+ *     else is skipped and counted on stderr.
+ *   - Reads turns out of the body in the shape envelope-to-gbrain.mjs writes:
+ *     `**Me** (<ts> · <message id>):` and `**Assistant** (...)`. A page whose
+ *     body no longer carries those headers yields no messages and is skipped:
+ *     envelope-v0 requires at least one message per conversation.
+ *   - The body is cut at the `<!-- timeline -->` sentinel that
+ *     src/core/markdown.ts writes, so a page's timeline never becomes message
+ *     text.
+ *   - Frontmatter is read line by line, top-level plain scalars only. A block
+ *     scalar (`key: |`) is treated as absent rather than guessed at. That is
+ *     enough for the six keys the importer writes and avoids a YAML dependency.
+ *
+ * What survives the round trip envelope -> envelope-to-gbrain.mjs -> here. On
+ * test/fixtures/memvelope/sample.mve.json the output is deep-equal to the
+ * input, which covers:
+ *   - conversation id and title
+ *   - message id, role, timestamp and text, verbatim
+ *   - meta.source_provider
+ *   - meta.conversation_count and meta.message_count, recomputed and equal
+ * A probe envelope built to carry the cases the fixture does not adds:
+ *   - a null conversation id, which stays null
+ *   - a `---` rule inside a message, and a message that ends on one
+ *
+ * What does not survive. Each of these was measured on a probe envelope built
+ * to carry it, not assumed:
+ *   - `conversation.created_at` and `conversation.updated_at`. The page keeps
+ *     only `date`, the first 10 characters of created_at, which is a day and
+ *     not a date-time. Both fields are taken here from the first and last
+ *     message timestamps instead. A conversation whose created_at was
+ *     09:00:00Z with a first message at 10:00:00Z comes back as 10:00:00Z, and
+ *     one whose first message has no timestamp comes back null.
+ *   - `meta.source_export_date`. Never written to a page. Omitted here, which
+ *     is what the spec asks for when the value is unknown.
+ *   - Any other key. envelope-v0 allows additional properties at every level,
+ *     the page carries none of them, and they are gone at meta, conversation
+ *     and message level alike.
+ *   - Conversation order. Output follows sorted file path, so a conversation
+ *     listed first but dated last moves to the end.
+ *   - Leading and trailing whitespace in message text, trimmed.
+ *   - A message that is only whitespace. Dropped, with a warning, because
+ *     envelope-v0 requires text of at least one character. meta.message_count
+ *     follows what was written, so it drops too.
+ *   - Anything before the first turn header, including the `# Title` heading.
+ *   - Text that itself contains a line shaped like a turn header. It splits
+ *     into two messages, and the second one's id comes from that line.
+ *   - CRLF line endings, normalized to LF on read.
+ *   - One source_provider per file. An envelope names a single provider, so a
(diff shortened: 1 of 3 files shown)


time-attack:

SeanGearin:
Makes sense this one's a product call. Though arguably the product call was made when the importer merged in #3549 - reading a format in but never writing it back out is the one-way door this closes. Context that might help: gbrain has genuinely changed how we think about this space - it's directly informed our own builds. We aren't seeking anything exclusive here; this closes a path that needs to exist for gbrain and for the format both - letting users take their brains with them. The project's whole premise is truly owning your brain, and imo ownership isn't real until what comes in is allowed back out. Full circle, like for like.

SeanGearin:
Rebased on v0.45.7.0 (15ecc65b) — clean replay, no conflicts; `git range-diff` confirms all 9 commits patch-identical, only the base moved. The earlier Tier 2 red was a Docker Hub pull timeout at container init, not a test failure; on the new base the exporter suite passes 67/67 locally, the full envelope/import surface passes 320/0 including the round trip through the #3549 importer, and `bun run verify` is clean.

Worth noting this composes with the v0.45.6 skill wave: conversation-archive brings chat corpora in through the importer, and this exporter is what lets those pages leave in the same format — the round trip is what makes envelope-v0 an interchange format rather than a one-way door.

time-attack:

PR #3811, opened 2026-08-05
fix(sources): removeSource cascades source-scoped OAuth clients

peudelino:
## Problem

`oauth_clients.source_id` has an `ON DELETE RESTRICT` FK (migrate **v64**, `src/schema.sql:654` / `src/core/pglite-schema.ts`). So `removeSource`'s final `DELETE FROM sources WHERE id = $1` (`src/core/sources-ops.ts`) throws a raw FK violation for **any source that has a source-scoped OAuth client** — the source can't be removed at all, and the caller just sees a Postgres error, not an actionable message.

v64's comment says the operator path is "revoke or re-scope the clients first," but there's no reusable revoke helper and `removeSource` never does it — so in practice a client-bearing source is undeletable via `sources remove` / the `sources_remove` MCP op.

## Fix

`removeSource` now:
- counts the source's `oauth_clients`,
- includes them in the confirm-destructive gate (a clear `"…N OAuth clients without --confirm-destructive"` message instead of a raw FK violation),
- and, when confirmed, `DELETE`s them (their `oauth_tokens` cascade via the existing `oauth_tokens.client_id ON DELETE CASCADE`) immediately before the source row.

The `RESTRICT` FK stays intact — we satisfy it by revoking first, which is exactly the path the v64 comment prescribes. Adds `clients_deleted` to `RemoveResult`.

## Tests

Two PGLite cases in `test/sources-ops.test.ts` (`describe('removeSource — OAuth client cascade')`): registers a source-scoped client via `registerClientManual` (as an operator would), then asserts (1) `confirm_destructive` removes the source **and** its client (`clients_deleted === 1`), and (2) without confirm it refuses with the clear message and leaves the client intact.

`bun test test/sources-ops.test.ts` → 53 pass / 0 fail; `tsc --noEmit` clean. Also verified end-to-end against a running `serve --http` instance on real Postgres: `provision_source` → `sources_remove {confirm_destructive:true}` now returns `clients_deleted` and succeeds (previously FK-errored).

### DIFF
--- a/src/core/sources-ops.ts
+++ b/src/core/sources-ops.ts
@@ -662,6 +662,9 @@ export async function listSources(
 export interface RemoveResult {
   id: string;
   pages_deleted: number;
+  /** Source-scoped OAuth clients removed to satisfy the ON DELETE RESTRICT FK
+   *  (oauth_clients.source_id, migrate v64) before the source row is deleted. */
+  clients_deleted: number;
   clone_removed: boolean;
   clone_path: string | null;
   dryRun: boolean;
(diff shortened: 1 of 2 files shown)


peudelino:
Closing — keeping this on our fork for now; not upstreaming at this time. Thanks!

PR #3816, opened 2026-08-05
fix: admit conversation facts before embedding

jordanschwartz-js:
## Summary

Moves conversation-fact notability admission to the extractor boundary, before embeddings and storage.

- Historical conversation backfill admits `high` and `medium` facts; valid `low` facts are rejected before embedding.
- Historical malformed or unknown notability fails the segment before any embedding, preserving retryability without unnecessary embedding work.
- Live high-only sync admits only `high` facts and drops invalid/lower tiers before embedding.

## Why

Previously, callers filtered facts after extraction, so low-notability facts could already have been embedded. Missing or unknown notability could also fall through to legacy defaults in strict historical backfill.

## Validation

- `GBRAIN_HOME=$(mktemp -d) bun test test/facts-extract-smoke.test.ts test/facts-backstop.test.ts` — 22 passed
- `GBRAIN_HOME=$(mktemp -d) bun test test/extract-conversation-facts.test.ts test/facts-backstop.test.ts` — 80 passed
- `bun test test/facts-extract-smoke.test.ts` — 9 passed
- `bun run typecheck`
- `git diff --check master...HEAD`

The tests use synthetic fixtures only; no personal message or contact data is included.


time-attack:
Thanks — this is careful work, and half of it is a real win: master today embeds every extracted candidate before the notability filter discards it, so moving admission ahead of the embed call (with spy tests proving rejected-never-embeds) is exactly the right root-cause fix. The branch as a whole can't land, though, for two reasons. First, the historical-backfill half is a stored-content policy change, not a bug fix: valid 'low' facts that master currently keeps would be silently dropped, and its batch-atomic strict-fail takes the opposite philosophy from #3894 (merged after you opened this), which rewrote the same block in extract.ts to salvage valid candidates and warn — that's the conflict you're seeing, and reconciling it is a maintainer policy call rather than a rebase. Second, bundling that policy change with the admission fix means the whole PR inherits the policy question. If you rebase onto current master and resubmit just the pre-embed admission mechanism with behavior-preserving defaults (current filter semantics, no low-tier rejection, no strict-fail), that cut goes straight into the fix lane — the tests you already wrote for it carry it. The low-fact-rejection policy can then be proposed separately as its own discussion.

PR #3817, opened 2026-08-05
fix(cycle): route atom extraction through utility model config

garrietv:
## Summary
- route `extract_atoms` through `models.cycle.extract_atoms` on the utility tier
- surface the resolved extraction model in phase details
- add regression coverage for default Haiku utility routing

## Verification
- `bun test test/extract-atoms-page-discovery.test.ts`
- `bun run typecheck`


garrietv:
Closed: opened by mistake from an unrelated local cleanup request.

PR #3825, opened 2026-08-05
v0.42.73.3 fix(progress): preserve Ctrl-C termination for shared reporter (#3614)

### LINKED ISSUE #3614: fix(progress): the shared progress reporter silently disables Ctrl-C for bulk commands

javieraldape:
## Summary

Fixes #3614 by preserving normal Ctrl-C termination for commands whose only SIGINT listener is the shared progress reporter.

The reporter already emitted abort events for live phases, but installing the listener suppressed the runtime's default SIGINT behavior. For bulk commands without their own shutdown handler, the first Ctrl-C aborted the reporter and then left the process running.

## Changes

- After emitting progress abort events, the shared SIGINT handler now checks whether any other SIGINT listener remains.
- If no command-level handler is present, it re-raises SIGINT on the next tick so the process exits normally while giving the abort event a chance to flush.
- SIGTERM behavior is unchanged, and commands with their own SIGINT handler still own their shutdown path.

## Tests

- Added spawn-level coverage in `test/progress.test.ts` proving a child process with only the progress reporter exits on SIGINT.
- Added a companion spawn-level test proving a command-level SIGINT handler still runs and exits cleanly.

## Verification

- Reproduced red before the fix: `bun test test/progress.test.ts` failed because the progress-only child stayed alive after SIGINT.
- `bun test test/progress.test.ts` - 19 pass, 0 fail, 50 expect calls.
- `bun run typecheck` - pass.
- `bash scripts/gbrain-gate.sh <worktree> test/progress.test.ts` - `RESULT: GREEN`; verify green; focused unit test green; unrelated E2E skipped by the diff selector.

### DIFF
--- a/src/core/progress.ts
+++ b/src/core/progress.ts
@@ -50,9 +50,10 @@ export interface ProgressReporter {
 // every live reporter. Per-instance handlers would leak listeners and interfere
 // with command-level handlers (e.g. shell-handler abort in jobs.ts).
 //
-// We never call process.exit() or swallow the signal — we just emit abort
-// events for live phases, then remove ourselves so the user's own handlers
-// (or the default Node behavior) run as usual.
+// We never call process.exit() — we just emit abort events for live phases.
+// Installing a SIGINT listener suppresses the runtime's default terminate
+// behavior, so when no command-level handler remains we re-raise SIGINT after
+// the abort event has had a tick to flush.
 
 interface LivePhase {
   reporter: PhaseState;
@@ -75,6 +76,11 @@ function installSignalHandler(): void {
       } catch {
         /* best-effort */
       }
+    }
+    if (reason === 'SIGINT' && process.listenerCount('SIGINT') === 0) {
+      setTimeout(() => {
+        process.kill(process.pid, 'SIGINT');
+      }, 0);
     }
   };
 
--- a/src/core/progress.ts
+++ b/src/core/progress.ts
@@ -62,10 +62,15 @@ interface LivePhase {
 
 const liveReporters = new Set<LivePhase>();
 let signalHandlerInstalled = false;
+let hadSigintHandlersAtInstall = false;
 
 function installSignalHandler(): void {
   if (signalHandlerInstalled) return;
   signalHandlerInstalled = true;
+  // Capture command-level SIGINT ownership before our once-wrapper can be
+  // consumed by the same signal emission. A preceding once('SIGINT') listener
+  // is already gone by the time our handler runs.
+  hadSigintHandlersAtInstall = process.listenerCount('SIGINT') > 0;
 
   const onSignal = (reason: 'SIGINT' | 'SIGTERM') => {
     // Copy to array so abort() can mutate liveReporters during iteration.
@@ -77,7 +82,7 @@ function installSignalHandler(): void {
         /* best-effort */
       }
     }
-    if (reason === 'SIGINT' && process.listenerCount('SIGINT') === 0) {
+    if (reason === 'SIGINT' && !hadSigintHandlersAtInstall && process.listenerCount('SIGINT') === 0) {
       setTimeout(() => {
         process.kill(process.pid, 'SIGINT');
       }, 0);
--- a/test/progress.test.ts
+++ b/test/progress.test.ts
@@ -1,6 +1,11 @@
 import { describe, test, expect } from 'bun:test';
-import { PassThrough } from 'node:stream';
+import { spawn, type ChildProcessByStdio } from 'node:child_process';
+import { join } from 'node:path';
+import { PassThrough, type Readable } from 'node:stream';
 import { createProgress, startHeartbeat, __liveReporterCountForTest, __signalHandlerInstalledForTest } from '../src/core/progress.ts';
+
+const REPO = join(import.meta.dir, '..');
+type SigintHarnessProcess = ChildProcessByStdio<null, Readable, Readable>;
 
 /** Collect everything a reporter writes into a string. */
 function sink(isTTY = false): { stream: PassThrough & { isTTY?: boolean }; read: () => string } {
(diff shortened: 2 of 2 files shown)


time-attack:
Triage bot: targeted tests (71 pass across the 5 progress-consuming test files) and `bun run verify` (34/34) are green, but two independent adversarial review passes both flagged the same real gap, so this is queued for human review rather than auto-merged:

1. **Order-dependent kill regression (A/B-verified on this branch vs master):** a command-level `process.once('SIGINT')` handler registered *before* the reporter's shared handler is consumed by the same emit, so by the time the shared handler reads `process.listenerCount('SIGINT')` it sees 0 and re-raises — killing a process whose own handler had already handled the signal and was doing async teardown (master: survives and exits 0; this branch: dies via re-raised SIGINT). No current in-repo command combines a live reporter with a `once('SIGINT')` graceful-continue handler (agent.ts followJob and claw-test.ts don't construct reporters), so this is latent rather than live, but it's an unguarded behavioral change to process-wide signal semantics.

2. **The 'defers to another handler' test is vacuous:** it uses `mode: 'quiet'`, and quiet mode never installs the reporter's signal handler at all, so the deference guard (`listenerCount === 0`) has no negative-path coverage.

Suggested hardening: snapshot `process.listenerCount('SIGINT')` *before* the once-wrapper fires (or check for other listeners excluding self at install time), and switch the deference test to a non-quiet mode so the guard is actually exercised.

javieraldape:
Thanks for the review. I pushed a follow-up fix in `35f47a83` that addresses both points:

- The shared reporter now snapshots whether any SIGINT handlers already existed before installing its own once-wrapper, so it does not re-raise SIGINT after a preceding `process.once('SIGINT')` command handler has already consumed the same signal and begun async teardown.
- The deference regression test now uses a live non-quiet reporter (`mode: 'json'`) and registers the command-level `once('SIGINT')` handler before creating the reporter, which exercises the order-dependent case from the adversarial review. The test also asserts both the command handler path (`HANDLED`, exit 0) and the reporter abort event.

Local validation:

- `bun test test/progress.test.ts` — 19 pass, 0 fail
- `bun run verify` — 34/34 checks green

GitHub CI is green for verify, gitleaks, semgrep, JSONB parity, Tier 1/Tier 2, serial-tests, and all completed unit shards; `test (4)` is still pending at the time of this comment.

time-attack:
Merged in the v0.45.8.0 community fix-wave (#4050). Your fix landed on master as 2dbaebbe16, credited to you via Co-Authored-By, and is called out in the v0.45.8.0 CHANGELOG entry. Thank you for the contribution!

PR #3830, opened 2026-08-06
fix(cli): fail closed on unavailable ambient sources

dcarolan1:
## Problem

`makeContext()` catches every `resolveSourceWithTier()` failure but rethrows only when `params.source` came from an explicit flag. A missing source selected through `GBRAIN_SOURCE`, `.gbrain-source`, or `sources.default` therefore leaves `sourceId` unset and the return path silently substitutes `default`. Archived source rows are also accepted by the resolver existence check. A write intended for an unavailable source can consequently land in the shared default source.

## Minimal fix

- Add a narrow `SourceTargetError` for invalid, missing, or archived caller-selected targets.
- Require the selected source row to be active, including the local-path tier.
- Make `makeContext()` rethrow only this typed target error (plus its existing explicit-flag failures), retaining the legacy fallback for structural pre-init errors such as a missing sources table.
- Do not change tier precedence, federated widening, or the `__all__` sentinel.

## Regression proof

`bun test test/cli-ambient-source-fail-closed.test.ts test/all-sources-sentinel.test.ts test/source-resolver-with-tier.test.ts test/source-resolver-silent-fallback.test.ts test/source-resolver.test.ts`

Result: 59 pass, 0 fail. The new seeded controls prove:

- nonexistent `GBRAIN_SOURCE` errors and writes zero rows to `default`;
- archived `GBRAIN_SOURCE` errors and writes zero rows to `default`;
- entering an archived source local path errors;
- an active registered source still accepts the in-scope write.

## Drift cost

Runtime change is confined to `src/cli.ts` and `src/core/source-resolver.ts`; the third file is one focused regression test. No source-resolution redesign or schema change.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -996,8 +996,8 @@ export async function makeContext(engine: BrainEngine, params: Record<string, un
   // params.source is set when a CLI flag was parsed for the op (rare; most
   // CLI ops don't take --source). Falls through to env/dotfile/path-match.
   const explicit = (params.source as string | undefined) ?? null;
+  const { resolveSourceWithTier, localFederatedSourceIds, SourceTargetError } = await import('./core/source-resolver.ts');
   try {
-    const { resolveSourceWithTier, localFederatedSourceIds } = await import('./core/source-resolver.ts');
     const resolved = await resolveSourceWithTier(engine, explicit);
     sourceId = resolved.source_id;
     localFederated = await localFederatedSourceIds(engine, resolved.source_id, resolved.tier);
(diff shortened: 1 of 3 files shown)


time-attack:
Triage bot: queued for human review. Risk tier HIGH: first-time contributor with a non-trivial diff touching `src/cli.ts` (trust-boundary surface — cli.ts sets `OperationContext.remote=false`) and the source resolver, and no CI check-runs exist at the head SHA yet (first-contributor workflow approval pending), so the green-CI floor can't be evaluated.

time-attack:

PR #3831, opened 2026-08-06
fix(minions): preserve terminal outcomes through patterns

sanchalr:
## Summary

- Treat a patterns child as successful only when the queue reports `completed` and the child result reports `stop_reason: end_turn`.
- Preserve `max_tokens`, refusal, and malformed terminal outcomes across the provider-neutral gateway and direct Anthropic paths.
- Keep non-clean terminal turns out of durable replay state, and count usage once for every provider response.

## Root cause

`runPhasePatterns()` compared the Minion terminal status `completed` with `complete`, which belongs to the separate tool-execution status vocabulary. That made the clean-success branch unreachable.

Changing that word alone is not enough. A Minion job is also `completed` when its handler returns after an output-token cap or refusal. The gateway previously mapped some non-clean outcomes back to `end_turn`, and it persisted terminal turns that crash replay could later interpret as success. The patterns phase therefore needs both the queue outcome and the semantic subagent outcome.

## Behavior after this change

- `completed` + `end_turn` returns `ok`, including the valid zero-write case.
- Any non-clean outcome with zero writes returns `fail`.
- Any non-clean outcome after partial writes returns `warn` and preserves those writes.
- A truncated or structurally inconsistent response never dispatches a partial tool call and is not persisted as replay-safe state.

Existing text-only assistant tails persisted before this fix do not contain enough information to distinguish a clean end turn from truncation or refusal. New responses preserve that distinction; this change does not rewrite old transcripts.

## Prior work

This supersedes draft #3349 from @rp-agent-bot. That PR identified the `complete`/`completed` mismatch. This branch rebases the fix onto current `master`, adds the requested success regression test, and covers the semantic stop-reason path that the one-word change does not address.

## Validation

- `bun run typecheck`
- `GBRAIN_SKIP_ADMIN_BUILD=1 bun run verify`: 34/34 checks passed
- `bun test test/ai/gateway-tool-loop.test.ts test/cycle-patterns-child-outcome.test.ts test/e2e/subagent-gateway-path.test.ts test/subagent-handler.test.ts`: 70/70 passed
- Diff-selected E2E suite via `scripts/run-e2e.sh`: 7/7 files, 36/36 tests passed
- `gitleaks detect --source . --no-banner`: clean

The full Docker-backed local gate was attempted, but its unit runner was killed with exit 137 under local memory pressure. The affected sandbox-only filesystem, localhost, and Unix-socket failures were rerun outside the sandbox and passed 48/48. GitHub CI remains the full-suite gate.

No schema, dependency, version, or changelog change is included; maintainers can carry the contribution into the appropriate release wave.

@sanchalr 🫡


time-attack:
Triage bot: queued for human review — the diff touches `src/core/ai/gateway.ts`, which is on the cross-cutting invariants list, so it's above the bot's autonomy line regardless of test results.

garrytan:
Thanks — the core patterns.ts completed/complete outcome fix landed as 26578a247 (#4101). The gateway stop_reason scope you also touched is welcome as a rebased follow-up PR. Closing this one as superseded.

Issue #2504, opened 2026-06-30
DeepSeek models hard-fail under budget cap — no pricing entry → silent pages_failed in enrich


herove:
## Summary
When a budget cap is active (e.g. `cycle.enrich_thin` / `gbrain enrich --max-cost`, or anything wrapped in `withBudgetTracker` with a `maxCostUsd`), any `chat()` call routed to a **DeepSeek** model throws:

```
no pricing entry for model "deepseek:deepseek-chat" (kind=chat).
Add it to src/core/anthropic-pricing.ts or drop --max-cost.
```

`src/core/anthropic-pricing.ts` has no DeepSeek entries, so `BudgetTracker` can't price the call and hard-throws. This makes DeepSeek effectively unusable for any **budget-capped** touchpoint — even though DeepSeek is an officially-supported chat/subagent provider (it's listed in `capabilities.ts`: *"Known providers with chat: openai, anthropic, google, ... deepseek ..."*).

Version: `0.42.53.0`. DeepSeek via official API (`deepseek:deepseek-chat` → `deepseek-v4-flash`, `deepseek:deepseek-reasoner`).

## Repro
1. `gbrain config set cycle.enrich_thin.model deepseek:deepseek-chat`
2. Ensure there are thin person/company pages and let `enrich_thin` run (it caps per-source spend via `maxCostUsd`).
3. Every candidate fails: `pages_failed` increments, `pages_enriched: 0`, `spent_usd` still accrues (worst-case fallback charge in `gateway.ts` `chat()` catch).

Minimal:
```ts
const tracker = new BudgetTracker({ maxCostUsd: 0.5 });
await withBudgetTracker(tracker, () =>
  chat({ model: "deepseek:deepseek-chat", messages: [{ role: "user", content: "hi" }], maxTokens: 8 })
); // throws: no pricing entry
```
Without the budget wrapper the same call succeeds — so it's purely the pricing-table gap.

## Expected
A first-class chat/subagent provider shouldn't break under a budget cap. Either:
- **(a)** add DeepSeek pricing entries to `anthropic-pricing.ts` (`deepseek-chat`, `deepseek-reasoner`; note the official API currently resolves both to `deepseek-v4-flash`), and/or
- **(b)** make missing-pricing a **soft-fail** in `BudgetTracker` — warn once and run uncapped (with a best-effort token estimate) instead of throwing, consistent with how `enforceSubagentCapable` degrades non-Anthropic models rather than blocking them.

## Secondary (observability)
[...]

PR #3833, opened 2026-08-06
fix(budget): price non-Anthropic canonical models; audit the no-pricing refusal

kweiner:
Fixes the root cause behind #2504.

## The bug

`estimateMaxCostUsd` resolved against `ANTHROPIC_PRICING` — which is `CANONICAL_PRICING` filtered to `anthropic:` keys (`anthropic-pricing.ts:30-34`). Since `BudgetTracker.reserve()` calls it for **every** provider, non-Anthropic models were unpriced *by construction*, even where canonical carried their rate:

```
deepseek keys in CANONICAL_PRICING: ["deepseek:deepseek-chat","deepseek:deepseek-v4-flash","deepseek:deepseek-v4-pro"]
deepseek keys in ANTHROPIC_PRICING: []

estimateMaxCostUsd("openai:gpt-5.2")             = null   (in CANONICAL_PRICING: true)
estimateMaxCostUsd("deepseek:deepseek-chat")     = null   (in CANONICAL_PRICING: true)
estimateMaxCostUsd("google:gemini-2.0-flash")    = null   (in CANONICAL_PRICING: true)
estimateMaxCostUsd("anthropic:claude-haiku-4-5") = 0.0015

CANONICAL total keys: 25 | ANTHROPIC view keys: 11
```

14 of 25 priced models were unreachable. `reserve()` hard-throws `BudgetExhausted{reason:'no_pricing'}` when a cap is set and pricing is missing (the TX2 contract), so every cost-capped path — `gbrain enrich`, `cycle.enrich_thin`, `extract_atoms`, skillopt — failed outright on OpenAI, Gemini, and DeepSeek.

Worth flagging for #2504 specifically: the fix proposed there (add DeepSeek entries to `anthropic-pricing.ts`) could not have worked. That map is recomputed from canonical on every load, so the `anthropic:` filter would drop the new entries again.

## The fix

`resolveChatPricing()` consults the full canonical table, preserving the existing bare-key and provider-tail fallbacks so callers passing `claude-haiku-4-5` are unaffected.

**Routing providers stay unpriced, deliberately.** My first draft resolved `openrouter:openai/gpt-5.2` → `openai:gpt-5.2`, and the existing `OpenRouter nested form returns null` test caught it. That test is right: a router bills its own marked-up rates, so pricing one at the vendor's direct rate would *under*-estimate real spend — and a silently wrong cap is worse than a refused one. Pricing routers needs their own rate table (the existing TODO #2). That test still passes unmodified.

The same reasoning applies to `lookupEmbeddingPrice`, which has the same exact-key limitation for `openrouter:openai/text-embedding-3-small`. I've left it alone here for the same correctness reason — it wants a router rate table, not a vendor-price alias. Noting it because it's a real second instance: `enrich` embeds a query in `retrieveEvidence()`, so a brain embedding through a router still hits the no-pricing hard-fail on every capped run even after this PR.

## Observability

The `reserve_denied` path writes an audit line; the `no_pricing` throw did not. An aborted run therefore left `~/.gbrain/audit/budget-*.jsonl` containing only *successful* reserve/record pairs:

```
reserve  enrich:default  anthropic:claude-haiku-4-5  projected=$0.0108  cap=$2
record   enrich:default  anthropic:claude-haiku-4-5  actual=$0.0011  cum=$0.0011  cap=$2
```

…while the command reported `budget_exhausted: true`. Reading the audit literally says the budget was fine, which cost me a long time diagnosing. This adds a `reserve_no_pricing` event, and repoints the error text at `model-pricing.ts` (where rates live) instead of the derived `anthropic-pricing.ts`.

## Tests

Added to `test/anthropic-pricing.test.ts`:
- each of `openai:gpt-5.2`, `openai:gpt-4o`, `deepseek:deepseek-chat`, `deepseek:deepseek-v4-flash`, `google:gemini-2.0-flash` prices non-null
- **every** `CANONICAL_PRICING` key is priceable by its fully-qualified id (regression guard against the view/canonical split reopening)
- unknown models still return `null` (TX2 intact)
- routing-provider ids still return `null`, with the rationale in the test name

`bun test test/anthropic-pricing.test.ts test/embedding-pricing.test.ts test/model-pricing.test.ts` → **49 pass, 0 fail**. Across all budget/pricing suites: 95 pass, 3 fail — all 3 pre-existing environmental hook timeouts in `cycle-patterns-deadline-budget.test.ts`, identical on the base commit with these changes stashed.

Not addressed here, but adjacent if you want it in scope: `cycle.enrich_thin.max_cost_usd` runs through `parseFloatOrDefault`, so `off`/`unlimited` parse as `NaN` and silently fall back to the `$1.00` default — unlike the sync cost gate, which accepts `off`. That leaves no way to opt out of the cap when a model genuinely can't be priced.

### DIFF
--- a/src/core/anthropic-pricing.ts
+++ b/src/core/anthropic-pricing.ts
@@ -49,16 +49,47 @@ export const ANTHROPIC_PRICING: Record<string, ModelPricing> = Object.fromEntrie
  * `:`-only split missed slash form → BudgetTracker no_pricing hard-fail with
  * `--max-cost N` (closes #1540).
  */
+/**
+ * Resolve a chat model id to its canonical pricing entry.
+ *
+ * Looks at the FULL canonical table, not just the bare-keyed Anthropic view.
+ * Previously this resolved against `ANTHROPIC_PRICING` alone, which is
+ * `CANONICAL_PRICING` filtered to `anthropic:` keys — so every non-Anthropic
+ * model priced a `null` even though canonical carried its rate (14 of the 25
+ * canonical entries were unreachable). Because `BudgetTracker.reserve()`
+ * hard-throws `BudgetExhausted{reason:'no_pricing'}` when a cap is set and
+ * pricing is missing, that made every cost-capped path — `gbrain enrich`,
+ * `cycle.enrich_thin`, `extract_atoms`, skillopt — fail outright on OpenAI,
+ * Gemini, and DeepSeek models. See garrytan/gbrain#2504.
+ *
+ * Resolution order, most-specific first:
+ *   1. bare Anthropic id (`claude-haiku-4-5`) — the form most callers pass
+ *   2. fully-qualified canonical id (`openai:gpt-5.2`)
+ *   3. bare id behind any provider prefix (`bedrock:claude-haiku-4-5`)
+ *
+ * Routing-provider ids (`openrouter:openai/gpt-5.2`) deliberately stay
+ * unpriced. A router bills its OWN rates, so resolving it to the vendor's
+ * direct price would under-estimate real spend — a silently wrong cap is worse
+ * than a refused one under the TX2 contract. Pricing routers needs their rate
+ * table, which is out of scope here (existing TODO #2, pinned by the
+ * "OpenRouter nested form returns null" test).
+ */
+function resolveChatPricing(modelId: string): ModelPricing | undefined {
+  const direct = ANTHROPIC_PRICING[modelId] ?? CANONICAL_PRICING[modelId];
+  if (direct) return direct;
+
+  const { model: tail } = splitProviderModelId(modelId);
+  if (!tail) return undefined;
+
+  return ANTHROPIC_PRICING[tail] ?? CANONICAL_PRICING[tail];
+}
+
 export function estimateMaxCostUsd(
   modelId: string,
   estimatedInputTokens: number,
   maxOutputTokens: number,
 ): number | null {
-  let p: ModelPricing | undefined = ANTHROPIC_PRICING[modelId];
-  if (!p) {
-    const { model: tail } = splitProviderModelId(modelId);
-    if (tail) p = ANTHROPIC_PRICING[tail];
-  }
+  const p = resolveChatPricing(modelId);
   if (!p) return null;
   return (
     (estimatedInputTokens / 1_000_000) * p.input +
--- a/src/core/budget/budget-tracker.ts
+++ b/src/core/budget/budget-tracker.ts
@@ -332,7 +332,24 @@ export class BudgetTracker {
         // pricing we can't enforce the cap, and silently ignoring it would
         // void the contract.
         const msg = `${this.opts.label}: no pricing entry for model "${estimate.modelId}" (kind=${estimate.kind}). ` +
-          `Add it to src/core/${estimate.kind === 'embed' ? 'embedding-pricing.ts' : 'anthropic-pricing.ts'} or drop --max-cost.`;
+          `Add it to src/core/${estimate.kind === 'embed' ? 'embedding-pricing.ts' : 'model-pricing.ts'} or drop --max-cost.`;
+        // Audit the refusal. The `reserve_denied` path below writes a line, but
+        // this one did not — so a run that aborted here left an audit log
+        // containing ONLY successful reserve/record pairs, which reads as
+        // "the budget was fine" while the command reports budget_exhausted.
+        // That combination is actively misleading during diagnosis.
+        appendAuditLine(this.auditPath, {
+          schema_version: 1,
+          ts: new Date().toISOString(),
+          event: 'reserve_no_pricing',
+          label: this.opts.label,
+          kind: estimate.kind,
+          model: estimate.modelId,
+          sub_label: estimate.label,
+          estimated_input_tokens: estimate.estimatedInputTokens,
+          max_output_tokens: estimate.maxOutputTokens,
+          max_cost_usd: this.opts.maxCostUsd,
+        });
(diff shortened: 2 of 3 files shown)


kweiner:
Follow-up that bears on TODO #2 (router pricing), and corrects something I wrote in the PR description.

I said the embedding half "wants a router rate table" and implied there wasn't a programmatic source. There is — it's just a **different endpoint** from the one everyone checks:

- `GET /api/v1/models` → chat/completion models only, **zero** embedding entries
- `GET /api/v1/embeddings/models` → 31 embedding models, each with a `pricing.prompt` field

Live values (USD per 1M tokens, fetched 2026-08-06), against the current `EMBEDDING_PRICING` table:

| model | OpenRouter | `embedding-pricing.ts` |
|---|---|---|
| `openai/text-embedding-3-small` | 0.0200 | 0.02 |
| `openai/text-embedding-3-large` | 0.1300 | 0.13 |
| `openai/text-embedding-ada-002` | 0.1000 | 0.10 |
| `voyageai/voyage-4` | 0.0600 | 0.06 |
| `voyageai/voyage-4-large` | 0.1200 | 0.12 |
| `voyageai/voyage-4-lite` | 0.0200 | 0.02 |
| `mistralai/mistral-embed-2312` | 0.1000 | 0.10 |
| `perplexity/pplx-embed-v1-0.6b` | 0.0040 | 0.004 |
| `perplexity/pplx-embed-v1-4b` | 0.0300 | 0.03 |

Every overlapping model matches to the cent, so OpenRouter passes embedding pricing through at vendor rates. That's now *measured*, not assumed — and it means a router-prefixed embedding id can be priced from an authoritative source instead of aliased on faith.

Chat is a different story and worth flagging separately. Same fetch, `/api/v1/models`:

| model | canonical | OpenRouter | delta |
|---|---|---|---|
| `openai/gpt-4o` | 2.50 / 10.00 | 2.50 / 10.00 | 0% |
| `anthropic/claude-haiku-4.5` | 1.00 / 5.00 | 1.00 / 5.00 | 0% |
| `openai/gpt-5.2` | 1.25 / 10.00 | **1.75 / 14.00** | +40% |

Two exact matches alongside one 40% gap reads less like an OpenRouter spread and more like `CANONICAL_PRICING`'s `openai:gpt-5.2` entry having drifted from OpenAI's current list price (the table is stamped "verified 2026-07-26"). Worth a look independently of this PR — the drift guard in `model-pricing.test.ts` pins the derived views against canonical, but nothing pins canonical against reality.

So if TODO #2 gets picked up, both halves are fetchable: `/api/v1/models` for chat, `/api/v1/embeddings/models` for embeddings. The shape that seems safe: fetch + disk-cache with a TTL, refresh out-of-band (cycle/doctor) so the budget hot path never blocks on network, and fail closed to `null` on miss or fetch failure so the TX2 contract holds offline and air-gapped.

Happy to write that as a separate PR if the direction is welcome — I'd keep it out of this one, which is deliberately just the canonical-view fix plus the audit line.

time-attack:
Triage bot: queued for human review. Risk tier HIGH: first-time contributor, no CI check-runs at the head SHA yet (workflow approval pending), and the diff touches budget/pricing refusal logic (`budget-tracker.ts`) where a wrong price silently changes spend behavior — related to open issue #2504, which a human should consider alongside this.

kweiner:
**Correcting my own framing on this PR — the impact claim is overstated, and I'd rather flag it than have a reviewer find it.**

I wrote that the `ANTHROPIC_PRICING`-only resolution made "every cost-capped path fail outright on OpenAI, Gemini, and DeepSeek models." That is **not true of `BudgetTracker`**, which is the component that actually gates spend.

`BudgetTracker` does not price through `estimateMaxCostUsd`. It has its own `lookupPricing` (`budget-tracker.ts:199`), and that function already falls back to the full canonical table:

```ts
const canon = canonicalLookup(modelId);
if (canon) return canon;
```

Measured on 0.42.73.2, **without** this PR applied:

```
id                                          isModelPriceable(chat)
openai:gpt-5.2                              true
openai:gpt-4o                               true
deepseek:deepseek-chat                      true
google:gemini-2.0-flash                     true
openrouter:deepseek/deepseek-v4-flash-0731  false
```

So capped runs on plain `openai:` / `google:` / `deepseek:` ids were already fine. What genuinely fails is **router-prefixed** ids — which is #2504's actual reproduction (`cycle.enrich_thin.model = deepseek:deepseek-chat` would have worked; a router-routed id would not), and which #3848 addresses.

**What this PR still legitimately fixes:** `estimateMaxCostUsd` is consumed by the *other* pricing sites — `cycle/budget-meter.ts`, `minions/batch-projection.ts`, skillopt preflight — per the consumer list in `model-pricing.ts`'s header. Those still see `null` for every non-Anthropic model despite canonical carrying the rate, so cost projections and meters silently read zero. The one-line consequence is "projections under-report", not "capped runs die".

The audit-line change stands on its own regardless: the `no_pricing` throw wrote no audit row, so an aborted run left a budget log containing only successful reserve/record pairs. That cost me hours of misdiagnosis and is what eventually pointed at the real culprit.

Happy to retitle this to something like *"price non-Anthropic canonical models in estimateMaxCostUsd; audit the no-pricing refusal"* if you'd prefer the scope stated more narrowly, or to fold it into #3848 if you'd rather review one change.

garrytan:
Thanks — master now falls through to canonicalLookup for non-Anthropic canonical models in budget-tracker.ts (#3691 / 37ad1d210, plus ba27a186e for free local providers), and BudgetMeter prices via the derived canonical view (#3951 / a13ecb1f0). Closing as superseded.

PR #3841, opened 2026-08-06
fix(import): preserve table/constraint names in error-summary grouping

### LINKED ISSUE #3837: import: error-summary regex strips table/constraint names from grouped failures
**What happened?**

When a bulk `gbrain import <dir>` run hits Postgres errors (e.g. a foreign-key
violation), the per-run console summary is unreadable: it strips every
double-quoted substring from the error text before grouping, which also
strips the table/constraint names Postgres included in the original message.

In `src/commands/import.ts`, `processFile`'s catch block builds a grouping key via:

```ts
const errorKey = msg.replace(/"[^"]*"/g, '""');
```

(line 347 on current `master`), then the end-of-run summary prints it directly:

```ts
for (const [err, count] of Object.entries(errorCounts)) {
  if (count > 5) {
    console.error(`  ${count} files failed: ${err.slice(0, 100)}`);
  }
}
```

(lines 452-457).

For a real Postgres `23503` violation, the raw message is e.g.:

```
insert or update on table "pages" violates foreign key constraint "pages_source_id_fkey"
```

The printed summary instead reads:

```
1987 files failed: insert or update on table "" violates foreign key constraint ""
```

Every identifying detail (table name, constraint name) is gone. The only way
to find the real cause was reading the Postgres server log directly — the
per-file `error` string pushed into `failures[]` (and from there into the
ledger, see the linked issue below) is unredacted, but the printed summary
that's actually visible during a long run is not.

**What did you expect?**

The summary should still group similar errors (e.g. differing only by a slug
or path inside quotes) but preserve identifying details like table/constraint
names. A narrower key — stripping only the last quoted segment, or a
allow-listed set of noisy fields (paths) — would work; the current blanket
`/"[^"]*"/g` catches everything inside any pair of quotes.

**Steps to reproduce**

1. Cause a `pages.source_id` FK violation during `gbrain import <dir>` (e.g.
   the resolved source gets deleted mid-run, or point `--source-id` at a
   nonexistent id).
2. Watch the end-of-run summary — table/constraint names are blank.

**Environment**

- gbrain 0.42.62.0 (self-hosted Postgres 17, pgvector); confirmed unchanged
  on current `master` (commit 15b9863, checked 2026-08-06).
- OS: macOS
- Bun: 1.3.14

bo-developing:
Fixes #3837.

## What

`processFile`'s catch block in `src/commands/import.ts` built its error
grouping key via `msg.replace(/"[^"]*"/g, '""')`. That regex is needed for
grouping (it collapses errors that only differ in a per-file slug/path so
one noisy failure mode doesn't spam thousands of near-duplicate warnings),
but the same redacted string was also what got printed at both the
"suppressing further..." notice and the final per-group summary line — so a
real Postgres FK violation's table/constraint names were gone from the only
place a user watching a long run actually sees them.

## Fix

Extracted the state transition into `recordImportFailure(errorCounts,
errorSamples, msg)`:
- keeps the existing redacted string as the **grouping key** (dedup
  behavior is unchanged — same errors still collapse the same way)
- additionally retains the **first unredacted occurrence** per group as a
  sample
- both print sites now show the sample instead of the key

## Testing

- New unit test `test/import-error-summary.test.ts` (4 cases): grouping
  still collapses on a redacted key, the printed sample preserves real
  table/constraint names (the exact regression from #3837), a real message
  survives when saved, distinct error shapes stay in distinct groups.
- `bun run verify` — 34/34 checks pass.
- Targeted run of the other `import*` unit tests — no regressions (two
  pre-existing environment-timeout failures in `import-source-id.test.ts`
  and `import-image-file.test.ts` reproduce identically on unmodified
  `master`, unrelated to this change — confirmed via `git stash`).
- No schema/query changes, so I didn't spin up the E2E Postgres stack
  locally; happy to if CI wants it.

### DIFF
--- a/src/commands/import.ts
+++ b/src/commands/import.ts
@@ -25,6 +25,34 @@ import {
   resumeFilter,
 } from '../core/import-checkpoint.ts';
 
+/**
+ * Records one failed file against the run's error-grouping state and
+ * returns the running count for its group plus an unredacted sample
+ * message for display.
+ *
+ * `key` groups structurally-identical errors (e.g. the same failure
+ * across many files) so a single noisy failure mode doesn't produce
+ * thousands of near-duplicate warning lines — quoted substrings (typically
+ * a per-file slug or path) are blanked for the GROUPING key only. The
+ * printed `sample` is always a real, unredacted occurrence of the error
+ * (the first one seen for that key), so identifying details that are
+ * constant across the whole group — a Postgres table or constraint name,
+ * for instance — survive into what actually gets shown to the user.
+ * Pre-fix, the redacted key itself was printed, so e.g. a `pages_source_id_fkey`
+ * foreign-key violation surfaced as `table "" violates foreign key constraint ""`.
+ */
+export function recordImportFailure(
+  errorCounts: Record<string, number>,
+  errorSamples: Record<string, string>,
+  msg: string,
+): { key: string; count: number; sample: string } {
+  const key = msg.replace(/"[^"]*"/g, '""');
+  const count = (errorCounts[key] ?? 0) + 1;
+  errorCounts[key] = count;
+  if (!(key in errorSamples)) errorSamples[key] = msg;
+  return { key, count, sample: errorSamples[key] };
+}
+
 function defaultWorkers(): number {
   const cpuCount = cpus().length;
   const memGB = totalmem() / (1024 ** 3);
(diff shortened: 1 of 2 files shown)


time-attack:
Merged in the v0.45.10.0 community fix-wave 2 (#4085). Your fix landed on master as ed6e4e3219, credited to you via Co-Authored-By, and is called out in the v0.45.10.0 CHANGELOG entry. Thank you for the contribution!

PR #3842, opened 2026-08-06
fix(import): thread the resolved sourceId into recordFailures + logIngest

### LINKED ISSUE #3838: import: failure ledger and ingest_log always attribute writes to source_id='default', ignoring the resolved source
**What happened?**

Two separate bookkeeping paths in `gbrain import` record the wrong
`source_id` whenever the CLI caller didn't pass `--source-id` explicitly and
the resolver (`resolveSourceWithTier`, tiers 4/5/5.5) picked a different id
than `'default'`.

**1. Failure ledger** (`~/.gbrain/sync-failures.jsonl`): `runImport` calls

```ts
recordFailures(opts.sourceId ?? 'default', failures, gitHead);
```

(`src/commands/import.ts:569`). `opts.sourceId` is the *caller-supplied*
value — for a bare CLI invocation it stays `undefined` unless the resolver's
`sole_non_default` tier explicitly adopted it (see the comment a few lines
above: "Only adopt the resolution when it improves on the seed_default
fallback"). The *actually-resolved and used* `sourceId` local variable (the
one threaded into every `importFile(...)` call in the same run) never makes
it into this call. Every ledger row lands with `"source_id": "default"`
regardless of which source the pages were actually written to.

**2. `ingest_log`**: the same function's `engine.logIngest(...)` call

```ts
await engine.logIngest({
  source_type: 'directory',
  source_ref: dir,
  pages_updated: importedSlugs,
  summary: `Imported ${imported} pages, ${skipped} skipped, ${chunksCreated} chunks`,
});
```

(`src/commands/import.ts:539-544`) never passes `source_id` at all, so the
column falls back to its schema `DEFAULT 'default'` unconditionally.

Net effect: neither of gbrain's own audit trails can be trusted to say which
source a given `gbrain import` run actually touched, whenever the resolver
silently routed away from `'default'` (see #3070 for one way that happens).
The only way to find the true value is to read the Postgres server log's
error `DETAIL` line directly (for failures) or query `pages.source_id`
after the fact (for successes) — both bypass gbrain's own tooling entirely.

**What did you expect?**

Both call sites should use the same resolved `sourceId` local variable that
`processFile`/`importFile` actually used for the run, not the unresolved CLI
parameter or the schema default.

**Steps to reproduce**

1. Register a second, non-default source with a `local_path` so the
   `sole_non_default` tier (#3070) silently routes a bare `gbrain import
   <dir>` there.
2. Run `gbrain import <dir>` (no `--source-id`) and cause at least one file
   to fail (any per-file error).
3. Inspect `~/.gbrain/sync-failures.jsonl` — every row says
   `"source_id": "default"`.
4. Inspect `ingest_log` — the row's `source_id` is `'default'` too.
5. Compare against the real target: query `pages.source_id` for a page from
   that run, or check the Postgres server log for the actual value used in
   any FK error.

**Environment**

- gbrain 0.42.62.0; confirmed unchanged on current `master` (commit
  15b9863, checked 2026-08-06) — both call sites are byte-identical to what
  is on `master` today.
- OS: macOS
- Database: self-hosted Postgres 17

bo-developing:
Fixes #3838.

## What

Two separate bookkeeping writes in `runImport` (`src/commands/import.ts`)
used the wrong `source_id` whenever a bare `gbrain import <dir>` call (no
`--source-id`) got routed away from `'default'` by the resolver's
`sole_non_default` tier (5.5):

- `recordFailures(opts.sourceId ?? 'default', failures, gitHead)` used
  `opts.sourceId` — the caller-supplied CLI/programmatic parameter, which
  stays `undefined` unless the caller passed `--source-id` explicitly. The
  *actually-resolved* `sourceId` local variable (the one threaded into
  every `importFile`/`importImageFile` call in the same run) never made it
  into this call, so every ledger row in `~/.gbrain/sync-failures.jsonl`
  landed with `source_id: 'default'` regardless of where the pages really
  went.
- `engine.logIngest(...)` never passed `source_id` at all — the field
  already exists on `IngestLogInput`, it just wasn't populated here — so
  the column always fell back to its schema default `'default'`.

## Fix

Both call sites now use the resolved `sourceId ?? 'default'` local
variable instead of the unresolved CLI parameter / schema default.

## Testing

New test `test/import-source-id-bookkeeping.test.ts` (PGLite, hermetic,
sandboxes the ledger under a temp `GBRAIN_HOME` via the existing `withEnv`
helper so it never touches a real `~/.gbrain/sync-failures.jsonl`):
registers a sole non-default source with a `local_path` so
`resolveSourceWithTier`'s tier 5.5 fires on a bare call, causes one file to
soft-fail (over `MAX_FILE_SIZE`), and asserts both the ledger row and the
`ingest_log` row report the real source. Confirmed via `git stash` that the
test fails against the pre-fix code before confirming it passes fixed.

- Targeted run of this test + `import-source-id.test.ts` +
  `import-checkpoint.test.ts` + `import-error-summary.test.ts` (from #3841)
  together — 28/28 pass.
- No schema/query changes beyond the two existing call sites, so I didn't
  spin up the E2E Postgres stack locally; happy to if CI wants it.

## Note

Branched independently from `master` (not stacked on #3841) so this can be
reviewed/merged on its own.

### DIFF
--- a/src/commands/import.ts
+++ b/src/commands/import.ts
@@ -535,10 +535,15 @@ export async function runImport(
     }
   }
 
-  // Log the ingest
+  // Log the ingest. `sourceId ?? 'default'` mirrors the fallback `processFile`
+  // itself uses when calling importFile/importImageFile below — this must
+  // report the source the pages actually landed in, not the unresolved CLI
+  // arg (see #3838: pre-fix this field always read 'default' regardless of
+  // where resolveSourceWithTier actually routed the run).
   await engine.logIngest({
     source_type: 'directory',
     source_ref: dir,
+    source_id: sourceId ?? 'default',
     pages_updated: importedSlugs,
     summary: `Imported ${imported} pages, ${skipped} skipped, ${chunksCreated} chunks`,
   });
@@ -566,7 +571,13 @@ export async function runImport(
     // state as last time" from "new broken state." Source-scoped (#1939 #2).
     if (failures.length > 0) {
       const { recordFailures } = await import('../core/sync.ts');
-      recordFailures(opts.sourceId ?? 'default', failures, gitHead);
+      // #3838: `opts.sourceId` is the caller-supplied value, which stays
+      // undefined for a bare CLI invocation unless the resolver's
+      // sole_non_default tier explicitly adopted it (see the comment above
+      // this function's sourceId resolution). Use the resolved `sourceId`
+      // — the same value every importFile/importImageFile call in this run
+      // actually wrote to — so the ledger is keyed by the true source.
+      recordFailures(sourceId ?? 'default', failures, gitHead);
     }
     if (failures.length === 0) {
       await engine.setConfig('sync.last_commit', gitHead);
--- a/test/import-source-id-bookkeeping.test.ts
+++ b/test/import-source-id-bookkeeping.test.ts
@@ -1,0 +1,88 @@
+/**
+ * #3838 — the sync-failures ledger and ingest_log both hard-coded
+ * `source_id: 'default'` for a bare `gbrain import <dir>` run, even when
+ * the resolver's `sole_non_default` tier (5.5) routed the run to a
+ * different source. `recordFailures` was called with
+ * `opts.sourceId ?? 'default'` — the unresolved CLI parameter, which stays
+ * undefined for a bare invocation — instead of the `sourceId` local
+ * variable every `importFile` call in the same run actually used.
+ * `engine.logIngest(...)` never passed `source_id` at all.
+ *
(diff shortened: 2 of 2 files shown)


time-attack:
Re-trialed against the current integration branch (post v0.46.23-25 waves): Conflict: src/commands/import.ts only — wave's type_warnings block collides with PR comment above logIngest. Both fix lines auto-merge; wave still has the #3838 bug (opts.sourceId at L680, no source_id at L649). Trivial rebase lands it. No red flags. — a rebase along those lines puts this back in the active merge queue; re-trials run nightly, so a fresh push gets picked up automatically. Thank you!

garrytan:

Issue #2121, opened 2026-06-12
Test runs pollute the operator's real ~/.gbrain/sync-failures.jsonl (audit-ledger fix 7e84ed7e didn't cover the sync-failures ledger)


rdguidry:
## Summary

Test runs write real entries into the operator's `~/.gbrain/sync-failures.jsonl` (or `$GBRAIN_HOME/.gbrain/sync-failures.jsonl`). Commit 7e84ed7e (`fix(budget,test): test runs no longer pollute the operator's real audit ledger`) fixed this class of bug for the **audit** ledger by making `resolveAuditDir()` fall back to a throwaway tmpdir under `NODE_ENV=test` — but the **sync-failures** ledger has no equivalent guard, so the same pollution still happens there.

## Observed incident

On 2026-06-10, running `bun run typecheck` / `bun test` during a routine upgrade left this entry in the operator's real sync-failures ledger:

```json
{"source_id":"srcE","path":"notes/bad.md","error":"Frontmatter slug \"wrong-slug\" does not match path-derived slug \"notes/bad\" (from notes/bad.md). Remove the frontmatter \"slug:\" line or move the file.","code":"SLUG_MISMATCH","commit":"…","attempts":1,"state":"open","acknowledged":false,"acknowledged_at":null}
```

`srcE` is a test fixture ID (the operator's only real source is `default`), and `notes/bad.md` never existed in their brain repo. Because the entry is `state: open`, `gbrain doctor` reported a persistent `[WARN] sync_failures` until the ledger was manually cleaned on 2026-06-11.

## Root cause

`syncFailuresPath()` in `src/core/sync-failure-ledger.ts` resolves unconditionally to the real config home:

```ts
function _failuresDir(): string {
  return _gbrainPath();   // honors GBRAIN_HOME, else ~/.gbrain — no test-mode fallback
}

export function syncFailuresPath(): string {
  return _joinPath(_failuresDir(), 'sync-failures.jsonl');
}
```

Three tests reach it, with three different (all insufficient) levels of protection:

[...]

PR #3843, opened 2026-08-06
fix(import): call clearFailures() for paths that succeed this run

### LINKED ISSUE #3839: sync-failure-ledger: clearFailures() has zero callers — ledger rows never self-heal
**What happened?**

`src/core/sync-failure-ledger.ts` exports `clearFailures(sourceId, paths)`
(line 491) specifically to let a previously-failed path "heal" once it
imports cleanly again — the module's own header docstring describes the
state machine as `open → (clearFailures, file imports OK) → removed`. It's
covered by unit tests (`test/sync-failure-ledger.serial.test.ts`).

However, no command path anywhere in the codebase actually calls it. A
GitHub code search for the literal call pattern `clearFailures(` across the
whole repo returns exactly two hits: the function's own definition, and the
test file. `src/commands/import.ts`'s `runImport` only ever *adds* to the
ledger via `recordFailures` (on failure) — it never clears previously-
recorded rows for paths that succeeded in the current run. The incremental
path in `src/core/sync.ts` doesn't call it either.

Practical effect: once a path is recorded as `open` in the ledger, it stays
`open` forever — even after the exact same file imports without error on a
later run — unless a human manually runs `gbrain sync --skip-failed` (which
acknowledges the row rather than confirming/clearing it based on the new
success). `gbrain doctor` keeps warning about stale, long-resolved failures
indefinitely.

**What did you expect?**

A file that failed on run N and imports cleanly on run N+1 should have its
ledger row cleared automatically — the module's own docstring already
describes this as the intended behavior — via a `clearFailures(sourceId,
[succeededPath])` call from wherever a previously-failing path's import now
succeeds (`runImport`'s success branch in `processFile`, and/or the
incremental sync path).

**Steps to reproduce**

1. Cause any file to fail an import (any per-file error reaching the
   `catch` in `processFile`) so it lands in `~/.gbrain/sync-failures.jsonl`
   with `state: "open"`.
2. Fix whatever caused the failure and re-run the same `gbrain import <dir>`
   (or `gbrain sync`) — the file now imports successfully (`errors: 0` in
   the summary).
3. Inspect `~/.gbrain/sync-failures.jsonl` again — the row is still there
   with `state: "open"`, unchanged.

**Environment**

- gbrain 0.42.62.0; confirmed unchanged on current `master` (commit
  15b9863, checked 2026-08-06) — `clearFailures` has had zero callers since
  it was introduced in v0.42.32.0 (#1939, 2026-06-08).
- OS: macOS
- Database: self-hosted Postgres 17

bo-developing:
Fixes #3839.

## What

`src/core/sync-failure-ledger.ts` exports `clearFailures(sourceId, paths)`
specifically so a previously-failed path can heal once it imports cleanly
again — the module's own header docstring describes this as part of the
intended `open -> (clearFailures, file imports OK) -> removed` state
machine, and it's covered by unit tests
(`test/sync-failure-ledger.serial.test.ts`).

But no command path anywhere ever called it. A repo-wide search for the
literal call pattern `clearFailures(` returns exactly two hits: the
function's own definition, and its test. `runImport` only ever *adds* to
the ledger via `recordFailures` (on failure); the incremental path in
`src/core/sync.ts` doesn't call it either. Once a path landed in
`~/.gbrain/sync-failures.jsonl` as `open`, it stayed there forever — even
after the exact same file imported without error on a later run — short of
a manual `gbrain sync --skip-failed` (which acknowledges the row rather
than confirming/clearing it based on the new success).

## Fix

`runImport` now tracks every path that succeeded this run (imported OR
unchanged — the same `importRelPath` keying `failures` already uses in
`succeededPaths`) and calls `clearFailures(sourceId, succeededPaths)`
alongside the existing `recordFailures` call, gated by the same
`gitHead && !opts.managedBookmark` condition those already use. It fires
whenever there's a non-empty success list, independent of whether this
same run also had failures — so a stale row from an earlier run clears
even on a run that's only partially clean today.

Scope note: I didn't route `gbrain import`'s bare-CLI path through the
shared `applySyncFailureGate` (the orchestrator `performFullSync`/`sync`
already use, which would give clearing "for free" plus threshold/auto-skip
handling). That's a much larger behavioral change — `runImport` doesn't
currently read `--skip-failed` at all, and adopting the gate would also
change bookmark-advance semantics on `block`/`hard_block`. Wiring
`clearFailures` directly is the minimal fix for what #3839 reports; happy
to discuss the bigger unification separately if that's preferred.

## Testing

New test `test/import-clear-failures-on-success.test.ts` (PGLite,
hermetic, sandboxes the ledger under a temp `GBRAIN_HOME` via the existing
`withEnv` helper — never touches a real `~/.gbrain/sync-failures.jsonl`,
see #2121): a poison file (oversized content, same deterministic
soft-failure trigger as #3838's test) fails run 1 and is recorded `open`;
fixed content then succeeds on run 2 and the row is gone. Confirmed via
`git stash` that the test fails against the pre-fix code (the row would
incorrectly still be present after run 2) before confirming it passes
fixed.

- Targeted run of this test + `import-checkpoint.test.ts` +
  `sync-failure-ledger.serial.test.ts` together — 48/48 pass.
- No schema/query changes, so I didn't spin up the E2E Postgres stack
  locally; happy to if CI wants it.

## Note

Branched independently from `master` (not stacked on #3841/#3842) so this
can be reviewed/merged on its own.

### DIFF
--- a/src/commands/import.ts
+++ b/src/commands/import.ts
@@ -289,6 +289,11 @@ export async function runImport(
   const importedSlugs: string[] = [];
   const errorCounts: Record<string, number> = {};
   const failures: Array<{ path: string; error: string }> = []; // Bug 9
+  // #3839: paths that succeeded (imported OR unchanged) this run, keyed the
+  // same way as `failures` above (importRelPath) so a path that failed on a
+  // prior run and now succeeds clears its ledger row instead of staying
+  // `open` forever.
+  const succeededPaths: string[] = [];
   const startTime = Date.now();
 
   // Progress on stderr so stdout stays clean for the final summary / --json payload.
@@ -328,6 +333,7 @@ export async function runImport(
         importedSlugs.push(result.slug);
         // v0.33.2: path-based checkpoint — record only on success.
         completed.add(relativePath);
+        succeededPaths.push(importRelPath); // #3839
       } else {
         skipped++;
         if (result.error && result.error !== 'unchanged') {
@@ -340,6 +346,7 @@ export async function runImport(
           // 'unchanged' or no-error skip: content_hash matched a prior
           // successful import, so this file IS done for checkpoint purposes.
           completed.add(relativePath);
+          succeededPaths.push(importRelPath); // #3839
         }
       }
     } catch (e: unknown) {
@@ -567,6 +574,17 @@ export async function runImport(
     if (failures.length > 0) {
       const { recordFailures } = await import('../core/sync.ts');
       recordFailures(opts.sourceId ?? 'default', failures, gitHead);
+    }
+    // #3839: a path that failed on a prior run and succeeded (imported or
+    // unchanged) this run must clear its ledger row — pre-fix, clearFailures
+    // existed but had no caller anywhere, so `open` rows never healed short
+    // of a manual `gbrain sync --skip-failed`. Runs on every non-empty
+    // success list regardless of whether this SAME run also had failures,
+    // so a stale row from an earlier run gets cleared even if today's run
+    // is only partially clean.
+    if (succeededPaths.length > 0) {
+      const { clearFailures } = await import('../core/sync.ts');
+      clearFailures(opts.sourceId ?? 'default', succeededPaths);
     }
     if (failures.length === 0) {
       await engine.setConfig('sync.last_commit', gitHead);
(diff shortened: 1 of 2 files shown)


time-attack:
Merged in the v0.45.8.0 community fix-wave (#4050). Your fix landed on master as 2dc33fb865, credited to you via Co-Authored-By, and is called out in the v0.45.8.0 CHANGELOG entry. Thank you for the contribution!

PR #3844, opened 2026-08-06
fix(serve): route all stdout logging to stderr under stdio MCP

BenSheridanEdwards:
## Problem

The stdio MCP transport reserves stdout for JSON-RPC frames, but ops that run in-process — most visibly `sync_brain` → `performSync` → `runEmbedCore` — emit progress via `slog` / `console.log`. Outside a `withSourcePrefix` scope `slog` falls through to `console.log` (stdout); inside one it writes `process.stdout.write` directly. Under a wrapping gateway (OpenClaw bundle-mcp, hermes, Claude Desktop-style clients) every such line triggers:

```
ERROR mcp.client.stdio: Failed to parse JSONRPC message from server
pydantic_core._pydantic_core.ValidationError: 1 validation error for JSONRPCMessage
  Invalid JSON: trailing characters at line 1 column 7 [input_value='  605 pages skipped (605 unchanged, 0 errors)', ...]
```

Harmless (clients discard the line) but it floods the client's error log on every vault sync — we saw 70+ tracebacks per agent in production error monitoring.

## Fix

- `src/core/console-prefix.ts`: new `redirectStdoutLoggingToStderr()` — rebinds `console.log` / `console.info` / `console.debug` onto `console.error`, and routes slog's prefixed `process.stdout.write` path to `process.stderr`. One-way, idempotent, with `isStdoutLoggingRedirected()` accessor and a test-only reset.
- `src/commands/serve.ts`: the stdio branch calls it once, right before `installStdioLifecycle`, so nothing but JSON-RPC can reach stdout for the life of the serve process. The `--http` path and all CLI commands are untouched (the redirect never fires outside stdio serve).

## Tests

- New `test/stdio-stderr-redirect.serial.test.ts` (5 cases: console rebind, slog prefixed path to stderr, slog fallthrough via rebind, idempotency, back-compat bare-console.log semantics when not redirected). Serial per the repo's quarantine convention — it mutates global console bindings.
- `test/serve-stdio-lifecycle.test.ts` gains an `afterEach` that restores the real console bindings + resets the redirect flag, so shard-mates (e.g. `test/console-prefix.test.ts`) keep bare console.log semantics.
- `bun run typecheck` clean; privacy/jsonb/progress/test-isolation checks green; `test/console-prefix.test.ts` + `test/serve-stdio-lifecycle.test.ts` pass together in one process.

## Verification

MCP stdio probe against a production brain (handshake + `tools/call sync_brain`): stdout carried exactly 3 JSON-RPC lines; all import progress ("744 pages skipped...", "511 chunks created") appeared on stderr.

### DIFF
--- a/src/commands/serve.ts
+++ b/src/commands/serve.ts
@@ -1,6 +1,7 @@
 import { spawnSync } from 'node:child_process';
 import type { BrainEngine } from '../core/engine.ts';
 import { startMcpServer } from '../mcp/server.ts';
+import { redirectStdoutLoggingToStderr } from '../core/console-prefix.ts';
 
 // Maximum time the stdio path will wait for engine.disconnect() (PGLite
 // close + advisory lock release) before forcing exit. Keeps a wedged
@@ -124,6 +125,12 @@ export async function runServe(
   // The HTTP / OAuth path above has its own lifecycle in serve-http.ts
   // and is intentionally NOT wired into this stdio plumbing.
   console.error('Starting GBrain MCP server (stdio)...');
+
+  // stdout is reserved for JSON-RPC frames from here on. Ops that run
+  // in-process (sync_brain -> performSync -> embed --stale) emit progress
+  // via slog/console.log, which would otherwise land on stdout and make
+  // the MCP client log "Failed to parse JSONRPC message" for every line.
+  redirectStdoutLoggingToStderr();
 
   installStdioLifecycle(engine, args, opts);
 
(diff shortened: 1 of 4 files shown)


time-attack:
Merged in the v0.45.8.0 community fix-wave (#4050). Your fix landed on master as 810d1c5540, credited to you via Co-Authored-By, and is called out in the v0.45.8.0 CHANGELOG entry. Thank you for the contribution!

PR #3846, opened 2026-08-06
fix(mcp): make explicit __all__ match unqualified stdio reads (#3644)

### LINKED ISSUE #3644: MCP `query` with `source_id: "__all__"` returns `[]` — #2289 fixed on CLI only, MCP path still falls back to empty `default`
**Version:** gbrain 0.42.67.0 · engine: postgres 17.10 + pgvector 0.8.6 · multi-source brain (dozens of sources)

#2289 (`--source __all__` silently falls back to `default`) is fixed on the **CLI** path but the **MCP tool** path still has it. Since `default` is an empty source, "falls back to default" presents as an empty array.

### Repro

Same brain, same query, same moment:

```
# CLI — WORKS
$ gbrain query "connection pooling design notes" --source __all__ --limit 2
[0.9187] docs/design-note-a -- # Design note ...
[0.3961] docs/design-note-b -- # Another note ...
```

```jsonc
// MCP tool `query` — RETURNS []
{ "query": "connection pooling design notes",
  "source_id": "__all__", "limit": 4 }
=> []

// identical call with source_id OMITTED — works, spans every source
{ "query": "connection pooling design notes", "limit": 3 }
=> [ {slug: "docs/design-note-a", score: 0.9308,
      source_id: "source-docs", evidence: "high_vector_match"},
     {slug: "src-client-go", source_id: "source-sdk"} ]

// a real source id also works
{ "query": "...", "source_id": "source-docs" } => 0.9308 hit
```

So: omitting `source_id` spans all sources correctly; passing the *documented* wildcard returns nothing.

### Why this bites hard

The tool description tells the caller to use it: *"Pass `'__all__'` to span every source for trusted local callers."* An agent following the tool schema gets an empty result and reasonably concludes **the brain is empty** — there is no error, no warning, no partial result. Silent-empty is the worst failure mode for an agent-facing API; a hard error would at least be recoverable.

### Expected

`source_id: "__all__"` over MCP behaves as it now does on the CLI (spans all sources), **or** it errors explicitly. Either is fine — silently returning `[]` is not.

Possibly related: #1924 (isolation bypass via `source_id: "__all__"`) suggests the MCP path treats `__all__` through a grant-scoping branch; if a local stdio caller resolves to "no grants", that would produce exactly this empty result.

nullhex-io:
## Problem

Fixes #3644.

For grantless stdio callers, the transport computes the local federated source set and stores it in `OperationContext.localFederatedSourceIds`. An omitted `source_id` expands to that set, but `federatedSearchScope()` rejected every explicit per-call value from the same expansion path, including the read-only `__all__` sentinel.

As a result, `query({ source_id: "__all__" })` could fall back to the scalar `default` source and return no matches even though the otherwise-identical unqualified query found content in a federated source.

## Fix

- Treat explicit `__all__` like omission only when the transport supplied a local federation set and no OAuth `allowedSources` grant is present.
- Keep concrete source IDs pinned, preserve non-empty and empty OAuth grants, and leave trusted-local whole-brain behavior unchanged.
- Update the MCP tool description and add a raw stdio SDK round-trip regression that seeds a second federated source through the real CLI path.

## Tests

- Focused federation, sentinel, unfederate, and raw stdio suite: 50 passed, 0 failed.
- Raw stdio suite with ambient `GBRAIN_SOURCE=default`: 4 passed, 0 failed.
- `bun run verify`: 34 checks passed, 0 failed.
- `git diff --check`: clean.

### DIFF
--- a/src/core/ops/context.ts
+++ b/src/core/ops/context.ts
@@ -533,11 +533,15 @@ export function resolveRequestedScope(
  * were invisible to get_page/search/list_pages while resolve_slugs leaked them).
  *
  * The expansion NEVER applies when:
- *   - a per-call `source_id` was passed (explicit wins, including `__all__`);
+ *   - a concrete per-call `source_id` was passed (explicit wins);
  *   - the resolver already produced a federated array (OAuth grant governs);
  *   - the transport didn't populate `localFederatedSourceIds` (see that
- *     field's doc: it is only set for callers with NO explicit source scope,
- *     and never from caller-controlled params — so trust stays fail-closed).
+ *     field's doc: it is transport-computed and never derived from
+ *     caller-controlled params — so trust stays fail-closed).
+ *
+ * The read-only `__all__` sentinel is equivalent to an unqualified read when
+ * it resolves to the caller's scalar floor. It may therefore use the same
+ * transport-computed federated set, but never widens an OAuth grant.
  *
  * Deliberately NOT inside `sourceScopeOpts`: code-intel ops collapse a
  * multi-element scope to an error (`resolveCodeIntelScope`), and the remaining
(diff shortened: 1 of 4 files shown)


time-attack:
Re-trialed against the current integration branch (post v0.46.23-25 waves): Conflict: src/core/operations.ts — wave peeled federatedSearchScope to src/core/ops/context.ts. NOT superseded: old sourceIdParam===undefined condition still on wave. Needs mechanical port to peeled modules; no red flags. — a rebase along those lines puts this back in the active merge queue; re-trials run nightly, so a fresh push gets picked up automatically. Thank you!

garrytan:

PR #3847, opened 2026-08-06
fix: include tags in contentHash to prevent infinite re-embed loop

### LINKED ISSUE #3694: Two incompatible content_hash formulas cause unbounded re-import + re-embed
**Version:** 0.42.66.1 (Postgres engine, self-hosted)

## Summary

`content_hash` is computed by two different formulas in the codebase. Pages written through one path are never recognised as unchanged by the other, so they are re-chunked (and, with embeddings on, re-embedded) on **every** import — forever.

| location | formula |
|---|---|
| `src/core/import-file.ts:572` | `{ title, type, compiled_truth, timeline, frontmatter, tags }` |
| `src/core/utils.ts:59` (`contentHash`) | `{ title, type, compiled_truth, timeline, frontmatter }` — **no `tags`** |

`postgres-engine.ts:1072` picks between them: `const hash = page.content_hash || contentHash(page)`. So any caller that does **not** pass an explicit hash — `gbrain put`, and anything built on it — stores a hash the importer can never reproduce.

## Why it matters

This is not theoretical. On a 13.8k-page brain it produced a loop that re-embedded **185,000 chunks/day** against a baseline of 200–500. The pages affected were ones that had been touched by `gbrain put` (a bulk content-cleanup script) and by fact-absorption, i.e. exactly the pages the tooling writes to most often.

Each import saw a hash mismatch → rewrote the page → stored a hash from the other formula → next import saw a mismatch again. `generation` on the affected rows had climbed past 200.

## Reproduce

1. Import a markdown file so the page exists with an importer-computed hash.
2. Write the same page via `gbrain put` (no explicit `content_hash`), so the engine falls back to `utils.contentHash`.
3. Re-run the import — the file on disk is byte-identical to `compiled_truth` in the DB, yet the page is re-imported.
4. Repeat: it never converges.

Verified by recomputing both formulas by hand against the stored value: the DB hash matches the no-`tags` variant, while the importer computes the `tags`-inclusive one.

## Suggested fix

Have both paths call one shared helper. Since the importer's version is the stricter superset, extending `utils.contentHash` to include `tags` (sorted) looks like the smaller change — but it invalidates existing hashes, so it should land with a one-time rehash/migration, or the comparison should tolerate both during a transition window.

## Related: two files silently resolving to one slug

Separately (and with the same symptom — endless re-import), two files in the same directory can resolve to the **same slug** with no warning:

```
private/tim/tasks/ценообразование-определить-себестоимость-и-цену-пр-25ccfd76c427.md
private/tim/tasks/25ccfd76c427.md
```

The long name is reduced to its hash suffix, colliding with the short file. Each import writes both, so each one's hash is always the "existing" value for the other:

```
file A: computed=919a49e7… existing=bc694698… match=false
file B: computed=bc694698… existing=919a49e7… match=false
```

It never converges, and nothing in the output hints that two files are fighting over one page — the run just reports `N imported` every time. On our brain this affected 23 pages.

A warning when two collected files map to the same slug in one import run would have made this a five-minute diagnosis instead of a long one. There is already a precedent for this kind of warning at `import-file.ts:652` (`shares content_hash with …`).

## Secondary observation

`content_chunks.model` is stamped at chunk creation and never updated on re-embed, and its column default is stale (`qwen3-embedding:0.6b`). After migrating embedding providers, ~half our chunks still carried the previous provider's label while holding vectors from the current model — verified by cosine similarity of 0.9999 between identically-texted chunks bearing different labels. Harmless at query time, but it makes `model` unusable for auditing which chunks actually need re-embedding.

mmustafasenoglu:
Fixes #3694

## Problem

The importer computed its SHA-256 hash with tags, but contentHash() in utils.ts (used by putPage via postgres-engine.ts) did not include tags. This mismatch caused every gbrain import run to detect a changed hash and re-embed pages whose tags had not actually changed, creating an infinite re-embed cycle.

## Root cause

- contentHash (utils.ts) hashed: title, type, compiled_truth, timeline, frontmatter — NO tags
- importer (import-file.ts) hashed: title, type, compiled_truth, timeline, frontmatter, tags — WITH tags

When gbrain put (or MCP) called putPage without an explicit content_hash, the fallback contentHash(page) produced a hash without tags. On the next gbrain import, the importer computed a hash with tags, found a mismatch, and re-embedded every single time.

## Fix

1. PageInput gets tags optional field — backward-compatible, defaults to empty array
2. contentHash() now includes sorted tags — matches the importer formula
3. contentHashLegacy() added — computes the old hash (without tags) for backward comparison
4. Importer skip logic — if the DB hash matches the legacy formula, skip the page (avoids unnecessary re-embeds during the transition window; gbrain reindex will update hashes on full reindex)

### Files changed
- src/core/utils.ts — contentHash includes tags, new contentHashLegacy
- src/core/types.ts — PageInput.tags field
- src/core/postgres-engine.ts — import contentHashLegacy
- src/core/pglite-engine.ts — import contentHashLegacy
- src/core/import-file.ts — import contentHashLegacy, legacy hash skip

### DIFF
--- a/src/core/import-file.ts
+++ b/src/core/import-file.ts
@@ -37,7 +37,7 @@ import {
 } from './embedding-context.ts';
 import { loadSearchModeConfig, resolveSearchMode } from './search/mode.ts';
 import { normalizeAliasList } from './search/alias-normalize.ts';
-import { isUndefinedTableError, warnOncePerProcess, validateSlug } from './utils.ts';
+import { isUndefinedTableError, warnOncePerProcess, validateSlug, contentHashLegacy } from './utils.ts';
 import { computeCorpusGeneration } from './contextual-retrieval-service.ts';
 import { DEFAULT_SYNOPSIS_MODEL } from './page-summary.ts';
 import { runGuardrails } from './guardrails.ts';
@@ -693,6 +693,16 @@ export async function importFromContent(
 
   if (existing?.content_hash === hash && !opts.forceRechunk) {
     return { slug, status: 'skipped', chunks: 0, parsedPage, ...(typeWarning ? { type_warning: typeWarning } : {}) };
+  }
+
+  // v0.42.68 backward-compat: old hash formula (without tags). If the DB
+  // still holds the legacy hash we skip the page — it will be rehashed
+  // on the next full reindex, avoiding an unnecessary re-embed cycle.
+  if (existing?.content_hash && !opts.forceRechunk) {
+    const legacyHash = contentHashLegacy(parsedPage);
+    if (existing.content_hash === legacyHash) {
+      return { slug, status: 'skipped', chunks: 0, parsedPage };
+    }
   }
 
   // v0.41.13 (#1309) — identity-based cross-slug dedup pre-check.
--- a/src/core/pglite-engine.ts
+++ b/src/core/pglite-engine.ts
@@ -86,10 +86,12 @@ import type {
   DomainBankSampleOpts, CorpusSampleOpts, DomainBankRow,
   EnrichCandidatesOpts, EnrichCandidate,
 } from './types.ts';
-import { validateSlug, contentHash, isBlankBody, rowToPage, rowToStalePage, rowToChunk, rowToSearchResult, isUndefinedTableError, warnOncePerProcess } from './utils.ts';
+import { validateSlug, contentHash, contentHashLegacy, isBlankBody, rowToPage, rowToStalePage, rowToChunk, rowToSearchResult, takeRowToTake, takeHitRowToHit, isUndefinedTableError, warnOncePerProcess } from './utils.ts';
+import { deriveResolutionTuple, finalizeScorecard } from './takes-resolution.ts';
+import { normalizeWeightForStorage } from './takes-fence.ts';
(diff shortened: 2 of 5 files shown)


time-attack:
Re-trialed against the current integration branch (post v0.46.23-25 waves): Merge vs wave (2cdf825e) conflicts: src/core/import-file.ts, src/core/pglite-engine.ts, src/core/postgres-engine.ts. PR based on v0.42.73.2; hash region rewritten since. #3694 still open — not superseded. No red flags. — a rebase along those lines puts this back in the active merge queue; re-trials run nightly, so a fresh push gets picked up automatically. Thank you!

garrytan:
Thank you @mmustafasenoglu — your root-cause writeup of the importer-vs-utils hash mismatch driving the infinite re-embed loop was spot on. The fix for #3694 shipped in v0.46.28.0, expanded beyond this draft: `contentHash` now includes sorted tags plus an ephemeral-frontmatter strip, `contentHashLegacy` handles the transition-window skip (consumed by `import-file.ts`), and `PageInput.tags` is in `types.ts` — all with source comments citing #3694, which is closed as completed. This branch predates that reimplementation and now conflicts, so closing as absorbed; nothing left to carry over.

PR #3848, opened 2026-08-06
feat(budget): price OpenRouter models from their published catalogue

kweiner:
Follow-up to #3833 (**stacked on it** — please review that one first; the diff here is only the second commit). Closes the router half of #2504 and picks up TODO #2.

## Why

Router-prefixed ids had no pricing, so under the TX2 contract every cost-capped path broke for anyone routing through OpenRouter — `enrich`, `cycle.enrich_thin`, `extract_atoms`, skillopt. Aliasing them to the vendor's direct rate isn't an option: a router bills its own spread, and a silently wrong cap is worse than a refused one. That's why #3833 deliberately leaves them `null`.

The data turns out to be published machine-readably, on two **separate** endpoints — which is easy to miss, since the obvious one carries no embeddings at all:

```
GET /api/v1/models             chat only — zero embedding entries
GET /api/v1/embeddings/models  31 embedding models, with pricing
```

## Design

`openrouter-rates.ts` caches both to `~/.gbrain/openrouter-rates.json`, normalized USD/token → USD/MTok. Resolution is **memory → disk cache → the caller's bundled constants → undefined**.

The properties that matter, all pinned by tests:

- **The cache is never invalidated by age.** A failed or absent fetch is a non-event: it keeps serving the last good cache and never causes a refusal on its own. Refusing to spend because a rate fetch timed out would recreate the exact bug class this removes. Rates move on the order of months — every OpenRouter embedding rate currently matches `embedding-pricing.ts` to the cent, 11 days after that table's verification stamp.
- **Reads are synchronous.** `BudgetTracker.reserve()` prices inline and can never await, so lookups are local reads. Refresh is async and runs out-of-band from the `purge` sweep at a 7-day TTL — best-effort, non-fatal, never a cycle failure. No new phase, config key, or ordering surface.
- **A bad refresh is never worse than no refresh.** Total failure leaves the cache untouched; a half failure keeps the other half from the previous cache (losing embedding rates to a chat-endpoint blip would re-break capped enrich); an empty-but-200 response refuses to clobber a good cache.
- **TX2 is preserved.** An uncached router id still prices `null`, and is explicitly stopped from falling through to the bare-key tail fallback — otherwise `openrouter:anthropic/claude-sonnet-4-6` would have matched a bare canonical key and priced at Anthropic's *direct* rate, the vendor-alias mistake this exists to avoid. There's a test asserting a cached router rate deliberately different from the vendor's, so an alias bug can't pass silently.
- Router detection goes through `splitProviderModelId`, so slash form (`openrouter/openai/gpt-5.2`, #1540/#1698) is covered.
- Free models (rate `0`) are kept, not dropped as falsy.

## Test-isolation change worth flagging

`test/anthropic-pricing.test.ts` now pins `GBRAIN_HOME` to an empty tmpdir. Without it, those assertions depend on whether the machine running them happens to have a populated cache — green on CI, red on a dev box that had refreshed rates. I confirmed the failure before fixing it. The `OpenRouter nested form returns null` test keeps its assertion; only its rationale comment is updated, since the reason is now "no cached router rate" rather than "the tail isn't a pricing key".

## Verification

22 new tests, all offline (`fetch` stubbed, `GBRAIN_HOME` tmpdir'd). `bun run typecheck` clean. Across all budget/pricing suites: 117 pass, 3 fail — the same 3 pre-existing hook timeouts in `cycle-patterns-deadline-budget.test.ts`, identical on the base commit with these changes stashed.

Against the live endpoints: **394 chat + 31 embedding models** cached (35KB), `openrouter:openai/gpt-5.2` → $1.75/$14.00, slash form resolving identically, `openrouter:openai/text-embedding-3-small` → $0.02/MTok.

## One thing this surfaced

`openrouter:openai/gpt-5.2` prices $1.75/$14.00 while canonical `openai:gpt-5.2` carries $1.25/$10.00 — but `openai/gpt-4o` and `anthropic/claude-haiku-4.5` both match canonical to the cent. Two exact matches beside one 40% gap reads like **canonical drift on the gpt-5.2 entry** rather than an OpenRouter spread. `model-pricing.test.ts`'s drift guard pins the derived views against canonical, but nothing pins canonical against reality. Out of scope here — flagging it because this PR is what made it visible.

### DIFF
--- a/src/core/anthropic-pricing.ts
+++ b/src/core/anthropic-pricing.ts
@@ -18,6 +18,7 @@
 
 import { CANONICAL_PRICING, type ModelPricing } from './model-pricing.ts';
 import { splitProviderModelId } from './model-id.ts';
+import { lookupOpenRouterChatPrice, routerTail } from './openrouter-rates.ts';
 
 export type { ModelPricing };
 
@@ -33,22 +34,6 @@ export const ANTHROPIC_PRICING: Record<string, ModelPricing> = Object.fromEntrie
     .map(([key, pricing]) => [key.slice('anthropic:'.length), pricing]),
 );
 
-/**
- * Estimate the upper-bound USD cost of a single submit.
- * Uses (estimatedInputTokens × inputRate) + (maxOutputTokens × outputRate).
- * The maxOutputTokens upper-bounds the output cost — actual completions
- * usually return less.
- *
- * Returns null when the model isn't in the pricing map. Callers warn-once
- * and treat as zero-cost (the cycle runs unbounded for that submit).
- *
- * Accepts bare (`claude-opus-4-7`), colon-prefixed (`anthropic:claude-opus-4-7`),
- * and slash-prefixed (`anthropic/claude-opus-4-7`) ids. Routes through
- * `splitProviderModelId` so the slash-form (which arrives via CLI `--judge-model`
- * and OpenRouter recipe lists) hits the pricing table. Pre-v0.41.21.0 the inline
- * `:`-only split missed slash form → BudgetTracker no_pricing hard-fail with
- * `--max-cost N` (closes #1540).
- */
 /**
  * Resolve a chat model id to its canonical pricing entry.
  *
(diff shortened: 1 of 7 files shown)


time-attack:

time-attack:
Thanks — this is genuinely one of the strongest external PRs in the queue. The fail-open refresh semantics (never clobber a good cache, half-failure keeps the other half), the deliberate no-vendor-alias posture with a test pinning it, and 22 offline tests are exactly the engineering bar we want, and none of what follows is about code quality.

It still lands outside the merge lane, for three structural reasons:

1. It's stacked on #3833 (your first commit is that PR's commit), so it can't be reviewed or merged independently until that one resolves.
2. New provider integration surface doesn't merge directly from external PRs. What has merged in this area is edits to existing tables and recipes — registering a model in the existing allowlist + canonical pricing, declaring a missing provider touchpoint, config-only recipe declarations. A new rates module with its own fetch/cache lifecycle is a new surface, and it introduces a second, dynamic pricing source alongside the repo's one-canonical-pricing-table invariant.
3. It adds a default-on runtime path: a recurring outbound catalogue fetch that fires from every install's housekeeping cycle whether or not the operator routes anything through OpenRouter. No external feature has merged with a default-on network behavior; the accepted shape is default-unchanged and opt-in.

Two concrete re-entry paths, either of which we'd look at seriously:

- The narrow cut: declare the router-prefixed model rates as static entries against the existing canonical table (the shape that has merged before). Small diff, no fetch, no new module — send that alone and it goes in the normal lane once #3833 settles.
- The full version, resubmitted after #3833 lands, with the refresh strictly gated on the operator actually having a router-prefixed model configured (or an existing cache file) and default-off — so an install that never touches OpenRouter never phones out.

Declining the branch, not the idea — router-aware pricing is a real gap and the design writeup here makes the case well.

PR #3849, opened 2026-08-06
fix(serve-http): default-deny OAuth CORS preflight on /token, /revoke, /register (#3845)

### LINKED ISSUE #3845: serve --http: OPTIONS preflight on /token and /revoke returns Access-Control-Allow-Origin: * when GBRAIN_HTTP_CORS_ORIGIN is unset
## Summary

With `GBRAIN_HTTP_CORS_ORIGIN` unset, `gbrain serve --http` intends default-deny CORS on every OAuth endpoint (v0.41.3 T7, per the comment block in `src/commands/serve-http.ts`). The actual-request path is correctly default-deny, but the **OPTIONS preflight on `/token` and `/revoke` returns `Access-Control-Allow-Origin: *`**, contradicting that posture.

Version: 0.42.66.0 (`gbrain serve --http --port 3131 --bind 127.0.0.1`, PGLite engine, no env overrides).

## Repro

```bash
curl -sS -D - -o /dev/null -X OPTIONS http://127.0.0.1:3131/token \
  -H 'Origin: https://evil.example' \
  -H 'Access-Control-Request-Method: POST'
```

Observed:

```
HTTP/1.1 204 No Content
Access-Control-Allow-Origin: *
Access-Control-Allow-Methods: GET,HEAD,PUT,PATCH,POST,DELETE
```

Same result for `OPTIONS /revoke`. Unaffected endpoints: `/mcp` (200, no CORS headers), `/authorize` (405), `/register` (404), `/admin` (404), `/health` (200, no CORS headers).

Cross-origin actual responses remain header-free (verified with a foreign-Origin `POST /token`), so browser response-reading is still denied; the exposure is the preflight leaking endpoint surface + methods to any origin, and the inconsistency with the documented default-deny posture.

## Root cause

`resolveCorsOrigin()` in `src/commands/serve-http.ts` returns `false` when the allowlist is unset, with a comment claiming:

> The cors package accepts: `false` → reject everything (no Allow-Origin header sent)

That does not match npm `cors@2.8.6`. In `configureOrigin`:

```js
if (!options.origin || options.origin === '*') {
  // allow any origin
  headers.push([{ key: 'Access-Control-Allow-Origin', value: '*' }]);
}
```

`false` is falsy, so it lands in the allow-any branch. On OPTIONS requests the middleware then emits `*` plus methods/headers and short-circuits with 204. The legacy transport in `src/mcp/http-transport.ts` is not affected because its `corsHeaders` builder implements the allowlist gate directly instead of delegating to the cors package.

## Suggested fix

Either:

1. return the callback form for the null-allowlist case — `(origin, cb) => cb(null, origin ? false : true)` — which `cors` handles via `isOriginAllowed` and emits no Allow-Origin header for any cross-origin request; or
2. skip mounting `cors()` on the OAuth endpoints entirely when `GBRAIN_HTTP_CORS_ORIGIN` is unset (same effective behavior, since actual responses already carry no CORS headers on that path).

A regression test asserting `OPTIONS /token` and `OPTIONS /revoke` carry **no** `access-control-allow-origin` header when the env var is unset would lock this down.

bunlongheng:
Fixes #3845.

## Bug

With `GBRAIN_HTTP_CORS_ORIGIN` unset, `gbrain serve --http` intends default-deny CORS on every OAuth endpoint (v0.41.3 T7). The actual-request path is default-deny, but the **OPTIONS preflight on `/token` and `/revoke` returns `Access-Control-Allow-Origin: *`**, leaking the endpoint surface + methods to any web origin and contradicting the documented posture.

```
$ curl -sS -D - -o /dev/null -X OPTIONS http://127.0.0.1:3131/token \
    -H 'Origin: https://evil.example' -H 'Access-Control-Request-Method: POST'
HTTP/1.1 204 No Content
Access-Control-Allow-Origin: *
Access-Control-Allow-Methods: GET,HEAD,PUT,PATCH,POST,DELETE
```

## Root cause (deeper than the issue's suggested one-liner)

The gate `app.use('/token', cors(corsOAuthOptions))` runs first, but for a denied / default-deny origin **cors@2.8.x does not emit a header and does not short-circuit the preflight — it calls `next()`**. In `cors/lib/index.js`, a falsy `origin` option (our `resolveCorsOrigin(null) === false`) makes `originCallback` null, so the middleware just `next()`s; the allowlist function form does the same via `cb(null, false)` for a denied origin.

Control then reaches the MCP SDK's `mcpAuthRouter`, whose token/revoke/register handlers each mount a **bare `cors()` (origin `*`) as their first middleware**:

```
# @modelcontextprotocol/sdk 1.29.0 — auth/handlers/token.ts, revoke.ts, register.ts
router.use(cors());
router.use(allowedMethods(['POST']));
```

That downstream `cors()` answers the OPTIONS preflight with `*`. `/mcp` and `/authorize` are **not** shadowed (no downstream `cors`; `/authorize` runs `allowedMethods` first → 405), which matches the issue's observations exactly.

I verified that the issue's suggested `resolveCorsOrigin` change (callback form for the null case) **does not fix it** — gbrain's gate still `next()`s for a denied origin, so the SDK's `*` still wins the preflight. Reproduced with an isolated express app mirroring the mount order:

| variant | evil OPTIONS `/token` | claude OPTIONS `/token` |
|---|---|---|
| current / suggested one-liner | `204` + `aco: *` | `204` + `aco: *` |
| **this PR** (allowlist unset) | `204` + `aco: none` | `204` + `aco: none` |
| **this PR** (allowlist=`claude.ai`) | `204` + `aco: none` | `204` + `aco: https://claude.ai` |

## Fix

`mountOAuthCorsGate` wraps `cors()` so the gate **owns** the preflight: `cors()` only reaches the wrapper's callback when it did not short-circuit itself (i.e. the origin was denied, or it is a real non-OPTIONS request). A denied `OPTIONS` is terminated here with a header-free `204`, so the SDK's `cors()` never runs. Allowed origins are still short-circuited by `cors()` with the reflected `Origin`, and real requests fall through unchanged. Applied to the three SDK-shadowed endpoints (`/token`, `/revoke`, `/register`); `/mcp` and `/authorize` keep the plain gate.

## Tests

Extended `test/serve-http-cors.test.ts` with integration coverage that mounts the **real** exported gate in front of a bare-cors SDK stub (same ordering as `serve-http.ts`) and asserts:

- allowlist unset → `OPTIONS /token` and `OPTIONS /revoke` carry **no** `Access-Control-Allow-Origin` (the regression)
- allowlist set → non-listed Origin preflight is denied (no `Allow-Origin`)
- allowlist set → listed Origin preflight reflects that Origin
- real `POST` still reaches the downstream handler

```
$ bun test test/serve-http-cors.test.ts
 18 pass  0 fail  34 expect() calls
```

`tsc --noEmit` is clean; the neighboring `serve-http-*` unit suites (58 tests) pass. No behavior change for same-origin requests, allowed origins, or actual (non-OPTIONS) requests.

### DIFF
--- a/src/commands/serve-http.ts
+++ b/src/commands/serve-http.ts
@@ -12,7 +12,7 @@
 
 import express from 'express';
 import type { Socket } from 'net';
-import type { Request, Response, NextFunction } from 'express';
+import type { Request, Response, NextFunction, RequestHandler } from 'express';
 import cookieParser from 'cookie-parser';
 import cors from 'cors';
 import rateLimit from 'express-rate-limit';
@@ -406,7 +406,11 @@ export function parseCorsAllowlistOAuth(): Set<string> | null {
 /**
  * Build a `cors.CorsOptions['origin']` value from the allowlist. The cors
  * package accepts:
- *   - `false` → reject everything (no Allow-Origin header sent)
+ *   - `false` → NOT an Allow-Origin header of "none"; cors@2.8.x treats a
+ *     falsy `origin` option as "no CORS gate" and simply calls `next()`
+ *     without setting or short-circuiting anything (see the mismatch note on
+ *     `mountOAuthCorsGate` below). We keep `false` for the null-allowlist case
+ *     because the gate is enforced by `mountOAuthCorsGate`, not by cors.
  *   - `(origin, cb) => cb(null, boolean)` → dynamic per-request check
  * We use the function form when an allowlist is set so the value of the
  * Allow-Origin header echoes the request Origin (RFC 6454) instead of a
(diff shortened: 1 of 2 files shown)


time-attack:

PR #3850, opened 2026-08-06
fix(mcp): treat null / empty-string optional params as absent at dispatch

SeanGearin:
**The bug.** An MCP client that fills an optional param with `""` or `null` instead of omitting it gets a silent wrong answer. Against a fresh PGLite brain with one fact, over the real stdio transport at `15b9863d`:

```
recall {}              →  {"facts": [ … 1 row … ]}
recall {since: ""}     →  {"facts": [], "total": 0}      isError: false
recall {since: null}   →  {"facts": [], "total": 0}      isError: false
```

No error is raised, so the model concludes the brain is empty and moves on. This is the failure mode worth fixing: a loud rejection would at least prompt a retry.

**Why it happens.** `validateParams` reads `null` as absent for the required-param check, and `""` type-checks fine as a string, so both reach the handler. Handlers then disagree about what absence looks like — `recall` guards `entity` and `session_id` with `typeof p.x === 'string' && p.x.length > 0`, but branches on `p.since !== undefined`. The since-branch parses `""` to no date and returns nothing. The same asymmetry exists wherever a handler tests an optional param with `!== undefined`.

**Who hits it.** Models other than Claude routinely emit `""`/`null` for optional params rather than omitting the key — it shows up immediately when a gbrain MCP surface is driven from agent harnesses (OpenClaw, Hermes) or from any smaller model. The client is not doing anything illegal; the JSON is well-formed and type-correct.

**The fix.** One normalization pass in `dispatchToolCall` — the single path both MCP transports share — so every handler sees one canonical absence instead of each guarding for itself. Strips `null` (any type) and `""` (string-typed) for **optional declared params only**.

Deliberately narrow, each pinned by a test:
- Required params untouched: `get_page {slug: null}` still fails "Missing required parameter: slug".
- Type-mismatched junk untouched: `recall {limit: ""}` still fails "must be a number".
- Undeclared keys untouched. Caller objects never mutated (copy-on-write, no-op inputs return the same object).

**Not in this PR**, happy to add if you want them: coercing `"50"` → `50` / `"false"` → `false` for type-mismatched optionals (changes accepted-value semantics, felt like your call), and treating the literal strings `"null"` / `"undefined"` as absent (I only found `""`/`null` in practice, and did not want to widen your parser on speculation).

**Tests.** `test/mcp-dispatch-optional-params.test.ts` — 12 tests against a real `PGLiteEngine` through `dispatchToolCall`, covering the tolerated idioms end to end plus the three non-goals. Reverting just the call site turns the two `since` tests red. `bun run verify` is green (34/34), as are all 12 existing test files that touch the dispatch layer.

### DIFF
--- a/src/mcp/dispatch.ts
+++ b/src/mcp/dispatch.ts
@@ -199,6 +199,43 @@ export function validateParams(op: Operation, params: Record<string, unknown>): 
   return null;
 }
 
+/**
+ * Normalize the absent-idioms real MCP clients send for OPTIONAL params
+ * before validation and dispatch: `null` (any type) and `''` (string-typed
+ * params) become "not provided".
+ *
+ * Why: non-Claude models routinely fill optional params with `""` or `null`
+ * instead of omitting them. `validateParams` already reads null as absent
+ * for the required-param check (see above); handlers do not — some guard
+ * with `typeof p.x === 'string' && p.x.length > 0` (entity / session_id in
+ * `recall`), others with `p.x !== undefined` (since in `recall`), so
+ * `recall {since: ""}` silently returned zero facts where the same call
+ * with `since` omitted returns rows. Normalizing once at the shared
+ * dispatch layer gives every handler one canonical absence instead of
+ * per-handler guards.
+ *
+ * Deliberately narrow:
+ *   - Required params are untouched — null on a required param still fails
+ *     validation loudly, and `''` on a required string still reaches the
+ *     handler exactly as before.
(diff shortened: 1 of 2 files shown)


time-attack:
Merged in the v0.45.8.0 community fix-wave (#4050). Your fix landed on master as 0a34ced5d7, credited to you via Co-Authored-By, and is called out in the v0.45.8.0 CHANGELOG entry. Thank you for the contribution!

PR #3851, opened 2026-08-06
fix(recall): SQL-side grep + bare-token entity resolution for hosts/projects dirs

miroslavb:
## What

Two `recall` retrieval defects found by a live memory eval (retrieval-QA probes over a brain with ~20k extracted conversation facts):

**1. `grep` filtered client-side AFTER the limit.** On a high-cardinality entity (hundreds of facts), the matching fact sits outside the newest-N window, so `recall --entity X --grep needle` returns empty while the fact exists — indistinguishable from "not in memory". The filter now lands in SQL (`fact ILIKE`, wildcards escaped via the existing `escapeLikePattern`) in all three fact-list methods of **both engines** (parity preserved); `FactListOpts.grep` documents the contract. The handler's client-side pass survives only for the supersessions branch, which bypasses `FactListOpts`.

**2. Bare tokens with a canonical `hosts/<token>` page fell through to slugify.** Prefix expansion only tried `people/` and `companies/`, and its LIKE pattern matched *suffixed* children only (`<dir>/<token>-%`), so an exact bare-child page (`hosts/boxname`) was invisible. `hosts/` and `projects/` joined `PREFIX_EXPANSION_DIRS`, and the expansion SQL now also matches `<dir>/<token>` exactly. A docs-heavy namespace (`infra/`) was tried and deliberately excluded — its analysis pages (`infra/<token>-audit-…`) trip the ambiguity gate and re-break resolution; the comment in the code pins that reasoning.

## Verification

- Typecheck clean on top of upstream master (0 errors).
- `entity-resolve-slug-fallback` / `recall-extensions` / `facts-engine` suites green (35 tests).
- Verified live over MCP on a production brain: both previously-empty probes (`grep` on a hundreds-of-facts entity; bare host token) now return the facts.

### DIFF
--- a/src/core/engine.ts
+++ b/src/core/engine.ts
@@ -552,6 +552,15 @@ export interface FactListOpts {
    * are returned. Remote (untrusted) callers must supply ['world'].
    */
   visibility?: FactVisibility[];
+  /**
+   * Case-insensitive substring filter on fact text, applied IN SQL (before
+   * limit). A client-side post-limit grep silently returns nothing for
+   * high-cardinality entities — the match may sit outside the newest-N
+   * window (found by the 2026-08-06 memory eval on an entity with hundreds
+   * of facts). LIKE wildcards in the needle are escaped; plain substring
+   * semantics only.
+   */
+  grep?: string;
 }
 
 /** Per-source operational health snapshot consumed by `gbrain doctor`. */
--- a/src/core/entities/resolve.ts
+++ b/src/core/entities/resolve.ts
@@ -111,7 +111,13 @@ function isBareName(raw: string): boolean {
   return true;
 }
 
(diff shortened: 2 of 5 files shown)


time-attack:
Re-trialed against the current integration branch (post v0.46.23-25 waves): Conflicts: src/core/operations.ts, pglite-engine.ts, postgres-engine.ts — v0.46 facade peel moved fact-list SQL to *-engine/facts.ts + ops/*. NOT superseded: wave lacks FactListOpts.grep + hosts/projects expansion. Diff benign. — a rebase along those lines puts this back in the active merge queue; re-trials run nightly, so a fresh push gets picked up automatically. Thank you!

garrytan:

PR #3855, opened 2026-08-07
fix(security): fail-closed facts visibility — omitted remote flag reads world-only

jeanpierre121:
## Problem

`findTrajectory`'s visibility filter keys on `opts.remote === true` (both engines), and the `think` / `find_trajectory` op handlers thread it as `remote: ctx.remote === true`. That polarity is fail-open: a context that loses its `remote` flag (a widening cast, a caller that forgets to thread it) reads `visibility='private'` fact rows.

This contradicts the repo's own stated trust posture. `put_page`'s coercion a few hundred lines up says it directly: *"Fail-closed — anything not strictly local is remote (matches CV6 / v0.26.9 F7b posture)"* — and F7b closed exactly this cast-bypass class for whoami. The facts read path is the one surface still keying trust on strict-true.

## Fix

- Both engines: `remoteFilter = opts.remote !== false`
- Ops layer: `remote: ctx.remote !== false` in the `think` and `find_trajectory` handlers
- Every trusted local caller now says so explicitly (`remote: false`): CLI `think`, the auto-think dream cycle, `founder-scorecard`, `eval-trajectory` (`eval-longmemeval` already did)
- `TrajectoryOpts.remote` / `ThinkOpts.remote` docs state the fail-closed contract
- The facts meta-hook (`meta-hook.ts`) was already fail-closed and is untouched

Behavior is unchanged for every caller that threads the flag: `remote: true` still filters to world, `remote: false` still sees all rows. Only the ambiguous-context default moves — from private-leak to world-only.

## Tests

- `engine-find-trajectory.test.ts`: the `remote default (undefined) is treated as trusted` pin is inverted to pin fail-closed (undefined → world-only, private rows hidden)
- Trajectory/think test callers gain explicit `remote: false` so each test keeps its original intent
- 71 tests across the 8 affected files pass; `tsc --noEmit` clean

## Noted, deliberately out of scope

`operations.ts` has one more strict-true trust site in the same class: the fence-strip trigger `const isUntrustedReader = ctx.remote === true` (v0.32.2). An ambiguous context skips fence stripping there. Left out to keep this PR's surface reviewable; happy to follow up.

### DIFF
--- a/src/commands/eval-trajectory.ts
+++ b/src/commands/eval-trajectory.ts
@@ -139,6 +139,8 @@ export async function runEvalTrajectory(engine: BrainEngine, args: string[]): Pr
       since: parsed.since,
       until: parsed.until,
       limit: parsed.limit,
+      // Fail-closed trust: local CLI must say so explicitly.
+      remote: false,
     });
     const { regressions, drift_score } = computeTrajectoryStats(points);
     result = {
--- a/src/commands/founder-scorecard.ts
+++ b/src/commands/founder-scorecard.ts
@@ -305,6 +305,8 @@ export async function runFounder(engine: BrainEngine, args: string[]): Promise<v
       kind: 'metric',
       since: windowSince,
       until: windowUntil,
+      // Fail-closed trust: local CLI must say so explicitly.
+      remote: false,
     });
     let takes: Take[] = [];
     try {
--- a/src/commands/think.ts
+++ b/src/commands/think.ts
@@ -133,6 +133,9 @@ prints what would have been the input (exit 0).
     try {
       result = await runThink(engine, {
         question, anchor, rounds, save, take, model, since, until,
+        // Fail-closed trust: local CLI must say so explicitly, or trajectory
+        // injection degrades to visibility='world' rows.
+        remote: false,
         // #1698: explicit --model → hard error on an unresolvable model (no silent
         // degrade to the no-LLM stub). Omitting --model keeps the graceful default path.
         modelExplicit: !!model,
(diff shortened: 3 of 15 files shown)


jeanpierre121:
Updated after an adversarial cross-model review of the diff:

- Added an operation-level cast-bypass pin in `test/operations-find-trajectory.test.ts`: a context whose `remote` was never threaded (Partial<> cast) must read world-only through the `find_trajectory` handler. This pins the ops-layer coercion itself — if it ever regresses to `ctx.remote === true`, the undefined context would become an explicit `remote: false` and bypass the engine default; the engine-level pin alone would not catch that.
- The `think` handler's coercion is the character-identical expression and shares the engine-level pin; a handler-level pin there needs LLM stubbing beyond this PR's footprint, so I left it out deliberately.
- The review also suggested a live-Postgres assertion for the polarity (all new visibility pins run on PGLite). Both engines' `remoteFilter` lines are textually identical after this change; happy to add a Postgres E2E pin if your CI harness has a live Postgres for it.

72 tests across the 8 affected files pass; `tsc --noEmit` clean.

time-attack:
Re-trialed against the current integration branch (post v0.46.23-25 waves): Conflicts: operations.ts, pglite-engine.ts, postgres-engine.ts (v0.46.9.1 peel moved code). Ops half landed fail-closed (ops/takes.ts:224, ops/insights.ts:291); engine flip in */facts.ts + CLI remote:false still needed — port, not merge. — a rebase along those lines puts this back in the active merge queue; re-trials run nightly, so a fresh push gets picked up automatically. Thank you!

garrytan:

PR #3857, opened 2026-08-07
feat(write-through): add sync.write_through opt-out flag for DB-only brains

DanCanadian:
## What

Adds a `sync.write_through` config flag (opt-out, default unchanged) checked at the top of `writePageThrough` before any FS or DB work:

```bash
gbrain config set sync.write_through false   # every page write is DB-only
gbrain config set sync.write_through true    # (or unset) today's behavior
```

Skips surface as a new `skipped: 'disabled_by_config'` variant so callers (`put_page` MCP result, capture, brainstorm/lsd --save) report *why* no file appeared instead of silently not writing.

## Why

When the brain's `default` source is a **shared working tree** — a team's actual code repository indexed for code-aware recall — every `put_page` drops a rendered `.md` at the repo root. Each agent session journaling to the brain leaves stray root-level files that pollute `git status` for every collaborator and every CI diff.

Today there is no supported way out: unsetting the source's `local_path` kills sync entirely, and the #2018 leak-guard only covers the *cross-source* topology. We hit this running gbrain against a live monorepo (the exact setup INSTALL_FOR_AGENTS.md encourages): 3 stray root files in one day across parallel agent sessions.

Only the explicit string `'false'` disables — mirroring the existing string-compare convention (`search.mcp_keyword_only`, `autopilot.*`). The DB row remains the durable sink; `gbrain sync` reconciliation is unaffected.

## Tests

TDD (red first): three new cases in `test/write-through.test.ts` —
- flag off gates the `sync.repo_path` branch: `disabled_by_config`, nothing on disk, DB row intact;
- flag off gates the per-source `local_path` branch too (the check runs before target resolution);
- explicit `'true'` / unset keeps the default write-through behavior.

`bun test test/write-through.test.ts test/ingestion/put-page-write-through.test.ts` → **22 pass / 0 fail**.

### DIFF
--- a/src/core/config.ts
+++ b/src/core/config.ts
@@ -963,6 +963,10 @@ export const KNOWN_CONFIG_KEYS: readonly string[] = [
   'sync',
   'sync.repo_path',
   'sync.last_commit',
+  // Opt-out for the put_page/capture disk write-through (write-through.ts):
+  // 'false' makes every page write DB-only. For brains whose host repo is a
+  // shared working tree where stray root-level .md artifacts are unwanted.
+  'sync.write_through',
   // Gateway-native subagent loop toggle (routes subagent jobs through the
   // provider-agnostic gateway.toolLoop for non-Anthropic providers). The
   // subagent handler's error message tells users to `config set` this, so it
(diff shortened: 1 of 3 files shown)


time-attack:

time-attack:

PR #3858, opened 2026-08-07
v0.42.74.0 feat(ai): add Atlas Cloud provider recipe

binyangzhu000-sudo:
## Summary

- register Atlas Cloud as an OpenAI-compatible provider for chat and query expansion
- isolate authentication to `ATLASCLOUD_API_KEY` and use the official `https://api.atlascloud.ai/v1` endpoint
- expose the catalog-backed `deepseek-ai/deepseek-v4-pro` model with tools, structured outputs, and its 1,048,576-token context window
- reuse the existing fail-open DeepSeek `reasoning_content` compatibility fetch
- add provider registry, auth, model-resolution, and CLI readiness regression coverage

## Design notes

- `supports_subagent_loop` remains `false` until replay stability is proven.
- Recipe-level pricing is omitted because account terms can vary.
- Existing providers and configuration paths are unchanged.

## Validation

- `bun test test/ai/recipe-atlascloud.test.ts test/providers.test.ts test/ai/recipes-existing-regression.test.ts` (32 pass, 0 fail)
- `bun run typecheck`
- `bun run verify` (34 pass, 0 fail)
- `bun run build` (1,757 modules bundled and compiled)
- live public Atlas Cloud catalog metadata check for `deepseek-ai/deepseek-v4-pro`

`bun run ci:local:diff` could not start its test phase because the local Docker CLI blocked on `docker ps`; it was stopped after the mandatory verification, focused tests, typecheck, and build had all passed.


time-attack:

time-attack:
Thanks for this — the recipe itself is genuinely well made: it follows the openai-compat pattern faithfully, isolates the credential in its own env var, reuses the existing DeepSeek reasoning_content compatibility fetch, keeps subagent replay conservatively off, and ships a shape test plus a KEY_FILES entry. The gateway APIs it references all check out.

What gates it isn't the code. New provider recipes are a product-surface decision we make deliberately, and our bar for AI-provider PRs is a live roundtrip against the actual endpoint — chat, expansion, tools, and structured outputs on the real model — because a recipe pinned to a third-party API that no CI exercises is exactly how a repo ends up shipping something wrong silently. We don't hold an Atlas Cloud account, so we can't verify the model or context-window claims ourselves, and the one provider recipe we've merged from an external contributor came with the author's own field-validation evidence.

So we're declining this branch as-is, not the idea. Two things would change the answer on a resubmission: (1) include your own roundtrip receipts — transcripts or a small script output showing chat/expansion/tools/structured-outputs working against deepseek-ai/deepseek-v4-pro on a real key; (2) drop the VERSION/package.json/CHANGELOG bump commits — release bumps happen at ship time here, and yours is already conflicting with master. With those, this becomes a straightforward maintainer product call rather than an unverifiable vendor add.

PR #3860, opened 2026-08-07
v0.42.73.3 fix(autopilot): ignore foreign PIDs in stale locks (#2503)

### LINKED ISSUE #2503: autopilot lock can false-positive when stale PID is reused by unrelated process

javieraldape:
## Summary

Fixes #2503 by making the autopilot lock check reject reused PIDs from unrelated live processes.

Previously, both `gbrain autopilot` startup and `gbrain status` trusted `kill -0` alone. If a stale `autopilot.lock` pointed at a PID that now belonged to another process, autopilot exited as if another daemon was running and status reported a false healthy state.

## Changes

- Adds a shared autopilot-lock classifier that combines PID liveness with a lightweight command-line identity probe.
- Keeps the conservative behavior when process identity cannot be inspected, so a genuinely live daemon is not stolen on unknown platforms or permission failures.
- Uses the shared classifier in `decideLockAcquisition` so foreign live PIDs are treated as stale locks.
- Uses the same classifier in `gbrain status`, so a foreign PID no longer reports `autopilot.running=true`.

## Tests

- Extends `test/autopilot-lock.test.ts` for live autopilot holders, foreign PID reuse, and unknown process identity.
- Adds `test/status-autopilot-lock.test.ts` to pin `gbrain status`'s JSON-facing lock interpretation.

## Verification

- `bun test test/autopilot-lock.test.ts test/status-autopilot-lock.test.ts` - 12 pass, 0 fail, 19 expect calls.
- `bash scripts/gbrain-gate.sh <worktree> test/autopilot-lock.test.ts test/status-autopilot-lock.test.ts` - `RESULT: GREEN`; verify green; focused unit tests green; unrelated E2E skipped by the diff selector.

### DIFF
--- a/src/commands/autopilot.ts
+++ b/src/commands/autopilot.ts
@@ -25,6 +25,11 @@ import { execSync } from 'child_process';
 import type { BrainEngine } from '../core/engine.ts';
 import { loadPreferences } from '../core/preferences.ts';
 import { loadConfig, loadConfigFileOnly, saveConfig, gbrainPath as gbrainHomePath } from '../core/config.ts';
+import {
+  classifyAutopilotLockHolder,
+  type AutopilotLockProbeDeps,
+  isPidAlive,
+} from '../core/autopilot-lock.ts';
 import { ChildWorkerSupervisor } from '../core/minions/child-worker-supervisor.ts';
 import { VERSION } from '../version.ts';
 import {
@@ -244,19 +249,12 @@ export function shouldSpawnAutopilotWorker(args: string[]): boolean {
   return !args.includes('--no-worker');
 }
 
-export function isPidAlive(pid: number): boolean {
-  if (!Number.isFinite(pid) || pid <= 0) return false;
-  try {
-    process.kill(pid, 0);
-    return true;
-  } catch (error: unknown) {
-    return (error as NodeJS.ErrnoException).code === 'EPERM';
-  }
-}
+export { isPidAlive };
 
 export function decideLockAcquisition(
   lockPath: string,
   currentPid: number,
+  deps: AutopilotLockProbeDeps = {},
 ): { action: 'acquire' } | { action: 'exit'; holderPid: number } | { action: 'takeover'; reason: string } {
   if (!existsSync(lockPath)) return { action: 'acquire' };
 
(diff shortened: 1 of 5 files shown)


time-attack:
Automated triage (adversarial review) found a demonstrated blocker before merge: on a source-checkout daemon whose command line reads `bun src/cli.ts autopilot --repo ...` (relative path in argv, a normal launch shape), the new classifier reports `alive-foreign` and `decideLockAcquisition` returns takeover — i.e. the fix steals the lock from a genuinely live autopilot daemon, risking two concurrent daemons against one PGLite store. Reproduced live against a real spawned daemon at the PR head. The identity regex is simultaneously under-inclusive (fork/renamed checkouts, relative argv) and over-inclusive. Suggest treating live-PID + unrecognized-command as exit unless corroborated by lock-file age/metadata (see db-lock.ts's age-gated grace-window pattern). Routed to maintainer review rather than auto-merge.

javieraldape:
Maintenance update:

- Rebased this branch onto current `master` to resolve the merge conflict.
- Resolved the conflict in `src/commands/autopilot.ts` by keeping current master CLI path/config changes and the PR autopilot-lock holder classification import.
- Pushed new head: `8fc6c32a`.

Validation:
- `bun test test/autopilot-lock.test.ts test/status-autopilot-lock.test.ts` — 13 pass, 0 fail
- `bun run typecheck` — pass
- `bun run verify` — 38/38 checks green

GitHub CI has started on the new head; I will include its observed final state in the maintenance report once checks settle or time out.

time-attack:
Merged in the v0.45.10.0 community fix-wave 2 (#4085). Your fix landed on master as 5087507de0, credited to you via Co-Authored-By, and is called out in the v0.45.10.0 CHANGELOG entry. Thank you for the contribution!

PR #3861, opened 2026-08-07
v0.42.73.3 fix(autopilot): honor source pull policy (#3835)

### LINKED ISSUE #3835: `sync` and `autopilot-cycle` job handlers read opposite-polarity pull keys, so the autopilot freshness dispatcher cannot suppress a pull
## Summary

Two job handlers express the same concept with inverted keys:

```ts
// src/commands/jobs.ts:1417  worker.register('sync', ...)
const noPull = !!job.data.noPull;

// src/commands/jobs.ts:1859  worker.register('autopilot-cycle', ...)
// Pull default: legacy `true` for back-compat; explicit boolean wins.
const pull = typeof job.data.pull === 'boolean' ? job.data.pull : true;
```

The autopilot freshness dispatcher (`src/commands/autopilot.ts`, the
`queue.add('sync', ...)` with idempotency key `autopilot-sync:<id>:<slot>`) sends **neither** key.
`noPull` is therefore `false` and every dispatched sync attempts a `git pull`.

For a source whose config has no `remote_url`, that pull should never be attempted at all. Where the
working tree's `origin` is a local filesystem path it also cannot succeed, so each sync completes
`status: "partial"`, `reason: "pull_failed"`.

The trap is that the obvious fix does not work. Passing `pull: false` from the dispatcher is inert,
because the `sync` handler does not read `pull`. Verified on a live brain:

```json
// job data
{ "pull": false, "repoPath": "/home/gbrain/brain", "sourceId": "default" }
// job result
{ "status": "partial", "reason": "pull_failed", "fromCommit": "6cc617ee", "toCommit": "6cc617ee" }
```

Switching the same dispatcher to `noPull: true` fixed it immediately:

```json
{ "noPull": true, ... }  ->  { "status": "up_to_date", "embed_skip_reason": null }
```

## Environment

- gbrain 0.42.73.2, bun, Debian 13.6, Postgres engine, pgvector 0.8.0
- One federated source `default`, `local_path` `/home/gbrain/brain`, no `remote_url`
- Working tree fed by `git push` to a bare repo on the same host plus a `post-receive` hook that
  runs `gbrain sync --no-pull`, so `origin` is a local path

## Steps to reproduce

1. Register a local-path source with no `remote_url`.
2. Let the autopilot run a freshness cycle.
3. Inspect the dispatched `sync` job: `result.status` is `partial`, `result.reason` is
   `pull_failed`, every cycle.

## Suggested fix

Either send `noPull: !remoteUrl` from the freshness dispatcher, or, better, make the two handlers
agree on one key so this class of mistake stops recurring. `autopilot-fanout.ts:439` already sends
`pull: !!remoteUrl` for `autopilot-cycle` jobs, which is correct for that handler and wrong for
`sync`, which is easy to miss when reading the two side by side.

## Related

`src/core/sources-load.ts:27` notes "Postgres returns object; PGLite returns JSON string. Parse via
`parseSourceConfig`." `autopilot-fanout.ts:433` reads `src.config?.remote_url` directly, so on
PGLite it yields `undefined` and `pull` resolves to `false` for every source. A PGLite user with a
real `remote_url` would silently never pull. Both call sites should use `parseSourceConfig`.

RerankerGuo:
Fixes #3835.

## Problem

Autopilot's freshness dispatcher queued `sync` jobs without either pull-policy
key. The sync handler interpreted that as "pull", so local-only sources tried
to pull every cycle and ended in `partial / pull_failed`.

The adjacent fan-out path also read `source.config.remote_url` directly. That
worked for Postgres objects but not PGLite's JSON-string config shape.

## Changes

- Use `pull` as the canonical positive-polarity job field in both sync and
  autopilot-cycle handlers.
- Preserve queued jobs that still carry the inverse legacy `noPull` field;
  explicit `pull` takes precedence.
- Derive pull policy through the shared source-config parser in both autopilot
  dispatch paths.
- Add focused policy, PGLite-config, and dispatcher wiring regressions.
- Document the resulting job and source-config contracts in `KEY_FILES.md`.

## Verification

- `bun test test/job-pull-policy.test.ts test/autopilot-fanout.test.ts test/autopilot-fanout-wiring.test.ts test/sources-load.test.ts`
  - 61 pass, 0 fail
- `bun run verify`
  - 34 pass, 0 fail
- `bun run typecheck`
  - pass
- `bun run test:slow`
  - 41 pass, 0 fail
- `git diff --check`
  - pass

The full unit command reached 7,676 passes before two unrelated failures in
untouched retrieval-reflex and worker-registry tests; two shards also hit the
runner's 1,500-second environment timeout. Both failures reproduce when their
files are run alone. Docker is unavailable on this machine, so the
Docker-backed E2E gate was not run.

## Scope

This does not change pull behavior for legacy payloads that omit both fields:
they continue to default to pulling.


ghost (commented):
Why¿?

PR #3862, opened 2026-08-07
v0.42.73.3 fix(sync): honor file transport escape hatch (#3836)

### LINKED ISSUE #3836: `GBRAIN_GIT_ALLOW_FILE_TRANSPORT=1` does not apply to `pullRepo`, so the documented escape hatch has no effect on sync
## Summary

`durableSsrfFlags()` honours the escape hatch:

```ts
// src/core/git-remote.ts:396
function durableSsrfFlags(): string[] {
  const fileAllow = process.env.GBRAIN_GIT_ALLOW_FILE_TRANSPORT === '1' ? 'always' : 'never';
  ...
}
```

and its docstring states the hatch exists "so self-hosted local-filesystem remotes and the test
suite can use the file transport."

But `pullRepo`, the function `sync.ts` actually calls, uses the hardcoded constant instead:

```ts
// src/core/git-remote.ts:219
export function pullRepo(repoPath: string, opts: { timeoutMs?: number } = {}): void {
  const args: string[] = ['-C', repoPath, ...GIT_SSRF_FLAGS, 'pull', ...];
```

```ts
// src/core/git-remote.ts:44
export const GIT_SSRF_FLAGS = [
  '-c', 'http.followRedirects=false',
  '-c', 'protocol.file.allow=never',   // unconditional
  '-c', 'protocol.ext.allow=never',
] as const;
```

So an operator who sets `GBRAIN_GIT_ALLOW_FILE_TRANSPORT=1` to support a local-filesystem origin
gets no change in behaviour on the sync path, with no warning. The variable was set here on
2026-07-31, is present in the daemon's `/proc/<pid>/environ`, and had no effect for five days.

Plain `git fetch origin master` in the same working tree succeeds, which confirms the refusal comes
from gbrain's flags rather than from git or the repository.

## Environment

Same as issue 1.

## Suggested fix

Have `pullRepo` build its global flags through the same helper that honours the escape hatch, or
document explicitly that the variable covers only the durability fetch/pull/push paths and not
`sync`. The current split is easy to misread, since the hatch's docstring describes precisely the
use case that `sync` represents.

## Notes

Happy to open a PR for either.

RerankerGuo:
Fixes #3836.

## Problem

`GBRAIN_GIT_ALLOW_FILE_TRANSPORT=1` was documented for self-hosted
filesystem remotes, but `pullRepo` and the sync cost-preview fetch still used
the immutable `protocol.file.allow=never` flag set. The opt-in therefore had
no effect on normal sync.

## Changes

- Keep the exported clone/remote-discovery flags strict.
- Share one runtime SSRF flag builder across operations on existing checkouts:
  sync pull, cost-preview fetch, and durability pull/push.
- Keep file transport denied by default; only the exact value `1` enables it.
- Preserve redirect blocking, external-helper blocking, no-prompt behavior,
  `--no-recurse-submodules`, and `pull --ff-only`.
- Add default-deny and explicit-opt-in regression coverage.

## Verification

- `bun test test/git-remote.test.ts` — 38 pass, 0 fail
- `bun test --max-concurrency=1 test/git-remote-durable.serial.test.ts test/sync-pull-failed-anchor.serial.test.ts` — 14 pass, 0 fail
- `bun run typecheck` — pass
- `bun run verify` — 34 pass, 0 fail
- `git diff --check` — pass

Docker is unavailable on this machine, so the Docker-backed E2E gate was not
run.

## Security boundary

Clone and remote discovery remain HTTPS-only and continue using the immutable
strict flag set. This change affects only commands operating on an existing
checkout after the operator explicitly sets the documented environment
escape hatch.


time-attack:
Automated triage: full test ladder green (targeted 51+11 pass, `bun run verify` 34/34), and two independent adversarial reviews both concluded no articulable defect — the default (env unset) git flag set was proven byte-identical before/after by live argv probes at every call site, clone/ls-remote discovery stays strict, and the new opt-in test fails on the merge-base. This PR is held for maintainer sign-off only because it edits SSRF-protection flag construction (security-sensitive path is above the autonomous-merge line), not because any problem was found.

time-attack:

click once rebased.

PR #3863, opened 2026-08-07
v0.42.73.3 fix(code): index decorated Python definitions (#3821)

### LINKED ISSUE #3821: code-def returns 0 for every decorated Python symbol — decorated_definition handled in none of four sites, while TS export_statement is handled in all four
## Summary

Every decorated Python definition is missing from the code index. `gbrain code-def`,
`code-refs` and `code-callers` return `count: 0` for them — the same confident empty
answer a genuinely nonexistent symbol produces, with no warning anywhere.

The cause is that `decorated_definition` is handled in **none** of the four places that
matter, while the equivalent TypeScript wrapper `export_statement` is handled in **all
four**. In tree-sitter-python a decorated definition parses as:

```
decorated_definition
├── decorator
└── function_definition   (field name: `definition`)
```

so the real `function_definition` never surfaces where the chunker looks for it.

This is not a niche construct: `@property`, `@staticmethod`, `@classmethod`,
`@dataclass`, `@pytest.fixture`, `@app.route`, `@click.command` and every framework tool
decorator disappear. On the repo I hit this with, 79 of 1464 definitions are decorated —
including the entire public tool surface, which is precisely what you go looking for.

Related in symptom, not cause: #3356 (Dart, ABI mismatch), #3602 (C#), #3707 (empty
result conflates "graph not built" with "outside your grant").

## Reproduce

```python
# probe.py — bodies must be substantial, see note below
@property
def decorated_one():
    ...40 lines...

def plain_one():
    ...40 lines...
```

```
gbrain sources add probe --path .
gbrain sync --strategy code --source probe
gbrain code-def decorated_one   # count: 0
gbrain code-def plain_one       # count: 1
```

**Note for whoever reproduces this:** keep the function bodies large. With short bodies
the sibling-merging path produces a single chunk with `symbolType: "merged"` and
`symbolName: null`, and *both* functions vanish — which looks like a different bug and
sent me down a wrong path for a while.

## Root cause — four sites

| # | File | What is there | What is missing |
|---|---|---|---|
| 1 | `src/core/chunkers/code.ts`, `TOP_LEVEL_TYPES.python` | `function_definition`, `class_definition`, … | `decorated_definition` — so the node is never selected as a chunk boundary |
| 2 | `src/core/chunkers/code.ts`, `extractSymbolName` | unwraps `childForFieldName('declaration')` (TS `export_statement`) | Python's wrapper field is named `definition`, not `declaration` |
| 3 | `src/commands/code-def.ts`, `findCodeDef` | `symbol_type IN (…, 'export statement')` | `'decorated definition'` is not in the allow-list |
| 4 | `src/core/chunkers/code.ts`, `NESTED_EMIT_CONFIG.python` | `childTypes: {function_definition}` | a decorated *method* sits in the class body as `decorated_definition` |

Sites 1–3 are indivisible. Fixing 1 alone yields chunks with `symbol_name` NULL, i.e.
exactly the shape of #769. Sites 1–3 fix module-level functions; site 4 is needed for
methods.

Worth flagging: the comment above the SQL branch of `extractSymbolName` states that the
`statement` wrapper is *"unique to SQL among gbrain's 37 grammars (Step 0 inspection
2026-05-24)"*. Python's `decorated_definition` is the same shape, so that inspection
missed it — which may be the reason the Python case was never wired up.

## Patch

```diff
--- a/src/core/chunkers/code.ts
+++ b/src/core/chunkers/code.ts
@@ TOP_LEVEL_TYPES
   python: new Set([
+    'decorated_definition',
     'function_definition', 'class_definition',
     'import_statement', 'import_from_statement', 'assignment',
   ]),

@@ NESTED_EMIT_CONFIG
   python: {
     parentTypes: new Set(['class_definition']),
-    childTypes: new Set(['function_definition']),
+    childTypes: new Set(['function_definition', 'decorated_definition']),
   },

@@ extractSymbolName
   const declaration = node.childForFieldName('declaration');
   if (declaration) { ... }
+
+  const definition = node.childForFieldName('definition');
+  if (definition) {
+    const nested = extractSymbolName(definition);
+    if (nested) return nested;
+  }

--- a/src/commands/code-def.ts
+++ b/src/commands/code-def.ts
-       AND cc.symbol_type IN ('${DEF_TYPES.join("','")}', 'export statement')
+       AND cc.symbol_type IN ('${DEF_TYPES.join("','")}', 'export statement', 'decorated definition')
```

Chunks then carry `symbolType: "decorated definition"`. Emitting the inner node's type
(`function` / `class`) instead would be tidier and would drop the change at site 3, at the
cost of losing the "this is decorated" signal — your call which you prefer.

## Verification

Applied to a real 1464-definition Python repo (symbol and file names generalised below),
then `gbrain reindex-code --source <src> --yes --force`:

| Symbol | Decorator | Before | After |
|---|---|---|---|
| `tool_one` | `@mcp.tool()` | count 0 | `module_a.py:36` |
| `tool_two` | `@mcp.tool()` | count 0 | 2 hits, two modules |
| `tool_three` | framework decorator | count 0 | `module_b.py:131` |
| `tool_four` | framework decorator | count 0 | 2 hits |
| `cached_value` | `@property` | count 0 | `module_c.py:247` |
| `_helper` | `@staticmethod` | count 0 | `module_d.py:17` |
| four undecorated functions + one class, same modules | none | count 1 each | count 1 each — unchanged |

Six recovered, five unaffected, no regression.

## Three side findings

1. **`reindex-code` without `--force` is a no-op after a chunker change.** The source
   files are unchanged, so `content_hash` short-circuits and it reported
   `0 reindexed, 71 skipped`. The `sources.chunker_version` gate that exists for exactly
   this did not fire — a local edit does not move that version, so anyone testing a
   chunker patch will conclude their fix does nothing.

2. **`code-def` reports the decorator's line, not the `def`'s** — line 36 where the `def`
   is on 37. Correct for the `decorated_definition` node, surprising at the call site.

3. **A `count: 0` is indistinguishable from "not indexed".** This is the part that cost
   the most time, and it is the same complaint as #3707. Anything that separated "no such
   symbol" from "this file/construct was never chunked" would have turned a multi-hour
   investigation into a one-line answer.

## Environment

gbrain 0.42.73.2 (`15b9863d`), bun 1.3.11, Windows 11 / MSYS bash, PGLite engine,
`voyage:voyage-code-3` @ 1024d, 72 code pages / 1523 chunks, all embedded.

RerankerGuo:
## Summary

- index tree-sitter Python `decorated_definition` nodes at module and class scope
- derive symbol names and canonical function/class types from the wrapper `definition` field while retaining decorators in the chunk text
- bump `CHUNKER_VERSION` to 5 so existing code pages are re-chunked

Closes #3821.

## Root cause

The Python semantic-node and nested-child allowlists only recognized bare `function_definition` / `class_definition` nodes. Tree-sitter wraps every decorated definition in `decorated_definition`, so top-level decorated symbols fell through to anonymous text chunks and decorated methods were omitted from class expansion.

## Evidence

Before the fix, the new regression tests failed in all three observable paths:

- chunker: 31 pass, 2 fail; neither the top-level decorated function nor decorated method had a symbol chunk
- PGLite `findCodeDef`: 10 pass, 1 fail; `decorated_health_check` returned zero results

After the fix:

- focused chunker, version-gate, and PGLite definition tests: 47 pass, 0 fail
- `bun test test/e2e/code-indexing.test.ts`: 24 pass, 0 fail
- `bun run typecheck`: pass
- `git diff --check`: pass

The first CI run exposed two stale version-4 assertions in `test/chunker-version-gate.test.ts`; the updated head includes that gate and reproduces it locally in the 47/47 focused run above.

Docker-backed E2E was not run locally because Docker is unavailable; the affected indexing path is covered end-to-end with PGLite and will also run in CI.

## Scope

The change is limited to Python semantic chunk boundaries, wrapper metadata extraction, the required chunker-version invalidation, regression tests, and current-state architecture documentation. Other language grammars and query semantics are unchanged.


time-attack:
Automated triage (adversarial review, A/B differential probe of both chunker versions) confirms the decorated-def fix works for functions/methods, but found a real regression before merge: decorators on a decorated *class* (e.g. `@dataclass`) vanish from every emitted chunk — the pre-fix chunker indexed that text, the PR head does not, while KEY_FILES.md claims all decorators are retained. Because CHUNKER_VERSION=5 forces a corpus-wide re-chunk + re-embed, the text loss propagates to every user's brain at real embedding cost. Suggested fix: use the decorated_definition wrapper's range for the class scope-header chunk (mirror the TS export_statement pattern) and add decorated_definition unwrapping to findNestableParent/collectImmediateNestedChildren and splitLargeNode. Routed to maintainer review rather than auto-merge.

RerankerGuo:
Addressed the decorated-class regression in `95e64c68` and rebased onto current master.

- class scope headers now use the outer `decorated_definition` range, preserving `@dataclass`
- nested child classification unwraps `decorated_definition`, so decorated nested classes recurse into their methods
- large-node splitting unwraps the definition body and keeps the wrapper/declaration prefix in the first split

Red control: the three production-shaped regressions failed 33 pass / 3 fail before the fix. Afterward, the expanded code-indexing suite is 94/94, `bun run typecheck` passes, and `bun run verify` is 34/34. New GitHub CI is running.

PR #3867, opened 2026-08-07
fix: source-scope get_recent_salience and find_anomalies

alexey-metaengage:
`get_recent_salience` and `find_anomalies` are the two v0.29 read ops that don't thread the caller's source scope into the engine, unlike the other read ops (`list_pages`, `search`, `traverse_graph`, `find_experts`, `find_orphans`), which all resolve `sourceScopeOpts(ctx)`. This brings the two in line so a source-bound caller sees only its granted sources.

## Change
- Add `sourceId` / `sourceIds` to `SalienceOpts` and `AnomaliesOpts` (canonical precedence: federated array wins over scalar).
- Thread `sourceScopeOpts(ctx)` through both op handlers.
- Add the source-filter clause to both engines' SQL. For `findAnomalies` the same filter is applied to all four cohort queries (tag/type × baseline/today) so the baseline and today windows share one scope and the anomaly math stays self-consistent.
- Trusted local callers (`ctx.remote === false`, empty scope) are unchanged: full-brain results as before.
- `get_recent_transcripts` is intentionally left alone — it is `localOnly` + remote-gated, so it never serves remote callers.

## Test
`test/e2e/salience-anomalies-source-isolation-pglite.test.ts` (PGLite, no `DATABASE_URL` needed): seeds two sources and asserts scalar scoping excludes the other source, the federated array returns the union, an unscoped call still spans both, and both op handlers thread `ctx.sourceId` end-to-end.

### DIFF
--- a/src/core/operations.ts
+++ b/src/core/operations.ts
@@ -3837,11 +3837,18 @@ const get_recent_salience: Operation = {
   },
   handler: async (ctx, p) => {
     const recencyBias = p.recency_bias === 'on' ? 'on' : 'flat';
+    // Scope by the caller's source (canonical sourceScopeOpts ladder: federated
+    // array > scalar > nothing), matching find_orphans/find_experts. Pre-fix
+    // this op returned brain-wide salience regardless of a source-bound OAuth
+    // client's grant — a read leak in the v0.34.1 (#861) source-isolation class
+    // that the v0.29 salience/anomaly batch missed. Trusted local callers
+    // (ctx.remote === false) still get the empty scope = full brain.
     return ctx.engine.getRecentSalience({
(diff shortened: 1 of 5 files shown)


time-attack:
Re-trialed against the current integration branch (post v0.46.23-25 waves): Conflicts: operations.ts, pglite-engine.ts, postgres-engine.ts (wave peeled them into ops/salience.ts + engine/salience.ts). NOT superseded — sourceScopeOpts still missing on branch; re-port fix to peeled modules. Skim clean. — a rebase along those lines puts this back in the active merge queue; re-trials run nightly, so a fresh push gets picked up automatically. Thank you!

garrytan:

PR #3869, opened 2026-08-07
fix: MCP file ops use the connected OperationContext engine, not the global DB singleton

dpaluy:
### Problem

The `file_list`, `file_upload`, and `file_url` MCP operations reached for the module-global `db.getConnection()` singleton instead of the engine carried on the operation context:

```ts  
const sql = db.getConnection();  
```

In the long-running `gbrain serve` / MCP dispatch path, that global is intentionally never initialized — it throws
`connect() has not been called` — so every file operation through MCP failed, even though the operation's own `ctx.engine` was already connected and ready.

This was the same engine-ownership bug class as the operation-context wiring, but the file operations were missed: their handlers took `(_ctx, p)` and ignored the context entirely.

### Fix

Each of the three handlers now resolves SQL from the context engine via the
existing `sqlQueryForEngine(ctx.engine)` helper:

```ts  
handler: async (ctx, p) => {  
  const { sqlQueryForEngine } = await import('./sql-query.ts');  
  const sql = sqlQueryForEngine(ctx.engine);  
  ...  
}  
```

The dynamic import mirrors the pattern already used by the other operations and avoids a top-level import cycle.

### Test

Adds `test/file-upload-engine-context.test.ts` — a regression test that dispatches `file_upload`, `file_list`, and `file_url` through `dispatchToolCall` with a connected `PGLiteEngine` and asserts the full upload → list → url round-trip succeeds. Prior to this fix the first dispatch failed with the uninitialized-singleton error.

Remote dispatch passes an explicit `sourceId: 'default'` to satisfy the v0.42.73 write-scope fence (#3809).

### Verification

- `bunx tsc --noEmit` — clean
- `bun test test/file-upload-engine-context.test.ts` — 1 pass / 0 fail (6 expects)

## Checklist

- [x] Typecheck passes
- [x] Regression test added and passing

### DIFF
--- a/src/core/operations.ts
+++ b/src/core/operations.ts
@@ -3155,8 +3155,9 @@ const file_list: Operation = {
   },
   scope: 'admin',
   localOnly: true,
-  handler: async (_ctx, p) => {
-    const sql = db.getConnection();
+  handler: async (ctx, p) => {
+    const { sqlQueryForEngine } = await import('./sql-query.ts');
+    const sql = sqlQueryForEngine(ctx.engine);
     const slug = p.slug as string | undefined;
     const rows = slug
       ? await sql`SELECT id, page_slug, filename, storage_path, mime_type, size_bytes, content_hash, created_at FROM files WHERE page_slug = ${slug} ORDER BY filename LIMIT ${FILE_LIST_LIMIT}`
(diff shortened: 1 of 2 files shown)


time-attack:
Merged in the v0.45.8.0 community fix-wave (#4050). Your fix landed on master as f7d63c7159, credited to you via Co-Authored-By, and is called out in the v0.45.8.0 CHANGELOG entry. Thank you for the contribution!

PR #3876, opened 2026-08-08
v0.42.74.1 fix(migrate): preserve cross-source link targets (#3859)

### LINKED ISSUE #3859: migrate discards to_source_id for cross-source links
## Summary

In gbrain `v0.42.72.1` (commit `6d1232d5a67c9ab7086d00aaa4d362324919fb3d`), `gbrain migrate --to supabase --url <target-url>` aborts when the source data contains a link whose target page belongs to a different source.

## Reproduction

1. Create a brain with at least two sources.
2. Add a link from a page in source A to a page in source B.
3. Run:

   ```text
   gbrain migrate --to supabase --url <target-url>
   ```

The migration reaches the link-copy phase and fails on the cross-source link.

## Expected behavior

The migration should preserve the link's target source and complete successfully. A link returned by `getLinks` includes `to_source_id`; that value should be used for the destination source, with the current page source as the fallback only when the field is absent.

## Actual behavior

`migrate-engine.ts:420` constructs the copied link with both `fromSourceId` and `toSourceId` set to `page.source_id`, discarding `link.to_source_id` returned by `getLinks` (`engine.ts:162`, `postgres-engine.ts:3066`). The source-qualified validation in `addLink` (`postgres-engine.ts:2944-2950`) then throws and aborts the migration at the first cross-source link.

A local throwaway-copy workaround using `toSourceId: link.to_source_id ?? page.source_id` allowed the migration to complete, which indicates the destination field is the missing value rather than a data incompatibility.

## Request

Please preserve `link.to_source_id` when constructing migrated links and add a regression test covering a cross-source link.

RerankerGuo:
## Summary

- preserve each migrated link target source instead of rebinding it to the origin page source
- use the target `(source_id, slug)` key when skipping links whose destination page failed migration
- retain the origin-source fallback for legacy link rows without `to_source_id`

Closes #3859.

## Root cause

The link-copy phase passed `page.source_id` as both `fromSourceId` and `toSourceId`. A link from `media-corpus::media/x/post-123` to `default::concepts/widget` therefore looked up `media-corpus::concepts/widget` on the target and failed even though the correct default-source page had already migrated. The failed-page filter made the same source assumption.

## Evidence

Before the fix, the PGLite regression failed with:

```text
addLink failed: page "media/x/post-123" (source=media-corpus) or "concepts/widget" (source=media-corpus) not found
```

After the fix:

- `bun test test/e2e/multi-source-bug-class.test.ts`: 9 pass, 0 fail, 52 assertions
- `bun run typecheck`: pass
- `git diff --check`: pass

`bun run test:full` was also attempted: 7,203 pass / 88 fail / 17 skip. Two unrelated shards hit the repository 1,500s timeout, and the reported failures were outside this change (primarily tests writing sandbox-blocked `~/.gbrain` state). No changed test failed; GitHub CI remains the authoritative full-suite run.

## Scope

This only changes engine-migration link copying. Page migration, source catalog copying, manifest format, and same-source link behavior are unchanged.


time-attack:
Merged in the v0.45.10.0 community fix-wave 2 (#4085). Your fix landed on master as 1243a7d3bb, credited to you via Co-Authored-By, and is called out in the v0.45.10.0 CHANGELOG entry. Thank you for the contribution!

PR #3878, opened 2026-08-08
v0.42.75.0 feat: add Voyage reranking through OpenRouter

mikez93:
## Summary

- add opt-in `voyageai/rerank-2.5-lite` and `voyageai/rerank-2.5` support through the existing OpenRouter rerank transport
- price Voyage by input token while preserving Cohere's per-search billing and NVIDIA's free tier
- make reranker accounting conservative and provider-usage-aware, including malformed paid responses
- validate and normalize provider results before returning them to search
- preserve bundled schema packs in Bun-compiled binaries, including checkouts whose paths contain spaces
- keep the default reranker unchanged

This provides an alternative migration path for #3657 without changing existing installations automatically. The schema-pack fallback fixes a production packaging defect found while validating the compiled Voyage build.

## Why

OpenRouter already exposes both current Voyage rerankers, but GBrain's strict reranker recipe allowlist rejected them. The implementation also needed billing metadata that distinguishes token-priced, per-search, and mixed catalogs so cost caps remain accurate across providers.

Bun's compiled binary does not package YAML files reached only through dynamically constructed filesystem paths. As a result, a compiled GBrain could advertise bundled schema packs but fail to load them. The patch statically embeds all bundled manifests as an immutable fallback while retaining disk-first loading for source installations and user-defined packs.

## Test coverage

- 105 focused reranker, recipe, gateway, budget, and schema-pack tests pass
- typecheck and production build pass
- 31 machine-recorded repository verification checks pass
- 19 documentation and recipe checks pass; generated `llms-full.txt` is deterministic
- clean-path serial suite: 909 pass, 5 fail; the same five failures reproduce unchanged on the base commit (cross-modal and autocut/unified-multimodal tests)
- compiled-binary smoke: every bundled schema pack loads; `schema show`, `schema validate`, and active-pack resolution pass
- no database migration or persistence path is changed, so DB E2E was not required

## Review notes

- both colon and slash model ID forms are covered
- Voyage's repeated-query token formula is covered, including conservative dense-text reservation
- authoritative provider usage is retained even when response rows are malformed
- malformed, duplicate, out-of-range, and non-finite result rows fail closed at the provider boundary while search keeps its existing fail-open behavior
- bundled schema manifests are disk-first with embedded fallback; user-authored packs remain filesystem-backed
- the existing transport-wide reranker response-size ceiling remains separate follow-up work; this patch does not expand that surface

## Configuration

```bash
gbrain config set search.reranker.model openrouter:voyageai/rerank-2.5-lite
gbrain config set search.reranker.enabled true
gbrain models doctor --json
```


ghost (approved):

time-attack:

time-attack:
Thanks for this — it is genuinely one of the stronger contributions in the queue: the motivation is real (#3657, the hosted default reranker is sunsetting with signups already closed), the billing_unit/mixed-catalog pricing model is the right shape, and the fail-closed provider-row validation comes with focused tests. Declining as-is on fit, not quality. Provider/model additions to the reranker catalog are a product-surface call we don't take directly from the queue (same lane as #3587), the rerank result-validation change alters result ordering for every provider in edge cases and so needs an eval-replay receipt before it can ride a wave (#3465 lane), and the branch bundles a second concern. That second concern is the salvageable part: the compiled-binary schema-pack fallback fixes a real packaging defect — a compiled gbrain advertising bundled packs it cannot load — and if you send that alone (embedded.ts + load-active.ts + the serial test, no version bump), it goes straight into the fix lane. If you want to re-open the Voyage path afterwards: rebase onto current master without a version bump, include a receipt (curl or recorded doctor run) showing OpenRouter serving voyageai/rerank-2.5-lite on /api/v1/rerank with the results[] envelope, and split or explicitly call out the change that makes the current openrouter Cohere default hard-fail under --max-cost where it previously priced at the pseudo-rate — that is a behavior regression for capped installs and needs its own decision. Appreciate the care that went into this one.

Issue #3856, opened 2026-08-07
pglite->postgres migration leaves the old engine store (2x DB size) on disk forever — no doctor signal, no cleanup path


Masashi-Ono0611:
## Why I'm opening this (human-written intent, per CONTRIBUTING)

> When i was creating my monthly gbrain buckup, i figured out that it had unnecessary storages from previous month. Therefore, i would like to remove it

(That paragraph is written by me, the operator, verbatim. 😀)

## Screenshot
<img width="371" height="268" alt="Image" src="https://github.com/user-attachments/assets/e9759e49-5dbe-4172-8024-0239eca43396" />

## What happened

A brain that migrated `pglite -> postgres` on 2026-05-24 kept BOTH of these on disk, untouched, for 2.5 months:

- `~/.gbrain/brain.pglite/` — the abandoned old engine store (453 MB)
- `~/.gbrain/brain.pglite.pre-migrate-20260524/` — the pre-migration safety copy (another 453 MB)

Together that was **906 MB ≈ 2× the live DB**, and 97% of the `~/.gbrain` directory — with `config.json` reading `"engine": "postgres"` the whole time. Nothing ever surfaced it:

- `doctor` says nothing (engine is healthy, so it is — but the corpse of the previous engine sits right next to it).
- No command offers a cleanup, and no doc says when the pre-migrate copy is safe to drop.
- The mtimes are frozen at the migration minute, so the dirs LOOK important ("some database") to a human skimming `ls`.

## Why it mattered (real cost, not hypothetical)

The `~/.gbrain` directory is what I snapshot, encrypt, and push to permanent paid storage (Arweave) monthly. The leftovers silently rode along in every snapshot: discovering and excluding them cut the snapshot **898 MB -> 459 MB and the monthly push cost roughly in half (~$26 -> ~$13)**. Until the manual dig, that was recurring money paid to permanently archive a dead engine's corpse — twice.

Evidence from the live install (the old dirs also survive in our daily rsync rotation, which is where this capture reads them from — we have since deleted the originals by hand):

```
== oldest kept daily backup (2026-07-25) still carries the abandoned pglite engine dir ==
453M    ~/Backups/gbrain/2026-07-25/brain.pglite

== engine has been postgres since the 2026-05-24 migration ==
"engine": "postgres"
~/Backups/gbrain/2026-07-25/brain.pglite  last-modified=May 24 11:38:42 2026

== workaround: exclude the leftovers from encrypted offsite snapshots ==
[...]

Issue #3697, opened 2026-07-31
Remediation text names commands that do not exist, hard-fail, or falsely report success


time-attack:
Error messages, doc snippets, and CHANGELOG "To take advantage" blocks routinely name commands that do not exist, hard-fail on missing flags, or report success while the problem is still live. Remediation text is the one string nobody ever runs.

## Evidence

From an adversarial review pass over 32 PRs, five separate instances:

- **#3519 and #3569** both print `gbrain reinit-pglite` as the recovery step. `reinit-pglite.ts:212` hard-fails on missing `--embedding-model` / `--embedding-dimensions`, so a user following the hint gets a usage error while their brain is still broken. The correct string already exists in the tree at `pglite-engine.ts:222`. #3519 was rejected for this, and #3569 re-shipped it.
- **#3512** tells the user to set `GBRAIN_E2E_ALLOW_DB`, which `run-e2e.sh:81` unsets.
- **#3517** offers `git ls-files --eol | grep -c 'w/crlf'` as the verification step. It misses `w/mixed`, prints `0`, and tells the contributor the fix worked when it did not.
- **#3545** documents `gbrain purge-deleted`, which is not the command grammar, and a `put_raw_data` example that omits a required parameter.

**And one of mine, which is the clearest case.** In #3618 I wrote a CHANGELOG remediation block telling operators to confirm a leftover process was gone with `gbrain doctor`. Executed against a live holder:

```
pgrep -fl "gbrain serve"  ->  (nothing — the process runs as `bun src/cli.ts serve`)
gbrain doctor             ->  exit 0
                              Overall health score: 95/100. All checks OK (some warnings).
```

`doctor` scored the brain **higher** with the orphan live (95) than without it (85). It has no check for the PGLite filesystem write lock at all — `stale_locks` inspects DB rows with `ttl_expires_at`, which is a different thing. Anyone following my instructions would have seen green and concluded they were fine.

Worth noting: the repo already emits the correct diagnostic. A write command fails with a precise error naming the holding PID and the lock path. The remediation text just pointed somewhere else.

## Why this class persists

Every other kind of string in the repo gets exercised by something. Remediation text is emitted only on a failure path a test rarely reaches, and read only by a user who is already stuck. So it rots silently, and it rots in the worst possible place — the moment someone is trying to recover.

The failure modes are distinct and all present above:
1. the command does not exist
2. the command exists but hard-fails without flags the message omits
[...]

PR #3454, opened 2026-07-27
fix(doctor,docs): warn that the npm name 'gbrain' is unrelated + detect a shadowing npm install (#505)

time-attack:
Closes #505 — thanks to @mdcruz88 for the report.

GBrain is not distributed on npm; the npm package named `gbrain` is an unrelated package. Anyone who reflexively runs `npm i -g gbrain` / `bun add -g gbrain` gets something that is not this project, and it can shadow the real binary on PATH. A defensive npm publish is not possible, so this mitigates on our side: documentation + detection.

## What's in here

- **README.md** — `> [!WARNING]` block at the top of the Install section: gbrain is not on npm, the npm name is unrelated, install/upgrade only via the documented paths (`bun install -g github:garrytan/gbrain`, or `git clone` + `bun install && bun link`), plus the removal commands if the npm install already happened.
- **INSTALL_FOR_AGENTS.md** — same warning adapted for the agent install protocol (Step 1), so an agent following the doc never runs the registry install.
- **`gbrain doctor` check `npm_squat`** (ops category) — runs `which -a gbrain`, classifies every hit via pure helpers in `src/core/npm-squat-check.ts`:
  - `real`: compiled `bun build --compile` binary (native magic bytes), or a script whose nearest `package.json` is this project's (repository field `garrytan/gbrain`, or this repo's `bin: { gbrain: "src/cli.ts" }` shape carried verbatim by `github:` installs / `bun link`).
  - `foreign`: nearest `package.json` named `gbrain` but not this project.
  - `broken`: dangling symlink. `unknown` is fail-open (no false alarms).
  - WARN when a foreign/broken entry wins on PATH (or shadows the real one), with the exact removal + reinstall commands; ok-with-note when a foreign install sits behind the real one; silent skip when `gbrain` isn't on PATH at all.
- **Unit tests** — `test/npm-squat-check.test.ts`, tmp-dir fixtures (fake package.json files, symlinks, magic-byte binaries); no network, no real npm install.
- llms bundles regenerated (README/INSTALL_FOR_AGENTS are inlined into `llms-full.txt`).

## Receipts

- `bun run typecheck` → EXIT=0
- `bun test test/npm-squat-check.test.ts` → EXIT=0 (14 pass / 0 fail)
- `bun test test/doctor-categories.test.ts` → EXIT=0 (drift guard picks up the new check name)
- `bun test test/doctor-home-dir-in-worktree.test.ts` → EXIT=0 (neighboring doctor section)
[...]

PR #3879, opened 2026-08-08
v0.42.74.1 fix(doctor): surface abandoned PGLite stores after an engine migration (#3856)

Masashi-Ono0611:
Addresses ask 1 of #3856 (detection only — the cleanup command and the retention policy for pre-migration copies, asks 2–3, stay maintainer decisions and are deliberately NOT in this PR).

> When i was creating my monthly gbrain buckup, i figured out that it had unnecessary storages from previous month. Therefore, i would like to remove it.
>
> In this PR, the issue should be detected by doctor.

## What

In the default layout, a `pglite -> postgres` migration leaves `brain.pglite/` (the old engine's store) under the gbrain home forever — dead weight that nothing surfaces, silently riding along in any backup that archives the home dir. On the reporting brain the abandoned store (together with a sibling copy created by my own external backup tooling) totaled roughly **2× the live DB** and was paid onto Arweave for 2.5 months (898 MB → 459 MB snapshot, ~half the monthly cost, once excluded by hand). Details + live receipts on #3856.

New warn-only doctor check, `pglite_leftovers`, in the npm-squat shape (#505 / #3454):

- **Pure helper** `src/core/pglite-leftovers-check.ts` (filesystem-only, no network, no shelling out) + a fail-open integration in `doctor.ts` + `OPS_CHECK_NAMES` registration.
- **Warns only when the DURABLE file config says `engine: "postgres"`** — the engine is read straight from `config.json`, not `loadConfig()`, because a transient `DATABASE_URL` (#427) can make a live PGLite brain resolve as postgres for one process, and deletion advice must never rest on an env override. `pglite` means the store is live; anything else (missing/unreadable config, future engines) skips.
- **Silent while a migration is in flight**: if `migrate-manifest.json` exists under the gbrain home, the check returns `skip` without assessing anything. With an interrupted `postgres -> pglite` migration the durable engine is still `postgres` while `brain.pglite` is the LIVE migration target (#3194: the engine only flips on clean completion), and deleting it would strand the manifest's completed set so a re-run silently skips those pages. In that state no deletion advice is ever produced. (Added after review — credit to the adversarial triage pass for the scenario.)
- **Scope is exactly `brain.pglite`** — the default-layout store directory (a brain initialized with a custom `--path` keeps its store elsewhere and is simply never claimed). ALL `brain.pglite.*` siblings (hand-made `.bak` copies, external backup-tooling artifacts, etc.) have unknown provenance and are not claimed. (An earlier revision also claimed `brain.pglite.pre-migrate-*`; review established that no version of gbrain ever creates that name — it came from my own ops tooling — so the pattern was dropped.)
- **The walk cannot lie, and is bounded against pathological store trees**: iterative `opendirSync` streaming with one shared assessment-wide budget (20k entries), never follows symlinks (top-level or nested), so an ordinary layout cannot pull unrelated data into the count, and any unreadable entry marks the size as an explicit floor (`>=`, rounded **down**) instead of reporting an unreadable multi-GB store as `0 B`.
- **The message sticks to what the walk measured**: per-dir floor-marked size and the directory's own mtime labeled `dir mtime` (file contents can change without touching it) — and a **manual** remediation that tells the operator to verify the postgres data first, then delete by hand, precisely because the check has not verified provenance or reclaimability itself. No CLI command is named that does not exist (#3697), and the check itself never deletes anything.

## Tests

13 unit tests (`test/pglite-leftovers-check.test.ts`, tmp-dir fixtures, npm-squat test shape): manifest-present → skip (brain.pglite may be the live migration target), name-scope matching (`brain.pglite` only), a pre-migrate-named sibling never claimed and alone yields ok, all fail-open skips (pglite engine / unknown engine / missing home), ok path on a clean postgres brain, file-not-a-dir, symlinked store not claimed, the abandoned-store warn signature with receipts asserted and siblings staying unclaimed (including a guard that the message never invents a `gbrain cleanup`-style command), nested-symlink escape guard, unreadable-subdir incomplete floor (the mode-000 `readSync` EACCES path is measured, not assumed), and the injectable walk budget.

`bun run typecheck` clean; `test/doctor-categories.test.ts` + `test/doctor-brain-checks-score.test.ts` pass with the new registration.


Masashi-Ono0611:
# screenshot
Screenshot 2026-08-08 at 2 34 58 PM

Masashi-Ono0611:
Applied this to a production brain (postgres engine, 2.7k pages) ahead of review:

- On the real brain (leftovers already hand-deleted per #3856): `[OK] pglite_leftovers: No abandoned PGLite store under the gbrain home (engine: postgres).`
- On a scratch GBRAIN_HOME seeded with brain.pglite + brain.pglite.pre-migrate-20260524: the WARN fires with per-dir sizes, dir mtime, and the manual-only remediation.

Both sides exercised live, not just in the unit suite.

time-attack:
Automated triage: tests green (108 pass targeted + verify 34/34), but both independent adversarial review passes found the same real defect before merge: with an interrupted postgres->pglite migration, durable engine stays `postgres` while `brain.pglite/` is the LIVE migration target (config only flips on a clean run, #3194) — this check then labels it a deletable leftover. If the user deletes it, the surviving `migrate-manifest.json` makes the re-run silently skip every previously-completed page, producing an incomplete brain. Also: `brain.pglite.pre-migrate-*` is a directory name no version of gbrain appears to have ever created (full-history pickaxe found no creation site), so the message's provenance claim is unverified. Suggested fix: skip or soften the check when `migrate-manifest.json` exists, and drop/justify the pre-migrate pattern. Routed to maintainer review rather than auto-merge.

Masashi-Ono0611:
Both findings were correct — thank you for the executed probes, this was exactly the review this check needed.

Fixed in 56157e35:

1. **Interrupted-migration false positive**: the check now returns `skip` whenever `migrate-manifest.json` exists under the gbrain home. Verified against migrate-engine.ts: the manifest is only cleared — and the durable engine only flips — on clean completion (#3194 round 2), so while the manifest exists, `brain.pglite` can be the live migration target even with durable engine=postgres. In that state the check refuses to assess leftovers at all (no deletion advice, ever). Pinned by a new fixture: manifest + brain.pglite both present → skip, and the message contains no deletion language.

2. **Pre-migrate pattern dropped**: you are right that no version of gbrain creates `brain.pglite.pre-migrate-*` — the directory on my brain came from my own external ops script, and my header comment's provenance claim was wrong. The check now claims exactly `brain.pglite` and nothing else; all `brain.pglite.*` siblings are documented as unknown-provenance and out of scope. Pinned by tests: a pre-migrate-named sibling is never claimed, and alone yields ok.

13 tests pass (48 asserts), typecheck clean. PR body updated to match the narrowed scope.

time-attack:
Merged in the v0.45.8.0 community fix-wave (#4050). Your fix landed on master as e795324ec5, credited to you via Co-Authored-By, and is called out in the v0.45.8.0 CHANGELOG entry. Thank you for the contribution!

PR #3888, opened 2026-08-08
v0.42.74.1 fix(doctor): parse flow-style skill triggers (#3887)

### LINKED ISSUE #3887: doctor: MECE_GAP remediation says "Add a triggers: array" for skills that already have one
### Summary

`gbrain doctor`'s `resolver_health` check emits `MECE_GAP` findings whose remediation text tells you to add a `triggers:` array to skills **that already have one**. Following the ACTION verbatim is a guaranteed no-op.

### Version

`gbrain 0.42.74.0` (commit `a948dfd6`), Postgres backend, Linux arm64.

### What happens

```
[WARN] resolver_health: 3 issue(s): 0 error(s), 3 warning(s)
  → MECE_GAP: capability-audit
    ACTION: Add a triggers: array to the frontmatter of skills/capability-audit/SKILL.md
  → MECE_GAP: speaking-prep
    ACTION: Add a triggers: array to the frontmatter of skills/speaking-prep/SKILL.md
  → MECE_GAP: transcript-insights
    ACTION: Add a triggers: array to the frontmatter of skills/transcript-insights/SKILL.md
```

All three skills already declare `triggers:`. A representative one (content generalised):

```yaml
---
name: transcript-insights
description: "Extract insights and actionable takeaways from video/audio transcripts."
platforms: [linux, macos, windows]
triggers: [analyze this transcript, podcast takeaways, video transcript insights, interview
    takeaways, policy mechanisms from transcript]
---
```

`skill_conformance` passes for the same skills in the same run (`3/3 skills pass`), so the frontmatter is being read successfully somewhere.

### Why this matters

The ACTION is the only guidance the check gives, and it does not describe what `MECE_GAP` actually requires. Acting on it literally means adding a key that is already present — the warning persists, and the natural conclusion is either "the fix didn't work" or, worse, "I fixed it" while the score is unchanged. Since `doctor --fast` is a common thing to script a health score off, a permanently-unfixable warning quietly costs 10 points with no actionable path.

Concretely: `resolver_health` warnings put Skill checks at 90/100 and `doctor --fast` at 85/100, with no documented way to reach 100 short of understanding what MECE coverage means.

### Ruled out

- **Not the CRLF/`extractTriggers` issue from #1150** (closed). These files are LF — verified with `cat -A` (lines terminate `$`, not `^M$`) and the frontmatter fence is `---\n` (`od -c` → `-   -   -  \n`).
- **Not YAML flow-sequence wrapping.** Reflowing a wrapped `triggers: [...]` onto a single line did not change the finding. One of the three (`speaking-prep`) is already a single-line quoted flow sequence and is flagged identically.

### What would help

Either:

1. Make the ACTION describe the actual requirement — what does `MECE_GAP` need beyond a present `triggers:` array? (Non-overlapping coverage? A minimum trigger count? A `RESOLVER.md` routing table entry?) — or
2. If the array *is* present but failing to parse, say that specifically (`triggers: present but not parsed`) rather than `Add a triggers: array`, so the two failure modes are distinguishable.

A short note in the docs on what MECE coverage means for a small skill set would also resolve it. If a handful of narrow skills simply cannot be "collectively exhaustive" by construction, saying so would turn a permanent warning into an understood one.

### Secondary observation (reported as observed, not characterised)

The check does not appear deterministic. On the same install, with the same files and no intervening changes, consecutive `gbrain doctor` runs sometimes report the 3 `MECE_GAP` findings and sometimes report `[OK] resolver_health: 0 skills, all reachable`. I could not pin down what flips it, so I am noting it rather than claiming it — happy to gather more if useful.

RerankerGuo:
## Summary

- parse SKILL.md array fields with the existing js-yaml dependency, including single-line and wrapped flow sequences
- route doctor trigger extraction through the shared skill frontmatter parser so it cannot disagree with the trigger index
- preserve CRLF normalization and the tolerant regex fallback for partially malformed frontmatter

Closes #3887.

## Root cause

`resolver_health` used a separate `extractTriggers()` regex that only accepted block-list syntax. Valid YAML such as `triggers: [analyze this transcript, podcast takeaways]`, including a flow sequence wrapped across lines, therefore returned an empty array. The gap detector then emitted `MECE_GAP` and the incorrect action to add a field that already existed, while other skill checks used the shared parser and succeeded.

## Evidence

Before the fix, the focused test file produced 41 pass / 2 fail: both single-line and wrapped flow-sequence trigger lists returned `[]`.

After the fix:

- `bun test test/check-resolvable.test.ts test/skill-brain-first.test.ts test/skill-trigger-index.test.ts test/check-resolvable-openclaw-compact.test.ts`: 125 pass, 0 fail
- `bun run typecheck`: pass
- `bun run build:llms`: pass, no generated diff
- `git diff --check`: pass

## Scope

This does not change MECE coverage rules, warning severity, doctor scoring, or auto-fix behavior. It only makes the existing SKILL.md array fields accept valid YAML flow syntax consistently.


time-attack:
Merged in the v0.45.10.0 community fix-wave 2 (#4085). Your fix landed on master as 136fc109c1, credited to you via Co-Authored-By, and is called out in the v0.45.10.0 CHANGELOG entry. Thank you for the contribution!

Issue #3877, opened 2026-08-08
Feature: positive slug-prefix candidate restriction (pre-HNSW) + one-embedding multi-lane merge for curated/authority subsets


vinsunfeng:
**Tested versions:** 0.42.73.2 (15b9863d) and 0.42.74.0 (a948dfd6) — identical retrieval behavior on this surface.

### Use case (generic)
I have a small curated subset of pages — authoritative reference/definition documents — that I want to surface ALONGSIDE regular search results. They are identifiable by a **slug prefix** (e.g. `docs/<topic>/`). The problem: their semantic similarity to many queries is moderate (cosine ~0.4–0.5), so they lose rank competition against high-volume event/incident pages and never enter the HNSW candidate pool (the inner `hnsw_candidates` CTE's `ORDER BY ... LIMIT`). A post-HNSW source boost cannot recover chunks the pool never selected.

### Observed API/source behavior
- `SearchOpts` exposes `exclude_slug_prefixes` (subtractive hard-exclude) and `include_slug_prefixes` (opt-back-in FROM the hard-exclude set). Neither is a **positive** prefix inclusion that restricts candidates TO a prefix set.
- `SearchOpts.sourceId/sourceIds` IS applied pre-HNSW inside the `hnsw_candidates` CTE on both engines (`AND p.source_id = ANY(...)`) — but it filters by `source_id`, and my curated subset is distinguished by slug prefix, not source. So `sourceIds` cannot select it.
- `searchVector(embedding: Float32Array, opts?)` already accepts a precomputed embedding, but `hybridSearch` embeds the query once internally and runs a single vector lane (plus separate multimodal/image branches). There is no public way to compose a second, prefix-restricted vector lane on the SAME embedding while preserving the native RRF/dedup/rerank/cache/fallback pipeline.

### Expected capability
One or both of:
1. A positive prefix restriction on `SearchOpts` (e.g. `restrict_slug_prefixes?: string[]`) applied pre-HNSW in the `hnsw_candidates` CTE on both Postgres and PGlite — so a subset can be queried directly.
2. A supported way to run two vector lanes (general + prefix/source-restricted) on a single embedded query and merge them through the native ranking/dedup/cache/fallback pipeline, ideally with a relevance floor so only genuinely related subset pages are admitted (avoiding false surface on unrelated queries).

### Questions
[...]

PR #3890, opened 2026-08-08
feat(search): add configurable reserved-lane retrieval

vinsunfeng:
> Refs #3877.

### Problem
A curated slug-prefix subset (reference pages, authoritative definitions — any caller-chosen prefix) often loses rank competition against high-volume pages and never enters the HNSW candidate pool, so it can't surface alongside regular results. A post-HNSW source boost can't recover chunks the pool never selected.

### What this adds (all additive, default-OFF)

**Engine primitives (usable independently):**
- `SearchOpts.restrict_slug_prefixes?: string[]` — positive slug-prefix inclusion applied **pre-HNSW** in the inner `hnsw_candidates` CTE on both Postgres and PGlite (`buildRestrictClause`, sibling of `buildHardExcludeClause`).
- `SearchOpts.min_raw_score?: number` — configurable **raw-cosine** floor applied in the `scored` CTE **before** the source-factor multiplier (true raw-cosine gate, independent of `GBRAIN_SOURCE_BOOST`).

**Composition (one-embedding bounded-lane merge):**
- `HybridSearchOpts.reservedLanePrefixes?: string[]` — non-empty enables a second vector lane ("Lane B") on the **same query embedding** (one-embed invariant: reuses the precomputed vector, no re-embed). Also configurable via `GBRAIN_RESERVED_LANE_PREFIXES` env (comma-separated, deployment-supplied — **no application-specific literal** in production source).
- `HybridSearchOpts.reservedLaneMinRawScore?: number` — per-call floor override (default `GBRAIN_RESERVED_LANE_FLOOR=0.5`, conservative placeholder pending pilot).
- Prefix resolution: trims whitespace, deduplicates, validates (rejects absolute `/`, traversal `..`, glob `*`/`?`, empty). Invalid ⇒ lane disabled (fail-closed).
- **Bounded merge**: at most **one** deduplicated relevant candidate in a **reserved tail slot** (never rank-0, `limit>=2`, replaces only the Lane-A tail when full, `limit<2` ⇒ Lane-A only). Lane-A top + relative order preserved. Lane-B failure (`try/catch`) ⇒ exact Lane-A.
- Reserved-lane-on **skips the query cache** (Option B — `skipCache` predicate; `KNOBS_HASH_VERSION` unchanged).

### Primitive vs composition — maintainer may accept or split
This PR includes both the engine primitives (`restrict_slug_prefixes`, `min_raw_score`) AND the composition (`reservedLanePrefixes` → bounded-tail merge in `hybridSearch`). If preferred, the primitives can land first (they're independently useful) and the composition in a follow-up.

### Compatibility / rollback
- Default OFF (empty/unset prefixes) ⇒ **byte-identical legacy** SQL and behavior. `min_raw_score` unset ⇒ empty clause. No `mode.ts`, no schema/migration, no re-embed.
- Rollback: `git revert` (single commit) or image swap to prior — no DB state to undo.

### Tests
- **32 serial** (both PGLite + Postgres where the engine seam applies): cache-skip + `KNOBS=15`; Lane-B throw → exact Lane-A + one-embed spy; bounded-merge (irrelevant exclusion, boosted-irrelevant exclusion, relevant survival in reserved tail, Lane-A top+order, bounded one-candidate, `limit<2`); genericization (arbitrary `trusted/` prefix, two-prefix deterministic, env parsing, whitespace/dedup normalization, `reservedLaneMinRawScore` configurable, empty-legacy, invalid fail-closed, src grep no-literal contract).
- **12-case seam** (positive-prefix filter + raw-score floor) × PGLite + Postgres (parity).
- **75 focused regressions** (query-cache-knobs-hash, sql-ranking, search-lang-symbol-kind).
- `tsc --noEmit` exit 0; `verify` 34/34 (typecheck/privacy/jsonb/source-id/test-isolation).

### serve-http E2E gap
`serve-http-takes-holders` e2e requires a running `gbrain serve --http` + brain; it is **skipped** in the dev harness (environmental). This patch has **zero overlap** with `serve-http`/`takes_holders`/`oauth-provider` (the v0.42.74.0 security files). The e2e **must run at deployment smoke** in a serve-http environment before go-live.

### Questions for maintainer
1. Naming: `reservedLanePrefixes` — acceptable, or prefer `governedLanePrefixes` / `authorityLanePrefixes`? Optionally, the test filenames currently use `governed-*` development names (`_governed-lane-cases.ts`, `governed-source-lane*.test.ts`, `governed-lane.serial.test.ts`) — happy to rename to `reserved-*` if preferred for consistency.
2. `min_raw_score`: keep as a general `SearchOpts` (any caller can floor), or scope to the reserved lane only?
3. Bounded merge: "at most one, reserved tail, Lane-A preserved" — acceptable, or prefer configurable candidate count?
4. Default floor: `GBRAIN_RESERVED_LANE_FLOOR=0.5` placeholder — OK to ship conservative pending pilot, or leave unset?

### File manifest (9 files, 980+/6-)
| file | status |
|------|--------|
| src/core/types.ts | M |
| src/core/search/sql-ranking.ts | M |
| src/core/postgres-engine.ts | M |
| src/core/pglite-engine.ts | M |
| src/core/search/hybrid.ts | M |
| test/_governed-lane-cases.ts (shared seam factory) | A |
| test/governed-source-lane.test.ts | A |
| test/governed-source-lane-postgres.test.ts | A |
| test/search/governed-lane.serial.test.ts | A |

### DIFF
--- a/src/core/pglite-engine.ts
+++ b/src/core/pglite-engine.ts
@@ -77,7 +77,7 @@ import { GBrainError, PAGE_SORT_SQL, ENRICH_ORDER_SQL } from './types.ts';
 import { finalizeLastSeen } from './chronicle/last-seen.ts';
 import { computeAnomaliesFromBuckets } from './cycle/anomaly.ts';
 import { resolveBoostMap, resolveHardExcludes } from './search/source-boost.ts';
-import { buildSourceFactorCase, buildHardExcludeClause, buildVisibilityClause, buildRecencyComponentSql, buildBestPerPagePoolCte, buildOrFallbackWebsearchQuery } from './search/sql-ranking.ts';
+import { buildSourceFactorCase, buildHardExcludeClause, buildRestrictClause, buildVisibilityClause, buildRecencyComponentSql, buildBestPerPagePoolCte, buildOrFallbackWebsearchQuery } from './search/sql-ranking.ts';
 import { unverifiedExtractionFragment } from './extraction-review.ts';
 import { shouldExcludeFromOrphanReporting, loadOrphanPolicyOverrides } from './orphan-policy.ts';
 import { LINK_EXTRACTOR_VERSION_TS } from './link-extraction.ts';
@@ -2109,9 +2109,19 @@ export class PGLiteEngine implements BrainEngine {
     const sourceFactorCaseOnSlug = buildSourceFactorCase('slug', boostMap, opts?.detail, 'unverified_stub');
     const hardExcludePrefixes = resolveHardExcludes(opts?.exclude_slug_prefixes, opts?.include_slug_prefixes);
     const hardExcludeClause = buildHardExcludeClause('p.slug', hardExcludePrefixes);
+    // Governed-authority lane: positive PRE-HNSW prefix inclusion (mirror of
+    // postgres-engine). Empty ⇒ '' ⇒ no-op ⇒ byte-identical legacy SQL.
+    const restrictClause = buildRestrictClause('p.slug', opts?.restrict_slug_prefixes);
     const innerLimit = offset + Math.max(limit * 5, 100);
 
     const params: unknown[] = [vecStr, innerLimit, limit, offset];
+    // min_raw_score: raw-cosine floor in the scored CTE (pre-sourceFactor, both
+    // engines) — true raw-cosine gate, independent of source boosts. Mirror of postgres.
+    let rawScoreFloorCte = '';
+    if (opts?.min_raw_score != null && Number.isFinite(opts.min_raw_score)) {
+      params.push(opts.min_raw_score);
+      rawScoreFloorCte = `WHERE raw_score >= $${params.length}`;
+    }
     let extraFilter = '';
     if (opts?.language) {
       params.push(opts.language);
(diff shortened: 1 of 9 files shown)


time-attack:

time-attack:
Thanks for this — it's genuinely one of the better-constructed external PRs in the queue: both-engine parity, escaping routed through the existing LIKE-literal builder, default-OFF byte-identical SQL, cache-skip instead of a knobs-hash bump, and a real parity suite on both engines. None of what follows is about the quality of the work.

We hold a default-close bar on new product surface here, and this adds three new retrieval knobs plus a new hybridSearch composition lane — permanent API surface the repo then owns forever. On top of that, anything that touches the retrieval path (hybrid.ts, sql-ranking.ts, both engines' searchVector) has to clear the eval bar before it can land: `gbrain eval replay` receipts (Jaccard@k, top-1 stability, latency delta), not just a green unit suite. And at 980 lines across three distinct knobs, this is three concerns in one PR — I can't honestly review it as a unit.

The right sequencing is the one you already set up: #3877 is the design conversation, and the four open questions there (naming, whether min_raw_score is general SearchOpts or lane-scoped, the merge bound, the default floor) are exactly the calls that have to be made before code. Closing this PR while that discussion is open — if the answer there is yes to any piece, a resubmission following that guidance, one knob per PR with eval-replay receipts attached, goes straight into the review lane. That path has landed before and I'd be glad to see it here.

PR #3891, opened 2026-08-08
v0.42.74.1 fix(sync): acknowledge auto-skipped failures (#3829)

### LINKED ISSUE #3829: doctor recommends `sync --skip-failed` for auto_skipped failures, but --skip-failed only acknowledges open rows
The doctor `sync_failures` check says: "run `gbrain sync --skip-failed` on the host to acknowledge." But `acknowledgeFailures()` in src/core/sync-failure-ledger.ts skips any row whose state is not `open`. Rows in `auto_skipped` state are exactly the ones doctor keeps warning about, and no CLI command can clear them.

Repro on my brain: 7 failures, all state=auto_skipped. Ran `gbrain sync --skip-failed` and `gbrain sync --all --skip-failed`. Neither printed an "Acknowledged" line, count stayed 7, doctor kept warning.

Second dead end on top: my failures belong to the default source, which is DB-only (no local_path). The other documented exit from auto_skipped is "the file imports cleanly on a future sync," which can never happen for a repo-less source. So the rows are permanently stuck and the only fix is hand-editing ~/.gbrain/sync-failures.jsonl.

Suggested fixes, either would do:
1. Let --skip-failed acknowledge auto_skipped rows too (a human running it is expressing exactly that intent), or
2. Add an explicit `gbrain sync failures ack` that works without a repo, and update the doctor fix text.

gbrain 0.42.73.2, Windows 11.

RerankerGuo:
## Summary

- let the existing `gbrain sync --skip-failed` path acknowledge chronic `auto_skipped` rows as well as fresh `open` rows
- preserve source scoping and the sentinel-path refusal
- document the explicit human-ack transition in the ledger state machine

Closes #3829.

## Root cause

`doctor` tells operators to run `gbrain sync --skip-failed` for unresolved failures, including `auto_skipped` rows. That flag delegates to `acknowledgeFailures()`, but the function skipped every state except `open`. A chronic failure therefore remained doctor-visible forever, especially for DB-only sources that have no repository path to re-import successfully.

## Evidence

The regression uses the real ledger transition (`recordFailures` -> `autoSkipFailures` -> `acknowledgeFailures`) for two sources.

Before the fix:

- `bun test test/sync-failure-ledger.serial.test.ts`: 23 pass, 1 fail
- scoped acknowledgment returned `count: 0` and left the target row `auto_skipped`

After the fix:

- `bun test test/sync-failure-ledger.serial.test.ts test/sync-failures.test.ts`: 84 pass, 0 fail
- `bun run typecheck`: pass
- `bun run build:llms`: pass, no generated diff
- `git diff --check`: pass

## Scope

The behavior change is one state predicate. It does not alter auto-skip thresholds, bookmark advancement, sentinel handling, or cross-source isolation. Open PR #3583 also touches `sync-failure-ledger.ts`, but only inserts a separate restore helper before this function; it does not change acknowledgment semantics.


time-attack:
Merged in the v0.45.10.0 community fix-wave 2 (#4085). Your fix landed on master as fd0e371d5b, credited to you via Co-Authored-By, and is called out in the v0.45.10.0 CHANGELOG entry. Thank you for the contribution!

PR #3892, opened 2026-08-08
v0.42.74.1 fix(extract): summarize wrapped inline citations by paragraph (#3664)

### LINKED ISSUE #3664: An inline [Source: ..., YYYY-MM-DD] citation materialises a timeline entry whose summary is a prose fragment
## Summary

An inline `[Source: ...]` citation ending in a date — the format `skills/_brain-filing-rules.md:81` makes mandatory for every fact — materialises a timeline entry whose summary is the surrounding prose, sometimes a mid-sentence fragment.

## Measured

A page was written following `skills/_brain-filing-rules.md:78`: *"Every fact written to a brain page must carry an inline `[Source: ...]` citation"*, in the documented shape `[Source: User, {context}, YYYY-MM-DD]`.

After `gbrain sync`, `get_timeline` on that page returned four rows. Two were the intended bullets from its `## Timeline` section and parsed correctly. Two were not written as timeline entries at all:

```json
{"id":32,"date":"2026-07-30T00:00:00.000Z","source":"User, session",
 "summary":"Not written by us."}
{"id":33,"date":"2026-07-30T00:00:00.000Z","source":"User, session",
 "summary":"product fit for commercial use. That app is"}
```

Row 33's summary is a fragment cut mid-sentence.

Both trace to prose lines of the shape:

```
<sentence>. [Source: User, session, 2026-07-30]
```

A citation written on its own line — `[Source: git log, both repos, 2026-07-30]` — did **not** produce a row. So the trigger appears to involve a citation trailing prose on the same line, though I have not isolated which element matters.

## What I have not established

- Whether extracting dated citations into the timeline is intended. If it is, the fragment summary still looks wrong.
- The exact parse rule. I measured which inputs produced rows, not why.
- Whether removing the prose removes the row. I removed unrelated test content and re-synced; these two rows persisted, but that test does not isolate reconciliation behaviour.

## Why it matters

Two rules in the same document pull against each other. Following the mandatory citation rule on a page that also has a `## Timeline` section silently adds entries the author did not write, and a reader of the timeline cannot tell them apart from real ones.

## Environment

gbrain 0.42.67.0, commit c6dc0adf26a2d20df1147d2ec87c8922ca86d410, Postgres engine, single markdown source.

javieraldape:
## Summary

Fixes #3664 by making inline citation timeline extraction summarize the surrounding paragraph instead of only the physical line containing `[Source: ..., YYYY-MM-DD]`.

Previously, a wrapped sentence could create a timeline row whose summary was just the second half of the sentence. That made normal citation-following pages produce misleading dated fragments.

## Changes

- Adds a shared inline-citation timeline parser that groups markdown paragraph blocks before stripping citation markers.
- Keeps the existing skip behavior for timeline bullets that already produce a Format 1 entry, so a bullet with its own citation is not double-filed.
- Reuses the shared parser from the filesystem extraction path so db-source and fs-source citation extraction stay aligned.
- Adds focused regression coverage for wrapped citation paragraphs in both extraction surfaces.

## Tests

- Extends `test/link-extraction.test.ts` for the db-source/ingest parser.
- Extends `test/extract.test.ts` for `extractTimelineFromContent`.

## Verification

- `bun test test/extract.test.ts test/link-extraction.test.ts` - 203 pass, 0 fail, 471 expect calls.
- `bash scripts/gbrain-gate.sh <worktree> test/extract.test.ts test/link-extraction.test.ts` - `RESULT: GREEN`; verify green; focused unit tests green; diff-mapped E2E green with 19 pass, 0 fail.

### DIFF
--- a/src/commands/extract.ts
+++ b/src/commands/extract.ts
@@ -39,6 +39,7 @@ import {
   extractFrontmatterLinks, isGlobalBasenameEnabled, LINK_EXTRACTOR_VERSION_TS,
   WIKILINK_BASENAME_LINK_TYPE,
   buildBasenameIndex, queryBasenameIndex, stripCodeBlocks,
+  parseInlineCitationTimelineEntries,
   type UnresolvedFrontmatterRef, type LinkCandidate,
 } from '../core/link-extraction.ts';
 import { createProgress } from '../core/progress.ts';
@@ -566,23 +567,11 @@ export function extractTimelineFromContent(content: string, slug: string): Extra
   // carries its own [Source: ...] citation, and re-extracting it would file
   // a duplicate entry under a different (source, summary) shape that the
   // DB-level uniqueness cannot collapse.
-  const citationPattern = /\[Source:\s*([^\]]+?),\s*(\d{4}-\d{2}-\d{2})\s*\]/g;
   const bulletLinePattern = /^-\s+\*\*\d{4}-\d{2}-\d{2}\*\*\s*\|/;
-  for (const line of content.split(/\r?\n/)) {
-    if (bulletLinePattern.test(line)) continue;
-    const lineMatches = [...line.matchAll(citationPattern)];
-    if (lineMatches.length === 0) continue;
-    // Strip every citation marker from the line to leave the annotated text.
-    const summary = line
-      .replace(/\[Source:[^\]]*\]/g, '')
-      .replace(/^[-*>#\s]+/, '')
-      .replace(/\s+/g, ' ')
-      .trim()
-      .slice(0, 300);
-    if (!summary) continue; // a bare citation with no surrounding text is not an event
-    for (const m of lineMatches) {
-      entries.push({ slug, date: m[2], source: m[1].trim().slice(0, 200), summary });
-    }
+  for (const entry of parseInlineCitationTimelineEntries(content, {
+    skipLine: (line) => bulletLinePattern.test(line),
+  })) {
+    entries.push({ slug, date: entry.date, source: entry.source, summary: entry.summary });
   }
 
   return entries;
--- a/src/core/link-extraction.ts
+++ b/src/core/link-extraction.ts
@@ -14,7 +14,12 @@
 import type { BrainEngine } from './engine.ts';
 import type { PageType, EffectiveDateSource } from './types.ts';
 import { ensureWellFormed } from './text-safe.ts';
+import { stripCodeBlocks } from './markdown-code.ts';
+import { parseInlineCitationTimelineEntries } from './timeline-citations.ts';
 import { slugifyPath } from './sync.ts';
+
+export { stripCodeBlocks } from './markdown-code.ts';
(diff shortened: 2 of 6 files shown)


time-attack:
Triage bot: CI is green and the wrapped-paragraph case works, but an adversarial regression pass found a real data-loss case this introduces. The skip semantics changed from skip-the-matching-LINE to skip-the-whole-PARAGRAPH: a prose continuation line directly under a timeline bullet now joins the bullet's paragraph and the ENTIRE paragraph is skipped, so its citation — which master extracts — silently vanishes. A/B-verified repro:

```markdown
- **2025-03-18** | Meeting notes
Follow-up decision recorded. [Source: memo, 2025-03-20]
```

master (fs path): extracts the 2025-03-20 entry. This PR: `[]`. Same drop on the db path (2 entries → 1). Code-fence content can also leak into paragraph summaries (`citationParagraphs` doesn't strip fenced blocks; the old per-line matcher rarely hit them, and `parseTimelineEntries`' bullet pass runs `stripCodeBlocks` for the bullet formats but this citation pass reads raw content). Flagging for changes rather than merging: probably flush-paragraph-on-skip-line (drop only the matching line, keep the rest of the paragraph) plus stripping code fences before paragraph assembly.

javieraldape:
Addressed the triage finding in 8106b20b.

What changed:
- Inline citation paragraph assembly now strips fenced/inline code before scanning for `[Source: ..., YYYY-MM-DD]` citations.
- Skip semantics now drop only the matching timeline-bullet line and flush the current paragraph, so a prose citation directly below a timeline bullet is still extracted instead of losing the whole paragraph.
- Added regression coverage for both `parseTimelineEntries` and the filesystem `extractTimelineFromContent` path.

Validation:
- `bun test test/link-extraction.test.ts test/extract.test.ts` (207 pass)
- `bun run typecheck`
- `GBRAIN_VERIFY_TIMEOUT=240 bun run verify` (34/34 pass)

time-attack:
Trial-merge is clean and typechecks, but on the merged state this PR's own test 'keeps a prose citation directly under a timeline bullet' fails (extract.test.ts) — the v0.46.23.0 relative-link changes altered the surrounding behavior. Please rebase and adjust; fixes #3664 is still wanted.

javieraldape:
Maintenance update after rebasing onto current `master`.

What changed:
- Updated the filesystem timeline regression to account for current `master` retaining delimiterless timeline bullets while still preserving the prose citation directly below the bullet.
- Moved markdown code stripping and inline citation timeline parsing into sibling core modules so `src/core/link-extraction.ts` stays below the module-size ratchet.

Validation:
- `env -u DATABASE_URL -u GBRAIN_DATABASE_URL bun test test/extract.test.ts test/link-extraction.test.ts` (211 pass)
- `env -u DATABASE_URL -u GBRAIN_DATABASE_URL bun run typecheck`
- `env -u DATABASE_URL -u GBRAIN_DATABASE_URL GBRAIN_VERIFY_TIMEOUT=240 bun run verify` (54/54 pass)
- GitHub checks on `013cf744` reached terminal state with all non-skipped checks passing.

time-attack:

PR #3898, opened 2026-08-08
fix(search): tokenize multi-word CJK queries instead of matching the whole string

yangwhale:
## The bug

`PGLiteEngine._searchKeywordCJK` binds the ENTIRE query as one indivisible
`ILIKE` substring. Any multi-token query containing CJK returns zero rows
unless the corpus holds that exact string verbatim, spaces included.

```
query "记忆 系统"   ->  ILIKE '%记忆 系统%'   ->  0 rows
```

Chinese prose writes `记忆系统` with no space, so nothing matches. On a brain
with ~350 pages of Chinese content, every two-word query came back empty.

Single tokens work (`记忆` alone is a valid substring), which is why this went
unnoticed: it only bites once a user types two words, and it surfaces as an
empty array. Empty reads as "nothing in the brain", not as "the tokenizer
cannot express this query". There is no error, no warning, no log line.

Mixed queries route here too, since `hasCJK("cron 定时")` is true, so the
ASCII half is collateral damage on the same path.

Reproduction on a brain containing the text `记忆系统需要定期回收`:

| query | before | after |
|---|---|---|
| `记忆` | hit | hit |
| `记忆 系统` | **0 rows** | hit |
| `cron 定时` | **0 rows** | hit |
| `memory system` (ASCII, FTS path) | hit | hit |

## The fix

Each whitespace-separated token becomes its own `ILIKE` conjunct, and the
per-token bigram-frequency scores sum.

AND rather than OR: with OR a two-token query would be strictly noisier than
either token alone, which is the wrong direction for a query the user made
*more* specific.

Single-token queries produce byte-identical arithmetic to before, so existing
CJK behavior is untouched. Both bindings per token are preserved from the
original codex C8 fix (LIKE-escaped for the conjunct, raw for the
`position()`/`replace()` ranking), and `ESCAPE '\'` still applies per conjunct.
Parameter numbering is now derived from token count, so `limit`/`offset` and
the `extraFilter` pushes downstream stay correctly indexed.

Postgres engine is untouched. It goes through `websearch_to_tsquery`, which is
a separate problem (needs pgroonga / zhparser) and out of scope here.

## Tests

Five cases added to `test/pglite-engine.test.ts`, all failing before this change:

- multi-token CJK against spaceless prose
- mixed CJK+ASCII (`cron 定时`)
- AND semantics (a chunk matching only one token is excluded)
- score summation (denser match outranks sparser)
- three-token query, which guards the dynamic `$N` numbering

Verified by mutation rather than by a green check:

- reverting the tokenizer (splitting on `\x00` instead of `\s+`) turns exactly
  those five red, 100 pass / 5 fail
- flipping `AND` to `OR` turns the AND-semantics case red **and** the
  pre-existing `LIKE-meta-char escape` case red, 103 pass / 2 fail

`114 pass / 0 fail` in `test/pglite-engine.test.ts` with the fix in place.
`bun run typecheck` clean.

### DIFF
--- a/src/core/pglite-engine.ts
+++ b/src/core/pglite-engine.ts
@@ -1874,18 +1874,48 @@ export class PGLiteEngine implements BrainEngine {
     },
   ): Promise<SearchResult[]> {
     const { limit, offset, innerLimit, sourceFactorCase, hardExcludeClause, visibilityClause, detailFilter, opts, dedup } = ctx;
-    const qRaw = query;
-    if (qRaw.length === 0) return [];
-    const qLike = escapeLikePattern(qRaw);
+    // Whitespace tokenization. The original matched the ENTIRE query as one
+    // indivisible ILIKE substring, so any multi-token query containing CJK
+    // returned empty unless the corpus held that exact string verbatim —
+    // spaces included. "记忆 系统" demanded a literal "记忆 系统"; Chinese
+    // prose writes "记忆系统". The failure was silent: an empty array reads
+    // as "nothing in the brain", not "the tokenizer can't express this".
+    //
+    // Mixed queries route here too (hasCJK is true for "cron 定时"), so the
+    // ASCII half was collateral damage on the same code path.
+    //
+    // Each token becomes its own ILIKE conjunct; scores sum across tokens.
+    // Single-token queries produce arithmetic identical to pre-fix.
+    const tokens = query.split(/\s+/).filter(t => t.length > 0);
+    if (tokens.length === 0) return [];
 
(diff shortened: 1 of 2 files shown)


time-attack:
Closing as superseded: #4299 (the same _searchKeywordCJK rewrite plus dedup, contiguous-run bonus, and the engine peel) has been landed on the v0.46.25.0 integration branch. The two hard-conflict on the same function, and #4299's version carries the extra hardening. Thank you for identifying and attacking the CJK substring bug — your diagnosis was correct and is credited in the fix lineage.

Issue #3665, opened 2026-07-30
Tests that pass on master: require a discrimination result before a fix is reviewable


time-attack:
A fix's test is only worth anything if it **fails without the fix**. Several PRs here have shipped or nearly shipped tests that pass identically on master — they add green checkmarks and CI time while proving nothing, and they make the next reviewer more confident, not less.

## Evidence

From a review pass over 32 open PRs in which 13 of 15 CI-green PRs failed adversarial review:

- **#3563** — **7 of its 8 new tests pass on master.** Verified by restoring the old tokenizer in both the source and the test fixture and re-running: the tests still passed. The PR's actual behavior change was almost entirely unpinned.
- **#3573** — 2 of 10 tests pass *vacuously* when the script under test is absent, because `bash` exits 127 and the assertion is only `≠ 0`. The other 8 do discriminate, so this one is partial.
- **#3548** — `const FLAGGED = ['warn','fail']` makes every status assertion date-tolerant, so nothing anywhere pins the post-sunset behavior. That is precisely why the PR's most serious blocker was invisible to its own suite.

Three more of the reviewed PRs were flagged for having no test at all on the specific path they changed (#3583 on three of four new call sites, #3561 on the only state in which its new code silently fails, #3574 on both new consented-apply surfaces).

This is not carelessness. Writing a test that passes is easy; confirming it *would have failed* requires a second, deliberate step that nothing currently asks for.

## Why this is worse than no test

A test that passes both ways is not neutral. It:

- makes the PR look better-covered than it is, so reviewers probe less;
- gets re-bin weighted into the LPT shards (`scripts/test-weights.json`) and costs CI time forever;
- will still pass after a future refactor silently breaks the behavior, so it actively delays discovery.

## What would fix it

**1. Make the discrimination result a required field in the PR template.** One line, filled by the author:

> Discrimination test: stubbed `<old behavior>` back into `<file:line>`, ran `<test file>` → `N pass / M fail`. Restored → all pass.

That is the whole ask. It is a two-minute step that authors can do, and it converts an invisible property into a reviewable claim. Every genuinely-good PR in this review round already did it naturally and said so.

**2. Consider a `git stash`-based helper** — `scripts/check-test-discriminates.sh <test-file> <source-file>` that reverts the source hunk, runs the test file, asserts non-zero exit, and restores. Not wired into CI (it cannot know which hunk is "the fix"), but a one-command local check makes the required field trivial to fill honestly.

[...]

PR #119, opened 2026-04-14
fix: preserve CJK paths in gbrain sync (core.quotepath=false)

vinsew:
## Summary

`gbrain sync` silently drops files with non-ASCII names from the sync manifest. Git's default `diff --name-status` output wraps CJK/unicode paths in double quotes with octal byte escapes:

```
A	\"inbox/2026-04-14 22_38 \350\256\260\345\275\225.md\"
```

`buildSyncManifest` treats that entire string as the literal path, downstream filesystem lookups fail, and the file is silently dropped. The user sees `added: 0, chunksCreated: 0` in the sync result even though git has the file committed, and `gbrain search` can't find the content. The cron log reports success because nothing technically errored.

This is the sync-layer counterpart to the same CJK root cause class fixed in #98 (query expansion), #114 (chunker), and #115 (slugify): ASCII-only assumptions baked into a fourth part of the codebase.

## Reproduction

```bash
mkdir /tmp/brain && cd /tmp/brain && git init
gbrain init
echo \"# test content\" > \"inbox/测试文件.md\"
git add . && git commit -m \"add test\"
gbrain sync --repo .
# → Output: { \"added\": 0, \"chunksCreated\": 0 }     ← bug
# → But `git log --stat` clearly shows the file was added.
# → `gbrain list` shows no new page; `gbrain search \"测试\"` returns nothing.
```

Real-world case: Apple Notes exports produce names like `2026-04-14 22_38 记录-个人智能体_原文.md` — the space + CJK combo guarantees git quoting, and those files never make it into the brain via the live-sync cron.

## Fix

One-line change to the `git()` helper in `src/commands/sync.ts`: prepend `-c core.quotepath=false` before the subcommand.

```ts
// before
execFileSync('git', ['-C', repoPath, ...args], ...)

// after
execFileSync('git', ['-c', 'core.quotepath=false', '-C', repoPath, ...args], ...)
```

`core.quotepath=false` tells git to emit paths as-is (UTF-8) in diff/log output. Setting it at the helper level means all current and future git invocations through this code path are covered — not just `diff`.

## Impact

- **2 files changed**, **+38 / -1 lines** (1-line code fix + explanatory comment + 3 tests)
- **Zero behavior change for ASCII-only paths** (same flag, same output format)
- **CJK filenames — with or without spaces — now sync correctly**
- Unblocks Chinese/Japanese/Korean users whose incremental sync has been silently partial
- Also fixes Apple Notes export (spaces + CJK) which is a common ingest pattern

[...]

PR #3899, opened 2026-08-08
fix(sync): unquote git C-style-quoted paths in buildSyncManifest (#3897)

### LINKED ISSUE #3897: sync: git C-style-quoted paths are dropped from the manifest silently (any path containing a `"`)
## Summary

`buildSyncManifest` (`src/core/sync.ts`) takes git's path column verbatim. git
C-style-quotes any path containing `"`, `\` or a control character, so such a
path arrives with surrounding quotes and backslash escapes, fails
`isMarkdownFilePath()`, and is discarded — with no error, no warning, and no
counter. The sync output is byte-identical to a clean run.

Affects add, modify, delete and rename alike.

Observed on `0.42.62.0` (installed from `github:garrytan/gbrain#f72de97`), and
still present on `master` @ `0b47afbf402a`: `buildSyncManifest` does
`trimmed.split('\t')` and pushes the column as-is, `isMarkdownFilePath` is
`path.endsWith('.md') || path.endsWith('.mdx')`, and there is no unquoting
anywhere in `src/`.

## Why `core.quotepath=false` does not prevent it

`buildGitInvocation` (`src/commands/sync.ts`) always passes
`core.quotepath=false`, and the comment there reads as though this normalises
all paths. It does not: `core.quotepath` governs **octal-escaping of non-ASCII
bytes only** (the case #119 fixed). Quoting of `"`, `\` and control characters
is unconditional in git — no config disables it.

This is why the bug looks rare and stays hidden: a vault full of accented and
CJK names syncs perfectly, and only a name with a literal quote is lost.

## Reproduction

```bash
git init repo && cd repo
mkdir people
printf 'x\n' > 'people/Marcus "Esports" Howard.md'
git add -A && git commit -m one
printf 'y\n' >> 'people/Marcus "Esports" Howard.md'
git add -A && git commit -m two
```

git emits (verified with `od -c`, using gbrain's own invocation):

```
M \t " p e o p l e / M a r c u s  \ " E s p o r t s \ "   H o w a r d . m d "
```

i.e. `M\t"people/Marcus \"Esports\" Howard.md"`.

`gbrain sync` then reports `up_to_date, modified: 0` and the page is never
updated.

## Root cause

1. `src/core/sync.ts` — `buildSyncManifest`: `trimmed.split('\t')`, path taken
   verbatim. No unquoting anywhere in `src/`.
2. `src/core/sync.ts` — `isMarkdownFilePath`: `path.endsWith('.md')`. The
   quoted path ends `.md"`, so it returns false.
3. The entry is dropped from the manifest with no diagnostic.

## Impact observed in production

A 14.7k-page vault, 2026-08-08:

- 41 tracked files carry such a name (19 under `people/`, 22 under
  `companies/` — mostly Russian company names like `ПАО "Ростелеком"` and
  nicknames like `Christian "Raz" Kippelt`). The quotes are meaningful; they
  cannot be renamed away.
- A vault pass deleted 45 pages. `sync_brain` reported `deleted: 44`. The
  45th, `people/Jason "Jay" Strand.md`, stayed live in the brain with
  `deleted_at: null` after its file was gone — a page answering searches with
  content that no longer exists anywhere.
- It was caught only because that pass happened to know the exact expected
  count. Nothing in the tool's output distinguishes it from success.

## Suggested fix

Unquote at the single point paths enter the manifest, so the extension filter
and every downstream consumer (slug resolution, file reads) see the real name.
Decode octal escapes as **bytes**, utf-8 decoding once at the end — one
codepoint can span several `\NNN` escapes, so per-escape decoding produces
mojibake.

A local patch (function `unquoteGitPath` plus four call sites in
`buildSyncManifest`) is running here and is unit-tested against the byte
sequences above. It was verified end-to-end against the live 14.7k-page vault
on 2026-08-08: an edit to a quoted-filename page that previously synced as
`up_to_date, modified: 0` now reports the edit and updates the page
(`content_hash aa492ab8… → d1663259…`), and the commit range that previously
lost one deletion — reporting 44 of 45 — now reports all 45. Entry counts
reconcile exactly: git reports 93 `.md` entries for the range, gbrain
processes 93.

Happy to send it as a PR against `master` if the approach looks right.

## A second, independent suggestion

Whatever the fix, a dropped manifest entry should not be silent. Emitting a
count of entries git reported vs entries processed — or simply warning on any
path that fails the extension filter after unquoting — would have surfaced
this on day one instead of after a page count happened to disagree.

SergeyShol:
Fixes #3897.

## The bug

`buildSyncManifest` takes git's path column verbatim. git C-style-quotes any
path containing `"`, `\` or a control character, so such a path arrives as
`"people/Jason \"Jay\" Strand.md"` — surrounding quotes included — ends `.md"`,
fails `isMarkdownFilePath()`, and is dropped from the manifest with no error, no
warning and no counter. The sync output is byte-identical to a clean run.

`core.quotepath=false` (#119) does not prevent this: it governs octal-escaping
of non-ASCII bytes only. Quoting of `"`, `\` and control characters is
unconditional in git — no config disables it. That is why the bug hides: a vault
full of accented and CJK names syncs perfectly, and only a name with a literal
quote is lost.

Observed in production on a 14.7k-page vault (2026-08-08): 41 tracked files
carry such a name; a pass that deleted 45 pages had `sync_brain` report
`deleted: 44`, and `people/Jason "Jay" Strand.md` stayed live in the brain with
`deleted_at: null` after its file was gone. It was caught only because that pass
happened to know the exact expected count.

## The fix

`unquoteGitPath()` — unquote at the single point every path enters the manifest,
so the extension filter and every downstream consumer (slug resolution, file
reads) see the real name. Four call sites in `buildSyncManifest`: add/modify,
delete, and both sides of a rename.

Octal escapes are decoded as **bytes** and utf-8 decoded once at the end — one
codepoint can span several `\NNN` escapes, so per-escape decoding produces
mojibake. An unquoted path is returned unchanged, so this is a no-op for the
common case and nothing about existing behaviour changes.

## Tests

9 new tests in `test/sync.test.ts`: `unquoteGitPath` unit coverage (passthrough,
quote/backslash/control escapes, multi-byte octal), manifest coverage for
add/modify/delete/rename, and one test that builds a real git repo with a quoted
filename, runs gbrain's own `buildGitInvocation`, asserts the exact quoted bytes
git emits, and pushes them through `buildSyncManifest`.

**Discrimination result** (per CONTRIBUTING / #3665) — with `unquoteGitPath`
stubbed to `return raw`, the 9 new tests fail and the 73 pre-existing ones in
that file pass:

```
(fail) unquoteGitPath > strips the wrapping quotes and unescapes embedded ones
(fail) unquoteGitPath > unescapes a literal backslash
(fail) unquoteGitPath > decodes single-character escapes
(fail) unquoteGitPath > decodes octal escapes as bytes, utf-8 decoding once at the end
(fail) unquoteGitPath > resulting path passes the extension filter
(fail) buildSyncManifest — C-style-quoted paths > unquotes add, modify and delete entries
(fail) buildSyncManifest — C-style-quoted paths > unquotes both sides of a rename
(fail) buildSyncManifest — C-style-quoted paths > quoted entries survive the syncable filter
(fail) buildSyncManifest — C-style-quoted paths > real git output for a quoted filename reaches the manifest
 73 pass
 9 fail
```

With the fix in place: `82 pass, 0 fail` in `test/sync.test.ts`, and
`bun run verify` is `pass=34 fail=0`.

The same change has been running against the live 14.7k-page vault since
2026-08-08: an edit to a quoted-filename page that previously synced as
`up_to_date, modified: 0` now reports the edit and updates the page, and the
commit range that previously lost one deletion now reports all 45. Entry counts
reconcile exactly — git reports 93 `.md` entries for the range, gbrain processes
93.

## Not included

The issue's second suggestion — warning when a manifest entry is dropped, rather
than dropping it silently — is deliberately left out of this PR. It changes
output on paths this one doesn't touch, and is worth its own change.

## Test-suite note

`bun run test` on this machine also reports 4 pre-existing failures in
`test/search/autocut-integration.serial.test.ts` and
`test/worker-registry.serial.test.ts`. They reproduce identically on unmodified
`master` @ `0b47afbf402a`, so they are not from this change.

### DIFF
--- a/src/core/sync.ts
+++ b/src/core/sync.ts
@@ -102,6 +102,54 @@ const CODE_EXTENSIONS = new Set<string>([
  *   D       path/to/deleted-file.md
  *   R100    old/path.md     new/path.md
  */
+/**
+ * Undo git's C-style path quoting.
+ *
+ * git quotes any path containing `"`, `\` or a control character, and it does
+ * so unconditionally: `core.quotepath=false` (buildGitInvocation in
+ * commands/sync.ts, #119) only suppresses octal-escaping of NON-ASCII bytes.
+ * A path like `people/Jason "Jay" Strand.md` therefore reaches
+ * buildSyncManifest as `"people/Jason \"Jay\" Strand.md"` — surrounding quotes
+ * included — so it ends `.md"`, fails isMarkdownFilePath(), and is dropped from
+ * the manifest with no error and no counter.
+ *
+ * Octal escapes are decoded as BYTES and utf-8 decoded once at the end: a
+ * single codepoint can span several \NNN escapes, so per-escape decoding
+ * produces mojibake.
+ *
+ * An unquoted path is returned unchanged, so this is a no-op for the common
+ * case.
+ */
+export function unquoteGitPath(raw: string): string {
+  if (raw.length < 2 || !raw.startsWith('"') || !raw.endsWith('"')) return raw;
+  const body = raw.slice(1, -1);
+  const simple: Record<string, number> = {
+    n: 10, t: 9, r: 13, '"': 34, '\\': 92, a: 7, b: 8, f: 12, v: 11,
+  };
+  const bytes: number[] = [];
+  let i = 0;
+  while (i < body.length) {
+    const c = body[i];
+    if (c === '\\' && i + 1 < body.length) {
+      const nxt = body[i + 1];
+      if (nxt in simple) {
+        bytes.push(simple[nxt]);
+        i += 2;
+        continue;
+      }
+      const trio = body.slice(i + 1, i + 4);
+      if (trio.length === 3 && /^[0-7]{3}$/.test(trio)) {
+        bytes.push(parseInt(trio, 8));
+        i += 4;
+        continue;
+      }
+    }
+    for (const b of new TextEncoder().encode(c)) bytes.push(b);
+    i += 1;
+  }
+  return new TextDecoder('utf-8').decode(new Uint8Array(bytes));
+}
+
 export function buildSyncManifest(gitDiffOutput: string): SyncManifest {
   const manifest: SyncManifest = {
     added: [],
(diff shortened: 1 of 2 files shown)


time-attack:
Merged in the v0.45.8.0 community fix-wave (#4050). Your fix landed on master as ce156eb8ed, credited to you via Co-Authored-By, and is called out in the v0.45.8.0 CHANGELOG entry. Thank you for the contribution!

PR #3900, opened 2026-08-08
fix(features): treat heartbeat-backed integrations as configured

mitche10:
## Problem
The \`gbrain features\` command's 'Set Up Integrations' recommendation decided whether an integration was configured by checking ONLY if its secret env vars were exported in the current shell:

\`\`\`ts
const unconfigured = RECIPE_META.filter(r => !r.secrets.every(s => process.env[s]));
\`\`\`

Integrations installed through Hermes (and other hosts) persist a \`setup_complete\` event to \`~/.gbrain/integrations/<id>/heartbeat.jsonl\` rather than exporting secrets into the shell. Result: a fully-installed integration with a real heartbeat was falsely reported as 'not configured'. On one real deployment this reported 7/7 recipes as unconfigured when 4 were actually live.

## Fix
A recipe is now considered configured if EITHER its secrets are present in the environment OR a real \`setup_complete\` heartbeat exists on disk. Added \`integrationHeartbeatExists(id)\` which reads the heartbeat jsonl and confirms a successful setup event.

## Verification
- Typecheck passes (\`bun build --no-bundle\`).
- Ran the actual command against a live brain: previously reported 7 unconfigured, now correctly reports only the 3 recipes with no heartbeat file (Voice to Brain, Meeting Sync, Ngrok Tunnel).

This is a non-destructive, strictly-more-accurate check — it can only reduce false 'not configured' reports.

### DIFF
--- a/src/commands/features.ts
+++ b/src/commands/features.ts
@@ -72,6 +72,32 @@ function saveOffers(offers: FeatureOffersFile) {
   } catch { /* best-effort */ }
 }
 
+// Real-world check: did this integration actually get set up?
+// Heartbeat files are written by Hermes during install (and by other hosts),
+// so they are the source of truth for "is this wired up" — not just whether
+// a secret env var happens to be exported in this shell.
+function integrationHeartbeatExists(recipeId: string): boolean {
+  try {
+    const dir = join(process.env.HOME || '', '.gbrain', 'integrations', recipeId);
+    const heartbeat = join(dir, 'heartbeat.jsonl');
+    if (!existsSync(heartbeat)) return false;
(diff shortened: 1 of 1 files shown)


time-attack:
Re-trialed against the current integration branch (post v0.46.23-25 waves): conflict: src/commands/features.ts. Wave 190cef5f rewrote same hunk (any-of secrets); PR's heartbeat=>configured check absent on wave — overlapping, not superseded. Rework must keep any-of; scrub 'Hermes'. No red flags. — a rebase along those lines puts this back in the active merge queue; re-trials run nightly, so a fresh push gets picked up automatically. Thank you!

garrytan:

PR #3909, opened 2026-08-09
fix(ai): match hyphenated Qwen3-Embedding ids for dimensions passthrough

mikez93:
## Problem

The Qwen3-Embedding matryoshka `dimensions` passthrough in `dimsProviderOptions` only matches Ollama's colon-tag naming (`qwen3-embedding:8b`). OpenRouter and other HF-style routers serve the same family as `qwen/qwen3-embedding-8b` — org prefix plus hyphenated size.

Result today: with `embedding_model: openrouter:qwen/qwen3-embedding-8b` and a narrower configured width (e.g. 1536), the `dimensions` parameter is silently never sent, the provider returns the model's native width (4096 for 8B), and the brain hard-fails with a dim mismatch at first embed — even though both the model and the engine fully support the truncation.

## Fix

Match the bare hyphenated form alongside the colon-tag form, reusing the `bareModelId` org-prefix strip already used by the `text-embedding-3` branch a few lines up. Native-width suppression (the vLLM HTTP-400 guard) extended to both spellings.

Verified live against OpenRouter: `qwen/qwen3-embedding-8b` honors `dimensions` at 1536/2000/1024 exactly, and a full 8.7k-chunk brain migration via `gbrain migrate embeddings` completes with the patch applied.

## Tests

`test/ai/dims-qwen3-native.test.ts` extended: hub-form truncation, hub-form native-width suppression, bare hyphenated form, unknown hyphenated variant fall-through. Full `test/ai/` suite green on this branch (492 pass).

Related: #3584 (query-side Instruct template) composes with this — that PR covers the asymmetric-query gap, this one covers reaching the family through hub-form routers at all.

### DIFF
--- a/src/core/ai/dims.ts
+++ b/src/core/ai/dims.ts
@@ -309,9 +309,12 @@ export function dimsProviderOptions(
       // provider serving it) supports Matryoshka truncation via `dimensions`.
       // Native sizes: 0.6B=1024, 4B=2560, 8B=4096. Without `dimensions`,
       // Ollama returns the native size and brains configured for narrower
-      // widths hard-fail with a dim-mismatch error. Pattern match the bare
-      // model name + any `:tag` (e.g. `qwen3-embedding:4b`, `qwen3-embedding:0.6b`).
-      if (modelId === 'qwen3-embedding' || modelId.startsWith('qwen3-embedding:')) {
+      // widths hard-fail with a dim-mismatch error. Two naming schemes reach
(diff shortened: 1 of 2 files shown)


time-attack:
Merged in the v0.45.8.0 community fix-wave (#4050). Your fix landed on master as 8ecd52022e, credited to you via Co-Authored-By, and is called out in the v0.45.8.0 CHANGELOG entry. Thank you for the contribution!

PR #3915, opened 2026-08-09
fix(cycle): price synthesize_concepts estimates from the canonical table

Masashi-Ono0611:

> the price table seems fixed for Sonnet even thou we can run with a cheaper model such as gpt-5.2. therefore, i've modified the price mapping so that the price estimation would work correctly.

## What

`synthesize_concepts`' T1/T2 cost estimates hardcode Sonnet rates (`* 3.0` / `* 15.0` with a "Sonnet at ~$3/M input + $15/M output" comment — `src/core/cycle/synthesize-concepts.ts:207-209`), which violates the repo invariant that all paid-cloud chat prices live once in `CANONICAL_PRICING` and everything else derives from it.

On a non-Sonnet model the estimate diverges from reality: with `openai:gpt-5.2` ($1.25/$10 vs Sonnet's $3/$15) it runs 1.5×–2.4× high depending on the input/output token mix, so the `$1.50` budget gate can fire early and drop concept generation to the deterministic template, and the inflated figure is then persisted into receipts (`:268`) and extract rollups (`:282`) as if it were real spend.

## Fix

- Price from `canonicalLookup(result.model)` — the model that actually answered (`ChatResult.model` carries `provider:modelId`).
- Canonical miss falls back to Sonnet-tier rates **derived from** `canonicalLookup('anthropic:claude-sonnet-4-6')` (non-throwing), following the `skillopt/preflight.ts` `lookupPrice` precedent. This preserves the pre-fix behavior exactly for unknown models and keeps the budget gate armed. The BudgetMeter warn-and-bypass shape was considered and not used here: it would newly allow unmetered runs and $0 receipts for unknown models.
- The `$1.50` gate threshold, concept generation, and model resolution are untouched.

## Relationship to #3718

Complementary, not overlapping. #3718 fixes *which model is called* (honour `models.dream.synthesize`); this PR fixes *what the call is priced at*. The hunks don't intersect — both patches were verified to apply cleanly together. Once #3718 lands, non-Sonnet models actually flow through this path, which makes this fix more load-bearing.

## Tests

- Discriminating test — "non-Sonnet canonical model is priced at its own rates (openai:gpt-5.2)" — fails on pre-fix code, passes post-fix (revert-check executed, 29 pass / 1 fail pre-fix).
- Miss-policy test + Sonnet control test. Expectations are derived from the canonical table so future price updates don't break them, with a guard asserting that gpt-5.2 and Sonnet rates actually differ (the discriminating test cannot silently degenerate).
- `bun test test/cycle/extract-atoms-synthesize-concepts.test.ts` 30 pass / 0 fail; adjacent suites (synthesize-concepts-progress, model-pricing) 22 pass; `bun run typecheck` clean.

### DIFF
--- a/src/core/cycle/synthesize-concepts.ts
+++ b/src/core/cycle/synthesize-concepts.ts
@@ -30,8 +30,20 @@ import { chat as gatewayChat, isAvailable } from '../ai/gateway.ts';
 // source-boost's 1.3× 'concepts/' weighting can actually reach them.
 import { importFromContent } from '../import-file.ts';
 import { serializeMarkdown } from '../markdown.ts';
+import { canonicalLookup, type ModelPricing } from '../model-pricing.ts';
 
 const DEFAULT_BUDGET_USD = 1.5;
+// Canonical-miss policy — mirrors skillopt/preflight.ts's lookupPrice:
+// assume Sonnet-tier pricing for models absent from CANONICAL_PRICING.
+// Conservative and non-throwing; keeps the budget gate effective (and
+// matches this file's pre-canonical behavior) instead of letting an
+// unpriced model run unmetered. The rates are DERIVED from the canonical
+// table (never hand-copied — CLAUDE.md invariant); the literal pair only
+// fires if the Sonnet key itself ever leaves the table.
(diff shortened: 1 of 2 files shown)


Masashi-Ono0611:
### Screenshot
Screenshot 2026-08-09 at 3 39 06 PM

PR #3916, opened 2026-08-09
fix(think): extend the 16K output headroom to OpenAI reasoning models

Masashi-Ono0611:

> when i was running "gbrain think" with a reasoning model by OpenAI such as gpt5.6, i figured out that the token window was wider only with Anthropic Claude-5 family. And i believe that it should be available with other new models. Therefore, i've created this PR.

## What

`gbrain think` clamps output at 4,000 tokens unless the model matches the Anthropic Claude-5 thinking regex, which gets a 16,000 headroom (`src/core/think/index.ts:165-178`). That headroom exists because thinking-by-default models spend part of the output budget on reasoning before the visible answer.

OpenAI reasoning models (the gpt-5 family and the numbered o-series) are in exactly the same situation — reasoning tokens draw from the same `max_tokens` budget — but they fall through to the 4,000 default, so a think run routed at them starves its visible synthesis.

## Fix

A second, deliberately narrow regex: `/^openai[:/](?:gpt-5|o[0-9]+)(?:[.-]|$)/i`, with `-chat` snapshots (non-reasoning) excluded. The 4,000/16,000 defaults are unchanged and no config knob is added.

A recipe-level "reasoning" capability declaration doesn't exist today (grepped `recipes/openai.ts` and `types.ts` — the only hit is a DeepSeek `reasoning_content` comment), so this follows the existing model-regex precedent in the same function rather than inventing new schema surface.

## Scope

- `gpt-5-chat-latest` and other `-chat` snapshots intentionally stay at 4K (pinned by test).
- `codex-mini-latest` and other non gpt-5/o-series ids intentionally stay at 4K (pinned by test).
- openrouter-nested ids don't match (boundary-tested), and boundary look-alikes (`gpt-50`, `o3foo`, bare ids) are covered.

## Tests

- Discriminating test — OpenAI reasoning ids get 16,000 — fails on pre-fix code (revert-check executed: 2 pass / 1 fail pre-fix, 3 pass post-fix).
- Anthropic Claude-5 control (16,000, unchanged) and non-reasoning control (4,000, unchanged).
- 3 tests / 29 asserts in the file; `bun run typecheck` clean.

### DIFF
--- a/src/core/think/index.ts
+++ b/src/core/think/index.ts
@@ -171,8 +171,19 @@ const DEFAULT_MAX_OUTPUT_TOKENS = 4000;
 // keeps 4000.
 const THINKING_DEFAULT_MAX_OUTPUT_TOKENS = 16000;
 const THINKING_BY_DEFAULT_MODEL_RE = /^anthropic[:/]claude-[a-z0-9]+-5(?:[.-]|$)/i;
+// OpenAI reasoning models spend output budget on internal reasoning tokens
+// the same way — reasoning tokens are billed as output and count against
+// `max_tokens` — so they get the same headroom. Deliberately scoped to the
+// gpt-5 family and the numbered o-series only; anything else (gpt-4o, the
+// non-reasoning `*-chat` snapshots like gpt-5-chat-latest, other providers'
+// reasoning models routed through their own recipes) keeps the conservative
+// 4000 default.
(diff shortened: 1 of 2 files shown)


Masashi-Ono0611:
### Screenshot
Screenshot 2026-08-09 at 3 30 31 PM

Issue #2778, opened 2026-07-13
bug(subagent): fenced allowlist omits add_timeline_entry (the canonical timeline op); 4096-token turn cap + unrecovered tool_use truncation make pages >~12KB unwritable


alexputici:
**What happened?**

> Deployment context (applies to a batch of reports filed today, all from the same production operation): ~1 week of intensive operation (2026-07-06 → 2026-07-12) of a ~7,000-page brain (6,907 indexed pages / 8,945 chunks / 100% embedded, single source) on a Mac mini against Supabase Postgres (PgBouncer transaction mode). Full native surface plus a thin custom layer: `gbrain autopilot` (launchd, KeepAlive), Minions with five sense collectors dispatching idempotency-keyed jobs, fenced triage/apply subagents from plugin defs, remote MCP over a cloudflared tunnel, hourly git backup lane. Every report was hit in real operation, has receipts (job numbers, log lines, source line numbers), and most have a local workaround deployed whose retirement is the point of filing. All source references verified against v0.42.58.0 (a25209bb).

Three interacting defects in the fenced subagent write path make it impossible for a fenced subagent to update any page beyond ~12KB:

1. **The allowlist omits gbrain's own canonical timeline op.** `BRAIN_TOOL_ALLOWLIST` (`src/core/minions/tools/brain-allowlist.ts:48-71`) includes `put_page` but not `add_timeline_entry` — with no justifying comment (the only write-exclusion comment in the file, lines 57-59, argues for excluding LINK writes specifically). Yet the compiled-truth guide (`docs/guides/compiled-truth.md:71-88`) prescribes `add_timeline_entry` for ALL timeline appends ("timeline is NOT passed" to put — put is for compiled-truth rewrites only), and `engine.addTimelineEntry` is an O(entry) idempotent insert — `ON CONFLICT (page_id, date, summary, source) DO NOTHING` (`postgres-engine.ts:3387-3392`) — that works at any page size. The allowlist blocks the canonical surface and leaves full-page-echo `put_page` as the only write path.

2. **The default subagent loop caps every turn at `max_tokens: 4096`** (`src/core/minions/handlers/subagent.ts:483`, the default/legacy path — the gateway loop is opt-in per lines 186-187).

[...]

PR #3917, opened 2026-08-09
fix(gateway): propagate finish_reason=length through toolLoop as a non-clean stop

Masashi-Ono0611:

> when i was running toolLoop by OpenAI models on GBrain subagents, GBrain always recognizes normal result even thou the actual result was "finish_reason=length". Therefore, i would like to fix this issue to let it show the actual ones.

## What

When a gateway-routed subagent turn ends with the provider's `finish_reason=length` (output truncated), `toolLoop` reports it as a clean end: the zero-tool-calls exit path sets `stopReason = 'end'` unconditionally and never consults `chatResult.stopReason` — and the `ToolLoopStopReason` union has no `'length'` member, so the truncation is structurally impossible to propagate. Downstream, the subagent handler maps `'end'` to `stop_reason='end_turn'`, so a truncated run is recorded as a clean success.

`mapStopReason` itself already maps the provider's length signal correctly; it is only the loop's exit classification that drops it.

## Fix

- Add `'length'` to `ToolLoopStopReason`; the zero-tool-calls exit returns it when `chatResult.stopReason === 'length'`. All other exit paths unchanged.
- The subagent handler maps `'length'` to the **existing** `'max_tokens'` stop reason (the #2778 "result text is TRUNCATED" marker) — record-only, the job is not failed. This is deliberately the same policy the legacy Anthropic-direct path already applies to a zero-tool-use `max_tokens` stop, so the two paths now agree instead of one of them silently reporting success.
- The handler's stop-reason mapping is now an exhaustive `Record<ToolLoopStopReason, SubagentStopReason>` (a future stop-reason addition fails typecheck instead of silently falling through to a success value).
- `skillopt`'s `Trajectory.stop_reason` was a hand-written copy of the gateway union; it now imports the type from the source (no consumer branches on it).

## Relationship to the synthesize completion stamp

This PR makes truncation *visible* to every consumer of `stop_reason`. Whether synthesize should still stamp completion over non-clean children is a separate question tracked elsewhere; nothing here changes completion semantics.

## Tests

- Red-first: "toolLoop propagates length stop reason" and "handler maps length → max_tokens" both fail on pre-fix code (revert-check executed: 17 pass / 2 fail pre-fix, 19 pass / 0 fail post-fix).
- Controls pin that a length stop with tool calls still continues the loop, and that clean ends stay `end_turn`.
- Adjacent suites (skillopt, subagent-handler, crash-replay, resume-reconciliation, gateway-chat): 290 pass. `bun run typecheck` clean.

Related to #2778 (the `max_tokens` truncation marker this reuses).

### DIFF
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -3645,7 +3645,7 @@ export interface ToolLoopOpts {
   onHeartbeat?: (event: string, data: Record<string, unknown>) => void;
 }
 
-export type ToolLoopStopReason = 'end' | 'max_turns' | 'refusal' | 'content_filter' | 'aborted' | 'unrecoverable';
+export type ToolLoopStopReason = 'end' | 'length' | 'max_turns' | 'refusal' | 'content_filter' | 'aborted' | 'unrecoverable';
 
 export interface ToolLoopResult {
   finalText: string;
@@ -3750,7 +3750,11 @@ export async function toolLoop(opts: ToolLoopOpts): Promise<ToolLoopResult> {
     );
 
     if (toolCalls.length === 0) {
-      stopReason = 'end';
+      // An output-cap stop (finish_reason=length / max-tokens) is NOT a clean
+      // end — the text is truncated. Propagate it so the caller can record a
+      // non-clean stop_reason instead of a silent success (the gateway-loop
+      // analog of the legacy path's #2778 handling).
+      stopReason = chatResult.stopReason === 'length' ? 'length' : 'end';
       finalText = chatResult.text;
       break;
     }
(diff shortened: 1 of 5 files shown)


Masashi-Ono0611:
### Screenshot
Screenshot 2026-08-09 at 3 00 55 PM

Masashi-Ono0611:
Closing — this is now fully superseded by upstream's own fix. v0.46.19.0 (#4216 wave, PR #4253 / "5 subagent honesty fixes") shipped an equivalent \`stopReason === 'length' ? 'max_tokens' : ...\` mapping natively in \`subagent.ts\` and \`skillopt/types.ts\` already imports \`ToolLoopStopReason\` from gateway.ts — functionally identical outcome to this PR's change. Verified line-for-line while rebasing a local patch-stack through v0.46.19.0 today. No action needed on this PR; the behavior it proposed is already live on master.

Issue #3910, opened 2026-08-09
grade_takes: a judge output parse failure is minted as a durable 'unresolvable' verdict — cached forever, indistinguishable from a real verdict, never surfaced


Masashi-Ono0611:
**Type:** bug
**Component:** `dream` cycle / `grade_takes`
**Version:** master `f15480b9` (v0.42.75.0)
**Engine:** any

## Summary

#3763 reports the propose_takes side of a failure-handling gap: an unparseable extractor response is *not* memoized, so the negative-result cache never converges and the phase re-bills forever. `grade_takes` has the **opposite policy for the same failure class**, with the opposite failure mode:

`defaultJudge` catches an unparseable judge response and returns a synthetic verdict — `{verdict: 'unresolvable', confidence: 0, reasoning: 'judge_output_parse_failed'}` (`src/core/cycle/grade-takes.ts:330-339`). The caller cannot tell this apart from a genuine evidence-based verdict: it lands in `take_grade_cache` (`:559-566` — the row persists verdict/confidence/applied, **not** reasoning), and every later cycle hits the idempotency lookup and skips the take (`:457-467`, keyed on `(take_id, prompt_version, judge_model_id, evidence_signature)`).

Net effect: **one transient formatting miss permanently freezes grading for that take** under the current judge model + prompt version, and the frozen row is indistinguishable from a real "insufficient evidence" verdict.

Three aggravators:

1. **The code's own justification doesn't hold.** The comment above the return (`:331-333`) says the row lands in the cache "so the operator sees the LLM's parse failure surfaced via warnings" — but no warning is pushed on this path. `judge_output_parse_failed` appears nowhere else in the file; the only warning in the grade loop is for a *thrown* judge error (`:487-490`). A parse failure is fully silent: the phase reports ok, `cache_hits` climbs, nothing names the take.
2. **"Re-grade when evidence changes" never fires with the shipped retriever.** `evidence_signature` is derived from the evidence text, and the default retriever is the static placeholder (#2811) whose output is deterministic per take. Under a stable judge model + prompt version, the suppression is effectively permanent, not "until evidence improves."
3. **It is inconsistent with the thrown-error path in the same loop.** A judge call that *throws* → warning + continue, no cache row, retried next cycle. A judge call that *returns garbage* → silent, durable tombstone. Two failure modes of the same class, opposite policies, one function.

[...]

Issue #3763, opened 2026-08-02
propose_takes: the #2106/#2136 negative-result cache is not written for unparseable extractor output, so the re-billing loop returns at 100% of pages


MarcvsTvllivs:
**Type:** bug
**Component:** `dream` cycle / `propose_takes`
**Version:** gbrain 0.42.72.1
**Engine:** PostgreSQL

## Summary

#2106 and #2136 fixed the case where a zero-take page never got an idempotency row and was re-billed every cycle. The fix memoizes a **cleanly parsed empty array** via a tombstone. It does not memoize an **unparseable** response.

Any model whose output fails to parse therefore reproduces the exact loop those two issues closed, except at 100% of pages instead of the zero-take subset. On my brain that ran for 60 hours: 92 cycles, 8,425 extractor submits, roughly 14,400 provider calls, **0 rows written**, and the phase reported `status: ok` on every one of those cycles.

This is not a request for the config gate from #3033. I understand that posture and am not re-litigating it. This is the negative-result cache having one uncovered branch.

## Mechanism

`defaultExtractor` throws when the output does not parse (`core/cycle/propose-takes.ts:313`), the caller catches it, pushes a warning, and continues to the next page (`:552`). No tombstone is written.

The comment at `:305-313` explains why, and the reasoning is sound in isolation: a transient parse failure must not permanently suppress a page that genuinely has claims, so only a cleanly parsed `[]` is memoized. The gap is that **nothing distinguishes a transient failure from a deterministic one**. A model that can never satisfy the output contract fails identically on every page, on every cycle, and each failure is treated as a fresh blip worth retrying in full.

With no tombstone there is no row, so the next cycle's lookup on `(source_id, page_slug, content_hash, prompt_version)` misses again and the same pages are re-sent. The idempotency cache is structurally unreachable in this state.

## What I measured

92 cycles, 2026-07-31 02:42 to 2026-08-02 17:14, from `minion_jobs.result` and `.gbrain/audit/dream-budget-*.jsonl`. Chat model was an OpenAI-compatible mandatory-reasoning model (`xiaomi-mimo-v2-5`, Venice-hosted, via `OPENAI_BASE_URL`).

| metric | value |
|---|---|
| cycles | 92 |
| extractor submits | 8,425 |
| provider calls including retries | ~14,400 |
| wall clock in `propose_takes` | 30.9 h |
| share of all cycle AI time | 99.97% (every other phase combined: 0.01 h) |
[...]

PR #3918, opened 2026-08-09
fix(dream): do not permanently cache truncated or degenerate significance verdicts

Masashi-Ono0611:

> When i was using a reasoning such as gpt-5.6 sol on GBrain dream, the verdict result ignored all transcripts as this 200 token windows got filled by only the reasoning. As my current workaround, i keep using sonnet5 but i would like to switch it to other model. Therefore, i am trying to fix this issue.

## What

`judgeSignificance` — the dream cycle's transcript-significance verdict — has four defects that compose into one severe failure on reasoning models (`src/core/cycle/synthesize.ts`, line refs at master `f15480b9`):

1. `max_tokens: 200` hardcoded (`:1093`) — a reasoning model spends that budget on reasoning before emitting the verdict JSON;
2. the adapter pins `stop_reason: 'end_turn'` (`:1029`) — truncation is invisible upstream;
3. an unparseable verdict silently becomes `worth_processing: false` (`:1114`) — no warning, no distinction from a genuine "not worth processing";
4. that false is permanently cached in `dream_verdicts` (`:494`) — the transcript is never re-judged.

Net effect: point a reasoning model at the verdict touchpoint and a transcript whose verdict response gets truncated or fails to parse is **silently rejected, forever**, while the phase reports ok. (Hit as a near-miss on my brain while routing dream to OpenAI models; avoided by pinning the verdict back to a non-reasoning model.)

## Fix

- The adapter propagates the real `ChatResult.stopReason` (`length` → `max_tokens`; `refusal`/`content_filter` → `refusal`; `other` stays `end_turn` — a commented, deliberate scoping decision, see below).
- `judgeSignificance` raises the cap to 1,024 (repo precedent: grade-takes judge 600, eval contradictions judge 1,024) and returns `unreliable?: 'truncated' | 'refusal' | 'unparseable'`. A truncated response is unreliable even when it happens to parse; a non-boolean `worth_processing` counts as unparseable.
- The verdict loop **skips the cache write** for unreliable results and emits a stderr warning naming the transcript, the reason, and that the next cycle will re-judge. Clean verdicts cache exactly as before.

This matches the failure-handling shape the codebase already uses for judge errors in grade-takes (warn + continue, no cache row), and addresses the synthesize member of the same degenerate-result-caching class as #3763 (propose_takes) and #3910 (grade_takes).

## Deliberately out of scope

`mapStopReason` (gateway.ts) folds the AI SDK's `error`/`unknown` finish reasons into `other`; gateway.ts is untouched here because a broad open PR is reworking that file. The residual risk is narrow — an error-terminated response would also need a complete boolean-bearing JSON body to be cached (unparseable is already not cached) — and labeling `error`/`unknown` distinctly in `mapStopReason` is the natural follow-up.

## Tests

- 12 new tests (judge unit ×8, adapter propagation ×1, e2e cache behavior ×3). 10 fail on pre-fix code (revert-check executed: 7 unit + 1 adapter + 2 e2e red pre-fix); the remaining 2 controls (a clean parseable verdict stays cacheable, pinned at both the unit and e2e level) pass on both sides.
- Full affected suites: 81 pass / 0 fail; `bun run typecheck` clean.

### DIFF
--- a/src/core/cycle/synthesize.ts
+++ b/src/core/cycle/synthesize.ts
@@ -491,7 +491,22 @@ export async function runPhaseSynthesize(
       }
       try {
         const verdict = await judgeSignificance(judge, t, config.verdictModel);
-        await engine.putDreamVerdict(t.filePath, t.contentHash, verdict);
+        if (verdict.unreliable) {
+          // Degenerate judgement (truncated, refused/content-filtered, or
+          // unparseable LLM output). Do
+          // NOT write it to dream_verdicts: a cached `worth_processing:
+          // false` is permanent for this content hash, and a brain whose
+          // verdict model reliably truncates (e.g. a reasoning model under
+          // a tight budget) would silently reject every transcript forever.
+          // Log + skip so the next cycle re-judges.
+          process.stderr.write(
+            `[dream] verdict for ${t.basename} was ${verdict.unreliable} ` +
+            `(${verdict.reasons.join('; ')}); not caching in dream_verdicts — ` +
+            `next cycle will re-judge ${t.filePath}\n`,
+          );
+        } else {
+          await engine.putDreamVerdict(t.filePath, t.contentHash, verdict);
+        }
         verdicts.push({ filePath: t.filePath, worth: verdict.worth_processing, reasons: verdict.reasons, cached: false });
         if (verdict.worth_processing) worthProcessing.push(t);
       } catch (e) {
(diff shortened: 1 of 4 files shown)


Masashi-Ono0611:
### Screenshot
Screenshot 2026-08-09 at 2 50 54 PM

ghost:
.

time-attack:
Triage bot: floor + e2e green (targeted, sibling, verify, dream e2e all pass) and one adversarial pass found no doubt. The second pass raised one real scope question that makes this a maintainer call rather than a bot merge: the fix is prospective-only. Brains ALREADY poisoned by cached degenerate verdicts keep their permanent `worth_processing: false` rows (fingerprintable as `reasons = ['judge response unparseable']` in dream_verdicts), and the untouched cache-read path trusts them forever — no invalidation, backfill, or read-side bypass ships here, so existing installs stay stuck until content changes rehash. Options: ship as-is + follow-up sweep, or add a one-time invalidation of rows whose reasons match the degenerate fingerprints. Related: #3910 reports the same failure class in grade_takes' cache. Queued for your decision; otherwise this looked solid.

Masashi-Ono0611:
Thanks — agreed the fix is prospective-only, and I'd rather not bolt the invalidation onto this PR. Some field data from a single-user install that has been running dream nightly since May, in case it helps the call:

| | rows |
|---|---:|
| `dream_verdicts` total | 1,405 |
| `reasons = ['judge response unparseable']` | 28 (all `worth_processing = false`) |
| `worth_processing = false` **and** `reasons = []` | 12 |

The 28 span `2026-05-24 .. 2026-08-09` — i.e. it accrues continuously rather than being one bad night, and the newest landed the day before this PR.

**The proposed fingerprint misses a second degenerate shape.** Pre-fix, the parse does `const worth = parsed.worth_processing === true`, so *any* JSON that parses but lacks a boolean `true` — `{}`, `{"worth_processing": "true"}`, a truncation that happens to close — returns `{worth_processing: false, reasons: []}` and gets cached as a real verdict. That's the 12 rows above. This PR already covers that case prospectively (the "a JSON object without a boolean `worth_processing` is NOT a verdict" branch), so its residue is just as permanent as the unparseable one.

Caveat on using `reasons = []` as an invalidation key: it is **not** exact — a model can legitimately answer `{"worth_processing": false, "reasons": []}`. But the blast radius of over-invalidating is one re-judge per row, not data loss, so an over-inclusive key looks like the right trade here.

So: ~2.8% of verdict rows on this install are permanently stuck, across two shapes rather than one. My suggestion is to ship this as-is and do the invalidation as its own change that covers both shapes — and, if #3910 is the same class in `grade_takes`, one sweep covering both caches rather than two partial ones. Happy to send that as a follow-up PR if you want it.

time-attack:
Merged in the v0.45.8.0 community fix-wave (#4050). Your fix landed on master as 3fa0a5acb5, credited to you via Co-Authored-By, and is called out in the v0.45.8.0 CHANGELOG entry. Thank you for the contribution!

Issue #1151, opened 2026-05-18
Subagent worker dead-letters successfully-completed jobs when final message is text-only assistant summary


zscgeek:
## Symptom

Dream subagent finishes its synthesis task — writes pages via `brain_put_page` tool_use, then emits a final text-only assistant message summarizing what was written. Worker tries to "resume" from that terminal state, calls `messages.create` with the conversation ending on assistant role.

Anthropic API returns:

```
400 invalid_request_error
"This model does not support assistant message prefill.
 The conversation must end with a user message."
```

Job marked `dead` after 3 retry attempts despite **all pages already successfully written** in earlier turns.

## Reproduction

Run `gbrain dream` on an Otter transcript corpus where the subagent does its work and naturally concludes. In our test run on 27 transcripts (Sonnet 4.6, Anthropic Tier 2), ~7% of subagent jobs hit this. Concrete evidence from one job:

- **Job 190** (status: dead, error: `400 ... This model does not support assistant message prefill`)
- Inspected `subagent_messages` for `job_id = 190`:
  - 15 messages persisted
  - Message 15 (last) = role=`assistant`, content = final text summary listing the slugs that were written
  - Slugs verified to exist in `pages` table:
    - `wiki/originals/ideas/2026-05-17-ai-smart-intern-failure-modes-fd3ece`
    - `wiki/originals/ideas/2026-05-17-claude-escalation-list-leakage-fd3ece`
    - `wiki/personal/reflections/2026-05-17-hubspot-cleanup-avoidance-fd3ece`

The job's `error_text` field shows the 400 prefill error. The work was complete; the worker just didn't recognize it.

## Root cause (suspected location)

`src/core/minions/handlers/subagent.ts:240-260` — replay reconciliation logic.

The current code checks if the last persisted assistant message has pending tool_uses. If yes, it dispatches the tools. If no, it falls through to calling `messages.create` again — but with the conversation ending on an assistant role, which Sonnet 4.6 rejects (older Sonnet allowed assistant-prefill; 4.6 doesn't).

The bug: the worker doesn't recognize "final assistant message has `stop_reason: end_turn` and no pending tool_uses" as terminal. It should mark the job complete instead of attempting another model call.

## Proposed fix

Check the last assistant message's `stop_reason`:

- `stop_reason === 'end_turn'` and no pending tool_uses → conversation is complete, mark job `completed`, do not call `messages.create`
[...]

PR #3919, opened 2026-08-09
fix(minions): verify the finally-resolved model at the subagent gate — the models.subagent config path bypassed capability checks

Masashi-Ono0611:

> i found that "gbrain config set models.subagent" didn't actually check the availability of the model for tool calling.
> Therefore, i've solved the issue by this PR.

## What

The subagent handler's capability gate only fires for jobs that carry an explicit `data.model`. When `data.model` is omitted, the handler resolves `models.subagent` from config — and that resolution path (`resolveModel`'s explicit-key branch) never passes through `enforceSubagentCapable`'s silent-fallback protection, which only guards the inherited branches (`models.default` / tier / env). Net effect: **a model selected via `models.subagent` config reaches `gateway.toolLoop()` without enforcement.** `classifyCapabilities()` is still called on the resolved model, but its verdict only decides whether to inject prompt-cache control — nothing rejects an incapable model before the loop runs.

Reproduction (against master):

```
gbrain config set models.subagent minimax:MiniMax-M2   # recipe declares supports_tools: false
# submit a subagent job without data.model
# → queue gate passes, handler gate never fires, tool-incapable model enters the loop
```

This bypass predates and is independent of any classification change — it lets through the existing `unusable:no_tools` class today.

## Fix

Classify the **finally-resolved** model at the handler gate, whichever branch produced it. Following the #1151 precedent, a job whose replay is already terminal (last persisted message is an assistant turn with real text and no pending tool dispatch — both the legacy `tool_use` and gateway `tool-call` vocabularies are recognized) is exempt, so re-pointing `models.subagent` after submission cannot dead-letter already-completed work.

## Tests

- Red-first: the config-path dispatch rejection fails on pre-fix code (revert-check executed).
- Terminal-replay exemption and pending-`tool-call` replay rejection are both pinned.
- `bun test test/subagent-handler.test.ts` 31 pass / 0 fail on master + this change alone; `bun run typecheck` clean.

## Scope

Deliberately small and self-contained (2 files, +137/−14). A follow-up that makes the classifier honor recipes' declared `chat.supports_subagent_loop` (currently unenforced) is implemented and ready on top of this gate; I'll open it once this lands so its review diff stays clean.

### DIFF
--- a/src/core/minions/handlers/subagent.ts
+++ b/src/core/minions/handlers/subagent.ts
@@ -48,6 +48,7 @@ import {
   logSubagentHeartbeat,
 } from './subagent-audit.ts';
 import { resolveModel, isAnthropicProvider, TIER_DEFAULTS } from '../../model-config.ts';
+import { splitProviderModelId, normalizeModelId } from '../../model-id.ts';
 import { resolveAnthropicKey } from '../../ai/anthropic-key.ts';
 import { buildSystemPrompt, DEFAULT_SUBAGENT_SYSTEM } from '../system-prompt.ts';
 import { toolLoop as gatewayToolLoop } from '../../ai/gateway.ts';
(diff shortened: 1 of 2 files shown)


Masashi-Ono0611:
### Screenshot
Screenshot 2026-08-09 at 3 48 07 PM

time-attack:
Triage bot: the gate-on-resolved-model idea is right and all three new tests pass, but an adversarial pass executed a differential repro showing this bricks a previously-working config: a BARE Anthropic id in `models.subagent` (e.g. `claude-sonnet-4-6`, no `provider:` prefix). On master that runs fine on the default legacy path (`isAnthropicProvider` has an explicit bare-id branch, and the handler's own `DEFAULT_MODEL` is the bare `'claude-sonnet-4-6'`). With this PR, `classifyCapabilities` on the resolved bare id yields `unknown` and the handler throws `references an unknown provider` for a config the rest of the codebase deliberately supports. The dream cycle hit this exact shape and normalizes before queueing (src/core/cycle/synthesize.ts:617: prefix `anthropic:` onto bare `claude-*` ids) — the same normalization belongs here before the verdict, or the gate should treat bare `claude-*` as Anthropic like `isAnthropicProvider` does. Flagging for changes rather than merging.

Masashi-Ono0611:
Fixed in a0038924 — thank you for the executed differential repro. The bare-id case was real and my gate broke it.

The gate now normalizes before classifying, and only for the verdict:

```ts
const modelForVerdict = splitProviderModelId(model).provider === null && isAnthropicProvider(model)
  ? normalizeModelId(model)
  : model;
const verdict = alreadyTerminal ? 'ok' : classifyCapabilities(modelForVerdict);
```

I reached for the existing `normalizeModelId` helper (`src/core/model-id.ts`) rather than re-inlining the `includes(':') ? … : 'anthropic:' + …` shape from synthesize.ts:617 — that inline mangles slash form (`anthropic/claude-sonnet-4-6` → `anthropic:anthropic/claude-sonnet-4-6`), which #1698 already fixed centrally.

Narrow on both axes, deliberately:

- Only bare ids that `isAnthropicProvider()` recognizes get normalized. A bare non-Anthropic id (`gpt-5`) still classifies as `unknown` and is still refused; a second new test pins that boundary.
- Normalization feeds the verdict only. The resolved `model` value is untouched, so nothing downstream moves — the legacy path still strips any prefix and calls the Messages API with the bare id, exactly as on master.

Controls executed on the new test ("a BARE Anthropic id in models.subagent still runs"):

| ref | result |
|---|---|
| merge-base f15480b9 | PASS |
| 44981e53 (the head you reviewed) | FAIL — `references an unknown provider`, verbatim |
| a0038924 | PASS |

33/33 in `test/subagent-handler.test.ts`, `bun run typecheck` clean, `bun run verify` 34/34, CI green on the branch (22 success, 1 skipped).

One residual I am not fixing here, disclosed rather than left silent: with `agent.use_gateway_loop` ON, a bare id passes this gate and then fails inside the gateway, where `parseModelId` throws "missing a provider prefix". That is master's behavior as well — this PR neither improves nor worsens it — so making bare ids work on the gateway path is a separate change, out of scope for the regression fix.

time-attack:
Merged in the v0.45.8.0 community fix-wave (#4050). Your fix landed on master as 94ec7e31e0, credited to you via Co-Authored-By, and is called out in the v0.45.8.0 CHANGELOG entry. Thank you for the contribution!

PR #3912, opened 2026-08-09
v0.42.79.0 fix(think,takes): think --take persists dual-plane; takes embeddings get a writer, dims, and an embed lane

garrytan:
## Summary

Wave 4 of the sequenced backlog plan: think-ops. The 4-agent audit re-verified all 18 planned items against current master — 10 already fixed (think gather scoping via #2739, excerpt selection via #3197, supersede lookup via #3275, model resolution, and more), 3 obsolete — leaving `think --take` (#2556), the takes-embedding hole (#2089), and a doc-truth residual. Closes #2556. Closes #2089.

**`think --take` actually persists (#2556)**
- The flag was declared and validated, then consumed by nothing — a silent no-op on CLI and MCP. `persistThinkTake` now appends the synthesis as the anchor page's next take row (claim=answer flattened + bounded at `THINK_TAKE_CLAIM_MAX_CHARS=2000` with `TAKE_CLAIM_TRUNCATED`, kind=take, holder=brain, source='gbrain think').
- **Dual-plane** via new shared `src/core/takes-append.ts`: the page's takes fence (markdown, source of truth) allocates the row number and the DB mirrors it — the red team proved a DB-only row is deterministically overwritten by the next fence-allocated `takes add` (`ON CONFLICT DO UPDATE`), so `cmdAdd` and `persistThinkTake` now share one append path that can't drift. Headless/DB-only brains fall back with an explicit `TAKE_FILE_PLANE_UNAVAILABLE` posture warning.
- Honesty contract mirrors `--save`: empty synthesis or missing anchor exits non-zero; `take_row`/`take_inserted` in `--json` and human output. Remote callers stay blocked; two `ctx.remote` forwards align to the fail-closed `!== false` pattern. CLI resolves the ambient source (env/dotfile tiers) so duplicate slugs across sources land on the caller's-context page; the op handler maps the federated grant dialect explicitly.
- Design from community PR #2618 (its successor #3164 was declined for bundled feature scope; the maintainer confirmed this half legitimate) — credited.

**Takes embeddings exist now (#2089)**
- `takes.embedding` shipped in v37 with **no writer anywhere** and a hardcoded `VECTOR(1536)` (takes isn't in schema.sql, so dim templating never reached it) — `searchTakesVector` and think's takes-recall arm were structurally dead on every brain, and non-1536 brains had a column that could never match their vectors.
[...]

PR #3901, opened 2026-08-08
v0.42.75.0 fix(pglite): in-place WAL auto-repair for the macOS Aborted() startup crash (#2575, #223, #1670)

garrytan:
## Summary

Fixes the #1 macOS deployment blocker (#2575, #223, #1670): PGLite's default engine crashing at startup with `RuntimeError: Aborted()`. The "macOS 26 WASM bug" was a misdiagnosis — an unclean shutdown (typically the OS-upgrade reboot) tears the write-ahead log, and every subsequent open dies replaying it. gbrain now repairs a torn brain in place, automatically, with data preserved.

**Core recovery (`src/core/pglite-resetwal.ts`, `pglite-repair.ts`)**
- pg_resetwal-in-TypeScript ported from the rejected upstream PR electric-sql/pglite#994 (@yestheboxer), run against gbrain's own data dir — no fork. Fail-closed PG17 layout validation, atomic+durable writes, whole-`pg_wal`-dir rename backup, pg_control CRC verification (refuses to launder corrupt counters).
- Orchestrator: validate → backup → reset → retry-once → restore-on-failure, as a never-throws engine seam.

**Auto-repair + surfaces**
- `PGLiteEngine.connect()` self-heals a `wasm-abort` on a persistent dir, prints a loud notice, keeps the lock. Kill-switch `GBRAIN_PGLITE_WAL_REPAIR=off`.
- `gbrain pglite-repair` (`--dry-run`/`--yes`/`--json`/`--path`) — the deliberate manual surface.
- `gbrain doctor` `pglite_data_dir` check diagnoses an unopenable PGLite brain from disk.
- `gbrain reinit-pglite` now defaults embedding flags from config so the rebuild rung works bare.

**Safety guardrails** (from pre-landing + Claude & Codex adversarial review)
- Repair never runs under a possibly-live writer: reaped-lock refusal, cross-process reap quarantine, `postmaster.pid` liveness refusal (protects a native Postgres dir too), ESRCH-only death verdict, atomic lock writes.
- Episode-scoped backups (never stack copies; gutted backups lose their pin; stale >24h not reused), cooldown that also caps repaired-only crash loops, honest `failed`/`failed-not-restored` reporting.
- Destructive command rejects unknown/valueless args; symlink confinement on every path component.

**Docs:** README / INSTALL / ENGINES rewritten to the real diagnosis + recovery ladder (native-Postgres recipe kept, credit @roysaurav); KEY_FILES current-state; two follow-up TODOs filed.

[...]

PR #3921, opened 2026-08-09
v0.42.80.0 fix(sync,facts): absolute repo paths at every ingress; durable facts-absorb at exit drain

### LINKED ISSUE #2108: facts:absorb fire-and-forget jobs are aborted by the 1s exit-drain — silent 100% fact loss on batch put_page ingest with non-instant chat models

### LINKED ISSUE #2114: gbrain import silently overwrites sync.repo_path — dream full-reimport fallback clobbers the configured brain repo

garrytan:
## Summary

The final combined wave (5+6) of the sequenced backlog plan. The two cluster audits found 38 of 40 remaining items already fixed on master or deliberately declined — this PR ships the two survivors plus review hardening. Closes #2114. Closes #2108. Supersedes #2530 (re-land with credit).

> **⚠ Land order:** this PR's migration is **v127**; the in-flight Wave 4 PR #3912 carries **v126**. **Merge #3912 first.** The migration runner stamps the highest applied version, so landing this first would make a later v126 silently unreachable on upgraded brains (the #2038 class). If #3912 is closed instead, this is unaffected (the runner is gap-tolerant — verified).

**Repo paths absolute at every ingress (re-land of approved PR #2530, + #2114)**
- A relative stored anchor (`sync.repo_path`, `sources.local_path`, job payloads, generated wrapper scripts) resolves against the daemon's cwd — typically `/` under launchd — and gbrain syncs the wrong tree, up to reconcile-deleting real pages (the incident class behind the mass-delete valve). The PR was approved ("Thanks — merging this") and never merged; the audit confirmed the invariant never landed by other means.
- `src/core/repo-path.ts`: `resolveRepoArg` (absolute realpath at ingress) + `requireAbsoluteStoredPath` (loud refusal at read time, remediation hint parameterized to where the value actually lives — config anchor vs `sources.local_path` vs immutable job payload). Wired at `sync --repo`/`performSyncInner`/`writeSyncAnchor`, `autopilot` (3 sites incl. generated wrappers), `jobs` handlers, `sources add --path`/`--clone-dir` (+ the overlap check), `config set sync.repo_path` (normalizes at set time), and `import`.
- Migration **v127** clears legacy relative anchors so the next sync errors with the exact fix command instead of guessing a directory. Renumbered from the PR's 120; see the land-order note above.
- #2114 folded in: `gbrain import` no longer silently repoints an existing anchor — set-if-unset, `--set-repo-path` to override, and an `anchor: kept|seeded|repointed` field in the `--json` summary.
- Credit @panda850819 (#2530).

**Durable facts-absorb for put_page (#2108)**
- A page write queues fire-and-forget facts extraction; if the process exited while the chat was in flight, the page persisted and its facts silently never landed — no error, no retry (the backstop's durable minion-job path existed but was gated to CLI one-shots).
- Remote/MCP put_page now passes `preferDurableAbsorb: ctx.remote !== false` (fail-closed); the backstop prefers the durable job when a live worker **on the right queue** exists (probe is queue-aware + 5s-cached; `readWorkers()` execs `ps` per entry, so the cache keeps it off the hot path), falling back in-process otherwise — unconditional durable submission would park jobs forever on worker-less installs.
- On a genuine drain-abort of an in-process absorb, it re-submits itself as a durable job (content-hash idempotency dedups; the job row is the retry marker — no new table) with a loud stderr line. Abort classification requires `AbortError`/`signal.aborted` — a routine Postgres "current transaction is aborted" can't masquerade as an exit drain. A bonus gap found in testing: an abort landing before the chat starts returned zeros *without throwing* — handled.

**Plus:** the `putPage` RETURNING-empty PGLite regression test (#2586 residual, the maintainer-approved minimal shape).

## Pre-Landing Review

Combined red-team/adversarial pass: 9 findings, 6 fixed in-wave:
- **Critical:** the initial worker-liveness probe was queue-blind — an install running only `jobs work --queue shell` would satisfy it, park every durable job forever, and skip the in-process fallback: the exact #2108 loss re-created. Fixed (`.some(w => w.queue === FACTS_ABSORB_QUEUE)`) with real-registry tests.
- **Critical:** v127's "every consumer errors loudly" claim was false for the extract/backlinks job handlers (`?? '.'` cwd fallback) and doctor — all four sites now refuse null/relative anchors loudly; the autopilot-cycle checkout-less contract preserved (null stays null, only relative fails).
- `config set sync.repo_path .` could re-seed the class past the migration — normalized at set time; `isAnchorOwnedSyncPath` no longer realpaths a relative anchor against cwd.
- Remediation hints parameterized per storage context; abort message-sniffing removed; import anchor decision machine-readable.
- Deferred with rationale (TODOS.md "Sync/facts follow-ups"): doctor check for waiting jobs on worker-less queues, doctor scan for pre-fix launchd wrappers carrying relative `--repo` (a migration can't reach files), minion-jobs retention trade-off (completed rows double as idempotency memory), CI guard for migration-version-exceeds-master's-max, facts-absorb PROTECTED_JOB_NAMES posture note.
- Codex remained unavailable (5-minute timeouts throughout the day); Claude adversarial passes carried cross-model duty.

## Design Review

No frontend files changed — design review skipped.

## Eval Results

No prompt-related files changed — evals skipped.

## Plan Completion

Waves 5+6 audits: 40 items — 38 fixed on master / obsolete / deliberately declined (verified with refs; banked at `~/.gstack/projects/garrytan-gbrain/wave5-audit-20260809.json` + `wave6-audit-20260809.json`), 2 survivors shipped here. Scope check CLEAN. This completes the six-wave sequence planned 2026-07-06.

## TODOS

No existing items completed. Filed "Sync/facts follow-ups" (5 items).

## Documentation

- **docs/architecture/KEY_FILES.md** — new per-file entries for `src/core/repo-path.ts` (absolute-at-ingress invariant, consumer map, migration v127) and `src/core/facts/backstop.ts` (#2108 durable facts-absorb: queue-aware probe with 5s TTL memo, content-hash idempotent re-queue on drain-abort, AbortError-only classification); updated `sync.ts`, `import.ts`, `jobs.ts`, `autopilot.ts`, `migrate.ts`, `sources-ops.ts` entries to current behavior.
- Verified current, no changes needed: README, live-sync/repo-architecture/system-of-record guides, GBRAIN_VERIFY, qm-harness, skills (all repo-path examples already absolute); progress-events (no new phases); TESTING.md.
- Gates: check-key-files-current-state green, verify 34/34, llms bundles byte-identical.

### Documentation Debt
- `--set-repo-path` (import) has reference coverage but no README how-to example (the import usage string names it).

## Test plan

- [x] Full unit suite: **16,396 pass / 2 skip / 2 flakes** — both flakes are master's own WAL-repair tests (#3901 files, untouched by this diff), standalone-green 16/16 and 25/25; the known PGLite parallel-shard contention class
- [x] `bun run typecheck` clean; `bun run verify` 34/34 (before and after the red-team batch)
- [x] Targeted: repo-path 22, facts-backstop 20 (real-registry probe tests), jobs handlers 17, migrate 165, minions 187, pglite-engine 111 — all EXIT=0
- [x] Migration v127: clears relative anchors both engines, idempotent, LIKE/regex escaping verified against raw TEXT config values


time-attack:

click once rebased.

PR #3923, opened 2026-08-09
v0.42.75.1 fix(import): reject malformed YAML frontmatter (#3708)

### LINKED ISSUE #3708: parseMarkdown silently discards frontmatter when a string value has an unquoted colon (e.g. any "Re: ..." email title)
## Summary

`parseMarkdown` (`src/core/markdown.ts:100`) silently discards an entire frontmatter block when a string value contains an unquoted `: ` (colon-space) — the single most common shape for an email reply subject (`Re: ...`). The YAML parse fails inside `gray-matter`, `parseMarkdown` swallows the error, and falls back to slug-derived title + `type: concept`. No error, no warning, and the resulting page silently looks like a "double frontmatter" corruption (the original, perfectly well-formed frontmatter block gets folded into the page body, underneath a freshly-synthesized generic wrapper), even though only one frontmatter block was ever written.

I initially spent a long time diagnosing this as an ingestion-side bug (an agent generating malformed double frontmatter) before isolating it to a pure, deterministic `parseMarkdown` parse failure — no DB, no async, no ingestion pipeline involved.

## Minimal repro

```ts
import { parseMarkdown } from './src/core/markdown.ts';

const content = `---
type: imessage
title: Text with Jane Oh re: October 24 booking
---

Body text.`;

console.log(parseMarkdown(content, 'test.md'));
// {
//   frontmatter: {},
//   compiled_truth: '---\ntype: imessage\ntitle: Text with Jane Oh re: October 24 booking\n---\n\nBody text.',
//   type: 'concept',
//   typeExplicit: false,
//   title: 'Test',   // slug-derived, not "Text with Jane Oh re: October 24 booking"
//   ...
// }
```

Quoting the value fixes it completely:

```ts
title: "Text with Jane Oh re: October 24 booking"
// -> frontmatter parses correctly, type: 'imessage', typeExplicit: true,
//    title: 'Text with Jane Oh re: October 24 booking'
```

Same mechanism reproduces with any string frontmatter field containing an unquoted colon, not just `title`. Confirmed reproducing at `v0.42.66.0` and current master (`v0.42.68.1` at time of filing) — not a recent regression.

## Real-world impact

This isn't an edge case — `Re:` is close to universal in email reply subjects, so any agent (or human) writing a raw email subject line as an unquoted `title:` will hit this on essentially every reply thread. Concretely, in a real brain I was working with, two email pages that kept appearing "corrupted" with what looked like double frontmatter — `title: Re: Ken & #10 update...` and `title: Re: Coverage Clarification & Policy Review...` — were both hitting exactly this. The agent generating the content was blamed for the corruption before I traced it back to `parseMarkdown` itself.

## Root cause

`parseMarkdown` catches the YAML parse error but has no failure path other than silently treating the whole document as bodyless:

```ts
let parsed: ReturnType<typeof matter> | null = null;
let yamlParseError: Error | null = null;
try {
  parsed = matter(content);
} catch (e) {
  yamlParseError = e as Error;
}
...
const frontmatter = (parsed?.data ?? {}) as Record<string, unknown>;
const body = parsed?.content ?? content;
```

`yamlParseError` is captured but only surfaced when `opts.validate` is explicitly passed (`gbrain frontmatter validate` / lint path). The default `parseMarkdown` call used by `gbrain put` / `put_page` (`src/core/operations.ts` → `importFromContent` → `parseMarkdown`) never passes `opts.validate`, so the parse failure is invisible at write time — the put succeeds, reports `status: "created_or_updated"`, and the page is silently wrong.

## Suggested fix

`gbrain put`/`put_page` should not succeed silently on a YAML parse failure inside the frontmatter fence. Two options, not mutually exclusive:

1. Run `opts.validate` (or at minimum surface `yamlParseError`) on the default `put`/`put_page` path, not only on the explicit `frontmatter validate` command, and reject or warn loudly when the leading frontmatter fence fails to parse as YAML.
2. Since this is specifically triggered by unquoted strings containing `: `, a pre-check that quotes ambiguous scalar values before handing content to `gray-matter` would fix it silently for the common case, though (1) is the more robust fix since it also catches other YAML syntax errors, not just this one shape.

Happy to sketch a PR for either direction if useful.

## Test cases

- unquoted string value containing `: ` anywhere → currently silently swallows the whole frontmatter block; should reject or warn, not fall back silently
- same value, quoted → already works correctly (confirms the fix direction)
- applies to any string frontmatter field, not just `title`

javieraldape:
## Summary

Fixes #3708 by making malformed YAML frontmatter fail closed during Markdown import instead of being treated as page body.

Previously, an unquoted scalar containing `: ` could make frontmatter parsing fall back to empty metadata. `put_page` and file import could then accept the write, infer a generic title/type, and store the original frontmatter fence inside `compiled_truth`.

## Changes

- Validates fenced YAML directly when `parseMarkdown(..., { validate: true })` is used, so validation does not depend only on `gray-matter` throwing.
- Makes `importFromContent` return `status: "error"` before DB work when the leading frontmatter fence has invalid YAML.
- Makes `importFile` reject invalid YAML before frontmatter inference can wrap the bad fence in synthesized metadata.
- Adds focused regression coverage for invalid unquoted `: ` frontmatter and the quoted-value success path.

## Tests

- `bun test test/import-yaml-frontmatter.test.ts`
- `bun test test/markdown.test.ts test/import-yaml-frontmatter.test.ts`
- `bun test test/import-file.test.ts`
- `bash scripts/gbrain-gate.sh <worktree> test/markdown.test.ts test/import-file.test.ts test/import-yaml-frontmatter.test.ts` - `RESULT: GREEN`; verify green; focused unit tests green with 77 pass, 0 fail, 189 expect calls; diff-mapped E2E skipped as unrelated.

### DIFF
--- a/src/core/import-file.ts
+++ b/src/core/import-file.ts
@@ -224,7 +224,7 @@ export interface ImportResult {
    * Parsed page content. Present for status='imported' AND status='skipped'
    * (skip happens when content is identical to existing page; auto-link still
    * needs to run for reconciliation in case links table drifted from page text).
-   * Absent only on status='error' (early payload-size rejection).
+   * Absent on early rejection before a page can be parsed.
    */
   parsedPage?: ParsedPage;
   /** Content-quality gate (issue #1699): true when the page landed with a
@@ -238,6 +238,13 @@ export interface ImportResult {
 }
 
 const MAX_FILE_SIZE = 5_000_000; // 5MB
+
+function invalidYamlFrontmatterError(parsed: ReturnType<typeof parseMarkdown>): string | null {
+  const yamlError = parsed.errors?.find((error) => error.code === 'YAML_PARSE');
+  if (!yamlError) return null;
+  const detail = yamlError.message.replace(/^YAML parse failed:\s*/, '').trim();
+  return `Invalid YAML frontmatter: ${detail}. Quote scalar values that contain ": " or fix the frontmatter block.`;
+}
 
 /**
  * Import content from a string. Core pipeline:
@@ -343,7 +350,14 @@ export async function importFromContent(
     };
   }
 
-  const parsed = parseMarkdown(content, slug + '.md', { activePack: opts.activePack });
+  const parsed = parseMarkdown(content, slug + '.md', {
+    validate: true,
+    ...(opts.activePack ? { activePack: opts.activePack } : {}),
+  });
+  const frontmatterError = invalidYamlFrontmatterError(parsed);
+  if (frontmatterError) {
+    return { slug, status: 'error', chunks: 0, error: frontmatterError };
+  }
 
   // v0.42 (#1699 trust boundary): strip gate-owned markers from UNTRUSTED
   // input. parseMarkdown preserves every frontmatter key except type/title/
@@ -1092,6 +1106,17 @@ export async function importFromFile(
     });
   }
 
+  const preInferenceParsed = parseMarkdown(content, relativePath, { validate: true });
+  const preInferenceFrontmatterError = invalidYamlFrontmatterError(preInferenceParsed);
+  if (preInferenceFrontmatterError) {
+    return {
+      slug: slugifyPath(relativePath),
+      status: 'skipped',
+      chunks: 0,
+      error: preInferenceFrontmatterError,
+    };
+  }
+
   // v0.22.8 — Frontmatter inference: if the file has no frontmatter and
   // inference is enabled, synthesize it from the filesystem path + content.
   // This turns bare markdown files into fully-typed, dated, tagged pages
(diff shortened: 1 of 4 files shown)


time-attack:
Triage bot: floor tests, sibling tests, verify, and CI are all green, and the #3708 fix itself is correct — but adversarial review found a real blast-radius question: a markdown doc that opens with a `---` thematic break/epigraph (not frontmatter) now fails import with a YAML_PARSE error on every sync run (verified: gray-matter throws on that shape), where master imported it with the fence folded into the body. Whether lenient\u2192fail-closed for legitimate leading-`---` corpora is acceptable, or needs a warn-only/override path, is a maintainer call. Queued for review rather than auto-merged.

time-attack:
Merged in the v0.45.8.0 community fix-wave (#4050). Your fix landed on master as 23c7b0eb16, credited to you via Co-Authored-By, and is called out in the v0.45.8.0 CHANGELOG entry. Thank you for the contribution!

PR #3949, opened 2026-08-10
fix(frontmatter): drop the shipped rule binding a personal folder to a sensitive category

Masashi-Ono0611:
## Problem

`DIRECTORY_RULES` in `src/core/frontmatter-inference.ts` ships one Apple Notes entry whose
`pathPrefix` is a single real individual's folder and whose `tags` assign a category that gbrain
itself treats as protected — `discoverTranscripts` drops transcripts matching that vocabulary
before any LLM call, and `src/core/types.ts` lists a corresponding page type.

The Privacy rule in `CLAUDE.md` says public artifacts must not reference real people, and names
comments in checked-in code as public artifacts. A shipped default rule is stronger than a
comment: it is data every install evaluates, and here it encodes a private fact about a named
person. That rule ships to everyone who installs gbrain, not only to the brain it was written for.

I hit this while reading `DIRECTORY_RULES` to work out the filename convention for importing my
own Apple Notes corpus — the rule table is the documentation for that convention, so operators do
read it.

## What changed

**Remove the rule.** This is close to behaviour-neutral for the operator whose folder it
describes. The generic `apple notes/` rule already assigns type, source and date, and derives a
tag from the first-level folder name, so title/date/type/source are unchanged and a
folder-derived tag remains. What is lost is one hardcoded semantic tag, which belongs in that
operator's own notes rather than in the default table.

Verified rather than assumed: `inferFrontmatter` on a path under that prefix returns the same
`type` / `source` / `date` / `title` before and after, and a grep for the folder segment and its
tag slug across the whole tree returns 0 remaining references.

**Add a guard so the shape cannot return.** No `DIRECTORY_RULES` entry may carry a tag in the
project's own sensitivity vocabulary. Two details worth flagging for review:

- It matches with `compileExcludePatterns`, the production compiler, not a local equality check.
  Bare words compile to `\b<word>\b`, so hyphenated and multi-word spellings are caught the same
  way the runtime catches them. An exact-equality guard let those spellings through with CI green
  (raised in review before this landed). Boundary semantics are production-identical and no
  wider — `psychotherapy` is not caught, exactly as the runtime filter treats it.
- It reports rule **indices**, not prefixes. An offending prefix is by definition the kind of
  string this test exists to keep out of public artifacts, and a failing assertion is printed into
  CI logs.

To make that matching possible without hand-copying the vocabulary, the inline default in
`synthesize.ts` moves to an exported `DEFAULT_EXCLUDE_PATTERNS` beside the compiler it feeds.
Runtime behaviour is unchanged: same two entries, and `dream.synthesize.exclude_patterns` still
overrides it.

## Verification (on 130d321d / v0.42.76.0)

- `bun test test/frontmatter-inference.test.ts test/page-type-exhaustive.test.ts test/cycle-synthesize.test.ts` → **85 pass / 0 fail**
- Red check: restoring the rule from `upstream/master` → **38 pass / 1 fail**. (My first red check was
  void — it stashed against this branch's own commit, so the rule was never actually restored. Redone
  against master.)
- Guard boundary behaviour, measured: `therapy` / `therapy-notes` / `medical records` / `Therapy`
  caught; `psychotherapy` / `gstack` / `yc` / `pitch-notes` not caught.
- `bun run typecheck` → clean
- `bun run verify` → 34/34 green
- `bun run build:flag-registry` → no diff (the new import introduces no flag literals)

## Not included

There are also five Apple Notes rules whose output is byte-identical to what the generic rule
already produces — dead entries. Deliberately left out so this PR stays about the one thing;
happy to send that cleanup separately if useful.

### DIFF
--- a/src/core/cycle/synthesize.ts
+++ b/src/core/cycle/synthesize.ts
@@ -40,7 +40,7 @@ import { MinionQueue } from '../minions/queue.ts';
 import { waitForCompletion, TimeoutError } from '../minions/wait-for-completion.ts';
 import { makeSubagentHandler } from '../minions/handlers/subagent.ts';
 import type { MinionJobInput, MinionJobContext, MinionHandler, SubagentHandlerData } from '../minions/types.ts';
-import { discoverTranscripts, type DiscoveredTranscript } from './transcript-discovery.ts';
+import { discoverTranscripts, DEFAULT_EXCLUDE_PATTERNS, type DiscoveredTranscript } from './transcript-discovery.ts';
 import { serializeMarkdown, serializePageToMarkdown } from '../markdown.ts';
 import type { Page, PageType } from '../types.ts';
 import { validateSourceId } from '../utils.ts';
@@ -836,7 +836,7 @@ async function loadSynthConfig(engine: BrainEngine): Promise<SynthConfig> {
     DEFAULT_SUBAGENT_WAIT_TIMEOUT_MS,
   );
 
-  let excludePatterns: string[] = ['medical', 'therapy'];
+  let excludePatterns: string[] = [...DEFAULT_EXCLUDE_PATTERNS];
   if (excludeStr) {
     try {
       const parsed = JSON.parse(excludeStr);
--- a/src/core/cycle/transcript-discovery.ts
+++ b/src/core/cycle/transcript-discovery.ts
@@ -120,6 +120,14 @@ export function isDreamOutput(content: string, bypass = false): boolean {
 }
 
 /**
+ * Sensitivity categories dream excludes from transcript ingestion by default.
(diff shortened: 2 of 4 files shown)


time-attack:
Triage bot: passed all gates (floor tests, sibling tests, verify, two independent adversarial reviews) but now conflicts with #3952 which merged first — both touch frontmatter-inference.ts and its test. Needs a rebase; will re-attempt merge on the next run once rebased.

time-attack:
Merged in the v0.45.8.0 community fix-wave (#4050). Your fix landed on master as a849d833eb, credited to you via Co-Authored-By, and is called out in the v0.45.8.0 CHANGELOG entry. Thank you for the contribution!

Issue #2964, opened 2026-07-18
[bug] dream cycle sync phase fails nightly on a never-git-initialized legacy brain dir; doctor's 'ok' is a false-negative coincidence


Masashi-Ono0611:
## Problem

On a legacy single-source personal brain (pre-v0.18 default, resolved via the
`sync.repo_path` config key rather than a `sources` table row — see
`resolveBrainDir` in `src/commands/dream.ts`), the nightly `dream cycle`'s
`sync` phase fails **every single run** with:

```
[InternalError/UNKNOWN] Not inside a git repository: <brainDir>. GBrain sync
requires a git-initialized repo (or a subdirectory of one).
```

(older gbrain versions before #2942 surfaced the same failure as `Not a git
repository: <path>. GBrain sync requires a git-initialized repo.`)

The brain dir itself was never `git init`-ed (e.g. it predates git-backed
sync, or was rsync'd from another machine without its `.git`), so
`discoverGitRoot` (`src/commands/sync.ts`) throws unconditionally, and the
dream cycle reports `✗ sync failed` in the nightly log — chronically, night
after night, with no self-recovery.

## Why this is confusing to debug

`gbrain doctor`'s `sync_freshness` / `sync_consolidation` checks
(`src/commands/doctor.ts`) report "ok / N/A" for the exact same brain, which
reads as "sync isn't needed here" — but that's a coincidence, not a
diagnosis: those checks key off `SELECT ... FROM sources WHERE local_path IS
NOT NULL`, and a legacy `sync.repo_path`-based brain has zero rows in the
`sources` table. Doctor's "ok" isn't verifying anything about the actual
sync phase's git precondition — it's just not looking at this brain dir at
all. So `doctor` says everything's fine while the nightly cycle log says
sync is failing, for unrelated reasons.

## Impact

Sync isn't a federation-only mechanism — it's the general-purpose importer
that picks up local markdown edits made directly under the brain dir. A
brain stuck in this state silently stops ingesting hand-edited files (while
still working fine for content added via `gbrain put_page`/ingest, which
don't go through this path), with only a nightly "sync failed" log line
(easy to miss / normalize-away) as the symptom.

## Suggested fix direction

Self-heal the common case: when `discoverGitRoot` fails for the *default*
sync path (no `--source` / no registered `sources` row — i.e. gbrain owns
this directory outright, unlike a locally-registered external
`sources add --path` entry a user pointed gbrain at), run `git init` in
[...]

PR #3964, opened 2026-08-10
fix(sync): never write a baseline commit over an already-populated repo

NidTamil:
## Problem

`createSyncBaselineCommit` (the #2964 sync self-heal) is meant to write the *first* commit of a genuinely unborn/uninitialized repo. But both of its call sites infer "unborn" from a **failure to observe** git state — site 1 when `discoverGitRoot` throws, site 2 when `git rev-parse HEAD` throws — and site 1 then runs `git init --quiet` (a no-op reinit on an already-initialized repo) followed by an **unconditional** baseline commit.

Those probes also fail *transiently* on a fully-populated repo: a 30s timeout on a large brain, or a concurrent `gbrain sync` holding a git lock. When that happens the transient failure is mistaken for "empty repo", and the self-heal stacks a spurious `gbrain: initial commit (auto-init by sync)` commit on top of real history. Because the baseline does `read-tree --empty` + `git add -A`, on a case-insensitive filesystem (Windows / default macOS) it also **re-cases the whole tree** (e.g. `projects/` → `Projects/`), corrupting tracked paths.

Observed in the wild: two stacked auto-init commits plus a full-tree case-flip on a populated repo, recovered by hand.

## Fix (defense in depth)

1. **Fail-closed chokepoint guard** in `createSyncBaselineCommit`: probe `git rev-parse --verify --quiet HEAD` and proceed **only** on a clean unborn signal (exit 1, no signal, empty stderr). A born HEAD *or* an ambiguous probe failure (timeout / lock / other) both refuse. Because the guard's own probe is subject to the same transient failures as the callers, failing closed on *ambiguity* (not just born-HEAD) is what makes the backstop real. The born/unborn/ambiguous decision is a pure, unit-tested helper `classifyHeadProbeError`.
2. **Site-1 re-probe**: on a `discoverGitRoot` throw, re-probe once before healing — resolves → use it (covers the subdir-anchored `#753/#774` monorepo case); still throws but `.git` present → surface the original error (never heal over a corrupt/unreadable repo); genuinely non-git → `git init` + baseline. Also closes a pre-existing nested-repo-shadowing edge for subdir anchors.

One non-incident edge is documented in-code and left as-is: an orphan branch (`git switch --orphan`) in a repo with history elsewhere probes as unborn (no stacking on real history, no re-case of other branches).

## Tests

- **Regression**: `createSyncBaselineCommit` on a populated repo now refuses (HEAD unmoved, no re-cased paths).
- **Site-1 fail-closed**: a `.git` gitlink pointing nowhere → both probes throw → refuse, no init/commit.
- **Pure unit file** for `classifyHeadProbeError` (unborn / timeout / lock / malformed inputs), no fault injection.

`createSyncBaselineCommit` + `classifyHeadProbeError` are exported for the direct unit tests. Existing `sync-git-autoinit` / `sync-discover-git-root-stderr` / `sync-monorepo` suites stay green and `typecheck` is clean on current `master`.

I left the `VERSION` / `CHANGELOG` bump out so you can slot it into a release wave — happy to add one if you'd prefer.

### DIFF
--- a/src/commands/sync.ts
+++ b/src/commands/sync.ts
@@ -1038,7 +1038,71 @@ export function discoverGitRoot(inputPath: string): string {
  * gbrain.yml, or a semantic overlap) propagates — better to leave this
  * self-heal wedged with a clear error than commit unknown content.
  */
-function createSyncBaselineCommit(repoPath: string): void {
+/**
+ * Classify a caught `git rev-parse --verify --quiet HEAD` failure for the
+ * baseline-commit guard. A genuinely UNBORN HEAD makes git exit with status
+ * EXACTLY 1 and nothing on stderr — precisely what `--verify --quiet` emits
+ * for an unresolvable HEAD (verified empirically: unborn => exit 1, empty
+ * stderr; born => exit 0). Every OTHER failure shape — a 30s timeout (killed
+ * by signal, so `status` is null), an index/ref lock (`fatal: Unable to
+ * create ...lock`, non-empty stderr), or any nonzero-but-not-1 exit — is NOT
+ * proof the repo is empty; it is the transient-probe-failure class that
+ * corrupted the live brain on 2026-08-10. Return 'unborn' ONLY for the clean
+ * signal so the caller fails CLOSED on 'ambiguous'. Pure (no I/O) so the
+ * distinction is unit-testable without fault injection.
+ */
+export function classifyHeadProbeError(err: unknown): 'unborn' | 'ambiguous' {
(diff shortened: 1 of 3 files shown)


time-attack:
Triage bot: first-time contributor + 200-line change on the sync baseline path, and CI has not run at this head. Routed to maintainer review (HIGH tier) — no autonomous action taken.

time-attack:
Merged in the v0.45.8.0 community fix-wave (#4050). Your fix landed on master as 0a1890bbf8, credited to you via Co-Authored-By, and is called out in the v0.45.8.0 CHANGELOG entry. Thank you for the contribution!

PR #3971, opened 2026-08-10
fix(embed): retry 502 Bad Gateway during embed --stale backfill (#3966)

### LINKED ISSUE #3966: embed --stale crashes on 502 Bad Gateway instead of retrying (free-tier NIM intermittent outages kill bulk backfill)
## Summary

`gbrain embed --stale` terminates the process on HTTP 502 Bad Gateway responses from the NVIDIA NIM embedding endpoint, rather than retrying the way it does for 429 rate-limit responses. On the free tier, NIM returns 502 intermittently under sustained bulk load (backend queue full / transient overload), so a long embed backfill job reliably dies mid-run with no recovery.

## Environment

- gbrain: 0.42.75.0
- Provider: `nvidia:nvidia/nv-embed-v1`
- Mode: `gbrain embed --stale --catch-up --batch-size 32`
- DB: PGLite
- ~54,000 stale chunks to embed (bulk backfill after embedding provider was misconfigured)

## What happens

After processing several hundred to a few thousand pages successfully, the process receives a 502 and exits:

```
[embed.pages] 2200/6760 (32%)
...
  [rate-limit] attempt 1/5, waiting 52448ms...   ← 429s are retried
  Error embedding 0-inbox/personal/2025-11-04-...: [embed(nvidia:nvidia/nv-embed-v1)] Bad Gateway
                                                   ← 502 is fatal; process exits here
```

The 502 is transient — the same endpoint returns 200 immediately after a fresh restart, confirming it's not a credentials or content issue. But gbrain's retry loop only handles 429; a 502 falls through to a crash path.

## Expected behavior

502 should be treated as a retriable transient error with the same backoff logic as 429. At minimum, `embed --catch-up` / `embed --stale` should log the failure, skip or defer the problem chunk, and continue — not exit the process.

## Workaround

Wrapping the command in a shell restart loop that re-invokes `gbrain embed --stale` after each crash (and clears the PGLite lock left behind) recovers correctly. The lock cleanup step is necessary because the process exits without releasing it:

```bash
while true; do
    REMAINING=$(gbrain embed --stale --dry-run 2>/dev/null | grep -oP '(?<=Would embed )\d+')
    [ -z "$REMAINING" ] || [ "$REMAINING" -eq 0 ] && break
    # clear stale PGLite lock if needed
    rm -rf /path/to/brain.pglite/.gbrain-lock
    gbrain embed --stale --catch-up --batch-size 32
    sleep 30
done
```

## Notes

- 429 responses log `[rate-limit] attempt N/5, waiting Xms...` and are retried correctly — the same behaviour for 502 would fix this entirely
- The PGLite lock is not released on unclean exit; a crash + lock means the next invocation also fails until the lock is manually removed

pmn4:
## Summary
- Extend `embedBatchWithBackoff` to retry transient HTTP 502/503/504 gateway errors with the same backoff loop already used for 429 rate limits
- Prevents bulk `gbrain embed --stale --catch-up` runs from permanently failing chunks on intermittent NVIDIA NIM outages

Fixes #3966

## Test plan
- [x] `bun test test/embed.serial.test.ts` — new 502 retry case passes
- [x] `bun test test/embed-partial-failure-3037.serial.test.ts` — sustained 502 retries batch but does not fan out per-chunk
- [x] `bun run verify` — full pre-push gate green

### DIFF
--- a/src/commands/embed.ts
+++ b/src/commands/embed.ts
@@ -1352,12 +1352,14 @@ async function embedAllStale(
 
 /**
  * v0.33.3: rate-limit-aware embedBatch wrapper.
+ * #3966: also retries transient gateway errors (502/503/504) that NIM and
+ * similar providers emit under sustained bulk load.
  *
  * The OpenAI SDK has built-in retry with exponential backoff, but its
  * backoff window (max ~4s) is too short for TPM (tokens-per-minute)
  * rate limits on large pages (~90K tokens).  This wrapper catches
- * 429-shaped errors, parses the retry delay from the error message
- * (e.g. "Please try again in 248ms"), and sleeps before retrying.
+ * 429-shaped and gateway-overload errors, parses the retry delay from
+ * the error message (e.g. "Please try again in 248ms"), and sleeps before retrying.
  *
  * v0.33.4 hardening (codex + re-review findings):
  *   - D4: detect 429 via the wrapped error's `cause.status` (the gateway's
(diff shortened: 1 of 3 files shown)


time-attack:

PR #3977, opened 2026-08-10
fix(chunker,edges): Dart code intelligence — grammar ABI 15 vs pinned runtime, and no call graph (#3356)

### LINKED ISSUE #3356: code-def/callers/callees silently return 0 for all Dart symbols — bundled Dart grammar is tree-sitter ABI 15, but pinned web-tree-sitter@0.22.6 supports only ABI 13–14
## Summary

On any Dart/Flutter repo, `gbrain code-def <symbol>` returns 0 for **every** symbol — top-level functions *and* classes — even though the source is fully indexed and semantic search works. Same for `code-callers` / `code-callees`. The root cause is a **tree-sitter ABI mismatch**: the bundled `tree-sitter-dart.wasm` is compiled to **ABI 15**, but gbrain pins `web-tree-sitter@0.22.6`, whose runtime supports only **ABI 13–14**. `parser.setLanguage(dartGrammar)` throws, the throw is swallowed by a silent `catch`, and every `.dart` file degrades to text chunks with `symbol_name = NULL` — which is exactly the column `code-def` queries.

This is **not** the "missing `TOP_LEVEL_TYPES['dart']` entry" it looks like at first glance — that map is real but **downstream and unreachable**, because `setLanguage` throws before the map is ever consulted.

## Impact

- **Every Dart/Flutter repo** loses `code-def` / `code-callers` / `code-callees`. `code-refs` and semantic `search` / `query` are unaffected (they scan `chunk_text`, not `symbol_name`).
- **Silent** — no error, no warning. It looks like a successful index (`0 errors`), so users burn time force-reindexing before realizing the symbols were never extracted.
- A full grammar audit (below) shows the silent-failure surface is **dart + elm + ql** (3 of 36 shipped grammars).

## Root cause (with file:line)

Pipeline in `src/core/chunkers/code.ts`:

1. `.dart → 'dart'` is correctly mapped (`code.ts:453`), and the grammar is wired + present on disk (`code.ts:233` `dart: { … embeddedPath: G_DART }`; `src/assets/wasm/grammars/tree-sitter-dart.wasm`, 962 KB).
2. `Language.load(dartGrammar)` **succeeds** (this is what makes it look fine — and why semantic search works).
3. **`parser.setLanguage(grammar)` at `code.ts:599` THROWS** `Incompatible language version 15. Compatibility range 13 through 14.` — and it sits *outside* the inner `try` (which starts at `code.ts:601` and only guards `parseWithTimeout`).
4. The throw propagates to the **outer `catch` at `code.ts:722`**, which is a bare `} catch {` with **no logging** and simply `return { chunks: fallbackChunks(source, filePath, language, opts), edges: [] }`.
5. `fallbackChunks` emits text chunks with `symbolName: null, symbolType: 'module'` (`code.ts:907`). Execution **never reaches** the `TOP_LEVEL_TYPES[language]` lookup at `code.ts:615`, so the missing `dart` map key is irrelevant.
6. `code-def` queries `WHERE cc.symbol_name = $1 AND cc.symbol_type IN (…DEF_TYPES…)` (`src/commands/code-def.ts:70-74`) — no re-parse, it trusts the stored `symbol_name`. For Dart every chunk has `symbol_name = NULL` → 0 results.

`code-refs` survives because it matches `AND cc.chunk_text ILIKE $1` (`src/commands/code-refs.ts:61`), independent of `symbol_name`.

## Reproduction (deterministic)

Using gbrain's own embedded runtime + grammars (`web-tree-sitter@0.22.6`):

```js
import mod from 'web-tree-sitter';
const P = mod.default || mod;
await P.init({ locateFile: () => './src/assets/wasm/tree-sitter.wasm' });
for (const g of ['dart', 'typescript', 'swift']) {
  const lang = await P.Language.load(`./src/assets/wasm/grammars/tree-sitter-${g}.wasm`);
  const p = new P();
  try { p.setLanguage(lang); p.parse('class A {}'); console.log(g, 'abi', lang.version, 'OK'); }
  catch (e) { console.log(g, 'abi', lang.version, 'THREW:', e.message); }
}
```

```
dart        abi 15  THREW: Incompatible language version 15. Compatibility range 13 through 14.
typescript  abi 14  OK
swift       abi 13  OK
```

## Full grammar audit (bonus — the complete silent-failure surface)

Loading + `setLanguage` on all 36 shipped grammars under the pinned 0.22.6 runtime:

- **OK (34):** bash, c, c_sharp, cpp, css, elisp, elixir, embedded_template, go, html, java, javascript, json, kotlin, lua, objc, ocaml, php, python, rescript, ruby, rust, scala, solidity, sql, swift, systemrdl, tlaplus, toml, tsx, typescript, vue, yaml, zig (all ABI 13–14).
- **BROKEN (3):**
  - `dart` — ABI **15**, `setLanguage` throws (this issue).
  - `elm` — `Language.load` fails: `Out of bounds memory access`.
  - `ql` — `Language.load` fails: `Out of bounds memory access`.

## Why bumping `web-tree-sitter` alone does NOT fix it

I tested the current grammars against `web-tree-sitter@0.26.11` (ABI range 13–15, which *would* accept the ABI-15 Dart grammar). **All grammars fail to load** under 0.26 — `typescript`, `swift`, and `dart` all throw a `dylink metadata` error. The `tree-sitter-wasms@0.1.13` grammar wasms were built for the older loader and are incompatible with the ≥0.25 runtime. So a runtime bump requires a **coordinated rebuild of the entire grammar set**, not a one-line dependency change.

## Proposed fix

**Primary (minimal, Dart-only, no runtime change):** ship `tree-sitter-dart.wasm` **rebuilt at ABI ≤ 14** (matching the pinned 0.22.6 runtime — e.g. build tree-sitter-dart with a 0.22.x-era `tree-sitter build --wasm`), then add the two map entries:

```ts
// src/core/chunkers/code.ts — TOP_LEVEL_TYPES (~line 283)
dart: new Set([
  'class_definition', 'enum_declaration', 'mixin_declaration',
  'extension_declaration', 'type_alias', 'function_signature',
]),
```

```ts
// src/commands/code-def.ts — DEF_TYPES (~line 48): add so mixin/extension surface
// (normalizeSymbolType maps *_declaration → "mixin declaration" / "extension declaration")
'mixin declaration', 'extension declaration',
```

`class_definition → class`, `enum_declaration → enum`, `function_signature → function`, `type_alias → type` are already covered by `normalizeSymbolType` (`code.ts:1189`) + existing `DEF_TYPES`.

> ⚠️ The Dart node-type names above are the canonical tree-sitter-dart shape and should be **confirmed against the ABI≤14 grammar you build** — I couldn't parse a `.dart` file to verify them (the ABI-15 wasm is rejected by both the 0.22 and 0.26 runtimes). In particular, top-level Dart functions are known to parse as `function_signature` (+ a sibling `function_body`) rather than a single wrapper — verify `childForFieldName('name')` resolves on whichever node carries the name.

Also rebuild the `elm` and `ql` grammars (separate load failure).

**Alternative (heavier, fixes elm/ql/dart at once):** bump `web-tree-sitter` to ≥0.25 **and** replace `tree-sitter-wasms` with a matching, freshly-built grammar set + swap the embedded runtime asset `src/assets/wasm/tree-sitter.wasm`; then re-validate all 34 currently-working grammars.

## Independent robustness fix (recommended regardless)

The outer catch at `src/core/chunkers/code.ts:722` is `} catch {` with no logging. A single `console.warn` there (grammar/`setLanguage` failure → falling back to text chunks for `<file>`) would have surfaced all three failures immediately and guards against silent ABI drift for future grammars.

## Environment

gbrain `0.42.64.0` · `web-tree-sitter@0.22.6` · `tree-sitter-wasms@0.1.13` · Bun 1.3.14 · Linux arm64. Verified 2026-07-24.

Jey2311:
Fixes #3356.

## The bug, in one line

The bundled `tree-sitter-dart.wasm` is tree-sitter ABI **15**; `web-tree-sitter` is pinned at **0.22.6**, which accepts **13–14**. `Language.load()` *succeeds*, so the grammar looks wired and semantic search keeps working — but `parser.setLanguage()` throws, lands in a bare `catch`, and every `.dart` file degrades to text chunks with `symbol_name = NULL`. `code-def` reads that column, so it returned **0 for every Dart symbol in every Dart repo**, silently, on an index reporting `0 errors`.

Reproduced on `v0.44.0.0` with the repo's own runtime and grammars:

```
dart        abi 15  THREW: Incompatible language version 15. Compatibility range 13 through 14.
typescript  abi 14  OK
swift       abi 13  OK
```

## Why not bump the runtime

Because it is not a one-line dependency change. Under `web-tree-sitter@0.26.11` the **entire** vendored grammar set fails to load with a `dylink metadata` error — `typescript` and `swift` included — so bumping the runtime means rebuilding all 36 grammars in the same PR. This PR keeps the runtime pinned and fixes the one grammar that disagrees with it.

## What this changes

**1. The grammar, regenerated at ABI 14.** Same upstream source `tree-sitter-wasms` already builds from (`UserNobody14/tree-sitter-dart`, pinned at `be07cf7`), regenerated with `tree-sitter generate --abi 14`. `scripts/build-dart-wasm.sh` reproduces the committed binary, so the blob is auditable rather than trusted.

Measured against **214 real Dart files** (a Flutter app + a shared package):

| grammar | files parsed | parse errors |
|---|---|---|
| current (ABI 15) | **0** — `setLanguage` throws | n/a |
| npm `tree-sitter-dart@1.0.0` (ABI 14) | 214 | **760**, across 79 files |
| this PR (`be07cf7`, `--abi 14`) | 214 | **0** |

The middle row is worth calling out: the npm package publishes a pre-generated ABI-14 `parser.c`, so it looks like the obvious zero-build fix. It is an older grammar that predates Dart 3 — it fails on `library;`, `abstract interface class`, and switch expressions. It passed every unit test I had before I ran it against real code.

**2. `TOP_LEVEL_TYPES.dart` + the matching `code-def` `DEF_TYPES` entries.** `class_definition` / `enum_declaration` / `type_alias` / `function_signature` already normalize into existing `DEF_TYPES` via `normalizeSymbolType`; `mixin_declaration`, `extension_declaration`, `getter_signature` and `setter_signature` fall through `type.replace(/_/g, ' ')` and needed listing. `DEF_TYPES` is hoisted to module scope (verbatim, no edits) so a test can pin both halves — the chunker emitting `"mixin declaration"` while `DEF_TYPES` omits it is a green pipeline returning 0 rows, which is the shape of the original bug.

**3. A Dart-specific chunk-range fix that #3356 does not mention.** Dart is the only shipped grammar that makes `function_body` a **sibling** of the signature rather than a child:

```
[function_signature] "int topLevel(int a)"
[function_body]      "=> a + 1;"
```

Chunks are built solely from `semanticNodes` and source not covered by one is never emitted, so adding `function_signature` to the map *alone* indexes every top-level function as its signature and **drops the body**. Measured on a real file: `clampNameSlot` came out as lines 410–417 instead of 410–442. The symbol *count* is identical either way, so the test asserts on the body text rather than on a count.

**4. The silent fallback now warns, once per language.** This is why the bug survived: `} catch { return fallbackChunks(...) }` with no logging is indistinguishable from "this language has no semantic nodes". Note `elm` and `ql` still fail to *load* under the pinned runtime today, just as quietly — this PR does not fix them, but it makes them audible.

## Testing

`test/chunkers/code-dart.test.ts`, 5 tests. Every one was mutation-tested rather than assumed:

| mutation | result |
|---|---|
| restore the ABI-15 grammar | 3 fail |
| disable body absorption | 1 fail |
| remove `dart` from `TOP_LEVEL_TYPES` | 2 fail |
| remove the Dart `DEF_TYPES` entries | 1 fail |
| remove the `warnParseFailure` call | 1 fail |
| remove the warn-once dedup | 1 fail |

`bun run verify` — 34/34 green. `bun test test/chunkers/` — 76 pass. `check:wasm` (which compiles the binary and asserts real semantic chunks) passes.

Two things I did **not** verify, stated rather than glossed:

- `test/code-def-refs.test.ts` fails in my environment with a `beforeEach` hook timeout. It fails identically on a pristine `origin/master`, so it is a missing local Postgres, not this change — but I have not seen it green.
- Files whose declarations are all small still merge into one anonymous chunk with `symbolName: null` at the default `chunkSizeTokens: 300` (`mergeSmallSiblings`). That is pre-existing, language-agnostic behaviour, unchanged here; it means "every Dart symbol is now extractable" is true per-declaration, not per-file.

---

## Second commit: the call graph (`code-callers` / `code-callees`)

The first commit fixes `code-def`. It does **not** fix `code-callers`/`code-callees`, which #3356's title also covers — I verified that on a patched install rather than assuming, and initially proposed it as a follow-up. It turned out small enough to belong here, so this PR now closes the issue rather than half of it.

`CALL_CONFIG` names a call NODE type and reads the callee from a field or the first named child. `tree-sitter-dart` has **no call node**: an invocation is a flat run of siblings with the argument list to the *right* of the name.

```
bare(1)         identifier "bare" · selector(argument_part)
obj.method(2)   identifier "obj" · selector(unconditional_assignable_selector
                  identifier "method") · selector(argument_part)
obj..m()        cascade_section(cascade_selector identifier "m", argument_part)
new Widget(2)   new_expression(type_identifier "Widget", arguments)
```

So `extractDartCallEdges` finds the argument list and reads the nearest name to its left, with cascades and `new_expression` handled directly since they do not follow that rule. Dart is deliberately **not** added to `RECEIVER_RESOLUTION_LANGS` — bare-token emit, the same scope Ruby/Go/Rust/Java/Kotlin already have.

**Measured before:** `chunkCodeTextFull` returned 0 edges for a 442-line Dart file and 94 for a comparable Python one; `edges-backfill` reported `341 chunks walked, 0 resolved, 0 ambiguous, 0 unmatched` — nothing to resolve, rather than resolution failing.

**Measured after:** 10,755 call edges across 214 real Dart files, 193 with at least one. End-to-end on a patched install, after `reindex-code` + `edges-backfill`: `measureLabel` 8 callers / 5 callees, `styleFor` 10 / 2, `BambooApiClient` 8 / 23, and `clampNameSlot → measureLabel` is a real call in the source. Those edges come back `resolved: false` — as do Python's on the same brain (`lambda_handler`: 0 of 49 resolved), so this is parity with an already-supported language, not a Dart shortfall.

### One defect worth naming, because only mutation testing found it

My first version walked left through *any* non-naming selector looking for a name. On `a()()` — one call to `a`, then a call on its anonymous result — that emitted `['a', 'a']`: two edges for one definition. All eleven other shapes passed, and the suite was green. A selector that names nothing now stops the walk, and `a()()` → `['a']` / `list[0]()` → `[]` is pinned by its own test.

### Testing

`test/dart-call-edges.test.ts`, 14 cases covering bare / method / null-aware / constructor / named constructor / explicit `new` / generic / awaited / chained / cascade / nested-argument calls, plus the anonymous-result and non-call-selector cases. Six mutations, six distinct failures:

| mutation | result |
|---|---|
| Dart never dispatched to the new extractor | 13 fail |
| cascade branch removed | 1 fail |
| `new_expression` branch removed | 1 fail |
| anonymous-result guard removed | 1 fail |
| emit on **any** selector, not just argument-bearing | 6 fail |
| `conditional_assignable_selector` unhandled | 1 fail |

`bun run verify` 34/34 green on both commits; `test/edge-extractor.test.ts` still passes, so the other nine configured languages are untouched.


Jey2311:
Scope correction on my own claim, before a reviewer has to find it.

This PR says "Fixes #3356", and that issue's title covers `code-def` **and** `code-callers`/`code-callees`. **It only fixes `code-def` (and `code-refs`, which already worked).** `code-callers`/`code-callees` still return 0 for Dart after this change — I verified that on a patched install rather than assuming the fix was complete.

The reason is unrelated to the ABI mismatch, and I want to save the next person the detour I took. My first guess was that the edges just needed a `dream` pass. Wrong, and `edges-backfill` settles it without spending an LLM call:

```
[edges-backfill] source=bamboo-and-chill done: 341 chunks walked, 0 resolved, 0 ambiguous, 0 unmatched
```

`0 unmatched` is the decisive number — failed resolution would have counted unmatched edges. There were simply no raw edges to resolve. Confirmed at the source: `chunkCodeTextFull` returns **0 edges** for a 442-line Dart file and **94** for a comparable Python one.

The cause is `CALL_CONFIG` in `src/core/chunkers/edge-extractor.ts`, which lists **10 of 36 languages** — so callers/callees are empty for 26 languages, not for Dart in particular.

And Dart is not a one-line addition to that map. Every configured language has a call *node* whose callee is a field or the first named child. `tree-sitter-dart` has no call node at all — `bare(1)` parses as flat **siblings** inside `expression_statement`:

```
expression_statement
  identifier "bare"
  selector
    argument_part
      arguments
```

with `obj.method(2)` inserting a `selector > unconditional_assignable_selector > identifier` between the receiver and the argument list. The current `CallConfig` shape (`callNodeTypes` + `calleeFieldName` / `calleeFirstNamedChild`) cannot express that; it needs a new extractor form that walks preceding siblings. That felt like a different PR than a grammar-ABI fix, so I left it out rather than widen this one — happy to follow up if you'd want it, and equally happy to retitle this PR or drop the `Fixes` keyword to `Refs` if you'd rather #3356 stay open until callers/callees work too.

Jey2311:
Retracting the scope caveat above: **this PR now covers `code-callers`/`code-callees` too**, so `Fixes #3356` is accurate again and no retitle is needed.

The follow-up I described turned out to be ~90 lines, and the diagnosis in that comment is what it is built on — `tree-sitter-dart` has no call node, so the callee is found positionally instead of structurally. Details and measurements are in the updated description; the second commit stands alone and is easy to drop if you would rather take the grammar fix by itself.

One correction to what I wrote above, from actually building it: I said adding Dart "needs a new extractor form that walks preceding siblings". The walk turned out to be the trap rather than the solution — walking left through any non-naming selector double-counts `a()()` as two calls to `a`. The rule that works stops at the first selector that names nothing.

time-attack:
Automated triage: this adds new product surface, so it is queued for the maintainer's own product decision rather than being auto-actioned. Labeled `triage:new-feature`.

time-attack:

PR #3980, opened 2026-08-11
fix(config): read the config table once instead of once per key

hexsprite:
## The problem

`gbrain stats` against a Supabase-hosted brain took **7.7s**. `pg_stat_statements` for that single run:

- 54 queries
- **2.3ms** of total server time
- 44 of them `SELECT value FROM config WHERE key = $1`

The database does no work. The CLI pays one network round trip per config key, sequentially, before the command starts.

The same command against a local PGLite brain takes 0.26s. That is why this survived: on an embedded database each read costs microseconds and the pattern is invisible. It only appears once the brain is a network hop away.

## Where the reads come from

Two startup paths, both on every command:

| Path | Reads | Cost on a hosted brain |
|---|---|---|
| `loadConfigWithEngine()` | ~24 keys, DB plane merged over file/env | ~1.5s |
| `reconfigureGatewayWithEngine()` | 6 models x a precedence chain of up to 4 keys, plus alias expansion | **3.1s** |

The gateway path was the bigger half and the less obvious one — `resolveModel` looks cheap at the callsite, and it is called six times in a loop.

## The fix

The config table is a handful of rows, so read it once and answer every key from memory.

- `engine.getAllConfig()` — one `SELECT key, value FROM config`, implemented on both engines (parity).
- `src/core/config-snapshot.ts` — wraps it as either a raw map (`loadConfigSnapshot`) or a `ConfigReader` (`snapshotConfigReader`).
- `resolveModel` / `resolveAlias` now take a `ConfigReader` instead of a `BrainEngine`. `getConfig` is the only method they ever called, so this is a widening, not a break.

## What does not change

- Same keys, same precedence (env > file > DB > defaults), same per-key quiet failure.
- The snapshot lives for **one resolution pass**, not for the process. A `gbrain config set` from another process is picked up by the next command, exactly as before.
- When the bulk read is unavailable — a brain mid-migration with no config table, or an engine implemented outside this repo — every read falls back to the old per-key path.

## Result

Measured against a Supabase brain (transaction pooler, us-east), median of 3:

```
gbrain stats   7.7s  ->  1.9s
```

Remaining time is connect (~0.5s), the migration probe (~0.15s), CLI boot (~0.5s), and the actual work.

## Tests

`test/config-db-plane-round-trips.test.ts` pins the **round-trip count**, not the wall clock — a timing assertion would be flaky in CI, while a call counter is deterministic. Seven cases: one bulk read instead of per-key; prefix maps resolved from the same snapshot; precedence unchanged; file/env still beats the DB plane; an engine without `getAllConfig` still resolves every key; a throwing bulk read degrades to per-key; a brain with no config table at all still loads.

- `bun run typecheck` — clean
- Config, gateway, and model-resolution suites pass

## Note for reviewers

I went at this with `Promise.all` batching first. It did help (7.7s -> 5.5s), but it was the wrong lever: postgres.js opens a connection per concurrent query up to the pool cap, so 18 parallel reads opened 9 physical connections, each paying TLS + auth + a full `pg_type` catalog load. Reading once removes the round trips *and* the connection pressure. Mentioning it in case the batching approach looks tempting later.

### DIFF
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -52,6 +52,7 @@ import {
   openrouterRequiresExplicitPromptCache,
 } from './recipes/openrouter.ts';
 import { resolveModel, TIER_DEFAULTS } from '../model-config.ts';
+import { snapshotConfigReader } from '../config-snapshot.ts';
 import { parseLlmJson } from '../llm-json.ts';
 import type { BrainEngine } from '../engine.ts';
 import { dimsProviderOptions } from './dims.ts';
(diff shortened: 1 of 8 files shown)


time-attack:
Re-trialed against the current integration branch (post v0.46.23-25 waves): Conflicts vs wave: src/core/ai/gateway.ts, src/core/model-config.ts. PR base v0.43.0.0; wave rewrote resolveModel->resolveModelDetailed + tier/discovery flow. NOT superseded (wave lacks bulk config read). No red flags. — a rebase along those lines puts this back in the active merge queue; re-trials run nightly, so a fresh push gets picked up automatically. Thank you!

garrytan:

PR #3987, opened 2026-08-11
fix(pages): purge-deleted dry-run shares the delete predicate — the listPages cap under-counts

### LINKED ISSUE #3594: Dry-run is unsafe in at least 3 commands: destructive call precedes the guard, or the default contradicts the docs

Masashi-Ono0611:
## Problem

`gbrain pages purge-deleted --dry-run` and the real purge use two different predicates, so the dry-run count is not a guarantee of what the destructive run will do.

- **Dry-run** (`src/commands/pages.ts`, pre-fix): enumerates `engine.listPages({ includeDeleted: true, limit: 10000 })` and filters in JS against `Date.now()`:

  ```ts
  const candidates = await engine.listPages({ includeDeleted: true, limit: 10000 });
  const cutoff = Date.now() - olderThanHours * 60 * 60 * 1000;
  const wouldPurge = candidates.filter(
    (p) => p.deleted_at && p.deleted_at instanceof Date && p.deleted_at.getTime() < cutoff,
  );
  ```

- **Real run** (`purgeDeletedPages` in both engines): an unbounded DELETE on the DB clock:

  ```sql
  DELETE FROM pages
  WHERE deleted_at IS NOT NULL
    AND deleted_at < now() - ($1 || ' hours')::interval
  RETURNING slug
  ```

The 10000-row cap is consumed by **live** pages too (`includeDeleted: true` returns everything, default sort `updated_desc`). On a brain past 10000 pages, any soft-deleted row that sorts beyond the first 10000 `updated_desc` results falls off the enumeration, and the dry-run under-reports — in the extreme, dry-run says 0 while the real run deletes N. The two paths also read different clocks (JS `Date.now()` vs DB `now()`).

Same broader dry-run honesty/safety class as previously merged fixes: #3594 ("Dry-run is unsafe in at least 3 commands"), #3955 (dry-run progress told the wrong scale), #3956 (dry-run flag unwired, actually mutated), #1575 (destructive default contradicting its doc).

## Fix

Route the dry-run through the engine method: `purgeDeletedPages(olderThanHours, { dryRun: true })` runs a SELECT with the **identical WHERE clause** (same cutoff arithmetic, same DB `now()` clock source) in both engines. The predicate text is kept adjacent to the DELETE in each engine rather than extracted to a helper, and the new parity tests pin that the two branches select the same rows. It returns `{ slugs, count, pages }`, where the optional `pages` field carries `slug + deleted_at` so the CLI keeps its `deleted_at=` display line. The destructive path is byte-for-byte unchanged; existing callers (`cycle.ts`, `jobs.ts`, `operations.ts`) are unaffected (optional second parameter, optional response field).

The CLI dry-run branch in `src/commands/pages.ts` now calls this instead of the capped enumeration.

## Tests

`test/pages-soft-delete.test.ts` (PGLite, always runs) — new describe:
- dry-run mutates nothing (row count unchanged before/after)
- dry-run slug set == the set a subsequent real run actually deletes; live and inside-window pages in neither
- dry-run `pages` carry `deleted_at` as a `Date` (CLI display contract)
- **red contrast**: reproduces the pre-fix predicate verbatim at small scale (3 recently-updated live pages + 1 old soft-deleted page, enumeration capped at 3) — the old filter misses the victim; the predicate-sharing dry-run and the real DELETE both catch it
- dry-run clamps negative hours (no crash, finite count)

`test/e2e/purge-deleted-dryrun-postgres.test.ts` (new, DATABASE_URL-gated) — same parity assertions on the live Postgres engine, per the engine-parity invariant.

Red-green verified: with `src/` reverted to master and the new tests kept, 4/5 fail (the reverted dry-run actually deletes rows); with the fix, 20/20 pass. `bun run typecheck` clean.

## Scope / open question

This PR aligns the dry-run with the destructive run and deliberately adds no new feature surface; the dry-run's own behavior does change (uncapped enumeration, DB clock, previously missed candidates now listed) — that change is the fix. One thing I noticed but left out: `purge-deleted` has no way to scope by source (the DELETE spans all sources). Whether that should exist feels like a product call, so I'm flagging it here rather than implementing it.

### DIFF
--- a/src/commands/pages.ts
+++ b/src/commands/pages.ts
@@ -30,19 +30,17 @@ async function runPurgeDeleted(engine: BrainEngine, args: string[]): Promise<voi
   const json = args.includes('--json');
 
   if (dryRun) {
-    // Use listPages with includeDeleted to enumerate the recoverable set, then
-    // count how many would be purged given the cutoff. Stays read-only.
-    const candidates = await engine.listPages({ includeDeleted: true, limit: 10000 });
-    const cutoff = Date.now() - olderThanHours * 60 * 60 * 1000;
-    const wouldPurge = candidates.filter(
-      (p) => p.deleted_at && p.deleted_at instanceof Date && p.deleted_at.getTime() < cutoff,
-    );
+    // Same engine method, same WHERE predicate, same DB now() clock as the
+    // real purge — only the verb differs (SELECT, stays read-only). The old
+    // listPages enumeration capped at 10000 rows (live pages included), so
+    // brains past the cap under-reported the purge set.
+    const preview = await engine.purgeDeletedPages(olderThanHours, { dryRun: true });
     if (json) {
-      console.log(JSON.stringify({ dry_run: true, older_than_hours: olderThanHours, count: wouldPurge.length, slugs: wouldPurge.map((p) => p.slug) }, null, 2));
+      console.log(JSON.stringify({ dry_run: true, older_than_hours: olderThanHours, count: preview.count, slugs: preview.slugs }, null, 2));
       return;
     }
-    console.log(`(dry-run) Would purge ${wouldPurge.length} page(s) soft-deleted more than ${olderThanHours}h ago.`);
-    for (const p of wouldPurge) console.log(`  ${p.slug}  deleted_at=${p.deleted_at?.toISOString()}`);
+    console.log(`(dry-run) Would purge ${preview.count} page(s) soft-deleted more than ${olderThanHours}h ago.`);
+    for (const p of preview.pages ?? []) console.log(`  ${p.slug}  deleted_at=${p.deleted_at.toISOString()}`);
     return;
   }
 
--- a/src/core/engine.ts
+++ b/src/core/engine.ts
@@ -819,8 +819,18 @@ export interface BrainEngine {
    * v0.26.5 — hard-delete pages whose `deleted_at` is older than the cutoff.
    * Called by the autopilot purge phase and by the `gbrain pages purge-deleted`
    * CLI escape hatch. Cascades through existing FKs.
+   *
+   * `opts.dryRun: true` runs a SELECT with the SAME WHERE predicate (same
+   * cutoff arithmetic, same DB `now()` clock source) instead of the DELETE,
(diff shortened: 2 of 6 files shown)


time-attack:

PR #3992, opened 2026-08-11
fix(oauth): require explicit DCR client approval

ShintaroKawakami:
## Summary
- Dynamic registrations stay pending until an operator atomically approves the complete client metadata and source grants.
- OAuth/MCP/ingest paths fail closed for pending or unscoped remote clients.
- v126 is manually gated; v127 preserves response-type metadata without replaying destructive work.

## Validation
- OAuth/DCR focused suites, migration gate isolation, and verify checks passed locally.
- HTTP E2E remains pending a DATABASE_URL-backed environment.

## Deployment note
This PR does not authorize production migration, credential containment, or DCR re-enable.


time-attack:
Thanks for this — the fail-closed direction for dynamic client registration is the right instinct, the test coverage is substantial, and you kept engine parity across all three schema surfaces, which most contributors miss. But we can't take this as one PR. It bundles three separate product decisions: (1) migration v126 resets every existing OAuth client to pending and revokes all live tokens, which breaks every deployment on upgrade; (2) the new manual-migration gate makes every gbrain command refuse to start after upgrade until `apply-migrations --yes`, including embedded-engine personal brains that never touch OAuth; (3) the new CHECK constraint changes what a valid approved client row is, altering the existing register-client contract. We decline bundles regardless of how sound the individual pieces are — each of those is a deliberate call we have to make on its own, not ride in on a hardening PR. Two mechanical blockers as well: the #1553 citations throughout the code point at an unrelated closed change, so they need to be fixed or dropped, and by your own description the HTTP E2E path is untested. Here's the cut that IS fix-lane worthy: new dynamic registrations land pending with zero scopes, plus the `approve-client` operator command — no reset of existing clients, no token revocation, no startup gate, no constraint change. Send that alone, small, with a test that fails on unpatched master, and it goes in the fix lane. Declining the branch, not the idea.

PR #3993, opened 2026-08-11
fix(embed): detect chunkless pages with non-empty content in embed --stale

Masashi-Ono0611:
### Symptom

`gbrain embed --stale` derives its stale set entirely from `content_chunks`
rows — NULL embeddings, plus page-signature drift in signature-aware mode
(via `countStaleChunks` / `listStaleChunks`). A page written directly
via `putPage` that never went through the chunking step has **zero**
`content_chunks` rows — there is no row to go stale, so the page is
invisible to `embed --stale` forever, no matter how many times it runs.

We hit this dogfooding a single-user install: an enrichment path calls
`engine.putPage` directly to write an entity-stub page (title + a short
body), bypassing the normal import/chunk pipeline. The stub page landed in
the brain, `get_page` could read it back, but it never appeared in
content-query results — `embed --stale` reported "0 stale found"
indefinitely because there was nothing in `content_chunks` for it to find.
(The page-grain title arm in `searchTitles` can still surface a chunkless
page via its synthetic `chunk_id 0` row when the query hits the title/
`search_vector` — see `postgres-engine.ts:1962-1970` — so this is a
content/semantic-retrieval gap specifically, not total invisibility.)

There is already a **manual** repair path for a related staleness class:
`gbrain doctor` warns on pages where `contextual_retrieval_mode IS NULL`,
and `gbrain reindex --markdown` sweeps that same predicate
(`reindex.ts:118,135`) and re-chunks each page with
`forceRechunk: true` — `importFromFile` for file-backed pages,
`importFromContent` for DB-only pages (`reindex.ts:215-220,242-246`) —
which would also catch a chunkless page, since it never got a
`contextual_retrieval_mode` stamp either. But that sweep is
operator-invoked: you have to run `gbrain doctor`, notice the warning, and
then remember to run `gbrain reindex --markdown`. `embed --stale` is the
loop that runs unattended (cron/routine "get everything embedded") with no
human in that path to notice anything. This PR adds the same detection as
an automatic safety net inside that unattended loop, rather than relying on
someone reaching `reindex --markdown` by way of `doctor`.

This PR is the detection-side safety net only. It does not touch the
stub-generation call site itself (that's a separate, narrower fix in the
enrichment path) — the goal here is that regardless of how a chunkless page
with real content ends up in the brain, `embed --stale` finds and repairs
it.

### What changed

- **New `BrainEngine` methods** (parity in `src/core/pglite-engine.ts` and
  `src/core/postgres-engine.ts`): `countChunklessPagesWithContent` /
  `listChunklessPagesWithContent`. Predicate: `deleted_at IS NULL AND
  (compiled_truth <> '' OR timeline <> '') AND NOT EXISTS (content_chunks
  row)`, using the canonical `EMBED_SKIP_FILTER_FRAGMENT` /
  `QUARANTINE_FILTER_FRAGMENT` to exclude pages carrying those frontmatter
  markers. Both are **excluded on purpose** — already
  intentionally-chunkless by design (the content-quality gate drops chunks
  for junk/oversized pages), not the drift this fix targets. Re-chunking
  them would just recreate the state the gate produced. (`timeline` is
  included, not just `compiled_truth`, because the healer chunks both
  independently, mirroring `embedPage`.)
- **`healChunklessPages` sweep wired into `embedAllStale`**
  (`src/commands/embed.ts` — wired inside `embedAllStale`, so it runs
  wherever a stale-mode embed runs: every `runEmbedCore({ stale: true })`
  caller, i.e. the CLI's `--stale`, sync's auto-embed, cycle embed, the
  embed-backfill job, migrations, and the ops path; `--all` keeps its
  embedding behavior — see Scope boundaries). Gated behind a pre-flight `countChunklessPagesWithContent`
  call so a healthy brain pays one extra `SELECT count(*)` and does no
  further work. When chunkless pages are found, they're chunked locally
  (mirroring the existing `embedPage()` chunkless branch — same `chunkText`
  calls, same `compiled_truth`/`timeline` split) with `embedding = NULL`.
  The new chunk rows flow into the *same* `--stale` pass via the existing
  NULL-embedding cursor — no separate embed step needed. `dryRun` chunks
  in-memory only (no writes), matching `embedPage`'s dry-run contract; the
  affected dry-run log lines report the cumulative `result.would_embed`
  instead of a bare stale count that would otherwise understate the total.
  Each page's heal is try/caught and isolated (`EmbedResult.failures` +
  `failure_samples`) so one broken chunkless page can't abort the whole
  `--stale` run before the normal NULL-embedding pass even starts. Batch
  size is 50 (not this file's usual 2000) because each row here carries a
  full page body. Shares ONE `GBRAIN_EMBED_TIME_BUDGET_MS` wall-clock
  window with the main stale loop (not two independent 30-minute ones),
  mirrors `--catch-up`, honors the caller's pacer, and an abort during
  healing stops the run before any further write path runs.
- Added `EmbedResult.chunkless_pages_healed` (additive field).
- `docs/architecture/KEY_FILES.md`'s `embed.ts` entry updated to describe
  this as current `--stale` behavior (repo convention: reference docs
  describe current state, no versioned append-log). `llms.txt`/
  `llms-full.txt` regenerated (byte-identical — `KEY_FILES.md` is
  link-only in that bundle).

### Known limitation (not fully closed — read before merging)

The write path narrows, but does **not eliminate**, a check-then-write
race: immediately before writing, the healer re-fetches the live page and
re-checks for chunks, but a concurrent writer landing chunks in the exact
gap between that check and the `upsertChunks` call can still have them
overwritten with this sweep's (stale) content snapshot — `upsertChunks`
treats its input as the full desired chunk set and deletes anything not
in it. This is the **same** unmitigated window `embedPage`'s existing
single-page chunkless branch already ships with (that branch doesn't even
have the re-check this PR adds) — not a new race class, and the blast
radius is bounded: the page isn't deleted or corrupted, just re-chunked
from stale content, self-healing on the next real write. Full atomicity
would need a new engine primitive (transactional/version-guarded
conditional write) shared by every `upsertChunks` caller, which is out of
scope for a chunkless-page safety net. See the `healChunklessPages`
docstring for the full writeup. Flagging this explicitly rather than
letting the code's re-check read as a complete fix.

### Engine parity

Both engines implement the identical `buildChunklessPagesWhere` predicate
and the same keyset-pagination shape as the existing
`listStalePagesForExtraction` sibling. Verified against a **real Postgres**
(`docker-compose.test.yml`, not just PGLite) via
`test/e2e/engine-parity.test.ts` — all 30 parity tests pass, including the
new `chunkless-page-with-content detection: Postgres ↔ PGLite parity` test.

### Tests

- `test/embed-stale-chunkless-pages.serial.test.ts` (new, 13 tests):
  engine-level detection (chunked/empty/quarantined/embed_skip exclusion,
  timeline-only pages); end-to-end `embed --stale` healing (chunked +
  embedded in one pass), no regression on pre-existing stale chunks,
  quarantine/embed_skip pages staying untouched, dry-run non-mutation, a
  healthy-brain guard (asserts no healing occurs when no chunkless pages
  exist; the single pre-flight `SELECT count(*)` is by design and not
  itself asserted), a concurrent-writer race
  simulation (asserts the concurrent write survives, not the sweep's stale
  snapshot), and per-page failure isolation (one broken chunkless page
  doesn't block an unrelated stale chunk from embedding).
- `test/embed.serial.test.ts` (+1 test, mock-engine based): a caller abort
  firing mid-heal stops the run before `invalidateStaleSignatureEmbeddings`.
- `test/e2e/engine-parity.test.ts` (+1 test, DATABASE_URL-gated): Postgres
  ↔ PGLite parity for the detection predicate.
- Confirmed red→green at each of the three review-fix rounds (reverted
  the relevant implementation file(s) to the pre-fix commit, confirmed the
  new test(s) failed; restored the fix, confirmed they passed).
- No regressions: `test/embed.serial.test.ts`, `test/embed-stale.serial.test.ts`,
  `test/embed-partial-failure-3037.serial.test.ts`,
  `test/embed-stale-chunkless-pages.serial.test.ts` — 66 tests, all pass.
  Also ran the real-PGLite embed-adjacent suites (`migrate-embeddings-boundary`,
  `migrate-embeddings-flow`, `jobs-embed-background-dryrun`, `v0_37_gap_fill`,
  `sync-deleted-slug-embed`) — one pre-existing 3-test failure in
  `migrate-embeddings-flow.serial.test.ts` reproduces identically on
  unmodified `upstream/master` (confirmed by re-running with this diff's
  implementation files reverted) when run in the same combined process; a
  pre-existing cross-file gateway-state isolation issue, unrelated to this
  change — the file passes cleanly in isolation both before and after.
- `bun run typecheck` clean throughout.
- Went through three rounds of `codex exec` review (`model_reasoning_effort=high`);
  each round's Critical/Warning findings were fixed and re-verified before
  the next round (concurrent-writer race narrowing + per-page failure
  isolation + timeline-only predicate + canonical embed_skip/quarantine
  fragments in round 1; per-page try/catch, live-refetch-before-write,
  smaller batch size, doc accuracy in round 2; shared wall-clock budget,
  abort propagation, and honest race-window documentation in round 3). The
  "Known limitation" section above is the one finding carried forward as
  an accepted, documented, bounded risk rather than fixed outright — see
  that section for why.

### Scope boundaries

- Does not touch the enrichment/stub-generation call site.
- Does not change existing `--stale` behavior for NULL-embedding chunks —
  only adds detection for the zero-chunk-row case.
- `--all`'s embedding behavior is unchanged (the heal sweep runs only on
  stale-mode runs — every `runEmbedCore({ stale: true })` caller); its
  `EmbedResult` does gain the additive
  `chunkless_pages_healed: 0` field, like every other embed result shape.


Masashi-Ono0611:
Borrowed this patch onto a live single-user install (Postgres brain, v0.44.0.0 + patch): ran the PR's test files against a clone of the patched live install — the 13 new `embed-stale-chunkless-pages.serial` tests and the full `embed.serial` suite all pass (49 pass total), including end-to-end healing, quarantine/embed_skip exclusion, dry-run non-mutation, the healthy-brain guard, the concurrent-writer race simulation, and per-page failure isolation. Scope note: the `engine-parity` e2e file can't execute in that sandboxed clone (no `DATABASE_URL` there by design), so cross-engine parity is covered by the real-Postgres run reported in the PR body (the repo's CI e2e lane does not include the engine-parity file), not by this live pass. Receipt retained locally.

time-attack:
Merged in the v0.45.10.0 community fix-wave 2 (#4085). Your fix landed on master as 8a626999f0, credited to you via Co-Authored-By, and is called out in the v0.45.10.0 CHANGELOG entry. Thank you for the contribution!

PR #3996, opened 2026-08-11
v0.44.0.1 fix(eval): resolve BrainBench defaults from package root

philip-rossoneri:
## Summary

- resolve the bundled BrainBench fixture, gold, and baseline defaults from the GBrain package root
- keep help output stable and portable instead of exposing machine-specific absolute paths
- resolve the receipt git SHA from the package root and suppress irrelevant git errors
- add a subprocess regression that runs the bare command from an unrelated working directory

## Reproduction

Before this change, a global Bun/GitHub install failed when `gbrain eval brainbench` was launched outside the package checkout because the default corpus paths were interpreted relative to the caller cwd. Explicit `--fixtures` and `--gold` paths worked.

## Verification

- `bun test test/eval-brainbench-e2e.test.ts` (22 pass)
- `bun run typecheck`
- `bun run verify` (34 checks pass)
- external-cwd bare CLI regression passes

### DIFF
--- a/src/commands/eval-brainbench.ts
+++ b/src/commands/eval-brainbench.ts
@@ -16,13 +16,14 @@
  */
 
 import { readFileSync, writeFileSync, existsSync, mkdirSync } from 'node:fs';
-import { dirname } from 'node:path';
+import { dirname, join } from 'node:path';
 import { execSync } from 'node:child_process';
 import { flushThenExit } from '../core/cli-force-exit.ts';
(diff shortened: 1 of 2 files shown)


time-attack:
Merged in the v0.45.8.0 community fix-wave (#4050). Your fix landed on master as 30c81b709c, credited to you via Co-Authored-By, and is called out in the v0.45.8.0 CHANGELOG entry. Thank you for the contribution!

PR #3998, opened 2026-08-11
fix(engine): NUL-sanitize pages body columns + chunk_text (both engines)

joshuaarcher:
## Problem

`putPage` binds `title` / `compiled_truth` / `timeline` raw, and `upsertChunks` binds `chunk_text` raw, in **both** `postgres-engine.ts` and `pglite-engine.ts`. A raw NUL byte (`0x00`) anywhere in a page body aborts the INSERT with:

```
ERROR: invalid byte sequence for encoding "UTF8": 0x00
```

Postgres `text` (and PGLite, which is Postgres under the hood) cannot store `0x00`, so a single bad byte silently drops the **entire** document — the page never persists. This is common in large ingests of real-world content (transcripts, scraped pages, exported chat logs); in one deployment it dropped ~240 documents before the cause was traced.

## Root cause

#2011 added NUL + lone-surrogate sanitization (`sanitizeForJsonb`) for the free-text of **links, timeline entries, and takes** — but the same treatment was never applied to the `pages` table's own body columns (`title`, `compiled_truth`, `timeline`) or to `content_chunks.chunk_text`. Those four fields still reach the driver raw, so the NUL/lone-surrogate class that #2011 fixed everywhere else still aborts a page write.

## Fix

Adds a small null-safe helper beside the existing sanitizers in `batch-rows.ts`:

```ts
export const sanitizeText = <T extends string | null | undefined>(s: T): T =>
  (typeof s === 'string' ? (sanitizeForJsonb(s) as T) : s);
```

and wraps the four body fields with it in both engines. Same free-prose-body policy as `sanitizeForJsonb` (strip NUL + well-form lone surrogates; never applied to identity/security fields). `null` / `undefined` pass through unchanged so nullable columns keep their semantics.

## Verification

- Current build **rejects** a document containing a `0x00` with the exact error above.
- Patched build **ingests** it (NUL stripped) and it persists — confirmed via `get_page` and a direct DB row check, on real Postgres.
- `tsc --noEmit` clean. 18 insertions across 3 files; no behavior change for content that was already valid.

### DIFF
--- a/src/core/batch-rows.ts
+++ b/src/core/batch-rows.ts
@@ -87,6 +87,18 @@ export const stripNul = (s: string): string => (s.includes('\0') ? s.replace(/\0
  */
 export const sanitizeForJsonb = (s: string): string => ensureWellFormed(stripNul(s));
 
+/**
+ * Null-safe free-prose sanitizer for plain TEXT / TEXT[] columns (title,
+ * compiled_truth, timeline, chunk_text). Same NUL + lone-surrogate cleanup as
+ * `sanitizeForJsonb` — a raw 0x00 in a page body otherwise aborts the INSERT
+ * with `invalid byte sequence for encoding "UTF8": 0x00` (Postgres text cannot
+ * store NUL), and a lone surrogate is likewise rejected. Passes null/undefined
+ * through unchanged so nullable columns keep their semantics. Same policy as
+ * sanitizeForJsonb: free-prose body only, NEVER identity/security fields.
+ */
+export const sanitizeText = <T extends string | null | undefined>(s: T): T =>
+  (typeof s === 'string' ? (sanitizeForJsonb(s) as T) : s);
+
 /** One links row, keys === the jsonb_to_recordset column list. */
 export interface LinkRow {
   from_slug: string;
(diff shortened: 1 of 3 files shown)


Masashi-Ono0611:
Ran into this exact failure mode independently — a local transcript-ingest pipeline had files permanently skipped on every run with the exact `invalid byte sequence for encoding "UTF8": 0x00` error, traced to a stray NUL byte in source content.

Traced it to the same root cause reading the code: `putPage`/`upsertChunks` bind `title`/`compiled_truth`/`timeline`/`chunk_text` raw in both engines, so anything that reaches those paths with a NUL byte hits this. Sanitizing at the `batch-rows.ts` layer (alongside the existing `sanitizeForJsonb` from #2011) looks like the right root-cause location — it catches every write path (markdown import, code import, sync, direct API writes), not just one importer.

No blockers found reviewing the diff itself — small, scoped, consistent with the existing sanitizer policy, and CI is green. Would be good to see this land.

time-attack:
Re-trialed against the current integration branch (post v0.46.23-25 waves): Merge impossible: unrelated histories (fork history rewritten; no merge-base with wave). Cherry-pick of lone commit 0ce1ebbc conflicts in src/core/{pglite,postgres}-engine.ts; batch-rows.ts clean. Not superseded; benign diff. — a rebase along those lines puts this back in the active merge queue; re-trials run nightly, so a fresh push gets picked up automatically. Thank you!

garrytan:

PR #4000, opened 2026-08-11
fix(mcp,search): honor per-call source_id on federated read ops

pavelpp-topia:
## Problem

`search`, `get_page`, `list_pages`, and `resolve_slugs` never declared a `source_id` parameter, so an agent had **no per-call way to narrow the source scope** — the argument was silently dropped by `validateParams` (`src/mcp/dispatch.ts`). Only `query` accepted it.

All four route through `federatedSearchScope(ctx)`, and when a transport populates `localFederatedSourceIds` (stdio MCP without `GBRAIN_SOURCE`; no-grant HTTP tokens — #2561/#3242), an unqualified read spans **every** federated source. With no per-request `source_id`, the only lever to pin one source was a server-wide `GBRAIN_SOURCE` env pin — static, not agent-steerable.

Since pages are owned per-source (`(source_id, slug)` uniqueness), the same slug exists across sources, so an inability to scope is both a relevance and a cross-source-read problem.

## Fix

Add a `source_id` param to `search`, `get_page`, `list_pages`, and `resolve_slugs` and thread it through `federatedSearchScope(ctx, sourceIdParam)`, mirroring the existing `query` op. Trust + grant enforcement come for free via `resolveRequestedScope`:

- an out-of-grant `source_id` is rejected for remote callers (`permission_denied`),
- `__all__` keeps its local (whole-brain) / remote (grant-limited) semantics,
- omitting `source_id` preserves the federated-span default (no behavior change for existing callers).

This brings all five federated read ops (`query` + these four) to parity.

## Tests

Extends `test/local-federated-search-scope.test.ts`: param declared on each op, each narrows to one source over a federated ctx, and remote callers cannot escape their grant via `source_id`. Full file 27/27 pass; `bun run typecheck` clean.

## Notes

No VERSION/CHANGELOG bump — per `docs/RELEASING.md` the maintainer's fix-wave `/ship` allocates the version and writes the CHANGELOG entry at ship time.


Masashi-Ono0611:
Black-box confirmation from a real brain (Postgres engine, v0.45.11.0), in case it is useful for this PR.

The decisive case is a **source id that does not exist**. Calling the op directly:

```
gbrain call search '{"query":"test","source_id":"apple-notes","limit":3}'      -> 3 results
gbrain call search '{"query":"test","source_id":"zzz-nonexistent","limit":3}'  -> the same 3 results
gbrain call search '{"query":"test","limit":3}'                                -> the same 3 results
```

All three return an identical result set, and the returned slugs come from a different source than the one requested. A nonexistent id producing the same results as no id at all rules out "the scope was applied and happened to match" — the parameter is dropped before it reaches the scope, exactly as this PR describes.

One thing worth noting for whoever reviews the surface question: **the CLI no longer shows this symptom.** `gbrain search --source-id X` is rejected up front by strict flag validation (`unknown flag --source-id for 'gbrain search'`, exit 1), because the flag validator derives its allowlist from `op.params` — the same declaration this PR adds. So today the silent behavior survives only on the paths that bypass that validator: the MCP tool and `gbrain call search`. That also means the fix here has a second effect that may not be obvious from the diff: declaring the parameter makes the CLI flag start working, rather than erroring.

No opinion offered on the implementation; the reason I looked at this at all is that `search` scoping matters for the source-isolation invariant, and I wanted to know which surfaces are affected before assuming my own results were scoped.

time-attack:

pavelpp-topia:
Rebased onto current master - the branch is mergeable again.

time-attack:
Re-trialed against the current integration branch (post v0.46.23-25 waves): Conflict: src/core/ops/pages.ts — get_page hunk superseded by #4329 (parseSourceIdParam source_id landed). search/list_pages/resolve_slugs source_id hunks still novel; rebase on wave pattern. No red flags. — a rebase along those lines puts this back in the active merge queue; re-trials run nightly, so a fresh push gets picked up automatically. Thank you!

garrytan:

PR #4007, opened 2026-08-11
fix(facts): recall since/order follows event time (COALESCE(valid_from, created_at))

Masashi-Ono0611:
## The bug

`extract-conversation-facts` run as a batch over many pages writes facts whose `created_at` collapses into per-transaction groups — each segment's insert transaction stamps all its facts with the same clock value, so a large run produces big blocks of identical timestamps. `recall --since 48h` ("what happened yesterday") then sorts and filters by *extraction time*, not by when the underlying events occurred, and surfaces facts from the wrong day.

Measured on a brain where one backfill produced 880 facts: `recall --since 48h` returned the top of a single earlier session, not the recent work.

## The fix

A `FactListOpts.eventTime` engine option, off by default at the engine layer. When set, `listFactsSince`'s `since` comparison and `ORDER BY` use `COALESCE(valid_from, created_at)` — event time — instead of creation time, on both engines. The `recall` op and the local CLI recall path opt in unconditionally — **that is the user-visible fix: `recall`'s `--since` filtering and ordering now follow event time** (facts without `valid_from` fall back to `created_at`, so they behave exactly as before).

The engine default stays creation-time, so the one other caller — `src/core/facts/meta-hook.ts`, the hot-memory injection cache (24h window re-ranked by decayed confidence) — deliberately keeps its existing semantics.

Migration v126 adds `idx_facts_since_event_time`, mirroring `idx_facts_since` (same leading column, same partial predicate). Without it the common recall shape — epoch cutoff, no entity, `ORDER BY … LIMIT n` — degrades from an index scan that stops at n rows into a full scan of the source plus a sort.

## Verification

- **Engine layer** (`test/facts-engine.test.ts`, plus a `DATABASE_URL`-gated Postgres mirror at `test/e2e/facts-recall-event-time-postgres.test.ts`): the flag unset keeps `created_at` semantics; set, it drops a backdated event out of a 48h window and flips `ORDER BY` to event time.
- **Op layer** (`test/facts-recall-event-time-op.test.ts`): drives the real op through `dispatchToolCall` with two facts carrying extraction-time `created_at` values and divergent `valid_from`. The engine tests alone stayed green when the op stopped passing the flag, so they could not catch the wiring regressing.
- **Discriminating**: removing `eventTime: true` from `src/core/operations.ts` turns both op-level tests red (0 pass / 2 fail); restoring it returns 19 pass / 0 fail across both files.
- `bun run typecheck` clean.

## Scope

The behavior change is deliberately scoped to the `recall` surfaces (op + local CLI); no new CLI flag, no config dimension, and the engine flag stays off for every other caller. Whether to expose a user-facing toggle back to creation-time ordering is left as a maintainer call.

### DIFF
--- a/scripts/module-size-limits.tsv
+++ b/scripts/module-size-limits.tsv
@@ -15,7 +15,7 @@ src/core/cycle.ts	3080	ratchet	grown v0.46.20.0 phase-boundary split (resolveCyc
 src/commands/serve-http.ts	3194	ratchet	grown v0.46.23.0 review fix wave: github-kind webhook gating + case-insensitive repo match
 src/commands/jobs.ts	3027	ratchet	grown v0.46.22.0 phone wave: github source sync job (#4104)
 src/core/search/hybrid.ts	2629	ratchet	grown v0.46.23.0 review fix wave: supersede scoping + edge-existence gate
-src/core/engine.ts	2351	ratchet	grown v0.46.23.0: putPage allowEmptyOverwrite contract (#4335)
+src/core/engine.ts	2364	ratchet	grown v0.46.23.0: putPage allowEmptyOverwrite contract (#4335) + eventTime JSDoc
 src/commands/autopilot.ts	2301	ratchet
 src/commands/extract.ts	2161	ratchet
 src/commands/extract-conversation-facts.ts	2045	ratchet	grown v0.46.x per-provider gateway-unavailable copy
--- a/src/commands/recall.ts
+++ b/src/commands/recall.ts
@@ -426,8 +426,17 @@ async function fetchRowsLocal(
     });
   }
   if (resolvedSince) {
+    // Post-review fix: `--since-last-run`/`--watch` resolve `resolvedSince`
+    // from a cursor written as the PRIOR RUN's wall-clock start time
+    // (`writeCursor(sourceId, tStart, ...)` in runRecallOnce — creation-time
+    // semantics). Comparing that cursor against event time
+    // (COALESCE(valid_from, created_at)) drops rows created after the cursor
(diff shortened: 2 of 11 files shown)


time-attack:

time-attack:
Trial-merge against current master conflicts after the v0.46.23.0 facts changes. Still wanted (opt-in eventTime recall ordering) — please rebase over master; its pair #4008 rebases on top of it.

Masashi-Ono0611:
Follow-up from a Codex self-review of this branch's rebase — not blocking, flagging for consideration before merge:

Migration v133 (`facts_event_time_index`) creates the new index with plain `CREATE INDEX IF NOT EXISTS`:

```sql
CREATE INDEX IF NOT EXISTS idx_facts_since_event_time
  ON facts (source_id, (COALESCE(valid_from, created_at)) DESC)
  WHERE expired_at IS NULL;
```

Other migrations that add indexes to potentially large existing tables in this codebase (e.g. `pages_dedup_partial_index` at v95) use the `CREATE INDEX CONCURRENTLY` + `transaction: false` + pre-drop-invalid-remnant pattern on Postgres, since a plain `CREATE INDEX` takes an exclusive lock that blocks writes for the duration of the build on a large `facts` table. This migration doesn't follow that pattern.

Given `facts` can grow large in an active brain, might be worth switching this one to the CONCURRENTLY pattern (with the PGLite branch staying a plain `CREATE INDEX`, as usual). Happy to send a follow-up patch if useful — didn't want to bundle an unrelated migration change into this PR without checking first.

Masashi-Ono0611:
Rebased over master at 280099d1 — mergeable, CLEAN. Ready for re-trial.

time-attack:

PR #4008, opened 2026-08-11
fix(facts,search): exclude extraction audit checkpoint rows from recall

Masashi-Ono0611:
## What

`extract-conversation-facts` writes durable audit checkpoint rows
(`EXTRACTION_COMPLETE` / `EXTRACTION_NOT_APPLICABLE`) into the `facts` table
to mark batch-run progress (`writeTerminalAuditRow` / `writeNonExtractableAuditRow`
in `src/commands/extract-conversation-facts.ts`). Their `created_at` is
the audit write time — within each page's run they land after that page's
facts — so right after a batch run they can occupy the newest-N fetch window
for trusted callers (no visibility filter), and `recall --since` (and the
no-filter arm) can return audit checkpoint rows instead of real user facts. With a small `--limit` and a multi-page batch run, the
audit rows alone can fill the window and the real facts never surface.

This PR adds `FactListOpts.excludeAuditRows?: boolean` and wires it into
every trusted `recall` read path (other internal fact readers such as
`facts/meta-hook.ts` do not pass the flag and keep their existing view):

- Both engines (`src/core/postgres-engine.ts`, `src/core/pglite-engine.ts`)
  honor it as a SQL `WHERE fact NOT IN ('EXTRACTION_COMPLETE',
  'EXTRACTION_NOT_APPLICABLE')` clause, consistently across all three
  `FactListOpts` consumers — `listFactsByEntity`, `listFactsBySession`,
  `listFactsSince` — not just the one call site that originally surfaced the
  bug.
- The `recall` MCP op (`src/core/operations.ts`) passes it on every arm
  (entity / session / since / no-filter), plus keeps a client-side filter as
  belt-and-braces defense in depth.
- The `gbrain recall` CLI's local (non-thin-client) path
  (`src/commands/recall.ts` `fetchRowsLocal`) also passes it — this path
  calls the engine directly and does **not** go through the `recall` op
  handler, so it needed its own fix (caught in review; see below).

Off by default, so any existing caller that intentionally wants the full row
set (e.g. an internal checkpoint audit) is unaffected.

## Scope boundary (deliberate)

This PR is narrow on purpose:

- **Only** `excludeAuditRows`. A separate, unrelated concern (in-SQL event-time
  ordering via a `FactListOpts.eventTime` field, and in-SQL grep via
  `FactListOpts.grepText`) is explicitly **not** included here — different
  problem, different PR.
- **Not** attempting to separate audit checkpoint rows into their own table.
  That's a bigger design question (schema surface, migration, backward
  compat for existing checkpoint state) that I'm leaving to maintainer
  judgment rather than bundling into a bug-fix PR. This PR only changes what
  `recall`'s read paths return — it does not change how or where the audit
  rows are written.

## Testing

- `bun run typecheck`: clean.
- Engine-level parity tests (PGLite always-on + Postgres gated by
  `DATABASE_URL`, mirrored files): `test/facts-separation-pglite.test.ts` /
  `test/e2e/facts-separation-postgres.test.ts`. Cover: basic exclude/include
  toggle, exact-match-not-substring (a real fact merely *containing* the
  audit literal as a substring survives), and `listFactsByEntity` /
  `listFactsBySession` honoring the flag.
- Op-level end-to-end test (new file, `test/facts-recall-audit-rows.test.ts`):
  exercises the `recall` MCP op via `dispatchToolCall` (since arm with a
  `limit` smaller than the seeded audit-row count — the actual crowd-out
  failure mode — plus the no-filter arm and a world-visible-audit-row
  defense-in-depth case for `remote=true`), **and** the `gbrain recall` CLI's
  local path via `runRecall` directly (isolated from the operator's real
  `~/.gbrain/config.json` via `GBRAIN_HOME` -> a fresh temp dir, so the test
  deterministically exercises `fetchRowsLocal` instead of possibly routing
  through `callRemoteTool` on a thin-client-configured machine).
- Discriminativeness is a reproducible property of the tests: flip the
  `recall` op's since and no-filter `excludeAuditRows: true` call sites in
  `src/core/operations.ts` to `false` and their two SQL-dependent op-level
  tests go red (one per call site); restore them and they pass.
- `bun run typecheck` clean; the full suite runs in CI (all checks green on
  this PR).
- Postgres e2e tests skip gracefully (no `DATABASE_URL` in this sandbox) —
  PGLite side is fully exercised; the Postgres SQL is the same shape as the
  existing PGLite-parity pattern in this file.

### DIFF
--- a/scripts/module-size-limits.tsv
+++ b/scripts/module-size-limits.tsv
@@ -15,7 +15,7 @@ src/core/cycle.ts	3080	ratchet	grown v0.46.20.0 phase-boundary split (resolveCyc
 src/commands/serve-http.ts	3194	ratchet	grown v0.46.23.0 review fix wave: github-kind webhook gating + case-insensitive repo match
 src/commands/jobs.ts	3027	ratchet	grown v0.46.22.0 phone wave: github source sync job (#4104)
 src/core/search/hybrid.ts	2629	ratchet	grown v0.46.23.0 review fix wave: supersede scoping + edge-existence gate
-src/core/engine.ts	2351	ratchet	grown v0.46.23.0: putPage allowEmptyOverwrite contract (#4335)
+src/core/engine.ts	2365	ratchet	grown v0.46.23.0: putPage allowEmptyOverwrite contract (#4335) + #4008 recall-exclude-audit-rows fix
 src/commands/autopilot.ts	2301	ratchet
 src/commands/extract.ts	2161	ratchet
 src/commands/extract-conversation-facts.ts	2045	ratchet	grown v0.46.x per-provider gateway-unavailable copy
--- a/src/commands/extract-conversation-facts.ts
+++ b/src/commands/extract-conversation-facts.ts
@@ -94,6 +94,7 @@ import { assertFactsEmbeddingDimMatchesConfig } from '../core/embedding-dim-chec
 import { writeReceipt, shortRunId } from '../core/extract/receipt-writer.ts';
 import { upsertExtractRollup } from '../core/extract/rollup-writer.ts';
 import { ALLOWED_TYPES, type AllowedType } from '../core/facts/conversation-types.ts';
+import { TERMINAL_AUDIT_SOURCE, NON_EXTRACTABLE_AUDIT_SOURCE } from '../core/facts/audit-sources.ts';
 
 // Re-exported verbatim so existing importers (this file's own helpers below
 // and this file's tests) keep working unchanged; doctor.ts, jobs.ts,
@@ -103,6 +104,12 @@ import { ALLOWED_TYPES, type AllowedType } from '../core/facts/conversation-type
 // own CLI flag surface.
 export { ALLOWED_TYPES };
 export type { AllowedType };
+
+// Re-exported for existing importers (test/extract-conversation-facts.test.ts,
+// test/doctor-conversation-facts-backlog.test.ts, src/eval/brainbench/metrics/write-back.ts).
+// The values themselves now live in ../core/facts/audit-sources.ts — see that
+// leaf module's docstring for why (engine-live static-import requirement).
+export { TERMINAL_AUDIT_SOURCE, NON_EXTRACTABLE_AUDIT_SOURCE };
 
 // ---------------------------------------------------------------------------
 // Tunables (exported for tests).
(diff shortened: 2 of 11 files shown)


time-attack:

Masashi-Ono0611:
Rebased onto current master (93 commits, zero conflicts) and pushed one substantive change on top.

**The filter now keys on the writing source rather than the fact text.** It was
`fact NOT IN ('EXTRACTION_COMPLETE', 'EXTRACTION_NOT_APPLICABLE')`; it is now an exclusion on the two
audit `source` values. Two reasons:

1. A genuine user fact whose text happens to be exactly `EXTRACTION_COMPLETE` was being hidden. The
   payload is user data; the writer is what actually identifies a marker.
2. A third audit marker added later would silently slip past a text-based filter. Keying on the
   writer covers it automatically.

The two source constants moved to a leaf module `src/core/facts/audit-sources.ts` so both engines can
import them with a static top-level import — `pglite-engine.ts` and `postgres-engine.ts` are
engine-live paths and `scripts/check-engine-dynamic-import.sh` forbids runtime dynamic `import()`
there. `extract-conversation-facts.ts` imports from the leaf and re-exports both names unchanged, so
its existing importers are untouched (verified: the two test files and
`src/eval/brainbench/metrics/write-back.ts` all still pass without edits).

**Opt-in semantics are unchanged.** `excludeAuditRows` stays opt-in with the same explicit call sites;
callers that do not ask see no behavior change. Only the predicate moved.

Checked rather than assumed:
- `source TEXT NOT NULL` in the facts DDL, so the exclusion cannot be defeated by SQL NULL comparison.
- `listSupersessions` on BOTH engines takes a narrower opts type with no `excludeAuditRows` field and
  never received the filter, before or after. That is deliberate — it is the supersession audit log
  and should keep showing every superseded row, including markers.

The new regression test is the point of the change: a fact with text `EXTRACTION_COMPLETE` but a
normal source must still be returned. Against the old text-based predicate the suite goes 7 fail / 2
pass; with the source-based predicate, 9 pass / 0 fail. Also `bun run typecheck` and `bun run verify`
(43/43) exit 0, and shard 5 is 1572 pass / 0 fail.

Context for the record: I briefly had a second PR open for this same defect (#4134) — my duplicate,
now closed. This change is the one improvement it carried.

time-attack:
Trial-merge conflicts against current master alongside its pair #4007 (same 5 files, v0.46.23.0 facts changes). Please rebase after #4007 — both remain queue candidates.

Masashi-Ono0611:
Rebased over master at a7ceb547 — mergeable. The only red job is the pre-existing test/e2e/sync-delegation-under-serve.serial.test.ts Pin 4 flake (also failing on unmodified master), not caused by this change. Ready for re-trial.

time-attack:

PR #4010, opened 2026-08-11
fix(pglite): batch code edge inserts below bind limit

kyle944:
## Summary

Batch PGLite code-edge inserts below its signed-16-bit parameter limit.

A generated TypeScript declaration produced 5,826 unresolved edges in one statement (34,956 bind parameters). Above 32,767 parameters PGLite reported success but poisoned the session, causing later imports to silently fail.

This caps batches at 30,000 parameters: 5,000 unresolved edges or 4,285 resolved edges.

## Validation

- `bun test test/code-edges.test.ts` (9 passing), including a 5,462-edge regression followed by a successful page write
- `bun test test/import-code-edges-source-id.test.ts` (2 passing)
- `bun run typecheck`
- `git diff --check`

### DIFF
--- a/src/core/pglite-engine.ts
+++ b/src/core/pglite-engine.ts
@@ -95,6 +95,12 @@ import { hasCJK, escapeLikePattern } from './cjk.ts';
 
 type PGLiteDB = PGlite;
 
+// PGLite's parameter bridge corrupts the current engine session when a single
+// statement crosses the signed-int16 bind ceiling (32,767 parameters). Keep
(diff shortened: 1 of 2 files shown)


time-attack:
Merged in the v0.45.8.0 community fix-wave (#4050). Your fix landed on master as dc6e61b07f, credited to you via Co-Authored-By, and is called out in the v0.45.8.0 CHANGELOG entry. Thank you for the contribution!

PR #4011, opened 2026-08-11
fix(code-intel): route unqualified graph queries to the code-bearing source on federated brains

kyle944:
## Problem

On a brain with two federated sources — a notes/vault source and a code source — the four
graph ops (`code_callers`, `code_callees`, `code_blast`, `code_flow`) return
`status: "not_built"` for symbols that are demonstrably indexed, while `code_def` and
`code_refs` answer correctly for the same symbol.

Reproduced on a live PGLite brain: `code_def` finds `fires_in_window` at
`routines/weekly-rollup/expected_runs.py:67`, and the call graph holds both of its caller
edges, but `code_callers` over MCP reports `not_built` with zero callers. The CLI answers
fine, which makes it look like a stale index rather than a routing bug.

Two independent causes, one per commit.

### 1. Graph ops collapse to a source that holds no code

`code_def` / `code_refs` are brain-wide. The graph four are single-source by design (the
engine API and the traversal cache key take one `sourceId`), so an unqualified call
collapses to the caller's scalar scope — the seed `default` source, which on a vault+code
brain contains no code chunks. The readiness probe then honestly reports `not_built` for
that scope, and the agent concludes the graph was never built.

`routeCodeIntelScope` re-routes instead of widening (widening isn't available to a
single-source traversal): when the collapsed source has no code chunks and exactly one
source in the caller's federated read set does, traverse that one. This is the
multi-source cousin of the CLI's existing `sole_non_default` tier (#1434).

Fail-closed, in precedence order — explicit `source_id` or `all_sources` wins; an already
brain-wide scope is untouched; no federated read set (a granted token) never widens; a
collapsed source that *does* have code stands; 0 or 2+ candidate sources is ambiguous and
keeps the original scope so readiness reports honestly; a probe error keeps the original
scope.

### 2. `serve --http` never threaded the federated read set

Even with the above, the fix couldn't fire under `serve --http`. The legacy HTTP transport
(`src/mcp/http-transport.ts`) and stdio (`src/mcp/server.ts`) both thread
`localFederatedSourceIds` into dispatch for no-grant callers per #3242; the SDK-transport
dispatch site in `serve-http.ts` never did. The same legacy bearer token therefore saw
federated pages on two transports and scalar `default` on the third — a pre-existing
visibility split that reaches well beyond code intel.

`AuthInfo` gains `hasSourceGrant`, set only by the legacy verify path in `oauth-provider.ts`
(mirroring the field the legacy transport already computes). OAuth clients leave it
undefined and never widen, so the widening condition matches the other two transports
exactly.

## Testing

- 6 new unit tests pin the re-route matrix: re-routes to the sole code-bearing source;
  does not re-route when the collapsed source has code; explicit `source_id` always wins;
  ambiguous (2+ code sources) keeps the original scope; no federated set never widens;
  probe failure fails closed.
- 300 tests green across the scope-resolver, code-intel, code-graph-readiness, transport
  and OAuth suites; `bun run typecheck` clean.
- Verified end to end against a live two-source PGLite brain over MCP: the same unqualified
  `code_callers` call that returned `not_built` now returns `ready` with both real caller
  edges.

## Note

Branched off `master` and independent of #4010 (PGLite edge batching) — no overlapping
files.

### DIFF
--- a/src/commands/serve-http.ts
+++ b/src/commands/serve-http.ts
@@ -2056,12 +2056,29 @@ export async function runServeHttp(engine: BrainEngine, options: ServeHttpOption
       // verifyAccessToken. The env-fallback is gone.
       const tokenSourceId = authInfo.sourceId ?? 'default';
 
+      // #3242 parity: the legacy-transport and stdio dispatch sites widen a
+      // no-grant caller's unqualified reads across the federated source set
+      // (localFederatedSourceIds); this SDK-transport site never did, so the
+      // same token saw federated pages over /mcp on one serve mode and scalar
+      // 'default' on the other. hasSourceGrant === false is set ONLY for
+      // legacy bearer tokens with no operator source grant (oauth-provider);
+      // granted tokens and OAuth clients never widen. Best-effort: a resolver
+      // failure keeps the scalar scope.
+      let localFederated: string[] | undefined;
+      if (authInfo.hasSourceGrant === false && tokenSourceId) {
+        try {
+          const { localFederatedSourceIds } = await import('../core/source-resolver.ts');
+          localFederated = await localFederatedSourceIds(engine, tokenSourceId, 'seed_default');
+        } catch { /* scalar scope stands */ }
+      }
+
       let toolResult: Awaited<ReturnType<typeof dispatchToolCall>>;
       try {
         toolResult = await dispatchToolCall(engine, name, params as Record<string, unknown> | undefined, {
           remote: true,
           takesHoldersAllowList: tokenAllowList,
           sourceId: tokenSourceId,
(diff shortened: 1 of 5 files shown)


time-attack:

time-attack:
Re-trialed against the current integration branch (post v0.46.23-25 waves): Conflicts: src/core/operations.ts (now peeled to ops/*), src/commands/serve-http.ts. serve-http #3242 half superseded (#4132 adoption, v0.46.14.0); routeCodeIntelScope graph re-route half unlanded — rework into ops/code-intel.ts. — a rebase along those lines puts this back in the active merge queue; re-trials run nightly, so a fresh push gets picked up automatically. Thank you!

garrytan:

PR #4012, opened 2026-08-11
feat(takes): human-review proposal queue + fence parser fix for proposal kinds (data-loss guard)

erionjuniordeandrade-a11y:
## What

- **Fence parser reads proposal kinds.** `parseTakesFence` only knew the seed kinds (`fact|take|bet|hunch`), but the proposal pipeline emits `prediction|judgment|bet`. Accepting a proposal wrote a fence row the parser could not read.
- **Fence rewrites refuse to drop rows.** The unknown-kind warning was swallowed (`void warnings`), and the next rewrite of the same fence re-rendered from the parsed subset — silently erasing the previously accepted take. Observed as real data loss on a live brain: accept proposal A on a page, accept proposal B on the same page, A's row is gone from markdown and the derived DB row is clobbered. `upsertTakeRow`/`supersedeRow` now throw instead of rewriting a fence that contains rows the parser dropped.
- **Take-proposal review surface**: source-scoped `takes proposals` listing with evidence status (`current|stale|missing`) + bounded excerpt, explicit `takes accept|reject <id> --by <reviewer>` (idempotent; repeat decisions are no-ops), current-page content-hash check before promotion, and a read-only `takes_proposals_list` op for MCP with remote evidence redaction.
- **`takes_proposal_create` op**: lets an external reasoning layer (e.g. a Claude operator replacing the internal LLM extractor) enter the review queue. Content hash is stamped server-side from the page's current indexed content, creation is idempotent per (page, content, prompt_version, claim), and nothing is promoted without the manual review above.
- **Source-scoped `extract_facts` enumeration**: a source-scoped run no longer enumerates every slug in the brain.

## Tests

- Regression tests for the parser vocabulary, the no-clobber guard on both rewrite paths, proposal creation/dedup/kind validation, and the review CLI.
- Ran on top of current master (52306ed4): `bun run typecheck` clean; takes/fence/proposal/extract suites 135 pass / 0 fail.

## Notes for the wave

The no-clobber guard changes `upsertTakeRow`/`supersedeRow` from silent-warning to throw when the existing fence has unparseable rows — any caller that relied on rewriting past a malformed fence will now surface the malformation instead of erasing it.

### DIFF
--- a/docs/takes-vs-facts.md
+++ b/docs/takes-vs-facts.md
@@ -21,6 +21,32 @@ The epistemological layer. WHO believes WHAT, with confidence weight and time.
 
 **Query surface:** `gbrain takes list`, `gbrain takes search`, `gbrain think`
 
+### Proposal review is a separate step
+
+The `propose_takes` cycle writes candidates to `take_proposals`; it does not
+promote them into canonical takes. Review the queue explicitly:
+
+```bash
+gbrain takes proposals --json
+gbrain takes accept <proposal-id> --by <reviewer>
+gbrain takes reject <proposal-id> --by <reviewer>
+```
+
+The queue is source-scoped, includes the proposing model/run and current
+evidence status, and refuses promotion when the source page has changed or
+has no registered local repository. Acceptance writes the Markdown take row
+and the derived DB row together under the page lock; repeating the same
+decision is idempotent. `takes.bootstrap_enabled` remains opt-in and is not
+required for manual proposal review.
+
+External reasoning layers (e.g. a Claude operator) enter the queue through
+the `takes_proposal_create` op: the content hash is stamped server-side from
+the page's current indexed content, creation is idempotent per (page,
+content, prompt_version, claim), and nothing is promoted without the manual
+review above. Proposal kinds are `prediction | judgment | bet`; the fence
+parser reads these alongside the seed kinds, and fence rewrites refuse to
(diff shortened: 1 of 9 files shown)


time-attack:
Thanks — this is careful work, and the core finding is real. We verified against master: the fence parser's kind vocabulary doesn't include the kinds the proposal cycle emits, and the rewrite paths silently re-render from the parsed subset, which is a genuine silent-data-loss bug. That portion — the parser vocabulary fix, the no-clobber throw on both rewrite paths, and the source-scoped extract_facts enumeration, plus their regression tests — is exactly the class of bug we take. Send those alone and they go in the fix lane.

The branch as a whole is a close, though, for two reasons that have nothing to do with the quality of the work. First, it bundles that fix with a substantial new feature (a review-queue CLI, a new read op, and a new remote write op), and we don't merge bundles regardless of how sound the individual pieces are — one concern per PR is the bar. Second, new operation surface — especially a new remote write path — is a default no here; it's a product decision we make deliberately, not something we take by PR, however well it's built.

For what it's worth, the review-queue design points the right direction (proposals stay unpromoted until an explicit human decision), and the queue you're reading is currently write-only, so the gap you identified is real. If the maintainer opts into that surface, a resubmission scoped to it alone would get a real review. Meanwhile: the ~160-line fix cut is invited now.

PR #4017, opened 2026-08-11
feat(doctor): policy-aware checks — pack-gated atoms backlog and opt-in takes report ok instead of warn

erionjuniordeandrade-a11y:
## What

Makes two doctor checks policy-aware so intentionally-disabled features stop dragging the health score, following the existing `checkHiddenBySearchPolicy` pattern (policy-derived state reports `ok`, not `warn`). Both derive the classification from EXISTING signals — no new config keys, no suppression list.

- **`extract_atoms_backlog`**: when the active pack does not declare the `extract_atoms` phase, the backlog is a configuration choice, not a fault — the check now reports `ok` with the count and the manual `--drain` command in the message. Behavior when the pack DOES declare the phase is unchanged.
- **`takes_count`**: the check already read `takes.bootstrap_enabled` but only changed the message. `0 takes` with bootstrap OFF (the A12 two-gate consent resting state) now reports `ok`; `0 takes` with bootstrap ON still warns with the remediation step.

Motivation: on an install that deliberately runs without the internal LLM phases, doctor's score punishes the policy itself — we observed the score *drop* after the system objectively improved, because every policy-off feature costs 5 points forever. Score should reflect actionable risk.

## Open design question (deliberately NOT included)

`cycle_freshness` has the same problem on cycle-less installs, but there is no clean existing signal: the pack `phases:` list only gates lens phases (`manifest-v1.ts`), the 17 core phases always run, and the default `gbrain-base` pack declares no `phases:` at all — so gating freshness on "empty phases" would silence real staleness for default installs. Left for a maintainer call (e.g. an explicit `cycle.enabled` config, or gating on cycle-eligibility of the sources).

## Tests

- New: undeclared-phase backlog (large and small) → `ok`; `checkTakesCount` bootstrap-off → `ok`, bootstrap-on → `warn`.
- On top of current master: `bun run typecheck` clean; `doctor-extract-atoms-backlog`, `onboard-full-flow`, `doctor-cycle-freshness`, `doctor-categories` → 43 pass / 0 fail; pack-gating/onboard regression suites → 39 pass / 0 fail.

### DIFF
--- a/src/commands/doctor.ts
+++ b/src/commands/doctor.ts
@@ -3895,12 +3895,13 @@ export async function checkDbOnlyCollectorCollision(
 /**
  * issue #1678 — extract_atoms_backlog doctor check.
  *
- * Closes the "silent backlog" gap: extract_atoms is pack-gated, so on a brain
- * whose active pack doesn't declare the phase it NEVER runs in the routine
- * cycle and pages accumulate forever with zero signal (the cycle reports a
- * clean `skipped`). This check counts the eligible-but-unextracted pages and,
- * when the pack doesn't run the phase AND the backlog is real, WARNs with the
- * exact `--drain` command.
+ * Counts the eligible-but-unextracted pages and reports whether the active
+ * pack runs extract_atoms at all. When the pack doesn't declare the phase,
+ * that is a policy choice (intentionally off) — the check reports `ok` with
+ * the count + the manual `--drain` command instead of warning, so an
+ * intentionally-disabled feature stops dragging the health score. When the
+ * pack DOES declare the phase, the routine cycle drains the backlog in
+ * bounded batches — informational `ok` either way.
  *
  * PAGE-BACKLOG-ONLY (Codex #11): extract_atoms also discovers transcript files
  * at runtime; this counts DB pages only — labeled in details. No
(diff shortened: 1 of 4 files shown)


time-attack:
Thanks for this — the problem statement is real (a deliberate policy choice shouldn't drag the health score down) and the tests are pinned coherently on both halves. Closing as-is, though, because as a unit this is a default-behavior policy change to doctor's reporting rather than a bug fix, and we hold a default-close bar on those regardless of implementation quality.

The extract_atoms_backlog half is the blocker: that warn exists because of a real incident where a backlog grew silently for weeks precisely when the pack didn't declare the phase, and the ask that created the check was 'never let I-did-nothing report as a clean success with no backlog visibility.' Two things break the 'undeclared = intentional' framing: the default pack declares no phases at all, so on stock installs undeclared is the shipped default rather than a choice anyone made; and the phase-declaration lookup fail-opens to false on registry errors, so a broken pack registry would report doctor ok. If we want policy-awareness here it needs an explicit operator signal (e.g. a config key acknowledging the phase is off), not intent inferred from absence.

The checkTakesCount half is a different story: bootstrap is an explicit opt-in, so 0 takes with it off really is the designed resting state, and warning on it is a misleading diagnostic. Send that piece alone with its test and it goes in the fix lane. Closing the branch, not the idea.

PR #4019, opened 2026-08-11
feat(ai): add ChatGPT OAuth provider for GPT-5.6

ravehorn:
## Summary
- add a ChatGPT-subscription Codex OAuth chat recipe for GPT-5.6 Luna, Terra, and Sol, with Luna first and max reasoning
- keep GBrain as the sole tool dispatcher while disabling Codex shell, web, MCP, plugins, skills, memory, telemetry, and other host capabilities
- preserve separate NVIDIA expansion and embedding routes and price subscription chat at zero incremental API-token cost

## Security
- require a dedicated owner-only OAuth home and pinned Codex CLI 0.147.0
- refuse API-key accounts, inherited credentials, model fallback, custom providers, redirect URLs, remote thread config, prompt telemetry, and unexpected protocol requests
- serialize the rotating OAuth store with abort-safe cross-process locking and clean ephemeral state on every exit path

## Validation
- 126 focused tests passed with 850 expectations after rebasing onto v0.44.1.0
- 34/34 repository verification checks passed
- TypeScript check and diff check passed
- independent security review found no remaining P0-P3 issue

Production login, configuration, and service activation are intentionally outside this PR.


ravehorn:
VPS staging evidence (no production cutover):

- candidate: GBrain `0.44.1.0`, official base `fc310db3ea152e821ddb245a800956f15b8ed4fb` + this PR head `be7bb2f35cb57887b8627a3fbbeae83be6e7a630`
- `bun run typecheck`: pass
- focused provider/lock/NVIDIA tests: 46/46 pass (199 expectations)
- production remains on 0.42.62; no database migration or service restart performed

The live ChatGPT device authorization and real Luna/Terra/Sol calls remain separate runtime gates.

time-attack:
Thanks — this is genuinely one of the most carefully engineered provider PRs the repo has seen: the hermetic app-server stub suite, the env-allowlist and effective-config isolation checks, and the abort-safe lock reuse are all real craft, and none of it exists on master.

Declining on lane, not quality. This is the same shape as #3587 (Codex-CLI OAuth brokering), and the call there holds: routing a consumer-subscription OAuth session through a pinned third-party CLI subprocess is a vendor-relationship and terms-of-service posture gbrain doesn't carry, and the exact-version pin (with its per-version disabled-features audit) is a maintenance treadmill that breaks on every upstream release. External provider integrations of this class live in downstream agents, not core — the craft is fine; it's the mission fit that fails.

Two pieces of this branch would be welcome on their own in the fix lane:
1. The cross-process lock abort-safety change, with its 'aborting a waiter never reaps the live holder' test — send that alone and it goes in.
2. The claude-cli tool_use/tool_result rendering change — but as its own PR, with the tool-use instruction text updated to describe the new format and a claude-cli-side test pinning it. Bundled here it silently changes an existing provider's prompts, which we don't take in a ride-along (#3543).

Appreciate the staging receipts and the test discipline — if you re-cut either piece above, reference this PR and it'll get a fast look.

PR #4020, opened 2026-08-11
chore(cli): regenerate flag registry — --full drift is failing #2185 on master

DarkNightForge:
`bun run build:flag-registry` on a clean `master` already emits `--full` for eight commands that the committed registry lacks, so the `#2185` freshness guard is **failing on master right now**:

```
book-mirror  calibration  capture  enrich  forget  onboard  protocol  recall
```

Identical single-flag delta across eight unrelated commands — the flag (or the generator's detection of it) landed without a regeneration.

The failing test's own message is the fix:

```
test/cli-flag-validation.test.ts:224
// Any listed command means: run `bun run build:flag-registry` and commit.
expect(stale).toEqual([]);
```

So that's all this is. The diff is only `'--full'` inserted alphabetically into those eight entries — no hand edits.

### Verification

`bun test test/cli-flag-validation.test.ts` → **24 pass / 0 fail** (was 23/1 on master).

### Why it's separate

Split out of a larger branch so it can land on its own — it's mechanical, it un-reds `master`, and it's unrelated to the bug fixes in the sibling PRs. Worth merging first, since any other PR's CI will show this same failure until it does.

PR #4021, opened 2026-08-11
fix(test-runner): OOM rescue fires on text a test printed; ANSI breaks every shard count

DarkNightForge:
Three coupled defects in `scripts/run-unit-parallel.sh`. The third one launched a full 372-file serial re-run on a machine with 19 GB free, triggered by the test that exists to verify OOM rescue.

## 1. ANSI codes break every count

`bun_summary_count` matches `$1 ~ /^[0-9]+$/`, but the log line as bun writes it to a pipe is:

```
ESC[0mESC[32m 5385 passESC[0m
```

so `$1` is the escape blob, never the integer. Run the parser against a real shard log:

```
$ awk '$1 ~ /^[0-9]+$/ && $2 == "pass" { t += $1 } END { print t+0 }' .context/test-shards/shard-1.log
0
```

That shard had **5385 passes**. Every `shard N/M: pass=0 fail=0 skip=0` line this script has printed is this bug — including on runs where thousands of tests passed.

## 2. The failure marker is stale

Bun 1.3 emits `✗ `. The script matches `/^\(fail\) /` only, which appears **nowhere** in a current shard log. So `failing_files_in_log`, the failure-block extraction, and the rescue lane's real-vs-phantom classification all match nothing — and the rescue note renders as the malformed `oom_rescue_failed=0real`, which is exactly what `test/scripts/run-unit-parallel.test.ts` has been failing on:

```
Expected to contain: "oom-rescue (serial, confirmed real)"
Received:            "... | oom_rescue_failed=0real"
```

The meta-test wasn't flaky. It was correctly reporting this.

## 3. OOM detection is fooled by text a test printed

Detection was a bare `grep -qE "$OOM_RE"` over the whole shard log. The signature is ordinary text, so any failing test that *prints* it trips the detector — and one does: the rescue lane's own meta-test writes fixtures containing `Original error: Out of memory` and asserts on the nested runner's output. Once (2) made that assertion fail, bun printed the captured payload as a `Received:` diff, the fixture text landed in the **parent** shard log, and the parent concluded a real OOM had occurred.

(2) is the root cause of (3): a passing meta-test prints no diff, so nothing pollutes the log. (3) is hardened anyway, because the class is real.

## Fixes

- `strip_ansi` helper applied to every log parse. ESC is built with `printf` rather than written `\x1b`, since BSD sed (macOS) doesn't understand `\x`.
- Both marker vocabularies accepted — `✗ `/`✓ ` and `(fail) `/`(pass) ` — so this works across bun versions rather than pinning to one.
- OOM detection moved into `oom_signature_in_log`, which excludes assertion payloads (`Expected:` / `Received:` lines). A real OOM aborts the file and the signature appears as runtime output on its own line; it never appears **only** inside a quoted diff.

## Verification — both directions

The safety net still has to catch real OOMs, so that was tested explicitly, not assumed.

| Check | Result |
| --- | --- |
| Fixed parser vs. real shard logs | `pass=5385 fail=1`, `pass=5205 fail=1 skip=2`, `pass=5102 fail=0` — matches hand extraction; old parser said `0/0/0` |
| `oom_signature_in_log` on the polluted log | no OOM — false positive gone |
| `oom_signature_in_log` on a synthetic `error: Out of memory` log | **detected** — net still armed |
| `test/scripts/run-unit-parallel.test.ts` | **18 pass / 0 fail** (was 17/1), including *"a deterministic failure carrying the OOM signature re-fails serially and stays red"* |
| all of `test/scripts/` | **181 pass / 0 fail** |
| `bash -n`, `--dry-run` smoke | clean |

That meta-test passing is the important one: real failures are still not absolved by the rescue lane.

## Note

CI here will also show `cli-flag-validation` red until #4020 lands — that's pre-existing `master` drift, unrelated to this change.

### DIFF
--- a/scripts/run-unit-parallel.sh
+++ b/scripts/run-unit-parallel.sh
@@ -284,6 +284,23 @@ grep_count() {
   echo "${n:-0}"
 }
 
+# strip_ansi: emit FILE with SGR color sequences removed.
+#
+# Every parser below reads a log that `bun test` wrote to a pipe with color
+# still enabled, so its lines carry SGR escapes:
+#     ESC[0mESC[32m 5385 passESC[0m
+#     ESC[0mESC[31m✗ESC[0m ESC[0msome test name
+# Parsing those raw makes awk's $1 the escape blob rather than the count or the
+# marker — which silently produced `pass=0 fail=0` on shards that had thousands
+# of passes, and stopped every `✗`/`(fail)` pattern below from ever matching.
+# The ESC is built with printf rather than written as `\x1b`, because BSD sed
+# (macOS) does not understand `\x` escapes.
+strip_ansi() {
+  local esc
+  esc=$(printf '\033')
+  sed "s/${esc}\[[0-9;]*[a-zA-Z]//g" "$1"
+}
+
 # bun_summary_count: parses Bun's summary lines (one per `bun test` invocation
 # inside a shard — there's only one when we pass an explicit file list).
 # Looks for ` N pass` / ` N fail` / ` N skip` patterns and sums them across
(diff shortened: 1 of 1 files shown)


time-attack:

DarkNightForge:
Rebased onto current master (`dd99e40c`) — clean, no conflicts. Shell-only change; `bash -n` clean.

Workflow runs are queued as `action_required` (fork PR), so CI still needs a maintainer approval to actually run.

time-attack:

PR #4022, opened 2026-08-11
fix(files,capture,delete): commands report success while storing nothing, corrupting, or deleting originals

DarkNightForge:
Four commands reported success while doing nothing — or, in one case, while destroying data. They share one shape: **storage/disk-dependent work that proceeds as if its precondition held.**

Found while answering a much smaller question ("can I attach a PDF?"), then widened by auditing for siblings.

## 1. `files upload` — phantom attachments

```ts
if (config?.storage) { ...upload... }    // no backend → silently skipped
await sql`INSERT INTO files ...`         // row written regardless
console.log(`Uploaded: ${storagePath}`)  // reports success
```

`files` has **no blob column**, so with no backend there is nowhere for bytes to go. Observed: `Uploaded: inbox/test/img.png (3 KB)`, a metadata row written, a filesystem-wide `find` for the payload returning nothing, and `files signed-url` on that exact path answering `No storage backend configured.`

## 2. `files sync` — never uploaded at all

Inserted a row per file and reported `N uploaded` **with no storage call anywhere in the function**. Every row it produced was a phantom, even on a correctly configured brain.

## 3. `files verify` — false-passes, hiding 1 and 2

```ts
if (!row.content_hash || !row.storage_path) mismatches++;  // tautological for any INSERT
else verified++;
console.log(`${verified} files verified, 0 mismatches, 0 missing`);  // both counts literal
```

`missing` was declared and never incremented. The one command whose job is catching un-stored files reported a phantom row as **verified**.

## 4. `files redirect` — data loss

The destructive path: it unlinks local originals and leaves `supabase://` pointers. It built `storage` only `if (config?.storage)` and made the "verify remote exists before deleting locals" check conditional on that — so with **no backend the guard was skipped entirely** and the loop deleted originals while writing pointers to bytes that had never been uploaded.

## 5. `capture --file <pdf>` — silent corruption

The CV10 comment documented the hole it left open — *"Known limit: a PNG-without-NUL-in-first-8KB slips through... magic-byte allowlist closes this hole"* — and that allowlist was never built. An ASCII-armored PDF contains **zero NUL bytes**, so it passed the scan, was UTF-8-decoded into U+FFFD mojibake, and stored as the page body with `title: '%PDF-1.4'`. The document's real text, living in FlateDecode streams, was **gone**, while the command reported `created_or_updated`.

## 6. `delete_page` — deleted pages resurrect

DB-only, so the write-through `.md` outlived the page. On any brain committed on a timer (a snapshot cron, `sources harden`'s post-commit push) the orphan gets committed *after* deletion and the next `gbrain sync` restores the page. `restore_page` was correspondingly asymmetric.

---

## Fixes

- **One shared `requireStorageBackend()`** gates `upload` / `sync` / `mirror` / `redirect`, failing *before* any DB write or unlink. `sync` now actually uploads.
- **`verify` really verifies**: `storage.exists()` per row, real counts, nonzero exit. Rows with no backend report **UNVERIFIABLE** rather than "verified".
- **`detectBinarySignature()`**: the 29-signature magic-byte allowlist CV10 specified, running before the NUL scan (both kept). Weak ASCII magics (`MZ`, `BM`, `ID3`) are deliberately **omitted** — real text can begin with those, and those formats carry NULs early anyway, so including them would trade a missed binary for a refused valid capture.
- **`deletePageThrough()`**, sharing a factored `resolveWriteThroughTarget()` with the write path so the two planes cannot disagree about which file backs a page. `delete_page` and `restore_page` both report `write_through`.

**Guards.** Two source scans fail the build if anyone reintroduces `if (config?.storage)` or a hardcoded verify count. Both were confirmed to *fail* against a deliberately reintroduced regression, rather than passing vacuously.

## Breaking-change note

`upload` / `sync` / `redirect` now **refuse** where they previously exited 0, and `verify` exits nonzero when rows exist with no backend. That's the point — the old behavior minted phantom rows and, in `redirect`, deleted originals — but anyone scripting against a backend-less install will see a hard failure instead of a silent no-op, and a green CI may correctly turn red.

## Verification

Unit: 75 pass across the three touched test files, 97 across delete/restore-adjacent suites, `tsc --noEmit` clean.

End-to-end against a live Postgres brain:

| Check | Result |
| --- | --- |
| `capture --file doc.pdf` | refuses, exit 1, no page created, suggests `pdftotext` |
| `pdftotext … \| capture --stdin` | text preserved — unique token present 3×, was **0×** |
| `files upload` | refuses; **0 phantom rows** |
| `files sync` | refuses; no rows |
| `files redirect` (mirrored fixture) | refuses; **original intact, no pointer written** |
| `delete` | `write_through: {removed: true}`; file gone |
| `restore` | `write_through: {written: true}`; file back, content intact |

Does not touch retrieval (search ranking, RRF, embeddings, intent classification, query expansion, source boost, or the `query`/`search` handlers), so per CONTRIBUTING the `gbrain eval replay` gate doesn't apply.

## Note

CI here will also show `cli-flag-validation` red until #4020 lands — pre-existing `master` drift, unrelated to this change.

### DIFF
--- a/src/commands/capture.ts
+++ b/src/commands/capture.ts
@@ -172,9 +172,12 @@ function defaultSlug(content: string, now: Date = new Date(), type?: string): st
  * Returns the 0-indexed byte offset of the first NUL, or -1 if clean.
  * Caller decides the error shape (message vs JSON envelope).
  *
- * Known limit: a PNG-without-NUL-in-first-8KB slips through. v0.39
- * magic-byte allowlist (per CV10-B + TODOS.md) closes this hole. The
- * 8KB ceiling bounds the scan cost to ~microseconds even on huge files.
+ * The 8KB ceiling bounds the scan cost to ~microseconds even on huge files.
+ *
+ * This scan alone is NOT sufficient — a container with no NUL in its head (an
+ * ASCII-armored PDF being the common case) passes it. That hole is closed by
+ * `detectBinarySignature`, which runs first; keep both, since the NUL scan
+ * still catches binary formats that carry no signature in this list.
  */
 export function detectBinaryNullByte(buf: Buffer): number {
   const limit = Math.min(buf.length, 8 * 1024);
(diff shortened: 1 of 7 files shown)


time-attack:

DarkNightForge:
Rebased onto current master (`dd99e40c`). Two conflicts, both because master's twin-file fix landed in the same region this PR refactors. Flagging the resolution since it's more than a mechanical replay.

**`src/core/operations.ts`** — import union (`WriteThroughResult` from master + `deletePageThrough` from here).

**`src/core/write-through.ts`** — this PR extracts target resolution into a shared `resolveWriteThroughTarget()` so the write and delete planes can't disagree about which file backs a page. Master meanwhile added the `source_path` / `source_uri` preference to the *inline* version. Taking either side of the conflict would have lost one of the two fixes, so master's logic is **ported into the extracted helper** instead. Master's five `[REGRESSION twin]` tests pass unmodified against the refactor.

**One deliberate change at the seam.** Master's lookup filtered `AND deleted_at IS NULL`. That's correct inline, but `deletePageThrough` runs *after* `softDeletePage` has stamped the row — so in the shared helper that predicate returns no row on the delete plane, falls back to the slug-derived twin, and reports a clean `file_not_present` while the real artifact stays on disk for the next timer-based commit to resurrect. That is precisely the silent-no-op class this PR exists to close, reached through the delete plane instead of the write plane.

Dropped the predicate — `CONSTRAINT pages_source_slug_key UNIQUE (source_id, slug)` guarantees at most one row per slug, so there's no live/deleted ambiguity to resolve. Added a regression test to pin it:

    [REGRESSION twin] resolves the recorded source_path even AFTER the row is soft-deleted

It fails with the predicate restored and passes without, so it's a real guard rather than a restatement of the implementation.

Local: `bun run typecheck` clean; `write-through.test.ts` + `files.test.ts` 69 pass / 0 fail. CI runs are queued as `action_required` pending maintainer approval.

DarkNightForge:
@time-attack — this one plus #4021 and #4023 are rebased onto `dd99e40c` and in a mergeable state whenever fix-wave 3 is being assembled.

Deliberately not asking for a CI approval: per `docs/RELEASING.md` ("Community PR wave process") external PRs land by cherry-pick into a collector branch rather than by merging the fork branch — which is how #4018 landed in wave 2. If that's still the flow then the gated runs don't matter and the diffs can be read directly.

Recording the verification the gated CI would otherwise provide, since all four workflows on each PR are sitting at `action_required` (and per the same doc, fork PRs wouldn't receive base-repo secrets even once approved):

- **#4021** — shell-only change, `bash -n` clean
- **#4023** — 149 pass / 0 fail across the 9 affected files under the exact ambient env from the report; 10 fail with the branch's own `GBRAIN_TEST_KEEP_AMBIENT_ENV=1` escape hatch, so the scrub is doing the work
- **#4022** (this one) — `bun run typecheck` clean; `write-through.test.ts` + `files.test.ts` 69 pass / 0 fail

The one that genuinely wants a reviewer's eye is this PR. Master's twin-file fix landed inside the exact region it refactors, so the rebase ports master's logic into the extracted `resolveWriteThroughTarget()` rather than taking either side of the conflict, and it drops a `deleted_at IS NULL` predicate that becomes a silent no-op once the helper is shared with the delete plane. Full reasoning in the comment above. Happy to split that predicate change into its own PR if it's easier to review separately from the extraction.

No urgency — just making sure they're not sitting conflicted when the next wave goes out.

time-attack:
Re-trialed against the current integration branch (post v0.46.23-25 waves): CONFLICT (4 files): capture.ts, operations.ts, write-through.ts, test/write-through.test.ts. Not superseded — core fixes absent on wave; #4258 resolvePageWriteTarget overlaps its resolver. Rebase needed. No red flags. — a rebase along those lines puts this back in the active merge queue; re-trials run nightly, so a fresh push gets picked up automatically. Thank you!

garrytan:

PR #4023, opened 2026-08-11
fix(test): unit suite isn't hermetic against the operator's env — 12 spurious failures

DarkNightForge:
`scripts/run-e2e.sh` already scrubs `CONDUCTOR_* / MCP_* / OPENCLAW_* / GBRAIN_*` before bun starts, with a comment naming the failure mode exactly:

> A dev shell or a Conductor workspace exports … `GBRAIN_*` config overrides (e.g. a stray `GBRAIN_BRAIN_ID`, `GBRAIN_SOURCE`, `GBRAIN_*_THRESHOLD`) that would silently change test behavior — making "hermetic" E2E non-hermetic and its failures unreproducible across machines.

The unit lane never got that treatment. Measured on one operator machine, the ambient environment alone produced **12 failures with no defect in the product**:

| Ambient value | Failures |
| --- | --- |
| `GBRAIN_SOURCE=default` | 8 in `source-resolver-with-tier` + `source-resolver-silent-fallback`, 1 in `extract-fs-source-id` — resolution short-circuits at tier `env` before reaching the tier under test (`Expected: "seed_default" / Received: "env"`) |
| `GBRAIN_CYCLE_FRESHNESS_WARN_HOURS=10` | 1 in `doctor-cycle-freshness` — it moves the very threshold the test asserts (11 pass / 0 fail once unset) |
| `ANTHROPIC_` / `OPENAI_` / `OPENROUTER_API_KEY` | 2 in `memory-verbs-conformance`, 2 in `mcp-eval-capture`, 2 in `facts-classify` — tests pinning provider-*unavailable* branches take the live-provider path |

A suite that is red for environmental reasons is worse than a slow one: it trains you to ignore failures.

## Why per-test cleanup doesn't fix the API-key group

`test/helpers/legacy-embedding-preload.ts` snapshots `env: { ...process.env }` into the gateway, and `gateway.isAvailable()` ends at:

```ts
return required.every(k => !!_config!.env[k]);   // the snapshot — not process.env
```

So a key present when the snapshot is taken makes the gateway "available" for the rest of the process, and a test that later does `delete process.env.OPENAI_API_KEY` does **not** undo it. `mcp-eval-capture` deletes the key in its own setup and still failed.

That's why the new preload must load **first** in `bunfig.toml` — strip before the snapshot, or don't bother. Load order is load-bearing and commented as such.

## Why a preload rather than a shell scrub

`run-e2e.sh`'s approach only protects `bun run test`. A preload also protects `bun test test/foo.test.ts`, which is how anyone iterates on a single file. The cost is that this file has to own the E2E gate, since `run-e2e.sh` also invokes `bun test` and inherits the preload list.

**Happy to move it into the runner scripts instead** if you'd rather keep the harness shell-side — it's a small rewrite and this is the most opinionated piece here.

## The E2E gate

Provider keys are left **alone** when argv names a `test/e2e/` path (that runner passes one file at a time). Deliberately fail-safe in that direction: a misdetection keeps keys — E2E keeps working, at worst a unit file stays non-hermetic — rather than pulling credentials out from under a live-provider suite. The operator-context scrub is ungated, since E2E already does it and it's idempotent there.

Verified directly with doctored argv:

| argv | provider keys | `GBRAIN_SOURCE` |
| --- | --- | --- |
| `test/facts-classify.test.ts` | stripped | stripped |
| `test/e2e/cjk-roundtrip.test.ts` | **kept** | stripped |

## Also in here

`facts-classify` asserted the classifier-unavailable branch via a *comment* — `// Without API key in test env, isAvailable('chat') is false`. It now pins that precondition with `configureGateway({ env: {} })`. Runtime there went **1401ms → 141ms**: those "pure threshold math" unit tests had been making real billable provider calls on any machine with keys configured.

## Scope choices

- `_API_KEY$` pattern rather than a hardcoded list of the 21 provider vars, so new recipes are covered without editing this file.
- **Not** stripped: provider `*_BASE_URL` and `AZURE_OPENAI_USE_ENTRA`. Same shape, but no observed failure, and someone may point tests at a local Ollama on purpose. Left pending evidence rather than widened on speculation.
- Hatches: `GBRAIN_TEST_KEEP_AMBIENT_ENV=1`, `GBRAIN_TEST_KEEP_PROVIDER_KEYS=1`, `GBRAIN_DEBUG_PRELOAD=1` (logs names, never values).

## Verification

With all ambient vars still exported: **54 pass** across the four previously-failing files, **51 pass** across `source-resolver-*` / `extract-fs-source-id` / `cli-flag-validation`, `tsc --noEmit` clean.

## Note

CI here will also show `cli-flag-validation` red until #4020 lands — pre-existing `master` drift, unrelated to this change.

### DIFF
--- a/bunfig.toml
+++ b/bunfig.toml
@@ -23,4 +23,18 @@ timeout = 60_000
 # were appending rows into the operator's real ~/.gbrain/sync-failures.jsonl,
 # which `gbrain doctor` reads and warns on. See
 # test/helpers/sync-failures-preload.ts.
-preload = ["./test/helpers/legacy-embedding-preload.ts", "./test/helpers/audit-dir-preload.ts", "./test/helpers/sync-failures-preload.ts"]
+#
+# hermetic-env-preload.ts: drop the operator's ambient CONDUCTOR_/MCP_/OPENCLAW_/
+# GBRAIN_ overrides and provider *_API_KEYs, mirroring the scrub run-e2e.sh
+# already performs for the E2E lane. Measured on one operator machine, the
+# ambient environment alone produced 12 failures with no defect in the product
+# (a stray GBRAIN_SOURCE, a GBRAIN_CYCLE_FRESHNESS_WARN_HOURS, and three
+# exported provider keys). Provider keys are left alone when argv names a
+# test/e2e/ path, since that lane needs live providers.
+#
+# ORDER IS LOAD-BEARING: it must precede legacy-embedding-preload, which
+# snapshots `env: {...process.env}` into the gateway. isAvailable() reads that
+# snapshot, not process.env — so keys must be gone BEFORE the snapshot is taken.
+# (This is why a test's own `delete process.env.OPENAI_API_KEY` does not make the
+# gateway unavailable.)
+preload = ["./test/helpers/hermetic-env-preload.ts", "./test/helpers/legacy-embedding-preload.ts", "./test/helpers/audit-dir-preload.ts", "./test/helpers/sync-failures-preload.ts"]
--- a/test/facts-classify.test.ts
+++ b/test/facts-classify.test.ts
@@ -9,12 +9,48 @@
  *   - 4-strategy parse fallback for malformed JSON
  */
 
-import { describe, test, expect } from 'bun:test';
+import { describe, test, expect, beforeAll, afterAll } from 'bun:test';
 import {
   cosineSimilarity,
   classifyAgainstCandidates,
 } from '../src/core/facts/classify.ts';
+import { configureGateway, resetGateway } from '../src/core/ai/gateway.ts';
 import type { FactRow } from '../src/core/engine.ts';
+
+/**
+ * Two tests below pin the NO-CHAT-PROVIDER branch (classifier unavailable →
+ * cosine fallback). They used to just assume it, with a comment reading
+ * "Without API key in test env, isAvailable('chat') is false".
+ *
+ * That assumption is false on any machine where the operator exports provider
+ * keys: `test/helpers/legacy-embedding-preload.ts` builds the gateway config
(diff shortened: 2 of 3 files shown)


time-attack:

DarkNightForge:
Rebased onto current master (`dd99e40c`).

One conflict, in `bunfig.toml`: master added `sync-failures-preload.ts` to the preload list while this branch was adding `hermetic-env-preload.ts`. Resolved as a **union**, with the hermetic preload FIRST — its position is load-bearing (it must precede `legacy-embedding-preload.ts`, which snapshots `env: {...process.env}` into the gateway). The two scratch-dir preloads run after the scrub, so they mint fresh dirs rather than inheriting an ambient one; that's the same relationship this PR already had with `audit-dir-preload.ts`, so `sync-failures-preload.ts` slots in beside it.

Verified against the exact ambient env from the PR description
(`GBRAIN_SOURCE`, `GBRAIN_CYCLE_FRESHNESS_WARN_HOURS`, provider keys):

- with the fix: **149 pass / 0 fail** across the 9 affected files, `sync-failures.test.ts` and `gbrain-home-isolation.test.ts` included
- with `GBRAIN_TEST_KEEP_AMBIENT_ENV=1` (the escape hatch): **10 fail** — so the scrub is doing the work, and the hatch works

CI runs are queued as `action_required` pending maintainer approval.

time-attack:
Re-trialed against the current integration branch (post v0.46.23-25 waves): Conflict: bunfig.toml preload list (branch grew to 7 preloads). Key-strip half superseded by landed provider-keys-preload (same hatch names); GBRAIN_SOURCE/CONDUCTOR/MCP context scrub still unlanded — rebase, don't drop. — a rebase along those lines puts this back in the active merge queue; re-trials run nightly, so a fresh push gets picked up automatically. Thank you!

garrytan:

Issue #923, opened 2026-05-12
gbrain frontmatter validate walks node_modules/, inflating MISSING_OPEN counts in doctor


donogeme:
## Summary

`gbrain frontmatter validate <path>` (and by extension the `frontmatter_integrity` check in `gbrain doctor`) walks every subdirectory under `<path>`, including `node_modules/`, `.git/`, and other non-content directories. This produces large false-positive counts of `MISSING_OPEN` warnings on any source repo that vendors npm packages.

The sibling `gbrain frontmatter generate` walker filters these directories correctly. The two walkers should behave consistently.

## Repro

\`\`\`bash
mkdir test-source && cd test-source
npm init -y && npm i express
gbrain sources add test --path "\$(pwd)"
gbrain doctor 2>&1 | grep frontmatter_integrity
\`\`\`

Expected: 0 issues (no content yet).
Actual: tens to hundreds of \`MISSING_OPEN\` issues, all inside \`node_modules/\`.

I hit this on a real-world setup pointing gbrain at a TypeScript project (Bun + node_modules + a docs/ folder). 7 real content issues got drowned in 226 false positives from npm package READMEs.

## Root cause

Two walkers in [\`src/commands/frontmatter.ts\`](https://github.com/garrytan/gbrain/blob/master/src/commands/frontmatter.ts):

- **\`generate\` walker** (line ~421) correctly skips:
  \`\`\`ts
  if (entry === '.git' || entry === 'node_modules' || entry === '.obsidian') continue;
  \`\`\`

- **\`validate\` walker** (lines ~243–264) has no such filter. It pushes every directory onto its DFS stack and only filters per-file via \`isSyncable\`, which doesn't reject the \`node_modules\` prefix.

Result: doctor's \`frontmatter_integrity\` check (which uses the validate path) reports inflated counts that operators cannot resolve via \`--fix\` since the files aren't theirs. The warning becomes load-bearing noise that erodes trust in doctor's output.

## Suggested fix

Add the same skip predicate to the validate walker:

\`\`\`ts
for (const name of entries) {
  if (name === '.git' || name === 'node_modules' || name === '.obsidian') continue;
  // ...existing per-entry logic
}
\`\`\`

Both walkers should ideally share a single helper (\`shouldSkipDirEntry(name)\`) to prevent this from drifting again.

## Environment

- gbrain v0.33.0
- Windows 11, Bun 1.3.13
- Postgres engine (Supabase pooler)
[...]

Issue #202, opened 2026-04-18
extract: --source fs walker does not respect isSyncable prefix exclusions


jamebobob:
Bumped into this last night setting up a quarantine namespace (`_pending/`) for ambient captures from a signal-detector-style plugin. Extended `isSyncable()` in `src/core/sync.ts` to exclude the prefix and sync behaved. Then I ran `gbrain extract all --source fs` to check the graph layer and got "154 pages walked" on a brain that only has 61 authoritative pages. Spent a while convinced we had a quarantine leak before noticing extract uses a different walker.

`src/commands/extract.ts::walkMarkdownFiles()` skips leading-dot directories and leading-underscore *files*, but not leading-underscore *directories*. So `_pending/originals/foo.md` gets walked even when sync correctly skips it.

Nothing is getting corrupted. Extract writes don't land as real links because auto-link only creates edges between endpoints that exist in the DB, and quarantined content isn't indexed. But the page count is misleading and extract is doing work on content I asked it not to touch.

One-line fix, matching the file-level behavior at dir level:

```ts
for (const entry of readdirSync(d)) {
  if (entry.startsWith('.')) continue;
  if (entry.startsWith('_')) continue;  // new
  // ...
}
```

Or, less fragile long-term: lift a shared `shouldWalk(path)` helper out of `isSyncable` and have both walkers call it, so they can't drift apart again.

Related:
- #160 (quarantine gate proposal): user-defined quarantine only works if both sync and extract respect it.
- #132 (sync deadlock on large modification sets): adjacent "walks more than it should" class.
- #184 (normalize extract slugs and dir parsing): touches the same file, worth coordinating.

time-attack:
Closing as fixed on master (see 4446e9f9). Please reopen if still reproducible on current HEAD.

PR #4027, opened 2026-08-12
sync/import: --include-hidden waives the dot-prefix prune for named globs

dndungu:
## Problem

`pruneDir`'s leading-dot heuristic (skip `.git`, `.obsidian`, and any other dot-prefixed directory) is unconditional and currently unreachable by any CLI flag, `--include`/`--exclude` glob, or per-source config. Both classifiers that decide what's syncable check it *before* any include/exclude glob is even evaluated:

- `classifySync` (core/sync.ts) — drives `sync`'s incremental path, full-sync dry-run, and full-sync reconcile-deletes.
- `isCollectibleForWalker` (commands/import.ts) — drives `import`'s git-ls-files fast path, which is what every git-tracked source actually uses (the FS-walk fallback is only reached for non-git directories).

Concretely: a project that keeps schema'd, git-tracked content under a dot-prefixed directory by convention (e.g. [kazi-org/dira](https://github.com/kazi-org/dira)'s `.dira/entries/`) is silently unsyncable and unimportable on every git-tracked gbrain source, with no workaround. `gbrain doctor` and `gbrain query` give no signal that anything was skipped — the pages are just absent.

Worth noting: the two classifiers above independently reimplement the same segment-level prune check. That's the exact shape of bug that let `node_modules` drift out of sync before `PRUNE_DIR_NAMES` existed (#923, #202) — a fix to one and not the other silently prunes differently.

## Fix

- New exported `isPathPruned(path, includeHidden?)` in `core/sync.ts` — the single function both `classifySync` and `isCollectibleForWalker` now route through, so they can't drift from each other again.
- New `--include-hidden <glob>` CLI flag on `sync` (repeatable, same glob dialect as `--exclude`). When a path is pruned *only* because a segment is dot-prefixed, and the full path matches one of these globs, the dot-prune is waived. `PRUNE_DIR_NAMES` (`node_modules`/`vendor`/`dist`/`build`/`venv`) and `*.raw` are **never** waivable — the flag reaches past the leading-dot heuristic specifically, not past generated/vendored-tree exclusions.
- Threaded through every collection call site that needed it, including the full-sync reconcile-deletes pass — missing it there would have made freshly-included pages look "gone" on the very next sync and tripped the mass-delete valve.
- Cannot combine with `sync --all` (same restriction as `--exclude` and `--src-subpath` — these scope a single invocation).

## Known scope gap (documented in the code)

This does not reach `pruneDir`'s other, live-filesystem-walk callers — `import`'s fallback walker for non-git directories, and `storage.ts`'s `walkBrainRepo`. Those decide per-directory during real recursive descent and would need prefix-of-a-glob reasoning to honor the same waiver safely (a directory named `.dira` has to stay open for descent to reach `.dira/entries/*.md` even though `.dira` alone doesn't match that glob). Every `gbrain sources add`'d git repository — virtually all of them — goes through the git-listing fast path this PR covers, so the gap is scoped to importing a bare, non-git directory.

## Tests

`test/sync-include-hidden.test.ts`, new:
- `include-hidden-waives-dot-prune-when-glob-matches` — fails against pre-fix `master` (`isPathPruned` doesn't exist there; `unsyncableReason`/`isSyncable` ignore the option and always prune).
- `default-still-prunes-dot-dirs-without-include-hidden` — regression guard: `.git/notes.md` stays pruned when no flag is given, unaffected by an unrelated `--include-hidden` pattern.
- A case proving `PRUNE_DIR_NAMES`/`*.raw` are never waivable even with a matching glob.
- A case proving the waiver is scoped to the matching path, not every dot-dir.
- Two integration tests against a real temp git repo via the exported `collectSyncableFiles` (import's actual entry point): default excludes the hidden-dir page while collecting the ordinary one; `--include-hidden` includes both.

`bun run typecheck` clean. Full regression pass across the existing `sync`/`import` test files: 466/468 (the 2 stragglers are PGLite `beforeEach` hook timeouts from running 44 test files in a single process — both pass cleanly re-run in isolation, unrelated to this change).

## Not done here

CLI help text updated for `sync --include-hidden`; `import`'s own top-level CLI doesn't currently expose `--exclude` either, so I didn't add a new top-level flag there — the underlying `collectSyncableFiles`/`CollectOpts` support is there if/when that's wired up. `estimateSourceTreeTokens`'s cost-preview walker doesn't thread `includeHidden` (a preview accuracy nit, not a correctness issue). Happy to extend either if useful.


time-attack:

time-attack:
Re-trialed against the current integration branch (post v0.46.23-25 waves): Conflicts vs wave 2cdf825e: src/commands/import.ts(5), src/commands/sync.ts(1), cli-flag-registry.generated.ts(3, regenerable). Wave's onExcluded param at same call sites; composable rebase. Not superseded; no red flags. — a rebase along those lines puts this back in the active merge queue; re-trials run nightly, so a fresh push gets picked up automatically. Thank you!

garrytan:

PR #4035, opened 2026-08-12
fix(context-engine): reflex ignores turn delivered via `prompt` (empty `messages`)

OrcunJoyoLabs:
## Symptom

The Retrieval Reflex in the OpenClaw context engine never fires on live gateway turns. Salient entities in the current user turn that resolve to existing brain pages are silently not injected. The reflex only works in test harnesses that populate `messages`.

## Root cause

`assemble()` in `src/core/context-engine.ts` declares a `prompt?: string` param in its interface but only destructures `{ messages, tokenBudget, availableTools, citationsMode }`. The reflex reads the current turn's text exclusively from `messages` via `getLastUserText` / `getWindowTurns` / `getPriorContextText`.

Some OpenClaw runtimes — notably the codex-app-server in 2026.7.x — deliver the current user turn via `prompt` with an **empty** `messages` array. On those runtimes the reflex has nothing to read, so it never resolves any entity.

## Fix

When `messages` is empty, synthesize a single user turn from `prompt` and feed that (`effectiveMessages`) to the reflex resolution. The deterministic live-context block and the pass-through return path are untouched — they still return the original `messages`.

## Backward compatibility

The synthesis only activates when `messages` is empty. Whenever `messages` is non-empty (every existing runtime and all current tests), `effectiveMessages === messages`, so this is a no-op on those paths.

## Tests

Adds two regression tests to `test/retrieval-reflex.test.ts` (addressing the "zero tests" review blocker):

- **`prompt`-only turn (empty `messages`) still fires the reflex** — the codex-app-server delivery shape. Proven meaningful with a before/after run: it **fails** on the pre-fix `assemble()` (which reads only `messages`) and **passes** after the fix.
- **`prompt` is ignored when `messages` is non-empty** — guards the no-double-count / backward-compat claim by asserting an entity named only in `prompt` never reaches the resolver when `messages` is populated.

Full file: `24 pass, 0 fail`.

## Verification

Confirmed on a live gbrain v0.44.1.0 / OpenClaw 2026.7.1 instance: the reflex began firing on real gateway turns only after this change. The bug is still present on current `master` (v0.45.2.0) — `assemble()` there still destructures without `prompt` and passes `messages` straight into the reflex.

### DIFF
--- a/src/core/context-engine.ts
+++ b/src/core/context-engine.ts
@@ -661,9 +661,19 @@ export function createGBrainContextEngine(ctx: {
       return { ingested: true };
     },
 
-    async assemble({ messages, tokenBudget, availableTools, citationsMode }) {
+    async assemble({ messages, tokenBudget, availableTools, citationsMode, prompt }) {
       // Lazy SDK load on first method call (was top-level await pre-L0-B).
       await ensureSdkLoaded();
+
+      // Some OpenClaw runtimes (e.g. the codex-app-server in 2026.7.x) deliver
+      // the current user turn via `prompt` with an empty `messages` array.
+      // Synthesize a single user turn so the Retrieval Reflex still sees the
+      // text (the deterministic live-context/pass-through path is unaffected).
+      const effectiveMessages = (Array.isArray(messages) && messages.length > 0)
+        ? messages
+        : (typeof prompt === 'string' && prompt.trim()
+            ? ([{ role: 'user', content: prompt }] as typeof messages)
+            : messages);
 
       // 1. Generate deterministic context (<5ms, zero LLM calls)
       const liveCtx = generateLiveContext(workspaceDir);
@@ -681,11 +691,11 @@ export function createGBrainContextEngine(ctx: {
       // nothing salient resolves. Detect + point, never auto-dump bodies.
       const reflexAddition = await buildReflexAddition({
         workspaceDir,
-        currentUserText: getLastUserText(messages),
-        priorContextText: getPriorContextText(messages),
+        currentUserText: getLastUserText(effectiveMessages),
+        priorContextText: getPriorContextText(effectiveMessages),
         // v0.43 (#2095): rolling window — assistant-introduced entities and
         // named-antecedent follow-ups from recent turns now resolve too.
-        windowTurns: getWindowTurns(messages),
+        windowTurns: getWindowTurns(effectiveMessages),
         resolveEntities: ctx.resolveEntities,
       });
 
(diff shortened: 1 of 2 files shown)


time-attack:
Triage bot: routed to human review after adversarial analysis. The bug shape is real (ContextEngine declares prompt?: string at context-engine.ts:63 but assemble() drops it, so a prompt-only turn never reaches the reflex) and the fix avoids double-counting (prompt is ignored whenever messages is non-empty). Two blockers for auto-merge: (1) zero tests despite a trivially extendable retrieval-reflex.test.ts suite, and (2) the core claim about codex-app-server delivering turns via prompt with empty messages is unverifiable from this repo. Adding a small test for the prompt-only path would likely unblock this.

OrcunJoyoLabs:
Thanks for the adversarial pass — both points are fair. Addressed in `9dfa9c27`:

**Blocker 1 (zero tests).** Added two tests to `test/retrieval-reflex.test.ts`:

- `prompt`-only turn with empty `messages` still fires the reflex (the codex-app-server delivery shape). I verified it's a real guard, not a rubber stamp: reverting `context-engine.ts` to the pre-fix `assemble()` makes this test **fail** (`0 pass, 1 fail`), and restoring the fix makes it **pass**.
- `prompt` is ignored when `messages` is non-empty — asserts an entity named only in `prompt` never reaches the resolver, which locks in the no-double-count / backward-compat behavior.

Full file is `24 pass, 0 fail`.

**Blocker 2 (codex-app-server claim unverifiable from this repo).** Correct — the `prompt`-with-empty-`messages` delivery lives in the OpenClaw host, not in gbrain, so it can't be proven from this tree. Two things bound the risk regardless:

1. The `ContextEngine` interface already declares `prompt?: string` on `assemble()` (context-engine.ts:63), so the contract explicitly anticipates a turn arriving via `prompt`. The old code accepted the param and silently dropped it — this fix just honors it.
2. The change is a strict no-op whenever `messages` is non-empty (new test 2 pins this), so even if a given runtime never uses the `prompt` path, nothing regresses.

The live-runtime half of the claim is validated out-of-repo: on a gbrain v0.44.1.0 / OpenClaw 2026.7.1 gateway the reflex only began firing on real turns after this change.

time-attack:
Re-trialed against the current integration branch (post v0.46.23-25 waves): Conflict: src/core/context-engine.ts assemble() destructure — PR adds `prompt`, wave added sessionId/sessionKey (Cathedral 5). NOT superseded: wave still ignores prompt. Trivial union fix; test file merged clean. No red flags. — a rebase along those lines puts this back in the active merge queue; re-trials run nightly, so a fresh push gets picked up automatically. Thank you!

OrcunJoyoLabs:
@time-attack seems like the commit is going in with the next release - assuming no action items from me for rebase?

garrytan:

Issue #3660, opened 2026-07-30
deepseek recipe declares supports_prompt_cache: false; DeepSeek caches by default, and a warning built on the flag steers operators off the provider


exGeni:
## Summary

`src/core/ai/recipes/deepseek.ts` declares `supports_prompt_cache: false`. DeepSeek's context caching is on by default and demonstrably works. The false declaration drives a warning that prints on **every CLI invocation** and advises operators to move off the provider for a reason that is not true.

## The declaration contradicts itself in the same file

```ts
// src/core/ai/recipes/deepseek.ts
:98    supports_prompt_cache: false,
:100   cost_per_1m_input_usd: 0.14, // deepseek-v4-flash cache-miss baseline
```

Line 100 names a *cache-miss* price two lines below a declaration that there is no cache.

## What the vendor says

DeepSeek's API documentation, context caching guide:

> The DeepSeek API Context Caching on Disk Technology is enabled by default for all users, allowing them to benefit without needing to modify their code.

https://api-docs.deepseek.com/guides/kv_cache

## Measured against the live API

Two identical requests, `deepseek-chat`, same 379-token prompt, seconds apart:

```
call 1: prompt_tokens=379 prompt_cache_hit_tokens=0   prompt_cache_miss_tokens=379
call 2: prompt_tokens=379 prompt_cache_hit_tokens=256 prompt_cache_miss_tokens=123
```

The API returns `prompt_cache_hit_tokens` / `prompt_cache_miss_tokens` per call, and the second request hit cache without any client-side change.

## Consequence

`src/core/model-config.ts:273` and `src/commands/doctor.ts:3089` print, on every invocation:

```
[models] tier.subagent resolved to "deepseek:deepseek-v4-flash" via "models.default" — provider does not support prompt caching. The loop will run hot (cost scales linearly with conversation length). For lower cost on long loops, set models.tier.subagent to an Anthropic model.
```

Two problems with that line, independent of each other:

1. Its premise is false, and the advice it gives follows from the false premise.
2. It lands on **stdout**, ahead of the output of read-only commands. `gbrain sources`, `gbrain models list` and others emit it before their own output, so naive parsing of any command breaks.

## Scope of the claim

To be precise about what this report does and does not assert:

- It does **not** ask gbrain to implement Anthropic-style explicit cache-control breakpoints. DeepSeek's caching is server-side and automatic; there is nothing for a client to steer.
[...]

PR #3904, opened 2026-08-08
v0.45.16.0 fix(recipes,ai,config): provider freshness — live Google/OpenAI defaults, pullable ollama tags, native base-url config plane

garrytan:
## Summary

Wave 2 of the sequenced backlog plan: provider-compat freshness. Before implementing, a 6-agent audit re-verified all 21 planned items against current master — 16 had already landed via the v0.42.58.0 provider-agnostic wave and follow-ups (#2820, #2800, #3449, #3451, #3080, #2373, #2936), so this PR ships only the 5 surviving gaps plus the guards that keep the class closed. Closes #2507.

**Recipe freshness (the dead-default purge)**
- Google recipe: the 2.0-flash family was shut down 2026-06-01, so every chat/expansion model in the recipe was dead — a Google-only install could not chat. Now the GA 3.x line (`gemini-3.6-flash` / `3.5-flash` / `3.5-flash-lite`); the retiring 2.5 family is deliberately unlisted anywhere a default resolves.
- OpenAI recipe: `gpt-5.2` dropped off the live price sheet with the GPT-5.6 GA (2026-07-09) → chat is 5.6-only (`terra`/`luna`/`sol`); `gpt-4o-mini` stays in expansion but leaves chat (the touchpoint-wide 1M `max_context_tokens` would over-report its 128K window 8x). OpenRouter's `models[0]` — what `gbrain init` auto-pick persists — swapped from the dead `openai/gpt-5.2` to the OR-verified `openai/gpt-5.6-terra`.
- `DEFAULT_ALIASES`: `gemini` pointed at a shut-down preview id for a month; now `google:gemini-3.6-flash` (no GA pro-class Gemini exists — documented). `gpt` → `openai:gpt-5.6-sol`, tracking OpenAI's own `gpt-5.6` alias.
- `CANONICAL_PRICING`: rows for every new default; `gpt-5.5` reconciled to the live sheet ($4/$16 → $5/$30); retired ids stay priced so historical usage/audit rows keep pricing — including `gemini-2.0-flash-exp`, the exact `models[0]` string older env-auto-inited Google brains persisted, which would otherwise trip the budget tracker's fail-closed `no_pricing` path. Default panels (cross-modal slot A, takes-quality) follow. All rates verified against live provider sheets 2026-08-08 (ai.google.dev, OpenAI platform docs, OpenRouter).
[...]

PR #4037, opened 2026-08-12
fix(recipes): providers that cache prompts automatically no longer read as cache-less (#3660)

Masashi-Ono0611:
## What

`supports_prompt_cache` answers one question: *does this provider cache prompts
at all?* `capabilities.ts` reads it to classify a chat model as `ok` vs
`degraded:no_caching`, and `enforceSubagentCapable` turns `degraded:no_caching`
into operator-facing advice — *"provider does not support prompt caching. The
loop will run hot ... For lower cost on long loops, set `models.tier.subagent`
to an Anthropic model."*

Two recipes declared `false` while their providers cache by default, so that
advice fired against providers that were already caching — and pointed at a more
expensive one.

## The two providers

**DeepSeek** (`recipes/deepseek.ts`) — the subject of #3660. Context caching is
on by default for every account, the API reports
`prompt_cache_hit_tokens` / `prompt_cache_miss_tokens`, and there is no client
opt-in to make. The recipe contradicted itself: the same touchpoint's
`cost_per_1m_input_usd` is annotated *"cache-miss baseline"*, which already
presupposes a cache.

**Native OpenAI** (`recipes/openai.ts`) — named in #3660's triage as the same
class. OpenAI caches prompt prefixes automatically. gbrain already implements
the OpenAI half of this: `openAIPromptCacheKey()` derives a `prompt_cache_key`
routing hint for `native-openai` recipes, and its own docstring opens with
*"OpenAI caches prefixes automatically"* — while the recipe two files away said
the provider has no prompt caching.

OpenAI gets a **per-model predicate**, not a blanket `true`: automatic caching
starts at the `gpt-4o` / `gpt-4.1` / o-series generation, and recipe model lists
are advisory, so any id a user configures reaches the flag. A blanket `true`
would suppress the cost warning on a legacy id like `gpt-4-turbo` that really
does run hot — the same wrong-advice failure, pointing the other way.

Every branch is boundary-anchored (family token, then `-`, `.` or end), so an id
that merely shares a prefix — `gpt-4oops`, `o9anything`, a custom base-url alias
— does not inherit the family's answer. Fine-tunes (`ft:<base>:…`) resolve
through their base model. Unrecognized ids return `false` on purpose: an
unnecessary warning is cheaper than a silently hidden cost regression.

OpenRouter routing variants (`:online`, `:nitro`, `:floor`) are stripped before
delegating, since they are an OpenRouter concept rather than part of the
upstream id.

DeepSeek stays a plain boolean because there the behavior is a property of the
whole provider, not of the model.

An allow-list will always trail OpenAI's naming — an id outside the listed
families (a new alias, a custom base-url name) reads as `false` and earns one
unnecessary warning. That is the deliberate direction of failure, and adding a
family later is a one-line change with a test row.

**One thing I could not verify, flagged rather than assumed:** how the newest
GPT-5.6 generation places its implicit cache breakpoint in a *growing* tool
loop. If it anchors on the latest message rather than falling back to a matching
earlier prefix, a long subagent loop could miss more often than the "caches
automatically" framing suggests, and gbrain currently emits only
`prompt_cache_key` (a routing hint), not an explicit breakpoint. That does not
change the older generations or DeepSeek, and it is a separate question from the
stale declaration this PR fixes — but if you know the answer, it is worth
deciding whether the predicate should be narrower at the top end.

## The inconsistency this removes

The repo already counts automatic server-side caching as `true`:
`openrouterSupportsPromptCache()` returns `true` for `openai/gpt-*` routes,
commented *"OpenAI chat routes cache automatically"*.

So before this change, the **same OpenAI model** classified as:

| Route | verdict |
|---|---|
| `openrouter:openai/gpt-5.2` | `ok` |
| `openai:gpt-5.2` | `degraded:no_caching` |

`test/ai/capabilities.test.ts` asserted both, a few assertions apart in the same
`describe` block.

### …and does not create the mirror image of it

Fixing only the native recipes would leave `deepseek:deepseek-v4-flash` as `ok`
while `openrouter:deepseek/*` stayed `degraded:no_caching` — the same bug
pointing the other way. So `openrouterSupportsPromptCache()` gains the
`deepseek/` family, which is also what #3660's triage suggests ("Consider the
same for openrouter's deepseek/* routes").

For OpenAI the two routes now **share one predicate**: the OpenRouter branch
strips the `openai/` prefix and defers to the same function the native recipe
uses. Agreement is structural rather than two lists that happen to match today —
and it fixes the routed side's own over-claim, since the old branch returned
`true` for every `openai/gpt-*`, legacy ids included.

The predicate stays an allow-list, not a blanket true: the "does not mark every
OpenRouter route as cache-capable" test keeps its job, re-pointed at a family
the predicate does not list. And a new test makes the underlying rule explicit
— *a provider reachable two ways reports the same caching verdict either way* —
so native and routed views cannot silently drift apart again.

## Wire safety

Once the flag resolves `true`, a `cacheSystem: true` caller makes the gateway
pass `system` as a `SystemModelMessage` carrying
`providerOptions.anthropic.cacheControl`.

That is not a new shape — it is exactly what the shipped
`openrouter:openai/*` route produces today, captured from the transport args:

```
openrouter:openai/gpt-5.2  system = {"role":"system","content":"SYS",
                                    "providerOptions":{"anthropic":{"cacheControl":{"type":"ephemeral"}}}}
openai:gpt-4o-mini         system = (identical)
```

The Anthropic-namespace marker is inert off Anthropic: the AI SDK routes
`providerOptions` by provider key, so an `anthropic` entry never reaches an
OpenAI request body. That is the same reason the OpenRouter OpenAI route can
already carry it safely. A new test pins that the native and OpenRouter shapes
**agree** rather than diverging, so a future change cannot quietly split them.

The "leaves `system` a bare string" path is untouched — see Scope below for how
its test picks a model without smuggling in a vendor claim.

## Everything the verdict feeds

`classifyCapabilities` has four consumers, so here is the complete blast radius
rather than just the advice surface:

| Consumer | Before | After |
|---|---|---|
| `model-config.ts` `enforceSubagentCapable` | warns "provider does not support prompt caching … set `models.tier.subagent` to an Anthropic model" | no warn for these providers |
| `doctor.ts` `subagent_capability` | same message as a `warn` check | `ok` |
| `minions/queue.ts` + `handlers/subagent.ts` submit gates | reject only on `unusable:no_tools` / `unknown` | unchanged — `degraded:*` was never a refusal |
| `handlers/subagent.ts:959` `cacheSystem = verdict === 'ok' \|\| 'degraded:no_parallel'` | `false` for these providers | **`true`** — the loop now asks for caching |

That last row is the only behavior change beyond advice, and it is the point of
the fix: the subagent loop stops assuming these providers cannot cache. On the
wire it resolves to the shape documented above — inert off Anthropic, identical
to what the OpenRouter OpenAI route already sends.

## Scope

Only `deepseek` and `openai` change (plus the OpenRouter routes that must move
with them).

**Known remaining cases, named rather than quietly left behind:**

- `moonshot` — #3660's triage lists it as the same class. Not verified here, so
  not flipped on a guess.
- `google` — Gemini does have caching (implicit on recent models, plus explicit
  context caching), so the recipe's `false` is likely stale for the same reason.
  The mechanism to fix it now exists here — a per-model predicate — but drawing
  the right generation boundary for Gemini needs research I have not done, and
  guessing it is what this PR argues against. Left for someone who can state the
  boundary with a source.
  Where the tests needed a recipe that declares no caching, they use `google`
  and say so explicitly — including a guard assertion that fails loudly if that
  declaration is ever corrected, so the test gets re-pointed instead of quietly
  testing nothing.

Noticed while working here but deliberately **not** touched, since it predates
this change and the diff does not go near it: `ChatResult.usage`'s doc says the
`cache_*` counters are "present only when the active provider returned them
(Anthropic)", while the implementation has long also read the AI SDK's
provider-neutral `usage.cachedInputTokens` for OpenAI-compatible routes. Making
non-Anthropic caching real here makes that wording more visible, but correcting
it belongs in its own change.

This PR also does not touch the overloaded-flag question the triage raises
(splitting "caches automatically" from "needs an explicit `cache_control`
marker"). That is a design call for the maintainer; this PR corrects factual
declarations and leaves the flag's shape as-is.

## Sequencing note

This overlaps textually with the open #3904 (provider freshness): that PR
rewrites the same `openai` chat touchpoint block (`models`,
`max_context_tokens`, prices) and renames the model ids these tests use
(`openai:gpt-5.2` → `gpt-5.6-terra`, `google:gemini-1.5-pro` →
`gemini-3.6-flash`). Nothing here changes the same *fields* it changes — this
touches one flag in that block, plus test model-id strings — so whichever lands
first, the other is a mechanical rebase. Happy to rebase on request. (The
predicate here already returns `true` for `gpt-5.6-*`, so that rename needs no
follow-up on this side.)

(Worth noting for that PR's own scope: after its rename, the "recipe that
declares no caching" example becomes a Gemini 3.x id, which makes the stale
`google` declaration more visible. The guard assertion added here fails loudly
if that declaration is ever corrected, so the test gets re-pointed rather than
silently testing nothing.)

## Tests

- `capabilities.test.ts` — OpenAI reads as cache-capable; `classifyCapabilities`
  returns `ok` for both auto-caching providers; **new** test pinning the
  generation boundary (`gpt-4o`/`gpt-5.x`/o-series cache, `gpt-4-turbo` and
  older do not); **new** invariant *a provider reachable two ways reports the
  same caching verdict either way*, including a legacy id that must be
  cache-less on both routes.
- `gateway-chat.test.ts` — the registry invariant enumerates which recipes
  declare caching, instead of hardcoding "only Anthropic + OpenRouter". It is
  scoped to *current declarations*, with the known-stale ones called out in the
  test itself so it does not read as a claim that everything else is cache-less.
- `gateway-cache-breakpoint.test.ts` — **new** native-vs-OpenRouter parity test;
  **new** case pinning that a cacheable-but-implicit route (DeepSeek) gets the
  breakpoint *without* the Anthropic rewrite header; the no-caching case
  re-pointed, with a guard assertion on the declaration it depends on.
- `recipe-openrouter.test.ts` — `deepseek/*` now cacheable; a non-listed family
  keeps the negative case honest.
- `model-config.serial.test.ts` — the cost-warn test re-pointed at
  `google:gemini-1.5-pro`, plus a new test asserting an auto-caching provider is
  accepted **without** the warn.

**Red check:** reverting the three declaration sites (both recipe flags and the
OpenRouter branch) turns 10 tests red, including the new "no warn",
generation-boundary and route-parity assertions.

**Green:** 103 pass / 0 fail across the five directly affected files;
`bun run verify` is 37/37; the full unit suite is green apart from one failure
that reproduces identically on an unmodified checkout.

Addresses #3660.

### DIFF
--- a/scripts/module-size-limits.tsv
+++ b/scripts/module-size-limits.tsv
@@ -9,7 +9,7 @@ src/core/postgres-engine.ts	5837	ratchet	grown v0.46.23.0: putPage empty-overwri
 src/core/pglite-engine.ts	5881	ratchet	grown v0.46.23.0: putPage empty-overwrite guard (#4335)
 src/core/migrate.ts	668	region-exempt	append-only MIGRATIONS array grows freely; runner logic is ratcheted
 src/commands/sync.ts	4327	ratchet	grown v0.46.24.0 serve-delegated sync: onProgress seam + help text
-src/core/ai/gateway.ts	4428	ratchet	watchlist; merged: dream-wave turn permits + key-aware effective-model reconfigure + env-only worker refresh + discovery endpoint-trust docstring
+src/core/ai/gateway.ts	4439	ratchet	watchlist; merged: dream-wave turn permits + key-aware effective-model reconfigure + env-only worker refresh + discovery endpoint-trust docstring + #4037 prompt-cache fix
 src/cli.ts	3499	ratchet	grown v0.46.24.0: serve-delegated sync preflight hook
 src/core/cycle.ts	3080	ratchet	grown v0.46.20.0 phase-boundary split (resolveCyclePhases + normalizeQueuedSourcePhases + deriveStatus exclusion filter)
 src/commands/serve-http.ts	3194	ratchet	grown v0.46.23.0 review fix wave: github-kind webhook gating + case-insensitive repo match
--- a/src/core/ai/capabilities.ts
+++ b/src/core/ai/capabilities.ts
@@ -30,9 +30,17 @@ export interface ProviderCapabilities {
   supportsToolCalling: boolean;
 
   /**
-   * Anthropic-style ephemeral prompt cache markers honored. When false, the
-   * loop runs hot (no cache_control injection) and per-turn costs scale
-   * linearly with conversation length. Doesn't break the loop; just costs more.
+   * The provider caches prompt prefixes at all — by either mechanism:
+   * automatically server-side (OpenAI, DeepSeek; nothing to attach, and the
+   * Anthropic-namespace marker the gateway adds is inert on them), or when the
+   * request carries explicit `cache_control` markers (Anthropic). When false,
+   * the loop runs hot and per-turn costs scale linearly with conversation
+   * length. Doesn't break the loop; just costs more.
+   *
+   * This is deliberately "does it cache", not "does it honor our markers":
+   * `enforceSubagentCapable` and `doctor` use it to decide whether to warn an
+   * operator off a provider for cost reasons, and that advice is wrong for a
+   * provider that caches without being asked.
    */
   supportsPromptCaching: boolean;
 
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -3169,8 +3169,13 @@ export interface ChatOpts {
   maxTokens?: number;
   abortSignal?: AbortSignal;
   /**
-   * Anthropic-specific: cache the system prompt + last tool def. Silently
-   * ignored on providers without `supports_prompt_cache`.
+   * Ask for the stable prefix (system prompt + last tool def) to be cached.
+   * Silently ignored on providers whose recipe declares no prompt caching.
+   *
+   * Only Anthropic reads the resulting `cache_control` markers. Providers that
+   * cache prefixes automatically (OpenAI, DeepSeek) need no markers, and the
+   * Anthropic-namespace `providerOptions` this attaches never reach their
+   * request body — the AI SDK routes provider options by provider key.
    */
   cacheSystem?: boolean;
 }
@@ -3344,8 +3349,10 @@ function mapStopReason(
 
 /**
  * Run one chat completion turn. Provider-neutral wrapper over Vercel AI SDK's
- * `generateText`. Tool-use blocks are normalized; cache_control markers are
- * applied only on Anthropic when `cacheSystem: true`.
+ * `generateText`. Tool-use blocks are normalized. `cacheSystem: true` engages
+ * the caching path on any provider whose recipe declares prompt caching; the
+ * `cache_control` markers it attaches are read only by Anthropic, and are inert
+ * on providers that cache prefixes automatically.
  *
  * Crash-resumable replay is the caller's responsibility (subagent.ts persists
  * blocks via the provider-neutral schema landing in commit 2a).
@@ -3842,7 +3849,11 @@ export interface ToolLoopOpts {
   /** Per-turn max output tokens. Default 4096. */
   maxTokens?: number;
   abortSignal?: AbortSignal;
-  /** Apply Anthropic cache_control to system + last tool. Silently ignored elsewhere. */
+  /**
+   * Ask for the stable prefix (system + last tool) to be cached. Forwarded to
+   * `chat()`; see `ChatOpts.cacheSystem` for what each provider does with it.
+   * Silently ignored on recipes that declare no prompt caching.
+   */
   cacheSystem?: boolean;
 
   /** Crash-replay state. When set, the loop resumes from the recorded position. */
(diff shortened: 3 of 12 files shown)


time-attack:

time-attack:
Trial-merge against current master now conflicts in test/ai/capabilities.test.ts, test/ai/gateway-chat.test.ts, and scripts/module-size-limits.tsv (v0.46.23.0 touched all three). The fix itself is wanted (fixes #3660) — please rebase and it goes straight back into the queue.

Masashi-Ono0611:
Rebased over master at 498e4dd6 — mergeable, CLEAN. Ready for re-trial.

time-attack:

Issue #3730, opened 2026-08-01
resolveEntitySlug never consults page_aliases — a page declaring the exact alias still fails to resolve


wright-io:
## Summary

`resolveEntitySlug` never consults `page_aliases` — so a page that already declares the exact alias being resolved still fails to resolve, and facts written under that name become phantom slugs.

Environment: gbrain 0.42.59.0, Postgres engine.

## Detail

- `page_aliases` is populated from each page's frontmatter `aliases:` (`src/core/import-file.ts:867-876`). Our brain has 210 aliases across 95 pages.
- `src/core/entities/resolve.ts` contains **zero** references to it. Its four stages are exact-slug → bare-name prefix expansion → fuzzy trigram → slugify.
- `page_aliases` is currently consulted only by search (`src/core/search/hybrid.ts:648`).

Observed: `people/rob-macauslan` declares alias "Robert MacAuslan" and `people/trish-wheeler` declares "Patricia Wheeler" — yet facts extracted under those names became phantom slugs, forcing a human to confirm a mapping the brain already knew.

## Suggested fix

Add an alias stage right after exact-slug. Two constraints we found necessary when implementing this as a local patch:

1. Require an **unambiguous** match — an alias claimed by 2+ pages (e.g. "Dave") must not resolve.
2. Require **multi-token** aliases — a bare given name declared on one page means that person is *sometimes called* that, not that every "Scott" is them (this mirrors the resolver's own comment at `resolve.ts:69-71`: "a shared first name is not specific enough").

Also match the alias in its **reversed and namespace-stripped** shapes: `normalizeAlias` keeps punctuation (`"Wheeler, Patricia"` → `"wheeler, patricia"`), so a literal-only comparison misses the very forms extractors emit.

We are running this as a local patch; happy to PR it if useful.

wright-io:
Verified still present on current `main` (as of 2026-08-01), not just 0.42.59: `src/core/entities/resolve.ts` on main contains zero references to `page_aliases` or `resolveAliases` — the four stages remain exact-slug → bare-name prefix expansion → fuzzy trigram → slugify, and the alias table populated by `import-file.ts` is still consulted only by search. Line numbers in the body above are per the 0.42.59 release.

wright-io:
Correcting one sentence in my own report before it misleads anyone.

[...]

PR #4052, opened 2026-08-12
feat(entities): save-time entity resolution — alias-aware resolver cascade

SeanGearin:
## Problem

`save_facts` and the conversation-facts backfill mint `entity_slug` values without consulting the alias substrate the brain already maintains. Surface forms get slugified more or less verbatim, so one person accumulates fragmented slugs over time. In a production brain here, one contact ended up spread across 11 slugs, plus a 1,068-fact bucket under a typo slug. The `slug_aliases` and `page_aliases` tables exist and are populated, but `resolveSlugWithAlias` had zero callers on any write path. The write side and the alias substrate never met.

## Change

Two alias hops added to the shared resolver cascade in `resolveEntitySlugWithSource`, ahead of the existing exact/fuzzy/prefix/fallback chain and without weakening any of it:

1. Slug-shaped input: exact live page first (unchanged), then a source-scoped `slug_aliases` redirect so old slugs land on their canonical page. Tag: `alias_redirect`.
2. Any input that misses above: normalized (`normalizeAlias`) exact match against source-scoped `page_aliases`. Exactly one target resolves with tag `alias_match`. Multiple targets are recorded as `ambiguous_aliases` metadata and resolution falls through: the write path never guesses between two people.

The conversation-facts backfill now resolves each extracted `entity_slug` through the same resolver before row construction, best-effort per fact. An ordinary resolver failure keeps the raw value and increments `resolution_errors`, while aborts and `BudgetExhausted` still propagate. `fallback_slugify` outcomes are counted per run so operators can see phantom-slug minting as it happens instead of discovering it later. Resolution telemetry commits only after the data batch persists.

## Failure-mode hardening

- A permission-denied (SQLSTATE 42501) or missing-table (42P01) alias probe degrades to the pre-alias cascade with a once-per-process warning, rather than throwing - the connecting role needs SELECT (and any applicable row-security policy) on the alias tables, and a deployment without that access keeps writing exactly as before. No schema migration ships with this change: alias-table access for non-owner roles is a deployment decision, deliberately left to operators.
- Transaction-pinned alias probes are isolated with savepoints on both engines, so a failed probe cannot poison the surrounding transaction. This was verified with real-database constructions on PGLite for both SQLSTATEs. The PGLite savepoint stack is serialized per pinned handle because concurrent probes could otherwise roll back past one another.

## What this does not do

No LLM or embedding calls anywhere on the path: resolution is SQL-only. There is no interactive disambiguation on the write path, since ambiguity falls through deterministically and is surfaced as metadata. Backstop's fallback policy is untouched beyond what the shared resolver gives it.

## Testing

24 new tests across two files. Highlights: the lowercase bare-name case where slug-shaped input must still reach `page_aliases` (an LLM extractor emitting `kendall` should land on the canonical person, not mint a phantom), multi-alias fall-through without guessing, real-PGLite transaction recovery for both SQLSTATEs inside `engine.transaction()`, concurrent-probe serialization, and the same denied-probe recovery reproduced against a live Postgres 17 server (savepoint isolation held; the transaction stayed usable), budget/abort propagation in the backfill, and no telemetry for unpersisted facts.

`bun run typecheck` clean. `bun run verify` 38/38. Focused affected suites 140/140 on the branch. The no-alias-rows regression surface is pinned: existing resolver behavior, including globally-unambiguous bare-name expansion and the 0.7 fuzzy floor, is unchanged.

One field note: while this branch was being verified, the production brain running the pre-change resolver filed a fact about "entity resolution" under an unrelated `debt-resolution` page via a trigram match on the word "resolution". That is the exact failure class the alias-first cascade addresses.

### DIFF
--- a/scripts/module-size-limits.tsv
+++ b/scripts/module-size-limits.tsv
@@ -18,7 +18,7 @@ src/core/search/hybrid.ts	2629	ratchet	grown v0.46.23.0 review fix wave: superse
 src/core/engine.ts	2351	ratchet	grown v0.46.23.0: putPage allowEmptyOverwrite contract (#4335)
 src/commands/autopilot.ts	2301	ratchet
 src/commands/extract.ts	2161	ratchet
-src/commands/extract-conversation-facts.ts	2045	ratchet	grown v0.46.x per-provider gateway-unavailable copy
+src/commands/extract-conversation-facts.ts	2090	ratchet	grown v0.46.x per-provider gateway-unavailable copy + save-time resolution caller
 src/core/import-file.ts	2000	ratchet	grown v0.46.11.0 five-issue wave
 src/core/cycle/synthesize.ts	2702	ratchet	merged: five-issue wave + dream-wave C7 peel + C8 manifest + C9 mode/telemetry/backfill + adversarial fix batches
 src/commands/embed.ts	1963	ratchet
--- a/src/commands/extract-conversation-facts.ts
+++ b/src/commands/extract-conversation-facts.ts
@@ -94,6 +94,12 @@ import { assertFactsEmbeddingDimMatchesConfig } from '../core/embedding-dim-chec
 import { writeReceipt, shortRunId } from '../core/extract/receipt-writer.ts';
 import { upsertExtractRollup } from '../core/extract/rollup-writer.ts';
 import { ALLOWED_TYPES, type AllowedType } from '../core/facts/conversation-types.ts';
+import {
+  emptySaveTimeResolutionCounts,
+  formatSaveTimeResolutionCounts,
+  mergeSaveTimeResolutionCounts,
+  resolveExtractedEntitiesForSave,
+} from '../core/entities/resolve-on-save.ts';
 
 // Re-exported verbatim so existing importers (this file's own helpers below
 // and this file's tests) keep working unchanged; doctor.ts, jobs.ts,
@@ -367,6 +373,10 @@ export interface ExtractConversationFactsResult {
   segments_processed: number;
   facts_extracted: number;
   facts_inserted: number;
+  /** Entity values that reached the shipped deterministic fallback slug path. */
+  fallback_slugify_count: number;
+  /** Entity values kept raw after a best-effort resolution failure. */
+  resolution_errors: number;
   budget_exhausted?: boolean;
   spent_usd?: number;
 }
@@ -1053,6 +1063,7 @@ async function processPage(
   let newestEnd: string | null = null;
   let segmentsThisPage = 0;
   let pageInsertedTotal = 0;
+  const pageResolution = emptySaveTimeResolutionCounts();
 
   for (const seg of segments) {
     if (state.segmentLimit > 0 && segmentsThisPage >= state.segmentLimit) break;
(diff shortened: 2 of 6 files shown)


SeanGearin:
Follow-up commit `1a821f3a`: hardening from a later adversarial review round against a downstream deployment of this feature — alias targets are now validated against a live page before resolving (a dangling alias could previously stamp facts with a slug no live page owns), `setPageAliases` is safe inside pinned transactions, and the Postgres savepoint recovery is serialized per transaction to match the PGLite discipline. Red-first regressions for all three; full verify green.

time-attack:
Triage bot: this adds new product surface, so it's queued for the maintainer's own product decision rather than automated review. No action needed from you.

SeanGearin:
Second follow-up `38a5291b`, two fixes from running this feature in production downstream:

- The resolver's candidate pools now admit only entity pages (schema-pack-derived predicate, falling open to the bundled gbrain-base taxonomy when no pack resolves). This fixes a class that exists in the base resolver as well: the fuzzy arm matches any page by title with no page-type predicate, so a non-entity page whose title collides with an entity surface (e.g. a note titled "sean") can win resolution — reproducible on the base branch via the extraction/backstop paths. The alias hops this PR adds get the same live-entity check on their targets.
- The new predicate itself could collapse to an empty type set in pack-less environments and silently disable resolution — that one is specific to this PR, now failing open to the default taxonomy with a warning.

Red-first regressions for both, on both engines; full verify green.

SeanGearin:
```
This rescope closes the extract-skip hole on the display-name lane by calling shipped `alias_exact` at save time. Conversation-facts extract writes through `insertFacts` and never calls `writeSingleFact`, so a display name never reached the resolver. The new `resolve-on-save` helper is a thin caller of `resolveEntitySlugWithSource` before that insert.

It deliberately does not cover the old-slug redirect lane. Original #4052 hopped `slug_aliases` via `resolveSlugWithAlias` and tagged `alias_redirect`. Adversarial review of this tree found `resolveSlugWithAlias` still has zero write-path callers. After this commit, extract and `writeSingleFact` match on display names. Both still miss `slug_aliases`. A slug-shaped old slug such as `people/brian-old` is stored as `fallback_slugify` instead of redirecting to the live page.

Propose a focused follow-up PR that extends the shipped resolver with that one hop. The production symptom it closes is old slugs minting phantom pages on write paths, including `writeSingleFact`.

Rebased onto current master (v0.46.21.0). The parallel `alias_match` / `alias_redirect` cascade is gone. The `slug_aliases` hop is gone. Savepoints and the tenant GRANT are gone too. The schema-pack entity-type filter from `38a5291b` is not in this update.

Tests on this tree: `tsc --noEmit` clean; save-time hook file 5/5; PGLite extract integration 7/7; master's five `alias_exact` tests untouched (blob `a311588e`) and `test/entity-resolve.test.ts` 32/32.
```

time-attack:
Status: deliberately NOT adopted in the v0.46.23.0 wave — your engine-side conflict resolutions were sound, but the alias-resolution vocabulary (alias_match/alias_redirect vs master's alias_exact from #3730) is a product decision that deserves a deliberate maintainer landing rather than a squash-time call. This stays open pending that decision; everything else from the wave is merged. Thank you for the patience.

SeanGearin:
One clarification in case the wave notes were drafted before yesterday's push: the PR no longer contains the alias_match/alias_redirect cascade. It was rescoped onto master's shipped alias_exact (head a7aa2f33, details in my comment above), so the only open product question left is the narrower one from the follow-up proposal: whether master wants the old-slug redirect lane at all. One input for whenever you take that up: resolveSlugWithAlias currently has zero write-path callers on master, so old slugs mint phantom pages on every write path today. Not urgent on my end. Happy to wait for the deliberate landing either way.

time-attack:
Re-trialed against the current integration branch (post v0.46.23-25 waves): Conflict only in scripts/module-size-limits.tsv (PR bumps extract-conversation-facts ceiling 2045->2090; wave rewrote adjacent rows). Not superseded: resolve-on-save.ts+tests absent on wave. No red flags. Trivial re-resolve. — a rebase along those lines puts this back in the active merge queue; re-trials run nightly, so a fresh push gets picked up automatically. Thank you!

garrytan:

PR #4056, opened 2026-08-12
search: supersession-aware rerank — down-rank + SUPERSEDED flag for superseded pages

rspinabellaagent:
## Motivation
`recall.ts` already honors `superseded_by` / `expired_at` on the **facts** table, but
`hybridSearch`'s post-fusion stages had no equivalent awareness at the **page** level. A
page that is the target of a `supersedes` link — i.e. a newer/canon page has explicitly
superseded it — still ranked purely on semantic score, so stale material could out-rank
the current canon that replaced it. Agents doing retrieval (and the contradiction probe)
had no signal that a hit was superseded.

## Change
A new final post-fusion stage, `applySupersedeDownrank`, modeled on the existing
`applyAliasResolvedBoost`:

- One source-agnostic index lookup on `links` where `link_type = 'supersedes'` and
  `to_page_id = ANY(result page_ids)` (a page is *superseded* when it is the `to` side of
  a `supersedes` edge; `A supersedes B` → `from = A` canon, `to = B` stale).
- For each superseded result: stamp `superseded = true`, `superseded_by = <canon slug>`,
  and multiply `score` by `SUPERSEDE_PENALTY` (0.5).
- Runs **last** so the penalty applies to the fully-boosted score; **fail-soft** per the
  established per-stage contract (a brain with no `supersedes` edges, or a pre-`links`
  schema, gets 0 rows / a caught throw → no-op).

`SearchResult` gains three optional, back-compatible fields — `superseded`,
`superseded_by`, `supersede_penalty` — surfaced for `gbrain search --explain`, the
contradiction probe, and agent renderers. The annotation is authoritative even in
reranked modes, where the cross-encoder owns the final head order and the score penalty
governs head entry / non-reranked ordering.

## Why this shape
- Mirrors `applyAliasResolvedBoost` (same composite-lookup + stamp pattern, inverse sign)
  so it slots into the existing staged pipeline with no new plumbing or opts.
- No new migration, no schema change — reads the existing `links` + `pages`.
- Additive fields only; every existing caller is unaffected when there are no
  `supersedes` edges.

## Testing
- `tsc --noEmit` clean (strict), on current `master`.
- Verified live against a production brain: a page that was the #1 organic hit
  (`base_score 0.896`) and the target of a `supersedes` link was demoted to rank 4
  (`score 0.464`) with `superseded=true`, `supersede_penalty=0.5`,
  `superseded_by=<canon slug>`; the superseding canon page ranked #1 on the same query.
  `sources`, `capture`, `embed`, and the contradiction eval were all unaffected.

## Open questions for the maintainer
- Penalty is a fixed `0.5` — happy to make it a `mode`/config knob if you'd prefer it tunable.
- Deliberately does **not** re-order within the cross-encoder reranked head (to avoid
  perturbing autocut's rerank-score-cliff logic); the SUPERSEDED flag is the guaranteed
  signal there. Open to a post-rerank demotion variant if you want stronger ordering.

### DIFF
--- a/src/core/search/hybrid.ts
+++ b/src/core/search/hybrid.ts
@@ -614,6 +614,80 @@ export async function runPostFusionStages(
   } catch {
     // Non-fatal; preserves the per-stage contract.
   }
+
+  // supersession stage — runs LAST so the penalty applies to the fully-boosted
+  // score. Down-ranks results whose page is the target of a `supersedes` link
+  // (a newer/canon page supersedes it) and stamps `superseded`/`superseded_by`
+  // for --explain, the contradiction probe, and agent renderers. The page-level
+  // analogue of the superseded_by/expired_at awareness recall.ts applies to
+  // facts. Fail-open per the per-stage contract: a brain with no `supersedes`
+  // edges (or a pre-links schema) finds 0 rows / throws and no-ops.
+  try {
+    await applySupersedeDownrank(results, engine);
+  } catch {
+    // Non-fatal; preserves the per-stage contract.
+  }
+}
+
+/**
+ * Down-rank results whose page is superseded by a newer/canon page, and stamp
+ * the SUPERSEDED annotation.
+ *
+ * A page X is "superseded" when it is the `to_page_id` of a `supersedes` link
+ * (`A supersedes B` → from=A canon, to=B stale). This is the page-level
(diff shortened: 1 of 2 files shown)


time-attack:
Triage bot: this adds new product surface, so it's queued for the maintainer's own product decision rather than automated review. No action needed from you.

time-attack:

PR #4066, opened 2026-08-13
fix(bootstrap): refuse to bootstrap directly into $HOME

Masashi-Ono0611:
## Summary

`gbrain bootstrap`'s `--workspace` defaults to `process.cwd()`. An SSH
session frequently lands in `$HOME` before the operator `cd`s into a
project, so an unqualified `gbrain bootstrap <subcommand>` run can
silently target the home directory.

Observed (real-world, not hypothetical) fallout of running bootstrap at
`$HOME`:
- Later bootstrap phases stage `git add -A` over the resolved workspace.
  At home-directory scale this hung for 50+ minutes in practice.
- The generated `.gitignore` template is not a guaranteed superset of
  every sensitive dotfile `$HOME` may hold (e.g. `~/.ssh/`), so a
  home-directory-scale `git add -A` risks staging secrets.

## Fix

`resolveWorkspace()` in `src/commands/bootstrap.ts` now throws a new
`BootstrapError('HOME_WORKSPACE')` when the resolved `--workspace` (or
its `cwd` default) IS the operator's home directory — compared via
`realpath` so a symlinked `$HOME` or a symlinked cwd (e.g. macOS's
`/var` → `/private/var` tmp roots) still matches. The home value used
is `process.env.HOME`, falling back to `os.homedir()` only when `HOME`
is unset or empty (matches the existing `src/core/preferences.ts:home()` /
`src/commands/upgrade.ts` house pattern for Bun's `os.homedir()`, which
caches its value at process start).

**Known, documented limitation:** this does NOT protect against a
process already launched with `HOME` pointed away from the real account
home — Bun's `os.homedir()` mirrors the process-start `HOME` value
rather than doing an independent OS/uid lookup, so both would read back
the same overridden value in that scenario. This guard covers the
real-world trigger described above (an SSH session whose shell sets
`HOME` normally and whose `cwd` defaults there); it is not a general
sandboxing mechanism. Called out explicitly in the `resolveWorkspace()`
comment and in `docs/architecture/KEY_FILES.md` rather than overstating
the protection.

Subcommand validation and the workspace-free `cloud-setup-script`
printer now dispatch BEFORE workspace resolution, so they keep working
from any cwd including `$HOME` (previously `resolveWorkspace()` ran
before the unknown-subcommand check, so both would have been
incorrectly caught by the new guard too). `cloud-setup-script` keeps its
own try/catch so an asset-load failure still returns exit 1 with the
usual message + support hint instead of throwing out of `runBootstrap()`.

The refusal is still appended to `<home>/bootstrap/install.jsonl`
(keyed by the rejected candidate path, since there is no confirmed
workspace to attribute it to otherwise) — same audit trail every other
`BootstrapError` gets.

### (A) hard-stop vs (B) `--force` flag

Went with **(A) — a simple error stop**, not a new `--force`-style flag.
This keeps the change a minimal defect fix (existing, unguarded
data-loss/secret-exposure risk) rather than adding a new opt-in CLI
surface. The fix is always the same: re-run with `--workspace <dir>`
pointing at the actual project directory.

## Scope

Workspace resolution + this safety guard only. Expanding the
`.gitignore` template's dotfile coverage is a related but separate
concern and intentionally NOT addressed here.

## Testing

- `bun test test/bootstrap-dispatcher.serial.test.ts` — 31/31 pass,
  including 8 new tests for the guard: cwd-defaults-to-$HOME refused,
  explicit `--workspace $HOME` refused, non-home workspace unaffected
  (no regression), `cloud-setup-script` still works from `$HOME`,
  unknown subcommand from `$HOME` still exits 2 (not masked), `HOME`
  unset falls back to `os.homedir()` without crashing, a symlinked
  `$HOME` is refused in both directions (realpath equivalence), and a
  mutating subcommand refused at `$HOME` makes zero exec-runner calls /
  writes nothing / but IS logged to `install.jsonl`.
- `bun test test/bootstrap-status.serial.test.ts` — 16/16 pass (no
  regression from the dispatcher restructuring).
- `bash scripts/check-test-isolation.sh` — OK.
- `bash scripts/check-key-files-current-state.sh` + `bun run build:llms`
  — docs/architecture/KEY_FILES.md updated to current-state, bundle
  regenerated (no diff — that doc is link-only in the bundle).
- Reviewed with `codex exec -c model_reasoning_effort=high` across three
  passes; all Critical/Warning-equivalent (P2/P3) findings addressed
  (workspace-free subcommands bypassing the guard, the overridden-HOME
  comparison not being independently verified, install-log audit trail
  being skipped on the early refusal, and the `cloud-setup-script`
  exit-code contract regression introduced while fixing the first
  finding).


time-attack:
Triage bot: routed to human review. Adversarial pass verified the guard itself is correct (realpath comparison handles symlinked $HOME, APFS casing, trailing slashes; new tests are red on master, green on head) BUT found a real recovery-path concern: users who ALREADY bootstrapped into $HOME — the exact victims this PR cites — appear locked out of `bootstrap uninstall`/`status` because the guard fires before dispatch with no exemption or override, and the refusal message gives cleanup guidance that doesn't apply to them. Suggest exempting read-only/cleanup subcommands (status, uninstall) from the guard, or adding an explicit override flag. Queued for the maintainer.

Masashi-Ono0611:
The recovery-path concern is real and is now fixed. Thank you — I had the guard right about the
dangerous action and wrong about the boundary.

`resolveWorkspace()` now takes the subcommand and skips the `HOME_WORKSPACE` throw for a narrow
exemption set, `HOME_WORKSPACE_GUARD_EXEMPT = { status, uninstall }`. I chose the exemption over an
override flag deliberately: a flag is new surface, and it would also put the burden on precisely the
user who is already confused about why the tool refuses them.

The classification is by what each subcommand actually does to the workspace, not by name:

- `status` only reads and prints a report.
- `uninstall` removes exactly `receipt.created_paths`, each containment-checked in
  `core/bootstrap/uninstall.ts` — no `git add`, no commit, no push, no workspace-wide scan.
- everything else (`interview` / `render` / `repo` / `hooks` / `verify` / `attach`) stages, commits,
  pushes, or writes identity/registration files into the workspace, so all of it stays refused.

The refusal message no longer gives advice that cannot apply to the person reading it. It now names
the reachable path first, and only then suggests `--workspace` for the case where someone genuinely
meant a different directory.

Verified by removal, not by reasoning: reverting only `src/commands/bootstrap.ts` while keeping the
new tests fails exactly the three new cases (status-at-$HOME, uninstall-at-$HOME, and the refusal
message naming the recovery path); restoring it is 42 pass / 0 fail. The guard's own tests are
untouched and still pass — including `a mutating subcommand refused at $HOME makes no runner calls,
writes nothing, and IS logged`, which is the assertion that the exemption did not quietly widen the
hole. `bun run typecheck` and `bun run verify` (43/43) both exit 0.

Rebased onto current master while I was in there, so this is `MERGEABLE` again.

time-attack:

Issue #3045, opened 2026-07-21
conversation_format_coverage 0/50: built-in patterns don't match the bulleted "- **Speaker:** text" meeting-transcript shape (colon inside bold, no timestamp)


andreineacsu:
## Summary

On a real brain, `doctor` `conversation_format_coverage` reports 50/50 sampled conversation-eligible pages (100%) match NO built-in parser pattern (`_no_match=50`), and `conversation_facts_backlog` is 556 eligible pages. All eligible pages are `type=meeting`, whose `## Transcript` turns use the bulleted `- **Speaker:** utterance` shape — colon inside the bold markers, no per-line timestamp — which none of the built-in patterns recognize. Running `extract-conversation-facts` against 0% parser coverage extracts no real facts while (per #2610) still able to write terminal audit rows, so the backlog silently "drains" with zero facts learned.

## Evidence (gbrain 0.42.63, one live store)

- `doctor`: `conversation_format_coverage: 50/50 conversation pages (100.0%) match NO built-in pattern. Breakdown: _no_match=50`.
- `doctor`: `conversation_facts_backlog: 556 eligible pages without extraction`.
- `gbrain conversation-parser scan <meeting-slug>`: `phase: no_match`, `patterns_scored: 16`, `messages: 0`, `unmatched_lines: 317`.
- The transcript-turn shape (synthetic, structurally identical to the real pages):
  ```
  ## Transcript
  - **Alex Carter:** we should ship the wedge first
  - **Priya Rao:** agreed, and pull the review forward
  ```
  Leading `- ` list bullet, `**Name:**` with the colon inside the bold, no `(HH:MM)` / date token.
- Pages carry no `raw_transcript` frontmatter sidecar (frontmatter is type/title/ingested_at/source_kind/ingested_via), so the sidecar-preference path from the earlier parser fix doesn't apply — the transcript lives in the compiled `## Transcript` body.

## How this differs from the closed parser issues (searched open+closed)

- #1897 (raw_transcript sidecar + plain `Speaker A:` lines): our pages have no sidecar and use bold + a leading bullet.
- #1533 (line-counting on `imessage-slack` pages): that pattern needs a parenthetical `YYYY-MM-DD` datetime; our lines carry no timestamp.
- #2756 (iMessage type allowlist + 12h time-only + multi-day headers): meeting is already allow-listed; this shape has no time token at all and a leading `- ` bullet with colon-inside-bold.

[...]

PR #4068, opened 2026-08-13
fix(conversation-parser): handle ChatGPT export multi-paragraph turns

Masashi-Ono0611:
## Problem

ChatGPT's web-export → Markdown conversion shape anchors every turn with a
literal `**You:**` / `**ChatGPT:**` label, followed by a blank line and then
a multi-paragraph reply (often 20-30+ lines of prose before the next
anchor). On a real 156-line export with only 4 real anchor lines, running
`gbrain conversation-parser scan` reports `phase: no_match, messages: 0,
unmatched_lines: 99` — the page never gets extracted.

Root cause: the existing `bold-name-no-time` pattern (`/^\*\*(?!\[)(.+?):\*\*\s*(.*)$/`,
`multi_line: false`) is the only builtin that can match a bare
`**Label:** text` anchor line, but it treats every following prose line as
unrelated, non-matching evidence for density purposes. 4 real anchors among
156 non-blank lines ≈ 2.6% density, well below `SCORING_MIN_ACCEPTANCE =
0.05`, so the page is rejected outright.

`bold-name-no-time` is intentionally narrow this way — it carries an
explicit **BROAD-REGEX GUARD** comment: because `**Label:** text` is a
common prose idiom (`**Note:**`, `**Owner:**`, etc.), a notes page with a
few bold labels clustered in its first 10 lines could otherwise
false-positive as a conversation. It defends against this via
`score_full_body: true`, pinned by an existing regression test (`F1` in
`test/conversation-parser/parse.test.ts`).

## Fix (option a — new scoped pattern, chosen for reasons below)

Adds a new builtin pattern, `chatgpt-export-you-chatgpt`, declared just
before `bold-name-no-time` in `BUILTIN_PATTERNS`. Its speaker capture is a
**closed, literal two-value enumeration** (`You` or `ChatGPT` only) —
unlike `bold-name-no-time`'s open `(.+?)` capture, it can never match an
arbitrary `**Label:** text` prose idiom, so it structurally cannot reopen
the BROAD-REGEX GUARD `bold-name-no-time` depends on.

I considered adjusting `bold-name-no-time`'s own scoring instead (option
b), but that pattern's breadth is exactly what its guard exists to
constrain — loosening its density math to tolerate long continuations would
weaken the guard for every caller of that pattern, not just the ChatGPT
shape. A new, narrowly-scoped pattern keeps the blast radius to the two
literal labels this format actually uses.

To correctly absorb the multi-paragraph reply body without the same
density problem, the new pattern composes three gates (`multi_line: true`
plus three `PatternEntry` scoring flags — the latter two new, added in this
PR, both opt-in and no-ops for every existing pattern):

- **`score_continuations_as_body`** (existing mechanism, mirrors
  `bold-time-dash`): excludes non-`**`-prefixed reply lines from the
  density denominator entirely, so long multi-paragraph replies don't
  dilute the anchor ratio.
- **`score_continuations_min_distinct_speakers: 2`** (new field): a bare
  `**You:**`/`**ChatGPT:**` heading — unlike `bold-time-dash`'s bold-name +
  valid-24h-time + dash anchor — is a plausible label in ordinary prose
  ABOUT ChatGPT. Without this, a page with just one such heading (or
  several repeats of the same role) would get the same density immunity a
  genuine two-party exchange gets. This requires BOTH roles to actually
  appear.
- **`score_continuations_max_preamble_lines: 5`** (new field): distinct-
  speaker count alone still lets one genuine You/ChatGPT pair — both roles
  present — through when merely embedded somewhere inside an otherwise
  unrelated long document (a tutorial, a "how I use ChatGPT" article).
  Requiring the FIRST anchor to land within the first 6 scored lines
  (index `<= 5`, tolerating a short title/heading before the transcript
  starts) keeps that immunity scoped to pages that look like a real export
  from the top.
- **`score_full_body: true`** (existing mechanism, mirrors
  `bold-name-no-time`): belt-and-suspenders full-body recompute of the
  winning candidate, so acceptance never depends on where the first anchor
  happens to land inside the 10-line head-pass window.

### BROAD-REGEX GUARD safety — how it's preserved

- The exact `bold-name-no-time` F1 notes-page fixture (`**Attendees:**` /
  `**Date:**` / `**Goal:**` + 80 prose lines) is reused in a new test: the
  new pattern's enumerated regex cannot match any of those labels at all,
  so `anchorCandidates = 0` and `scorePatternFull` returns exactly `0`
  regardless of any continuation-scoring mechanism.
- Two additional false-positive classes that the plain
  `score_continuations_as_body` mechanism would otherwise open (found
  during adversarial multi-model review, see below) are closed by the two
  new gates: a solitary `**You:**` heading opening a long document, and a
  genuine You/ChatGPT pair merely embedded deep inside an unrelated
  document. Both are pinned by new regression tests.
- `bold-name-no-time` is unaffected for every label except the two literal
  `You`/`ChatGPT` enumeration values: a bare `**You:**`/`**ChatGPT:**` anchor
  now resolves to this new pattern instead, by design (the declaration-order
  tie-break above). `bold-paren-time` and `telegram-bracket` are unaffected —
  non-shadowing tests confirm `**Alice Example:**`, `**Assistant:**`,
  `**User:**`, `**You** (00:00): text`, and `**[18:37] 👤 You:** hello` all
  still resolve to their existing patterns.

## Review process

Went through 3 rounds of adversarial review via `codex exec -c
model_reasoning_effort=high` before submission:

- **Round 1**: 2 Warnings — (1) the plain `score_continuations_as_body`
  mechanism grants full density immunity to any body containing a
  qualifying anchor, regardless of surrounding prose (closed by adding
  `score_continuations_min_distinct_speakers`); (2) the original regression
  test's numbers didn't actually cross `bold-name-no-time`'s acceptance
  floor (fixed: increased reply length to 40 lines/turn, giving `4/84 ≈
  0.048 < 0.05`, and the test now asserts this directly via
  `scorePatternFull`). Plus 1 Nit (tightened `quick_reject`).
- **Round 2**: No Critical findings, 1 Warning — the residual "one genuine
  pair embedded anywhere" false-positive class (closed by adding
  `score_continuations_max_preamble_lines`). Plus 2 Nits (both fixed:
  avoid counting an empty/undefined capture as a distinct speaker; avoid
  the `exec()` allocation for patterns that don't need distinct-speaker
  tracking).
- **Round 3**: **No Critical or Warning findings.** 1 Nit (boundary test
  coverage + a comment-wording precision fix) — both addressed.

## Scope

Matching-mechanism only. `gbrain conversation-parser validate`'s
`simple_pattern` compilation (referenced in issue #3045) is a separate,
unimplemented TODO stub and is intentionally **out of scope** for this PR.

## Testing

- `bun test test/conversation-parser/` — 139 pass, 0 fail across 5 files
  (includes the 9 new tests for this pattern: basic multi-paragraph parse,
  the original density-floor defect reproduction, notes-page
  BROAD-REGEX-GUARD reuse, two non-shadowing checks, solitary/repeated-role
  heading, embedded-pair, and the exact preamble-boundary test)
- `bun run typecheck` — clean
- `bun run check:conversation-parser` (fixture-corpus CI gate against
  `all.jsonl`) — 22/22 fixtures pass
- `bash scripts/check-fixture-privacy.sh` — OK
- `bash scripts/check-key-files-current-state.sh` — OK (reference docs
  updated to describe current state, per this repo's own maintenance rule)

Opening as **Draft** per house convention — happy to switch to Ready once
CI is green.


time-attack:
Triage bot: all gates passed EXCEPT merge cleanliness — this PR now conflicts with master after #4005 (markdown-heading pattern) merged today; KEY_FILES.md needs conflict resolution and the pattern count/enumeration should reflect both new patterns (19). One of two independent adversarial reviews verified the parser change itself is safe (gated scoring additions are undefined-guarded no-ops for other patterns; 9 adversarial probes held; new tests are red on master). Please rebase onto current master and let CI re-run — this should be a quick re-queue.

Masashi-Ono0611:
Merged current master into the branch (711e3291) per the triage note. The KEY_FILES.md conflict is resolved: the conversation-parser paragraph now reads 19-pattern and enumerates both new patterns (markdown-heading-turn from #4005 and chatgpt-export-you-chatgpt from this PR). conversation-parser suite 142 pass, extract-conversation-facts baseline 66 pass, typecheck clean, CI green on the merged head (23 successful checks).

Masashi-Ono0611:
Rebased onto current master; this is `MERGEABLE` again.

On the conflict you flagged: it turned out not to be a content conflict at all. #4005 appends
`markdown-heading-turn` at the END of `BUILTIN_PATTERNS`, while this PR inserts
`chatgpt-export-you-chatgpt` mid-array (declared just before `bold-name-no-time`, per that block's
own design-order comment), so the two edits do not overlap and git's three-way merge resolved them
without markers. What had gone stale was the branch, not the patterns.

Verified rather than assumed, since a duplicate pattern id makes the module refuse to load:
19 unique ids, no duplicates, and the declaration order is preserved
(`chatgpt-export-you-chatgpt` still precedes `bold-name-no-time`). #4005's three
`markdown-heading-turn` tests pass alongside this PR's, so the two patterns coexist rather than one
shadowing the other.

- load-bearing test — `REGRESSION: long multi-paragraph replies no longer starve the density floor`: pass
- 11 conversation-parser test files: 263 pass / 0 fail
- `bun run typecheck` and `bun run verify` (43/43): both exit 0

One thing I noticed and deliberately did NOT touch, since it is pre-existing and outside a rebase's
remit: inline comments in `builtins.ts` and `parse.ts` still say "18 hand-vetted" / "18 candidates".
The real count is 19 now, and it was already off by one before #4005 landed. No test asserts the
literal number. Happy to sweep it in a separate one-line PR if you want it fixed.

time-attack:

PR #4070, opened 2026-08-13
fix(consolidate): fetch pending facts before limit

### LINKED ISSUE #4057: consolidate: bucket eligibility counts unconsolidated facts, but the per-bucket fetch is capped at 100 active rows with no consolidated_at filter and no offset — buckets over 100 active facts skip forever
Verified against v0.45.0.0 (`d35c9c9`). All line references are to that tag.

## Summary

The `consolidate` phase certifies a bucket as eligible using a count of its
**unconsolidated active** facts, then fetches a window of its **active** facts —
capped at 100, with no `consolidated_at` predicate and no pagination. When one
`(source_id, entity_slug)` bucket has more than 100 active facts, that window can
be filled mostly with already-consolidated rows. The in-memory re-filter then
drops below the threshold and the bucket is counted in `buckets_skipped`, even
though it genuinely satisfies the phase's own documented rule.

This is not a request to make the thresholds tunable. #3042 was closed on the
grounds that the phase "behaves exactly as its documented rules specify", and
that is the right standard — this report is that under this condition it does
*not* meet that standard. The eligibility query and the fetch query disagree
about which rows they are counting, so a bucket that passes the documented rule
is skipped anyway. #3824 covers the min-bucket and cluster-size gates; this is a
separate defect and does not depend on either threshold's value.

## Mechanism

**1. The scan counts unconsolidated active facts.**
`src/core/cycle/phases/consolidate.ts:69-75`:

```sql
SELECT source_id, entity_slug, COUNT(*)::int AS count
FROM facts
WHERE consolidated_at IS NULL
  AND expired_at IS NULL
  AND entity_slug IS NOT NULL
GROUP BY source_id, entity_slug
HAVING COUNT(*) >= ${minPerBucket}
```

A bucket reaching this loop is certified: it has at least `minPerBucket`
unconsolidated, unexpired facts.

**2. The fetch asks for a different row set.**
`consolidate.ts:102-105` calls `listFactsByEntity(..., { activeOnly: true, limit: 100 })`.
That query filters on `expired_at IS NULL` only — there is no `consolidated_at`
predicate:

- Postgres: `src/core/postgres-engine.ts:4607-4614` (`WHERE source_id … AND entity_slug … AND expired_at IS NULL … ORDER BY valid_from DESC, id DESC LIMIT … OFFSET …`)
- PGLite: `src/core/pglite-engine.ts:4622-4633` → `_listFacts`, `src/core/pglite-engine.ts:4889-4895` (same `activeOnly` → `expired_at IS NULL`, same ordering)

**3. Consolidated facts stay in that row set.**
`consolidateFact` sets only `consolidated_at` and `consolidated_into`; it never
sets `expired_at` (`postgres-engine.ts:4731-4734`, `pglite-engine.ts:4833-4838`,
and the interface contract at `src/core/engine.ts:1895-1899`). So every fact ever
consolidated for an entity remains "active" and continues to occupy a slot in the
100-row window, forever.

**4. The cap is hard and cannot be raised by the caller.**
`limit: 100` is already at `MAX_SEARCH_LIMIT = 100` (`src/core/engine.ts:650`),
and both engines route the value through `clampSearchLimit(opts.limit, 50, MAX_SEARCH_LIMIT)`
(`engine.ts:653-657`, `postgres-engine.ts:4601`, `pglite-engine.ts:4889`). Passing
a larger limit would be silently clamped back to 100.

**5. No cursor advances across cycles.**
`consolidate.ts:102-105` passes no `offset`, and both engines default it to 0
(`postgres-engine.ts:4602`, `pglite-engine.ts:4890`). Every cycle re-reads the
same 100 newest-by-`valid_from` rows.

**6. The in-memory filter then starves the bucket.**
`consolidate.ts:107` applies `f.consolidated_at == null` to the truncated window,
and `consolidate.ts:108-111` increments `buckets_skipped` when fewer than
`minPerBucket` survive.

Net: for an entity with more than 100 active facts, unconsolidated facts that sort
outside the 100 newest by `valid_from DESC, id DESC` are never examined, and no
mechanism ever advances past them.

Note that `valid_from` is the fact's semantic validity date, not its insertion
time — the phase's own age gate treats it that way (`consolidate.ts:114-120`). A
newly inserted fact carrying an older `valid_from` therefore sorts *below* the
window on arrival and is unreachable from the moment it is written.

**Secondary effect on the age gate.** `consolidate.ts:114-116` computes `oldest`
from the truncated set. When truncation removes the genuinely oldest
unconsolidated facts, the gate at `consolidate.ts:117` evaluates a newer "oldest"
than the bucket actually has, so it can skip on age as well. Same root cause.

## Reproduction

1. Pick one `(source_id, entity_slug)` bucket. Give it more than 100 active facts.
2. Let a cycle consolidate some of them, so a substantial share of the newest 100
   by `valid_from` carry a non-null `consolidated_at`. (Equivalently: insert
   unconsolidated facts whose `valid_from` is older than the 100 newest rows.)
3. Ensure at least `minFactsPerBucket` unconsolidated active facts exist in the
   bucket overall — i.e. the scan at `consolidate.ts:69-75` still certifies it.
4. Run the phase.

Expected: the bucket is processed, since it satisfies the documented count rule.
Actual: fewer than `minFactsPerBucket` unconsolidated rows survive the filter at
`consolidate.ts:107`, and the bucket lands in `buckets_skipped`. It does so on
every subsequent cycle, with no state that would ever let it recover.

The condition requires more than 100 active facts for a single
`(source_id, entity_slug)` pair. Below that, the window covers the whole bucket
and the phase behaves as documented.

## Minimal fix

Align the fetch's predicate with the scan's. Add an unconsolidated-only option to
`FactListOpts` (`src/core/engine.ts:543-555`) that pushes `consolidated_at IS NULL`
into the `WHERE` in both engines, and have `consolidate.ts:102-105` pass it. The
100-row window is then drawn from exactly the rows the scan counted, and the
in-memory re-filter at `consolidate.ts:107` becomes redundant.

A bucket with more than 100 unconsolidated facts would still process 100 per
cycle, but it would make progress every cycle instead of never — and the next
cycle's window would be drawn from what remains.

If a change to the engine surface is unwanted, the alternative is to paginate in
the phase: `FactListOpts` already carries `offset` (`engine.ts:547`) and both
engines honour it (`postgres-engine.ts:4602/4614`, `pglite-engine.ts:4890`), so
`consolidate.ts` could page until it has collected the `count` the scan reported.
That costs extra round-trips on large buckets, which is why aligning the
predicate seems preferable.

Either way the invariant worth asserting is that the rows the phase evaluates are
the rows the phase counted.

arisgysel-design:
## Summary

- add an `unconsolidatedOnly` predicate to the shared facts-list options in both engines
- apply the predicate before ordering and the 100-row consolidation limit
- cover the starvation case with 100 newer consolidated facts hiding three older pending facts
- verify the predicate-before-limit behavior against real Postgres

Fixes #4057.

## Why

The consolidate phase selected eligible buckets by counting active facts with `consolidated_at IS NULL`, but fetched each bucket from all active facts and filtered only after `LIMIT 100`. Consolidated rows could fill the entire window, so an eligible bucket was skipped forever and its older pending facts were never evaluated.

The fetch now uses the same predicate as the eligibility query. The work window therefore contains only pending facts; the existing cap, ordering, age gate, and clustering behavior remain unchanged.

## Verification

- red-first regression: 100 consolidated facts backed by a real take plus three older pending facts produced `facts_consolidated = 0` before the fix and `3` after it
- `bun test test/consolidate-valid-until.test.ts test/facts-engine.test.ts test/facts-separation-pglite.test.ts test/build-llms.test.ts` — 34 pass, 0 fail
- `bun run typecheck` — pass
- `gitleaks dir . --redact --no-banner` and `gitleaks git --redact --no-banner` — pass

### DIFF
--- a/docs/TESTING.md
+++ b/docs/TESTING.md
@@ -163,6 +163,7 @@ The quarantine has grown to dozens of files — treat it as debt: every addition
 
 Unit tests and what they cover:
 
+- `test/facts-engine.test.ts` / `test/consolidate-valid-until.test.ts` — facts-list filtering and consolidate correctness (#4057): `unconsolidatedOnly` is applied before the 100-row limit, so newer consolidated facts cannot permanently hide older pending facts; the phase regression seeds 100 consolidated rows plus three older pending rows and requires all three to progress.
 - `test/markdown.test.ts` — frontmatter parsing; `splitBody` sentinel precedence, horizontal-rule preservation, `inferType` wiki subtypes.
 - `test/chunkers/recursive.test.ts` — chunking.
 - `test/parity.test.ts` — operations contract parity.
@@ -267,6 +268,8 @@ Unit tests and what they cover:
 ### E2E test inventory
 
 E2E tests live in `test/e2e/` and run against real Postgres+pgvector (require `DATABASE_URL`), except where noted as PGLite in-memory (no `DATABASE_URL` needed).
+
+- `test/e2e/facts-separation-postgres.test.ts` — real-Postgres parity for cross-session facts, supersession, and the pre-limit `unconsolidatedOnly` predicate used by consolidation.
 
 - `bun run test:e2e` runs Tier 1 (mechanical, all operations, no API keys). Includes dedicated cases for the postgres-engine `addLinksBatch` / `addTimelineEntriesBatch` bind path — postgres-js's JSONB bind (`jsonb_to_recordset(($1::jsonb)->'rows')`) differs from PGLite's and gets its own coverage.
 - `test/e2e/search-quality.test.ts` — search quality against PGLite (no API keys, in-memory).
--- a/src/core/cycle/phases/consolidate.ts
+++ b/src/core/cycle/phases/consolidate.ts
@@ -99,12 +99,11 @@ export async function runPhaseConsolidate(
       try { await opts.yieldDuringPhase(); } catch { /* keepalive errors non-fatal */ }
     }
 
-    const facts = await engine.listFactsByEntity(b.source_id, b.entity_slug, {
+    const unconsolidated = await engine.listFactsByEntity(b.source_id, b.entity_slug, {
       activeOnly: true,
+      unconsolidatedOnly: true,
       limit: 100,
     });
-    // Re-filter to unconsolidated since listFactsByEntity returns all active.
-    const unconsolidated = facts.filter(f => f.consolidated_at == null);
     if (unconsolidated.length < minPerBucket) {
       bucketsSkipped += 1;
       continue;
--- a/src/core/engine.ts
+++ b/src/core/engine.ts
@@ -543,6 +543,8 @@ export interface NewFact {
 export interface FactListOpts {
   /** Hide expired_at IS NOT NULL rows. Default true. */
   activeOnly?: boolean;
+  /** Return only facts where consolidated_at IS NULL. Default false. */
+  unconsolidatedOnly?: boolean;
   limit?: number;
   offset?: number;
   /** Restrict to specific kinds. Default: all kinds. */
--- a/src/core/pglite-engine.ts
+++ b/src/core/pglite-engine.ts
@@ -4908,6 +4908,9 @@ export class PGLiteEngine implements BrainEngine {
     if (opts.activeOnly !== false) {
       whereParts.push(`expired_at IS NULL`);
     }
+    if (opts.unconsolidatedOnly === true) {
+      whereParts.push(`consolidated_at IS NULL`);
+    }
     if (opts.kinds && opts.kinds.length > 0) {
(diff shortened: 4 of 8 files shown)


time-attack:

time-attack:
Re-trialed against the current integration branch (post v0.46.23-25 waves): Conflict: src/core/pglite-engine.ts + postgres-engine.ts (wave peeled listFactsByEntity into engine/facts.ts). Not superseded — post-limit re-filter bug still live. Mechanical rebase to peeled modules. No red flags. — a rebase along those lines puts this back in the active merge queue; re-trials run nightly, so a fresh push gets picked up automatically. Thank you!

garrytan:

PR #4071, opened 2026-08-13
fix(gateway): two Anthropic-compatible-provider defects that fail silently

arusa4ok:
Two independent fixes for running gbrain against an **Anthropic-compatible** endpoint rather than `api.anthropic.com`. Both were found on z.ai/GLM, both fail *silently* — no error line, no failed op — which is why they survived a long time. Neither changes behaviour on the native Anthropic path.

Happy to split this into two PRs if you'd prefer.

---

### 1. `expand()` degrades to the bare query — name the schema

`generateObject` turns the Zod schema into a tool. Without `schemaName`/`schemaDescription` the emitted tool carries **no `description` field**. `api.anthropic.com` tolerates that; at least one compatible endpoint does not — it ignores `tool_choice: { type: 'tool' }` and replies with markdown-fenced JSON as ordinary text.

`generateObject` then finds no object, `result.object` is `undefined`, and:

```ts
expansions = result.object?.queries ?? [];
```

quietly yields `[]`, so **query expansion silently reduces to the original query on every call**. Nothing is logged, because the `catch` only reports `AIConfigError`.

Measured against z.ai, deterministic both ways: **with** a description 3/3 `tool_use`; **without** it 3/3 `end_turn` + text.

This lands on the `recipe.implementation !== 'openai-compatible'` branch specifically. The two `openai-compatible` branches already recover through `viaText()`; the native branch has no fallback, so naming the schema is its only guard.

### 2. A missing `toolCallId` wedges the tool loop one turn later

The same class of provider intermittently returns a tool-call part with `toolCallId` **undefined**. AI SDK v6's `ModelMessage` schema requires a non-empty string, so the bad turn is invisible when produced and throws

> The messages do not match the ModelMessage[] schema

the moment it is replayed as history on turn N+1 — permanently wedging the job. We observed ~528 identical failures on one recurring subagent job before locating it.

`ensureToolCallId()` synthesizes a stable id **at the source** (`chat()` block-normalization) so the same id reaches both the persisted tool-call block and its matching tool-result block. It is also applied defensively in `toModelMessages`, so rows already persisted with a poisoned id stop throwing on replay.

---

**Scope:** `src/core/ai/gateway.ts` + one test file. No config, schema or dependency changes.

**Tests:** `bun test test/ai/` → 494 pass / 0 fail. The full suite has one pre-existing failure (`dream-postgres.serial` A1 per-source scope, `PGLite not connected`) which reproduces identically on unmodified `master` — 8 pass / 1 fail on both, checked side by side rather than assumed.

---

*Separately, and not part of this PR:* `provider_base_urls.<provider>` is documented and works for openai-compatible recipes, but is **inert for the two native providers** — `resolveNativeBaseUrl` reads only `cfg.env.<PROVIDER>_BASE_URL` and the native branches never consult `cfg.base_urls`. I wrote the obvious one-line fold and then withdrew it: the base URL and the API key travel through different planes, so folding only the URL can pair a config-plane endpoint with an env-plane key belonging to a different vendor. That seemed worth reporting rather than patching blind — happy to open it as an issue if useful.

### DIFF
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -23,7 +23,7 @@
 
 import { embed as aiEmbed, embedMany, generateObject, generateText, jsonSchema } from 'ai';
 import { AsyncLocalStorage } from 'node:async_hooks';
-import { createHash } from 'node:crypto';
+import { createHash, randomUUID } from 'node:crypto';
 import { listRecipes } from './recipes/index.ts';
 import { createOpenAI } from '@ai-sdk/openai';
 import { createGoogleGenerativeAI } from '@ai-sdk/google';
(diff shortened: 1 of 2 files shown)


time-attack:

time-attack:
Re-trialed against the current integration branch (post v0.46.23-25 waves): Conflicts in src/core/ai/gateway.ts (2 hunks): wave's #2373 expansion restructure vs schemaName fix; #4201 providerOptions map vs ensureToolCallId. Neither fix landed on wave (not superseded). Needs rebase. No red flags. — a rebase along those lines puts this back in the active merge queue; re-trials run nightly, so a fresh push gets picked up automatically. Thank you!

garrytan:

PR #4072, opened 2026-08-13
fix(doctor): effective_date_health re-parses frontmatter instead of trusting JSONB key-existence

Masashi-Ono0611:
## Summary

`gbrain doctor`'s `effective_date_health` check flags pages whose `effective_date_source =
'fallback'` "despite parseable frontmatter date". Its SQL used JSONB `?` (key-existence) as a
proxy for "has a parseable date": `frontmatter ? 'event_date' OR frontmatter ? 'date' OR
frontmatter ? 'published'`. Key-existence is not the same as "parses to a valid date" — an
empty string, `null`, or an unparseable value all satisfy `?` and still get counted.

This causes the warning to over-report and never clear: `gbrain reindex-frontmatter --dry-run`
(which runs the real `computeEffectiveDate` precedence chain) correctly finds nothing to fix for
those rows, so the doctor count and the reindex-frontmatter count disagree, and the warning
persists no matter how many times reindex-frontmatter runs.

## Changes

- `src/commands/doctor.ts`: the `effective_date_health` check's "fell back despite parseable
  date" arm now fetches the `effective_date_source = 'fallback'` candidate rows and re-runs
  `computeEffectiveDate()` — the same function `gbrain reindex-frontmatter` uses — over each one
  in JS, instead of trusting the JSONB key-existence predicate as a count. The call passes
  `filename: null` deliberately, so the row is only counted when the **frontmatter** fields
  (`event_date`/`date`/`published`) alone resolve to a valid date, independent of whichever
  source wins the full precedence chain:
  - a row whose frontmatter value is empty/unparseable but whose slug carries a `YYYY-MM-DD`
    prefix (so the full chain would resolve a *filename*-derived date) is correctly excluded —
    the message specifically says "despite parseable **frontmatter** date".
  - conversely, a `daily/`/`meetings/`-prefixed row with a genuinely parseable frontmatter date
    is still counted even though `computeEffectiveDate`'s filename-first precedence for those
    prefixes would otherwise resolve the *full* chain to `source: 'filename'` and mask it. This case is pinned by a dedicated test (`daily/2024-03-15-standup` + `{ date: '2024-04-01' }`).
- `src/core/backfill-effective-date.ts`: exported the existing `parseFrontmatter()` helper so
  the doctor check normalizes the raw `pages.frontmatter` JSONB value the same way the backfill
  walk does before feeding it to `computeEffectiveDate`.
- `test/doctor-effective-date-health.test.ts` (new): drives `buildChecks` against a real PGLite
  engine, covering:
  - a fallback row with an empty-string frontmatter date → not flagged
  - a clean brain (no pages) → reports ok
  - a fallback row with an unparseable frontmatter date → not flagged
  - a fallback row with a genuinely parseable frontmatter date → still flagged (true positive
    preserved), with the exact count asserted
  - a mix of empty/garbage/real fallback rows → only the real one is counted
  - a fallback row with no date-ish frontmatter keys at all → not flagged
  - a fallback row with an empty frontmatter date but a date-prefixed slug (`daily/2024-03-15-...`)
    → not flagged, since only a filename-derived date resolves, not a frontmatter field
  - a fallback row under `daily/` with an empty-vs-real distinction: a **real** frontmatter date
    on a `daily/`-prefixed, date-named slug → still flagged, confirming the filename-precedence
    shadowing case above doesn't suppress a genuine frontmatter date

**Known, intentional scope limit:** this check only ever flags rows where the *frontmatter*
itself is parseable. A `fallback` row whose frontmatter has no usable date at all, but whose
slug/filename carries a valid date, is a real "reindex-frontmatter would still update this row"
case that this specific check does not surface — because its message is scoped to frontmatter
dates, not filename dates. That's unchanged from the check's original intent (the sentinel column
comment above it), just now correctly evaluated.

## Test plan

- [x] `bun test test/doctor-effective-date-health.test.ts` — 8 pass, 0 fail
- [x] `bun test test/effective-date.test.ts` — 21 pass, 0 fail (no change to `computeEffectiveDate` itself)
- [x] `scripts/check-privacy.sh` — clean
- [x] `scripts/check-test-isolation.sh` — clean (1167 non-serial unit files scanned)
- [x] `scripts/check-pglite-embedded.sh` — clean

### DIFF
--- a/scripts/module-size-limits.tsv
+++ b/scripts/module-size-limits.tsv
@@ -3,7 +3,7 @@
 # Raising a ceiling is a conscious, reviewer-visible act. Lower ceilings in
 # the same commit as any peel (the guard fails on >50 lines of stale slack).
 # Columns: path	max_lines	policy	note
-src/commands/doctor.ts	4291	ratchet	peel target: containment sprint C8-C13; grown v0.46.x keyless not_configured state
+src/commands/doctor.ts	4348	ratchet	peel target: containment sprint C8-C13; grown v0.46.x keyless not_configured state + #4072 frontmatter re-parse
 src/core/operations.ts	303	ratchet	peel target: containment sprint C4-C7
 src/core/postgres-engine.ts	5816	ratchet	peel target: containment sprint C15; grown v0.46.11.0 five-issue wave + retrieval wave
 src/core/pglite-engine.ts	5744	ratchet	peel target: containment sprint C15; grown v0.46.11.0 five-issue wave + retrieval wave
--- a/src/commands/doctor.ts
+++ b/src/commands/doctor.ts
@@ -19,6 +19,8 @@ import { gbrainPath, loadConfig } from '../core/config.ts';
 import { dirname, join } from 'path';
 import { existsSync, readFileSync, readdirSync, statSync } from 'fs';
 import { resolveEnvNumber, resolveHoursEnv } from '../core/env-number.ts';
+import { computeEffectiveDate } from '../core/effective-date.ts';
+import { parseFrontmatter } from '../core/backfill-effective-date.ts';
 import { hnswIndexExpected, hnswMaxDimsForType } from '../core/vector-index.ts';
 import { VERSION as GBRAIN_BINARY_VERSION } from '../version.ts';
 // Peeled doctor modules (containment sprint): each is a verbatim move out of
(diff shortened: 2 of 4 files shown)


Masashi-Ono0611:
The `test (9)` failure (`per-source line prefix under withSourcePrefix`) is the known shard-composition flake tracked in #4061 — `test/sweep.test.ts`'s ENG-5 serve harness leaves `console-prefix.ts`'s stdout-redirect global flipped for later files in the same shard when it lands ahead of `console-prefix.test.ts`. This PR only touches `src/commands/doctor.ts`, `src/core/backfill-effective-date.ts`, and the new `test/doctor-effective-date-health.test.ts` — none of which are involved in that leak. Will retrigger once #4061 merges.

time-attack:

PR #4077, opened 2026-08-13
fix(dream): propagate cooperative abort through synthesis

Kkkkkuro:
## Summary

- propagate the cycle `AbortSignal` through synthesize and patterns phases
- forward cancellation into judge gateway calls, inline PGLite subagents, and completion waits
- stop new cache, fan-out, provenance, reverse-write, summary, and completion-stamp work after cancellation
- cancel active child jobs with a truthful `cancelled` outcome instead of surfacing cancellation as provider failure or success
- add regression coverage for judge transport, active inline children, patterns reverse-write, and PGLite fixture stability

## Why

The cycle runner already accepted a cooperative cancellation signal, but the synthesize path did not carry it all the way through judge calls, subagent execution, completion waits, or the gateway transport. A cancelled cycle could therefore continue working during the force-eviction grace period and potentially write derived state after cancellation.

This change threads the existing signal through those boundaries and adds explicit abort checks before every downstream write boundary. The sibling patterns path receives the same treatment.

## Test plan

Run serially with production database/provider environment variables removed:

- `bun test test/cycle-patterns-source-scope.test.ts test/cycle-abort.test.ts` — 18 pass, 0 fail
- `bun test test/cycle/synthesize-gateway-adapter.test.ts test/cycle-abort.test.ts test/e2e/dream-synthesize-pglite.test.ts` — 46 pass, 0 fail
- focused synthesize/patterns regression suite — 79 pass, 0 fail
- `bun run typecheck`
- `bun run check:search-path`
- `git diff HEAD^ HEAD --check`

The PGLite-heavy fixtures use explicit 60-second setup timeouts and guarded teardown so cold schema initialization does not create false failures or mask the original setup error.

Related: #1972

### DIFF
--- a/src/core/cycle.ts
+++ b/src/core/cycle.ts
@@ -1858,6 +1858,7 @@ export async function runCycle(
           // (explicit --source wins, else derived from the checkout dir).
           sourceId: cycleSourceId,
           once: opts.onceForPhase === 'synthesize',
+          signal: opts.signal,
         }));
         result.duration_ms = duration_ms;
         phaseResults.push(result);
@@ -2070,6 +2071,7 @@ export async function runCycle(
           // source owns the page, which is what doctor reports as
           // multi_source_drift.
           sourceId: cycleSourceId,
+          signal: opts.signal,
         }));
         result.duration_ms = duration_ms;
         phaseResults.push(result);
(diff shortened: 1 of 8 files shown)


time-attack:
Re-trialed against the current integration branch (post v0.46.23-25 waves): Conflicts: src/core/cycle.ts, cycle/patterns.ts, cycle/synthesize.ts, test/cycle-abort.test.ts. Not superseded — wave still lacks dream abort plumbing, but drain peeled to inline-drain.ts; PR needs rework. No red flags. — a rebase along those lines puts this back in the active merge queue; re-trials run nightly, so a fresh push gets picked up automatically. Thank you!

garrytan:

PR #4078, opened 2026-08-13
fix(ai): honor chat.supports_subagent_loop in capability classification

Masashi-Ono0611:
## Summary

`classifyCapabilities()` reads a recipe's `chat.supports_tools` but ignores its
`chat.supports_subagent_loop` flag. The OpenRouter recipe has declared
`supports_subagent_loop: false` since v0.37.6.0 (tool_use_id stability across
crashes/replays can't be guaranteed through the proxy — the recipe's own header
documents this), but nothing enforced it: an `openrouter:*` model submitted for a
subagent job passed the capability gate and ran the loop the recipe says it
cannot safely run.

This is the deliberately-held-back follow-up to #3919 (the handler-entry gate
verifying the finally-resolved model), which landed on master in the v0.45.8.0
fix-wave (#4050, commit 94ec7e31e0). This branch is rebased on top of that
landed state — the diff contains only the capability-classification change.

## Changes

- `src/core/ai/capabilities.ts`: `classifyCapabilities()` returns a new
  `unusable:no_subagent_loop` verdict when the resolved recipe's chat
  touchpoint declares `supports_subagent_loop: false`. `ProviderCapabilities`
  normalizes it as `chat.supports_subagent_loop === true`, so the verdict is
  fail-closed; that is safe because `supports_subagent_loop` is a REQUIRED
  field on `ChatTouchpoint` (`src/core/ai/types.ts`), so every recipe with a
  chat touchpoint declares it explicitly; the recipes in the registry that
  don't mention the flag are the embedding/rerank-only ones, which have no
  chat touchpoint to classify. `unusable:no_tools`
  is still checked first, so a provider lacking tools keeps its existing,
  more specific verdict.
- `src/core/minions/queue.ts`: the submit gate maps the new verdict to a
  specific rejection message (names the recipe flag instead of the generic
  no-tools wording).
- `src/core/minions/handlers/subagent.ts`: the dispatch-time gate (from #3919)
  refuses on the same verdict, so config-resolved models get the same
  enforcement as explicit `data.model`.
- `src/commands/doctor.ts`: the `subagent_capability` check reports the new
  verdict with the recipe-flag wording (warn-only, no new check added).
- `src/core/model-config.ts`: the subagent-tier resolver treats the new verdict
  the same way it already treats `unusable:no_tools` / `unknown` — fall back to
  `TIER_DEFAULTS.subagent` and warn, with wording naming the recipe flag.
- `src/commands/agent.ts`: help text only — the gateway-loop precondition now
  says a qualifying recipe must declare `supports_subagent_loop: true`, not
  just support tool calling.
- `src/core/ai/recipes/openrouter.ts` / `TODOS.md` / KEY_FILES: doc text
  updated to describe the enforced (not just declared) state, including the
  follow-up TODO now naming both gates a future relaxation would need to touch.

## Test plan

- [x] `bun test test/ai/capabilities.test.ts test/ai/recipe-openrouter.test.ts test/subagent-handler.test.ts test/doctor.test.ts test/agent-cli.test.ts test/model-config.serial.test.ts` — 230 pass, 0 fail (measured before the review amend; that amend changed doc comments plus one user-visible string in `src/core/minions/queue.ts` — `litellm-proxy` → `litellm`, matching the recipe's actual id — and no test asserts that wording)
- [x] `bun test test/ai/capabilities.test.ts test/subagent-handler.test.ts test/model-config.serial.test.ts` — 75 pass, 0 fail (re-run after the amend)
- [x] `bun run build:llms` — regenerated (KEY_FILES.md text changed)

Red-green: the new classification, queue, resolver, and handler assertions fail
on master (the flag is ignored there); the terminal-replay case is a
compatibility guard (#1151 precedent — already-terminal jobs replay without a
provider call and must not be dead-lettered by a config repoint), not a red
test.


time-attack:

PR #4080, opened 2026-08-13
fix(dream): resolve context limits from model recipes

Kkkkkuro:
## Summary

- make the official provider recipes and model resolver the single source of truth for chat context limits
- use recipe-declared context limits when sizing synthesis chunks, including DeepSeek v4 Flash/Pro at 1M tokens
- preserve explicit model overrides, provider defaults, and the conservative 180K fallback for genuinely unknown models
- keep capability reporting aligned with the same resolver

## Motivation

`src/core/cycle/synthesize.ts` maintained a private context-size table. Models absent from that table silently fell back to 180K even when their official recipe declared a larger context. In particular, both DeepSeek v4 chat models declare a 1M context window, but synthesis sized chunks as though they had 180K.

This removes the duplicate table and resolves context limits from the canonical recipe metadata instead.

## Resolution order

1. explicit runtime prompt-token override
2. recipe model-level `model_context_tokens`
3. recipe provider-level `max_context_tokens`
4. explicit conservative 180K fallback for unknown model/provider ids

The Anthropic model-specific 1M entries are moved from the private synthesis table into recipe metadata; other Anthropic models retain the existing 200K provider default.

## Test plan

- RED reproduced the DeepSeek 1M model being truncated to the 180K fallback
- chunk budget suite: 7 passed, 0 failed
- focused model/context suite: 46 passed, 0 failed
- PGLite synthesis chunking: 6 passed, 0 failed
- `bun run typecheck`
- `bun run check:search-path`
- `git diff HEAD^ HEAD --check`

All test runs cleared database and provider environment variables.

## Scope

This PR is independent of #4077. Both touch `src/core/cycle/synthesize.ts`, but in separate hunks; a three-way merge against their shared base is conflict-free. It does not include cooperative-abort changes, verdict-cache/dry-run work, thinking configuration changes, or production configuration changes.

### DIFF
--- a/src/core/ai/capabilities.ts
+++ b/src/core/ai/capabilities.ts
@@ -21,7 +21,7 @@
  * decisions don't depend on it.
  */
 
-import { resolveRecipe } from './model-resolver.ts';
+import { resolveChatContextTokens, resolveRecipe } from './model-resolver.ts';
 import { listRecipes } from './recipes/index.ts';
 import { AIConfigError } from './errors.ts';
 
@@ -105,7 +105,7 @@ export function getProviderCapabilities(modelString: string): ProviderCapabiliti
     // a `supports_thinking` field later without breaking this helper (it'll
     // just keep returning false until a recipe sets it).
     supportsThinking: false,
-    maxContext: chat.max_context_tokens ?? 128_000,
+    maxContext: resolveChatContextTokens(modelString) ?? 128_000,
   };
 }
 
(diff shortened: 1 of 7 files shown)


time-attack:
Re-trialed against the current integration branch (post v0.46.23-25 waves): Conflict: src/core/ai/capabilities.ts (wave #4172 supportsThinking block overlaps PR's maxContext line). NOT superseded — MODEL_CONTEXT_TOKENS map still hardcoded in synthesize.ts. Trivial rebase fixes it; no red flags. — a rebase along those lines puts this back in the active merge queue; re-trials run nightly, so a fresh push gets picked up automatically. Thank you!

garrytan:

PR #4083, opened 2026-08-13
fix(cli): reach real usage block for gbrain auth --help

### LINKED ISSUE #4003: `gbrain auth --help` is a dead end — redirects to `gbrain --help`, which doesn't list the auth group
### Summary

`gbrain auth --help` is a dead end: it prints a stub that redirects to `gbrain --help`, which does not document the `auth` command group at all. The only ways to discover `auth register-client` are to run bare `gbrain auth` (no flag) or read the source.

### Reproduce

`gbrain 0.42.63.0`:

```
$ gbrain auth --help
Usage: gbrain auth

gbrain auth - run gbrain --help for the full command list.
```

Follow that instruction:

```
$ gbrain --help | grep -i '^AUTH\|auth create\|register-client'
(nothing — there is no AUTH section, and no auth command is listed)
```

Bare invocation, however, has the real help:

```
$ gbrain auth
GBrain Token Management

Usage:
  gbrain auth create <name> ...
  gbrain auth register-client <name> ...
  gbrain auth revoke-client <client_id> ...
  gbrain auth test <url> --token <token> ...
```

### Why it matters

Other parts of gbrain instruct users to run `gbrain auth register-client`, e.g. the pre-flight failure hint in `gbrain init --mcp-only`:

> Hint: the host operator can run `gbrain auth register-client <name> --grant-types client_credentials --scopes read,write,admin` to mint fresh credentials.

A host operator who reaches for `--help` to check the flags is sent to a page that does not mention the command. `--help` is the reflex; `<command>` with no args is not.

Same behavior on `0.42.59.0`, so it is not a recent regression.

### Minor, but adjacent

The hint above passes scopes comma-separated (`--scopes read,write,admin`), which the parser rejects:

```
$ gbrain auth register-client "x" --grant-types client_credentials --scopes read,write
Error: Unknown scope "read,write". Allowed: admin, agent, read, sources_admin, users_admin, write.
```

Space-separated works (`--scopes "read write"`). Either the hint should be corrected or the parser should accept commas — the error message is clear, but the documented invocation is copy-paste-broken.

### Suggested fix

- Route `gbrain auth --help` / `-h` to the same output as bare `gbrain auth`.
- List the `auth` group in the top-level `gbrain --help`.
- Fix the comma-separated `--scopes` example in the `init --mcp-only` hint text (and anywhere else it appears).

abhiramasonny:
## Problem

`gbrain auth --help` prints a generic one-line stub instead of the real usage block. `-h` hits the same stub. `gbrain --help` at the top level never lists `auth` either, even though `init.ts`'s own error hints point operators at `gbrain auth register-client ...` when auth fails.

## Root cause

`src/cli.ts` routes `--help` for any `CLI_ONLY` command through a generic short-circuit (`printCliOnlyHelp`), unless the command is also listed in `CLI_ONLY_SELF_HELP`. `auth` was missing from that second set. `runAuth`'s own `default:` case in `src/commands/auth.ts` already prints a full usage block — `--help` matches no real subcommand name, so it falls through to `default` — but the short-circuit above intercepts the call before `runAuth` ever runs.

`gbrain auth` itself works fine. This is a `--help`-only routing gap, the same class of bug ~15 other `CLI_ONLY` commands (`capture`, `sync`, `schema`, `bootstrap`, ...) hit before, each fixed by adding the command to `CLI_ONLY_SELF_HELP`.

## Fix

- Add `auth` to `CLI_ONLY_SELF_HELP` in `src/cli.ts`, so `--help`/`-h` reaches `runAuth`.
- Add an `auth` entry to the top-level `gbrain --help` output, in the ADMIN section next to `connect` and `serve --http` — the other two commands touching the same token/OAuth surface.

## Tests

Added three cases to `test/cli-help-discoverability.test.ts`, the file already used for ~15 prior fixes of this exact bug class:
- `gbrain auth --help` reaches the real usage block, not the stub
- `gbrain auth -h` does too
- `gbrain --help` lists `auth`

Verified red→green: reverted the `src/cli.ts` change with the tests still in place, confirmed all three failed with the exact pre-fix symptom, restored the fix, confirmed all three pass.

Commands run:
```
bun test test/cli-help-discoverability.test.ts   # 16 pass, 0 fail
bun run verify                                    # 38/38 checks green (incl. typecheck)
bun run test                                      # 486 pass, 10 fail
```

The 10 `bun run test` failures are the same pre-existing PGLite/sources-ops flake I traced on my previous PR (#4081) — identical test names, identical `oom-rescue ... confirmed real` tag, unrelated files (`test/pglite-engine.test.ts`, `test/sources-ops.test.ts`). Reproduces on a clean, unmodified `upstream/master` checkout with none of this branch's changes applied.

## Limitations

Left the third complaint from the original report alone — comma-separated `--scopes` parsing. That one's already fixed by merged PR #3990 and works fine on current master.

Fixes #4003.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -153,6 +153,11 @@ const CLI_ONLY_SELF_HELP = new Set([
   // would leave that help dead code behind the generic stub (the init.ts:117
   // trap ENG-2 names).
   'bootstrap', 'hook', 'sweep',
+  // #4003: auth ships its own detailed usage (token/OAuth-client commands +
+  // flags) in its `default:` switch case, hit whenever the subcommand isn't
+  // one of create/list/revoke/etc — including --help. Without this entry the
+  // generic short-circuit fires first and that usage block is dead code.
+  'auth',
 ]);
 
 /**
@@ -3177,6 +3182,9 @@ ADMIN
     --public-url URL                 Public issuer URL (required behind proxy/tunnel)
   connect <mcp-url> --token <t>      Wire Claude Code to a remote gbrain (bearer token)
         [--install] [--json]         Print the paste-ready command, or --install to run it
+  auth <create|list|revoke|...>      Manage legacy tokens + OAuth 2.1 clients
+  auth --help                        Full subcommand list (register-client,
+                                     rescope-client, revoke-client, permissions, test, ...)
   watch [--json]                     Push-based context: pipe conversation turns in,
                                      volunteered brain pages stream out (#2095)
   call <tool> '<json>'               Raw tool invocation
--- a/src/commands/auth.ts
+++ b/src/commands/auth.ts
@@ -633,34 +633,7 @@ export function parseAuthCreateArgs(rest: string[]): { name: string; takesHolder
   return { name: positional || '', takesHolders };
 }
 
-export async function runAuth(args: string[]): Promise<void> {
-  const [cmd, ...rest] = args;
-  switch (cmd) {
-    case 'create': {
-      // v0.28: optional --takes-holders world,garry,brain (default: world only)
-      const parsed = parseAuthCreateArgs(rest);
-      await create(parsed.name, { takesHolders: parsed.takesHolders });
-      return;
-    }
-    case 'list': await list(); return;
-    case 'revoke': await revoke(rest[0]); return;
-    case 'permissions': {
-      // gbrain auth permissions <name> set-takes-holders world,garry
-      await permissions(rest[0] || '', rest[1] || '', rest[2]);
-      return;
-    }
-    case 'register-client': await registerClient(rest[0], rest.slice(1)); return;
-    case 'rescope-client': await rescopeClient(rest[0], rest.slice(1)); return;
-    case 'revoke-client': await revokeClient(rest[0]); return;
-    case 'test': {
-      const tokenIdx = rest.indexOf('--token');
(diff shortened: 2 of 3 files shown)


time-attack:
Automated triage finding (routed to maintainer, not auto-merged): adding `auth` to `CLI_ONLY_SELF_HELP` changes behavior beyond help text. `runAuth` only prints usage when the FIRST arg is unknown, so post-fix `gbrain auth create foo --help` / `gbrain auth revoke foo --help` skip the generic help short-circuit and actually EXECUTE the subcommand — minting or revoking a real bearer token instead of showing help. Pre-fix those invocations were inert (generic stub). Suggested completion: add a `--help`/`-h` early-return inside `runAuth` (the sync.ts pattern) and extend the tests to cover subcommand+--help. Also note: CI has not run at the current head SHA (workflows sit at action_required awaiting maintainer approval).

abhiramasonny:
Confirmed — `gbrain auth create foo --help` would've minted a real token. Fixed in b51eb951: added an early `--help`/`-h` check in `runAuth` before the subcommand switch (matches `sync.ts`'s existing pattern), checked via `args.includes()` so a trailing flag anywhere gets caught. Added regression tests for create/revoke/register-client plus the `-h` variant.

garrytan:

PR #4092, opened 2026-08-13
fix(core): isPathContained() rejects valid paths on Windows

### LINKED ISSUE #3895: isPathContained() trailing-separator check is backslash/forward-slash mismatched — breaks path containment on Windows
## Summary

`src/core/path-confine.ts:45` (`isPathContained`) builds the
trailing-separator-normalized parent prefix like this:

    const parentWithSep = resolvedParent.endsWith('\') ? resolvedParent : resolvedParent + '/';
    return resolvedChild === resolvedParent || resolvedChild.startsWith(parentWithSep);

It checks for a trailing **backslash** but appends a trailing **forward
slash** when none is found. On Windows, `realpathSync()` returns
backslash-delimited paths (e.g. `C:\Users\me\project`), so:
  - `resolvedParent.endsWith('\')` is false for virtually every real path
    (paths don't end in a separator to begin with),
  - the code appends `/` instead of `\`, producing a mixed-separator
    prefix like `C:\Users\me\project/`,
  - a genuine child path (`C:\Users\me\project\skills`, backslash-delimited)
    never starts with that mixed prefix,
  - so containment incorrectly returns `false` for a normal, non-symlinked
    child directory.

## Confirmed impact

Directly verified via `bun -e` against `isPathContained()`:

    isPathContained('C:/Users/me/project/skills', 'C:/Users/me/project')
    → false   (should be true — plain subdirectory, no symlink involved)

This breaks `autoDetectSkillsDir()`'s tier 1b (`cwd_walk_up`) in
`src/core/repo-root.ts:165`, which gates on this same check. Concretely:
after installing a policy skill via `gbrain integrations install
retrieval-reflex --target <repo>` (which writes `<repo>/skills/
retrieval-reflex/SKILL.md` successfully), `gbrain doctor` still reports:

    "resolver_health": {"status":"warn","message":"Could not find skills directory"}
    "retrieval_reflex_health": {"status":"warn", details: {"policy_skill_installed": false}}

even though the file is present exactly where it's expected, one directory
below `startDir`.

## Scope

`isPathContained` is used in 3 files: `path-confine.ts` (definition),
`core/repo-root.ts` (skills-dir auto-detection — confirmed broken above),
and `core/sources-ops.ts` (source `local_path` containment — not
independently verified broken, but shares the same buggy primitive, worth
checking).

Not affected: `.gbrain-mount`/`.gbrain-source` dotfile trust-gating uses a
separate function, `isTrustedDotfile()` (lstat-based symlink/ownership
check, no path-prefix string comparison), so the brain/source routing fix
from #3576 is unaffected by this bug.

## Environment

- OS: Windows 11
- gbrain: v0.42.73.2 (current origin/master as of this report)
- Path under test: a project folder inside OneDrive
  (`C:\Users\...\OneDrive\Documents\...`), though the bug is a plain
  string-separator mismatch and should reproduce on any Windows path.

## Suggested fix

    const sep = resolvedParent.includes('\\') ? '\\' : '/';
    const parentWithSep = resolvedParent.endsWith(sep) ? resolvedParent : resolvedParent + sep;

or simpler, use `path.sep` (already imported in most of this codebase)
instead of hardcoding `/`.

abhiramasonny:
## Problem

`isPathContained()` incorrectly rejects legitimate subdirectories on Windows, breaking `gbrain doctor`'s skills-dir resolution (`resolver_health: "Could not find skills directory"` even when the file exists exactly where expected), source `local_path` containment checks, and uninstall path-safety gates. Two independent users confirmed this on real Windows installs; a maintainer triage bot reproduced it by inspection and left the issue open.

## Root cause

`isPathContained()` builds a "parent + separator" prefix like this:

```ts
const parentWithSep = resolvedParent.endsWith('/') ? resolvedParent : resolvedParent + '/';
return resolvedChild === resolvedParent || resolvedChild.startsWith(parentWithSep);
```

`resolvedParent` and `resolvedChild` come from `realpathSync()`, which returns native-OS-separator paths — backslash-delimited on Windows. A Windows realpath essentially never ends in a separator to begin with, so `.endsWith('/')` is false there just as often as `.endsWith('\\')` would be. The check always falls through to appending a literal `/`, producing a mixed-separator prefix like `C:\Users\me\project/`. A real backslash-delimited child (`C:\Users\me\project\skills`) never starts with that prefix, so containment fails for every ordinary, non-symlinked Windows subdirectory — not just the sibling-prefix case (`/foo` vs `/foobar`) the trailing-separator logic exists to guard.

`isPathContained()` is the shared confinement primitive behind four call-site families: `repo-root.ts`'s `autoDetectSkillsDir` (the `gbrain doctor` symptom above), `sources-ops.ts`'s `local_path` clone-root containment, `bootstrap/uninstall.ts`'s path-safety gates, and `claude-code-jsonl.ts`'s transcript-path containment. All four silently fail-closed on Windows for perfectly legitimate paths.

## Fix

Same idiom `isWriteTargetContained` already uses in this file, and `isWithinRoot` in `sync.ts` already uses for the analogous `sync` path-safety check (merged in #3415): `path.relative()` instead of a manually-appended separator.

Split the pure comparison out into a new exported `isResolvedContained(resolvedChild, resolvedParent, pathMod?)`, so `isPathContained()` (which does the `realpathSync` + fail-closed error handling) delegates to it. The extra parameter is what makes the Windows case testable — `pathMod` defaults to the real `path` module, but a test can pass `path.win32` to pin Windows separator semantics without touching the filesystem or mocking anything global.

A prior PR (#3284) proposed a similar refactor and was closed as a "no-op on the platforms we ship" — but that PR only demonstrated impact through `sync.ts`'s `isPathSafe`, which #3415 had already fixed independently the same day. It didn't address `isPathContained`'s other three call sites, none of which route through `sync.ts`. Issue #3895's own repro (`gbrain doctor` failing skills-dir resolution) goes through `repo-root.ts`, not `sync.ts` — that's the evidence the earlier "no-op" verdict was missing, and why a maintainer bot re-verified this issue as real 11 days after #3284 closed.

## Tests

`test/path-confine.test.ts`'s existing `isPathContained` describe block (5 tests, all run against the real POSIX filesystem) already passes unchanged — this file's own doc comment for that block already noted it can't catch a Windows-specific bug on POSIX CI.

Added a new `isResolvedContained — Windows separator semantics (#3895)` describe block (6 tests) that calls the new pure helper directly with `path.win32`: real subdir contained, parent-itself contained, sibling directory NOT contained, unrelated directory NOT contained, parent-of-parent NOT contained, and a sanity check that the default `pathMod` (POSIX on this dev box) still matches `isPathContained`'s existing behavior.

Verified red→green: reverted just `src/core/path-confine.ts` with the test file in place — the module failed to load (`isResolvedContained` export missing), confirming the tests actually exercise the fix rather than passing vacuously. Restored the fix, confirmed green.

Also ran the tests for all four call sites (`repo-root.test.ts`, `sources-ops.test.ts`, `claude-code-jsonl.test.ts`, `bootstrap-uninstall.serial.test.ts`) to confirm no behavioral change on POSIX.

Commands run:
```
bun test test/path-confine.test.ts                                          # 41 pass, 0 fail (6 new)
bun test test/repo-root.test.ts test/sources-ops.test.ts                    # 115 pass, 1 pre-existing fail (unrelated)
bun test test/claude-code-jsonl.test.ts test/bootstrap-uninstall.serial.test.ts  # 43 pass, 0 fail
bun run verify                                                              # 38/38 checks green (incl. typecheck)
bun run test                                                                # 17886 pass, 11 fail
```

The `bun run test` failures are the same pre-existing PGLite/sources-ops flake traced on my three previous PRs in this repo (#4081, #4083, #4084), plus one additional transient flake in `test/extract-atoms-chunk-embed.test.ts` — a file that landed in this exact `upstream/master` sync moments before I started this branch, unrelated to path confinement, and not classified "confirmed real" by the test runner's own rescue-retry mechanism (it recovered on retry). None of the 11 touch `path-confine.ts` or its call sites.

## Limitations

No real Windows machine or CI runner available to validate end to end — the regression coverage is a unit-level Windows-semantics test via `path.win32`, not a real Windows filesystem run. The reporter's own suggested fix (a manual `path.sep`-aware check) and this PR's fix produce equivalent results for the cases in the issue; I went with `path.relative()` specifically to match the two already-reviewed precedents in this codebase rather than write new separator-handling logic.

Fixes #3895.

### DIFF
--- a/src/core/path-confine.ts
+++ b/src/core/path-confine.ts
@@ -20,13 +20,46 @@
  */
 
 import { realpathSync, existsSync, type Stats } from 'fs';
-import { resolve as resolvePath, relative, isAbsolute, dirname, basename, join } from 'path';
+import { resolve as resolvePath, relative, isAbsolute, dirname, basename, join, sep } from 'path';
+
+interface RelativePathMod {
+  relative(from: string, to: string): string;
+  isAbsolute(path: string): boolean;
+  sep: string;
+}
+
+const DEFAULT_PATH_MOD: RelativePathMod = { relative, isAbsolute, sep };
 
 /**
- * Symlink-safe path confinement: realpath BOTH sides, then a separator-aware
- * prefix check. A plain `startsWith()` on un-resolved paths would let a
- * `parent/skills` symlink → `/etc` (or `$GBRAIN_HOME/clones/<id>` → `/etc`)
- * bypass the boundary; resolving first defeats that.
+ * Pure `path.relative`-based containment check on two ALREADY-RESOLVED
+ * paths. Split out from `isPathContained` (#3895) so the Windows-separator
+ * case is unit-testable on POSIX CI: pass `path.win32` as `pathMod` to pin
+ * Windows semantics without touching the filesystem or the real `path`
+ * module used everywhere else in this process.
+ *
+ * `resolvedParent.endsWith('/') ? ... : resolvedParent + '/'` (the pre-fix
+ * check) always fell through to appending a literal `/` on Windows, because
+ * `realpathSync` returns native-OS-separator paths and a backslash-delimited
+ * path essentially never ends in `/`. `relative()` resolves through the
+ * path module's own separator instead, so it works on whichever platform
+ * `pathMod` describes. Matches `isWriteTargetContained` below and
+ * `isWithinRoot` in `sync.ts` (#3415) — same idiom, not new logic.
+ */
+export function isResolvedContained(
+  resolvedChild: string,
+  resolvedParent: string,
+  pathMod: RelativePathMod = DEFAULT_PATH_MOD,
+): boolean {
+  if (resolvedChild === resolvedParent) return true;
+  const rel = pathMod.relative(resolvedParent, resolvedChild);
+  return rel !== '' && rel !== '..' && !rel.startsWith('..' + pathMod.sep) && !pathMod.isAbsolute(rel);
+}
+
+/**
+ * Symlink-safe path confinement: realpath BOTH sides, then
+ * `isResolvedContained`. A plain `startsWith()` on un-resolved paths would
+ * let a `parent/skills` symlink → `/etc` (or `$GBRAIN_HOME/clones/<id>` →
+ * `/etc`) bypass the boundary; resolving first defeats that.
  *
  * Returns true iff `child` exists AND its realpath is `parent`'s realpath or a
  * real subtree of it. Returns false if either path is unresolvable (missing /
@@ -41,9 +74,7 @@ export function isPathContained(child: string, parent: string): boolean {
   } catch {
     return false; // missing / unresolvable path → not contained
   }
-  // Append a separator so /foo doesn't match /foobar.
-  const parentWithSep = resolvedParent.endsWith('/') ? resolvedParent : resolvedParent + '/';
-  return resolvedChild === resolvedParent || resolvedChild.startsWith(parentWithSep);
+  return isResolvedContained(resolvedChild, resolvedParent);
 }
 
 /**
--- a/test/path-confine.test.ts
+++ b/test/path-confine.test.ts
@@ -15,10 +15,10 @@ import {
   mkdtempSync, mkdirSync, writeFileSync, rmSync, symlinkSync, chmodSync,
   lstatSync, type Stats,
 } from 'fs';
-import { join } from 'path';
+import { join, win32 as win32Path } from 'path';
 import { tmpdir } from 'os';
 import {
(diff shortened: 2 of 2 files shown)


time-attack:
Superseded: the same isPathContained() Windows-separator fix landed in v0.46.14.0 (wave PR #4226, commit 864dec4f, adopting PR #4103 by @MohammedAlkindi). Master now uses the OS `sep` with CI-runnable win32 shape tests, and issue #3895 is closed. Thanks for the thorough write-up.

Issue #2555, opened 2026-07-03
MCP get_chunks returns [] for pages readable via federated grant (outside the token's floor source)


uppifyagency:
Legacy bearer token federated over `["generale","default","libri","prova"]` with floor source `generale` (permissions.source_id array), v0.42.52.0 @ bb2e88c, remote MCP:

- `get_page("clienti/acme")` (page lives in source `prova`) → **works** (federated read)
- `search` → returns that page's chunk fine
- `get_chunks("clienti/acme")` → **`[]`** (empty array, no error)

Either get_chunks doesn't apply the federated-read grant (scoping bug), or cross-source chunk reads are intentionally floor-only — in which case an explicit `permission_denied` would beat an empty array, which reads as "this page has no chunks".

Same trust-boundary family as #2529 (legacy bearer scoping for takes_holders).

time-attack:
## Severity corrected: p3 → **P1**

This issue was labelled p3 ("cosmetic, docs-only, or a rare configuration"). That was wrong. It was re-assessed by reproducing it against current master.

Reproduced on master 3c61e255 with PGLite: built ctx exactly as http-transport.ts does for a legacy bearer token with permissions.source_id=["generale","default","libri","prova"] (parseLegacyTokenScope → sourceId 'generale' + allowedSources array), page 'clienti/acme' in source 'prova' with 1 chunk. get_page → FOUND (federated read works), resolve_slugs → finds it, get_chunks → [] silently; get_versions → 0, get_raw_data → [] (same class). The [] is a bug, not fail-closed-by-design: the grant explicitly includes 'prova', so the read is authorized — the op just ignores the grant, and the empty array misleads the caller into 'this page has no chunks'. Then applied open PR #3550's diff (applies clean to master) and re-ran: get_chunks → the prova chunk, get_versions → 1, get_raw_data → the row. #3550 also adds tests covering exactly this ('get_chunks returns only in-grant chunks'). Merging #3550 closes this issue. p1 not p3: documented remote read op silently broken for any federated-grant caller for every page outside the floor source, failure invisible, no chunk-level workaround (op takes only slug). Not p0: under-read direction, no disclosure or data loss. No duplicate open issue (#1260 is brain-mount get_page, different axis).

[...]

Issue #2544, opened 2026-07-03
Postgres egress: getChunks selects cc.* (unused embeddings) + per-put_page getAllSlugs full scan


kochiagency-netizen:
## Summary

Running gbrain against a hosted Postgres (Supabase) accumulated ~20 GB egress in a month, tracked to two query patterns in the Postgres path. Both are read-side byte-over-wire issues, not correctness bugs.

## 1. `getChunks` fetches embedding columns it discards

`PostgresEngine.getChunks()` (`src/core/postgres-engine.ts`, ~L2187) uses `SELECT cc.*`. `content_chunks` has three vector columns — `embedding` (1536-d), `embedding_image` (1024-d), `embedding_multimodal` (1024-d), ~18 KB/row combined — that this path never uses: `rowToChunk` is called without `includeEmbedding`, so they're parsed to `null`/dropped. `getChunksWithEmbeddings()` is the separate path that actually needs them.

Fix: select the explicit non-embedding column list. Same applies to the PGLite sibling `PgliteEngine.getChunks()` (`src/core/pglite-engine.ts`, ~L2102) for parity, though PGLite is in-process so it's memory, not egress.

Patch (Postgres):
```sql
SELECT cc.id, cc.page_id, cc.chunk_index, cc.chunk_text, cc.chunk_source,
       cc.model, cc.token_count, cc.embedded_at, cc.language, cc.symbol_name,
       cc.symbol_type, cc.start_line, cc.end_line, cc.parent_symbol_path,
       cc.doc_comment, cc.symbol_name_qualified
FROM content_chunks cc ...
```

## 2. `getAllSlugs` full-table scan once per put_page

`SlugResolver.ensureBasenameIndex()` (`src/core/link-extraction.ts`, ~L905) calls `engine.getAllSlugs()` — `SELECT slug FROM pages` — once per extract run / put_page to build the basename→slug fuzzy-match index. Over a bulk ingest this is N_pages × N_ingested rows. In our brain: ~584 slugs × ~1,972 runs ≈ 1.18 M rows, the single largest egress line in `pg_stat_statements`.

The index is cached within a resolver instance but the resolver is rebuilt per put_page. Options: cache the slug snapshot across an ingest batch, build the basename index once per batch run, or scope the fetch to slugs referenced by the page's links. (This one has wikilink-resolution correctness implications, so it needs more care than #1.)

## Environment
- gbrain v0.42.38.0, Postgres engine (Supabase, `postgres` role via pooler)

[...]

PR #4095, opened 2026-08-14
fix(import,engines): restore code-chunk embedding reuse

TriNguyen1110:
## Problem

The user-visible promise of incremental chunking is the one
`incremental-chunking.test.ts` states: "daily autopilot costs cents not
dollars." Since v0.42.76.0 that promise is silently broken — every code edit
re-embeds the whole file, so users pay for vectors they already own, on every
sync, without any error to notice. Two stacked defects:

**1. #2544 unplugged the cache.** The egress trim removed `cc.embedding` from
`getChunks`' column list — but `importCodeFile`'s incremental-reuse cache reads
`matched.embedding` from exactly that call. Since 2026-08-09 it is always
`undefined`, every chunk misses, and every code chunk re-embeds on any content
change. The comment justifying the trim ("rowToChunk discards embeddings at
this call site") is true for six of the seven callers; this is the seventh.

**2. The cache key was positional.** `${chunk_index}:${chunk_text}` — and
`chunk_text` starts with the `[Lang] path:N-M symbol` header. Any edit that
changes the line count above a symbol shifts both halves while the body stays
byte-identical. Measured on three files in this repo (insert one function at
the top): 0/5, 0/5, 0/8 cache hits, where 4/5, 4/5, 6/8 were recoverable.

Neither defect was catchable by the existing suite: **every test that calls
`importCodeFile` passes `noEmbed`**, so the reuse path had never executed under
test — `incremental-chunking.test.ts` stayed green while the feature it was
written to protect died.

## This PR edits the #2544 structural pin — deliberately, and it gets stricter

`get-page-federated-scope.test.ts` pins `getChunks`' SELECT shape and failed on
this change — working as designed. The pin now enforces a two-way invariant
that is stronger than the old one:

- no `cc.*`, all 14 scalar columns present (unchanged),
- `cc.embedding` may appear **only** behind the `includeEmbedding` conditional
  — any unconditional fetch is the #2544 regression returning,
- the gated opt-in path must **exist**, so a half-revert that strands the flag
  fails too.

`getChunksWithEmbeddings` was considered and rejected: it drops source scoping
when `sourceId` is absent and bypasses `withScopedReadTransaction`, i.e. it
would reopen the federated-scope hole #2555 closed. The opt-in flag keeps the
scope precedence, the `'default'` fallback, and RLS binding identical —
`rowToChunk` already had the `includeEmbedding` parameter waiting.

## Fix

- `engine.ts` + both engines: `getChunks(slug, { includeEmbedding })`, default
  false — the egress trim stands for every other caller.
- `chunkers/code.ts`: `stripChunkHeader()` extracted (the same regex existed
  inline twice, in `buildMergedChunk` and `capOversizedChunks`).
- `embed-reuse.ts` (new, leaf): pure `planEmbeddingReuse()` — buckets stored
  chunks by header-stripped body, FIFO per body so two byte-identical bodies
  never collapse onto one vector.
- `import-file.ts`: the reuse loop consumes the plan; the reused row carries
  the stored `model` stamp forward with its vector.

No schema change, no migration, no `CHUNKER_VERSION` bump, and nothing about
*what gets embedded* changes — existing vectors stay valid.

## Trade-offs, stated plainly

- **Egress:** this one call site fetches vectors again. Justified because these
  vectors are what the cache consumes, and it is once per page at import — a
  miss costs an embedding API call; the fetch costs bytes.
- **Header drift:** a reused vector was computed when its header read `62-86`
  and the row now says `64-88`. Two number tokens out of a few hundred; the
  body — the semantic payload — is byte-identical.
- **token_count** is carried from the matched row, so cost reporting can be off
  by the few tokens a line-number change costs.

## Verification

- New wiring test embeds for real (seeded vectors, provider keys cleared via
  `withEnv`): **fails on master** with the `[gbrain] embedding failed …`
  signature — `embedBatch` reached — and the seeded vectors nulled; passes with
  the fix, vectors byte-identical. It also pins that plain `getChunks` still
  hides vectors, so the #2544 trim is asserted, not assumed.
- 7 unit tests on the pure planner (shifted headers, shifted index, changed
  body, duplicate bodies, null-embedding rows).
- `bun run verify` 38/38 green; targeted suites 106 pass / 0 fail on top of
  v0.45.11.0.

Out of scope, deliberately: the markdown path (`importFromContent` has no reuse
cache at all — a separate change with its own contextual-wrapper questions),
and removing line numbers from the embedded header text (would invalidate every
stored code vector; a future re-embed wave gets it free).

### DIFF
--- a/src/core/chunkers/code.ts
+++ b/src/core/chunkers/code.ts
@@ -808,12 +808,24 @@ function mergeSmallSiblings(chunks: CodeChunk[], chunkTarget: number): CodeChunk
   return merged;
 }
 
+/**
+ * The structured header `buildChunk` prepends: "[Lang] path:N-M symbol\n\n".
+ * It carries line numbers, so it is volatile across edits above a symbol —
+ * `src/core/embed-reuse.ts` keys the embedding-reuse cache on the stripped body.
+ */
+export const CHUNK_HEADER_RE = /^\[[^\]]+\] [^\n]+\n\n/;
+
+/** Chunk text minus its header, or unchanged when there is no header. */
+export function stripChunkHeader(text: string): string {
+  return text.replace(CHUNK_HEADER_RE, '');
+}
+
 function buildMergedChunk(group: CodeChunk[], index: number): CodeChunk {
   const first = group[0]!;
   const last = group[group.length - 1]!;
   // Strip each chunk's structured header line when merging so the combined
   // body reads like the original source. Header is always "[Lang] path:N-M symbol".
-  const bodies = group.map((c) => c.text.replace(/^\[[^\]]+\] [^\n]+\n\n/, ''));
+  const bodies = group.map((c) => stripChunkHeader(c.text));
   const mergedBody = bodies.join('\n\n');
   const header = `[${displayLang(first.metadata.language)}] ${first.metadata.filePath}:${first.metadata.startLine}-${last.metadata.endLine} merged (${group.length} siblings)`;
   return {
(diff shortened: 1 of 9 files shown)


time-attack:

PR #4105, opened 2026-08-14
fix(bootstrap): enforce same-turn memory writeback

garrytan-agents:
## Problem

GBrain context injection is read-side only. A long-lived custom agent workspace could retrieve context on every turn while never persisting newly learned durable facts. Fresh bootstrap templates already described writeback, but upgrades and skillpack scaffolds preserve user-owned `AGENTS.md`, and `bootstrap verify` did not audit that contract. The result was a silent one-way memory loop: useful context today, stale context tomorrow.

## Error Log

This was a missing invariant rather than a thrown runtime exception. The structural reproduction on a legacy workspace was:

```json
{
  "ok": false,
  "agentsPath": "/tmp/example-workspace/AGENTS.md",
  "mode": "missing",
  "detail": "same-turn write-back contract missing; context can be injected without durable capture"
}
```

Before this change, `gbrain bootstrap verify` had no check that could surface this condition.

## What We Tried

1. Audited the context engine. It assembles retrieval context but intentionally does not author durable facts.
2. Audited conversation fact extraction. The batch extractor is useful for backfills, but it is opt-in and does not guarantee same-turn persistence.
3. Audited skillpack scaffold behavior. Scaffolds correctly refuse to overwrite user-owned `AGENTS.md`, so silently mutating that file during install would violate the ownership model.
4. Kept the ownership invariant and added an explicit, additive repair path with backup and idempotency.

## Solution

- Add a shared onboarding contract that makes the read, detect, write, verify memory loop explicit for every installed skill bundle.
- Teach fresh bootstrap templates to use `remember` with provenance for atomic facts and structured brain operations for richer knowledge.
- Add `gbrain bootstrap contract` to audit existing workspaces.
- Add `gbrain bootstrap contract --repair` to append a marker-owned writeback block only after backing up the original `AGENTS.md`.
- Make `gbrain bootstrap verify` fail when the same-turn writeback contract is absent and point to the repair command.
- Refuse repair for symlinked `AGENTS.md` and keep repeated repairs idempotent.

| Workspace state | Audit | Repair |
|---|---|---|
| Fresh integrated gate | Pass | No-op |
| Previously repaired marker block | Pass | No-op |
| Semantically equivalent custom gate | Pass | No-op |
| Legacy one-way contract | Fail | Backup plus additive block |
| Missing `AGENTS.md` | Fail | Refuse; render first |
| Symlinked `AGENTS.md` | Fail | Refuse without mutation |

## Results

Before: retrieval could be healthy while durable capture was absent, and verification still passed.

After: bootstrap verification treats writeback as a first-class contract; existing workspaces get an explicit migration, while fresh installs and default skillpacks carry the always-on memory loop.

## Testing

- 115 tests passed across bootstrap contract, render, verify, dispatcher, attach, Codex door, scaffold, and end-to-end skillpack flows.
- `bun run typecheck`
- `bash scripts/check-bootstrap-templates.sh`
- Public diff and PR metadata passed the PII scanner.

### DIFF
--- a/skills/_AGENT_README.md
+++ b/skills/_AGENT_README.md
@@ -50,6 +50,25 @@ carries the disambiguation rules for overlapping matches. If the two disagree,
 frontmatter wins. (There is no machine-managed block inside `RESOLVER.md` or
 `AGENTS.md`; that pattern was retired.)
 
+## The always-on memory loop
+
+Routing a requested skill is only half the job. On every inbound message,
+apply the bundled `signal-detector` contract directly (or through a host-native
+background lane when the user has authorized delegation), then close the turn
+with same-turn write-back:
+
+1. Read relevant entity/project context before answering.
+2. Detect durable new facts, preferences, decisions, commitments, relationships,
+   and original thinking in the user's message and in the resulting discussion.
+3. Write atomic facts with GBrain's `remember` verb and explicit provenance.
+   Route richer knowledge through `brain-ops` (`put_page`, timeline, links).
+4. Verify with `recall`, `entity`, or `get_page` before claiming the write landed.
+
+Context injection is read-side automation; it does not, by itself, capture what
+the agent just learned. A turn that learned something durable and recorded
+nothing is incomplete. Skip only transient logistics, acknowledgments, and facts
+already verified as present.
+
 ## When the user invokes a skill
 
 Read the entire `skills/<slug>/SKILL.md` file. Follow its `## Phases`,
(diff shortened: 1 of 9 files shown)


garrytan:
Closing the loop: this landed via wave-L (v0.47.3.0, #4663) with explicit credit — the CHANGELOG entry reads "(#4105 by @garrytan-agents)". Master carries `gbrain bootstrap contract [--repair]`, the `writeback_contract` check in `src/core/bootstrap/verify.ts`, `src/core/bootstrap/contract.ts`, and templates that route atomic facts through `remember` with explicit provenance. The "silent one-way memory loop" framing in this PR is what made the invariant legible enough to adopt — thank you. Closing as shipped.

PR #4109, opened 2026-08-14
fix: clarify source-boundary mutations

jared-voss:
## Summary

- preserve same-source graph mutation isolation for remote MCP clients
- return endpoint-specific `permission_denied` or `page_not_found` diagnostics when `add_link` / `add_timeline_entry` targets a page outside the caller's write source
- keep source probing limited to the caller's allowed read scope and avoid disclosing soft-deleted foreign pages
- make link and timeline endpoint resolution atomic with insertion so concurrent hard deletes cannot report false success
- lock referenced Postgres page rows with `FOR KEY SHARE` so concurrent hard deletes cannot leak raw FK violations
- make Postgres and PGlite `addLink` misses identify `from` vs `to`
- release the behavior as `0.45.15.0`, rebased on upstream `0.45.14.0`

## Why

Federated `get_page` can read a page in another granted source, but graph mutations intentionally use the caller's scalar write source. The old engine errors reported the page as missing and surfaced as `internal_error`, which made intentional source isolation look like data loss.

## Verification

- `bun test test/source-boundary-mutation-errors.test.ts test/source-id-tx-regression.test.ts test/source-scope-resolver.test.ts test/operations-fuzzy-source-scope.test.ts test/mcp-dispatch-optional-params.test.ts test/sync-delta.test.ts test/openclaw-plugin-manifest.test.ts` — 94 passed / 491 assertions
- `DATABASE_URL=... bun test test/e2e/source-boundary-mutation-postgres.test.ts` — 2 real-Postgres deletion-race tests passed
- `bun run typecheck`
- `bun run build:llms`
- `bash scripts/check-key-files-current-state.sh`
- `bun run check:bootstrap-tag`
- `bun run check:bootstrap-templates`
- `bun test test/openclaw-plugin-manifest.test.ts`
- `git diff --check`

Repository-level `bun run verify` completed 37/39 checks locally. The two failures were compiled-artifact environment checks (`check:wasm` terminated with 137; `check:pglite-embedded` produced empty output). Hosted CI will rerun the complete upstream matrix on the rebased head.

## Compatibility

- no schema migration
- same-source writes remain unchanged
- same-source soft-deleted graph endpoints retain existing engine behavior
- hidden and deleted foreign pages remain indistinguishable from absence


time-attack:

garrytan:

PR #4111, opened 2026-08-14
fix(ai): claude-cli scrubs cloud-auth routing env vars, not just the three ANTHROPIC_* keys

Masashi-Ono0611:
## What

`src/core/ai/providers/claude-cli-language-model.ts` spawns `claude --print` as a subprocess for this recipe's language-model adapter. It scrubs the child env of `ANTHROPIC_API_KEY` / `ANTHROPIC_AUTH_TOKEN` / `ANTHROPIC_BASE_URL` before spawning, specifically to guarantee the subprocess authenticates via its own subscription OAuth session rather than an inherited API key (`test/claude-cli-recipe.test.ts` pins this as "subscription-only auth").

That scrub doesn't cover Claude Code's cloud-backend switch flags. Claude Code supports routing through Amazon Bedrock, Google Cloud's Agent Platform (Vertex AI), the Bedrock Mantle endpoint, Microsoft Foundry, and Claude Platform on AWS — each gated by its own `CLAUDE_CODE_USE_*` environment variable, all of which take priority over subscription OAuth when set. If any of these flags is present in the parent process's environment (e.g. inherited from a shell profile, CI runner, or another tool in the same environment), it survives the existing scrub and the subprocess silently routes through a paid cloud backend instead of the subscription auth this recipe is built around — the same class of gap the existing `ANTHROPIC_API_KEY` scrub exists to close, just on a different set of variables.

Per the official docs, each backend requires its switch flag explicitly:
- Bedrock — `CLAUDE_CODE_USE_BEDROCK` ([amazon-bedrock](https://code.claude.com/docs/en/amazon-bedrock))
- Vertex AI / Google Cloud's Agent Platform — `CLAUDE_CODE_USE_VERTEX` ([google-vertex-ai](https://code.claude.com/docs/en/google-vertex-ai))
- Bedrock Mantle endpoint — `CLAUDE_CODE_USE_MANTLE` ([amazon-bedrock](https://code.claude.com/docs/en/amazon-bedrock))
- Microsoft Foundry — `CLAUDE_CODE_USE_FOUNDRY` ([microsoft-foundry](https://code.claude.com/docs/en/microsoft-foundry))
- Claude Platform on AWS — `CLAUDE_CODE_USE_ANTHROPIC_AWS` ([claude-platform-on-aws](https://code.claude.com/docs/en/claude-platform-on-aws), which states explicitly: *"Claude Platform on AWS is opt-in even when AWS credentials are present in your environment"*)

Since every backend is opt-in behind its switch flag, clearing the five flags is sufficient — the provider-specific credential vars (`AWS_*`, `ANTHROPIC_VERTEX_PROJECT_ID`, `ANTHROPIC_FOUNDRY_RESOURCE`, `ANTHROPIC_AWS_WORKSPACE_ID`, `ANTHROPIC_BEDROCK_BASE_URL`, etc.) are inert without their flag regardless of presence. This PR does not touch those credential vars, to keep the diff to the minimal set that closes the gap.

## Test

`test/claude-cli-recipe.test.ts`'s existing "subscription-only auth" test now also sets all five `CLAUDE_CODE_USE_*` flags in the parent env before invoking the model, and asserts all eight scrubbed vars come back `UNSET` in the spawned child's env (verified via a stub `claude` binary that dumps its own env to a file). Confirmed red→green: reverting only the source change against this test fails (`1 fail`); with the fix, `bun test test/claude-cli-recipe.test.ts` passes (`20 pass`, `0 fail`). `bun run typecheck` is clean.

## Note

Gap identified during a Codex-assisted security review of our own subscription-launch wrappers (2026-08-15).

### DIFF
--- a/src/core/ai/providers/claude-cli-language-model.ts
+++ b/src/core/ai/providers/claude-cli-language-model.ts
@@ -219,10 +219,22 @@ function runClaude(
     // (subscription), never via an inherited API key. Without this, an
     // ANTHROPIC_API_KEY in gbrain's env (the exact setup this recipe is meant
     // to replace) silently flips billing to per-token API usage.
+    //
+    // Also scrub the CLAUDE_CODE_USE_* backend-switch flags: Bedrock, Vertex
+    // AI, Mantle, Microsoft Foundry, and Claude Platform on AWS are each
+    // gated by one of these, take priority over subscription OAuth when set,
+    // and route billing through a cloud account instead. Clearing the switch
+    // is sufficient — provider-specific creds (AWS_*, ANTHROPIC_VERTEX_*,
+    // ANTHROPIC_FOUNDRY_*, ANTHROPIC_AWS_*, ...) are inert without it.
     const env = { ...process.env };
     delete env.ANTHROPIC_API_KEY;
     delete env.ANTHROPIC_AUTH_TOKEN;
     delete env.ANTHROPIC_BASE_URL;
+    delete env.CLAUDE_CODE_USE_BEDROCK;
+    delete env.CLAUDE_CODE_USE_VERTEX;
+    delete env.CLAUDE_CODE_USE_MANTLE;
+    delete env.CLAUDE_CODE_USE_FOUNDRY;
+    delete env.CLAUDE_CODE_USE_ANTHROPIC_AWS;
     const child = spawn(claudeBin(), args, {
       stdio: ['pipe', 'pipe', 'pipe'],
       cwd: ensureCleanCwd(),
--- a/test/claude-cli-recipe.test.ts
+++ b/test/claude-cli-recipe.test.ts
@@ -377,19 +377,24 @@ describe('claude-cli LanguageModel — context isolation', () => {
     });
   });
 
-  test('scrubs ANTHROPIC_* credentials from the child env (subscription-only auth)', async () => {
+  test('scrubs ANTHROPIC_* credentials and cloud-auth backend switches from the child env (subscription-only auth)', async () => {
     await withStubEnv(async () => {
       await withEnv(
         {
           ANTHROPIC_API_KEY: 'sk-should-never-leak',
           ANTHROPIC_AUTH_TOKEN: 'tok-should-never-leak',
           ANTHROPIC_BASE_URL: 'https://proxy.should.never.leak',
+          CLAUDE_CODE_USE_BEDROCK: '1',
+          CLAUDE_CODE_USE_VERTEX: '1',
+          CLAUDE_CODE_USE_MANTLE: '1',
+          CLAUDE_CODE_USE_FOUNDRY: '1',
+          CLAUDE_CODE_USE_ANTHROPIC_AWS: '1',
         },
         async () => {
           const envLog = join(stubDir, 'env.log');
           const envStub = [
             '#!/bin/sh',
-            `printf "key=%s\\ntoken=%s\\nbase=%s\\n" "\${ANTHROPIC_API_KEY:-UNSET}" "\${ANTHROPIC_AUTH_TOKEN:-UNSET}" "\${ANTHROPIC_BASE_URL:-UNSET}" > "${envLog}"`,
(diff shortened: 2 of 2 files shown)


time-attack:
Automated triage: floor gate passed — CI green at head, targeted tests pass, `bun run verify` clean in an isolated worktree at the pinned SHA. However, this PR modifies `src/core/ai/providers/claude-cli-language-model.ts`, which triggers the mandatory live-roundtrip tier (a real chat() call through the modified provider plus the legacy path — unit-shape assertions do not satisfy it). That tier could not be executed in this environment: driving the real CLI path against an isolated GBRAIN_HOME, the model-reachability probe resolves `chat` to `anthropic:claude-sonnet-4-6` even with `chat_model` explicitly set to `claude-cli:claude-sonnet-4-6` (the DB-plane value shadows the file plane at runtime), and no ANTHROPIC_API_KEY is present in this sandbox. A skipped required tier is treated as a block rather than a silent pass, so this is routed to human review. Nothing negative was found — the roundtrip simply was not runnable here.

Masashi-Ono0611:
Live-roundtrip data point from a production brain (same environment as the note on #4120): this PR's diff has been applied to the installed v0.45.11.0 tree since 2026-08-15T01:31Z, with `chat` = `claude-cli:claude-sonnet-5` on subscription session auth. Since 2026-08-15T03:14Z, 54 real `chat()` calls have gone through the modified provider, all `request_status = 'succeeded'` in `chat_usage_log` — the extended scrub set does not break the normal auth path.

Stated caveat: this environment does not set any of the newly-scrubbed cloud-routing vars (`CLAUDE_CODE_USE_BEDROCK` / `VERTEX` / …), so these calls exercise the scrub as a no-op on absent vars. They satisfy the "real chat() call through the modified provider" tier, but they do not demonstrate the hijack-prevention branch itself. Happy to run a forced roundtrip with `CLAUDE_CODE_USE_BEDROCK=1` set in the parent env if that would be useful.

Masashi-Ono0611:
Follow-up on the offer in my previous comment — I ran the forced-hijack contrast, and it demonstrates the prevention branch directly.

Same parent environment for both runs: `CLAUDE_CODE_USE_BEDROCK=1` exported, on the machine described in my previous comment (patched provider in the installed v0.45.11.0 tree).

1. **Raw `claude` CLI (no scrub), `claude -p "Say OK" --model claude-sonnet-5`:** fails with `API Error: Could not load credentials from any providers` — the AWS credential-chain error, i.e. the env var is live and reroutes the raw CLI to Bedrock.
2. **`chat()` through this PR's provider, same parent env:** the spawned subprocess does NOT attempt Bedrock — it proceeds down the normal claude-auth path. In this particular test context (a non-interactive SSH session) it then stopped at `Not logged in · Please run /login`, because the CLI's OAuth credentials live in the macOS keychain, which is unreachable from a non-interactive SSH session. That stop is an artifact of the test transport, not of the provider: the same patched provider completed 54/54 successful production calls (subscription auth) since the patch was applied on 2026-08-15, per the `chat_usage_log` data in my previous comment.

So the contrast is: unscrubbed → Bedrock credential-chain error; scrubbed → normal claude auth route. That is the hijack-prevention behavior this PR extends, observed rather than inferred.

garrytan:
Folding this into the current fix wave with credit (adopted at the pinned patch; the subscription-only contract gets a CHANGELOG callout). Will close this PR when the wave lands.

Masashi-Ono0611:
Closing — this landed already, in a hardened form. Commit \`864dec4f\` (v0.46.14.0 fix-wave, #4226) explicitly credits this PR: "Adopted from PR #4111 by @Masashi-Ono0611 (pinned patch)... the existing scrub already deleted ANTHROPIC_API_KEY/AUTH_TOKEN/BASE_URL for exactly this reason." The maintainer went further than my original diff — a prefix-wipe loop over all \`CLAUDE_CODE_USE_*\` keys instead of five individual deletes, so future flags in that family are covered automatically. Confirmed while rebasing this branch: the resulting tree is byte-identical to \`upstream/master\`. No further action needed on this PR.

PR #4112, opened 2026-08-14
fix(models): doctor chat probe honors slow-start providers instead of a hardcoded 5s abort

Masashi-Ono0611:
## Symptom

`gbrain models doctor` reports the `claude-cli:` chat touchpoint as `unknown` on essentially every run, even on a fully healthy CLI + subscription:

```
✘ chat              claude-cli:claude-sonnet-5     unknown (5004ms)
    claude-cli adapter aborted
```

Real-world observation: doctor reported `unknown@5004ms` and the *very next* real call through the same provider (a normal chat/agent turn) succeeded fine. The probe was never actually testing reachability — it was testing whether the `claude --print` subprocess could start, authenticate, and return in under 5 seconds, which it routinely doesn't (CLI process start + loading user-level `CLAUDE.md` takes ~5-6s even when nothing is wrong).

## Root cause

`probeModel()` in `src/commands/models.ts` hardcoded a flat `5000` for its `AbortController` timeout, for every provider, with no way for a slow-start provider to declare it needs more headroom.

## Fix

Same shape as the existing reranker probe's recipe-default timeout fallback (`resolveLiveRerankerTimeoutMs` / `src/core/search/mode.ts`, already on master): a recipe can declare `touchpoints.<chat|expansion>.default_timeout_ms`, and the probe uses it instead of the flat default.

- `src/core/ai/types.ts`: added `default_timeout_ms?: number` to `ChatTouchpoint` and `ExpansionTouchpoint` (mirrors the field already on `RerankerTouchpoint`).
- `src/core/ai/recipes/claude-cli.ts`: declared `default_timeout_ms: 30_000` on its `chat` touchpoint (30s comfortably covers the observed 5-6s cold start without masking a genuinely dead/unauthenticated CLI for anywhere near that long; claude-cli has no expansion touchpoint, so nothing else changes).
- `src/commands/models.ts`: new exported `resolveChatProbeTimeoutMs(modelStr, touchpoint)` reads `recipe.touchpoints[touchpoint]?.default_timeout_ms ?? 5000` (fail-open to 5000 on any resolution error, same shape as the reranker helper). `probeModel()` calls it once and uses the resolved value for both the abort timer and the abort-error message, instead of the literal `5000`.

No config-key override exists for chat/expansion timeouts (unlike `search.reranker.timeout_ms`), so the precedence chain here is intentionally simpler than the reranker one: recipe declaration, else the unchanged historical 5000ms. Every provider other than `claude-cli` is unaffected — same 5000ms as before.

Scope is deliberately narrow to this one probe. `probeEmbeddingReachability` (same file, same hardcoded-5000ms pattern) is left untouched — no embedding touchpoint currently declares this kind of slow-start behavior, and this PR is a single logical change.

## Tests

- `test/models-doctor-chat-probe-timeout.test.ts` (fast, zero-wait): pins `resolveChatProbeTimeoutMs` — claude-cli → 30000, every other provider/touchpoint → unchanged 5000, unresolvable model string → fails open to 5000.
- `test/models-doctor-chat-probe-timeout.slow.test.ts` (per `docs/TESTING.md`'s `*.slow.test.ts` convention, since these use real multi-second waits): end-to-end proof via injected `deps.chat` mocks that stall ~5.5s — claude-cli now succeeds past the old 5s cliff, every other provider is provably unchanged (aborts at exactly 5000ms, exact message pinned).

`bun run typecheck` and `bun run verify` (39/39 checks) both pass; targeted + `bun run test:slow` all green.

### DIFF
--- a/src/commands/models.ts
+++ b/src/commands/models.ts
@@ -564,7 +564,7 @@ export async function probeEmbeddingReachability(deps: ProbeDeps = {}): Promise<
  *
  * Pre-fix `probeModel` hardcoded 5000ms for every provider. That's fine for
  * a plain network round-trip, but `claude-cli:` dispatches through a
- * `claude --print` subprocess (CLI cold start + user-level CLAUDE.md load),
+ * `claude -p (print mode)` subprocess (CLI cold start + user-level CLAUDE.md load),
  * which routinely takes 5-6s even when healthy — so the probe aborted on
  * every run and reported 'unknown — claude-cli adapter aborted', not
  * because the model was actually unreachable. Mirrors the reranker probe's
--- a/src/commands/models.ts
+++ b/src/commands/models.ts
@@ -557,13 +557,40 @@ export async function probeEmbeddingReachability(deps: ProbeDeps = {}): Promise<
   }
 }
 
+/**
+ * Resolve the chat/expansion probe timeout: the recipe's declared
+ * `touchpoints.<kind>.default_timeout_ms` when set, else the probe's
+ * historical flat 5000ms.
+ *
+ * Pre-fix `probeModel` hardcoded 5000ms for every provider. That's fine for
+ * a plain network round-trip, but `claude-cli:` dispatches through a
+ * `claude --print` subprocess (CLI cold start + user-level CLAUDE.md load),
+ * which routinely takes 5-6s even when healthy — so the probe aborted on
+ * every run and reported 'unknown — claude-cli adapter aborted', not
+ * because the model was actually unreachable. Mirrors the reranker probe's
+ * recipe-default fallback (`resolveLiveRerankerTimeoutMs` / mode.ts), but
+ * simpler: unlike `search.reranker.timeout_ms`, there's no config-key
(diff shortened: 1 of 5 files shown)


time-attack:

garrytan:
Folding this into the current fix wave with credit (adopted verbatim at the pinned patch). Will close this PR when the wave lands.

Masashi-Ono0611:
Closing — this landed already. Commit \`08f2a050\` ("fix(models): doctor chat probe honors slow-start providers instead of a hardcoded 5s abort") shipped in the v0.46.14.0 fix-wave (#4226), explicitly credited: "Adopted from PR #4112 by @Masashi-Ono0611 (patch sha256 pinned pre-assembly)... Co-authored-by: Masashi-Ono0611". CHANGELOG.md also credits it by name. Confirmed \`default_timeout_ms\` logic and both test files are identical between this branch and current \`upstream/master\` while attempting a rebase today (git even dropped both commits as empty patches). No further action needed on this PR.

PR #4120, opened 2026-08-15
fix(ai): claude-cli reports cachedInputTokens so cache reads stop counting as zero

Masashi-Ono0611:
## Symptom

Every `claude-cli`-routed gateway call reports `cache_read_tokens: 0`, regardless of
actual prompt-cache reads. Sampled 20/20 `claude-cli` gateway calls in one usage window
show `input_tokens=2` (min=max=2, i.e. structurally near-zero) and `cache_read_tokens=0`
for the whole sample, while a parallel `openai:gpt-5.6` recipe in the same window shows
avg `input_tokens=27,408` and a cumulative `cache_read_tokens` of 1,844,509 — consistent
with the cache-read count being zeroed out structurally rather than genuinely absent.

## Mechanism

`doGenerate()` in `src/core/ai/providers/claude-cli-language-model.ts` builds the
returned `LanguageModelV2` usage object from `inputTokens`/`outputTokens`/`totalTokens`
only. It never mapped `result.usage?.cache_read_input_tokens` — a field the
`ClaudeJsonResult` type already declares, and which `claude --print --output-format json`
already returns — onto the AI SDK's `cachedInputTokens` field.

Downstream, `src/core/ai/gateway.ts:3455` builds the reported `cache_read_tokens` as:

```ts
cache_read_tokens: Number(anthropicCache.cacheReadInputTokens ?? anthropicCache.cache_read_input_tokens ?? usage.cachedInputTokens ?? 0),
```

`claude-cli` never returns `providerMetadata`, so `anthropicCache` is always `{}`. With
`cachedInputTokens` also always `undefined` (this bug), the chain always fell through to
the literal `0` — for every call, permanently, regardless of real cache behavior.

## Fix

Map `result.usage?.cache_read_input_tokens` onto `cachedInputTokens` in the returned
usage object (`Number()`-coerced), leaving it `undefined` — not `0` — when the CLI omits
the field, so `gateway.ts`'s own `?? 0` normalization stays the single place that decides
the omitted-vs-zero default.

**Scope note — `cache_creation_input_tokens` intentionally untouched.** `LanguageModelV2Usage`
has no corresponding field for cache-creation tokens, and wiring it through would require a
`claude-cli`-specific branch in `gateway.ts`'s usage assembly (a separate, larger design
decision). Keeping this PR to the one already-supported field it can surface cleanly.

## Tests

Added to `test/claude-cli-recipe.test.ts` using the file's existing `GBRAIN_CLAUDE_CLI_BIN`
shell-stub pattern: present-zero, present-nonzero (pins the fix), and omitted/undefined.

**Red-green, both directions verified locally:**
- Fix removed (temporarily reverted just the source file, tests kept): `bun test
  test/claude-cli-recipe.test.ts` → 20 pass / 2 fail. Both failures are on
  `cachedInputTokens` assertions: the present-zero assertion in the existing "returns a
  single text content block with usage + stop finish reason" test (`Expected: 0,
  Received: undefined`), and the new "maps cache_read_input_tokens onto
  usage.cachedInputTokens" test (`Expected: 9001, Received: undefined`). The
  omitted/undefined test passes either way, as expected — it pins the deliberate
  `undefined` semantics, not the defect itself.
- Fix restored: 22 pass / 0 fail (62 `expect()` calls).
- `bun run typecheck` (`tsc --noEmit`): exit 0.

## Draft

Opening as Draft — happy to switch to Ready once maintainer bandwidth allows a look;
no urgency implied.

### DIFF
--- a/src/core/ai/providers/claude-cli-language-model.ts
+++ b/src/core/ai/providers/claude-cli-language-model.ts
@@ -392,7 +392,12 @@ export class ClaudeCliLanguageModel implements LanguageModelV2 {
   async doGenerate(options: LanguageModelV2CallOptions): Promise<{
     content: LanguageModelV2Content[];
     finishReason: 'stop' | 'length' | 'content-filter' | 'tool-calls' | 'error' | 'other' | 'unknown';
-    usage: { inputTokens: number | undefined; outputTokens: number | undefined; totalTokens: number | undefined };
+    usage: {
+      inputTokens: number | undefined;
+      outputTokens: number | undefined;
+      totalTokens: number | undefined;
+      cachedInputTokens: number | undefined;
+    };
     warnings: never[];
   }> {
     const { systemText, userPrompt } = renderPrompt(options.prompt);
@@ -422,6 +427,14 @@ export class ClaudeCliLanguageModel implements LanguageModelV2 {
     const inputTokens = result.usage?.input_tokens;
     const outputTokens = result.usage?.output_tokens;
     const totalTokens = (inputTokens ?? 0) + (outputTokens ?? 0);
+    // `cache_creation_input_tokens` is deliberately NOT surfaced here — the AI
+    // SDK's LanguageModelV2Usage has no corresponding field, and folding it in
+    // would need a claude-cli-specific branch in the gateway's usage assembly
+    // (src/core/ai/gateway.ts). Out of scope for this fix.
+    const cachedInputTokens =
+      result.usage?.cache_read_input_tokens !== undefined
+        ? Number(result.usage.cache_read_input_tokens)
+        : undefined;
 
     return {
       content,
@@ -430,6 +443,7 @@ export class ClaudeCliLanguageModel implements LanguageModelV2 {
         inputTokens,
         outputTokens,
         totalTokens: inputTokens !== undefined && outputTokens !== undefined ? totalTokens : undefined,
+        cachedInputTokens,
       },
       warnings: [],
     };
--- a/test/claude-cli-recipe.test.ts
+++ b/test/claude-cli-recipe.test.ts
@@ -120,6 +120,54 @@ describe('claude-cli LanguageModel — text-only round trip', () => {
       expect(result.content[0]).toEqual({ type: 'text', text: 'hello world' });
       expect(result.usage.inputTokens).toBe(12);
       expect(result.usage.outputTokens).toBe(34);
+      // baseEnvelope's default cache_read_input_tokens is 0 (present, not
+      // omitted) — pins "present zero" as distinct from the omitted/undefined
+      // case covered below.
+      expect(result.usage.cachedInputTokens).toBe(0);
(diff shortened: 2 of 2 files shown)


time-attack:
Automated triage: floor gate passed — CI green at head, targeted tests pass, `bun run verify` clean in an isolated worktree at the pinned SHA. However, this PR modifies `src/core/ai/providers/claude-cli-language-model.ts`, which triggers the mandatory live-roundtrip tier (a real chat() call through the modified provider plus the legacy path — unit-shape assertions do not satisfy it). That tier could not be executed in this environment: driving the real CLI path against an isolated GBRAIN_HOME, the model-reachability probe resolves `chat` to `anthropic:claude-sonnet-4-6` even with `chat_model` explicitly set to `claude-cli:claude-sonnet-4-6` (the DB-plane value shadows the file plane at runtime), and no ANTHROPIC_API_KEY is present in this sandbox. A skipped required tier is treated as a block rather than a silent pass, so this is routed to human review. Nothing negative was found — the roundtrip simply was not runnable here.

Masashi-Ono0611:
Live-roundtrip data from a production brain — the tier the automated triage reported it could not run in its sandbox.

Setup: real brain (Postgres engine), `chat` routed through `claude-cli:claude-sonnet-5` on subscription session auth. This PR's diff was applied onto the installed v0.45.11.0 tree at 2026-08-15T03:13:59Z (file mtime of the patched provider).

Measured directly from `chat_usage_log` (SQL over the live DB, `provider_id LIKE '%claude-cli%'`):

- Before the patch: 19 succeeded claude-cli calls (first claude-cli row ever: 2026-08-14T16:42Z), every one with `cache_read_tokens = 0`.
- First call after the patch, 12 seconds later (2026-08-15T03:14:11Z): `cache_read_tokens = 3289`.
- Since then: 54 calls, all `request_status = 'succeeded'`, 39 of them with nonzero cache reads (up to 8,757 tokens on recent rows, e.g. `2026-08-15T15:51:45Z  model=claude-cli:claude-sonnet-5  input=2  output=4  cache_read=8757`). 15 rows still record 0 — I have not root-caused those individually, so I can only say the field is no longer structurally zero, not that every zero is a true miss.

So the `cache_read_input_tokens → cachedInputTokens` mapping is exercised end-to-end through the real CLI path — real `chat()` calls through the modified provider — and the usage rows flip from structurally-zero to recording real cache reads, with no auth or request regressions observed.

garrytan:

Masashi-Ono0611:
Closing — this landed on master already. Two commits carry this exact change (\`1803a0ff\`, authored by me, and \`3421d7ce\` by @garrytan, both titled "claude-cli reports cachedInputTokens so cache reads stop counting as zero") — confirmed present and functioning in v0.46.19.0's \`claude-cli-language-model.ts\` while rebasing a local patch-stack through it today. No further action needed on this PR.

PR #4122, opened 2026-08-15
fix(dream): --input on an already-synthesized transcript says why it skipped instead of reporting clean

Masashi-Ono0611:
## Summary

`gbrain dream --input <file>` (which implies `--phase synthesize`) prints only:

```
Brain is healthy. 1 phase(s) checked in 0.2s.
```

and exits, with zero indication of what happened, whenever the synthesize phase
does no *new* work on an already-handled transcript. Confirmed on master: this
happens both when (a) the transcript's content hash already has a completed job
under the v2 idempotency key — `MinionQueue.add()`'s pre-insert
`SELECT ... WHERE idempotency_key = $1` fast path (the `ON CONFLICT (idempotency_key)
... DO NOTHING` branch only fires for a concurrent-submit race that slips past that
SELECT) transparently returns the pre-existing completed job, no new pages — and
when (b) the D8 legacy-key-family completion check in `runPhaseSynthesize`
(`src/core/cycle/synthesize.ts`) determines every worth-processing transcript was
already synthesized under the pre-v2 key shape (reason
`already_synthesized_legacy_chunked` / `already_synthesized_legacy_single_chunk`,
recorded in `details.skips`).

Note: the significance-verdict lookup (`dream_verdicts`, keyed on full file path +
content hash) runs *before* the job-queue idempotency check (keyed on source +
basename + content hash) in `runPhaseSynthesize`. So "no LLM call" in (a) is
narrower than it first looks: it holds when the same path is re-run unchanged *and*
the prior verdict was cached. Two cases still call the judge before the queue ever
dedupes — re-running the same content from a different path (verdict key misses),
and a prior verdict that came back `unreliable` (deliberately not cached, so it is
re-judged). In all of them the job-queue reuse itself remains LLM-free; it is the
step in front of it that can spend.

Two root causes, both purely in status-derivation / display code — this PR does
**not** touch job-queue idempotency semantics or synthesize's skip logic:

1. **`deriveStatus()`** (`src/core/cycle.ts`) computes `anyWork` from `totals.*`
   but never looked at `totals.transcripts_processed` / `totals.synth_pages_written`.
   A synthesize-only cycle (which is exactly what `--input` always runs) never
   touches sync/embed/lint/etc, so those other totals stay zero even when
   synthesize genuinely processed a transcript or wrote pages — forcing
   `status: 'clean'` regardless of what synthesize actually did.
2. **`printHuman()`** (`src/commands/dream.ts`) short-circuits on
   `status === 'clean'` and prints only the "Brain is healthy" line — it never
   inspects `report.phases[].details.skips`, which is exactly where
   synthesize's D8/D5 skip reasons already live (already computed and returned,
   just never surfaced to the terminal).

## Fix

- `deriveStatus()`: added `totals.transcripts_processed > 0 || totals.synth_pages_written > 0`
  to the existing `anyWork` OR-chain — a real synthesize outcome now flips
  `status` to `'ok'` instead of `'clean'`.
- `printHuman()`: in the `status === 'clean'` branch, before printing the healthy
  line, collect any `details.skips` entries across all phases and print them as
  `  - <phase>: <filePath> (<reason>)` lines. A cycle that skipped a transcript
  for a real reason (D8/D5) but did zero new work still says *why*, instead of
  just "Brain is healthy."
- Both files get a `__testing` export (mirrors the existing pattern already used
  in `src/core/cycle/synthesize.ts`) so the fix is pinned by pure-function unit
  tests with no DB required.

`--json` keeps the same shape, but not the same value in one field: because `deriveStatus()` feeds `CycleReport.status`, a synthesize-only cycle that actually did work now reports a non-`clean` status in JSON as well as in the human output. That is the intended correction rather than a side effect — the old value said "nothing happened" about a run that had processed a transcript. No new fields, no removed fields.

## Testing

New file `test/dream-clean-status-skip-visibility.test.ts` (7 tests, pure
functions, no DB):

- `transcripts_processed > 0` alone (queue idempotency reuse, no new pages) →
  `deriveStatus` returns `'ok'`, not `'clean'`
- `synth_pages_written > 0` alone → same
- genuinely idle synthesize (0 transcripts, 0 pages) still reports `'clean'`
  (no regression)
- a D8/D5 skip-only run (`transcripts_processed: 0`, non-empty `details.skips`)
  still derives `'clean'` — `deriveStatus` correctly stays conservative;
  surfacing *why* is `printHuman`'s job
- `printHuman` on a clean report with non-empty `details.skips` prints the skip
  reason, not just "Brain is healthy"
- `printHuman` on a clean report with empty/absent skips still prints only the
  healthy line (no regression)
- `printHuman` on the now-`'ok'` case prints the existing per-phase summary +
  totals line

Discrimination check (test fails against the pre-fix code): reverted both
source files to `upstream/master` locally and re-ran the new test file — it
fails to even load (`Export named '__testing' not found in module
'.../src/commands/dream.ts'`), confirming the test exercises the fix, not a
tautology. Restored the fix and the file is 7/7 pass.

```
bun test test/dream-clean-status-skip-visibility.test.ts
 7 pass
 0 fail
```

Also ran the full related suite (dream-cli-flags, dream.test,
cycle-enabled-phase-completeness, cycle-stamp-write-failure,
e2e/dream-synthesize-pglite, e2e/dream-synthesize-chunking,
e2e/dream-cycle-phase-order-pglite, cli-dream-engine-warn,
cycle-dream-output-root, dream-dir-source-stamp): 112/112 pass, 0 fail. Combined
with the new file's 7/7, that's 119/119 pass, 0 fail across all 11 files.
`bun run typecheck` and `bun run verify` both pass.

## Scope

No new CLI flags, no new config keys, no schema/migration changes, no writes
added — this is a report-honesty fix: fields that were already computed
(`details.skips`, `totals.transcripts_processed`) were being computed and then
discarded before they reached the user (dropped by `deriveStatus()`'s `anyWork`
check and `printHuman()`'s `status === 'clean'` short-circuit); this PR only
changes whether they're surfaced, not how they're computed.

Opening as Draft for review.

### DIFF
--- a/src/commands/dream.ts
+++ b/src/commands/dream.ts
@@ -445,9 +445,29 @@ function printHuman(report: CycleReport) {
   }
 
   if (report.status === 'clean') {
+    // A 'clean' cycle can still carry a skip reason worth surfacing — e.g.
+    // synthesize's D8 legacy-key / D5 oversize-chunk skips leave
+    // transcripts_processed/synth_pages_written at 0 (so deriveStatus sees
+    // no activity) while `details.skips` names exactly why each transcript
+    // was passed over. Without this, `--input <already-handled-file>`
+    // prints only "Brain is healthy" with no indication anything was
+    // examined and skipped.
+    const skipLines: string[] = [];
+    for (const p of report.phases) {
+      const skips = (p.details as { skips?: Array<{ filePath: string; reason: string }> } | undefined)?.skips;
+      if (Array.isArray(skips)) {
+        for (const s of skips) {
+          skipLines.push(`  - ${p.phase}: ${s.filePath} (${s.reason})`);
+        }
+      }
+    }
     console.log(
       `Brain is healthy. ${report.phases.length} phase(s) checked in ${(report.duration_ms / 1000).toFixed(1)}s.`,
     );
+    if (skipLines.length > 0) {
+      console.log('Skipped:');
+      for (const line of skipLines) console.log(line);
+    }
     return;
   }
 
@@ -479,6 +499,14 @@ function printHuman(report: CycleReport) {
     );
   }
 }
+
+// ── Test-only export ───────────────────────────────────────
+// `__testing` re-exports otherwise-private helpers so unit tests can pin
+// CLI output behavior without spawning a subprocess. Not part of the
+// runtime contract.
+export const __testing = {
+  printHuman,
+};
 
 // ─── CLI entry ─────────────────────────────────────────────────────
 
--- a/src/core/cycle.ts
+++ b/src/core/cycle.ts
@@ -2715,6 +2715,22 @@ function deriveStatus(phases: PhaseResult[], totals: CycleReport['totals']): Cyc
     // (resolve_symbol_edges). Without these, an edges-only cycle reports 'clean'
     // — indistinguishable from "nothing happened" even when N edges resolved.
     totals.edges_resolved > 0 ||
-    totals.edges_ambiguous > 0;
+    totals.edges_ambiguous > 0 ||
+    // `gbrain dream --input <file>` implies `--phase synthesize` (a
+    // synthesize-only cycle never touches sync/embed/etc, so those totals
+    // stay zero even on a genuinely productive run). Without these two, a
+    // real synthesize outcome — new pages written, or an already-completed
+    // transcript quietly reusing its prior job via the queue's
+    // idempotency_key dedupe — is indistinguishable from "nothing happened".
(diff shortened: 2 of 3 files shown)


garrytan:

time-attack:
Thanks @Masashi-Ono0611! This was adopted with credit in the v0.46.14.0 fix wave (commit 864dec4f): dream --input now says why it skipped, and the test ships as test/dream-clean-status-skip-visibility.test.ts. Closing as superseded — fresh PRs are always welcome.

Issue #4121, opened 2026-08-15
expansion spend bypasses the instrumented chat boundary — expansion_applied=true, zero rows in the budget audit


Masashi-Ono0611:
## What

`expand()` in `src/core/ai/gateway.ts` calls the AI SDK's `generateObject`/`generateText` directly instead of going through `chat()`. `chat()` is where `tracker.record()` runs, so **multi-query expansion spend is invisible to the budget tracker and to any usage accounting built on that boundary** — the call happens, the ledger stays silent.

## Evidence (real brain, Postgres engine, v0.45.11.0)

A single `query` op with expansion enabled produces:

| Signal | Value |
|---|---|
| `eval_candidates.expansion_applied` for that query | `true` |
| Chat-boundary usage rows recorded for that query | `0` |
| Occurrences of `expan*` across three weeks of `~/.gbrain/audit/budget-*.jsonl` | `0` |

The audit files are not empty over that window — they carry many rows with `sub_label: "gateway.chat"` from other phases (fact extraction, dream phases, embeddings). Expansion never appears in them.

That pairing is what makes this reportable rather than a guess: the search path records that expansion ran, and the accounting path has no trace of it. I mention it because I initially drew the wrong conclusion from the ledger alone — "expansion isn't firing" — and only found the real shape after checking `expansion_applied`.

## Why it matters

- Budget caps under-count. A phase running under `withBudgetTracker` with a `--max-usd` ceiling is enforced against a total that omits the expansion call for every query it makes.
- The `query` op enables expansion by default independent of the search mode, so this is the common path rather than an opt-in corner.
- Reconciling gbrain's own numbers against a provider invoice leaves an unattributable gap.

`generateOcrText()` looks like the same shape (a direct `generateText()` outside the instrumented boundary), so a fix here may want to consider whether the seam should be shared rather than patched per call site. I have not measured that one.

## Secondary observation, same area

`gbrain models doctor`'s expansion probe calls `chat({ model: expansionModel, ... })`. That confirms the model is reachable, but it does not exercise `expandQuery()` → `gatewayExpand()`, which is the path that actually runs in production (and which selects `generateObject` or `generateText` depending on the provider's declared structured-output support). A probe that stays green while the real expansion path is unaccounted is a weaker signal than it appears. Noting it here rather than as a second issue, since a fix likely touches the same seam.

## Suggested direction (not a PR yet)

[...]

PR #4124, opened 2026-08-15
fix(ai): expansion calls record to the budget tracker instead of spending invisibly

Masashi-Ono0611:
## Symptom

`expand()` (query expansion) spend is invisible to `BudgetTracker`, even inside a `withBudgetTracker()` scope. `gateway.chat()` composes with the ambient tracker via an internal `_recordBudget` closure that calls `tracker.record({..., label: 'gateway.chat'})`. `expand()` calls the AI SDK's `generateObject()` / `generateText()` directly and never goes through that closure, so its usage was never recorded.

Observed on a real v0.45.11.0 brain: for a single `query` op where `eval_candidates.expansion_applied: true` was recorded, the same-window chat-boundary usage rows were 0, and grepping 3 weeks of `~/.gbrain/audit/budget-2026-W{31,32,33}.jsonl` for `expan*` returned 0 hits — while the same files carry many `sub_label: "gateway.chat"` rows from other phases. Re-measuring this exact number isn't necessary to evaluate the fix; it's cited to show the gap is real, not theoretical.

Related to #4121

## Fix

Resolves the ambient tracker inside `expand()` via the existing exported `getCurrentBudgetTracker()` and records on every **successful** LLM call site (there are 3: native-provider `generateObject`, the openai-compatible structured-output `generateObject` path, and the schemaless `generateText` "viaText" fallback), tagged with `label: 'gateway.expand'` so it's distinguishable in the audit JSONL (`sub_label`) instead of silently merging into `gateway.chat`. Fails open — no ambient tracker means no-op — and swallows `BudgetExhausted` (TX1) the same way `chat()`'s `_recordBudget` does (surfaced on the next `reserve()`, not thrown out of `expand()`).

One new test-only transport seam is added (`_generateObjectTransport`), and `viaText()` is rewired from a direct `generateText()` call to the already-existing `_generateTextTransport` so all 3 call sites are interceptable in tests, mirroring the `_chatTransport` / `_generateTextTransport` / `_rerankTransport` pattern already used repeatedly in this file. They default to the real SDK functions in production and are reset in `clearGatewayState()`.

## Deliberately out of scope

- **No refactor of `expand()` to go through `chat()`.** `chat()`'s retry/providerMetadata/stop-reason handling is a much larger surface than an accounting fix warrants.
- **`gbrain doctor`'s expansion probe is untouched** (separate concern from #4121).
- **Failed LLM calls are not recorded** — only successful calls. Mirrors the issue's own scope (successful expansion spend going unrecorded). Recording failures the way `chat()` does would need a pessimistic-fallback input/output-token estimate (`chat()` has `estimateChatInputTokens()` + a precomputed `maxOutputTokens` to build one on error; `expand()` has neither today), which is a materially bigger diff. Confirmed via review that this omission is honest — a failed call simply produces no audit row, it's never double-counted or misattributed to `gateway.chat`.

## Tests

`test/core/budget/expand-records-budget.test.ts` (new, 5 tests) — hermetic, no network/provider/env touched:
- Native `generateObject` path records usage under the ambient tracker, with the correct `sub_label`/model/token counts in the audit JSONL, and moves `tracker.totalSpent`.
- `expand()` outside any `withBudgetTracker` scope stays a no-op (back-compat).
- `BudgetExhausted` from `record()` is swallowed — `expand()` still returns its best-effort result.
- openai-compatible schemaless (`viaText`) path records usage with the same `sub_label`.
- openai-compatible structured-output `generateObject` path records usage — verified it's NOT reached via `viaText()` (no shipped recipe currently opts into `supports_structured_outputs`, so this uses the existing `__setTestRecipesForTests()` seam, already used by `no-batch-cap-suppression.serial.test.ts`, to register a synthetic recipe rather than mutating any shipped one).

Red-green, stated precisely because the two halves differ in strength:

- **Main fix**: reverting `gateway.ts` to upstream while keeping the new test does not produce a failing assertion — the module fails to load (`Export named '__setGenerateObjectTransportForTests' not found`, exit 1), because the test seam ships with the fix. That proves the test cannot pass without the change, but it is a load error rather than a red assertion.
- **Structured-output branch**: removing just that one `recordExpansionUsage()` call site produces the sharper signal — exactly that test fails (`callsRecorded 1 -> 0`) while the other four stay green. Restoring it returns 5/5.

```
bun test test/ai/ test/core/budget/   -> 532 pass / 0 fail
bun run typecheck                      -> clean
bun run verify                         -> 39/39 checks green
```

## Review

Passed 3 rounds of independent Codex review (`model_reasoning_effort=high`). Round 1 flagged missing coverage for the structured-output `generateObject` branch (added). Round 2 flagged that the added test mutated the shipped `litellm` recipe singleton with a shape-inexact restore (switched to the established `__setTestRecipesForTests()` synthetic-recipe seam). Round 3: 0 Critical, 0 Warning.

### DIFF
--- a/src/core/ai/gateway.ts
+++ b/src/core/ai/gateway.ts
@@ -160,6 +160,11 @@ type EmbedManyFn = typeof embedMany;
 let _embedTransport: EmbedManyFn = embedMany;
 type GenerateTextFn = typeof generateText;
 let _generateTextTransport: GenerateTextFn = generateText;
+// Test-only seam for expand()'s structured-output SDK call. Mirrors
+// _generateTextTransport (see __setGenerateObjectTransportForTests). Never
+// swapped in production — expand() always calls the real generateObject.
+type GenerateObjectFn = typeof generateObject;
+let _generateObjectTransport: GenerateObjectFn = generateObject;
 // v0.41.6.0 D1: tests that install a transport stub also pass the
 // embedding-creds preflight, matching the chat-transport fast-path
 // pattern. Set when __setEmbedTransportForTests is called with a
@@ -615,6 +620,7 @@ function clearGatewayState(): void {
   _shrinkState.clear();
   _embedTransport = embedMany;
   _generateTextTransport = generateText;
+  _generateObjectTransport = generateObject;
   _embedTransportInstalled = false;
   _chatTransport = null;
   _warnedRecipes.clear();
@@ -670,6 +676,17 @@ export function __setEmbedTransportForTests(fn: EmbedManyFn | null): void {
  */
 export function __setGenerateTextTransportForTests(fn: GenerateTextFn | null): void {
   _generateTextTransport = fn ?? generateText;
+}
+
+/**
+ * Test-only seam for expand()'s generateObject call (the structured-output
+ * path used for native providers and openai-compatible recipes that declare
+ * supportsStructuredOutputs). Same shape as __setGenerateTextTransportForTests.
+ *
+ * @internal exported for tests; not part of the public gateway API.
+ */
+export function __setGenerateObjectTransportForTests(fn: GenerateObjectFn | null): void {
+  _generateObjectTransport = fn ?? generateObject;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


time-attack:

Masashi-Ono0611:
Live verification from the same production brain as my notes on #4120/#4111 — the ambient-tracker path exercised for real this time.

Before this run, `grep 'gateway.expand' ~/.gbrain/audit/budget-*.jsonl` matched zero rows across every week file. On this brain that is expected rather than incriminating — search mode is the balanced default (expansion off), so routine search flows here never trigger expansion. Which is also why this verification had to drive `expand()` directly instead of waiting for a nightly.

With this PR's diff applied to the installed v0.45.11.0 tree, I ran `expand()` inside a `withBudgetTracker` scope (a `BudgetTracker` with `maxCostUsd: 0.05`, mirroring how a cycle carries its tracker):

- `expand("what changed in the retrieval architecture last week")` returned 5 queries, tracker `totalSpent` = $0.000542.
- The audit row landed in `budget-2026-W33.jsonl`: `event=record, sub_label=gateway.expand, model=anthropic:claude-haiku-4-5-20251001, input_tokens=327, output_tokens=43, actual_cost_usd=0.000542, max_cost_usd=0.05`.

One operational note for anyone else verifying: a plain nightly won't exercise this on a default install — with search mode `balanced`/`conservative` the expansion knob is off, so routine search flows never call `expand()`. Driving `expand()` directly inside `withBudgetTracker` is the direct way to see the row.

garrytan:

garrytan:
Thank you @Masashi-Ono0611 — this fix has been absorbed into the v0.46.13.0 fix wave (#4227) with credit in the CHANGELOG. We applied your record-on-success diff at all three expand sites (with the transport seam) and extended it with pessimistic failed-path records, OCR instrumentation, and a NaN-guarded usage normalizer. Closing this PR as superseded by #4227 — the fix ships, authored-in with your credit.

Masashi-Ono0611:
Closing — explicitly absorbed by \`cc949945\` ("fix(ai): expansion and OCR spend land in the budget tracker — absorb PR #4124 + failure-path records (#4121)"), which says in its body: "Absorbs PR #4124 by @Masashi-Ono0611". Confirmed the gateway.expand budget-tracker recording is live in v0.46.19.0 while rebasing a local patch-stack through it today. No further action needed on this PR.

PR #4129, opened 2026-08-15
fix(path-confine): Windows path containment - use path.relative instead of '/' string prefix

alexsanchezw:
## Problem

`isPathContained` (src/core/path-confine.ts) appends a hardcoded `'/'` to the resolved parent before the `startsWith` check. On Windows, `realpathSync` returns backslash-separated paths (`C:\Users\...`), so the prefix never matches and the containment check rejects **every** legitimate child path.

Observable symptom: `gbrain hook session-end` always degrades with `transcript_outside_projects_dir` on Windows, so the dream-corpus transcript pipeline never ingests anything on that platform.

## Repro

```js
const parent = realpathSync(join(homedir(), '.claude', 'projects'));
// => C:\Users\me\.claude\projects
const child = realpathSync(join(parent, 'proj', 'session.jsonl'));
child.startsWith(parent + '/');  // false — always
```

## Fix

Replace the string-prefix idiom with `path.relative()`: empty result = same dir; `..` prefix or absolute result = outside. This handles `\` vs `/`, case-insensitive drive letters, and cross-drive paths, and keeps `/foo` from matching `/foobar`. Fail-closed behavior (unresolvable paths = not contained) is unchanged.

Verified on Windows 11: direct child OK, parent itself OK, sibling with common prefix rejected, `..` traversal rejected, other drive rejected, lowercase drive letter OK, mixed separators OK. `test/path-confine.test.ts` goes from 7 failing to 6 on Windows (the remaining 6 are EPERM symlink-creation failures needing admin rights, unrelated).

### DIFF
--- a/src/core/path-confine.ts
+++ b/src/core/path-confine.ts
@@ -20,7 +20,7 @@
  */
 
 import { realpathSync, existsSync, type Stats } from 'fs';
-import { resolve as resolvePath, relative, isAbsolute, dirname, basename, join } from 'path';
+import { resolve as resolvePath, relative, isAbsolute, dirname, basename, join, sep } from 'path';
 
 /**
  * Symlink-safe path confinement: realpath BOTH sides, then a separator-aware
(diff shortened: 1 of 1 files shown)


time-attack:

time-attack:
Superseded: the hardcoded-'/' bug in isPathContained was fixed on master in v0.46.14.0 (commit 864dec4f, adopting PR #4103's sep-based fix, which the wave commit names as closing this duplicate). The hook session-end transcript containment path (claude-code-jsonl.ts) now works on Windows.

PR #4132, opened 2026-08-15
fix(serve-http): thread localFederatedSourceIds for no-grant legacy tokens (#3242 parity)

kyle944:
## The bug

#3242 gave the legacy HTTP transport and the stdio dispatch site the same rule: a caller carrying no operator source grant has its unqualified reads widened across the federated source set. The SDK transport in `src/commands/serve-http.ts` never got it.

So one legacy bearer token sees different data depending on which serve mode answers. On a brain whose code lives in a non-default source, `/mcp` under `serve --http` collapses to scalar `default` and returns nothing, while the same token on the legacy transport resolves the federated set and answers. Nothing errors — the reads just come back empty, which reads as "not indexed yet" rather than as a scoping bug.

## The fix

Populate `localFederatedSourceIds` at the SDK transport site too, gated the same way the sibling transports gate it. `AuthInfo` gains `hasSourceGrant`, mirroring `AuthResult.hasSourceGrant` in `src/mcp/http-transport.ts`, and `GBrainOAuthProvider` sets it from `permissions.source_id`.

The decision moved into `noGrantFederatedScope()` in `src/core/source-resolver.ts`, beside the resolver it wraps, so it is testable and so a third transport needing the same rule has one place to call.

## On the gate

It is `hasSourceGrant === false`, not a falsy check, and that is the part worth reviewing. The flag has three meaningful states: `false` is the legacy-bearer no-grant floor and may widen, `true` is an operator-set scope, and `undefined` is an OAuth client. Simplifying to `hasSourceGrant ||` would hand every OAuth client the full federated set. `test/no-grant-federated-scope.test.ts` covers that case specifically, and flipping the gate fails it.

Resolution stays best-effort: a sources table that cannot be read leaves the scalar scope standing rather than failing the request, matching the surrounding transport behaviour. That path is covered too.

## Verification

`bun run verify` green (43/43), `tsc --noEmit` clean, and the five federated-scope suites pass together (73 tests). New coverage is 6 cases in `test/no-grant-federated-scope.test.ts`; the `KEY_FILES.md` entry for `source-resolver.ts` records the new export and its pin.

Found while upgrading a brain from 0.44.0.0 to v0.45.15.0, where code-intel queries over `/mcp` answered on one serve mode and not the other.


kyle944:
Overlap worth flagging before anyone reviews this: the same serve-http commit also rides along on #4011, which is primarily the code-intel routing fix. I opened this separately after reading that PR's title rather than its commits.

What this branch adds over the copy in #4011 is the coverage — `noGrantFederatedScope()` extracted so the gate is testable, and six cases pinning which callers may widen, including the OAuth-client case that a falsy gate would break.

Take whichever is easier. If you would rather have it all in #4011, say so and I will move the test commit there and close this. I left #4011 alone rather than force-pushing a branch already under review.

time-attack:

garrytan:

time-attack:
Adopted with credit in v0.46.14.0 (864dec4f): the SDK transport now threads localFederatedSourceIds via the shared fail-closed noGrantFederatedScope gate (src/commands/serve-http.ts, source-resolver.ts:422), with your coverage landed as test/no-grant-federated-scope.test.ts. Thank you @kyle944!

PR #4133, opened 2026-08-15
fix(cli): route sources --help to its own usage block instead of the circular generic stub

Masashi-Ono0611:
## Summary

`gbrain --help` tells the reader, in the SOURCES block (`src/cli.ts`):

```
  sources --help                     Full subcommand list (rename, default, attach,
                                     current, federate, set-cr-mode, webhook, harden, ...)
```

`gbrain sources --help` answered `gbrain sources - run gbrain --help for the full command list.`
— pointing straight back. `sources` was in `CLI_ONLY` but not in `CLI_ONLY_SELF_HELP`, so the
dispatcher's generic stub fired before `runSources` could reach the complete `printHelp()` it
already ships. The eight subcommands the top-level help names as discoverable only there were
discoverable from neither side.

This follows the same shape the tip commit used for `jobs` in #4125, including the guard that
keeps a nested `--help` from dispatching.

## What changed

1. `sources` added to `CLI_ONLY_SELF_HELP` and `SELF_HELP_WITHOUT_ENGINE`, so `sources --help`
   reaches its own usage block and works on a machine with no brain configured (the state a
   reader running `--help` is most likely to be in).
2. A nested-help guard in `runSources` before the subcommand switch. Step 1 alone would have been
   worse than the bug: `hasHelpFlag()` matches `--help`/`-h` in any position while `runSources`
   dispatches on `args[0]`, so `gbrain sources detach --help` would have run `runDetach()` — which
   takes no engine and calls `unlinkSync` on the caller's `.gbrain-source` — instead of printing
   help. Caught in review before this was pushed; the regression test below fails without the
   guard and its captured output shows the actual deletion.
3. `sources webhook --help` is served from a hoisted `SOURCES_WEBHOOK_HELP` constant shared with
   `runWebhook`'s own `--help` case, so the guard does not swallow the one subcommand that has
   richer help. This was already unreachable before this PR (the generic stub caught it too) — it
   is not a regression this PR introduces, it is the same class this PR is fixing.
4. `webhook` added to `printHelp()`. Auditing the eight subcommands the top-level help promises,
   it was the only one missing.

## Verification

`bun test test/cli-help-discoverability.test.ts` 19 pass / 0 fail; `bun run typecheck` and
`bun run verify` both exit 0.

Red-green, compiled binary, `GBRAIN_HOME` pointing at a nonexistent dir:

| command | before | after |
|---|---|---|
| `sources --help` | generic stub | full usage block, exit 0 |
| `sources list --help` | generic stub | full usage block (does not dispatch into `runList`) |
| `sources webhook --help` | generic stub | webhook's own set/show/rotate/clear block |
| `sources detach --help` (dir with `.gbrain-source`) | generic stub | help printed, dotfile survives |

Related to #3686 and #4003, which report the same generic-stub class for other command groups;
this PR touches only `sources`.


time-attack:
Automated triage: the fix itself verified cleanly — the circular stub is reproducible on master and gone at this head, exit codes are unchanged across every sources subcommand's help path, the detach-with-dotfile guard holds live, and the supply-chain scan is clean. However, two independent adversarial review passes both converged on the same concrete gap: this PR adds `sources` to `SELF_HELP_WITHOUT_ENGINE` but does not add it to `HELP_WITHOUT_BRAIN` in `test/cli-help-without-brain.serial.test.ts`, which the map's own doc comment (src/cli.ts:178-180) declares mandatory, and which the #4125 jobs precedent this PR cites as its template did do. The PR's own new tests don't cover that property hermetically (they spread process.env without scrubbing `DATABASE_URL`/`GBRAIN_DATABASE_URL` and spawn without `--no-env-file`), so the no-brain contract for `sources` is durably pinned nowhere. The fix is one line in the serial test, and a reviewer confirmed it would pass today. Routing to human review with this finding rather than auto-merging; happy path would be the author pushing that one-line test addition.

Masashi-Ono0611:
Pushed 0ce1bf2e with both gaps closed: `sources` is added to `HELP_WITHOUT_BRAIN` in `test/cli-help-without-brain.serial.test.ts` (one line, mirroring the #4125 jobs precedent), and this PR's own test spawns are now hermetic — `DATABASE_URL`/`GBRAIN_DATABASE_URL` are deleted from the spread env and the spawns pass `--no-env-file`, copying the serial test's established idiom. Serial suite: 8 pass / 0 fail (7 HELP_WITHOUT_BRAIN entries incl. the new `sources`, plus the tripwire). Discoverability file: 19 pass / 0 fail. Typecheck clean.

garrytan:

Masashi-Ono0611:
Closing — this landed already. Commit \`864dec4f\` (v0.46.14.0 fix-wave, #4226) explicitly credits: "Adopted from PR #4133 by @Masashi-Ono0611 (pinned patch)". Confirmed while attempting a rebase today — every conflict resolved to content byte-identical with \`upstream/master\`, and the branch ended up with zero commits ahead once rebased. No further action needed on this PR.

PR #4134, opened 2026-08-15
fix(facts): exclude internal extraction audit rows from fact reads

Masashi-Ono0611:
## Summary

`extract-conversation-facts` writes a per-page terminal marker into the `facts` table for every
page it finishes — `EXTRACTION_COMPLETE` (`TERMINAL_AUDIT_SOURCE`) or `EXTRACTION_NOT_APPLICABLE`
(`NON_EXTRACTABLE_AUDIT_SOURCE`). The insert's own comment says it plainly:

> this is internal audit state, not user-facing knowledge

But the rows are written with `kind: 'fact'`, `entity_slug: null`, and no `expired_at`, and
`FactListOpts` offered only `activeOnly` / `limit` / `offset` / `kinds` / `visibility`. None of
those can separate a marker from a fact: the kind is the same, the row is never expired, and
visibility is unset. So there was no way for a reader to ask for knowledge and not get audit state.

Every non-engine caller of `listFactsByEntity` / `listFactsSince` / `listFactsBySession` is a
knowledge surface — the `recall` op, the `entity` verb's card, ambient turn-context, the facts
meta-hook, the consolidate phase, and the `recall` CLI. There are no diagnostic callers.
`extract-conversation-facts`'s own resume path reads the markers through `engine.executeRaw`, not
these methods, so it is unaffected.

## Scale, measured on a real brain

165 of 1,517 active facts (10.9%) are markers. During the two batch-extraction hours they were
13.7% and 8.4% of the facts created in that hour. One minute-level window created 73 facts of
which 73 were markers — a newest-50 read issued there returns internal audit state and no
knowledge at all.

## What changed

`FactListOpts.excludeAuditRows`, defaulting to true — the same shape and default as the
neighbouring `activeOnly`, which already hides a non-user-facing row class by default. Both
engines filter it across the three list methods. `excludeAuditRows: false` restores the previous
behavior. The two source constants moved to `src/core/facts/audit-sources.ts` so both engines can
import them statically (engine-live paths must not use runtime dynamic `import()`);
`extract-conversation-facts.ts` re-exports them under their existing names, so current importers
are untouched.

`listSupersessions` deliberately does NOT filter. It is the supersession audit log — it should
keep showing every superseded row. PGLite reaches it through the shared `_listFacts` helper, so it
passes `excludeAuditRows: false` explicitly to match Postgres. Review caught this as an
engine-parity divergence before it was pushed; it is now pinned by a test on both engines.

Deliberately out of scope: `countUnconsolidatedFacts` and `getFactsHealth` (both engines) still
count markers, so `include_pending` and the doctor health numbers can exceed what a default read
returns. Changing a health surface's numbers is a separate behavior change and is not folded in
here.

## Verification

`test/facts-engine.test.ts` 24 pass; `test/extract-conversation-facts.test.ts` +
`test/doctor-conversation-facts-backlog.test.ts` 69 pass; `bun run typecheck` and `bun run verify`
(39/39) both exit 0.

Coverage was verified by removal, not assumed:

| reverted | result |
|---|---|
| all three engine files | 5 failures across since / byEntity / bySession |
| only PGLite's `listSupersessions` override | 1 failure |
| only Postgres's three predicates | 3 failures in the dual-engine block |

The last row is the one that matters — the new `DATABASE_URL`-gated block in
`test/e2e/engine-parity.test.ts` seeds identical rows into real Postgres and PGLite side by side,
so a one-engine-only regression cannot pass. It was run green against a real pgvector container,
and skips gracefully without `DATABASE_URL`.

### DIFF
--- a/src/commands/extract-conversation-facts.ts
+++ b/src/commands/extract-conversation-facts.ts
@@ -93,6 +93,16 @@ import { withRefreshingLock, LockUnavailableError } from '../core/db-lock.ts';
 import { assertFactsEmbeddingDimMatchesConfig } from '../core/embedding-dim-check.ts';
 import { writeReceipt, shortRunId } from '../core/extract/receipt-writer.ts';
 import { upsertExtractRollup } from '../core/extract/rollup-writer.ts';
+import {
+  TERMINAL_AUDIT_SOURCE,
+  NON_EXTRACTABLE_AUDIT_SOURCE,
+} from '../core/facts/audit-sources.ts';
+
+// Re-exported verbatim so existing importers (eval write-back, this file's
+// own tests, the doctor backlog tests) keep working unchanged. Moved to
+// src/core/facts/audit-sources.ts so both engines can filter these rows out
+// of listFacts* reads via a static import (see that file for why).
+export { TERMINAL_AUDIT_SOURCE, NON_EXTRACTABLE_AUDIT_SOURCE };
 
 // ---------------------------------------------------------------------------
 // Tunables (exported for tests).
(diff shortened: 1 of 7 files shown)


Masashi-Ono0611:
Closing this as a duplicate of my own earlier #4008, which reports the same defect, touches the
same files, is already green and mergeable, and has been in human review since 08-13. Two open PRs
from the same author for one defect is not something a maintainer should have to arbitrate.

My apologies for the noise — the cause was a hole in my own duplicate check. I searched the issue
tracker and listed my own issues with `--include-prs=false`, so an existing *PR* of mine on the
same defect was structurally invisible to the check. Fixed on my side.

The one substantive difference is worth carrying over, and I will propose it on #4008 rather than
here: this branch filtered by the audit rows' `source` (the two constants in
`extract-conversation-facts.ts`) where #4008 filters by the `fact` text
(`EXTRACTION_COMPLETE` / `EXTRACTION_NOT_APPLICABLE`). Filtering on the writer rather than the
payload is the more durable predicate — it cannot hide a genuine user fact that happens to carry
that exact text, and a third audit string added later is covered automatically instead of silently
slipping through.

Nothing else here is lost: #4008 already implements the exclusion opt-in with explicit call sites,
which is the more conservative shape and does not change behavior for callers that do not ask for
it.

Issue #2756, opened 2026-07-11
conversation-facts cannot run on iMessage exports: type allowlist omits imessage, 12h time-only pattern missing, multi-day date headers not anchored


notjbg:
## Summary

`extract-conversation-facts` can never run on iMessage exports, even though gbrain's own imessage-sync integration produces them. Three independent gaps stack up:

1. **Type allowlist omits iMessage.** `ALLOWED_TYPES` in `src/commands/extract-conversation-facts.ts` is `['conversation', 'meeting', 'slack', 'email']` — the `imessage` / `imessage-daily` types the integration writes are excluded, so extraction never considers those pages at all.
2. **No pattern for the time-only 12-hour message shape.** iMessage exports produce `**Speaker** (9:04 AM): text` — bold speaker, paren-wrapped 12h wall-clock time **without a date**, colon outside the bold. This falls in the gap between the `imessage-slack` builtin (requires a full `YYYY-MM-DD` inside the parens) and `bold-paren-time` (24h only, rejects an AM/PM suffix). Result: `no_match` on every line.
3. **Multi-day exports anchor every message to one fallback date.** iMessage pages commonly carry a frontmatter `date_range` plus per-day `## YYYY-MM-DD` section headers. The parser ignores markdown date headers, so every message in a multi-day page gets the single frontmatter fallback date — and facts inherit extraction-time `valid_from` instead of the conversation's real time.

## Why it matters

iMessage is a first-party source (imessage-sync ships with gbrain), and conversation-facts is the feature that makes conversational sources queryable. Together these gaps mean the entire iMessage corpus silently contributes zero facts — no error, no warning, just `no_match`/skip.

## Tested fix (running in production since 2026-06-01)

We've been running the three patches below on a ~43K-page Postgres brain since June 1 (currently v0.42.58.0), reapplied across every upgrade. Conversation-facts extraction on iMessage pages works end-to-end with them; the new pattern's test_negative fixtures pin the non-shadowing guarantees (24h shapes still fall through to `bold-paren-time`, full-date shapes to `imessage-slack`).

(Test fixtures below are sanitized — phone numbers and message text are placeholders.)

<details>
<summary>Patch 1 — builtins.ts: <code>bold-paren-time-12h</code> PatternEntry</summary>

```diff
--- a/src/core/conversation-parser/builtins.ts
+++ b/src/core/conversation-parser/builtins.ts
@@ -176,6 +176,54 @@
     ],
     source_doc:
[...]

Issue #2185, opened 2026-06-14
parseFlag silently accepts unknown flags (e.g. --dry-run on init --migrate-only no-ops and applies real migrations)


ethanbeard:
## Summary

`parseFlag` resolves flags with a naive `args.indexOf(flag)`, which returns `-1` cleanly for any flag the function doesn't know about. Unknown flags are therefore silently accepted and ignored — no warning, no error, non-zero exit. This is a foot-gun for safety-critical flags that a caller assumes will prevent side effects.

## Concrete incident

```
gbrain init --migrate-only --dry-run
```

This applied real migrations. `--dry-run` is not a parsed flag on `init`, so `parseFlag`/the flag check silently no-op'd it, and `--migrate-only` ran for real. A command that the operator believed was a no-op preview mutated the schema (in our case 92 → 115, 23 migrations). The command returned success, which told the operator nothing about whether `--dry-run` had any effect.

## Root cause

`src/commands/agent.ts:25` (and the same pattern elsewhere, e.g. `src/commands/book-mirror.ts:81`):

```ts
function parseFlag(args: string[], flag: string): string | undefined {
  // indexOf-based lookup; unknown flags are simply never found and never rejected
}
```

There is no strict-flag layer that rejects or warns on unrecognized flags.

## Suggested resolution

Add a strict-flag validation pass per subcommand that errors (or at least warns loudly) on unrecognized flags, especially for commands with destructive/irreversible side effects. At minimum, `--dry-run` should be a recognized no-op-or-error everywhere it's plausibly expected, rather than silently ignored.

Observed on v0.42.36.0.

PR #4135, opened 2026-08-15
fix(facts): make the conversation-type allowlist single-source instead of 5 hand-copied lists

Masashi-Ono0611:
## Summary

The conversation-facts type allowlist — `conversation`, `meeting`, `slack`, `email`, `imessage`,
`imessage-daily` — is hand-copied into five places across four files:

| # | site |
|---|---|
| 1 | `src/commands/extract-conversation-facts.ts` — `export const ALLOWED_TYPES` (canonical) |
| 2 | `src/commands/doctor.ts` — the default when `cycle.conversation_facts_backfill.types` is unset |
| 3 | `src/commands/doctor.ts` — `const allowedTypes` in the parse-coverage sampler |
| 4 | `src/commands/jobs.ts` — inline `.includes(t)` filter |
| 5 | `src/commands/sources.ts` — `FACTS_BACKFILL_ALLOWED` |

It has already drifted. In `jobs.ts`, one line below a runtime filter that accepts all six values,
the result was cast as `('conversation' | 'meeting' | 'slack' | 'email')[] | undefined` — the
four-element set from before `imessage`/`imessage-daily` were added. To be precise about severity:
this was **type-level only**. The `as` assertion is unchecked, the narrower array type is
assignable to the wider parameter under array covariance, and the values flowed through as
strings, so `typecheck` passed before and after and no runtime behavior differed. It was a
declared type that had stopped describing its data.

CLAUDE.md already states the principle this violates, for pricing:

> One canonical chat-pricing table ... Every other table ... is a DERIVED view, never a
> hand-copied duplicate — so cross-table price drift is structurally impossible.

Same situation, without the discipline. The stale cast is exactly the drift that rule predicts.

## What changed

The six values move to a new leaf module, `src/core/facts/conversation-types.ts`, exporting
`ALLOWED_TYPES` and the derived `AllowedType`. `extract-conversation-facts.ts` imports from there
and re-exports both names verbatim, so its existing importers are untouched. Sites 2-5 import the
leaf. `jobs.ts`'s cast is replaced by a `(t): t is AllowedType` type guard, so the narrowing is
checked rather than asserted. `doctor.ts` site 2 keeps its config-overridable-default semantic
exactly — only where the default comes from changed.

**The set of allowed types is unchanged.** This PR changes where the six values are defined, not
what they are.

### Why a leaf module and not a direct import

The first version of this PR had sites 2-5 import the constant straight from
`extract-conversation-facts.ts`. CI caught the problem: `scripts/generate-flag-registry.ts` scans
a command module's own text plus one level of its relative imports for flag-shaped literals, and
treats static and dynamic imports identically. Importing anything from that command therefore
attributed its entire flag surface to the importers — the regenerated registry gained about ten
entries each on `doctor`, `repos` and `sources` (`--segment-limit`, `--types`, `--slug`,
`--sleep`, `--override-disabled`, and friends).

That registry backs #2185's strict unknown-flag validation, so the naive version would have made
three commands start accepting flags they had been correctly rejecting — a silent widening of
their surface, in a PR that claims to change nothing about behavior. Regenerating the registry to
get a green build would have baked that in. The leaf module has no flag surface of its own, so the
committed registry is byte-identical before and after.

## Verification

`bun run build:flag-registry` followed by `git diff --stat` on the generated registry is empty —
that is the load-bearing check for the paragraph above, and it is pinned by a new test that
reproduces the generator's own relative-import scan for `doctor`/`repos`/`sources`/`jobs` rather
than hardcoding an assumed flag list.

`test/conversation-facts-type-allowlist-drift.test.ts` is modelled on `test/model-pricing.test.ts`'s
drift guard, which asserts each derived view equals canonical. It also asserts the leaf itself
carries no flag-shaped literals, so the fix cannot silently regress.

Red-green: reverting the four source files and deleting the leaf makes the drift test fail hard.
Against the earlier hand-copied state the structural guards failed while the value-level
assertions passed — which is the point: the values had not drifted yet, only the source had, and a
value-only test would never have caught the `jobs.ts` cast.

`bash scripts/test-shard.sh 5 10` 1755 pass / 0 fail; `test/cli-flag-validation.test.ts` 25/25;
the drift test plus `test/extract-conversation-facts.test.ts` and
`test/doctor-conversation-facts-backlog.test.ts` 88 pass / 0 fail; `bun run typecheck` exit 0;
`bun run verify` 43/43 exit 0.

Related to #2756, which added `imessage`/`imessage-daily` and is where the cast fell behind.

### DIFF
--- a/src/commands/doctor.ts
+++ b/src/commands/doctor.ts
@@ -1,5 +1,8 @@
 import type { BrainEngine } from '../core/engine.ts';
 import { REPAIR_SOURCE_CONFIG_SQL } from '../core/source-config-sql.ts';
+// Leaf module (no flag surface of its own) — see that file for why this
+// isn't imported from extract-conversation-facts.ts directly (#4135).
+import { ALLOWED_TYPES } from '../core/facts/conversation-types.ts';
 import { setCliExitVerdict } from '../core/cli-force-exit.ts';
 import * as db from '../core/db.ts';
 import { LATEST_VERSION, getIdleBlockers } from '../core/migrate.ts';
@@ -3831,9 +3834,8 @@ export async function computeConversationFactsBacklogCheck(
     const typesRaw = await engine.getConfig(
       'cycle.conversation_facts_backfill.types',
     );
-    // Default mirrors extract-conversation-facts.ts's ALLOWED_TYPES — the
-    // single source of truth for the conversation-facts type allowlist.
-    const { ALLOWED_TYPES } = await import('./extract-conversation-facts.ts');
+    // Default mirrors ALLOWED_TYPES — the single source of truth for the
+    // conversation-facts type allowlist.
     let types: string[] = [...ALLOWED_TYPES];
     if (typesRaw) {
       try {
@@ -6277,7 +6279,7 @@ export async function buildChecks(
       const { readConversationBodyForParsing } = await import('../core/conversation-parser/body.ts');
       const { parseConversation } = await import('../core/conversation-parser/parse.ts');
       // Single source of truth for the conversation-facts type allowlist.
-      const { ALLOWED_TYPES: allowedTypes } = await import('./extract-conversation-facts.ts');
+      const allowedTypes = ALLOWED_TYPES;
       // PageFilters supports singular `type` only; iterate the allowed types
       // and cap at ~50/each to land at ~200 total max.
       const sample: import('../core/types.ts').Page[] = [];
--- a/src/commands/doctor.ts
+++ b/src/commands/doctor.ts
@@ -3831,7 +3831,10 @@ export async function computeConversationFactsBacklogCheck(
     const typesRaw = await engine.getConfig(
       'cycle.conversation_facts_backfill.types',
     );
-    let types = ['conversation', 'meeting', 'slack', 'email', 'imessage', 'imessage-daily'];
+    // Default mirrors extract-conversation-facts.ts's ALLOWED_TYPES — the
+    // single source of truth for the conversation-facts type allowlist.
+    const { ALLOWED_TYPES } = await import('./extract-conversation-facts.ts');
+    let types: string[] = [...ALLOWED_TYPES];
     if (typesRaw) {
       try {
         const parsed = JSON.parse(typesRaw);
@@ -6273,7 +6276,8 @@ export async function buildChecks(
     try {
       const { readConversationBodyForParsing } = await import('../core/conversation-parser/body.ts');
       const { parseConversation } = await import('../core/conversation-parser/parse.ts');
-      const allowedTypes = ['conversation', 'meeting', 'slack', 'email', 'imessage', 'imessage-daily'] as const;
+      // Single source of truth for the conversation-facts type allowlist.
+      const { ALLOWED_TYPES: allowedTypes } = await import('./extract-conversation-facts.ts');
       // PageFilters supports singular `type` only; iterate the allowed types
       // and cap at ~50/each to land at ~200 total max.
       const sample: import('../core/types.ts').Page[] = [];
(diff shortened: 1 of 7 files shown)


time-attack:
Automated triage: the consolidation itself verified clean — two independent reviews confirmed element-by-element (via `git show origin/master:`) that all five hand-copied lists were identical six-element sets in the same order, so no call site widened or narrowed; the `as const` type surface is preserved (and jobs.ts upgrades an unchecked cast to a real type guard); no remaining hand-copied literal exists repo-wide; supply-chain scan clean. Both reviews nevertheless landed on the same concrete problem via mutation testing: the drift guard is a byte-exact literal/order matcher, so it does NOT catch the harmful direction of drift. Demonstrated on a pristine head worktree — re-adding a hand-copied list in permuted order passes 19/19; re-adding it as a five-element subset (dropping `imessage-daily`) passes 19/19 despite real behavior drift; and reverting jobs.ts's runtime filter to four elements passes 19/19 AND `bun run typecheck` exits 0 while silently dropping requested types. Only a byte-exact canonical-order re-add trips it. That falsifies the PR body's claim that the fix cannot silently regress, which is the stated justification for this being a fix rather than a cosmetic refactor. Suggested: assert set-equality/derivation (or a length/subset check) instead of a literal regex, `Object.freeze()` the exported constant, and consider pulling `src/core/cycle/conversation-facts-backfill.ts` into the guarded set. Routing to human review.

Masashi-Ono0611:
Pushed 992f56d1 implementing all three suggestions, and re-ran the three mutations from the triage to confirm the guard now catches them:

- The two byte-exact regexes are replaced by a semantic detector (`findSuspectTypeLists()`): it scans array literals and string-literal union chains and flags any whose elements intersect the canonical set in ≥2 places, order- and length-insensitive. jobs.ts additionally gets a wiring assertion that its runtime filter references `ALLOWED_TYPES`. Mutation results at this head: permuted 6-element re-add → 1 test fails; 5-element subset re-add (dropping `imessage-daily`) → 2 tests fail; jobs.ts filter narrowed to 4 elements → 2 tests fail (typecheck alone still exits 0, but the wiring test now closes that hole).
- `ALLOWED_TYPES` is `Object.freeze()`d, with a test asserting `isFrozen` and that `push` throws; the `as const` type surface is preserved (compile-time probe: string→AllowedType assignment still rejected, `ALLOWED_TYPES.length` statically 6).
- `src/core/cycle/conversation-facts-backfill.ts` is pulled into the guarded set — and running the new detector over it immediately caught a live instance of the bug class: its header comment still documented the `cycle.conversation_facts_backfill.types` default as the stale pre-#2756 four-element list while the code default is `[...ALLOWED_TYPES]` (six). Fixed in the same commit, and its import now comes from the leaf module rather than the command-module re-export.

118 pass / 0 fail across the six affected test files, typecheck clean.

garrytan:

Masashi-Ono0611:
Closing — this landed already. Sub-commit \`0ee6e886\` ("fix(facts): conversation-type allowlist becomes single-source instead of 5 hand-copied lists") shipped in the v0.46.14.0 fix-wave (\`864dec4f\`, #4226), explicitly credited: "Adopted from PR #4135 by @Masashi-Ono0611 (pinned patch, applied via 3-way merge)". This includes the follow-up hardening (\`findSuspectTypeLists\`, \`Object.isFrozen\` check, backfill coverage) requested in review and pushed as \`992f56d1\` — that's also on master. You'd noted on this PR you'd close it once the wave landed; looks like that got missed. Confirmed while attempting a rebase today: \`src/core/facts/conversation-types.ts\` and every consumer (doctor.ts/jobs.ts/sources.ts/extract-conversation-facts.ts) already match this PR's design on \`upstream/master\`. No further action needed on this PR.

Issue #4140, opened 2026-08-15
AbortController drops the reason when abort() runs in a timer-scheduled microtask — the lock-steal flake is the symptom, cycle.ts:2718 is the consequence


Masashi-Ono0611:
## The correctness gap

`src/core/cycle.ts:2718` decides how a lock steal is reported:

```ts
const stolenFired = stolen?.signal.aborted === true
  && stolen.signal.reason instanceof LockStolenError   // <-- this
  && externalSignal?.aborted !== true;
if (stolenFired) { lockStolenAbort = true; … } else { throw e; }
```

That middle term can be `false` even on a genuine steal, because the runtime intermittently drops
the abort reason (evidence below). When it does, the branch **inverts**: instead of the structured
`reason: 'lock_stolen'` partial report, `runCycle` takes the `else` and rethrows — handing
`jobs.ts` / autopilot exactly the exception-classification burden the comment above that block says
this design removes. The feature does not degrade; on that path it presents as not existing, with
no error pointing at why.

Two more sites read the same value: `:1906` and `:1909` (`raceStolen`) reject with
`stolen.signal.reason`, i.e. reject with `undefined` when it is dropped.

I have measured the runtime behavior only. I have NOT observed a production lock steal hitting
this path — the consequence above is read off the code, not from an incident.

The visible symptom today is a test: `test/db-lock-fencing.test.ts` →
`startCycleLockRefresher … aborts the controller with LockStolenError when a fenced refresh returns
false` fails roughly 1 run in 20 for the whole file on unmodified `master`, and has already failed
in CI (`test (8)` on
[run 31867462707](https://github.com/garrytan/gbrain/actions/runs/31867462707/job/94970717302)), so
it is not macOS-only. That run was on an unrelated PR of mine whose diff touches neither the test
nor `cycle.ts`.

## The underlying behavior

`controller.abort(reason)` lands with `signal.aborted === true` but `signal.reason === undefined`,
when the `abort()` call happens inside a promise continuation scheduled from a timer callback —
exactly the shape `startCycleLockRefresher` uses.

Plain Bun, no gbrain code. `MyErr` stands in for `LockStolenError`; the structure mirrors
`setInterval` → `void (async () => …)()` → `abort(new …)`.

```ts
class MyErr extends Error {}
let lost = 0;
for (let i = 0; i < 400; i++) {
  const c = new AbortController();
  const t = setInterval(() => {
    if (c.signal.aborted) return;
    void (async () => {
[...]

PR #4141, opened 2026-08-15
fix(cycle): key the lock-steal decision on the aborted flag, not the abort reason

Masashi-Ono0611:
## What this fixes

`runCycle` decided whether a lock steal reports a structured partial or rethrows with:

```ts
const stolenFired = stolen?.signal.aborted === true
  && stolen.signal.reason instanceof LockStolenError   // <-- this
  && externalSignal?.aborted !== true;
if (stolenFired) { … } else { throw e; }
```

The middle term is not load-bearing. The `stolen` controller never leaves `runCycle`, and the only
two things that can abort it — `startCycleLockRefresher` and the `onStolen` callback — both pass a
`LockStolenError`. So the aborted flag already *is* the steal signal; re-deriving the answer from
the reason only adds a way to get it wrong.

It does get it wrong. The runtime intermittently leaves `signal.reason` undefined once an abort has
settled, for aborts raised from a microtask continuation scheduled by a timer callback — exactly the
refresher's shape. The branch then **inverts**: a genuine steal takes the `else` and rethrows,
handing `jobs.ts` / autopilot the exception-classification burden the `lock_stolen` partial exists
to remove. It fails on the low-frequency path and presents as the feature not existing rather than
as an error.

Details and a reproduction that uses no gbrain code are in #4140.

## The change

`isLockStolenAbort(stolen, external)` keys on the aborted flag and preserves the external-abort
conjunct verbatim. The reason is still preferred for the log message when it survived — now guarded
rather than dereferenced, since the old `instanceof` gate was the only thing making that
dereference safe. `raceStolen`'s two rejections fall back to a constructed `LockStolenError` so a
raced phase can never reject with `undefined`.

**Exactly one input's classification changes.** Evaluating the old expression and the new predicate
on identical inputs:

| input | old (master) | new |
|---|---|---|
| steal, reason **dropped** | `false` | **`true`** |
| steal, reason intact | `true` | `true` |
| no steal | `false` | `false` |
| steal + external abort | `false` | `false` |
| steal dropped + external abort | `false` | `false` |

The predicate takes structural `{aborted, reason}` types rather than `AbortSignal`. That is not
incidental: `abort()` and `abort(undefined)` both produce a `DOMException`, so `reason === undefined`
is unreachable through `AbortController` and the regression is otherwise untestable except by
waiting for the flake. Duck-typed signals are already the documented convention here — see
`anyAbortSignal`'s comment about test stubs.

## The test flake

`test/db-lock-fencing.test.ts`'s refresher test asserted `signal.reason` after its poll loop, which
is the one property the runtime does not guarantee; it failed roughly 1 run in 20 and has turned CI
red on unrelated PRs. It now captures the reason **in the abort listener** and asserts on that, so
the producer assertion is kept rather than weakened — the refresher must still hand over a real
`LockStolenError`, not merely abort.

That shape came out of review, and it works for a measurable reason. In one process, interleaved:

| | late read `undefined` |
|---|---|
| no abort listener attached | 41/300, then 49/300 |
| with an abort listener attached | **0/300** (captured value was always a `LockStolenError`) |

Attaching a listener appears to keep the reason from being dropped. That is also why the production
fix does not lean on it: whether a listener happens to be attached to `stolen.signal` varies —
`anyAbortSignal` attaches one only when the caller passed a signal, and `raceStolen` only during a
raced phase.

## Verification

`test/db-lock-fencing.test.ts` 15 pass / 0 fail, and 10 consecutive runs with 0 failures (it was
~1-in-20 before). `test/cycle-abort.test.ts` + `test/cycle-any-abort-signal.test.ts` 15 pass;
`test/cycle-lock-steal.serial.test.ts` 2 pass; `bun run typecheck` and `bun run verify` (43/43) both
exit 0. The source-string guards in `cycle-abort.test.ts` still hold — 21 `checkAborted(cycleSignal)`
occurrences and the literal `anyAbortSignal([externalSignal, stolen.signal])` are untouched.

On red-green, precisely: reverting only `src/core/cycle.ts` produces a module-load error (the export
is new), not an assertion failure, so it demonstrates that the symbol is new rather than that the
logic changed. The behavioral evidence is the table above, evaluated on the same inputs.

## Out of scope

- Diagnosing the runtime behavior itself; #4140 asks that question and deliberately does not answer it.
- `checkAborted` (`cycle.ts:912-919`) also consumes a reason and degrades it to a generic `Error`.
  Pre-existing, already non-crashing, untouched here.
- `CycleReport.reason`'s doc comment does not list `'lock_stolen'` even though the code sets it —
  a doc/code drift from #4127. Flagging rather than folding in.
- Whether the steal should travel in `signal.reason` at all is the maintainer's design call. This
  change only stops the branch from inverting when it does not.

### DIFF
--- a/src/core/cycle.ts
+++ b/src/core/cycle.ts
@@ -605,6 +605,26 @@ export function cycleLockIdFor(sourceId?: string): string {
 }
 
 /**
+ * Non-throwing companion to `cycleLockIdFor`, for LOG LABELS only (the
+ * LockStolenError messages + the steal abort log line in runCycle).
+ *
+ * runCycle needs the label on paths that never validate `opts.sourceId`:
+ * lock-free phase selections acquire no lock at all (e.g.
+ * `gbrain dream --phase orphans --source __all__` — the `__all__` sentinel
+ * deliberately fails strict validation), and the engine-null file-lock path
+ * never routes the sourceId through `acquireDbCycleLock`. Throwing there
+ * would crash a run that never needed a lock id. NEVER use this for an
+ * actual lock acquisition — `cycleLockIdFor`'s throw IS the defense there.
+ */
+function cycleLockIdLabelFor(sourceId?: string): string {
+  try {
+    return cycleLockIdFor(sourceId);
+  } catch {
+    return `${LEGACY_CYCLE_LOCK_ID}:${String(sourceId)}`;
+  }
+}
+
+/**
  * Acquire the DB-backed cycle lock for a given source.
  *
  * Pre-v0.38 this file had its own copy of the UPSERT-with-TTL SQL for both
@@ -1931,7 +1951,11 @@ export async function runCycle(
   const cycleSignal: AbortSignal | undefined = combinedSignal
     ? combinedSignal.signal
     : (stolen?.signal ?? externalSignal);
-  const cycleLockId = cycleLockIdFor(opts.sourceId);
+  // Label only — the non-throwing variant, because this line runs even for
+  // lock-free phase selections where opts.sourceId was never validated (a
+  // `cycleLockIdFor` throw here crashed `--source __all__` runs that never
+  // needed a lock id). Real acquisition validated above via acquireDbCycleLock.
+  const cycleLockId = cycleLockIdLabelFor(opts.sourceId);
   const stopRefresher: (() => void) | undefined = lock && stolen
     ? startCycleLockRefresher(lock, stolen, cycleLockId)
     : undefined;
--- a/src/core/cycle.ts
+++ b/src/core/cycle.ts
@@ -871,6 +871,41 @@ function resolveCycleLockRefreshMs(): number {
   return CYCLE_LOCK_REFRESH_INTERVAL_MS;
 }
 
+/**
+ * Did runCycle's private steal controller fire, and is this a steal rather than
+ * an external abort?
+ *
+ * Keyed on the ABORTED FLAG, not on `reason instanceof LockStolenError`. The
+ * `stolen` controller never leaves runCycle: the only two things that can abort
+ * it are startCycleLockRefresher (which passes `new LockStolenError(lockId)`)
+ * and the `onStolen` callback (typed to take a LockStolenError). So its aborted
+ * flag IS the steal signal, and re-deriving that answer from the reason only
+ * adds a way to get it wrong.
+ *
+ * It does get it wrong: the runtime can deliver `aborted === true` with
+ * `reason === undefined` when abort() runs in a microtask continuation
+ * scheduled from a timer callback — exactly startCycleLockRefresher's shape.
+ * Gating on the reason then INVERTS this branch, so a real steal rethrows
+ * instead of reporting the structured `lock_stolen` partial that exists to
+ * spare daemon callers that classification.
+ *
+ * The external-abort conjunct is preserved: a caller-initiated abort keeps the
+ * throw-out contract even if a steal races it.
+ *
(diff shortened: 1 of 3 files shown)


time-attack:
Automated triage: the core lock-steal fix reads sound (keying on the private controller's aborted flag is safe — every abort of that controller carries LockStolenError, and the external-signal conjunct still routes external cancellation to the throw path), and both raceStolen sites were fixed consistently. However, adversarial review found a real regression introduced by an incidental refactor: hoisting `cycleLockIdFor(opts.sourceId)` to unconditional evaluation (~cycle.ts:1934) means `--source __all__` / `GBRAIN_SOURCE=__all__` on lock-free phase selections (e.g. `gbrain dream --phase orphans --source __all__`) now throws an uncaught `Invalid source_id` crash — verified by running the identical input against master (clean) and this head (throws). Suggested fix: compute the lock id lazily, only in the `lock && stolen` branch and the catch-block fallback, or use a non-throwing formatter for the log string. Routing to human review with this finding attached.

Masashi-Ono0611:
Fixed in 9ff4c67b. The regression reproduced exactly as described — a test at the pre-fix head fails with `Invalid source_id: "__all__"` thrown from `cycleLockIdFor` via the hoisted call at cycle.ts:1934. I took the non-throwing-formatter option: a private `cycleLockIdLabelFor()` now feeds the three consumers of that value (the lock-refresher label, the `stolenReason()` fallback, and the catch-block log), all of which are message-only — `LockStolenError.lockId` is never dispatched on — while real acquisition (`acquireDbCycleLock` → `cycleLockIdFor`) keeps its throwing validation unchanged. This also keeps the engine-null file-lock path (which never validates `opts.sourceId`) crash-free, which pure laziness alone would not have. Added a regression test: `__all__` + a lock-free phase selection (`orphans`) completes without throwing. 33 pass / 0 fail across the four lock/abort cycle test files, typecheck clean.

Masashi-Ono0611:
One more data point on how often this bites, from an unrelated PR.

PR #3720 (`config set` unknown-flag validation — touches nothing in the cycle or
lock paths) has been sitting on a red `test (8)` shard. The failure is this one:

```
(fail) startCycleLockRefresher (Tier-1 #1 + D5.11) >
       aborts the controller with LockStolenError when a fenced refresh returns false
Expected constructor: [class LockStolenError extends Error]
Received value: undefined
```

On that PR's tree merged with current master, `bun test test/db-lock-fencing.test.ts`
run three times in a row gave **fail / fail / pass**.

So the dropped `reason` is not only a self-inflicted flake inside the lock work —
it currently reads as a red CI shard on contributor PRs that have no relationship
to it, which is the expensive failure mode (the natural read is "my change broke
something", and the natural response is a blind re-run).

No change requested here; this PR already fixes it. Recording it because it
argues for landing sooner rather than later.

garrytan:

Masashi-Ono0611:
Closing — this landed already. Commit \`af4d7d64\` ("fix(cycle): key the lock-steal decision on the aborted flag, not the abort reason") shipped in the v0.46.14.0 fix-wave (\`864dec4f\`, #4226), explicitly credited: "Adopted from PR #4141 by @Masashi-Ono0611 (pinned patch)", closing #4140. You'd noted on this PR you'd close it once the wave landed — looks like that got missed. Confirmed \`isLockStolenAbort()\`/\`cycleLockIdLabelFor()\`/the stolen-reason fallback are byte-identical between this branch and current \`upstream/master\` while attempting a rebase today. No further action needed on this PR.

PR #4144, opened 2026-08-15
v0.45.18.1 fix(sync): accept Windows path casing and index Astro/Svelte (#4044)

### LINKED ISSUE #4044: Windows path casing rejects valid syncs; .astro/.svelte silently unindexed
Two independent defects found running gbrain 0.44.0.0 on Windows across a 12-source brain.

## 1. Case-sensitive path comparison refuses valid syncs (regression in 0.44.0.0)

After upgrading, every markdown source failed with:

```
Sync scope C:\Users\s_lak\projects\Tally resolves outside git repo
C:\Users\s_lak\Projects\Tally. Refusing to sync: possible path traversal
via --src-subpath.
```

Those are the same directory. Windows is case-insensitive; the traversal guard
compares the source's stored `local_path` against the git-discovered root as raw
strings, so `projects` != `Projects` reads as traversal.

**Scope:** 7 of 12 sources — every one registered by hand before the orchestrator
started writing canonical casing. The 5 written by the orchestrator were fine,
which made it look like a single-repo problem for some time. A nightly scheduled
sync was also failing this way with no one watching.

**Suggested fix:** normalise both sides before comparing — `fs.realpathSync.native()`,
or case-fold on `win32`.

**Workaround** (no `sources update`/repoint verb exists, and `sources remove`
destroys the source's pages): correct `local_path` directly in the store.

## 2. `.astro` and `.svelte` files are never indexed

`detectCodeLanguage` in `src/core/chunkers/code.ts` maps 35 extensions including
`.vue`, but omits `.astro` and `.svelte`. Both return `null`, so those files are
skipped entirely.

**Effect**, measured on two repos:

| repo | framework | pages indexed | component files indexed |
|---|---|---|---|
| Astro site | Astro | 20 | 0 of 15 `.astro` |
| Mail client | SvelteKit | 142 | 0 of 14 `.svelte` |

Both sources sync clean and report 100% embed coverage. Code search answers
confidently — from config, tests and scripts, which is all that parsed. `.tsx`
indexes fine (15 and 27 pages on two other repos), which is why this went
unnoticed for weeks. A search that silently omits the product is harder to catch
than one that errors, because the caller cannot tell the answer is partial.

**On the fix:** `tree-sitter-wasms` ships no svelte or astro grammar, so proper
support is larger work. But mapping both to the already-bundled `html` grammar
would be two lines and turns *invisible* into *searchable*. Symbol extraction
inside `<script>` blocks would be weak; that seems a fair trade against zero
coverage, and matches how `.vue` is already handled.

---

Happy to open a PR for either if the suggested direction looks right.

javieraldape:
## Summary

Fixes #4044.

This tightens two ingest/indexing edges from the issue:

- sync path containment now accepts Windows paths that resolve to the same directory even when the stored source path and git-discovered root differ only by casing, while still rejecting sibling escapes
- Astro and Svelte files are classified as code by routing them through the existing HTML tree-sitter grammar, so they become indexed instead of being silently skipped
- markdown code-reference extraction now recognizes .astro and .svelte paths to stay aligned with the code classifier

## Tests

- `bun test test/sync-path-containment.test.ts test/chunkers/code.test.ts test/sync-classifier-widening.test.ts test/link-extraction-code-refs.test.ts` - 71 pass
- `env -u DATABASE_URL -u GBRAIN_DATABASE_URL GBRAIN_TEST_ALLOW_DATABASE_URL=1 bash scripts/gbrain-gate.sh <worktree> test/sync-path-containment.test.ts test/chunkers/code.test.ts test/sync-classifier-widening.test.ts test/link-extraction-code-refs.test.ts` - RESULT: GREEN; verify green, 71 focused unit tests pass, 72 mapped e2e tests pass

### DIFF
--- a/src/commands/sync.ts
+++ b/src/commands/sync.ts
@@ -1246,6 +1246,10 @@ export function createSyncBaselineCommit(repoPath: string): void {
  * returns backslash paths there, and a literal `rootReal + '/'` prefix can
  * never match one. A sibling (`root-evil`) is rejected because `relative`
  * yields `../root-evil`, and a cross-drive path because it yields an absolute.
+ * On Windows this intentionally follows the platform path semantics exposed by
+ * `win32.relative`, including case-insensitive segment matching; detecting rare
+ * per-directory case-sensitive NTFS siblings would require filesystem identity
+ * checks rather than a pure path check.
  */
 interface PathContainmentOps {
   relative(from: string, to: string): string;
--- a/src/commands/sync.ts
+++ b/src/commands/sync.ts
@@ -1247,10 +1247,23 @@ export function createSyncBaselineCommit(repoPath: string): void {
  * never match one. A sibling (`root-evil`) is rejected because `relative`
  * yields `../root-evil`, and a cross-drive path because it yields an absolute.
  */
-export function isWithinRoot(childReal: string, rootReal: string): boolean {
+interface PathContainmentOps {
+  relative(from: string, to: string): string;
+  isAbsolute(path: string): boolean;
+  sep: string;
+}
+
+const nativePathContainmentOps: PathContainmentOps = { relative, isAbsolute, sep };
+
+export function isWithinRoot(
+  childReal: string,
+  rootReal: string,
+  pathOps: PathContainmentOps = nativePathContainmentOps,
+): boolean {
   if (childReal === rootReal) return true;
-  const rel = relative(rootReal, childReal);
-  return rel !== '' && rel !== '..' && !rel.startsWith('..' + sep) && !isAbsolute(rel);
+  const rel = pathOps.relative(rootReal, childReal);
+  if (rel === '') return true;
+  return rel !== '..' && !rel.startsWith('..' + pathOps.sep) && !pathOps.isAbsolute(rel);
 }
 
 /**
--- a/src/core/chunkers/code.ts
+++ b/src/core/chunkers/code.ts
@@ -464,6 +464,7 @@ export function detectCodeLanguage(filePath: string, content?: string): Supporte
   if (lower.endsWith('.sh') || lower.endsWith('.bash')) return 'bash';
   if (lower.endsWith('.css')) return 'css';
   if (lower.endsWith('.html') || lower.endsWith('.htm')) return 'html';
+  if (lower.endsWith('.astro') || lower.endsWith('.svelte')) return 'html';
   if (lower.endsWith('.vue')) return 'vue';
   if (lower.endsWith('.json')) return 'json';
   if (lower.endsWith('.yaml') || lower.endsWith('.yml')) return 'yaml';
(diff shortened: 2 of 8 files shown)


MohammedAlkindi (commented):
Ran this on Windows 11 (10.0.26200), bun 1.3.14, against `main` at `4922905`.

**Tests:** the four files this PR touches are **71 pass / 0 fail** on the branch; the three that exist on `main` are **49 pass / 0 fail** there. No regressions, and the injectable `PathContainmentOps` is a nice touch — it means the win32 behaviour is testable from a Linux CI runner, which is exactly the class of thing that otherwise only shows up on a user's machine.

**The core change is right, and it fixes more than the description claims.** The semantic move is `rel === ''` going from *reject* to *accept*. I checked what actually produces `''`:

```
win32.relative(C:\Users\p\Projects\Tally, C:\Users\p\projects\Tally)  ->  ""     casing only
win32.relative(C:\a\b\,                   C:\a\b)                     ->  ""     trailing separator
win32.relative(C:\a\root,                 C:\a\root-evil)             ->  "..\root-evil"
win32.relative(C:\a\root,                 C:\a)                       ->  ".."
win32.relative(C:\a,                      D:\a)                       ->  "D:\a"
posix.relative(/a/b/,                     /a/b)                       ->  ""     trailing separator
posix.relative(/a/B,                      /a/b)                       ->  "../b" POSIX stays case-SENSITIVE
```

So the escapes all still reject — sibling, parent and cross-drive are unaffected. And `''` also covers a **trailing-separator** mismatch on *both* platforms, which the old code rejected too. That's a second real fix riding along; might be worth a line in the description or a test, since it isn't obvious from "Windows path casing" that `/a/b/` vs `/a/b` was also failing.

**One edge I'd want your read on, because this is a containment check.** `win32.relative` is *unconditionally* case-insensitive, but NTFS case sensitivity is per-directory since Windows 10 1803 (`fsutil file setCaseSensitiveInfo`), and directories created from WSL frequently have it set. In such a directory `C:\a\Foo` and `C:\a\foo` are genuinely two different directories — and `isWithinRoot` would now answer `true` for a child that is actually outside the root.

Concretely: root `C:\proj\Data` with a sibling real directory `C:\proj\data` on a case-sensitive-flagged parent. Pre-change that returned `false`; post-change it returns `true`.

I have **not** reproduced this — setting the case-sensitivity flag needs an elevated `fsutil`, which I didn't run — so treat it as a question rather than a defect. It may well be out of scope: the flag is rare, and the alternative (stat-ing both paths and comparing file IDs) is a much heavier check for a hot path. But since this function's whole job is rejecting escapes, it seemed worth naming explicitly rather than leaving implied.

**Small note on the diff itself:** `if (childReal === rootReal) return true;` is now redundant — the `rel === ''` branch below covers identical paths on both platforms. Harmless as a fast path, and arguably clearer to keep, just flagging that it is no longer load-bearing.

---

*Disclosure: I used an AI assistant while reviewing. The bun runs and the `path.relative` table above were executed by me on the Windows machine described. The NTFS per-directory case-sensitivity edge is reasoned from the diff and explicitly not reproduced, which is why it's phrased as a question.*

garrytan:

time-attack:
Thanks @javieraldape — adopted with credit in the v0.46.14.0 wave (#4226, commit 864dec4f): Windows path-casing containment and .astro/.svelte indexing are on master. Closing as superseded; please open fresh PRs against latest master if anything from #4044 remains.

PR #4150, opened 2026-08-15
fix(transcripts): remove 512MB hard cap on Hermes store import

### LINKED ISSUE #4149: transcripts ingest: Hermes store 512MB cap blocks multi-profile clusters
## Problem

`gbrain transcripts ingest` imposes a **512 MB hard cap** on Hermes SQLite stores (`state.db`):

```
error: /path/to/state.db: hermes store too large for import: 1894327792 bytes incl. sidecars (cap 536870912)
```

In multi-profile Hermes setups (common for agent clusters), each profile maintains its own `state.db`. On a typical cluster with 5 active profiles:

| Profile | state.db size | Importable? |
|---------|--------------|-------------|
| herme (default) | 1.5 GB | ❌ blocked |
| agent_arch | 1.8 GB | ❌ blocked |
| aevum | 571 MB | ❌ blocked |
| reviewer | 267 MB | ✅ OK |
| researcher | 133 MB | ✅ OK |

**3 out of 5 profiles are blocked**, and these are precisely the most active ones (herme handles daily user interaction, agent_arch manages cluster governance, aevum handles enterprise tasks).

## Why the cap exists (hypothesis)

The cap likely protects against accidentally loading a massive store into memory. However, the adapter already processes sessions incrementally (`session-atomic` per the v0.46.0.0 release notes), so the total file size should not be a memory concern.

## What the Hermes store contains

Each `state.db` is a SQLite database with:
- `sessions` table: session metadata (model, token counts, timestamps)
- `messages` table: full conversation turns including **tool call results** (32,479 tool messages in the default profile alone)
- FTS indexes for search

The tool results (file contents, search results, API responses) represent significant accumulated knowledge that is currently invisible to the brain.

## Suggested fix

1. **Remove or significantly raise the cap** (e.g., to 4 GB) — the adapter already reads incrementally, so file size is not a memory issue
2. Alternatively, add a `--store-cap` flag so users can opt into larger stores
3. Or add `--limit N` support for session-count-based batching as a workaround

## Current workaround

We use a custom `brain-agent-loop` hook that exports Hermes sessions as `.txt` files, which gbrain autopilot then consumes via the dream pipeline. This works but:
- Requires maintaining custom hook code
- Produces less structured output (plain text vs. structured session pages)
- Loses tool call result context that the native adapter would preserve

## Environment

- gbrain: v0.46.0.0 (upgraded today)
- Hermes Agent: multi-profile cluster on Linux x64
- Engine: Postgres (remote)
- 5 active profiles, 3 exceeding 512 MB

justemu:
## Summary

Remove the 512MB hard cap (`HERMES_DB_HARD_CAP`) on Hermes store files during transcript import. The default is now `Infinity` (no limit).

## Problem

The Hermes adapter rejected stores larger than 512MB (including WAL/SHM sidecars). In multi-profile Hermes setups, this blocks the most active profiles:

| Profile | state.db size | Importable |
|---------|--------------|------------|
| herme (default) | 1.5 GB | ❌ blocked |
| agent_arch | 1.8 GB | ❌ blocked |
| aevum | 571 MB | ❌ blocked |
| reviewer | 267 MB | ✅ |
| researcher | 133 MB | ✅ |

The tool call results stored in these databases (32,479+ tool messages per profile) represent accumulated knowledge that is currently invisible to the brain.

## Why the cap is unnecessary

The adapter reads sessions **incrementally** via SQLite queries — it does not load the entire store into memory. The only real constraint is temp disk space for the copy-then-read pattern, which is:
- Inherited from the OS (typically many GB)
- Already bounded by the `maxBytes` opt-in via `ParseSessionsOpts`
- Not meaningfully different at 512MB vs 2GB

## Changes

- `src/core/transcripts/hermes.ts`: Set `HERMES_DB_HARD_CAP` from `512 * 1024 * 1024` to `Infinity`
- Updated comments to reflect the new default and explain the `maxBytes` opt-in
- Enhanced error message to guide users toward `maxBytes` when a cap is needed

## Backward compatibility

- Users who never hit the 512MB cap see no behavior change
- Users who need a cap can still pass `maxBytes` via `ParseSessionsOpts`
- The `HERMES_DB_HARD_CAP` export is preserved (value changed, not removed)

Closes #4149

### DIFF
--- a/src/core/transcripts/hermes.ts
+++ b/src/core/transcripts/hermes.ts
@@ -49,8 +49,8 @@ export const HERMES_SPEC_TARGET: HostSpecTarget = {
     'context window). PROVISIONAL: no populated production sample verified.',
 };
 
-/** Hard cap for the store copy (FTS indexes make legitimate stores large). */
-export const HERMES_DB_HARD_CAP = 512 * 1024 * 1024;
+/** Default cap for the store copy — Infinity removes the limit. Pass maxBytes via opts to override. */
+export const HERMES_DB_HARD_CAP = Infinity;
 
 const SQLITE_MAGIC = 'SQLite format 3\u0000';
 
@@ -113,16 +113,16 @@ export const hermesAdapter: TranscriptAdapter = {
   async *parse(path: string, opts: ParseSessionsOpts = {}): AsyncGenerator<ParsedSession, FileDiagnostics> {
     const cap = opts.maxBytes ?? HERMES_DB_HARD_CAP;
     const size = statSync(path).size;
-    // The cap bounds the TOTAL copied (db + sidecars) — a runaway WAL can
-    // dwarf the main file, and only capping the db would let the copy blow
-    // through temp storage while advertising a 512MB bound.
+    // The cap bounds the TOTAL copied (db + sidecars). Default is Infinity
+    // (no limit) since the adapter reads sessions incrementally. Pass
+    // maxBytes via opts to cap when temp disk space is constrained.
     let totalBytes = size;
     for (const suffix of ['-wal', '-shm']) {
       if (existsSync(path + suffix)) totalBytes += statSync(path + suffix).size;
     }
     if (totalBytes > cap) {
       throw new Error(
-        `hermes store too large for import: ${totalBytes} bytes incl. sidecars (cap ${cap})`,
+        `hermes store too large for import: ${totalBytes} bytes incl. sidecars (cap ${cap}) — pass a higher maxBytes or omit the cap`,
       );
     }
 


time-attack:
Automated triage: **routing to maintainer** — this is a policy call, not a bug fix.

The change flips `HERMES_DB_HARD_CAP` from `512 * 1024 * 1024` to `Infinity`, so the default becomes unbounded and callers must opt *in* to a cap via `opts.maxBytes`.

Why it isn't auto-mergeable:

1. **It removes a safety limit rather than fixing incorrect behavior.** The original comment says the cap bounds the *total* copied (db + `-wal` + `-shm`) so a runaway WAL can't blow through temp storage. Defaulting to `Infinity` reinstates exactly that failure mode for every caller that doesn't pass `maxBytes` — an unbounded temp-disk copy, which fails worse (host out of disk) than the current explicit error.
2. **No test accompanies the production change.** Nothing pins the new default or the opt-in path.
3. The new error text ("pass a higher maxBytes or omit the cap") is unreachable via the default now, since the default can't be exceeded.

The underlying report (#4149, multi-profile clusters over 512MB) is legitimate — a bigger default, or a size-aware streaming path that doesn't copy sidecars wholesale, may be the better shape than removing the bound. Deferring the judgment call to you.

garrytan (changes requested):
Thanks — the block is real and your diagnosis of the incremental read path is correct, but Infinity-as-default removes the temp-disk protection the cap exists for (copy-then-read duplicates db + WAL/SHM into tmpdir, which is often RAM-backed tmpfs; a runaway WAL can dwarf the main file).

Requested changes:
1. Thread the cap instead of deleting it — add a maxStoreBytes/--store-cap option through TranscriptsIngestOpts and pass { maxBytes } at ingest.ts:197. Today ParseSessionsOpts.maxBytes is unreachable from the CLI, so this PR's backward-compat path and new error-message advice can't actually help `gbrain transcripts ingest` users.
2. Raising the default (e.g. to 4GB) is fine — but keep it finite.
3. Add a unit test for the over-cap rejection and the override path.

Heads-up: the current fix wave lands the cap-threading half (optional validated override, per-adapter defaults preserved, checkpoint fingerprinting) with credit to your report on #4149 — happy to have this PR rebase onto it for the default-raise discussion.

garrytan:
Thanks @justemu — the multi-profile size table made the need concrete, and that need is legitimate. But defaulting `HERMES_DB_HARD_CAP` to Infinity removes a deliberate temp-disk safety bound (the cap covers the db plus runaway WAL/SHM sidecars in the copy-then-read path) with no tests, and it makes the new "pass a higher maxBytes" error text unreachable. The underlying ask (#4149, closed completed) was solved the bounded way in v0.46.14.0: `gbrain transcripts ingest --max-bytes 4gb <store>` is a validated operator escape hatch while the safe default stays. Closing on that basis — if you hit a store the escape hatch can't handle, please file it and we'll revisit the ceiling.

Issue #4155, opened 2026-08-15
claude-cli subagent jobs die on uniq_subagent_tools_use_id — the model mints the tool_use_id and cannot keep it unique across turns


Masashi-Ono0611:
## Symptom

The dream cycle's `patterns` phase produces zero pattern pages when `models.dream.patterns` (or the
resolved subagent model) is a `claude-cli:*` model. The subagent job runs, does real work, then dies:

```
minion_jobs id=… name=subagent status=dead attempts_made=3 max_attempts=3
error_text: duplicate key value violates unique constraint "uniq_subagent_tools_use_id"
```

```
[cycle] phase 'patterns' ran 262s, exceeding the 30s worker force-evict deadline
  ✗ patterns    pattern-detection subagent job … ended 'dead'; nothing was written
      [InternalError/PATTERNS_CHILD_DEAD] subagent job … outcome 'dead' with zero pattern pages written
```

Three retries, all the same, then dead. On the previous nights, with the same phase on an OpenAI
model, it produced pattern pages normally.

## What is actually colliding

`claude-cli` tool calls arrive with short, repeating ids. From `subagent_tool_executions` on a live
brain, grouped by provider:

| `provider_id` | `tool_use_id` | rows | distinct ids |
|---|---|---|---|
| `claude-cli` | `toolu_01` | 6 | 1 |
| `claude-cli` | `toolu_02` | 6 | 1 |
| `claude-cli` | `toolu_03` | 2 | 1 |
| `anthropic` | `toolu_015RAe…` | 1 | 1 |
| `anthropic` | `toolu_015cWm…` | 1 | 1 |

Every `anthropic` row carries a distinct full-length id. Every `claude-cli` row carries one of three
short ones. The rows written by the job that died:

```
job_id | provider_id | tool_name        | tool_use_id  | status
  …    | claude-cli  | brain_list_pages | toolu_01     | complete
  …    | claude-cli  | brain_get_page   | toolu_check1 | complete
  …    | claude-cli  | brain_search     | toolu_02     | complete
```

## Why they repeat

There is no counter in gbrain minting these — the model mints them, and the transport makes it
impossible for it to do so reliably.

`src/core/ai/providers/claude-cli-language-model.ts:113` instructs the model to invent the id, by
example:

```
'{"id": "<unique tool call id, like toolu_01ABC>", "name": ..., "input": ...}'
```

`claude-cli:` is not the Anthropic API — it shells out to `claude --print` as a **fresh subprocess
[...]

Issue #4152, opened 2026-08-15
Dream synth: two-stage cascade — cheap-model triage gates frontier-model synthesis


garrytan-agents:
## Problem

Dream synth-v2 currently fans out one **Opus-5, 30-turn subagent per file** in every registered corpus, unconditionally. Real numbers from a live deployment (2026-08-15):

- Aug 12 → corpus `straylight-brain`: 1,216 jobs
- Aug 14 → corpus `camera-roll-2026-08-13`: 1,237 jobs
- Aug 15 → corpus `zion-brain`: 1,257 jobs
- Queue state: **3,798 waiting subagent jobs**, drain ~150/day best case at concurrency 3 (avg job 173s unloaded, ~4,770s under host load). Intake ~1,250/day → backlog grows ~1,100/day, unbounded.

Every job pays full frontier-model price regardless of content. Most conversation transcripts are logistics/small talk with little durable knowledge — the pipeline spends its most expensive model on triage work.

Related: #4145 (eviction of long-running handlers under queue pressure — this proposal removes most of that pressure at the source).

## Proposal: two-stage cascade

**Stage 1 — triage (cheap model, e.g. Haiku-class, 1–2 turns):**
- One triage job per file: score noteworthiness (0–1), classify content type, extract candidate segments (spans/quotes) + entity candidates.
- Persist the triage verdict (score, segments, model, version) so it is auditable and re-runnable when thresholds change.
- Idempotency: `dream:triage-v1:<corpus>:filename:<file>:<hash>`.

**Stage 2 — synthesis (frontier model, existing synth-v2 prompt):**
- Fired **only** by a passing triage verdict (score ≥ threshold, default e.g. 0.5).
- Triage output injected into the synth prompt as a segment map (what's noteworthy and why) — fewer wasted turns hunting signal in raw transcripts, likely lower `max_turns` needed.
- Idempotency unchanged: `dream:synth-v2:...`, but submission gated on triage.

## Why this beats a flat per-source fanout cap

- **Intrinsic throttling:** triage jobs complete in seconds, so the queue drains them at any concurrency; synthesis intake automatically shrinks to match actual signal density instead of corpus size.
- **Cost:** at a plausible 10–25% pass rate, frontier-model spend drops 75–90% while coverage stays 100% (everything is at least triaged).
- **Quality:** the expensive model starts from extracted segments, not thread sludge.

## Design details / knobs

[...]

PR #4156, opened 2026-08-15
fix(minions): keep the gateway loop's tool-execution rows unique when the provider reuses a tool_use_id

Masashi-Ono0611:
Addresses the crash reported in #4155: a `claude-cli` subagent job dies on
`uniq_subagent_tools_use_id` because the model mints its own `tool_use_id` and structurally cannot
keep it unique across turns (fresh `claude --print` subprocess per `doGenerate`, and `renderPrompt()`
strips ids from the replayed context, so it is never shown the ones it already used). What comes
back is the prompt's example with the entropy dropped — `toolu_01`, `toolu_02`, again every turn.

## The change

One line of substance, in the gateway-loop write path only (`onToolCallStart`,
`src/core/minions/handlers/subagent.ts`): the value bound into the constrained `tool_use_id` column
is now

```ts
const storedToolUseId = `${providerToolCallId}#${messageIdx}.${ordinal}`;
```

`(message_idx, ordinal)` is exactly the D11 surrogate key this row already conflicts on
(`ON CONFLICT (job_id, message_idx, ordinal)`), and `message_idx` increases monotonically per job,
so the stored value is unique per job by construction no matter what the provider sends. The
provider's own string stays visible as the prefix, which keeps it useful for the debugging that
`gateway.ts` says it is for.

**Suffixing with `ordinal` alone does not work**, which is worth recording because it is the obvious
first idea: `ordinal` is `callIdx`, the index *within a turn*, and it restarts at `0` on every new
assistant turn. Against a provider that names its first tool call `toolu_01` on every turn, that
yields the identical string `toolu_01#0` each time — the same collision with a new label. That was
caught by reading `gateway.ts:3546/3649/3678` rather than by the tests, since the regression test as
first written happened to use one turn.

No schema change, no migration.

## What this does and does not touch

The legacy Anthropic-direct path (`persistToolExecPending`/`Complete`/`Failed`) is unchanged — it
uses `ON CONFLICT (job_id, tool_use_id)` and depends on the constraint's current semantics. The real
Anthropic API returns globally-unique ids, so it does not hit this.

Two behavioral notes a reviewer should have:

- **`reconcileGatewayReplay`'s `execByKey` exact match now always misses for gateway-loop rows** —
  it builds `${message_idx}:${tool_use_id}` from the stored column and matches against the raw id
  from `content_blocks`. It falls through to the positional fallback
  (`execByMsg.get(message_idx)?.[callIdx]` plus the tool-name guard), which the code already
  anticipates and which does not depend on the id string. Verified by the full
  `subagent-gateway-resume-reconciliation` suite, 7/7.
- **The D5 legacy-row shim does not engage** for these rows: `gbrain_tool_use_id` is always non-null
  for D11 rows, so `stableKey` resolves via `gbrainId` and never reaches the `legacy:` branch.

Adjacent and deliberately left alone: `reconcileGatewayReplay` also calls the legacy persist helpers
with the *raw* provider id (`subagent.ts:1251-1265`), so a crash-heal re-dispatch spanning two turns
that share a raw id could still collide there. That is a pre-existing gap in a different path, not
this defect, and folding it in would widen the diff past what #4155 reports.

## Verification

New regression test in `test/e2e/subagent-gateway-path.test.ts`: two tool calls across **two turns**
both carrying provider id `toolu_01`, asserting both rows persist with distinct stored ids.

Red-green, with the fix committed first and the src then reverted to master:

- **red** — 6 pass / 2 fail. Real failures, not module-load: the updated assertion mismatched
  (`Received: "provider-tc-1"`), and the new test hit the actual database error
  `duplicate key value violates unique constraint "uniq_subagent_tools_use_id"`,
  `Key (job_id, tool_use_id)=(1, toolu_01) already exists.`
- **green** — 8 pass / 0 fail.

Ten-file subagent neighbour sweep (each path confirmed to exist first, reported file count checked
against the count passed): 120 pass / 41 skip / 0 fail. `bun run typecheck` exit 0;
`bun run verify` 44/44 exit 0; `engine-parity` 0 fail — the write goes through one shared
`engine.executeRaw`, so there is no per-engine variant to keep in step.

The two design questions #4155 raises — whether the provider should stop trusting model-authored ids
altogether, and whether `uniq_subagent_tools_use_id` should be retired repo-wide once the legacy loop
moves onto the same ordinal scheme — are deliberately not answered here.


time-attack:
Automated triage: **routing to maintainer** — CI is fully green, but the change looks incomplete on the read side.

The fix stops the `uniq_subagent_tools_use_id` collision by storing a composed id instead of the raw provider one:

```
const storedToolUseId = `${providerToolCallId}#${messageIdx}.${ordinal}`;
```

That does make the stored value unique per job. The problem is that `tool_use_id` is not write-only — several readers still look it up by the **raw provider id**, and those lookups will now miss:

- `src/core/minions/transcript.ts:128` builds `toolById` keyed on the stored `tool_use_id`, then `:180` looks up `toolById.get(block.id)` where `block.id` is the raw provider id. Composed keys never match, so rendered transcripts silently lose every `→ status:` / output line.
- `src/core/minions/transcript.ts:202` — same mismatch makes `tool_result` blocks fall into the "no matching tool_use in this transcript" branch and dump raw.
- `src/core/minions/handlers/subagent.ts:370` — `priorToolByUseId` is keyed the same way and used on the replay path.
- `src/core/minions/handlers/subagent.ts:1257` — `execByKey.get(`${msg.message_idx}:${call.toolCallId}`)` uses the raw id. It does have an ordinal-position fallback on the next line, which likely masks this in the e2e test; that fallback is also gated on `tool_name` matching, so it isn't a general substitute.

Net effect: the dead-letter is fixed, but crash-replay dedupe may fall back to positional matching and transcript rendering degrades — for **all** providers, not just claude-cli, since the composition is unconditional.

Worth considering whether the raw id should be kept in `tool_use_id` with the disambiguation applied to the constraint or a separate column instead. Not merging automatically.

Masashi-Ono0611:
Reworked per the review (d92764b6, then merged with master as ddf2f48d — the migration is renumbered to v131 since #4152/#4145 landed 129/130 first). `tool_use_id` now stores the raw provider id verbatim, and migration v131 drops the job-wide `uniq_subagent_tools_use_id UNIQUE (job_id, tool_use_id)` constraint (also removed from the schema files for fresh installs). Row identity was already carried by the INSERT's own conflict target `(job_id, message_idx, ordinal)` — the job-wide unique encoded the false assumption that provider ids are unique within a job, and a narrower unique (adding message_idx) would still dead-letter a turn that reuses one id twice, so no replacement constraint is added. The legacy `persistToolExec*` helpers that targeted the dropped constraint via `ON CONFLICT (job_id, tool_use_id)` now key by `(job_id, message_idx, tool_use_id)`, stamp the ordinal, and use the stable-id constraint as the conflict backstop; their settlement UPDATEs target exactly one row (own ordinal first, then pending) instead of every row sharing the raw id.

Readers now do message_idx-based canonical matching: the transcript renderer resolves executions by `(message_idx, tool_use_id)` with stamped-ordinal-first resolution (previously the id-keyed map was last-write-wins, so a repeated raw id mis-attributed status/output across turns — the core of your finding), echoed `tool_result` ownership is checked against the nearest preceding assistant turn, and the legacy replay map is keyed the same way, which also stops a repeating provider id from short-circuiting a NEW turn into a prior turn's stored output. `reconcileGatewayReplay`'s `execByKey` was already message_idx-scoped and is correct again with raw ids stored.

Tests: the #4155 regression test now asserts two rows persist with the raw id intact and distinct message_idx; a new transcript regression test pins per-turn attribution when the same raw id appears in two turns (red against the previous reader); `test/subagent-tool-ledger.test.ts` pins the ledger helpers' duplicate-id and settled-winner-downgrade guards; `test/migrations-v131.test.ts` covers fresh-init and upgrade, including a positive control that the duplicate insert throws while the constraint exists.

Two pre-existing gaps found during review are NOT addressed here to keep the diff scoped: the renderer doesn't splice the gateway path's provider-neutral tool-call blocks (it predates this PR), and a lease-expired zombie worker can still double-execute a tool (the original `ON CONFLICT ... DO NOTHING` form had the same property).

Masashi-Ono0611:
Screenshot of the live probe backing the comment above — the exact #4155 dead-letter shape (same job, same raw provider id, two turns) persists 2 rows on a live Postgres brain with the v131 migration applied; the probe rows are deleted afterwards (cleanup leftover = 0).

live probe: duplicate raw id persists 2 rows, cleanup 0

garrytan:
Folding this into the current fix wave with credit (adopted at pinned patch sha256 ec36452a; migration verified free as v131 at HEAD, schema-bootstrap + migrate guards green). Will close this PR when the wave lands.

garrytan:
Thank you @Masashi-Ono0611 — this fix has been absorbed into the v0.46.13.0 fix wave (#4227) with credit in the CHANGELOG. We took your rewritten approach exactly (drop the job-wide constraint via migration v131, keep raw provider ids, rekey readers to (message_idx, tool_use_id)) and extended it with a provider-side always-mint in the claude-cli adapter so no weak-id provider depends on the constraint being gone, plus a mixed-vintage resume test. Closing this PR as superseded by #4227 — the fix ships, authored-in with your credit.

PR #4163, opened 2026-08-15
fix: parse sparse multiline transcript imports

richtheworld:
## Summary

- treat the native `imessage-slack` transcript shape as continuation-aware during format scoring
- keep the existing false-positive guard: continuation scoring only activates after two anchors or when the first line is an anchor
- add a round-trip regression for sparse, multi-line coding turns

## Problem

The native transcript importer preserves embedded newlines. In coding sessions with a few long turns, valid message anchors can make up less than 5% of non-blank lines, so the conversation parser returns `no_match` for pages emitted by GBrain itself.

## Validation

- `bun test test/transcript-render.test.ts test/conversation-parser/parse.test.ts` (92 pass)
- `bun run typecheck`
- `bun run check:conversation-parser` (20/20 fixtures)
- verified an affected imported Codex page changes from `no_match` to `imessage-slack` with 20 messages

### DIFF
--- a/src/core/conversation-parser/builtins.ts
+++ b/src/core/conversation-parser/builtins.ts
@@ -74,7 +74,13 @@ export const BUILTIN_PATTERNS: readonly PatternEntry[] = [
     date_source: 'inline',
     time_format: '12h_ampm',
     timezone_policy: 'inline_utc',
-    multi_line: false,
+    // Transcript imports preserve embedded newlines in each turn. Treat
+    // non-anchor lines as message continuations when scoring so a small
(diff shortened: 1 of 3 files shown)


richtheworld:
@greptileai review

garrytan:

time-attack:
Thanks @richtheworld — adopted with credit in the v0.46.14.0 wave (#4226, commit 864dec4f): the imessage-slack builtin now scores continuations as body, so sparse multiline transcripts parse. Closing as superseded; fresh PRs against latest master are always welcome.

PR #4166, opened 2026-08-15
fix(think): gather stream failures surface typed warning codes instead of stderr-only

Masashi-Ono0611:
## Problem

`runGather` (`src/core/think/gather.ts`) catches per-stream failures in all four
retrieval streams — hybrid pages, takes-keyword, takes-vector, graph walk — and
correctly degrades fail-open with labeled stderr diagnostics. But no typed code
reaches the think result's `warnings[]`. `diagnostics` carries counts and a sanitizer flag but no error signal, so
`pagesFromHybrid: 0` is ambiguous for MCP/remote callers whose stderr goes to
server logs: did the stream error, or was it legitimately empty?

Sibling degradations in the same pipeline already follow the code-only-on-the-wire
idiom ("D6: code-only on the wire; raw exception text goes to server logs"):

- `QUESTION_EMBED_FAILED` — `src/core/think/index.ts:392`
- `CALIBRATION_FETCH_FAILED` — `src/core/think/index.ts:445`
- `TRAJECTORY_INJECTION_FAILED` — `src/core/think/index.ts:527`

Gather's four streams are top-level retrieval arms that degrade fail-open without any typed code.

## Fix

Align gather with the existing mechanism — additive only:

- `ThinkGatherResult` gains `warnings: string[]`; each existing per-stream catch
  additionally pushes a machine-stable code: `GATHER_HYBRID_FAILED`,
  `GATHER_TAKES_KEYWORD_FAILED`, `GATHER_TAKES_VECTOR_FAILED`,
  `GATHER_GRAPH_FAILED`.
- `runThink` folds them into `ThinkResult.warnings` (same fold pattern as the
  `resolveCitations` warnings at `src/core/think/index.ts:695-696`).

Unchanged by design: the fail-open behavior, the stderr diagnostics text, and the
`diagnostics` counts.

## Tests

`test/think-pipeline.serial.test.ts` — new describe with a delegating engine
wrapper (listed methods reject, the rest hit the real PGLite engine):

- clean run reports no gather warnings
- a throwing hybrid stream surfaces `GATHER_HYBRID_FAILED` and stays fail-open
  (takes still gathered)
- each failing stream maps to its own code (all four at once, empty results)
- `runThink` folds gather warnings into `ThinkResult.warnings`

Red-green verified: without the src change the 4 new tests fail (31 pass / 4 fail);
with it, 35 pass / 0 fail.

- targeted suite: 35/35 pass (DATABASE_URL unset, PGLite)
- `bun run typecheck`: clean
- `bun run verify`: 44/44 checks green

### DIFF
--- a/src/core/think/gather.ts
+++ b/src/core/think/gather.ts
@@ -48,6 +48,13 @@ export interface ThinkGatherResult {
   takes: TakeHit[];
   /** Graph nodes — slugs reachable from anchor within graphDepth. Empty when no anchor. */
   graphSlugs: string[];
+  /**
+   * Machine-stable warning codes for per-stream failures (D6: code-only on the
+   * wire; the raw error text goes to stderr/server logs). Lets MCP/remote
+   * callers distinguish "stream errored" from "stream legitimately returned 0"
+   * — the diagnostics counts alone can't. Empty when every stream succeeded.
+   */
+  warnings: string[];
   /** Diagnostics for telemetry / `--explain` path (Lane D follow-up). */
   diagnostics: {
     pagesFromHybrid: number;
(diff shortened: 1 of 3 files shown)


Masashi-Ono0611:
Live data point from the production brain described in my comments on #4120/#4111: I applied this PR's diff to the installed v0.45.11.0 tree (clean `patch -p1` apply, 3 files) and drove `runGather` on the live install with an engine proxy whose every method rejects:

```
[gbrain] searchTitles arm failed (fail-open, title candidates skipped): injected failure for live verification
[think.gather] hybrid stream failed: injected failure for live verification
WARNINGS: ["GATHER_TAKES_KEYWORD_FAILED","GATHER_HYBRID_FAILED"]
PAGES: 0 TAKES: 0
```

The typed codes surface next to the stderr diagnostics (which this PR leaves untouched — the diff is additive-only, +84/-0), with fail-open intact (empty results, no throw). This particular injection activated two of the four codes — the graph arm needs an `anchor` (none was passed), and the takes-vector arm's code did not appear in this run (not root-caused here) — so the remaining two codes are covered by the PR's tests rather than by this injection.

Issue #4177, opened 2026-08-16
fix(context): register reflex telemetry writes before background drain


calebhicks:
## Summary

`logDeliveredReflexPointers()` dynamically imports the volunteer-event sink before it registers the pending INSERT. A caller that immediately invokes `awaitPendingVolunteerEventWrites()` can snapshot an empty pending set, return `unfinished: 0`, and continue teardown before the reflex event is registered or written.

This currently makes two tests fail on clean `v0.46.6.0` / commit `8bf23ab`, even with provider API keys removed from the environment:

- `test/retrieval-reflex.test.ts`: expected one `context_volunteer_events` row, received zero
- `test/volunteer-events-delivery.test.ts`: expected one captured INSERT, received zero

## Reproduction

```sh
env -u OPENAI_API_KEY -u ANTHROPIC_API_KEY \
  bun test test/volunteer-events-delivery.test.ts test/retrieval-reflex.test.ts
```

Result: 28 pass, 2 fail.

The race is deterministic in the focused run:

1. `logDeliveredReflexPointers()` starts `void import('./volunteer-events.ts').then(...)`.
2. `awaitPendingVolunteerEventWrites()` runs before the import callback registers a promise.
3. The drain sees `pendingVolunteerEventWrites.size === 0` and returns success.
4. The event write is scheduled later, outside the drain snapshot and potentially after CLI teardown begins.

## Expected

The fire-and-forget write must be registered synchronously before `logDeliveredReflexPointers()` returns, matching the documented background-work drain contract and the already synchronous hook-delivery path.

## Suggested native fix

Use a static import of `logVolunteerEventsFireAndForget` / `volunteerEventRowsFrom`, or otherwise synchronously register a promise that encompasses the import and eventual INSERT. Add a regression that calls the logger followed immediately by the drain and observes the row.

This issue was discovered while validating an unrelated job-control patch; the failure reproduces without that patch and should be fixed separately.

time-attack:
Fixed in v0.46.23.0 (07f5d28d): logDeliveredReflexPointers registers the volunteer-event write synchronously (static import, retrieval-reflex.ts:545), per the drain contract. Your keyless repro passes 47/47 on master. Reopen if telemetry rows are still lost at shutdown.

PR #4179, opened 2026-08-16
v0.46.6.1 fix(jobs): make job-control dry runs non-destructive

### LINKED ISSUE #4174: pause_job and resume_job mutate state during dry_run and lack mutating metadata
## Summary

Four minion job-control operations omit `mutating: true`. Two of them,
`pause_job` and `resume_job`, also omit the universal `ctx.dryRun` guard used
by adjacent job-control operations. A caller invoking either operation with
`dry_run: true` therefore performs the real pause/resume.

Observed on v0.46.2.0 (`src/core/operations.ts`):

- `cancel_job` and `retry_job` declare `mutating: true`.
- `replay_job` and `send_job_message` honor `ctx.dryRun`, but omit
  `mutating: true`.
- `pause_job` and `resume_job` have `scope: 'admin'`, but neither declares
  `mutating: true` nor checks `ctx.dryRun` before calling `MinionQueue`.
- The operations were introduced together in `d8613366`; searches under
  `test/` find no direct coverage for either operation.

Admin scope still prevents read-scoped callers from reaching these operations,
and the slug-bound-client fence also denies non-read scopes. This is not a
reported authorization bypass. The defect is the false dry-run contract and
incorrect operation metadata.

## Reproduction

Against a disposable brain with a pausable job:

1. Call `pause_job` with the job id and `dry_run: true`.
2. Read the job back.
3. The job is paused instead of remaining unchanged.

The same shape applies to `resume_job`.

## Expected

- All four state-changing operations declare `mutating: true`.
- `dry_run: true` returns a receipt without changing job state.
- Regression tests prove real calls still transition state and dry-run calls do
  not.
- A registry invariant test prevents future state-changing operations from
  silently omitting mutation metadata where the contract can be determined.

calebhicks:
## Summary

- Mark `pause_job`, `resume_job`, `replay_job`, and `send_job_message` as mutating operations so catalogs, bound-client filters, and approval flows see their real behavior.
- Make `pause_job` and `resume_job` honor the operation-layer dry-run contract by returning preview receipts without changing queue state.
- Add queue-backed regressions that prove previews preserve state and real calls still perform the transition.

Closes #4174.

## Scientific reproduction

Before the fix, the focused regression showed:

- all four operations had `mutating === undefined`
- a pause dry run changed `waiting` to `paused`
- a resume dry run changed `paused` to `waiting`

After the fix, the same contract suite passes 17/17 and checks the persisted job status before and after both preview and real calls.

## Verification

- `bun test test/operations-trust-boundary.test.ts` — 17 pass, 0 fail
- release-stamp and bootstrap regressions — 88 pass, 0 fail
- `bun run typecheck` — pass
- `bun run verify` — 50/50 checks green
- `git diff origin/master...HEAD --check` — pass

The full unit lane reached 17,533 passes with 12 failures that reproduce independently of this branch. Ten are the ambient-provider-key isolation class already addressed by #4023; two are the reflex telemetry drain race captured in #4177. Neither failing test imports or exercises the files changed here.

## Adversarial review

- Tried authority expansion: the operations remain `admin`-scoped; no token gains a new capability.
- Checked bound-client behavior: these operations were already non-read by scope, so truthful mutation metadata does not create a new denial class.
- Checked preview parity: the new pause/resume receipts match the existing replay/message dry-run convention.
- Checked false-green tests: regressions assert persisted queue state, not only the returned receipt.

## Release

This is the `0.46.6.1` micro fix. `VERSION`, package/plugin manifests, runbook/template stamps, and the changelog are synchronized.


time-attack:

garrytan:

PR #4180, opened 2026-08-16
v0.46.6.2 fix(cli): wait for complete piped JSON delivery

### LINKED ISSUE #3423: call get_page intermittently truncates captured stdout at 65,536 bytes
## Summary

On Gbrain `0.42.63.0` with Bun `1.3.14` on Linux, `gbrain call get_page`
intermittently exits 0 after writing exactly 65,536 bytes of an incomplete JSON
response to an actively consumed captured stdout pipe. A second call for the
same deterministic synthetic page writes the complete response when stdout is
an unnamed regular temporary file.

## Runnable reproduction

From the repository root:

```bash
scripts/reproduce-gbrain-get-page-pipe-truncation
```

The command creates an isolated disposable PGLite Brain, indexes one
deterministic synthetic page, compares pipe and regular-file capture, prints
only sizes and structural status, and removes the disposable state. It does not
print page content or temporary paths.

## Observed

One reproduced content-free receipt:

```json
{"cleanup":"completed","pipe":{"bytes":65536,"content_complete":false,"exit_code":0,"json_valid":false},"regular_file":{"bytes":102488,"content_complete":true,"exit_code":0,"json_valid":true},"status":"reproduced"}
```

Three consecutive standalone runs produced that exact receipt. A separate
four-run observation on the same versions produced one valid complete
102,488-byte pipe response under concurrent test load, followed by three exact
65,536-byte truncations. The complete-pipe branch is also treated as a valid
observation:

```json
{"cleanup":"completed","pipe":{"bytes":102488,"content_complete":true,"exit_code":0,"json_valid":true},"regular_file":{"bytes":102488,"content_complete":true,"exit_code":0,"json_valid":true},"status":"not-reproduced"}
```

This is intermittent, but the truncated branch consistently stops at exactly
65,536 bytes and still returns exit code 0. The regular-file control remains
valid and content-complete.

## Expected

`gbrain call get_page` should deliver its complete JSON response to an actively
consumed captured stdout pipe before a successful process exit. An upstream
regression should exercise a response above 64 KiB through a pipe while
preserving bounded shutdown.

## Downstream containment

The caller currently uses an unnamed temporary-file capture before applying
strict JSON shape, identity, source, and full-content checks. The downstream
verified synthetic input boundary remains 499,000 bytes; this report does not
request a retrieval-threshold or product-scope change.

calebhicks:
## Summary

- wait on the callback attached to the actual non-empty CLI payload for `--tools-json` and formatted local/remote operation results
- fail nonzero after a bounded timeout or stream error instead of reporting success after partial delivery
- pin the behavior with a real producer-to-slow-consumer kernel-pipe regression

Closes #3423.

## Reproduction and verification

Before the fix, a deliberately backpressured `gbrain --tools-json` pipeline consistently produced this content-free receipt:

```text
producer=0 consumer=1 bytes=65536 valid_json=false
```

Raising the existing flush grace to five seconds did not change the result. The empty `write('', callback)` fence was being elided by Bun; attaching the callback to the actual non-empty payload is the native delivery signal that survives backpressure.

After the fix:

- the slow-consumer regression receives the exact same 118,818 bytes as a regular-file control and parses valid JSON
- `gbrain --tools-json | jq empty` succeeds 12/12 times
- a consumer that never reads makes the producer fail nonzero after the two-second bound
- focused tests: 30 passed, 0 failed (65 assertions)
- `bun run typecheck`: passed
- `bun run verify`: 50/50 checks passed

The complete unit suite was also run: 17,539 passed and 21 skipped. Two known failures reproduce the separately filed reflex-telemetry drain race (#4177). One unrelated repository-durability test failed once in the serial lane and passed 23/23 when rerun in isolation. All tests added or changed by this branch are green.

## Adversarial review

- attacked silent false success: timeout and EPIPE reject, so the central CLI catch owns the nonzero verdict
- attacked byte drift: the integration test compares the slow pipe's byte count to an independent regular-file control, then requires valid JSON
- attacked test-induced buffering: the producer and consumer are connected directly by Bash, not through a parent-process stream buffer
- attacked hung consumers: the write wait is bounded by the existing two-second flush guard
- attacked payload disclosure: the slow-consumer fixture emits only byte count and JSON validity, never the payload


time-attack:

garrytan:
Closing as superseded: the 64KiB piped-stdout truncation (#3423) was fixed in v0.46.23.0 (07f5d28d). Thank you — verified during the Wave K triage sweep at 975886a3.

PR #4181, opened 2026-08-16
v0.46.6.3 fix(context): register reflex telemetry before drain

### LINKED ISSUE #4177: fix(context): register reflex telemetry writes before background drain

calebhicks:
## Summary

- replace the reflex delivery path's dynamic import with synchronous registration in the existing volunteer-event sink
- keep the INSERT fire-and-forget, bounded by the existing background-work drainer, and fail-open
- move the shared reflex rationale into the telemetry module so the static dependency remains acyclic at runtime

Closes #4177.

## Red-to-green evidence

On clean v0.46.6.0, immediately logging a delivered reflex pointer and draining the sink produced two deterministic failures:

```text
28 pass, 2 fail
expected context_volunteer_events rows: 1
received: 0
```

The regression now asserts that the pending-write count is already one before `logDeliveredReflexPointers()` returns. After the fix:

- focused reflex/telemetry tests: 30 passed, 0 failed (92 assertions)
- broader reflex/volunteer surface: 81 passed, 0 failed (235 assertions)
- typecheck: passed
- release verifier: 50/50 checks passed
- exhaustive suite: 18,980 passed, 21 skipped, 0 failed (1,228 main-shard files plus 142 process-isolated files)

## Adversarial review

- attacked late registration: the test peeks the real sink synchronously before invoking the drain
- attacked runtime import cycles: `volunteer-events.ts` retains only a type-only dependency on `retrieval-reflex.ts`; the runtime dependency is one-way
- attacked delivery latency: the INSERT remains fire-and-forget; only promise registration is synchronous
- attacked telemetry failures leaking into context delivery: synchronous assembly errors and asynchronous INSERT failures remain swallowed
- attacked double-counting: logging remains at the existing accept-side delivery seam; the resolver itself still writes nothing


time-attack:

garrytan:
Closing as superseded: the reflex-telemetry late-registration race (#4177) was fixed in v0.46.23.0 (07f5d28d). Thank you — verified during the Wave K triage sweep at 975886a3.

Issue #4182, opened 2026-08-16
smoke-test: replace GNU-only grep -P for macOS portability


calebhicks:
## Summary

`scripts/smoke-test.sh` extracts the doctor score with `grep -oP`, but the BSD `grep` shipped by macOS does not support `-P`. A native `gbrain smoke-test` therefore emits a noisy `grep: invalid option -- P` diagnostic on macOS even when the doctor check itself succeeds.

## Reproduction

Run `gbrain smoke-test` on macOS with the normal BSD grep. The database check prints a fallback `?/100`, while stderr contains the unsupported-option error.

## Expected

Use a POSIX-portable extraction path (`sed`, `awk`, or shell parsing) so the shipped smoke test is clean on every supported host.

Discovered while adding a hermetic regression for #4175; it is independent of worker detection and should remain a separate fix.

time-attack:
Automated triage: **reproduced on current master** — labeling `triage:bug-verified`.

Confirmed at `scripts/smoke-test.sh:98`:

```bash
SCORE=$(echo "$DOCTOR_OUT" | grep -oP 'Health score: \K[0-9]+' || echo '?')
```

`grep -P` is a GNU extension. macOS ships BSD grep, which rejects `-P` outright, so the `|| echo '?'` swallows it and the smoke test silently reports `SCORE=?` on every Mac rather than failing loudly. Same class as the `timeout`-is-GNU-only portability gap.

`sed -E` or `awk` would be the portable replacement here. Left open for a fix pass.

PR #4183, opened 2026-08-16
v0.46.6.4 fix(smoke): respect native worker supervisors

### LINKED ISSUE #4175: smoke-test auto-fix starts a duplicate shell-enabled worker on supervised installs
## Summary

`gbrain smoke-test` decides whether a worker exists solely by checking
`/tmp/gbrain-worker.pid`. On an install whose workers are managed by the native
supervisor/launchd path and use per-brain supervisor pidfiles, the check reports
the healthy worker missing and auto-starts a second unmanaged worker.

The duplicate is launched with `GBRAIN_ALLOW_SHELL_JOBS=1`, concurrency 2, a
global `/tmp/gbrain-worker.pid`, and no per-brain supervisor contract. This is
unsafe for shared or multi-brain hosts and makes the documented native smoke
test unusable there.

Observed on v0.46.2.0 and confirmed unchanged on current v0.46.6.0 `main`.

## Reproduction

On a host with a healthy supervisor-managed worker:

1. Confirm the brain's native supervisor pidfile exists and its process is
   alive.
2. Confirm `/tmp/gbrain-worker.pid` does not exist.
3. Read or run test 4 in `scripts/smoke-test.sh`.

The script takes the auto-fix branch and starts another `jobs work` process.
On a four-brain host, all four managed supervisor pidfiles can be present while
the single legacy `/tmp/gbrain-worker.pid` is absent.

## Expected

- The smoke test recognizes the native supervised-worker topology, including
  per-brain pidfile configuration, before attempting any fix.
- A read-only smoke invocation never creates an unmanaged duplicate worker.
- Any worker-start repair is explicit, topology-aware, and preserves the
  configured shell-job policy rather than forcing it on.
- Regression coverage models a healthy supervisor-managed worker with no
  `/tmp/gbrain-worker.pid` and proves no process is spawned.

calebhicks:
## Summary

- ask `gbrain jobs supervisor status --json` for worker health, inheriting its brain-scoped PID-file and live DB-lock detection
- retain legacy bare-worker PID detection for compatibility, while flagging supervisor-plus-legacy as a duplicate topology
- remove the smoke test's automatic unmanaged `jobs work` launch and its unconditional shell-job enablement
- document the explicit native repair: `gbrain jobs supervisor start --detach`
- replace GNU-only `grep -P` health-score extraction with POSIX `sed` for macOS

Closes #4175 and #4182.

## Red-to-green evidence

The hermetic regression runs the shipped shell script with a fake gbrain CLI that reports a healthy supervisor and no legacy PID file.

Before the fix, the script ignored supervisor status, invoked `jobs work`, wrote a legacy PID file, and failed after that unmanaged process exited.

After the fix:

- healthy supervisor, no legacy PID: passes without invoking `jobs work`
- no supervisor and no legacy PID: fails with the explicit native supervisor command and creates nothing
- healthy supervisor plus live legacy PID: fails as a duplicate topology and creates nothing
- the macOS/BSD toolchain reports the parsed doctor score without an unsupported `grep` warning
- focused tests: 3 passed, 0 failed (14 assertions)
- typecheck: passed
- shell syntax check: passed
- release verifier: 50/50 checks passed

## Adversarial review

- attacked split-home/per-brain installs: detection delegates to the existing native status surface, including its queue DB-lock fallback
- attacked silent authority expansion: the smoke path no longer starts any daemon or enables shell jobs
- attacked duplicate masking: a live native supervisor and live legacy PID produce a failure, not a green result
- attacked stale PID reuse: the duplicate message tells the operator to verify ownership before stopping anything; the smoke test never sends a signal
- attacked test interference: regressions override the legacy PID path and Bun path inside a temporary directory and never touch global PID files or live processes
- attacked platform drift: the harness asserts both the parsed score and clean stderr using the shipped macOS toolchain


time-attack:

garrytan:

PR #4186, opened 2026-08-16
fix(config): let DB-plane eval.* reach the runtime gate (#1475)

Masashi-Ono0611:
`gbrain config set eval.capture true` is accepted, persisted, and read back by
`gbrain config get` as `true` — and changes nothing. Capture stays off unless
`GBRAIN_CONTRIBUTOR_MODE=1` is also exported. Reported in #1475 in May, and
re-confirmed there on 0.46.1.0 on 2026-08-15.

**Part of #1475** — this closes the CLI half only. See "What this does not fix".

## Two halves, and why fixing one is not enough

1. `loadConfigWithEngine` merges a fixed set of DB-plane keys. It had no
   `eval.*` branch. Both keys are in `KNOWN_CONFIG_KEYS`, which is why the
   write is accepted in the first place, and `config get` queries
   `engine.getConfig` directly, which is why it reads back.

2. `makeContext` built `ctx.config` from the sync, file-only `loadConfig()`.
   `connectEngine` *does* re-merge after connect — but it keeps the result to
   itself (env stashes + gateway reconfigure) and returns only the engine. The
   gate the runtime consults is `isEvalCaptureEnabled(ctx.config)`.

Measured by reverting each half against the new runtime test file (7 cases):

| tree | result |
|---|---|
| merge branch only, no plumbing | **5 fail** — including all 3 runtime-symptom cases |
| plumbing only, no merge branch | **4 fail** |
| both | 7 pass |

So a patch that adds the merge branch alone — the natural reading of the issue
— looks right and does not move the symptom. That is why the test asserts on
`isEvalCaptureEnabled(ctx.config)` rather than on either half's output.

## No new config reads on the host brain

Open PR #3980 reports 54 queries for a single `gbrain stats` against a hosted
brain, 44 of them `SELECT value FROM config WHERE key = $1`, and attributes
~1.5s of the 7.7s to `loadConfigWithEngine`'s per-key reads. Re-merging inside
`makeContext` would have doubled the merge's share of that. Instead
`connectEngine` publishes the merge it already performs on a `WeakMap` keyed by
engine, and `makeContext` consumes it.

Measured here: **25 config reads uncached, 1 cached** — the survivor being the
source resolver's `sources.default`, which is not merge work. The test asserts
the cached key set (`['sources.default']`) rather than a bare count, so a
future reader can tell an intentional addition from a regression.

**Two cases still pay the fallback reads.** `connectEngine` returns a mounted
engine (`--brain <mount>`) before reaching the publish, so mounted commands
fall through to the merge in `makeContext`. And `connectEngine` wraps its
merge in a non-fatal `catch`, so if the merge itself throws there the map stays
empty and `makeContext` retries — correct either way, just not free. That
is also why the map is keyed by engine rather than being a module global: it
keeps a mount on the correct fallback instead of serving it the host brain's
merged config.

## Strict boolean parsing on these two keys

`dbBool` maps every non-empty value other than the exact string `'true'` to
**false**. Measured on this tree, with the value stored by `config set` and
read back through the merge: `'tru'`, `'TRUE'`, `'1'`, `'yes'` all arrive as
`false`.

Routing `eval.scrub_pii` through it would mean `gbrain config set
eval.scrub_pii TRUE` silently **disables PII scrubbing** — a footgun this
change would newly activate, since those values were inert while no merge
branch existed. So `eval.*` parses strictly: anything that is not exactly
`true`/`false` is treated as unset and the documented default applies.

Deliberately scoped to the two keys added here. Five other keys still use the
loose helper (`embedding_multimodal`, `embedding_image_ocr`, and three
`content_sanity.*`); tightening those changes behavior for existing brains and
belongs in its own PR. Flagging it rather than folding it in.

## What this does not fix

**MCP, `gbrain call`, and background subagents still read the file plane only.**
`src/mcp/dispatch.ts:378` builds its context with `loadConfig()`, and
`src/core/minions/handlers/subagent.ts:196` does the same. That matters here
because #1475's own workaround is a systemd drop-in on `gbrain-mcp.service` and
`gbrain-worker.service` — so the reporter's deployment is not covered by this
PR.

I did not extend it there because the fix shape is different and the choice is
yours: `buildOperationContext` is synchronous, and an MCP server is long-lived,
so "merge once at boot" trades the bug for staleness after a later `config set`.
That is a policy question about how often a long-lived process re-reads DB
config, not a defect. Happy to implement whichever way you want it.

The same staleness caveat applies to the WeakMap here in principle; it does not
bite today because a CLI invocation builds one context and exits.
`docs/eval-capture.md` now says which surfaces honor the DB plane and which do
not, instead of stating flatly that `config set` does not work.

## Tests

`test/eval-capture-db-plane.serial.test.ts` (new, 7 cases) — the runtime
composition: DB-only `true` turns the gate on with `CONTRIBUTOR_MODE` unset; DB
`false` turns it off *with* `CONTRIBUTOR_MODE=1`; a control proving capture
stays off with no DB value (so the first case cannot pass on a build that
enables capture unconditionally); `scrub_pii` on the same path; the read-count
assertion; a source-level guard on the publish; and fail-open when the config
table is missing.

`test/loadConfig-merge.test.ts` (+7 cases) — the merge contract: both
directions, file-over-DB precedence, no spurious `cfg.eval` container,
non-fatal on a throwing `getConfig`, and the strict-parsing table (7 rejected
spellings plus `true`/`false` controls, so strictness cannot be satisfied by
ignoring the keys).

One weak spot, stated plainly: the read-count case primes the WeakMap through a
`__testing` seam, so it pins the consumer, not `connectEngine`'s publish. The
companion assertion for the publish is a source-string check — it fires if the
line is deleted (verified), but a comment containing the same literal would
satisfy it. `connectEngine` is neither exported nor runnable without a live
brain, so I did not find a behavioural way to cover it without changing that
function's signature; if you would rather it were exported for test access, say
so and I will.

`bun run verify` 50/50 and `bun run typecheck` clean locally; CI is the
authority.

### DIFF
--- a/docs/eval-capture.md
+++ b/docs/eval-capture.md
@@ -155,6 +155,16 @@ Resolution order (most explicit wins):
 `eval.scrub_pii: false` to preserve raw query text (only if you control
 the brain's distribution).
 
-`gbrain config set eval.capture false` does **not** work — that
-command writes the DB-plane config, and the MCP server reads the
-file-plane. Edit the JSON directly or use the env var.
+`gbrain config set eval.capture <true|false>` applies to **CLI commands**
+(`gbrain query`, `gbrain recall`, …): the CLI merges the DB plane into the
+operation context it hands the capture gate.
+
+It does **not** yet apply to the **MCP server, `gbrain call`, or background
+subagents** — those build their context from the file plane only, so a
+DB-plane value is stored, echoed back by `gbrain config get`, and ignored.
+For those, edit `~/.gbrain/config.json` directly or export
+`GBRAIN_CONTRIBUTOR_MODE=1` in the service environment.
+
+Only the exact strings `true` and `false` are recognised on the DB plane.
+Anything else (`TRUE`, `1`, `yes`, a typo) is treated as unset and the
+default applies — it does not silently mean "false".
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -1281,6 +1281,32 @@ export function findUnknownOpFlag(op: Operation, args: string[]): string | null 
   return null;
 }
 
+/**
+ * #1475 — the DB-plane merge `connectEngine` performs after connect, published
+ * for `makeContext` so the operation context carries it too.
+ *
+ * Why a map rather than a module-level variable: a process can hold more than
+ * one engine (`--brain <mount>` connects a second one), and the host brain's
+ * merged config must never be served for a mounted brain's context. Weak so a
+ * disconnected engine does not pin its config for the life of the process.
+ */
+const MERGED_CONFIG_BY_ENGINE = new WeakMap<BrainEngine, GBrainConfig>();
+
+/**
+ * @internal Exported for test/eval-capture-db-plane.serial.test.ts.
+ *
+ * Publishing the merge is the whole point of the map — if the `set` in
+ * connectEngine is ever dropped, makeContext silently falls back to
+ * re-merging and every command pays the per-key reads twice with no test
+ * going red. The seam lets a test prime the map the way connectEngine does
+ * and observe that no further reads happen.
+ */
+export const __testing = {
+  publishMergedConfig(engine: BrainEngine, config: GBrainConfig): void {
+    MERGED_CONFIG_BY_ENGINE.set(engine, config);
+  },
+};
+
 export async function makeContext(engine: BrainEngine, params: Record<string, unknown>): Promise<OperationContext> {
   // v0.31.8 (D11): resolve sourceId via the canonical 6-tier chain. Honors
   // --source / GBRAIN_SOURCE / .gbrain-source / path-match / brain default /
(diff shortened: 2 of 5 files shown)


time-attack:
Automated triage: **routing to maintainer** — tests are green but adversarial review found a likely production regression.

Floor gate passed: targeted tests 7/7, `bun run verify` clean, CI green.

**The concern.** `src/cli.ts:1349-1351` states the fallback re-merge path only runs for "tests with a stub engine." That looks incorrect for **mounted brains**:

- `src/cli.ts:3062` — `connectEngine` early-returns `connectMountEngine(brainId)` for any non-host brain
- `src/cli.ts:3136` — `MERGED_CONFIG_BY_ENGINE.set(engine, merged)` sits in the host branch only, so mounts never publish
- `src/cli.ts:1363` — mount engines therefore always miss the cache and take the full re-merge

Measured on this branch: ~24 additional `getConfig` round-trips per `makeContext` on an unpublished engine (`embedding_*`, `content_sanity.*`, `dream.*`, `eval.*`). Master's equivalent (`src/cli.ts:1315`) is a synchronous `loadConfig()` with zero DB reads, so on a remote Postgres brain this adds roughly a second per command.

The `src/core/config.ts` merge itself looks well-scoped to the eval namespace and precedence is not widened — the concern is confined to the cli.ts wiring. Publishing the merged config in `connectMountEngine` too (or narrowing the fallback) would likely resolve it.

Not merging automatically. Flagging for a human call since this is a hot path.

Masashi-Ono0611:
Pushed a fixup addressing this (568c90a6, CI green): the fallback re-merge in `makeContext` was indeed reachable far beyond the stub-engine case — as you measured, a mounted-brain engine always missed the cache and took the full per-key re-merge on every command that reaches `makeContext`, since `connectMountEngine` never published to `MERGED_CONFIG_BY_ENGINE`.

Fix shape: `connectMountEngine` now registers its engine in a `MOUNT_ENGINES` WeakSet (the same engine-keyed pattern as `MERGED_CONFIG_BY_ENGINE`), and the fallback skips any engine in that set. Mounted brains keep the file/env-only config `makeContext` built pre-#1475 — zero additional config-merge DB reads, matching master. Host brains are unchanged, so DB-plane `eval.*` still reaches the runtime gate.

This also closes a trust gap the re-merge would have opened: `connectMountEngine`'s docstring promises a mount's DB-plane model config is never merged into the caller's AI gateway, but the #1475 fallback would have merged the mount's DB-plane config into the caller's `ctx.config`. Skipping the merge for mount engines keeps the caller's op-side gates governed by the caller's own config.

Added a regression test pinning the skip (no DB-plane merge, no config leak for a mount-marked engine) plus a source-level wiring guard on the `MOUNT_ENGINES.add(handle.engine)` line, and rewrote the comment block that claimed the fallback "only runs for tests with a stub engine."

Masashi-Ono0611:
Screenshot of the live run backing the comment above — the PR's two test files executed on the installed tree with this PR's diff applied: 39 pass / 0 fail.

live run: 39 pass / 0 fail on the installed tree

time-attack:

